PR #150, opened 2016-02-22
Add ready deque to scheduler

### LINKED ISSUE #148: New Worker not used by previously submitted map
To roughly reproduce, set up a scheduler on one machine with `dscheduler`, start up several workers with `dworker`, and then do

```python
from distributed import Executor

executor = Executor('larry:8786')
L = executor.map(long_running_function, inputs)
```
where the number of `inputs` is greater than the number of processes available among the workers. 

Adding an additional worker with `dworker` later is recognized by the executor, but none of the pending operations appear to be sent to that new worker. New submissions using e.g. `Executor.map`, however, do make use of the new worker.

Not sure if this is technically a bug, but it would be cool if adding more workers helps move pending operations from a `map` further along.

mrocklin:
Previously we assigned a task to a worker once that task was ready to
run.  It would then linger in that worker's stack, possibly for a long
time.  If in the future another worker capable of running that task
became free we had no mechanism for it to steal work.  There was good
reason for this, it's tricky to find good tasks to steal in constant
time.

However, in the common case of leaf tasks that have neither dependent
tasks nor restrictions it makes sense to keep these ready tasks in a
common pool from which all workers can draw.  This is the new `ready`
collection on the scheduler, stored as a FIFO deque.

This enables workers to load balance more effectivley in this common
case and provides newly arrived workers with a pool of available work.

Fixes #148 

Screencast of things working:  https://www.youtube.com/watch?v=x7ehRDYBing&feature=youtu.be

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -11,7 +11,7 @@ from tornado.ioloop import IOLoop
 from tornado import gen
 
 from .core import Server, rpc, write
-from .utils import get_ip
+from .utils import get_ip, ignoring
 
 
 logger = logging.getLogger(__name__)
@@ -201,4 +201,5 @@ def run_worker(q, ip, center_ip, center_port, ncores, nanny_port,
             q.put({'port': worker.port, 'dir': worker.local_dir})  # pragma: no cover
 
     loop.add_callback(start)  # pragma: no cover
-    loop.start()  # pragma: no cover
+    with ignoring(KeyboardInterrupt):
+        loop.start()  # pragma: no cover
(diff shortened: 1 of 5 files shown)


mrocklin:
@dotsdl can you see if this branch works well for you?

dotsdl:
Very nice! Giving it a spin shortly. I'll let you know by this evening if I see anything unexpected.

dotsdl:
@mrocklin on first play it looks like it's working as expected. `Executor` instances also show the right number of workers at any given time, from any given python session. I've got plenty of work to throw at this, so I'll see if I notice any issues over the next couple of days. Thanks!

mrocklin:
OK.  I plan to merge this then and we can work from there.

PR #153, opened 2016-02-23
Make sure worker selected IP is on the same network as scheduler

ogrisel:
If the dworker host has many network interfaces with complex setups (e.g.
docker swarm or VM bridge networks) it is unlikely that the worker will select
the interface that has an address that is visible from the scheduler. Instead,
we change the `get_ip` function to be able to target the center ip address to select
an ip that is visible from it automatically.

Alternatively make it possible to explicitly pass a specific IP or hostname to
the dworker commandline argument.

### DIFF
--- a/distributed/cli/dworker.py
+++ b/distributed/cli/dworker.py
@@ -13,15 +13,14 @@ from tornado import gen
 
 logger = logging.getLogger('distributed.dworker')
 
-ip = get_ip()
-
 
 @click.command()
 @click.argument('center', type=str)
 @click.option('--port', type=int, default=0,
               help="Serving port, defaults to randomly assigned")
 @click.option('--host', type=str, default=None,
-              help="Serving host defaults to %s" % ip)
+              help="Serving host. Defaults to an ip address that can hopefully"
+                   " be visible from the center network.")
 @click.option('--nthreads', type=int, default=0,
               help="Number of threads per process. Defaults to number of cores")
 @click.option('--nprocs', type=int, default=1,
@@ -29,7 +28,8 @@ ip = get_ip()
 @click.option('--no-nanny', is_flag=True)
 def main(center, host, port, nthreads, nprocs, no_nanny):
     try:
-        center_ip, center_port = center.split(':')
+        center_host, center_port = center.split(':')
+        center_ip = socket.gethostbyname(center_host)
         center_port = int(center_port)
     except IndexError:
         logger.info("Usage:  dworker center_host:center_port")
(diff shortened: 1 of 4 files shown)


distributed/cli/dworker.py
```diff
@@ -41,16 +41,19 @@ def main(center, host, port, nthreads, nprocs, no_nanny):
     if not nthreads:
         nthreads = _ncores // nprocs
 
-    if nprocs > 1 and port != 0:
-        raise ValueError("Can not specify a port when using multiple processes")
-
```
mrocklin: This check still seems valid to me.  I should add more tests around the command line interface.

distributed/utils.py
```diff
@@ -31,10 +30,9 @@ def funcname(func):
         return str(func)
 
 
-def get_ip():
-    return [(s.connect(('8.8.8.8', 80)), s.getsockname()[0], s.close())
-        for s in [socket.socket(socket.AF_INET, socket.SOCK_DGRAM)]][0][1]
-
+def get_ip(host='8.8.8.8', port=80):
+    return [(s.connect((host, port)), s.getsockname()[0], s.close())
+            for s in [socket.socket(socket.AF_INET, socket.SOCK_DGRAM)]][0][1]
```
mrocklin: Nice.

mrocklin:
Thank you for this.  I hadn't thought of the more complex case.  

Can you think of a convenient way to test this sort of change?  "No" is a fine answer.  Happy to merge regardless.

distributed/cli/dworker.py
ogrisel: Yes but this is a near duplicate of a similar block 10 lines above it this can never be executed.

mrocklin: Ah, indeed you're correct.

mrocklin:
+1.  Merging soon if no comments.

ogrisel:
> Can you think of a convenient way to test this sort of change? "No" is a fine answer. Happy to merge regardless.

I have a [docker-compose.yml config](https://github.com/ogrisel/distributed-compose) that might be adapted to run on a single docker daemon (by removing the overlay network) on a travis docker service.

FYI I have been testing this on a [Rackspace Carina](getcarina.com) cluster. I have another issue now that seems unrelated to distributed but used to work yesterday...

I would be nice to write a demo kubernetes config to run on openshift / Google Compute Engine / Azure too.

distributed/utils.py
ogrisel: BTW I think we might want to fallback to `socket.gethostbyname(socket.gethostname())` if `host` is left to the default `8.8.8.8` and is unreachable (not connected to internet). But this should probably be addressed in a separate PR.

ogrisel:
> FYI I have been testing this on a Rackspace Carina cluster. I have another issue now that seems unrelated to distributed but used to work yesterday...

I just tried again and it works perfectly now. It was probably a bug at the Carina infra level.

PR #155, opened 2016-02-25
WIP: S3 chunks

martindurant:
Reading in chunks and reading delimited works.

Marked as WIP for the following reasons:
- boto3 objects cannot be shared between threads, apparently, so have to be re-established for each task. Together with repeated calls when finding delimiter boundaries, this makes small reads very slow.
- if block size is small compared to the typical record size, you get many blank responses (i.e., `b''`) - is this desirable?
- needs some sanity checking on what should be expected behaviour for possible offsets etc.
- no file-like layer implemented. That may be useful.
- read CSV is an obvious next step


mrocklin:
* What is the rough cost for a single boto read?  
* Does the caching in `get_s3` help here or is this cost associated to setting up a read rather than a connection?  FWIW I've found that this cost this tends to be a lot less painful when on EC2. 
*  I think that the small block size case will be rare.  If records are too large then I think that blank responses is appropriate.  I expect blocksizes to be in the 100-500 MB range
*  Hooray for sanity checks

martindurant:
Any idea what's causing failures in some builds and not others?

mrocklin:
The latter two builds only test the HDFS tests.  

The error that we're seeing looks like it's due to using tornado unsafely in two threads.

mrocklin:
Travis.ci machines are *much* slower than your personal machine, and so catch concurrency bugs much more effectively.

mrocklin:
`v.compute()` will call `e.get`, which is synchronous.  Inside of the testing coroutine you'll need to use the asynchronous API, I recommend the following:

```python
futures = e.compute(values)
results = yield e._gather(futures)
```

mrocklin:
Sorry for the dual API.  Tornado coroutines are really awesome, but they also steepen the learning curve.

distributed/s3.py
```diff
@@ -65,6 +66,66 @@ def get_list_of_summary_objects(bucket_name, prefix='', delimiter='',
...
+        try:
+            return ob.get(Range="bytes=%i-%i" %
+                          (offset, offset + length - 1))['Body'].read()
+        except ClientError:
+            return b''
+    start = seek_delimiter(ob, offset, delimiter)
+    end = seek_delimiter(ob, offset + length, delimiter)
+    if start == end:
+        return b''
+    return ob.get(Range="bytes=%i-%i" % (start, end - 1))['Body'].read()
```
mrocklin: If each individual read is expensive then it might make sense to initially read most of the bytes straight away with a big buffer on either side.  E.g., if I ask for the byte range `1000000->2000000`, then perhaps we just read `900000->2100000` straightaway and only go seeking if we don't find a delimiter.

martindurant:
Specific answer:
Every access to S3 seems to take at least 200ms (from my machine); I think the boto3 connection object only stores the metadata, not the connection. The open file objects do store more, including an unserializable lock object, so I am creating them within the tasks rather than passing them around.

distributed/s3.py
```diff
@@ -65,6 +67,66 @@ def get_list_of_summary_objects(bucket_name, prefix='', delimiter='',
...
+            piece = ob.get(Range="bytes=%i-%i" %
+                           (offset, offset + seeklength))['Body'].read()
+        except ClientError:
+            return ob.get()['ContentLength']
+        data += piece
+        try:
+            i = data.index(delimiter)
+            return offset + i + len(delimiter)
+        except ValueError:
+            offset += seeklength
```
mrocklin: Does this handle all of the cases that `hdfs3.utils.seek_delimiter` handles?  I sometimes get a bit nervous whenever I see two copies of difficult logic.

mrocklin:
Those are really useful numbers to know.  Once we have an open s3 file object presumably repeated requests of the same file are relatively cheap?

martindurant:
No, unfortunately not - each `get()` with a new bytes range incurs this cost. Streaming reads from one of these is then fast (the HTTP channel stays open), but you can only read sequentially, not seek.

mrocklin:
OK, so we should expect 200ms overhead per future, regardless of if we have to read a little bit more unexpectedly to get to the next endline.  Is my understanding correct?

mrocklin:
Oh, nevermind, you said each call to `get`, which appears to also take a fixed length.

mrocklin:
I think that my approach to this problem would be something like the following:

Given approximate desired range a to b

*  Get range a to b + delta
*  Shove bytes into a BytesIO
*  call `hdfs3.utils.seek_delimiter` on the bytesio with the range a to b and the line delimiter
*  If we get to the end of the BytesIO file (no delimiter found) then pull in a much larger delta buffer

martindurant:
This is not how `read_block` in HDFS3 works now, but I suspect that that benefits from at least OS-level disk read caching. Is it time yet to consider a pipeline-oriented refactor:
- set of providers of (delimited) blocks of bytes (or BytesIOs), at least one block per input where any metadata/header is in one of those blocks (S3, HDFS, HTTP, g/luster ...)
- set of compressions
- set of bytes to data-structure interpreters (csv, json, avro, parquet ...)

mrocklin:
Sure.  I guess my thought here is that `read_bytes` is the first step of that.  I think I agree that after we have a couple of fully compatible `read_bytes` functions then we might want to pull out the common logic for file formats.  This hasn't been possible before because hdfs was blocked and s3 wasn't.

Regarding the difference between hdfs3.read_block and what is presented above: I think that if we can create a file interface where seek and read are both somewhat cheap then that's best.  What I'm hearing here is that that isn't the case for S3.  In that case I think we can efficiently emulate the file interface in the common case by reading a bit extra and using BytesIO.  This allows us to reuse error-prone code (I really don't want to maintain multiple copies of the endline seeking stuff for the next few years) and is, I think, near-optimally efficient.

martindurant:
@mrocklin , the new file-like object works, and caches data for minimal number of accesses. `read_bytes` does not yet depend on it - sending this now to see if it's what you had in mind.

`hdfs3.utils.read_block` can successfully read records (tested manually); but to use the very same code would require moving things around. I'm not sure how to integrate this.

A natural extension of the file object would be a filesystem object, as for hdfs3, and then fuse integration. Is there any value in these? (e.g., instead of an s3-specific `read_bytes` calling `get_list_of_summary_objects`, a generic `read_bytes` calls `ls` on a filesystem object)

A couple of suggestions for delimited reading
- `seek_delimiter` should bail after some given length: if there is no delimiter within the block of interest, there's no point finding it further in the file. There might not be any delimiter at all, and then the whole file would get scanned twice before returning an empty block
- `read_block` should include the final delimiter, since some bytes readers like avro will specifically look for it (for CSV, as read by pandas, it doesn't matter)

mrocklin:
I think that the S3 file interface looks great!  It's reassuring that the `hdfs3.utils.read_block` function works out-of-the-box.

For now my suggestion would be to just copy-paste `hdfs3.utils.read_block` to `distributed.utils.read_block` and deal with code duplication.  Maybe we find a nice way to rearrange things in the future after we have a few of these.  Maybe at some point we remove the `HDFile.read_block` method?

A file system object for S3 would make me personally happy.  I rarely remember how to use boto a day or two after I relearn it to solve some problem.  For the moment my inclination is to let this linger unless we've identified a pressing need for it.  Perhaps raise an issue?  If we go this route eventually it might make sense to break off the functionality to a separate project a la `hdfs3`.

> seek_delimiter should bail after some given length: if there is no delimiter within the block of interest, there's no point finding it further in the file. There might not be any delimiter at all, and then the whole file would get scanned twice before returning an empty block. 

I'm not overly concerned about this case.  In the common case of JSON, CSV, Avro I think we'll run into delimiters fairly frequently.  I don't mind being inefficient in the uncommon case.  I do mind being incorrect in the uncommon case.  We would need to add a bit of complexity to handle cases where we bailed out incorrectly.

> It would be easy to add an HTTP file object, but there a filesystem would be hard (no directory listing)

Agreed, that would be cool.  HTTP supports starting a GET request at a particular byte, right? Suggest we wait a bit until the S3 work firms up.

> read_block should include the final delimiter, since some bytes readers like avro will specifically look for it (for CSV, as read by pandas, it doesn't matter)

OK, I hadn't thought of that.  Presumably this will come up in writing tests for things like Avro?

martindurant:
Added in HTTPFile object which works just like the S3 one. This will need some test data - I was playing, for example, with `https://raw.githubusercontent.com/mwaskom/seaborn-data/master/raw/exercise.csv` .

Added in the functions from hdfs3.utils - can plug in now and make the s3 case look more like the hdfs case (considering how long a filesystem would take). The http case will only work with an explicit list of URLs.

mrocklin:
Cool.  For testing you might consider using a mock like `https://pypi.python.org/pypi/requests-mock` or even the stdlib Python http server. 

My preference of course would be to see the existing functionality solidified. well tested, and merged before branching out to new things like filesystem abstractions.  In particular it would be really cool to get the S3 stuff merged quickly so that we can point people to fun datasets with `dec2`.

mrocklin:
There is also moto, if you wanted to test against S3 without S3.  We didn't use this for the original effort, because moto didn't play nicely with tornado, but that's not such a big deal here.  https://github.com/spulec/moto

Not sure that's worth it though if you already have a testing solution that works for you.

martindurant:
@mrocklin : I decided the most elegant way to proceed was to make the S3FileSystem after all. Here is a version that works by interactive checking. However, although `read_block` works using the code from hdfs3, `read_bytes` does not work out of the box, as it makes some hdfs-only assumptions. Would you like to help with with making `read_bytes` backend agnostic? Then the very same tests as for hdfs (the file system and the distributed part) should work here too - but we would need a bucket with write privileges.

One note: writing uses multipart upload, which allows for multiple writes (ended by a call to `flush()` or `close()`), but this requires that every part is >5MB in size. The alternative would be to only ever allow a single write per open file; other more complex alternatives here https://github.com/boto/boto3/issues/256

mrocklin:
My guess is that `read_bytes` will share a common interface among storage technologies but won't be backend agnostic.  My guess is that it will need to be particular to each storage system.  For example in systems like HDFS we need to add restrictions to run tasks on the correct nodes.  Also the implementations are fairly small (`hdfs.read_csv` is 29 lines) so I don't see a whole lot of benefit from adding super-structure here.

I do like the idea of adding generic tests to define how `read_bytes` functions should work.  One approach would be to make an old-fashioned `unittest.TestCase` class.

This PR is getting fairly old and has a large amount of disparate functionality within it.  This has me worried about it going stale.  I think that it's time to test, solidify, and merge in some of the base parts otherwise we'll end up with something sufficiently out of date that it will never get merged in.  At this stage it might also be worthwhile splitting some of these contributions into separate PRs just so that we can start merging the more mature parts.   Perhaps the functionality could be split into multiple PRs of `s3.read_bytes`, `http read_bytes` and `s3 file system`.

I recommend starting with tests for `s3.read_bytes` without trying to merge things in a clever way.  Just straight and simple tests in a new PR so that we can get `s3.read_bytes` in happily.

distributed/tests/test_s3.py
```diff
@@ -27,59 +27,112 @@
...
+@gen_cluster(timeout=60)
+def test_read_bytes(s, a, b):
+    e = Executor((s.ip, s.port), start=False)
+    yield e._start()
 
-    keys = get_list_of_summary_objects(test_bucket_name, 'nested/', anon=True)
+    futures = read_bytes(test_bucket_name+'/test/accounts.*', lazy=False)
+    assert len(futures) >= len(files)
+    results = yield e._gather(futures)
+    assert set(results).issuperset(set(files.values()))
```
mrocklin: Why is this true?  Isn't it possible for the blocks of bytes to be cut differently than the files on S3?

distributed/tests/test_s3.py
```diff
@@ -27,59 +27,112 @@
...
     yield e._start()
-
-    futures = read_bytes(test_bucket_name, prefix='test/', anon=True,
+    bs = 15
+    vals = read_bytes(test_bucket_name+'/test/account*', blocksize=bs)
+    assert len(vals) == sum([(len(v) // bs + 1) for v in files.values()])
+    futures = e.compute(vals)
+    results = yield e._gather(futures)
+    assert sum(len(r) for r in results) == sum(len(v) for v in
+               files.values())
```
mrocklin: Is it reasonable to test the stronger case that `b''.join(results) == b''.join(files.values())`?

distributed/tests/test_s3.py
```diff
@@ -27,59 +27,112 @@
...
+    import io
+    e = Executor((s.ip, s.port), start=False)
+    yield e._start()
+    bs = 15
+    futures = read_bytes(test_bucket_name+'/test/accounts*',
+                         lazy=False, blocksize=bs, delimiter=b'\n')
+    results = yield e._gather(futures)
+    res = [r for r in results if r]
+    data = [io.BytesIO(o).readlines() for o in files.values()]
+    assert set(res) == set(data[0] + data[1])
```
mrocklin: Why is this just the first and second element rather than all of them?  Why the use of `set`?

mrocklin: As above it might be nice to verify that the bytes are equivalent, that all result blocks end in the proper delimiter (if indeed, we want to end with the delimiter as you suggested earlier).

We might also want to try iterating over a few values of `bs`

distributed/tests/test_s3.py
```diff
@@ -27,59 +27,112 @@
...
+
+    # delimiter not at the end
+    d = b'}'
```
mrocklin: Ah, good thinking

distributed/utils.py
```diff
@@ -301,3 +301,115 @@ def iterator_to_queue(seq, maxsize=0):
...
+            i = full.index(delimiter)
+            file.seek(file.tell() - (len(full) - i) + len(delimiter))
+            return
+        except ValueError:
+            pass
+        last = full[-len(delimiter):]
+
+
+def read_block(f, offset, length, delimiter=None):
+    """ Read a block of bytes from a file
```
mrocklin: Should we bring over the tests for these functions as well?

distributed/httpfile.py
```diff
@@ -0,0 +1,95 @@
+# -*- coding: utf-8 -*-
+import requests
```
mrocklin: Would you mind making a new PR with just the `s3.read_block` work so that it's easier to merge quickly?

distributed/utils.py
```diff
@@ -301,3 +301,115 @@ def iterator_to_queue(seq, maxsize=0):
...
+
+    Parameters
+    ----------
+    fn: string
+        Path to filename on HDFS
```
mrocklin: This docstring is out of date (probably my fault)

distributed/httpfile.py
mrocklin: Or alternatively rewrite history on this branch to exclude these commits for now.

distributed/s3.py
```diff
@@ -27,147 +33,741 @@
...
+        delimiter boundaries that follow the locations ``offset`` and ``offset
+        + length``.  If ``offset`` is zero then we start at zero.  The
+        bytestring returned will not include the surrounding delimiter strings.
+
+        If offset+length is beyond the eof, reads to eof.
+
+        Parameters
+        ----------
+        fn: string
+            Path to filename on HDFS
```
mrocklin: Typo with hdfs here

distributed/s3.py
```diff
@@ -27,147 +33,741 @@
...
+        Examples
+        --------
+        >>> s3.read_block('data/file.csv', 0, 13)  # doctest: +SKIP
+        b'Alice, 100\\nBo'
+        >>> s3.read_block('data/file.csv', 0, 13, delimiter=b'\\n')  # doctest: +SKIP
+        b'Alice, 100\\nBob, 200'
+
+        See Also
+        --------
+        hdfs3.utils.read_block
```
mrocklin: same

distributed/s3.py
```diff
@@ -27,147 +33,741 @@
...
+        return out
+
+    def write(self, data):
+        if self.mode != 'wb':
+            raise ValueError('File not in write mode')
+        partno = len(self.part_info['Parts']) + 1
+        part = self.s3.s3.upload_part(Bucket=self.bucket, Key=self.key,
+                                      PartNumber=partno,
+                                      UploadId=self.mpu['UploadId'], Body=data)
+        self.part_info['Parts'].append({'PartNumber': partno, 'ETag': part['ETag']})
```
mrocklin: We'll have to either test or remove this extra functionality before merging.  What would be best?

distributed/tests/test_s3.py
martindurant: The default blocksize is very large compared to the file, so you always get all the data, one future per file. There is a separate test for blocksize smaller than the files.

distributed/tests/test_s3.py
martindurant: There are only two entries, so this is all of them, but iterating was awkward. Set accounts for not being able to rely on the order of the files.
Will try with different values of bs.

mrocklin: Does `read_bytes` guarantee lexicographic ordering of files on S3?  This would be good to test.  I recommend the following:

    assert b''.join(results) == b''.join(files[k] for k in sorted(files))

mrocklin:
Closing in favor of #162  and #159

PR #149, opened 2016-02-21
Change protocol from cloudpickle to msgpack

mrocklin:
I've been changing distributed from cloudpickle to msgpack for message serialization (though keeping cloudpickle for function and data serialization for tasks).  This has been for a few reasons

1.  Cloudpickle *might* be a bottleneck (profiling is hard at this stage)
2.  Cloudpickle introduces some security issues in the scheduler
3.  Msgpack allows us, in the future, to use the scheduler for other languages, like R

Unfortunately msgpack understands less of Python than cloudpickle does, so this change was expensive to implement and messes things up a bit.  For example, msgpack coreces strings to bytestrings, so everywhere in the code that said this `d['status']` now says  this `d[b'status']` and other bit like that.

The transformation is mostly done.  It does not seem to speed up benchmarks considerably, so our bet that this was a bottleneck didn't pay off.  It does make us seem more serious and opens up other language client/workers, which is good.  It will slow down development for the near future though as we pick up lingering bugs from the change and teach ourselves to only use constructs that can flow through msgpack well.

My plan is to see how much I can clean this up for the next half day and then make a decision.  I would welcome thoughts from people that have experience with virtues of various message protocols.  Should we be doing something else?

PR #158, opened 2016-02-28
Add run function

### LINKED ISSUE #156: Direct function evaluation
For diagnostic purposes you sometimes want to run a function outside of the task scheduling framework and get an immediate result.  We can use the scheduler's broadcast functionality to manage this.

mrocklin:
This lets users run arbitrary functions on all workers immediately.
This does not interact with the task scheduling system.

Fixes #156 

cc @quasiben @ogrisel

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -142,9 +142,9 @@ class Future(WrappedKey):
         """
         return sync(self.executor.loop, self._exception)
 
-    def cancel(self, block=False):
+    def cancel(self):
         """ Returns True if the future has been cancelled """
-        return self.executor.cancel([self], block=False)
+        return self.executor.cancel([self])
 
     def cancelled(self):
         """ Returns True if the future has been cancelled """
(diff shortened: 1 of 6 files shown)


ogrisel:
Maybe it would be worth adding a `workers=None` kwarg to run the function only a on subset of the workers, along with some documentation.

Would it make it possible / useful to return futures (or maybe a single future instance) instead of imposing a blocking sync barrier on the caller? A use case would be: make sure that some library is installed or some work folder exists on all the worker and then scheduler some work with submit without blocking.

distributed/executor.py
```diff
@@ -834,6 +834,27 @@ def cancel(self, futures, block=False):
         return sync(self.loop, self._cancel, futures, block=False)
```
ogrisel: This should probably be `block=block`.

mrocklin: Ah, thanks.  I've removed this functionality.  Doesn't seem necessary.

mrocklin:
> Would it make it possible / useful to return futures (or maybe a single future instance) instead of imposing a blocking sync barrier on the caller? A use case would be: make sure that some library is installed or some work folder exists on all the worker and then scheduler some work with submit without blocking.

I can imagine cases where it would be useful.  However, for now I would prefer to keep this operation separate from the task scheduling system.  It operates differently.  The task scheduling system provides various guarantees like rerunning the function if a worker goes down, cancelling jobs, etc..  This doesn't do any of that, it's just an imperative function call, probably for use with side effects.

We could consider adding a broadcast keyword in the submit call in the future, that's a larger change though that would require us to add more state to the scheduler.  I hesitate to do this before necessary.

ogrisel:
Alright for non-blocking call. Indeed this should not go through the usual scheduling logic because it's primarily meant for operations with side-effects.

What about the first part of my remark:

> Maybe it would be worth adding a workers=None kwarg to run the function only a on subset of the workers, along with some documentation.

mrocklin:
Both are good points.  I've expanded the docstring for `run` and placed it into the API docs.  

I'd like to wait on #149 to get in before implementing the `workers=None` fix.

mrocklin:
I've added the `workers=` keyword argument to `Executor.run`

PR #159, opened 2016-03-02
S3 backend

martindurant:
With bytes, text and avro support (not csv, no test files).
Read-only file system for now.


distributed/s3.py
```diff
@@ -2,16 +2,22 @@
...
+import sys
 
 import boto3
-from botocore.handlers import disable_signing
+from botocore.exceptions import ClientError
 from tornado import gen
 
 from dask.imperative import Value
+from dask.base import tokenize
+from distributed.utils import read_block, seek_delimiter
```
mrocklin: Use relative imports within source files.  See the `.utils` import you've added a few lines below.

distributed/s3.py
```diff
@@ -26,148 +32,635 @@
 
 get_s3_lock = threading.Lock()
 
+def split_path(path):
+    path = path.lstrip('s3://')
```
mrocklin: This isn't quite what you want:

```python
In [1]: 's3://sally'.lstrip('s3://')
Out[1]: 'ally'
```

I suggest using `startswith` and slicing.

mrocklin: It'd also be great to see a doctest in this function showing what it does.

distributed/s3.py
```diff
@@ -26,148 +32,635 @@
...
+        raise NotImplementedError('S3 implemented as read-only')
+
+    def rm(self, path, recursive=True):
+        raise NotImplementedError('S3 implemented as read-only')
+
+    def exists(self, path):
+        return bool(self.ls(path))
+
+    def copy(self, path1, path2):
+        raise NotImplementedError('S3 implemented as read-only')
```
mrocklin: I would prefer to remove these from the interface for now.  Otherwise we're just setting users up for disappointment.

distributed/s3.py
```diff
@@ -26,148 +32,635 @@
...
+    logger.debug("Read %d blocks of binary bytes from %s", len(names), fn)
     if lazy:
-        values = [Value(name, [{name: (read_content_from_keys, bucket_name,
-                                       key, anon)}])
-                  for name, key in zip(names, keys)]
+        executor._send_to_scheduler({'op': 'update-graph',
+                                     'tasks': {},
+                                     'dependencies': set(),
+                                     'keys': [],
+                                     'client': executor.id})
```
mrocklin: This send_to_scheduler call is not necessary.  It was used in hdfs to communicate restrictions, which don't exist here.

distributed/s3.py
```diff
@@ -26,148 +32,635 @@
...
+                s3pars=s3pars, delimiter=delimiter)
 
 
-def read_text(bucket_name, prefix='', path_delimiter='', encoding='utf-8',
-        errors='strict', lineterminator='\n', executor=None, anon=None,
-        collection=True, lazy=True, compression=None):
-    """
-    Read lines of text from S3
+def read_block_from_s3(filename, offset, length, s3pars={}, delimiter=None):
+    s3 = S3FileSystem(**s3pars)
```
mrocklin: Just to verify, this is fast to do after the first time, yes?

distributed/s3.py
```diff
@@ -26,148 +32,635 @@
...
+
+
+def avro_to_df(b, av):
+    """Parse avro binary data with header av into a pandas dataframe"""
+    import pandas as pd
+    return pd.DataFrame(data=avro_body(b, av))
+
+
+@gen.coroutine
+def _read_avro(path, executor=None, fs=None, lazy=True, **kwargs):
```
mrocklin: Thoughts on removing all of this avro and csv stuff for now?

distributed/s3.py
martindurant: The S3FileSystem does cache connections and directory entries; but actually boto3 establishes new HTTP connections for everything it does.

distributed/tests/test_s3.py
```diff
@@ -27,60 +27,154 @@
...
+    assert fn not in s3.ls('distributed-test/')
+    assert fn in s3.ls('distributed-test/nested/')
+    assert fn in s3.ls('distributed-test/nested')
+    assert fn in s3.ls('s3://distributed-test/nested/')
+    assert fn not in s3.glob('distributed-test/')
+    assert fn not in s3.glob('distributed-test/*')
+    assert fn in s3.glob('distributed-test/nested')
+    assert fn in s3.glob('distributed-test/nested/*')
+    assert fn in s3.glob('distributed-test/nested/file*')
+    assert fn in s3.glob('distributed-test/*/*')
```
mrocklin: These are great.  Thanks!

I recommend splitting some of these off into separate tests.  One for glob, one for ls, one for cat/head/tail, etc..  This helps us to quickly isolate what breaks when we make changes.

distributed/s3.py
martindurant: CSV I removed due to no tests (but appears to work)
Seems a pity, it works surprisingly easy. Could be kept elsewhere...

distributed/tests/test_s3.py
```diff
@@ -27,60 +27,154 @@
...
+    fn = 'test/accounts.1.json'
+    data = files[fn]
+    with s3.open('/'.join([test_bucket_name, fn])) as f:
+        seek_delimiter(f, b'}', 0)
+        assert f.tell() == 0
+        f.seek(1)
+        seek_delimiter(f, b'}', 5)
+        assert f.tell() == data.index(b'}') + 1
+        seek_delimiter(f, b'\n', 5)
+        assert f.tell() == data.index(b'\n') + 1
```
mrocklin: Can we verify also that it will seek from current location.  So for example if we seek to `data.index(b'\n') + 1` and then call `seek_delimiter` again do we end up at the new location?

distributed/tests/test_s3.py
```diff
@@ -27,60 +27,154 @@
...
+    yield e._start()
+    for bs in [5, 15, 45, 1500]:
+        futures = read_bytes(test_bucket_name+'/test/accounts*',
+                             lazy=False, blocksize=bs, delimiter=b'\n')
+        results = yield e._gather(futures)
+        res = [r for r in results if r]
+        assert all(r.endswith(b'\n') for r in res)
+        ourlines = b''.join(res).split(b'\n')
+        testlines = b"".join(files.values()).split(b'\n')
+        assert set(ourlines) == set(testlines)
```
mrocklin: I think if you use `b''.join(files[k] for k in sorted(files))` you'll be able to test order here as well

distributed/tests/test_s3.py
```diff
@@ -27,60 +27,154 @@
...
+        ourlines = b"".join(results).split(b'\n')
+        testlines = b"".join(files.values()).split(b'\n')
+        assert set(ourlines) == set(testlines)
+    yield e._shutdown()
+
+
+@gen_cluster(timeout=60)
+def test_read_bytes_delimited(s, a, b):
+    e = Executor((s.ip, s.port), start=False)
+    yield e._start()
```
mrocklin: If you're interested, I've added an `executor=True` keyword argument to `gen_cluster` to automate the construction and shutdown of the exectuor:

```python
@gen_cluster(timeout=60, executor=True)
def test_read_bytes_delimited(e, s, a, b):
   ...
```

distributed/utils.py
```diff
@@ -352,3 +352,115 @@ def convert(task):
...
+
+        offset = start
+        length = end - start
+
+        # if length and not eof:
+        #    length -= len(delimiter)
+
+    f.seek(offset)
+    bytes = f.read(length)
+    return bytes
```
mrocklin: This function could use its test companion from the hdfs3 library

mrocklin:
This is looking great to me so far

distributed/s3.py
mrocklin: Yeah, we should keep working on it in another branch and merge it when it's ready.

distributed/utils.py
martindurant: Sorry, what do you mean?

mrocklin: We should copy over this test from hdfs3: https://github.com/dask/hdfs3/blob/master/hdfs3/tests/test_utils.py#L32-L52

If a future developer decides to change this function then something needs to yell at them.

distributed/hdfs.py
```diff
@@ -78,7 +80,7 @@ def read_bytes(fn, executor=None, hdfs=None, lazy=True, delimiter=None,
         offsets = [max([o, 1]) for o in offsets]
     lengths = [d['length'] for d in blocks]
     workers = [[h.decode() for h in d['hosts']] for d in blocks]
-    names = ['read-binary-%s-%d-%d' % (fn, offset, length)
+    names = ['read-binary-hdfs3-%s-%s' % (fn, tokenize(offset, length, delimiter, not_zero))
```
mrocklin: Can we add a regression test for this?  Perhaps something like 

```python
A = read_bytes(..., delimiter=b'\n')
B = read_bytes(..., delimiter=b'foo')
assert [a.key for a in A] != [b.key for b in B]
```

distributed/s3.py
```diff
@@ -26,148 +32,526 @@
...
+    >>> split_path("s3://mybucket/path/to/file")
+    ("mybucket", "path/to/file")
+    """
+    if path.startswith('s3://'):
+        path = path[5:]
+    if '/' not in path:
+        return path, ""
+    else:
+        bits = path.split('/')
+        return bits[0], '/'.join(bits[1:])
```
mrocklin: I think this can be `path.split('/', 1)`

```python
In [1]: '1,2,3,4,5'.split(',')
Out[1]: ['1', '2', '3', '4', '5']

In [2]: '1,2,3,4,5'.split(',', 1)
Out[2]: ['1', '2,3,4,5']
```

distributed/s3.py
```diff
@@ -26,148 +32,526 @@
...
-        Name of S3 bucket like ``'my-bucket'``
-    prefix: string
-        Prefix of key name to match like ``'/data/2016/``
-    path_delimiter: string (optional)
-        Delimiter like ``'/'`` to define implicit S3 directory structure
+    fn: string
+        location in S3
     executor: Executor (optional)
         defaults to most recently created executor
+    s3: HDFileSystem (optional)
```
mrocklin: typo

distributed/s3.py
martindurant: Ah, I undid that because `maxsplit` is not in py2

distributed/utils.py
mrocklin: ping

distributed/s3.py
mrocklin:
```python
In [1]: '1,2,3,4,5'.split(',', 1)
Out[1]: ['1', '2,3,4,5']

In [2]: import sys; sys.version
Out[2]: '2.6.9 |Continuum Analytics, Inc.| (unknown, Aug 21 2014, 18:28:52) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]'
```

distributed/s3.py
```diff
@@ -26,148 +32,526 @@
...
+            files = self._ls(path)
+        except ClientError:
+            files = []
+        if path:
+            pattern = re.compile(path + '/[^/]*.$')
+            files = [f for f in files if pattern.match(f['Key']) is not None]
+            if not files:
+                try:
+                    files = [self.info(path)]
+                except OSError:
```
mrocklin: In Python 2 this raises an IOError

martindurant: Yes, that's a bug, should be IOError

distributed/tests/test_s3.py
```diff
@@ -27,60 +27,171 @@
...
+        seek_delimiter(f, b'}', 5)
+        assert f.tell() == data.index(b'}') + 1
+        seek_delimiter(f, b'\n', 5)
+        assert f.tell() == data.index(b'\n') + 1
+        f.seek(1, 1)
+        ind = data.index(b'\n') + data[data.index(b'\n')+1:].index(b'\n') + 1
+        seek_delimiter(f, b'\n', 5)
+        assert f.tell() == ind + 1
+
+
```
mrocklin: This test is slow.  We should decorate it with the `slow` decorator in `utils_test.py`.

Alternatively, it would be interesting to learn to use `moto`, which would really help with things like this.  If you want to dive into S3 more seriously it might be worth investing some time here.

distributed/s3.py
mrocklin: It's probably both.  This is a common point of difference between Python 2 and 3.

I recommend `except (OSError, IOError):`

distributed/tests/test_s3.py
martindurant: I don't think I am interested in doing more with S3 than we currently have in scope. As you said, though, tests should be much faster within AWS.

mrocklin: Moto would also facilitate things like writing and would make our tests more self-describing, rather than relying on pre-seeded buckets.

I'm just thinking aloud here about moto though.  I think we should get this in and then think about using moto as we expand towards writing and more filesystem-y features.

distributed/utils.py
martindurant: Yes, so we want to keep that final delimiter - the test needs to be modified.

mrocklin: OK, either way it should be explicit.  We need to include that test within this PR.

distributed/hdfs.py
```diff
@@ -21,7 +22,7 @@
 
 
 def walk_glob(hdfs, path):
-    if '*' not in path and hdfs.info(path)['kind'] == 'directory':
+    if '*' not in path and hdfs.info(path).get('kind', '') == 'directory':
```
mrocklin: What brought this up?

mrocklin: To be pedantic here, in an ideal world no source change should really ever happen without a similar change in tests.  Is it ok if I revert this change?  If not why not?

distributed/s3.py
```diff
@@ -26,148 +32,525 @@
...
+            if bucket == '':
+                # list of buckets
+                if self.anon:
+                    # cannot list buckets if not logged in
+                    return []
+                files = self.s3.list_buckets()['Buckets']
+                for f in files:
+                    f['Key'] = f['Name']
+                    f['Size'] = 0
+                    del f['Name']
```
mrocklin: What is the reasoning behind this?

distributed/s3.py
```diff
@@ -26,148 +32,525 @@
...
+                self.cache = self.s3.s3.get_object(Bucket=self.bucket, Key=self.key,
+                                                Range='bytes=%i-%i' % (start, self.end - 1)
+                                                )['Body'].read()
+            if start < self.start:
+                new = self.s3.s3.get_object(Bucket=self.bucket, Key=self.key,
+                                         Range='bytes=%i-%i' % (start, self.start - 1)
+                                         )['Body'].read()
+                self.start = start
+                self.cache = new + self.cache
+            if end > self.end:
```
mrocklin: When does this ever occur?  I can't seem to trigger this code path.

distributed/hdfs.py
martindurant: My apologies, this should be reverted - it is a remnant of trying to get the hdfs read_bytes to work with s3fs directly

mrocklin: It was reverted (or rather I just didn't bring along the change) in #162

distributed/s3.py
martindurant: To make a user's buckets look like top-level directories. That normalises (to my mind) a path like s3://mybucket/path/file ; you can do a ls('/') and see apparent directories.

distributed/s3.py
martindurant: You can request any block, including past the end of the file (like with BytesIO); that should return empty bytestrings and not error.

mrocklin: I wasn't able to trigger this.  What is a code snippet that might trigger this code?

martindurant: Here is the snippet - and it broke! `self.s3.s3.get_object.get(` should b `self.s3.s3.get_object(`

```python
s3 = distributed.s3.S3FileSystem()
f = s3.open('distributed-test/test/accounts.1.json', block_size=20)
while f.read(10):
    pass
f.seek(5000)
f.read(10)
```

Each time you go past the current (very small) cache, you need that code path; also seeking past the end of the file. `read_block` explicitly checks the file size instead.

mrocklin: Great.  Added in https://github.com/mrocklin/distributed/commit/de7550b847db1e7b12f739962baafad5fd153afd

mrocklin:
Stale, closing

PR #160, opened 2016-03-02
Add Executor.{ncores,who_has,has_what}

mrocklin:


minrk:
Great! This will be useful for debugging with IPython. Can workers have names that I could use/set, so that it would be easy to get from the `who_has` result to the IPython engine ID?

mrocklin:
That would be good to add.  Do you have thoughts on how to assign user-defined or auto-generated names?

mrocklin:
Mostly I'm curious about how to assign names in a setting where workers can show up and disappear dynamically, and proper user interfaces to help provide hints as to what they want names to be.

How do you resolve this in `ipyparallel`?

minrk:
I would do it in the Worker constructor, and default to the address if not given:

```python
Worker(name='alice')
```

I'm not sure about passing that around the cluster and resolving names, though. I can also store the addresses in IPython when I make the workers, so that it's easier.

minrk:
> How do you resolve this in ipyparallel?

In IPython, Engine IDs are an integer counter, assigned by the Hub when the engine registers itself. Engines cannot name themselves. There's a routing ID associated with engine IDs used for zmq purposes.

mrocklin:
OK, naming a worker on creation makes sense to me.  Presumably we err if the name is taken?  I can add this.  It will probably be in another PR though.

minrk:
Yup, I'd balk on name collision. Another PR makes sense, I'm just trying to think with my "debugging with IPython" hat on.

mrocklin:
That hat looks good on you :)

PR #164, opened 2016-03-04
S3 csv

mrocklin:
This copies the read_csv functionality over from HDFS to S3.

This is more-or-less a copy-paste-edit job.  It might be worthwhile looking into how to generalize this in the future.  There was a fair amount of code duplication.

### DIFF
--- a/distributed/hdfs.py
+++ b/distributed/hdfs.py
@@ -109,8 +109,9 @@ def bytes_read_csv(b, **kwargs):
 
 
 @gen.coroutine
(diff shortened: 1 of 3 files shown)


mrocklin:
@martindurant when you have time if you could give this a quick review that would be appreciated.

mrocklin:
@martindurant are we good to go with this?

martindurant:
All good, +1

mrocklin:
Thanks for the attentive review.  That was very helpful.

PR #165, opened 2016-03-04
Worker aliases

mrocklin:
This allows the assignment of names to workers either through the programming or command line interfaces:

```python
Worker(ip, port, name='alice')
```
```
$ dworker ip:port --name alice
```

These names can be used directly within computational calls like map or submit

```python
e.map(func, sequence, workers='alice')
```

I haven't yet propagated this throughout all user options, nor have I added sufficient state to make it constant time.

cc @minrk @seibert

### DIFF
--- a/distributed/center.py
+++ b/distributed/center.py
@@ -37,9 +37,8 @@ class Center(Server):
         Set of keys owned by a particular worker
     *   ``ncores:: {worker: int}``
         Number of cores per worker
-    *   ``worker_services:: {worker: {str: port}}``:
-        Ports of other running services on each worker.
-        E.g. ``{('192.168.1.100', 8000): {'http': 9001, 'nanny': 9002}}``
+    *   ``worker_info``:: {worker: {str: data}}``:
+        Information on each worker
 
     Workers and clients check in with the Center to discover available resources
 
@@ -74,10 +73,10 @@ class Center(Server):
      ('bob', 8788): 4,
      ('charlie', 8788): 4}
 
-    >>> center.worker_services # doctest: +SKIP
-    {('alice', 8788): {'nanny', 8789},
-     ('bob', 8788): {'nanny', 8789},
-     ('charlie', 8788): {'nanny', 8789}}
+    >>> center.worker_info  # doctest: +SKIP
+    {('alice', 8788): {'services': {'nanny', 8789}},
+     ('bob', 8788): {'services': {'nanny', 8789}},
+     ('charlie', 8788): {'servies': {'nanny', 8789}}}
 
     See Also
     --------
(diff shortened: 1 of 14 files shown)


seibert:
Very nice!  Is it assumed that worker names are unique?  If I gave all my workers with GPUs the name "gpu", could I use this feature to direct tasks to that set of workers?

mrocklin:
At the moment if you did that then bad things would happen (I've added a WIP prefix to this PR).  I plan to enforce alias uniqueness, but perhaps we want to have something else, like group membership.  

If I added groups would you actively use them?

mrocklin:
Of course, you can always specify a set of workers in any map/submit call.  So it's still possible to achieve the same result you want, even with old code, by arranging the groups on the client side.

seibert:
OK, I assumed these were meant to be unique aliases, but wanted to check.  I don't have an immediate need for groups, but could imagine them as being a nice way to manage sets of workers with special resources.  I'd throw it on the wishlist, but not devote any time to it now.

mrocklin:
OK.  Honestly it's not so much about devoting time to it; it's easy to do.  It's more about increasing inertia.  I've noticed that the cost to make a change to the scheduler is increasing slightly over time.

Specifying task location is a cool and novel feature though.  I'm quite happy to facilitate it if useful.  I just want to avoid making features that no one actively uses.

seibert:
My logic with groups is that as soon as you have one special node, it is easy to imagine having more than one special node.  However, these sorts of groups could easily be managed on the client (just put all the GPU nodes into a list and pass that around), so I'm not sure if making the scheduler aware the group has a name is valuable.

I think the biggest benefit to setting task locations will come with computation graphs that are most efficient with a heterogenous cluster.  One set of tasks in the graph get 5x from a GPU, but the rest can be run on much cheaper CPU-only systems.

seibert:
And this isn't even exclusive to GPUs.  There might be tasks that just work better with a lot of RAM, so pinning them to a "bigmem" group is beneficial.

mrocklin:
To be clear, we already support pinning tasks to sets of workers (see [docs](http://distributed.readthedocs.org/en/latest/locality.html#user-control)).  What's changing here is the ability to refer to these workers by user created names, rather than by IP addresses or hostnames.

minrk:
If you call it a name/alias, enforcing uniqueness makes sense. For a use case like `gpu`, a different but related concept of groups would make sense. I could imagine both being useful, though you can always implement the groups concept client-side by manually creating collections of names.

seibert:
Yeah, I'm just thinking out loud how I might end up with a group of servers I might want to give a name to (rather than just an individual).   The one benefit for having the group definition server-side would be so that a system administrator could add nodes to special groups that are documented for users of the cluster.  Then the user only needs to know that the admin created a "gpu" group, but not what is in it.

seibert:
But, as I said before, I don't have a burning need for groups right at this moment, so I don't want to add to its priority level.  :)

PR #166, opened 2016-03-04
Avoid error in map thread

mrocklin:
Previously we exposed a bare StopIteration error in
Exectuor._threaded_map.  Now we catch and break appropriately.

```python
In [1]: from distributed import Executor

In [2]: e = Executor('127.0.0.1:8786')

In [3]: items = enumerate(range(10))

In [4]: futures = e.map(lambda x: x, items)

In [5]: next(futures).result()
Out[5]: (0, 0)
```

@sklam I suspect that your problem was actually with `futures[0]`.  Because you give map an iterator, its giving you back an iterator.  The error you saw, I think, was benign but has now been cleaned up regardless.

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -526,9 +526,13 @@ class Executor(object):
             raise NotImplementedError()
 
         while True:
-            args = [get(q) for q in qs_in]
+            try:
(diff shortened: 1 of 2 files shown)


quasiben:
+1

PR #167, opened 2016-03-04
Add back S3 avro functions, with tests

martindurant:
Tests are largely modelled on read_csv, but with some additions for
the avro helper functions.

### DIFF
--- a/distributed/tests/test_s3.py
+++ b/distributed/tests/test_s3.py
@@ -371,6 +371,12 @@ def test_read_csv(e, s, a, b):
     result = yield e.compute(df2.id.sum())._result()
     assert result == 1 + 2 + 3 + 4 + 5 + 6
 
+    df2 = yield _read_csv('distributed-test/csv/2015/',
+                          collection=True, lazy=False, blocksize=20)
+    f = e.compute(df2.amount.sum())
+    result = yield f._result()
+    assert result == (100 + 200 + 300 + 400 + 500 + 600)
+
 
 @gen_cluster(timeout=60, executor=True)
 def test_read_csv_gzip(e, s, a, b):
@@ -389,7 +395,7 @@ def test_read_csv_sync(loop):
     dd = pytest.importorskip('dask.dataframe')
     with cluster() as (s, [a, b]):
         with Executor(('127.0.0.1', s['port']), loop=loop) as e:
-            df = read_csv('distributed-test/csv/2015/')
+            df = read_csv('distributed-test/csv/2015/', lazy=True)
             assert isinstance(df, dd.DataFrame)
             assert list(df.columns) == ['name', 'amount', 'id']
             f = e.compute(df.amount.sum())
@@ -482,3 +488,5 @@ def test_read_avro_sync(loop):
             assert list(df.columns) == ['amount', 'name']
             f = e.compute(df.amount.sum())
             assert f.result() == (100 + 200 + 300 + 400) * 2
+            df = read_csv('distributed-test/csv/2015/', lazy=False)
+            assert df.amount.sum().compute() == (100 + 200 + 300 + 400 + 500 + 600)
--- a/distributed/tests/test_s3.py
+++ b/distributed/tests/test_s3.py
@@ -1,6 +1,9 @@
-import pytest
-import json
+from __future__ import print_function, division, absolute_import
+
 import io
+import json
+from math import ceil
+import pytest
 
 import boto3
 from tornado import gen
@@ -45,6 +48,16 @@ def make_avro():
     writer(b, schema, records)
     b.seek(0)
     return b.read()
+
+csv_files = {'2014-01-01.csv': (b'name,amount,id\n'
+                                b'Alice,100,1\n'
+                                b'Bob,200,2\n'
+                                b'Charlie,300,3\n'),
+             '2014-01-02.csv': (b'name,amount,id\n'),
+             '2014-01-03.csv': (b'name,amount,id\n'
+                                b'Dennis,400,4\n'
+                                b'Edith,500,5\n'
+                                b'Frank,600,6\n')}
 
 
 @pytest.yield_fixture
@@ -244,7 +257,6 @@ def test_read_bytes_lazy(e, s, a, b):
 
 @gen_cluster(timeout=60, executor=True)
 def test_read_text(e, s, a, b):
-    pytest.importorskip('dask.bag')
     import dask.bag as db
     from dask.imperative import Value
 
@@ -268,8 +280,23 @@ def test_read_text(e, s, a, b):
     assert all(isinstance(v, Future) for v in text)
 
 
+@gen_cluster(timeout=60, executor=True)
+def test_read_text_blocksize(e, s, a, b):
+    for bs in [20, 27, 12]:
+        b = yield _read_text(test_bucket_name+'/test/accounts*', lazy=True,
+                             blocksize=bs, collection=True)
+        assert b.npartitions == sum(ceil(len(b) / bs) for b in files.values())
+
+
+@gen_cluster(timeout=60, executor=True)
+def test_read_text_compression(e, s, a, b):
+    b = yield _read_text('distributed-test/csv/gzip/', compression='gzip')
+    result = yield e.compute(b)._result()
+    assert result == [line for k in sorted(csv_files)
+                           for line in csv_files[k].decode().split('\n')]
+
+
 def test_read_text_sync(loop):
-    pytest.importorskip('dask.bag')
     import dask.bag as db
     with cluster() as (s, [a, b]):
         with Executor(('127.0.0.1', s['port']), loop=loop) as e:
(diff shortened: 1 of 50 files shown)


distributed/tests/test_s3.py
```diff
@@ -393,3 +394,91 @@ def test_read_csv_sync(loop):
...
+    schema = header['meta']['avro.schema'].decode()
+    schema = json.loads(schema)
+    assert ['amount', 'name'] == sorted(f['name'] for f in schema['fields'])
+    sync = header['sync']
+    assert len(sync) == 16
+    bit = byte[byte.index(sync)+len(sync):]
+    assert avro_body(bit, header) == records
+    bit = byte[byte.index(sync)+len(sync):-len(sync)]
+    assert avro_body(bit, header) == records
+    assert avro_body(b"", header) == []
```
mrocklin: Ideally we would be able to test this function without relying on S3.  Perhaps now would be a good time to think about refactoring out formats?  We're not particularly time constrained on the avro work (no one is actively asking for it) so the choice between copy-paste and refactor might change here (we may want to refactor.)

mrocklin:
Looks like you need to add a `pip install fastavro` line to the travis.ci script

distributed/tests/test_s3.py
```diff
@@ -393,3 +394,91 @@ def test_read_csv_sync(loop):
...
+    bit = byte[byte.index(sync)+len(sync):]
+
+    df = avro_to_df(bit, header)
+    assert ['amount', 'name'] == sorted(df.columns)
+    assert len(df) == 4
+    assert 100 + 200 + 300 + 400 == df.amount.sum()
+    df = avro_to_df(b"", header)
+    assert len(df) == 0
+    df = avro_to_df(b"", header, cols=['amount', 'name'])
+    assert ['amount', 'name'] == list(df.columns)
```
mrocklin: Same here, this function takes bytes.  Ideally a unittest relies only on the functionality we intend to test.  At the moment this test, which I think intends to test `avro_to_df` will break if any of the S3 code breaks as well.

mrocklin:
In general this looks fine to me.  I suggest that we push to get things to pass, merge, and then think about refactoring out formats.

martindurant:
Sorry, where do I put that `pip`? I agree that we should push this as 'avro-in-s3' and then, if we feel it's useful, plan that refactor. Otherwise I would probably do the S3-write piece next with moto (unless you have a good suggestion for a blog post).

mrocklin:
There is a `.travis.yml` file in the repository.  There are various other `pip install` lines, you would just copy one of those.

martindurant:
`distributed/tests/test_s3.py::test_read_bytes_delimited` passes on my py2

mrocklin:
You could try one of the following to try to diagnose the issue further:

1.  Try to create an environment very much like the travis.ci environment.  Create clean conda environments with the same conda/pip install commands, watch out for local config files, etc..
2.  Add a `-s` flag to the `py.test` call in the `.travis.yml` file to get debug messages in the tests in travis.ci and then rebase it out later.

I generally find the first option to provide a faster iteration time.

martindurant:
With local new travis-like env and py27, get one failed test (`distributed/diagnostics/tests/test_widgets.py::test_fast`), and before getting to test_s3, get internalerrors which appear caused by
`INTERNALERROR> EMFILE: [Too many open files]: getcwd()`
(note that I am on mac, not ubuntu like travis).

Run alone, no failures in `distributed/tests/test_s3.py`.

This will take a little digging...

mrocklin:
OK, that's probably an internal issue.  We must be doing something wrong with how we close out sockets.  I can look into this.

mrocklin:
See #169

mrocklin:
It would help me to be able to diagnose the issue if you can provide the errors you're getting.

martindurant:
PY3 failures

test_defaults 
UnboundLocalError: local variable 'e' referenced before assignment
(in finally block of test code - does not have gen_cluster decorator?)

test_sync_compute 
OSError: [Errno 24] Too many open files
at distributed/utils_test.py:149: in cluster
q = Queue()

test_executor_with_scheduler
tornado.iostream.StreamClosedError: Stream is closed
logged errors:
File "/Users/mdurant/Downloads/distributed/distributed/utils.py", line 229, in log_errors
    yield
  File "/Users/mdurant/Downloads/distributed/distributed/client.py", line 113, in gather_from_workers
    raise KeyError(*bad_keys)
;
distributed.core - ERROR - ('Key exists in wrong number of places', 'inc-0f41327e4a8b32b59c636e4724d4cc6a', 2)
;
and some tornado raise errors on result

test_map_iterator_sync
OSError: [Errno 24] Too many open files
at with Executor(('127.0.0.1', s['port']), loop=loop) as e
from socket __init__
[could it be that earlier errors leave lockfiles/sockets around?]

test_Future_exception_sync
OSError: [Errno 24] Too many open files
at with Executor(('127.0.0.1', s['port']), loop=loop) as e:

test_coerce_address
tornado.ioloop.TimeoutError: Operation timed out after 10 seconds
at loop.run_sync(cor, timeout=timeout)
See logged message: distributed.scheduler - INFO - Worker failed from closed stream: 127.0.0.2:57034

mrocklin:
Is this all the relevant information from the printout or are you redacting some things?

mrocklin:
For example I would love to know which `test_defaults` that error refers to and a line number if it was printed.  Generally an unadulterated copy-paste is appreciated.

martindurant:
This is the abridged version. I could email you the whole thing. I suspect the default file limits on mac may bear some responsibility - just set some parameters and running again. Still, can't see how we have more than a few dozen locks and sockets open.

mrocklin:
I recommend copy-pasting the entire error section here or, if it's too large for github to accept, put it into a gist and paste the link to the gist here.  It's nice to keep things out in the open and in a permanent record.

mrocklin:
If you're interested I'm debugging this by printing out the number of file descriptors before and after all of the `gen_cluster` tests.  Here is a diff


```python
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -359,7 +359,7 @@ def str_graph(dsk):
 
 import logging
 logging.basicConfig(format='%(name)s - %(levelname)s - %(message)s',
+                    level=logging.CRITICAL)

         def test_func():
+            import psutil
+            proc = psutil.Process()
+            start = proc.num_fds()
             IOLoop.clear_instance()
             loop = IOLoop()
             loop.make_current()
@@ -415,6 +422,8 @@ def gen_cluster(ncores=[('127.0.0.1', 1), ('127.0.0.1', 2)], timeout=10,
                 loop.run_sync(lambda: end_cluster(s, workers))
                 loop.stop()
                 loop.close()
+                end = proc.num_fds()
+                print("Start: %d, End: %d" % (start, end))
 
         return test_func
     return _
```

Then I run with `-s` and see things like the following:

```
PASSED
distributed/tests/test_scheduler.py::test_restart <- distributed/utils_test.py distributed.scheduler - CRITICAL - Lost all workers
distributed.nanny - CRITICAL - Unable to unregister with center None. Nanny: ('127.0.0.1', 48522), Worker: 127.0.0.1:41628
distributed.nanny - CRITICAL - Unable to unregister with center None. Nanny: ('127.0.0.1', 56642), Worker: 127.0.0.1:38804
distributed.scheduler - CRITICAL - Lost all workers
Start: 683, End: 708
PASSED
distributed/tests/test_scheduler.py::test_scheduler <- distributed/utils_test.py Start: 708, End: 713
PASSED
distributed/tests/test_scheduler.py::test_multi_queues <- distributed/utils_test.py Start: 713, End: 695
PASSED
distributed/tests/test_scheduler.py::test_delete_data <- distributed/utils_test.py Start: 695, End: 697
PASSED
distributed/tests/test_scheduler.py::test_rpc <- distributed/utils_test.py Start: 697, End: 697
PASSED
distributed/tests/test_scheduler.py::test_filtered_communication <- distributed/utils_test.py Start: 697, End: 705
PASSED
distributed/tests/test_scheduler.py::test_remove_worker_from_scheduler <- distributed/utils_test.py Start: 705, End: 705
```

martindurant:
https://gist.github.com/martindurant/6c32deabc37edd75e18d

mrocklin:
Great, that printout greatly aids identifying the two issues present there, which are actually unrelated to the open files issue.  The libhdfs3 failure is because we should be guarding the hdfs3 import better in tests.

The test_defaults error is because the CLI tests require that the executables like `dscheduler` be installed in the system.  We should either change these tests to point to the local executables within the `cli/` directory or you should install the library before testing; perhaps with `py.test setup.py develop`.

martindurant:
With HDFS guard and installed CLI, get these remaining issues:
https://gist.github.com/martindurant/2656908684d3965b80c2

Note that checks above claim to pass (but I didn't push anything new).

mrocklin:
Linux allows more file descriptors per process than OSX by default.  Defaults are something like Linux 1024, OSX: 128.  It's not surprising that travis.ci is passing even while the problem persists.  This is a genuine problem though.  I'm going through and working to resolve it now.

This error: `distributed.utils - ERROR - fd 49 already registered` is surprising and novel.

mrocklin:
@martindurant can you pull from my cleaup-streams branch and try again?

Or alternatively

```
git more update mrocklin
git cherry-pick 4ecf69a
```

https://github.com/mrocklin/distributed/commit/4ecf69a43ccd572103301ed2d5e87c5dd17906b1

distributed/tests/test_s3.py
mrocklin: Ping

distributed/tests/test_s3.py
mrocklin: Ping

mrocklin:
This appears to be stale.  Closing.

PR #169, opened 2016-03-06
Cleanup streams on garbage collecting rpc

mrocklin:
We were running into too many open file descriptors.  This *may* resolve the problem.

@martindurant you might want to merge this branch into your branch and try again.  You probably know git well enough to do this, but just in case here are some useful commands.

```
git remote add mrocklin git@github.com:mrocklin/distributed.git
git remote update mrocklin
git merge mrocklin/cleanup-streams # merge this branch into your current active branch
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ install:
   - source activate test-environment
   - conda install pytest coverage tornado toolz dill futures dask ipywidgets psutil bokeh requests
   - pip install git+https://github.com/dask/dask.git --upgrade
+  - conda install -c pandas pandas=v0.18.0rc1
 
   # Install distributed
   - python setup.py install
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -234,8 +234,8 @@ def connect(ip, port, timeout=3):
     client = TCPClient()
     start = time()
     while True:
+        future = client.connect(ip, port, max_buffer_size=MAX_BUFFER_SIZE)
         try:
-            future = client.connect(ip, port, max_buffer_size=MAX_BUFFER_SIZE)
             stream = yield gen.with_timeout(timedelta(seconds=timeout), future)
             raise Return(stream)
         except StreamClosedError:
@@ -385,6 +385,9 @@ class rpc(object):
             self.streams[stream] = True  # mark as open
             raise Return(result)
         return send_recv_from_rpc
+
+    def __del__(self):
+        self.close_streams()
 
 
 def coerce_to_address(o, out=str):
--- a/distributed/diagnostics/progressbar.py
+++ b/distributed/diagnostics/progressbar.py
@@ -82,6 +82,10 @@ class ProgressBar(object):
     def _draw_stop(self, **kwargs):
         pass
 
+    def __del__(self):
+        with ignoring(AttributeError):
+            self.stream.close()
+
 
 class TextProgressBar(ProgressBar):
     def __init__(self, keys, scheduler=None, interval=0.1, width=40,
@@ -145,7 +149,6 @@ class ProgressWidget(ProgressBar):
         self.bar_text.value = '<div style="padding: 0px 10px 0px 10px; text-align:right;">%d / %d</div>' % (ndone, all)
 
 
-
 class MultiProgressBar(object):
     def __init__(self, keys, scheduler=None, func=key_split, interval=0.1, complete=False):
         self.scheduler = get_scheduler(scheduler)
@@ -198,6 +201,10 @@ class MultiProgressBar(object):
 
     def _draw_stop(self, **kwargs):
         pass
+
+    def __del__(self):
+        with ignoring(AttributeError):
+            self.stream.close()
 
 
 class MultiProgressWidget(MultiProgressBar):
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -252,7 +252,7 @@ class Executor(object):
     def __init__(self, address, start=True, loop=None, timeout=3):
         self.futures = dict()
         self.refcount = defaultdict(lambda: 0)
-        self._should_close_loop = start and not loop
+        self._should_close_loop = loop is None and start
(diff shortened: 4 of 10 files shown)


martindurant:
> On Mar 6, 2016, at 18:23, Matthew Rocklin <notifications@github.com> wrote:
> 
> git remote update mrocklin

I don’t seem to have access to this?

mrocklin:
Odd, you should be able to use the HTTP version then

    git remote add mrocklin https://github.com/mrocklin/distributed.git

mrocklin:
OK, we're cleaning up file iostreams much more aggressively throughout tests.  The dominant source of lingering file descriptors (as measured by `psutil.Process.num_fds`) seems to be from creating Processes which are then terminated later on.  I'm not sure if this counts towards our limit or not or if there is a way around it.

mrocklin:
This seems to stay under 80 file descriptors total with 20 to 40 being typical throughout tests.

@martindurant can I ask you to give this a whirl on your OSX machine?

mrocklin:
Testing with 

```
+++ b/distributed/utils.py
@@ -359,7 +359,7 @@ def str_graph(dsk):
 
 import logging
 logging.basicConfig(format='%(name)s - %(levelname)s - %(message)s',
+                    level=logging.CRITICAL)
 
 # http://stackoverflow.com/questions/21234772/python-tornado-disable-logging-to-stderr
 stream = logging.StreamHandler(sys.stderr)
diff --git a/distributed/utils_test.py b/distributed/utils_test.py
index 4840406..c078d7c 100644
@@ -147,6 +147,8 @@ def run_nanny(q, center_port, **kwargs):
 
 @contextmanager
 def cluster(nworkers=2, nanny=False):
+    import psutil
+    startps = psutil.Process().num_fds()
     if nanny:
         _run_worker = run_nanny
     else:
@@ -199,6 +201,8 @@ def cluster(nworkers=2, nanny=False):
         for fn in glob('_test_worker-*'):
             shutil.rmtree(fn)
         loop.close(all_fds=True)
+        endps = psutil.Process().num_fds()
+        print("Start: %d, End: %d" % (startps, endps))
 
 
 @contextmanager
@@ -415,6 +419,8 @@ def gen_cluster(ncores=[('127.0.0.1', 1), ('127.0.0.1', 2)], timeout=10,
         cor = gen.coroutine(func)
 
         def test_func():
+            import psutil
+            startps = psutil.Process().num_fds()
             IOLoop.clear_instance()
             loop = IOLoop()
             loop.make_current()
@@ -435,6 +441,8 @@ def gen_cluster(ncores=[('127.0.0.1', 1), ('127.0.0.1', 2)], timeout=10,
                 loop.run_sync(lambda: end_cluster(s, workers))
                 loop.stop()
                 loop.close(all_fds=True)
+                endps = psutil.Process().num_fds()
+                print("Start: %d, End: %d" % (startps, endps))
```

martindurant:
Would you like me to test that one commit or the current state of the branch?

mrocklin:
I would be interested to see how this branch fares on OSX in isolation from other changes.  I would also be interested in hearing if it crashes when merged into your avro branch.  If the latter occurs, then it would be interesting to implement something like what's in the diff above and see the result of `py.test distributed --verbose -s` so that we can see which tests are leaking file descriptors.

```
git remote update mrocklin
git checkout mrocklin/cleanup-streams
```

mrocklin:
Or more generally, I'm looking to check if the underlying problem is solved.  I'm happy with any approach that determines that.  Note that I didn't touch Queues at all.  Things seems to be fine on my end without it.

martindurant:
(with default number of open files, 256)
https://gist.github.com/martindurant/92f670cb4054905690e5

mrocklin:
Ah, sorry, I should have been more direct.  The output from tests is only really helpful if you implement the diff shown above in [this comment](https://github.com/dask/distributed/pull/169#issuecomment-193890173)

I've implemented that diff in a commit and pushed to a separate testing branch on github.  So you should get those lines if you do the following:

```
git remote update mrocklin
git cherry-pick 6431c0b
```

mrocklin:
After that, the output of `py.test distributed -s --verbose` should look pretty useful to you.  Please let me know if it doesn't

martindurant:
Gist updated

mrocklin:
Any thoughts or observations on the results?

martindurant:
Don't see any too-many-files errors now, but the number hovers in the low 200s from test_executor on, so whether or not it hits the tight mac limit may depend on specific invocation. 
I don't now know if the timeout/hang at the end is related or not.

mrocklin:
I wouldn't be surprised to learn that the OSX limit is 256, so maybe it's hitting that limit and then staying under?  On my machine I didn't get above 80.  So something is different.

Two thoughts:

1.  Try the timeout test on its own with `py.test distributed/test_executor.py::test_name`.  See if it timeouts when run in isolation
2.  Try to hunt down some theme among the tests that increase the number of open file descriptors?

I'll do the second in a while if you don't beat me to it. (working on csv formatting at the moment)

martindurant:
1. All passed in `test_executor` except `test_Future_exception_sync`(due to extra line in the output with number of fds). At soon as some nannies are started, the fds are always above 150, sometimes around 200, one instance at 250 in`test_start_is_idempotent`.

mrocklin:
Here is a printout from my machine.  I don't seem to be running into the same problem: https://gist.github.com/mrocklin/dbc32977fdabbc1ad794

mrocklin:
... which is unfortunate.  It's tricky to resolve something without a physically failing machine in front of you.

mrocklin:
I'd like to resolve conflicts and merge this.  @martindurant you had some fixes on top of this somewhere right?  Where did those end up?

martindurant:
https://github.com/martindurant/distributed/tree/cleanup-streams

mrocklin:
OK, I've rebased, merged in @martindurant 's commits, and added timeouts to the `join` calls.  Merging this on passed tests

PR #171, opened 2016-03-08
Format CSV

mrocklin:
This adds a new formats submodule with a read_csv function intended to share code between s3/hdfs and to separate out tricky logic.

`formats.csv.read_csv` consumes the following:

    Parameters
    ----------
    block_lists: list of lists of futures of bytes
        The lists of bytestrings with each list corresponding to one logical file
    header: bytestring
        The header, found at the front of the first file, to be prepended to
        all blocks
    head: pd.DataFrame
        An example Pandas DataFrame to be used for metadata
    kwargs: dict
        Keyword arguments to pass down to ``pd.read_csv``
    lazy: boolean, optional (defaults to True)
    collection: boolean, optional (defaults to True)

We then use this within S3 as a proof of concept.  I'll move on to hdfs3 shortly if people think that this is a reasonable step forward.

### DIFF
--- a/distributed/formats/csv.py
+++ b/distributed/formats/csv.py
@@ -1,0 +1,91 @@
+from __future__ import print_function, division, absolute_import
+
+from io import BytesIO
+
+from dask import do
+from dask.dataframe import from_imperative
+import pandas as pd
+
+from .compression import compressors, decompressors
+
+from ..executor import default_executor, ensure_default_get
+from ..utils import ensure_bytes, log_errors
+
+
+def bytes_read_csv(b, header, kwargs):
+    """ Convert a block of bytes to a Pandas DataFrame
+
+    Parameters
+    ----------
+    b: bytestring
+        The content to be parsed with pandas.read_csv
+    header: bytestring
+        An optional header to prepend to b
+    kwargs: dict
+        A dictionary of keyword arguments to be passed to pandas.read_csv
+
+    See Also:
+        distributed.formats.csv.read_csv
+    """
+    with log_errors():
+        compression = kwargs.pop('compression', None)
+        b2 = decompressors[compression](b)
+        bio = BytesIO()
+        if header:
+            if not header.endswith(b'\n') and not header.endswith(b'\r'):
+                header = header + ensure_bytes(kwargs.get('lineterminator', b'\n'))
+            bio.write(header)
+        bio.write(b2)
+        bio.seek(0)
+        return pd.read_csv(bio, **kwargs)
+
+
+def read_csv(block_lists, header, head, kwargs, lazy=True, collection=True,
+        executor=None):
+    """ Convert blocks of bytes to a dask.dataframe or other high-level object
+
+    This accepts a list of lists of futures/values of bytes where each list
+    corresponds to one file, and the futures/values of bytes concatenate to
+    comprise the entire file, in order.
+
+    Parameters
+    ----------
+    block_lists: list of lists of futures of bytes
+        The lists of bytestrings with each list corresponding to one logical file
+    header: bytestring
+        The header, found at the front of the first file, to be prepended to
+        all blocks
+    head: pd.DataFrame
+        An example Pandas DataFrame to be used for metadata
+    kwargs: dict
+        Keyword arguments to pass down to ``pd.read_csv``
+    lazy: boolean, optional (defaults to True)
+    collection: boolean, optional (defaults to True)
+
+    Returns
+    -------
+    A dask.dataframe, or list of futures or values, depending on the value of
+    lazy and collection.
+    """
+    executor = default_executor(executor)
+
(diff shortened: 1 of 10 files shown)


distributed/formats/csv.py
```diff
@@ -0,0 +1,89 @@
+import pandas as pd
```
mrocklin: @jcrist the code in this file may interest you.  I'm retrying the `read_csv` problem with the approach laid out in https://github.com/dask/dask/issues/1022

mrocklin:
@martindurant if you have a chance this could use review

distributed/formats/csv.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    from io import BytesIO
+    import pandas as pd
+    with log_errors():
+        compression = kwargs.pop('compression', None)
+        b2 = decompressors[compression](b)
+        bio = BytesIO()
+        if header:
+            if not header.endswith(b'\n') and not header.endswith(b'\r'):
+                header = header + ensure_bytes(kwargs.get('lineterminator', b'\n'))
```
martindurant: It doesn't appear obvious to me that passing around the text header is the right way - shouldn't it be a list of column names (strings) or perhaps an empty dataframe?

distributed/formats/tests/test_csv.py
```diff
@@ -0,0 +1,134 @@
...
+    assert df.id.sum() == 1 + 2 + 3
+
+
+
+@gen_cluster(executor=True)
+def test_read_csv(e, s, a, b):
+    bytes = [files[k] for k in sorted(files)]
+    gzbytes = [gzip_compress(b) for b in bytes]
+    kwargs = {}
+    head = bytes_read_csv(files['2014-01-01.csv'], '', {})
```
martindurant: Is the expectation that the user does this every time when calling `read_csv`? This makes it one step longer than assuming the file and/or keywords had everything required.

distributed/formats/tests/test_csv.py
```diff
@@ -0,0 +1,134 @@
...
+
+@gen_cluster(executor=True)
+def test_kwargs(e, s, a, b):
+    blocks = [files[k] for k in sorted(files)]
+    blocks = yield e._scatter(blocks)
+    blocks = [[b] for b in blocks]
+    kwargs = {'usecols': ['name', 'id']}
+    head = bytes_read_csv(files['2014-01-01.csv'], '', kwargs)
+
+    df = read_csv(blocks, header, head, kwargs, lazy=False, collection=True)
```
martindurant: I'm assuming a mismatch between header, head and kwargs would cause this to fail

distributed/formats/csv.py
mrocklin: Unfortunately we're bound by what the `pd.read_csv` function can do for us.  We can't give it a dataframe and say "please make something like this" we need to give it a file object and a set of keyword arguments.  Experience with blaze and with dask.dataframe is that it's extremely difficult to produce the right keyword arguments to produce the desired result.

distributed/formats/tests/test_csv.py
mrocklin: The intention is that the user never sees this code.  This is for use by other internal modules, like `s3` and `hdfs`, who will have their own, simpler, `read_csv` functions.

distributed/formats/tests/test_csv.py
mrocklin: Yes, as in the previous comment, this is for internal use only.

martindurant:
What will the main end-user function look like here? Something like this?
`read_csv('hdfs://myhost/path/files*csv', usecols=['name', 'id'])`
where we decide whether to grab blocks of bytes from hdfs, s3 or wherever?

distributed/formats/tests/test_csv.py
martindurant: OK, sorry, didn't see this. The same patterns will work for avro, not so sure about parquet (which works by byte offsets in a file, the header being at the end).

mrocklin:
I don't know, perhaps?

At the moment I'm content with `s3.read_csv`, `hdfs.read_csv`.  I agree that at some point it would be nice to dispatch on protocol or whatever, I don't yet see an immediate pressing need for it though (no one is asking for it.)

This current refactor motivated by the need for robust keyword argument handling, rather than by a desire to unify the API.  I just didn't want to write the same error-prone logic twice.

I'd like to wait until people are actively using `s3.read_csv` and `hdfs3.read_csv` and having difficulty with them before devoting significant resources to their merger.

martindurant:
OK, agreed. I don't have anything further to add - the code looks good.

mrocklin:
I've started using `formats.csv`  in `hdfs.read_csv`.  The result ended up being fairly nice.  I've removed the WIP label.  Merging on passed tests if no comments

PR #173, opened 2016-03-09
Handle unserializable exceptions/tracebacks

mrocklin:
Previously we would fail to send back exceptions/tracebacks that were
not serializable.  This occurred for pathological exceptions and for
tracebacks from recursion limit errors.  Now we are robust to these issues,
sending either string versions or nothing at all.


mrocklin:
Merging soon if no comments

PR #175, opened 2016-03-09
Remove _read_foo functions

mrocklin:
Remove the asynchronous `_read_foo` functions from the s3/hdfs APIs.

We now handle things better so that we don't actually need to go asynchronous at any point.  All metadata collection happens synchronously (and quickly) through the hdfs or s3 APIs.


mrocklin:
Merging soon if no comments

PR #174, opened 2016-03-09
S3 with writing

martindurant:
with retried on downloads

PR #176, opened 2016-03-09
S3fs

mrocklin:
Continuation of #174 

This separates the S3 filesystem code from the S3 read_bytes/text/csv code. 

It also includes retry logic for `S3File.read`

Notably it does not include S3 functionality beyond that.


mrocklin:
@martindurant this could use your review, particularly the `Update s3 retry code` commit

distributed/s3fs.py
```diff
@@ -0,0 +1,424 @@
...
+        tok = tokenize(anon, key, secret, kwargs)
+        if tok not in self._conn:
+            logger.debug("Open S3 connection.  Anonymous: %s",
+                         self.anon)
+            if self.anon:
+                from botocore import UNSIGNED
+                conf = Config(connect_timeout=self.connect_timeout,
+                              read_timeout=self.read_timeout,
+                              signature_version=UNSIGNED)
+                s3 = boto3.Session().client('s3', config=conf)
```
martindurant: The non-anon version below allows for extra kwargs. We never test any (I don't know what they could be).

martindurant:
The serialisation thing is completely unconnected?

mrocklin:
The serialization thing came up because boto3's Exception types, notably `ClientError` don't serialize well and so I wasn't able to actually tell what was going on.  So yes, unconnected to the code but highly connected to the solved problem.

distributed/s3fs.py
mrocklin: Lets expand this in later work.  I'd like to get this in quickly.

martindurant:
OK, I'm 95% sure I won't have more to add, but it is too late to be sure.

PR #189, opened 2016-03-15
StreamClosedError: Stream is closed is raised by executor.run instead of expected exception

ogrisel:
Here is a non regression test to reproduce the issue. This is the part of the test log that show the expected exception being logged and the unexpected `StreamClosedError` being raised to the user.

```
distributed.core - ERROR - informative message
Traceback (most recent call last):
  File "/Users/ogrisel/code/distributed/distributed/core.py", line 177, in handle_stream
    result = yield gen.maybe_future(handler(stream, **msg))
  File "/Users/ogrisel/venvs/py35/lib/python3.5/site-packages/tornado/gen.py", line 1008, in run
    value = future.result()
  File "/Users/ogrisel/venvs/py35/lib/python3.5/site-packages/tornado/concurrent.py", line 232, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
  File "/Users/ogrisel/venvs/py35/lib/python3.5/site-packages/tornado/gen.py", line 267, in wrapper
    result = func(*args, **kwargs)
  File "/Users/ogrisel/venvs/py35/lib/python3.5/types.py", line 243, in wrapped
    coro = func(*args, **kwargs)
  File "/Users/ogrisel/code/distributed/distributed/worker.py", line 311, in run
    result = function(*args, **kwargs)
  File "/Users/ogrisel/code/distributed/distributed/tests/test_executor.py", line 2176, in raise_exdistributed.core - ERROR - informative message
Traceback (most recent call last):
  File "/Users/ogrisel/code/distributed/distributed/core.py", line 177, in handle_stream
    result = yield gen.maybe_future(handler(stream, **msg))
  File "/Users/ogrisel/venvs/py35/lib/python3.5/site-packages/tornado/gen.py", line 1008, in run
    value = future.result()
  File "/Users/ogrisel/venvs/py35/lib/python3.5/site-packages/tornado/concurrent.py", line 232, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
  File "/Users/ogrisel/venvs/py35/lib/python3.5/site-packages/tornado/gen.py", line 267, in wrapper
    result = func(*args, **kwargs)
  File "/Users/ogrisel/venvs/py35/lib/python3.5/types.py", line 243, in wrapped
    coro = func(*args, **kwargs)
  File "/Users/ogrisel/code/distributed/distributed/worker.py", line 311, in run
    result = function(*args, **kwargs)
[...]

PR #190, opened 2016-03-15
Catch exceptions in Executor.run

minrk:
re-raise them locally, rather than aborting without any reply. Avoids StreamClosedError when run would fail.

I think it makes the most sense to add this kind of error handling to [a higher level](https://github.com/dask/distributed/blob/3a49b048ae277448748fa6a14f2a4de61ed60907/distributed/core.py#L180) in the RPC dispatch, but I wasn't sure how to do that without making a mess, so I added it directly to run handler.

Folow-up on #189, including @ogrisel's test and making it pass.

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -876,12 +876,18 @@ class Executor(object):
     @gen.coroutine
     def _run(self, function, *args, **kwargs):
         workers = kwargs.pop('workers', None)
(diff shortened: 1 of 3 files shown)


mrocklin:
This looks good to me.  Thank you both. 

I'll think about how to handle exceptions more generally in `rpc` in future work.

Issue #191, opened 2016-03-15
Handle errors robustly within RPC


mrocklin:
Currently this just provides non-informative StreamClosedErrors.  It would be good to send back the exception and reraise locally

PR #192, opened 2016-03-15
Raise uncaught exceptions through rpc

mrocklin:


mrocklin:
Fixes #191 

This wouldn't have resolved #190 but I'm actually not sure that it should have.  It's not clear how Scheduler.broadcast should handle worker exceptions.

PR #194, opened 2016-03-16
Bokeh DataTable output of status.json http endpoint

canavandl:
potential UI element for distributed:

screen shot 2016-03-16 at 5 55 00 pm

Notebook code to run:
```python
from distributed import Executor, progress

e = Executor('10.20.20.182:8786')
e

from distributed.diagnostics.status_monitor import Status_Monitor
Status_Monitor(('10.20.20.182', 9786))

import dask.array as da
x = da.random.normal(10, 0.1, size=(1000000000,), chunks=(1000000,))
y = (x + 1).mean()
future = e.compute(y)
```

PR #200, opened 2016-03-21
Status diagnostics

mrocklin:
This does a few things:

1.  Pulls some bokeh web UI work from #194 by @canavandl
2.  Adds `Scheduler.host_info` state onto the scheduler that tracks manages cpu, memory, latency, etc. across workers grouped by host
3.  Split up the old `status.json` route to workers (now improved) and tasks
4.  Split up @canavandl's work to separate functions, unfortunately drop the timeseries plot for now
5.  Add a call to `bokeh serve` directly within the `dscheduler` command line application, so this starts up automatically

```
mrocklin@workstation:~/workspace/distributed$ dscheduler 
distributed.scheduler - INFO - Start Scheduler at:        192.168.1.141:8786
distributed.scheduler - INFO -            http at:        192.168.1.141:9786
distributed.scheduler - INFO -  Start Bokeh UI at:        http://localhost:5006/status/
```

image


distributed/cli/dscheduler.py
```diff
@@ -42,11 +45,23 @@ def main(center, host, port, http_port):
         loop.run_sync(scheduler.sync_center)
     scheduler.start(port)
 
+    try:
+        import bokeh
+    except ImportError:
+        pass
+    else:
+        logger.info
```
canavandl: This need to be remove?

distributed/diagnostics/status_monitor.py
```diff
@@ -0,0 +1,128 @@
...
+        self.client = AsyncHTTPClient()
+
+        self.loop = loop or IOLoop.current()
+        self.loop.add_callback(self.update)
+        self._pc = PeriodicCallback(self.update, self.interval, io_loop=self.loop)
+        self._pc.start()
+
+    def _ipython_display_(self, **kwargs):
+        show(self.output)
+        # show(vplot(self.task_table, self.worker_table))
```
canavandl: Probably remove this.

canavandl:
Looks good to me.

What's the current barrier to the timeseries plot (maintaining plot state?). Let me know if I can help.

mrocklin:
You mentioned that the plot wasn't updating correctly.  For the moment I've just taken it out rather than try to diagnose the issue.  We can add it, or other plots, as we get them up and running, I just wanted something minimal that I would feel comfortable merging soon.

We might also want to find a general solution to maintaining history on more of these values.  We should probably find a way to do this across sessions.

distributed/cli/dscheduler.py
mrocklin: Fixed

distributed/diagnostics/status_monitor.py
mrocklin: Fixed

mrocklin:
Some general thoughts for future changes

*  Column width adjustment
*  We should move beyond text tables for some of these things.  The task table could be changed into a multi-progress bar
*  The worker data can be split between fast changing (cpu, processing) and slowly changing (total memory, processes, latency).  Many of the fast-changing attributes are amenable to visual plots rather than tables.
*  We could adapt the MultiProgress plugin currently used for progressbars to instead track all tasks in all sections (ready, waiting, finished, ...)  This might be fun.

distributed/diagnostics/bokeh/status/server_lifecycle.py
```diff
@@ -0,0 +1,44 @@
...
+                           'deque': deque(maxlen=1000),
+                           'times': deque(maxlen=1000),
+                           'condition': Condition()}
+    server_context.add_periodic_callback(lambda: http_get('workers'), 1000)
+
+    messages['tasks'] = {'interval': 100,
+                         'deque': deque(maxlen=1000),
+                         'times': deque(maxlen=1000),
+                         'condition': Condition()}
+    server_context.add_periodic_callback(lambda: http_get('tasks'), 100)
```
mrocklin: @canavandl 

OK, so now we grab the messages from the scheduler on a per-process basis, rather than a per-session/tab basis.  Moreover, we record the last thousand messages we've seen from the scheduler so that, if necessary, plots can query message history.

Now, rather than waiting on a `client.fetch` future, code should wait on `condition.wait()` for their particular message stream.  They'll be woken up every time a new message comes in.

mrocklin: @canavandl your feedback on this would be welcome.  For the plots you're thinking about how does this feel?  What could be made less awkward about it?  Is this sufficient state so that future plots can manage without tracking their own state?

canavandl: I like the concept of message deques that are managed by the scheduler and not bokeh server sessions.

In terms of plotting, are you imagining visualizing only the deque data (~16min at 1000 messages @ 1 message/sec default) or appending data the plot's CDS upon a new bokeh server session?

mrocklin: We could go either way.  We could also increase the number of messages we store.

canavandl: So how long (in terms of time) of processes are you intending on visualizing? 4hr? 24hr? More? Obviously we intend on making the sampling interval configurable, but I'm curious.

mrocklin: Short answer is I don't know.  For CPU/Memory usage I'm usually only interested in the last few seconds.  I can't anticipate other use cases well though.

mrocklin:
I'm feeling fairly content with what's here.  I think we're at a good point to merge in after I rebase a few things down.

Some possible visual next steps might include the following:

*  Change the task table to be more visually appealing
*  Combine the `processing` dicts like `{inc: 1, add: 2}` into something visual that shows off what the cluster as a whole is working on (not broken down per-worker)
*  Clean up the column widths (though this is probably low priority until we know we're going to keep the tables)
*  Plot cpu/memory/disk/net activity for selected workers.  I'm not quite sure of the best way to do this.  Do we want different page that focuses on a worker?  Do we want to show many workers plots at once?  Perhaps based on the selection of workers in the worker table?

@canavandl can you think of other opportunities here?

Some data gathering next steps include the following:

*  Collect start/stop times of all tasks in the scheduler (this is large) and stream them to the bokeh server (probably via tcp) to get something like the following: https://www.youtube.com/watch?v=sUTYlnCcgTg&feature=youtu.be
*  Serve worker log files (this is a common pain point)
*  Gather more Scheduler statistics like tasks/second, cpu load, active clients

mrocklin:
OK, merging for now, mostly in order to sync up to master and avoid a long-running branch.

PR #203, opened 2016-03-23
Task stream

mrocklin:
This creates a stream of task-completed/failed messages that we send to the bokeh server.  Ideally we then produce a streaming visualization of task-completion times similar to the `dask.diagnostics.Profiler`.

cc @canavandl @jcrist

### DIFF
--- a/distributed/diagnostics/bokeh/status/main.py
+++ b/distributed/diagnostics/bokeh/status/main.py
@@ -6,7 +6,6 @@ import json
 
 from bokeh.plotting import curdoc, vplot
 from toolz import valmap
-from tornado import gen
 
 from distributed.diagnostics.status_monitor import (
         worker_table_plot, worker_table_update, task_table_plot,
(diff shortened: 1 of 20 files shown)


mrocklin:
Updated to include example messages in the docstring: 

```python
+        {'key': 'mykey', 'worker': 'host:port', 'status': status,
 +         'compute-start': time(), 'compute-stop': time(),
 +         'transfer-start': time(), 'transfer-stop': time(),
 +         'other': 'junk'}
 +
```

Where ``status`` is either 'OK', or 'error'

The bokeh server now maintains the last 100k such messages from the scheduler, updated every 100ms.

mrocklin:
Transfer start/stop are omitted if there was no data transfer

mrocklin:
I've merged this work with @canavandl 's work at https://github.com/dask/distributed/pull/204.

Here is a fancy gif

bokeh-diagnostic

mrocklin:
Rebased and merged manually

PR #204, opened 2016-03-23
Add cpu and memory profiler per worker

canavandl:
Todo:

Remove try/except loop in callback re: message parsing
Add labels
Add styling

ping @mrocklin

### DIFF
--- a/distributed/diagnostics/bokeh/status/main.py
+++ b/distributed/diagnostics/bokeh/status/main.py
@@ -44,8 +44,8 @@ resource_source, resource_plot = resource_profile_plot()
 def resource_update():
     with log_errors():
         yield messages['workers']['condition'].wait()
-        worker_buffer = messages['workers']['deque']
-        times_buffer = messages['workers']['times']
+        worker_buffer = list(messages['workers']['deque'])
+        times_buffer = list(messages['workers']['times'])
         resource_profile_update(resource_source, worker_buffer, times_buffer)
 doc.add_periodic_callback(resource_update, messages['workers']['interval'])
 
--- a/distributed/diagnostics/bokeh/status/main.py
+++ b/distributed/diagnostics/bokeh/status/main.py
@@ -10,6 +10,10 @@ from tornado import gen
 from distributed.diagnostics.status_monitor import (
         worker_table_plot, worker_table_update, task_table_plot,
         task_table_update)
+from distributed.diagnostics.worker_monitor import (
+        resource_profile_plot, resource_profile_update
+)
+
 from distributed.utils import log_errors
 import distributed.diagnostics
 
@@ -35,4 +39,14 @@ def task_update():
         task_table_update(task_source, msg)
 doc.add_periodic_callback(task_update, messages['tasks']['interval'])
 
-doc.add_root(vplot(worker_table, task_table))
+resource_source, resource_plot = resource_profile_plot()
+@gen.coroutine
+def resource_update():
+    with log_errors():
+        yield messages['workers']['condition'].wait()
+        worker_buffer = messages['workers']['deque']
+        times_buffer = messages['workers']['times']
+        resource_profile_update(resource_source, worker_buffer, times_buffer)
+doc.add_periodic_callback(resource_update, messages['workers']['interval'])
+
+doc.add_root(vplot(worker_table, task_table, resource_plot))
--- a/distributed/diagnostics/tests/test_worker_monitor.py
+++ b/distributed/diagnostics/tests/test_worker_monitor.py
@@ -13,15 +13,13 @@ def test_resource_monitor_plot(s, a, b):
     while any('last-seen' not in v for v in s.host_info.values()):
         yield gen.sleep(0.01)
 
-    times_buffer = deque([datetime.datetime(2016, 3, 23, 14, 31, 4, 426461),
-                          datetime.datetime(2016, 3, 23, 14, 31, 5, 430657),
-                          datetime.datetime(2016, 3, 23, 14, 31, 6, 432290)],
-                          maxlen=1000)
-    workers_buffer = deque([{},
-                            {'10.10.20.86': {'cpu': 15.9, 'memory-percent': 63.0}},
-                            {'10.10.20.86': {'cpu': 14.9, 'memory-percent': 64.0},
-                             '10.10.20.87': {'cpu': 13.9, 'memory-percent': 64.0}}],
-                             maxlen=1000)
+    times_buffer = [datetime.datetime(2016, 3, 23, 14, 31, 4, 426461),
+                    datetime.datetime(2016, 3, 23, 14, 31, 5, 430657),
+                    datetime.datetime(2016, 3, 23, 14, 31, 6, 432290)]
+    workers_buffer = [{},
+                      {'10.10.20.86': {'cpu': 15.9, 'memory-percent': 63.0}},
+                      {'10.10.20.86': {'cpu': 14.9, 'memory-percent': 64.0},
+                       '10.10.20.87': {'cpu': 13.9, 'memory-percent': 64.0}}]
 
     source, plot = resource_profile_plot()
     resource_profile_update(source, workers_buffer, times_buffer)
--- a/distributed/diagnostics/tests/test_worker_monitor.py
+++ b/distributed/diagnostics/tests/test_worker_monitor.py
@@ -1,0 +1,41 @@
+from distributed.diagnostics.worker_monitor import (resource_profile_plot,
+        resource_profile_update)
+from distributed.diagnostics.scheduler import workers, tasks
+from distributed.utils_test import gen_cluster
+
+from collections import deque
+import datetime
+from tornado import gen
+
+
+@gen_cluster()
+def test_resource_monitor_plot(s, a, b):
+    while any('last-seen' not in v for v in s.host_info.values()):
+        yield gen.sleep(0.01)
+
+    times_buffer = deque([datetime.datetime(2016, 3, 23, 14, 31, 4, 426461),
+                          datetime.datetime(2016, 3, 23, 14, 31, 5, 430657),
+                          datetime.datetime(2016, 3, 23, 14, 31, 6, 432290)],
+                          maxlen=1000)
+    workers_buffer = deque([{},
+                            {'10.10.20.86': {'cpu': 15.9, 'memory-percent': 63.0}},
+                            {'10.10.20.86': {'cpu': 14.9, 'memory-percent': 64.0},
+                             '10.10.20.87': {'cpu': 13.9, 'memory-percent': 64.0}}],
+                             maxlen=1000)
+
+    source, plot = resource_profile_plot()
+    resource_profile_update(source, workers_buffer, times_buffer)
+
+    assert source.data['workers'] == ['10.10.20.86', '10.10.20.87']
+    assert len(source.data['cpu']) == 2
+    assert source.data['cpu'][0] == ['null', 15.9, 14.9]
+    assert source.data['cpu'][1] == ['null', 'null', 13.9]
+
+    assert source.data['times'][0] == ['null',
+                                       datetime.datetime(2016, 3, 23, 14, 31, 5, 430657),
+                                       datetime.datetime(2016, 3, 23, 14, 31, 6, 432290)]
+    assert source.data['times'][1] == ['null',
+                                       'null',
+                                       datetime.datetime(2016, 3, 23, 14, 31, 6, 432290)]
+    assert len(source.data['times']) == 2
+    assert len(source.data['memory-percent']) == 2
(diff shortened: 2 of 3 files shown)


canavandl:
Default styling aside - you see that there are two sets of lines (profiling two workers)

screen shot 2016-03-23 at 6 48 41 pm

distributed/diagnostics/worker_monitor.py
```diff
@@ -0,0 +1,35 @@
...
+    for name in ['cpu', 'memory-percent']:
+        try:
+            data[name] = [[msg[w][name] for msg in worker_buffer if msg != {} and w in msg] for w in workers]
+        except KeyError:
+            pass
+
+    data['workers'] = workers
+    data['times'] = [[t for i, t in enumerate(times_buffer) if worker_buffer[i] != {}] for w in workers]
+
+    source.data.update(data)
```
mrocklin: Can I ask for a test similar to what is in `diagnostics/tests/test_status_monitor.py`.  I also find that writing to a test first before interacting with bokeh makes me develop much more quickly.  The test can setup everything and try it far faster than I can by manually setting up scheduler, workers, etc..

To test a particular testing file I run the following command from within the repository's root directory.

    py.test distributed/diagnostics/tests/test_status_monitor.py

If you copy the model in that file will spin up a scheduler, workers, and executor for you and shut them down cleanly in around a second.

distributed/diagnostics/worker_monitor.py
```diff
@@ -0,0 +1,35 @@
...
+def resource_profile_update(source, worker_buffer, times_buffer):
+    data = defaultdict(list)
+
+    workers = sorted(worker_buffer[-1], reverse=True)
+
+    import pdb; pdb.set_trace()
+
+    for name in ['cpu', 'memory-percent']:
+        try:
+            data[name] = [[msg[w][name] for msg in worker_buffer if msg != {} and w in msg] for w in workers]
```
mrocklin: If a worker cuts out for a bit or comes in at the beginning or end of the period are we in a bit of trouble here?  I suspect that the times values won't line up well with these values.

mrocklin: One could pad with zeros I like in the following I suppose:

```python
workers = sorted(worker_buffer[-1], reverse=True)
cpu = {w: [] for w in workers}
memory = {w: [] for w in workers}
for msg in worker_buffer[-n:]:
    for w, L in cpu.items():
        L.append(msg.get(w, 0))
    for w, L in memory.items():
        L.append(msg.get(w, 0))

times = times_buffer[-n:]
```

mrocklin: Padding with zeros seems a bit janky though maybe.  

Does this require separate plots overlaid on the same figure?  Is that even possible to do just by updating the source?

distributed/diagnostics/worker_monitor.py
canavandl: Sure - I shouldn't have been so lazy and written tests while I was coding.

distributed/diagnostics/worker_monitor.py
canavandl: Good catch. 
The two things (AFAICT) to do are:
* have the worker list be the set of all workers in the buffer (not just in the last message)
* pass `None` values instead of zero-pad (which will cause Bokeh to render gaps in the line instead of zeroes)

Doing both should handle the cases of working appearing being added/removed anytime during the plotting.

mrocklin: Oh, that sounds great.

canavandl: missing points docs: 

http://bokeh.pydata.org/en/latest/docs/user_guide/plotting.html#missing-points

canavandl:
Currently hitting a Bokeh serialization error. I would expect this to work using nan values, as it does in non-server implementations. I'll ping the Bokeh team.

```bash
distributed.utils - ERROR - Out of range float values are not JSON compliant
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/distributed-1.9.1-py3.4.egg/distributed/utils.py", line 229, in log_errors
    yield
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/distributed-1.9.1-py3.4.egg/distributed/diagnostics/bokeh/status/main.py", line 49, in resource_update
    resource_profile_update(resource_source, worker_buffer, times_buffer)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/distributed-1.9.1-py3.4.egg/distributed/diagnostics/worker_monitor.py", line 33, in resource_profile_update
    source.data = data
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/core/properties.py", line 714, in __setattr__
    super(HasProps, self).__setattr__(name, value)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/core/properties.py", line 456, in __set__
    self._real_set(obj, old, value)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/core/properties.py", line 446, in _real_set
    self._trigger(obj, old, value, hint)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/core/properties.py", line 397, in _trigger
    obj.trigger(self.name, old, value, hint)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/model.py", line 112, in trigger
    super(Model, self).trigger(attr, old, new, hint)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/util/callback_manager.py", line 104, in trigger
    self._document._notify_change(self, attr, old, new, hint)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 523, in _notify_change
    self._trigger_on_change(ModelChangedEvent(self, model, attr, old, new, serializable_new, hint))
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 507, in _trigger_on_change
    self._with_self_as_curdoc(invoke_callbacks)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 491, in _with_self_as_curdoc
    return f()
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 506, in invoke_callbacks
    cb(event)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 475, in <lambda>
    self._callbacks[receiver] = lambda event: event.dispatch(receiver)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 56, in dispatch
    super(ModelChangedEvent, self).dispatch(receiver)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 43, in dispatch
    receiver._document_patched(self)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/server/session.py", line 177, in _document_patched
    self._pending_writes.append(connection.send_patch_document(event))
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/server/connection.py", line 42, in send_patch_document
    msg = self.protocol.create('PATCH-DOC', [event])
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/server/protocol/__init__.py", line 46, in create
    return self._messages[msgtype].create(*args, **kwargs)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/server/protocol/messages/patch_doc.py", line 35, in create
    patch_json = document.create_json_patch_string(events)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/document.py", line 827, in create_json_patch_string
    return serialize_json(json)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/site-packages/bokeh/core/json_encoder.py", line 97, in serialize_json
    return json.dumps(obj, cls=encoder, allow_nan=False, indent=indent, separators=separators, sort_keys=True, **kwargs)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/json/__init__.py", line 237, in dumps
    **kw).encode(obj)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/json/encoder.py", line 192, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/lcanavan/anaconda3/envs/distributed/lib/python3.4/json/encoder.py", line 250, in iterencode
    return _iterencode(o, 0)
ValueError: Out of range float values are not JSON compliant
bokeh.util.tornado - ERROR - Error thrown from periodic callback: ValueError('Out of range float values are not JSON compliant',)
```

canavandl:
Huh - apparently setting the value to `"null"` gets around the serialization step and works on the JS side. That's probably a bug, but works for now :)

mrocklin:
Are you using `np.nan` or `None`?  I would expect `None` to flow through well.

canavandl:
I was using `float('nan')`, the serialization also choked on `None`.

canavandl:
Here's with a new worker being added at 52sec

screen shot 2016-03-23 at 8 52 04 pm

mrocklin:
Very cool

distributed/diagnostics/bokeh/status/main.py
```diff
@@ -35,4 +39,14 @@ def task_update():
...
 doc.add_periodic_callback(task_update, messages['tasks']['interval'])
 
-doc.add_root(vplot(worker_table, task_table))
+resource_source, resource_plot = resource_profile_plot()
+@gen.coroutine
+def resource_update():
+    with log_errors():
+        yield messages['workers']['condition'].wait()
+        worker_buffer = messages['workers']['deque']
+        times_buffer = messages['workers']['times']
```
mrocklin: Might want to call `list` on both of these two deques before passing them down to another function, just for cleanliness's sake.

canavandl:
now looks like:

screen shot 2016-03-24 at 12 54 29 pm

distributed/diagnostics/tests/test_worker_monitor.py
```diff
@@ -0,0 +1,41 @@
...
+
+    times_buffer = deque([datetime.datetime(2016, 3, 23, 14, 31, 4, 426461),
+                          datetime.datetime(2016, 3, 23, 14, 31, 5, 430657),
+                          datetime.datetime(2016, 3, 23, 14, 31, 6, 432290)],
+                          maxlen=1000)
+    workers_buffer = deque([{},
+                            {'10.10.20.86': {'cpu': 15.9, 'memory-percent': 63.0}},
+                            {'10.10.20.86': {'cpu': 14.9, 'memory-percent': 64.0},
+                             '10.10.20.87': {'cpu': 13.9, 'memory-percent': 64.0}}],
+                             maxlen=1000)
```
mrocklin: I would change these to straight lists.

distributed/diagnostics/bokeh/status/main.py
```diff
@@ -35,4 +39,14 @@ def task_update():
...
+        resource_profile_update(resource_source, worker_buffer, times_buffer)
```
mrocklin: Do we want the entire history?  We could either:

1.  Take only the last n messages (probably also much cheaper this way)
2.  Use a DataRange1d follow/follow_interval and xwheel_zoom

mrocklin:
The code here looks pretty slick to me

mrocklin:
Have you tried this with multiple hosts (like `127.0.0.{1,2,3,4,5,6}`)?  It would be interesting to see how scraggly things get.

canavandl:
I'll do that now ^

canavandl:
with 10 workers:

screen shot 2016-03-24 at 1 08 00 pm

it gets yellower w/ time because i'm adding more workers that are being added on top (and the new workers are on the yellow end of the color palette i'm using). Not bad for now, though different palettes/alpha might be better in future.

mrocklin:
I think I'm somewhat against colors for this.  It's a strong visual signal that often doesn't easily correspond to what I'm looking for (general activity in time).

I would suggest a single color for CPU of all workers and a single color for memory of all workers with some moderate alpha value.  Eventually we could control which workers we want to see by selecting workers in the worker_table.

canavandl:
now: 

screen shot 2016-03-24 at 2 01 10 pm

mrocklin:
This looks good to me.  I'm going to take it, make a few alterations learned in #203 , and then try to combine the two along the same time axis.

mrocklin:
Fixed the py27 issue (future imports), rebased, and merged manually into master.

PR #205, opened 2016-03-26
Depend on external s3fs library

mrocklin:
We pulled out the s3fs module to a separate library.  Now we depend on it.


mrocklin:
Playing fast and loose today.  Merging.

PR #208, opened 2016-03-28
Canavandl/responsive layout

canavandl:
* Make plots responsive (fill container and shrink down to minimum dimension, while maintaining aspect ratio). Note - THE `responsive=True` DOES NOT WORK IN FIREFOX IN THE BOKEH 0.11.1 RELEASE. We've fixed the issue in master (link), but the current release has this bug. Also note: The tables aren't responsive, just the plots.

Merging this PR, makes (in my opinion) the UI better in Chrome because it's responsive but worse in Firefox because the resulting plots are massive and cause one's CPU usage to skyrocket. I don't know if that's a non-starter.

* Nicely formats the table cells (formats memory parameters in bytes, percentages as percentages, gives reasonable significant figures)

* Prepends worker plot legend items with "Avg " 

screen shot 2016-03-28 at 5 37 36 pm

### DIFF
--- a/distributed/diagnostics/status_monitor.py
+++ b/distributed/diagnostics/status_monitor.py
@@ -19,7 +19,7 @@ with ignoring(ImportError):
     from bokeh.palettes import Spectral11
     from bokeh.models import (ColumnDataSource, FactorRange, DataRange1d,
             HoverTool)
-    from bokeh.models.widgets import DataTable, TableColumn
+    from bokeh.models.widgets import DataTable, TableColumn, NumberFormatter
     from bokeh.plotting import vplot, output_notebook, show, figure
     from bokeh.io import curstate, push_notebook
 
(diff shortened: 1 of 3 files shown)


canavandl:
With nicely formatted worker_monitor plot y_axis:

screen shot 2016-03-28 at 5 54 52 pm

distributed/diagnostics/status_monitor.py
```diff
@@ -117,11 +117,20 @@ def worker_table_plot(width=600, height="auto", **kwargs):
...
     fast_names = ['workers', 'cpu', 'memory-percent', 'processing',
             'disk-read', 'disk-write', 'network-send', 'network-recv']
     fast = DataTable(source=source, columns=[columns[n] for n in fast_names],
                      width=width, height=height, **kwargs)
+    fast.columns[1].formatter = NumberFormatter(format='0.0 %')
+    fast.columns[2].formatter = NumberFormatter(format='0.0 %')
+    fast.columns[4].formatter = NumberFormatter(format='0 b')
+    fast.columns[5].formatter = NumberFormatter(format='0 b')
+    fast.columns[6].formatter = NumberFormatter(format='0 b')
+    fast.columns[7].formatter = NumberFormatter(format='0 b')
```
mrocklin: Oh, I was wondering about how to do this

birdsarah:
@canavandl 

> Note - THE responsive=True DOES NOT WORK IN FIREFOX IN THE BOKEH 0.11.1 RELEASE. 

Are you sure? I have no knowledge of that and it's working for me on, for example: http://cecp.mit.edu/pages/co2-air-quality.html

mrocklin:
Verified that firefox chokes on my machine.

mrocklin:
Is there a way to constrain a plot to not be larger than screen size or is aspect ratio, once fixed, always fixed.

mrocklin:
Thoughts on table sizes?  Is responsiveness not doable?

canavandl:
> Is there a way to constrain a plot to not be larger than screen size or is aspect ratio, once fixed, always fixed.

The aspect ratio is always fixed. So the plot won't get wider than the screen (well, the container the plot div is in) and, when shrunk, won't get smaller than max(width, 100px).

> Thoughts on table sizes? Is responsiveness not doable?

No. Bokeh doesn't support responsive widgets, though that more an oversight than any technical limitation.

mrocklin:
Alrighty.  Merging!

birdsarah:
responsive doesn't play nicely with VBox/HBox - this is probably the source of the issues.

PR #210, opened 2016-03-29
Guard imports against pandas absence

### LINKED ISSUE #209: pandas is now a mandatory dependency for dscheduler
If pandas is not installed I get an infinite loop with the following logs:

```
tornado.httpclient.HTTPError: HTTP 500: Internal Server Error
bokeh.util.tornado - ERROR - Error thrown from periodic callback: HTTPError(500, 'Internal Server Error', HTTPResponse(_body=None,buffer=<_io.BytesIO object at 0x7f17202a1e08>,code=500,effective_url='http://localhost:9786/tasks.json',error=HTTPError(...),headers=<tornado.httputil.HTTPHeaders object at 0x7f17204ffe48>,reason='Internal Server Error',request=<tornado.httpclient.HTTPRequest object at 0x7f1720578da0>,request_time=0.005105733871459961,time_info={}))
distributed.utils - ERROR - No module named 'pandas'
  File "/volatile/ogrisel/code/distributed/distributed/utils.py", line 229, in log_errors
    yield
  File "/volatile/ogrisel/code/distributed/distributed/http/scheduler.py", line 79, in get
    from ..diagnostics.scheduler import tasks
  File "/volatile/ogrisel/code/distributed/distributed/diagnostics/scheduler.py", line 6, in <module>
    import pandas as pd
ImportError: No module named 'pandas'
```

If think this is not intended? The diagnostics callback should probably be disabled when  pandas is not available, no?

mrocklin:
cc @ogrisel 

Fixes #209

### DIFF
--- a/distributed/diagnostics/scheduler.py
+++ b/distributed/diagnostics/scheduler.py
@@ -3,7 +3,6 @@ from __future__ import print_function, division, absolute_import
 from datetime import datetime
 import os
 
-import pandas as pd
 from toolz import countby, concat, dissoc
 
 from ..utils import key_split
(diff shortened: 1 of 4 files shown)


ogrisel:
Thanks!

PR #221, opened 2016-04-12
Reduce Protocol overhead

mrocklin:
Memory copies and the use of msgpack on large bytestrings ends up being surprisingly costly.  This PR adds framing to the protocol and will eventually replace msgpack with a custom format for large bytestring data, such as come from `gather`.  This complicates things but results in significant performance improvements.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -141,6 +141,7 @@ class Server(TCPServer):
 
         Coroutines should expect a single IOStream object.
         """
+        stream.set_nodelay(True)
         ip, port = address
         logger.info("Connection from %s:%d to %s", ip, port,
                     type(self).__name__)
@@ -225,7 +226,11 @@ def write(stream, msg):
         except Exception as e:
             logger.exception(e)
             raise
-        yield stream.write(struct.pack('L', len(msg)) + msg)
+        if len(msg) > 100000:  # long message, avoid memcpy
+            yield stream.write(struct.pack('L', len(msg)))
+            yield stream.write(msg)
+        else:                  # short message, avoid tornado overhead
+            yield stream.write(struct.pack('L', len(msg)) + msg)
         logger.debug("Written %s", orig)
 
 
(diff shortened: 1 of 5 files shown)


minrk:
In IPython parallel, we handle serialization of numpy arrays and potentially other types specially, in a manner that potentially eliminates *all* in-memory copies.

Preparing an object for pickling (called canning in IPython parallel) means wrapping it in an object that knows how to reconstruct the original. By default, this does nothing, and pickle (or dill/cloudpickle) is used. But objects can register custom serialization as:

1. an object with a `get_object` method
2. a list of 0-many memory buffers

The list of buffers is the key for avoiding unnecessary copies, and allows objects with an arbitrary number of memory chunks to be serialized without unnecessary copies of those chunks. This all derives from the concept we got from zeromq that "a message is a list of frames," but you still get the benefit if you do your own framing.

In the case of numpy arrays, we pack the array's info (shape, dtype), but send the array's body raw. Then we can reconstruct the array on the receiving side.

The gist:

- on the wire, a message is a sequence of frames (destination: `zmq.send/recv_multipart`)
- serializing an object returns a wrapper object that knows how to reconstruct the original (default: no-op, pickle)
- that object may have a `.buffers` attribute, which is a list of memory chunks (bytes, memory views, etc.)
- those buffers are popped off the wrapper before pickling, and sent as a separate frame

So on the wire, you could get:

```python
[
  b'pickled object A',
  b'pickled Canned(B) with 2 buffers',
  b'B buffer 0',
  b'B buffer 1',
  b'picked Canned(C) with 1 buffer',
  b'C buffer',
]
```


[Some illustration of serialization in IPython parallel](https://nbviewer.jupyter.org/gist/minrk/c510cddb2797c7c736835c5accadc6d1)

mrocklin:
Yeah, I was looking at `CannedArray` a bit yesterday.  We're approaching something like that.  I'm not sure how close we'll get.  

What's here now performs decently well on 100MB/s network, which is our common case.  I have a somewhat hacky alternative that uses a *single* memory copy without special casing (we accept calls to Pickle generally) that I might use in the moderate future.  I need to play with what's here a bit first in real world settings to figure out what current bottlenecks are.

PR #234, opened 2016-04-18
[WIP] Test worker stops without informing stream

mrocklin:
This is a resilience fail.  We probably need to heartbeat to resolve this.


mrocklin:
Stalled.  Closing for now.

PR #239, opened 2016-04-19
Added 'linux2' as a valid platform option for client test

ofirr:


mrocklin:
Perhaps `'linux' in sys.platform`?

This same pattern occurs many times throughout the codebase.  Is it correct to replace it in all occurrences?

mrocklin:
Thanks.  This is in.

PR #241, opened 2016-04-19
Pass scheduler arguments to bokeh server with json file

mrocklin:
We pass arguments like host, port, http-port, etc from the scheduler to the bokeh server using a temporary local JSON file.


mrocklin:
cc @nfaggian are you able to test this in your setup to see if it helps to resolve the problem?

PR #243, opened 2016-04-19
Expose replication to users through Executor

mrocklin:
Examples
--------

```python
>>> x = e.submit(func, *args) 
>>> e.replicate([x])  # send to all workers
>>> e.replicate([x], n=3)  # send to three workers
>>> e.replicate([x], workers=['alice', 'bob'])  # send to specific workers
>>> e.replicate([x], n=1, workers=['alice', 'bob'])  # send to one of specific workers
>>> e.replicate([x], n=1)  # reduce replications
```

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -1220,6 +1220,52 @@ class Executor(object):
         """
         sync(self.loop, self._rebalance, futures, workers),
 
+    @gen.coroutine
+    def _replicate(self, futures, n=None, workers=None, branching_factor=2):
+        futures = futures_of(futures)
+        yield _wait(futures)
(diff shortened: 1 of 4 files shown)


mrocklin:
@broxtronix if you have a minute your thoughts on the API here are welcome.

broxtronix:
Interesting... I imagine the use case here is to make certain intermediate results more highly available on the cluster?  Does this cause each worker to run the function `func()`, or does just one worker run it and then copy it to other workers?

In any case, this API seems plenty flexible and very useful!

mrocklin:
Replication triggers copying of data once computed.

PR #246, opened 2016-04-21
Add network throughput plot

### LINKED ISSUE #245: Add Network send/recv to web UI plot
It would be nice to see network activity over time.

We already capture this data in the same way we capture the CPU/Memory data that we already plot, so the job of figuring out how to move the data into the `ColumnDataSource` isn't hard.  What *is* tricky is that we don't have a percentage, but instead have actual MB/s bandwidth counts.  We would have to add a second Y-axis (this is doable) and think a bit about how best to set the axes.

This might be an interesting task for anyone looking to get into the Bokeh Web UI, notably @martindurant

martindurant:
Actually plots the mean of the network-send values.

Note that assumes the update interval is 0.5s, but this could change;
should decide on a  single place to define the value.

The y-axis max value gets fixes to the biggest throughput seen so far.

Fixes #245

### DIFF
--- a/distributed/bokeh/tests/test_worker_monitor.py
+++ b/distributed/bokeh/tests/test_worker_monitor.py
@@ -21,8 +21,10 @@ def test_resource_monitor_plot(s, a, b):
     times_buffer = [1000, 1001, 1003]
     workers_buffer = [{},
                       {'10.10.20.86': {'cpu': 15.9, 'memory-percent': 63.0}},
-                      {'10.10.20.86': {'cpu': 14.9, 'memory-percent': 64.0},
-                       '10.10.20.87': {'cpu': 13.9, 'memory-percent': 64.0}}]
+                      {'10.10.20.86': {'cpu': 14.9, 'memory-percent': 64.0,
+                                       'network-send': 2**16},
+                       '10.10.20.87': {'cpu': 13.9, 'memory-percent': 64.0,
+                                       'network-send': 2**17}}]
 
     source, plot = resource_profile_plot()
     resource_profile_update(source, workers_buffer, times_buffer)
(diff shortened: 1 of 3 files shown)


distributed/bokeh/tests/test_worker_monitor.py
```diff
@@ -37,11 +39,17 @@ def test_resource_monitor_plot(s, a, b):
     assert len(source.data['times']) == 2
     assert len(source.data['memory-percent']) == 2
 
+    assert len(source.data['network-send']) == 2
+    assert source.data['network-send'][1] == ['null', 'null', 2**17]
```
martindurant: Am I correct in thinking this is the raw input from one of the workers?

distributed/bokeh/worker_monitor.py
```diff
@@ -59,6 +67,11 @@ def resource_append(lists, msg):
     for k in ['cpu', 'memory-percent']:
         lists[k].append(mean(pluck(k, L)) / 100)
 
+    try:
+        net = mean(pluck('network-send', L)) if list(pluck(k, L)) else 0
+    except KeyError:
```
martindurant: This was required because network-send was not included in some messages (particularly the first message at worker startup)

distributed/bokeh/worker_monitor.py
```diff
@@ -59,6 +67,11 @@ def resource_append(lists, msg):
...
+        net = 0
+    lists['network-send'].append(net / 2**20 / 0.5)
```
martindurant: This is where the value of the update interval is assumed, because the scale is a rate, not a ratio as the other lines.

distributed/bokeh/worker_monitor.py
mrocklin: You can provide `pluck` a default value, like 0

distributed/bokeh/worker_monitor.py
mrocklin: Can we use the time since last message (I think it's recorded somewhere in there) to determine the duration?

mrocklin: This is scaling by MB?  Presumably things go crazy if we go above 100MB/s?

distributed/bokeh/tests/test_worker_monitor.py
mrocklin: It looks like `resource_profile_update` is no longer in use.  Previously we used to show the cpu/memory use of *all* workers.  This has been replaced.  Presumably this test and the associated function can be removed.

mrocklin: In general I apologize for the state of all this.  It was done in a rush.

distributed/bokeh/worker_monitor.py
martindurant: The scale should look fine at 1000s of MB - could play with the label formatter.

distributed/bokeh/worker_monitor.py
```diff
@@ -60,6 +68,13 @@ def resource_append(lists, msg):
         lists[k].append(mean(pluck(k, L)) / 100)
 
     lists['time'].append(mean(pluck('time', L)) * 1000)
+    net = mean(pluck('network-send', L, 0))
+    if len(lists['time']) >= 2:
+        t1, t2 = tail(2, lists['time'])
```
mrocklin: `lists['time']` is indexible so `lists['time'][-2:]` will be both faster and probably more readable for most Python prorgrammers.

mrocklin:
This seems to work quite well for me.  Nice work!

distributed/bokeh/worker_monitor.py
martindurant: Spice wasn't allowed, integer index was.

mrocklin: Ah, right.  Deques.  Using two integer indexes will still be much faster.  `t1, t2 = lists['time'][-2], lists['time'][-1]`

This deque can get long and this function is called very often.  This line makes the computation linear over a long list rather than constant time.

martindurant: Yep, already ahead of you; I thought we were fixing the queues to be not-too-long, but this is fine now.

mrocklin: ah ha, great.

mrocklin:
Merging.   Thanks @martindurant !

PR #252, opened 2016-05-02
Add blosc as compressor

mrocklin:
This sometimes performs better or worse than LZ4.  

Currently I'm leaning towards including it but keeping LZ4 as default.

### DIFF
--- a/distributed/protocol.py
+++ b/distributed/protocol.py
@@ -63,6 +63,14 @@ with ignoring(ImportError):
     default_compression = 'lz4'
 
 
+with ignoring(ImportError):
+    import blosc
+    blosc.set_nthreads(2)
+    compressions['blosc'] = {'compress': lambda b: blosc.compress(b, clevel=5),
+                           'decompress': blosc.decompress}
+    default_compression = 'blosc'
+
+
 def dumps(msg):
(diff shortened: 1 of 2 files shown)


distributed/protocol.py
```diff
@@ -63,6 +63,14 @@
     default_compression = 'lz4'
 
 
+with ignoring(ImportError):
+    import blosc
+    blosc.set_nthreads(2)
```
ogrisel: Have you tried to play with increasing this to see its impact on compression speed?

ogrisel:
I think LZ4 is a reasonable default to but AFAIK it does not support multi-threading while blosc does. Not sure if multithreading helps so much in practice.

distributed/protocol.py
mrocklin: I found that benefits of compression fall off a bit after the first few cores.  I'm also concerned about cores fighting for resources.  Often our cores are busy doing actual work.

mrocklin:
My plan is to leave this PR as is until I run into a problem where compression significantly helps and then do some benchmarking against that and other problems.

mrocklin:
Closing this for now.  I haven't yet run into a case where the performance difference warrants the added complexity.

PR #253, opened 2016-05-02
Rewrite data ingest to use recent dask work 

mrocklin:
We recently moved a lot of logic from distributed to dask in https://github.com/dask/dask/pull/1116 .  This is the matching PR that removes and redirects logic appropriately.


mrocklin:
Blocking on https://github.com/dask/hdfs3/issues/77

mrocklin:
@martindurant are you ok with removing Avro for a while?  I'd rather remove it that move it over.  It doesn't appear to get any use.

martindurant:
OK, but I definitely think that this and other formats should feature in the long run. 
Given it's simple, could it just be documentation on how one might load some arbitrary format of the header+block style, with avro as the example?

mrocklin:
I think the thing to move this code to dask once its needed by anyone who asks for it.  Hopefully that occurs after we come across a faster in-memory library.  Right now Avro loading is very very slow in Python.  Even CSV is a far faster choice.

If you're attached to this code we can also just move it over to dask.  I just didn't want to bother given that this code doesn't seem to have any users.

martindurant:
Get rid of it - better not include than to have something slow/incomplete. I wonder if I should look into cyavro at some point... Or maybe feather/arrow will sidestep this anyway.

mrocklin:
Yes, the goal is that someone makes a nice Avro -> Arrow conversion library in C with hooks to Python.  Then we would resurrect the code that picks out proper delimited bytes and pass those bytes to that faster function.

PR #255, opened 2016-05-03
Allow a whitelist of IPs to dscheduler for Bokeh

jseabold:


mrocklin:
Looks good to me.  I've also been considering making the bokeh server accept connections on all hostnames in the future, see https://github.com/bokeh/bokeh/pull/4052 .  

I'd be curious to hear your thoughts on that.

mrocklin:
Thoughts on the following test?

```python
@pytest.mark.skipif(not sys.platform.startswith('linux'),
                    reason="Need 127.0.0.2 to mean localhost")
def test_bokeh_whitelist():
    pytest.importorskip('bokeh')

    try:
        proc = Popen(['dscheduler', '--whitelist', '127.0.0.2', '127.0.0.3'], stdout=PIPE, stderr=PIPE)
        e = Executor('127.0.0.1:%d' % Scheduler.default_port)

        while True:
            line = proc.stderr.readline()
            if b'Bokeh UI' in line:
                break

        start = time()
        while True:
            try:
                for name in ['127.0.0.2', '127.0.0.3']:
                    response = requests.get('http://%s:8787/status/' % name)
                    assert response.ok
                break
            except:
                sleep(0.1)
                assert time() < start + 5

    finally:
        with ignoring(Exception):
            e.shutdown()
        with ignoring(Exception):
            os.kill(proc.pid, signal.SIGINT)
```

jseabold:
For my purposes, I'll stick with the whitelist concept. The less default-to-open-everything, the better around here. Allowing '*' or something may be useful or a `--bokeh-open` which is off by default.

I may have some PRs along these lines for dec2 soon :)

mrocklin:
Closing.  Thanks!

mrocklin:
Sounds grand.  Just an FYI, the whitelist in bokeh server is for the server name, not for whitelisted clients (you probably know this, but it's a common misconception.)

Currently the distributed scheduler and workers are wide open on both ends.  No security has been attempted whatsoever.  So far everyone has been acting in fairly closed environments where this hasn't been an issue.  Happy to start to address it if it is now an issue.

jseabold:
Re: Bokeh whitelist. I'm still getting my head around this, but I needed to
whitelist my container host IP to get the forwarding to work correctly from
the outside. Otherwise I was getting a 403 refused.

Re: security. Good to know, and I was operating under this assumption and
in a similar environment. I'll possibly make some PRs around the edges but
no show stoppers so far.

PR #261, opened 2016-05-05
Avoid calling get_ip unnecessarily

### LINKED ISSUE #259: Requires internet connection
I was surprised that `dscheduler` and even `dscheduler --help` require an internet connection. Maybe a try/except around `global_ip` that falls back to localhost with a warning? Things should still work locally right?

mrocklin:
This takes a moment and can fail if no network interface is
available.

Fixes #259


mrocklin:
Verified that this works for me locally with networking turned off.  Merging soon if no comments.

cc @jseabold

PR #262, opened 2016-05-05
Test and fix #260 - Executor.map on Iterator hangs.

### LINKED ISSUE #260: Executor.map on Iterator hangs
The helper thread spawned when Executor.map is invoked on an Iterator never puts a StopIteration into the queue, so the iterator returned (queue_to_iterator) blocks indefinitely.
I will attach a pull request with a test which shows this and a fix (the same fix, btw, is already implemented in scatter, which follows the same pattern).

Note that without the bugfix, the test hangs, and because it is blocking in multithreading, a timeout on the event loop doesn't help. I hope to reimplement this with coroutines and PR soon.

amosonn:
https://github.com/dask/distributed/issues/260

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -524,6 +524,7 @@ class Executor(object):
             try:
                 args = [get(q) for q in qs_in]
             except StopIteration:
+                q_out.put(StopIteration)
                 break
             f = self.submit(func, *args, **kwargs)
             q_out.put(f)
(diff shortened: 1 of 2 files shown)


mrocklin:
This looks good to me.  Merging on passed tests.

mrocklin:
This is in.  Thanks @amosonn !

PR #263, opened 2016-05-05
Changed queue_to_iterator to pass StopIteration instances.

amosonn:
Now, instead of passing the StopIteration class to signal the end of the queue, we will pass a  StopIteration instance. Since in many cases, this queue is populated by some upstream iterator, we can simple pass the StopIteration that iterator issued downstream.

I updated the current test which checks that scattering StopIteration doesn't break, now it also scatters a StopIteration instance.

I also added another test which checks that indeed the proper StopIteration (with args) is being raised downstream.

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -523,8 +523,8 @@ class Executor(object):
         while True:
             try:
                 args = [get(q) for q in qs_in]
-            except StopIteration:
-                q_out.put(StopIteration)
+            except StopIteration as e:
+                q_out.put(e)
                 break
             f = self.submit(func, *args, **kwargs)
             q_out.put(f)
@@ -781,9 +781,8 @@ class Executor(object):
             else:
                 try:
                     L = [next(q_or_i)]
-                except StopIteration:
-                    L = [StopIteration]
-                    qout.put(StopIteration)
+                except StopIteration as e:
+                    qout.put(e)
                     break
 
             futures = self.scatter(L, **kwargs)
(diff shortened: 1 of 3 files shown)


mrocklin:
How would this work if the user wanted to manage the futures themselves such as in this toy example?

```python
seq = e.scatter(iter(data))
skip_seq = (future for i, future in enumerate(seq) if i % 2 == 0)  # take every other future
```

amosonn:
If you change the iterators as you demonstrated, you're bound to lose the "return value" (args of the  StopIteration) anyway, the same is true if you take the original iterator and do that:
```
seq = iter(data)
skip_seq = (future for i, future in enumerate(seq) if i % 2 == 0)  # take every other future
```
However, this wouldn't break anything, you'd just get an empty StopIteration instead of one with arguments.
I think in general, it is more polite to move around StopIteration instances than the class itself; keeping the value is just a nice bonus, e.g. for getting stack traces.

(The cooler thing I had in mind was that if you run map on something which is coroutine-like, the output is also coroutine-like, except with Futures of the remote jobs. I don't know if this quite makes sense, but it might.)

amosonn:
The breaking of the test is because I used py3 syntax inside the new test; just need to fix the return to raise StopIteration.

mrocklin:
Alright, looks good to me.  Merging on passing tests.

amosonn:
Tests fixed.

mrocklin:
Great.  Merging.

PR #278, opened 2016-05-17
Tune work stealing to be data size aware

mrocklin:
Work stealing can be inefficient if it pulls large datasets around.  We avoid these cases.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1440,8 +1440,10 @@ class Scheduler(Server):
                                        io_loop=self.loop)
         except gen.TimeoutError:
             logger.warn("Heartbeat failed for %s", host)
+            return
         except Exception as e:
             logger.exception(e)
+            return
 
         end = default_timer()
         end_time = time()
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -835,7 +835,9 @@ class Scheduler(Server):
             if address not in self.processing:
                 return
             with ignoring(AttributeError):
-                self.worker_streams[address].stream.close()
+                stream = self.worker_streams[address].stream
+                if not stream.closed():
+                    stream.close()
 
             host, port = address.split(':')
 
@@ -1377,7 +1379,8 @@ class Scheduler(Server):
         except (StreamClosedError, IOError, OSError):
             logger.info("Worker failed from closed stream: %s", ident)
         finally:
-            stream.close()
+            if not stream.closed():
+                stream.close()
             self.remove_worker(address=ident)
 
     @gen.coroutine
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -34,6 +34,8 @@ from .utils import (All, ignoring, clear_queue, get_ip, ignore_exceptions,
 
 logger = logging.getLogger(__name__)
 
+BANDWIDTH = 100e6
+
 
 class Scheduler(Server):
     """ Dynamic distributed task scheduler
@@ -185,11 +187,13 @@ class Scheduler(Server):
         self.ready = deque()
         self.unrunnable = set()
         self.idle = set()
-        self.maybe_ready = set()
+        self.maybe_idle = set()
         self.who_has = defaultdict(set)
         self.deleted_keys = defaultdict(set)
         self.who_wants = defaultdict(set)
         self.wants_what = defaultdict(set)
+
+        self.saturated = set()
 
         self.exceptions = dict()
         self.tracebacks = dict()
@@ -385,11 +389,84 @@ class Scheduler(Server):
                 self.unrunnable.add(key)
             else:
                 self.stacks[new_worker].append(key)
-                self.maybe_ready.add(new_worker)
+                self.maybe_idle.add(new_worker)
                 # self.ensure_occupied(new_worker)
         else:
             self.ready.appendleft(key)
             # self.ensure_idle_ready()
+
+    def should_steal(self, key, bandwidth=None):
+        """ Is a key good for stealing from its chosen worker?
+
+        It must have the following attributes
+
+        1.  Not have too many dependencies
+        2.  Not be restricted to run on that worker
+        3   Take less time to transfer than to compute
+        """
+        bandwidth = bandwidth if bandwidth is not None else BANDWIDTH
+        if len(self.dependencies[key]) > 10:
+            return False
+        if key in self.restrictions and key not in self.loose_restrictions:
+            return False
+
+        nbytes = sum(self.nbytes[k] for k in self.dependencies[key])
+        transfer_time = nbytes / bandwidth
+        try:
+            compute_time = self.task_duration[key_split(key)]
+            return transfer_time < compute_time
+        except KeyError:
+            return False
+
+    def work_steal(self, bandwidth=None):
+        bandwidth = bandwidth if bandwidth is not None else BANDWIDTH
+        if not self.idle or not self.saturated:
+            return
+
+        thieves = set()  # Output list
+
+        idle = iter(self.idle)  # we will walk down these two sequences
+        remove_idle = set()
+        saturated = iter(self.saturated)
+        remove_saturated = set()
+
+        thief = next(idle)
+        victim = next(saturated)
+        try:
+            while True:
+                if victim == thief:
+                    raise ValueError()
+                thieves.add(thief)  # add to output
+                n = (self.ncores[thief]  # number of tasks to consume
+                  - len(self.processing[thief])
+                  - len(self.stacks[thief]))
+                stack = self.stacks[victim]
+                while n > 0 and stack:
+                    key = stack.popleft()
+                    if self.should_steal(key):
+                        self.stacks[thief].append(key)
+                        n -= 1
+                    else:
+                        stack.appendleft(key)  # replace task in victim's stack
+                        remove_saturated.add(victim)
+                        victim = next(saturated)
+                        break
+
+                if not stack:
+                    remove_saturated.add(victim)
+                    victim = next(saturated)
+
+                if n <= 0:
+                    remove_idle.add(thief)
+                    thief = next(idle)
+        except StopIteration:
+            pass
+        for worker in remove_saturated:
+            self.saturated.remove(worker)
+        for worker in remove_idle:
+            self.idle.remove(worker)
+        logger.debug('Stolen tasks for %d workers', len(thieves))
+        return thieves
 
     def ensure_idle_ready(self):
         """ Run ready tasks on idle workers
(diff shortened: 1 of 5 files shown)


distributed/scheduler.py
```diff
@@ -478,8 +478,11 @@ def keyfunc(w):
                 if k <= 0:
                     continue
                 tasks = stack[:k]
-                good = [t for t in tasks if t not in self.restrictions]
-                bad = [t for t in tasks if t in self.restrictions]
+                bad = [t for t in tasks if t in self.restrictions or
+                        sum(self.nbytes.get(d, 1000) for d in
+                            self.dependencies[t]) > 1000000]
+                bads = set(bad)
+                good = [t for t in tasks if t not in bads]
```
mrocklin: This was a hack and needs to be replaced.

mrocklin:
@jcrist @martindurant if either of you are interested in eventually learning more about the distributed scheduler, this PR is somewhat bite-sized and could use review.

mrocklin:
OK, this fails in some workloads (takes up 80% of our scheduling budget) because it's non-constant per new task.  When we have many idle and saturated workers with tasks that can't easily be shared (too much data transfer to warrant the movement) we end up cycling though all workers over and over again on each update.

I'll change things around to only respond to newly saturated or newly idle workers.

mrocklin:
OK, this is efficient for the shuffle problem (which is particularly demanding)

distributed/scheduler.py
```diff
@@ -385,12 +387,81 @@ def mark_ready_to_run(self, key):
...
             else:
                 self.stacks[new_worker].append(key)
-                self.maybe_ready.add(new_worker)
+                self.maybe_idle.add(new_worker)
                 # self.ensure_occupied(new_worker)
         else:
             self.ready.appendleft(key)
             # self.ensure_idle_ready()
 
+    def should_steal(self, key, bandwidth=100e6):
```
martindurant: Is there any attempt to estimate the bandwidth?

distributed/scheduler.py
```diff
@@ -385,12 +387,81 @@ def mark_ready_to_run(self, key):
...
+        """ Is a key good for stealing from its chosen worker?
+
+        It must have the following attributes
+
+        1.  Not have too many dependencies
+        2.  Not be restricted to run on that worker
+        3   Take less time to transfer than to compute
+        """
+        if len(self.dependencies[key]) > 10:
```
martindurant: Could be tunable parameter. I have no good feeling for how many dependencies is "too much" - plus we don't know how many of those might also be satisfied on other machines.

distributed/scheduler.py
```diff
@@ -385,12 +387,81 @@ def mark_ready_to_run(self, key):
...
+            return False
+        if key in self.restrictions and key not in self.loose_restrictions:
+            return False
+
+        nbytes = sum(self.nbytes[k] for k in self.dependencies[key])
+        transfer_time = nbytes / bandwidth
+        compute_time = self.task_duration.get(key_split(key), 1)
```
martindurant: I understand task_duration to be how long previous versions of this same operation took for different inputs. Maybe should have some reasonable limits on it? The default time of 1s seems long, will always cause a steal for most data.

martindurant:
I see no check for whether the potential thief is on the same machine as the victim, whereas I would have thought that's important for communication bandwidth.

Do you plan to also steal based on memory exhaustion sometime?

docs/source/work-stealing.rst
```diff
@@ -0,0 +1,86 @@
...
+This is a performance optimization and not required for correctness.  Work
+stealing provides robustness in many ad-hoc cases, but can also backfire when
+we steal the wrong tasks and reduce performance.
+
+
+Task criteria for stealing
+--------------------------
+
+If a task has been specifically restricted to run on particular workers (such
+as is the case when special hardware is required) then we do not steal.
```
martindurant: Stealing amongst the set of allowed workers is prohibited for a hard restriction?

distributed/scheduler.py
mrocklin: No

distributed/scheduler.py
mrocklin: It could be, yes.  Generally dependencies tend to be in the 1-4 range or in the 1000s range, so comparing to ten is a fairly discriminating test.  It's also usually ok if we have false negatives.

distributed/scheduler.py
mrocklin: Good point.

mrocklin: Generally we have an estimate after we run the first task, but this does open us up to bad behavior on initial tasks.  Perhaps we just shouldn't steal until we have a guess.

mrocklin:
> I see no check for whether the potential thief is on the same machine as the victim, whereas I would have thought that's important for communication bandwidth.

True, although in practice intra-node communication (300MB/s + serialization) is not hugely different from inter-node communication (100MB/s + serialization) so at the moment I'm tempted to avoid caring.

> Do you plan to also steal based on memory exhaustion sometime?

Perhaps, as applications demand.

docs/source/work-stealing.rst
mrocklin: It's not prohibited, just not implemented.  Generally there is a high cost for adding any logic here (error prone, very performance sensitive) so I've tended to avoid some of the more fringe cases.  They're definitely valid, they just haven't shown yet themselves to be worth the development and maintenance cost.

mrocklin:
The general lesson so far has been that stealing work definitely helps robustness for computations.  Lots of computations are pretty terrible without it.  However at the same time it introduces a lot of complexity that is hard to reason about and verify.  It's been a bit of a rough ride so far.

distributed/scheduler.py
ogrisel: Maybe using a module level constant and/or an environment variable would allow people with high end hardware to tune this to better reflect their empirically observed bandwidth.

distributed/tests/test_executor.py
```diff
@@ -2829,3 +2829,59 @@ def test_scheduler_saturates_cores_random(e, s, a, b):
...
+
+    slow = [e.submit(slowinc, x, delay=0.1, pure=False) for i in range(4)]
+    yield _wait([slow])
+
+    assert b.data  # not empty
+
+
+@gen_cluster(executor=True, ncores=[('127.0.0.1', 1)] * 10)
+def test_worksteal_many_thieves(e, s, *workers):
+    np = pytest.importorskip('numpy')
```
ogrisel: numpy is unused here.

distributed/scheduler.py
mrocklin: Fixed this.   We don't steal when we don't have timing information.

distributed/scheduler.py
mrocklin: Added a module level constant.  I think we should specify these sorts of things with config files eventually when there are other such parameters like choice of compression.  

At the moment I'm not too concerned with bandwidth for work-stealing.  For most tasks I've seen the computation / communication ratio is either very high or very low.

distributed/tests/test_executor.py
mrocklin: Fixed

mrocklin:
Future work here I think is to have workers send tasks back to the scheduler once it's clear that they're over-burdened.  I've run into cases where this occurs, especially when the length of functions varies significantly.   This could be done by maintaining excess tasks on the worker in a queue and running a periodic callback to filter out any task that had not been sent to the thread-pool-executor since the previous iteration.

mrocklin:
If there are no other comments then I may merge this tomorrow.

ogrisel:
> Future work here I think is to have workers send tasks back to the scheduler once it's clear that they're over-burdened. I've run into cases where this occurs, especially when the length of functions varies significantly. This could be done by maintaining excess tasks on the worker in a queue and running a periodic callback to filter out any task that had not been sent to the thread-pool-executor since the previous iteration.

That's an interesting idea. I think it would be worth maintaining a suite of workload that highlight both common and pathological scheduling behaviors to make sure that we do not introduce performance regressions when refactoring the scheduling logics.

docs/source/work-stealing.rst
```diff
@@ -0,0 +1,86 @@
...
+All operations in the scheduler endeavor to be computed in constant time (or
+linear time relative to the number of processed tasks.)  We can pull from the
+bottom of the stack in constant time by implementing each worker's stack as a
+``collections.deque``.  However, we currently do not maintain the data
+structures necessary to efficiently find the most occupied workers.  Common
+solutions, like maintaining priority queue of workers by stack length add a
+``log(n)`` cost to the common case.
+
+Instead we allow Python to iterate through the set of saturated workers however
+it finds to be the most efficient.
```
ogrisel: I don't understand the phrasing of this sentence. Could you please try to make it more explicit?

distributed/scheduler.py
```diff
@@ -626,9 +689,13 @@ def ensure_occupied_ready(self, worker):
 
         self._check_idle(worker)
 
+    def issaturated(self, worker, latency=5e-3):
```
ogrisel: I think you should add a docstring to explain this heuristic:

- what this method it used for,
- what does "saturated" mean (and how this definition matches the code of the function).

mrocklin:
>  I think it would be worth maintaining a suite of workload that highlight both common and pathological scheduling behaviors to make sure that we do not introduce performance regressions when refactoring the scheduling logics.

This sounds good to me.  Can you recommend a benchmark suites from a couple of other projects?  I'd like to see how other people do this.  What do you suggest?

distributed/scheduler.py
mrocklin: Fixed I think

docs/source/work-stealing.rst
mrocklin: Rephrased

mrocklin:
Currently my solution to avoid performance regressions is to add unit tests to ensure specific behavior:

```python
@gen_cluster(executor=True, ncores=[('127.0.0.1', 1)] * 2)
def test_even_load_on_startup(e, s, a, b):
    x, y = e.map(inc, [1, 2])
    yield _wait([x, y])
    assert len(a.data) == len(b.data) == 1
```

However, I usually add these only *after* I notice that something is broken.  I've been testing applications by hand.  It'd be nice to automate alerts here.

minrk:
[Pandas](https://github.com/pydata/pandas/tree/master/asv_bench) uses [asv](http://pydata.github.io/pandas/), which is something I've had some positive experiences with, but haven't managed to fully integrate into a project. I tried to use it myself for pyzmq, but I couldn't get around the hangs that would happen for certain non-functional stages in history.

mrocklin:
I'm planning to handle performance benchmarks after this.  

This branch holds some fairly important fixes.  I'd like to merge.  Any further comments?

ogrisel:
LGTM. I have not tested this branch on a real cluster / workload but I guess we can always fine tune later.

ogrisel:
BTW +1 for asv as a bench framework. While I have never used it myself, it looks really neat.

PR #284, opened 2016-05-23
Never call scheduler.sync_center

ogrisel:
`scheduler.sync_center` has been removed is the latest scheduler refactoring but there were 2 occurrences remaining in the code base.


mrocklin:
Thanks!

PR #296, opened 2016-05-27
Better check for Windows platform when finding bokeh

danring:
Slight modification to #270 -- both `darwin` and `cygwin` match `if 'win' in sys.platform`, so instead check for `'win'` as a prefix.


mrocklin:
Ah.  Thank you for the fix.

mrocklin:
This is in.  Thanks!

PR #299, opened 2016-06-01
Stop progress widget on cancel

martindurant:


martindurant:
A couple of lines in scheduler still appear in the PR from elsewhere - please check whether they are supposed to be there or not.

martindurant:
Hm, my test passes (and it works interactively in the notebook), but the test suite hands in collections.

mrocklin:
Perhaps remove the extra lines in the scheduler and see what happens?

mrocklin:
The collections test additionally serve as a good stress test.  They often fail when the scheduler has something very slightly out of alignment.  The spurious lines in the scheduler code could be just such a misalignment.

PR #302, opened 2016-06-03
Terminating bokeh only if it is properly started

spirali:
Fixes the following exception:
`UnboundLocalError: local variable 'bokeh_proc' referenced before assignment`
It occurs when bokeh is not properly started


mrocklin:
This generally looks good to me.  The error surprises me.  I'll take a look in a bit.

PR #303, opened 2016-06-03
Fix health monitoring on machines without physical disks

spirali:
`psutil.disk_io_counters()` throws `RuntimeException` when there is no physical disk in the machine


mrocklin:
This looks good to me.  I'm curious, what do your computational nodes look like?

spirali:
The computer is SGI cluster Salomon: http://www.top500.org/system/178525
More details about hardware can be found here: https://docs.it4i.cz/salomon/hardware-overview-1

PR #305, opened 2016-06-04
Fix race condition when intermediate task finish after state refresh

hussainsultan:
/pr @mrocklin


mrocklin:
Failing tests are due, I think, to our improvement of the validate function.  The extra check that we added exposed a couple of other slight inconsistencies.  I'll clean these up and merge afterwards.

mrocklin:
This has been fixed by #306 .  Closing.

PR #313, opened 2016-06-06
Create new IOLoops for each call to dworker/dscheduler

### LINKED ISSUE #312: Programmatically launched dworkers fail to terminate cleanly
If a dworker is launched with:

```python
from distributed.cli import dworker
dworker.main.main(args=['127.0.0.1:8786'], prog_name='dworker', standalone_mode=False)
```

Then everything will run fine until an exit is attempted.  If a Ctrl+C is issued, then the following error message is displayed:

```console
distributed.worker - INFO -       Start worker at:            127.0.0.1:44731
distributed.worker - INFO -              nanny at:            127.0.0.1:43297
distributed.worker - INFO -               http at:            127.0.0.1:33287
distributed.worker - INFO - Waiting to connect to:            127.0.0.1:8786
distributed.worker - INFO -         Registered to:            127.0.0.1:8786
distributed.core - INFO - Connection from 127.0.0.1:45160 to Worker
distributed.core - INFO - Connection from 127.0.0.1:45162 to Worker
distributed.nanny - INFO - Nanny 127.0.0.1:43297 starts worker process 127.0.0.1:44731
^C
distributed.nanny - INFO - Unregister worker 127.0.0.1:44731 from center
distributed.nanny - INFO - Nanny 127.0.0.1:43297 kills worker process 127.0.0.1:44731
distributed.dworker - ERROR - Failed to cleanly kill worker 127.0.0.1:43297
```

Depending on the setup, if another dworker is launched in the same way in the same process, an IOError will be thrown because an I/O loop is already running.  

Setting the kwarg `standalone_mode=True` is not really an option because click will raise a SystemExit (see [the docs](http://click.pocoo.org/5/api/#click.BaseCommand.main)) and take the whole process with it.

Using `make_context()` and `invoke()` or using the callback directly suffer from the the same issues as presented above.

This was discovered while implementing scopatz/xonsh#1125, and so the associated code may be found there.  This is needed because the easiest and most direct way to makes sure that distributed has access to the xonsh builtins is to simply launch dworkers from within a xonsh process.

Thanks in advance!

mrocklin:
Fixes #312

cc @scopatz


mrocklin:
@scopatz can you see if this works for you?

mrocklin:
Hrm, nevermind.  I'll have to clean up a couple of things.  We were probably depending on the global loop in a couple of places.

minrk:
If you want to ensure a new IOLoop is created, but still returned by later access to `.current()`, you can use:

```python
loop =  IOLoop()
loop.make_current()
```

But only do this if you are sure that the existing loop returned by `IOLoop.current()` is defunct or hasn't been created yet. Since you call `.start()` later on, that should be okay.

Alternatively, the same issue may be fixed if you did `IOLoop.clear_current()` in your `finally:` block after `loop.start()`.

scopatz:
Thanks for taking into this @mrocklin. Let me know when you want me to try it out.

mrocklin:
To be honest I'm fairly ignorant of Tornado + multiprocessing bits.  I screwed around for a while on this but ran into issues.  

There is a non-trivial chance that this one-line change will have a decent effect on your problem.  If it is easy for you to try then I'd give it a shot.

mrocklin:
Any luck with this @scopatz ?  

Alternatively can you suggest a test that might replicate your situation?

mrocklin:
Merging soon if no comments

Issue #285, opened 2016-05-23
Support workers= in persist and compute


mrocklin:
Sometimes we want to restrict tasks to land on a subset of our cluster.

martindurant:
Noticing that in some existing functions (scatter), workers is a list of (host, port), i.e,. specific processed, but in others (submit) it is hosts only. Is there a preference for persist and compute? 
Should submit be brought into line with scatter, or vice-versa?

mrocklin:
We want to support just hostname when we think about restrictions based on hardware (e.g. does this machine have a GPU or access to particular blocks of HDFS data.)  We want to support hostname+port when we think about particular worker processes, perhaps to target specific memory spaces.  We also often want to support name aliases like `'alice'` that workers may have.

In principle all functions probably want to support all cases.  I don't currently know how comprehensive this is.  I suspect that most functions on the scheduler handle this logic on their own, which probably isn't best.

PR #315, opened 2016-06-06
Persist worker keyword

martindurant:
Beginning of thoughts on #285

### DIFF
--- a/distributed/cli/dscheduler.py
+++ b/distributed/cli/dscheduler.py
@@ -50,6 +50,7 @@ def main(center, host, port, http_port, bokeh_port, show, _bokeh, bokeh_whitelis
                           services={('http', http_port): HTTPScheduler})
     scheduler.start(port)
 
+    bokeh_proc = None
     if _bokeh:
         try:
             import bokeh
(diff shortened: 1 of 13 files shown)


mrocklin:
In principle this looks fine to me.  Some thoughts:

1.  We should add support for compute
2.  We should verify that both `a.ip` and `a.address` and `[a.ip]` all work
3.  I'm a bit curious what the behavior is if we call persist on the same collection twice with different `workers=` .  I would hope that `restrictions` would merge on the scheduler but I doubt that this is current behavior.
4.  It's not entirely clear if actual use cases want this or if they want the ability to restrict certain parts of the computation to different worker sets.

mrocklin:
But, we should probably nail down point 4 above before spending much effort on points 1-3

martindurant:
Agreed we need to make a plan on number 4).
From the code (not tested), I think in the case for 3), the most recent wins. This kind of thing would be important, since (once it's implemented) one can call compute() on something persisted/submitted, and pass different restrictions too.

mrocklin:
This is stalled.  Closing for now.

PR #317, opened 2016-06-07
Removed duplicated `hdfs.read_text` implementation

frol:


mrocklin:
This function should definitely be removed but we should probably wait a version or two to pass through a deprecation cycle.  Currently the function warns the user of the deprecation and redirects them to `db.read_text`.  Same with `distributed.hdfs.read_csv`

frol:
@mrocklin Please, look into the source file there are TWO copies of `read_text` function!

mrocklin:
Oh, wow.  Whoops!  Please excuse my lack of attention here.

PR #325, opened 2016-06-10
make joblib frontend cancel all pending tasks on first error

ogrisel:
This leverages new backend API introduced in joblib/joblib#361.

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -4,7 +4,7 @@ from joblib._parallel_backends import ParallelBackendBase, AutoBatchingMixin
 from joblib.parallel import register_parallel_backend
 from tornado import gen
 
-from .executor import Executor, _wait, Future
+from .executor import Executor, _wait
 
 
 class DistributedBackend(ParallelBackendBase, AutoBatchingMixin):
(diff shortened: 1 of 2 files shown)


distributed/joblib.py
```diff
@@ -35,5 +35,11 @@ def callback_wrapper():
         future.get = future.result  # monkey patch to achieve AsyncResult API
         return future
 
+    def abort_everything(self, ensure_ready=True):
+        # Tell the executor to cancel any task submitted via this instance
+        # as joblib.Parallel will never access those results.
+        executor = self.executor
+        executor.cancel([Future(k, executor) for k in executor.futures])
```
ogrisel: @mrocklin the current cancel method does not accept future keys as argument hence the need to manually wrap as a Future instance. I wonder if the `Executor.cancel` API should be changed to make it possible not have to create those intermediate objects. Not a big deal though.

mrocklin: It's an easy change to make.  Usually it's not a big deal because users usually don't manipulate keys directly.

Even here I might suggest that the `DistributedBackend` should explicitly track the futures that it produces from `apply_async` in a separate list.

mrocklin: This would also allow you to test the early abort functionality by asserting that the status of all futures was `cancelled`.

mrocklin: Should I make this change?  It would be easy to do.

mrocklin: I'd like to get this in.  I'm going to go ahead and do this.  We can see how it looks afterwards.

mrocklin: https://github.com/ogrisel/distributed/pull/1

ogrisel:
Indeed this is better I rebased/merged your commit on top of this. Feel free to squash the two commits.

PR #339, opened 2016-06-16
Deploy

mrocklin:
This adds a way to programatically control a local cluster.  This is both for convenience and to establish an interface for future projects on other systems (ssh, mesos, etc.)

Examples
------

```python
In [1]: from distributed.deploy import Local
In [2]: c = Local(n_workers=0, scheduler_port=12345)
In [3]: c.start_worker(ncores=2, separate_process=True)
In [4]: c.start_worker(ncores=5, separate_process=False)
In [5]: c
Out[5]: LocalCluster(192.168.1.141:12345, workers=2, ncores=7)
```

We can use this object as a direct input to the Executor.  Anything with a `scheduler_address` field is accepted.

```python
In [7]: from distributed import Executor
In [8]: e = Executor(c)
In [9]: e
Out[9]: <Executor: scheduler=192.168.1.141:12345 processes=2 cores=7>
```

By default a `Local` cluster creates a bunch of single-threaded process workers

```python
In [11]: c = Local(scheduler_port=2345)
```

And, by default, the Executor produces a `Local` cluster if given no other arguments

```python
In [12]: e = Executor()
In [13]: e
Out[13]: <Executor: scheduler=192.168.1.141:54574 processes=8 cores=8>
```

cc @ogrisel


mrocklin:
If this works well then I think we should copy the same approach for the existing Knit/YARN and SSH solutions

distributed/deploy/local.py
```diff
@@ -0,0 +1,107 @@
...
+
+from ..utils import sync, ignoring, All
+from ..executor import Executor
+from ..nanny import Nanny
+from ..scheduler import Scheduler
+from ..worker import Worker, _ncores
+
+
+class Local(object):
+    def __init__(self, n_workers=None, cores_per_worker=None, processes=True, loop=None, start=True, scheduler_port=8786, **kwargs):
```
ogrisel: Why did you choose to start processes instead of theads by default? This is not consistent with the default behavior of the `dask-worker` command.

mrocklin: So this is a bit confusing.  There are three relevant keyword arguments in this decision:

1.  n_workers
2.  cores_per_worker: similar to `--nthreads` in `dask-worker`
3.  processes: similar to `--nanny` in `dask-worker`.  Switches between creating a Worker (threadpool in this process) and a Nanny (threadpool in other process)

So if n_workers was 1 and cores_per_worker was 4 and processes was true this would start a four-threaded worker in a separate process (using a nanny).  Can you recommend better names for these keyword arguments?

ogrisel: Ok, I read too quickly. I am wondering if `threads_per_worker` would not be more explicit but then I assume we should use the "threads" terminology consistently throughout the code base and IIRC you intentionally switched from "threads" to "cores" in the `Executor` string representation in the past.

Maybe for the `processes` kwarg, it would be better to name it `use_nanny_processes` or just `used_nanny` to make that more explicit?

A docstring to explain the meaning of those parameters would help also :)

mrocklin: I've switched to threads here for now.  Also changed `processes=` to `nanny=`.  Added docstring.

mrocklin:
OK, so this PR exposed some Tornado IOLoop things that we were doing slightly uncleanly in tests, which led to a lot of unrelated work.  I may end up rebasing these out.  Until then sorry for the noise.

distributed/tests/test_joblib.py
```diff
@@ -32,14 +32,22 @@ def test_simple(loop):
             seq = Parallel()(delayed(inc)(i) for i in range(10))
             assert seq == [inc(i) for i in range(10)]
 
+            from joblib.parallel import get_active_backend
+            ba, _ = get_active_backend()
+            ba.executor.shutdown()
```
mrocklin: @ogrisel it would be nice for the backends to have the ability to clean themselves up

mrocklin:
OK, this is cleaned up. There are two major changes in this PR:

1.  The `LocalCluster` object discussed in the header comment.
2.  A stricter handling of IOLoops within the codebase, and fixing various issues that arose from this change.

Merging on passed tests.  I think that this significantly lowers the barrier to newcomers screwing around with the library

mrocklin:
Tests uncovered an intermittent failure that is, I think, unrelated.  Gist of error log here: https://gist.github.com/mrocklin/46f89ae44528f17a8311cd249802058f

PR #342, opened 2016-06-16
Fixed worker crashing with bad __repr__ for args

amosonn:
Previously, when a worker job failed, it tried to log the args and kwargs which led to a failure. In order to do so, it invoked the __repr__ of said arguments (indirectly, via __str__ of args and kwargs), and if those were broken, crashed.
Now it try..excepts these castings, and puts a stock value when they fail.
Includes test.

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -6,6 +6,8 @@ import os
 import shutil
 import sys
 import traceback
+import logging
+import re
 
 import pytest
 from toolz import pluck
@@ -17,7 +19,7 @@ from distributed.center import Center
 from distributed.core import rpc, dumps, loads, connect, read, write
 from distributed.scheduler import Scheduler
 from distributed.sizeof import sizeof
-from distributed.worker import Worker, error_message
+from distributed.worker import Worker, error_message, logger
 from distributed.utils import ignoring
 from distributed.utils_test import (loop, _test_cluster, inc, gen_cluster,
         slow, slowinc, throws, current_loop)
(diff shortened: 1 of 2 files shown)


distributed/worker.py
```diff
@@ -760,3 +763,41 @@ def apply_function(function, args, kwargs):
...
+    """ Convert args to a string, allowing for some arguments to raise
+    exceptions during conversion and ignoring them.
+    """
+    sargs = "("
+    for arg in args:
+        try:
+            sarg = repr(arg)
+        except:
+            sarg = "< could not convert arg to str >"
+        sargs += sarg + ", "
```
mrocklin: It is slightly more efficient if you make a list of strings and then create the result in one go with `join` as in `'(' + ', '.join(strings) + ')'` instead.  Otherwise we create many copies.  This isn't a big deal though (number of args is usually small).

distributed/tests/test_worker.py
```diff
@@ -55,6 +55,84 @@ def test_health():
...
+                                    who_has={'x': [a.address]})
+        assert not b.active
+        assert response['status'] == 'error'
+        assert isinstance(loads(response['exception']), ZeroDivisionError)
+        if sys.version_info[0] >= 3:
+            assert any('1 / 0' in line
+                      for line in pluck(3, traceback.extract_tb(
+                          loads(response['traceback'])))
+                      if line)
+        # TODO: test log item.
```
mrocklin: You could either redirect the `logger` using the logging module or else use the `capsys` pytest fixture to capture stderr.  See `distributed/diagnostics/tests/test_progress.py:def check_bar_completed` for an example.

distributed/worker.py
amosonn: My consideration was that we want to lazily repr() things, so if we run out of character limit we don't repr redundant things, I guess this can be achieved with length arithmetic as well. I will rewrite.

distributed/tests/test_worker.py
amosonn: will check.

distributed/worker.py
mrocklin: Oh sure, that makes sense.  Do as you think is best.

amosonn:
I added some code snippet to test the logging. It seems the right tool for the job is testfixtures, see https://pythonhosted.org/testfixtures/logging.html
I of course leave at your discretion to choose whether to adopt it; since there is quite a lot of logging in distributed, I think good logging tests could go a long way.

distributed/tests/test_worker.py
```diff
@@ -55,6 +56,109 @@ def test_health():
         assert 'network-send' in d
 
 
+def test_worker_bad_args(loop, capsys):
```
mrocklin: Can probably drop `capsys` with your `MockLoggingHandler`

distributed/tests/test_worker.py
```diff
@@ -55,6 +56,109 @@ def test_health():
...
+        if sys.version_info[0] >= 3:
+            assert any('1 / 0' in line
+                      for line in pluck(3, traceback.extract_tb(
+                          loads(response['traceback'])))
+                      if line)
+        assert hdlr.messages['warning'][0] == " Compute Failed\n" \
+            "Function: bad_func\n" \
+            "args:     (< could not convert arg to str >)\n" \
+            "kwargs:   {'k': < could not convert arg to str >}\n"
+
```
mrocklin: Are the lines below this comment relevant for this change?

distributed/tests/test_worker.py
```diff
@@ -55,6 +56,109 @@ def test_health():
         assert 'network-send' in d
 
 
+def test_worker_bad_args(loop, capsys):
+    @gen.coroutine
+    def f(c, a, b):
```
mrocklin: You may be interested in switching this out for a slightly newer testing style

```python
@gen_cluster
def test_worker_bad_str_args(c, a, b):
    ...
```

See test_executor.py for examples.  I should probably rewrite the tests in this file in the new style to avoid issues like this.

Feel free to ignore.  I can handle this later.

mrocklin:
testfixtures looks cool.  A quick google search also yields https://pypi.python.org/pypi/pytest-capturelog 

This deserves some thought.

distributed/tests/test_worker.py
amosonn: the exception checks are copy-pasted from the other test, but I wanted to leave them to make sure it is indeed the division-by-zero that break the job, not the failure to repr().
The last assert is the check that indeed the logging worked.

mrocklin: I think that you're referring to the lines above this comment.  The lines below this comment seem to be unrelated?  I could be confused here.

amosonn:
Yeah, I don't have much to say about either, no experience with testing logging either way :). But it seems the capturelog is easier to use and also more pytest ready.

distributed/tests/test_worker.py
amosonn: Ah, sorry, I thought it was odd that you wrote above, but these were the lines that were shown next to your comment.
The lines below are to check that none of the workers died as a result of processing the sad object - so they're still accepting jobs and processing them normally. Perhaps a comment is in place. I will add.

mrocklin:
This looks good to me.  The errors that we're seeing confuse me a bit.  They don't appear to be related.

amosonn:
Yeah, they also come and go: the last build succeeded (before adding the
comments). One of them is sort of a timeout, so maybe increasing that
timeout will solve it, and I have a hunch that the other is related.

mrocklin:
Verified that the `test_bad_tasks_fail` test is unrelated.  Feel free to bump the timeout to 20 seconds.  I'll do this in another branch soon as well.

amosonn:
Increasing the timeouts indeed solves test_bad_tasks_fail.
However, test_broken_worker_during_computation also fails, sometimes, and I haven't managed to understand why. But it still seems unrelated.

amosonn:
Seems that increasing the timeout still doesn't really help. So probably something is amiss. What do you recommend next?

mrocklin:
I've run into this on my own work and am handling at least some of it.  I recommend that we just let this PR sit tight for a couple of days after I've resolved some internal issues.

amosonn:
Cool, so I'll throw away the false timeout fixes.

If you have some pointer to throw me at I can try to give it a shot.

mrocklin:
I've merged in a fairly large change.  I apologize that this caused some merge conflicts on your branch.  I've gone through the rebase on by bad-worker-args if you'd like to reset to the changes.  Alternatively I can just merge from there.

mrocklin:
Ah, the branch name is `worker-bad-args`

```
git remote add mrocklin git@github.com:mrocklin/distributed.git
git remote update mrocklin
git reset --hard mrocklin/worker-bad-args
```

https://github.com/mrocklin/distributed/tree/worker-bad-args

amosonn:
Thanks!
I wasn't sure exactly what happened, so I've just rebased it myself on top of master.

amosonn:
However, the logging tests now fail on travis, even though they still pass on my machine. Should I just drop them for now?

mrocklin:
Thanks @amosonn !  This is in.

PR #353, opened 2016-06-23
Replaced 'if not in' and KeyError with setdefault, discard and pop.

amosonn:
The code of the scheduler has many places where it needs to initialize some keys by adding a set(), or clean by removing them from a set. In some places the former is done with a "if key not in...", in others with a "container[key] = container.get(key, default)"; the latter is sometimes done with "if key in", in other times with "except KeyError".
However, there are builtin methods, such as dict.setdefault, dict.pop and set.discard, which do exactly what it required, and are probably more efficient. So I tried switching to them wherever applicable.
It might be argued that this decreases readability - up to you to decide.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -584,10 +584,7 @@ class Scheduler(Server):
         for dep in deps:
             if dep in self.waiting:
                 s = self.waiting[dep]
-                try:
-                    s.remove(key)
-                except KeyError:
-                    pass
+                s.discard(key)
                 if not s:  # new task ready to run
                     self.mark_ready_to_run(dep)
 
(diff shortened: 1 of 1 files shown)


mrocklin:
Interesting.  I had never used `dict.setdefault` or `set.discard`.  I have no strong feelings on the readability issue; it might be argued that these *increase* readability as well assuming that the readers are familiar with these methods.  

Regarding performance I doubt that there is a huge impact here (profiling is generally in order for performance-based changes) and I'm not confident that it'd be a positive change.  Try-excepts and conditionals are pretty speedy in Python.  

I should publish the scripts I use to measure internal performance.

mrocklin:
I'm going to close this for now.  I think it's a fine idea but I think that we lack sufficient motivation to make a change like this.

PR #356, opened 2016-06-23
Update Web Interface to Bokeh 0.12

mrocklin:
Among several other things this greatly enhances layout

cc @birdsarah


mrocklin:
I pushed a small commit that removed a fake circle in the task-stream plot.  Historically this was here in order to make sure that there was *some* data on the plot at first render.  This used to be cleaned up by the `follow_interval` logic.  Now it sticks around, distorting the x-range.

Removing it is fine except that we fail on first render.  Everything looks fine after a resize event.  Suggestions on how to better handle this?

birdsarah:
@mrocklin - solved the template thing - packaging problem.

birdsarah:
> Removing it is fine except that we fail on first render. Everything looks fine after a resize event. Suggestions on how to better handle this?

All is well it was broken because of bad plot code.

birdsarah:
screen shot 2016-06-24 at 1 07 37 am

There's still a bunch things I'd like to do, but I think this is usable. @masoninman had the good idea of switching to progress rings to save space along the bottom - but that would require a lot more fiddling and we'd still not be able to do squashable layouts.

Please check I haven't broken the progress bars. I reversed the yaxis directions in the hope that they would add from top to bottom - but I may just have broken them - in which case I'm guessing you can restore this yourself.

The modal is more than I have time for now. 

HTH.

birdsarah:
And the messy reset is clearly a bug - so thanks for the assist on that!

birdsarah:
If you can clean the title up on the Progress plot that would make me very happy!

mrocklin:
Title cleaned up.  Happy to perform token amount of work here :)

mrocklin:
At first glance everything looks fantastic here.  I was unable to break things after a quick 10 minutes of playing.  Will stress it further in a bit

mrocklin:
Yup, things are still awesome.  I'm unable to break things. 

image

mrocklin:
Including a "before" picture for posterity

image

mrocklin:
So, some general thoughts:

1.  We've lost overall vertical space for the task stream plot, which has the most information.  Presumably we could regain this if we switched out of stretch mode. 
2.  We've lost some horizontal space on the left side of the progress bars
3.  Splitting the resource plot into two generally seems like a good move, however there is a bit of excess space in between the two plots taken up by a title (which could maybe move to the y-axis label?) and the x-axis, (which could maybe be removed if we just depended on the x-axis of the network plot?)

Given time constraints we should probably not deal with these now (or I should learn more Bokeh and deal with them myself).  @birdsarah assuming that you're about to check out can you provide a list of things you think would be good to do here in the future?

birdsarah:
> We've lost overall vertical space for the task stream plot, which has the most information. Presumably we could regain this if we switched out of stretch mode. 

yes

> We've lost some horizontal space on the left side of the progress bars

that's not necessary I was just futzing - you should be able to tweak that

>      Splitting the resource plot into two generally seems like a good move, however there is a bit of excess space in between the two plots taken up by a title (which could maybe move to the y-axis label?) and the x-axis, (which could maybe be removed if we just depended on the x-axis of the network plot?)

yaxis is too short for a label. x_axis removal will be a good move. there maybe some styling tweaks possible too

PR #359, opened 2016-06-27
Reify iterators in as_completed

### LINKED ISSUE #358: Results not returned when mapping over an iterator with as_completed
When I map a function over an iterator with `as_completed`, nothing is returned.

```python3
from distributed import Executor, as_completed

def print_and_return(x):
    print(x)
    return x

e = Executor()
things = iter(range(10))
futures = as_completed(e.map(print_and_return, things))
# Prints out the numbers 0 through 9...
list(futures)
# Returns an empty list!
```

The problem goes away if I map over a list instead:
```python3
things = list(range(10))
futures = as_completed(e.map(print_and_return, things))
# Prints the numbers 0 through 9...
list(futures)
# Returns a list of 10 futures
```

It's also fine without the `as_completed`:
```python3
things = iter(range(10))
futures = e.map(print_and_return, things)
# Prints out 0 through 9...
list(futures)
# Returns a list of 10 futures
```

This is on e0b21d1 on Python 3.5.1.

mrocklin:
Previously we would burn an input iterator in sanity checks.

Now we reify iterators to a list and then proceed.

Fixes #358

cc @Futrell

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -1593,6 +1593,7 @@ def as_completed(fs):
 
     This function does not return futures in the order in which they are input.
     """
+    fs = list(fs)
     if len(set(f.executor for f in fs)) == 1:
         loop = first(fs).executor.loop
     else:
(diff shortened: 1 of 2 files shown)


njsmith:
Looks good to me

Issue #351, opened 2016-06-21
dask-submit


hussainsultan:
Often times, the Client cannot be run on the same machine as the Scheduler; it may also not be possible to replicate the same environment as scheduler machine especially on EC2 e.g. EC2 machines may have access to S3 buckets while the executor machine may not. In the event, the client machine cannot replicate the environments, being able to ship a script to the client may be super helpful.

Akin to `spark-submit`, the Client lives with Scheduler node and can accept a python script over ssh/http etc.

hussainsultan:
looking at `distributed.deploy`, it appears that much of the ssh machinery is already there. I am happy to take this on if this makes sense.

mrocklin:
How would you design this?  Is this adding a service to the scheduler to create an on-demand client locally?  Is this running a client alongside the scheduler all the time that accepts scripts, runs them, and then returns results?

hussainsultan:
I think minimal step would be to run the client on on-demand basis on the same machine as scheduler when a user submits a job. Much of this code could live outside of scheduler and user will be in-charge of figuring out how outputs are handled e.g. output to s3 for persistence, etc. 

More sophisticated solution could be running an IPython kernel alongside scheduler that can run and return results in an interactive manner - this maybe more complex and i am not sure if we need this yet.

mrocklin:
OK.  Those both sound like good ideas to me.

One solution would be to subclass distributed.core.Server to create a
RemoteClient(Server) object.  This is the same common subclass of Scheduler
and Worker.  We would then send the script over TCP using the same
communication machinery that everything else uses.  Tracking the code in
Worker.upload_file might be useful.

hussainsultan:
ahh. interesting, that sounds pretty good. i will take this on in next couple of days.

hussainsultan:
Closing!

hussainsultan:
Sorry wrong issue. Reopening!

mrocklin:
This seems to be resolved

hussainsultan:
Thanks!

PR #360, opened 2016-06-28
Feature dask submit [#351]

hussainsultan:
This is work in progress and i am putting it here for general review and CI feedback.
### TODO

- [x] Add traceback to `dask-submit`
- [x] Add `dask-remote` cli
- [x] Make cli an executable 
- [x] Figure out how to test the CLI best

current status:

local cluster setup:

```
$ dask-scheduler
$ dask-worker 127.0.0.1:8786
$ dask-remote
```

```
$ dask-submit 127.0.0.1:8788 script.py 
b'11\n'
```
where 

script.py:
````python
from distributed import Executor

def inc(x):
    return x + 1

if __name__=='__main__':
    executor = Executor('10.34.13.61:8786')
    x = executor.submit(inc, 10)
    print(x.result())
```


distributed/cli/dask_submit.py
```diff
@@ -0,0 +1,59 @@
...
+        return self
+
+    @gen.coroutine
+    def _start(self):
+        self.remote_client = rpc(ip=self.ip, port=self.port)
+
+    def run(self, filepath=None):
+        remote_file = os.path.basename(filepath)
+        self.remote_client.upload_file(filename=remote_file, source_file_path=filepath)
+        return self.remote_client.execute(filename=remote_file)
```
mrocklin: I believe that this could all be collapsed down to a single coroutine.  Perhaps something like the following:

```python
@gen.coroutine
def submit(ip, port, filepath):
    rc = rpc(ip=ip, port=port)
    yield rc.upload_file(filename=remote_file, source_file_path=filepath)
    result = rc.execute(filename=remote_file)
    raise gen.Return(result)
```

distributed/remote.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def start(self,port=0):
+        self.loop.add_callback(self._start, port)
+        logger.info("Remote Client is running at {0}".format(port))
+
+
+    @gen.coroutine
+    def execute(self, stream=None, filename=None):
+        script_path = os.path.join(self.local_dir, filename)
+        cmd = 'python {0}'.format(script_path)
```
mrocklin: Might want to try `sys.executable` to get the same `python` from which this was run.

distributed/cli/dask_scheduler.py
```diff
@@ -52,7 +53,9 @@ def main(center, host, port, http_port, bokeh_port, show, _bokeh, bokeh_whitelis
     scheduler = Scheduler(center, ip=ip,
                           services={('http', http_port): HTTPScheduler},
                           loop=loop)
+    remote_client = RemoteClient(ip=ip)
     scheduler.start(port)
+    remote_client.start(port=remote_client_port)
```
mrocklin: Another option here is that we could treat `RemoteClient` as a service, similar to how `HTTPScheduler` is handled now.

mrocklin: Regardless these should possibly use the same ioloop

distributed/cli/dask_submit.py
hussainsultan: ahh..this  works!. Thanks!

hussainsultan:
@mrocklin Any ideas on whats causing Travis to fail - the tests run fine on all Python versions that i could try locally.

jcrist:
If you look at the travis failures (https://travis-ci.org/dask/distributed/jobs/140944830), it looks like it's due to the recent bokeh release. The recent release had some api changes that will need to be fixed.

hussainsultan:
@jcrist thanks! makes sense.

mrocklin:
Bokeh issues are being handled in https://github.com/dask/distributed/pull/356

hussainsultan:
Is it advisable to use specific version for packages[ here](https://github.com/dask/distributed/blob/master/.travis.yml#L51) to avoid this in future?

mrocklin:
That'd be reasonable.  If you want to pin the bokeh version for your PR to
get tests to pass that'd be fine with me.

mrocklin:
You should revert the `pin` commit now

hussainsultan:
@mrocklin I think i could use another review at this point. I believe most of the basic functionality that i am looking for is their. The big assumption is that all the dependencies of a submitted script are already present at scheduler machine, which is fine for now.

mrocklin:
Alrighty.  I'll probably have some time tonight or tomorrow morning.

distributed/cli/dask_submit.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@click.command()
+@click.argument('scheduler_address', type=str, required=True)
+@click.argument('filepath', type=str, required=True)
+def submit(scheduler_address, filepath):
+    @gen.coroutine
+    def f():
+        scheduler_host, remote_client_port = scheduler_address.split(':')
+        rc = rpc(ip=scheduler_host, port=remote_client_port)
```
mrocklin: You should be able to pass scheduler_address directly into `rpc` and avoid the logic above

distributed/cli/tests/test_dask_submit.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import print_function, division, absolute_import
+from distributed.utils_test import valid_python_script, popen, invalid_python_script
+
+def test_dask_submit_cli(valid_python_script, invalid_python_script):
+    with popen(['dask-scheduler', '--port', '8989']) as sched:
+            with popen(['dask-submit', '127.0.0.1:8788', str(valid_python_script)]) as submit:
```
mrocklin: Some whitespace alignment issues here.  Looks like you're slightly over-indented here.

distributed/remote.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def upload_file(self, stream, filename=None, source_file_path=None):
+        out_filename = os.path.join(self.local_dir, filename)
+        with open(source_file_path) as source_file:
+            data = source_file.read()
+        if isinstance(data, unicode):
+            data = data.encode()
+        with open(out_filename, 'wb') as f:
+            f.write(data)
+            f.flush()
```
mrocklin: Probably don't need to flush if you're using the contextmanager

mrocklin: Should we up the file, perhaps with `atexit`?

distributed/cli/dask_submit.py
```diff
@@ -0,0 +1,40 @@
...
+        remote_file = os.path.basename(filepath)
+        yield rc.upload_file(filename=remote_file, source_file_path=filepath)
```
mrocklin: Test that the status of this is `'OK'`?

distributed/remote.py
```diff
@@ -0,0 +1,53 @@
...
+        process = subprocess.Popen(cmd, shell=True,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE)
+        out, err = process.communicate()
+        raise gen.Return({'stdout': out, 'stderr': err})
+
+    def upload_file(self, stream, filename=None, source_file_path=None):
+        out_filename = os.path.join(self.local_dir, filename)
+        with open(source_file_path) as source_file:
+            data = source_file.read()
```
mrocklin: It appears that this function expects the file to already be on the local file system?  This confuses me.  I thought that the intention was for the remote client to accept the bytes of a file as one of the arguments and then write these bytes to a file.

hussainsultan: `RemoteClient.execute` is called after `upload_file` has been called already. `upload_file` sends the bytes and writes them to the worker local disk ( /tmp by default?) that is then executed by RemoteClient. Is this inline with your expectations?

mrocklin: What you describe is indeed in line with my expectations.  However looking at `upload_file` it looks like it's expecting a filename, rather than the bytes from that file.

hussainsultan: ahh yes. I was supposed to make it bytes instead of filename - sorry about that!

distributed/cli/tests/test_dask_submit.py
```diff
@@ -0,0 +1,23 @@
...
+                    break
+
+        with popen(['dask-submit', '127.0.0.1:8788', str(invalid_python_script)]) as submit:
+            for x in range(5):
+                line = submit.stdout.readline()
+                if b'Traceback' in line:
+                    break
+
+
+def test_submit_a_distributed_script(executor_contract_script):
```
hussainsultan: @mrocklin this test currently fails and i am not sure why. A trivial python script works fine but if i try to connect to scheduler via Executor, it never successfully establishes a connection. 

```
assert 'scheduler="127.0.0.1:8786" processes=0 cores=0' in [b'b\'Traceback (most recent call last):\\n  File "/var/folders/p4/6kv0q3sn02b5h9w1myf7l52rrhtbc2/T/client-mjpqip6q/di...ine 297, in wait\\n    gotit = waiter.acquire(True, timeout)\\nKeyboardInterrupt\\n\'\n', b'', b'', b'', b'', b'', ...]
```
any ideas on what might be going on?

hussainsultan: script
```python
executor = Executor('127.0.0.1:8786')      #hangs                               
print(executor)                         
```

mrocklin: Does it fail when run in isolation?  Is it possible that this port is occupied by a previous test?  Does sched have anything interesting in its stderr logs?

hussainsultan: Submitting a script on local machine without RemoteClient works. I am beginning to wonder if `subprocess.Popen` inside a coroutine is doing something interesting. The scheduler logs start showing stuff like:

```
bokeh.command.subcommands.serve - INFO - Starting Bokeh server on port 8787 with applications at paths ['/status', '/tasks']
distributed.core - INFO - Connection from 10.34.13.61:50499 to Scheduler
distributed.core - INFO - Connection from 10.34.13.61:50500 to Scheduler
distributed.core - INFO - Connection from 10.34.13.61:51052 to RemoteClient
bk_script_5585e068b2024763b0e64497d612d069 - WARNING - workers http route failed
bk_script_5585e068b2024763b0e64497d612d069 - WARNING - http route tasks failed

```

hussainsultan: quick update: I was able to make it this work if RemoteClient runs in its own `IOLoop`. The above issue is probably because of sharing the same `IOLoop` with Scheduler, Executor and RemoteClient. I tried doing this by running RemoteClient in isolation in a new IOLoop and things work as expected. I have the following questions:

* is there a way in tornado to run two separate IOLoops from dask-scheduler? I have been trying to do that but i dont think i have been successful.
* it also makes me wonder if `dask-remote` that kicks off the RemoteClient should be its own cli and people explicitly run side by side a scheduler

Let me know your thoughts! Thanks!

hussainsultan: cc: @jcrist

distributed/cli/dask_submit.py
hussainsultan: do you mean if `rpc` works or test submit coroutine? would `test_dask_submit_cli` not cover the case?

hussainsultan:
@mrocklin the pr is complete in terms of immediately desired functionality but there is a lot of commit noise. I will close it after a review and any discussion and open a new one with all the garbage removed. Thank you!

distributed/cli/dask_submit.py
```diff
@@ -0,0 +1,41 @@
...
+@click.command()
+@click.argument('scheduler_address', type=str, required=True)
+@click.argument('filepath', type=str, required=True)
+def submit(scheduler_address, filepath):
+    @gen.coroutine
+    def f():
+        scheduler_host, remote_client_port = scheduler_address.split(':')
+        rc = rpc(ip=scheduler_host, port=remote_client_port)
+        remote_file = os.path.basename(filepath)
+        bytes_read = open(filepath, "rb").read()
```
mrocklin: We should use a context manager here to close the file

```python
with open(filepath, 'rb') as f:
    bytes_read = f.read()
```

distributed/cli/dask_submit.py
```diff
@@ -0,0 +1,41 @@
...
+        yield rc.upload_file(filename=remote_file, bytes=bytes_read)
+        result = yield rc.execute(filename=remote_file)
+        if result['stdout']:
+            print(result['stdout'])
+        if result['stderr']:
+            print(result['stderr'])
```
mrocklin: Perhaps we should write these to the `sys.stdout` and `sys.stderr` of this process?

mrocklin: Instead of printing both to stdout that is

distributed/cli/tests/test_dask_submit.py
```diff
@@ -0,0 +1,20 @@
...
+                for x in range(5):
+                    line = submit.stdout.readline()
+                    if b'hello world!\\n' in line:
+                        break
+
+            with popen(['dask-submit', '127.0.0.1:8788', str(invalid_python_script)]) as submit:
+                for x in range(5):
+                    line = submit.stdout.readline()
+                    if b'Traceback' in line:
+                        break
```
mrocklin: If this doesn't occur then we still pass.  Perhaps `assert x < 4` ?

distributed/remote.py
```diff
@@ -0,0 +1,50 @@
...
+        process = subprocess.Popen(cmd, shell=True,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE)
+        out, err = process.communicate()
+        raise gen.Return({'stdout': out, 'stderr': err})
+
+    def upload_file(self, stream, filename=None, bytes=None):
+        out_filename = os.path.join(self.local_dir, filename)
+        if isinstance(bytes, unicode):
+            bytes = bytes.encode()
```
mrocklin: Perhaps we just force the payload to be `bytes`?

Also perhaps the name should be something like `file_contents` or `file_payload` rather than `bytes`

distributed/cli/dask_submit.py
```diff
@@ -0,0 +1,41 @@
...
+logger = logging.getLogger('distributed.dask_submit')
+
+
+@click.command()
+@click.argument('scheduler_address', type=str, required=True)
+@click.argument('filepath', type=str, required=True)
+def submit(scheduler_address, filepath):
+    @gen.coroutine
+    def f():
+        scheduler_host, remote_client_port = scheduler_address.split(':')
```
mrocklin: Is this the address of the scheduler or the address of the `dask-remote` process?

hussainsultan: this is dask-remote address. needs to be updated! thanks

distributed/remote.py
hussainsultan: i didnt like bytes name. I like `file_payload `

distributed/remote.py
```diff
@@ -0,0 +1,50 @@
...
+        handlers = {'upload_file': self.upload_file, 'execute': self.execute}
+        super(RemoteClient, self).__init__(handlers, io_loop=self.loop, **kwargs)
+
+    @gen.coroutine
+    def _start(self, port=0):
+        self.listen(port)
+
+    def start(self, port=0):
+        self.loop.add_callback(self._start, port)
+        logger.info("Remote Client is running at {0}".format(port))
```
mrocklin: We should probably include both the hostname/ip and the port like `ip:port`.  I often find that for interactive/local use it's very valuable to have a copy-pasteable address print to the screen.

distributed/cli/tests/test_dask_submit.py
hussainsultan: I am generally not happy how these tests are run. I tried a few ways to get `assert` statements here but i couldnt make travis work with them well. I will continue thinking about this. It seems like a general enough test type that it will help to have it in a nicer state.

distributed/remote.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @gen.coroutine
+    def execute(self, stream=None, filename=None):
+        script_path = os.path.join(self.local_dir, filename)
+        cmd = '{0} {1}'.format(sys.executable, script_path)
+        process = subprocess.Popen(cmd, shell=True,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE)
+        out, err = process.communicate()
+        raise gen.Return({'stdout': out, 'stderr': err})
```
mrocklin: Does popen give the exit code?  If so then we should probably communicate that as well.

distributed/cli/dask_submit.py
hussainsultan: good idea!

distributed/cli/dask_submit.py
hussainsultan: yes! Thanks

distributed/cli/tests/test_dask_remote.py
```diff
@@ -0,0 +1,7 @@
+from distributed.utils_test import popen
+
+
+def test_dask_remote():
+    with popen(['dask-remote', '--port', '9999']) as remote:
```
hussainsultan: @mrocklin any ideas on how this could be tested better? I was hoping their was a way to test if TCP communication is working.

mrocklin:
> @mrocklin the pr is complete in terms of immediately desired functionality but there is a lot of commit noise. I will close it after a review and any discussion and open a new one with all the garbage removed. Thank you!

We could also just squash it if you're comfortable with that.

mrocklin:
This could use a doc page

distributed/cli/tests/test_dask_remote.py
mrocklin: I would probably just skip this to be honest.  Or I would run the `RemoteClient` object in a normal, non-popen test using `@gen_cluster` and a scheduler.  I find the popen tests to be incredibly finicky.

distributed/cli/tests/test_dask_remote.py
```diff
@@ -0,0 +1,29 @@
+from mock import Mock
+
+from click.testing import CliRunner
+
+from distributed.cli.dask_remote import remote, _remote
```
mrocklin: You import a module that registers signal handlers.  This might cause problems?

mrocklin: I recommend pulling valuable functions out of the cli module.  We might want to keep that module as thin as possible.

mrocklin:
Another option I would be tempted by to identify the issue is to start a new PR in which you add code in slowly, module by module, and see which addition causes the havoc in TravisCI

hussainsultan:
@mrocklin This could use another review. Thank you!

mrocklin:
I'll try to take a look later tomorrow or early this weekend.  Sorry for
the delay.  Somewhat slammed at the moment.  I'd definitely like to get
this in soon though to make sure it doesn't go stale again.

hussainsultan:
Thank you. no rush at all.

distributed/submit/remote_cli.py
```diff
@@ -0,0 +1,34 @@
...
+    if ':' in host and port == 8788:
+        host, port = host.rsplit(':', 1)
+        port = int(port)
+    ip = socket.gethostbyname(host)
+    remote_client = client(ip=ip, loop=loop)
+    remote_client.start(port=port)
+    loop.start()
+    loop.close()
+    remote_client.stop()
+    logger.info("End remote client at %s:%d", host, port)
```
mrocklin: I apologize, my comments from before that it's nice to bring out a lot of logic from the CLI module weren't entirely clear.  

What I meant to say was that it's nice to have tests that don't rely on cli technology, like click and opening processes, but instead rely on the code that those cli modules use.  In the case of submit I suspect that what I'm suggesting is that we create tests based on `RemoteClient` and the `_submit` coroutine, not that we bring the click stuff out from the cli directory into other directories.  Click stuff should definitely stay in CLI.  Starting and stopping IOLoops should probably stay there as well.  Ideally we find a way to test most of the logic of our code without using these technologies; that way we can rely on the testing framework to set up and close down IOLoops and processes for us cleanly.  This should reduce testing frustration.

mrocklin: In an effort to avoid hierarchy I would recommend putting the remote and submit non-cli code into a single .py file in the base directory rather than having a `remote` subdirectory.

mrocklin: Thoughts on either point?

hussainsultan: Sorry i been away here. I think your points make sense. Regarding tests, i am pretty happy with https://github.com/dask/distributed/pull/360/files#diff-82ff7723602029eaf2609772606f30edR10. Is that what you meant?

mrocklin: I would have the `_submit` coroutine return values for stderr and stdout and have the CLI code emit those to sys.stderr/sys.stdout.  To me the CLI and sys.stderr/sys.stdout go together while coroutines and returning values go together.  I would keep the test but remove the capsys stuff and just check the return value of `_submit`.  

I have to admit though that I'm not as familiar with what's going on here as you are and I'm not paying a whole lot of attention to this.  I may be giving bad advice.

hussainsultan: That makes sense. I agree stdout writing should be at cli level. This should be easy to fix. Thanks!

hussainsultan:
closing in favor of #404

Issue #280, opened 2016-05-18
Is it possible to get into a task resubmission loop?


nfaggian:
I am finding sometimes it looks like I can get stuck waiting for a task that never finishes, I see logs like this:

~~~
distributed.worker - WARNING - Could not find data for ['aggregate-355f7592-270a-4b7a-acc4-afd53476e263']
~~~

which seems to be closely followed by a "re-execution" of a task. Is this a feature? worth noting that I see this when using a distributed scheduler in dask, not when it is local.

mrocklin:
Getting stuck in any way is definitely a bug.  

If it can't find data then possibly the worker with the data has died?  Are your workers all healthy?  I run into problems like this when some of my workers run out of distributed memory.

mrocklin:
I don't really expect this, but are you able to provide an example that I can replicate?

nfaggian:
I can share the basic pattern I am using, which is this:

~~~

from dask.delayed import delayed

@delayed
def load(json_file):
    # mutate json data 
   return json_dict

@delayed
def run(json_dict):
   # os.subprocess call to execute a simulator
   # mutate json dict 
   return json_dict

@delayed
def normalize(json_dict):
   # perform some data manipulation (data pointed to in dict)
   # mutate json dict 
   return json_dict

@delayed 
def aggregate(data):
     return data

workflow = aggregate([ normalize(run(load(x))) for x in json_file_list])

~~~

I guess this is a good opportunity to ask for advice too -- is the way I am using delayed a bit flawed. It seems really easy to break a large task if any one of the workflow steps breaks.

mrocklin:
Nothing here seems particularly scary.  I wouldn't recommend mutating inputs, especially if multiple functions rely on the same data, but that doesn't appear to be the case here.

Are your machines running out of RAM by any chance?

nfaggian:
The machines are not running out of ram. Was wondering if it was anything to do with how I am executing the simulators: 

~~~
def execute_command(cmd, prefix=None, log_output=True):
    """
    Executes a system call and logs output.
    """

    try:
        LOG.info("Executing: {0}".format(cmd))
        output = subprocess.check_output(
            cmd, shell=True, stderr=subprocess.STDOUT)
        if log_output:
            LOG.info("Output: \n {0}".format(
                "\n> ".join([""] + output.decode().split("\n"))))
    except subprocess.CalledProcessError as ex:

        # An exception occurred, capture log output.
        log_content = None
        if prefix is not None:
[...]

PR #361, opened 2016-07-05
Make multi-progress bars in notebooks indicate failure. 

nfaggian:
Picked up on the TODO that was listed in the multiprogress bar,  #280.

I suspect this could give a false impression of the entire computation not being completed, which is not true if you use the "delayed" method for forming a list of tasks to execute and only one fails.


mrocklin:
Thoughts on how we can add a test for this?

nfaggian:
I used a notebook but is it possible to setup a mock display loop somehow?

mrocklin:
While it's probably hard to test the display you *can* test the attributes of the widgets themselves.  Perhaps see `test_progressbar.py` to see if there are tests there that you could copy?

nfaggian:
Yep - I will have a go at doing exactly that. What are your thoughts on having a "warn" level, which simply states that some component of the calculation failed?

mrocklin:
Seems reasonable

nfaggian:
Apologies for slow progress here.. I will tackle this tomorrow.

nfaggian:
@mrocklin - possibly silly question but is just running py.test in the src directory supposed to work? or is it best to use travis?

mrocklin:
I usually run the following from the main repository directory (one up from the source directory)

    py.test distributed --verbose

If you want the cli tests to pass you'll also have to install the library ahead of time

nfaggian:
@mrocklin thanks -- trying to figure out how to create a "delayed" graph for the tests. Any examples?

mrocklin:
The bulk of the tests are in https://github.com/dask/distributed/blob/master/distributed/tests/test_executor.py .  Searching in that file will probably yield some nice examples.

mrocklin:
Closing as stale

PR #370, opened 2016-07-12
Add Executor.start_ipython

minrk:
- Starts an IPython kernel on worker(s).
- Doesn't work for local (ThreadPool) workers due to singleton checks in IPython.
-  Returns connection info for connecting Jupyter clients to the kernel.
- IPython runs in its own thread on the worker due to requiring the zmq poller to be in use by tornado.
- `e.start_ipython([workers], magic_names=['w1', 'w2'])` defines IPython magics for each specified worker that performs execution on the worker. Any output from the worker, including images and HTML output, will be properly redisplayed (widgets not included).
- `e.start_ipython([worker], qtconsole=True)` starts a QtConsole connected to the worker(s)

TODO:

- [x] what actions do we want the Executor to take? Return connected Client objects? Open qtconsole and/or jupyter console? Hook up local IPython execution to the worker?
- [x] what entrypoints should we have for starting IPython? Should there be methods on Futures to interact with the worker that ran them?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ install:
   # Install dependencies
   - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION
   - source activate test-environment
-  - conda install pytest coverage tornado toolz dill futures dask ipywidgets psutil bokeh requests joblib
+  - conda install pytest coverage tornado toolz dill futures dask ipywidgets psutil bokeh requests joblib mock ipykernel jupyter_client
   - pip install git+https://github.com/dask/dask.git --upgrade
   - pip install git+https://github.com/joblib/joblib.git --upgrade
   - pip install git+https://github.com/dask/s3fs.git --upgrade
--- a/distributed/_ipython_utils.py
+++ b/distributed/_ipython_utils.py
@@ -89,7 +89,7 @@ def register_worker_magic(connection_info, magic_name='worker'):
     ip.register_magic_function(remote, magic_kind='cell', magic_name=magic_name)
 
 
-def connect_qtconsole(connection_info, name=None):
+def connect_qtconsole(connection_info, name=None, extra_args=None):
     """Open a QtConsole connected to a worker who has the given future
     
     - identify worker with who_has
@@ -102,7 +102,10 @@ def connect_qtconsole(connection_info, name=None):
 
     path = os.path.join(runtime_dir, name + '.json')
     write_connection_file(path, **connection_info)
-    Popen(['jupyter', 'qtconsole', '--existing', path])
+    cmd = ['jupyter', 'qtconsole', '--existing', path]
+    if extra_args:
+        cmd.extend(extra_args)
+    Popen(cmd)
     def _cleanup_connection_file():
         """Cleanup our connection file when we exit."""
         try:
--- a/distributed/_ipython_utils.py
+++ b/distributed/_ipython_utils.py
@@ -1,0 +1,113 @@
+"""Utilities for integrating with IPython
+
+These functions should probably reside in Jupyter and IPython repositories,
+after which we can import them instead of having our own definitions.
+"""
+
+from __future__ import print_function
+
+import atexit
+import os
+try:
+    import queue
+except ImportError:
+    # Python 2
+    import Queue as queue
+from subprocess import Popen
+import sys
+from uuid import uuid4
+
+from tornado.gen import TimeoutError
+
+from IPython import get_ipython
+from jupyter_client import BlockingKernelClient, write_connection_file
+from jupyter_core.paths import jupyter_runtime_dir
+
+
+OUTPUT_TIMEOUT = 10
+
+
+def run_cell_remote(ip, kc, cell):
+    """Run a cell on a KernelClient
+
+    Any output from the cell will be redisplayed in the local session.
+    """
+    msg_id = kc.execute(cell)
+
+    in_kernel = getattr(ip, 'kernel', False)
+    if in_kernel:
+        socket = ip.display_pub.pub_socket
+        session = ip.display_pub.session
+        parent_header = ip.display_pub.parent_header
+
+    while True:
+        try:
+            msg = kc.get_iopub_msg(timeout=OUTPUT_TIMEOUT)
+        except queue.Empty:
+            raise TimeoutError("Timeout waiting for IPython output")
+
+        if msg['parent_header'].get('msg_id') != msg_id:
+            continue
+        msg_type = msg['header']['msg_type']
+        content = msg['content']
+        if msg_type == 'status':
+            if content['execution_state'] == 'idle':
+                # idle means output is done
+                break
+        elif msg_type == 'stream':
+            stream = getattr(sys, content['name'])
+            stream.write(content['text'])
+        elif msg_type in ('display_data', 'execute_result', 'error'):
+            if in_kernel:
+                session.send(socket, msg_type, content, parent=parent_header)
+            else:
+                if msg_type == 'error':
+                    print('\n'.join(content['traceback']), file=sys.stderr)
+                else:
+                    sys.stdout.write(content['data'].get('text/plain', ''))
+        else:
+            pass
+
+
+def register_worker_magic(connection_info, magic_name='worker'):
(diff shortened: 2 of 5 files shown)


mrocklin:
> what actions do we want the Executor to take?

I can imagine users wanting a variety of things.  Is there a convenient way to give them this choice?  For example can IPython Client objects be used to open qtconsoles or jupyter consoles?

> what entrypoints should we have for starting IPython?

First off I think would be an Executor method.  Strictly speaking Futures aren't tied to particular workers.  They might run on multiple workers or the data might be moved around.  Though of course selecting the recent worker of a future is likely to be a desired use case.

minrk:
> Is there a convenient way to give them this choice? For example can IPython Client objects be used to open qtconsoles or jupyter consoles?

The connection info returned is all we need to do any of these things, we don't need to instantiate a Client to do it. It's mainly a matter of what APIs we want to present. Do we want a single command to spawn a qtconsole connected to a worker, for instance? Jupyter doesn't, for instance, have a direct API for quickly "launch a qtconsole, given connection info," though we could add it (either here or there, it would be very brief).

For the API on the Executor, should it take workers directly, or take Futures, so users don't need to be responsible for figuring out the right worker given a Future? Or should it be both (explicit worker and figure it out from a Future)?

The most batteries-included magic API could be:

```python
e.qtconsole_for_first_failure(list_of_futures)
```

which could find the first Future in a list that failed, identify the Worker, start IPython, and open a local qtconsole connected to it.

The most basic version is what I have in this PR: given worker(s), start IPython and get the connection info back, leaving it to the user to use that connection info to do something.

We can really do anything and everything in between, so it's a matter of what do we think people will actually use, and how many options should we give them.

minrk:
Who knows what worker(s) were involved in a Future? How would I query the worker that failed to run a given task?

I suppose another question to ask is if we want to also send information back to the worker to be available in the IPython namespace, such as the original task, the key of the failing future, etc.

mrocklin:
The minimalist in me wants to start by just including documentation in the `e.start_ipython` method on how to initialize various jupyter solutions, seeing how these get used, and then add keyword arguments to `start_ipython` as use dictates.  My guess is that it's much easier to add functionality later than to maintain unused options.  The people starting ipython workers are probably the sort that are willing to read docstrings.  I don't have a good feel for how heavy this process is though.  Can you give a brief outline of the available options?

At the moment it looks like no one tracks where a task was run when it failed.  Historically it's hasn't been necessary to maintain correct scheduling.  The *pure* question that we might want to ask is "what workers *could* run this task given the data that they have now?"

minrk:
> I don't have a good feel for how heavy this process is though. Can you give a brief outline of the available options?

I'm writing up some examples now. There's more boilerplate than perhaps there should be, but we can address much of that with patches adding utility functions to Jupyter packages. The general options are:

- Instantiate a KernelClient to get a Python API to run code in the IPython instance on the worker.
- Spawn jupyter qtconsole attached to the kernel
- IPython magics to switch local IPython execution to run on the worker (extension of the KernelClient case). I've written basically this before, as IPython Parallel's `%autopx` and `%%px` magics.

The gist of starting a qtconsole with the info we have right now is:

1. write out connection info to a file
2. spawn `jupyter qtconsole --existing -f that-file.json`

With this PR exactly as-is, one could use this function to spawn a qtconsole locally, connected to the worker corresponding to a future:

```python
import subprocess
from jupyter_client import write_connection_file
from jupyter_core.paths import jupyter_runtime_dir

def qtconsole_for_future(future):
    e = future.executor
    worker = e.who_has(future)[future.key][0]
    infos = e.start_ipython(worker)

    runtime_dir = jupyter_runtime_dir()

    for worker, info in infos.items():
        path = os.path.join(runtime_dir, worker.replace(':', '-') + '.json')
        write_connection_file(path, **info)
        subprocess.Popen(['jupyter', 'qtconsole', '--existing', '-f', path])

futures = e.map(lambda x: x, range(5))
qtconsole_for_future(futures[0])
```

with the caveat that it only works when `who_has` finds a worker for the future, so it's only for tasks that *succeeded*.

>  The pure question that we might want to ask is "what workers could run this task given the data that they have now?"

I guess the real question is: what do you want to do with IPython for debugging dask? What information do you have at the client level, how do you pick a worker to interact with, and what do you want to send to the worker to do so?

nfaggian:
What about a "dask-cli" that lets you attach to the ipython instance in a terminal? I could imagine this would be really useful as a terminal based tool to monitor the cluster and attach to sessions? silly?

minrk:
This now defines two entrypoints:

1. `e.start_ipython([workers], magic_names=['w1', 'w2'])`
    defines IPython magics for each specified worker that performs execution on the worker. Any output from the worker, including images and HTML output, will be properly redisplayed (widgets not included).
2. `e.start_ipython([worker], qtconsole=True)`
   starts a QtConsole connected to the worker(s)

@nfaggian `jupyter console` could be that interface. This API returns the connection information needed, so similar to the `qtconsole` implementation here `dask-cli worker` would be:

```python
info_map = e.start_ipython(worker)
write_connection_file(connection_file_path, **info_map[worker])
start_jupyter_console(['--existing', connection_file_path)
```

minrk:
I've updated [the gist](https://nbviewer.jupyter.org/gist/minrk/6b75313e5fad6fb99fffefbdecd8e870) where I'm experimenting with these.

minrk:
I added `qtconsole_args` for passing through extra arguments to the qtconsole command. I used this in my lightning talk today to ssh tunnel through the scheduler to the worker.

mrocklin:
Thoughts on how to test this?

mrocklin:
### Future work

- [ ] Support `magic_names='workers-*'` or support suffixes
- [ ] Type-check magic_names
- [ ] Run an IPython kernel on the scheduler as a service alongside HTTP and Bokeh

### Questions

* How do we test?

* Is it possible to parametrize a magic based on data in local scope?  For example I can imagine the following workflow

```python
info = e.start_ipython(worker)[worker]

%remote info print("Hello")
```

This would avoid the issue of dynamically creating and managing magics but might be difficult on the IPython side?

minrk:
With IPython installed, I can test that the connection works and I can run code. I think I can minimally exercise the magics and qtconsole paths with mocks.

minrk:
I've added tests that exercise connecting to the kernel and sending code (this actually sends and receives messages), as well as two tests with mocks for registering magics and launching the qtconsole. Since I mock out IPython, it doesn't really test that we are calling IPython APIs correctly, but it should at least exercise the code path.

> Is it possible to parametrize a magic based on data in local scope?

It is possible to parameterize based on the IPython interactive namespace, but not in general the local scope (e.g. inside functions / modules). Based on this:

```python
info = e.start_ipython(worker)[worker]
%remote info print("Hello")
```

That would suggest `start_ipython(worker)` with no `magic_names` should register the parameterized IPython magic by default (if IPython is found to be running)? That should be quite doable. The magic would look something like:


```python
_clients = {}

def remote_magic(ip, line):
    info_name, line = line.split(None, 1)
    info = ip.user_ns[info_name]
    client = _clients.get(info, new_client_with(info)) # pseudo-code, but you get the gist
    run_cell(client, line)
```

mrocklin:
I think that having the `%remote my_connection_info` magic would be good.  However I'm also ecstatic with what we have here and I think we'll learn a lot more after some use.  I'm going to go ahead and merge now.  My apologies on the long delay.  This looks really cool to me.

minrk:
@mrocklin I had the `%remote` magic working in manual usage, but the tests weren't passing, so I'll open a PR for that as soon as I figure out what's funky in the tests. One question: when would you like the magic to be registered? There are a few ways:

1. on first call of `ex.start_ipython()`
2. explicitly, with e.g. `%load_ext dask.distributed`, or a function call (`%remote` is not tied to any executor) which might include any other IPython extensions we come up with.

mrocklin:
Both?  

In favor of a `%load_ext` solution I can imagine the `%remote` magic being generally useful to other projects outside of `dask.distributed` that are capable of producing connection info dicts.  

In favor of the on-the-first-call solution it's pretty convenient for how I can imagine it being used.

PR #376, opened 2016-07-14
[WIP] fix/bokeh_whitelist

quasiben:
PR fixes a small annoyance with whitelisting for distributed.  Previously, if one type

```
dask-scheduler --bokeh-port=9000 --bokeh-whitelist="*"
```

This would only proxy for host connections on `"*:9000`.  If a user is proxied (as often is the case inside docker containers) then it is likely they will be hitting some other host *usually* on port 80.  With this PR, a user can now type the following

```
dask-scheduler --bokeh-port=9000 --bokeh-whitelist="*" --prefix=/some_prefix
```

And expect all `host:port` combos to be available hosts and an optional prefix

### DIFF
--- a/distributed/bokeh/application.py
+++ b/distributed/bokeh/application.py
@@ -21,7 +21,8 @@ logger = logging.getLogger(__file__)
 
 class BokehWebInterface(object):
     def __init__(self, host='127.0.0.1', http_port=9786, tcp_port=8786,
-                 bokeh_port=8787, bokeh_whitelist=[], log_level='info', show=False):
+                 bokeh_port=8787, bokeh_whitelist=[], log_level='info',
+                 show=False, prefix=None, use_xheaders=False):
         self.port = bokeh_port
         ip = socket.gethostbyname(host)
 
@@ -29,6 +30,7 @@ class BokehWebInterface(object):
                  '127.0.0.1',
                  ip,
                  host]
+
         with ignoring(Exception):
             hosts.append(socket.gethostbyname(ip))
         with ignoring(Exception):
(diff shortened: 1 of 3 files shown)


quasiben:
Happy to fix tests but perhaps my solution is undesirable since now `--bokeh-whitelist` will need to include the `port` if defining a specific `host`

`--bokeh-whitelist='127.9.8.10:9000'`

mrocklin:
@quasiben what is the right thing to do here?

quasiben:
@mrocklin, I'd advocate for my PR -- we want to allow for users to control both IP and port

mrocklin:
Question for you.  Is there an input we can add to just always whitelist everything all the time and do away with all this?

quasiben:
Would you also want x-headers turned on as well ?

mrocklin:
I don't know what that means.  Let me rephrase.  Is there an appropriate nuclear option to get rid of this problem for good?  This kind of solution might not be appropriate given Bokeh's security guarantee's but may be appropriate given dask.distributed's weaker guarantees.

mrocklin:
OK, sounds like I should rebase and merge?  I'm mostly just trusting you to know if this should go in or not.

quasiben:
I don't think there is an appropriate nuclear option.  Behind proxies you will often still need to define a prefix.

quasiben:
I think this should go and feel free to ping me if this becomes an issue

mrocklin:
Rebased.  @quasiben can you take a look and verify that this is what you intended?

quasiben:
yup, looks good

PR #384, opened 2016-07-22
Move worker_info from defaultdict to dict

mrocklin:
Previously we had some errant entries in scheduler.worker_info.
Hopefully this change helps to isolate their origin.

cc @frol


mrocklin:
OK, I've constructed a new test that brings up and tears down workers pretty aggressively while trying to compute something large and have uncovered some race conditions.  Fixing.

PR #388, opened 2016-07-26
Avoid multiple heartbeats

mrocklin:
Previously when the network between worker and scheduler was clogged
(say due to a large data transfer) heartbeats would pile up waiting to
be sent.  These multiple heartbeats would expand the pool of open
connections, causing "Too many open files" errors.

Now we implement a simple semaphore to avoid sending a heartbeat when
one is already sent out.  This appears to fix the problem.

Fixes https://github.com/dask/dask/issues/1411


mrocklin:
cc @akcheung

PR #397, opened 2016-08-01
Protocol supports large bytes in nested objects

### LINKED ISSUE #393: Large bytestrings making their way into msgpack
Msgpack does not handle large bytestrings well.  Usually we evict large bytestrings from messages and send them along separately (see http://matthewrocklin.com/blog/work/2016/04/14/dask-distributed-optimizing-protocol).  

From https://github.com/dask/dask/issues/1411#issuecomment-235655955 it appears that this isn't happening during a `scatter` operation.  This may be because the large bytestrings aren't at the top-level of the message, but instead in a `data=` key.  

If this is the case then there are a few options:

1.  Dump scattered keys into the top-level keys of the message (and watch out for key conflicts)
2.  Teach the protocol to look more deeply into sub dictionaries (but be careful that this doesn't affect performance in the many-small message case)
3.  Special case a few keys, like `data` and `payload`

This might be a good learning issue.  Pinging @jcrist

jcrist:
Previously the distributed protocol would only shard large bytestrings
if they were values in the top level dictionary. This would cause
problems if there were large bytestrings in nested objects.

To overcome this, we now shard all large bytestrings if they meet the
following requirements:

- The top level of the message is a dictionary
- They are contained in any level of `dict`/`list` nesting. Other
  objects aren't traversed.

Fixes #393.

### DIFF
--- a/distributed/protocol.py
+++ b/distributed/protocol.py
@@ -27,17 +27,15 @@ outside of this module though and is not baked in.
 from __future__ import print_function, division, absolute_import
 
 import random
-import struct
 
 try:
     import pandas.msgpack as msgpack
 except ImportError:
     import msgpack
 
-from toolz import first, keymap, identity, merge
+from toolz import identity, get_in
 
 from .utils import ignoring
-from .compatibility import unicode
 
 
 compressions = {None: {'compress': identity,
@@ -64,35 +62,77 @@ with ignoring(ImportError):
     default_compression = 'lz4'
 
 
+def _extract_big_bytes(x, big, path=()):
+    if type(x) is dict:
+        for k, v in x.items():
+            if isinstance(v, (list, dict)):
+                _extract_big_bytes(v, big, path + (k,))
+            elif type(v) is bytes and len(v) >= 2**31:
+                big[path + (k,)] = v
+                del x[k]
+    elif type(x) is list:
+        for k, v in enumerate(x):
+            if isinstance(v, (list, dict)):
+                _extract_big_bytes(v, big, path + (k,))
+            elif type(v) is bytes and len(v) >= 2**31:
+                big[path + (k,)] = v
+                x[k] = None
+
+
 def dumps(msg):
     """ Transform Python value to bytestream suitable for communication """
-    small_header = {}
-
+    big = {}
+    # Only dicts can contain big values
     if isinstance(msg, dict):
-        big = {k: v for k, v in msg.items()
-                    if isinstance(v, bytes) and len(v) > 1e6}
-    else:
-        big = False
-    if big:
-        small = {k: v for k, v in msg.items() if k not in big}
-    else:
-        small = msg
+        _extract_big_bytes(msg, big)
+    small_header, small_payload = dumps_msgpack(msg)
+    if not big:
+        return small_header, small_payload
 
-    frames = dumps_msgpack(small)
-    if big:
-        frames += dumps_big_byte_dict(big)
+    # Shard the big segments
+    shards = []
+    res = {}
+    for k, v in list(big.items()):
+        L = []
+        for i, j in enumerate(range(0, len(v), 2**30)):
+            key = '.shard-%d-%s' % (i, k)
+            res[key] = v[j: j + 2**30]
+            L.append(key)
+        shards.append((k, L))
 
-    return frames
+    keys, values = zip(*res.items())
+
+    compression = []
+    values2 = []
+    for v in values:
+        fmt, vv = maybe_compress(v)
+        compression.append(fmt)
+        values2.append(vv)
+
+    header = {'encoding': 'big-byte-dict',
+              'keys': keys,
+              'compression': compression,
+              'shards': shards}
+
+    return [small_header, small_payload,
+            msgpack.dumps(header, use_bin_type=True)] + values2
 
 
 def loads(frames):
     """ Transform bytestream back into Python value """
-    header, payload, frames = frames[0], frames[1], frames[2:]
-    msg = loads_msgpack(header, payload)
+    small_header, small_payload, frames = frames[0], frames[1], frames[2:]
+    msg = loads_msgpack(small_header, small_payload)
 
     if frames:
-        big = loads_big_byte_dict(*frames)
-        msg.update(big)
+        header = msgpack.loads(frames[0], encoding='utf8')
+
+        values2 = [compressions[c]['decompress'](v)
+                   for c, v in zip(header['compression'], frames[1:])]
+        lk = dict(zip(header['keys'], values2))
+
+        for k, keys in header['shards']:
+            v = b''.join(lk.pop(kk) for kk in keys)
+            get_in(k[:-1], msg)[k[-1]] = v
 
     return msg
 
@@ -109,7 +149,7 @@ def byte_sample(b, size, n):
 
 
 def maybe_compress(payload, compression=default_compression, min_size=1e4,
-        sample_size=1e4, nsamples=5):
+                   sample_size=1e4, nsamples=5):
     """ Maybe compress payload
 
     1.  We don't compress small messages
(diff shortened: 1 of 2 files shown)


jcrist:
Note that `dumps` now is destructive to large messages. This could be overcome easily with a `deepcopy` of the original message or less easily by explicitly copying the message in the traversal. However, I don't think this should be a problem internally (could be wrong though).

distributed/protocol.py
```diff
@@ -64,35 +62,77 @@
...
+            elif type(v) is bytes and len(v) >= 2**31:
+                big[path + (k,)] = v
+                del x[k]
+    elif type(x) is list:
+        for k, v in enumerate(x):
+            if isinstance(v, (list, dict)):
+                _extract_big_bytes(v, big, path + (k,))
+            elif type(v) is bytes and len(v) >= 2**31:
+                big[path + (k,)] = v
+                x[k] = None
```
mrocklin: Do you have any thoughts on how to make this pure and pass through the original object in the common case?

distributed/protocol.py
```diff
@@ -64,35 +62,77 @@
...
+
+
 def dumps(msg):
```
mrocklin: I recommend that `dumps` now gets a keyword argument `big=False` that tells it when to look for big bytestrings.

jcrist: In what places would you set it to True?

mrocklin: Hrm, that's a good point, this is fairly far buried within other calls, like rpc.  

In practice the channels that receive large bytestrings are scatter, gather, worker-worker communication and occasionally update-graph.  But this is probably not something you should dive into at the moment.

Alternative question: how expensive is this change?  What is the overhead on dumping small messages?

jcrist: Using these messages as examples

```python
from distributed.protocol import dumps, loads

# Unrealistic and intentionally bad
msg = {'a': list('abcdefghijklmnopqrstuvwxyz'),
       'b': dict(zip(range(1000), range(1000))),
       'c': 'foo',
       'd': 'bar',
       'e': [{'a': 1} for i in range(100)]}

# More realistic
msg2 = {'a': 'foo',
        'b': 'bar',
        'c': 1.5,
        'd': ['foo', 'bar']}

s = dumps(msg)
s2 = dumps(msg2)
```
This PR:

```python
In [1]: %timeit dumps(msg)
1000 loops, best of 3: 590 µs per loop

In [2]: %timeit loads(s)
10000 loops, best of 3: 49.9 µs per loop

In [3]: %timeit dumps(msg2)
The slowest run took 6.03 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 6.48 µs per loop

In [4]: %timeit loads(s2)
The slowest run took 9.48 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.16 µs per loop
```

Master:

```python
In [1]: %timeit dumps(msg)
10000 loops, best of 3: 50.2 µs per loop

In [2]: %timeit loads(s)
10000 loops, best of 3: 49.4 µs per loop

In [3]: %timeit dumps(msg2)
The slowest run took 7.68 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 4.04 µs per loop

In [4]: %timeit loads(s2)
The slowest run took 12.41 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.06 µs per loop
```

In summary:
- No overhead on loads for realistic or unrealistic cases
- Dumps is slower. Depending on the depth/breadth of the tree this overhead may be small or not.

I haven't spent any time optimizing this though, I can probably find some small speedups at least.

mrocklin: These times are probably somewhat non-trivial.  Current peak performance is around 5000 tasks per second, which gives us a 200 us budget per task. 

Here is a benchmarking script that can be run with snakeviz

https://gist.github.com/5b2b6dcc92d760003f1d9b393f8d4b13

```
python -m cProfile -o prof.out distributed-benchmark.py
snakeviz prof.out
```

jcrist: I find no noticeable change between this PR and master when running that benchmark.

Trying all my usual python micro-optimization tricks, I can't get a noticeable speedup on the intentionally bad case. I think the only way to speed this up (if needed) would be to do less work. One option would be to set a limit on the depth at which we traverse, another would be to add a boolean (as you suggested above) and use full traversal only where necessary. We could also only traverse certain keys.

mrocklin: These benchmarks may also be of interest: https://gist.github.com/67fbb7c77288c110cd1018a01526044e

mrocklin: OK, maybe we accept this running all the time for the near future and optimize when it becomes a bottleneck.

jcrist: Timings from the array benchmark (normalized for same number of calls, since they slightly differ between runs).

```
This PR:
- dumps cumulative time: 0.6042 s
- loads cumulative time: 0.5028 s

Master:
- dumps cumulative time: 0.5738 s
- loads cumulative time: 0.3064 s
```

jcrist: I think you have a better intuition here about the cost of this change. I'd trust your judgement. I'm skeptical that we can improve performance here without reducing the amount of work done, but that can be left for later until it's necessary.

distributed/protocol.py
```diff
@@ -64,35 +63,89 @@
...
+            if isinstance(v, (list, dict)):
+                _extract_big_bytes(v, big, path + (k,))
+            elif type(v) is bytes and len(v) >= 2**31:
+                big[path + (k,)] = v
+    elif type(x) is list:
+        for k, v in enumerate(x):
+            if isinstance(v, (list, dict)):
+                _extract_big_bytes(v, big, path + (k,))
+            elif type(v) is bytes and len(v) >= 2**31:
+                big[path + (k,)] = v
```
mrocklin: I wonder if there is a way to do this without deepcopy in the common case.  Is it possible to walk through the data structure and on-the-fly either build a new thing or let the current thing pass through based on whether or not changes are necessary.

I wouldn't put too much thought into this, just a question.

jcrist: We don't deepcopy in the default case (see the function above this).

mrocklin: Ah!  just so I get this right you mutate, and then you check to see if you mutated and if so you put the stuff back in?

jcrist: Not quite. We crawl through `msg` and find all the big strings and paths - these are then put into `big`. After that, if `big` isn't empty, we `deepcopy` the original `msg`, and then remove all the big strings from the copy. Cheap in the common case when `big` is empty, but more expensive than just mutating `msg` originally.

mrocklin: Great.  When big is non-empty we'll have much larger costs to hide behind.

distributed/protocol.py
```diff
@@ -64,35 +63,89 @@
...
+                t[path[-1]] = None
+    return x, big
+
+
+def _extract_big_bytes(x, big, path=()):
+    if type(x) is dict:
+        for k, v in x.items():
+            if isinstance(v, (list, dict)):
+                _extract_big_bytes(v, big, path + (k,))
+            elif type(v) is bytes and len(v) >= 2**31:
```
mrocklin: We want to extract bytestrings that are even moderately large.  Probably anything larger than 1MB.  It would be nice to add tests to ensure that this was the case.

In particular, the following test might be useful:

```python
data = np.random.randint(0, 255, dtype='u1', size=(2000000)).data.tobytes()
msg = {'x': [{'y': data}]}
frames = dumps(msg)
assert any(data == frame for frame in frames)
```

jcrist: This isn't currently the case, but it could be. There are two numbers that are important here:
- The size to pull a bytestring out of the "small" header
- The size to shard each bytestring

Currently these are set to `2**31` and `2**30` respectively. What do you suggest for these numbers instead?

mrocklin: Off the cuff I'll suggest `2**20` and `2**28`

jcrist: Fixed.

mrocklin:
This is in.  Thanks @jcrist

PR #398, opened 2016-08-02
Refactor Scheduler

mrocklin:
This is a large refactor of the distributed scheduler.

### Problem

* Recently we've found deadlocks in the scheduler.  These have become increasingly time consuming to identify and resolve.
* No one feels comfortable operating on the scheduler code except for myself

Historically as we've added more and more features we convolute the scheduler code more and more.  This has led to a difficult and time consuming situation where resilience and performance are hard to guarantee, and hard to fix once an issue has been identified.

The previous architecture was a set of scheduler methods that called each other in a difficult-to-reason-about way that were triggered by a few events.

### Solution

This PR adds a bit of top-down design to the management of scheduler state.  This has a few positive effects:

1.  The resulting code is cleaner and easier to follow
2.  The design of the scheduler is easier to talk about.  What were previously conventions and ways of thinking about what's happening have become actual function calls
3.  Much of the state changes happen in a central place.  This enables consistent logging, and runtime checking, which greatly accelerate the debugging and development process.

### Transitions and Stimuli

To learn more about the new design you should read the doc page that comes along with this PR

* Docs: https://github.com/mrocklin/distributed/blob/scheduler-refactor/docs/source/scheduler.rst
* Code: https://github.com/mrocklin/distributed/blob/scheduler-refactor/distributed/scheduler.py


mrocklin:
Scheduler doc page added.  Feedback on the transitions/stimuli design appreciated.

mrocklin:
This is ready for review.  I recommend the following:

1.  Read the doc page
2.  Read through the full scheduler.py, not the diff

mrocklin:
@minrk @ogrisel @frol this may interest you as well

frol:
@mrocklin This looks neat! I will give it a try tomorrow!

jcrist:
Reading through the scheduler code refactor, find myself a bit confused by this section: https://github.com/mrocklin/distributed/blob/03fd3eec3debc8c21472222fa56122b357046fa2/distributed/scheduler.py#L2056-L2064

From what I understand, if there isn't a single function implementing the desired transition, you first transition to 'released', then go from released to finish. I assume this is just a special case due to some internals, and not a general solution (e.g. there's no time where you'd want to transition two states when not passing through 'released')?

Also, is this copy needed? https://github.com/mrocklin/distributed/blob/03fd3eec3debc8c21472222fa56122b357046fa2/distributed/scheduler.py#L2061

(I'd do these as line comments, but you can't on a PR this big it seems).

mrocklin:
Correct about `released`.  We're in an awkward `n^2` situation where we may find ourselves writing transition functions between every pair of states.  Alternatively we could have some sort of graph traversal algorithm.  At the moment we have a traversal algorithm that goes through a commonly accessible state, `'released'`. 

In the future I think that we should make all edges explicit, even if the many transition functions themselves just look like the following:

```python
def transition_foo_bar(self, key):
    a = self.transition(key, 'intermediate')
    b = self.transition(key, 'bar')
    return merge(a, b)
```

mrocklin:
`a.copy()` exists because we've placed `a` into the `transition_log` here: https://github.com/mrocklin/distributed/blob/03fd3eec3debc8c21472222fa56122b357046fa2/distributed/scheduler.py#L2069 (when it was called `recommendations` in the previous call to `transition`)

jcrist:
Just to clarify, the call to `logger.info` may not have written `recommendations` in the previous call yet, so we don't want to mutate it?

mrocklin:
Ah, I pointed to the wrong line: https://github.com/mrocklin/distributed/blob/03fd3eec3debc8c21472222fa56122b357046fa2/distributed/scheduler.py#L2066

mrocklin:
@jcrist large scale feedback is also appreciated.  Especially feedback of the form "I didn't really understand X"

mrocklin:
Barring further comments I plan to merge this tomorrow morning (roughly 12 hours from now)

PR #404, opened 2016-08-04
Feature dask submit [#351, supersedes #360]

### LINKED ISSUE #351: dask-submit

necaris:
Supersedes https://github.com/dask/distributed/pull/360 from @hussainsultan , also closes #351 

Submitting a PR directly for simplicity


docs/source/submitting-applications.rst
```diff
@@ -0,0 +1,51 @@
...
+`script.py`:
+
+.. code-block:: python
+   from distributed import Executor
+
+   def inc(x):
+        return x + 1
+
+   if __name__=='__main__':
+        executor = Executor('10.210.220.181:8786')
```
mrocklin: We should probably change this to `'127.0.0.1:8786'` to make it more general and because it's likely to target the same machine

distributed/submit.py
```diff
@@ -0,0 +1,85 @@
...
+    if ':' in host and port == 8788:
+        host, port = host.rsplit(':', 1)
+        port = int(port)
+    ip = socket.gethostbyname(host)
+    remote_client = client(ip=ip, loop=loop)
+    remote_client.start(port=port)
+    loop.start()
+    loop.close()
+    remote_client.stop()
+    logger.info("End remote client at %s:%d", host, port)
```
mrocklin: I'm a bit curious about how to put this on the RemoteClient object itself or in the cli script.  For now though I'd just like to get this merged so I say we leave this for future work.

distributed/cli/dask_worker.py
```diff
@@ -16,18 +16,22 @@
...
         exit(1)
 
-signal.signal(signal.SIGINT, handle_signal)
-signal.signal(signal.SIGTERM, handle_signal)
+
+def install_signal_handlers():
+    # NOTE: We can't use the generic install_signal_handlers() function from
+    # distributed.cli.utils because we're handling the signal differently.
+    signal.signal(signal.SIGINT, handle_signal)
+    signal.signal(signal.SIGTERM, handle_signal)
```
mrocklin: Should we just inline these directly rather than wrap them in a function?

necaris: Sure, I'd be happy to define the handler function inline if that's clearer -- I don't think there's much in it, though.

mrocklin: It just avoids an indirection, so where we currently have `install_signal_handlers` we call these two lines instead.

necaris: Addressed in acc4c2b :-)

mrocklin:
I've squashed this into two commits (one for each author) and merged.  Closing.

Thanks @hussainsultan and @necaris !  This is in.

PR #408, opened 2016-08-08
Raise exception in Executor when send before start

### LINKED ISSUE #401: Executor fails when not started
When starting the executor asynchronously, it can happen that tasks are submitted before it has been started, resulting in:

```
Exception raised while submitting task:
'Executor' object has no attribute 'scheduler_stream'
```

It would be good to have a way of querying the executor status, to see if it is currently accepting tasks or not.

mrocklin:
This solidifies the `.status` attribute of the Executor to have the
three values of ``None, 'running', 'closed'``.  It also raises
exceptions if the user tries to send messages to a closed scheduler.

Fixes #401

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -302,6 +302,9 @@ class Executor(object):
         self.status = 'running'
 
     def _send_to_scheduler(self, msg):
+        if self.status is not 'running':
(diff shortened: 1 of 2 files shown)


mrocklin:
cc @stefanv

stefanv:
This is very helpful, thank you.

PR #409, opened 2016-08-08
Start Diagnostics server by default

### LINKED ISSUE #402: ENH: start diagnostics when launching Executor from client code
effectively do this:

provide an option in the ``Executor`` (prob should default to True?) to auto-start diagnostics
```
e = Executor()
e.cluster.start_diagnostics_server()
```

mrocklin:
Previously when calling `Executor()` or `LocalCluster()` we wouldn't
start up a bokeh server by default, instead asking users to find
`e.cluster.start_diagnostics_server()`.  Now we start this by default.

Additionally, we used to pass information via a `.dask-web-ui.json` file
which was written to the local directory, but searched for in the
install directory.  Now we write this to a `~/.dask/` directory instead
for global access.

Finally we turn off bokeh logging to clean up the interactive experience
and remove some unused bokeh plots (tables) that were using deprecated
functionality and causing warnings to pop up.

Fixes #402

### DIFF
--- a/distributed/bokeh/application.py
+++ b/distributed/bokeh/application.py
@@ -19,10 +19,15 @@ binname = os.path.join(os.path.dirname(sys.argv[0]), binname)
 
 logger = logging.getLogger(__file__)
 
+dask_dir = os.path.join(os.path.expanduser('~'), '.dask')
+if not os.path.exists(dask_dir):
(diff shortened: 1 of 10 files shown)


mrocklin:
cc @jreback

PR #411, opened 2016-08-09
PR: Implement writing to hdfs3

martindurant:
(following https://github.com/dask/dask/pull/1346 )

Depends on https://github.com/dask/hdfs3/pull/89


martindurant:
(will fail until the hdfs3 PR is merged)

distributed/tests/test_hdfs.py
```diff
@@ -375,3 +377,35 @@ def test_deterministic_key_names(e, s, a, b):
...
+        out = write_bytes(values, path, hdfs=hdfs)
+        futures = e.compute(out)
+        results = yield e._gather(futures)
+        assert len(hdfs.ls('/tmp/test/')) == 5
+
+        sample, vals = read_bytes('hdfs:///tmp/test/*.part',
+                                    hdfs=hdfs, lazy=True)
+        futures = e.compute(vals)
+        results = yield e._gather(futures)
+        assert set(values) == set(results)
```
mrocklin: Can we drop the call to `set` here?  It looks like these operations might preserve order

mrocklin:
Why does this require serialization again?  In the common case shouldn't the file and the bytes-writing occur on the same machine?

martindurant:
Yes, they ought to occur on the same machine (and process), and the scheduler will tend to make this the case - but it's not a hard necessity. There are two delayed steps to the writing, one to open the file and one to write to it. I cannot conceive of how to make it one stage and still support streamed writing.

mrocklin:
Perhaps we have an alternative route that does not support streamed writing.  A route that takes a path and a bytestring and writes those bytes to that path.

martindurant:
You mean write_bytes and stream_bytes, and the user has to choose? 

How about at https://github.com/dask/dask/blob/master/dask/bytes/core.py#L144 instead of opening the file, we make a delayed opener, or pass the function+args to `write_block_to_file` instead of the file object, would that work?

mrocklin:
Something like that double approach, yes.  Users working at this low level will, I expect, be able to handle the choice of being given a file object or sending in a bytestring

These two interfaces seem intuitive to me:

    * `path -> writable file`
    * `path, bytes -> None`

There are probably other nice interfaces though.  I don't know if these would be more or less clear than a delayed opener.

martindurant:
OK, I can do it by passing down the opener function, which in some ways is simpler. That means a change over in dask, so that we do the same for all backends. The `open_files_write`, still using delayed, is then bypassed by available to anyone who wants it, and streamed writing is still supported.

mrocklin:
Just to make sure I understand, you're now registering how to open a single file for each backend?  Then the streaming-files or write-bytes solutions can both independently reference this collection?

martindurant:
Yes, just the opener, so that it can be passed to the writing function in dask which, as before, can handle bytes _or_ streaming. Should I put up that PR to demonstrate?

mrocklin:
It seems to make sense to me.  Obviously a PR to dask at some point with that idea would be good :)

martindurant:
https://github.com/dask/dask/pull/1453/

distributed/hdfs.py
```diff
@@ -45,6 +45,24 @@ def read_block_from_hdfs(filename, offset, length, host=None, port=None,
     return bytes
 
 
+def open_file_write(paths, hdfs=None, **kwargs):
+    """ Open list of files using delayed """
+    if hdfs is None:
+        from hdfs3 import HDFileSystem
+        hdfs = HDFileSystem(kwargs.get('host'), kwargs.get('port'))
+    out = [delayed(hdfs.open)(path, 'wb') for path in paths]
+    return out
```
mrocklin: We can probably drop this

mrocklin:
The failures we're seeing in the test suite are unrelated.  I'm slowly working through them.

distributed/hdfs.py
martindurant: I thought worth keeping around for some unknown purpose (I think the local version remains) - but of course untested and unused as it is.

martindurant:
Only failure: `distributed/diagnostics/tests/test_progress.py::test_many_Progress` on py27 (hdfs=True)

mrocklin:
Probably not an issue.  Looking into it.  I've been fixing bugs like this that have crept into master.  Fix is already in https://github.com/dask/distributed/pull/415 .  Safe to restart the test.  Doing so now.

PR #416, opened 2016-08-11
Joblib backend improvements

### LINKED ISSUE #414: Distributed.joblib code fails to affect sklearn
The comments in [this stackoverflow answer](http://stackoverflow.com/questions/38601026/easy-way-to-use-parallel-options-of-scikit-learn-functions-on-hpc/38814491#38814491) raise concerns about the effectiveness of `distributed.joblib` to parallelize vanilla sklearn code.  It appears that sklearn ships with its own version of Joblib, which the plugin registration in `distributed.joblib` does not affect.

It would be good to test sklearn functionality and, if necessary, add plugin registration to `sklearn.externals.joblib` in the same way we do to normal `joblib`.

jcrist:
A user may have joblib installed on its own, or bundled with
scikit-learn. Further, the joblib backend only works with joblib >=
0.10.0. This PR implements logic to handle both of these cases.

- Register with `joblib` *and/or* `sklearn.externals.joblib` if
  installed and meets the minimum version.
- Throw a nice error if neither installed or up to date.
- Parametrize joblib tests to test from both joblib and sklearn.

Updated the docs to reflect this as well.

Fixes #414.

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -1,10 +1,36 @@
 from __future__ import print_function, division, absolute_import
 
-from joblib._parallel_backends import ParallelBackendBase, AutoBatchingMixin
-from joblib.parallel import register_parallel_backend
+from distutils.version import LooseVersion
+
 from tornado import gen
 
 from .executor import Executor, _wait
+from .utils import ignoring
+
+
+# A user could have installed joblib, sklearn, both, or neither. Further, only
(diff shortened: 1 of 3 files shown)


mrocklin:
cc @ogrisel 

Also @minrk this same issue may affect `ipyparalell`

minrk:
Thanks, I'll see if I should do something similar.

PR #417, opened 2016-08-12
Don't register hdfs backend if hdfs3 not installed

jcrist:
Previously importing `distributed.hdfs` would cause `dask.core.bytes` to have `hdfs` registered *even if* `hdfs3` wasn't installed. This was causing problems with the function that checks if the backend is available, leading to hard to diagnose bugs. This was discovered in https://github.com/dask/dask/pull/1464, where the test failed only if the doctests were run *and* the whole test suite was run (ordering dependent). A hack to deal with the issue was added in https://github.com/dask/dask/pull/1464/commits/a353995fbe39822876e8cc35ac800180ffeac83b, but it's not ideal.

This fixes the issue, only registering the hdfs backend if hdfs3 is installed.

### DIFF
--- a/distributed/hdfs.py
+++ b/distributed/hdfs.py
@@ -2,20 +2,16 @@
 from __future__ import print_function, division, absolute_import
 
 import logging
-import json
 from math import log
 import os
-import io
 from warnings import warn
 
-from dask.delayed import delayed
-from dask.base import tokenize
 import dask.bytes.core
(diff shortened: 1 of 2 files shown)


mrocklin:
So this would cause doctests to fail, which I'm surprised to learn we don't do at all in travis.ci.  

Merging.  I may end up replacing this with an explicit check around `if hdfs3` in the moderate future.

PR #421, opened 2016-08-16
Avoid dropped message

mrocklin:
This was reported by a private user.  There are some cases where a message appears to be dropped when gathering data between workers.

The requester sends off hundreds of requests to various other workers, all but one comes back, and the worker waits forever for the missing message.  Looking at traces it looks like the requester definitely sent the message and the network on the other side definitely received it, however the worker itself never actually called `read` on the message to respond to it correctly.

Replacing the rpc mechanism with straight IOStreams seems to remove the issue.  It's unclear why however; this may point to a deeper issue within our communication system.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -116,7 +116,8 @@ def gather_from_workers(who_has, deserialize=True, rpc=rpc, close=True,
             else:
                 raise KeyError(*bad_keys)
 
-        coroutines = [rpc(address).get_data(keys=keys, close=close)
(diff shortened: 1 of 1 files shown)


mrocklin:
This was happening fairly regularly with something like 700 active workers and, I think, gathering from them all

mrocklin:
I believe that this was independently resolved by either of the following PRs

*  https://github.com/dask/distributed/pull/476
*  https://github.com/dask/distributed/pull/462

Issue #128, opened 2016-02-11
Resource Manager Interface


mrocklin:
Resource Manager Interface
==========================

*The first step is hard, but standing up is way harder.*

Deploying distributed on a remote cluster is a large hurdle between the
curious user and serious adoption.  Efforts to improve deployment can
significantly improve community uptake and value.

Unfortunately, there are a large number of deployment solutions so there is no
single piece of code we can write to satisfy most potential users.  We have
direct relations with people who employ the following deployment technologies:

1.  YARN
2.  Mesos
3.  SLURM
4.  Torque
5.  SGE
6.  SSH
7.  Local processes

Supporting each of these is doable in isolation, but when we consider
supporting all in aggregate we desire a common framework.

Additionally, near-future versions of the distributed scheduler may want to
dynamically request and release worker resources depending on load.  Providing
a single resource management interface to the scheduler would allow integration
with several technologies in a sane and maintainable manner.

**Question:** *What does such a resource management interface look like?*

Example Interface
-----------------

Let us consider the following signals from scheduler to resource manager (RM)
and back.

*  Scheduler -> RM: Please give me more workers
*  Scheduler -> RM: I no longer need these workers
*  RM -> Scheduler: I plan to take back this worker very soon
*  RM -> Scheduler: This worker has unexpectedly died

We could implement this as a Python ResourceManger object that has fuctions for
"Please give me more workers" and "I no longer need these workers" that the
scheduler can call as well as callbacks that the scheduler provides on
ResourceManager creation for "I plan to take back this worker soon" and "this
worker has unexpectedly died".

```python
class ResourceManager(object):
    def __init__(self, on_reclaiming_worker, on_worker_died, **kwargs):
        ...

    def request_workers(self, nworkers, **kwargs):
        ...

    def release_workers(self, worker_ids):
        ...
```

We might then subclass this interface for processes, SSH, Mesos, Yarn, etc..

Questions
---------

Different resource management systems employ different information.  Yarn
allows the specification of CPUs and Memory and the use of containers, HPC job
schedulers have fixed job durations, Mesos operates a bit differently from all.

The example operations provided above may not fit all resource managers.  Some
may provide only a subset of this functionality while some may have a superset.
How do we balance the needs to specialize to a particular system while also
maintaining compatibility with many systems and keeping code simple.

cc @hussainsultan @broxtronix @quasiben @danielfrg

mrocklin:
[...]

PR #428, opened 2016-08-17
[WIP] Cluster scheduling support for joblib 

mwojcikowski:
Make joblib "auto-magicaly" schedule dask-workers on queue system. This prototype works on our Slurm installation. Some comments/input might be useful. Does anyone, but me, need such functionality?

Finally, this is just a prototype, final version should have:

- [ ]  scheduling software abstraction layer - move it out as separate class (see next point)
- [ ]  add support for other schedulers PBS/Torque/SGE/UGE etc. or just use [clusterlib](https://github.com/clusterlib/clusterlib)
- [ ]  document additional arguments, general cleanup

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -1,5 +1,9 @@
 from __future__ import print_function, division, absolute_import
 
+from subprocess import check_output
+from time import sleep
+from tempfile import NamedTemporaryFile
+
 from joblib._parallel_backends import ParallelBackendBase, AutoBatchingMixin
 from joblib.parallel import register_parallel_backend
 from tornado import gen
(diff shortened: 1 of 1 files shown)


mrocklin:
Hi.  I'm glad to see interest in using Dask on SLURM and other job schedulers.  This kind of functionality is probably useful beyond just Joblib, so it probably shouldn't be so tightly coupled to this code.  Eventually it would be nice to have a few of these living in the `distributed/deploy/` submodule.  

Could I ask you to take a look at `distributed/deploy/local.py` and share your thoughts if something with the same interface would be feasible for SLURM?

mwojcikowski:
I have used [ipython-cluster-helper](https://github.com/roryk/ipython-cluster-helper), which is similar but has it's own quirks.

This commit is a proof-of-concept for me right now, so I'm more than welcome to re-think and redesign stuff.

The interface looks easy to implement, but current approach has one advantage - it produces one job. On one hand it's beneficial, if you get the resource then you have it all. On the other hand, all the resources are blocked for 100% of time (lets call it `greedy)`. This can be waste some of them if your workload is not balanced. The `lazy` approach would produce one scheduler job + n worker jobs on demand.

In my mind the sweet-spot is to queue all required resources at beginning and free them near the end when no more jobs are needed, although it's the "embarrassingly parallel joblib loops" speaking.

Keep in mind that on **shared** clusters you're not guarantied to get all resources at once, thus, in my opinion, you should rather queue whole dask cluster at once.

Ping #128 for any other thoughts.

mrocklin:
Closing this for now.

PR #430, opened 2016-08-17
Adapt work stealing to the number of idle workers

mrocklin:
If there are many more idle workers than saturated ones then we
accept a higher communication to computation ratio imbalance.


mrocklin:
In hindsight this should probably be proportional to the stack size and average task duration of the victim worker

mrocklin:
Closing in favor of #458

Issue #333, opened 2016-06-14
Memory doesn't get freed up after changing reference


hussainsultan:
```python
df = dd.read_csv('s3://xx/*.csv')
e.persist(df) #first result persists in memory

df = dd.read_csv('s3://yy/*.csv')
e.persist(df)
```

the memory on the nodes grows and df now references to the new data.However, it appears that original df doesnt get cleared in the worker memory. Is this expected behavior?

mrocklin:
First, `e.persist` returns a new dataframe, so you should actually say the following:

```python
# e.persist(df)  # don't do this
df = e.persist(df)  # do this instead
```

Second, if you're using IPython here then its possible that it's holding onto the anonymous result in its cache.

Does this problem persist if you provide names to things as suggested above?  If so does it still persist if you explicitly call `del` or `e.cancel` on the dataframe resulting from the `persist` call?

It could also be that you have some other result in RAM.  For example maybe you've done something like the following:

    total = e.compute(df.column.sum())

This would also keep `df` around in distributed ram.  

The scheduler keeps around all data for which any client has any future that depends on that data.  If the scheduler is holding onto RAM that you don't think it should then you should look around for futures (or persisted collections of futures) that depend on that data and delete them.  Alternatively, you should be able to cancel the collection, which should percolate out to all dependent futures.

hussainsultan:
ahh. it was Jupyter holding it in its cache. Thanks! Does this mean that if my Client disconnects, all the persisted data should go away?

mrocklin:
Yes

mrocklin:
Well, unless some other client is holding onto the exact same data

hussainsultan:
I disconnected all my Clients but the memory doesnt get freed up all the way. I looked at the scheduler logs and it registers my client to disconnect.

mrocklin:
That is surprising.  I would want to see more information about what's going on within the scheduler or a reproducible example, but that's probably difficult to produce.  We should probably expose more diagnostic information about the scheduler-client relationship just as we currently expose information about the scheduler-worker relationship.  

You can always call `e.restart()` which forces a clearing of all state and a hard reset of all workers.

hussainsultan:
I also think the desired behavior for long-running dask cluster is that even if all clients have disconnected, we should have a mechanism to keep some keys around that can then be referenced again. My understanding was flawed since i was assuming the above is true.

`e.restart` actually does clear everything.

mrocklin:
[...]

PR #431, opened 2016-08-17
Add coroutine to synchronize worker data

mrocklin:
It's possible that some extra data is getting lost on the workers.
This adds a scheduler coroutine that does a proper synchronization with
the worker.

It's not clear yet whether or not this is necessary.  If it is
necessary then we can put it in a long running PeriodicCallback.

Fixes https://github.com/dask/distributed/issues/333


mrocklin:
cc @quasiben @martindurant @jcrist for review

This is a nice PR to review.  It's isolated to a small bit of functionality and yet also touches a few of the core aspects to how things work.  This uses a bit of Tornado.

PR #433, opened 2016-08-18
Convert progress bars to annuli

mrocklin:
This allows us to fit more progress formation in the same horiztonal
space.

TODO
----

- [x] Figure out how to snap the view on the wedges.  Currently when the plot comes up we need to pan and zoom around to get a good fit.
- [x] Align names (inc, dec, ...) more consistently just below wedges.  This seems to shift a bit based on aspect ratio
- [ ] Style check with a viz person
- [ ] Maybe add a line between done and final counts, e.g. `1000 / 1000`

image

### DIFF
--- a/distributed/bokeh/status/main.py
+++ b/distributed/bokeh/status/main.py
@@ -10,7 +10,7 @@ from toolz import valmap
 
 from distributed.bokeh.status_monitor import progress_plot, task_stream_plot
 from distributed.bokeh.worker_monitor import resource_profile_plot
-from distributed.diagnostics.progress_stream import progress_quads
+from distributed.diagnostics.progress_stream import progress_wedge
 from distributed.utils import log_errors
 import distributed.bokeh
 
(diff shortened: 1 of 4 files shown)


mrocklin:
cc @birdsarah

mrocklin:
OK, the first issue is that bokeh isn't considering the radius of the annular wedges.  Here is how it looks by default when you first open the page:

image

mrocklin:
Second problem is that the distance between the text depends on the aspect ratio of the plot.  

image

mrocklin:
In the both images we see that we snap to all of the text elements.

In the second image, when there are only two annular wedges, we see that the distance between the lower text and the wedge is much smaller

mrocklin:
Turns out that I can control a lot of this by setting ranges explicitly.  This forces us to accept only a fixed number of progress wedges, but does look nice.

mrocklin:
OK, this looks decent now: 

image

canavandl:
It may be worth looking into using `Label` annotations instead of Text glyphs. The former can be placed in screen units (pixels) instead of data coordinates, which sometimes makes alignment easier.

http://bokeh.pydata.org/en/latest/docs/reference/models/annotations.html#bokeh.models.annotations.Label

mrocklin:
I think the problem with alignment wasn't with the text, it was with the annular wedges.  When the aspect ratio in pixels was off from the aspect ratio in coordinates the annular wedges are forced to either stop being circular on the screen or else choose one of the dimensions to apply their radius to.  They do the latter, choosing the smaller dimension.  This causes the annular wedges to shrink, causing the relative spacing to text to be quite different.

mrocklin:
Superceded by https://github.com/dask/distributed/pull/448  Closing.

PR #434, opened 2016-08-18
Add refresh button to the tasks web interface

mrocklin:
This adds a refresh button that updates the tasks in the tasks/
pane of the web interface.

Currently this takes up half of the space because columns in `stretch_both` mode split space evenly.  This is a known not-yet-implemented feature in Bokeh.  Perhaps there is a way around, given that the button itself doesn't need to scale and can possibly be placed outside of the general layout space?

cc @birdsarah

### DIFF
--- a/distributed/bokeh/tasks/main.py
+++ b/distributed/bokeh/tasks/main.py
@@ -4,6 +4,8 @@ from __future__ import print_function, division, absolute_import
 
 from toolz import valmap
 from bokeh.io import curdoc
+from bokeh.layouts import column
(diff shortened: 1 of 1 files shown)


birdsarah:
> Perhaps there is a way around, given that the button itself doesn't need to scale and can possibly be placed outside of the general layout space?

Yes - and with regular templating that would be easy - but server templating is not yet full fleshed out - suggest opening an issue (there may already be one)

mrocklin:
The tasks interface now refreshes itself, just on a slower rate than the status page.  Closing.

PR #435, opened 2016-08-18
Attempt to use panels/tabs in bokeh web interface

mrocklin:
This creates a single panel Tabs object around the status layout.
If we can get this to work then I'd like to merge all of the various views into
a single page.

Currently sizing isn't playing nicely.

cc @birdsarah


birdsarah:
Tabs are known to only work in fixed mode. It's an open issue on bokeh.

mrocklin:
OK, will hold off on this for now

Issue #437, opened 2016-08-18
bokeh plot memory display


caseyjlaw:
It would be helpful to have the bokeh web UI display memory use per worker and per task group. Memory intensive applications can get stuck (for a bit) retrying a computation that is too memory intensive. The current display does not make it clear that a memory limit is killing the computation.

mrocklin:
There are a few ways in which we could slice and dice this data and display it.

1.  A worker table showing, among other things, the total amount of memory consumed as a percentage (we used to have this)
2.  A simple plot showing the distribution of bytes among task groups, like 50% read_csv, 25% groupby, 25% sum, although in a more visual form.
3.  Something that tries to cross the two, like the plot of part 2 for every worker
4.  Something else ...

There are undoubtedly many ways to do this.  Sketches for what you think would be useful are most welcome.  I'm fooling around with the web interface now, so this issue is timely

caseyjlaw:
For my issue, I think the per-worker memory needs to be clear. So I think option 1 or 3 is best.
I'll elaborate on #438.

mrocklin:
I think that #438 is strictly option 2.  I think that in the end we'll want many such views.  I'll probably stand up option 1 in the near future.

mrocklin:
@caseyjlaw says 

> I wonder if the per-task-group memory bar could be accompanied by a per-worker memory bar? I'm imagining a bar in the same units and same design as the per-task-group, but showing how many bytes are used in each worker. The total number of bytes represented in both bars would be the same, so they'd be pretty easy to interpret at a glance.

So this would show us the relative distribution of bytes across the workers.  More or less it would let us quickly answer the question "are my bytes spread evenly across my workers."  

However, usually when I'm curious about workers and memory the question I want to answer is "How much memory is being used on my workers?" and, a corollary "Are any of my workers close to being saturated?"  I think that the plot style in #438 is bad at showing absolute quantities like these.  I think that we should consider other styles.  We have a lot of power with Bokeh, I think that we really just need to think about the right kind of plot.  I genuinely encourage sketches here.  I've been sitting down with paper and pen for a bit.

caseyjlaw:
I think you have summarized what I want well. I suppose stacked histograms are pretty bad at showing absolute values. I thought that with the hover label that bokeh does well, it would be easy to extract absolute values. Another consideration is that any per-worker visualization needs to work in the n=1  and n=many cases.
How about using lines instead of bars for per-worker memory? So maybe (sorry for ascii):

[...]

PR #438, opened 2016-08-18
Memory progress plot

mrocklin:
Add a simple plot to the status page to show memory consumption grouped by task label

image

Related to https://github.com/dask/distributed/issues/437

cc @caseyjlaw

cc @birdsarah for aesthetic suggestions (though please feel free to ignore if you're busy)

### DIFF
--- a/distributed/bokeh/status/main.py
+++ b/distributed/bokeh/status/main.py
@@ -8,9 +8,10 @@ from bokeh.io import curdoc
 from bokeh.layouts import column, row
 from toolz import valmap
 
-from distributed.bokeh.status_monitor import progress_plot, task_stream_plot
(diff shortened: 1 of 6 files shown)


caseyjlaw:
This is a nice addition. In my experiene, the per-worker memory state is also good for beginners (like me) to see.
I wonder if the per-task-group memory bar could be accompanied by a per-worker memory bar? I'm imagining a bar in the same units and same design as the per-task-group, but showing how many bytes are used in each worker. The total number of bytes represented in both bars would be the same, so they'd be pretty easy to interpret at a glance.

mrocklin:
I'm going to suggest that we keep more general worker-memory conversation over at #437 .  Will copy your comment there and respond.

caseyjlaw:
No problem.

PR #440, opened 2016-08-18
Worker diagnostics

mrocklin:
This adds back a simplified version of the worker table to the web diagnostics.  It also adds a simple plot showing memory percentage use of all workers and a Paragraph element that we might eventually populate with further information.

### DIFF
--- a/distributed/bokeh/application.py
+++ b/distributed/bokeh/application.py
@@ -12,7 +12,7 @@ from ..utils import ignoring
 
 dirname = os.path.dirname(distributed.__file__)
 paths = [os.path.join(dirname, 'bokeh', name)
-         for name in ['status', 'tasks']]
+         for name in ['status', 'tasks', 'workers']]
 
 binname = 'bokeh.exe' if sys.platform.startswith('win') else 'bokeh'
 binname = os.path.join(os.path.dirname(sys.argv[0]), binname)
--- a/distributed/bokeh/status/main.py
+++ b/distributed/bokeh/status/main.py
@@ -21,29 +21,6 @@ WIDTH = 600
 messages = distributed.bokeh.messages  # global message store
 doc = curdoc()
 
-"""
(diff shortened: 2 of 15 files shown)


mrocklin:
Relates to #437 cc @caseyjlaw care to give this a spin and recommend changes?

image

caseyjlaw:
I have not built distributed before, but it seems to work with a 'python setup.py install'. Starting scheduler and workers up reports no errors.
Loading web ui at localhost:8787/status looks new, but loading the workers view gives me a 404 error.

mrocklin:
Hrm, that's surprising.  You can also install from a git branch with 

    pip install git+https://github.com/mrocklin/distributed.git@worker-diagnostic --upgrade

Perhaps give that a try?

caseyjlaw:
Closer. With that build, the workers page loads, but I only see an empty table during and after a computation. The tasks and status pages look fine.

screen shot 2016-08-19 at 2 40 39 pm

mrocklin:
Hrm, I ran into this as well on my machine but was hoping that it was an environment problem (several other similar issues ended up being the result of stale versions lying around.)  It's nice to have the issue confirmed. I'll have a closer look.

It may end up working for you if you tear down and bring back up the scheduler a few times.  Seems to be caused by a race condition somewhere.

caseyjlaw:
The workers page never worked for showed info, even after several restarts
of workers and scheduler. My default conda Python is 2, fwiw.

mrocklin:
OK.  Good to know.  Thanks for giving it a shot

mrocklin:
@caseyjlaw I believe that the no-show plot issue has been resolved

caseyjlaw:
It works! 
Very nice and I am already learning that a memory limit is not causing one of my issues...

mrocklin:
Hooray for visual diagnostics!

PR #445, opened 2016-08-21
Dynamic submission of tasks from tasks

mrocklin:
This allows tasks to submit and depend on other tasks.  We emit and depend on new tasks by yielding dask collections.

### Examples

```python
def range(x, y):
    a, b = yield delayed(min)(x, y), delayed(max)(x, y)  # deploy new tasks to scheduler
    diff = yield delayed(sub)(a, b)  # deploy new task to scheduler
    return abs(diff)

future = e.submit(range, 10, 20)
```

```python
@delayed(pure=True)
def fib(n):
    if n in (0, 1):
        return n
    else:
        a, b = yield fib(n - 1), fib(n - 2)
        return a + b

future = e.compute(fib(10))
```

We spawn new tasks by yielding dask collections.  These tasks can themselves spawn more tasks (as shown in the fib example).  The final task waits for these tasks to finish before moving on to the next yield point.  This looks like tornado or asyncio. 

A worker identifies a self-expanding tasks as generator functions.  It uses `next` and `send` to run bits of the function piece by piece.   

This works with both submit and with compute/persist.  This only works with the distributed scheduler.  We will need to provide loud disclaimers that this doesn't work with the single-machine scheduler.

### Scheduling

This works by tricking the scheduler, and altering state like `tasks` and `dependencies` that was previously considered immutable.  Instead of getting a task-completed message we get a task-evolved message and tweak the scheduler state so that the submitted key is pushed back a bit, now with more dependencies that need to be computed.

This feels dangerous.  However we're able to achieve self-expanding tasks with much less invasion into the scheduler than was previously anticipated.

### TODO

- [x] Resilience to failure
- [x] Avoid moving generators around - this is like `restrictions` but has to be more about "always compute this task where this other task lives" rather than "always compute this task on these machines"
- [ ] Publish data through workers, not the task graph.  Currently data is submitted directly with the graph.  Instead we should put data into the local worker data dictionary and include a key reference.  This will avoid worker-scheduler I/O, data on the scheduler, and will enable data sharing more easily.

### DIFF
--- a/distributed/diagnostics/tests/test_widgets.py
+++ b/distributed/diagnostics/tests/test_widgets.py
@@ -85,7 +85,8 @@ def test_progressbar_widget(s, a, b):
                                              'y': (inc, 'x'),
                                              'z': (inc, 'y')}),
                    keys=['z'],
-                   dependencies={'y': {'x'}, 'z': {'y'}})
+                   dependencies={'y': {'x'}, 'z': {'y'}},
+                   client='alice')
 
     progress = ProgressWidget(['z'], scheduler=(s.ip, s.port))
     yield progress.listen()
@@ -109,7 +110,8 @@ def test_multi_progressbar_widget(s, a, b):
                    keys=['e'],
                    dependencies={'x-2': ['x-1'], 'x-3': ['x-2'],
                                  'y-1': ['x-3'], 'y-2': ['y-1'],
-                                 'e': ['y-2']})
+                                 'e': ['y-2']},
+                   client='alice')
 
     p = MultiProgressWidget(['e'], scheduler=(s.ip, s.port))
     yield p.listen()
@@ -145,7 +147,8 @@ def test_multi_progressbar_widget_after_close(s, a, b):
                    keys=['e'],
                    dependencies={'x-2': {'x-1'}, 'x-3': {'x-2'},
                                  'y-1': {'x-3'}, 'y-2': {'y-1'},
-                                 'e': {'y-2'}})
+                                 'e': {'y-2'}},
+                   client='alice')
 
     p = MultiProgressWidget(['x-1', 'x-2', 'x-3'], scheduler=(s.ip, s.port))
     yield p.listen()
@@ -218,7 +221,8 @@ def test_multibar_complete(s, a, b):
                    keys=['e'],
                    dependencies={'x-2': {'x-1'}, 'x-3': {'x-2'},
                                  'y-1': {'x-3'}, 'y-2': {'y-1'},
-                                 'e': {'y-2'}})
+                                 'e': {'y-2'}},
+                   client='alice')
 
     p = MultiProgressWidget(['e'], scheduler=(s.ip, s.port), complete=True)
     yield p.listen()
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -608,7 +608,8 @@ class Scheduler(Server):
         return recommendations
 
     def stimulus_task_evolves(self, key=None, worker=None, tasks=None,
-            dependencies=None, arg_key=None, new_key=None, **kwargs):
+            dependencies=None, arg_key=None, new_key=None, wrapped_task=None,
+            **kwargs):
         with log_errors():
             assert self.task_state[key] == 'processing'
             self.task_state[new_key] = self.task_state.pop(key)
@@ -618,7 +619,8 @@ class Scheduler(Server):
                 self.dependents[dep].add(new_key)
                 self.waiting_data[dep].remove(key)
                 self.waiting_data[dep].add(new_key)
-            self.tasks[new_key] = self.tasks.pop(key)
+            del self.tasks[key]
+            self.tasks[new_key] = wrapped_task
             self.waiting_data[new_key] = set()
             self.rprocessing[new_key] = self.rprocessing.pop(key)
             for w in self.rprocessing[new_key]:
@@ -629,7 +631,7 @@ class Scheduler(Server):
                                        {new_key: 'memory'}))
 
             self.update_graph(client=None, tasks=tasks, keys=[key],
-                    dependencies=dependencies, restrictions={key: [worker]})
+                    dependencies=dependencies)
 
             recommendations = self.transition(new_key, 'memory', worker=worker, **kwargs)
             if self.task_state[new_key] == 'memory':
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -606,6 +606,34 @@ class Scheduler(Server):
 
         return recommendations
 
+    def stimulus_task_evolves(self, key=None, worker=None, tasks=None,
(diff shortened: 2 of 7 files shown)


mrocklin:
Cleaner example with fib.  Running all the way through:

```python
In [1]: from distributed import Executor, LocalCluster

In [2]: c = LocalCluster(nanny=False)

In [3]: e = Executor(c)

In [4]: from dask import delayed

In [5]: @delayed(pure=True)
def fib(n):
    if n in (0, 1):
        return n
    else:
        a, b = yield fib(n - 1), fib(n - 2)
        return a + b
   ...:     

In [6]: future = e.compute(fib(10))

In [7]: future.result()
Out[7]: 55

In [8]: e.cluster.scheduler.task_state
Out[8]: 
{'fib-229df9ea4f4fec7db9c5a3aa3eda9836': 'released',
 'fib-229df9ea4f4fec7db9c5a3aa3eda9836-0': 'released',
 'fib-2c898d43edd43c283c3b917ee7c66ecf': 'released',
 'fib-2c898d43edd43c283c3b917ee7c66ecf-0': 'released',
 'fib-3521178b6c42fe6f9df0eda05516c100': 'released',
 'fib-3521178b6c42fe6f9df0eda05516c100-0': 'released',
 'fib-376601cb9ba805ab61dbf926d2c40bf0': 'released',
 'fib-376601cb9ba805ab61dbf926d2c40bf0-0': 'released',
 'fib-7a47039e60d9bec34d255bfa551dee17': 'released',
 'fib-7a47039e60d9bec34d255bfa551dee17-0': 'released',
 'fib-d6726ffda3ca281823c223e0ae356ad8': 'released',
 'fib-d6726ffda3ca281823c223e0ae356ad8-0': 'released',
 'fib-dd75da11fb7e34fbbd2ecdb8cdcbadc2': 'released',
 'fib-dd75da11fb7e34fbbd2ecdb8cdcbadc2-0': 'released',
 'fib-e2b35668f0946bfe7e253abbccca2dd2': 'released',
 'fib-e2b35668f0946bfe7e253abbccca2dd2-0': 'released',
 'fib-eb075c767f71bc544013ba2a8f7f7724': 'released',
 'fib-eb075c767f71bc544013ba2a8f7f7724-0': 'released',
 'fib-f4a93e5c116037a1dce773c8e73e38cf': 'released',
 'fib-f8b91cdb7d59d9c7aaa72747fa87ec39': 'released',
 'finalize-2214302ab34e529fc2a73373b84f9647': 'memory',
 'tuple-0c765d81-8bd2-4a5d-aa46-6b0942a33214': 'released',
 'tuple-15e7a6af-75ad-42fa-9b64-d21af0f0b109': 'released',
 'tuple-991c42a0-1446-4300-9d90-9898426a26dc': 'released',
 'tuple-ce94f462-26ba-454c-96c4-eed6d3f0812d': 'released',
 'tuple-d2609dd1-f822-4f7f-9483-be0d706d66eb': 'released',
 'tuple-d5964310-1f75-4164-94ea-8b58f226ab39': 'released',
 'tuple-ebeea06e-403d-49d1-8522-b0b923098282': 'released',
 'tuple-ee096349-e633-4a30-bf10-00a932f2f3fc': 'released',
 'tuple-f66b8701-5b71-407c-9c13-ad4af00b7d05': 'released'}
```

mrocklin:
cc @ogrisel @minrk @shoyer @broxtronix for feedback.  This adds a significant feature to raw task scheduling.  Your critical feedback here would be welcome.

seibert:
Question about the API: The yield blocks until all the tasks are complete, right?  So if I want to programmatically generate a number of independent tasks to run in parallel, I need to collect them into a tuple and yield them all at once?

mrocklin:
Hrm, interesting.  Yes at the moment you couldn't submit tasks, wait on just a few, and then wait on others later.  I can see how that would be an interesting addition.

seibert:
I only ask because I'm using CUDA's dynamic parallelism as a mental guide for how this might look.  In that API, launching subtasks and blocking until it they are complete are separate operations.

mrocklin:
Yeah, so too in dask.distributed's normal Executor model...

mrocklin:
I wonder how far we could get with a global `client` or `executor` object.

eriknw:
Interesting, and Stan brings up a very good point.  Perhaps the values sent to the generator (the left hand side of `= yield`) could be `Future` objects from an `Executor`.  At this point, though, it may be clearer just to use an `executor` object directly.

I think a use case to target is a split-apply-combine for which you don't know the number of groups *a priori*, and to allow this computation to be done both cpu- and memory-efficient.

What happens if the generator returns a `Delayed` object?  Is this sensible?

mrocklin:
> Interesting, and Stan brings up a very good point. Perhaps the values sent to the generator (the left hand side of = yield) could be Future objects from an Executor. At this point, though, it may be clearer just to use an executor object directly.

I agree.  I'm now in favor of trashing this implementation, and instead giving every `Worker` an `Executor` that it can use to submit jobs if necessary.  My initial impression of this was that it was a heavy solution.  On further reflection though I think that we can make it very lightweight.  It also feels cleaner than what is proposed here.  There are some drawbacks though, tasks will take up a worker slot while they wait for other tasks.  In the case of highly spreading computations, like a non-memoizing `fib`, it's entirely possible that you could fill the cluster with tasks that were just waiting on other tasks.  

> I think a use case to target is a split-apply-combine for which you don't know the number of groups a priori, and to allow this computation to be done both cpu- and memory-efficient.

Yes, generally any time when the computation depends on data values something like this becomes valuable.  The previous (and still very viable) solution would have been to have the worker report back to the client (through some small future) the necessary information to construct the rest of the graph.  This is still reasonable.  We're just cutting out one network roundtrip

seibert:
Yeah, this is the same problem CUDA had with their dynamic parallelism, and so they had to implement the ability (roughly speaking) for a task to be unscheduled off the hardware in order to ensure tasks could continue to make progress.

mrocklin:
I'd be tempted to completely ignore that consideration until it came up in practice

seibert:
I believe the unscheduling happened at the point whether the task blocked waiting for results.

ogrisel:
Indeed this is a very important feature.

> There are some drawbacks though, tasks will take up a worker slot while they wait for other tasks.

If this pattern is repeated `n_workers` time that could trigger a global deadlock no? One should ensure that the workers are waiting actively in the sense that can steal the jobs they submitted instead of just waiting.

mrocklin:
> If this pattern is repeated n_workers times that could trigger a global deadlock no? 

Yes, that is correct.

mrocklin:
I've implemented the "Executor's everywhere" solution in https://github.com/dask/distributed/pull/471

It is prone to the deadlock issue stated above.

mrocklin:
Closing this in favor of #471 which now seems to resolve all issues, including deadlock concerns.

Issue #418, opened 2016-08-12
Connection refused when trying to view Web UI using SSH port forwarding


stsievert:
I'm getting started with distributed and running it on a remote cluster, and am trying to verify that it's talking between two different compute machines in a cluster.

Specifically, I run the commands

``` shell
scott@local-machine: ssh -L 8000:localhost:8787 scott@remote-server-1
scott@remote-server-1: dask-scheduler
scott@local-machine: ssh scott@remote-server-2
scott@remote-server-2: dask-worker <ip addr from dask-scheduler>
```

Using this, I can run dask scripts -- I believe it's working, but would like to see. On my local machine I try to access `http://localhost:8000/status` to see the Dask Web UI. When I do this, I see the below error on `remote-machine-1`:

``` bash
channel 3: open failed: connect failed: Connection refused
channel 3: open failed: connect failed: Connection refused
channel 3: open failed: connect failed: Connection refused
```

I've played with it, but I'm not sure where to go next. I would appreciate seeing the [Web UI docs] being expanded on to mention this issue.

[Web UI docs]:http://distributed.readthedocs.io/en/latest/web.html

mrocklin:
> I would appreciate seeing the Web UI docs being expanded on to mention this issue.

I'd be happy to expand the documentation.  But first, what do you think the problem is?  I don't know what your network looks like enough to be able to reasonably solve your problem.  Networks are hard.  Are you able to see other web servers that you host on your network?  How about bokeh server applications?  Are you able to get things running ok on your local machine?

stsievert:
> Are you able to get things running ok on your local machine?

When running on my local machine, I can see the Web UI.

> Are you able to see other web servers that you host on your network? 

With the commands below, I can access a Jupyter notebook on my local machine by going to `http://localhost:8000`.

``` python
scott@local-machine: ssh -L 8000:localhost:8888 scott@remote-server-1
scott@remote-server-1: jupyter notebook
... uses links on remote machine to start jupyter notebook on remote machine ...
```

> what do you think the problem is?

I think the problem is network-related -- somehow SSH tunneling isn't working. I would guess that I need to bind 144.92.142.191:8787 (the IP address given by `dask-scheduler`) to localhost:8787 on the remote machine then SSH in with `ssh -L 8000:localhost:8787`?

mrocklin:
You could try passing the address by which you're calling the server to the bokeh whitelist as follows:

    dask-scheduler --bokeh-whitelist localhost:8000

[...]

PR #447, opened 2016-08-22
DOC: mentions SSH port forwarding on web ui page

stsievert:
I added a couple paragraphs about what I experienced in #418. I give an example of port forwarding.

### DIFF
--- a/docs/source/web.rst
+++ b/docs/source/web.rst
@@ -38,6 +38,30 @@ tasks.  For the most recent 20000 tasks visit
 http://my-scheduler-address:8787/tasks , although beware that this page is not
 updated in real time.
 
+Viewing Web UI when running distributed on remote cluster
+---------------------------------------------------------
+Viewing the Web UI on a local machine when `distributed` is running on a remote
+cluster is possible through SSH port forwarding. Typically, this look something
+like
+
+.. code::
+
+   local$ ssh -L 8000:localhost:8787 user@remote
+   remote$ dask-scheduler
+
+It is then possible to go to ``localhost:8000`` and see Dask Web UI.
+
+This can be used in conjuction with port forwarding on the Jupyter notebook
+port.
+
+.. code::
+
+   local$ ssh -L 8001:localhost:8888 user@remote  # 8888 or whatever port Jupyter is running on
+   remote$ jupyter notebook
+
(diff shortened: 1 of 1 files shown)


mrocklin:
Cool.  One concern I have is that people may interpret this as the preferred way to reach the web interface, rather than simply pointing their browser to an accessible HTTP route.  If we include this in our documentation then we need to make sure that it is appropriately scoped so that novice users don't trip up.

stsievert:
Addressed. I added a section on the default behavior. I said if that doesn't work, one way to get around it is via port forwarding.

docs/source/web.rst
```diff
@@ -38,6 +38,50 @@ tasks.  For the most recent 20000 tasks visit
...
+.. code:: bash
+
+   local$ ssh -L 8000:localhost:8787 user@remote
+   remote$ dask-scheduler  # now, the web UI is visible at localhost:8000
+   remote$ # continue to set up dask if needed -- add workers, etc
+
+It is then possible to go to ``localhost:8000`` and see Dask Web UI.
+
+This can be used in conjuction with port forwarding on the Jupyter notebook
+port.
```
mrocklin: Why the note here about Jupyter notebooks?

stsievert: Because it's how I'm using distributed and I think it's useful.

mrocklin: Hrm, perhaps generalize to say something like.  "This same approach is not specific to dask.distributed, but can be used by any service that operates over a network, such as Jupyter notebooks."

Alternatively, perhaps you could write a tiny external blogpost about how many python services like Jupyter and Dask are now providing web interfaces and how you can access them through ssh tunneling.  Then we could link to that page from these docs.

There are enough varied users of Dask that operate in such different ways that I want to make sure we don't get pulled in the game of documenting them all.  I'm afraid that things will start to become confusing as novice users have to figure out what documentation is essential and what documentation was particular to a specific way of using the system.

stsievert: I added Jupyter as an example.

> you could write a tiny external blogpost

I've been drafting that post (and filed this issue while preparing to write it). I plan to cover how I use dask with UW–Madison's cluster and my personal setup (web UI + remote Jupyter notebook).

mrocklin:
Merging!  Thanks for getting this in.

PR #453, opened 2016-08-23
Publish datasets on scheduler

### LINKED ISSUE #444: Publish collections/futures
From one client/executor I may want to create a dataset and then share it with my colleagues working from other clients/executors.  Currently I can pickle my dask collection, send it to them, they unpickle it, and through the creation of Futures they get registered as clients to the data.  However this pickling-and-sending process is inconvenient.  We could publish collections to the scheduler itself

### Client 1

```python
e.publish_dataset(df, 'analysis')
```

### Client 2

```python
df = e.collect_dataset('analysis')
```

### Scheduler

```python
>>> s.published_datasets
{'analysis': {'thing-to-send-to-client': ...,
              'relevant-keys': {set of keys}}
```

This is completely orthogonal functionality for the scheduler, but fairly trivial to build.  It should be a couple of handlers and a dictionary on the scheduler, a couple of methods on the Executor, and a sane way to collect futures and things that wrap futures into something that we can store in the dictionary.

This would be an interesting and moderately challenging task for someone like @jcrist or @martindurant if they have interest.  It would touch major pieces of code peripherally but it isn't conceptually or algorithmically tricky.

martindurant:
Allows for the publication of datasets on the scheduler

### Client 1

```python
df = dd.read_csv('s3://...')
df = e.persist(df)

e.publish_dataset(df, 'my-data')
```

### Client 2

```python
df = e.get_dataset('my-data')
```

Fixes #444

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -1015,16 +1015,16 @@ class Executor(object):
         """
         sync(self.loop, self.scheduler.unpublish_dataset, name=name)
 
-    def published_datasets(self):
+    def list_datasets(self):
         """ Returns list of named datasets referenced in the scheduler
 
         See publish_dataset()
         """
-        return sync(self.loop, self.scheduler.get_published_keys)
+        return sync(self.loop, self.scheduler.list_datasets)
 
     @gen.coroutine
-    def _get_published_dataset(self, name):
-        out = yield self.scheduler.get_published_dataset(name=name,
+    def _get_dataset(self, name):
+        out = yield self.scheduler.get_dataset(name=name,
                                                          client=self.id)
         data, keys = out['data'], out['keys']
 
@@ -1032,8 +1032,12 @@ class Executor(object):
             data = loads(data)
         raise Return(data)
 
-    def get_published_dataset(self, name):
-        return sync(self.loop, self._get_published_dataset, tokey(name))
+    def get_dataset(self, name):
+        """ Fetch named dataset from the scheduler.
+
+        See publish_dataset()
+        """
+        return sync(self.loop, self._get_dataset, tokey(name))
 
     @gen.coroutine
     def _run(self, function, *args, **kwargs):
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -990,9 +990,8 @@ class Executor(object):
     def _publish_dataset(self, data, name):
         keys = [tokey(f.key) for f in futures_of(data)]
         yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
-                                             data=pickle.dumps(data),
+                                             data=dumps(data),
                                              client=self.id)
-
 
     def publish_dataset(self, data, name):
         """
@@ -1002,8 +1001,7 @@ class Executor(object):
         ----------
         data: list of futures or dask collection
         name: str or container
-            handle by which the stored data should be known. If not a string,
-            will be wrapped by tokey() to deterministically turn into one.
+            handle by which the stored data should be known.
 
         Returns
         -------
@@ -1011,27 +1009,18 @@ class Executor(object):
         """
         sync(self.loop, self._publish_dataset, data, name)
 
-    @gen.coroutine
-    def _unpublish_dataset(self, name):
-        yield self.scheduler.unpublish_dataset(name=name)
-
     def unpublish_dataset(self, name):
         """
         Remove reference to data on the scheduler
         """
-        sync(self.loop, self._unpublish_dataset, name)
-
-    @gen.coroutine
-    def _published_datasets(self):
-        resp = yield self.scheduler.get_published_keys()
-        raise Return(resp)
+        sync(self.loop, self.scheduler.unpublish_dataset, name=name)
 
     def published_datasets(self):
         """ Returns list of named datasets referenced in the scheduler
 
         See publish_dataset()
         """
-        return sync(self.loop, self._published_datasets)
+        return sync(self.loop, self.scheduler.get_published_keys)
 
     @gen.coroutine
     def _get_published_dataset(self, name):
@@ -1040,7 +1029,7 @@ class Executor(object):
         data, keys = out['data'], out['keys']
 
         with temp_default_executor(self):
-            data = pickle.loads(data)
+            data = loads(data)
         raise Return(data)
 
     def get_published_dataset(self, name):
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -989,9 +989,10 @@ class Executor(object):
     @gen.coroutine
     def _publish_dataset(self, data, name):
         keys = [tokey(f.key) for f in futures_of(data)]
-        yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
+        out = yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
                                              data=dumps(data),
                                              client=self.id)
+        raise Return(out)
 
     def publish_dataset(self, data, name):
         """
@@ -1007,7 +1008,7 @@ class Executor(object):
         -------
         None
         """
-        sync(self.loop, self._publish_dataset, data, name)
+        return sync(self.loop, self._publish_dataset, data, name)
 
     def unpublish_dataset(self, name):
         """
@@ -1024,10 +1025,11 @@ class Executor(object):
 
     @gen.coroutine
     def _get_dataset(self, name):
-        out = yield self.scheduler.get_dataset(name=name,
-                                                         client=self.id)
+        out = yield self.scheduler.get_dataset(name=name, client=self.id)
         data, keys = out['data'], out['keys']
 
+        if not data:
+            raise Return(None)
         with temp_default_executor(self):
             data = loads(data)
         raise Return(data)
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -992,7 +992,8 @@ class Executor(object):
         out = yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
                                              data=dumps(data),
                                              client=self.id)
-        raise Return(out)
+        if out['status'] != "OK":
+            raise KeyError(name)
 
     def publish_dataset(self, data, name):
         """
@@ -1008,7 +1009,7 @@ class Executor(object):
         -------
         None
         """
-        return sync(self.loop, self._publish_dataset, data, name)
+        sync(self.loop, self._publish_dataset, data, name)
 
     def unpublish_dataset(self, name):
         """
@@ -1029,7 +1030,7 @@ class Executor(object):
         data, keys = out['data'], out['keys']
 
         if not data:
-            raise Return(None)
+            raise KeyError(name)
         with temp_default_executor(self):
             data = loads(data)
         raise Return(data)
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -989,8 +989,8 @@ class Executor(object):
     @gen.coroutine
     def _publish_dataset(self, data, name):
         keys = [tokey(f.key) for f in futures_of(data)]
-        yield self.scheduler.publish_data(keys=keys, name=tokey(name),
-                                          data=pickle.dumps(data))
+        yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
+                                             data=pickle.dumps(data))
 
 
     def publish_dataset(self, data, name):
@@ -1024,7 +1024,8 @@ class Executor(object):
 
     @gen.coroutine
     def _get_published_dataset(self, name):
-        out = yield self.scheduler.get_published_data(name=name, client=self.id)
+        out = yield self.scheduler.get_published_dataset(name=name,
+                                                         client=self.id)
         data, keys = out['data'], out['keys']
 
         with temp_default_executor(self):
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -1012,6 +1012,16 @@ class Executor(object):
         sync(self.loop, self._publish_dataset, data, name)
 
     @gen.coroutine
+    def _unpublish_dataset(self, name):
+        yield self.scheduler.unpublish_dataset(name=name)
+
+    def unpublish_dataset(self, name):
+        """
+        Remove reference to data on the scheduler
+        """
+        sync(self.loop, self._unpublish_dataset, name)
+
+    @gen.coroutine
     def _published_datasets(self):
         resp = yield self.scheduler.get_published_keys()
         raise Return(resp)
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -2,12 +2,14 @@ from __future__ import print_function, division, absolute_import
 
 from collections import defaultdict, Iterator, Iterable
 from concurrent.futures._base import DoneAndNotDoneFutures, CancelledError
+from contextlib import contextmanager
 import copy
 from datetime import timedelta
 from functools import partial
 import logging
 import os
 import sys
+import pickle
 from time import sleep
 import uuid
 from threading import Thread
@@ -191,6 +193,15 @@ class Future(WrappedKey):
         if not self._cleared and self.executor.generation == self._generation:
             self._cleared = True
             self.executor._dec_ref(tokey(self.key))
+
+    def __getstate__(self):
+        return self.key
+
+    def __setstate__(self, key):
+        e = default_executor()
+        Future.__init__(self, key, e)
+        e._send_to_scheduler({'op': 'update-graph', 'tasks': {},
+                              'keys': [tokey(self.key)], 'client': e.id})
 
     def __del__(self):
         self.release()
@@ -974,6 +985,54 @@ class Executor(object):
         futures: list of Futures
         """
         return sync(self.loop, self._cancel, futures)
+
+    @gen.coroutine
+    def _publish_dataset(self, data, name):
+        keys = [tokey(f.key) for f in futures_of(data)]
+        yield self.scheduler.publish_data(keys=keys, name=tokey(name),
+                                          data=pickle.dumps(data))
+
+
+    def publish_dataset(self, data, name):
+        """
+        Store a named reference to the data in the scheduler for other executors
+
+        Parameters
+        ----------
+        data: list of futures or dask collection
+        name: str or container
+            handle by which the stored data should be known. If not a string,
+            will be wrapped by tokey() to deterministically turn into one.
+
+        Returns
+        -------
+        None
+        """
+        sync(self.loop, self._publish_dataset, data, name)
+
+    @gen.coroutine
+    def _published_datasets(self):
+        resp = yield self.scheduler.get_published_keys()
+        raise Return(resp)
+
+    def published_datasets(self):
+        """ Returns list of named datasets referenced in the scheduler
+
+        See publish_dataset()
+        """
+        return sync(self.loop, self._published_datasets)
+
+    @gen.coroutine
+    def _get_published_dataset(self, name):
+        out = yield self.scheduler.get_published_data(name=name, client=self.id)
+        data, keys = out['data'], out['keys']
+
+        with temp_default_executor(self):
+            data = pickle.loads(data)
+        raise Return(data)
+
+    def get_published_dataset(self, name):
+        return sync(self.loop, self._get_published_dataset, tokey(name))
 
     @gen.coroutine
     def _run(self, function, *args, **kwargs):
@@ -1872,3 +1931,20 @@ def futures_of(o, executor=None):
             raise CancelledError(bad)
 
     return list(futures)
+
+
+@contextmanager
+def temp_default_executor(e):
+    """ Set the default executor for the duration of the context
+
+    Parameters
+    ----------
+    e : Executor
+        This is what default_executor() will return within the with-block.
+    """
+    old_exec = default_executor()
+    _global_executor[0] = e
+    try:
+        yield
+    finally:
+        _global_executor[0] = old_exec
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -990,7 +990,8 @@ class Executor(object):
     def _publish_dataset(self, data, name):
         keys = [tokey(f.key) for f in futures_of(data)]
         yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
-                                             data=pickle.dumps(data))
+                                             data=pickle.dumps(data),
+                                             client=self.id)
 
 
     def publish_dataset(self, data, name):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -929,7 +929,6 @@ class Scheduler(Server):
         """ Remove client from network """
         logger.info("Remove client %s", client)
         self.client_releases_keys(self.wants_what.get(client, ()), client)
-        self.check_published_datasets(client)
         with ignoring(KeyError):
             del self.wants_what[client]
 
@@ -1599,8 +1598,7 @@ class Scheduler(Server):
     def publish_dataset(self, stream=None, keys=None, data=None, name=None,
                         client=None):
         self.client_wants_keys(keys, 'published-%s' % name)
-        self.datasets[name] = {'data': data, 'keys': keys,
-                                     'clients': {client}}
+        self.datasets[name] = {'data': data, 'keys': keys}
 
     def unpublish_dataset(self, stream=None, name=None):
         out = self.datasets.pop(name, {'keys': []})
@@ -1610,18 +1608,7 @@ class Scheduler(Server):
         return list(sorted(self.datasets.keys()))
 
     def get_dataset(self, stream, name=None, client=None):
-        self.datasets[name]['clients'].add(client)
-        data = self.datasets[name].copy()
-        del data['clients']
-        return data
-
-    def check_published_datasets(self, client):
-        for key, data in self.datasets.copy().items():
-            with ignoring(KeyError):
-                data['clients'].remove(client)
-            if not(data['clients']):
-                del self.datasets[key]
-                self.client_releases_keys(data['keys'], 'published-%s' % key)
+        return self.datasets[name]
 
     #####################
     # State Transitions #
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1598,15 +1598,13 @@ class Scheduler(Server):
 
     def publish_dataset(self, stream=None, keys=None, data=None, name=None,
                         client=None):
-        for k in keys:
-            self.who_wants[k].add('published-%s' % name)
+        self.client_wants_keys(keys, 'published-%s' % name)
         self.published_data[name] = {'data': data, 'keys': keys,
                                      'clients': {client}}
 
     def unpublish_dataset(self, stream=None, name=None):
         out = self.published_data.pop(name, {'keys': []})
-        for k in out['keys']:
-            self.who_wants[k].remove('published-%s' % name)
+        self.client_releases_keys(out['keys'], 'published-%s' % name)
 
     def get_published_keys(self, *args):
         return list(sorted(self.published_data.keys()))
@@ -1621,10 +1619,9 @@ class Scheduler(Server):
         for key, data in self.published_data.copy().items():
             with ignoring(KeyError):
                 data['clients'].remove(client)
-                if not(data['clients']):
-                    del self.published_data[key]
-                    for k in data['keys']:
-                        self.who_wants[k].remove('published-%s' % key)
+            if not(data['clients']):
+                del self.published_data[key]
+                self.client_releases_keys(data['keys'], 'published-%s' % key)
 
     #####################
     # State Transitions #
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -223,7 +223,7 @@ class Scheduler(Server):
         self.exceptions = dict()
         self.tracebacks = dict()
         self.exceptions_blame = dict()
-        self.published_data = dict()
+        self.datasets = dict()
         self.stealable = [set() for i in range(12)]
         self.stealable_unknown_durations = defaultdict(set)
 
@@ -1599,28 +1599,28 @@ class Scheduler(Server):
     def publish_dataset(self, stream=None, keys=None, data=None, name=None,
                         client=None):
         self.client_wants_keys(keys, 'published-%s' % name)
-        self.published_data[name] = {'data': data, 'keys': keys,
+        self.datasets[name] = {'data': data, 'keys': keys,
                                      'clients': {client}}
 
     def unpublish_dataset(self, stream=None, name=None):
-        out = self.published_data.pop(name, {'keys': []})
+        out = self.datasets.pop(name, {'keys': []})
         self.client_releases_keys(out['keys'], 'published-%s' % name)
 
     def list_datasets(self, *args):
-        return list(sorted(self.published_data.keys()))
+        return list(sorted(self.datasets.keys()))
 
     def get_dataset(self, stream, name=None, client=None):
-        self.published_data[name]['clients'].add(client)
-        data = self.published_data[name].copy()
+        self.datasets[name]['clients'].add(client)
+        data = self.datasets[name].copy()
         del data['clients']
         return data
 
     def check_published_datasets(self, client):
-        for key, data in self.published_data.copy().items():
+        for key, data in self.datasets.copy().items():
             with ignoring(KeyError):
                 data['clients'].remove(client)
             if not(data['clients']):
-                del self.published_data[key]
+                del self.datasets[key]
                 self.client_releases_keys(data['keys'], 'published-%s' % key)
 
     #####################
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1598,10 +1598,10 @@ class Scheduler(Server):
     def publish_dataset(self, stream=None, keys=None, data=None, name=None,
                         client=None):
         if name in self.datasets:
-            return {'success': 'false'}
+            return {'status': 'failed'}
         self.client_wants_keys(keys, 'published-%s' % name)
         self.datasets[name] = {'data': data, 'keys': keys}
-        return {'success':  'true'}
+        return {'status':  'OK'}
 
     def unpublish_dataset(self, stream=None, name=None):
         out = self.datasets.pop(name, {'keys': []})
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -928,6 +928,7 @@ class Scheduler(Server):
         """ Remove client from network """
         logger.info("Remove client %s", client)
         self.client_releases_keys(self.wants_what.get(client, ()), client)
+        self.check_published_datasets(client)
         with ignoring(KeyError):
             del self.wants_what[client]
 
@@ -1594,15 +1595,26 @@ class Scheduler(Server):
 
             return result
 
-    def publish_dataset(self, stream=None, keys=None, data=None, name=None):
-        self.published_data[name] = {'data': data, 'keys': keys}
-        # TODO: add guards against deletion of keys?
+    def publish_dataset(self, stream=None, keys=None, data=None, name=None,
+                        client=None):
+        self.published_data[name] = {'data': data, 'keys': keys,
+                                     'clients': {client}}
 
     def get_published_keys(self, *args):
         return list(sorted(self.published_data.keys()))
 
     def get_published_dataset(self, stream, name=None, client=None):
-        return self.published_data[name]
+        self.published_data[name]['clients'].add(client)
+        data = self.published_data[name].copy()
+        del data['clients']
+        return data
+
+    def check_published_datasets(self, client):
+        for key, data in self.published_data.copy().items():
+            with ignoring(KeyError):
+                data['clients'].remove(client)
+                if not(data['clients']):
+                    del self.published_data[key]
 
     #####################
     # State Transitions #
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -263,8 +263,8 @@ class Scheduler(Server):
                          'rebalance': self.rebalance,
                          'replicate': self.replicate,
                          'start_ipython': self.start_ipython,
-                         'get_published_keys': self.get_published_keys,
-                         'get_published_dataset': self.get_published_dataset,
+                         'list_datasets': self.list_datasets,
+                         'get_dataset': self.get_dataset,
                          'publish_dataset': self.publish_dataset,
                          'unpublish_dataset': self.unpublish_dataset}
 
@@ -1606,10 +1606,10 @@ class Scheduler(Server):
         out = self.published_data.pop(name, {'keys': []})
         self.client_releases_keys(out['keys'], 'published-%s' % name)
 
-    def get_published_keys(self, *args):
+    def list_datasets(self, *args):
         return list(sorted(self.published_data.keys()))
 
-    def get_published_dataset(self, stream, name=None, client=None):
+    def get_dataset(self, stream, name=None, client=None):
         self.published_data[name]['clients'].add(client)
         data = self.published_data[name].copy()
         del data['clients']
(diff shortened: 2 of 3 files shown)


martindurant:
@mrocklin , I feel I am getting this totally wrong! ;)

mrocklin:
Totally wrong would be great.  I think that the main objective here is learning.

Looking...

distributed/executor.py
```diff
@@ -1798,3 +1829,62 @@ def futures_of(o, executor=None):
             raise CancelledError(bad)
 
     return list(futures)
+
+
+@contextmanager
+def serialize_for_publish(x):
```
mrocklin: This approach seems dangerous.  We might instead try to make a copy with `from copy import copy`

distributed/executor.py
```diff
@@ -975,6 +976,36 @@ def cancel(self, futures):
...
+        None
+        """
+        with serialize_for_publish(data) as out:
+            sdata, keys = out
+            self._send_to_scheduler({'op': 'publish-data', 'keys': keys,
+                                     'data': sdata, 'name': name})
+
+    @property
+    def published_datasets(self):
+        if hasattr(self, '_loop_thread'):
```
mrocklin: Why did you need this check?

distributed/scheduler.py
```diff
@@ -1554,6 +1556,15 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
 
             return result
 
+    @property
+    def published_data_keys(self):
+        return set(self.published_data.keys())
+
+    def publish_data(self, keys=None, data=None, name=None):
+        print('puiblish!', keys, data, name)
+        self.published_data[name] = (data, keys)
```
mrocklin: FWIW I recommend a dictionary here (dictionaries everywhere!) just to be more informative for other people coming to the code.

distributed/executor.py
martindurant: I saw it elsewhere - should you not check that the scheduler is around and listening before trying to pull a response? In any case, this doesn't work yet.

distributed/executor.py
martindurant: The idea was to make a serializable version temporarily within the context manager. There are probably some nice functions in your util code to walk around dask and dask-like objects; but things already persisted should tend to be shallow and easy to work with.

distributed/tests/test_executor.py
```diff
@@ -3576,3 +3576,25 @@ def test_idempotence(s, a, b):
...
+    assert d
+
+
+@gen_cluster(executor=True)
+def test_publish_simple(e, s, a, b):
+    data = yield e._scatter(range(12))
+    e.publish_dataset(data, 'data')
+    yield gen.sleep(0.2)
+
+    assert s.published_data
```
mrocklin: You should be able to make a pair of executors to test this.  Here is an example test:

```python
@gen_cluster()
def test_two_consecutive_executors_share_results(s, a, b):
    from random import randint
    e = Executor((s.ip, s.port), start=False)
    yield e._start()

    x = e.submit(randint, 0, 1000, pure=True)
    xx = yield x._result()

    f = Executor((s.ip, s.port), start=False)
    yield f._start()

    y = f.submit(randint, 0, 1000, pure=True)
    yy = yield y._result()

    assert xx == yy

    yield e._shutdown()
    yield f._shutdown()
```

distributed/scheduler.py
martindurant: OK (and sorry about the stray debug print)

distributed/tests/test_executor.py
martindurant: This is, effectively, recreating the whole graph for each executor and noticing that a key matching that output already exists. I was wanting to pass the keys themselves around. 
The assert above is to show that some reference is being stored on the scheduler.

distributed/executor.py
mrocklin: I recommend that you start with just a single collection:

    e.publish_dataset('name', my_collection)

You can get the keys from any dask collection, I think, by using the `._keys()` method.  There is also the `distributed.executor.py::futures_of` function, which may be useful.

My approach to making a serializable version would be to make a copy of the collection object and then swap out the dask attributes.

```python
from copy import copy
my_collection = copy(my_collection)
my_collection.dask = {}
```

However this won't work on Delayed values, which hold their dask graphs elsewhere.  Maybe we need a generic `._metadata()` method for dask objects or something.  Pinging @jcrist for whenever he gets back to civilization.

distributed/tests/test_executor.py
mrocklin: Yeah, sorry, the test is copy-pasted from the test suite.  The relevant part is how to create and shut down executors manually, so that you can have two of them.  This is how I would recommend testing that we can move collections between Executors.

mrocklin:
```python
@gen_cluster()
def test_your_test(s, a, b):
    e = Executor((s.ip, s.port), start=False)
    yield e._start()

    f = Executor((s.ip, s.port), start=False)
    yield f._start()

    # Your code

    yield e._shutdown()
    yield f._shutdown()
```

martindurant: OK, but that's for the next test - this one just says that the data made it to the scheduler. Getting it back is another matter!
As an aside, start=True would create a blocking version of the Executor?

mrocklin: yeah, sorry, this comment was intended to be in the same space of the tests, not referring to this test in particular.  I ought to have been more clear from the get-go.

The `start=True` keyword will call `sync(loop, e._start)`, which is bad because we can't use any synchronous calls within the `@gen_cluster` tests.  I wouldn't worry about this for now.  You rarely need to care.  We only have to deal with this this time because we need multiple executors, and so are slightly outside of the abilities of `@gen_cluster`.

distributed/executor.py
martindurant: That's effectively what I'm doing, but perhaps yours is more explicit and not likely to cause strange behaviour from exceptions.
I was planning on avoiding Delayed objects - if the user wants to publish those, they have to be submitted and the future used instead.

martindurant:
OK, so futures and collections (only bag tried so far) go into the scheduler, can be listed from another executor, can be fetched back. When fetched, the futures are recreated. However, the scheduler doesn't know that these futures exist in the other executor, so they remain "pending" (even if the keys are ready).

How to make the connection? I can see `Scheduler.client_wants_keys` (which should also protect the data from deletion), but I don't see how to get the scheduler to report on the current state of the futures when they are created.

mrocklin:
```python
        for key in keys:
            if self.task_state[key] in ('memory', 'erred'):
                self.report_on_key(key)
```

This will inform all clients watching a key of that key's current state.

martindurant:
OK, seems to be working.
I don't like the calls to dumps and loads, but the (bag's) keys are true tuples in the executor, and strings of those on the scheduler, and replying on the normal packing sometimes made tuples into lists.

The following message is displayed for test_publish_roundtrip:
```
Exception ignored in: <generator object add_client at 0x113df9f68>
RuntimeError: generator ignored GeneratorExit
```

distributed/scheduler.py
```diff
@@ -1565,6 +1570,30 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
...
+        keys = [str(k) for k in self.published_data[name]['keys']]
+        self.client_wants_keys(keys=keys, client=client)
+        return pickle.dumps(self.published_data[name])
+
+    def poll_keys(self, stream, keys=None):
+        keys = pickle.loads(keys)
+        for key in keys:
+            key = str(key)
+            if self.task_state[key] in ('memory', 'erred'):
+                self.report_on_key(key)
```
mrocklin: What is the reason for this function?

distributed/scheduler.py
```diff
@@ -1565,6 +1570,30 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
...
+
+    def get_published_data(self, stream, name=None, client=None):
+        keys = [str(k) for k in self.published_data[name]['keys']]
+        self.client_wants_keys(keys=keys, client=client)
+        return pickle.dumps(self.published_data[name])
+
+    def poll_keys(self, stream, keys=None):
+        keys = pickle.loads(keys)
+        for key in keys:
+            key = str(key)
```
mrocklin: See the `to_key` function

distributed/scheduler.py
```diff
@@ -1565,6 +1570,30 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
...
+    def get_published_keys(self, *args):
+        return self.published_data_keys
+
+    def get_published_data(self, stream, name=None, client=None):
+        keys = [str(k) for k in self.published_data[name]['keys']]
+        self.client_wants_keys(keys=keys, client=client)
+        return pickle.dumps(self.published_data[name])
+
+    def poll_keys(self, stream, keys=None):
+        keys = pickle.loads(keys)
```
mrocklin: It is ideal to keep `pickle` out of the scheduler.  It's a vulnerability and stops us from supporting other systems downstream.  Generally we always call `tokey` before sending anything to the scheduler.  `tokey` ensures that a key is msgpack serializable.

distributed/tests/test_executor.py
```diff
@@ -3633,3 +3633,65 @@ def f():
...
+def test_publish_simple(s, a, b):
+    e = Executor((s.ip, s.port), start=False)
+    yield e._start()
+    f = Executor((s.ip, s.port), start=False)
+    yield f._start()
+
+    data = yield e._scatter(range(3))
+    e.publish_dataset(data, 'data')
+    yield gen.sleep(0.2)
+    assert 'data' in s.published_data
```
mrocklin: Tests that depend on sleep invariably fail some times.  Search for `assert time() < start` for an idiom to manage time-dependent tests.

distributed/tests/test_executor.py
```diff
@@ -3633,3 +3633,65 @@ def f():
...
+    bag = db.from_sequence([0, 1, 2])
+    bagp = e.persist(bag)
+
+    assert len(futures_of(bagp)) == 3
+
+    e.publish_dataset(bagp, 'data')
+    yield gen.sleep(0.2)
+
+    # check that serialization didn't affect original bag's dask
+    assert len(futures_of(bagp)) == 3
```
mrocklin: Could also check that the keys agree, at least under the mapping of `tokey`.

mrocklin:
In general the `distributed` parts of this seem pretty good.  

The serialization bits concern me.  For example I believe that they will fail on dask.delayed objects.  This might require some cooperation on the dask side to do well.  I'm not sure how best to achieve this.

martindurant:
Is it not reasonable to only publish things that have already been submitted to the scheduler?

I'll look into what `tokey` does and if I can avoid pickle.

distributed/scheduler.py
martindurant: The scheduler needs to tell the executor which is pulling the dataset about the state of the futures contained. This can only happen after the executor has received the data object itself and built the futures, hence the need for two calls, one after the other.

mrocklin:
> Is it not reasonable to only publish things that have already been submitted to the scheduler?

That seems reasonable.  In case this is in response to the "I'm worried about how we serialize dask things" comment I think that it's a separate issue.  I suspect that if you tried to do what you're doing now with dask.delayed objects that were backed by Futures (much as you're doing with bag) then things would break.

distributed/scheduler.py
mrocklin: I recommend having the executor send an `'update-graph'` message with the relevant keys and an empty dask graph.  I suspect that this will work.  We should clean up the semantics around that.

Knowing what the current state of the futures is probably won't suffice.  We'll need to register the client as being interested in these particular keys so that it gets future updates as well.

mrocklin: the update-graph route will accomplish that for us.

martindurant: The line `self.client_wants_keys(keys=keys, client=client)` does this above, during the first call. I did play with update-graph too, but didn't get anywhere (yet)

mrocklin: I guess the issue here is that the Executor doesn't yet know that it cares about these keys and so treats the messages as spurious?

martindurant: That's exactly it: in the first instance, the scheduler knows about which futures are concerned and which executor will want them, but the executor hasn't had the info yet. Next, the executor gets the info and builds futures, but it needs to get an update on their statuses, which will not happen without asking, unless their statuses change.

mrocklin: I recommend having the Executor send up the `'update-graph'` message.  This is a predictable behavior that will be around for a while.  Generally it's nice to reuse the same routes when available.  We know that this will work many moons from now.  I am less confident that the `client_wants_keys` method will remain untouched.

distributed/executor.py
```diff
@@ -193,6 +193,22 @@ def release(self):
             self._cleared = True
             self.executor._dec_ref(tokey(self.key))
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.update({'event': self.event.is_set(), 'status': self.status})
+        del state['executor']
+        return state
```
mrocklin: If we can get away with it I recommend only sending along the Key, and then recreating everything else on the other side.  State like the event or status are not guaranteed to be valid on the other end.

distributed/executor.py
```diff
@@ -193,6 +193,22 @@ def release(self):
...
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        e = default_executor()
+        event = Event()
+        if state.pop('event'):
+            event.set()
+        e.futures[tokey(state['key'])] = {'event': event,
+                                          'status': state.pop('status')}
+        state['executor'] = e
```
mrocklin: This could also send the message up to the scheduler of "Hey this executor cares about this key now!"

     e._send_to_scheduler({'op': 'update-graph': 'tasks': {}, 'keys': [tokey(key)], 'client': e.id})

distributed/tests/test_executor.py
```diff
@@ -3784,3 +3786,34 @@ def test_stress_scatter_death(e, s, *workers):
...
+    yield e._start()
+    f = Executor((s.ip, s.port), start=False)
+    yield f._start()
+
+    future = e.submit(lambda: 1)
+
+    with temp_default_executor(f):
+        future2 = pickle.loads(pickle.dumps(future))
+        assert future2.executor is f
+        assert tokey(future2.key) in f.futures
```
mrocklin: Would also be good to verify that `result = yield future2._result()` finishes.

distributed/executor.py
mrocklin: They're also non-essential.  Generally a rule of thumb is that pickling is a good time to take a hard look at the intrinsic state of an object, removing everything else that is around just for efficiency's sake.

distributed/executor.py
mrocklin: This would also probably remove the need for the `poll_keys` route.

distributed/executor.py
```diff
@@ -974,6 +992,45 @@ def cancel(self, futures):
...
+        name: str
+            handle by which the stored data should be known
+
+        Returns
+        -------
+        None
+        """
+        keys = [tokey(f.key) for f in futures_of(data)]
+        self._send_to_scheduler({'op': 'publish-data', 'keys': keys,
+                                 'data': pickle.dumps(data), 'name': name})
```
mrocklin: Suggest wrapping `name` in `tokey` as well, just so that we can support tuple-names like `('account-data', 43)`

distributed/executor.py
```diff
@@ -974,6 +992,45 @@ def cancel(self, futures):
...
+        data: list of futures or dask collection
+        name: str
+            handle by which the stored data should be known
+
+        Returns
+        -------
+        None
+        """
+        keys = [tokey(f.key) for f in futures_of(data)]
+        self._send_to_scheduler({'op': 'publish-data', 'keys': keys,
```
mrocklin: Perhaps we should move away from the term `publish-data` to `publish-dataset` to avoid conceptual overlap with `scatter`

distributed/executor.py
mrocklin: Or we could do this wrapping on the scheduler side when we try to put `name` into a dictionary.  The underlying issue is that `msgpack` will turn tuples into lists, which can't be used as keys in dicts.

distributed/executor.py
mrocklin: Ah, I also recommend using `yield self.scheduler.publish_data(keys=keys, data=..., name=name)` instead.

There are two channels from the Executor to the Scheduler.  One is the `Executor.scheduler` rpc object for general, infrequent, synchronous communication.  The other is the `Executor.scheduler_stream` object for rapid high-volume communications (like 10000 calls to submit).  

In this case we probably want to return only after we've actually sent the data up and we're not in a high message volume case, so we probably want to use the non-specialized `Executor.scheduler` solution.  This will also require moving the `publish-data` route in the scheduler to the normal handlers dict

distributed/executor.py
martindurant: I was surprised to find that this works, and removes the need to call client_wants_keys too. Update-graph does more than I thought.

distributed/tests/test_executor.py
martindurant: With the changes to _send_to_scheduler, (not yet pushed), there is a failure here, f looking as if "cancelled" and not appearing in the dependencies dict - but this is timing, it appears immediately after.

mrocklin: 'cancelled' is the case when `f.key not in e.futures` .  On deserialization of the `Future` we should add an entry to `e.futures` with an empty event and status.  This normally happens in `Future.__init__`

        if tkey not in executor.futures:
            executor.futures[tkey] = {'event': Event(), 'status': 'pending'}

mrocklin: One approach here would be to just call the constructor on deserialization

martindurant: That exact line is in my `__setstate__`; is there any reason during `_result` that it might be deleted? The new future is now created with a clear Event and pending, as opposed to before, where these attributed were copied over.

distributed/executor.py
mrocklin: I recommend that you implement these as follows:

```python
def __getstate__(self):
    return self.key

def __setstate__(self, key):
    Future.__init__(self, key, default_executor())
```

martindurant: Hm, that now causes a timeout on `yield future2._result()` - future looks like it's pending even with the call to _send_to_scheduler, which I still kept in `__setstate__` after the `__init__`

mrocklin: One sec, giving this a shot locally

martindurant: Do you want me to push what I have now?

mrocklin: Sure, especially if you have a failing test.  Currently running through tests with this:

```python
    def __getstate__(self):
        return self.key 

    def __setstate__(self, key):
        e = default_executor()
        Future.__init__(self, key, e)
        e._send_to_scheduler({'op': 'update-graph',
                              'tasks': {},
                              'keys': [tokey(key)],
                              'client': e.id})
```

mrocklin: You need to send the `'client'` as `e.id` (a uuid string) rather than `e`.  If you run tests with the `-s` flag you'll get some extra debugging output, notably this error occurred:

```
distributed.core - ERROR - can't serialize <Executor: scheduler="127.0.0.1:42236">
  File "/home/mrocklin/workspace/distributed/distributed/core.py", line 274, in write
    frames = protocol.dumps(msg)
  File "/home/mrocklin/workspace/distributed/distributed/protocol.py", line 105, in dumps
    small_header, small_payload = dumps_msgpack(msg)
  File "/home/mrocklin/workspace/distributed/distributed/protocol.py", line 212, in dumps_msgpack
    payload = msgpack.dumps(msg, use_bin_type=True)
  File "/home/mrocklin/Software/anaconda/lib/python3.4/site-packages/pandas/msgpack/__init__.py", line 44, in packb
    return Packer(**kwargs).pack(o)
  File "pandas/msgpack/_packer.pyx", line 223, in pandas.msgpack._packer.Packer.pack (pandas/msgpack/_packer.cpp:3434)
  File "pandas/msgpack/_packer.pyx", line 225, in pandas.msgpack._packer.Packer.pack (pandas/msgpack/_packer.cpp:3276)
  File "pandas/msgpack/_packer.pyx", line 213, in pandas.msgpack._packer.Packer._pack (pandas/msgpack/_packer.cpp:3006)
  File "pandas/msgpack/_packer.pyx", line 184, in pandas.msgpack._packer.Packer._pack (pandas/msgpack/_packer.cpp:2433)
  File "pandas/msgpack/_packer.pyx", line 220, in pandas.msgpack._packer.Packer._pack (pandas/msgpack/_packer.cpp:3158)
TypeError: can't serialize <Executor: scheduler="127.0.0.1:42236">
```

martindurant: OK, so the issue was that the call to update-graph from the second executor was happening before a yield had ensured that the call from the first executor was actually processed in the scheduler (sot hey key didn't exist inthe state dicts).
I'm not sure that this is a case we need to worry about, as the second executor cannot have enough information to generate the future by itself, it should only be called on things that the scheduler knows about.

distributed/executor.py
```diff
@@ -974,6 +985,46 @@ def cancel(self, futures):
...
+                                 'data': pickle.dumps(data),
+                                 'name': tokey(name)})
+
+    @gen.coroutine
+    def _published_datasets(self):
+        resp = yield self.scheduler.get_published_keys()
```
mrocklin: Perhaps this should be named `published_datasets` on the scheduler as well to align terms?  

We might also consider getting rid of this method and just using `sync(self.loop, self.scheduler.published_datasets)` directly.

distributed/executor.py
mrocklin: I see two things that might help avoid errors like this:

1.  Go through the actual process that users will go through in the test.  So rather than pass the name directly force yourself to query the scheduler for the names of the published datasets
2.  As suggested in https://github.com/dask/distributed/pull/453/files#r76600071 use and yield from the `.scheduler` rpc object rather than use the `._send_to_scheduler` method.  The `._send_to_scheduler` method is only really intended to be used by the computational functions that need high-volume fire-and-forget operation.  In this case we probably want to wait/block/yield until the dataset has been properly loaded up in the scheduler.

distributed/executor.py
martindurant: Trying that completely messed me up - that's what I had called the dict storing the published data :)

mrocklin: Ah! sorry.

mrocklin:
What happens if one executor publishes a dataset, then leaves, and then another executor comes by to get that dataset?  Can there be a test for this case, whatever the desired behavior?

martindurant:
The futures will be released and forgotten, so you get a key error when trying to reference them.
I would suggest that the expected behaviour should be to error - but then we should figure out how to remove the entry in `published_datasets` too.
 The alternative would to keep extra references to data on the scheduler, and data on the workers forever, or until explicitly released.

mrocklin:
We probably shouldn't get an error.  Either of the alternatives that you suggest sound good.  The first one is probably easier.  You could track who wants each dataset by adding an extra `'clients'` key-value pair in the published datasets dictionary on the scheduler alongside keys and data.  You could then iterate through this dict on every call to `Scheduler.remove_client`

martindurant:
Looking for clients disconnecting works on keeping/deleting published data; but this can't prevent some of the futures becoming obsolete in the meantime, e.g., by explicit `del` in the executors.

mrocklin:
Hrm, you're right.  One solution to this would be to have the scheduler create a fake "client".

Perhaps when you publish a dataset you publish it forever.  That would allow you to publish a dataset, leave, and then come back and open it later.  This would require at least the following:

1.  When we publish a dataset add all of the keys into `who_wants/wants_what` with a new 'published' client.  You might have to play around a bit with `Scheduler.report` to avoid trying to send messages to the fake 'published' client.
2.  Add a `remove_dataset` route

martindurant:
who_wants will need entries for each published dataset that references it, since there may be some overlap between futures in various published datasets

mrocklin:
Perhaps we consider each published dataset a different client?  That would take care of collisions in who_wants

distributed/tests/test_executor.py
```diff
@@ -3246,6 +3247,27 @@ def test_publish_roundtrip(s, a, b):
     yield f._shutdown()
     assert 'data' not in s.published_data
 
+    assert 'published-data' not in s.who_wants[data[0].key]
+
+    e = Executor((s.ip, s.port), start=False)
```
martindurant: Oops, this has no purpose now.

distributed/executor.py
```diff
@@ -976,6 +987,66 @@ def cancel(self, futures):
         return sync(self.loop, self._cancel, futures)
 
     @gen.coroutine
+    def _publish_dataset(self, data, name):
+        keys = [tokey(f.key) for f in futures_of(data)]
+        yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
+                                             data=pickle.dumps(data),
```
mrocklin: Should probably use the `dumps` function rather than `pickle.dumps`.  It handles a few special cases.

distributed/executor.py
```diff
@@ -976,6 +987,66 @@ def cancel(self, futures):
...
+                                             client=self.id)
+
+
```
mrocklin: One line of whitespace between methods

distributed/executor.py
```diff
@@ -976,6 +987,66 @@ def cancel(self, futures):
...
+    def publish_dataset(self, data, name):
+        """
+        Store a named reference to the data in the scheduler for other executors
+
+        Parameters
+        ----------
+        data: list of futures or dask collection
+        name: str or container
+            handle by which the stored data should be known. If not a string,
+            will be wrapped by tokey() to deterministically turn into one.
```
mrocklin: Lets keep the existence of `tokey` internal and not expose it to users if we don't have to.

distributed/executor.py
```diff
@@ -976,6 +987,66 @@ def cancel(self, futures):
...
+
+        Returns
+        -------
+        None
+        """
+        sync(self.loop, self._publish_dataset, data, name)
+
+    @gen.coroutine
+    def _unpublish_dataset(self, name):
+        yield self.scheduler.unpublish_dataset(name=name)
```
mrocklin: Slight preference to remove this one-line coroutine and instead inline `self.scheduler.unpublish_dataset` wherever it gets used.

distributed/executor.py
```diff
@@ -976,6 +987,66 @@ def cancel(self, futures):
...
+        out = yield self.scheduler.get_published_dataset(name=name,
+                                                         client=self.id)
+        data, keys = out['data'], out['keys']
+
+        with temp_default_executor(self):
+            data = pickle.loads(data)
+        raise Return(data)
+
+    def get_published_dataset(self, name):
+        return sync(self.loop, self._get_published_dataset, tokey(name))
```
mrocklin: Is there redundancy in the above functions?

distributed/executor.py
martindurant: The reason was that when looking at the set of published keys, it'll look like a string. I'm happy to remove the sentence.

distributed/executor.py
martindurant: You mean the coroutine versus sync? I was following this pattern everywhere, to make them callable with yield in the test (but not used from anywhere else).
If you mean published_datasets versus get_published_dataset, the former gives the list of keys stored, the latter actually fetches one of them.

mrocklin: Ah, after looking below I see the difference.  I recommend the following names:

*  get_published_keys -> list_datasets
*  get_published_dataset -> get_dataset
*  publish_dataset

The term "key" generally means the name of a task.  It would be nice to not overload this term.  I suggest that we now marry the term "dataset" for all functions related to what you're doing now.  The term "list" might be a nice alternative to "get", when we're just getting the names of available datasets.

distributed/scheduler.py
```diff
@@ -1590,6 +1596,36 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
...
+                        client=None):
+        for k in keys:
+            self.who_wants[k].add('published-%s' % name)
+        self.published_data[name] = {'data': data, 'keys': keys,
+                                     'clients': {client}}
+
+    def unpublish_dataset(self, stream=None, name=None):
+        out = self.published_data.pop(name, {'keys': []})
+        for k in out['keys']:
+            self.who_wants[k].remove('published-%s' % name)
```
mrocklin: Whenever we manipulate `who_wants` we must also manipulate `wants_what`.  These two dicts must always be in accordance.  Same with dependencies/dependents, who_has/has_what, etc..

mrocklin: And actually I strongly recommend using the `client_releases_keys` function for all of this.  We want to centralize logic like this to avoid hard-to-find errors.

distributed/tests/test_executor.py
```diff
@@ -3317,3 +3407,96 @@ def test_scatter_compute_store_lose_processing(e, s, a, b):
...
+
+    n = 40
+    xs = [da.random.random((100, 100), chunks=(5, 5)) for i in range(n)]
+    ys = [x + x.T for x in xs]
+    z = da.atop(vsum, 'ij', *concat(zip(ys, ['ij'] * n)))
+
+    future = e.compute(z.sum())
+
+    result = yield future._result()
+    assert isinstance(result, float)
```
mrocklin: Something screwy happened when you merged with master.  These tests shouldn't be here.

distributed/executor.py
martindurant: How does that look, without a `@coroutine`, how does it get called in the test?

mrocklin: The rpc methods, like `self.scheduler.unpublish_dataset(...)` operate as coroutines.  So where-ever you used `self._unpublish_dataset(name)` you would now use `self.scheduler.unpublish_dataset(name=name)`

For example you could implement `unpublish_dataset` as follows:

```python
def unpublish_dataset(self, name):
￼    sync(self.loop, self.scheduler.unpublish_dataset, name=name)
```


￼

distributed/scheduler.py
```diff
@@ -1590,6 +1596,33 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
...
+        del data['clients']
+        return data
+
+    def check_published_datasets(self, client):
+        for key, data in self.published_data.copy().items():
+            with ignoring(KeyError):
+                data['clients'].remove(client)
+            if not(data['clients']):
+                del self.published_data[key]
+                self.client_releases_keys(data['keys'], 'published-%s' % key)
```
mrocklin: These last two lines are just `unpublish_dataset` yes?  If so then we should probably reuse that function here.  

Alternatively, if datasets no longer disappear when clients disappear then maybe we no longer need to track which clients care about which datasets and this entire function can be removed.

distributed/tests/test_executor.py
```diff
@@ -3202,6 +3204,94 @@ def f():
...
+    yield f._shutdown()
+    assert 'data' not in s.published_data
+
+    assert 'published-data' not in s.who_wants[data[0].key]
+
+
+@gen_cluster(executor=False)
+def test_unpublish(s, a, b):
+    e = Executor((s.ip, s.port), start=False)
+    yield e._start()
```
mrocklin: If you only need one Executor in a test then I recommend setting `executor=True` in the `gen_cluster` call and adding an `e` argument to `test_unpublish(e, s, a, b)`

distributed/tests/test_executor.py
```diff
@@ -3202,6 +3204,94 @@ def f():
...
+    data = yield e._scatter([0, 1, 2])
+    yield e._publish_dataset(data, 'data')
+
+    key = data[0].key
+    del data
+
+    yield e.scheduler.unpublish_dataset(name='data')
+
+    assert 'data' not in s.published_data
+    assert not s.who_wants[key]
```
mrocklin: This depends on the garbage collector running and the release-key message getting up to the scheduler before the unpublish_dataset message gets back, which may not always occur as expected.  I recommend checking this within a timed loop

```python
start = time()
while key in s.who_wants:
    yield gen.sleep(0.01)
    assert time() < start + 5
```

There are various models of doing this throughout the tests.

distributed/tests/test_executor.py
```diff
@@ -3202,6 +3204,94 @@ def f():
...
+
+    yield e._publish_dataset(bagp, 'data')
+
+    # check that serialization didn't affect original bag's dask
+    assert len(futures_of(bagp)) == 3
+
+    result = yield f._get_dataset('data')
+
+    out = yield f.compute(result)._result()
+    assert out == [0, 1, 2]
```
mrocklin: Would be nice to `assert set(out.dask) == set(bagp.dask)`

distributed/scheduler.py
```diff
@@ -1590,6 +1596,33 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
 
             return result
 
+    def publish_dataset(self, stream=None, keys=None, data=None, name=None,
+                        client=None):
+        self.client_wants_keys(keys, 'published-%s' % name)
+        self.published_data[name] = {'data': data, 'keys': keys,
+                                     'clients': {client}}
```
martindurant: Good point. Should it be disallowed (error) by the executor before even passing in the data?

mrocklin: The executor may not know about what keys are available if it hasn't synchronized with the scheduler recently.  Even if we do try to synchronize with the scheduler then we open ourselves up to race conditions.  I recommend checking this on the scheduler side, where everything is sequential.

distributed/scheduler.py
martindurant: On who_wants/wants_what, this to ensure consistency for future-proofing, rather than any material difference at this moment? 
I think I had been operating under the assumption to avoid touching the other internal structures as much as possible, but I see this is faulty. Probably with the addition of unpublish, can now stop counting executors wanting a published dataset and regard each dataset as an ordinary client (without a stream), as you suggested some time before.

distributed/tests/test_executor.py
martindurant: Doesn't the yield on unpublish ensure that that method runs in its entirety before returning control? It directly calls client_releases_keys, which edits who_wants - not by state transition or secondary action.

distributed/scheduler.py
martindurant: Then the best would be passing back a result "succeeded" - in fact, that flag could be passed back as-is to the user without an exception.

distributed/scheduler.py
mrocklin: > On who_wants/wants_what, this to ensure consistency for future-proofing, rather than any material difference at this moment? 

Various parts of the code assume that those two dicts are always in sync.  They should be considered a single data structure that is somewhat cumbersome to update.  I would expect some things to err if these get out of sync.  The fact that things aren't erring is somewhat concerning.  I should add checks here into the runtime validation methods.

distributed/tests/test_executor.py
mrocklin: Ah, sorry, let me rephrase.  Would this test still be OK without the `del data` statement?  We should not assume that `del data` occurs when we want it to.

distributed/scheduler.py
mrocklin: > Probably with the addition of unpublish, can now stop counting executors wanting a published dataset and regard each dataset as an ordinary client (without a stream), as you suggested some time before.

This does seem like a nice solution, and remove the need to think about this problem at all.

distributed/scheduler.py
mrocklin: Yes, that would be good.  I recommend the response `'OK'` or `{'status': 'OK'}` as that is used in a few other places.

You could also raise an error.  I *think* that all of the pieces are in place to propagate that error from the scheduler coroutine to the caller.  I could be wrong though.

distributed/tests/test_executor.py
martindurant: OK got it - yes, the del of data must have been processed, otherwise the key would still be referenced. I'll put in the timing thing.

distributed/scheduler.py
```diff
@@ -1598,30 +1597,21 @@ def get_nbytes(self, stream=None, keys=None, summary=True):
 
     def publish_dataset(self, stream=None, keys=None, data=None, name=None,
                         client=None):
+        if name in self.datasets:
+            return {'success': 'false'}
         self.client_wants_keys(keys, 'published-%s' % name)
-        self.published_data[name] = {'data': data, 'keys': keys,
-                                     'clients': {client}}
+        self.datasets[name] = {'data': data, 'keys': keys}
+        return {'success':  'true'}
```
mrocklin: Suggest `{'status': 'OK'}`

This is the common response for other routes within the system.

distributed/tests/test_executor.py
```diff
@@ -3213,9 +3213,12 @@ def test_publish_simple(s, a, b):
     yield f._start()
 
     data = yield e._scatter(range(3))
-    yield e._publish_dataset(data, 'data')
-    assert 'data' in s.published_data
-    assert e.id in s.published_data['data']['clients']
+    out = yield e._publish_dataset(data, 'data')
+    assert out['success'] == 'true'
```
mrocklin: I suggest that, on the client side this return None if successful and raise a `KeyError(name)` otherwise.

mrocklin: The status messages are useful internally but are probably not something that the user wants to see

distributed/tests/test_executor.py
```diff
@@ -3241,19 +3244,12 @@ def test_publish_roundtrip(s, a, b):
...
 
-    yield e._shutdown()
-    assert e.id not in s.published_data['data']['clients']
-
-    yield f._shutdown()
-    assert 'data' not in s.published_data
-
-    assert 'published-data' not in s.who_wants[data[0].key]
+    result = yield f._get_dataset(name='nonexistent')
+    assert result is None
```
mrocklin: Recommend raising a `KeyError` in this case

distributed/tests/test_executor.py
```diff
@@ -3287,6 +3289,8 @@ def test_publish_bag(s, a, b):
     assert len(futures_of(bagp)) == 3
 
     result = yield f._get_dataset('data')
+    assert set(result.dask.keys()) == set(bagp.dask.keys())
+    assert [f.key for f in result.dask.values()] == [f.key for f in bagp.dask.values()]
```
mrocklin: Recommend using curly braces here.  The output order of iterating over dictionaries is not guaranteed.

mrocklin:
Playing around with this a bit on my own machine this generally feels pretty good.

Some small rough edges:

At first I did the following:

    e.publish_dataset('x', my_dataset)

Which ends up being the wrong order.  My intuition didn't match.  Thoughts on which order is preferred?  Also thoughts on supporting keyword arguments like the following?:

    e.publish_dataset(x=my_dataset, y=dataset_2)

When I publish a dask collection:

```python
df = dd.demo.make_timeseries('2000', '2001', {'value': float, 'name': str, 'id': int}, freq='2H', partition_freq='1D', seed=1)
e.publish_dataset(df, 'df')
```

The scheduler isn't aware of any keys within the collection.  This is because there are no futures within this object.  Perhaps this is cleanest, however an alternative would be to persist all collections beforehand.  This can get tricky though because now we're trying to dive through some list of things to find dask collections.

mrocklin:
Should we warn if there are no keys present in what the user sends up perhaps?

mrocklin:
cc @hussainsultan you've asked for things like this before.  Would you mind giving this a spin and sharing your thoughts?

martindurant:
So a summary of my thoughts:
- I think it's fine to store dask objects which don't have futures - they get de/serializes fine, and can be useful as a data prescription rather than existing data (and the keys they generate upon evaluation may or may not be on the workers). The documentation needs to be very clear here.
- however, we do need some check on the input data; probably if you got the inputs the wrong way around, you end up storing a string, with a dataset name that is a string version of the data you wanted to store... So, inputs should be list of futures or dask objects, not arbitrary data
- I don't have a string preference on which way around the inputs to publish should be, but of course the way I wrote it reflects what I thought was natural. The syntax of kwargs might get around this, but again, this is just style preferences and I don't think anyone will be confused.

mrocklin:
OK, I'm going to go ahead and merge this.  I'll make another pull request with some documentation and some tests for informative error raising.  I may switch to using kwargs in publish to start, mostly because it's easier to add args later if we need to, but harder to take them away.

PR #458, opened 2016-08-24
Re-implement work-stealing

### LINKED ISSUE #456: Thoughts on Work Stealing
I've been thinking about work stealing recently.  I thought I'd put down some thoughts.

### Situation

Recall that the scheduler sends a worker about as many tasks as it has cores.  Excess tasks that are known to be particularly well suited to the worker, either because it has most of the necessary data or because of a user-specified restriction, are kept in the `stacks`, a list associated with each worker.  Additionally, there is a common pool of tasks from which any worker may draw.

The need for work stealing arises when both of the following conditions hold:

1.  Some workers are *saturated*, meaning that those workers are busy processing as many tasks as they have cores, and they also have some excess tasks in their stack.
2.  Some workers are *idle*, meaning that they are not processing sufficient tasks, and so necessarily their stack and the common ready pool are both empty.

In these cases we may choose to pull tasks from the stack of a *saturated* worker onto the stack of an *idle* worker.  The idle worker *steals* from the saturated worker.

Recall though that in some sense this is sub-optimal; there there was an original reason why the stolen task was assigned to the saturated worker.  We're going to end up moving excess data or violating a loose user-provided restriction in order to achieve load balancing.

### Objectives

We want to find "good" tasks to steal among the stacks of saturated workers.  A task can be good in a few ways:

1.  **Computation/communication time ratio**:  The ratio of its computation time to its data transfer time is high.  We would like tasks that take minutes to compute who only have a few bytes of data that can be transferred in a millisecond.
2.  **Steal from the back of the stack**: Workers put preferred tasks at the front of their stacks, leaving stale unwanted tasks at the back.  This practice aids low-memory-footprint operation.  The tasks at the front of the stack are fresh for the worker, and something related to the worker's current computations.  The tasks at the back of the stack are of lower priority for that worker, and so are more polite to steal.
3.  **Steal from the rich**.  Some workers have longer stacks than others.  We prefer to steal from these overburdened workers.

### Performance

So we could formulate an objective function and search over the stacks for tasks that satisfy this objective function.  However, the set of tasks in the stacks may be quite large (in the millions) and we may want to steal frequently.  As with all scheduling operations we would like to keep the administrative costs of work-stealing low, preferably even as the number of tasks increases.  This constant-time restriction stops us from considering traditional sorting techniques, like keeping around a heap.  This is especially important because work stealing often never comes up.  In the common case we would like to have low cost.

### Three approaches

Lets consider a few approaches that optimize for different things.  None of these will be ideal but they may bring us closer to a solution.

### Keep a sorted collection of time ratios

We want to steal tasks that are expensive to compute but cheap to move.  So as we move any task into the stacks we compute this ratio (using estimated costs) and put it into a heap.  If the task leaves the stacks then we remove it from the heap.  Each operation takes `log(n)` time for `n` the number of tasks in the stacks.  When we want to steal we just take the top element of the heap, a fast operation.

This solution optimizes computation/communication time ratio.  It does not select tasks low in the worker's stack, nor does it steal from the rich.  However, it doesn't do a particularly bad job at these either.  All things equal we're likely to pick a task from the middle of a stack and we're more likely to steal from the rich than from the poor.  In practice work stealing often arises in heavily imbalanced situations, where a small number of workers hold all of the pending tasks.

### Keep a pair of sets

In order to avoid the logarithmic time cost of the heap, we instead keep around two sets, "good to steal" and "not good to steal", and put tasks into each bin according to whether or not their time ratio is over a certain threshold.  This is faster, but results in less optimal task selection.  Like the sorted collection it doesn't take from the bottom of the stack nor does it steal from the rich.

We can extend this approach to "Keeping an ordered collection of k sets", say one for each order of magnitude of time ratio.  We can still add and remove tasks at constant time but now have more granularity about how good a task is to steal.  This also lets us respond to desperate times by stealing worse tasks when the task distribution is heavily lopsided.  This solution starts to look similar to non-comparative integer sorting techniques like Radix sort.

### Pull from the end

To optimize for "steal from the back" and "steal from the rich" we might find the worker with the largest stack (at a linear search through the number of workers, typically less than 1000) and then take the task from the end of that worker's stack.  This does not optimize for the time ratio but we could add in a quick check to only steal if it's reasonable and just not steal otherwise.

mrocklin:
This maintains a list of sets of tasks to steal, ranked by
compute/communication ratios.

Fixes #456 

Doc page: https://github.com/mrocklin/distributed/blob/work-steal-2/docs/source/work-stealing.rst

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4,6 +4,7 @@ from collections import defaultdict, deque, OrderedDict
 from datetime import datetime, timedelta
 import logging
 import math
+from math import log
 import os
 import pickle
 import random
@@ -137,7 +138,8 @@ class Scheduler(Server):
         Which clients want each key.  The active targets of computation.
     * **nbytes:** ``{key: int}``:
         Number of bytes for a key as reported by workers holding that key.
-
+    * **stealable:** ``[[key]]``
+        A list of stacks of stealable keys, ordered by stealability
     * **ncores:** ``{worker: int}``:
         Number of cores owned by each worker
     * **idle:** ``{worker}``:
@@ -221,6 +223,8 @@ class Scheduler(Server):
         self.exceptions = dict()
         self.tracebacks = dict()
         self.exceptions_blame = dict()
+        self.stealable = [set() for i in range(12)]
+        self.stealable_unknown_durations = defaultdict(set)
 
         # Worker state
         self.ncores = dict()
@@ -1649,6 +1653,7 @@ class Scheduler(Server):
                 else:
                     self.stacks[new_worker].append(key)
                     self.maybe_idle.add(new_worker)
+                    self.put_key_in_stealable(key)
                     self.task_state[key] = 'stacks'
             else:
                 self.ready.appendleft(key)
@@ -1678,6 +1683,7 @@ class Scheduler(Server):
             self.rprocessing[key].add(worker)
             self.occupancy[worker] += duration
             self.task_state[key] = 'processing'
+            self.remove_key_from_stealable(key)
 
             logger.debug("Send job to worker: %s, %s", worker, key)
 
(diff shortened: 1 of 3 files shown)


mrocklin:
cc @akcheung , this might interest you

PR #459, opened 2016-08-25
DOC: adds note to start at least one worker

stsievert:
Added a note in quickstart.rst that at least one worker had to be started.


mrocklin:
Merged.   Thanks!

PR #461, opened 2016-08-26
Connect to ipython kernel in scheduler

mrocklin:
This starts an IPython kernel within the scheduler and adds a
convenience function to the Executor to start and connect to it.

I've tried to copy the worker tests faithfully but this is behaving oddly for me.

### DIFF
--- a/distributed/_ipython_utils.py
+++ b/distributed/_ipython_utils.py
@@ -91,7 +91,7 @@ def register_worker_magic(connection_info, magic_name='worker'):
 
 def connect_qtconsole(connection_info, name=None, extra_args=None):
     """Open a QtConsole connected to a worker who has the given future
-    
+
     - identify worker with who_has
     - start IPython kernel on the worker
     - start qtconsole connected to the kernel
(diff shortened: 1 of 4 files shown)


mrocklin:
@minrk when you have a free moment can I ask you to take a look at this?  I'm also getting strange behavior when I use this locally with the qtconsole.  

No rush.  This work is cleanly isolated away from everything else that is happening.

minrk:
@mrocklin I just opened #464, which I had 90% done at SciPy, but couldn't figure out something in the tests until today (at EuroSciPy), where running any *two* IPython tests would fail, but each would individually succeed.

I suspect it is what you are referring to, and seems to be fixed by the `zmq_ctx` fixture I added in #464.

I also cleaned up a few things that I found from the first version, so it might make sense to base this one on #464.

minrk:
Also given this PR, I moved `start_ipython` to a function in `_ipython_utils`, so you should be able to import that in the scheduler rather than copying it.

mrocklin:
Closing in favor of #464

PR #464, opened 2016-08-27
add %remote magic, Executor.start_ipython_scheduler

minrk:
As discussed at SciPy, but it took me forever to figure out why a test wasn't passing.
registered when Executor.start_ipython is called during an IPython session.

%remote loads connection info from a dict in the IPython namespace, e.g.

    info = e.start_ipython(worker)[worker]
    %remote info print(worker.data)

### DIFF
--- a/distributed/_ipython_utils.py
+++ b/distributed/_ipython_utils.py
@@ -155,7 +155,7 @@ def register_remote_magic(magic_name='remote'):
 
 def connect_qtconsole(connection_info, name=None, extra_args=None):
     """Open a QtConsole connected to a worker who has the given future
-    
+
     - identify worker with who_has
     - start IPython kernel on the worker
     - start qtconsole connected to the kernel
--- a/distributed/_ipython_utils.py
+++ b/distributed/_ipython_utils.py
@@ -15,9 +15,12 @@ except ImportError:
     import Queue as queue
 from subprocess import Popen
 import sys
+from threading import Thread
 from uuid import uuid4
 
 from tornado.gen import TimeoutError
+from tornado.ioloop import IOLoop
+from threading import Event
 
 from IPython import get_ipython
 from jupyter_client import BlockingKernelClient, write_connection_file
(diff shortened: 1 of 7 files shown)


minrk:
@mrocklin since I'm sprinting at EuroSciPy, I went ahead and folded #461 into this. I can revert that, if you would prefer. Now you've got:

1. `%remote` magic for running code on a remote kernel (worker or scheduler) based on an info dict
2. `Executor.start_ipython_scheduler` which registers the `%scheduler` magic

Question:

The way you set `magic_name='scheduler'` by default means that the *default* behavior of `start_ipython_scheduler` is to register the scheduler magic. This might be the cause of the failures you were seeing in tests, because this will fail if IPython isn't running. Options I see:

`start_ipython_scheduler()` registers its magic:

1. only if `magic_name` is given by the user
1. if nothing is given *and* IPython is running
1. always unless `(magic_name=None)` (current behavior, perhaps not intended?)

All are doable, just need to pick.

minrk:
Tests are failing because mock isn't available in the hdfs context. I don't fully understand how to get it there.

mrocklin:
Sorry for the delay here.  Spending much of this weekend away from the computer.  

Option two sounds like a convenient choice to me at the moment.  There is less ambiguity here about what the magic should be named than in the worker case.  

The HDFS environment is handled here:

* https://github.com/dask/distributed/blob/master/continuous_integration/Dockerfile#L13
* https://github.com/dask/distributed/blob/master/continuous_integration/Dockerfile#L20

Thank you for sprinting on this during EuroScipy by the way.  The magics have been super useful for debugging.  I used to do a number of tricks in order to get remote workers and schedulers that were easy to inspect.  This solution is way easier and folds into normal deployment schemes.

minrk:
Gotcha, thanks. It now does option 2 and tests are passing.

distributed/_ipython_utils.py
```diff
@@ -114,3 +178,72 @@ def _cleanup_connection_file():
...
+        app.kernel.pre_handler_hook = noop
+        app.kernel.post_handler_hook = noop
+        app.kernel.start()
+        app.kernel.loop = IOLoop.instance()
+        # save self in the IPython namespace as 'worker'
+        # inject things into the IPython namespace
+        if ns:
+            app.kernel.shell.user_ns.update(ns)
+        from tornado.ioloop import PeriodicCallback
+        pc = PeriodicCallback(lambda : print(os.getpid(), file=sys.__stdout__), 1000, io_loop=zmq_loop)
```
mrocklin: Why this callback?

mrocklin:
The `%remote` and `%scheduler` magic both work quite well locally.  This feels really slick.

The qtconsole is unhappy on my machine, a Jupyter QTConsole window shows up, but it's just a white box.  This happens on master though as well, so it's quite possibly just an issue with my local environment.

distributed/_ipython_utils.py
minrk: super-professional print-statement debugging that I neglected to clean up. Gone now.

minrk:
@mrocklin I think that's a bug in recent qtconsole connecting to running kernels. Try pressing `enter` and see if you get a prompt.

mrocklin:
Yup indeed.  Now I feel a little foolish :)

mrocklin:
This looks good to me.  Merging.

Thanks @minrk!

PR #466, opened 2016-08-28
New progress bars

birdsarah:
@mrocklin - weekend surprise - awesome progress bars.

screen shot 2016-08-27 at 10 06 43 pm

The claims:
- These are made by your very own custom model - `ProgressBar`
- They are real html elements and will all be the same size growing the page as they need to.
- They are colored so they form a legend for the other two plots
- The legending means that I've cleaned up the memory plot.
- I also cleaned up the hover styling slightly.
- Much less data processing and less data going to the front end (see distributed/diagnostics/progress_stream.py)

The caveats:
- This makes a whole bunch of new dom elements. Having tried it out, it doesn't seem that performance is a problem. But if it is, we could optimize this.
- I also haven't done the thing where the progress bar shows the unallocated memory (or whatever the light blue bar thing is) because I wasn't entirely sure about that. But we can certainly iterate.
- I also haven't done anything about displaying errored counts - but of course we can.

### DIFF
--- a/distributed/bokeh/progress_bar.py
+++ b/distributed/bokeh/progress_bar.py
@@ -25,8 +25,6 @@ class ProgressBarView extends Widget.View
             done = data.done[i]
             all = data.all[i]
             percent = parseInt(done/all * 100)
-
-            # Can use bar_class to set other colors
             if percent == 100
                 bar_class = 'bar finished'
             else
@@ -39,7 +37,7 @@ class ProgressBarView extends Widget.View
                         <span class='count'>#{done}/#{all}</span>
                     </div>
                     <div class='meter'>
-                        <span class='#{bar_class}' style='width: #{percent}%'></span>
+                        <span class='#{bar_class}' style='width: #{percent}%; background-color: #{data.color[i]}'></span>
                     </div>
                 </li>
             ")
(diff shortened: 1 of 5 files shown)


mrocklin:
Hooray weekend surprises!  Very cool.  Some comments:

### Memory plot

Very cool.  Much cleaner.  No concerns here so moving on.

### Progress bars

At first I was a bit concerned about using color in the progress bars, thinking that it would be a bit too much.  The effect here is much nicer than I expected though, +1.

I'm somewhat concerned about space.  On my screen I'm running low on vertical space.  The new bars are 2-3 times taller than before so I end up scrolling down to see what's going on.  The number of bars can easily increase to the dozens (or hundreds in a community setting, but I'm not sure we're going to ever find a way to satisfy this.)  In the previous solution I ended up resolving this by:

1.  Putting the text within the bar
2.  Minimizing vertical padding
3.  Having the bars stack more horizontally based on how many bars there were
4.  Sorting the bars by number of total tasks

I'm not yet seeing the hover.  Should I be on a recent dev build of Bokeh?

How established are the new custom models?

mrocklin:
Here is another computation that will generate a few more bars.

```python
from distributed import Executor, wait
e = Executor('127.0.0.1:8786')
import dask.dataframe as dd

df = dd.demo.make_timeseries('2000', '2001',
        {'value': float, 'name': str, 'id': int},
        freq='10s', partition_freq='1D', seed=1)

df = df[(df.value > 0) & (df.value < 1)]

a = df.groupby(df.name).id.mean()
b = df.groupby(df.name).value.mean()

aa, bb = e.persist([a, b])
wait([aa, bb])

from time import sleep
sleep(10)
```

mrocklin:
Probably going to stick to pure python bokeh for a while.  Closing.

PR #470, opened 2016-08-30
Add IPython integration docs

mrocklin:
cc @minrk

### DIFF
--- a/docs/source/ipython.rst
+++ b/docs/source/ipython.rst
@@ -3,10 +3,14 @@ IPython Integration
 
 Dask.distributed integrates with IPython in three ways:
 
-1.  You can launch a Dask.distributed cluster from an IPyParallel cluster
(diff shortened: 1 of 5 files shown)


docs/source/ipython.rst
```diff
@@ -0,0 +1,84 @@
...
+This is possible with the Client.become_distributed_ method::
+
+   $ ipcluster start
+
+.. code-block:: python
+
+   >>> from ipyparallel import Client
+   >>> c = Client()  # connect to IPyParallel cluster
+
+   >>> e = c.become_distributed()  # start dask on top of IPyParallel
```
minrk: While we are here, do you have a preference for the name of this method? Should I have called it `.become_dask`?

mrocklin: Yes, that would probably be nicer.

PR #471, opened 2016-08-30
Launch tasks from other tasks

mrocklin:
This allows the spawning of tasks from other tasks.  We don't add any new scheduling functionality. Instead we make it more convenient to start a client from a task.  

```python
In [1]: from distributed import Executor, local_executor

In [2]: e = Executor()

In [3]: def fib(n):
   ...:     if n < 2:
   ...:         return n
   ...:     else:
   ...:         with local_executor() as ee:
   ...:             a, b = ee.submit(fib, n - 1), ee.submit(fib, n - 2)
   ...:             a, b = ee.gather([a, b])
   ...:             return a + b
   ...:         

In [4]: future = e.submit(fib, 100)

In [5]: future
Out[5]: <Future: status: finished, type: int, key: fib-7890e9f06d5f4e0a8fc7ec5c77590ace>

In [6]: future.result()
Out[6]: 354224848179261915075
```

See associated doc page for more information.

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -12,6 +12,7 @@ from .nanny import Nanny
 from .scheduler import Scheduler
 from .utils import sync
 from .worker import Worker
+from .worker_executor import local_executor
 
 try:
     from .collections import futures_to_collection
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -205,8 +205,8 @@ class Server(TCPServer):
                 try:
                     handler = self.handlers[op]
                 except KeyError:
-                    result = 'No handler found: ' + op.encode()
-                    logger.warn(result)
+                    result = "No handler found: %s" % op
+                    logger.warn(result, exc_info=True)
                 else:
                     logger.debug("Calling into handler %s", handler.__name__)
                     try:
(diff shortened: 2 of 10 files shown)


mrocklin:
Supercedes #471

mrocklin:
There are two ways in which this can deadlock.  

1.  All cores in the cluster are waiting on these tasks
2.  All threads in an individual worker are waiting on these tasks while additional tasks are already in the thread-pool-executor's work queue.

The first case is unfortunate, but somewhat controllable and unlikely.  

The second case can happen more easily and is harder to control for.  Because we're not using coroutines/generator functions we don't have an easy way to jump out of the task and free up the thread.  I think our only option here is to add another thread to the ThreadPoolExecutor.  This can be done dynamically by changing private variables however the standard `concurrent.futures.ThreadPoolExecutor` can't scale back down, so adding more space like this is a somewhat irreversible operation.  That is unless of course we want to reimplement the ThreadPoolExecutor, which is fairly doable.

mrocklin:
I've resolved the more serious of the deadlocks by allowing tasks that call `get_executor()` to leave the common thread pool, making space for additional work.  This required forking the `concurrent.futures.ThreadPoolExecutor`.

mrocklin:
OK, now scatter dumps data directly to the local worker's data dict, bypassing the scheduler.

Additionally I've removed the first deadlock by temporarily increasing the number of slots on any worker that is running a client.

mrocklin:
I'm now pretty happy with this.  Merging on passed tests

Issue #463, opened 2016-08-26
Configuration File


mrocklin:
Continuation of https://github.com/dask/distributed/issues/58

I think it's now time to have a configuration file.  There are a few options that may be nicer to manage on a per-machine basis rather than in various command line options (though these will remain dominant) and hard coded settings.

Here are a few:

1.  Logging levels for dask
2.  Logging levels for the bokeh web application
3.  Compression 
4.  Ports for the scheduler, json, web interface, etc..
5.  Whitelisted ports for bokeh (though this is now open by default)
6.  Whether or not to use PDB when an error occurs (I use this for debugging)

Some open questions:

1.  Where do we put this file?  I'm thinking `~/.dask/config`
2.  What format do we use, JSON, YAML, TOML, INI?
3.  Are there other options that people find themselves often setting that we would want to include?  We could also just include *all* options available through the CLI
4.  Desired nesting level?  For example

    ```python
    'scheduler': {'port': 8786,
                  'bokeh': 8787}, 
    ...
    ```
    vs 
    ```python
    'scheduler-port': 8786,
    'scheduler-bokeh': 8787,
    ...
    ```

@quasiben I would value your feedback in particular here.

I don't have much scar tissue on this topic.

mrocklin:
Currently going ahead with YAML.  So far I'm only putting in options that I use personally.  Planning to wait until people need more to add more:

```yaml
logging:
  distributed: info
  distributed.executor: warning
  bokeh: critical

compression: auto

# Scheduler specific options

bandwidth: 100000000    # 100 MB/s estimated worker-worker bandwidth
allowed-failures: 3     # number of retries before a task is considered bad
pdb-on-err: False       # enter debug mode on scheduling error
transition-log-length: 100000
```

quasiben:
*All formats are bad*

With that said, I've found yaml for config files to be not as a bad as other.  YAML is:

- human editable/readable
- supports comments

YAML is a bit better with some extra tooling like:

- https://yaml.readthedocs.io/en/latest/overview.html -- comment preservation and round trip support
- http://schematics.readthedocs.io/en/latest/ -- schema/type validation

mrocklin:
Implemented in https://github.com/dask/distributed/pull/472

minrk:
Personally, I like having nesting for grouping related config:

```yaml
scheduler:
   port: 123
```

I don't know what the scope of your configurability would be, though.

[...]

PR #472, opened 2016-08-31
Add config.yaml file

mrocklin:
This puts a new config.yaml file in `~/.dask/config.yaml` on import if it
is not present.  These options gets loaded into a module-level global,
``distributed.config`` and are used in various places where constants
used to be hard coded.  So far I've only added support for constants
that I change frequently.

At the moment the config file looks like the following:

```yaml
logging:
  distributed: info
  distributed.executor: warning
  bokeh: critical

compression: auto

 # Scheduler specific options

bandwidth: 100000000    # 100 MB/s estimated worker-worker bandwidth
allowed-failures: 3     # number of retries before a task is considered bad
pdb-on-err: False       # enter debug mode on scheduling error
transition-log-length: 100000
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,5 +6,6 @@ include README.rst
 include LICENSE.txt
 include MANIFEST.in
 include requirements.txt
+include distributed/config.yaml
 
 prune docs/_build
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import print_function, division, absolute_import
 
+from .config import config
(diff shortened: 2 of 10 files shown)


mrocklin:
Fixes #463

quasiben:
Are you concerned at all with invalid data?  A fine answer would be, "not right now"

mrocklin:
Should I be?  Presumably things would err in a sensible way?

quasiben:
I suppose things err in a nice way.  If users have troubles I think we could easily fix

PR #473, opened 2016-08-31
support complex keys in rebalance and replicate

mrocklin:
http://stackoverflow.com/questions/39254182/replicate-a-dataset-with-dask-to-all-workers

cc @hussainsultan


hussainsultan:
@mrocklin wonderful! thank you!

PR #474, opened 2016-09-01
Datasets 2

mrocklin:
Follows on https://github.com/dask/distributed/pull/453

cc @martindurant 

1.  Uses keyword arguments to name datasets
2.  Raises errors in scheduler functions, depending on these propagating down to the executor
3.  Tests errors for informative phrases
4.  Adds documentation page
5.  Tweaks docstrings to align well with sphinx documentation

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -987,58 +987,89 @@ class Executor(object):
         return sync(self.loop, self._cancel, futures)
 
     @gen.coroutine
-    def _publish_dataset(self, data, name):
-        keys = [tokey(f.key) for f in futures_of(data)]
-        out = yield self.scheduler.publish_dataset(keys=keys, name=tokey(name),
-                                             data=dumps(data),
-                                             client=self.id)
-        if out['status'] != "OK":
-            raise KeyError(name)
+    def _publish_dataset(self, **kwargs):
+        coroutines = []
+        for name, data in kwargs.items():
+            keys = [tokey(f.key) for f in futures_of(data)]
+            coroutines.append(self.scheduler.publish_dataset(keys=keys,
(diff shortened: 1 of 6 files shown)


distributed/executor.py
```diff
@@ -987,58 +987,101 @@ def cancel(self, futures):
...
-        if out['status'] != "OK":
-            raise KeyError(name)
-
-    def publish_dataset(self, data, name):
+    def _publish_dataset(self, **kwargs):
+        coroutines = []
+        for name, data in kwargs.items():
+            keys = [tokey(f.key) for f in futures_of(data)]
+            coroutines.append(self.scheduler.publish_dataset(keys=keys,
+                name=tokey(name), data=dumps(data), client=self.id))
```
martindurant: Do we want to protect against arbitrary sending of large objects, or something that is completely un-dask?

mrocklin: I don't know a nice way to check for this without reaching down into the protocol level.  Thoughts?

martindurant: How about only allow two cases: is a sequence where type of every item is future, or has .dask attribute. Easy to spoof, of course.

mrocklin: In general I like starting out with tightly scoped things like this.  In this case there are some things, like lists or dictionaries of `dask.delayed` objects that would not be covered here.  I'm a little tempted by the simplicity of "we just shove anything in there and its up to you not to misuse it."

mrocklin:
I'm planning to remove some dead code and rename Executor -> Client soon.  For that I'd like to have everything merged in to avoid conflicts.  I'm happy to continue work on this in another PR.  Merging this for now though.

PR #484, opened 2016-09-02
Add workers= keyword to e.persist

mrocklin:
Fixes https://github.com/dask/distributed/issues/285

cc @martindurant @nfaggian if you can provide feedback on the plan proposed in [the associated doc page](https://github.com/mrocklin/distributed/blob/persist-restrictions/docs/source/locality.rst#specify-workers-with-computepersist) that would be helpful.

### DIFF
--- a/distributed/executor.py
+++ b/distributed/executor.py
@@ -1237,18 +1237,28 @@ class Executor(object):
         results2 = pack_data(keys, results)
         return results2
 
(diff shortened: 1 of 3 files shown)


nfaggian:
LGTM. I would make immediate use of this work.

mrocklin:
Well then lets give this a shot.  We can iterate a bit if things aren't right.

nfaggian:
Great. I will give this a go today at work.

Issue #475, opened 2016-09-01
Dump excess data to disk


mrocklin:
One way for Dask to be robust to running out of memory would be to dump data to disk.  There are a few problems within this problem:

1.   How do we identify when we are running low on memory?  We can use `psutil` and watch the memory use of this process or of the entire machine.  We can also be given a budget and track the nbytes of our keys and dump when we overflow
2.  How do we choose which keys to dump?  We want keys that are infrequently accessed and expensive to keep around.  
3.  How do we serialize data for efficient dumping to disk?  Pickle?  Compression?  Perhaps the `dumps` function we already use for network communications would be appropriate here.

mrocklin:
There is a `class Buffer(MutableMapping)` in [zict](https://github.com/mrocklin/zict) that might be useful for creating a two-level in-memory/on-disk store with LRU semantics weighted by number of bytes.

PR #485, opened 2016-09-03
Add Spill-to-disk policy for workers

mrocklin:
This enacts a few changes:

First, workers can spill excess data to disk.  They use the core.dumps system
for writing data (usually just pickle) and then compress with the default
compression.  This uses ``zict.Buffer``, which uses an LRU mechanism to dump
excess data to a directory of files.

By default this is off.  You can turn it on by setting `--memory-limit=auto` in
dask-worker or `--memory-limit=1e9` or whatever to specify that it should spill
after using up one gigabyte.

Currently this uses `sizeof` to determine the amount of RAM that data is taking
up.  This is error prone with custom objects but fine as long as most things
are numpy arrays or pandas dataframes or whatnot.

Excess data is kept in the worker directory, which can now grow to be very
large.  We appropriately are now much more careful about cleaning this up when
workers close down.

We pass more information to the diagnostics so that disk-loads show up in the
task stream plot.  They show up as a slightly different shade of red than
worker-worker data transfers.

Fixes https://github.com/dask/distributed/issues/475

I'll write up a doc page for this soon.


mrocklin:
cc @hussainsultan

hussainsultan:
Thanks. I will check test this out on my end!

Issue #477, opened 2016-09-01
LocalCluster() fails to start Diag Server


m-ric:
A simple:
```
import distributed
c = LocalCluster()
```
fails to start diagserver with following error. Running `LocalCluster(diagnostics_port=None)` does not show the error.
```
In [1]: import distributed

In [2]: c = distributed.LocalCluster()
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-9953140d5917> in <module>()
----> 1 c = distributed.LocalCluster()

/home/johndoe/.local/lib/python2.7/site-packages/distributed/deploy/local.pyc in __init__(self, n_workers, threads_per_worker, nanny, loop, start, scheduler_port, silence_logs, diagnostics_port, services, **kwargs)
     99         if diagnostics_port is not None:
    100             self.start_diagnostics_server(diagnostics_port,
--> 101                                           silence=silence_logs)
    102 
    103     def __str__(self):

/home/johndoe/.local/lib/python2.7/site-packages/distributed/deploy/local.pyc in start_diagnostics_server(self, port, show, silence)
    179                 http_port=self.scheduler.services['http'].port,
    180                 bokeh_port=port, show=show,
--> 181                 log_level=logging.getLevelName(silence).lower())
    182 
    183     @gen.coroutine

/home/johndoe/.local/lib/python2.7/site-packages/distributed/bokeh/application.pyc in __init__(self, host, http_port, tcp_port, bokeh_port, bokeh_whitelist, log_level, show, prefix, use_xheaders, quiet)
     81             import subprocess
     82             self.process = subprocess.Popen(args, stdout=subprocess.PIPE,
---> 83                     stderr=subprocess.PIPE)
     84 
     85         if not quiet:

/usr/lib/python2.7/subprocess.pyc in __init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags)
    708                                 p2cread, p2cwrite,
    709                                 c2pread, c2pwrite,
--> 710                                 errread, errwrite)
    711         except Exception:
    712             # Preserve original exception in case os.close raises.

/usr/lib/python2.7/subprocess.pyc in _execute_child(self, args, executable, preexec_fn, close_fds, cwd, env, universal_newlines, startupinfo, creationflags, shell, to_close, p2cread, p2cwrite, c2pread, c2pwrite, errread, errwrite)
   1325                         raise
[...]

PR #488, opened 2016-09-06
This fixes issue dask/distributed#477

dsidi:
Getting the dirname from argv[0] sometimes fails. For example, if running unit tests in PyCharm utrunner is called from a directory that may not contain bokeh

`/home/dsidi/miniconda/envs/FOO/bin/python /home/dsidi/.local/lib/pycharm-2016.2/helpers/pycharm/utrunner.py path_to_test_file.py ...`


mrocklin:
Seems fine to me.  Merging!  Thanks @dsidi

PR #490, opened 2016-09-06
Import joblib classes from public module

lesteve:
for ParallelBackendBase and AutoBatchingMixin.

Both classes have been exposed in joblib.parallel in joblib/joblib@1eb46ca which is in joblib 0.10.2.

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -15,11 +15,11 @@ from .utils import ignoring
 joblib = sk_joblib = None
 with ignoring(ImportError):
     import joblib
-    if LooseVersion(joblib.__version__) < '0.10.0':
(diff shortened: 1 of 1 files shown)


mrocklin:
In that same file a bit higher up we check the version number of joblib and eventually raise an error if the version number is too low.  We should probably bump that version to 0.10.2.

mrocklin:
We should do the same for the `sklearn` version number.

lesteve:
Good point I missed that !

lesteve:
Done, the scikit-learn version is still correct, since joblib 0.10.2 will be bundled with scikit-learn 0.18.

mrocklin:
Great.  Merging on passed tests

PR #493, opened 2016-09-06
Ensure bokeh application shuts down

jcrist:
Due to reference cycles `Executor.__del__` won't be run on cleanup. When
using a `LocalCluster` and not run as a contextmanager, this method was
responsible for shutting down the bokeh diagnostics application. To
remedy this, the bokeh application will now shut itself down upon
connection failure to the scheduler.

### DIFF
--- a/distributed/bokeh/tasks/server_lifecycle.py
+++ b/distributed/bokeh/tasks/server_lifecycle.py
@@ -3,8 +3,10 @@
 from __future__ import print_function, division, absolute_import
 
 from collections import deque
+import sys
 import json
 import os
+import logging
 from time import time
 
 from tornado import gen
@@ -16,7 +18,9 @@ from distributed.core import read
 from distributed.diagnostics.eventstream import eventstream
 from distributed.bokeh.status_monitor import task_stream_append
 import distributed.bokeh
-from distributed.utils import log_errors
+
+
+logger = logging.getLogger(__name__)
 
 client = AsyncHTTPClient()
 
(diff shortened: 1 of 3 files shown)


distributed/bokeh/tasks/server_lifecycle.py
```diff
@@ -32,32 +35,34 @@
...
+                    index.append(i)
+                    i += 1
+                    if msg.get('transfer_start') is not None:
+                        index.append(i)
+                        i += 1
+                    if msg.get('disk_load_start') is not None:
+                        index.append(i)
+                        i += 1
+                    task_stream_append(rectangles, msg, workers)
+    except Exception as e:
```
mrocklin: In the case of a `StreamClosedError` we probably don't want to exit without logging the error.

mrocklin: Err, we probably *want* to exit without logging the error

distributed/bokeh/tests/test_application.py
```diff
@@ -29,3 +29,15 @@ def test_BokehWebInterface(loop):
...
+def test_bokeh_shutsdown_without_cluster___del__(loop):
+    c = LocalCluster(2, loop=loop, scheduler_port=0,
+                     services={('http', 0): HTTPScheduler},
+                     diagnostic_port=None)
+    proc = c.diagnostics.process
+    # don't run the del, as it isn't ever run in python < 3.5 due to cycles
+    c.__del__ = lambda self: None
+    del c
+    sleep(0.1)
+    assert proc.poll() is not None
```
mrocklin: In general trusting `sleep`s on travis.ci is tough.  Instead I recommend the following pattern:

```python
start = time()
while proc.poll() is None:
    sleep(0.01)
    assert time() < start + 5
```

jcrist:
Addressed all comments.

mrocklin:
Merged

PR #500, opened 2016-09-08
Remove first message from 'progress' message stream

mrocklin:
We now just avoid computing anything if we haven't yet received a message

cc @hussainsultan


hussainsultan:
Thanks. I am going to try this out.

PR #505, opened 2016-09-09
WIP: worker environments

jcrist:
Todo:
- [x] `Client.register_environment` updates state on scheduler and workers
- [x] environment state updated when workers are added
- [x] environment state updated when workers are removed
- [x] tests for managing environment state
- [x] scheduler takes environments into account when assigning tasks
- [x] `Client.submit` can take environment names in `workers=`
- [ ] Tests for scheduling with environments

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -35,6 +35,7 @@ from .utils_comm import WrappedKey, unpack_remotedata, pack_data
 from .compatibility import Queue as pyQueue, Empty, isqueue
 from .core import (read, write, connect, coerce_to_rpc, dumps,
         clean_exception, loads)
+from .environments import Environment
 from .worker import dumps_function, dumps_task
 from .utils import (All, sync, funcname, ignoring, queue_to_iterator,
         tokey, log_errors, str_graph)
@@ -1089,6 +1090,19 @@ class Client(object):
         Client.list_datasets
         """
         return sync(self.loop, self._get_dataset, tokey(name))
+
+    @gen.coroutine
+    def _register_environments(self, environments):
+        envs = {}
+        for name, env in environments.items():
+            if not isinstance(env, Environment):
+                env = Environment(condition=env)
+            envs[name] = dumps(env)
+        yield self.scheduler.register_environments(environments=envs)
+
+    def register_environments(self, **kwargs):
+        """Register environments with the scheduler."""
+        return sync(self.loop, self._register_environments, kwargs)
 
     @gen.coroutine
     def _run(self, function, *args, **kwargs):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1092,17 +1092,26 @@ class Client(object):
         return sync(self.loop, self._get_dataset, tokey(name))
 
     @gen.coroutine
-    def _register_environments(self, environments):
-        envs = {}
-        for name, env in environments.items():
-            if not isinstance(env, Environment):
-                env = Environment(condition=env)
-            envs[name] = dumps(env)
-        yield self.scheduler.register_environments(environments=envs)
+    def _register_environment(self, name, env):
+        if not isinstance(env, Environment):
+            env = Environment(condition=env)
+        envs = {name: dumps(env)}
+        exceptions = yield self.scheduler.register_environments(environments=envs)
+        if exceptions:
+            raise loads(exceptions[name][0])  # only raise the first exception
 
-    def register_environments(self, **kwargs):
-        """Register environments with the scheduler."""
-        return sync(self.loop, self._register_environments, kwargs)
+    def register_environment(self, name, env):
+        """Register an environment with the scheduler.
+
+        Parameters
+        ----------
+        name : str
+            The name of the environment.
+        env : Environment or callable
+            The environment to register. If a callable, will be used as the
+            ``condition`` method of the environment.
+        """
+        return sync(self.loop, self._register_environment, name, env)
 
     @gen.coroutine
     def _run(self, function, *args, **kwargs):
--- a/distributed/environments.py
+++ b/distributed/environments.py
@@ -1,0 +1,22 @@
+from __future__ import absolute_import, division, print_function
+
+
+class Environment(object):
+    def __init__(self, condition=None, setup=None):
+        self._condition = condition
+        self._setup = setup
+
+    def condition(self):
+        if self._condition:
+            return self._condition()
+        return True
+
+    def setup(self):
+        if self._setup:
+            self._setup()
+
+    def __getstate__(self):
+        return (self._condition, self._setup)
+
+    def __setstate__(self, state):
+        self._condition, self._setup = state
--- a/distributed/environments.py
+++ b/distributed/environments.py
@@ -14,9 +14,3 @@ class Environment(object):
     def setup(self):
         if self._setup:
             self._setup()
-
-    def __getstate__(self):
-        return (self._condition, self._setup)
-
-    def __setstate__(self, state):
-        self._condition, self._setup = state
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1654,19 +1654,28 @@ class Scheduler(Server):
             for name in environments:
                 if name in self.environments:
                     raise KeyError("Environment %s already exists" % name)
-            for name, env in environments.items():
-                self.environments[name] = env
-                self.environment_workers[name] = set()
 
         if workers is None:
             workers = list(self.worker_info)
 
-        responses = yield [self.rpc(addr=worker).register_environments(environments=environments)
-                           for worker in workers]
+        results = yield [self.rpc(addr=worker).register_environments(environments=environments)
+                         for worker in workers]
 
-        for worker, added in zip(workers, responses):
-            for name in added:
-                self.environment_workers[name].add(worker)
+        exceptions = {}
+        for worker, result in zip(workers, results):
+            for name, val in result.items():
+                if isinstance(val, bool):
+                    if name not in self.environments:
+                        self.environments[name] = environments[name]
+                        self.environment_workers[name] = set()
+                    if val:
+                        self.environment_workers[name].add(worker)
+                else:
+                    if name not in exceptions:
+                        exceptions[name] = [val]
+                    else:
+                        exceptions[name].append(val)
+        raise Return(exceptions)
 
     def change_worker_cores(self, stream=None, worker=None, diff=0):
         """ Add or remove cores from a worker
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1197,9 +1197,11 @@ class Scheduler(Server):
         """
         if not self.ncores:
             raise ValueError("No workers yet found.")
+        workers2 = list(self.workers_set(workers))
         if workers is not None:
-            workers = [self.coerce_address(w) for w in workers]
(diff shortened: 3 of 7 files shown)


distributed/scheduler.py
```diff
@@ -1267,20 +1280,9 @@ def restart(self, environment=None):
...
-                workers = list(self.ncores)
-            else:
-                workers = []
-        if hosts is not None:
-            for host in hosts:
-                if host in self.host_info:
-                    workers.extend([host + ':' + port
-                            for port in self.host_info[host]['ports']])
-        # TODO replace with worker_list
+        workers = self.workers_set(workers)
```
mrocklin: Hooray!

distributed/tests/test_scheduler.py
```diff
@@ -872,6 +872,29 @@ def test_coerce_address():
     yield [w._close() for w in [a, b, c]]
 
 
+@gen_test()
+def test_workers_set():
+    s = Scheduler(validate=True)
+    s.start(0)
```
mrocklin: Recommend 

```python
@gen_cluster(ncores=[])
def test_workers_set(s):
    a = ...
```

distributed/tests/test_scheduler.py
```diff
@@ -872,6 +872,29 @@ def test_coerce_address():
...
+    a = Worker(s.ip, s.port, name='alice', ip='127.0.0.1')
+    b = Worker(s.ip, s.port, name=123, ip='127.0.0.2')
+    c = Worker(s.ip, s.port, name='charlie', ip='127.0.0.2')
+    yield [a._start(), b._start(), c._start()]
+    # Manually setup environment state, just for this test
+    s.environments['test'] = b'not a real pickle string'
+    s.environment_workers['test'] = set([a.address, c.address])
```
mrocklin: FWIW I used to do unit-testy things like this and eventually gave it up and just used clients all the time.  The internal state changes more quickly than the API and I found myself editing tests very often.  I'm not against this code by the way, it's great and arguably the "right" way to do things, just sharing my experience.  I now use clients all the time

distributed/tests/test_client.py
```diff
@@ -3340,6 +3341,82 @@ def test_publish_bag(s, a, b):
...
+    assert len(s.environment_workers['never_passes']) == 0
+
+    # Remove worker
+    yield w._close()
+    start = time()
+    while not w.status == 'closed':
+        yield gen.sleep(0.01)
+        assert time() - start < 5
+
+    assert w.address not in s.environment_workers['myenv']
```
mrocklin: Nice test!

distributed/environments.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def setup(self):
+        if self._setup:
+            self._setup()
+
+    def __getstate__(self):
+        return (self._condition, self._setup)
+
+    def __setstate__(self, state):
+        self._condition, self._setup = state
```
mrocklin: Were these serialization methods necessary?

distributed/scheduler.py
```diff
@@ -1635,6 +1637,37 @@ def get_dataset(self, stream, name=None, client=None):
...
             raise KeyError("Dataset '%s' not found" % name)
 
+    @gen.coroutine
+    def register_environments(self, stream=None, environments=None, workers=None):
+        if environments is None:
+            environments = self.environments
+        else:
+            for name in environments:
+                if name in self.environments:
+                    raise KeyError("Environment %s already exists" % name)
```
mrocklin: Perhaps only err if the value is different?

distributed/scheduler.py
```diff
@@ -1635,6 +1637,37 @@ def get_dataset(self, stream, name=None, client=None):
...
+
+        results = yield [self.rpc(addr=worker).register_environments(environments=environments)
+                         for worker in workers]
+
+        exceptions = {}
+        for worker, result in zip(workers, results):
+            for name, val in result.items():
+                if isinstance(val, bool):
+                    if name not in self.environments:
+                        self.environments[name] = environments[name]
```
mrocklin: Perhaps do `self.environments.upate(environments)` early on?  This looks complex-ish.

For example, if we our Scheduler has no workers when this is run with then we don't hold onto the environment for future workers.

mrocklin: Could also do the `self.environment_workers[name] = set()` bit early on as well

distributed/scheduler.py
```diff
@@ -236,6 +236,10 @@ def __init__(self, center=None, loop=None,
         self.saturated = set()
         self.occupancy = dict()
 
+        # Environment state
+        self.environments = {}
+        self.environment_workers = {}
```
mrocklin: Can I ask that we also include environment_workers in `self.identity()`

mrocklin: I can imagine wanting this information in diagnostics

distributed/scheduler.py
```diff
@@ -1635,6 +1637,37 @@ def get_dataset(self, stream, name=None, client=None):
...
+                        self.environments[name] = environments[name]
+                        self.environment_workers[name] = set()
+                    if val:
+                        self.environment_workers[name].add(worker)
+                else:
+                    if name not in exceptions:
+                        exceptions[name] = [val]
+                    else:
+                        exceptions[name].append(val)
+        raise Return(exceptions)
```
mrocklin: We might want to return environment counts back to the client here.  I expect people will want feedback on how many gpus machines they have and such or if their condition was ever satisfied.

distributed/tests/test_scheduler.py
jcrist: Fixed, as well as the test I copied this from.

distributed/scheduler.py
jcrist: Fixed.

distributed/scheduler.py
jcrist: Ooh, nice catch. Fixed.

distributed/scheduler.py
jcrist: That's doable. Since these can change as workers are added, perhaps making `environment_workers` accessible on the client might be nice as well (would probably call it `environments` and make it a property that returns a dictionary).

distributed/scheduler.py
jcrist: Fixed

distributed/client.py
```diff
@@ -1091,6 +1092,28 @@ def get_dataset(self, name):
         return sync(self.loop, self._get_dataset, tokey(name))
 
     @gen.coroutine
+    def _register_environment(self, name, env):
+        if not isinstance(env, Environment):
+            env = Environment(condition=env)
```
mrocklin: we could also do this with keyword arguments, like the following:

```python
def register_environment(self, name, environment=None, condition=None):
    ...
```

distributed/environments.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import, division, print_function
+
+
+class Environment(object):
+    def __init__(self, condition=None, setup=None):
+        self._condition = condition
+        self._setup = setup
```
mrocklin: Perhaps default to the following:

```python
self.condition = condition or lambda: True
self.setup = setup or lambda: None
```

and then do away with the method definitions?

distributed/scheduler.py
```diff
@@ -336,7 +341,8 @@ def identity(self, stream):
              'id': str(self.id),
              'workers': list(self.ncores),
              'services': {key: v.port for (key, v) in self.services.items()},
-             'workers': dict(self.worker_info)}
+             'workers': dict(self.worker_info),
+             'environments': valmap(set, self.environment_workers)}
```
mrocklin: Msgpack may not like the sets.  Recommend lists instead.

distributed/environments.py
jcrist: Then you can't subclass without overriding `__init__`.

distributed/scheduler.py
```diff
@@ -502,17 +508,28 @@ def add_worker(self, stream=None, address=None, keys=(), ncores=None,
             if self.ncores[address] > len(self.processing[address]):
                 self.idle.add(address)
 
-            for key in list(self.unrunnable):
-                r = self.restrictions.get(key, [])
-                if address in r or host in r or name in r:
-                    self.transitions({key: 'released'})
```
mrocklin: Why was this removed?

mrocklin: Ah, I see that you've pulled it out below.  Will take a look.  

Generally some history here is that for a while I pulled everything out into little small functions like this.  All of the indirection soon became very hard to follow.  I tried to keep everything fairly shallow after the refactor.

mrocklin: There are obviously pros and cons here.

jcrist: It's called in 2 places, so I figured it could be its own tiny method.

mrocklin: Yeah, I can't think of a cleaner way.  In that case I suppose I'm just bringing it up to bring it up.  Shallowness is a design goal of sorts when convenient.

distributed/scheduler.py
```diff
@@ -711,6 +727,9 @@ def remove_worker(self, stream=None, address=None):
             if address in self.saturated:
                 self.saturated.remove(address)
 
+            for worker_sets in self.environment_workers.values():
+                worker_sets.discard(address)
```
mrocklin: Can you add a test inside of `Scheduler.validate` to check that all workers in `self.environment_workers` are also in`self.worker_info`.  Don't bother with `scheduler.py:validate()` (which is the prominent function in Scheduler.validate)

distributed/scheduler.py
```diff
@@ -1635,6 +1644,38 @@ def get_dataset(self, stream, name=None, client=None):
         else:
             raise KeyError("Dataset '%s' not found" % name)
 
+    @gen.coroutine
+    def register_environments(self, stream=None, environments=None, workers=None):
+        if environments is None:
+            environments = self.environments
```
mrocklin: What is the importance of this case?

distributed/scheduler.py
```diff
@@ -2651,23 +2697,32 @@ def coerce_address(self, addr):
...
+                try:
+                    w = self.coerce_address(w)
+                except Exception:
+                    continue
+                if ':' in w:
+                    out.add(w)
+                else:
+                    out.update('%s:%s' % (w, port)
+                               for port in self.host_info[w].get('ports', ()))
+        return out
```
mrocklin: Yes, I can see how the hostname environment-name ambiguity will start to hurt here

distributed/scheduler.py
```diff
@@ -2685,7 +2740,7 @@ def start_ipython(self, stream=None):
 
 
 def decide_worker(dependencies, stacks, processing, who_has, has_what, restrictions,
-                  loose_restrictions, nbytes, key):
+                  loose, nbytes, key):
```
mrocklin: This seems like a good move.  I would expect a lot of `test_scheduler.py` tests to get angry with you though :)

mrocklin:
Considering names for a moment.  I'm colliding a bit here with the recent conda work that allows the client to upload conda environments for all of the workers to restart themselves in.  This feature is unpublished and so can be easily renamed, but the current name is `upload_environment`.  We'll need to change one or both terms.

Thoughts on names for either of the two concepts currently called environments?

mrocklin:

PR #507, opened 2016-09-10
DOC: adds info about OMP_NUM_THREADS to FAQ

stsievert:
Added a short section about the bug about `OMP_NUM_THREADS` to the FAQ. I mention the solution and point to [the related SO question] for more detail.

[the related SO question]:http://stackoverflow.com/questions/39422092/error-with-omp-num-threads-when-using-dask-distributed


mrocklin:
Hrm, part of my programming brain is that likes for solutions to always be in one place is firing off little warning signals right now.  I like the idea of calling out important StackOverflow quesitons and answers here but I'm wary of repeating their contents.  I'd rather focus edits on the StackOverflow question itself and have a very small link here to it.  Something like:

*  For problems with OMP_NUM_THREADS, see <LINK>

stsievert:
> likes for solutions to always be in one place

:+1: I wasn't sure how you wanted to handle this repeated information, but I like pointing to SO. That also shows that the docs aren't the only place to find info.

Issue #290, opened 2016-05-25
Longitudinal Benchmarks with asv


mrocklin:
In conversation [here](https://github.com/dask/distributed/pull/278#issuecomment-221188100) it was suggested by @minrk and @ogrisel that we set up something like https://github.com/spacetelescope/asv to monitor performance of a few important computation patterns over time.  This would help us to identify performance regressions as we change the scheduler.

mrocklin:
I already have a bunch of benchmarks that I use when profiling performance.  The goal of this issue should be first to set up sufficient infrastructure so that it is easy for others to add benchmarks in the future.

dhirschfeld:
In case there's desire to have a real-world test case I'm posting below a very simple Monte-Carlo GBM simulation. I've got similar code running on a distributed cluster so have a vested-interest in making sure it runs efficiently! I'm hopeful that a real-world use-case such as this can inform future optimisation efforts (though I wonder if it's too simple to be of any use?)


<details>
<summary>
<b><tt>multivariate_normal(covariance, ndates, nsims, nchunks)</tt></b>
</summary>

```python
def multivariate_normal(covariance, ndates, nsims, nchunks):
    """Returns correlated random normal variables according to the supplied ``covariance``
    
    Parameters
    ----------
    covariance : ndarray[float](ndims, ndims)
        The size of the covariance is determined by ``ndims`` - the number
        of processes being simulated
    
    ndates : int
        The size of the time axis for the simulation
    
    nsims : int
        The number of simulations for each time period
    
    nchunks : Tuple[int, int, int]
        The size of each chunk
    
    Returns
    -------
    rvs : ndarray[float](ndims, ndates, nsims)
        The correlated rvs st ``np.cov(rvs.reshape(ndims, -1)) ~= covariance``
    
    """
    ndims = covariance.shape[0]
    U, s, Vh = np.linalg.svd(covariance)
    rvs = da.random.normal(size=(ndims, ndates, nsims), chunks=(ndims, 1+ndates//nchunks, nsims))
    rvs -= rvs.mean(axis=-1)[..., None]
    rvs /= rvs.std(axis=-1)[..., None]
    rvs = da.dot(U @ np.diag(np.sqrt(s)), rvs.reshape(2, -1)).reshape(rvs.shape)
    return rvs
```
</details>


<details>
<summary>
<b><tt>simulate_gbm(F, sigma, Te, rvs)</tt></b>
</summary>

```python
[...]

PR #524, opened 2016-09-14
Tune scheduler performance

mrocklin:
I put the scheduler through a small dask.array benchmark and ran into a
few places where performance could be improved:

1.  Remove log_errors in decide_worker
2.  Keep reverse dictionary for stealable keys
3.  Remove logger.debug lines in tight stimulus/transition functions
4.  Remove Scheduler.ready.remove in transition_ready_released
5.  Small optimizations on key_split

### DIFF
--- a/distributed/diagnostics/progress.py
+++ b/distributed/diagnostics/progress.py
@@ -273,7 +273,10 @@ class AllProgress(SchedulerPlugin):
             self.all[k].remove(key)
             if not self.all[k]:
                 del self.all[k]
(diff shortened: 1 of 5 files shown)


minrk:
Have you thought about setting up [asv](http://asv.readthedocs.io/en/latest/) to track changes like this?

mrocklin:
Yes.  Actually you recommended it earlier :)

 https://github.com/dask/distributed/issues/290

Issue #499, opened 2016-09-08
Can not connect to the scheduler with large number of workers


hussainsultan:
I set up a cluster with 25 nodes x32 cores with workers running in threaded mode. The scheduler logs show the following error for each of the workers.

```
Sep 08 18:42:23 conda[15999]: yield
Sep 08 18:42:23 conda[15999]: File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/distributed/bokeh/status/main.py", line 62, in progress_update
Sep 08 18:42:23[15999]: nb = nbytes_bar(msg['nbytes'])
Sep 08 18:42:23 [15999]: KeyError: 'nbytes'
Sep 08 18:42:23 [15999]: distributed.utils - ERROR - 'nbytes'
Sep 08 18:42:23  conda[15999]: Traceback (most recent call last):
Sep 08 18:42:23conda[15999]: File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/distributed/utils.py", line 207, in log_errors
Sep 08 18:42:23  conda[15999]: yield
Sep 08 18:42:23  conda[15999]: File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/distributed/bokeh/status/main.py", line 62, in progress_update
```

Since this, i also cannot connect to the Scheduler

```
e =  Executor('127.0.0.1:8786', set_as_default=True)
```

```
distributed.utils - ERROR - Could not connect to 127.0.0.1:8786
Traceback (most recent call last):
  File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/distributed/core.py", line 303, in connect
    stream = yield gen.with_timeout(timedelta(seconds=timeout), future)
  File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/tornado/gen.py", line 1015, in run
    value = future.result()
  File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/tornado/concurrent.py", line 237, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 3, in raise_exc_info
tornado.gen.TimeoutError: Timeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/root/.miniconda/envs/dask_env/lib/python3.5/site-packages/distributed/executor.py", line 371, in ensure_connected
    stream = yield connect(self.scheduler.ip, self.scheduler.port)
[...]

PR #526, opened 2016-09-15
Use fraction of hard connection limit

mrocklin:
Previously the scheduler would limit the number of ad-hoc connections to
512 by default.  Now we change the soft limit of the process to use up
to half of the available open files and we set the ad-hoc connection
limit to half of that.  On linux by default this is 65k / 4.

Maybe Fixes https://github.com/dask/distributed/issues/499

cc @hussainsultan

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -5,6 +5,7 @@ import json
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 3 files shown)


hussainsultan:
I tested that this has been fixed. Thank you! I still couldn't get the `%scheduler` magic work thought:(

PR #532, opened 2016-09-16
DOC: clarifies setting up network in quick start

stsievert:
I ran into an issue today where someone was trying to setup the hard way because it was on the top of the page.

After he started using the easy way, he asked "why would I use the hard way?". I added a line explaining the use case.


mrocklin:
Seems like a good change to me

PR #534, opened 2016-09-16
DOC: faq mentions nested job submission

stsievert:
I've fielded a question from two other users about nested job submission -- they ask if it's possible to submit jobs within jobs that are already submitted.

I would clarify the question, but look at the docs I changed -- they say what I'm trying to say here.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -15,6 +15,15 @@ for larger modules.
 __ http://stackoverflow.com/questions/39295200/can-i-use-functions-imported-from-py-files-in-dask-distributed
 .. _API docs: http://distributed.readthedocs.io/en/latest/api.html#distributed.executor.Executor.upload_file
 
+Can I launch a tasks from a task?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This could be while averaging inside another function submitted with
+dask.distributed. For more detail and explanation, see
+http://distributed.readthedocs.io/en/latest/task-launch.html#submit-tasks-from-worker
+
+
+
 Too many open file descriptors?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


mrocklin:
I think I'd rather keep this style of Question and Answer documentation on StackOverflow.  

I recommend asking our colleague to ask a question on StackOverflow with the #Dask tag.  I subscribe to this tag and am fairly timely if the question is well worded.

mrocklin:
Also, you may want to take a look at http://distributed.readthedocs.io/en/latest/task-launch.html#submit-tasks-from-worker

mrocklin:
Or you could also ask-and-answer a question on stackoverflow if you preferred to provide the answer yourself

stsievert:
Thanks for that.

I tried writing a SO question to answer, but I realized I would just point to the docs again (which include an answer and good technical explanation). I've revised the commit to only include the question and point to the docs.

docs/source/faq.rst
```diff
@@ -15,6 +15,15 @@ for larger modules.
 __ http://stackoverflow.com/questions/39295200/can-i-use-functions-imported-from-py-files-in-dask-distributed
 .. _API docs: http://distributed.readthedocs.io/en/latest/api.html#distributed.executor.Executor.upload_file
 
+Can I launch a tasks from a task?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This could be while averaging inside another function submitted with
+dask.distributed. For more detail and explanation, see
```
mrocklin: Citing one example application seems odd here.  It could be used for many things.  I still think we should just rely on the Table of Contents for this.  Is there a way to improve the main webpage so that people interested in topics like this will the relevant page more easily?

stsievert: Maybe swapping the order of "build understanding" and "additional features" so "additional features" is on top? This looks like it makes sense for what those two sections cover too.

And maybe include a link in the FAQ pointing to "additional features"?

mrocklin:
Closing as stale

PR #535, opened 2016-09-16
Scheduler can retire workers cleanly

mrocklin:
This allows the scheduler to retire workers cleanly by replicating data
on doomed workers to other ongoing machines.


mrocklin:
Superceded by #536

PR #536, opened 2016-09-16
Adaptive Deployments

mrocklin:
This adds an interface to adaptively deploy dask workers based on scheduler load.  

Full doc page is here: https://github.com/mrocklin/distributed/blob/worker-nbytes/docs/source/adaptive.rst

### DIFF
--- a/distributed/deploy/__init__.py
+++ b/distributed/deploy/__init__.py
@@ -1,5 +1,6 @@
 from ..utils import ignoring
 
 from .local import LocalCluster
+from .adaptive import Adaptive
 with ignoring(ImportError):
     from .ssh import SSHCluster
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -1,0 +1,109 @@
+from __future__ import print_function, division, absolute_import
+
+import logging
+from ..utils import log_errors
+
+from tornado import gen
+from tornado.ioloop import PeriodicCallback
+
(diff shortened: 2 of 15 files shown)


mrocklin:
@hussainsultan @kszucs this could use your feedback.  In particular the docpage linked to from the top and the example Marathon implementation included at the bottom of that docpage

kszucs:
It looks good to me. Health-checks and port definitions would be nice though. 

Is there any advantage if using --nprocs option for colocated workers?

```py
if 'cpus' in kwargs:
    args.extend(['--nprocs', str(int(kwargs['cpus']))])
```

mrocklin:
@kszucs how would you recommend integrating docker containers into this solution?

Submitting the scheduler is tricky.  We use information on the scheduler to determine how many workers to create.  It's convenient to have it locally.

Looking into health checks and ports now

kszucs:
The interface of marathon client is pretty ugly, so:

```python
MarathonApp(
    cmd='...', 
    container=MarathonContainer(
        docker=MarathonDockerContainer(
            image='dask/distributed', 
            force_pull_image=True)))
```

It translates to https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-apps

You need to enable docker containerizer in mesos_slaves like You [did](https://github.com/mrocklin/dask-marathon/blob/master/docker-compose.yml#L69) :)

PR #537, opened 2016-09-18
Add ssl configurations for Server and rpc connections

hussainsultan:
Add minimal SSL configuration options for `dask.distributed` cluster.  

For production deployments, this PR allows for setting up different TLS certificate for every dask worker node in your cluster, scheduler node and client node. It assumes that all certificates are signed by a single "root/cluster" CA authority.  This does not include verification of certificates  when presented.

This PR does not add SSL for Bokeh Interface.  Needs docs.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -22,10 +22,10 @@ from tornado.locks import Event
 from tornado.tcpserver import TCPServer
 from tornado.tcpclient import TCPClient
 from tornado.ioloop import IOLoop
-from tornado.iostream import IOStream, StreamClosedError
+from tornado.iostream import IOStream, StreamClosedError, SSLIOStream
 
 from .compatibility import PY3, unicode, WINDOWS
-from .utils import get_traceback, truncate_exception, ignoring
+from .utils import get_traceback, truncate_exception, ignoring, create_ssl_context
 from . import protocol
 
 pickle_types = [str, bytes]
@@ -132,12 +132,12 @@ class Server(TCPServer):
     default_port = 0
 
     def __init__(self, handlers, max_buffer_size=MAX_BUFFER_SIZE,
-            connection_limit=512, **kwargs):
+            connection_limit=512, ssl_context_creator= create_ssl_context, **kwargs):
         self.handlers = assoc(handlers, 'identity', self.identity)
         self.id = str(uuid.uuid1())
         self._port = None
         self.rpc = ConnectionPool(limit=connection_limit)
-        super(Server, self).__init__(max_buffer_size=max_buffer_size, **kwargs)
+        super(Server, self).__init__(max_buffer_size=max_buffer_size, ssl_options=ssl_context_creator(), **kwargs)
 
     @property
     def port(self):
@@ -299,7 +299,7 @@ def connect(ip, port, timeout=3):
     client = TCPClient()
     start = time()
     while True:
-        future = client.connect(ip, port, max_buffer_size=MAX_BUFFER_SIZE)
+        future = client.connect(ip, port, max_buffer_size=MAX_BUFFER_SIZE, ssl_options=create_ssl_context())
         try:
             stream = yield gen.with_timeout(timedelta(seconds=timeout), future)
             stream.set_nodelay(True)
@@ -608,7 +608,7 @@ def coerce_to_rpc(o, **kwargs):
     if isinstance(o, (bytes, str, tuple, list)):
         ip, port = coerce_to_address(o, out=tuple)
         return rpc(ip=ip, port=int(port), **kwargs)
-    elif isinstance(o, IOStream):
+    elif isinstance(o, IOStream) or isinstance(o, SSLIOStream):
         return rpc(stream=o, **kwargs)
     elif isinstance(o, rpc):
         return o
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -46,6 +46,11 @@ def invalid_python_script(tmpdir_factory):
     local_file.write("a+1")
     return local_file
 
+@pytest.fixture(scope='session')
+def certs(tmpdir_factory):
+    local_file = tmpdir_factory.mktemp('data').join('test.cert')
+    local_file.write("test")
+    return local_file
 
 @pytest.yield_fixture
 def current_loop():
--- a/distributed/config.yaml
+++ b/distributed/config.yaml
@@ -11,3 +11,7 @@ allowed-failures: 3     # number of retries before a task is considered bad
 pdb-on-err: False       # enter debug mode on scheduling error
 transition-log-length: 100000 
 
+# SSL Options
+ssl-enabled: False
+ssl-key-file:
+ssl-cert-file:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -22,7 +22,7 @@ from tornado import gen
 from tornado.gen import Return
 from tornado.queues import Queue
 from tornado.ioloop import IOLoop, PeriodicCallback
-from tornado.iostream import StreamClosedError, IOStream
+from tornado.iostream import StreamClosedError, IOStream, SSLIOStream
 
 from dask.compatibility import PY3, unicode
 from dask.core import reverse_dict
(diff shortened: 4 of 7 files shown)


hussainsultan:
currently Travis is failing on `test_stress_creation_and_deletion` . Looking into if this PR changed anything.

distributed/core.py
```diff
@@ -132,12 +132,12 @@ class Server(TCPServer):
...
 
     def __init__(self, handlers, max_buffer_size=MAX_BUFFER_SIZE,
-            connection_limit=512, **kwargs):
+            connection_limit=512, ssl_context_creator= create_ssl_context, **kwargs):
         self.handlers = assoc(handlers, 'identity', self.identity)
         self.id = str(uuid.uuid1())
         self._port = None
         self.rpc = ConnectionPool(limit=connection_limit)
-        super(Server, self).__init__(max_buffer_size=max_buffer_size, **kwargs)
+        super(Server, self).__init__(max_buffer_size=max_buffer_size, ssl_options=ssl_context_creator(), **kwargs)
```
mrocklin: This indirection seems unnecessary.  If possible I'd like to stick as close as possible to the Tornado API here.  I don't want to invent and maintain new API unless it's very convenient.  Is this very convenient?

Instead of taking in a thing that creates a set of options when called perhaps we can instead just take in the set of options?  If your goal is to default to config/environment variables then perhaps we can use a value like `ssl_options=True`?

```python
def __init__(..., ssl_options=False, ...):
    
    if ssl_options is True:
        ssl_options = ... default_options ...
    ...
    super(Server, self).__init__(..., ssl_options=ssl_options)
```

distributed/core.py
```diff
@@ -608,7 +608,7 @@ def coerce_to_rpc(o, **kwargs):
     if isinstance(o, (bytes, str, tuple, list)):
         ip, port = coerce_to_address(o, out=tuple)
         return rpc(ip=ip, port=int(port), **kwargs)
-    elif isinstance(o, IOStream):
+    elif isinstance(o, IOStream) or isinstance(o, SSLIOStream):
```
mrocklin: `isinstance(o, (IOStream, SSLIOStream))`

mrocklin: Although actually it looks like `SSLIOStream` inherits from `IOStream`, so this should not be necessary.

```python
In [2]: SSLIOStream.mro()
Out[2]: 
[tornado.iostream.SSLIOStream,
 tornado.iostream.IOStream,
 tornado.iostream.BaseIOStream,
 object]
```

distributed/utils_test.py
```diff
@@ -46,6 +46,11 @@ def invalid_python_script(tmpdir_factory):
     local_file.write("a+1")
     return local_file
 
+@pytest.fixture(scope='session')
+def certs(tmpdir_factory):
+    local_file = tmpdir_factory.mktemp('data').join('test.cert')
+    local_file.write("test")
+    return local_file
```
mrocklin: This appears to only be used once.  Perhaps just include it directly in that one test instead of creating the extra fixture?

mrocklin:
What is the process to use SSL with dask.distributed today without any modifications?  It looks like I would need to pass `ssl_options` to all of the Scheduler/Worker calls and this keyword argument would trickle down to `tornado.TCPServer`.  Is this correct?

distributed/core.py
hussainsultan: ahh. makes sense. I will take these out.

distributed/utils_test.py
hussainsultan: cool.

hussainsultan:
Yes. Thats correct.

mrocklin:
In that case I wonder if we could handle all of this by just putting logic in the `dask-scheduler` and `dask-worker` command line utilities.  Alternatively, if there is extra logic that needs to be added maybe try adding it to Tornado?  I would be happy if the core of dask.distributed could pass by the question of SSL entirely.  If we can get away without having to think about SSL while maintaining the project I would be happy.

distributed/core.py
hussainsultan: dependency injection for `ssl_context_creator` felt ok to me since it makes it clear that configurations are separated out. Perhaps, i haven't fully understood your concern regarding maintaining different API. Wouldn't `ssl_option=False` in `__init__` a deviation already?

I anticipate that `create_ssl_options` method to evolve into less trivial case and would like to keep it separate from the `Server` object.

mrocklin: It could continue remaining separate from the Server object if we stick with the Tornado default of passing in `ssl_options` rather than a function that gives us these options.  It would just be on the caller of `Server` to call the function rather than sending the function-to-be-called to `Server`.

Alternatively, what is the benefit to passing around an options-creator rather than the options themselves?  What do we gain by delaying the function call?

mrocklin: This is a deviation from Tornado.  I don't yet see the reason motivating the deviation.  I don't have much experience with SSL though, so I don't know if, for example, you have to figure out ssl_options after something has already run in the `Server.__init__` code or something.

hussainsultan:
i toyed around with the idea of `dask-scheduler`, and `dask-worker` as the way to configure it. It is certainly possible however, this means configuration for `rpc` object to connect with `ssl_options` along with other subclasses of `Server` . I found Server with dependency injection the best way to accomplish it but agree with keeping `distributed.core` simple. Let me play around with this some more because its certainly worth it to refactor it that way.

mrocklin:
What needs to change in `rpc`?

What subclasses would be necessary?  Are we able to pass down `Scheduler(..., ssl_options=...)` and `Worker(..., ssl_options=...)` where we call `Scheduler` and `Worker` currently?

hussainsultan:
rpc calls connect() method that will  need to be aware of SSL

mrocklin:
Are the `ssl_options` in `TCPClient.connect` the exact same options that will be passed to `TCPServer.__init__`?

mrocklin:
I should go read up on SSL

hussainsultan:
`ssl_options` are the same for `TCPClient.connect`.  If dask is deployed in a multi node setting then for each worker the certs could be configured to be different based on the global configs and that lets us create per worker certs which is more secure. I may have  a need for Client to have different ssl_options though but thats for another day.

distributed/core.py
hussainsultan: ahh. i see. I dont have to configure it after some `Server` state. The reason it is a method was mostly because it lets me test the server object with a trivial method.

would this be better:
```python
def __init__( ...):


    ssl_options = ssl_context_creator() 
    ...
    super(Server, self).__init__(..., ssl_options=ssl_options)
```

mrocklin: I still don't understand why we don't just rely on kwarg pass through?  

```python
    def __init__(..., **kwargs):
        super(Server, self).__init__(..., **kwargs)
```

Why do we need to specifically call an `ssl_context_creator` function at all?  Why can't the caller pass in `ssl_options` from the `Schduler(..., ssl_options=...)` constructor and have them trickle through?

Perhaps you can share an example of how you're doing this now and why this change makes things easier?

mrocklin: Yeah, so I think at the moment I'm mostly confused because I don't understand the pain you're currently trying to solve.  If I were to solve this problem (that I admittedly know little about) I would probably check for ssl certificates whereever I called Scheduler/Worker and then send them in at that point.  My guess is that this would be at the CLI level.

hussainsultan: Got it. Your point around  not doing this in server object makes sense. Let me make some changes tomorrow.

hussainsultan: And thank you. This is helpful.

pitrou:
Hi Hussain, 

I have a bunch of comments.

On certificates: Tornado does check certificates by default, see http://www.tornadoweb.org/en/stable/iostream.html#tornado.iostream.IOStream.start_tls. It is probably reasonable to configure the same cert and key for all server ports on the scheduler. Dask clients and workers are clients (in the TCP sense), so we don't need to configure a private key there (additional security might perhaps require that the scheduler only accept connections from authenticated clients), but we need to give the clients a suitable CA cert (or the server cert itself) to validate the server's identity.

On tests: I think we need actual connection tests, not just simple mocking.

On configuration: as mentioned, we will need to configure the CA cert on all clients and workers. This probably warrants a dedicated CLI option. Same for the server (one option for each of cert and key file). An explicit "enable SSL" flag may not be needed if we enable it whenever a cert file is specified.

hussainsultan:
Thanks you @pitrou . How should we handle direct Client to worker communication? Wouldn't that require certificates and private key present in worker? I might be missing something but thought i'd ask. Thanks!

hussainsultan:
btw, i intend to clean this up soon but havent been able to find time for it. I am running this branch in production with single root cert and per worker certs/key pair. It appears that is ok with infosec folks at my work but i agree that worker should be treated as a client and scheduler as server in TCP nomenclature.

mrocklin:
FWIW clients and workers rarely communicate directly.  They usually use the scheduler as an intermediary.

mrocklin:
@pitrou thoughts on the right way to support SSL within Dask?

pitrou:
The main question is how to handle configuration and what people expect for security guarantees (i.e. do all peers authenticate each other? perhaps only the scheduler and the workers are authenticated? etc.).

hussainsultan:
closing this in favor of #866

PR #541, opened 2016-09-20
Fix bug when creating the log-directory

Chilipp:
``dask-ssh --log-directory .`` raises an error because the ``'{logdir}'`` directory is not existent because the correct path is not inserted in the logging command. Instead it creates a directory with the name ``'{logdir}'``. This commit inserts the correct directory in the shell command


mrocklin:
Looks great.  Thanks!

PR #542, opened 2016-09-20
Only increase open file limit on linux

mrocklin:
Previously we set the open file limit to half of the hard limit.  This
used to fail on OSX because they set their hard limit at infinity and
their soft cap at what appears to be 2**13 (though this doesn't seem
documented).

Now we don't set anything on osx and stick to defaults.  Given that OSX
users typically don't run into open file limits, this should be ok


canavandl:
This resolves the file ``ValueError: current limit exceeds maximum`` issue I was experiencing.

PR #544, opened 2016-09-21
Add deterministic ids to bokeh ui elements

canavandl:
Gives Bokeh UI elements deterministic ids (by default are uuids), so that they can be queried from the bokeh server process.


mrocklin:
Looks good to me.  Merging on passed tests.

canavandl:
Please hold before merging, i might have to change some ids

canavandl:
yes - it works as intended. this is ready to merge @mrocklin

PR #553, opened 2016-09-26
Stack durations

mrocklin:
This adds an estimate of the amount of time in the stack and uses that estimate when deciding which worker a task should use.  This improves load balancing.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2177,11 +2177,13 @@ class Scheduler(Server):
             if self.task_state[key] == 'stacks':  # TODO: non-linear
                 for w in self.stacks:
                     if key in self.stacks[w]:
-                        i = self.stacks[w].index(key)
-                        del self.stacks[w][i]
-                        duration = self.stack_durations[w][i]
-                        del self.stack_durations[w][i]
-                        self.stack_duration[w] -= duration
+                        for i, k in enumerate(self.stacks[w]):
+                            if k == key:
+                                del self.stacks[w][i]
+                                duration = self.stack_durations[w][i]
+                                del self.stack_durations[w][i]
+                                self.stack_duration[w] -= duration
+                            break
 
             self.released.add(key)
             self.task_state[key] = 'released'
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -106,6 +106,10 @@ class Scheduler(Server):
     * **processing:** ``{worker: {key: cost}}``:
         Set of keys currently in execution on each worker and their expected
         duration
+    * **stack_durations:** ``{worker: [ints]}``:
+        Expected durations of stacked tasks
+    * **stacks_duration:** ``{worker: int}``:
+        Total duration of all tasks in each workers stack
     * **rprocessing:** ``{key: {worker}}``:
         Set of workers currently executing a particular task
     * **who_has:** ``{key: {worker}}``:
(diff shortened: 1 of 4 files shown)


mrocklin:
@martindurant @jcrist this could use brief review.  This shows an example of tracking additional state in the scheduler to improve scheduling.  This was motivated by heavy imbalances detected in some dataframe workloads.

distributed/scheduler.py
```diff
@@ -1818,14 +1832,17 @@ def transition_waiting_ready(self, key):
...
-                        key)
+                        self.stack_duration, self.processing, self.who_has,
+                        self.has_what, self.restrictions,
+                        self.loose_restrictions, self.nbytes, self.ncores, key)
                 if not new_worker:
                     self.unrunnable.add(key)
                     self.task_state[key] = 'no-worker'
                 else:
                     self.stacks[new_worker].append(key)
+                    duration = self.task_duration.get(key_split(key), 0.5)
```
martindurant: The task durations are as before, an average over previous tasks of the same type?

distributed/scheduler.py
```diff
@@ -2860,18 +2894,19 @@ def decide_worker(dependencies, stacks, processing, who_has, has_what, restricti
...
-                len(has_what.get(w, ())))
-    worker = min(workers, key=objective)
-    return worker
+    # Select worker that will finish task first
+    best_time = 1e100
+    best_worker = None
+    for w in workers:
+        comm_bytes = sum([nbytes.get(k, 1000) for k in dependencies[key]
+                          if w not in who_has[k]])
+        stack_time = stack_duration[w] / ncores[w]
```
martindurant: This is the actual usage for the durations - but the objective function changed too. Is the only effective change here that we measure the execution time of the stack instead of the number of tasks?

martindurant: It may be nice to still somehow formulate is as an "objective" function, to encourage users to provide their own for some other use cases.

martindurant:
I saw no way to update the times in the stack durations as new information comes in about other similar tasks that have completed. Perhaps that would be another layer of complexity, but the task estimate could be very wrong, especially when some of the tasks have the default 0.5s duration.

distributed/scheduler.py
mrocklin: Previously we minimized based on communication bytes and then only used stack-length to break ties.  Now with stack durations we can meaningfully compare the two, so both end up being used in the initial minimization.

mrocklin: Formulating this as an objective function is much cleaner.  Thanks for the suggestion.  Fixed.

mrocklin:
> I saw no way to update the times in the stack durations as new information comes in about other similar tasks that have completed. Perhaps that would be another layer of complexity, but the task estimate could be very wrong, especially when some of the tasks have the default 0.5s duration.

That's correct.  This may be suboptimal at times.  Deques don't offer constant time random access.  I don't see a simple way to keep stack_durations accurately up to date cheaply, so at the moment we're stuck to a somewhat out-dated estimate.  I'm not overly concerned here though.  

The 0.5s default could use some deeper thinking. It would be nice to figure out what the conservative choice is here.

distributed/scheduler.py
mrocklin: Yes, exponential weighted moving average based on calling `key_split` on the keyname.

PR #554, opened 2016-09-28
Individual app routes for bokeh models

canavandl:
This PR branch is intended to work with https://github.com/dask/dask-labextension/pull/3 as a proof of concept of working with the distributed ui inside of JupyterLab. It was discovered that a deficiency in Bokeh, related to selecting specific models from within a document containing a nested layout, made resizing Bokeh elements not work. This is a **short-term workaround** until the Bokeh deficiency is addressed (https://github.com/bokeh/bokeh/issues/5244).

This PR branch contains a lot of (hopefully) unnecessary code to provide individual app routes for each UI element and (definitely) has a lot of repeated code.

### DIFF
--- a/distributed/bokeh/application.py
+++ b/distributed/bokeh/application.py
@@ -15,7 +15,7 @@ from ..utils import ignoring
 from ..compatibility import logging_names
 from ..config import config
 
-dask_labextension_paths = ['resource_profile', 'network_profile']
+dask_labextension_paths = ['resource_profile', 'network_profile', 'memory_usage']
 
 dirname = os.path.dirname(distributed.__file__)
 paths = [os.path.join(dirname, 'bokeh', name)
--- a/distributed/bokeh/application.py
+++ b/distributed/bokeh/application.py
@@ -15,14 +15,9 @@ from ..utils import ignoring
 from ..compatibility import logging_names
 from ..config import config
 
-dask_labextension_paths = [
-    'resource_profile', 'network_profile', 'memory_usage', 'task_stream',
-    'progress_stream', 'processing', 'worker_memory', 'worker_table',
-    ]
-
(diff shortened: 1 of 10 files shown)


canavandl:
So I think moving each plot into it's own app route may be the solution for the jupyter lab extension for the middle term - the bokeh issue with selecting individual models from a layout seems difficult at the moment. I *think* that the additional app routes, if not actively being queries, shouldn't add any notable CPU overhead.

However, this branch has a ton of duplicated code that needs to be cleaned up before merging.

canavandl:
I've removed the added dask-labextension app routes from the web UI index page, so the traditional UI should look (and perform?) identically as before. I've also tried to minimize the amount of duplicate code for the individual app routes.

ping @mrocklin for review/feedback on the advisability of merging this. The Bokeh issue of being unable to select non-layout wrapped models from an existing app route seems difficult and unlikely to get addressed in the next month or two, short of some political maneuvering.

canavandl:
I'm also experiencing build timeouts/failures and I don't have permissions to restart the builds.

mrocklin:
Hrm, one thought to the code duplication would be to put the update functions in some other place.  One option would be to include them in the same function that generates the source and plot

### Before

    source, plot = make_plot(...)
￼
### After

    source, plot, callback = make_plot(...)

Or maybe this is enough to have us switch to some class-based thing?

mrocklin:
> So I think moving each plot into it's own app route may be the solution for the jupyter lab extension for the middle term - the bokeh issue with selecting individual models from a layout seems difficult at the moment. I think that the additional app routes, if not actively being queries, shouldn't add any notable CPU overhead.

Is this still the case?

Thoughts on using scripts rather than directories?

Thoughts on how to package up redundant code?

PR #560, opened 2016-10-04
Add Future.add_done_callback

mrocklin:
Fixes #1608

@dhirschfeld

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -151,6 +151,15 @@ class Future(WrappedKey):
         """
         return sync(self.client.loop, self._exception)
 
+    def add_done_callback(self, fn):
+        """ Call callback on future when callback has finished
+
+        The callback ``fn`` should take the future as its only argument.  This
+        will be called regardless of if the future completes successfully,
+        errs, or is cancelled
+        """
+        self.client.loop.add_callback(done_callback, self, fn)
+
     def cancel(self):
         """ Returns True if the future has been cancelled """
         return self.client.cancel([self])
@@ -223,6 +232,14 @@ class Future(WrappedKey):
             return '<Future: status: %s, key: %s>' % (self.status, self.key)
 
     __repr__ = __str__
+
+
+@gen.coroutine
+def done_callback(future, callback):
+    """ Wait on future, then call callback """
+    while future.status == 'pending':
+        yield future.event.wait()
+    callback(future)
 
 
 @partial(normalize_token.register, Future)
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -3635,3 +3635,19 @@ def test_distribute_tasks_by_ncores(c, s, a, b):
     yield _wait(futures)
 
     assert len(b.data) > 2 * len(a.data)
+
+
+@gen_cluster(client=True)
+def test_add_done_callback(c, s, a, b):
+    x = c.submit(inc, 1)
+
+    L = []
+    def f(future):
+        L.append(future.key)
+        L.append(future.status)
+
+    x.add_done_callback(f)
+
+    yield _wait(x)
+
+    assert L == [x.key, x.status]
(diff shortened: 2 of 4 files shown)


dhirschfeld:
LGTM and I think it's useful to follow the `concurrent.futures` api however my usecase is to wait for a list of futures to complete so I don't think I can use this for what I'm after. 

I think adding a callback to a list of futures is also useful functionality so it might be convenient to have a corresponding method `Client.add_done_callback(futures, callback)` as well at the method on the Future itself.

The technique in #1608 does work for this but having the method would be more discoverable for tornado noobs like myself.

minrk:
Does it make sense to inherit from concurrent.futures.Future (or tornado.concurrent.Future)?

Tornado has a [multi](http://www.tornadoweb.org/en/stable/gen.html#tornado.gen.multi) function for quickly dealing with multiple futures:

```python
future = gen.multi([f, f, f])
future.add_done_callback(cb)
```

or quickly adding a callback when all of a collection of futures are done:

```python
gen.multi([f, f, f]).add_done_callback(cb)
```

dhirschfeld:
@mrocklin - I'm getting an exception on import on this branch:
```python-traceback

  File "C:/dev/code/sandbox/test_dask.py", line 8, in <module>

  File "c:\dev\packages\distributed\distributed\__init__.py", line 5, in <module>
    from .deploy import LocalCluster

  File "c:\dev\packages\distributed\distributed\deploy\__init__.py", line 3, in <module>
    from .local import LocalCluster

  File "c:\dev\packages\distributed\distributed\deploy\local.py", line 15, in <module>
    from ..scheduler import Scheduler

  File "c:\dev\packages\distributed\distributed\scheduler.py", line 11, in <module>
    import resource

ImportError: No module named 'resource'
```
It  seems this is a unix-only module?

dhirschfeld:
It looks like the windows equivalent is `_getmaxstdio`/`_setmaxstdio`?

image
https://msdn.microsoft.com/en-us/library/kdfaxaay(v=vs.140).aspx

http://docs.activestate.com/activepython/2.7/pywin32/win32file___getmaxstdio_meth.html
http://docs.activestate.com/activepython/2.7/pywin32/win32file___setmaxstdio_meth.html
image

image

Looking at the [commit](https://github.com/dask/distributed/commit/b2cc90a88be5246231325c773c1d19e7997df3da) where that was implemented i'm not quite sure how to translate that to windows as I'm not sure (really have no idea) if there's an equivalent soft-limit concept.

dhirschfeld:
Thanks, @minrk - that looks like a handy convenience function. With this PR you can implement a similar workflow in distributed like below:
```python
futures = client.map(func, *args, **kwargs)
future = client.submit(lambda *args: args, futures)
future.add_done_callback(cb)
```

mrocklin:
@dhirschfeld is the Windows resource issue particular to this branch and not to master?

mrocklin:
@minrk have you found benefits from subclassing from `concurrent.futures.Future`?  Presumably things like `concurrent.futures.as_completed` will start to work cross-backend?  This seems reasonable to do.  I'd want to check out the ramifications first though.

Something like `gen.multi` could work.  It's a bit tricky in the distributed case because a `Future` sort of promises that all of the data is on one worker, which isn't always something we want.  We could also create a bunch of convenience functions, but the deviation here is a bit of a pain.

Anyway, I'm going to go ahead and merge this.  It seems innocuous enough and is inline with concurrent.futures.

dhirschfeld:
@mrocklin - it's present on master too. I opened #564 to track it

minrk:
> have you found benefits from subclassing from concurrent.futures.Future?

Mainly that I don't have to bother making sure that I reimplement the full API. I only implement the extensions to the API, and all of the common API I get to inherit. There's also the theoretical advantage of some library having `isinstance(f, Future)` checks, but I'm not sure that's come up.

PR #565, opened 2016-10-07
documentation on strategies for Windows architecture setups

jreback:


mrocklin:
Well isn't that interesting.  I guess I can now be lazy about file descriptors on Windows too :)

mrocklin:
Thanks for the well written writeup

docs/source/setup.rst
```diff
@@ -174,3 +174,21 @@ This is typically enforced through external means, such as by having a network
...
 typically employed by cluster administrators.
+
+Windows
+~~~~~~~
+
+.. note::
+
+  - Running a ``dask-scheduler`` on Windows architectures is supported for only a
+    limited number of workers (roughly 100). This is a detail of the underlying tcp server
+    implementation and is discussed `here`__.
```
dhirschfeld: I'm not an expert on this topic but my understanding is that this is purely a consequence of *tornado* not properly supporting Windows (which they fully admit) and since distributed is built on top of tornado it inherits its limitations.

IIUC the issue is that rather than implementing an event loop using IOCP as [asyncio does](https://docs.python.org/3/library/asyncio-eventloops.html#asyncio.ProactorEventLoop) they just fall back on select which on Windows isn't performant and is limited by Python's choice of [`FD_SETSIZE`](https://github.com/python/cpython/blob/master/Modules/selectmodule.c#L29-L37) value. 

I read the above as stating that the problem is inherent with Windows whereas I think it's more accurate to say that it's due to limitations in both *tornado* and the Windows compilation of Python. From [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms739169%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396):
> *If an application is designed to be capable of working with more than 64 sockets using the select and WSAPoll functions, the implementor should define the manifest FD_SETSIZE in every source file before including the Winsock2.h header file*

Interestingly I've come across this before in zeromq/libzmq#1165. They've now increased `FD_SETSIZE` to 16k - zeromq/libzmq#2035.  Perhaps, given that there's no recourse except recompiling, Python should also set a similar limit to avoid running out of sockets - in this day and age 512 sockets just doesn't cut it!

dhirschfeld: In case it's not clear - I'm happy for this PR to go in as it's an accurate description of where we are today with tornado/Python and gives a workaround.

PR #580, opened 2016-10-13
Fix #579: avoid race condition when closing streams

### LINKED ISSUE #579: Sporadic hangs under Windows
When running the test suite under Windows, it hangs randomly in some tests. Before the hang, a traceback similar to the following is reported:
```
Exception in thread Thread-58:
  File "c:\miniconda3\envs\dask35\lib\threading.py", line 914, in _bootstrap_inner
    self.run()
  File "c:\miniconda3\envs\dask35\lib\threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "c:\miniconda3\envs\dask35\lib\site-packages\tornado\ioloop.py", line 862, in start
    event_pairs = self._impl.poll(poll_timeout)
  File "c:\miniconda3\envs\dask35\lib\site-packages\tornado\platform\select.py", line 63, in poll
    self.read_fds, self.write_fds, self.error_fds, timeout)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

pitrou:
Closing streams from threads other than the running loop can
cause race conditions from inside the loop (the sockets become
invalid while polling for them).  The race condition only seems
to manifest under Windows, though it is theoretically OS-independent.

The fix involves two changes:
- no explicit closing from `__del__`, since `__del__` can be called
  in basically any Python thread
- in Client.shutdown(), close streams from the running loop rather
  than the calling thread

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -593,7 +593,15 @@ class ConnectionPool(object):
     def collect(self):
         logger.info("Collecting unused streams.  open: %d, active: %d",
                     self.open, self.active)
-        for k, streams in list(self.available.items()):
+        for streams in list(self.available.values()):
+            for stream in streams:
+                stream.close()
+
+    def close(self):
+        for streams in list(self.available.values()):
+            for stream in streams:
+                stream.close()
+        for streams in list(self.occupied.values()):
             for stream in streams:
                 stream.close()
 
--- a/distributed/bokeh/tests/test_application.py
+++ b/distributed/bokeh/tests/test_application.py
@@ -13,7 +13,7 @@ from distributed.utils import ignoring
 def test_BokehWebInterface(loop):
     with LocalCluster(2, loop=loop, scheduler_port=0,
                       services={('http', 0): HTTPScheduler},
-                      diagnostic_port=None) as c:
+                      diagnostics_port=None) as c:
         with pytest.raises(Exception):
             response = requests.get('http://127.0.0.1:8787/status/')
         with BokehWebInterface(
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -550,15 +550,12 @@ class Client(object):
         --------
         Client.restart
         """
+        # XXX handling of self.status here is not thread-safe
         if self.status == 'closed':
             return
-        self.status = 'closed'
-        with ignoring(AttributeError):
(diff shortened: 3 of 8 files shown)


distributed/core.py
```diff
@@ -462,9 +462,6 @@ def send_recv_from_rpc(**kwargs):
             raise gen.Return(result)
         return send_recv_from_rpc
 
-    def __del__(self):
-        self.close_streams()
```
mrocklin: Did removing this have any negative consequences?

mrocklin: OK, lets remove for now and see what happens then :)

pitrou: Well, looking at the Tornado source code, IO streams actually don't seem to feature a custom garbage collection mechanism, so perhaps `__del__` is really needed. I'll take a look.

pitrou: Hmm, spawning asynchronous code from `__del__` is fraught with dangers (because it can revive the object being del'ed). I think we should therefore not resinstate those methods, and instead make sure we call the `close_rpc` or `close_streams` methods from owning objects.

pitrou:
There is an unexpected failure in the bokeh tests:
https://travis-ci.org/dask/distributed/jobs/167440774#L1570

pitrou:
There's something weird with `distributed/bokeh/tests/test_application.py::test_BokehWebInterface`: it starts two bokeh apps on the same port 8787 (one implicitly through LocalCluster, and one implicitly in the test). I wonder why that is?

pitrou:
Ha, it's a typo in the test: `diagnostic_port` instead of `diagnostics_port`. The `**kwargs` passthrough in `LocalCluster.__init__` didn't catch it :-)

jreback:
let me give this a try on my large windows cluster tomorrow. having failures on ``.restart()`` of the whole cluster (but I can't seem to repro in the test code, though it isn't using many machines).

jreback:
thanks for your fixes on windows @pitrou !

pitrou:
> thanks for your fixes on windows @pitrou !

I'm hoping we can get AppVeyor CI for distributed once we tackle all issues.

mrocklin:
Rebased and added another commit to close ConnectionPool objects on server close.

Actually, I'm quite surprised that I can force-push to your branch.  The new Github policies break some of my expectations.

PR #584, opened 2016-10-13
Clean up rpc objects explicitly

mrocklin:
Previously we depended on `__del__` and garbage collection to clean up
rpc objects.  This caused some difficulties, particularly on Windows.

Now we clean up all rpc objects explicitly.  Most of this PR commit is
just tweaking other tests to be more robust.

We test this by explicitly checking the rpc count in the gen_cluster decorator
and the cluster testing context manager.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -395,6 +395,8 @@ class rpc(object):
 
     >>> remote.close_streams()  # doctest: +SKIP
     """
+    active = 0
+
     def __init__(self, arg=None, stream=None, ip=None, port=None, addr=None,
             timeout=3):
         ip, port = ip_port_from_args(arg=arg, addr=addr, ip=ip, port=port)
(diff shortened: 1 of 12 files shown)


mrocklin:
cc @pitrou

mrocklin:
I tracked down the restart issues.  Tests run locally.

jreback:
who hoo!

I assume I should deploy both on the scheduler and worker nodes?

mrocklin:
For restart things this is most relevant on the scheduler.

jreback:
hmm, isn't there the change in the nanny code?

mrocklin:
Ah, yes, you're correct

jreback:
@mrocklin this so fixed windows ``.restart()``!

now very fast! (and works!)

jreback:
thank you @pitrou for suggesting this method!

PR #585, opened 2016-10-14
Retry on new http port if socket.error

mrocklin:
We already do this in core.Server.  We should do it in http.core.MyApp as well


mrocklin:
I can't think of an easy way to robustly test this.  Merging as is.

PR #597, opened 2016-10-19
Fix CLI tests under Windows

pitrou:
Note that process cleanup seems quite different under Windows than it is under Unix. A child process can apparently continue running even though the parent disappeared. Making things really robust is out of the picture for this PR.

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -81,13 +81,15 @@ def main(host, port, http_port, bokeh_port, show, _bokeh,
         except Exception as e:
             logger.warn("Could not start Bokeh web UI", exc_info=True)
 
-    loop.start()
-    loop.close()
-    scheduler.stop()
-    if bokeh_proc:
-        bokeh_proc.close()
+    try:
+        loop.start()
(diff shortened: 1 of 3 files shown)


pitrou:
With this PR, all distributed tests now pass under Windows here (Python 3.5, Windows 7 64-bit VM). The whole test suite takes 700 s.

jreback:
this probably fixes a bug I have been seeing. a work-flow like this (using master)

- start scheduler, workers with nanny
- do work
- .restart works fine
- hard kill scheduler
- restart scheduler
- hard kill workers
- restart workers
- .restart hangs & we have some windows zombie workers.

The reason for this is to load a new version of packages to both scheduler and workers. I should prob NOT leave work on the cluster when doing this...hahh (not sure if that does it). But it *seems* that causes zombies (as opposed to a clean cluster where .restart is fine).

mrocklin:
This looks good to me.  Merging soon if no further comments.

@pitrou any desire to set up appveyor again for dask/distributed ?

pitrou:
Yes, i'll take a look at enabling AppVeyor once this is merged.

PR #598, opened 2016-10-19
Add AppVeyor configuration

pitrou:
This also fixes some failing tests on AppVeyor.


pitrou:
Unfortunately there's still a failing test: https://ci.appveyor.com/project/pitrou/distributed/build/1.0.9#L921

pitrou:
Note the proposed AppVeyor configuration also uploads test results in XML format, to give another view (in addition to raw logs) as shown here: https://ci.appveyor.com/project/pitrou/distributed/build/1.0.9/tests

Nothing terrific, though.

mrocklin:
Seeing the durations of each test can be somewhat useful at least

mrocklin:
I'm also unable to reproduce this test.  I've tried running the failing in a loop many times on my linux workstation as well as running the entire stress test suite many times in a loop.

I'm tempted to punt on this part of this test for Windows.

pitrou:
I wonder if someone else can give it a try. @jreback perhaps?

jreback:
i will give this a run tomorrow

but don't wait on. me if urgent!

mrocklin:
I've been experiencing similar failures on travis-ci: https://travis-ci.org/dask/distributed/jobs/169339745

I suspect that we're still not closing down all rpc objects perfectly, leading to hard-to-reproduce failures.  I think that this is not isolated to Windows.

pitrou:
The `__del__` I added in this PR should display a warning if a rpc object wasn't closed. However, that assumes the `__del__` method is actually called...

pitrou:
Now that I look at the Travis-CI failure, `rpc.active` being 14 at the beginning of the test hints that *many* rpc objects weren't closed properly, right?

mrocklin:
> Now that I look at the Travis-CI failure, rpc.active being 14 at the beginning of the test hints that many rpc objects weren't closed properly, right?

We only test this in tests running `@gen_cluster` or `with cluster(...)`.  I didn't bother to clean up the other tests that don't use one of these two testing harnesses, so I'm not surprised by a slowly rising number.

mrocklin:
Is the failure intermittent or predictable.  If it is intermittent then it's probably related to the other similar errors we've been getting recently.  If so then I think we can probably merge this and I can try to hunt down the lingering rpc leak sometime next week.

pitrou:
It seems entirely deterministic on AppVeyor (happens every time). It would be nice if @jreback managed to reproduce it as well :-)
Note that I haven't tested with the exact same package configuration as on AppVeyor, yet. Perhaps I should try that...

pitrou:
Build success here (lucky?):
https://ci.appveyor.com/project/pitrou/distributed/build/1.0.10

mrocklin:
Can we trigger a few appveyor builds to see if there are intermittent errors?

pitrou:
Yes, I'll do so.

pitrou:
Verdict over 4 runs: 2 successes, 2 have a failed test with `rpc.active` being the wrong value.

pitrou:
Verdict over 4 runs: 2 successes, 2 have a failed test with `rpc.active` being the wrong value.

mrocklin:
I'm tempted to not test rpc.active on windows for now.  I would like to get appveyor in and while rpc.active is an issue, it's not huge.

Do you recall the tests where it failed?

pitrou:
It's test_cancel_stress in both cases: https://ci.appveyor.com/project/pitrou/distributed/build/1.0.13 and https://ci.appveyor.com/project/pitrou/distributed/build/1.0.14.
By the way, should AppVeyor be configured with the dask account (rather than my personal one)?

mrocklin:
I suspect that you don't want to be the only person on call to handle Windows issues in the future :)

pitrou:
I'd rather not :-)

mrocklin:
Perhaps we just skip the cancel_stress test on Windows then?  That seems more precise.

pitrou:
Ok, I've run three builds after the test_cancel_stress skipping and they were all successful. i'm frankly a bit surprised.

mrocklin:
We could keep going.  I wouldn't be surprised if some errors show up less frequently than 30% of the time.

mrocklin:
What is the status on this?

pitrou:
There was a timeout error in the latest build (see above), I wonder if I set the test timeout to a too short value. Will try to up it.

pitrou:
Or perhaps it's just the usual sporadic hang in `test_iterator_gather`...

mrocklin:
What is the status on this?

pitrou:
> What is the status on this?

There are random hangs when running the test suite which I don't manage to reproduce locally. They seem unrelayed to the Tornado issue. Latest example:
https://ci.appveyor.com/project/dask-ci/distributed/build/1.0.2#L444

mrocklin:
Are they in a few tests that we can xfail for now and fix in the future?  Or are they in many tests?

pitrou:
As far as I can see from the last three builds, the hangs are bit all over the place, unfortunately.

mrocklin:
Is this a recent change?

pitrou:
If we consider 275721dca91a43b749f47e45fe06d4fd36d31a5b to be the first changeset where the hangs happen (it's a merge from master), then unless I'm mistaken the differences with parent 32bd8b938e702d9adb731bf5d3a5ec12184412f9 are only in `distributed.bokeh`. Specifically, it seems PR #621.

pitrou:
I managed to unstuck the AppVeyor builds. Now the only issue remaining are a couple of failures when checking `rpc.active`.

mrocklin:
By the way, I was able to trigger more rpc.active failures by adding a delay to my local network.  Some of these went away with https://github.com/dask/distributed/pull/692/commits/e27af40c5ff03f6c812b7d1d9652f15d01e811b3

pitrou:
> Some of these went away with e27af40

That would deserve a comment in the code, no?

pitrou:
Wow, a green build.

mrocklin:
We could, yes.  I wouldn't mind waiting a bit to verify that this does actually solve the problem.  My evidence so far is only anecdotal, so there is only a commit log message.

pitrou:
Since this is now working (except for the usual sporadic failures perhaps), the few code changes here are ready for reviewing.

distributed/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+def _run_psutil(method_name, default):
+    try:
+        import psutil
+    except ImportError:
+        return default
+    try:
+        return getattr(psutil, method_name)()
+    except RuntimeError:
+        # This can happen on some systems (e.g. no physical disk in worker)
+        return default
```
mrocklin: Perhaps we should move the psutil check outside?

```python
try:
    from psutil import disk_io_counters, net_io_counters
except importError:
    disk_io_counters = net_io_counters = _empty_named_tuple
```

Given the comment on no physical disk I suspect that the solution above may be overly simplistic.  Perhaps we could try running each method once here?

PR #601, opened 2016-10-20
Improve file handling

mrocklin:
We used to generate files in two cases:

1.  the `~/.dask/config.yaml' file
2.  the `~/.dask/.dask-web-ui.yaml` file that held the scheduler address

These would break in some cases when the file system was used simultaneously by several systems.  

In the first case we are now robust to overwrites

In the second case we have switched to using command line arguments in bokeh

### DIFF
--- a/distributed/bokeh/application.py
+++ b/distributed/bokeh/application.py
@@ -9,7 +9,7 @@ import sys
 
 import bokeh
 import distributed.bokeh
-from toolz import get_in
+from toolz import get_in, concat
 
 from ..utils import ignoring
 from ..compatibility import logging_names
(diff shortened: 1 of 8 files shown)


jreback:
very nice! 

since these are generated by the workers, shouldn't this also be an argument to the actual Worker creation?

mrocklin:
Workers should only collide on the second commit, when auto-generating config files.  It's unlikely that anything in the config file is of high relevance to the worker if you're generating it anyway (nothing was set up ahead of time) so in this case it doesn't matter so much what happens, just that it doesn't break.

quasiben:
+1

Issue #604, opened 2016-10-24
Custom Serialization


mrocklin:
### Current state

Currently functions that move data first call pickle.dumps on that data before they return the result, which then gets sent through a socket using our [protocol](http://distributed.readthedocs.io/en/latest/protocol.html) (msgpack + large byte buffers).  Usually the large bytestrings end up getting plucked out of the message and sent at the end as separate frames.  This approach is simple, but has a few problems:
1.  Pickle does not always provide ideal performance
   -  For example on numpy there are two extraneous memory copies
   -  For example it can be quite slow on on pandas object dtype 
   -  For example lists of long-form-text can be more efficiently encoded as msgpack
2.  We sometimes want to compress differently for different types, for example by using blosc on numpy arrays
3.  Some types, like netCDF4.Dataset can not be pickled, but could be represented with filename/datapath pairs (useful for https://github.com/pydata/xarray/issues/798)
4.  Reliance on Pickle forces workers and clients to all have the same software environment.  Having established serialization formats would be a first step towards mutli-language worker inter-operation (a surprisingly frequently requested feature in enterprise)
5.  The fact that we call pickle.dumps in many places means that it's difficult to capture information or change policies about serialization.  Future changes like destination specific serialization (https://github.com/dask/distributed/issues/400) or better diagnostics might want a more centralized location that can make serialization decisions based on a variety of factors.

Generally speaking how we choose to serialize can depend on many things including the type of the object, the location/environment/bandwidth-distance of the destination, etc..  Additionally, for optimal performance we can't just turn data into bytes, leave it around, and then later stick it on a wire with a bunch of other bytes.  We sometimes want more control.
### Proposal

When a function returns a value that it knows is destined to end up on a wire it no longer serializes immedately.  Instead it wraps the data in a `Serialize` object.  

``` python
class Serialize(object):
    def __init__(self, data):
        self.data = data
```

This object serves as a signal to the protocol that it should treat this data in a special way.  At the last possible moment before we ship data to the wire, the protocol searches for these objects within the message, pulls them out as separate frames (as we do today for large bytestrings), and calls a custom serialization method on them.  The custom serialization function will be determined by type to start with, but may eventually grow to also include information about the destination, such as the location or software environment.  The information about the custom serialization will go into the header that we use now for large bytestrings.  

Custom serialization methods will provide a list of `bytes` or `memoryview` objects that will be written linearly to a socket.  Deserialization methods will get a list of `memoryview` objects in return.  

[...]

PR #606, opened 2016-10-25
Support custom serialization

mrocklin:
This changes the protocol to support custom serialization methods on a per-type basis.

Fixes https://github.com/dask/distributed/issues/604 which also includes an explanation of the approach taken here.

As a proof of concept I implemented a custom serialization for numpy ndarrays that both uses blosc (if available) and avoids the two memory copies in pickle.dumps.


mrocklin:
This could use review if anyone has some free time:

I recommend starting at the docpage and the protocol directory, with a focus on the numpy example: 

* https://github.com/mrocklin/distributed/blob/protocol-serialize/docs/source/serialization.rst
* https://github.com/mrocklin/distributed/tree/protocol-serialize/distributed/protocol
* https://github.com/mrocklin/distributed/blob/protocol-serialize/distributed/protocol/numpy.py

cc @pitrou @minrk @ogrisel

distributed/protocol/core.py
```diff
@@ -0,0 +1,221 @@
...
+            header['keys'].append(key)
+            out_frames.extend(frames)
+
+        for key, (head, frames) in pre.items():
+            head['lengths'] = list(map(len, frames))
+            header['headers'][key] = head
+            header['keys'].append(key)
+            out_frames.extend(frames)
+
+        out_frames = [bytes(f) for f in out_frames]
```
mrocklin: This possible copy should become unnecessary if we can get something like https://github.com/tornadoweb/tornado/pull/1691 in to tornado

distributed/protocol/compression.py
```diff
@@ -0,0 +1,113 @@
...
+
+with ignoring(ImportError):
+    import blosc
+    compressions['blosc'] = {'compress': partial(blosc.compress, clevel=5,
+                                                 cname='lz4'),
+                             'decompress': blosc.decompress}
+
+
+default = config.get('compression', 'auto')
+if default != 'auto':
```
pitrou: 'auto' doesn't do anything?

distributed/scheduler.py
```diff
@@ -1657,26 +1657,27 @@ def feed(self, stream, function=None, setup=None, teardown=None, interval=1, **k
...
-        try:
-            while True:
-                if state is None:
-                    response = function(self)
-                else:
-                    response = function(self, state)
-                yield write(stream, response)
-                yield gen.sleep(interval)
-        except (OSError, IOError, StreamClosedError):
+        with log_errors(pdb=True):
```
pitrou: I suppose we shouldn't keep this in production code.

distributed/protocol/pickle.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    try:
+        result = pickle.dumps(x, protocol=pickle.HIGHEST_PROTOCOL)
+        if len(result) < 1000:
+            if b'__main__' in result:
+                return cloudpickle.dumps(x, protocol=pickle.HIGHEST_PROTOCOL)
+            else:
+                return result
+        else:
+            if isinstance(x, pickle_types) or b'__main__' not in result:
```
pitrou: I'm unsure what `pickle_types` is here?

mrocklin: Originally these were types for which pickle definitely worked and where we didn't want to use cloudpickle.  This notably included numpy arrays.  I must have lost the variable while moving around code.  Clearly this code isn't covered though and needs to be fixed.

distributed/protocol/serialize.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+# Teach serialize how to handle bytestrings
+def serialize_bytes(obj):
+    header = {}  # no special metadata
+    frames = [obj]
+    return header, frames
+
+def deserialize_bytes(header, frames):
+    return b''.join(frames)  # the frames may be cut up in transit
```
pitrou: When you say "in transit", it's in case some level of compression is enabled?

mrocklin: There are various reasons to cut up large bytestrings.  I think in the current case we do need this for compression, yes.

Arguably we should be able to handle this internally though without the custom functions being aware of it.

distributed/protocol/core.py
```diff
@@ -0,0 +1,195 @@
...
+            if 'compression' not in head:
+                frames = frame_split_size(frames)
+                compression, frames = zip(*map(maybe_compress, frames))
+                head['compression'] = compression
+            head['lengths'] = list(map(len, frames))
+            header['headers'][key] = head
+            header['keys'].append(key)
+            out_frames.extend(frames)
+
+        for key, (head, frames) in pre.items():
```
mrocklin: Ideally the frames should already be compressed

pitrou:
I'm a bit confused trying to follow the layers of protocol encoding. Whose responsiblity is it to call `to_serialize()`?

mrocklin:
> I'm a bit confused trying to follow the layers of protocol encoding. Whose responsiblity is it to call to_serialize()?

Application code like the worker and client functions.  For example

```python
class Worker:
    def get_data(self, keys):
        return {key: to_serialize(self.data[key]) for key in keys}
```

distributed/protocol/serialize.py
mrocklin: This is maybe fixed in https://github.com/dask/distributed/pull/606/commits/05198a8dfb8d2cf58150adb34ee7c9341aad3368

distributed/protocol/core.py
mrocklin: This is now tested in https://github.com/dask/distributed/pull/606/commits/9d688472c09cf84a83af0f0ba0f68454dcbf2ece

distributed/protocol/pickle.py
mrocklin: Fixed

distributed/protocol/compression.py
mrocklin: If `config['compression'] == 'auto'` then we use the `default_compression` determined above during the import process.

pitrou:
> Application code like the worker and client functions.

What happens if a dask graph produces e.g. intermediate results as Numpy arrays (think about rechunk()). Do those get the special serialization treatment as well when shuffled around between workers?

mrocklin:
Yes, worker-worker communication is actually handled by exactly the function I used above:

```python
class Worker:
    def get_data(self, keys):
        return {key: to_serialize(self.data[key]) for key in keys}
```

So those messages contain data in custom serialized form.  Any time you *don't* wrap a part of a message within `to_serialize` you are implying that it can be serialized with msgpack.

mrocklin:
I've added a test to demonstrate inter-worker communication works fine.

distributed/protocol/serialize.py
```diff
@@ -0,0 +1,176 @@
...
+    Examples
+    --------
+    >>> from distributed import Scheduler
+    >>> typename(Scheduler)
+    'distributed.scheduler.Scheduler'
+    """
+    return typ.__module__ + '.' + typ.__name__
+
+
+def serialize(x):
```
pitrou: I suppose the case where x is e.g. a tuple of Numpy arrays (instead of a single array) isn't covered by this... Does such a case never happen.

mrocklin: That's a good point.  We would fall back to pickle in that case.  Perhaps we should dive within standard containers.

mrocklin: I think I'm going to wait on this until it comes up in practice.  There are some tricky things to consider (how to avoid type checking all entries of long lists of text for example) and I'd like to ensure that what's in here handles itself well.

minrk:
Nice! This is a lot like IPython's custom serialization. How do you propagate the custom deserializer to peers, so they know how to deserialize a custom-serialized message?

mrocklin:
Each piece of serialized data comes along with a header that mentions the kind used, the compression per frame, the length per frame, etc..  Currently we use serialization by type and use the `'modulename.classname'` string as an identifier.

minrk:
Right - I think I meant how does the field in `deserializers` get populated on the destination? I get how the key gets there, just not the value.

mrocklin:
Ah, we currently assume that it is imported on the client.  

This works fine for types that we manage, like numpy arrays.  

We could probably use something like [worker environments](https://github.com/dask/distributed/pull/505) (in a stalled PR) to distribute custom serialization functions

mrocklin:
I would like to merge this soon.

distributed/utils_comm.py
```diff
@@ -76,8 +77,6 @@ def gather_from_workers(who_has, deserialize=True, rpc=rpc, close=True,
         bad_addresses |= {v for k, v in rev.items() if k not in response}
         results.update(merge(response))
 
-    if deserialize:
```
pitrou: Does this map the `deserialize` parameter is now obsolete on this function?

distributed/utils.py
```diff
@@ -492,6 +492,10 @@ def ensure_bytes(s):
     """
     if isinstance(s, bytes):
         return s
+    if isinstance(s, memoryview):
+        return s.tobytes()
+    if isinstance(s, buffer):
```
pitrou: `buffer` doesn't exist on Python 3.

distributed/utils_comm.py
mrocklin: Removed

distributed/utils.py
mrocklin: Guarded.  Also added this to tests

mrocklin:
Merging.

distributed/protocol/numpy.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import print_function, division, absolute_import
+
+import sys
+
+import numpy as np
+
+try:
+    import blosc
+    n = blosc.set_nthreads(2)
```
jakirkham: Why was this hard-coded to `2`?

mrocklin: We want some benefit from threads, but don't want to take them all (dask itself does parallelism)  Two seemed like a decent default (more decent than "all").  Do you have other thoughts on what this should be?

mrocklin: Yeah, we could place this in the `~/.dask/config` file.  Users can also call `blosc.set_threads` themselves.  It's possible that blosc itself should have some place to make this configurable instead.

jakirkham: > Yeah, we could place this in the `~/.dask/config` file.

This seems like the most desirable option.

mrocklin: Can I interest you in raising an issue or perhaps a short PR?  Relevant files is distributed/config.py and an example is here:

    distributed/scheduler.py:BANDWIDTH = config.get('bandwidth', 100e6)

jakirkham: Agreed done in issue ( https://github.com/dask/distributed/issues/1054 ).

PR #608, opened 2016-10-26
distributed UI cleanup

canavandl:
WIP


distributed/bokeh/plots/status.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+
+        self.root.add_tools(
+            hover, ResetTool(), PanTool(dimensions="width"), WheelZoomTool(dimensions="width")
+        )
+
+        # Required for update callback
+        self.task_stream_index = [0]
+
+    def update(self, messages):
```
canavandl: Needs docstring

canavandl:
ping @mrocklin for:

1. General feedback on proposed class structure (``Element`` object that creates layoutable bokeh `root` in init method and has update method). I'll probably tweak variable names and add docstrings later. 
2. In terms of testing, is there an existing util to create a `distributed.bokeh.messages` fixture? I could probably write it myself, but my goal is to ensure that if the messages dict changes, then the plot tests also break.

distributed/bokeh/plots/status.py
```diff
@@ -0,0 +1,102 @@
+from bisect import bisect
```
mrocklin: from __future__ import print_function, division, absolute_import

distributed/bokeh/plots/status.py
```diff
@@ -0,0 +1,102 @@
...
+
+from bokeh.models import (
+    ColumnDataSource, Plot, Datetime, DataRange1d, Rect, LinearAxis,
+    DatetimeAxis, Grid, BasicTicker, HoverTool, BoxZoomTool, ResetTool,
+    PanTool, WheelZoomTool, Title
+)
+
+from distributed.utils import log_errors
+
+class Element(object):
```
mrocklin: Thoughts on expanding this to `DashboardElement`?

distributed/bokeh/plots/status.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self):
+        """
+        Should be overridden by subclasses.
+
+        self.source ColumnDataSource
+        self.root should a layout-able object, a Plot, Layout, Widget or WidgetBox
+        """
+        self.source = None
+        self.root = None
```
mrocklin: The term `root` wasn't immediately obvious to me.  Is this the standard Bokeh term for a layout-able object?  Something like `plot` might be more suggestive if it's not also incorrect in this case.

canavandl: yes, ``root`` is the Bokeh term for a layout-able object, which encompasses plots, widgets and layouts (rows, columns and widgetboxes) of everything. I'll try to expand the docstring to explain this better and consider changing the property name.

TomAugspurger:
One thing I ran into today (because I'm silly): it'd be nice to have a status indictor for whether the connection to the scheduler is alive. I restarted the client / scheduler and spent a couple minutes figuring out why the plots weren't updating. I needed to refresh the dashboard page.

mrocklin:
That happens to me as well.  This is also a significant issue with the JLab plugin, where there doesn't seem to be a way to "refresh" a panel to retrigger the connection to a newly created scheduler.

canavandl:
It'd be trivial to add a tiny bit of front-end code to poll the scheduler and add an indicator whether the connection is live or not. As for the JupyterLab plugin, I'll think about a solution.

mrocklin:
To be clear, the non-responding webpage issue probably means that the Bokeh server is down too (it is usually married to the dask scheduler)

distributed/bokeh/components.py
```diff
@@ -0,0 +1,256 @@
...
+
+from bisect import bisect
+
+from bokeh.models import (
+    ColumnDataSource, Plot, Datetime, DataRange1d, Rect, LinearAxis,
+    DatetimeAxis, Grid, BasicTicker, HoverTool, BoxZoomTool, ResetTool,
+    PanTool, WheelZoomTool, Title, Range1d, Quad, Text, value
+)
+
+from distributed.diagnostics.progress_stream import progress_quads, nbytes_bar
```
canavandl: should migrate these methods as "parse_message" methods on relevant DashboardComponent class.

distributed/diagnostics/progress_stream.py
```diff
@@ -152,15 +162,50 @@ def progress_quads(msg, nrows=8, ncols=3):
...
-from bokeh.palettes import Spectral11, Spectral9, viridis
-import random
-task_stream_palette = list(viridis(25))
-random.shuffle(task_stream_palette)
-
-import itertools
-counter = itertools.count()
-@memoize
-def incrementing_index(o):
-    return next(counter)
```
mrocklin: Your travis errors are because of this.  You can't inline the counter.  It needs to stay outside of the function definition.

mrocklin:
OK, I'd like to merge this soon.  I think that future work can happen cleanly on top.

mrocklin:
I've rebased and merged this manually.  Closing.

PR #610, opened 2016-10-31
Spill to LMDB for small evicted worker data

### LINKED ISSUE #525: Spill to disk is inefficient for small byte-values
As of #485 workers can spill excess data to disk.  They currently evict tasks from a fixed pool of memory to disk with a simple LRU policy.  Task results are stored on disk as single files.  This is currently done using the composable MutableMappings in a tiny project, [zict](https://github.com/mrocklin/zict).

```python
if memory_limit:
    from zict import Func, Buffer, File
    path = os.path.join(self.local_dir, 'storage')
    memory = dict()
    disk = File(path)
    disk = Func(dumps_to_disk, loads_from_disk, disk)
    self.data = Buffer(memory, disk, int(float(memory_limit)), weight)  # LRU from memory to disk
else:
    self.data = {}
```

This policy is decent for large results (larger than say, a megabyte) but will slow down when trying to save many small results due to file system overhead.  This case specifically occurs when performing shuffles, such as with dask.array rechunkings, or dask.bag/dataframe joins, where we have possibly millions of very small results.  Saving thousands of tiny files is *very* slow on most file systems.  [See blogpost](http://matthewrocklin.com/blog/work/2015/12/29/disk-bandwidth).

Single-machine key-value store databases provide decently efficient random write and read access to small values on disk.  There are *many* solutions here.  I'm somewhat partial to embedded databases like LevelDB (which now has many more modern competitors) that run inside the same process because they don't require significant setup, run in a well contained way, and are easy to use.

So one solution here would be to optionally replace the `File(MutableMapping)` from zict that uses the filesystem with a MutableMapping that was backed by some more efficient embedded database.  This would require learning a bit about different key-value stores and choosing a good one based on the many-small writes/reads workloads that dask.distributed might push onto it, finding a way to combine it with the LRU scheme (or some other scheme) used currently (this is easy if you can make it follow a `MutableMapping` interface), and then hooking up the necessary keywords to allow users to access this option easily.

pitrou:
This needs https://github.com/dask/zict/pull/5 and https://github.com/dask/zict/pull/6 to be merged first.

Fixes #525.


distributed/worker.py
```diff
@@ -129,11 +129,18 @@ def __init__(self, scheduler_ip, scheduler_port, ip=None, ncores=None,
...
             try:
-                from zict import Buffer, File, Func
+                from zict import Buffer, File, Func, LMDB, Sieve
             except ImportError:
                 raise ImportError("Please `pip install zict` for spill-to-disk workers")
             path = os.path.join(self.local_dir, 'storage')
-            storage = Func(dumps_to_disk, loads_from_disk, File(path))
+            file_storage = Func(dumps_to_disk, loads_from_disk, File(path))
+            lmdb_path = os.path.join(self.local_dir, 'lmdb')
+            lmdb = Func(dumps_to_disk, loads_from_disk, LMDB(lmdb_path))
```
pitrou: I'm mostly being cautious here, in case LMDB ends up slowish for very large values on some systems (Windows? special filesystems? who knows :-)).

distributed/worker.py
```diff
@@ -129,11 +129,18 @@ def __init__(self, scheduler_ip, scheduler_port, ip=None, ncores=None,
...
+            def sieve_func(k, v):
+                return 'small' if sys.getsizeof(v) < 32 * 1024 else 'large'
```
pitrou: Ok, I'll fix this.

pitrou:
There are weird build stalls on Travis, I wonder what's happening exactly.

mrocklin:
I have not seen that issue before

distributed/tests/test_worker.py
```diff
@@ -521,6 +521,37 @@ def test_spill_to_disk(e, s):
...
+
+    w = Worker(s.ip, s.port, loop=s.loop, memory_limit=100)
+    yield w._start()
+
+    x = e.submit(np.random.randint, 0, 255, size=500, dtype='u1', key='x')
+    yield _wait(x)
+    y = e.submit(np.random.randint, 0, 255, size=500, dtype='u1', key='y')
+    yield _wait(y)
+    z = e.submit(np.random.randint, 0, 255, size=1000000, dtype='u1', key='z')
+    yield _wait(z)
```
mrocklin: FYI you can wait on multiple futures at once

```python
x = 
y = 
z = 
yield _wait([x, y, z])
```

distributed/tests/test_worker.py
```diff
@@ -521,6 +521,37 @@ def test_spill_to_disk(e, s):
...
+    assert set(w.data.slow.mappings['small']) == {x.key, y.key}
+    assert set(w.data.slow.mappings['large']) == {z.key}
+
+    def get_files(path):
+        print(os.listdir(path))
+        return [s for s in os.listdir(path)
+                if os.path.isfile(os.path.join(path, s))]
+
+    files = get_files(os.path.join(w.local_dir, 'storage'))
+    assert files == ['z']
```
mrocklin: I don't understand this.  There is a single file named `z`?

distributed/worker.py
```diff
@@ -129,11 +129,18 @@ def __init__(self, scheduler_ip, scheduler_port, ip=None, ncores=None,
...
+                from zict import Buffer, File, Func, LMDB, Sieve
             except ImportError:
                 raise ImportError("Please `pip install zict` for spill-to-disk workers")
             path = os.path.join(self.local_dir, 'storage')
-            storage = Func(dumps_to_disk, loads_from_disk, File(path))
+            file_storage = Func(dumps_to_disk, loads_from_disk, File(path))
+            lmdb_path = os.path.join(self.local_dir, 'lmdb')
+            lmdb = Func(dumps_to_disk, loads_from_disk, LMDB(lmdb_path))
+            def sieve_func(k, v):
+                return 'small' if sizeof(v) < 32 * 1024 else 'large'
```
mrocklin: We might want to increase the size of this.  32k might still be fairly small.  One way to think about this is "how many files is a gigabyte".  In worst case here we get 30000 files in a directory, which I'm pretty sure will negatively affect the file system.

distributed/tests/test_worker.py
pitrou: Oh, thanks.

distributed/tests/test_worker.py
pitrou: Yes. `x` and `y` are stored in the LMDB store, only `z` is too large.

distributed/worker.py
pitrou: You're right. I think upping it to ~10 MB is fair.

pitrou:
Ok, the current status is that HDFS tests hang: https://travis-ci.org/dask/distributed/jobs/172615019#L3414
A priori, we cannot rule out a possible nasty interaction between HDFS and LMDB. How easy is it to set up HDFS on a home computer?

mrocklin:
Assuming that you're on Linux and don't mind using Docker

    cd distributed/continuous_integration
    docker build .

pitrou:
Oh, I hadn't noticed the README in the `continuous_integration` directory. It turns out surprisingly easy, thanks :)

mrocklin:
Thank @danielfrg

pitrou:
Right, the problem appears to be simply that some dependencies are missing on the Docker container.

pitrou:
The Travis tests now finally pass, except for one intermittent failure in an `rpc.active` check.

pitrou:
The Travis tests now finally pass, except for one intermittent failure in an `rpc.active` check.

pitrou:
How do we handle the added dependency on `python-lmdb`? Do we add it explicitly somewhere?

mrocklin:
We probably leave it as purely optional for now.  We should include a note in the installation documentation.  We might consider making another conda or PyPI package like `distributed-full` in the future.

pitrou:
Ok, I've updated the docs. Also fixed some missing links.

docs/source/install.rst
```diff
@@ -31,6 +31,19 @@ To install distributed from source, clone the repository from `github
...
+---------------------
+
+To use the :ref:`spill-to-disk <spill-to-disk>` feature, you'll need
+to install the Python bindings to the LMDB database.  Using ``conda``::
+
+   conda install -c conda-forge python-lmdb
+
+Or using ``pip``::
+
+   pip install lmdb
```
mrocklin: Does this include the system library as well?

docs/source/protocol.rst
```diff
@@ -175,7 +175,7 @@ For administrative messages like updating status msgpack is sufficient.
 However for large results or Python specific data, like NumPy arrays or Pandas Dataframes, or
 for larger results we need to use something else to convert Python objects to
 bytestrings.  Exactly how we do this is described more in the
-`:doc:Serialization documentation <serialize.rst>`.
+:doc:`Serialization documentation <serialization>`.
```
mrocklin: Thank you for cleaning up my mess

.travis.yml
```diff
@@ -34,7 +34,7 @@ before_install:
         popd
         docker run -d -p 8020:8020 -p 50070:50070 -v $(pwd):/distributed distributed-hdfs
         export CONTAINER_ID=$(docker ps -l -q)
-        sleep 60  # Wait for namenode and datanode
+        sleep 10  # Wait for namenode and datanode
```
mrocklin: Travis.ci can be very slow.  Even at 60s this is still the source of some intermittent failures.  In the future it would be nice to poll instead.

pitrou: Ok, I'll restore the 60s timeout then :-)

mrocklin:
I've merged master into this branch

A quick benchmark on a small dataset actually shows a performance regression from using LMDB over just the file system


```
dask-scheduler
dask-worker localhost:8786 --nthreads 4 --memory-limit 1000000
```
```python
c = Client('localhost:8786')

x = da.random.random((20000, 20000), chunks=(20000, 20))
y = c.persist(x)
```

### On Master
```python
%time z = c.persist(y.rechunk((1000, 1000)).rechunk((20, 20000))); wait(z)
CPU times: user 5.6 s, sys: 268 ms, total: 5.87 s
Wall time: 1min 37s
```

### LMDB branch
```python
%time z = c.persist(y.rechunk((1000, 1000)).rechunk((20, 20000))); wait(z)
CPU times: user 5.57 s, sys: 52 ms, total: 5.62 s
Wall time: 3min 37s
```

However, this dataset also fits comfortably in RAM, so the file system is probably doing a lot of buffering against disk here.

mrocklin:
This was 3GB spread over around 40k tasks and two full copy passes, so there were around 20k tasks on disk at once.

pitrou:
> However, this dataset also fits comfortably in RAM, so the file system is probably doing a lot of buffering against disk here.

Hmm, when running this benchmark here (8 GB RAM machine, of which ~6 available), it does a massive amount of disk I/O. Both reads and writes. I'll post the numbers when I finish running it.

pitrou:
Here are the results. I also measured the total disk I/O for the process using `psutil`.

### LMDB
```
Process read 40.0 GB, wrote 13.2 GB

real	6m52.276s
user	3m4.576s
sys	0m23.204s
```

### File
```
Process read 2.1 GB, wrote 21.2 GB

real	2m31.277s
user	1m10.764s
sys	0m29.044s
```

It's interesting that LMDB writes less data but reads a *lot* more. Since reads are typically blocking while writes can be performed in the background, this should explain the runtimes. A pity: watching `vmstat` shows LMDB making good use of the SSD's bandwidth (achieving 150 to 300 MB/s).

As for the much larger user CPU time for LMDB, I'm not sure, but that may be LMDB's bookkeeping overhead under high stress.

(note: I'm running client, scheduler and workers in the same process here)

pitrou:
This seems to show that the filesystem (depending on OS specifics -- I'm using ext4) isn't as bad as we assumed it be, and LMDB has performance characteristics that are not as favourable as we hoped.

mrocklin:
CPU overhead might also be compression?

I'll try to benchmark a bit more in the future.  I'm curious to see how the file system performs on larger computations.  

Are there any relevant settings within LMDB that we should check?

pitrou:
> CPU overhead might also be compression?

I don't think LMDB has built-in compression. If it's dask compression, then it should be the same overhead in the two cases, no?

> Are there any relevant settings within LMDB that we should check?

The only relevant one seems to be `writemap`. I've tried to force it to False, but that didn't change the numbers significantly.
See https://lmdb.readthedocs.io/en/release/#environment-class

mrocklin:
This seems to be not as important as we thought.  Closing.

PR #613, opened 2016-11-01
Pass extra LocalCluster arguments to Worker

pitrou:
This seems to be an oversight -- `LocalCluster` accepts arbitrary additional arguments but didn't do anything with them.
Also fixes a typo in tests (`diagnostic_port` isn't a recognized argument, `diagnostics_port` is).


mrocklin:
+1

PR #620, opened 2016-11-02
Add h5py custom serialization

mrocklin:
This allows H5Py dataset objects to be used in dask.distributed, despite not being picklable.  We use the recent custom serialization change in #606 to encode a dataset object as the filename and datapath.  

This is a proof of concept for solving https://github.com/pydata/xarray/issues/798

cc @shoyer if would be good to get your input on this before this weekend

### DIFF
--- a/distributed/protocol/h5py.py
+++ b/distributed/protocol/h5py.py
@@ -1,0 +1,40 @@
+from __future__ import print_function, division, absolute_import
+
+from .serialize import register_serialization
+
+from ..utils import log_errors, ensure_bytes
+
+
+def serialize_h5py_file(f):
+    if f.mode != 'r':
+        raise ValueError("Can only serialize read-only h5py files")
+    return {'filename': f.filename}, []
+
+
+def deserialize_h5py_file(header, frames):
+    import h5py
+    return h5py.File(header['filename'], mode='r')
+
+
+register_serialization('h5py._hl.files.File', serialize_h5py_file,
+                       deserialize_h5py_file)
+
+
+
+def serialize_h5py_dataset(x):
+    header, _ = serialize_h5py_file(x.file)
+    header['name'] = x.name
+    return header, []
+
+
+def deserialize_h5py_dataset(header, frames):
+    import h5py
+    file = deserialize_h5py_file(header, frames)
+    return file[header['name']]
+
+
+register_serialization('h5py._hl.dataset.Dataset', serialize_h5py_dataset,
+                       deserialize_h5py_dataset)
+
+register_serialization('h5py._hl.group.Group', serialize_h5py_dataset,
+                       deserialize_h5py_dataset)
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -44,7 +44,10 @@ def register_serialization(cls, serialize, deserialize):
     serialize
     deserialize
     """
-    name = typename(cls)
+    if isinstance(cls, type):
+        name = typename(cls)
+    elif isinstance(cls, str):
+        name = cls
     serializers[name] = serialize
     deserializers[name] = deserialize
 
--- a/distributed/protocol/tests/test_h5py.py
+++ b/distributed/protocol/tests/test_h5py.py
@@ -1,0 +1,81 @@
+import pytest
+
+h5py = pytest.importorskip('h5py')
+
+from distributed.protocol import serialize, deserialize, dumps, loads
+
+from distributed.utils import tmpfile
+
+
+def test_serialize_deserialize_file():
+    with tmpfile() as fn:
+        with h5py.File(fn, mode='a') as f:
+            f.create_dataset('/x', shape=(2, 2), dtype='i4')
+        with h5py.File(fn, mode='r') as f:
+            g = deserialize(*serialize(f))
+            assert f.filename == g.filename
+            assert isinstance(g, h5py.File)
+            assert f.mode == g.mode
+
+            assert g['x'].shape == (2, 2)
+
+
+def test_serialize_deserialize_group():
+    with tmpfile() as fn:
+        with h5py.File(fn, mode='a') as f:
+            f.create_dataset('/group1/group2/x', shape=(2, 2), dtype='i4')
+        with h5py.File(fn, mode='r') as f:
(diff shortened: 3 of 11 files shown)


mrocklin:
@jcrist I'm running into issues here with `extract_serialize` (previously `extract_big_bytes`) because calling deepcopy on things often ends up calling pickle.dumps on them and now I'm trying to avoid calling pickle on things.  My planned solution is to make a deepcopy function that dives through lists, dicts, sets, tuples, etc..  Do you see any problem with this approach?

distributed/protocol/h5py.py
```diff
@@ -0,0 +1,33 @@
...
+
+import h5py
+
+from .serialize import register_serialization
+
+from ..utils import log_errors, ensure_bytes
+
+
+def serialize_h5py_file(f):
+    return {'filename': f.filename, 'mode': f.mode}, []
```
mrocklin: Alright, I should change then to only work with read-only files?

shoyer: That would be my inclination, yes. Require `mode in {'r', 'r+'}`.

But what then do you do with writeable files to ensure they don't get passed around?

h5py does have support for "parallel HDF5" using MPI but it seems a good deal more complex (I've never used it):
http://docs.h5py.org/en/latest/mpi.html

jcrist:
> Do you see any problem with this approach?

No, that sounds fine to me.

mrocklin:
OK, I've switched out `deepcopy` for `container_copy` (ping @jcrist for review)

I'm now erring whenever mode isn't strictly `'r'`.  h5py seems to turn everything else into `'r+'`.

Tests pass locally.

distributed/protocol/serialize.py
```diff
@@ -159,6 +160,54 @@ def __init__(self, header, frames):
         self.frames = frames
 
 
+def container_copy(c):
+    if isinstance(c, list):
```
jcrist: I might turn these into explicit type checks (`if type(c) is list` as done in `_extract_serialize`) for a slight speedup. Up to you. Otherwise, this looks good to me.

distributed/protocol/h5py.py
jakirkham: Why not just restrict it to reading outright? It seems like `r+` has the same risks attached as a writable files.

mrocklin:
@shoyer  I've added support and minimal testing for groups.  Review appreciated.

shoyer:
netCDF group support looks good to me.

jakirkham:
So I understand the value of serializing/deserializing datasets. However, I don't really understand the use case for serializing/deserializing groups. Could you please explain where you plan to use this a bit more?

mrocklin:
Its for completeness and to help serialize variables within groups.  Do you
have a particular concern you'd like to express?

mrocklin:
I would like to merge this soon.  @shoyer other than concerns about import names is there anything you think should be addressed here before a merge?

shoyer:
Indeed, my only concern is using private import names. You can fix later, but it would be good to do before you consider this feature "stable". Otherwise, looks good to me.

mrocklin:
I've rebased this into three commits.  Will merge into master once tests pass.

jakirkham:
My main concern is we are moving in the direction of serializing file objects, which I also see in the code. Unlike datasets it is less clear when we might be done with these. So we don't know if we can close the underlying files. Though it is unclear to me that we are actually closing the underlying files in those cases either.

This is important as `h5py.File` objects don't close the file when all references to them are removed. Instead they expect the user to close them when they are done. Though `h5py` will try to make some effort to close these lingering files when the process is done. In the interim, we are keeping a lot of lingering open files around that we don't need.

mrocklin:
Yes I agree that opening HDF5 files without a plan to close them is unfortunate.  However at the moment the ability to open a few HDF5 files locally and have those same HDF5 files open elsewhere as needed is very pragmatic, which I believe overweighs the lack of cleanliness. 

However, if you have thoughts for a better solution and the time and energy to implement then that would be very welcome.  Otherwise I'm planning to merge this soon.

jakirkham:
Sorry I updated my comment above. Recent versions of `h5py` (`2.4.0`+) automatically close files for us. So this should be fine.

I'm less familiar with the other file formats here like NetCDF. So can't comment on how those behave.

Generally I agree this is something to iterate on. Was merely concerned that we trap ourselves by exposing objects that we don't need yet and would be significantly hard to clean up manually. It seems better for us to expose the subset we more often need datasets, variables, etc. Those should prove easier to cleanup if we have to do this manually later.

jakirkham:
Do you have any idea as to when this might show up in a release? Would be very interested in playing with it once it does.

mrocklin:
I plan to do a micro release later this week

jakirkham:
That sounds great! Thanks. Please keep me posted.

distributed/protocol/tests/test_h5py.py
```diff
@@ -0,0 +1,81 @@
...
+def test_h5py_serialize(c, s, a, b):
+    with tmpfile() as fn:
+        with h5py.File(fn, mode='a') as f:
+            x = f.create_dataset('/group/x', shape=(4,), dtype='i4',
+                                 chunks=(2,))
+            x[:] = [1, 2, 3, 4]
+            f.flush()
+        with h5py.File(fn, mode='r') as f:
+            dset = f['/group/x']
+            x = da.from_array(dset, chunks=x.chunks)
```
jakirkham: Should this be `dset.chunks`?

jakirkham: Please see PR ( https://github.com/dask/distributed/pull/652 ).

jakirkham:
Are there changes needed from `dask` to use this functionality? My reason for asking is I tried to do a simple operation that crashed and am not sure if it is worth a bug report at this stage or not.

mrocklin:
Everything should be fine from the most recent release of dask.  But I
could be wrong and a bug report would be welcome.

jakirkham:
Ok, I came up with a simple reproducer and put as much info as seemed relevant in issue ( https://github.com/dask/dask/issues/1777 ). Please let me know if you have any questions or suggestions.

PR #623, opened 2016-11-02
Use a default StreamHandler for all loggers

dhirschfeld:
- Remove use of basicConfig which can affect the root logger

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -360,7 +360,7 @@ def str_graph(dsk, extra_values=()):
 import logging
 from .compatibility import logging_names
 
-handler = logging.StreamHandler(sys.stdout)
+handler = logging.StreamHandler(sys.stderr)
 handler.setLevel(logging.INFO)
 fmt = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
 handler.setFormatter(fmt)
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -360,7 +360,7 @@ def str_graph(dsk, extra_values=()):
 import logging
 from .compatibility import logging_names
 
-handler = logging.StreamHandler()
+handler = logging.StreamHandler(sys.stdout)
 handler.setLevel(logging.INFO)
 fmt = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
 handler.setFormatter(fmt)
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -359,12 +359,17 @@ def str_graph(dsk, extra_values=()):
 
 import logging
 from .compatibility import logging_names
-logging.basicConfig(format='%(name)s - %(levelname)s - %(message)s',
-                    level=logging.INFO)
+
+handler = logging.StreamHandler()
+handler.setLevel(logging.INFO)
+fmt = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
+handler.setFormatter(fmt)
 
 for name, level in config.get('logging', {}).items():
     LEVEL = logging_names[level.upper()]
-    logging.getLogger(name).setLevel(LEVEL)
+    logger = logging.getLogger(name)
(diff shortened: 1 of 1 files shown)


mrocklin:
The stream handler apparently isn't dumping out to stdout, hence the error.

dhirschfeld:
The builds which failed all seemed to hang on `test_bokeh` - not sure if that's related?

```
distributed/cli/tests/test_dask_scheduler.py::test_bokeh 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

distributed/utils.py
```diff
@@ -359,12 +359,17 @@ def convert(task):
 
 import logging
 from .compatibility import logging_names
-logging.basicConfig(format='%(name)s - %(levelname)s - %(message)s',
-                    level=logging.INFO)
```
dhirschfeld: AFAICS it's not a method on a logging instance?

mrocklin: Oh well.  Hrm, perhaps this is intended to be global and you're expected to just overwrite other people?  Perhaps call basicConfig yourself after importing `dask.distributed`?

My next step here would be to ask on stackoverflow what the right way to do this would be.

mrocklin:
The `test_bokeh` test is waiting on logging information to come out of stdout or stderr.  Apparently the StreamLogger isn't dumping to stdout or stderr.

dhirschfeld:
I tried passing `sys.stdout` to the StreamHandler - we'll see what travis thinks of that...

pitrou:
@dhirschfeld, you can probably install bokeh and run the tests locally if you don't want to wait for Travis every time.

distributed/utils.py
```diff
@@ -359,12 +359,17 @@ def convert(task):
...
+
+handler = logging.StreamHandler(sys.stderr)
+handler.setLevel(logging.INFO)
+fmt = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
+handler.setFormatter(fmt)
 
 for name, level in config.get('logging', {}).items():
```
pitrou: By default, this won't configure any logger, though? You should add a default configuration for the "distributed" logger, it will be inherited by any child loggers (any "distributed.FOO.BAR").

distributed/utils.py
```diff
@@ -359,12 +359,17 @@ def convert(task):
...
     LEVEL = logging_names[level.upper()]
-    logging.getLogger(name).setLevel(LEVEL)
+    logger = logging.getLogger(name)
+    logger.setLevel(LEVEL)
+    logger.addHandler(handler)
 
 # http://stackoverflow.com/questions/21234772/python-tornado-disable-logging-to-stderr
 stream = logging.StreamHandler(sys.stderr)
```
pitrou: Logging only critical messages seems heavy-handed. This will even silence exceptions when `add_callback` callbacks are called. I would recommend setting the level to ERROR instead.

distributed/utils.py
pitrou: My point is it doesn't configure any logger if you have nothing in the logging configuration.

dhirschfeld: Right, you probably wouldn't ever want all output to be silenced. I'm assuming that the test is failing because no loggers are configured so I've pushed a further change to fall back to `basicConfig` in that instance. We'll see what travis thinks of that...

I've just been using the GitHub web-interface so its easier for me to simply wait for travis. I can squash the commits if desired when the test pass.

pitrou:
@dhirschfeld, if you are struggling with this PR, do you want me to take over?

dhirschfeld:
Please do! I'm at a bit of a loss as to how to both solve my problem and
not break the tests.

There's not too much here so if it's easier to close this and start afresh
feel free to do so also...

pitrou:
I'm going close this PR as obsoleted by #645.

PR #626, opened 2016-11-03
bump version to 1.14.0

mrocklin:


mrocklin:
I've run into a message passing regression from the serialization work.  Holding off on this for a bit.

PR #628, opened 2016-11-06
Update bokeh plots

mrocklin:
*  Fix size of memory plot
*  Add height of task stream plot (maybe not a good idea)

### DIFF
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -88,7 +88,10 @@ class TaskStream(DashboardComponent):
         )
 
         self.root.add_tools(
-            hover, ResetTool(), PanTool(dimensions="width"), WheelZoomTool(dimensions="width")
+            hover,
+            ResetTool(reset_size=False),
+            PanTool(dimensions="width"),
+            WheelZoomTool(dimensions="width")
         )
 
         # Required for update callback
@@ -392,7 +395,7 @@ class WorkerTable(DashboardComponent):
         mem_plot = Plot(
             title=Title(text="Memory Usage (%)"), toolbar_location=None,
             x_range=Range1d(start=0, end=1), y_range=Range1d(start=-0.1, end=0.1),
-            plot_height=80, **kwargs
+            **kwargs
         )
 
         mem_plot.add_glyph(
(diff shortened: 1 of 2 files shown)


mrocklin:
cc @canavandl can you give this four-line PR a quick glance to check if anything seems off to you?

distributed/bokeh/components.py
```diff
@@ -402,6 +402,7 @@ def __init__(self, **kwargs):
 
         mem_plot.add_layout(LinearAxis(), 'left')
         mem_plot.add_layout(LinearAxis(), 'below')
+        mem_plot.yaxis.visible = False
```
canavandl: remove L404 instead

canavandl:
What's the intent of settings the plot_height in the element constructor instead of passing it in as a kwarg? I think doing so might impact the responsive resizing for JLab

mrocklin:
Currently in the workers/ pane the `mem_plot` plot is very large.  We want it to take up only a small sliver of space.

Similarly we're getting strange effects when the user presses "reset" on the task stream plot in the status/ page.  The plot increases in vertical size.

canavandl:
I think you should be able to constrain the mem_plot size by passing ``plot_height=300`` here: https://github.com/dask/distributed/blob/master/distributed/bokeh/workers/main.py#L21

I'll look into the task stream plot.

canavandl:
It appears that ResetTool accepts a `reset_size` kwarg bool. While you described behavior sounds like a Bokeh bug, it may be resolved by adding ``reset_size=False`` to the ResetTool constructor in: https://github.com/dask/distributed/blob/master/distributed/bokeh/components.py#L91

mrocklin:
Changes applied.  Lookin' sharp.

distributed/bokeh/components.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, n_rectangles=1000, **kwargs):
         self.root = Plot(
             title=Title(text="Task Stream"), id='bk-task-stream-plot',
             x_range=x_range, y_range=y_range, toolbar_location="above",
-            min_border_right=35, **kwargs
+            min_border_right=35, plot_height=300, **kwargs
```
canavandl: also pass as kwarg here?

mrocklin: I've removed it entirely.  This was only here as a hack to avoid the reset issue

canavandl:
LGTM

PR #636, opened 2016-11-08
Numpy dtypes

### LINKED ISSUE #635: Cannot serialize datetime64/timedelta64 numpy arrays
```
distributed.core - INFO - Unserializable Message: {'finalize-c29ed21fd1e79d8eb6b3ff89cf2ae354': <Serialize: ['2010-01-01T00:00:00.000000000' '2010-01-02T00:00:00.000000000'
 '2010-01-03T00:00:00.000000000']>}
distributed.core - ERROR - cannot include dtype 'M' in a buffer
  File "/Users/shoyer/dev/distributed/distributed/core.py", line 228, in write
    frames = protocol.dumps(msg)
  File "/Users/shoyer/dev/distributed/distributed/protocol/core.py", line 43, in dumps
    for key, value in data.items()
  File "/Users/shoyer/dev/distributed/distributed/protocol/core.py", line 44, in <dictcomp>
    if type(value) is Serialize}
  File "/Users/shoyer/dev/distributed/distributed/protocol/serialize.py", line 103, in serialize
    header, frames = serializers[name](x)
  File "/Users/shoyer/dev/distributed/distributed/protocol/numpy.py", line 58, in serialize_numpy_ndarray
    frames = [x.data]
ValueError: cannot include dtype 'M' in a buffer
```

The buffer protocol fails on these (sigh). You can view as int64, but then you'll need to keep track of the original type (including time units) separately.

This came up in some of my integration tests with xarray.

mrocklin:
Fixes #635  

cc @shoyer

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -47,15 +47,17 @@ def serialize_numpy_ndarray(x):
               'strides': x.strides,
               'shape': x.shape}
 
+    data = x.view('u1').data
+
     if blosc:
-        frames = frame_split_size([x.data])
+        frames = frame_split_size([data])
         if sys.version_info.major == 2:
             frames = [ensure_bytes(frame) for frame in frames]
         frames = [blosc.compress(frame, typesize=size,
                                  cname='lz4', clevel=5) for frame in frames]
         header['compression'] = ['blosc'] * len(frames)
     else:
-        frames = [x.data]
+        frames = [data]
 
     header['lengths'] = [x.nbytes]
 
--- a/distributed/protocol/tests/test_numpy.py
+++ b/distributed/protocol/tests/test_numpy.py
@@ -32,6 +32,8 @@ def test_serialize():
          np.random.random(5).astype('f4'),
          np.arange(5).astype('M8[us]'),
          np.arange(5).astype('M8[ms]'),
+         np.arange(5).astype('m8'),
+         np.arange(5).astype('m8[s]'),
          np.arange(5).astype('c16'),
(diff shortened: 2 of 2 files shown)


mrocklin:
Added m8 tests.  Thanks @jreback .  Any other suggestions?

distributed/protocol/numpy.py
```diff
@@ -47,15 +47,17 @@ def serialize_numpy_ndarray(x):
               'strides': x.strides,
               'shape': x.shape}
 
+    data = x.view('u1').data
```
shoyer: I'm a little nervous about using view with a different dtype itemsize. See this warning from the docs:

> For a.view(some_dtype), if some_dtype has a different number of bytes per entry than the previous dtype (for example, converting a regular array to a structured array), then the behavior of the view cannot be predicted just from the superficial appearance of a (shown by print(a)). It also depends on exactly how a is stored in memory. Therefore if a is C-ordered versus fortran-ordered, versus defined as a slice or transpose, etc., the view may give different results.

It might be just fine for merely getting access to the `data` buffer, though.

jreback:
numpy scalars

e.g. np.array(5)

jreback:
*maybe* big/little endian

e.g. >i8 and <i8

distributed/protocol/numpy.py
mrocklin: Yeah, I think that these would all be tricky situations if the itemsizes were relatively prime or if you actually cared about indexing afterwards. I think we're safe turning it into a view of singleton bytes though.

PR #639, opened 2016-11-08
Add export tool to Task Stream plot

mrocklin:


mrocklin:
OK, I could use some help here from anyone with some JS or CoffeeScript experience.  

The goal is to build a custom Bokeh tool that, when clicked asks the server to construct static standalone HTML and then deliver that HTML to the user through the client.

What's here is apparently visible in Firefox and Safari but not visible in Chrome.  In Firefox we can then save the resulting page but it comes down as separate HTML, js, and css files.

A couple of people have recommended constructing a link with an octet stream and then clicking it.  That might be a nice way to go.  Unfortunately I personally have no idea how to accomplish this in coffee script.  Help would be appreciated if anyone has time.

mrocklin:
If anyone wants to test this locally:
conda install -c bokeh bokeh
git clone git@github.com:mrocklin/distributed.git
cd distributed
git checkout export-tool
python setup.py install

ipython
c = Client()  # starts up the bokeh server and a small "cluster" on your laptop
```

Navigate to http://127.0.0.1:8787/status and then click the save icon on the center plot.

**Comment by canavandl** [comment]
```
...
  export: () ->
    if @model.content?
      blob = new Blob([@model.content], {type: "text/plain"})
      url = window.URL.createObjectURL(blob);

      a = document.createElement("a")
      a.id = "bk-export-tool-link"
      a.style = "display: none"
      a.href = url
      a.download = 'plot.html'
      document.body.appendChild(a)
      a.click()

      document.getElementById('bk-export-tool-link').remove()
      window.URL.revokeObjectURL(url);
```

mrocklin:
This appears to work like a charm on my machine.  Nice @canavandl !

canavandl:
You may wan to rename the ``a.download`` attribute to something better than "plot.html", so the downloaded filename is something more helpful.

canavandl:
Also doesn't seem to work on FireFox or Safari. Let me look into it.

mrocklin:
I'm tempted to merge this as is.  I could really use this feature and I suspect we'd find other issues during experimentation.

canavandl:
Go for it. And we can open an issue to track the browser problem

PR #640, opened 2016-11-08
Add --no-reconnect flag to dask-worker

mrocklin:
Usually when a worker loses communication with the scheduler it sticks around trying to reconnect.  This adds an option to workers, nannies, and the dask-worker CLI to avoid this feature if desired.  

This tends to be useful in some cluster deployments when you want the workers to clean themselves up gracefully when the scheduler dies.

    dask-worker scheduler:8786 --no-reconnect


jreback:
ohh this is nice

so one could bounce the scheduler and everything should die?
even if it's the same port ?

mrocklin:
Yes, if you destroy the scheduler then the workers should die on their own.  The choice of port should not have an effect.

Issue #614, opened 2016-11-01
Efficient Pandas serialization


mrocklin:
We need an efficient way to serialize Pandas Dataframes.  As of #606 we can now customize this beyond just pickle.  There are a number of different data regimes here including pure numeric data, very compressible numeric data like time series, text data with repeats, categoricals, long text data etc..  We care both about fast encoding and about fast compression for larger results.  We want something that has minimal overhead on small dataframes (important for shuffling).  

Several options come to mind:

1.  Pickle + lz4: what we use now.  Actually not that bad.  Pickle deduplicates text data, handles numeric data without too much overhead, etc..  It's a shame that we can't compress per type.
2.  NumPy serialization + blosc on blocks + pickle or msgpack + lz4 on text: This is similar to what we do in partd.  A custom approach can get significant speed boosts in some cases.
3.  Arrow: we could give arrow a shot.  I suspect that it handles encoding intelligently.  It would be nice if it were to provide different compressions per column but we could probably push a bit upstream if this went into use.
4.  Other thoughts?

@jreback @wesm @shoyer 

It would be good to make this decision with a benchmark in hand.  It would be good both to get people's opinions on solutions we should consider as well as some benchmarks that are representative of data that they care about.

wesm:
I've spent a lot of time lately working on DataFrame serialization with as little overhead as possible. I'm happy to assist as best I can with development to support this use case and benchmarking to support the decision making. I recently did some rudimentary benchmarking on uncompressed numeric data, see slide 24 in 

http://www.slideshare.net/wesm/python-data-wrangling-preparing-for-the-future

the benchmarking code (which I can clean up and publish if so desired) looks like:

```
import pyarrow as pa
import pyarrow.io as io
import pyarrow.ipc as ipc

def write_arrow(df):
    batch = pa.RecordBatch.from_pandas(df)

    buf = BytesIO()
    writer = ipc.ArrowFileWriter(buf, batch.schema)
    writer.write_record_batch(batch)
    writer.close()
    return buf.getvalue()


def read_arrow(buf):
    reader = ipc.ArrowFileReader(buf)
    return [reader.get_record_batch(i).to_pandas()
            for i in range(reader.num_record_batches)]
```

[...]

PR #643, opened 2016-11-09
WIP: Use Arrow for Pandas serialization

mrocklin:
This adds experimental support for custom Pandas serialization using Arrow.

There is some missing functionality here.  I'm not sure if this is due to things missing within Arrow or my naive use of arrow (I'm just copy-pasting a solution from #614 ):

1.  Indexes
2.  Categoricals
3.  Series without an extra copy

cc @wesm

### DIFF
--- a/distributed/protocol/pandas.py
+++ b/distributed/protocol/pandas.py
@@ -1,0 +1,48 @@
+from __future__ import print_function, division, absolute_import
+
+from .serialize import register_serialization
+
+import pandas as pd
+from io import BytesIO
+
+import pyarrow as pa
+import pyarrow.io as io
+import pyarrow.ipc as ipc
+
+
+def serialize_pandas_dataframe(df):
+    batch = pa.RecordBatch.from_pandas(df)
+    buf = BytesIO()
+    writer = ipc.ArrowFileWriter(buf, batch.schema)
+    writer.write_record_batch(batch)
+    writer.close()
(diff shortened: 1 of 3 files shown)


distributed/protocol/pandas.py
```diff
@@ -0,0 +1,48 @@
...
+                       deserialize_pandas_dataframe)
+
+
+def serialize_pandas_series(s):
+    return serialize_pandas_dataframe(s.to_frame())  # unfortunate copy here
+
+
+def deserialize_pandas_series(header, frames):
+    df = deserialize_pandas_dataframe(header, frames)
+    return df[df.columns[0]]  # unfortunate copy here
```
jreback: this is not a copy :)

distributed/protocol/pandas.py
```diff
@@ -0,0 +1,48 @@
...
+            for i in range(reader.num_record_batches)][0]  # what is each batch?
+
+
+register_serialization(pd.DataFrame,
+                       serialize_pandas_dataframe,
+                       deserialize_pandas_dataframe)
+
+
+def serialize_pandas_series(s):
+    return serialize_pandas_dataframe(s.to_frame())  # unfortunate copy here
```
jreback: this is also not s copy iirc (if i am not remembering correctly it's easy to make this not a copy)

jreback:
for 1) you can do preserve index by
.reset_index() and record the column

if u want

PR #644, opened 2016-11-09
(WIP) Make filesystem wrapper for hdfs to match changes in dask bytes

martindurant:
Tests relying on previous functions will fail - should all be calling dask.bytes.core now.

### DIFF
--- a/continuous_integration/Dockerfile
+++ b/continuous_integration/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:trusty
+FROM ubuntu:14.04
 
 # conda
 RUN apt-get update && apt-get install -y -q curl bzip2 git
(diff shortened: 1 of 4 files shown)


mrocklin:
Are you resolving these tests or should I?  (happy to do it)

martindurant:
I would appreciate some help here

mrocklin:
OK, almost everything works.  We do have an issue though because HDFS isn't thread or process safe.  Previously we used a lock file (locket) in the `read_bytes` function.  Now that we're depending on open files I'm not sure how best to handle this.  Thoughts welcome.

martindurant:
I remember the thread-unsafeness but process too? If thread, we can keep a lock around (in the HDFileSystem class?), and grab this in `OpenFile.__enter__`?

mrocklin:
Process too.  If you look through the diff for `locket` you'll find the missing code.

It might not be just opening a file.  It might also be starting the HDFileSystem class.

To this end we might need to make a special open that also opens up the file system, rather than passing along the method.  Or perhaps we lock both OpenFile.__enter__ and the `__setstate__` method of the HDFileSystem so that it reinitializes safely?

martindurant:
Didn't we use to have a stress test against hdfs3 to look for this sort of thing?

mrocklin:
Perhaps?  The current tests themselves expose the issue though.

martindurant:
Locket added to libhadfs3 calls: https://github.com/dask/hdfs3/pull/89

martindurant:
The failures are because of the changes not yet being merged on the dask side

mrocklin:
Lets hold off on merging this until I make a micro-release for dask/distributed.

Issue #599, opened 2016-10-20
Intermittent testing hangs


mrocklin:
These show up infrequently
- [ ] distributed/tests/test_client.py::test_iterator_gather
- [ ] distributed/tests/test_joblib.py::test_simple
- [ ] distributed/tests/test_ipython.py::test_start_ipython_scheduler 
- [ ] distributed/tests/test_stress.py::test_stress_scatter_death 
- [ ] distributed/deploy/tests/test_adaptive.py::test_adaptive_local_cluster_multi_workers

pitrou:
I managed to reproduce the test_joblib under the pytest-timeout plugin. Here is the output (log + stack traces). Weirdly, it looks like there's a deadlock in Tornado's IO loop over the callback lock...
https://gist.github.com/pitrou/c406b81a4c0cc0e04335604ec9443029

pitrou:
(method to reproduce:
`py.test -v -x --timeout-method=thread --timeout=6 --count=1000 distributed/tests/test_joblib.py`
you'll need the pytest-timeout and pytest-repeat plugins)

mrocklin:
Didn't know you could get that kind of output.  That's handy.  

@ogrisel do custom joblib backends use multiple threads?  Which methods are possibly called concurrently?

pitrou:
it may be a bug or limitation in Tornado too. According to [this](https://github.com/leporo/tornado-redis/issues/59), `add_callback` is not reentrant and deadlocks when the lock is taken again...

pitrou:
Ok, I determined that it is indeed a deadlock in Tornado's `add_callback` due to reentrant invocation through a `__del__`. See the following traceback excerpt:

```
  File "/home/antoine/distributed/distributed/joblib.py", line 54, in apply_async
    future = self.client.submit(func, *args, **kwargs)
  File "/home/antoine/distributed/distributed/client.py", line 643, in submit
    loose_restrictions, priority={skey: 0})
  File "/home/antoine/distributed/distributed/client.py", line 1241, in _graph_to_futures
    'priority': priority})
  File "/home/antoine/distributed/distributed/client.py", line 345, in _send_to_scheduler
    self.loop.add_callback(self.scheduler_stream.send, msg)
  File "/home/antoine/tornado/tornado/ioloop.py", line 964, in add_callback
    stack_context.wrap(callback), *args, **kwargs))
  File "/home/antoine/tornado/tornado/stack_context.py", line 278, in wrap
[...]

PR #646, opened 2016-11-09
[EXP] Background cleanup thread

pitrou:
This is an attempt to implement the background cleanup thread suggested in issue #599. Only objects with a finalizer that doesn't need to reference `self` can be converted to this idiom.

This does fix the test_joblib hangs even without https://github.com/tornadoweb/tornado/pull/1876.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -33,6 +33,7 @@ from .batched import BatchedSend
 from .utils_comm import WrappedKey, unpack_remotedata, pack_data
 from .compatibility import Queue as pyQueue, Empty, isqueue
 from .core import (read, write, connect, coerce_to_rpc, clean_exception)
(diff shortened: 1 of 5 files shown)


mrocklin:
I haven't looked extensively at this yet, but my first impression is "is this worth it?"  How pervasive is the issue exposed in the joblib test?  How error prone is this solution?  I feel like we might be taking on more complexity.

pitrou:
I'm not sure it is. I'm posting it here for reference and also to test it on the CI builders, but I don't feel comfortable merging it into the source tree (hence the "[EXP]" -- for experimental -- in the title).

mrocklin:
Can we close this without merging?  It looks like this will be handled upstream.

pitrou:
Agreed?

PR #648, opened 2016-11-10
Fix #647: clarify usage of add_done_callback()

### LINKED ISSUE #647: future.result() freezes in add_done_callback()
Hi,

below is a simplified example of how calling `future.result()` in `add_done_callback` freezes.

```
c = Client()

def sleep():
    time.sleep(2)
    return 1

def cb(f):
    print(f)
    print('Result should be {}'.format(sleep()))
    print('Result was', f.result())

d = delayed(sleep)()
d.compute() # 1

c.compute(d).result() # 1

f2 = c.compute(d)
f2.add_done_callback(cb)
# <Future: status: finished, type: int, key: finalize-e24717a1ea49f34025d9f69a8d4b9e3e>
# Result should be 1
```

You never get the `f.result()` in `cb`. Execution freezes.

pitrou:
Also add tests for coroutine usage.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -157,7 +157,18 @@ class Future(WrappedKey):
 
         The callback ``fn`` should take the future as its only argument.  This
         will be called regardless of if the future completes successfully,
-        errs, or is cancelled
+        errs, or is cancelled.
+
+        The callback is called from the event loop, so it should only
+        use coroutine variants of blocking methods, e.g.::
+
+            @gen.coroutine
+            def cb(fut):
+                res = yield fut._result()
+                print("Got result:", res)
+
+            my_future.add_done_callback(cb)
+
         """
         self.client.loop.add_callback(done_callback, self, fn)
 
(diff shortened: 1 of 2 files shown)


mrocklin:
We might consider changing the API around a bit to call the function in a separate thread, perhaps using a single-threaded `concurrent.futures.ThreadPoolExecutor`

pitrou:
That's a possibility. What is the use case for `add_done_callback`?

mrocklin:
People were asking for it to be there.  Pragmatically I think it's also nice for people who want to run a local function when a future finishes, for example for logging.

pitrou:
I was wondering if people were expecting to pass a coroutine or a regular function but, since the main distributed API is blocking, they are probably expecting to pass a regular blocking function.

pitrou:
Alternative solution in PR #656.

pitrou:
Obsoleted by #656, closing.

PR #656, opened 2016-11-12
Execute a Future's done callback in a separate thread

### LINKED ISSUE #647: future.result() freezes in add_done_callback()

pitrou:
Fixes #647.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1,6 +1,7 @@
 from __future__ import print_function, division, absolute_import
 
 from collections import defaultdict, Iterator, Iterable
+from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures._base import DoneAndNotDoneFutures, CancelledError
 from contextlib import contextmanager
 import copy
@@ -72,6 +73,9 @@ class Future(WrappedKey):
     --------
     Client:  Creates futures
     """
+    _cb_executor = None
+    _cb_executor_pid = None
+
     def __init__(self, key, client):
         self.key = key
         tkey = tokey(key)
@@ -154,8 +158,15 @@ class Future(WrappedKey):
         The callback ``fn`` should take the future as its only argument.  This
         will be called regardless of if the future completes successfully,
         errs, or is cancelled
+
+        The callback is executed in a separate thread.
         """
-        self.client.loop.add_callback(done_callback, self, fn)
+        cls = Future
+        if cls._cb_executor is None or cls._cb_executor_pid != os.getpid():
+            cls._cb_executor = ThreadPoolExecutor(1)
+            cls._cb_executor_pid = os.getpid()
+        self.client.loop.add_callback(done_callback, self,
+                                      partial(cls._cb_executor.submit, fn))
 
     def cancel(self):
         """ Returns True if the future has been cancelled """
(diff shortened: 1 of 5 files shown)


pitrou:
I've first tried to store the Future objects directly on the Client, but ran into nasty GC problems which weren't worth trying to solve or workaround (IMO).

distributed/worker_client.py
```diff
@@ -9,10 +9,10 @@
 from dask.base import tokenize
 from tornado import gen
 
-from .client import Client, Future, pack_data, unpack_remotedata
+from .client import AllExit, Client, Future, pack_data, unpack_remotedata
 from .sizeof import sizeof
 from .threadpoolexecutor import secede
-from .utils import log_errors, sync, tokey, ignoring
+from .utils import All, log_errors, sync, tokey, ignoring
```
pitrou: Note this was a bug that was hidden by the overly broad `ignoring(Exception)` in `WorkerClient._gather`.

distributed/client.py
```diff
@@ -158,22 +158,36 @@ def add_done_callback(self, fn):
...
         will be called regardless of if the future completes successfully,
         errs, or is cancelled
+
+        The callback is executed in a separate thread.
         """
-        self.client.loop.add_callback(done_callback, self, fn)
+        cls = Future
+        if cls._cb_executor is None or cls._cb_executor_pid != os.getpid():
+            cls._cb_executor = ThreadPoolExecutor(1)
+            cls._cb_executor_pid = os.getpid()
```
mrocklin: The PID checking here is in case we fork a process that has Futures?

distributed/client.py
```diff
@@ -235,9 +246,42 @@ def __str__(self):
...
+
+    def lose(self):
+        self.status = 'lost'
+        self.event.clear()
+
+    def set_error(self, exception, traceback):
+        self.status = 'error'
+        self.exception = exception
+        self.traceback = traceback
+        self.event.set()
```
mrocklin: Is there a clean way to combine this into the `Future` class itself?  It would be nice to avoid a deep class hierarchy.  I suspect that this is somewhat difficult, but thought that I would ask just in case.

distributed/client.py
pitrou: Yes. I'm not sure that happens, but better safe than sorry.

distributed/client.py
pitrou: As I said, i tried to store Future objects directly on the Client, but ended up with nasty GC issues (`__del__` could be called even though the Future was still alive). Hence the two-class approach.
Note there is no inheritance here, so the class hierarchy remains shallow.

Without storing Futures on the Client, I find that it's neater to have a state class than have Client much with raw dictionaries represent the state.

pitrou:
Is there something else you want me to do here?

mrocklin:
Fine by me.  Merge away.

PR #661, opened 2016-11-14
Rework BatchedSend logic

pitrou:
Does away with the timeout and looking up a private attribute on IOStream.
Refs PR #653.


distributed/batched.py
```diff
@@ -38,87 +39,76 @@ class BatchedSend(object):
...
-            self.last_send = write(self.stream, payload)
-        except Exception as e:
-            logger.exception(e)
-            raise
+            try:
+                yield write(self.stream, payload)
+            except Exception:
+                logger.exception("Error in batched write")
+                break
+            self.next_deadline = self.loop.time() + self.interval
```
mrocklin: We might want to base the next deadline on when we started the last send rather than when we finsihed it.

distributed/batched.py
```diff
@@ -38,87 +39,76 @@ class BatchedSend(object):
...
+        self.waker.set()
+        yield self.stopped.wait()
         try:
-            if self.stream._write_buffer:
-                yield self.last_send
             if self.buffer:
+                if self.next_deadline is not None:
+                    delay = self.next_deadline - self.loop.time()
+                    if delay > 0:
+                        yield gen.sleep(delay)
```
mrocklin: Why this added delay?

distributed/tests/test_batched.py
```diff
@@ -37,7 +37,7 @@ def handle_stream(self, stream, address):
                 self.count += 1
                 yield write(stream, msg)
             except StreamClosedError as e:
-                pass
+                return
```
mrocklin: I would expect this to be a syntax error in Python 2

mrocklin:
This looks pretty nice to me

distributed/batched.py
pitrou: Well, I don't know. What are the intended semantics?

distributed/batched.py
pitrou: It mirrors the `yield self.last_send` that was here previously. Perhaps I'm misunderstanding the intent :-)

distributed/batched.py
pitrou: Then the `yield write` wasn't really useful in the previous version? There's no need to wait on the write future if we want to base the deadline on the start of the write operation.

Related question: what is with `gen.with_timeout(timedelta(seconds=0.01),...)` in `d.core.write`?

mrocklin: That is related to what was going on here and is also a possible source of error.  Some explanation here: https://github.com/dask/distributed/pull/653

pitrou: Oh, right. We can remove it.

pitrou:
By the way, `BatchedStream` doesn't seem used anymore, perhaps we should remove it?

mrocklin:
Removing it sounds fine to me.  Generally I am happy to yield to your judgment on anything related to this issue.  I suspect that you have a lot more experience here than I do.

mrocklin:
Testing failures here are unrelated.  Addressing them in https://github.com/dask/distributed/pull/662 .

distributed/core.py
```diff
@@ -196,58 +196,42 @@ def handle_stream(self, stream, address):
...
+    for frame in frames[:-1]:
+        stream.write(frame)
 
-        while stream._write_buffer:
-            try:
-                yield gen.with_timeout(timedelta(seconds=0.01), futures[-1])
-                break
-            except gen.TimeoutError:
-                pass
+    yield stream.write(frames[-1])
```
mrocklin: I *think* that this could fail if we write to the same stream in another coroutine.  Ideally we shouldn't do this.  Normally the `rpc` class creates new streams as necessary to handle concurrent communications to the same destination.  All cases that I can find when a coroutine writes directly to a stream it creates and owns that stream exclusively.

Still though, we were running into problems in the wild that suggested that this might be an issue.

pitrou: I'm not sure how that's different from the old code, though? It would also wait on `futures[-1]` and only catch timeout errors.

mrocklin: Somewhere some coroutine is stuck waiting on `yield write(...)`.  This first occurred on `yield self.last_write` within BatchedSend, and resulted in messages waiting in the worker's message buffer.

pitrou: You're right. That's because `IOStream.write` can forget previous futures. So how about we don't wait for the write at all? We could simply yield `gen.moment` so that write() remains a coroutine...

mrocklin: I think we would want to yield on the write if we were going to apply backpressure.  We're not doing this yet at other stages though so yes, I suspect that that would work fine.  We're moving the data pile-up from the BatchedSend buffer to the Tornado write buffer, which is probably appropriate anyway.

pitrou: Turns out we must wait for the write() to be issued before closing the stream. This is gonna be a bit hairy...

mrocklin: More or less hairy than polling on the _write_buffer?

pitrou: I think the solution is to flush the stream explicitly before closing. Let me try it out.

distributed/core.py
```diff
@@ -229,11 +229,36 @@ def write(stream, msg):
...
+        yield stream.write(b'')
+
+
+@gen.coroutine
+def close(stream):
+    """Close a stream safely.
+    """
+    if not stream.closed():
+        try:
+            flush(stream)
```
pitrou: Yes, you're right, my bad.

distributed/core.py
```diff
@@ -229,11 +229,36 @@ def write(stream, msg):
...
     yield gen.moment
 
 
+@gen.coroutine
+def flush(stream):
+    """Flush the stream's output buffer.
+    This is recommended before closing the stream.
+    """
+    if stream.writing():
+        yield stream.write(b'')
```
mrocklin: Are we confident that this expectation is fulfilled?

mrocklin: I'm mostly concerned about us as users.  Only very advanced dask/distributed users should use read/write/close directly.  

However, given that we've had problems reported it's possible that we aren't handling everything well internally.  

I have no problem with read/write/close as an API generally.

mrocklin:
It would be good to develop some tests to stress communication in a few ways.  However, I'm not entirely sure how this needs to be stressed.  One thing that I've found to be useful in the past is to change `300000` to `1000000` in `distributed/tests/test_batched.py::test_sending_traffic_jam`.

mrocklin:
Thoughts on ignoring the RuntimeError around the h5py test?

mrocklin:
This all seems fine to me.  +1

pitrou:
> Thoughts on ignoring the RuntimeError around the h5py test?

I would hope h5py merges the pull request that would fix the issue.

mrocklin:
I do not expect h5py to merge or release quickly.

PR #662, opened 2016-11-14
Use SerializableLock in hdf5 test

mrocklin:
I believe that this was the cause of an intermittent error in #661 .  This forces dependence on dask master.


mrocklin:
@shoyer is it sufficient to have a per-process lock for HDF5 reading or do we need a per-machine lock?

shoyer:
I'm pretty sure per process is enough.

pitrou:
I've submitted a fix upstream at https://github.com/h5py/h5py/pull/776

PR #670, opened 2016-11-14
Use ndarray instead of frombuffer for numpy serialization

teoliphant:
These are a few minor edits I made while reviewing dask serialization in preparation for a talk.  

the numpy.ndarray constructor is exactly the right approach to building a new array from the data stored in headers + bytestring.   Behavior should be equivalent, but it avoids using "stride_tricks" which is not a standard API.

### DIFF
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -40,8 +40,8 @@ def dumps(msg):
                if type(value) is Serialized}
 
         data = {key: serialize(value.data)
-                     for key, value in data.items()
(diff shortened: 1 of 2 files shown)


mrocklin:
Looks good to me.  Merging on passed tests.

mrocklin:
While you're looking at this issue, if you have other thoughts for corner cases they would be most welcome.  Current list is here: https://github.com/dask/distributed/blob/master/distributed/protocol/tests/test_numpy.py#L29-L57

PR #673, opened 2016-11-15
Fix #607: make Tornado coroutines serializable

### LINKED ISSUE #607: New tornado generators are not serializable
I was working on tornado master for a while (4.5) and found that coroutines were not serializable, even with cloudpickle.

pitrou:
As a bonus, serializing coroutines is now much faster and produces a smaller payload.

### DIFF
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -1,9 +1,13 @@
 from __future__ import print_function, division, absolute_import
 
+import inspect
+from io import BytesIO
 import logging
 import pickle
 
 import cloudpickle
+
+from tornado import gen
 
 from ..utils import ignoring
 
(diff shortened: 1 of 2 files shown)


mrocklin:
The reason to avoid putting this into cloudpickle directly is that tornado is a bit of a special case?

pitrou:
I don't know, it depends on @rgbkrk's preference. The code involves a fair amount of hacks :-)

mrocklin:
I think that cloudpickle is entirely hacks.  Also, there isn't really a single core maintainer of cloudpickle.  It's a bunch of people like us that have some small amount of need for the project.  You are about as likely to maintain cloudpickle in the future as anyone else.

pitrou:
Still, I don't have push access to cloudpickle, so the current maintainers would have to give their opinion whether they'd like to support this kind of code (possibly a bit more hackish, since we have to avoid importing tornado unconditionally in cloudpickle).

mrocklin:
cc @minrk @ogrisel .  Do we want this code to support Tornado coroutines in cloudpickle?

distributed/protocol/pickle.py
```diff
@@ -19,6 +23,59 @@
...
+@gen.coroutine
+def _tornado_coroutine_sample():
+    yield
+
+def is_tornado_coroutine(func):
+    """
+    Return whether *func* is a Tornado coroutine function.
+    Running coroutines are not supported.
+    """
+    return func.__code__ is _tornado_coroutine_sample.__code__
```
minrk: How reliable is this check? What sort of change in `@gen.coroutine` would break this `is` check?

minrk:
I'd be AOK with supporting tornado coroutines in cloudpickle if I were more confident in the detection of the coroutine (i.e. if https://github.com/tornadoweb/tornado/pull/1890 is merged). I'd love it even more if tornado itself handled the pickle/unpickle of coroutines.

mrocklin:
cc @ajdavis

distributed/protocol/pickle.py
pitrou: It would need `gen.coroutine` to return something else than a wrapper function. Apparently this has been the case since the beginning, see https://github.com/tornadoweb/tornado/commit/58b0dabd285458043bec38a1861372bd93f6ee27

pitrou:
> I'd be AOK with supporting tornado coroutines in cloudpickle if I were more confident in the detection of the coroutine (i.e. if tornadoweb/tornado#1890 is merged).

I can understand that. From distributed's point of view, if tornadoweb/tornado#1890 isn't merged before the Tornado 4.5 release, we'll need our own workaround.

pitrou:
On the good news front, https://github.com/tornadoweb/tornado/pull/1890 has now been merged.

rgbkrk:
Great news!

pitrou:
I've submitted https://github.com/cloudpipe/cloudpickle/pull/69

pitrou:
This is superseded by the aforementioned cloudpickle PR. Closing.

Issue #676, opened 2016-11-15
str / bytes mismatch in keys


pitrou:
I noticed the following ignored exception during e.g. `distributed/tests/test_scheduler.py::test_add_worker`. It seems at some point some of the code paths were converted to use bytestrings as keys (why?), but not all of them, and Python 3 obviously doesn't like it.

(also, some keys can also be tuples... is there a hard rule?)

To me, str keys would feel more natural. Is there a reason I'm missing?

```
ERROR:tornado.application:Future <tornado.concurrent.Future object at 0x7f8fa40a2630> exception was never retrieved: Traceback (most recent call last):
  File "/home/antoine/tornado/tornado/gen.py", line 1043, in run
    yielded = self.gen.send(value)
  File "/home/antoine/distributed/distributed/worker.py", line 563, in compute_one
    self.data[key] = result.pop('result')
  File "/home/antoine/miniconda3/envs/dask35/lib/python3.5/site-packages/zict/buffer.py", line 61, in __setitem__
    if key in self.slow:
  File "/home/antoine/miniconda3/envs/dask35/lib/python3.5/_collections_abc.py", line 601, in __contains__
    self[key]
  File "/home/antoine/miniconda3/envs/dask35/lib/python3.5/site-packages/zict/func.py", line 39, in __getitem__
    return self.load(self.d[key])
  File "/home/antoine/miniconda3/envs/dask35/lib/python3.5/site-packages/zict/file.py", line 34, in __getitem__
    with open(os.path.join(self.directory, key), 'rb') as f:
  File "/home/antoine/miniconda3/envs/dask35/lib/python3.5/posixpath.py", line 89, in join
    genericpath._check_arg_types('join', a, *p)
  File "/home/antoine/miniconda3/envs/dask35/lib/python3.5/genericpath.py", line 145, in _check_arg_types
    raise TypeError("Can't mix strings and bytes in path components") from None
TypeError: Can't mix strings and bytes in path components
```

pitrou:
@mrocklin

mrocklin:
This was me being dumb.  I didn't realize that msgpack would return strings
to us if we asked it to.  All changes were made and then most were reverted.

PR #677, opened 2016-11-15
Fix some silenced exceptions during the test suite

### LINKED ISSUE #675: Serialization turns tuples into lists
```
>>> tup = 1,2
>>> from distributed.protocol import core
>>> core.loads(core.dumps(tup))
[1, 2]
```

It produces ignored exceptions in tests such as `distributed/tests/test_scheduler.py::test_remove_worker_from_scheduler`:

```
ERROR:tornado.application:Future <tornado.concurrent.Future object at 0x7f707583ef98> exception was never retrieved: Traceback (most recent call last):
  File "/home/antoine/tornado/tornado/gen.py", line 304, in wrapper
    yielded = next(result)
  File "/home/antoine/distributed/distributed/worker.py", line 540, in compute_one
    self.active.add(key)
TypeError: unhashable type: 'list'
```

pitrou:
Fixes #675 and #676.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -33,7 +33,7 @@ from .core import (rpc, connect, read, write, MAX_BUFFER_SIZE,
         Server, send_recv, coerce_to_address, error_message)
 from .utils import (All, ignoring, clear_queue, get_ip, ignore_exceptions,
         ensure_ip, get_fileno_limit, log_errors, key_split, mean,
-        divide_n_among_bins)
+        divide_n_among_bins, validate_key)
 from .utils_comm import (scatter_to_workers, gather_from_workers)
 from .versions import get_versions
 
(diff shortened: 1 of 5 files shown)


distributed/scheduler.py
```diff
@@ -758,12 +758,13 @@ def remove_worker(self, stream=None, address=None, safe=False):
             del self.worker_bytes[address]
 
             for key in self.has_what.pop(address):
-                self.who_has[key].remove(address)
-                if not self.who_has[key]:
-                    if key in self.tasks:
-                        recommendations[key] = 'released'
-                    else:
-                        recommendations[key] = 'forgotten'
+                if key in self.who_has:
```
mrocklin: Why was this change necessary?  These two dictionaries should always be synchronized.

pitrou: In `transition_memory_released`, there is the following:
```
            workers = self.who_has.pop(key)
            for w in workers:
                if w in self.worker_info:  # in case worker has died
                    self.has_what[w].remove(key)
```

I was seeing an ignored KeyError before the change.

mrocklin: I would prefer to hunt down why these two were not matched.  I can try to find this sometime tomorrow.  What test was triggering this failure?  Generally I prefer to remain very strict in the scheduler.

pitrou: `distributed/tests/test_client.py::test_scatter_hash` and `distributed/tests/test_client.py::test__scatter_types`.

pitrou: I've removed the workaround.

PR #678, opened 2016-11-15
Fix test_stress_communication

pitrou:


mrocklin:
We should consider xfail'ing  `distributed/tests/test_worker_failure.py::test_broken_worker_during_computation` for a while.

pitrou:
Speaking of which, it seems that whenever a rpc remains active after a test, it's always connected to a non-localhost address (e.g. "192.168.x.x"). This seems to hint that there's a stream coroutine waiting for the final close notification to come.

pitrou:
To be more preicse on the `test_broken_worker_during_computation` failures:
* `gather_from_workers` creates an rpc to the Nanny's worker
* the Nanny's worker is configured as using the external address ("192.168.x.x" here), not localhost
* when the worker is killed, `gather_from_workers` may take a little time to notice
* in the mean time, the test function has returned and the `rpc.active` test fails...

So this seems to mean the failure is harmless.

PR #684, opened 2016-11-16
use cPickle in protocol in Python 2

mrocklin:


pitrou:
Looks like an obvious improvement to me.

PR #685, opened 2016-11-17
New HDFS container

pitrou:
Main differences compared to the previous approach:
* re-use an existing hadoop image
* tests are run from the host against the container

As a result, the build is quite faster. The random hangs haven't disappeared, though :-)


mrocklin:
cc @danielfrg

pitrou:
I believe this is ready for review.

mrocklin:
Also cc @koverholt @quasiben in case they happen to have more time than @danielfrg

pitrou:
Bythe way, the only gotcha here is that I had to add the container's hostname to `/etc/hosts`. It seems the client uses that hostname to make some connections, even though the initial connection is to `localhost:8020`. That doesn't sound like a problem here, though I'd be delighted if someone can explain me a way to avoid it (especially for local testing).

danielfrg:
LGTM this also removes some code/setup scripts that are annoying to maintain :)

I don't think there is a way to avoid the `/etc/hosts` thing but i would love to be proven wrong. The only tip I have to make it a little bit better on travis is to use this: https://docs.travis-ci.com/user/hosts

pitrou:
@danielfrg, thanks for the tip! I'm merging now.

PR #692, opened 2016-11-21
WIP - Communication stress test and small fixes

mrocklin:
This adds a communication heavy stress test with the following features:

1.  Many tasks (50k)
2.  Lots of data within the graph (many numpy arrays)
3.  Repeated submission of these graphs with low delay

I tested this and the rest of the test suite under moderate network delay (20ms roundtrip) and cleaned up a few small issues but wasn't able to find anything significant.  This was an effort to identify the cause of a report of dropped messages under similar workloads.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -278,6 +278,11 @@ class FutureState(object):
         self.traceback = traceback
         self.event.set()
 
+    def __str__(self):
+        return '<%s: %s>' % (self.__class__.__name__, self.status)
+
+    __repr__ = __str__
+
 
 @gen.coroutine
 def done_callback(future, callback):
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -137,8 +137,8 @@ class Server(TCPServer):
         """
         stream.set_nodelay(True)
         ip, port = address
-        logger.info("Connection from %s:%d to %s", ip, port,
-                    type(self).__name__)
+        logger.debug("Connection from %s:%d to %s", ip, port,
+                     type(self).__name__)
         try:
             while True:
                 try:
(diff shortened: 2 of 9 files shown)


mrocklin:
I'll try this with larger parameters under a repeated loop next.

distributed/tests/test_stress.py
```diff
@@ -166,3 +166,25 @@ def test_stress_communication(c, s, *workers):
 
     result = yield future._result()
     assert isinstance(result, float)
+
+
+@slow
+@gen_cluster(client=True, ncores=[('127.0.0.1', 1)] * 8, timeout=10000)
```
pitrou: I hope 10000 s. is not the order of magnitude of this test's duration? :-)

mrocklin: No, it's on the order of minutes on my machine.  It also doesn't yet actually find any issues though.  This is still a work in progress.

distributed/tests/test_stress.py
```diff
@@ -166,3 +166,25 @@ def test_stress_communication(c, s, *workers):
...
+def test_communication_heavy_graph(c, s, *workers):
+    da = pytest.importorskip('dask.array')
```
pitrou: By the way, it seems distributed already imports Numpy and other stuff unconditionally:
```
>>> import distributed
>>> import sys
>>> "numpy" in sys.modules
True
>>> "pandas" in sys.modules
True
```

mrocklin: It tries to import them if available but is intended to work without them.

mrocklin:
Feedback welcome on this.  I'd like to merge in the small fixes without the stress test.

pitrou:
No objection from me.

PR #693, opened 2016-11-21
Scale auto memory limit based on fraction of cores

mrocklin:
Previously every Worker would get 60% of the total memory by default.
This would cause problems when we launched many single-threaded workers
on larger compute nodes.

Now we scale the available memory to match the fraction of cores that we
claim.


pitrou:
LGTM.

PR #695, opened 2016-11-21
Run slow tests on Travis

pitrou:
Let's see if anything is broken.


pitrou:
The answer is yes :-)

pitrou:
There are two failures remaining: in `test_open_close_many_workers[Nanny-10-20]` and `test_missing_worker`.

mrocklin:
As written `test_mising_worker` should genuinely fail.  We used to be more lenient about inconsistent internal state but now we are pretty strict.  This test creates inconsistent internal state.

mrocklin:
I don't know what's up with the file descriptor test, other than that we have a few problems like that.  We're slightly unclean in our handling of file descriptors.

distributed/tests/test_worker.py
```diff
@@ -573,8 +573,8 @@ def test_Executor(c, s):
 
 
 @slow
-@gen_cluster(client=True, ncores=[('127.0.0.1', 1)])
-def test_str(c, s, w):
```
pitrou: This was shadowing the other test named `test_str`.

distributed/tests/test_worker_failure.py
```diff
@@ -49,10 +50,14 @@ def test_gather_then_submit_after_failed_workers(loop):
             total = c.submit(sum, L)
             wait([total])
 
-            (_, port) = first(c.scheduler.who_has[total.key])
```
pitrou: This was buggy.

pitrou:
Well, I *seem* to have made the slow tests somehow more stable now.

distributed/tests/test_client.py
```diff
@@ -3141,16 +3143,16 @@ def test_reconnect(loop):
...
         start = time()
         while len(c.ncores()) != 1:
-            sleep(0.01)
-            assert time() < start + 5
+            sleep(0.05)
+            assert time() < start + 15
 
         x = c.submit(inc, 1)
         assert x.result() == 2
+        print(c.status, c.ncores())
```
mrocklin: print statement

PR #696, opened 2016-11-22
Add communication stress test

mrocklin:
This sends a large heavy graph up to the scheduler


pitrou:
Is there a fundamental difference with test_stress_communication()? If no, how about keeping only one of both?

PR #703, opened 2016-11-25
WIP: extend protocol docs

mrocklin:


mrocklin:
Here are some other notes.  I'm not sure if I want them in long-term docs or not though:

How to make a dask worker and client
====================================

Client side
-----------

A client sends the following message to the scheduler:

```python
{'op': 'register-client',
 'client': 'unique-client-identifier'}
```

It receives an acknowledgement message:

```python
{'op': 'stream-start'}
```

This same stream will be used to send and receive messages for tasks.

### Send computations

When the client wants to inform the scheduler of new computations it sends an
update-graph message:

```python
{'op': 'update-graph',
 'tasks': {key-names: stuff-to-send-to-worker},
 'keys': [list of keys we care about]
 'dependencies: {key: [keys]},
 'client': 'unique-client-identifier'}
```

Where ``stuff-to-send-to-worker`` is language-specific.  In Python this is
usually dictionaries of bytes like the following:

```python
{
  'function': pickle.dumps(add),
  'args': pickle.dumps([1, 2])
  'kwargs': {}
}
```

or
```python
{
  'task': pickle.dumps((add, 1, 2))
}
```

However you choose to encode this will be send directly to the worker at the
right time.

There are more options here, but we'll keep things minimal for now.

### Receive updates

This same stream gets updates from the scheduler the fate of its keys.  There
are a few messages based on what happens to the computation:

```python
{
  'op': 'key-in-memory',
  'key': key,
}

{
  'op': 'task-erred',
  'key': key,
  'exception': binary-blob-from-worker,
  'traceback': binary-blob-from-worker,
}

{
  'op': 'cancelled-key',
  'key': key,
}
```

These messages will come in batches as small lists of messages.

### Gather Data

On separate connections (not the high-volume socket) the client can request
computed data

```python
{
  'op': 'gather',
  'keys': [list of keys]
}
```

It's worth noting that all keys are assumed to be strings.


## Worker


The worker sends a message to the scheduler registering itself:

```python
{
  'op': 'register',
  'address': (ip, port),
  'ncores': int,
  'keys': [],
  'name': an optional alias,
  'nbytes': {},
  'now': current time, as integer seconds,
  'memory_limit': integer number of bytes,
  'host_info': dictionary about host health, {} for now,
}
```

This stream should hear back a general acknowledgement from the scheduler.  It
is nice to repeat this message periodically as a heartbeat.

### Listen for work


The worker should be a TCP Server listening for new connections.  When it gets
a connection with the following message then it should listen on that socket
for a feed of tasks:

```python
{
  'op': 'compute-stream'
}
```

These tasks come in lists of small messages like the following:

```python
{
  'op': 'compute-task',
  'key': key,
  'the-stuff-you-put-in-update-graph': ...
  'who_has': {key: [workers that have key]}
}
```

It is then the job of the worker to take this information, gather data from
peer workers (given location information in ``who_has``) and execute the task.

### Report Back

When it finishes execution the worker should hold onto the data, associated to
the given key, and report back to the scheduler along the same channel.

```python
{
  'status': 'OK',
  'key': key,
  'nbytes': int,
  'type': bytes-to-send-to-client,
  'compute_start': integer time,
  'compute_stop': integer time,
  'thread': thread ID of the core that ran the computation (for diagnostics)
}

The worker should not send back the result, but instead keeps it available for
workers or the scheduler to collect as necessary.

### Serve Data

Separately the worker should serve data to anyone that asks for it.  These
requests will come in on separate sockets.

```python
{
  'op': 'get_data',
  'keys': [list of keys]
}
```

Responses should be a mapping of keys to serialized values:

```python
{
  'x': bytes,
  'y': bytes,
  'z': bytes,
}
```

Similarly the worker will respond to messages to delete data

```python
{
  'op': 'delete_data',
  'keys': [list of keys]
}
```

The worker is expected to hold onto data until it is asked to delete it.

Issue #698, opened 2016-11-22
Redesign Workers


mrocklin:
While trying to add support for [resource constraints](https://github.com/dask/distributed/pull/694) I got blocked on trying to add more logic to the Worker.  The worker is probably the oldest bit of code still living in dask/distributed and could use a refactor to a more modern design.  This would increase complexity but would also allow us to put more logic into the workers and increase performance and flexibility in a few critical areas.

### Opportunities for improvement

*   Local task scheduling.  For example we might wait until a local resource like a GPU is ready for use.
*   Better overlap communication and computation.  At the moment we get a batch of tasks, perform all inter-worker communication for those tasks, and then start processing.  Tasks with smaller dependencies may be unnecessarily delayed by larger ones.
*   Prioritize tasks to saturate the CPU.  For example we might reorder communication to start long-running small-byte tasks first while we communicate larger results in the background.
*   Limit number of outgoing connections.  This might benefit shuffle performance as well as help to avoid out-of-file-descriptor issues.
*  More comprehensive worker diagnostics for future visualization
*  Eventual support for multiprocessing within a single worker
*  The scheduler doesn't know which data is in RAM and which data is on disk, sometimes resulting in sub-optimal task ordering.
*  Smoother consumption.  We can probably decrease peak scheduling load and improve communication bandwidth if we put more tasks on the worker at once.  This requires us to be able to send tasks back to the scheduler if we become overwhelmed.

### Current design

Currently the worker is straightforward.  There is a `ThreadPoolExecutor`, a `MutableMapping`, and  a `TCPServer` that waits on instructions.  When someone asks for a computation we gather data from workers, put that data into the MutableMapping, submit a task to the ThreadPoolExecutor, wait for it to finish, put the result in the `MutableMapping` and send back a job-finished message.  Other than the MutableMapping and the ThreadPoolExecutor there is no state.

This design is straightforward and fairly robust.  There have been few bugs in the worker.

### Proposed design

We keep the MutableMapping and ThreadPoolExecutor, but also add a pipeline through which tasks proceed.  This will look like a much simpler version of the scheduler.

As tasks arrive we will need to track their data dependencies, including which are in `.data` and which are yet to be received.  

New data dependencies will also need to be tracked, including where we believe them to be.  Given the outstanding data dependencies and a fixed number of connections (say, 10-50) we prioritize which workers to contact, asking for which data.  

[...]

PR #704, opened 2016-11-25
Worker Redesign

mrocklin:
This replaces the compute logic in the worker with something a bit more complex.  See #698 for more details.


mrocklin:
Running into the following error in `connect`

```python
distributed.worker - ERROR - 'NoneType' object has no attribute 'setsockopt'
  File "/home/mrocklin/workspace/distributed/distributed/worker.py", line 1157, in gather_dep
    stream = yield future
  File "/home/mrocklin/Software/anaconda/lib/python3.5/site-packages/tornado/gen.py", line 1015, in run
    value = future.result()
  File "/home/mrocklin/Software/anaconda/lib/python3.5/site-packages/tornado/concurrent.py", line 237, in result
  File "<string>", line 3, in raise_exc_info
  File "/home/mrocklin/Software/anaconda/lib/python3.5/site-packages/tornado/gen.py", line 1024, in run
  File "/home/mrocklin/workspace/distributed/distributed/core.py", line 327, in connect
    set_tcp_timeout(stream)
  File "/home/mrocklin/workspace/distributed/distributed/core.py", line 224, in set_tcp_timeout
    sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
AttributeError: 'NoneType' object has no attribute 'setsockopt'
```

mrocklin:
This is now ready for review

pitrou:
Given the scheduler is inheritently single-process and could become a bottleneck in large deployments, it sounds like a good idea to delegate some scheduling responsibilities to the workers. The new organization of Worker methods seems more easily readable as well. I'll come up with more detailed comments later.

pitrou:
(the `test_stress_scatter_death` hang on Travis-CI looks new?)

mrocklin:
Yeah, I'm figuring it out now.  The combination of stress tests and run-time tests (for example `Worker.validate_state`) are pretty useful in finding odd corner cases.  Most of the time spent on this refactor has been identifying these cases with the help of tests like `test_stress_scatter_death`.

distributed/worker.py
```diff
@@ -920,3 +717,666 @@ def run(worker, stream, function=None, args=(), kwargs={}):
...
+                             self.total_connections)
+
+                key = self.data_needed[0]
+
+                if key not in self.tasks:
+                    self.data_needed.popleft()
+                    continue
+
+                if self.task_state.get(key) != 'waiting':
+                    self.log.append((key, 'communication pass'))
```
pitrou: Out of curiosity, when would this happen?

distributed/worker.py
```diff
@@ -920,3 +717,666 @@ def run(worker, stream, function=None, args=(), kwargs={}):
...
+                for dep in deps:
+                    if not self.who_has.get(dep):
+                        logger.info("Can't find dependencies for key %s", key)
+                        self.cancel_key(key)
+                        continue
+
+                self.log.append(('gather-dependencies', key, deps))
+
+                while deps and len(self.connections) < self.total_connections:
+                    self.gather_dep(deps.pop())
```
pitrou: `gather_dep()` is a coroutine, so it doesn't have a chance to affect the connections array right now.

distributed/worker.py
mrocklin: 1.  If the key was removed by some other event `key not in self.task_state` but wasn't removed from `data_needed` (removal from the deque is expensive)
2.  If they key was jumped to some other state, such as `'memory'` if we gather the key as part of a dependency for  some other task

Neither of these occur during normal operation

distributed/worker.py
mrocklin: My understanding is that it runs as a normal function until it hits a yield point

distributed/worker.py
```diff
@@ -920,3 +717,666 @@ def run(worker, stream, function=None, args=(), kwargs={}):
...
+
+        if key not in self.nbytes:
+            self.nbytes[key] = sizeof(value)
+
+        for dep in self.dependents.get(key, ()):
+            if dep in self.waiting_for_data:
+                if key in self.waiting_for_data[dep]:
+                    self.waiting_for_data[dep].remove(key)
+                if not self.waiting_for_data[dep]:
+                    self.waiting_for_data[dep]
```
pitrou: A `del` is missing.

distributed/worker.py
mrocklin: I would be happy avoiding this though and calling this with `loop.add_callback`.  Tornado code is difficult to profile otherwise.

distributed/worker.py
```diff
@@ -920,3 +717,666 @@ def run(worker, stream, function=None, args=(), kwargs={}):
...
+                            self.cancel_key(key)
+                    return
+                worker = random.choice(list(self.who_has[dep]))
+                ip, port = worker.split(':')
+                try:
+                    future = connect(ip, int(port))
+                    self.connections[future] = True
+                    stream = yield gen.with_timeout(timedelta(seconds=3),
+                                                    future)
+                except AttributeError:  # set_sock_opt sometimes fails
```
pitrou: I'd rather know how to reproduce this. setsockopt() should never fail, or at least it should only fail deterministically.

distributed/worker.py
```diff
@@ -920,3 +717,666 @@ def run(worker, stream, function=None, args=(), kwargs={}):
...
+                    continue
+                if total_bytes + self.nbytes[d] > self.target_message_size:
+                    break
+                deps.add(d)
+                total_bytes += self.nbytes[d]
+
+            self.connections[stream] = deps
+            for d in deps:
+                if d in self.in_flight:
+                    self.in_flight[d].add(stream)
```
pitrou: I'm a bit surprised here. Can several streams be requesting the same datum simultaneously?

mrocklin: Given the current checks I don't think so.  I just tend to do this by default now for everything.

distributed/worker.py
```diff
@@ -920,3 +717,666 @@ def run(worker, stream, function=None, args=(), kwargs={}):
...
+                ip, port = worker.split(':')
+                try:
+                    future = connect(ip, int(port))
+                    self.connections[future] = True
+                    stream = yield gen.with_timeout(timedelta(seconds=3),
+                                                    future)
+                except AttributeError:  # set_sock_opt sometimes fails
+                    pass
+                except (gen.TimeoutError, StreamClosedError):
+                    with ignoring(KeyError):  # other coroutine may have removed
```
pitrou: I'm a bit surprised we don't log anything here. In normal operation, a connection error sounds important to report.

distributed/worker.py
pitrou: > My understanding is that it runs as a normal function until it hits a yield point

It looks like you're right. It is labelled an optimization in Tornado's source code:
```
                # Inline the first iteration of Runner.run.  This lets us
                # avoid the cost of creating a Runner when the coroutine
                # never actually yields, which in turn allows us to
                # use "optional" coroutines in critical path code without
                # performance penalty for the synchronous case.
```

distributed/worker.py
pitrou: Perhaps assert it cannot happen instead? This would let such bugs appear clearly.

distributed/worker.py
mrocklin: I plan to switch this out though for something else, just to improve our ability to profile the worker.  Ideally we can isolate most of the logic into non-tornado code.

pitrou:
The `ThreadPoolExecutor` is limited to `ncores`, so it doesn't look like transitioning a task to `long-running` would actually free its thread for another task.

mrocklin:
Have you seen this: https://github.com/dask/distributed/blob/master/distributed/threadpoolexecutor.py ?

pitrou:
No, I hadn't. That explains it :-)

distributed/worker.py
mrocklin: Resolved missing del by removing line (it gets removed elsewhere)

distributed/worker.py
mrocklin: This has been removed

distributed/worker.py
mrocklin: We now assert that no data element is simultaneously in two communications

distributed/worker.py
```diff
@@ -298,7 +298,11 @@ def _close(self, report=True, timeout=10):
             v.stop()
         self.rpc.close()
         self.status = 'closed'
-        self.stop()
+        with ignoring(gen.TimeoutError):
+            yield gen.with_timeout(timedelta(seconds=1),
+                    [close(stream) for stream in self._listen_streams])
+        for stream in self._listen_streams:
+            stream.close()
```
mrocklin: @pitrou any thoughts on why this would cause more file descriptors to be left dangling?

The test `distributed/tests/test_client.py::test_open_close_many_workers[Nanny-10-20]` is now failing when we create and close many `Nanny`s.

pitrou: What if you remove the timeout?

mrocklin: The problem persists.  It also persists if you remove the calls to `close(stream)` entirely and just call `stream.close()`.

pitrou: It may be that closing the stream is confusing another coroutine which is using the stream at the same time. My question is: why did you want to add this?

mrocklin: In `distributed/tests/test_stress.py::test_stress_scatter_death` we close workers during computation.  We were finding that connections to killed workers were becoming non-responsive but still remained open.  The living worker trying to gather data from the dead worker would wait for a very long time on a living but ineffective stream.

distributed/tests/test_worker2.py
```diff
@@ -0,0 +1,83 @@
+
```
mrocklin: It was convenient to start, but I'll merge them now.

distributed/core.py
```diff
@@ -49,6 +49,9 @@ def get_total_physical_memory():
 MAX_BUFFER_SIZE = get_total_physical_memory()
 
 
+CommunicationErrors = (StreamClosedError, socket.error, IOError, RPCClosed)
```
mrocklin: @pitrou do you see any problems with doing this?

mrocklin:
@pitrou if you get a chance I could use some help tracking down a failure in `distributed/tests/test_stress.py::test_stress_scatter_death`

If you try this test a few times (I tend to run it in a loop) you'll eventually land in a pdb statement.  We seem to be waiting on a message that never finishes.  Useful things to look at include the following:

*  `s.processing` will show that there is still an outstanding task
*  `ws` is all of the currently living workers.  One will still have a key in flight
*  `ws[address].connections` will have a stream that is still open where `address` is the address of the waiting worker 
*  `first(ws[address].connections).socket.getpeername()` is the location from which we are gathering this data
*  `workers` holds all of the workers.  If you get the right worker and then check `w._listen_streams` you'll have all of the outstanding streams from that worker.

Sometimes the remote stream appears to be in a bad state.  Calling `.getpeername()` raises an OSError and yet the stream itself does not register as closed.  If I add a periodic callback to watch all streams and close them if this occurs then the frequency of failures decreases but doesn't go away.  The same situation can occur even when the sockets of both streams appear to be ok.

Anyway, if you have some time to give this a shot I'd appreciate any thoughts you might come up with.  I've been trying things here for a long while without much success.

distributed/core.py
mrocklin: Now using `EnvironmentError`.

mrocklin: RPCClosed ends up being convenient.  I don't now recall the original motivation.  I suspect that it was difficult to always ensure that the system hadn't closed down in between every yield point and using exceptions was simpler.

pitrou:
I've tried it out. I even got into a case where the socket was still believed to be connected:
```
(Pdb) pp ws
{'127.0.0.1:33244': <Worker: 127.0.0.1:33244, threads: 1, running: 0, ready: 0, in-flight: 1, waiting: 1>,
 '127.0.0.1:34060': <Worker: 127.0.0.1:34060, threads: 1, running: 0, ready: 0, in-flight: 0, waiting: 0>,
 '127.0.0.1:34650': <Worker: 127.0.0.1:34650, threads: 1, running: 0, ready: 0, in-flight: 0, waiting: 0>}
(Pdb) w1 = ws['127.0.0.1:33244']
(Pdb) pp w1.connections
{<tornado.iostream.IOStream object at 0x7fad78099438>: {'c49ce813f188e6493e3d9a4885782a0f'}}
(Pdb) s1 = first(w1.connections)
(Pdb) pp s1
<tornado.iostream.IOStream object at 0x7fad78099438>
(Pdb) pp s1.socket
<socket.socket fd=28, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 45564), raddr=('127.0.0.1', 34650)>
(Pdb) pp s1.socket.getpeername()
('127.0.0.1', 34650)
(Pdb) w2 = ws['127.0.0.1:34060']
(Pdb) pp w2.connections
{}
(Pdb) w3 = ws['127.0.0.1:34650']
(Pdb) pp w3.connections
{}
(Pdb) !import socket
(Pdb) pp s1.socket.getsockopt(socket.SOL_SOCKET, socket.SO_ERROR)
0
```
Yet the connection didn't appear anymore in `netstat -tnp`.

pitrou:
I stand corrected: the connection does appear in `netstat -tnp`:
```
(Pdb) pp s1.socket.getpeername()
('127.0.0.1', 34650)
(Pdb) pp s1.socket.getsockname()
('127.0.0.1', 45564)
```
And the netstat output shows there's some data in the recv queue on the other end (i.e. on the worker being gathered from):
```
$ netstat -tnp
[...]
tcp      111      0 127.0.0.1:34650         127.0.0.1:45564         ESTABLISHED 10223/python    
[...]
```

pitrou:
https://github.com/tornadoweb/tornado/issues/1901

pitrou:
I would recommend applying the following patch: https://gist.github.com/pitrou/f9a19e3b162213ca70169c28582e65e2

mrocklin:
You should have permissions to push directly to this branch

pitrou:
Ok, I'll try to.

mrocklin:
@pitrou any other comments on this PR?  I'm tempted to merge this soon.

pitrou:
I just pushed minor style cleanups. I'm ok on the PR, as I said it makes the worker's operation more understandable (though more complicated as well :-)).

mrocklin:
Yeah, I expect we'll run into a few frustrating issues in the near future.

pitrou:
At least we don't seem to see any sporadic failures.

mrocklin:
OK, merging.

PR #705, opened 2016-11-28
ENH: add Client.get_versions(check=False)

### LINKED ISSUE #664: ENH: add client.get_versions(check=True)
xref https://github.com/dask/distributed/pull/596

if check == True (default), then raise an error message showing differences in versions (e.g. distributed between workers and scheduler). makes debugging versions differences much easier.

open to different reporting options back to user (e.g. maybe just print out diffs)

jreback:
option to provide error checking .get_versions()

closes #664

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2007,7 +2007,6 @@ class Client(object):
         result = {'scheduler': scheduler, 'workers': workers, 'client': client}
 
         if check:
-
             # we care about the required & optional packages matching
             extract = lambda x: merge(result[x]['packages'].values())
             client_versions = extract('client')
@@ -2022,17 +2021,17 @@ class Client(object):
                                          client=cv,
                                          scheduler=sv))
 
-            for wkr, worker in workers.items():
-                worker_versions = merge(worker['packages'].values())
+            for w, d in workers.items():
+                worker_versions = merge(d['packages'].values())
                 for pkg, cv in client_versions.items():
                     wv = worker_versions[pkg]
                     if wv != cv:
                         raise ValueError("package [{package}] is version [{client}] "
-                                         "on client and [{worker}] on worker [{wkr}]!".format(
+                                         "on client and [{worker}] on worker [{w}]!".format(
                                              package=pkg,
                                              client=cv,
                                              worker=wv,
-                                             wkr=wkr))
+                                             w=w))
 
         return result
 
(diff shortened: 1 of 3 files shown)


distributed/client.py
```diff
@@ -1998,7 +2004,37 @@ def f(worker=None):
...
+            for pkg, cv in client_versions.items():
+                sv = scheduler_versions[pkg]
+                if sv != cv:
+                    raise ValueError("package [{package}] is version [{client}] "
+                                     "on client and [{scheduler}] on scheduler!".format(
+                                         package=pkg,
+                                         client=cv,
+                                         scheduler=sv))
+
+            for wkr, worker in workers.items():
```
mrocklin: We often use the pronouns `w` or `worker` to mean a worker address, like `127.0.0.1:8786`.  Recommend changing `wkr` to w or worker and changing worker to some other pronoun like `response` or `d`

distributed/client.py
```diff
@@ -1998,7 +2004,37 @@ def f(worker=None):
                 return None
 
         workers = sync(self.loop, self._run, f)
-        return {'scheduler': scheduler, 'workers': workers, 'client': client}
+        result = {'scheduler': scheduler, 'workers': workers, 'client': client}
+
+        if check:
+
```
mrocklin: Remove excess whitespace

distributed/tests/test_client.py
```diff
@@ -3281,6 +3281,10 @@ def test_get_versions(loop):
             for k, v in v['workers'].items():
                 assert v is not None
 
+            # smoke test for versions
+            # that this does not raise
+            c.get_versions(check=True)
```
mrocklin: Would prefer these comments after the code rather than on top.

jreback:
updated.

note that this is *not* extensively tested......

mrocklin:
It's hard to test things that explicitly need different software environments

jreback:
can add to release-notes.md if you'd like (ideally you'd add a new section for the next release when releasing FYI)

mrocklin:
Yeah, we're not very disciplined about asking contributors to add to release-notes.  I tend to just do things manually at each release.  However adding something yourself would be very appreciated.

jreback:
http://dask.pydata.org/en/latest/develop.html is slightly outdated I think. should be ``realease-notes.md``? (or is this different for dask & distributed)?

mrocklin:
Different between the two unfortunately.

jreback:
updated

jreback:
would it be useful to instead of a ``check=False`` have

``check_required``, ``check_optional`` and ``check`` (which would override the others if True)?

PR #706, opened 2016-11-28
Do not attempt configuring timeouts on a closed stream

pitrou:


pitrou:
@mrocklin, this should fix the AttributeErrors you are seeing.

PR #707, opened 2016-11-28
Skip joblib and threaded gather tests

mrocklin:
These block intermittently, slow down development,
and are low priority.  We choose to skip them for now.

Thoughts on this @pitrou ?


pitrou:
That sounds reasonable (AFAIR, the joblib test hang is resolved by the fix to Tornado's add_callback).

jakirkham:
xref: https://github.com/dask/distributed/issues/599 ?

PR #709, opened 2016-11-29
Worker bokeh

mrocklin:
Adds a Bokeh server to every worker.  Nothing fancy here yet.  But it's quite easy to add diagnostics.

image

@pitrou this may interest you (if you ever wanted to play with Bokeh)


mrocklin:
@pitrou any thoughts on what might be causing the Python 2.7 errors on this PR?  

https://travis-ci.org/dask/distributed/jobs/180385905
https://travis-ci.org/dask/distributed/jobs/180380932
https://travis-ci.org/dask/distributed/jobs/180274371

I'm only able to reproduce this locally when I run the full test suite rather than just these tests in isolation.  I'm only able to reproduce in Python 2.  

We *are* adding Bokeh, which runs on the Tornado IOLoop.  I don't believe we ever let it manage the IOLoop though.

pitrou:
https://travis-ci.org/dask/distributed/jobs/180385905#L1462 looks like the IO loop was closed from another thread than the thread it was running on.

There's an unrelated issue on AppVeyor, probably because Windows timers don't have enough granularity:
```
 File "C:\projects\distributed\distributed\worker.py", line 434, in get_data
    'bandwidth': total_bytes / (stop - start)
ZeroDivisionError: float division by zero
```

mrocklin:
OK, I found it.  `bokeh.server.server.Server` creates a `bokeh.server.tornado.BokehTornado` which, on construction adds a signal handler for `SIGTERM` that calls `self.stop`, which stops the `IOLoop`.

Reported to Bokeh a couple days ago here: https://github.com/bokeh/bokeh/issues/5494

cc @bryevdv

PR #713, opened 2016-12-01
Add status monitor to worker

mrocklin:
This adds a simple status monitor to all Server objects.  It serves that data on the worker through Bokeh

### DIFF
--- a/distributed/system_monitor.py
+++ b/distributed/system_monitor.py
@@ -1,3 +1,5 @@
+from __future__ import print_function, division, absolute_import
+
 from collections import deque
 import psutil
 from time import time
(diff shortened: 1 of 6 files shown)


distributed/system_monitor.py
```diff
@@ -0,0 +1,53 @@
...
+        self.time.append(now)
+
+        self.read_bytes.append(read_bytes)
+        self.write_bytes.append(write_bytes)
+
+        self.count += 1
+
+        return {'cpu': cpu, 'memory': memory, 'num_fds': num_fds, 'time': now,
+                'count': self.count, 'read_bytes': read_bytes, 'write_bytes':
+                write_bytes}
```
mrocklin: @pitrou is there anything else you think it would be useful to track over time on each of the workers?

mrocklin:
## Future Work

- [ ] Legends for in/out transfer plots
- [ ] Cross filter scale_width layout
- [ ] Hover tips for cross filter

PR #714, opened 2016-12-01
Fix #712: better Windows timer

### LINKED ISSUE #712: Need Window Timer
We need a timer for Windows that has the following properties:

1.  Fast to run (microsecond or less) - like `time.time` or `timeit.default_timer`
2.  Roughly globally consistent (to within minutes) - like `time.time` but not `timeit.default_timer`
3.  Microsecond precision - like `timeit.default_timer`, but not `time.time`

I wonder if there is a way to combine `time.time` and `timeit.default_timer` to achieve the desired result.  Or is there an established alternative?

cc @pitrou @jreback

pitrou:
I only changed the Worker to use this new timer, since that's the only place where we seem to desire better timer resolution.

### DIFF
--- a/distributed/metrics.py
+++ b/distributed/metrics.py
@@ -3,7 +3,7 @@ from __future__ import print_function, division, absolute_import
 import collections
 from functools import wraps
 import sys
-import time
+import time as timemod
 
 
 _empty_namedtuple = collections.namedtuple("_empty_namedtuple", ())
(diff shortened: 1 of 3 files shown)


mrocklin:
Thoughts on naming this just `time`?  It's one less term and true in the common case.

mrocklin:
It would also allow us to easily change existing code without renaming `time` -> `wall_clock` everywhere.

PR #717, opened 2016-12-03
Scheduler redesign

### LINKED ISSUE #711: Redesign task placement and work stealing
Now that the workers are more intelligent we may want to revisit some of our scheduling policies.
In particular if we give workers more tasks at once they can be more efficient, for example by batching communications and overlapping that time with computation of other tasks.  

Currently the scheduler only feeds workers as many tasks as they have cores (plus some buffer), and keeps the rest of the tasks waiting on the scheduler in the `.stacks`.  This was nice because we used those stacks to inform work stealing when some workers went idle while others were still saturated.  Now we might want to remove these stacks, and instead send all tasks to the workers immediately once they're ready.  The scheduler will still be aware of when work stealing is necessary (some workers have too many tasks while some workers have too few) but won't have the information necessary to decide which tasks to move.  However this decision can probably be negotiated among the workers themselves.  Moving this decision to the workers also lets us be less careful when it comes to efficiency; we no longer have to operate in linear time.

This removes state like the following from the scheduler:

* **stacks:** ``{worker: [keys]}``:
    List of keys waiting to be sent to each worker
* **stack_durations:** ``{worker: [ints]}``:
    Expected durations of stacked tasks
* **stacks_duration:** ``{worker: int}``:
     Total duration of all tasks in each workers stack

It adds a new scheduler-worker message about moving tasks and adds a new handshake between workers.

Additionally host or worker constraints will be somewhat problematic.  The scheduler knows the constraints of each task but the workers may not.  They'll have to check the scheduler for this information.

mrocklin:
This is a significant change to the scheduler.  In an effort to saturate hardware we move many more tasks onto the workers at once, rather than buffering them on the scheduler.  This allows workes allrs to more consistently saturate their CPUs and their network.  However it also limits the ability of the scheduler to perform load balancing (`stacks`, `ready`, etc..)  Instead we now rely on the workers to steal data among themselves.  

The scheduler periodically checks task occupancy on all workers and recommends that some workers steal from certain others.  The workers negotiate the transfer of tasks themselves and report their decision back to the scheduler.  Decisions are less globally optimal but can now take more time because the scheduling logic has been somewhat distributed to the workers themselves.

Fixes #711


distributed/scheduler.py
```diff
@@ -1550,57 +1524,59 @@ def replicate(self, stream=None, keys=None, n=None, workers=None,
                     self.add_keys(address=w, keys=list(gathers[w]))
 
     def workers_to_close(self, memory_ratio=2):
```
pitrou: Unrelated to the changes in this PR, but I feel a docstring and/or explanation at top would make this method more understandable.

mrocklin: Any feedback is welcome.

distributed/scheduler.py
```diff
@@ -2978,40 +2784,52 @@ def start_ipython(self, stream=None):
             )
         return self._ipython_kernel.get_connection_info()
 
+    def worker_objective(self, key, worker):
+        """
+        Objective function to determine which worker should get the key
+
+        Minimize expected start time.  If a tie then break with data storate.
+        """
+        comm_bytes = sum([self.nbytes.get(k, 1000) for k in self.dependencies[key]
```
pitrou: I think this is not the first time the `1000` default value pops up, perhaps a constant would make things easier?

distributed/scheduler.py
```diff
@@ -3034,46 +2851,46 @@ def decide_worker(dependencies, stacks, stack_duration, processing, who_has,
...
-        stack_time = stack_duration[w] / ncores[w]
-        start_time = comm_bytes / BANDWIDTH + stack_time
-        return start_time
+    if objective is None:
+        def objective(w):
+            comm_bytes = sum([nbytes.get(k, 1000) for k in dependencies[key]
+                              if w not in who_has[k]])
+            stack_time = occupancy[w] / ncores[w]
+            start_time = comm_bytes / BANDWIDTH + stack_time
+            return (start_time, len(has_what[w]))
```
pitrou: This is disturbingly similar to the `worker_objective()` method, except that it uses `len(has_what[worker])` as a proxy for memory occupation. Is this deliberate?

mrocklin: This was a crutch while I rewrote scheduler tests.  I forgot to remove it afterwards.  Thanks for pointing it out.  Removed.

PR #729, opened 2016-12-06
Channels, and scheduler extensions

mrocklin:
Motivation
----------

This PR adds a basic mechanism for multiple clients to communicate between each other with a shared sequence of futures onto which any client can append and any client can view.  This opens up some opportunities for complex workloads, especially when paired with [local_client](http://distributed.readthedocs.io/en/latest/task-launch.html#submit-tasks-from-worker).

Additionally, this PR attempts to add this new abstraction separately from the scheduler.  We create a pair of objects for the scheduler and client with their own set of messages that they pass back and forth along the normal communication channels.    This isolates this logic to a single place, keeps the scheduler uncluttered, and provides an opening for future experimentation for different (hopefully better) solutions to similar problems.

So again, two objectives:

1.  Direct objective to implement a shared sequence of futures between multiple clients
2.  Meta objective to establish a pattern to extend the scheduler cleanly

Now, more details about channels

Channel
-------

A changing stream of futures shared between clients

Several clients connected to the same scheduler can communicate a sequence
of futures between each other through shared *channels*.  All clients can
append to the channel at any time.  All clients will be updated when a
channel updates.  The central scheduler maintains consistency and ordering
of events.

Examples
--------

Create channels from your Client:

```python
>>> client = Client('scheduler-address:8786')
>>> chan = client.channel('my-channel')
```

Append futures onto a channel

```python
>>> future = client.submit(add, 1, 2)
>>> chan.append(future)
```

A channel maintains a collection of current futures added by both your
client, and others.

```python
>>> chan.futures
deque([<Future: status: pending, key: add-12345>,
       <Future: status: pending, key: sub-56789>])
```

You can iterate over a channel to get back futures.

```python
>>> for future in chan:
...     pass
```

### Example with `local_client`

```python
import random, time, operator
from distributed import Client, local_client

def emit(name):
    with local_client() as c:
       chan = c.channel(name)
       while True:
           future = c.submit(random.random, pure=False)
            chan.append(future)
            time.sleep(1)

def combine():
    with local_client() as c:
        a_chan = c.channel('a')
        b_chan = c.channel('b')
        out_chan = c.channel('adds')
        for a, b in zip(a_chan, b_chan):
            future = c.submit(operator.add, a, b)
            out_chan.append(future)

client = Client()

emitters = (client.submit(emit, 'a'), client.submit(emit, 'b'))
combiner = client.submit(combine)
chan = client.channel('adds')


for future in chan:
    print(future.result())
   ...:     
1.782009416831722
...
```

### TODO

- [ ] Documentation
- [ ] Some sort of `StopIteration` and `Exception` signal
- [ ] A proper asynchronous API

### DIFF
--- a/distributed/channels.py
+++ b/distributed/channels.py
@@ -34,6 +34,7 @@ class ChannelScheduler(object):
                     'channel-append': self.append}
 
         self.scheduler.compute_handlers.update(handlers)
+        self.scheduler.extensions['channels'] = self
 
     def subscribe(self, channel=None, client=None, maxlen=None):
         logger.info("Add new client to channel, %s, %s", client, channel)
@@ -86,7 +87,7 @@ class ChannelClient(object):
     def __init__(self, client):
         self.client = client
         self.channels = dict()
-        self.client._channel_handler = self
+        self.client.extensions['channels'] = self
 
         handlers = {'channel-append': self.receive_key}
 
--- a/distributed/channels.py
+++ b/distributed/channels.py
@@ -106,6 +106,40 @@ class ChannelClient(object):
 
 
 class Channel(object):
+    """
+    A changing stream of futures shared between clients
+
+    Several clients connected to the same scheduler can communicate a sequence
+    of futures between each other through shared *channels*.  All clients can
+    append to the channel at any time.  All clients will be updated when a
+    channel updates.  The central scheduler maintains consistency and ordering
+    of events.
+
+    Examples
+    --------
+
+    Create channels from your Client:
+
+    >>> client = Client('scheduler-address:8786')
+    >>> chan = client.channel('my-channel')
+
+    Append futures onto a channel
+
+    >>> future = client.submit(add, 1, 2)
+    >>> chan.append(future)
+
+    A channel maintains a collection of current futures added by both your
+    client, and others.
+
+    >>> chan.futures
+    deque([<Future: status: pending, key: add-12345>,
+           <Future: status: pending, key: sub-56789>])
+
+    You can iterate over a channel to get back futures.
+
+    >>> for future in chan:
+    ...     pass
+    """
     def __init__(self, client, name, maxlen=None):
         self.client = client
         self.name = name
@@ -120,6 +154,7 @@ class Channel(object):
                                         'client': self.client.id})
 
     def append(self, future):
+        """ Append a future onto the channel """
         self.client._send_to_scheduler({'op': 'channel-append',
                                         'channel': self.name,
                                         'key': tokey(future.key)})
@@ -138,6 +173,9 @@ class Channel(object):
             self._thread_condition.notify_all()
 
     def flush(self):
+        """
+        Wait for acknowledgement from the scheduler on any pending futures
+        """
         while self._pending:
             sleep(0.01)
 
--- a/distributed/channels.py
+++ b/distributed/channels.py
@@ -1,5 +1,8 @@
+from __future__ import print_function, division, absolute_import
+
 from collections import deque, defaultdict
 from functools import partial
+import logging
 from time import sleep
 import threading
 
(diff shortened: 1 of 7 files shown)


mrocklin:
cc @jcrist @martindurant @quasiben @pitrou

Also cc @dan-blanchard I hear that this sort of thing might interest you.  I would be very interested in more useful ways to set up this sort of abstraction (we have a lot of freedom here).

martindurant:
Is there a specific use case in mind for sharing futures? I don't immediately see what the examples achieve.

Does this supercede the dataset publication, which also recreates futures within clients, but in an ad-hoc way?

mrocklin:
Channels are, I think, sort of a minimum viable streaming abstraction.  For example if you wanted to watch live data on some ever-changing stream (like a Kafka queue) then this would help.  Dataset publication wasn't designed with rapid changes, getting a backlog of history, etc. in mind.

I think that the dataset publication mechanism would be a good candidate to be re-implemented in this way.  It's a self-contained set of messages and state in the scheduler that could probably be pulled out to a plugin like this.  I think that this would be useful to make both the scheduler and the dataset publication code easier to maintain.

There are probably several other things like Channels and dataset publication out there.  This gets into the "meta" goal of the PR, to establish a convention for adding similar systems.

mrocklin:
The approach taken here (not channels, but an as-yet-unwritten peer to channels) would also be sufficient, I think, to implement Rx cc @rgbkrk

martindurant:
Could channels be used to emulate Spark's "accumulator" concept: a global shared quantity that workers can add to in a associative way.

mrocklin:
I don't think so.  Channels move futures (abstract things) between clients.  Spark's accumulators maintain state on a per-worker basis.  

I'm confident that Dask needs the accumulator optimization though.

martindurant:
If I understand accumulators, they only maintain one global state each on the whole cluster, the workers merely issue messages with a value to add into the accumulator. This is probably off-topic, though, for this PR.

pitrou:
Is it some kind of publish-subscribe mechanism you have in mind?

mrocklin:
> If I understand accumulators, they only maintain one global state each on the whole cluster, the workers merely issue messages with a value to add into the accumulator. This is probably off-topic, though, for this PR.

One could build something to maintain global state that supported commutative changes, yes.

mrocklin:
> Is it some kind of publish-subscribe mechanism you have in mind?

Not sure if this is directed at me.  If so then yes, I suppose that this PR implements something similar to pub-sub.  If we wanted a somewhat more pure pub-sub system it would be very similar to this I think.  

The larger point here is that we can extend the scheduler and client with patterns like pub-sub without actually diving into the scheduler or client code much.

rgbkrk:
As a light implementation, I like what's present here for the channels as a stream of futures. I love having this layer of abstraction.

On the pseudo-Rx side, do you think this implementation will lend well in the future to being able to compose the overall stream with operators declaratively?

After reading this PR I finally went through and looked more deeply into how RxPy works and this is the most interesting example I can see that uses a scheduler: https://github.com/ReactiveX/RxPY/blob/master/examples/autocomplete/autocomplete_asyncio.py

I also tried my hand with using raw Futures (which misses the benefit of a scheduler):

```
import concurrent
import time

import rx

tp = concurrent.futures.thread.ThreadPoolExecutor()

def slow_add(a,b):
    time.sleep(1)
    return a + b

obs = rx.Observable.merge(
    tp.submit(slow_add, 1, 2),
    tp.submit(slow_add, 2, 2),
    tp.submit(slow_add, 5, 4),
    tp.submit(slow_add, 1, 1),
)

odd = obs.filter(lambda x: x % 2 == 1)

def p(val):
    print("--> {}\n".format(val))

sub = odd.subscribe(p)
```

RxPy could certainly use with more narrative documentation, it's a little hard (even as an Rx user) how to map this into Python.

As for what's here, looking forward to trying it out!

PR #735, opened 2016-12-07
Break early if gathering cancelled result

mrocklin:
Previously in rare cases when a future was cancelled during a gather
the gather would not break.


mrocklin:
This also resolves an intermittent error that popped up on `test_stress_scatter_death`

PR #738, opened 2016-12-07
Add TDigest counters to worker and scheduler

mrocklin:
cc @jcrist this is why I want TDigest

image

### DIFF
--- a/distributed/bokeh/worker.py
+++ b/distributed/bokeh/worker.py
@@ -13,7 +13,8 @@ from bokeh.models import (
 )
 from bokeh.models.widgets import DataTable, TableColumn, NumberFormatter
 from bokeh.plotting import figure
-from toolz import frequencies
+from bokeh.palettes import viridis
+from toolz import frequencies, merge
 
 from .components import DashboardComponent
 from .core import BokehServer, format_bytes, format_time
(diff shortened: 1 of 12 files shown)


distributed/bokeh/worker.py
```diff
@@ -421,6 +421,40 @@ def update(self):
...
+        self.figures[name] = fig
+        fig.yaxis.visible = False
+        return fig
+
+    def update(self):
+        for name, figure in self.figures.items():
+            counter = self.server.counters[name]
+            xs = [counter.quantile(i / 100) for i in range(1, 100)]
+            try:
+                ys = [1 / (xs[i + 1] - xs[i]) for i in range(len(xs) - 1)]
```
mrocklin: I'm not sure that this is the right way to get a sense for the distribution visually out of the quantiles.  Recommendations welcome here.  In particular this fails when two centroids overlap.

mrocklin:
Updated image in header comment

mrocklin:
And a gif!

bokeh-counters

distributed/counter.py
```diff
@@ -0,0 +1,67 @@
...
+            for i in range(len(self.intervals) - 1):
+                frac = 0.2 * self.intervals[0] / self.intervals[i]
+                part = self.components[i].scale(frac)
+                rest = self.components[i].scale(1 - frac)
+
+                self.components[i + 1].merge(part)
+                self.components[i] = rest
+
+        def size(self):
+            return sum(d.size() for d in self.components)
```
mrocklin: @jcrist can I ask for your review on this?

mrocklin:
merging soon if no further comments

distributed/bokeh/worker.py
```diff
@@ -422,6 +424,123 @@ def update(self):
...
+
+    def update(self):
+        with log_errors():
+            for name, figure in self.digest_figures.items():
+                digest = self.server.digests[name]
+                d = {}
+                for i, d in enumerate(digest.components):
+                    if d.size():
+                        try:
+                            xs = [d.quantile(i / 100) for i in range(1, 99)]
```
jcrist: Here's one for a normal distribution:

hist2

It would be simple enough to create a `hist` method on `TDigest` that takes in a number of bins, or the bin locations, and produces the `y`s for the centers. I may add this anyway.

mrocklin:
That would be useful

distributed/bokeh/worker.py
jcrist: Done, there's now a `histogram` method that mirrors numpy's. Can now just do `t.histogram(nbins)`, `t.histogram(nbins, range=(min, max))`, or `t.histogram(bin_edges)`.

This returns `hist` and `bin_edges`, where `len(bin_edges) = len(hist) + 1`. To get the centers, do `mids = bin_edges[:-1] + (bin_edges[1:] - bin_edges[:-1]) / 2`. If you know the bin widths are all equal, then this simplifies down to `mids = bin_edges[:-1] + bin_width`.

PR #740, opened 2016-12-09
Scaling improvements

mrocklin:


distributed/protocol/core.py
```diff
@@ -8,7 +8,10 @@
 except ImportError:
     import pandas.msgpack as msgpack
 
-from toolz import get_in
+try:
+    from cytoolz import get_in
+except ImportError:
+    from toolz import get_in
```
eriknw: You know, if you required `toolz >= 0.8.1`, you could do `from tlz import get_in` instead of doing `try-except`.

PR #746, opened 2016-12-12
Add futures to requirements.txt for python 2.7

rbubley:
distributed/client.py  has a dependency on concurrent.futures; for python 2.7 the futures module seems to be needed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ psutil
 locket
 zict >= 0.1.0
 sortedcollections
+futures; python_version < '3.0'


mrocklin:
This *should* be handled already in setup.py.

Is the syntax you're using in requirements.py the standard syntax expected by pip?  I'm not familiar with it.

rbubley:
I see where it is in setup.py, but depending on the tools one uses, requirements.txt may be read instead.

The format used is as per the pip docs (http://pip.readthedocs.io/en/stable/reference/pip_install/#requirement-specifiers) – apparently it has worked since pip version 6.0 (December 2014)

mrocklin:
Alrighty then.  Merging.  Thanks @rbubley !

PR #749, opened 2016-12-13
Adding documentation for the new Channels feature

strets123:


docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
+Channels: Futures Shared Between Clients
+===================================
```
mrocklin: This bar should match the number of characters in the title

Rather than "Futures Shared" how about "Share Futures" to keep the voice active.

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Often it is desirable to respond to events from outside the distributed cluster 
+or to instantiate a new client in order to check on the progress of a set of tasks.
```
mrocklin: Would prefer to start with a one-line description of what channels do, then going into examples, then go into a greater description.  Your one line description below "A changing stream of futures shared between clients" looks great to me.  Can we elevate it to the first line?

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+===================================
+
+Often it is desirable to respond to events from outside the distributed cluster 
+or to instantiate a new client in order to check on the progress of a set of tasks. 
+The channels feature makes these and many other workflows possible. 
+It also allows the Dask Scheduler to be extended in a clean way using the normal
+Distributed task submission.
+
+Channel:
+--------
```
mrocklin: Prefer to avoid colons in titles

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+of events.
+
+Examples
+--------
+
+Create channels from your Client:
+
+.. code-block:: python
+>>> client = Client('scheduler-address:8786')
+>>> chan = client.channel('my-channel')
```
mrocklin: I expect sphinx to be upset with this formatting.  I usually have to do have an extra line between `code-block` and the code.  I also have to indent by three spaces.  I'm often wrong about sphinx though and would be happy is this was valid.

In case I'm not wrong, you should be able to build docs locally by 

```
cd docs
make html
```

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+
+You can ensure that a channel is fully processed using the flush method
+
+.. code-block:: python
+>>> client = Client('scheduler-address:8786')
+>>> chan = client.channel('my-channel')
+>>> future2 = client.submit(time.sleep,2)
+>>> chan.append(future2)
+>>> chan.flush()
+
```
mrocklin: We should probably also mention the `stop` method here.  You can close a channel for all clients.  This will also stop them from iterating (otherwise they continue forever).

We could also mention that `chan.futures` contains a Python `deque` if we want to capture a snapshot.

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Example with local_client
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Using channels with local client allows for a more decoupled version of what is possible with :doc:`Data Streams with Queues<queues>` in that independent worker clients can build up a set of results which can be read later by a different client. This opens up Dask/Distributed to being integrated in a wider application environment similar to other python task queues such as celery,
```
mrocklin: In an effort to be kind to other projects, we should probably capitalize Celery and provide a link

```
... such as celery_.

.. _celery: http://www.celeryproject.org/
```

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+from distributed import Client, local_client
+from tornado import gen
+
+def emit(name):
+    with local_client() as c:
+       chan = c.channel(name)
+       while True:
+           future = c.submit(random.random, pure=False)
+            chan.append(future)
+            gen.sleep(1)
```
mrocklin: This will need to be a normal `time.sleep`, not `gen.sleep`.  This function is not a tornado coroutine.

docs/source/channels.rst
```diff
@@ -0,0 +1,99 @@
...
+import random, time, operator
+from distributed import Client, local_client
+from tornado import gen
+
+def emit(name):
+    with local_client() as c:
+       chan = c.channel(name)
+       while True:
+           future = c.submit(random.random, pure=False)
+            chan.append(future)
```
mrocklin: unaligned indentation here

docs/source/channels.rst
strets123: OK, got it, I had misunderstood the local_client documentation.

strets123:
Thanks for reviewing, I have hopefully covered off most of your comments. A couple of questions:

Is there room for an example using the "Submitting Applications" functionality to listen for events or does this risk deadlocking the cluster.

What happens to a running loop and any active futures if stop is called or the deque emptied? Does the StopIteration exception need to be handled?

mrocklin:
> Is there room for an example using the "Submitting Applications" functionality to listen for events or does this risk deadlocking the cluster.

There is room for more examples.  I don't know of any mechanism that would cause a deadlock here.

> What happens to a running loop and any active futures if stop is called or the deque emptied? Does the StopIteration exception need to be handled?

If a channel is "stopped" then future appends will fail, the deques within channels will stop changing, and any future iteration will continue until it reaches the last appended future.  Deques will retain their futures, which will themselves persist in distributed memory until all those futures are forgotten on all clients and the scheduler.

strets123:
Ah,  I had not fully understood how Dask Remote works, thanks. I will create an example using this.

So to avoid [garbage collection](file:///work/distributed/docs/build/html/journey.html#step-8-garbage-collection) a future needs to be in scope of a client but it will be if the local client is currently iterating that channel. So effectively you did not have to change garbage collection to get channels working.

docs/source/channels.rst
```diff
@@ -0,0 +1,140 @@
...
+
+If you wish to persist the current status of tasks outside of the distributed
+cluster (e.g. to take a snapshot in case of shutdown) you can copy a channel full
+of futures as it is a python deque_.
+
+.. _deque: https://docs.python.org/3.5/library/collections.html#collections.deque`
+
+.. code-block:: python
+
+    channelcopy = copy.deepcopy(chan.futures)
```
mrocklin: This should work and is a bit simpler:

    futures = list(chan.futures)

docs/source/channels.rst
```diff
@@ -0,0 +1,140 @@
...
+You can iterate over a channel to get back futures.
+
+.. code-block:: python
+
+    >>> anotherclient = Client('scheduler-address:8786')
+    >>> chan = anotherclient.channel('my-channel')
+    >>> for future in chan:
+    ...     pass
+
+You can ensure that a channel is fully processed using the ``flush`` method
```
mrocklin: This only relevant when writing.  So perhaps "When done writing, call `flush` to wait until your appends have been fully registered with the scheduler."

docs/source/channels.rst
```diff
@@ -0,0 +1,140 @@
...
+
+    >>> client = Client('scheduler-address:8786')
+    >>> chan = client.channel('my-channel')
+    >>> future2 = client.submit(time.sleep,2)
+    >>> chan.append(future2)
+    >>> chan.flush()
+
+
+
+
```
mrocklin: Should probably have only a couple of lines of whitespace here.

docs/source/channels.rst
```diff
@@ -0,0 +1,140 @@
...
+Example with local_client
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Using channels with local client allows for a more decoupled version
```
mrocklin: Perhaps link here from `local_client` to http://distributed.readthedocs.io/en/latest/task-launch.html#submit-tasks-from-worker (or the appropriate in-document sphinx link as you do below with queues).  The casual reader may not know what "local client" means otherwise.

strets123:
I have added the local client link and a motivating example to further demonstrate the functionality.

docs/source/channels.rst
```diff
@@ -0,0 +1,253 @@
...
+        return data
+
+    def hold_completed_futures():
+        with local_client() as c:
+            output_channel = c.channel("output")
+            for future in output_channel:
+                print("holding a future")
+                time.sleep(3600)
+                if future.status != "finished":
+                    print(future.status)
```
mrocklin: If your goal is to ensure that futures stick around then you could probably also create a channel with `maxlen=None`

docs/source/channels.rst
```diff
@@ -0,0 +1,253 @@
...
+
+    client = Client('scheduler-address:8786')
+    chan = client.channel('input')
+    dask_graph = {
+                    "tracking_key_1" :(func, arg1, arg2),
+                    "tracking_key_2" :(func2, "tracking_key_1")
+                 }
+    future_list = client.scatter([("<worflow_run_id>", dask_graph)])
+    chan.append(future_list[0])
+    chan.flush()
```
mrocklin: Would it be reasonable to just submit the task now rather than send it to someone else to submit?

docs/source/channels.rst
strets123: Thanks so much for the tips and patience, I will try to clean this up and create a blog post in the next couple of weeks.

docs/source/channels.rst
strets123: Thinking about it, given that the get is happening on a remote worker, this makes more sense and combined with the channel maxlength should lead to a stable solution.

mrocklin:
I recommend that we drop the example for now and merge the explanatory text.

strets123:
OK, done, thanks again for your help.

mrocklin:
Merged.  Thanks @strets123 !

Issue #751, opened 2016-12-15
__str__() of Nanny can fail


rbubley:
In `nanny.py`:
`__str__()` of `Nanny` relies on `worker_address`, which in turn relies on `worker_port`. This is initially set to `None`, so it can result in:

`TypeError: %d format: a number is required, not NoneType`

mrocklin:
Indeed.  It would be good to relax that assumption.  

Can I interest you to submitting a small PR to resolve the issue?

mrocklin:
Resolved

PR #753, opened 2016-12-15
Update nanny.py to address Issue #751

rbubley:


mrocklin:
Nice.  Two comments:

1.  It might be nice to visually signify that the address is incomplete because we have not connected.  Perhaps we could put `None` in where the port usually is?  I think that we could accomplish this by changing the `%d` to `%s` in the original line
2.  Can you provide a tiny test that would have failed before?  Probably this just means adding a small `test_nanny_no_port` function in `tests/test_nanny.py` that creates a Nanny and calls `str` on it.

rbubley:
I though that reporting `<Nanny: 127.0.0.1:None, threads: 8>`, as per your suggestion, would be unclear. But feel free to amend the PR if you disagree.

mrocklin:
Fair point

So now we have `<Nanny: 127.0.0.1 (No worker port assigned), threads: 8>`

Which seems fine to me.  If there were a way to make this more concise so that it could more easily fit into interactive sessions or logs that would be nice, but I'm probably being overly picky.  Thoughts on any of the following?

    <Nanny: 127.0.0.1:no-port, threads: 8>
    <Nanny: 127.0.0.1:unassigned, threads: 8>
    <Nanny: 127.0.0.1:0000, threads: 8>

mrocklin:
Looks good to me.  Thanks for going the extra round.  Merged!

PR #756, opened 2016-12-20
Fix #731: add ability to run coroutine on workers

### LINKED ISSUE #731: Asynchronous long-running tasks on workers
Dask is able to reason about and run long-running tasks on workers that themselves launch other tasks.  This enables some powerful and novel workflows.  http://distributed.readthedocs.io/en/latest/task-launch.html#extended-example

This system launches a separate thread for the long-running task to divorce it from the normal thread pool.  It then creates a local Client object and uses the synchronous API with that client.  

There may be some benefit to launching a long-running task inside of the main worker thread, running on the same IOLoop as the administrative worker code.  This has a few potential benefits:

1.  It might be less hacky than the current separate thread solution
2.  It would give the long-running task thread-safe access to the worker
3.  It would enable experimentation with asynchronous APIs in Python 3.5, which might bear long term fruit.

This would be useful for example in cases where we know the worker is sitting on some other data source, like a Kafka queue, and we want it to just pump that data into the system forever.  

cc @pitrou this sounds like the sort of thing that might interest you.

pitrou:
The added API mirrors `Client.run` (though without the `nanny` parameter as I'm not sure whether that's useful). The API returns the coroutine's return value from each worker, though of course the caller can just as well ignore that.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1291,18 +1291,23 @@ class Client(object):
     @gen.coroutine
     def _run_coroutine(self, function, *args, **kwargs):
         workers = kwargs.pop('workers', None)
+        wait = kwargs.pop('wait', True)
         responses = yield self.scheduler.broadcast(msg=dict(op='run_coroutine',
                                                 function=dumps(function),
                                                 args=dumps(args),
-                                                kwargs=dumps(kwargs)),
+                                                kwargs=dumps(kwargs),
+                                                wait=wait),
                                                 workers=workers)
-        results = {}
-        for key, resp in responses.items():
-            if resp['status'] == 'OK':
-                results[key] = resp['result']
-            elif resp['status'] == 'error':
-                six.reraise(*clean_exception(**resp))
-        raise Return(results)
+        if not wait:
+            raise Return(None)
+        else:
+            results = {}
+            for key, resp in responses.items():
+                if resp['status'] == 'OK':
+                    results[key] = resp['result']
+                elif resp['status'] == 'error':
+                    six.reraise(*clean_exception(**resp))
+            raise Return(results)
 
     def run_coroutine(self, function, *args, **kwargs):
         """
@@ -1319,6 +1324,8 @@ class Client(object):
              a Python 3.5+ async function)
         *args: arguments for remote function
         **kwargs: keyword arguments for remote function
+        wait: boolean (default True)
+            Whether to wait for coroutines to end.
         workers: list
             Workers on which to run the function. Defaults to all known workers.
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1288,6 +1288,43 @@ class Client(object):
         """
         return sync(self.loop, self._run, function, *args, **kwargs)
 
+    @gen.coroutine
+    def _run_coroutine(self, function, *args, **kwargs):
+        workers = kwargs.pop('workers', None)
+        responses = yield self.scheduler.broadcast(msg=dict(op='run_coroutine',
+                                                function=dumps(function),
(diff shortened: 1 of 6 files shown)


pitrou:
I'm going to add a test with a `async def` function.

distributed/client.py
```diff
@@ -1288,6 +1288,43 @@ def run(self, function, *args, **kwargs):
...
+        function: a coroutine function
+            (typically a function wrapped in gen.coroutine or
+             a Python 3.5+ async function)
+        *args: arguments for remote function
+        **kwargs: keyword arguments for remote function
+        workers: list
+            Workers on which to run the function. Defaults to all known workers.
+
+        """
+        return sync(self.loop, self._run_coroutine, function, *args, **kwargs)
```
mrocklin: Is it possible to check the input function to see if it is a coroutine and, if so, use the `Client.run` method for both cases?

mrocklin:
This looks cool.  Eventually it would be useful to also submit coroutines as long-running tasks as well.  This would allow them to be tracked by the normal task-based system and gain access to all of the standard restrictions and load balancing features to apply as well.

distributed/client.py
pitrou: Only on Tornado 4.5+, so I think we still need a separate API.

pitrou:
> Eventually it would be useful to also submit coroutines as long-running tasks as well.

Ok, I'll take a look and file another PR when that's ready.

distributed/client.py
```diff
@@ -1288,6 +1288,43 @@ def run(self, function, *args, **kwargs):
...
+                                                function=dumps(function),
+                                                args=dumps(args),
+                                                kwargs=dumps(kwargs)),
+                                                workers=workers)
+        results = {}
+        for key, resp in responses.items():
+            if resp['status'] == 'OK':
+                results[key] = resp['result']
+            elif resp['status'] == 'error':
+                raise loads(resp['exception'])
```
mrocklin: Grep for `six.reraise` in the codebase.  We tend to also ship around the remote traceback.

distributed/tests/test_client.py
```diff
@@ -2188,6 +2188,38 @@ def func(x, y=10):
...
+def test_run_coroutine_sync(loop):
+    with cluster() as (s, [a, b]):
+        with Client(('127.0.0.1', s['port']), loop=loop) as c:
+            result = c.run_coroutine(geninc, 2, delay=0.01)
+            assert result == {'127.0.0.1:%d' % a['port']: 3,
+                              '127.0.0.1:%d' % b['port']: 3}
+
+            result = c.run_coroutine(geninc, 2,
+                                     workers=['127.0.0.1:%d' % a['port']])
+            assert result == {'127.0.0.1:%d' % a['port']: 3}
```
mrocklin: We may not want to block here.  I'm not sure what correct behavior should be.

distributed/client.py
pitrou: I didn't know tracebacks were shipped. Generally they aren't very picklable, but perhaps cloudpickle has special support for them?

distributed/tests/test_client.py
pitrou: Right, we could add an optional `wait` argument to avoid blocking.

distributed/client.py
mrocklin: We use https://pypi.python.org/pypi/tblib

pitrou: Neat, thanks.

mrocklin:
Looks good to me.  Merging soon if no further comments.

PR #779, opened 2017-01-03
use extras_require for conditional dependencies

### LINKED ISSUE #771: distributed-1.15.0rc1 seems wrongly requiring "futures" from a Python-3.6 installation
Collecting futures (from distributed>=1.14; extra == "complete"->dask[complete]->-r C:\Winpython\basedir36

minrk:
needed for wheels to properly embed [conditional dependencies](https://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies).

Turns

    req; python < '3.0'

into

    {':python < '3.0': [req]}

closes #771

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,3 +5,6 @@ versionfile_source = distributed/_version.py
 versionfile_build = distributed/_version.py
 tag_prefix =
 parentdir_prefix = distributed-
+
+[bdist_wheel]
+universal=1
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,18 @@ import sys
 import versioneer
 
 requires = open('requirements.txt').read().strip().split('\n')
+install_requires = []
+extras_require = {}
+for r in requires:
+    if ';' in r:
+        # requirements.txt conditional dependencies need to be reformatted for wheels
+        # to the form: `'[extra_name]:condition' : ['requirements']`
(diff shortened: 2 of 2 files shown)


minrk:
I also made the wheel universal here, because embedding these conditional dependencies correctly should mean there's no difference between the two wheels. Is there any other reason for the wheel to be different for py2/3?

setup.cfg
```diff
@@ -5,3 +5,6 @@ versionfile_source = distributed/_version.py
 versionfile_build = distributed/_version.py
 tag_prefix =
 parentdir_prefix = distributed-
+
+[bdist_wheel]
+universal=1
```
mrocklin: What does this do exactly?  When I build I ususally do `python setup.py install sdist bdist_wheel --universal`, which seems to suffice.  Honestly I'm not sure what purpose the setup.cfg file serves here.  My knowledge of Python build tools is embarrassingly low.

mrocklin:
Thank you for taking this on @minrk

setup.cfg
minrk: > When I build I ususally do python setup.py install sdist bdist_wheel --universal

It's another way of specifying the same flag in a more permanent way, so you don't have to remember to pass the argument every time. The resulting wheel is the same if you've been doing `--universal` already.

Issue #782, opened 2017-01-04
Bug: typo in Client.unpublish_dataset()


bschreck:
should be:

self.scheduler.publish_delete

not

self.scheduler.publist_delete

PR #784, opened 2017-01-04
fixed publist_delete : publish_delete typo in client.py

bschreck:
Fixes typo #782 that prevents using Client.unpublish_dataset()

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,0 +1,11 @@
+joblib >= 0.10.2
+mock >= 2.0.0
+pandas >= 0.19.2
+numpy >= 1.11.0
+bokeh >= 0.12.3
+requests >= 2.12.4
+pyzmq >= 16.0.2
+ipython >= 5.0.0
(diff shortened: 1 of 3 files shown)


mrocklin:
Thanks for the correction @bschreck .

It's a bit concerning that this didn't come up during tests.  I guess we only tested the asynchronous versions of these functions.  Do you have any interest in also submitting a small test that exercises this fix to make sure that something like this doesn't occur in the future?

bschreck:
Sure I'll put 1 or 2 in now. By the way, it looks like there are several testing dependencies not in requirements.txt. Would it make sense to have a dev-requirements.txt with Mock, pandas, joblib, etc?

mrocklin:
Adding dev-requirements sounds sensible to me

mrocklin:
This looks great to me.  Thank you for your work.

Looks like there was an unrelated testing failure.  I've restarted the test and will investigate separately.

PR #794, opened 2017-01-07
Ensure communicating continues even on bad key

mrocklin:
It would be useful to have a test for this.


mrocklin:
Actually, the thing to do here is probably to avoid using the `changed=` flag and instead remove keys from `data_needed` if their dependencies are still uncertain.  This would make the while loop much more authoritative and remove a lot of the uncertainty here.

PR #802, opened 2017-01-09
Use worker event loop in local client

mrocklin:
There were race conditions otherwise.


mrocklin:
Failures here are due to the recent bokeh update.

PR #803, opened 2017-01-10
Avoid start_loop keyword in Bokeh 0.12.4

mrocklin:


heroxbd:
This might not be correct.  bokeh.__versoin__ is conda_version instead of bokeh version.

heroxbd:
Also the string comparison does not work for 0.9.7:
`
bokeh.__version__ <= '0.12.3'`

```
python -c 'print "0.9.7" <= "0.12.3"'
False
```

Issue #776, opened 2017-01-02
Should we use ZeroMQ?


mrocklin:
I've been wrestling with network communication recently as our bottlenecks have become increasingly communication focused.  I'd like to get feedback from people like @minrk and @pitrou on whether or not it is worth exploring switching our communication stack over to ZeroMQ.

Our Communication Types
--------------------

Dask.distributed engages in the following kinds of communication.  Everything is point-to-point.

1.  Frequent small messages along long-running connections.  We might dump in thousands of 100 byte messages in a second.  Currently we batch manually.  Presumably ZeroMQ would do this for us allowing us to remove some of the `BatchedSend` infrastructure.
2.  Peer to peer data movement.  These are small-to-large messages (1kb to 10GB) along ephemeral worker-to-worker connections.  Sometimes these overlap.  Here we're concerned about not having this communication block or be blocked by other executing code, establishing connections quickly (or else caching connections in a way that respects open file limits), and possibly negotiating multiple messages at once on the same socket.
3.  Periodic small messages for heart beats.

Advantages of ZeroMQ
---------------------

1.  We get to remove `BatchedSend` and timeout handling
2.  We get zero-copy for free (though this is probably not that big of a deal)
3.  We maybe get Infiniband for free?
4.  We get pub-sub, though I'm not sure how we would use this currently?

Questions
--------

1.  How does ZeroMQ handle hard failures on the other end of the socket?
2.  We are sometimes bound by open file handles.  Does one Request-Reply socket live on one TCP Socket?
3.  We currently suffer a delay of 10ms-100ms to establish TCP connections through Tornado.  Is this likely to change with ZeroMQ?  (Though in truth this is more likely something we're doing wrong internally)
4.  Are there opportunities for ZeroMQ to hold on to "closed" connections in the background while still respecting open file limits?  (I suppose we could also do this connection pooling in application code)
4.  Assuming that we're using pyzmq's Tornado connection, is there anything we should be concerned about if the event loop is blocked for a non-trivial amount of time?  Obviously we can't dump any new messages during this time, but is ZeroMQ handling things nicely in the background in a separate system thread or is it blocked?
5.  If we install pyzmq's event loop can we still overlap with other Tornado applications like Bokeh?

Implementation
---------------

What is the right way to experiment with this?  Presumably we change around `core.connect`, `core.read` and `core.write` and we switch out all `BatchedSend` streams with normal streams.  Is there anything else?  

[...]

PR #810, opened 2017-01-12
IO layer refactor

pitrou:
This refactor creates a distinct `distributed.comm` namespace which hosts an abstracted communications layer, mediated using URIs and several top-level functions (mainly `connect()` and `listen()`, plus some address-handling functions). The default transport is `tcp` which handles both IPv4 and IPv6. `zmq` is provided as a proof-of-concept but no effort has been done (yet?) to prove that it works with distributed at large: it is only tested using the internal communications API.

A number of small issues are also addressed on the side. One notable issue is the fact that the scheduler and workers would always listen on all addresses, even if asked to identify as `127.0.0.1`.


distributed/comm/tcp.py
```diff
@@ -0,0 +1,254 @@
...
+                         timeout * 1000)
+            TCP_USER_TIMEOUT = 18  # since Linux 2.6.37
+            sock.setsockopt(socket.SOL_TCP, TCP_USER_TIMEOUT, timeout * 1000)
+    except EnvironmentError as e:
+        logger.warn("Could not set timeout on TCP stream: %s", e)
+
+
+class TCP(Comm):
+
+    def __init__(self, stream, peer_addr, deserialize=True):
```
mrocklin: Should comms think about deserialization?

distributed/comm/zmq.py
```diff
@@ -0,0 +1,218 @@
...
+    return sock
+
+
+def set_socket_options(sock):
+    """
+    Set common options on a ZeroMQ socket.
+    """
+    # XXX set context default options instead?
+    sock.set(zmq.RECONNECT_IVL, -1)  # disable reconnections
+    sock.set(zmq.IPV6, True)
```
mrocklin: I'm generally curious about this choice.  What was the motivation?

distributed/comm/zmq.py
```diff
@@ -0,0 +1,218 @@
...
+    @gen.coroutine
+    def read(self, deserialize=None):
+        if self.sock is None:
+            raise CommClosedError
+        if deserialize is None:
+            deserialize = self.deserialize
+
+        frames = yield self.sock.recv_multipart(copy=False)
+        msg = from_frames([f.buffer for f in frames], deserialize=self.deserialize)
+        raise gen.Return(msg)
```
mrocklin: Well that was easy :)

distributed/comm/zmq.py
```diff
@@ -0,0 +1,218 @@
...
+            assert req['op'] == 'zmq-connect'
+
+            cli_sock = make_socket(zmq.DEALER)
+            set_socket_options(cli_sock)
+            cli_port = bind_to_random_port(cli_sock, self.ip)
+
+            resp = {'zmq-url': make_zmq_url(self.ip, cli_port)}
+            yield self.sock.send_multipart([envelope] + to_frames(resp))
+
+            address = 'zmq://<unknown>'  # XXX
```
mrocklin: Is the address the same as `envelope` above?

distributed/comm/zmqimpl.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        #print("-- remove_handler:", self._sock)
+        self.io_loop.remove_handler(self._sock)
+
+
+class Context(zmq.Context):
+
+    @property
+    def _socket_class(self):
+        return _AsyncSocket
```
mrocklin: Is this file re-implementing what you found to be inefficient in pyzmq?

distributed/scheduler.py
```diff
@@ -2733,34 +2749,43 @@ def remove_resources(self, worker):
...
-        Coerce possible input addresses to canonical form
+        Coerce possible input addresses to canonical form.
+        *resolve* can be disabled for testing with fake hostnames.
 
-        Handles lists, strings, bytes, tuples, or aliases
+        Handles strings, tuples, or aliases.
         """
-        if isinstance(addr, list):
-            addr = tuple(addr)
+        # XXX how many address-parsing routines do we have?
```
mrocklin: Yeah, sorry about this.  There is some history here that is no longer relevant.  I also wouldn't mind reducing the kinds of inputs this takes.

mrocklin:
The approach here seems reasonable to me.  Thank you for pushing early; I was curious.

distributed/comm/zmq.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @gen.coroutine
+    def read(self, deserialize=None):
+        if self.sock is None:
+            raise CommClosedError
+        if deserialize is None:
+            deserialize = self.deserialize
+
+        frames = yield self.sock.recv_multipart(copy=False)
+        msg = from_frames([f.buffer for f in frames], deserialize=self.deserialize)
```
minrk: deserialize=deserialize

pitrou: Woo, thank you.

distributed/comm/zmqimpl.py
pitrou: Yes, and also avoids https://github.com/zeromq/pyzmq/issues/963

distributed/comm/zmq.py
pitrou: No, `envelope` is some opaque endpoint id attributed by ZeroMQ.

distributed/comm/zmq.py
pitrou: I'm hoping this way ZeroMQ might actually inform us of disconnects. Though at this point I'm considering the ZeroMQ transport a mere experiment, as ZeroMQ's characteristics doesn't make it very friendly to work with.

distributed/comm/tcp.py
pitrou: I got the impression that some comms were deserialize-always and others deserialize-never, though I've also kept the deserialize argument on `read()` for now. Which approach do you think is better?

If you're talking about the more general approach of pickling and unpickling in read() and write(), yes, I think it's desirable: imagine we write a new transport usable between threads of the same process, in this case we won't want to pickle anything, simple pass objects around in a queue.

distributed/comm/zmq.py
minrk: You won't be informed of disconnects by disabling reconnects. Sockets will simply stop triggering POLL events. I wouldn't recommend using this setting. If you want to be notified of such events, the `monitor` system must be used.

distributed/comm/zmq.py
minrk: zmq calls the routing id 'IDENTITY', and is a uid for routing messages to peers. It can be set manually with `socket.IDENTITY = b'whoiam', but you have to be careful to avoid collisions if you do this. The default is either a monotonic counter or random id.

distributed/comm/zmqimpl.py
minrk: I hope I can look forward to a PR to pyzmq when this is done!

distributed/comm/tcp.py
mrocklin: Ah, yes, good point.

mrocklin:
Whoa, tests are passing :)

distributed/scheduler.py
```diff
@@ -367,11 +348,42 @@ def identity(self, stream):
...
+                service = v(self, io_loop=self.loop)
+                service.listen((listen_ip, port))
+                self.services[k] = service
+            except Exception as e:
+                logger.info("Could not launch service: %r", (k, port),
+                            exc_info=True)
+
+    def stop_services(self):
+        for service in self.services.values():
+            service.stop()
```
mrocklin: We might eventually be able to unify this on `Server`.  Both Scheduler and Worker should have similar needs here.

mrocklin: I refer both to start_services and stop_services

distributed/scheduler.py
```diff
@@ -392,10 +404,19 @@ def start(self, port=8786, start_queues=True):
...
+                self.listen(('', addr_or_port))
+                self.start_services()
+            else:
+                self.listen(addr_or_port)
+                self.ip, _ = get_address_host_port(self.listen_address)
+                self.start_services(self.ip)
 
             self.status = 'running'
-            logger.info("  Scheduler at: %20s:%s", self.ip, self.port)
+            logger.info("  Scheduler at: %24s", self.address)
```
mrocklin: This may not be aligned 

```
mrocklin@carbon:~$ dask-scheduler 
distributed.scheduler - INFO -   Scheduler at:  tcp://10.22.112.88:8786
distributed.scheduler - INFO -        http at:              0.0.0.0:9786
distributed.scheduler - INFO -       bokeh at:              0.0.0.0:8788
distributed.bokeh.application - INFO - Web UI: http://127.0.0.1:8787/status/
```

distributed/tests/test_stress.py
```diff
@@ -179,7 +179,7 @@ def test_stress_communication(c, s, *workers):
         pytest.skip("file descriptor limit too low and can't be increased :"
                     + str(e))
 
-    n = 40
+    n = 20
```
mrocklin: Why this change?

distributed/worker.py
```diff
@@ -368,11 +388,10 @@ def get_data(self, stream, keys=None, who=None):
             self.digests['get-data-load-duration'].add(stop - start)
         start = time()
         try:
-            compressed = yield write(stream, msg)
-            yield close(stream)
```
mrocklin: Was this unnecessary before as well?

mrocklin:
Do the ZeroMQ comms work now?  If so it would be useful to see basic tests with them.  I would also play with this now if it was clear how to set it up.

distributed/tests/test_stress.py
pitrou: `n = 40` made it too slow to run on my machine. I can change it back, though.

distributed/worker.py
pitrou: Actually, I think I've removed it by mistake. Let me see.

pitrou: Scratch that. If I add the `close` back, a timeout error is raised in `test_client_with_scheduler`.

distributed/scheduler.py
```diff
@@ -51,6 +53,9 @@
 DEFAULT_DATA_SIZE = config.get('default-data-size', 1000)
 
 
+# XXX avoid inheriting from Server? there is some large potential for confusion
+# between base and derived attribute namespaces...
```
pitrou: Scheduler and other heavy objects (objects with many of their own attributes and methods) inherit from `Server`, which makes it more confusing when looking up attributes (does it stem from Scheduler or Server?) and also makes it possible to shadow some Server attributes with some Scheduler attributes by mistake (that may actually have been the case with `address` or `port`). Using composition would solve that issue.

distributed/nanny.py
```diff
@@ -418,18 +346,12 @@ def run():
...
-        return proc.is_alive()
+    return proc is not None and proc.is_alive()
 
 
 def join(proc, timeout):
-    if proc is None or isinstance(proc, subprocess.Popen):
-        return
-    proc.join(timeout)
+    if proc is not None:
+        proc.join(timeout)
```
pitrou: They still handle the case where proc is `None`. I don't know if that's a good enough motivation.

distributed/core.py
```diff
@@ -784,6 +660,6 @@ def clean_exception(exception, traceback, **kwargs):
         exception = protocol.pickle.loads(exception)
     if isinstance(traceback, bytes):
         traceback = protocol.pickle.loads(traceback)
-    if isinstance(traceback, str):
-        traceback = None
+    elif isinstance(traceback, string_types):
+        traceback = None  # why?
```
pitrou: I see. I'll update the comment.

distributed/core.py
```diff
@@ -615,30 +487,35 @@ class ConnectionPool(object):
...
     Parameters
     ----------
     limit: int
-        The number of open streams to maintain at once
+        The number of open comms to maintain at once
     deserialize: bool
         Whether or not to deserialize data by default or pass it through
     """
     def __init__(self, limit=512, deserialize=True):
+        # Total number of open comms
```
pitrou: Ok, fair enough.

distributed/core.py
```diff
@@ -118,93 +128,113 @@ def __init__(self, handlers, max_buffer_size=MAX_BUFFER_SIZE,
...
+        """
+        The port number this Server is listening on.
+        """
         if not self._port:
-            try:
-                self._port = first(self._sockets.values()).getsockname()[1]
-            except StopIteration:
-                raise OSError("Server has no port.  Please call .listen first")
+            _, loc = parse_address(self.address)
+            _, self._port = parse_host_port(loc)
```
pitrou: No, not always. This part isn't properly abstracted. One would have to review if the `self.port` attribute is still used, and whether/how to eliminate it...

distributed/comm/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from .core import connect, listen, CommClosedError
+# Register transports
+from . import tcp
+try:
+    import zmq as _zmq
+except ImportError:
+    pass
+else:
+    # XXX this registers the ZMQ event loop, event if ZMQ is unused...
+    from . import zmq
```
pitrou: I haven't seen any when running a very simple TCP benchmark. At the bottom of each event loop is a simple, relatively cheap, system call (either poll() or epoll()).

distributed/client.py
```diff
@@ -354,9 +356,10 @@ def __init__(self, address=None, start=True, loop=None, timeout=3,
         self._pending_msg_buffer = []
         self.extensions = {}
 
-        if hasattr(address, 'scheduler_address'):
+        if isinstance(address, Cluster):
```
pitrou: Hm, I see.

distributed/cli/tests/test_dworker.py
```diff
@@ -60,20 +59,20 @@ def test_no_nanny(loop):
 @pytest.mark.parametrize('nanny', ['--nanny', '--no-nanny'])
 def test_no_reconnect(nanny, loop):
     with popen(['dask-scheduler', '--no-bokeh']) as sched:
-        wait_for_port('127.0.0.1:8786')
-        with popen(['dask-worker', '127.0.0.1:8786', '--no-reconnect', nanny,
+        wait_for_port(('127.0.0.1', 8786))
+        with popen(['dask-worker', 'tcp://127.0.0.1:8786', '--no-reconnect', nanny,
```
pitrou: Yes, it is.

mrocklin:
Just ran this on a cluster on my standard set of demonstrations.  Everything worked fine except that the notebook progress bars did not display.  Looking through the changes to distributed/diagnostics now.

mrocklin:
Just ran this on a cluster on my standard set of demonstrations.  Everything worked fine except that the notebook progress bars did not display.  Looking through the changes to distributed/diagnostics now.

mrocklin:
OK, the progress bar thing may be related to the fact that `IOLoop.current()._running` is now `False` when called in the method `MultiProgressWidget._ipython_display_`.  Previously loop was the currently running Jupyter IOLoop, which we could expect to be running.  I'm not sure why this would have changed.

```python
    def _ipython_display_(self, **kwargs):
        IOLoop.current().add_callback(self.listen)  # IOLoop.current()._running is False
        return self.widget._ipython_display_(**kwargs)
```

@minrk

pitrou:
> I'm not sure why this would have changed.

This may be because the ZMQ IO loop is being used. What happens if you comment out the zmq import lines in `distributed/comm/__init__.py`?

mrocklin:
Yes, removing the zmq import resolves things.  It also resolves the lingering zombie workers issue.

I do get this undesired log message

```python
distributed.core - WARNING - Lost connection to 'tcp://127.0.0.1:46144' while sending result for op 'feed': BrokenPipeError: [Errno 32] Broken pipe
```

pitrou:
It sounds like you shutdown the cluster while it still busy doing things? Otherwise it shouldn't get a connection close while sending an operation's result, I suppose.

mrocklin:
@pitrou any feedback on the following comments:

Do the ZeroMQ comms work now? If so it would be useful to see basic tests with them. I would also play with this now if it was clear how to set it up.


I think that it would be educational to try using this with some other comm system, either ZeroMQ or queues (if we make a queue system quickly). I expect some issues to arise when trying to use a new comm system.

pitrou:
To try out ZMQ, you can change the `DEFAULT_SCHEME` value in `distributed/comm/core.py` from "tcp" to "zmq". It seems to create some issues with msgpack ("can't serialize array").

pitrou:
Ok, I've fixed the issue. I can now run a benchmark similar to https://github.com/dask/distributed/issues/776#issuecomment-270000969 (I only adjusted the input sizes) and see that `zmq` is 15% faster than `tcp`.

mrocklin:
OK, cool.  I find it exciting that you are able to switch from TCP to ZeroMQ and have everything still work smoothly.

Can you share your setup for the benchmark?  I would like to try locally.

pitrou:
It's really your script as posted, I just reduced the size for `x` to have it run faster.

pitrou:
> OK, cool. I find it exciting that you are able to switch from TCP to ZeroMQ and have everything still work smoothly.

"Everything" is an overstatement. Some tests will definitely fail (the ZMQ layer doesn't detect hard connection closes, for example). Basic functionality should hopefully be there, though.

mrocklin:
By my benchmarks we're still spending a large amount of time waiting on `zmqimpl.py:160(_handle_send)`.  It takes around 10s over 8000 calls for me.  Is this backpressure from ZeroMQ?  Something else?

mrocklin:
OK, I did a bit more tweaking and benchmarking.  Here is what I did:

1.  I turned off all compression (set blosc to false in protocol/numpy.py and set `compression: null` in ~/.dask/config.yaml`)
2.  I removed the `ensure_bytes` call on all frames in protocol/core.py

I found that we spend the same amount of time sending bytes (around 15s on the original benchmark) in both ZeroMQ and TCP.

We spend far less time receiving bytes, around 14s in TCP and 2s in ZeroMQ.  We lose around 8 seconds of overhead in Tornado and around 2s of overhead by being able to avoid the memoryview->bytes copy in our own code.

The 15s wait on zeromq send is unfortunate.  Perhaps this is the no-copy issue in ZeroMQ?  Perhaps this is backpressure?

pitrou:
> By my benchmarks we're still spending a large amount of time waiting on zmqimpl.py:160(_handle_send). It takes around 10s over 8000 calls for me.

I can't reproduce that (I have 4 cores on a Linux machine). However, cProfile isn't threads-aware and that may be the effect of pyzmq releasing the GIL and letting another thread take it. cProfile would include the time waiting for the GIL to be available again in the function's runtime.

mrocklin:
I also have four cores on a Linux machine.

pitrou:
As an experiment, you could try to disable no-copy in `ZMQ.write()` in `distributed.comm.zmq`.

pitrou:
As an aside, for profiling, it could be nice to have a single-thread mode for workers, to avoid the GIL switching issues mentioned above. Perhaps by passing a special Executor?

mrocklin:
So workers would block when computing?

pitrou:
Yes, `submit()` could for example execute the function synchronously.

mrocklin:
Seems reasonable for debugging

mrocklin:
I'm digging into this a bit more and I find that I don't get the same slow communication I was getting when I wrote the "Should we use ZeroMQ" issue.  In that issue I was observing ~30-50 MB/s bandwidth between eight workers.  Now I observe closer to 100-200 MB/s, even without this PR.  I've gone back to git history on that day and confirmed the difference.

Bisecting history actually takes me all the way up to the current numpy memoryview PR.  I think that previously we were being very inefficient when determining if we should compress when using n-dimensional data.  We thought that we were taking out a small 10000 bytes sample when actually we were taking out a large 10000 row sample from the array and compressing it to determine if compressing was a good idea.

mrocklin:
Also, to share my benchmark, 

I create a scheduler

    dask-scheduler

I create eight worker processes

    dask-worker localhost:8786 --nprocs 8 --nthreads 1

I open up http://localhost:8789/main

And I run this script

```python
c = Client('localhost:8786')
import dask.array as da
x = da.random.randint(0, 255, size=(1000, 1000, 1000),
                                  chunks=(5, 1000, 1000)).astype('uint8')

x = c.persist(x)
wait(x)
f = c.compute(x.mean(axis=0))

wait(f)
```

mrocklin:
It would be nice to develop a plan to merge this PR.  Some questions:

1.  Do we want to merge the ZeroMQ work?
2.  How do we want to handle git history?  It seems wrong both to squash all of the commits into one and to merge them all as is.
3.  What happens next?
    *  Custom IOStream
    *  Try to improve Tornado's IOStream 
    *  Build Queue Comm for single-machine work
4.  Does this need developer docs?

@pitrou other thoughts?

pitrou:
1. I think it's beneficial to keep an example of another transport based on something with slightly different semantics. There's the question of whether we import it (since it replaces the event loop with ZMQ's).
2. I think squashing is fine.
3. I plan on 1) Queue comm 2) Custom IOStream
4. Thanks for reminding me. Yes we should at least update the docs in https://distributed.readthedocs.io/en/latest/foundations.html

pitrou:
I've added the docs. Also it seems test failures are gone now :-)

mrocklin:
I would like to issue a micro release before we merge this.  I will try to accomplish that today.

distributed/comm/core.py
```diff
@@ -0,0 +1,292 @@
...
+
+# Listener classes
+
+listeners = {
+    #'tcp': ...,
+    # 'zmq': ...,
+    }
+
+
+DEFAULT_SCHEME = 'tcp'
```
mrocklin: We might want to pull this value from `config`.  This would make it easier to switch to ZeroMQ without changing source.

distributed/comm/__init__.py
mrocklin: Given recent problems we probably don't want to import this by default.

distributed/comm/tcp.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    if exc.real_error is not None:
+        # The stream was closed because of an underlying OS error
+        exc = exc.real_error
+        raise CommClosedError("%s: %s" % (exc.__class__.__name__, exc))
+    else:
+        raise CommClosedError(str(exc))
+
+
+class TCP(Comm):
```
mrocklin: This should probably have a docstring

distributed/comm/core.py
```diff
@@ -0,0 +1,292 @@
...
+        The listening address as a URI string.
+        """
+
+    @abstractproperty
+    def contact_address(self):
+        """
+        An address this listener can be contacted on.  This can be
+        different from `listen_address` if the latter is some wildcard
+        address such as 'tcp://0.0.0.0:123'.
+        """
```
mrocklin: It might be good to add these classes and their methods as API documentation within the docs.  They provide an informative explanation of what one needs to do to implement another transport mechanism.

mrocklin: Ah, I see that you've done this already.

mrocklin:
It would be nice to remove this line from `distributed/protocol/core.py:dumps()`

    out_frames = [ensure_bytes(f) for f in out_frames]

This will probably require us to add an `ensure_bytes` call within the tcp `write` method.  This avoids an unnecessary memory copy for transports that support memoryviews.

distributed/protocol/core.py
```diff
@@ -75,7 +75,7 @@ def dumps(msg):
             header['keys'].append(key)
             out_frames.extend(frames)
 
-        out_frames = [ensure_bytes(f) for f in out_frames]
+        out_frames = [f for f in out_frames]
```
mrocklin: Can this line be removed?

distributed/comm/__init__.py
pitrou: So, now the question is how we test it... Do we use an environment variable, some kind of configuration option? I'm not sure we can safely change IOLoops during the test suite...

mrocklin:
In distributed/bokeh/scheduler.py we add a TapTool to a plot so that, whenever a user clicks on a rectangle corresponding to a worker, it redirects them to that worker's bokeh page.  The code to do this is currently written as follows:

                bokeh_addresses = [worker.split(':')[0] + ':' + str(wi[worker]['services']['bokeh'])
                                   for worker in workers]

This solution no longer works.  We find ourselves somewhere less-than-useful.

pitrou:
> This solution no longer works. We find ourselves somewhere less-than-useful.

Oh, sorry about that. I'll fix it.

mrocklin:
I've tried out the ZeroMQ solution on a larger problem on a cluster.  It is not robust and the cluster often deadlocks.  I suspect that we would run into errors on the `test_stress.py` tests.

pitrou:
The TapTool regression should be fixed now (I also added a test). Could you check?

distributed/comm/__init__.py
mrocklin: I suggest that we extend `distributed.config` to also collect all environment variables named `DASK_FOO` and put `foo` into the global config dict.  We could then test ZeroMQ by changing environment variables (which is easy on travis-ci).

I guess we would need to special case the import, `if default_transport == 'zmq': import ...`.

Thoughts?

pitrou:
Le 27/01/2017 à 18:13, Matthew Rocklin a écrit :
> 
> I suggest that we extend |distributed.config| to also collect all
> environment variables named |DASK_FOO| and put |foo| into the global
> config dict. We could then test ZeroMQ by changing environment variables
> (which is easy on travis-ci).

Ok, this sounds reasonable.  I've pushed an implementation of this, and
one Travis builds get DASK_EXPERIMENTAL_ZMQ=1.

mrocklin:
My guess is that ZeroMQ won't pass tests.  I don't think we necessarily need it to pass tests for us to merge this.  

I think that Travis has the ability to add new elements to the test matrix that don't need to pass to get a green light.

pitrou:
The option doesn't switch to default transport to ZMQ, it just enables the ZMQ transport and the ZMQ event loop. The tests have been passing before with that, so they shouldn't fail :-)

mrocklin:
Ah, I see

mrocklin:
Now that version 1.15.2 is released I would like to see this merged quickly.  Merging soon has a few benefits:

1.  It frees @pitrou from having to constantly merge in updates from master
2.  It gets more exposure to possible regressions
3.  It lets us start working on other comms like Queues or our own TCP stream (if we can verify that that is valuable)

mrocklin:
@pitrou how do you feel about this PR?  What remains to be done?

pitrou:
Nothing important really. There may be bits and pieces worth further cleaning up, but that can be tackled in further PRs.

mrocklin:
Well, this is in.  Lets see what happens :)

mrocklin:
Thank you @pitrou for taking this on.

minrk:
I will investigate if there are any zmq tuneups I can provide, if I get a chance.

PR #821, opened 2017-01-18
Fix retire_workers(close=True)

mrocklin:
This was previously untested and faulty

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -445,16 +445,23 @@ class Scheduler(Server):
         """
         with log_errors():
             original = worker
-            self.remove_worker(address=original)
-            with ignoring(KeyError):
+            try:
                 nanny_port = self.worker_info[worker]['services']['nanny']
-                ip, port = worker.split(':')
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -445,16 +445,23 @@ def close_worker(self, stream=None, worker=None):
...
+                address = '%s:%s' % (ip, nanny_port)
+            else:
+                address = worker
 
-            with rpc(worker) as r:
+            with rpc(addr=address) as r:
                 yield r.terminate(report=False)
 
-            self.remove_worker(address=original)
+            self.remove_worker(address=worker)
```
pitrou: May I ask why remove_worker() is called a second time here?

mrocklin: remove_worker is idempotent.  There is a slight chance that we have received an out-of-order heartbeat (which is just an add-worker message) during the yield.  I didn't actually experience this, nor do I have confidence that it would be an issue, but it is cheap and safe to call remove_worker the second time.

PR #823, opened 2017-01-20
Set rpc status while deleting

mrocklin:
We're getting the following error, showing that rpc.active decreases
below zero.  This change is just a blind guess as to the cause.

    File "C:\projects\distributed\distributed\utils_test.py", line 326, in check_active_rpc
      assert rpc.active == rpc_active
    AssertionError: assert -1 == 0

@pitrou, any thoughts?

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -563,6 +563,7 @@ class rpc(object):
     def __del__(self):
         if self.status != 'closed':
             rpc.active -= 1
+            self.status = 'closed'
             n_open = sum(not stream.closed() for stream in self.streams)
             if n_open:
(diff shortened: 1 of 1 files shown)


pitrou:
That would certainly be a possibility, though unlikely. Do you know how to reproduce the issue?
In any case, the change here doesn't harm...

pitrou:
I said it's unlikely, but it's actually plausible. If there's a reference cycle, and the rpc's `__del__` is called first, and another object's `__del__` calls the rpc's `close_rpc` method, then that can happen.

pitrou:
Another thing to look at in the refactor :-)

PR #826, opened 2017-01-23
WIP: worker lifetime initial commit

jcmgray:
This adds a `--lifetime` command line option to `dask-worker`, such that the worker retires itself after this time, replicating its data to other workers. This adds a degree of resilience to e.g. long running computations on batch-systems where the workers are submitted as jobs with limited run time.

For example:

```bash
$ dask-worker ... --lifetime="hours=23 mins=55"
```

Some comments/questions:
- the name '--lifetime', maybe '--retire-after' would be more explicit? Also the specifier is currently chosen to match the --resources argument, rather than e.g. 23h55m.
- Due to the timing nature the tests are a little slow but possibly fragile if sped up.
- Docs: should this be in the resilience section or somewhere else?

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -76,9 +76,13 @@ def handle_signal(sig, frame):
               help="Internal use only")
 @click.option('--resources', type=str, default='',
               help='Resources for task constraints like "GPU=2 MEM=10e9"')
+@click.option('--lifetime', type=str, default='',
+              help='Lifetime of worker before it should retire itself, '
+                   'specified e.g. "d=2.5 h=12 m=45 s=59" for days, hours, '
+                   'minutes and seconds respectively.')
 def main(scheduler, host, worker_port, http_port, nanny_port, nthreads, nprocs,
-         nanny, name, memory_limit, pid_file, temp_filename, reconnect,
-         resources, bokeh, bokeh_port, local_directory):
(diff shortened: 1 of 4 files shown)


jcmgray:
Looks already like the tests need to be much more robust ..

distributed/cli/dask_worker.py
```diff
@@ -167,9 +186,27 @@ def f():
...
+            with rpc(ip=nannies[0].scheduler.ip,
+                     port=nannies[0].scheduler.port) as scheduler:
+                workers = ([n.worker_address for n in nannies
+                            if n.process and n.worker_port] if nanny else
+                           [n.address for n in nannies])
+                scheduler.retire_workers(workers=workers, remove=False)
+            retired[0] = True
+            logger.info("Worker retired")
+
+        loop.add_callback(retire)
```
mrocklin: It would be nice to see this logic (or really any logic) within the Worker class rather than within the command line interface.  It is also *much* easier to test the worker class directly.  http://distributed.readthedocs.io/en/latest/develop.html#writing-tests

jcmgray: Yes that makes sense regarding tests - I will have a go at moving it in when I get a chance. The only complication that comes to mind is that to survive restarts etc. the time might have to be converted to an absolute time.

mrocklin:
Closing as stale.  Please reopen if work picks up again.

PR #827, opened 2017-01-23
Added bindall option to dask-ssh.

### LINKED ISSUE #824: Problem with dask-ssh in combination with hostnames
When starting a distributed cluster with dask-ssh, the hostnames supplied via commandline or hostfile are passed to the workers with the `--host` parameter. This might be useful for the bokeh interface on the scheduler, but in our case the worker binds to the local interface only (the hostname is listed in the `/etc/hosts` with `127.0.0.1`), so a connection from the scheduler to the worker is not possible (maybe I am wrong and this is somehow not needed, but I receive errors about failed connects).

A possible solution is to provide IP-Adresses, because with that the worker binds to the correct interface. For me this is less optimal, as I cannot easily see where I start my workers.

My quick and dirty fix was to remove the `host` option from the worker start call.

So I would like to suggest adding an option to `dask-ssh` to prevent using the `host` option when spawning a worker (as this works perfectly fine in internal networks where some open ports are not such a big problem).

If you are interested in this `dask-ssh` option, I would implement this in a proper way and create a PR.

malteger:
With the default settings dask-workers spawned by dask-ssh where
listening only on the adress which resolved to the given hostname.
This can now be changed.
Closes #824.

### DIFF
--- a/distributed/cli/dask_ssh.py
+++ b/distributed/cli/dask_ssh.py
@@ -29,12 +29,13 @@ from distributed.cli.utils import check_python_3
               help="Port to use for SSH connections.")
 @click.option('--ssh-private-key', default=None, type=str,
               help="Private key file to use for SSH connections.")
-
+@click.option('--nohost', is_flag=True,
+              help="Do not pass the hostname to the worker.")
 @click.option('--log-directory', default=None, type=click.Path(exists=True),
               help="Directory to use on all cluster nodes for the output of dask-scheduler and dask-worker commands.")
 @click.pass_context
 def main(ctx, scheduler, scheduler_port, hostnames, hostfile, nthreads, nprocs,
-          ssh_username, ssh_port, ssh_private_key, log_directory):
+          ssh_username, ssh_port, ssh_private_key, nohost, log_directory):
     try:
         hostnames = list(hostnames)
         if hostfile:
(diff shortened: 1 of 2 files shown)


distributed/cli/dask_ssh.py
```diff
@@ -29,12 +29,13 @@
               help="Port to use for SSH connections.")
 @click.option('--ssh-private-key', default=None, type=str,
               help="Private key file to use for SSH connections.")
-
+@click.option('--bindall', is_flag=True,
+              help="Let the worker bind to all available interfaces.")
```
mrocklin: Ah, please note that the use of `--host` is not a bind address, but rather the name by which the worker refers to itself.  

When a worker registers with the scheduler it provides an address that other workers should use to contact it in the future.  That is the meaning of the value provided by `--host`.

malteger: If I understand the code correctly, this is not true. If no host option is given, an interface which can reach the scheduler is automatically choosen. But if you provide a host option, `socket.gethostbyname` is used ([see here](https://github.com/dask/distributed/blob/master/distributed/cli/dask_worker.py#L142)).

This causes the real problem, if you specify a host like `foo` and your hosts file says that `127.0.0.1` is also known as `foo`, the worker will use `127.0.0.1` as the interface address.

After thinking about it a little bit, it might be better to rename the option to something like `--nohost` or `--bindauto`.

malteger: But the name thing is also true, so without `--host` you get only IP addresses in bokeh.

mrocklin: `--nohost` seems like a nice name to me.

mrocklin:
Looks good to me.  Thanks for doing this @malteger !

PR #828, opened 2017-01-23
Atomically create the destination config.yaml

pitrou:


pitrou:
@jcrist

distributed/config.py
```diff
@@ -24,8 +24,13 @@ def ensure_config_file(destination=dask_config_path):
                     os.mkdir(os.path.dirname(destination))
                 except FileExistsError:
                     pass
-            shutil.copy(default_path, destination)
-
+            # Atomically create destination
+            tmp = '%s.tmp.%d' % (destination, os.getpid())
+            shutil.copy(default_path, tmp)
```
jcrist: Might be nice to add a comment explaining why we do this instead of just copying directly.

distributed/config.py
```diff
@@ -24,8 +24,13 @@ def ensure_config_file(destination=dask_config_path):
...
+            try:
+                os.rename(tmp, destination)
+            except OSError:
+                os.unlink(tmp)
```
jcrist: Slight request to use `os.remove` instead. I was unfamiliar with `unlink`, and per the docs they are equivalent (https://docs.python.org/3.6/library/os.html#os.unlink).

pitrou: Ha, fair enough.

distributed/config.py
pitrou: Ok, I added an explanation.

pitrou:
I'm merging now. Please reopen if this doesn't fix the sporadic failures in Dask.

PR #829, opened 2017-01-24
As completed

mrocklin:
This extends the concurrent.futures `as_completed` iterator to allow us to add new futures in the middle of the computation.  This is useful for some upcoming machine learning work.

cc @pitrou do you have some time to give this a quick review?

### DIFF
--- a/distributed/tests/test_as_completed.py
+++ b/distributed/tests/test_as_completed.py
@@ -52,14 +52,17 @@ def test_as_completed_add(loop):
     with cluster() as (s, [a, b]):
         with Client(('127.0.0.1', s['port']), loop=loop) as c:
             total = 0
-            futures = c.map(inc, range(20))
+            expected = sum(map(inc, range(10)))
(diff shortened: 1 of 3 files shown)


distributed/tests/test_as_completed.py
```diff
@@ -0,0 +1,80 @@
...
+        with Client(('127.0.0.1', s['port']), loop=loop) as c:
+            total = 0
+            futures = c.map(inc, range(20))
+            ac = AsCompleted(futures)
+            for future in ac:
+                total += future.result()
+                if random.random() < 0.5:
+                    future = c.submit(add, future, 10)
+                    ac.add(future)
+            assert total > sum(map(inc, range(10)))
```
pitrou: The test could be a bit more precise and assertive (perhaps by keeping a record of all expected results?).

mrocklin: Fixed (I think)

PR #832, opened 2017-01-26
Use numpy.nbytes rather than memoryview.nbytes

mrocklin:
This avoids a python 2 issue

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -50,7 +50,7 @@ def serialize_numpy_ndarray(x):
 
     data = x.view('u1').data
 
-    if blosc and data.nbytes > 1e5:
+    if blosc and x.nbytes > 1e5:
         frames = frame_split_size([data])
         if sys.version_info.major == 2:
             frames = [ensure_bytes(frame) for frame in frames]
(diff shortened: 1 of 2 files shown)


pitrou:
Looks fine to me.

mrocklin:
Running into other issues here that you might be able to help with actually @pitrou.  I like memoryviews, but I'm having trouble creating a very simple memory view with a linear shape and itemsize=1.

I can get to itemsize=1 with `my_numpy_array.view('u1').data`, but I'm not sure how to flatten out a memoryview's shape in a way that doesn't involve a copy on the numpy side.

mrocklin:
I want this:

```python
x = np.ones((10, 10, 10, 10, 10, 10, 10))
x.flatten().view('u1').data
```

But flatten involves a copy.  I have yet to find find API either in numpy or the Python memoryview that lets me do this without a copy.

mrocklin:
Ah, ravel

pitrou:
The cast() method should work (but perhaps not on Python 2):
```
>>> x = np.ones((10, 10, 10, 10, 10, 10, 10))
>>> m = memoryview(x)
>>> m.shape, m.nbytes, m.itemsize
((10, 10, 10, 10, 10, 10, 10), 80000000, 8)
>>> m = m.cast('B')
>>> m.shape, m.nbytes, m.itemsize
((80000000,), 80000000, 1)
```

pitrou:
`array.ravel()` will make a copy if necessary (if the array isn't contiguous), while `memoryview.cast()` will refuse to work on non-contiguous arrays.

mrocklin:
I think we want ravel then.  We're willing to copy.  https://github.com/dask/distributed/pull/833

PR #833, opened 2017-01-26
Ravel bytes before creating memoryview

mrocklin:


distributed/protocol/tests/test_numpy.py
```diff
@@ -49,6 +49,7 @@ def test_serialize():
          np.array([(1, 'abc')], dtype=[('x', 'i4'), ('s', object)]),
          np.zeros(5000, dtype=[('x%d'%i,'<f8') for i in range(4)]),
          np.zeros(5000, dtype='S32'),
+         np.zeros((1, 1000, 1000)),
```
pitrou: I think it would be good to add a non-contiguous array to that list.
(for example `np.arange(12)[::2]`).

PR #835, opened 2017-01-27
Test that forked workers have different random seeds

mrocklin:
This is currently just a failing test.

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -314,8 +314,20 @@ def run_worker_fork(q, scheduler_addr, ncores, nanny_port,
     """
     Create a worker by forking.
     """
+    import os
+    import random
     from distributed import Worker  # pragma: no cover
+    from distributed.metrics import time
     from tornado.ioloop import IOLoop  # pragma: no cover
+    random.seed((time(), os.getpid()))
+    try:
+        import numpy as np
+    except ImportError:
+        pass
+    else:
+        np.random.seed(hash((time(), os.getpid(), random.randint(0, 2**30)))
+                       % 2**32)
+
     IOLoop.clear_instance()  # pragma: no cover
     loop = IOLoop()  # pragma: no cover
     loop.make_current()  # pragma: no cover
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -5,6 +5,7 @@ import logging
 from numbers import Integral, Number
 from operator import add
 import os
+import random
 import re
 import shutil
 import sys
@@ -12,13 +13,14 @@ import traceback
 
 from dask import delayed
 import pytest
+import numpy as np
 from toolz import pluck, sliding_window
 import tornado
 from tornado import gen
 from tornado.ioloop import TimeoutError
 
 from distributed.core import rpc, connect
-from distributed.client import _wait
+from distributed.client import _wait, Client
 from distributed.scheduler import Scheduler
 from distributed.metrics import time
 from distributed.protocol import to_serialize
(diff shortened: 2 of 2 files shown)


distributed/tests/test_worker.py
```diff
@@ -651,3 +652,12 @@ def test_hold_onto_dependents(c, s, a, b):
     yield gen.sleep(0.1)
 
     assert x.key in b.data
+
+
+def test_random_seed(loop):
+    with cluster() as (s, [a, b]):
+        with Client(('127.0.0.1', s['port']), loop=loop) as c:
+            x = c.submit(random.randint, 0, 2**31, pure=False)
+            y = c.submit(random.randint, 0, 2**31, pure=False)
```
pitrou: Should probably be testing for `np.random` as well (since it uses a separate random state).

mrocklin:
Pushed a solution.  Interestingly sending random.randint or np.random.randint will never produce different results because these are methods on existing RandomState objects.  The random state data gets sent along with the methodname when pickled.

pitrou:
What's the status on this? Do you want me to take over?

> Interestingly sending random.randint or np.random.randint will never produce different results because these are methods on existing RandomState objects. The random state data gets sent along with the methodname when pickled.

Interesting indeed...

mrocklin:
Any feedback on the solution here?  Anything we should do that we aren't doing?

distributed/nanny.py
```diff
@@ -314,8 +314,20 @@ def run_worker_fork(q, scheduler_addr, ncores, nanny_port,
     """
     Create a worker by forking.
     """
+    import os
+    import random
     from distributed import Worker  # pragma: no cover
+    from distributed.metrics import time
     from tornado.ioloop import IOLoop  # pragma: no cover
+    random.seed((time(), os.getpid()))
```
pitrou: Using `random.seed()` without any argument makes a best effort to seed with enough high-quality entropy.

distributed/nanny.py
```diff
@@ -314,8 +314,20 @@ def run_worker_fork(q, scheduler_addr, ncores, nanny_port,
...
+    try:
+        import numpy as np
+    except ImportError:
+        pass
+    else:
+        np.random.seed(hash((time(), os.getpid(), random.randint(0, 2**30)))
```
pitrou: Ditto here, just using `np.random.seed()` fills the random state with system-provided entropy.

mrocklin:
Feel free to take over

distributed/nanny.py
```diff
@@ -314,8 +314,20 @@ def run_worker_fork(q, scheduler_addr, ncores, nanny_port,
...
     Create a worker by forking.
     """
+    import os
+    import random
     from distributed import Worker  # pragma: no cover
+    from distributed.metrics import time
     from tornado.ioloop import IOLoop  # pragma: no cover
+    random.seed((time(), os.getpid()))
+    try:
+        import numpy as np
```
pitrou: I'm not sure, but instead of trying to import, we could just check for the module in `sys.modules`. That way, you don't load Numpy if it's not already in use.

distributed/nanny.py
pitrou: By the way, AFAICT, multiprocessing already does this for us. Only `np.random` requires our intervention.

pitrou:
One more comment: how about we centralize this "after fork" code somewhere?

mrocklin:
> One more comment: how about we centralize this "after fork" code somewhere?

Fine by me.  Where else it it used?

pitrou:
Right now, nowhere else, but is dask's process pool susceptible to the same issue?

pitrou:
See https://github.com/dask/distributed/pull/953 for an alternative take on this.

pitrou:
Closing as  #953 got merged.

PR #837, opened 2017-01-30
Add IP keyword to LocalCluster to enable remote access

mrocklin:
Previously the scheduler in the localcluster could only receive connections
from localhost.  This adds a keyword to enable remote access.

This is important because LocalCluster is often a base for other cluster
implementations.

cc @pitrou

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -38,6 +38,8 @@ class LocalCluster(object):
     silence_logs: logging level
         Level of logs to print out to stdout.  ``logging.CRITICAL`` by default.
         Use a falsey value like False or None for no change.
+    ip: string
+        IP range on which the scheduler will listen, defaults to only localhost
     kwargs: dict
         Extra worker arguments, will be passed to the Worker constructor.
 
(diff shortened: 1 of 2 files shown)


distributed/deploy/local.py
```diff
@@ -38,6 +38,8 @@ class LocalCluster(object):
     silence_logs: logging level
         Level of logs to print out to stdout.  ``logging.CRITICAL`` by default.
         Use a falsey value like False or None for no change.
+    ip: string
+        IP range on which the scheduler will listen, defaults to only localhost
```
pitrou: It's an IP address, not an IP range.

distributed/deploy/tests/test_local.py
```diff
@@ -223,3 +224,21 @@ def test_silent_startup(capsys, loop):
     assert not out
     for line in err.split('\n'):
         assert 'worker' not in line
+
+
+def test_only_local_access(loop):
+    from distributed.http import HTTPScheduler
+    import requests
```
pitrou: It seems requests isn't actually used?

distributed/deploy/tests/test_local.py
```diff
@@ -223,3 +224,21 @@ def test_silent_startup(capsys, loop):
...
+    with LocalCluster(scheduler_port=0, silence_logs=False,
+            services={('http', 3485): HTTPScheduler}, diagnostics_port=None,
```
pitrou: Perhaps you intended to test this port as well below.

Issue #838, opened 2017-01-30
Gathering results of queued futures


mackaiver:
Hello,

similar to the example provided in the docs I'm trying to achieve some form of data streaming using Queues. I have a producer running in a thread putting things into a queue. After some computation I want to gather the results using a call to `client.gather(queue_of_futures)`. This seems to work more or less until I start producing data faster than it is being consumed. Nothing ever gets written to the results. I guess this is intended behaviour and I need to wait until all futures on all workers  have finished computation. In that case I'd need to know when to start producing data again. To my understanding this would mean pausing the data stream which is unfortunate. Is there a way to circumvent these limitations? 

Here is a small working example:

```
from queue import Queue
import time
from distributed import Client
from threading import Thread

e = Client('127.0.0.1:8786')


def load_data(q):
    i = 0
    while True:
        q.put(i)
        time.sleep(0.5)
        i += 1


def inc(x):
    time.sleep(0.2)
    return x + 1


def double(x):
    time.sleep(0.1)
    return 2 * x


input_q = Queue(maxsize=1000)
remote_q = e.scatter(input_q)
inc_q = e.map(inc, remote_q)
double_q = e.map(double, inc_q)
result_q = e.gather(double_q)


load_thread = Thread(target=load_data, args=(input_q,), daemon=True)
load_thread.start()

while True:
    #this waits forever in case the producer is too fast
    print('latest result: {}'.format(result_q.get()))

```

mrocklin:
Interesting.  So currently it looks like the workers respond in a first-in-last-out fashion.  This tends to be ideal for interleaving computations as they come in from the scheduler.  However you've clearly identified a fail-case where, for dynamic workflows, this causes undesired behavior.  

I'll look at how we handle priorities and see if there is a way to maintain the old interleaving behavior while still maintaining first-in-first-out based on time of submission.  Thanks for raising the issue.

For others looking at this issue, the example provided can reproduce *if* you significantly increase the duration of the computations.

PR #840, opened 2017-01-30
First in first out on client submission

mrocklin:
Given more tasks than available worker threads, we must choose between them.

There is a general tension between first-in-first-out, which prioritizes tasks
submitted earlier to tasks submitted later, and first-in-last-out, which
prioritizes tasks that have just become ready.  Generally FIFO is better for
streaming or linear computations while LIFO is better for complex computations
and reducing memory footprint.

Now the scheduler's policy is to be FILO within any atomic submission of tasks
(like compute or persist) but to be FIFO between calls.

Fixes https://github.com/dask/distributed/issues/838

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1425,6 +1425,11 @@ class Client(object):
         for k, v in dsk3.items():
             dependencies[k] |= get_dependencies(dsk3, task=v)
 
+        if priority is None:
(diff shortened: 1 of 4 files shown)


mrocklin:
cc @mackaiver

mackaiver:
This works great. Thanks for the quick response 👍

PR #855, opened 2017-02-03
Task stream no labels

mrocklin:
cc @martindurant @brendancol


mrocklin:
image

martindurant:
The number of ticks probably depends on the number of cores.

mrocklin:
It does, we used to have two ticks per core.  I wasn't sure why.

Issue #860, opened 2017-02-07
h5py lock-up on array load


martindurant:
When rewriting the dask-tutorial to use distributed, find that the dask-array example with a hdf5 file freezes, presumably in lock contention. I have not isolated the problem, but it does not happen with the default scheduler. I unravelled da.from_array a bit to try to make it clearer

This first version passes, where `dset` is global
```python
fn = 'random.hdf5' # generated by dask-tutorial/prep.py, random_array
f = h5py.File(fn, mode='r')
dset = f['/x']

def get_part(off):
    return dset[off:off+1000000]

@gen_cluster(client=True)
def test_hdf(c, s, a, b):
    parts = [delayed(get_part)(r) for r in range(0, 10000000, 1000000)]
    bits = [da.from_delayed(part, shape=(1000000, ), dtype='f4')
            for part in parts]
    x = da.concatenate(bits)
    f = c.compute(x.sum())
    out = yield c._gather(f)
```

but if you pass `dset`, it fails (dset itself pickles OK)
```python
def get_part(off, dset):
    return dset[off:off+1000000]

@gen_cluster(client=True)
def test_hdf_direct(c, s, a, b):
    f = h5py.File(fn, mode='r')
    dset = f['/x']

    parts = [delayed(get_part)(r, dset) for r in range(0, 10000000, 1000000)]
    bits = [da.from_delayed(part, shape=(1000000, ), dtype='f4')
            for part in parts]
    x = da.concatenate(bits)
    f = c.compute(x.sum())
    out = yield c._gather(f)
```

seibert:
Not sure whether you are running multiple threads in the worker process, but the HDF5 library has some global state that isn't thread-safe (even when accessing files read-only) unless it has been built with `--enable-threadsafe`.  Last I checked, the Anaconda build of HDF5 doesn't turn on this flag as it also can reduce performance (IIRC).

mrocklin:
from_array should introduce a lock by default

martindurant:
Even calling from_array with lock=True still showed the error.

martindurant:
Interestingly, I am not seeing the problem with `dataframe.read_hdf`

mrocklin:
Looking at this example more closely now.  So the problem here is that h5py objects aren't picklable (nor are they ever likely to be).  Fortunately, we have defined our own serialization for these objects, but that only works if the objects are top-level in the graph

    {'x': dset}  # this should be ok
    {'x': (myfunc, dset)}  # this would fail

[...]

PR #861, opened 2017-02-07
Add failing test for h5py serialization

mrocklin:
See https://github.com/dask/distributed/issues/860


mrocklin:
Oddly it looks like we're getting lists from BatchedSend in `handle_comm`.  I'm not sure why this would occur.

mrocklin:
@martindurant this may now be resolved.  Was a tricky case to find.  Thank you for reporting the issue.

martindurant:
Thank you. Seems pretty subtle.

mrocklin:
L[-0] .  Not the first time that has bitten me.

PR #862, opened 2017-02-07
POC: Cluster monitor

martindurant:
Set of methods on LocalCluster, and bokeh dashboard to execute them. Bokeh dashboard will be available after `cluster.start_cluster_monitor(port)`.

Can create, restart and stop workers, which can be nannied or not (restarting a raw worker means killing it and starting a new one). Can also run restart and stop on the scheduler entry.

The plan is to have the dashboard and log buttons take you to the relevant information for the selected item (or simply remove these buttons...). Extra information could be given about the workers, such as is done in the normal dashboards.

All the cluster implementations then could gain these few methods, if they live in the same process as the scheduler. Container-based clusters will allow you to create and stop new containers, rather than individual processes, and maybe dynamic policies could go on this monitor too.


martindurant:
Thought on how to extend this for other cluster deployments in http://distributed.readthedocs.io/en/latest/setup.html

The general pattern for some of the deployments (kubernetes, marathon, ssh) is to run a process locally and connect to remote services/resource providers, to set up containers with dask-workers. That means that the information on logins and how to make new containers (or ssh IP/keys) remains local, not where the scheduler is.

A better model (knit, drmaa), assumes we are running on the scheduler machine, which has the right configuration to talk to the resource manager. For the monitor, we want the scheduler in the same process as the controller object, like LocalCluster. In fact, this could be a single object like in the [adaptive example](http://distributed.readthedocs.io/en/latest/adaptive.html#marathon-an-example). How to set up this scheduler machine can be another problem later down the road.

jakirkham:
This looks really interesting. Wonder if it could be combined with the `Cluster` ABC?

mrocklin:

PR #863, opened 2017-02-07
This adds a better approximation to the memory consumption of scipy sparse matrices

demianw:
By adding new dispatches to the sizeof function we provide better support for persisting sparse matrix objects to a distributed cluster.

### DIFF
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -68,7 +68,7 @@ def register_pandas():
 def register_spmatrix():
     from scipy import sparse
     @sizeof.register(sparse.dok_matrix)
-    def sizeof_spmatrix(s):
+    def sizeof_spmatrix_dok(s):
         return s.__sizeof__()
 
     @sizeof.register(sparse.spmatrix)
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -62,3 +62,18 @@ def register_pandas():
         p = int(i.memory_usage())
         obj = len(i) * 100 if i.dtype == object else 0
         return int(p + obj) + 1000
+
+
+@sizeof.register_lazy("scipy")
+def register_spmatrix():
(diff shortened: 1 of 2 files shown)


distributed/sizeof.py
```diff
@@ -62,3 +62,18 @@ def sizeof_pandas_index(i):
...
+    from scipy import sparse
+    @sizeof.register(sparse.dok_matrix)
+    def sizeof_spmatrix_dok(s):
+        return s.__sizeof__()
+
+    @sizeof.register(sparse.spmatrix)
+    def sizeof_spmatrix(s):
+        return sum(
+            sizeof(v) for v in s.__dict__.values()
+        )
```
mrocklin: How slow is this in the common case?  Are these values all of a certain dtype?  If so then are there other ways to compute this more quickly?  For example by returning `len(s.__dict__) * len(first(s.values()))`

Also in cases like this it might be nicer to put this code on one line if it fits.

demianw: It's very fast, just a couple milliseconds,, as the number of elements in the dictionary around 5 or 6 in each `spmatrix` and 2 or 3 are `ndarray`s depending on the sparse matrix type.
If this is a problem, an alternative would be to overestimate it as 
`len(s.__dict__) * int(s.data.nbytes)`
as `s.data` is a numpy ndarray in all cases except in  the `dok_matrix` type

mrocklin: Ah, I didn't realize that the elements of this dict were numpy arrays.  What you have here seems great.

mrocklin:
This looks pretty good to me.  Thank you for taking this on.

demianw:
Thanks for this libraries, they're great! 

So, is this PR merge-ready or should I change something?

mrocklin:
Merged.  Thanks @demianw !

PR #866, opened 2017-02-10
Add SSL Support 

mariusvniekerk:
@mrocklin This is a pretty early rough idea.  Can you tell me what else I missed?

This is a pretty simple passthrough similar to #537


mrocklin:
cc @pitrou @hussainsultan @ebolyen , all of whom are much more qualified than I am to answer this quesiton :)

ebolyen:
I think this PR is missing a way to authenticate the `Client` object to the `Scheduler` as it doesn't pass `**kwargs` around like the other entities. (I'm assuming mutual authentication is the goal here.)

distributed/comm/core.py
```diff
@@ -238,7 +238,7 @@ def resolve_address(addr):
 
 
 @gen.coroutine
-def connect(addr, timeout=3, deserialize=True):
+def connect(addr, timeout=3, deserialize=True, ssl_options=None):
```
pitrou: I think this is the wrong approach. SSL will not be applicable to all transports. Instead, I think SSL (I'd rather call it TLS btw) should be exposed as a distinct transport from TCP.

mariusvniekerk: On the tornado side though its basically just this additional flat to the TCPServer/Client (which is how jupyter notebook, for example, does it https://github.com/jupyter/notebook/blob/900e77d1b70ec7663f45e566404f2ede964b99c1/notebook/notebookapp.py#L1107 )

Alternatively I can make it change the urls that it spits out to tls:// instead of tcp:// if the cert/key things are specified.

pitrou: Yes, but the `distributed.comm` layer abstracts away the transport mechanism, so you cannot assume that the notion of SSL options has a meaning for transports other than TCP.

mariusvniekerk: We do have to add a transport-kind specific option of passing additional parameters around.  
Not sure where the best place for that is other than a **kwargs.

New to this codebase.

pitrou: That's a good point. My initial idea was to specify the TLS configuration in the config file, which would have avoided passing extra arguments. Is it more convenient to specify the TLS configuration on the command line instead?

mariusvniekerk: I think it is more convenient for users (although having it behave like jupyter where both modes work would be even nicer).  

The main reason there is that you may not want it to be global for all dask runs, since there is presumably some overhead in the encryption.  Also not sure if there is a single config file or if that can be specified per job.

As a point of comparison, Apache Spark allows passing these as command line arguments to  `spark-submit`

mrocklin:
What is the status on this?  cc @mariusvniekerk @pitrou

distributed/comm/core.py
pitrou: As a matter of fact, we could still have command-line options but have them override parameters in the global `config` dictionary. But that can be implemented later as well...

pitrou:
Another high-level question: this seems to only address client-to-scheduler connections? Am I missing something? Is the PR incomplete perhaps?

mrocklin:
I'm chatting with @mariusvniekerk about this now.  We're thinking that it might make sense for all Servers to hold onto a `connection_kwargs` dict that they pass into all connect and rpc calls as well as their initial `listen` call.  

After this is in place it would be relatively easy to add `ssl_options=` into this dict within the constructor or CLI utilities.  Hopefully this would help us to avoid having a bunch of smaller options (like `scheduler.ssl=True`) scattered throughout the codebase.  

However, this also means that there won't be an identifiable difference between the addresses of servers that use TCP connections and the addresses of servers that require SSL enabled TCP connections.  We require that SSL is "all or nothing".  

@pitrou, any thoughts on this?

pitrou:
The `connection_kwargs` idea makes sense. That said, we can combine it with different addresses for TLS endpoints, such that logs and the like clearly can tell whether a connection is "secure" or not.

mrocklin:
Adding a different address type seems sensible to me.  I think it would be nice to see if we can do SSL strictly on the current TCP system with just options.  This would be, I think, a nice demonstration that we're able to pass through options to the underlying transport.  

I think that after this works it would be interesting to add an explicit comm type for TLS endpoints to see how cheaply we can add a new comm type.

(though in this case, I'm out of my element, and so may be missing things.)

mariusvniekerk:
I think we want to wait for #896 to drop properly?

pitrou:
I'm not sure I understand the question. What do you mean by "drop properly"?
But, yes, it would be better to wait for that PR (or rebase on it, but given we squash when merging PRs, you'll be getting conflicts after the PR gets merged).
We might also want to wait for #919, but that's less important (i.e. less conflict-prone).

mariusvniekerk:
@pitrou I mean be merged in.

mariusvniekerk:
So relevant tests seem to be passing.  Seems there is some weirdness grabbing some conda packages.

mrocklin:
Overall this seems great to me.  Especially from a maintenance burden perspective this appears to be free :)

I'm curious about the handling of TCP vs TLS.  What are your thoughts on the protocol switching here?

mrocklin:
How does one test SSL like this?  Both when setting things up manually in python and when using the CLI?

The `gen_cluster` test decorator takes in keywords for the scheduler and workers that might be helpful here.  Presumably it would also be useful to do at least a single test with the CLI directly using `dask-scheduler` and `dask-worker`.

Have you tried this out on your own?  Do things run well for you anecdotally?

mrocklin:
Also, it's going to make me pretty happy if you can add SSL support to Dask in 132 changed lines of code :)

mrocklin:
cc @hussainsultan now would be a good time for you to check in on this, if only briefly, just to make sure that this would satisfy your needs.

mariusvniekerk:
i still need to write some tests for the ssl pieces.  Figure i can use gen_cluster to make it work?

mrocklin:
This page may be of some use:
http://distributed.readthedocs.io/en/latest/develop.html#writing-tests

On Fri, Mar 3, 2017 at 8:53 PM, Marius van Niekerk <notifications@github.com

distributed/tests/test_ssl.py
```diff
@@ -0,0 +1,69 @@
...
+    @gen_cluster(
+        scheduler_kwargs={'connection_kwargs': ssl_kwargs},
+        worker_kwargs={'connection_kwargs': ssl_kwargs},
+        client=True)
+    def f(c, s, a, b):
+
+        assert isinstance(c, Client)
+        assert isinstance(c, Scheduler)
+        assert isinstance(a, Worker)
+        assert isinstance(b, Worker)
```
mrocklin: You can drop these.  This was just for the example so that new devs could see what they were.

distributed/tests/test_ssl.py
```diff
@@ -0,0 +1,69 @@
...
+
+        future = c.submit(inc, 1)
+        assert future.key in c.futures
+
+        # result = future.result()  # This synchronous API call would block
+        result = yield future._result()
+        assert result == 2
+
+        assert future.key in s.tasks
+        assert future.key in a.data or future.key in b.data
```
mrocklin: What would happen if we added a new worker without SSL certificates?

distributed/tests/test_ssl.py
```diff
@@ -0,0 +1,69 @@
...
+    p.communicate()
+    assert p.returncode == 0
+
+    return {'ssl_options': {
+        'certfile': certfile,
+        'keyfile': keyfile,
+    }}
+
+
+def test_ssl(ssl_kwargs, loop):
```
mrocklin: Hrm, I had never thought of doing this to combine pytest fixtures with the current decorator approach.  Neat.

distributed/tests/test_ssl.py
mariusvniekerk: well right now this doesn't quite work yet.  The error messages suggest that some tcp communication is trying to happen somewhere.  My tornado debug powers are sadly less strong than desired.

mariusvniekerk: Also the way things work right now is pieces get very very unhappy if you mix ssl and non-ssl

mrocklin: For debugging is there anything I can help with?

Adding `pdb-on-err: True` to your `~/.dask/config.yaml` might help in some cases.

mariusvniekerk: Seems like the workers get kicked out of the scheduler for some reason.

mariusvniekerk: Was missing a connection_kwargs in a connect call from the scheduler

mariusvniekerk:
Okay @mrocklin i'm not quite sure why the single test that is failing right now is failing.  Seems to make little sense to me.

mrocklin:
There are two failures, one in broken_worker_during_... which is my fault and a current intermittent failure that I thought we had resolved.  The other is something odd happening in ZeroMQ tests.  My guess is that something you're doing isn't passing around addresses correctly. @pitrou would be able to comment better on this topic, but I suspect he's away on vacation for a few days.

mariusvniekerk:
Sweet.  And we are green

distributed/tests/test_ssl.py
mrocklin: It would also be nice to test the CLI to ensure that the whole system functions well.

mrocklin: See `distributed/cli/tests/test_dask_scheduler.py`

mrocklin:
In general I'm pretty happy with all of this.  Do you mind waiting a few days before merging?  I'd like to give @pitrou a chance to look this over and I think he's on vacation until Monday.

mariusvniekerk:
Fine with me @mrocklin.  Unsure whether we want to call this out in the docs yet?

mrocklin:
A small docpage entry would be nice, if only so that we could hand it around to a few people and get them to try it out and offer feedback.  I would focus on the minimum text necessary to use things rather than design choices in how it works.

mrocklin:
@pitrou if you have some time this PR could use your review

pitrou:
Yes, I'm doing that.

docs/source/ssl.rst
```diff
@@ -0,0 +1,23 @@
+.. _ssl:
+
+SSL
```
pitrou: I'd prefer to call this "TLS" everywhere. "SSL" is admittedly more well-known, but it's been a legacy name for 15 years...

distributed/cli/dask_remote.py
```diff
@@ -9,8 +9,12 @@
 @click.option('--host', type=str, default=None,
               help="IP or hostname of this server")
 @click.option('--port', type=int, default=8788, help="Remote Client Port")
-def main(host, port):
-    _remote(host, port)
+@click.option('--certfile', default=None,
```
pitrou: These options (and the corresponding logic) is duplicated several times, maybe it would be good to factor them out.

distributed/cli/dask_worker.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import print_function, division, absolute_import
 
 import atexit
+import ssl
```
pitrou: This import seems unused?

distributed/cli/utils.py
```diff
@@ -69,3 +71,12 @@ def uri_from_host_port(host_arg, port_arg, default_port):
     addr = unparse_address(scheme, loc)
 
     return addr
+
+
+def create_ssl_context(certfile, keyfile):
+    if certfile and keyfile:
```
pitrou: This doesn't seem right. Theoretically, you can create a context with only a certfile and no keyfile (for example if the certfile also contains the private key).

If that's not allowed, then passing only one of two options should give an error instead of silently falling back on an insecure transport.

distributed/cli/utils.py
```diff
@@ -69,3 +71,12 @@ def uri_from_host_port(host_arg, port_arg, default_port):
...
+        ssl_ctx = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
```
pitrou: The theoretical difficulty here is that both client and server connections will be created using the same context, even though you specified `SERVER_AUTH`. Nowadays this doesn't matter so much, but it may matter in subsequent revisions of the standard `ssl` module.

mariusvniekerk: Chose this since it seemed like pretty much all dask things were both clients and servers and it seemed to work at least.  Given that the `ssl.Purpose` was reintroduced into python 3.4 I'm unclear how much this matters.

distributed/comm/tcp.py
```diff
@@ -211,34 +211,47 @@ def closed(self):
         return self.stream is None or self.stream.closed()
 
 
+def _protocol_name(connection_kwargs):
+    """
+    The communication protocol that this connection is using
```
pitrou: I think this is really the wrong way of doing it. If the user asked for `tcp`, it should be given a TCP connection, regardless of what `connection_kwargs` is. Conversely, asking for `tls` should always give a TCP connection (and perhaps bail out if no suitable cert and private key is given, at least for a server connection).

distributed/tests/test_ssl.py
```diff
@@ -0,0 +1,63 @@
...
+        '-passout', 'pass:',
+        '-keyout', 'test.key',
+        '-out', 'test.pem'],
+        cwd=ssl_dir.dirname,
+    )
+
+    p.communicate()
+    assert p.returncode == 0
+
+    return {'ssl_options': {
```
pitrou: I'd rather make it mandatory for it to be a SSLContext. I understand that Tornado accepts several different things for compatibility reasons, but we are not tied by backwards compatibility here.

distributed/tests/test_ssl.py
```diff
@@ -0,0 +1,63 @@
...
+def ssl_kwargs(tmpdir):
+
+    ssl_dir = tmpdir.mkdir("ssl")
+    dirname = ssl_dir.dirname
+
+    keyfile = os.path.join(dirname, 'test.key')
+    certfile = os.path.join(dirname, 'test.pem')
+
+    p = subprocess.Popen([
+        'openssl', 'req',
```
pitrou: Running this at each test invocation makes it mandatory to have OpenSSL installed to run the tests, right?
If so, I'd rather have cert and key files added to the tests directory, and a small script added to regenerate them when needed.
(OpenSSL isn't usually installed on Windows, for instance)

distributed/cli/utils.py
pitrou: The `ssl` module chooses slightly different parameters for servers and clients. Right now the differences may not matter much, but they can start diverging in the future.
At least there should be a comment here about this possible gotcha.

distributed/tests/test_ssl.py
mariusvniekerk: I thought about generating them using python, but that started looking painful pretty quickly.  Pregenerating them sounds like a decent idea.  Guess we should also explicitly tell people to NOT use them for anything other than test purposes.

distributed/client.py
```diff
@@ -343,7 +343,7 @@ class Client(object):
     distributed.scheduler.Scheduler: Internal scheduler
     """
     def __init__(self, address=None, start=True, loop=None, timeout=3,
-                 set_as_default=True, scheduler_file=None):
+                 set_as_default=True, scheduler_file=None, connection_kwargs=None):
```
pitrou: Instead of passing `connection_kwargs` everywhere (here, in the scheduler, in the workers...), perhaps those parameters should be stored as configuration options? That's what we do for other global config options.

distributed/tests/test_ssl.py
pitrou: Yes, generating TLS certificates is generally tedious :-(

docs/source/ssl.rst
mariusvniekerk: we can refer to is in the docs as TLS/SSL?  Like python standard library.

pitrou: Yes, of course.

distributed/client.py
mariusvniekerk: Primary reason for doing it this way is to provide some means of passing in other connection parameters which could be useful for different communication backends.

pitrou: Yes, but we could use the same strategy for those other connection parameters.  We already have `tcp-timeout` in Dask's `config.yml`.

distributed/cli/dask_remote.py
mariusvniekerk: not that sure how easy it is to refactor that out with click, since it relies on things being in the function arguments.

mrocklin: FWIW I prefer keeping the click code somewhat flat (rather than refactored with dependencies).  I think that this keeps things straightforward and easy for new developers to quickly understand, even if slightly repetitive.

distributed/client.py
mrocklin: Depending on the config.yaml file is probably a bad choice for anything except debugging options.  It is annoying to guarantee this presence of this file in all deployment scenarios.  Job schedulers often make it far easier to pass keywords to scripts than to populate environments with files. 

However, it would be nice not to have to specify `--certfile` etc.  whenever starting up dask-workers.  I can see how this would be tedious.  I wonder if there is some way to initially seed connection_kwargs with a section in the config file.

pitrou: There's a misunderstanding :-) I am not asking to depend on `config.yaml`, I am asking (or suggesting) for `--certfile` and `--keyfile` to set a key in the global `config` dictionary instead of being passed as additional options to many internal APIs.

mrocklin: Ah I see, that does make more sense.  I guess the question then becomes about the following tradeoff

1.  The inconvenience of passing around `connection_kwargs` and the long term maintenance costs of this.
2.  The likelihood that we will want to have different connection kwargs within the same process.  

@pitrou seems to have taken on the maintenance of distributed/comm, so he probably has the most say here, especially on point 1.  

On point number two I don't know of anyone who would do this today, but it doesn't seem that unlikely.  People today are already building some strange things.

mariusvniekerk: We probably need to add a flag to the cli to set the default communication method

mrocklin:
@mariusvniekerk I take it you're now in agreement with @pitrou about using globals rather than passing connection kwargs explicitly?

distributed/client.py
```diff
@@ -354,7 +354,7 @@ def __init__(self, address=None, start=True, loop=None, timeout=3,
         self.status = None
         self._pending_msg_buffer = []
         self.extensions = {}
-        self.scheduler_file = scheduler_file
+        self.scheduler_file = scheduler_fileF
```
mrocklin: Typo

mrocklin:
It seems odd to me to depend on global config to determine SSL behavior.  This seems awkward to do programatically (for example if someone creates workers in python) and it seems restrictive if you want to have different systems live within the same process.

pitrou:
> This seems awkward to do programatically (for example if someone creates workers in python) and it seems restrictive if you want to have different systems live within the same process.

Ok, I can understand that. We can try living with `connection_kwargs` and see what happens.

mariusvniekerk:
@mrocklin it basically WAS a global config effectively. @pitrou 's suggested change probably makes it easier to have something like Scheduler using TLS and all workers using TCP.

Essentially we just need to make it easier to choose the communication protocol for a worker/server

mrocklin:
> @mrocklin it basically WAS a global config effectively. @pitrou 's suggested change probably makes it easier to have something like Scheduler using TLS and all workers using TCP.

OK, I'm happy being overruled here if both of you agree.  (after all, you both have more experience on this part of the code than I do). 

I would like to see this pushed in at some point.  What else should be done here?

mariusvniekerk:
Right now it's failing and I'm not quite sure why

mrocklin:
> @mrocklin it basically WAS a global config effectively. @pitrou 's suggested change probably makes it easier to have something like Scheduler using TLS and all workers using TCP.

@mariusvniekerk can you explain this?  I think that previously I could host two schedulers in the same process and give them both different certfiles.  Now I can no longer do this.  

I was talking about globals per-process.  I think that you're talking about globals per-cluster, which is also a valid thing to think about.  I'm curious how does the change from connection_kwargs to using the global config allow different protocols between different nodes within the cluster?

distributed/client.py
mrocklin: > Right now it's failing and I'm not quite sure why

Just in case you didn't see I thought I'd bring this to your attention again @mariusvniekerk

distributed/cli/dask_scheduler.py
```diff
@@ -46,8 +43,18 @@
               help="File to write connection information. "
               "This may be a good way to share connection information if your "
               "cluster is on a shared network file system.")
+@click.option('--certfile', default=config.get('tls-certfile'),
+              help='path to certfile to use for ssl connection')
+@click.option('--keyfile', default=config.get('tls-keyfile'),
+              help='path to keyfile to use for ssl connection')
+@click.option('--default-scheme', type='str', default=config.get("default-scheme"),
+              "Default comminiocation scheme to use.")
```
mrocklin: Running tests locally shows that this needs a keyword name, like `help="Default communication ..."` (also you have a prose typo).

mrocklin: And replace `type='str'` with `type=str`.  This needs to be a type rather than text.

mrocklin:
I'm getting this traceback.  Do we need to make tcp a default somewhere?

```python
Print from stderr
=================

b'distributed.scheduler - INFO - -----------------------------------------------\n'
b'Traceback (most recent call last):\n'
b'  File "/home/mrocklin/anaconda/bin/dask-scheduler", line 11, in <module>\n'
b"    load_entry_point('distributed', 'console_scripts', 'dask-scheduler')()\n"
b'  File "/home/mrocklin/workspace/distributed/distributed/cli/dask_scheduler.py", line 118, in go\n'
b'    main()\n'
b'  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/click/core.py", line 722, in __call__\n'
b'    return self.main(*args, **kwargs)\n'
b'  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/click/core.py", line 697, in main\n'
b'    rv = self.invoke(ctx)\n'
b'  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/click/core.py", line 895, in invoke\n'
b'    return ctx.invoke(self.callback, **ctx.params)\n'
b'  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/click/core.py", line 535, in invoke\n'
b'    return callback(*args, **kwargs)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/cli/dask_scheduler.py", line 86, in main\n'
b'    scheduler.start(addr)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/scheduler.py", line 413, in start\n'
b'    self.listen(addr_or_port)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/core.py", line 197, in listen\n'
b'    deserialize=self.deserialize)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/comm/core.py", line 185, in listen\n'
b'    backend = registry.get_backend(scheme)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/comm/registry.py", line 65, in get_backend\n'
b'    % (scheme, sorted(backends)))\n'
b"ValueError: unknown address scheme 'None' (known schemes: ['inproc', 'tcp', 'tls'])\n"
```

mariusvniekerk:
@mrocklin whilst you could previously host two different schedulers with different certs, you could not have the situation where you wanted to run the scheduler under tls and the workers with normal tcp.

mrocklin:
This still fails for me locally:

```python
==================================================== FAILURES =====================================================
____________________________________________________ test_ssl _____________________________________________________

ssl_kwargs = {'certfile': '/home/mrocklin/workspace/distributed/distributed/tests/test.pem', 'keyfile': '/home/mrocklin/workspace/distributed/distributed/tests/test.key'}
loop = <tornado.platform.epoll.EPollIOLoop object at 0x7fa749e440b8>

    def test_ssl(ssl_kwargs, loop):
    
        backup = {}
        for k in ['default-scheme', 'tls-certfile', 'tls-keyfile']:
>           backup[k] = config[k]
E           KeyError: 'default-scheme'

```

mariusvniekerk:
So I think that's because you have an old config file

mrocklin:
@mariusvniekerk are you able to run tests locally?

Running 

```
py.test  -s distributed/cli/tests/test_dask_scheduler.py::test_defaults
```

I get

```
Print from stderr
=================

b'Traceback (most recent call last):\n'
b'  File "/home/mrocklin/Software/anaconda/bin/dask-scheduler", line 11, in <module>\n'
b"    load_entry_point('distributed==1.16.0+52.g9ead925', 'console_scripts', 'dask-scheduler')()\n"
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 565, in load_entry_point\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2598, in load_entry_point\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2258, in load\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2264, in resolve\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/distributed-1.16.0+52.g9ead925-py3.6.egg/distributed/cli/dask_scheduler.py", line 51\n'
b'    "Default communication scheme to use.")\n'
b'                 ^\n'
b'SyntaxError: positional argument follows keyword argument\n'
```

You may have to install the library to get `dask-scheduler` and friends on your path.  See https://github.com/dask/distributed/issues/951 , which may be an issue.

mrocklin:
Still getting

```python

Print from stderr
=================

b'distributed.scheduler - INFO - -----------------------------------------------\n'
b'Traceback (most recent call last):\n'
b'  File "/home/mrocklin/Software/anaconda/bin/dask-scheduler", line 11, in <module>\n'
b"    load_entry_point('distributed', 'console_scripts', 'dask-scheduler')()\n"
b'  File "/home/mrocklin/workspace/distributed/distributed/cli/dask_scheduler.py", line 118, in go\n'
b'    main()\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/click/core.py", line 722, in __call__\n'
b'    return self.main(*args, **kwargs)\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/click/core.py", line 697, in main\n'
b'    rv = self.invoke(ctx)\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/click/core.py", line 895, in invoke\n'
b'    return ctx.invoke(self.callback, **ctx.params)\n'
b'  File "/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/click/core.py", line 535, in invoke\n'
b'    return callback(*args, **kwargs)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/cli/dask_scheduler.py", line 86, in main\n'
b'    scheduler.start(addr)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/scheduler.py", line 413, in start\n'
b'    self.listen(addr_or_port)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/core.py", line 197, in listen\n'
b'    deserialize=self.deserialize)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/comm/core.py", line 185, in listen\n'
b'    backend = registry.get_backend(scheme)\n'
b'  File "/home/mrocklin/workspace/distributed/distributed/comm/registry.py", line 65, in get_backend\n'
b'    % (scheme, sorted(backends)))\n'
b"ValueError: unknown address scheme 'None' (known schemes: ['inproc', 'tcp', 'tls'])\n"
```

mariusvniekerk:
@pitrou I'm not quite sure why test_ssl is hanging.  Can you take a look if you have some time?

pitrou:
@mariusvniekerk, if you set the log level to "debug" in `config.yaml`, you'll see that the test loops doing:
```
distributed.comm.core - DEBUG - sleeping on connect
distributed.comm.core - DEBUG - sleeping on connect
distributed.comm.tcp - DEBUG - Setting TCP keepalive: nprobes=10, idle=10, interval=2
distributed.comm.tcp - DEBUG - Setting TCP user timeout: 30000 ms
distributed.core - DEBUG - Connection from 'tls://127.0.0.1:54774' to Scheduler
distributed.comm.tcp - DEBUG - Setting TCP keepalive: nprobes=10, idle=10, interval=2
distributed.comm.tcp - DEBUG - Setting TCP user timeout: 30000 ms
distributed.core - DEBUG - Connection from 'tls://127.0.0.1:54776' to Scheduler
distributed.core - DEBUG - Lost connection to 'tls://127.0.0.1:54776' while reading message: SSLError: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:749). Last operation: None
distributed.core - DEBUG - Lost connection to 'tls://127.0.0.1:54774' while reading message: SSLError: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:749). Last operation: None
```

By the way, from a methodology standpoint, I think you should first add lower-level TLS tests in `distributed/comm/tests` before trying to debug higher-level tests.

mariusvniekerk:
@pitrou So i guess thats because i am using a saved self-signed cert?  Can you see if i'm making it correctly?

pitrou:
> So i guess thats because i am using a saved self-signed cert?

Yes, it's probably that. By default, `ssl.create_default_context()` loads the system's CA certificates, and obviously we can't automate creating a test cert/key pair that would be recognized by those CA certificates. One possibility would be to call `load_verify_locations` on the SSLContext used for testing (see https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations).

mariusvniekerk:
Testing with ssl_ctx.verify_mode = ssl.CERT_NONE passes

pitrou:
That's an option, but we'd like to test that certificate checks work in the nominal case :-)

mariusvniekerk:
What's the easiest way to generate a ca for self signing.  I'm not really an expert here.

quasiben:
FWIW, I wrote a post on testing with SSL: http://quasiben.github.io/blog/2016/11/21/proxy-ssl-setup/  I don't have a lot of time at the moment.  If I find some I'd be happy to help

mrocklin:
The widget test failure is odd.  I would say that this is probably an intermittent failure that we just haven't seen before, but it came up three times, which seems particularly unlikely.  I can take a look locally sometime tomorrow.

mrocklin:
I've added a fix for the progress bar tests.  This exposed a small race condition that was there before in the tests (not in the actual code).  I've fixed the race condition, but it *is* somewhat odd that this PR caused the race condition to occur more frequently.  Perhaps some other part of the code has slowed down in the common (non-ssl) case?

distributed/utils_test.py
```diff
@@ -784,3 +795,54 @@ def captured_handler(handler):
...
+    keyfile = os.path.join(root, 'tests', 'test.key')
+
+    assert os.path.exists(certfile)
+    assert os.path.exists(keyfile)
+
+    from .config import config
+
+    backup = {}
+    for k in ['default-scheme', 'tls-certfile', 'tls-keyfile']:
+        backup[k] = config[k]
```
mrocklin: Should this be converted to `config.get(k, None)` to make these tests robust to this entry missing in the config file?  Generally we've so far not assumed that the config file has any values and instead inline defaults in code where appropriate.

mrocklin: I've made this change, and others, in d98768a .  Review appreciated.

mrocklin:
I would like to add a test for the command line interface.  Is there a nice way to generate temporary certfile and keyfiles?

mrocklin:
I've added a CLI test for SSL.  It fails and I'm not sure why.  Any recommendations @mariusvniekerk @pitrou ?

mariusvniekerk:
So this is mostly because the cert we make for testing is not  use the correct certificate authority.  Thus failing.

mrocklin:
Is there a way to do an integrative test here?  This would make me feel much more comfortable about maintaining this long term without accidentally breaking it.

mariusvniekerk:
So @quasiben seems to have done a bunch of tests for ssl things.  We can probably adapt that.

mrocklin:
What is the minimum amount of infrastructure we need to do an end-to-end test on travis?  Can we accomplish this without setting up Docker containers?

quasiben:
You don't need docker containers to test SSL.  It looks like you have the right idea, generate an SSL cert and use that cert in the test.  Is that not working ?

mrocklin:
Comments from myself and then @mariusvniekerk  above

> mrocklin: I've added a CLI test for SSL. It fails and I'm not sure why. Any recommendations @mariusvniekerk @pitrou ?

> mariusvniekerk: So this is mostly because the cert we make for testing is not use the correct certificate authority. Thus failing.

quasiben:
Apologies, I should've looked up.  I'm not sure why there is a `ConnectionRefusedError` error.  I was expecting to see something related to `SSLError`.  @mariusvniekerk, I didn't know it was possible to generate a cert without a Certificate Authority (CA).  Usually, I see that being created first, then generating the cert

hussainsultan:
You can just create a self signed certificate without a CA that should be good enough for testing.  I will look into this more soon.

mariusvniekerk:
probably need to create a self signed cert for `localhost` and ensure that the connection is using a hostname instead of an ip?

mrocklin:
I've added code to push the tls config settings through Nanny and Bokeh Application.  Can anyone provide the necessary statements to create self-signed certs for localhost as Marius suggests?

mrocklin:
I would like to release soon and I would like to have an experimental version of TLS support in that release.  This has become a bit of a priority for me.

mrocklin:
So locally I've created key/certfile using 

    openssl req -x509 -sha256 -nodes -newkey rsa:2048 -days 365 -keyout localhost.key -out localhost.crt

Now when I setup a scheduler and worker I get the following when the worker tries to connect:

```
tornado.application - ERROR - Exception in callback (<socket.socket fd=8, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 8786)>, <function wrap.<locals>.null_wrapper at 0x7f0946001ae8>)
  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/tornado/ioloop.py", line 887, in start
    handler_func(fd_obj, events)
  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/tornado/stack_context.py", line 275, in null_wrapper
    return fn(*args, **kwargs)
  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/tornado/netutil.py", line 272, in accept_handler
    callback(connection, address)
  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/tornado/tcpserver.py", line 244, in _handle_connection
    do_handshake_on_connect=False)
  File "/home/mrocklin/anaconda/lib/python3.6/site-packages/tornado/netutil.py", line 522, in ssl_wrap_socket
    return context.wrap_socket(socket, **kwargs)
  File "/home/mrocklin/anaconda/lib/python3.6/ssl.py", line 401, in wrap_socket
    _context=self, _session=session)
  File "/home/mrocklin/anaconda/lib/python3.6/ssl.py", line 764, in __init__
    raise ValueError("check_hostname requires server_hostname")
ValueError: check_hostname requires server_hostname
^Cdistributed.scheduler - INFO - End scheduler at 'tls://localhost:8786'
```

Does this error mean anything to anyone?

mariusvniekerk:
Yes. You need to make the scheduler etc listen on a hostname not an ip afaik.

mrocklin:
I start the scheduler and worker with 

```
python -m distributed.cli.dask_scheduler --certfile ~/localhost.crt --keyfile ~/localhost.key --no-bokeh --host localhost
python -m distributed.cli.dask_worker tls://127.0.0.1:8786 --keyfile ~/localhost.key --certfile ~/localhost.crt --host localhost --no-nanny
```

The errors happen on the scheduler side.  

On the CLI the address renders with IP addresses (@pitrou is it easy to change this to use the hostname if provided?) but I dug in a bit and it looks like `TCPServer.listen` is being called with the correct `localhost` name, at least in the worker case.  

@pitrou any thoughts?

mrocklin:
Sorry, I meant to say "at least in the scheduler case" above

pitrou:
Ok, so this kind of issue is why I think we should first figure out what the "recommended" setup would be, instead of rushing an implementation.

My suggestion would be:
* a single dedicated CA cert for all TLS connections
* a TLS key/cert pair for the scheduler, signed by the CA cert
* a shared TLS key/cert pair for all workers, signed by the CA cert
* no hostname checking, as that's not necessary in this setup (unless you want to protect against something truly weird, e.g. a worker masquerading as the scheduler)

I don't know what @ebolyen, @hussainsultan and @mariusvniekerk have in mind, or if my proposal would be compatible with their needs.

ebolyen:
@pitrou what you describe would work perfectly fine for us, we just need some kind of mechanism to prevent unauthorized code execution. In this case TLS acts as a glorified HMAC digest (and the encryption is nice too).

This doesn't matter to us, but would the certs still need to signed for the context of TLS-client *and* TLS-server in your setup? Or is it suggesting that each connection-type is configurable independently (e.g. worker to worker could be configured differently avoiding the dual-context cert)?

mrocklin:
I had a nice chat with @quasiben about this, which also helped to slightly improve my understanding of how SSL is used in practice.  I think he is going to check in with someone else who might have more experience.

Some questions arose:

1.  Given that the clients don't serve anything does it make sense for them to have certificates?
2.  How painful is it to provide certificates to all workers?  This is especially a question when there are many workers on a large pool of compute resources. 
3.  Are there other options for inter-worker encryption that might make more sense?

pitrou:
> Given that the clients don't serve anything does it make sense for them to have certificates?

If they don't, anybody can submit code to execute or get data from the workers.

> How painful is it to provide certificates to all workers?

They can all use the same cert. This is a regular deployment problem, IMHO.

> Are there other options for inter-worker encryption that might make more sense?

That depends. What specific problem would those options solve that SSL doesn't?

mrocklin:
> > Given that the clients don't serve anything does it make sense for them to have certificates?

> If they don't, anybody can submit code to execute or get data from the workers.

I guess stepping back even further, what security guarantees are we trying to provide here?  This is again a question for @hussainsultan @ebolyen @mariusvniekerk .

When I think of how we use SSL in HTTPS it seems strange for the client to have certificates.  Usually identity of the client is handled by some other means.  SSL seems to handle identity of the server and encryption between client and server.

pitrou:
> When I think of how we use SSL in HTTPS it seems strange for the client to have certificates. Usually identity of the client is handled by some other means.

That makes sense because most HTTPS sites have no a priori knowledge of who might connect to them (and they might simply not care, when clients merely browse the public area of the HTTPS site). But a Distributed deployment generally isn't available to everyone, only to a controlled set of users.

Of course, we could have a notion of usernames and passwords to authenticate Distributed clients, but that's much more involved for us than authenticating them using TLS certs.

ebolyen:
> When I think of how we use SSL in HTTPS it seems strange for the client to have certificates. Usually identity of the client is handled by some other means. SSL seems to handle identity of the server and encryption between client and server.

TLS/SSL has two modes, typically it is only used for server authentication, but the protocol supports both server *and* client. Even browsers can do this, so it is possible (and valid) to create a webserver that authenticates the clients connecting to it, but it's so rarely used that no one knows how to add a client cert to their browser. 

I don't think there is much value in doing server authentication against a scheduler, because it doesn't grant you much other than the guarantee that you are talking to the *right* scheduler (all HTTPS usually does).

What I think is more useful in this context, is the ability authenticate entities that can submit arbitrary code because that is actually a pretty privileged operation. That's where client-authentication in TLS is useful (though rarely used). There are of course other ways to solve this problem (baking authentication into the protocol for example), but TLS is attractive because it can accomplish this by just wrapping TCP frames without changing your underlying protocol.

mrocklin:
OK, that makes sense to me.  Thanks both for the explanation.

pitrou:
> I don't think there is much value in doing server authentication against a scheduler, because it doesn't grant you much other than the guarantee that you are talking to the right scheduler

It would guarantee the client isn't the victim of a man-in-the-middle attack, where someone would intercept the communication in order to capture data computed on the cluster (for example if some legitimate user is trying to submit jobs from home or a public Wifi network).

mariusvniekerk:
@pitrou @mrocklin So what do we want do with this one?

I think the route we want long term is that we want things like client-key worker-key scheduler-key config values.  That probably requires adding some kind of role flag to the various rpc methods so that we know which value to go retrieve.  I am probably not the right person to make that degree of change to this codebase.

pitrou:
@mariusvniekerk, I have started my own version of this branch.

What to do for the various RPC functions is a good question. I think `rpc()` and `ConnectionPool.connect()` will have to grow an additional argument to pass connection parameters (such as TLS params). It's up to higher layers (such as the Scheduler and Worker objects) to decide what those connection parameters are, depending on which peer is contacted.

mariusvniekerk:
@pitrou Yeah that sounds correct.  Guess we just need to add a PeerType enum for that as well

mrocklin:
Another approach to this is happening over at https://github.com/dask/distributed/pull/1034

mrocklin:
Closing in favor of #1034

PR #871, opened 2017-02-13
WIP: specifying resources in the hostfile

jmontoyam:
Hi @mrocklin,
what do you think about the proposed solution?

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -232,8 +232,16 @@ class SSHCluster(object):
 
         self.scheduler_addr = scheduler_addr
         self.scheduler_port = scheduler_port
-        self.nthreads = nthreads
-        self.nprocs = nprocs
+
+        if not isinstance(nthreads, list):
+            self.nthreads = [nthreads] * len(worker_addrs)
+        else:
+            self.nthreads = nthreads
+
+        if not isinstance(nprocs, list):
+            self.nprocs = [nprocs] * len(worker_addrs)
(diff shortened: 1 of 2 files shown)


distributed/deploy/ssh.py
```diff
@@ -256,8 +264,9 @@ def __init__(self, scheduler_addr, scheduler_port, worker_addrs, nthreads = 0, n
 
         # Start worker nodes
         self.workers = []
-        for i, addr in enumerate(worker_addrs):
-            self.add_worker(addr)
+        for addr, nthreads, nprocs in zip(worker_addrs, self.threads,
+                                          self.nprocs):
+            self.add_worker(addr, nthreads, nprocs)
```
mrocklin: Ideally we would make this general to all keywords.  I expect users to become confused why only certain keywords are respected in this way.

jmontoyam: Thanks for the feedback @mrocklin, what you mean is that I should take into account that the user might want to use the hostfile to specify any feature of the worker?, right?

mrocklin: Yes, that it what I mean.  Or, if not all features, it should at least be pretty trivial to extend this to almost all features, ideally without a long chain of if-else statements.  I hesitate to add any user-facing feature where some things work and some very related things don't work.  This tends to fill my inbox with confusing user reports :)

mrocklin:
Closing as stale.  Please reopen if work picks up again

PR #874, opened 2017-02-14
Test large uneven frames through whole pipeline

mrocklin:
Previously we weren't testing moving large amounts of data through
non-deserializing channels.  This was problematic because we set up an
erroneous assertion in a previous PR.  This removes that assertion and
adds a regression test.


mrocklin:
This is experiencing a possibly unrelated test failure.  Going to try to track it down before merging.

PR #876, opened 2017-02-17
Maintain fortran ordering in serialization

mrocklin:
Add order information in numpy serialization

This seems suboptimal though, because we're coercing to C order, sending over a wire, and then coercing back to Fortran order.  Is there a way to just move the data in whichever way it prefers?  Odd issues arise if we drop the `ascontiguousarray` call.

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -43,7 +43,10 @@ def serialize_numpy_ndarray(x):
     else:
         dt = x.dtype.str
 
-    if np.isfortran(x) or not x.shape:
+    if not x.shape:
+        strides = x.strides
+        data = x.ravel().view('u1').data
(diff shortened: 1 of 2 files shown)


mrocklin:
cc @teoliphant

JohnCrickett:
Should one or more of the test parameters to test_dumps_serialize_numpy not have order='F' set to properly test this?

mrocklin:
Do these parameters suffice?

         np.asfortranarray(np.random.random((5, 5))),
         np.asfortranarray(np.random.random((5, 5)))[::2, :],
         np.asfortranarray(np.random.random((5, 5)))[:, ::2],

JohnCrickett:
I wrote that before the second commit :) 

Looks reasonable to me.

PR #877, opened 2017-02-17
Add scheduler address file

mrocklin:
This makes it easier to deploy Dask on systems that have a shared network file system by writing connection information to a file

```
$ dask-scheduler --scheduler-file foo.json
$ dask-worker --scheduler-file foo.json
```
```python
c = Client(scheduler_file='foo.json')
```

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -42,8 +42,10 @@ logger = logging.getLogger('distributed.scheduler')
               help="User xheaders in bokeh app for ssl termination in header")
 @click.option('--pid-file', type=str, default='',
               help="File to write the process PID")
+@click.option('--scheduler-file', type=str, default='',
+              help="File to write connection information")
 def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
-         bokeh_whitelist, prefix, use_xheaders, pid_file):
+         bokeh_whitelist, prefix, use_xheaders, pid_file, scheduler_file):
 
     if pid_file:
         with open(pid_file, 'w') as f:
@@ -70,7 +72,8 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
         with ignoring(ImportError):
             from distributed.bokeh.scheduler import BokehScheduler
             services[('bokeh', bokeh_internal_port)] = BokehScheduler
-    scheduler = Scheduler(loop=loop, services=services)
+    scheduler = Scheduler(loop=loop, services=services,
+                          scheduler_file=scheduler_file)
     scheduler.start(addr)
 
     bokeh_proc = None
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -43,7 +43,9 @@ logger = logging.getLogger('distributed.scheduler')
 @click.option('--pid-file', type=str, default='',
               help="File to write the process PID")
 @click.option('--scheduler-file', type=str, default='',
-              help="File to write connection information")
+              help="File to write connection information. "
+              "This may be a good way to share connection information if your "
+              "cluster is on a shared network file system.")
 def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
          bokeh_whitelist, prefix, use_xheaders, pid_file, scheduler_file):
 
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -80,9 +80,9 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
 
     bokeh_proc = None
     if _bokeh:
-        if bokeh_port == 0:
-            bokeh_port = open_port()
-        try:
+        if bokeh_port == 0:          # This is a hack and not robust
+            bokeh_port = open_port() # This port may be taken by the OS
+        try:                         # before we successfully pass it to Bokeh
             from distributed.bokeh.application import BokehWebInterface
             bokeh_proc = BokehWebInterface(http_port=http_port,
                     tcp_port=scheduler.port, bokeh_port=bokeh_port,
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -13,7 +13,7 @@ import click
 
 import distributed
 from distributed import Scheduler
-from distributed.utils import ignoring
+from distributed.utils import ignoring, open_port
 from distributed.http import HTTPScheduler
 from distributed.cli.utils import (check_python_3, install_signal_handlers,
                                    uri_from_host_port)
@@ -78,6 +78,8 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
 
     bokeh_proc = None
     if _bokeh:
+        if bokeh_port == 0:
+            bokeh_port = open_port()
         try:
             from distributed.bokeh.application import BokehWebInterface
             bokeh_proc = BokehWebInterface(http_port=http_port,
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -2,6 +2,7 @@ from __future__ import print_function, division, absolute_import
 
 import atexit
 from datetime import timedelta
+import json
 import logging
 import os
 import shutil
@@ -43,7 +44,7 @@ def handle_signal(sig, frame):
 
 
 @click.command()
-@click.argument('scheduler', type=str)
+@click.argument('scheduler', type=str, required=False)
 @click.option('--worker-port', type=int, default=0,
               help="Serving worker port, defaults to randomly assigned")
 @click.option('--http-port', type=int, default=0,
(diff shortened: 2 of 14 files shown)


mrocklin:
cc @pitrou when you get a chance (no rush) can I ask you to take a brief look at the solution in https://github.com/dask/distributed/pull/877/commits/bbd9c58048372537254b37a2d597fb484634d5dd ?

This has to do with using zero-port values in comms.

mrocklin:
I'm not sure how best to handle port zero here.  Waiting on comments from @pitrou (who may still be out on vacation).  Planning to let this simmer until Monday otherwise.

mrocklin:
@bnaul @pwolfram @rabernat these changes may interest you if you want to deploy dask on a SGE/SLURM/etc. cluster more easily.

distributed/cli/tests/test_cli_utils.py
```diff
@@ -17,6 +17,7 @@ def test_uri_from_host_port():
     assert f('', 456, 123) == 'tcp://:456'
     assert f('', None, 123) == 'tcp://:123'
     assert f('', None, 0) == 'tcp://'
+    assert f('', 0, 123) == 'tcp://:0'
```
pitrou: Uh, no. This should really be spelt `tcp://` (i.e. the unknown port number is omitted).

pitrou: To elaborate on this, we don't want to have several different representations for a given endpoint address. The CLI can silently convert `0` to `None` if desired, though.

distributed/cli/dask_scheduler.py
```diff
@@ -70,11 +72,14 @@ def del_pid_file():
...
             services[('bokeh', bokeh_internal_port)] = BokehScheduler
-    scheduler = Scheduler(loop=loop, services=services)
+    scheduler = Scheduler(loop=loop, services=services,
+                          scheduler_file=scheduler_file)
     scheduler.start(addr)
 
     bokeh_proc = None
     if _bokeh:
+        if bokeh_port == 0:
+            bokeh_port = open_port()
```
pitrou: We should really let Bokeh open its port for real here. The "open port" might be reused by something else otherwise (even in another unrelated process)...

distributed/utils.py
```diff
@@ -618,3 +618,18 @@ def shutting_down(l=_shutting_down):
     to early bind this function rather than look it up when calling it,
     since at shutdown module globals may be cleared.
     """
+
+
+def open_port(host=''):
+    """ Return a probably-open port
+
+    There is a chance that this port will be taken by the operating system soon
```
pitrou: Yes, but another process (or even another library function, e.g. imagine setting up logging into a socket) might take it instead of the intended service. The right way to do this is really to let the intended service (e.g. Bokeh) bind on port 0 and then query the actual port. If Bokeh doesn't support this, then it should :-)

distributed/cli/tests/test_cli_utils.py
pitrou: (of course we may want to change the canonical representation if that really makes things better, but that doesn't look to be the case here)

distributed/cli/dask_worker.py
```diff
@@ -76,9 +77,12 @@ def handle_signal(sig, frame):
               help="Internal use only")
 @click.option('--resources', type=str, default='',
               help='Resources for task constraints like "GPU=2 MEM=10e9"')
+@click.option('--scheduler-file', type=str, default='',
+              help='Filename to JSON encoded scheduler information.'
```
pitrou: There's probably a missing space after the dot.

distributed/scheduler.py
```diff
@@ -422,6 +426,17 @@ def start(self, addr_or_port=8786, start_queues=True):
             for k, v in self.services.items():
                 logger.info("%11s at: %25s", k, '%s:%d' % (self.ip, v.port))
 
+        if self.scheduler_file:
+            with ignoring(ImportError):
```
pitrou: I think it would be nice to refactor scheduler file handling into separate functions (see also client.py and dask_worker.py).

Also why the `ImportError`?

distributed/scheduler.py
```diff
@@ -422,6 +426,17 @@ def start(self, addr_or_port=8786, start_queues=True):
...
+                with open(self.scheduler_file, 'w') as f:
+                    json.dump(self.identity(), f, indent=2)
+
+            fn = self.scheduler_file  # remove file when we close the process
+            def del_scheduler_file():
+                if os.path.exists(fn):
+                    os.remove(fn)
+            atexit.register(del_scheduler_file)
```
pitrou: Ideally the lifetime would be the Scheduler object's, not the process'...

distributed/cli/tests/test_cli_utils.py
mrocklin: As written previously sending this address to `start` would result in a server choosing the default port, not a random port.  What should be default behavior for `start('tcp://')`?

rbubley:
Probably worth commenting in the docs that although this might be suitable for true network shares, it could cause problems on systems where files are instead synchronised, e.g. with rsync, or by using a cloud file store provider (like dropbox or google drive).

distributed/cli/dask_scheduler.py
mrocklin: The problem is that bokeh is running in a separate process, and we would like to communicate the address back to the main process.  This choice was made as an easy hack that solved the existing problem most of the time.  A better solution would be ideal though.

distributed/utils.py
mrocklin: Bokeh supports this, it's getting the information back that's a problem.  perhaps that isn't necessary though.

distributed/scheduler.py
mrocklin: Is there a way to reliably call a function on the death of an object?  `__del__` doesn't seem to work in all cases.

distributed/cli/tests/test_cli_utils.py
pitrou: Is this `Scheduler.start()`? If you are passing it `tcp://`, it binds to a random port:
```
>>> from distributed import Scheduler
>>> s = Scheduler()
>>> s.start('tcp://')
distributed.scheduler - INFO -   Scheduler at:  tcp://192.168.1.68:39230
<tornado.concurrent.Future at 0x7f4272531160>
>>> s.listen_address
'tcp://0.0.0.0:39230'
>>> s.port
39230
>>> s.address
'tcp://192.168.1.68:39230'
```

distributed/cli/dask_scheduler.py
pitrou: AFAIU, our user's usecase is for programmatic starting of a scheduler, so something that works most of the time isn't really satisfactory. Two solutions I can think of:
1. have Bokeh grow something like the scheduler file argument
2. allow Bokeh to take an existing (bound) socket

The first solution sounds less hairy than the second.

distributed/scheduler.py
pitrou: We should really backport `wearef.finalize` from Python 3. I can do that in a separate PR.

distributed/cli/tests/test_cli_utils.py
mrocklin: The originally reported problem was here

```
mrocklin@carbon:~/workspace/distributed$ dask-scheduler  --port 0
distributed.scheduler - INFO -   Scheduler at:  tcp://192.168.1.115:8786
distributed.scheduler - INFO -        http at:              0.0.0.0:9786
distributed.scheduler - INFO -       bokeh at:              0.0.0.0:8788
distributed.bokeh.application - INFO - Web UI: http://127.0.0.1:8787/status/
```

It looks like this is due to `uri_from_host_port`

```python
In [1]: from distributed.cli.utils import uri_from_host_port

In [2]: uri_from_host_port('', 0, 8786)
Out[2]: 'tcp://:8786'
```

Which is used in dask-scheduler

pitrou: Yes, what I mean is that `uri_from_host_port` can return `tcp://` here, it doesn't have to return `tcp://:0`.

distributed/scheduler.py
mrocklin: ImportError has been removed.  I'm lightly in favor of leaving this code here.  I think that there are two competing objectives here:

1.  Pull out blocks of self-related code from high level control flow.  We don't want readers to have to sift through the details of scheduler files when looking at basic instantiation code.
2.  Reduce the amount of indirection that has to happen when looking through a codebase.  We don't want readers to have to navigate to many different parts of the file to understand how something works.

I tend to favor the second objective in cases like these, particularly because these are only a few lines, there is no repetition of this code elsewhere, and the code is easy the skip over when reading (it isn't highly connected to surrounding lines.  However, my preference here is only slight.  I'm happy to change this if your preference is stronger.

distributed/cli/tests/test_cli_utils.py
mrocklin: OK, I think that this is now fixed in `unparse_host_port`

distributed/cli/dask_worker.py
mrocklin: Fixed missing space.

mrocklin:
Thanks for the comment @rbubley , see https://github.com/dask/distributed/pull/877/commits/c65b2a4aa0e8a81ab43bfde22bc019b670c559a5

mrocklin:
I plan to merge this soon if there are no objections.

distributed/scheduler.py
pitrou: Since `weakref.finalize` is now available as a compatibility backport, perhaps you can try to use that?

distributed/cli/dask_scheduler.py
pitrou: Perhaps at least comment that this is an unreliable hack? :-)

distributed/scheduler.py
mrocklin: Now using finalize

distributed/cli/dask_scheduler.py
mrocklin: Added a disclaimer about open_port as a comment

PR #878, opened 2017-02-17
Add serializer for keras models

### LINKED ISSUE #841: Serialize Keras Model
For Dask.distributed to work well with a deep learning library like Keras, it may be helpful for Dask to know how to serialize Keras models.

It looks like there are `keras.models.Model.save_weights_to_hdf5_group` and `keras.models.load_model` functions that store/load a Keras model to and from an HDF5 file.  We can probably copy most of this logic to create Dask serialize/deserialize functions that convert a Model to a series of frames of bytes. See [dask serialization docs for detailed information](http://distributed.readthedocs.io/en/latest/protocol.html#serializing-data).

Additionally, we would want to extend the `sizeof` function in `distributed.sizeof` to quickly provide a rough estimate of how many bytes are contained within a model (hopefully without having to serialize everything).

If we have these two pieces then Keras models should be able to move freely throughout Dask clusters.

This issue does require skill and attention to detail, but fortunately does not require much expertise in Dask itself.

bnaul:
Fixes #841. Turned out to be easier than expected: `model._updated_config()` is what's used inside `model.save()` to extract the metadata, and `get_weights()` returns a list of `numpy` arrays of weight parameters.

Only weird bit I see is the repeated registrations: depending on how the model is instantiated, it could have any of those types (they all inherit from `keras.engine.training.Model` but it didn't seem to work if I didn't register all three; maybe there's a better way though?).

### DIFF
--- a/distributed/protocol/keras.py
+++ b/distributed/protocol/keras.py
@@ -6,6 +6,11 @@ from ..utils import log_errors, ensure_bytes
 
 
 def serialize_keras_model(model):
+    import keras
+    if keras.__version__ < '1.2.0':
+        raise ImportError("Need Keras >= 1.2.0. "
+                "Try pip install keras --upgrade --no-deps")
+
     header = model._updated_config()
     weights = model.get_weights()
     headers, frames = list(zip(*map(serialize, weights)))
--- a/distributed/protocol/keras.py
+++ b/distributed/protocol/keras.py
@@ -10,7 +10,7 @@ def serialize_keras_model(model):
     weights = model.get_weights()
     headers, frames = list(zip(*map(serialize, weights)))
     header['headers'] = headers
-    header['lengths'] = [len(L) for L in frames]
+    header['nframes'] = [len(L) for L in frames]
     frames = [frame for L in frames for frame in L]
     return header, frames
 
@@ -19,7 +19,7 @@ def deserialize_keras_model(header, frames):
     from keras.models import model_from_config
     n = 0
     weights = []
-    for head, length in zip(header['headers'], header['lengths']):
+    for head, length in zip(header['headers'], header['nframes']):
         x = deserialize(head, frames[n: n + length])
         weights.append(x)
         n += length
--- a/distributed/protocol/keras.py
+++ b/distributed/protocol/keras.py
@@ -1,0 +1,36 @@
+from __future__ import print_function, division, absolute_import
+
+from .serialize import register_serialization, serialize, deserialize
+
+from ..utils import log_errors, ensure_bytes
+
+
+def serialize_keras_model(model):
+    header = model._updated_config()
+    weights = model.get_weights()
+    headers, frames = list(zip(*map(serialize, weights)))
+    header['headers'] = headers
+    header['lengths'] = [len(L) for L in frames]
+    frames = [frame for L in frames for frame in L]
+    return header, frames
+
+
+def deserialize_keras_model(header, frames):
+    from keras.models import model_from_config
+    n = 0
+    weights = []
+    for head, length in zip(header['headers'], header['lengths']):
+        x = deserialize(head, frames[n: n + length])
+        weights.append(x)
+        n += length
+    model = model_from_config(header)
+    model.set_weights(weights)
+    return model
+
+
+register_serialization('keras.engine.training.Model', serialize_keras_model,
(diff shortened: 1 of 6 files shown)


mrocklin:
Cool find.  The only other thing to do here is to convert the numpy arrays into bytes-like objects. You can probably do this with something like the following:

```python
weights = model.get_weights()
headers, frames = list(zip(*map(serialize, weights)))
header['headers'] = headers
header['lengths'] = [len(L) for L in frames]
frames = [frame for L in frames for frame in L]
```

On the deserialize side:

```python
n = 0
weights = []
for head, length in zip(header['headers'], header['lengths']):
    x = deserialize(head, frames[n: n + length])
    weights.append(x)
    n += length
```

mrocklin:
The full chain test would probably be something like the following:

```python
data = {'x': to_serialize(model)}
frames = dumps(data)
result = loads(frames)
... test against data['x'] and result['x']
```

Adding this test would reveal the issue referred to above.

distributed/protocol/keras.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def serialize_keras_model(model):
+    header = model._updated_config()
+    weights = model.get_weights()
+    headers, frames = list(zip(*map(serialize, weights)))
+    header['headers'] = headers
+    header['lengths'] = [len(L) for L in frames]
+    frames = [frame for L in frames for frame in L]
+    return header, frames
```
mrocklin: OK, the failure here is my fault.  Dask is already using the term `'lengths'` and so there is a conflict.  Naming this something else should work.  However I've also identified another internal issue.  Resolving that now.

mrocklin:
This is in.  Thanks @bnaul !

amelius15:
I was just wondering what the right way to utilize keras with dask.distributed is. I'm getting errors no matter what I try. What I've tried include:

`result = client.map(model.predict, data)
result = client.submit(lambda x,y: x.predict(y), model, data)`

but no matter what I try, I get Theano errors when I run that don't occur when I run it without dask.distributed, and seems to be related to not having the model working properly. Could it be something to do with it being serialized in some way?

My model is:
`
        model = Sequential()
        model.add(Conv2D(filters=filters, kernel_size=(1, 5), padding='same', activation=None, input_shape=input_shape))
        model.add(GaussianNoise(0.1))
        model.add(Activation('relu'))
        model.add(MaxPooling2D(pool_size=(1, 2)))
        model.add(Dropout(0.2))
        model.add(Conv2D(filters=filters, kernel_size=(4, 1), padding='same', activation='relu'))
        model.add(MaxPooling2D(pool_size=(4, 1)))
        model.add(Reshape((-1, len_seq)))
        model.add(Dropout(0.2))
        model.add(Bidirectional(LSTM(n_lstm, return_sequences=True)))
        model.add(Bidirectional(LSTM(n_lstm)))
        model.add(Dropout(0.2))
        model.add(Dense(2, activation='sigmoid'))
`
and the error I get is:

> theano.gof.fg.MissingInputError: Long error message
> Input 0 of the graph (indices start from 0), used to compute if{}(keras_learning_phase, Elemwise{add,no_inplace}.0, Elemwise{add,no_inplace}.0), was not provided and not given a value. Use the Theano flag exception_verbosity='high', for more information on this error. 

This seem to only happen with dask.distributed, doesn't ever occur when running it with cpu natively. 

@mrocklin @bnaul I know this might not be the right place to ask the question, but you two are the only link between Keras and Dask.distributed I've found on the entire Internet. 

Thanks!

mrocklin:
FYI I am away this week with very sporadic internet contact.

@bnaul do you have any interest in engaging here?

bnaul:
Hi @amelius15,

It seems that there are still issues with `keras` and multiple threads, at least with the `tensorflow` backend (see e.g. fchollet/keras#5896 and fchollet/keras#5640). Someone in the latter thread mentioned that they were able to do inference in parallel by persisting the model structure to disk and re-loading it but I haven't confirmed that this works. In my own work I worked around the problem by using a process pool instead of thread pool; I'm not sure how you're launching the workers but you could try `--nprocs <n>` to see if that helps. If not, I'm afraid there isn't quite enough info here to be able to debug; maybe you could post a self-contained gist that recreates your error?

I've been meaning to spend some time exploring the different possible options here, curious to hear what else you find!

PR #880, opened 2017-02-20
Report connection errors in local_client()

pitrou:


pitrou:
I'm not sure how to write a test for this. It's not obvious how to force local_client() to fail connecting while keeping a functional Scheduler which can report the Future's outcome to the main Client.

pitrou:
@mrocklin, what do you think about this approach?

mrocklin:
This seems fine to me.  I like that you add a timeout to `local_client`.  That seems like a good solution that gives users control when they know they will send out many local_client tasks.

PR #881, opened 2017-02-20
Fix distributed for ShareDict

pitrou:
`ShareDict.copy()` is not defined. Since copying should remove the sharing, we could define it to return a plain `dict` instead, but that would be a slightly weird API. Conversely, returning a ShareDict with a single underlying `dict` would be a tad sub-optimal...


pitrou:
@mrocklin

mrocklin:
Looks good to me.  Perhaps we should separately consider implementing `ShareDict.copy`

PR #883, opened 2017-02-20
Add weakref.finalize() backport

pitrou:


distributed/compatibility.py
```diff
@@ -68,3 +68,146 @@ def isqueue(o):
     from functools import singledispatch
 except ImportError:
     from singledispatch import singledispatch
+
+try:
+    from weakref import finalize
+except ImportError:
+    # Backported from Python 3.6
```
mrocklin: Do we need to include a license for this?

pitrou: I usually don't and assume people are reasonable enough to understand this means the code is covered by the said project's license. The PSF license is quite large and would have to be shipped as a separate file in any case (but it's basically a permissive open source license).

mrocklin: OK, I'm happy to defer to your judgement here

Issue #865, opened 2017-02-09
Uploading zip module


bmaisonn:
Hi,
I'm using distributed.upload_file to upload some .py files that i need to use in my worker.
Also I would like to import a module that I zipped.
[ZipImport](https://docs.python.org/3.5/library/zipimport.html) gives the possibility to do that easily however it's not implemented in worker.upload_file yet.
Do you mind if I had this functionality as per egg archives ?
Except if i missed something, the code is straightforward (See the change [here](https://github.com/bmaisonn/distributed/commit/08e09de63250901e2df865db02a0b5a8a279eae2))
Regards,
Bertrand

mrocklin:
Sounds good to me.  I wasn't aware of zipimport.  Your solution looks good to me.  Is it easy to add a test for this functionality as well?

bmaisonn:
I suppose that test_client is the right place for that, right ?
I see an 'egg' file in test folder, can I add a zip file too ?
Thx

mrocklin:
Adding a small file there seems reasonable, yes.

mrocklin:
Creating and reliably deleting a file (such as with the tmpfile
contextmanager) is slightly preferred if convenient.

pitrou:
Closing after the functionality was merged in #886.

PR #886, opened 2017-02-20
Support to upload modules from zip files

bmaisonn:
Fix issue (#865).

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -14,7 +14,7 @@ import sys
 from threading import Thread, Semaphore
 from time import sleep
 import traceback
-import zipfile
+import zipfile, sys
 
 import mock
 import pytest
@@ -1102,42 +1102,45 @@ def test_upload_file(c, s, a, b):
         import myfile
         return myfile.f()
 
-    with tmp_text('myfile.py', 'def f():\n    return 123') as fn:
-        yield c._upload_file(fn)
-
-    sleep(1)  # TODO:  why is this necessary?
-    x = c.submit(g, pure=False)
-    result = yield x._result()
-    assert result == 123
-
-    with tmp_text('myfile.py', 'def f():\n    return 456') as fn:
-        yield c._upload_file(fn)
-
-    y = c.submit(g, pure=False)
-    result = yield y._result()
-    assert result == 456
-
+    try:
+        for value in [123, 456]:
+            with tmp_text('myfile.py', 'def f():\n    return {}'.format(value)) as fn:
+                yield c._upload_file(fn)
+        
+            x = c.submit(g, pure=False)
+            result = yield x._result()
+            assert result == value
+    finally:
+        # Ensure that this test won't impact the others
+        if 'myfile' in sys.modules:
+            del sys.modules['myfile']
+            
 @gen_cluster(client=True)
 def test_upload_file_zip(c, s, a, b):
     def g():
         import myfile
         return myfile.f()
 
-    with tmp_text('myfile.py', 'def f():\n    return 123') as fn_my_file, \
-         tmp_text('init.py', '') as fn_init:
-        try:
-            with zipfile.ZipFile('myfile.zip', 'w') as z:
-                for fn in (fn_my_file, fn_init):
-                    z.write(fn, arcname=os.path.basename(fn))
-            yield c._upload_file('myfile.zip')
-        finally:
-            if os.path.exists('myfile.zip'):
-                os.remove('myfile.zip')
-
-    sleep(1)  # TODO:  why is this necessary?
-    x = c.submit(g, pure=False)
-    result = yield x._result()
-    assert result == 123
+    try:
+        for value in [123, 456]:
+            with tmp_text('myfile.py', 'def f():\n    return {}'.format(value)) as fn_my_file:
+                with zipfile.ZipFile('myfile.zip', 'w') as z:
+                    z.write(fn_my_file, arcname=os.path.basename(fn_my_file))
+                yield c._upload_file('myfile.zip')
+                    
+                x = c.submit(g, pure=False)
+                result = yield x._result()
+                assert result == value
+    finally:
+        # Ensure that this test won't impact the others
+        if os.path.exists('myfile.zip'):
+            os.remove('myfile.zip')
+        if 'myfile' in sys.modules:
+            del sys.modules['myfile']
+        for path in sys.path:
+            if os.path.basename(path) == 'myfile.zip':
+                sys.path.remove(path)
+                break
 
 @gen_cluster(client=True)
 def test_upload_large_file(c, s, a, b):
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -14,6 +14,7 @@ import sys
 from threading import Thread, Semaphore
 from time import sleep
 import traceback
+import zipfile
 
 import mock
 import pytest
@@ -1116,6 +1117,27 @@ def test_upload_file(c, s, a, b):
     result = yield y._result()
     assert result == 456
 
+@gen_cluster(client=True)
+def test_upload_file_zip(c, s, a, b):
+    def g():
+        import myfile
+        return myfile.f()
+
+    with tmp_text('myfile.py', 'def f():\n    return 123') as fn_my_file, \
+         tmp_text('init.py', '') as fn_init:
+        try:
+            with zipfile.ZipFile('myfile.zip', 'w') as z:
+                for fn in (fn_my_file, fn_init):
+                    z.write(fn, arcname=os.path.basename(fn))
+            yield c._upload_file('myfile.zip')
+        finally:
+            if os.path.exists('myfile.zip'):
+                os.remove('myfile.zip')
+
+    sleep(1)  # TODO:  why is this necessary?
+    x = c.submit(g, pure=False)
+    result = yield x._result()
+    assert result == 123
 
 @gen_cluster(client=True)
 def test_upload_large_file(c, s, a, b):
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -465,6 +465,12 @@ class WorkerBase(Server):
                         reload(import_module(pkg.project_name))
                     if not pkgs:
                         logger.warning("Found no packages in egg file")
+                if ext == '.zip':
+                    logger.info("Reload module %s from .zip file", name)
+                    if out_filename not in sys.path:
+                        sys.path.insert(0, out_filename)
+                    name = name.split('-')[0]
+                    reload(import_module(name))
             except Exception as e:
                 logger.exception(e)
                 return {'status': 'error', 'exception': dumps(e)}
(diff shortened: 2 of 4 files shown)


distributed/tests/test_client.py
```diff
@@ -1116,6 +1117,27 @@ def g():
...
+        try:
+            with zipfile.ZipFile('myfile.zip', 'w') as z:
+                for fn in (fn_my_file, fn_init):
+                    z.write(fn, arcname=os.path.basename(fn))
+            yield c._upload_file('myfile.zip')
+        finally:
+            if os.path.exists('myfile.zip'):
+                os.remove('myfile.zip')
+
+    sleep(1)  # TODO:  why is this necessary?
```
mrocklin: What happens if this is not present?

distributed/tests/test_client.py
```diff
@@ -1116,6 +1117,27 @@ def g():
...
+    x = c.submit(g, pure=False)
```
mrocklin: You probably don't need `pure=False` here.

distributed/tests/test_client.py
```diff
@@ -1116,6 +1117,27 @@ def g():
...
     assert result == 456
 
+@gen_cluster(client=True)
+def test_upload_file_zip(c, s, a, b):
+    def g():
+        import myfile
+        return myfile.f()
+
+    with tmp_text('myfile.py', 'def f():\n    return 123') as fn_my_file, \
+         tmp_text('init.py', '') as fn_init:
```
mrocklin: Why the file `init.py`?

distributed/tests/test_client.py
pitrou: The one second resolution may depend on the filesystem. Try `importlib.invalidate_caches()` instead. (see https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches ). It isn't needed on Python 2.

distributed/tests/test_client.py
```diff
@@ -1116,6 +1117,27 @@ def g():
     result = yield y._result()
     assert result == 456
 
+@gen_cluster(client=True)
+def test_upload_file_zip(c, s, a, b):
+    def g():
+        import myfile
+        return myfile.f()
+
+    with tmp_text('myfile.py', 'def f():\n    return 123') as fn_my_file, \
```
pitrou: I would really prefer this kind of test to create and use a temporary directory. You'll also need to temporarily insert the directory at the front of `sys.path`.
(you can find an example of such pattern in https://github.com/python/cpython/blob/master/Lib/test/test_import/__init__.py#L198-L215 )

distributed/tests/test_client.py
```diff
@@ -1116,6 +1117,27 @@ def g():
     result = yield y._result()
     assert result == 456
 
+@gen_cluster(client=True)
+def test_upload_file_zip(c, s, a, b):
```
pitrou: The test would be more robust if it scrubbed out `myfile` from `sys.modules` at the start and at the end.

distributed/worker.py
```diff
@@ -465,6 +465,12 @@ def upload_file(self, comm, filename=None, data=None, load=True):
                         reload(import_module(pkg.project_name))
                     if not pkgs:
                         logger.warning("Found no packages in egg file")
+                if ext == '.zip':
+                    logger.info("Reload module %s from .zip file", name)
+                    if out_filename not in sys.path:
+                        sys.path.insert(0, out_filename)
+                    name = name.split('-')[0]
```
pitrou: Just a question: why this line?

distributed/worker.py
```diff
@@ -465,6 +465,12 @@ def upload_file(self, comm, filename=None, data=None, load=True):
...
+                    reload(import_module(name))
```
pitrou: `invalidate_caches` should probably be called just before this line.

distributed/tests/test_client.py
bmaisonn: Actually you're right. I thought that the init.py was necessary to make a valid module but actually it's not. I tried to remove it and the test passed as well.
Thx for the reveiw

distributed/worker.py
bmaisonn: I kept this line from ('.py', '.pyc') case
```
                if ext in ('.py', '.pyc'):
                    logger.info("Reload module %s from .py file", name)
                    name = name.split('-')[0]
                    reload(import_module(name))
```
Actually I don't know what it is for, so i can remove it in the zip case

distributed/tests/test_client.py
bmaisonn: There's no case where we call invalidate_cache in the upload_file method. 
Should I add it in all cases ? 
In the same way the sleep call is in all the upload_file tests.
Should I remove it from all the tests ?

```
                if ext in ('.py', '.pyc'):
                    logger.info("Reload module %s from .py file", name)
                    name = name.split('-')[0]
                    reload(import_module(name))
                if ext == '.egg':
                    import pkg_resources
                    sys.path.append(out_filename)
                    pkgs = pkg_resources.find_distributions(out_filename)
                    for pkg in pkgs:
                        logger.info("Load module %s from egg", pkg.project_name)
                        reload(import_module(pkg.project_name))
                    if not pkgs:
                        logger.warning("Found no packages in egg file")
                if ext == '.zip':
                    logger.info("Reload module %s from .zip file", name)
                    if out_filename not in sys.path:
                        sys.path.insert(0, out_filename)
                    invalidate_caches()
                    reload(import_module(name))
```

distributed/tests/test_client.py
bmaisonn: In the same way `pure=False` is used in the other upload_file test but i agree that it could be removed.
Should I remove it from the other tests ?

distributed/tests/test_client.py
bmaisonn: I'm not sure to get this one.
This module is uploaded locally on workers , right ? The workers aren't restarted before each test ?

distributed/tests/test_client.py
pitrou: You should at least try it and see it if works, yes :-)

distributed/tests/test_client.py
pitrou: With `gen_cluster`, the workers should run in the current process, so they would inherit the current (process-wide) import state. For example, perhaps the `myfile` module seen in this test actually comes from another one of the tests here.

It seems all tests for `upload_file` are written in this style. I think it would deserve fixing.

bmaisonn: ok, so i'm going to apply all those modifications to all tests :)

bmaisonn: all the upload_file_* tests i mean

bmaisonn:
It starts to be confusing here...
I modified my test to integrate your reviews (Using invalid_cache, remove sleep and pure=False). 
I also modified the test to make the upload twice with different code and check if the modifications are correctly taken into account (As it's done in test_upload_file):

```
@gen_cluster(client=True)
def test_upload_file_zip(c, s, a, b):
    def g():
        import myfile
        return myfile.f()

    try:
        for value in [123, 456]:
            with tmp_text('myfile.py', 'def f():\n    return {}'.format(value)) as fn_my_file:
                with zipfile.ZipFile('myfile.zip', 'w') as z:
                    z.write(fn_my_file, arcname=os.path.basename(fn_my_file))
                yield c._upload_file('myfile.zip')
                    
                x = c.submit(g)
                result = yield x._result()
                assert result == value
    finally:
        if os.path.exists('myfile.zip'):
            os.remove('myfile.zip')
        if 'myfile' in sys.modules:
            del sys.modules['myfile']
        for path in sys.path:
            if os.path.basename(path) == 'myfile.zip':
                sys.path.remove(path)
                break
```

With this code the second test `assert result == 456` fails as `result == 123`. So the second file isn't updated.
The same happens in test_upload_file is I remove pure=False OR sleep(1).
I really don't understand why it fails here ...

mrocklin:
You need the pure=False if you're doing this twice and expect to get
different results.  Otherwise the second call will just reference the first.

bmaisonn:
Ok, i misunderstood the meaning of pure.
It works now
Thx

distributed/tests/test_client.py
bmaisonn: I added the invalid_caches and commonalized some code:
 
```
try:
                name, ext = os.path.splitext(filename)
                names_to_import = []
                if ext in ('.py', '.pyc'):
                    names_to_import.append(name)
                if ext in ('.egg', '.zip'):
                    if out_filename not in sys.path:
                        sys.path.insert(0, out_filename)
                    if ext == '.egg':
                        import pkg_resources
                        pkgs = pkg_resources.find_distributions(out_filename)
                        for pkg in pkgs:
                            names_to_import.append(pkg.project_name)
                    elif ext == '.zip':
                        names_to_import.append(name)
                
                if not names_to_import:
                    logger.warning("Found nothing to import from %s", filename)
                else:
                    invalidate_caches()
                    for name in names_to_import:
                        logger.info("Reload module %s from %s file", name, ext)
                        reload(import_module(name))
```
I removed the sleep in my test case and it works fine (But it also works without the call to invalidate_caches).
However the test still fails in the 'test_upload_file' case which only uploads .py or .pyc files

bmaisonn: maybe I found the answer here : 
http://stackoverflow.com/questions/8122734/pythons-imp-reload-function-is-not-working

pitrou: That may be the case indeed. You may want to try to remove the cached bytecode file. On Python 3, see https://docs.python.org/3/library/importlib.html#importlib.util.cache_from_source . On Python 2, you will need to reimplement that function yourself.

bmaisonn: Ok i'll try this. It starts to be much bigger than the original change :)
By the way, I'm not used to write code for Python2/Python3
Is there a good way to do that ?  What I see on forums is something like that:

```
            if hasattr(importlib, "invalidate_caches"):
                importlib.invalidate_caches()
```

Is that right ?

pitrou: Yes, though we have a `distributed.compatibility` module where you could add a `invalidate_import_caches` function.

bmaisonn: I made the changes to delete the pyc file
It works fine now without the sleep(1) call

bmaisonn:
I commited a new change to integrate the reviews
Both tests pass in python 2 and python 3

distributed/worker.py
```diff
@@ -452,19 +452,31 @@ def upload_file(self, comm, filename=None, data=None, load=True):
...
+                            names_to_import.append(pkg.project_name)
+                    elif ext == '.zip':
+                        names_to_import.append(name)
+                
+                if not names_to_import:
+                    logger.warning("Found nothing to import from %s", filename)
+                else:
+                    for name in names_to_import:
+                        logger.info("Reload module %s from %s file", name, ext)
+                        invalidate_caches()
```
pitrou: Just a nit: you can pull `invalidate_caches()` out of the `for` loop.

pitrou:
Thanks a lot!
There's one thing missing: you should update the `Client.upload_file` docstring.

bmaisonn:
I moved the call to invalidate_caches and updated the documentation
Thx for you help!

pitrou:
I don't have anything to add here, perhaps @mrocklin wants to take a last look.

mrocklin:
Happy to trust you two on this.  Thank you both for the effort and review.

PR #887, opened 2017-02-21
In-process communication transports

pitrou:
This adds a new communication transport named `inproc` that is backed by in-process queues. The aim is to later allow local cluster instances to use such communication channels instead of TCP, to reduce the I/O overhead. That will be a bit more involved than I initially assumed, though, due to the notion of host and port being ingrained in some places (especially the scheduler).

### DIFF
--- a/distributed/comm/inproc.py
+++ b/distributed/comm/inproc.py
@@ -1,0 +1,230 @@
+from __future__ import print_function, division, absolute_import
+
+from collections import namedtuple
+import itertools
+import os
+import sys
+import threading
+import weakref
+
+from tornado import gen, locks
+from tornado.ioloop import IOLoop
+from tornado.queues import Queue
+
+from ..compatibility import finalize
+from ..utils import get_ip
+from .core import (connectors, listeners, Comm, Listener, CommClosedError,
+                   )
+
+
+ConnectionRequest = namedtuple('ConnectionRequest',
+                               ('c2s_q', 's2c_q', 'c_loop', 'c_addr',
+                                'conn_event', 'close_request'))
+
+_Close = object()
+
+
+class Manager(object):
+
+    def __init__(self):
+        self.listeners = weakref.WeakValueDictionary()
+        self.addr_suffixes = itertools.count(1)
+        self.ip = get_ip()
+        self.lock = threading.Lock()
+
+    def add_listener(self, addr, listener):
+        with self.lock:
+            if addr in self.listeners:
+                raise RuntimeError("already listening on %r" % (addr,))
+            self.listeners[addr] = listener
+
+    def remove_listener(self, addr):
+        with self.lock:
+            del self.listeners[addr]
+
+    def get_listener_for(self, addr):
+        with self.lock:
+            self.validate_address(addr)
+            return self.listeners.get(addr)
+
+    def new_address(self):
+        return "%s/%d/%s" % (self.ip, os.getpid(), next(self.addr_suffixes))
+
+    def validate_address(self, addr):
+        """
+        Validate the address' IP and pid.
+        """
+        ip, pid, suffix = addr.split('/')
+        if ip != self.ip or int(pid) != os.getpid():
+            raise ValueError("inproc address %r does not match host (%r) or pid (%r)"
+                             % (addr, self.ip, os.getpid()))
+
+
+global_manager = Manager()
+
+def new_address():
+    """
+    Generate a new address.
+    """
+    return 'inproc://' + global_manager.new_address()
+
+
+class InProc(Comm):
+    """
+    An established communication based on a pair of in-process queues.
+    """
+
+    def __init__(self, peer_addr, read_q, write_q, write_loop,
+                 close_request, deserialize=True):
+        self._peer_addr = peer_addr
+        self.deserialize = deserialize
+        self._read_q = read_q
+        self._write_q = write_q
+        self._write_loop = write_loop
+        self._closed = False
+        # A "close request" event shared between both comms
+        self._close_request = close_request
+
+        self._finalizer = finalize(self, self._get_finalizer())
+        self._finalizer.atexit = False
+
+    def _get_finalizer(self):
+        def finalize(write_q=self._write_q, write_loop=self._write_loop,
+                     close_request=self._close_request):
+            if not close_request.is_set():
+                logger.warn("Closing dangling queue in %s" % (r,))
+                close_request.set()
+                write_loop.add_callback(write_q.put_nowait, _Close)
+
+        return finalize
+
+    def __repr__(self):
+        return "<InProc %r>" % (self._peer_addr,)
+
+    @property
+    def peer_address(self):
+        return self._peer_addr
+
+    @gen.coroutine
+    def read(self, deserialize=None):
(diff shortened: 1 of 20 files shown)


distributed/comm/inproc.py
```diff
@@ -0,0 +1,335 @@
...
+        q = self._q
+        if q:
+            return q[0]
+        elif default is not self._omitted:
+            return default
+        else:
+            raise QueueEmpty
+
+
+def _maybe_deserialize(msg):
```
pitrou: I would rather avoid such code in the communication layer. @mrocklin, do you think it's ok to put this in `distributed.protocol`?

pitrou: Or perhaps you want to suggest another approach.

distributed/comm/tests/test_comms.py
```diff
@@ -440,3 +640,130 @@ def test_tcp_many_listeners():
...
+
+    q.put_nowait(in_value)
+    out_value = yield comm.read()
+    yield comm.close()
+    check_out(out_value)
+
+@gen.coroutine
+def check_deserialize(addr):
+    # Create a valid Serialized object
+    # (if using serialize(), it will lack a compression header)
```
pitrou: Is this by design? Do you think we can change `distributed.protocol.loads` to assume no compression if the compression header is absent?

distributed/comm/inproc.py
mrocklin: I don't have a strong opinion on this.

distributed/comm/tests/test_comms.py
mrocklin: If no compression information is present in the header I would assume that the data is not compressed.  Is this not normally default behavior?

pitrou: No, `loads()` always calls `decompress()` which fails with a KeyError if `compression` is not found in the header.

mrocklin: Avoiding that behavior seems sensible to me

distributed/comm/tests/test_comms.py
```diff
@@ -188,12 +203,104 @@ def client_communicate(key, delay=0):
...
+    futures = [client_communicate(key=i, delay=0.001) for i in range(N)]
+    yield futures
+    assert set(l) == {1234} | set(range(N))
+
+    assert len(client_addresses) == N + 1
+    assert listener.contact_address not in client_addresses
+
+
+def run_coro(func, *args, **kwargs):
+    return func(*args, **kwargs)
```
mrocklin: `from .compatibility import apply` ?

distributed/protocol/serialize.py
```diff
@@ -195,6 +202,14 @@ def deserialize(self):
         frames = decompress(self.header, self.frames)
         return deserialize(self.header, frames)
 
+    def __eq__(self, other):
+        return (isinstance(other, Serialized) and
+                other.header == self.header and
+                other.frames == self.frames)
+
+    def __ne__(self, other):
+        return not (self == other)
```
mrocklin: Why was defining `eq` necessary?  Should we also now define `__hash__`?

distributed/tests/test_core.py
```diff
@@ -150,29 +152,58 @@ def listen_on(cls, *args):
...
             yield assert_can_connect_locally_6(server.port)
 
+    # InProc
 
-def test_rpc(loop):
-    @gen.coroutine
-    def f():
-        server = Server({'ping': pingpong})
-        server.listen(8883)
+    with listen_on(Server, 'inproc://') as server:
```
mrocklin: The inproc stuff still seems pretty experimental to me.  I hesitate to integrate it into our core tests.  It would be nice to make sure that we can extract this cheaply in the future if we decide we don't want to keep it around.

pitrou:
> It would be useful to include an integration test that actually uses inproc for some trivial dask.array work.

As stated, it doesn't work yet. The followup PRs start doing the cleanup and refactor work needed to make it happen.

distributed/comm/tests/test_comms.py
pitrou: Uh, why not indeed. I guess I never use `apply` :-)

distributed/protocol/serialize.py
pitrou: Just for the unit tests. It has no use in the rest of the code base. `__hash__` isn't necessary as we don't use those objects as keys in associative containers.

distributed/tests/test_core.py
pitrou: That was meant to verify that the basic `core` infrastructure works with inproc. I think a "grep inproc" would easily point the places where it's used :-)

distributed/protocol/serialize.py
mrocklin: We don't in this codebase, that's true, but including objects in containers is generally useful.  I'm fairly confident that I've put serialized objects in containers before while testing.  I'm somewhat confident that this will come up for someone in the future when building something atypical.  If it is cheap to keep this door open then I would prefer to do so.

distributed/tests/test_core.py
mrocklin: Ah, I see.  I was misreading things because of how the diff was laid out.  I thought that you were replacing existing tests with inproc tests.  Looking at the flat file makes it more clear that this isn't occurring.

distributed/protocol/serialize.py
pitrou: I see. For `Serialize` objects we can easily piggyback on the original object's `__hash__` method. For `Serialized` object it's quite dangerous to assume hashability while the original object may not be hashable.

pitrou:
> Merge this and then merge those later?

This would have my preference. Reordering is not really doable: the refactors are based on this PR.

distributed/protocol/serialize.py
mrocklin: Concerned about mutation?  I'm generally willing to assume that the user won't change values after giving them to us.  We don't really have a reasonable way of proceeding correctly if we don't make this assumption.

pitrou: My reasoning was that if a `Serialized` object is somehow meant to represent the original object, then its hashability should also reflect the original object's hashability.

To be frank, I'm not sure what hashability really brings for `Serialize` and `Serialized` objects, but I'm happy to defer to your judgement.

mrocklin: I don't have a specific example concern in this case, I've just been bitten by projects in the past where people define `__eq__` for convenience and then suddenly I can't use those objects with other python containers.  I now avoid `__eq__` by a strong habit unless there is a good API reason to implement it (such as with dask.array).  So this may just be an irrational avoidance on my part, but it's still something I'd prefer to avoid if it's not particularly costly to avoid having an `__eq__` method.

pitrou: Ok, so I think I'll remove `Serialized.__eq__` (which doesn't seem to be needed actually) and also implement `Serialize.__hash__`.

mrocklin:
I would also still like to release before merging this if possible.  I will try to get that done tonight.

mrocklin:
This has a merge conflict.  I'm +1 otherwise.

PR #894, opened 2017-02-22
Local exception replay

### LINKED ISSUE #868: Exception reporting/rerunning
When a task errors, all tasks that depend upon it also error, and the user finds this out when they do a `gather` on some end-product. Whilst the full traceback is printed, which is often enough to pinpoint the problem, the user cannot debug without running the whole graph locally using `get_sync`. 
What we would like is to be able to run just the one task that failed, locally, with the specific inputs/corner case that caused the failure, without having to run the whole computation locally. The scheduler already has the information about which task failed (i.e., erred even though none of the inputs had), and the inputs that it needed (which are probably still in memory, or can be recreated on the cluster).

One step around this is to use submit/futures instead of delayed (because futures report their status), but that is a less flexible programming paradigm, and doesn't fit with collections, nor do we know the inputs of a given future unless there is a local data structure representing the graph that delayed would have created.

martindurant:
Fixes #868 

Could probably do with more complex test cases.

Example
------

```python
In [2]: c = Client()
In [3]: def div(x, y):
   ...:     return x / y
   ...: 
In [4]: def inc(x):
   ...:     return x + 1
   ...: 
In [5]: x = c.submit(div, 1, 0)
In [6]: y = c.submit(inc, x)
In [7]: y.result()  # should err

In [8]: pdb
Automatic pdb calling has been turned ON

In [9]: c.recreate_error_locally(y)
> <ipython-input-3-398a43a7781e>(2)div()
      1 def div(x, y):
----> 2     return x / y

ipdb> pp x
1
ipdb> pp y
0
```

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -380,6 +380,8 @@ class Client(object):
 
         from distributed.channels import ChannelClient
         ChannelClient(self)  # registers itself on construction
+        from distributed.recreate_exceptions import ExceptionsClient
+        ExceptionsClient(self)
 
     def __str__(self):
         if hasattr(self, '_loop_thread'):
@@ -2380,59 +2382,6 @@ class Client(object):
             dsk = merge(c.dask for c in collections)
 
         return dsk
-
-    @gen.coroutine
-    def _get_futures_error(self, future):
-        """
-        Ask the scheduler details of the sub-task of the given failed future
-
-        Parameters
-        ----------
-        future: future that failed
-
-        Returns
-        -------
-        The function that failed, its arguments, and the keys that it depends
-        on.
-        """
-        futures = futures_of(future)
-        deps, cause, task = yield self.scheduler.cause_of_failure(
-                keys=[f.key for f in futures])
-        function, args, kwargs = Worker._deserialize(None, **task)
-        raise gen.Return((function, args, kwargs, deps))
-
-    @gen.coroutine
-    def _get_specific_keys(self, keys):
-        # Inform scheduler that we want these keys.
-        futures = self._graph_to_futures({}, keys)
-
-        # Get data from remote workers, pack into args and kwargs
-        raise gen.Return(self._gather(futures))
-
-    def recreate_error_locally(self, future):
-        """
-        For a failed calculation, perform the blamed task locally for debugging.
-
-        Parameters
-        ----------
-        future: future that failed
-            The same thing as was given to ``gather``, but came back with
-            an exception/stack-trace.
-
-        Returns
-        -------
-        Nothing; the function runs and should raise an exception, allowing
-        the debugger to run.
-        """
-        function, args, kwargs, deps = sync(self.loop, self._get_futures_error,
-                                            future)
-        data = sync(self.loop, self._get_specific_keys, deps)
-        data = dict(zip(deps, data))
-        args = pack_data(args, data)
-        kwargs = pack_data(kwargs, data)
-
-        # run the function, hopefully trigger exception.
-        func(*args, **kwargs)
 
 Executor = Client
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -915,7 +915,7 @@ class Client(object):
             bad_keys = set()
             for key in keys:
                 if (key not in self.futures or
-                    self.futures[key].status in failed):
+                        self.futures[key].status in failed):
                     exceptions.add(key)
                     if errors == 'raise':
                         try:
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -380,8 +380,8 @@ class Client(object):
 
         from distributed.channels import ChannelClient
         ChannelClient(self)  # registers itself on construction
-        from distributed.recreate_exceptions import ExceptionsClient
-        ExceptionsClient(self)
+        from distributed.recreate_exceptions import ReplayExceptionClient
+        ReplayExceptionClient(self)
 
     def __str__(self):
         if hasattr(self, '_loop_thread'):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -36,7 +36,7 @@ from .compatibility import Queue as pyQueue, Empty, isqueue
 from .core import connect, rpc, clean_exception, CommClosedError
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
-from .worker import dumps_function, dumps_task
+from .worker import dumps_function, dumps_task, Worker
 from .utils import (All, sync, funcname, ignoring, queue_to_iterator,
         tokey, log_errors, str_graph)
 from .versions import get_versions
@@ -2381,6 +2381,58 @@ class Client(object):
 
         return dsk
 
+    @gen.coroutine
+    def _get_futures_error(self, future):
+        """
+        Ask the scheduler details of the sub-task of the given failed future
+
+        Parameters
+        ----------
+        future: future that failed
+
+        Returns
+        -------
+        The function that failed, its arguments, and the keys that it depends
+        on.
+        """
+        futures = futures_of(future)
+        deps, cause, task = yield self.scheduler.cause_of_failure(
+                keys=[f.key for f in futures])
+        function, args, kwargs = Worker._deserialize(None, **task)
+        raise gen.Return((function, args, kwargs, deps))
+
+    @gen.coroutine
+    def _get_specific_keys(self, keys):
+        # Inform scheduler that we want these keys.
+        futures = self._graph_to_futures({}, keys)
+
+        # Get data from remote workers, pack into args and kwargs
+        raise gen.Return(self._gather(futures))
+
+    def recreate_error_locally(self, future):
+        """
+        For a failed calculation, perform the blamed task locally for debugging.
+
+        Parameters
+        ----------
+        future: future that failed
+            The same thing as was given to ``gather``, but came back with
+            an exception/stack-trace.
+
+        Returns
+        -------
+        Nothing; the function runs and should raise an exception, allowing
+        the debugger to run.
+        """
+        function, args, kwargs, deps = sync(self.loop, self._get_futures_error,
+                                            future)
+        data = sync(self.loop, self._get_specific_keys, deps)
+        data = dict(zip(deps, data))
+        args = pack_data(args, data)
+        kwargs = pack_data(kwargs, data)
+
+        # run the function, hopefully trigger exception.
+        func(*args, **kwargs)
 
 Executor = Client
 
--- a/distributed/recreate_exceptions.py
+++ b/distributed/recreate_exceptions.py
@@ -1,0 +1,98 @@
+from __future__ import print_function, division, absolute_import
+
+import logging
+from tornado import gen
+from .client import futures_of
+from.worker import _deserialize
+
+logger = logging.getLogger(__name__)
+
+
+class ExceptionScheduler(object):
+    """ A plugin for the scheduler to recreate exceptions locally
+
+    This adds the following routes to the scheduler
+
+    *  channel-subscribe
+    *  channel-unsubsribe
+    *  channel-append
+    """
+    def __init__(self, scheduler):
+        self.scheduler = scheduler
+
+        handlers = {'cause_of_failure': self.cause_of_failure}
+
+        self.scheduler.handlers.update(handlers)
+        self.scheduler.extensions['exceptions'] = self
+
+    def cause_of_failure(self, *args, keys=None, **kwargs):
+        for key in keys:
+            if key in self.scheduler.exceptions_blame:
+                cause = self.scheduler.exceptions_blame[key]
+                # cannot serialize sets
+                return {'deps': list(self.scheduler.dependencies[cause]),
+                        'cause': cause,
+                        'task': self.scheduler.tasks[cause]}
+
+
+class ExceptionsClient(object):
+    def __init__(self, client):
+        self.client = client
+        self.client.extensions['exceptions'] = self
+        # monkey patch
+        self.client.recreate_error_locally = self.recreate_error_locally
+        self.client._get_futures_error = self._get_futures_error
+        self.client.get_futures_error = self.get_futures_error
+
+    @property
+    def scheduler(self):
+        return self.client.scheduler
+
+    @gen.coroutine
+    def _get_futures_error(self, future):
+        futures = futures_of(future)
+        out = yield self.scheduler.cause_of_failure(
+            keys=[f.key for f in futures])
+        deps, cause, task = out['deps'], out['cause'], out['task']
+        function, args, kwargs = _deserialize(**task)
+        raise gen.Return((function, args, kwargs, deps))
+
+    def get_futures_error(self, future):
+        """
+        Ask the scheduler details of the sub-task of the given failed future
+
+        Parameters
+        ----------
+        future: future that failed
+
+        Returns
+        -------
+        The function that failed, its arguments, and the keys that it depends
+        on.
+        """
+        return sync(self.loop, self._get_futures_error, future)
+
+    def recreate_error_locally(self, future):
+        """
+        For a failed calculation, perform the blamed task locally for debugging.
+
+        Parameters
+        ----------
+        future: future that failed
+            The same thing as was given to ``gather``, but came back with
+            an exception/stack-trace.
+
+        Returns
+        -------
+        Nothing; the function runs and should raise an exception, allowing
+        the debugger to run.
+        """
+        function, args, kwargs, deps = get_futures_error(future)
+        futures = self._graph_to_futures({}, deps)
+        data = self.gather(futures)
+        data = dict(zip(deps, data))
+        args = pack_data(args, data)
+        kwargs = pack_data(kwargs, data)
+
+        # run the function, hopefully trigger exception.
+        func(*args, **kwargs)
--- a/distributed/recreate_exceptions.py
+++ b/distributed/recreate_exceptions.py
@@ -91,14 +91,25 @@ class ReplayExceptionClient(object):
         """
         Ask the scheduler details of the sub-task of the given failed future
 
+        When a future evaluates to a status of "error", i.e., an exception
+        was raised in a task within its graph, we an get information from
+        the scheduler. This function gets the details of the specific task
+        that raised the exception and led to the error, but does not fetch
+        data from the cluster or execute the function. To actually run the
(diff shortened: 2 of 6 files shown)


martindurant:
Question: are the methods I added useful in their own right, should they have their own tests too?

When we are OK with the above, I will write documentation before merger.

martindurant:
Note on Travis failure: apparently in py2, calling the class method is strict: "TypeError: unbound method _deserialize() must be called with Worker instance as first argument (got NoneType instance instead)". Actually, `self` is not used at all, and this is a static method - any objection to decorating it as such?

distributed/client.py
```diff
@@ -2381,6 +2381,58 @@ def collections_to_dsk(collections, optimize_graph=True, **kwargs):
...
+        function, args, kwargs = Worker._deserialize(None, **task)
+        raise gen.Return((function, args, kwargs, deps))
+
+    @gen.coroutine
+    def _get_specific_keys(self, keys):
+        # Inform scheduler that we want these keys.
+        futures = self._graph_to_futures({}, keys)
+
+        # Get data from remote workers, pack into args and kwargs
+        raise gen.Return(self._gather(futures))
```
mrocklin: This doesn't need to be a coroutine.  It never yields.

It might also be enough to replace calls to `_get_specific_keys` with just the graph_to_futures and _gather calls directly?  I think that there is a cost to adding a new method.  Future readers will have to backtrack to find what it does.  I think that this indirection makes less sense when the method is short and when it is used in relatively few places.

distributed/client.py
```diff
@@ -2381,6 +2381,58 @@ def collections_to_dsk(collections, optimize_graph=True, **kwargs):
...
+        """
+        function, args, kwargs, deps = sync(self.loop, self._get_futures_error,
+                                            future)
+        data = sync(self.loop, self._get_specific_keys, deps)
+        data = dict(zip(deps, data))
+        args = pack_data(args, data)
+        kwargs = pack_data(kwargs, data)
+
+        # run the function, hopefully trigger exception.
+        func(*args, **kwargs)
```
mrocklin: If possible it would be nice to make two such functions, one that is a coroutine and has all of the logic, and then a second that just calls `sync` on the first.  This makes testing easier and also lets people use this functionality if they're using the asynchronous/tornado API (which happens in advanced situations).

distributed/scheduler.py
```diff
@@ -2909,6 +2910,13 @@ def worker_objective(self, key, worker):
         start_time = comm_bytes / BANDWIDTH + stack_time
         return (start_time, self.worker_bytes[worker])
 
+    def cause_of_failure(self, stream=None, keys=None):
+        for key in keys:
+            if key in self.exceptions_blame:
+                cause = self.exceptions_blame[key]
+                # cannot serialize sets
+                return (list(self.dependencies[cause]), cause,
+                        self.tasks[cause])
```
mrocklin: We might consider sending back a dict, which might be a bit more self-documenting.

distributed/client.py
```diff
@@ -2381,6 +2381,58 @@ def collections_to_dsk(collections, optimize_graph=True, **kwargs):
...
+        Returns
+        -------
+        The function that failed, its arguments, and the keys that it depends
+        on.
+        """
+        futures = futures_of(future)
+        deps, cause, task = yield self.scheduler.cause_of_failure(
+                keys=[f.key for f in futures])
+        function, args, kwargs = Worker._deserialize(None, **task)
+        raise gen.Return((function, args, kwargs, deps))
```
mrocklin: Is there value to returning the dependencies?

distributed/client.py
mrocklin: I think it would be useful to test this function

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,98 @@
...
+
+import logging
+from tornado import gen
+from .client import futures_of
+from.worker import _deserialize
+
+logger = logging.getLogger(__name__)
+
+
+class ExceptionScheduler(object):
```
mrocklin: "Exceptions" may be a bit too broad of a term to claim for this functionality.  Dask handles exceptions in a variety of ways.

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,98 @@
...
+    """ A plugin for the scheduler to recreate exceptions locally
+
+    This adds the following routes to the scheduler
+
+    *  channel-subscribe
+    *  channel-unsubsribe
+    *  channel-append
```
mrocklin: These appear to be copied over

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    def __init__(self, scheduler):
+        self.scheduler = scheduler
+
+        handlers = {'cause_of_failure': self.cause_of_failure}
+
+        self.scheduler.handlers.update(handlers)
```
mrocklin: Perhaps combine the two lines above

    self.scheduler.handlers['cause_of_failure'] = self.cause_of_failure

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @property
+    def scheduler(self):
+        return self.client.scheduler
+
+    @gen.coroutine
+    def _get_futures_error(self, future):
+        futures = futures_of(future)
+        out = yield self.scheduler.cause_of_failure(
+            keys=[f.key for f in futures])
```
mrocklin: What is the behavior here if only one of the futures has failed?  Is this tested?

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        function, args, kwargs, deps = get_futures_error(future)
+        futures = self._graph_to_futures({}, deps)
+        data = self.gather(futures)
+        data = dict(zip(deps, data))
+        args = pack_data(args, data)
+        kwargs = pack_data(kwargs, data)
+
+        # run the function, hopefully trigger exception.
+        func(*args, **kwargs)
```
mrocklin: If convenient it would be nice to have this as an asynchronous function first, and then wrap it in sync to produce the synchronous one.  We try to keep both APIs complete.

distributed/tests/test_client.py
```diff
@@ -3750,3 +3750,37 @@ def test_dont_clear_waiting_data(c, s, a, b):
...
+
+    f = c.compute(tot)
+    yield _wait(f)
+    assert f.status == 'error'
+
+    function, args, kwargs, deps = yield c._get_futures_error(f)
+    assert function.__name__ == 'makes_error'
+    assert args ==  (0,)
+    with pytest.raises(ZeroDivisionError):
+        function(*args, **kwargs)
```
mrocklin: Given the use of futures_of above it would be nice to also include an example where a collection fails.  Perhaps something like the following:

```python
b = db.range(10, npartitions=4)
b = b.map(lambda x: 1 / x)
b = c.persist(b)

with pytest.raises(NotImplementedError):
    yield c._recreate_error_locally(b)
```

martindurant:
The failing test seems to pull a non-serialized task from scheduler.cause_of_failure, within a bag.reify.

Note that the async _recreate_error_locally does *not* execute the failing function by design, since that would not actually enable debugging.

martindurant:
Tries manually with a  dataframe, and it does the right thing. Am I right in thinking that this is a bag-specific fuse effect?

martindurant:
NB: this feels pretty hacky, I don't know if there are other possible cases. Also, recreate_error.exec_tuple must exist in some form elsewhere, I just don't know where to find it.

distributed/recreate_exceptions.py
```diff
@@ -124,6 +131,13 @@ def recreate_error_locally(self, future):
...
+        out = sync(self.client.loop, self._recreate_error_locally, future)
+        if len(out) == 3 and isinstance(out[2], dict):
+            func, args, kwargsfunc(*args, **kwargs)
+        else:
+            exec_tuple(out)
+
+
+def exec_tuple(t):
+    newt = (exec_tuple(v) if isinstance(v, tuple) else v for v in t[1:])
+    return t[0](*newt)
```
mrocklin: This feels like it might not handle all corner cases.  I recommend using `dask.get` here instead to ensure that we cover the full spec.  The relevant internal function is probably `dask/async.py::_execute_task`, though relying on public API (dask.get) is probably nicer.

distributed/recreate_exceptions.py
```diff
@@ -124,6 +131,13 @@ def recreate_error_locally(self, future):
...
         the debugger to run.
         """
-        func, args, kwargs = sync(
-                self.client.loop, self._recreate_error_locally, future)
-        func(*args, **kwargs)
+        out = sync(self.client.loop, self._recreate_error_locally, future)
+        if len(out) == 3 and isinstance(out[2], dict):
+            func, args, kwargsfunc(*args, **kwargs)
+        else:
+            exec_tuple(out)
```
mrocklin: It would be nice to replicate exactly whatever logic the worker does here.  Otherwise I expect we'll run into confusing corner cases.

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,143 @@
...
+        cause: the key that failed
+        task: the definition of that key
+        deps: keys that the task depends on
+        """
+
+        keys = kwargs.pop('keys', [])
+        for key in keys:
+            if isinstance(key, list):
+                key = tuple(key)
+                key = tuple(key)
```
mrocklin: duplicate?

mrocklin: Perhaps drop the check and just call 

    key = tuple(key)  # ensure not a list from msgpack

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,143 @@
...
+        # only get errors for futures that errored.
+        futures = [f for f in futures_of(future) if f.status == 'error']
+        out = yield self.scheduler.cause_of_failure(
+            keys=[f.key for f in futures])
+        deps, cause, task = out['deps'], out['cause'], out['task']
+        if isinstance(task, dict):
+            function, args, kwargs = _deserialize(**task)
+            raise gen.Return((function, args, kwargs, deps))
+        else:
+            raise gen.Return(task)
```
mrocklin: At some point we need to normalize between these two cases.  Doing so earlier might result in overall cleaner code.  Perhaps do something like the following?

```python
if not isinstance(task, dict):
    function = dask.get
    args = ({'x': task}, 'x')
    kwargs = {}

raise gen.Return((function, args, kwargs, deps))
```

martindurant: Thank you, that's just the kind of logic I was looking for - but I'll have a read in Worker and see how it is done there too.

martindurant: It turns out this is exactly what `_deserialize()` does if passed `task=`.

martindurant:
@mrocklin , would you say functionality is complete here? Should this PR include docs, and if so, where would they fit in?

mrocklin:
First thing to do would be to ensure that the docstring is comprehensive and then add it to the API documentation.

We should probably build a debug doc at some point that includes the various ways to get information when things go wrong.

distributed/tests/test_client.py
```diff
@@ -3751,6 +3751,117 @@ def test_dont_clear_waiting_data(c, s, a, b):
...
+
+
+@gen_cluster(client=True)
+def test_get_futures_error(c, s, a, b):
+    def dec(x):
+        return x - 1
+
+    def makes_error(x):
+        return 1 / x
```
mrocklin: dec and div are already implemented from `utils_test`

distributed/tests/test_client.py
```diff
@@ -3751,6 +3751,117 @@ def test_dont_clear_waiting_data(c, s, a, b):
...
+    yield _wait(f)
+    assert f.status == 'error'
+
+    function, args, kwargs = yield c._recreate_error_locally(f)
+    assert function.__name__ == 'makes_error'
+    assert args ==  (0,)
+    with pytest.raises(ZeroDivisionError):
+        function(*args, **kwargs)
+
+    # with futures
```
mrocklin: Separate test?

distributed/tests/test_client.py
```diff
@@ -3751,6 +3751,117 @@ def test_dont_clear_waiting_data(c, s, a, b):
...
+        if x == 0 :
+            raise ValueError
+        return x
+    df2 = df.a.map(make_err)
+    f = c.compute(df2)
+    yield _wait(f)
+    function, args, kwargs = yield c._recreate_error_locally(f)
+    with pytest.raises(ValueError):
+        function(*args, **kwargs)
+
```
mrocklin: It would also be good to test the synchronous API

martindurant:
I am unclear on how this would be folded into `gather`, since the exception handling actually happens in `_gather` (i.e., asynchronously), but the execution of the function should be synced; I could imagine passing `errors='skip'` into `_gather` (to get back the list of futures and know which one failed), but then all futures will be evaluated just for the sake of debugging one failed one. 
I think I'm happy to not to add this feature, seeing it as necessarily interactive functionality.

mrocklin:
OK

distributed/recreate_exceptions.py
```diff
@@ -91,14 +91,25 @@ def get_futures_error(self, future):
         """
         Ask the scheduler details of the sub-task of the given failed future
 
+        When a future evaluates to a status of "error", i.e., an exception
+        was raised in a task within its graph, we an get information from
+        the scheduler. This function gets the details of the specific task
+        that raised the exception and led to the error, but does not fetch
+        data from the cluster or execute the function. To actually run the
+        function locally and be able to debug, see
+        ``client.recreate_error_locally``.
```
mrocklin: If you add this in a `See Also` section then you'll get an actual link in the API docs (this is worth checking out if you haven't already)

distributed/recreate_exceptions.py
```diff
@@ -116,6 +127,14 @@ def recreate_error_locally(self, future):
         """
         For a failed calculation, perform the blamed task locally for debugging.
 
+        This operation should be performed after a future (result of ``gather``,
+        ``compute``, etc) comes back with a status of "error", if the stack-
+        trace is not informative enough to diagnose the problem. The specific
+        task (part of the graph pointing to the future) responsible for the
+        error will be fetched from the scheduler, together with the values of
+        it's inputs. The function will then be executed, so that ``pdb`` can
+        be used for debugging.
```
mrocklin: Maybe a quick `Examples` section?

docs/source/api.rst
```diff
@@ -36,6 +34,15 @@ API
    Client.upload_file
    Client.who_has
 
+.. currentmodule:: distributed.recreate_exceptions
```
martindurant: @mrocklin This makes the two methods appear below the others, which isn't optimal, but probably OK.

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,171 @@
...
+        Examples
+        --------
+        >>> x0 = delayed(dec)(2)         # doctest: +SKIP
+        >>> y0 = delayed(dec)(1)         # doctest: +SKIP
+        >>> x = delayed(div)(1, x0)      # doctest: +SKIP
+        >>> y = delayed(div)(1, y0)      # doctest: +SKIP
+        >>> tot = delayed(sum)(x, y)     # doctest: +SKIP
+        >>> f = c.compute(tot)           # doctest: +SKIP
+        >>> wait(f)                      # doctest: +SKIP
+        >>> #  f's status becomes "error"
```
mrocklin: Perhaps the following:

```python
>>> future = c.submit(div, 1, 0)
>>> future.status
'error'
>>> c.recreate_error_locally(future)

If you're in IPython you might take this opportunity to use pdb

>>> %pdb
Automatic pdb calling has been turned ON

>>> c.recreate_error_locally(future)
      1 def div(x, y):
----> 2     return x / y
ipdb> 
```

Themes of changes:

1.  We start with the simplest possible example, rather than expecting users to read much
2.  We use input/output pairs rather than comments (such as with `>>> future.status` and `ZeroDivisionError`)
3.  I've never referred to pdb in an example before, but in this case it seems like the main point.

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,171 @@
...
+        >>> c.recreate_error_locally(f)  # doctest: +SKIP
+        >>> #  a real ZeroDivisionError occurs in `div` with inputs (1, 0)
+
+
```
mrocklin: Single line break rather than double

distributed/recreate_exceptions.py
```diff
@@ -0,0 +1,171 @@
...
+        Parameters
+        ----------
+        future : future that failed
```
mrocklin: Future or collection?

distributed/tests/test_client.py
```diff
@@ -3751,6 +3751,120 @@ def test_dont_clear_waiting_data(c, s, a, b):
...
+        with Client(s['address'], loop=loop) as c:
+            x0 = c.submit(dec, 2)
+            y0 = c.submit(dec, 1)
+            x = c.submit(div, 1, x0)
+            y = c.submit(div, 1, y0)
+            tot = c.submit(sum, x, y)
+            f = c.compute(tot)
+
+            wait(f)
+            assert f.status == 'error'
```
mrocklin: What happens if we don't have the wait?  Do things still work?

docs/source/api.rst
mrocklin: Agreed

distributed/tests/test_client.py
martindurant: Quite right, it breaks; it should *not* work with no errored futures yet, but it should return an appropriate message, I suppose.

mrocklin: We could also choose to wait on the input

martindurant: True; also, should it be a ValueError if the user tries this on a future/set of futures that contain no error?

mrocklin: I don't have a strong opinion here.  Given that normal behavior is to receive an error it might be even be a stronger signal if no error occurs.

mrocklin:
This looks good to me.  Merging soon if there are no other comments.

mrocklin:
Merged.  Thanks @martindurant , I think that a number of users will really enjoy this feature.  It has often been requested.

PR #896, opened 2017-02-23
Improve address handling abstraction

pitrou:
This removes some of the roadblocks for using non-IP based protocols for local communication (notably, the notion of a "port" does not exist with `inproc` communications).

Based on PR #888.

### DIFF
--- a/distributed/bokeh/background/server_lifecycle.py
+++ b/distributed/bokeh/background/server_lifecycle.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
 from __future__ import print_function, division, absolute_import
 
-from collections import deque
 import json
 import logging
-import os
 import sys
 
 from tornado import gen
@@ -51,6 +49,8 @@ def http_get(route):
 
 
 last_index = [0]
+
+
 @gen.coroutine
 def workers():
     """ Get data from JSON route, store in messages deques """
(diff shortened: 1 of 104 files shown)


distributed/bokeh/background/server_lifecycle.py
```diff
@@ -124,11 +126,6 @@ def task_events(interval, deque, times, index, rectangles, workers, last_seen):
         pass  # don't log CommClosedErrors
     except Exception as e:
         logger.exception(e)
-    finally:
```
pitrou: Note this block was a no-op, as `sys.exit` always raises SystemExit.

mrocklin:
My first impression is that things seem clean, but also that the class hierarchy around comms is growing significantly more complex.  I think that now would be a good time to build real documentation around it so that others can easily understand this code after reading sufficient prose.

pitrou:
Thank you, good point about the documentation. i'll add some to make the architecture more understandable.

pitrou:
I think I've improved the documentation now. I'm not sure how informative that is, so please tell me if you think something is missing.

mrocklin:
cc @jcrist , who has expressed an interest to learn more about dask/distributed communication

docs/source/communications.rst
```diff
@@ -0,0 +1,98 @@
...
+Communications
+==============
+
+Workers, the Scheduler, and Clients communicate by sending each other
+Python objects (such as :ref:`protocol` messages or user data).
+The communication layer handlers appropriate encoding and shipping
+of those Python objects between the distributed endpoints.  The
+communication layer is able to select between different transport
+implementations, depending on user choise or (possibly) internal
+optimizations.
```
mrocklin: handlers -> handles

choise -> choice

pitrou:
@jcrist, do you want to take a look soon or should I merge?

mrocklin:
I think that @jcrist is busy today.  I recommend merging.

PR #898, opened 2017-02-23
Display deserialization time in task stream

mrocklin:
cc @stuartarchibald


stuartarchibald:
Thanks @mrocklin this will be helpful. Patch looks good.

PR #908, opened 2017-02-27
Client._gather will now reraise original Attribute and Key Errors

bschreck:
Before, if a task had an AttributeError or KeyError, the except block in reraise would catch these-- that except block is only meant for the FutureState object not having exception or traceback attribute, or the key no longer being inside self.futures.


mrocklin:
That looks like a good change to me.  Sorry for the Python 3 failures in CI.  Those are unrelated and will be cleaned up shortly.  

Are you able to set up a small test to ensure that this change doesn't revert by accident in the future?

mrocklin:
@bschreck any interest in adding a small test to this PR?

bschreck:
Yes I will- sorry for the delay, been busy this week

mrocklin:
I can't very well fault you if you're working for free :)

bschreck:
How does that test look?

mrocklin:
This looks good to me.  Thanks for following up @bschreck

PR #909, opened 2017-02-28
TST: Skip more bokeh-requiring tests if unavailable.

QuLogic:
Tests are currently a bit divided on this; some parts skip if bokeh is not available and some don't. Sometimes this is even inconsistent within the same file. With this change, bokeh is now completely optional for tests. It would be nice if CI checked this though, but I don't know if that change should be made here.


mrocklin:
Thanks for this.  The CI failures are unrelated and will be cleaned up shortly.

PR #911, opened 2017-02-28
Add mytestegg to manifest.

QuLogic:
Not including this file means tests break when installed from PyPI.


mrocklin:
Merged.  Thanks @QuLogic !

Issue #913, opened 2017-02-28
Please disable forkserver for py3 by config flag


tumb1er:
After #687 distributed client is not starting anymore from celery task.
That's because celery/billiard is patching multiprocessing module with it's own implementation.
* distributed==1.14.3 works correctly
* distributed>=1.15.0,<=1.15.2 freezes completely
* distributed==1.16.0 crashes with:
```python
# lots of stacks skipped
  File "/data/quokka/virtualenv/lib/python3.5/site-packages/tornado/gen.py", line 285, in wrapper
    yielded = next(result)
  File "/data/quokka/virtualenv/lib/python3.5/site-packages/distributed/nanny.py", line 207, in instantiate
    self.process.start()
  File "/usr/local/lib/python3.5/multiprocessing/process.py", line 105, in start
    self._popen = self._Popen(self)
  File "/usr/local/lib/python3.5/multiprocessing/context.py", line 281, in _Popen
    return Popen(process_obj)
  File "/usr/local/lib/python3.5/multiprocessing/popen_forkserver.py", line 36, in __init__
    super().__init__(process_obj)
  File "/usr/local/lib/python3.5/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/usr/local/lib/python3.5/multiprocessing/popen_forkserver.py", line 47, in _launch
    reduction.dump(prep_data, buf)
  File "/usr/local/lib/python3.5/multiprocessing/reduction.py", line 59, in dump
    ForkingPickler(file, protocol).dump(obj)
  File "/data/quokka/virtualenv/lib/python3.5/site-packages/billiard/process.py", line 330, in __reduce__
    'Pickling an AuthenticationString object is '
TypeError: Pickling an AuthenticationString object is disallowed for security reasons
```

So, flag for forkserver is necessary.

how to reproduce:
1. define celery task 
```python

@celery.task
def some_task():
   c = distributed.Client()
```
2. start celery worker
3. send a task
```python
some_task.delay()
```

mrocklin:
Should be resolved in #914

PR #914, opened 2017-02-28
Add forkserver as configuration option

mrocklin:
Fixes https://github.com/dask/distributed/issues/913

cc @pitrou for review (should be fast)


pitrou:
No problem from me.

PR #918, opened 2017-03-01
More tests for poor serialization

mrocklin:


distributed/tests/test_client.py
```diff
@@ -3770,3 +3770,62 @@ def test_retire_workers(c, s, a, b):
     while a.status != 'closed':
         yield gen.sleep(0.01)
         assert time() < start + 5
+
+
+@gen_cluster(client=True)
+def test_robust_unserializable(c, s, a, b):
+    class Foo(object):
+        def __getstate__(self):
+            raise Exception('hello')
```
pitrou: IMHO, a robustness test would be more thorough if raising (and checking for) a specific exception class, perhaps even specific to this test.

mrocklin: Fixed

PR #919, opened 2017-03-01
Use inproc transport in LocalCluster

pitrou:
If the parameters allow (i.e. no explicit scheduler port is asked for, and nanny is false), use inproc for distributed comms inside a LocalCluster, otherwise use tcp as usual.

Based on PR #896.

### DIFF
--- a/distributed/comm/addressing.py
+++ b/distributed/comm/addressing.py
@@ -1,0 +1,115 @@
+from __future__ import print_function, division, absolute_import
+
+import six
+
+from ..config import config
+from ..utils import ensure_ip
+
+
+DEFAULT_SCHEME = config.get('default-scheme', 'tcp')
+
+
+def parse_address(addr):
+    """
+    Split address into its scheme and scheme-dependent location string.
+    """
+    if not isinstance(addr, six.string_types):
+        raise TypeError("expected str, got %r" % addr.__class__.__name__)
+    scheme, sep, loc = addr.rpartition('://')
+    if not sep:
+        scheme = DEFAULT_SCHEME
+    return scheme, loc
+
+
+def unparse_address(scheme, loc):
+    """
+    Undo parse_address().
+    """
+    return '%s://%s' % (scheme, loc)
+
+
+def normalize_address(addr):
+    """
+    Canonicalize address, adding a default scheme if necessary.
+    """
+    return unparse_address(*parse_address(addr))
+
+
+def parse_host_port(address, default_port=None):
+    """
+    Parse an endpoint address given in the form "host:port".
+    """
+    if isinstance(address, tuple):
+        return address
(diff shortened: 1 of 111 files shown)


mrocklin:
@shoyer this may affect XArray users positively.  In particular I think it would give single-machine XArray users access to the distributed scheduler without moving data between client and workers.  This would increase complexity, but would give access to newer diagnostics.  It would not significantly increase communication costs.

distributed/deploy/tests/test_local.py
```diff
@@ -46,11 +47,36 @@ def test_procs(loop):
...
 
+def test_transports():
+    """
+    Test the transport chosen by LocalCluster depending on arguments.
+    """
+    with LocalCluster(1, nanny=False, silence_logs=False) as c:
+        assert c.scheduler_address.startswith('inproc://')
+        assert c.workers[0].address.startswith('inproc://')
+        with Client(c.scheduler.address) as e:
+            assert e.submit(inc, 4).result() == 5
```
mrocklin: Sorry for the delay in reviewing.  I thought I had pressed the "submit review" button hours ago...

distributed/comm/addressing.py
```diff
@@ -119,15 +119,24 @@ def get_address_host(addr):
...
     return backend.get_address_host(loc)
 
 
+def get_local_address_for(addr):
+    """
+    Get the local listening address suitable for reaching *addr*.
+
+    For instance, trying to reach an external TCP address will return
+    a local TCP address that's routable to that external address.
+    """
```
pitrou: I'm not very happy with the wording either... It's a generic equivalent of `get_ip()`, roughly:
```
>>> get_local_address_for('tcp://8.8.8.8:1234')
'tcp://192.168.1.68'
>>> get_local_address_for('tcp://127.0.0.1:1234')
'tcp://127.0.0.1'
```

This won't ever try to choose another address or change transports. When we want to do that, we'll have to define another API.

distributed/deploy/tests/test_local.py
pitrou: Ok, will do.

mrocklin: See my pull request to your fork

distributed/comm/addressing.py
pitrou: @mrocklin, do you have any suggestion for a better wording?

mrocklin: I'm not sure I have the expertise in networking to provide an accurate definition.  The examples you've provided would probably help.  

Generally this might be a nice pattern to repeat throughout `distributed/comm/addressing.py`.  When I look at this file I find myself daunted by the variety of functions here that do slightly different things.

mrocklin:
+1 from me.  Merging this soon if tests pass.

PR #920, opened 2017-03-01
Add flake8 to travis.yml

mrocklin:
cc @mariusvniekerk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ install:
   # Install dependencies
   - conda create -q -n test-environment python=$PYTHON
   - source activate test-environment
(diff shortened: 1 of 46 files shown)


.travis.yml
```diff
@@ -85,6 +85,7 @@ script:
       else
         py.test -m "not avoid_travis" distributed $PYTEST_OPTIONS;
       fi;
+    - flake8 distributed/*.py distributed/{bokeh,protocol,deploy}/*.py # no tests yet
```
pitrou: Is there a way to have it run over everything *except* tests, instead of manually listing subpackages?

mrocklin: Probably.  I also think that we should eventually include tests.  I think that we should aim to change this to `flake8 distributed` in the near future.  I wanted to finish what I had done so far and emit a PR though (I probably won't work on this much more today and don't want a PR like this growing stale (which will happen quickly))

jcrist: Yes. You can just specify `flake8 --exclude tests distributed`. Can also add this to the `setup.cfg` as:

```
[flake8]
exclude = tests
```

Issue #921, opened 2017-03-01
Log and display cluster events


mrocklin:
We maintain structured logs about various fine grained events on each task in `Scheduler.transition_log`,  `Worker.log`, `Stealing.log`, etc..  these are invaluable both in tracking down issues with methods like `Scheduler.transition_story(*keys)` and `Worker.story(*keys)`, and in producing visuals.

As people use clusters more dynamically we may find value in recording events around workers, clients, consistency checks, etc..  Ideally we find a consistent way to do this with many kinds of events that makes creating plots or query functions easy in the future.

So some questions:

1.  What are good events to track?
2.  What are good consistency checks to track?
3.  How do we organize these events for easy querying and visualization?
4.  What visuals and queries would be of pragmatic use?

PR #922, opened 2017-03-01
Add Event logs

mrocklin:
This provides data structures for us to add more structured logging that we can more easily query and visualize.

See https://github.com/dask/distributed/issues/921


mrocklin:
@pitrou can I ask you to review the events logging here?  (you can safely ignore the bokeh plots, which are experimental at this point)

In particular do you have thoughts on

1.  Better ways to structure logs (currently a dict of deques)
2.  Other events we should be logging

PR #931, opened 2017-03-03
Pandas serialization

mrocklin:
This implements a custom (and more efficient) serializer for Pandas dataframes and series.  It works by calling `serialize` on each series's values attribute and then merging the results.  This redirects either to numpy (common case) or pickle (in cases like categoricals) so we can get benefits in the common case without covering the entire pandas abstraction.

I'm not sure that we should merge this; I suspect that there are quite a few corner cases that it does not handle.  It's useful to have it up here for benchmarking and review purposes.

### DIFF
--- a/distributed/protocol/pandas.py
+++ b/distributed/protocol/pandas.py
@@ -1,0 +1,100 @@
+from __future__ import print_function, division, absolute_import
+
+import pickle
+import sys
+
+import pandas as pd
+
+from .serialize import register_serialization, serialize, deserialize
+
+
+def serialize_pandas_dataframe(df):
+    head = pickle.dumps(df.head(0))
+    headers = []
+    framess = []
+    compression = [None]
+    lengths = [len(head)]
+    for column in df.columns:
+        x = df[column].values
+        header, frames = serialize(x)
+        headers.append(header)
+        framess.append(frames)
+        compression.extend(header.pop('compression', [None] * len(frames)))
+        lengths.extend(header.pop('lengths', [len(f) for f in frames]))
+
+
+    # TODO: avoid if trivial index
+    x = df.index.values
+    index_header, index_frames = serialize(x)
+    compression.extend(index_header.pop('compression',
+                                        [None] * len(index_frames)))
+    lengths.extend(index_header.pop('lengths', [len(f) for f in index_frames]))
+
+    framess.append(index_frames)
+
+    header = {'frame-counts': [len(f) for f in framess],
+              'headers': headers,
+              'index-header': index_header,
+              'lengths': lengths}
+
+    if any(compression):
+        header['compression'] = compression
+
+    return header, [head] + sum(framess, [])
+
+
+def deserialize_pandas_dataframe(header, frames):
+    head = pickle.loads(frames[0])
+    n = 1
+    d = {}
+    for column, h, count in zip(head.columns, header['headers'], header['frame-counts']):
+        x = deserialize(h, frames[n:n + count])
+        n += count
+        d[column] = x
+
+    index = deserialize(header['index-header'], frames[n:])
+
+    df = pd.DataFrame(d, columns=head.columns, index=index)
+    df.index.name = head.index.name
+    return df
+
+
+def serialize_pandas_series(s):
+    value_header, value_frames = serialize(s.values)
+    index_header, index_frames = serialize(s.index)
+
+    compression = []
+    lengths = []
+    for h, f in [(value_header, value_frames), (index_header, index_frames)]:
+        compression.extend(h.pop('compression', [None] * len(f)))
+        lengths.extend(h.pop('lengths', [len(ff) for ff in f]))
+
+    header = {'name': s.name,
+              'value-header': value_header,
+              'index-header': index_header,
+              'n_value_frames': len(value_frames),
+              'index-name': s.index.name,
+              'lengths': lengths}
+
+    if any(compression):
+        header['compression'] = compression
+
+    return header, value_frames + index_frames
+
+
+def deserialize_pandas_series(header, frames):
+    values = deserialize(header['value-header'], frames[:header['n_value_frames']])
+    index = deserialize(header['index-header'], frames[header['n_value_frames']:])
+
+    return pd.Series(values,
+                     name=header['name'],
+                     index=pd.Index(index, name=header['index-name']))
+
+
+register_serialization(pd.DataFrame,
+                       serialize_pandas_dataframe,
+                       deserialize_pandas_dataframe)
+
+register_serialization(pd.Series,
+                       serialize_pandas_series,
+                       deserialize_pandas_series)
--- a/distributed/protocol/tests/test_pandas.py
+++ b/distributed/protocol/tests/test_pandas.py
@@ -2,6 +2,7 @@ from __future__ import print_function, division, absolute_import
 
 from zlib import crc32
 
+import numpy as np
 import pandas as pd
 import pandas.util.testing as tm
 import pytest
@@ -32,13 +33,16 @@ dfs = [
     pd.DataFrame({'x': [b'a', b'b', b'c']}),
     pd.DataFrame({'x': pd.Categorical(['a', 'b', 'a'], ordered=True)}),
     pd.DataFrame({'x': pd.Categorical(['a', 'b', 'a'], ordered=False)}),
+    pd.Series(np.arange(10000000)),
+    pd.DataFrame({'x': np.arange(10000000)}),
     tm.makeCategoricalIndex(),
     tm.makeCustomDataframe(5, 3),
     tm.makeDataFrame(),
     tm.makeDateIndex(),
     tm.makeMissingDataframe(),
     tm.makeMixedDataFrame(),
-    tm.makeObjectSeries(),
+    pytest.mark.xfail(tm.makeObjectSeries(),
+                      reason='date to timestamp conversion'),
     tm.makePeriodFrame(),
     tm.makeRangeIndex(),
     tm.makeTimeDataFrame(),
@@ -48,10 +52,28 @@ dfs = [
 
 
 @pytest.mark.parametrize('df', dfs)
-def test_dumps_serialize_numpy(df):
+def test_serialize_pandas(df):
     header, frames = serialize(df)
     if 'compression' in header:
         frames = decompress(header, frames)
     df2 = deserialize(header, frames)
 
-    assert_eq(df, df2)
+    if isinstance(df, pd.DataFrame):
+        tm.assert_frame_equal(df, df2)
+    elif isinstance(df, pd.Series):
+        tm.assert_series_equal(df, df2)
+    else:
+        assert_eq(df, df2)
+
+
+@pytest.mark.parametrize('df', dfs)
+def test_dumps_pandas(df):
+    frames = dumps({'x': to_serialize(df)})
+    df2 = loads(frames)['x']
+
+    if isinstance(df, pd.DataFrame):
+        tm.assert_frame_equal(df, df2)
+    elif isinstance(df, pd.Series):
+        tm.assert_series_equal(df, df2)
+    else:
+        assert_eq(df, df2)
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -15,6 +15,10 @@ from ..utils  import ignoring
 def _register_numpy():
     from . import numpy
 
+@partial(register_serialization_lazy, "pandas")
+def _register_pandas():
+    from . import pandas
+
 @partial(register_serialization_lazy, "h5py")
 def _register_h5py():
     from . import h5py


distributed/protocol/pandas.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def serialize_pandas_dataframe(df):
+    head = pickle.dumps(df.head(0))
+    headers = []
+    framess = []
+    compression = [None]
+    lengths = [len(head)]
+    for column in df.columns:
+        x = df[column].values
```
jreback: using ``.values`` will cause coercion to numpy datatypes for most things (except categoricals), IOW, datetimes w/tz are converted to UTC (and thus this is not idempotent). Better to use ``df[column]._values`` which preserves the structure.

distributed/protocol/tests/test_pandas.py
```diff
@@ -32,13 +33,16 @@
     pd.DataFrame({'x': [b'a', b'b', b'c']}),
     pd.DataFrame({'x': pd.Categorical(['a', 'b', 'a'], ordered=True)}),
     pd.DataFrame({'x': pd.Categorical(['a', 'b', 'a'], ordered=False)}),
+    pd.Series(np.arange(10000000)),
```
jreback: should test various dtypes (int16, float32), datetimes, timedelta, period (``pd.period_range``), datetime w/tz (pd.date_range(...., tz='US/Eastern')

jreback:
```
In [12]: df = tm.makeMixedDataFrame().assign(E=lambda df: df.D.dt.tz_localize('US/Eastern'), B=lambda df: df.B.astype(int), F=lambda df: df.C.astype('category'))

In [13]: df
Out[13]: 
     A  B     C          D                         E     F
0  0.0  0  foo1 2009-01-01 2009-01-01 00:00:00-05:00  foo1
1  1.0  1  foo2 2009-01-02 2009-01-02 00:00:00-05:00  foo2
2  2.0  0  foo3 2009-01-05 2009-01-05 00:00:00-05:00  foo3
3  3.0  1  foo4 2009-01-06 2009-01-06 00:00:00-05:00  foo4
4  4.0  0  foo5 2009-01-07 2009-01-07 00:00:00-05:00  foo5

In [14]: df.dtypes
Out[14]: 
A                       float64
B                         int64
C                        object
D                datetime64[ns]
E    datetime64[ns, US/Eastern]
F                      category
dtype: object

In [15]: df.to_dict(orient='series')
Out[15]: 
{'A': 0    0.0
 1    1.0
 2    2.0
 3    3.0
 4    4.0
 Name: A, dtype: float64, 'B': 0    0
 1    1
 2    0
 3    1
 4    0
 Name: B, dtype: int64, 'C': 0    foo1
 1    foo2
 2    foo3
 3    foo4
 4    foo5
 Name: C, dtype: object, 'D': 0   2009-01-01
 1   2009-01-02
 2   2009-01-05
 3   2009-01-06
 4   2009-01-07
 Name: D, dtype: datetime64[ns], 'E': 0   2009-01-01 00:00:00-05:00
 1   2009-01-02 00:00:00-05:00
 2   2009-01-05 00:00:00-05:00
 3   2009-01-06 00:00:00-05:00
 4   2009-01-07 00:00:00-05:00
 Name: E, dtype: datetime64[ns, US/Eastern], 'F': 0    foo1
 1    foo2
 2    foo3
 3    foo4
 4    foo5
 Name: F, dtype: category
 Categories (5, object): [foo1, foo2, foo3, foo4, foo5]}
```

mrocklin:
Closing this for now.  This isn't a high priority and we're waiting for Arrow/Pandas to handle this on their end.

ogrisel:
@mrocklin the fact that `worker.data.slow` is falling back to `cloudpickle.dumps` is causing large memory usage spiked when evicting data when working with partitions that are on the order of 1GB.

It would be great to have a no-copy serialization mechanism, at least for spilling data to disk on workers. This PR looked promising as it could access the underlying numpy data blocks as sliceable memory views and therefore pass those as slices of memoryview to msgpack to do no-copy serialization to disk.

I am not sure how pyarrow could allow some no-copy view on pandas 0.21 dataframes. It seems that `pyarrow.Table.from_pandas(df)` makes at least one memory copy (based on some `psutil` experiments).

mrocklin:
It is possible to resurrect this work.  It was put on hold partially because there wasn't a motivating use case.  It sounds like now we do have a motivating use case.

I would be more comfortable with this direction if we were to whitelist when it can be used.  For example I expect it to be error prone when dealing with categoricals, timestamps with timezones, and so on.  

> I am not sure how pyarrow could allow some no-copy view on pandas 0.21 dataframes. It seems that `pyarrow.Table.from_pandas(df)` makes at least one memory copy (based on some psutil experiments).
 Delete branch

@wesm @xhochy do either of you have thoughts on this?  My understanding here is the same as @ogrisel 's, that today Arrow is probably not a suitable mechanism for a pandas -> list of memoryviews conversion in a zero-copy manner, and that this is probably out of scope near term.  Is this understanding accurate?

mrocklin:
Also @xhochy I think that @ogrisel is trying to resolve problems related to what @bluenote10 ran into regarding memory use on Pandas dataframes.

xhochy:
Zero-copy serialization of Pandas DataFrames to Arrow will only be able for a very narrow scope:
 * integer columns: ✅ 
 * float columns without nulls (or treating nan as non-null in the case of nans): ✅ 
 * string columns: ❌ 
 * categorical columns: ❌ 
 * boolean columns: ❌ 
 * datetime columns: ✅  (not sure if we do this atm)
 * object columns: ❌ 

We can serialise all these types of columns, some can also be reconstructed without any loss of information/type but zero-copy is quite hard as there are some minor differences in the memory layout of Arrow and Pandas.

wesm:
I'm open to building something outside the box to help here. Per http://arrow.apache.org/blog/2017/10/15/fast-python-serialization-with-ray-and-arrow/, we can serialize general Python lists and dicts containing NumPy arrays. So without _too_ much effort, we could disassemble the BlockManager into a dictionary using a custom serializer and then do a zero-copy reconstruction from the resulting buffer.

wesm:
I can prototype something and put up a patch if that sounds good?

mrocklin:
I would be very happy to have others deal with this problem :)

To be clear on our objectives, we're looking to avoid any significant memory allocation.  My understanding of the blogpost above is that it is still about turning Python objects into bytes, which we don't want to do (except for metadata). 

> and then do a zero-copy reconstruction from the resulting buffer.

Again, to be clear, it probably isn't feasible to construct or consume a single buffer here (this would, I think, imply concatenation and so memory allocation).  Instead I would assume that we would need to pass around a sequence of bytes-like objects where those bytes-like objects were directly taken from the blocks.

wesm:
Hopefully this helps explain what is going on:

```
In [1]: import pyarrow as pa

In [2]: import numpy as np

In [3]: data = {i: np.random.randn(1000, 1000) for i in range(10)}

In [4]: prepped = pa.serialize(data)

In [5]: %timeit prepped = pa.serialize(data)
27.8 µs ± 3.26 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [6]: as_buf = prepped.to_buffer()

In [7]: %timeit as_buf = prepped.to_buffer()
11.5 ms ± 89.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [8]: data_deserialized = pa.deserialize(as_buf)

In [9]: %timeit data_deserialized = pa.deserialize(as_buf)
34.8 µs ± 227 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

The operation `pyarrow.serialize` is a zero copy operation, unless you have `PyObject*` arrays. The act `to_buffer` creates a single memoryview-compatible object containing the whole payload. You can also write the stream to a file-like object, too

The action `pa.deserialize` is zero-copy (assuming that you don't have any object arrays, which require unpickling), returning the original data, but containing memory references to the input buffer

So what I'm proposing is to decompose a pandas DataFrame into a collection of NumPy arrays representing the BlockManager, and then pushing them through this zero-copy serialization machinery

mrocklin:
Dask's communication machinery deals with bytes or memoryview objects, it doesn't use file-like objects.  Is it possible to pass through the memoryview object without an allocation?

For example in NumPy arrays we do something like the following:

```python
def serialize_numpy(x):
    header = {'shape': x.shape, 'dtype': x.dtype, 'strides': x.strides}
    frames = [x.data]
    return header, frames
```

This PR did more or less the same thing, except that the header was much more complex and the list of frames was much longer (probably one per block or one per column).

wesm:
That would be a little bit more work, but it's certainly possible and would be valuable. The purpose of ARROW-1509 https://issues.apache.org/jira/browse/ARROW-1509 is to make the framing of the serialized payload less monolithic in the way you describe

cc @pcmoritz @robertnishihara @cpcloud

pitrou:
> The action to_buffer creates a single memoryview-compatible object containing the whole payload. 

Which is a copy, right? Can you instead grow a `to_buffers()` method which returns a list of buffers without copying data?

mrocklin:
If Arrow is able to efficiently convert Python things into header + list of bytes/memoryviews without allocating much, and then take that same representation and convert back to Python things then it would be easy to integrate into Dask's communication machinery (and presumably, anything that relies on sockets/tornado/asyncio)

We have an immediate need for this for Pandas dataframes in particular due to memory issues when dealing with large dataframes in small containers.  Do you have a sense for when this work might start/complete?

wesm:
> Which is a copy, right? Can you instead grow a to_buffers() method which returns a list of buffers without copying data?

yes, this is exactly something that ARROW-1509 will permit quite easily

In theory this is all about a day or so of work for someone that knows what they are doing. I will create a JIRA to scope this for 0.8.0, so I think it can be completed in the next 2 weeks -- whenever it's merged it will appear in Linux nightlies (https://anaconda.org/twosigma/pyarrow) so it can be validated more easily by others without having to set up the build environment. 

I can do the development work myself if no one else steps up to take a look, doesn't seem like it will be that difficult.

mrocklin:
Awesome.  I'll be quite glad if this can help us resolve our memory issues.  Thanks for engaging here @wesm

wesm:
I have some pressing stuff to wrap up this week so the earliest I could take a look would be early next week. Will keep you in the loop

wesm:
Tracking JIRAs

* https://issues.apache.org/jira/browse/ARROW-1783
* https://issues.apache.org/jira/browse/ARROW-1784

wesm:
I haven't gotten to this yet, but it's on deck for either sometime this weekend or first thing next week (Thanksgiving week being quiet will be good for writing a lot of patches)

mrocklin:
This would definitely be nice, but it isn't critical to Dask.  No major rush on our end.

ogrisel:
Just a heads up to let you know that I am working on a generic improvement in cloudpickle itself to implement memoryview-based nocopy semantics for large pydata objects: https://github.com/cloudpipe/cloudpickle/pull/138

This could be useful for other nested numpy arrays objects (e.g. scipy sparse matrices, big scikit-learn estimators such as random forests...).

wesm:
That's cool, this sounds like a very similar strategy to what we've been doing in the Arrow serializer

pitrou:
> Just a heads up to let you know that I am working on a generic improvement in cloudpickle itself to implement memoryview-based nocopy semantics for large pydata objects

This is a bit different though, as it only does "nocopy" pickling if you pickle to a file-like object (i.e. call `dump()` rather than `dumps()`). Which will be tricky if your file object is non-blocking :-)

This is an inherent limitation of pickle, as it outputs a single stream of bytes, while the pyarrow approach returns a list of buffers.

ogrisel:
> This is an inherent limitation of pickle, as it outputs a single stream of bytes, while the pyarrow approach returns a list of buffers.

One of the goals of my PR is to actually support the "list of buffers" patterns in cloudpickle. Check the use of `ChunksAppender` in the `test_nocopy_numpy_array` test:

https://github.com/cloudpipe/cloudpickle/pull/138/files#diff-ad0a048ef2f4c0dc3eaf2baa553b134eR980

The call to `materialize()` happens after the end of the pickling process and is meant to simulate the kind of async access to the buffers done by tornado in dask distributed.

This is actually the motivation to not call `.release()` in the frame buffer of the Python pickler in upstream CPython and use a new BytesIO instance for each protocol 4 frame instead.

pitrou:
> This is actually the motivation to not call .release() in the frame buffer of the Python pickler in upstream CPython and use a new BytesIO instance for each protocol 4 frame instead.

I'm worried about the implications in terms of buffer lifetime. This will need a bit more analysis IMHO.

ogrisel:
FYI I have updated cloudpipe/cloudpickle#138 and I can now get nocopy list-of-buffers semantics for pandas data frames, numpy arrays and scipy sparse matrices under CPython 3.5 and 3.6.

mrocklin:
This is really impressive.  Have you tried this with dask?  If so then are the issues you were seeing before about memory spikes when writing to disk being resolved?

ogrisel:
I have not yet tried with dask, I am battling with understanding the details of refcounting to find a way to ensure that my mutable bytes hack is safe.

But I am pretty sure that this new cloudpickle + the throttled GC will solve the memory usage spikes when workers spill to disk.

ogrisel:
I think I found a solution. I plan to run some tests with dask next week.

jakirkham:
Obviously this is an old thread and things have changed since then, but wanted to link this comment ( https://github.com/dask/distributed/issues/614#issuecomment-661617412 ) showing how zero-copy serialization with pickle protocol 5 efficiently handles Pandas serialization today. Though this is not unique to Pandas and can work with anything that supports pickle protocol 5

PR #935, opened 2017-03-08
Use Connection Pool for Inter-worker communication

mrocklin:
Previously we made new connections any time any worker wanted to request data
from another worker.  Now we use the pre-existing connection pool for this
purpose.  This lets us avoid reconnection costs.

We had this feature before but removed it due to errors that we couldn't track
down.  Given how much has changed since them I'm comfortable adding this back
in, but we should be aware of the potential problem.

### DIFF
--- a/distributed/diagnostics/progress_stream.py
+++ b/distributed/diagnostics/progress_stream.py
@@ -175,21 +175,18 @@ def color_of_message(msg):
 
 
 colors = {'transfer': 'red',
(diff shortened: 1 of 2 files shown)


mrocklin:
This will also probably affect SSL users after #866 by @mariusvniekerk

mrocklin:
Merging this soon if there are no objections

PR #936, opened 2017-03-09
Sizeof dataframe tweaks

mrocklin:
Two main changes:

1.  Sample the dataframe for objects when receiving an object dtype column.  Previously we assumed that these were text with length 100.  Now we select a sample of twenty elements and call sizeof on those.
2.  Reduce overhead from calling `DataFrame.memory_usage`

### DIFF
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -42,6 +42,8 @@ def register_pandas():
     import numpy as np
 
     def object_size(x):
+        if not len(x):
+            return 0
         sample = np.random.choice(x, size=20, replace=True)
         sample = list(map(sizeof, sample))
         return sum(sample) / 20 * len(x)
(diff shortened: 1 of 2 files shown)


distributed/sizeof.py
```diff
@@ -39,29 +39,40 @@ def sizeof_numpy_ndarray(x):
...
+        return sum(sample) / 20 * len(x)
 
     @sizeof.register(pd.DataFrame)
     def sizeof_pandas_dataframe(df):
-        p = int(df.memory_usage(index=True).sum())
-        obj = int((df.dtypes == object).sum() * len(df) * 100)
-        if df.index.dtype == object:
-            obj += len(df) * 100
-        return int(p + obj) + 1000
+        p = sum(c.memory_usage(index=False) for _, c in df.iteritems())
```
jreback: fyi using deep=True would be more accurate (though somewhat more expensive)

mrocklin: We removed this because it was too expensive.  We would like to perform this operation at well under a millisecond.

jreback: that's what i figured

PR #938, opened 2017-03-10
Add nanny's process to processes_to_close

mrocklin:
I've noticed that dask-workers hang around for a while if I create a
LocalCluster in a notebook and then restart the kernel.  This may help.

cc @pitrou for review


mrocklin:
Merging soon if there are no other comments.

pitrou:
This looks ok to me. Apparently `processes_to_close` wasn't ever populated before?

mrocklin:
Indeed

PR #939, opened 2017-03-10
Added directory upload support

jacksonloper:
This gives the `Client.upload_file` function the ability to upload directories.  It also tries to import the directory as a python package, and reload it if it is uploaded twice.  

If there is a good use-case for uploading directories that *aren't* python packages, we might not want to automatically try to import every directory that is uploaded.  But I can't think of such a use-case.  After all, if you're talking about uploading a large amount of data to disk, you absolutely do *not* want to use this, since it will duplicate the data for each worker (even two workers on the same node).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ services:
 
 env:
   matrix:
-    - PYTHON=2.7 PACKAGES="blosc futures faulthandler"
-    - PYTHON=3.5 COVERAGE=true PACKAGES=blosc CRICK=true
+    - PYTHON=2.7 PACKAGES="python-blosc futures faulthandler"
+    - PYTHON=3.5 COVERAGE=true PACKAGES=python-blosc CRICK=true
     - PYTHON=3.6
     - HDFS=true PYTHON=2.7
-    - HDFS=true PYTHON=3.5 PACKAGES=blosc
+    - HDFS=true PYTHON=3.5 PACKAGES=python-blosc
 
 matrix:
   fast_finish: true
--- a/distributed/protocol/tests/test_protocol.py
+++ b/distributed/protocol/tests/test_protocol.py
@@ -214,6 +214,8 @@ def test_dumps_loads_Serialized():
     assert result == result3
 
 
+@pytest.mark.skipif(sys.version_info[0] < 3,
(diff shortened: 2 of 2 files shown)


mrocklin:
I suspect that flake8 will complain at you a bit here.

Also, it would be very useful to include tests with this to ensure that all desired behaviors will remain in the future even if other people start screwing around with this code.

jacksonloper:
For tests, should I programmatically generate files and directories in a temporary folder and then upload them?  Or should I store a set of eggs and zips and folders and .py and .pyc files somewhere in this repo, and then upload those?  If the latter, what would be the best way to get access to those files from the test functions?

mrocklin:
Programmatically generating files is probably easier.  If you search `distributed/tests/test_client.py` for `upload_file` you'll see some examples that you can copy.  Most of our convenience functions for temporary files are just for files.  You might just make a temporary directory with the `tempfile` modules, do stuff, and then shutil.rmtree the directory afterwards

```python
dir = tempfile.mkdtemp()
# create files in os.path.join(dir, filename)
try:
    do tests
finally:
    shutil.rmtree(dir)
```

mrocklin:
@jacksonloper were you still interested in this or should I close?

jacksonloper:
Sorry... just finished phd defense (thanks to dask!), moving to new york
city...

I will get to it before december, but that is all I can say :).  Feel free
to close and I will submit another pr later with the tests included.

Best,
Jackson


On Jun 1, 2017 9:03 AM, "Matthew Rocklin" <notifications@github.com> wrote:

@jacksonloper <https://github.com/jacksonloper> were you still interested
in this or should I close?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/dask/distributed/pull/939#issuecomment-305486702>, or mute
the thread

mrocklin:
Congratulations!

jacksonloper:
I guess if you reset from upstream and `git push origin master --force` it closes all your PRs.  Learn something new every day...

mrocklin:
Heh, the thing to do here is probably not to develop on the master branch.  I always check out a new branch before starting work

    git checkout -b upload-directories

Issue #905, opened 2017-02-26
Use protocol serialization to store excess worker data to disk


mrocklin:
When workers run out of RAM they dump data to disk.  Currently they use pickle/cloudpickle and then do light compression.  We should change this so that we instead use the same serialization we use for our network protocol.  This has a few advantages:

1.  We can get added performance on types like NumPy arrays that we have special cased
2.  We can serialize data types that are Dask-serializable but not pickle/cloudpickle serializable, like Keras models, HDF5 files, etc.

The relevant functions here are distributed/protocol/serialize.py::serialize and distributed/worker.py::dumps_to_disk/loads_from_disk

PR #946, opened 2017-03-14
Use protocol for on-disk serialization

mrocklin:
Fixes https://github.com/dask/distributed/issues/905


distributed/protocol/serialize.py
```diff
@@ -8,7 +8,13 @@
 except ImportError:
     from toolz import valmap, get_in
 
+try:
+    import msgpack
+except ImportError:
+    import pandas.msgpack as msgpack
```
jreback: FYI this changes in master to ``pandas.io.msgpack`` (for >= 0.20.0)

PR #947, opened 2017-03-14
Perform message serialization in a Thread Pool

mrocklin:
This adds around 80us to every write, but avoids contention if
serialization is blocking and releases the GIL.

I don't think that this is worthwhile today.  I intend to close this immediately.  I just wanted to push it up here for future reference.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -1,5 +1,6 @@
 from __future__ import print_function, division, absolute_import
 
+from concurrent.futures import ThreadPoolExecutor
 from datetime import timedelta
 import errno
 import logging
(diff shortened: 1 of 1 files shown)


pitrou:
Interesting. I assume serialization is releasing the GIL mostly for compression? Perhaps deferring to a given thread should be done at that point whenever the data is big enough?

mrocklin:
Dask won't know if the data is big enough until after it has serialized.  We could send in a `big=True` flag from the worker when we know that this might be an issue.

Compression is one point, yes, so too is serialization sometimes, especially if we switch to something like Arrow for dataframes.

But I don't have any benchmarks that say that this is a good idea, so I suggest that we keep this here for now.

PR #952, opened 2017-03-16
Support async for protocol in as_completed

mrocklin:
This just defines `__aiter__` and `__anext__` protocols for `as_completed`.  

What is the right way to test Python-3 only code?  Can we put these in separate files that are ignored by pytest in Python 2?  I imagine that this is tricky because just parsing things will raise syntax errors.  cc @pitrou

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2573,10 +2573,6 @@ class AsCompleted(object):
             yield self.event.wait()
         return self.queue.get()
 
-
-
-
-
     next = __next__
 
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2571,6 +2571,7 @@ class AsCompleted(object):
             raise StopAsyncIteration  # flake8: noqa
         while self.queue.empty():
             yield self.event.wait()
+        sleep(0.0000000001)
         raise gen.Return(self.queue.get())
 
     next = __next__
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2523,6 +2523,7 @@ class AsCompleted(object):
         self.queue = pyQueue()
         self.lock = Lock()
         self.loop = loop or default_client().loop
+        self.event = Event()
 
         if futures:
             for future in futures:
@@ -2536,6 +2537,7 @@ class AsCompleted(object):
             if not self.futures[future]:
                 del self.futures[future]
             self.queue.put_nowait(future)
+            self.event.set()
 
     def add(self, future):
         """ Add a future to the collection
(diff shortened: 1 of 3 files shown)


pitrou:
I would suggest putting the Python 3-specific functions in a separate module (that's not named `test_xxx.py`) and then importing them from the actual tests only under Python 3.

mrocklin:
I'll give that a shot.  thanks.

mrocklin:
@pitrou any insight on what is causing the Windows failure?

pitrou:
I have no idea, sorry.

distributed/client.py
```diff
@@ -2571,6 +2571,7 @@ def __anext__(self):
             raise StopAsyncIteration  # flake8: noqa
         while self.queue.empty():
             yield self.event.wait()
+        sleep(0.0000000001)
```
mrocklin: @pitrou any thoughts on why this might be necessary on Windows but not Linux?

Adding a sleep of any positive number seemed to work.  Adding a sleep of 0 failed (thing hang).  Adding a print works.  Adding `1 + 1` hangs.

All of these fail on my local VM, so I don't think it's just an issue of CI services being slow enough to catch race conditions.

mrocklin:
I'm tempted to merge this even with the confusing nano-sleep in there.

distributed/client.py
pitrou: Did you get a sense of where it hangs?

pitrou: For the record, there are subtle differences between `yield`-based coroutines and `await`-based coroutines (the former will always yield to the scheduler while the second might not, which can create deadlocks). I'm not really an expert on `await`. I also don't know whether it's ok to make `__await__` a `yield`-based coroutine. Perhaps @bdarnell has more insight on this.

mrocklin: No.  You might be able to get something from [one of the failed tracebacks](https://ci.appveyor.com/project/daskdev/distributed/build/1.0.871) however

pitrou:
Looking at the tracebacks in https://ci.appveyor.com/project/daskdev/distributed/build/1.0.869#L1286 and https://ci.appveyor.com/project/daskdev/distributed/build/1.0.871#L1045, the main thread seems to be timing out in places which aren't blocking.

This seems to hint that perhaps `__aiter__` and `__anext__` protocol is implemented wrongly here. Here is Tornado does it for `WaitIterator` in `tornado/gen.py` (https://github.com/tornadoweb/tornado/blob/master/tornado/gen.py#L480-L488):

```
    @coroutine
    def __aiter__(self):
        raise Return(self)

    def __anext__(self):
        if self.done():
            # Lookup by name to silence pyflakes on older versions.
            raise getattr(builtins, 'StopAsyncIteration')()
        return self.next()
```

distributed/client.py
```diff
@@ -2551,12 +2556,24 @@ def add(self, future):
...
             if not self.futures and self.queue.empty():
                 raise StopIteration()
         return self.queue.get()
 
+    @gen.coroutine
+    def __anext__(self):
+        if not self.futures and self.queue.empty():
+            raise StopAsyncIteration  # flake8: noqa
+        while self.queue.empty():
+            yield self.event.wait()
```
bdarnell: This event is set once and never cleared, so once you've read one thing from the queue this turns into a busy loop. Maybe that's starving out some other thread that needs to run on windows? 

It looks like you want a `tornado.locks.Condition` instead.

mrocklin: Ah, indeed, silly me.  Fixed.  Thank you for catching my errors @bdarnell .

PR #953, opened 2017-03-16
Ensure that random generators are re-seeded in worker processes

pitrou:
Supersedes #835.

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -326,10 +326,14 @@ class Nanny(Server):
 def run_worker_fork(q, scheduler_addr, ncores, nanny_port,
                     worker_ip, worker_port, local_dir, **kwargs):
     """
-    Create a worker by forking.
+    Create a worker in a forked child.
     """
+    from dask.multiprocessing import initialize_worker_process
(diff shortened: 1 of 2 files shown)


pitrou:
This PR needs https://github.com/dask/dask/pull/2094 in Dask.

distributed/nanny.py
```diff
@@ -326,10 +326,14 @@ def monitor_resources(self, comm, interval=1):
...
                     worker_ip, worker_port, local_dir, **kwargs):
     """
-    Create a worker by forking.
+    Create a worker in a forked child.
     """
+    from dask.multiprocessing import initialize_worker_process
     from distributed import Worker  # pragma: no cover
     from tornado.ioloop import IOLoop  # pragma: no cover
+
+    initialize_worker_process()
```
mrocklin: I would like to guard against old versions of Dask here.  It turns out to be very useful to have some flexibility in version mismatch (it's nice to have distributed work with dask of one or two versions ago).  Can we put the import and this call in a try-except ImportError block?

pitrou: Of course, will do.

PR #965, opened 2017-03-23
Ensure that run can return non-msgpack serializable results

mrocklin:


distributed/tests/test_client.py
```diff
@@ -2195,6 +2195,23 @@ def test_run(c, s, a, b):
     assert results == {}
 
 
+@gen_cluster(client=True)
+def test_run_serialize(c, s, a, b):
```
pitrou: Perhaps add a comment and issue number explaining what the test is about?

distributed/tests/test_client.py
```diff
@@ -2195,6 +2195,23 @@ def test_run(c, s, a, b):
...
+    futures = c.map(inc, range(10))
+    yield _wait(futures)
+
+    func = lambda dask_scheduler=None: dask_scheduler.who_has
```
pitrou: Is it required for `dask_scheduler` to be an optional argument?

mrocklin: Apparently no.  Fixed

distributed/tests/test_client.py
mrocklin: I've expanded the title to be more informative and changed the code to be, I think, more self explanatory.  There was no related github issue.  The problem was reported in private conversation.

pitrou:
Great, thank you.

PR #971, opened 2017-03-26
Return both future and result in as_completed

mrocklin:
Many as_completed workflows look like this:

```python
ac = as_completed(futures)
for future in ac:
    result = future.result()
    ...
    new_future = client.submit(func, *args, **kwargs)
    ac.add(new_future)
```

Adding the blocking `.result()` call into this loop can significantly impact responsiveness.  It is sometimes useful to gather asynchronously in the background and only yield the result as it is gathered.

```python
ac = as_completed(futures, results=True)
for future, result in ac:
    # result = future.result()
    ...
    new_future = client.submit(func, *args, **kwargs)
    ac.add(new_future)
```

Any thoughts or suggestions?  This is a deviation away from the concurrent.futures API.

This affects @moody-marlin .  @pitrou may have thoughts on whether or not this API is appropriate.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2528,7 +2528,7 @@ class AsCompleted(object):
     4
     3
     """
-    def __init__(self, futures=None, loop=None, results=False):
+    def __init__(self, futures=None, loop=None, with_results=False):
         if futures is None:
             futures = []
         self.futures = defaultdict(lambda: 0)
@@ -2536,7 +2536,7 @@ class AsCompleted(object):
         self.lock = Lock()
         self.loop = loop or default_client().loop
         self.condition = Condition()
-        self.results = results
+        self.with_results = with_results
 
         if futures:
             for future in futures:
@@ -2545,13 +2545,13 @@ class AsCompleted(object):
     @gen.coroutine
     def track_future(self, future):
         yield _wait(future)
-        if self.results:
+        if self.with_results:
             result = yield future._result()
         with self.lock:
             self.futures[future] -= 1
             if not self.futures[future]:
                 del self.futures[future]
-            if self.results:
+            if self.with_results:
                 self.queue.put_nowait((future, result))
             else:
                 self.queue.put_nowait(future)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2518,8 +2518,17 @@ class AsCompleted(object):
     6
     12
     24
+
+    Optionally wait until the result has been gathered as well
+
+    >>> ac = as_completed([x, y, z], results=True)  # doctest: +SKIP
+    >>> for future, result in ac:  # doctest: +SKIP
+    ...     print(result)  # doctest: +SKIP
+    2
+    4
+    3
     """
-    def __init__(self, futures=None, loop=None):
+    def __init__(self, futures=None, loop=None, results=False):
         if futures is None:
             futures = []
         self.futures = defaultdict(lambda: 0)
@@ -2527,6 +2536,7 @@ class AsCompleted(object):
         self.lock = Lock()
         self.loop = loop or default_client().loop
         self.condition = Condition()
+        self.results = results
 
         if futures:
             for future in futures:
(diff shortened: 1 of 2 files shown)


mrocklin:
I thought about this problem for a while.  I started doing something more complex that tried to intelligently handle gathering from iterators or async-iterators.  I think that this would be an interesting project, but decided to scale back a bit and solve my immediate problem first.

eriknw:
+1.  I like the functionality and spelling, although I would probably change the keyword argument name to `with_results=True`.

I'm not too concerned about deviating from `concurrent.futures`.

If it's cheap and easy to do so, I would love to have a method such as `as_completed.has_results()` or `is_ready()` so one could extend your example loop to do the following:
```python
ac = as_completed(futures, results=True)
for future, result in ac:
    # handle future and maybe add new futures
    ...
    # do additional stuff while waiting on results
    while not ac.has_results():
        ...
```
Can and should this be done with `ac.queue.empty()`?  I think this is a useful primitive operation to have, as there are a lot of variants to the above example.  I would use it and `with_results=True` today if they were available.

mrocklin:
@eriknw, what is your application?

eriknw:
> @eriknw, what is your application?

An asynchronous pattern search.  Being able to cheaply query whether results are available or not would help with two things:

1. Advance the algorithm during the run-up time.  For example, suppose we want to have up to 100 tasks running at all times, but we begin receiving results after only 50 tasks have been submitted.  I would like to advance the algorithm based on the results received before submitting all 100 tasks, which, in this case, would be very easy to do algorithmically.  The run-up time here may be slow (let's say between 0.1 to 60 seconds) due to a one-time initialization cost.  The next 100 tasks will be generated *much* faster than the first 100 tasks, so we're confident we can use all the cores *eventually*.  The more cores we want to use, the more pressure and benefit there will be to advance the algorithm during run-up time.

2. The ability to process many completed tasks at once.  For example, one could do
```python
while not_done:
    results = []
    while ac.has_results():
        results.append(next(ac))
    # process results...
```

mrocklin:
Short term operating on `ac.queue` would certainly work

mrocklin:
Though I do agree that adding `has_results` or `result_count` and `get_batch` or something would be useful.

mrocklin:
I think that my suggestion would be to use `ac.queue` as you mention above and see what behaviors you find useful in practice. Then lets encode those as API.

eriknw:
Thanks, I'll give `ac.queue` a try.  I hope this wasn't too much of a tangent for this thread.

mrocklin:
Quite the contrary.  I think it's great, and I'm happy to know that improvements here will be immediately to someone that isn't me :)  Generally I'm very curious about how to make asynchronous workflows like this both usable and fast.  

The *right* solution to many of these problems is to use tornado or async/await syntax.  However I don't expect most users to pick this up easily.  The blocking syntax seems to be pretty approachable, but we need to make sure that we're doing the right productive work in the background.

pitrou:
Isn't your proposal much like `Client.gather`? Except that `Client.gather` gathers in-order while your proposal gathers opportunistically (probably reordered).

Intuitively I'd favour a separate API for this, except that I don't have a nice name to suggest.

pitrou:
By the way I think it's ok to augment the `as_completed` API. `asyncio` already does so for its own purposes. I'm a little less comfortable with changing the return type based on an argument's value, but practicality may beat purity here.

mrocklin:
Yes, my first solution to this was to start by making `Client.gather` act intelligently on `AsyncIterators`.  This lead to thinking about how to a general set of intelligent operations on `AsyncIterators` for windowing, batching, etc..  I stopped thinking about this just because I was going in too deeply. 

However, just improving client.gather on asynciterators is probably a good idea just on its own.  I'll see if I can write something up quickly.

mrocklin:
OK, so an alternative to this PR that is available today is the following:

```python
ac = as_completed(futures)
ac1, ac2 = itertools.tee(2, ac)
results = client.gather(ac2)
for future, result in zip(ac1, results):
    ...
```

This uses an extra thread, but that's fine.  It is a bit ugly though and probably not clear to the average user.

I'm tempted to let this change go in as written.  I think that this is likely to be a common case.  I wouldn't mind solving sequences-of-futures more thoroughly in the future, but I think it's a larger endeavor.

mrocklin:
@pitrou any objections to merging this?

pitrou:
No. Please do!

PR #975, opened 2017-03-27
Scheduler close also closes nannies

mrocklin:
This resolves an intermittent testing failure

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -503,9 +503,18 @@ class Scheduler(Server):
         self.status = 'closing'
         logger.debug("Cleaning up coroutines")
 
-        for w, comm in list(self.worker_comms.items()):
-            with ignoring(AttributeError):
-                yield comm.close()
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -503,9 +503,18 @@ def cleanup(self):
         self.status = 'closing'
         logger.debug("Cleaning up coroutines")
 
-        for w, comm in list(self.worker_comms.items()):
-            with ignoring(AttributeError):
-                yield comm.close()
+        close_nannies = {w: self.rpc(self.get_worker_service_addr(w, 'nanny')).terminate()
```
pitrou: Unrelated to this PR, but I'm wondering why the scheduler has to close both workers and nannies. Is it a historical accident or a deliberate decision? Intuitively, closing e.g. the worker to close explicitly should also tell the nanny to exit (or the reverse).

mrocklin: Yes, I can see how that might be cleaner.  Looking into this further it looks like the worker *already* tells the nanny to close by default.

PR #977, opened 2017-03-27
Add more tests for channels

### LINKED ISSUE #976: channel blocks on iteration and does not update data
I have the simple task of submitting a future from a client which then another client needs to get and see if it finished. The initial client is destroyed shortly after submitting and adding the future to the channel (flask application). This seems from the documentation to be the perfect task for the channel concept. However I have the issue that the second client blocks forever when iterating over the channel:

```
def add(self, a, b):
        time.sleep(1)
        return a+b

with Client('127.0.0.1:8786') as client:
            chan = client.channel("some-id")
            future = client.submit(self.add, 2, 2)
            chan.append(future)
 
            future = client.submit(self.add, 4, 6)
            chan.append(future)
            chan.flush()

with Client('127.0.0.1:8786') as other_client:
            chan = other_client.channel("some-id")
          
            for f in chan:
                 print(f)


```

The above blocks in iterating over the channel. Interestingly chan.data is empty immediately after channel creation, i.e. the chan.data is not refreshed when connecting to an existing channel. One ugly trick is the following:

```
with Client('127.0.0.1:8786') as other_client:
        chan = other_client.channel("some-id")
          
        def foo():
            pass

        foo = self.client.submit(foo)
        chan.append(foo)
        chan.flush()
 
        # the next two lines need a long explanation
        for t in chan:
            break
 
        for f in chan.data:
            for f in chan:
                 print(f)
```

Submit a dummy future and break on the iteration on chan. This updates the chan.data with the fresh future status. However this cannot be the right way to do it.

I appreciate to know if I am using channels wrong or this behaviour is wrong.

mrocklin:
Fixes #976

### DIFF
--- a/distributed/tests/test_channels.py
+++ b/distributed/tests/test_channels.py
@@ -225,9 +225,9 @@ def test_channel_gets_updates_immediately_2(loop):
 
             with Client(s['address']) as c2:
                 x2 = c.channel('x')
-                future = c.submit(inc, 1)
-                x.append(future)
-                x.flush()
+                future = c2.submit(inc, 1)
+                x2.append(future)
+                x2.flush()
 
             future = next(iter(x))
             assert future.result() == 2
--- a/distributed/tests/test_channels.py
+++ b/distributed/tests/test_channels.py
@@ -202,3 +202,32 @@ def test_values(c, s, a, b):
     assert list(x2.data) == data
 
     yield c2._shutdown()
+
+
+def test_channel_gets_updates_immediately(loop):
+    with cluster() as (s, [a, b]):
+        with Client(s['address']) as c:
+            x = c.channel('x')
+            future = c.submit(inc, 1)
+            x.append(future)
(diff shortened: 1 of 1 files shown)


distributed/tests/test_channels.py
```diff
@@ -202,3 +202,32 @@ def test_values(c, s, a, b):
...
+            assert future.result() == 2
+
+
+def test_channel_gets_updates_immediately_2(loop):
+    with cluster() as (s, [a, b]):
+        with Client(s['address']) as c:
+            x = c.channel('x')
+
+            with Client(s['address']) as c2:
+                x2 = c.channel('x')
```
pitrou: I might misunderstand the intent here, but do `x2` and `c2` actually have a use in this test?

mrocklin: Whoops.  Yes, that's a mistake.

PR #984, opened 2017-03-30
Task stream cleanup

mrocklin:
This changes the task stream diagnostic in two ways:

1.  Whenever it restarts it starts at time zero (Fixes https://github.com/dask/dask/issues/2106 cc @jbednar)
2.  Avoids jumping backwards in time, which occurred before if a task had been previously computed long ago.  This was visually jarring

### DIFF
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -64,7 +64,7 @@ class TaskStream(DashboardComponent):
             worker=[], y=[], worker_thread=[], alpha=[])
         )
 
-        x_range = DataRange1d()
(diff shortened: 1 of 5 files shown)


jbednar:
Sounds great, thanks!!

jbednar:
The code looks reasonable to me, and makes good sense, but I'm not otherwise familiar with this codebase.

mrocklin:
As an FYI this will only be available on port 8788/status in master for a
while, not 8787/status.  By next release we should hopefully have swapped
this over to 8787

PR #986, opened 2017-03-30
Fail gracefully if we are unable to write config file

mrocklin:
We don't need this file to exist, and sometimes we don't have permission.

cc @quasiben


quasiben:
👍

PR #989, opened 2017-03-31
Support connections between Python 2/3 client/scheduler

mrocklin:
This changes the Python 2 implementation to always use unicode.
This allows a Python 2 scheduler to support Python 3 client/workers and
the reverse.

This required the following changes:

1. from __future__ import unicode_literals everywhere
2. from .compatibility import unicode as str in many files

cc @pitrou @jcrist for thoughts

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -1,4 +1,4 @@
-from __future__ import print_function, division, absolute_import
+from __future__ import print_function, division, absolute_import, unicode_literals
 
 from datetime import timedelta
 import errno
@@ -124,7 +124,10 @@ class TCP(Comm):
         def finalize(stream=self.stream, r=repr(self)):
             if not stream.closed():
                 logger.warn("Closing dangling stream in %s" % (r,))
-                stream.close()
+                try:
+                    stream.close()
+                except socket.error:
+                    pass
 
         return finalize
 
@@ -166,6 +169,9 @@ class TCP(Comm):
         stream = self.stream
         if stream is None:
             raise CommClosedError
+
+        if isinstance(msg, dict) and any(isinstance(k, bytes) for k in msg):
+            print(msg)
 
         # IOStream.write() only takes bytes objects, not memoryviews
         frames = [ensure_bytes(f) for f in to_frames(msg)]
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -1,4 +1,4 @@
-from __future__ import print_function, division, absolute_import
+from __future__ import print_function, division, absolute_import, unicode_literals
 
 from .config import config
 from .core import connect, rpc
--- a/distributed/_ipython_utils.py
+++ b/distributed/_ipython_utils.py
@@ -4,7 +4,7 @@ These functions should probably reside in Jupyter and IPython repositories,
 after which we can import them instead of having our own definitions.
 """
 
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
 
 import atexit
 import os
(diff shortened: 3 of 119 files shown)


distributed/comm/tcp.py
```diff
@@ -124,7 +124,10 @@ def _get_finalizer(self):
         def finalize(stream=self.stream, r=repr(self)):
             if not stream.closed():
                 logger.warn("Closing dangling stream in %s" % (r,))
-                stream.close()
+                try:
+                    stream.close()
+                except socket.error:
```
pitrou: Which error specifically did you get here? It would be nice at least to comment about it.

pitrou:
I'm not sure that's a good idea. pickling compatibility between Python 2 and 3 is only guaranteed for common built-in types (and even then, only for pickle protocol 2 AFAIR). Trying to guarantee a broader compatibility will tax us maintenance-wise and may also lead to suboptimal performance.

Another possibility would be to declare the Python version at the first message when establishing a connection, and erroring out if the two endpoints' versions are detected to be incompatible.
(adding a first message to declare the distributed protocol and/or library version is a good idea anyway)

mrocklin:
The objective isn't to provide cross version support for data or functions, only for administrative data.  The clients and workers are still expected the match precisely, this only increases flexibility between the scheduler on one side and the clients/workers on the other.

The scheduler today never unpacks any Python language specific information (mostly).  In principle the scheduler could be written in another language entirely (like C or Go) or could serve non-Python workers and clients (for example for R or Julia) if they existed.  

The main change here is to ensure that administrative messages like `{'status': 'OK'}` use text rather than bytestrings.  Previously this would depend on the Python version used.  Python 2 would send bytestrings while Python 3 would send text.  Msgpack represents these differently.

mrocklin:
The two short-term impacts that I see are the following:

1.  We can use a PyPy2 scheduler with Python3 workers and clients
2.  People will run into this error less frequently

However I do agree with you that we should probably run something like the current `client.get_versions(check=True)` at startup and at least warn if versions are mismatched.

martindurant:
At some point you did an experiment with pypy and found that it gave a decent speedup to the scheduler. Do you know if that is still the case after many iterations of optimization?

pitrou:
> We can use a PyPy2 scheduler with Python3 workers and clients

I think we can reasonably hope that PyPy3 will become robust enough in the near future. I don't think PyPy users really like being stuck with Python 2, despite some of the ideology that's sometimes brandied about.

mrocklin:
PyPy may be a bit of a distraction here.  

I think the relevant question of this PR is "Should we spend development effort to ensure that Python 2 and Python 3 are consistent in how they send text in administrative messages?"

pitrou:
> I think the relevant question of this PR is "Should we spend development effort to ensure that Python 2 and Python 3 are consistent in how they send text in administrative messages?"

Is there an easy way to do it without the `unicode_literals` import? The latter can be a subtle bug magnet, due to behaviour differences between str and unicode even on Python 2 (for example, unicode filenames aren't very well handled on Python 2).

mrocklin:
Explicit unicode literals in relevant places with u'foo' ?

pitrou:
> Explicit unicode literals in relevant places with u'foo' ?

Well, that's reasonable, though it's easy to overlook some of them, since we don't really test this situation in our CI setup. As some best-effort thing, though, it sounds ok.

mrocklin:
This came up in conversation recently.  FWIW I'm against adding unicode literals throughout the codebase.  This doesn't seem important enough to be worth the hassle.

pitrou:
As long as someone else does the work and pledges to maintain it I'd have no issue with it. Otherwise agreed that it's not worth *our* hassle ;-)

mrocklin:
I suspect that the maintenance here would happen in every PR that touches
string literals.  It's also a step backwards in terms of Python 2->3
conversion.

PR #993, opened 2017-04-03
Add separate_thread keyword to worker_client

mrocklin:


vincentschut:
Interesting. Would you mind to explain the usecase to a noob like me? My first reaction is that this would inevitably lead to deadlocks once you'd have a task spawning a new worker_client(separate_thread=False) on each available worker...

mrocklin:
Yes, if you're going to be building large trees of long running tasks then you don't want to mess with this keyword argument, and should stick to the default.

However sometimes people use this feature without wanting to use long-running tasks.  They just want a client for a little while in a normal task.  For these people this feature is useful.

vincentschut:
Clear, thanks.

PR #994, opened 2017-04-03
Add batches method to as_completed

mrocklin:
Thoughts on this interface @eriknw ?

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2600,6 +2600,34 @@ class AsCompleted(object):
 
     next = __next__
 
+    def batches(self):
+        """
+        Yield all finished futures at once rather than one-by-one
+
+        This returns an iterator of lists of futures or lists of
+        (future, result) tuples rather than individual futures or individual
+        (future, result) tuples.  It will yield these as soon as possible
+        without waiting.
+
+        Examples
+        --------
+
+        >>> for batch in as_completed(futures).batches():
+        ...     results = client.gather(batch)
+        ...     print(results)
+        [4, 2]
+        [1, 3, 7]
+        [5]
+        [6]
+        """
+        seq = iter(self)
+        while True:
+            batch = [next(seq)]
+            with self.lock:
+                while not self.queue.empty():
+                    batch.append(self.queue.get())
+            yield batch
+
 
 as_completed = AsCompleted
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2626,9 +2626,8 @@ class AsCompleted(object):
             batch = [next(self)]
         else:
             batch = []
-        with self.lock:
-            while not self.queue.empty():
-                batch.append(self.queue.get())
+        while not self.queue.empty():
+            batch.append(self.queue.get())
         return batch
 
     def batches(self):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2600,6 +2600,37 @@ class AsCompleted(object):
 
     next = __next__
 
+    def next_batch(self, block=True):
+        """ Get next batch of futures from as_completed iterator
+
+        Parameters
+        ----------
+        block: bool, optional
+            If True then wait until we have some result, otherwise return
+            immediately, even with an empty list.  Defaults to True.
+
+        Examples
+        --------
+        >>> ac = as_completed(futures)  # doctest: +SKIP
+        >>> client.gather(ac.next_batch())  # doctest: +SKIP
+        [4, 1, 3]
+
+        >>> client.gather(ac.next_batch(block=False))  # doctest: +SKIP
+        []
+
+        Returns
+        -------
+        List of futures or (future, result) tuples
+        """
+        if block:
+            batch = [next(self)]
+        else:
+            batch = []
+        with self.lock:
+            while not self.queue.empty():
+                batch.append(self.queue.get())
+        return batch
+
     def batches(self):
         """
         Yield all finished futures at once rather than one-by-one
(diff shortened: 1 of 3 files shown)


eriknw:
It appears as though this blocks in order to return at least one item in the batch.  I don't want any blocking at all.  Typically, I expect to have separate code paths for (1) get all completed tasks without waiting (may be empty), and (2) wait for one result.

mrocklin:
Right, this is intended to be used within a for loop that waits until it has something to process.

I'm curious, in your application what work do you do when there isn't anything yet finished?  I assumed that you would be waiting here.

eriknw:
I understand you want to support the loop:
```python
for batch in as_completed(futures).batches():
    ...
```
This loop does not work for me.  Instead, my outer loop is basically `while not finished:`.

> I'm curious, in your application what work do you do when there isn't anything yet finished? I assumed that you would be waiting here.

I submit new tasks.  If I can't submit new tasks (e.g., max tasks have been submitted), I wait for a single result.

If one or more results are available, I retrieve them all (no waiting).  At this point, I may either submit new tasks, process the results, or both.

Perhaps `as_completed` is the wrong construct for my application.

eriknw:
Give me a few more moments to think if your proposed loop will work for my application.

eriknw:
The logic in my application is somewhat delicate and evolved organically.  Let me try to sketch it out:
```python
while not finished:
    if is_next_iteration:
        is_next_iteration = False
        ... # advance the algorithm

    # Submit tasks
    while self.can_submit() and (self.must_submit() or not self.results_available()):
        ...
        submit(new_task)
        ...
        if some_condition:
            is_next_iteration = True
            break

    if is_next_iteration:
        continue

    # Collect all completed tasks
    while as_completed.has_results():
        future, result = next(as_completed)
        ...
        processing.append((future, result))
        # self.can_submit() may change to True

    # Wait for one task if necessary
    if not self.can_submit() and not processing:
        future, result = next(as_completed)
        ...
        processing.append((future, result))

    # Process all results
    if processing and not self.must_submit():
        for item in processing:
            ...
            if some_condition:
                is_next_iteration = True
        processing.clear()
```
Stopping condition logic not shown.

mrocklin:
OK, so one improvement would be to factor the logic from `batches` into an explicit `next_batch` method that would be the batch-equivalent to calling `next`.

mrocklin:
Is the section `    # Wait for one task if necessary` necessary?  What happens if a bit of extra work has come in while you've been collecting?  Is this an important race condition?

I guess I'm still thinking that something like the following is often sufficient.  I don't think I understand the example above well enough to say so definitively however.

```python
batches = as_completed.batches()

while not stopping_condition():
    do stuff
    batch = next(batches)
    do stuff with batch
    for ...
        as_completed.add(...)
```

eriknw:
I could get behind a `next_batch` method that doesn't wait, although I still want a way to check whether results are available.

> Is the section `# Wait for one task if necessary` necessary? What happens if a bit of extra work has come in while you've been collecting? Is this an important race condition?

What race condition?  The code block denoted by `# Wait for one task if necessary` is the *only* place where we wait for anything to happen.  No other part of the algorithm is able to proceed.  If we deleted this code block, the algorithm would behave the same: it would wait for the next result (or results if many came in at the same time).  However, it would do so by running in a CPU hot-loop until `while as_completed.has_results():` (the previous code block) became True.

To reiterate, I view the acts of
1. retrieving all available results without waiting, and
2. waiting for one result

to be very distinct.  I want them both.  I don't want to merge them.  Well, I suppose these could be merged if a `timeout` argument was provided (I would pass `0` when I don't want to wait, and `None` when I want to wait), but I don't need a timeout per se.

eriknw:
> I guess I'm still thinking that something like the following is often sufficient. [...]

Probably.  It's not sufficient for my current needs though.  I could actually refit my algorithm to follow this flow and it would converge just fine.  There is a burn-in period for my algorithm, though, which is the main reason I went through the extra effort to stop submitting tasks if there are results available to process.

distributed/client.py
```diff
@@ -2600,6 +2600,59 @@ def __anext__(self):
...
+
+        Returns
+        -------
+        List of futures or (future, result) tuples
+        """
+        if block:
+            batch = [next(self)]
+        else:
+            batch = []
+        with self.lock:
```
eriknw: Why do you need to lock here?  `self.queue` operations already use locks.

mrocklin: I suppose general paranoia.  You're right though.  I think we've used the lock historically because we also check `self.futures` which is not threadsafe and is modified in a separate thread.  That's not the case in this code though (except implicitly in `next`) .  Removing.

eriknw:
Thanks!

One of the delicate features of the above algorithm sketch is that it *is* possible to leave the "submit" loop and not need to acquire any new results from `as_completed`.  But, we would always like to get results if there are any available.  I'm sorry this was unclear.

mrocklin:
> One of the delicate features of the above algorithm sketch is that it is possible to leave the "submit" loop and not need to acquire any new results from as_completed. But, we would always like to get results if there are any available. I'm sorry this was unclear.

As I understand it `next_batch` as written satisfies this.  If there are results, it returns them.  If `block=False` and there are no results, it returns `[]`.  I may still be confused about your intent though.

eriknw:
Yeah, `next_batch(block=False)` satisfies this.  I was trying to clarify any previous confusion regarding the given algorithm.

+1.

eriknw:
`next_batch` made my application cleaner.  Check out the commit above that referenced this issue.  Thanks for your effort here @mrocklin.

mrocklin:
:)

PR #94, opened 2016-02-02
Add CompatibleExecutor

minrk:
a concurrent.futures-compatible Executor subclass. Only difference from Executor at this point: map yields results instead of returning list of Futures.

I'm not at all tied to the class name, but I couldn't think of a better one. Happy to change it if you have a better idea.

closes #91

PR #997, opened 2017-04-04
Add concurrent.futures-compatible Executor

pitrou:
The CompatibleExecutor added in #94 isn't compatible with wait() or as_completed() functions from the concurrent.futures package.

We replace it with a better implementation that returns concurrent.futures Futures.

### DIFF
--- a/distributed/cfexecutor.py
+++ b/distributed/cfexecutor.py
@@ -1,0 +1,151 @@
+from __future__ import print_function, division, absolute_import
+
+from functools import partial
+import weakref
+
+import six
+
+from toolz import merge
+
+from tornado import gen
+
+from concurrent.futures import Future, Executor
+
+from .utils import sync
+
+
+@gen.coroutine
+def _cascade_future(future, cf_future):
+    """
+    Coroutine that waits on future, then transmits its outcome to cf_future.
+    """
+    result = yield future._result(raiseit=False)
+    status = future.status
+    if status == 'finished':
+        cf_future.set_running_or_notify_cancel()
+        cf_future.set_result(result)
+    elif status == 'cancelled':
+        cf_future.cancel()
+        cf_future.set_running_or_notify_cancel()
+    else:
+        cf_future.set_running_or_notify_cancel()
+        try:
+            six.reraise(*result)
+        except BaseException as exc:
+            cf_future.set_exception(exc)
+
+
+@gen.coroutine
+def _wait_on_futures(futures):
+    try:
+        yield [futures]
+    except Exception:
+        # One future errored, ignore
+        pass
+
+
+class ClientExecutor(Executor):
+    """
+    A concurrent.futures Executor that executes tasks on a distributed Client.
+    """
+
+    _allowed_kwargs = frozenset(['pure', 'workers', 'resources', 'allow_other_workers'])
+
+    def __init__(self, client, **kwargs):
+        assert set(kwargs) <= self._allowed_kwargs, "unsupported kwargs to ClientExecutor"
+        self._client = client
+        self._futures = weakref.WeakSet()
+        self._shutdown = False
+        self._kwargs = kwargs
+
+    def _wrap_future(self, future):
+        """
+        Wrap a distributed Future in a concurrent.futures Future.
+        """
+        cf_future = Future()
+        # Support cancelling task through .cancel() on c.f.Future
+        def cf_callback(cf_future):
+            if cf_future.cancelled() and future.status != 'cancelled':
+                future.cancel()
+
+        cf_future.add_done_callback(cf_callback)
+
+        self._client.loop.add_callback(_cascade_future, future, cf_future)
+        return cf_future
+
+    def submit(self, fn, *args, **kwargs):
+        """Submits a callable to be executed with the given arguments.
+
+        Schedules the callable to be executed as fn(*args, **kwargs) and returns
+        a Future instance representing the execution of the callable.
+
+        Returns:
+            A Future representing the given call.
+        """
+        if self._shutdown:
+            raise RuntimeError('cannot schedule new futures after shutdown')
+        future = self._client.submit(fn, *args, **merge(self._kwargs, kwargs))
+        self._futures.add(future)
+        return self._wrap_future(future)
+
+    def map(self, fn, *iterables, timeout=None, chunksize=1):
+        """Returns an iterator equivalent to map(fn, iter).
+
+        Args:
+            fn: A callable that will take as many arguments as there are
+                passed iterables.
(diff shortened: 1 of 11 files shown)


docs/source/conf.py
```diff
@@ -36,6 +36,7 @@
     'sphinx.ext.viewcode',
     'sphinx.ext.autosummary',
     'sphinx.ext.extlinks',
+    'sphinx.ext.intersphinx',
```
mrocklin: Do we need to add this to docs/requirements.txt for readthedocs?

distributed/cfexecutor.py
```diff
@@ -0,0 +1,165 @@
...
+
+import weakref
+
+import six
+
+from toolz import merge
+
+from tornado import gen
+
+from concurrent.futures import Future, Executor, TimeoutError
```
mrocklin: Given that these terms are used elsewhere in the codebase to mean different things it might be wise to instead use the namespace explicitly.  Perhaps the following?

```python
import concurrent.futures as cf

cf.Future, cf.Executor, cf.TimeoutError
```

distributed/cfexecutor.py
```diff
@@ -0,0 +1,165 @@
...
+
+from .metrics import time
+from .utils import sync
+
+
+@gen.coroutine
+def _cascade_future(future, cf_future):
+    """
+    Coroutine that waits on future, then transmits its outcome to cf_future.
```
mrocklin: What kind of future is the first?  is it a Dask future?

distributed/cfexecutor.py
```diff
@@ -0,0 +1,165 @@
...
+    for fut in futures:
+        try:
+            yield fut
+        except Exception:
+            pass
+
+
+class ClientExecutor(Executor):
+    """
+    A concurrent.futures Executor that executes tasks on a distributed Client.
```
mrocklin: distributed Client -> dask.distributed Client

distributed/cfexecutor.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    _allowed_kwargs = frozenset(['pure', 'workers', 'resources', 'allow_other_workers'])
+
+    def __init__(self, client, **kwargs):
+        assert set(kwargs) <= self._allowed_kwargs, "unsupported kwargs to ClientExecutor"
```
mrocklin: Perhaps an explicit `raise ValueError(...)` instead so that this sticks around even when assertions are turned off.

distributed/client.py
```diff
@@ -48,6 +49,15 @@
 _global_client = [None]
 
 
+def _timeout_wrapper(timeout, coro_func, *args, **kwargs):
+    coro = coro_func(*args, **kwargs)
+    if timeout is None:
+        return coro
+    else:
+        # XXX should this reraise concurrent.futures.TimeoutError?
+        return gen.with_timeout(timedelta(seconds=timeout), coro)
```
mrocklin: I suppose that the answer here is "yes" if this is what gets raised to the user

mrocklin: Oh, I see, this is in the dask side, not the concurrent futures side.  I would leave it as a tornado TimeoutError for now.

distributed/client.py
```diff
@@ -107,9 +117,14 @@ def done(self):
         """ Is the computation complete? """
         return self.event.is_set()
 
-    def result(self):
-        """ Wait until computation completes. Gather result to local process """
-        result = sync(self.client.loop, self._result, raiseit=False)
+    def result(self, timeout=None):
+        """ Wait until computation completes. Gather result to local process.
+
+        If `timeout` seconds are elapsed before returning, a TimeoutError
```
mrocklin: I think that sphinx uses restructured text by default, so you'll need double backticks if you want to use code highlighting.

distributed/client.py
mrocklin: Thoughts on adding a `timeout=` keyword to sync instead of using this function?

distributed/client.py
```diff
@@ -691,6 +714,23 @@ def shutdown(self, timeout=10):
...
 
+    def get_executor(self, **kwargs):
+        """ Return a concurrent.futures Executor for submitting tasks
+        on this Client.
+
+        Parameters
+        ----------
+        **kwargs:
+            Any submit()- or map()- compatible arguments, such as
+            `workers` or `resources`.
```
mrocklin: Why not include these on the submit or map calls themselves?

mrocklin: We can still be compatible with an interface even if we're a superset

distributed/tests/test_client.py
```diff
@@ -246,6 +222,11 @@ def test_thread(loop):
             x = c.submit(inc, 1)
             assert x.result() == 2
 
+            x = c.submit(slowinc, 1, delay=0.3)
+            with pytest.raises(gen.TimeoutError):
+                x.result(timeout=0.01)
```
mrocklin: Cool.  I'm glad to see this.

distributed/client.py
pitrou: I thought about that, but presently `timeout` can be one of the arguments forwarded to sync's `func` callable argument.

distributed/client.py
pitrou: People would typically use ClientExecutor to pass it to code expecting a c.f.Executor, i.e. ignorant of our additional arguments.

distributed/client.py
mrocklin: Perhaps `callback_timeout` or something else?

pitrou: Ok, will do.

mrocklin:
It looks like this breaks dask imports:

```python
ImportError                               Traceback (most recent call last)
<ipython-input-1-29a650dc246d> in <module>()
----> 1 from dask.distributed import Client

/home/mrocklin/workspace/dask/dask/distributed.py in <module>()
      2 from __future__ import absolute_import, division, print_function
      3 
----> 4 from distributed import (Client, Executor, progress, as_completed, wait,
      5         LocalCluster, Scheduler, Worker, nanny, local_client, config,
      6         CompatibleExecutor)

ImportError: cannot import name 'CompatibleExecutor'
```

If possible I'd like to deprecate the term `CompatibleExecutor` for a version.  This could be as simple as the following:

```python
def CompatibleExecutor(*args, **kwargs):
    raise Exception("This has been moved to the Client.get_executor() method")
```

mrocklin:
Something we should be aware of is that "Executor" is used by Spark to mean "the thing on the worker that performs tasks".  Migrating Spark users may find the "get_executor" method name confusing.  I don't have a good solution to this, I just thought I'd bring it up.

mrocklin:
Other than the backwards compatibility issue this looks good to me

PR #999, opened 2017-04-05
Improve diagnostics for joblib backend

jcrist:
Previously the task name in all joblib related calls was just `"joblib"`. This changes it to the function name of the called function.

One potential future change might be to include the batch size in the keyname.  This might be nice, as the batch size is used for work-stealing, and is mildly related to the time it takes for the task to complete. However, different parameters can also wildly affect the compute time, so this may not be useful in real applications.


mrocklin:
Thanks @jcrist !

PR #1000, opened 2017-04-06
Add OS X Travis builder

pitrou:


pitrou:
Looks like the OS X builder is too slow for `test_open_close_many_workers[Worker-100-5] `, or it genuinely hangs on that test... Perhaps someone can test on their local OS X machine?

mrocklin:
This test takes around 7-8 seconds on my local OS X machine

mrocklin:
And passes

pitrou:
Ah. Perhaps we need to disable slow tests on Travis OS X?

pitrou:
Ok, it seems there are a lot of failures on the OS X builder. Some of them are timeouts, some of them are other errors. I also see "Too many open files" in some of the printed out log excerpts.

pitrou:
I can't push this any further as I don't have an OS X machine to diagnose these errors. Someone else may take over.

pitrou:
Things seem much better now. We're down to one failing test on OS X, and hopefully zero soon.

mrocklin:
This looks fine to me.  Looks like it has conflicts with master

pitrou:
Weird, I don't see any conflicts here. Is a Github glitch?

mrocklin:
Either way it sounds like you should press the green button.  Either it will work or it won't.

PR #1004, opened 2017-04-07
Localcluster cleanup

mrocklin:
This does four things:

1.  Replace the nanny= keyword with processes= keyword
2.  Pass keywords from the Client down the LocalCluster if started with `Client(**kwargs)`  (cc @pelson)
3.  Replace the separate process Bokeh server with the in-process one.  I want to start with this before changing it out on dask-scheduler
4.  Adds bokeh servers to all of the workers by default

Everything seems fine so far.  However I'm now seeing this (cc @pitrou):

```python

In [2]: client = Client()

In [3]:                                                                         
Do you really want to exit ([y]/n)? 
/home/mrocklin/Software/anaconda/lib/python3.6/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 20 leaked semaphores to clean up at shutdown
  len(cache))
```

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -1,10 +1,12 @@
 from __future__ import print_function, division, absolute_import
 
+import atexit
 import logging
 import math
 from threading import Thread
 from time import sleep
 import warnings
+import weakref
 
 from tornado import gen
 from tornado.ioloop import IOLoop
@@ -114,6 +116,8 @@ class LocalCluster(object):
 
         if start:
             sync(self.loop, self._start, ip)
+
+        clusters_to_close.add(self)
 
     def __str__(self):
         return ('LocalCluster(%r, workers=%d, ncores=%d)' %
@@ -234,6 +238,13 @@ class LocalCluster(object):
     def close(self):
         """ Close the cluster """
         if self.status == 'running':
+            for w in self.workers:
+                self.loop.add_callback(self._stop_worker, w)
+            for i in range(10):
+                if not self.workers:
+                    break
+                else:
+                    sleep(0.01)
             self.status = 'closed'
             if self.loop._running:
                 sync(self.loop, self._close)
@@ -287,3 +298,12 @@ class LocalCluster(object):
             return self.scheduler.address
         except ValueError:
             return '<unstarted>'
+
+
+clusters_to_close = weakref.WeakSet()
+
+
+@atexit.register
+def close_clusters():
+    for cluster in clusters_to_close:
+        cluster.close()
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -4,6 +4,7 @@ import logging
 import math
 from threading import Thread
 from time import sleep
+import warnings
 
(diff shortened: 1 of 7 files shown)


mrocklin:
So it looks like I can both remove the semaphore issues (at least from here) *and* reduce the client startup time from 2.0s to 0.6 seconds by using multiprocessing fork rather than the forkserver.  For the purposes of interactively constructed Nannies through `Client()` this seems like a win.  

Is there any reason why I would prefer to keep using forkserver in a single-machine context?

martindurant:
Weren't there various one-process things that shouldn't be allowed to fork (thinking libhdfs3 state).

mrocklin:
Yes, I believe that libhdfs3 was the original motivation for the
forkserver.  This is unlikely to be an issue when using `Client()` though.
One approach here would be to have Nanny accept a multiprocessing context
object that defaulted normally to the forkserver but that LocalCluster
would redirect to the multiprocessing module itself (which uses fork).
This would still be override-able if someone wanted to say use LocalCluster
to read HDFS data.  At the moment this seems like a corner case though.

martindurant:
So could we get around this with careful documentation ("If you use hdfs3 with LocalCluster/Client(), be sure to create processes before importing hdfs3") or a runtime check, rather than passing multiprocessing contexts.

pitrou:
Using forkserver is orthogonal to single-machine or multiple-machine contexts. The default *fork* method can produce various hard to debug issues when inheriting system resources (such as file descriptors, sockets...) in child processes.

As for the semaphore leak issue, chances are the child resources are not cleaned up properly. Semaphores appear in synchronization primitives such as multiprocessing Queues. There may be some of them in the Process itself. This means probably you're not shutting down the children cleanly.

mrocklin:
Are there ways to get more information about which semaphores are not being
cleaned up?

mrocklin:
Resolved (at least for LocalCluster) in latest commit.  

I would still like to get the startup time for Client down if possible

```python

In [2]: %time client = Client()
CPU times: user 528 ms, sys: 52 ms, total: 580 ms
Wall time: 2.41 s
```

mrocklin:
Second time around is less, but still a while

```python
In [3]: %time client = Client()
CPU times: user 112 ms, sys: 0 ns, total: 112 ms
Wall time: 1.26 s
```

mrocklin:
OK, going to go ahead and merge this soon if no one has objections.

distributed/deploy/local.py
```diff
@@ -27,11 +30,10 @@ class LocalCluster(object):
     ----------
     n_workers: int
         Number of workers to start
+    processes: bool
+        Whether to use processes (True) or threads (False).  Defaults to True
```
pitrou: By the way, why is the default true? Because of better performance on some workloads?

mrocklin: Yes.  Anecdotally I've observed that this is more often what people seem to want and expect.

Changing this would be very convenient for us though.

Issue #1006, opened 2017-04-07
Specify network interface rather than IP


mrocklin:
Some clusters have multiple overlapping networks.  For example an HPC cluster may have both ethernet and infiniband.  Currently users can specify which network to use by identifying each worker by the proper IP address associated to that network

```
dask-worker ... --host 192.168.100.1
dask-worker ... --host 192.168.100.2
dask-worker ... --host 192.168.100.3
```

However it can be challenging to specify each IP address individually for large cluster deployments, especially if set up by someone with modest networking knowledge (which is common for scientific users).  Are there other ways that we can support preferentially choosing particular networks, such as by providing network interface names or providing an address that they should try to connect to?

cc @pitrou @quasiben for networking help

mrocklin:
Currently we do something like the following internally:

```python
sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
sock.connect(('8.8.8.8', 80))
sock.getsockname()
```

Perhaps this can be resolved by inserting an address that is more specific than `'8.8.8.8'`.

rabernat:
If we could specify an interface argument, that would solve the problem. Like

``
dask-worker --interface eth0
``

the worker script could then look up the IP to use, doing something like this: http://stackoverflow.com/questions/24196932/how-can-i-get-the-ip-address-of-eth0-in-python

mrocklin:
Proof of concept in #1007

PR #1007, opened 2017-04-07
Add --interface keyword to cli

mrocklin:
Fixes https://github.com/dask/distributed/issues/1006

cc @rabernat

```
dask-scheduler --interface ib0  # identify over infiniband network
dask-worker scheduler:8786 --interface ib0
```

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -13,7 +13,7 @@ import click
 
 import distributed
 from distributed import Scheduler
-from distributed.utils import ignoring, open_port
+from distributed.utils import ignoring, open_port, get_ip_interface
 from distributed.http import HTTPScheduler
 from distributed.cli.utils import (check_python_3, install_signal_handlers,
                                    uri_from_host_port)
@@ -33,6 +33,8 @@ logger = logging.getLogger('distributed.scheduler')
               required=False, help="Launch Bokeh Web UI")
 @click.option('--host', type=str, default='',
               help="IP, hostname or URI of this server")
+@click.option('--interface', type=str, default=None,
+              help="Preferred network interface like 'eth0' or 'ib0'")
 @click.option('--show/--no-show', default=False, help="Show web UI")
 @click.option('--bokeh-whitelist', default=None, multiple=True,
               help="IP addresses to whitelist for bokeh.")
@@ -47,7 +49,8 @@ logger = logging.getLogger('distributed.scheduler')
               "This may be a good way to share connection information if your "
               "cluster is on a shared network file system.")
 def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
-         bokeh_whitelist, prefix, use_xheaders, pid_file, scheduler_file):
+         bokeh_whitelist, prefix, use_xheaders, pid_file, scheduler_file,
+         interface):
 
     if pid_file:
         with open(pid_file, 'w') as f:
@@ -63,6 +66,9 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
         soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
         limit = max(soft, hard // 2)
         resource.setrlimit(resource.RLIMIT_NOFILE, (limit, hard))
+
+    if interface and not host:
+        host = get_ip_interface(interface)
 
     addr = uri_from_host_port(host, port, 8786)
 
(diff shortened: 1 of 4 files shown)


mrocklin:
@rabernat are you able to test this on your system?

distributed/cli/dask_scheduler.py
```diff
@@ -64,6 +67,9 @@ def del_pid_file():
         limit = max(soft, hard // 2)
         resource.setrlimit(resource.RLIMIT_NOFILE, (limit, hard))
 
+    if interface and not host:
```
pitrou: We should probably disallow passing both host and interface, instead of sliently ignoring one of them.

distributed/utils.py
```diff
@@ -128,6 +128,22 @@ def get_ipv6(host='2001:4860:4860::8888', port=80):
     return _get_ip(host, port, family=socket.AF_INET6, default='::1')
 
 
+import socket
+import fcntl
+import struct
+
+def get_ip_interface(ifname):
+    # http://stackoverflow.com/a/24196955/616616 by @martin-konecny
```
pitrou: Wow. How about using [psutil.net_if_addrs](https://pythonhosted.org/psutil/#psutil.net_if_addrs) instead?
Or, if that doesn't work for us, there's the dedicated [netifaces](https://pypi.python.org/pypi/netifaces) package.

distributed/cli/tests/test_dask_scheduler.py
```diff
@@ -186,6 +186,20 @@ def test_multiple_workers(loop):
                         assert time() < start + 10
 
 
+def test_interface(loop):
+    with popen(['dask-scheduler', '--no-bokeh', '--interface', 'lo']) as s:
```
pitrou: This test will probably have to be skipped under Windows...

distributed/cli/dask_scheduler.py
mrocklin: Fixed, though not tested

distributed/utils.py
mrocklin: OK, using psutil

distributed/cli/tests/test_dask_scheduler.py
mrocklin: Now skipping on windows

mrocklin:
@jakirkham do you have any insight on what is going on with the windows builds here?

rabernat:
> @rabernat are you able to test this on your system?

Should be able to give it a try tomorrow.

distributed/utils.py
```diff
@@ -128,6 +128,11 @@ def get_ipv6(host='2001:4860:4860::8888', port=80):
     return _get_ip(host, port, family=socket.AF_INET6, default='::1')
 
 
+def get_ip_interface(ifname):
+    import psutil
+    return psutil.net_if_addrs()[ifname][0].address
```
pitrou: psutil doesn't specify in which order the results are returned, so you should really select the result for which the `family` member is `AF_INET` (or error out if there is None).

pitrou:
The AppVeyor issue looks like either a `conda` issue or an issue with some conda-forge packages (or both). i've asked on the conda channel.

jakirkham:
Not sure if this is the problem, but this seems [strange]( https://ci.appveyor.com/project/daskdev/distributed/build/1.0.1008#L170 ). Probably worth raising over at the feedstock.

pitrou:
If you merge the latest changes from master, it will hopefully fix the AppVeyor issue.

rabernat:
I tested this on my cluster and it worked!

My impression is that it went faster on the infiniband interface, but can you suggest a specific test I could run to benchmark communication performance?

mrocklin:
That was what I was about to suggest.  

OK, cool.  So it looks like this had the desired effect *and* a user (@rabernat) was able to use it effectively without confusion.

I plan to merge this soon.  

@rabernat I think that the next thing to do here is to compare the bandwidth you got to what you should expect on your network.  `client.scatter(x).gather()` involves four transfers `client->scheduler->worker->scheduler->client`.  So if you move 100MB in 1s then that means that we're observing 400MB/s bandwidths.  So two questions:

1.  What bandwidth are you seeing?  `data.nbytes / transfer_time / 1e6`
2.  What do you expect to see?  Your IT might know this or you or your colleagues might have a general understanding.

weiliu620:
I use the `--interface` for both scheduler and worker, but could not get it to work on InfiniBand. 

I know little about computer network concept, but I check the results of `get_ip_interface` and it is the correct address of the `ib0` interface.  My understanding is the scheduler just use this IP address to set up the inter-process communication (with Tornado?). 

So, where should I look to debug?  I'll also work with my system admin.

jakirkham:
Could you please migrate this to a new issue @weiliu620? Expect it will have low visibility here.

Also would recommend including what you tried and what versions of things you are using in that issue.

PR #1009, opened 2017-04-10
Try to fix AppVeyor builds

pitrou:


pitrou:
Turns out the issue is with either `conda` or some conda-forge packages. Will wait for more guidance from the conda folks.

mrocklin:
@jakirkham fixed similar issues that just occurred on dask/dask.  I
wouldn't be surprised to learn that they had similar causes.

jakirkham:
This also has the same [issue]( https://ci.appveyor.com/project/daskdev/distributed/build/1.0.1006#L170 ) that the other AppVeyor build did.

I don't know if this `qt` package issue is the problem or a different issue. Though it just strikes me as weird that the `qt` package has issues installing and then every package can't be imported. It's almost like a rollback should have occurred, but did not. So the environment is in some sort of incomplete state.

pitrou:
It seems updating conda may fix the issue.

PR #1010, opened 2017-04-11
Add ticker to CurrentLoad components to set ticks at nice memory amts

canavandl:
Example:

screen shot 2017-04-11 at 11 25 28 am

Are there other plots that should be updated as well?

### DIFF
--- a/distributed/bokeh/scheduler.py
+++ b/distributed/bokeh/scheduler.py
@@ -13,7 +13,7 @@ from bokeh.application.handlers.function import FunctionHandler
 from bokeh.layouts import column, row
 from bokeh.models import ( ColumnDataSource, DataRange1d, HoverTool, ResetTool,
         PanTool, WheelZoomTool, TapTool, OpenURL, Range1d, Plot, Quad, Text,
-        value, LinearAxis, NumeralTickFormatter)
+        value, LinearAxis, NumeralTickFormatter, BasicTicker)
 from bokeh.models.widgets import DataTable, TableColumn
 from bokeh.plotting import figure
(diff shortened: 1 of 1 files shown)


distributed/bokeh/scheduler.py
```diff
@@ -288,6 +288,7 @@ def __init__(self, scheduler, width=600, **kwargs):
                         x='nbytes-half', y='y',
                         width='nbytes', height=1,
                         color='nbytes-color')
+            nbytes.axis[0].ticker = BasicTicker(mantissas=[1,256,512], base=1024)
```
mrocklin: I'm curious, can you explain what this is doing so that I can replicate this in the future?

mrocklin:
> Are there other plots that should be updated as well?

Nothing yet that is likely to stick around.

distributed/bokeh/scheduler.py
canavandl: from the docs:

```
    ''' Generate "nice" round ticks at any magnitude.

    Creates ticks that are "base" multiples of a set of given
    mantissas. For example, with ``base=10`` and
    ``mantissas=[1, 2, 5]``, the ticker will generate the sequence::

        ..., 0.1, 0.2, 0.5, 1, 2, 5, 10, 20, 50, 100, ...

    '''
```

With the base arg of 1024, it should nicely handle the kb -> mb -> gb transitions. The mantissas specify that 1*b, 0.25*b, 0.5*b steps are desired. (Let me know if i'm not explaining it well)

mrocklin:
Thanks @canavandl !  Merging.

PR #1014, opened 2017-04-12
Partner to https://github.com/dask/dask/pull/2207

martindurant:
Massage paths to deal with name/password/host/port .


martindurant:
Replaces https://github.com/dask/dask/pull/2171

pitrou:
Perhaps it would be useful to add tests for such subtleties?

mrocklin:
This looks good to me

Issue #1013, opened 2017-04-12
Preloading a module in every worker processes


bm371613:
I would like to propose a new `--preload` option for `dask-worker`. It would let the user pass a module string (`foo.bar`) and then the module would be loaded in every worker process.

Rationale
---
The preloaded module could initialize some resources, later used by tasks. Lazy initialization is not an option if workers might often be stopped or started and the initialization is too slow for a task to risk waiting.

Would you accept a PR with that feature?

mrocklin:
In general I'm in favor of the idea of preloading state.  Because many people start dask in different way I think that this should be deeper than the command line interface to start.  This is similar to https://github.com/dask/distributed/issues/495 and PR https://github.com/dask/distributed/pull/505  (never merged).

I suspect that the scheduling bits of this make them somewhat more challenging.  I can see two (of many) paths forward

1.  Try to push the current PR forward
2.  Implement something simpler in such a way that it could be easily extended to the full scope of environments in the future

Thoughts @bm371613 ?

bm371613:
These `Environment`s would not solve my problem, as they are registered by a client. I would require an environment to be there before any client connects. One of the two would have two be possible:

- the scheduler can be started with a `required_environments` option that are later set up for every worker that connects
- the worker can be started with a `initial_environments` option, and then it tells the scheduler what is available already

As environments are objects, they cannot be passed as CLI options to scheduler/worker easily. They could be initialized in a module which would be preloaded and in that sense, what I propose could work together with #495. Other than that, I would see these problems separately.

mrocklin:
Environments as defined in that issue are stored on the scheduler and are
tried against workers as they arrive.  How the environments arrive on the
scheduler is an interesting question but generally not the hard part of the
problem.  Presumably in your case you would set up the scheduler, register
an environment with it somehow, either with custom code or by spinning up a
short lived client.  Does starting a short-lived client pose a problem?

Running some arbitrary python script at startup as you suggest is also a
solution.  I'm slightly less excited by it just because it feels
tacked-on.  This sort of thing does come up decently often though where
people want to slightly modify the setup process.  It would be interesting
to think about how we would want to arrange this.  If you want to put
together a proposal I'd be happy to look at it.

What would your script do precisely?  Does it contain a particular function
that we call?  How does it learn about the scheduler or worker that it is
supposed to modify?  Does it return something?  Perhaps the worker again?
Should it expect the event loop to be running or not running at the time
when it is called?

[...]

PR #1016, opened 2017-04-13
Preloading a module in every worker processes

bm371613:
Implements #1013

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -9,12 +9,12 @@ import tempfile
 
 import click
 
-import distributed.preloading
 from distributed import Scheduler
 from distributed.utils import ignoring, open_port, get_ip_interface
 from distributed.http import HTTPScheduler
 from distributed.cli.utils import (check_python_3, install_signal_handlers,
                                    uri_from_host_port)
+from distributed.preloading import preload_modules
 from tornado.ioloop import IOLoop
 
 logger = logging.getLogger('distributed.scheduler')
@@ -94,9 +94,7 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
     scheduler = Scheduler(loop=loop, services=services,
                           scheduler_file=scheduler_file)
     scheduler.start(addr)
-    distributed.preloading.preload(preload,
-                                   parameter=scheduler,
-                                   file_dir=local_directory)
+    preload_modules(preload, parameter=scheduler, file_dir=local_directory)
 
     bokeh_proc = None
     if _bokeh:
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -113,7 +113,7 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
         except Exception as e:
             logger.warn("Could not start Bokeh web UI", exc_info=True)
 
-    logger.info('      Local Directory: %26s', local_directory)
+    logger.info('Local Directory: %26s', local_directory)
     logger.info('-' * 47)
     try:
         loop.start()
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -63,9 +63,14 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
                 os.remove(pid_file)
         atexit.register(del_pid_file)
 
-    local_directory = local_directory or tempfile.mkdtemp(prefix='scheduler-')
-    if not os.path.exists(local_directory):
-        os.mkdir(local_directory)
+    local_directory_created = False
+    if local_directory:
+        if not os.path.exists(local_directory):
+            os.mkdir(local_directory)
+            local_directory_created = True
+    else:
+        local_directory = tempfile.mkdtemp(prefix='scheduler-')
+        local_directory_created = True
     if local_directory not in sys.path:
         sys.path.insert(0, local_directory)
 
@@ -120,7 +125,7 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
         scheduler.stop()
         if bokeh_proc:
             bokeh_proc.close()
-        if os.path.exists(local_directory):
+        if local_directory_created:
             shutil.rmtree(local_directory)
 
         logger.info("End scheduler at %r", addr)
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -4,9 +4,11 @@ import atexit
 import json
 import logging
 import os
+import shutil
 import socket
 import subprocess
 import sys
+import tempfile
 from time import sleep
 
 import click
@@ -48,11 +50,13 @@ logger = logging.getLogger('distributed.scheduler')
               help="File to write connection information. "
               "This may be a good way to share connection information if your "
               "cluster is on a shared network file system.")
+@click.option('--local-directory', default='', type=str,
+              help="Directory to place scheduler files")
 @click.option('--preload', type=str, multiple=True,
               help='Module that should be loaded by each worker process like "foo.bar"')
 def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
          bokeh_whitelist, prefix, use_xheaders, pid_file, scheduler_file,
-         interface, preload):
+         interface, local_directory, preload):
 
     if pid_file:
         with open(pid_file, 'w') as f:
@@ -62,6 +66,12 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
             if os.path.exists(pid_file):
                 os.remove(pid_file)
         atexit.register(del_pid_file)
+
+    local_directory = local_directory or tempfile.mkdtemp(prefix='scheduler-')
+    if not os.path.exists(local_directory):
+        os.mkdir(local_directory)
+    if local_directory not in sys.path:
+        sys.path.insert(0, local_directory)
 
     if sys.platform.startswith('linux'):
         import resource   # module fails importing on Windows
@@ -87,7 +97,9 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
             services[('bokeh', bokeh_internal_port)] = BokehScheduler
     scheduler = Scheduler(loop=loop, services=services,
                           scheduler_file=scheduler_file)
-    distributed.preloading.preload(preload, scheduler)
+    distributed.preloading.preload(preload,
+                                   parameter=scheduler,
+                                   file_dir=local_directory)
     scheduler.start(addr)
 
     bokeh_proc = None
@@ -105,6 +117,7 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
         except Exception as e:
             logger.warn("Could not start Bokeh web UI", exc_info=True)
 
+    logger.info('      Local Directory: %26s', local_directory)
     logger.info('-' * 47)
     try:
         loop.start()
@@ -113,6 +126,8 @@ def main(host, port, http_port, bokeh_port, bokeh_internal_port, show, _bokeh,
         scheduler.stop()
         if bokeh_proc:
             bokeh_proc.close()
+        if os.path.exists(local_directory):
+            shutil.rmtree(local_directory)
 
         logger.info("End scheduler at %r", addr)
 
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -11,7 +11,7 @@ from time import sleep
 
 import click
 
-import distributed
+import distributed.preloading
 from distributed import Scheduler
 from distributed.utils import ignoring, open_port, get_ip_interface
 from distributed.http import HTTPScheduler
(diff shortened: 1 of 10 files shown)


mrocklin:
Some comments:

1.  Thanks for taking the trouble to push this though the Nanny
2.  This needs tests
3.  We probably need to implement this both for the worker and for the scheduler
4.  I expect that many users will want to use a script file somewhere not on their path.  How should we handle this?
5.  There are use cases where the user will want to modify the worker or scheduler (such as @vincentschut's request to take different behavior based on worker resources.  Thoughts on how to support this?

bm371613:
1.-2. sure thing
3. I would argue it is not (as) need - scheduler does not execute users code, does it? But I don't mind.
4. If you have control over `dask-worker` invocation (the options), you have control over the path.
5. **Simple answer**: you need to pass different resource options for workers to have different resources, so you can pass different modules as well. **More convenient solution**: the worker might call a `module.__init_dask_worker__` function if present, passing some parameters or even the worker itself. That would be an extension of proposed feature, not a replacement. If you know what that interface should be like right away I can implement it now, otherwise I would leave the basic "just execute" interface in this PR, and the extension would be implemented after it is discussed by whoever needs it.

mrocklin:
> 3. I would argue it is not (as) need - scheduler does not execute user's code, does it? But I don't mind.

No argument about it not being as needed.  I do expect it to be needed though.  Having symmetry between interfaces ends up being pretty useful "X supports foo, but when I tried it on Y it didn't work."

> 4. If you have control over dask-worker invocation (the options), you have control over the path.

This may be true, but it's also very inconvenient for many users who don't understand PYTHONPATH or who don't have access to the directory where they install software.  I think that if the user provides a file with a `.py` extension then it would be useful to copy the script to the worker's temporary directory (`self.local_dir`) which we place on the path already and then import.  You might want to look at `Worker.upload_file` for some of the other cases that have come up in the past.

> the worker might call a module.__init_dask_worker__ function if present, passing some parameters or even the worker itself

This seems useful to me and may solve the problem that @vincentschut brought up.  I've also seen this situation arise in other deployments where people want to install plugins to the scheduler or workers to, for example, run extra diagnostics.

What do you think about something like the following:

```python
def dask_setup(server):
    ...

def dask_teardown(server):
    ...
```

bm371613:
OK, but when should be `dask_teardown` called? What should we promise about it?

Originally the module was supposed to be imported before the first task is executed. What happened to the worker instance did not matter, module is there as long as the process is alive, no need to tear down anything if the process is killed.

mrocklin:
Maybe we would use `atexit.register`?

mrocklin:
> no need to tear down anything if the process is killed

You might want to close files, database connections, write "worker closed" etc..

bm371613:
re out of pythonpath: with your solution, we could get `--preload /tmp/foo/resource.py --preload /tmp/bar/resource.py`. The file names clash (together with a std module too), to we probably should copy them to `self.local_dir` under random names. Is that OK?

So for `foo.bar` we promise to load `foo.bar` from python path as `foo.bar`, for `foo/bar.py` we promise only to execute it.

mrocklin:
Instead of using random names we might also raise an informative error message if there are collisions.

vincentschut:
Thanks guys, this looks very promising!

Re my bonus question: calling module.init_dask_worker(worker) would certainly do it. Or even just making sure the worker instance is present and accessible as something like 'worker = thread_state.execution_state['worker']' would already suffice to query its resources I suppose?

The path/random names thing is not so much an issue for me. I can make sure the module is present and in the python path. I'm already passing classes around anyway. It might matter for others, though.

distributed/cli/dask_scheduler.py
```diff
@@ -11,7 +11,7 @@
 
 import click
 
-import distributed
```
bm371613: that import was not used

distributed/worker.py
```diff
@@ -232,6 +234,8 @@ def start_services(self, listen_ip=''):
     def _start(self, addr_or_port=0):
         assert self.status is None
 
+        preload(self.preload, self)
```
mrocklin: We might want to call this after we call `listen` just so that the worker has a proper address.  We probably want to call this before `start_services` though.

bm371613:
Unless you have more comments, the PR is ready.

distributed/preloading.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def load_file(path):
+    """ Loads modules for a file (.py, .pyc, .zip, .egg) """
+    directory, filename = os.path.split(path)
+    name, ext = os.path.splitext(filename)
+    names_to_import = []
+    if ext in ('.py', '.pyc'):
+        if directory not in sys.path:
+            sys.path.insert(0, directory)
```
mrocklin: This is slightly dangerous.  They may have other scripts in this directory that we don't want on the path.  Instead, we might consider copying the script somewhere that *is* on the path.  This might be a worker's `local_directory` or perhaps the current working directory.  Thoughts?

distributed/preloading.py
```diff
@@ -0,0 +1,87 @@
...
+import sys
+from importlib import import_module
+
+from .compatibility import reload, invalidate_caches, cache_from_source
+
+
+logger = logging.getLogger(__name__)
+
+
+def load_file(path):
```
mrocklin: It looks like the `upload_file` method depends on this function in `preloading.py`.  Because this function is more general than just the preloading feature perhaps we should instead move it to utils.py?

distributed/tests/test_worker_client.py
```diff
@@ -1,7 +1,10 @@
 from __future__ import print_function, division, absolute_import
```
mrocklin: The worker_client testing file is for other features  (see http://distributed.readthedocs.io/en/latest/task-launch.html#submit-tasks-from-worker).  Putting these tests in their own `test_preload.py` file might be better.

distributed/tests/test_worker_client.py
```diff
@@ -165,3 +168,57 @@ def mysum():
...
+    with open(path, 'w') as f:
+        f.write(PRELOAD_TEXT)
+    with cluster(worker_kwargs={'preload': [path]}) as (s, workers),\
+            Client(s['address'], loop=loop) as c:
+
+        assert c.run(check_worker) == {
+            worker['address']: worker['address']
+            for worker in workers
+        }
+    shutil.rmtree(tmpdir)
```
mrocklin: This cleanup code should be in a try-finally block so that we clean up even if the test fails.

distributed/tests/test_worker_client.py
```diff
@@ -165,3 +168,57 @@ def mysum():
...
+
+    with cluster(worker_kwargs={'preload': ['worker_info']})\
+            as (s, workers), Client(s['address'], loop=loop) as c:
+
+        assert c.run(check_worker) == {
+            worker['address']: worker['address']
+            for worker in workers
+        }
+    sys.path.remove(tmpdir)
+    shutil.rmtree(tmpdir)
```
mrocklin: Can we also test that this works for the scheduler?

distributed/cli/dask_scheduler.py
mrocklin: Thank you for cleaning up.  We've been slowly extending flake8 testing throughout the codebase but haven't gotten here yet.

mrocklin:
This will also need a small section in the docs.  Perhaps here:

-  http://distributed.readthedocs.io/en/latest/setup.html
-  https://github.com/dask/distributed/blob/master/docs/source/setup.rst

distributed/preloading.py
bm371613: For all current uses the directory already is on the path. The only reason I added this is that this function takes care of the path for `zip` and `egg`. I will just remove this line an everything will still work, only the behaviour will be less consistent: "this function takes care of the path if you pass `zip` or `egg`, but for `py` and `pyc` you need to do it yourself.

distributed/cli/dask_scheduler.py
```diff
@@ -110,6 +122,8 @@ def del_pid_file():
         scheduler.stop()
         if bokeh_proc:
             bokeh_proc.close()
+        if os.path.exists(local_directory):
```
pitrou: If I were misunderstanding the meaning of `--local-directory` and passing an existing directory, I would probably not like Dask to delete it. Is this necessary?

distributed/preloading.py
pitrou: Perhaps simply restore the original `sys.path` when done?

distributed/preloading.py
pitrou: A name more specific than `load_file` would be welcome IMO. `import_file` perhaps?

distributed/cli/dask_scheduler.py
bm371613: This is the way `Worker` handles its local directory, I just did the same thing.

distributed/preloading.py
```diff
@@ -0,0 +1,87 @@
...
+        logger.warning("Found nothing to import from %s", filename)
+    else:
+        invalidate_caches()
+        for name in names_to_import:
+            logger.info("Reload module %s from %s file", name, ext)
+            loaded.append(reload(import_module(name)))
+    return loaded
+
+
+def preload(names, parameter=None, file_dir=None):
```
pitrou: Ditto here. `preload_modules` perhaps?

distributed/preloading.py
mrocklin: I agree that we should handle the path.  I'm concerned that we are also including everything else in that directory in the path when we may not want to.  One way to handle the path without including all other contents of the directory would be to copy the file to the server's `local_directory` attribute.  Although now that I see @pitrou 's solution that might be even simpler.

distributed/cli/dask_scheduler.py
mrocklin: It sounds like we should review what we do there generally.  You could probably avoid this issue by following @pitrou 's suggestion rather than mine and simply remove the directory of the file after importing it.

distributed/cli/tests/test_dask_scheduler.py
```diff
@@ -255,3 +254,49 @@ def test_bokeh_port_zero(loop):
...
+    if isinstance(scheduler, Scheduler):
+        print("setup")
+
+def dask_teardown(scheduler):
+    if isinstance(scheduler, Scheduler):
+        print("teardown")
+"""
+
+
+def test_preload_file(loop):
```
bm371613: Any idea why this fails on windows? I have no experience with Python on Windows. The Travis build is fine, one job failed and it seems to be unrelated (and non-deterministic).

https://ci.appveyor.com/project/daskdev/distributed/build/1.0.1040#L1412

mrocklin: I don't know precisely, but I'm not surprised.  Windows handles things differently.  Maybe a file needs to be flushed?  My recommendation is to avoid this problem by staying within Python and away from the OS.  For example you might attach a new attribute/flag to the scheduler and then run a function on the scheduler from the client to verify that it has arrived

```python
def f(dask_scheduler):
    assert dask_scheduler.my_flag

c.run_on_scheduler(f)
```

bm371613:
Thanks for help, PR ready again.

BTW some CI checks fail randomly, but you probably know that.

docs/source/setup.rst
```diff
@@ -231,3 +231,13 @@ the workers.
...
+
+Customizing initialization
+--------------------------
+
+Both ``dask-scheduler`` and ``dask-worker`` support a ``--preload`` option that
+allows custom initialization of each scheduler/worker respectively. A module
+or python file passed as a ``--preload`` value is guaranteed to be imported
+before establishing any connection. A ``dask_setup(service)`` function is called
+if found, with a ``Scheduler`` or ``Worker`` instance as the argument. As the
+service stops, ``dask_teardown(service)`` is called if present.
```
mrocklin: We might consider adding a small example here:

```
As an example, consider the following file that creates a :doc:`scheduler plugin <plugins>` and registers it with the scheduler

.. code-block:: python
   # scheduler-setup.py
   from distributed.diagnostics.plugins import SchedulerPlugin

   class MyPlugin(SchedulerPlugin):
       def add_worker(self, scheduler=None, worker=None, **kwargs):
           print("Added a new worker at", worker)
   
   def dask_setup(scheduler):
       plugin = MyPlugin(scheduler)
       scheduler.add_plugin(plugin)
   
We can then run this preload script by referring to its filename (or module name if it is on the path) when we start the scheduler::

   dask-scheduler --preload scheduler-setup.py
```

mrocklin:
This looks good to me.  I'm going to wait a while to see if @pitrou has any comments.  But I'm happy to merge afterwards.

distributed/utils.py
```diff
@@ -696,3 +700,46 @@ def open_port(host=''):
...
+        if directory not in sys.path:
+            tmp_python_path = directory
+        names_to_import.append(name)
+        # Ensures that no pyc file will be reused
+        cache_file = cache_from_source(path)
+        if os.path.exists(cache_file):
+            os.remove(cache_file)
+    if ext in ('.egg', '.zip'):
+        if path not in sys.path:
+            sys.path.insert(0, path)
```
pitrou: Should this set `tmp_python_path` as well?

pitrou: Ok, I understand this case is different (we're adding a given package to the path, not a whole directory). My bad.

pitrou:
This looks good to me too.

mrocklin:
OK, merging.  Thank you for following through with this @bm371613 .  I know that this was probably more involved than you were originally planning for.  I really like the result though, I think it will help out in a number of cases.

Issue #1020, opened 2017-04-17
Support asyncio's await interface


kszucs:
Currently distributed Client can be used with asyncio the following way:

```python
from tornado.platform.asyncio import BaseAsyncIOLoop, to_asyncio_future

ioloop = BaseAsyncIOLoop(asyncio_loop)

# init distributed client
client = Client('tcp://localhost:8786', loop=ioloop, start=False)
await to_asyncio_future(app.client._start())

dask_future = client.submit(inc, 10)
result = await to_asyncio_future(future._result())

await to_asyncio_future(app.client._shutdown())
```

It would be nice to just `await client.submit(inc, 10)` like discussed in https://github.com/channelcat/sanic/pull/644#discussion_r111733247

PR #1021, opened 2017-04-17
Support running on asyncio ioloop

mrocklin:
See #1020 

cc @kszucs @pitrou

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -26,6 +26,7 @@ from dask.core import flatten, get_dependencies
 from dask.compatibility import apply, unicode
 from dask.context import _globals
 from toolz import first, groupby, merge, valmap, keymap
+import tornado
 from tornado import gen
 from tornado.gen import Return, TimeoutError
 from tornado.locks import Event, Condition
@@ -36,7 +37,7 @@ from .batched import BatchedSend
 from .utils_comm import WrappedKey, unpack_remotedata, pack_data
 from .cfexecutor import ClientExecutor
 from .compatibility import Queue as pyQueue, Empty, isqueue
-from .core import connect, rpc, clean_exception, CommClosedError
+from .core import connect, rpc, clean_exception, CommClosedError, clean_loop
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
 from .worker import dumps_task
@@ -262,7 +263,11 @@ class Future(WrappedKey):
     __repr__ = __str__
 
     def __await__(self):
-        return self._result().__await__()
+        tornado_future = self._result()
+        if isinstance(self.client.loop, tornado.platform.asyncio.BaseAsyncIOLoop):
+            return tornado.platform.asyncio.to_asyncio_future(tornado_future).__await__()
+        else:
+            return tornado_future.__await__()
 
 
 class FutureState(object):
@@ -365,7 +370,7 @@ class Client(object):
         self.futures = dict()
         self.refcount = defaultdict(lambda: 0)
         self._should_close_loop = loop is None and start
-        self.loop = loop or IOLoop() if start else IOLoop.current()
+        self.loop = clean_loop(loop, new=start)
         self.coroutines = []
         self.id = str(uuid.uuid1())
         self.generation = 0
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -23,7 +23,7 @@ import socket
 import dask
 from dask.base import tokenize, normalize_token, Base, collections_to_dsk
 from dask.core import flatten, get_dependencies
-from dask.compatibility import apply, unicode
+from dask.compatibility import apply, unicode, PY3
 from dask.context import _globals
 from toolz import first, groupby, merge, valmap, keymap
 import tornado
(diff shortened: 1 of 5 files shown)


distributed/tests/py3_test_client.py
```diff
@@ -37,3 +38,30 @@ def test_as_completed_async_for(c, s, a, b):
...
+        await to_asyncio_future(worker._start())
+
+        client = Client(scheduler.address, loop=ioloop, start=False)
+        await to_asyncio_future(client._start())
+
+        future = client.submit(inc, 1)
+        result = await future
+        assert result == 2
+
+        await to_asyncio_future(client._shutdown())
```
kszucs: https://github.com/dask/distributed/blob/master/distributed/client.py#L659

What do You think about decorating gen.coroutine methods with an additional to_asyncio_future based on the same ioloop typecheck?

kszucs: Then we could avoid using to_syncio_future outside.

mrocklin: I think it sounds like potentially a good idea.  I probably won't pursue this myself near-term.  Is this something that you would like to try out?

kszucs:
Something like this?

Decorator:

```python
def aiocompatible(method):
    @wraps(method)
    def converter(self, *args, **kwargs):
        tornado_future = method(self, *args, **kwargs)
        if self.is_asyncio_loop():  # needs a method 
            return to_asyncio_future(tornado_future)
        else:
            return tornado_future
    return converter
```

Usage:

```python
    @aiocompatible
    @gen.coroutine
    def _start(self, timeout=3, **kwargs):
        if self.cluster is not None:
            ...
```

mrocklin:
If we're going to end up wrapping every coroutine with two decorators then maybe it makes sense to collapse these into one

```python
@coroutine
def _start(...):
    ...
```

distributed/client.py
```diff
@@ -262,7 +263,11 @@ def __str__(self):
     __repr__ = __str__
 
     def __await__(self):
-        return self._result().__await__()
+        tornado_future = self._result()
+        if isinstance(self.client.loop, tornado.platform.asyncio.BaseAsyncIOLoop):
```
mrocklin: Given that both libraries have similar terms (ioloop, future, etc.) I've chosen to be fully explicit here.

kszucs:
Then this functions should go to compatibility.py, right?
Also eliminating the need of type checking in `__a*` methods.

mrocklin:
That could make sense

mrocklin:
Planning to merge this soon if there are no objections.

kszucs:
I took a try for implementing the auto-converting `coroutine` decorator, but distributed internally expects tornado yieldables instead of asyncio futures - so a simple decorator won't solve the compatibility issue.

An easier approach would be to create an AsyncioClient. What do You think?

mrocklin:
I don't have enough practical experience with asyncio to have an informed
opinion here.

kszucs:
Me neither. Do You know someone who has?

mrocklin:
if I were you I would try things out and try to learn how best to do this.
You could also ask on stackoverflow if you can reduce the problem enough.

kszucs:
I'm doing that by asking You. Wrapping gen.coroutine methods with to_asyncio_future seems to work. Converting back and forth asyncio/tornado futures automatically prividing tornado corutine and asyncio compatible interface at the same time seems too complex and unreasonable. 

I'll create an example AsyncioClient.

mrocklin:
@pitrou if you get a chance can you look this over and merge if you think it's appropriate?

distributed/client.py
```diff
@@ -262,7 +263,11 @@ def __str__(self):
     __repr__ = __str__
 
     def __await__(self):
-        return self._result().__await__()
+        tornado_future = self._result()
+        if PY3 and isinstance(self.client.loop, tornado.platform.asyncio.BaseAsyncIOLoop):
```
pitrou: Can we assume `tornado.platform.asyncio` is already imported?

distributed/core.py
```diff
@@ -672,3 +673,27 @@ def clean_exception(exception, traceback, **kwargs):
...
+
+    if isinstance(ioloop, tornado.ioloop.IOLoop):
+        return ioloop
+    if not ioloop:
+        if new:
+            return tornado.ioloop.IOLoop()
+        else:
+            return tornado.ioloop.IOLoop.current()
+    try:
+        import asyncio
```
pitrou: I think it would be better not to import this unconditionally, as it can lengthen startup times. You could check for "asyncio" in `sys.modules`.

distributed/core.py
```diff
@@ -672,3 +673,27 @@ def clean_exception(exception, traceback, **kwargs):
...
+    except ImportError:
+        pass
+    else:
+        if isinstance(ioloop, asyncio.BaseEventLoop):
```
pitrou: If not, it returns `None`?

kszucs:
Is it still actual after providing an explicit interface for asyncio?

PR #1028, opened 2017-04-19
Minor Client interface inconsistencies

kszucs:
1. There is a couple of places where `Return` is used instead of `gen.Return`
2. `reconnect` and `ensure_connected` methods are not prefixed like the rest of the coroutines


mrocklin:
Sorry for my delay in merging this.  Thank you @kszucs !

PR #1032, opened 2017-04-25
Updated runtime requirements and changes in using the lz4 compression api

jneines:
I've had a lot of DeprecationWarnings regarding lz4 in my jupyter notebook environment and noticed that dask distributed was the source of these. Instead of simply suppressing these I thought that updating distributed to use the new api would be more sufficient.

When testing  the updated version in a new conda based environment I noticed that paramiko and bokeh were missing as necessary runtime requirements for dask-ssh and the bokeh based status pages, so I updated these requirements as well.

Hope these changes are useful.

### DIFF
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -52,12 +52,12 @@ with ignoring(ImportError):
     import lz4
 
     def _fixed_lz4_decompress(data):
-        # lz4.LZ4_uncompress() doesn't accept memoryviews
+        # lz4.block.decompress() doesn't accept memoryviews
         if isinstance(data, memoryview):
             data = data.tobytes()
-        return lz4.LZ4_uncompress(data)
+        return lz4.block.decompress(data)
 
-    compressions['lz4'] = {'compress': lz4.LZ4_compress,
+    compressions['lz4'] = {'compress': lz4.block.compress,
                            'decompress': _fixed_lz4_decompress}
     default_compression = 'lz4'
 
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -53,12 +53,12 @@ with ignoring(ImportError):
 
     try:
         # try using the new lz4 API
-        lz4_compress=lz4.block.compress
-        lz4_decompress=lz4.block.decompress
+        lz4_compress = lz4.block.compress
+        lz4_decompress = lz4.block.decompress
     except AttributeError as err:
         # fall back to old one
-        lz4_compress=lz4.LZ4_compress
-        lz4_decompress=lz4.LZ4_uncompress
+        lz4_compress = lz4.LZ4_compress
+        lz4_decompress = lz4.LZ4_uncompress
 
     def _fixed_lz4_decompress(data):
         # helper to bypass missing memoryview support in lz4
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -68,7 +68,7 @@ with ignoring(ImportError):
 
     compressions['lz4'] = {'compress': lz4_compress,
                            'decompress': _fixed_lz4_decompress}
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for updating the `lz4` function calls.  Do you know how recently this change was made in the `lz4` library by any chance?  Is it worth warning users if they have an old version?  (my guess is no)

Regarding bokeh and paramiko we leave these as optional dependencies for people who want to depend on the `distributed` library without requiring many dependencies.

I can merge in just the `compression.py` commit if you like or you can remove the `requirements.txt` commit and we can merge this PR.

jneines:
I've removed bokeh and paramiko from the runtime requirements again. I understand that they are not strictly necessary to operate distributed. I'm using a scenario depending on both on a day to day basis, so when testing the changes in a fresh environment my setup directly failed due to these packages missing. So I thought they have simply be forgotten. Should have known better ;-)

The deprecated label of the lz4 functions went toplevel end of December 2016. It has been a warning since end of may 2016. 

Gesendet mit der WEB.DE iPad App

Am 25.04.17 um 15:04 schrieb Matthew Rocklin

mrocklin:
Looks like the Python 3.4 tests will fail with this change.  Presumably there isn't a sufficiently recent package.  Perhaps this would benefit from a compatibility function that checks versions?

```python
if lz4.__version__ > = ...:
    lz4_compress = lz4.block.compress
else:
    lz4_compress = lz4....
```

jneines:
I'll have a look at it. 

Gesendet mit der WEB.DE iPad App

Am 26.04.17 um 17:13 schrieb Matthew Rocklin

jneines:
There does not  seem to be a tuple based version exported by the lz4 module. This makes a reliable version check a bit tricky. I could either rely on just comparing version strings, which would work until a "0.10.x" release, explicitly write code to convert the version string to a list of ints for better comparison, or rely on an existing implementation for this from another package, such as distutils, introducing another runtime requirement.  

My favorite would be relying on distutils, although it feels a bit odd to have it as a runtime requirement.

What do you think?

Regarding the failing Py34 tests. I've seen the py3_test_client.py script failing as the used async/await syntax has not been available in Python 3.4. Next thing to circumvent?

Gesendet mit der WEB.DE iPad App

Am 26.04.17 um 17:13 schrieb Matthew Rocklin

pitrou:
@jneines, you can just use a try/except block:
```
try:
    lz4_compress = lz4.block.compress
except AttributeError:
    lz4_compress = lz4....
```

jneines:
@pitrou That's a reasonable approach. I've updated the pull request accordingly.

distributed/protocol/compression.py
```diff
@@ -51,15 +51,24 @@ def _fixed_snappy_decompress(data):
...
         if isinstance(data, memoryview):
             data = data.tobytes()
-        return lz4.LZ4_uncompress(data)
+        return lz4_decompress(data)
 
-    compressions['lz4'] = {'compress': lz4.LZ4_compress,
+    compressions['lz4'] = {'compress': lz4_compress,
                            'decompress': _fixed_lz4_decompress}
-    default_compression = 'lz4'
+     default_compression = 'lz4'
```
pitrou: Looks like this misindent broke the CI runs.

jneines: fixed the misindent along with a notice to myself not to do online edits it github repositories using mobile devices anymore.

pitrou:
Looking at the CI results, there are a couple style issues that need resolving. Probably very minor.

distributed/protocol/compression.py
mrocklin: :)

pitrou:
Thank you! The AppVeyor seems to have hit a sporadic error, which is unrelated to this PR. I'm going to merge.

PR #1034, opened 2017-04-25
TLS support

pitrou:
The following interfaces are added to make TLS support possible:
* a `connection_args` argument to `connect()` and `listen()` to allow passing a SSLContext object
* a set of `config.yaml` entries
* a Security object that's initialized from the global config, and can be modified later
* a set of new command-line options (`--tls-ca-file`, `--tls-cert`, `--tls-key`)
* a DASK_CONFIG environment variable to use another config file than `~/.dask/config.yaml` (useful in tests)

Skipped:
- [ ] Fix Bokeh apps to use connection args
- [ ] Fix dask-submit and dask-remote to support new CLI options and pass connection args


distributed/config.yaml
```diff
@@ -17,6 +17,20 @@ multiprocessing-method: forkserver
 # Communication options
 tcp-timeout: 30         # seconds delay before calling an unresponsive connection dead
 default-scheme: tcp
+require-encryption: False   # whether to require encryption on non-local comms
```
pitrou: @jbednar, @stuartarchibald, these are the configuration file entries that I'm proposing to add. Note `ciphers` is an arbitrary cipher string in OpenSSL format.

distributed/cli/dask_scheduler.py
```diff
@@ -21,16 +22,19 @@
 
 
 @click.command()
-@click.option('--port', type=int, default=None, help="Serving port")
+@click.option('--host', type=str, default='',
+              help="URI, IP or hostname of this server")
+@click.option('--port', type=int, default=None, help="Serving port (deprecated, --host is preferred)")
+@click.option('--tls-ca-file', type=str, default=None, help="CA cert(s) file for TLS (in PEM format)")
```
pitrou: @jbednar, @stuartarchibald, these are the 3 CLI options that I'm proposing to add to `dask-scheduler` and `dask-worker`. They should override corresponding options in the config file.

mrocklin:
cc @mariusvniekerk

mrocklin:
FWIW the separate-process bokeh app is on its way out.  I also wouldn't prioritize dask-remote at all if it is expensive.  As far as I know it is almost unused.

mariusvniekerk:
I've been following this branch for a while.  It looks like it covers most of the points that we wanted to have.  

I need to take some more time to read through all of it properly.

distributed/cli/dask_scheduler.py
```diff
@@ -21,16 +22,19 @@
 
 
 @click.command()
-@click.option('--port', type=int, default=None, help="Serving port")
+@click.option('--host', type=str, default='',
+              help="URI, IP or hostname of this server")
+@click.option('--port', type=int, default=None, help="Serving port (deprecated, --host is preferred)")
```
mrocklin: FWIW I find specifying `--port` to be useful in practice.

mariusvniekerk: Agreed. Specifying both is pretty standard I think. Very useful when running multiple schedulers on one node.

distributed/config.yaml
stuartarchibald: Thanks, this looks good. Perhaps add a comment to the tail of the `ciphers` section to note the cipher string format?

distributed/cli/dask_scheduler.py
stuartarchibald: This seems good, thanks. Perhaps set the `type` as `type=click.Path` (or `File`) with appropriate args opposed to it being a string?

pitrou:
> FWIW the separate-process bokeh app is on its way out.

Is there a pending PR for this? Even if Bokeh becomes in-process, I must still adapt its code to use the required connection_args.

distributed/cli/dask_scheduler.py
pitrou: Ok, I'll remove the deprecation notice.

distributed/cli/dask_scheduler.py
pitrou: I didn't know about `click.Path`, I'll take a look.

distributed/config.yaml
pitrou: Either that, or we document it in the documentation.

mrocklin:
> Is there a pending PR for this? 

It has been removed for local cluster (`client = Client()`) and I'm using it heavily in development.  I'll remove it from `dask-scheduler` after I'm confident that all bugs have been worked out.  I expect this to happen in a month or so.  If there is a period where TLS users don't have access to Bokeh I think that that is OK.

> Even if Bokeh becomes in-process, I must still adapt its code to use the required connection_args.

Why is this?  The bokeh applications will not need to communicate to any remote server to collect diagnostic data.  The only communication is between the Bokeh server and the web clients in the browser.  My hope is that we can rely on Bokeh to handle this.

pitrou:
> Why is this? The bokeh applications will not need to communicate to any remote server to collect diagnostic data.

Oh. Yes, you're right. I thought that you would still be opening a TCP communication to the scheduler.

mrocklin:
The benefit of in-process is that we can access the data directly without figuring out how to move data around.  Bokeh and the dask scheduler are now on the same IOLoop.

docs/source/communications.rst
```diff
@@ -30,6 +30,10 @@ source tree:
 * ``tcp`` is the main transport; it uses TCP sockets and allows for IPv4
   and IPv6 addresses.
 
+* ``tls`` is a secure transport using the well-known `TLS protocol`_ over
```
pitrou: Someone may want to review the docs here and below. @stuartarchibald  perhaps?

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
+.. _configuration:
+
+=============
+Configuration
+=============
+
+As with any distributed computation system, taking full advantage of
+Dask distributed sometimes require configuring.  Some options can be passed
```
stuartarchibald: s/require configuring/requires configuration/ ?

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+as :ref:`API <api>` parameters and/or as command line options to the
```
stuartarchibald: s/as command/command/

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+
+
+Communication options
+---------------------
+
+``compression``
+"""""""""""""""
+
+This key configures the desired compression scheme when transferring data
+over the cluster.  The default value, "auto", applies heuristics to try and
```
stuartarchibald: should that be `over the network` or similar?

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+then Dask will refuse establishing any clear-text communications (for example
+over TCP without TLS), forcing you to use a secure transport such as
+:ref:`TLS <tls>`.
+
+
+``tcp-timeout``
+"""""""""""""""
+
+The default "timeout" on TCP sockets.  If a remote endpoint is unresponsive
+(at the TCP layer, not at the distributed layer) for at least this seconds,
```
stuartarchibald: s/this/the specified number of/

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+the communication is considered closed.  This helps detect endpoints that
+have been killed or disconnected abruptly.
```
stuartarchibald: s/or/or have/

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+
+``tls``
+"""""""
+
+This key configures :ref:`TLS <tls>` communications.  Several sub-keys are
+recognized:
+
+* ``ca-file`` configures the CA certificate file used to authenticate
+  and authorize all endpoints
+* ``ciphers`` restricts allowed ciphers on TLS communications
```
stuartarchibald: Perhaps mention the ciphers are in OpenSSL format and their interpreted order is that of OpenSSL (I assume it is?!)?

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+have been killed or disconnected abruptly.
+
+
+``tls``
+"""""""
+
+This key configures :ref:`TLS <tls>` communications.  Several sub-keys are
+recognized:
+
+* ``ca-file`` configures the CA certificate file used to authenticate
```
stuartarchibald: Should this section mention it must be a path to the cert/key files?

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+  and authorize all endpoints
```
stuartarchibald: Full stop at end of line.

docs/source/configuration.rst
stuartarchibald: Full stop at end of line.

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+* ``ciphers`` restricts allowed ciphers on TLS communications
+
+Each kind of endpoint has a dedicated endpoint sub-key: ``scheduler``,
+``worker`` and ``client``.  Each endpoint sub-key also supports several
+sub-keys:
+
+* ``cert`` configures the certificate file for the endpoint
```
stuartarchibald: Full stop at end of line.

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+* ``key`` configures the private key file for the endpoint
```
stuartarchibald: Full stop at end of line.

docs/source/configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+""""""""""""""""""""
+
+The number of retries before a "suspicious" task is considered bad.
+A task is considered "suspicious" if the worker died while executing it.
+
+
+``bandwidth``
+"""""""""""""
+
+The estimated bandwidth, in bytes per second, from worker to worker.  This
```
stuartarchibald: Assume this means network bandwidth?

docs/source/tls.rst
```diff
@@ -0,0 +1,58 @@
+.. _tls:
+
+
+TLS/SSL
+=======
+
+Currently dask distributed has experimental support for TLS/SSL communication,
+providing mutual authentication and encryption of communications between cluster
+endpoints (Clients, Schedulers and Workers).
```
stuartarchibald: Is it worth noting auth+encrypt is the supported mode and they are not implemented separately?

docs/source/tls.rst
```diff
@@ -0,0 +1,58 @@
...
+private key, or whether all endpoints share the same, or whether each
+endpoint kind (Client, Scheduler, Worker) gets its own certificate / key pair.
+Unless you have extraordinary requirements, however, the CA certificate
+should probably be the same for all endpoints.
+
+One can also pass additional parameters:
+
+* a set of allowed *ciphers*, if you have strong requirements as to which
+  algorithms are considered secure;  this setting's value should be an
+  `OpenSSL cipher string <https://www.openssl.org/docs/man1.1.0/apps/ciphers.html>`_;
```
stuartarchibald: Worth noting about chips with hardware support for given ciphers?

docs/source/tls.rst
```diff
@@ -0,0 +1,58 @@
...
+* whether to *require encryption*, to avoid using plain TCP communications
+  by mistake.
+
+How to pass those parameters can be done in several ways:
```
stuartarchibald: Perhaps... `The passing of these parameters can be done in several ways` ?

docs/source/tls.rst
pitrou: I think that's too specific. AES is normally selected by default and is hardware-accelerated on recent CPUs.

stuartarchibald: Apologies, I was ambiguous in my statement. What I meant was, is it worth noting that some hardware has special support for some ciphers? Not what that hardware is or what those ciphers are, just that it exists and so users may want to investigate?

pitrou: Yes. I think I've addressed all the review comments now.

docs/source/configuration.rst
pitrou: The detailed explanation for these settings is in the TLS chapter.

docs/source/tls.rst
```diff
@@ -0,0 +1,85 @@
...
+  `OpenSSL cipher string <https://www.openssl.org/docs/man1.1.0/apps/ciphers.html>`_;
+* whether to *require encryption*, to avoid using plain TCP communications
+  by mistake.
+
+All those parameters can be passed in several ways:
+
+* through the :ref:`configuration file <configuration>` ``.dask/config.yaml``;
+* if using the command line, through options to ``dask-scheduler`` and
+  ``dask-worker``;
+* if using the API, through a ``Security`` object.
```
mrocklin: It would be convenient to provide a couple of examples here

distributed/security.py
```diff
@@ -0,0 +1,139 @@
...
+        - require_encryption
+        - tls_ca_file
+        - tls_ciphers
+        - tls_client_key
+        - tls_client_cert
+        - tls_scheduler_key
+        - tls_scheduler_cert
+        - tls_worker_key
+        - tls_worker_cert
+    """
```
mrocklin: Is it likely that we will be asked to support other security systems in the future other than TLS?  If so then is the API that we're setting up here going to cause problems if we have to grow in other direction?

distributed/tests/test_tls_functional.py
```diff
@@ -0,0 +1,172 @@
...
+@gen_tls_cluster(client=True)
+def test_worker_client_executor(c, s, a, b):
+    def mysum():
+        with worker_client() as c:
+            with c.get_executor() as e:
+                return sum(e.map(double, range(30)))
+
+    future = c.submit(mysum)
+    result = yield future._result()
+    assert result == 30 * 29
```
mrocklin: I'm curious, did implementing these tests reveal issues?  I'm curious if we should anticipate other issues in the future.  What do other developers need to know when building peripheral functionality?  Are there general suggestions like "always use `self.rpc` instead of `rpc`"?

docs/source/tls.rst
```diff
@@ -0,0 +1,96 @@
...
+  by mistake.
+
+All those parameters can be passed in several ways:
+
+* through the Dask :ref:`configuration file <configuration>`;
+* if using the command line, through options to ``dask-scheduler`` and
+  ``dask-worker``;
+* if using the API, through a ``Security`` object.  For example, here is
+  how you might configure a ``Security`` object for client use::
+
```
mrocklin: Should probably use `.. code-block:: python` below.

distributed/tests/test_tls_functional.py
pitrou: I had to fix `worker_client()`, but the rest went fine. Either you use `self.rpc`, or you explicitly pass a `connection_args` argument to functions that take one.

distributed/security.py
pitrou: I don't think we're likely to support other secure transports, no. However, perhaps we'll add other security features such as permissions?
I'm not entirely sure about the `Security` object's API, but it should be decently flexible still.

docs/source/tls.rst
pitrou: Done.

stuartarchibald:
I've given this a run through testing a load of things manually. One thing I think is possibly a bug is a setup like:
```
machine1 #> dask-scheduler --host tls://ip:port --scheduler-file ~/.sched
```
```
machine2 #> dask-worker --host tls://ip:port --scheduler-file ~/.sched
```
which gives a backtrace from `machine2` (the worker) ending:
```
File "<install path>/distributed/comm/registry.py", line 71, in get_backend
    raise ValueError("unknown address scheme %s (known schemes: %s)"
ValueError: unknown address scheme '[tls' (known schemes: ['inproc', 'tcp', 'tls'])
```
Stepping with pdb, seems like the address passed through is the string `"[tls://ip:port]"` and the parser pulls the scheme from this along with the leading square bracket hence the unknown scheme as there's no match.

pitrou:
Can you give the exact `tls://ip:port` string? There may be a parsing issue somewhere.

stuartarchibald:
I've tested:
* The `DASK_CONFIG` for a global config with a number of `TLS` based options set, works as expected.
* The cipher suites used are used as specified.
* Trying to start schedulers and workers with mismatched `tls://`/`tcp://` schemes and also trying to start anything without `tls://` schemes when `require-encryption: True` is set, all of which seem to work as expected.

An observation:
* The TLS related arg names (`--tls-foo`) used in the command line args to the `dask-` launchers don't match those which are present in the `Security` object. The CA cert also has `file` in its name whereas the client cert and client key do not, I'm not sure this matters.

stuartarchibald:
In reply to https://github.com/dask/distributed/pull/1034#issuecomment-299243975: The string is like:
`--host tls://A.B.C.D:7000` there is nothing unusual about it that I can see.

pitrou:
About the TLS arg names: when invoking a specific `dask-` launcher, Dask knows for which role those arguments are (for example scheduler or worker). The Security object can get option for all roles (especially when filled from the config file), and people might want to emit distinct certs and keys for client, worker or scheduler nodes.

As for `tls-ca-file`, OpenSSL also allows a directory format for CA certificates, so this is a provision in case we want to support that format.

stuartarchibald:
Thanks for the explanation in https://github.com/dask/distributed/pull/1034#issuecomment-299248223, good points. Happy to leave as is.

pitrou:
I've pushed a fix for the `dask-worker --host tls://ip:port` error. However, dask-worker is currently confusing as there are different options impacting this (`--worker-port`, `--nanny-port`, `--host`, `--interface`) and how they interact is not clear. We should improve this in a further PR.

mrocklin:
Should we merge this?

pitrou:
I'm ok with merging, though you may want to ask other people to try it out if they have time.

mrocklin:
@minrk in this PR we add TLS/SSL support to Dask.  Do you have thoughts on how to best integrate this cleanly into the `start_ipython` methods?  (also, any feedback you have on this PR would be welcome.)

mrocklin:
Do we know anyone else who is likely to test this within their system?

pitrou:
> Do we know anyone else who is likely to test this within their system?

I have no idea. I expect there would be a significant amount of users interested in this, still I don't know of any specific people to ask. Perhaps we should mention this PR on a community channel?

mrocklin:
I've mentioned it in gitter a couple of times.  Any objection to my tweeting about it?

pitrou:
No objection.

bkreider:
I don't really use dask, but I'm very excited by this TLS support.  Reading through this it looks like all the right questions were asked (and answered).  

These are probably dumb questions:

1.  Does this support client certs?  ie: Server certs encrypt transport and identify the server to the client.  A client cert would provide authentication and authorization as well.

2. Do we have a crypto export notification in the project?  I know that Anaconda has an EAR crypto declaration, but that only covers anaconda.  Obviously, IANAL so you should get appropriate advice.  

https://www.thoughtworks.com/insights/blog/encryption-open-source-and-export-control
> Under Section 740.13(e) of License Exception TSU, publicly available encryption source code may be exported without a license, so long as the notification requirement is met (and updated accordingly). This exception is not limited only to those who distribute their software under an open-source license for free, it is also applies to code that is licensed for a fee or royalty. Hence, making your source code publicly available has the double benefit of simplifying your compliance with the EAR and making your software safer and more trustworthy, since anyone can examine it to ensure there are no mistakes or backdoors.

Note that the US laws are very broad according to the link above:

> The EAR purports to apply to all people, anywhere on earth. The EAR considers anything that was developed in the US, incorporates technology developed in the US, or is transshipped through the US to be “US-origin.”

pitrou:
> Does this support client certs?

Not only, but it mandates them, i.e. both ends have to be authenticated to each other using TLS certs validated by the CA cert.

> Do we have a crypto export notification in the project?

No. I'm unsure how to ask about this, or whether it's necessary at all.

bkreider:
> Not only, but it mandates them, i.e. both ends have to be authenticated to each other using TLS certs validated by the CA cert.

Very nice.

mrocklin:
@teoliphant mentioned that because all this does is `import ssl` it may not be necessary to be concerned about export control.

mrocklin:
I'm inclined to merge this soon.  I think that we've gotten good feedback from a different groups with different needs.  This feedback seems to be fairly positive with minor edits.  I suspect that we'll get more feedback in the future if we merge.

mrocklin:
@pitrou thoughts?

pitrou:
I'll need to fix the current merge conflicts, but otherwise +1 for merging.

mrocklin:
I'm looking into the intermittent test failures now

mrocklin:
I've been able to reliably trigger `test_tls_functional.py::test_work_stealing` but haven't been able to trigger the same test in `test_steal.py`.  If we merge this then we're probably adding another significant intermittent test.  Looking now at ways to relax this test.

mrocklin:
Merging this later today if there are no further comments.

mrocklin:
Thanks @pitrou for diving into this and thanks to others for reviewing and providing feedback.

PR #1036, opened 2017-04-26
Handling repeated signals in cli

mrocklin:
This uses the `IOLoop.add_callback_from_signal` method rather than `IOLoop.add_callback` method when running from a signal handler.  

This also simply exits when calling a signal handler twice.

However I'm not sure how best to test this.  Playing around interactively I wasn't able to convince myself that the `sys.exit` branch was ever taken.

cc @pitrou for advice


mrocklin:
I've merged in just the first commit here.  Closing.

PR #1038, opened 2017-04-26
Reevaluate occupancy periodically

mrocklin:
Our estimate of task runtimes can change.  This can throw off our ability to
worksteal considerably.  Unfortunately it's not clear that there is a constant
time way to update the relevant state.

This PR experiments with an alternative.  We periodically scan through our
worker state when we have excess cpu power.   This lets us do non-constant-time
updates while still staying within performance bounds.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -445,6 +445,8 @@ class Scheduler(Server):
 
             finalize(self, del_scheduler_file)
 
+        self.loop.add_callback(self.reevaluate_occupancy)
+
         return self.finished()
 
     @gen.coroutine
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -3022,6 +3024,76 @@ def worker_objective(self, key, worker):
...
+
+                    duration = time() - last
+                    if duration > 0.005:  # 5ms since last release
+                        yield gen.sleep(duration * 5)  # 25ms gap
+                        last = time()
+
+    def _reevaluate_occupancy_worker(self, worker):
+        """ See reevaluate_occupancy """
+        w = worker
+        processing = self.processing[w]
```
pitrou: Can this raise KeyError if the worker was removed during `yield gen.sleep(0.100)`?

distributed/scheduler.py
```diff
@@ -3022,6 +3024,76 @@ def worker_objective(self, key, worker):
...
+        for key in processing:
+            duration = self.task_duration.get(key_split(key), 0.5)
+            processing[key] = duration
+            new += duration
+            for dep in self.dependencies[key]:
+                if dep not in self.has_what[w]:
+                    nbytes += self.nbytes.get(key, 0)
+
+        comm = nbytes / BANDWIDTH
+        self.occupancy[w] = max(new, comm)
```
pitrou: I'm not sure why communication duration is taken into account here? Can't other tasks run while the dependencies are being fetched?

distributed/scheduler.py
```diff
@@ -3022,6 +3024,76 @@ def worker_objective(self, key, worker):
...
+        self.total_occupancy += new - old
+        self.check_idle_saturated(w)
+
+        if new > old * 1.5:
```
pitrou: It would be nice to explain that heuristic :-)

distributed/scheduler.py
mrocklin: Yes, this overlapping ability is why I chose `max` rather than `sum`.  Not taking into account communication costs when they exceed computation costs can result in overly-aggressive stealing.

distributed/scheduler.py
mrocklin: Added a short comment, but mostly the code under this line can be expensive.  We don't want it to happen each time.

distributed/scheduler.py
mrocklin: Fixed

mrocklin:
Hrm, an alternative to this PR would be to do the necessary changes on significant changes to task_duration.  Presumably the number of changes that increase a task duration by say 20% are rather limited.

mrocklin:
I would like to release before merging this.  I like the approach here but am somewhat cautious about it.  Apologies for the delay.

PR #1043, opened 2017-04-27
Revert addition of terminate to joblib backend

jcrist:
Due to how joblib calls terminate, this can fail if you ever call
`Parallel` twice inside the `parallel_backend` contextmanager. Until
this is fixed in joblib, we'll need to let the gc cleanup handle
shutdown.


jcrist:
Related to https://github.com/joblib/joblib/issues/335.

mrocklin:
Fine with me.

Issue #1031, opened 2017-04-24
Long list of socket errors after client shutdown


bschreck:
If i initialize a client with a LocalCluster and then shut it down like this:
    c = Client()
    c.shutdown()

I get many repeated errors that look like:

distributed.comm.tcp - WARNING - Closing dangling stream in <TCP 'tcp://127.0.0.1:57338'>
Traceback (most recent call last):
  File "/Users/bschreck/FeatureLabs/featuretools/ENV/lib/python2.7/site-packages/distributed/compatibility.py", line 216, in _exitfunc
    f()
  File "/Users/bschreck/FeatureLabs/featuretools/ENV/lib/python2.7/site-packages/distributed/compatibility.py", line 140, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "/Users/bschreck/FeatureLabs/featuretools/ENV/lib/python2.7/site-packages/distributed/comm/tcp.py", line 127, in finalize
    stream.close()
  File "/Users/bschreck/FeatureLabs/featuretools/ENV/lib/python2.7/site-packages/tornado/iostream.py", line 434, in close
    self.io_loop.remove_handler(self.fileno())
  File "/Users/bschreck/FeatureLabs/featuretools/ENV/lib/python2.7/site-packages/tornado/ioloop.py", line 732, in remove_handler
    fd, obj = self.split_fd(fd)
  File "/Users/bschreck/FeatureLabs/featuretools/ENV/lib/python2.7/site-packages/tornado/ioloop.py", line 652, in split_fd
    return fd.fileno(), fd
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 174, in _dummy
    raise error(EBADF, 'Bad file descriptor')
socket.error: [Errno 9] Bad file descriptor

I tried figuring out exactly why but not sure. It's very annoying when I write test cases that need to shutdown the cluster. If I wrap the line it refers to in comm/tcp.py with a try/except socket.error block these go away. I'm not sure if there's actually something to do to prevent the errors from cropping up in the first place, or if the best solution is just to try/except

mrocklin:
Yeah, I've been seeing these in tests as well.  FWIW these are not present
on Python 3.  We should probably figure out what's going on regardless
though.  Does anyone have experience on what this might be?

[...]

Issue #1042, opened 2017-04-27
Local cluster fails to close on Client.shutdown()


bschreck:
I assume this is a bug but maybe it's what's supposed to happen. When I initialize a Client with no arguments (so that it initializes a LocalCluster as well), and then call shutdown(), the cluster does not shutdown properly. This is because the client shuts down it's IOLoop first, and then at the end tries to close the cluster. The cluster's close method checks if its IOLoop is running before actually closing.

However, moving the cluster.close() call to the start of Client.shutdown() does trick:

    def shutdown(self, timeout=10):
                if self.status == 'closed':
                    return
        # call this before closing loop (cluster runs on same ioloop as self)
        with ignoring(AttributeError):
            self.cluster.close()
        sync(self.loop, self._shutdown, fast=True)
        assert self.status == 'closed'
        ...

mrocklin:
Not shutting down the cluster when calling Client().shutdown() seems like a
bug to me.  Any interest in submitting a PR?

PR #1045, opened 2017-04-28
Client shutdown cluster properly & suppress tornado errors

bschreck:
Fixes issues #1031 and #1042 

I'm not sure how to test because the LocalCluster.close() method is implemented as coroutine callbacks which are not guaranteed to finish by the end of the Client.shutdown() method. I found two tests in test_client that are supposed to test shutdown but they both don't load the LocalCluster and so don't test to see if it shut down.


mrocklin:
Hey @bschreck, thanks for working on this.  

I suspect that the `all_fds` flag is required to properly clean up file descriptors in tests quickly enough so that we don't overwhelm the testing machine.  Any suggestions?

bschreck:
Hmm yeah I guess that issue can just go unresolved until a solution that passes the test is figured out. I'll just remove that change for now.

mrocklin:
Sorry, looks like the error is still there.  My analysis with all_fds may not have been accurate.

mrocklin:
This has been merged in https://github.com/dask/distributed/commit/e9357c9d84b9af960eb9413cb2509a606cd12dce

PR #1052, opened 2017-05-02
Add __aenter__ and __aexit__ methods to Client

mrocklin:
Companion to #1029 

cc @kszucs I'm trying to see if I can trigger the errors you get with a normal tornado solution.  Also, stealing some of your logic back into the main client codebase.

### DIFF
--- a/distributed/tests/py3_test_client.py
+++ b/distributed/tests/py3_test_client.py
@@ -40,18 +40,19 @@ def test_as_completed_async_for(c, s, a, b):
 
 
 def test_async_with(loop):
-    results = []
-    ns = {}
+    result = None
+    client = None
+    cluster = None
     async def f():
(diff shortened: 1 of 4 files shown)


distributed/tests/py3_test_client.py
```diff
@@ -37,3 +37,21 @@ def test_as_completed_async_for(c, s, a, b):
     yield f()
 
     assert set(results) == set(range(1, 11))
+
+
+def test_async_with(loop):
+    results = []
```
pitrou: Since this is a Python 3-only test file, you don't need these tricks and can simply use `nonlocal` variables instead.

mrocklin:
Merging this soon.  Will then rebase https://github.com/dask/distributed/pull/1029 on top of it and then merge that in as well hopefully.

PR #1053, opened 2017-05-03
Pre release fixes

mrocklin:


distributed/worker.py
```diff
@@ -161,8 +160,8 @@ def heartbeat(self):
             self.heartbeat_active = True
             logger.debug("Heartbeat: %s" % self.address)
             try:
-                if proc:
-                    memory_info = proc.memory_info()
+                if psutil:
+                    memory_info = psutil.Process().memory_info()
```
pitrou: I think instantiating a new Process object each time this method is called is a bit wasteful. Doing it in _start() would be sufficient.

mrocklin: The costs are fairly well below our current threshold.  Also this happens on the workers where computation resources are plentiful and it only happens every second or so.

```python
In [1]: import psutil

In [2]: %time _ = psutil.Process()
CPU times: user 514 µs, sys: 21 µs, total: 535 µs
Wall time: 356 µs

In [3]: %time for i in range(1000): psutil.Process()
CPU times: user 23 ms, sys: 11.8 ms, total: 34.8 ms
Wall time: 31.7 ms
```

pitrou: I was worried that it might be slower on other platforms but it seems fine on Windows as well:
```
>>> %timeit _ = psutil.Process()
The slowest run took 11.32 times longer than the fastest. This could mean that a
n intermediate result is being cached.
100000 loops, best of 3: 6.76 µs per loop
```

PR #1055, opened 2017-05-04
Close things to avoid logging errors

mrocklin:
Previously closing down a LocalCluster would result in a number of
errors being emitted to the console.  Now we close things in a different
order and check closing status to avoid these errors.

I tried to test this for a long while but failed to capture logging
information in pytest.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -705,6 +705,7 @@ class Client(object):
         if self.status == 'closed':
             return
 
+        self.status = 'closing'
         if self._start_arg is None:
             with ignoring(AttributeError):
                 self.cluster.close()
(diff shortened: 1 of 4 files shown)


pitrou:
> I tried to test this for a long while but failed to capture logging information in pytest.

`captured_logger()` in `utils_test` should help you with that.

mrocklin:
I've added a test with captured_logger, however it doesn't fail previous to this PR.  This is the case with my other attempts as well.  I've tried the pytest `capsys` fixture as well as creating a logging Handler that captures emitted messages.

pitrou:
> I've added a test with captured_logger, however it doesn't fail previous to this PR.

Does the log get printed out? Perhaps capturing doesn't work, for whatever reason.

mrocklin:
> Does the log get printed out? Perhaps capturing doesn't work, for whatever reason.

My first guess is that pytest is doing something magical here.

pitrou:
Apart from silencing output when a test is successful (which you can disable with `-s`), I'm not sure it does. `captured_logger` is used in another test.

mrocklin:
I plan to merge this soon even without proper testing.  My apologies.

PR #1065, opened 2017-05-08
Add test to check that tasks restart after all worker loss

mrocklin:
Trying to track down a reported bug.


mrocklin:

PR #1067, opened 2017-05-08
Update `keras` serializer for `tensorflow.contrib` version

bnaul:
`keras` now comes bundled as part of the default `tensorflow` install via `tf.contrib.keras`: models instantiated that way look like `tensorflow.contrib.keras.python.keras.models.Sequential`.

Perhaps there's a better way to handle this? All six of the calls to `register_serialization` here are for classes that have basically the same API.

### DIFF
--- a/distributed/protocol/keras.py
+++ b/distributed/protocol/keras.py
@@ -31,9 +31,7 @@ def deserialize_keras_model(header, frames):
     return model
 
 
-register_serialization('keras.engine.training.Model', serialize_keras_model,
-                       deserialize_keras_model)
-register_serialization('keras.models.Model', serialize_keras_model,
-                       deserialize_keras_model)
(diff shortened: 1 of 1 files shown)


jcrist:
> Perhaps there's a better way to handle this? All six of the calls to register_serialization here are for classes that have basically the same API.

Perhaps loop over a list of all class names? Alternatively, could modify `register_serialization` to take a list of names and do the looping there instead.

mrocklin:
Thanks for doing this @bnaul .  Are there particular tensorflow or keras version constraints that we should be aware of?

bnaul:
`keras` was added to `tf.contrib` in `tensorflow==1.1.0`; at the moment it's just an exact copy of `keras` except for the `theano` functionality. If you have a lower version of `tensorflow` then this module won't exist but that's fine.

PR #1071, opened 2017-05-09
fix StopIteration behaviour in AsCompleted.batches

### LINKED ISSUE #1069: PEP 479 compatibility
[PEP 479](https://www.python.org/dev/peps/pep-0479/) changes style of exit from a generator. Basically, instead of
```python
raise StopIteration
```
one should
```
return
```
Currently, in Python 3.6 `AsCompleted.batches()` generator causes:
> DeprecationWarning: generator 'AsCompleted.batches' raised StopIteration

In Python 3.7 this will be an error, not a warning.

slavoutich:
PEP 479: generators should `return` instead of `raise StopIteration`

Fixes #1069

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2693,7 +2693,10 @@ class AsCompleted(object):
         [6]
         """
         while True:
-            yield self.next_batch(block=True)
+            try:
+                yield self.next_batch(block=True)
+            except StopIteration:
+                return
 
 
 as_completed = AsCompleted


pitrou:
Wouldn't it better to fix `next_batch` to return an empty list instead of raising StopIteration?

slavoutich:
It's rather matter of taste. Either we catch exception, or check next_batch for length. Exception is catched once, and `len(self.next_batch()) > 0` will be checked every iteration.

pitrou:
`next_batch` isn't an iterator, so I wouldn't expect it to raise `StopIteration` in any case. Either return an empty list, or raise an error such as `ValueError` or `EOFError`.

slavoutich:
I'd prefer raising in that case, but I think that `ValueError` or `EOFError` are even worse here than `StopIteration`. `GeneratorExit` might be better.

pitrou:
`GeneratorExit`, as `StopIteration`, is not supposed to be raised by normal functions.  Actually `GeneratorExit` is not even an `Exception` subclass, so an `except Exception` clause won't catch it...

slavoutich:
Correct. Anyway, I would prefer either stay with `StopIteration`, or create custom exception, otherwise people can accidentally enter infinite loop, when use `AsCompleted.next_batch()` directly.

mrocklin:
@pitrou ok to merge?  Are there any tests that would be appropriate here?

pitrou:
I don't think it's easy to test this, except perhaps try to catch and check warnings on 3.6+. I'm going to merge.

PR #1077, opened 2017-05-12
check for len(workers) in Scheduler.retire_workers

slavoutich:
In some scenarios workers set in retire_workers function gets empty,
    which then causes an exception in set.union, which gets no
    arguments. This commit adds check for number of workers in that
    place.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1791,8 +1791,11 @@ class Scheduler(Server):
                         pass
 
             workers = set(workers)
-            keys = set.union(*[self.has_what[w] for w in workers])
-            keys = {k for k in keys if self.who_has[k].issubset(workers)}
+            if len(workers) > 0:
+                keys = set.union(*[self.has_what[w] for w in workers])
+                keys = {k for k in keys if self.who_has[k].issubset(workers)}
+            else:
+                keys = set()
 
             other_workers = set(self.worker_info) - workers
             if keys:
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -1791,8 +1791,11 @@ def retire_workers(self, comm=None, workers=None, remove=True, close=False,
                         pass
 
             workers = set(workers)
-            keys = set.union(*[self.has_what[w] for w in workers])
-            keys = {k for k in keys if self.who_has[k].issubset(workers)}
+            if len(workers) > 0:
+                keys = set.union(*[self.has_what[w] for w in workers])
+                keys = {k for k in keys if self.who_has[k].issubset(workers)}
+            else:
+                keys = {}
```
mrocklin: Note that the literal `{}` is a dict.  You may want `set()` instead.

slavoutich: Ah yeah, sorry, will fix in a minute.

slavoutich: Running tests now, if Ok -- will push with test also.

mrocklin:
I think that all this is missing is a small test.  @slavoutich was the comment above enough information to help you with this?

mrocklin:
> Sorry for delay, at the conference now. Should be fine now.

Thank you for following through.  I've modified the test slightly and pushed.  Will merge when tests pass.

mrocklin:
This is in.  Thanks @slavoutich !

PR #1081, opened 2017-05-12
Implement manual timeout for popen wait

mrocklin:
This resolves the test_bokeh_zero_port intermittent error.  In Python
2.7 we Process.wait() lacks a timeout= keyword argument.  We implement
this manually with Process.poll and a sleep-while loop.

Additionally we add a small sleep in the bokeh-zero-port test to
minimize the frequency of the process hang (the original cause of which
remains unknown).


mrocklin:

PR #1086, opened 2017-05-18
Move import of `infer_storage_options`

jcrist:
`dask.utils.infer_storage_options` moved to `dask.bytes.utils.infer_storage_options`. We use a try/except block to remain compatible with older dask releases. See https://github.com/dask/dask/pull/2346.

Also deleted unused logger in `hdfs.py`.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1700,10 +1700,10 @@ class Client(Node):
 
         Examples
         --------
-        >>> from dask import do, value
+        >>> from dask import delayed
         >>> from operator import add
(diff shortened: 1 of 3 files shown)


martindurant:
makes sense.

jcrist:
Unrelated, is there a reason for this being in `distributed` instead of in `dask.bytes.hdfs` (besides development history)?

martindurant:
Partly historical, but HDFS is different since it rarely makes sense in a single-machine setting, and there is some special code to find out where in the cluster the data blocks live.

PR #1092, opened 2017-05-22
AsyncIO API cleanups

ajdavis:


distributed/asyncio.py
```diff
@@ -72,43 +61,45 @@ class AioClient(Client):
...
-    The AioClient can receive an asyncio loop explicitly:
-
-    >>> loop = asyncio.get_event_loop()  # doctest: +SKIP
-    >>> client = AioClient('127.0.0.1:8786', loop=loop)  # doctest: +SKIP
+    Provide cluster's scheduler address on initialization::
 
-    Start the client:
+        client = AioClient('127.0.0.1:8786')
 
-    >>> await client.start()  # doctest: +SKIP
```
ajdavis: Not valid syntax. I decided to show code blocks instead of an interactive session.

distributed/asyncio.py
```diff
@@ -117,17 +108,15 @@ class AioClient(Client):
     """
     _Future = AioFuture
 
-    def __init__(self, *args, loop=None, start=True, set_as_default=False,
-                 **kwargs):
```
ajdavis: If you're using this from an asyncio application, it's your job to start the loop, not AioClient's. Furthermore, both asyncio and Tornado are rapidly moving away from this idiom where you pass a "loop" argument. Tornado just deleted all "ioloop" arguments on master, and Yury said yesterday that asyncio is deleting most "loop" arguments circa Python 3.7.

distributed/asyncio.py
```diff
@@ -146,11 +135,9 @@ def __del__(self):
         if self.status == 'running':
             return
 
-        if self._make_current:
-            self.loop.make_current()
+        self.status = 'running'
```
ajdavis: Set status before starting to connect. Fixes race if several coroutines await client.start() with the same client concurrently.

distributed/asyncio.py
```diff
@@ -166,14 +153,6 @@ def __del__(self):
             await to_asyncio_future(future)
             self.cluster.status = 'closed'
 
-        if self._make_current:
-            IOLoop.clear_current()
-
-    async def run_coroutine(self, function, *args, **kwargs):
-        tornfn = to_tornado(function)
-        future = self._run_coroutine(tornfn, *args, **kwargs)
-        return await to_asyncio_future(future)
```
ajdavis: Duplicated below.

distributed/asyncio.py
```diff
@@ -191,6 +170,9 @@ def __del__(self):
     replicate = to_asyncio(Client._replicate)
     start_ipython_workers = to_asyncio(Client._start_ipython_workers)
 
+    def __enter__(self):
+        raise RuntimeError("Use AioClient in an 'async with' block, not 'with'")
```
ajdavis: Prevents a deadlock if someone does "with AioClient() as client" and it calls the base class's "__enter__".

mrocklin:
cc @kszucs @pitrou

ajdavis:
I've reproduced the new Travis failure on a local Linux VM:

```
    def pytest_pyfunc_call(pyfuncitem):
        """
        Run coroutines in an event loop instead of a normal function call.
        """
        fast = pyfuncitem.config.getoption("--fast")
        if asyncio.iscoroutinefunction(pyfuncitem.function):
            existing_loop = pyfuncitem.funcargs.get('loop', None)
            with _passthrough_loop_context(existing_loop, fast=fast) as _loop:
                testargs = {arg: pyfuncitem.funcargs[arg]
                            for arg in pyfuncitem._fixtureinfo.argnames}

>               task = _loop.create_task(pyfuncitem.obj(**testargs))
E               AttributeError: 'EPollIOLoop' object has no attribute 'create_task'
```

Doesn't happen on Mac? I'll investigate.

distributed/asyncio.py
```diff
@@ -127,10 +127,6 @@ def __init__(self, *args, **kwargs):
     async def __aexit__(self, type, value, traceback):
         await self.shutdown()
 
-    def __del__(self):
-        if self.status == 'running':
-            self.loop.asyncio_loop.run_until_complete(self.shutdown(fast=True))
```
ajdavis: This can't work. If AioClient is used in an asyncio application, the loop is always running, so calling run_until_complete will raise an exception. It's the user's job to use "async with" or otherwise guarantee shutdown.

distributed/nanny.py
```diff
@@ -151,8 +151,7 @@ def _kill(self, comm=None, timeout=10):
             except CommClosedError:
                 pass
             except BaseException as e:
-                if (not self.loop._closing and self.loop._running and not
-                        self.loop._stopped):
+                if self.loop._running:
```
ajdavis: I hope this is ok, I'm not sure why the "if" had three checks beforehand. The AioLoop shim only supports "_running", and it seems better to only check "_running" here, rather than add shims for "_closing" and "_stopped".

distributed/tests/py3_test_asyncio.py
```diff
@@ -12,11 +12,11 @@
 from distributed.utils_test import slowinc
 
 from distributed.asyncio import AioClient, AioFuture, as_completed, wait
-from distributed.utils_test import inc, div, loop
+from distributed.utils_test import inc, div
 
 
-async def test_asyncio_start_shutdown(loop):
-    c = AioClient(loop=loop, processes=False)
+async def test_asyncio_start_shutdown():
```
ajdavis: AioClient no longer takes a "loop" parameter, it just uses the current asyncio event loop.

kszucs:
LGTM Thanks for the fixes @ajdavis

distributed/asyncio.py
pitrou: Hmm... This raises the question: can `__del__` even be called if some AioClient coroutines are still running?
In other words, perhaps we can put a `assert self.status != 'running'` in `__del__`?

distributed/nanny.py
pitrou: Yes, reducing use of private APIs is always good.

distributed/asyncio.py
pitrou: I agree that explicitly passing a loop object smells like a bad idiom. If a current thread's loop is not adequate, then perhaps the loop setup is wrong.

distributed/asyncio.py
ajdavis: Here's an experiment that does what you propose:

https://github.com/ajdavis/distributed/pull/1

It passes Travis and has no AssertionErrors logged from destructors. Is it worth doing? Does adding a `__del__` to a class in Python 3 put any more pressure on the garbage collector or is that problem completely solved?

pitrou: Thanks for the experiment. After thinking a bit about it, it would probably be overkill.

distributed/asyncio.py
```diff
@@ -166,13 +149,7 @@ def __del__(self):
...
             self.cluster.status = 'closed'
 
-        if self._make_current:
-            IOLoop.clear_current()
-
-    async def run_coroutine(self, function, *args, **kwargs):
-        tornfn = to_tornado(function)
-        future = self._run_coroutine(tornfn, *args, **kwargs)
-        return await to_asyncio_future(future)
+        BaseAsyncIOLoop.clear_current()
```
pitrou: Is it worth having a `try...finally` block so that `clear_current` is always called?

ajdavis: Done.

ajdavis:
New patch, adds a "try-finally" block to `AioClient.shutdown` to ensure the shim IOLoop is always uninstalled.

pitrou:
Thank you! I'm merging this now.

PR #1094, opened 2017-05-23
Serialize more dtypes

pitrou:
Structured dtypes (kind 'V') can have additional complication which require more care than is currently done.

Also remove unnecessary `stride_tricks` calls.

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -1,7 +1,6 @@
 from __future__ import print_function, division, absolute_import
 
 import numpy as np
-from numpy.lib import stride_tricks
 
 try:
     import blosc
@@ -52,16 +51,17 @@ def serialize_numpy_ndarray(x):
         dt = (0, x.dtype.str)
 
     if not x.shape:
+        # 0d array
         strides = x.strides
         data = x.ravel().view('u1').data
-    elif np.isfortran(x):
+    elif x.flags.c_contiguous or x.flags.f_contiguous:
+        # Avoid a copy and respect order when unserializing
         strides = x.strides
-        data = stride_tricks.as_strided(x, shape=(np.prod(x.shape),),
-                                           strides=(x.dtype.itemsize,)).view('u1').data
+        data = x.ravel(order='K').view('u1').data
     else:
         x = np.ascontiguousarray(x)
         strides = x.strides
-        data = x.ravel().view('u1').data
+        data = x.view('u1').data
 
     header = {'dtype': dt,
               'shape': x.shape,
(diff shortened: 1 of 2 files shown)


pitrou:
@mrocklin

distributed/protocol/numpy.py
```diff
@@ -33,22 +32,36 @@ def serialize_numpy_ndarray(x):
         frames = [pickle.dumps(x)]
         return header, frames
 
+    # We cannot blindly pickle the dtype as some may fail pickling,
+    # so we have a mixture of strategies.
```
mrocklin: Does cloudpickle work in all cases?  (might be simpler if so)

pitrou: cloudpickle doesn't have specific code for dtypes, so I doubt so.

mrocklin: Should we also raise an issue upstream in numpy?

distributed/protocol/numpy.py
```diff
@@ -33,22 +32,36 @@ def serialize_numpy_ndarray(x):
...
+        # Preserving all the information works best when pickling
+        try:
+            # Only use stdlib pickle as cloudpickle is slow when failing
+            # (microseconds instead of nanoseconds)
+            dt = (1, pickle.pickle.dumps(x.dtype))
+            pickle.loads(dt[1])  # does it unpickle fine?
+        except Exception:
+            # dtype fails pickling => fall back on the descr if reasonable.
+            if x.dtype.type is not np.void or x.dtype.alignment != 1:
+                raise
```
mrocklin: Does this occur?  Are there known cases where our serialization fails?

distributed/protocol/numpy.py
```diff
@@ -33,22 +32,36 @@ def serialize_numpy_ndarray(x):
...
+            else:
+                dt = (0, x.dtype.descr)
     else:
-        dt = x.dtype.str
+        dt = (0, x.dtype.str)
```
mrocklin: Is this effectively a boolean?  If so should we use a bool?  Or is the thought here that we might have to have other schemes in the future?

mrocklin:
I'm generally +1 on this.  I have some questions, but they shouldn't stop merging.

distributed/protocol/numpy.py
pitrou: https://github.com/numpy/numpy/issues/9159

distributed/protocol/numpy.py
pitrou: Yes, for example: `np.dtype([('a', 'i1'), ('b', 'm8')], align=True)`

distributed/protocol/numpy.py
pitrou: Yes, it's a boolean. I don't think it matters much but we can switch to a boolean if you prefer.

PR #1095, opened 2017-05-23
Fix DeprecationWarnings and ResourceWarnings

pitrou:
Recent py.test versions display Python warnings at the end of a test run, which helps us chase those.


pitrou:
Note some of those warnings can be due to Bokeh (https://github.com/bokeh/bokeh/issues/6317) or cloudpickle (https://github.com/cloudpipe/cloudpickle/pull/88).

Issue #1088, opened 2017-05-19
Memory error on worker can cause dropped keys


jcrist:
Memory errors on worker can cause a key to be dropped from `Worker.data`, resulting in further computation to use the literal key instead of the referenced data. See https://github.com/dask/dask/issues/2361.

mrocklin:
This is probably on me to fix.  If I recall correctly I think that this issue occurs when dumping a piece of data to disk fails.  I think that we silently fail in this case.  Instead we should probably propagate out some informative exception to all dependent tasks.

mrocklin:
(although others are certainly welcome to play with this)

mrocklin:
This reproduces a similar issue:

```python
class Bad(object):
    def __getstate__(self):
        raise Exception()

    def __sizeof__(self):
        return int(100e9)

from dask.distributed import Client
Client().submit(future, Bad).result()
```

jakirkham:
> If I recall correctly I think that this issue occurs when dumping a piece of data to disk fails.

Yeah, in the xref'd issue above I had been writing out results to disk.

mrocklin:
This is being resolved in #1096

PR #1096, opened 2017-05-25
Worker propagates errors when writing to disk

mrocklin:
Fixes https://github.com/dask/distributed/issues/1088

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -772,3 +772,26 @@ def test_fail_write_to_disk(c, s, a, b):
     futures = c.map(inc, range(10))
     results = yield c._gather(futures)
     assert results == list(map(inc, range(10)))
+
+
+@gen_cluster(client=True, worker_kwargs={'memory_limit': 1000})
+def test_fail_write_many_to_disk(c, s, a, b):
+    a.validate = False
+    b.validate = False
+    class Bad(object):
+        def __init__(self, x):
+            pass
+
+        def __getstate__(self):
+            raise TypeError()
+
+        def __sizeof__(self):
+            return 500
+
+    futures = c.map(Bad, range(10))
+    future = c.submit(lambda *args: 123, *futures)
+
+    yield _wait(future)
+
+    with pytest.raises(Exception) as info:
+        yield future._result()
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -768,3 +768,7 @@ def test_fail_write_to_disk(c, s, a, b):
 
     with pytest.raises(TypeError):
         yield future._result()
+
+    futures = c.map(inc, range(10))
+    results = yield c._gather(futures)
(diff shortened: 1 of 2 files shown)


distributed/tests/test_worker.py
```diff
@@ -736,3 +736,21 @@ def test_worker_dir(c, s, a, b):
...
+        def __sizeof__(self):
+            return int(100e9)
+
+    future = c.submit(Bad)
+    yield _wait(future)
+
+    assert future.status == 'error'
+
+    with pytest.raises(TypeError):
+        yield future._result()
```
pitrou: Perhaps test that we can submit another task after this one?

mrocklin: Done

mrocklin:
@rabernat this may affect you

distributed/worker.py
```diff
@@ -1365,8 +1366,17 @@ def transition_executing_done(self, key, value=no_value):
...
+                    self.put_key_in_memory(key, value, transition=False)
+                except Exception as e:
+                    logger.info("Failed to put key in memory", exc_info=True)
+                    msg = error_message(e)
+                    self.exceptions[key] = msg['exception']
+                    self.tracebacks[key] = msg['traceback']
+                    self.task_state[key] = 'error'
+                    out = 'error'
+
                 if key in self.dep_state:
```
pitrou: Should this still be executed on error?

mrocklin:
OK, I've added a test that fails in a way that I'm not sure how best to fix.

When a piece of data fits in memory, and then later gets pushed to disk what should we do?  Should we keep it in memory?  Should we change the state from `memory` to `error`?

pitrou:
If we wanted to push it out from memory, then my intuition is that it should end up in the `error` state.

mrocklin:
I find myself a little bit stalled here.  Adding a transition from memory->error on the scheduler side is somewhat expensive for such a rarely occurring case.  It's hard to justify.  At the same time it's not clear that a very clean solution exists otherwise.  Still thinking about this.

pitrou:
>  Adding a transition from memory->error on the scheduler side is somewhat expensive for such a rarely occurring case.

Expensive in what terms? If the case is rarely occurring, do we mind the expense?

mrocklin:
Expensive in terms of development and maintenance.

mrocklin:
OK, I've handled this by generalizing a bit.  

The workers already had a system that would flag when they couldn't obtain a dependency.  Previously they would kick this back up to the scheduler giving up and ask the scheduler to recompute elsewhere.  The original need for this behavior was removed a while ago (handled through other means).  

Now instead of asking the scheduler to compute elsewhere they say that the task has erred.  This covers our current case (and others).  I don't think that it will affect any existing situation, but we should be aware of this change if bugs come in soon afterwards.  Stress tests pass.

mrocklin:
Merging this soon if there are no further comments.

PR #1098, opened 2017-05-25
Test function serialization issues 

mrocklin:
We've received a few issues that occur when workers are unable to deserialize functions.  This PR adds some tests to try to track this down.  I am not yet able to to isolate a fail case.


mrocklin:
This didn't succeed in finding the reported error, but it's been up long enough that I'd like to just merge it in.

ogrisel:
You should also add a test for a call to submit with undeserializable arguments (but not the callable) and also for callables that return undeserializable results.

mrocklin:

PR #1099, opened 2017-05-25
Fix warning when garbage collecting AioClient

pitrou:
AioClient would inherit `Client.__del__` and print out such warnings:
```
  /home/antoine/distributed/distributed/client.py:594: RuntimeWarning: coroutine 'AioClient.shutdown' was never awaited
    self.shutdown()
```


pitrou:
@ajdavis

ajdavis:
Thanks!

PR #1104, opened 2017-05-25
Avoid "RuntimeError: current IOLoop already exists"

ajdavis:


distributed/tests/py3_test_asyncio.py
```diff
@@ -12,22 +12,30 @@
...
+from tornado.ioloop import IOLoop
+from tornado.platform.asyncio import BaseAsyncIOLoop
+
 from distributed.asyncio import AioClient, AioFuture, as_completed, wait
 from distributed.utils_test import inc, div
 
 
 async def test_asyncio_start_shutdown():
+    # AioClient has not yet installed its AioLoop shim.
+    assert IOLoop.current(instance=False) is None
```
pitrou: Wouldn't that fail the same way `make_current=True` currently fails?

ajdavis: Agreed. I removed this, and now I'm investigating the Travis failure.

distributed/tests/py3_test_asyncio.py
```diff
@@ -241,7 +247,6 @@ def f(x, y=10):
 async def test_asyncio_cancel_tuple_key():
     async with AioClient(processes=False) as c:
         x = c.submit(inc, 1, key=('x', 0, 1))
-        await x
```
mrocklin: Why was this removed?

ajdavis: Hi Matt, I removed this as an experiment to debug the timeout error in Travis. I had a theory we shouldn't await a Future, then cancel it, which turned out not to be relevant. I'm still trying to figure out the cause of the error.

ajdavis:
Thanks for waiting - I've got a repro and perhaps a fix of the timeout error we got on the Travis build of this PR. Just need to clean it up a bit over the weekend.

ajdavis:
This is ready for review again. I replaced the aiohttp test wrapper with my own, which does not exhibit the intermittent timeout that aiohttp's does. I believe the Travis failure on the latest build is not related to my changes.

mrocklin:
> I believe the Travis failure on the latest build is not related to my changes.

Confirmed.  Replicated on master

mrocklin:
Any further comments on this @pitrou or is this good to merge?

pitrou:
@ajdavis, can we remove the dependency on `aiohttp` if we're always using the `coro_test` decorator?

Issue #801, opened 2017-01-09
Scheduler hangs randomly


csdenboer:
Hey,

I'm facing problems using parallel_backend. For some reason, the scheduler hangs randomly after few calls. I'm calling parallel_backend as follows:

`with parallel_backend('dask.distributed', scheduler_host='127.0.0.1:8786'):
    ret = Parallel()(fd(X, s, **kwds) for s in _gen_even_slices(len(X), n_jobs))`

I have 2 8-core servers where I start dask-worker as follows:
`dask-worker 10.1.0.4:8786 --nprocs 7 --nthreads 1 --no-bokeh`
They are idle at the moment the scheduler starts hanging. When I cancel the python call after it hangs, the following stacktrace is printed:

Traceback (most recent call last):
  File "/home/christian/.conda/envs/nedtrain35/lib/python3.5/site-packages/joblib/parallel.py", line 684, in retrieve
^C  File "/home/christian/.conda/envs/nedtrain35/lib/python3.5/site-packages/distributed/client.py", line 110, in result
    result = sync(self.client.loop, self._result, raiseit=False)
  File "/home/christian/.conda/envs/nedtrain35/lib/python3.5/site-packages/distributed/utils.py", line 161, in sync
    e.wait(1000000)
  File "/home/christian/.conda/envs/nedtrain35/lib/python3.5/threading.py", line 549, in wait
    signaled = self._cond.wait(timeout)
  File "/home/christian/.conda/envs/nedtrain35/lib/python3.5/threading.py", line 297, in wait
    gotit = waiter.acquire(True, timeout)
KeyboardInterrupt

The following versions are installed:
- dask                      0.13.0                   py35_0    conda-forge
- distributed               1.15.0                   py35_0    conda-forge
- tornado                   4.4.1                    py35_0

mrocklin:
Hi @csdenboer, are you able to provide a small reproducible example that we can use to generate the failure ourselves?

csdenboer:
It's pretty hard to come up with a small reproducible example. Is there somewhere a log where I can see why a worker failed?

csdenboer:
It does not seem to be the scheduler that is hanging, but the python process from which parallel_backend is called. It 'hangs' after a call to Parallel and the Python process from which it is called starts consuming 150% CPU. This is always the case after many calls did succeed.

mrocklin:
> Is there somewhere a log where I can see why a worker failed?

[...]

PR #1108, opened 2017-05-30
Make worker robust to failures in sizeof function

mrocklin:
Fixes https://github.com/dask/distributed/issues/801 cc @gouthambs

### DIFF
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -54,10 +54,9 @@ def register_pandas():
 
     @sizeof.register(pd.DataFrame)
     def sizeof_pandas_dataframe(df):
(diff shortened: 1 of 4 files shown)


mrocklin:
This should possibly be merged commit-by-commit

distributed/sizeof.py
```diff
@@ -88,3 +91,16 @@ def sizeof_spmatrix(s):
...
+
+
+def safe_sizeof(obj, default_size=1e6):
+    """ Safe variant of sizeof that captures and logs exceptions
+
+    This returns a default size of 1e6 if the sizeof function fails
+    """
+    try:
+        return sizeof(obj)
+    except:
```
pitrou: This should be `except Exception` so that we don't silence KeyboardInterrupts.

PR #1112, opened 2017-05-31
Make test_silent_startup more robust

pitrou:
Launch code in a subprocess to avoid cleanup logs from dangling coroutines or GC runs.


pitrou:
I think I managed to track down the sporadic EBADF issue in accept().

pitrou:
As for the BatchedSend log, I determined that it's failing when trying to background-send a `{'op': 'close'}` message.

PR #1113, opened 2017-05-31
More as completed methods

jcrist:
Add two new methods to `AsCompleted`:

- `is_empty`: returns True if there no waiting futures, False otherwise
- `update`: a bulk version of `add`

I'm not set on the names. Since we have `add`, I used `update` instead of `extend` to better mirror a `set`. For `is_empty`, the `Queue` class has an `empty` method, but I don't like that name as it could be confused with `clear`.

I also thought about implementing `__bool__` as `not self.is_empty()`, but didn't yet. Could go either way on that.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2649,6 +2649,11 @@ class AsCompleted(object):
             self.futures[future] += 1
         self.loop.add_callback(self.track_future, future)
 
+    def is_empty(self):
+        """Return True if there no waiting futures, False otherwise"""
+        with self.lock:
+            return not self.futures and self.queue.empty()
+
     def __iter__(self):
         return self
 
@@ -2656,9 +2661,8 @@ class AsCompleted(object):
         return self
 
     def __next__(self):
-        with self.lock:
-            if not self.futures and self.queue.empty():
-                raise StopIteration()
+        if self.is_empty():
+            raise StopIteration()
         return self.queue.get()
 
     @gen.coroutine
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2620,8 +2620,7 @@ class AsCompleted(object):
         self.with_results = with_results
 
         if futures:
-            for future in futures:
-                self.add(future)
+            self.update(futures)
 
     @gen.coroutine
     def track_future(self, future):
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -2656,9 +2668,8 @@ def __aiter__(self):
         return self
 
     def __next__(self):
-        with self.lock:
-            if not self.futures and self.queue.empty():
-                raise StopIteration()
+        if self.is_empty():
```
jcrist: May also be able to replace the similar check in `__anext__`. That one doesn't use a lock though, and I'm not sure if that was intentional or not.

distributed/client.py
```diff
@@ -2638,16 +2638,28 @@ def track_future(self, future):
                 self.queue.put_nowait(future)
             self.condition.notify()
 
+    def update(self, futures):
+        """ Add multiple futures to the collection.
+
+        The added futures will emit from the iterator once they finish"""
+        for f in futures:
+            with self.lock:
```
mrocklin: Should we swap the order of the loop and the lock here?

distributed/client.py
mrocklin: Yeah, the async code doesn't need locks as long as we're consistent between yield/await points.

distributed/tests/test_as_completed.py
```diff
@@ -65,6 +65,22 @@ def test_as_completed_add(loop):
             assert total == expected
 
 
+def test_as_completed_update(loop):
+    with cluster() as (s, [a, b]):
+        with Client(s['address'], loop=loop) as c:
+            total = 0
+            todo = list(range(10))
+            expected = sum(map(inc, todo))
+            ac = AsCompleted([])
```
mrocklin: FWIW we've been using the `as_completed` syntax everywhere.  Actually, I wouldn't mind renaming the class to `as_completed`.

distributed/client.py
```diff
@@ -2638,16 +2638,28 @@ def track_future(self, future):
                 self.queue.put_nowait(future)
             self.condition.notify()
 
+    def update(self, futures):
```
pitrou: `__init__` can now call `update`, by the way.

distributed/client.py
jcrist: Oop, that's what I meant to do. Fixed.

distributed/client.py
jcrist: Thanks, fixed.

mrocklin:
Merged.  Thanks @jcrist !

PR #1115, opened 2017-06-01
Some async-await friendly changes

mrocklin:
Commit messages contain descriptions.

cc @kszucs @ajdavis

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -15,7 +15,6 @@ import logging
 from numbers import Number
 import os
 import sys
-import threading
 from time import sleep
 import uuid
 from threading import Thread, Lock
@@ -38,7 +37,7 @@ from tornado.queues import Queue
 from .batched import BatchedSend
 from .utils_comm import WrappedKey, unpack_remotedata, pack_data
 from .cfexecutor import ClientExecutor
-from .compatibility import Queue as pyQueue, Empty, isqueue
+from .compatibility import Queue as pyQueue, Empty, isqueue, get_thread_identity
 from .core import connect, rpc, clean_exception, CommClosedError
 from .node import Node
 from .protocol import to_serialize
@@ -329,6 +328,10 @@ def done_callback(future, callback):
 def normalize_future(f):
     return [f.key, type(f)]
 
+def in_ioloop():
+    """ Are we running within the current IOLoop? """
+    return IOLoop.current()._running and IOLoop.current()._thread_ident == get_thread_identity()
+
 
 class AllExit(Exception):
     """Custom exception class to exit All(...) early.
@@ -395,7 +398,7 @@ class Client(Node):
 
         if loop is None:
             self._should_close_loop = None
-            if IOLoop.current()._running and IOLoop.current()._thread_ident == threading.get_ident():
+            if in_ioloop():
                 self.loop = IOLoop.current()
             else:
                 self.loop = IOLoop()
@@ -464,7 +467,7 @@ class Client(Node):
         pc = PeriodicCallback(lambda: None, 1000, io_loop=self.loop)
         self.loop.add_callback(pc.start)
         _set_global_client(self)
-        if self.loop._thread_ident == threading.get_ident():
+        if in_ioloop():
             self._started = self._start(**kwargs)
         else:
             sync(self.loop, self._start, **kwargs)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -15,6 +15,7 @@ import logging
 from numbers import Number
 import os
 import sys
+import threading
 from time import sleep
 import uuid
 from threading import Thread, Lock
@@ -374,13 +375,11 @@ class Client(Node):
     """
     _Future = Future
 
-    def __init__(self, address=None, start=True, loop=None, timeout=5,
+    def __init__(self, address=None, loop=None, timeout=5,
                  set_as_default=True, scheduler_file=None,
-                 security=None, **kwargs):
+                 security=None, start=None, **kwargs):
         self.futures = dict()
         self.refcount = defaultdict(lambda: 0)
-        self._should_close_loop = loop is None and start
-        self.loop = loop or (IOLoop() if start else IOLoop.current())
         self.coroutines = []
         self.id = str(uuid.uuid1())
         self.generation = 0
@@ -393,6 +392,16 @@ class Client(Node):
         assert isinstance(self.security, Security)
         self.connection_args = self.security.get_connection_args('client')
         self._connecting_to_scheduler = False
+
+        if loop is None:
+            self._should_close_loop = None
+            if IOLoop.current()._running and IOLoop.current()._thread_ident == threading.get_ident():
+                self.loop = IOLoop.current()
+            else:
+                self.loop = IOLoop()
+        else:
+            self._should_close_loop = False
+            self.loop = loop
 
         if hasattr(address, "scheduler_address"):
             # It's a LocalCluster or LocalCluster-compatible object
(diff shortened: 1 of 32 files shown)


distributed/client.py
```diff
@@ -328,6 +328,10 @@ def done_callback(future, callback):
 def normalize_future(f):
     return [f.key, type(f)]
 
+def in_ioloop():
+    """ Are we running within the current IOLoop? """
+    return IOLoop.current()._running and IOLoop.current()._thread_ident == get_thread_identity()
```
mrocklin: @kszucs @ajdavis @pitrou is there a better way to do this?  Is this generally possible?

mrocklin: And is there a way to do this with Tornado-wrapped asyncio event loops.

pitrou: We might want to propose new Tornado APIs to get this in a clean way.

distributed/client.py
```diff
@@ -394,6 +396,16 @@ def __init__(self, address=None, start=True, loop=None, timeout=5,
         self.connection_args = self.security.get_connection_args('client')
         self._connecting_to_scheduler = False
 
+        if loop is None:
+            self._should_close_loop = None
+            if in_ioloop():
+                self.loop = IOLoop.current()
```
pitrou: How about:
```
self.loop = IOLoop.current(instance=False) or IOLoop()
```
(`IOLOoop.current(instance=False)` returns None if there is no registered IOLoop for this thread)

distributed/client.py
```diff
@@ -394,6 +396,16 @@ def __init__(self, address=None, start=True, loop=None, timeout=5,
         self.connection_args = self.security.get_connection_args('client')
         self._connecting_to_scheduler = False
 
+        if loop is None:
+            self._should_close_loop = None
```
pitrou: Why not True here? It seems you can close a non-started loop:
```
>>> IOLoop().close()
>>> 
```

distributed/client.py
pitrou: At least the `._running` flag, since we're using it elsewhere too.

distributed/client.py
kszucs: This proposal seems cleaner.

mrocklin: I agree, however there are some odd cases where this changes behavior in ways that I don't totally understand.  In particular when starting many clients on the same IOLoop running in a separate thread such as often occurs when using `worker_client`.

pitrou: Do you have a simple reproducer?

mrocklin: `py.test distributed/tests/test_channels.py`

pitrou: You mean on this PR?

mrocklin: Yes, a variety of tests will fail if we make this change.  I haven't yet dug into exactly why.

pitrou: Ok, I've reproduced.

pitrou: Oh... Well, I understand why: the logic is faulty. We're trying to get the loop for the current thread in `Client.__init__`, but then we call `Client.start()` which starts that loop in another thread. That's bound to fail in various ways depending on whether the loop is already used somewhere else.

I think we need a rethink of how Client chooses, initializes and starts its event loop.
Also, I think we should deprecate the explicit `loop` argument (not only in `Client.__init__`, but in other places too), because it seems both useless and fragile.

mrocklin: I'm fine rethinking things, though would prefer to do that thinking in a separate PR (and possibly much later).  I'd like to push on if possible.  Also happy for other people to take this on.

I'm curious about your thoughts on the explicit loop argument.  Can you explain more about how it is useless and fragile in our case?

pitrou: > I'd like to push on if possible. Also happy for other people to take this on.

Fair enough. I'll try that next week.

> I'm curious about your thoughts on the explicit loop argument. Can you explain more about how it is useless and fragile in our case?

As for useless, I don't see what use case it solves: for sanity's sake, you always want a distinct IO loop per thread, and that's already Tornado's model (if you use `IOLoop.current(instance=False) or IOLoop()`, for example).

As for dangerous, people may well pass the wrong loop if they don't understand what we're doing with it. Not to mention it makes our loop-handling code more complex.

mrocklin: My reasoning for it (which I'm happy to forget if it doesn't make sense) is that we often want to specify an IOLoop living on a different thread.  For example when we make a `LocalCluster` we also make a thread running an IOLoop.  We want to tell the client to use the same particular IOLoop running on that particular thread.

pitrou: I see. Yes, that's a reasonable use case, though it does complicate things a bit.

mrocklin: Generally Dask is a bit different from most async applications because it uses the event loop, but it often uses it in a separate thread to support blocking style code for interactive scientific users.

pitrou: The only two objects those people would use directly are Client() and LocalCluster(), right?

mrocklin: Normal users yes, although there are also plenty of advanced people doing strange things.  They might want to construct their own systems and manage their own IOLoops.

I have been frequently surprised by what people do and far deep they go into internals.  I would be cautious of any change that removed flexibility.

distributed/client.py
mrocklin: We don't want to close loops that were given to us already running.  We only want to close those loops that we started ourselves.

pitrou: Oh, right.

distributed/client.py
mrocklin: Reported upstream in https://github.com/tornadoweb/tornado/issues/2070

mrocklin:
Is there anything else to do here?

mrocklin:
I plan to merge this soon if there are no further comments

mrocklin:
Looks like this fails within the Jupyter notebook, where `IOLoop.current()` is the ZMQIOLoop used by Jupyter generally.

mrocklin:
I'm going to punt and add an explicit `asynchronous=` keyword or possibly create an `AsyncClient` class.

pitrou:
What would the AsyncClient do?

mrocklin:
```python
from .client import Client

class AsyncClient(Client):
        kwargs['asynchronous'] = True

    gather = Client._gather
    get_dataset = Client._get_dataset
    publish_dataset = Client._publish_dataset
    replicate = Client._replicate
    rebalance = Client._rebalance
    restart = Client._restart
    run = Client._run
    run_coroutine = Client._run_coroutine
    run_on_scheduler = Client._run_on_scheduler
    scatter = Client._scatter
    shutdown = Client._shutdown
    start_ipython_workers = Client._start_ipython_workers
    upload_file = Client._upload_file
```

```python
from distributed.async_client import AsyncClient
from distributed.utils_test import gen_cluster

    
@gen_cluster()
def test_async_client(s, a, b): 
    c = yield AsyncClient(s.address)
    future = yield c.scatter(1)
    result = yield c.gather(future)
    assert result == 1
    yield c.shutdown()
```

mrocklin:
Remove underscores

pitrou:
What would be the plan: instruct people to use `AsyncClient` for coroutine use of a client? If so, then I'm all for it (though I would prefer another name, to avoid proximity with `AioClient`).

mrocklin:
Do you have a proposed name?  We can also rename `AioClient`.

pitrou:
Also, hopefully we can separate sync invocation support from the base client functionality (so have a BaseClient that all derived Client classes inherit from perhaps).

pitrou:
I was thinking `CoroClient` or `TornadoClient`.

mrocklin:
Hrm, those names refer to implementation details that may not be important to our average user.  The Async term is, I think, slightly more recognizable to semi-technical users.

mrocklin:
Although, I suppose one could raise the question of how much we expect average users to use this, which may be relatively little for the near term.

pitrou:
Well, if people want to use the async APIs, they have to know whether they are running on Tornado or asyncio (because a `gen.coroutine` probably won't run on asyncio). If they are using the sync APIs, the class would still be named Client.

pitrou:
And, yes, I don't expect average users to have a clear notion of cooperative concurrency such as `async def` functions or Tornado coroutines.

PR #1117, opened 2017-06-02
Distributed Queues

mrocklin:
This adds a multi-consumer multi-producer queue mediated by the scheduler.  

```python
In [1]: from distributed import Client, Queue

In [2]: client = Client()

In [3]: q = Queue('x')

In [4]: future = client.submit(lambda x: x + 1, 10)

In [5]: q.put(future)

In [6]: q.put(['hello', 'world'])

In [7]: client2 = Client(client.scheduler.address)

In [8]: q2 = Queue('x', client=client2)

In [9]: q2.get()
Out[9]: <Future: status: pending, key: <lambda>-7c3d5d0b0f151b79059d85d7dda814b0>

In [10]: _.result()
Out[10]: 11

In [11]: q2.get()
Out[11]: ['hello', 'world']

In [12]: q2.get(timeout=1)
TimeoutError: 
```

This builds on #1115 

cc @MLnick you may prefer these to channels for sending updates from workers to your parameter server.  They will be nicer in that they won't hold on to the updates in memory forever, but will release them quickly.

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -8,6 +8,7 @@ from .client import (Client, Executor, CompatibleExecutor,
                      wait, as_completed, default_client)
 from .tornado_client import TornadoClient
 from .nanny import Nanny
+from .queues import Queue
 from .scheduler import Scheduler
 from .utils import sync
 from .worker import Worker, get_worker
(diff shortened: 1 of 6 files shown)


MLnick:
@mrocklin thanks looks interesting. So quick question - does each consumer get an independent view of the queue? Or once any consumer pops the head, all other consumers will see that (i.e. it is like a mutable shared variable)?

mrocklin:
When any consumer gets a value it is burned from the queue.  This queue is semantically equivalent to the `queues.Queue` object in the standard library.

Issue #1122, opened 2017-06-02
Usage of ssl.Purpose introduced Python 2.7.9+ requirement


bluenote10:
Trying to run the latest devel branch on Ubuntu 14.04 (Python 2.7.6) fails with:

```
Traceback (most recent call last):
  File "/home/fabian/.virtualenvs/dask/bin/dask-scheduler", line 11, in <module>
    load_entry_point('distributed', 'console_scripts', 'dask-scheduler')()
  File "/home/fabian/github/distributed/distributed/cli/dask_scheduler.py", line 172, in go
    main()
  File "/home/fabian/.virtualenvs/dask/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/fabian/.virtualenvs/dask/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/fabian/.virtualenvs/dask/local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/fabian/.virtualenvs/dask/local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/fabian/github/distributed/distributed/cli/dask_scheduler.py", line 134, in main
    security=sec)
  File "/home/fabian/github/distributed/distributed/scheduler.py", line 209, in __init__
    self.connection_args = self.security.get_connection_args('scheduler')
  File "/home/fabian/github/distributed/distributed/security.py", line 126, in get_connection_args
    d['ssl_context'] = self._get_tls_context(tls, ssl.Purpose.SERVER_AUTH)
```

It looks like ea2405e541551b694b0fa095ef1da81d1cfb383e has introduced a dependence on newer Python 2.7 versions, because `ssl.Purpose` was only introduced in 2.7.9 (according to [docs](https://docs.python.org/2/library/ssl.html#ssl.create_default_context)). This breaks distributed's compatibility with older systems like Ubuntu 14.04 or Debian 7 (Python 2.7.3).

pitrou:
Yes. We use not only that, but the whole `SSLContext` API which is new in 2.7.9. I don't intend to write specific fallback code for pre-2.7.9 Pythons, unless some client wants to pay for it, as there's a non-trivial development and maintenance cost.

[...]

PR #1123, opened 2017-06-02
minimal fix for Python 2.7.6 compatibility

bluenote10:
This is the minimal code change that works around issue #1122.

I have no experience with SSL, so this need review from maybe @pitrou.

Note that this is only a partial fix, because `ssl.create_default_context` is also not available in Python 2.7.6, but without specifying a `ca_file` and `cert`, it is at least possible to get distributed running again.

### DIFF
--- a/distributed/security.py
+++ b/distributed/security.py
@@ -126,6 +126,7 @@ class Security(object):
         """
         d = {}
         tls = self.get_tls_config_for_role(role)
+        # Ensure backwards compatibility (ssl.Purpose is Python 2.7.9+ only)
         purpose = ssl.Purpose.SERVER_AUTH if hasattr(ssl, "Purpose") else None
         d['ssl_context'] = self._get_tls_context(tls, purpose)
         d['require_encryption'] = self.require_encryption
@@ -138,6 +139,7 @@ class Security(object):
         """
         d = {}
         tls = self.get_tls_config_for_role(role)
+        # Ensure backwards compatibility (ssl.Purpose is Python 2.7.9+ only)
         purpose = ssl.Purpose.CLIENT_AUTH if hasattr(ssl, "Purpose") else None
         d['ssl_context'] = self._get_tls_context(tls, purpose)
         d['require_encryption'] = self.require_encryption
--- a/distributed/security.py
+++ b/distributed/security.py
@@ -123,7 +123,8 @@ class Security(object):
         """
         d = {}
         tls = self.get_tls_config_for_role(role)
-        d['ssl_context'] = self._get_tls_context(tls, ssl.Purpose.SERVER_AUTH)
+        purpose = ssl.Purpose.SERVER_AUTH if hasattr(ssl, "Purpose") else None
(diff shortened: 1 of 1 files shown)


distributed/security.py
```diff
@@ -123,7 +123,8 @@ def get_connection_args(self, role):
         """
         d = {}
         tls = self.get_tls_config_for_role(role)
-        d['ssl_context'] = self._get_tls_context(tls, ssl.Purpose.SERVER_AUTH)
+        purpose = ssl.Purpose.SERVER_AUTH if hasattr(ssl, "Purpose") else None
+        d['ssl_context'] = self._get_tls_context(tls, purpose)
```
pitrou: I would rather have `_get_tls_context` return None (or, better, a dummy object that will raise an explicit error if one of its APIs is being used?) if the `ssl` module doesn't have the required APIs.

pitrou: Also, we can pass a string to `_get_tls_context` to work around the fact that `ssl` doesn't have the required `Purpose` enum.

bluenote10: I had a quick check: All other usages (there are only 3 if I didn't miss anything) should be okay and covered. How do you prefer to communicate the error? I went for raising a runtime error with a more explicit message.

bluenote10: Passing a string would only be important if we actually want to create a context, right? But then we are into the problem of re-implementing `create_default_context`. In other words: If there is no `Purpose`, any attempt to create a context _should_ fail, right?

distributed/security.py
```diff
@@ -105,8 +105,11 @@ def _get_config_for_role(self, category, role, fields, per_role_fields):
 
     def _get_tls_context(self, tls, purpose):
         if tls.get('ca_file') and tls.get('cert'):
-            ctx = ssl.create_default_context(purpose=purpose,
-                                             cafile=tls['ca_file'])
+            try:
+                ctx = ssl.create_default_context(purpose=purpose,
+                                                 cafile=tls['ca_file'])
+            except AssertionError:
```
pitrou: I'm curious, where does the `AssertionError` come from? Did you mean `AttributeError`?

distributed/security.py
pitrou: The RuntimeError is fine, thank you.

distributed/security.py
bluenote10: yes thanks, blind spot ;)

mrocklin:
Is there anything left to do here?

distributed/security.py
```diff
@@ -123,7 +126,8 @@ def get_connection_args(self, role):
         """
         d = {}
         tls = self.get_tls_config_for_role(role)
-        d['ssl_context'] = self._get_tls_context(tls, ssl.Purpose.SERVER_AUTH)
+        purpose = ssl.Purpose.SERVER_AUTH if hasattr(ssl, "Purpose") else None
```
pitrou: I'd like to see a comment added here to mention compatibility with old Python 2s.

pitrou:
Ok, thank you @bluenote10 !

PR #1124, opened 2017-06-02
Async cleanup

mrocklin:
Follows on https://github.com/dask/distributed/pull/1115

- This adds a TornadoClient (cc @ajdavis who requested this)
- This adds an explicit `asynchronous=` keyword to Client for use in async situations
- This slightly changes the behavior and startup of the AioClient (cc @kszucs) 

I would also like to rename `AioClient` to `AsyncIOClient`.  Any objections?

### DIFF
--- a/distributed/asyncio.py
+++ b/distributed/asyncio.py
@@ -117,27 +117,21 @@ class AioClient(Client):
 
         loop = asyncio.get_event_loop()
 
-        # "distributed" expects to call Tornado's IOLoop.current() and get the
-        # main loop, with a "_running" property. Install asyncio's loop instead.
+        # dask.distributed expects to call Tornado's IOLoop.current() and get
+        # the main loop, with a "_running" property. Install asyncio's loop
+        # instead.
         ioloop = AioLoop(loop)
-        super().__init__(*args, loop=ioloop, set_as_default=False, **kwargs)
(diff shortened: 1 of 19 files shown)


pitrou:
> I would also like to rename AioClient to AsyncIOClient. Any objections?

No objection from me, though the `aio` prefix is quite well-ingrained as a reference to the `asyncio` library (there are libraries named `aiohttp`, `aioftp`, etc.).

mrocklin:
OK, if it's a standard then I'll retract my question unless others think that it's a good idea.

distributed/asyncio.py
```diff
@@ -117,27 +117,21 @@ def __init__(self, *args, **kwargs):
...
 
     async def __aenter__(self):
-        await self.start()
+        await to_asyncio_future(self._started)
         return self
 
     async def __aexit__(self, type, value, traceback):
         await self.shutdown()
 
-    async def start(self, timeout=5, **kwargs):
```
ajdavis: Now AioClient inherits Client's "start" method. Will that work with an asyncio application?

mrocklin: I think so?  Tests pass.  We now await on the tornado `_started` future, which we appropriately wrap with `to_asyncio_future`.

PR #1126, opened 2017-06-03
Minor style improvements for Bokeh apps

bluenote10:
A few changes to the Bokeh apps:

* All apps use the template to give them a more consistent look.
* Subtle stylistic tweaks to make the dashboards look less rough.
* Styles don't use the full window width (uses CSS media query, it is wider on low resolution, and narrower on wide screens). This was a major usability issue for me, because e.g. from the four workers plots, I could only see ~1.5 of them on a fullscreen 1080p window. To avoid constant scrolling, it was necessary to resize the window or zoom-out all the time. I have chosen a width which allows to see ~3.5 of the worker plots at once, which gives a good overview.
* Small CSS fixes, like wrong font properties or wrong container heights. It's not perfect yet due to some Bokeh quirks, but most issues should be resolved.
* Removed the code + templates in the submodules `status`, `workers`, and `tasks`. As far as I can see that should be dead/legacy code, right?

I'll try to attach a screenshot to show the visual result.


bluenote10:
screenshot from 2017-06-03 10 55 03

mrocklin:
Thank you for this.  It's good to have more people work on the diagnostic pages, especially people with some web styling experience.  The failing test `test_BokehWebInterface` is testing the legacy diagnostic pages, which you have removed.  I agree that it's probably time that we can remove these safely.  You should also remove the `--bokeh-external-port` flag in cli/dask_scheduler.py` as well as the `bokeh/application.py` file.

mrocklin:
Do we want to limit the scale_width behavior from before?  I'm seeing this by default on my screen:

image

mrocklin:
For comparison: 

image

mrocklin:
Ah I see, this is your second bullet point from above.  FWIW my typical behavior here is to change the browser width to suit my needs.  Sometimes I do want to focus on one particular plot so I expand the browser window and scroll appropriately.

Also cc @canavandl for styling comments

bluenote10:
Is your screen width larger than 1920? On a 1080p display the page should have 80% width: 

screenshot from 2017-06-03 13 28 23

For higher resolutions it is set to 60%, which is probably too narrow. Maybe I go for 80% as well.

mrocklin:
My screen width is 2560.  People also use regularly Dask on large hi-res monitors that have even larger width.

bluenote10:
Sure, I was using the UI yesterday on a high-res monitor, where the large plot size was even a bigger issue. I have added the media query part to allow fine-tuning. I'll be back on my high-res monitor on Tuesday to check again.

mrocklin:
Thanks again for working on this.  The dashboard is well loved and very important, but most people who use Dask aren't very familiar with web styling (myself included).  It's great to have more people thinking about and working on it.

bluenote10:
I have worked in the review comments. I was hesitating to just delete this test, but if it was only testing the old bokeh app, I hope that is fine. What I don't understand though is why the test didn't fail with an `ImportError`.

mrocklin:
General question: now that we include the header in all pages do you have any thoughts on how to reduce vertical whitespace or improve the styling generally?  Personally I like having the Dask Logo but I find that I don't use many of the links.

mrocklin:
I plan to merge this soon.  

I think that in the future we should consider improving the template.

bluenote10:
You're right, the links are mainly relevant for first time users. From a usability perspective links to the other dashboard would be more helpful to speed up navigating between them. Personally I don't care about the space lost on the top of the page (and I like seeing the Dask logo there), because I can always scroll down. For maximizing screen estate, one might consider putting the links to other dashboards in a side-board though. But yes, this can be optimized later.

mrocklin:
Merged.  Thanks @bluenote10 !

mrocklin:
Yes, I agree that replacing the docs links with other dashboard pages would be good.  I also like seeing the Dask logo.  Perhaps a [standard navigation bar](https://www.google.com/search?site=&tbm=isch&source=hp&biw=1600&bih=767&q=navigation+bar&btnG=Search+by+image&oq=exascale&gs_l=img.3..0l2j0i24l8.1603.2938.0.3163.8.8.0.0.0.0.187.784.3j5.8.0....0...1ac.1.31.img..2.6.468.txrajucGCEY&gws_rd=ssl) in the header?

PR #1130, opened 2017-06-04
Scatter update

mrocklin:
This includes a few fixes:

1.  We add a `direct=` keyword to scatter which pushes data directly from the client to the workers without going through the scheduler.  Better names welcome.
2.  We now scatter singleton numpy arrays and pandas dataframes (and all other Iterables that are not the standard builtin list/tuple/set/etc.) as single elements rather than lists
3.  We ensure that tokenization happens on the client side rather than on the scheduler.  This may have been causing undue serialization/CPU costs on the scheduler.

```python
In [2]: client = Client()
In [3]: futures = client.scatter([1, 2, 3], direct=True)  # this avoids the scheduler

In [4]: import numpy as np
In [5]: x = np.ones(5)
In [6]: future = client.scatter(x)
In [7]: future
Out[7]: <Future: status: finished, type: ndarray, key: 1ec89731b5394a422f6f47679b292d54>
```

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1147,44 +1147,38 @@ class Client(Node):
             d = yield self._scatter(keymap(tokey, data), workers, broadcast)
             raise gen.Return({k: d[tokey(k)] for k in data})
 
+        if isinstance(data, type(range(0))):
+            data = list(data)
+        input_type = type(data)
+        names = False
         unpack = False
-        if isinstance(data, dict):
-            data2 = valmap(to_serialize, data)
-            types = valmap(type, data)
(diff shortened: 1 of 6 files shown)


martindurant:
AM I right in thinking that the previous behaviour - passing everything through the scheduler - remains the default? I ask because there are implications for setting up a remote cluster, needing to expose the worker ports to the outside.

mrocklin:
Yes, that remains the default.  Yes, this is because it is not guaranteed that client machines will have access to workers directly.

mrocklin:
I'm inclined to merge this.  I plan to clean things up a bit later.

Issue #1134, opened 2017-06-05
Better HTML repr for Client, Futures


mrocklin:
Currently we represent the client and futures as follows with text:

```python
In [12]: client
Out[12]: <Client: scheduler='tcp://127.0.0.1:43109' processes=4 cores=4>

In [13]: future
Out[13]: <Future: status: finished, type: int, key: add-6bc144bc0ae4e5f1e17a1b8e5cafe9c2>
```

Many of our users use Jupyter notebooks.  Should we provide a richer HTML-based repr in this context?  If so what content should we add and how should we style it?

PR #1136, opened 2017-06-05
Add HTML repr methods for Future and Client

mrocklin:
image

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -489,7 +489,7 @@ class Client(Node):
                 address = 'http://%s:%d' % (host, port)
             text += "  <li><b>Dashboard: </b><a href='%(web)s' target='_blank'>%(web)s</a>\n" % {'web': address}
 
-        text += "</ul>\n\n"
+        text += "</ul>\n"
 
         if info:
             workers = len(info['workers'])
@@ -500,13 +500,17 @@ class Client(Node):
                      "<ul>\n"
                      "  <li><b>Workers: </b>%d</li>\n"
                      "  <li><b>Cores: </b>%d</li>\n"
-                     "  <li><b>Memory: </b>%s</li>\n") % (
-                             workers, cores, memory)
+                     "  <li><b>Memory: </b>%s</li>\n"
+                     "</ul>\n") % (workers, cores, memory)
 
-        # return ('<table border="0">\n'
-        #         "<tr><td>%s</td><td>%s</td></tr></table>") % (text, text2)
+            return ('<table style="border: 2px solid white;">\n'
+                    '<tr>\n'
+                    '<td style="vertical-align: top; border: 0px solid white">\n%s</td>\n'
+                    '<td style="vertical-align: top; border: 0px solid white">\n%s</td>\n'
+                    '</tr>\n</table>') % (text, text2)
 
-        return text + text2
+        else:
+            return text
 
     def start(self, asynchronous=None, **kwargs):
         """ Start scheduler running in separate thread """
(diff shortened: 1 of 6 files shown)


mrocklin:
Recommendations for styling or desired content are very welcome.

Fixes #1134

pitrou:
Are there general guidelines for when an object should provide a HTML pretty-printer on Jupyter? On the one hand, this makes the appearance obviously nicer, on the other hand it brings it farther from the experience of a CLI-based prompt.

(also, people may not like the additional vertical overhead in the Client case above)

minrk:
libraries like pandas and sympy often put the behavior of such things behind an option, so that users can opt-in/out of HTML rendering (depending on what authors prefer as default).

distributed/client.py
```diff
@@ -454,6 +470,44 @@ def __str__(self):
...
+            text += "  <li><b>Dashboard: </b><a href='%(web)s' target='_blank'>%(web)s</a>\n" % {'web': address}
+
+        text += "</ul>\n\n"
+
+        if info:
+            workers = len(info['workers'])
+            cores = sum(w['ncores'] for w in info['workers'].values())
+            memory = sum(w['memory_limit'] for w in info['workers'].values())
+            memory = format_bytes(memory)
+            text2 = ("<h3>Cluster</h3>\n"
```
TomAugspurger: Could this be a defined on `Cluster`, so that `client.cluster` has the same HTML repr for its info?

mrocklin:
> (also, people may not like the additional vertical overhead in the Client case above)

I agree with this generally and would like to find a better arrangement.  I tried arranging things into an HTML table (see commented code), however the border around the table looked bad and I wasn't able to remove it.  Recommendations welcome.

mrocklin:
> The only other thing that comes to my mind is the status / health of individual workers.

This can get tricky in larger deployments.  My original thought here was to provide a table that grouped similar workers together ("you have 10 workers with 4 cores each and 20 workers with 1 core each") but decided that in the common case workers are pretty uniform, so this might not be useful.  I now think that the right thing to do here is to provide some synopsis values (like total memory) and leave the rest for the dashboard for now.

distributed/client.py
```diff
@@ -454,6 +470,44 @@ def __str__(self):
...
+                     "<ul>\n"
+                     "  <li><b>Workers: </b>%d</li>\n"
+                     "  <li><b>Cores: </b>%d</li>\n"
+                     "  <li><b>Memory: </b>%s</li>\n") % (
+                             workers, cores, memory)
+
+        # return ('<table border="0">\n'
```
pitrou: Have you tried `<table style="border: 0px;">`?

distributed/client.py
```diff
@@ -454,6 +470,44 @@ def __str__(self):
...
+
+        text += "</ul>\n\n"
+
+        if info:
+            workers = len(info['workers'])
+            cores = sum(w['ncores'] for w in info['workers'].values())
+            memory = sum(w['memory_limit'] for w in info['workers'].values())
+            memory = format_bytes(memory)
+            text2 = ("<h3>Cluster</h3>\n"
+                     "<ul>\n"
```
pitrou: The `<ul>` tag should be closed at the end.

mrocklin: Fixed

distributed/client.py
mrocklin: Currently we're getting this information by asking the scheduler over TCP, not by looking at the cluster object.  This should generalize well to cases where the scheduler is not present in the same process.

However I agree that it would be nice to also have an html repr for the `LocalCluster`  object and drop that in here instead if it is available.

mrocklin:
image

mrocklin:

PR #1138, opened 2017-06-06
Close comm on EOFError

pitrou:
If we get an EOFError when unserializing a distributed message, it's likely something went wrong with the communication.  We close the communication and also log the frames if not too large.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -8,6 +8,7 @@ import ssl
 import struct
 import sys
 
+import tornado
 from tornado import gen, netutil
 from tornado.iostream import IOStream, StreamClosedError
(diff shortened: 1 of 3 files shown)


distributed/comm/tcp.py
```diff
@@ -173,8 +180,10 @@ def write(self, msg):
         if stream is None:
             raise CommClosedError
 
-        # IOStream.write() only takes bytes objects, not memoryviews
-        frames = [ensure_bytes(f) for f in to_frames(msg)]
+        if self._iostream_allows_memoryview:
```
pitrou: This is unrelated but came by while reading this code.

mrocklin: This is useful.  Thanks.

PR #1140, opened 2017-06-06
Make client threadsafe for common operations

mrocklin:
This tests thread safety for common client operations.  To be honest, most of the tests ran fine previous to these changes, they just raised more warnings.  We add a lock around any operation that mutates or depends on a consistent view of the state of futures from the main thread.  All operations that merely sync down to the IOLoop thread are probably safe already.

I'm inclined not to merge this until after the next release.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -255,10 +255,9 @@ class Future(WrappedKey):
         return self._state.type
 
     def release(self):
-        with self.client._lock:
-            if not self._cleared and self.client.generation == self._generation:
-                self._cleared = True
-                self.client._dec_ref(tokey(self.key))
+        if not self._cleared and self.client.generation == self._generation:
+            self._cleared = True
+            self.client._dec_ref(tokey(self.key))
 
     def __getstate__(self):
         return self.key
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -255,9 +255,10 @@ class Future(WrappedKey):
         return self._state.type
 
     def release(self):
-        if not self._cleared and self.client.generation == self._generation:
-            self._cleared = True
-            self.client._dec_ref(tokey(self.key))
+        with self.client._lock:
+            if not self._cleared and self.client.generation == self._generation:
+                self._cleared = True
+                self.client._dec_ref(tokey(self.key))
 
     def __getstate__(self):
         return self.key
@@ -579,8 +580,7 @@ class Client(Node):
         if self.status is 'running':
             self.loop.add_callback(self.scheduler_comm.send, msg)
         elif self.status is 'connecting':
-            with self._lock:
-                self._pending_msg_buffer.append(msg)
+            self.loop.add_callback(self._pending_msg_buffer.append, msg)
         else:
             raise Exception("Client not running.  Status: %s" % self.status)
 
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -251,9 +251,10 @@ def type(self):
         return self._state.type
 
     def release(self):
-        if not self._cleared and self.client.generation == self._generation:
-            self._cleared = True
-            self.client._dec_ref(tokey(self.key))
+        with self.client._lock:
+            if not self._cleared and self.client.generation == self._generation:
+                self._cleared = True
+                self.client._dec_ref(tokey(self.key))
```
pitrou: What is the likelihood that a decref operation triggers other lock-taking operations (e.g. by way of garbage collection)?

pitrou:
I would be a bit cautious about this. Promising thread-safety may increase maintenance costs for us in the future, or prove to be tricky to achieve. For the time being, I would rather document that the Client class is not meant to be thread-safe and people should use their own thread-safety mechanisms if they need thread-safe operation.

mrocklin:
The motivation for this change is that if we start doing more dynamic client-on-worker workloads then every thread on every worker may want to have its own client to communicate to the scheduler.  This both significantly increases the number of open connections and becomes tricky when users rely on global state like the `client.get` method that gets used in `x.compute()`.

distributed/client.py
mrocklin: After briefly looking at the code I think that we are safe here.  The client has much less state than the scheduler and is generally much easier to reason about.

mrocklin:
I'm running into a number of cases where the Python 2 test cases fail.  In particular it appears that during the `start_ipython_workers` tests (which fail occasionally normally) we now fail quite regularly.  One of the threads is stuck waiting on a lock when garbage collecting futures.

pitrou:
Oh, yes, garbage collection is typically asynchronous and therefore one should be careful taking any locks there.

pitrou:
This is true on Python 3 as well, of course, but perhaps the timings are different and therefore the problem doesn't show up on our CI.

mrocklin:
> Oh, yes, garbage collection is typically asynchronous and therefore one should be careful taking any locks there.

It's odd that things hang though.  Somehow a lock was acquired and never released.

pitrou:
> It's odd that things hang though. Somehow a lock was acquired and never released.

I don't think that's what's happening.  Probably a GC run happens while the lock is taken, the GC collects a Future which then tries to re-take the Lock in its `__del__`.

Using `loop.add_callback` in `__del__` or in `release` might be the solution.

mrocklin:
> Probably a GC run happens while the lock is taken, the GC collects a Future which then tries to re-take the Lock in its __del__.

Why would this cause tests to hang?  Or maybe it isn't hanging here, this is just the transient state of things when the timeout triggers?

mrocklin:
> Using loop.add_callback in __del__ or in release might be the solution.

I'll take a look

mrocklin:
> Using loop.add_callback in del or in release might be the solution.

This now seems odd because in `__del__` we now create a new reference to `self`.  How does Python respond in cases like this?  It seems as though the object will never be deleted or it will be deleted too soon.  The alternative here seems to be to set a flag (which we do) and to use another lock.

pitrou:
If you revive the object in `__del__`, it will then remain alive (until all references are lost again). I would recommend against this. Can you arrange so that no reference to `self` is needed? `release` is very simple so you can inline and tweak it, for example:
```
    def __del__(self):
        if not self._cleared:
            def clear(client=self.client, gen=self._generation, key=tokey(self.key)):
                if client.generation == gen:
                    client._dec_ref(key)
            self.client.loop.add_callback(clear)
```

mrocklin:
I resolve the issue a different way.  There was a corner case in which we tried to acquire the lock within a function where the lock was already acquired.  This has been changed so that that code remains safe by always running within the event loop.

PR #1141, opened 2017-06-06
Add navbar to bokeh pages

mrocklin:
This adds a basic navigation bar to the bokeh template

This still needs work.  In particular we need to populate the navbar with the
correct links for the different pages.

image


mrocklin:
Does anyone know if it is possible to partially render a jinja template?  I want to populate the list of links to include, but also need to hand Bokeh a proper unrendered template object.

mrocklin:
Ok, I've figured out how to use template variables in bokeh.  I think everything works ok here.  Style comments and css love would be appreciated.

PR #1142, opened 2017-06-07
FIX: load traceback fails task status not set to error

3kwa:
While taking dask.distributed for a serious spin I noticed that some task status were not set to error despite the job failing.

> distributed.protocol.pickle - INFO - Failed to deserialize

The problem is caused by the tblib raising an AttributeError when loads(traceback) is called - I am calling R from Python in the task which could explain why there is a traceback issue with tasks raising RRuntimeError.

> AttributeError: ("'Traceback' object has no attribute 'tb_next'", ...)

Setting traceback to None when that happens seems to be a better option than having the tasks hang in there as pending despite having erred.

Not sure how to go about writing an automated test for that corner case - happy to give it a shot just need some guidance.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -706,7 +706,10 @@ class Client(Node):
             except TypeError:
                 exception = Exception("Undeserializable exception", exception)
             if traceback:
-                traceback = loads(traceback)
+                try:
+                    traceback = loads(traceback)
+                except AttributeError:
+                    traceback = None
             else:
                 traceback = None
             state.set_error(exception, traceback)


mrocklin:
Hi @3kwa .  Thanks for the fix.  

A couple of questions: 

1.  Should we log the error?  Is this useful or annoying to users?  Perhaps log on debug level (which is usually ignored) with `logger.debug("Failed to deserialize traceback", exc_info=True)`?
2.  Should we be catching other kinds of exceptions?

3kwa:
Pleasure @mrocklin - thanks for dask!

1. I am not sure it is necessary since it raises an exception which the pickle protocol logger already output as info.
2. I am not very familiar with all the exceptions tblib could raise so can only say maybe.

mrocklin:
Thanks @3kwa !

PR #1152, opened 2017-06-08
Use normal Client API in asynchronous mode

mrocklin:
Previously we had two parallel APIs for synchronous/blocking and
asynchronous/non-blocking operation with underscored functions being
asynchronous:

    client.gather(...)  # synchronous
    yield/await client._gather(...)  # asynchronous

Now that we ask the user to explicitly provide an asychronous=True flag
we can make this choice for them and use the same function name in both cases.

    client = yield Client(asynchronous=True)
    yield/await client.gather(...)

This was a common mistake among new users who found their programs hanging.
This also allows us to have a single set of public facing functions.

cc @kszucs @ajdavis


mrocklin:
I am now inclined to remove the `TornadoClient` object

pitrou:
Should `AioClient` reuse this infrastructure instead of redeclaring all methods?

mrocklin:
I think it will need to redeclare them anyway do handle the tornado-future -> asyncio-future conversion.

mrocklin:
I've added docs and removed the `TornadoClient`

ajdavis:
I like the idea. I hope that users will discover their mistake easily if they use a Client in a coroutine but forget to set "asynchronous". Right now, the "Client()" constructor will deadlock here in "sync()":

```
    while not e.is_set():
        e.wait(1000000)
```

I can't think of a way to make that less error-prone.

Is the "start" method properly updated for the new design? It seems like "start" has a redundant "asynchronous" parameter that should be removed. Additionally, the Client parameter "start" has been unused since a few commits ago.

mrocklin:
I also can't think of a way to make `Client()` less error prone while within an event loop. 

I've added an informative error if the user uses `start=` pointing them to the `asynchronous=` keyword.

PR #1155, opened 2017-06-09
lz4: first try passing memoryview before fallback

pitrou:
Current versions of the lz4 binding don't support memoryview inputs, but later versions should (see https://github.com/python-lz4/python-lz4/pull/38/).

### DIFF
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -158,7 +158,7 @@ def maybe_compress(payload, min_size=1e4, sample_size=1e4, nsamples=5):
     else:
         nbytes = len(payload)
 
-    if blosc and type(payload) is memoryview and payload.itemsize > 1:
+    if blosc and type(payload) is memoryview:
         # Blosc does itemsize-aware shuffling, resulting in better compression
         compressed = blosc.compress(payload, typesize=payload.itemsize,
                                     cname='lz4', clevel=5)
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -60,13 +60,28 @@ with ignoring(ImportError):
         lz4_compress = lz4.LZ4_compress
         lz4_decompress = lz4.LZ4_uncompress
 
-    def _fixed_lz4_decompress(data):
-        # helper to bypass missing memoryview support in lz4
-        if isinstance(data, memoryview):
-            data = data.tobytes()
-        return lz4_decompress(data)
+    # helper to bypass missing memoryview support in current lz4
+    # (fixed in later versions)
 
-    compressions['lz4'] = {'compress': lz4_compress,
+    def _fixed_lz4_compress(data):
+        try:
+            return lz4_compress(data)
+        except TypeError:
+            if isinstance(data, memoryview):
+                return lz4_compress(data.tobytes())
+            else:
+                raise
+
+    def _fixed_lz4_decompress(data):
+        try:
+            return lz4_decompress(data)
+        except TypeError:
+            if isinstance(data, memoryview):
+                return lz4_decompress(data.tobytes())
+            else:
+                raise
+
+    compressions['lz4'] = {'compress': _fixed_lz4_compress,
                            'decompress': _fixed_lz4_decompress}
     default_compression = 'lz4'
 
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -158,7 +158,8 @@ def maybe_compress(payload, min_size=1e4, sample_size=1e4, nsamples=5):
     else:
         nbytes = len(payload)
 
-    if blosc and type(payload) is memoryview:
+    if blosc and type(payload) is memoryview and payload.itemsize > 1:
+        # Blosc does itemsize-aware shuffling, resulting in better compression
         compressed = blosc.compress(payload, typesize=payload.itemsize,
                                     cname='lz4', clevel=5)
         compression = 'blosc'
--- a/distributed/protocol/tests/test_protocol.py
+++ b/distributed/protocol/tests/test_protocol.py
@@ -7,6 +7,7 @@ import pytest
 
 from distributed.protocol import (loads, dumps, msgpack, maybe_compress,
         to_serialize)
+from distributed.protocol.compression import compressions
 from distributed.protocol.serialize import (Serialize, Serialized,
                                             serialize, deserialize)
 from distributed.utils_test import slow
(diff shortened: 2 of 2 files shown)


pitrou:
@mrocklin, would you know why there's a special case for memoryviews and blosc?

mrocklin:
We typically get memoryviews from numpy arrays which have dtype information.  Blosc handles this dtype information well by doing a bit-shuffle by the itemsize first.

distributed/protocol/compression.py
```diff
@@ -60,13 +60,28 @@ def _fixed_snappy_decompress(data):
...
 
-    compressions['lz4'] = {'compress': lz4_compress,
+    def _fixed_lz4_decompress(data):
+        try:
+            return lz4_decompress(data)
+        except TypeError:
+            if isinstance(data, memoryview):
+                return lz4_decompress(data.tobytes())
+            else:
+                raise
```
mrocklin: Is there a cost to this exception?  If we know an explicit version then it might be better to perform this check once:

```python
if lz4.__version__ > ...:
    _fixed_lz4_compress = lz4_compress
else:
    def _fixed_lz4_compress(data):
        ...
```

pitrou: The cost is negligible:
```python
>>> m = memoryview(b'.')
>>> def f(x):
...:    try:
...:        return lz4.block.compress(x)
...:    except TypeError:
...:        pass
...:    
>>> f(m)
>>> %timeit f(m)
1.4 µs ± 9.66 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

pitrou:
If the user chose a compression scheme over than "auto", is it really desirable to switch to blosc?

mrocklin:
Possibly, yes.  Sometimes blosc is very good.  Here is an artificially good case for blosc.

```python
In [1]: import numpy as np

In [2]: import lz4

In [3]: import blosc

In [4]: x = np.arange(10000000)

In [5]: b = x.data.tobytes()

In [6]: %time len(lz4.block.compress(b))
CPU times: user 164 ms, sys: 4 ms, total: 168 ms
Wall time: 167 ms
Out[6]: 40014270

In [7]: %time len(blosc.compress(x.data, typesize=x.data.itemsize, clevel=5, cname='lz4'))
CPU times: user 52 ms, sys: 0 ns, total: 52 ms
Wall time: 15.7 ms
Out[7]: 1364898
```

martindurant:
Hah, that must be the delta filter. Should then apply run-length-encoding, this data should require almost no bytes at all.
Which reminds me, is there any interest for [numcodecs](http://numcodecs.readthedocs.io/en/latest/delta.html) for a wider range of (numerical) compressions, including blosc, with a consistent API? There the delta filter can be explicitly passed. I don't believe there is functionality to automatically pick the "best" option for given data.

distributed/protocol/compression.py
```diff
@@ -143,7 +158,8 @@ def maybe_compress(payload, min_size=1e4, sample_size=1e4, nsamples=5):
     else:
         nbytes = len(payload)
 
-    if blosc and type(payload) is memoryview:
+    if blosc and type(payload) is memoryview and payload.itemsize > 1:
+        # Blosc does itemsize-aware shuffling, resulting in better compression
```
mrocklin: It also supports memoryviews in older lz4 versions (on which versions is this active?  I get errors even fairly recently) and uses multithreading.  I'm inclined to use blosc for memoryviews if available.  Is there any reason not to use it?

pitrou: But then why not use it for everything? I don't really understand the strategy here.

pitrou:
More broadly, I think I'm misunderstanding things here:
* if the user wants a particular compression algorithm (i.e. not "auto"), why force "blosc" silently? presumably they have their reasons to choose something specific
* "auto" chooses a fixed scheme by default ("lz4" preferably), but sometimes overrides it with a dynamic data-driven choice ("choice")

I would have expected the user's choise (other than "auto") to always be respected, and I would expect "auto" to be either fully static (e.g. always "lz4" preferred) or fully dynamic (one could imagine sampling a smallish piece of data and try all algorithms on it).

jonathanunderwood:
FYI I've just cut a new release of python-lz4 with the changes from @pitrou to allow dealing with memoryviews etc incorporated - see version 0.10.0.

jonathanunderwood:
Ps. also note that there's some tuning you can do with the lz4 bindings to trade off speed and compression ratio. If speed is your primary concrn, you should set `mode='fast'` and tune `acceleration`.  See:

http://python-lz4.readthedocs.io/en/latest/lz4.block.html

mrocklin:
@pitrou your concerns seem valid.  FWIW I've never known a user to specify a compression algorithm.  I think that the default should be lz4 most of the time but blosc when we think it's likely to make a difference.  Anecdotally I've found blosc to be better when we have numeric data such as comes from numpy arrays or pandas dataframes and lz4 to be a better choice when data comes from other sources.  Anecdotally I've also found that receiving a memoryview is a strong signal that our data is coming from numeric data, hence the current approach.

I wouldn't mind changing this approach, but I think that we shouldn't bother until it is requested by a user or we have evidence that it improves performance.

mrocklin:
As always, thank you for maintaining `python-lz4` @jonathanunderwood .  It's really appreciated.

jonathanunderwood:
No problem. Actually, I hadn't seen blosc before this thread - I notice it natively leverages LZ4, so I am slightly surprised that there's any situation where python-lz4 adds much benefit!

mrocklin:
Blosc is more heavyweight.  Where it's pre-compressors make sense it's quite valuable, when they don't make sense it can have worse performance.  Also, not everyone has blosc installed.  LZ4 seems to be more common.

pitrou:
Yes, the reason we support several compressors is that some may be more cumbersome to install than others. By the way, it might be worthwhile to investigate [zstandard](https://facebook.github.io/zstd/) as well.

mrocklin:
As far as I can tell this is low-priority though.  It might become more important as people start using dask on high performance networks though (which seems to be happening slowly).

pitrou:
For high-performance networks lz4 or no compression is probably preferable. zstandard would allow to achieve much better compression ratios at the price of a slightly higher compression cost.

pitrou:
Another point worth considering is that as long as CPU use is maxed out, it's better to offload some resource consumption from the CPU to the network (i.e. if the CPU is already a bottleneck, perhaps do less compression).

pitrou:
It might be interesting to know what the trend of network performance vs. CPU density improvements is.

mrocklin:
>  For high-performance networks lz4 or no compression is probably preferable.

Yes, I entirely agree.  And for low-performance networks (like EC2) it's probably better to use this over LZ4.  If communication heavy workloads start showing up more then making these sorts of decisions may become useful.

> it's better to offload some resource consumption from the CPU to the network (i.e. if the CPU is already a bottleneck, perhaps do less compression).

Agreed.  Anecdotally my experience is that when communication is high we are almost always communication bound.  In other words, when we are computation bound there is usually very little communication, so this tends not to matter as much.  I can think of some upcoming workloads (like https://github.com/dask/dask/issues/2225) that would engage both at the same time though.

pitrou:
> I can think of some upcoming workloads (like dask/dask#2225) that would engage both at the same time though.

Thanks for reminding me :-)

distributed/protocol/compression.py
mrocklin: Copying text from a longer comment earlier

> Anecdotally I've found blosc to be better when we have numeric data such as comes from numpy arrays or pandas dataframes and lz4 to be a better choice when data comes from other sources. Anecdotally I've also found that receiving a memoryview is a strong signal that our data is coming from numeric data, hence the current approach.

I still think that we should retain the old behavior.  This would come up, for example, when passing around image data (`uint8`) where I think that blosc would still be a better choice.

PR #1157, opened 2017-06-11
Remove Python 3.4 run from Travis-CI

pitrou:
This may help speed up builds a bit.
Also remove the ASYNCIO environment variable which doesn't seem examined anywhere.
Also taking the liberty of adding `setup.py` classifiers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,8 @@ services:
 env:
   matrix:
     - PYTHON=2.7 PACKAGES="blosc futures faulthandler"
-    - PYTHON=3.4 COVERAGE=true DASK_EXPERIMENTAL_ZMQ=1
-    - PYTHON=3.5 CRICK=true ASYNCIO=true
-    - PYTHON=3.6 PACKAGES=blosc ASYNCIO=true
+    - PYTHON=3.5 COVERAGE=true PACKAGES=blosc CRICK=true DASK_EXPERIMENTAL_ZMQ=1
+    - PYTHON=3.6
     - HDFS=true PYTHON=2.7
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -41,6 +41,19 @@
                 'distributed.http'],
       long_description=(open('README.rst').read() if os.path.exists('README.rst')
                         else ''),
+      classifiers=[
+        "Development Status :: 5 - Production/Stable",
```
pitrou: Perhaps I'm being optimistic?

mrocklin: Seems fine to me.  This seems project seems stable relative to most other projects.

.travis.yml
```diff
@@ -7,9 +7,8 @@ services:
 env:
   matrix:
     - PYTHON=2.7 PACKAGES="blosc futures faulthandler"
-    - PYTHON=3.4 COVERAGE=true DASK_EXPERIMENTAL_ZMQ=1
-    - PYTHON=3.5 CRICK=true ASYNCIO=true
-    - PYTHON=3.6 PACKAGES=blosc ASYNCIO=true
+    - PYTHON=3.5 COVERAGE=true PACKAGES=blosc CRICK=true DASK_EXPERIMENTAL_ZMQ=1
```
mrocklin: Thoughts on dropping ZMQ?

pitrou: No hard feelings from me. It we drop testing it, though, we can probably drop it from the code base as well. What do you think?

mrocklin: I think it's value is in verifying that comms is extensible.  However now that `inproc` exists that may suffice.  I'm inclined to remove it just to reduce the codebase.

mrocklin:
Merged

PR #1161, opened 2017-06-12
Minor typo in documentation

ThomasA:
It said 'torando.tcpserver.TCPServer' somewhere instead of 'tornado.tcpserver.TCPServer'.


pitrou:
Actually, this section should be rewritten, since those objects are not `TCPServer` objects anymore.

ThomasA:
OK, I'm afraid I can't help with that

mrocklin:
> OK, I'm afraid I can't help with that

Yeah, nor would I expect you to :)  This looks great to me.  Merging on passed tests (I had to reset things, due to an unrelated issue).

mrocklin:
Thanks @ThomasA !   This is in.

PR #1185, opened 2017-06-18
Update queues

### LINKED ISSUE #1180: Calling get() on a variable returns a future which does not have the latest state
Simple. Submit a future and add it to a variable. Then, from another client "peek" at the variable, i.e. don't call result() just look at the state and see if it is finished by calling variable.get()

Behavior: sometimes the state is fresh while other times it takes long to actually get the latest state of the future. 

By calling result on the future everything is fine. However I can't block  and need to return. Only when the future is finished can the caller block to get the result.

`   def inc(self, x):
        return x + 1

    def dummy(self, x):
        pass

    def test_variables_2(self):
        with Client(cfg.current_config.DISTRIBUTED_SCHEDULER) as client:
            L = client.map(self.inc, range(1000))
            f = client.submit(self.dummy, L)

            v = Variable(name="var", client=client)
            v.set(f)

            client.compute(f)
    
        while True:
            with Client(cfg.current_config.DISTRIBUTED_SCHEDULER) as client:
                v = Variable(name="var", client=client)
                f = v.get()

                if f.status == 'finished':
                    break
                else:
                    print('not finished')
                    time.sleep(1)``
`

mrocklin:
Each commit here has a decent explanation.  

1.  Support getting batches of elements from queues
2.  Get status of future immediately (Fixes #1180 )

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -94,7 +94,7 @@ class Future(WrappedKey):
     _cb_executor = None
     _cb_executor_pid = None
 
-    def __init__(self, key, client, inform=False):
+    def __init__(self, key, client, inform=False, state=None):
         self.key = key
         self._cleared = False
         tkey = tokey(key)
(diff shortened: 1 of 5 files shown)


mrocklin:
Fixes #1180

Issue #1164, opened 2017-06-13
Asyncio interface for Variable 


kszucs:
I guess it's enough to wrap _get and _set with to_asyncio, just not sure how to handle the default client. Right now, AioClient can't be set as default.

Ideas?

mrocklin:
Set the client explicitly?  Have AioClient claim the default client global slot?

jakirkham:
This is now unnecessary given PR ( https://github.com/dask/distributed/pull/1458 ) is in, correct?

PR #1186, opened 2017-06-19
AioClient as default and AioVariable

kszucs:
Trying to solve #1164


kszucs:
@mrocklin How should I handle https://github.com/dask/distributed/blob/master/distributed/client.py#L2973 ?

mrocklin:
Sorry for the delay.  No particular thoughts on setting the default client.  It seems like a good idea to me?  What are your thoughts?

Looks like the test here is failing.

mrocklin:
Superceded by #1458 .  Closing

PR #1194, opened 2017-06-20
Tell helper thread to end in AsyncProcess.join()

pitrou:

PR #1195, opened 2017-06-20
Allow explicit close of AsyncProcess

pitrou:
This will reap the helper thread, regardless of possible reference cycles delaying garbage collection.


pitrou:
This is an alternative to #1194.

pitrou:
The 2.7 failure (in `test_threadsafe_compute`) is very unlikely to be related. I'm going to restart the job.

PR #1199, opened 2017-06-21
Fix #1158: stress socket backlog

### LINKED ISSUE #1158: Add a configuration entry for TCP socket backlog
This may help write tests which would stress saturating the connection accept queue.

pitrou:


pitrou:
Unfortunately the test fails quite consistently.

mrocklin:
Any thoughts here on how/if we should make Dask robust to tasks like this?

pitrou:
One easy thing would be to make the connect-timeout configurable, which would help our users with large clusters. Generally there is a tension between retrying a certain number of times and returning early when an error is detected. Making it configurable will let users choose whichever is best of them.

pitrou:
As for the situation in the test, it's quite unusual so I'm not sure whether we should do something against it (what?).

mrocklin:

PR #1201, opened 2017-06-23
Client and worker_client refactor

mrocklin:
This is a major rewrite of how we handle tasks that create clients for complex dynamic workloads.

Previously we would start up a new client every time someone used `worker_client`

```python
def func():
    with worker_client() as c:
        ...
```

We change this in two large ways:

1.  In API we split this into two functions, `get_client` and `secede`.
2.  In implementation we keep around the same client forever

Reusing the same client and requesting the same client from many threads required us to be a bit clever locks, which is probably error prone.  This also exposed some thread-unsafety issues in our ThreadPoolExecutor fork.

This also removes channels.  It was a pain to maintain the same behavior when all of the worker_clients were in fact the same client.


mrocklin:
Rebased to more sensible commits.  Going to merge this soon.

adamklein:
Can I nest future objects into say a dictionary? E.g, instead of `client.submit(f, x)` where `x` is a future, can I have `client.submit(f, {'x': x})`? I tried this approach and get:
   
    Exception ignored in: <object repr() failed>                                                                                                                   
      File "/home/aklein/src/distributed/distributed/client.py", line 282, in __del__                              
        self.release()                                       
      File "/home/aklein/src/distributed/distributed/client.py", line 267, in release                           
       with self.client._lock:       
    AttributeError: 'Future' object has no attribute 'client'

mrocklin:
This brings up a good question of when do we want to materialize futures to track dependencies and when do we want to serialize them without tracking dependencies.  

In the case you describe above I would actually expect that to create a dependency on `x`, not serialize it.

In the case you brought up on Stack Overflow it's actually not clear to me what best behavior is here.  Do we want to finalize the result and create a dependency or move around the collection-holding-futures.

mrocklin:
Here is the behavior I get when trying your example:

```python

In [2]: client = Client()
 
In [3]: future = client.scatter(123)

In [4]: future2 = client.submit(lambda x: x, {'x': future})

In [5]: future2.result()
Out[5]: {'x': 123}
```

adamklein:
I guess I'm a little confused by what we mean when we say `support serialization of dask collections` or a future. For instance, this doesn't work for a dask dataframe, but it does for a concrete value.

	def task1():
		return pd.DataFrame(np.random.rand(10,10))


	def task2(df):
		return df

	def testme():
		with Client() as client:
			ddf = dd.from_delayed([client.submit(task1)])
			ddf = client.persist(ddf)
			fut = client.scatter(ddf)
			#fut = client.scatter(123)
			fut2 = client.submit(task2, fut)
			print(client.compute(fut2, sync=True))

because i get

	ValueError: No global client found

What I want to do is send the handle to the ddf or future to the worker client and manipulate it there. Is this still not possible at the moment? I'd be happy with a way to wrap ddf so it's like `client.submit(task2, serialize(ddf))` - do I have access to such a magical `serialize` function?

adamklein:
Also, why does this fail?

	def task1():
		return pd.DataFrame(np.random.rand(10,10))


	def task2(df):
		return df


	def testme():
		with Client() as client:
			ddf = dd.from_delayed([client.submit(task1)])
			fut2 = client.submit(task2, ddf)
			print(client.compute(fut2, sync=True))

adamklein:
I guess even more fundamentally, I thought I'm replicating your test here, but this fails with the same `ValueError: No global client found`:

        with Client() as client:
            import dask.array as da
            x = da.arange(10, chunks=(5,)).persist()

            def f(x):
                assert isinstance(x, da.Array)
                return x.sum().compute()

            future = client.submit(f, x)
            result = client.compute(future, sync=True)
            print(result)

adamklein:
This *almost* works:

            client1 =>

            ddf = dd.from_delayed(dfs)
            ddf = client.persist(ddf)

            # hold reference to future so it isn't ejected from memory
            from distributed.protocol.pickle import dumps
            storage[ddf_name] = (ddf, dumps(ddf))
            ....
            worker client =>

            from distributed.protocol.pickle import loads
            ddf = loads(ddf_string)
            ....

Sometimes it falls over with:

	  File "/home/aklein/src/architect/architect/core/model.py", line 386, in run_job                                                                                    
		model.initialize(run_context, upstream)                                                                                                                          
	  File "/home/aklein/src/architect/architect/models/manipulators/add_columns.py", line 47, in initialize                                                             
		self.df = self.model( run_context )                                                                                                                              
	  File "/home/aklein/src/architect/architect/core/model.py", line 162, in __call__                                                                                   
		return pull_from_cluster_memory(self, run_context)                                                                                                               
	  File "/home/aklein/src/architect/architect/core/scheduler.py", line 222, in pull_from_cluster_memory                                            
		df = client.compute(ddf.loc[rc.start:rc.end, :], sync=True)                                                                             
	  File "/home/aklein/src/distributed/distributed/client.py", line 1985, in compute                                                          
		result = self.gather(futures)                                                                                                                              
	  File "/home/aklein/src/distributed/distributed/client.py", line 1317, in gather                                                                              
		direct=direct, local_worker=local_worker)                                                                                                                                      
	  File "/home/aklein/src/distributed/distributed/client.py", line 508, in sync                                                                                                     
	  File "/home/aklein/src/distributed/distributed/utils.py", line 234, in sync                                                                                                      
		six.reraise(*error[0])                                                                                                                                                         
	  File "/lib/python3.5/site-packages/six.py", line 686, in reraise                                                                                    
		raise value                                                                                                                                                                                                                    
	  File "/home/aklein/src/distributed/distributed/utils.py", line 223, in f                                                                                                                                                         
		result[0] = yield make_coro()                                                                                                                                                                                                  
	  File "/lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                                                                                               
		value = future.result()                                                                                                                                                                                                        
	  File "/lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                                                                                      
	  File "<string>", line 4, in raise_exc_info                                                                                                                                       
	  File "/lib/python3.5/site-packages/tornado/gen.py", line 1063, in run                                                                                                         
		yielded = self.gen.throw(*exc_info)                                                                                                                                            
	  File "/home/aklein/src/distributed/distributed/client.py", line 1197, in _gather                                                                                                                                                 
		traceback)                                                                                                                                                                                                                     
	  File "/lib/python3.5/site-packages/six.py", line 685, in reraise                                                                                    
		raise value.with_traceback(tb)                                                                                                                                                                                                 
	  File "/home/aklein/src/distributed/distributed/protocol/pickle.py", line 59, in loads                                                                                                                      
		return pickle.loads(x)                                                                                                                                                                                                         
	AttributeError: Can't get attribute 'apply_and_enforce' on <module 'dask.dataframe.core' from '/home/aklein/src/dask/dask/dataframe/core.py'>                

I know I'm doing some weird stuff, but it feels very close :)

mrocklin:
It looks like clients aren't being created on demand.  I'll need to take a look.  This probably won't happen any time today though.  I appreciate the failing tests.

adamklein:
No problem, appreciate the work, it's very promising! Have a great pre-4th of july weekend.

mrocklin:
@adamklein the recent commits may interest you

mrocklin:
This now feels ready to me.  @adamklein if you have a chance to try to break this or @pitrou if you have a chance to review I would appreciate it.

mrocklin:
I'm considering where to put documentation for this.  I may want to put this in the dask/dask repository.

adamklein:
Giving it a look this morning!

adamklein:
This is looking excellent. I'm running into timeout issues on my large task return values but I think that's the problem I hacked before that I simply lost b/c of the worker_client refactor, and I just have to throw in back in there.  I think it's a result of my scheduler being overworked and unable to communicate in a timely fashion. Running locally I see it taking a lot of CPU and memory. Here is that stack trace just for information purposes, but again, think it has nothing to do with your changes (or e.g. the changes regarding de/serialization on a separate thread).

	  File "/lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/comm/core.py", line 181, in connect         
		quiet_exceptions=EnvironmentError)                                                                                                                               
	  File "/lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                                
		value = future.result()                                                                                                                                          
	  File "/lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                                                        
	  File "<string>", line 4, in raise_exc_info                                                                                                                                                                         
	tornado.gen.TimeoutError: Timeout                                                                                                                                  
																																										 
			result = f.result()                                                                                                                                            
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/client.py", line 152, in result             
			self._result, raiseit=False, callback_timeout=timeout)                                                                                                         
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/utils.py", line 234, in sync               
			six.reraise(*error[0])                                                                                                                                           
		  File "lib/python3.5/site-packages/six.py", line 686, in reraise                                                                     
			raise value                                                                                                                                                    
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/utils.py", line 223, in f                 
			result[0] = yield make_coro()                                                                                                                                   
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                               
			value = future.result()                                                                                                                                         
		  File "lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                      
		  File "<string>", line 4, in raise_exc_info                                                                                                                         
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1063, in run                                                               
			yielded = self.gen.throw(*exc_info)                                                                                                                             
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/client.py", line 177, in _result           
			result = yield self.client._gather([self])                                                                                                                       
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                               
			value = future.result()                                                                                                                                       
		  File "lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                      
		  File "<string>", line 4, in raise_exc_info                                                                                                                         
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1063, in run                                                                
			yielded = self.gen.throw(*exc_info)                                                                                                                              
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/client.py", line 1245, in _gather           
			response = yield self.scheduler.gather(keys=keys)                                                                                                                                                 
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                                                                       
			value = future.result()                                                                                                                                                                                          
		  File "lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                     
		  File "<string>", line 4, in raise_exc_info                                                                                                                         
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1063, in run                                                                
			yielded = self.gen.throw(*exc_info)                                                                                                                              
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/core.py", line 423, in send_recv_from_rpc    
			comm = yield self.live_comm()                                                                                                                                     
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                                
			value = future.result()                                                                                                                                           
		  File "lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                        
		  File "<string>", line 4, in raise_exc_info                                                                                                                         
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1063, in run                                                                
			yielded = self.gen.throw(*exc_info)                                                                                                                              
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/core.py", line 399, in live_comm            
			connection_args=self.connection_args)                                                                                                                           
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1055, in run                                                               
			value = future.result()                                                                                                                                          
		  File "lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result                                                      
		  File "<string>", line 4, in raise_exc_info                                                                                                                         
		  File "lib/python3.5/site-packages/tornado/gen.py", line 1063, in run                                                                
			yielded = self.gen.throw(*exc_info)                                                                                                                            
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/comm/core.py", line 190, in connect         
			_raise(error)                                                                                                                                                   
		  File "lib/python3.5/site-packages/distributed-1.17.1+36.g007cbf73-py3.5.egg/distributed/comm/core.py", line 173, in _raise         
			raise IOError(msg)                                                                                                                                              
		OSError: Timed out trying to connect to 'tcp://172.22.174.50:8786' after 5 s: connect() didn't finish in time

mrocklin:
Do you have a sense for what is slowing down your scheduler?  It seems odd to me to see 5s of delay, even under very heavy load (thousands of workers).

You may be interested in [this comment](https://github.com/dask/distributed/issues/1225#issuecomment-312497112).

mrocklin:
Documentation added in https://github.com/dask/dask/pull/2501

adamklein:
I do not know what's slowing down the scheduler, but I see it taking >10Gb memory + 90+% cpu so it's churning hard on something.  I do see in `worker_client(timeout=3, ...` that the timeout function is no longer used. Where does it inherit that timeout information from now, in the client workers? I'll also try the crick + counters hint from your other thread, thank you for pointing that out.

distributed/worker.py
```diff
@@ -2103,6 +2131,40 @@ def story(self, *keys):
...
+        get_client
+        """
+        with self._lock:
+            if self._client:
+                return self._client
+
+            try:
+                from .client import default_client
+                client = default_client()
+            except Exception:
```
pitrou: It would be nice to know why/when this fails. `except Exception` is a bit vague (and might hide programming errors?).

distributed/worker.py
```diff
@@ -2103,6 +2131,40 @@ def story(self, *keys):
...
+                pass
+            else:
+                if client.scheduler.address == self.scheduler.address:
+                    self._client = client
+
+            if not self._client:
+                from .client import Client
+                asynchronous = get_thread_identity() == self.loop._thread_ident
```
pitrou: Why not let the user pass the `asynchronous` argument instead?

distributed/worker.py
```diff
@@ -2119,6 +2181,82 @@ def get_worker():
...
     See Also
     --------
+    get_client
+    worker_client
+    """
+    try:
+        return thread_state.execution_state['worker']
+    except AttributeError:
+        if _global_workers:
+            return _global_workers[-1]()
```
pitrou: What if this returns None?

distributed/worker.py
```diff
@@ -2119,6 +2181,82 @@ def get_worker():
...
+        else:
+            raise
```
pitrou: Don't you want to raise a more informative error message here?

distributed/worker.py
```diff
@@ -2119,6 +2181,82 @@ def get_worker():
...
+
+
+def get_client(address=None):
+    """ Get a client while within a task
```
pitrou: Will it be asynchronous or not? It doesn't seem obvious that this always returns the same kind of client...

distributed/client.py
```diff
@@ -486,13 +488,29 @@ def __init__(self, address=None, loop=None, timeout=5,
...
+
+            Client(asynchronous=True)
+
+        and if we are currently in the same thread as the event loop.
+        """
+        result = self._asynchronous
+        try:
+            if get_thread_identity() != self.loop._thread_ident:
+                result = False
+        except AttributeError:  # AsyncIOLoop doesn't have _thread_ident
```
pitrou: Does it mean we return a bogus result in this case?

distributed/client.py
```diff
@@ -486,13 +488,29 @@ def __init__(self, address=None, loop=None, timeout=5,
...
-        self.start(timeout=timeout, asynchronous=asynchronous)
+        self.start(timeout=timeout)
 
-        from distributed.channels import ChannelClient
-        ChannelClient(self)  # registers itself on construction
         from distributed.recreate_exceptions import ReplayExceptionClient
         ReplayExceptionClient(self)
 
+    @property
+    def asynchronous(self):
```
pitrou: I think either the name of this property and/or its semantics and/or its docstring should be changed, since all three seem to be saying different things :-)

distributed/client.py
```diff
@@ -578,14 +597,16 @@ def start(self, asynchronous=None, **kwargs):
...
             while not self.loop._running:
                 sleep(0.001)
+
         pc = PeriodicCallback(lambda: None, 1000, io_loop=self.loop)
         self.loop.add_callback(pc.start)
         _set_global_client(self)
-        if asynchronous:
+        self.status = 'connecting'
+
+        if self.asynchronous:
```
pitrou: I must admit I fail to understand the intent of the `.asynchronous` property. If the user passed `asynchronous=True`, we may still start an event loop in a separate thread?

mrocklin: The problem arises because we reuse the same client in both contexts.  Consider the following:

```python
def g():
    # I run in a worker thread
    client = get_client()
    return future.result()

async def f():
    async with Client(asynchronous=True, processes=False) as c:
        await c.submit(g)
```

The client in both cases is the same.  We tell the client originally that it is operating in asynchronous mode.  However when it finds itself in the `g` task it realizes that it definitely can't be in asynchronous mode and so switches over within that context.

Ideally we would have a  function that could tell us if we "were in the event loop" but this isn't doable.  So instead we take a signal from the user (this is `asynchronous=True`) but disregard it when it's obviously false.

This approach would currently fail if we were to start a synchronous client and then try to use it asynchronously, but seems to work in all other settings.

distributed/client.py
mrocklin: I've expanded the docstring.  Maybe this helps?  Suggestions welcome.

distributed/client.py
mrocklin: It means that we just go with what the user provided when they defined the client.  We don't overrule them when we know that the situation has changed.

distributed/worker.py
mrocklin: Fixed

distributed/worker.py
mrocklin: We now iterate from most to least recent.  If none are found then we raise.

mrocklin:
@pitrou any response on the asynchronous comments?  In general I agree that this isn't entirely clean.  I do think it's significantly cleaner than before.  I'm not sure yet how to improve on the situation.

adamklein:
@mrocklin I am actually seeing a *major* memory leak in the scheduler on my example in which I am beating up the scheduler. That seems to be what is leading to things to fall over. I rolled back to 1.17.1 and it disappears. I'll try to bisect to figure out on what commit it happens. I don't think I'll be able to get to it today though - most likely 7/5

mrocklin:
I appreciate you digging into it.  Let us know what you find

adamklein:
I did a git bisect and 068c571cfe7aeed16ff3a4bb01541e403e0b4a2f is the commit where my dask stress test starts to make the scheduler blow out memory (and stress the CPU to boot). Could it have to do with the changes in worker_client updating its data dictionary? That's the only thing that looks related to data/memory management when I look through the code.

distributed/client.py
pitrou: > The client in both cases is the same. 

Does it have to? :-) I don't think it would be terribly ruinous to restrict that asynchronous clients cannot be re-used in a synchronous context, and vice-versa.

mrocklin:
@adamklein can you say a bit more about what your stress test is doing?  For example, it would be useful to know if it is scatter/gathering a large amount of data between worker-clients.

distributed/client.py
mrocklin: There is some value to reducing the number of active clients.  I agree though that, especially because of the failure of synchronous-intended clients to work in asynchronous environments, we do need to resolve this.  

I think I'm still hoping for a way to help identify whether we should or should not act in an asynchronous manner that is separate from defining a client as asynchronous or not.

My inclination here is to merge things roughly as they are now.  I think that the user-facing bits of this are decent.  We have a known failure in one case but it's unlikely to be encountered by most users.  This PR is large and I'd like to get it in soon-ish.  I don't have strong confidence that I can find a good solution in the next day or two with other things going on.

mrocklin:
@adamklein if you can try this branch again with the recent commit I'd be
obliged.  No rush.

adamklein:
@mrocklin Unfortunately my example seems to still leak memory and run at high CPU with your latest commit. I'm seeing about 10x performance drop in CPU and memory usage so that I can't do what I used to. I have tasks launching tasks and passing 100Mb's to Gbs as task results between workers. This is certainly an example I'd like to show you offline when everything is in place.

distributed/client.py
mrocklin: One issue that arises is that when we deserialize futures or queues or variables we don't know if the client should be asynchronous or synchronous.  The user doesn't have a chance to tell us in this case.

mrocklin:
@adamklein it would be interesting to see if the scheduler is reporting high communication volume.  There is a plot at `http://scheduler-address:8787/system` that might be of interest

adamklein:
@mrocklin The answer is yes, there is high communication volume. Here are my snapshots: off current dask master =>

snap1

Off the latest PR of #1201 =>
snap2

adamklein:
PS I am using `on_completed`, which I see calls into `self.client._gather([self])` around line 177 of client.py; this has default parameter `direct=False`, and probably `local_worker=None` as well. I assume we want either direct=True or local_worker=(something). Granted this is just from reading code, I haven't done any tracing so I could be totally off base here.

adamklein:
PPS, I just confirmed if I put `direct=True` into `self.client._gather([self], direct=True)` in the above mentioned line in client.py, the problem disappears. I don't know if that has any negative consequences, but it reverts the behavior to what happens on the master branch in terms of memory and bandwidth utilization.

mrocklin:
I have rebased commits to be somewhat more clean.  Merging on passed tests.

mrocklin:
> PPS, I just confirmed if I put direct=True into self.client._gather([self], direct=True) in the above mentioned line in client.py, the problem disappears. I don't know if that has any negative consequences, but it reverts the behavior to what happens on the master branch in terms of memory and bandwidth utilization.

It looks like `Future.result` would suffer from this as well.  I'll try to resolve it in another PR after this gets merged.

mrocklin:
@adamklein thanks for tracking that down.

PR #1206, opened 2017-06-27
Fix #1204: allow logging config using logging.dictConfig()

### LINKED ISSUE #1204: Improve configuration options regarding logging
Python's standard library offers two useful ways to initialize loggers from a [file](https://docs.python.org/2/library/logging.config.html#logging.config.fileConfig)
or a [dictionary](https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig).

Dask's `config.yaml` currently contains a `logging` field. I propose to change the field so that it can be fed to the `logging.config.dictConfig` function. The relevant part of the default `config.yaml` would then for example look like this:

```yaml
logging:
  formatters:
    default:
      format: '%(name)s - %(levelname)s - %(message)s'
  handlers:
    console:
      class: logging.StreamHandler
      stream: ext://sys.stderr
  loggers:
    distributed:
      level: info
      handler: console
      formatter: default
    distributed.client:
      level: warning
      handler: console
      formatter: default
    bokeh:
      level: warning
      handler: console
      formatter: default
    tornado:
      level: critical
      handler: console
      formatter: default
    tornado.application:
      level: error
      handler: console
      formatter: default
``` 

To also gain the possibility to provide a file config, it could either be fed to `logging.config.fileConfig` in case it is a string. Or another field like `logging-file-config` could be introduced for that purpose.

The downside is increasing complexity of the config file. But gaining more flexibility in setting up logging would be a huge gain.

What are your thoughts on this?

pitrou:


pitrou:
Still need to add docs.

PR #1211, opened 2017-06-28
Add support for logging file config

jneuff:


distributed/tests/test_config.py
```diff
@@ -204,3 +206,60 @@ def test_logging_extended():
...
+level=INFO
+handlers=console
+qualname=foo
+
+[logger_foo_bar]
+level=ERROR
+handlers=console
+qualname=foo.bar
+"""
+    with tempfile.NamedTemporaryFile() as logging_config:
```
pitrou: You should pass `mode="w"` since you're writing a `str` to it.

distributed/tests/test_config.py
```diff
@@ -204,3 +206,60 @@ def test_logging_extended():
...
+        logging_config.file.write(logging_config_contents)
+        logging_config.file.seek(0)
```
pitrou: I'm not sure why you're doing this. Perhaps you want `flush()`?

distributed/tests/test_config.py
```diff
@@ -204,3 +206,60 @@ def test_logging_extended():
...
+        dask_config = {'logging-file-config': logging_config.name}
+        with new_config(dask_config):
```
pitrou: You should really use a separate process, as in the tests above. Otherwise, subsequent tests will get an unexpected logging configuration.

pitrou:
This looks mostly good to me!
As for the Windows test failure, I think you'll have to bite the bullet and create the file yourself using `mkstemp()`, because `NamedTemporaryFile` doesn't seem usable for that under Windows. To quote the Python docs:
>  Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)

jneuff:
Thanks for pointing that out! I still stick to `NamedTemporaryFile` as it provides a higher level interface.

pitrou:
I took the liberty of doing a tiny cleanup :) Now merging. Thank you for posting this PR!

PR #1217, opened 2017-06-29
[WIP] Rewrite batched communications

pitrou:
This was prompted by the BatchedSend-induced memory leak. The new design doesn't have a background-running loop but only schedules a callback when required, which should remove the core issue.

Also the new "batched" object is an actual Comm, which would ideally simplify and streamline things. Unfortunately the requirement to instantiate a "batched" before having an underlying Comm reintroduces some of the complication I hoped to dispose of.

### DIFF
--- a/distributed/channels.py
+++ b/distributed/channels.py
@@ -49,14 +49,14 @@ class ChannelScheduler(object):
 
         comm = self.scheduler.comms[client]
         for type, value in self.deques[channel]:
-            comm.send({'op': 'channel-append',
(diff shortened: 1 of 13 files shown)


pitrou:
For now there are spurious "Comm closed" exceptions in the logs that need to be cleaned up before this can go in...

pitrou:
Debugging this is rather tedious and boring so I'll let it rot.

mrocklin:

PR #1218, opened 2017-06-29
Offload (de)serialization to a separate thread

pitrou:
This is conditioned on the input size, to avoid increasing the overhead of processing
small administrative messages.

### DIFF
--- a/distributed/comm/utils.py
+++ b/distributed/comm/utils.py
@@ -1,43 +1,88 @@
 from __future__ import print_function, division, absolute_import
 
+from concurrent.futures import ThreadPoolExecutor
 import logging
 import socket
 
+from tornado import gen
+
 from .. import protocol
-from ..utils import get_ip, get_ipv6
+from ..compatibility import finalize
+from ..sizeof import sizeof
+from ..utils import get_ip, get_ipv6, nbytes
 
 
 logger = logging.getLogger(__name__)
 
 
+# Offload (de)serializing large frames to improve event loop responsiveness
+FRAME_OFFLOAD_THRESHOLD = 10 * 1024 ** 2   # 10 MB
+
+
+class Offloader(object):
+
+    def __init__(self):
+        self.executor = ThreadPoolExecutor(max_workers=1)
+        self._finalizer = finalize(self, self.executor.shutdown)
+
+    def submit(self, fn, *args, **kwargs):
+        return self.executor.submit(fn, *args, **kwargs)
+
+
+_offloader = Offloader()
+
+
+def offload(fn, *args, **kwargs):
+    return _offloader.submit(fn, *args, **kwargs)
+
+
+@gen.coroutine
 def to_frames(msg):
     """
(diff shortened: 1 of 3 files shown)


distributed/comm/utils.py
```diff
@@ -1,43 +1,88 @@
...
+
+    def submit(self, fn, *args, **kwargs):
+        return self.executor.submit(fn, *args, **kwargs)
+
+
+_offloader = Offloader()
+
+
+def offload(fn, *args, **kwargs):
+    return _offloader.submit(fn, *args, **kwargs)
```
mrocklin: Why the need for the `Offloader` class here?  Would a raw `ThreadPoolExecutor` work just as well?  Is this necessary because of the `finalize` cleanup code?  Does finalize require being part of a class?

distributed/comm/utils.py
```diff
@@ -1,43 +1,88 @@
...
 
 
+# Offload (de)serializing large frames to improve event loop responsiveness
+FRAME_OFFLOAD_THRESHOLD = 10 * 1024 ** 2   # 10 MB
+
+
+class Offloader(object):
+
+    def __init__(self):
+        self.executor = ThreadPoolExecutor(max_workers=1)
```
mrocklin: We might want a couple of workers here?  We do accept multiple messages concurrently.  Having multiple threads here might help fast messages get through in the presence of large messages.

pitrou: Well, that depends. On the workers, that might overcommit worker threads? (do we care about this issue?)

The problem is we need a heuristic. Something like `min(4, cpu_cores())`?

distributed/comm/utils.py
pitrou: There is no real need, but I find it nice to avoid too many global variables.

pitrou:
(Travis-Ci seems not to trigger any builds from Github anymore)

distributed/comm/utils.py
mrocklin: We're going to have a global `_offloader` regardless.  Adding another for the finalizer doesn't seem that bad to me.

I generally prefer to stick to using common types for things when possible.  I personally find it difficult to quickly understand code when it creates many of its own abstractions.

distributed/comm/utils.py
mrocklin: In my experience heavy communication and heavy computation rarely overlap.  In the case that they do I'd rather have communication take precedence.  I think it makes sense to have somewhere between 2-4 communication threads.  Although really all of this is probably premature before we find some application that cares.

pitrou: Well, having multiple threads early might help expose any potential issues.

mrocklin:
This looks good to me.  +1.  

I ran both a basic small-data benchmark and a more communication heavy benchmark (with chunks > size 10MB) and things seemed about the same.

pitrou:
Thank you. I'm merging now.

PR #1219, opened 2017-06-29
Improve Comm repr() and add Comm.local_address

pitrou:


pitrou:
Just a simple change with a couple additional tests.

Issue #1205, opened 2017-06-26
Questions around emptying and resuming channels


MathMagique:
Hi! I have workflow that is expected to have some side effects such as writing results into a database. Scheduling the necessary jobs using the `Client` is easy enough, but I do not want to keep my Python script/web request/whatever uses the `Client` running for as long as it takes to compute all jobs.

The [documentation](http://distributed.readthedocs.io/en/latest/channels.html) suggests to create a `Channel` and to append any future worth doing (even when no "real" client seems to care). I did that, and it works as expected.

However, this leads to the channel raking up more and more futures.
* Am I assuming correctly that the results associated with these futures will be kept in memory/on disk around the cluster?
* I cannot find a way to empty a channel, e.g., via `channel.reset()` or something similar.
* When I tried to reset the whole cluster with `client.restart()`, I noticed that channels are not emptied. Instead, the stored futures are listed with `status: pending`.

I also used `channel.stop()` in the (false) hope that it would clean off its contents. I had to realize, though, that I can no longer `append()` to the channel, getting a weirdly used `StopIteration` exception I would otherwise expect from a `next()` method. 
* Is there a way to resume a channel?
* Even after resetting the cluster with `cluster.restart()`, the channel appears to be `stopped`.

MathMagique:
_Side remark that does not answer any of the questions above, but provides an alternative solution to the use case_

In the meantime, I also found out about `client.publish_dataset()` and [related functions](http://distributed.readthedocs.io/en/latest/publish.html), which seems better for my use case.

mrocklin:
Yeah, channels aren't perfect for this use case.  It would be nice to have a fire-and-forget option for tasks, but this doesn't yet exist.

For the moment you might try Queues, which are probably a bit more sensible than Channels.  Once someone pulls an element from the queue it leaves the queue entirely.  You could even set up a long-running task to pull from the queue constantly.

```python
def pull_from_queue(name):
    with worker_client() as c:
        q = Queue(name, client=c)
        while True:
            q.get()

future = client.submit(pull_from_queue, queue_name)
```

This is still a bit of a hack, but should result in better behavior.

MathMagique:
After playing around some, I noticed two issues with your suggestion:
* The watcher would never shut down
* The watcher itself needs to be kept alive to do the cleanup. Otherwise, tasks are kept alive by the queue indefinitely.

I slightly modified your code and came up with the following:

```
from dask.distributed import worker_client, Queue, wait
[...]

PR #1221, opened 2017-06-29
Add fire_and_forget function

mrocklin:
Run tasks at least once, even if we release the futures

Under normal operation Dask will not run any tasks for which there is not an
active future (this avoids unnecessary work in many situations).  However
sometimes you want to just fire off a task, not track its future, and expect it
to finish eventually.  You can use this function on a future or collection of
futures to ask Dask to complete the task even if no active client is tracking
it.

The results will not be kept in memory after the task completes (unless there
is an active future) so this is only useful for tasks that depend on side
effects.

Parameters
----------

obj: Future, list, dict, dask collection
    The futures that you want to run at least once

Examples
--------

    >>> fire_and_forget(client.submit(func, *args))

Fixes https://github.com/dask/distributed/issues/1205

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -5,7 +5,7 @@ from .core import connect, rpc
 from .deploy import LocalCluster
 from .diagnostics import progress
 from .client import (Client, Executor, CompatibleExecutor,
-                     wait, as_completed, default_client)
+                     wait, as_completed, default_client, fire_and_forget)
 from .nanny import Nanny
 from .queues import Queue
 from .scheduler import Scheduler
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2967,6 +2967,36 @@ def futures_of(o, client=None):
     return list(futures)
 
 
+def fire_and_forget(obj):
+    """ Run tasks at least once, even if we release the futures
+
+    Under normal operation Dask will not run any tasks for which there is not
+    an active future (this avoids unnecessary work in many situations).
+    However sometimes you want to just fire off a task, not track its future,
+    and expect it to finish eventually.  You can use this function on a future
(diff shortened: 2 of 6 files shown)


MathMagique:
Hi @mrocklin! Thank you very much for adding the feature so quickly! I have a small comment. Since I agree that fire and forget is only useful for functions with side effects, should fire and forget then imply `pure=False`?

With the syntax you have chosen, enforcing `pure=False` for the user is difficult, as `fire_and_forget` acts on already existing futures. A `fire_and_forget=True` parameter to `submit()` could easily enforce `pure=False`. Alternatively, you might just add a `pure=False` to the examples in the documentation ;-).

mrocklin:
> Since I agree that fire and forget is only useful for functions with side effects, should fire and forget then imply pure=False?

While I think that this is mostly useful for functions with side-effects I'm not inclined to force `pure=False` behavior just yet.  I agree that this is common case but I think that we should wait a bit to see how it's used before removing this option.  

Adding a `fire_and_forget=` keyword to submit makes sense, although we would also have to add it to `compute`, `persist`, and `map`.  I'm similarly inclined to start with the separate function and see how it goes, before adding it into the base API.  

Regardless, I agree with you that we could use better documentation around all of this stuff.  I'll try to write up a more comprehensive docpage around dynamic graph generation after this and #1201 have been merged in.  Does this sound like a sufficient first step?

MathMagique:
Of course, that sounds like a reasonable thing to do. I am very happy that you took the time to implement fire and forget at all in such a short time.

mrocklin:
It ended up being much easier than expected.  This has been a
long-requested feature.  It is nice to get it out there.

jlowin:
Excited to see this added, thanks @mrocklin!

Issue #1223, opened 2017-06-30
AssertionError when using sparse dask array with the distributed scheduler


ogrisel:
Code to reproduce:

```python
import dask, distributed
import dask.array as da
import sparse

client = distributed.Client('localhost:8786')
dask.set_options(get=client.get)

x = da.random.random((100, 10), chunks=(10, 10))
x[x < 0.95] = 0
s = x.map_blocks(sparse.COO)
s.sum(axis=0)[:10].compute()
```

I get on the client:

```
ValueError: Could not find dependent ('sum-d463f48d7b3a2898552ef145502597fe', 0, 0).  Check worker logs
```

The tracebacks in the output of the worker seem to be messed up by concurrent threads, but here is the end of one of them:

```
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/tornado/gen.py", line 1069, in run
    yielded = self.gen.send(value)
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/distributed/comm/tcp.py", line 169, in read
    msg = from_frames(frames, deserialize=self.deserialize)
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/distributed/comm/utils.py", line 30, in from_frames
    return protocol.loads(frames, deserialize=deserialize)
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/distributed/protocol/core.py", line 120, in loads
    fs = merge_frames(head, fs)
  File "/home/ogrisel/.virtualenvs/py36/lib/python3.6/site-packages/distributed/protocol/utils.py", line 58, in merge_frames
    assert sum(lengths) == sum(map(len, frames))
AssertionError
```

Versions:

- dask 0.15.0
- distributed 1.17.1
- sparse 0.1.0

Note: this problem cannot be reproduced when running in dask without the distributed scheduler.

mrocklin:
Thank you for the excellent bug report.  I've reproduced locally.  Checking it out now.

ogrisel:
no pbm, thanks for investigating, I was too lazy :)

PR #1224, opened 2017-06-30
Use nbytes more consistently throughout protocol code

mrocklin:
Historically we always moved bytestrings around and so used the `len`
function to determine how many bytes there were.  In the last release we
started supporting memoryviews with itemsizes greater than one.  These
itemsizes were used as hints for the compression code.  We started using
a new function, `nbytes`, which provided the bytes of an object
regardless of type (for bytestrings, buffers, and memoryviews).

However we didn't smoothly transition all uses of `len` to `nbytes`.  In
particular some of the sparse array code wasn't transitioned over.  This
commit goes through the protocol more thoroughly replacing uses of `len`
with `nbytes`.

Fixes https://github.com/dask/distributed/issues/1223

cc @ogrisel

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -213,7 +213,7 @@ class TCP(Comm):
             stream = None
             convert_stream_closed_error(self, e)
 
-        raise gen.Return(sum(map(len, frames)))
+        raise gen.Return(sum(map(nbytes, frames)))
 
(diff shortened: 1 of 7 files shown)


ogrisel:
LGTM, thanks for the quickfix :)

mrocklin:
Good error reports tend to produce quick fixes :)

PR #1228, opened 2017-07-03
Warn on long event loop delay

mrocklin:
We emit a warning whenever the event loop is delayed by more than one
second.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -133,8 +133,14 @@ class Server(object):
 
     def _measure_tick(self):
         now = time()
-        self.digests['tick-duration'].add(now - self._last_tick)
+        diff = now - self._last_tick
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @kuanb @adamklein do we want to do something like this?  This would make your problems more evident, but also somewhat loud in the logs

adamklein:
I'll add it in and see what kind of verbosity I see.

adamklein:
I like this. It's not too bad at all. I definitely get warnings. I just don't know what causes them... yet

kuanb:
Sorry I am late to the party on this one - I just wanted to say I second @adamklein. This looks really helpful, esp in helping cue the user into roughly where the issue is emanating from. Thanks.

PR #1229, opened 2017-07-03
Avoid using lock in Future.__del__

mrocklin:
This caused some odd deadlocks


pitrou:
That will only make things marginally better. `__del__` can be called at any point, so it could be called while inside `decref` or `incref`. To avoid such issues, a possible solution should be to use `loop.add_callback` in `__del__` (at the cost of delaying releasing resources slightly).

mrocklin:
`__del__` now adds a decref call onto the event loop.  Merging soon if there are no further comments.

PR #1232, opened 2017-07-04
Fix #1158: Make socket backlog configurable

### LINKED ISSUE #1158: Add a configuration entry for TCP socket backlog

pitrou:
Like PR #1199, but without the additional stress test.


mrocklin:
+1 from me

Issue #1230, opened 2017-07-03
Accept unicode or byte string keys on streams


nicoleepp:
It would be great if the dask.distributed scheduler could accept byte or unicode keys and not just strings of type `str` as checked by `validate_key`.

This issue came up while integrating the dask.distributed scheduler with julia client/workers in https://github.com/invenia/DaskDistributedDispatcher.jl, since julia msgpack will encode strings as either byte strings or unicode strings. Currently, this integration will not be able to support python 3.5 since it adds a `bytes` class. Python 2 works since it treats byte strings as `str` types.

Sample error when using python 3.5:
```distributed.scheduler - ERROR - Unexpected key type <class 'bytes'> (value: b'Int64-4635863187378904022')```

mrocklin:
Sure. I'll take a look at this tomorrow.   Thanks for bringing it up.  Is
there anything else we can do to assist Invenia?

mrocklin:
I threw up an experimental PR here: https://github.com/dask/distributed/pull/1234

We'll see if anything breaks.  You might also consider resolving this on the Julia side.  Python's msgpack library added text-as-text a few years ago.  Msgpack itself apparently extended it's spec to include this.

iamed2:
@mrocklin How about unicode keys? 

The `String` type in Julia is a unicode string, which would map to `unicode` in Python 2. Right now we use `String` in Julia in most cases except when dealing with `dask.distributed` keys, where we need to use byte arrays (ends up being an opaque `Vector{UInt8}` type). We can get this efficiently without copies but having byte arrays around instead of strings still makes the code harder to read and harder to debug.

mrocklin:
First, have you considered just using Python 3?  Python 2 is near end-of-life.

mrocklin:
Alternatively, you might submit a PR similar to #1234 that supports unicode keys in Python 2.  If it doesn't impact performance then I would have no issue merging such a PR.  I personally though am somewhat slammed.

iamed2:
I tried in https://github.com/dask/distributed/pull/1328 but a bunch of tests started freezing

PR #1234, opened 2017-07-04
Support bytestrings as keys

mrocklin:
This is useful to other languages where msgpack may interpret text as
bytes.

Fixes https://github.com/dask/distributed/issues/1230

cc @nicoleepp

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -4292,5 +4292,16 @@ def test_identity(c, s, a, b):
     assert s.id.lower().startswith('scheduler')
 
 
+@gen_cluster(client=True)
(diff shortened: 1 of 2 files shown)


mrocklin:
@nicoleepp if you get a chance to try this out before we merge to see if it helps with Invenia I'd appreciate it.

    pip install git+https://github.com/mrocklin/distributed@bytes-keys

nicoleepp:
I'll test it asap and let you know how it goes.

nicoleepp:
Tested this for the DaskDistributedScheduler use case and it solves the issue with not being able to use python3. 👍  Thanks for taking a look at this.

PR #1235, opened 2017-07-05
Add memory monitor periodic callback

mrocklin:
This periodically checks memory according to psutil and, if we come close to
exceeding limits it pushes data down to disk.


mrocklin:

PR #1241, opened 2017-07-06
Fix testing of blosc compression

QuLogic:
~~Based on [`test_compress_numpy`](https://github.com/dask/distributed/blob/master/distributed/protocol/tests/test_numpy.py#L143), [`test_compression_takes_advantage_of_itemsize`](https://github.com/dask/distributed/blob/master/distributed/protocol/tests/test_numpy.py#L196), and [`test_maybe_compress_memoryviews`](https://github.com/dask/distributed/blob/master/distributed/protocol/tests/test_protocol.py#L217), I believe that blosc is expected to be the default when available. However, it appears that the line setting it as default is missing, and is now added by this PR.~~

I'm a bit confused why a) this never failed on Travis, and b) the latest build of master on Travis is 8 months old.

Install the correct package on Travis and skip a test that shouldn't be run on Python 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ services:
 
 env:
   matrix:
-    - PYTHON=2.7 PACKAGES="blosc futures faulthandler"
-    - PYTHON=3.5 COVERAGE=true PACKAGES=blosc CRICK=true
+    - PYTHON=2.7 PACKAGES="python-blosc futures faulthandler"
+    - PYTHON=3.5 COVERAGE=true PACKAGES=python-blosc CRICK=true
     - PYTHON=3.6
     - HDFS=true PYTHON=2.7
-    - HDFS=true PYTHON=3.5 PACKAGES=blosc
+    - HDFS=true PYTHON=3.5 PACKAGES=python-blosc
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


QuLogic:
Ah, I think I see the problem with Travis; `blosc` is being used but in conda(-forge), that's the C library! So blosc is never really being tested.

QuLogic:
Unfortunately, now that blosc is truly enabled, we run into the problem that random tests deadlock somehow and that's a bit deeper than my knowledge goes.

QuLogic:
It seems that `test_compression_2` is the only one that _doesn't_ assume that blosc takes priority.

mrocklin:
We don't want blosc to be the default.  We currently use it only when passed a `memoryview`.

mrocklin:
It *would* be good to add the `python-blosc` package though.  That was a genuine error.

QuLogic:
Oh, I see; I think the skips were added after the release, which is why I still see them as errors. Based on the corrected Travis run, I think `test_maybe_compress_memoryviews` should be skipped on Python 2 as well.

mrocklin:
As always, thanks @QuLogic !  My apologies for the delay in merging this.

PR #1244, opened 2017-07-07
Add failing test for coro_test on OSX

mrocklin:
This works fine on my Linux machine, but fails on my OSX machine.

cc @kszucs


mrocklin:
This also means that all other asyncio tests are failing.  This test also failed at the last release.

PR #1253, opened 2017-07-11
Allow recreate on persisted collections

martindurant:
Previously only worked for single futures (which could have been pointing
to a compute of a collection), now works for persisted collections.

Suggestions for how to test for a non-persisted collection and improve
the error message?

Suggestions on whether there should be an action taken if local recreation
doesn't raise anything (this has happened to me over credentials).

(thanks, @jcrist )

### DIFF
--- a/distributed/recreate_exceptions.py
+++ b/distributed/recreate_exceptions.py
@@ -3,7 +3,7 @@ from __future__ import print_function, division, absolute_import
 import logging
 from tornado import gen
 from .client import futures_of, _wait
-from .utils import sync
+from .utils import sync, tokey
 from .utils_comm import pack_data
 from .worker import _deserialize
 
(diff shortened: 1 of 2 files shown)


mrocklin:
> Suggestions for how to test for a non-persisted collection 

What would this look like?  Presumably this works now for `client.compute(collection)` (because you say that it works well for futures) and if it now works well for `client.persist(collection)` then what is left?

> and improve the error message?

What is the error message like currently?

martindurant:
> non-persisted collection

The function works for a single future

    f = c.submit(func, args)   # f comes back with an error

or a persisted collection containing an errored future (gets first one if there are multiple)

    df2 = c.persist(df)   # one or more futures of df2 has error

but not with the un-persisted `df`. The error message is `ValueError: No errored futures passed
`

mrocklin:
What are your thoughts on this @martindurant ?  What do you think we should do here?

martindurant:
I'd say the change here is good and should be merged; the questions were for any further improvement in the messaging of what's going on.

Issue #1015, opened 2017-04-12
dask-worker --memory-limit not working?


matyasselmeci:
Limiting the memory used by a dask-worker using the --memory-limit option seems to have no effect.

Using Python 2.7.5 on a Scientific Linux 7.3 host, with dask 0.14.1 and distributed 1.16.1, I set up a scheduler and a single worker as follows:
```
dask-scheduler --scheduler-file /tmp/schedfile &
sleep 5
dask-worker --no-nanny --no-bokeh --nprocs 1 --nthreads 1 --memory-limit=250e6 \
    --scheduler-file /tmp/schedfile &
```
I ran the following client code to test:
```
import dask, dask.distributed, dask.array
cli = dask.distributed.Client(scheduler_file='/tmp/schedfile')
x = dask.array.random.random((2000, 2000), chunks=(10, 10))
y = x.T * x
print(cli.gather(cli.compute(y.sum()))
```
and watched the memory usage via `ps -e -o rss,command | grep python | grep -v grep`.

I expected the RSS of the dask-worker process to stay below 250K (or just slightly above), but it was around 450K for most of the calculation. With larger array sizes, the RSS rises even more.

Is this a bug? If not, what is the best way to guarantee that a dask-worker does not go over a hard limit, memory-wise?

pitrou:
> I expected the RSS of the dask-worker process to stay below 250K (or just slightly above), but it was around 450K for most of the calculation.

You mean 250MB, right?

matyasselmeci:
Yes, my mistake. `ps` reports memory in K.

mrocklin:
I can't immediately explain why you would be seeing significantly more memory use, however I should note that the `--memory-limit` mechanism isn't a hard limit.  Instead it is a limit at which Dask starts to push data to disk.  It doesn't track the process at all, it only tracks the data that it keeps in the worker's data storage dictionary.  When the estimated number of bytes in this dictionary exceeds the limit then it starts writing to the filesystem.

So many things could be going on here, maybe the executing functions are using lots of memory.  Maybe the filesystem is using memory (though I wouldn't expect this to be assigned to the process).  Maybe Python isn't freeing memory quickly after garbage collection.

If you want a hard limit on the process you'll probably be better served using operating system utilities like cgroups.

matyasselmeci:
When Dask pushes data to disk, does it write to the `/tmp/worker-XXXXXX/storage` directory it creates? I was watching it while running the test, but I didn't see any files.

[...]

Issue #1179, opened 2017-06-16
Series.values.compute() leads to "TypeError: can't concat buffer to bytearray"


bluenote10:
Running a local `dask-scheduler` + `dask-worker` pair, the following code leads to a crash of the worker:

```
import dask.dataframe as dd
from dask.distributed import Executor
e = Executor('127.0.0.1:8786', set_as_default=True)
df = pd.DataFrame({"A": [1, 2, 3] * 10})
ddf = dd.from_pandas(df, npartitions=3)
ddf["A"].values.compute()
```

The worker crashes with:

```
Traceback (most recent call last):
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/distributed/core.py", line 259, in handle_comm
    result = yield result
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/distributed/worker.py", line 439, in get_data
    compressed = yield comm.write(msg)
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/tornado/gen.py", line 292, in wrapper
    result = func(*args, **kwargs)
  File "/home/PHI-TPS/fkeller/.virtualenvs/dask/local/lib/python2.7/site-packages/distributed/comm/tcp.py", line 196, in write
    stream.write(frame)
[...]

PR #1255, opened 2017-07-13
Added explicit worker GC (throttled)

bluenote10:
This PR adds explicit garbage collections to mitigate the memory issues observed in https://github.com/dask/distributed/issues/1015 and https://github.com/dask/zict/issues/19.

The implementation is throttled so that it does not lead to excessive GC calls. As discussed in https://github.com/dask/zict/issues/19 calling GC from the inner scopes like `put_key_in_memory` or `update_data` would not collect the current `value`. The ideal solution would trigger it on the outermost scope whenever a large value has been persisted after releasing the reference to the value. However these places are tricky to find. Currently I'm triggering from `release_key`, `gather`, and `execute`, which is hopefully frequent enough.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -39,6 +39,8 @@ from .config import config
 from .metrics import time
 
 
+thread_state = threading.local()
+
 logger = logging.getLogger(__name__)
 
 
@@ -222,11 +224,13 @@ def sync(loop, func, *args, **kwargs):
             if main_tid == get_thread_identity():
                 raise RuntimeError("sync() called from thread of running loop")
             yield gen.moment
+            thread_state.asynchronous = True
             result[0] = yield make_coro()
         except Exception as exc:
             logger.exception(exc)
             error[0] = sys.exc_info()
         finally:
+            thread_state.asynchronous = False
             e.set()
 
     loop.add_callback(f)
@@ -236,6 +240,28 @@ def sync(loop, func, *args, **kwargs):
         six.reraise(*error[0])
     else:
         return result[0]
+
+
+@contextmanager
+def set_thread_state(**kwargs):
+    old = {}
+    for k in kwargs:
+        try:
+            old[k] = getattr(thread_state, k)
+        except AttributeError:
+            pass
+    for k, v in kwargs.items():
+        setattr(thread_state, k, v)
+    try:
+        yield
+    finally:
+        for k in kwargs:
+            try:
+                v = old[k]
+            except KeyError:
+                delattr(thread_state, k)
+            else:
+                setattr(thread_state, k, v)
 
 
 @contextmanager
@@ -824,6 +850,28 @@ def format_bytes(n):
     if n > 1e3:
         return '%0.2f kB' % (n / 1000)
     return '%d B' % n
+
+
+def asciitable(columns, rows):
+    """Formats an ascii table for given columns and rows.
+
+    Parameters
+    ----------
+    columns : list
+        The column names
+    rows : list of tuples
+        The rows in the table. Each tuple must be the same length as
+        ``columns``.
+    """
+    rows = [tuple(str(i) for i in r) for r in rows]
+    columns = tuple(str(i) for i in columns)
+    widths = tuple(max(max(map(len, x)), len(c))
+                   for x, c in zip(zip(*rows), columns))
+    row_template = ('|' + (' %%-%ds |' * len(columns))) % widths
+    header = row_template % tuple(columns)
+    bar = '+%s+' % '+'.join('-' * (w + 2) for w in widths)
+    data = '\n'.join(row_template % r for r in rows)
+    return '\n'.join([bar, header, bar, data, bar])
 
 
 if PY2:
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -948,23 +948,6 @@ else:
         else:
             return frame.nbytes
 
-
-class ThrottledGC(object):
-    """
-    Allows to run throttled garbage collection as a countermeasure to e.g.:
-    https://github.com/dask/zict/issues/19
-    """
-    def __init__(self, min_interval_in_sec=1.0):
-        self.min_interval_in_sec = min_interval_in_sec
-        self.last_collect = time()
-        #gc.collect()
-
-    def collect(self, force_gc=False):
-        new_time = time()
-        if force_gc or new_time - self.last_collect > self.min_interval_in_sec:
-            gc.collect()
-            self.last_collect = new_time
-
 @contextmanager
 def time_warn(duration, text):
     start = time()
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -957,7 +957,7 @@ class ThrottledGC(object):
     def __init__(self, min_interval_in_sec=1.0):
         self.min_interval_in_sec = min_interval_in_sec
         self.last_collect = time()
-        gc.collect()
+        #gc.collect()
 
     def collect(self, force_gc=False):
         new_time = time()
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -858,7 +858,7 @@ class ThrottledGC(object):
         self.last_collect = time.time()
         gc.collect()
 
-    def collect(self, force_gc):
+    def collect(self, force_gc=False):
         new_time = time.time()
         if force_gc or new_time - self.last_collect > self.min_interval_in_sec:
             gc.collect()
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -14,7 +14,6 @@ import os
 import re
 import shutil
 import socket
-import time
 from importlib import import_module
 
 import six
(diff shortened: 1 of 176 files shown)


distributed/worker.py
```diff
@@ -1563,12 +1566,7 @@ def put_key_in_memory(self, key, value, transition=True):
         if stop - start > 0.020:
             self.startstops[key].append(('disk-write', start, stop))
 
-        if key not in self.nbytes:
-            self.nbytes[key] = sizeof(value)
-
-        if key not in self.types:
-            self.types[key] = type(value)
-
+        self.nbytes[key] = sizeof(value)
```
bluenote10: Note: I changed this because it is probably safer to just update the `nbytes` value instead of relying that it can't change. For `self.types` this was already done that way and the `if` was probably a left-over.

mrocklin:
Sorry for the delay here, I was away at a conference.

A couple of concerns:

1.  This puts GC on the main event loop thread.  Does this matter?  I wouldn't be surprised if `gc.collect()` held the GIL explicitly while collecting, but if it doesn't then it would be far more preferable to keep this off of the main thread.
2.  Rather than make a new class for this perhaps we can just add it as a `PeriodicCallback`

bluenote10:
Some thoughts:

1. I would be very surprised if the GC does not hold the GIL during collection, i.e., it will have to briefly stop the event loop anyway.
2. I was thinking about a periodic solution as well, but in the end, it's not an ideal solution. On the one hand the period needs to be fairly small, because already ~1 sec of heavy spill-to-disk-swaps can produce large amounts of uncollected garbage, causing workers to go out-of-memory. On the other hand, triggering with 1 second intervals would probably have a measurable on CPU-bound tasks. A smart trigger would avoid these issues.

distributed/worker.py
mrocklin: We assume immutability.  Additionally sizeof can take some time while type is free.

distributed/utils.py
```diff
@@ -13,6 +14,7 @@
 import re
 import shutil
 import socket
+import time
```
mrocklin: from .metrics import time

The windows `time.time` function has a 1s precision limit.

distributed/worker.py
```diff
@@ -1809,6 +1809,10 @@ def release_key(self, key, cause=None, reason=None, report=True):
                 self.batched_stream.send({'op': 'release',
                                           'key': key,
                                           'cause': cause})
+            self._throttledGC.collect(
+                force_gc=True if nbytes_to_free > 10 * 2**20 else False
```
mrocklin: Could probably drop the ternary statement and just replace with the condition.

    force_gc = nbytes_to_free > 10 * 2**20

mrocklin:
I'm generally fine with the code here.  Unfortunately it looks like the tests are not.  This change causes a variety of tests to fail in interesting ways.  I suspect some subtle multi-threading interaction.

bluenote10:
I was trying to reproduce the problems locally, but the tests just work for me. There was one remaining issue regarding the usage of `nbytes_to_free`, but now the set of failed tests seems to have changed again (one of the failures is rather a Travis timeout, right?). Do the tests fail randomly or is there a clear pattern? How can we solve this?

mrocklin:
Yeah, welcome to the wonderful world of concurrent debugging.  The virtual machines on travis-ci are *very* slow and so are quite good at catching subtle bugs that sneak by on faster machines.  I recommend trying to run a few of the failing tests in a for loop like the following:

    for i in {1..100}; do distributed/tests/test_stress.py::test_stress_1 --pdb ; done

This might help you catch some of the errors on your local machine.

mrocklin:
It's also entirely possible though that calling `gc.collect` in this way is subtly dangerous though and that we'll need to try another approach.

bluenote10:
How can garbage collection be dangerous? As far as I can see the explicit garbage collection happens in places where an automatic garbage collection could trigger anyway, so it's not possible to rely on GC not happening, right? Or am I missing that there are periods with `gc.disable()`?

I was rerunning all the failed tests (from the Travis Python 2.7 build) in a brute-force attempt (see commands below). I couldn't get the tests to fail except for `distributed/tests/test_steal.py::test_steal_expensive_data_slow_computation`. However, locally this test fails due to the buffer concatenation issue https://github.com/dask/distributed/issues/1179, which is not the reason why it fails on Travis.

<details>

Command lines for running the failed tests in brute force:

```
for i in {1..100}; do py.test "distributed/tests/test_client.py::test_open_close_many_workers[Worker-100-5]" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_scheduler.py::test_balance_many_workers" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_scheduler.py::test_balance_many_workers_2" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_scheduler.py::test_correct_bad_time_estimate" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_worksteal_many_thieves" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_dont_steal_unknown_functions" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_new_worker_steals" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_work_steal_no_kwargs" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_balance_without_dependencies" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_steal_twice" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_steal.py::test_accept_old_result_if_stolen" --pdb --runslow ; done
for i in {1..100}; do py.test "distributed/tests/test_stress.py::test_stress_1" --pdb --runslow ; done
```
</details>

mrocklin:
> How can garbage collection be dangerous?

Yeah I don't know.  I suspect some subtle interaction.

mrocklin:
Any further thoughts on this @bluenote10 ?

bluenote10:
I'm still puzzled why the tests fail on travis, but not locally. The logs on travis also don't look like actual test failures, rather errors related to testing infrastructure. I'm trying to get more output from travis now to see if this can help.

mrocklin:
There are a number of new failures in the tests, I suspect due to some change on travis-ci.  I pushed a trivial change here to help us get a better understanding of the kinds of tests that are failing on master: https://github.com/dask/distributed/pull/1317

mrocklin:
I've merged this into master.  We'll see if problems persist.

mrocklin:
Sorry, I've merged master into this.  Not the other way around.

mrocklin:
It looks like you've gotten tests to pass, but only by removing all of the actual functionality of the PR :/

Do you have any thoughts about how to trigger GC safely?  Or other thoughts about how to achieve the same results while only relying on the standard GC operation?

bluenote10:
I'm still wondering if the tests failing by explicit GC may hint to a potential cause of the memory leaks. Also, some failing tests were related to work stealing, and I think you already spotted an issue there. I will try to systematically re-enable GC here (after merging in the work stealing fix) and look at the pattern of failing tests again. However I'll have to work on something else for a while, so it might take a bit until I can pick this up again.

mrocklin:
OK, thank you for your continued engagement on this.  Sorry it took me a while to engage myself.

distributed/utils.py
```diff
@@ -898,6 +899,22 @@ def nbytes(frame):
...
+    def __init__(self, min_interval_in_sec=1.0):
+        self.min_interval_in_sec = min_interval_in_sec
+        self.last_collect = time()
+        #gc.collect()
+
+    def collect(self, force_gc=False):
+        new_time = time()
+        if force_gc or new_time - self.last_collect > self.min_interval_in_sec:
+            gc.collect()
+            self.last_collect = new_time
```
ogrisel: As @pitrou said in another PR, GC can legitimately take more than several hundred milliseconds (maybe even seconds) on interpreters loaded with lots of small python objects (e.g. in dask.bags with nested constructs). Therefore you might want to consider a strategy that records the time taken by the last call to `gc.collect()` and the time elapsed since after the end to last call to `gc.collect()` so as to only call `gc.collect()` if less than 5% of the time is spent inside the `gc.collect()` calls in aggregate instead of using a fixed absolute value for `min_interval_in_sec`.

Also using a [time.monotonic()](https://docs.python.org/3/library/time.html#time.monotonic) instead of `time.time()` under Python 3.5 and later will probably spare some rare but potentially hard to debug issues in case of leap second events.

bluenote10: Yes, good point, I think I'll just use your implementation, no need for two throttled GC implementations ;). I'll still have to check if this PR is needed at all with you modification.

mrocklin:

PR #1259, opened 2017-07-14
Fix slow marker when conftest is missing.

QuLogic:
The old fallback does not return anything, breaking anything that might be using the return value, like a marked parametrization.

An example is `test_dynamic_workloads_sync` in `distributed/tests/test_client.py`.

### DIFF
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -423,8 +423,7 @@ try:
                 not pytest.config.getoption("--runslow"),
                 reason="need --runslow option to run")
 except (AttributeError, ValueError):
-    def slow(*args):
-        pass
+    slow = pytest.mark.slow
 
 
 from tornado import gen


QuLogic:
BTW, should conftest be in the PyPI tarball?

distributed/utils_test.py
```diff
@@ -423,8 +423,7 @@ def disconnect_all(addresses, timeout=3):
                 not pytest.config.getoption("--runslow"),
                 reason="need --runslow option to run")
 except (AttributeError, ValueError):
-    def slow(*args):
-        pass
+    slow = pytest.mark.slow
```
TomAugspurger: This could cause issues with downstream projects who import distributed and run tests with the `--strict` option. See https://github.com/pandas-dev/pandas/issues/16680. The solution was to register the marker in our `setup.cfg` like https://github.com/pandas-dev/pandas/pull/16797/files

QuLogic: How about this gets moved to somewhere in `tests`? Right now this is broken when `conftest.py` is missing (like in the PyPI tarballs.)

mrocklin:
Would this all be resolved if we distributed the conftest.py file?

TomAugspurger:
That's what pandas ended up doing. We had to sense we have `pd.test()` to run the test suite.

QuLogic:
But does Pandas add an extra option? In Matplotlib, we couldn't add `conftest.py` in both packages and the top-level (i.e., distributed, but not part of a package), so we didn't bother with adding any options. This might not affect `distributed` though since it doesn't need both.

mrocklin:

Issue #1262, opened 2017-07-17
Handle missing extensions gracefully


jlowin:
There are a couple places ([1](https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L1311), [2](https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L3206)) where the scheduler tries to access `self.extensions['stealing']`. This will raise a KeyError if the `Stealing` extension has been disabled. We should handle these cases more gracefully.

(This is a tracking issue for my original [comment](https://github.com/dask/distributed/commit/afe66d5c7223ef2bf5501262c84a8b2ca1f11cb6#commitcomment-23135608))

PR #1263, opened 2017-07-17
Handle missing extensions

jlowin:
Addresses issue #1262 -- guards against direct access of `extensions['stealing']` in the event that the `Stealing` extension was disabled.

Adds a test for the access in `handle_long_running`, but I didn't find a way to trap the error caused by the access in `_reevaluate_occupancy`.

### DIFF
--- a/distributed/tests/test_worker_client.py
+++ b/distributed/tests/test_worker_client.py
@@ -217,3 +217,31 @@ def test_local_client_warning(c, s, a, b):
     future = c.submit(func, 10)
     result = yield future
     assert result == 11
+
+
+def test_secede_without_stealing_issue_1262():
+    """
+    Tests that seceding works with the Stealing extension disabled
+    https://github.com/dask/distributed/issues/1262
+    """
+
+    # turn off all extensions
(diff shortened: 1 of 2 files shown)


distributed/tests/test_worker_client.py
```diff
@@ -217,3 +217,31 @@ def func(x):
...
+    extensions = []
+
+    # run the loop as an inner function so all workers are closed
+    # and exceptions can be examined
+    @gen_cluster(client=True, scheduler_kwargs={'extensions': extensions})
+    def secede_test(c, s, a, b):
+        def func(x):
+            with worker_client() as wc:
+                y = wc.submit(lambda: 1 + x)
+                return wc.gather(y)
```
mrocklin: Looks like this isn't Python 2 compatible.  You'll have to use `raise gen.Return(...)`

jlowin: Whoops -- I was a little overoptimistic in my Python 3-ing.

mrocklin:
Thanks @jlowin !  This is in.

PR #1265, opened 2017-07-17
Change become_distributed to become_dask

jakirkham:
It use to be the case that one should call `become_distributed` to get a Distributed cluster to start on top of ipyparallel. However that method is deprecated. So the correct way to do this with ipyparallel now is to run `become_dask`.

cc @minrk

ref: http://ipyparallel.readthedocs.io/en/6.0.1/changelog.html#id23


mrocklin:
Looks good.  Thanks as always @jakirkham !

PR #1269, opened 2017-07-18
Futures_of only dives within objects that have a .dask mapping

mrocklin:


martindurant:
I can't see any downside to this extra check.

Issue #899, opened 2017-02-23
Feature request: Worker port ranges and/with nprocs support


stuartarchibald:
I think it would be quite useful for `dask-worker` to support `--nprocs` with `--foo-port=[range]` so that systems with specific port ranges opened can have a number of processes opened from one call applied to a specified range opposed to having to write a shell loop to do the same.

Further, general support for port ranges and failing over to the next available port in a range if the current is taken would probably be quite useful in a similar situation as that described above.

Thanks.

mrocklin:
Does anyone with some networking background know of a good way to ask the OS to connect to a port within a certain range?

cc @pitrou

pitrou:
@stuartarchibald, I assume you mean `--nprocs=N --start-port=X` will bind to ports X, X+1, X+2, etc? Are you expecting a specific pattern (e.g. first all worker ports, then all Bokeh ports), or is it indifferent?

mrocklin:
I suspect he's looking for something like `--port=5000:6000`.  Often people are told that they need to stay within a particular range of ports.

stuartarchibald:
I think being able to express a port range for each component would be ideal thanks. Something like:
```
dask-worker --nprocs=16 --worker-port=5000:5100 --http-port=6000:6100
```
which would start 16 worker processes bound to free ports between 5000 and 5100 and the same with `http` but `s/5/6/g`. This is with view of use cases where open ports are restricted to some particular range but some external reason requires computation to be undertaken in separate processes.

mrocklin:
I see two options:

1.  Find a way to ask the operating system for an unused port within a range
2.  Assume that the system giving us the port range has already made them free for us, and just choose an increasing sequence (erring loudly if this fails)

stuartarchibald:
I think either of these would be fine, the latter, with consistent and accessible complaining (for easy parsing), is probably easiest place to start?

pitrou:
If we change the syntax only slightly, we can fit this feature in the URL syntax, for example `tcp://0.0.0.0:6000-6099` would listen somewhere in the [6000, 6099] port range. What do you think?

pelson:
The addition of a port range would be awesome. I'm in the situation where I have a range of ports opened on my firewall, and I just want to use any available port in the given range.

[...]

PR #1278, opened 2017-07-21
1258: Added option to specify different listen and contact IPs for workers

### LINKED ISSUE #1258: Differentiate between binding and advertising address for dask workers
Currently the bind and the advertise ip of dask-workers is the same. In case the dask-worker is running inside a docker container in bridge mode these addresses will not be the same. While the worker binds to some internal address the scheduler needs to communicate to it using some external IP.

Currently we only have the option of settings the host address or the interface on which the Dask-worker listens on but nothing about which address is advertised to the scheduler.

sohaibiftikhar:
This issue fixes #1258 

Currently the bind and the advertise ip of dask-workers is the same. In case the dask-worker is running inside a docker container in bridge mode these addresses will not be the same. While the worker binds to some internal address the scheduler needs to communicate to it using some external IP. Currently we only have the option of settings the host address or the interface on which the Dask-worker listens on but nothing about which address is advertised to the scheduler.

After the fix we now have an option `--contact-addr` that helps achieve this. This needs to specified as `<address>:<port>`. In case this is not specified it defaults to the older behaviour. Run  `dask-worker --help` for more details on the new options.

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -49,14 +49,12 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 @click.option('--bokeh/--no-bokeh', 'bokeh', default=True, show_default=True,
               required=False, help="Launch Bokeh Web UI")
 @click.option('--listen-address', type=str, default=None,
-        help="The <protocol>://<address>:<port> on which the worker binds to. "
-                   "The options --worker-port and --host are not compatible "
-                   "with this option. Cannot be used with --nprocs>1")
+        help="The address to which the worker binds. "
+             "Example: tcp://0.0.0.0:9000")
 @click.option('--contact-address', type=str, default=None,
-        help="The <protocol>://<address>:<port> the worker advertises to the scheduler "
-                   "for communication with it and other workers. If specified "
-                   "must also specify --listen-address. "
-                   "If not specified uses the --listen-address")
+        help="The address the worker advertises to the scheduler for "
+             "communication with it and other workers. "
+             "Example: tcp://127.0.0.1:9000")
 @click.option('--host', type=str, default=None,
               help="Serving host. Should be an ip address that is"
                    " visible to the scheduler and other workers. "
@@ -96,7 +94,6 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 @click.option('--preload', type=str, multiple=True,
               help='Module that should be loaded by each worker process '
                    'like "foo.bar" or "/path/to/foo.py"')
-
 def main(scheduler, host, worker_port, listen_address, contact_address,
          http_port, nanny_port, nthreads, nprocs, nanny, name,
          memory_limit, pid_file, reconnect, resources, bokeh,
@@ -190,7 +187,7 @@ def main(scheduler, host, worker_port, listen_address, contact_address,
     loop = IOLoop.current()
 
     if nanny:
-        kwargs = {'worker_port': worker_port}
+        kwargs = {'worker_port': worker_port, 'listen_address': listen_address}
         t = Nanny
     else:
         kwargs = {}
@@ -230,7 +227,7 @@ def main(scheduler, host, worker_port, listen_address, contact_address,
                  services=services, name=name, loop=loop, resources=resources,
                  memory_limit=memory_limit, reconnect=reconnect,
                  local_dir=local_directory, death_timeout=death_timeout,
-                 preload=preload, security=sec, contact_addr = contact_address,
+                 preload=preload, security=sec, contact_address=contact_address,
                  **kwargs)
                for i in range(nprocs)]
 
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -19,6 +19,7 @@ from distributed.http import HTTPWorker
 from distributed.metrics import time
 from distributed.security import Security
 from distributed.cli.utils import check_python_3, uri_from_host_port
+from distributed.comm import get_address_host_port
 
 from toolz import valmap
 from tornado.ioloop import IOLoop, TimeoutError
@@ -47,9 +48,20 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
               help="Bokeh port, defaults to 8789")
 @click.option('--bokeh/--no-bokeh', 'bokeh', default=True, show_default=True,
               required=False, help="Launch Bokeh Web UI")
+@click.option('--listen-address', type=str, default=None,
+        help="The <protocol>://<address>:<port> on which the worker binds to. "
+                   "The options --worker-port and --host are not compatible "
+                   "with this option. Cannot be used with --nprocs>1")
+@click.option('--contact-address', type=str, default=None,
+        help="The <protocol>://<address>:<port> the worker advertises to the scheduler "
+                   "for communication with it and other workers. If specified "
+                   "must also specify --listen-address. "
+                   "If not specified uses the --listen-address")
 @click.option('--host', type=str, default=None,
               help="Serving host. Should be an ip address that is"
                    " visible to the scheduler and other workers. "
+                   "See --listen-address and --contact-address if you "
+                   "need different listen and contact addresses. "
                    "See --interface.")
 @click.option('--interface', type=str, default=None,
               help="Network interface like 'eth0' or 'ib0'")
@@ -84,20 +96,17 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 @click.option('--preload', type=str, multiple=True,
               help='Module that should be loaded by each worker process '
                    'like "foo.bar" or "/path/to/foo.py"')
-def main(scheduler, host, worker_port, http_port, nanny_port, nthreads, nprocs,
-         nanny, name, memory_limit, pid_file, reconnect,
-         resources, bokeh, bokeh_port, local_directory, scheduler_file,
-         interface, death_timeout, preload, bokeh_prefix,
-         tls_ca_file, tls_cert, tls_key):
+
+def main(scheduler, host, worker_port, listen_address, contact_address,
+         http_port, nanny_port, nthreads, nprocs, nanny, name,
+         memory_limit, pid_file, reconnect, resources, bokeh,
+         bokeh_port, local_directory, scheduler_file, interface,
+         death_timeout, preload, bokeh_prefix, tls_ca_file,
+         tls_cert, tls_key):
     sec = Security(tls_ca_file=tls_ca_file,
                    tls_worker_cert=tls_cert,
                    tls_worker_key=tls_key,
                    )
-
-    if nanny:
-        port = nanny_port
-    else:
-        port = worker_port
 
     if nprocs > 1 and worker_port != 0:
         logger.error("Failed to launch worker.  You cannot use the --port argument when nprocs > 1.")
@@ -110,6 +119,40 @@ def main(scheduler, host, worker_port, http_port, nanny_port, nthreads, nprocs,
     if nprocs > 1 and not nanny:
         logger.error("Failed to launch worker.  You cannot use the --no-nanny argument when nprocs > 1.")
         exit(1)
+
+    if contact_address and not listen_address:
+        logger.error("Failed to launch worker. "
+                     "Must specify --listen-address when --contact-address is given")
+        exit(1)
+
+    if nprocs > 1 and listen_address:
+        logger.error("Failed to launch worker. "
+                     "You cannot specify --listen-address when nprocs > 1.")
+        exit(1)
+
+    if  (worker_port or host) and listen_address:
+        logger.error("Failed to launch worker. "
+                     "You cannot specify --listen-address when --worker-port or --host is given.")
+        exit(1)
+
+    try:
+        if listen_address:
+            (host, worker_port) = get_address_host_port(listen_address, strict=True)
+
+        if contact_address:
+            # we only need this to verify it is getting parsed
+            (_, _) = get_address_host_port(contact_address, strict=True)
+        else:
+            # if contact address is not present we use the listen_address for contact
+            contact_address = listen_address
+    except ValueError as e:
+        logger.error("Failed to launch worker. " + str(e))
+        exit(1)
+
+    if nanny:
+        port = nanny_port
+    else:
+        port = worker_port
 
     if not nthreads:
         nthreads = _ncores // nprocs
@@ -187,7 +230,7 @@ def main(scheduler, host, worker_port, http_port, nanny_port, nthreads, nprocs,
                  services=services, name=name, loop=loop, resources=resources,
                  memory_limit=memory_limit, reconnect=reconnect,
                  local_dir=local_directory, death_timeout=death_timeout,
-                 preload=preload, security=sec,
+                 preload=preload, security=sec, contact_addr = contact_address,
                  **kwargs)
                for i in range(nprocs)]
 
--- a/distributed/comm/addressing.py
+++ b/distributed/comm/addressing.py
@@ -22,8 +22,10 @@ def parse_address(addr, strict=False):
         raise TypeError("expected str, got %r" % addr.__class__.__name__)
     scheme, sep, loc = addr.rpartition('://')
     if strict and not sep:
-        raise ValueError("Invalid url scheme. Must be of the form <protocol>://<host>:<port>. "
-                        "Got %s" % addr)
+        msg = ("Invalid url scheme. "
+               "Must include protocol like tcp://localhost:8000. "
+               "Got %s" % addr)
+        raise ValueError(msg)
     if not sep:
         scheme = DEFAULT_SCHEME
     return scheme, loc
--- a/distributed/comm/addressing.py
+++ b/distributed/comm/addressing.py
@@ -9,16 +9,21 @@ from . import registry
 DEFAULT_SCHEME = config.get('default-scheme', 'tcp')
 
 
-def parse_address(addr):
+def parse_address(addr, strict=False):
     """
     Split address into its scheme and scheme-dependent location string.
 
     >>> parse_address('tcp://127.0.0.1')
     ('tcp', '127.0.0.1')
+
+    If strict is set to true the address must have a scheme.
     """
     if not isinstance(addr, six.string_types):
         raise TypeError("expected str, got %r" % addr.__class__.__name__)
     scheme, sep, loc = addr.rpartition('://')
+    if strict and not sep:
+        raise ValueError("Invalid url scheme. Must be of the form <protocol>://<host>:<port>. "
+                        "Got %s" % addr)
     if not sep:
         scheme = DEFAULT_SCHEME
     return scheme, loc
@@ -96,17 +101,17 @@ def unparse_host_port(host, port=None):
         return host
 
 
-def get_address_host_port(addr):
+def get_address_host_port(addr, strict=False):
     """
     Get a (host, port) tuple out of the given address.
-
+    For definition of strict check parse_address
     ValueError is raised if the address scheme doesn't allow extracting
     the requested information.
 
     >>> get_address_host_port('tcp://1.2.3.4:80')
     ('1.2.3.4', 80)
     """
-    scheme, loc = parse_address(addr)
+    scheme, loc = parse_address(addr, strict=strict)
     backend = registry.get_backend(scheme)
     try:
         return backend.get_address_host_port(loc)
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -11,7 +11,7 @@ from tornado import gen
 from tornado.ioloop import IOLoop, TimeoutError
 from tornado.locks import Event
 
-from .comm import get_address_host, get_local_address_for
+from .comm import get_address_host, get_local_address_for, unparse_host_port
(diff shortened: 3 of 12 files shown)


mrocklin:
It looks like this is failing tests.  I suspect that some addresses are not switched over.

More generally, the help-text of the `--host` keyword seems to imply that it is the address that should be advertised out to network peers.  I suspect that it was playing both roles until now.  Separating these roles into two keywords might require some thought (and changes to existing help strings).

For example, `worker.address` currently points to the value given by `--host`.  Should it point to the address instead?  If we're asking the user to provide both `--host` and `--advertise-address` then what are valid combinations here?  Are there better terms that we should adopt like binding address and advertising address?  General thoughts are welcome here.  I suspect that this might involve a bit of conversation before a clean solution is found.

cc @pitrou (in case he's around)

bansalr:
To start the conversation:

1. --host and the pair (--advertise-address, --bind-address ) should be mutually exclusive. 
2. If you supply --advertise you have to supply --bind and vice versa.
3. If you supply --host you cannot supply either of --advertise or --bind
4. In case only --host is supplied, internally you can set both --bind and --advertise to --host.

rbubley:
While I don't disagree with the need for the discussions @mrocklin and @bansalr have started, the bug that is causing the failing tests seem to be coming from a different area: it's just that the advertising address isn't being parsed properly.

Replacing (in line 214 of worker.py):
```
            self.advertise_addr = self.advertise_addr + ":" + str(get_address_host_port(self.address)[1])
```

with
```
            self.advertise_addr = unparse_host_port(*parse_host_port(self.advertise_addr, default_port= str(get_address_host_port(self.address)[1])))

```

and changing the relevant  `import` line to:  
```
from .comm import get_address_host, get_address_host_port, get_local_address_for, parse_host_port, unparse_host_port
```
should fix (at least one?) of the test failures.

@sohaibiftikhar, perhaps you could make these changes to this PR?

mrocklin:
It looks like `distributed/comms` refers to these concepts as `listen_address` and `contact_address`.  It might be wise for us to be internally consistent and use those names if they are appropriate.

sohaibiftikhar:
@mrocklin Thanks for the discussion. I already knew this was not perfect and hence put it under WIP. Was not able to spend time with it lately due to other commitments but will look at it during the weekend. What do you have to say about @bansalr 's suggestion?
@rbubley Basically line 214 is a hack. I knew this while writing it. I need to get more familiar with the codebase to do it in a more elegant fashion.

rbubley:
Thanks @sohaibiftikhar. If you don't object, would you mind making the two changes to the PR I suggested, (as I don't think I can conveniently to it on GitHub) – as at least we can then see if it passes all the tests. We can make it more elegant after that...

bansalr:
Thank you @sohaibiftikhar. Much appreciated.

The above proposal updated for @mrocklin's suggestion to use --listen and --contact address.

1. --host and the pair (--listen-address, --contact-address ) should be mutually exclusive.
2. If you supply --listen-address you have to supply --contact-address and vice versa.
3. If you supply --host you cannot supply either of --listen-address or --contact-address
4. If --host is supplied, internally set both --contact-address and --listen-address to --host.
5. Finally, change internal implementation to only use --contact-address and --listen-address

@mrocklin - is this a good solution?

pitrou:
I would like to tweak the proposal slightly:

1. `--host` becomes a legacy option that is incompatible with the others (perhaps issue a deprecation warning?)
2. `--listen-address` can be specified alone; the contact address is then derived as it is today (IIRC, by using `getsockname`)
3. both `--listen-address` and `--contact-address` must be fully qualified URIs, i.e. `127.0.0.1:1234` is rejected (but `tcp://`, `tcp://127.0.0.1:1234`, etc. are all accepted)

pitrou:
As a more general remark, I was not aware that NAT was in use in datacenter-style setups, such as seems to happen with Docker. If that's not entirely exotical, perhaps a deployment section in the docs would be useful as well.

mrocklin:
For those who aren't aware @pitrou built most of the comms infrastructure in Dask and knows far more than I do here.  I recommend following his recommendations.

mrocklin:
@pitrou thanks for chiming in

bansalr:
@pitrou Its just not docker but also EC2 and GCE. Dask on ec2 can only work if the scheduler and worker are in the same ec2 VPC. VPC can only in one AWS region. If the scheduler is outside the VPC, e.g. in another AWS region or on an non ec2 machine, the workers will not be able to connect to the scheduler.

See for example [#1281](https://github.com/dask/distributed/issues/1281).

sohaibiftikhar:
@pitrou The problem is I think you cannot specify ports with the listen/contact address as in case of multiple `--nprocs` this gets a bit involved (currently you cannot specify worker port as well when `nprocs` is greater than one. Correct me if I am wrong.

pitrou:
@sohaibiftikhar I think you are right. But I'm not sure how you intend to configure port forwarding if you don't know the exact ports up front... (unless Docker magically forwards everything unchanged)

sohaibiftikhar:
@pitrou No docker cannot do this magically unless you run on `--host`. I will use a much simpler setup. If I need multiple processes I will launch them on the same machine manually rather than doing it with the `--nprocs` since behind the scenes it does the same. Then I have explicit control over which ports I am using.

rbubley:
Defining ports in the nprocs > 1 case is considered in Issue #899.

sohaibiftikhar:
@rbubley I think the issue is slightly different. While it does deal with controlling the ports the worker uses but for forwarding you still don't want a range of IP addresses to be forwarded.

bansalr:
@sohaibiftikhar Do you need a hand on this? We have some time this week and cannot use Dask on ec2 so high priority for us. Happy to pick this up where you leave off.

mrocklin:
It seems reasonable to me to disallow the `--nprocs` keyword in this situation.  I would expect mature deployments to specify workers individually anyway.

sohaibiftikhar:
@bansalr I'm sorry I have been slow in taking this on. I hope to make a definitive commit by tomorrow. Frankly, I don't know how I can divide this anyway. It's not a lot of changes just that my unfamiliarity with the codebase makes it a little slow.

bansalr:
@sohaibiftikhar no worries. you seem to have it under control. look forward to testing it out when you commit. thanks for working on this.

sohaibiftikhar:
@pitrou Could you throw a little light on why we need fully qualified URIs of the format like `<protocol>://<host>:<port>` and not just `<host>:<port>`?

sohaibiftikhar:
Is this latest failure random? My own Travis build seems to have succeeded: https://travis-ci.org/sohaibiftikhar/distributed/builds/259371095.

pitrou:
@sohaibiftikhar because distributed doesn't support only TCP but other protocols as well (for example TLS). The `host:port` form is a remnant of earlier versions which only knew about TCP, but I would like that habit to disappear.

sohaibiftikhar:
@pitrou Currently it should support both ways of specifying. Would you like it to be strict?
This would require some parsing code which I had largely till now avoided because of its subtlety.

bansalr:
@sohaibiftikhar, thanks!

sohaibiftikhar:
@mrocklin Could you please review the changes/suggest improvements (if any)? What can I do to move this forward?

distributed/cli/dask_worker.py
```diff
@@ -47,9 +48,21 @@
...
               required=False, help="Launch Bokeh Web UI")
+@click.option('--listen-addr', type=str, default=None,
+        help="The <address>:<port> on which the worker binds to. "
+                   "The options --worker-port and --host are not compatible "
+                   "with this option. Cannot be used with --nprocs>1")
+@click.option('--contact-addr', type=str, default=None,
+        help="The <address>:<port> the worker advertises to the scheduler "
+                   "for communication with it and other workers. For eg., "
+                   "in the case of using workers behind NAT. If specified "
+                   "must also specify listen-address. "
```
mrocklin: Should these helpstrings include protocol, like tcp or tls (or other)?  

Should we expand the name to `--listen-address` rather than `--listen-addr`?

I would remove the reference to the example case of a NAT in the helpstring.  This seems to be specific to a single case and not necessarily representative.

distributed/cli/dask_worker.py
```diff
@@ -107,6 +122,32 @@ def main(scheduler, host, worker_port, http_port, nanny_port, nthreads, nprocs,
         logger.error("Failed to launch worker.  You cannot use the --name argument when nprocs > 1.")
         exit(1)
 
+    if contact_addr and not listen_addr:
+        logger.error("Failed to launch worker. "
+                     "Must specify --listen-addr when --contact-addr is given")
+        exit(1)
```
mrocklin: This check should maybe be `(not contact_attr) != (not listen_addr)` for symmetry

pitrou:
> Currently it should support both ways of specifying. Would you like it to be strict?

Yes, I'd rather have it strict.

> This would require some parsing code which I had largely till now avoided because of its subtlety.

The parsing code should largely be there. If some additional functionality is needed, you can add it to `distributed.comm`.

mrocklin:
Any thoughts on how to test this?  Maybe we can use the fact that all of `127.0.0.*` points to localhost, so we could presumably specify different addresses like `--contact-address 127.0.0.2 --listen-address 127.0.0.3` and check that everyone gets the correct information.  The tests in `distributed/cli/dask_scheduler.py` are probably worth looking at.  You might also find the `client.scheduler_info()` function useful to get information about the cluster.

bansalr:
We are in the process of testing it for the ec2 use case where the
scheduler is running outside ec2. We hope to have results by tomorrow.

pitrou:
On the topic of docstrings, URIs needn't be of the form `scheme://host:port`, i.e. there are protocols which don't have a notion of host or port (currently only `inproc`, see https://github.com/dask/distributed/blob/master/distributed/comm/inproc.py#L55).

distributed/comm/addressing.py
```diff
@@ -9,16 +9,21 @@
...
     ('tcp', '127.0.0.1')
+
+    If strict is set to true the address must have a scheme.
     """
     if not isinstance(addr, six.string_types):
         raise TypeError("expected str, got %r" % addr.__class__.__name__)
     scheme, sep, loc = addr.rpartition('://')
+    if strict and not sep:
+        raise ValueError("Invalid url scheme. Must be of the form <protocol>://<host>:<port>. "
+                        "Got %s" % addr)
```
mrocklin: Not all protocols have this structure.  Consider for example the `inproc://` protocol.

sohaibiftikhar: > Any thoughts on how to test this

I had an idea about specifying a `listen-address` and an `advertise-address` and then checking if the process is actually connecting to the listen-address and the scheduler is actually connecting on the advertised-address. I think if I can somehow find out the process-id of the worker process this should not be hard. I am not sure about the scheduler part. Manually I just verify through the logs but not sure how to do it in this case.

> 1. For workers would people want to specify an address but not a port, asking for that to be randomly assigned by the OS? If so then does the current system accomplish this?

I believe currently, this is explicit. You need to specify the ports. I will post an update after checking.
> 2. Do we also want these keywords for the scheduler?

I don't think so. You can bind it to 0.0.0.0 and do the necessary redirection because the scheduler does not advertise its IP by itself.
> 3. Not all protocols have this structure. Consider for example the inproc:// protocol

I understand this. But the way the `cli/dask_worker.py` is written it expects a port? I can make the change but I am not clear about how to test this (even manually). I mean start the service on `inproc` address (is this supported?) but the scheduler and worker process are anyway on different machines and even if not there would be more workers that may need to communicate with this worker that could be on a different machine?

mrocklin: Regarding testing, please take a look at the tests in `distributed/cli/tests/test_dask_worker.py`.  You may want to use `client.scheduler_info()` to get information about the scheduler and workers.  You can also check for various attributes using [client.run](http://distributed.readthedocs.io/en/latest/api.html#distributed.Client.run)

> I believe currently, this is explicit. You need to specify the ports. I will post an update after checking.

So there aren't situations where workers would want to identify themselves by different hostnames but wouldn't know their ports ahead of time?  

> I understand this. But the way the cli/dask_worker.py is written it expects a port? 

This was true historically though I think that @pitrou has been trying to move away from this and keep addresses general.  For example we've been asked to look into supporting communication through other protocols like MPI.  This hasn't happened yet of course, so it's not something that you should feel obliged to optimize for, but it would be best to make sure that we don't exclude in the future.

bansalr:
We have been testing this for the last few days and this now allows workers running in ec2 to connect to a scheduler running outside ec2. 

Each worker has to be individually created (no --nprocs 1 and --nthreads 1) and the port specified for each worker. 

Great work @sohaibiftikhar, thank you!

distributed/comm/addressing.py
pitrou: @sohaibiftikhar, the problem here is only the error message you adopted for the ValueError.

sohaibiftikhar:
@bansalr I don't think you need the threads option?

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

mrocklin:
I've pushed up a test.  It fails showing that we currently ignore everything but the port in the `--listen-address` keyword.

mrocklin:
Or, perhaps it's just that I'm confused about how `--listen-address` should operate

mrocklin:
How should we handle the nanny port in this context?  Does this require a second contact/listen address pair?

mrocklin:
Beware that I've force-pushed to this branch to resolve conflicts with master.  @sohaibiftikhar I hope that you don't mind the intrusion into your fork.

sohaibiftikhar:
@pitrou No, it's good you added the tests. I will start looking into this again from Thursday if it is not wrapped up by then. Are there still any open concerns?

mrocklin:
It would be good to verify that these changes haven't affected your use in practice.  It would also be good to think about all the possible cases where this might be desired and ensure that we cover those cases.

mrocklin:
@pitrou this could use your review if you are available

distributed/worker.py
```diff
@@ -266,32 +270,48 @@ def _start(self, addr_or_port=0):
...
             self.ip = get_address_host(self.address)
+            try:
+                listen_host = get_address_host(addr_or_port)
+            except ValueError:
+                listen_host = addr_or_port
+
+        if '://' in listen_host:
+            protocol, listen_host = listen_host.split('://')
+            if protocol == 'inproc':
+                listen_host = 'localhost'
```
mrocklin: It's unfortunate that I'm hard-coding a particular protocol here.  Any suggestions on how to do this better?

pitrou: Can you explain why this is required?
Also, at this point, is `self.ip` still used for something?

distributed/worker.py
```diff
@@ -266,32 +270,48 @@ def _start(self, addr_or_port=0):
...
         self.name = self.name or self.address
         preload_modules(self.preload, parameter=self, file_dir=self.local_dir)
         # Services listen on all addresses
         # Note Nanny is not a "real" service, just some metadata
         # passed in service_ports...
-        self.start_services()
+        self.start_services(listen_host)
+
+        try:
+            listening_address = '%s%s:%d' % (self.listener.prefix, listen_host, self.port)
```
pitrou: Uh, why is this useful?

distributed/scheduler.py
```diff
@@ -410,16 +410,19 @@ def get_versions(self, comm):
...
-    def start_services(self, listen_ip=''):
+    def start_services(self, host):
         for k, v in self.service_specs.items():
             if isinstance(k, tuple):
                 k, port = k
             else:
                 port = 0
 
+            if host == '0.0.0.0':
+                host = ''  # for IPv6
```
pitrou: When/why is this useful?

distributed/scheduler.py
```diff
@@ -450,12 +453,17 @@ def start(self, addr_or_port=8786, start_queues=True):
                 # as it would prevent connecting via 127.0.0.1.
                 self.listen(('', addr_or_port), listen_args=self.listen_args)
                 self.ip = get_ip()
+                listen_host = ''
             else:
                 self.listen(addr_or_port, listen_args=self.listen_args)
                 self.ip = get_address_host(self.listen_address)
+                listen_host = self.ip
+
+            if self.ip == '0.0.0.0':
```
pitrou: Can `self.ip` *ever* be "0.0.0.0"? AFAIR it should be an actual IP address, not a wildcard address...

distributed/worker.py
mrocklin: We pass `listen_host` down to the services (http, bokeh).  We still use `self.ip` occasionally when we need to refer to this particular host.

The code above that creates `listen_host` is incorrect in the case of `inproc://` we sometimes get long addresses like `127.0.0.1/382`

distributed/worker.py
mrocklin: We log it 

```
mrocklin@carbon:~$ dask-worker localhost:8786 --contact-address tcp://127.0.0.1:9000 --listen-address tcp://:9000
distributed.nanny - INFO -         Start Nanny at: 'tcp://127.0.0.1:39718'
distributed.worker - INFO -       Start worker at:       tcp://127.0.0.1:9000
distributed.worker - INFO -          Listening to:                tcp://:9000
distributed.worker - INFO -              nanny at:                     :39718
distributed.worker - INFO -               http at:                     :33794
distributed.worker - INFO -              bokeh at:                      :8789
```

distributed/scheduler.py
mrocklin: This failed otherwise

```python
def test_defaults(loop):
    with popen(['dask-scheduler', '--no-bokeh']) as proc:

        @gen.coroutine
        def f():
            # Default behaviour is to listen on all addresses
            yield [
                assert_can_connect_from_everywhere_4_6(8786, 2.0),  # main port
                assert_can_connect_from_everywhere_4_6(9786, 2.0),  # HTTP port <<<<----------
                ]

        loop.run_sync(f)
```

distributed/worker.py
pitrou: So, this is the address passed by the user? Why do we have to recreate it?

distributed/scheduler.py
mrocklin: It can be.  You're right that we should address this rather than work around it.  I'll investigate.

distributed/scheduler.py
pitrou: I see. Where does "0.0.0.0" come from? Something must have changed which made that value appear...

distributed/worker.py
mrocklin: It can be created a couple different ways, including using `--host`.

mrocklin:
Stepping onto plane.  Going offline for a while.

distributed/nanny.py
```diff
@@ -161,6 +164,13 @@ def instantiate(self, comm=None):
 
         Blocks until the process is up and the scheduler is properly informed
         """
+        if self._listen_address:
+            start_arg = self._listen_address
+        else:
+            host = self.listener.bound_address[0]
```
pitrou: `bound_address` is not an attribute of the Listener ABC.

distributed/scheduler.py
pitrou: Ok, I understand. We used to call `start_services` without an argument, so `host` would always be empty. I agree this change ok, though I think `listen_ip` really was a better name than `host`.

distributed/worker.py
pitrou: Have you tried using `get_address_host` instead?

distributed/worker.py
pitrou: If you agree with breaking the Worker API, perhaps the Worker class may only accept fully-formed URIs and the CLI scaffolding be left to the dask-worker script. Then by construction the Worker class knows the listen URI.

Of course, a similar change could also be made for the Scheduler class.

sohaibiftikhar:
@mrocklin So I tested this out with a few different combinations (manually) and it worked fine. While you seem to have taken over, tell me if you need a hand.

distributed/worker.py
mrocklin: This works.  Fixed

distributed/scheduler.py
mrocklin: Changed host to listen_ip

distributed/scheduler.py
mrocklin: I've changed this check from `self.ip` over to `listen_host`

distributed/nanny.py
mrocklin: Hrm, is there a more consistent way to construct the listening address?  I suppose this problem goes away if we become more consistent about using contact and listening addresses.

distributed/worker.py
mrocklin: I think that breaking the API for the Worker and Nanny classes seems reasonable.  The people I know who use these APIs directly are usually pretty expert and capable of changing.  I'm inclined to put off this change for a while though just for personal time reasons.

mrocklin:
> While you seem to have taken over, tell me if you need a hand.

Right now what you're doing, review and testing in different and creative situations, is probably the most valuable.

mrocklin:
I plan to merge this in 24 hours if there are no further comments.

mrocklin:
This is in.  Thank you everyone for your efforts on this.

PR #1279, opened 2017-07-21
Improve `Client.get_versions`

jcrist:
This fixes two issues:

- `get_versions` now works even if `cloudpickle` versions mismatch. This was caused by applying a function in a closure, which required cloudpickle to serialize. We now just use `get_versions` directly.

- Fully describe all mismatched versions. Previously we only errored on the first found mismatch, which led to playing whack-a-mole with getting version compatibility. We now provide a nicely formatted error message describing all version incompatibilities with the client. Only workers/scheduler that mismatch the client are reported.

**Example:**

```
In [3]: client.get_versions(check=True)
<ipython-input-3-13767d2792b6> in <module>()
----> 1 client.get_versions(check=True)

~/Code/distributed/distributed/client.py in get_versions(self, check)
   2566                 raise ValueError("Mismatched versions found\n"
   2567                                  "\n"
-> 2568                                  "%s" % ('\n\n'.join(errs)))
   2569
   2570         return result

ValueError: Mismatched versions found

cloudpickle
+-------------------------+---------+
|                         | version |
+-------------------------+---------+
| client                  | 0.3.1   |
| tcp://10.10.20.47:52687 | 0.2.2   |
+-------------------------+---------+

distributed
+-----------+--------------------------+
|           | version                  |
+-----------+--------------------------+
| client    | 1.18.0+15.gafbd453       |
| scheduler | 1.18.0+14.gef90d9e.dirty |
+-----------+--------------------------+

pandas
+-------------------------+---------+
|                         | version |
+-------------------------+---------+
| client                  | 0.20.3  |
| tcp://10.10.20.47:52687 | 0.19.2  |
+-------------------------+---------+
```

Note that there currently aren't any tests, as I wasn't sure how to test this. Could break the error formatting code out into a function and test that, but not sure if it's worth it.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -49,7 +49,7 @@ from .security import Security
 from .sizeof import sizeof
 from .worker import dumps_task, thread_state, get_client, get_worker
 from .utils import (All, sync, funcname, ignoring, queue_to_iterator,
-        tokey, log_errors, str_graph, key_split, format_bytes)
+        tokey, log_errors, str_graph, key_split, format_bytes, asciitable)
 from .versions import get_versions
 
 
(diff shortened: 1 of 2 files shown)


mrocklin:
This looks great.  Merging

Issue #1294, opened 2017-07-26
TypeError: 'Future' object is not iterable when using dask.bag.Bag.accumulate with get=dask.get


ogrisel:
Here is a reproduction case (I tried to simplify it but could not get rid of the scikit-learn dependency even tough it seems unrelated):

```
pip install scipy scikit-learn 
```

```python
import numpy as np
import scipy.sparse as sp
from sklearn.linear_model import SGDClassifier
import dask
import dask.bag as db

n_features = int(1e3)
n_informative = int(1e2)

n_chunks = int(1e1)
chunk_size = int(1e2)

rng = np.random.RandomState(42)
true_coef = rng.randn(n_features)
true_coef[n_informative:] = 0


def make_chunk(n_samples, true_coef, chunk_idx, format='csr',
               density=1e-3, noise=1e-1):
    rng = np.random.RandomState(chunk_idx)
    n_features = true_coef.shape[0]
    input_data = sp.rand(n_samples, n_features, format=format,
                         density=density, random_state=rng)
    noise = rng.normal(loc=0, scale=noise, size=n_samples)
    target = input_data.dot(true_coef).ravel() + noise
    return chunk_idx, input_data, (target > 0).astype(np.int32)


def scan_fit(model, next_chunk):
    chunk_idx, X, y = next_chunk
    return model.partial_fit(X, y, classes=[0, 1])


if __name__ == "__main__":
    from distributed import Client
    client = Client()
    b = db.from_sequence([(chunk_size, true_coef, i)
                          for i in range(n_chunks)])
    b = b.starmap(make_chunk).persist()

    model = SGDClassifier(loss='log', penalty='l1')
    last_delayed = b.accumulate(scan_fit, initial=model).to_delayed()[-1]

    # Using the distributed scheduler works:
    print("distributed scheduler")
    print(last_delayed.compute()[0])

    # Trying to use the single-threaded dask scheduler on a pipeline of delayed
    # operations defined **after** the distributed client was initialized:
    print("sequential scheduler")
    print(last_delayed.compute(get=dask.get)[0])
```

The second call to `compute` (the one with `get=dask.get`) triggers the following exception:

```
Traceback (most recent call last):
  File "/home/ogrisel/tmp/dask_bug.py", line 51, in <module>
    print(last_delayed.compute(get=dask.get)[0])
  File "/home/ogrisel/code/dask/dask/base.py", line 98, in compute
    (result,) = compute(self, traverse=False, **kwargs)
[...]

PR #1296, opened 2017-07-26
Don't send closing messages on closed comm

mrocklin:
When client closes it sends out a bunch of final messages to the scheduler
cleaning up its various resources.  However if the scheduler connection is
already dead then this generates annoying error messages.  This was a common
occurrence when closing a LocalCluster.

Reported in https://github.com/dask/distributed/issues/1294

### DIFF
--- a/distributed/batched.py
+++ b/distributed/batched.py
@@ -7,7 +7,6 @@ from tornado import gen, locks
 from tornado.ioloop import IOLoop
 
 from .core import CommClosedError
-from .utils import ignoring
 
 
 logger = logging.getLogger(__name__)
--- a/distributed/batched.py
+++ b/distributed/batched.py
@@ -65,9 +65,11 @@ class BatchedSend(object):
     @gen.coroutine
     def _background_send(self):
         while not self.please_stop:
-            with ignoring(gen.TimeoutError):
+            try:
                 yield self.waker.wait(self.next_deadline)
                 self.waker.clear()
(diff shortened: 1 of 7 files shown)


mrocklin:
@ogrisel

ogrisel:
I confirm that fixes my issue. Cannot you find an easy way to add a non-regression test for this?

ogrisel:
The travis failures are real apparenty.

mrocklin:
Yeah, I've been tracking down a failure today that snuck in recently.

Regarding tests the clearest way to test this would probably be to look at logs.  I've had issues testing logging in the past.  Do you have any recommendations or examples where you've seen this done well?

mrocklin:
Ah, no.  These failures are genuine

ogrisel:
Besides registering a specific logger in a test fixture and grepping the byte buffer...

mrocklin:
> Besides registering a specific logger in a test fixture and grepping the byte buffer...

If you had a pointer to someplace that had done this well I would be happy to learn from such an example.  My attempts to do this in the past have either not worked or felt like a hack.

ogrisel:
I assume it's possible to do a test fixture that does:

```python
import logging
import io

CAPTURED_LOGS = io.BytesIO()
logging.basicConfig(level=logging.DEBUG, stream=CAPTURED_LOGS)
```

(I have not tested it).

ogrisel:
LGTM.

PR #1298, opened 2017-07-27
(WIP) tracing to JSON

martindurant:
Note the places where binary and sets had to be stripped out for JSON
serialization not to fail.

Something similar could be done for workers and clients, and the resultant
files would be relatively easy to merge together by timestamp (if they could be pulled to the same location - or maybe dask does this for us so long a the workers stay alive?).

Example output: https://gist.github.com/martindurant/f9be1fa145c6691b8661310830574bea

### DIFF
--- a/distributed/diagnostics/tracer.py
+++ b/distributed/diagnostics/tracer.py
@@ -1,0 +1,115 @@
+from __future__ import print_function, division, absolute_import
+
+import logging
+import json
+import time
+from tornado import gen
+
+from .plugin import SchedulerPlugin
+from ..utils import key_split
+
+
+logger = logging.getLogger(__name__)
+
+
+class TraceLoggerPlugin(SchedulerPlugin):
+    def __init__(self, scheduler):
+        self.log = None
+        logger.debug('Set up Scheduler trace logger')
+        self.scheduler = scheduler
+        self.scheduler.handlers['start_trace'] = self.start_trace
+        self.scheduler.handlers['stop_trace'] = self.stop_trace
+
+    def start_trace(self, *args, **kwargs):
+        log_file = kwargs.pop('log_file', 'trace.log')
+        logger.debug('Start trace to file: ' + log_file)
+        log = logging.getLogger('scheduler_trace')
+        json_handler = logging.handlers.RotatingFileHandler(filename=log_file)
+        json_handler.setFormatter(JSONFormatter())
+        log.addHandler(json_handler)
+        log.setLevel(logging.DEBUG)
+        self.log = log
+        self.log.debug({'log': 'start'})
+        self.scheduler.add_plugin(self)
+
+    def stop_trace(self, *args, **kwargs):
+        logger.debug('Stop trace')
+        self.log.debug({'log': 'stop'})
+        self.scheduler.extensions.pop('trace', None)
+        self.log = None
+
+    def transition(self, key, start, finish, *args, **kwargs):
+        if self.log:
+            kwargs.pop('type', None)
+            self.log.debug({'key': key, 'start': start, 'finish': finish,
+                           'args': args, 'kwargs': kwargs, 'log': 'transition'})
+
+    def update_graph(self, scheduler, dsk=None, keys=None,
+                     restrictions=None, **kwargs):
+        if self.log and dsk or keys:
+            kwargs2 = kwargs.copy()
+            kwargs2.pop('tasks', None)
+            keys = list(keys or [])
+            self.log.debug({'log': 'update_graph', 'dsk': dsk, 'keys': keys,
+                            'restrictions': restrictions, 'kwargs': kwargs2})
+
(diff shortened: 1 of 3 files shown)


distributed/diagnostics/tracer.py
```diff
@@ -0,0 +1,115 @@
...
+        logger.debug('Stop trace')
+        self.log.debug({'log': 'stop'})
+        self.scheduler.extensions.pop('trace', None)
+        self.log = None
+
+    def transition(self, key, start, finish, *args, **kwargs):
+        if self.log:
+            kwargs.pop('type', None)
+            self.log.debug({'key': key, 'start': start, 'finish': finish,
+                           'args': args, 'kwargs': kwargs, 'log': 'transition'})
```
mrocklin: There are many transitions here that you may not want to record.  I typically filter here for transitions that leave the task in a finished state (memory, error)

distributed/diagnostics/tracer.py
```diff
@@ -0,0 +1,115 @@
...
+        logger.debug('Set up Scheduler trace logger')
+        self.scheduler = scheduler
+        self.scheduler.handlers['start_trace'] = self.start_trace
+        self.scheduler.handlers['stop_trace'] = self.stop_trace
+
+    def start_trace(self, *args, **kwargs):
+        log_file = kwargs.pop('log_file', 'trace.log')
+        logger.debug('Start trace to file: ' + log_file)
+        log = logging.getLogger('scheduler_trace')
+        json_handler = logging.handlers.RotatingFileHandler(filename=log_file)
```
mrocklin: Is there a benefit to using the loging module rather than writing to a file directly?

distributed/diagnostics/tracer.py
```diff
@@ -0,0 +1,115 @@
...
+                           'args': args, 'kwargs': kwargs, 'log': 'transition'})
+
+    def update_graph(self, scheduler, dsk=None, keys=None,
+                     restrictions=None, **kwargs):
+        if self.log and dsk or keys:
+            kwargs2 = kwargs.copy()
+            kwargs2.pop('tasks', None)
+            keys = list(keys or [])
+            self.log.debug({'log': 'update_graph', 'dsk': dsk, 'keys': keys,
+                            'restrictions': restrictions, 'kwargs': kwargs2})
```
mrocklin: Pushing the entire dask graph might be problematic.  Especially when it contains gigabyte arrays.  I recommend switching to a synopsis here.

martindurant:
The question I'm asking by posting this WIP is: is this "trace" output useful, compared to normal debug-level logging? How much to write, whether to filter for particular keys, and whether to save (pieces of) whole graphs could all be options.
Saving whole graphs and several of the scheduler state variables would be similar to "checkpoints", and could be another optional output, done at set intervals.

mrocklin:
Serializing and saving the entire graph could take minutes of time in real-world situations.  This would almost certainly crash the cluster.

I recommend trying to build functions to produce particular outputs for particular sets of data that we already collect.  I can think of two at the moment:

1.  The contents of the task stream plot, which are currently recorded in a deque in the existing `TaskStreamPlugin`
2.  The contents of the `.events` collection

These are already recorded so you wouldn't need to muck with logging handlers, writing to json, etc..  You would just need to call a function on the scheduler that collects this data and hands it to the user.

However, I'm also somewhat hesitant to see this work done without an active driving user.  I certainly wouldn't trust myself to know what to produce without having an active use case to guide me.  I know that I personally would almost certainly make the wrong choices about what to collect and how.

martindurant:
The driver is the somewhat nebulous requests at scipy from people who want to know how a computation actually went down, or an an idea of how the whole system works. Of course, that's a very wide potential parameter space.

The most basic representation might be the animated graph that seems to be quite appealing for people that could be made quite easily from this data (but now with duration and bytes).

More generally, the more data that's saved, the more diagnostics could be done. Examples include: the scheduler started to try to push a task to worker X, but work didn't start on worker X until after a lag of 2s; what was X doing during this time? When I do a simple round-trip from a client, what parts of the system contribute tot he latency? (these for consolidated logs from all components).

I think I'll leave this PR here for a few days and think about options.

mrocklin:
I would start this process by opening an issue on github and asking people for their thoughts on concrete plots and analyses they would want to run.  Probably you would have to iterate with them to get to a few specific deliverables.  Then I would write a small implementation and hand it to them to try it out and give feedback.  Generally I've found that writing things in isolation of user feedback tends to produce work that is neat, but not especially critical to anyone.

mrocklin:
I would be much more comfortable if this work had a couple of users who were trying things out on their problems as development occurred.  Personally I would hold off on spending development hours until we found users willing to engage in this way.

I know that @shughes-uk has expressed an interest in this.  I wonder if this is still a driving need for her and if she has the time to engage here.

mrocklin:

PR #1301, opened 2017-08-01
fixing docstring of read_block function

evanfwelch:
I am in the middle of debugging an authetntication to S3 error and just found it

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -614,7 +614,7 @@ def read_block(f, offset, length, delimiter=None):
 
     Parameters
     ----------
-    fn: string
+    f: string
         Path to filename on S3
     offset: int
         Byte offset to start read
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -614,8 +614,8 @@ def read_block(f, offset, length, delimiter=None):
 
     Parameters
     ----------
-    f: string
-        Path to filename on S3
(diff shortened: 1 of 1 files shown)


martindurant:
You mean that `f` in the function is an open file, not a file-name to be opened?

distributed/utils.py
```diff
@@ -611,7 +611,7 @@ def read_block(f, offset, length, delimiter=None):
 
     Parameters
     ----------
-    fn: string
+    f: string
```
mrocklin: `f` is commonly used for a file while `fn` is commonly used for a filename.  If the intent is for this to be a filename then we might instead change f->fn everywhere.  If this is commonly used for non-file paths then we might instead use the term `path` everywhere.

@martindurant any preferences?

martindurant: In the code that follows, it is definitely a file (not a string), so `f`. I suspect that in a previous version, it was indeed a string, and this is vestigial.

evanfwelch:
Thanks for your attention here, I should have been more clear. Yeah f is
definitely a file in this function so just fixing the docstring to reflect
that. I like the f vs fn convention and have noticed it elsewhere.

mrocklin:
I've updated the docstring to point out that this is a file, and not a string.  Merging on passed tests.

PR #1302, opened 2017-08-01
Deprecate Executor

mrocklin:
This adds a warning.

It is also a bit unclean because Executor is now a function rather than
the class itself, so isinstance checks and the like will stop working
immediately.

https://stackoverflow.com/questions/45449038/what-is-the-difference-between-calling-dask-distrubuted-executor-and-dask-distri/45449190#45449190

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2777,10 +2777,11 @@ class Client(Node):
         return collections_to_dsk(collections, *args, **kwargs)
 
 
-def Executor(*args, **kwargs):
+class Executor(Client):
     """ Deprecated: see Client """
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -2773,7 +2774,10 @@ def collections_to_dsk(collections, *args, **kwargs):
         return collections_to_dsk(collections, *args, **kwargs)
 
 
-Executor = Client
+def Executor(*args, **kwargs):
```
TomAugspurger: Could you make `Executor` a subclass of `Client` that raises the warning in `__init__` like

```python
class Executor(Client):
    warnings.warn(...)
```

PR #1311, opened 2017-08-03
Additional flexibility for Adpative

TomAugspurger:
This starts to make `deploy.Adaptive` more reusable by subclasses like `dask_drmaa.Adaptive` by giving the subclass more places to override the default behavior. Currently the extra points are

- Whether to scale down (e.g. a subclass can choose not to change the cluster size if it's been less than `n` seconds since the last task finished)
- How to scale up (a subclass may have tasks with some resource constraints)
- Why we should scale up (currently are we cpu-bound? or are we memory-bound?).

I'll add additional tests and docs tomorrow, but everything should be backwards compatible.

### DIFF
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -20,6 +20,11 @@ class Adaptive(object):
     scheduler: distributed.Scheduler
     cluster: object
         Must have scale_up and scale_down methods/coroutines
+    startup_cost : int, default 1
+        Factor representing how costly it is to start an additional worker.
+        Affects quickly to adapt to high tasks per worker loads
+    scale_factor : int, default 2
+        Factor to scale by when it's determined additional workers are needed
 
     Examples
     --------
@@ -28,38 +33,123 @@ class Adaptive(object):
     ...         """ Bring worker count up to n """
     ...     def scale_down(self, workers):
     ...        """ Remove worker addresses from cluster """
+
+    Notes
+    -----
+    Subclasses can override :meth:`Adaptive.should_scale_up` and
+    :meth:`Adaptive.should_scale_down` to control when the cluster should be
+    resized. The default implementation checks if there are too many tasks
+    per worker or too little memory available (see :meth:`Adaptive.needs_cpu`
+    and :meth:`Adaptive.needs_memory`).
+
+    :meth:`Adaptive.get_scale_up_kwargs` method controls the arguments passed to
+    the cluster's ``scale_up`` method.
     '''
-    def __init__(self, scheduler, cluster, interval=1000, startup_cost=1):
+
+    def __init__(self, scheduler, cluster, interval=1000, startup_cost=1,
+                 scale_factor=2):
         self.scheduler = scheduler
         self.cluster = cluster
         self.startup_cost = startup_cost
+        self.scale_factor = scale_factor
         self._adapt_callback = PeriodicCallback(self._adapt, interval,
                                                 self.scheduler.loop)
         self.scheduler.loop.add_callback(self._adapt_callback.start)
         self._adapting = False
 
+    def needs_cpu(self):
+        """
+        Check if the cluster is CPU constrained (too many tasks per core)
+
+        Notes
+        -----
+        Returns ``True`` if the occupancy per core is some factor larger
+        than ``startup_cost``.
+        """
+        total_occupancy = self.scheduler.total_occupancy
+        total_cores = sum(self.scheduler.ncores.values())
+
+        if total_occupancy / (total_cores + 1e-9) > self.startup_cost * 2:
+            logger.info("CPU limit exceeded [%d occupancy / %d cores]",
+                        total_occupancy, total_cores)
+            return True
+        else:
+            return False
+
+    def needs_memory(self):
+        """
+        Check if the cluster is RAM constrained
+
+        Notes
+        -----
+        Returns ``True`` if  the required bytes in distributed memory is some
+        factor larger than the actual distributed memory available.
+        """
+        limit_bytes = {w: self.scheduler.worker_info[w]['memory_limit']
+                        for w in self.scheduler.worker_info}
+        worker_bytes = self.scheduler.worker_bytes
+
+        limit = sum(limit_bytes.values())
+        total = sum(worker_bytes.values())
+        if total > 0.6 * limit:
+            logger.info("Ram limit exceeded [%d/%d]", limit, total)
+            return True
+        else:
+            return False
+
     def should_scale_up(self):
+        """
+        Determine whether additional workers should be added to the cluster
+
+        Returns
+        -------
+        scale_up : bool
+
+        Notes
+        ----
+        Additional workers are added whenever
+
+        1. There are unrunnable tasks and no workers
+        2. The cluster is CPU constrained
+        3. The cluster is RAM constrained
+
+        See Also
+        --------
+        needs_cpu
+        needs_memory
+        """
         with log_errors():
             if self.scheduler.unrunnable and not self.scheduler.ncores:
                 return True
 
-            total_occupancy = sum(self.scheduler.occupancy.values())
-            total_cores = sum(self.scheduler.ncores.values())
+            needs_cpu = self.needs_cpu()
+            needs_memory = self.needs_memory()
 
-            if total_occupancy / (total_cores + 1e-9) > self.startup_cost * 2:
-                return True
-
-            limit_bytes = {w: self.scheduler.worker_info[w]['memory_limit']
-                            for w in self.scheduler.worker_info}
-            worker_bytes = self.scheduler.worker_bytes
-
-            limit = sum(limit_bytes.values())
-            total = sum(worker_bytes.values())
-
-            if total > 0.6 * limit:
+            if needs_cpu or needs_memory:
                 return True
 
             return False
+
+    def should_scale_down(self):
+        """
+        Determine whether any workers should potentially be removed from
+        the cluster.
+
+        Returns
+        -------
+        scale_down : bool
+
+        Notes
+        -----
+        ``Adaptive.should_scale_down`` always returns True, so we will always
+        attempt to remove workers as determined by
+        ``Scheduler.workers_to_close``.
+
+        See Also
+        --------
+        Scheduler.workers_to_close
+        """
+        return True
 
     @gen.coroutine
     def _retire_workers(self):
@@ -67,10 +157,30 @@ class Adaptive(object):
             workers = yield self.scheduler.retire_workers(remove=True,
                     close_workers=True)
 
-            logger.info("Retiring workers %s", workers)
-            f = self.cluster.scale_down(workers)
-            if gen.is_future(f):
-                yield f
+            if workers:
+                logger.info("Retiring workers %s", workers)
+                f = self.cluster.scale_down(workers)
+                if gen.is_future(f):
+                    yield f
+
+    def get_scale_up_kwargs(self):
+        """
+        Get the arguments to be passed to ``self.cluster.scale_up``.
+
+        Notes
+        -----
+        By default the desired number of total workers is returned (``n``).
+        Subclasses should ensure that the return dictionary includes a key-
+        value pair for ``n``, either by implementing it or by calling the
+        parent's ``get_scale_up_kwargs``.
+
+        See Also
+        --------
+        LocalCluster.scale_up
+        """
+        instances = max(1, len(self.scheduler.ncores) * self.scale_factor)
+        logger.info("Scaling up to %d workers", instances)
+        return {'n': instances}
 
     @gen.coroutine
     def _adapt(self):
@@ -80,13 +190,13 @@ class Adaptive(object):
         self._adapting = True
         try:
             if self.should_scale_up():
-                instances = max(1, len(self.scheduler.ncores) * 2)
-                logger.info("Scaling up to %d workers", instances)
-                f = self.cluster.scale_up(instances)
+                kwargs = self.get_scale_up_kwargs()
+                f = self.cluster.scale_up(**kwargs)
                 if gen.is_future(f):
                     yield f
 
-            yield self._retire_workers()
+            if self.should_scale_down():
+                yield self._retire_workers()
         finally:
             self._adapting = False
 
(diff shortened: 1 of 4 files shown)


distributed/deploy/adaptive.py
```diff
@@ -80,13 +132,13 @@ def _adapt(self):
         self._adapting = True
         try:
             if self.should_scale_up():
```
martindurant: Should probe should_scale_up/down before calling any adaption functions? I know that the actions are not immediate, but I could imagine scaling up here and immediately scaling down a few lines later.

TomAugspurger:
I wonder if the `should_scale_up` and `should_scale_down` methods should be coroutines? I could imagine the `Adaptive` querying the cluster manager for how busy it currently is, and deciding based on that whether or not to scale up / down.

mrocklin:
In most cluster resource mangers I know of you just ask for resources and the resource manager either gives you or doesn't give you resources.  You generally don't ask for its state.

Regardless, if would not be hard to make the `should_scale` methods coroutines in the future

distributed/deploy/adaptive.py
TomAugspurger: Could you say a bit more about your concern? Or if I can try to guess, the concern is that you get into a situation of competing `scale_up` and `scale_down` calls?

The `should_scale_down` method may not be the clearest name, it's more like `should_check_with_the_scheduler_if_there_are_any_workers_to_scale_down`. It's that "check with the scheduler" part that matters. The default is always `True`, so always check. And then the default behavior of the scheduler is to retire any workers that are idle.

The new behavior should be identical to the previous behavior. It's just adding a point where a subclass can overrride.

TomAugspurger:
@mrocklin can you clear up a question for me: when can / should the `.workers` attribute of a `cluster` and `scheduler` differ?

On master, if you run

```python
import logging
import time

from distributed.deploy import Adaptive


logger = logging.getLogger("distributed.deploy.adaptive")

cluster = LocalCluster(n_workers=0)
client = Client(cluster)
adaptive = Adaptive(cluster.scheduler, cluster)


def f(x):
    time.sleep(.1)
    return x


logger.setLevel(logging.INFO)
tasks = client.map(f, range(100))
```

And then check `adaptive.cluster, adaptive.scheduler`, you'll see they have different number of workers

```python
In [3]: adaptive.cluster, adaptive.scheduler
Out[3]:
(LocalCluster('tcp://127.0.0.1:51049', workers=2, ncores=16),
 <Scheduler: "tcp://127.0.0.1:51049" processes: 1 cores: 8>)
```

This causes us to repeatedly call `Adaptive.cluster.scale_up(2)`, since the scheduler only knows about 1 worker. But in the cluster has two workers, so the `scale_up` does nothing.

mrocklin:
There is no established pattern for what `cluster.workers` should include.  Feel free to invent one.  I would expect there to be some lag between when the cluster launches/asks for resources and when the scheduler sees them.

distributed/deploy/adaptive.py
```diff
@@ -20,6 +20,11 @@ class Adaptive(object):
     scheduler: distributed.Scheduler
     cluster: object
         Must have scale_up and scale_down methods/coroutines
+    startup_cost : int, default 1
+        Factor representing how costly it is to start an additional worker.
+        Affects quickly to adapt to high tasks per worker loads
+    scale_factor : int, default 2
```
TomAugspurger: Earlier I had this as a class attribute, but I moved it to be an argument to `__init__` to be more in line with `startup_cost` (it's hard coded to 2 on master)

distributed/deploy/adaptive.py
```diff
@@ -28,49 +33,154 @@ class Adaptive(object):
...
+    def needs_cpu(self):
+        """
+        Check if the cluster is CPU constrained (too many tasks per core)
+
+        Notes
+        -----
+        Returns ``True`` if the occupancy per core is some factor larger
+        than ``startup_cost``.
+        """
+        total_occupancy = sum(self.scheduler.occupancy.values())
```
TomAugspurger: There's now a `Scheduler.total_occupancy`, should I use that instead?

distributed/deploy/adaptive.py
```diff
@@ -28,49 +33,154 @@ class Adaptive(object):
...
         self.scheduler = scheduler
         self.cluster = cluster
         self.startup_cost = startup_cost
+        self.scale_factor = scale_factor
         self._adapt_callback = PeriodicCallback(self._adapt, interval,
                                                 self.scheduler.loop)
         self.scheduler.loop.add_callback(self._adapt_callback.start)
         self._adapting = False
 
+    def needs_cpu(self):
```
mrocklin: Thoughts on other needs?  For example `needs_resources`?  Should we think about generalizing this or is this premature?

distributed/deploy/adaptive.py
```diff
@@ -28,49 +33,154 @@ class Adaptive(object):
...
+                        for w in self.scheduler.worker_info}
+        worker_bytes = self.scheduler.worker_bytes
+
+        limit = sum(limit_bytes.values())
+        total = sum(worker_bytes.values())
+        if total > 0.6 * limit:
+            logger.info("Ram limit exceeded [%d/%d]", limit, total)
+            return True
+        else:
+            return False
```
mrocklin: Just thinking out loud here, but we might consider emitting the amount of RAM we need here rather than just "more please".  This might help systems to adapt more quickly.  

More generally this brings up the thought that maybe we should specify targets rather than ask for more/less.  This is beyond the scope of this PR though.

distributed/deploy/adaptive.py
```diff
@@ -28,49 +33,154 @@ class Adaptive(object):
...
+        -------
+        scale_up : bool
+
+        Notes
+        ----
+        Additional workers are added whenever
+
+        1. There are unrunnable tasks and no workers
+        2. The cluster is CPU constrained
+        3 The cluster is RAM constrained
```
mrocklin: `3.`

distributed/deploy/adaptive.py
```diff
@@ -28,49 +33,154 @@ class Adaptive(object):
...
+        -----
+        ``Adaptive.should_scale_down`` always returns True, so we will always
+        attempt to remove workers as determined by
+        ``Scheduler.workers_to_close``.
+
+        See Also
+        --------
+        Scheduler.workers_to_close
+        """
+        return True
```
mrocklin: This seems awkward

distributed/deploy/adaptive.py
TomAugspurger: > More generally this brings up the thought that maybe we should specify targets rather than ask for more/less. This is beyond the scope of this PR though.

Agreed that targets are probably better. I'm essentially doing that via the `get_scale_up_kwargs`: in the dask-drmaa PR on top of this: https://github.com/TomAugspurger/dask-drmaa/blob/8021d288fee20d5e464532505753c0fe74583ba8/dask_drmaa/adaptive.py#L68

distributed/deploy/adaptive.py
TomAugspurger: Agreed :) My usecase is allowing a subclass to define something more useful. For example, an Adaptive strategy that won't scale down too soon after the last task finished:

```python
import datetime
from distributed.diagnostics.plugin import SchedulerPlugin
from distributed.deploy.adaptive import Adaptive


class LastTask(SchedulerPlugin):
    "Record the time of the last task"
    def __init__(self):
        self.last_task = datetime.datetime.now()

    def transition(self, key, start, finish):
        if start == 'waiting' and finish == 'processing':
            self.last_task = datetime.datetime.now()


plugin = LastTask()
cluster = LocalCluster(n_workers=0)
client = Client(cluster)
cluster.scheduler.add_plugin(plugin)


class DownsizeSlowly(Adaptive):

    def should_scale_down(self):
        delta = datetime.datetime.now() - plugin.last_task
        if delta > datetime.timedelta(seconds=5):
            print("Scaling down")
            return True
        else:
            print("too soon: ", delta)


adapt = DownsizeSlowly(cluster.scheduler, cluster)


f = client.submit(lambda x: x + 1, 10)
```

Do you have a more useful default in mind? Or a better way override the default behavior? I suppose an alternative is to have all the scale-down logic on a separate `interval`, though I could see that getting complicated.

distributed/deploy/adaptive.py
TomAugspurger: These two cover the previous behavior, and my needs in https://github.com/dask/dask-drmaa/pull/35

Even with a `needs_resources`, we may still want individual `needs_*` methods. In the dask-drmaa PR, the current behavior checking for memory is fine, so I only needed to override `needs_cpu`.

I think generalizing may be premature (unless you can think of other resource restrictions we should be checking). Subclasses who have their own requirements can check those in their own `should_scale_up` method I think.

distributed/deploy/adaptive.py
TomAugspurger: I played around with this a bit. the basic idea was to let `needs_*` return the key-value pairs of `{'restriction': 'desired'}`, and collate those in `should_scale_up`. These would eventually make their way to `cluster.scale_up`.

The tricky part is ensuring that the eventual call to `start_workers` (only) gets the arguments it's expecting. I'll try a bit more, but haven't been able to come up with a clean solution yet.

TomAugspurger:
Sorry, just seeing that there's a CI failure here.

I'll see if I can figure out what's going on.

TomAugspurger:
Ahh, it might be related to cloudpickle 0.4.0; I'm able to reproduce the Appveyor failure on just python 2.7 with cloudpickle 0.4.0 on master:


```
pytest distributed/deploy/tests/test_local.py -k test_move_unserializable_data  --verbose -r s --timeout-method=thread --timeout=10 --durations=20
=================================================================== test session starts ===================================================================
platform darwin -- Python 2.7.13, pytest-3.2.1, py-1.4.34, pluggy-0.4.0 -- /Users/taugspurger/Envs/py27/bin/python2.7
cachedir: .cache
rootdir: /Users/taugspurger/Envs/py27/lib/python2.7/site-packages/distributed, inifile:
plugins: timeout-1.2.0, faulthandler-1.3.1
timeout: 10.0s method: thread
collected 28 items

distributed/deploy/tests/test_local.py::test_move_unserializable_data
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Captured stderr ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
distributed.scheduler - INFO -   Scheduler at: inproc://192.168.7.27/21019/1
distributed.worker - INFO -       Start worker at: inproc://192.168.7.27/21019/2
distributed.worker - INFO - Waiting to connect to: inproc://192.168.7.27/21019/1
distributed.worker - INFO -               Threads:                          8
distributed.worker - INFO -                Memory:                   10.31 GB
distributed.worker - INFO -       Local Directory: /var/folders/hz/f43khqfn7b1b1g8z_z6y3bsw0000gp/T/worker-Rt6Wql
distributed.scheduler - INFO - Register inproc://192.168.7.27/21019/2
distributed.scheduler - INFO - Starting worker compute stream, inproc://192.168.7.27/21019/2
distributed.worker - INFO -         Registered to:    inproc://192.168.7.27/21019/1
distributed.scheduler - INFO - Receive client connection: Client-eda6fa0f-876d-11e7-921b-186590cd1c87
distributed.protocol.pickle - INFO - Failed to deserialize �ccloudpickle.cloudpickle
_rehydrate_skeleton_class
q(ccloudpickle.cloudpickle
_builtin_type
qTypeTypeq�qRqUlockqhU
ObjectTypeq�qR�q        }q
U__doc__q
         T�A lock object is a synchronization primitive.  To create a lock,
call the PyThread_allocate_lock() function.  Methods are:

acquire() -- lock the lock, possibly blocking until it can be obtained
release() -- unlock of the lock
locked() -- test whether the lock is currently locked

A lock is not owned by the thread that locked it; another thread may
unlock it.  A thread attempting to lock a lock that it has already locked
will block until another thread unlocks it.  Deadlocks may ensue.q
Rq}q(U  __enter__qc__builtin__                                    s�q
getattr
qhh�qRqUlockedqhhh�qRqU
                       acquire_lockqhhh�qRqU
                                            locked_lockqhhh�qqUacquireqhhh�qRqU
                                                                               release_lockq hhh �q!Rq"Ureleaseq#hhh#�q$Rq%__exit__q&hhh&�q'Rq(utR.
  File "/Users/taugspurger/Envs/py27/lib/python2.7/site-packages/distributed/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
AttributeError: ("type object 'lock' has no attribute '__enter__'", <built-in function getattr>, (<class 'distributed.protocol.pickle.lock'>, '__enter__'))
distributed.client - ERROR - ("type object 'lock' has no attribute '__enter__'", <built-in function getattr>, (<class 'distributed.protocol.pickle.lock'>, '__enter__'))
  File "/Users/taugspurger/Envs/py27/lib/python2.7/site-packages/distributed/distributed/client.py", line 832, in _handle_report
    handler(**msg)
  File "/Users/taugspurger/Envs/py27/lib/python2.7/site-packages/distributed/distributed/client.py", line 842, in _handle_key_in_memory
    type = loads(type)
  File "/Users/taugspurger/Envs/py27/lib/python2.7/site-packages/distributed/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
AttributeError: ("type object 'lock' has no attribute '__enter__'", <built-in function getattr>, (<class 'distributed.protocol.pickle.lock'>, '__enter__'))
```

TomAugspurger:
The failure [here](https://travis-ci.org/dask/distributed/jobs/267560234#L2150) in `test_stress_scatter_death). I've re-triggered that worker.

mrocklin:
That is a known intermittent failure.  I'll take a look at it now

mrocklin:
@TomAugspurger this all looks fine to me.  You should feel free to merge when you feel it's appropriate.

TomAugspurger:
👍  thanks. I'll rebase my dask-drmaa PR on top of this and do another round on it to make sure all the necessary pieces are in place before merging.

PR #1317, opened 2017-08-07
Clean file descriptor and memory leaks

mrocklin:


mrocklin:
@pitrou if you have a moment can you take a look at the failures in travis for this PR?  We're getting out of memory errors on master when forking new processes.

mrocklin:
It looks like we're leaking file descriptors

distributed/process.py
```diff
@@ -119,39 +125,26 @@ def _start():
...
+                _call_and_set_future(loop, msg['future'], process.terminate)
+            elif op == 'stop':
+                break
+            else:
+                assert 0, msg
+
+    @classmethod
+    def _watch_process(cls, selfref, process, state, q):
+        r = repr(selfref())
+        while True:
```
pwaller: Is there a good reason to poll here instead of simply waiting for the exit with `process.join`?

mrocklin: We now use process.join *after* we poll for when the process starts.  We could probably do this more cleanly, but I'm going to punt for now.

distributed/process.py
```diff
@@ -119,58 +125,45 @@ def _start():
...
             logger.debug("[%s] created process with pid %r" % (r, state.pid))
 
-        def _process_one_message(timeout):
-            try:
-                msg = q.get(timeout=timeout)
-            except Empty:
-                pass
+        while True:
+            # Periodic poll as there's no simple way to poll a threading Queue
+            # and a mp Process at the same time.
```
pwaller: This comment is now obsolete, no?

pwaller:
There is a method, `_do_start` which now appears upon reading to be dead code - I guess it was dead before this PR but complicates understanding what's happening. Likewise with `_do_terminate`.

I assume the poll for process start is because you can't `process.join` a process before it has started?

A more straightforward architecture to analyse would be to start the process watcher thread once `process.start()` has been called, then the watcher a life cycle that better matches the process (rather than starting when the `AsyncProcess` is created).

My suggestion would be to move the `_watch_process_thread` construction out of `__init__ -> _start_threads` and instead do it immediately after `process.start()` in `_watch_message_queue.start`. Then the watcher the thread can go straight into `process.join`.

mrocklin:
I've removed `_do_start/terminate` and moved thread-starting until after we start the process.

distributed/process.py
```diff
@@ -107,70 +97,61 @@ def _run(cls, target, args, kwargs):
...
-                _maybe_notify_exit(exitcode)
-                break
-
-        while True:
-            # Idle poll
-            try:
-                _process_one_message(timeout=1e6)
-            except SystemExit:
-                return
+                q.put({'op': 'stop'})
```
pitrou: So what happens if e.g. someone calls `terminate` after the process has stopped and you have stopped the message queue thread?

distributed/process.py
```diff
@@ -107,70 +97,61 @@ def _run(cls, target, args, kwargs):
...
-
-        def _maybe_notify_exit(exitcode):
+                assert 0, msg
+
+    @classmethod
+    def _watch_process(cls, selfref, process, state, q):
+        r = repr(selfref())
+        process.join()
+        exitcode = process.exitcode
+        if exitcode is not None:
```
pitrou: This should always be true after join(), so you can turn this into an assert.

distributed/process.py
```diff
@@ -107,70 +97,61 @@ def _run(cls, target, args, kwargs):
...
+            logger.debug("[%s] process %r exited with code %r",
+                         r, state.pid, exitcode)
+            state.is_alive = False
+            state.exitcode = exitcode
+            # Make sure the process is removed from the global list
+            # (see _children in multiprocessing/process.py)
+            process.join(timeout=0)
```
pitrou: join() was already called above, so it's pointless to call it again :-)

pitrou:
I took a brief look at the AsyncProcess changes (see comments).

mrocklin:
Thanks for the feedback @pitrou .  Some of these are handled in https://github.com/dask/distributed/pull/1324

distributed/process.py
mrocklin: I don't immediately see an issue here.  All `{'op': 'stop'}` does is stop the message queue thread, so this wouldn't have a large impact.  You know this part of the code better than I do though.  If you see something amiss then please speak up.

PR #1319, opened 2017-08-07
Remove bare `except:` blocks and test that none exist.

eriknw:
If a bare except is desired, do `except: # noqa: E722`.

`pycodestyle` is used, because this is a recent feature not yet available in `pep8` or `flake8` utilities.

I was mostly lazy and used `Exception` rather than more specific exceptions, but this is still better than bare `except:`.


mrocklin:
FYI the dask/distributed repository has been having some testing difficulties.  The travis-ci tests are likely to fail through no fault of your own.

mrocklin:
Internal issues have now been resolved. Merging. Thanks @eriknw !

PR #1324, opened 2017-08-11
Clean up AsyncProcess handling

mrocklin:
We did a hasty merge in #1317 .  This includes a few follow on changes.

cc @pitrou

### DIFF
--- a/distributed/process.py
+++ b/distributed/process.py
@@ -150,7 +150,6 @@ class AsyncProcess(object):
                 self._loop.add_callback(self._on_exit, exitcode)
         finally:
             self = None  # lose reference
-            q.put({'op': 'stop'})
 
     def start(self):
         """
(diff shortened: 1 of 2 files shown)


distributed/process.py
```diff
@@ -136,22 +136,21 @@ def _watch_process(cls, selfref, process, state, q):
...
+        # Make sure the process is removed from the global list
+        # (see _children in multiprocessing/process.py)
+        # Then notify the Process object
+        self = selfref()  # only keep self alive when required
+        try:
+            if self is not None:
+                self._loop.add_callback(self._on_exit, exitcode)
+        finally:
+            self = None  # lose reference
+            q.put({'op': 'stop'})
```
pitrou: The problem here is that the message queue won't process any more messages. So it the user issues, says, `yield process.terminate()` afterwards, it will never return.

mrocklin: Is there an approach that you recommend here?  To be honest I'm not very familiar with this part of the code.

pitrou: Just let the finalizer do its job. It will trigger on close() or on garbage collection.

mrocklin:

PR #1328, opened 2017-08-11
WIP: Allow keys to be unicode

iamed2:
This is my first attempt, using `six`. 

A bunch of the tests freeze now and I have no idea why.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -386,7 +386,7 @@ def key_split_group(x):
     typ = type(x)
     if typ is tuple:
         return x[0]
-    elif typ is str:
+    elif typ is str or typ is six.text_type:
         if x[0] == '(':
             return x.split(',', 1)[0].strip('()"\'')
         elif len(x) == 32 and re.match(r'[a-f0-9]{32}', x):
@@ -535,8 +535,8 @@ def tokey(o):
     >>> tokey(1)
     '1'
     """
-    t = type(o)
-    if t is str or t is bytes:
+    typ = type(o)
+    if typ is six.binary_type or typ is six.text_type:
         return o
     else:
         return str(o)
@@ -546,9 +546,9 @@ def validate_key(k):
     """Validate a key as received on a stream.
     """
     typ = type(k)
-    if typ is not str and typ is not bytes:
+    if typ is not six.text_type and typ is not six.binary_type:
         raise TypeError("Unexpected key type %s (value: %r)"
-                        % (type(k), k))
+                        % (typ, k))
 
 
 def _maybe_complex(task):
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -356,49 +356,90 @@ except ImportError:
 else:
     key_split = lru_cache(100000)(key_split)
 
+if PY3:
+    def key_split_group(x):
+        """A more fine-grained version of key_split
 
-def key_split_group(x):
-    """A more fine-grained version of key_split
-
-    >>> key_split_group('x')
-    'x'
-    >>> key_split_group('x-1')
-    'x-1'
-    >>> key_split_group('x-1-2-3')
-    'x-1-2-3'
-    >>> key_split_group(('x-2', 1))
-    'x-2'
-    >>> key_split_group("('x-2', 1)")
-    'x-2'
-    >>> key_split_group('hello-world-1')
-    'hello-world-1'
-    >>> key_split_group(b'hello-world-1')
-    'hello-world-1'
-    >>> key_split_group('ae05086432ca935f6eba409a8ecd4896')
-    'data'
-    >>> key_split_group('<module.submodule.myclass object at 0xdaf372')
-    'myclass'
-    >>> key_split_group(None)
-    'Other'
-    >>> key_split_group('x-abcdefab')  # ignores hex
-    'x-abcdefab'
-    """
-    typ = type(x)
-    if typ is tuple:
-        return x[0]
-    elif typ is unicode or typ is str:
-        if x[0] == '(':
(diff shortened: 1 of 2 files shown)


distributed/utils.py
```diff
@@ -383,10 +383,9 @@ def key_split_group(x):
     >>> key_split_group('x-abcdefab')  # ignores hex
     'x-abcdefab'
     """
-    typ = type(x)
-    if typ is tuple:
+    if isinstance(x, tuple):
         return x[0]
-    elif typ is str:
+    elif isinstance(x, six.string_types):
```
mrocklin: We're bound by performance here.  This function is called *many* times.  The use of `type(x) is foo` is intentional. 
 This is significantly faster than calling `isinstance`.

iamed2: Ah okay I was wondering about that. I will amend.

iamed2:
Second attempt, freezing is gone, client tests passed for me locally.

distributed/utils.py
```diff
@@ -386,7 +386,7 @@ def key_split_group(x):
     typ = type(x)
     if typ is tuple:
         return x[0]
-    elif typ is str:
+    elif typ is str or typ is six.text_type:
```
mrocklin: This is going to sound a bit extreme, but I'd like to avoid the attribute access here.  I recommend importing `unicode` from `distributed.compatibility` and using that instead

distributed/utils.py
```diff
@@ -535,8 +535,8 @@ def tokey(o):
     >>> tokey(1)
     '1'
     """
-    t = type(o)
-    if t is str or t is bytes:
+    typ = type(o)
+    if typ is six.binary_type or typ is six.text_type:
```
mrocklin: the text type is the standard in Python 3, so should be the first checked here

distributed/utils.py
iamed2: I'm not familiar with performance optimization in Python so I'll defer completely to you! I'll do that.

distributed/utils.py
iamed2: Python 3 is the priority, got it 👍

distributed/utils.py
```diff
@@ -386,7 +386,7 @@ def key_split_group(x):
     typ = type(x)
     if typ is tuple:
         return x[0]
-    elif typ is str:
+    elif typ is unicode or typ is str:
```
mrocklin: In Python 3 this is redundant.  In Python 2 the bytes case is handled just below.  Would everything work just as well if you dropped the second condition here?  Alternatively we could always have two separate functions that we define based on if we're in Python 2 or Python 3.

distributed/utils.py
```diff
@@ -535,8 +535,8 @@ def tokey(o):
     >>> tokey(1)
     '1'
     """
-    t = type(o)
-    if t is str or t is bytes:
+    typ = type(o)
+    if typ is unicode or typ is bytes:
```
mrocklin: Thanks for the consistent renaming

distributed/utils.py
iamed2: In Python 2, wouldn't this case have handled `bytes` and the case below would only handle `bytes` on Python 3? Was that unintentional? I was trying to replicate the behaviour of the original.

distributed/utils.py
iamed2: I am glad that you appreciated my accident 😉

distributed/utils.py
mrocklin: Ah, I see.  OK, that makes sense.

iamed2: I could still split into a PY2 and a PY3 method if you think it's worth it

mrocklin: Can we time this for the common cases of `('x', 0, 1)` and `'x-123'` and check that things don't slow down too much?  If not then lets just go with this.

iamed2: On master:
```
>>> timeit.repeat("key_split_group('x-123')", setup="from distributed.utils import key_split_group")
[0.9205930233001709, 0.8670709133148193, 0.822983980178833]
>>> timeit.repeat("key_split_group(('x', 0, 1))", setup="from distributed.utils import key_split_group")
[0.4636390209197998, 0.4523329734802246, 0.4530479907989502]
```
On this branch:
```
>>> timeit.repeat("key_split_group('x-123')", setup="from distributed.utils import key_split_group")
[0.9638810157775879, 0.9432499408721924, 0.9448020458221436]
>>> timeit.repeat("key_split_group(('x', 0, 1))", setup="from distributed.utils import key_split_group")
[0.47734498977661133, 0.4366590976715088, 0.4450240135192871]
```
This is the time for running 1000000 executions, three times (the default settings for `timeit.repeat`).

mrocklin: OK, so it looks like a 5-20% slowdown.  This *could* go in, but if you're willing it might be nicer to build separate Python 2 and Python 3 versions of this function and decide which to use on import time.

```python
if PY3:
    def ...
        ...
else:
    def ...
```

iamed2: Sure, I'll do that.

iamed2:
I think this has addressed all comments. Is there anything I can do about the failing tests, or are those expected right now?

mrocklin:
Taking a look now

mrocklin:
This is in.  Thanks @iamed2 !

iamed2:
Thank you!

iamed2:
When should we look for a new release containing this PR?

mrocklin:
I'm cleaning things up now in hopes of a release soon.  It's difficult to say though.  Fixing bugs causes code churn which sometimes unearths more bugs.

iamed2:
Sure thing 👍

PR #1330, opened 2017-08-13
Add timeout to ThreadPoolExecutor.shutdown

mrocklin:
This significantly reduces memory leaks during testing


pitrou:
I am surprised it would reduce memory leaks. Have you found out why?

mrocklin:
> I am surprised it would reduce memory leaks. Have you found out why?

No.  This surprised me as well.  Something was keeping the state of the thread in memory.

mrocklin:
I plan to merge this soon if there are no further comments

PR #1339, opened 2017-08-16
Try custom container for hdfs testing

martindurant:
(to fix HDFS test errors)


martindurant:
@mrocklin , do I need to do something to turn back on the tests you allowed to fail?

mrocklin:
I've pushed a commit that should reactivate things (I think).

martindurant:
Thank you. Cancelling earlier builds.

martindurant:
@mrocklin : it went green after several attempts. Actually it did need a [change](https://github.com/dask/hdfs3/commit/388ad644490825f835062c8d657702eb340aebb1) in hdfs3: loading of libhdfs3 with LIBHDFS3_CONF set to a file that doesn't exist is what caused the seg-fault.
So now the changes here could be reverted or ignored, but conversely, testing of hdfs3 and knit could be changed to be based on the same docker image base. I have no strong feeling: the images were made to ease my docker test cycle, but it would be good to have consistency across projects.

mrocklin:
I don't recommend looking to me for an answer on this.  You're probably more experience here.  From my perspective you're in charge and responsible here.

continuous_integration/Dockerfile
```diff
@@ -1,17 +1,7 @@
-FROM sequenceiq/hadoop-docker:2.7.1
+FROM mdurant/docker_images:hadoop
```
martindurant: Yes - it's the name of the github repo, though (which makes more sense).

mrocklin: It seems odd to me to use the version as a name.  What will you do if you need to update the version?  I suspect (but am not sure) that this goes against common naming conventions for docker containers.

martindurant: They are different tags, which is not necessarily the same as version (e.g., ubuntu:trusty gets you the latest, security updated version of trusty, not some specific release). I suppose dockerhub allows multiple entries to pull from the same github repo, would it be better to call it mdurant/dev and mdurant/hadoop (:latest assumed)?

mrocklin:
@martindurant what is the status here?

martindurant:
@mrocklin , I changed the names as you suggested and did a bit of clean up of the previous framework. I shied away from making the decision that this was the way to go, though, I can see points on both sides. If this were to be merged, it would require updating the ci instructions README.

mrocklin:
Presumably leaving it in limbo is the wrong choice regardless :)

What is the decision that has to be made?  It seems to me that you are likely the most qualified person to make decisions here.

martindurant:
Listing considerations here for posterity.
Three projects would be impacted: this one, hdfs3 and knit.

pros of using unified docker framework
- single place to look for problems (currently the three ci setups are each different)
- a good docker environment to have locally if needing to debug in the container
- static base for testing, not rebuilt on every ci run
- somewhat faster

cons
- will not pick up new package versions of dependencies unless poked
- will require some maintenance forever (more than the scripts in the current ci folders?)
- knit is not tested like the others, since tests need to run inside the container to access the yarn CLI; perhaps other important differences between the repos
- image includes plenty of things that are non-minimal, such as full gcc build system

martindurant:
Based on the above, I don't feel a reason to make this change at the moment. Should it become more obvious, it can always be resurrected at a later date.

mrocklin:
Thanks for taking on the decision @martindurant!  I really appreciate your
work here.

Issue #1322, opened 2017-08-09
OOMKill of ipython kernel results in orphaned processes


pwaller:
I'm using jupyter-notebook 5.0.0, ipython 6.1.0. Python3.5.1, distributed 1.18.0 and dask `0.15.1+33.gaafeedc`.

I'm working on fairly large datasets which I expect to graze the RAM ceiling. Under various circumstances, my ipython kernel crashes (for example, during an OOM).

This leaves me in a state with ~2 python processes per CPU core, e.g:

```
~$ ps aux | grep 8691
pwaller   8691  0.3  0.2 573092 41156 ?        Sl   12:47   0:14 /usr/bin/python3 -B -c from multiprocessing.forkserver import main; main(77, 85, ['distributed', 'pkg_resources'], **{'sys_path': [...]})
```

In sum, they remain taking 10GiB of resident memory indefinitely. I'm regularly finding myself in this situation of having to kill these processes. Is this normal?

Clearly, I also need to figure out why the kernel is crashing in the first place (it's OOMing, but why is still a mystery to me), but I do have an expectation that if it crashes there aren't large amounts of resources left tied up in this way.

Another key problem with this is that it is not clear that these resources are still tied up except that it results in more OOMs when I try to continue my analysis later (or unless I keep my eye on the memory usage constantly).

Apologies if I missed something obvious in the documentation.

mrocklin:
cc @pitrou any thoughts on this (sorry to rely on you a bit lately)

pitrou:
> This leaves me in a state with ~2 python processes per CPU core, e.g:

That sounds expected. It it not obvious for a child process to get notified when the parent dies, and there's no solution that I can immediately think of.

mrocklin:
@pwaller you say that you have two subprocesses attached to a local IPython process.  Does this mean that you're using the local cluster, such as with `client = Client()`?

pwaller:
@pitrou 
> It it not obvious for a child process to get notified when the parent dies, and there's no solution that I can immediately think of.

Two ideas:

* Pass a pipe to the child, have the child do a blocking read from the pipe (nothing ever writes). When the read completes, this indicates the parent process has gone, because there are no remaining writers on the write side.

[...]

PR #1345, opened 2017-08-20
Exit workers immediately if parent process exits

pwaller:
This is my response to @mrocklin in [#1322](https://github.com/dask/distributed/issues/1322#issuecomment-322317319).

---

Prior to this commit, if the parent process was killed (for example due
to out-of-memory 'OOM'), then the child processes would never exit.

This could result in (potentially large) resources being tied up until
they are manually released, for example by an operator finding and killing
the processes.

This commit introduces a mechanism that enables children to determine
that the parent went away, and does not rely on the parent having time
to send any signals - so it works if the parent is SIGKILL'd.

The mechanism is described in detail in the code comments in this patch.

The test program I use to see the behaviour before and after this
commit. Without this commit, 1GiB of memory is leaked, and must be freed
with something like `pkill -f forkserver` (where `forkserver` uniquely
matches the command-line of what you intend to kill, which `forkserver`
does for me).

```python
import os
import signal
import sys

import distributed
from dask import delayed

def leak_memory():
    # Leak approx 1GiB of RAM (8 bytes per slot)
    sys.leak = [None for i in range(int(2**30 / 8))]
    return 42

if __name__ == "__main__":
    c = distributed.Client()
    q = delayed(leak_memory)()
    ret = q.compute()
    # Kill self. This results in the children remaining around since
    # they aren't signalled that we are done with them.
    os.kill(os.getpid(), signal.SIGKILL)
```

Possible fix for #1322.
/cc @pitrou @mrocklin

### DIFF
--- a/distributed/process.py
+++ b/distributed/process.py
@@ -3,6 +3,7 @@ from __future__ import print_function, division, absolute_import
 import atexit
 from datetime import timedelta
 import logging
+import os
 import sys
 import threading
 import weakref
@@ -48,8 +49,21 @@ class AsyncProcess(object):
         self._state = _ProcessState()
         self._loop = loop or IOLoop.current(instance=False)
 
+        # _keep_child_alive is the write side of a pipe, which, when it is
+        # closed, causes the read side of the pipe to unblock for reading. Note
+        # that it is never closed directly. The write side is closed by the
+        # kernel when our process exits, or possibly by the garbage collector
+        # closing the file descriptor when the last reference to
+        # _keep_child_alive goes away. We can take advantage of this fact to
+        # monitor from the child and exit when the parent goes away unexpectedly
+        # (for example due to SIGKILL). This variable is otherwise unused except
+        # for the assignment here.
+        parent_alive_pipe, self._keep_child_alive = mp_context.Pipe(duplex=False)
+
         self._process = mp_context.Process(target=self._run, name=name,
-                                           args=(target, args, kwargs))
+                                           args=(target, args, kwargs,
+                                                 parent_alive_pipe,
+                                                 self._keep_child_alive))
         _dangling.add(self._process)
         self._name = self._process.name
         self._watch_q = PyQueue()
@@ -91,8 +105,47 @@ class AsyncProcess(object):
         self._exit_future.set_result(exitcode)
 
     @classmethod
-    def _run(cls, target, args, kwargs):
+    def _immediate_exit_when_closed(cls, parent_alive_pipe):
+        """
+        Immediately exit the process when parent_alive_pipe is closed.
+        """
+        def monitor_parent():
+            try:
+                # The parent_alive_pipe should be held open as long as the
+                # parent is alive and wants us to stay alive. Nothing writes to
+                # it, so the read will block indefinitely.
+                parent_alive_pipe.recv()
+            except EOFError:
+                # Parent process went away unexpectedly. Exit immediately. Could
+                # consider other exiting approches here. My initial preference
+                # is to unconditionally and immediately exit. If we're in this
+                # state it is possible that a "clean" process exit won't work
+                # anyway - if, for example, the system is getting bogged down
+                # due to the running out of memory, exiting sooner rather than
+                # later might be needed to restore normal system function.
+                # If this is in appropriate for your use case, please file a
+                # bug.
+                os._exit(-1)
+            else:
+                # If we get here, something odd is going on. File descriptors
+                # got crossed?
+                raise RuntimeError("unexpected state: should be unreachable")
+
+        t = threading.Thread(target=monitor_parent)
+        t.daemon = True
+        t.start()
+
+    @classmethod
+    def _run(cls, target, args, kwargs, parent_alive_pipe, _keep_child_alive):
+        # On Python 2 with the fork method, we inherit the _keep_child_alive fd,
+        # whether it is passed or not. Therefore, pass it unconditionally and
+        # close it here, so that there are no other references to the pipe lying
+        # around.
+        _keep_child_alive.close()
+
         # Child process entry point
+        cls._immediate_exit_when_closed(parent_alive_pipe)
+
         threading.current_thread().name = "MainThread"
         target(*args, **kwargs)
 
--- a/distributed/tests/test_asyncprocess.py
+++ b/distributed/tests/test_asyncprocess.py
@@ -255,7 +255,7 @@ def test_child_main_thread():
     yield proc.join()
     n_threads = q.get()
     main_name = q.get()
-    assert n_threads == 1
+    assert n_threads == 2
     assert main_name == "MainThread"
     q.close()
     q._reader.close()
--- a/distributed/tests/test_asyncprocess.py
+++ b/distributed/tests/test_asyncprocess.py
@@ -16,7 +16,7 @@ from tornado.locks import Event
 from distributed.metrics import time
 from distributed.process import AsyncProcess
 from distributed.utils import ignoring, mp_context
-from distributed.utils_test import gen_test
+from distributed.utils_test import gen_test, pristine_loop
 
 
 def feed(in_q, out_q):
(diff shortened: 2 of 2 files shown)


pwaller:
I just made some tweaks to the patch in the last few minutes, but I'm now done and won't touch it again until after receiving comments.

distributed/process.py
```diff
@@ -90,9 +104,42 @@ def _on_exit(self, exitcode):
...
+                # Parent process went away unexpectedly. Exit immediately. Could
+                # consider other exiting approches here. My initial preference
+                # is to unconditionally and immediately exit. If we're in this
+                # state it is possible that a "clean" process exit won't work
+                # anyway - if, for example, the system is getting bogged down
+                # due to the running out of memory, exiting sooner rather than
+                # later might be needed to restore normal system function.
+                # If this is in appropriate for your use case, please file a
+                # bug.
+                os._exit(-1)
```
mrocklin: Does this allow `atexit` calls to run?  We do some cleanup of temporary files.

distributed/process.py
```diff
@@ -90,9 +104,42 @@ def _on_exit(self, exitcode):
...
+            else:
+                # If we get here, something odd is going on. File descriptors
+                # got crossed?
+                raise RuntimeError("unexpected state: should be unreachable")
+
+        t = threading.Thread(target=monitor_parent, daemon=True)
```
mrocklin: The `daemon=True` keyword fails in Python 2, which we still support.

pwaller: Fixed. I added a commit on the end to preserve the history which can be sqaush'd away during merge or I can do it beforehand, as you prefer.

distributed/process.py
pwaller: Hm, this is a tricky one. Partially, I made this choice of exiting because it is the easy choice: it works. (A quick attempt with the other methods don't work as I'd expect, so there would be more work to do to make this work.)

It is correct that atexit handlers don't run; `finalizers` also don't run. On the one hand this is bad (in the case of temporary files it's another potential resource leak). On the other hand it gets us to our intended final state more quickly (not hogging resources). It's also the case that this is currently used in an extreme situation: the parent process has gone unexpectedly. Maybe the child processes should follow suit?

I suppose you're talking about the pid files, yes? As far as I can tell, those should be unaffected, because the process which is `_exit`ing is a thing managed by an `AsyncProcess`, not the main thread. The main thread in my use case has been `Killed` - so would leak its pidfile in any case, if you had one.

pwaller:
> Are you able to add a test for this? I imagine that this is somewhat challenging, but if you can accomplish this I think that it's much more likely that this will continue to function well in the future.

Yes, I have thought about this, and I thought about it more in response to your comment. I can see how it's possible: have the test invoke a process which creates a `Client`, and then verify that this new process's sub-processes die when the process which created the `Client` dies through `SIGKILL`.

But I don't have time to execute on this for the moment. I may get enough uninterrupted time later, but I may not.

I have just submitted two new patches to fix test failures seen on Travis. I'll try to remember to check up n them and fix any further issues it notices. Please ping me if you see failed travis tests and no response from me for a while.

pitrou:
@pwaller, could you add a dedicated test for this feature?

distributed/process.py
pwaller: Hm. One issue with this: the thing sending the signals will also receive it. So it will receive the `SIGINT`, and not likely run the `SIGKILL` later. I could go the route of masking `SIGINT` with `signal.pthread_kill`, but then we stray from portability. Thoughts?

pitrou: Oh, you're right. It's probably overkill.

pwaller:
I found the time to put together a test. You were right @mrocklin, it was challenging ;-). I hope the chatty code comments are warranted. My code doesn't usually look this comment-dense, but it is sufficiently subtle I think it benefits in this case.

The test is designed so that if it fails, it doesn't leak a process, nor does it block all other tests for too long - at most 5 seconds. I've also put effort into thinking about how it will run on a busy or slow system - I believe it should function reliably, since the short wait doesn't occur in parallel with anything expensive. 

On my system, the test takes about 1s to run when working correctly.

I do not know if this test will work on windows or not. I rely on being able to pass file descriptors from one process to another so that the test can learn whether the child process exited correctly or not. This might work? If not, I guess the test will need to be masked there, and also the mechanism introduced inthis PR itself will not function there.

For reviewing the test, one can cherry-pick 0a52c70 onto the master branch and observe the test failure.

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +300,124 @@ def test_terminate_after_stop():
...
     yield proc.terminate()
+
+
+def _worker_process(worker_ready, child_pipe):
+    # child_pipe is the write-side of the children_alive pipe held by the
+    # test process. When this _worker_process exits, this file descriptor should
+    # have no references remaining anywhere and be closed by the kernel. The
+    # test will therefore be able to tell that this process has exited by
+    # reading children_alive.
+    _ = child_pipe
```
pitrou: I don't see any reason for this line.

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +300,124 @@ def test_terminate_after_stop():
...
+    """ Check that a child process started by AsyncProcess exits if its parent
+    exits.
+
+    The motivation is to ensure that if an AsyncProcess is created and the
+    creator process dies unexpectedly (e.g, via Out-of-memory SIGKILL), the
+    child process and resources held by it should not be leaked.
+
+    The child should monitor its parent and exit promptly if the parent exits.
+
+    [test process] -> [parent using AsyncProcess (dies)] -> [worker process]
```
pitrou: Nice diagram :-)

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +300,124 @@ def test_terminate_after_stop():
...
+                 \                                          /
+                  \________ <--   child_pipe   <-- ________/
+    """
+    ctx = multiprocessing.get_context('spawn')
```
pitrou: I'm curious: why do you need this? This is Python 3-only by the way.

pitrou: Also, we have `mp_context` in `distributed.utils` which is probably what we want to test against.

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +300,124 @@ def test_terminate_after_stop():
...
+
+        # If this assert fires, then something went wrong. Either the child
+        # should write into the pipe, or it should exit and the pipe should be
+        # closed (which makes it become readable).
+        assert readable
+
+        try:
+            # This won't block due to the above 'assert readable'.
+            result = children_alive.recv()
+        except EOFError:
```
pitrou: `with pytest.raises(EOFError)` would be a shorter spelling of the whole try/except/else block.

pitrou:
> I do not know if this test will work on windows or not. I rely on being able to pass file descriptors from one process to another so that the test can learn whether the child process exited correctly or not. This might work?

When using `multiprocessing.Pipe`, it should work on Windows, yes.

mrocklin:
The test failures on travis-ci are unrelated and have been very recently fixed in master.  I recommend pulling in the master branch to avoid them.

distributed/tests/test_asyncprocess.py
pwaller: Hm. On the other hand, I want to lead any poor debugger to follow the "result" value from the exception message back to whence it came from grep. Does pytest.raises print the return value in the same way?

distributed/tests/test_asyncprocess.py
pwaller: I believe this is needed because of what I observed when I used the default `forkserver` - when I used the forkserver, the `child_pipe` passed into the `_parent_process` is not literally the same `fd` as the one that ends up being passed in to the `_worker`, so the signalling mechanism doesn't work. If you look closely at line 340 and also the diff to `AsyncProcess.__init__`, I had to also use the `spawn` method in `AsyncProcess`. This does unfortunately mean that this doesn't test the forking method ordinarily used, but I couldn't see a quick way to get it to work there. Any advice?

distributed/tests/test_asyncprocess.py
pitrou: Probably not (though py.test does lot of unexpected magic).

pwaller:
> I recommend pulling in the master branch to avoid them.

* Rebased against master.
* Amended to remove `_ = child_pipe` assignment.

I believe I have replied to all comments. Please ping me if I missed something.

distributed/tests/test_asyncprocess.py
pwaller: Removed.

distributed/tests/test_asyncprocess.py
pwaller: I prefer sticking with the less magical version if that's OK?

pwaller:
Sorry, I forgot that I still have to make this work on python2. Will see if I can squeeze that in later.

distributed/tests/test_asyncprocess.py
pitrou: I'm a bit surprised. Ideally `EOFError` should be raised even though the fd was recreated in the child through whatever means (and the official multiprocessing test suite seems [to support it](https://github.com/python/cpython/blob/master/Lib/test/_test_multiprocessing.py#L2736)).

In any case, your PR *has to* work with `forkserver` since that's what we're using :-) Making it work only with `spawn` is a little use...

distributed/tests/test_asyncprocess.py
pitrou: That's fine with me.

pwaller:
I think my theory is incorrect. I can't figure out where the additional file descriptor is coming from. What I see is that the pipe associated with `parent_alive_pipe` somehow ends up as two file descriptors in the worker process on python 2, and just one on python 3. This means that the file descriptor is not closed, and therefore the mechanism doesn't work correctly.

pitrou:
@pwaller, I would expect it's the `keep_child_alive` fd that survives in the child. In which case, the solution should be to pass `keep_child_alive` explicitly to the child and call the `close()` method on it there?

(for the record, Python 2 uses `fork` unconditionally on Unix, which is why you're seeing this)

pwaller:
I had just arrived at the same solution. Patch incoming.

pwaller:
PTAL. Believe it should work on Python2 now, and I removed my use of `spawn`.

distributed/process.py
```diff
@@ -91,8 +105,47 @@ def _on_exit(self, exitcode):
...
+        t = threading.Thread(target=monitor_parent)
+        t.daemon = True
+        t.start()
+
+    @classmethod
+    def _run(cls, target, args, kwargs, parent_alive_pipe, _keep_child_alive):
+        # On Python 2 with the fork method, we inherit the _keep_child_alive fd,
+        # whether it is passed or not. Therefore, pass it unconditionally and
+        # close it here, so that there are no other references to the pipe lying
+        # around. Therefore, we pass
```
pitrou: Seems like this sentence is unterminated.

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +300,120 @@ def test_terminate_after_stop():
...
+    # test will therefore be able to tell that this process has exited by
+    # reading children_alive.
+
+    # Signal to parent process that this process has started and made it this
+    # far. This should cause the parent to exit rapidly after this statement.
+    worker_ready.close()
+
+    # The parent exiting should cause this process to os._exit from a monitor
+    # thread. This sleep should never return.
+    shorter_timeout = 2.5 # timeout shorter than that in the spawning this.
```
pitrou: A missing word perhaps in "the spawning this"?

distributed/tests/test_asyncprocess.py
```diff
@@ -2,6 +2,7 @@
 
 from datetime import timedelta
 import gc
+import multiprocessing
```
pitrou: I believe the multiprocessing import is now obsolete?

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +300,120 @@ def test_terminate_after_stop():
...
+                              args=(worker_ready, child_pipe))
+
+        yield worker.start()
+
+        # Close our reference to the worker_ready pipe, so that the child has
+        # the only one.
+        worker_ready.close()
+
+        # Wait for the child process to have started.
+        yield PipeIOStream(await_worker_start.fileno()).read_until_close()
```
pitrou: `PipeIOStream` will probably not work on Windows. We will probably have to use a blocking call instead, such as `await_worker_start.read()`. But then the Pipe trick is not really useful anymore, and a `multiprocessing.Event` object may work just as well (and make the code a bit simpler).

pwaller: I did the switch to Event. It causes python to show this at the end on python3:

```
/usr/lib/python3.5/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 5 leaked semaphores to clean up at shutdown
  len(cache))
```

You also say the pipe trick is not useful - is that to say that it won't work at all on windows? If that is the case, then the main function won't work on windows.

distributed/tests/test_asyncprocess.py
pwaller: Removed.

distributed/tests/test_asyncprocess.py
pwaller: Meant to say "spawning test" rather than "spawning this". Fixed.

distributed/process.py
pwaller: Removed.

pwaller:
PTAL.

distributed/tests/test_asyncprocess.py
pwaller: Sorry for the confusing comment, I meant to say: "The main purpose of this PR won't work on windows". (Maybe that's OK? It's only a safety net for a crash condition.)

pitrou: The semaphore tracker warning is probably ok since we're brutally crashing a process without closing the synchronization primitives it created.

As for the pipe trick, using `Pipe.read()` and relying on `EOFError` should be ok (the AppVeyor run will tell us if there's a problem anyway). What wouldn't work is calling `fileno()` and expecting Tornado's `PipeIOStream` to work with it, since Unix fds and Windows handles work differently.

pwaller:
@pitrou I have little recent windows dev experience. The test fails on Appveyor with:

```
_______________ test_asyncprocess_child_teardown_on_parent_exit _______________
  File "C:\projects\distributed\distributed\tests\test_asyncprocess.py", line 390, in test_asyncprocess_child_teardown_on_parent_exit
    readable = children_alive.poll(short_timeout)
  File "C:\Miniconda35-x64\envs\testenv\lib\multiprocessing\connection.py", line 257, in poll
    return self._poll(timeout)
  File "C:\Miniconda35-x64\envs\testenv\lib\multiprocessing\connection.py", line 328, in _poll
    _winapi.PeekNamedPipe(self._handle)[0] != 0):
BrokenPipeError: [WinError 109] The pipe has been ended
____________________ test_broken_worker_during_computation ____________________
```

A little unfortunate that the poll sees a `BrokenPipeError` I see that this is only a Python3 thing - is there a simple way to catch it or WinError in a portable manner? I assume that we may also get this error instead of EOFError from the `recv`.

pwaller:
Sorry, I mean `OSError` or `BrokenPipeError`, if it is available. I have made an attempted fix in 8286ab8, let's see what appveyor thinks. Does it test python2 on windows? I don't have the ability to test there otherwise.

@pitrou with respect to the semaphore warning introduced by using the Event - should I:

* go back to using a pipe (but not using `PipeIOStream`) - which did not have the warning
* find a way of ignoring the warning (not sure this is a good idea)
* live with the warning as is

pitrou:
> Does it test python2 on windows?

No, it doesn't. For making the test Python 2 compatible, i'd suggest catching `EnvironmentError` instead of `BrokenPipeError`. Perhaps also assert that we're under Windows, as `poll()` shouldn't raise on Unix.

pitrou:
As for the warning, we should simply live with it.

pwaller:
`EnvironmentError` caught in 9d53fd1, which is an amendment to the final commit.

pwaller:
Missed one. Amended again in 2a9d9f1.

pwaller:
Tests all passed.

distributed/tests/test_asyncprocess.py
```diff
@@ -297,3 +299,124 @@ def test_terminate_after_stop():
...
+
+        try:
+            # This won't block due to the above 'assert readable'.
+            result = children_alive.recv()
+        except EOFError:
+            pass # Test passes.
+        except EnvironmentError:
+            # Windows can raise BrokenPipeError. EnvironmentError is caught for
+            # Python2/3 portability.
+            assert sys.platform.startswith('win'), "should not raise on windows"
```
pitrou: "Should not raise on Windows" should probably be "Should not raise on non-Windows".

distributed/tests/test_asyncprocess.py
```diff
@@ -12,11 +12,13 @@
 import pytest
 from tornado import gen
 from tornado.locks import Event
+from tornado.iostream import PipeIOStream
```
pitrou: I suspect this import isn't used anymore.

pitrou:
LGTM apart from two nits above.

pwaller:
Fixed above two nits, squashed to two commits (one for new feature, one for new test).

mrocklin:
Thanks for your efforts here @pwaller and for the review @pitrou .  I've merged this into master.

PR #1358, opened 2017-08-24
ENH: Worker Environments

TomAugspurger:
Adds the notion of worker environments to the client and scheduler.
These allow for specifying certain setup and teardown methods that
should be called when workers enter or exit the pool. In the future
environments could allow for scheduling decisions, e.g. restricting
certain tasks to certain environments.

```python
>>> client.register_worker_environment("my-env",
...                                    condition=lambda: True)
```

This is pretty rough at the moment.

### DIFF
--- a/distributed/environments.py
+++ b/distributed/environments.py
@@ -1,6 +1,3 @@
-from tornado import gen
-
-
 class WorkerEnvironment(object):
     """A worker environment.
 
@@ -21,7 +18,6 @@ class WorkerEnvironment(object):
     def __repr__(self):
         return "<WorkerEnvironment({})>".format(self._name or '')
 
-    @gen.coroutine
     def condition(self):
         """A function be run on a worker determining whether
         it qualifies for this environment.
@@ -36,16 +32,15 @@ class WorkerEnvironment(object):
         in the environment.
         """
         if self._condition:
-            yield gen.maybe_future(self._condition())
+            result = self._condition()
+            return result
         return True
 
-    @gen.coroutine
     def setup(self):
         """Run on the worker when it's added to the environment"""
         if self._setup:
             return self._setup()
 
-    @gen.coroutine
     def teardown(self):
         """Run on the worker when it's removed from the environment"""
         if self._teardown:
--- a/distributed/environments.py
+++ b/distributed/environments.py
@@ -1,0 +1,52 @@
+from tornado import gen
+
+
+class WorkerEnvironment(object):
+    """A worker environment.
+
+    Parameters
+    ----------
+    condition : callable
+    setup : callable
+    teardown : callable
+    """
+
+    _name = None  # Used only for a nicer repr, and nothing else
+
+    def __init__(self, condition=None, setup=None, teardown=None):
+        self._condition = condition
+        self._setup = setup
+        self._teardown = teardown
+
+    def __repr__(self):
+        return "<WorkerEnvironment({})>".format(self._name or '')
+
+    @gen.coroutine
+    def condition(self):
+        """A function be run on a worker determining whether
+        it qualifies for this environment.
+
+        Returns
+        -------
+        qualifies : bool
+
+        Notes
+        -----
+        By default, the just returns True, meaning every worker falls
+        in the environment.
+        """
+        if self._condition:
+            return self._condition()
+        return True
+
+    @gen.coroutine
+    def setup(self):
+        """Run on the worker when it's added to the environment"""
+        if self._setup:
+            return self._setup()
+
+    @gen.coroutine
+    def teardown(self):
+        """Run on the worker when it's removed from the environment"""
+        if self._teardown:
+            return self._teardown()
--- a/distributed/environments.py
+++ b/distributed/environments.py
@@ -36,7 +36,7 @@ class WorkerEnvironment(object):
         in the environment.
         """
         if self._condition:
-            return self._condition()
+            yield gen.maybe_future(self._condition())
         return True
 
     @gen.coroutine
--- a/distributed/tests/test_worker_environments.py
+++ b/distributed/tests/test_worker_environments.py
@@ -6,7 +6,7 @@ from distributed.metrics import time
 from distributed.utils_test import gen_cluster, cluster, loop  # noqa
 
 class MyEnv(WorkerEnvironment):
-    """Test worker environmnet"""
+    """Test worker environment"""
     data = 0
 
     def condition(self):
@@ -20,9 +20,9 @@ class MyEnv(WorkerEnvironment):
 
 
 @gen_cluster(client=True)
-def test_environmnets_register(c, s, a, b):
+def test_environment_register(c, s, a, b):
     # Register environment
-    yield c._environments_register('myenv', MyEnv())
+    yield c._environment_register('myenv', MyEnv())
     assert len(s.worker_environments) == 1 and 'myenv' in s.worker_environments
     assert len(s.environment_workers['myenv']) == 2
     assert 'myenv' in a.environments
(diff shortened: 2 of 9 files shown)


distributed/tests/test_worker_environments.py
```diff
@@ -0,0 +1,96 @@
...
+
+    assert w.address in s.environment_workers['myenv']
+    assert len(s.environment_workers['never_passes']) == 0
+
+    # Remove worker
+    yield w._close()
+    start = time()
+    while not w.status == 'closed':
+        yield gen.sleep(0.01)
+        assert time() - start < 5
```
mrocklin: Is this timing block necessary?

distributed/tests/test_worker_environments.py
```diff
@@ -0,0 +1,96 @@
...
+
+    assert w.address not in s.environment_workers['myenv']
+
+
+@gen_cluster(client=True)
+def test_register_condition_for_worker_raises(c, s, a, b):
+    r = c._register_worker_environment('myenv', lambda: True)
+    with pytest.raises(TypeError) as m:
+        r.result()
```
mrocklin: FWIW I *always* use the `yield` syntax everywhere throughout the project.

distributed/worker.py
```diff
@@ -2233,6 +2238,47 @@ def _get_client(self, timeout=3):
...
         return self._client
 
+    @gen.coroutine
+    def register_worker_environments(self, stream=None, environments=None):
+        result = {}
+        for name, env in environments.items():
+            if name not in self.environments:
+                try:
+                    env = pickle.loads(env)
+                    meets_condition = yield self.executor_submit('condition', env.condition)
```
mrocklin:
```python
from .comm.utils import offload
meets_condition = yield offload(env.condition)
```

We should move this to `.utils`

distributed/worker.py
```diff
@@ -2233,6 +2238,47 @@ def _get_client(self, timeout=3):
...
+                    # TODO: Should this be in some kind of "while running" loop with
+                    # a timeout?
+                    if gen.is_future(meets_condition):
+                        meets_condition = meets_condition.result()
```
mrocklin: Prefer `yield` if that's ok

distributed/worker.py
```diff
@@ -2233,6 +2238,47 @@ def _get_client(self, timeout=3):
...
+                    if meets_condition:
+                        # TODO: This currently blocks. I expect that that is
+                        # not desirable, as a long setup task may look like a
+                        # worker that's become unresponsive.
```
mrocklin: This doesn't seem to block anymore.  I suspect that this comment is out of date.

distributed/worker.py
```diff
@@ -2233,6 +2238,47 @@ def _get_client(self, timeout=3):
...
+                        yield self.executor_submit('setup', env.setup)
```
mrocklin: What if setup/condition are coroutines?  This might not function well.

Actually, I'm tempted to just run them locally as is without a separate thread.  If people care about non-blocking behavior then they should write coroutines.  I think that we can expect a bit of expertise from anyone using Environments.  Most of the time this probably won't be an issue.

TomAugspurger:
> `starting all methods with environment_` 

Do you prefer `environment_register` or `worker_environment_register`? Do you anticipate any non-worker environments in the future?

mrocklin:
environment_register sounds good to me

TomAugspurger:
Just to make sure I understand the running of methods on workers correctly:

> Actually, I'm tempted to just run them locally as is without a separate thread. If people care about non-blocking behavior then they should write coroutines. I think that we can expect a bit of expertise from anyone using Environments. Most of the time this probably won't be an issue.

What would that look like on `Worker.environment_register`? Since distributed is calling the function or coroutine, aren't we responsible for calling it on the main thread or not? Perhaps I misunderstood.

---

My latest push simplifies things:

- Renamed all `register_worker_environmnet` style methods to `environment_register` style
- Removed essentially all exception handling, letting `.rpc` handle it for us
- Added environment de-registration
- Make all methods (de)register a single environment at once (this was the case for the `Client` already). There's a small utility method on the scheduler for registering many environments, when a new worker is added
- Made `WorkerEnvironment` methods regular functions (not coroutines).
- Call `WorkerEnvironment` methods on the main thread

I'm not sure if those last two are desirable or not.

distributed/environments.py
```diff
@@ -0,0 +1,52 @@
...
+        -------
+        qualifies : bool
+
+        Notes
+        -----
+        By default, the just returns True, meaning every worker falls
+        in the environment.
+        """
+        if self._condition:
+            return self._condition()
```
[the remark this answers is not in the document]
TomAugspurger: FYI, `gen.maybe_future` seems to be deprecated: http://www.tornadoweb.org/en/stable/gen.html#tornado.gen.maybe_future.

TomAugspurger:
Moving [this](https://github.com/dask/distributed/pull/1358#discussion_r135114979) comment up, since I haven't addressed it yet.

# Summary of Changes

Equality Semantics

> This will happen concurrently with other things. We might want to avoid either having the worker start work or having it even be known in the scheduler until the environment membership is known.
> 
> I see two options:
> 
> 1. Change add_worker and remove_worker to be coroutines. Use `self.rpc(address).register_environment(...)` before we kick things off. This is best because the worker won't be assigned any work until this comes back.
> 2. Add environments to the handshake that occurs in the beginning of handle_worker (look for 'compute-stream').
> This could be part of the original handshake or could come directly afterwards or before. This is simpler, doesn't give the worker any work before its environment suite is known, but might inadvertently schedule work beforehand.
>
> Actually, if all scheduling is going to be environment inclusive (e.g. run this task on a GPU node) then the second case might suffice.

docs/source/worker-environments.rst
```diff
@@ -0,0 +1,84 @@
...
+   Determines whether that worker qualifies for the environment.
+3. A ``setup`` a function of no arguments to be run on the worker when they
+   join an environment
+4. A ``teardown`` function of no arguments to be run on the worker when they
+   leave the environment
+5. A ``state`` variable (TODO: what exactly is this? Tentatively, each worker
+   has a dictionary ``environment_state :: name -> Dict``, where you can get and
+   put whatever you want in that nested ``Dict``)
+
+TODO: Can / should the setup or teardown methods take the ``dask_worker`` itself?
```
mrocklin: This sounds useful to me.

docs/source/worker-environments.rst
```diff
@@ -0,0 +1,84 @@
...
+1. A unique name like ``"high-memory"`` or ``"database-access"``
+2. A ``condition``, a function of no arguments to be called on the worker.
+   Determines whether that worker qualifies for the environment.
+3. A ``setup`` a function of no arguments to be run on the worker when they
+   join an environment
+4. A ``teardown`` function of no arguments to be run on the worker when they
+   leave the environment
+5. A ``state`` variable (TODO: what exactly is this? Tentatively, each worker
+   has a dictionary ``environment_state :: name -> Dict``, where you can get and
+   put whatever you want in that nested ``Dict``)
```
mrocklin: Is having state necessary?  Perhaps drop this but include examples of environment instances with state?

docs/source/worker-environments.rst
```diff
@@ -0,0 +1,84 @@
...
+   client.environment_register(name='high-memory',
+                               condition=has_high_memory)
+
+
+Or class-based style by passing an instance of :class:`WorkerEnvironment`:
+
+.. code-block:: python
+
+   client.environment_register(name='high-memory',
+                               environment=HighMemory)
```
mrocklin: This example needs a class definition?

docs/source/worker-environments.rst
mrocklin: In `client.run` we magically look for the special `dask_worker` argument, and if it is there then we populate it accordingly.  I'm not sure if we want to repeat this pattern, but it has so far worked out well.

mrocklin:

PR #1361, opened 2017-08-26
Avoid storing large messages in message log

mrocklin:
The test here still fails.  I'm not yet sure why.

#2625


mrocklin:
We have to send one more message down the comm.  I'm ok with this.

PR #1370, opened 2017-08-31
Support unicode keys in pack_data

mrocklin:
In https://github.com/dask/distributed/pull/1328 we started allowing
Python 2 to use unicode keys, mostly to support Julia.  However this
change missed a the pack_data function used in worker.py which still
only checked for string-based keys.  This arose any time whenever
a task depended on a unicode key.

Fixes https://github.com/pydata/xarray/issues/1540

cc @iamed2


iamed2:
Drat, sorry!

PR #1372, opened 2017-09-04
Cleaning up intro paragraph

EdrianI:
"from local machine may not be work." to "from local machine may not work."


mrocklin:
Thanks @EdrianI !

PR #1383, opened 2017-09-08
Offload upload_file Disk IO to separate thread

mrocklin:
cc @adamklein


mrocklin:
This helps, especially with slow disk, but unfortunately we still have a dependence on the size of the data from network I/O, which still blocks a bit.

mrocklin:
More broadly this serves as a pattern for how to offload I/O tasks to a separate Python thread.

Issue #1377, opened 2017-09-06
SSLContext Error when programmatically closing down a worker


ywangd:
I tried to setup SSL connections following the instruction [here](https://distributed.readthedocs.io/en/latest/tls.html) and it was successful (version 1.18.0). 

One issue that I have experienced is that a `SSLContext` error is thrown when closing down a worker programmatically via `scheduler.retire_workers`. The stacktrace is rather long so I'll just show the last few lines:
```
File "/opt/local/anaconda2/envs/mdsdispatcher/lib/python2.7/site-packages/distributed/comm/tcp.py", line 268, in _expect_tls_context
    raise TypeError("TLS expects a `ssl_context` argument of type "
TypeError: TLS expects a `ssl_context` argument of type ssl.SSLContext (perhaps check your TLS configuration?)
```

This can be fixed by adding a `connection_args` argument to https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L542

i.e. replace  
```python
with rpc(address) as r:
```
with 
```python
with rpc(address, connection_args=self.connection_args) as r:
```

I am not all that clear about how communication is done with distributed. But the same issue could potentially exist in other places where `rpc(...)` is used, e.g. https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L1611

Thanks!

mrocklin:
Yes, I think that your assessment of the situation is accurate.  Do you have any interest including that fix in a PR?

ywangd:
Pull request here #1385
Thanks!

Corni:
Hi, this issue looks like it is already fixed and should be closed, shouldn't it?
If this is not true, please sorry for the noise!

mrocklin:
Thank you @Corni !  Closing

PR #1385, opened 2017-09-11
Fix issue #1377 - missing SSLContext

ywangd:
Add missing connection arguments to rpc init calls so that SSL
connection is correctly handled when retiring/restarting workers.


mrocklin:
Thanks for sending this in @ywangd !

PR #1386, opened 2017-09-12
Resolve #1382 by removing hostname from dns cache on worker removal

### LINKED ISSUE #1382: Scheduler not robust to worker dns changes
If a worker connects to the scheduler and gives a name rather than an IP address, and then subsequently the worker's IP address changes, the scheduler still tries to connect using the old IP address.

The scheduler appears to resolve the name and then cache the looked up address, leading to this issue.

Perhaps the scheduler should invalidate the cache when a worker registers with the scheduler?

rbubley:
Add explicit memoize to ensure_ip, enable deletion of a cached ip, and delete it when removing a worker.

### DIFF
--- a/distributed/comm/inproc.py
+++ b/distributed/comm/inproc.py
@@ -327,8 +327,11 @@ class InProcBackend(Backend):
         self.manager.validate_address(loc)
         return self.manager.ip
 
-    def resolve_address(self, loc, clear_cache):
+    def resolve_address(self, loc):
         return loc
+
+    def clear_address_cache(self, loc):
+        pass
 
     def get_local_address_for(self, loc):
         self.manager.validate_address(loc)
--- a/distributed/comm/inproc.py
+++ b/distributed/comm/inproc.py
@@ -327,7 +327,7 @@ class InProcBackend(Backend):
         self.manager.validate_address(loc)
         return self.manager.ip
 
-    def resolve_address(self, loc):
+    def resolve_address(self, loc, clear_cache):
         return loc
 
     def get_local_address_for(self, loc):
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -473,8 +473,12 @@ class BaseTCPBackend(Backend):
     def get_address_host_port(self, loc):
         return parse_host_port(loc)
 
-    def resolve_address(self, loc):
+    def resolve_address(self, loc, clear_cache=False):
         host, port = parse_host_port(loc)
+        if clear_cache:
+            result = ensure_ip(host)
+            ensure_ip.clear_cache(host)
+            return unparse_host_port(result, port)
         return unparse_host_port(ensure_ip(host), port)
 
     def get_local_address_for(self, loc):
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -473,13 +473,13 @@ class BaseTCPBackend(Backend):
     def get_address_host_port(self, loc):
         return parse_host_port(loc)
 
-    def resolve_address(self, loc, clear_cache=False):
+    def resolve_address(self, loc):
         host, port = parse_host_port(loc)
-        if clear_cache:
-            result = ensure_ip(host)
-            ensure_ip.clear_cache(host)
-            return unparse_host_port(result, port)
         return unparse_host_port(ensure_ip(host), port)
+
+    def clear_address_cache(self, loc):
+        host, port = parse_host_port(loc)
+        ensure_ip.clear_cache(host)
 
     def get_local_address_for(self, loc):
         host, port = parse_host_port(loc)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3161,6 +3161,8 @@ class Scheduler(ServerNode):
         return addr
 
     def clear_ip_cache(self, addr):
+        if addr in self.aliases:
+            addr = self.aliases[addr]
         if isinstance(addr, tuple):
             addr = unparse_host_port(*addr)
         if not isinstance(addr, six.string_types):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3163,7 +3163,10 @@ class Scheduler(ServerNode):
     def clear_ip_cache(self, addr):
         if isinstance(addr, tuple):
             addr = unparse_host_port(*addr)
-            clear_address_cache(addr)
+        if not isinstance(addr, six.string_types):
+            raise TypeError("addresses should be strings or tuples, got %r"
+                            % (addr,))
+        clear_address_cache(addr)
 
 
     def coerce_hostname(self, host):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -904,12 +904,13 @@ class Scheduler(ServerNode):
             del self.worker_comms[address]
             del self.ncores[address]
             self.workers.remove(address)
-            del self.aliases[self.worker_info[address]['name']]
+            _ = self.coerce_address(self.worker_info[address]['name'], clear_cache=True)
             del self.worker_info[address]
             if address in self.idle:
                 self.idle.remove(address)
             if address in self.saturated:
                 self.saturated.remove(address)
+
 
             recommendations = OrderedDict()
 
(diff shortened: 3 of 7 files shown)


distributed/comm/inproc.py
```diff
@@ -327,7 +327,7 @@ def get_address_host(self, loc):
         self.manager.validate_address(loc)
         return self.manager.ip
 
-    def resolve_address(self, loc):
+    def resolve_address(self, loc, clear_cache):
```
mrocklin: It seems odd to have this keyword here where it doesn't make much sense.

distributed/comm/tcp.py
```diff
@@ -473,8 +473,12 @@ def get_address_host(self, loc):
     def get_address_host_port(self, loc):
         return parse_host_port(loc)
 
-    def resolve_address(self, loc):
+    def resolve_address(self, loc, clear_cache=False):
         host, port = parse_host_port(loc)
+        if clear_cache:
+            result = ensure_ip(host)
+            ensure_ip.clear_cache(host)
+            return unparse_host_port(result, port)
```
mrocklin: Maybe the thing to do here is to invalidate the cache before calling `ensure_ip`.  That might result in slightly simpler code.

distributed/scheduler.py
```diff
@@ -904,13 +904,14 @@ def remove_worker(self, comm=None, address=None, safe=False, close=True):
             del self.worker_comms[address]
             del self.ncores[address]
             self.workers.remove(address)
-            del self.aliases[self.worker_info[address]['name']]
+            _ = self.coerce_address(self.worker_info[address]['name'], clear_cache=True)
```
mrocklin: It seems odd to me to call coerce address in order to invalidate a cache.  I would also expect any call to coerce address to add the new value back into the cache after invalidation.  

In either case do we still delete the name from self.aliases?

distributed/scheduler.py
```diff
@@ -3144,15 +3145,18 @@ def coerce_address(self, addr, resolve=True):
         """
         # XXX how many address-parsing routines do we have?
         if addr in self.aliases:
-            addr = self.aliases[addr]
+            if clear_cache:
+                del self.aliases[addr]
```
mrocklin: Deleting the aliases should only happen during a remove_worker event.  Placing this code here confuses me.

distributed/utils.py
```diff
@@ -477,31 +477,49 @@ def silence_logging(level, root='distributed'):
...
+                return results[0][4][0]
+
+        raise exc
 
-    raise exc
+    def clear_cache(self, hostname):
+        try:
+            del self.ip_cache[hostname]
+        except KeyError:
+            pass
```
mrocklin: This class seems a little bit special purpose.  Long term I've opened https://github.com/pytoolz/toolz/issues/383 , which I think would be a slightly cleaner way to do this.  Short term I'm still thinking about this.

distributed/scheduler.py
rbubley: The caching happens after the logic to strip off the protocol and port – the choice was either to have it in coerce_address, which already manages this stripping, or to duplicate this logic in a separate function. (Or to factor out this logic and use it in both).

distributed/scheduler.py
rbubley: You are right, and this is actually redundant here.

rbubley:
For testing, I wonder if we could use two different loopback addresses.

mrocklin:
> For testing, I wonder if we could use two different loopback addresses.

Interesting.  How would this work?

rbubley:
> > For testing, I wonder if we could use two different loopback addresses.

> Interesting. How would this work?

Conceptually:
```
dask-worker --listen-address 127.0.0.1:9000 --contact-address localhost:9000
```
Wait a few seconds and then kill it.

Reroute localhost by editing /etc/hosts

Plumb the new address into the loopback
```
sudo ifconfig lo0 127.0.0.2 netmask 255.0.0.0 up
```

Restart worker with new details:
```
dask-worker --listen-address 127.0.0.2:9000 --contact-address localhost:9000
```

Check for errors.
Reset network configuration to original settings.

distributed/scheduler.py
```diff
@@ -3158,6 +3160,12 @@ def coerce_address(self, addr, resolve=True):
 
         return addr
 
+    def clear_ip_cache(self, addr):
+        if isinstance(addr, tuple):
+            addr = unparse_host_port(*addr)
+            clear_address_cache(addr)
```
mrocklin: What should happen otherwise?  Should we raise a NotImplementedError?  The silent behavior here is probably dangerous for future developers.

distributed/utils.py
```diff
@@ -477,31 +477,49 @@ def silence_logging(level, root='distributed'):
...
-    --------
-    >>> ensure_ip('localhost')
-    '127.0.0.1'
-    >>> ensure_ip('123.123.123.123')  # pass through IP addresses
-    '123.123.123.123'
-    """
-    # Prefer IPv4 over IPv6, for compatibility
-    families = [socket.AF_INET, socket.AF_INET6]
-    for fam in families:
+class Ensure_ip(object):
```
mrocklin: I recommend camel case for classes

https://www.python.org/dev/peps/pep-0008/#class-names

mrocklin:
Rewriting /etc/hosts sounds somewhat invasive and would, I think, this would require sudo priveleges.

Are there any other options?

rbubley:
I agree it's invasive and would need sudo privileges, but if that test were configured, say, to only run on Travis, would that be a problem?

distributed/utils.py
rbubley: Yes, but in this case Ensure_ip is a callable class masquerading as a function, and thus keeps us from needing to change the calls to ensure_ip throughout the rest of the project, plus changing it back if the upstream issue at Tools/memoize progresses...

mrocklin: To be explicit, all I'm asking here is renaming the class to `class EnsureIP(object)` to align with PEP8.  Function name consistency should be maintained by the `ensure_ip` instance that you define just below.

rbubley: Doh. Fixed.

distributed/scheduler.py
rbubley: We should largely mirror the logic in coerce_address, as that is how addresses get fed into caches. (But exclude the check for aliases, as (1) they are pre-resolved, and (2) this would miss the underlying cache that would still be there after the alias gets deleted)

mrocklin:
> I agree it's invasive and would need sudo privileges, but if that test were configured, say, to only run on Travis, would that be a problem?

We give up sudo privileges on travis-ci.

distributed/scheduler.py
```diff
@@ -904,13 +904,15 @@ def remove_worker(self, comm=None, address=None, safe=False, close=True):
             del self.worker_comms[address]
             del self.ncores[address]
             self.workers.remove(address)
+            self.clear_ip_cache(self.worker_info[address]['name'])
```
mrocklin: I think that you don't want name here.  I think that you want host.  Name is assigned aliases when they exist.

mrocklin: Can we call `ensure_ip.clear_cache(self.worker_info[address]['host'])` here directly and avoid touching the comms layer and all of the `clear_address_cache` definitions?

distributed/utils.py
```diff
@@ -477,31 +477,54 @@ def silence_logging(level, root='distributed'):
...
+            except socket.gaierror as e:
+                exc = e
+            else:
+                self.ip_cache[hostname] = results[0][4][0]
+                return results[0][4][0]
+
+        raise exc
+
+    def clear_cache(self, hostname):
+        # depending on how clear_cache has been called, hostname could be either a key or a value in the cache
```
pitrou: I don't think that's very useful. Why not simply invalidate based on the key? Besides, looking up based on value is a O(n) operation.

distributed/scheduler.py
pitrou: I think @mrocklin has a point. Also, it would be nice to add a comment here explaining why we're doing this.

pitrou:
As for testing, there's a simpler route: check that the hostname/IP association is in the cache once the worker has registered, and that it's removed from the cache after the worker has unregistered.

rbubley:
Useful comments @pitrou. I will come back to this and revise as per suggestions. Some of the rationale for the way it was done was because the cache can end up with not just hostname:ip entries, but also ip:ip entries, but on reflection there is no especial benefit in clearing out the latter ones. 

It will be a few weeks before I can turn to this, however.

pitrou:
> It will be a few weeks before I can turn to this, however.

No problem! This isn't a high-priority issue so we can take the time to do things right :-)

mrocklin:

PR #1397, opened 2017-09-15
Restart worker if memory use goes above 95%

mrocklin:


mrocklin:
This is a WIP.  We need to add some documentation, and clean up keyword/CLI arguments

PR #1398, opened 2017-09-15
Track worker's memory use

mrocklin:
This augments the `sizeof` spill to disk behavior with psutil tracking.

If we rise above 70% memory use, start dumping data to disk.

If we rise above 80% memory use, stop execution of new tasks

cc @bluenote10

### DIFF
--- a/distributed/diagnostics/profile.py
+++ b/distributed/diagnostics/profile.py
@@ -73,7 +73,7 @@ def process(frame, child, state, stop=None):
      'children': {'...'}}
     """
     ident = identifier(frame)
-    if frame.f_back is not None and (stop is None or stop != frame.f_back.f_code.co_name):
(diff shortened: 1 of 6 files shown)


mrocklin:
This is a WIP.  We need to add some documentation, and clean up keyword/CLI arguments

Also see https://github.com/dask/distributed/pull/1397

martindurant:
Is this the time we should solve the problem that psutil doesn't see the true memory available to a docker container, but the host's total?

mrocklin:
We're measuring the memory use of the process, not of the entire machine.

martindurant:
Unless otherwise specified, worker processes set their memory limits based on the apparent available memory, no?

mrocklin:
Yes.  We do make that guess.  In mature deployments people should specify
numbers explicitly.

martindurant:
Also on diagnostics boards, we report machine usage... well this obviously doesn't belong here.

Issue #1391, opened 2017-09-13
Considering adding support for env variables


barrachri:
Hi, I was playing with distributed and for both dask-worker and dask-scheduler you have to specify an IP:PORT (generally).

Considering the use with docker images and containers I think it would be useful to use also env variables.

Would you consider to add something like:
```
DASK_WORKER_IP=your_ip
DASK_WORKER_PORT=your_port
```

```
DASK_SCHEDULER_IP=your_ip
DASK_SCHEDULER_PORT=your_port
```
to specify them?

mrocklin:
I'm curious, is there precedent here?  Are there other distributed systems
that work in this way?

danielfrg:
Most applications allows config from env variables. For spark you have for example: https://github.com/apache/spark/blob/master/conf/spark-env.sh.template

Its probably more a convenience thing when you have a lot of settings in a lot of nodes.

danielfrg:
This is probably actually already there but maybe its not documented?

https://github.com/dask/distributed/blob/master/distributed/config.py#L66-L70

barrachri:
Cool, I'll check it.

I'll send a PR to improve the docs in case is not documented.

mrocklin:
OK, if this is standard practice then that is a good argument to add it.  We don't currently support it but could fairly easily.  The code that @danielfrg  is pointing to automatically collects all environment variables that start with DASK_ and puts them into the global distributed.config.config dictionary.  We would want to check the contents of that directory and use those variables if they exist in in the startup scripts distributed/cli/dask_worker.py and distributed/cli/dask_scheduler.py .  

@barrachri is this something for which you would be interested in submitting a PR?

barrachri:
Absolutely.

I'll send a PR probably during this weekend.

What do you mean by 
> We would want to check the contents of that *directory*
?

In case you have both env vars and cli parameters which one should be considered?

mrocklin:
Feel free to ignore that comment.

Regarding env and cli parameters, what are your thoughts?

barrachri:
I'd say that generally cli parameters have higher priority.

So if both are specified we will use the cli parameters, in case there is a collision logging a warning should be enough.

hussainsultan:
In my general experience, ENV variables should take precedence over the config file variables. CLI should take precedence over both.

barrachri:
So:
cli > env > conf

PR #1401, opened 2017-09-16
Add env vars

barrachri:
With this PR you can specify for both workers and scheduler an IP:PORT using env vars `DASK_SCHEDULER_IP` and `DASK_SCHEDULER_PORT`.

https://github.com/dask/distributed/issues/1391

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -1,6 +1,7 @@
 from __future__ import print_function, division, absolute_import
 
 import atexit
+from distutils.util import strtobool
 from functools import partial
 import logging
 import os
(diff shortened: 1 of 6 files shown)


distributed/cli/dask_scheduler.py
```diff
@@ -111,6 +112,15 @@ def del_pid_file():
         limit = max(soft, hard // 2)
         resource.setrlimit(resource.RLIMIT_NOFILE, (limit, hard))
 
+    # check if DASK_SCHEDULER_IP and DASK_SCHEDULER_PORT are defined
+    if not host:
+        dict_view = config.keys
+        if PY2:
+            dict_view = config.viewkeys
+        if {'scheduler-ip', 'scheduler-port'} <= dict_view():
```
mrocklin: The above seems more complex than it needs to be.  If I were to write this I would say `if 'scheduler-ip' in config and 'scheduler-port' in config:`

distributed/cli/dask_worker.py
```diff
@@ -191,8 +191,11 @@ def del_pid_file():
         t = Worker
 
     if not scheduler and not scheduler_file:
-        raise ValueError("Need to provide scheduler address like\n"
-                         "dask-worker SCHEDULER_ADDRESS:8786")
+        try:
+            scheduler = ":".join((config['scheduler-ip'], config['scheduler-port']))
+        except KeyError:
+            raise ValueError("Need to provide scheduler address like\n"
+            "dask-worker SCHEDULER_ADDRESS:8786 or with environment variables")
```
mrocklin: If we're going to list environment variables as an option then we probably need to explain them.  In the interests of brevity I recommend just avoiding this change.

distributed/cli/tests/test_dask_worker.py
```diff
@@ -165,3 +166,15 @@ def func(dask_worker):
...
+def test_scheduler_with_env_vars(loop):
+    envs = os.environ.copy()
+    ip = envs['DASK_SCHEDULER_IP'] = '127.0.0.1'
+    port = envs['DASK_SCHEDULER_PORT'] = '8786'
+    address = "tcp://{}:{}".format(ip, port).encode('utf-8')
+    with popen(['dask-scheduler', '--no-bokeh']) as sched:
+        with popen(['dask-worker',
+                    '--no-bokeh'], env=envs) as worker:
+                    assert any(address in worker.stderr.readline()
+                       for i in range(15))
```
mrocklin: The indentation here is odd.

distributed/config.py
```diff
@@ -64,6 +64,12 @@ def load_config_file(config, path):
 
 
 def load_env_vars(config):
+    """
+    Scan environment variables for `DASK_`.
```
mrocklin: See http://dask.pydata.org/en/latest/develop.html#docstrings for style conventions

barrachri:
I was checking Spark and it seems they still do in the [old way](https://github.com/apache/spark/blob/master/conf/spark-env.sh.template).

Did you think about naming the vars in this way?

`DASK_SCHEDULER_LISTEN_ADDRESS`
`DASK_SCHEDULER_CONTACT_ADDRESS`

mrocklin:
Sorry for the delay in response.  In general this looks good to me.  

> Did you think about naming the vars in this way?

>     DASK_SCHEDULER_LISTEN_ADDRESS
>     DASK_SCHEDULER_CONTACT_ADDRESS

Yes.   More generally it seems that some groups prefer to specify keywords with environment variables.  Great, we should support that.  Now how do we go about supporting this generally across all command line arguments.   Presumably we should have a convention like `DASK_SCHEDULER_PORT` -> `--port` or `DASK_PORT` -> `--port`.  Is there a way to do this for all command line arguments that isn't particularly annoying or are we best off with just a brute force approach?  Are there pre-existing conventions for this sort of thing?

barrachri:
About the names for the variables, I prefer something more explicit like `DASK_SCHEDULER_PORT` instead of `DASK_PORT`.

I'll give a deeper look inside click, but I think you can resolve this only in the boring way....
In case, would it be a new PR or you prefer to add new commits to this one?

mrocklin:
The same PR would be fine to me.  Thank you for taking this on.

barrachri:
Sorry for the delay.

I didn't find an easier solution than the one inside the last commit.

With `click` you can do something like this
```python
@click.command()
@click.option('--host', type=str, default='', help="URI, IP or hostname of this server")
@click.pass_context
def cli(ctx, debug):
    print(ctx.params) #dict
```
and you will get a dictionary, but you will miss the specs about types, so you will still have to hardcode when you cast from str to another type.

jakirkham:
Found myself reaching for this functionality recently. Agree it is desirable. ;)

FWIW Click appears to have a way to take [environment variable for options]( http://click.pocoo.org/5/options/#values-from-environment-variables ). Should point out this makes prefixing trivial thanks to `auto_envvar_prefix` and works exactly as imagined. :)

mrocklin:

Issue #1409, opened 2017-09-20
Use RSS or VMS to track memory?


mrocklin:
We recently added mechanisms to track a worker's memory use in order to slow down or terminate that worker to avoid out-of-memory errors.  When tracking memory we can look at either RSS or VMS memory use.

```python
[1]: import psutil

In [2]: psutil.Process().memory_info()
Out[2]: pmem(rss=34463744, vms=289349632, shared=5316608, text=4096, lib=0, data=179593216, dirty=0)
```

Which should we be tracking?  What do job schedulers track?

VMS is a bit unfortunate in that it includes 300MB of Python and 600MB including numpy, pandas, etc. which are sometimes just forked from the main process.

RSS is a bit unfortunate in that it seems not to include data allocated by NumPy.

```python
In [3]: import numpy as np

In [4]: psutil.Process().memory_info()
Out[4]: pmem(rss=43642880, vms=358273024, shared=8069120, text=4096, lib=0, data=186109952, dirty=0)

In [5]: x = np.empty(int(1e9), dtype='u1')

In [6]: psutil.Process().memory_info()
Out[6]: pmem(rss=43642880, vms=1358274560, shared=8069120, text=4096, lib=0, data=1186111488, dirty=0)
```

Given this I think we have to use VMS.  I'm not very familiar with what is commonly tracked by job schedulers that might kill a job though, so I thought I'd put this up here for general comment.

cc @davidedelvento

seibert:
VMS will unfortunately be very misleading with any process that uses CUDA.  When the CUDA driver is initialized, it expands the virtual memory space to allow GPU memory to be mapped into the process's memory space.  Example:
``` python
In [1]: import psutil

In [2]: psutil.Process().memory_info()
Out[2]: pmem(rss=30953472, vms=212426752, shared=8904704, text=4096, lib=0, data=99287040, dirty=0)

In [3]: import numba.cuda

In [4]: c = numba.cuda.get_current_device()  # trigger initialization of CUDA

In [5]: psutil.Process().memory_info()
Out[5]: pmem(rss=160231424, vms=22376509440, shared=111820800, text=4096, lib=0, data=22016393216, dirty=0)
```
The more GPU memory you have, the larger the effect.  I'm not sure how to work around this...

mrocklin:
Are there cases other than CUDA where this occurs.

Do you have suggestions on how best to track memory use at the process level?

martindurant:
[...]

PR #1411, opened 2017-09-20
Use RSS when tracking memory 

mrocklin:
Fixes https://github.com/dask/distributed/issues/1409


TomAugspurger:
Haven't looked at the changes, but this does fix the issue I was seeing with workers being killed at startup.

PR #1414, opened 2017-09-21
Add try-except to scheduler when logging worker errors (Resolves #1396)

### LINKED ISSUE #1396: Consider making Scheduler non-reliant on modules used by client/worker
I currently have a process where the client and the workers all have access to some custom Python modules, but the scheduler runs on a clean box with nothing by distributed and its requirements installed.

This works fine, except when a worker raises an exception in a class defined in the custom module. In this case, I get:

```
No module named custom
  File "/usr/local/lib/python2.7/dist-packages/distributed/scheduler.py", line 1432, in handle_worker
    worker, clean_exception(**msg)[1])
  File "/usr/local/lib/python2.7/dist-packages/distributed/core.py", line 711, in clean_exception
    exception = protocol.pickle.loads(exception)
  File "/usr/local/lib/python2.7/dist-packages/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
ImportError: No module named custom    
```

rbubley:
Worker errors may involve classes loaded from packages which do not exist on the scheduler. Currently, the scheduler tries to parse these exceptions, which may (in an unpickle) try to import a module that doesn't exist on the scheduler.

The alternative would be to send the pre-rendered error from the worker to the scheduler, but as it can be useful for the client to receive the actual Exception object (e.g. for viewing the traceback), this would involve sending both the original Exception and the rendered version.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1426,11 +1426,16 @@ class Scheduler(ServerNode):
                     for msg in msgs:
                         if msg == 'OK':  # from close
                             break
-
                         if 'status' in msg and 'error' in msg['status']:
-                            logger.error("error from worker %s: %s",
+                            try:
+                                logger.error("error from worker %s: %s",
                                          worker, clean_exception(**msg)[1])
-
+                            except ImportError:
+                                # clean_exception may try to unpickle an object that requires a package that does not
+                                # exist on the scheduler but does exist on the worker – this attempt would generate an
+                                # ImportError
+                                logger.error("error from worker %s",
+                                         worker)
                         op = msg.pop('op')
                         if op:
                             self.correct_time_delay(worker, msg)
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -1426,11 +1426,16 @@ def handle_worker(self, worker):
                     for msg in msgs:
                         if msg == 'OK':  # from close
                             break
-
                         if 'status' in msg and 'error' in msg['status']:
-                            logger.error("error from worker %s: %s",
+                            try:
+                                logger.error("error from worker %s: %s",
                                          worker, clean_exception(**msg)[1])
-
```
mrocklin: Why catch `ImportError`'s specifically?  Are there not other cases where deserializing the exception might cause an exception?  In fact, we might be able to test this by creating an exception that can be serialized, but fails on deserialization.

```python
class UnDeserializableException(Exception):
    def __getstate__(self):
        return 
    def __setstate__(self, state):
        raise Exception('Some unique text to search for in a test')
```

rbubley: You are right. Unpickling can generate a variety of Exceptions. I'll broaden it.

PR #1415, opened 2017-09-21
Give Client a name which shows up in scheduler logs.

rbubley:
Add name to Client, to enable scheduler logs to show which client messages are about.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -408,6 +408,9 @@ class Client(Node):
         Optional security information
     asynchronous: bool (False by default)
         Set to True if this client will be used within a Tornado event loop
+    name: string (optional)
+        Gives the client a name that will be included in logs generated on
+        the scheduler for matters relating to this client
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for the submission.  A couple of notes:

1.  This needs tests
2.  What happens if multiple clients choose the same name?  Presumably this should err?

rbubley:
I'll add a test. 

There is no need to err if multiple clients choose the same name – distributed will still work just fine, it's just that the log messages won't be as useful. (As the name defaults to None, this is precisely what will happen if multiple Clients happen not to set the name.)

mrocklin:
If a client leaves then `remove_worker` will be triggered for that client and all clients with that name will no longer receive messages.  I suspect that there will also be other undesired effects in other situations.  Generally the scheduler was written with the assumption that IDs are unique.

rbubley:
This PR keeps the ids unique: the name is concatenated onto the existing unique id - so even if two clients use the same name, the id remains unique.

mrocklin:
Ah I see.  I misread.  Thank you for clarifying.

Are you comfortable adding a small test?

rbubley:
Testing that the Client still works both with and without the new parameter is straightforward. 

Not sure how easy it is to test that the scheduler generates log messages which include the client's name - any suggestions welcome.

mrocklin:
You should first check that the name is actually in the client's id.  

```python
client = yield Client(s.address, asynchronous=True, name='foo')
assert 'foo' in client.id
```

If you care about logging then you might also search the codebase for `captured_logger`

    with captured_logger(logging.getLogger('distributed.scheduler')) as logger:

mrocklin:
Thanks @rbubley .  Merging.

PR #1442, opened 2017-10-04
Deal with restart timeouts

mrocklin:
Noticed that sometimes restarts were hanging in odd ways.  This is a few attempts to address this.  Needs tests and more work.


mrocklin:
cc @adamklein

PR #1450, opened 2017-10-05
Add Client.current

mrocklin:
cc @TomAugspurger @seibert


distributed/client.py
```diff
@@ -520,6 +520,11 @@ def __init__(self, address=None, loop=None, timeout=5,
         from distributed.recreate_exceptions import ReplayExceptionClient
         ReplayExceptionClient(self)
 
+    @classmethod
```
martindurant: Isn't this a static method, as opposed to a class method?

mrocklin: Fixed

PR #1458, opened 2017-10-09
[WIP] Asyncio rewrite

kszucs:
I'm trying to take advantage of `asynchronous` feature of Client to implement AioClient, AioFuture, AioVariable more consistently.  This PR is more or less a prototype.

I've just realized, that most the public facing API of `Future` and `Variable` and `Client` objects are handled through `Client.sync` which implicitly takes care of asynchronous case - an exception is `Future.result`.

Ideally the whole asyncio compatibility could be implemented in `Client.sync` preventing subclassing the `Client` causing dependency injection problems like `client._Future` I've previously introduced. Also with subclassed `AioVariable` and `AioFuture` unpickling would be really cumbersome.

@mrocklin thoughts?

related to #1186


distributed/client.py
```diff
@@ -155,19 +155,7 @@ def result(self, timeout=None):
...
-                result = gen.with_timeout(timedelta(seconds=timeout), result)
-            return result
-        result = sync(self.client.loop,
-                      self._result, raiseit=False, callback_timeout=timeout)
-        if self.status == 'error':
-            six.reraise(*result)
-        elif self.status == 'cancelled':
-            raise result
-        else:
-            return result
```
mrocklin: We wrote this in the more verbose way for a reason, though I can't today remember why.  All the tests pass and I'm unable to reproduce any negative behavior with this change so presumably whatever the reason was no longer applies :)

mrocklin:
In general this seems very nice.  Thank you for working on this @kszucs .  What are your thoughts?

distributed/client.py
mrocklin: Ah, in the case of an error the traceback that we give to users is much much smaller with the implementation in master.

```python
In [5]: future.result()
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-5-a3ab6647db03> in <module>()
----> 1 future.result()

/home/mrocklin/workspace/distributed/distributed/client.py in result(self, timeout)
    164                       self._result, raiseit=False, callback_timeout=timeout)
    165         if self.status == 'error':
--> 166             six.reraise(*result)
    167         elif self.status == 'cancelled':
    168             raise result

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    683             value = tp()
    684         if value.__traceback__ is not tb:
--> 685             raise value.with_traceback(tb)
    686         raise value
    687 

<ipython-input-1-398a43a7781e> in div()
      1 def div(x, y):
----> 2     return x / y

```

This branch

```python
In [19]: future.result()
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-19-a3ab6647db03> in <module>()
----> 1 future.result()

/home/mrocklin/workspace/distributed/distributed/client.py in result(self, timeout)
    156         is raised.
    157         """
--> 158         return self.client.sync(self._result, callback_timeout=timeout)
    159 
    160     @gen.coroutine

/home/mrocklin/workspace/distributed/distributed/client.py in sync(self, func, *args, **kwargs)
    545             return future
    546         else:
--> 547             return sync(self.loop, func, *args, **kwargs)
    548 
    549     def __str__(self):

/home/mrocklin/workspace/distributed/distributed/utils.py in sync(loop, func, *args, **kwargs)
    239         e.wait(1000000)
    240     if error[0]:
--> 241         six.reraise(*error[0])
    242     else:
    243         return result[0]

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    684         if value.__traceback__ is not tb:
    685             raise value.with_traceback(tb)
--> 686         raise value
    687 
    688 else:

/home/mrocklin/workspace/distributed/distributed/utils.py in f()
    227             yield gen.moment
    228             thread_state.asynchronous = True
--> 229             result[0] = yield make_coro()
    230         except Exception as exc:
    231             logger.exception(exc)

/home/mrocklin/workspace/tornado/tornado/gen.py in run(self)
   1053 
   1054                     try:
-> 1055                         value = future.result()
   1056                     except Exception:
   1057                         self.had_exception = True

/home/mrocklin/workspace/tornado/tornado/concurrent.py in result(self, timeout)
    236         if self._exc_info is not None:
    237             try:
--> 238                 raise_exc_info(self._exc_info)
    239             finally:
    240                 self = None

/home/mrocklin/workspace/tornado/tornado/util.py in raise_exc_info(exc_info)

/home/mrocklin/workspace/tornado/tornado/gen.py in run(self)
   1067                             exc_info = None
   1068                     else:
-> 1069                         yielded = self.gen.send(value)
   1070 
   1071                     if stack_context._state.contexts is not orig_stack_contexts:

/home/mrocklin/workspace/distributed/distributed/client.py in _result(self, raiseit)
    165                                   self._state.traceback)
    166             if raiseit:
--> 167                 six.reraise(*exc)
    168             else:
    169                 raise gen.Return(exc)

/home/mrocklin/Software/anaconda/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    683             value = tp()
    684         if value.__traceback__ is not tb:
--> 685             raise value.with_traceback(tb)
    686         raise value
    687 

<ipython-input-1-398a43a7781e> in div()
      1 def div(x, y):
----> 2     return x / y

```

mrocklin: I've pushed a test to your branch that has a test for this.  It should trigger a failure here.  I *think* that the right thing is to go back to the previous code, though there might also be some more clever solution.

kszucs:
Thanks! I'm a bit concerned [about `closing()` an async client](https://github.com/dask/distributed/blob/master/distributed/client.py#L953). 
What will happen with `self.cluster` and `self.loop`?

mrocklin:
It looks like we close the cluster explicitly: https://github.com/dask/distributed/blob/master/distributed/client.py#L927

kszucs:
Ohh [this](https://github.com/dask/distributed/blob/master/distributed/client.py#L965) has misled me.

kszucs:
@mrocklin If I fallback to the previous implementation I'd need to manually check whether that future is an asyncio or tornado one originating from a tornado or an asyncio client.

Could You suggest another workaround for limiting the traceback?

mrocklin:
What was wrong with your intermediate solution?

kszucs:
A [lot](https://travis-ci.org/dask/distributed/jobs/285604738), I handled both the async and sync cases the same way.

kszucs:
@mrocklin I've managed to solve it.

mrocklin:
This looks good to me.  Ready to merge?

kszucs:
If You say so :)

PR #1462, opened 2017-10-11
Add rejoin function partner to secede

### LINKED ISSUE #1459: Add rejoin operation to rejoin thread pool
Currently the `secede` function allows a task to remove itself from the worker's current thread pool, opening up a space for more tasks.

We might consider an inverse operation, `rejoin` that blocks until a new spot in the thread pool has opened up.  This would enable long-running task computations to avoid contention of many threads computing at once.

First suggested by @adamklein in https://github.com/dask/distributed/issues/1342

also cc @ogrisel

mrocklin:
This blocks until a thread can rejoin the thread pool

Fixes #1459 

This is a work in progress.  I'm not yet certain that this is always safe.

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -10,6 +10,7 @@ from .client import (Client, Executor, CompatibleExecutor,
 from .nanny import Nanny
 from .queues import Queue
 from .scheduler import Scheduler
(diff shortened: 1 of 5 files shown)


mrocklin:
Merging this soon if there are no further comments.

Issue #1460, opened 2017-10-10
Default behavior when calling compute within a task


mrocklin:
We might consider the following changes when code calls `dask.compute` from within a task:

1.  Automatically connect to a worker-client and reuse the known cluster
2.  Automatically secede from the thread pool to allow more tasks (possibly our own) to use those threads
3.  Automatically rejoin the thread pool on completion #1459 

cc @ogrisel

PR #1465, opened 2017-10-11
Nested compute

mrocklin:
This makes nested compute calls reuse the scheduler cleanly.  

cc @ogrisel 

Fixes https://github.com/dask/distributed/issues/1460

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -458,7 +458,7 @@ class Client(Node):
         self.coroutines = []
         self.id = type(self).__name__ + ('-' + name + '-' if name else '-') + str(uuid.uuid1(clock_seq=os.getpid()))
         self.generation = 0
-        self.status = None
+        self.status = 'newly-created'
         self._pending_msg_buffer = []
         self.extensions = {}
         self.scheduler_file = scheduler_file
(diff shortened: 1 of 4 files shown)


mrocklin:
https://github.com/dask/dask/pull/2762

ogrisel:
It seems to work fine although the tasks view of the web UI is a bit confusing. Here is an example output on a cluster with 5 worker threads. The client submit 5 top level tasks that spawn 3 nested tasks each:

image

Here is the code:

```python
import dask
import time


c = Client('localhost:8786')


def slow_return(x):
    time.sleep(0.1)
    return x


def slow_sum(stuff):
    return sum(dask.compute(dask.delayed(slow_return)(x) for x in stuff)[0])


dask.compute(dask.delayed(slow_sum)(range(3)) for _ in range(5))
```

It's hard to tell which task was run on which worker from the layout on the y axis.

mrocklin:
Yes, the lines are indexed by worker ID and thread ID.  Now whenever you call dask.compute that thread leaves the thread pool for a while and a new thread takes its place.  The web UI will look a bit weird, but I'm not sure what we can do about this.

ogrisel:
Maybe it could be ordred by thread names (optionally + thread ids to enforce deterministic ordering in case of duplicate names) could use the seceded thread names as prefix for the newly created threads that are spawn upon secession?

mrocklin:
Two issues:

1.  We might then have to change the y-coordinates of existing rectangles in the task stream, which is doable but slightly inefficient when there are many rectangles (such as in the `/tasks` page)  (this can take several seconds)
2.  Most of the tasks created by the call to `compute` will go onto existing threads.  We will spawn at most one new thread which will likely go away soon.  It might be nicer to keep it at the bottom of the plot.

ogrisel:
Alright, it's not a big deal and should not impact many workloads anyway.

PR #1476, opened 2017-10-17
Reschedule

mrocklin:
This enables tasks to ask to be rescheduled by the scheduler by raising a special `Reschedule` exception.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -752,6 +752,7 @@ def apply_function(function, args, kwargs, execution_state, key,
     except Exception as e:
         msg = error_message(e)
         msg['op'] = 'task-erred'
+        msg['actual-exception'] = e
     else:
         msg = {'op': 'task-finished',
                'status': 'OK',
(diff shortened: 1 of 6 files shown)


distributed/worker.py
```diff
@@ -2592,3 +2600,18 @@ def secede():
...
+
+    Raising this exception will stop the current execution of the task and ask
+    the scheduler to reschedule this task, possibly on a different machine.
+
+    This does not guarantee that the task will move onto a different machine.
+    The scheduler will proceed through its normal heuristics to determine the
+    optimal machine to accept this task.  The machine will likely change if the
+    load across the cluster has significantly changed since first scheduling
+    the task.
+    """
```
TomAugspurger: How does this interact with the scheduler's tracking of suspicious tasks? I assume a task that raises a Reschedule still counts as a failure, but it'd be good to document that here.

mrocklin: It does not count as a failure.  A task that raises Reschedule will not be considered more suspicious.

Issue #1447, opened 2017-10-04
Website to navigate internal scheduler information about tasks


mrocklin:
When things break in complex situations advanced users often resort to querying scheduler and worker state, often using code like the following:

```python
client.run_on_scheduler(lambda dask_scheduler: dask_scheduler.processing(worker))
client.run_on_scheduler(lambda dask_scheduler: dask_scheduler.transition_story(key))
client.run_on_scheduler(lambda dask_scheduler: dask_scheduler.transition_story(other_key))
...
```

We can probably provide better navigation here by providing a static HTML interface with many hyperlinks to allow users to jump around.  This might help to reduce frustration in complex situations.

PR #1478, opened 2017-10-18
Add scheduler HTML routes to bokeh server

mrocklin:
This adds a new set of routes to the HTTP server that are focused on exposing internal operation of the scheduler.  These generated templated HTML.  They do not update like the Bokeh plots but they are easy to build information dense pages cheaply and they are highly networked.

This needs additional exposure to more scheduler state
This needs some CSS love


mrocklin:
Fixes https://github.com/dask/distributed/issues/1447

mrocklin:
I'm testing this out by running simple code like the following:
```python
client = Client(processes=True)
import dask.array as da
x = da.random.random((10000, 10000), chunks=(1000, 1000)).persist()
y = x + x.T
y = y.persist()
```

And then navigating to http://localhost:8786/scheduler/workers.html

mrocklin:
image

mrocklin:
Updated.  I dumped the bulma CSS framework in there and messed around with it a bit.

image

image

PR #1482, opened 2017-10-19
Capture logs

mrocklin:
We now track the last 10000 logged records for both the scheduler and
worker modules.  These are accessible through normal routes on the
workers, scheduler, and client.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -28,7 +28,7 @@ from .batched import BatchedSend
 from .comm import (normalize_address, resolve_address,
                    get_address_host, unparse_host_port)
 from .compatibility import finalize
-from .config import config
+from .config import config, log_format
 from .core import (rpc, connect, Server, send_recv,
                    error_message, clean_exception, CommClosedError)
 from . import profile
@@ -36,7 +36,7 @@ from .metrics import time
 from .node import ServerNode
 from .security import Security
 from .utils import (All, ignoring, get_ip, get_fileno_limit, log_errors,
-                    key_split, validate_key, no_default)
+                    key_split, validate_key, no_default, DequeHandler)
 from .utils_comm import (scatter_to_workers, gather_from_workers)
 from .versions import get_versions
 
@@ -49,6 +49,10 @@ from .variable import VariableExtension
 
 
 logger = logging.getLogger(__name__)
+deque_handler = DequeHandler(n=config.get('log-length', 10000))
+deque_handler.setFormatter(logging.Formatter(log_format))
+logger.addHandler(deque_handler)
+
 
 BANDWIDTH = config.get('bandwidth', 100e6)
 ALLOWED_FAILURES = config.get('allowed-failures', 3)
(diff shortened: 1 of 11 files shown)


distributed/utils.py
```diff
@@ -986,3 +986,14 @@ def format_time(n):
...
+
+class DequeHandler(logging.Handler):
+    """ A logging.Handler that records records into a deque """
+    def __init__(self, *args, **kwargs):
+        n = kwargs.pop('n', 10000)
+        self.deque = deque(maxlen=n)
+        super(DequeHandler, self).__init__(*args, **kwargs)
+
+    def emit(self, record):
+        self.deque.append(record)
```
mrocklin: @pitrou do you know of a cleaner or more standard way to do this?

distributed/scheduler.py
```diff
@@ -3334,6 +3340,20 @@ def get_profile_metadata(self, comm=None, workers=None, merge_workers=True,
 
         raise gen.Return({'counts': counts, 'keys': keys})
 
+    def get_logs(self, comm=None, n=None):
+        if n is None:
+            L = list(deque_handler.deque)
+        else:
+            L = deque_handler.deque
+            L = [L[-i] for i in range(min(n, len(L)))]
```
pitrou: Not sure, but `list(deque_handler.deque)[-n:]` looks simpler (and potentially faster) to me.

distributed/utils.py
pitrou: Perhaps you want to subclass [QueueHandler](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.QueueHandler), or copy and adapt its implementation (it's quite simple). It allows to return structured yet picklable information.

mrocklin:
Also do you have any thoughts on why tests might be failing here?  These work well locally.

pitrou:
No idea, sorry :-/

distributed/scheduler.py
mrocklin: It's only faster above 1000 elements,.  I suspect that most entries for n will be fairly small (otherwise people will just use `n=None`)

```python
In [1]: from collections import deque

In [2]: d = deque(range(10000))

In [3]: %timeit [d[-i] for i in range(100)]
100000 loops, best of 3: 8.78 µs per loop

In [4]: %timeit [d[-i] for i in range(1000)]
The slowest run took 4.67 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 88.7 µs per loop

In [5]: %timeit [d[-i] for i in range(10000)]
1000 loops, best of 3: 1.52 ms per loop

In [6]: %%timeit
   ...: L = list(d)
   ...: [L[-i] for i in range(100)]
   ...: 
The slowest run took 4.87 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 54.9 µs per loop

In [7]: %%timeit
   ...: L = list(d)
   ...: [L[-i] for i in range(1000)]
   ...: 
10000 loops, best of 3: 128 µs per loop

In [8]: %%timeit
   ...: L = list(d)
   ...: [L[-i] for i in range(10000)]
   ...: 
The slowest run took 4.30 times longer than the fastest. This could mean that an intermediate result is being cached.
1000 loops, best of 3: 684 µs per loop
```

PR #1484, opened 2017-10-19
Add default keyword argument to get_metadata

mrocklin:
This allows for semantics similar to `dict.get`

    >>> client.get_metadata('unknown-key', None)
    None

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -36,7 +36,7 @@ from .metrics import time
 from .node import ServerNode
 from .security import Security
 from .utils import (All, ignoring, get_ip, get_fileno_limit, log_errors,
-                    key_split, validate_key)
+                    key_split, validate_key, no_default)
 from .utils_comm import (scatter_to_workers, gather_from_workers)
 from .versions import get_versions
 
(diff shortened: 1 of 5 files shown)


distributed/scheduler.py
```diff
@@ -2175,11 +2175,17 @@ def set_metadata(self, stream=None, keys=None, value=None):
         except Exception as e:
             import pdb; pdb.set_trace()
 
-    def get_metadata(self, stream=None, keys=None):
+    def get_metadata(self, stream=None, keys=None, default=no_default):
         metadata = self.task_metadata
         for key in keys[:-1]:
             metadata = metadata[key]
-        return metadata[keys[-1]]
+        try:
```
adamklein: I don't think it's going to have a material impact, but perhaps it's more efficient to write:

    return metadata[keys[-1]] if default == no_default else metadata.get(keys[-1], default)

mrocklin: I slightly prefer the try-except approach.

If your goal is to avoid exceptions in the scheduler I don't yet think that this is likely to be an issue, at least not if it's entirely contained within a function

```python
In [1]: d = {}

In [2]: %%timeit 
   ...: try:
   ...:     d['x']
   ...: except KeyError:
   ...:     pass
   ...: 
The slowest run took 15.78 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 316 ns per loop
```

PR #1485, opened 2017-10-19
Accelerate tests

mrocklin:
This avoids some slow-downs in our testing machinery


pitrou:
Are the AppVeyor issues new here?

PR #1488, opened 2017-10-20
Aggressively GC on high memory usage in worker

ogrisel:
This is a possible fix for some of the spurious worker restarts reported in the discussion of dask/dask#2470.

I ran the test script of https://github.com/dask/dask/issues/2470#issuecomment-338117445 with additional logs before and after the call to `gc.collect()` and I confirm that many unnecessary restarts can be avoided this way when memory is tight.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -22,6 +22,7 @@ import tblib.pickling_support
 import tempfile
 import threading
 import warnings
+import gc
 
 from .compatibility import cache_from_source, getargspec, invalidate_caches, reload
 
@@ -44,7 +45,7 @@ try:
 except ImportError:
     thread_state = threading.local()
 
-logger = logging.getLogger(__name__)
+logger = _logger = logging.getLogger(__name__)
 
 
 no_default = '__no_default__'
@@ -1000,3 +1001,50 @@ class DequeHandler(logging.Handler):
 
     def emit(self, record):
         self.deque.append(record)
+
+
+class ThrottledGC(object):
+    """Wrap gc.collect to protect against excessively repeated calls.
+
+    Allows to run throttled garbage collection in the workers as a
+    countermeasure to e.g.: https://github.com/dask/zict/issues/19
+
+    collect() does nothing when repeated calls are so costly and so frequent
+    that the thread would spend more than max_in_gc_frac doing GC.
+
+    warn_if_longer is a duration in seconds (10s by default) that can be used
+    to log a warning level message whenever an actual call to gc.collect()
+    lasts too long.
+    """
+    def __init__(self, max_in_gc_frac=0.05, warn_if_longer=10, logger=None):
+        self.max_in_gc_frac = max_in_gc_frac
+        self.warn_if_longer = 10
+        self.last_collect = time()
+        self.last_gc_duration = 0
+        self.logger = logger if logger is not None else _logger
+
+    def collect(self):
+        # In case of non-monotonicity in the clock, assume that any Python
+        # operation lasts at least 1e-6 second.
+        collect_start = time()
+        elapsed = max(collect_start - self.last_collect, 1e-6)
+        if self.last_gc_duration / elapsed < self.max_in_gc_frac:
+            self.logger.debug("Calling gc.collect(). %0.3fs elapsed since "
+                              "previous call.", elapsed)
+            gc.collect()
+            self.last_collect = collect_start
+            self.last_gc_duration = max(time() - collect_start, 1e-6)
+            if self.last_gc_duration > self.warn_if_longer:
+                self.logger.warning("gc.collect() took %0.3fs. This is usually"
+                                    " a sign that the some tasks handle too"
+                                    " many Python objects at the same time."
+                                    " Rechunking the work into smaller tasks"
+                                    " might help.",
+                                    self.last_gc_duration)
+            else:
+                self.logger.debug("gc.collect() took %0.3fs",
+                                  self.last_gc_duration)
+        else:
+            self.logger.debug("gc.collect() lasts %0.3fs but only %0.3fs "
+                              "elapsed since last call: throttling.",
+                              self.last_gc_duration, elapsed)
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -44,7 +44,7 @@ from .threadpoolexecutor import ThreadPoolExecutor, secede as tpe_secede
 from .utils import (funcname, get_ip, has_arg, _maybe_complex, log_errors,
                     ignoring, validate_key, mp_context, import_file,
                     silence_logging, thread_state, json_load_robust, key_split,
-                    format_bytes, DequeHandler)
+                    format_bytes, DequeHandler, ThrottledGC)
 from .utils_comm import pack_data, gather_from_workers
 
 _ncores = mp_context.cpu_count()
@@ -1089,6 +1089,7 @@ class Worker(WorkerBase):
 
         self.log = deque(maxlen=100000)
         self.validate = kwargs.pop('validate', False)
+        self.gc = ThrottledGC(logger=logger)
 
         self._transitions = {
             ('waiting', 'ready'): self.transition_waiting_ready,
@@ -2172,21 +2173,22 @@ class Worker(WorkerBase):
             return
         self._memory_monitoring = True
         total = 0
+
         proc = psutil.Process()
-        frac = proc.memory_info().rss / self.memory_limit
+        memory = proc.memory_info().rss
+        frac = memory / self.memory_limit
 
         # Pause worker threads if above 80% memory use
         if self.memory_pause_fraction and frac > self.memory_pause_fraction:
             if not self.paused:
-                logger.warn("Worker is at %d percent memory usage.  "
-                            "Stopping work. "
+                logger.warn("Worker is at %d%% memory usage. Pausing worker.  "
                             "Process memory: %s -- Worker memory limit: %s",
                             int(frac * 100),
                             format_bytes(proc.memory_info().rss),
                             format_bytes(self.memory_limit))
                 self.paused = True
         elif self.paused:
-            logger.warn("Worker at %d percent memory usage. Restarting work. "
+            logger.warn("Worker is at %d%% memory usage. Resuming worker. "
                         "Process memory: %s -- Worker memory limit: %s",
                         int(frac * 100),
                         format_bytes(proc.memory_info().rss),
@@ -2198,8 +2200,8 @@ class Worker(WorkerBase):
         if self.memory_spill_fraction and frac > self.memory_spill_fraction:
             target = self.memory_limit * self.memory_target_fraction
             count = 0
-
-            while proc.memory_info().rss > target:
+            need = memory - target
+            while memory > target:
                 if not self.data.fast:
                     logger.warn("Memory use is high but worker has no data "
                                 "to store to disk.  Perhaps some other process "
@@ -2209,9 +2211,17 @@ class Worker(WorkerBase):
                                 format_bytes(self.memory_limit))
                     break
                 k, v, weight = self.data.fast.evict()
+                del k, v
                 total += weight
                 count += 1
                 yield gen.moment
+                memory = proc.memory_info().rss
+                if total > need and memory > target:
+                    # Issue a GC to ensure that the evicted data is actually
+                    # freed from memory and taken into account by the monitor
+                    # before trying to evict even more data.
+                    self.gc.collect()
+                    memory = proc.memory_info().rss
             if count:
                 logger.debug("Moved %d pieces of data data and %s to disk",
                              count, format_bytes(total))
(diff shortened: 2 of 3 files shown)


ogrisel:
I don't think there is an easy way to add a test for this.

distributed/worker.py
```diff
@@ -2174,20 +2175,25 @@ def memory_monitor(self):
         total = 0
         proc = psutil.Process()
         frac = proc.memory_info().rss / self.memory_limit
+        if frac > min(self.memory_pause_fraction, self.memory_spill_fraction):
+            # Force a GC event so as to get more accurate memory usage
+            # measurements in case GC is lagging.
+            gc.collect()
```
pitrou: How often is `memory_monitor` executed? `gc.collect` can become quite costly if there are many live objects.

ogrisel:
I broken the tests by editing the other warning messages. I will push a fix.

distributed/worker.py
ogrisel: `gc.collect` is only executed when we are above the high memory fraction. This should not incur any significant performance impact when the memory usage is normal (lower than 70% of the limit by default I think).

pitrou: But imagine a worker that's permanently above that fraction. Now you're running the GC over and again (even if it doesn't release any memory)...

ogrisel: Here is the kind of log I got when running the linked snippet with the 1GB memory limit on the worker with additional logs before and after GC:

https://gist.github.com/ogrisel/6cb75c91d515601c4a659675506ea9ff

I had the code running for 10min or so. So it would not get triggered too often but saved a spurious restart.

ogrisel: > But imagine a worker that's permanently above that fraction. Now you're running the GC over and again (even if it doesn't release any memory)...

The default `memory_monitor_interval` is 200ms according to the default value of the kwarg. I could add a 5s cool-down period between two consecutive gc calls but that will make the code more complex / hackish.

pitrou: Well, It wouldn't be extraordinary for `gc.collect` alone to take 200 ms. Running it on such a short period doesn't sound like a good idea.

Also, I wonder why `gc.collect` is supposed to fix your memory issue at all. Do you have cyclic references somewhere?

mrocklin:
FYI I'm also tracking down some intermittent failures that have been popping up recently.  I haven't seen these before though.

mrocklin:
See also https://github.com/dask/distributed/pull/1255 

cc @bluenote10

distributed/worker.py
ogrisel: It seems to reduce the variance of the measurement of the memory usage of the worker and avoid triggering a worker restart when it's not necessary. If the worker get restarted, all the intermediate tasks are canceled and restarted from scratch which can be very expensive and can even cause the whole "macro-task" to never complete.

mrocklin: We should at least call `gc.collect()` when first transitioning to a paused state.  This occurs below after the `if not self.paused` statement.  I'm curious if this suffices to avoid your problem.

mrocklin: We may also want to call gc.collect() regularly when memory is use is high, I agree with @pitrou though that we should probably control this a little bit.  @bluenote10 's solution in https://github.com/dask/distributed/pull/1255 might be worth looking into.

ogrisel:
The failure on travis in `test_pause_executor` that fails to be paused is probably caused by my change that causes some sort of race condition in this test.

distributed/worker.py
ogrisel: Calling `gc.collect()` in the paused state is apparently not enough. nanny has still a high probability of reaching it's 95% threshold without the worker having the time to see the internal monitory 80% paused threshold first.

I have tried to make nanny less trigger happy by adding a patience counter for consecutive violations but then the nanny leak test tend to fail with my computer close to freezing for a couple of seconds: the nanny OOM killer became too slow to detect the memory leak.

mrocklin: We might call `gc.collect()` *if* we haven't called it fairly recently?  Taking the implementation from #1255 :

```python
class ThrottledGC(object):
    """
    Allows to run throttled garbage collection as a countermeasure to e.g.:
    https://github.com/dask/zict/issues/19
    """
    def __init__(self, min_interval_in_sec=1.0):
        self.min_interval_in_sec = min_interval_in_sec
        self.last_collect = time()
  
    def collect(self, force_gc=False):
        new_time = time()
        if force_gc or new_time - self.last_collect > self.min_interval_in_sec:
            gc.collect()
            self.last_collect = new_time
```

mrocklin:
I believe that you can safely ignore Python 2 nanny test failures: https://github.com/dask/distributed/issues/1491

distributed/worker.py
ogrisel: I have a tried a less aggressive / smarter strategy with extensive logs here: f3981a53a81af4ddfef97f8ed58a9cd61e3dd76c . It kind of works but I still get a random OOM after a while.

mrocklin:
By "nanny test failures" I mean any test in `test_nanny.py` or any test that starts with `@gen_cluster(..., Worker=Nanny)`

distributed/worker.py
ogrisel: At least that last option gets rid of the following erroneous: warning I get most often when running the original snippet on master: "Memory use is high but worker has no data to store to disk.  Perhaps some other process leaking memory?"

mrocklin: Nanny OOMs aren't entirely avoidable if your tasks can quickly generate a moderate fraction of your available memory.  It will always be possible that they can conspire to generate enough data to send you over very quickly.

There are costs and benefits to both approaches.

ogrisel: I will implement the `ThrottledGC` in this PR. But I am not sure if should better implement the strategy of this PR or the one in f3981a5.

mrocklin: I don't have any thoughts on which strategy if better.  If you think what you have in https://github.com/dask/distributed/commit/f3981a53a81af4ddfef97f8ed58a9cd61e3dd76c is nicer then go with that.

ogrisel:
I have updated this PR to do less aggressive GC in the worker (only when it has a chance to be beneficial) while also implementing a throttling strategy based on the fraction of the time in GC over the time between 2 consecutive calls.

I run a bunch of long-running tasks that process chunks with a low memory-limit and it seems to work as expected. The frequency of actual GC events dependes on the value of the memory-limit but I no longer get the misleading "Perhaps some other process is leaking memory?" warning. I get the "Pausing" / "Resuming" warnings from time to time as expected. I have added debug logs to make it possible to use grep to check that the gc is actually called from time to time (e.g. every couple of seconds in my scripts, while each gc duration is on the order of 100ms).

distributed/utils.py
```diff
@@ -1000,3 +1001,42 @@ def __init__(self, *args, **kwargs):
...
+class ThrottledGC(object):
+    """Wrap gc.collect to protect against excessively repeated calls.
+
+    Allows to run throttled garbage collection in the workers as a
+    countermeasure to e.g.: https://github.com/dask/zict/issues/19
+
+    throttled_gc.collect() does nothing when repeated calls are so costly and
+    so frequent that the thread would spend more than max_in_gc_frac doing GC.
+    """
+    def __init__(self, max_in_gc_frac=0.2, warn_if_longer=10, logger=None):
```
mrocklin: Do you have any thoughts on the default value for `max_in_gc_frac=0.2`?  Spending 20% of our time in garbage collection seems unpleasant.

mrocklin: The `warn_if_longer=` parameter could also use some mention in the docstring

ogrisel: >  Spending 20% of our time in garbage collection seems unpleasant.

It's an upper-bound. In most situations memory-limit is high enough that GC will almost never be called at all.

But in cases where the worker is under heavy memory pressure, I find it better to spend that time in the GC than run the risk to be spuriously killed by nanny for no valid reason. Note that it's not a smoothed average, just based on the last 2 calls to collect.

ogrisel: I will try to change the fraction to 5% and run some experiments.

ogrisel: In the meantime, I have pushed an improvement to the docstring as requested. Let me know what you think.

mrocklin: > It's an upper-bound. In most situations memory-limit is high enough that GC will almost never be called at all.

OK, that makes sense to me. 

> In the meantime, I have pushed an improvement to the docstring as requested. Let me know what you think.

If we're spending 10s in gc then that is probably also something that we should know about.  Thoughts on reducing this to 1s?

distributed/worker.py
```diff
@@ -2209,9 +2214,17 @@ def memory_monitor(self):
...
                     break
                 k, v, weight = self.data.fast.evict()
+                del k, v
                 total += weight
                 count += 1
                 yield gen.moment
+                memory = proc.memory_info().rss
+                if total > need and memory > target:
+                    # Issue a GC to ensure that the evicted data is actually
+                    # freed from memory and taken into account by the monitor
```
pitrou: Is it expected for evicted data to have cyclic references?

distributed/utils.py
```diff
@@ -1000,3 +1001,49 @@ def __init__(self, *args, **kwargs):
...
+        collect_start = time()
+        elapsed = max(collect_start - self.last_collect, 1e-6)
+        if self.last_gc_duration / elapsed < self.max_in_gc_frac:
+            self.logger.debug("Calling gc.collect(). %0.3fs elapsed since "
+                              "previous call.", elapsed)
+            gc.collect()
+            self.last_collect = collect_start
+            self.last_gc_duration = max(time() - collect_start, 1e-6)
+            if self.last_gc_duration > self.warn_if_longer:
+                self.logger.warning("gc.collect() took %0.3fs. This is usually"
```
pitrou: Hmm, I'm curious why you think this is useful. Do you know a use case where that warning would have helped?

distributed/utils.py
```diff
@@ -1000,3 +1001,49 @@ def __init__(self, *args, **kwargs):
...
+    countermeasure to e.g.: https://github.com/dask/zict/issues/19
+
+    collect() does nothing when repeated calls are so costly and so frequent
+    that the thread would spend more than max_in_gc_frac doing GC.
+
+    warn_if_longer is a duration in seconds (10s by default) that can be used
+    to control to log a warning level message whenever an actual call to
+    gc.collect() lasts too long.
+    """
+    def __init__(self, max_in_gc_frac=0.2, warn_if_longer=10, logger=None):
```
pitrou: Nitpicking, but allowing 20% of the time inside the GC sounds like a lot.

distributed/utils.py
ogrisel: To warn datascientists that are inadvertently processing a list of 100 million small dicts at a time in each task instead of processing more smaller lists of a couple of million each instead. For instance parsing large JSON files in large chunks instead of small chunks. 

I assume that large GC pauses might make the worker non-responsive. But if you think this is useless we can get rid of this warning.

distributed/worker.py
ogrisel: I don't think it's going to be the case for typical workloads but users are free to ship whatever data structures they like into `delayed` function calls. Why kind of problems do you anticipate?

distributed/utils.py
ogrisel: Would 5% be a reasonable default?

distributed/utils.py
pitrou: > I assume that large GC pauses might make the worker non-responsive.

It will. But what if using many objects is deliberate and required by the workload? Do we want to warn against any possible programming mistake?

distributed/worker.py
pitrou: I simply think it's bad hygiene to add `gc.collect` calls "just in case" :-)

distributed/utils.py
pitrou: It certainly sounds better...

distributed/utils.py
mrocklin: I think that users would appreciate getting a warning if GC is taking seconds.  They have responded well to similar feedback that we have provided in the past (like long event loop pauses).

distributed/worker.py
ogrisel: For the record, this is the kind of actual case that was observed by dask-distributed users:

https://github.com/dask/zict/issues/19#issuecomment-313669395

ogrisel:
I have reworked this PR to address the comments. I now only trigger throttled GC when it's promising, that is after data eviction.

I decided to know longer call the throttled GC after a paused as it's better to call it after eviction, otherwise the throttling mechanism might prevent interesting collections.

mrocklin:
Merged.  

At some point I might want to look at also calling `self.gc.collect` when we transition into a paused state.  This requires some care to avoid the throttling mechanism.  We might get around this by just calling the normal `gc.collect()` here, which might be ok given that this should occur infrequently.

PR #1489, opened 2017-10-20
Avoid double-scheduling in work stealing

mrocklin:
This avoids double-scheduling tasks when work stealing by first checking
with the victim worker to verify that it has not yet started execution.
If it has started then we reverse the steal operation, if it has not
started then we cleanly start the new task only after the victim worker
has removed the task.

This passes most tests (those tests within stealing.py that if fails are
reasonable failures) but typically this sort of change triggers some
knock-on effects.

### DIFF
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -192,12 +192,14 @@ class WorkStealing(SchedulerPlugin):
             raise
 
     def move_task_confirm(self, key=None, worker=None, state=None):
-        if self.scheduler.task_state.get(key) != 'processing':
-            return
         try:
             d = self.in_flight.pop(key)
             thief = d['thief']
             victim = d['victim']
+            if self.scheduler.task_state.get(key) != 'processing':
+                self.scheduler.occupancy[thief] = sum(self.scheduler.processing[thief].values())
+                self.scheduler.occupancy[victim] = sum(self.scheduler.processing[victim].values())
+                return
 
             # One of the pair has left, punt and reschedule
             if (thief not in self.scheduler.workers or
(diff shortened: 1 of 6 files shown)


distributed/stealing.py
```diff
@@ -186,6 +191,54 @@ def move_task(self, key, victim, thief):
...
+                self.scheduler.check_idle_saturated(thief)
+                self.scheduler.check_idle_saturated(victim)
+
+            # Victim was waiting, has given up task, enact steal
+            elif state in ('waiting', 'ready'):
+                self.scheduler.rprocessing[key] = thief
+                del self.scheduler.processing[victim][key]
+                self.scheduler.processing[thief][key] = d['thief_duration']
+                self.remove_key_from_stealable(key)
+                self.put_key_in_stealable(key)
```
adamklein: Do you really mean to `remove_key_from_stealable` and then immediately `put_key_in_stealable`?

adamklein: Empirically seems to make no difference, but the logic escapes me

distributed/worker.py
```diff
@@ -1197,6 +1197,10 @@ def on_closed():
                         self.release_key(report=False, **msg)
                     elif op == 'delete-data':
                         self.delete_data(**msg)
+                    elif op == 'steal-request':
+                        self.steal_request(**msg)
+                    elif op == 'steal-confirm':
+                        self.steal_confirm(**msg)
```
adamklein: Not implemented ... not necessary?

distributed/worker.py
```diff
@@ -1902,6 +1907,17 @@ def update_who_has(self, who_has):
...
+    def steal_request(self, key):
+        state = self.task_state[key]
+
+        response = {'op': 'steal-response',
+                    'key': key,
+                    'state': state}
+        self.batched_stream.send(response)
+
+        if state in ('ready', 'waiting'):
+            self.release_key(key)
```
adamklein: No need for `self.release_key(key, reason='stolen')` ?

PR #1494, opened 2017-10-23
Make nanny tell worker to GC itself on high memory usage

ogrisel:
This PR is built on top of #1488.

#1488 should make worker free memory on a regular basis when reaching the 70% limit using an unthrottled GC and without any warning in the logs.

However, sometimes the nanny monitor might detect the 95% limit before the worker gets the opportunity to use its throttled GC. In this case, the nanny would restart the worker directly.

Instead, this PR sends a USR1 signal (POSIX only) to make the worker process do a non-throttled GC. If the memory usage of the worker has not decreased in the following 200ms, then the worker is restarted.

On Windows, this PR does nothing.

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -75,6 +75,7 @@ class Nanny(ServerNode):
         self.name = name
         self.quiet = quiet
         self.auto_restart = True
+        self._memory_monitoring = False
 
         if memory_limit == 'auto':
             memory_limit = int(TOTAL_MEMORY * min(1, self.ncores / _ncores))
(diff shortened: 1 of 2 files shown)


ogrisel:
I don't understand why you need the flag in the memory_monitor of the worker: there is no call to `gen.sleep` in that method.

distributed/worker.py
```diff
@@ -2172,61 +2172,63 @@ def memory_monitor(self):
...
-                logger.debug("Moved %d pieces of data data and %s to disk",
-                             count, format_bytes(total))
+                    if total > need and memory > target:
+                        # Issue a GC to ensure that the evicted data is actually
+                        # freed from memory and taken into account by the monitor
+                        # before trying to evict even more data.
+                        self.gc.collect()
+                        memory = proc.memory_info().rss
+                if count:
+                    logger.debug("Moved %d pieces of data data and %s to disk",
```
pitrou: "data data"?

distributed/nanny.py
```diff
@@ -245,15 +248,34 @@ def _():
...
-        frac = memory / self.memory_limit
-        if self.memory_terminate_fraction and frac > self.memory_terminate_fraction:
-            logger.warn("Worker exceeded 95% memory budget.  Restarting")
-            self.process.process.terminate()
+        self._memory_monitoring = True
+        try:
+            memory = psutil.Process(self.process.pid).memory_info().rss
+            frac = memory / self.memory_limit
+            if self.memory_terminate_fraction and frac > self.memory_terminate_fraction:
+                if hasattr(signal, 'SIGUSR1'):
```
pitrou: Might be nice to use a module-level constant for this, e.g. `GC_SIGNAL = getattr(signal, 'SIGUSR1', None)`

distributed/nanny.py
```diff
@@ -245,15 +248,34 @@ def _():
...
+                    # Try to be nice and tell the worker to free some memory.
+                    logger.warn("Worker exceeded 95% memory budget.  Forcing GC")
+                    os.kill(self.process.process.pid, signal.SIGUSR1)
+
+                    # Give some time to the worker to GC.
+                    yield gen.sleep(0.2)
```
pitrou: If the worker is busy and has to spill a lot to disk, I wonder if 200ms is really sufficient? Is it important to react quickly?

distributed/nanny.py
```diff
@@ -245,15 +248,34 @@ def _():
...
-            logger.warn("Worker exceeded 95% memory budget.  Restarting")
-            self.process.process.terminate()
+        self._memory_monitoring = True
+        try:
+            memory = psutil.Process(self.process.pid).memory_info().rss
+            frac = memory / self.memory_limit
+            if self.memory_terminate_fraction and frac > self.memory_terminate_fraction:
+                if hasattr(signal, 'SIGUSR1'):
+                    # Try to be nice and tell the worker to free some memory.
+                    logger.warn("Worker exceeded 95% memory budget.  Forcing GC")
```
pitrou: Rather than "95%", display the actual `frac` value?

mrocklin:
From in-person conversations it seemed like we might be able to find another way to make this work that didn't rely on sending the signal.  I think we briefly discussed just increasing the frequency of the `memory_monitor` check.  I don't yet know how expensive this is likely to be.

pitrou:
Looking up the `memory_info` for an arbitrary process looks reasonably cheap on Linux:
```
>>> %timeit p.memory_info()
21.2 µs ± 1.92 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Note you needn't reinstantiate the `psutil.Process` every time, you can just keep it around.

mrocklin:
Slightly different timings

```python
In [1]: import psutil

In [2]: %timeit proc = psutil.Process()
The slowest run took 21.47 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 22.7 µs per loop

In [3]: proc = psutil.Process()

In [4]: %timeit proc.memory_info()
The slowest run took 9.92 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 16.9 µs per loop
```

But still, if we reduce the worker's callback time to 50ms or so this remains very small. I suspect that Tornado itself is more likely to be an issue (but even then I think things would be fine).

If we instead just increase the worker's memory_monitor frequency then we should probably start to be careful about dumping too many warning messages into the logs.

ogrisel:
Yes I am still not convinced what's the best way to tackle this. We could also monitory the variance of the memory over a longer window (e.g. a couple of seconds of historical memory recordings) and be tolerant if it's only a one time spike above the 95% threshold while otherwise oscillating below the threshold.

pitrou:
What's the overarching goal in this? Kill a rogue task? Protect against memory leaks in 3rd party libraries? Prevent the machine from swapping? Something else?

ogrisel:
I suppose: make it easy for the user to control the resources allocated to dask workers on a machine that is also used for other things and in particular preventing accidental swapping especially when using dask on a local developer machine like a laptop.

ogrisel:
I think the solution proposed in #1527 is better. I suggest to cluse this PR.

PR #1496, opened 2017-10-24
Tiny fix: make use of constructor argument in ThrottledGC

bluenote10:


mrocklin:
This is obviously a good change.  I'm curious, what are your thoughts on what the default value should be here.  10s seems long to me.  Do you see something different in practice?

bluenote10:
Yes you're right, 10s for GC is very unusual. So far I haven't been able to trigger the message, but my experiments also did not involve a large number of small Python objects. I would probably be interested in seeing everything larger 1s or 2s (still slow enough that the logging itself is not a bottleneck).

mrocklin:
Thanks @bluenote10 !  Merged

PR #1498, opened 2017-10-24
Improve documentation for scheduler states

pitrou:
This categorizes the state variable in three groups, adds a table of where tasks appear depending on their state, another table listing worker state changes in each task transition, and improves some wording choices in my POV :-)

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -j4
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
--- a/docs/source/scheduling-state.rst
+++ b/docs/source/scheduling-state.rst
@@ -163,10 +163,12 @@ get updated at each state transition.
 * **processing:** ``{worker: {key: cost}}``:
 
    Keys that are currently allocated to a worker.  This is keyed by worker
-   address and contains the expected cost in seconds of running that task.
+   address and contains the expected cost in seconds of running each task,
+   summing both the task's expected computation time and the expected
+   communication time of its result.
 
-   Multiple tasks can typically be submitted to a worker in advance and the
-   worker will run them eventually, depending on its execution resources
+   Multiple tasks may be submitted to a worker in advance and the worker
+   will run them eventually, depending on its execution resources
    (but see :doc:`work-stealing`).
 
 * **rprocessing:** ``{key: worker}``:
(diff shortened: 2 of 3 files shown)


docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -j4
```
pitrou: This will build docs in parallel.

docs/source/scheduling-state.rst
```diff
@@ -34,30 +34,36 @@ State Variables
...
-      scheduler these are stored serialized, as they were sent from the client,
-      so it looks more like ``{'function': b'\x80\x04\x95\xcb\...', 'args':
-      b'...', }``
+   * ``{'function': inc, 'args': (1,), 'kwargs': {}}``; a dictionary with the
+     function, arguments, and keyword arguments (kwargs).  However in the
+     scheduler these are stored serialized, as they were sent from the client,
+     so it looks more like ``{'function': b'\x80\x04\x95\xcb\...', 'args':
+     b'...', }``
 
-    * ``{'task': (inc, 1)}``: a tuple satisfying the dask graph protocol.  This
```
pitrou: Putting `'task'` there seemed to be a bug.

docs/source/scheduling-state.rst
```diff
@@ -70,153 +76,197 @@ Each of the following is a dictionary keyed by task name (described below):
...
-    precedence.  The second part is determined by the client, and is a way to
-    prioritize tasks within a large graph that may be important, such as if
-    they are on the critical path, or good to run in order to release many
-    dependencies.  This is explained further in :doc:`Scheduling Policy
-    <scheduling-policies>`
+   These dictionaries are a subset of ``dependencies`` and ``dependents``
+   respectively, as they only track keys that are still in play.
 
-    A key's priority is only used to break ties, when many keys are being
-    considered for execution.  The priority does *not* determine running order,
```
pitrou: "The priority does *not* determine running order" sounds wrong to me, otherwise what would it be used for? The goal is to start computing some tasks earlier than others...

docs/source/scheduling-state.rst
```diff
@@ -375,12 +385,29 @@ Notes:
...
+Transition                           Affected worker state
+==================================== ==========================================================
+released → waiting                   occupancy, idle
+waiting → processing                 occupancy, idle, used_resources
+waiting → memory                     idle, worker_bytes
+processing → memory                  occupancy, idle, used_resources, worker_bytes
+processing → erred                   occupancy, idle, used_resources
+processing → released                occupancy, idle, used_resources
+memory → released                    worker_bytes
+memory → forgotten                   worker_bytes
```
mrocklin: The task in `tasks` itself is also forgotten if you want to include this

mrocklin: I am actually a bit surprised that the right column of this table is not more dense.  I would have expected more worker and task state to be affected.  Ah, I see that you are only listing worker state, never mind.

mrocklin:
In general this looks great to me.  I agree with all of the wording changes.  Thank you for going through this.

mrocklin:
I'm +1 on this PR and happy to merge any time.

pitrou:
Cool, thank you :-)

mrocklin:
Thank you!  This seems much clearer to me than before.

PR #1508, opened 2017-10-27
Add dict-like class to get/set datasets

mivade:
I find dict-style access much friendlier than explicitly using getters and setters.

Are there other dict methods that should be added to this?


mrocklin:
Note that this currently fails tests.  

See http://distributed.readthedocs.io/en/latest/develop.html#test

mrocklin:
In general I agree that having a MutableMapping style interface could be useful.  I suspect that this is difficult to achieve with the async API (which you're currently using for testing).  I suspect that this is only feasible with the synchronous API (see tests with `with cluster()`).

mrocklin:
It would also be useful to see if this object could inherit from `collections.MutableMapping`

mivade:
Thanks for the pointers. I had not actually been aware of `collections.MutableMapping`, but that looks like a good base class for this.

mivade:
@mrocklin I still need to switch tests over to use the synchronous API. I think you are correct that it is not immediately obvious how this would work with the async API as is, although presumably it can be used if I implement `__aiter__`

mrocklin:
This looks good to me.  A few final thoughts:

Can we move the implementation to `publish.py` and then import it into `client.py`?  I'm trying to start keeping non-core functionality out of client/scheduler/worker.py.

Any thoughts on a repr for this?  Perhaps listing a few keys if the client isn't in asynchronous mode?  See the `if self._loop_thread is not None:` in `Client.__str__` for a pattern to copy if you think a repr would be useful.

We might consider doing this for `Variables` and `Queues` as well sometime in the future:  

-  http://distributed.readthedocs.io/en/latest/api.html#distributed.Variable
-  http://distributed.readthedocs.io/en/latest/api.html#distributed.Queue

Though this could obviously be in a separate PR

mrocklin:
Merging this.  Thanks @mivade !

PR #1509, opened 2017-10-30
Avoid forkserver in PyPy

mrocklin:


pitrou:
What would be the reason? Does PyPy not support it?

mrocklin:
Correct.  PyPy3 doesn't seem to support forkserver.  This came up when I tried running a scheduler benchmark.

mrocklin:
@pitrou ok to merge?

pitrou:
Yes!

PR #1512, opened 2017-10-31
Remove some uses of IOLoop._running

pitrou:
Also some small cleanups.

One question revolves around the `loop` fixture. It asks the loop to stop at the end in case the test didn't stop already... and it turns out some tests don't seem to stop the loop indeed. Is it an error for a test not to do so?

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -256,6 +256,112 @@ def sync(loop, func, *args, **kwargs):
         return result[0]
 
 
+class LoopRunner(object):
+    """
+    A helper to start and stop an IO loop.
+
+    Parameters
+    ----------
+    loop: IOLoop (optional)
+        If given, this loop will be re-used, otherwise an appropriate one
+        will be looked up or created.
+    asynchronous: boolean (optional, default False)
+        If false (the default), the loop is meant to run in a separate
+        thread and will be started if necessary.
+        If true, the loop is meant to run in the thread this
+        object is instantiated from, and will not be started automatically.
+    """
+    def __init__(self, loop=None, asynchronous=False):
+        if loop is None:
+            if asynchronous:
+                self._loop = IOLoop.current()
+            else:
+                # We're expectinf the loop to run in another thread,
+                # avoid re-using this thread's assigned loop
+                self._loop = IOLoop()
+            self._should_close_loop = True
+        else:
+            self._loop = loop
+            self._should_close_loop = False
+        self._asynchronous = asynchronous
+        self._loop_thread = None
+        self._started = False
+
+    def start(self):
+        """
+        Start the IO loop if required.  The loop is run in a dedicated
+        thread.
+
+        If the loop is already running, this method does nothing.
+        """
+        if self._asynchronous or self._loop_thread is not None:
+            self._started = True
+            return
+
+        loop_evt = threading.Event()
+        done_evt = threading.Event()
+        in_thread = [None]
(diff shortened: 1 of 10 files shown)


pitrou:
@mrocklin

mrocklin:
> and it turns out some tests don't seem to stop the loop indeed. Is it an error for a test not to do so?

Most tests will not stop the IOLoop.  It is ok if they do or if they don't.

distributed/utils_test.py
```diff
@@ -73,24 +74,32 @@ def invalid_python_script(tmpdir_factory):
...
+@pytest.fixture
+def loop_in_thread():
+    with pristine_loop() as loop:
+        thread = threading.Thread(target=loop.start,
+                                  name="test IOLoop")
+        thread.daemon = True
+        thread.start()
+        loop_started = threading.Event()
+        loop.add_callback(loop_started.set)
+        loop_started.wait()
```
mrocklin: This pattern seems much cleaner than our other approaches.  We might consider doing this in client.py as well.

pitrou: Yes, I think I can refactor a bit more code in that PR.

distributed/utils.py
```diff
@@ -253,6 +256,112 @@ def f():
...
+            assert start_exc[0] is None, start_exc
+            self._loop_thread = thread
+
+    def stop(self, timeout=10):
+        """
+        Stop and close the loop if it was created by us.
+        Otherwise, just mark this object "stopped".
+        """
+        # XXX it would be more logical to stop an user-created loop
+        # if we started it, but that would break the current distributed API
```
mrocklin: In which situations would this break user code?  Some breakage may be acceptable.

pitrou: Well, the problem is it currently breaks some of our tests. For example this one:
```python
def test_Current(loop):
    with cluster() as (s, [a, b]):
        with Client(s['address'], loop=loop) as c:
            assert Client.current() is c
        with pytest.raises(ValueError):
            Client.current()
        with Client(s['address'], loop=loop) as c:
            assert Client.current() is c
```

As soon as the first `with` end, the loop is closed, but it's passed again to the second `Client` instead, so the test hangs.

I don't know if there use cases for continuing to use a loop after you closed the client that started it, but this is how it currently works.

mrocklin: If this is inconvenient then it might be reasonable to drop it.  The next version will likely be a 0.x.0 bump, so we can break a few things.

pitrou: It's not technically inconvenient, it just seems a bit inconsistent to have such assymetry in the start/stop responsabilities. But it probably wouldn't save any complexity to make it different.

mrocklin: If you think it's a good idea then go for it.

pitrou:
Will merge if no further comments.

pitrou:
The one crucial remaining use of `IOLoop._running` is in the `sync()` function. I haven't found a reliable way to do without it.

mrocklin:
That looks like a rarely used code branch to me.  Running tests now to see
where it comes up.

mrocklin:
I think that we can just remove this functionality.  I suspect that it does not come up in practice.

pitrou:
> I think that we can just remove this functionality. I suspect that it does not come up in practice.

Actually, it comes up in the `loop` fixture, since we want to ensure the loop is stopped before closing it.

PR #1513, opened 2017-10-31
Use new dask interface

mrocklin:
This uses the new dask interface created in https://github.com/dask/dask/pull/2748

cc @jcrist

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -24,7 +24,7 @@ import warnings
 import weakref
 
 import dask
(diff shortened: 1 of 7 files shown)


jcrist:
Overall this seems fine to me. This does break backwards compatibility with older versions of dask. Not sure how you feel about that.

If you wanted, backwards compatibility could be maintained by defining small helper functions for getting the graph, keys, and finalizers, depending on the version of dask.

mrocklin:
I think I'm ok with a hard dependency and synchronized release.  This seems big enough.

PR #1518, opened 2017-11-02
Tornado 5 fixes and workarounds

pitrou:
Most of the test suite seems to pass with this. IPython-related tests fail with some timeouts (perhaps ZMQ doesn't like Tornado 5?).

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -532,13 +532,7 @@ class Client(Node):
         though the client was originally created in asynchronous mode we may
         find ourselves in contexts when it is better to operate synchronously.
         """
-        result = self._asynchronous
-        try:
-            if self.loop._thread_ident and get_thread_identity() != self.loop._thread_ident:
-                result = False
-        except AttributeError:  # AsyncIOLoop doesn't have _thread_ident
-            pass
-        return result
+        return self._asynchronous and self.loop is IOLoop.current()
 
     def sync(self, func, *args, **kwargs):
         asynchronous = kwargs.pop('asynchronous', None)
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -205,6 +205,22 @@ def All(*args):
     raise gen.Return(results)
 
 
+def is_loop_running(loop):
+    """
+    Return whether an IOLoop is running.
+    """
+    # Tornado < 5.0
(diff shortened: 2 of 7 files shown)


distributed/utils.py
```diff
@@ -205,6 +205,22 @@ def All(*args):
     raise gen.Return(results)
 
 
+def is_loop_running(loop):
```
pitrou: Of course, this is not ideal, but gets us running until a decision is made on https://github.com/tornadoweb/tornado/pull/2085

distributed/client.py
```diff
@@ -532,13 +532,7 @@ def asynchronous(self):
         though the client was originally created in asynchronous mode we may
         find ourselves in contexts when it is better to operate synchronously.
         """
-        result = self._asynchronous
-        try:
-            if self.loop._thread_ident and get_thread_identity() != self.loop._thread_ident:
```
mrocklin: Could this also be handled by the LoopRunner?  We could capture the thread of the IOLoop at the same time we wait for the `threading.Event`

pitrou: That only works if the loop wasn't started by the user.

pitrou: Also, the same pattern occurs in `worker.py` below which doesn't have a LoopRunner.

mrocklin: This code is useful because it allows us to use clients that were created to work asynchronously safely outside of an asynchronous environment.

```python
@gen.coroutine
def f():
    client = yield Client(asynchronous=True)
    ...

client = ... get the same client ...
client.gather(...)  # this identifies that it is outside of the IOLoop and uses the blocking API
```

There are likely cleaner ways around this, such as by creating multiple clients or asking users to explicitly annotate all calls with `asynchronous=True/False`.  I'm -1 on the second option (users don't understand this).  The first option is possible, but would require some work on our part.

pitrou: I suspect my replacement code below should work, since `self.loop is IOLoop.current()` is false if you call it from another thread.

mrocklin: OK, you're right, that's probably nicer regardless.  Thanks!

pitrou:
I'm going to merge if no further comments.

PR #1523, opened 2017-11-02
Move thread_state in dask

jcrist:
Previously this was in the top namespace. In https://github.com/dask/dask/pull/2858 this is moved  to `dask.context`, adjust the import accordingly.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -44,7 +44,7 @@ from .metrics import time
 
 
 try:
-    from dask import thread_state
+    from dask.context import thread_state
 except ImportError:
     thread_state = threading.local()
 


distributed/utils.py
```diff
@@ -44,7 +44,7 @@
 
 
 try:
-    from dask import thread_state
+    from dask.context import thread_state
```
jcrist: There has been one distributed release, and no dask release since this was added. The users this backwards incompatible change affects are those running the new distributed release with dask master. If we wanted to be 100% backwards compatible we could check for `thread_state` in `dask.base` if it's not found in `dask.context`. Didn't seem worth it to me though.

mrocklin: Yeah, I wouldn't worry about it.  The next release will probably pin each to the other.

PR #1524, opened 2017-11-02
Allow automatic retries of a failed task

### LINKED ISSUE #852: Explicitly retry failed tasks
Currently if a task fails it is marked with an exception and that state is final.  If the user then changes an external reason for the failure they may want to rerun that failed task and restart all dependent tasks.

It would be convenient to have a `retry` operation that explicitly enabled this behavior.

pitrou:
Fix #852


distributed/scheduler.py
```diff
@@ -261,6 +263,7 @@ def __init__(
         self.worker_restrictions = dict()
         self.resource_restrictions = dict()
         self.loose_restrictions = set()
+        self.retries = defaultdict(lambda: 0)
```
mrocklin: It might be nice to keep this as a normal dict.  The common case is likely to be no retries at all.  It would be nice not to keep the extra `{key: 0}` around in that case.

distributed/tests/test_client.py
```diff
@@ -136,6 +136,34 @@ def test_map_keynames(c, s, a, b):
...
 @gen_cluster(client=True)
+def test_map_retries(c, s, a, b):
+    args = [[ZeroDivisionError("one"), 2, 3],
+            [4, 5, 6],
+            [ZeroDivisionError("seven"), ZeroDivisionError("eight"), 9]]
+
+    x, y, z = c.map(*map_varying(args), retries=2)
+    assert (yield x) == 2
+    assert (yield y) == 4
+    assert (yield z) == 9
```
mrocklin: I didn't know that `(yield foo)` was valid syntax.   That's useful.  Thanks!

distributed/tests/test_client.py
```diff
@@ -136,6 +136,84 @@ def test_map_keynames(c, s, a, b):
...
+
+    # Sanity check
+    x = c.persist(delayed(varying(args))())
+    fut = c.compute(x)
+    with pytest.raises(ZeroDivisionError) as exc_info:
+        yield fut
+    exc_info.match("one")
+
+    x = c.persist(delayed(varying(args))())
+    fut = c.compute(x, retries=1)
```
pitrou: Note you have to use `retries=1` on the compute() call. Do you think it's worthwhile having it as an argument on persist() too?

distributed/utils_test.py
```diff
@@ -211,6 +212,57 @@ def slowidentity(*args, **kwargs):
         return args
 
 
+def varying(items):
```
pitrou: I was surprised I had to use such an elaborate hack for this use case. Do you know of a simpler way?

distributed/client.py
```diff
@@ -2096,9 +2114,16 @@ def compute(self, collections, sync=False, optimize_graph=True,
         if resources:
             resources = self.expand_resources(resources)
 
+        if retries:
+            # Each task unit may potentially fail, allow retrying all of them
+            retries = {name: retries for name in itertools.chain(dsk, dsk2)}
+        else:
+            retries = None
```
mrocklin: This breaks a little bit from the convention we have for `resources`, where users can specify parts of their computation that have different resource constraints.

See this test for compute

```python
@gen_cluster(client=True, ncores=[('127.0.0.1', 1, {'resources': {'A': 1}}),
                                  ('127.0.0.1', 1, {'resources': {'B': 1}})])
def test_compute(c, s, a, b):
    x = delayed(inc)(1)
    y = delayed(inc)(x)

    yy = c.compute(y, resources={x: {'A': 1}, y: {'B': 1}})
    yield wait(yy)

    assert b.data
```

Arguably we should publish this convention in docs more prominently.  I suspect that the current `expand_resources` function already has the logic you would need here.  (that function should probably be renamed and made private).

distributed/tests/test_client.py
mrocklin: > Note you have to use retries=1 on the compute() call

I'm not sure I understand this statement

> Do you think it's worthwhile having it as an argument on persist() too?

I think so.  To me they should offer the same capabilities.

distributed/utils_test.py
mrocklin: No, not really.  I haven't thought about this much.  Though elaborate, it does appear to get the job done :)

distributed/tests/test_client.py
pitrou: So the following should show similar behavior?
```
x = c.persist(...)
fut = c.compute(x, retries=N)
```
vs.
```
x = c.persist(..., retries=N)
fut = c.compute(x)
```
or should there be a difference?

distributed/client.py
pitrou: Ok, I improved `retries` handling to allow both an int or a dict of per-task specifications.

distributed/client.py
```diff
@@ -2959,8 +2984,21 @@ def expand_resources(resources):
...
+    @classmethod
+    def _expand_retries(cls, retries, all_keys):
+        if retries and isinstance(retries, dict):
+            return cls._expand_per_key_specification(retries)
+        elif isinstance(retries, Integral):
+            # Each task unit may potentially fail, allow retrying all of them
+            return {name: retries for name in all_keys}
+        else:
+            raise TypeError("`retries` should be an integer or dict, got %r"
+                            % (type(retries,)))
```
mrocklin: Would it be possible to have the exact same behavior for resources and retries?  I'm curious if there are reasons why the behaviors here should not be identical.

distributed/tests/test_client.py
mrocklin: I think so, yes, except that in the first case it's possible that the tasks would have already failed before the second `update_graph` call made it to the scheduler.

distributed/client.py
pitrou: Do you mean you don't want the `retries=3` shortcut in case you want the same value for all futures? Or something else?

mrocklin: It seems like a very valuable shortcut to me.  I wonder if we might also have the same shortcut for resources as well like `resources={'GPU': 1}`.  I'm actually not sure what the current behavior for this is.

mrocklin: So, to be clear, my objective here is consistent behavior across the different keywords more than an adherence to previous behaviors.

pitrou: Oh, I see. Do you want it as part of this PR?

mrocklin: If you're up for it, yes.  We should probably try to maintain API consistency as part of the normal dev process.

pitrou: I'll do.

pitrou: Hmm, so there is semantic ambiguity here. If I pass `{'GPU': 1}` for several tasks at once, does each task need a GPU for itself, or do they only need one GPU for all of them? Probably the former, but just wanting to check.

rbubley: IMHO a logical semantic would be to have { ‘GPU’ : True } meaning that the task requires that there is a GPU, that could potentially be used by multiple tasks, whereas { ‘GPU’ : 1} would mean that the task required 1 GPU for its exclusive use. ( C.f { ‘DiskSpace’ : 1e9 } to mean a task needed 1GB of disk space for its exclusive use.)

mrocklin: Presumably this same ambiguity exists for retries as well?

pitrou: @mrocklin I don't really see how.

pitrou: @rbubley `1 == True` means it may get confusing :-)

rbubley: @pitrou - pah! Don’t you want lots of isinstance(x, bool) to disambiguate?

pitrou:
@mrocklin Do you want to take another look here?

mrocklin:
This looks good to me.  Merging.  I think that this will make many people happy :)

PR #1527, opened 2017-11-03
Make worker memory_monitor collect data with cyclic references

ogrisel:
This PR:

- calls `collect()` systematically after any eviction without relying on the weight measurements,
- call the memory pause check nested in the eviction loop in case the workers threads grow the memory faster than the eviction mechanism frees the memory,
- adds a quick stress test to check that large number of tasks with cyclic refs are properly evicted with 

Note: the `sizeof` / `weight` measurements are completely blind to nested.

This is a more robust alternative to #1494 (nanny sending signal to trigger GC).

Note that this does not fix the all the memory problems reported by @bluenote10  as the eviction mechanism itself is very memory hungry because it makes in-memory copies of the data before writing to disk. This is very problematic when working with arrays or dataframe blocks that are a couple GB each.

### DIFF
--- a/distributed/tests/test_stress.py
+++ b/distributed/tests/test_stress.py
@@ -5,11 +5,13 @@ from datetime import timedelta
 from operator import add
 import random
 import sys
+import os
 from time import sleep
 
 from dask import delayed
 import pytest
 from toolz import concat, sliding_window
+import psutil
 
 from distributed import Client, wait, Nanny
 from distributed.config import config
@@ -20,6 +22,17 @@ from distributed.utils_test import (gen_cluster, cluster, inc, slowinc,
 from distributed.utils_test import loop # flake8: noqa
 from distributed.client import wait
 from tornado import gen
+
+
+def get_folder_size(folder):
+    total_size = os.path.getsize(folder)
+    for item in os.listdir(folder):
+        itempath = os.path.join(folder, item)
+        if os.path.isfile(itempath):
+            total_size += os.path.getsize(itempath)
+        elif os.path.isdir(itempath):
+            total_size += get_folder_size(itempath)
+    return total_size
 
 
 @gen_cluster(client=True)
@@ -266,7 +279,7 @@ def test_no_delay_during_large_transfer(c, s, w):
     assert nbytes[-1] < (x.nbytes * 1.2) / 1e6
 
 
-def test_nanny_no_terminate_on_cyclic_ref():
+def test_nanny_no_terminate_on_cyclic_ref(tmpdir):
     from time import sleep
 
     class BigCyclicRef(object):
@@ -275,7 +288,11 @@ def test_nanny_no_terminate_on_cyclic_ref():
             self.data = b'0' * size
             self.cyclic_ref = self
 
-    w_kwargs = {'memory_limit': int(1e9)}
+    worker_folder = str(tmpdir)
+    memory_limit = int(1e9)
+    size = int(5e7)
+    n_tasks = 40
+    w_kwargs = {'memory_limit': memory_limit, 'local_dir': worker_folder}
     with cluster(nworkers=1, nanny=True, worker_kwargs=w_kwargs) as (s, [a]):
         with Client(s['address']) as c:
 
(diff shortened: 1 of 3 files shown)


mrocklin:
> Note that this does not fix the all the memory problems reported by @bluenote10 as the eviction mechanism itself is very memory hungry because it makes in-memory copies of the data before writing to disk. This is very problematic when working with arrays or dataframe blocks that are a couple GB each.

Can you expand on this?  What copies are you referring to?  I'd be happy to do more work here if it would help.

distributed/tests/test_stress.py
```diff
@@ -264,3 +264,42 @@ def test_no_delay_during_large_transfer(c, s, w):
...
+            while True:
+                sleep(0.1)
+                # Ensure that the worker process has not been restarted by the
+                # nanny memory monitor.
+                assert get_worker_pids() == pids
+                assert time() < start + 30
+                if all(f.done() for f in futures):
+                    break
+            len(futures[0].result().data) == size
+            del futures
```
mrocklin: Do we also want to test the memory use of the worker processes here to verify that they are below their specified limit?

distributed/tests/test_worker.py
```diff
@@ -30,7 +30,6 @@
 from distributed.utils_test import (inc, mul, gen_cluster, div, dec,
                                     slow, slowinc, gen_test, cluster,
                                     captured_logger)
-from distributed.utils_test import loop # flake8: noqa
```
mrocklin: It looks like this change is causing test failures

ogrisel: Indeed, I must have deleted this line by mistake.

ogrisel:
> >   This is very problematic when working with arrays or dataframe blocks that are a couple GB each.

> Can you expand on this? What copies are you referring to? I'd be happy to do more work here if it would help.

I have not yet conducted a systematic investigation but I believe that the following lines are problematic:

```python
            path = os.path.join(self.local_dir, 'storage')
            storage = Func(serialize_bytelist, deserialize_bytes, File(path))
            target = int(float(self.memory_limit) * self.memory_target_fraction)
            self.data = Buffer({}, storage, target, weight)
```

This means that we hold the bytes of the results of the call to `serialize_bytelist` in memory before streaming them to disk.

Instead we would probably need something like:

```python
            path = os.path.join(self.local_dir, 'storage')
            storage = StreamableFile(path, serialize_bytechunks, deserialize_bytes)
            target = int(float(self.memory_limit) * self.memory_target_fraction)
            self.data = Buffer({}, storage, target, weight)
```

Where `serialize_bytechunks` returns an iterator on chunks of bytes to be written to disk before extracting the next chunk. Chunk would be a couple of tens of MB maximum to avoid memory allocations of large temporary data structures.

distributed/tests/test_stress.py
ogrisel: Good suggestion, done in my last commit. I also added a check on the worker `local_dir` folder size.

ogrisel:
This change is causing `test_nanny` to fail, at least on appveyor. I need to check what assumptions are actually made by that test.

mrocklin:
I would expect the current `serialize_bytelist` operation to work well for numpy arrays, where we're just passing around the `my_array.data` memoryview, but poorly for dataframes, where we call pickle.dumps.  If this is true then we might consider being more intelligent about serializing pandas dataframes by passing the memoryviews of the blocks directly.

ogrisel:
Indeed, the simulated shuffle workload suggested by @bluenote10 for instance in https://github.com/dask/distributed/issues/1015#issuecomment-313401739 is using dataframes so this might be the cause of the problem. I also encountered what looks like a similar issue when working with large `df.read_csv` and `df.read_parquet` which also deal with large dataframes.

I plan to work on writing another stress test to investigate this family of cases.

distributed/worker.py
```diff
@@ -2182,73 +2182,86 @@ def execute(self, key, report=False):
...
-                    logger.warn("Memory use is high but worker has no data "
-                                "to store to disk.  Perhaps some other process "
-                                "is leaking memory?  Process memory: %s -- "
-                                "Worker memory limit: %s",
-                                format_bytes(proc.memory_info().rss),
-                                format_bytes(self.memory_limit))
+                    self._throttled_gc.collect()
+                    yield gen.moment
+                    memory = proc.memory_info().rss
+                    self._check_pause_resume(memory)
```
mrocklin: Why do we repeat these checks?  A (not necessarily better) alternative would be to wait for the next iteration of the `memory_monitor`.

mrocklin:
@ogrisel is there anything I can do to help here?

mrocklin:

PR #1534, opened 2017-11-06
Remove all uses of IOLoop._running

pitrou:
Monkey-patch `IOLoop.start` in the `loop` fixture to know when the loop is started / stopped. Also other fixes.

This seems to work more or less with Tornado changeset cccd5fa and Python 3.6 (excluding IPython- and Bokeh-related functionality).


pitrou:
Looks like this breaks with Tornado < 5 :-(

pitrou:
Note to self: we use `sync` in at least some `__repr__` implementations.

pitrou:
For the record: ipython issue reported at https://github.com/ipython/ipykernel/issues/277

conftest.py
```diff
@@ -1,6 +1,19 @@
 # https://pytest.org/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option
+import os
 import pytest
 
+
+# Enable more logging and checks (https://docs.python.org/3/library/asyncio-dev.html)
+os.environ["PYTHONASYNCIODEBUG"] = "1"
```
mrocklin: You mentioned that this makes things run a bit slowly.  Is there a cost to this?

distributed/tests/test_client.py
```diff
@@ -165,6 +165,46 @@ def test_Future_exception_sync(loop):
...
 
 
+@gen_cluster(client=True)
+def test_Future_release(c, s, a, b):
+    # Released Futures should be removed timely from the Client
+    x = c.submit(div, 1, 1)
+    yield x
+    x.release()
+    yield gen.moment
+    assert not c.futures
```
mrocklin: Does tornado definitely go through all callbacks before coming back to this one?

distributed/tests/test_client.py
```diff
@@ -165,6 +165,46 @@ def test_Future_exception_sync(loop):
...
+            wait_for(lambda: not c.futures, timeout=0.1)
+
+            x = c.submit(slowinc, 1, delay=0.5)
+            x.release()
+            wait_for(lambda: not c.futures, timeout=0.1)
+
+            x = c.submit(div, 1, 0)
+            x.exception()
+            x.release()
+            wait_for(lambda: not c.futures, timeout=0.1)
```
mrocklin: I recommend increasing these timeouts for travis-ci.  It can occasionally be surprisingly slow.

distributed/utils.py
```diff
@@ -205,26 +206,14 @@ def All(*args):
...
-                        % (type(loop),))
-
-
 def sync(loop, func, *args, **kwargs):
     """
     Run coroutine in loop running in separate thread.
     """
+    # Tornado's PollIOLoop doesn't raise when using closed, do it ourselves
+    if isinstance(loop, PollIOLoop) and getattr(loop, '_closing', False):
+        raise RuntimeError("IOLoop is closed")
```
mrocklin: Why was this necessary?  Who uses `PollIOLoop`?  Windows?

pitrou: Python 2 on Tornado 5, and all users of Tornado 4 :-)

distributed/tests/test_client.py
pitrou: Not sure I understand the question... which callbacks are you talking about?

distributed/utils.py
mrocklin: Ah, indeed

```python
In [1]: from tornado.ioloop import IOLoop

In [2]: type(IOLoop.current()).mro()
Out[2]: 
[tornado.platform.epoll.EPollIOLoop,
 tornado.ioloop.PollIOLoop,
 tornado.ioloop.IOLoop,
 tornado.util.Configurable,
 object]
```

I was on master when I wrote that comment

distributed/tests/test_client.py
mrocklin: In `x.release()` we add a callback to `_dec_ref`, then we wait a `gen.moment`, then we assume that that callback has run.  Is this a safe assumption?

pitrou: Yes, `gen.moment` should run all ready callbacks.

distributed/tests/test_client.py
pitrou: Ok, will do.

conftest.py
pitrou: With Tornado 4.5.2 (PollIOLoop) I get: 948 passed, 91 skipped, 9 xfailed, 3 xpassed, 1 warnings in 452.45 seconds
With Tornado git master (AsyncIOLoop, debug enabled) I get: 2 failed, 939 passed, 98 skipped, 11 xfailed, 1 xpassed, 1 warnings in 749.41 seconds
With Tornado git master (AsyncIOLoop, debug disabled) I get: 1 failed, 940 passed, 98 skipped, 8 xfailed, 4 xpassed, 1 warnings in 394.44 seconds

(this is Python 3.6, without the slow tests)

Moreover, it seems that in debug mode, the py.test process' RSS balloons up to 3GB at the end of the test suite. So we'd better disable debug mode after all.

pitrou:
@mrocklin would you be ok with merging this?

mrocklin:
Generally yes.  I think the question is whether or not we want to wait for a release or push it in now.  I think that there are good reasons for either.  It's a biggish change and so we may want to have it in master for a while before releasing.  However we're also releasing a 0.X.0 release, so it would be nice to include large changes.  

Do you have any thoughts or preferences here?

pitrou:
I think the 0.X.0 is a good opportunity to exercise those changes. Our users have probably learnt to be a bit aventurous :-)

mrocklin:
Then I'm +1 on this.  Feel free to merge at will.

PR #1536, opened 2017-11-07
Avoid doing I/O in repr() or str()

pitrou:
Also standardize on defining `__repr__`.
`__str__` automatically falls back on `__repr__` unless overriden.

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -1694,7 +1694,7 @@ def test_repr(loop):
             for func in funcs:
                 text = func(c)
                 assert c.scheduler.address in text
-                assert '2' in text
+                assert '3' in text
 
         for func in funcs:
             text = func(c)
--- a/distributed/batched.py
+++ b/distributed/batched.py
@@ -56,10 +56,8 @@ class BatchedSend(object):
         self.comm = comm
         self.loop.add_callback(self._background_send)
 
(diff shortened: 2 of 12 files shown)


distributed/tests/test_client.py
```diff
@@ -1694,7 +1694,7 @@ def test_repr(loop):
             for func in funcs:
                 text = func(c)
                 assert c.scheduler.address in text
-                assert '2' in text
```
pitrou: This would succeed simply because the address of the scheduler is always `tcp://127.0.0.1:something` ;-)

pitrou:
@mrocklin I'd like to merge this quickly.

mrocklin:
I often use the repr to check how many cores/RAM I have in the cluster.  This can change during the course of a client's lifetime.  It is valuable to check in with the scheduler in the repr.

pitrou:
Well, there is a periodic callback to update that information every 2 seconds. That should be good enough?

pitrou:
We can't do I/O in `__repr__` as it can be invoked at any time. For example, asyncio in debug mode invokes `repr()` on callback arguments... if you do I/O by calling back into asyncio at that point, things completely break.

mrocklin:
Ah I missed that, yes, that should be good enough.

pitrou:
Cool, thanks :)

PR #1538, opened 2017-11-07
Protect transition_processing_released from worker failure

mrocklin:
An ill-timed worker failure could cause a task to go into an ill-defined state.
This could use a test

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -840,7 +840,7 @@ class Scheduler(ServerNode):
                          worker, self.task_state.get(key), key,
                          self.who_has.get(key))
             if worker not in self.who_has.get(key, ()):
-                self.worker_comms[worker].send({'op': 'release-task', 'key': key})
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -2747,7 +2747,11 @@ def transition_processing_released(self, key):
                 self.total_occupancy -= duration
                 self.check_idle_saturated(w)
                 self.release_resources(key, w)
-                self.worker_comms[w].send({'op': 'release-task', 'key': key})
+                try:
+                    self.worker_comms[w].send({'op': 'release-task',
+                                               'key': key})
+                except EnvironmentError:
```
pitrou: It's specifically a CommClosedError, right?

mrocklin: Replaced with `CommClosedError`

mrocklin:
I've cleaned this up a bit and applied it more consistently throughout the codebase.  Merging later today if there are no further comments.

PR #1543, opened 2017-11-09
Add a disposable workspace facility

pitrou:


pitrou:
Curious about the test results on Windows and macOS.

pitrou:
Open question: do we have users with complicated network storage setups where file locking may be unsupported?

mrocklin:
> Open question: do we have users with complicated network storage setups where file locking may be unsupported?

@davidedelvento may have an answer to this

davidedelvento:
Not sure what you mean by "complicated network storage setups". On our HPC infrastructure the only network storage we have is spectra-whatever, aka with its old name GPFS. See https://www2.cisl.ucar.edu/resources/storage-and-file-systems/glade-file-spaces for details

pitrou:
@davidedelvento the main question is whether the filesystem in question supports file locking. Unfortunately, the docs you point to don't seem to say anything about that :-)

pitrou:
I think this PR is technically ready now (barring CI validation perhaps). I also added a config option to disable it for setups where file locking is problematic. I'm still not 100% sure this is required or desirable, but it does remove most test leftovers here.

distributed/locket.py
```diff
@@ -0,0 +1,198 @@
+# Vendored up-to-date copy of locket.py
+# Based on https://github.com/mwilliamson/locket.py/pull/8
```
mrocklin: I'm curious if the locket maintainers would be willing to push a release quickly so that we could avoid vendoring this project.

cc @mwilliamson @eryksun

Fine either way, just thought I'd ask

mrocklin:
I quickly played with this PR and found the experience to be present.  I liked both that all of the `worker-*` directories were isolated in a separate directory and that they were cleanly handled even after a SIGTERM event (and a subsequent loading of the next workers).

I'll admit that I haven't looked over the implemention in detail.  What I did see was clear though.

distributed/locket.py
mwilliamson: I'm a bit busy at the moment and probably don't have time to review the pull request, so probably not I'm afraid.

mrocklin:
Merging this in.  Thanks @pitrou

PR #1545, opened 2017-11-10
Keep timings consistent between processing, occupancy and total_occupancy. Fixes #1544

### LINKED ISSUE #1544: No tasks on scheduler yet total_occupancy > 0
I observed an example yesterday of the scheduler having no remaining tasks, yet total_occupancy was > 0. This mattered, because I was using Adaptive, and it meant there was a continuous cycle of spin up and start a new worker, only for it to be immediately closed.

Relevant log snippet below:
```
2017-11-09 20:01:13  distributed.deploy.adaptive (scheduler): CPU limit exceeded [353 occupancy / 0 cores]
2017-11-09 20:01:13  distributed.deploy.adaptive (scheduler): Scaling up to 1 workers
2017-11-09 20:01:17  distributed.comm.tcp (worker-group-1-7kl0): Setting TCP keepalive: nprobes=10, idle=10, interval=2
2017-11-09 20:01:17  distributed.comm.tcp (worker-group-1-7kl0): Setting TCP user timeout: 30000 ms
2017-11-09 20:01:17  distributed.comm.tcp (worker-group-1-7kl0): TLS connection with 'tls://scheduler:8786': protocol=TLSv1.2, cipher=ECDHE-RSA-AES256-GCM-SHA384, bits=256
2017-11-09 20:01:17  distributed.scheduler (scheduler): Register tls://10.154.0.5:32819
2017-11-09 20:01:17  distributed.scheduler (scheduler): Starting worker compute stream, tls://10.154.0.5:32819
2017-11-09 20:01:18  distributed.scheduler (scheduler): Closing worker tls://10.154.0.5:32819
2017-11-09 20:01:18  distributed.core (worker-group-1-7kl0): Message from 'tls://10.154.0.4:33068': {'reply': False, 'op': 'close'}
2017-11-09 20:01:18  distributed.scheduler (scheduler): Remove worker tls://10.154.0.5:32819
2017-11-09 20:01:18  distributed.scheduler (scheduler): Lost all workers
2017-11-09 20:01:18  distributed.scheduler (scheduler): Removed worker tls://10.154.0.5:32819
2017-11-09 20:01:18  distributed.deploy.adaptive (scheduler): Retiring workers ['tls://10.154.0.5:32819']
```

I suspect this is because in the scheduler, `transition_processing_released` (unlike in `transition_processing_memory`), there is no check to see if there is any outstanding processing, following which occupancy is set to zero. But before proposing a PR, I thought I would see if someone with a better understanding of the scheduler concurred.

rbubley:


mrocklin:
Is it feasible to test this behavior?  "No" is a fine answer here, but it's worth considering.

rbubley:
The trouble is I'm not sure what makes the "no-task occupancy >0"  behaviour occur: I've observed it in the wild and tried a few simple ways to replicate the problem but without success. So without a replicable case I'm not sure what to test for. A test would be something like the below...

```python
@gen_cluster(client=True)
def test_occupancy_cleardown(c, s, a, b):
    futures = c.map(...) # Do something, but what?

    yield wait(futures)
    assert abs(s.total_occupancy) < 1e6
```

mrocklin:
Perhaps we could inject an incorrect occupancy value directly and then test that things zero out afterwards?

distributed/scheduler.py
```diff
@@ -647,6 +647,10 @@ def add_worker(self, comm=None, address=None, keys=(), ncores=None,
                 self.has_what[address] = set()
                 self.worker_bytes[address] = 0
                 self.processing[address] = dict()
+                try:
+                    self.total_occupancy -= self.occupancy[address]
+                except KeyError:
+                    pass
                 self.occupancy[address] = 0
```
mrocklin: Under what conditions would we get a `KeyError` here?

distributed/scheduler.py
```diff
@@ -1411,6 +1415,7 @@ def handle_long_running(self, key=None, worker=None, compute_duration=None):
 
         worker = self.rprocessing[key]
         self.occupancy[actual_worker] -= self.processing[actual_worker][key]
+        self.total_occupancy -= self.processing[actual_worker][key]
```
mrocklin: Nice catch!

distributed/scheduler.py
rbubley: It’s actually the usual case: when the worker is first added, this KeyError/pass will be taken; it’s only not a KeyError when workers are re-added (and possibly some other circumstances?)

mrocklin: Ah I see, well then I'll change my question to, under what conditions would `address` be in `self.processing` but not in `self.occupancy`?  Should we perhaps avoid this situation altogether and consider it an error if one exists but not the other?

rbubley: I’m actually now wondering if it is possible to ever be anything but the KeyError path..

rbubley: I don't think the code guarantees that add_worker can never be called a second time for a worker, without the occupancy having been cleared first, so I think the try..except is necessary.

mrocklin: We definitely call `add_worker` many times during a worker's lifetime (heartbeats call add_worker).  But the `address not in self.processing` should be enough to protect this part from repeated calls.  `self.processing[worker]` is created on first instantiation and only destroyed on `remove_worker`.

rbubley: Yes, I agree. The test is redundant; I'll remove it.

mrocklin:
I can get this to fail locally.

```python
    @gen_cluster(client=True)
    def test_occupancy_cleardown(c, s, a, b):
        futures = c.map(slowinc, range(100), delay=0.01)
        yield wait(futures)
    
        for w in s.occupancy:
            s.occupancy[w] += 2
            s.total_occupancy += 2
    
        futures = c.map(slowinc, range(200), delay=0.01)
        yield wait(futures)
    
>       assert abs(s.total_occupancy) < 0.01
E       assert 2.0011097265165354 < 0.01
E        +  where 2.0011097265165354 = abs(2.0011097265165354)
E        +    where 2.0011097265165354 = <Scheduler: "tcp://127.0.0.1:43251" processes: 2 cores: 3>.total_occupancy
```

By placing a breakpoint where the assert would fail I can get this:

```python
-> import pdb; pdb.set_trace()
(Pdb) pp s.total_occupancy
1.9985174130045762
(Pdb) pp s.occupancy
{'tcp://127.0.0.1:35869': 0, 'tcp://127.0.0.1:39805': 0}
```

So at some point these two pieces of state are able to fall out of sync.

mrocklin:
My guess is that this is due to the use of `occupancy` in stealing.py, which currently doesn't adjust `total_occupancy` in lines 207 and 208.

rbubley:
I can't make it fail locally, and the travis/appveyor set-ups don't fail either. Is there a test you can think of that would make the failure you are seeing happen elsewhere? In the meantime, I'll look at stealing.py to fix the issue you mentioned.

mrocklin:
I just ran the test in a loop on my machine

    for i in {1..100}; do py.test distributed/tests/test_scheduler.py::test_occupancy_cleardown --pdb; done

rbubley:
OK thanks. I'll track it down. I think there are some other issues in stealing.py with regards to the occupancy calculations, too, which I am currently looking into.

mrocklin:
I appreciate you tracking this down.  It's nice to find other people who are comfortable diving into the scheduler :)

rbubley:
OK: so it transpired that there were a couple of transitions in stealing that didn't preserve the invariants sum(processing[w]) == occupancy[w] and sum(occupancy) == total_occupancy. This was because the changes were split between `move_task_request` and `move_task_confirm`, so if something else happened in between, the invariance could be lost.

To fix this,

distributed/stealing.py
```diff
@@ -165,15 +165,11 @@ def move_task_request(self, key, victim, thief):
...
             victim_duration = self.scheduler.processing[victim][key]
-            self.scheduler.occupancy[victim] -= victim_duration
-            self.scheduler.total_occupancy -= victim_duration
 
             thief_duration = self.scheduler.task_duration.get(key_split(key), 0.5)
             thief_duration += sum(self.scheduler.nbytes[key] for key in
                                   self.scheduler.dependencies[key] -
                                   self.scheduler.has_what[thief]) / BANDWIDTH
-            self.scheduler.occupancy[thief] += thief_duration
-            self.scheduler.total_occupancy += thief_duration
```
mrocklin: Unfortunately it's important that we change the occupancy when we first try to steal.  This lets the load balancing machinery know when it should stop trying to move tasks around.

rbubley: Hmm. But this leads to undesirable behaviour elsewhere. 

What if `balance` were to look at an augmented version of occupancy, created by putting together info from occupancy and in_flight?

mrocklin: > What if balance were to look at an augmented version of occupancy, created by putting together info from occupancy and in_flight?

That seems reasonable to me.  This would also allow us to uncomment the final check in `Scheduler.validate_state`.

distributed/scheduler.py
```diff
@@ -3159,8 +3159,9 @@ def reschedule(self, key=None, worker=None):
     # Assigning Tasks to Workers #
     ##############################
 
-    def check_idle_saturated(self, worker):
-        occ = self.occupancy[worker]
+    def check_idle_saturated(self, worker, occ=0):
```
mrocklin: I recommend `occ=None` here and `if occ is not None:`  Otherwise the meaning of `occ=0` is ambiguous between "there is no occupancy" and "please use the default behavior"

mrocklin: I'm slightly concerned that other operations that call `check_idle_saturated` will now use the incorrect value in `Scheduler.occupancy` if they are called between requesting and confirming a steal.

distributed/stealing.py
```diff
@@ -178,6 +179,14 @@ def move_task_request(self, key, victim, thief):
...
                                    'victim_duration': victim_duration,
                                    'thief_duration': thief_duration}
+            try:
+                self.in_flight_occupancy[victim] -= victim_duration
+            except KeyError:
+                self.in_flight_occupancy[victim] = -victim_duration
+            try:
+                self.in_flight_occupancy[thief] += thief_duration
+            except KeyError:
+                self.in_flight_occupancy[thief] = thief_duration
```
mrocklin: It might make sense to define `self.in_flight_occupancy = defaultdict(lambda: 0)` instead of using these try-excepts.

distributed/scheduler.py
rbubley: Agreed on occ=None.

If I understand the code correctly, as written now, the task is effectively owned by the victim until the theft is confirmed; before ownership was sort of blurred between request and confirmation. So, what bad things might happen with other things looking at check_idle_saturated? The only thing I could think of was that the thief gets killed - and move_task_confirm already tests for this.

distributed/stealing.py
rbubley: Agreed.

distributed/scheduler.py
mrocklin: It might be placed back into idle or saturated, which would trigger additional work stealing.  This may not be that bad though.

distributed/tests/test_scheduler.py
```diff
@@ -824,6 +824,14 @@ def test_learn_occupancy_2(c, s, a, b):
     assert 50 < s.total_occupancy < 200
 
 
+@gen_cluster(client=True)
+def test_occupancy_cleardown(c, s, a, b):
+    futures = c.map(slowinc, range(100), delay=0.01)
+    yield wait(futures)
+
+    assert abs(s.total_occupancy) < 0.01
```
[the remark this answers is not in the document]
rbubley: With the tests now in validate_state, injecting fake occupancy figures (without also injecting them to processing) will cause an AssertionError. And that just tests that validate_state is working, which is not what we are trying to test.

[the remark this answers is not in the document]
mrocklin: I've pushed up a small change that might work.  Please let me know if it's ok with you.

rbubley:
Looks good.

mrocklin:
This is in.  Thanks for all your work @rbubley !

Issue #1519, opened 2017-11-02
Accelerate tests


mrocklin:
From this record on travis-ci these are our slowest tests:

```
========================== slowest 20 test durations ===========================
41.75s call     distributed/tests/test_scheduler.py::test_file_descriptors
35.96s call     distributed/tests/test_stress.py::test_close_connections
20.66s call     distributed/tests/test_stress.py::test_cancel_stress_sync
15.86s call     distributed/tests/test_client.py::test_open_close_many_workers[Worker-100-5]
13.79s call     distributed/tests/test_client.py::test_open_close_many_workers[Nanny-10-20]
13.57s call     distributed/tests/test_stress.py::test_cancel_stress
12.33s call     distributed/tests/test_stress.py::test_stress_creation_and_deletion
10.35s call     distributed/tests/test_variable.py::test_race
10.19s call     distributed/tests/test_client.py::test_missing_worker
9.80s call     distributed/tests/test_client.py::test_threadsafe_get
9.41s call     distributed/tests/test_client.py::test_threadsafe_compute
7.74s call     distributed/deploy/tests/test_local.py::test_defaults
7.40s call     distributed/tests/test_asyncio.py::test_asyncio_restart
6.92s call     distributed/tests/test_stress.py::test_stress_gc[func1-1000]
6.87s call     distributed/tests/test_client.py::test_secede_balances
5.88s call     distributed/tests/test_worker.py::test_worker_death_timeout
5.08s call     distributed/tests/test_queues.py::test_race
4.98s call     distributed/tests/test_nanny.py::test_nanny_terminate
4.63s call     distributed/tests/test_scheduler.py::test_non_existent_worker
4.41s call     distributed/tests/test_worker_failure.py::test_gather_then_submit_after_failed_workers
```

It may be worth looking at a few of the slower ones to see if they can be accelerated.

pitrou:
Do you remember why `test_file_descriptors` was written the way it is? Does it attempt to reproduce a particular situation?

mrocklin:
It looks like it's trying to test a few different situations in one test..  It wants to verify that FDs don't increase without reason.  It wants to ensure that we only add one FD per worker when we do work, even if that work has many small communications.  The last few checks seem odd to me, they're more lenient than I would have expected.

pitrou:
Update from a recent PR:
```
========================== slowest 20 test durations ===========================
42.38s call     distributed/tests/test_stress.py::test_close_connections
[...]

PR #1549, opened 2017-11-13
Issue #1519: Make test_file_descriptors faster

pitrou:
Also fix an assertion error in transition_released_erred()


pitrou:
@mrocklin

distributed/scheduler.py
```diff
@@ -2234,6 +2234,7 @@ def transition_released_waiting(self, key):
 
             for dep in self.dependencies[key]:
                 if dep in self.exceptions_blame:
+                    del self.waiting[key]
```
mrocklin: Instead we might move the `self.waiting[key] = set()` line below this for loop.

PR #1550, opened 2017-11-13
Fix test_stress.py::test_close_connections

pitrou:
The loop which closes communications at random wasn't even entered (here at least) as the "processing" check occurred before the scheduler had actually started to spawn tasks (thus s.processing was empty).

Note that the test passes nominally, but it dumps some validation warnings in the Worker. @mrocklin

### DIFF
--- a/distributed/tests/test_stress.py
+++ b/distributed/tests/test_stress.py
@@ -220,18 +220,18 @@ def test_stress_steal(c, s, *workers):
 
 
 @slow
-@gen_cluster(ncores=[('127.0.0.1', 1)] * 10, client=True, timeout=120)
+@gen_cluster(ncores=[('127.0.0.1', 1)] * 20, client=True, timeout=120)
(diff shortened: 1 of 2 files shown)


mrocklin:
Thank you for identifying this.  This warning in particular is concerning (we've tried to remove all of these cases):

   distributed.scheduler - WARNING - Unexpected worker completed task, likely due to work stealing.  Expected: tcp://127.0.0.1:34629, Got: tcp://127.0.0.1:34467, Key: ('transpose-25893ce5a91e42a327c7ad41b84ec24b', 7, 0)

I guess it makes sense though if the worker is temporarily removed and then returned, or if only some of the communications are severed.

Although really it would be interesting to investigate more of them as well.  Is this something you feel comfortable doing?

pitrou:
> Although really it would be interesting to investigate more of them as well. Is this something you feel comfortable doing?

Not yet unfortunately.

PR #1554, opened 2017-11-14
Add HDFS wrapper for pyarrow compat

jcrist:
Add a wrapper for `DaskHDFileSystem` to provide an arrow compatible
interface to our hdfs filesystem. Adds a test using parquet to ensure
compatibility. Depends on https://github.com/dask/dask/pull/2881.

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -64,7 +64,8 @@ if [[ $HDFS == true ]]; then
     pip install -q git+https://github.com/dask/hdfs3 --upgrade
 fi;
 
-pip install -q git+https://github.com/dask/dask.git --upgrade
+# TODO: revert to install from dask after this PR is merged
+pip install -q git+https://github.com/jcrist/dask.git@arrow-hdfs-works --upgrade
 pip install -q git+https://github.com/joblib/joblib.git --upgrade
 pip install -q git+https://github.com/dask/s3fs.git --upgrade
 pip install -q git+https://github.com/dask/zict.git --upgrade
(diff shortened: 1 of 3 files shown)


mrocklin:
Moderate-term I wouldn't mind removing the HDFS code in this repo to somewhere separate.  It might be unwise to add more code here and expect it to survive for a long while.

mrocklin:
This would be more work, but I would be much happier to pull the hdfs code (and testing infrastructure) out of this repository and create a separate `dask-hdfs` package.

jcrist:
I'm hoping this code is quite temporary, and won't last longer than a release or two. Ideally we'd settle on a standard interface that could live in each project and wouldn't require these wrappers.

jcrist:
> This would be more work, but I would be much happier to pull the hdfs code (and testing infrastructure) out of this repository and create a separate dask-hdfs package.

Actually, this wouldn't be too bad to do, and a change like this would be good to get in before the next big release. I'll try and set this up tonight/tomorrow.

mrocklin:
Really my hope was that you would just split things off :)  But, if now isn't a convenient time to do so then yes, throwing it in here for testing purposes seems pragmatic.

mrocklin:
Hooray!

mrocklin:
I will be very glad to see the HDFS code live in its own repository.  Hopefully it will also reduce our testing infrastructure burden.

stuckeyr:
Ah, I was in the middle of testing, but things have gone awry. I checked out dask/dask#2881 and dask/distributed#1554 but must have made a mistake as I now get:
```
NotImplementedError: Using pyarrow with a 'DaskHDFileSystem' filesystem object
```
I might wait till @jcrist has set up the new package...

jcrist:
The dask filesystem and the pyarrow compatibility wrapper are both in https://github.com/dask/hdfs3/pull/144. Closing.

PR #1555, opened 2017-11-14
Remove explicit pycodestyle checks

jcrist:
As of the most recent release of flake8, pycodestyle is now called
directly by flake8. Remove explicit install and check call from ci.


pitrou:
Thanks :-)

Issue #1002, opened 2017-04-06
Distributed memory management plugin


mrocklin:
It would be useful to have a plugin that tracked and managed distributed memory.

Currently results are copied around to various workers as necessary for computation.  Our current policy is to avoid movement when possible and keep duplicated data around until it is no longer necessary with the thought that data that ends up being duplicated organically is likely in high demand.  This can currently be manually controlled by two synchronous operations:

1.  `rebalance(futures)`: takes a set of futures and tries to balance them around a set of workers so that all workers store roughly the same number of bytes
2.  `replicate(futures)`: Change the number of replicas for results.  This is commonly used to spread around commonly needed intermediate results efficiently (this performs a tree-scatter) or to clean up replicated data (n=1) and increase free space

Currently people do this by hand.  This is error prone for a number of reasons and generally rare among non-expert users. It would be interesting to consider a scheduler plugin that watched current memory usage and responded dynamically by moving data around in the background.  Operations like `rebalance`, and `replicate` might then set objectives for this dynamic system to accomplish rather than perform the operation explicitly.

There are a few objectives when moving memory around:

1.  Respect replication constraints set by users (came up in https://github.com/dask/distributed/issues/3184)
2.  Redistribute data evenly across workers
3.  Free up memory, particularly when we have highly replicated data
4.  But still respect replication, particularly when it is cheap or frequently used

I think that this would improve memory usage when we're constrained.  I think that it shouldn't impact task scheduling much if we act in a safe manner like verifying that copied data has been moved to one location before deleting it from its origin, etc..  It may be challenging to do this efficiently in linear time, but we could also punt if we notice that the scheduler is otherwise busy.

sjperkins:
I have a scenario where the `align` functionality  would be useful. Briefly, I compute a function taking many, large, chunked inputs and producing a single [Chi-squared value](https://en.wikipedia.org/wiki/Chi-squared_test) output. Chi-squared's of each individual chunk are summed together to form the final Chi-squared value. All my inputs are stored as dask arrays in an xarray Dataset.

I *think* that I'm observing a lot of data transfer of the large input chunks (and spilling to disk when workers start running out of memory), mostly because the inputs to each individual function are not automatically aligned on each worker.

As mentioned above, I'm considering persisting each input chunk by hand to avoid data movement and spillage.

I had a couple of thoughts on this:

[...]

PR #1559, opened 2017-11-15
Group tasks into partitions which all execute on the same worker

sjperkins:
This is a very basic implementation of an idea that I think would be useful. I'm creating the PR to create discussion and decide whether it would be useful to take it further.

Worker restrictions on `Client.get`, `Client.compute`, `Client.persist` etc. are hard physical restrictions.

It would be useful to allow some flexibility and instead group tasks together into a partition which all execute on the same worker. This would give the scheduler flexibility when choosing which worker to schedule the first task in the partition. Once the first task is scheduled, all tasks within that partition would then have to execute on that worker, ensuring data locality.

 In the following pseudo-code for e.g. A, B and C each have four chunks of size 250. 

```python
A = dask.array<ones, shape=(1000,), dtype=float64, chunks=(250,250,250,250)>
B = dask.array<ones, shape=(1000,), dtype=float64, chunks=(250,250,250,250)>
C = dask.array<ones, shape=(1000,), dtype=float64, chunks=(250,250,250,250)>

D = A + B + C
```

When submitting the task graph representing D, it would be useful to assign the first chunk in each array to a partition, the second to another and so on, to ensure data locality.

At present, this is specified as *another* restriction in `Client.get`

```python
f = client.get({'x' :1, 'y':2}, ['x', 'y], partitions={'x':'a', 'y':'a'})
```

I would like to fold the `partitions` functionality into `worker_restrictions` but the worker/hostname handling is sticky and I didn't want to mess with it too much. Something like `workers=['Alice', 'tcp://123.456.789.10.11.12:8080/', 'parta']`, where `parta` is treated as a partition if no matching worker name or host name is matched.

I could also see this being renamed to `task_group` for e.g.

I'd welcome your input on this -- I'd imagine integrating this with the rest of the scheduling functionality would be tricky.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1071,6 +1071,7 @@ class Client(Node):
         key = kwargs.pop('key', None)
         pure = kwargs.pop('pure', True)
         workers = kwargs.pop('workers', None)
+        partitions = kwargs.pop('partitions', None)
         resources = kwargs.pop('resources', None)
         retries = kwargs.pop('retries', None)
         allow_other_workers = kwargs.pop('allow_other_workers', False)
@@ -1110,7 +1111,8 @@ class Client(Node):
         futures = self._graph_to_futures(dsk, [skey], restrictions,
                                          loose_restrictions, priority={skey: 0},
                                          resources={skey: resources} if resources else None,
-                                         retries={skey: retries} if retries else None)
+                                         retries={skey: retries} if retries else None,
+                                         partitions={skey: partitions} if partitions else None)
 
         logger.debug("Submit %s(...), %s", funcname(func), key)
 
@@ -1904,7 +1906,7 @@ class Client(Node):
 
     def _graph_to_futures(self, dsk, keys, restrictions=None,
                           loose_restrictions=None, priority=None,
-                          resources=None, retries=None):
+                          resources=None, retries=None, partitions=None):
         with self._lock:
             keyset = set(keys)
             flatkeys = list(map(tokey, keys))
@@ -1952,11 +1954,12 @@ class Client(Node):
                                      'resources': resources,
                                      'submitting_task': getattr(thread_state, 'key', None),
                                      'retries': retries,
+                                     'partitions': partitions,
                                      })
             return futures
 
     def get(self, dsk, keys, restrictions=None, loose_restrictions=None,
-            resources=None, sync=True, asynchronous=None, **kwargs):
+            resources=None, partitions=None, sync=True, asynchronous=None, **kwargs):
         """ Compute dask graph
 
         Parameters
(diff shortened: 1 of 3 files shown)


mrocklin:
My first impressions after reading your description, but not yet your implemenation, are that I'm curious if there are other ways to acheive this same behavior.  What is your final objective here?  Do you *need* these tasks to be executed on the same machines or are you just looking to mostly-align datasets for more efficient computation in the future.  

For example, another possible solution to this would be for the task scheduler to do a little bit of look-ahead when assigning tasks to workers and try to see which tasks might be related in the future.  This isn't necessarily easier than the approach that you have laid out here, but does demonstrate that there might be other solutions worth investigating.

mrocklin:
This problem might also be resolvable by using diamond fusion, the `ave_width=` keyword in [fuse](http://dask.pydata.org/en/latest/optimize.html#dask.optimize.fuse)

Something like 

```python
with dask.set_options(fuse_ave_width=5):
    ...
```

This would lump the related tasks in A, B, and C into single tasks and just compute D directly.  You would lose parallelism, and wouldn't have access to A, B, and C afterwards, but it would probably work well today.

sjperkins:
@mrocklin Thanks for nudging me towards the optimisation features. You're right, I can probably get 90% of the way by tending the graph. I'd glanced over them at some point, but never in depth.

I still think there's some intersection between what `align` in https://github.com/dask/distributed/issues/1002 and task partitioning could achieve. I think one of the pros of partioning is that the alignment/grouping is specified at graph/task submission, rather than as a secondary step.

mrocklin:
> I still think there's some intersection between what align in #1002 and task partitioning could achieve. I think one of the pros of partioning is that the alignment/grouping is specified at graph/task submission, rather than as a secondary step.

Yes, I agree, and I see value in this.  However anything this deep in the scheduler inevitably becomes fairly expensive to do correctly.  For example what happens if the worker goes down, what happens if two tasks in the same partition run at the same time on different workers, how does this interact with work stealing, how will this react with future features, and so on.  This isn't to say that we shouldn't pursue this, just that it'll be tricky.

Also, I'm hopeful that in some moderate future we might be able to solve this in a more automatic way that doesn't require as much input from the user.  

One approach here would be to use task priorities provided by [dask.order](https://github.com/dask/dask/blob/master/dask/order.py) more carefully.  The `dask.order` function  currently assigns a numeric priority to tasks that is actually probably pretty close to how you want them (the same chunks in A, B, and C will all have very close priorities).  When we get a new batch of tasks we assign tasks somewhat randomly among the idle workers.  If we were to change this allocation of tasks just after `update_graph` then this might achieve most of what you want without much cost.  We would be compteting against other interests though, like tasks with actually different priorities.

mrocklin:

PR #1561, opened 2017-11-15
Move hdfs functionality to hdfs3

jcrist:
Move the hdfs & dask integration out of distributed.hdfs. This removes
the need for managing hdfs infrastructure in the distributed tests, and
consolidates related code.

See https://github.com/dask/hdfs3/pull/144.


jcrist:
I'm a bit baffled by why the tests are now failing, and am unable to reproduce locally. None of it looks related to changes I've made here. Thoughts @mrocklin?

mrocklin:
No thoughts.  It could be that something changed upstream.  I would try pushing a trivial change and making a PR to test master.  

Have you tried a fresh testing environment with updated dependencies?

jcrist:
This took forever to track down: https://github.com/travis-ci/travis-ci/issues/8770. For some reason our tests fail when run in a container, but not on a VM. Looks to be networking related. The issue I reported to travis was masking this implicit environment change, and made debugging trickier. For now I've set `sudo: required`, which fixes things.

mrocklin:
@pitrou might find that interesting

pitrou:
@jcrist can you remove the sudo line and add `-x` to the pytest arguments so that we see what the `test_comms` failure is?

.travis.yml
```diff
@@ -1,16 +1,12 @@
 language: generic
-sudo: false
+sudo: required
```
pitrou: Can you add a comment referencing https://github.com/dask/distributed/pull/1563?

jcrist: Done

.travis.yml
```diff
@@ -1,16 +1,14 @@
 language: generic
-sudo: false
+# sudo shouldn't be required, but currently tests fail when run in a container
+# on travis instead of a vm. See https://github.com/dask/distributed/pull/1563.
```
QuLogic: The tornado PR linked from this one was closed as it appeared that Travis was fixed; should this workaround be reverted?

jcrist: No, the workaround is still needed. That PR was closed as we were done debugging.

QuLogic: Not the distributed PR, the [tornado one](https://github.com/tornadoweb/tornado/pull/2199).

jcrist: Looks like travis no longer has containerized environments, so I suspect that this could be removed. However, since the `sudo` setting no longer changes the build time it's not really a high priority to change it. Feel free to submit a PR if you want to.

PR #1566, opened 2017-11-17
Fix key_split for tuple keys

kelvyang:
We use tuples as keys in our computations, but we don't use hyphens. This causes the bokeh web interface to display weird names. I believe it's because the bokeh messages are all of type `str`, so it's doing `key_split(key)` on a `key` that looks like a `tuple` but is really a `str`.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -481,6 +481,8 @@ def key_split(s):
     'x'
     >>> key_split("('x-2', 1)")
     'x'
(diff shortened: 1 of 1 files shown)


mrocklin:
This makes sense to me and I think that this is the right place to put this change.  It's nicely protected to only occur in the tuple case.  Merging

Thanks @kelvyang !

PR #1570, opened 2017-11-18
Wrap 'import ssl' statements with try-except block 

### LINKED ISSUE #1569: Catching ImportError for ssl-less environments?
Hi Matthew, dask team,

I've been digging into dask recently and am super impressed. So, thanks for all the work you guys are putting into these awesome libraries. I was amazed at how easy it was to get setup.

I've been looking at ways to use dask in a project at work where SSL is intentionally disabled due to some federal compliance requirements. In this case, I have no SSL module (despite it being in the standard lib).

I noticed that tornado has a pattern to catch ssl import errors:

```
try:
    import ssl
except ImportError:
    # ssl is not available on Google App Engine.
    ssl = None
```
(https://github.com/tornadoweb/tornado/blob/0e869e3f4735685cb2460aac499b509e6f5cc58a/tornado/tcpserver.py#L32)

In this case, adding such a change would allow for us to use distributed despite not having access to the ssl module. If you think this is a reasonable request, I'd love to make a PR. If not, we can simply fork distributed since it is such a small change, however, I thought it wouldn't hurt to ask in case you find this small change amenable.

Best,
Xander

metasyn:
For ssl-crippled environments. Fixes #1569

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -3,9 +3,13 @@ from __future__ import print_function, division, absolute_import
 import errno
 import logging
 import socket
-import ssl
 import struct
 import sys
+
+try:
+    import ssl
+except ImportError:
+    ssl = None
 
 import tornado
 from tornado import gen, netutil
--- a/distributed/security.py
+++ b/distributed/security.py
@@ -1,6 +1,9 @@
 from __future__ import print_function, division, absolute_import
 
-import ssl
+try:
+    import ssl
+except ImportError:
+    ssl = None
 
 from . import config
 
--- a/distributed/tests/test_security.py
+++ b/distributed/tests/test_security.py
@@ -1,8 +1,12 @@
 from __future__ import print_function, division, absolute_import
 
 from contextlib import contextmanager
-import ssl
 import sys
+
+try:
+    import ssl
+except ImportError:
+    ssl = None
 
 import pytest
 from tornado import gen
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -14,7 +14,6 @@ import re
 import shutil
 import signal
 import socket
-import ssl
 import subprocess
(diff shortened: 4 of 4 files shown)


mrocklin:
This is fine with me.  Lets wait a day to make sure that no one (like @pitrou) has any objections.

pitrou:
Sounds ok to me.

mrocklin:
Thanks @metasyn !  This is in.

jakirkham:
Late to the party here. This seems reasonable.

Though was a little surprised not to also see blocks of code like so.

```python
if ssl:
    ...
```

Are errors (e.g. `AttributeError`) already caught somehow if `ssl` is `None` or are those code paths just not traversed for you?

metasyn:
@jakirkham I thought about needing to do that that, but after a little testing, I was able to use the `parallel_backend` context manager with no issues with only the simple catches.

In `comm/tcp.py` - the ssl context is only needed when tls is being used (`_expect_tls_context`
In `security.py` - the ssl context is also only needed when tls is being used (`_get_tls_context`) - Its also used in the `get_connection_args` and `get_listen_args`, but doesn't seem like an issue:
```
        purpose = ssl.Purpose.SERVER_AUTH if hasattr(ssl, "Purpose") else None
```

So yes, they're not traversed for me if I'm not using tls. It also seems that having a crippled ssl module would just fail at a different point if I used tls (as some other error, rather than an import error). 

If ya'll think adding a more defensive check for ssl presence is needed, I'm happy to make an additional PR :)

jakirkham:
Thanks for the clarification. No follow up needed for me. Just curious was all.

If this does become a more common use case, we might want to convert exceptions into something more readable (e.g. "TLS cannot be enabled without `ssl` module."). Though am happy to see someone add that when it becomes more of an issue.

PR #1573, opened 2017-11-20
Avoid PeriodicCallback double-starts

mrocklin:


distributed/scheduler.py
```diff
@@ -506,6 +506,12 @@ def del_scheduler_file():
 
         self.loop.add_callback(self.reevaluate_occupancy)
 
+        def start_pcs():
```
pitrou: Perhaps this infrastructure should be in distributed/core.py?

mrocklin: I've moved this to core.py and applied that logic consistently across, scheduler.py, worker.py, and nanny.py

mrocklin:
Any other thoughts here @pitrou ?

PR #1579, opened 2017-11-22
Fix large-messages test

mrocklin:
This possibly fixes an intermittent failure described here
https://github.com/dask/distributed/issues/966#issuecomment-346228536

I was unable t reproduce the failure locally, so this is a blind shot.


pitrou:
We still need some kind of timeout if we don't want the test to hang, I think.

mrocklin:
Ah, I may have misunderstood the purpose of the timeout here

mrocklin:

PR #1583, opened 2017-11-22
DOC: note about when secede useful

stsievert:
It took me a while to parse the docstring for `secede`, and this helps resolve that (also, `get_client` is super useful!).

It looks like a function that has called `secede` may be moved to another compute node. I think that's adequately covered by the docstring.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2625,7 +2625,10 @@ def secede():
     """
     Have this task secede from the worker's thread pool
 
-    This opens up a new scheduling slot and a new thread for a new task.
+    This opens up a new scheduling slot and a new thread for a new task. This
+    enables the client to schedule tasks on this node, which is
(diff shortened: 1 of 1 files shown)


distributed/worker.py
```diff
@@ -2625,7 +2625,9 @@ def secede():
     """
     Have this task secede from the worker's thread pool
 
-    This opens up a new scheduling slot and a new thread for a new task.
+    This opens up a new scheduling slot and a new thread for a new task. This
+    enables the client to schedule another task while waiting for the
+    ``client.gather`` to finish.
```
mrocklin: The user isn't necessarily using `gather` here.  It might be best to generalize this comment a bit.

stsievert: How does that look? (sorry, it's not showing up in GitHub UI)

mrocklin:
Thanks @stsievert !

Issue #1517, opened 2017-11-01
Retire old HTTP servers


mrocklin:
I would like to retire the separate `http` Tornado server that currently serves JSON data about the scheduler and workers.  I don't think that these are commonly used.  We may in the future decide to integrate these routes (or perhaps better routes) into the single Bokeh server that is now growing its own simple routes.

This issue is here to collect counter-arguments or objections if any exist.

bluenote10:
Ah too bad, seeing this too late to raise objections :). 

Our Dask integration is relying on the HTTP JSON interface, and we are now looking for a work-around to migrate to the latest distributed version. Unless I'm missing something the only option now is to write a meta-package which

- offers a modified CLI
- contains the old HTTP server code
- adds the `HTTPService` instance to the `services` dictionary in the CLI before launching the scheduler.

Or is there a better alternative?

Out of curiosity, was there anything wrong with the old implementation, which motivated the removal? I assume that simply moving the old implementation as-is into the already complex Bokeh server would probably not be desired either. What do you think about extending the CLI so that user provided services can be added to the `services` dict. On first glance this could enable maintaining the HTTP interface as a third party addons.

mrocklin:
Lack of obvious use.  It was also falling behind.  If I were to build that again (which is a possibility) it was would be designed very differently.  

In your situation I would probably copy the code from previous versions and initialize them in a preload script: http://distributed.readthedocs.io/en/latest/setup.html#customizing-initialization

I'm curious, which parts of these servers were you using?

bluenote10:
We were using the `/tasks.json` endpoint for auto-scaling purposes.

PR #1584, opened 2017-11-23
Add JSON task counts route

mrocklin:
See conversation at the end of #1517 

cc @bluenote10 more contributions here would be welcome.  Hopefully this gives a good pattern that others can follow.

### DIFF
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -114,14 +114,31 @@ class CountsJSON(RequestHandler):
         self.write(response)
 
 
+class IdentityJSON(RequestHandler):
+    def get(self):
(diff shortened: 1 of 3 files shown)


bluenote10:
Great, that would be very helpful, thanks a lot!

Comparing this to the old implementation of `task.json` event handler (`def tasks(s)` in `diagnostics/scheduler.py`): Would it be possible to add something like `'waiting': len(self.server.waiting)` or is that no longer available?

mrocklin:
Sure, all of this is fairly easy.  I encourage you to take a look at the Scheduler constructor https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L210 to see what else might be around

PR #1586, opened 2017-11-23
Clear deque handlers after each test

pitrou:
This eases memory consumption a bit. Ideally we would instead have a new deque handler for each separate Worker or Scheduler instance...

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -489,6 +489,7 @@ class Client(Node):
 
         # Communication
         self.security = security or Security()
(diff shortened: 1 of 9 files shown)


mrocklin:
Associating a handler to a worker/scheduler makes sense to me if you prefer that solution.

pitrou:
> Associating a handler to a worker/scheduler makes sense to me if you prefer that solution.

Ok, I did that.

Issue #465, opened 2016-08-27
Error if joblib used within Executor.map (PicklingError)


stsievert:
I have a function I would like to run over many different inputs. I'm using `Executor.map` which provides the natural interface for this.

In the function I call with many different inputs, I call another function many times. In my case, this serves the purpose of obtaining some average performance. The natural interface to use is a joblib for-loop. See the code below for an example of what I'm describing.

My first thought to get around this is to use an example in the docs to have double mapped functions. However, I'd rather not further complicate my function definitions and have to pass `Executor.map` into `repeated_tests` below.

I have verified that the example in the docs of using distributed.joblib for the joblib interface worked.

``` python
def model_performance(x):
    """ returns 0 or 1 depending if model succeed or failed """
    time.sleep(2)
    return np.random.randint(2)

def repeated_tests(x, n=10):
    """ calls model_performance n times to get average performance """
    y = Parallel(n_jobs=-1)(delayed(model_performance)(x)
                            for _ in range(n))
    return sum(y) / n
    
with joblib.parallel_backend('distributed', scheduler_host=host):
    model_params = [1, 2, 3, 4]
    
    # normal
    a = map(repeated_tests, model_params)
    print(list(a))
    
    # dask.distributed
    b = e.map(repeated_tests, model_params)
    b = e.gather(b)
    print(b)
```

The error message I get is below, but the full stack trace can be found in [a gist]

``` python
PicklingError: Can't pickle <function model_performance at 0x7f0fac1c8950>: attribute lookup model_performance on __main__ failed
```

[a gist]:https://gist.github.com/stsievert/43391861410203ebbec6511ef3d0e4e4

mrocklin:
Yeah, you're implicitly trying to pickle up a connection to the scheduler, which will be tricky.  At first glance I see three options:

1.  Use normal, non-distributed joblib within `repeated_tests`.  Maybe start your workers with fewer threads to give joblib some more parallelism space.
2.  Open up an Executor within `repeated_tests()` or use the `with joblib.parallel_backend('distributed', ...)` context manager within `repeated_tests()`.  You'll have to pass the scheduler host information to that function.  It's a bit unconventional to have tasks in workers spin up short-lived Executors, but it should be ok.
3.  Just use the Executor to submit all the work up front:

    ```python
    a = [[e.submit(model_performance, x, pure=False) for i in range n]
          for n in model_params]
    b = e.gather(b)
    ```

stsievert:
[...]

PR #1587, opened 2017-11-23
WIP: DOC: adds note about client.compute in get_client

stsievert:
This is documentation of https://github.com/dask/distributed/issues/465#issuecomment-346450238.

Is `dask.delayed` responsible for `client.compute` doing "everything correctly" in https://github.com/dask/distributed/issues/465#issuecomment-346450238? That is, would `client.map(delayed(f), X)` do the same thing as `[client.compute(delayed(f)(x)) for x in X]` in terms of freeing up scheduling slots when `client = get_client()`? Clarifying this could help simply the docs.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2584,6 +2584,12 @@ def get_client(address=None, timeout=3):
 
     This client connects to the same scheduler to which the worker is connected
 
+    Submitting new jobs with the client returned by this function will likely
+    work best with ``client.compute``, especially if no computation can proceed
+    until task completion and the tasks take a significant amount of time.
+    However, calls to ``secede`` and ``rejoin`` `may` be more adventageous for
+    particular use cases.
+
     Examples
     --------
     >>> def f():
@@ -2601,6 +2607,7 @@ def get_client(address=None, timeout=3):
     get_worker
     worker_client
     secede
+    compute
     """
     try:
         worker = get_worker()


distributed/worker.py
```diff
@@ -2584,6 +2584,12 @@ def get_client(address=None, timeout=3):
 
     This client connects to the same scheduler to which the worker is connected
 
+    Submitting new jobs with the client returned by this function will likely
+    work best with ``client.compute``, especially if no computation can proceed
+    until task completion and the tasks take a significant amount of time.
+    However, calls to ``secede`` and ``rejoin`` `may` be more adventageous for
+    particular use cases.
```
mrocklin: Why is client.compute better than client.submit or client.map or client.persist here?  Two issues:

1.  I don't think that this is true, (I think that you're thinking of dask.compute) 
2.  This is likely very situation-specific.  People use `get_client` in many different situations.  I've found that it's hard to give generic advice like this and have it not confuse people in the future.

mrocklin:
@stsievert anything I can do to help resolve this?

distributed/worker.py
stsievert: > I think that you're thinking of dask.compute

Thanks for catching that, I typo'd client.compute.

> This is likely very situation-specific ... it's hard to give generic advice

Maybe the docs should list common uses you've seen? i.e., something like "Submitting jobs with this client has seen a couple common uses: (1) using dask.compute, which does ... and results in .... This is most applicable when X or Y." Then maybe a section on warnings or links to SO questions?

This might be better as a separate doc page.

stsievert:
@mrocklin I've replied on the comment thread above. Feel free to make edits to this PR too.

distributed/worker.py
mrocklin: Yes, moving this to a separate doc page might be nice.

mrocklin:
> @mrocklin I've replied on the comment thread above. Feel free to make edits to this PR too.

I'm unlikely to make changes here.  In general it's difficult to call out a particular case in a docstring.  My inclination here would be to close and think about how we can restructure the docs to provide common patterns here.  Unfortunately I'm unlikely to be of much help here near-term, but I would love it if you or anyone else took a crack at it.

stsievert:
> but I would love it if you or anyone else took a crack at it.

I'll try to make this happen. I'm planning on writing a blog post on  `get_client` and my application, and I'll include a section that can serve as (at least part of) the documentation.

> My inclination here would be to close and think about how we can restructure the docs to provide common patterns here.

My thoughts echo this.

mrocklin:
Grand.  Thank you for this.  

Also, I look forward to seeing what you've been working on.

stsievert:
> I look forward to seeing what you've been working on.

@mrocklin here's a blog post I wrote up on my use case: http://stsievert.com/blog/2018/01/04/dask-get-client/

mrocklin:
Cool blogpost.  If you're looking at adapting parameter selection it may be that you don't need `get_client` but could instead get away with `as_completed`.  

- Look at this section, the last code example in particular: http://dask.pydata.org/en/latest/futures.html#waiting-on-futures
-  http://matthewrocklin.com/blog/work/2017/04/19/dask-glm-2 (look for Asynchronous ADMM)

mrocklin:
Also https://github.com/eriknw/dask-patternsearch

stsievert:
> but could instead get away with `as_completed`

I could. I was looking into that, but discovered `get_client` along the way, which was cleaner for me.

> http://matthewrocklin.com/blog/work/2017/04/19/dask-glm-2 https://github.com/eriknw/dask-patternsearch

Thanks for the links. I was trying to get involved in the ADMM Dask work but couldn't find the time.

PR #1590, opened 2017-11-28
Fix #1589: Customize process title for subprocesses

### LINKED ISSUE #1589: Customize process title for subprocesses
When you have a bunch of dask subprocesses (such as workers), the reading of the `ps` command output is not very informative as it shows the multiprocessing command-line. It could be useful to instead set the process title to an informative string such as `dask-worker at tcp://127.0.0.1:1234`.

The [setproctitle](https://pypi.python.org/pypi/setproctitle) module provides the desired functionality.

pitrou:


pitrou:
For some reason this PR appears to make Travis-CI test runs quite slower... Needs investigation.

pitrou:
@mrocklin, I'm not sure how desirable you think it is. I personally find that it makes the `ps` output more useful when you have some Dask-launched subprocesses around.

mrocklin:
I think that your judgement here is probably better than mine regardless.  +1 from me

PR #1591, opened 2017-11-28
add dask-kubernetes to setup docs

mrocklin:
cc @martindurant


martindurant:
Good point.

I think it should strictly be Google Container Engine, whereas Google Compute is node provisioning, not necessarily for kubernetes.

mrocklin:
Updated, pointing to Google Kubernetes Engine.  This is what came up in a
quick web search.  Is Google Container Engine a better term?

martindurant:
Well darn gosh, they renamed it :)
The commands are still "gcloud container ...", but I guess they didn't want to mess up the API.

Issue #854, opened 2017-02-03
Accelerate Scheduler with Cython, PyPy, or C


mrocklin:
We are sometimes bound by the administrative of the distributed scheduler.  The scheduler is Pure-Python, and a bundle of core data structures (lists, sets, dicts).  It generally has an overhead of a few hundred microseconds per task.  When graphs become large (hundreds of thousands) this overhead can become troublesome.

There are a few potential solutions:

1.  Use Cython in a few places
2.  Run the entire scheduler in PyPy (workers, clients, and user code can still be in CPython)
3.  Rewrite everything in C/Go/Julia/whatever

Generally efforts here have to be balanced with the fact that the scheduler will continue to change, and we're likely to continue writing it in Python, so any performance improvement would have the extra constraint that it can't add significant development inertia or friction.

Here are a couple of cProfile-able scripts that stress scheduler performance: https://gist.github.com/mrocklin/eb9ca64813f98946896ec646f0e4a43b

mrocklin:
Here is another, more real-world benchmark: http://dask.pydata.org/en/latest/scheduler-choice.html

Though I plan to try to clean up performance on this one up a bit in pure python first.

kszucs:
Any experience running with PyPy? How much is the difference? 

I think it's a bit complicated to run the scheduler on pypy and the workers with cpython (if you want numpy or pandas). 

According to @GaelVaroquaux 's testimonial on Cython:

> You guys rock! In scikit-learn, we have decided early on to do Cython, rather than C or C++. That decision has been a clear win because the code is way more maintainable. We have had to convince new contributors that Cython was better for them, but the readability of the code, and the capacity to support multiple Python versions, was worth it.

With Cython it's easier to achieve c-like performance without leaving the python origins. I had great experience with it too.

mrocklin:
My perspective is the following:

The advantage of PyPy is that we would maintain a single codebase.  We would also speedup everything for free (including tornado).  I'm not normally excited about PyPy.  I think that dask's distributed scheduler is an interesting exception.  PyPy seems to be fairly common among web projects.  The distributed scheduler looks much more like a web project than a data science project.

The advantage of Cython is that users can install and run it normally using their normal tool chain.  Also the community that works on Dask has more experience with Cython than with PyPy.  The disadvantage is that we will need to maintain two copies of several functions to support non-Cython users, and things will grow out of date.  I do not intend to make Cython a hard dependency.

The advantage of C is that it would force us to reconsider our data structures.

[...]

PR #1594, opened 2017-11-29
Refactor scheduler to use TaskState objects rather than dictionaries

pitrou:
This refactors the scheduler to use TaskState, WorkerState, and ClientState objects rather than a forest of dictionaries.  It was originally planned in order to make Dask more amenable to acceleration with compilers like PyPy and Cython, but is also probably a more accessible organization for new developers.

See https://github.com/dask/distributed/issues/854


distributed/scheduler.py
```diff
@@ -239,78 +474,115 @@ def __init__(
         # Communication state
         self.loop = loop or IOLoop.current()
         self.worker_comms = dict()
+        # XXX rename to client_comms?
```
mrocklin: +1

distributed/scheduler.py
```diff
@@ -922,76 +1239,72 @@ def remove_worker(self, comm=None, address=None, safe=False, close=True):
...
 
             address = self.coerce_address(address)
             host = get_address_host(address)
 
+            ws = self.workers[address]
+
             self.log_event(['all', address], {'action': 'remove-worker',
                                               'worker': address,
-                                              'processing-tasks': self.processing[address]})
+                                              'processing-tasks': ws.processing})
```
mrocklin: It occurs to me that we should make a copy of the dict here (unrelated to this PR though).

distributed/scheduler.py
```diff
@@ -2422,61 +2830,59 @@ def transition_waiting_processing(self, key):
...
+                        recommendations[dts.key] = 'processing'
+
+            for dts in ts.dependencies:
+                s = dts.waiters
+                s.discard(ts)
+                if not s and not dts.who_wants:
+                    recommendations[dts.key] = 'released'
+
+            if not ts.waiters and not ts.who_wants:
+                # XXX what about 'fire-and-forget'?
```
mrocklin: Fire-and-forget should just be another client in `ts.who_wants`, no?

distributed/scheduler.py
```diff
@@ -69,6 +70,242 @@
...
+    Transform a set of worker states into a set of worker keys.
+    """
+    return {ws.worker_key for ws in workers}
+
+
+def _legacy_task_key_dict(task_dict):
+    """
+    Transform a dict of {task state: value} into a dict of {task key: value}.
+    """
+    return {ts.key: value for ts, value in task_dict.items()}
```
mrocklin: I suppose that these will eventually go away as you work through more of the peripheral modules?

distributed/bokeh/scheduler_html.py
```diff
@@ -93,23 +93,43 @@ def get(self, key):
...
     def get(self):
+        scheduler = self.server
+        erred = 0
+        nbytes = 0
+        ncores = 0
+        memory = 0
+        processing = 0
+        waiting = 0
+        waiting_data = 0
+        for ts in scheduler.task_states.values():
```
mrocklin: This is a little unfortunate.  I'm curious to see in what other situations we lose constant time diagnostics.

distributed/scheduler.py
pitrou: Ideally yes. Some though are used a lot in tests, so that will require a fair bit of manual fixing (unless there's a way to automate the changes).

distributed/scheduler.py
pitrou: Well, yes, but fire-and-forget doesn't require the task to remain alive.

distributed/bokeh/scheduler_html.py
pitrou: AFAIR, diagnostics and bokeh are the only places.

pitrou:
Just so that I don't forget it, a big update of the docs is needed.

mrocklin:
cc @fijal and @scoder for perspectives from PyPy and Cython respectively (if they're interested).  If you plan to look at this PR then I would start in the definition of `TaskState` in scheduler.py and then at the definition of `transition_processing_memory` as a representative function.

distributed/bokeh/scheduler.py
```diff
@@ -19,7 +19,7 @@
 from bokeh.plotting import figure
 from bokeh.palettes import Viridis11
 from bokeh.io import curdoc
-from toolz import pipe
+from toolz import frequencies, pipe
```
mrocklin: We should use cytoolz here for performance reasons.  

```python
try:
    from cytoolz import ...
except ImportError:
    from toolz import ...
```

distributed/bokeh/scheduler.py
```diff
@@ -109,13 +109,15 @@ def __init__(self, scheduler):
...
         with log_errors():
             s = self.scheduler
-            d = {'Tasks': [len(s.tasks)],
-                 'Stored': [len(s.who_has)],
-                 'Processing': ['%d / %d' % (len(s.rprocessing), s.total_ncores)],
-                 'Waiting': [len(s.waiting)],
-                 'No Worker': [len(s.unrunnable)],
-                 'Erred': [len(s.exceptions)],
-                 'Released': [len(s.released)]}
+            freqs = frequencies(ts.state for ts in s.task_states.values())
```
mrocklin: This could get ugly when we get to a million tasks (which we should treat as normal)

```python
In [1]: class A(object):
   ...:     pass
   ...: 

In [2]: objs = []

In [3]: for i in range(1000000):
   ...:     obj = A()
   ...:     obj.x = 'a'
   ...:     objs.append(obj)
   ...:     

In [4]: from toolz import frequencies

In [5]: %time _ = frequencies([o.x for o in objs])
CPU times: user 140 ms, sys: 0 ns, total: 140 ms
Wall time: 139 ms
```

So we might have to drop this diagnostic

mrocklin: Some of these numbers we can get from other state, right? 

    {'memory': scheduler.who_has, 'erred': scheduler.exceptions_blame, ...}

distributed/bokeh/scheduler.py
mrocklin: Or not, on the benchmark below it only helps by 20-30%, which still isn't good enough to continue with that diagnostic as is.

distributed/bokeh/scheduler.py
pitrou: That other state are just pseudo-mappings that query the internal `task_states`.
If the diagnostic is important, though, we can always maintain aggregates in the scheduler. It shouldn't be very costly (but, of course, always risks losing sync if there's a bug :-)).

pitrou: See https://github.com/dask/distributed/pull/1594/files#diff-6a3108098afcddf88b38652ae1e383b5R240 for what I mean.

pitrou: What do you think: should we keep running counts in the scheduler or should we remove those diagnostics?

mrocklin: I think that it depends on how many cases there are like this.  If this is the only one then lets just remove the diagnostic

mrocklin: In worst-case we can move this out to a counting plugin.

pitrou: There's `CountsJSON` and `StateTable`.

mrocklin: I recommend that we stop using `StateTable` and leave `CountsJSON` as you have it now.  `StateTable` is called periodically (500ms), `CountsJSON` is up to the user. 

cc @bluenote10 I believe that this affects you.  Note that this operation will start becoming more expensive.  If you have million+ node graphs then I recommend not calling this route more often than once a second.

mrocklin:
At some point we should also check the difference in memory use in the scheduler when there are about a million tasks.

pitrou:
I can't test with a million tasks, but here is a benchmark with 32k tasks:
https://gist.github.com/pitrou/68a306cbb4aaa2860e146aed76422935

Here I get (with git master):
```
* RSS: 358 MB
** Duration: 18.003 s. real, 15.780 s. CPU time
```
and with this PR:
```
* RSS: 434 MB
** Duration: 18.279 s. real, 15.762 s. CPU time
```
(both measurements with CPython 3.6, 64-bit, on Ubuntu 16.04 with 4 CPU cores)

So, while the runtime is similar, this PR clearly consumes more memory. Given the unloaded memory consumption seems about 100 MB, it means there is a ~30% increase in memory consumption (from 258 MB to 334 MB).

Another crude estimate is that, on this benchmark, git master consumes about 7800 bytes per task, and this PR about 10000 bytes per task (on the scheduler). This means around 10GB RSS would be needed for a million tasks *on this benchmark*.

Also note the baseline consumption of a minimal `TaskState` object (with empty dependencies, who_has, etc.):
```
>>> ts = scheduler.TaskState('foo', None)
>>> sys.getsizeof(ts)
224
>>> sys.getsizeof(ts) + sum(sys.getsizeof(getattr(ts, k)) for k in ts.__class__.__slots__)
1944
```

mrocklin:
Can you estimate how this might change with Cython or PyPy? 

Do you have any thoughts on how to reduce this overhead?

pitrou:
(note also that if up the task count to 128k, on this benchmark, there are phases where the scheduler seems to run all alone, perhaps busy stealing tasks? and there are messages such as "Event loop was unresponsive for 29.47s" displayed...)

pitrou:
I doubt PyPy or Cython would be able to improve memory consumption. There's a small opportunity for improvement by making the per-task sets optional (mainly `waiting_on`, `waiters`, `who_wants` and `has_what`), i.e. making them potentially None, but it would be at best a 10-15% improvement IMHO, and would complicate the code.

fijal:
Note that PyPy has way more compact objects - the memory can be significantly smaller here (at least for million tasks).

mrocklin:
@fijal if you're interested it seems like this branch is in a state where it might be useful to test against PyPy again.  Do you still have the time and interest in this?  Do you still have the benchmarks that we were using earlier?

Both I and @pitrou have tried things out separately.  I got some modest improvement over CPython but it was similar to what we were seeing previously.  If you're interested in giving this a shot I'd appreciate it.  If you have feedback on how to make this codebase more amenable to accelaration with PyPy we'd love to hear it.

fijal:
Hi Matthew

I do have interest, but absolutely not right now. I tried to get the
benchmark to run (it's a Python 3 benchmark so I had to do few tweaks) and
then it somehow didn't run. I doubt I'll have more time than 5 minutes in
the next couple weeks

Cheers,
fijal

pitrou:
> Note that PyPy has way more compact objects - the memory can be significantly smaller here

What would be the explanation? Smaller dicts and sets? I'm not sure object size really comes in play here. A bare TaskState object is 224 bytes (with slots) while the total consumption is around 10000 bytes per task.

mrocklin:
> I do have interest, but absolutely not right now. I tried to get the benchmark to run (it's a Python 3 benchmark so I had to do few tweaks) and then it somehow didn't run. 

Everything should be Python 2/3 compatible.  We test both continuously.

> I doubt I'll have more time than 5 minutes in the next couple weeks

Good to know.  Good luck!

mrocklin:
Thoughts on renaming `task_states` to `tasks`?  It's shorter and aligns with `.workers` and `.clients`.

mrocklin:
I've done some renaming within test_scheduler and test_client.  Yeah, that's not pleasant work.  It reminds me of long ago, when we used to unit-test the scheduler internals.  Every change to the internal machinery resulted in a painful test-rewriting process.

pitrou:
Perhaps the solution is to shield that internal state behind introspection APIs such as `Scheduler.get_task_status`? We could add more of them.

mrocklin:
> Perhaps the solution is to shield that internal state behind introspection APIs such as  Scheduler.get_task_status? We could add more of them.

So far my solution has been to try to avoid writing tests that explicitly check Scheduler state and instead check correct operation with Client API, which changes more slowly.  The first half of test_client was more painful than the second half.  

I don't think that there is any concrete action we should take at time time, I'm just thinking out loud.

mrocklin:
@pitrou in your opinion what remains to be done here?

Some questions:

1.  Are you satisfied with current scheduling state documentation?  If so then I may take a pass over it.
2.  Are there legacy mappings that we should add back in for a release or two?

pitrou:
> Are you satisfied with current scheduling state documentation? If so then I may take a pass over it.

Yes, but I'd appreciate your reviewing it.

> Are there legacy mappings that we should add back in for a release or two?

Yes, I removed some of them when I had finished suppressing all usage of them but that was probably a mistake.

> what remains to be done here?

Other than the above, nothing IMHO.

docs/source/scheduling-state.rst
```diff
@@ -30,385 +32,445 @@ system used to perform this tracking.
...
-   * ``(inc, 1)``: a tuple satisfying the `dask graph protocol
-     <http://dask.pydata.org/en/latest/graphs.html>`_.  This
-     again is stored serialized.
+In addition to the literal state, though, other information needs to be
+kept and updated about each task.  Individual task state is stored in an
+object named :class:`TaskState` and consists of the following information:
 
-   These are the values that will eventually be sent to a worker when the task
-   is ready to run.
+.. class:: TaskState
```
mrocklin: Would it be possible to move this documentation to the class docstring?

pitrou: I'll try to. Hopefully that will render correctly.

mrocklin:
> Yes, I removed some of them when I had finished suppressing all usage of them but that was probably a mistake.

If you have time to add these back in that would be helpful.

I'm proposing a micro-release by the end of the day.  After that I'm comfortable merging this.  I might take another pass through tests to see if there are other things to clean up.

pitrou:
> If you have time to add these back in that would be helpful.

I can do that on Monday.

mrocklin:
In the future if memory becomes an issue then we might consider using tuples or lists instead of sets for some of the TaskState attributes.  In the common case the number of dependencies/dependents can be quite small and our most common operation is iterating over the collection.

pitrou:
That's a possibility. Though if you look at the measurements above I wonder if other things come into play - 10 kB per task doesn't really match the basic size of a TaskState... and the sets were already there before (they were dict values rather than attribute values).

mrocklin:
Do you have any suggestions on how to determine the origin of memory costs?

pitrou:
First I would disable or minimize any sort of persistent logging (`transition_log` etc.). Then perhaps [tracemalloc](https://docs.python.org/3/library/tracemalloc.html) can help diagnose what is going on.

pitrou:
Btw, disabling stealing seems to make runtimes much more predictable and memory consumption lower as well.

pitrou:
If I monitor memory consumption just before and just after update_graph(), I get the following for 32767 tasks:
* before: 194 MB
* after: 342 MB (and that number doesn't really grow afterwards)

so each task actually costs 4.5 kB *in the scheduler itself* (notwithstanding stealing and other stuff).

More generally, I think any potentially costly operation could be wrapped in CPU and memory measurements, so that we have the option of logging what happens.

mrocklin:
OK to merge from me.

distributed/scheduler.py
```diff
@@ -722,104 +1239,148 @@ def update_graph(self, client=None, tasks=None, keys=None,
...
-                    self.exceptions_blame[key] = self.exceptions_blame[dep]
+        for ts in sorted(runnables, key=operator.attrgetter('priority')):
+            if ts.state == 'released' and ts.run_spec:
+                recommendations[ts.key] = 'waiting'
+
+        for ts in touched_tasks:
+            for dts in ts.dependencies:
+                if dts.exception_blame:
+                    ts.exception_blame = dts.exception_blame
                     recommendations[key] = 'erred'
```
jakirkham: Guessing `key` here should now be `ts.key`. Fixing in PR ( https://github.com/dask/distributed/pull/1900 ).

PR #1595, opened 2017-11-29
Fix error from get_client() with no global client

li-dan:
Fix TypeError from get_client() with no global client. This is supposed
to raise a ValueError.

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -4612,6 +4612,17 @@ def test_get_client(c, s, a, b):
         del distributed.tmp_client
 
 
+def test_get_client_no_cluster():
+    # Clean up any global workers added by other tests. This test requires that
+    # there are no global workers.
+    from distributed.worker import _global_workers
+    del _global_workers[:]
+
+    msg = 'No global client found and no address provided'
+    with pytest.raises(ValueError, match=r'^{}$'.format(msg)):
+        get_client()
+
+
 @gen_cluster(client=True)
 def test_serialize_collections(c, s, a, b):
     da = pytest.importorskip('dask.array')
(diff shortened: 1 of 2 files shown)


distributed/tests/test_client.py
```diff
@@ -4612,6 +4612,12 @@ def f(x):
         del distributed.tmp_client
 
 
+def test_get_client_no_cluster():
+    msg = u'No global client found and no address provided'
```
mrocklin: Seems fine to me.  I recommend dropping the unicode string prefix here.  The dask codebase defaults to Python 3.

li-dan: Okay, fixed.

mrocklin:
+1 from me.  Waiting for travis-ci

li-dan:
Some of the builds failed. It seems that `get_worker` is not raising `ValueError`. Is there a way to set up the test so that there are no global workers?

mrocklin:
> Some of the builds failed. It seems that get_worker is not raising ValueError. Is there a way to set up the test so that there are no global workers?

Hrm, that's unfortunate.  Two possible solutions come to mind

```python
def test_foo():
    from distributed.workers import _global_workers
    _global_workers.clear()
    # the rest of your code
```

Or, if we wanted to do this more holistically, we would modify the `gen_cluster` and `cluster` functions in `utils_test.py` to always clear this collection after every test.  *Almost* all tests operate with one of these functions, so this is a nice way to enforce hygiene across the test suite.

li-dan:
I've made the smaller change for now. Let's see if that fixes the builds.

mrocklin:
It looks like this fails on Python 2.  My apologies for the bad suggestion.  You probably want the following instead:

    del _global_workers[:]

li-dan:
I made the fix, and all builds are passing now.

mrocklin:
Awesome.  Thank you for following through with this @li-dan !  Merging.

Issue #1606, opened 2017-12-04
scheduler.retire_workers can mark tasks on retired workers as suspicious_tasks


rbubley:
For example, when Adaptive scales down workers, it calls `scheduler.retire_workers(remove=True, close_workers=True)`

`scheduler.remove_worker` has an argument, `safe=True` that would prevent this, and this is indeed used, but prior to this a call is made to `scheduler.close_worker`, which in turn calls `scheduler.remove_worker` without passing on the `safe=True` flag.

mrocklin:
Thanks for reporting.   A fix would be welcome here.

rbubley:
Closed via #1607

PR #1607, opened 2017-12-04
Don't mark tasks as suspicious when retire_workers called. Addresses #1606.

rbubley:


mrocklin:
Thanks!  The fix here looks great.  Can I also entice you to add a small test?  I recommend using a `gen_cluster` style test http://distributed.readthedocs.io/en/latest/develop.html#writing-tests

distributed/tests/test_scheduler.py
```diff
@@ -716,6 +716,7 @@ def test_retire_workers(c, s, a, b):
     workers = yield s.retire_workers()
     assert not workers
 
+    assert all(v==0 for v in s.suspicious_tasks.values())
```
mrocklin: This probably would have been this way regardless.  I think that we need a task to be actively running when we take down the worker.  In this case we just have data sitting there but no tasks running.

Maybe something something like the following?

```python
futures = c.map(slowinc, range(10), delay=0.05, workers=b.address)
future = c.submit(slowinc, 100, delay=0.5, workers=a.address, allow_other_workers=True)
yield gen.sleep(0.2)
yield s.retire_workers()  # I'm hoping that this closes worker a

assert all(v==0 for v in s.suspicious_tasks.values())
```

Although I suppose that if you're also going to stop `retire_workers` from closing workers that have running tasks then this might go away anyway?

rbubley: Yes and no. That change will make this go away in the sense that `retire_workers()` should no longer choose to retire worker a, but we can force this in this test with `yield s.retire_workers(workers=a.address)`. This is leaving open the possibility that `retire_workers` gets called by some other function either in the future or in some sub-class.

mrocklin: OK great, I had forgotten about the ability to force retirement.  Presumably the example above could be simplified to something like the following:

```python
future = c.submit(slowinc, 100, delay=0.5, workers=a.address, allow_other_workers=True)
yield gen.sleep(0.2)
yield s.retire_workers(workers=[a.address])

assert all(v==0 for v in s.suspicious_tasks.values())
```

distributed/tests/test_scheduler.py
```diff
@@ -717,6 +717,15 @@ def test_retire_workers(c, s, a, b):
     assert not workers
 
 
+@gen_cluster(client=True, timeout=1000)
+def test_retire_workers_no_suspicious_tasks(c, s, a, b):
+    c.submit(slowinc, 100, delay=0.5, workers=a.address, allow_other_workers=True)
```
mrocklin: You will need to hold on to this future.  Otherwise the task will be garbage-collected.

mrocklin: If flake8 complains at you then I recommend verifying that it actually computes after you retire `a` by yielding on it and checking the value.

distributed/tests/test_scheduler.py
```diff
@@ -717,6 +717,15 @@ def test_retire_workers(c, s, a, b):
     assert not workers
 
 
+@gen_cluster(client=True, timeout=1000)
```
mrocklin: Were you getting timeouts on this?

rbubley: No. I'll remove that timeout argument.

mrocklin:
Looks good to me.  Merging.  Thanks @rbubley !

Issue #1604, opened 2017-12-04
Adaptive can seek to scale up and down simultaneously


rbubley:
I observed in logs some occasions where adaptive scales down the cluster and then soon after, scales back up again. 

Looking at the logic for when scaling up and scaling down can happen: 
* scale down :not all workers are processing, when workers are idle and memory is not an issue
* scale up; and when scaling up can happen (by default when  total_occupancy > 2* total_cores

it becomes apparent that both of these are satisfiable simultaneously, e.g.

Three tasks: two with occupancy 4 and one with occupancy 1, running on 4 cores:

* Core 1 : occupancy 4
* Core 2 : occupancy 4
* Core 3 : occupancy 1
* Core 4 : occupancy 0

In this scenario, both the scale_up criteria and the scale_down criteria are satisfied.

(There will be more subtle cases where the scale_up condition only becomes satisfied after a scale_down operation is completed, of course.)

mrocklin:
Thanks for the well written issue.  Do you have any recommendations on how to refactor the Adaptive class?  Changes here would be welcome.

rbubley:
There are two things to do, I think:

1. Detect if we are trying to scale_up and scale_down simultaneously (and if so, do neither).
I think it could most easily be down by changing should_scale_down to return true only if scheduler.workers_to_close was non-empty. Then explicitly test in `_adapt` whether should_scale_up and should_scale_down are both true.

2. Change the logic so that this doesn't occur with the default parameters
I think this one will be covered by a subsequent fix to #1605.

mrocklin:
I believe that this has been resolved.  Closing.

PR #1608, opened 2017-12-04
Disallow simultaneous scale up and down in Adaptive. Fix for #1604

rbubley:


distributed/deploy/adaptive.py
```diff
@@ -188,14 +188,19 @@ def _adapt(self):
...
-                f = self.cluster.scale_up(**kwargs)
-                if gen.is_future(f):
-                    yield f
-
-            if self.should_scale_down():
-                yield self._retire_workers()
+            should_scale_up = self.should_scale_up()
+            should_scale_down = self.should_scale_down()
+            if should_scale_up and should_scale_down:
+                logger.info("Attempting to scale up and scale down simultaneously.")
```
mrocklin: Given that we know that this can occur in normal operation we should probably demote this to `logger.debug` level.  Otherwise users might think that there is something wrong with their system, rather than with our algorithms :)

mrocklin:
You had an example situation in your issue.  Perhaps that would make a good test here?

distributed/deploy/adaptive.py
rbubley: Maybe I'm overthinking it, but we should change the algorithm so that it *shouldn't* happen in normal operation. Seeing this message highlights that there is a flaw in the scaling algorithm being used, whether that be a subclass of Adaptive or the default version.

With regards to fixing the algorithm in Adaptive, so that we don't see it in normal use, this should mostly be fixed by #1609, although I think we may need a couple more changes to finish eliminate the corner cases: (these should probably go in a separate PR)

1. Don't scale up to have more cores than  there are tasks.
2. If we see should_scale_up and should_scale_down both True, try forcing a steal rebalance, and check again.

mrocklin: > With regards to fixing the algorithm in Adaptive, so that we don't see it in normal use, this should mostly be fixed by #1609, although I think we may need a couple more changes to finish eliminate the corner cases:

OK, if this is true then I agree that leaving this at INFO probably makes sense

rbubley:
If, given changes in other PRs it becomes hard to engineer a situation where `should_scale_up` and `should_scale_down` are both true, it seems a bit pointless to start up an adaptive cluster, force it to enter this sort of state, and then check that it logs this. Can you think of something more useful to test?

mrocklin:
You might verify that the workers that we start with are the same as the workers that we end with.  No unnecessary worker death or creation should have occurred.

rbubley:
I'm not convinced that this test actually works. @mrocklin, would welcome your thoughts - the whole of the testing framework is a little obscure to me.

mrocklin:
Thanks for the ping.  You should be using `gen_cluster` here, which will launch and clean up a cluster and client for you.  Check http://distributed.readthedocs.io/en/latest/develop.html#writing-tests for an example.

mrocklin:
Merging this in.  Thanks for all the recent activity @rbubley !

jakirkham:
Wondering if this fixes the problem described in issue ( https://github.com/dask/dask-drmaa/issues/41 ).

PR #1609, opened 2017-12-04
Do not include processing workers in workers_to_close. Fixes #1605

### LINKED ISSUE #1605: Adaptive can close active workers
I observed in logs some occasions where adaptive scales down the cluster and closes active workers, lengthening the overall computation time.

Looking at the logic for when scaling down can happen: 
* scale down :not all workers are processing, when workers are idle and memory is not an issue

It seems that the semantics of scheduler.idle are not what scheduler.workers_to_close assumes they are:

In the workers_to_close docstring, it says that it will only close workers that are idle (not running anything), yet the code shows that if there is any worker not running anything, it can close any worker marked as idle. 

Workers are marked as idle, however, are not just those workers with no processing: check_idle_saturated marks workers as idle also if they have less than half the average occupancy.

So, in this example:

Three tasks: two with occupancy 4 and one with occupancy 1, running on 4 cores:

* Worker 1 : occupancy 4
* Worker 2 : occupancy 4
* Worker 3 : occupancy 1
* Worker 4 : occupancy 0

In this scenario, both Workers 3 and 4 would be marked as idle and closed.

rbubley:


rbubley:
I will add a test, shortly.

mrocklin:
> I will add a test, shortly.

:)

rbubley:
I don't think that AppVeyor build failure is connected to this PR.

mrocklin:
Agreed.  I'll take a look.  Restarting that test

mrocklin:
This is in.  Thanks @rbubley !

PR #1610, opened 2017-12-04
Use environment variable for scheduler address if present

mrocklin:
This makes `Client()` use the `DASK_SCHEDULER_ADDRESS` if available.  Set
`Client(scheduler_env_variable=False)` to retain original behavior.

This ends up being useful when deploying on distributed systems.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -449,9 +449,6 @@ class Client(Node):
         Claim this scheduler as the global dask scheduler
     scheduler_file: string (optional)
         Path to a file with scheduler information if available
-    scheduler_env_variable: bool (optional)
-        If True then use DASK_SCHEDULER_ADDRESS for scheduler address if it
-        exists.  True by default.
     security: (optional)
         Optional security information
     asynchronous: bool (False by default)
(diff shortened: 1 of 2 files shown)


pitrou:
Is there a reason to add the `scheduler_env_variable` parameter instead of having people pass an `address` explicitly?

mrocklin:
In some deployments the scheduler address is known by an automated service setting up an environment, but not known by the user.  In this case a common way to pass information is for the deployment service to populate environment variables.

pitrou:
Understood, I was just asking why you added the `scheduler_env_variable` override.

mrocklin:
Ah I see, my mistake.  Behavior under this branch is that `Client()` connects to a remote client even if no address is given.  This may not always be preferred.  People may not want this magic, or they may want to create a local cluster.

pitrou:
I see. How about having `address='<from-env>'` be the default, and let people pass `address=None`?

mrocklin:
The `'address='<from-env>'` option seems odd to me because this isn't the common case.  Placing this as the default of the first argument seems too prominent.  I agree generally though, that the API choice here could be improved, I'm trying to think of an alternative.

mrocklin:
OK, my solution is just to remove this optoin for now.  If a problem arises then we can determine a convention then.  Instead I've added logging in order to alert the user.

mrocklin:
Merging this soon

PR #1611, opened 2017-12-04
[skip ci] Add Cloud deployment documentation

mrocklin:
This depends on a number of PRs currently in-flight

### DIFF
--- a/docs/source/setup.rst
+++ b/docs/source/setup.rst
@@ -201,6 +201,91 @@ See the :doc:`EC2 quickstart <ec2>` for information on the ``dask-ec2`` easy
 setup script to launch a canned cluster on EC2.
 
 
+Cloud
+-----
+
+It is easy to launch a Dask cluster and Jupyter notebook server on cloud
+resources using Kubernetes_ and Helm_.
+
+.. _Kubernetes: https://kubernetes.io/
+.. _Helm: https://helm.sh/
+
+
+Setup Cluster
+~~~~~~~~~~~~~
+
+First you will need to set up an account with one of Google, Amazon, or
+Microsoft Azure,  and launch a cluster running Kubernetes.  This is
+straightforward by following the instructions at `Zero to JupyterHub
+<http://zero-to-jupyterhub.readthedocs.io/en/latest/>`_ that focus on
+Kubernetes and Helm.  You do not need to follow all of these instructions.
+JupyterHub is not necessary to deploy Dask:
+
+- `Creating a Kubernetes Cluster <https://zero-to-jupyterhub.readthedocs.io/en/v0.4-doc/create-k8s-cluster.html>`_
+- `Setting up Helm <https://zero-to-jupyterhub.readthedocs.io/en/v0.4-doc/setup-helm.html>`_
+
+Detailed instructions are available for all major Cloud providers.
+
+
+Helm Install Dask
+~~~~~~~~~~~~~~~~~
+
+Now, you can launch Dask on your cluster using the Dask Helm_ chart::
+
+   helm install stable/dask
+
+This will also deploy a Jupyter server that you can use to launch computations.
+This will deploy your application on the cluster and print out instructions to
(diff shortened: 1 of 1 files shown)


docs/source/setup.rst
```diff
@@ -201,6 +201,91 @@ See the :doc:`EC2 quickstart <ec2>` for information on the ``dask-ec2`` easy
...
+
+.. _Kubernetes: https://kubernetes.io/
+.. _Helm: https://helm.sh/
+
+
+Setup Cluster
+~~~~~~~~~~~~~
+
+First you will need to set up an account with one of Google, Amazon, or
+Microsoft Azure,  and launch a cluster running Kubernetes.  This is
```
TomAugspurger: Although it isn't "Cloud", I found minikube helpful for testing things out locally before trying on Google. Both `helm install stable/dask-distributed` and `helm install stable/spark` worked for me locally.

docs/source/setup.rst
```diff
@@ -201,6 +201,91 @@ See the :doc:`EC2 quickstart <ec2>` for information on the ``dask-ec2`` easy
...
+   helm upgrade iced-mule stable/dask \
+     --set worker.pipPackages="gcsfs dask-ml" \
+     --set jupyter.pipPackages="gcsfs dask-ml"
+     --set worker.condaPackages="-c conda-forge xarray pyarrow" \
+     --set jupyter.condaPackages="-c conda-forge xarray pyarrow"
+
+Conda packages are installed before pip packages.  These values are appended
+directly to the ``conda install`` and ``pip install`` commands and so you can
+also include flags, such as specifying conda channels.
+
```
jcrist: There's another "Kubernetes" section on line 305 below, not sure if it should be removed or merged with this.

docs/source/setup.rst
mrocklin: Cool.  I wasn't aware of minikube.  Should we mention here in these docs or is this just an FYI?

mrocklin:
Current work on dask helm chart: https://github.com/kubernetes/charts/pull/2914

docs/source/setup.rst
mrocklin: Yeah, this and the dask-kubernetes section probably require some conversation to see what we want to do going forward

docs/source/setup.rst
```diff
@@ -201,6 +201,91 @@ See the :doc:`EC2 quickstart <ec2>` for information on the ``dask-ec2`` easy
...
+   helm upgrade iced-mule stable/dask \
+     --set worker.pipPackages="gcsfs dask-ml" \
+     --set jupyter.pipPackages="gcsfs dask-ml"
+
+You can also install conda packages alongside pip packages.  For example we
+might want to conda-install XArray and Arrow::
+
+   helm upgrade iced-mule stable/dask \
+     --set worker.pipPackages="gcsfs dask-ml" \
```
martindurant: gcsfs is now on conda-forge too (here and above)

martindurant:
cc @danielfrg

docs/source/setup.rst
TomAugspurger: Not sure... I think a parenthetical "or minikube for local testing" may be worthwhile to ease the burden of entry for people looking to test things out. It seems to be the officially blessed version of "local kubernetes", so I don't see much downside.

docs/source/setup.rst
mrocklin: I think that the decision here probably depends on how smooth we can make the Helm solution.  Right now I don't think it's sufficiently smooth, but perhaps we can get it there?

jcrist: The section I'm talking about is below the Google Cloud section, and looks like it lists some previous (and possibly outdated?) work.

docs/source/setup.rst
mrocklin: My intent was to show off both.  Personally I use pip for gcsfs just for speed.  Conda can take a while.

PR #1615, opened 2017-12-06
Parse bytes given in --memory-limit

mrocklin:
This allows users to pass values like 5000MB instead of 5e9


adamklein:
This broke my `LocalCluster(memory_limit=None)` code. What's the right way to do this now?

mrocklin:
Fixed in https://github.com/dask/distributed/pull/1639

PR #1616, opened 2017-12-07
Fix deprecation warning from logger.warn()

bnaul:
Had to cancel a job because I was getting inundated with:
```
/home/brettnaul/miniconda3/lib/python3.6/site-packages/distributed/worker.py:2232: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  format_bytes(self.memory_limit))
```
Think this should be fine but feel free to close if it isn't backwards compatible for some reason.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -154,11 +154,11 @@ class Server(object):
         diff = now - self._last_tick
         self._last_tick = now
         if diff > config.get('tick-maximum-delay', 1000) / 1000:
-            logger.warn("Event loop was unresponsive for %.2fs.  "
-                        "This is often caused by long-running GIL-holding "
-                        "functions or moving large chunks of data. "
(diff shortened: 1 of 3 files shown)


mrocklin:
Fine by me

jakirkham:
AppVeyor failure looked unrelated. So restarted it.

mrocklin:
Yeah, we've been having an increasing number of issues with Appveyor (our fault, obviously).  Thanks for restarting.  I'm trying to reduce these.

Merging!

Thanks @bnaul

jakirkham:
Maybe so. Though I don't think this one was our fault. Seems to reference something about JUnit, which I'm pretty sure we don't use. In case you are curious, linked the problematic log below.

ref: https://ci.appveyor.com/project/daskdev/distributed/build/1.0.2700

PR #1618, opened 2017-12-08
Add min and max worker flags to adaptive cluster

jacobtomlinson:
### Description
Add min and max worker flags to adaptive cluster.

We've been using adaptive clusters on Kubernetes (over EC2) and have found an initial spin up time before tasks start calculating due to the first couple of scale up calls taking time to complete.

This PR adds a minimum number of workers to keep alive in the adaptive cluster meaning that tasks start processing immediately.

We've also added a maximum number of workers while we were at it.

#### Example usage
```
dask-scheduler --port 8786 --bokeh-port 8787 --min-workers 10 --max-workers 100 --preload /opt/scheduler/adaptive.py
```

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -61,10 +61,14 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
               help="Directory to place scheduler files")
 @click.option('--preload', type=str, multiple=True,
               help='Module that should be loaded by each worker process like "foo.bar" or "/path/to/foo.py"')
+@click.option('--min-workers', type=int, default=0,
+              help="Minimum number of workers required in an adaptive cluster")
+@click.option('--max-workers', type=int, default=1e100,
+              help="Maximum number of workers required in an adaptive cluster")
 def main(host, port, bokeh_port, show, _bokeh,
          bokeh_whitelist, bokeh_prefix, use_xheaders, pid_file, scheduler_file,
          interface, local_directory, preload, tls_ca_file, tls_cert,
-         tls_key):
+         tls_key, min_workers, max_workers):
     enable_proctitle_on_current()
     enable_proctitle_on_children()
 
@@ -118,7 +122,8 @@ def main(host, port, bokeh_port, show, _bokeh,
                                                       prefix=bokeh_prefix)
     scheduler = Scheduler(loop=loop, services=services,
                           scheduler_file=scheduler_file,
-                          security=sec)
+                          security=sec, min_workers=min_workers,
+                          max_workers=max_workers)
     scheduler.start(addr)
     preload_modules(preload, parameter=scheduler, file_dir=local_directory)
 
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -118,7 +118,8 @@ class Adaptive(object):
         needs_memory
         """
         with log_errors():
-            if self.scheduler.unrunnable and not self.scheduler.ncores:
+            if len(self.scheduler.ncores) < self.scheduler.min_workers or \
+                    (self.scheduler.unrunnable and not self.scheduler.ncores):
                 return True
 
             needs_cpu = self.needs_cpu()
@@ -177,7 +178,9 @@ class Adaptive(object):
         --------
         LocalCluster.scale_up
         """
-        instances = max(1, len(self.scheduler.ncores) * self.scale_factor)
+        instances = max(1, self.scheduler.min_workers,
+                        len(self.scheduler.ncores) * self.scale_factor)
+        instances = min(instances, self.scheduler.max_workers)
         logger.info("Scaling up to %d workers", instances)
         return {'n': instances}
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -217,6 +217,8 @@ class Scheduler(ServerNode):
             validate=False,
             scheduler_file=None,
             security=None,
+            min_workers=0,
+            max_workers=1e100,
             **kwargs):
 
         self._setup_logging()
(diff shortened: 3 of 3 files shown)


mrocklin:
Thanks for this, I'm glad to see people experimenting with these policies on Kubernetes.  Two main comments regarding the code here:

1.  This could use tests
2.  It would be good to avoid modifying the scheduler directly.  We've intentionally kept the Adaptive code separate from the Scheduler code and it would be useful to keep it that way in the short term.  I expect other Adaptive strategies to arise and we can't have all of them adding attributes to the scheduler.  Do you think it would be possible to put the min/max_workers values on the Adaptive class and specify them in your preload script?

For Kubernetes in particular I've been thinking that maybe the right thing to do here isn't to scale-up/scale-down but rather to set a target number of workers given what we know about the workload.  For example, we can look at `Scheduler.total_occupancy` and get a pretty good estimate of how many CPU-seconds we need.  This is larger in scope that this PR, but might be worth considering.

niallrobinson:
Hi Matt (I co-wrote the PR with @jacobtomlinson ),

1) That's fair - we thought you might say that
2) We agree, but without a fairly significant refactor, we could see how to do this, as a lot of the logic about add/removing nodes is contained in the Scheduler. (I guess this is what you're driving at with "larger scope than this PR")

(btw I'm presenting some of this stuff at AGU on Monday, if you're about - talk called "Jade: using on-demand cloud analysis to give scientists back their flow")

rbubley:
I've also been looking at extending Adaptive, and there are various strategies for which `scheduler.workers_to_close()` is too inflexible. 

Perhaps this function itself should be moved into the Adaptive class, as it that seems to be its only function? This would mean that future improvements in Adaptive strategies would be less likely to need to touch the scheduler.

jacobtomlinson:
I agree with @rbubley. 

We did start with `min_workers` and `max_workers` in the adaptive class but found ourselves basically reimplementing `workers_to_close` as it wasn't flexible enough for what we were trying to do.

mrocklin:
I've raised a separate issue for adaptive kubernetes clusters here, just in case this conversation extends beyond the lifetime of this PR.  https://github.com/dask/distributed/issues/1619

mrocklin:
> We did start with min_workers and max_workers in the adaptive class but found ourselves basically reimplementing workers_to_close as it wasn't flexible enough for what we were trying to do.

Perhaps we can improve `workers_to_close`?  What was restrictive?

jacobtomlinson:
Replying from mobile.

It was the fact that we don’t want to scale down if removing the list
returned from `workers_to_close` brings us below our minimum. We can check
thy logic fine in the `should_scale_down` method and remove some from the
list to avoid going below the minimum. However then when retiring the
`workers_to_close` is called again by `retire_workers` and gets the list
again.

mrocklin:
I would be fine with a `minimum` keyword in the `retire_workers` method if that would solve things.

rbubley:
A slightly more general way would be to have a list of do_not_retire workers.

If you only care about a numeric minimum you could just select some arbitrarily, but in some cases there may be distinguished workers that you want to keep alive: perhaps you get a discount if the same worker stays alive for a long time, or perhaps these workers are using harvested time from computers doing something else, etc.

This could be achieved with a function parameter, but could also be achieved by using the existing worker resources functionality - which in some ways might make more sense: they could declare themselves with a resource of “always-on”, or something.

mrocklin:
Have you had concrete experience with systems that need this functionality or is this more hypothetical?

rbubley:
I wouldn't say 'need': it's a want. Both of the examples I gave are live issues on our roadmap.

asford:
Would you mind looking at #1632 and checking if the modifications made in that PR could be used in the resolution of this issue?

It would, at a high level, move the question of adaptive scaling behavior to `Adaptive` subclasses as  previously suggested by providing an override hook in `Adaptive` but maintain compatibility with the current scheduler API.

As @rbubley suggested, it seems a *touch* strange that this logic lives in the scheduler but I can see a compelling argument for providing basic scale-down behavior in the scheduler interface, as is currently provided by `workers_to_close` and `retire_workers`. Adding the ability to hook into and override this behavior via `Adaptive` seems preferable to extending the core scheduler class with arbitrarily complex scaling logic.

jacobtomlinson:
@asford I see #1632 as a complimentary PR to this one. I need to make some changes to this one anyway before it can be merged, however I get the feeling progress will be slow this week with AGU.

asford:
It seems that one of the primary reasons this change touches `scheduler.py` is to provide support for command-line configuration for your autoscaling logic. I've run into the same issue, but don't think it's a great idea to overload `scheduler` as a configuration container and/or rely on modification of the `dask-scheduler` CLI.

Would a change allowing pass-through arguments to modules loaded via `--preload` be a workable solution for you? I'd envision this functioning in a manner similar to a compiler, in which subtool-specific flags may be specified on the command line. For example, `clang -Xlinker --foo=bar` passes `--foo=bar` as arguments to `ld`. In this case your interface might look like:

````
dask-scheduler --preload scheduler_autoscaling.py -Xpreload --min-workers=2 -Xpreload --max-workers=10
````

Where `scheduler_autoscaling.py` is a application specific plugin you've written that handles your scaling logic.

mrocklin:
Passing through keyword arguments to the preload script makes sense to me.  In general I'm usually in favor of improving the functioning of escape hatches like preload scripts.

asford:
@jacobtomlinson Just to clarify, I think this PR and #1632 are compatible and I would be stoked to see all of this logic land somewhere in `distributed` together.

jacobtomlinson:
@asford Yes I think that would be a good option. Let me know when it's in and I'll update this PR to make use of it.

jacobtomlinson:
@jhamman yes keen to revisit if @mrocklin is still happy with the proposal

mrocklin:
In general I agree that adding min/max workers to Adaptive is a good idea

mrocklin:
I've implemented min/max in Adaptive here: https://github.com/dask/distributed/pull/1797

Note that this doesn't pass anything through the scheduler.  This just focuses on the Adaptive piece in isolation.

jhamman:
I suppose this can be closed now?

mrocklin:
Thanks for starting this @jacobtomlinson !

PR #1623, opened 2017-12-11
[EXP] Cythonize scheduler state

pitrou:
This is an experiment into Cythonizing basic parts of the scheduler:
* individual state classes become C extension classes
* attributes are type-annotated
* interesting local variables in transition functions are type-annotated


pitrou:
So, the current result for this work:
* ~0% change on a Dask scheduler benchmark
* ~0% on `client.ClientSuite.time_trivial_tasks` in the distributed benchmarks suite

Quite underwhelming.

pitrou:
I also encountered several Cython issues along the way:
* https://github.com/cython/cython/issues/2038 (means it can be difficult to test the pure Python and Cython-compiled version at the same time)
* https://github.com/cython/cython/issues/2039
* https://github.com/cython/cython/issues/2040 (perhaps lost optimization opportunities)
* https://github.com/cython/cython/issues/2042 (perhaps lost optimization opportunities)

distributed/schedulerstate.py
```diff
@@ -0,0 +1,1419 @@
...
+        'who_wants',
+        'exception',
+        'traceback',
+        'exception_blame',
+        'suspicious',
+        'retries',
+        'nbytes',
+        )
+    if cython.compiled:
+        key = cython.declare(object, visibility='public')
```
scoder: You can also spell this `cython.declare(key=object, visibility='public')`. Avoids an assignment and might be more convenient.

distributed/schedulerstate.py
```diff
@@ -0,0 +1,1419 @@
...
+        workers = cython.declare(object, visibility='public')
+        idle = cython.declare(object, visibility='public')
+        saturated = cython.declare(set, visibility='public')
+        total_occupancy = cython.declare(cython.double, visibility='public')
+
+    def __init__(self):
+        self.tasks = dict()
+        self.unrunnable = set()
+        self.task_duration = {prefix: 0.00001 for prefix in fast_tasks}
+        self.unknown_durations = defaultdict(set)
```
scoder: You have to declare all attributes for a cclass. Suggests that there is probably something not working here.

distributed/schedulerstate.py
```diff
@@ -0,0 +1,1419 @@
...
+            if send_worker_msg:
+                self.worker_send(w, send_worker_msg)
+
+    @cython.ccall
+    @cython.locals(ts=TaskState, ws=WorkerState, dts=TaskState, cs=ClientState)
+    def _add_to_memory(self, ts, ws, recommendations, type=None):
+        """
+        Add *ts* to the set of in-memory tasks.
+        """
+        if self.validate:
```
scoder: This attribute seems undefined.

pitrou:
I tried again with Cython git master (which solves https://github.com/cython/cython/issues/2040 and https://github.com/cython/cython/issues/2042) and don't seem any significant difference, unfortunately.

distributed/schedulerstate.py
pitrou: Actually, it's defined on `Scheduler` which derives `BaseSchedulerState`.

scoder: If a base class uses it, why would a subclass define it?
Anyway, just be aware that Cython has to generate generic Python attribute access code in this case. You can easily see that with `cython -a`.

pitrou: Yeah, I understand that. This PR is an experiment, the goal is not to write well-architected code, but to judge how much Cython can improve performance on critical routines.

scoder:
Did you run `cython -a` to see how Cython interprets your critical code? It could well be, for example, that it isn't aware of some of the extension types in the code that processes them (outside of the classes themselves), or something like that.

And, sorry for asking the obvious (especially asking *you* the obvious), but since you tried a couple of things now and found that the effect of compilation was really zero, are you sure that your benchmarks are really CPU-bound, that they test realistic scenarios, and that the code you are compiling is really the critical code? It happens from time to time that compilation makes no difference whatsoever, but it usually hints at this: 1a) people aren't actually compiling the code that dominates the runtime, and/or 1b) the time is spent mostly inside of data structure operations (CPython, NumPy, ...) and not in their own code that they are trying to speed up.

pitrou:
> Did you run cython -a to see how Cython interprets your critical code?

No, but I looked at the generated C code. It seemed reasonable to me (i.e. attribute accesses are de-virtualized and concrete C API functions are used where expected). Of course, I only skimmed through the thousands of lines of generated C code...

> are you sure that your benchmarks are really CPU-bound, that they test realistic scenarios, and that the code you are compiling is really the critical code?

They are CPU-bound indeed. They are not realistic in themselves, more like a stress test; i.e. there *may* be bottlenecks elsewhere in some real-world situations (but, given the use cases for distributed, it is probably hopeless to try and model a real-world workload as a reasonably short to run benchmark).

The code I compiled is "more or less" the critical code... but runtime is spread over many more functions (i.e. there isn't a single well-identified bottleneck). I would not expect Cython to save more than 30% (and that would already be very nice). As an aside, PyPy doesn't seem to fare better as far as I can tell (or perhaps it needs several minutes just to warmup its JIT?) :-)

pitrou:
`cython -a` seems to give useful information, though, I'll take another look :-)

mrocklin:
We were eventually able to get something like 1.5x-2x with PyPy after some JIT warming with the old scheduler layout.  I think I got something similar with the newer codebase.

mrocklin:
Benchmark and results with PyPy2:

https://gist.github.com/75f7b4ef2a2c24352d39bc1cf966f04f

mrocklin:
Forgot to include CPython times.  

```
--------------------
Create graph: 0.302 seconds
Execute graph: 3.639 seconds
transitions per second: 9002.4
--------------------
Create graph: 0.299 seconds
Execute graph: 3.910 seconds
transitions per second: 10475.2
--------------------
Create graph: 0.312 seconds
Execute graph: 4.020 seconds
transitions per second: 10187.1
--------------------
Create graph: 0.298 seconds
Execute graph: 4.078 seconds
transitions per second: 10043.3
```

This is with Python 3.6.  PyPy2 gets around 15k transitions per second while CPython gets around 10k.

pitrou:
Update: with the latest Cython changes, I now get a 4% improvement.

pitrou:
Last Cython issue that could help this is https://github.com/cython/cython/issues/2048 . Currently all set iteration uses the generic C APIs, and there's quite a bit of set iteration in `schedulerstate.py`.

pitrou:
I tried Linux `perf` on the benchmark. Learnings:
* 30% of scheduler time is spent on scheduler transitions and their descendants
* 18% of scheduler time is spent inside the garbage collector (note part of this may be during transitions, so this number doesn't add up with the number above)
* 13% of scheduler time is Python interpreter loop overhead (inside `_PyEval_EvalFrameDefault`)

jakirkham:
Have you retested this with Cython 0.28+? Asking as it contains the set iteration optimizations.

mrocklin:

PR #1625, opened 2017-12-12
Add function for process CPU time and thread CPU time

pitrou:
Use thread CPU time in ThrottledGC.


pitrou:
There are probably other places where to switch to one of those functions (notably in `worker.py`).

pitrou:
@mrocklin @ogrisel

ogrisel:
LGTM a posteriori :)

PR #1626, opened 2017-12-12
Incorrect sizeof() result for dicts and memoryviews

pitrou:
Note that unfortunately `sizeof()` is quite slow, and I'm not sure it's a good idea to call it in `distributed.comm`.

### DIFF
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -28,8 +28,13 @@ def sizeof_default(o):
 @sizeof.register(tuple)
 @sizeof.register(set)
 @sizeof.register(frozenset)
-def sizeof_python_collection(seq):
+def sizeof_python_container(seq):
     return getsizeof(seq) + sum(map(sizeof, seq))
+
+
+@sizeof.register(dict)
+def sizeof_python_mapping(dct):
+    return getsizeof(dct) + sum(map(sizeof, dct)) + sum(map(sizeof, dct.values()))
 
 
 @sizeof.register_lazy("numpy")
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -1,6 +1,8 @@
 from __future__ import print_function, division, absolute_import
 
+from functools import reduce
 import logging
+import operator
 import sys
 
 from dask.utils import Dispatch
@@ -35,6 +37,15 @@ def sizeof_python_container(seq):
 @sizeof.register(dict)
 def sizeof_python_mapping(dct):
     return getsizeof(dct) + sum(map(sizeof, dct)) + sum(map(sizeof, dct.values()))
+
+
+@sizeof.register(memoryview)
+def sizeof_memoryview(mem):
+    try:
+        return getsizeof(mem) + mem.nbytes
+    except AttributeError:
+        # Python 2
+        return getsizeof(mem) + reduce(operator.mul, mem.shape, mem.itemsize)
 
 
 @sizeof.register_lazy("numpy")
--- a/distributed/tests/test_sizeof.py
+++ b/distributed/tests/test_sizeof.py
@@ -13,6 +13,9 @@ def test_base():
 
 def test_containers():
     assert sizeof([1, 2, [3]]) > (getsizeof(3) * 3 + getsizeof([]))
+    b = b"x" * 100
+    assert sizeof({b}) >= (getsizeof(b) + getsizeof(set()))
+    assert sizeof({b: b + b'z'}) >= (2 * getsizeof(b) + getsizeof({}))
 
 
 def test_numpy():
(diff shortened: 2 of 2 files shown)


pitrou:
Here is a synthetic example showing how counter-productive it can be to call `sizeof` each time we send a message on the wire:
```
>>> d = {str(i): (i, b"x"*i*10,) for i in range(40)}
>>> %timeit msgpack.packb(d)
15.1 µs ± 61.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>> %timeit distributed.protocol.dumps(d)
28.7 µs ± 121 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
>>> %timeit sizeof(d)
151 µs ± 577 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

@mrocklin, what do you think?

mrocklin:
The alternative here I guess would be to specifically state which messages to offload?

On a different kind of message this has less of an impact.

```python
In [1]: msg = {'status': b'OK', 'x': 123, 'y': 1.23}

In [2]: import msgpack

In [3]: from distributed.sizeof import sizeof

In [4]: %timeit msgpack.packb(msg)
The slowest run took 90.06 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.06 µs per loop

In [5]: %timeit sizeof(msg)
The slowest run took 117.97 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 708 ns per loop
```

I'm curious if this shows up when profiling workloads.  If not then I'm inclined to ignore it if the alternative is to manually annotate which serialization calls to offload.

pitrou:
Right, `sizeof` will be cheap on extremely small messages.

> I'm curious if this shows up when profiling workloads.

I suppose it would depend which workload? It's still not clear to me in which cases we can have a "rich" message where `sizeof` would be expensive (i.e. a large container or nested container), but `update_graph` is probably the likeliest candidate. Now, `update_graph` is serialized on the client, where chewing up CPU cycles in the IO loop is probably less important...

mrocklin:
> I suppose it would depend which workload?

Yeah.  The fact that I haven't seen this before gives me some confidence that these workloads are not particularly common.  Messages tend to either be very short (like the one in my example) or contain large numpy arrays or pandas dataframes (where cost to transfer will overwhelm other costs).  Update_graph is a good counter example.  In short though I'm not really that worried about this.  A fix would be nice, but I doubt that it would significantly impact usability.

pitrou:
Well, this PR degrades performance on `client.ClientSuite.time_trivial_tasks` by 15-20%.

mrocklin:
OK, yeah, that seems relevant

mrocklin:
OK, so some options:

1.  Explicitly mark `write` calls that we might want to offload (probably this is mostly restricted to inter-worker data communication and scattering?)
2.  Always offload, this adds 10s of microseconds to every call, but avoids the occasional 100microsecond burst (my guess is that this isn't worth it, but I'm not sure)
3.  ??

pitrou:
Other thought: until now, since `sizeof` on `dict` was incorrect, we probably didn't offload when serializing (except when a dict *alone* took larger than 10MB, and that would be a very large dict: around 10 million elements on Python 3.6 here), and it didn't seem to hurt very much. So perhaps we should disable offloading on serializing? On deserializing, we measure the size of a couple frames, so speed is not an issue.

pitrou:
(original offloading PR was https://github.com/dask/distributed/pull/1218 and was motivated by a private bug report from a customer -- see link there)

fjetter:
Any progress on this? Is the discussion around the offloading still relevant or did this change in the past 2 years?

While I'm not an expert on this topic, it feels like the question when and how to offload the serialisation is a topic on its own but not relevant to the implementation of dict/memoryview serialisation, is it?

`sizeof` (or rather `nbytes`) is used all over the place in the scheduler and it feels to me that a more accurate calculation of the size of objects would help the scheduler.

mrocklin:
(Antoine, you may want to unsubscribe here if you're no longer interested.  I'll try to avoid using your username in case it re-subscribes you)

The fail case and cost that Antoine points to is relevant.  We can't deal with 100us costs on every message.  However, maybe it makes sense to recurse if the dict is relatively small, perhaps less than 10 elements.  That could cover cases like `{"op": "get-data", "data": large-numpy-array}` or something similar, while avoiding expensive overhead when operating on large dictionaries with many small elements.

mrocklin:
It looks like he's also referring to a benchmark here, which might be good to test against.  

https://github.com/dask/dask-benchmarks/blob/595fd08d1ef8df1229202811179e4915efaeb2a3/distributed/benchmarks/client.py#L27-L39

mrocklin:
This was done in another PR

PR #1630, opened 2017-12-13
Remove recent message log in BatchedSend

pitrou:
This could keep alive many objects, while entirely unused.


mrocklin:
What is the concern about keeping alive objects?  We've ensured that these objects don't take up that much space when serialized.

mrocklin:
It is used when debugging

pitrou:
My concern is to avoid any possible cause of uncontrolled memory usage. It might be mostly theoretical.

> We've ensured that these objects don't take up that much space when serialized.

Well, since we use compression, we may end up with a serialized payload much smaller than the original object:
```
>>> payload = {'data': to_serialize(np.zeros(100*1024*1024, dtype=np.int8))}
>>> frames = dumps(payload)
>>> sizeof(frames)
470843
```

(I won't comment on Blosc's compression efficiency here :-))

mrocklin:
> My concern is to avoid any possible cause of uncontrolled memory usage. It might be mostly theoretical.

This seems like pre-mature optimization to me.  These logs have been useful in the past.  They have not yet been demonstrated to be harmful in practice.  I'm hesitant to remove them.

distributed/batched.py
```diff
@@ -50,7 +49,6 @@ def __init__(self, interval, loop=None):
         self.batch_count = 0
         self.byte_count = 0
         self.next_deadline = None
-        self.recent_message_log = deque(maxlen=100)
```
mrocklin: I am inclined to keep this functionality, but set `maxlen=` to some configuration option that defaults to zero

pitrou: Yes, I suppose it can be made configurable, as long as it's 0 by default ;-)

PR #1631, opened 2017-12-13
Remove recent messages log in worker

pitrou:
This could keep alive many objects, while entirely unused.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1061,7 +1061,6 @@ class Worker(WorkerBase):
         self.long_running = set()
 
         self.batched_stream = None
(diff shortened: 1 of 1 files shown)


mrocklin:
I'm not entirely against removing this, but it is used when debugging.

pitrou:
As in #1630, my concern was to avoid any possible cause of uncontrolled memory usage. Not any practical observation, so this may be entirely pointless.

mrocklin:
I'm inclined to close until this comes up in practice.

pitrou:
No problem from me :-)

Issue #1619, opened 2017-12-08
Adaptive cluster for Kubernetes


mrocklin:
In https://github.com/cedadev/jasmin-dask/blob/master/docker/k8s-adaptive.py there is an adaptive cluster system for Kubernetes, I believe jointly developed by @mkjpryor-stfc ,  @jacobtomlinson, and @niallrobinson .  This was brought up in https://github.com/dask/distributed/pull/1618 .  I'm also interested in this so I thought I'd raise an issue to discuss potential design.

mrocklin:
OK, so looking at the Jasmin implemenation in particular here are some questions/comments:

1.  I've never used the `kubernetes` Python package before, how do you like it?
2.  It seems odd to me to have a Kubernetes spec defined in Python rather than in some yaml file. I'm new to this space though and so don't really know what's common.
3.  How can we generalize something like this to work with other dask-kubernetes deployments?  I would expect different groups to be somewhat opinionated about how they set things up.
4.  In my ideal world when we create a Kubernetes Dask cluster manager we provide it with the name of some group of pods/containers and have it just change the number of active replicas rather than hand-crafting individual pods on its own
5.  However, I really like the ability of the Jasmin cluster manager to destroy particular pods that are identified by the `Scheduler.close_workers` method
6.  What kinds of permissions do you need to give to the scheduler container so that it can perform kubernetes operations like this?

mrocklin:
I'm inclined to make a different adaptive strategy that targets a particular number of workers based on time estimates in `Scheduler.total_occupancy` and on current memory use.  This would avoid a slow ramp-up of workers and would instead jump directly to an appropriate number.

mkjpryor-stfc:
@mrocklin: Here's my take on the points you made:

1. It's OK - better than making the API requests myself. However, the API is generated by Swagger from the API documentation, so it is very verbose and not very "Pythonic". A bespoke Python API might be better but risks falling out of step with what is currently a very rapidly changing API.

2. There are some dynamic aspects to the pod spec, defined using environment variables to the scheduler, which is why it is defined in Python. I suppose you could template it instead - Jinja would probably make most sense from Python - but including and configuring Jinja seemed like a lot of effort for one template.

3. I think it is already quite configurable - you can set namespace, labels, image, image pull policy, and worker resources are all configurable using environment variables to the scheduler. We could add more.

[...]

PR #1632, opened 2017-12-14
Add support for "grouped" adaptive scaling and adaptive behavior overrides.

asford:
_This is a vaguely strawman proposal. I'm *certainly* happy to hear implementation feedback from maintainers, and admit that this conflicts with the open PR #1618, but believe these changes may offer a minimally invasive approach to support custom adaptive logic without further changes to score scheduler class._

- [x] Determine if additional test coverage is required.
- [x] Maybe extended out-of-the-box implementation with "min size" and "max size" logic and hostname-based grouping via command-line flag.
- [ ] Update documentation.

This pull intended to improve support for adaptive scaling on clusters in which there is not a 1-to-1 mapping between workers and underlying cluster resources (a "node"). This may occur in which a resource restriction, such as node startup time, fixed node resource cost or node count limitation, require that multiple workers are hosted on a single node. In these contexts adaptive scaling should logically occur on a per-node basis, rather than a per worker basis, as closing a single worker on a node does not allow removal of the underlying cluster resource. 

This patch attempts to provide the minimal required extension of the core scheduler "workers_to_close" logic required to perform scale-down analysis in a grouped context. Worker activity and memory use are considered on a grouped basis, and groups are selected on an all-or-none basis for retirement. Identification of a worker "group" is deferred to a user-specified grouping function, as the desired behavior is *challenging to anticipate*. A simple example of hostname-based grouping is provided in documentation.

Extends the `Adaptive` base class with an overridable hook for `workers_to_close`, allowing optional specification of a grouping key as well as post-facto analysis of the candidate retirement list. This hook may be used to provide support for minimum cluster size logic, which is demonstrated via a test. This implementation *may* be relevant for #1618.

### DIFF
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -25,6 +25,8 @@ class Adaptive(object):
         Affects quickly to adapt to high tasks per worker loads
     scale_factor : int, default 2
         Factor to scale by when it's determined additional workers are needed
+    **kwargs:
+        Extra parameters to pass to Scheduler.workers_to_close
 
     Examples
     --------
@@ -47,7 +49,7 @@ class Adaptive(object):
     '''
 
     def __init__(self, scheduler, cluster, interval=1000, startup_cost=1,
-                 scale_factor=2):
+                 scale_factor=2, **kwargs):
         self.scheduler = scheduler
         self.cluster = cluster
         self.startup_cost = startup_cost
@@ -55,6 +57,7 @@ class Adaptive(object):
         self._adapt_callback = PeriodicCallback(self._adapt, interval)
         self.scheduler.loop.add_callback(self._adapt_callback.start)
         self._adapting = False
+        self._workers_to_close_kwargs = kwargs
 
     def needs_cpu(self):
         """
@@ -150,7 +153,7 @@ class Adaptive(object):
         """
         return len(self.workers_to_close()) > 0
 
-    def workers_to_close(self):
+    def workers_to_close(self, **kwargs):
         """
         Determine which, if any, workers should potentially be removed from
         the cluster.
@@ -168,7 +171,9 @@ class Adaptive(object):
         --------
         Scheduler.workers_to_close
         """
-        return self.scheduler.workers_to_close()
+        kw = dict(self._workers_to_close_kwargs)
+        kw.update(kwargs)
+        return self.scheduler.workers_to_close(**kw)
 
     @gen.coroutine
     def _retire_workers(self):
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -140,20 +140,41 @@ class Adaptive(object):
 
         Notes
         -----
-        ``Adaptive.should_scale_down`` always returns True, so we will always
-        attempt to remove workers as determined by
-        ``Scheduler.workers_to_close``.
+        ``Adaptive.should_scale_down`` defaults to dispatching to
+        ``Adaptive.workers_to_close``, returning True if any workers to close
+        are specified.
 
         See Also
         --------
         Scheduler.workers_to_close
         """
-        return len(self.scheduler.workers_to_close()) > 0
+        return len(self.workers_to_close()) > 0
+
+    def workers_to_close(self):
+        """
+        Determine which, if any, workers should potentially be removed from
+        the cluster.
+
+        Returns
+        -------
+        workers: [worker_name]
+
+        Notes
+        -----
+        ``Adaptive.workers_to_close`` dispatches to Scheduler.workers_to_close(),
+        but may be overridden in subclasses.
+
+        See Also
+        --------
+        Scheduler.workers_to_close
+        """
+        return self.scheduler.workers_to_close()
 
     @gen.coroutine
     def _retire_workers(self):
         with log_errors():
-            workers = yield self.scheduler.retire_workers(remove=True,
+            workers = yield self.scheduler.retire_workers(workers=self.workers_to_close(),
+                                                          remove=True,
                                                           close_workers=True)
 
             if workers:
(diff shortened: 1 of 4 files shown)


asford:
Given the current updates being made to the scheduler, I will hold off on merging `master` into this pull until I've feedback and +1 from maintainers on the implementation. I will respond to implementation feedback working from the current release `1.20.2`, so that I can continue testing from a stable base in my environment.

mrocklin:
It seems to me that this PR has two separable parts.  

1.  Making the `workers_to_close` call in Adaptive configurable so that others can subclass Adaptive and provide their own logic
2.  Changing `Scheduler.workers_to_close` to support particular grouping logic

I'm entirely in favor of part 1.  In general creating escape hatches so that others can implement custom logic outside of Dask is almost always good.

I'm hesitant about part 2.  At first glance this seems like logic for a specific case entering the mainline codebase.  I would want to see this show up in a few different use cases before it became part of the main code.  That being said, you've done a nice job of making this change minimal, and it doesn't seem to conflict with the existing flow of that operation.  I'm not saying no here, just expressing some concern.

asford:
I completely agree, this is logically separable and needs a little thought before merging.

From my perspective (1) is a must-have, but I can work from there to developing a working implementation for my use case.

On (2), I agree with you that this feels a bit too specialized to live in the scheduler. My options to begin were to (a) modify this function or (b) duplicate the implementation with modifications at a higher level. My gut instinct was that (b) was a ugly solution, but your assessment is spot on. I see the implementation options as:

- (A) **Stay in scheduler.** Non-ideal, as you don't want to add application-specific logic in the class, and there no reason to say that this feature should be privileged over any other.

- (B) **Lift to adaptive.** This would provide the feature "out of the box" with distributed, but would require duplication of the basic `workers_to_close` logic. I think it would be reasonable to extend `adaptive` to provide generic, mutually compatible support for more scaling features and could envision further development occurring there.

- (C) **Extract into an adaptive subclass.** I would prefer that the generic framing of this logic live somewhere within `distributed` so that I can target a reasonable API for my deployment specific plugin and provide testing for the feature within your CI. This might be a "`incubate/contrib`" level feature, implying that you wouldn't guarantee forward compatibility, but I don't know if you've a process in place to support that.

My preference would be to go with (B) and lift this logic into the `Adaptive` class. Would that be an acceptable solution? Would it then make sense to "feature freeze" `Scheduler.workers_to_close` and/or deprecate it?

mrocklin:
OK, short term I'm fine with (A) if you want this to go directly into the Scheduler, perhaps with a note in the docstring that it is experimental.  It seems innocuous enough to add and may help others.  

However, given the flux around this part of the codebase, if you all are building long-term systems I do recommend pulling out some of your logic into custom Adaptive subclasses if you need stability.  

For the API of the grouping function itself we might just expect a new `WorkerState` object.  Currently this doesn't include the `worker_info`, but this might change: https://github.com/dask/distributed/issues/1641

mrocklin:
> My preference would be to go with (B) and lift this logic into the Adaptive class. Would that be an acceptable solution? Would it then make sense to "feature freeze" Scheduler.workers_to_close and/or deprecate it?

My turnaround preference for (a) rather than (b) is to avoid churn short term.  Yes though I agree that we might consider deprecating or factoring workers_to_close out of the Scheduler.   This can happen later though.

mrocklin:
I'm happy to be argued away from this approach though

mrocklin:
I'm moving information into WorkerState objects here: https://github.com/dask/distributed/pull/1661

I suggest that after this is done we change the signature as follows:

```diff
-        group_key: Callable(worker_address, worker_info)
+        group_key: Callable(WorkerState)
```

mrocklin:
I've rebased, updated to use `WorkerState` objects, and force pushed.  @asford I apologize for abusing your fork here.  I hope you don't mind.  If you have a chance to review I would appreciate it.

mrocklin:
I plan to merge this soon if there are no further comments.

asford:
I've had a chance to work with the implementation in a real deployment and now agree with you that this implementation is likely the easiest solution for the time being. I've had to make significant changes to the adaptive layer for my deployment and don't see any reason conflate this feature with those issues.

@mrocklin Don't worry about the force-push; I'll be able to continue working from my original commits for the time being and move to this implementation on the next release. Is there an expected timeline for the next release?

mrocklin:
It sounds like I'm starting down the same path.  My current work is here: https://github.com/pangeo-data/pangeo/pull/56

I'd be curious to hear about any additional issues that you run across.  My apologies for the slow uptake on our end recently. 

> Is there an expected timeline for the next release?

My current guess is in 1-2 weeks.

asford:
No problems on uptake, I've definitely been off-and-on for these pulls as well.

What would be the easiest way for me to document this for you? I can lift the dask-related work out of a private repo within a few days, which may be a good starting point for discussion. I'm still ramping-up on dask and understanding how I can stabily integrate this into our current software stack, but I've a reasonably stable autoscaled deployment running on compute engine.

mrocklin:
> What would be the easiest way for me to document this for you? 

This would be above and beyond, so I'd be happy to accept anything.  My ideal form would be a public-facing blogpost.  Personal e-mail or whatnot would be fine too.  An open-and-shut github issue is probably somewhere in between.  

> but I've a reasonably stable autoscaled deployment running on compute engine.

This seems to be a common theme these days.  This would be, as far as I'm aware, the fourth such dask-on-cloud system I've seen in the last couple months.  Here are some related issues:

- https://github.com/pangeo-data/pangeo/pull/57
- https://github.com/dask/distributed/issues/1619
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/326

This Kubernetes deployment seems to be the furthest along: https://github.com/cedadev/jasmin-dask/blob/master/docker/k8s-adaptive.py cc @mkjpryor-stfc

PR #1633, opened 2017-12-14
Disable OS X build on Travis-CI

pitrou:
It has been broken for quite some time ("language: python" is not supported on OS X builds) and I don't think there's any motivation among us to fix it.


mrocklin:
Fine by me

PR #1635, opened 2017-12-14
Add diagnosis of GC behaviour (Python 3)

pitrou:
This will log a warning when the GC takes too much CPU over several calls, and will log an informational message if the GC releases a lot of memory at once.

### DIFF
--- a/distributed/metrics.py
+++ b/distributed/metrics.py
@@ -10,6 +10,11 @@ _empty_namedtuple = collections.namedtuple("_empty_namedtuple", ())
 
 
 def _psutil_caller(method_name, default=_empty_namedtuple):
+    """
+    Return a function calling the given psutil *method_name*,
+    or returning *default* if psutil is not present.
+    """
+    # Import only once to avoid the cost of a failing import at each wrapper() call
     try:
         import psutil
     except ImportError:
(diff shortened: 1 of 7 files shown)


pitrou:
@mrocklin any thoughts on this?

mrocklin:
I haven't taken a look at it yet.  Given the amount of work here I doubt that I will get to this any time in the near future.  If you are only mildy confident that this is a good addition then I recommend merging.  If not then maybe we can convince @bluenote10 or @ogrisel to take a look at this.

pitrou:
I think the only possible concern here is the potential performance impact of having a callback invoked at each GC attempt. But there doesn't seem to be any significant slowdown in the scheduler benchmark.

bnaul:
@pitrou @mrocklin Any thoughts on parametrizing these warnings via the config? My worker logs are just thousands of repetitions of:
```
distributed.utils_perf - INFO - full garbage collection released 39.18 MB from 121182 reference cycles (threshold: 10.00 MB)
```
I could filter them manually but it seems like there are other similar config options already so it might make sense there.

mrocklin:
Parametrizing warnings in the config sounds good to me.  

I can also imagine having some sort of exponential backoff for the warning that would increase the time allowed until the next warning if it kept happening.

PR #1636, opened 2017-12-15
Warn on serializing large data in Client

mrocklin:
This raises a warning when serializing large amounts of data in the client.
It prints the object and the size in bytes.  It points users towards
Client.scatter on large objects.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -10,6 +10,7 @@ from pickle import PicklingError
 import random
 import threading
 import sys
+import warnings
 import weakref
 
 from dask.core import istask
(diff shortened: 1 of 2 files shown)


martindurant:
If the object is large, is there not at least some chance that its str is large too, and should not be dumped to the console?

mrocklin:
Indeed, hence these lines: 

```python
+        s = str(obj)
 +        if len(s) > 70:
 +            s = s[:50] + ' ... ' + s[-15:]
```

I should add this explicitly in the tests though.  Thanks!

distributed/worker.py
```diff
@@ -695,16 +696,39 @@ def dumps_task(task):
...
+
+def warn_dumps(obj, dumps=pickle.dumps, limit=1e6):
+    """ Dump an object to bytes, warn if those bytes are large """
+    b = dumps(obj)
+    if not _warn_dumps_warned[0] and len(b) > limit:
+        _warn_dumps_warned[0] = True
+        s = str(obj)
+        if len(s) > 70:
+            s = s[:50] + ' ... ' + s[-15:]
+        warnings.warn("Large object of size %s detected in task graph: \n"
```
jcrist: Perhaps add the type in as well if you're slicing up the `str` repr of the object?

mrocklin: I did this at first, but in practice the type is just a tuple of args.

PR #1638, opened 2017-12-15
DOC: get_client and dask.compute example

stsievert:
The page on launching tasks from workers was stale and only mentioned `worker_client`. This refreshes that and mentions `dask.compute` and `get_client`. It also tries to mention why one might choose a particular method.

I am unsure if I am missing any important points, and I feel the descriptions of `dask.compute` and `get_client` are too terse. Could someone look it over and suggest changes?

### DIFF
--- a/docs/source/task-launch.rst
+++ b/docs/source/task-launch.rst
@@ -116,9 +116,9 @@ object that connects to the scheduler. There are three options for this:
 dask.delayed
 ~~~~~~~~~~~~
 
-`dask.delayed`_ behaves as normal: it submits the functions to the graph,
-optimizes for less bandwidth/computation and gathers the results.
-For more detail, see `dask.delayed`_.
+The Dask delayed behaves as normal: it submits the functions to the graph,
+optimizes for less bandwidth/computation and gathers the results.  For more
+detail, see `dask.delayed`_.
 
 .. code-block:: python
 
@@ -149,9 +149,9 @@ For more detail, see `dask.delayed`_.
 Getting the client on a worker
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The :py:func:`distributed.get_client` function provides a normal Client object that gives full
-access to the dask cluster, including the ability to submit, scatter, and
-gather results.
+The :py:func:`get_client <distributed.get_client>` function provides a normal
+Client object that gives full access to the dask cluster, including the ability
+to submit, scatter, and gather results.
 
 .. code-block:: python
 
@@ -177,9 +177,10 @@ once. Each task does not communicate to the scheduler that they are waiting on
 results and are free to compute other tasks. This can deadlock the cluster if
 every scheduling slot is running a task and they all request more tasks.
 
-To avoid this deadlocking issue we can use :py:func:`distributed.secede` and
-:py:func:`distributed.rejoin`. These functions will remove and rejoin the
-current task from the cluster respectively.
+To avoid this deadlocking issue we can use :py:func:`secede
+<distributed.secede>` and :py:func:`rejoin <distributed.rejoin>`. These
+functions will remove and rejoin the current task from the cluster
+respectively.
 
 .. code-block:: python
 
@@ -197,9 +198,10 @@ current task from the cluster respectively.
 Connection with context manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The :py:func:`distributed.worker_client` function performs the same task as
-:py:func:`distributed.get_client`, but is implemented as a context manager.
-Using ``worker_client`` as a context manager ensures proper cleanup on the
+The :py:func:`worker_client <distributed.worker_client>` function performs the
+same task as :py:func:`get_client <distributed.get_client>`, but is implemented
+as a context manager.  Using :py:func:`worker_client
+<distributed.worker_client>` as a context manager ensures proper cleanup on the
 worker.
 
 .. code-block:: python
@@ -222,11 +224,11 @@ worker.
         result = future.result()
         print(result)  # prints "55"
 
-Tasks that invoke ``worker_client`` are conservatively assumed to be *long
+Tasks that invoke :py:func:`worker_client <distributed.worker_client>` are conservatively assumed to be *long
 running*.  They can take a long time blocking, waiting for other tasks to
 finish, gathering results, etc. In order to avoid having them take up
 processing slots the following actions occur whenever a task invokes
-``worker_client``.
+:py:func:`worker_client <distributed.worker_client>`.
 
 1.  The thread on the worker running this function *secedes* from the thread
     pool and goes off on its own.  This allows the thread pool to populate that
--- a/docs/source/task-launch.rst
+++ b/docs/source/task-launch.rst
@@ -76,36 +76,24 @@ and so this only added a few milliseconds to our total processing time.
 Extended Example
 ~~~~~~~~~~~~~~~~
 
-This example computing the Fibonacci numbers creates tasks that submit tasks
-that submit tasks that submit other tasks, etc. This example is a good example
-of when the APIs below should be used.
+Computing the Fibonacci numbers creates involves a recursive function. When the
+function is run, it calls itself using values it computed. We will use this as
+an example throughout this documentation to illustrate different techniques of
+submitting tasks from tasks.
 
 .. code-block:: python
 
-   In [1]: from distributed import Client
-
-   In [2]: client = Client()
-
-   In [3]: def fib(n):
+   In [1]: def fib(n):
       ...:     if n < 2:
       ...:         return n
-      ...:     a = fib(n - 1)  # we want to submit this function to the client
-      ...:     b = fib(n - 2)  # we want to submit this function to the client
+      ...:     a = fib(n - 1)
+      ...:     b = fib(n - 2)
       ...:     return a + b
 
-   In [4]: future = client.submit(fib, 100)
+   In [2]: fib(10)
+   Out[2]: 55
 
-   In [5]: future
-   Out[5]: <Future: status: finished, type: int, key: fib-7890e9f06d5f4e0a8fc7ec5c77590ace>
-
-   In [6]: future.result()
-   Out[6]: 354224848179261915075
-
-This example is a bit extreme and spends most of its time establishing client
-connections from the worker rather than doing actual work, but does demonstrate
-that even pathological cases function robustly. We will use this example later
-on for the various interfaces.
-
+We will use this example to show the different interfaces.
 
 Submit tasks from worker
 ------------------------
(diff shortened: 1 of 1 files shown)


mrocklin:
In general I think that this is a great improvement, and I think that the use of the Fibonnaci function is a good example to use throughout all of these operations.  

I've been trying to consolidate usage information over at the dask/dask documentation, and in particular there is a bit over at the [Futures section](http://dask.pydata.org/en/latest/futures.html#submit-tasks-from-tasks) there.  Some of the content in this PR might also be relevant there.  At some point in the moderate future I might want to deprecate usage docs here.

stsievert:
Those docs provide a good introduction. Do we want to provide an overview of when to choose between `dask.compute`, `get_client` and `worker_client`?

> deprecate usage docs here.

Can you expand on that?

mrocklin:
> Can you expand on that?

We'll probably move this documentation from distributed.readthedocs.io to dask.pydata.org

docs/source/task-launch.rst
```diff
@@ -58,21 +58,54 @@ on ``data`` remotely and then break up data into its various elements.
...
+   In [2]: client = Client()
+
+   In [3]: def fib(n):
+      ...:     if n < 2:
+      ...:         return n
+      ...:     a = fib(n - 1)  # we want to submit this function to the client
+      ...:     b = fib(n - 2)  # we want to submit this function to the client
+      ...:     return a + b
+
+   In [4]: future = client.submit(fib, 100)
```
mrocklin: Does this actually finish?  `fib(100)` is pretty crazy

stsievert: Looks like it got left in there by mistake. The old docs used `worker_client` in this function.

mrocklin: Right, the text just below this is also a bit odd.  It makes it sound like we're using Dask here but we aren't.  We might avoid all use of Dask in this first example of "the thing we're trying to parallelize"

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+      scheduling slot (the worker that submitted the jobs is labeled as "maybe
+      long running").
+* ``worker_client``
+    * Intended to submit long running jobs (the worker is labeled as "long
+      running")
+    * Implemented as a context manager, which provides some ease of use.
+    * Will ``secede`` from the current worker, and likely transition to another
+      worker.
+    * Establishing a connection to the scheduler takes roughly 10–20ms. It's
+      wise for the submitted jobs to be several times longer than this.
```
mrocklin: Many of the details here assume that the reader already knows a lot about how these things work.  I might leave most of these notes for later.  Instead I would start with "here are three options and how to use them" then maybe "here is how you might choose between those options" finally finishing with "by the way, here are some technical details that you should beware of if you've made it this far in the document.

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
-           elements = e.scatter(data)
-           futures = e.map(process_element, elements)
-           analysis = e.submit(aggregate, futures)
-           result = analysis.result()
-       return result
+    @delayed
+    def fib(n):
+        if n < 2:
+            return n
+        a, b = compute(fib(n-1), fib(n-2))
```
mrocklin: I would split this out so that we can comment that the results of calling `fib` are delayed

```python
a = fib(n - 1)  # these calls are delayed 
b = fib(n - 2)
a, b = compute(a, b)  # execute both in parallel
```

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+        return a + b
 
-    analysis = e.submit(process_all, data)  # spawns many tasks
+    client = Client()  # to change default dask scheduler
```
mrocklin: I recommend starting the Client at the top of this script and changing the comment to "these features require the newer dask.distributed scheduler"

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+    assert fib(4).compute() == 3
 
-This approach is somewhat complex but very powerful.  It allows you to spawn
-tasks that themselves act as potentially long-running clients, managing their
-own independent workloads.
+Look at the documentation of ``dask.compute`` for more detail.
```
mrocklin: The compute documentation doesn't include much information on this kind of workflow.  Instead I might add a link to the [dask.delayed](http://dask.pydata.org/en/latest/delayed.html) documentation at the top, perhaps inline in some comment like "we can use [dask.delayed](http://dask.pydata.org/en/latest/delayed.html) and dask.compute to launch computations from within tasks."

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
 
-Extended Example
-~~~~~~~~~~~~~~~~
+``get_client``
+~~~~~~~~~~~~~~
 
-This example computing the Fibonacci numbers creates tasks that submit tasks
-that submit tasks that submit other tasks, etc..
+``get_client`` is a lower-leve interface than ``worker_client``. It exposes
```
mrocklin: typo `leve` -> `level`.  

We haven't yet gotten to `worker_client` and so shouldn't mention it.  We also haven't gotten to `secede` and `rejoin`.  I recommend doing a simple example with `get_client` alone, saying that it gives us the ability to do all of the client operations like submit, scatter, gather, etc.. from within a task.

Then I would provide a cautionary note about flooding all of the available scheduling slots, and then expand the example with secede and rejoin around the gather to show how they can resolve this problem.

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+    from distributed import Client, get_client, secede, rejoin
 
-   In [2]: client = Client()
+    def fib(n):
+        if n < 2:
+            return n
+        client = get_client()
+        jobs = client.map(fib, [n-1, n-2])
+        secede()  # so we don't take up a scheduling slot
+        return sum(client.gather(jobs))
```
mrocklin: I would resist the temptation to combine the gather and return steps.  I might even through in a `rejoin` operation just to be explicit

```python
a_future = client.submit(fib, n - 1)
b_future = client.submit(fib, n - 2)
secede()
a, b = client.gather([a_future, b_future])
rejoin()
return a + b
```

Note that I also switched to using `a, b` and `client.submit` to maintain parity with the previous example above.

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+cluster and no computation will be performed.
 
-   In [5]: future
-   Out[5]: <Future: status: finished, type: int, key: fib-7890e9f06d5f4e0a8fc7ec5c77590ace>
+``rejoin`` is also available, and may be more advantageous if more work is done
+(and not just a ``return`` statement).
 
-   In [6]: future.result()
-   Out[6]: 354224848179261915075
+``dask.worker_client``
```
mrocklin: I would use `worker_client` rather than `dask.worker_client`.  The latter implies that worker_client is available in the dask namespace.

docs/source/task-launch.rst
```diff
@@ -84,67 +117,122 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
 
+.. code-block:: python
 
-Technical details
-~~~~~~~~~~~~~~~~~
+   from distributed import worker_client
+   import time
+
+   def fib(n):
+       time.sleep(1)  # tasks are assumed to be long running
```
mrocklin: This statement may confuse new readers.  Looking at this with my naive-user persona I think "Oh, worker_client assumes that my tasks need to sleep for a while."

stsievert: What about the below?

``` python
def fib(n):
    time.sleep(1)  # to simulate a long running function.
    # ...
```

stsievert:
Thanks for the review. I'll revise it soon, hopefully later tonight.

> We'll probably move this documentation from distributed.readthedocs.io to dask.pydata.org

I'd like to see something on distributed.readthedocs.io, at least a link to the relevant section of dask.pydata. My use has only involved dask.distributed, not dask.

mrocklin:
> My use has only involved dask.distributed, not dask

I would like to remove this distinction.  The fact that there are two different repositories is just for developmental ease.  I would like to see people perceive this as one project.

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+       ...:     a_future = client.submit(fib, n - 1)
+       ...:     b_future = client.submit(fib, n - 2)
+       ...:     secede()
+       ...:     a, b = client.gather([a_future, b_future])
+       ...:     rejoin()
+       ...:     return a + b
+
+Using ``secede`` can possibly have the worker move to a different node if it's
+long running. Transferring state between nodes can be costly time-wise if
+there's a lot of internal state (e.g., with a large matrix).
```
stsievert: Is this accurate? Are there other things to watch out?

stsievert:
I've reorganized – thanks for your comments @mrocklin.

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
-ensure proper cleanup on the worker.
+object that connects to the scheduler. There are three options for this:
+
+1. ``dask.delayed`` and ``dask.compute``
+2. ``get_client`` with ``secede`` and ``rejoin``
+3. ``worker_client``
+
+.. _Scheduling Policies: https://distributed.readthedocs.io/en/latest/scheduling-policies.html
+.. _Scheduling State: https://distributed.readthedocs.io/en/latest/scheduling-state.html
+.. _The compute function: https://dask.pydata.org/en/latest/scheduler-overview.html#the-compute-function
```
mrocklin: It's not clear that these do anything here.  You may want to build locally 

```
cd doc
make html
```

Also, when referring to internal documentation you should refer to it relatively.  Search the codebase for `:doc:`

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+
+``dask.delayed``
+~~~~~~~~~~~~~~~~
```
mrocklin: I recommend that we avoid code syntax in titles.

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
 
-    analysis = e.submit(process_all, data)  # spawns many tasks
+    In [4]: @delayed
+       ...: def fib(n):
+       ...:     if n < 2:
+       ...:         return n
+       ...:     # We can use dask.delayed and dask.compute to launch
+       ...:     # computation from within tasks
+       ...:     a = fib(n-1)  # these calls are delayed
+       ...:     b = fib(n-2)
```
mrocklin: pep8, add spaces around binary operators

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
-~~~~~~~~~~~~~~~~
+.. _dask.delayed: https://dask.pydata.org/en/latest/delayed.html
 
-This example computing the Fibonacci numbers creates tasks that submit tasks
-that submit tasks that submit other tasks, etc..
+``get_client``
+~~~~~~~~~~~~~~
+
+``distributed.get_client`` allows individual workers to access the ``client``
+that the scheduler uses:
```
mrocklin: This is worded oddly.  The scheduler doesn't use clients.  How about

"The get_client function provides a normal Client object that gives full access to the dask cluster, including the ability to submit, scatter, and gather results"

Ideally we would also link to `get_client` in the internal API (there is some sphinx magic to do this) and drop the `distributed` prefix (I'd like to avoid people referring to this as a separate library).

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
-      ...:             return a + b
-      ...:
+    In [3]: client = Client()
+
+    In [4]: future = client.submit(fib, 10)
+
+    In [5]: future.result()
+    Out[5]: 55
+
+This can deadlock the scheduler if too many nodes request jobs at once. To
```
mrocklin: I would add in a `"However,"` to start here.  We might also explain a bit more about the mechanism that causes deadlocking.  It's not just requesting jobs that's the problem, it's that they tasks doing the requesting, like our call to `fib(55)` are taking up a slot in a worker and if enough tasks like that occur then all tasks will just be waiting for slots that will never appear.

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+correct that, we can use ``secede`` and ``rejoin``. These functions will remove
+and rejoin the current node from the cluster respectively.
```
mrocklin: node -> task

docs/source/task-launch.rst
mrocklin: No, the task will remain on the same worker.

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+       ...:     return a + b
+
+Using ``secede`` can possibly have the worker move to a different node if it's
+long running. Transferring state between nodes can be costly time-wise if
+there's a lot of internal state (e.g., with a large matrix).
+
+``worker_client``
+~~~~~~~~~~~~~~~~~~~~~~
+
+``worker_client`` is a convenience function to do this for you so that you
```
mrocklin: I generally try to avoid starting sentences with code highlighting.  Perhaps "The `worker_client` function ..."

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+don't have to pass around connection information.  However you must use this
+function ``worker_client`` as a context manager to ensure proper cleanup on the
+worker.
```
mrocklin: My naive reader persona is now confused about what you mean by connection information.  Wait, is there something I'm supposed to pass around?  Why didn't I have to do that before?

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+
+    In [1]: from distributed import worker_client
 
-   In [4]: future = e.submit(fib, 100)
+    In [2]: import time
 
-   In [5]: future
-   Out[5]: <Future: status: finished, type: int, key: fib-7890e9f06d5f4e0a8fc7ec5c77590ace>
+    In [3]: def fib(n):
+       ...:     time.sleep(1)  # to simulate a long running function
```
mrocklin: I'm still confused why this time.sleep is here and not elsewhere.  I'm generally in favor of it not being around, but it should at least be consistent.

Also, it might be nice to have these code samples not in IPython syntax so that they can be easily copy-pasted.  I'm not sure what standard conventions are for this though.

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
+       ...:     if n < 2:
+       ...:         return n
+       ...:      with worker_client() as client:
+       ...:          a_future = client.compute(fib, n - 1)
+       ...:          b_future = client.compute(fib, n - 2)
```
mrocklin: Why do we now use client.compute rather than client.submit?

docs/source/task-launch.rst
```diff
@@ -84,71 +105,131 @@ We can submit tasks from other tasks.  This allows us to make decisions while
...
-that even pathological cases function robustly.
+    In [5]: future = client.submit(fib, 10)
 
+    In [6]: future.result()
+    Out[6]: 55
 
-Technical details
-~~~~~~~~~~~~~~~~~
+This allows you to spawn tasks that themselves act as potentially long-running
+clients, managing their own independent workloads.
```
mrocklin: This is mostly the same in the cases above as well.  Worker_client doesn't do anything other than call get_client, secede, and rejoin. This is also exactly what dask.compute does.  You're right that it does send the "long-running" signal up, but this rarely affects real workloads.

docs/source/task-launch.rst
```diff
@@ -163,6 +244,6 @@ processing slots the following actions occur whenever a task invokes
 Because of this behavior you can happily launch long running control tasks that
 manage worker-side clients happily, without fear of deadlocking the cluster.
```
mrocklin: But this is also true of the secede/rejoin and dask.compute approaches above, no?

docs/source/task-launch.rst
stsievert: That's reassuring. I now realize the `transition_*` functions are to transition state, not machines.

docs/source/task-launch.rst
stsievert: I found a way to link to the `get_client` internal API and remove the `distributed` prefix.

docs/source/task-launch.rst
stsievert: I was following the style in this document – I think there was another function with some IPython syntax. I've moved the `fib` example to Python script syntax.

stsievert:
I've refreshed the doc page, and have tried to strip out the old docs in the `worker_client` section. I've built the HTML docs and made sure the links worked/looked as expected.

stsievert:
What else need to be done on this?

mrocklin:
Thanks for your patience here @stsievert .  This looks great to me.  Merging.

PR #1640, opened 2017-12-15
Make Future._repr_html_ work with tuple keys

### LINKED ISSUE #1637: Future repr issue in Jupyter Notebook
When trying to inspect a `Future` in the Jupyter Notebook, an issue cropped up trying to represent it. Looks like it assumes a `key` is a `str` when it is in fact a `tuple`. Running `distributed` version 1.20.1.

```python
AttributeError                            Traceback (most recent call last)
/opt/conda3/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

/opt/conda3/lib/python3.6/site-packages/distributed/client.py in _repr_html_(self)
    334                 typ = str(self.type)
    335             text += '<font color="gray">type: </font>%s, ' % typ
--> 336         text += '<font color="gray">key: </font>%s' % html_escape(self.key)
    337         return text
    338 

/opt/conda3/lib/python3.6/html/__init__.py in escape(s, quote)
     17     translated.
     18     """
---> 19     s = s.replace("&", "&amp;") # Must be done first!
     20     s = s.replace("<", "&lt;")
     21     s = s.replace(">", "&gt;")

AttributeError: 'tuple' object has no attribute 'replace'

<Future: status: finished, type: Array, key: ('store-rechunk-merge-9d56828f48fdb9af2a81e6b10a09d976', 3, 3, 0)>
```

mrocklin:
Fixes #1637

cc @jakirkham

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -334,7 +334,7 @@ class Future(WrappedKey):
             except AttributeError:
                 typ = str(self.type)
             text += '<font color="gray">type: </font>%s, ' % typ
-        text += '<font color="gray">key: </font>%s' % html_escape(self.key)
+        text += '<font color="gray">key: </font>%s' % html_escape(str(self.key))
         return text
 
     def __await__(self):
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -271,11 +271,21 @@ def test_persist_retries(c, s, a, b):
 
 @gen_cluster(client=True)
 def test_future_repr(c, s, a, b):
+    x = c.submit(inc, 10)
     for func in [repr, lambda x: x._repr_html_()]:
-        x = c.submit(inc, 10)
         assert str(x.key) in func(x)
         assert str(x.status) in func(x)
         assert str(x.status) in repr(c.futures[x.key])
+
+
+@gen_cluster(client=True)
+def test_future_tuple_repr(c, s, a, b):
+    da = pytest.importorskip('dask.array')
+    y = da.arange(10, chunks=(5,)).persist()
+    f = futures_of(y)[0]
+    for func in [repr, lambda x: x._repr_html_()]:
+        for k in f.key:
+            assert str(k) in func(f)
 
 
 @gen_cluster(client=True)


distributed/client.py
```diff
@@ -334,7 +334,7 @@ def _repr_html_(self):
             except AttributeError:
                 typ = str(self.type)
             text += '<font color="gray">type: </font>%s, ' % typ
-        text += '<font color="gray">key: </font>%s' % html_escape(self.key)
+        text += '<font color="gray">key: </font>%s' % html_escape(str(self.key))
```
jakirkham: Not to be pedantic, but do you think this should be `str` or `repr`?

jakirkham:
Thanks. LGTM. Asked one question above.

distributed/client.py
mrocklin: Pedantry is welcome.  When would this matter?

In practice keys are only of type str, bytes, or tuple, but this isn't enforced so a sane default here would be good.

jakirkham: Not in any current cases that I can think of.

There is a little bit of subtlety between `repr` and `str`. The former typically returns something that can be evaluated to get something equivalent. Also `repr` and `str` behave slightly differently with `bytes` vs. `str` on Python 2 or Python 3. 

<details>
<summary>Python 2</summary>

```python
In [1]: repr(b"a")
Out[1]: "'a'"

In [2]: repr(u"a")
Out[2]: "u'a'"

In [3]: str(b"a")
Out[3]: 'a'

In [4]: str(u"a")
Out[4]: 'a'
```

</details>

<details>
<summary>Python 3</summary>

```python
In [1]: repr(b"a")
Out[1]: "b'a'"

In [2]: repr(u"a")
Out[2]: "'a'"

In [3]: str(b"a")
Out[3]: "b'a'"

In [4]: str(u"a")
Out[4]: 'a'
```

</details>

mrocklin: `str` it is

Issue #1620, opened 2017-12-08
RuntimeError: Event loop is closed after running tests on Debian.


detrout:
Hi,

For some reason when the tests finish running with Debian autopkgtests there's an exception that looks like its happening during cleanup. (Since the report of what tests passed happened).

distributed: 1.20.2
dask: 0.16.0
python: 3.6.4rc1

root@amarana:/tmp/autopkgtest.8SWqzJ/autopkgtest_tmp
```
============================== 4 tests deselected ==============================
 987 passed, 82 skipped, 4 deselected, 8 xfailed, 2 xpassed, 1 warnings in 417.17 seconds 
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/distributed/deploy/local.py", line 237, in close
    self.loop.add_callback(self._stop_worker, w)
  File "/usr/lib/python3/dist-packages/tornado/platform/asyncio.py", line 149, in add_callback
    functools.partial(stack_context.wrap(callback), *args, **kwargs))
  File "/usr/lib/python3.6/asyncio/base_events.py", line 620, in call_soon_threadsafe
    self._check_closed()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 357, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

autopkgtests is testing the installed package, and not running in the source tree.

I copied conftest.py from the source tree into a temporary directory and I'm rusing the following py.test command

```
LC_ALL=C.UTF-8 LANG=C.UTF-8 http_proxy= \
  $py $(which py.test-3)  -m "not avoid_travis" . /usr/lib/python3/dist-packages/distributed \
  --verbose -r s --timeout-method=thread --timeout=300 --durations=20
```

(The environment variables are for oddnesses of the test chroot and not likely to be relevant).

adding "." to before the path to the distributed install directory is so py.test picks up the conftest.py module.

My most successful workaround was to add this:

```
--- ../tree/distributed/deploy/local.py	2017-12-07 21:48:26.000000000 +0000
+++ /usr/lib/python3/dist-packages/distributed/deploy/local.py	2017-12-08 18:38:06.757564588 +0000
@@ -242,8 +242,10 @@
                     sleep(0.01)
             del self.workers[:]
             self._loop_runner.run_sync(self._close, callback_timeout=timeout)
             self._loop_runner.stop()
+        except RuntimeError as e:
+            logger.error("%s", e)
[...]

PR #1645, opened 2017-12-19
Catch the RuntimeError caused if the _loop_runner is already shutdown

detrout:
This is a fix for https://github.com/dask/distributed/issues/1620


mrocklin:
Closing in favor of #1648

PR #1657, opened 2017-12-22
Add client heartbeats

### LINKED ISSUE #1655: Dask distributed client doesn't keep the connection alive
We connect client machine to a distributed dask scheduler, submit a long running task graph and eventually call progress() to wait for completion.
In this case dask client doesn't seem to keep the connection alive.
If the connection goes through ELB or any other element that has an idle timeout, the connection is disconnected and the job is dropped (in case of ELB max timeout is 4000s).

mrocklin:
Fixes #1655

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -456,6 +456,8 @@ class Client(Node):
     name: string (optional)
         Gives the client a name that will be included in logs generated on
         the scheduler for matters relating to this client
+    heartbeat_interval: int
+        Time in milliseconds between heartbeats to scheduler
 
     Examples
     --------
@@ -484,7 +486,7 @@ class Client(Node):
     def __init__(self, address=None, loop=None, timeout=5,
                  set_as_default=True, scheduler_file=None,
                  security=None, asynchronous=False,
-                 name=None, **kwargs):
+                 name=None, heartbeat_interval=None, **kwargs):
 
         self.futures = dict()
         self.refcount = defaultdict(lambda: 0)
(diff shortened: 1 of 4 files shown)


mrocklin:
cc @pitrou for review

distributed/config.yaml
```diff
@@ -58,3 +58,5 @@ worker-memory-terminate: 0.95  # fraction at which we terminate the worker
 
 log-length: 10000  # default length of logs to keep in memory
 log-format: '%(name)s - %(levelname)s - %(message)s'
+
+client-heartbeat-interval: 5000  # milliseconds between client heartbeats
```
pitrou: I'd rather have seconds everywhere :-)

distributed/client.py
```diff
@@ -456,6 +456,8 @@ class resembles executors in ``concurrent.futures`` but also allows
     name: string (optional)
         Gives the client a name that will be included in logs generated on
         the scheduler for matters relating to this client
+    heartbeat_interval: int
+        Time in milliseconds between heartbeats to scheduler
```
pitrou: Ditto here.

distributed/config.yaml
mrocklin: Yeah, I was thinking about this as I was writing this.  Historically we have preferred milliseconds when we're just passing through to Tornado or Bokeh and they expect seconds.  However we also use seconds when we do our own thing.  This has plenty of opportunity for confusion.

As a solution I suggest that we make a `parse_time` function and use strings like `"5000 ms"` in configuration files in the future.  

Smoothly transitioning will be tricky no matter what solution we choose.

mrocklin: > Tornado or Bokeh and they expect seconds

Should be 

Tornado or Bokeh and they expect milliseconds

pitrou: I suspect Bokeh copied Tornado's initial API mistake. But, well, no need to accumulate our own debt here :-)

mrocklin: Yeah, I'm in favor of changing this.  This is a larger task though and will need to be done consistently across the codebase.

mrocklin: @pitrou any objection to holding off on this change until we change it globally?  For the moment this choice matches current conventions in the codebase.

I'm happy to commit to changing to seconds or strings everywhere prior to the next 0.X.0 release.

pitrou:
Ok from me.

mrocklin:
Merging.  Thanks for the review @pitrou 

Follow up issue on time units in https://github.com/dask/distributed/issues/1660

PR #1661, opened 2017-12-24
Centralize more state into WorkerState

mrocklin:
This moves information like `Scheduler.worker_info` into `WorkerState.info` and other attributes on WorkerState.  

We also rename `worker_key` to `address`.


pitrou:
This sounds fine to me.

PR #1666, opened 2017-12-28
Skip unhashable functions test

mrocklin:
This fails with the new cloudpickle release on Python 2.7

I have no desire to identify and fix the problem on Python 2.7.  This is non-critical functionality.  Skipping the test.


mrocklin:
Reported upstream at https://github.com/cloudpipe/cloudpickle/issues/144

PR #1670, opened 2017-12-28
Remove threaded map(), scatter() and gather()

pitrou:
Those are difficult to implement without creating a runaway thread, and the functionality is little used nowadays, so remove it.


mrocklin:
We should also remove http://distributed.readthedocs.io/en/latest/queues.html

mrocklin:
Or perhaps remove it from the index and add text to that page saying that the feature has been removed.

mrocklin:
In various cases we might instead point people to the [streamz](http://streamz.readthedocs.io/en/latest/) project, which is probably the successor here.

PR #1671, opened 2017-12-28
Avoid infinite reevalute_occupancy() coroutine

pitrou:
The infinite coroutine seems to be the cause of many leaked Scheduler objects.
This change saves 20% memory on the test suite here (from 500MB down to 400 MB).

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3990,8 +3990,7 @@ class Scheduler(ServerNode):
     # Cleanup #
     ###########
 
-    @gen.coroutine
-    def reevaluate_occupancy(self):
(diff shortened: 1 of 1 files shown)


pitrou:
A possible explanation for the leak is https://github.com/tornadoweb/tornado/issues/2229

mrocklin:
If this is the main source of the leak then it might make sense to just
merge this new piece and leave the removal of threaded map/scatter/gather
for later (if ever?)

pitrou:
Those are two different leaks. The fact that threaded map/scatter/gather can leave some threads around deserves fixing in itself, IMO.

pitrou:
That said, I've rebased on master.

PR #1672, opened 2017-12-29
Allow to use `retries=` argument in ClientExecutor

rqx:
Now `retries=` argument will be passed automatically to `submit()` or `map()`

### DIFF
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -287,7 +287,7 @@ def run_for(duration, timer=time):
 
 
 # This dict grows at every varying() invocation
-_varying_dict = collections.Counter()
+_varying_dict = collections.defaultdict(int)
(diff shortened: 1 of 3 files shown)


pitrou:
Thank you! This looks good on the principle.

distributed/utils_test.py
```diff
@@ -287,9 +287,10 @@ def run_for(duration, timer=time):
 
 
 # This dict grows at every varying() invocation
-_varying_dict = {}
+_varying_dict = collections.Counter()
```
pitrou: `collections.defaultdict(int)` looks more idiomatic here.

mrocklin:
I've changed Counter to defaultdict on @pitrou 's suggestion and am now merging.  

Thanks @rqx !

PR #1679, opened 2018-01-05
Add base_url parameter to Bokeh handling

mrocklin:
This adds a `base_url` to all links within the Bokeh dashboard.  So that instead of serving at `/foo` we now serve at `/some_base_url/foo` .  Most links within our application seem to propagate fine, however we seem to have lost the actual Bokeh plots themselves.  The template renders but the plots don't.  

cc @bryevdv if you have thoughts on this I would welcome help here.

### DIFF
--- a/distributed/bokeh/scheduler.py
+++ b/distributed/bokeh/scheduler.py
@@ -19,7 +19,7 @@ from bokeh.models.widgets import DataTable, TableColumn
 from bokeh.plotting import figure
 from bokeh.palettes import Viridis11
 from bokeh.io import curdoc
-from toolz import pipe
+from toolz import pipe, merge
 try:
     import numpy as np
 except ImportError:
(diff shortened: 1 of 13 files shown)


mrocklin:
Ah, this seems to work well if I also pass along the `prefix=` keyword to Bokeh.  I guess I'm confused about the relationship between these parameters.

mrocklin:
Perhaps I am mistaken and the prefix and base_url arguments should be merged.  I wonder what Jupyter does here.

bryevdv:
`--base-url` is not part of Bokeh. It seems to be duplicated of `--prefix` which does this:
```
The Bokeh server can also add an optional prefix to all URL paths.
This can often be useful in conjunction with "reverse proxy" setups.

.. code-block:: sh

    bokeh serve app_script.py --prefix foobar

Then the application will be served under the following URL:

.. code-block:: none

    http://localhost:{DEFAULT_PORT}/foobar/app_script
```

mrocklin:
OK, thanks for the comment @bryevdv .  Replaced with `prefix` everywhere.

PR #1705, opened 2018-01-23
Handle nested parallelism in distributed.joblib

TomAugspurger:
This implements the new API introduced in
https://github.com/joblib/joblib/pull/538 for to ensure nested parallelism works
correctly.

This breaks our API as previously `DaskDistributedBackend` had a `client` attribute
with the `Client` that was used. I believe (though I may be wrong) that the backend object gets serialized in nested calls, and clients are not serializable.

Nested parallelism will be workers creating a `DaskDistributedBackend`, with no
'scheduler_host'. To avoid deadlocks, worker threads will secede and rejoin.

cc @ogrisel.

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -9,7 +9,7 @@ from tornado import gen
 
 from .client import Client, _wait
 from .utils import ignoring, funcname, itemgetter
-from . import get_client, secede
+from . import get_client, secede, rejoin
 
 # A user could have installed joblib, sklearn, both, or neither. Further, only
 # joblib >= 0.10.0 supports backends, so we also need to check for that. This
@@ -186,7 +186,11 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
             pass
 
         yield
-        # No need to rejoin here
+
+        try:
+            rejoin()
+        except AttributeError:
+            pass
 
 
 for base in _bases:
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -1,5 +1,7 @@
 from __future__ import print_function, division, absolute_import
 
+import contextlib
+
 from distutils.version import LooseVersion
 from uuid import uuid4
 
@@ -7,6 +9,7 @@ from tornado import gen
 
 from .client import Client, _wait
 from .utils import ignoring, funcname, itemgetter
+from . import get_client, secede, rejoin
 
 # A user could have installed joblib, sklearn, both, or neither. Further, only
 # joblib >= 0.10.0 supports backends, so we also need to check for that. This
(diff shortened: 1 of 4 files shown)


mrocklin:
I wonder if maybe we should just use `get_client()` all the time.  Perhaps it was a mistake to create a client within this wrapper.  Perhaps we should expect the user to create a client externally and either pass that in or else depend on the `get_client()` global default.  Thoughts?

distributed/joblib.py
```diff
@@ -130,17 +157,40 @@ def callback_wrapper():
...
+            secede()
+        except ValueError:
+            pass
+
+        yield
+
+        try:
+            rejoin()
+        except AttributeError:
+            pass
```
mrocklin: Should we rejoin here?  Typically this is necessary if there is additional computational work to do in the task.  That may not be the case here.  It's harmless either way, but might add a tiny bit of unwanted delay.

distributed/tests/test_joblib.py
```diff
@@ -140,3 +145,38 @@ def test_joblib_scatter(loop, joblib):
...
     assert y.count == 1
     assert z.count == 4
+
+
+@pytest.mark.parametrize('joblib', joblibs)
+def test_nested_backend_context_manager(loop, joblib):
+    if joblib is None or LooseVersion(joblib.__version__) <= "0.11.0":
+        pytest.skip("Joblib >= 0.11.1 required.")
+    Parallel = joblib.Parallel
+    delayed = joblib.delayed
```
mrocklin: Slight preference to leave these as fully namespaced, especially given the double use of `delayed`

distributed/tests/test_joblib.py
```diff
@@ -140,3 +145,38 @@ def test_joblib_scatter(loop, joblib):
...
+
+    def get_nested_pids():
+        return Parallel(n_jobs=2)(delayed(os.getpid)() for _ in range(2))
+
+
+    with cluster() as (s, [a, b]):
+        with joblib.parallel_backend('dask.distributed', loop=loop,
+                                     scheduler_host=s['address']) as (ba, _):
```
mrocklin: Right, so about the "lets just always use the default client" argument from above, this would become the following:

```python
with Client(s['address'], loop=loop) as client:
    with joblib.parallel_backend('dask') as (ba, _):
        ...
```

distributed/joblib.py
TomAugspurger: I wasn't sure about that either. To make sure I understand, when the worker thread `secedes`, the scheduler will tell the worker to make a new thread to take its place? In that case, yes we should be OK just not rejoining.

mrocklin: Right, the thread leaves the thread pool, leaving it with `n - 1` threads.  We make a new thread to take it's place.  Relevant code here: https://github.com/dask/distributed/blob/master/distributed/threadpoolexecutor.py#L94-L105

```python
def secede(adjust=True):
    """ Have this thread secede from the ThreadPoolExecutor
    See Also
    --------
    rejoin: rejoin the thread pool
    """
    thread_state.proceed = False
    with threads_lock:
        thread_state.executor._threads.remove(threading.current_thread())
        if adjust:
            thread_state.executor._adjust_thread_count()
```

TomAugspurger:
>  Perhaps we should expect the user to create a client externally and either pass that in or else depend on the get_client() global default.

Ah, I was getting a bit confused earlier about when `get_client` was raising a ValueError. I was thinking the `ValueError` only came when you were on a worker, but it could also be when the user hasn't created a client. I'll think more about requiring an existing client. I'm 50/50 right now.

Maybe @jcrist has thoughts here too.

mrocklin:
> Ah, I was getting a bit confused earlier about when get_client was raising a ValueError. I was thinking the ValueError only came when you were on a worker, but it could also be when the user hasn't created a client. I'll think more about requiring an existing client. I'm 50/50 right now.

I think that we can evolve into this API somewhat smoothly.  I might suggest the following interface:

```python
def __init__(self, scheduler_host=None, client=None, **kwargs):
    if client is None:
        if scheduler_host is not None:
            client = Client(scheduler_host, **kwargs)
        else:
            client = get_client()
    ...
```

mrocklin:
I've pushed my proposed changes to this fork/branch.  Please feel free to accept or reject them as you like.

mrocklin:
+1 from me

Also it looks like we need some of these changes due to a new release of joblib?

TomAugspurger:
The change at
https://github.com/dask/distributed/pull/1705/files#diff-bf3f88e462cfc84fc9199b072c6b4e6bR66
will be required for the new joblib.

Everything else changing in joblib should be backwards compatible.

distributed/joblib.py
```diff
@@ -136,11 +163,31 @@ def callback_wrapper():
...
+        # See 'joblib.Parallel.__call__' and 'joblib.Parallel.retrieve' for how
+        # this is used.
+        try:
+            secede()
+        except ValueError:
+            # We are not a worker, i.e. not in nested parallelism.
+            pass
+
+        yield
+        # No need to rejoin here
```
ogrisel: Why? Won't that create too many threads on the distributed worker nodes?

Also, is it valid to call `secede()` several times in the same thread? That would happen if there are several consecutive calls to `Parallel()` in a nested function. We should probably add a test for that case.

distributed/tests/test_joblib.py
```diff
@@ -140,3 +135,46 @@ def test_joblib_scatter(loop, joblib):
...
+
+@pytest.mark.parametrize('joblib', joblibs)
+def test_nested_backend_context_manager(loop, joblib):
+    if joblib is None or LooseVersion(joblib.__version__) <= "0.11.0":
+        pytest.skip("Joblib >= 0.11.1 required.")
+    Parallel = joblib.Parallel
+    delayed = joblib.delayed
+
+    def get_nested_pids():
+        return Parallel(n_jobs=2)(delayed(os.getpid)() for _ in range(2))
```
ogrisel: It would be more interesting to have:

```python
def get_nested_pids():
    pids = set(Parallel(n_jobs=2)(delayed(os.getpid)() for _ in range(2)))
    return pids.union(Parallel(n_jobs=2)(delayed(os.getpid)() for _ in range(2)))
```

distributed/joblib.py
mrocklin: If we're likely to do more work in this thread after completing then yes, we should `rejoin`.

@ogrisel I suspect that you're thinking of cases like the following?

```python
results = Parallel()(...)

# do work

results2 = Parallel()(...)
```

distributed/tests/test_joblib.py
mrocklin: This was a good test.  It failed before and passes with the most recent commit.

distributed/joblib.py
ogrisel: yes

Issue #1702, opened 2018-01-22
Problems with client.compute for multi-dimensional collections with resource


caseyjlaw:
The instructions for [computing on collections](http://distributed.readthedocs.io/en/latest/resources.html) define a syntax for using specific resources, such as:
`z.compute(resources={tuple(y._keys()): {'GPU': 1}}, optimize_graph=False)`

It seems these instructions fail for collections that are multi-dimensional. For example:
```
> dd = delayed(np.random.randint)(0, 10, (10,10))
> arr = dask.array.from_delayed(dd, (10, 10), int)
> future = cl.compute(arr, resources={tuple(arr.__dask_keys__()): {'GPU': 1}}, optimize_graph=False)
TypeError: unhashable type: 'list'
```

This is fixed if the keys are indexed ndim-1 times:

```
> future = cl.compute(arr, resources={tuple(arr.__dask_keys__()[0]): {'READER': 1}}, optimize_graph=False)

> future
Out[35]: <Future: status: finished, type: ndarray, key: finalize-3bb181ad379f56fc58168aeb1b16f918>

> future.result()
Out[36]: 
array([[4, 8, 2, 9, 8, 2, 9, 4, 4, 2],
       [8, 1, 1, 5, 2, 0, 5, 4, 6, 3],
       [5, 3, 6, 3, 1, 6, 4, 3, 8, 5],
       [3, 8, 0, 6, 7, 2, 4, 8, 1, 0],
       [7, 0, 6, 8, 6, 9, 8, 2, 5, 1],
       [7, 0, 4, 1, 4, 8, 5, 9, 8, 9],
       [2, 9, 0, 3, 1, 5, 6, 5, 6, 3],
       [2, 5, 7, 2, 2, 2, 8, 2, 0, 4],
       [2, 6, 6, 0, 6, 3, 2, 0, 8, 6],
       [0, 8, 0, 3, 7, 3, 6, 6, 5, 1]])
```

Furthermore, it is possible to require ill-formatted resources that are silently ignored. Here is an example with a resource that does not exist (no 'GPU' set for workers in this case):

```
> future = cl.compute(arr, resources={arr: {'GPU': 1}}, optimize_graph=False)

> future
Out[41]: <Future: status: finished, type: ndarray, key: finalize-56585de7b2c9bd58e45431f2891ab592>

> future.result()
Out[42]: 
array([[2, 1, 3, 2, 9, 9, 1, 4, 8, 3],
[...]

PR #1706, opened 2018-01-23
compute resources handle multi-dimensional collections

caseyjlaw:
Adds a check for bad input and also flattens the dask_keys list to properly handle multi-dimensional collections.
See also #1702.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3036,14 +3036,17 @@ class Client(Node):
         Expand a user-provided task key specification, e.g. in a resources
         or retries dictionary.
         """
-        if not isinstance(k, tuple):
-            k = (k,)
-        for kk in k:
-            if dask.is_dask_collection(kk):
-                for kkk in flatten(kk.__dask_keys__()):
-                    yield tokey(kkk)
-            else:
-                yield tokey(kk)
+        if isinstance(k, tuple) or isinstance(k, str) or dask.is_dask_collection(k):
+            if not isinstance(k, tuple):
+                k = (k,)
+            for kk in k:
+                if dask.is_dask_collection(kk):
+                    for kkk in flatten(kk.__dask_keys__()):
+                        yield tokey(kkk)
+                else:
+                    yield tokey(kk)
+        else:
+            raise TypeError("key should be tuple, string, or dask collection")
 
     @classmethod
     def _expand_retries(cls, retries, all_keys):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3040,7 +3040,7 @@ class Client(Node):
             k = (k,)
         for kk in k:
             if dask.is_dask_collection(kk):
-                for kkk in list(flatten([kk])):
+                for kkk in flatten(kk.__dask_keys__()):
                     yield tokey(kkk)
             else:
                 yield tokey(kk)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3037,10 +3037,18 @@ class Client(Node):
         or retries dictionary.
         """
         if not isinstance(k, tuple):
+            if not dask.is_dask_collection(k):
+                raise TypeError("key should be a tuple of dask collection")
             k = (k,)
         for kk in k:
             if dask.is_dask_collection(kk):
-                for kkk in kk.__dask_keys__():
+
+                # flatten nested lists for multi-dim data
+                keys = kk.__dask_keys__()
+                while isinstance(keys[0], list):
+                    keys = keys[0]
+
+                for kkk in keys:
                     yield tokey(kkk)
             else:
                 yield tokey(kk)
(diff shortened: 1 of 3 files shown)


distributed/client.py
```diff
@@ -3037,10 +3037,18 @@ def _expand_key(cls, k):
...
+                raise TypeError("key should be a tuple of dask collection")
             k = (k,)
         for kk in k:
             if dask.is_dask_collection(kk):
-                for kkk in kk.__dask_keys__():
+
+                # flatten nested lists for multi-dim data
+                keys = kk.__dask_keys__()
+                while isinstance(keys[0], list):
+                    keys = keys[0]
```
mrocklin: What about `keys[1]`?

Also, you might want to use `dask.core.flatten`

distributed/client.py
```diff
@@ -3037,10 +3037,18 @@ def _expand_key(cls, k):
         or retries dictionary.
         """
         if not isinstance(k, tuple):
+            if not dask.is_dask_collection(k):
+                raise TypeError("key should be a tuple of dask collection")
```
mrocklin: This isn't true.  Users are also allowed to put actual keys here as well.

caseyjlaw: Ah, ok.
By my reading, any kind of object can be input and it will be wrapped in a tuple. However, if that object can't be used to make a proper key, the requirement will be silently ignored.
Is there a definition of a proper key?

distributed/client.py
```diff
@@ -3040,7 +3040,7 @@ def _expand_key(cls, k):
             k = (k,)
         for kk in k:
             if dask.is_dask_collection(kk):
-                for kkk in kk.__dask_keys__():
+                for kkk in list(flatten([kk])):
```
mrocklin: You probably want to continue using `kk.__dask_keys__()` like you were before.  The `flatten` function is there just to collapsed the nested list of lists to a single list.

mrocklin: And sorry for the delay on review.  I appreciate your contribution here and your patience.

caseyjlaw: No problem at all. I'm amazed that you can keep track of everything going on on this project!
My new test is still failing because it is looking for a key for the dask.array.from_delayed() object. However, I can't figure out what that key looks like in worker.data. Any tip on how to check that?

distributed/tests/test_resources.py
```diff
@@ -261,3 +261,20 @@ def test_dont_optimize_out(c, s, a, b):
...
+@gen_cluster(client=True, ncores=[('127.0.0.1', 1, {'resources': {'A': 1}}),
+                                  ('127.0.0.1', 1, {'resources': {'B': 1}})])
+def test_compute_multidim(c, s, a, b):
+    da = pytest.importorskip('dask.array')
+    np = pytest.importorskip('numpy')
+    x = delayed(np.random.randint)(0, 1, (5,5))
+    y = da.from_delayed(x, (5,5), int)
+
+    xx = c.compute(x, resources={x: {'A': 1}},)
+    yy = c.compute(y, resources={y: {'A': 1}},)
```
mrocklin: Looking at the test failures in travis-ci it seems like this is illegal because `y` is not hashable?

```python
    @gen_cluster(client=True, ncores=[('127.0.0.1', 1, {'resources': {'A': 1}}),
                                      ('127.0.0.1', 1, {'resources': {'B': 1}})])
    def test_compute_multidim(c, s, a, b):
        da = pytest.importorskip('dask.array')
        np = pytest.importorskip('numpy')
        x = delayed(np.random.randint)(0, 1, (5,5))
        y = da.from_delayed(x, (5,5), int)
    
        xx = c.compute(x, resources={x: {'A': 1}},)
>       yy = c.compute(y, resources={y: {'A': 1}},)
E       TypeError: unhashable type: 'Array'
```

caseyjlaw: That is not surprising, but the weird thing is that the travis-ci test fails in a different way. The travis-ci test fails in a way that I can reproduce locally, but I don't understand `data` property of the worker.
I've removed the assertion related to dask.array. It passes locally now.

distributed/client.py
mrocklin: Keys get transformed by `distributed.utils.tokey` which mostly just converts them into strings.

distributed/tests/test_resources.py
mrocklin: That assertion was the main reason for this PR.  We genuinely can't pass `resources={x: ...}` where x is a dask array because dask arrays are not hashable.  The flatten logic you've added in this PR probably needs to be called on the user side.  Alternatively, we could introduce some sort of list of key-value-pairs alternative input to `resources=` to handle non-hashable keys.

caseyjlaw:
Thanks for looking into this. I have not had much luck developing a good test, so I'll close the PR.

PR #1710, opened 2018-01-26
TST: Added another nested parallelism test

TomAugspurger:
Here's a workload @ogrisel and I developed when (unnecessarily) debugging https://github.com/joblib/joblib/issues/615

It's a useful test to demonstrate that our backend is indeed overridden by the `ThreadingBackend` when the `Parallel` call says that it requires shared memory. The test ensure that follows through a chain of Parallel calls with prefers -> requires -> function.


distributed/tests/test_joblib.py
```diff
@@ -180,3 +180,42 @@ def test_errors(loop, joblib):
...
+            # Require threads, should be threading
+            with joblib.parallel_backend('dask') as (ba, _):
+                result = joblib.Parallel(n_jobs=2)(joblib.delayed(outer)(
+                    joblib, require='sharedmem') for _ in range(1))
+                assert isinstance(result[0][0][0],
+                                  joblib.parallel.ThreadingBackend)
+
+
+
+
```
ogrisel: PEP8: 2 blank lines instead of 4

distributed/tests/test_joblib.py
```diff
@@ -180,3 +180,42 @@ def test_errors(loop, joblib):
...
+def outer(joblib, require):
+    return joblib.Parallel(n_jobs=2, prefer='threads')(
+        joblib.delayed(middle)(joblib, require) for _ in range(1)
+    )
+
```
ogrisel: PEP8: 2 blank lines instead of 1.

distributed/tests/test_joblib.py
```diff
@@ -180,3 +180,42 @@ def test_errors(loop, joblib):
...
+            with joblib.parallel_backend('dask') as (ba, _):
+                result = joblib.Parallel(n_jobs=2)(joblib.delayed(outer)(
+                    joblib, require='sharedmem') for _ in range(1))
+                assert isinstance(result[0][0][0],
+                                  joblib.parallel.ThreadingBackend)
+
+
+
+
+def outer(joblib, require):
```
ogrisel: Maybe `require` should be renamed to `nested_require` to make it explicit that this will be used in a nested parallel call and not in the outer calls.

ogrisel:
Actually this cannot be merged before joblib is released with the new backend hints and constraints API from https://github.com/joblib/joblib/pull/602.

ogrisel:
Also this test should probably be protected to be skipped with joblib < 0.12.

distributed/tests/test_joblib.py
mrocklin: I've re-enabled flake8 testing on test files

TomAugspurger:
Updated.

@ogrisel what's the plan for the next joblib version? I'm currently skipping these tests for `joblib <= '0.11.0'`. My local version is at `'0.11.1.dev0'`. Will the nested parallelism changes be in 0.11.1 or 0.12.0?

ogrisel:
I have a bug to fix and will probably release a beta tomorrow.

ogrisel:
I fixed my bug but I still have a random failure on our CI and I won't have time to investigate this week. Maybe next week for the joblib release.

distributed/tests/test_joblib.py
```diff
@@ -176,6 +176,43 @@ def test_errors(loop, joblib):
     assert "create a dask client" in str(info.value).lower()
 
 
+def test_correct_nested_backend(loop, joblib):
+    if LooseVersion(joblib.__version__) <= "0.11.0":
```
ogrisel: Should be `LooseVersion("0.11.0")` I think.

ogrisel:
joblib 0.12 was released a while ago so this PR could be completed now.

ogrisel:
Some tests are randomly failing but those are not the new joblib connector tests.

TomAugspurger:
Thanks for fixing this up!

Issue #1376, opened 2017-09-05
Check versions on connection


mrocklin:
A decent fraction of the errors that we see (15%?) are due to mismatched versions.  We might consider silently running `client.get_versions(check=True)` on startup and issuing a warning if things don't match well.

jonnor:
Added a comment about one occurrence of this issue on the PR: https://github.com/dask/distributed/pull/1720#issuecomment-425733996

costrouc:
Yes please this would be extremely helpful. When running dask on Kubernetes it is too easy to get a version mismatch from separate conda solves for each docker image (dask worker, dask scheduler, jupyterlab user) without pinning.

jrbourbeau:
I think this was added in https://github.com/dask/distributed/pull/2627 and is included in `distributed >= 2.9.2`

jrbourbeau:
Just confirmed that a warning is logged when there are package version mismatches (when using the current `distributed` master branch). For example:

```
distributed.worker - WARNING - Mismatched versions found

distributed
+---------------------------+---------------------------+
|                           | version                   |
+---------------------------+---------------------------+
| This Worker               | 2.11.0+21.g05ca3fb1.dirty |
| scheduler                 | 2.12.0                    |
| tcp://192.168.0.102:64357 | 2.11.0+21.g05ca3fb1.dirty |
+---------------------------+---------------------------+

msgpack
+---------------------------+---------+
|                           | version |
+---------------------------+---------+
| This Worker               | 0.6.1   |
| scheduler                 | 1.0.0   |
| tcp://192.168.0.102:64357 | 0.6.1   |
+---------------------------+---------+
```

jrbourbeau:
Closed by https://github.com/dask/distributed/pull/2627

PR #1720, opened 2018-01-29
Add check_versions= keyword to Client

mrocklin:
This calls get_versions(check=True) when starting up the clien

See #1376

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -456,6 +456,9 @@ class Client(Node):
         the scheduler for matters relating to this client
     heartbeat_interval: int
         Time in milliseconds between heartbeats to scheduler
+    check_versions: bool (True)
+        Whether or not to check versions of local environment against the
+        workers.  Good to ensure safety, but slightly delays startup time.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


mrocklin:
It's not clear that this is a good idea as it is currently implemented.  We are actually ok with a decent amount of version skew between client and scheduler.  I've left the default as False for now, but I'm afraid that this will never get used if we don't actually turn it on.  I'm tempted to merge this with `check=True` after the upcoming release and let people play with it a bit on the dev version.

jonnor:
Just experienced a problem which seem to have been due to compatibility issues with different version on Scheduler and Client. Had 0.18.1 on scheduler (default now on Kubernetes via Helm), and 0.19.2 on client (default now on pip/conda). When I updated on the cluster, issue went away.

Would have been nice if there would be some hints about this. If an error is too much, maybe Client could  warn on (minor) version mismatch, or maybe print Client/Scheduler versions in the interactive representation.

Example error below:

```
/usr/lib/python3.7/site-packages/distributed/client.py in _gather(self, futures, errors, direct, local_worker)
   1446                         try:
-> 1447                             st = self.futures[key]
   1448                             exception = st.exception
KeyError: "('concatenate-2994d66cd08840071f42891a2be9ebc4', 131, 0)"

CancelledError                            Traceback (most recent call last)
<ipython-input-169-569b98d17bdd> in <module>()
      1 #train_F = train_F.persist()
----> 2 train_X = train_F.compute()
      3 train_X[0,:]
...
/usr/lib/python3.7/site-packages/six.py in reraise(tp, value, tb)
    691             if value.__traceback__ is not tb:
    692                 raise value.with_traceback(tb)
--> 693             raise value
    694         finally:
    695             value = None

CancelledError: ('concatenate-2994d66cd08840071f42891a2be9ebc4', 131, 0)
```

mrocklin:

Issue #1740, opened 2018-02-03
IPython test failures


mrocklin:
We've had an intermittent test failure for the IPython code for a long while.  Here is a traceback

```python
=================================== FAILURES ===================================
__________________________ test_start_ipython_remote ___________________________
loop = <tornado.platform.epoll.EPollIOLoop object at 0x7fd1842a12e8>
zmq_ctx = <zmq.sugar.context.Context object at 0x7fd165bc74a8>
    @pytest.mark.ipython
    @need_functional_ipython
    def test_start_ipython_remote(loop, zmq_ctx):
        from distributed._ipython_utils import remote_magic
        with cluster(1) as (s, [a]):
            with Client(s['address'], loop=loop) as e, mock_ipython() as ip:
                worker = first(e.ncores())
                ip.user_ns['info'] = e.start_ipython_workers(worker)[worker]
                remote_magic('info 1')  # line magic
>               remote_magic('info', 'worker')  # cell magic
distributed/tests/test_ipython.py:143: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
distributed/_ipython_utils.py:140: in remote_magic
    run_cell_remote(ip, kc, cell)
distributed/_ipython_utils.py:48: in run_cell_remote
    msg = kc.get_iopub_msg(timeout=OUTPUT_TIMEOUT)
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/jupyter_client/client.py:81: in get_iopub_msg
    return self.iopub_channel.get_msg(*args, **kwargs)
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/jupyter_client/blocking/channels.py:50: in get_msg
    ready = self.socket.poll(timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <zmq.sugar.socket.Socket object at 0x7fd165dceac8>, timeout = 10000
flags = 1
    def poll(self, timeout=None, flags=POLLIN):
        """poll the socket for events
    
            The default is to poll forever for incoming
            events.  Timeout is in milliseconds, if specified.
    
            Parameters
            ----------
            timeout : int [default: None]
                The timeout (in milliseconds) to wait for an event. If unspecified
[...]

PR #1743, opened 2018-02-07
cleanup remote_magic client cache after tests

minrk:
should avoid stale clients being reused during tests

This *might* help with #1740


mrocklin:
We'll see.  Thanks for looking into this @minrk !

Issue #1687, opened 2018-01-11
Using interface=ib0 interferes with serving bokeh


caseyjlaw:
(After asking around with my sysadmins, I think this is not just related to my network configuration. But I could be wrong.)

I am using infiniband (--interface=ib0 for dask-scheduler and dask-worker) on a small linux cluster. The only change I needed was to refer to the host by its IP rather than IP, as identified by at scheduler startup. After that, I can see all processes start well and seem to run tasks, etc..

However, after this change I cannot find the bokeh web server. The scheduler reports an IP address and port for the bokeh server. The IP address is the same as that of the scheduler and the port is 1 higher, as usual. But using my usual ssh tunneling to that address:port does not expose the bokeh server. 

My sysadmins noted that the infiniband is configured to handle TCP/IP traffic, but that is only intended to communicate with other devices on infiniband. Outside traffic is not supported. Their suggestion is to explcitly tell the web server the device to listen to. Is this supported by an existing scheduler option?

mrocklin:
I agree with your sysadmins and think that this *should* be supported.  Unfortunately it currently is not.  

The offending line is here: https://github.com/dask/distributed/blob/7b987c2ba29ba2fe0acc7f7dd087d351ee076129/distributed/scheduler.py#L985

We should be able to pass in that `listen_ip` address from the command line.  This will require snaking it in through a few function calls.  This will require changing the `Scheduler(services=)` keyword in such a way that provides adding this option and is also backwards compatible.

If you have *any* interest in taking this on that would be quite welcome.

rabernat:
I encountered this same error on my local cluster. I worked around it by *not* using infiniband for the scheduler but continuing to use it for the workers.

caseyjlaw:
@rabernat Thanks for that. I don't have a lot of data moving to my scheduler, so that works for me.
I looked over the changes, but don't feel comfortable enough with a PR. I'll go ahead without ib0 on the scheduler.

caseyjlaw:
@rabernat I definitely feel encouraged to submit a PR. After reviewing the code, I don't think I understand the theory well enough to do so.

PR #1744, opened 2018-02-07
Allow service ports to be specified as a (host, port) instead

bmerry:
My main use case is to pass `diagnostics_port=(host, port)` to
LocalCluster to run a world-accessible Bokeh server even when the
scheduler uses inproc://. However, the design should be more general
e.g. it is a step towards dask/distributed#1687.

I haven't written any tests or documented it yet - I'd like to get
feedback on whether this is the right approach before investing in
those.

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1045,51 +1045,29 @@ def test_correct_bad_time_estimate(c, s, *workers):
 
 
 @pytest.mark.skipif(not sys.platform.startswith('linux'),
-                    reason="Need 127.0.0.2 to mean localhost")
-@gen_test(timeout=None)
-def test_service_hosts_match_scheduler():
-    pytest.importorskip('bokeh')
-    from distributed.bokeh.scheduler import BokehScheduler
-    services = {('bokeh', 0): BokehScheduler}
-
-    s = Scheduler(services=services)
-    yield s.start('tcp://0.0.0.0')
-
-    sock = first(s.services['bokeh'].server._http._sockets.values())
-    assert sock.getsockname()[0] in ('::', '0.0.0.0')
-    yield s.close()
-
-    for host in ['tcp://127.0.0.2', 'tcp://127.0.0.2:38275']:
-        s = Scheduler(services=services)
-        yield s.start(host)
-
-        sock = first(s.services['bokeh'].server._http._sockets.values())
-        assert sock.getsockname()[0] == '127.0.0.2'
-        yield s.close()
-
-
-@pytest.mark.skipif(not sys.platform.startswith('linux'),
                     reason="Need 127.0.0.* to mean localhost")
-@gen_test(timeout=None)
-def test_service_custom_host():
+@gen_test()
+def test_service_hosts():
     pytest.importorskip('bokeh')
     from distributed.bokeh.scheduler import BokehScheduler
-    services = {('bokeh', ('127.0.0.3', 0)): BokehScheduler}
 
-    s = Scheduler(services=services)
-    yield s.start('tcp://0.0.0.0')
+    for port in [0, ('127.0.0.3', 0)]:
+        for url, expected in [('tcp://0.0.0.0', ('::', '0.0.0.0')),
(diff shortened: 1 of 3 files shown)


mrocklin:
In principle this seems reasonable to me.

mrocklin:
My apologies for the delay in response.  I'm glad that you're taking this on.

mrocklin:
It would be good to see things like tests and doc changes around this code change.

bmerry:
> It would be good to see things like tests and doc changes around this code change.

Sure, I'll get to that next now that you're said you're happy in principle.

I haven't started looking at the testing framework yet. Can you suggest which area I should look at to add tests?

bmerry:
I've added documentation for the `LocalCluster` use case. I'm very new to distributed and haven't read most of the docs, so I'm not sure if there are other places the documentation needs to be extended.

I added a test as well. It's a bit of a copy-paste job from the test immediately above it. Let me know if you'd prefer me to try to combine them into one parametrized test.

distributed/tests/test_scheduler.py
```diff
@@ -1068,6 +1068,30 @@ def test_service_hosts_match_scheduler():
...
+def test_service_custom_host():
+    pytest.importorskip('bokeh')
+    from distributed.bokeh.scheduler import BokehScheduler
+    services = {('bokeh', ('127.0.0.3', 0)): BokehScheduler}
+
+    s = Scheduler(services=services)
+    yield s.start('tcp://0.0.0.0')
+
+    sock = first(s.services['bokeh'].server._http._sockets.values())
+    assert sock.getsockname()[0] in ('::', '0.0.0.0')
```
mrocklin: I'm curious, why wouldn't this be the hostname that we specified above?  (My understanding of networking here is poor)

distributed/tests/test_scheduler.py
```diff
@@ -1068,6 +1068,30 @@ def test_service_hosts_match_scheduler():
         yield s.close()
 
 
+@pytest.mark.skipif(not sys.platform.startswith('linux'),
+                    reason="Need 127.0.0.* to mean localhost")
+@gen_test(timeout=None)
```
mrocklin: Can we remove `timeout=None` to re-enable the standard 10s timeout?

distributed/tests/test_scheduler.py
bmerry: You're right, it should, and it's confirmed by the Travis failure.

It seems that when I run `py.test distributed --verbose` it's silently exiting during `distributed/deploy/tests/test_ssh.py::test_cluster` without indicating any sort of failure, and hadn't even reached my new test. So I was falsely assuming it was passing.

distributed/tests/test_scheduler.py
bmerry: I can - it was copy-pasted from the test above it, which is very similar. Should I remove it from both tests?

mrocklin: Yes, that sounds good

distributed/tests/test_scheduler.py
```diff
@@ -1068,6 +1068,30 @@ def test_service_hosts_match_scheduler():
...
+    assert sock.getsockname()[0] == '127.0.0.3'
+    yield s.close()
+
+    for host in ['tcp://127.0.0.2', 'tcp://127.0.0.2:38275']:
+        s = Scheduler(services=services)
+        yield s.start(host)
+
+        sock = first(s.services['bokeh'].server._http._sockets.values())
+        assert sock.getsockname()[0] == '127.0.0.3'
+        yield s.close()
```
mrocklin: You might also be able to clean up this test a bit, perhaps by merging the for loop with the example above?  Also, does it make sense to try with each of these scheduler start hosts in particular?  I get the copy-paste history, but we should also do some thinking about cases that might be more relevant to your situation.

bmerry: I've given it another go, merging it with the test from which it was copy-pasting, and also reducing the amount of copy-paste in that test while I'm at it.

mrocklin:
This seems good to me.  Merging.  Thanks for the analysis and for the fix @bmerry !

PR #1756, opened 2018-02-13
Add graph layout engine and bokeh plot

mrocklin:
A real-time graph layout system.  I expect this to scale into the low thousands of nodes real-time.  [Screencast](https://youtu.be/jBS10KCyRZM)

graph-layout


mrocklin:
cc @shughes-uk you've asked for things like this in the past

shughes-uk:
Can't wait!

mkhoshle:
How is the graph telling that it is waiting and what does memory mean? Does that mean that it is waiting for memory to fetch the data and bring it to first level cache? 
Is it possible to know what is the root-cause for the wait (gray bar)?

mrocklin:
Those colors refer to the state of each task in the scheduler.  See
http://distributed.readthedocs.io/en/latest/scheduling-state.html#task-state

mkhoshle:
Thanks!

PR #1757, opened 2018-02-13
Use new LZ4 API

### LINKED ISSUE #1755: lz4 new api
I believe the code here is faulty: https://github.com/dask/distributed/blob/cc908e7725410278c3a2ba08bd05675c186479e1/distributed/protocol/compression.py#L56-66. You need to import lz4.block in order to use the new api. Right now this always fall back. This is how I would fix it.

```python
with ignoring(ImportError):
    import lz4

    try:
        # try using the new lz4 API
        import lz4.block
        lz4_compress = lz4.block.compress
        lz4_decompress = lz4.block.decompress
    except ImportError:
        # fall back to old one
        lz4_compress = lz4.LZ4_compress
        lz4_decompress = lz4.LZ4_uncompress

thrasibule:
closes #1755

### DIFF
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -58,9 +58,10 @@ with ignoring(ImportError):
 
     try:
         # try using the new lz4 API
+        import lz4.block
         lz4_compress = lz4.block.compress
         lz4_decompress = lz4.block.decompress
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @thrasibule !  Merging

PR #1766, opened 2018-02-15
Use set.discard to avoid KeyErrors in stealing

mrocklin:
See https://github.com/pangeo-data/pangeo/issues/99#issuecomment-365201840

I can't figure out why this is necessary, but it seems to be.  cc @jhamman


jhamman:
Thanks @mrocklin, I'll give it a try today.

PR #1774, opened 2018-02-20
[WIP] - Submit priority

mrocklin:
Repeated calls to submit should not base their priority on order, but rather on graph structure.


mrocklin:

PR #1777, opened 2018-02-20
specify IOLoop for stealing PeriodicCallback

mrocklin:
Previously this would use IOLoop.current, which, when deploying from a Jupyter notebook, was the notebooks IOLoop.  This would turn off work stealing during `compute` calls.

Fixes https://github.com/pangeo-data/pangeo/issues/118

### DIFF
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -45,7 +45,8 @@ class WorkStealing(SchedulerPlugin):
             self.add_worker(worker=worker)
 
         pc = PeriodicCallback(callback=self.balance,
-                              callback_time=100)
+                              callback_time=100,
+                              io_loop=self.scheduler.loop)
         self._pc = pc
         self.scheduler.periodic_callbacks['stealing'] = pc
         self.scheduler.plugins.append(self)


distributed/stealing.py
```diff
@@ -45,7 +45,8 @@ def __init__(self, scheduler):
             self.add_worker(worker=worker)
 
         pc = PeriodicCallback(callback=self.balance,
-                              callback_time=100)
+                              callback_time=100,
+                              io_loop=self.scheduler.loop)
```
jakirkham: Am surprised this works with Tornado 5.0. Looks like they dropped the [`io_loop` argument to `PeriodicCallback`]( http://www.tornadoweb.org/en/branch5.0/ioloop.html#tornado.ioloop.PeriodicCallback ).

Having a similar discussion about problems this causes for Tornado 5.0 usage in PR ( https://github.com/dask/dask-drmaa/pull/61 ).

mrocklin: from distributed.utils import PeriodicCallback

jakirkham: Got it. Thanks. :)

PR #1778, opened 2018-02-21
add to_scheduler_file to Client

### LINKED ISSUE #1741: More convenient way to create the scheduler-file
The concept of a "scheduler file" is a nice, convenient way to share scheduler information when working on a system where you won't know the scheduler address a priori. I would like to suggest a simple method, probably on `Client`, that creates such a "scheduler file" with a simple interface like this:

```Python
client = Client()
client.to_scheduler_file('path-to-a-file.json')
```

then from another session one could easily connect to that same set of resources via:

```Python
client = Client(scheduler_file='path-to-a-file.json')
```

Under the hood this would just be something like 

```Python
json.dump(client.scheduler_info(), 'path-to-a-file.json', indent=2)
```

jhamman:
This adds a new method to `Client` that writes a "scheduler_file". This file can be used when instantiating new Clients using the `scheduler_file` keyword argument.  This is purely a convenience method that facilitates sharing scheduler information via a filesystem. Example usage:

In one python session:
```Python
client = Client()
client.to_scheduler_file('test.json')
```

In a second session:
```Python
client = Client(scheduler_file='test.json')
```

closes #1741

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2767,8 +2767,25 @@ class Client(Node):
         self.sync(self._update_scheduler_info)
         return self._scheduler_identity
 
-    def to_scheduler_file(self, scheduler_file):
-        """Dump the info in this file to a scheduler file"""
+    def write_scheduler_file(self, scheduler_file):
+        """ Write the scheduler information to a json file.
+
+        This facilitates easy sharing of scheduler information using a file
+        system. The scheduler file can be used to instantiate a second Client
+        using the same scheduler.
+
+        Parameter
+        ---------
+        scheduler_file: str
+            Path to a write the scheduler file.
+
+        Examples
+        --------
+        >>> client = Client()
+        >>> client.to_scheduler_file('scheduler.json')
+        # connect to previous client's scheduler
+        >>> client2 = Client(scheduler_file='scheduler.json')
+        """
 
         if self.scheduler_file:
             raise ValueError('Scheduler file already set')
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2782,11 +2782,10 @@ class Client(Node):
         Examples
         --------
         >>> client = Client()
-        >>> client.to_scheduler_file('scheduler.json')
+        >>> client.write_scheduler_file('scheduler.json')
         # connect to previous client's scheduler
         >>> client2 = Client(scheduler_file='scheduler.json')
         """
-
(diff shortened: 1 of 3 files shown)


jhamman:
As far as I can tell, the test failure on Travis is not related to this PR.

distributed/tests/test_client.py
```diff
@@ -3716,6 +3717,22 @@ def test_scheduler_info(loop):
...
 
+def test_to_scheduler_file(loop):
+    with cluster() as (s, [a, b]):
+        with Client(s['address'], loop=loop) as c:
+            info = c.scheduler_info()
+            with tmpfile('json') as scheduler_file:
+                c.to_scheduler_file(scheduler_file)
+                with Client(scheduler_file=scheduler_file) as c2:
+                    info2 = c2.scheduler_info()
+                    assert info['address'] == info2['address']
```
mrocklin: I recommend that we test `c.scheduler.address == c2.scheduler.address`

distributed/client.py
```diff
@@ -2767,6 +2767,17 @@ def scheduler_info(self, **kwargs):
         self.sync(self._update_scheduler_info)
         return self._scheduler_identity
 
+    def to_scheduler_file(self, scheduler_file):
+        """Dump the info in this file to a scheduler file"""
```
mrocklin: Unless we're going to write stand-alone documentation about this somewhere (and this seems small enough to me that we can skip it) we should expand this docstring to include the motivation to use this as well as a small example, maybe like the following:

```python
client = Client()
client.to_scheduler_file('scheduler.json')

client2 = Client(scheduler_file='scheduler.json')  # connect to previous client's scheduler
```

distributed/client.py
```diff
@@ -2767,6 +2767,17 @@ def scheduler_info(self, **kwargs):
         self.sync(self._update_scheduler_info)
         return self._scheduler_identity
 
+    def to_scheduler_file(self, scheduler_file):
```
mrocklin: Thoughts on calling this `write_scheduler_file`?  The prefix `to_` is often used in projects like pandas and xarray when we convert the data in the object to a new format.  That isn't quite the case here.

jhamman:
@mrocklin -- all green here.

mrocklin:
Thanks @jhamman !  Merged

Issue #1718, opened 2018-01-29
.pyz support in upload


bmaisonn:
Hi,
In the same way that we can upload .py, .egg and .zip files, i'd like to upload .pyz files.
For information .pyz is similar to .zip except that it's standard and can be easily distinguished from non-python zip files:
https://docs.python.org/3/library/zipapp.html
If accepted i can take care of the change.
Regards,
Bertrand

mrocklin:
I have no objection to extending the auto-import feature to all standard
python file formats.  I was unfamiliar with pyz.  Thanks for taking this on.

mrocklin:
@bmaisonn any progress here?

bmaisonn:
Hi!
Sorry it's implemented but i lacked of time to merge.
I'll do it quickly.
Bertrand

bmaisonn:
Created a pull request here:
https://github.com/dask/distributed/pull/1781

PR #1781, opened 2018-02-21
#1718 .pyz support in upload 

bmaisonn:
In the same way that distributed supports upload of .py and .egg files, it now supports to upload .pyz files.


distributed/tests/test_worker.py
```diff
@@ -213,6 +213,37 @@ def g(x):
...
+
+    aa = rpc(a.address)
+    bb = rpc(b.address)
+    with open(local_file, 'rb') as f:
+        payload = f.read()
+    yield [aa.upload_file(filename=pyzname, data=payload),
+           bb.upload_file(filename=pyzname, data=payload)]
+
+    assert os.path.exists(os.path.join(a.local_dir, pyzname))
+    assert os.path.exists(os.path.join(b.local_dir, pyzname))
```
mrocklin: Thoughts on using `c.upload_file` instead here?  This might be good both to simplify the testing process and also to ensure that everything works at the user-level API.

bmaisonn: You're right.
I've to admit that my test is inspired from the egg case. Should I change this one too ?

bmaisonn: actually `c.upload_file` fails because it can't find mytest.pyz
Maybe it's due to the way we create the test cluster, i'm going to investigate this

```

self = <Client: not connected>, filename = 'mytest.pyz', raise_on_error = True

    @gen.coroutine
    def _upload_file(self, filename, raise_on_error=True):
>       with open(filename, 'rb') as f:
E       FileNotFoundError: [Errno 2] No such file or directory: 'mytest.pyz'

distributed\client.py:2340: FileNotFoundError`
```

mrocklin: You could probably pass it the full path that you create in `local_file`

mrocklin: Changing the egg case would be welcome as well.

bmaisonn: Better late than never!
I made the the appropriate changes. It works fine fo the .pyz case.
However for .egg case, i had to rename the egg from 'mytestegg-1.0.0-py3.4.egg' to 'testegg-1.0.0-py3.4.egg'. In import_file, the module to load is determined using pkg_resource which returns 'mytestegg' instead of 'testegg'.
Not sure if the issue is due to the egg naming or in the way import_file determines the list of modules to load...

distributed/tests/test_worker.py
```diff
@@ -183,18 +183,13 @@ def g():
...
     assert not os.path.exists(os.path.join(b.local_dir, eggname))
     assert a.local_dir != b.local_dir
 
-    aa = rpc(a.address)
-    bb = rpc(b.address)
-    with open(local_file, 'rb') as f:
-        payload = f.read()
-    yield [aa.upload_file(filename=eggname, data=payload),
-           bb.upload_file(filename=eggname, data=payload)]
+    yield c.upload_file(filename=eggname)
```
mrocklin: From the test failures it looks like this should maybe be `local_file` instead?  It seems like it's having trouble finding the `eggname` file.  I'm curious, are you able to run tests locally and, if so, do these tests work locally for you?

mrocklin:
Thanks for following up with this!

mrocklin:
It also looks like there are a couple linting issues.  You can see these with flake8

```
pip install flake8
flake8 distributed
```

http://dask.pydata.org/en/latest/develop.html

mrocklin:
This looks good to me.  Thanks for following through @bmaisonn !

PR #1783, opened 2018-02-22
Add `futures_of` to API docs

jakirkham:
Includes `futures_of` in the API docs. Also makes sure `fire_and_forget` has its function documented since it is already listed.


mrocklin:
+1  

Thanks!

Issue #1760, opened 2018-02-14
FFT on modestly large array results in KilledWorker


wtbarnes:
# System Info
OS: GNU/Linux x86_64
Cores: 64
Memory: 252 GB
distributed 1.21.0
dask 0.17.0

# Minimum Example
``` python
# Import needed libraries
import numpy as np
import distributed
import dask.array
# Create local cluster
client = distributed.Client()
# Create Numpy array
shape = (500, 500, 1000)
big_array= np.random.rand(*shape)
# Create Dask array with appropriate chunk size
chunks = (int(shape[0]/20), int(shape[1]/20), shape[2])
big_dask_array = dask.array.from_array(big_array, chunks=chunks)
# Compute FFT
fft = dask.array.fft.rfft(big_dask_array, axis=2,  n=(2 * shape[2] + 1))
foo = fft.compute()
```

# Explanation and Traceback
When calculating a fast Fourier transform on a modestly large array (that still fits in memory), I get repeated error messages that look like,

```python
tornado.application - ERROR - Exception in callback <bound method Nanny.memory_monitor of <Nanny: tcp://127.0.0.1:32854, threads: 1>>
Traceback (most recent call last):
  File "/storage-home/w/wtb2/anaconda3/envs/synthesizar/lib/python3.6/site-packages/tornado/ioloop.py", line 1026, in _run
    return self.callback()
  File "/storage-home/w/wtb2/anaconda3/envs/synthesizar/lib/python3.6/site-packages/distributed/nanny.py", line 251, in memory_monitor
    self.process.process.terminate()
AttributeError: 'NoneType' object has no attribute 'terminate'
```
followed by the error (and traceback) seen below. Looking at the Bokeh dashboard for the local cluster, the jobs either hang or continually restart.

I'm a bit confused as to why the worker is being killed. The array sizes I'm using here easily fit into memory. Adjusting the chunk size either way does not seem to make a difference . The traceback does not provide much insight. Any idea what is going on here?

Reducing the array shape to `(475,475,1000)`, I don't seem to see this error. The tipping point seems to be around a `(495,495)`.

```python
---------------------------------------------------------------------------
KilledWorker                              Traceback (most recent call last)
<ipython-input-7-2e30612ecfe9> in <module>()
----> 1 foo = fft_1.compute()

~/anaconda3/envs/synthesizar/lib/python3.6/site-packages/dask/base.py in compute(self, **kwargs)
    141         dask.base.compute
    142         """
[...]

PR #1785, opened 2018-02-23
Reduces LocalCluster logging threshold from CRITICAL to WARN.

andyljones:
As per the discussion in #1760

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -34,7 +34,7 @@ class LocalCluster(object):
     scheduler_port: int
         Port of the scheduler.  8786 by default, use 0 to choose a random port
     silence_logs: logging level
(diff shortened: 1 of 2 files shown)


mrocklin:
One possible issue here is that we'll get exceptions from the workers, which might get a bit noisy.  

```python

In [2]: client = Client()

In [3]: future = client.submit(lambda x, y: x / y, 1, 0)

In [4]: distributed.worker - WARNING -  Compute Failed
Function:  <lambda>
args:      (1, 0)
kwargs:    {}
Exception: ZeroDivisionError('division by zero',)

distributed.scheduler - ERROR - error from worker tcp://127.0.0.1:33205: divisio
```

I suggest that we merge this in and see how people working on git-master like it (or not). 

Merging.

Issue #1770, opened 2018-02-16
distributed.nanny cannot start worker process


seanstorey:
**start dask-worker tcp://192.168.1.251:8786**

outputs

distributed.nanny - INFO -     Start nanny at: 'tcp://192.168.1.251:51329'
distributed.nanny - INFO - Failed to start worker process.  Restarting
distributed.nanny - INFO - Failed to start worker process.  Restarting
...

task manager shows the worker processes are being spawned but stop after a couple of seconds 

**start dask-worker tcp://192.168.1.251:8786 --no-nanny** 

runs correctly connecting to the distributed.scheduler. 

Using dask==0.17.0, distributed==1.21.0, tornado==4.5.3 on Windows 10 Professional with 

logging:
  distributed: info
  distributed.client: warning
  distributed.nanny: debug
  distributed.nanny: debug

Is it possible to redirect the logs to file, particularly for the spawned worker processes ?

mrocklin:
Logs are sent to stderr, you can redirect them as follows:

    dask-worker ... 2> log.txt

mrocklin:
I don't generally have any helpful information on what might be going on.  You might also consider trying one version back to see if this was newly introduced.

nav1n:
I have the same issue, but worked around it by extending the timeout found in nanny.py within the start method of the WorkerProcess class from 5 to 10 seconds:
```
                yield self.process.start()
                if self.status == 'starting':
                    yield gen.with_timeout(timedelta(seconds=10),
                                           self._wait_until_running())
```
I see this log message when the worker starts up:
`distributed.core - WARNING - Event loop was unresponsive in Nanny for 5.54s.  This is often caused by long-running GIL-holding functions or moving large chunks of data. This can cause timeouts and instability.`

mrocklin:
Hrm, that's interesting feedback.  Ideally there wouldn't be any hard-coded timeouts like that (I see that there are a few more floating around).  

Presumably you're on a system where startup times are longer, both due to Windows' spawn and possibly also due to things like network filesystems, etc..  This reliance on a fixed timeout seems like a bug to me.

seanstorey:
Thanks nav1n, I had found this and tried seconds=10 with out success though have also tried it at seconds=20 again without luck. I'm following the code / adding logging but as it works with -no-nanny it hasn't been on the top of my to-do list :-) though I will investigate further

jakirkham:
Does 1.21.1 work for you @seanstorey?

[...]

PR #1723, opened 2018-01-30
Clean up IOLoop handling in PeriodicCallbacks

mrocklin:
In Tornado 4 the IOLoop is specified at creation time.
This gave PeriodicCallbacks the wrong IOLoop when running in the console.

PR #1651, opened 2017-12-21
Add user-defined priorities

mrocklin:
Fixes https://github.com/dask/dask/issues/2860

Or at least it would if more work were put in here.  This is a minimal example.

PR #1793, opened 2018-02-28
Avoid restarting workers when nanny waits

mrocklin:
Fixes https://github.com/dask/distributed/issues/1770

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -461,7 +461,6 @@ class WorkerProcess(object):
                 yield gen.sleep(delay)
                 continue
 
-
     @gen.coroutine
     def _wait_until_connected(self):
         delay = 0.05
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -455,7 +455,7 @@ class WorkerProcess(object):
                 return
             try:
                 msg = self.init_result_q.get_nowait()
-                assert msg == 'started'
+                assert msg == 'started', msg
                 return
             except Empty:
                 yield gen.sleep(delay)
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -333,7 +333,7 @@ class WorkerProcess(object):
             return
 
         while True:
-            # FIXME: this sometimes stalls in _wait_until_running
+            # FIXME: this sometimes stalls in _wait_until_connected
             # our temporary solution is to retry a few times if the process
             # doesn't start up in five seconds
             self.init_result_q = mp_context.Queue()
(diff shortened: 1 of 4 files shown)


distributed/nanny.py
```diff
@@ -356,14 +356,17 @@ def start(self):
                 yield self.process.start()
                 if self.status == 'starting':
                     yield gen.with_timeout(timedelta(seconds=5),
```
jakirkham: Could we add this timeout to the config file?

distributed/nanny.py
```diff
@@ -356,14 +356,17 @@ def start(self):
...
-                                           self._wait_until_running())
+                                           self._wait_until_started())
             except gen.TimeoutError:
                 logger.info("Failed to start worker process.  Restarting")
                 yield gen.with_timeout(timedelta(seconds=1),
```
jakirkham: Not sure if it is needed, but it might be nice if this was configurable too.

mrocklin: Yeah, ideally all of the timeouts everywhere would be configurable.  I'm not yet sure how best to do this.

distributed/nanny.py
jakirkham: Pointing this one out as people seem to already be hacking the code to change this anyways as illustrated by [this comment]( https://github.com/dask/distributed/issues/1770#issuecomment-368921388 ) and [this one]( https://github.com/dask/distributed/issues/1770#issuecomment-369013195 ). Even gave it a try myself while doing some debugging earlier. Of course naming is hard, maybe `worker_startup_wait_time`?

mrocklin:
This timeout shouldn't be necessary.  There are other things going on that
are necessitating the hacking around.

distributed/nanny.py
mrocklin: This is no longer the same timeout.  This is the time for the process to start and send a message back, not the time it takes to connect (which is what is was for the last version)

jakirkham:
Hmm...so what is the motivation for the timeout?

mrocklin:
Very occasionally in our tests worker processes don't start up.  I don't know why.  I spent a couple days looking at this a few weeks ago and could come up with anything.

jakirkham:
Is there an issue or PR that I can read about this?

mrocklin:
Unfortunately it appears that I didn't take very good notes.  The change
was made here:

https://github.com/dask/distributed/pull/1723

And something similar around the same time was discussed here:

https://github.com/dask/distributed/pull/1651#issuecomment-361468976

Although thinking more about that now I think that that might be a separate
issue, which is still live if you rename

distributed/tests/test_failed_workers.py
to distributed/tests/test_worker_failed.py :)

PR #1794, opened 2018-03-01
msgpack-python --> msgpack (Name change on pypi) (may fix #1913)

### LINKED ISSUE #1913: problem importing msgack-python
When I try to run dask-mpi I get the following error:

```
arun@apersaud:~>mpirun n0-4 dask-mpi --scheduler-file /home/$USER/scheduler.json
mpirun: cannot start dask-mpi on n1: invalid node
  File "/usr/bin/dask-mpi", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3088, in <module>
    @_call_aside
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3072, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3101, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 574, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 892, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 778, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'msgpack-python' distribution was not found and is required by distributed
```
However, I have msgpack-python installed.

```
arun@apersaud:~>python3
Python 3.6.4 (default, Jan 03 2018, 13:52:55) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import msgpack
>>> msgpack.version
(0, 5, 6) 
```
I'm running linux (opensuse tumbleweed) with
python3-distributed version     : 1.21.6
and
python3-dask version     : 0.17.2

Any idea what is going wrong here?

rbubley:


mrocklin:
Something is up with tests here.  I'm not sure what.  Have you verified that this works locally?

rbubley:
I checked that

```
pip uninstall msgpack-python
pip install msgpack
```

worked.

Maybe it’s something more subtle.

mrocklin:
I've tried restarting the test in travis-ci to see if that helps.  Given that the same issue also arose on appveyor my guess is that it won't though. 

The next thing I recommend would be to create the testing environment locally.  You can do this by going through the continuous_integration/travis/install.sh script.  This is a little bit laborious, but a useful way to get fairly exactly to what is happening on travis-ci.

rbubley:
In the Appveypr logs it is still trying to load msgpack-python, will look for where this is referred to.

rbubley:
That will probably be it. The PR only changed `requirements`; the CI set-ups explicitly contain the pip package names too. I’ll amend the PR to change the CI set-ups to match.

mrocklin:
Hey cool, you got linux tests to pass :)

rbubley:
Looking at the AppVeyor logs, it doesn't obviously look like this failure is connected to this PR. Any ideas?

mrocklin:
*Maybe*.  The CLI tests (which are what is failing) are generally bad about giving us logs (which is why we use them rarely) and about hanging.  I've restarted the Appveyor test to see if it repeats.  It isn't showing up on other PRs going on right now though, so it may be a genuine issue.

rbubley:
I have a suspicion that it stems from the AppVeyor build trying to install both `msgpack` and `msgpack-python`. This will be easier to resolve once this name change has been sorted out at [conda-forge](https://github.com/conda-forge/msgpack-python-feedstock/issues/5), so this should be parked until then.

mrocklin:
cc @jakirkham , any thoughts?

jakirkham:
As to the failures? Not ATM.

As to msgpack in conda-forge, don't think conda-forge will rename it (just closed that issue) as Anaconda is not planning to rename it (nor do they want to). Wouldn't want to deviate from Anaconda unless there was a pretty compelling reason. PyPI name change for a package available in multiple languages isn't one for me at least. Happy to change the URL to get the package from PyPI under its new name though (with a version bump). PR for this would be welcome.

mrocklin:
Checking in on this.  Is there anything that we should be doing here?

rbubley:
For the time being, it's just a housekeeping issue - nothing is yet broken from this.
I have suggested a fix at conda forge <https://github.com/conda-forge/msgpack-python-feedstock/issues/5>, which *may* resolve this issue here – but not being either a conda or windows user it's hard for me to test.

mrocklin:
It's pretty easy to [download miniconda](https://conda.io/miniconda.html) test things, and then blow it away.  I do this for testing things like 32 bit architectures regularly.  One of Conda's main selling points is that it doesn't touch your system Python at all, so it's safe to install ephemerally.

rbubley:
Are you sure, @jakirkham ? As far as I can see conda-forge[ is still pointing to](https://github.com/conda-forge/msgpack-python-feedstock/blob/master/recipe/meta.yaml) `msgpack-python` on pypi not to `msgpack`; it's just the version that has been bumped.

jakirkham:
The two versions are the same AFAICT. Are you seeing differences between those two Python packages?

rbubley:
I've not looked. But the conda-forge recipe is definitively pointing at the deprecated target, so it will need to be fixed eventually.

TBH, I still have no idea why this PR is failing on appveyor – there is nothing in the logs that I can see relating to the msgpack name change: fixing the conda-forge problem was simply a guess. I'll try merging from current master into this branch and see if it still errors.

jakirkham:
Should add that I'm not the maintainer of that feedstock, but have been helping out a bit. So would ping the maintainer in that issue to get his feedback.

lesteve:
Sorry I did not see this PR when I opened #1927. For the record I was getting a similar error on AppVeyor and I don't have a good explanation either ...

lesteve:
@rbubley if you could add "fix #1913" (I am not 100% sure about that it does fix it but reasonably confident) to your PR description, this would be great. This way #1913 will be closed when this PR is merged.

lesteve:
It's hard to exactly be sure just from the AppVeyor log, but I believe this is very similar to what is happening in #1913: you can import `msgpack` but as far as `pkg_resources` is concerned it is called `mspack-python` (conda package name). When you try to execute scripts (e.g. `test_dask_scheduler.py::test_defaults`, which fails, does execute `dask-scheduler`), `pkg_resources` can not find `msgpack` and raises a `DistributionNotFoundError`.

I would think that a diff like this would fix the error (does `pip install msgpack` in AppVeyor):
```diff
--- a/continuous_integration/setup_conda_environment.cmd
+++ b/continuous_integration/setup_conda_environment.cmd
@@ -49,6 +49,7 @@ call activate %CONDA_ENV%
 %PIP_INSTALL% git+https://github.com/dask/zict --upgrade
 
 %PIP_INSTALL% pytest-repeat pytest-timeout pytest-faulthandler sortedcollections
+%PIP_INSTALL% msgpack
 
 @rem Display final environment (for reproducing)
 %CONDA% list
```

The situation is a bit dodgy though and I am not sure about the best approach. Here are the main caveats I can see:
* if you are installing distributed from source (wheels are fine and conda packages too), you need to do `pip install msgpack`. It will be done automatically (`msgpack` is in `install_requires` with these PR changes) unless you use `pip install --no-deps`.
* I am not sure about the other potential implications of mixing two packages (with potentially one coming from conda and the other one from pip) with different distribution names under the same site-packages folder.

jakirkham:
That's related to the how `msgpack`/`msgpack-python` does naming. The `conda` package is merely using the `msgpack-python` flavor currently. This could be changed to use the `msgpack` flavor instead if someone would like to submit a PR.

ref: https://github.com/msgpack/msgpack-python/blob/0.5.6/setup.py#L108-L111
ref: https://github.com/conda-forge/msgpack-python-feedstock/blob/7c7c8c9764d559e3e7c55bb1783912c39ba4a543/recipe/meta.yaml#L13-L14

lesteve:
Doing `pip install msgpack` rather than `conda install msgpack-python` on AppVeyor seems to fix the issue. See this [AppVeyror build](https://ci.appveyor.com/project/lesteve/distributed/build/1.0.3) on my fork.

I agree that changing the `name` field from `msgpack-python` to `msgpack` in the conda recipe will be needed too.

PR #1797, opened 2018-03-02
Adaptive min max

mrocklin:
Replaces https://github.com/dask/distributed/pull/1618  cc @jacobtomlinson @jhamman 

This adds minimum and maximum keywords to the `Adaptive` class to set limits on available workers.


mrocklin:
There are still some issues with this.  Maximum isn't implemented.

jhamman:
Thanks @mrocklin - I'm looking forward to using this!

PR #1804, opened 2018-03-03
Use string timedeltas everywhere

mrocklin:
Previously when taking time-delta inputs we usually took numbers that
represented either seconds or milliseconds, depending on if we were
using system python or tornado/bokeh commands. 

```python
>>> Adaptive(..., interval=1000, startup_cost=1)
```

Now we normalize everything to strings or timedelta objects and parse them.

```python
>>> Adaptive(..., interval='1000 ms', startup_cost='1s')
```

This is clearer and more explicit.

We maintain backwards compatibility by supporting a default= keyword
to parse_timedelta that lets us choose a default unit.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -8,6 +8,7 @@ import functools
 import json
 import logging
 import multiprocessing
+from numbers import Number
 import operator
 import os
 import re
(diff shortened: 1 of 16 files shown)


distributed/utils.py
```diff
@@ -1130,6 +1130,65 @@ def parse_bytes(s):
...
+    --------
+
+    >>> parse_timedelta('3s')
+    3
+    >>> parse_timedelta('3.5 seconds')
+    3.5
+    >>> parse_timedelta('300ms')
+    0.3
+    """
+    s = s.replace(' ', '')
```
jreback: you know about pd.to_timedelta right?

mrocklin:
I do and would love to use it, but we don't have a strict dependency on
Pandas

mrocklin:
cc @jacobtomlinson and @jhamman for feedback from folks doing this operationally.

jhamman:
@mrocklin - I don't have any specific comments on the implementation but I am appreciative of the clarity/consistency this brings across the code base, particularly in the `Adaptive` startup cost.

jacobtomlinson:
Yeah looks good to me. cc @niallrobinson

PR #1807, opened 2018-03-05
Scale adaptive based on total_occupancy

mrocklin:


rbubley:
@mrocklin, did you mean to have
```
instances = max(1,
                        len(self.scheduler.workers) * self.scale_factor,
                        target - len(self.scheduler.workers))
```
rather than
```
instances = max(1,
                        len(self.scheduler.workers) * self.scale_factor,
                        target)
```
? 

It would seem to me that the `target` is the number of instances needed to finish in `task_duration` – or are you trying to force some sort of interaction with `self.scale_factor`?

mrocklin:
I want to scale up either by the scale factor, or to the target, whichever is larger.

mrocklin:
This is all pretty arbitrary though, I suspect that we'll change this around a lot in the coming months.  Did you try this out and experience bad behavior?

rbubley:
If you want to scale to the larger of the two, then I think you shouldn’t subtract away from `target`, as this should return the desired number of instances not the desired increase.

I haven’t had problems with this as I had previously sub-classed it with similar functionality; rather I was reviewing it versus my sub-class.

mrocklin:
Indeed.  I think you're right, we should just be using target there.

PR #1812, opened 2018-03-06
option to choose remote Python

svenkreiss:
The default continuous to be `sys.executable`. For remote systems with a different file layout, this can now be set to `python` or `python3` or `some/other/path/python3`.


mrocklin:
This seems sensible to me.  Thanks!

PR #1813, opened 2018-03-06
Add cluster superclass

mrocklin:
This defines shared functionality that can be used across clusters.  Today this is mostly adaptivity and ipython widgets.  This should be useful for a variety of cluster objects:

image


mrocklin:
@jhamman I was trying to play with this with PBSCluster on Geyser but had difficulty logging in.  If you're around and can either play with this or point me to the right instructions to get an interactive notebook please let me know.

PR #1815, opened 2018-03-06
Fix bug in recreate_error when dependencies are dropped

mrocklin:


distributed/recreate_exceptions.py
```diff
@@ -91,7 +91,7 @@ def _get_futures_error(self, future):
             raise gen.Return((function, args, kwargs, deps))
         else:
             function, args, kwargs = _deserialize(task=task)
-            raise gen.Return((function, args, kwargs, []))
+            raise gen.Return((function, args, kwargs, deps))
```
mrocklin: @martindurant any objection to this change?

martindurant: no objection

mrocklin: Thanks!

PR #1817, opened 2018-03-07
Add percentage to profile plot

bnaul:
Percentage of total runtime seems like a very useful metric (I always end up computing it by hand) so why not add it to the tooltip?

image


mrocklin:
Looks great.  Thanks @bnaul !

mrocklin:
Yup. I just found this useful in practice.  That didn't take long :)

PR #1828, opened 2018-03-10
Add fifo timeout to control priority generation 

mrocklin:
Previously every call to update_graph would decrement the priority of
those tasks.  This resulted in a first-in-first-out behavior.
This was good both for multi-client workloads, and for consistent
ordering.

However, this got in the way of common workflows where people would
call persist rapidly in sucession:

    x = x.persist()
    y = y.persist()  # <-- these are all lower priority

Now we only decrement priority when enough time has passed.  
This time is different for realtime apis (map/submit) and dask collections
(compute/persist)

-  submit/map: 100ms
-  compute/persist: 60s

These are controllable with the `fifo_timeout=` keyword

Additionally we re-institute a last-in-first-out policy on workers
that serves only as a tie breaker, such as when using client.submit.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1114,6 +1114,8 @@ class Client(Node):
         priority: Number
             Optional prioritization of task.  Zero is default.
             Higher priorities take precedence
+        fifo_timeout: str timedelta (default '100ms')
+            Allowed amount of time between calls to consider the same priority
 
         Examples
         --------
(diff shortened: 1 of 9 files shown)


mrocklin:
Fixes https://github.com/dask/dask/issues/3268

jakirkham:
Is it possible to set things like `priority` globally? Related are we able to set `priority` for a context?

mrocklin:
Currently no.  I can imagine several reasons for wanting to do this though.

Issue #1837, opened 2018-03-14
No response when connecting client to dask-scheduler


rudvlf0413:
I ran dask scheduler as dask-scheduler --scheduler-file /users/ddd/workspace/fff/tmp/scheduler.json and scheduler.json was created file in that directory

and here are contents of scheduler.json:
{
  "type": "Scheduler",
  "id": "Scheduler-695028c5-521c-455b-a4f7-a84dcbc35c8e",
  "address": "tcp://***:8786",
  "services": {
    "bokeh": 8787
  },
  ...
}

also, dask-worker --scheduler-file /users/ddd/workspace/fff/tmp/scheduler.json

I try to connect client to dask scheduler in network file system using below code:
client = Client(scheduler_file='/users/ddd/workspace/fff/tmp/scheduler.json')

but it doesn't work (no any response or it may be in infinite loop).

I think client can't connect to dask-scheduler. However, I don't know how to solve this problem...

How can I solve this problem? Is there anyone who had similar problem?

mrocklin:
To help diagnose your problem I recommend including logs from your scheduler, workers, and client.

rudvlf0413:
@mrocklin 

I capture logs of scheduler and worker as below.
<img width="1195" alt="2018-03-15 1 30 58" src="https://user-images.githubusercontent.com/6302455/37416278-b57173d6-27f0-11e8-9702-da781f16d745.png">
<img width="1264" alt="2018-03-15 1 31 21" src="https://user-images.githubusercontent.com/6302455/37416282-b79dc7cc-27f0-11e8-9791-56208b3c73a3.png">

and scheduler.json
<img width="545" alt="2018-03-15 1 31 34" src="https://user-images.githubusercontent.com/6302455/37416310-ca3e3c22-27f0-11e8-8cc9-9ee92d46fd47.png">

and then I got these errors
<img width="1153" alt="2018-03-15 1 29 56" src="https://user-images.githubusercontent.com/6302455/37416325-d21895f0-27f0-11e8-8b8d-6744504d6134.png">

Do you know why I got these errors? :(

rudvlf0413:
I'm currently using
dask==0.17.1
distributed==1.21.3
tornado==5.0
python 3.6.4

mrocklin:
No, I don't know.  The two most common causes of this kind of issue are:

   1. You have different versions of Python/Dask/... running on the
   different machines (this is worth double-checking even if you are confident)
[...]

PR #1842, opened 2018-03-19
Highlight that asynchronous=True should only be used with async/await…

mrocklin:
… functions

Fixes https://github.com/dask/distributed/issues/1837

### DIFF
--- a/docs/source/asynchronous.rst
+++ b/docs/source/asynchronous.rst
@@ -1,20 +1,21 @@
 Asynchronous Operation
 ======================
 
-Dask.distributed can operate as a fully asynchronous framework and so
-interoperate with other highly concurrent applications.  Internally Dask is
-built on top of Tornado coroutines but also has a compatibility layer for
-asyncio (see below).
+Dask can run fully asynchronously and so interoperate with other highly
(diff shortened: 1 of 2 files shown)


docs/source/asynchronous.rst
```diff
@@ -1,10 +1,9 @@
 Asynchronous Operation
 ======================
 
-Dask.distributed can operate as a fully asynchronous framework and so
-interoperate with other highly concurrent applications.  Internally Dask is
-built on top of Tornado coroutines but also has a compatibility layer for
-asyncio (see below).
+Dask run fully asynchronously and so interoperate with other highly concurrent
+applications.  Internally Dask is built on top of Tornado coroutines but also
+has a compatibility layer for asyncio (see below).
```
martindurant: Should mention the transparent creation of threads, if not using async, somewhere?

mrocklin: No objection to mentioning thread creation within the docs.  I don't think that here is the place though (this is about running within async-await functions).  This hasn't come up in a user issue that I'm aware of, so it's somewhat far down my priority list at the moment.

PR #1848, opened 2018-03-22
[WIP] - Restart if nanny's checks fail

mrocklin:
This is an attempt to track down the following error:

```
  File "/home/b.weinstein/miniconda3/envs/pangeo/lib/python3.6/site-packages/distributed/nanny.py", line 471, in _wait_until_started
    assert msg == 'started', msg
AssertionError: {'address': 'tcp://172.16.194.178:37068', 'dir': '/home/b.weinstein/dask-worker-space/worker-5ipv17oi'}
```

PR #1849, opened 2018-03-22
Publish temp_default_client

jakirkham:
Adds the `temp_default_client` function as part of the public API. Can be useful in cases where a user wants to guarantee what `Client` is used for submission, but does not want to alter what the default `Client` may be set as elsewhere.

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -4,7 +4,7 @@ from .config import config
 from .core import connect, rpc
 from .deploy import LocalCluster, Adaptive
 from .diagnostics import progress
-from .client import (Client, Executor, CompatibleExecutor,
+from .client import (Client, Executor, CompatibleExecutor, temp_default_client,
                      wait, as_completed, default_client, fire_and_forget,
                      Future, futures_of)
 from .lock import Lock
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -46,6 +46,7 @@ API
 .. currentmodule:: distributed
 
 .. autosummary::
+   temp_default_client
    worker_client
    get_worker
    get_client
@@ -156,6 +157,7 @@ Other
 
 .. currentmodule:: distributed
 
+.. autofunction:: distributed.temp_default_client
 .. autofunction:: distributed.worker_client
 .. autofunction:: distributed.get_worker
 .. autofunction:: distributed.get_client


mrocklin:
Should we rename this to `temporary_default_client`?  That might be easier to understand.  This is so infrequently used that abbreviating the name may not have much value.

jakirkham:
IDK. `temp` is a pretty common abbreviation (e.g. [`tempfile`]( https://docs.python.org/3/library/tempfile.html )). That said, no strong opinions from me.

jakirkham:
Hmm...AppVeyor tests have some sort of time out issue cropping up. Don't think it is related... 🤔

mrocklin:
Yeah, tests are in a sorry state at the moment.  There are a few
intermittent failures, I think due to something related to
https://github.com/dask/distributed/pull/1848 , but I'm not sure.  I've
been restarting tests recently when they fail.

jakirkham:
No worries. Just wanted to make sure we weren't stumbling into something here. Restarted AppVeyor to see if it passes this time.

mrocklin:
OK, next question: do we need this in the top-level API?  Or is it advanced
enough that it should stay within distributed.client ?

jakirkham:
Am ok with either.

The main thing that motivated me to look for this is some functions accept an optional `client` argument and others just grab the `default_client`. Having a standard way of saying, "use this as the `default_client` for this block of code", should effectively address both cases. Right now `temp_default_client` seems to do this well. For that reason it seems nice to have this available in the same space as `default_client`.

Long term we might want to standardize functions a bit by having them always optionally taking a `client` and falling back to `default_client` if not otherwise set (possibly with some special cases like `futures_of`). Though saving some boiler plate and/or handling nested contexts on this front with `temp_default_client` would expect remains useful.

On a totally different note, we might want to override the default `get` in this context as well. Though I may be overthinking things.

jakirkham:
Thoughts? Happy to tweak as needed.

jakirkham:
Could you please let me know what you think here, @mrocklin? Am open to changing the strategy as long as this is part of the public API in the end.

mrocklin:
Sorry for the prolonged silence.  Taking another look now.

mrocklin:
I still have a preference to keep this out of the top-level API, mostly because I percieve a cost to flooding that namespace and this function is so infrequently used.  I think that it is best to keep it in distributed.client until we see more use of it (a quick search on github shows no uses other than in nanshe and copies of this codebase)

> Right now temp_default_client seems to do this well. For that reason it seems nice to have this available in the same space as default_client.

Yeah, this does make sense.  Part of me wishes that I hadn't put default_client in the top-level namespace now :).  I still think that we should wait and that people should import from `distributed.client` today.

There is still the question of "how do we signal that this is dependable API?" or, more deeply, "do we want to signal that this is dependable API?".  I don't see any reason why this function would change or go away in the future, but we should also recognize that this project is still at a stage where it does experience internal churn. It may be that we just don't promise long-term support on functions like these.

So, for the moment my inclination is not to merge this, and to ask you to depend on internal API, shifting when it shifts (which will probably be never) rather than make the promise that it will not shift except with a deprecation cycle.

mrocklin:
Push-back on that idea is welcome.

jakirkham:
I'd really like to see some option to solve this problem become part of the public API. The cost incurred by not having a function like this join the public API is a bit high especially for me personally. If there is another one that you would prefer to make part of the public API, that would be good to discuss in issue ( https://github.com/dask/dask/issues/3708 ).

mrocklin:
Hey, any thoughts on using dask.config for this?

```python
with dask.config.set({'distributed.default-client': my_client}):
    ...
```

`default_client` would also check this first, before going to the collection of weakrefs.

I like that this reuses the existing API to temporarily set things.

jakirkham:
Think PR ( https://github.com/dask/dask/pull/4062 ) addressed this. Closing.

PR #1850, opened 2018-03-22
Use ALL_COMPLETED variable

jakirkham:
Not sure if there is some obvious reason not to do this...


jakirkham:
> There are obvious benefits to using the variable as well, like forced consistency, avoiding misspellings, etc..

That's what I was thinking about. ;)

<br>

Ok, feel free to merge or close as you see fit.

PR #1852, opened 2018-03-22
Remove started timeout in nanny

mrocklin:
This was introduced to avoid testing failures where processes sometimes
just don't start.  Unfortunately what we thought was a generous timeout
here is too long for some HPC systems.

Lets try removing this safety check and see if tests continue to fail.


jakirkham:
Is this still needed?

mrocklin:
Maybe.  If I can find out why this was needed in the first place then it
would be useful to remove this feature altogether.  It's an odd situation.

mrocklin:
OK, this is generally responding to travis-ci much better than current master.  We've also had at least two external reports that it makes things somewhat nicer.  Merging.

PR #1855, opened 2018-03-23
Add security support to LocalCluster.

mariusvniekerk:
This makes it easier for users to simulate a tls enabled cluster
locally.

### DIFF
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -464,5 +464,6 @@ def test_local_tls(loop):
              exception_class=RuntimeError,
              )
 
+
 if sys.version_info >= (3, 5):
     from distributed.deploy.tests.py3_test_deploy import *  # noqa F401
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -17,7 +17,9 @@ from distributed import Client, Worker, Nanny
 from distributed.deploy.local import LocalCluster
 from distributed.metrics import time
 from distributed.utils_test import (inc, gen_test, slowinc,
+                                    assert_cannot_connect,
                                     assert_can_connect_locally_4,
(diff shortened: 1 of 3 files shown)


distributed/deploy/tests/test_local.py
```diff
@@ -444,5 +444,17 @@ def test_adapt_then_manual(loop):
...
+def test_local_tls(loop):
+    from distributed.utils_test import tls_only_security
+    security = tls_only_security()
+    with LocalCluster(scheduler_port=0, silence_logs=False, security=security,
+                      diagnostics_port=None, ip='tls://0.0.0.0', loop=loop) as c:
+        sync(loop, assert_can_connect_from_everywhere_4_6, c.scheduler.port,
+             protocol='tls',
+             connection_args=security.get_connection_args('client'),
+             timeout=1,
+             )
```
mrocklin: Thoughts on adding some trivial assertions here like `assert c.scheduler.address.startswith('tls://')` or is the idea that the `assert_can_connect_from_everywhere` function should handle that for us?

mrocklin:
In general this seems great to me.  I may not be well qualified to give technical review here though.

distributed/deploy/tests/test_local.py
mariusvniekerk: that was mostly my thought.  The other one that i should probably add is asserting that non-tls communication does not work,  and i guess more in the way of docs

mrocklin:
This is causing a couple failures in tests.  They appear to be genuine.

mariusvniekerk:
@mrocklin This seems to be passed now

mariusvniekerk:
ah whoops, i was testing with docker :P

mrocklin:
This is in.  My apologies for the late merge @mariusvniekerk , I was away last week. 

I also noticed that this is your first official code contribution to dask.distributed.  Welcome!

PR #1865, opened 2018-03-27
Extend nanny started timeout to 30s

mrocklin:
Sometimes worker processes just don't start.  We don't know why.
Previously we set up a timeout of five seconds such that if a worker
process does not at least start in that time we restart the process
and try again.  This timeout is slow on some systems, notably HPC
clusters with slow network file systems, and also windows machines when
they try to start many workers simultaneously.

This commit extends that timeout to thirty seconds and makes it
configurable with the ``nanny-started-timeout`` config option.

Additionally, sometimes wires seem to get crossed and we get messages
from previous attempts.  We now restart cleanly in these cases.

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -21,7 +21,7 @@ from .process import AsyncProcess
 from .proctitle import enable_proctitle_on_children
 from .security import Security
 from .utils import (get_ip, mp_context, silence_logging, json_load_robust,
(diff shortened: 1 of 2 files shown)


jakirkham:
Thanks for doing this @mrocklin.

Agree with the change as I've also ran into this issue on HPC sometimes. Plus having it be configurable seems like a good idea.

PR #1882, opened 2018-04-01
Adds has_dataset method to Client

jakirkham:
Adds the `has_dataset` method to `Client`.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1826,6 +1826,18 @@ class Client(Node):
         """
         return self.sync(self.scheduler.publish_list, **kwargs)
 
+    def has_dataset(self, name, **kwargs):
+        """
+        Checks for named dataset available on the scheduler
+
+        See Also
+        --------
+        Client.publish_dataset
+        Client.get_dataset
+        Client.list_datasets
+        """
+        return self.sync(self.scheduler.publish_has, name=name, **kwargs)
+
     @gen.coroutine
     def _get_dataset(self, name):
         out = yield self.scheduler.publish_get(name=name, client=self.id)
--- a/distributed/publish.py
+++ b/distributed/publish.py
@@ -75,7 +75,7 @@ class Datasets(MutableMapping):
         self.__client.unpublish_dataset(key)
 
     def __contains__(self, key):
-        return key in self.__client.list_datasets()
+        return self.__client.has_dataset(key)
 
     def __iter__(self):
         for key in self.__client.list_datasets():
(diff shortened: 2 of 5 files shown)


distributed/client.py
```diff
@@ -1826,6 +1826,18 @@ def list_datasets(self, **kwargs):
...
+        """
+        Checks for named dataset available on the scheduler
+
+        See Also
+        --------
+        Client.publish_dataset
+        Client.get_dataset
+        Client.list_datasets
+        """
+        return name in self.list_datasets()
```
mrocklin: Perhaps we should just leave this to the user to do on their own?

jakirkham: Adjusted this to push the check into `PublishExtension`.

Though would be better great if `get_dataset` behaved a bit more like `get` for a `Mapping` instance by taking a `default` value. ATM checking for a dataset that doesn't exist even if one uses a `try...except` triggers a lengthy amount of logging about the `name` in question not being present.

jakirkham:
Any thoughts on the updated version?

mrocklin:
Everything here things consistent to me.  I think my main question here is where do we draw the line between functions that we implement and add to Client/Scheduler and functions that we leave to the user.  There is some cost to keeping around several methods on the Client.  I would not want to implement all of the MutableMapping API on the Client for published datasets, for example.

As an alternative, I wonder if we might separate out the publish API to a separate `MutableMapping` interface.  Someone suggested this previously (maybe you?).  I would be much less conservative here if this API was not strongly attached to the Client.

martindurant:
Don't we already have this as `'mydataname' in client.datasets`? I agree that `client.datasets.has` would be less intrusive than `client.has_dataset`. In fact, I suspect that the publish functions, as opposed to the `dataset` attribute access, have been kept around only for backward compatibility.

mrocklin:
Ah ha, yes, that thing.

jakirkham:
An alternative would be PR ( https://github.com/dask/distributed/pull/1887 ), which might be better for my use case. However it seems to not quite exhibit the behavior desired yet. Also that PR has a better description of the motivation.

jakirkham:
True. Was looking at the `MutableMapping` option as well. Does look nicer for my use case. Only realized that was an option today.

Think the main issue left is the `get_dataset` trace. PR ( https://github.com/dask/distributed/pull/1888 ) should improve on that (if we are ok with that).

Given all of this, will close this PR out. Thanks everyone. :)

PR #1885, opened 2018-04-01
Fix numeric environment variable configuration

### LINKED ISSUE #1884: Numeric environment variable configuration is broken
I recently found that a pipeline was throwing this error:
```
  File "/usr/local/lib/python3.6/site-packages/distributed/utils.py", line 623, in log_errors
    yield
  File "/usr/local/lib/python3.6/site-packages/distributed/scheduler.py", line 1591, in remove_worker
    if ts.suspicious > self.allowed_failures:
TypeError: '>' not supported between instances of 'int' and 'str'
```
The environment variable `DASK_ALLOWED_FAILURES=20` had been set. In a new terminal I ran the following:
```pya
>>> LocalCluster().scheduler.allowed_failures
'20'
```
There appears to be nothing in the code which ensures that the type of a configuration variable conforms to the expected type.

I'd be happy to submit a PR to fix this but I'm not really sure what form it should take. Some ideas:
1. Remove environment variable configuration completely (it doesn't seem to be documented?)
2. Use `pyyaml` to interpret environment variables the same way yaml interprets them
    - (introduces are hard dependency on `pyyaml`)
3. Approximate the above by just automatically casting all env var config variables to ints where possible
4. Make a `get_config` which casts its return value to be the same type as the provided default; replace `config.get` everywhere. e.g.
```py
def get_config(key, default=None):
    if default is not None:
        return type(default)(config.get(key, default=default))
    return config.get(key)
```
5. Change the `config` dict to be an instance of a subclass of dict which has `get()` overridden to behave as suggested above
6. Change `config.py` to use the example `yaml` file to correctly cast all config variables
7. Cast variables correctly at the point of `get` everywhere
    - e.g. `ALLOWED_FAILURES = int(config.get('allowed-failures', 3))`

I think I'd probably go with 3, which seems like the lowest impact solution.

Of course, the workaround for now is to just use a yaml configuration file.

Spacerat:
This fixes #1884 by trying to parse environment variables as integers, floats or booleans if possible. I allow booleans to be represented with any case to stay consistent with PyYAML.

I've also added a test of `load_env_vars()`, and a helper to temporarily set environment variables.

(A small note: this will actually make environment variables more compatible with YAML than PyYAML is, since `yaml.load('1e+10')` incorrectly returns the string '1e+10'.)

### DIFF
--- a/distributed/config.py
+++ b/distributed/config.py
@@ -67,7 +67,24 @@ def load_env_vars(config):
     for name, value in os.environ.items():
         if name.startswith('DASK_'):
             varname = name[5:].lower().replace('_', '-')
-            config[varname] = value
+            config[varname] = _parse_env_value(value)
+
+
+def _parse_env_value(value):
+    """ Convert a string to an integer, float or boolean (in that order) if possible. """
+    bools = {
+        'true': True,
+        'false': False
+    }
+    try:
+        return int(value)
+    except ValueError:
+        pass
+    try:
+        return float(value)
+    except ValueError:
+        pass
+    return bools.get(value.lower(), value)
 
 
 def _initialize_logging_old_style(config):
--- a/distributed/tests/test_config.py
+++ b/distributed/tests/test_config.py
@@ -9,8 +9,8 @@ import os
 import pytest
 
 from distributed.utils_test import (captured_handler, captured_logger,
-                                    new_config, new_config_file)
-from distributed.config import initialize_logging, set_config, config
+                                    new_config, new_config_file, new_environment)
+from distributed.config import initialize_logging, set_config, config, load_env_vars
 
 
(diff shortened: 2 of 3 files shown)


Spacerat:
Ah, thanks for the heads up! I amended the commit, so I'm passing my own test & flake8. However an unrelated test seems to have failed. Do you think it could be intermittent failure?

mrocklin:
Yes.  That's likely intermittent.  My apologies for the testing friction.  I've restarted it.  This looks good to me.  Merging on a green light from Travis.

mrocklin:
This is in.  Thanks @Spacerat !

Spacerat:
Great - thanks for getting this in @mrocklin !

PR #1887, opened 2018-04-02
WIP: Add option to ignore error in get_dataset

jakirkham:
When calling `get_dataset` with a non-existent name, the resulting exceptions (including logging thereof) is a bit loud. This remains the case when catching and handling the `KeyError` that propagates upward. To make this a little nicer, add an `error` keyword argument to `get` and `get_dataset` to ensure that the `KeyError` can be optionally suppressed. This silences the logging while allowing users to easily check if they got what they expected.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1827,8 +1827,10 @@ class Client(Node):
         return self.sync(self.scheduler.publish_list, **kwargs)
 
     @gen.coroutine
-    def _get_dataset(self, name):
-        out = yield self.scheduler.publish_get(name=name, client=self.id)
+    def _get_dataset(self, name, **kwargs):
+        out = yield self.scheduler.publish_get(
+            name=name, client=self.id, **kwargs
+        )
 
         with temp_default_client(self):
             data = loads(out['data'])
--- a/distributed/publish.py
+++ b/distributed/publish.py
@@ -40,11 +40,11 @@ class PublishExtension(object):
         with log_errors():
             return list(sorted(self.datasets.keys()))
 
-    def get(self, stream, name=None, client=None):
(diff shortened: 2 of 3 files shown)


distributed/client.py
```diff
@@ -1827,8 +1827,10 @@ def list_datasets(self, **kwargs):
         return self.sync(self.scheduler.publish_list, **kwargs)
 
     @gen.coroutine
-    def _get_dataset(self, name):
-        out = yield self.scheduler.publish_get(name=name, client=self.id)
+    def _get_dataset(self, name, **kwargs):
+        out = yield self.scheduler.publish_get(
```
martindurant: If the only purpose is to sanitise the stack-trace, then a check on the dataset's existence could be made here before attempting to get, raising KeyError?

jakirkham: So then raise here as well for backcompat? Wouldn't that still make things noisy?

martindurant: I was thinking raise here only, and since this is a local exception from the client, it will not get deeply buried.

jakirkham: Ok that could work then. Never totally sure how things like Tornado deal with these issues. Plus not super familiar with what happens in this part of the codebase. 😄

Though maybe we do away with the check entirely and just call `get`. If it doesn't exist, it will return `None`, which is easy to check.

martindurant: Both sounds equally uncomplicated to me (checking before calling get, versus using `None` for not found), on an API that is not user-facing, so it's just a question of style.

jakirkham: Thanks for the advice. PR ( https://github.com/dask/distributed/pull/1888 ) should do the trick. Please let me know what you think. :)

PR #1888, opened 2018-04-02
Santize get_dataset trace

jakirkham:
Closes https://github.com/dask/distributed/pull/1887

Bump up where `KeyError` is raised in the `get_dataset` stack to cutdown on noise in the logs while preserving the expected behavior in the user facing API.

cc @martindurant

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1829,6 +1829,8 @@ class Client(Node):
     @gen.coroutine
     def _get_dataset(self, name):
         out = yield self.scheduler.publish_get(name=name, client=self.id)
+        if out is None:
+            raise KeyError("Dataset '%s' not found" % name)
 
         with temp_default_client(self):
             data = loads(out['data'])
--- a/distributed/publish.py
+++ b/distributed/publish.py
@@ -42,10 +42,7 @@ class PublishExtension(object):
 
     def get(self, stream, name=None, client=None):
         with log_errors():
-            if name in self.datasets:
-                return self.datasets[name]
-            else:
-                raise KeyError("Dataset '%s' not found" % name)
+            return self.datasets.get(name, None)
 
 
 class Datasets(MutableMapping):


martindurant:
I'm OK with it.
Could use a little test.
@mrocklin , `raise` is enough in a coroutine, does not need some `gen` ?

jakirkham:
What would you suggest testing? FWIW there are tests for `get_dataset` when the dataset does and does not exist already.

Always wonder the same thing with Tornado. Seems [just raising is fine]( https://stackoverflow.com/a/43942549 ).

martindurant:
Ah, so indeed you replaces the complex exception with the cleaner one, and so of course the tests are happy :) In that case, all good with me.

jakirkham:
Great. FWIW would be ok check the log too, but would need advice on how to approach it.

jakirkham:
Hmm...so this is still a bit noisier than I'd like. Will see about moving the exception to `get_dataset` instead.

jakirkham:
Seems that is not possible as `_get_dataset` returns a `Future`. Suggestions?

For context, this is the red text that shows up any time we check for a named dataset that doesn't exist. This happens regardless of whether we catch the `KeyError` thrown outside the Tornado `IOLoop`.

```python
distributed.utils - ERROR - "Dataset 'raw' not found"
  File "/opt/conda3/lib/python3.6/site-packages/distributed/utils.py", line 238, in f
    result[0] = yield make_coro()
  File "/opt/conda3/lib/python3.6/site-packages/tornado/gen.py", line 1099, in run
    value = future.result()
  File "/opt/conda3/lib/python3.6/site-packages/tornado/gen.py", line 1113, in run
  File "/opt/conda3/lib/python3.6/site-packages/distributed/client.py", line 1833, in _get_dataset
    raise KeyError("Dataset '%s' not found" % name)
KeyError: "Dataset 'raw' not found"
```

mrocklin:
You want to control line 240 in utils.py.  We don't necessarily want to log every error that occurs with sync.  I'm actually inclined to just remove this line entirely and see what happens.  We always raise the exception anyway, so it doesn't seem necessary to log it as well.

jakirkham:
That feels a little heavy-handed to me. Wouldn't be surprised if someone came and gripped about that change after the fact. Is there any way for us to fix this issue locally?

mrocklin:
https://github.com/dask/distributed/pull/1894

jakirkham:
Think this is good to go.

PR #1889, opened 2018-04-02
Drop `__contains__` from `Datasets`

jakirkham:
By default, the `Mapping` interface provides a method for `__contains__`,  which `MutableMapping` inherits. So there is no need for this method to be included explicitly on `Datasets`.

### DIFF
--- a/distributed/publish.py
+++ b/distributed/publish.py
@@ -68,9 +68,6 @@ class Datasets(MutableMapping):
     def __delitem__(self, key):
         self.__client.unpublish_dataset(key)
 
-    def __contains__(self, key):
-        return key in self.__client.list_datasets()
-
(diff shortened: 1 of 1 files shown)


jakirkham:
Blocked by PR ( https://github.com/dask/distributed/pull/1888 ).

mrocklin:
@jakirkham what is the status here?

jakirkham:
Just now rechecked using a development copy of Distributed, looks like the log noise is gone. IOW running the code below generates no logging messages.

```python
try:
    client.get_dataset("foo")  # No "foo" dataset published
except KeyError:
    pass
```

jakirkham:
Should be good to merge :)

PR #1894, opened 2018-04-03
Don't log errors in sync

mrocklin:
We raise anyway, so lets let that process be enough.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -237,7 +237,6 @@ def sync(loop, func, *args, **kwargs):
             thread_state.asynchronous = True
             result[0] = yield make_coro()
         except Exception as exc:
-            logger.exception(exc)
             error[0] = sys.exc_info()
         finally:
             thread_state.asynchronous = False


jakirkham:
Thanks Matt.

So this partially addresses it. Appears that [line 627]( https://github.com/mrocklin/distributed/blob/532263602113f02a501cf574509ae8655f943d5d/distributed/utils.py#L627 ) also logs exceptions. Once that is also dropped, there still appears to be a warning from the scheduler with the traceback on the scheduler end included followed by an actual traceback outside the scheduler. Not sure what causes that last one to be logged.

Independently there is a question of whether we are ok disabling all of these logging features with exceptions. Personally wouldn't want to mess anyone up if they are relying on this logging of exceptions. We could soften this to an option if we think they are used. Though you would know better than I if these are used somewhere.

mrocklin:
So if you don't think we should log KeyErrors in `PublishExtension.get` then you should remove the `with log_errors()` context manager in that function.  I might recommend explicitly including an info or debug level logging event on a missing key though if we think it's somewhat errant behavior.

I'm going to avoid that in this PR.  This is a larger and more broadly sweeping change.

PR #1897, opened 2018-04-04
Add count method to as_completed

mrocklin:
Thoughts on other possible names for `count`?  I also considered using `__len__`


martindurant:
Could be more explicit, like `nrunning` (do we also keep count of how many have been completed, like for a progress bar? I guess not).

mrocklin:
I've updated the docstring to be more explicit

PR #1900, opened 2018-04-04
Fix `UnboundLocalError` for `key`

jakirkham:
Appears `key` in this context was not necessarily defined, which caused some errors to bubble up from the scheduler. From context, it seems like this should be `ts.key`. So this changes it to `ts.key`.


mrocklin:
Interesting.  How did you run across this?  I'm somewhat surprised that this was caught neither by tests nor a linter.

jakirkham:
From running into a different error that I've yet to diagnose.

jakirkham:
FWIW `key` is used in `for`-loops earlier. Just never defined independently of a looping variable. So that could have confused the code linter maybe.

PR #1903, opened 2018-04-04
Intern result of key_split

mrocklin:
We leak strings.  This isn't a huge deal, but its noticable after running streaming computations for a long while.  This is small attempt to fix a small part of that leak.

### DIFF
--- a/distributed/compatibility.py
+++ b/distributed/compatibility.py
@@ -14,6 +14,7 @@ if sys.version_info[0] == 2:
 
     reload = reload
     unicode = unicode
+    intern = intern
     PY2 = True
     PY3 = False
     ConnectionRefusedError = OSError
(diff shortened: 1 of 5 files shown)


jakirkham:
So I'm not very familiar with `intern`ing strings in Python. However [the docs seem to suggest we need to "keep a reference to the return value"]( https://docs.python.org/3/library/sys.html#sys.intern ). Given this, should we be caching these somewhere as well? If this is already happening somewhere and I'm merely unaware, feel free to ignore this comment.

mrocklin:
I think it's ok.  I think that Python cleans things up on its end, though I'm not sure.

This turns out to be a bigger problem in Python 2, where interning unicode is not supported.

mrocklin:

PR #1904, opened 2018-04-04
Extend default timeout to 10s

mrocklin:
As we see people operate in more extreme situations (larger clusters,
larger graphs, weirder codebases) we experience longer periods of
non-communication.  It seems sensible to extend the default timeout to
10s.  We also double the default timeout for restarts, which are
particularly problematic.

### DIFF
--- a/distributed/bokeh/tests/test_scheduler_bokeh.py
+++ b/distributed/bokeh/tests/test_scheduler_bokeh.py
@@ -349,7 +349,7 @@ def test_GraphPlot_clear(c, s, a, b):
         assert time() < start + 5
 
 
-@gen_cluster(client=True)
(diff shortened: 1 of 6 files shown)


dhirschfeld:
As a case in point I often see intermittent failures of the form:
```
OSError: Timed out trying to connect to 'tcp://scheduler:8786' after 3 s: connect() didn't finish in time
```
So I'm hoping this PR may help fix that. I hadn't gotten around to increasing the timeout myself as the problem was harmless due to automatic retries in my codebase.

jakirkham:
For clarity, this only changes the default. It is [user configurable currently]( https://github.com/dask/distributed/blob/1.21.5/distributed/config.yaml#L54 ).

PR #1908, opened 2018-04-07
Fix PR cross references' format

jakirkham:
Some of the PR cross references had the wrong format. So when Sphinx generates the docs, these don't link to the GitHub PR. Fixed to match the expected format to ensure the cross references work as intended.


mrocklin:
Ah, whoops.  Indeed.  Thanks @jakirkham

PR #1917, opened 2018-04-13
DOC: new lines needed for doctest rendering

lesteve:
From https://distributed.readthedocs.io/en/latest/local-cluster.html:
image
From "Add a new worker to the cluster" the rendering is off.

With this change:
image


mrocklin:
Looks great.  Thank you!  Merging.

PR #1927, opened 2018-04-19
Replace msgpack-python by msgpack

### LINKED ISSUE #1913: problem importing msgack-python

lesteve:
This is the new name of the package on PyPI. The [msgpack-python PyPI page](https://pypi.org/project/msgpack-python/) says "This package is deprecated. Install msgpack instead."

This may just fix #1913 as mentioned in https://github.com/dask/distributed/issues/1913#issuecomment-382798935.

Close #1794.

### DIFF
--- a/continuous_integration/setup_conda_environment.cmd
+++ b/continuous_integration/setup_conda_environment.cmd
@@ -30,7 +30,6 @@ call deactivate
     joblib ^
     jupyter_client ^
     mock ^
-    msgpack-python ^
     psutil ^
     pytest=3.1 ^
     python=%PYTHON% ^
@@ -49,6 +48,7 @@ call activate %CONDA_ENV%
 %PIP_INSTALL% git+https://github.com/dask/zict --upgrade
 
 %PIP_INSTALL% pytest-repeat pytest-timeout pytest-faulthandler sortedcollections
+%PIP_INSTALL% msgpack
 
 @rem Display final environment (for reproducing)
 %CONDA% list
--- a/continuous_integration/setup_conda_environment.cmd
+++ b/continuous_integration/setup_conda_environment.cmd
@@ -30,7 +30,7 @@ call deactivate
     joblib ^
     jupyter_client ^
     mock ^
-    msgpack ^
+    msgpack-python ^
     psutil ^
     pytest=3.1 ^
     python=%PYTHON% ^
--- a/continuous_integration/setup_conda_environment.cmd
+++ b/continuous_integration/setup_conda_environment.cmd
@@ -30,7 +30,7 @@ call deactivate
     joblib ^
     jupyter_client ^
     mock ^
-    msgpack-python ^
+    msgpack ^
     psutil ^
     pytest=3.1 ^
     python=%PYTHON% ^
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -60,7 +60,7 @@ pip install -q git+https://github.com/dask/dask.git --upgrade --no-deps
 pip install -q git+https://github.com/joblib/joblib.git --upgrade --no-deps
 pip install -q git+https://github.com/dask/s3fs.git --upgrade --no-deps
 pip install -q git+https://github.com/dask/zict.git --upgrade --no-deps
-pip install -q sortedcollections msgpack-python --no-deps
+pip install -q sortedcollections msgpack --no-deps
 pip install -q keras --upgrade --no-deps
 
 if [[ $CRICK == true ]]; then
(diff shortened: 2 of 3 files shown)


mrocklin:
This looks fine to me.  cc local packaging expert, @jakirkham

lesteve:
Hmmm not sure about the CI failures, I'll need to investigate but I'll retrigger the CIs just in case.

After opening the PR, I realised that msgpack was not available through conda at the moment. There is an issue in conda-forge related to this renaming https://github.com/conda-forge/msgpack-python-feedstock/issues/5.

continuous_integration/setup_conda_environment.cmd
```diff
@@ -30,7 +30,7 @@ call deactivate
     joblib ^
     jupyter_client ^
     mock ^
-    msgpack-python ^
+    msgpack ^
```
jakirkham: So this part is using `conda` where it is still named `msgpack-python`. Seems to be causing the CI failure. Reverting this part should fix that issue.

jakirkham:
Yeah the package is not likely to renamed in `conda-forge` for basically the same reason it won't be renamed in `defaults`. Namely there are many `msgpack` implementations for different languages and we need a clean way to disambiguate them as we do package stuff for multiple languages.

continuous_integration/setup_conda_environment.cmd
lesteve: I would have guessed this is only used in Windows though, so that does not really explain the Travis failure. I pushed the change anyway.

lesteve:
> Yeah the package is not likely to renamed in conda-forge for basically the same reason it won't be renamed in defaults

Thanks for the details. I am wondering how this is going to pan out with the change of name in PyPI and the entry point issue detailed in #1913. If you have some suggestions on whether #1913 can be solved in setup.py I am interested.

lesteve:
Right so as often I did not look thoroughly enough in the existing issues and it turns out there is already a PR for that. Closing in favour of #1794.

mrocklin:
Is this ok to go in?  cc @jakirkham

jakirkham:
Thanks @lesteve. :)

jakirkham:
Now wondering if that was a question for me or @lesteve. Sorry if I merged early.

lesteve:
This was fine to merge as far as I was concerned. I may try to do the renaming on the conda-forge side at one point.

mrocklin:
Great.  Glad that this is in.  Thanks @lesteve !

rbubley:
There is a already a conda-forge PR. Once that’s merged, the work-around here can probably be simplified.

lesteve:
> Once that’s merged, the work-around here can probably be simplified.

Agreed, once https://github.com/conda-forge/msgpack-python-feedstock/pull/7 is merged, we can go back to using conda to install msgpack-python on AppVeyor.

PR #1937, opened 2018-04-26
Accept KeyError when closing tornado IOLoop in tests

mrocklin:
Previously this would raise a confusing KeyError.

See https://github.com/tornadoweb/tornado/issues/2367

We ignore this for now, which *seems* innocuous.


mrocklin:
Fast-merging this for now.  Tests on other PRs are failing due to this.

PR #1947, opened 2018-04-28
Support submit keywords in joblib

mrocklin:
This allows for the use of keywords like retries=, priority=, and so on
in the Joblib Backend

```python
with joblib.parallel_backend('dask', retries=10):
    ...
```

We store the submit keyword arguments on the DaskParallelBackend

Fixes https://github.com/dask/dask-ml/issues/164


mrocklin:
Merging this in 24 hours if there are no further comments

PR #1948, opened 2018-04-30
Overhaul configuration

mrocklin:
This is a large overhaul of our configuration.  It depends on https://github.com/dask/dask/pull/3432 which centralizes configuration code in dask/dask.

In particular this does the following:

1.  The config dict moves from distributed.config to dask.config.config
2.  We now depend on the config values in distributed/distributed.yaml being in the config, allowing us to drop default values that were previously sprinkled around the codebase
3.  We change our configuration keys to be more hierarchical, so rather than `{'scheduler-wait-time': ...}` we get values more like `{'scheduler': {'wait-time': ...}}`. 
4.  We establish a mechanism to migrate old config names to new ones.


mrocklin:
This could use review by people familiar with configuration systems.  

cc @jakirkham @minrk @jacobtomlinson if you have time.

distributed/security.py
```diff
@@ -49,7 +49,7 @@ class Security(object):
     __slots__ = tuple(_fields)
 
     def __init__(self, **kwargs):
-        self._init_from_dict(config)
+        self._init_from_dict(dask.config.config)
```
jacobtomlinson: Did you mean to double `.config` here?

distributed/scheduler.py
```diff
@@ -1357,7 +1355,7 @@ def update_graph(self, client=None, tasks=None, keys=None,
         if isinstance(user_priority, Number):
             user_priority = {k: user_priority for k in tasks}
 
-        priority = priority or order(tasks)  # TODO: define order wrt old graph
+        priority = priority or dask.order.order(tasks)  # TODO: define order wrt old graph
```
jacobtomlinson: Did you mean to double `.order` here?

distributed/security.py
mrocklin: Yes, the dask.config module has a config dict.  See https://github.com/dask/dask/pull/3432 for the co-PR

distributed/scheduler.py
mrocklin: Yes, same as above.  This change is somewhat orthogonal to this PR, I just decided to start using namespaces more heavily rather than direct imports.

mrocklin:
Tests will fail hard on this PR until https://github.com/dask/dask/pull/3432 is merged.  The next release will have a strict dependence between dask/dask and dask/distributed.

Issue #1952, opened 2018-05-02
distributed.Client.processing method hangs when called from async code 


hdoupe:
I'm tinkering around with building a [REST api in front of a dask cluster](https://github.com/hdoupe/async-dask-api). I'd like to be able to query the dask scheduler to see which jobs are still running. However, the `distributed.Client.processing` method appears to hang when called from asynchronous code. It appears to work just fine with synchronous code. I have a simple tornado application that demonstrates this:

python versoin: 3.6.5
tornado version: 5.0.2
distributed version: 1.21.6
dask version: 0.17.2

```
import json

from tornado.web import Application, RequestHandler
from tornado.ioloop import IOLoop

from distributed import Client

class AsyncTaskStatus(RequestHandler):

    async def get(self):
        print('GET')
        client = await Client(asynchronous=True)
        print('got client', client)
        print('client', client._scheduler_identity)

        # gets hung up here
        processing = await client.processing()

        print('processing tasks', processing)

        self.write(json.dumps({addr: list(processing[addr]) for addr in processing}))


class SyncTaskStatus(RequestHandler):

    def get(self):
        print('GET')
        client = Client()
        print('got client', client)
        print('client', client._scheduler_identity)

        # works just fine
        processing = client.processing()

        print('processing tasks', processing)

        self.write(json.dumps({addr: list(processing[addr]) for addr in processing}))


class Health(RequestHandler):

    async def get(self):
        print('GET-HEALTH')
        self.write('healthy...')


def make_app():
    return Application(
        [(r'/asynctaskstatus', AsyncTaskStatus),
         (r'/synctaskstatus', SyncTaskStatus),
         (r'/health', Health)],
        default_host='0.0.0.0',
        debug=True,
        autoreload=True
    )

if __name__ == "__main__":
    print('starting up app...')
    app = make_app()
    app.listen(8888)
    IOLoop.current().start()
```

API calls:
```
HDoupe-MacBook-Pro:async-dask-api henrydoupe$ http http://localhost:8888/synctaskstatus
HTTP/1.1 200 OK
Content-Length: 116
Content-Type: text/html; charset=UTF-8
Date: Wed, 02 May 2018 21:35:50 GMT
Etag: "51ddcc8412a5a82c61422a58b84aab85d7f847e6"
Server: TornadoServer/5.0.2

{
    "tcp://127.0.0.1:55274": [],
    "tcp://127.0.0.1:55278": [],
    "tcp://127.0.0.1:55280": [],
[...]

PR #1962, opened 2018-05-06
Use self.sync so Client.processing works in asynchronous context

hdoupe:
This PR attempts to close issue #1952 using the fix suggested by @mrocklin in this [comment](https://github.com/dask/distributed/issues/1952#issuecomment-386140014).

I also added a couple test cases that hang without this fix. Please let me know if there is a better way to include these tests or if there are other tests that should be added.

Also, `Client.stacks` does not use the `Client.sync` method either. Should I go ahead and use `self.sync` there, too?

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -3354,6 +3354,12 @@ def test_get_processing(c, s, a, b):
     x = yield c.scheduler.processing(workers=[a.address])
     assert isinstance(x[a.address], list)
 
+    x = yield c.processing()
+    assert set(x) == {a.address, b.address}
+
+    x = yield c.processing(workers=[a.address])
+    assert isinstance(x[a.address], list)
+
 
 @gen_cluster(client=True)
 def test_get_foo(c, s, a, b):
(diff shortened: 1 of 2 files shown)


distributed/tests/test_client.py
```diff
@@ -3354,6 +3354,12 @@ def test_get_processing(c, s, a, b):
     x = yield c.scheduler.processing(workers=[a.address])
     assert isinstance(x[a.address], list)
```
mrocklin: We can probably remove this code above now that `client.processing` works directly.

hdoupe:
Great. No problem, this is a really cool project, and I am glad that I could help out.

I removed the stacks method and to the best of my knowledge, all references to it.

distributed/tests/test_client.py
hdoupe: Done. I glanced through the other tests in `test_client.py`. A lot of them access methods similar to `processing` like `has_what` through the `scheduler` attribute. I'm not sure if that's by design or if that's how it was done in an earlier version of the API.

mrocklin:
Merged.  Thanks for contributing this @hdoupe !

Issue #1487, opened 2017-10-20
Print newline after text progress bar completes


jakirkham:
In the notebook, when the text progress bar completes, it does not currently print a newline. The result is anything printed afterwards gets smashed into the same line against the progress bar. Further if additional text progress bars are created, they simply overwrite the previous text progress bar. This makes hard to tell how many steps have been completed in a computation. To fix these issues, it would be nice to print a newline right after the text progress bar completes.

mrocklin:
You're talking about `dask.distributed.progress`, yes?  Not `dask.diagnostics.ProgressBar`?

I've never used the text progress bar in the notebook, instead preferring ipywidgets.  I'm curious, why do you find yourself using the text version?

Of course we should fix it regardless.

jakirkham:
Correct.

Because we were unable to get the widgets to (optionally) block yet. We discussed this before, but a long time ago in issue ( https://github.com/dask/distributed/issues/926 ). May be worth revisiting that again at some point, but can't revisit it personally ATM.

Agreed. Think it should be easy to fix. Wouldn't mind taking a look if no one else does, but can't do it soon. So figured this issue should be mentioned before I forget. :)

mrocklin:
Sounds good

mrocklin:
I believe that this is now resolved.

jakirkham:
Is there a PR with the fix somewhere?

mrocklin:
Oh I'm sorry, I saw the "merged" PR link above and assumed that it was a dask issue.  I see now that this was a nanshe PR.

jakirkham:
Sorry was just doing that for my own bookkeeping. Didn't mean to cause confusion.

jakirkham:
Resolved in PR ( https://github.com/dask/distributed/pull/1976 ).

PR #1976, opened 2018-05-09
insert newline by default after TextProgressBar

ptooley:
PR to add functionality requested in issue #1487.

New option to progress() allows the old behaviour to be reproduced but default to adding a newline as that seems to be more likely what most people will want.

### DIFF
--- a/distributed/diagnostics/progressbar.py
+++ b/distributed/diagnostics/progressbar.py
@@ -103,12 +103,11 @@ class ProgressBar(object):
 
 class TextProgressBar(ProgressBar):
     def __init__(self, keys, scheduler=None, interval='100ms', width=40,
-                 loop=None, complete=True, start=True, newline=True):
+                 loop=None, complete=True, start=True):
         super(TextProgressBar, self).__init__(keys, scheduler, interval,
                                               complete)
         self.width = width
         self.loop = loop or IOLoop()
-        self.newline = newline
 
         if start:
             loop_runner = LoopRunner(self.loop)
@@ -126,9 +125,8 @@ class TextProgressBar(ProgressBar):
             sys.stdout.flush()
 
     def _draw_stop(self):
-        if self.newline:
-            sys.stdout.write('\n')
-            sys.stdout.flush()
+        sys.stdout.write('\n')
+        sys.stdout.flush()
 
 
 class ProgressWidget(ProgressBar):
@@ -350,9 +348,6 @@ def progress(*futures, **kwargs):
     complete: bool (optional)
         Track all keys (True) or only keys that have not yet run (False)
         (defaults to True)
-    newline: bool (optional)
-        Insert a newline after the progress bar. Ignored in notebooks.
-        (defaults to True)
 
     Notes
     -----
@@ -368,7 +363,6 @@ def progress(*futures, **kwargs):
     notebook = kwargs.pop('notebook', None)
     multi = kwargs.pop('multi', True)
     complete = kwargs.pop('complete', True)
-    newline = kwargs.pop('newline', True)
     assert not kwargs
 
     futures = futures_of(futures)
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for doing this @ptooley !

Three (or two, depending) small comments:

1.  Can we test this functionality?  Hopefully some of the tests in distributed/diagnostics/tests/test_progressbar.py can be adapted here
2.  Do we want to preserve the old functionality?
3.  If we do then can we rename the keyword to `newline=True` to avoid the double-negative

ptooley:
Agree, double-negatives bad, fixed.

I don't see a reason not to keep the old functionality around, someone may well be using it and it introduces minimal additional code.

As far as testing goes, the newline breaks the current tests because of the way the output is captured.  `check_bar_completed` should grab the second-last line now rather than the last.  I am inclined to fix the current tests for the new default case and add a single extra test of the `newline=False` case.  Does that sound logical to you?

mrocklin:
That sounds fine to me.  

> I don't see a reason not to keep the old functionality around, someone may well be using it and it introduces minimal additional code.

One reason not to keep it around is an accrual of functionality and increased maintenance costs.  If you think that this is likely to be preferred then I'm inclined to remove the old functionality until someone arrives who needed the older functionality.

mrocklin:
Looks like some failing tests here, in particular I just tried this locally and got the following:

```python
~/workspace/distributed/distributed/diagnostics/progressbar.py in listen(self)
     75             if response['status'] in ('error', 'finished'):
     76                 yield self.comm.close()
---> 77                 self._draw_stop(**response)
     78                 break
     79 

TypeError: _draw_stop() got an unexpected keyword argument 'all'
```

ptooley:
Yeah, sorry, missed the **kwargs in the signature.  Unfortunately I haven't out how to run the tests locally. Pytest just throws a bunch of errors.

mrocklin:
> Pytest throws an AttributeError trying to access dask.config.config

Testing dask/distributed master requires dask/dask master.  You might want to install from github

    pip install git+https://github.com/dask/dask.git --upgrade --no-deps

mrocklin:
It looks like there are still some failing tests:

```
distributed/diagnostics/tests/test_progressbar.py:71: 
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/tornado/ioloop.py:581: in run_sync
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/tornado/gen.py:1113: in run
distributed/diagnostics/tests/test_progressbar.py:65: in f
    check_bar_completed(capsys)
capsys = <_pytest.capture.CaptureFixture object at 0x7fafc64730f0>, width = 40
    def check_bar_completed(capsys, width=40):
        out, err = capsys.readouterr()
        # trailing newline so grab next to last line for final state of bar
>       bar, percent, time = [i.strip() for i in out.split('\r')[-2].split('|')]
E       ValueError: not enough values to unpack (expected 3, got 1)
```

You can also see these by clicking on the `details` link next to the travis-ci check and then following down to one of the four builds.

mrocklin:
Hi @ptooley , any luck with this?  If you need help running tests locally let me know.  I'd be happy to help.

jakirkham:
Maybe it's not worth adding a test for something so simple. Will likely take significantly more effort than the fix itself.

ptooley:
Sorry for the radio silence, real life got in the way of productive coding...

I got the tests running in a conda env and the breakage was just a typo from me in the end.  I have rebased my changes on latest master and hopefully they come up good on the CI.

ptooley:
So digging into the remaining test failure locally getting sporadic failures (~5% of runs) on the current master as well as my branch.  Additionally the offending test doesn't touch my changes.

mrocklin:
Indeed.  That's related to a recent PR.  I'll push up a fix shortly.

mrocklin:
The intermittently failing test is resolved in https://github.com/dask/distributed/pull/2112

Merging this.  Thank you for following through @ptooley !

jakirkham:
Thanks @ptooley :)

PR #1979, opened 2018-05-13
Remove the reference to dask-ec2 from docs.

j-bennet:
dask-ec2 is deprecated, so that's misleading.


mrocklin:
Agreed.  Thanks @j-bennet !

In the future we might consider removing this entire section in favor of http://dask.pydata.org/en/latest/setup.html

PR #1980, opened 2018-05-15
add comm to packages

mrocklin:
Any suggestions on cleaner ways to ensure that we don't forget things like this in the future @jakirkham ?


jakirkham:
We could build the package and run the tests against the built package.

One way to do this would be to include a conda recipe for development purposes. The tests could then be run on the conda package as part of the test phase.

We could do something similar with a pip installed version of the package. Would just need to be more careful about making sure the cloned copy of the code doesn't contaminant the testing environment.

jcrist:
> We could do something similar with a pip installed version of the package. Would just need to be more careful about making sure the cloned copy of the code doesn't contaminant the testing environment.

I think dropping the `-e` from the `pip install` command installing distributed (end of `install.sh`), and `cd`ing to a different directory before running the tests (removes git repo from pythonpath) should work.

Something like:

```
cd docs  # Just some other directory
py.test ../distributed
```

PR #1989, opened 2018-05-18
Pause execution during communication if under memory pressure

mrocklin:
This avoids run-away processes when the critical path is communication
bound and there is ample available work that fills up memory.


mrocklin:

PR #1992, opened 2018-05-21
Add worker_key parameter to Adaptive

mrocklin:
This allows adaptive clusters to intelligently close down groups of
workers based on some logical association.

See https://github.com/dask/dask-jobqueue/pull/63 for motivation

cc @jhamman

### DIFF
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -41,6 +41,9 @@ class Adaptive(object):
     target_duration: timedelta or str, default "5s"
         Amount of time we want a computation to take.
         This affects how aggressively we scale up.
+    worker_key: Callable[WorkerState]
+        Function to group workers together when scaling down
+        See Scheduler.workers_to_close for more information
     minimum: int
         Minimum number of workers to keep around
     maximum: int
@@ -84,8 +87,9 @@ class Adaptive(object):
 
     def __init__(self, scheduler, cluster, interval='1s', startup_cost='1s',
                  scale_factor=2, minimum=0, maximum=None, wait_count=3,
-                 target_duration='5s', **kwargs):
+                 target_duration='5s', worker_key=lambda x: x, **kwargs):
         interval = parse_timedelta(interval, default='ms')
+        self.worker_key = worker_key
         self.scheduler = scheduler
         self.cluster = cluster
         self.startup_cost = parse_timedelta(startup_cost, default='s')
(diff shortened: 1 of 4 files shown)


mrocklin:
@jhamman any luck with this?  My current plan is to wait on you to try things.  No rush, I just wanted to make sure that you weren't waiting on me to merge this.

jhamman:
@mrocklin - So far so good. I'm testing this branch along with my mods in dask/dask-jobqueue#63 and things are working much better. I'll give this PR a quick review now.

distributed/deploy/adaptive.py
```diff
@@ -223,7 +227,8 @@ def workers_to_close(self, **kwargs):
     @gen.coroutine
     def _retire_workers(self, workers=None):
         if workers is None:
-            workers = self.workers_to_close()
+            workers = self.workers_to_close(key=self.worker_key,
+                                            minimum=self.minimum)
```
jhamman: what is the logic for passing in minimum here? Should you also pass in maximum?

distributed/deploy/local.py
```diff
@@ -145,12 +146,24 @@ def __repr__(self):
...
+        asynchronous = kwargs.pop('asynchronous', None)
+        if asynchronous or self._asynchronous or getattr(thread_state, 'asynchronous', False):
+            callback_timeout = kwargs.pop('callback_timeout', None)
+            future = func(*args, **kwargs)
+            if callback_timeout is not None:
+                future = gen.with_timeout(timedelta(seconds=callback_timeout),
+                                          future)
+            return future
+        else:
+            return sync(self.loop, func, *args, **kwargs)
```
jhamman: Is this related to the grouped worker issue? (I'm not following)

jhamman:
@mrocklin - save my few questions above, this seems to working and I'd like to see it merged in the next few days. Currently, this is blocking https://github.com/dask/dask-jobqueue/pull/63.

distributed/deploy/adaptive.py
mrocklin: Minimum here corresponds to the minimum number of workers that we want to keep.  The `workers_to_close` function is a slightly nicer place to handle this than outside.  For example if `workers_to_close` tells us that we need to close all of our workers, we then look at minimum and see that we need one, which worker should we keep?  Also, which *other* workers should we keep that are in the same group.  Suddenly we have to replicate the grouping logic.

Maximum doesn't make sense because we're only reducing the number of workers, not increasing them.

distributed/deploy/local.py
mrocklin: Something was slightly nicer if we switched to using self.sync here (this is a common pattern in the client).  I decided to clean this up as part of this PR.

mrocklin:
I'm fine to merge this any time.  I'll give you a chance to respond to recent commentary but I'll merge after you give the go-ahead.

jhamman:
Thanks @mrocklin - go ahead and merge whenever now.

PR #1995, opened 2018-05-22
Allow `name` to be explicitly passed in publish_dataset

mariusvniekerk:
We can now make datasets that don't have string names

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1806,7 +1806,7 @@ class Client(Node):
             def add_coro(name, data):
                 keys = [tokey(f.key) for f in futures_of(data)]
                 coroutines.append(self.scheduler.publish_put(keys=keys,
-                                                             name=dumps(name),
+                                                             name=name,
                                                              data=to_serialize(data),
                                                              client=self.id))
 
@@ -1890,7 +1890,7 @@ class Client(Node):
         --------
         Client.publish_dataset
         """
-        return self.sync(self.scheduler.publish_delete, name=dumps(name), **kwargs)
+        return self.sync(self.scheduler.publish_delete, name=name, **kwargs)
 
     def list_datasets(self, **kwargs):
         """
@@ -1905,7 +1905,7 @@ class Client(Node):
 
     @gen.coroutine
     def _get_dataset(self, name):
-        out = yield self.scheduler.publish_get(name=dumps(name), client=self.id)
+        out = yield self.scheduler.publish_get(name=name, client=self.id)
         if out is None:
             raise KeyError("Dataset '%s' not found" % name)
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1799,20 +1799,23 @@ class Client(Node):
                          force=force)
 
     @gen.coroutine
-    def _publish_dataset(self, *args, name=None, **kwargs):
+    def _publish_dataset(self, *args, **kwargs):
         with log_errors():
             coroutines = []
 
             def add_coro(name, data):
                 keys = [tokey(f.key) for f in futures_of(data)]
                 coroutines.append(self.scheduler.publish_put(keys=keys,
-                                                             name=tokey(name),
+                                                             name=dumps(name),
                                                              data=to_serialize(data),
                                                              client=self.id))
 
+            name = kwargs.pop('name', None)
             if name:
                 if len(args) == 0:
-                    raise ValueError("If name is provided, data must be provided as args not keyword args")
+                    raise ValueError(
+                        "If name is provided, expecting call signature like"
+                        " publish_dataset(df, name='ds')")
                 # in case this is a singleton, collapse it
                 elif len(args) == 1:
                     args = args[0]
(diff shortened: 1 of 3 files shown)


martindurant:
Have you used `client.datasets`? Since that uses set/get item notation, you can already use it to specify arbitrary keynames for datasets.

Not that there's anything wrong with what is here - but the docs should specify that providing a name is only valid if one dataset is being passed.

distributed/client.py
```diff
@@ -1831,7 +1848,7 @@ def publish_dataset(self, **kwargs):
         Client.unpublish_dataset
         Client.persist
         """
-        return self.sync(self._publish_dataset, **kwargs)
+        return self.sync(self._publish_dataset, *args, name=name, **kwargs)
```
mariusvniekerk: Ugh this isn't valid python 2.  Suggestions?

mrocklin: kwargs['name'] = name

martindurant:
Then perhaps there should be a test to show the alternate usage. The example has a string name.

mariusvniekerk:
So these fail atm because tuple seems to turn into list?  What is the correct way to avoid that?

mariusvniekerk:
well i pickle and unpickle the names, so you should be able to interpret the state on the scheduler. since they should only be pickled during transit

mrocklin:
> well i pickle and unpickle the names, so you should be able to interpret the state on the scheduler. since they should only be pickled during transit

Yeah, only by unpickling though, which is less than ideal.  

I understand the concern about msgpack deserializing tuples as unhashable lists though.  I don't yet know what to do about this.

mrocklin:
One rather extreme approach would be to use to have msgpack always unpack arrays as tuples rather than lists

```python
In [1]: import msgpack

In [2]: msgpack.dumps([1, 2, 3])
Out[2]: b'\x93\x01\x02\x03'

In [3]: msgpack.loads(msgpack.dumps([1, 2, 3]), use_list=False)
Out[3]: (1, 2, 3)

In [4]: msgpack.loads(msgpack.dumps([1, 2, 3]), use_list=True)
Out[4]: [1, 2, 3]
```

This would almost certainly require changes through the codebase, but hopefully not in that many places.  It's a bit of a yak-shave, but might turn out to be a good choice generally.

mrocklin:
This PR might be more feasible now.  I think that we should remove the calls to pickle.dumps/loads

mariusvniekerk:
Rebased this and removed the serialization.

mariusvniekerk:
@mrocklin The test failures here seem unrelated?

mrocklin:
Yeah, the windows failure appears to be new, and is consistently affecting all PRs. 

The error is intermittent, I'm looking at it now.  

The flake8 linting errors appear to be genuine.

```
distributed/publish.py:4:1: F401 '.protocol.pickle.dumps' imported but unused
distributed/publish.py:4:1: F401 '.protocol.pickle.loads' imported but unused
```

mrocklin:
Appveyor issue possibly solved with https://github.com/dask/distributed/pull/2060

distributed/tests/test_publish.py
```diff
@@ -37,6 +37,24 @@ def test_publish_simple(s, a, b):
...
+        data = yield c.scatter(range(3))
+        out = yield c.publish_dataset(data, name=name)
+        assert name in s.extensions['publish'].datasets
+        assert isinstance(s.extensions['publish'].datasets[name]['data'], Serialized)
+
+        datasets = yield c.scheduler.publish_list()
+        assert name in datasets
+
+    yield c.close()
+    yield f.close()
```
mrocklin: This should probably use a try-finally block to clean up the clients even if the test fails in the future.

distributed/tests/test_publish.py
```diff
@@ -37,6 +37,24 @@ def test_publish_simple(s, a, b):
...
+
+
```
mrocklin: Can I also ask for a test with the `client.datasets` dictionary?

```python
In [10]: client.datasets[('a', 1)] = 123
<ipython-input-10-b1bdb9334cdd> in <module>()
----> 1 client.datasets[('a', 1)] = 123

~/workspace/distributed/distributed/publish.py in __setitem__(self, key, value)
     61 
     62     def __setitem__(self, key, value):
---> 63         self.__client.publish_dataset(**{key: value})
     64 
     65     def __delitem__(self, key):

TypeError: publish_dataset() keywords must be strings
```

distributed/tests/test_publish.py
mariusvniekerk: Should this be done by the @gen_cluster rather than relying on the test to do it?

mariusvniekerk: I'd say thats out of scope for this pr.  I can make another one for that, since that might be handy to simplify some of the test suite

mrocklin:
This is in.  Thanks @mariusvniekerk .  Sorry for the delay

PR #1997, opened 2018-05-23
Exclude gc statistics under PyPy

mariusvniekerk:
Under pypy3 the gc module doesn't contain the things that dask expects.
For now treat it the same as PY2

### DIFF
--- a/distributed/compatibility.py
+++ b/distributed/compatibility.py
@@ -79,6 +79,8 @@ if sys.version_info[0] == 3:
     logging_names.update(logging._nameToLevel)
 
 
+import platform
+PYPY = platform.python_implementation().lower() == 'pypy'
 WINDOWS = sys.platform.startswith('win')
 
 
--- a/distributed/utils_perf.py
+++ b/distributed/utils_perf.py
@@ -5,7 +5,7 @@ import gc
 import logging
 import threading
 
-from .compatibility import PY2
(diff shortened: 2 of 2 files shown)


mariusvniekerk:
Under pypy3 starting up a scheduler throws the following exceptions

```Python traceback
/usr/local/site-packages/distributed/utils_perf.py in enable(self)
   153
   154         cb = self._gc_callback
-> 155         assert cb not in gc.callbacks
   156         # NOTE: a global ref to self is saved there so __del__ can't work
   157         gc.callbacks.append(cb)

AttributeError: module 'gc' has no attribute 'callbacks'
```

distributed/compatibility.py
```diff
@@ -79,6 +79,8 @@ def isqueue(o):
     logging_names.update(logging._nameToLevel)
 
 
+import platform
+PYPY = platform.python_implementation().lower() == 'pypy'
```
mrocklin: Do pypy and pypy3 look the same here?

mrocklin:
This looks good to me.  Thanks for submitting this @mariusvniekerk

distributed/compatibility.py
mrocklin: We don't support Python 2.6 anyway

mrocklin:
Thanks @mariusvniekerk !

PR #1999, opened 2018-05-23
First draft of PubSub

mrocklin:
This allows for direct communication between workers using a
publish/subscribe pattern.


mrocklin:
OK, when running locally (but over normal tcp sockets) two workers on my machine can roundtrip 1000 messages between each other in around two seconds, so we're seeing around 1ms latency in communication.

mrocklin:
OK, this seems usable to me.  I'm going to merge it in soonish.  I'm sure that we'll find things to fix, but I suspect that it will have better exposure while in master.

I've added docstrings for documentation.  I plan to handle narrative documentation in dask/dask within the futures documentation pages.

PR #2000, opened 2018-05-23
Use tuples in msgpack

mrocklin:
This makes msgpack decode arrays as tuples rather than lists, allowing them to be used within containers like set and dict that expect hashable elements.

cc @mariusvniekerk 

I took a quick look at this but then got side tracked by other things.  I thought I'd push it up here in case you were interested in carrying on.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3040,7 +3040,7 @@ class Client(Node):
         if check:
             # we care about the required & optional packages matching
             def to_packages(d):
-                return dict(sum(d['packages'].values(), []))
+                return dict(sum(map(list, d['packages'].values()), []))
             client_versions = to_packages(result['client'])
             versions = [('scheduler', to_packages(result['scheduler']))]
             versions.extend((w, to_packages(d))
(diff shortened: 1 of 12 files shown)


rbubley:
It seems that the unit tests in msgpack-python use a different approach to the issue of types.

See, e.g. https://github.com/lbolla/msgpack-python/commit/c8042ea99c3fa6641628e054d9a6ad6e4162b24a

mariusvniekerk:
I like the approach in that and we can implement set, tuple, frozenset which is probably good enough for now?

mrocklin:
Because I'm somewhat occupied this week the best solution might be to
submit  a PR of your own to the master branch, possibly including my
commits or not, as you like.  You could also submit a PR to my PR's branch.

mrocklin:
OK, hypothetically this passes tests.  There are some things that concern me though:

1.  We are inconsistent about our use of tuples and lists.  We use tuples everywhere *except* if the tuple contained a Serilized object, in that case we use a list
2.  We actually *have* to do this, because we only traverse through lists to find serialized objects.

I would not be surprised if this came back to bite us later.

mrocklin:
Well, lets see what happens.  Merging.

PR #2001, opened 2018-05-23
Skip click locales check

mrocklin:
Previously we raised an informative error pointing to documentation

Now we just skip the check entirely.  This will cause issues if people
try to use unicode inputs, but is otherwise safe.

See https://github.com/pallets/click/issues/448#issuecomment-246029304

cc @danielfrg can you see any major issues with this sort of hack?

### DIFF
--- a/distributed/cli/utils.py
+++ b/distributed/cli/utils.py
@@ -8,27 +8,36 @@ from distributed.comm import (parse_address, unparse_address,
 
 
 py3_err_msg = """
-Your terminal does not properly support unicode text required by command line
-utilities running Python 3.  This is commonly solved by specifying encoding
-environment variables, though exact solutions may depend on your system:
+Warning: Your terminal does not set locales.
+
+If you use unicode text inputs for command line options then this may cause
+undesired behavior.  This is rare.
+
+If you don't use unicode characters in command line options then you can safely
+ignore this message.  This is the common case.
+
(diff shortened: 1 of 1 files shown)


danielfrg:
I like it but I imagine it will generate some errors in some systems, those errors are probably going to be harder to debug but maybe they are less frequent than what you find right now with the check (i image a lot of people hit it and mention it).

In general I am +1

mrocklin:
Thank you for the response.  Can you give an example of a system where this
might cause an error?

On Wed, May 23, 2018 at 12:05 PM, Daniel Rodriguez <notifications@github.com

mariusvniekerk:
Should we rather warn and continue?  I've found this useful to know when my base container doesn't have locales set up properly

AlexSchuy:
For what it's worth, this would be great for us. We haven't been able to get the dask scripts that use click to work since moving to python 3 and are a bit flummoxed.

mrocklin:
It seems like the right thing to do is to modify our message from before
and keep it as a warning rather than an exception.

I'm busy for the next week and might not get to this.  But if anyone wants
to take this over they would be welcome.  I think that it would make a
number of people quite happy.

mrocklin:
This now warns with what I hope is an informative but-not-too-scary message

mrocklin:
Merging in 24 hours if there are no further comments.

PR #2011, opened 2018-05-28
Use timeouts during `Client._ensure_connected`

martindurant:


distributed/tests/test_client.py
```diff
@@ -5288,6 +5288,22 @@ def test_client_timeout_2():
...
 
 
+@gen_test()
+def test_client_active_bad_port():
+    import subprocess
+    import shlex
+    if PY3:
+        p = subprocess.Popen(shlex.split('python -m http.server 8000'))
+    else:
+        p = subprocess.Popen(shlex.split('python -m SimpleHTTPServer 8000'))
```
mrocklin: I'm curious, is it possible to do this from within python rather than shelling out?  Perhaps `from http.server import HTTPServer` and then something following?

Alternatively you could also take this as an opportunity to learn how to set up a Tornado web server, which is probably a good investment and would integrate more smoothly into the test suite.

http://www.tornadoweb.org/en/stable/httpserver.html#http-server

martindurant: I'm *sure* it's possible, but there at the least I would have to either not use `get_test` in order to handle the loop directly, or perhaps just try the same thing in a separate (normal) thread.

mrocklin: You can get the loop directly by accessing the global singleton

```python
from tornado.ioloop import IOLoop
io_loop = IOLoop.current()
```

However this shouldn't be necessary.  All tornado code will get the current IOLoop by default.  You only need to manually specify an IOLoop if you are doing things in separate threads, which we shouldn't need to do here.

martindurant:
Failures:
- `test_errors[joblib]`, is this a joblib version thing?
- `test_lock_sync` (py2 only), no idea

Neither seems to be related to changes here.

distributed/tests/test_client.py
```diff
@@ -5288,6 +5288,19 @@ def test_client_timeout_2():
...
+def test_client_active_bad_port():
+    import tornado.web
+    application = tornado.web.Application([
+        (r"/", tornado.web.RequestHandler),
+    ])
+    application.listen(8000)
+    with dask.config.set({'distributed.comm.timeouts.connect': '10ms'}):
+        c = Client('127.0.0.1:8000', asynchronous=True)
+        with pytest.raises((TimeoutError, IOError)):
+            yield c
```
mrocklin: The other tests are failing because there is a lingering client.  If you want these tests to pass you can either close this client (`yield c.close(fast=True)`) or you can wait for https://github.com/dask/distributed/pull/2012 , which should solve this problem generally.

mrocklin:
Thanks @martindurant !

PR #2013, opened 2018-05-29
Replace deprecated workers.iloc with workers.values()

grantjenks:
Replace deprecated workers.iloc with workers.values().

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -14,7 +14,7 @@ import random
 import six
 
 import psutil
-from sortedcontainers import SortedSet, SortedDict
+import sortedcontainers
 try:
     from cytoolz import frequencies, merge, pluck, merge_sorted, first
 except ImportError:
@@ -836,7 +836,7 @@ class Scheduler(ServerNode):
         self.clients['fire-and-forget'] = ClientState('fire-and-forget')
 
         # Worker state
-        self.workers = SortedDict()
+        self.workers = sortedcontainers.SortedDict()
         for old_attr, new_attr, wrap in [
                 ('ncores', 'ncores', None),
                 ('worker_bytes', 'nbytes', None),
@@ -852,7 +852,7 @@ class Scheduler(ServerNode):
             setattr(self, old_attr,
                     _StateLegacyMapping(self.workers, func))
 
-        self.idle = SortedSet(key=operator.attrgetter('address'))
+        self.idle = sortedcontainers.SortedSet(key=operator.attrgetter('address'))
         self.saturated = set()
 
         self.total_ncores = 0
@@ -3147,7 +3147,7 @@ class Scheduler(ServerNode):
                 worker = min(self.workers.values(),
                              key=operator.attrgetter('occupancy'))
             else:  # dumb but fast in large case
-                worker = self.workers[self.workers.iloc[self.n_tasks % len(self.workers)]]
+                worker = self.workers.values()[self.n_tasks % len(self.workers)]
 
         if self.validate:
             assert worker is None or isinstance(worker, WorkerState), (type(worker), worker)


grantjenks:
@mrocklin Just a heads-up that v2.0.3 of Sorted Containers will deploy today and emit a warning for ``sorted_dict.iloc``

I've replaced the ``iloc`` attribute with ``scheduler.workers_values`` in this pull request to eliminate the warning.

Some the CI test failures look flaky. But there's one relevant failure where ``min(self.workers_values, ...)`` raises an exception because the iterable is empty. I think the same problem existed before with ``iloc``.

distributed/scheduler.py
```diff
@@ -837,6 +837,7 @@ def __init__(
 
         # Worker state
         self.workers = SortedDict()
+        self.workers_values = self.workers.values()
```
mrocklin: Why this change?  Is this expensive?  More expensive than converting it to a set?

mrocklin:
Thank you for doing this @grantjenks .  We have been having a few intermittent failures in our tests recently, though neither of these seem familiar.  I can take a closer look at this next week.

Regarding the warning, does it warn on every use of `iloc` or just on first use?  I ask because we call this potentially thousands of times per second, which might become problematic.

grantjenks:
``.iloc`` is now a property. It warns on every access. In my testing, I only saw it emitted only once per line of code where it appeared. Is that what you mean?

Regarding the added ``.workers_values`` attribute: a call to ``.values()`` creates a `SortedValuesView` which requires a memory allocation. By caching a reference, you avoid that memory allocation. That's exactly what ``.iloc`` did but now you hold the reference rather than `SortedDict` holding the reference. The sorted dict values view does not behave as a Set. It behaves as a Sequence. There's one place where a conversion is necessary.

grantjenks:
The `warnings` module is smarter than that. Given a program like:

```python
"warnings_test.py"
import sortedcontainers as sc

sd = sc.SortedDict(zip(range(1000), range(1000)))

sd.iloc[0]

for i in range(1000):
    sd.iloc[i]
```

If you run python with no options then nothing is printed:

    >>> python warnings_test.py

But if you run with "-Wd" then two warnings are printed:

```
$ python -Wd warnings_test.py 
warnings_test.py:5: DeprecationWarning: sorted_dict.iloc is deprecated. Use SortedDict.keys() instead.
  sd.iloc[0]
warnings_test.py:8: DeprecationWarning: sorted_dict.iloc is deprecated. Use SortedDict.keys() instead.
  sd.iloc[i]
```

And if you run with "-Wa" then 1,001 warnings are printed:

```
$ python -Wa warnings_test.py 
warnings_test.py:5: DeprecationWarning: sorted_dict.iloc is deprecated. Use SortedDict.keys() instead.
  sd.iloc[0]
warnings_test.py:8: DeprecationWarning: sorted_dict.iloc is deprecated. Use SortedDict.keys() instead.
  sd.iloc[i]
... x999 more warnings ...
```

The options are documented at: https://docs.python.org/3/library/warnings.html#the-warnings-filter

mrocklin:
Ah.  I'm very glad to hear it!

mrocklin:
The Python 2 failure here seems to be genuine: 

```python
(Pdb) pp self.workers
SortedDict({'tcp://127.0.0.1:36125': <Worker 'tcp://127.0.0.1:36125', memory: 0, processing: 2>, 'tcp://127.0.0.1:38015': <Worker 'tcp://127.0.0.1:38015', memory: 0, processing: 1>})
(Pdb) pp self.workers_values
[]
(Pdb) pp self.workers.values()
[<Worker 'tcp://127.0.0.1:36125', memory: 0, processing: 2>,
 <Worker 'tcp://127.0.0.1:38015', memory: 0, processing: 1>]
```

Any thoughts on how this might occur?

mrocklin:
I got that with the following diff

```diff
mrocklin@carbon:~/workspace/distributed$ git diff
diff --git a/distributed/scheduler.py b/distributed/scheduler.py
index 3859075..f02f90e 100644
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3137,6 +3137,8 @@ class Scheduler(ServerNode):
             else:  # dumb but fast in large case
                 worker = self.idle[self.n_tasks % len(self.idle)]
         else:
+            if not self.workers_values:
+                import pdb; pdb.set_trace()
             if len(self.workers) < 20:  # smart but linear in small case
                 worker = min(self.workers_values,
                              key=operator.attrgetter('occupancy'))
```

And then running this test

    py.test distributed/bokeh/tests/test_components.py::test_profile_plot

grantjenks:
I can't reproduce locally. How do I run the tests? I keep getting:

```
...
  File "/Users/grantj/repos/distributed/distributed/__init__.py", line 3, in <module>
    from . import config
  File "/Users/grantj/repos/distributed/distributed/config.py", line 13, in <module>
    config = dask.config.config
AttributeError: 'module' object has no attribute 'config'
```
 whenever I use Python 2. (No problem with Python 3.)

If I were to guess at what was causing the error then I would say the version of sortedcontainers that you're using looks like version 1. In Version 1, the ".values()" call would've returned an empty list and set "workers_values" in the initializer. The telltale sign is the type of these values:

```
(Pdb) pp self.workers_values  # <-- Stale, set in Schedule.__init__
[]
(Pdb) pp self.workers.values()  # <-- Should not return a list.
[<Worker 'tcp://127.0.0.1:36125', memory: 0, processing: 2>,
 <Worker 'tcp://127.0.0.1:38015', memory: 0, processing: 1>]
```

Those are both Python lists...  but they shouldn't be. In Python 2, the semantics of ".values()" is to return a list but in Python 3, the semantics are to return a view. Rather than lists, you should see something like:

```
>>> import sortedcontainers
>>> sd = sortedcontainers.SortedDict({'a': 0, 'b': 1})
>>> sd.values()
SortedValuesView(SortedDict({'a': 0, 'b': 1}))
```

In version 2, sortedcontainers adopts the Python 3 semantics. Is there some way you'd be getting version 1 of sortedcontainers?

grantjenks:
One more thought, are you using conda? Maybe the issue is there and conda is still shipping an older version of sortedcontainers. Looks like that would certainly be the case on win-32: https://anaconda.org/conda-forge/sortedcontainers

mrocklin:
The testing setup is defined here: https://github.com/dask/distributed/blob/master/continuous_integration/travis/install.sh

To test you'll also need the master version of `dask/dask`, available here: git@github.com:dask/dask.git

To me it sounds like we might want to build in some version flexibility for dask's use of sortedcontainers so that things work smoothly with older versions back to a certain point.  We do this for most core dependencies (numpy, tornado, pandas, ...) because it's sometimes hard to force users to newer versions quickly due to other libraries that might depend on them, but don't update as quickly.

grantjenks:
Re: older versions of sortedcontainers -- That's fair. I'll update the pull request.

I still can't reproduce locally. I tried:

```
$ export PYTHON=2
$ source continuous_integration/travis/install.sh 
... installs miniconda with dependencies ...
$ python --version
Python 2.7.15 :: Anaconda, Inc.
$ python -m pytest distributed/bokeh/tests/test_components.py::test_profile_plot
= test session starts =
platform darwin -- Python 2.7.15, pytest-3.1.3, py-1.5.3, pluggy-0.4.0
rootdir: /Users/grantj/repos/distributed, inifile:
plugins: timeout-1.2.1, repeat-0.4.1, faulthandler-1.5.0
collected 1 item 

distributed/bokeh/tests/test_components.py .

= 1 passed in 1.15 seconds =
```

Do I need to clone `dask` and put it on my Python path?

mrocklin:
Hrm, interesting.  I've updated my test environment and indeed this problem went away.  

Unfortunately I nuked my old testing environment and so now am having difficulties reproducing.

distributed/scheduler.py
```diff
@@ -655,6 +655,12 @@ def _task_key_or_none(task):
     return task.key if task is not None else None
 
 
+if sc.__build__ < 0x020000:  # for backwards compatibility
```
mrocklin: I'm curious about the definition of `__build__` here.  Would it make sense to check something like `sc.__version__ < '2.0.0'`, which seems more common?

grantjenks: It would be more readable but strings are flawed for these kinds of comparisons. I wouldn't want version "11.0.0" to be less than "2.0.0".

The `__build__` attribute makes the same guarantees as [`sys.hexversion`](https://docs.python.org/3.6/library/sys.html#sys.hexversion) in the standard library: 1) version encoded as a single integer; 2) guaranteed to increase with each version.

For prior art on naming, see the `requests` module: https://github.com/requests/requests/blob/master/requests/__version__.py#L9

jhamman: How about using `LooseVersion`?

```Python
from distutils.version import LooseVersion
LooseVersion(sc.__version__) < LooseVersion("2.0.0"):
```
https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

grantjenks: `LooseVersion` is a good idea. Thanks for the tip! I updated the code.

mrocklin:
Sorry for the long delay in response here.  Work things have gotten in the way.

I'm looking now at the benefits of keeping SortedValues around vs the code complexity cost of having an extra values object in state.  At first glance it looks as though creating the values is somewhat cheap, relative to access.

```python
In [1]: import sortedcontainers

In [2]: d = sortedcontainers.SortedDict()

In [3]: for i in range(1000):
   ...:     d[str(i)] = i
   ...:     

In [4]: %timeit d.values()
327 ns ± 14.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: v = d.values()

In [6]: %timeit v[50]
977 ns ± 35.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

So I find myself inclined to just call `.values()` frequently.  @grantjenks are there things that I might be overlooking here?

Also, I made a couple small style changes.  Do you mind if I push directly to your branch?

grantjenks:
Using `scheduler.workers_values` or `scheduler.workers.values()` is each perfectly fine. The semantics are identical between the two. The former trades a bit of memory for a bit of speed (25% faster) and is nearly exactly how `.iloc` worked previously. An earlier thread had said that performance was important in the scheduler code but if there's no benchmark then a few hundred nanoseconds will likely go unnoticed.

You are welcome to push directly to my branch. I'm not sure if I need to do something to permit that.

mrocklin:
> Using scheduler.workers_values or scheduler.workers.values() is each perfectly fine. The semantics are identical between the two. The former trades a bit of memory for a bit of speed (25% faster) and is nearly exactly how .iloc worked previously. An earlier thread had said that performance was important in the scheduler code but if there's no benchmark then a few hundred nanoseconds will likely go unnoticed.

OK, I've reverted things to create the view each time.  Our budget is something like 100us per task, so I figure we live with the 300ns cost for now until it becomes an issue.

I'm going to go ahead and merge this.  Thank you so much for helping out here @grantjenks .  It'll be nice to have you in the authors list.

grantjenks:
Glad to help out and have another user!

PR #2014, opened 2018-05-29
Avoid reference cycle in joblib backend

mrocklin:
Previously we attached an instance method to the instance

    future.get = future.result

This caused a reference cycle.
Now we use weakrefs to avoid this cycle.

cc @ogrisel @gaelvaroquaux


GaelVaroquaux:
Your linter is failing because you have a left-over debug import.

PR #2020, opened 2018-05-31
Automatically scatter large arguments in joblib connector

ogrisel:
This requires: https://github.com/joblib/joblib/pull/689 to work efficiently but should not crash or degrade performance when used with an older joblib.

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -83,12 +83,7 @@ class _WeakKeyDictionary:
             # as soon as the object used as key is garbage collected.
             def on_destroy(_):
                 del self._data[key]
-            try:
-                ref = weakref.ref(obj, on_destroy)
-            except TypeError:
-                # Some interned objects such as str cannot have a weakref:
-                # store the object directly
-                ref = obj
+            ref = weakref.ref(obj, on_destroy)
         self._data[key] = ref, value
 
     def __len__(self):
@@ -181,7 +176,6 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
         # The explicit call to clear is required to break a cycling reference
         # to the futures.
         self.call_data_futures.clear()
-        self.call_data_futures = None
 
     def effective_n_jobs(self, n_jobs):
         return sum(self.client.ncores().values())
@@ -205,13 +199,11 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
                 if f is None and arg_id in self.data_futures:
                     f = self.data_futures[arg_id]
 
-                if f is None and call_data_futures is not None:
+                elif f is None and call_data_futures is not None:
                     try:
                         f = call_data_futures[arg]
                     except KeyError:
-                        if (call_data_futures is not None
-                                and is_weakrefable(arg)
-                                and sizeof(arg) > 1e6):
+                        if is_weakrefable(arg) and sizeof(arg) > 1e6:
                             # Automatically scatter large objects to some of
                             # the workers to avoid duplicated data transfers.
                             # Rely on automated inter-worker data stealing if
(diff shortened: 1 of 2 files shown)


ogrisel:
@TomAugspurger if you have a chance it would be great to update the docker image of your pangeo instance to test nested parallelism with this new feature (this requires upgrading the joblib branch too).

ogrisel:
I ran some tests on @TomAugspurger's pangeo instance with 30 workers  (randomized search for a random forests) and everything looks fine: the CPUs get saturated and there is very little white space in the task stream view of the diagnostics page of the dask scheduler.

ogrisel:
I also checked that all the futures are garbage collected by the end of the grid search: `client.futures` is empty.

distributed/joblib.py
```diff
@@ -40,6 +41,63 @@
...
 
+def is_weakrefable(obj):
+    try:
+        weakref.ref(obj)
+        return True
+    except TypeError:
+        return False
+
+
+class _WeakKeyDictionary:
```
jcrist: Any reason not to subclass from [MutableMapping](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping), which would turn this into a full `dict` interface?

distributed/joblib.py
```diff
@@ -116,47 +174,80 @@ def get_nested_backend(self):
...
+                arg_id = id(arg)
+                if arg_id in itemgetters:
+                    yield itemgetters[arg_id]
+                    continue
+
+                f = None
+                if f is None and arg_id in self.data_futures:
+                    f = self.data_futures[arg_id]
+
+                if f is None and call_data_futures is not None:
```
jcrist: Should this be `elif f is None...` instead?

distributed/joblib.py
```diff
@@ -116,47 +174,80 @@ def get_nested_backend(self):
...
+                    try:
+                        f = call_data_futures[arg]
+                    except KeyError:
+                        if (call_data_futures is not None
```
jcrist: `call_data_futures` will always be not `None` due to the `if` check above, this check is not needed.

distributed/joblib.py
```diff
@@ -116,47 +174,80 @@ def get_nested_backend(self):
...
+                                and is_weakrefable(arg)
```
jcrist: IIUC, ` _WeakKeyDictionary` can handle non-weakrefable objects, so this check is not needed?

distributed/joblib.py
```diff
@@ -116,47 +174,80 @@ def get_nested_backend(self):
...
         return self.effective_n_jobs(n_jobs)
 
+    def start_call(self):
+        self.call_data_futures = _WeakKeyDictionary()
+
+    def stop_call(self):
+        # The explicit call to clear is required to break a cycling reference
+        # to the futures.
+        self.call_data_futures.clear()
+        self.call_data_futures = None
```
jcrist: Why set this to `None`? If `call_data_futures` always exists and is just cleared after running, then you never need to check if it's `None` below.

distributed/joblib.py
ogrisel: I would need to implement `__delitem__` (easy) and `__iter__` not as simple but we have no use for those. So I don't think it's worth it.

distributed/joblib.py
ogrisel: Actually, it's an oversight on my side when implementing the `_WeakKeyDictionary`. I don't think it's possible to safely handle non-weakrefable objects like dict instances: if the id of a recently collected dict argument is reused we will compute the wrong thing. I will make the `_WeakKeyDictionary` refuse to accept non-weakrefable keys.

distributed/joblib.py
ogrisel: Alright.

ogrisel:
@jcrist I think I addressed your comments. Let's see of the CI likes it.

ogrisel:
I think the travis failure on `test_administration` is unrelated to this PR.

mrocklin:
Yeah, it looks like a couple of intermittent failures have crept in.
Several PRs are failing today.

TomAugspurger:
CI seems green now.

mrocklin:
I think @jcrist gave a +1 in person.  Merging this now.

PR #2030, opened 2018-06-04
Better auto-scatter tests for the joblib connector

ogrisel:
Improve the tests for the new joblib auto-scatter feature to better test details of its behavior.

I ran the tests locally against joblib master and it works as expected.


mrocklin:
Thanks for following up @ogrisel

Issue #2034, opened 2018-06-07
KeyError in scheduler for task that's submitted from within a joblib.Parallel call


TomAugspurger:
I'm seeing a key error in the distributed scheduler, where a tasks is apparently forgotten while it has dependents.

code:

<details>

```python
import dask  # noqa
import distributed
import logging
from dask_ml.datasets import make_classification
from dask_ml.wrappers import Incremental
from dask_ml.metrics import get_scorer
from sklearn.linear_model import SGDClassifier

import sklearn.model_selection
import dask_ml.model_selection  # noqa
import dask_ml.joblib  # noqa
from sklearn.externals import joblib
import numpy as np  # noqa

from distributed import Client

handler = logging.FileHandler("bug.log", mode="w")
FORMAT = '[%(asctime)-15s] %(filename)s - %(lineno)d - %(message)s'
formatter = logging.Formatter(FORMAT)
handler.setLevel(logging.INFO)
handler.setFormatter(formatter)

loggers = [
    logging.getLogger(__name__),
    logging.getLogger('distributed.client'),
    logging.getLogger('distributed.core'),
    logging.getLogger('distributed.scheduler'),
]
for logger in loggers:
    logger.addHandler(handler)
    logger.setLevel(logging.INFO)

logger = loggers[0]

# KeyError: "('astype-lt-getitem-fe947b8291fa5d0276b0b63d123fa1f0', 1)"


def main():
    logger.info('#' * 80)
    logger.info("Starting")
    logger.info("#" * 80)
    cluster = distributed.LocalCluster(n_workers=2, processes=False)
    client = Client(cluster)  # noqa
    dask.config.set(**{'distributed.scheduler.transition-log-length': 1000})

    X, y = make_classification(n_samples=100_000, chunks=50_000)
    logger.info("%0.2f GBs, %s blocks", X.nbytes / 1e9, X.numblocks)

    clf = SGDClassifier(random_state=0)
    scorer = get_scorer('accuracy', compute=True)
    inc = Incremental(clf, scoring='accuracy')
    param_grid = {
        'alpha': [0.1, 10],
    }

    gs1 = sklearn.model_selection.GridSearchCV(inc, param_grid, iid=False,
                                               refit=False, verbose=2, cv=2,
                                               scoring=scorer,
                                               n_jobs=-1)

    with joblib.parallel_backend('dask'):
        gs1.fit(X, y, classes=[0, 1])
    print("Random... OK")
    client.close()


[...]

PR #2036, opened 2018-06-07
Only avoid recomputation if key exists

mrocklin:
Fixes https://github.com/dask/distributed/issues/2034

This does not yet include a test

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1355,8 +1355,9 @@ class Scheduler(ServerNode):
                 ts = self.tasks[key]
                 for dep in dependencies[key]:
                     if all(d in done for d in dependents[dep]):
-                        done.add(dep)
-                        stack.append(dep)
+                        if dep in self.tasks:
+                            done.add(dep)
+                            stack.append(dep)
 
             for d in done:
                 del tasks[d]


mrocklin:
The fact that this was necessary is odd.  This code goes through the submitted tasks, and finds those that don't need to be computed because either they are already in memory, or because the tasks that they are here for, are also already in memory.

```python
        if already_in_memory:
            dependents = dask.core.reverse_dict(dependencies)
            stack = list(already_in_memory)
            done = set(already_in_memory)
            while stack:  # remove unnecessary dependencies
                key = stack.pop()
                ts = self.tasks[key]
                for dep in dependencies[key]:
                    if all(d in done for d in dependents[dep]):
                        if dep in self.tasks:
                            done.add(dep)
                            stack.append(dep)

            for d in done:
                del tasks[d]
                del dependencies[d]
```

The change we made is to short-circuit this process if all of the dependents are already done, but the task itself is not yet known.  This is odd.  How can a task's dependents be done if that task itself is not known?  There might be something up with how the dask-joblib backend is naming tasks that creates collisions?

mrocklin:
@TomAugspurger any feedback on this?

TomAugspurger:
Nothing insightful to add. A comment explaining the check, with a link back here may be helpful for future maintainers.

I've verified that it avoids the exception I was seeing.

PR #2042, opened 2018-06-08
Rely on dask implementation of sizeof

mrocklin:
Previously we copied these implementations between dask/dask and
dask/distributed.  Now we depend explicitly on the dask/dask
implementation.

Depends on https://github.com/dask/dask/pull/3582


jakirkham:
Happy to see this code moving into Dask proper.

PR #2052, opened 2018-06-12
Have worker data transfer wait until recipient acknowledges

mrocklin:
Previously when a worker sent data it would think it was finished as soon as the data was dumped to the socket.  Now we wait until we hear an acknowledgement from the recipient that the transfer is complete.  This helps with our diagnostics a bit *may* help avoid backing up a bunch of memory on the OS level, and also assists in future GPU work, where the sending side wants to wait until deserialization on the recipient side has finished.

### DIFF
--- a/distributed/tests/test_utils_comm.py
+++ b/distributed/tests/test_utils_comm.py
@@ -1,5 +1,6 @@
 from __future__ import print_function, division, absolute_import
 
+import pytest
 
 from distributed.core import rpc
 from distributed.utils_test import gen_cluster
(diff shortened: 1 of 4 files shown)


seibert:
This could also be useful for shared memory data exchange on the CPU (i.e. via Plasma or other mechanisms) in scenarios where having multiple worker processes on the CPU is necessary.  That's still pretty hypothetical, but the GPU use case is really a special case of shared memory communication more broadly.

sklam:
@mrocklin, it's not clear to me how to use this with custom serialization.  My GPU dataframe serialization would send the data as a IPC handle and would need to wait for the deserializer to finish using the IPC handle.

mrocklin:
> My GPU dataframe serialization would send the data as a IPC handle and would need to wait for the deserializer to finish using the IPC handle.

Right, all this PR does is guarantee to you that the data will stay in scope on the sender's side for as long as your deserialization code has not returned.

mrocklin:
`human`

mrocklin:
Is this enough for your needs @sklam ?

sklam:
I thk so.  I will be integrating this in a branch of dask_gdf in the next few days.  The dask_gdf data transfer is my current top priority task.

sklam:
I have success with this patch in dask_gdf.

PR #2054, opened 2018-06-12
Provide communication context to serialization functions

mrocklin:
This allows comms to provide information about the communication, for example the address of the receiving end, to serialization functions.

This is primarily motivated by the support of side-channel communication for GPUs, but has also come up when we've wanted to know if we're writing to disk or sending along a comm.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -216,7 +216,9 @@ class TCP(Comm):
 
         frames = yield to_frames(msg,
                                  serializers=serializers,
-                                 on_error=on_error)
+                                 on_error=on_error,
+                                 context={'sender': self._local_addr,
+                                          'recipient': self._peer_addr})
 
         try:
             lengths = ([struct.pack('Q', len(frames))] +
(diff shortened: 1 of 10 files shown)


sklam:
I'm trying this but I can't get the `context` into my serializers.  It looks like my type-based serializers are invoked by `dask_dumps` which doesn't pass along the `context`.  Custom serialization family is also not getting the serializers from `Client(serializers=...)`

mrocklin:
I've added a context= keyword to dask_dumps, though it's entirely untested at the moment, so don't expect much.

Can you say more about how the serialization family approach isn't working?  Are you registering the serialization family with the register_serialization_family function?  Are you providing the name to Client(serializers=)?  What are you checking when you say this sentence?

> Custom serialization family is also not getting the serializers from Client(serializers=...)

mrocklin:
Also, thanks for being a guinea pig here.

sklam:
I registered and provided the name to `Client(serializers=['gdf', 'dask', 'pickle')` and didn't see it get used.  I put a `pdb.set_trace()` inside `dask_dumps` and walk up the stack.  I saw `serializers` is `None` throughout as if the `Client` didn't propagate that info.

mrocklin:
Was the latest change effective @sklam ?

sklam:
Yes, I'm able to use `register_serialization`.

sklam:
I have success with this patch in dask_gdf.

PR #2057, opened 2018-06-13
Set `gen_cluster`s `Client` as the default one

jakirkham:
Needed for PR ( https://github.com/dask/dask/pull/3561 ).


distributed/utils_test.py
```diff
@@ -779,7 +779,8 @@ def coro():
                             args = [s] + workers
                             if client:
                                 c = yield Client(s.address, loop=loop, security=security,
-                                                 asynchronous=True, **client_kwargs)
+                                                 set_as_default=True, asynchronous=True,
```
mrocklin: I may not understand.  This is true by default.

jakirkham:
You're right. Think something was off with my local dev environment as CI worked fine without it.

PR #2063, opened 2018-06-17
Allow bokeh port to be overridden in _repr_html_

### LINKED ISSUE #2062: Bokeh port seems to be hardcoded to 8787 in _repr_html_
My bokeh status page is running 63220 however the Client html repr uses a port of 8787.

It's not immediately obvious in the below code where the 8787 is coming from and I haven't had time to investigate properly though I assume it's from some default config?
https://github.com/dask/distributed/blob/3715fd128eaa5dc6d027789470a5966b8d880253/distributed/client.py#L691-L700

dhirschfeld:
Fixes #2062

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -522,6 +522,7 @@ class Client(Node):
             timeout = parse_timedelta(timeout, 's')
         self._timeout = timeout
 
+        self._local_config = dict()
         self.futures = dict()
         self.refcount = defaultdict(lambda: 0)
         self.coroutines = []
@@ -536,8 +537,6 @@ class Client(Node):
         self._startup_kwargs = kwargs
         self.cluster = None
         self.scheduler = None
-        self.scheduler_address = None
-        self.bokeh_port = None
         self._scheduler_identity = {}
         self._lock = threading.Lock()
         self._refcount_lock = threading.Lock()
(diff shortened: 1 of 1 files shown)


dhirschfeld:
I'm not sure if this is the best fix, or even a good fix but it solves the problem I identified in #2062. If there's a better way to accomplish that I'm happy to make any changes required...

mrocklin:
I'm disinclined to tack on attributes to the client for relatively infrequent special cases like this.  I think that this is bad for maintenance long term.  That said, I can see situations in which this would be valuable.  I suggest two optoins:

1.  Try using the global config and see if that satisfies your short-term needs.  The following should work as long as you aren't managing multiple clusters within the same process:

        dask.config.set({'distributed.dashboard.link': 'http://{host}:your_port/status'})

2.  If that doesn't work, then we include a local config dictionary on the client, which gets added into the global config whenever it does things

    ```python
    class Client:
        def __init__(...):
            self._local_config = dict()
    ```

    This probably provides the kind of customization that you want, but in a way that is a bit more flexible and avoids the proliferation of a thousand tiny attributes over time.

But I recommend seeing if 1 solves your needs operationally first if possible.

dhirschfeld:
Going forward I will definitely want to have more than one dask cluster in the same process so whilst 1. will work in the interim, I will want some way of configuring this on a per-client basis.

I had thought of implementing the generic config method in 2. but figured I'd implement the minimum I required and get your input first.

> *I'm disinclined to tack on attributes to the client for relatively infrequent special cases like this*

Agreed, it's a bit of an ugly hack. Maybe not the `scheduler_address` attribute since it pretty much defines the `Client` so could deserve a top-level attribute but the `bokeh_port` was a bit whiffy and begs the question of where you stop adding random top-level attributes to the class.

dhirschfeld:
There's no rush on this - for some other reasons as well, I've gone with a custom subclass and overridden the `_repr_html_` there. I still think it would be a useful feature to be able to override the global config though.

distributed/client.py
```diff
@@ -680,24 +682,29 @@ def _repr_html_(self):
...
+                urlparts = urlparse(scheduler.address)
+                if urlparts.scheme == 'inproc':
+                    host = 'localhost'
+                else:
+                    host = urlparts.hostname
+                port = info['services']['bokeh']
+                template = dask.config.get('distributed.dashboard.link')
+                ns = os.environ.copy()
+                ns['host'] = host
+                ns['port'] = port
```
dhirschfeld: Since there are multiple ports in a distributed deployment perhaps this would be better renamed to `bokeh_port` to allow future usage of the `_local_config` without fear of name clashes?

mrocklin:

PR #2064, opened 2018-06-19
Allow adaptive to exist without a cluster

mrocklin:
This allows the Adaptive object to exist on the scheduler and make
recommendations with an external route without being attached to an
explicit cluster object.

For motivation see conversation in https://github.com/dask/dask-yarn/issues/1


mrocklin:
Tests will fail here.  This is up here for discussion in dask-yarn for the moment.

Issue #2058, opened 2018-06-14
KeyError in distributed joblib


TomAugspurger:
I haven't been able to reproduce the locally yet.

This is on distributed, dask, & dask-ml master, and the scikit-learn / joblib we used at the sprint.

I'm doing

```python
%%time
with joblib.parallel_backend("dask"):
    gs.fit(X, y, classes=[0, 1])
```

with X and y being dask arrays (so can't pre-scatter).

The behavior I observe is

1. Make the call, tasks show up in the dashboard
2. A short time later, the tasks go black / gray, indicating they failed
3. The notebook is hanging
4. ctrl-C to interrupt (the keyboard interrupt you see in the exception below)

```
tornado.application - ERROR - Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x7fc37ad6fbf8>, <Future finished exception=AssertionError("yield from wasn't used with future",)>)
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/distributed/joblib.py", line 204, in maybe_to_futures
    f = call_data_futures[arg]
  File "/opt/conda/lib/python3.6/site-packages/distributed/joblib.py", line 67, in __getitem__
    ref, val = self._data[id(obj)]
KeyError: 140477560921584

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/tornado/ioloop.py", line 759, in _run_callback
    ret = callback()
  File "/opt/conda/lib/python3.6/site-packages/tornado/stack_context.py", line 276, in null_wrapper
    return fn(*args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/tornado/ioloop.py", line 780, in _discard_future_result
    future.result()
  File "/opt/conda/lib/python3.6/site-packages/tornado/gen.py", line 1113, in run
    yielded = self.gen.send(value)
  File "/opt/conda/lib/python3.6/site-packages/distributed/joblib.py", line 244, in callback_wrapper
    callback(result)  # gets called in separate thread
  File "/home/jovyan/src/scikit-learn/sklearn/externals/joblib/parallel.py", line 326, in __call__
    self.parallel.dispatch_next()
  File "/home/jovyan/src/scikit-learn/sklearn/externals/joblib/parallel.py", line 746, in dispatch_next
    if not self.dispatch_one_batch(self._original_iterator):
[...]

PR #2066, opened 2018-06-20
BUG: Normalize address before comparison

TomAugspurger:
Closes https://github.com/dask/distributed/issues/2058

### DIFF
--- a/distributed/tests/test_worker_client.py
+++ b/distributed/tests/test_worker_client.py
@@ -297,3 +297,17 @@ def test_worker_client_rejoins(c, s, a, b):
 
     result = yield c.submit(f)
     assert result
+
+
+@gen_cluster(client=True)
+def test_submit_different_names(c, s, a, b):
+    # https://github.com/dask/distributed/issues/2058
+    da = pytest.importorskip('dask.array')
+    c = yield Client('localhost:' + s.address.split(":")[-1], loop=s.loop,
+                     asynchronous=True)
+    with c:
+        X = c.persist(da.random.uniform(size=(100, 10), chunks=50))
+        yield wait(X)
+
+        fut = yield c.submit(lambda x: x.sum().compute(), X)
+        assert fut > 0
--- a/distributed/tests/test_worker_client.py
+++ b/distributed/tests/test_worker_client.py
@@ -299,15 +299,17 @@ def test_worker_client_rejoins(c, s, a, b):
     assert result
 
 
-@gen_cluster(client=True)
-def test_submit_different_names(c, s, a, b):
+@gen_cluster()
+def test_submit_different_names(s, a, b):
     # https://github.com/dask/distributed/issues/2058
     da = pytest.importorskip('dask.array')
     c = yield Client('localhost:' + s.address.split(":")[-1], loop=s.loop,
                      asynchronous=True)
-    with c:
+    try:
         X = c.persist(da.random.uniform(size=(100, 10), chunks=50))
         yield wait(X)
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
cc @mrocklin 

The test is good, but I'm not sure if this is the appropriate fix or not. My idea was to always normalize user-provided `address`es. `resolve_address` seemed to fit the bill.

If you like the fix, then I'll go through `worker.py` and `client.py` to always resolve the address on user-defined input.

mrocklin:
I don't have strong thoughts about how this could backfire.  @pitrou is the expert here.  I'm inclined to go ahead with it if the test suite doesn't complain.

TomAugspurger:
I notice that other places accepting an `address` also take a `resolve_address` keyword argument. e.g. `Scheduler.add_worker`. I'll plan on following that approach.

mrocklin:
The windows failure is interesting.  I haven't seen it before and *may* be related?  I'm not sure.

distributed/tests/test_worker_client.py
```diff
@@ -297,3 +297,17 @@ def f():
...
+    # https://github.com/dask/distributed/issues/2058
+    da = pytest.importorskip('dask.array')
+    c = yield Client('localhost:' + s.address.split(":")[-1], loop=s.loop,
+                     asynchronous=True)
+    with c:
+        X = c.persist(da.random.uniform(size=(100, 10), chunks=50))
+        yield wait(X)
+
+        fut = yield c.submit(lambda x: x.sum().compute(), X)
+        assert fut > 0
```
mrocklin: I recommend avoiding the use of the normal context manager in async tests.  If you were to use `async with c` that would work well, but we can't use that within tests that have to be within python 2 friendly files.  Instead I recommend using try/finally.  

Also, if you're not going to use the client created in the gen_cluster decorator then you probably want the following:

```python
@gen_cluster()
def test_foo(s, a, b):
    c = yield Client(...)

    try:
        ...
    finally:
        yield c.close()
```

There are also some examples of this in test_client.py

TomAugspurger: OK, thanks. I was just using the context manager so that it was cleaned up.

mrocklin: Understood, but the context manager is synchronous API.  There is no way for us to `yield c.close()` within the normal context mananger.  I think that it just starts the close coroutine, but doesn't wait for it.  You'll either need to use `async with` or explicitly call `c.close()` to be sure that it closes up.

TomAugspurger:
Agreed that the windows test could be related. I'll see if it fails this time around.

TomAugspurger:
Windows passed this time.

The Python 2 travis run failed: https://travis-ci.org/dask/distributed/jobs/394751774#L1637

distributed/tests/test_worker_client.py
```diff
@@ -297,3 +297,19 @@ def f():
...
+
+@gen_cluster()
+def test_submit_different_names(s, a, b):
+    # https://github.com/dask/distributed/issues/2058
+    da = pytest.importorskip('dask.array')
+    c = yield Client('localhost:' + s.address.split(":")[-1], loop=s.loop,
+                     asynchronous=True)
+    try:
+        X = c.persist(da.random.uniform(size=(100, 10), chunks=50))
+        yield wait(X)
```
mrocklin: Fine for now, but in the future I recommend `foo.persist()` over `c.persist(foo)`

mrocklin: Just for aesthetic reasons really

PR #2071, opened 2018-06-24
Add test for as_completed for loops in Python 2

mrocklin:


distributed/compatibility.py
```diff
@@ -71,6 +73,7 @@ def iscoroutinefunction(func):
     from gzip import compress as gzip_compress
     ConnectionRefusedError = ConnectionRefusedError
     FileExistsError = FileExistsError
+    StopAsyncIteration = StopAsyncIteration
```
jakirkham: This seems to be causing a test failure in Dask on Python 3.4 as `StopAsyncIteration` was not defined until Python 3.5.

ref: https://travis-ci.org/dask/dask/jobs/396392846
ref: https://www.python.org/dev/peps/pep-0492/

mrocklin: Dask.distributed doesn't support Python 3.4.  We might consider solving this on the dask side by avoiding `distributed` tests somehow

PR #2072, opened 2018-06-25
Support TB and PB in format bytes

mrocklin:


martindurant:
Nice to see that this is now required :)

PR #2088, opened 2018-07-02
Update example for stopping a worker

jakirkham:
The `remove_worker` syntax is outdated. Update it to use `stop_worker`, which is the current syntax.

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -70,7 +70,7 @@ class LocalCluster(Cluster):
 
     Shut down the extra worker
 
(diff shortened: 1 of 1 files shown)


mrocklin:
For consistency with the newer cluster interface we might actually want to use `scale_down`

jakirkham:
IMHO both methods are valuable. If one wants to launch just one more worker, `scale_up`/`scale_down` is not the right path. One still is better of using `start_worker`/`stop_worker`.

mrocklin:
OK

PR #2092, opened 2018-07-03
Allow worker to refuse data requests with busy signal

mrocklin:
This allows workers to say "I'm too busy right now" when presented with
a request for data from another worker.  That worker then waits a bit,
queries the scheduler to see if anyone else has that data, and then
tries again.  The wait time is an exponential backoff.

Pragmatically this means that when single pieces of data are in high
demand that the cluster will informally do a tree scattering.  Some workers
will get the data directly while others wait on the busy signal.  Then other
workers will get from them, etc..  We used to ask users to do this explicitly
with the following:

    client.replicate(future)
    or
    client.scatter(data, broadcast=True)

And now the replicate/broadcast step is no longer strictly necessary. (though
some scattering of local data still is).

Machines on the same host are given some preference, and so should be able to
sneak in more easily.

Currently this has two issues:

1.  We need to unify the configuration with the total_connections parameter
    (which does the same thing, but in the opposite direction)
2.  We don't test the same-host behavior (this is hard because we're currently
    getting host information from the socket.)

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -19,7 +19,9 @@ distributed:
   worker:
     multiprocessing-method: forkserver
     use-file-locking: True
-    max-connections: 10     # maximum simultaneous outgoing connections
+    connections:            # Maximum concurrent connections for data
+      outgoing: 50          # This helps to control network saturation
+      incoming: 10
 
     profile:
       interval: 10ms        # Time between statistical profiling queries
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -19,6 +19,7 @@ distributed:
   worker:
     multiprocessing-method: forkserver
     use-file-locking: True
+    max-connections: 10     # maximum simultaneous outgoing connections
 
     profile:
       interval: 10ms        # Time between statistical profiling queries
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -564,7 +564,7 @@ def test_clean_nbytes(c, s, a, b):
 
 @gen_cluster(client=True, ncores=[('127.0.0.1', 1)] * 20)
 def test_gather_many_small(c, s, a, *workers):
-    a.total_connections = 2
+    a.total_out_connections = 2
     futures = yield c._scatter(list(range(100)))
 
     assert all(w.data for w in workers)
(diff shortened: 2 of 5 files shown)


mrocklin:
cc @ogrisel @seibert

distributed/worker.py
```diff
@@ -1381,28 +1398,32 @@ def transition_dep_waiting_flight(self, dep, worker=None):
                 pdb.set_trace()
             raise
 
-    def transition_dep_flight_waiting(self, dep, worker=None):
+    def transition_dep_flight_waiting(self, dep, worker=None, busy=False):
```
mrocklin: This keyword should be inverted and changed to something like `remove`, which probably makes more sense locally

ogrisel:
> Machines on the same host 

you mean workers on the same host?

distributed/tests/test_worker.py
```diff
@@ -1180,3 +1180,19 @@ def test_wait_for_outgoing(c, s, a, b):
...
+    yield wait(x)
+
+    with dask.config.set({'distributed.worker.max-connections': 1}):
+        futures = [c.submit(len, x, pure=False, workers=[w.address])
+                   for w in workers[1:]]
+
+        yield wait(futures)
+
+    assert len(workers[0].outgoing_transfer_log) < 18
+    assert sum(not not w.outgoing_transfer_log for w in workers) >= 3
```
ogrisel: `not not` is weird. wouldn't `bool(w.outgoing_transfer_log)` work?

ogrisel: or `sum(1 for w in workers if len(w.outgoing_transfer_log) > 0)` that might be even more explicit.

distributed/tests/test_worker.py
```diff
@@ -1180,3 +1180,19 @@ def test_wait_for_outgoing(c, s, a, b):
...
+    x = c.submit(np.random.random, 1000000, workers=[workers[0].address])
+    yield wait(x)
+
+    with dask.config.set({'distributed.worker.max-connections': 1}):
+        futures = [c.submit(len, x, pure=False, workers=[w.address])
+                   for w in workers[1:]]
+
+        yield wait(futures)
+
+    assert len(workers[0].outgoing_transfer_log) < 18
```
ogrisel: Where does the 18 and the 3 come from? Where they measured from empirical runs?

distributed/distributed.yaml
```diff
@@ -19,6 +19,7 @@ distributed:
   worker:
     multiprocessing-method: forkserver
     use-file-locking: True
+    max-connections: 10     # maximum simultaneous outgoing connections
```
ogrisel: Did you have the opportunity to run some IO intensive benchmark/stress test on a "real" cluster (e.g. on GCP) to measure the impact of that setting on the overall completion time of a data bottlenecked set of tasks?

If you do I would be curious to see the empirical arity of the tree structure of the resulting broadcasting for different values of max-connections.

distributed/tests/test_worker.py
mrocklin: Two issues:

1.  We may start to run out of RAM on travis with 1e8 * 20 bytes
2.  Compression will make the transfers too fast

I'm not very concerned about the cost of creating the random array the first time.  I don't think that this will affect the number of secondary worker-to-worker transfers.  However I may not fully understand your meaning.

distributed/distributed.yaml
mrocklin: No, I haven't yet done any benchmarking.  I'll play a bit on my local machine.  I may get to trying this out on a larger cluster, but that's uncertain.  Instead, I suspect that we will end up changing the default value over time.

distributed/tests/test_worker.py
mrocklin: Fixed to be more explicit.  I'm using `len(... for ... if ...)`

mrocklin:
The behavior here is as I would expect.  I'm comfortable with this from a performance perspective, though I also think that we'll end up wanting to tune this default in the future by a factor of 2-3.  

There is still some administrative cleanup to do I think.

ogrisel:
Thanks for the benchmarks, that seems to work fine :)

PR #2094, opened 2018-07-04
Adjust worker doc after change in config file location and treatment

apatlpo:


docs/source/worker.rst
```diff
@@ -156,16 +156,17 @@ Workers use a few different policies to keep memory use beneath this limit:
...
    # Set any of the lower three values to False to turn off the behavior entirely
-   worker-memory-target: 0.60  # target fraction to stay below
-   worker-memory-spill: 0.70  # fraction at which we spill to disk
-   worker-memory-pause: 0.80  # fraction at which we pause worker threads
-   worker-memory-terminate: 0.95  # fraction at which we terminate the worker
+   memory:
+     target: 0.60  # target fraction to stay below
+     spill: 0.70  # fraction at which we spill to disk
+     pause: 0.80  # fraction at which we pause worker threads
+     terminate: 0.95  # fraction at which we terminate the worker
```
mrocklin: Thanks for doing this!  One request: 

Can I ask that we add the full tree here to give users context on where these changes should go:

```yaml
distributed:
  worker:
    memory:
      target: ...
```

apatlpo: done

mrocklin:
Can I ask you to make the same change in the dask/dask PR?  At some point we should probably deprecate one of these documents for the other to avoid duplication.

apatlpo:
sure, done

PR #2096, opened 2018-07-05
removed hardcoded value of memory terminate fraction from a log message

bm371613:


mrocklin:
This looks good to me.  Thanks @bm371613 !

PR #2100, opened 2018-07-07
Cull in graph_to_futures

mrocklin:
Currently dask.delayed objects aren't culled prior to sending up to the
scheduler.  This can result in too much communication.

This adds a cull step in the Client, but this should maybe be handled instead
by the collections.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2094,6 +2094,7 @@ class Client(Node):
                       and k not in keyset}
             if values:
                 dsk = dask.optimization.inline(dsk, keys=values,
+                                               inline_constants=False,
                                                dependencies=dependencies)
 
             d = {k: unpack_remotedata(v) for k, v in dsk.items()}
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -25,7 +25,7 @@ import weakref
 
 import dask
 from dask.base import tokenize, normalize_token, collections_to_dsk
-from dask.core import flatten, get_dependencies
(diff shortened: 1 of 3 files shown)


mrocklin:
This adds five microseconds per task for what I believe to be a relatively infrequently useful thing.  (for context, recall that our budget is currently around 200-300us) Not sure that this is worth it.

```python
In [1]: import dask.array as da

In [2]: x = da.random.random((100000, 100000), chunks=(1000, 1000))

In [3]: y = x + x.T - x.mean(axis=0)

In [4]: from dask.optimization import cull

In [5]: dsk = dict(y.dask)

In [6]: %time _ = cull(dsk, y.__dask_keys__())
CPU times: user 287 ms, sys: 3.84 ms, total: 291 ms
Wall time: 290 ms

In [7]: from dask.core import get_dependencies

In [8]: %time _ = {k: get_dependencies(dsk, k) for kk in y.__dask_keys__() for k
   ...:  in kk}
CPU times: user 48.4 ms, sys: 0 ns, total: 48.4 ms
Wall time: 47.9 ms

In [9]: 0.290 / len(dsk) * 1e6  # microseconds per task
Out[9]: 5.420560747663551
```

It might would be nicer to handle this in `delayed.__dask_optimize__`

jakirkham:
IIUC `dask.delayed` has started culling. ( https://github.com/dask/dask/pull/3856 ) So this can be closed.

PR #2101, opened 2018-07-07
Pre-compute dependencies for inline

mrocklin:
This moves around computation a bit in graph_to_futures to avoid
computing dependencies twice.


jsignell:
Looks like this just needs another merge and then it's good to go. Is that the case?

PR #2114, opened 2018-07-12
Joblib backend improvements

ogrisel:
@TomAugspurger I wonder if the change in the threshold of the size to trigger the auto-scattering could fix the case we observed together at scipy.

If so I would love to get your confirmation that using 1 kB instead of 1 MB as the threshold is a good idea.

### DIFF
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -196,7 +196,7 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
                     continue
 
                 f = None
-                if f is None and arg_id in self.data_futures:
+                if arg_id in self.data_futures:
                     f = self.data_futures[arg_id]
 
                 elif f is None and call_data_futures is not None:
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -204,12 +204,13 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
                         f = call_data_futures[arg]
                     except KeyError:
                         if is_weakrefable(arg) and sizeof(arg) > 1e3:
-                            # Automatically scatter large objects to some of
-                            # the workers to avoid duplicated data transfers.
-                            # Rely on automated inter-worker data stealing if
-                            # more workers need to reuse this data concurrently
-                            # beyond the initial broadcast arity.
-                            [f] = self.client.scatter([arg], broadcast=3)
+                            # Automatically scatter large objects to one of the
+                            # workers to avoid duplicated data transfers. Rely
+                            # on automated inter-worker data stealing if more
+                            # workers need to reuse this data concurrently as
+                            # well as the ability of a worker to refuse data
+                            # requests with busy signal.
+                            [f] = self.client.scatter([arg], broadcast=1)
                             call_data_futures[arg] = f
 
                 if f is not None:
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -210,7 +210,7 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
                             # workers need to reuse this data concurrently as
                             # well as the ability of a worker to refuse data
                             # requests with busy signal.
-                            [f] = self.client.scatter([arg], broadcast=1)
+                            [f] = self.client.scatter([arg])
                             call_data_futures[arg] = f
 
                 if f is not None:
--- a/distributed/joblib.py
+++ b/distributed/joblib.py
@@ -203,7 +203,7 @@ class DaskDistributedBackend(ParallelBackendBase, AutoBatchingMixin):
                     try:
                         f = call_data_futures[arg]
                     except KeyError:
-                        if is_weakrefable(arg) and sizeof(arg) > 1e6:
+                        if is_weakrefable(arg) and sizeof(arg) > 1e3:
                             # Automatically scatter large objects to some of
                             # the workers to avoid duplicated data transfers.
                             # Rely on automated inter-worker data stealing if


distributed/joblib.py
```diff
@@ -196,20 +196,21 @@ def maybe_to_futures(args):
...
-                            # beyond the initial broadcast arity.
-                            [f] = self.client.scatter([arg], broadcast=3)
+                        if is_weakrefable(arg) and sizeof(arg) > 1e3:
+                            # Automatically scatter large objects to one of the
+                            # workers to avoid duplicated data transfers. Rely
+                            # on automated inter-worker data stealing if more
+                            # workers need to reuse this data concurrently as
+                            # well as the ability of a worker to refuse data
+                            # requests with busy signal.
+                            [f] = self.client.scatter([arg], broadcast=1)
```
mrocklin: You can just remove the `broadcast=` keyword here.

ogrisel: Explicit is better than implicit, no?

mrocklin: If you like.  No strong thoughts.  One argument against is we tend not to explicitly state defaults.  Being fully explicit and listing all keywords would be verbose.  There is history here though so listing this keyword might be a good idea.  Again, no strong thoughts.

TomAugspurger:
This seems to cause a deadlock on the script we were playing with

```pytb
~/sandbox/distributed/distributed/joblib.py in _to_func_args(self, func)
    223         tasks = []
    224         for f, args, kwargs in func.items:
--> 225             args = list(maybe_to_futures(args))
    226             kwargs = dict(zip(kwargs.keys(), maybe_to_futures(kwargs.values())))
    227             tasks.append((f, args, kwargs))

~/sandbox/distributed/distributed/joblib.py in maybe_to_futures(args)
    211                             # well as the ability of a worker to refuse data
    212                             # requests with busy signal.
--> 213                             [f] = self.client.scatter([arg], broadcast=1)
    214                             call_data_futures[arg] = f
    215 

~/sandbox/distributed/distributed/client.py in scatter(self, data, workers, broadcast, direct, hash, maxsize, timeout, asynchronous)
   1771                              broadcast=broadcast, direct=direct,
   1772                              local_worker=local_worker, timeout=timeout,
-> 1773                              asynchronous=asynchronous, hash=hash)
   1774 
   1775     @gen.coroutine

~/sandbox/distributed/distributed/client.py in sync(self, func, *args, **kwargs)
    651             return future
    652         else:
--> 653             return sync(self.loop, func, *args, **kwargs)
    654 
    655     def __repr__(self):

~/sandbox/distributed/distributed/utils.py in sync(loop, func, *args, **kwargs)
    268     else:
    269         while not e.is_set():
--> 270             e.wait(10)
    271     if error[0]:
    272         six.reraise(*error[0])

/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py in wait(self, timeout)
    549             signaled = self._flag
    550             if not signaled:
--> 551                 signaled = self._cond.wait(timeout)
    552             return signaled
    553 

/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py in wait(self, timeout)
    297             else:
    298                 if timeout > 0:
--> 299                     gotit = waiter.acquire(True, timeout)
    300                 else:
    301                     gotit = waiter.acquire(False)

KeyboardInterrupt: 
```

when running

```python
from urllib.parse import urlparse
import webbrowser
from sklearn.datasets import make_classification
from sklearn.ensemble import RandomForestClassifier
from sklearn import model_selection
from sklearn.externals import joblib


def open_dashboard(client):
    info = client.scheduler_info()
    base = info['address']
    port = info['services']['bokeh']

    # there's probably already a way to get this URL.
    url = "http://{}:{}/status".format(urlparse(base).netloc.split(":")[0],
                                       port)
    webbrowser.open(url)


def main():
    client = Client()
    open_dashboard(client)

    X, y = make_classification(
        n_samples=1000,
        random_state=0,
        n_informative=5,
        class_sep=0.5,
    )

    params = {
        'n_estimators': [10, 50, 100],
        'max_depth': [2, 5, None],
        'max_features': [0.1, 0.25, 0.5, None],
        'min_samples_leaf': [1, 2, 3, 5],
    }

    rf = RandomForestClassifier(random_state=0)
    search = model_selection.RandomizedSearchCV(
        rf, params, n_iter=100, cv=5, n_jobs=-1,
        return_train_score=False,
        refit=False,
        iid=False, random_state=0
    )

    with joblib.parallel_backend('dask'):
        search.fit(X, y)


if __name__ == '__main__':
    main()
```

TomAugspurger:
Seeing some KeyErrors in the traceback like

```pytb
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 204, in maybe_to_futures
    f = call_data_futures[arg]
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 67, in __getitem__
    ref, val = self._data[id(obj)]
KeyError: 4536700080

  File "distributed-joblib.py", line 54, in <module>
  File "distributed-joblib.py", line 50, in main

  File "/Users/taugspurger/sandbox/scikit-learn/sklearn/model_selection/_search.py", line 661, in fit
    cv.split(X, y, groups)))
  File "/Users/taugspurger/sandbox/scikit-learn/sklearn/externals/joblib/parallel.py", line 952, in __call__
    while self.dispatch_one_batch(iterator):
  File "/Users/taugspurger/sandbox/scikit-learn/sklearn/externals/joblib/parallel.py", line 791, in dispatch_one_batch
    self._dispatch(tasks)
  File "/Users/taugspurger/sandbox/scikit-learn/sklearn/externals/joblib/parallel.py", line 748, in _dispatch
    job = self._backend.apply_async(batch, callback=cb)
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 235, in apply_async
    func, args = self._to_func_args(func)
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 225, in _to_func_args
    args = list(maybe_to_futures(args))
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 213, in maybe_to_futures
    [f] = self.client.scatter([arg], broadcast=1)
  File "/Users/taugspurger/sandbox/distributed/distributed/client.py", line 1773, in scatter
    asynchronous=asynchronous, hash=hash)
  File "/Users/taugspurger/sandbox/distributed/distributed/client.py", line 653, in sync
  File "/Users/taugspurger/sandbox/distributed/distributed/utils.py", line 270, in sync
    e.wait(10)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 551, in wait
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 299, in wait
KeyboardInterrupt
tornado.application - ERROR - Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x10e782598>, <Future finished exception=AssertionError("yield from wasn't used with future",)>)
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 204, in maybe_to_futures
    f = call_data_futures[arg]
  File "/Users/taugspurger/sandbox/distributed/distributed/joblib.py", line 67, in __getitem__
    ref, val = self._data[id(obj)]
KeyError: 4535866832
```

ogrisel:
@TomAugspurger I found the cause of the problem we observed during the sprint. To fix it, we need to wrap the nested call inside a `with parallel_backend('dask')`. I find it easier to do this change directly in the joblib code base. I think I would like to deprecate the joblib connector copy in the distributed package once this is done.

Here is the PR: https://github.com/joblib/joblib/pull/722 (I still need to move over the tests there).

ogrisel:
The only private API we use from dask.distributed is the `_wait` function. I wonder if we could rewrite that part to only use public API from distributed.

mrocklin:
You should be able to use the public wait in an async context.  Everything
should be fine if you replace _wait with wait.

ogrisel:
I tried to replace `_wait` by `wait` and that freezes the tests.

mrocklin:
Interesting.  That means that `wait` thinks its being called in a blocking context, rather than in a coroutine/async-def function.  Usually it's pretty good about detecting one over the other.

We could add an `asynchronous=` keyword to the `wait` function to force async operation and use that.  This would imply a dependence on the next version of `distributed`.  We could also just do

```python
result = yield future
```

But this will cause a data transfer, which we do again later for the same future, so we would need to merge these two.

The `_wait` function is also semi-public.  It was underscored not because it was private, but due to a previous convention that all blocking functions had non-blocking variants that were prefixed by underscores (an unfortunate overlap of conventions).  Today we just trust that the client can figure out if it is in an async or blocking situation.

ogrisel:
I am fine with adding a dependency on the next version of distributed as soon as it's ready and use the private `_wait` in the mean time.

Issue #2110, opened 2018-07-12
Serialize data within tasks


mrocklin:
Currently if we submit data within a task like the following:

```python
x = np.array([1, 2, 3])
future = client.submit(func, x)
```

Then we construct a task like `(func, x)` and then call pickle on this task.  *We don't do custom serialization once we contruct tasks*.  This is mostly because stuff in tasks is rarely large, and traversing tasks can increase our overhead in the common case.  Typically we encourage people to use dask.delayed or something similar to mark data

```python
x = np.array([1, 2, 3])
x = dask.delayed(x)
future = client.submit(func, x)
```

But this is error prone and requires rarely held expertise.  

We might again consider traversing arguments.

dhirschfeld:
**Update:**
Scattering *does* work, but only if you ensure the custom serialization is imported on all workers.

```python
obj_fut = client.scatter(obj)
```
...will fail with the below error if `distributed.protocol.arrow` isn't *explicitly* imported.
```python-traceback
  File "C:\dev\src\distributed\distributed\protocol\serialize.py", line 157, in serialize
    raise TypeError(msg)
TypeError: Could not serialize object of type RecordBatch
```
If you import `distributed.protocol.arrow` in the client process but not in the workers it fails with the below error:
```python-traceback
  File "c:\dev\src\distributed\distributed\core.py", line 448, in send_recv
    raise Exception(response['text'])
Exception: Serialization for type pyarrow.lib.RecordBatch not found
```

...so to get it to actually work I need to run:
```python
def init_arrow():
    from distributed.protocol import arrow
    return None

init_arrow()
client.run(init_arrow)
obj_fut = client.scatter(obj)
fut = client.submit(echo, obj_fut)
result = fut.result()
assert obj.equals(result)
```

mrocklin:
Yes, something like preload is probably the right way to handle this today, assuming that it's not already in library code.  

Eventually it would be nice to allow clients to register functions to be run at worker start time with the scheduler that could be passed to workers as they start up.

1.  Client registers `func` with the scheduler as a preload operation
2.  Scheduler holds on to `func`
3.  Scheduler also sends `func` to all workers to have them run `func`
4.  Workers run `func`
5.  New worker arrives, tells Scheduler that it exists
6.  As part of saying "Hi" the scheduler also gives the worker `func` and tells it to run it

[...]

PR #2115, opened 2018-07-13
 Add custom serialization support for pyarrow 

dhirschfeld:


dhirschfeld:
Tests incoming - just wanted to get this up here for consideration in the meantime

dhirschfeld:
**c.c.** @wesm, @xhochy - this may be of interest and you're certainly better qualified than me to comment on the implementation!

mrocklin:
> c.c. @wesm, @xhochy - this may be of interest and you're certainly better qualified than me to comment on the implementation!

To be explicit, the objective would be to get something like a `bytes`, `buffer`, or ideally `memoryview` object out as cheaply as possible.  A list of such objects would also be welcome if data isn't contiguous.

mrocklin:
So if I were to do this for a pandas dataframe I would probably pull the `.data` attribute from each block in the block manager separately, and then put a lot of metadata into the header.

mrocklin:
So the solution where you scatter explicitly should work fine, but when working with coroutines you'll need to yield the scatter result.  This shouldn't be an issue if you use the normal API.

The rest of the tests here are related to https://github.com/dask/distributed/issues/2110 .  It's also worth noting that this shouldn't be an issue for worker-to-worker transfers.  So if your tasks generate and then pass around arrow objects then things will also be fine with this change.  This is only an issue when you push data into Dask.

xhochy:
In future, we will support pickling these objects. The return values of `__reduce__` should be the necessary buffers packed into nested tuples. Would that be sufficient for dask or do you need to have a flat list of buffers?

dhirschfeld:
***xref:***
 - https://issues.apache.org/jira/browse/ARROW-1715
 - https://issues.apache.org/jira/browse/ARROW-2660 (https://github.com/apache/arrow/pull/2161)

dhirschfeld:
I think this is good to go now - all the tests, including the scatter pass ro me locally:

image

The other `Client.submit` tests have been removed as adding that functionality is being discussed in #2110.

Until `arrow` supports pickling itself this allows passing around `arrow` data-structures with `dask` so I think is useful new functionality.

If I've implemented it correctly I'd hope that it might be more efficient than pickling, at least until [PEP 574](https://www.python.org/dev/peps/pep-0574/) is passed and https://github.com/apache/arrow/pull/2161 is merged.

wesm:
If you don't mind giving me a chance to review I'll have a closer look tomorrow

distributed/protocol/arrow.py
```diff
@@ -0,0 +1,53 @@
...
+
+def serialize_batch(batch):
+    import pyarrow as pa
+    sink = pa.BufferOutputStream()
+    writer = pa.RecordBatchStreamWriter(sink, batch.schema)
+    writer.write_batch(batch)
+    writer.close()
+    buf = sink.get_result()
+    header = {}
+    frames = [buf.to_pybytes()]
```
wesm: This causes an extra memory copy. Can `frames` contain objects exporting the buffer protocol?

distributed/protocol/arrow.py
```diff
@@ -0,0 +1,53 @@
...
+    return header, frames
+
+
+def deserialize_batch(header, frames):
+    import pyarrow as pa
+    blob = frames[0]
+    reader = pa.RecordBatchStreamReader(pa.BufferReader(blob))
```
wesm: I opened ARROW-2859 to see if we can get rid of this `pa.BufferReader` detail

distributed/protocol/arrow.py
```diff
@@ -0,0 +1,53 @@
...
+
+def serialize_table(tbl):
+    import pyarrow as pa
+    sink = pa.BufferOutputStream()
+    writer = pa.RecordBatchStreamWriter(sink, tbl.schema)
+    writer.write_table(tbl)
+    writer.close()
+    buf = sink.get_result()
+    header = {}
+    frames = [buf.to_pybytes()]
```
wesm: ditto

distributed/protocol/arrow.py
mrocklin: That's a good question.  It looks like it can with small modification (pushed).  Presumably this means that PyArrow also works with sockets and Tornado IOStreams.  Hooray for consistent use of protocols.

wesm:
Awesome. Is there something we can do to make the Python API for Buffer more conforming?

distributed/protocol/arrow.py
```diff
@@ -0,0 +1,53 @@
...
+
+from .serialize import register_serialization
+
+
+def serialize_batch(batch):
+    import pyarrow as pa
+    sink = pa.BufferOutputStream()
+    writer = pa.RecordBatchStreamWriter(sink, batch.schema)
+    writer.write_batch(batch)
+    writer.close()
```
dhirschfeld: One improvement on the arrow side would be if `RecordBatchStreamWriter` was a context manager as that would avoid the need for an explicit close.

wesm: A fair point. ARROW-2863

mrocklin:
Merging this in a few hours if there are no further comments.

mrocklin:
> Awesome. Is there something we can do to make the Python API for Buffer more conforming?

No, this was due to ugliness on the Dask side.

mrocklin:
Thanks @dhirschfeld !  Merged.

dhirschfeld:
Thanks @mrocklin, @wesm & @xhochy - I'm excited about getting this support into dask! :)

PR #2116, opened 2018-07-13
Use type hints to further emphasize the custom serialization api

dhirschfeld:


mrocklin:
It's interesting to see types used in documentation like this.  I wouldn't have thought of doing it myself.  This looks good to me.  Merging.

(also, my apologies about the intermittent failures.  I hope to find time to clean up that test soonish.)

Issue #2111, opened 2018-07-12
Make it easier to capture task stream information


mrocklin:
People often want to get the data on the task stream in a structured format.  This is possible today with some pain.  Here is an example:

```python

In [1]: from dask.distributed import Client

In [2]: client = Client()

In [3]: # open dashboard to ensure that the scheduler captures information

In [4]: import dask.array as da

In [5]: x = da.random.random((10, 10), chunks=(5, 5))

In [6]: y = (x + x.T).persist()

In [7]: def f(dask_scheduler):
   ...:     from distributed.bokeh.task_stream import TaskStreamPlugin
   ...:     for plugin in dask_scheduler.plugins:
   ...:         if isinstance(plugin, TaskStreamPlugin):
   ...:             return plugin.buffer
   ...:         

In [8]: client.run_on_scheduler(f)
Out[8]: 
[{'key': "('random_sample-ecda85d77df96a666d2303bc1629d816', 0, 1)",
  'nbytes': 200,
  'startstops': (('deserialize', 1531409759.1762872, 1531409759.2346888),
   ('compute', 1531409759.2363048, 1531409759.2377546)),
  'status': 'OK',
  'thread': 140000297481984,
  'type': b'\x80\x04\x95\x15\x00\x00\x00\x00\x00\x00\x00\x8c\x05numpy\x94\x8c\x07ndarray\x94\x93\x94.',
  'worker': 'tcp://127.0.0.1:41153'},
 {'key': "('transpose-9114dbb1ddcfe7f13e190d36cd1c2016', 1, 0)",
  'nbytes': 200,
  'startstops': (('compute', 1531409759.241821, 1531409759.241839),),
  'status': 'OK',
  'thread': 140000297481984,
  'type': b'\x80\x04\x95\x15\x00\x00\x00\x00\x00\x00\x00\x8c\x05numpy\x94\x8c\x07ndarray\x94\x93\x94.',
  'worker': 'tcp://127.0.0.1:41153'},
 {'key': "('random_sample-ecda85d77df96a666d2303bc1629d816', 0, 0)",
  'nbytes': 200,
  'startstops': (('deserialize', 1531409759.1766217, 1531409759.240885),
   ('compute', 1531409759.241804, 1531409759.2433474)),
  'status': 'OK',
  'thread': 140000297481984,
[...]

PR #2120, opened 2018-07-16
Update dask-scheduler cli help text for preload

nicolls1:
As per discussion on #2111

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -60,8 +60,7 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
               help="Directory to place scheduler files")
 @click.option('--preload', type=str, multiple=True, is_eager=True,
               help='Module that should be loaded by the scheduler process  '
-                   'like "foo.bar" or "/path/to/foo.py". Scheduler plugins '
-                   'should be loaded here.')
+                   'like "foo.bar" or "/path/to/foo.py".')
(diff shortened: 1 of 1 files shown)


distributed/cli/dask_scheduler.py
```diff
@@ -59,8 +59,9 @@
 @click.option('--local-directory', default='', type=str,
               help="Directory to place scheduler files")
 @click.option('--preload', type=str, multiple=True, is_eager=True,
-              help='Module that should be loaded by each worker process '
-                   'like "foo.bar" or "/path/to/foo.py"')
+              help='Module that should be loaded by the scheduler process  '
+                   'like "foo.bar" or "/path/to/foo.py". Scheduler plugins '
+                   'should be loaded here.')
```
mrocklin: Thanks for changing worker -> scheduler here. 

I'm not convinced that we should call out the specific use case of scheduler plugins though.  Preload scripts are used for a variety of reasons, of which this is only one.

mrocklin: Should we remove this?

mrocklin:
Merging.  Thanks @nicolls1 !

distributed/cli/dask_scheduler.py
nicolls1: Yeah good point, removed.

mrocklin:
Also I notice that this is your first contribution to the project.  Welcome!

PR #2123, opened 2018-07-18
[WIP] Responsive bokeh templating

canavandl:
This PR
* Replaces the Status diagnostic page with one that uses single-element bokeh layouting and a responsive template

### DIFF
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -112,6 +112,8 @@ class TaskStream(DashboardComponent):
             export.register_plot(self.root)
             self.root.add_tools(export)
 
+        self.root = row(self.root, name='task_stream', sizing_mode='scale_width')
+
         # Required for update callback
         self.task_stream_index = [0]
 
--- a/distributed/bokeh/scheduler.py
+++ b/distributed/bokeh/scheduler.py
@@ -306,7 +306,9 @@ class CurrentLoad(DashboardComponent):
             self.nbytes_figure = nbytes
 
             processing.y_range = nbytes.y_range
-            self.root = row(nbytes, processing, sizing_mode='scale_width')
+            # self.root = row(nbytes, processing, sizing_mode='scale_width', name='current_load')
+            self.root_1 = nbytes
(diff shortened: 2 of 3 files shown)


distributed/bokeh/components.py
```diff
@@ -112,6 +112,8 @@ def __init__(self, n_rectangles=1000, clear_interval='20s', **kwargs):
             export.register_plot(self.root)
             self.root.add_tools(export)
 
+        self.root = row(self.root, name='task_stream', sizing_mode='scale_width')
```
canavandl: I think the desired pattern is to put the roots inside of ``bokeh.layout.row`` layouts with a responsive ``scale_width`` sizing mode and name, so that it's addressable in a template

distributed/bokeh/scheduler.py
```diff
@@ -306,7 +306,9 @@ def __init__(self, scheduler, width=600, **kwargs):
             self.nbytes_figure = nbytes
 
             processing.y_range = nbytes.y_range
-            self.root = row(nbytes, processing, sizing_mode='scale_width')
+            # self.root = row(nbytes, processing, sizing_mode='scale_width', name='current_load')
+            self.root_1 = nbytes
```
canavandl: We do have the benefit of setting multiple roots for things that share datasources/update callbacks.

mrocklin:
> Just the status page? I haven't opened an issue yet but both tasks and graph show up too small to be useful for me (on the last released):

I believe that the goal here is to apply this to all pages.  @canavandl pushed this up as a WIP PR at my request.

dhirschfeld:
On this pr the tasks view looks better - the graph takes up the whole page. The graph view is still small though.

The navbar links point to random `keen.io` urls:
image

On the status page there's no margin on the LHS which looks a little awkward:
image

...also all pages are too high for me - I have to scroll to see all the information.

I recognise this is just a WIP pr but thought I'd post my observations in case they're helpful.

canavandl:
superceded by #2150

PR #2146, opened 2018-07-30
Cleanup

mrocklin:
This includes a variety of changes to clean up the testing process.  Going commit-by-commit is probably wise.


mrocklin:
OK, things aren't perfect, but they're definitely better.  Merging this in.

PR #2147, opened 2018-07-30
Worker class

martindurant:
Ability to set the class used as the worker within Nanny, so that Nanny can be easily subclassed with different types of worker.

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -36,7 +36,6 @@ class Nanny(ServerNode):
     """
     process = None
     status = None
-    Worker = Worker  # default class to call in WorkerProcess
 
     def __init__(self, scheduler_ip=None, scheduler_port=None,
                  scheduler_file=None, worker_port=0,
@@ -44,7 +43,8 @@ class Nanny(ServerNode):
                  name=None, memory_limit='auto', reconnect=True,
                  validate=False, quiet=False, resources=None, silence_logs=None,
                  death_timeout=None, preload=(), preload_argv=[], security=None,
-                 contact_address=None, listen_address=None, **kwargs):
+                 contact_address=None, listen_address=None, worker_class=None,
+                 **kwargs):
         if scheduler_file:
             cfg = json_load_robust(scheduler_file)
             self.scheduler_addr = cfg['address']
@@ -62,6 +62,7 @@ class Nanny(ServerNode):
         self.death_timeout = death_timeout
         self.preload = preload
         self.preload_argv = preload_argv
+        self.Worker = Worker if worker_class is None else worker_class
 
         self.contact_address = contact_address
         self.memory_terminate_fraction = dask.config.get('distributed.worker.memory.terminate')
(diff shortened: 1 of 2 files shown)


distributed/tests/test_nanny.py
```diff
@@ -145,6 +145,31 @@ def test_close_on_disconnect(s, w):
...
+
+
+class Nanny2(Nanny):
+    # a subclass with alternate Worker class to create
+    Worker = Something
+
+
+@gen_cluster(client=True, ncores=[])
+def test_nanny_worker_class(c, s):
+    w = Nanny(s.ip, s.port, ncores=1, loop=s.loop)
```
mrocklin: You'll want to add `Worker=` to the constructor here.  

Alternatively, you might consider using gen_cluster for this with 

    gen_cluster(client=True, Worker=Nanny, worker_kwargs={'Worker': Something})

This is a bit nicer because it will clean up things slightly more reliably.

martindurant: ah ,sorry

distributed/nanny.py
```diff
@@ -44,7 +43,8 @@ def __init__(self, scheduler_ip=None, scheduler_port=None,
                  name=None, memory_limit='auto', reconnect=True,
                  validate=False, quiet=False, resources=None, silence_logs=None,
                  death_timeout=None, preload=(), preload_argv=[], security=None,
-                 contact_address=None, listen_address=None, **kwargs):
+                 contact_address=None, listen_address=None, worker_class=None,
```
mrocklin: Maybe we can set `worker_class=Worker` to avoid the check later?

I'm +1 regardless though if you want to merge @martindurant

martindurant: The current line fits in better with the style of the other optional arguments, so I would keep it this way. To me, None says something like "do not worry about this parameter, unless you specifically know you need it".

mrocklin: I don't think that that's convention across the ecosystem.  I think that default values are usually used rather than None.

mrocklin: None is typically used when a default value can't be known ahead of time because it depends on other arguments passed in.  In this case we know what the default value is, `Worker`, and so it's nice to specify it explicitly.  This both reduces arg handling code later on and also provides the user a suggestion in documentation and docstrings.

martindurant: It is certainly common :) particularly in very long lists like this, where one suspects it might eventually be replaced by `**kwargs` .  Thinking https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html

mrocklin: Sure, that looks like a fine example to look at.  Is there a keyword in that list that has a known default value that isn't listed as a default?

martindurant: To be sure, this was nothing more than a gut decision, I don't mind either way.

(encoding, dialect, line_terminator, float_precision, date_parser ...)

mrocklin: Looking at encoding as an example I'm unable to find a line like 

```python
if encoding is None:
    encoding = 'foo'
```

Instead there is logic in there

```python
def _read(filepath_or_buffer, kwds):
    """Generic reader of line files."""
    encoding = kwds.get('encoding', None)
    if encoding is not None:
        encoding = re.sub('_', '-', encoding).lower()
        kwds['encoding'] = encoding
```

I suspect, but have not verified, that the same is true with the others as well.

PR #2150, opened 2018-08-01
Canavandl/responsive bokeh

canavandl:
supercedes #2123 

This PR:
* creates a base.html template with a collapsable navbar
* creates a status.html page that extends base.html and individually lays out the Bokeh elements for the status view
* creates a simple.html page that extends base.html and all non-status views use


mrocklin:
Thanks for doing this @canavandl .  I'm excited about this work. 

A few questions:

-  How can we avoid external resources?  Should we bundle these up as part of the package build process?
-  How can we better support wide screens?  Is there a way to change the layout when the screen's aspect ratio changes significantly?
-  Where should we keep custom css?  I'll probably want to futz slightly with the coloring and size of the navbar

canavandl:
> How can we avoid external resources? Should we bundle these up as part of the package build process?

Yes - I believe you could add them to the static/ dir as part of the package build and have the templates load from there. I don't know enough about the distributed packaging, so you might have to suggest how to do it.

> How can we better support wide screens? Is there a way to change the layout when the screen's aspect ratio changes significantly?

I believe was can modify the layout so that the other plots are to the right of the task stream plot for large screens. I'll try it out later

> Where should we keep custom css? I'll probably want to futz slightly with the coloring and size of the navbar

We should add a custom.css file to the static/ dir and move the inline styles there (and load them normally). I'll make this change later also.

mrocklin:
@canavandl I'm interpretting the silence here as you enjoying some time away from the computer.  I may reach out to others if that's ok.

For anyone else that comes along there are a few active questions that could use someone's attention:

1.  How can we have a layout change between tall-and-skinny and short-and-fat aspect ratios

### Narrow

status-narrow svg

### Wide

status-wide svg

2.  How should we improve our build process to include the relevant bootstrap files when we package the project, but not include them in the repository?

distributed/bokeh/components.py
```diff
@@ -112,6 +112,8 @@ def __init__(self, n_rectangles=1000, clear_interval='20s', **kwargs):
             export.register_plot(self.root)
             self.root.add_tools(export)
 
+        self.root = row(self.root, name='task_stream', sizing_mode='scale_width')
```
mrocklin: @canavandl what was the motivation for this?  Is this just so that we can provide a name?  Is there a nicer way to do this on the original figure?

mrocklin:
Superceded by #2213 

Thanks again @canavandl .  This was fun to work on as a group.

Issue #2156, opened 2018-08-02
Investigate worker overhead


mrocklin:
Motivated by a desire for reduced latencies on the workers for [Actors](https://github.com/dask/distributed/pull/2133) (we found that 1ms things were taking 5ms) we added a thread that [statistically profiles the event loop](https://github.com/dask/distributed/pull/2144).  This showed overhead from a couple surprising sources:

1.  `psutil` and the `SystemMonitor`
2.  Tornado's `write_to_fd` which apparently isn't entirely non-blocking
3.  Tornado's `add_callback` overhead, see [this stack overflow question](https://stackoverflow.com/questions/51582394/which-functions-are-free-when-profiling-tornado-asyncio/51595426?noredirect=1#comment90212561_51595426)

I'm not sure how best to address these.  There are probably a few approaches:

1.  Check that we're using psutil appropriately, and that there isn't some better way to regularly poll system use at high-ish frequency (currently we poll every 500ms)
2.  Quantify the cause of `add_callback`, and see if there aren't some occasions where we can reduce our use of Tornado
3.  Investigate other concurrency frameworks, like asyncio + uvloop.  This sounds neat, but is likely expensive for many reasons.  I did try using uvloop + asyncio + tornado but it wasn't very effective.  The overhead appears to be higher in this stack so that uvloop doesn't seem to do much good.

mrocklin:
I can reduce the overhead of using `add_callback` by patching tornado to use asyncio.call_soon rather than `asyncio.call_soon_threadsafe` (see https://github.com/tornadoweb/tornado/issues/2463) and then annotating some uses of add_callback from within the event loop as being safe.

Waiting on that tornado issue to learn if there is a better way forward.

<details>

```diff
diff --git a/distributed/worker.py b/distributed/worker.py
index 48e4bb8..4a73b7c 100644
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -338,7 +338,7 @@ class WorkerBase(ServerNode):
         self.batched_stream = BatchedSend(interval='2ms', loop=self.loop)
         self.batched_stream.start(comm)
         self.periodic_callbacks['heartbeat'].start()
-        self.loop.add_callback(self.handle_scheduler, comm)
+        self.loop.add_callback(self.handle_scheduler, comm, threadsafe=True)
 
     @gen.coroutine
     def handle_scheduler(self, comm):
@@ -351,7 +351,8 @@ class WorkerBase(ServerNode):
         finally:
[...]

PR #2165, opened 2018-08-07
Add asyncio.streams Comm

mrocklin:
This only works with Python 3.7. I had to pull out and modify some asyncio code.  It also makes no attempt at TLS.

### DIFF
--- a/distributed/comm/_asyncio_utils.py
+++ b/distributed/comm/_asyncio_utils.py
@@ -1,0 +1,169 @@
+"""
+Asyncio.start_server is a coroutine, which stops ``TCPListener.start`` from
+being a normal function.  It turns out that the only reason it is a coroutine
+is to get a bit of information asynchronously::
+
+    async def _ensure_resolved(self, address, *,
+                               family=0, type=socket.SOCK_STREAM,
+                               proto=0, flags=0, loop):
+        host, port = address[:2]
+        info = _ipaddr_info(host, port, family, type, proto)
+        if info is not None:
+            # "host" is already a resolved IP.
+            return [info]
+        else:
+            return await loop.getaddrinfo(host, port, family=family, type=type,
+                                          proto=proto, flags=flags)
+
+
+We replace this with socket.getaddrinfo and make the thing return immediately.
+We also remove all sockets that aren't SOCK_STREAM type.
+"""
+
+
+
+import socket
+
+from asyncio.streams import _DEFAULT_LIMIT
+from asyncio.base_events import Server
+from asyncio import events, coroutines, protocols
+import os
+import sys
+import itertools
+
+def create_server(self, protocol_factory, host=None, port=None,
+                  *,
+                  family=socket.AF_UNSPEC,
+                  flags=socket.AI_PASSIVE,
+                  sock=None,
+                  backlog=100,
+                  ssl=None,
+                  reuse_address=None,
+                  reuse_port=None):
+    """Create a TCP server.
+
+    The host parameter can be a string, in that case the TCP server is bound
+    to host and port.
+
+    The host parameter can also be a sequence of strings and in that case
+    the TCP server is bound to all hosts of the sequence. If a host
+    appears multiple times (possibly indirectly e.g. when hostnames
+    resolve to the same IP address), the server is only bound once to that
+    host.
+
+    Return a Server object which can be used to stop the service.
+
+    This method is a coroutine.
+    """
+    if isinstance(ssl, bool):
+        raise TypeError('ssl argument must be an SSLContext or None')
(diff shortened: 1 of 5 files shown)


distributed/comm/_asyncio_utils.py
```diff
@@ -0,0 +1,186 @@
...
+        elif (isinstance(host, str) or
+              not isinstance(host, collections.abc.Iterable)):
+            hosts = [host]
+        else:
+            hosts = host
+
+        infos = [socket.getaddrinfo(host, port, family=family, flags=flags) for
+                 host in hosts]
+        infos = set(itertools.chain.from_iterable(infos))
+        infos = [info for info in infos if info[1] == socket.SocketKind.SOCK_STREAM]
```
mrocklin: I changed these three lines

mrocklin:
This does get us down to about a 600us roundtrip time, down from 800us, which is a bit of a win.

```python
   ...: client = Client()
   ...: client
   ...: 
   ...: 
Out[1]: <Client: scheduler='tcp://127.0.0.1:35289' processes=4 cores=4>

In [2]: async def f():
   ...:     for i in range(10000):
   ...:         await client.scheduler.identity()
   ...: %time client.sync(f)
   ...: 
   ...: 
CPU times: user 5.8 s, sys: 467 ms, total: 6.27 s
Wall time: 6.31 s
```

mrocklin:
Oooh, I can get this down to 470us if 

1.  I bundle all of the writes into one for small messages (this is definitely doable)
2.  I engage https://github.com/tornadoweb/tornado/issues/2463

At this point about 70% of scheduler time is spent in socket.send (see https://stackoverflow.com/questions/51731690/why-does-asyncio-spend-time-in-socket-senddata)

mrocklin:
OK, this could use review by someone who knows this stack better than I do.

This currently implements a functional `Comm` using asyncio streams.  This seems to perform a bit better when doing many small fast messages.  I haven't yet checked with larger messages (though that should be interesting).

There are a few challenges:

1.  asyncio's `start_server` function is asynchronous, which means that `Listener.start` would need to be asynchronous, which means that all of the Dask `Server.listen` methods would need to become asynchronous (they are currently not asynchronous) which would be a bit of a pain.  

    Currently my solution to this is to copy over and patch a couple functions in `asyncio` (see `distributed/comm/_asyncio_utils.py`), but this is fragile and only works for specific versions (the `Server` class changed a bit between Python 3.6 and 3.7).  

2.  I haven't figured out how to do IPV6.  I suspect that this is a simple matter of passing a different flag when creating a connections or starting a server.  

3.  Currently we fail `distributed/comm/tests/test_comms.py::test_tcp_comm_closed_implicit`.  When the server severs our connection we don't feel it on the client side.  

@pitrou if you have time to look things over here I would appreciate it.

mrocklin:
Current tests on the CI systems aren't that meaningful.  This only works in Python 3.7 due to copying over and modifying asyncio code.

mrocklin:
Oh, and I'm also somewhat blocked on this problem: https://stackoverflow.com/questions/51731690/why-does-asyncio-spend-time-in-socket-senddata

A very simple benchmark that sends a small message back and forth currently gets up to spending about 70% of its time in `socket.send`.

pitrou:
I also suggest you run a benchmark with large messages, because `asyncio` doesn't have all the optimizations that we added to Tornado.

pitrou:
> A very simple benchmark that sends a small message back and forth currently gets up to spending about 70% of its time in socket.send.

Just for the record, which profiler are you using? `cProfile` or a sampling profiler? Do you have other threads going on in the background?

mrocklin:
> Just for the record, which profiler are you using? cProfile or a sampling profiler? Do you have other threads going on in the background?

Sampling profiler.  The same one that we use for worker threads.  See https://github.com/dask/distributed/pull/2144

> I also suggest you run a benchmark with large messages, because asyncio doesn't have all the optimizations that we added to Tornado.

I agree.  My short-term plan is to include both, but have tornado be default.  I'm dealing with some workloads now that need relatively low-latency communication of many small messages.  Using this and *many* other tricks I can get round-trip message latency down to about 450us (see https://github.com/dask/distributed/issues/2156)

> And how much send calls is that per second or, another way to phrase it, what is the average duration of a send call?

That's a good question.  I'll find out.

distributed/comm/_asyncio_utils.py
```diff
@@ -0,0 +1,186 @@
+"""
```
mrocklin, line 1: This stuff is probably no longer necessary.  Since doing this work we've made Listeners optionaly awaitable, so we should be able to handle things appropriately on the Dask side.

PR #2168, opened 2018-08-08
Fix memory tooltip in workers dashboard

lesteve:
screenshot_20180808_151728

I used the same names as the CPU tooltip a few lines below (which works fine):
https://github.com/dask/distributed/blob/b16ee25506fd20ec5daa7d77e338e2725e778562/distributed/bokeh/scheduler.py#L991-L998

I double-checked that it was working too.


mrocklin:
Thanks @lesteve !

Issue #2117, opened 2018-07-14
Collect timeseries from workers, keep in scheduler and plot


mrocklin:
Currently we record the cpu and memory usage from workers from their heartbeats.  These end up on the main status page of the dashboard as bar charts.

It might be useful to extend this in two ways

1.  Have a separate view that includes these as time series.  Additionally, to support the 1000 worker case we'll need to actually plot something like quantiles instead of all of the values beyond a certain amount
2.  Allow users to specify functions that they would like to register to also be published with these time series.  This would allow people to quickly build things like diagnostics for special hardware like GPUs

cc @lesteve 

###  Register functions

Currently in `Worker.heartbeat` we include all of the values in `**self.monitor.recent()`, which includes things like CPU and memory use.  This gets collected into the WorkerState's `info` attribute on the scheduler.  We can put other things in here as well.

So the worker probably also needs some registry of functions to run along with `self.monitor.recent()`

```python
                response = yield self.scheduler.heartbeat_worker(
                    address=self.contact_address,
                    name=self.name,
                    now=time(),
                    memory_limit=self.memory_limit,
                    executing=len(self.executing),
                    in_memory=len(self.data),
                    ready=len(self.ready),
                    in_flight=len(self.in_flight_tasks),
                    **self.monitor.recent()
                    **{k: func(self) for k, func in self.diagnostic_functions})  # <<--- this is new!
```

Then we will need to have some mechanism for users to add these diagnostic functions.  This can be done in a few ways.  I'll try to write more in a bit.  Short term when testing I would just hardcode a couple of these in a test.

mrocklin:
Here is a simple test to verify that we can register a function on the workers and have that be collected on the scheduler.


```python
@gen_cluster(client=True)
def test_diagnostic_functions(c, s, a, b):
    def myfunc(worker):
        return 123 

    a.diagnostic_functions['foo'] = myfunc
    b.diagnostic_functions['foo'] = myfunc
    
    while 'foo' not in s.workers[a.address].info:
        yield gen.sleep(0.01)

    assert s.workers[a.address].info['foo'] = 123
```

mrocklin:
Closed by #2169

PR #2169, opened 2018-08-08
Add ability to log additional custom metrics from each worker

lesteve:
After some discussion with the people interested by this feature, they are fine with having the data as point value rather than as a timeseries so I started with that which I think is simpler. This addresses part of https://github.com/dask/distributed/issues/2117.

Basically I added an additional table in the Workers tab. This is how it looks:
screenshot_20180808_190422

The layout is a bit ugly (too much space between the two tables). No clue why ...

I'll add a few more comments inline, feed-back more than welcome!


lesteve:
This is how I test the dashboard by the way:

```py
import random
import subprocess

def my_function(worker):
    return random.random()


def nvidia_data(name):
    def dask_function(dask_worker):
        cmd = f'nvidia-smi --query-gpu={name} --format=csv,noheader'
        result = subprocess.check_output(cmd.split())
        return result.strip().decode()
    return dask_function


def register_custom_metrics(dask_worker):
    dask_worker.custom_metrics['tasks'] = my_function
    # only add this if you have nvidia-smi on your workers
    # for name in ['utilization.gpu', 'utilization.memory']:
    #    dask_worker.custom_metrics[name] = nvidia_data(name)


if __name__ == '__main__':
    cluster = LocalCluster()
    client = Client(cluster)
    print(cluster.scheduler.identity()['services']['bokeh'])

    client.run(register_custom_metrics)
```

distributed/worker.py
```diff
@@ -268,7 +273,9 @@ def heartbeat(self):
                     in_memory=len(self.data),
                     ready=len(self.ready),
                     in_flight=len(self.in_flight_tasks),
-                    **self.monitor.recent())
+                    custom_metrics_names=list(custom_metrics.keys()),
```
lesteve: I ended up adding the custom_metric_names as another field in the call-back. With this you know the metrics which were added, which allows to make the code downstream a lot simpler. Maybe there is a better way of passing the metrics names around though?

distributed/worker.py
```diff
@@ -268,7 +273,9 @@ def heartbeat(self):
...
+                    **self.monitor.recent(),
+                    **custom_metrics)
```
lesteve: Should we add custom_metrics to `_register_with_scheduler` too? The reason I am asking is because self.monitor.recent is passed in [there](https://github.com/dask/distributed/pull/2169/files#diff-3f78e3c9a9a81b6355c8e594dce0b8f3R323).

distributed/bokeh/scheduler.py
```diff
@@ -933,6 +933,9 @@ def __init__(self, scheduler, width=800, **kwargs):
         self.names = ['worker', 'ncores', 'cpu', 'memory', 'memory_limit',
                       'memory_percent', 'num_fds', 'read_bytes', 'write_bytes',
                       'cpu_fraction']
+        workers = self.scheduler.workers.values()
+        self.custom_metrics_names = list(workers[0].info.get('custom_metrics_names', []))
```
lesteve: This is the way I managed to get the custom_metrics_names, but maybe there is a better way to pass them around?

mrocklin: What if only some workers have some metrics.  I can imagine this being the case with a cluster that has some GPU workers and some non-GPU workers for example.

I'll suggest the following:

    names = sorted(sum([w.info.get('custom_metrics', []) for w in workers], []))

distributed/worker.py
```diff
@@ -180,6 +180,10 @@ def __init__(self, scheduler_ip=None, scheduler_port=None,
         self.service_ports = service_ports or {}
         self.service_specs = services or {}
 
+        if custom_metrics is None:
+            self.custom_metrics = {}
```
mrocklin: What if it isn't `None`?

mrocklin: I might instead do

```python
if custom_metrics is None:
    custom_metrics = {}
self.custom_metrics = custom_metrics
```

distributed/worker.py
mrocklin: I don't understand, why isn't the dictionary passed in `custom_metrics` above sufficient?

distributed/worker.py
mrocklin: Maybe.  It wouldn't hurt.  I don't think it's critical though.

distributed/bokeh/scheduler.py
lesteve: Good point, I didn't think of that. I'll add a test for this.

mrocklin: Looks like I forgot to form it into a set

    names = sorted({n for w in workers for n in w.info.get('custom_metrics', [])})

Presumably there is also dict.get call later on when populating the tables.

distributed/worker.py
lesteve: `custom_metrics_names` is useful for the downstream code to figure out which one are the "additional custom metrics" that should go in the secondary table. Without it, you need another way to figure that out because all of the metrics have the same status in the `worker_state.info` dictionary. One is to have a long hard-coded list metric to ignore (i.e. all the ones that comes from `SystemMonitor` plus some others IIRC).

distributed/worker.py
```diff
@@ -268,7 +275,8 @@ def heartbeat(self):
                     in_memory=len(self.data),
                     ready=len(self.ready),
                     in_flight=len(self.in_flight_tasks),
-                    **self.monitor.recent())
+                    custom_metrics_names=list(custom_metrics_names_no_overlap),
+                    **metrics)
```
mrocklin: I wonder if maybe we should just do `metrics=metrics` here, and avoid the ambiguity of the `**kwargs`.  This would require changing the `**info` in `Scheduler.heartbeat_worker` as well.

mrocklin: Thoughts?

lesteve: Maybe this is a good idea ... at the moment it is not that clear (unless you know the signature of course) which are variables of the function and which ones are metrics that end up in `worker_state.info`.

I noticed also that the current code is not robust against using other names like `in_flight`, `ready`, etc ... as names of custom metrics, I need to fix that.

If we go down the route of getting rid of the `**kwargs` we probably want to do the same for `add_worker`:
https://github.com/dask/distributed/blob/a1d4a9d698d760496e06c0790cf4a617065ccbb3/distributed/scheduler.py#L1228-L1230

mrocklin: Agreed

mrocklin: Is this something that you would be willing to do within this PR?

lesteve:
Two questions:
- should we log somewhere if a custom metric is ignored because the name is already used for a core metric, e.g. `cpu`, `in_flight`. Maybe it is a bit expensive to log something on each heartbeat?
- In a previous iteration, I noticed an interesting behaviour: if you add a custom metric and then remove it, it never gets removed from `worker_state.info` (which in some sense makes sense if you send a constant value you want to send it once and not repeat it in each heartbeat). Reading the code naively seem to go in the same direction, you just keep calling `worker_state.info.update` on each heartbeat: https://github.com/dask/distributed/blob/a1d4a9d698d760496e06c0790cf4a617065ccbb3/distributed/scheduler.py#L1209-L1210. The test showing the problem is below. The interesting thing is that I did not see this problem in a dashboard. I could add/remove metrics and the table would be be update just fine (after refreshing the browser tab of course). Any idea why there is this difference between the test and the dashboard settings? Is it worth investigating a bit more?

```py
@gen_cluster(client=True)
def test_WorkerTable_custom_metrics_with_different_metrics(c, s, a, b):
    def metric_port(worker):
        return worker.port
    a.custom_metrics['metric_a'] = metric_port
    b.custom_metrics['metric_b'] = metric_port
    while ('metric_a' not in s.workers[a.address].info or
           'metric_b' not in s.workers[b.address].info):
        yield gen.sleep(0.01)
    assert s.workers[a.address].info['metric_a'] == a.port
    assert s.workers[b.address].info['metric_b'] == b.port
    wt = WorkerTable(s)
    wt.update()
    assert 'metric_a' in wt.source.data
    assert 'metric_b' in wt.source.data
    assert all(wt.source.data.values())
    assert all(len(v) == 2 for v in wt.source.data.values())
    del b.custom_metrics['metric_b']
    # TODO: this times out. Understand why 'metrics_b' never disappear from .info
    while 'metric_b' in s.workers[b.address].info:
        yield gen.sleep(0.01)
```

distributed/worker.py
lesteve: I reckon it is quicker to do this in the same PR.

mrocklin:
> Maybe it is a bit expensive to log something on each heartbeat?

We could push this through some API like `Worker.register_metric` and that could perform validation?

> you just keep calling worker_state.info.update on each heartbeat:

Do you think it would make sense to replace the `info` object rather than update it?

mrocklin:
Is there anything placed into info just once?  If so then perhaps we should separate `info` into two dicts, `info` and `metrics`?

lesteve:
I think this is ready for another review.

A summary about the status. Happy to tackle limitations in other PRs depending on what seems of higher priority:
* custom info names that overlap with core info names (e.g. `cpu`, `in_memory`, ...) are not sent and there is no logging for this. You can always look at `worker_state.info['custom_info_names']` to see which custom info is recorded in the `worker_state`. Another way would be to add one level of nesting i.e. `worker_state.info['custom']['info_name']` rather than `worker_state.info['info_name']`. I did not really find a good way of having a `register_metric` that would do the validation and, it feels easier for the user if it is a plain dict.
* custom info that were not received on the scheduler side are removed from `s.workers[worker_address].info` (this addresses the problem I was seeing in https://github.com/dask/distributed/pull/2169#issuecomment-411784351). This logic only applies to custom info names. I am fine with this since core info names (`cpu`, `in_memory`, ...) are never removed over time and only custom info names may (too be honest custom info names being removed is a bit of a edge case for 95% of usages I suspect).
* on Python >= 3.6, the ordering of custom info names on the scheduler, and on the bokeh dashboard in in the same order as on the workers. This means that you can control the ordering when registering custom info. I spent some time on this (and the code is less straightforward with this constraint than without in one or two places) but maybe this is overkill, let me know what you think.
* no additional plots based on name convention either. In the GPU use case you could imagine having a similar plot to the CPU (utilization + memory). Happy to look at that in a separate PR.
* there is no formatter based on name convention. I can do that in a separate PR. You can always do the formatting in your custom info function so I would say this is not very high priority.

mrocklin:
I'll take a look later today.  I'm looking forward to it.  Thanks for
putting this together @lesteve

lesteve:
I just pushed a commit to retrigger the CIs. Full disclosure: I have seen a few failures that seemed intermittant:
* `test_learn_occupancy_2` on [AppVeyor](https://ci.appveyor.com/project/daskdev/distributed/build/1.0.3997#L2143) and [Travis (Python 3.5)](https://travis-ci.org/dask/distributed/jobs/415438662#L2883)
* `test_dont_hold_on_to_large_messages` on [AppVeyor](https://ci.appveyor.com/project/daskdev/distributed/build/1.0.3995#L2141)
* `test_task_stream` on [Travis (Python 3.5)](https://travis-ci.org/dask/distributed/jobs/415438662#L2726)

I haven't managed to reproduce the Travis ones so far.

mrocklin:
Yeah, sorry about the CI issues.  I've been cleaning things up recently.  Some of those should be resolved soonish, but this will be a long battle.

mrocklin:
My apologies for the delay here.  I didn't get to this before a workshop happened this week.  I should be back to normal operation on Friday.  This is near the top of my priority list.

lesteve:
No worries, I have plenty of other things to keep me busy in the mean time ;-).

mrocklin:
OK, I took a look through this.  My first impression is that I think that the handling of metrics is somewhat complex.  I think that this is mostly because it was complex before, there was a lot of legacy stuff placed into `WorkerState.info`.  I'm going to try to refactor things a bit at the core level and then push up here if that's ok.

mrocklin:
@lesteve your review on the recent commit would be welcome.  I've removed the stale names and sorted names bits.  I think that it might be simpler overall though.

mrocklin:
@lesteve checking in here.  Are you ok with my pushed changes?

lesteve:
I need to look at it a bit more before commenting in more details.

I kind of liked the ordering consistency between metric registration order and Workers tab table (Python 3.6+) but since you removed the `test_WorkerTable_custom_info_columns_order` test I guess you thought it was not worth the additional complexity ...

Also do we need to think about backward-compatibility? For example maybe someone was using `worker_state.info['host']` or another key.

mrocklin:
I'm not too worried about people depending on the WorkerState.info
attribute.  Anyone depending on the scheduler internals is used to getting
burned. I don't think this project is at a stage where it should be too
concerned with keeping around old processes.

However I am somewhat concerned about the external API that exposes this
information, which is the `Scheduler.identity` method, which does get used
a bit downstream.

distributed/worker.py
```diff
@@ -283,6 +280,15 @@ def heartbeat(self):
...
             logger.debug("Heartbeat skipped: channel busy")
 
+    def get_metrics(self):
+        core = dict(executing=len(self.executing),
+                    in_memory=len(self.data),
+                    ready=len(self.ready),
+                    in_flight=len(self.in_flight_tasks))
+        custom = {k: metric(self) for k, metric in self.metrics.items()}
+
+        return merge(core, custom, self.monitor.recent())
```
lesteve: The last one wins in `merge` right? So should it not be `merge(custom, self.monitor.recent(), core)`? Maybe we need to add `in_flight` to the test (I think the test uses `cpu` at the moment) so that we check that a metric in `custom` does not override anything in `core`.

distributed/bokeh/scheduler.py
```diff
@@ -927,12 +927,18 @@ class WorkerTable(DashboardComponent):
...
     """
+    excluded_names = {'executing', 'in_flight', 'in_memory', 'ready', 'time'}
 
     def __init__(self, scheduler, width=800, **kwargs):
         self.scheduler = scheduler
         self.names = ['worker', 'ncores', 'cpu', 'memory', 'memory_limit',
                       'memory_percent', 'num_fds', 'read_bytes', 'write_bytes',
                       'cpu_fraction']
+        workers = self.scheduler.workers.values()
+        # Keep self.custom_info_names in the same order as one the workers (Python 3.6+ only)
```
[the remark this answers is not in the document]
mrocklin: Ah, good point.  What are your thoughts on ordering?  I personally don't care that much, but for your intended application it may be more important.  You probably have a better sense of what is important here than I do.  I probably just default to promising as little as possible until necessary :)

distributed/bokeh/tests/test_scheduler_bokeh.py
```diff
@@ -286,6 +286,146 @@ def test_WorkerTable(c, s, a, b):
...
+
+@gen_cluster(client=True)
+def test_WorkerTable_different_metrics(c, s, a, b):
+    def info_port(worker):
+        return worker.port
+
+    a.metrics['info_a'] = info_port
+    b.metrics['info_b'] = info_port
+
+    while not all(any('info' in k for k in s.workers[w.address].metrics)
```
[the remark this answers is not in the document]
mrocklin: That would also work.  The `'info' in k` check is string contains, not dict contains.  We're checking to see if 'info' is in any key for all metrics dicts.

distributed/scheduler.py
```diff
@@ -1004,11 +1021,11 @@ def get_worker_service_addr(self, worker, service_name):
         Returns None if the service doesn't exist.
         """
         ws = self.workers[worker]
-        port = ws.info['services'].get(service_name)
+        port = ws.services.get(service_name)
         if port is None:
             return None
```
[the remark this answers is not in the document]
mrocklin: I don't disagree.  I suggest that if we handle this we handle it elsewhere though.

distributed/worker.py
mrocklin: Good point

distributed/bokeh/scheduler.py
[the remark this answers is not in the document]
lesteve: I like the fact that you can control the ordering by how you register metrics. I don't think it is that crucial in practice though, especially since it is in alphabetical order you can always rename your metric if you really care.

lesteve:
Thanks for the flake8 fix, if there is anything else I can do, let me know!

lesteve:
Looking at the doc, maybe adding a small subsection about the "Workers" tab in https://distributed.readthedocs.io/en/latest/web.html#plots could be an option. After that a few lines could be added to explain how to register custom metrics.

Do you want this to be done in this PR?

mrocklin:
A lot of that page now seems out of date.  We might consider adding to http://dask.pydata.org/en/latest/diagnostics-distributed.html which is a bit more up-to-date.  

I think that long term I would rather integrate documentation into the dashboard itself.  I'm hoping that this is one of the things that becomes possible after https://github.com/dask/distributed/pull/2150 gets in.

lesteve:
> I think that long term I would rather integrate documentation into the dashboard itself. I'm hoping that this is one of the things that becomes possible after #2150 gets in.

That would be nice indeed!

> A lot of that page now seems out of date. We might consider adding to http://dask.pydata.org/en/latest/diagnostics-distributed.html which is a bit more up-to-date.

Quickly looking at the dask doc, it seems very high-level and links to the distributed doc for more details, so I am slightly in favour of including it in the distributed doc (even if some of it is out of date). My last commit added a "Per-worker resources" section (see snapshot below for how it looks). Feel free to revert this commit if you think this PR is better without it!

image

mrocklin:
Looks great.  And we can pull it out into new docs if a good location arises.

Merging

lesteve:
Great, nice to see this merged!

mrocklin:
It's a nice feature, resulted in some useful refactoring, and hopefully
also gave you a sense for how things work internally a bit better, which
I'm pretty happy about.

lesteve:
Working on this PR was very useful indeed to get a better picture about how things fit together!

distributed/bokeh/scheduler.py
```diff
@@ -328,13 +328,13 @@ def update(self):
                 else:
                     processing_color.append('blue')
 
-            nbytes = [ws.info['memory'] for ws in workers]
+            nbytes = [ws.metrics['memory'] for ws in workers]
```
mrocklin: @jhamman you'll probably want to be aware of this change for https://github.com/dask/dask-kubernetes/pull/101

PR #2171, opened 2018-08-08
fix scale /  avoid returning coroutines

jhamman:
This reworks the scale down block of `Cluster.scale()` to avoid calling coroutines. 

xref: https://github.com/dask/dask-jobqueue/pull/97#discussion_r208428234


jhamman:
I'm seeing what I think is an unrelated test failure in `distributed/tests/test_actor.py::test_compute`. Is this a know issue?

mrocklin:
Not a known issue.  But that test was added recently, so I wouldn't be surprised if it was unrelated.

mrocklin:
Thanks @jhamman

Issue #2166, opened 2018-08-07
--memory-limit is per process, not per invocation of dask-worker.


rainwoodman:
After connecting the workers to the scheduler, it appears that dask-worker forked nprocs processes, and each has the limit given by the --memory-limit passed to dask-worker.

The documentation wasn't very clear about this, and it was suggested nprocs is semantically equivalent to nthreads [1] which implied the memory-limit per forked process to be 1 / nprocs of the --memory-limit argument. Perhaps the current behavior shall be changed to make processes more equivalent to threads, or the document may need to be clarified that the limit is per process, not per invocation of dask-worker.

[1] https://distributed.readthedocs.io/en/latest/worker.html#thread-pool

mrocklin:
I agree that the `--nprocs` parameter is a bit confusing, especially with
the interactions with `--memory-limit`.  I still have to check what the
behavior is each time when using this parameter.  As always, the challenge
to changing a parameter like this is doing it in a clean way that doesn't
break existing deployments.  Suggestions on the right path towards
something more sensible would be welcome.

rainwoodman:
There is a way to make sense of this without breaking current deployments: just state clearly the memory limit is per process, not per invocation of dask-worker. I think we do need to make clear that the number of  invocations of dask-worker differs from number of worker processes. I can file a PR.

lesteve:
I think his one has been closed by #2173.

@rainwoodman note you need to put the magical words (e.g. close #issueNumber) in the PR description to have the issue closed automatically when the PR is merged. Putting the magical words in a comment inside the PR does now work unfortunately ...

PR #2173, opened 2018-08-08
clarify dask-worker --nprocs

rainwoodman:
First commit. Will also modify the help string and docstrings.

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -60,15 +60,15 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 @click.option('--interface', type=str, default=None,
               help="Network interface like 'eth0' or 'ib0'")
 @click.option('--nthreads', type=int, default=0,
-              help="Number of threads per process.")
+              help="Number of threads in the pool per process.")
 @click.option('--nprocs', type=int, default=1,
-              help="Number of worker processes.  Defaults to one.")
+              help="Number of worker processes to spawn.  Defaults to one.")
 @click.option('--name', type=str, default='',
               help="A unique name for this worker like 'worker-1'. "
                    "If used with --nprocs then the process number "
                    "will be appended like name-0, name-1, name-2, ...")
 @click.option('--memory-limit', default='auto',
-              help="Bytes of memory that the worker can use. "
+              help="Bytes of memory per process that the worker can use. "
                    "This can be an integer (bytes), "
                    "float (fraction of total system memory), "
                    "string (like 5GB or 5000M), "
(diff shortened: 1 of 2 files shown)


rainwoodman:
Closes #2166

distributed/cli/dask_worker.py
```diff
@@ -60,15 +60,15 @@
 @click.option('--interface', type=str, default=None,
               help="Network interface like 'eth0' or 'ib0'")
 @click.option('--nthreads', type=int, default=0,
-              help="Number of threads per process.")
+              help="Number of threads in the pool per process.")
```
mrocklin: What does "in the pool" refer to here, the `ThreadPoolExecutor`?  If so I suggest that we drop this.  A novice user may not understand that threads come in pools or that dask uses a ThreadPool.  They just know that dask uses threads internally.

distributed/cli/dask_worker.py
```diff
@@ -60,15 +60,15 @@
...
 @click.option('--nprocs', type=int, default=1,
-              help="Number of worker processes.  Defaults to one.")
+              help="Number of worker processes to spawn.  Defaults to one.")
```
mrocklin: Sometimes we spawn, sometimes we fork, it might be better to avoid using this term explicitly.

docs/source/worker.rst
```diff
@@ -143,12 +146,13 @@ Memory Management
 
 Workers are given a target memory limit to stay under with the
 command line ``--memory-limit`` keyword or the ``memory_limit=`` Python
-keyword argument.::
+keyword argument, which sets the memory limit per worker processes launched
+by dask-workder ::
 
-    $ dask-worker tcp://scheduler:port --memory-limit=auto  # total available RAM
-    $ dask-worker tcp://scheduler:port --memory-limit=4e9  # four gigabytes
+    $ dask-worker tcp://scheduler:port --memory-limit=auto  # total available RAM on the node
```
mrocklin: I recommend using the term "machine" rather than "node" which is often used to mean other things as well.

distributed/cli/dask_worker.py
rainwoodman: I thought fork+exec is one way of spawning. `to launch`?

distributed/cli/dask_worker.py
rainwoodman: That's right.

docs/source/worker.rst
rainwoodman: Yes. We shall make sure the set of terminology is defined, if not yet.

mrocklin:
This looks good to me.  Thank you for these changes @rainwoodman !

Merging.

mrocklin:
Also, I see that this is your first commit to this repository.  Welcome!

I know you've already commited plenty of work to dask/dask, but it's nice to have more people over here too :)

rainwoodman:
Sure. This was done via the web interface.

PR #2175, opened 2018-08-09
Use dispatch for dask serialization

mrocklin:
This allows people to register serialization functions using decorator syntax

```python
@dask_serialize.register(np.ndarray)
def serialize_array(x):
    ...

@dask_deserialize.register(np.ndarray)
def serialize_array(header, frames):
    ...
```

This also means that inheritance turns on by default
(which is both good and bad)

### DIFF
--- a/distributed/protocol/keras.py
+++ b/distributed/protocol/keras.py
@@ -1,8 +1,11 @@
 from __future__ import print_function, division, absolute_import
 
-from .serialize import register_serialization, serialize, deserialize
+from .serialize import dask_serialize, dask_deserialize, serialize, deserialize
+
+import keras
 
 
+@dask_serialize.register(keras.Model)
 def serialize_keras_model(model):
     import keras
     if keras.__version__ < '1.2.0':
@@ -18,6 +21,7 @@ def serialize_keras_model(model):
     return header, frames
 
 
+@dask_deserialize.register(keras.Model)
 def deserialize_keras_model(header, frames):
     from keras.models import model_from_config
     n = 0
@@ -29,9 +33,3 @@ def deserialize_keras_model(header, frames):
     model = model_from_config(header)
     model.set_weights(weights)
     return model
-
-
-for module in ['keras', 'tensorflow.contrib.keras.python.keras']:
-    for name in ['engine.training.Model', 'models.Model', 'models.Sequential']:
-        register_serialization('.'.join([module, name]), serialize_keras_model,
-                               deserialize_keras_model)
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -2,6 +2,7 @@ from __future__ import print_function, division, absolute_import
 from functools import partial
 import traceback
 
+import dask
 from dask.base import normalize_token
 try:
     from cytoolz import valmap, get_in
@@ -17,41 +18,35 @@ from .compression import maybe_compress, decompress
 from .utils import unpack_frames, pack_frames_prelude, frame_split_size
 
 
-class_serializers = {}
-
 lazy_registrations = {}
+
+
+dask_serialize = dask.utils.Dispatch('dask_serialize')
+dask_deserialize = dask.utils.Dispatch('dask_deserialize')
 
 
 def dask_dumps(x, context=None):
     """Serialise object using the class-based registry"""
-    typ = typename(type(x))
-    if typ in class_serializers:
-        dumps, loads, has_context = class_serializers[typ]
-        if has_context:
-            header, frames = dumps(x, context=context)
-        else:
-            header, frames = dumps(x)
-        header['type'] = typ
-        header['serializer'] = 'dask'
-        return header, frames
-    elif _find_lazy_registration(typ):
-        return dask_dumps(x)  # recurse
+    type_name = typename(type(x))
+    try:
+        dumps = dask_serialize.dispatch(type(x))
+    except TypeError:
+        raise NotImplementedError(type_name)
+    if has_keyword(dumps, 'context'):
+        header, frames = dumps(x, context=context)
     else:
-        raise NotImplementedError(typ)
+        header, frames = dumps(x)
+
+    header['type'] = type_name
+    header['type-serialized'] = pickle.dumps(type(x))
+    header['serializer'] = 'dask'
+    return header, frames
 
 
 def dask_loads(header, frames):
-    typ = header['type']
-
-    if typ not in class_serializers:
-        _find_lazy_registration(typ)
-
-    try:
-        dumps, loads, _ = class_serializers[typ]
-    except KeyError:
-        raise TypeError("Serialization for type %s not found" % typ)
-    else:
-        return loads(header, frames)
+    typ = pickle.loads(header['type-serialized'])
+    loads = dask_deserialize.dispatch(typ)
+    return loads(header, frames)
 
 
 def pickle_dumps(x):
@@ -406,20 +401,20 @@ def register_serialization(cls, serialize, deserialize):
     serialize
     deserialize
     """
-    if isinstance(cls, type):
-        name = typename(cls)
-    elif isinstance(cls, str):
-        name = cls
-    class_serializers[name] = (serialize,
-                               deserialize,
-                               has_keyword(serialize, 'context'))
+    if isinstance(cls, str):
+        raise TypeError(
+            "Strings are no longer accepted for type registration. "
+            "Use dask_serialize.register_lazy instead"
+        )
+    dask_serialize.register(cls)(serialize)
+    dask_deserialize.register(cls)(deserialize)
 
 
 def register_serialization_lazy(toplevel, func):
     """Register a registration function to be called if *toplevel*
     module is ever loaded.
     """
-    lazy_registrations[toplevel] = func
+    raise Exception("Serialization registration has changed. See documentation")
 
 
 def typename(typ):
@@ -434,33 +429,19 @@ def typename(typ):
     return typ.__module__ + '.' + typ.__name__
 
 
-def _find_lazy_registration(typename):
-    toplevel, _, _ = typename.partition('.')
-    if toplevel in lazy_registrations:
-        lazy_registrations.pop(toplevel)()
-        return True
-    else:
-        return False
-
-
 @partial(normalize_token.register, Serialized)
 def normalize_Serialized(o):
     return [o.header] + o.frames  # for dask.base.tokenize
 
 
 # Teach serialize how to handle bytestrings
+@dask_serialize.register((bytes, bytearray))
 def _serialize_bytes(obj):
     header = {}  # no special metadata
     frames = [obj]
     return header, frames
 
 
+@dask_deserialize.register((bytes, bytearray))
 def _deserialize_bytes(header, frames):
     return frames[0]
-
-
-# NOTE: using the same exact serialization means a bytes object may be
-# deserialized as bytearray or vice-versa...  Not sure this is a problem
-# in practice.
-register_serialization(bytes, _serialize_bytes, _deserialize_bytes)
-register_serialization(bytearray, _serialize_bytes, _deserialize_bytes)
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -1,6 +1,5 @@
 from __future__ import print_function, division, absolute_import
 from functools import partial
-import numbers
 import traceback
 
 import dask
@@ -453,7 +452,12 @@ def _deserialize_bytes(header, frames):
 #########################
 
 
-basic = (numbers.Number, str)
+def _is_msgpack_serializable(v):
+    typ = type(v)
+    return (typ is str or typ is int or typ is float or
+            isinstance(v, dict) and all(map(_is_msgpack_serializable, v.values()))
+                                and all(typ is str for x in v.keys()) or
+            isinstance(v, (list, tuple)) and all(map(_is_msgpack_serializable, v)))
 
 
 def serialize_object_with_dict(est):
@@ -471,11 +475,7 @@ def serialize_object_with_dict(est):
         d = est.__dict__
 
     for k, v in d.items():
-        typ = type(v)
-        if (typ is str or typ is int or typ is float or
-                isinstance(v, dict) and all(typ is str or typ is int or typ is float for x in v.values())
-                                    and all(typ is str for x in v.keys()) or
-                isinstance(v, (list, tuple)) and all(typ is str or typ is int or typ is float for x in v)):
+        if _is_msgpack_serializable(v):
             header['simple'][k] = v
         else:
             if isinstance(v, dict):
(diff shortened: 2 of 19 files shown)


mrocklin:
@stsievert with this branch you should be able to do the following to serialize your objects:

```python
from distributed.protocol import register_attributes

register_attributes(your_custom_class)
```

This will traverse down `your_obj.__dict__` and serialize those, probably capturing the Keras model along the way.

mrocklin:
I've also added a torch serialization scheme for Dask here.  This is a bit faster than the pickle solution:

```python
In [1]: import torchvision

In [2]: from distributed.protocol import serialize, deserialize

In [3]: import pickle

In [4]: model = torchvision.models.resnet50()

In [5]: %time header, frames = serialize(model)
CPU times: user 46.9 ms, sys: 0 ns, total: 46.9 ms
Wall time: 46.3 ms

In [6]: %timeit serialize(model)
19.1 ms ± 1.14 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [7]: %timeit deserialize(header, frames)  # most of this seems to be torch.Tensor(numpy_array)
64.2 ms ± 1.11 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %time b = pickle.dumps(model)
CPU times: user 77.1 ms, sys: 68.2 ms, total: 145 ms
Wall time: 142 ms

In [9]: %timeit pickle.dumps(model)
108 ms ± 583 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [10]: %timeit pickle.loads(b)
111 ms ± 1.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

I ran a statistical profiler on this and we spend most of our time turning numpy arrays back into pytorch Tensors.  I do wish that there was a way to get memoryviews directly in and out of torch Tensors.  It's quite fast on the sending side though and a noticably faster overall.

cc @stsievert @lesteve

stsievert:
Thanks for this PR. I'll start using this later tonight.

> I've also added a torch serialization scheme for Dask here.

For serialization I'd expect a 1.1x speedup to be significant because serialization is so common. I don't know if I'd call this PR "a bit" faster by that measure. 

> most of this seems to be torch.Tensor(numpy_array)

I'm surprised this takes so long; PyTorch tensors and NumPy arrays share the same memory locations.

``` python
In [1]: import torch

In [2]: import numpy as np

In [4]: n = int(10e6)

In [5]: x = np.random.rand(n)

In [7]: %time _ = torch.Tensor(x)
CPU times: user 23.6 ms, sys: 15 ms, total: 38.6 ms
Wall time: 46.4 ms

In [8]: %time _ = torch.from_numpy(x)
CPU times: user 33 µs, sys: 104 µs, total: 137 µs
Wall time: 1.11 ms 

In [11]: y = torch.from_numpy(x)

In [12]: %time _ = y.numpy()
CPU times: user 35 µs, sys: 94 µs, total: 129 µs
Wall time: 969 µs

In [16]: %time _ = y.detach().numpy()  # no retain_grad anymore; direct copy
CPU times: user 21 µs, sys: 1 µs, total: 22 µs
Wall time: 26 µs

In [16]: %time _ = y.detach().numpy()
CPU times: user 21 µs, sys: 1 µs, total: 22 µs
Wall time: 26 µs

# begin %timeit calls

In [17]: %timeit _ = torch.Tensor(x)
15.1 ms ± 352 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [18]: %timeit _ = torch.from_numpy(x)
1.51 µs ± 6.73 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [19]: %timeit _ = y.numpy()
804 ns ± 16 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [20]: %timeit _ = y.detach().numpy()  # no retain_grad anymore; direct copy
1.12 µs ± 5.52 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

mrocklin:
`from_numpy` looks like a winner :)

If the original tensor has different device or requires_grad attributes what is the right way to assign them?

mrocklin:
With @stsievert 's `from_numpy` fix

```python
In [1]: from distributed.protocol import serialize, deserialize

In [2]: import torchvision

In [3]: model = torchvision.models.resnet50()

In [4]: %time header, frames = serialize(model)
CPU times: user 47.4 ms, sys: 265 µs, total: 47.6 ms
Wall time: 46.3 ms

In [5]: %timeit serialize(model)
18.3 ms ± 98.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit deserialize(header, frames)
6.72 ms ± 106 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

For the 100MB model that's around 4GB/s, which should make serialization free until we get to exotic networks.

However, we still need to make sure that we faithfully roundtrip the device and requires_grad attributes.

stsievert:
> has different device

PyTorch tensors can either be on the CPU or on a CUDA GPU.

To move a Tensor `x` from CUDA to the CPU (which is required for it to be a NumPy array), `x.cpu().numpy() if x.is_cuda`. I'd probably record `torch.device(x)` in the serialization (so it can be sent to the right device). The docs have an example of sending it to the right device: https://pytorch.org/docs/stable/tensor_attributes.html#torch.torch.device

``` python
>>> # Example of a function that takes in a torch.device
>>> cuda1 = torch.device('cuda:1')
>>> x = torch.randn((2,3), device=cuda1)
>>> x.device  # added
... device(type='cpu')

```

> requires_grad 

I think something like

``` python
def serialize(...):
    # this NoneType check on x.grad is what torch.optim uses
    if x.grad is not None and torch.abs(x.grad).sum() != 0:
        headers['has_grad'] = True
        frames += [serialize(x.grad)]
    ...
    return header, frames

def deserialize(header, frames):
    ...
    if header.get('has_grad', False):
        if has_gradient(frames):
            x.grad = get_gradient(frames)
        else:
            x.grad = torch.zeros_like(x)
    return x
```

I'd only maybe include the gradient sum check; I don't think I'd expect to serialize a torch model in the middle of training (but I might be wrong). I think I'd only do it after training is finished and `optimizer.zero_grad` is called.

stsievert:
I must be missing something about serialization if 9x speedups are possible – I'd think 1.1x speedups would be significant. What am I missing?

(my original question is in https://github.com/dask/distributed/pull/2175#issuecomment-411925613, and I think my edits to add the question got lost in the recent comments)

mrocklin:
My inclination is to include the gradient if its around.  It would be up to
the user to remove the gradient beforehand if they didn't want it sent.

If you agree, then can you suggest a simple way to give a tensor a gradient
for a test?

stsievert:
> can you suggest a simple way to give a tensor a gradient for a test?

Sure – `Tensor.grad` is another PyTorch Tensor, which means it can be assigned.

``` python
In [1]: y = torch.from_numpy(...)
In [22]: y
Out[22]:
tensor([ 5.5241e-01,  7.7753e-01,  5.6075e-01,  ...,  3.9952e-01,
         2.9477e-01,  2.4177e-01], dtype=torch.float64)

In [23]: y.grad

In [24]: y.grad = torch.zeros_like(y) + 1

In [25]: y.grad.mean()
Out[25]: tensor(1., dtype=torch.float64)
```

mrocklin:
> I must be missing something about serialization if 9x speedups are possible – I'd think 1.1x speedups would be significant. What am I missing?

This is just serialization.  For numeric data that is already pretty much good to send on the wire, this should be free.  For example for a numpy array we can just grab x.data and put it on a wire

```python
In [1]: import numpy as np

In [2]: x = np.random.random(100000000)

In [3]: x.data
Out[3]: <memory at 0x7f754c084648>

In [4]: socket.send(x.data)
```

The fact that serialization costs *anything* is a flaw.  Note that serialization doesn't include the cost of actually communication that data.

stsievert:
> This is just serialization.

D'oh. So the pickling of PyTorch Tensors can be improved. Any ideas on why PyTorch serialization is so slow?

lesteve:
Interesting, thanks for this! Any insights why pickle is slower?

mrocklin:
It isn't *that* slow.  It will never be as fast as just moving memoryviews around (pickle today requires that you construct a new bytesring).  For details though you would have to look at `torch.save`, which probably means looking at sourcecode.

mrocklin:
@jcrist if you have time can I ask for your review on this?  I recommend going commit by commit for the first four commits (the last three should be squashed into one).

distributed/protocol/serialize.py
```diff
@@ -434,33 +430,116 @@ def typename(typ):
...
+        d = est
+    else:
+        d = est.__dict__
+
+    for k, v in d.items():
+        typ = type(v)
+        if (typ is str or typ is int or typ is float or
+                isinstance(v, dict) and all(typ is str or typ is int or typ is float for x in v.values())
+                                    and all(typ is str for x in v.keys()) or
+                isinstance(v, (list, tuple)) and all(typ is str or typ is int or typ is float for x in v)):
```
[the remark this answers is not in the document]
jcrist: Given the above, perhaps a protocol like pickle5 (not necessarily using pickle5) might be better in the long run (I think one could be built without changing the existing per-type functions, just how we dispatch to them in distributed). My quick thought is that something like this could probably be built to piggyback on top of the existing `Pickler`/`Unpickler` classes, with each handling an additional memo of frames that are appended-to/extracted-from during normal pickle operations.

distributed/protocol/serialize.py
```diff
@@ -434,33 +430,116 @@ def typename(typ):
...
+        v = deserialize(h, f)
+        dd[k] = v
+
+    return obj
+
+
+dask_deserialize.register(dict)(deserialize_object_with_dict)
+
+
+def register_attributes(cls):
```
jcrist: This name feels a bit off to me. Not sure if `register_generic` would be better.

distributed/protocol/serialize.py
```diff
@@ -434,33 +430,116 @@ def typename(typ):
...
-# deserialized as bytearray or vice-versa...  Not sure this is a problem
-# in practice.
-register_serialization(bytes, _serialize_bytes, _deserialize_bytes)
-register_serialization(bytearray, _serialize_bytes, _deserialize_bytes)
+#########################
+# Descend into __dict__ #
+#########################
+
+
+basic = (numbers.Number, str)
```
jcrist: Unused?

distributed/protocol/keras.py
```diff
@@ -1,8 +1,11 @@
 from __future__ import print_function, division, absolute_import
 
-from .serialize import register_serialization, serialize, deserialize
+from .serialize import dask_serialize, dask_deserialize, serialize, deserialize
 
+import keras
 
+
+@dask_serialize.register(keras.Model)
```
stsievert: This line assumes that every Keras model is an instance of `keras.Model`, and there are cases where that's not true. Taking from their "[Getting started guide on Sequential models][1]:

[1]:https://keras.io/getting-started/sequential-model-guide/

``` python
>>> from keras.datasets import mnist
>>> from keras.models import Sequential
>>> from keras.layers import Dense, Dropout, Flatten, Conv2D, MaxPooling2D
>>> from keras import backend as K
>>> import keras
>>>
>>> model = Sequential()
>>> model.add(Conv2D(32, kernel_size=(3, 3),
>>>                  activation='relu',
>>>                  input_shape=(28, 28, 1)))
>>> model.add(Conv2D(64, (3, 3), activation='relu'))
>>> model.add(MaxPooling2D(pool_size=(2, 2)))
>>> model.add(Dropout(0.25))
>>> model.add(Flatten())
>>> model.add(Dense(128, activation='relu'))
>>> model.add(Dropout(0.5))
>>> model.add(Dense(10, activation='softmax'))
>>> type(model).mro()
... [keras.engine.sequential.Sequential,
... keras.engine.training.Model,
... keras.engine.network.Network,
... keras.engine.base_layer.Layer,
... object]
>>>
>>> model.compile(loss='categorical_crossentropy',
              optimizer=opt)
>>> type(model).mro()
... [keras.engine.sequential.Sequential,
... keras.engine.training.Model,
... keras.engine.network.Network,
... keras.engine.base_layer.Layer,
... object]
```

I think the functionality implemented at the end of this file should be preserved (`for module in ... for name in ...`). I tried adding these lines back in but couldn't get it working. Here's a gist of what I have: https://gist.github.com/stsievert/7389acd62eb6f6d151286214cb51c392

distributed/protocol/serialize.py
```diff
@@ -434,33 +430,116 @@ def typename(typ):
...
+    frames = []
+
+    if isinstance(est, dict):
+        d = est
+    else:
+        d = est.__dict__
+
+    for k, v in d.items():
+        typ = type(v)
+        if (typ is str or typ is int or typ is float or
```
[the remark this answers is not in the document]
mrocklin: Good thought.  Done.

distributed/protocol/serialize.py
mrocklin: Done

distributed/protocol/serialize.py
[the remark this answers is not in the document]
mrocklin: Agreed about loss os specialized containers.  This is already an issue with using msgpack.

That's an interesting idea about piggybacking on pickle.  What we're doing now is clearly reinventing a (fairly complex) wheel.

distributed/protocol/serialize.py
mrocklin: Removed

distributed/protocol/keras.py
mrocklin: It seems to work ok for me from your example

```python
In [5]: from distributed.protocol import serialize, deserialize

In [6]: header, frames = serialize(model)
2018-08-17 06:36:00.382137: I tensorflow/core/platform/cpu_feature_guard.cc:141] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX AVX2 FMA
2018-08-17 06:36:00.396155: I tensorflow/core/common_runtime/process_util.cc:69] Creating new thread pool with default inter op setting: 2. Tune using inter_op_parallelism_threads for best performance.

In [7]: model2 = deserialize(header, frames)

In [8]: model2
Out[8]: <keras.engine.sequential.Sequential at 0x7f4d62c50828>
```

I think that we're depending here on `keras.model_from_config` to work properly

stsievert: Hm. I'm still seeing `TypeError: can't pickle _thread.RLock objects` on e5792a341045366cd2149cdee85985e9deceff5d in `deserialize` (`serialize` works fine). An example of what I'm seeing is in  https://gist.github.com/stsievert/4f593614450a143bf1b342e1fdeadce9.

mrocklin: Oddly that notebook rolls through fine for me.  The only difference in versions I see is that keras is 2.2.2

stsievert: Whoops, looks like I had modifications (`.dirty` in `distributed.__version__`).

I have reverted back to e5792a341045366cd2149cdee85985e9deceff5d and tested with different Keras versions. It works for Keras 2.2.2, 2.2.0, 2.1.0, but fails with 2.0.0.

stsievert: And now https://github.com/dask/distributed/issues/2183 is what I'm seeing.

mrocklin: Reproduced.  Thank you.

mrocklin:
Merging this tomorrow if there are no further comments.

PR #2176, opened 2018-08-09
Add dashboard_link property

jacobtomlinson:
It is useful to be able to access the dashboard link programatically. For example if using dask-kubernetes outside of a notebook I would still like to be able to easily get a link to the dashboard.


mrocklin:
This makes sense to me.  Thanks @jacobtomlinson

mrocklin:
Also, be warned, our CI is going through some rough times.  Please bear with us :)

mrocklin:
Thanks @jacobtomlinson !

Issue #2187, opened 2018-08-16
KeyError in Scheduler.update_graph


sdrave:
Hi, there seems to be an issue with the cleanup logic added to `Scheduler.update_graph` in #1840.

For instance, when I execute

```python
import time

def foo():
    return 'foo'

def bar(x):
    return x + 'bar'

def baz(x):
    time.sleep(100)
    return x + 'baz'

if __name__ == '__main__':
    from distributed import Client, LocalCluster
    cluster = LocalCluster()
    c = Client(cluster)
    x = c.submit(foo)
    y = c.submit(bar, x)
    y.result()
    z = c.submit(baz, y)
    del y
    time.sleep(1)
    f = c.submit(bar, x)
```

I get with distributed 1.22.1 the following error:

```
Traceback (most recent call last):
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/distributed/core.py", line 320, in handle_comm
    result = yield result
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
    raise self._exception
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/tornado/gen.py", line 1147, in run
    yielded = self.gen.send(value)
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/distributed/scheduler.py", line 1924, in add_client
    yield self.handle_stream(comm=comm, extra={'client': client})
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
    raise self._exception
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/tornado/gen.py", line 1147, in run
    yielded = self.gen.send(value)
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/distributed/core.py", line 374, in handle_stream
    handler(**merge(extra, msg))
  File "/home/stephan/.virtualenvs/pymor/lib/python3.5/site-packages/distributed/scheduler.py", line 1357, in update_graph
    for dep in dependencies[key]:
KeyError: 'foo-b55e006af80e0cfdc1a300901102086d'
```

[...]

PR #2190, opened 2018-08-17
Fix KeyError in Scheduler.update_graph

sdrave:
This addresses #2187.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1366,8 +1366,7 @@ class Scheduler(ServerNode):
             done = set(already_in_memory)
             while stack:  # remove unnecessary dependencies
                 key = stack.pop()
-                ts = self.tasks[key]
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @sdrave .  This would work.

However, it might be better to figure out what happened and then hopefully resolve the underlying issue.  It would be interesting to investigate why were we looking for a key in dependencies when none exists.

sdrave:
Hi @mrocklin, not sure if I understand enough of the internals of distributed. 

Why should every dependency be again a key in `dependencies`? In my understanding, the failing dependency just doesn't belong to the newly submitted tasks. (It is not contained in the `tasks`, but it is a key in `self.tasks`.)

mrocklin:
Superceded by https://github.com/dask/distributed/pull/2196

PR #2191, opened 2018-08-17
Warn if desired port is already in use

mrocklin:


stsievert:
I'm not sure if this related, but I'm seeing this in notebooks when running this on master:

screen shot 2018-08-20 at 3 36 37 pm

I'm fine with the warning, but it'd be nice to have a dashboard link, even if it's on a different port.

I'm not aware of any other notebooks running, but apparently I have some orphan server running though I can't see it. I'm not sure how to kill it without restarting my machine. Could we add either (or both)

* directions on how to kill the other dashboard server (a button to kill it would be even better)
* an option to launch the notebook to a different port (the old behavior)

mrocklin:
Sure, we can try downgrading the exception to a warning.  Give https://github.com/dask/distributed/pull/2199 a try?

PR #2198, opened 2018-08-20
Add plot= keyword to get_task_stream

mrocklin:
```python
In [1]: from dask.distributed import Client, get_task_stream

In [2]: import dask.array as da

In [3]: client = Client()

In [4]: with get_task_stream(plot='save', filename='foo.html') as ts:
   ...:     da.random.random((10000, 10000), chunks=(1000, 1000)).std().compute()

In [5]: !ls foo.html
foo.html

In [6]: type(ts.figure)
Out[6]: bokeh.plotting.figure.Figure
```

cc @sklam

### DIFF
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -59,59 +59,12 @@ class TaskStream(DashboardComponent):
         """
         kwargs are applied to the bokeh.models.plots.Plot constructor
         """
-        clear_interval = parse_timedelta(clear_interval, default='ms')
         self.n_rectangles = n_rectangles
+        clear_interval = parse_timedelta(clear_interval, default='ms')
         self.clear_interval = clear_interval
         self.last = 0
 
-        self.source = ColumnDataSource(data=dict(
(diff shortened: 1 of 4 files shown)


mrocklin:
@sklam it would be good to get your feedback on this before we merge if you have time.

sklam:
This is great!  I have a few questions but none of them are stoppers.

* What happens if `filename` is not given to `.get_task_stream()`?  I thought it will show up as "unamed.html" or something like that, but that's not the case.  If `filename` is necessary, the docstring should call that out.
* What's the use of `ts.figure`?  I can't seem to plot it in the notebook w/ `bokeh.io.show(ts.figure)`.

mrocklin:
Good points.

For the filename I've added an explicit default.  Previously it would save to a file in `/tmp` that would have been hard to convey to the user.

For the figure it's just a bokeh figure.  You could, for example, save this figure your own way rather than using the convenience `plot='save'` command.  In principle you should also be able to display it to the notebook.  I'm not sure why you can't.  I would probably escalate that question up to Bokeh.

PR #2199, opened 2018-08-20
Downgrade exception to warning when reusing port

mrocklin:
cc @stsievert can you give this PR a try and see if it works for you?

### DIFF
--- a/distributed/bokeh/core.py
+++ b/distributed/bokeh/core.py
@@ -2,6 +2,7 @@ from __future__ import print_function, division, absolute_import
 
 from distutils.version import LooseVersion
 import os
+import warnings
 
 import bokeh
 from bokeh.server.server import Server
@@ -48,7 +49,8 @@ class BokehServer(object):
                                % port)
                     else:
                         msg = "Failed to start diagnostics server on port %d. " % port + str(exc)
-                    raise type(exc)(msg)
+                    warnings.warn('\n' + msg)
+                    port = 0
                 if i == 4:
                     raise
 
--- a/distributed/bokeh/core.py
+++ b/distributed/bokeh/core.py
@@ -45,7 +45,8 @@ class BokehServer(object):
                     if ("already in use" in str(exc) or  # Unix/Mac
                             "Only one usage of" in str(exc)):  # Windows
                         msg = ("Port %d is already in use. "
-                               "Perhaps you already have a cluster running?"
+                               "\nPerhaps you already have a cluster running?"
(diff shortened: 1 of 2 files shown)


stsievert:
Things have always worked okay. I updated the image to show dask.distributed’s path, but I think the image before also had the dashboard link.

lesteve:
I stumbled on this in master and I have to admit I find the warning slightly on the noisy side. It feels like if no `diagnostics_port` argument is passed into `Client`, the user does not care too much about which port the dashboard uses and there should be no warning. IMHO the warning should also say that another port has been chosen for the dashboard and other than this everything is completely fine.

I suspect people are rather likely to have multiple clusters on the same machine (multiple notebooks running, different clusters defined in multiple cells in the same notebook, or rerunning the same cell that creates a cluster) and may find it slightly annoying that they have to pass `diagnostics_port` to silence it. In dask-jobqueue I am not 100% sure there is a straightforward way to override `diagnostics_port` of the underlying `LocalCluster`.

I may well be missing something about adding a warning for this "port is already in use" case so I would be more than happy to hear about the reason behind it ...

mrocklin:
I made the change originally after sitting with novice users a few times and noticing the same behavior.  They would work in a notebook, re-run the cell that creates a local `Client()` then they would run some code and be confused why the dashboard wasn't changing. 

Anyone with some experience of using dask in a notebook is pretty used to avoiding the cell that creates the client or are accustomed to restarting the notebook every time (this is what I do).  I think that this no longer bothers us but is a very common stumbling block when setting up new users.

My hope was that this warning would give them a hint about what was going on.

I suspect that there is a way to make everyone happy though.  Does anyone have any suggestions that could both help the situation I describe above while also not being too noisy?

mrocklin:
@lesteve how is this?

```python

In [2]: client = Client(processes=False)

In [3]: client = Client(processes=False)
/home/mrocklin/workspace/distributed/distributed/bokeh/core.py:53: UserWarning: 
Port 8787 is already in use. 
Hosting the diagnostics dashboard on a random port instead.
  warnings.warn('\n' + msg)
```

lesteve:
> @lesteve how is this?

Looks like I did not hit Enter of my comment but tl;dr looks good to me!

For completeness, here it is:
I see I guess the novice use case kind of makes sense. I don't really have a good suggestion off the top of my head to make everyone happy. A config option would work but seems a bit overkill for this IMO.

I would say merge this one and see if someone else than me complains ;-).

PR #2201, opened 2018-08-21
Add a worker initialization function

### LINKED ISSUE #2140: Worker initialization function
As discussed in #2110 it would be able to register a function with the scheduler which is guaranteed to run on a worker when it first connects.

> Eventually it would be nice to allow clients to register functions to be run at worker start time with the scheduler that could be passed to workers as they start up.
>
> 1. Client registers func with the scheduler as a preload operation
> 2. Scheduler holds on to func
> 3. Scheduler also sends func to all workers to have them run func
> 4. Workers run func
> 5. New worker arrives, tells Scheduler that it exists
> 6. As part of saying "Hi" the scheduler also gives the worker func and tells it to run it
>
> If anyone is interested in implementing this let me know and I'll point to the various locations in the client, scheduler, and worker, where these changes would have to be made. It's a modest amount of work and would be a good introduction to the distributed system.

guillaumeeb:
Closes #2140 

So I think I've got a first working example here. Also tested on a real deployment. Happily waiting for any feedback! I've some particular doubts considering:
- the test I added, which does some import module as inititalization function. This is taken from the problem described in #2110 that led to #2140. I'm not sure the test on import is a very stable solution.
- the function names I used, not really conviced about `add_preload_function` or `register_preload_function`, any suggestion welcome.
- the use of coroutine and pickling, which I used when needed, but maybe not correctly.


distributed/scheduler.py
```diff
@@ -940,6 +942,7 @@ def __init__(
             'heartbeat_worker': self.heartbeat_worker,
             'get_task_status': self.get_task_status,
             'get_task_stream': self.get_task_stream,
+            'add_preload_function': self.register_preload_function
```
dhirschfeld: I think that whatever is chosen should be named the same on the client and the scheduler - I don't see any benefit to coming up with two different names for this?

For me I associate the functionality with *initializing* a worker whereas `preload` seems a bit more ambiguous - `pre-` what? Given that I think `register_init_func` is a better name.

guillaumeeb:
As expected, testing if imports are already done is not a safe test: `matplotlib` is not imported in worker process, but `bokeh` already is there (was not when running only the new test locally), which cause the test failure. 
I changed second test which is now affecting os.environ. I leave an import test as this was what triggered the initial issue, but complexified it a little by importing submodule.

If anyone has some better proposition, I will be glad to change that.

distributed/scheduler.py
guillaumeeb: Thanks for the comment, I did use two different names just to be sure I understood what was happening behind 🙂, just changed that and adopted your suggestion.

distributed/tests/test_worker.py
```diff
@@ -1198,3 +1198,73 @@ def test_avoid_oversubscription(c, s, *workers):
...
+
+    def mystartup2():
+        import os
+        os.environ['MY_ENV_VALUE'] = 'WORKER_ENV_VALUE'
+        return "Env set."
+
+    #Check that preload function has been run
+    def test_import():
+        import sys
+        return 'matplotlib.bezier' in sys.modules.keys()
```
dhirschfeld: *Aside*: `'matplotlib.bezier' in sys.modules` should be sufficient. Prob wouldn't bother changing it though unless you're making other changes also.

distributed/client.py
```diff
@@ -3390,6 +3390,26 @@ def get_task_stream(self, start=None, stop=None, count=None):
...
+        """
+        Registers and run an initialization for every worker
+
+        This registers a function to be called by every currently connected
+        workers, and that will be also called by any added worker after that.
+        Multiple functions can be added by multiple calls.
+
+        Parameters
+        ----------
+        function: Function to register and run
```
dhirschfeld: This should be `name : type` with the description on the line below. For me, and particularly with functions I find it very useful to include the signature - e.g.
```
        Parameters
        ----------
        function : callable() -> None
            Function to register and run
```
Also note the space around the `:` - at least in spyder it won't render correctly without that.

dhirschfeld:
Sorry about the small nits - they're mostly irrelevant in the scheme of things so feel free to ignore - I'm just a lurker and not a maintainer!

distributed/scheduler.py
```diff
@@ -3013,6 +3017,24 @@ def get_task_stream(self, comm=None, start=None, stop=None, count=None):
...
+        self.init_functions.append(loads(function))
+
+        responses = yield self.broadcast(msg=dict(op='run',
+                                                  function=function))
+        results = {}
+        for key, resp in responses.items():
+            if resp['status'] == 'OK':
+                results[key] = resp['result']
+            elif resp['status'] == 'error':
+                six.reraise(*clean_exception(**resp))
```
mrocklin: It would be good to test the exception behavior

```python
with pytest.raises(ZeroDivisionError):
    yield client.register_initialization(lambda: 1 / 0)
```

distributed/scheduler.py
```diff
@@ -3013,6 +3017,24 @@ def get_task_stream(self, comm=None, start=None, stop=None, count=None):
         ts = [p for p in self.plugins if isinstance(p, TaskStreamPlugin)][0]
         return ts.collect(start=start, stop=stop, count=count)
 
+    @gen.coroutine
+    def register_init_func(self, comm, function=None):
+        """ Registers a preload function, and call it on every worker """
+        if function is None:
+            raise gen.Return({})
+
+        self.init_functions.append(loads(function))
```
mrocklin: If possible, I think it would be better to never call `pickle.loads` in the scheduler.  The scheduler tries to never actually touch real Python code.  Instead, I would store the serialized bytes in the scheduler and then pass those on directly.  

I'll also offer a suggestion in a comment next to the client code.

distributed/client.py
```diff
@@ -3390,6 +3390,26 @@ def get_task_stream(self, start=None, stop=None, count=None):
         return self.sync(self.scheduler.get_task_stream, start=start,
                          stop=stop, count=count)
 
+    @gen.coroutine
+    def _register_init_func(self, function):
+        response = yield self.scheduler.register_init_func(function=dumps(function))
+        raise gen.Return(response)
```
mrocklin: Instead of serializing explicitly with pickle I recommend using `yield self.scheduler.register_init_func(function=to_serialize(function))`

I think that this is all that you will need to do.  Dask should handle serialization to the worker for you.  I should probably change this in `run` as well.

distributed/client.py
```diff
@@ -3390,6 +3390,26 @@ def get_task_stream(self, start=None, stop=None, count=None):
...
+        This registers a function to be called by every currently connected
+        workers, and that will be also called by any added worker after that.
+        Multiple functions can be added by multiple calls.
+
+        Parameters
+        ----------
+        function: Function to register and run
+        """
+        return self.sync(self._register_init_func, function,
+                         asynchronous=asynchronous)
```
mrocklin: This logic is simple enough that we probably don't need a separate coroutine, but can instead probably call the scheduler method directly here:

    return self.sync(self.scheduler.register_init_func, ...

You can then drop the `_register_init_func` method entirely.

distributed/worker.py
```diff
@@ -326,6 +326,11 @@ def _register_with_scheduler(self):
                 middle = (_start + _end) / 2
                 self.scheduler_delay = response['time'] - middle
                 self.status = 'running'
+
+                # Retrieve eventual init functions and run them
+                for init_function in pickle.loads(response['init-functions']):
+                    result = init_function()
```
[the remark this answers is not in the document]
dhirschfeld: Could we just not make the signature `callable(dask_worker: Worker) -> None` in all cases? It doesn't seem to me to be much of a burden to the writer of an initialization function to mandate that it takes the worker instance as an argument whether or not the function actually uses it.

distributed/tests/test_worker.py
mrocklin: Note that all of the workers are in this process, so these don't currently test what happens when we get a new worker.  I think that this can be resolved by having the function modify the worker itself.

distributed/worker.py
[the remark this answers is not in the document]
guillaumeeb: > There is also a pre-existing convention for this in client.run which has proven useful.

I don't see where it is located. It could help me do implement this part for already connected workers.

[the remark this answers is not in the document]
mrocklin: Search for `dask_worker` in `worker.py`.  It's under the `def run(...)` function.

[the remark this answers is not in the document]
guillaumeeb: Thanks, was not looking into the correct run() method...

distributed/client.py
guillaumeeb: As I use `broadcast` under the hood with `run` coroutine called, this actually don't work. I probably should modify the `run` function in worker.py, but then I need to modify it in `Client.run` too. Isn't it out of scope of this PR?

Stacktrace is as follow, happens when calling `client.register_init_func`:
````
distributed.core - ERROR - 'function' object is not subscriptable
  File "/home/eh/eynardbg/git/distributed/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
TypeError: a bytes-like object is required, not 'function'

  File "/home/eh/eynardbg/git/distributed/distributed/core.py", line 340, in handle_comm
    result = yield result
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/asyncio/futures.py", line 294, in result
    raise self._exception
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/tornado/gen.py", line 326, in wrapper
    yielded = next(result)
  File "/home/eh/eynardbg/git/distributed/distributed/worker.py", line 1025, in run
    function = pickle.loads(function)
  File "/home/eh/eynardbg/git/distributed/distributed/protocol/pickle.py", line 61, in loads
    logger.info("Failed to deserialize %s", x[:10000], exc_info=True)
TypeError: 'function' object is not subscriptable
distributed.core - ERROR - 'function' object is not subscriptable
  File "/home/eh/eynardbg/git/distributed/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
TypeError: a bytes-like object is required, not 'function'
````

Comment (mrocklin) [reply distributed/client.py]: OK, lets not use `to_serialize` then.  You should still be able to drop loads/dumps from within the scheduler though and just store serialized bytes.

File: distributed/scheduler.py (hunk 4 above)
Comment (guillaumeeb) [reply distributed/scheduler.py]: Added the test, but it don't run well. Stack trace is:
````
distributed.worker - INFO - Run out-of-band function '<lambda>'
distributed.worker - WARNING -  Run Failed
Function: <lambda>
args:     ()
kwargs:   {}
  File "/home/eh/eynardbg/git/distributed/distributed/worker.py", line 1036, in run
    result = function(*args, **kwargs)
  File "/home/eh/eynardbg/git/distributed/distributed/tests/test_worker.py", line 1273, in <lambda>
    yield c.register_init_func(lambda: 1 / 0)
distributed.worker - INFO - Run out-of-band function '<lambda>'
distributed.worker - WARNING -  Run Failed
Function: <lambda>
args:     ()
kwargs:   {}
  File "/home/eh/eynardbg/git/distributed/distributed/worker.py", line 1036, in run
    result = function(*args, **kwargs)
  File "/home/eh/eynardbg/git/distributed/distributed/tests/test_worker.py", line 1273, in <lambda>
    yield c.register_init_func(lambda: 1 / 0)
distributed.core - ERROR - 'Serialized' object has no attribute '__traceback__'
  File "/home/eh/eynardbg/git/distributed/distributed/core.py", line 340, in handle_comm
    result = yield result
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/asyncio/futures.py", line 294, in result
    raise self._exception
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/tornado/gen.py", line 1147, in run
  File "/home/eh/eynardbg/git/distributed/distributed/scheduler.py", line 3036, in register_init_func
    six.reraise(*clean_exception(**resp))
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/six.py", line 691, in reraise
    if value.__traceback__ is not tb:
AttributeError: 'Serialized' object has no attribute '__traceback__'

````

File: distributed/client.py (hunk 3 above)
Comment (guillaumeeb) [reply distributed/client.py]: Done.

File: distributed/worker.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (guillaumeeb) [reply distributed/worker.py]: Even if I put the signature with dask_worker in the docstring, I kept @mrocklin suggestion to be robbust to simple case where the function does'nt declare dask_worker.

File: distributed/scheduler.py (hunk 5 above)
Comment (guillaumeeb) [reply distributed/scheduler.py]: Done.

File: distributed/client.py (hunk 6 above)
Comment (guillaumeeb) [reply distributed/client.py]: Yep, that part with scheduler is OK.

File: distributed/client.py (hunk 7 above)
Comment (guillaumeeb) [reply distributed/client.py]: Done too.

File: distributed/scheduler.py (hunk 4 above)
Comment (guillaumeeb) [reply distributed/scheduler.py]: So here I will need some help, I really don't know what's wrong!

Comment (mrocklin) [reply distributed/scheduler.py]: I'll take a look

Comment (mrocklin) [reply distributed/scheduler.py]: So it's similar to the other situation with pickle.  The scheduler doesn't deserialize data by default.  Instead, it's better to pass along the results to the client and raise the exception there.  This is good for a few reasons:

1.  You want the exception to raise on the client anyway, because that is where the user is.
2.  The scheduler may not have the same versions as the client and workers (which must be the same) so deserializing the exception here may result in another error
3.  In the future the scheduler may be written in another language, so the exceptions may not make sense
4.  We tend to treat the scheduler with a greater sense of security, and so don't deserialize data here.

The object you're running into, `Serialized`, is something that holds onto serialized bytes and passes them along.  By default all of the communication channels running into the scheduler don't deserialize their data, and instead pass along these `Serialized` objects.

So you probably want to return whatever `Scheduler.broadcast` gives you and hand it back to the Client.  On the client side you'll want to handle the exception, much like is done within the `Client._run` method.  Now that we're doing more logic we may now actually need separate coroutine and blocking methods (like `_register_init_func` and `register_init_func`).  My apologies for pushing you away from those earlier.

Comment (guillaumeeb) [reply distributed/scheduler.py]: Thanks for the explanation, very helpful, fixed now.

Not sure I understand all the subtleties of subroutine and passing args or kwargs around, bu it will come with more practice!

File: distributed/worker.py [hunk 9]
```diff
@@ -332,6 +332,15 @@ def _register_with_scheduler(self):
...
                 self.status = 'running'
+
+                # Retrieve eventual init functions and run them
+                for init_function_bytes in response['init-functions']:
+                    init_function = pickle.loads(init_function_bytes)
+                    if has_arg(init_function, 'dask_worker'):
+                        result = init_function(dask_worker=self)
+                    else:
+                        result = init_function()
+                    logger.info('Init function %s ran: output=%s' % (init_function, result))
```
mrocklin: It looks like this code fails under the test `py.test distributed/tests/test_scheduler.py::test_worker_name`

mrocklin:
Regarding names how about something like the following:

-  `Client.register_worker_setup` / `Scheduler.worker_setups`
-  `Client.register_worker_initialization` / `Scheduler.worker_initializations`

I would like to avoid abbreviations like `init` and `func` if possible for clarity.  This does make things longer, but these methods are not likely to be called frequently, so I think that a long method name probably isn't that bad.

I would like to avoid the term "function" because I anticipate that in the future we will use this same API point to also accept classes with `setup`/`teardown` methods and it would be nice to future-proof.

Thoughts?  There are probably better names than what I suggested above.

dhirschfeld:
Since `__init__` is used to initialize every python class I think the intention is pretty clear despite the abbreviation but it's also not a big deal to use the longer `initialization`.

guillaumeeb:
Coming from Java, I'm really fine with long method names. At first glance, I would have preferred `initialization`, but since it is meant to also propose a `teardown` or equivalent function, `setup` / `teardown` looks better. I don't see what to put as opposite as `initialization`.

distributed/worker.py
guillaumeeb: Yes thanks, I ran the init functions even if scheduler response was not OK, I've move the init code.

mrocklin:
`register_worker_context`?  I think that a previous attempt at this called it an `environment` (though that might conflict with software environments).  

I'll give this some thought.  Suggestions would be welcome.

guillaumeeb:
I definitly prefer `register_worker_setup`.

guillaumeeb:
No idea of what causes the remaining error...

guillaumeeb:
Thanks @mrocklin, this was quite simple after all :).

mrocklin:
There are some intermittent failures in our CI.   Some of them I've cleaned up in master.  Some still exist.  My apologies for the confusion.

mrocklin:
> I definitly prefer `register_worker_setup`.

That works for me

mrocklin:
> Random failure in appveyor?

Perhaps.  I've restarted it.

> As discussed in #2140, should I also implement the register_teardown here?

I recommend that we merge this in and then consider a few options for teardown.

Also, cc @jcrist who implemented something similar a while ago.  He may have thoughts on the API.

mrocklin:
Also cc @lesteve , who I think might want to use this in combination with #2169

lesteve:
Nice! I guess that the main advantage of using `client.register_worker_setup` compared that `client.run` is when using adaptative.

mrocklin:
This looks good to me.  I've pinged @jcrist to give feedback on the API.  Otherwise I think that this is ready to go in.

distributed/client.py
```diff
@@ -3390,6 +3390,34 @@ def get_task_stream(self, start=None, stop=None, count=None):
...
+                                                           kwargs=dumps(kwargs))
+        results = {}
+        for key, resp in responses.items():
+            if resp['status'] == 'OK':
+                results[key] = resp['result']
+            elif resp['status'] == 'error':
+                six.reraise(*clean_exception(**resp))
+        raise gen.Return(results)
+
+    def register_worker_setup(self, function, *args, **kwargs):
```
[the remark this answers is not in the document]
guillaumeeb: Actually, I had to add this in order to be able to make things work with `dask_worker` optional keyword, which is filled on worker side. I don't really know why this is needed, as you say, the initial intent on client side is to only pass the function, and we don't register argument. So if someone can explain me what to do here, please do!

distributed/client.py
```diff
@@ -3390,6 +3390,34 @@ def get_task_stream(self, start=None, stop=None, count=None):
...
+        """
+        Registers and run an setup function for every worker
+
+        This registers a function to be called by every currently connected
+        workers, and that will be also called by any added worker after that.
+        Multiple functions can be added by multiple calls.
```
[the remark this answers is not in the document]
guillaumeeb: Thanks, that's perfect, and added.

guillaumeeb:
> - If you're adding multiple setup/teardown/etc... callbacks, perhaps these would be better enclosed inside a single api call (...)

You are probably right. Should we do this in this PR, or wait for another one with teardown support?

> - Would you ever want to unregister or query a callback? If so, should they be provided a name that can reference them (e.g. register_worker_context(name='database_connection', setup=setup, teardown=teardown)).

I have no strong opinion on this, it could probably happen...

distributed/client.py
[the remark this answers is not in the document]
jcrist: I can't think of a use case for the function taking arguments beyond the worker, and that complicates things (perhaps unnecessarily). If it's truly meant to be a setup/teardown function, it should have state fully enclosed. I'd prefer the setup functions be forced to have the signature `setup()` or `setup(dask_worker)`, rather than arbitrary signatures

I read

> Actually, I had to add this in order to be able to make things work with dask_worker optional keyword, which is filled on worker side.

as meaning the args/kwargs were needed somehow to play well with distributed's internals. I'm not sure why/how that would be, @mrocklin would know more.

jcrist:
> You are probably right. Should we do this in this PR, or wait for another one with teardown support?

I'd rather not add api and then change it later. If a method like:

```
register_worker_callbacks(self, setup=None)
```

was added here, then it could be expanded like

```
register_worker_callbacks(self, setup=None, teardown=None)
```

in a subsequent PR. I'm not sure if that's the best api though, no strong thoughts.

distributed/client.py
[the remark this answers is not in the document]
mrocklin: OK, I'm fine with keeping things constrained to just the single possible argument.  I'm also ok with forcing it as the input (all setup functions must accept a single argument).  

It's easy to expand these things later if we find that they're necessary.

mrocklin:
+1 on @jcrist 's suggestion above.

distributed/client.py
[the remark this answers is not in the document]
guillaumeeb: Not sure what the conclusion is here. Force dask_worker arg in fonction signature and no other args? Currently no args are stored in the scheduler context.

How should the register_worker_setup function signature look like on client side?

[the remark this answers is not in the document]
mrocklin: I think that the decisions is to not expose `*args` or `**kwargs` in the client or scheduler code yet.  

We should check the function for the `dask_worker` parameter in the worker code and apply it there if it exists.

guillaumeeb:
Okay, so no optional args here, only dask_worker as a required argument to the functions.

distributed/scheduler.py
```diff
@@ -3035,6 +3038,18 @@ def get_task_stream(self, comm=None, start=None, stop=None, count=None):
...
+    @gen.coroutine
+    def register_worker_callbacks(self, comm, setup=None):
+        """ Registers a setup function, and call it on every worker """
+        if setup is None:
+            raise gen.Return({})
+
+        self.worker_setups.append(setup)
+
+        responses = yield self.broadcast(msg=dict(op='run', function=setup,
+                                                  args=(), kwargs={}))
```
guillaumeeb: Not sure why, but if I don't force `args=(), kwargs={}` here, and don't use `*args, **kwargs` signature for the `register_worker_callbacks` to propagate them as before, I got the following exception when running `register_worker_callbacks`:

````
  File "/home/eh/eynardbg/git/distributed/distributed/core.py", line 340, in handle_comm
    result = yield result
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/asyncio/futures.py", line 294, in result
    raise self._exception
  File "/work/ADM/hpc/eynardbg/Outils/miniconda3/envs/dask_dev/lib/python3.5/site-packages/tornado/gen.py", line 326, in wrapper
    yielded = next(result)
  File "/home/eh/eynardbg/git/distributed/distributed/worker.py", line 1036, in run
    kwargs = pickle.loads(kwargs)
  File "/home/eh/eynardbg/git/distributed/distributed/protocol/pickle.py", line 61, in loads
    logger.info("Failed to deserialize %s", x[:10000], exc_info=True)
TypeError: unhashable type: 'slice'
````

Investigating, it seems like a kwargs contains a dask-worker key and is a plain non serialized dict. Really don't know how it gets passed to `worker.run()` method. That was the problem I tried to describe [here](https://github.com/dask/distributed/pull/2201#discussion_r213552607).

**Comment by mrocklin** [comment]
OK.  Merging this in.  Thanks @guillaumeeb !

Also, I notice that this is your first commit to this repository.  I'm glad to have your eyes on this code !

**Comment by guillaumeeb** [comment]
Thanks to you! I'm glad to finally contribute directly to this amazing tool.

**Comment by rainwoodman** [comment]
I noticed if I have two clients connected to the scheduler that ran register_worker_callbacks on the 'same' function, then I end up having two callback ran every time a worker is reconnected.

It doesn't feel like a desired behavior, as the setup callback is likely doing some module level  initialization that cloudpickle can miss.

I think we can add a 'label' argument to allow registering callback only once:

```
register_worker_callbacks(setup=func, key="label")
```
key will be published on the scheduler. If "label" is already registered on the scheduler (can be implemented with `publish` or `persist`, I suppose) then this function shall do nothing.

This also eases removal of callbacks, e.g.

```
remove_worker_callbacks(['setup'], key="label")
```

**Comment by mrocklin** [comment]
cc @lesteve , who I believe implemented that functionality originally.

**Comment by mrocklin** [comment]
Sounds good.  @rainwoodman do you have any interest in implementing this?

FWIW I also slightly prefer renaming key= to name=

**Comment by rainwoodman** [comment]
Sure. Let's do the `add` first. 

Here is my current implementation:

https://github.com/bccp/nbodykit/blob/mpiself/nbodykit/__init__.py#L76

The implementation will block and relies on the uniqueness of dataset names. 
I noticed the current implementation of `register_worker_callbacks` are mostly inside a few async functions.

Would this be a problem?

**Comment by rainwoodman** [comment]
I started working on this in #2391.

I think we do not need to use a lock or publish a global dataset for this to work, as the scheduler will hold its own list of callback functions.  I do have to refine the semantics of the setup callback a bit ( and I think it is still reasonable):

1. A setup callback function shall function properly if it is called several times. We shall always broadcast the setup callback function to all existing workers even if the provided name has already been registered, in order to maintain a consistent state between existing workers and newly joined workers. This means the following code shall work:

```
On Client 1:
client.register_worker_callback(setup=mylib.dask_worker_init, name='mylib.dask_worker_init')
```

```
Client 2:
client.register_worker_callback(setup=mylib.dask_worker_init, name='mylib.dask_worker_init')
```

Then `mylib.dask_worker_init` will be triggered twice on existing workers. It will be triggered only once on newly connected workers, using the version of `mylib.dask_worker_init` from whichever client that arrived the most recently to the scheduler. Since we require client and worker run the same/compatible version of software stack, when the versions are incompatible we can leave the behavior undefined.

2. I think we shall also guarantee the callback is not called concurrently by multiple threads on the same worker. I do not know if the current implementation (via broadcast and worker initialization) guarantees this. I highly suspect not. 

3. Upon a new worker's initialization, only the latest version of any named setup callback function is called. (one trigger per name) For the functions of different names, the order shall be undefined. 

The motivation for this change is I do not think we currently correctly provide the current guarantee on the triggering order of callback functions -- the time the request arriving to the scheduler will not always match the order the requests were sent; besides, the ordering guarantee is not useful when the requests are from multiple clients; especially if we provide 1).

mrocklin:
I'm curious about calling the same setup function twice.  I'm somewhat inclined to assume idempotence and, if the functions are the same (as defined by equality of their serialized form on the scheduler) then we don't bother asking workers who have run them before to run them again.  Thoughts?

guillaumeeb:
I think the simplest is to assume indempotence, but to always run the function registered on all workers when added.

For me @rainwoodman 3 points above are somewhat implicits defaults:

1. Our implementation assumes indempotence, so calling a function twice is OK.
1. I don't see why the calback would be called by multiple threads ?
1. We should only keep one function version givena name, so it will be triggered once.

I don't thinks this needs any change, does it?

PR #2202, opened 2018-08-21
Explicitly test worker dep waiting->memory transition

mrocklin:
Follows on to #2182 

This doesn't yet actually test the waiting-memory transition state.  I'm having trouble finding a way to touch that code.


quasiben:
This test is consistently failing on py37/tornado 6 in TravisIC on the test in this PR.  I am not sure why.  Locally on linux/mac this tests passes.

hendrikmakait:
I think this is stale by now.

PR #2204, opened 2018-08-21
Fix formatting when port is a tuple

lesteve:
I stumble on this while looking at a dask-jobqueue related issue. `port` can be a tuple as the code a few lines above testifies:

```py
try:
    service = v(self, io_loop=self.loop, **kwargs)
    if isinstance(port, tuple):
        service.listen(port)
    else:
        service.listen((listen_ip, port))
    self.services[k] = service
except Exception as e:
    warnings.warn("\nCould not launch service '%s' on port %s. " % (k, port) +
                    "Got the following message:\n\n" + str(e),
                    stacklevel=3)
```

On a side-note, I am always surprised to see people using `%` formatting (I personally find it annoying to have to think about which type the arguments are and then you have less than great error messages when you get it wrong ...), but that's completely fine of course ;-)

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1051,7 +1051,7 @@ class Scheduler(ServerNode):
                     service.listen((listen_ip, port))
                 self.services[k] = service
             except Exception as e:
-                warnings.warn("\nCould not launch service '%s' on port %d. " % (k, port) +
+                warnings.warn("\nCould not launch service '%s' on port %s. " % (k, port) +
                               "Got the following message:\n\n" + str(e),
                               stacklevel=3)
 


mrocklin:
> On a side-note, I am always surprised to see people using % formatting

I'm probably guilty of this.  Which style do you recommend?

lesteve:
I find `.format` works fine for me but I have heard some people did not like it (too verbose is a complaint I seem to remember somehow). I have to say that when I use `%`, I almost always use `%s`, because:
* I don't want to have to think about the types of my input
* I have to admit I don't remember off the top off my head the format specifier
* I have still to encounter a case where erroring during the formatting because of a type mismatch is more helpful than later on in the code (if you error later on in the code at least you get the warning/print/log statement with the value of the problematic variable)

PR #2209, opened 2018-08-24
Scale by number of cores and memory

### LINKED ISSUE #2208: Scale by number of cores or amount of memory
When creating a cluster object we currently scale by number of workers

    cluster = KubeCluster()
    cluster.scale(10)

Where `10` is the number of workers we want to have.  However, it is common for users to think about clusters in terms of number of cores or amount of memory, rather than in terms of number of dask workers

    cluster.scale(cores=100)
    cluster.scale(memory='1 TB')

What is the best way to achieve this uniformly across the dask deployment projects?  I currently see two approaches, though there are probably more that others might see.

1.  Establish a convention where clusters define information about the workers they will produce, something like the following:

    ```python
    >>> cluster.worker_info
    {'cores': 4, 'memory': '16 GB'}
    ```

    Then the core `Cluster.scale` method would translate this into number of workers and then call the subclass's `scale` method appropriately

2.  Let the downstream classes handle this themselves, but ask them all to handle it uniformly.  This places more burden onto downstream implementations, but also gives them more freedom to select worker types as they see fit based on their capabilities.

cc

-  @guillaumeeb (who has shown interest in doing this work) @lesteve @jhamman from `dask-jobqueue`
-  @jcrist from `dask-yarn`
-  @jacobtomlinson from `dask-kubernetes`

guillaumeeb:
Closes #2208.

Just an initial commit to show how it could look. An associated PR incomming on dask-jobqueue.
There work left with the following tasks:

 - [x] Add some test,
 - [x] Rework the widget?

Widget rework to be discussed.
Reviews welcome.


guillaumeeb:
Also ping @jcrist and @jacobtomlinson

mrocklin:
Given planned churn in this module and the number of other packages that this would impact I'm inclined to wait on this until a future redesign.  I wonder if changes should be isolated to downstream packages (like dask-jobqueue) until then?  I don't have particularly strong thoughts on this though, I haven't reviewed this in detail.

guillaumeeb:
This can certainly be isolated into dask-jobqueue, I will try to have a look and see if this can be done elegantly without just copy pasting and overriding `scale` or `adapt` method. I've not yet dealt a ot with super() calls in Python yet.

quasiben:
I believe recent SpecCluster work addressed scaling cores/memory

PR #2211, opened 2018-08-25
Describe what ZeroMQ is

mivade:
Description taken from https://en.wikipedia.org/wiki/ZeroMQ.


mrocklin:
Yup.  That's a great change.  Thank you.

PR #2233, opened 2018-09-05
as_completed touchups

### LINKED ISSUE #2065: as_completed using with_results option hangs getting next future if future errored
I ran into an issue where failed tasks were causing my code to hang waiting for `as_completed` to return the next result.  Here's an example of the scenario: 
```
from dask.distributed import as_completed


client = Client()

def error_out():
    raise RuntimeError()

future = client.submit(error_out)
iterator = as_completed([future], with_results=True)
iterator.next()
```

### LINKED ISSUE #2089: as_completed drops cancelled futures instead of treating them as completed
Dask's as_completed started dropping cancelled futures from PR #1366 . The original intent was to mirror concurrent.futures as_completed api, but it seems (from the docs, worth confirming) that concurrent.futures includes cancelled futures in the list returned. See https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed it says "Returns an iterator over the Future instances (possibly created by different Executor instances) given by fs that yields futures as they complete (finished or were cancelled)."

Discussion about the behavior took place in #2039 

The desired behavior is that successful, error and cancelled futures are all returned by as_completed in the order they reach that state which mirrors the concurrent.futures behavior (according to the docs!)

cicdw:
Attempting to get my feet wet contributing

- Closes #2089 
- Closes #2065

Not sure if this is issue worthy or not, but the following tests fail for me locally (and are unrelated to my changes):
- `distributed/cli/tests/test_dask_scheduler.py::test_pid_file` (times out)
- `distributed/deploy/tests/test_local.py::test_duplicate_clients` (doesn't warn)
- `distributed/cli/tests/test_dask_worker.py::test_nprocs_requires_nanny` (intermittently times out)

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3631,13 +3631,10 @@ class as_completed(object):
             if not self.futures[future]:
                 del self.futures[future]
             if self.with_results:
-                self.queue.put_nowait((future, result))
+                self.queue.put_nowait((future, result, future.status == 'error'))
             else:
                 self.queue.put_nowait(future)
             self._notify()
-        if all([self.with_results, self.raise_errors,
-                future.status == 'error']):
-            six.reraise(*result)
 
     def update(self, futures):
         """ Add multiple futures to the collection.
@@ -3677,13 +3674,22 @@ class as_completed(object):
     def __aiter__(self):
         return self
 
+    def _get_and_raise(self):
+        if self.with_results:
+            future, result, erred = self.queue.get()
+            if self.raise_errors and erred:
+                six.reraise(*result)
+            else:
+                return future, result
+        return self.queue.get()
+
     def __next__(self):
         while self.queue.empty():
             if self.is_empty():
                 raise StopIteration()
             with self.thread_condition:
                 self.thread_condition.wait(timeout=0.100)
-        return self.queue.get()
+        return self._get_and_raise()
 
     @gen.coroutine
     def __anext__(self):
@@ -3694,7 +3700,7 @@ class as_completed(object):
                 raise StopAsyncIteration
             yield self.condition.wait()
 
-        raise gen.Return(self.queue.get())
+        raise gen.Return(self._get_and_raise())
 
     next = __next__
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3610,10 +3610,7 @@ class as_completed(object):
         try:
             yield _wait(future)
         except CancelledError:
-            del self.futures[future]
-            if not self.futures:
-                self._notify()
-            return
+            pass
         if self.with_results:
             result = yield future._result()
         with self.lock:
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3631,7 +3631,7 @@ class as_completed(object):
             if not self.futures[future]:
                 del self.futures[future]
             if self.with_results:
-                self.queue.put_nowait((future, result, future.status == 'error'))
+                self.queue.put_nowait((future, result))
             else:
                 self.queue.put_nowait(future)
             self._notify()
@@ -3675,13 +3675,12 @@ class as_completed(object):
         return self
 
     def _get_and_raise(self):
+        res = self.queue.get()
         if self.with_results:
-            future, result, erred = self.queue.get()
-            if self.raise_errors and erred:
+            future, result = res
+            if self.raise_errors and future.status == 'error':
                 six.reraise(*result)
-            else:
-                return future, result
-        return self.queue.get()
+        return res
 
     def __next__(self):
         while self.queue.empty():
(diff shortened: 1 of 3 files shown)


distributed/client.py
```diff
@@ -3625,6 +3635,9 @@ def track_future(self, future):
             else:
                 self.queue.put_nowait(future)
             self._notify()
+        if all([self.with_results, self.raise_errors,
+                future.status == 'error']):
```
mrocklin: I have a slight style preference for two `and`s rather than one `all`.  Going over 80 characters is fine.

distributed/tests/test_as_completed.py
```diff
@@ -168,3 +170,39 @@ def test_async_for_py2_equivalent(c, s, a, b):
...
+            x = c.submit(throws, 1)
+            y = c.submit(sleep, 1)
+            z = c.submit(inc, 1)
+
+            ac = as_completed([x, y, z], with_results=True)
+            y.cancel()
+            res = list(ac)
+
+            futs, results = zip(*res)
+            assert futs == (y, x, z)
```
mrocklin: How confident are we of this ordering?  We might also pick out x, y, and z based on their keys.

distributed/tests/test_as_completed.py
```diff
@@ -168,3 +170,39 @@ def test_async_for_py2_equivalent(c, s, a, b):
...
+            assert x.status == 'error'
+            assert y.status == 'cancelled'
+            assert z.status == 'finished'
+
+            assert isinstance(results[0], CancelledError)
+            assert isinstance(results[1][1], RuntimeError)
```
mrocklin: Given that the default to `raise_errors` is True I'm surprised that we didn't get an exception above.  Why is this?

distributed/tests/test_as_completed.py
cicdw: Good question - I can update so it's order independent.

distributed/tests/test_as_completed.py
cicdw: The exception gets raised within the `track_future` coroutine, and is consequently trapped by `tornado`; I had played with raising in `__next__` / `__anext__` so that it would actually surface to the top, but couldn't figure out how to manage [keeping it within a coroutine](https://github.com/dask/distributed/issues/2065#issuecomment-402778890) at the same time.  

The only solution I had come up with was:
```python
    def _get(self):
        res = self.queue.get()
        if self.with_results:
            fut, result = res
            assert sync(self.loop, fut._result, raiseit=False) == result
            six.reraise(*result)
        return res

    def __next__(self):
        ...
        return self._get()
```

but I'm not sure how robust that is, or whether it even solves the race condition problem?

mrocklin: Maybe we pass a boolean `erred` in the tuple that we put into `self.queue`.  Whenver we get from that queue and we see that erred is True then we raise if self.raise_errors.

```python
def __next__(self):
    ...
    future, result, erred = self.queue.get()
    if erred and self.raise_errors:
        raise result
```

cicdw: Should I worry about adjusting `__anext__`?

mrocklin: Yes, you should probably do the same on both.  The async style tests, the ones that use `@gen_cluster` will cover the `__anext__` method.

distributed/client.py
```diff
@@ -3610,18 +3623,15 @@ def track_future(self, future):
...
         if self.with_results:
-            result = yield future._result()
+            result = yield future._result(raiseit=False)
         with self.lock:
             self.futures[future] -= 1
             if not self.futures[future]:
                 del self.futures[future]
             if self.with_results:
-                self.queue.put_nowait((future, result))
+                self.queue.put_nowait((future, result, future.status == 'error'))
```
mrocklin: Ah, your solution here points to the fact that including the third element of the tuple is unnecessary.  We can compute it from the first.  My apologies for leading you astray.  I would compute the `erred` value just in the `_get_and_raise` method as you've done it here.

cicdw: Ah ok, no worries!

distributed/tests/test_as_completed.py
```diff
@@ -168,3 +170,107 @@ def test_async_for_py2_equivalent(c, s, a, b):
...
+        y = c.submit(sleep, 1)
+
+        ac = as_completed([x, y])
+        first = yield ac.__anext__()
+        second = yield ac.__anext__()
+        result = [first, second]
+
+        assert result == [x, y]
+        assert x.status == 'error'
+        assert y.status == 'finished'
```
mrocklin: Two requests:

1.  Lets replace `sleep` by `inc` throughout in order to keep these tests fast if possible.  This will probably require making the tests insensitive to ordering issues as you did before
2.  Lets unindent the `gen_cluster` style tests one level

Otherwise things here look good to me.

mrocklin:
This is in.  Thank you for your contribution @cicdw , and welcome as a code contributor to the dask/distributed repository!

PR #2213, opened 2018-08-27
Dludwig/css grid

dsludwig:
Another attempt at #2150, using CSS grid to layout, and without Bootstrap or other external libraries.

PR #2239, opened 2018-09-06
Display system graphs nicely on different screen sizes

dsludwig:
A followup to #2213 - this makes the system graph page scale nicely.

screen shot 2018-09-05 at 6 03 13 pm
screen shot 2018-09-05 at 6 03 23 pm

This one is more amenable to flexbox, especially since there can be variable numbers of graphs here (well, 3 on Windows, 4 otherwise).

Thanks @canavandl for cleaning up the CSS into separate files!


mrocklin:
Thanks @dsludwig !

Issue #2126, opened 2018-07-18
KeyError on release / exception never retrieved


adamklein:
On lastest distributed master, there's some sort of race condition where I'm getting

```
[16:03:33.254 ERROR  ] Future <tornado.concurrent.Future object at 0x7fd1e96ae080> exception was never retrieved: Traceback (most recent call last):
  File "lib/python3.5/site-packages/tornado/gen.py", line 307, in wrapper
    yielded = next(result)
  File "src/distributed/distributed/worker.py", line 590, in delete_data
    self.release_key(key)
  File "src/distributed/distributed/worker.py", line 2016, in release_key
    self.dependents[dep].remove(key)
KeyError: '<task>'
```

Stemming from

            for dep in self.dependencies.pop(key, ()):
                if dep in self.dependents:
                    self.dependents[dep].remove(key)  # <--- here
                    if not self.dependents[dep] and self.dep_state[dep] in ('waiting', 'flight'):
                        self.release_dep(dep)

Cannot repro it simply yet. But the behavior is that K2 depends on K1;  K2 is being released; but K1 does not have K2 as recorded in its dependents set (instead, it has a K2' which *ought* to be K2, but has a different hash suffix).

mrocklin:
A reproducer would be welcome.  You might also want to look at the story of
the keys

    w.story(key, dep)

adamklein:
I see what's happening via print-debugging but still haven't reproduced simply yet.  Here's the gist:

```
1) submit task A, submit task B that depends on A
2) (re)submit task A, submit task C that depends on A 
 ... at this point, dependents[A] = {B, C} ...
3) task A is released via delete_data -> release_dep -> dependents.pop(A) => dependents[A] = {}
4) (re)submit task A, submit task D that depends on A
... at this point, dependents[A] ={D}
5) release_key(B) -> calls problematic logic
  for dep in self.dependencies[B]:   # gives A
      self.dependents[dep].remove(B)   # cannot remove B, because {B, C} were removed in step 3
```

adamklein:
Ok, I have a repro (finally). Every few iterations this will yield the KeyError:

```
import time

from distributed import LocalCluster, Client, worker_client, get_worker


def inc_func(x):
    return x + 1


def add_func(x, y):
    return x + y


def on_worker():
    with worker_client() as wc:
        pins = []
        addr = get_worker().address

        for k in range(8):
            fut = wc.submit(inc_func, 0, workers=[addr], allow_other_workers=False)
[...]

PR #2250, opened 2018-09-14
Discard dependent rather than remove

mrocklin:
There are some cases where Worker dependencies and dependents can get out of
sync, especially when we lose dependencies but the dependents are still around.

To reflect this we replace a set.remove with set.discard

Fixes https://github.com/dask/distributed/issues/2126

There is a test in #2126 but I was unable to turn it into something that could run happily in small time.


mrocklin:
cc @adamklein

PR #2256, opened 2018-09-18
Using Cluster._worker_key function both in scale or adaptive

guillaumeeb:
Solution for https://github.com/dask/dask-jobqueue/issues/152 in order to have grouped worker taken into account when using scale method.

If using grouped workers submission as in https://github.com/dask/dask-jobqueue, for example launching workers processes 4 by 4, this avoids yo-yoing between 4 and 0 workers when calling cluster.scale(1) several times.

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -31,6 +31,14 @@ class Cluster(object):
     This will provide a general ``scale`` method as well as an IPython widget
     for display.
 
+    Clusters might also provide the following attributes:
+
+    -   _adaptive_options: dict
+        default options to use when using adaptive mode
+    -   _worker_key: Callable(WorkerState)
+        Callable mapping a WorkerState object to a group, see
+        Scheduler.workers_to_close
+
     Examples
     --------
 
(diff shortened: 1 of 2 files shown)


distributed/deploy/cluster.py
```diff
@@ -31,6 +31,14 @@ def scale_down(self, workers: List[str]):
     This will provide a general ``scale`` method as well as an IPython widget
     for display.
 
+    Clusters might also provide the following attributes:
+
+    -   _adaptive_options: dict
+        default options to use when using adaptive mode
+    -   _worker_key: Callable(WorkerState)
+        Callable mapping a WorkerState object to a group, see
+        Scheduler.workers_to_close
```
mrocklin: So, this seems like we're adding a new convention to the Cluster class to be used by downstream projects.  Is this something that we would want to adhere to long-term?  This part of the project seems sufficiently volatile that guaranteeing this long terms seems difficult.

mrocklin: In general I feel like this is a patch over a larger problem.  I'm inclined to think more about how to redesign the Cluster object and how it interacts with the Scheduler, Adaptive, and Clients more holistically.

guillaumeeb: Sounds fair, I agree this is kind of patching things around.

I'm not particularly concerned about this particular issue, It could wait. I suspect redesigning cluster object and interfaces might take some time.

I'm more concerned about #2209, I would have liked to have a solution soon enough.

Issue #2235, opened 2018-09-05
Redesign Cluster Managers


mrocklin:
Over the past few months we've learned more about the requirements to deploy Dask on different cluster resource managers.  This has created several projects like `dask-yarn`, `dask-jobqueue`, and `dask-kubernetes`.   These projects share some code within this repository, but also add their own constraints. 
 Now that we have more experience, this might be a good time to redesign things both in the external projects and in the central API.

There are a few changes that have been proposed that likely affect everyone:

1.  We would like to separate the `Cluster` manager object from the Scheduler in some cases
2.  We would like to optionally change how the user specifies cluster size to incorporate things like number of cores or amount of memory.  In the future this might also expand to other resources like GPUs.
3.  We would like to build more UI elements around the cluster object, like JupyterLab server extensions

The first two will likely require synchronized changes to all of the downstream projects.  

### Separate ClusterManager, Scheduler, and Client

I'm going to start calling the object that starts and stops workers, like `KubeCluster`, `ClusterManager` from here on.

Previously the `ClusterManager`, `Scheduler`, and `Client` were started from within a notebook and so all lived in the same process, and could act on the same event loop thread.  It didn't matter so much which part did which action.  We're now considering separating all of these.  This forces us to think about the kinds of communication they'll have to do, and where certain decisions and actions take place.  At first glance we might assign actions like the following:

1.  ClusterManager
    -  Start new workers
    -  Force-kill old workers
    -  Have UI that allows users to specify a desired makeup of workers, either a specific number or an Adaptive range.  I'm going to call this the Worker Spec
2.  Scheduler
    - Maintain `Adaptive` logic, and determine how many workers should exist given current load
    - Choose workes to remove and remove those workers politely
3.  Client: normal operation, I don't think that anything needs to change here.  

Some of these actions depend on information from the others.  In particular:

-  The `ClusterManager` needs to send the following to the `Scheduler`
    - The Worker Spec (like amount of desired memory or adaptive range), whenever it changes
-  The `Scheduler` needs to send the following to the `ClusterManager`
    - The desired target number of workers given current load (if in adaptive mode)
    - An announcement every time a new worker arrives (to manage the difference between launched/pending and actively running workers)
    - Requests to force-kill and remove workers after it politely kills them

This is all just a guess.  I suspect that other things might arise when actually trying to build everything here.

### How to specify cluster size 

This came up in https://github.com/dask/distributed/issues/2208 by @guillaumeeb .  Today we ask users for a number of desired workers

[...]

PR #2258, opened 2018-09-19
Add Scheduler method to compute target cores and memory

mrocklin:
This moves logic from Adaptive onto the main Scheduler and is a small
tentative step towards restructuring deployment.

cc @jcrist

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6,6 +6,7 @@ from functools import partial
 import itertools
 import json
 import logging
+import math
 from numbers import Number
 import operator
 import os
@@ -957,7 +958,8 @@ class Scheduler(ServerNode):
             'heartbeat_worker': self.heartbeat_worker,
             'get_task_status': self.get_task_status,
             'get_task_stream': self.get_task_stream,
-            'register_worker_callbacks': self.register_worker_callbacks
+            'register_worker_callbacks': self.register_worker_callbacks,
+            'target_workers': self.target_workers,
         }
 
         self._transitions = {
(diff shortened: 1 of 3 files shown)


distributed/scheduler.py
```diff
@@ -957,7 +958,8 @@ def __init__(
             'heartbeat_worker': self.heartbeat_worker,
             'get_task_status': self.get_task_status,
             'get_task_stream': self.get_task_stream,
-            'register_worker_callbacks': self.register_worker_callbacks
+            'register_worker_callbacks': self.register_worker_callbacks,
+            'target_workers': self.target_workers,
```
guillaumeeb: Why making this a handler method that can be called remotely?

mrocklin:
> Not sure I understand why you would want to make target_workers an handler. In your proposed design, adaptive intelligence is still running alongside the scheduler, isn't it?

The decision of "how many workers should the cluster have" is best made by the scheduler, which has all of the information necessary to make this decision.  

However the actual creation and destruction of the workers may happen on a separate process (ClusterManager) and so it may want to ask the scheduler for the target number of workers.

> You are planning things still further than this

There is still plenty of work to do, but I don't anticipate working on this topic personally in the next month.

> with non python scheduler and so one?

I'm not sure I understand what you mean here by non-python scheduler

guillaumeeb:
> However the actual creation and destruction of the workers may happen on a separate process (ClusterManager) and so it may want to ask the scheduler for the target number of workers.

From previous exchanges, I had the impression that it was the Scheduler that should ask the ClusterManager to scale to a given number of cores, not the ClusterManager asking how many workers to launch, but it may be the correct solution, I'm not sure. It depends where Adaptive logic must be run.

> I'm not sure I understand what you mean here by non-python scheduler

In another issue, IIRC, you speak about the possibility of Scheduler non being a python piece of software in the future.

mrocklin:
> From previous exchanges, I had the impression that it was the Scheduler that should ask the ClusterManager to scale to a given number of cores, not the ClusterManager asking how many workers to launch, but it may be the correct solution, I'm not sure. It depends where Adaptive logic must be run.

Yeah, I don't have a strong opinion here.  My guess though is that it will be easier for a ClusterManager to contact the Scheduler rather than the other way.  The scheduler is already running a server and we're accustomed to contacting it.  If we can avoid setting up another server for the ClusterManager that sounds ideal.  The cluster manager will also start the scheduler, and so will probably know the address to contact it.

guillaumeeb:
Yep you are right, that's one important design décision for #2235:
- ClusterManager will also run adaptive Logic
- Scheduler must provide remote method to informé resources It needs.

jrbourbeau:
Closing as I think was implemented elsewhere

PR #2263, opened 2018-09-19
Closes #2025, do not pull data twice

### LINKED ISSUE #2025: Gathers & memory issues

adamklein:
Prevents pulling data twice when already there, per comment in code review.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1461,7 +1461,7 @@ class Client(Node):
                     else:
                         raise ValueError("Bad value, `errors=%s`" % errors)
 
-            keys = [k for k in keys if k not in bad_keys]
+            keys = [k for k in keys if k not in bad_keys and k not in data]
 
             if local_worker:  # look inside local worker
                 data.update({k: local_worker.data[k]


mrocklin:
Thanks @adamklein !  It's good to add you to the authors list.  Welcome!

Issue #2184, opened 2018-08-14
Feature w/prototype implementation: Per-task memory profiling


itamarst:
In our Dask Distributed setup, which is currently single machine, I've needed to:

1. Figure out per-task peak memory usage, so we can figure out how many cores to use.
2. Reduce memory usage in ways that improves parallelism, which also means reducing peak memory for tasks that run in parallel.

To do so I have implemented a scheduler plugin that dumps a CSV of per-task memory usage (minimum and peak). Currently it assumes one thread per process, and all-local processes.

In general this has been extremely useful in helping achieve the two goals above, and I'm wondering if this is something worth integrating (in better written & more general form) into Dask Distributed:

```python
class MemoryUsagePlugin(SchedulerPlugin):
    """Record max and min memory usage for a task.

    Assumptions:

    * One task per process: each process has a single thread running a single
      task at a time.

    Limitations:

    * Statistical profiling at 10ms resolution.
    """
    def __init__(self, log_directory: str, cluster: LocalCluster):
        super().__init__()
        f = open(os.path.join(log_directory, "memory_usage.csv"), "w",
                 buffering=1)
        self._csv = csv.writer(f)
        self._csv.writerow(["task_key", "min_memory_mb", "max_memory_mb"])
        self._cluster = cluster
        # Map between worker address ("tcp://127.0.0.1:344") and list of memory
        # usage in MBs. cluster.workers is actually a list of Nanny objects,
        # which are wrappers around the underlying worker process.
        self._memory_per_worker = {worker.process.worker_address: []
                                   for worker in cluster.workers}
        # Schedule memory update every 10ms:
        def update_memory_usage():
            pc = PeriodicCallback(self._io_update_worker_memory_use, 10)
            pc.start()
        cluster.loop.add_callback(update_memory_usage)

    def _io_update_worker_memory_use(self):
        """Record current memory usage for worker processes.

        Runs in Tornado event loop thread.
        """
        for worker in self._cluster.workers:
            if worker.process is None:
                continue
            process = worker.process.process
            if process is None:
                continue
            try:
                proc = psutil.Process(process.pid)
                memory_mb = sum([
                    p.memory_info().rss / (1024 * 1024)
                    for p in [proc] + list(proc.children(recursive=True))
                ])
            except psutil.NoSuchProcess:
                continue
            try:
                self._memory_per_worker[
                    worker.process.worker_address].append(memory_mb)
            except KeyError:
                # Probably has been removed.
                pass

[...]

PR #2265, opened 2018-09-20
Design sketch: per-task memory profiling

itamarst:
This is sketch of better implementation of #2184. Would love feedback on this design.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1121,6 +1121,8 @@ class Scheduler(ServerNode):
 
         setproctitle("dask-scheduler [%s]" % (self.address,))
 
+        self.restart_plugins()
+
         return self.finished()
 
     @gen.coroutine
@@ -2167,6 +2169,12 @@ class Scheduler(ServerNode):
 
         self.plugins.append(plugin)
 
+        if self.status == "running":
+            try:
+                plugin.restart(self)
+            except Exception as e:
+                logger.exception(e)
+
     def remove_plugin(self, plugin):
         """ Remove external plugin from scheduler """
         self.plugins.remove(plugin)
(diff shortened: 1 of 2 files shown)


mrocklin:
I think it's a fine design and I'm glad that you've found it helpful.

If your objective is to merge this into the mainline codebase though then I think I'm mostly against, mostly because this seems specific (single threaded only, writes to csv file) rather than being universally useful.  I'd like to avoid maintaining plugins in the core codebase if at all possible to keep maintenance costs down.  However, I definitely encourage you to publish it in a separate package.  The plugin mechanism was designed so that tooling like this could live separated from the mainline codebase but still have full access.

itamarst:
1. Single-threaded is unavoidable, AFAICT.
2. I am happy to fix CSV thing.

But, if you still think it's not general enough, that's fine, can be separate package. Do you think it's useful enough to be a dask-* project? If not, I'll publish it on my own.

I should point out though that the need for this plugin to get access to scheduler object remains; if I do this separately, will a issue+PR for that be accepted?

itamarst:
That is, the changes I had to make to `scheduler.py`, or some other variation thereof.

distributed/scheduler.py
```diff
@@ -2167,6 +2169,12 @@ def add_plugin(self, plugin=None, idempotent=False, **kwargs):
 
         self.plugins.append(plugin)
 
+        if self.status == "running":
+            try:
+                plugin.restart(self)
+            except Exception as e:
+                logger.exception(e)
```
mrocklin: I'm not sure we want to trigger the `restart` event for plugins on startup.  I'm not sure how expected this would be by everyone.  Instead I recommend providing the scheduler as an argument as you create the plugin.

mrocklin: I may not be fully understanding the intent here though.

itamarst: The plugin (as currently designed) wants to make a Client. The Client requires a running Scheduler. Typically you add a plugin before the scheduler is running... so you need some way to know when the scheduler started.

There might be some other approach.

itamarst:
I will turn this into an independent project, I guess. Any problem with calling it `dask-memusage`?

itamarst:
OK, it's now an independent project, just needs a release: https://github.com/itamarst/dask-memusage

mrocklin:
Cool.  I look forward to find an opportunity to try this out.

Issue #2274, opened 2018-09-28
question about client.upload_file


GreatYYX:
Hi,

Is it possible to upload all files (include txt and other non-py or non-egg files) in my current directory to worker and use them?

I tried to compress them to zip file, but zip is not a module (to import). Then, I manually unzipped it, dask showed file not found (for those txt files). Is it because these files are not importable?

Thanks

GreatYYX:
I think I found a solution: `upload_file` can still be used to upload files (order does matter, upload non-py files first). Then in worker, change it's cwd to it's temp dir. I'm wondering if there's an official way to switch working directory for each process.

GreatYYX:
closing, a possible solution is here https://github.com/dask/distributed/pull/2275#issuecomment-426801816

PR #2275, opened 2018-09-28
support directory change after starting new worker processes

GreatYYX:
When a new worker process starts, current working directory (cwd) will change to it.

#2274


mrocklin:
Sorry for the long delay in reviewing this.  I hesitate to add this as a command line option.  It seems a bit too special case to me so far.  However, if more people were to show up who ran into the same problem that you did then that would be a good argument to adding this.  

In the meantime you might consider registering a startup function with https://distributed.readthedocs.io/en/latest/api.html#distributed.Client.register_worker_callbacks

Thoughts?

GreatYYX:
Thanks for letting me know, startup function is a good alternative if it's not a general case I think.

PR #2284, opened 2018-10-03
remove blosc threading from protocol.numpy

mgeplf:
* removed test_compress_numpy, as it was failing if blosc & python3
  was installed, and not run otherwise

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -4,12 +4,6 @@ import fractions
 
 import numpy as np
 
-try:
-    import blosc
-    n = blosc.set_nthreads(2)
-except ImportError:
-    blosc = False
-
 from .utils import frame_split_size, merge_frames
 from .serialize import dask_serialize, dask_deserialize
 from . import pickle
--- a/distributed/protocol/tests/test_numpy.py
+++ b/distributed/protocol/tests/test_numpy.py
@@ -7,7 +7,7 @@ import numpy as np
 import pytest
 
 from distributed.protocol import (serialize, deserialize, decompress, dumps,
-                                  loads, to_serialize, )
+                                  loads, to_serialize, msgpack)
 from distributed.protocol.utils import BIG_BYTES_SHARD_SIZE
 from distributed.utils import tmpfile, nbytes
 from distributed.utils_test import slow, gen_cluster
@@ -136,6 +136,24 @@ def test_dumps_serialize_numpy_large():
                                      ([('a', 'i4', 200), ('b', 'f8')], 8)])
 def test_itemsize(dt, size):
     assert itemsize(np.dtype(dt)) == size
+
+
+@pytest.mark.skipif(sys.version_info[0] < 3,
+                    reason='numpy doesnt use memoryviews')
+def test_compress_numpy():
+    pytest.importorskip('lz4')
+    x = np.ones(10000000, dtype='i4')
+    frames = dumps({'x': to_serialize(x)})
+    assert sum(map(nbytes, frames)) < x.nbytes
+
+    header = msgpack.loads(frames[2], encoding='utf8', use_list=False)
+    try:
+        import blosc  # noqa: F401
+    except ImportError:
+        pass
+    else:
+        assert all(c == 'blosc' for c in
+                   header['headers'][('x',)]['compression'])
 
 
 def test_compress_memoryview():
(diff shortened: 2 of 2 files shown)


distributed/protocol/tests/test_numpy.py
```diff
@@ -138,24 +138,6 @@ def test_itemsize(dt, size):
...
-    assert sum(map(nbytes, frames)) < x.nbytes
-
-    header = msgpack.loads(frames[2], encoding='utf8', use_list=False)
-    try:
-        import blosc  # noqa: F401
-    except ImportError:
-        pass
-    else:
-        assert all(c == 'blosc' for c in
-                   header['headers'][('x',)]['compression'])
```
mrocklin: This seems to be running for me

```
$ py.test distributed/protocol/tests/test_numpy.py::test_compress_numpy
============================= test session starts ==============================
platform linux -- Python 3.6.4, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
rootdir: /home/mrocklin/workspace/distributed, inifile:
plugins: xdist-1.23.0, timeout-1.2.1, forked-0.2, cov-2.5.1
collected 1 item                                                               

distributed/protocol/tests/test_numpy.py .                               [100%]

=========================== 1 passed in 0.11 seconds ===========================
```

distributed/protocol/numpy.py
```diff
@@ -4,12 +4,6 @@
 
 import numpy as np
 
-try:
-    import blosc
-    n = blosc.set_nthreads(2)
-except ImportError:
-    blosc = False
```
mrocklin: It is important to keep this code.  Blosc can improve performance when moving large memoryviews around, but we don't want it taking up all of our cores.

mrocklin: > compression.py handles the blosc side of things now.

Ah, you're right.  +1 on this change.  Thanks for pushing back.

distributed/protocol/tests/test_numpy.py
mrocklin: > do you have blosc installed?

Yes, otherwise the test would have skipped.  Regardless, I think that the intent of this test is correct.  If it is failing for some reason then it would be good to dive into what is causing it to fail and resolve that.  Do you have any interest in figuring out why your data isn't compressing here?

distributed/protocol/numpy.py
mgeplf: No problem, apologize if I seemed curt.

distributed/protocol/tests/test_numpy.py
mgeplf: Sigh, I'm looking at the wrong code, I didn't have lz4 installed, so `test_compress_numpy` worked, but `test_compression_takes_advantage_of_itemsize` failed.  I will add an importskip for that, and add the test backin.

Sorry about that.

distributed/protocol/tests/test_numpy.py
```diff
@@ -192,6 +174,7 @@ def test_dumps_large_blosc(c, s, a, b):
 @pytest.mark.skipif(sys.version_info[0] < 3,
                     reason='numpy doesnt use memoryviews')
 def test_compression_takes_advantage_of_itemsize():
+    pytest.importorskip('lz4')
```
mgeplf: w/o blosc or lz4 installed:
```
$ py.test distributed/protocol/tests/test_numpy.py
52 passed, 4 skipped, 15 warnings in 0.53 seconds
```
with blosc installed, but not lz4, before:
```
1 failed, 52 passed, 3 skipped, 15 warnings in 0.94 seconds
```
with this change:
```
52 passed, 4 skipped, 15 warnings in 0.60 seconds
```

distributed/protocol/numpy.py
mrocklin: Curtness is welcome.  The "Thanks" comment was genuine.

mrocklin: (also, you didn't seem curt)

mrocklin:
Other than the deleted test these changes look good to me.

mgeplf:
Fixed the deleted tests, correctly this time...
Thanks again.

mrocklin:
Thanks @mgeplf .  This is in.  

Also, I see that this is your first commit to this repository.  Welcome!

mgeplf:
I feel very welcomed, thank you.

Next time, I hope it's something more substantive.

PR #2292, opened 2018-10-05
Adapt to new Bokeh selection for 1.0

mrocklin:
cc @bryevdv

### DIFF
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -476,7 +476,10 @@ class ProfileTimePlot(DashboardComponent):
 
         def ts_change(attr, old, new):
             with log_errors():
-                selected = self.ts_source.selected['1d']['indices']
+                if bokeh.__version__ < '1.0':
+                    selected = self.ts_source.selected['1d']['indices']
+                else:
+                    selected = self.ts_source.selected.indices
                 if selected:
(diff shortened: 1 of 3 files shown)


bryevdv:
I would recommend a semantic test rather than an eplicit version test, e.g. if `self.ts_source.selected.indices` it not undefined, prefer that

mrocklin:
Sure, done.

Mostly I wanted to ping you so that you were aware of the backwards incompatible change.  Totally reasonable given the version number change, just an FYI as I run into things.

bryevdv:
the "dict" usage was discouraged several releases ago, when `Selection` objects were first added. We don't have a good way to warn about that programmatically, though. That said, for read-only use cases there should be no change, we aren't dropping the compat dicts altogether until 2.0, are you saying you experienced an issue?

mrocklin:
```
  File "/home/mrocklin/workspace/distributed/distributed/bokeh/components.py", line 639, in ts_change
    selected = self.ts_source.selected['1d']['indices']
TypeError: 'Selection' object is not subscriptable
```

bryevdv:
Oh, perhaps they were removed on the Python side. They still exist in BokehJS since there is much specifically JS example code in the wild that uses them.

PR #2293, opened 2018-10-07
Expand async docs

### LINKED ISSUE #2085: Clarify async documentation
I'm just testing out the distributed async support and seem to have hit a bug:
```python
from tornado.ioloop import IOLoop


client = Client()

async def f():
    print('A')
    future = client.submit(lambda x: x + 1, 10)
    print('B')
    result = await client.gather(future, asynchronous=True)
    print('C')
    return result

res = IOLoop().run_sync(f)
print(res)
client.close()
```
results in:
```
A
B
<HANG FOREVER>
```

I'm on win64 running:
```
python      3.6.5
distributed 1.22.0+9.gdb758d0f
dask        0.18.0
```

dhirschfeld:
Closes #2085 

I was a little confused how to run an async function using an existing `Client` object. This clarifies the documentation with the extra information provided by @mrocklin in #2085.


dhirschfeld:
If you don't think this is an improvement or necessary, feel free to close, otherwise I'm happy to make any requested changes.

dhirschfeld:
Errors seem unrelated:
```
E       TypeError: 'DaskDistributedBackend' object is not iterable
```

mrocklin:
See https://github.com/dask/distributed/pull/2298

dhirschfeld:
Looks like unrelated TLS errors this time.

mrocklin:
Yes, see https://github.com/dask/distributed/pull/2331

docs/source/asynchronous.rst
```diff
@@ -107,6 +107,23 @@ Python 3 with Tornado
    from tornado.ioloop import IOLoop
    IOLoop().run_sync(f)
 
+If you want to reuse an externally created client you need to run the async
+function using the clients loop which can be done with the ``Client.sync``
+function - e.g.
```
mrocklin: Statements like reuse "externally created client" and "using the client's loop" seem odd to me.  For the first I'm not sure why the client is external or why we're reusing it.  I think that this is just a normal client.  In the second case I don't think that we can expect all users to understand what an event loop is. 

Instead, I recommend something like ...

> If you want to use an asyncrhonous function with a synchronous client (one made without the `asynchronous=True` keyword) then you should use the `Client.sync` function. 

I've also been warned off of using `e.g.` in docs.  Apparently it's not universally understood by many people for whom English is not their first language. 

I think that the example is great.

dhirschfeld:
@mrocklin - if I make the suggested change above would you be happy merging this? If not I guess we can close it...

mrocklin:
Sure.  Sorry for the long delay in response.  I'm just going through my e-mail backlog now.

dhirschfeld:
np - I'll try and find the time to updating this PR shortly...

PR #2298, opened 2018-10-07
Remove Joblib Dask Backend from codebase

mrocklin:
This code has moved to joblib itself

This commits removes all relevant code and tests, and raises an
informative error message instead.

Recent changes in the joblib version of this code was causing our tests to fail.  I figure we should probably stop including a copy here.  I'm fine with requiring users who want to use this feature to depend on the latest release.  cc @TomAugspurger and @ogrisel


TomAugspurger:
Did you consider just printing out a warning instead of raising an error? I was going to ask for that, but upon further consideration the only adjustment people need to make is to delete an import.

dhirschfeld:
As a user of this feature I say rip it out!

The longer a bad/incorrect/broken feature remains in the harder it is to extricate IMHE.

The error message is clear and explicit.

TomAugspurger:
Thanks!

jakirkham:
For context/posterity PR ( https://github.com/joblib/joblib/pull/722 ) added the Distributed backend to Joblib and shows up in 0.12.2. This version of Joblib was also vendored in scikit-learn with PR ( https://github.com/scikit-learn/scikit-learn/pull/11741 ), which is included in scikit-learn 0.20.0.

jakirkham:
Probably safe to remove the last vestiges now. ( https://github.com/dask/distributed/pull/3040 )

PR #2299, opened 2018-10-08
Add explicit retry method to Future and Client

### LINKED ISSUE #2272: Explicitly retry failed tasks
Sometimes a task fails due to something external, then we change that external thing and want to try the task again.  

```python
>>> future = client.submit(load_data, filename)
>>> future.result()
FileNotFoundError

# oops, move the file to the correct location

>>> client.retry(future) 
```

If I were to implement this I would probably do the following:

1.  Make a client side method that used `futures_of` to find all futures in an object (including things like dask collections), then collect the keys of those that have failed and send them up to the scheduler with `self._send_to_scheduler`
2.  Make a scheduler-side route from the client that corresponds to the message sent above.  Find all of the root causes of the failed futures (using `TaskState.exception_blame`) and transition all of those to the released or waiting state (not sure at the moment makes sense)
3.  Make a `transition_erred_released` (or waiting) method, and in that method carefully walk back all of the state necessary to put the task into a good runnable condition again
4.  Write several tests

This would be a good task for someone looking to get into the distributed scheduler.

mrocklin:
This allows for the explicit retrying of failed tasks

Fixes #2272

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -370,7 +370,7 @@ class LoopRunner(object):
             # Loop already running in other thread (user-launched)
             done_evt.wait(5)
             if not isinstance(start_exc[0], RuntimeError):
-                if not isinstance(start_exc[0]):  # track down infrequent error
+                if not isinstance(start_exc[0], Exception):  # track down infrequent error
                     raise TypeError("not an exception", start_exc[0])
                 raise start_exc[0]
             self._all_loops[self._loop] = count + 1, None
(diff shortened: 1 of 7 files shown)


distributed/utils.py
```diff
@@ -370,7 +370,7 @@ def run_loop(loop=self._loop):
             # Loop already running in other thread (user-launched)
             done_evt.wait(5)
             if not isinstance(start_exc[0], RuntimeError):
-                if not isinstance(start_exc[0]):  # track down infrequent error
+                if not isinstance(start_exc[0], Exception):  # track down infrequent error
```
ogrisel: @mrocklin  Maybe launching pytest on travis with the `--showlocals` flag would help debug this kind of failures:

https://travis-ci.org/dask/distributed/jobs/438918060#L2577

PR #2301, opened 2018-10-08
use worker SSL context when getting client from worker.

### LINKED ISSUE #2288: get_client() not working from TLS-enabled worker
Hi,

it seems that ``worker.get_client()`` does not work from a worker that uses SSL.

to reproduce use environment from #2282 and run the following small example:
```python
# use venv/bin/python so you have the same version as worker/scheduler
from distributed.security import Security

s = Security(tls_ca_file="dask.pem", tls_client_cert="dask.pem", tls_client_key="dask.pem")
c = Client("tls://localhost:8786", security=s)

# To validate that this is only a TLS issue use the following client (against a non-ssl dask environment) instead:
# c=Client("localhost:8786")

def schedule_from_client():
    # this is run on the worker
    from dask.distributed import get_client
    def schedule_from_worker():
        return "worker scheduled me"

    client = get_client()  # get a client from the worker and schedule the function above
    # schedule the function  using the client
    result = client.get({"schedule_from_worker": (schedule_from_worker,)}, ["schedule_from_worker"], sync=True, pure=False) 
    return result


result = c.get({"schedule_from_client": (schedule_from_client,)}, ["schedule_from_client"], sync=True, pure=False)
print(result)

```
**Versions**: ``distributed==1.23.2, dask==0.19.2, tornado==5.1.1``
**Expected Result:** (this is what you get running without TLS): ``[['worker scheduled me']]``
**Actual Result**: [full stacktrace](https://github.com/dask/distributed/files/2449869/output.txt)
```
[...] venv/local/lib/python2.7/site-packages/distributed/comm/tcp.py", line 303, in _expect_tls_context
    raise TypeError("TLS expects a `ssl_context` argument of type "
```
I'm going to try and build an even more minimal reproducer as @mrocklin did in #2282. If I succeed I'll adapt the issue.

**edit:** slight modification of the reproducer for readability.

gluap:
This fixes #2288 for me. I have no easy test for this yet. -- I wasn't able to reproduce #2288 in the reduced TLS setup used for the TLS unit tests so far. #2288 describes in more detail how to test the fix. Locally unit tests ran through successfully with python.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -543,7 +543,12 @@ class Client(Node):
         self.security = security or Security()
         self.scheduler_comm = None
         assert isinstance(self.security, Security)
-        self.connection_args = self.security.get_connection_args('client')
+
+        if name == 'worker':
+            self.connection_args = self.security.get_connection_args('worker')
+        else:
+            self.connection_args = self.security.get_connection_args('client')
+
         self._connecting_to_scheduler = False
         self._asynchronous = asynchronous
         self._should_close_loop = not loop


mrocklin:
Thanks for this change @gluap .  In general I'm quite happy to see that you've found a way to resolve this.

I share some of your concerns about the logic here, mostly because we're relying on an informal convention that the client's name is worker if it belongs to a worker.  I would not be surprised if some well-meaning developer changed this in the future and this functionality stopped working.  To defend against this I think that we should find a way to test this properly.  

You mentioned that you tried to set something up before but it didn't work, possibly because of some shared state.  I just tried modifying some of our testing infrastructure to include TLS in https://github.com/dask/distributed/pull/2302 but as you experienced I wasn't able to get a simple test to fail.  

```python
from distributed.utils_test import tls_client, tls_cluster, security

def test_get_client(tls_client):
    def f():
        c = get_client()
        print(c.security)
        print(c.scheduler.address)

        assert c.submit(lambda x: x + 1, 10).result() == 11

    tls_client.submit(f).result()
```

Perhaps these simple testing credentials aren't complex enough in some way?  

Do you have any suggestions on how we might improve our testing here so that we can make this fail before this PR goes in?

mrocklin:
I plan to merge this PR sometime today, test or not (there is a new release coming out and I'd like to make sure that this fix is included).  It would be useful to have it tested though.

@jcrist , given your newfound expertise in TLS would you be interested in taking a quick look at our TLS testing solution in `distributed/utils_test.py` ?  I wouldn't dive too deeply here, but perhaps you can identify something that we're doing incorrectly.

jcrist:
This has nothing to do with the certificates, and more to do with the worker, client, and scheduler all sharing the same Security context (which has roles for all 3 defined). If you heavily patch all the test functions so that the client gets a security context with only a client role, worker with only a worker role, etc... *and* remove all the global clients in the test session for this test (I don't think `set_as_default=False` works anymore btw) then things pass.

This patch kind of works, but you could probably do better:

```diff
diff --git a/distributed/client.py b/distributed/client.py
index ac425d7..6bca3bc 100644
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -560,7 +560,10 @@ class Client(Node):
         self.security = security or Security()
         self.scheduler_comm = None
         assert isinstance(self.security, Security)
-        self.connection_args = self.security.get_connection_args('client')
+        if name == 'worker':
+            self.connection_args = self.security.get_connection_args('worker')
+        else:
+            self.connection_args = self.security.get_connection_args('client')
         self._connecting_to_scheduler = False
         self._asynchronous = asynchronous
         self._should_close_loop = not loop
diff --git a/distributed/tests/test_tls_functional.py b/distributed/tests/test_tls_functional.py
index 14f3495..6b86a4b 100644
--- a/distributed/tests/test_tls_functional.py
+++ b/distributed/tests/test_tls_functional.py
@@ -8,12 +8,31 @@ from __future__ import print_function, division, absolute_import
 
 from tornado import gen
 
-from distributed import Nanny, worker_client, Queue
+from distributed import Nanny, worker_client, Queue, get_client
 from distributed.client import wait
 from distributed.utils_test import (gen_tls_cluster, inc, double, slowinc,
         slowadd)
 
 
+@gen_tls_cluster(client=True, client_kwargs={'set_as_default': False})
+def test_tls_get_client(c, s, a, b):
+    import distributed.client
+    old_global = distributed.client._global_clients
+    try:
+        # Clear the global clients
+        distributed.client._global_clients = {}
+
+        def f():
+            c = get_client()
+            return c.submit(lambda x: x + 1, 10).result()
+
+        x = c.submit(f)
+        result = yield x
+        assert result == 11
+    finally:
+        distributed.client._global_clients = old_global
+
+
 @gen_tls_cluster(client=True)
 def test_Queue(c, s, a, b):
     assert s.address.startswith('tls://')
diff --git a/distributed/utils_test.py b/distributed/utils_test.py
index c586e14..a96e096 100644
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -780,10 +780,13 @@ from .worker import Worker
 @gen.coroutine
 def start_cluster(ncores, scheduler_addr, loop, security=None,
                   Worker=Worker, scheduler_kwargs={}, worker_kwargs={}):
-    s = Scheduler(loop=loop, validate=True, security=security,
-                  **scheduler_kwargs)
+    scheduler_kwargs = dict(scheduler_kwargs)
+    worker_kwargs = dict(worker_kwargs)
+    scheduler_kwargs.setdefault('security', security)
+    worker_kwargs.setdefault('security', security)
+    s = Scheduler(loop=loop, validate=True, **scheduler_kwargs)
     done = s.start(scheduler_addr)
-    workers = [Worker(s.address, ncores=ncore[1], name=i, security=security,
+    workers = [Worker(s.address, ncores=ncore[1], name=i,
                       loop=loop, validate=True,
                       **(merge(worker_kwargs, ncore[2])
                          if len(ncore) > 2
@@ -879,8 +882,8 @@ def gen_cluster(ncores=[('127.0.0.1', 1), ('127.0.0.1', 2)],
                             if s is False:
                                 raise Exception("Could not start cluster")
                             if client:
-                                c = yield Client(s.address, loop=loop, security=security,
-                                                 asynchronous=True, **client_kwargs)
+                                c = yield Client(s.address, loop=loop, asynchronous=True,
+                                                 **merge({'security': security}, client_kwargs))
                                 args = [c] + args
                             try:
                                 future = func(*args)
@@ -1358,6 +1361,27 @@ def tls_only_security():
     return sec
 
 
+def tls_client_worker_scheduler():
+    ca_file = get_cert('tls-ca-cert.pem')
+    keycert = get_cert('tls-key-cert.pem')
+
+    def tls_config(role):
+        return new_config({'tls': {'ca-file': ca_file,
+                            role: {'cert': keycert}},
+                           'require-encryption': True})
+
+    with tls_config('client'):
+        client = Security()
+
+    with tls_config('worker'):
+        worker = Security()
+
+    with tls_config('scheduler'):
+        scheduler = Security()
+
+    return client, worker, scheduler
+
+
 def get_server_ssl_context(certfile='tls-cert.pem', keyfile='tls-key.pem',
                            ca_file='tls-ca-cert.pem'):
     ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH,
@@ -1392,5 +1416,11 @@ def bump_rlimit(limit, desired):
 
 def gen_tls_cluster(**kwargs):
     kwargs.setdefault('ncores', [('tls://127.0.0.1', 1), ('tls://127.0.0.1', 2)])
-    return gen_cluster(scheduler='tls://127.0.0.1',
-                       security=tls_only_security(), **kwargs)
+    worker_kwargs = kwargs.setdefault('worker_kwargs', {})
+    scheduler_kwargs = kwargs.setdefault('scheduler_kwargs', {})
+    client_kwargs = kwargs.setdefault('client_kwargs', {})
+    client_security, worker_security, scheduler_security = tls_client_worker_scheduler()
+    worker_kwargs['security'] = worker_security
+    scheduler_kwargs['security'] = scheduler_security
+    client_kwargs['security'] = client_security
+    return gen_cluster(scheduler='tls://127.0.0.1', **kwargs)
```

jcrist:
I think a better fix might be to always use the client role for the clients, but have the `Security` object fill in the client role with the worker role's info if no specific certs are specified for the client.

mrocklin:
@jcrist I don't have much experience here.  I'm happy to defer to your
judgement on this.

jcrist:
Then yes, that's what we should do. The patch above is mostly fixing the test suite, but it was a hack to get the tests to fail (and then pass). The global client thing in particular is a huge hack, I don't know how to fix that, you'd probably be a better fit then.

mrocklin:
Merging this for now.  It's an incremental improvement.  It would be good to fix tests around this at some point.

Issue #1072, opened 2017-05-10
[RFC] Resilience improvements


kszucs:
Are there any plans for a highly available distributed cluster?

How much effort would it take (or is it possible at all) to:
- periodically persist the state of the scheduler for failover?
- or use a persistent backend instead of memory?
- have standby schedulers (via leader election etc.)?

mrocklin:
I don't have any such plans.  Of course if you feel up to this you're welcome to do so.

I suspect that all of these would take a non-trivial amount of effort.

mrocklin:
To be more clear (I have slightly more time now)

> periodically persist the state of the scheduler for failover?

Your guess is as good as mine.  Don't forget that you'd also have to ensure that workers respected this state at least somewhat on recovery.

> or use a persistent backend instead of memory?

I don't know of any persistent system that operates fast enough for this

> have standby schedulers (via leader election etc.)?

Having a standby scheduler isn't hard.  Maintaining updates might be hard.  


My previous thought on how to solve this problem was to have everyone re-submit what they know about the state of things if the scheduler goes down and comes up again.  Clients resubmit computations, workers check in with what they have.  I don't know if this is best though.

Would love to see work here.  I'm personally unlikely to work on this in the near future without serious enticement.

kszucs:
This is a good read: http://mesos.apache.org/documentation/latest/high-availability-framework-guide/

If I understand correctly the mesos agents (workers in our case) recovers their state to the newly elected master. 

Do workers maintain such state?

mrocklin:
See
http://distributed.readthedocs.io/en/latest/worker.html#api-documentation
for worker state

pitrou:
A possible way to approach this would be to have the scheduler dump its state to workers periodically (perhaps choosing a small number of different workers each time, to reduce the overall load). Then when a new scheduler instance goes up, reconnecting workers can announce they have some saved state of an old scheduler (e.g. tagged with a timestamp and a unique id). If the scheduler recognizes itself, it asks the worker with the most recent timestamp and asks to receive the saved state.

pitrou:
@mrocklin do you know how many tasks can the scheduler of a large cluster typically keep track of at any given time?

mrocklin:
[...]

PR #2311, opened 2018-10-15
Resilient Scheduler state 

Dracio:
This is a first rev at address #1072 . The scheduler state is saved to disk every 1 second, if a file is provided. The logic in add_worker returns if a task is still "executing". The logic for picking up when the result is in memory breaks test "test_worker_breaks_and_returns". 

I have created 1 test for the scheduler task state file.

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -15,6 +15,7 @@ distributed:
     transition-log-length: 100000
     work-stealing: True     # workers should steal tasks from each other
     worker-ttl: null        # like '60s'. Time to live for workers.  They must heartbeat faster than this
+    persist-interval: 1s    # time to sync TaskState to disc
 
   worker:
     multiprocessing-method: forkserver
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -144,7 +144,7 @@ def serialize(x, serializers=None, on_error='message', context=None):
             tb = traceback.format_exc()
             break
 
-    msg = "Could not serialize object of type %s" % type(x).__name__
+    msg = "Could not serialize object of type %s." % type(x).__name__
     if on_error == 'message':
         frames = [msg]
         if tb:
@@ -154,7 +154,7 @@ def serialize(x, serializers=None, on_error='message', context=None):
 
         return {'serializer': 'error'}, frames
     elif on_error == 'raise':
-        raise TypeError(msg)
+        raise TypeError(msg, str(x)[:10000])
 
 
 def deserialize(header, frames, deserializers=None):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -16,6 +16,8 @@ import warnings
 
 import psutil
 import sortedcontainers
+import shelve
+from pathlib import Path
 try:
     from cytoolz import frequencies, merge, pluck, merge_sorted, first
 except ImportError:
@@ -770,6 +772,7 @@ class Scheduler(ServerNode):
             scheduler_file=None,
             security=None,
             worker_ttl=None,
+            persist_file=None,
             **kwargs):
 
         self._setup_logging()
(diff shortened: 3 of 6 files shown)


mrocklin:
Thanks for giving this a shot @Dracio .  I'm curious, have you used it in practice?  If so, how does it perform?

Dracio:
> 1. What happens when the events that occur in the 1s lag between saving state are important?  For example a worker may have finished or deleted work, but restarted scheduler doesn't know about it.  Does it wait forever or is there some mechanism for the worker to update the scheduler with this new information?

This occurs in the add_worker function. I added logic to have the worker send the current state of each task it has a record of. This logic still needs to be fleshed out to make sure all the attributes from taskstate and workerstate are correlated correctly. 

> 2. How slow is this when we have millions of tasks?

I will try and setup a good stress test. It will depend on the size of self.tasks.

inc0:
Quickly looking at this change, did you consider using external service (optionally) like Redis? Problem with using file as mechanism for state sync will be shared filesystem requirement for clustered schedulers. In cloud world it's not very trivial thing to setup shared fs that's reliable and safe. Having support for simple but fast protocol that Redis offers we can depend on it's replication model (which isn't ideal, but it's not horrible as well).

Dracio:
> Quickly looking at this change, did you consider using external service (optionally) like Redis? 

> More resilient option would be ETCD, but number of active tasks would be something to consider (it should scale to millions of keys).

These maybe opinions. This was only my first pass at addressing this problem.

> Problem with using file as mechanism for state sync will be shared filesystem requirement for clustered schedulers. In cloud world it's not very trivial thing to setup shared fs that's reliable and safe. Having support for simple but fast protocol that Redis offers we can depend on it's replication model (which isn't ideal, but it's not horrible as well). 

I don't understand this comment. I don't see any documentation on more than one scheduler for a cluster. The task state on the scheduler is the only one that is saved to disk.

inc0:
> I don't understand this comment. I don't see any documentation on more than one scheduler for a cluster. The task state on the scheduler is the only one that is saved to disk.

I'm looking at it from cloud perspective. To allow autohealing of scheduler in cloud, you won't have guarantee that it will restart on same node (especially if said node blows up). If we want to address autohealing of even single scheduler on different node with file-based dumps we'd need to either reconnect network attached storage if it's single writer of have it connected via multi-write. There are huge problems with ensuring that single writer is properly unattached from original node, so only real viable option is distributed, multi-write fs.

On the other hand, if we use k-v store like etcd or redis, this problem is solved (very well) on k-v storage layer. It is also quite a bit more performant than writing to disk (especially if we'd want to dump whole thing always) so we might not need to have intervals like that, or have them very short and add logic to only update statuses that changed.

mrocklin:
I like seeing the conversation on this topic, but generally speaking I don't think that persisting the scheduler state is the way to go.  It changes too rapidly for anything but in-memory on-node storage.  Instead I think that the following are the only pragmatic ways to achieve high availability with a Dask deployment:

1.  Don't, and ask people to resubmit computations should the single point of failure go down (spark style)
2.  Have workers and clients reconnect to the scheduler and share their view of the world.  Have the scheduler re-assemble what it can from these views.

Currently we do option 1.  Option 2 is feasible with effort.

I'm pessimistic on persisting state because it changes too rapidly, and because I think it will be difficult to handle small mismatches in understanding between schedulers and workes.

inc0:
> Have workers and clients reconnect to the scheduler and share their view of the world. Have the scheduler re-assemble what it can from these views.

Will workers have context of what remains to be scheduled?

Dracio:
> Combined with the clients, yes, I think that that's feasible. #594

When the client is disconnected, then you lose that information. If a machine reboots in the middle of the night, then it has no clue what tasks need to be scheduled.

mrocklin:
Correct.  However if a client disconnects then that user probably won't be expecting the result back.  This isn't always correct (for example with `fire_and_forget` computations) but is usully the case.

inc0:
> Correct. However if a client disconnects then that user probably won't be expecting the result back. 

Unless last scheduled task would be to dump data to .csv, then it still should work correct?

mrocklin:
I'm going to close this.  Work here has stalled, and I'm not sure that this direction is likely to be the long term solution.

PR #2315, opened 2018-10-18
Fix DeprecationWarning

gsheni:
- There is multiple `DepricationWarnings` with python 3.7. 
> 
> DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
- This pull request fixes these import warnings but maintain backwards compatibility.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1,10 +1,11 @@
 from __future__ import print_function, division, absolute_import
+from collections import defaultdict, deque, OrderedDict
 try:
     # Python 3
-    from collections.abc import defaultdict, deque, OrderedDict, Mapping, Set
+    from collections.abc import Mapping, Set
 except ImportError:
     # Python 2.7
-    from collections import defaultdict, deque, OrderedDict, Mapping, Set
+    from collections import Mapping, Set
 from datetime import timedelta
 from functools import partial
 import itertools
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1,11 +1,5 @@
 from __future__ import print_function, division, absolute_import
 from collections import defaultdict, deque, OrderedDict
-try:
-    # Python 3
-    from collections.abc import Mapping, Set
-except ImportError:
-    # Python 2.7
-    from collections import Mapping, Set
 from datetime import timedelta
 from functools import partial
 import itertools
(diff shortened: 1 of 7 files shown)


mrocklin:
Thank you for the contribution @gsheni .  I appreciate this.  I'm happy to merge this, although it might also be useful to put this in the `compatibility.py` module, where we tend to keep version-dependent imports like these.  Would you be interested in moving these imports to compatibility and then importing from there within the client.py and scheduler.py files?

(my apologies for the delay in response)

gsheni:
Moved the imports to `compatibility.py` and imported from there. 
I also noticed there was a double import of `from toolz import first` which I removed.

mrocklin:
This looks great.  Thank you @gsheni 

I searched through the codebase and found a few other occurrences of Mapping and Iterator.

```
mrocklin@carbon:~/workspace/distributed$ git grep import.*Mapping
distributed/publish.py:from collections import MutableMapping
distributed/scheduler.py:from collections import defaultdict, deque, OrderedDict, Mapping, Set
mrocklin@carbon:~/workspace/distributed$ git grep import.*Iterator
distributed/client.py:from collections import Iterator, defaultdict
distributed/tests/test_as_completed.py:from collections import Iterator
distributed/tests/test_client.py:from collections import Iterator, deque
distributed/tests/test_utils.py:from collections import Iterator
```

Is `MutableMapping` also deprecated?  If so we should probably give that the same treatment.

It also looks like there are occurrences of `Iterator` in the testing files.

mrocklin:
Ah, I should clarify that I ran grep while on master, not this PR so many of those are already taken care of here.  A few are not though, notably the `MutableMapping` case and the cases in the tests.

gsheni:
@mrocklin no problem.
I fixed the `MutableMapping` and `Mapping` instances as well. I also fix the additional instances of `Iterator`.

mrocklin:
Thanks!  Merging.

Also, I see that this is your first code contribution to this project.  Welcome and thank you for the contribution!

distributed/scheduler.py
```diff
@@ -20,7 +20,7 @@
     from cytoolz import frequencies, merge, pluck, merge_sorted, first
 except ImportError:
     from toolz import frequencies, merge, pluck, merge_sorted, first
-from toolz import valmap, first, second, compose, groupby
+from toolz import valmap, second, compose, groupby
```
jakirkham: Is there a reason to prefer these come from `toolz` only instead of optionally coming from `cytoolz` as well?

gsheni: There is not reason I can think of. I wasn't aware of cytoolz until now, but it seems that from the ReadME it would be worth implementing, and see if it breaks anything:
https://github.com/pytoolz/cytoolz#cytoolz

jakirkham: No worries. Went ahead and switched to optionally using `cytoolz` where available. ( https://github.com/dask/distributed/pull/3579 )

We've done that for similar functions in other places. So the risk of breaking is low.

PR #2325, opened 2018-10-29
Add preload script to conf

### LINKED ISSUE #2121: Add preload scripts to configuration
Users can provide a `--preload` script to their dask-worker or dask-scheduler commands to launch custom code during execution.  This is pragmatic when you want to extend Dask with plugins, but somewhat annoying to do every time.

We might add two entries to the configuration `distribued.scheduler.preloads` and `distributed.worker.preloads` that are lists of scripts or imports to pass to the Worker and Scheduler on startup.  This should be the default within the `worker.py::WorkerBase` and `scheduler.py::Scheduler` constructors.

This was first suggested, I think, by @lesteve

guillaumeeb:
Fixes #2121.

Still trying to dive inside codebase, so I pick up some simple issues. Two questions for this one:
- Did not find any test about preload scripts. Don't know how to build one... Should we?
- For Scheduler, preload script are only loaded in cli.dask_scheduler, so I made the modification here, is this OK?

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -127,9 +127,9 @@ def main(host, port, bokeh_port, show, _bokeh, bokeh_whitelist, bokeh_prefix,
                           security=sec)
     scheduler.start(addr)
     if not preload:
-        preload = dask.config.get('distributed.scheduler.preloads')
+        preload = dask.config.get('distributed.scheduler.preload')
     if not preload_argv:
-        preload_argv = dask.config.get('distributed.scheduler.preloads-argv')
+        preload_argv = dask.config.get('distributed.scheduler.preload-argv')
     preload_modules(preload, parameter=scheduler, file_dir=local_directory, argv=preload_argv)
 
     logger.info('Local Directory: %26s', local_directory)
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -1,6 +1,7 @@
 from __future__ import print_function, division, absolute_import
 
 import atexit
+import dask
 import logging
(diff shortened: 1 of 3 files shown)


distributed/cli/dask_scheduler.py
```diff
@@ -125,6 +126,10 @@ def del_pid_file():
                           scheduler_file=scheduler_file,
                           security=sec)
     scheduler.start(addr)
+    if not preload:
+        preload = dask.config.get('distributed.scheduler.preloads')
+    if not preload_argv:
+        preload_argv = dask.config.get('distributed.scheduler.preloads-argv')
```
mrocklin: Thoughts on preload vs preloads?  I'm not sure if we should pluralize this word, given the single-ness of the `--preload` CLI keyword.

guillaumeeb: Fine by me to go with preload, I'll update the PR accordingly.

guillaumeeb:
>  I think that it would be good to move this within the Scheduler and Worker classes

For Worker this is already the case. For Scheduler, there seems to be some reason about it, thus I'm not sure this is a trivial change. I did not find a simple way to do it when I looked at it.

mrocklin:
Test failures here seem unrelated.  See https://github.com/dask/distributed/pull/2331

mrocklin:
Sorry for the delay on this.  TLS issues were fixed in other PRs.  Things are green now.  Merging.

guillaumeeb:
No problem, I was planning to look at the tests you mentioned to see if I can add something. But this seems simple enough to merge it as it is, hopefully I did not messed up with default kwargs :slightly_smiling_face:

PR #2335, opened 2018-11-05
Add serializer for masked arrays

dkillick:
Currently masked array objects are serialized and deserialized by the NumPy serializer/deserializer, which can fall afoul of this peculiarity of the NumPy API:

```python
import numpy as np
x = np.arange(8)
mx = np.ma.masked_less(x, 5)

x.data
<memory at 0x7f1d3c72d048>

mx.data
array([0, 1, 2, 3, 4, 5, 6, 7])
```

This means that if a masked array gets passed to the NumPy serializer, not all of the frames are serialized properly to bytes, causing the error reported in https://github.com/dask/dask/issues/4163.

This adds a serializer and deserializer specifically for masked arrays, which splits masked arrays up to separately handle `data`, `mask` and `fill_value`.

Fixes https://github.com/dask/dask/issues/4163.

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -102,3 +102,37 @@ def deserialize_numpy_ndarray(header, frames):
                        strides=header['strides'])
 
         return x
+
+
+@dask_serialize.register(np.ma.core.MaskedArray)
+def serialize_numpy_maskedarray(x):
+    # Separate elements of the masked array that we need to deal with discretely.
+    data = x.data
+    mask = x.mask
+    fill_value = x.fill_value
+
+    # Make use of existing numpy serialization for the two ndarray elements of
+    # the masked array.
+    data_header, data_frames = serialize_numpy_ndarray(data)
+    mask_header, mask_frames = serialize_numpy_ndarray(mask)
+
+    header = {"data-header": data_header,
+              "mask-header": mask_header,
+              "fill_value": fill_value,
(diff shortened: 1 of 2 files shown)


dkillick:
@mrocklin I haven't added the code from the original error report as an extra test to this yet, because I wasn't quite sure where to put it. My thought was to add it to `distributed/tests/test_client.py`, but my concern was that NumPy isn't currently imported in that test module, which made me wonder if a NumPy-importing test wasn't quite the right fit for that test module?

distributed/protocol/numpy.py
```diff
@@ -102,3 +102,37 @@ def deserialize_numpy_ndarray(header, frames):
...
+              "mask-header": mask_header,
+              "fill_value": fill_value,
+              "nframes": [len(data_frames), len(mask_frames)]}
+    return header, data_frames + mask_frames
+
+
+@dask_deserialize.register(np.ma.core.MaskedArray)
+def deserialize_numpy_maskedarray(header, frames):
+    data_frames = frames[:header["nframes"][0]]
+    mask_frames = frames[header["nframes"][1]:]
```
mrocklin: I think you actually only need `header['nframes'][0]` here in both cases.  I suspect that you only need this split point between the two lists of frames.  I don't think that you need both lengths.

dkillick:
Super! Thanks for your help with this @mrocklin

Issue #2193, opened 2018-08-19
Traceback in Distributed logs from Graph Dashboard plugin


jakirkham:
Saw this traceback popup in the Distributed logs while looking at the Graph Dashboard. This was early in a very busy computation. So not much had completed yet.

```python
distributed.utils - ERROR - '<' not supported between instances of 'NoneType' and 'tuple'
Traceback (most recent call last):
  File "/opt/conda3/lib/python3.6/site-packages/distributed/utils.py", line 644, in log_errors
    yield
  File "/opt/conda3/lib/python3.6/site-packages/distributed/bokeh/scheduler.py", line 1107, in graph_doc
    graph = GraphPlot(scheduler, sizing_mode='stretch_both')
  File "/opt/conda3/lib/python3.6/site-packages/distributed/bokeh/scheduler.py", line 612, in __init__
    self.layout = GraphLayout(scheduler)
  File "/opt/conda3/lib/python3.6/site-packages/distributed/diagnostics/graph_layout.py", line 38, in __init__
    priority=priority)
  File "/opt/conda3/lib/python3.6/site-packages/distributed/diagnostics/graph_layout.py", line 42, in update_graph
    stack = sorted(dependencies, key=priority.get, reverse=True)
TypeError: '<' not supported between instances of 'NoneType' and 'tuple'
```

jrbourbeau:
Reopening as @ncclementi reported seeing this with the latest `distributed` release (2021.09.1). There's also this PR https://github.com/dask/distributed/pull/2337 which was an attempt to fix this issue

ncclementi:
I'll leave here a minimal reproducible example. The code comes form the dask-tutorial for reference. I run into this when using joblib with dask as a backend. 

```python
import dask.distributed
client = dask.distributed.Client(n_workers=4)

from sklearn.datasets import make_classification
from sklearn.svm import SVC
from sklearn.model_selection import GridSearchCV
import joblib

X, y = make_classification(n_samples=10000, n_features=4, random_state=0)

estimator = SVC(gamma='auto', random_state=0, probability=True)
param_grid = {
    'C': [0.001, 10.0],
    'kernel': ['rbf', 'poly'],
}

grid_search = GridSearchCV(estimator, param_grid, verbose=2, cv=2, n_jobs=-1)

with joblib.parallel_backend("dask", scatter=[X, y]):
    grid_search.fit(X, y)
```

When clicking on the Graph tab on the dashboard, I get the following traceback
```
[...]

PR #2337, opened 2018-11-06
Fix '<' not supported between instances of 'NoneType' and 'tuple'

NotSqrt:
See #2193 

The fix in 9b8e092 was not enough.

Thanks !

### DIFF
--- a/distributed/diagnostics/graph_layout.py
+++ b/distributed/diagnostics/graph_layout.py
@@ -39,7 +39,7 @@ class GraphLayout(SchedulerPlugin):
 
     def update_graph(self, scheduler, dependencies=None, priority=None,
                      **kwargs):
-        stack = sorted(dependencies, key=lambda k: priority.get(k, 0), reverse=True)
+        stack = sorted(dependencies, key=lambda k: priority.get(k) or tuple(), reverse=True)
         while stack:
             key = stack.pop()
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks for the proposed fix @NotSqrt 

In the Appveyor tests we run across this error

```
distributed.scheduler - ERROR - unorderable types: tuple() < int()
3317Traceback (most recent call last):
3318  File "C:\projects\distributed\distributed\scheduler.py", line 1561, in update_graph
3319    loose_restrictions=loose_restrictions)
3320  File "C:\projects\distributed\distributed\diagnostics\graph_layout.py", line 42, in update_graph
3321    stack = sorted(dependencies, key=lambda k: priority.get(k) or tuple(), reverse=True)
3322TypeError: unorderable types: tuple() < int()
3323
```

So unfortunately it looks like neither solution works.  We can have priorities that are both integers or tuples.  This might require more cleverness.

ghltshubh:
Any solution to this problem?

jrbourbeau:
As you can see, things have stalled here. Looking at the current `master` branch the relevant lines of code are still the same. If you're running into this `TypeError`, I recommend opening a new issue

dkoes:
For what it's worth, I had this same problem (dask 2021.02.0) and manually applying this patch restored the Task graph.

PR #2345, opened 2018-11-08
[WIP] Investigate worker-scheduler latency

mrocklin:
This branch does a few things to help investigate latency between the
scheduler and workers:

1.  Includes a benchmark script that uses the inproc:// protocol and
    computes a sequential computation
2.  Turns off all BatchedSend throttles, allowing communications to be
    sent rapidly one after the other
3.  Adds print statements throughout key sections of the code to
    highlight what parts of code take what durations

We find that fully sequential computations have milliseconds of latency,
and then when we look at the /profile-server dashboard we don't see much
activity.  It's not clear what's taking up all of our time.

Here is some sample output.  Each duration value corresponds to the amount of time spent since the last reported step (see the `tictoc` function defined in this PR).

```
scheduler send task to worker 123.50 us
worker accepts task 276.80 us
worker transitions key 61.27 us
worker submits task to threadpool 89.88 us
worker starts computing 134.47 us
worker reports done scheduler 200.27 us
scheduler task finished 214.58 us

scheduler send task to worker 123.26 us
worker accepts task 265.60 us
worker transitions key 32.66 us
worker submits task to threadpool 109.67 us
worker starts computing 162.60 us
worker reports done scheduler 213.38 us
scheduler task finished 323.06 us
```

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -40,7 +40,8 @@ from .proctitle import setproctitle
 from .security import Security
 from .utils import (All, ignoring, get_ip, get_fileno_limit, log_errors,
                     key_split, validate_key, no_default, DequeHandler,
-                    parse_timedelta, PeriodicCallback, shutting_down)
+                    parse_timedelta, PeriodicCallback, shutting_down, tictoc,
+                    format_time)
 from .utils_comm import (scatter_to_workers, gather_from_workers)
 from .utils_perf import enable_gc_diagnosis, disable_gc_diagnosis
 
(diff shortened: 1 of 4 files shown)


mrocklin:
The timing mechanism seems to add around 10us of cost

mrocklin:
> when we look at the /profile-server dashboard we don't see much activity. It's not clear what's taking up all of our time.

Ah, when looking at the system resources though we do find that Python is taking up 100% of the CPU though, so presumably our statistical profiler is missing things.

Issue #2351, opened 2018-11-12
Unable to import distributed 1.24.1 with tornado < 5 on python 3.6


rwedge:
It looks like some code depends on tornado 5.0

Steps to recreate:
1. Install packages
```bash
pip install distributed==1.24.1 tornado==4.5.3
```
2. Import distributed
```python
import distributed
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/__init__.py", line 5, in <module>
    from .actor import Actor, ActorFuture
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/actor.py", line 4, in <module>
    from .client import Future, default_client
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/client.py", line 44, in <module>
    from .batched import BatchedSend
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/batched.py", line 10, in <module>
    from .core import CommClosedError
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/core.py", line 20, in <module>
    from .comm import (connect, listen, CommClosedError,
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/comm/__init__.py", line 9, in <module>
    from .core import connect, listen, Comm, CommClosedError
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/comm/core.py", line 12, in <module>
    from ..utils import parse_timedelta
  File "/home/roy/venv/lib/python3.5/site-packages/distributed/utils.py", line 1390, in <module>
    asyncio.set_event_loop_policy(tornado.platform.asyncio.AnyThreadEventLoopPolicy())
AttributeError: module 'tornado.platform.asyncio' has no attribute 'AnyThreadEventLoopPolicy'
```

[Looks like](https://www.tornadoweb.org/en/stable/asyncio.html#tornado.platform.asyncio.AnyThreadEventLoopPolicy) 'AnyThreadEventLoopPolicy' was added in tornado 5.0.

rwedge:
@mrocklin Sure

PR #2353, opened 2018-11-12
set loop policy if tornado version 5 or above

rwedge:
As discussed in #2351, ignore a section of code if the version of Tornado is too old.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1384,7 +1384,7 @@ if 'asyncio' in sys.modules:
             isinstance(traitlets.config.Application.instance(), NotebookApp)
         )
 
-    if not jupyter_event_loop_initialized:
+    if not jupyter_event_loop_initialized and tornado.version_info[0] >= 5:
         import asyncio
         import tornado.platform.asyncio
         asyncio.set_event_loop_policy(tornado.platform.asyncio.AnyThreadEventLoopPolicy())
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1372,7 +1372,7 @@ def reset_logger_locks():
 
 
 # Only bother if asyncio has been loaded by Tornado
-if 'asyncio' in sys.modules:
+if 'asyncio' in sys.modules and tornado.version_info[0] >= 5:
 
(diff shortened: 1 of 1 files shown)


distributed/utils.py
```diff
@@ -1384,7 +1384,7 @@ def reset_logger_locks():
             isinstance(traitlets.config.Application.instance(), NotebookApp)
         )
 
-    if not jupyter_event_loop_initialized:
+    if not jupyter_event_loop_initialized and tornado.version_info[0] >= 5:
```
rwedge: If the previous code block is only necessary for this if statement it might make more sense to the tornado version check on 1375 with the asyncio check .

mrocklin: Yes, that's the case.  You're right that it's probably better to include this check above.  It's fine either way though.

.travis.yml
```diff
@@ -6,7 +6,7 @@ dist: trusty
 
 env:
   matrix:
-    - PYTHON=2.7 TESTS=true PACKAGES="python-blosc futures faulthandler lz4"
+    - PYTHON=2.7 TESTS=true PACKAGES="python-blosc futures faulthandler lz4 tornado=4.5"
```
rwedge: I did not encounter this specific bug with python 2.7 and tornado 4.5, probably because of the `if 'asyncio' in sys.modules` check.

mrocklin:
OK, so it looks like there are other issues with Tornado 4.5 due to TLS issues.  I've been going through history for a bit and running tests locally to see when these arrived.  I think that there are two easy solutions to this:

1.  Depend on Tornado >= 5
2.  Depend on Tornado >= 5 for TLS only

I'm inclined to just bump the minimum version requirement to Tornado 5.  I'm curious if this would affect people.  I'll post an issue and publicize it on Twitter to gain community feedback.

mrocklin:
That can happen separately from this though.  I'll remove the tornado 4.5 testing commit from this branch and merge it in.

mrocklin:
Thanks for flagging and resolving this @rwedge !

Also, I notice that this is your first contribution to this repository.  Welcome!

Issue #2359, opened 2018-11-15
Redirect / in dashboard to /status


mrocklin:
We've removed that bokeh index page (it was often incorrect).  We should either build our own index page or else redirect to /status

mikewheel:
I think I have this fixed to redirect to /status, I just don't know how to build distributed from source to test it out. How would I do that?

mrocklin:
Dask is pure Python, so doesn't need to be built.  You can run your code
within the root directory of the project, or pip install it from source.
http://distributed.dask.org/en/latest/develop.html

mikewheel:
Thanks for the tip @lesteve. Should we also close #2338 ?

mrocklin:
Thanks.  Done

PR #2361, opened 2018-11-15
Fixed the 404 error on the Scheduler Dashboard homepage

mikewheel:
See issue #2359

### DIFF
--- a/distributed/bokeh/core.py
+++ b/distributed/bokeh/core.py
@@ -33,6 +33,7 @@ class BokehServer(object):
                                      check_unused_sessions_milliseconds=500,
                                      allow_websocket_origin=["*"],
                                      use_index=False,
+                                     extra_patterns=[(r'/', web.RedirectHandler, {'url': '/status'})],
                                      **self.server_kwargs)
                 self.server.start()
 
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for doing this @mikewheel .  I have a naive question, if we're hosting this web server at some other endpoint, will this redirect work properly?  

To make the question concrete, people often host the dashboard behind a proxy using nbserverproxy, so they navigate to something like the following:

    https://my-webserver/proxy/8787/status

In this situation would this redirect still work?   Would `https://my-webserver/proxy/8787` redirect to `https://my-webserver/proxy/8787/status` ?

mikewheel:
You know to be honest I'm not sure. I would assume so, only because the app locations are all specified the same way in the existing code (that is, relative to '/' like '/status' or '/profile'). But I don't know and can only guess.

mrocklin:
OK, well then lets just merge this and see if people continue to have this problem in those situations then. 

Thanks again @mikewheel !

mrocklin:
Also, I notice that this is your first contribution to this repository.  Welcome!

PR #2364, opened 2018-11-18
Avoid warnings in pyarrow and msgpack

mrocklin:


distributed/protocol/arrow.py
```diff
@@ -11,7 +11,7 @@ def serialize_batch(batch):
     writer = pyarrow.RecordBatchStreamWriter(sink, batch.schema)
     writer.write_batch(batch)
     writer.close()
-    buf = sink.get_result()
+    buf = sink.getvalue()
```
xhochy: You might need to increase the minimal Arrow version for this.

mrocklin: Which version would you recommend?  Also, is that version commonly deployed?  My guess is that current Arrow users are early adopters and are ok with updating frequently.

mrocklin: It looks like this works with pyarrow 0.9.0, which seems fairly old to me today

xhochy: I would suggest to require 0.10 as minimal version which is in `pyarrow` terms already old.

mrocklin:
TLS CI failures are unrelated.  Raised at #2369 

Merging

PR #2373, opened 2018-11-26
Fixup serializing masked arrays

jcrist:
A few fixups for serializing masked arrays:

- Properly serialize `fill_value` across all dtypes
- Be more efficient when serializing masked arrays with nomask
- Improve test coverage across dtypes

Fixes https://github.com/dask/dask/issues/4225

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -105,33 +105,41 @@ def deserialize_numpy_ndarray(header, frames):
 
 @dask_serialize.register(np.ma.core.MaskedArray)
 def serialize_numpy_maskedarray(x):
-    # Separate elements of the masked array that we need to deal with discretely.
-    data = x.data
-    mask = x.mask
-    fill_value = x.fill_value
+    data_header, frames = serialize_numpy_ndarray(x.data)
+    header = {'data-header': data_header,
+              'nframes': len(frames)}
 
-    # Make use of existing numpy serialization for the two ndarray elements of
-    # the masked array.
-    data_header, data_frames = serialize_numpy_ndarray(data)
-    mask_header, mask_frames = serialize_numpy_ndarray(mask)
+    # Serialize mask if present
+    if x.mask is not np.ma.nomask:
+        mask_header, mask_frames = serialize_numpy_ndarray(x.mask)
+        header['mask-header'] = mask_header
+        frames += mask_frames
 
-    header = {"data-header": data_header,
-              "mask-header": mask_header,
-              "fill_value": fill_value,
-              "nframes": len(data_frames)}
-    return header, data_frames + mask_frames
+    # Only a few dtypes have python equivalents msgpack can serialize
+    if isinstance(x.fill_value, (np.integer, np.floating, np.bool_)):
+        serialized_fill_value = (False, x.fill_value.item())
+    else:
+        serialized_fill_value = (True, pickle.dumps(x.fill_value))
+    header['fill-value'] = serialized_fill_value
+
+    return header, frames
 
 
 @dask_deserialize.register(np.ma.core.MaskedArray)
 def deserialize_numpy_maskedarray(header, frames):
-    data_frames = frames[:header["nframes"]]
-    mask_frames = frames[header["nframes"]:]
     data_header = header["data-header"]
-    mask_header = header["mask-header"]
-
-    # Get the individual elements of the masked array in order to reconstruct.
+    data_frames = frames[:header["nframes"]]
     data = deserialize_numpy_ndarray(data_header, data_frames)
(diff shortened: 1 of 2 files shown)


distributed/protocol/numpy.py
```diff
@@ -105,33 +105,41 @@ def deserialize_numpy_ndarray(header, frames):
...
+    if x.mask is not np.ma.nomask:
+        mask_header, mask_frames = serialize_numpy_ndarray(x.mask)
+        header['mask-header'] = mask_header
+        frames += mask_frames
+
+    # Only a few dtypes have python equivalents msgpack can serialize
+    if isinstance(x.fill_value, (np.integer, np.floating, np.bool_)):
+        serialized_fill_value = (False, x.fill_value.item())
+    else:
+        serialized_fill_value = (True, pickle.dumps(x.fill_value))
```
mrocklin: I suggest that we just always pickle for simplicity.  I don't think that this will affect performance much.

mrocklin:
cc @dkillick

distributed/protocol/numpy.py
jcrist: Perhaps, although pickling a simple numpy scalar results in an unfortunate number of bytes:

```
In [12]: len(msgpack.dumps(pickle.dumps(x)))
Out[12]: 120

In [13]: len(msgpack.dumps(x.item()))
Out[13]: 9

In [14]: %timeit msgpack.dumps(pickle.dumps(x))
10.3 µs ± 157 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [15]: %timeit msgpack.dumps(x.item())
1.72 µs ± 22.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

distributed/protocol/tests/test_numpy.py
```diff
@@ -82,18 +79,20 @@ def test_dumps_serialize_numpy(x):
...
-    np.testing.assert_equal(data, y.data)
-    np.testing.assert_equal(mask, y.mask)
-    assert fill_value == y.fill_value
+@pytest.mark.parametrize('x', [
+    np.ma.masked_array([5, 6], mask=[True, False], fill_value=10, dtype='i4'),
+    np.ma.masked_array([5., 6.], mask=[True, False], fill_value=10, dtype='f4'),
+    np.ma.masked_array([5., 6.], mask=[True, False], fill_value=np.nan, dtype='f8'),
+    np.ma.masked_array([5., 6.], mask=np.ma.nomask, fill_value=np.nan, dtype='f8'),
+    np.ma.masked_array([True, False], mask=np.ma.nomask, fill_value=True, dtype='bool'),
+    np.ma.masked_array(['a', 'b'], mask=[True, False], fill_value='c', dtype='O')
```
dkillick: One more parameter that would be interesting to see here is 
```python
np.ma.masked_array([5, 6], mask=True, fill_value=10)
```
We've occasionally seen some odd behaviour stemming from NumPy in this particular case, which is something to do with how NumPy interprets the mask. I've just tried creating one of these locally, however, and it doesn't seem to be that unusual, so perhaps this has been improved in more recent NumPy versions.

You could also consider checking the default fill-values for different dtypes, which was implicitly being done before by making use of the default fill-values in the first two masked array parameters. Given, however, that these checks are effectively being done by the first two parameters here, and I doubt there's much difference between serializing `10.0` and `1.0e+20`, this probably isn't too important.

distributed/protocol/numpy.py
mrocklin: Hrm, that is surprisingly expensive.  OK, seems reasonable to me.

distributed/protocol/tests/test_numpy.py
dkillick: Apologies, I didn't demonstrate the situation I was thinking of correctly. Here's a functioning example, produced by not specifying the mask at all when creating a masked array:

```python
>>> np.ma.arange(10)
masked_array(data = [0 1 2 3 4 5 6 7 8 9],
             mask = False,
       fill_value = 999999)
```

(For reference, the same behaviour is also present using `dask.array.ma`.)

mrocklin: @dkillick just to make sure I understand, is there a test that you think should be added here?

From what I read above it sounds like you didn't actually find a concrete issue.  I may be misreading though and want to check in before I hit merge here.

mrocklin:
Merging prior to a release.  Thanks @jcrist and @dkillick

distributed/protocol/tests/test_numpy.py
dkillick: @mrocklin sorry I missed your comment. There's no concrete issue that I know of here, just the chance of some unexpected NumPy behaviour that I thought it might be good to test.

PR #2384, opened 2018-11-30
Serialize numpy.ma.masked objects properly

jcrist:
`numpy.ma.masked` is a singleton of type `np.ma.core.MaskedConstant`. Since this singleton is a subclass of `np.ma.core.MaskedArray`, it was erroneously being picked up by the serializer for `MaskedArray`s, leading to errors. We now register a serializer/deserializer for `MaskedConstant` directly, fixing the issue.

Fixes https://github.com/dask/dask/issues/4225#issuecomment-443035031

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -103,6 +103,16 @@ def deserialize_numpy_ndarray(header, frames):
         return x
 
 
+@dask_serialize.register(np.ma.core.MaskedConstant)
+def serialize_numpy_ma_masked(x):
+    return {}, []
+
+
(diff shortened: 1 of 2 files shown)


jcrist:
Failure appears to be unrelated. cc @mrocklin.

mrocklin:
Yes, that particular issue comes up from time to time.  It's quite annoying and slows down development, but I haven't been able to isolate it.  Someone is somehow sending a totally unrelated message to that comm.  This is the case with any error on this line `raise ValueError("Unexpected response", response)`

I've restarted that test for now.

mrocklin:
Thanks for resolving this swiftly @jcrist

PR #2386, opened 2018-11-30
Fix race condition in normalize_collection

### LINKED ISSUE #2385: Race condition between Client.normalize_collection and Future.release
It seems there's a race condition between futures being cleaned up, and futures being reused by a collection:

```python
from __future__ import absolute_import, division

from dask import array as da


if __name__ == '__main__':
    with Client() as client:
        # Repeat until failure
        while True:
            x = da.arange(2**20, chunks=2**10)
            y = x.persist()
            wait(y)
            # Start cleaning up futures...
            del y
            # Find existing futures and reuse them, *race condition here*
            normalized = client.normalize_collection(x)
            normalized.compute()
```

Result:

```
$ python test.py
  File "test.py", line 18, in <module>
    normalized.compute()
  File "/Users/jcrist/Code/dask/dask/base.py", line 156, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File "/Users/jcrist/Code/dask/dask/base.py", line 398, in compute
    return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
  File "/Users/jcrist/Code/dask/dask/base.py", line 398, in <listcomp>
    return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
  File "/Users/jcrist/Code/dask/dask/array/core.py", line 904, in finalize
    return concatenate3(results)
  File "/Users/jcrist/Code/dask/dask/array/core.py", line 3637, in concatenate3
    result[idx] = arr
ValueError: invalid literal for int() with base 10: "('arange-96617f71e88551ee72155f5966b630aa', 45)"
```

It looks like futures are found to still exist in `normalize_collection`, but are fully released before they can be reused by the `.compute()` call (resulting in keys being treated as raw strings instead).

jcrist:
This cleans up the use of locks in `Client`, and fixes a race condition between `Client.normalize_collection` and `Future.release`. We now have single reentrant-lock on the client, that guards the refcounts on all futures managed by that client. Any operation that changes this state,
or requires consistency throughout its operation, should acquire this lock before proceeding. We then guard the critical section of `normalize_collection` preventing a race condition with `Future.release`, where previously futures would be decrefed and freed and then still used in the collection later. A test is added to confirm the race condition is fixed.

Fixes #2385.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -543,8 +543,11 @@ class Client(Node):
         self.cluster = None
         self.scheduler = None
         self._scheduler_identity = {}
-        self._lock = threading.Lock()
-        self._refcount_lock = threading.Lock()
+        # A reentrant-lock on the refcounts for futures associated with this
+        # client. Should be held by individual operations modifying refcounts,
+        # or any bulk operation that needs to ensure the set of futures doesn't
+        # change during operation.
+        self._refcount_lock = threading.RLock()
         self.datasets = Datasets(self)
         self._serializers = serializers
         if deserializers is None:
@@ -1249,7 +1252,7 @@ class Client(Node):
 
         skey = tokey(key)
 
-        with self._lock:
+        with self._refcount_lock:
             if skey in self.futures:
                 return Future(key, self, inform=False)
 
@@ -2184,7 +2187,7 @@ class Client(Node):
                           loose_restrictions=None, priority=None,
                           user_priority=0, resources=None, retries=None,
                           fifo_timeout=0, actors=None):
-        with self._lock:
+        with self._refcount_lock:
             if resources:
                 resources = self._expand_resources(resources,
                                                    all_keys=itertools.chain(dsk, keys))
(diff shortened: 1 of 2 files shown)


jcrist:
I don't know if these errors are periodic or new, they appear to be unrelated. cc @mrocklin.

jcrist:
FWIW I don't expect this to cause much of a slowdown. The only case that might be affected that I can think of is if someone is manually calling `normalize_collection` many times on large collections, all of which match existing futures, while also manipulating futures in other threads (submitting/releasing). But in that case I would expect a user to have run into this issue before. All other cases are strictly more correct with respect to race conditions.

mrocklin:
I'm also not concerned about performance here.  I've grown respectful of some of the more complex codebases out there and their ability to uncover unexpected issues.  

> All other cases are strictly more correct with respect to race conditions.

OK, that's reassuring.

mrocklin:
Sorry for the delay in review here.  I pushed up a small change to the test.  Otherwise this looks good to me.

Issue #1949, opened 2018-05-01
Not monitoring the process tree, only the worker proccess


barakor:
When using subprocess, the memory usage is counted by the local scheduling system but not by distributed, this leads to 2 problems:
A) we can't monitor the processes from the Dashboard properly
B) distributed's handling of memory usage isn't brought into effect, and so the worker is killed by the scheduling system in some cases. 

Is this intentional?

guillaumeeb:
Stack Overflow says we can watch all the children too, but this will probably leads to potential performance penalty.

This could prove useful in some use cases I have with dask-jobqueue though, even if I'm not sure if distributed's handling of memory usage would be efficient here.

PR #2390, opened 2018-12-01
Danpf/track subprocess resources

danpf:
I'm not sure of the performance implications of this, but I've been using this to track my resource usage.

fix for https://github.com/dask/distributed/issues/1949

description:
Currently, Dask can only track the worker's cpu usage and memory usage. If you make another process via `subprocess` that process's resources will not be tracked.   This makes us track the resource usage of all process children instead of just the python process.

I'm not sure if this is everything that needs to be changed, but it appears to work for me. I'm guessing this is slightly slower since tests are failing.

### DIFF
--- a/distributed/system_monitor.py
+++ b/distributed/system_monitor.py
@@ -10,7 +10,7 @@ from .metrics import time
 class SystemMonitor(object):
     def __init__(self, n=10000):
         self.proc = psutil.Process()
-        self.proc_children = psutil.children(recursive=True)
+        self.proc_children = self.proc.children(recursive=True)
 
         self.time = deque(maxlen=n)
         self.cpu = deque(maxlen=n)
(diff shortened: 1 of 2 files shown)


mrocklin:
Hi @danpf !  Sorry for the delay in review.  In principle this approach seems sensible to me.  I agree though that the larger question of "do we want to do this" is probably worth investigating though.  I've found that the system monitor can add a non-trivial baseline cost to the scheduler that would be good to avoid.

danpf:
it definitely adds overhead from what I have seen. however, since you have to run `cpu_percent` at least twice to get a readout. (ie keep a list of children and only update it if necessary), you have to be a little clever about keeping a list of child processes and how you update it.

This isn't a really big deal PR, it's mostly just here for if you're interested.

PR #2391, opened 2018-12-02
WIP Allowing named worker-setups

rainwoodman:
1. Convert worker-setups to a dictionary.

2. Coroutines in scheduler takes a callbacks dict as input,
   where functions are serialized as strings/bytes.

3. When the name already exists,
   the newly registered setup function will immediately be run
   on all existing clients, but new clients will only see
   the updated version of the setup function.


rainwoodman:
This is a WIP.

I am having difficulty running the tests. pytest gives me strange errors:
```
  $ pytest distributed

  .....
    action = super(_ArgumentGroup, self)._add_action(action)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/argparse.py", line 1366, in _add_action
    self._check_conflict(action)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/argparse.py", line 1505, in _check_conflict
    conflict_handler(action, confl_optionals)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/argparse.py", line 1514, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument -L/--leaks: conflicting option string: --leaks
```

mrocklin:
Unfortunately I'm not familiar with that error message.  I recommend trying from a clean environment or making sure that you don't have other configuration lying around.  

I know that this is a WIP, but one thought on quick review is that it would be good to support backwards compatibility for the Client API.  In particular we might place the `name` paramter after `setup` and have it default to `distributed.utils.funcname(setup)` or something similar.

rainwoodman:
-  you mean running pytest in the code directory is supposed to work.

- I was thinking of an an API breakage for two reasons:

If we use an optional name it shall probably be the fully qualified function name to avoid accidental conflicts. In many cases it is impossible to uniquely define a fully qualified function name of the callback function, and therefore callers are forced to provide a name function anyways.

Also problematic is the the planned inclusion of other callback functions; when that happens, which function name do we use?

guillaumeeb:
I'm also against API breakage. Accidental conflicts could only come from several users using the same cluster, in this case it shall be clearly documented to use a name or label for the callback.

> Also problematic is the the planned inclusion of other callback functions; when that happens, which function name do we use?

I'm not sure I get this, we should record one callback per function, so one function name per callback?

rainwoodman:
OK. Let me figure what I can do.

distributed/tests/test_worker.py
```diff
@@ -1251,21 +1264,33 @@ def test_startup2():
...
-    assert len(response) == 2
+    names2 = yield c.register_worker_callbacks(setup=mystartup2)
+    assert len(names2) == 1
     assert len(s.worker_setups) == 2
 
     # Check it has been run
     result = yield c.run(test_startup2)
     assert list(result.values()) == [True] * 2
 
+    # unregister a preload function
```
rainwoodman: I changed 'register_worker_callbacks' to return the registered names to allow unregisterring callback functions with inferred names.

distributed/worker.py
```diff
@@ -610,23 +612,26 @@ def _register_with_scheduler(self):
...
+        # Retrieve eventual init functions (only worker-setups for now)
+        for name, function_bytes in response['worker-setups'].items():
+            self.worker_setups[name] = pickle.loads(function_bytes)
+
         self.batched_stream = BatchedSend(interval='2ms', loop=self.loop)
         self.batched_stream.start(comm)
         self.periodic_callbacks['heartbeat'].start()
         self.loop.add_callback(self.handle_scheduler, comm)
 
+        # run eventual init functions (only worker-setups for now)
```
rainwoodman: I moved this to run the worker init functions after heartbeat; also I am using the locak worker_setups directionary rather than the unpickled function directly.

distributed/scheduler.py
```diff
@@ -3090,16 +3091,45 @@ def get_task_stream(self, comm=None, start=None, stop=None, count=None):
         return ts.collect(start=start, stop=stop, count=count)
 
     @gen.coroutine
-    def register_worker_callbacks(self, comm, setup=None):
-        """ Registers a setup function, and call it on every worker """
-        if setup is None:
-            raise gen.Return({})
+    def register_worker_callbacks(self, comm, callbacks, names):
```
rainwoodman: I preserved the Client API, but changed the scheduler API in a in compatible way:

This is to avoid
```
register_worker_callbacks(self, comm, setup, setup_name, teardown, teardown_name):
```

distributed/client.py
```diff
@@ -3546,25 +3547,39 @@ def _get_task_stream(self, start=None, stop=None, count=None, plot=False,
...
+    def _register_worker_callbacks(self, setup=None, name=None):
+        callbacks = {}
+        names = {}
+        # prepare the callbacks and their names
+        if setup is not None:
+            serialized = dumps(setup)
+            callbacks['setup'] = serialized
+            if name is None:
+                h = md5(serialized)
+                name = funcname(setup) + '-' + h.hexdigest()
```
rainwoodman: This achieves the hashing based on function "body" @mrocklin  mentioned.

distributed/scheduler.py
guillaumeeb: I understand the problem, but I'm not entirely convinced by the proposed solution. I've not much to propose though...

Are you expecting that we register more than a setup and teardown callbacks at once in the future?

Could we use tuples instead of dict containing several callbacks? e.g stick to
```python 
register_worker_callbacks(self, comm, setup=None, teardown=None):
```

But where we expect `setup` or `teardown` to be `(name, function)`?

distributed/client.py
rainwoodman: Speed is not a concern here since it is only intended to be used at initialization. 

The hash produces two names for lib1.setup and lib2.setup; even when a name is not explicitly given. I would preferred `__qualname__`, but it is only defined in Python 3.3+.

distributed/scheduler.py
rainwoodman: I will rework the PR along this line.

distributed/scheduler.py
```diff
@@ -3091,15 +3092,32 @@ def get_task_stream(self, comm=None, start=None, stop=None, count=None):
...
         raise gen.Return(responses)
 
+    @gen.coroutine
+    def unregister_worker_callbacks(self, comm, setup=None):
+        # add the setup function to the list to run them on new clients.
+        if setup is not None:
+            name, func = setup
+            self.worker_setups.pop(name)
+
+        raise gen.Return(None)
```
mrocklin: This doesn't need to be a coroutine.  I recommend removing the `@gen.coroutine` line and replacing `raise gen.Return(...)` with `return ...`

mrocklin: To be clear, the reason here is that this method never `yield`s, so it doesn't need to be a coroutine

distributed/client.py
```diff
@@ -3563,19 +3592,43 @@ def register_worker_callbacks(self, setup=None):
...
+    def unregister_worker_callbacks(self, setup=None):
+        """ Unregisters a worker callback.
+
+            Parameters
+            ----------
+            setup : callable(dask_worker: Worker) -> None, or tuple of (name, callable)
+                The setup callback to remove; it will be matched by the name.
+                If a name is not given, then it is generated from the callable.
+        """
+        return self.sync(self._unregister_worker_callbacks, setup=setup)
```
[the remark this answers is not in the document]
rainwoodman: `Client.cancel` still goes through `Client._cancel`.  I had the impression it was convention to always have two interfaces of the same functionality, the coroutine version  and the non-coroutine version.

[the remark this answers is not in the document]
mrocklin: Ah, you're right.  Perhaps `list_datasets` then.  

We need to create a coroutine if there is a `yield` somewhere.  If not, then we're happy to skip it for simplicity.

distributed/scheduler.py
mrocklin: Sorry for not responding to this earlier.  I personally prefer asking for the `name=` keyword explicitly rather than passing it through as a tuple.  

```python
client.register_worker_callbacks(setup=setup_database, teardown=teardown_database, name='database')
```

I personally find passing tuples somewhat awkward.

@guillaumeeb what was your concern with the `name=` keyword solution?  That is it hard to provide ordered parameter compatibility going forward?

distributed/client.py
mrocklin: I recommend using `dask.base.tokenize` generally for hashing objects.  We tend to centralize our logic there.

mrocklin:
I apologize for the delay in review here.  I've been travelling for work a bit more than is ideal.

distributed/scheduler.py
rainwoodman: I liked the tuple because:

- each callback needs a different name. 
- it was  decided `register_worker_callbacks` can register multiple callbacks in one call.
- there is no clean way to assign names to callbacks with a single name argument and multiple callbacks.

rainwoodman:
Is there a way to `sync` a coroutine in the tests? (like self.sync). I am talking about this

```
@gen_cluster(client=True)
def test_register_worker_callbacks(c, s, a, b):
    
    ....

    def assert_mystartup1_called():
          r = ???.sync(client.run, lambda worker: hasattr(worker, 'mystartup1_value')
          return r.values()

    ....
    if not all(assert_mystartup1_called()): raise AssertionError(xxxxxx)

```

The test case can be hugely simplified if I can do this.

mrocklin:
> Is there a way to sync a coroutine in the tests?

The client.run method magically works as a coroutine if in a coroutine.  So you should be able to do the following:

    r = yield client.run(...)

distributed/scheduler.py
guillaumeeb: My concern was not about the `name=` keyword solution, but about the `callbacks=dict` solution, which I find kind of obscure even if documented. The tuple may not be better though.

I think I prefer explicit `register_worker_callbacks(self, comm, setup, setup_name, teardown, teardown_name)` as the one with only a dict of callbacks.

rainwoodman: There wouldn't be a problem if the function was instead:

`register_worker_callback(self, comm, callback_type, function , name)`

I doubt there is need to make it atomic registering several callbacks.

rainwoodman: Here is another possibility. 

```
register_worker_callbacks(self, comm, setup, name=None)
```

If `setup` takes two arguments `setup(worker, event)`, then we will call it as 
`setup(worker, 'setup')`
`setup(worker, 'teardown')`.

If `setup` takes one or zero argument, we only call it during setup.

mrocklin: Maybe when we add teardowns we decide that we need classes instead?

So today

```python
def register_worker_callbacks(self, comm, setup, name=None):
```

Where setup is a callable.

In the future

```python
def register_worker_callbacks(self, comm, cb, name=None):
```

Where cb is either a callable, in which case it is interpretted as a setup function, or it is a class with the following structure:

```python
class MyWorkerCallback:
    def __init__(self, worker):
        ...
    def setup(self):
        ...
    def tearDown(self):
        ...
    ...
```

In this way we continue to maintain the same structure today, provide a path towards supporting teardowns in the future, but don't have to rely on implicit tuple semantics.

Issue #2371, opened 2018-11-23
Scheduler memory increasing over time due to ever growing event log


fjetter:
We have distributed clusters deployed with an adaptive/dynamic worker scaling using mesos/aurora. We frequently down-scale these clusters when they are idling and scale them up when there is work to do. The newly assigned workers usually will have different addresses as the workers we had running in a previous upscale cycle.
The issue now is that the current implementation of event logging seems to remember an event log per worker address (e.g. [here](https://github.com/dask/distributed/blob/8fd95ede2d3672c40e38f86503a2e38a46a50fb4/distributed/scheduler.py#L1259))
While the event log per worker is limited by a `deque(maxsize=10k)` the number of potential distinct event logs may grow without boundary leading to increased memory consumption of the scheduling process over time when new clients (new clients will, by default, generate a UUID which they use as a name/identifier) or workers connect.
As far as I am aware, there is currently no automatic cleanup for these logs and no other way to free this memory than restarting the scheduler.

Is there a mechanism I missed? If not, how should an implementation for this look like?

May be related to #1795

mrocklin:
I'm curious, have you measured in some way to tell if this is the cause of your memory leak?

You might do something like the following to see if it reduces the resident memory significantly.

```python
def f(dask_scheduler):
    dask_scheduler.events.clear()

client.run_on_scheduler(f)
```

If indeed this is the cause of your exceessive memory use then one way to resolve this would be to clear the worker's logs from the scheduler, perhaps after some suitable delay.

```python
def remove_worker(...):
    
    @gen.coroutine
    def remove_worker_from_events():
        yield gen.sleep(3600)  # wait an hour
        if address not in self.workers:
            del self.events[address]

    self.loop.add_callback(remove_worker_from_events)
```

fjetter:
Unfortunately, the scheduler where we observed this _leak_ ran out of memory and is dead by now :)
I started another one and let it run for a day. Since I didn't know what to look for I started to inspect `dask_scheduler.__dict__` with something like:
```python
def track_memory(dask_scheduler):
    import sys
    from numbers import Number
    from collections import Set, Mapping, deque

    try: # Python 2
        zero_depth_bases = (basestring, Number, xrange, bytearray)
        iteritems = 'iteritems'
    except NameError: # Python 3
        zero_depth_bases = (str, bytes, Number, range, bytearray)
        iteritems = 'items'

    def getsize(obj_0):
        _seen_ids = set()
        try:
            """
[...]

PR #2394, opened 2018-12-03
Remove clients and workers from event log after removal

fjetter:
This fixes the issue discussed in #2371 

Not sure what a reasonable default would be for the cleanup interval. I went with an hour, as suggested in the ticket but I believe a much higher value (a day?) would also be just fine.

I thought about implementing this as a plugin but the plugin API currently doesn't say anything about the add/remove client. If the plugin would be preferable, though, one could consider to not cleanup clients at all since they do not log that much data (e.g. no continuous heartbeat)


distributed/config.py
```diff
@@ -25,6 +25,7 @@
     'allowed-failures': 'distributed.scheduler.allowed-failures',
     'bandwidth': 'distributed.scheduler.bandwidth',
     'default-data-size': 'distributed.scheduler.default-data-size',
+    'events-cleanup-delay': 'distributed.scheduler.events-cleanup-delay',
```
mrocklin: This dictionary is intended for backwards compatibility.  I don't think that we need this alias.  I would prefer that users today use the fully qualified name.

distributed/distributed.yaml
```diff
@@ -12,6 +12,9 @@ distributed:
     allowed-failures: 3     # number of retries before a task is considered bad
     bandwidth: 100000000    # 100 MB/s estimated worker-worker bandwidth
     default-data-size: 1000
+    # Number of seconds to wait until workers or clients are removed from the events log
+    # after they have been removed from the scheduler
+    events-cleanup-delay: 3600
```
mrocklin: I recommend that we use `"1h"` or `"1 hour"` here and make sure that we use `distributed.utils.parse_timedelta` wherever it is read.

distributed/scheduler.py
```diff
@@ -2054,6 +2065,16 @@ def remove_client(self, client=None):
                                       client=cs.client_key)
             del self.clients[client]
 
+        @gen.coroutine
+        def remove_client_from_events():
+            # If the client isn't registered anymore after the delay, remove from events
+            if client not in self.clients and client in self.events:
+                del self.events[client]
```
mrocklin: Style request.  Can I ask for a blank line between the function definition and the following line?

distributed/tests/test_scheduler.py
```diff
@@ -194,6 +194,39 @@ def test_remove_worker_from_scheduler(s, a, b):
...
+    assert b.address in s.events
+    assert b.address in s.ncores
+
+    s.remove_worker(address=a.address)
+    # Shortly after removal, the events should still be there
+    assert a.address in s.events
+    assert a.address not in s.ncores
+    s.validate_state()
+
+    yield gen.sleep(0.02)
```
mrocklin: The time here is short enough that it might fail intermittently.  This is particularly a problem on Travis-CI, which can be *very* slow sometimes.  Instead, I recommend the following pattern:

```python
start = time()
while a.address in s.events:
    yield gen.sleep(0.01)
    assert time() < start + 2
assert b.address in s.events
```

This says "this condition should happen soonish, but we don't mind if it's a second or two late" (which could be something like a `gc.collect` in a bad case.

distributed/tests/test_scheduler.py
```diff
@@ -194,6 +194,39 @@ def test_remove_worker_from_scheduler(s, a, b):
     s.validate_state()
 
 
+@gen_cluster(config={"distributed.scheduler.events-cleanup-delay": 0.01})
```
mrocklin: Slight preference for `"10 ms"` here and below to be explicit about times and units.

fjetter:
@mrocklin I think I fixed all remarks you had. Two questions from my side: 

- Do you maintain the changelog yourself or should I add an entry somewhere?
- Is there any place where the config parameters are documented, other than the `distributed.yml` file? Is this `yml` file exhaustive?

mrocklin:
- Do you maintain the changelog yourself or should I add an entry somewhere?

They're written during release.  We take the title of the PR as the entry, changing it if necessary.  You don't need to do anything here.

-  Is there any place where the config parameters are documented, other than the distributed.yml file? Is this yml file exhaustive?

Currently the YAML file is the only documentation.  Yes, the YAML file is exhaustive.

Issue #2392, opened 2018-12-02
max_map_len problem in the latest `msgpack` library (version==0.6.0)


ZhuozhaoLi:
The default `max_map_len` in the latest `msgpack` (version 0.6.0) is 32768, which limits the length of input in `client.map()` to this value. I need to roll back to `msgpack==0.5.6` to solve this problem

`File "/lib/python3.6/site-packages/distributed/protocol/core.py", line 197, in loads_msgpack
    return msgpack.loads(payload, use_list=False, **msgpack_raw_false)
  File "msgpack/_unpacker.pyx", line 187, in msgpack._cmsgpack.unpackb
ValueError: 40000 exceeds max_map_len(32768)`

mrocklin:
Interesting.  I have not seen this before.  I'm curious, do you know what
caused this?  Do you have any recommendations on how to resolve it?

ZhuozhaoLi:
> Interesting. I have not seen this before. I'm curious, do you know what caused this? Do you have any recommendations on how to resolve it?

Like I said, it was caused by the default value change of `max_map_len` in the latest version (0.6.0) of `msgpack`. See here: https://github.com/msgpack/msgpack-python/blob/b1d658e7a0aed4e723f9b6b238cc9f2f876e54d7/msgpack/fallback.py#L256.
I think you can reproduce this error if you update `msgpack` to 0.6.0 and try a `client.map` with more than 32768 inputs.

However, in `msgpack=0.5.6`, the value was `2**31-1` -- it is sufficiently large. So I think a quick way is to put `msgpack==0.5.6` in your `requirement.txt`?

PR #2399, opened 2018-12-05
proposed msgpack 0.6.0 fix

tjb900:
Possible fix for #2392

### DIFF
--- a/distributed/protocol/tests/test_protocol.py
+++ b/distributed/protocol/tests/test_protocol.py
@@ -137,15 +137,10 @@ def test_large_messages():
     assert (msg['y']['a'].data == msg2['y']['a']).all()
 
 
-@slow
 def test_large_messages_map():
-    np = pytest.importorskip('numpy')
-    psutil = pytest.importorskip('psutil')
+    import psutil
     if psutil.virtual_memory().total < 8e9:
-        return
-
-    if sys.version_info.major == 2:
-        return 2
+        pytest.skip("insufficient memory")
 
     x = {i: 'mystring_%d' % i for i in range(100000)}
 
--- a/distributed/protocol/tests/test_protocol.py
+++ b/distributed/protocol/tests/test_protocol.py
@@ -137,6 +137,23 @@ def test_large_messages():
     assert (msg['y']['a'].data == msg2['y']['a']).all()
 
 
+@slow
+def test_large_messages_map():
+    np = pytest.importorskip('numpy')
+    psutil = pytest.importorskip('psutil')
+    if psutil.virtual_memory().total < 8e9:
+        return
+
+    if sys.version_info.major == 2:
+        return 2
+
+    x = {i: 'mystring_%d' % i for i in range(100000)}
+
+    b = dumps(x)
+    x2 = loads(b)
+    assert x == x2
+
+
 def test_loads_deserialize_False():
     frames = dumps({'data': Serialize(123), 'status': 'OK'})
     msg = loads(frames)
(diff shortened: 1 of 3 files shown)


mrocklin:
Thanks for the fix @tjb900 !

A couple small comments:

-  It looks like the flake8 linter didn't like one of the lines here:

        distributed/protocol/serialize.py:61:30: E226 missing whitespace around arithmetic operator

-  Can you think of a good test that we can add to ensure that regressions don't occur in the future?

tjb900:
@mrocklin Thanks for the feedback.  Have fixed up a couple of things, and added a test which fails before the fix (although it seems unlikely that a future regression would hit exactly that particular scenario).

AFAICT the remaining test failures seem to be transient issues in the CI?

distributed/protocol/tests/test_protocol.py
```diff
@@ -137,6 +137,23 @@ def test_large_messages():
     assert (msg['y']['a'].data == msg2['y']['a']).all()
 
 
+@slow
+def test_large_messages_map():
+    np = pytest.importorskip('numpy')
```
mrocklin: Do we need the numpy check here?

distributed/protocol/tests/test_protocol.py
```diff
@@ -137,6 +137,23 @@ def test_large_messages():
     assert (msg['y']['a'].data == msg2['y']['a']).all()
 
 
+@slow
```
mrocklin: This test seems to run quickly enough for me.

```python
In [1]: from distributed.protocol.tests.test_protocol import dumps, loads                                          

In [2]: %time x = {i: 'mystring_%d' % i for i in range(100000)}                                                    
CPU times: user 37.7 ms, sys: 3.42 ms, total: 41.2 ms
Wall time: 40.6 ms

In [3]: %time b = dumps(x)                                                                                         
CPU times: user 40.9 ms, sys: 99 µs, total: 41 ms
Wall time: 40.7 ms

In [4]: x2 = loads(b)                                                                                              

In [5]: %time x2 = loads(b)                                                                                        
CPU times: user 35.4 ms, sys: 324 µs, total: 35.7 ms
Wall time: 34.7 ms

In [6]: %time assert x == x2                                                                                       
CPU times: user 4.94 ms, sys: 97 µs, total: 5.04 ms
Wall time: 5.08 ms
```

We can probably drop the `@slow` decorator

distributed/protocol/tests/test_protocol.py
```diff
@@ -137,6 +137,23 @@ def test_large_messages():
...
+def test_large_messages_map():
+    np = pytest.importorskip('numpy')
+    psutil = pytest.importorskip('psutil')
```
mrocklin: It looks like we now depend on `psutil`, so we can probably drop this.

distributed/protocol/tests/test_protocol.py
```diff
@@ -137,6 +137,23 @@ def test_large_messages():
...
+    if psutil.virtual_memory().total < 8e9:
+        return
+
+    if sys.version_info.major == 2:
+        return 2
```
mrocklin: I'm curious, what doesn't work about this test in Python 2?

mrocklin:
And I agree that those failures look unrelated.  We've been plagued by
these recently.  I'll take a look at them later today.

tjb900:
Fixed.  Most of these were due to lazy copying from the test above.
Have also changed it to explicitly skip (rather than silently pass) in the low-memory situation.

mrocklin:
This looks good to me.  Thank you for resolving this @tjb900 

Also, I notice that this is your first commit to this repository.  Welcome!

rsignell-usgs:
And chance to get a new minor release to include this fix?  This broke [one of the pangeo workshop demos ](https://github.com/pangeo-data/pangeo-tutorial-agu-2018/issues/12) here at AGU and it would be nice to fix it with a `dask` update rather than a `msgpack` rollback.

Issue #2398, opened 2018-12-04
help tracking thread explosions?


danpf:
I'm having a super weird problem with dask on my SLURMcluster using dask-jobqueue but i'm confident that's not the problem.

Actual problem:
Slurm is unable to submit jobs to some nodes because my jobs are generating a highly excessive number of processes(threads)

brief workflow
1. submit script to slurm
2. script from(1) generates a SLURMcluster (aka a dask client)
3. (2) submits to SLURM and makes the actual workers

Description of problem:
I will attatch 2 logfiles from `pstree -l -c -u danpf -a`
logfile `normal_op.txt` is from normal operation
Could someone help me figure out what all of these extra python processes are?
why does each forkserver need 6 threads, and why does each worker need 13 threads?
This is a node processer load load of approx 13 or so per worker thread. this is pretty high. Ideally it would be around 6-7 (or so my admin tells me, they would prefer it to be as low as possible)

However, after a worker fails, which seems to occur for various reasons (including but not limited to):
1. the main scheduler encountering an exception and dying
2. a SIGKILL command
3. a worker dying for some reason ( not too sure I get a lot of weird dask errors sometimes)

If i'm ssh'd into a node and i track the number of processes  / threads when this happens, I see an explosion of the number of threads from ~300-500 to ~7-8000. The result from `pstree` command from above after i've killed the master(main Client) thread is `cancel_op.txt`

At this point, I can't even ssh into a node because there's too many processes, and any time SLURM tries to submit something to that node, the command it submits dies almost instantly. (I'm unaware if this is limited to my submissions, or if i'm actually causing other people's submissions to die as well)

I guess the questions I  would like answered /help answering are:
1. what are all of these threads in the first place (normal per-worker load appears to be ~13 threads/processes) (is there a way to slim this down?)
2. what could be happening to cause these 'forkserver`   processes  to generate so many threads when they lose contact with the master process?
3. Would a different multiprocessing-method help with this? it wouldn't appear so from what i've read, but figured I would ask (it looks like only spawn, fork, and forkserver are options)

[normal_op.txt](https://github.com/dask/distributed/files/2646408/normal_op.txt)
[cancel_op.txt](https://github.com/dask/distributed/files/2646409/cancel_op.txt)

Thanks for any help!!!
Dan

jhamman:
First thing I would check is if your workers are mixing with OpenMP parallelization. I would try setting OpenMP threads to 1. 

```yaml
jobqueue:
  slurm:
[...]

PR #2403, opened 2018-12-07
Fix excess threading on missing connections

danpf:
When you start a worker, and there is no scheduler at the location you expect it to be, it continuously generates threads until it hits some limit (~300 for me) and then fails.
The problem is this `while True` loop.

we can solve this by making the future outside the loop, and then tracking that one instead of all of the ones from each `while` loop.

I  - think - this is an okay way to do this, but maybe someone that knows tornado more can say otherwise.

possible fix for https://github.com/dask/distributed/issues/2398

### DIFF
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -304,7 +304,7 @@ def test_comm_failure_threading():
             thread_count = threading.active_count()
             if thread_count > max_thread_count:
                 max_thread_count = thread_count
-        return max_thread_count
+        raise gen.Return(max_thread_count)
     original_thread_count = threading.active_count()
 
     # tcp.TCPConnector()
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -312,8 +312,8 @@ def test_comm_failure_threading():
     with pytest.raises(IOError):
         yield connect("tcp://localhost:28400", 0.052)
     max_thread_count = yield sleep_future
-    # 3 is the number set by BaseTCPConnector.executor (ThreadPoolExecutor)
-    assert max_thread_count <= 3 + original_thread_count
+    # 2 is the number set by BaseTCPConnector.executor (ThreadPoolExecutor)
+    assert max_thread_count <= 2 + original_thread_count
 
     # tcp.TLSConnector()
     sleep_future = sleep_for_60ms()
@@ -321,7 +321,7 @@ def test_comm_failure_threading():
         yield connect("tls://localhost:28400", 0.052,
                                  connection_args={'ssl_context': get_client_ssl_context()})
     max_thread_count = yield sleep_future
-    assert max_thread_count <= 3 + original_thread_count
+    assert max_thread_count <= 2 + original_thread_count
 
 
 @gen.coroutine
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -305,23 +305,22 @@ def test_comm_failure_threading():
             if thread_count > max_thread_count:
                 max_thread_count = thread_count
         return max_thread_count
+    original_thread_count = threading.active_count()
 
     # tcp.TCPConnector()
-    sleep_future = gen.convert_yielded(sleep_for_500ms())
+    sleep_future = sleep_for_500ms()
     with pytest.raises(IOError):
-        future_connect = gen.convert_yielded(connect("tcp://localhost:28400", 0.6))
-        yield future_connect
+        yield connect("tcp://localhost:28400", 0.6)
     max_thread_count = yield sleep_future
-    assert max_thread_count == 4
+    assert max_thread_count - original_thread_count == 2
 
     # tcp.TLSConnector()
-    sleep_future = gen.convert_yielded(sleep_for_500ms())
+    sleep_future = sleep_for_500ms()
     with pytest.raises(IOError):
-        future_connect = gen.convert_yielded(
-                connect("tls://localhost:28400", 0.6, connection_args={'ssl_context': get_client_ssl_context()}))
-        yield future_connect
+        yield connect("tls://localhost:28400", 0.6,
+                                 connection_args={'ssl_context': get_client_ssl_context()})
     max_thread_count = yield sleep_future
-    assert max_thread_count == 4
+    assert max_thread_count - original_thread_count == 2
 
 
 @gen.coroutine
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -312,8 +312,8 @@ def test_comm_failure_threading():
     with pytest.raises(IOError):
         yield connect("tcp://localhost:28400", 0.052)
     max_thread_count = yield sleep_future
-    # 2 is the number set by BaseTCPConnector.executor (ThreadPoolExecutor)
-    assert max_thread_count <= 2 + original_thread_count
+    # 3 is the number set by BaseTCPConnector.executor (ThreadPoolExecutor)
+    assert max_thread_count <= 3 + original_thread_count
 
     # tcp.TLSConnector()
     sleep_future = sleep_for_60ms()
@@ -321,7 +321,7 @@ def test_comm_failure_threading():
         yield connect("tls://localhost:28400", 0.052,
                                  connection_args={'ssl_context': get_client_ssl_context()})
     max_thread_count = yield sleep_future
-    assert max_thread_count <= 2 + original_thread_count
+    assert max_thread_count <= 3 + original_thread_count
 
 
 @gen.coroutine
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -297,9 +297,9 @@ def test_comm_failure_threading():
     """
 
     @gen.coroutine
-    def sleep_for_120ms():
+    def sleep_for_60ms():
         max_thread_count = 0
-        for x in range(120):
+        for x in range(60):
             yield gen.sleep(0.001)
             thread_count = threading.active_count()
             if thread_count > max_thread_count:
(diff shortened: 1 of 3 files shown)


mrocklin:
Interesting.  First, thanks for taking the initiative and submitting this.  I appreciate it.

I'm a little curious about that while loop.  Under what conditions are we iterating many times?  What is the EnvironmentError that you're getting that allows the loop to continue?

I am also a little surprised that calling `connector.connect` creates a new thread.  This is all using Tornado, which should be doing everything concurrently within a single thread using non-blocking sockets.

danpf:
> I'm a little curious about that while loop. Under what conditions are we iterating many times? What is the EnvironmentError that you're getting that allows the loop to continue?

I'm not entirely sure, but if the ip exists, but there's no dask scheduler there it basically keeps checking until timeout error just in case it's busy responding to something else.

> I am also a little surprised that calling connector.connect creates a new thread. This is all using Tornado, which should be doing everything concurrently within a single thread using non-blocking sockets.

unfortunately I have no idea. I have never used tornado before, but it looks like something to do with the way it does:
```
client =  TCPClient()
try:
    stream = yield client.connect(ip, port, max_buffer_size=...
```

mrocklin:
OK, I printed out the error in the `except EnvironmentError` block, and saw a bunch of these:

```
<distributed.comm.tcp.TCPConnector object at 0x7ff27414a898>: ConnectionRefusedError: [Errno 111] Connection refused
```

Given the sleep that's there it looks like we're trying this every 10ms.  I'm not surprised that this is somewhat challenging on the system, but I am surprised that it is manifesting as too many threads. 

However, this also means that once that connection fails, we will want to start a new one, so I'm not sure that we can pull the `connect` call outside of the while loop.  If you print out the future in that loop you'll find that it has failed.  The while loop at that point becomes somewhat useless after that point.  It just keeps waiting on a failed future.

Things still work for you because at some point this function stops entirely, and something above it tries it again.

danpf:
@mrocklin 
here's why it's making threads.

```
  File "/home/danpf/git/distributed/distributed/core.py", line 573, in send_recv_from_rpc      
    comm = yield self.live_comm()              
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/gen.py", line 326, in wrapper                                                                        
    yielded = next(result)                     
  File "/home/danpf/git/distributed/distributed/core.py", line 545, in live_comm               
    connection_args=self.connection_args)      
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/gen.py", line 326, in wrapper                                                                        
    yielded = next(result)                     
  File "/home/danpf/git/distributed/distributed/comm/core.py", line 182, in connect            
    **(connection_args or {}))                 
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/gen.py", line 326, in wrapper                                                                        
    yielded = next(result)                     
  File "/home/danpf/git/distributed/distributed/comm/tcp.py", line 331, in connect             
    **kwargs)                                  
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/gen.py", line 326, in wrapper                                                                        
    yielded = next(result)                     
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/tcpclient.py", line 226, in connect                                                                  
    addrinfo = yield self.resolver.resolve(host, port, af)                                     
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/gen.py", line 326, in wrapper                                                                        
    yielded = next(result)                     
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/netutil.py", line 378, in resolve                                                                    
    None, _resolve_addr, host, port, family)   

>        result = yield IOLoop.current().run_in_executor(
>            None, _resolve_addr, host, port, family)

  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/site-packages/tornado/platform/asyncio.py", line 166, in run_in_executor                                                   
    return self.asyncio_loop.run_in_executor(executor, func, *args)                            

  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/asyncio/base_events.py", line 737, in run_in_executor                                                                      
    executor.submit(func, *args), loop=self)   

>    def run_in_executor(self, executor, func, *args):
>        print("run in executor", executor)
>        self._check_closed()
>        if self._debug:
>            self._check_callback(func, 'run_in_executor')
>        if executor is None:
>            executor = self._default_executor
>            if executor is None:
>                executor = concurrent.futures.ThreadPoolExecutor()
>                self._default_executor = executor
>        return futures.wrap_future(
>            executor.submit(func, *args), loop=self)

  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/concurrent/futures/thread.py", line 160, in submit                                                                         
    self._adjust_thread_count()                
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/concurrent/futures/thread.py", line 179, in _adjust_thread_count                                                           
    self._initargs))                           
  File "/home/danpf/.local/share/pyenv/versions/3.7.0/lib/python3.7/threading.py", line 784, in __init__                                                                                       
    traceback.print_stack()                    
```

The solution (i think) is to somehow hook distributed.comm.tcp.BaseTCPConnector.connect() up with an arg that gives the TCPClient init call
your own threadpoolexecutor that respects our desired number of threads. it will need its own `submit` function. (maybe it already has one?)

mrocklin:
Hrm, that's really interesting.  Thank you for spending the effort to track that down.

It looks like that code should create and reuse a single ThreadPoolExecutor (it seems to be attaching it to the IOLoop for future reuse).  I'm surprised that it seems to be leaking threads.  One thing you could do here if you were interested is to verify that it continues to create executors, and if so why, and then raise that upstream to core CPython (if indeed you find that it is a bug).

My guess is that it will be difficult to pass a ThreadPoolExecutor down from Dask, through Tornado, to Asyncio (indeed it looks as though Tornado is hard-coding the `executor` argument to `None`).

You might also want to try Tornado master to see if they have resolved the problem upstream.

danpf:
This new implementation __ might __ be a possible option.  Some tests still fail though, so it is probably filling up the thread queue so much that it isn't able to get through all of them.

mrocklin:
Those errors might also be intermittent (there are unfortunately a few of these).

I don't know enough about tornado internals to be able to judge this PR effectively.  Do you have any thoughts on how hard it would be to add a test for this behavior?  Presumably it would call connect on an address that needed resolving, wait for a while, and then check that there were not significantly more threads than when we started.

danpf:
`convert_yielded` appears to be the same as `ensure_future`, which basically means 'start this right asap'. but don't block on this right now, you just have to make sure you yield / await it later.

previously on an 8 core machine:
```
E       AssertionError: DONE 42
E       assert 42 == 3
```
now the test passes.

from the tornado mailing list:
https://groups.google.com/forum/#!topic/python-tornado/qfHTGOjIUuo

> This now uses the asyncio "default" executor (for consistency with the rest of the asyncio ecosystem). You can customize this with either IOLoop.set_default_executor or its asyncio equivalent: http://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.set_default_executor
> 
> Note that the default executor has a high limit, but it does have a limit, so it won't keep spawning threads forever. Last time I checked it defaults to 5 threads per CPU. In general it makes sense to allow more than one thread to perform DNS lookups, although it's hard to say what the best limit is. 
> 

so in the future if this becomes a more complicated problem, it might be necessary to set the loop default executor to a specified number of threads, but for now allocating 2 threads to comms seems fine... I think 1 would also be fine, but just to be cautious....

distributed/comm/tests/test_comms.py
```diff
@@ -289,6 +289,41 @@ def client_communicate(key, delay=0):
...
+                max_thread_count = thread_count
+        return max_thread_count
+
+    # tcp.TCPConnector()
+    sleep_future = gen.convert_yielded(sleep_for_500ms())
+    with pytest.raises(IOError):
+        future_connect = gen.convert_yielded(connect("tcp://localhost:28400", 0.6))
+        yield future_connect
+    max_thread_count = yield sleep_future
+    assert max_thread_count == 4
```
mrocklin: I recommend testing that no more than a few threads were created since starting this test.  Otherwise this test will fail if some other test leaks a thread (which does happen with a few of our tests).

distributed/comm/tests/test_comms.py
```diff
@@ -289,6 +289,41 @@ def client_communicate(key, delay=0):
...
+        max_thread_count = 0
+        for x in range(500):
+            yield gen.sleep(0.001)
+            thread_count = threading.active_count()
+            if thread_count > max_thread_count:
+                max_thread_count = thread_count
+        return max_thread_count
+
+    # tcp.TCPConnector()
+    sleep_future = gen.convert_yielded(sleep_for_500ms())
```
mrocklin: I *think* that you can drop the `gen.convert_yielded` here.  Tornado semantics are a bit different from normal async-def semantics.  `gen.coroutine` objects start running immediately once called.

This is also true below with the `future_connect` lines.  I would merge them into a single `yield connect(...)` statement.

danpf:
@mrocklin 
I reduced it to 0.2s on my computer. 
we make an attempt (thread) every 10ms, so 5 attempts should be good enough to prove that it's working.

mrocklin:
The Python 2 failure is interesting.  I don't have a good idea of what is going on there unfortunately.  Do you have any idea about what's going on?

danpf:
darn hrm I don't know what to do since it still fails...  I haven't used python 2 in a while so i don't really know what's different...
image

mrocklin:
So if I bump the thread count up to 100 then things run smoothly, so maybe there is some issue deep within an old version of Tornado. 

My suggestion at this point would be to only assign a client to `BaseTCPConnector` in Python 3 and leave it as None otherwise.  Then do something like `client = BaseTCPConnector.client or TCPClient()` in the `connect` coroutine.  In Python 3 we'll do the right thing.  In Python 2 we'll do previous behavior.  This should solve your issue and be fixable in finite time.

Eventually when we drop Python 2 we can clear this out.

danpf:
thank you for fixing my silly extra inclusion!

mrocklin:
Merging.

Thanks for identifying and resolving this issue.  I hope that you run into many more so that you can fix them ;)

Issue #2401, opened 2018-12-06
Starting local cluster from within an embedded interpreter


uellue:
I've experienced this issue with the Python interpreter in a beta build for the upcoming Gatan Digital Micrograph release in particular, but it is probably applicable to other cases where a Python interpreter is embedded.

Starting a process-based local cluster with the [normal method](https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/executor/dask.py#L130) spawns new instances of the host application, at least on Windows. That is not the intended behavior when starting a cluster. As far as I could see, this is a result of the way how the `multiprocessing` module kicks off new processes.

As an alternative solution, we've developed a launch method that uses the `subprocess` module: https://github.com/LiberTEM/LiberTEM/blob/master/src/libertem/executor/dask.py#L148 

Since this is something not specific to LiberTEM, but any dask.distributed user from within an embedded interpreter, could this be something useful to add to dask.distributed? 

The code in LiberTEM is not 100 % perfect yet because it tends to open a swarm of temporary shell windows, but at least it works, [including in our tests for Linux and Windows Continuous Integration](https://github.com/LiberTEM/LiberTEM/blob/master/tests/test_local_cluster.py). Probably some Windows-specific parameters have to be set for `subprocess.Popen()` to prevent the stray shell windows.

itamarst:
I believe the solution is to use https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_executable - you can set a different Python interpreter for `multiprocessing` to use.

itamarst:
`multiprocessing` already has infrastructure for starting subprocesses, so just seems like extra maintenance work? There might be another use case, but if the suggestion above works this isn't it.

Perhaps the above should be referenced in Distributed documentation, though.

uellue:
Hi, just FYI, I unfortunately lost the possibility to test this because of other issues with the testing environment for the embedded interpreter. I hope I can get back to this later as soon as it works again.

uellue:
I confirmed that the method suggested by @itamarst works for my case. Thank you again! In #2409 there's a short info on that for the documentation.

PR #2409, opened 2018-12-12
Include information to launch local cluster from embedded interpreter

uellue:
See #2401

### DIFF
--- a/docs/source/local-cluster.rst
+++ b/docs/source/local-cluster.rst
@@ -27,6 +27,16 @@ an ``Client`` with no arguments:
    >>> client = Client()
    >>> client
    <Client: scheduler=127.0.0.1:8786 processes=8 cores=8>
+   
+Embedded interpreter
+--------------------
+
+You have to set the `correct executable for multiprocessing <https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_executable>`_ to launch a local cluster from within an embedded Python interpreter. Otherwise attempting to spawn workers will spawn new instances of the embedding application instead.
+
+.. code-block:: python
+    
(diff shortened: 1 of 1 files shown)


uellue:
Hm, are the failing tests somehow related to the documentation change? So far I could see only `distributed/tests/test_client.py::test_normalize_collection <- distributed/utils_test.py FAILED`, which doesn't seem to be related to documentation.

mrocklin:
Thanks for adding this @uellue and my apologies for the late review.  The text here seems fine, my only concern is that we're elevating what I suspect is a rare problem to a high level in the documentation.  I suspect that most people (99%?) won't need this bit of documentation, and yet that many people (20%?) who look for docs on `LocalCluster` will see this and think that they have to do something special and be intimidated by it.  

There is an odd tension between placing gotcha's like this in places where people will find them while also avoiding having them clutter up mainline code.  

I'm not sure how best to handle this.  Maybe a separate errata doc?  What are your thoughts?

uellue:
Hi @mrocklin, for me the most important thing is that Google finds it. Perhaps it would fit in a general section on information how distributed interacts with multiprocessing?

Where would be a good place to put this information?

mrocklin:
Perhaps a miscelleneous.rst documentation file that explans the situation of needing many small tidbits?  Honestly I'm not sure if that's a good idea either.  If anyone has any suggestions on how to solve this problem generally they would be welcome.

Another solution altogether is not to handle this in documentation, but to ask and answer a question on stack overflow.  I've done that in the past for particularly specialized questions.

uellue:
Nevermind, then. I'm closing this.

PR #2412, opened 2018-12-12
Add a blocking argument to Lock.acquire()

shoyer:


mrocklin:
This failure seems to be coming up in other PRs.  I'll track it down.

PR #2415, opened 2018-12-14
Fix #2312

### LINKED ISSUE #2312: Is bandwidth meant to be hard-coded for work stealing?
The network bandwidth parameter is hard-coded in [stealing.py](https://github.com/dask/distributed/blob/master/distributed/stealing.py):
`BANDWIDTH = 100e6`
as opposed to in [scheduler.py](https://github.com/dask/distributed/blob/master/distributed/scheduler.py) where it is set by the config:
`BANDWIDTH = dask.config.get('distributed.scheduler.bandwidth')`
Is there any particular reason it is hard-coded for work-stealing?

rainwoodman:
Doodling changes suggested in https://github.com/dask/distributed/issues/2312

I find the current stealing is too conservative.
I have a single worker with 1000 slow IO jobs yet 127 workers sitting idle.
This change would allow some customization -- e.g. I do know the bandwidth of the machine is way higher than 100 MB/s. more like 6GB/s, with latency a few hundred us. (https://www.alcf.anl.gov/user-guides/aries-network-xc40)

screenshot from 2018-12-13 17-07-38


distributed/stealing.py
```diff
@@ -15,8 +15,8 @@
 except ImportError:
     from toolz import topk
 
-BANDWIDTH = 100e6
-LATENCY = 10e-3
+BANDWIDTH = dask.config.get('distributed.scheduler.bandwidth')
+LATENCY = dask.config.get('distributed.scheduler.latency')
```
mrocklin: You'll have to add this to the `distributed.yaml` file that contains defaults.

distributed/config.py
```diff
@@ -24,6 +24,7 @@
 aliases = {
     'allowed-failures': 'distributed.scheduler.allowed-failures',
     'bandwidth': 'distributed.scheduler.bandwidth',
+    'latency': 'distributed.scheduler.latency',
```
mrocklin: This shouldn't be necessary.  These aliases are mostly just used for backwards compatibility with older versions.

quasiben:
@mrocklin @jcrist can you review and merge ?

distributed/distributed.yaml
```diff
@@ -21,6 +21,7 @@ distributed:
     worker-ttl: null        # like '60s'. Time to live for workers.  They must heartbeat faster than this
     preload: []
     preload-argv: []
+    latency: .01            # used in calculation of transfer_time for work stealing
```
mrocklin: It would be good to use times like "10 ms" here instead of float values.  (and then use `distributed.utils.parse_timedelta`)

mrocklin: Similarly bandwidth should be something like `100 MiB` and we should use `distributed.utils.parse_bytes`

quasiben:
Some of this PR has since be superseded by https://github.com/dask/distributed/pull/2658.  If @rainwoodman are you still interested in adding in a latency configuration option ?

rainwoodman:
I haven't been using dask recently; so no. Feel free to take over!

Issue #2420, opened 2018-12-17
potential issue with worker.release_dep


tjb900:
Hi!

Bringing this up here to get some feedback on how you think it should be handled - I'm happy to write a fix but obviously don't want to go in a direction that won't be accepted.

We are working in an adaptive-scaling environment, and are hitting a race condition where sometimes workers don't have what the scheduler thinks they have - and thus frees data that it shouldn't.  After a lot of time spent tracing and digging, this is, I think what is happening:

0. something (e.g. a client, or a future dependent task which is not yet eligible) needs `task1` to be kept alive
1. `worker1` computes `task1` (state: `memory`)
2. `worker2` starts work on a `task2` which depends on `task1`.  It obtains the dependency from worker1, and tells the scheduler (https://github.com/dask/distributed/blob/879cdb35c52135cd53ff89c36c485311b7801004/distributed/worker.py#L1643 or https://github.com/dask/distributed/blob/879cdb35c52135cd53ff89c36c485311b7801004/distributed/worker.py#L1226) that it has it (this notification can be delayed, however).
3. we try to scale the cluster down, and (for whatever reason) decide to close `worker1`.  The scheduler sees that both `worker1` and `worker2` have `task1`, and so does not make a copy anywhere else.
4. `worker1` closes; scheduler now has `task1.who_has = {worker2}`.
5. `worker2` then completes `task2`, and frees the dependency (`task1`) (in `release_dep`), and tells the scheduler that it freed it (but again, this notification can be delayed)
6. when the notification arrives at the scheduler that `worker2` has released `task1`, it causes it to be recomputed (if it can be - if `task1` was scattered data this would then be an error).

I added a log message to `worker.py` to see occurrences of these:

```
@@ -939,6 +939,13 @@ class Worker(ServerNode):
                 if k in self.actors:
                     from .actor import Actor
                     data[k] = Actor(type(self.actors[k]), self.address, k)
+                else:
+                    logger.warning(
+                        "peer %s asked me for key %s but I don't have it!",
+                        str(comm.peer_address),
+                        str(k))
+                    for rel in [l for l in self.log if k in l]:
+                        logger.warning("relevant log entry: %s", str(rel))
 
         msg = {'status': 'OK',
                'data': {k: to_serialize(v) for k, v in data.items()}}
```

leading to thousands of messages like (edited):

```
[...]

PR #2421, opened 2018-12-18
Don't make False add-keys report to scheduler

tjb900:
Possible fix for #2420, for now putting here mostly to trigger CI.

Cleans up (imho) a couple of things in the worker:
 * responsibility for maintaining the `in_flight_tasks` / `in_flight_workers` maps is entirely contained within the `transition_dep_*` functions
 * better handles the case where dependencies are released (due to e.g. stealing) after they are marked as in-flight but before or during the transfer.  In particular, removes an `add-keys` notification to the scheduler for keys that we might then throw in the trash due to being no-longer needed.


mrocklin:
Thank you for submitting this @tjb900 

It looks like there are a few smaller changes here.  In general we're pretty conservative with changes to the scheduling logic, so I encouarage you to isolate changes into different PRs if possible and also demonstrate that they improve concrete situations with tests if possible.  Often constructing a minimal test that shows off failing behavior can be difficult, but is quite useful to ensure that future changes don't subtly break things again.

tjb900:
Will do @mrocklin - thanks for the feedback.  Will keep this PR for the removal of the add-keys only.

mrocklin:
I think that we still want to inform the scheduler that we have this data in the common case that we decide to store it.

tjb900:
I'm pretty sure that's done inside `transition_dep_flight_memory`?

tjb900:
(still planning to come up with a test, too - so far I have just removed all of the other changes)

distributed/tests/test_worker.py
```diff
@@ -1271,3 +1271,50 @@ def test_startup2():
...
+    assert not a.executing
+    assert x.key in a.data
+
+    def bogus_task(bogus_input):
+        import time
+        time.sleep(1)
+        return 2
+
+    y = c.submit(bogus_task, x, workers=b.address)
+    yield c._cancel(y)
```
mrocklin: You can safely drop the underscore here.  The public functions will operate asynchronously in an asynchronous context.

mrocklin: I'm slightly concerned that `y` may never start.  Should we wait until `b` is executing something?

```python
while not b.executing:
    yield gen.sleep(0.01)
```

distributed/tests/test_worker.py
```diff
@@ -1271,3 +1271,50 @@ def test_startup2():
...
+
+    # The notification to the scheduler happens in a batched send
+    # Make sure it has time to arrive
+    yield gen.sleep(5)
+    try:
+        yield wait(y)
+    except:
```
mrocklin: I recommend `except Exception:` otherwise this catches `KeyboardInterrupts`

distributed/tests/test_worker.py
tjb900: Actually, I think it's important that `y` never starts - we need to cause `x` to be transferred from `a` to `b`, but then have `y` be cancelled while that transfer is occurring - so that `x` is never actually placed in memory on `b`.  This timing is hard to arrange in the test (and I suspect actually now even harder with #2428),

distributed/tests/test_worker.py
tjb900: Yeah, I agree this `except` is way too broad actually - it's really for the `CancelledError`.  Will fix.

tjb900:
@mrocklin tbh, I'm a little concerned that this test is going to end up so targeted at the current implementation of the worker that it is highly unlikely to catch a future regression.  On top of this, because it depends on the internal details of the worker rather than some kind of semi-stable API, it comes with a maintenance burden of updating the test to match any logic changes inside the worker.

Is there instead a longer running "test everything" test which we could augment to catch this situation?  For instance, by verifying after a large amount of work that the scheduler and workers' ideas of what they all have are consistent with each other?

mrocklin:
For `gather_dep` it might work to look at `Worker.in_flight_workers`.  This will be populated while the dependency is possibly in flight, although there is definitely a period when it's in `in_flight_workers` before the transfer has started.

> but note that we need to arrange for the cancellation notice from s to reach b before the data from a does

As a warning, you should count on random 3ish second delays from time to time.  A combination of slow containers on travis-ci and some intense GC can play havoc with any test that is depending on something happening within a finite window.

### Another approach?

I've tended to avoid intriciately orchestrated tests for this reason, and instead tend to try to find statistical tests.  Is there something we could do quickly a hundred times that would have some measurable negative effect?  

In particular, I wonder if we might draw inspiration from your original concern:

>  We are working in an adaptive-scaling environment, and are hitting a race condition where sometimes workers don't have what the scheduler thinks they have - and thus frees data that it shouldn't.

Is there a lightweight system we could put together that would simulate this?  Maybe we start and stop workers very quickly and submit and cancel tasks very quickly at the same time?  Then we verify that at the end (or throughout) that everything is as it should be?  My guess is that you could construct a system that would test what you've done here, and probably test a bunch of other things at the same time as well and have it finish in less time.  A test like this might also survive future refactorings and keep us honest for longer.  

Thoughts?

mrocklin:
Whoops, I wrote my answer while you were writing yours.  Sounds like we came to the same conclusion independently :)

mrocklin:
> Is there instead a longer running "test everything" test which we could augment to catch this situation? For instance, by verifying after a large amount of work that the scheduler and workers' ideas of what they all have are consistent with each other?

There are a variety of such tests in `test_stress.py`, but obviously none of them have caught this :)

I think that a new stress test that focused on highly adaptive clusters would be useful generally if you're interested in cooking something up (no obligation though).

mrocklin:
Checking in here @tjb900 .  Any thoughts?

tjb900:
Definitely keen to work on a test that works out the adaptive functionality - but am snowed at the moment, so it's kind of on hold sorry.

jrbourbeau:
Checking in here, @tjb900 just wondering what the status is on this PR. Do you plan to pick things up again?

tjb900:
Hi @jrbourbeau - sorry I've left this languishing for so long.  Realistically, I don't think there's a chance I'll get to it sorry.  Happy to close.

jrbourbeau:
I removed `test_worker_keeps_data_gh2420` and merged the current `master` branch into this branch, hope that's okay @tjb900

The proposed changes seem reasonable to me (I discussed this issue with @mrocklin yesterday). @mrocklin, when you get a moment, would you mind taking another look at this PR

PR #2453, opened 2019-01-08
Replace register_worker_callbacks with worker plugins

mrocklin:
These provide space for a bit more complexity, including teardown and optional idempotence.  This is still a work in progress.  Some things I'd like to do:

- [ ] deprecate `register_worker_callbacks`
- [ ] support simple setup functions in `register_worker_plugins`
- [ ] improve error handling in broadcast

cc @rainwoodman and @guillaumeeb .  Possible replacement for #2391


distributed/client.py
```diff
@@ -3593,7 +3593,57 @@ def register_worker_callbacks(self, setup=None):
...
+        return self.register_worker_plugin(_WorkerSetupPlugin(setup))
+
+    def register_worker_plugin(self, plugin=None, name=None):
+        """
+        Registers a lifecycle worker plugin for all current and future workers.
+
+        This registers a new object to handle setup and teardown handling for
+        workers in this cluster. The function will run immediately on all
+        currently connected workers. It will also be run upon connection by any
+        workers that are added in the future. Multiple setup functions can be
```
rainwoodman: 'Multiple setup functions'? You meant multiple plugins. The setup function will be in order, but shouldn't the teardowns to be in reverse?

distributed/worker.py
```diff
@@ -1971,6 +1975,29 @@ def run_coroutine(self, comm, function, args=(), kwargs=None, wait=True):
         return run(self, comm, function=function, args=args, kwargs=kwargs,
                    wait=wait)
 
+    @gen.coroutine
+    def plugin_add(self, comm=None, plugin=None, name=None):
```
rainwoodman: There are currently two ways of specifying the plugin name. I wonder if there really is a use case for the name attribute of a Plugin object.

rainwoodman:
I like this better than #2391.

distributed/worker.py
guillaumeeb: No strong opinion here.

distributed/client.py
mrocklin: Fixed.  Thanks both.

distributed/worker.py
mrocklin: I definitely have use cases for the name attribute.  Often the author of a plugin knows more about how it should be used than the user of it.

distributed/client.py
```diff
@@ -3593,7 +3593,57 @@ def register_worker_callbacks(self, setup=None):
...
             Function to register and run on all workers
         """
-        return self.sync(self._register_worker_callbacks, setup=setup)
+        return self.register_worker_plugin(_WorkerSetupPlugin(setup))
+
+    def register_worker_plugin(self, plugin=None, name=None):
+        """
+        Registers a lifecycle worker plugin for all current and future workers.
+
+        This registers a new object to handle setup and teardown handling for
```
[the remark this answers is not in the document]
mrocklin: Thanks.  Fixed

distributed/client.py
```diff
@@ -3593,7 +3593,57 @@ def register_worker_callbacks(self, setup=None):
...
+        ...     def teardown(self):
+        ...         pass
+
+        >>> plugin = MyPlugin(1, 2, 3)
+        >>> client.register_worker_plugin(plugin)
+
+        Parameters
+        ----------
+        setup : callable(dask_worker: Worker) -> None
+            Function to register and run on all workers
```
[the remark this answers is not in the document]
mrocklin: Thanks. Fixed.

mrocklin:
@jcrist should you find yourself with a moment this could also use your review

mrocklin:
Docs added in https://github.com/dask/dask/pull/4833

mrocklin:
I plan to merge this this afternoon if there are no further comments

distributed/client.py
```diff
@@ -3593,7 +3593,57 @@ def register_worker_callbacks(self, setup=None):
...
+        with the following signature.
+
+        Examples
+        --------
+        >>> class MyPluin:
+        ...     def __init__(self, *args, **kwargs):
+        ...         pass  # the constructor is up to you
+        ...     def setup(self, worker: dask.distributed.Worker):
+        ...         pass
+        ...     def teardown(self):
```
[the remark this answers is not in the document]
mrocklin: done

distributed/client.py
```diff
@@ -3876,7 +3866,85 @@ def register_worker_callbacks(self, setup=None):
...
+                exc = response["exception"]
+                typ = type(exc)
+                tb = response["traceback"]
+                six.reraise(typ, exc, tb)
+        raise gen.Return(responses)
+
+    def register_worker_plugin(self, plugin=None, name=None):
+        """
+        Registers a lifecycle worker plugin for all current and future workers.
+
```
TomAugspurger: General question: do you have thoughts on using `..versionadded:: <version>` directives? Don't need to do it here necessarily, but I ask since this is being referenced from the dask documentation, which may be on a different release cycle.

mrocklin: No strong thoughts.  I've never used them much, either as an author or reader.

My plan was just to merge the doc PR after this gets released.  

Happy with whatever though.

mrocklin: Ah, I see that you've since merged the doc PR :)  Shouldn't be a big deal either way

distributed/tests/test_worker_plugins.py
```diff
@@ -0,0 +1,68 @@
...
+
+@gen_cluster(client=True, worker_kwargs={"plugins": [MyPlugin(5)]})
+def test_create_on_construction(c, s, a, b):
+    assert len(a.plugins) == len(b.plugins) == 1
+    assert a._my_plugin_status == "setup"
+    assert a._my_plugin_data == 5
+
+
+@gen_cluster(client=True, worker_kwargs={"plugins": [MyPlugin(5)]})
+def test_idempotence_with_name(c, s, a, b):
```
TomAugspurger: Are the names of these two tests flipped (this one doesn't use plugin names, the other does?)

mrocklin: I think that we're good here.  The `MyPlugin` class has a `name` attribute, which is what causes the idempotence here.

In the test below, we change the name intentionally.

TomAugspurger: Ah, understood.

mrocklin:
Woot.  Thanks for merging @TomAugspurger !  I think that some version of this PR has existed for years now :)

PR #2465, opened 2019-01-16
Check for non-zero `length` first in `read` loop

jakirkham:
Since both the Python 2 and Python 3 cases need to handle the case of zero and non-zero length separately, pull this check outside of the Python 2 or 3 version checks to simplify the code a bit.


mrocklin:
Looks good to me.  Thanks @jakirkham !

PR #2472, opened 2019-01-21
DOC/API: Signature for base class write / read

TomAugspurger:
https://github.com/dask/distributed/blob/b818d788deea53b9d6a5fff332d92dcda3084b76/distributed/core.py#L455-L459
will unconditionally pass these arguments, so they should likely be
present in the base class's signature. I think they're generally good to
support if possible.


mrocklin:
Thanks @TomAugspurger

Issue #2473, opened 2019-01-21
Closing a client sets dask.config 'shuffle' and 'scheduler' to None


gsakkis:
```
import dask
import dask.dataframe as dd
import distributed
import pandas as pd
import numpy as np


def set_index(msg=''):
    print("{} shuffle: {}".format(msg, dask.config.get('shuffle', 'MISSING')))
    print("{} scheduler: {}".format(msg, dask.config.get('scheduler', 'MISSING')))
    ddf.set_index('A').compute()


if __name__ == '__main__':
    df = pd.DataFrame({'A': np.random.rand(100)})
    ddf = dd.from_pandas(df, npartitions=2)
    set_index("before client")
    with distributed.Client():
        set_index("inside client")
    set_index("after client")
```
```
$ python tmp/dask/shuffle_none.py 
before client shuffle: MISSING
before client scheduler: MISSING
inside client shuffle: tasks
inside client scheduler: dask.distributed
after client shuffle: None
after client scheduler: None
Traceback (most recent call last):
...
NotImplementedError: Unknown shuffle method None
```

mrocklin:
Thank you for the excellent error report @gsakkis .  This makes it very clear what is going on.  

This can be fixed in two places.

1.  We should probably remove the shuffle entry in dask.distributed if the value is None, rather than set it to None.  The relevant lines are here: 
    https://github.com/dask/distributed/blob/90758dcd79182ff7ce0cce4b4c29d46316349836/distributed/client.py#L1078-L1079

2.  Make dask.dataframe and dask.bag robust to shuffle being None, rather than missing

https://github.com/dask/dask/blob/27231f7f7f73ae2425ca8784412f0b3765d384b6/dask/dataframe/shuffle.py#L237

Probably something like 

```python
shuffle = shuffle or config.get('shuffle', None) or 'disk'
```

There might also be something for dask.bag as well

@gsakkis do you have any interest in implementing either or both of these?  (both preferred) . This should be relatively easy to do.

gsakkis:
I'll give it a shot though I'm not sure about the tests; at least for `distributed` they're too slow and quite a few of them fail (that's after pinning pytest to 3.x; with 4.x they don't run at all).

gsakkis:
> There might also be something for dask.bag as well

`dask.bag` actually guards against None and additionally determines shuffle based on scheduler:
```
        if shuffle is None:
            if 'distributed' in config.get('scheduler', ''):
                shuffle = 'tasks'
            else:
[...]

PR #2475, opened 2019-01-22
Properly reset dask.config scheduler and shuffle when closing the client

gsakkis:
Fixes https://github.com/dask/distributed/issues/2473

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -606,11 +606,8 @@ class Client(Node):
 
         self._start_arg = address
         if set_as_default:
-            self._previous_scheduler = dask.config.get('scheduler', None)
-            dask.config.set(scheduler='dask.distributed')
-
-            self._previous_shuffle = dask.config.get('shuffle', None)
-            dask.config.set(shuffle='tasks')
+            self._set_config = dask.config.set(scheduler='dask.distributed',
+                                               shuffle='tasks')
 
         self._stream_handlers = {
             'key-in-memory': self._handle_key_in_memory,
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -1074,9 +1071,9 @@ def _close(self, fast=False):
...
             self._scheduler_identity = {}
             with ignoring(AttributeError):
-                dask.config.set(scheduler=self._previous_scheduler)
-            with ignoring(AttributeError):
-                dask.config.set(shuffle=self._previous_shuffle)
+                # clear the dask.config set keys
+                with self._set_config:
+                    pass
             if self.get == dask.config.get('get', None):
                 del dask.config.config['get']
```
mrocklin: What if the user had specified something prior to starting the scheduler, like 

```python
dask.config.set(scheduler='threads')
with Client():
    pass
assert dask.config.get('scheduler') == 'threads'
```

mrocklin: Same comment with shuffle.

gsakkis: Just tested it, the assertion passes; the `with self._set_config` restores the previous settings at exit.

mrocklin: OK, you're right.  That's a nice solution.

mrocklin:
This looks good to me.  It looks like there is a test failure though.  You might have to slightly modify that test.

gsakkis:
I fixed the related test, the others that fail randomly are probably irrelevant (`test_io_loop_periodic_callbacks` on python 3.5 at one run, `test_log_tasks_during_restart` on 3.7 at another, with no code diff between them).

mrocklin:
Thanks @gsakkis !

Merging.

mrocklin:

Issue #2273, opened 2018-09-28
Spurious CancelledError on LocalCluster.close()


adamklein:
The following repro produces a spurious log error from tornado/asyncio on closing the LocalCluster:

```
[00:36:18.413 ERROR  ] Exception in Future <Future cancelled> after timeout
Traceback (most recent call last):
  File "../lib/python3.5/site-packages/tornado/gen.py", line 970, in error_callback
    future.result()
  File "../lib/python3.5/asyncio/futures.py", line 286, in result
    raise CancelledError
concurrent.futures._base.CancelledError
```

Here is the repro:

```
from joblib import parallel_backend

import distributed.joblib
from distributed import Client, LocalCluster


def run_example():
    lc = LocalCluster(n_workers=1)
    client = Client(lc)
    with parallel_backend('dask.distributed', scheduler_host=client.scheduler.address):
        pass
    client.close()
    lc.close()


if __name__ == "__main__":
    run_example()
```

Couldn't figure out how to silence or what exactly is wrong here. Appreciate any guidance.

Distributed versions is 1.23.1 on python 3.5.6 and joblib 0.12.3

```
dask                      0.19.1                   py35_0  
dask-core                 0.19.1                   py35_0  
tornado                   5.1              py35h14c3975_0 
```

mrocklin:
I'm getting a slightly different error locally (although I'm also on more recent versions):

```pytz
future: <Future finished exception=CommClosedError('in <closed TCP>: Stream is closed',)>
Traceback (most recent call last):
  File "/home/mrocklin/workspace/distributed/distributed/comm/tcp.py", line 177, in read
    n_frames = yield stream.read_bytes(8)
  File "/home/mrocklin/workspace/tornado/tornado/gen.py", line 1133, in run
    value = future.result()
tornado.iostream.StreamClosedError: Stream is closed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mrocklin/workspace/tornado/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
  File "/home/mrocklin/workspace/distributed/distributed/comm/tcp.py", line 198, in read
    convert_stream_closed_error(self, e)
  File "/home/mrocklin/workspace/distributed/distributed/comm/tcp.py", line 126, in convert_stream_closed_error
    raise CommClosedError("in %s: %s" % (obj, exc))
distributed.comm.core.CommClosedError: in <closed TCP>: Stream is closed
```

[...]

PR #2477, opened 2019-01-23
quiet cancelledError for scheduler shutdown

danpf:
`gen.with_timeout` unfortunately permanently adds a callback to futures, and when the timeout is over, it doesn't remove that callback.

with python asyncio futures, if you get the result of them when they are cancelled they raise `CancelledErrors`.....  This isn't caught by `error_callback` in `with_timeout`, and so it is printed.

we cancel the futures a few lines down.

for reference: (the tornado callback)

```
    def error_callback(future):
        try:
            future.result()
        except Exception as e:
            if not isinstance(e, quiet_exceptions):
                app_log.error("Exception in Future %r after timeout",
                              future, exc_info=True)
```

possible fix for https://github.com/dask/distributed/issues/2273

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1090,7 +1090,8 @@ class Client(Node):
             # This makes the shutdown slightly smoother and quieter
             with ignoring(AttributeError, gen.TimeoutError):
                 yield gen.with_timeout(timedelta(milliseconds=100),
-                                       self._handle_scheduler_coroutine)
+                                       self._handle_scheduler_coroutine,
+                                       CancelledError)
 
             if self.scheduler_comm and self.scheduler_comm.comm and not self.scheduler_comm.comm.closed():
                 yield self.scheduler_comm.close()


mrocklin:
Thanks for the possible fix @danpf .  Does this resolve the problem for you?  Lets see if others on that issue have any feedback.

danpf:
yes, I started getting this problem a few days ago, not really sure why.... now i don't see the error anymore.

distributed/client.py
```diff
@@ -1090,7 +1090,8 @@ def _close(self, fast=False):
             # This makes the shutdown slightly smoother and quieter
             with ignoring(AttributeError, gen.TimeoutError):
                 yield gen.with_timeout(timedelta(milliseconds=100),
-                                       self._handle_scheduler_coroutine)
+                                       self._handle_scheduler_coroutine,
+                                       CancelledError)
```
lesteve: Nitpicks:
* maybe add the parameter name for the last argument since it is optional in `gen.with_timeout`
* maybe use a tuple of length 1 to be closer to the `gen.with_timeout` signature (using `CancelledError` works because the argument is only used in a `isinstance` call)

i.e. something like this:
```py
yield gen.with_timeout(..., quiet_exceptions=(CancelledError,))`
```

lesteve:
Here is a test, as I tried to explain in the comment it passes by chance on master in about 10% of the cases (at least on my machine ...).

* in case it is more convenient, I opened #2485 with the fix in this PR, the test below and 
the fix for the nitpick I had above
* You may want to double-check whether I used the `loop` fixture correctly. I used it both in `LocalCluster` and `Client` but I have to admit I don't understand the subtleties involved here.
* The test takes about 4s, not sure it should be labelled slow or not

```py
def test_quiet_client_close_when_cluster_is_closed_before_client(loop):
    n_attempts = 5
    # Trying a few times to reduce the flakiness of the test. Without the bug
    # fix in #2477 and with 5 attempts, this test passes by chance in about 10%
    # of the cases.
    for _ in range(n_attempts):
        with captured_logger(logging.getLogger('tornado.application')) as logger:
            cluster = LocalCluster(loop=loop)
            client = Client(cluster, loop=loop)
            cluster.close()
            client.close()

        out = logger.getvalue()
        assert 'CancelledError' not in out
```

mrocklin:
Thanks @danpf and @lesteve .   This is in.

PR #2478, opened 2019-01-23
Pytest 4 support for tests

### LINKED ISSUE #2476: Support pytest 4
Our tests currently fail with pytest 4.  

This is probably because pytest changed how they handle `pytest.mark`s. 

This was handled in Dask with 8bd87a42cc59ff2f50633cbfca6805a01843e2fe

beberg:
Fixes #2476

### DIFF
--- a/distributed/tests/test_resources.py
+++ b/distributed/tests/test_resources.py
@@ -277,9 +277,10 @@ def test_full_collections(c, s, a, b):
 
 
 @pytest.mark.parametrize('optimize_graph', [
-    pytest.mark.xfail(True, reason="don't track resources through optimization"),
-    pytest.mark.skipif(WINDOWS, False, reason="intermittent failure"),
-])
+    pytest.param(True, marks=[
+        pytest.mark.xfail(True, reason="don't track resources through optimization"),
+        pytest.mark.skipif(WINDOWS, False, reason="intermittent failure")])
+    ])
 def test_collections_get(client, optimize_graph, s, a, b):
     da = pytest.importorskip('dask.array')
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,15 @@
 *.pyc
 *.py~
 build/
+dist/
 *.egg-info
 docs/build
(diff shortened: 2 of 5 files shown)


distributed/tests/test_resources.py
```diff
@@ -277,9 +277,10 @@ def test_full_collections(c, s, a, b):
 
 
 @pytest.mark.parametrize('optimize_graph', [
-    pytest.mark.xfail(True, reason="don't track resources through optimization"),
-    pytest.mark.skipif(WINDOWS, False, reason="intermittent failure"),
-])
+    pytest.param(True, marks=[
```
beberg: Not 100% on this syntax for combining multiple marks. It's marks with an "S", but I cannot find docs/example of more than one.

TomAugspurger: This looks correct. You can pass a single mark (not wrapped in a list) to `marks` as well.

mrocklin:
In general this looks good to me.  CI is showing two failures.  One seems to be unrelated and probably due to a conda issue (will likely fix itself when we next run CI)

The other appears to be a flake8 issue

    distributed/tests/test_resources.py:283:5: E123 closing bracket does not match indentation of opening bracket's line

(you can see this by clicking "Details" next to the travis-ci line above and then looking at the last line that says `LINT=true`)

mrocklin:
You can also verify flake8 locally by running 

    $ flake8 distributed

Much like how you would run py.test

    $ py.test distributed

distributed/tests/test_resources.py
beberg: It was originally two tests cases I misread and combines. Fixed in latest push.

beberg:
Python 3.5 is unhappy with pytest 4.1.1, looks like pytest 3.8.1 is the last one to have a satisfiable `python >=3.5,<3.6.0a0`.

mrocklin:
Short term I recommend that we just unpin pytest and let the solver figure it out.

PR #2482, opened 2019-01-25
Add client.profile(filename='myfile.html') option

mrocklin:
This allows users to save profile results directly to a sharable file

### DIFF
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -311,7 +311,7 @@ class ProfilePlot(DashboardComponent):
         state = profile.create()
         data = profile.plot_data(state, profile_interval)
         self.states = data.pop('states')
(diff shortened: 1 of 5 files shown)


mrocklin:
cc @TomAugspurger I don't have this for the profilers running on the event loops yet, but can probably get there soonish.  (which might be helpful for UCX profiling)

TomAugspurger:
Thanks. A sentence in the docs under http://distributed.dask.org/en/latest/diagnosing-performance.html#statistical-profiling would be helpful.

mrocklin:
Done

martindurant:
I thought there was also an intent here to use the same code to profile any process, either where the client is running or anything else, for general usefulness. Did I misunderstand, or did that end up elsewhere?

PR #2499, opened 2019-02-01
Create Prometheus Endpoint

### LINKED ISSUE #2332: Create Prometheus Endpoint
We might consider publishing metrics at an endpoint for [Prometheus](https://prometheus.io).  This would allow standard logging and visualization tools to easily get some information about Dask.

cc @jacobtomlinson 

It might be helpful to look at [our code to host normal static pages](https://github.com/dask/distributed/blob/master/distributed/bokeh/scheduler_html.py)

beberg:
Fixes #2332

Added as /metrics on the scheduler's port 8787. If that server is moved to another prefix Prometheus will break.

### DIFF
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -27,6 +27,7 @@ class Workers(RequestHandler):
         with log_errors():
             self.render('workers.html',
                         title='Workers',
+                        scheduler=self.server,
                         **toolz.merge(self.server.__dict__, ns, self.extra))
 
 
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -165,6 +165,7 @@ class IndividualPlots(RequestHandler):
 
 class PrometheusHandler(RequestHandler):
     import prometheus_client # keep out of global namespace
+
     def __init__(self, *args, **kwargs):
         super(PrometheusHandler, self).__init__(*args, **kwargs)
         self.workers = self.prometheus_client.Gauge('workers_total',
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -164,9 +164,10 @@ class IndividualPlots(RequestHandler):
 
 
 class PrometheusHandler(RequestHandler):
-    import prometheus_client # keep out of global namespace
-
     def __init__(self, *args, **kwargs):
+        import prometheus_client # keep out of global namespace
+        self.prometheus_client = prometheus_client
+
         super(PrometheusHandler, self).__init__(*args, **kwargs)
         self.workers = self.prometheus_client.Gauge('workers_total',
             'Total number of workers.',
--- a/distributed/bokeh/tests/test_scheduler_bokeh_html.py
+++ b/distributed/bokeh/tests/test_scheduler_bokeh_html.py
@@ -67,7 +67,6 @@ def test_prefix(c, s, a, b):
              scheduler_kwargs={'services': {('bokeh', 0):  BokehScheduler}})
 def test_prometheus(c, s, a, b):
     http_client = AsyncHTTPClient()
-    for suffix in ['metrics']:
-        response = yield http_client.fetch('http://localhost:%d/%s'
-                                           % (s.services['bokeh'].port, suffix))
-        assert response.code == 200
+    response = yield http_client.fetch('http://localhost:%d/metrics'
+                                       % s.services['bokeh'].port)
+    assert response.code == 200
--- a/distributed/bokeh/worker_html.py
+++ b/distributed/bokeh/worker_html.py
@@ -15,9 +15,10 @@ class RequestHandler(web.RequestHandler):
 
 
 class PrometheusHandler(RequestHandler):
-    import prometheus_client # keep out of global namespace
-
     def __init__(self, *args, **kwargs):
+        import prometheus_client # keep out of global namespace
+        self.prometheus_client = prometheus_client
+
         super(PrometheusHandler, self).__init__(*args, **kwargs)
         # Add metrics like this:
         # self.workers = self.prometheus_client.Gauge('memory_bytes',
--- a/distributed/bokeh/worker_html.py
+++ b/distributed/bokeh/worker_html.py
@@ -16,6 +16,7 @@ class RequestHandler(web.RequestHandler):
 
 class PrometheusHandler(RequestHandler):
     import prometheus_client # keep out of global namespace
+
     def __init__(self, *args, **kwargs):
         super(PrometheusHandler, self).__init__(*args, **kwargs)
         # Add metrics like this:
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -871,7 +871,7 @@ def gen_cluster(ncores=[('127.0.0.1', 1), ('127.0.0.1', 2)],
                                         Worker=Worker, scheduler_kwargs=scheduler_kwargs,
                                         worker_kwargs=worker_kwargs)
                                 except Exception as e:
-                                    logger.error("Failed to start gen_cluster, retryng", exc_info=True)
+                                    logger.error("Failed to start gen_cluster, retrying", exc_info=True)
                                 else:
                                     workers[:] = ws
                                     args = [s] + workers
--- a/distributed/bokeh/components.py
+++ b/distributed/bokeh/components.py
@@ -97,77 +97,77 @@ class TaskStream(DashboardComponent):
 
 
 def task_stream_figure(clear_interval='20s', **kwargs):
-        """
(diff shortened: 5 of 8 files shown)


distributed/bokeh/scheduler_html.py
```diff
@@ -5,6 +5,7 @@
 from tornado import escape
 from tornado import gen
 from tornado import web
+from prometheus_client import Gauge, generate_latest
```
mrocklin: We'll want to put this import within the prometheus handler I think so that the library doesn't strictly depend on it (we try to keep dependencies lightweight).

distributed/bokeh/scheduler_html.py
```diff
@@ -163,6 +164,23 @@ def get(self):
         self.write(result)
 
 
+class PrometheusHandler(RequestHandler):
+    # Construct Prometheus metrics
+    # https://prometheus.io/docs/introduction/overview/
```
mrocklin: If you'd like to provide a documentation string I recommend reading through http://docs.dask.org/en/latest/develop.html#docstrings

distributed/bokeh/scheduler_html.py
```diff
@@ -163,6 +164,23 @@ def get(self):
...
+    def get(self):
+        workers = Gauge('workers_total',
```
mrocklin: Because prometheus functions are only used a couple times in the codebase I'd prefer that we keep the explicit namespace and use the full `prometheus_client.Gauge` name.  That way future maintainers will have more context about what a Gauge object is without having to dereference and find the import.

mrocklin:
@jacobtomlinson @dsludwig you both may find this interesting

mrocklin:
The flake8 failures have been happening everywhere.  They updated recently and became slightly more sensitive.  Handling in 

The `test_gc_diagnosis_cpu_time` failure is new.   I'm taking a look.  I agree that it looks unrelated.

mrocklin:
... Handling in https://github.com/dask/distributed/pull/2500

mrocklin:
Adding the worker was faster than expected :)

distributed/bokeh/scheduler_html.py
beberg: makes sense, fixed scope and made explicit.

mrocklin:
In particular @jacobtomlinson @dsludwig @yuvipanda it would be good to get your sense on the following:

1.  Is this generally the right approach for Dask workers and scheduler to publish metrics?  Are there alternatives that we should consider?
2.  Operationally, what metrics would be valuable to you?

beberg:
Both scheduler and worker now have Prometheus endpoints. They return the built-in python metrics, and the scheduler has client and worker gauges. More metrics can be added in additional PRs, but we should not add any metrics we do not anticipate to be permanent, so we should review those 2 before merging.

yuvipanda:
Prometheus is the way to go! 

In the long run, I like defining metrics like https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/metrics.py

This lets you have a shared location that defines what the metrics are and they can then be incredmented / decremented in other places. It doesn't matter right now for this PR, since you're only exposing guages you already have. But as soon as you wanna use anything else, you're gonna have to define metrics in a central location. It might make sense to do so now than wait for another PR.

beberg:
Agreed @yuvipanda, @mrocklin wanted to keep this one as a surgical strike to get minimal Prometheus support in, and do a larger refactoring of a common metrics core feeding Bokeh & Prometheus later.

mrocklin:
Cc @quasiben this may interest you

mrocklin:
@yuvipanda are there metrics that you have found particularly valuable for JupyterHub?  Perhaps that experience can help us prioritize a bit.

@jacobtomlinson what do you all end up using within UK-Met?

distributed/bokeh/scheduler_html.py
```diff
@@ -163,6 +163,22 @@ def get(self):
         self.write(result)
 
 
+class PrometheusHandler(RequestHandler):
+    def get(self):
+        import prometheus_client
+        workers = prometheus_client.Gauge('workers_total',
```
jacobtomlinson: I'm not sure about defining this in here. Same for the other one below. Creating this object automatically registers it with the prometheus client. I could see this causing conflicts when accessing multiple times.

distributed/bokeh/worker_html.py
```diff
@@ -0,0 +1,34 @@
...
+        self.extra = extra or {}
+
+    def get_template_path(self):
+        return os.path.join(dirname, 'templates')
+
+
+class PrometheusHandler(RequestHandler):
+    def get(self):
+        import prometheus_client
+        #workers = prometheus_client.Gauge('memory_bytes',
```
jacobtomlinson: Is this comment to illustrate what could be added? If so it could be good to explicitly state that in the comment, however I'm not that keen on this approach.

mrocklin:
I'm somewhat sensitive about introducing prometeus metrics throughout the codebase and so like that they are optionally introduced in one spot, rather than managed as we modify state throughout the code.

This is because most contexts in which Dask is used don't have Prometheus active, so I'm sensitive both about adding a dependency, sprinkling in a new library throughout the codebase, and the cost of maintaining state generally (the scheduler is often running at 100% cpu usage). 

Happy for pushback.

jacobtomlinson:
> [I] ... like that they are optionally introduced in one spot, rather than managed as we modify state throughout the code.

This sounds reasonable to me. It will limit some of the things we can do with prometheus (you can time function calls for example using a decorator) but there are always performance overheads when adding instrumentation and wouldn't want to negatively impact distributed.

My comments were more about constructing the `Gauge` objects inside the tornado handler as it will result in duplicate metrics in the prometheus registry. They should be constructed once (perhaps as a constant in the same file) and then used within the handler.

I think basically all the metrics I suggested would be suitable as a gauge and could be updated on request from the handler. We could also use `Summary` metrics for things like version numbers and other metadata in the same way.

mrocklin:
OK, that makes sense to me.  So the proposal is to move the allocation of the Gauge objects to a separate file, but only update them in the request handler?

distributed/bokeh/scheduler_html.py
beberg: You're right about this, will move to class init.

distributed/bokeh/worker_html.py
beberg: Will move and comment.

distributed/bokeh/scheduler_html.py
```diff
@@ -163,6 +163,25 @@ def get(self):
         self.write(result)
 
 
+class PrometheusHandler(RequestHandler):
+    import prometheus_client # keep out of global namespace
```
mrocklin: This will still get run at import time.    This is a problem for two reasons

1.  It adds about 11ms to our import time.   This is small, but these things add up somewhat quickly
2.  The full dashboard code will err if the user doesn't have prometheus_client installed

You'll probably want this in the `__init__` function

distributed/bokeh/tests/test_scheduler_bokeh_html.py
```diff
@@ -60,3 +60,14 @@ def test_prefix(c, s, a, b):
...
         else:
             assert xml.etree.ElementTree.fromstring(body) is not None
+
+
+@gen_cluster(client=True,
+             check_new_threads=False,
+             scheduler_kwargs={'services': {('bokeh', 0):  BokehScheduler}})
+def test_prometheus(c, s, a, b):
+    http_client = AsyncHTTPClient()
+    for suffix in ['metrics']:
```
mrocklin: We can probably collapse the for loop here

distributed/bokeh/worker.py
```diff
@@ -678,6 +678,19 @@ def __init__(self, worker, io_loop=None, prefix='', **kwargs):
...
         return self.worker
+
+    def listen(self, *args, **kwargs):
+        super(BokehWorker, self).listen(*args, **kwargs)
+
+        from .worker_html import routes
+        handlers = [(self.prefix + '/' + url, cls, {'server': self.my_server, 'extra': self.extra})
+                    for url, cls in routes]
+
+        self.server._tornado.add_handlers(r'.*', handlers)
```
mrocklin: Thanks for doing this.  It's nice to have the infrastructure setup on the worker to add new routes in the future

distributed/utils_test.py
```diff
@@ -871,7 +871,7 @@ def coro():
                                         Worker=Worker, scheduler_kwargs=scheduler_kwargs,
                                         worker_kwargs=worker_kwargs)
                                 except Exception as e:
-                                    logger.error("Failed to start gen_cluster, retryng", exc_info=True)
+                                    logger.error("Failed to start gen_cluster, retrying", exc_info=True)
```
mrocklin: Thanks for cleaning this up.

mrocklin:
Merging this in.  Thanks @beberg !

PR #2510, opened 2019-02-06
Avoid thread testing for TCP ThreadPoolExecutor

mrocklin:
This includes two changes that were affecting CI

1.  We no longer check for new threads that are coming from the TCP Executor used by Tornado
2.  We increase the size of the pause_executor test, which seems to help a bit


mrocklin:
Things are looking better here.  Merging on passed tests.

mrocklin:
I've run travis-ci twice with no failures.  Restarting again...

mrocklin:
`distributed/diagnostics/tests/test_progress.py::test_many_progress` failed.  Reported here: https://github.com/dask/distributed/issues/2512

PR #2526, opened 2019-02-12
Support alternative --remote-dask-worker SSHCluster() and dask-ssh CLI

beberg:
Tested with distributed.cli.dask_worker, dask_cuda.dask_cuda_worker, and an invalid worker.

Fixes https://github.com/rapidsai/dask-cuda/issues/2

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -214,9 +214,10 @@ def start_worker(logdir, scheduler_addr, scheduler_port, worker_addr, nthreads, 
                  memory_limit,
                  worker_port,
                  nanny_port,
-                 remote_python=None):
+                 remote_python=None,
+                 remote_dask_worker=None):
 
-    cmd = ('{python} -m distributed.cli.dask_worker '
+    cmd = ('{python} -m {remote_dask_worker} '
            '{scheduler_addr}:{scheduler_port} '
            '--nthreads {nthreads} --nprocs {nprocs} ')
 
@@ -234,6 +235,7 @@ def start_worker(logdir, scheduler_addr, scheduler_port, worker_addr, nthreads, 
 
     cmd = cmd.format(
         python=remote_python or sys.executable,
+        remote_dask_worker=remote_dask_worker or 'distributed.cli.dask_worker',
         scheduler_addr=scheduler_addr,
         scheduler_port=scheduler_port,
         worker_addr=worker_addr,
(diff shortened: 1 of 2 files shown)


distributed/deploy/ssh.py
```diff
@@ -214,9 +214,10 @@ def start_worker(logdir, scheduler_addr, scheduler_port, worker_addr, nthreads,
                  memory_limit,
                  worker_port,
                  nanny_port,
-                 remote_python=None):
+                 remote_python=None,
+                 remote_dask_worker=None):
```
mrocklin:
```suggestion
                 remote_dask_worker='distributed.cli.dask_worker'):
```

distributed/deploy/ssh.py
```diff
@@ -234,6 +235,7 @@ def start_worker(logdir, scheduler_addr, scheduler_port, worker_addr, nthreads,
 
     cmd = cmd.format(
         python=remote_python or sys.executable,
+        remote_dask_worker=remote_dask_worker or 'distributed.cli.dask_worker',
```
mrocklin:
```suggestion
        remote_dask_worker=remote_dask_worker,
```

distributed/deploy/ssh.py
```diff
@@ -273,7 +275,8 @@ class SSHCluster(object):
     def __init__(self, scheduler_addr, scheduler_port, worker_addrs, nthreads=0, nprocs=1,
                  ssh_username=None, ssh_port=22, ssh_private_key=None,
                  nohost=False, logdir=None, remote_python=None,
-                 memory_limit=None, worker_port=None, nanny_port=None):
+                 memory_limit=None, worker_port=None, nanny_port=None,
+                 remote_dask_worker=None):
```
mrocklin:
```suggestion
                 remote_dask_worker='distributed.cli.dask_worker'):
```

beberg: I modeled remote_dask_worker off remote_python, where the innermost layer sets the default. Is there a reason to do remote_dask_worker differently?

mrocklin: When we provide keyword arguments these go into docstrings (which are useful in IDEs and interactive sessions) and into documentation.  

People tend to avoid placing system-specific values into keywords so that they don't get written into docs. 

So for `sys.executable` we avoid it because then we'll have something like `/home/beberg/miniconda/bin/python` in the docs (or whatever the readthedocs build service uses).  However for the constant value of `distributed.cli.dask_worker` it probably does make sense to use.

mrocklin:
Thanks @beberg !

naji-s:
Follow up on this, I am really not following how one can use this to create dask-cuda workers on SSH; seemingly this thread closes @mrocklin's feature request in https://github.com/rapidsai/dask-cuda/issues/2

PR #2541, opened 2019-02-24
Fix LocalCluster to not overallocate memory when overcommitting threads per worker

### LINKED ISSUE #2536: LocalCluster may overallocate memory
On a VM with 16GB of total memory:
- `Client(n_workers=1)` allocates 16GB for a single spawned worker
- `Client(n_workers=2)` allocates 8GB for each of the two spawned workers
- `Client(n_workers=4)` allocates 4GB for each of the four spawned workers
So far so good but
- `Client(n_workers=3)` allocates 8GB for each of the three spawned workers

Even worse, in the latter case there's no `MemoryError` or `KilledWorker` when the workers do end up needing more than the total memory. Instead the whole VM becomes unresponsive and has to be rebooted (couldn't even ssh to it).

gsakkis:
Fixes #2536

### DIFF
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -328,15 +328,17 @@ def test_remote_access(loop):
         sync(loop, assert_can_connect_from_everywhere_4_6, c.scheduler.port)
 
 
-def test_memory(loop):
-    with LocalCluster(scheduler_port=0, processes=False, silence_logs=False,
-                      diagnostics_port=None, loop=loop) as cluster:
+@pytest.mark.parametrize('n_workers', [None, 3])
+def test_memory(loop, n_workers):
+    with LocalCluster(n_workers=n_workers, scheduler_port=0, processes=False,
+                      silence_logs=False, diagnostics_port=None, loop=loop) as cluster:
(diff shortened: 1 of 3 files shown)


distributed/deploy/tests/test_local.py
```diff
@@ -328,15 +328,17 @@ def test_remote_access(loop):
...
+    with LocalCluster(n_workers=n_workers, scheduler_port=0, processes=False,
+                      silence_logs=False, diagnostics_port=None, loop=loop) as cluster:
         assert sum(w.memory_limit for w in cluster.workers) <= TOTAL_MEMORY
 
 
-def test_memory_nanny(loop):
-    with LocalCluster(scheduler_port=0, processes=True, silence_logs=False,
-                      diagnostics_port=None, loop=loop) as cluster:
+@pytest.mark.parametrize('n_workers', [None, 1, 2, 3, 4])
+def test_memory_nanny(loop, n_workers):
```
mrocklin: This test takes about a second to run, so adding the parameterization above adds something like 4s extra to the test suite.  If we can cut this down, perhaps by using two entries rather than five that would be helpful to keep tests fast.

The test above is faster because it doesn't create new processes.

gsakkis: Done, thanks!

PR #2544, opened 2019-02-27
fix double init of prometheus metrics

crepererum:
Tornado handlers might be initaliazed multiple times within the same
python interpreter process. Adding new prometheus gauges/counters
every time leads to `ValueError: Duplicated timeseries` errors because
`prometheus_client` has one single, global registry to track all known
metrics. Instead, use a class-level singelton instead.

### DIFF
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -170,13 +170,13 @@ class _PrometheusCollector(object):
 
     def collect(self):
         yield self.prometheus_client.core.GaugeMetricFamily(
-            'scheduler_workers_total',
-            'Total number of workers.',
+            'dask_scheduler_workers',
+            'Number of workers.',
             value=len(self.server.workers),
         )
         yield self.prometheus_client.core.GaugeMetricFamily(
-            'scheduler_clients_total',
-            'Total number of clients.',
+            'dask_scheduler_clients',
+            'Number of clients.',
             value=len(self.server.clients),
         )
 
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -163,23 +163,49 @@ class IndividualPlots(RequestHandler):
         self.write(result)
 
 
+class _PrometheusCollector(object):
+    def __init__(self, server, prometheus_client):
+        self.server = server
+        self.prometheus_client = prometheus_client
+
+    def collect(self):
+        yield self.prometheus_client.core.GaugeMetricFamily(
+            'scheduler_workers_total',
+            'Total number of workers.',
+            value=len(self.server.workers),
+        )
+        yield self.prometheus_client.core.GaugeMetricFamily(
+            'scheduler_clients_total',
+            'Total number of clients.',
+            value=len(self.server.clients),
+        )
+
+
 class PrometheusHandler(RequestHandler):
+    _initialized = False
+
     def __init__(self, *args, **kwargs):
-        import prometheus_client # keep out of global namespace
+        import prometheus_client  # keep out of global namespace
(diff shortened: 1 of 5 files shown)


mrocklin:
cc @beberg

beberg:
@crepererum Why is Tornado being initialized multiple times? Is the Prometheus issue the symptom of that underlying problem?

crepererum:
Good question. I'll try to figure that out ⏳

StephanErb:
Tornado creates a new RequestHandler object for each request (https://tornado.readthedocs.io/en/latest/guide/structure.html#overriding-requesthandler-methods). 

During the initialisation of the `PrometheusHandler` we thus re-add the Prometheus metrics in the (global) metrics registry. This is not allowed. The canonical way to solve this issue is by implementing a collector https://github.com/prometheus/client_python#custom-collectors

crepererum:
OK. So here is my plan for this PR: I'll migrate the code to a collector and try to add a test. Might take some days though...

crepererum:
Done. The newly added tests even test the mentioned problem (at least for the scheduler where it occurred). What I do not understand are the test failures though. Seems that workers sometimes do not start correctly. Is that flaky?

crepererum:
Done. The newly added tests even test the mentioned problem (at least for the scheduler where it occurred). What I do not understand are the test failures though. Seems that workers sometimes do not start correctly. Is that flaky?

mrocklin:
Thanks @crepererum .  Yes, we have a few flaky tests, particularly in Python 2.  I'll restart them and we'll see how they go.  

@StephanErb thank you for the suggestion earlier of using collectors.  Do you have time to give this a quick review?  That would be much appreciated if so.

distributed/bokeh/scheduler_html.py
```diff
@@ -163,23 +163,49 @@ def get(self):
...
 
 
+class _PrometheusCollector(object):
+    def __init__(self, server, prometheus_client):
+        self.server = server
+        self.prometheus_client = prometheus_client
+
+    def collect(self):
+        yield self.prometheus_client.core.GaugeMetricFamily(
+            'scheduler_workers_total',
```
StephanErb: The metric names in the PR do not follow the Prometheus naming conventions (https://prometheus.io/docs/practices/naming/):

* We need a common prefix for dask / distributed e.g. `dask_distributed_`
* The suffix `_total` must only be used for counters (i.e. only increasing). This is a gauge so we cannot not use the prefix.

distributed/bokeh/tests/test_scheduler_bokeh_html.py
```diff
@@ -68,6 +68,10 @@ def test_prefix(c, s, a, b):
...
-    response = yield http_client.fetch('http://localhost:%d/metrics'
-                                       % s.services['bokeh'].port)
-    assert response.code == 200
+
+    # request data twice since there once was a case where metrics got registered multiple times resulting in
+    # prometheus_client errors
+    for _ in range(2):
+        response = yield http_client.fetch('http://localhost:%d/metrics'
+                                           % s.services['bokeh'].port)
+        assert response.code == 200
```
StephanErb: As a smoke test, I would recommend to also check for the existence of at least one metric (e.g. anything starting with `dask_distributed`).

distributed/bokeh/scheduler_html.py
mrocklin: I'm going to suggest the name `dask` rather than `dask_distributed`.  I think that Prometheus will only be used with Dask's distributed scheduler.  I don't think that there is a need to add `_distributed` to the namespace.

mrocklin:
Great.  Thanks @crepererum for the fix and @StephanErb for the review!

Merging.

PR #2551, opened 2019-03-07
TST: Fetch executables from build root.

QuLogic:
When building a package, nothing is installed in the prefix, but to a build root. This means any tests that use the `popen` wrapper cannot find the executables (or find old ones if distributed is installed already).

This looks in the build root instead, if the `DESTDIR` environment variable is set.

### DIFF
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -992,7 +992,7 @@ def popen(args, **kwargs):
     if sys.platform.startswith('win'):
         args[0] = os.path.join(sys.prefix, 'Scripts', args[0])
     else:
-        args[0] = os.path.join(sys.prefix, 'bin', args[0])
+        args[0] = os.path.join(os.environ.get('DESTDIR', '') + sys.prefix, 'bin', args[0])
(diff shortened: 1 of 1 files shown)


mrocklin:
Will the `python` executable will be installed in a different location than Python CLI utilities?

Also, please forgive my ignorance here, but is `DESTDIR` a common standard?

QuLogic:
The final location will be the same, but during package building, the _new_ package gets installed into an isolated directory. The package is then created from this directory. Tests will not of course be able to find the executables in the final location at this time.

The `DESTDIR` variable is a [GNU convention for staged installs](https://www.gnu.org/prep/standards/html_node/DESTDIR.html) (also understood by, e.g., [CMake](https://cmake.org/cmake/help/latest/envvar/DESTDIR.html)). However, since it's really for the install step and not tests, per se, if you want to use a different name, that's fine too.

mrocklin:
OK, I'm going to claim ignorance here and trust you :)

Merging in 24 hours.

cc @jakirkham in case you have time to look things over (and please feel free to merge)

PR #2553, opened 2019-03-07
TST: Skip test_duplicate_clients without bokeh.

QuLogic:
Just like the other tests in this file that use bokeh.


mrocklin:
Thanks @QuLogic !

PR #2556, opened 2019-03-07
Handler blocking

### LINKED ISSUE #2550: Config setting to disallow `run_on_scheduler`
We run an adaptive deployment wherein the dask `Scheduler` runs in an environment with elevated permissions compared with the workers.  Consequently, we'd rather not have an easily accessible method for accidentally executing code sent from a worker on the scheduler.

I know that this functionality was introduced mainly for debugging purposes in https://github.com/dask/distributed/pull/808, so was hoping we could introduce a config setting that turned this feature off (probably by just implementing a simple `if / else` in https://github.com/dask/distributed/blob/master/distributed/scheduler.py#L3069).

I'm happy to make the PR but wanted to open this issue for discussion before doing so.

cicdw:
Closes #2550 by surfacing a `blocked_handlers` kwarg on a few different classes, which serves the purpose of ultimately preventing certain routes on the Scheduler from being called by workers.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -102,7 +102,9 @@ class Server(object):
             'connection_stream': self.handle_stream,
         }
         self.handlers.update(handlers)
-        self.blocked_handlers = blocked_handlers or []
+        if blocked_handlers is None:
+            blocked_handlers = dask.config.get('distributed.%s.blocked-handlers' % type(self).lower(), [])
+        self.blocked_handlers = blocked_handlers
         self.stream_handlers = {}
         self.stream_handlers.update(stream_handlers or {})
 
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -43,6 +43,12 @@ def get_total_physical_memory():
         return 2e9
 
 
+def raise_later(exc):
+    def _raise(*args, **kwargs):
+        raise exc
+    return _raise
+
+
 MAX_BUFFER_SIZE = get_total_physical_memory()
 
 tick_maximum_delay = parse_timedelta(dask.config.get('distributed.admin.tick.limit'), default='ms')
@@ -313,9 +319,6 @@ class Server(object):
 
                 try:
                     op = msg.pop('op')
-                    if op in self.blocked_handlers:
-                        msg = "The '{op}' handler has been explicitly disallowed in {obj}, possibly due to security concerns."
-                        raise ValueError(msg.format(op=op, obj=type(self).__name__))
                 except KeyError:
                     raise ValueError(
                         "Received unexpected message without 'op' key: " %
@@ -334,7 +337,12 @@ class Server(object):
 
                 result = None
                 try:
-                    handler = self.handlers[op]
+                    if op in self.blocked_handlers:
+                        _msg = "The '{op}' handler has been explicitly disallowed in {obj}, possibly due to security concerns."
+                        exc = ValueError(_msg.format(op=op, obj=type(self).__name__))
+                        handler = raise_later(exc)
+                    else:
+                        handler = self.handlers[op]
                 except KeyError:
                     logger.warning("No handler %s found in %s", op,
                                    type(self).__name__, exc_info=True)
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -103,7 +103,7 @@ class Server(object):
         }
         self.handlers.update(handlers)
         if blocked_handlers is None:
-            blocked_handlers = dask.config.get('distributed.%s.blocked-handlers' % type(self).lower(), [])
+            blocked_handlers = dask.config.get('distributed.%s.blocked-handlers' % type(self).__name__.lower(), [])
         self.blocked_handlers = blocked_handlers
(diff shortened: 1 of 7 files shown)


distributed/core.py
```diff
@@ -312,6 +313,9 @@ def handle_comm(self, comm, shutting_down=shutting_down):
 
                 try:
                     op = msg.pop('op')
+                    if op in self.blocked_handlers:
+                        msg = "The '{op}' handler has been explicitly disallowed in {obj}, possibly due to security concerns."
+                        raise ValueError(msg.format(op=op, obj=type(self).__name__))
```
mrocklin: I see that you've copied my suggestion here, foolishly assuming that I knew what I was talking about.  This will probably stop the server entirely.  We probably need to handle the error similar to how we handle errors in handlers below (look for when `handler` gets called around ten lines lower down).

One simple way to do this would be to return a handler that just errs and then reuse the stanard code below.  Dask should then route the error to the caller appropriately.

distributed/deploy/local.py
```diff
@@ -54,6 +54,9 @@ class LocalCluster(Cluster):
         Tornado gen.coroutines.  This should remain False for normal use.
     kwargs: dict
         Extra worker arguments, will be passed to the Worker constructor.
+    blocked_handlers: List[str]
+        A list of strings specifying a blacklist of handlers to disallow on the Scheduler,
+        e.g., ``['feed', 'run_function']``
```
mrocklin: Copy-editors sometimes scold me for using `e.g.` rather than `like`, which they say is more easily understood by non-native speakers.

distributed/deploy/tests/test_local.py
```diff
@@ -41,6 +42,13 @@ def test_simple(loop):
             assert e.loop is c.loop
 
 
+def test_local_cluster_supports_blocked_handlers(loop):
+    with LocalCluster(blocked_handlers=['run_function'], loop=loop) as c:
+        with Client(c) as client:
+            with pytest.raises(CommClosedError):
+                client.run_on_scheduler(lambda x: x, 42)
```
mrocklin: Yeah, so we shouldn't have a closed comm, we should get the `ValueError`.  This is related to the comment above about handling errors in core.py

distributed/core.py
cicdw: Oh great!  Will update.

cicdw: Should I also update the "unexpected message without 'op'" `ValueError` that is raised right below mine?

mrocklin: Lets leave that as is for now.  If we find that it's a concern then we'll handle it in another PR.

cicdw: With this change, the behavior is in line with what I was originally expecting -- thank you!

distributed/deploy/local.py
cicdw: updated

distributed/deploy/tests/test_local.py
cicdw: updated

cicdw:
Test failures appear to be unrelated to this PR?

mrocklin:
Maybe.  I haven't seen that one before, and its consistent across travis-ci and appveyor.  I'll take a look later today.

distributed/core.py
```diff
@@ -89,13 +95,14 @@ class Server(object):
...
 
-    def __init__(self, handlers, stream_handlers=None, connection_limit=512,
+    def __init__(self, handlers, blocked_handlers=None, stream_handlers=None, connection_limit=512,
                  deserialize=True, io_loop=None):
         self.handlers = {
             'identity': self.identity,
             'connection_stream': self.handle_stream,
         }
         self.handlers.update(handlers)
+        self.blocked_handlers = blocked_handlers or []
```
mrocklin:
```suggestion
        if blocked_handlers is None:
            blocked_handlers = dask.config.get('distributed.%s.blocked-handlers' % type(self).lower(), [])
        self.blocked_handlers = blocked_handlers
```

And then can you add `blocked-handlers` entries to `distributed/distributed.yaml` under the `distributed.worker` and `distributed.scheduler` sections?  (My hope is that editing this file will be self explanatory if you haven't done it before)

cicdw: Nice, I like that - allows for changing this behavior entirely from config, which was an initial desire.  Should I add a test for this init behavior (pulling from config)?

mrocklin: Ping on the following:

> And then can you add blocked-handlers entries to distributed/distributed.yaml under the distributed.worker and distributed.scheduler sections? (My hope is that editing this file will be self explanatory if you haven't done it before)

Relevant file is here: https://github.com/dask/distributed/blob/master/distributed/distributed.yaml

A test would be welcome, but is not obligatory

cicdw: Yup sorry for the delay; updated config + a test for the scheduler.

mrocklin:
This looks good to me.  Testing failurres are unrelated. 

Merging in.  Thanks @cicdw !

PR #2559, opened 2019-03-11
BUG: yield for coroutine handlers

TomAugspurger:
Without this change, I see the following when running [this](https://github.com/TomAugspurger/dask-perf/blob/master/bench_comm.py)
comm benchmark.

```
(ucx-dev) an.taugspurger@dgx05:~/ucx-dev-env/dask-perf$ python bench_comm.py
/home/nfs/an.taugspurger/ucx-dev-env/distributed/distributed/comm/tcp.py:436: RuntimeWarning: coroutine 'server_handle_comm' was never awaited
  self.comm_handler(comm)
```

In Antoine's original benchmark, tornado was used throughout (for the handlers). Perhaps tornado does something special in that case. But I believe that to handle `async def` functions, we'll need something like this patch.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -21,7 +21,7 @@ from tornado.tcpserver import TCPServer
 from ..compatibility import finalize, PY3
 from ..threadpoolexecutor import ThreadPoolExecutor
 from ..utils import (ensure_bytes, ensure_ip, get_ip, get_ipv6, nbytes,
-                     parse_timedelta, shutting_down, iscoroutinefunction)
+                     parse_timedelta, shutting_down)
 
 from .registry import Backend, backends
 from .addressing import parse_host_port, unparse_host_port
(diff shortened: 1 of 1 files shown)


distributed/comm/tcp.py
```diff
@@ -433,7 +433,9 @@ def _handle_stream(self, stream, address):
                      address, self.contact_address)
         local_address = self.prefix + get_stream_address(stream)
         comm = self.comm_class(stream, local_address, address, self.deserialize)
-        self.comm_handler(comm)
```
mrocklin: Can we always `yield` here?

mrocklin:
> Perhaps tornado does something special in that case. But I believe that to handle async def functions, we'll need something like this patch.

This is probably the biggest difference in semantics between tornado and asyncio.  In tornado calling a coroutine starts it off and gives you a future pointing to something that is actively running on the event loop.  In Python generally calling an `async def` function gives you a thing that must be awaited on before it starts up.

mrocklin:
In general this change seems fine to me.  One small comment.

distributed/comm/tcp.py
TomAugspurger: Yes, probably... IIUC, for a regular function, `yield self.comm_handler(comm)` would just yield the concrete return value, which should be harmless.

PR #2561, opened 2019-03-12
Add warning if creating scratch directories takes a long time

mrocklin:
This adds a context manager `warn_on_duration` which warns if an action
takes a surprisingly long time.

We then use the context manager when creating a worker's local
directory.

Fixes https://github.com/rapidsai/dask-cuda/issues/11
See also https://github.com/dask/dask-jobqueue/issues/193

### DIFF
--- a/distributed/tests/test_utils.py
+++ b/distributed/tests/test_utils.py
@@ -5,6 +5,7 @@ from functools import partial
 import io
 import socket
 import sys
+from time import sleep
 import traceback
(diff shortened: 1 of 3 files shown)


lkugler:
Hi, so one can try to use local_dir argument to LocalCluster(local_dir=...) with a directory directly on the machines harddrive?

mrocklin:
Yes

PR #2563, opened 2019-03-13
Ensure that default `remote_dask_worker` is used

willirath:
Currently, the cli will pass `remote_dask_worker=None` to the `SSHCluster` and hence override the default kwarg resulting in `dask-ssh` without the `--remote-dask-worker` flag being broken.

- [x] Add test for `distributed.cli.dask_ssh`?
- [ ] Anything else?

### DIFF
--- a/distributed/cli/dask_ssh.py
+++ b/distributed/cli/dask_ssh.py
@@ -47,13 +47,12 @@ from distributed.cli.utils import check_python_3
               help="Serving computation port, defaults to random")
 @click.option('--nanny-port', type=int, default=0,
               help="Serving nanny port, defaults to random")
-@click.option('--remote-dask-worker', default=None, type=str,
+@click.option('--remote-dask-worker', type=str,
+              default="distributed.cli.dask_worker",
               help="Worker to run. Defaults to distributed.cli.dask_worker")
 @click.pass_context
-def main(ctx, scheduler, scheduler_port, hostnames, hostfile, log_directory, **kwargs):
-
-    if kwargs.get("remote_dask_worker") is None:
-        kwargs.pop("remote_dask_worker")
+def main(ctx, scheduler, scheduler_port, hostnames, hostfile, log_directory,
+         **kwargs):
 
     try:
         hostnames = list(hostnames)
@@ -69,7 +68,8 @@ def main(ctx, scheduler, scheduler_port, hostnames, hostfile, log_directory, **k
         print(ctx.get_help())
         exit(1)
 
-    c = SSHCluster(scheduler, scheduler_port, hostnames, logdir=log_directory, **kwargs)
+    c = SSHCluster(scheduler, scheduler_port, hostnames, logdir=log_directory,
+                   **kwargs)
 
     import distributed
     print('\n---------------------------------------------------------------')
--- a/distributed/cli/dask_ssh.py
+++ b/distributed/cli/dask_ssh.py
@@ -52,6 +52,9 @@ from distributed.cli.utils import check_python_3
 @click.pass_context
 def main(ctx, scheduler, scheduler_port, hostnames, hostfile, log_directory, **kwargs):
 
+    if kwargs.get("remote_dask_worker") is None:
+        kwargs.pop("remote_dask_worker")
+
     try:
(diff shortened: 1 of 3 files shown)


distributed/deploy/ssh.py
```diff
@@ -295,7 +295,8 @@ def __init__(self, scheduler_addr, scheduler_port, worker_addrs, nthreads=0, npr
         self.memory_limit = memory_limit
         self.worker_port = worker_port
         self.nanny_port = nanny_port
-        self.remote_dask_worker = remote_dask_worker
+        if remote_dask_worker is None:
+            self.remote_dask_worker = 'distributed.cli.dask_worker'
```
mrocklin: When is it `None`?  Why wouldn't the default value be used?

Also, this implementation is a little bit buggy, you don't do anything with the value if it is not None.

willirath: None is the default coming from the cli script.

willirath: But perhaps this should be fixed in <https://github.com/dask/distributed/blob/master/distributed/cli/dask_ssh.py#L50>?

mrocklin: Yes, I think I'd prefer that we handle this there if possible.  In general I'd prefer to avoid hard-coding the default values for keywords twice.

willirath: I've reverted these and rewritten `cli.dask_ssh` to handle this using kwargs.

willirath:
The test runs into timeout errors.  I guess this is either because `popen` in <https://github.com/dask/distributed/pull/2563/files#diff-a04ded3c2c7e2929887bedf42a617636R16> never returns or is never terminated, because it sits in <https://github.com/dask/distributed/blob/master/distributed/cli/dask_ssh.py#L83>?

How do I handle that @mrocklin ?

distributed/cli/dask_ssh.py
```diff
@@ -50,9 +50,11 @@
 @click.option('--remote-dask-worker', default=None, type=str,
```
mrocklin: Lets change the default here with `default="distributed.cli.dask_worker"`

distributed/cli/dask_ssh.py
```diff
@@ -50,9 +50,11 @@
...
               help="Worker to run. Defaults to distributed.cli.dask_worker")
 @click.pass_context
-def main(ctx, scheduler, scheduler_port, hostnames, hostfile, nthreads, nprocs,
-         ssh_username, ssh_port, ssh_private_key, nohost, log_directory, remote_python,
-         memory_limit, worker_port, nanny_port, remote_dask_worker):
+def main(ctx, scheduler, scheduler_port, hostnames, hostfile, log_directory, **kwargs):
+
+    if kwargs.get("remote_dask_worker") is None:
+        kwargs.pop("remote_dask_worker")
```
mrocklin: Then we should be able to drop this

distributed/cli/dask_ssh.py
willirath: Done.  (But note that we're then setting the default remote worker in two locations: `cli.dask_ssh.main` and `deploy.SSHCluster.__init__`.

distributed/cli/dask_ssh.py
willirath: Done.

distributed/cli/dask_ssh.py
```diff
@@ -67,9 +69,7 @@ def main(ctx, scheduler, scheduler_port, hostnames, hostfile, nthreads, nprocs,
         print(ctx.get_help())
         exit(1)
 
-    c = SSHCluster(scheduler, scheduler_port, hostnames, nthreads, nprocs,
-                   ssh_username, ssh_port, ssh_private_key, nohost, log_directory, remote_python,
-                   memory_limit, worker_port, nanny_port, remote_dask_worker)
+    c = SSHCluster(scheduler, scheduler_port, hostnames, logdir=log_directory, **kwargs)
```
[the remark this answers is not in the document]
willirath: :)

willirath:
I've merged the current state of master in 0888917 to see if the timeout errors go away.

PR #2570, opened 2019-03-19
Fix Torando 6 test failures

mrocklin:
These were failing in Tornado 6

They were faulty anyway because the IOLoop had been run before being
used a second time.


mrocklin:
OK, it looks like the Tornado 6 release mucked about with our CI system a bit.  I've fixed a few issues that seemed to be pretty consistent and expliicitly pinned Tornado to both 5 and 6 now in different matrix elements.

mrocklin:
There are still some intermittent failures, but this is definitely a win.  Merging.

Issue #2513, opened 2019-02-07
Optionally include lower level stack information in profile plots


mrocklin:
The current statistical profiler can provide a ton of useful information about where user code spends time.  Unfortunately it only looks down through Python code.  Anything lower level is effectively opaque.  For people wanting to profile low level compiled code this can be a limitation.

Last year @sklam integrated libunwind with [stacktrace](https://github.com/numba/stacktrace) into Dask's statistical profiler here: https://github.com/dask/distributed/compare/master...sklam:stacktrace_integration?expand=1

Some work to do:

1.  Fix merge conflicts (probably simple)
2.  Add a test or two?
3.  Make it optional with configuration
4.  See how much performance cost arises from its use
5.  Try it out, report feedback, fix bugs

mrocklin:
@beberg this may interest you

pentschev:
I think items 1, 2 and 3 were properly tackled by #2575. As for bugs (item 5), the only known bug is described in #2608.

Regarding performance, from my tests I saw low-level profiler almost doubling execution time in some cases, which doesn't really surprise me a lot, since a much larger amount of functions can be expected to derive from C calls. However, I don't know if this information belongs somewhere, perhaps a warning in documentation?

Other than that, I think the scope of this issue has been fully tackled, so I think we can close this too.

mrocklin:
This is in.  Thanks @pentschev

PR #2575, opened 2019-03-21
Stacktrace integration

pentschev:
Solves #2513.


pentschev:
@mrocklin is CI failing? The failures look non-deterministic both in CI and locally for me.

mrocklin:
Yes, we have intermittent failures in this repository.  They've been particularly bad recently.  I'm trying to handle them in spare cycles, but in the meantime you should probably ignore them.

pentschev:
@mrocklin alright, thanks.

mrocklin:
Some of the CI failures look genuine though

<details>

```
______________________________ test_profile_plot _______________________________
    def test_func():
        del _global_workers[:]
        _global_clients.clear()
    
        reset_config()
    
        dask.config.set({'distributed.comm.timeouts.connect': '5s'})
        # Restore default logging levels
        # XXX use pytest hooks/fixtures instead?
        for name, level in logging_levels.items():
            logging.getLogger(name).setLevel(level)
    
        result = None
        workers = []
    
        with pristine_loop() as loop:
            with check_active_rpc(loop, active_rpc_timeout):
                @gen.coroutine
                def coro():
                    with dask.config.set(config):
                        s = False
                        for i in range(5):
                            try:
                                s, ws = yield start_cluster(
                                    ncores, scheduler, loop, security=security,
                                    Worker=Worker, scheduler_kwargs=scheduler_kwargs,
                                    worker_kwargs=worker_kwargs)
                            except Exception as e:
                                logger.error("Failed to start gen_cluster, retrying", exc_info=True)
                            else:
                                workers[:] = ws
                                args = [s] + workers
                                break
                        if s is False:
                            raise Exception("Could not start cluster")
                        if client:
                            c = yield Client(s.address, loop=loop, security=security,
                                             asynchronous=True, **client_kwargs)
                            args = [c] + args
                        try:
                            future = func(*args)
                            if timeout:
                                future = gen.with_timeout(timedelta(seconds=timeout),
                                                          future)
                            result = yield future
                            if s.validate:
                                s.validate_state()
                        finally:
                            if client:
                                yield c._close(fast=s.status == 'closed')
                            yield end_cluster(s, workers)
                            yield gen.with_timeout(timedelta(seconds=1),
                                                   cleanup_global_workers())
    
                        try:
                            c = yield default_client()
                        except ValueError:
                            pass
                        else:
                            yield c._close(fast=True)
    
                        raise gen.Return(result)
    
>               result = loop.run_sync(coro, timeout=timeout * 2 if timeout else timeout)
distributed/utils_test.py:909: 
../../../miniconda/envs/test-environment/lib/python3.7/site-packages/tornado/ioloop.py:532: in run_sync
../../../miniconda/envs/test-environment/lib/python3.7/site-packages/tornado/gen.py:742: in run
distributed/utils_test.py:890: in coro
    result = yield future
../../../miniconda/envs/test-environment/lib/python3.7/site-packages/tornado/gen.py:729: in run
    value = future.result()
../../../miniconda/envs/test-environment/lib/python3.7/site-packages/tornado/gen.py:742: in run
c = <Client: not connected>
s = <Scheduler: "tcp://127.0.0.1:42679" processes: 2 cores: 3>
a = <Worker: tcp://127.0.0.1:34976, closed, stored: 0, running: 0/1, ready: 0, comm: 0, waiting: 0>
b = <Worker: tcp://127.0.0.1:46675, closed, stored: 0, running: 0/2, ready: 0, comm: 0, waiting: 0>
    @gen_cluster(client=True, check_new_threads=False)
    def test_profile_plot(c, s, a, b):
        p = ProfilePlot()
        assert len(p.source.data['left']) <= 1
        yield c.map(slowinc, range(10), delay=0.05)
        p.update(a.profile_recent)
>       assert len(p.source.data['left']) > 1
E       assert 0 > 1
E        +  where 0 = len([])
distributed/bokeh/tests/test_components.py:34: AssertionError
----------------------------- Captured stderr call -----------------------------
distributed.scheduler - INFO -   Scheduler at:     tcp://127.0.0.1:42679
distributed.worker - INFO -       Start worker at:      tcp://127.0.0.1:34976
distributed.worker - INFO -          Listening to:      tcp://127.0.0.1:34976
distributed.worker - INFO - Waiting to connect to:      tcp://127.0.0.1:42679
distributed.worker - INFO -               Threads:                          1
distributed.worker - INFO -                Memory:                    7.84 GB
distributed.worker - INFO -       Local Directory: /home/travis/build/dask/distributed/dask-worker-space/worker-aldcrzzd
distributed.worker - INFO -       Start worker at:      tcp://127.0.0.1:46675
distributed.worker - INFO -          Listening to:      tcp://127.0.0.1:46675
distributed.worker - INFO - Waiting to connect to:      tcp://127.0.0.1:42679
distributed.worker - INFO -               Threads:                          2
distributed.worker - INFO -                Memory:                    7.84 GB
distributed.worker - INFO -       Local Directory: /home/travis/build/dask/distributed/dask-worker-space/worker-2d4uxt8t
distributed.scheduler - INFO - Register tcp://127.0.0.1:34976
distributed.scheduler - INFO - Register tcp://127.0.0.1:46675
distributed.scheduler - INFO - Starting worker compute stream, tcp://127.0.0.1:34976
distributed.scheduler - INFO - Starting worker compute stream, tcp://127.0.0.1:46675
distributed.worker - INFO -         Registered to:      tcp://127.0.0.1:42679
distributed.worker - INFO -         Registered to:      tcp://127.0.0.1:42679
distributed.scheduler - INFO - Receive client connection: Client-55f6756c-4bf4-11e9-8ffc-42010a140102
tornado.application - ERROR - Exception in callback <bound method Worker.trigger_profile of <Worker: tcp://127.0.0.1:34976, running, stored: 0, running: 1/1, ready: 4, comm: 0, waiting: 0>>
  File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/tornado/ioloop.py", line 907, in _run
    return self.callback()
  File "/home/travis/build/dask/distributed/distributed/worker.py", line 2232, in trigger_profile
    llframes = {ident: profile.ll_get_stack(ident) for ident in active_threads}
  File "/home/travis/build/dask/distributed/distributed/worker.py", line 2232, in <dictcomp>
    llframes = {ident: profile.ll_get_stack(ident) for ident in active_threads}
  File "/home/travis/build/dask/distributed/distributed/profile.py", line 401, in ll_get_stack
    from stacktrace import get_thread_stack, print_thread_stack
ModuleNotFoundError: No module named 'stacktrace'
```

</details>

You may have to modify the .travis.yml script or `continuous_integration/travis/install.sh`

pentschev:
Yes, some of them must be, but I was confused with some of them, I'll have
to check in detail.

pentschev:
I think we won't be able to use/test this on Windows workers, there seems to be no support for libunwind for it, at least not officially. It seems like there's [Clang support for libunwind on Windows](https://bcain-llvm.readthedocs.io/projects/libunwind/en/latest/). I'm thinking of for now either raising an exception when on Windows. Any thoughts?

mrocklin:
>  I'm thinking of for now either raising an exception when on Windows. Any thoughts?

Ideally our profiler would be robust to`stacktrace` not being available on the user's system.  Presumably a Windows user would just not be able to install `stacktrace` and things would gracefully not use the low level features.

pentschev:
> Presumably a Windows user would just not be able to install `stacktrace` and things would gracefully not use the low level features.

I agree with that, but I think there should at least be a warning when the user enables that feature, to inform that this isn't available, this would avoid someone trying to use such a feature having to figure out on his own that it isn't available.

mrocklin:
Yes, that seems reasonable to me.  It looks like some parts of dask use `dask.utils.import_required`

pentschev:
@mrocklin there were some failures in the last build, but unrelated to this PR.

From my tests, the low-level profiler works fairly well, but to be honest, I think it becomes a bit confusing, with too many calls that aren't really informative, these calls are usually OS or architecture-dependent.

Regarding the test I added, it isn't the best that could be as it check for `__restore_rt` or `_L_unlock_13`. At the moment I can't think of a way to test this that we could be entirely sure of the calls that happen inside, since very low-level calls may vary depending on installed software, like the kernel and libc versions, compiler used, etc. If you have an idea, please, let me know.

mrocklin:
cc @sklam @seibert @kkraus14 @jrhemstad do any of you have suggestions here?

sklam:
In the stacktrace tests, I just check for valid formatting of the output.  It cannot check for exact entries in the stack portability.  (well, unless one write an expensive C function for it to call)

distributed/tests/test_profile.py
```diff
@@ -46,6 +47,43 @@ def test_f():
...
+            if f is not None:
+                llprocess(f, None, state)
+
+    assert state['count'] == 100
+    d = state
+    while len(d['children']) == 1:
+        d = first(d['children'].values())
+
+    assert d['count'] == 100
+    assert '__restore_rt' or '_L_unlock' in str(d['description'])
```
sklam: This assert is always True.  

```
In [5]: 'a' or 0 in []
Out[5]: 'a'
```

it should be `assert '__restore_rt' in str(d['description']) or '_L_unlock' in str(d['description'])`

pentschev:
> unless one write an expensive C function for it to call

I thought about this, but even then you don't know what functions will be in the compiled code. You could, of course, check if your expensive function is called at some point, but you probably can't figure exactly where in the stack this call will be.

distributed/tests/test_profile.py
pentschev: Ah, right, thanks for spotting it!

distributed/tests/test_profile.py
```diff
@@ -46,6 +47,45 @@ def test_f():
     assert 95 < g['count'] + h['count'] <= 100
 
 
+@pytest.mark.skipif(
+        sys.version_info < (3, 5),
+        reason='stacktrace not supported on python < 3.5')
```
mrocklin: Perhaps instead we should have the following line within the function

```python
pytest.importorskip('stacktrace`)
```

That way if stacktrace does support lower versions in the future, or if we don't have stacktrace installed, then things will still work correctly.

distributed/worker.py
```diff
@@ -440,6 +440,8 @@ def __init__(self, scheduler_ip=None, scheduler_port=None,
         self.service_specs = services or {}
         self.metrics = dict(metrics) if metrics else {}
 
+        self.low_level_profiler = low_level_profiler
```
mrocklin: We might want to use configuration here as well.  

```python
dask.config.get('distributed.worker.low-level-profiler')
```

you would also have to add this to the `distributed.yaml` file in this repository.

distributed/tests/test_profile.py
pentschev: That's a much better idea, I'll do that. Thanks for the suggestion.

distributed/worker.py
pentschev: Ah, that sounds good. Should I remove then the parameter from `worker.__init__()` in favor of the configuration?

pentschev:
> Is there an obvious condition to filter on? Do we have a sense for what makes a line interesting?

At the moment, I don't see an obvious condition. I think in favor of reliability, until we can come up with a really good condition, what I'll do is scrape the stack and make sure at least one of these calls happens.

pentschev:
I'll try to get back to fixing the test this week. Unfortunately, merging master is probably not enough. :(

mrocklin:
I just gave this a try.  Very fun!  I did change the coloring a bit so that low-level functions ended up as light-gray to distinguish them a bit.  It looks like this.

Screen Shot 2019-04-08 at 1 18 54 PM

I also tried writing a test for this behavior, but low-level profiling things failed on my machine even without this change (I'm not sure why), which made this somewhat difficult.  I did add an unrelated test around the `plot_data` function in `test_basic`.  

```python
    @pytest.mark.skipif(
            sys.version_info < (3, 5),
            reason='stacktrace not supported on python < 3.5')
    @pytest.mark.skipif(
            WINDOWS,
            reason='no low-level profiler support for Windows available')
    def test_basic_low_level():
        def test_g():
            time.sleep(0.05)

        def test_f():
            for i in range(100):
                test_g()

        thread = threading.Thread(target=test_f)
        thread.daemon = True
        thread.start()

        state = create()

        for i in range(100):
            time.sleep(0.02)
            frame = sys._current_frames()[thread.ident]
            llframes = {thread.ident: ll_get_stack(thread.ident)}
            for f in llframes.values():
                if f is not None:
                    llprocess(f, None, state)

>       assert state['count'] == 100
E       assert 0 == 100
```

I also noticed occasional warnings when running in ipython, but am no long able to recreate them.

distributed/profile.py
```diff
@@ -355,3 +355,50 @@ def plot_figure(data, **kwargs):
...
+
+def llprocess(frames, child, state):
+    if not frames:
+        return
+    frame = frames.pop()
+    if frames:
+        state = llprocess(frames, frame, state)
+
+    addr = hex(frame.addr - frame.offset)
+    ident = ';'.join(map(str, (frame.name, '<c>', addr)))
```
mrocklin: Is "<c>" the right name here?  Are we likely to get information from other compiled languages like Fortran or C++?  If so, what is a good alternative here?  "<compiled>" ?  "<low-level>" ?

distributed/worker.py
```diff
@@ -2229,13 +2231,21 @@ def trigger_profile(self):
...
+        llframes = {}
+        if self.low_level_profiler:
+            llframes = {ident: profile.ll_get_stack(ident) for ident in active_threads}
         for ident, frame in frames.items():
             if frame is not None:
                 key = key_split(active_threads[ident])
-                profile.process(frame, None, self.profile_recent,
-                                stop='distributed/worker.py')
-                profile.process(frame, None, self.profile_keys[key],
+                init = profile.create()
```
mrocklin: This doesn't seem to be used

distributed/worker.py
```diff
@@ -2229,13 +2231,21 @@ def trigger_profile(self):
...
+                llframe = llframes.get(ident)
+
+                state = profile.process(frame, True, self.profile_recent,
                                 stop='distributed/worker.py')
+                state = profile.llprocess(llframe, None, state)
```
mrocklin: We might want to defend the `llframe` lines with a `if self.low_level_profiler` conditional.

distributed/profile.py
pentschev: I don't understand, where do you see ""?

distributed/worker.py
pentschev: You mean the `init` variable only, right? I think we still need the `profile.create()`

distributed/worker.py
pentschev: Yes, I'll do that, thanks for pointing it out!

distributed/worker.py
mrocklin: I think that we use `self.profile_recent` as the state object here instead.  We don't need to keep creating a new state I don't think

distributed/profile.py
mrocklin: Ah, this is because I was using text like `<foo>` which github chose wisely to hide.  I've now wrapped these in backticks above.  Hopefully now my comment from before makes more sense.

pentschev: Thanks for clarifying @mrocklin. Your question is indeed a good one, I can't answer about Fortran, but from C++ symbols we definitely will. I'm fine using `<low-level>`instead, but `<compiled>` feels confusing somehow.

distributed/worker.py
pentschev: I'll double-check that then, you may be right.

distributed/profile.py
mrocklin: `<low-level>` it is then!  (unless someone can think of something better)

distributed/worker.py
pentschev: Removing this, you were right in your comment @mrocklin. Thanks for noticing that.

distributed/worker.py
pentschev: After looking at the code, I remember why I didn't do that. `llframes` is an empty dict when `self.low_level_profiler = False`, therefore, calling `llframes.get(ident)` returns `None`, and `profile.llprocess()` starts with `if not frames: return`. We could still add the conditional here, but it would be redundant. I would normally not do that in code I write, but if you think it's useful for someone reading the code, it's fine by me.

pentschev:
So I changed the test now not for specific functions, but to ensure we have the `<low-level>` as key in the profile dict `children` and that `filename == '<low-level>'`. This does not check specifically for what functions were executed, but rather that the low-level profiler was executed.

I think this test may be sufficient to test that the low-level profiler is actually running until we can come up with a good idea, since having a deterministic test doesn't seem to be always possible nor portable.

What I can see as a possible semi-deterministic test is to write our own C function, for which we can predict its symbol name, but that will require significant work on the build, not to mention some possible additional dependencies, I think that's a lot for little benefit.

mrocklin:
@kkraus14 is there someone within NVIDIA who we could get to test this out and provide user feedback?  This seems like it's in a pretty good state to me.

pentschev:
@mrocklin I did use it myself for a few tasks. While I understand it's nice to have more people try it out, if you have something specific you'd like tested in mind already, let me know, so we can prevent taking too much of other people's time.

mrocklin:
Nothing in particular, I'm just looking for other people's perspectives.  Keith knows people who have asked for this in the past, and so I suspect that he has a better sense for what users want here than you or I do.

pentschev:
> Nothing in particular, I'm just looking for other people's perspectives. Keith knows people who have asked for this in the past, and so I suspect that he has a better sense for what users want here than you or I do.

Got it, thanks for clarifying!

mrocklin:
I've changed things to be configurable by the config system rather than command line arguments.  This is how we handle most of the other profile configuration, and is, I think, a bit easier to manage.

Also, here is an error that I got, though I suspect that this is more likely from further down in the stack than we're operating here.

```python

In [2]: client = Client(low_level_profiler=True)

In [3]: import dask.array as da

In [4]: x = da.random.random((10000, 10000), chunks=(1000, 1000))

In [5]: (x + x.T - x.mean(axis=0)).sum().compute()
libunwind: malformed __unwind_info at 0x116CF648C bad second level page  # <--- here
Out[5]: 50003173.58741013
```

pentschev:
> I've changed things to be configurable by the config system rather than command line arguments. This is how we handle most of the other profile configuration, and is, I think, a bit easier to manage.

Sure, no worries. I asked you [about that](https://github.com/dask/distributed/pull/2575#discussion_r269137329) some time ago, but didn't get a reply.

> Also, here is an error that I got, though I suspect that this is more likely from further down in the stack than we're operating here.

Hmm, I didn't see that error before. I'll try to reproduce it and see if I can fix it.

distributed/tests/test_profile.py
```diff
@@ -45,6 +46,47 @@ def test_f():
...
+    state = create()
+
+    for i in range(100):
+        frame = sys._current_frames()[thread.ident]
+        llframes = {thread.ident: ll_get_stack(thread.ident)}
+        for f in llframes.values():
+            if f is not None:
+                llprocess(f, None, state)
+
+    assert state['count'] == 100
```
mrocklin: This test always fails for me on my OSX machine at this point.  I'm not sure why.

Is there some case where the count would be quite a bit lower than 100?

pentschev: I removed a sleep call from the for loop that processes the stack, which was wrong of me doing.

However, that still leaves another problem, we're relying on sleep for predetermined times to ensure that the the async thread has completed an iteration, this seems to work fine for the python profiler, but not for the low-level one. My guess is that it takes a little longer to process the low-level stack, therefore causing not catching all the calls.

So, what I did now was to remove the async thread completely and rely only on the main thread's sleep, that works for me on Linux and OSX. Regardless, using sleep as a means of thread synchronization is not reliable and we should avoid doing this.

pentschev:
> ```python
> libunwind: malformed __unwind_info at 0x116CF648C bad second level page  # <--- here
> ```

I assume you also saw this issue on OSX, right? I'm able to reproduce this on OSX only, Linux works.

pentschev:
> > ```python
> > libunwind: malformed __unwind_info at 0x116CF648C bad second level page  # <--- here
> > ```
> 
> I assume you also saw this issue on OSX, right? I'm able to reproduce this on OSX only, Linux works.

It looks like this is a message from llvm's libunwind, so I guess it's possible to have that on Linux if one's using llvm.

@mrocklin do we have any estimates of the amount of people using Dask distributed on OSX? I'm inclined to say we should file a bug report here to let people know we are aware of this issue but not trying to fix it right now. The issue could be in one of several places (libunwind itself, stacktrace, and less likely, Dask distributed itself), and if we're not really using it right now, I'd prefer to hold on to fixing this.

pentschev:
Regardless of how we decide to move on to tackle this, I filed an issue https://github.com/dask/distributed/issues/2608.

mrocklin:
> I assume you also saw this issue on OSX, right? I'm able to reproduce this on OSX only, Linux works.

Yes

> do we have any estimates of the amount of people using Dask distributed on OSX?

Thousands?  It gets used pretty uniformly across platforms.  Obviously folks tend to use Linux/Windows more in the multi-node distributed case, but plenty of people use dask on their laptops first, and plenty of those folks like OSX machines.

> The issue could be in one of several places (libunwind itself, stacktrace, and less likely, Dask distributed itself), and if we're not really using it right now, I'd prefer to hold on to fixing this.

OK, thanks.  I've cc'ed @sklam (who wrote stacktrace, and probably has pretty deep familiarity with LLVM and libunwind).  Hopefully he has suggestions on how to help solve this.

pentschev:
@mrocklin thanks for the docs, I resolved the conflicts to help out too. :)

mrocklin:
Oh wow, I'm impressed.  I was working through the conflicts (which are
substantial).  Happy to give up on that though :)

pentschev:
I did it all on GH's web interface, it's pretty easy to visualize and fix them, unless you need to check distant parts of the code, plus, you don't need to merge master locally before doing that.

mrocklin:
This is in.  Thanks for the work @pentschev !

pentschev:
Thanks for the review, as well as all the fixes and improvements, @mrocklin!

Issue #2580, opened 2019-03-27
Configurable Bokeh Host


quasiben:
When creating a dask-scheduler with comm protocols others than TCP the bokeh host uses the same host defined in the scheduler:

> dask-scheduler --host=ucx://QQ.XX.YY.ZZ

`QQ.XX.YY.ZZ` is probably not accessible as a TCP endpoint for the bokeh dashboard.   In this case we should optionally allow users to define where (IP) the bokeh host should spin up

cc @mrocklin @jakirkham

mrocklin:
We deal with this today with the `inproc://` comm used for single-node computation here:

https://github.com/dask/distributed/blob/b36e87d14e75c0cbdd0e3292ac65db2a836f74bd/distributed/scheduler.py#L1115-L1119

This just pins the listen host address to localhost when using inproc.  This is probably not a very good long term general solution, but at least it points to the right place to solve this problem.

Perhaps we need to optionally ask the user for an address to listen on, either for all services (you could imagine other diagnostic servers running) or for each service individually.

We do something similar today in how we accept a `LocalCluster(..., diagnostics_port=)` keyword, or the `--bokeh-port` keyword in the CLI.  Perhaps rather than consume a port we should consume an address.

https://github.com/dask/distributed/blob/b36e87d14e75c0cbdd0e3292ac65db2a836f74bd/distributed/deploy/local.py#L127-L135

This issue also comes up when people use IP over Infiniband in HPC settings.  The host address we pick up from the IB network interface isn't the host we want to use to show the dashboard.  The linking between these two has made HPC users unhappy for a while as well.

So, no solutions yet there, but hopefully the context above gives some information for people to start thinking about what a proper solution would be.  I'd be more than happy to significantly change how we get this information from users.  The current keywords leave something to be desired.

quasiben:
resolved by https://github.com/dask/distributed/pull/2589

quasiben:
tested https://github.com/dask/distributed/pull/2589 and validated the fix

PR #2583, opened 2019-03-28
small change to use localhost for services when using ucx

quasiben:
This PR resolves https://github.com/dask/distributed/issues/2580.  I decided to not go down the route of adding bokeh-host at the moment.  There is a fair bit of infrastructure in place just above the `inproc` detection which could be overridden if `bokeh_host` is set.  Alternatively, the [BokehServer](https://github.com/dask/distributed/blob/master/distributed/bokeh/core.py#L21-L377) could be set with `bokeh_host` and ignore the incoming addr to listen on.  

With that said, the above change makes UCX-bokeh interactions passable and gets Boken running on localhost rather than a UCX addr

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1112,7 +1112,8 @@ class Scheduler(ServerNode):
             if listen_ip == '0.0.0.0':
                 listen_ip = ''
 
-            if isinstance(addr_or_port, str) and addr_or_port.startswith('inproc://'):
+            if isinstance(addr_or_port, str) and (addr_or_port.startswith('inproc://')
+                                                  or addr_or_port.startswith('ucx://')):
                 listen_ip = 'localhost'
 
             # Services listen on all addresses


mrocklin:
Two comments:

1.  I suggest that the check might be if the protocol is not in `{'tcp', 'tls'}` rather than whitelisting inproc and ucx
2.  This isn't quite correct.  For `inproc` we know that we can reach the system on `localhost` because inproc is somehow directly local.  However for protocols like ucx this isn't the case.  UCX is likely to be used on machines other than localhost.

mrocklin:
@quasiben is this covered by being able to provide a full dashboard host/ip in https://github.com/dask/distributed/pull/2589

quasiben:
if both PRs are merged the behavior will be that localhost is always used for the dashboard unless `--dashboard-address=XX.XXX.XXX.XX:8787` is defined.

In point 2. in your [comment above](https://github.com/dask/distributed/pull/2583#issuecomment-477783951), are you suggesting that 127.0.0.1 is available or that in some hardware configurations the device enabloing UCX will not be able to talk to the localhost NIC ?  Like UCXdev1 is a device sitting external to the box ?

mrocklin:
I think that this was resolved elsewhere.  Closing.

Issue #2581, opened 2019-03-27
Using dask with PyTorch (train a model)


muammar:
I find Dask very useful, and would like to use it to accelerate some pytorch models I am working on. A function to be accelerated would be the `forward()` function. I started doing some tests, but I get errors like these:

```
  File "/home/muammar/.local/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 81, in serialization_error_loads
    raise TypeError(msg)
TypeError: Could not serialize object of type Tensor.
Traceback (most recent call last):
  File "/home/muammar/.local/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 144, in serialize
    header, frames = dumps(x, context=context) if wants_context else dumps(x)
  File "/home/muammar/.local/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 38, in dask_dumps
    header, frames = dumps(x)
  File "/home/muammar/.local/lib/python3.7/site-packages/distributed/protocol/torch.py", line 10, in serialize_torch_Tensor
    header, frames = serialize(t.numpy())
RuntimeError: Can't call numpy() on Variable that requires grad. Use var.detach().numpy() instead.
```

If I detach the variable, then pytorch does not work with numpy arrays. My question would be -- what is the state of pytorch support? If it is of any help, I would be more than happy to try any code (or even write something if it is under my capabilities) because it would be very useful to have this support in Dask.  Thanks for this great python module.

mrocklin:
> My question would be -- what is the state of pytorch support? 

It's been as supported as people have made support :)  I'm not sure anyone has seen the issue that you've raised here.  If you're able to isolate it that would help.  I encourage you to find some minimal reproducible example that reproduces the error above.  Ideally this looks like the following:

```python
from distributed.protocol import serialize, deserialize

t = # make some pytorch tensor
deserialize(*serialize(t))
```

That works for most pytorch tensors that we've seen so far (see tests in distributed/protocol/torch.py), so presumably your tensors are different in some way.  It would be good to find out how.

muammar:
Thanks for your reply. I should have inspected in more details the `distributed.protocol` modules. 

> It's been as supported as people have made support :)

Yes, this is true :). 

> I'm not sure anyone has seen the issue that you've raised here. If you're able to isolate it that would help. I encourage you to find some minimal reproducible example that reproduces the error above. Ideally this looks like the following:

 I can reproduce the error above with this minimal script: 

```
import torch
import numpy as np
[...]

PR #2586, opened 2019-03-29
.detach_() pytorch tensor to serialize data as numpy array.

muammar:
This commit is related to #2581. Tensors with requires_grad=True cannot
be converted to numpy unless they are detached from the computational
graph first. To fix that, the tensor has to be detached as a leaf using
the .detach_() method. For more information see:

https://pytorch.org/docs/stable/autograd.html#torch.Tensor.detach_

### DIFF
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -36,4 +36,10 @@ def test_resnet():
 def test_deserialize_grad():
     a = np.random.rand(8, 1)
     t = torch.tensor(a, requires_grad=True, dtype=torch.float)
-    deserialize(*serialize(t))
+    t2 = deserialize(*serialize(t))
+
+    assert t2.requires_grad
+
+    a = np.around(a, decimals=6)
+    t2 = np.around(t2.detach_().numpy(), decimals=6)
+    assert a.all() == t2.all()
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -31,3 +31,10 @@ def test_resnet():
     header, frames = serialize(model)
     model2 = deserialize(header, frames)
     assert str(model) == str(model2)
+
+
+def test_deserialize_grad():
+    torch.set_printoptions(precision=10)
+    a = np.random.rand(8, 1)
+    t = torch.tensor(a, requires_grad=True, dtype=torch.float)
+    deserialize(*serialize(t))
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -34,7 +34,6 @@ def test_resnet():
 
 
 def test_deserialize_grad():
-    torch.set_printoptions(precision=10)
     a = np.random.rand(8, 1)
     t = torch.tensor(a, requires_grad=True, dtype=torch.float)
     deserialize(*serialize(t))
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -38,7 +38,4 @@ def test_deserialize_grad():
     t = torch.tensor(a, requires_grad=True, dtype=torch.float)
     t2 = deserialize(*serialize(t))
     assert t2.requires_grad
-
-    a = np.around(a, decimals=6)
-    t2 = np.around(t2.detach_().numpy(), decimals=6)
-    assert a.all() == t2.all()
+    assert np.allclose(a, t2.detach_().numpy())
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -37,7 +37,6 @@ def test_deserialize_grad():
     a = np.random.rand(8, 1)
     t = torch.tensor(a, requires_grad=True, dtype=torch.float)
     t2 = deserialize(*serialize(t))
-
     assert t2.requires_grad
 
     a = np.around(a, decimals=6)
--- a/distributed/protocol/torch.py
+++ b/distributed/protocol/torch.py
@@ -7,7 +7,7 @@ import numpy as np
 
 @dask_serialize.register(torch.Tensor)
 def serialize_torch_Tensor(t):
-    header, frames = serialize(t.detach_().numpy())
+    header, frames = serialize(t.detach().numpy())
     if t.grad is not None:
         grad_header, grad_frames = serialize(t.grad.numpy())
         header['grad'] = {'header': grad_header, 'start': len(frames)}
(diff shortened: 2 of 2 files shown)


mrocklin:
Thanks for the PR @muammar !

Can I ask you to include your example in the issue as a small test?  See https://github.com/dask/distributed/blob/master/distributed/protocol/tests/test_torch.py for examples.

Also cc @stsievert

muammar:
I hope I did this correctly. Running `py.test` called the function correctly.

stsievert:
I'd like to see `requires_grad` tested. Something like

``` python
x = torch.tensor([0, 0, 0], requires_grad=True)
y = deserialize(*serialize(x))
assert y.requires_grad
```

What's the motivation behind this PR? Are models being trained with PyTorch and Dask?

distributed/protocol/tests/test_torch.py
```diff
@@ -31,3 +31,10 @@ def test_resnet():
     header, frames = serialize(model)
     model2 = deserialize(header, frames)
     assert str(model) == str(model2)
+
+
+def test_deserialize_grad():
+    torch.set_printoptions(precision=10)
+    a = np.random.rand(8, 1)
+    t = torch.tensor(a, requires_grad=True, dtype=torch.float)
+    deserialize(*serialize(t))
```
mrocklin: The result of this line should be a PyTorch tensor that is mostly equivalent to `t`.  Is there some check we can run to verify that we have recreated an equivalent tensor object?

distributed/protocol/tests/test_torch.py
```diff
@@ -31,3 +31,10 @@ def test_resnet():
     header, frames = serialize(model)
     model2 = deserialize(header, frames)
     assert str(model) == str(model2)
+
+
+def test_deserialize_grad():
+    torch.set_printoptions(precision=10)
```
mrocklin: My guess is that this line was used for debugging.  Maybe we can remove it before we merge?

distributed/protocol/tests/test_torch.py
muammar: By doing this: 

```
import torch
import numpy as np
from distributed.protocol import serialize, deserialize

a = np.random.rand(8, 1)

t = torch.tensor(a, requires_grad=True, dtype=torch.float)

t2 = deserialize(*serialize(t))
print(a)
print(t2.numpy())

```

I get: 
```
[[0.90611823]
 [0.57207592]
 [0.14868348]
 [0.24763735]
 [0.25475143]
 [0.71949837]
 [0.19011521]
 [0.3211044 ]]
[[0.9061182 ]
 [0.5720759 ]
 [0.14868347]
 [0.24763735]
 [0.25475144]
 [0.7194984 ]
 [0.19011521]
 [0.3211044 ]]
```

They slightly differ on the 8th decimal. `assert` will thorw an error because of that and I thought that setting larger precission would help.

muammar: Getting to check if they are equivalent, I could use https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.around.html#numpy.around. And truncate to the 7th or 6th decimal.

muammar:
@stsievert that is correct. I am trying to train a model using pytorch and dask.

muammar:
> I'd like to see `requires_grad` tested. Something like
> 
> ```python
> x = torch.tensor([0, 0, 0], requires_grad=True)
> y = deserialize(*serialize(x))
> assert y.requires_grad
> ```

Thanks for bringing this up. The data is being serialized/deserialized correctly. However, `.detach_()` creates a copy of the tensor with `requires_grad=False`. Shouldn't we use `.detach()` instead? because how else we could verify the tensor requires grad.

stsievert:
> `.detach_()` creates a copy of the tensor with `requires_grad=False`. Shouldn't we use `.detach()` instead? because how else we could verify the tensor requires grad.

The `requires_grad` information will need be encoded, probably in the header. Maybe

``` python
def serialize_torch(x):
    # ...
    if x.requires_grad:
	    kwargs["requires_grad"] = True
        x.requires_grad_()
    header, frames = serialize(x.numpy())
    header.update(kwargs)
    # ...
```

with a matching thing for `deserialize` that looks at the `header`.

> I am trying to train a model using pytorch and dask.

I've had a good experience training using PyTorch models with Dask when using [skorch] (the official wrapper [ignite] might work too, but I'm not sure).

[skorch]:https://skorch.readthedocs.io/en/latest/
[ignite]:https://pytorch.org/ignite/

muammar:
> > `.detach_()` creates a copy of the tensor with `requires_grad=False`. Shouldn't we use `.detach()` instead? because how else we could verify the tensor requires grad.
> 
> The `requires_grad` information will need be encoded, probably in the header. Maybe
> 
> ```python
> def serialize_torch(x):
>     # ...
>     if x.requires_grad:
> 	    kwargs["requires_grad"] = True
>         x.requires_grad_()
>     header, frames = serialize(x.numpy())
>     header.update(kwargs)
>     # ...
> ```
> 
> with a matching thing for `deserialize` that looks at the `header`.
> 

Inspired by your comment I came up with https://github.com/dask/distributed/pull/2586/commits/d9143a0e200a6ed5ea67be51bc8844f1b3a6357d. It works correctly for `requires_grad` true and false. 

> > I am trying to train a model using pytorch and dask.
> 
> I've had a good experience training using PyTorch models with Dask when using [skorch](https://skorch.readthedocs.io/en/latest/) (the official wrapper [ignite](https://pytorch.org/ignite/) might work too, but I'm not sure).

Thanks for sharing!. I will look into it, but would be very glad if I would not need to add another dependency.

distributed/protocol/tests/test_torch.py
stsievert: > They slightly differ on the 8th decimal

I use `np.allclose` in these situations.

muammar: @stsievert that worked. Thanks for the suggestion.

mrocklin:
@stsievert does this look ok to you?  If so, I'm going to merge this in.

stsievert:
> does this look ok to you?

Yes. This PR more accurately represents the tensor that's communicated. The `require_grad` flag specifies that the gradients should be recorded when performing `loss.backwards()`. 

The gradients do eventually need to be connected to an optimizer. @muammar implemented a parameter server for that (https://github.com/dask/distributed/issues/2581#issuecomment-479024638).

mrocklin:
Thanks @muammar .  This is in!

muammar:
> Thanks @muammar . This is in!
> 


You are right, this is my first contribution to dask!! thanks for accepting the PR.

PR #2587, opened 2019-04-01
Support uploading files with multiple modules

### LINKED ISSUE #1819: Cannot properly upload files with multiple modules or modules with an underscore in their name
When using the client.upload_file method to upload an egg file, the upload fails if the name of the module and resulting egg file contain an underscore character. In addition, only the package whose name matches the name of the egg file is imported and reloaded.

For example, if you make an egg file from a module "test_module",

`client.upload_file("test_module-1.0-py3.6.egg") `

attempts to import and then reload a module "test-module" (and then fails as a result), even though the original module was "test_module".

In addition it does not appear import any other modules in this egg file.

This appears to be down to the implementation of the import_file method in utils.py (called from the worker). 

This uses pkg_resources.find_distributions to get the project_name from the egg file, and then assumes that this project name is the name of the module to import and reload later. However, the package name has had 'safe_name' called on it when it was created, which converts any non alpha-numeric characters into hyphens.

charmoniumQ:
Fixes #1819


distributed/tests/test_client.py
```diff
@@ -1550,6 +1551,38 @@ def g():
...
+        dir_contents = {
+                'setup.py': (
+                    'from setuptools import setup, find_packages\n'
+                    'setup(name="my_package", packages=find_packages())\n'
+                ),
+                'package_1/__init__.py': 'a = {}'.format(value),
+                'package_2/__init__.py': 'b = {}'.format(value),
+        }
+        with tmp_dir(dir_contents) as directory:
+            subprocess.check_call(['python', 'setup.py', 'bdist_egg'], cwd=directory)
```
mrocklin: Perhaps `sys.executable` here rather than `'python'`?

distributed/tests/test_client.py
```diff
@@ -1550,6 +1551,38 @@ def g():
...
+            egg_root = os.path.join(directory, 'dist')
+            print(os.listdir(directory))
+            print(os.listdir(os.path.dirname(egg_root)))
+            print(os.listdir(egg_root))
```
mrocklin: Probably want to remove these print statements

distributed/tests/test_client.py
```diff
@@ -1550,6 +1551,38 @@ def g():
...
+            egg_name = [fname for fname in os.listdir(egg_root) if fname.endswith('.egg')][0]
+            egg_path = os.path.join(egg_root, egg_name)
+            yield c.upload_file(egg_path)
+
+            x = c.submit(g, pure=False)
+            result = yield x
+            assert result == (value, value)
+    # this test won't impact the others, because the dir containg the
+    # egg is gone at this point.
```
mrocklin: Just to be clear, will this affect the user's environment?  After running this test will a user find that they have  new packages named `package_1` and `package_2`?

distributed/tests/test_client.py
mrocklin: What happens if the user doesn't have install privileges to where their Python is installed, such as if they're using their system Python?

distributed/utils.py
```diff
@@ -478,6 +478,28 @@ def tmp_text(filename, text):
...
+            pass # we don't care if dir already exists
+
+        with open(full_fname, 'w') as f:
+            f.write(text)
+
+    try:
+        yield directory
+    finally:
+        if os.path.exists(directory):
+            shutil.rmtree(directory)
```
mrocklin: There is already a `dask.utils.tmpdir`.  I think that `dask.utils.tmpfile` probably also works (it is robust to the name being used as a directory).

distributed/utils.py
```diff
@@ -1041,7 +1063,8 @@ def import_file(path):
             import pkg_resources
             pkgs = pkg_resources.find_distributions(path)
             for pkg in pkgs:
-                names_to_import.append(pkg.project_name)
+                # See https://github.com/dask/distributed/issues/1819
+                names_to_import.extend(pkg.get_metadata_lines('top_level.txt'))
```
mrocklin: cc @jakirkham, who probably knows more about packaging than I do.

distributed/tests/test_client.py
charmoniumQ: This does not _install_ `pakage_1/package_2`. It merely _compiles_ them into an egg. You can verify this by running a simple `setup.py` on a non-root account.

```
    $ cat setup.py
    from setuptools import setup, find_packages
    setup(
        name='my_package',
        packages=find_packages(),
    )
    $ python3 setup.py bdist_egg

```

distributed/tests/test_client.py
charmoniumQ: Each worker modifies their own `sys.path`. This change does not persist after the worker's process dies (they get closed by `gen_cluster`).

The only persisting side-effect is that [it writes a file to the worker's `local_dir`](https://github.com/dask/distributed/blob/master/distributed/worker.py#L688). I don't know if we care about this.

charmoniumQ: Wait... Are the workers spawned by `gen_cluster` running in different processes?

mrocklin: No, unless you provide the `Worker=Nanny` keyword argument.

mrocklin:
The Python 2 test failure looks possibly genuine.

```
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/ioloop.py:576: in run_sync
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/concurrent.py:261: in result
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/gen.py:1147: in run
distributed/utils_test.py:890: in coro
    result = yield future
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/gen.py:1133: in run
    value = future.result()
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/concurrent.py:261: in result
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/gen.py:1141: in run
    yielded = self.gen.throw(*exc_info)
distributed/tests/test_client.py:1588: in test_upload_file_egg
    yield c.upload_file(egg_path)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/gen.py:1133: in run
    value = future.result()
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/concurrent.py:261: in result
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/tornado/gen.py:1147: in run
self = <Client: not connected>
filename = '/tmp/tmpa0VuWy./dist/my_package-0.0.0-py2.7.egg'
raise_on_error = True
    @gen.coroutine
    def _upload_file(self, filename, raise_on_error=True):
        with open(filename, 'rb') as f:
            data = f.read()
        _, fn = os.path.split(filename)
        d = yield self.scheduler.broadcast(msg={'op': 'upload_file',
                                                'filename': fn,
                                                'data': to_serialize(data)})
    
        if any(v['status'] == 'error' for v in d.values()):
            exceptions = [v['exception'] for v in d.values()
                          if v['status'] == 'error']
            if raise_on_error:
>               raise exceptions[0]
E               ZipImportError: bad local file header: /home/travis/build/dask/distributed/dask-worker-space/worker-XKQzqJ/my_package-0.0.0-py2.7.egg
distributed/client.py:2666: ZipImportError
```

(if so, one solution to this would be to just wait a couple weeks, at which point we may have dropped Python 2)

charmoniumQ:
I think this is due to caching in Python's HTTP library or importing library. Either way, using a unique version (and thus unique egg-name) seems to fix the problem.

mrocklin:
Agreed that these test failures are unrelated .  Restarting them now.

distributed/utils.py
mrocklin: @jakirkham if you find yourself with a few free minutes can I ask you to look this over briefly?

mrocklin:
I've changed the title of the PR to something that I think reflects what you're trying to acheive @charmoniumQ , but please reword it if there is something better (this title becomes the commit message).  I've added the issue number in the comment section with the Github magic word "Fixes" in front of it so that they link properly.

I've asked @jakirkham to briefly review the module handling (should only take a couple minutes hopefully if everything is good), which I don't really know much about.

distributed/utils.py
mrocklin: I spoke with @jakirkham briefly and he didn't have a quick answer.  I think he'll take a longer look sometime today or tomorrow.

jakirkham: Thoughts on using [`pkgutil.iter_modules`]( https://docs.python.org/3/library/pkgutil.html#pkgutil.iter_modules )?

charmoniumQ: `pkgutil.iter_modules` is better it is more general: it applies to the zip and pyz case as well, and it works for custom importers. I've switched the code over to that

jakirkham: Glad to hear that works. Also nice to see the code simplification here 🙂

charmoniumQ:
The AppVeyor test failure is spurious AFAICT.

jakirkham:
Restarted AppVeyor.

jakirkham:
@mrocklin, any thoughts on the rest of this PR?

distributed/tests/test_client.py
```diff
@@ -1511,46 +1509,80 @@ def g():
...
-            yield c._upload_large_file(fn, remote_filename="x")
-            yield c._upload_large_file(fn2)
-
-            for w in [a, b]:
-                assert os.path.exists(os.path.join(w.local_dir, "x"))
-                assert os.path.exists(os.path.join(w.local_dir, "myfile2"))
-                with open(os.path.join(w.local_dir, "x")) as f:
-                    assert f.read() == "abc"
-                with open(os.path.join(w.local_dir, "myfile2")) as f:
-                    assert f.read() == "def"
```
mrocklin: What happened to this test?

mrocklin:
> @mrocklin, any thoughts on the rest of this PR?

Whoops.  Sorry for abandoning this @charmoniumQ .  I left a comment about the upload_large_files test but, other than that this seems fine to me, given that jakirkham has signed off.

jakirkham:
Poked AppVeyor again.

mrocklin:
Thanks for this @charmoniumQ .  Merging.

PR #2588, opened 2019-04-01
Use ensure_bytes in serialize_error

mrocklin:


mrocklin:
This came out of a conversation with @quasiben using UCX, which produces `memoryview` objects when receiving.

jakirkham:
We can avoid a copy if we use `codecs.decode`. Proposing that change in PR ( https://github.com/dask/distributed/pull/6274 )

Issue #2565, opened 2019-03-15
GPU-friendly loads / merge_frames


TomAugspurger:
Just throwing this up here for now, need to investigate more.

I'm working on a [distributed cudf](https://github.com/TomAugspurger/dask-perf/blob/master/bench_cudf_join.py) join using UCX. Things progress fine until, in the Client process, we attempt to deserialize some data (I think the final result?). We end up calling calling `loads` with `deserialize=True`: https://github.com/dask/distributed/blob/fb30c33562862f30864456766424b44a3e91aa5b/distributed/protocol/core.py#L101

which calls `merge_frames`: https://github.com/dask/distributed/blob/fb30c33562862f30864456766424b44a3e91aa5b/distributed/protocol/utils.py#L80

which attempt to convert the data to a byte string.

At this point in the client process, `frames` is a list of objects representing device memory. If possible (and I *think* it's possible), I'd like to avoid copying to the host here.

Actually, this may only be possible if the Client happens to have a GPU as well. In this case that's true, but not in general.

---

TODO:

1. figure out exactly where the client is calling this
2. ...

mrocklin:
Hrm, that does seem like incorrect behavior in the GPU memory case

It also means that these lines weren't triggered.  
https://github.com/dask/distributed/blob/fb30c33562862f30864456766424b44a3e91aa5b/distributed/protocol/utils.py#L60-L61

This is somewhat surprising, especially given that UCX handles message framing itself.  I wonder if this is a case where we should be using `nbytes` rather than `len`.

TomAugspurger:
Ahh... I completely missed those lines. That may actually explain it. I've been hacking up the serialization code to make things work. I likely have a bug in there.

TomAugspurger:
So, after looking at this further, there may still be some merit to this issue.

I'm sending a cudf DataFrame with three columns and no nulls, so that's a total of 3 numba device arrays (I ignore the index for now).

Somewhere on the send size (`frame_split_size` perhaps?) the three arrays are split into 9. When means that the condition for the early return at 
https://github.com/dask/distributed/blob/fb30c33562862f30864456766424b44a3e91aa5b/distributed/protocol/utils.py#L60-L61 is not met.

```python
(Pdb) len(frames)
72
(Pdb) len(lengths)
3
(Pdb) len(gpu_frames)
9
(Pdb) sum(x.nbytes for x in gpu_frames) == sum(lengths)
True
```

[...]

PR #2591, opened 2019-04-02
Add UCX Comm

quasiben:
This PR brings in much of the work done by @TomAugspurger with ucx/ucx-py with @mrocklin's help I added a small change to cudf protocol handling and general cleanup.


distributed/protocol/utils.py
```diff
@@ -55,6 +55,10 @@ def merge_frames(header, frames):
     if not frames:
         return frames
 
+    if any(x.__class__.__name__ == 'MemoryPointer' for x in frames):
+        # XXX
+        return frames
```
mrocklin: @TomAugspurger can you expand on the issue here?  I'm just curious if there is something we can start up now in parallel so that this can go away over time.

distributed/protocol/cupy.py
```diff
@@ -0,0 +1,39 @@
...
+        data = x.ravel()  # order='K'
+    else:
+        x = cupy.ascontiguousarray(x)
+        strides = x.strides
+        data = x.ravel()
+
+    dtype = (0, x.dtype.str)
+
+    header = x.__cuda_array_interface__.copy()
+    header['lengths'] = (x.nbytes,)  # one per stride
```
mrocklin: Ideally we wouldn't need this.  Dask should populate this automatically if the `nbytes` function works on the bytes-like objects returned as frames.

distributed/protocol/cupy.py
```diff
@@ -0,0 +1,39 @@
...
+    header['compression'] = (None,)  # TODO
```
mrocklin: You can probably also omit this.  Dask should handle this further down.

distributed/protocol/cudf.py
```diff
@@ -0,0 +1,74 @@
...
+            null_masks.append(frame)
+            null_counts[label] = col.null_count
+
+    arrays.extend(null_masks)
+
+    header = {
+        'is_cuda': len(arrays),
+        'subheaders': sub_headers,
+        # TODO: the header must be msgpack (de)serializable.
+        # See if we can avoid names, and just use integer positions.
```
mrocklin: Why avoid names?  For fear that they won't be msgpack-serializable?

distributed/deploy/tests/test_local.py
```diff
@@ -555,5 +555,20 @@ def _():
...
+
+
+def test_protocol_tcp(loop):
+    with LocalCluster(protocol='tcp', loop=loop, processes=False) as cluster:
+        assert cluster.scheduler.address.startswith('tcp://')
+
+
+def test_protocol_ip(loop):
+    with LocalCluster(ip='tcp://127.0.0.2', loop=loop, processes=False) as cluster:
+        assert cluster.scheduler.address.startswith('tcp://127.0.0.2')
```
mrocklin: I believe that these tests are now in master

distributed/comm/ucx.py
```diff
@@ -0,0 +1,336 @@
...
+
+    async def read(self, deserializers=None):
+        resp = await self.ep.recv_future()
+        obj = ucp.get_obj_from_msg(resp)
+        n_frames, = struct.unpack("Q", obj)
+        n_data_frames = n_frames - 2
+
+        gpu_frame_msg = await self.ep.recv_future()
+        gpu_frame_msg = gpu_frame_msg.get_obj()
+        is_gpu = struct.unpack("{}?".format(n_data_frames), gpu_frame_msg)
```
mrocklin: Can one message contain both GPU and non-GPU frames?

mrocklin: Ah, below I see that this is unpacked into another `is_gpu` variable.  We might want to use a different variable name here that helps us recognize that it is a list.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,336 @@
...
+    async def write(self, msg: dict, serializers=None, on_error: str = "message"):
+        frames = await to_frames(
+            msg, serializers=serializers, on_error=on_error
+        )  # TODO: context=
+        gpu_frames = b''.join([struct.pack("?", hasattr(frame, '__cuda_array_interface__'))
+                               for frame in frames])
+        size_frames = b''.join([struct.pack("Q", nbytes(frame)) for frame in frames])
+
+        frames = [gpu_frames] + [size_frames] + frames
+        nframes = struct.pack("Q", len(frames))
```
mrocklin: I wouldn't mind avoiding the suffix `_frames` above for `gpu_frames` and `size_frames`.  I typically think of a frame as a large buffer that we're sending.  These appear to be more metadata.  Perhaps `is_gpus` and sizes` or something similar?

distributed/comm/ucx.py
TomAugspurger: I had unpushed changes locally renaming this to `is_gpus` :)

distributed/protocol/cudf.py
TomAugspurger: That, and duplicate columns.

@quasiben sorry for the mess, https://github.com/TomAugspurger/distributed/tree/ucx%2Bhacky-serialization has a largeish update to serialization. I thought I had merged it into my `ucx` branch, but apparently not. I think the last 4 commits are relevant, but it's a bit of a mess. https://github.com/TomAugspurger/distributed/commits/ucx%2Bhacky-serialization

distributed/protocol/utils.py
TomAugspurger: I was in the midst of fixing this when baby came. https://github.com/TomAugspurger/distributed/blob/1307616d513cec751eca6bc3a9a267b5871ee127/distributed/protocol/utils.py#L67-L85 has the "fixed" version.

That fix currently relies Numba DeviceNDArray as a stand-in for a gpu-memoryview. I would recommend that we (dask / rapids / pydata community members) define a lightweight object that implements `__cuda_array_interface__`, which both distributed and ucx-py can rely on.

TomAugspurger: This is fixing https://github.com/dask/distributed/issues/2565

TomAugspurger: And looking at my local directory, I have an uncommitted `import pdb; pdb.set_trace()`, so this likely isn't complete done :)

@quasiben this may be worth looking into before that segfault we spoke about earlier. It's possible (but not certain) that this fixes it.

quasiben: Thanks for the tip!

quasiben:
With @TomAugspurger suggestion and a few edits tests still pass and naively doing an experiment with of

```python
N = 1_000
P = 1_000
X = da.random.uniform(size=(N, P), chunks=(N//100, P))
print(format_bytes(X.nbytes))
result = X + X.T
start = clock()
result.compute()
stop = clock()
print(result)
```

However, when doing a large join operation I am getting the following (and much more interesting error):
https://gist.github.com/quasiben/5fa1cdd37196a5a25877541e2814e229

notebook can be found here:
https://gist.github.com/mrocklin/a80d3de513c2c21dda515b7d89abd152

mrocklin:
Seeing zict means that you're probably spilling to disk.  I think that the current spill-to-disk solution uses `serialize` and expects that to return something that we can write to a file handle, which may not be the case currently.

Short term I recommend scaling down the computation a bit.

quasiben:
@Akshay-Venkatesh @jakirkham 
 I have a small gist related to this PR which demonstrates the segfault I am currently trying to resolve https://gist.github.com/quasiben/926d82fdcca861a131828b63edb9b1b7

mrocklin:
@quasiben I hope you don't mind, but I merged in master, @rjzamora 's change that points to `ucp` rather than `ucp_py`, and removed some print statements.  Now the following works pleasantly for me.

```python
In [1]: from dask.distributed import LocalCluster, Client

In [2]: cluster = LocalCluster(protocol='ucx', interface='ib0', dashboard_address="10.31.241.45:8787", local_dir='/raid/mrocklin')

In [4]: import dask.array as da

In [5]: x = da.random.random((10000, 10000))

In [6]: x.sum().compute()
Out[6]: 50005228.45232478
```

Although the `dashboard_address=` and `local_dir=` keywords are specific to my particular setup.  I chose the address of the publicly visible ethernet network interface for the dashboard address, and a local scratch disk for the `local_dir=`.

I'm still running into some performance issues.  I get 700MB/s rather than the 10GB/s on one of the ucx-py benchmarks (similar to what @rjzamora reported as well), but things do seem usable.  I had to build from source.

mrocklin:
@quasiben I'm more or less hijacking your branch.  I hope you don't mind :)

mrocklin:
The performance issues mentioned above seemed to go away in the benchmark when I moved to a different machine (dgx-15 for those in NVIDIA).  However performance while in Dask is still slowish.  I plan to set up a benchmark that just uses Dask comms (not the rest of Dask) to see if I can do some profiling and see what's up.

TomAugspurger:
> I plan to set up a benchmark that just uses Dask comms (not the rest of Dask) to see if I can do some profiling and see what's up.

https://github.com/TomAugspurger/dask-perf/blob/master/bench_comm.py (based on Antoine's) may be helpful.

quasiben:
don't mind at all!

mrocklin:
> https://github.com/TomAugspurger/dask-perf/blob/master/bench_comm.py (based on Antoine's) may be helpful.

Thanks for the pointer Tom!  Super helpful :)

distributed/comm/tests/test_ucx.py
```diff
@@ -0,0 +1,290 @@
...
+    msg = {"op": "ping", "data": to_serialize(df)}
+
+    await com.write(msg)
+    result = await serv_com.read()
+    data2 = result.pop("data")
+    assert result["op"] == "ping"
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("shape", [(100,), (10, 10), (4_947,)])
```
mrocklin: @pentschev I'm able to get this test to fail (it hangs actually) if I have an array of this size or larger (one fewer element and things are fine).  Does this sound familiar to you?  I've merged in your latest changes from master.

mrocklin: It looks like we get stuck here:

https://github.com/dask/distributed/pull/2591/files#diff-ca5c24c584111a291f763107e4065539R168

We finish the `self.ep.send_obj(frame)` call, but then get stuck waiting on the future, which never completes.  @Akshay-Venkatesh, does this sound familiar to you?  Is there some behavior that might change around `4946 * 8 == 39568` bytes?  We send a few frames before the main piece of data.  These have the following lengths:

```
(Pdb) pp list(map(len, frames[:-1]))
[4, 32, 0, 9, 259]
```

mrocklin: Just to be clear here, we're getting block on waiting for the *send* to complete, which seems odd to me.

Akshay-Venkatesh: The message size sticks out because it would likely be the range where rendezvous protocol is used. So unlike the short message sizes, if the corresponding receive isn't posted, then this send operation blocks until such time. Is the receive being posted?

Akshay-Venkatesh: Also, can you send me repro commands for `test_ucx.py`? Matt, looks you're able to get this test to fail (https://github.com/dask/distributed/pull/2591/files#r279561588) so if I can repro I maybe more useful.

mrocklin: (thinking about the behavior of recvs in dask, one sec)

In the meantime, this is maybe also relevant.  Here is how this manifests itself in practice.

```python

In [2]: client = Client(protocol='ucx', interface='ib0')

In [3]: import cupy

In [4]: x = cupy.random.randint(0, 255, size=30000, dtype='uint8')

In [5]: client.scatter(x).result()  # roundtrip
Out[5]: array([152, 251, 139, ...,   4, 136, 137], dtype=uint8)

In [6]: x = cupy.random.randint(0, 255, size=40000, dtype='uint8')

In [7]: client.scatter(x).result()  # roundtrip
[dgx15:72708:0:72858] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x28)
==== backtrace ====
    0  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(+0x21b50) [0x7fce860f4b50]
    1  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(+0x21dc4) [0x7fce860f4dc4]
    2  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_rkey_destroy+0x7) [0x7fce867f8b27]
    3  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_rndv_progress_rma_get_zcopy+0x445) [0x7fce86814375]
    4  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_rndv_rtr_handler+0x65d) [0x7fce8681880d]
    5  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libuct.so.0(uct_rc_mlx5_iface_progress+0x11e) [0x7fce8657146e]
    6  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_worker_progress+0x4a) [0x7fce867fde8a]
    7  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0xace4c) [0x7fce86b01e4c]
    8  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x73fb5) [0x7fce86ac8fb5]
    9  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x25e9d) [0x7fce86a7ae9d]
   10  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x477a3) [0x7fce86a9c7a3]
   11  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(PyCFunction_Call+0x65) [0x55baae8efb65]
   12  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x1eb70) [0x7fce86a73b70]
   13  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x46dc3) [0x7fce86a9bdc3]
   14  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyMethodDef_RawFastCallKeywords+0x305) [0x55baae92c785]
   15  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyCFunction_FastCallKeywords+0x21) [0x55baae92c801]
   16  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x544e) [0x55baae98887e]
   17  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyGen_Send+0x14c) [0x55baae92551c]
   18  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/lib-dynload/_asyncio.cpython-37m-x86_64-linux-gnu.so(+0xabc9) [0x7fcea9a24bc9]
   19  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyObject_FastCallKeywords+0x49b) [0x55baae92cd2b]
   20  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(+0x20abc3) [0x55baae9bdbc3]
   21  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyMethodDef_RawFastCallDict+0x199) [0x55baae8ef8b9]
   22  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyCFunction_FastCallDict+0x21) [0x55baae8efae1]
   23  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x5d21) [0x55baae989151]
   24  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallKeywords+0xfb) [0x55baae92b79b]
   25  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x6a0) [0x55baae983ad0]
```

mrocklin: Right, so in that test if I apply the following diff then things no longer hang, instead I get a segfault like what is shown above (in practice, there is always a pending recv).

```diff

-    await com.write(msg)
-    result = await serv_com.read()
+    _, result = await asyncio.gather(com.write(msg), serv_com.read())
```

Akshay-Venkatesh: @mrocklin for the `rkey_destroy` error, can you run the experiment by setting `UCX_MEMTYPE_CACHE=n UCX_RNDV_SCHEME=put_zcopy` env vars (in case you've not already). We've seen this before.

mrocklin: Hooray!

```python
In [2]: client = Client(protocol='ucx', interface='ib0')

In [3]: import cupy
In [4]: x = cupy.random.randint(0, 255, size=10000000, dtype='uint8')

In [5]: client.scatter(x).result()  # roundtrip
Out[5]: array([166, 159,  69, ..., 226,   2,   0], dtype=uint8)
```

OK, how should we handle this for end users?  Is this something that should be fixed upstream somewhere?  Is this where https://github.com/Akshay-Venkatesh/ucx-py/issues/83 comes in?

mrocklin: Also, it looks like there might be a size limit on passed frames?  (if so, we can happily handle this on our end).

```python

In [2]: client = Client(protocol='ucx', interface='ib0', n_workers=1)

In [3]: import cupy

In [4]: x = cupy.random.randint(0, 255, size=100_000_000, dtype='uint8')

In [5]: %timeit client.scatter(x, hash=False, direct=True).result()
1.26 s ± 7.58 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: x = cupy.random.randint(0, 255, size=1_000_000_000, dtype='uint8')

In [7]: %timeit client.scatter(x, hash=False, direct=True).result()
[1556581592.967171] [dgx15:81863:0]         ucp_mm.c:108  UCX  ERROR failed to register address (nil) length 1000000000 on md[2]=ib/mlx5_3: Invalid parameter
[1556581592.967206] [dgx15:81863:0]    ucp_request.c:263  UCX  ERROR failed to register user buffer datatype 0x8 address (nil) len 1000000000: Invalid parameter
[dgx15:81863:0:595] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace ====
    0  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(+0x21b50) [0x7fe96f58ab50]
    1  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(+0x21dc4) [0x7fe96f58adc4]
    2  /lib/x86_64-linux-gnu/libc.so.6(+0x14e156) [0x7fe9816e6156]
    3  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_rndv_data_handler+0xf3) [0x7fe96fcaec83]
    4  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libuct.so.0(uct_rc_mlx5_iface_progress+0x11e) [0x7fe96fa0746e]
    5  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_worker_progress+0x4a) [0x7fe96fc93e8a]
    6  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0xace4c) [0x7fe96ff97e4c]
    7  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x6dad9) [0x7fe96ff58ad9]
    8  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyMethodDef_RawFastCallKeywords+0x141) [0x5592534465c1]
    9  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyCFunction_FastCallKeywords+0x21) [0x559253446801]
   10  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(+0x20abc3) [0x5592534d7bc3]
   11  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyMethodDef_RawFastCallDict+0x199) [0x5592534098b9]
   12  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyCFunction_FastCallDict+0x21) [0x559253409ae1]
   13  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x5d21) [0x5592534a3151]
```

Also, now I really want NVLINK for ucx-py :)

Akshay-Venkatesh: It's a combination of both, Matt. The immediate action would be to set this env var by default such that whenever `ucp` module is loaded we export these env vars.  Long term is to get this fixed in UCX C library so that we don't have to set the env var explicitly. This is because `get` rndv protocol can yield much better overlap.

Akshay-Venkatesh: NVLINK PR has stalled recently. Will poll on the UCX devs now.

pentschev: Looks like you've already figured out what was the issue here, it isn't related to the deserialization PR from yesterday, is that right @mrocklin?

mrocklin: Correct.  This was something different.

mrocklin: Results on cupy arrays.  I get something like 3GB/s.

```python
from bench_comm import run_bench
import asyncio
import cupy

x = cupy.random.randint(0, 255, 200_000_000, dtype='uint8')

loop = asyncio.get_event_loop()
%prun loop.run_until_complete(run_bench('ucx', 0, 50, data=x))
```

```
         34155 function calls (34133 primitive calls) in 3.480 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      506    2.138    0.004    2.138    0.004 {method 'poll' of 'select.epoll' objects}
      301    0.719    0.002    0.720    0.002 {method 'recv_obj' of 'ucp_py.Endpoint' objects}
        1    0.450    0.450    0.450    0.450 {ucp_py.init}
        1    0.057    0.057    0.057    0.057 {ucp_py.get_endpoint}
      503    0.025    0.000    0.026    0.000 {method 'send_obj' of 'ucp_py.Endpoint' objects}
      506    0.008    0.000    3.479    0.007 base_events.py:1679(_run_once)
      824    0.007    0.000    1.323    0.002 {method 'run' of 'Context' objects}
      302    0.004    0.000    0.742    0.002 ucx.py:177(read)
      506    0.004    0.000    2.143    0.004 selectors.py:451(select)
      513    0.003    0.000    0.005    0.000 base_events.py:707(_call_soon)
      202    0.003    0.000    0.003    0.000 {method 'recv_future' of 'ucp_py.Endpoint' objects}
      301    0.003    0.000    0.047    0.000 ucx.py:154(write)
      201    0.002    0.000    0.002    0.000 {msgpack._cmsgpack.unpackb}
      203    0.002    0.000    0.026    0.000 gen.py:184(wrapper)
      202    0.002    0.000    0.002    0.000 {method 'recv' of '_socket.socket' objects}
      101    0.002    0.000    0.006    0.000 core.py:111(loads)
      824    0.002    0.000    1.325    0.002 events.py:86(_run)
```

distributed/comm/ucx.py
```diff
@@ -0,0 +1,360 @@
...
+        for frame in frames:
+            await self.ep.send_obj(frame)
+        return sum(map(nbytes, frames))
+
+    async def read(self, deserializers=None):
+        resp = await self.ep.recv_future()
+        obj = ucp.get_obj_from_msg(resp)
+        nframes, = struct.unpack("Q", obj[:8])
+
+        gpu_frame_msg = obj[8 : 8 + nframes]
```
quasiben: What do you think about adding a small note on why an offset of `8` ?

distributed/comm/tests/test_ucx.py
mrocklin: Sometimes I run into this issue:

```
[1556630301.486209] [dgx15:8953 :0]         ucp_mm.c:108  UCX  ERROR failed to register address (nil) length 200000000 on md[5]=ib/mlx5_0: Invalid parameter
[1556630301.486234] [dgx15:8953 :0]    ucp_request.c:263  UCX  ERROR failed to register user buffer datatype 0x8 address (nil) len 200000000: Invalid parameter
[dgx15:8953 :0:8953] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace ====
    0  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(+0x21b50) [0x7fd635615b50]
    1  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(+0x21dc4) [0x7fd635615dc4]
    2  /lib/x86_64-linux-gnu/libc.so.6(+0x14e156) [0x7fd644873156]
    3  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_rndv_data_handler+0xf3) [0x7fd635d39c83]
    4  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libuct.so.0(uct_rc_mlx5_iface_progress+0x11e) [0x7fd635a9246e]
    5  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucp.so.0(ucp_worker_progress+0x4a) [0x7fd635d1ee8a]
    6  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0xace4c) [0x7fd636022e4c]
    7  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x73fb5) [0x7fd635fe9fb5]
    8  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x25e9d) [0x7fd635f9be9d]
    9  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x477a3) [0x7fd635fbd7a3]
   10  python(PyCFunction_Call+0x65) [0x55fb22b88b65]
   11  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x1eb70) [0x7fd635f94b70]
   12  /home/nfs/mrocklin/ucx-py/ucp/_libs/ucp_py.cpython-37m-x86_64-linux-gnu.so(+0x46dc3) [0x7fd635fbcdc3]
   13  python(_PyMethodDef_RawFastCallKeywords+0x305) [0x55fb22bc5785]
   14  python(_PyCFunction_FastCallKeywords+0x21) [0x55fb22bc5801]
   15  python(_PyEval_EvalFrameDefault+0x544e) [0x55fb22c2187e]
```

Is this the result of sending around frames that are too large?  Any thoughts @Akshay-Venkatesh?

distributed/comm/ucx.py
mrocklin: Sure.  Done in ecdf3cb

mrocklin:
And, it looks like we can now move cudf dataframes somewhat fast again

```python
from bench_comm import run_bench
import asyncio
import cupy
import cudf

x = cupy.random.randint(0, 255, 100_000_000, dtype='uint8')
df = cudf.DataFrame({'x': x})

loop = asyncio.get_event_loop()
loop.run_until_complete(run_bench('ucx', 0, 50, data=df))
```

```
duration: 1.7278750967234373 => rate: 2893 MB/s
```

Still no NVLink (Akshay has a PR upstream for this), but at least now we're within a factor of Infiniband speeds

mrocklin:
Trying the cudf join computation, I run into this:

```
[1556632175.904788] [dgx15:11162]            event.c:499  UCX  WARN cuMemGetAddressRange(devPtr=0x7ffa22c00200) failed
[dgx15:11162:0:11185]       event.c:502  UCX Assertion `dptr == (void *)pbase' failed
==== backtrace ====
    0  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(ucs_fatal_error+0x107) [0x7ffb04574dd7]
    1  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucm.so.0(+0x914f) [0x7ffb0479b14f]
    2  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucm.so.0(ucm_cudaFree+0x6e) [0x7ffb0479a21e]
    3  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/site-packages/../../librmm.so(rmmFree+0x128) [0x7ffa5c930988]
    4  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/lib-dynload/../../libffi.so.6(ffi_call_unix64+0x4c) [0x7ffb0e0ecec0]
    5  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/lib-dynload/../../libffi.so.6(ffi_call+0x22d) [0x7ffb0e0ec87d]
    6  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/site-packages/_cffi_backend.cpython-37m-x86_64-linux-gnu.so(+0x1a194) [0x7ffa6c024194]
    7  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyObject_FastCallKeywords+0x49b) [0x55d9e869bd2b]
    8  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x537e) [0x55d9e86f77ae]
    9  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallKeywords+0xfb) [0x55d9e869a79b]
   10  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x4aa9) [0x55d9e86f6ed9]
   11  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalCodeWithName+0xbb9) [0x55d9e8638db9]
   12  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallDict+0x1d5) [0x55d9e86395d5]
   13  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x1e20) [0x55d9e86f4250]
   14  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalCodeWithName+0x2f9) [0x55d9e86384f9]
   15  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallDict+0x1d5) [0x55d9e86395d5]
```

cc @Akshay-Venkatesh

Akshay-Venkatesh:
> Trying the cudf join computation, I run into this:
> 
> ```
> [1556632175.904788] [dgx15:11162]            event.c:499  UCX  WARN cuMemGetAddressRange(devPtr=0x7ffa22c00200) failed
> [dgx15:11162:0:11185]       event.c:502  UCX Assertion `dptr == (void *)pbase' failed
> ==== backtrace ====
>     0  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucs.so.0(ucs_fatal_error+0x107) [0x7ffb04574dd7]
>     1  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucm.so.0(+0x914f) [0x7ffb0479b14f]
>     2  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/libucm.so.0(ucm_cudaFree+0x6e) [0x7ffb0479a21e]
>     3  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/site-packages/../../librmm.so(rmmFree+0x128) [0x7ffa5c930988]
>     4  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/lib-dynload/../../libffi.so.6(ffi_call_unix64+0x4c) [0x7ffb0e0ecec0]
>     5  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/lib-dynload/../../libffi.so.6(ffi_call+0x22d) [0x7ffb0e0ec87d]
>     6  /home/nfs/mrocklin/miniconda/envs/ucx-dev/lib/python3.7/site-packages/_cffi_backend.cpython-37m-x86_64-linux-gnu.so(+0x1a194) [0x7ffa6c024194]
>     7  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyObject_FastCallKeywords+0x49b) [0x55d9e869bd2b]
>     8  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x537e) [0x55d9e86f77ae]
>     9  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallKeywords+0xfb) [0x55d9e869a79b]
>    10  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x4aa9) [0x55d9e86f6ed9]
>    11  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalCodeWithName+0xbb9) [0x55d9e8638db9]
>    12  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallDict+0x1d5) [0x55d9e86395d5]
>    13  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalFrameDefault+0x1e20) [0x55d9e86f4250]
>    14  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyEval_EvalCodeWithName+0x2f9) [0x55d9e86384f9]
>    15  /home/nfs/mrocklin/miniconda/envs/ucx-dev/bin/python(_PyFunction_FastCallDict+0x1d5) [0x55d9e86395d5]
> ```
> 
> cc @Akshay-Venkatesh

@mrocklin Somehow missed this yesterday. Are you already running with `UCX_MEMTYPE_CACHE=n` set?

mrocklin:
> @mrocklin Somehow missed this yesterday. Are you already running with UCX_MEMTYPE_CACHE=n set?

Yes

mrocklin:
> @mrocklin This disables the use of nvlink, TCP, and shared memory. Just FYI.

Hrm, I'm not seeing the code that this comment refers to.  Is there a line in particular I should be looking at?

rjzamora:
Probably the `os.environ.setdefault("UCX_TLS", "rc,cuda_copy")`?

mrocklin:
What should it be?

rjzamora:
In other places the suggestion was: `UCX_TLS=rc,sm,tcp,cuda_copy` . - Is this case different?

Akshay-Venkatesh:
Sorry, Matt. Richard is right. Until we have a fix for `invalid device context`, we'll have to use the env vars that Richard has posted.

distributed/protocol/utils.py
mrocklin: Is there a test that triggers this?  Or, more directly, when does this come up?  cc @TomAugspurger

TomAugspurger:
I think https://github.com/dask/distributed/issues/2565#issuecomment-473412977 has the motivation, but I don’t recall details. Not sure about tests.

mrocklin:
> I think #2565 (comment) has the motivation, but I don’t recall details. Not sure about tests.

OK, for now I've turned off compression on all cuda data.  That should stop us from splitting up large frames.  It looks like currently there is a limit in ucx-py that keeps us under 2**31 bytes.  At first this seems to be limited by the `int` type of the `Message._length` attribute.  Changing that to long causes a segfault, so I'm probably missing something upstream in UCX.  

I've also added logic to not call ensure_bytes and `b''.join` if there is only one element in that list.

distributed/core.py
```diff
@@ -484,7 +487,9 @@ def handle_stream(self, comm, extra=None, every_cycle=[]):
                 pdb.set_trace()
             raise
         finally:
-            comm.close()  # TODO: why do we need this now?
+            yield comm.close()  # TODO: why do we need this now?
+            #             ^ Good question :) comm.close can be a
+            # coroutine, in which case this isn't doing anything.
```
mrocklin: Not true actually, tornado coroutines are automatically run in the background, even if they are not awaited/yielded.

mrocklin:
This now works-ish (at least when combined with some of the work in https://github.com/rapidsai/dask-cuda/pull/46).  I would like to get this to a point where we could merge it somewhat quickly.  

I don't mind things being a little rough if they are well isolated into files that aren't in the mainline code path (files like ucx.py, cuda.py, cudf.py and so on).  

There are a few TODO's left in the actual logic that I suspect are left by @TomAugspurger .  Is this something that you can look into this week Tom to see if they are still necessary?

distributed/comm/tests/test_ucx.py
```diff
@@ -0,0 +1,313 @@
...
+from distributed.utils_test import gen_test, loop, inc  # noqa: 401
+
+from .test_comms import check_deserialize
+
+
+HOST = ucp.get_address()
+
+# Currently having some issues with re-using ports.
+# Tests just hang. Still debugging.
+port_counter = itertools.count(13337)
```
TomAugspurger: I'm not sure, but this *may* be fixed by https://github.com/Akshay-Venkatesh/ucx-py/issues/79. If so we could remove this and just provide the host without the port, and let ucx provide a port for us.

distributed/comm/tests/test_ucx.py
```diff
@@ -0,0 +1,313 @@
...
+async def get_comm_pair(listen_addr, listen_args=None, connect_args=None, **kwargs):
+    q = asyncio.queues.Queue()
+
+    async def handle_comm(comm):
+        await q.put(comm)
+
+    # Workaround for hanging test in
+    # pytest distributed/comm/tests/test_ucx.py::test_comm_objs -vs --count=2
+    # on the second time through.
+    ucp._libs.ucp_py.reader_added = 0
```
TomAugspurger: @Akshay-Venkatesh do you recall if this was resolved? Is it the same as https://github.com/Akshay-Venkatesh/ucx-py/issues/69, or different?

distributed/comm/tests/test_ucx.py
```diff
@@ -0,0 +1,313 @@
...
+    # TODO:
+    # 1. ensure exceptions in handle_comm fail the test
+    # 2. Use dict in read / write, put seralization there.
+    # 3. Test peer_address
+    # 4. Test cleanup
+    async def f():
+        address = "ucx://{}:{}".format(HOST, next(port_counter))
+
+        async def handle_comm(comm):
+            # XXX: failures here don't fail the build yet
```
TomAugspurger: I think by "build" I meant "test run". It'd be good to check if this is still true by adding an `assert 0` in this `handle_comm`.

distributed/comm/tests/test_ucx.py
```diff
@@ -0,0 +1,313 @@
...
+
+    asyncio.run(f())
+
+
+@pytest.mark.asyncio
+async def test_ping_pong_data():
+    np = pytest.importorskip("numpy")
+
+    data = np.ones((10, 10))
+    # TODO: broken for large arrays
```
TomAugspurger: Maybe can remove now that compression is disabled?

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
+"""
+:ref:`UCX`_ based communications for distributed.
+
+See :ref:`communcations` for more.
```
TomAugspurger: typo: communications.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+from .core import Comm, Connector, Listener, CommClosedError
+from .registry import Backend, backends
+from .utils import ensure_concrete_host, to_frames, from_frames
+from ..utils import ensure_ip, get_ip, get_ipv6, nbytes
+
+import ucp
+
+import os
+
+os.environ.setdefault("UCX_RNDV_SCHEME", "put_zcopy")
```
TomAugspurger: I haven't followed this discussion closely. Do we plan to allow users to set this in the dask config file?

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+os.environ.setdefault("UCX_MEMTYPE_CACHE", "n")
+os.environ.setdefault("UCX_TLS", "rc,cuda_copy")
+
+logger = logging.getLogger(__name__)
+MAX_MSG_LOG = 23
+PORT = 13337
+
+# set in ~/.dask/config.yaml
+# or DASK_DISTRIBUTED__COMM__UCXADDRESS
+_PORT_COUNTER = itertools.count(PORT)
```
TomAugspurger: Can maybe be removed. I'm not sure whether or not the test suite covers the behavior I was seeing that required this.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+
+_INITIALIZED = False
+
+
+def _ucp_init():
```
TomAugspurger: It occurs to me now that this behavior of the global `_INITIALIZED` to know whether init has been called probably belongs in `ucp.init`.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+    global _INITIALIZED
+
+    if not _INITIALIZED:
+        ucp.init()
+        _INITIALIZED = True
+
+
+# ----------------------------------------------------------------------------
+# Addressing
+# TODO: Parts of these should probably be moved to `comm/addressing.py`
```
TomAugspurger: I'm not sure what this comment means anymore.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+    """
+    Parse an endpoint address given in the form "host:port".
+
+    >>> _parse_host_port("10.33.225.160:13337")
+    ("10.33.225.160", 13337)
+    """
+    if address.startswith("ucx://"):
+        _, address = _parse_address(address)
+
+    # if default port is None we select the next port availabe
```
TomAugspurger: This comment may be out of date.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+    default_port = default_port or random.randint(1024, 65000)
+    return parse_host_port(address, default_port=default_port)
+
+
+def _unparse_host_port(host, port=None):
+    return unparse_host_port(host, port)
+
+
+def get_endpoint_address(endpoint):
+    # TODO: ucx-py: 18
```
TomAugspurger: This is https://github.com/Akshay-Venkatesh/ucx-py/issues/61 I think, which is still open.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+
+        # finalizer?
+
+    @property
+    def local_address(self) -> str:
+        return self._local_addr
+
+    @property
+    def peer_address(self) -> str:
+        # XXX: This isn't quite for the server (from UCXListener).
```
TomAugspurger: I think this is also referring to https://github.com/Akshay-Venkatesh/ucx-py/issues/61, but I may be wrong

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+
+        return msg
+
+    def abort(self):
+        # breakpoint()
+        if self._ep:
+            ucp.destroy_ep(self._ep)
+            logger.debug("Destroyed UCX endpoint")
+            self._ep = None
+        # if self.listener_instance:
```
TomAugspurger: Not sure why this is commented out, or why it was maybe there earlier.

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+    def closed(self):
+        return self._ep is None
+
+
+class UCXConnector(Connector):
+    prefix = "ucx://"
+    comm_class = UCX
+    encrypted = False
+
+    client = ...  # TODO: add a client here?
```
TomAugspurger: I'm not sure what this is referring to (probably just copying TCPConnector)

distributed/comm/ucx.py
```diff
@@ -0,0 +1,366 @@
...
+            address = "ucx://" + address
+        self.address = address
+        self.ip, self.port = _parse_host_port(address, default_port=0)
+        self.comm_handler = comm_handler
+        self.deserialize = deserialize
+        self._ep = None  # type: ucp.Endpoint
+        self.listener_instance = None  # type: ucp.ListenerFuture
+        self._task = None
+
+        # XXX: The init may be required to take args like
```
TomAugspurger: I think this was a comment about the abc `Listener.__init__`, but I'm not sure anymore.

distributed/core.py
```diff
@@ -467,6 +467,9 @@ def handle_stream(self, comm, extra=None, every_cycle=[]):
                             if op == "close-stream":
                                 closed = True
                                 break
+                            # XXX: getting a KeyError here. Our
```
TomAugspurger: Not sure what this is referring to.

distributed/comm/tests/test_ucx.py
Akshay-Venkatesh: @TomAugspurger There was a fix missing for the test but the test is passing.

distributed/comm/ucx.py
mrocklin: @Akshay-Venkatesh has this happened?  is `ucp.init()` idempotent?

mrocklin: Ah, looks like yes: 9d6b99587b4d748f89256d1a9338afb463e27ea4

distributed/comm/tests/test_ucx.py
mrocklin: Done

mrocklin:
I've removed the WIP label.  Review appreciated.  I think that this is safe to go in.

distributed/comm/tests/test_ucx.py
Akshay-Venkatesh: @TomAugspurger I tested this yesterday and the issue hasn't been resolved. This still has to be fixed.

mrocklin:
OK, I'm merging this tomorrow if there are no further comments.

TomAugspurger:
I won't have time soon to look this over again, but based on my last review
I also think this is safe to go in.

TomAugspurger:
🎉

mrocklin:
Thank you @TomAugspurger, @quasiben, and @Akshay-Venkatesh for working on this.  I'm sure that there is still plenty more to do here, but it will be nice to have this in master.

distributed/protocol/core.py
```diff
@@ -176,6 +176,7 @@ def loads_msgpack(header, payload):
     See Also:
         dumps_msgpack
     """
+    header = bytes(header)
     if header:
         header = msgpack.loads(header, use_list=False, **msgpack_opts)
```
jakirkham, line 181: Do you recall why this was needed? Was this due to the `if header` line? Did `msgpack.dumps` need this? Or was it due to something else like potentially unusual types being passed in for `header` (like maybe a NumPy array)?

jakirkham: Ah I guess this is explained here ( https://github.com/dask/distributed/commit/44c1d5cc6a867dcaa055518b6c7615863f498ebd ).

PR #2601, opened 2019-04-09
Improved description accuracy of --memory-limit option.

javabrett:
Kicking-off some changes to try and make the memory limit options more accurate.  The 60% factor has not been around since ca6beed95b82618386c7d566099b3b82429d9b56 and there are other inaccuracies.

Changes likely imperfect, feedback welcome.

### DIFF
--- a/docs/source/worker.rst
+++ b/docs/source/worker.rst
@@ -70,7 +70,7 @@ thread per process::
    $ dask-worker scheduler:8786 --nprocs 8 --nthreads 1
 
 This will launch 8 worker processes each of which has its own
-ThreadPoolExecutor of size 1. 
+ThreadPoolExecutor of size 1.
 
 If your computations are external to Python and long-running and don't release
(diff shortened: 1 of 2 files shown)


docs/source/worker.rst
```diff
@@ -98,7 +98,9 @@ are the available options::
                             cores
      --nprocs INTEGER       Number of worker processes to launch.  Defaults to one.
      --name TEXT            Alias
-     --memory-limit TEXT    Number of bytes (per worker process) before spilling data to disk
+     --memory-limit TEXT    Maximum bytes of memory that this worker should use.
+                            Use None for unlimited, or 'auto' for
```
mrocklin:
```suggestion
                            Use 0 for unlimited, or 'auto' for
```

We can't pass `None` through the CLI.  `0` seems to work though.

mrocklin: I've gone ahead and pushed this.

PR #2603, opened 2019-04-10
Add resources to scheduler update_graph plugin

nicolls1:
I would like to know the resources a task is requesting in my scheduler plugin in the update_graph callback. Could it be added as so?

The docs just reference **kwargs so doesn't seem they need to be updated: https://distributed.dask.org/en/latest/plugins.html


mrocklin:
Seems reasonable.  Is there anything else that should be there as well?

nicolls1:
For what I am currently doing, no. Scanned the function and nothing stood out to me also.

mrocklin:
OK, merging in.  Thanks @nicolls1 !

PR #2604, opened 2019-04-10
RabbitMQ scheduler plugin

nicolls1:
A RabbitMQ plugin example for tracking tasks. I wanted to share if there is interest. Was mentioned in #2111 and open for suggestions.


mrocklin:
That's quite neat.  Thanks @nicolls1 !

I'll leave this up for a short while to see if there are any comments, but will be happy to merge afterwards.

PR #2606, opened 2019-04-10
Check self.dependencies when looking at dependent tasks in memory

deepthirajagopalan7:
This is an extension to https://github.com/dask/distributed/pull/2196 which fixes https://github.com/dask/distributed/issues/2187

As explained in https://github.com/dask/distributed/pull/2196,  only the scheduler would know the dependencies if the inputs were futures, and therefore the fix was to check both the input dependencies and self.dependencies.

This PR extends the same logic to handle dependencies of futures which are futures themselves and would be known only to the scheduler.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1425,7 +1425,11 @@ class Scheduler(ServerNode):
                 except KeyError:
                     deps = self.dependencies[key]
                 for dep in deps:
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks for the contribution @deepthirajagopalan7 .  This looks good to me.  Merging it in.

PR #2607, opened 2019-04-11
Add basic health endpoints to scheduler and worker bokeh.

amerkel2:
This adds a basic `/health` endpoint as used by many microservices/frameworks.

For now, this is just a basic check whether the service is up. This is intended as a starting point, more sophisticated checks whether the components are really healthy could be added on top of this later.

### DIFF
--- a/distributed/bokeh/scheduler_html.py
+++ b/distributed/bokeh/scheduler_html.py
@@ -210,6 +210,12 @@ class PrometheusHandler(RequestHandler):
         self.set_header('Content-Type', 'text/plain; version=0.0.4')
 
 
+class HealthHandler(RequestHandler):
+    def get(self):
+        self.write('ok')
+        self.set_header('Content-Type', 'text/plain')
+
+
 routes = [
         (r'info/main/workers.html', Workers),
         (r'info/worker/(.*).html', Worker),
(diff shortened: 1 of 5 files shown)


mrocklin:
Interesting.  I wasn't aware that this was a common convention (I know relatively little about the microservices world).

cc @quasiben and @beberg for visibility in case they have thoughts.

amerkel2:
The convention comes from the Apache Mesos and Mesos frameworks (Aurora, Marathon), cf. 
http://aurora.apache.org/documentation/0.8.0/user-guide/#http-health-checking-and-graceful-shutdown
https://mesosphere.github.io/marathon/docs/health-checks.html
http://mesos.apache.org/documentation/latest/health-checks/

quasiben:
`/health` or `/healthz` are common endpoints for health checks/liveness probes.  This could be extended to check if the scheduler/worker process are actually up and running -- but as you say, a good starting point and something for other systems to hook into

mrocklin:
OK great.  Merging in.  Thanks for the contribution @amerkel2 and the review @quasiben !

Also @amerkel2 I notice that this is your first code contribution to this repository.  Welcome!

mrocklin:
There is also a release happening later today, so this should be out relatively soon.

amerkel2:
Glad to hear that, thanks!

PR #2610, opened 2019-04-11
Add exception details to the exception to help debug (#2549)

### LINKED ISSUE #2549: KilledWorker exception should include worker name
If we don't include the worker name alongside the 'tcp://' address in the exception object, it's hard to get that info otherwise (that I can tell) since the worker has been removed from the scheduler.

Will you accept a PR to add that to the KilledWorker exception? Also, any reason that KilledWorker takes args (instead of kwargs), as in:

```
KilledWorker(k, address)
```

Can we make kwargs to identify these arguments?

cc @plbertrand @wyegelwel

plbertrand:


distributed/core.py
```diff
@@ -885,7 +885,7 @@ def clean_exception(exception, traceback, **kwargs):
     --------
     error_message: create and serialize errors into message
     """
-    if isinstance(exception, bytes):
+    if isinstance(exception, bytes) or isinstance(exception, bytearray):
```
mrocklin: I'm curious, why this change?

distributed/scheduler.py
```diff
@@ -1750,7 +1750,14 @@ def remove_worker(self, comm=None, address=None, safe=False, close=True):
...
                         del recommendations[k]
-                        e = pickle.dumps(KilledWorker(k, address))
+                        e = pickle.dumps(
+                            KilledWorker(task_key=k,
+                                         worker_address=address,
+                                         worker_name=ws.name,
+                                         worker_host=ws.host,
+                                         worker_local_directory=ws.local_directory,
+                                         worker_metrics=ws.metrics,
+                                         worker_pid=ws.pid))
```
mrocklin: So, this is a little odd because a `KilledWorker` error isn't specifically associated to one worker, but rather to a task that has killed a set of workers.  I guess though that it can be helpful to include information about one representative of that set.

I wonder if, rather than include each of these keywords individually (which locks us into a particularly rigid structure going forward) we might just include the `ws` object itself.  That way in the future every time one of these attributes changes you're not on the hook to fix things here as well.

distributed/scheduler.py
```diff
@@ -4505,4 +4512,7 @@ def heartbeat_interval(n):
 
 
 class KilledWorker(Exception):
-    pass
+    def __init__(self, *args, **kwargs):
+        super(KilledWorker, self).__init__(*args)
+        self.args = args
+        self.kwargs = kwargs
```
mrocklin: If we do as above and include just a single keyword `last_failed_worker` or something then it would be good to include that explicitly, rather than catch everything.

distributed/scheduler.py
plbertrand: This is what I did at first but it seems to fail to pickle on Python 2.7 (https://travis-ci.org/dask/distributed/jobs/518810169#L2679). I would prefer having ws itself.

I did think about "which locks us into a particularly rigid structure going forward" and to go around it was to pass them as kwargs on the exception. The downside is after an upgrade, if someone uses one of the variable that is removed, it will throw an exception. I'm not sure what is ideal here.

distributed/core.py
plbertrand: I couldn't nailed down exactly where it came from but it seems that when an exception has a lot of information, it's passed in two frames. When it is small enough, it is represented as bytes otherwise I believe the network code will concatenate those bytes into a bytearray. Because of that reason, the exception returned was unpickled.

distributed/scheduler.py
mrocklin: > I would prefer having ws itself.

I recommend that we figure out why `ws` isn't serializable and possible resolve that, if you have time.

plbertrand: I'm unable to reproduce the CI that you have. Is there a good starter point that I could look into to run the Python 2.7 tests?

mrocklin: I'm not sure I entirely understand.  I would clone the repository, which you've done, make a conda environment with python 2.7, install things, and then run `py.test distributed`

https://distributed.dask.org/en/latest/develop.html

From what you've said before it sounds like you've already gotten things to fail on 2.7, so maybe I'm not understanding your question correctly.

plbertrand: It fails in your continuous integration. I've been trying really hard to reproduce the environment properly here and had some very mixed results on the jankiest setup you can imagine. I was able to get the breakpoint to work only when another breakpoint somewhere else in the code was causing a delay long enough to cause the race condition for me to get the state in the scheduler. Now it stopped working without any explanation while I was doing a divide and conquer on worker state fields. The short list that is left to test is ws.metrics, ws.services and ws.processing. Perhaps I'll push until my divide and conquer is done but it's very painful. The procedure in the doc also doesn't work as is. For example tornado gets installed at version 6 even though it does not support python 2.7. It's also missing the pytest-timeout package among other things. Are you able to run the environment easily?

mrocklin: > I've been trying really hard to reproduce the environment properly here and had some very mixed results on the jankiest setup you can imagine. 

Heh, sorry to hear that you've been having a frustrating time.  Lets see if we can get past that.

First, please note that we have some intermittent failures (they've been quite annoying lately) so if the error looks entirely unrelated to what you're working on, then please don't stress about it.


> It fails in your continuous integration

Do you have a link to the failure?

> Now it stopped working without any explanation while I was doing a divide and conquer on worker state fields

In your situation I would use a debugger like pdb and look at the state that failed.


> The procedure in the doc also doesn't work as is. For example tornado gets installed at version 6 even though it does not support python 2.7

I'm quite surprised that you were able to install tornado 6 in a Python 2 environment.  That either sounds like a terrible bug in tornado's packaging (please report upstream if so) or possibly that you're building an environment in an odd way.

plbertrand:
@mrocklin Thanks a lot! Having the -1 did the trick for me, what did you stumble upon that prompted the clean() method on the WorkerState?

mrocklin:
I wanted to remove references to TaskState objects, which link to each other to refer to the entire graph.  Otherwise you would end up downloading way more state than you wanted.

mrocklin:
This is in.  Thanks @plbertrand !

PR #2612, opened 2019-04-14
versions.py: added dask_xgboost to optional_packages

javabrett:


distributed/versions.py
```diff
@@ -24,6 +24,7 @@
                      ('bokeh', lambda p: p.__version__),
                      ('lz4', lambda p: p.__version__),
                      ('dask_ml', lambda p: p.__version__),
+                     ('dask_xgboost', lambda p: p.__version__),
```
jcrist: `dask_xgboost` is part of dask-ml (while also being its own package). I'm not saying it shouldn't go here, but without some curation this list may also grow unbounded.

Curently you can do:

```python
client.get_versions(packages=['dask_xgboost'])
```

PR #2613, opened 2019-04-15
added missing whitespace to start_worker cmd

condoratberlin:
A whitespace was missing in front of the optional "nprocs" option. Aditionally changed the following optional options to a more readable way, where the separating whitespace is always in front of the additional command.


jcrist:
LGTM. Thanks for the PR!

PR #2614, opened 2019-04-15
Apply Black to standardize code styling

mrocklin:
This uses the [black](https://github.com/ambv/black) code styler to manage our style.  This PR includes two commits:

1.  Add a pre-commit hook to make this automatic, should people use `pre-commit install` at the base of this repository
2.  Apply `black` to the entire codebase.

The only concern I have about this is that it screws with `git blame`.  Future maintainers will have to probably come back to this commit, and then git blame from `HEAD^`


mrocklin:
Merging shortly if there are no comments. This has been up in an issue on dask/dask for a while.

.travis.yml
```diff
@@ -32,6 +32,7 @@ install:
 script:
   - if [[ $TESTS == true ]]; then source continuous_integration/travis/run_tests.sh ; fi
   - if [[ $LINT == true ]]; then pip install flake8 ; flake8 distributed ; fi
+  - if [[ $LINT == true ]]; then pip install black; black distributed ; fi
```
jcrist: This will format the files and only returns a non-zero exit code if there's an internal error. You probably want `--check` instead to check that they pass.

PR #2619, opened 2019-04-17
Catch RuntimeError to avoid serialization fail when using pytorch

muammar:
After updating one of my systems I realized that there was a problem in `distributed/protocol/torch.py` that made serialization to fail. In this PR, the problem is fixed by adding a try/except statement to fix the problem.

### DIFF
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -13,6 +13,7 @@ def test_tensor():
     t2 = deserialize(header, frames)
     assert (x == t2.numpy()).all()
 
+
 @pytest.mark.parametrize("requires_grad", [True, False])
 def test_grad(requires_grad):
     x = np.arange(10)
@@ -23,12 +24,13 @@ def test_grad(requires_grad):
 
     t2 = deserialize(*serialize(t))
 
-    assert (t2.requires_grad is requires_grad)
-    assert (t.requires_grad is requires_grad)
+    assert t2.requires_grad is requires_grad
+    assert t.requires_grad is requires_grad
     assert np.allclose(t2.detach().numpy(), x)
 
     if requires_grad:
         assert np.allclose(t2.grad.numpy(), 1)
+
 
 def test_resnet():
     torchvision = pytest.importorskip("torchvision")
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -14,15 +14,17 @@ def test_tensor():
     assert (x == t2.numpy()).all()
 
 
-def test_grad():
+@pytest.mark.parametrize("requires_grad", [True, False])
+def test_grad(requires_grad):
     x = np.arange(10)
-    t = torch.Tensor(x)
+    t = torch.tensor(x, dtype=torch.float, requires_grad=requires_grad)
     t.grad = torch.zeros_like(t) + 1
 
     t2 = deserialize(*serialize(t))
-    assert (t2.numpy() == x).all()
-    assert (t2.grad.numpy() == 1).all()
 
+    assert (t2.detach().numpy() == x).all()
+    assert (t2.grad.numpy() == 1).all()
+    assert (t2.requires_grad is requires_grad)
 
 def test_resnet():
     torchvision = pytest.importorskip("torchvision")
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -13,18 +13,22 @@ def test_tensor():
     t2 = deserialize(header, frames)
     assert (x == t2.numpy()).all()
 
-
 @pytest.mark.parametrize("requires_grad", [True, False])
 def test_grad(requires_grad):
     x = np.arange(10)
     t = torch.tensor(x, dtype=torch.float, requires_grad=requires_grad)
-    t.grad = torch.zeros_like(t) + 1
+
+    if requires_grad:
+        t.grad = torch.zeros_like(t) + 1
 
     t2 = deserialize(*serialize(t))
 
-    assert (t2.detach().numpy() == x).all()
-    assert (t2.grad.numpy() == 1).all()
     assert (t2.requires_grad is requires_grad)
+    assert (t.requires_grad is requires_grad)
+    assert np.allclose(t2.detach().numpy(), x)
+
+    if requires_grad:
+        assert np.allclose(t2.grad.numpy(), 1)
 
 def test_resnet():
     torchvision = pytest.importorskip("torchvision")
--- a/distributed/protocol/torch.py
+++ b/distributed/protocol/torch.py
@@ -7,7 +7,12 @@ import numpy as np
 @dask_serialize.register(torch.Tensor)
 def serialize_torch_Tensor(t):
     requires_grad_ = t.requires_grad
-    header, frames = serialize(t.detach_().numpy())
+
+    if requires_grad_:
+        header, frames = serialize(t.detach().numpy())
+    else:
+        header, frames = serialize(t.numpy())
+
     if t.grad is not None:
         grad_header, grad_frames = serialize(t.grad.numpy())
         header["grad"] = {"header": grad_header, "start": len(frames)}


mrocklin:
I'm curious, what caused the problem?  Is it worth adding a small test?

muammar:
> I'm curious, what caused the problem? Is it worth adding a small test?

When a torch tensor requires grad, then one can not use `.numpy()` to convert it to a numpy array. Instead, one has to use `.detach_()`. In our example, the gradient data is set to some arbitrary value but the tensor does not have the `.requires_grad` attribute set to true.  

Proof:

I undo the fix I submitted and just added this change:

```
diff --git a/distributed/protocol/tests/test_torch.py b/distributed/protocol/tests/test_torch.py
index 6cc8bb20..1e2190e8 100644
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -17,6 +17,7 @@ def test_tensor():
 def test_grad():
     x = np.arange(10)
     t = torch.Tensor(x)
+    t.requires_grad = True
     t.grad = torch.zeros_like(t) + 1
 
     t2 = deserialize(*serialize(t))

```

Running the test for the protocols fails with: 

```
distributed/protocol/tests/test_torch.py::test_grad FAILED                                                                                                                                                  [ 98%]
distributed/protocol/tests/test_torch.py::test_resnet PASSED                                                                                                                                                [ 99%]
distributed/protocol/tests/test_torch.py::test_deserialize_grad PASSED                                                                                                                                      [100%]

==================================================================================================== FAILURES =====================================================================================================
____________________________________________________________________________________________________ test_grad ____________________________________________________________________________________________________

    def test_grad():
        x = np.arange(10)
        t = torch.Tensor(x)
        t.requires_grad = True
        t.grad = torch.zeros_like(t) + 1
    
        t2 = deserialize(*serialize(t))
>       assert (t2.numpy() == x).all()
E       RuntimeError: Can't call numpy() on Variable that requires grad. Use var.detach().numpy() instead.

distributed/protocol/tests/test_torch.py:24: RuntimeError

```

Applying the change proposed in this PR makes everything work:

```
distributed/protocol/tests/test_sklearn.py::test_basic PASSED                                                                                                                                               [ 97%]
distributed/protocol/tests/test_torch.py::test_tensor PASSED                                                                                                                                                [ 97%]
distributed/protocol/tests/test_torch.py::test_grad PASSED                                                                                                                                                  [ 98%]
distributed/protocol/tests/test_torch.py::test_resnet PASSED                                                                                                                                                [ 99%]
distributed/protocol/tests/test_torch.py::test_deserialize_grad PASSED     
```

Does it make sense?

mrocklin:
cc @stsievert if you have time to review

distributed/protocol/torch.py
```diff
@@ -7,7 +7,12 @@
 @dask_serialize.register(torch.Tensor)
 def serialize_torch_Tensor(t):
     requires_grad_ = t.requires_grad
-    header, frames = serialize(t.detach_().numpy())
+
+    try:
+        header, frames = serialize(t.detach_().numpy())
+    except RuntimeError:
+        header, frames = serialize(t.numpy())
```
stsievert: Why not an use an if-statement instead? It's known when the error will be thrown: when `t.requires_grad` is true.

distributed/protocol/torch.py
```diff
@@ -7,7 +7,12 @@
 @dask_serialize.register(torch.Tensor)
 def serialize_torch_Tensor(t):
     requires_grad_ = t.requires_grad
-    header, frames = serialize(t.detach_().numpy())
+
+    try:
+        header, frames = serialize(t.detach_().numpy())
```
stsievert: I don't think we want to use `detach_` here because it modifies `t` in place:

``` python
>>> import torch
>>> x = torch.rand(4, requires_grad=True)
>>> y = x.detach_().numpy()
>>> x.requires_grad
False
```

A test should be added to make sure that `x.requires_grad` stays the same before/after serialization.

distributed/protocol/tests/test_torch.py
```diff
@@ -18,9 +18,10 @@ def test_grad():
     x = np.arange(10)
     t = torch.Tensor(x)
     t.grad = torch.zeros_like(t) + 1
+    t.requires_grad = True
```
stsievert: It'd be nice if this test could have both values of `requires_grad`. Maybe `pytest.mark.parametrize`?

distributed/protocol/torch.py
muammar: I changed this block to use an if-statement.

distributed/protocol/torch.py
muammar: I originally placed a `.detach()` but after some discussions, we decided to add `.detach_()` instead. The reason you said, and it made sense for me, was that `.detach_() ` creates a leaf and does not break the graph.  What we want is just to extract the data from the tensor object, and if it requires or not gradients, to be de/serialized correctly. This operation shouldn't perturb what the model expects from these tensors. Are we sure we have to get back to a plain `.detach()`. 

This would be the commit to address what's been reviewed here:

```
diff --git a/distributed/protocol/tests/test_torch.py b/distributed/protocol/tests/test_torch.py
index e6747e32..fa181a0d 100644
--- a/distributed/protocol/tests/test_torch.py
+++ b/distributed/protocol/tests/test_torch.py
@@ -21,8 +21,17 @@ def test_grad():
     t.requires_grad = True
 
     t2 = deserialize(*serialize(t))
-    assert (t2.detach_().numpy() == x).all()
+
+    assert (t2.detach().numpy() == x).all()
     assert (t2.grad.numpy() == 1).all()
+    assert (t2.requires_grad is True)
+
+    t.requires_grad = False
+    t3 = deserialize(*serialize(t))
+
+    assert (t3.detach().numpy() == x).all()
+    assert (t3.grad.numpy() == 1).all()
+    assert (t3.requires_grad is False)
 
 
 def test_resnet():
diff --git a/distributed/protocol/torch.py b/distributed/protocol/torch.py
index e4185468..a70e60b6 100644
--- a/distributed/protocol/torch.py
+++ b/distributed/protocol/torch.py
@@ -8,9 +8,9 @@ import numpy as np
 def serialize_torch_Tensor(t):
     requires_grad_ = t.requires_grad
 
-    try:
+    if requires_grad_:
         header, frames = serialize(t.detach_().numpy())
-    except RuntimeError:
+    else:
         header, frames = serialize(t.numpy())
 
     if t.grad is not None:

```

Thoughts :)?

distributed/protocol/tests/test_torch.py
muammar: I don't know how to do that :S I can investigate though.

stsievert: Something like

``` python
@pytest.mark.parametrize("requires_grad", [True, False])
def test_grad(requires_grad):
    x = torch.zeros(..., requires_grad=requires_grad)
```

distributed/protocol/torch.py
stsievert: > .detach_() creates a leaf and does not break the graph.

I think this is true for `detach()` too. I think https://github.com/dask/distributed/pull/2586#issuecomment-478116967 is concerned with correctly sending the `requires_grad` information.

Serialization should not implicitly modify the inputs.

distributed/protocol/tests/test_torch.py
```diff
@@ -18,10 +18,20 @@ def test_grad():
...
     t2 = deserialize(*serialize(t))
-    assert (t2.numpy() == x).all()
+
+    assert (t2.detach().numpy() == x).all()
     assert (t2.grad.numpy() == 1).all()
+    assert (t2.requires_grad is True)
+
+    t.requires_grad = False
+    t3 = deserialize(*serialize(t))
+
```
stsievert: I would test that serialization and de-serialization does not modify `t.requires_grad`. Adding this line below the serialization would test that:

``` python
assert t.requires_grad == requires_grad
```

distributed/protocol/torch.py
muammar: I changed it back to `detach()`.

distributed/protocol/tests/test_torch.py
muammar: Very cool. I did not know about this. It is the first time I am using pytest. I have done something similar.

distributed/protocol/tests/test_torch.py
muammar: Done, too.

muammar:
I updated the PR following all suggestions by @stsievert :). I also unified all changes in one single commit. Please, let me know if some other changes are needed. Thanks.

distributed/protocol/tests/test_torch.py
```diff
@@ -14,15 +14,17 @@ def test_tensor():
     assert (x == t2.numpy()).all()
 
 
-def test_grad():
+@pytest.mark.parametrize("requires_grad", [True, False])
+def test_grad(requires_grad):
     x = np.arange(10)
-    t = torch.Tensor(x)
+    t = torch.tensor(x, dtype=torch.float, requires_grad=requires_grad)
     t.grad = torch.zeros_like(t) + 1
```
stsievert: Suggestion:

``` diff
-    t.grad = torch.zeros_like(t) + 1
+    if requires_grad:
+        t.grad = torch.zeros_like(t) + 1
```

distributed/protocol/tests/test_torch.py
```diff
@@ -14,15 +14,17 @@ def test_tensor():
...
 
     t2 = deserialize(*serialize(t))
-    assert (t2.numpy() == x).all()
-    assert (t2.grad.numpy() == 1).all()
 
+    assert (t2.detach().numpy() == x).all()
+    assert (t2.grad.numpy() == 1).all()
+    assert (t2.requires_grad is requires_grad)
```
stsievert: I'd add a test to make sure that `t.requires_grad` isn't modified by serialization:

``` diff
+     assert t.requires_grad is requires_grad
```

stsievert: And maybe use `np.allclose` instead of `==`?

quasiben:
rerunning tests

mrocklin:
Checking in.  What's the status here?  @muammar are you blocked on anything?  It looks like @stsievert has made a few recommendations above.

distributed/protocol/tests/test_torch.py
muammar: This is done now.

muammar:
> Checking in. What's the status here? @muammar are you blocked on anything? It looks like @stsievert has made a few recommendations above.

Thanks for checking-in :). I am sorry I was not able to put this together before. The latest commit should (hopefully) comply with all requests by @stsievert.

stsievert:
@muammar my concerns are resolved, especially the biggest one: now, the tests make sure the `requires_grad` flag of the input array is unmodified.

Looks like these changes are complaining about linting issues and missing a run through Black. (the other issues are unrelated and in `test_client.py::test_profile_bokeh` and `test_local.py::test_ipywidgets`).

muammar:
> @muammar my concerns are resolved, especially the biggest one: now, the tests make sure the `requires_grad` flag of the input array is unmodified.
> 
> Looks like these changes are complaining about linting issues and missing a run through Black. (the other issues are unrelated and in `test_client.py::test_profile_bokeh` and `test_local.py::test_ipywidgets`).

Is there anything from my side I could do to help fix this?

stsievert:
> Is there anything from my side I could do to help fix this?

Definitely with the linting issue. Dask uses [black] and [flake8] for style, so I'd run

``` shell
$ pip install black flake8
$ cd /path/to/distributed
$ black .
$ flake8 .
```

The docs at https://docs.dask.org/en/latest/develop.html#style could be improved.

[black]:https://github.com/python/black 
[flake8]:https://github.com/python/black

muammar:
> > Is there anything from my side I could do to help fix this?
> 
> Definitely with the linting issue. Dask uses [black](https://github.com/python/black) and [flake8](https://github.com/python/black) for style, so I'd run
> 
> ```shell
> $ pip install black flake8
> $ cd /path/to/distributed
> $ black .
> $ flake8 .
> ```
> 
> The docs at https://docs.dask.org/en/latest/develop.html#style could be improved.

Oook! I did not understand you were referring to code style. I can do that and probably submit a different PR for that purpose (?).

stsievert:
> Oook! I did not understand you were referring to code style.

Thanks!

> I can do that and probably submit a different PR for that purpose (?).

I've put in #2680 for this.

muammar:
> > Oook! I did not understand you were referring to code style.
> 
> Thanks!
> 
> > I can do that and probably submit a different PR for that purpose (?).
> 
> I've put in #2680 for this.

I have done it now. By the way, I had no idea of the existence of black. Very interesting. I will use it on my projects.

mrocklin:
The windows failure is unrelated (handling in #2684)  

Thanks for the contribution @muammar and thanks for the review @stsievert .

Merging.

PR #2622, opened 2019-04-18
Add worker task processing on if present

nicolls1:
I would like to have worker info when a task transitions to processing in my scheduler plugin.
Currently the processing transition doesn't have any additional info:
> transition: key=test-a0abb189-4c26-41df-bdcd-6176ace3b534, start=waiting, finish=processing, *args=(), **kwargs={}

I want to track which nodes are occupied and have current tasks running for tracking/auto scaling. I have long running tasks that happen very infrequently and requires beefy cloud GPU machines and can bring them up only when required.

Assuming this is an idea that is willing to be entertained, I looked at possibilities and this is the one that stood out to me. It is consistent with the kwargs that is passed when a task is finished.

At the moment, not seeing other info that would be useful to have

Happy to do it another way also, thought this would be the best way to start a conversation. Thanks in advance!


distributed/scheduler.py
```diff
@@ -3965,6 +3965,8 @@ def transition(self, key, finish, *args, **kwargs):
                     self.tasks[ts.key] = ts
                 for plugin in list(self.plugins):
                     try:
+                        if self.tasks[key].processing_on is not None:
+                            kwargs['worker'] = self.tasks[key].processing_on.address
```
mrocklin: Thanks @nicolls1 !

Perhaps instead we might just place in the entire task state object, and let the plugin do what it wants with it.  Thoughts?

mrocklin: To be clear, I'm suggesting something like `kwargs['task_state'] = ts`

nicolls1: Thanks for the quick response. That sounds much cleaner and more future proof, will do!

distributed/scheduler.py
```diff
@@ -3965,9 +3965,7 @@ def transition(self, key, finish, *args, **kwargs):
                     self.tasks[ts.key] = ts
                 for plugin in list(self.plugins):
                     try:
-                        if self.tasks[key].processing_on is not None:
-                            kwargs['worker'] = self.tasks[key].processing_on.address
-                        plugin.transition(key, start, finish2, *args, **kwargs)
+                        plugin.transition(key, start, finish2, task_state=ts, *args, **kwargs)
```
mrocklin: I suspect that Python will yell at you here for placing `*args` after a named keyword argument.  You'll probably have to insert this into kwargs yourself.

nicolls1: Good point, wasn't giving errors but looking at it now args was always empty so probably why. Sorry, github didn't auto refresh the second post.

distributed/scheduler.py
```diff
@@ -3965,6 +3965,7 @@ def transition(self, key, finish, *args, **kwargs):
                     self.tasks[ts.key] = ts
                 for plugin in list(self.plugins):
                     try:
+                        kwargs['task_state'] = ts
```
mrocklin: Can we move this outside of the for loop?

mrocklin: (Sorry for the nitpick)

mrocklin: Gentle ping

nicolls1: Of course! thanks for the feedback, sorry been away over holiday weekend.

nicolls1:
It seems the test failures are unrelated, no? Happy to make updates to them if I broke something.

mrocklin:
It looks like some of the failures are genuine:

```python
    @gen_cluster(client=True)
    def test_eventstream_remote(c, s, a, b):
        base_plugins = len(s.plugins)
        comm = yield eventstream(s.address, interval=0.010)
    
        start = time()
        while len(s.plugins) == base_plugins:
            yield gen.sleep(0.01)
            assert time() < start + 5
    
        futures = c.map(div, [1] * 10, range(10))
    
        start = time()
        total = []
        while len(total) < 10:
            msgs = yield comm.read()
>           assert isinstance(msgs, tuple)
E           assert False
E            +  where False = isinstance({'exception': TypeError("can not serialize 'TaskState' object"), 'status': 'uncaught-error', 'text': "can not serialize 'TaskState' object", 'traceback': <traceback object at 0x7fe417971d08>}, tuple)
```

It looks like some existing plugins that we rely on send all of the information we give to them across a wire (I suspect that this one is for the progress bars).  It looks like sending `TaskState` objects is difficult, probably because they include some non-serializable object.

Looking at the state for TaskState objects it becomes clear that they include references to pretty much the entire graph, so probably the solution is to not send these things around.  If EventStream is the only problem, then you can probably fix that here: 

https://github.com/dask/distributed/blob/a3d2016a4fab9ef14ab5be0b0b722b59365b59ce/distributed/diagnostics/eventstream.py#L24-L28

We definitely don't want to include all of kwargs.  Probably we just want to exclude the `ts` keyword.

mrocklin:
Also, regarding linting, we started using the `black` code formatter.  You might want to run the following:

```
pip install black
black distributed
git commit -a -m "black linting"
```

nicolls1:
Okay so seems like I need to get the tests running locally to do this effectively. Got dask to test with only 5 errors but distributed is failing on test_simple and pretty much all and is running very slowly. Hopefully will have time to look into later this week

mrocklin:
You can also run 

    py.test distributed -k test_eventstream_remote

nicolls1:
When looking through the other plugins I found this:
https://github.com/dask/distributed/blob/master/distributed/stealing.py#L77

Just going to use that! Added a bit to the docs if you think it will be useful otherwise can be closed.

docs/source/plugins.rst
```diff
@@ -49,3 +49,26 @@ for more information on RabbitMQ and how to consume the messages.
        scheduler.add_plugin(plugin)
 
 Run with: ``dask-scheduler --preload <filename.py>``
+
+Accessing Full Task State
+-------------------------
+
+If you would like to access the full ``TaskState`` stored in the scheduler you
```
TomAugspurger: For ease of reference

```suggestion
If you would like to access the full :ref:`distributed.scheduler.TaskTate` stored in the scheduler you
```

TomAugspurger: For ease of linking

```suggestion
If you would like to access the full :ref:`distributed.scheduler.TaskTate` stored in the scheduler you
```

FYI @nicolls1, I was going to push this directly, but I don't want to mess up your master branch :) Typically you'd want to checkout a new branch from the latest master before starting on a PR.

nicolls1: Good to know thanks! Sorry, was traveling and didn't have a chance to do it myself. Thanks for changing!

TomAugspurger:
Thanks @nicolls1!

PR #2625, opened 2019-04-19
Add worker_class argument to LocalCluster

pentschev:
Allows instantiating workers subclassed from Worker (e.g., CUDAWorker) in LocalCluster.

@mrocklin

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -206,6 +206,8 @@ class LocalCluster(Cluster):
         if security:
             self.worker_kwargs["security"] = security
 
+        if not worker_class:
+            worker_class = Worker if not processes else Nanny
         self.worker_class = worker_class
 
         self.start(ip=ip, n_workers=n_workers)
@@ -284,13 +286,9 @@ class LocalCluster(Cluster):
             return
 
         if self.processes:
-            W = Nanny
-            kwargs["worker_class"] = self.worker_class
             kwargs["quiet"] = True
-        else:
-            W = self.worker_class or Worker
 
-        w = yield W(
+        w = yield self.worker_class(
             self.scheduler.address,
             loop=self.loop,
             death_timeout=death_timeout,
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -74,6 +74,8 @@ class LocalCluster(Cluster):
         Protocol to use like ``tcp://``, ``tls://``, ``inproc://``
         This defaults to sensible choice given other keyword arguments like
         ``processes`` and ``security``
+    worker_class: Worker
+        Worker class used to instantiate workers from.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


pentschev:
It looks like the failures are not related to this PR, but I'm not quite sure about it.

distributed/deploy/local.py
```diff
@@ -203,6 +206,8 @@ def __init__(
         if security:
             self.worker_kwargs["security"] = security
 
+        self.worker_class = worker_class
```
mrocklin:
```suggestion
        if not worker_class:
            if processes:
                worker_class = Nanny
            else:
                worker_class = Worker
        self.worker_class = worker_class
```

distributed/deploy/local.py
```diff
@@ -280,9 +285,10 @@ def _start_worker(self, death_timeout=60, **kwargs):
 
         if self.processes:
             W = Nanny
+            kwargs["worker_class"] = self.worker_class
             kwargs["quiet"] = True
         else:
-            W = Worker
+            W = self.worker_class or Worker
 
         w = yield W(
```
mrocklin:
```suggestion
        w = yield self.worker_class(
```

distributed/deploy/local.py
```diff
@@ -280,9 +285,10 @@ def _start_worker(self, death_timeout=60, **kwargs):
 
         if self.processes:
             W = Nanny
+            kwargs["worker_class"] = self.worker_class
             kwargs["quiet"] = True
         else:
-            W = Worker
+            W = self.worker_class or Worker
```
mrocklin: If you accept the other changes, then we can probably drop the conditional above (it's effectively now handled in the constructor)

We might also remove the `self.processes` attribute, which seems to be effectively replaced and generalized by `self.worker_class`.

We should have something like the following though to maintain previous behavior.

```python
if self.worker_class is Nanny:
    kwargs['quiet'] = True
```

pentschev: Sure, your changes look fine to me. I will push a commit.

mrocklin:
This could also use a test.  Maybe something like ...

```python
def test_worker_class(loop):
    class MyWorker(Worker):
        pass

    with LocalCluster(n_workers=2, loop=loop, worker_class=MyWorker, scheduler_port=0, dashboard_address=None) as cluster:
        assert all(isinstance(w, MyWorker) for w in cluster.workers)
```

pentschev:
Thanks for the suggestion, I was really unsure how to test this. I extended your test a bit to test both `Worker` and `Nanny`.

mrocklin:
I should have mentioned that we recently switched to using `black` for code styling.  It looks like you found that out.  I need to add docs for this in a more prominent place, but you may want to try installing [pre commit hooks](https://ml.dask.org/contributing.html#pre-commit-hooks)

pentschev:
> I should have mentioned that we recently switched to using `black` for code styling. It looks like you found that out. I need to add docs for this in a more prominent place, but you may want to try installing [pre commit hooks](https://ml.dask.org/contributing.html#pre-commit-hooks)

Ah nice, didn't know about this, thanks for that!

I think the errors are not related to this PR.

pentschev:
No worries, let me know if you need help with them.

Issue #2618, opened 2019-04-17
Add interface= keyword to LocalCluster


mrocklin:
This will help when starting up LocalCluster on a machine with exotic hardware, like infiniband.

PR #2629, opened 2019-04-19
Add interface= keyword to LocalCluster

mrocklin:
This is useful when you want to use a particular network interface, like
infiniband.

Fixes https://github.com/dask/distributed/issues/2618


mrocklin:
cc @quasiben

PR #2633, opened 2019-04-22
Limit test_spill_by_default memory, reenable it

pentschev:
Judging by the skip reason, I think this is a reasonable approach to ensure the test works, yet doesn't cause memory issues.

@mrocklin any objections?


quasiben:
ping @mrocklin

mrocklin:
Thanks @pentschev !

PR #2635, opened 2019-04-29
Updated logging module doc links from docs/2 to docs/3

javabrett:
Should probably gradually retire more Python 2 things, including doc-links.


martindurant:
Thank you, this is useful.

PR #2637, opened 2019-04-29
Fix deserialization of bytes chunks larger than 64MB

pentschev:
When chunks of `bytes` are larger than 64MB they get split into multiple 64MB frames within the serializer. When `_deserialize_bytes()` returns only the first frame, it truncates the object and it's unable to deserialize it. The solution is simply to join all the frames.

### DIFF
--- a/distributed/protocol/tests/test_serialize.py
+++ b/distributed/protocol/tests/test_serialize.py
@@ -192,7 +192,7 @@ def test_empty_loads_deep():
 
 
 def test_serialize_bytes():
-    for x in [1, "abc", np.arange(5)]:
+    for x in [1, "abc", np.arange(5), b"ab" * int(100e6)]:
         b = serialize_bytes(x)
         assert isinstance(b, bytes)
         y = deserialize_bytes(b)
--- a/distributed/protocol/tests/test_serialize.py
+++ b/distributed/protocol/tests/test_serialize.py
@@ -192,7 +192,7 @@ def test_empty_loads_deep():
 
 
 def test_serialize_bytes():
-    for x in [1, "abc", np.arange(5), b"ab" * int(100e6)]:
+    for x in [1, "abc", np.arange(5), b"ab" * int(40e6)]:
         b = serialize_bytes(x)
         assert isinstance(b, bytes)
         y = deserialize_bytes(b)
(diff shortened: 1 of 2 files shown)


pentschev:
I'm still writing a test for this, please hold on to merging until then.

distributed/protocol/tests/test_serialize.py
```diff
@@ -192,7 +192,7 @@ def test_empty_loads_deep():
 
 
 def test_serialize_bytes():
-    for x in [1, "abc", np.arange(5)]:
+    for x in [1, "abc", np.arange(5), b"ab" * int(100e6)]:
```
mrocklin:
```suggestion
    for x in [1, "abc", np.arange(5), b"ab" * int(40e6)]:
```

Bringing this down to 80MB still causes the test to fail without the fix, and also reduces the testing time here from about 2s to 1s.

mrocklin:
Thanks for the fix @pentschev .  I have a small suggestion, but am +1 regardless.  You should be able to press the "commit suggestion" button.

distributed/protocol/tests/test_serialize.py
mrocklin: (My apologies if shaving off 1s sounds trivial, but the test suite has thousands of these tests, so we try to be minimal when easy)

pentschev: Yes, I intentionally had it at 200MB to make sure we had more than 2 frames, but no objections in making it smaller, since it will still cover the issue we're fixing.

pentschev: > (My apologies if shaving off 1s sounds trivial, but the test suite has thousands of these tests, so we try to be minimal when easy)

Yes, I know, I also tend to reduce because of that, but I forgot to check how long it was for this one in particular.

pentschev:
@mrocklin the failure doesn't seem to be related to this PR.

mrocklin:
I actually don't see a failure at all.

Regardless, merging.

pentschev:
Thanks @mrocklin !

mrocklin:
Thanks for finding and resolving the bug @pentschev

PR #2639, opened 2019-04-29
Add timeout to Client._reconnect

### LINKED ISSUE #2582: Client reconnect should have a timeout
Currently if a scheduler goes down, but the client isn't closed, the client will keep trying to connect to the scheduler address forever. This should probably have a timeout - maybe the same timeout as the initial connect (`distributed.timeouts.connect`)?

Relevant code:

https://github.com/dask/distributed/blob/b36e87d14e75c0cbdd0e3292ac65db2a836f74bd/distributed/client.py#L871-L876

jcrist:
Previously if a client lost connection to the scheduler, it would try to
reconnect forever. We now use the same timeout as the initial connect.
On failure a nice message is logged and the client is shutdown.

Fixes #2582.

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -3711,6 +3711,7 @@ def test_reconnect_timeout(c, s):
         start = time()
         while c.status != "closed":
             yield c._update_scheduler_info()
+            yield gen.sleep(0.05)
             assert time() < start + 5, "Timeout waiting for reconnect to fail"
     text = logger.getvalue()
     assert "Failed to reconnect" in text
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -3704,6 +3704,18 @@ def test_reconnect(loop):
     c.close()
 
 
+@gen_cluster(client=True, ncores=[], client_kwargs={'timeout': 0.5})
+def test_reconnect_timeout(c, s):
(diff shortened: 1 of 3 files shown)


jcrist:
Failure appears to be unrelated.

distributed/tests/test_client.py
```diff
@@ -3704,6 +3704,18 @@ def test_reconnect(loop):
     c.close()
 
 
+@gen_cluster(client=True, ncores=[], client_kwargs={"timeout": 0.5})
+def test_reconnect_timeout(c, s):
+    with captured_logger(logging.getLogger("distributed.client")) as logger:
+        yield s.close()
+        start = time()
+        while c.status != "closed":
+            yield c._update_scheduler_info()
```
mrocklin: Maybe add a small sleep in here just so we don't overwhelm things (though probably it would be fine either way)

```suggestion
            yield c._update_scheduler_info()
            yield gen.sleep(0.05)
```

distributed/utils_test.py
```diff
@@ -974,7 +974,7 @@ def coro():
                                 if s.validate:
                                     s.validate_state()
                             finally:
-                                if client:
+                                if client and c.status != "closed":
```
mrocklin: Probably worth checking, but there might also be a `closing` status. 
```suggestion
                                if client and not c.status.startswith("clos"):
```

PR #2654, opened 2019-05-04
Use 'temporary-directory' from dask.config for Worker's directory

mrocklin:


jakirkham:
Doing the same thing for `Nanny` in PR ( https://github.com/dask/distributed/pull/3531 ).

PR #2655, opened 2019-05-05
Fixed comment regarding keeping existing level if less verbose

javabrett:
This behaviour changed in commit 6cc529979 .


mrocklin:
Thanks @javabrett

PR #2658, opened 2019-05-05
Learn bandwidth over time

mrocklin:
In order to schedule tasks intelligently we need to know how long
communications will take.  To do this, we need to estimate the bandwidth
of the network.  This can vary by orders of magnitude depending on
hardwware.

Previously we asked the user to specify this in configuration.

Now we learn it over time.  Each worker keeps an exponentially weighted
moving average for all of its data communications.  It sends this
information to the scheduler as part of the heartbeats (which include
lots of other diagnostic information).  The scheduler updates its own
measurement accordingly.

Eventually, I may want to make our model of bandwidth more complex, 
to include type and network information.  I thought I'd start here though.


mrocklin:
@quasiben you might enjoy reviewing this

distributed/worker.py
```diff
@@ -1842,6 +1846,7 @@ def gather_dep(self, worker, dep, deps, total_nbytes, cause=None):
 
                 total_bytes = sum(self.nbytes.get(dep, 0) for dep in response["data"])
                 duration = (stop - start) or 0.5
+                bandwidth = total_bytes / (duration - abs(self.scheduler_delay))
```
StephanErb: There is a non-zero chance this results in a division by zero.

I am wondering if we could simply leave out the `abs(self.scheduler_delay))` part?

* the bandwidth is an approximation anyway
* the code would be more robust
* it would be consistent with the metrics computed below

distributed/scheduler.py
```diff
@@ -1336,6 +1337,8 @@ def heartbeat_worker(
         host_info = host_info or {}
 
         self.host_info[host]["last-seen"] = local_now
+        frac = 1 / 20 / len(self.workers)
```
StephanErb: This is a bit over my head: Why do we need to divide by the number of workers?

On a large cluster with little activity, I fear that this could lead to very small convergence.

distributed/worker.py
mrocklin: Thanks for the comment @StephanErb .  What you say makes sense to me.  I've removed the scheduler_delay change.

distributed/scheduler.py
mrocklin: Each worker sends a heartbeat independently.  So the more workers there are, the more heartbeats there are.  This delay time means (I think) that our bandwidth value changes with a time scale of around 20 heartbeats.  This is probably about a minute of time, so should be fairly fast on the scheduler side.  The worker's heartbeat also has a time scale though, and it only changes when there are actual communications, so that might move more slowly.

mrocklin:
Thanks for the review @StephanErb .  I suspect that we will need to modify this in the future after we have more experience with it.  Please raise an issue if you notice something odd.

Issue #2568, opened 2019-03-18
CPU Usage Child Processes


CMCDragonkai:
I've noticed that when using `subprocess.run` on my dask tasks, the workers still report low CPU usage, when the child process is in fact maxing out my CPU usage. Does Dask track the forked/child process CPU usage under the worker CPU usage? If not, I think it should since child processes can occur when interfacing dask workers against programs written in non-python.

CMCDragonkai:
The task stream isn't able to keep track of the child process either, so it just looks like the task is blocked and there's no progress report.

mrocklin:
> Does Dask track the forked/child process CPU usage under the worker CPU usage? 

No, it doesn't

> If not, I think it should since child processes can occur when interfacing dask workers against programs written in non-python.

Do you have recommendations on how to do this well?  Currently we use the `psutil` module.  If there is something else we should be doing then suggestions would be welcome.

The relevant code is here:

https://github.com/dask/distributed/blob/fb30c33562862f30864456766424b44a3e91aa5b/distributed/system_monitor.py#L50

CMCDragonkai:
It looks like you would need to do something where you find the pid tree of the worker process and then acquire stats for each pid. This is transitive as child processes that produce more child processes would all come under the pid tree.

CMCDragonkai:
Here's an example of doing this in Node.js: https://github.com/soyuka/pidusage-tree

CMCDragonkai:
This is natively supported in the `psutil` package: https://unix.stackexchange.com/a/339071/56970

```
pid=2235; python3 -c "import psutil
for c in psutil.Process($pid).children(True):
  print(c.pid)"
```

The `True` ensures that it gets it recursively.

https://psutil.readthedocs.io/en/latest/#psutil.Process.children

Another way is to realise that process group on Linux will always encapsulate the entire process tree when launched by a shell. See: https://en.wikipedia.org/wiki/Process_group This means that on Linux, if you start each worker on its own process group. Then you can just query the PGID and get all processes part of that group. This is also useful for distribution of signals. However this may not be portable to Windows or other operating systems, whereas process tree via direct descendants may be more standard among all OS.

mrocklin:
Perhaps you'd like to submit a PR?  Also, will this affect performance?  We
do these checks relatively frequently, and they show up frequently in
overhead profiling.

CMCDragonkai:
[...]

PR #2662, opened 2019-05-06
Monitor worker children processes 

kforti:
add proc.children() memory info to proc memory info. #2568 

This pull request is an effort to gather more profiling data on all processes running on a worker node. I am not sure where the cpu monitoring takes place, and if this is also only monitoring cpu usage by the worker process and not the children processes.

p.s. this is my first ever pull request. Apologies for any obvious mistakes.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2393,7 +2393,7 @@ class Worker(ServerNode):
         total = 0
 
         proc = self.monitor.proc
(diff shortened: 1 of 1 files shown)


mrocklin:
Possible repeat of https://github.com/dask/distributed/pull/2390/ ?  It might be worth checking out that PR and checking in with @danpf on his thoughts.

PR #2663, opened 2019-05-06
Add waiting tasks to progress title bar

jrbourbeau:
It was brought to my attention that the progress bar examples in http://distributed.dask.org/en/latest/web.html#progress have the number of tasks waiting for dependencies in the title bar, but when using the distributed `master` branch, this doesn't seem to be the case. For instance, running the [example computation](http://distributed.dask.org/en/latest/web.html#example-computation) in the Web Interface docs gives

Screen Shot 2019-05-06 at 3 07 44 PM

As best I can tell, the number of waiting tasks was removed from the title bar in #1729. Not sure if this was removed intentionally and the gifs in the docs are just outdated, or if it was an unintentional change. As the number of waiting tasks can be useful when charting the overall progress of a computation, this PR adds waiting tasks back to the `TaskProgress` title text.

### DIFF
--- a/distributed/bokeh/scheduler.py
+++ b/distributed/bokeh/scheduler.py
@@ -1049,7 +1049,7 @@ class TaskProgress(DashboardComponent):
     def update(self):
         with log_errors():
             state = {"all": valmap(len, self.plugin.all), "nbytes": self.plugin.nbytes}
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @jrbourbeau !

PR #2664, opened 2019-05-06
DOC: Clean up reference to cluster object

michaelaye:
The current docs would overwrite the cluster objects, so it needs to get its own variable name.
As the Client is usually named `c`, I changed the cluster object to the name `cluster`

Solves #4779

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -82,19 +82,19 @@ class LocalCluster(Cluster):
 
     Examples
     --------
-    >>> c = LocalCluster()  # Create a local cluster with as many workers as cores  # doctest: +SKIP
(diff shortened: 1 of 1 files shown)


martindurant:
Fixes #4779  ( <- will automatically close your issue upon merger)

michaelaye:
Darn, I always confuse `solve` with `resolve` which also auto-closes.. ;)

jrbourbeau:
Fixes https://github.com/dask/dask/issues/4779

martindurant:
Hah, I got it wrong too :)

martindurant:
I believe the failure here is due to https://github.com/dask/dask/pull/4780

PR #2665, opened 2019-05-07
Cap total memory by the hard limit of the maximum resident memory

twoertwein:
Some systems enforce memory limits with ulimit (SLURM seems to enforce it this way, at least it sets the appropriate ulimit). `psutil.virtual_memory().total` does not take this limit into consideration. This patch makes sure that `TOTAL_MEMORY` is capped by the allowed maximal memory.

If you think it makes sense to add this feature, I can work on a test case for it.

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -25,7 +25,7 @@ from distributed.core import rpc
 from distributed.client import wait
 from distributed.scheduler import Scheduler
 from distributed.metrics import time
-from distributed.worker import Worker, error_message, logger
+from distributed.worker import Worker, error_message, logger, parse_memory_limit
 from distributed.utils import tmpfile, format_bytes
 from distributed.utils_test import (
     inc,
@@ -1401,3 +1401,25 @@ def test_host_address(c, s):
     assert "127.0.0.3" in n.address
     assert "127.0.0.3" in n.worker_address
     yield n.close()
+
+
+def test_resource_limit():
+    assert parse_memory_limit("250MiB", 1, total_cores=1) == 1024 * 1024 * 250
+
+    # get current limit
+    resource = pytest.importorskip("resource")
+    try:
+        hard_limit = resource.getrlimit(resource.RLIMIT_RSS)[1]
+    except OSError:
+        pytest.skip("resource could not get the RSS limit")
+    memory_limit = psutil.virtual_memory().total
+    if hard_limit > memory_limit or hard_limit < 0:
+        hard_limit = memory_limit
+
+    # decrease memory limit by one byte
+    new_limit = hard_limit - 1
+    try:
+        resource.setrlimit(resource.RLIMIT_RSS, (new_limit, new_limit))
+        assert parse_memory_limit(hard_limit, 1, total_cores=1) == new_limit
+    except OSError:
+        pytest.skip("resource could not set the RSS limit")
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -12,6 +12,7 @@ import threading
 import sys
 import warnings
 import weakref
+import psutil
 
 import dask
 from dask.core import istask
@@ -76,15 +77,7 @@ LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
 
 no_value = "--no-value-sentinel--"
 
-try:
-    import psutil
-
-    TOTAL_MEMORY = psutil.virtual_memory().total
-except ImportError:
-    logger.warning("Please install psutil to estimate worker memory use")
-    TOTAL_MEMORY = 8e9
-    psutil = None
-
+TOTAL_MEMORY = psutil.virtual_memory().total
 
 IN_PLAY = ("waiting", "ready", "executing", "long-running")
 PENDING = ("waiting", "ready", "constrained")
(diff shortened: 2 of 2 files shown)


distributed/worker.py
```diff
@@ -82,6 +83,14 @@
     logger.warning("Please install psutil to estimate worker memory use")
     TOTAL_MEMORY = 8e9
     psutil = None
+try:
+    import resource
+    # minimum of physical available memory and the hard process limit
+    hard_limit = resource.getrlimit(resource.RLIMIT_RSS)[1]
+    if hard_limit > 0:
+        TOTAL_MEMORY = min(TOTAL_MEMORY, hard_limit)
+except OSError:
```
mrocklin: When does this occur?

Also, should I be concerned about this behavior on OSX?  Presumably this would be handled by your use of `min`

```python
In [1]: import resource

In [2]: resource.getrlimit(resource.RLIMIT_RSS)[1]
Out[2]: 9223372036854775807
```

We also now strictly require psutil, so if you wanted to drop the`try-except` around the psutil import while you're here, that would be a welcome change :)

distributed/worker.py
```diff
@@ -74,6 +74,7 @@
 
 no_value = "--no-value-sentinel--"
 
+# determine maximal amout of memory
```
mrocklin:
```suggestion
# determine maximum amount of memory
```

distributed/worker.py
twoertwein: Thanks for your feedback!

["An OSError is raised on syscall failure."](https://docs.python.org/3/library/resource.html) I assume that could occur on Windows/exotic OSs?

Maybe OSX uses this high number as a proxy for unlimited - what does `ulimit -m` return for you? On my linux machine it returns -1 for unlimited.

Ok, I will remove try try-catch around psutil.

twoertwein:
I'm not exactly sure whether the memory of a child process (when you spawn workers) counts also towards the parent process's limit. If that is not the case, this patch could artificially restrict the useable memory: assume a limit of 5GB per process - you might be able to have 10 workers which have their own 5GB - but this patch might force them to only have 500MB.

mrocklin:
> I'm not exactly sure whether the memory of a child process (when you spawn workers) counts also towards the parent process's limit. If that is not the case, this patch could artificially restrict the useable memory: assume a limit of 5GB per process - you might be able to have 10 workers which have their own 5GB - but this patch might force them to only have 500MB.

Wouldn't the call to `resource.getrlimit` still return 5GB though?  Isn't that the only limit that we're applying in this case?

twoertwein:
Yes, each worker/process should see their own(?) 5GB as a limit when calling `resource.getrlimit`.

But based on how `TOTAL_MEMORY` is used here https://github.com/dask/distributed/blob/ce24d8d4ed626ebb569a88925263479bae104ecf/distributed/worker.py#L2932 it feels to me that `memory_limit='auto'`  could result in too tight memory limits as it is divided by the number of 'ideal' workers? It probably makes more sense to cap the returned limit of `parse_memory_limit` instead of directly capping `TOTAL_MEMORY`?

mrocklin:
Ah yes, that's a good point.  In that case we are looking for the total memory of the full machine, just as you say, so that we can estimate how much we should take up based on the number of threads we've been asked to use.  

Maybe we should apply the resource limit a bit further down in the call stack

twoertwein:
I made it part of `parse_memory_limit`

twoertwein:
>     2. There are style linting issues though.  I recommend running black.  `pip install black` and then `black distributed/` from the root of the repository
ok, I will fix them soon.

>     3. Are there easy ways for us to test this?
I could imaging a test case which first spawns a new process, then sets a low RSS limit, creates a worker/calls `parse_memory_limit`, compares the used/returned memory limit, and then kills the process. I think we need to create a sub-process, otherwise the new RSS limit might affect other tests. I will look into that on the weekend.

twoertwein:
the test in AppVeyor seems not have the `resource` module! I can put it in a try-except - import within `parse_memory_limit` with try-except? That would also fix the shadowing of a few `resource` loop variables.

mrocklin:
> the test in AppVeyor seems not have the resource module! I can put it in a try-except - import within parse_memory_limit with try-except? That would also fix the shadowing of a few resource loop variables.

Hrm, interesting.  I thought it was part of the standard library.  Using a try-except makes sense to me.  Also, beware, there are a couple of intermittent failures in our tests currently.  My apologies for the background noise.

twoertwein:
I pushed a test case. Since the test case includes a `except ImportError` I don't know whether the test is actually conducted. The test case works on my linux machine.

distributed/tests/test_worker.py
```diff
@@ -1402,3 +1402,25 @@ def test_host_address(c, s):
...
+            limit = 1024 * 1024 * 200
+            resource.setrlimit(resource.RLIMIT_RSS, (limit, limit))
+            if parse_memory_limit("250MiB", 1, total_cores=1) != limit:
+                os._exit(1)
+        except (ImportError, OSError):
+            pass
+        os._exit(0)
+    else:
+        # wait for child
+        assert os.waitpid(child, 0)[1] == 0
```
mrocklin: I have never used this API.  It was interesting to see it work.  Seems pretty slick to me.  

Our window CI currently has some issues.  I'm trying to fix it up today.  Hopefully I'll be able to rerun tests afterwards to make sure that things here pass well there.  If not then we'll have to skip some behavior based on OS.  I'll update here when that's done.

twoertwein: I also used it for the first time and was surprised how simple it is :)

Is it possible from within a test function to mark it as skipped? If `resource` is not found or an `OSError` occurs, it might be nice to signal that the test had to be skipped.

mrocklin: Yes, you can decorate the function with 

```python
@pytest.mark.skipif(sys.platform.startswith('win'), reason="Fork not available")
def test_foo():
```

Or you can call `pytest.skip("a reason")` from within the function. 

Judging by the windows tests (which work again) `os.fork` isn't available so we'll need to skip this.

Maybe

```python
@pytest.mark.skipif(not hasattr(os, 'fork'), reason="Fork not available")
def test_foo():
```

twoertwein: thank you! Lets see whether it works now.

It would probably be better to have a test case which also works on windows! I assume that would either mean to open a new python instance with subprocess or to simply change the RSS limit for the current process. I think it would be safe to decrease the limit to maximal memory -1 (if no limit is set) or current limit -1.

twoertwein:
I re-wrote the test to decrease the limit of the current process by one byte

mrocklin:
This seems good to me.  Merging.  

Thanks @twoertwein !  

PR #2671, opened 2019-05-08
Remove support for Iterators and Queues

mrocklin:
These add non-trivial code complexity, and don't seem to be commonly
used (based on bug reports and SO questions).  They're also a bit odd on
our tests (there are some lingering threads as a result.

This commit removes functionality for them and replaces them with
informative warnings pointing people towards normal for loops.


martindurant:
Since there is no deprecation cycle here, only the informative error, would it be worthwhile keeping the doc page for a little while, with some simple text saying this is no longer available? I can understand a "no" here, because perhaps it only adds confusion.

docs/source/queues.rst
```diff
@@ -1,194 +1,10 @@
...
-   <iframe width="560"
-           height="315"
-           src="https://www.youtube.com/embed/4IzuoV6XW_4?rel=0"
-           frameborder="0"
-           allowfullscreen>
-   </iframe>
+1.  Use normal for loops with Client.submit/gather and as_completed
+2.  Use :doc:`asynchronous async/await <asynchronous>` code and a few coroutines
+3.  Try out the `Streamz <https://streamz.readthedocs.io>`_ project,
+    which has Dask support
```
mrocklin: @martindurant

martindurant: ah, ok  - sorry

mrocklin:
Merging this tomorrow if there are no further comments

distributed/client.py
```diff
@@ -1486,20 +1466,10 @@ def map(self, func, *iterables, **kwargs):
         if all(map(isqueue, iterables)) or all(
             isinstance(i, Iterator) for i in iterables
         ):
```
dhirschfeld: docstring might need updating above?
```
iterables: Iterables, Iterators, or Queues
```

mrocklin: Thanks for the catch @dhirschfeld !

djhoese:
As mentioned on gitter, this feature was used by some people including me in the Satpy library. In the future, don't consider a lack of github issues or SO questions to mean people aren't using a feature. For me it worked perfectly which is why I didn't need to ask for help. Luckily the workaround for newer versions of dask in my exact use case is pretty simple.

I'm still seeing mentions of queue as a valid argument a couple places in the documentation. I'll make an issue for that...

mrocklin:
Thanks David!

Sorry to disrupt things on your end.  We *could* add this back in, but for
technical reasons I'd still prefer to keep it out of the codebase if you
can find other solutions.

H4dr1en:
Queue was a nice feature, I would be glad to see it back as well :)
Maybe it could be provided as a separated functional package (dask-queue)?

mrocklin:
If folks want to make a third party package they are more than welcome.

People interested in this functionality might also want to look at `as_completed` or Dask Queues

https://docs.dask.org/en/latest/futures.html

PR #2672, opened 2019-05-08
Add release procedure doc

jrbourbeau:
Just points to the `dask/dask` release procedure, but still nice to have


jrbourbeau:
Should have used `[skip ci]`...canceling builds

mrocklin:
Thanks @jrbourbeau !

PR #2673, opened 2019-05-08
Add WeakSet _instances attributes to all classes

mrocklin:
These help us track leaking workers, schedulers, and clients


mrocklin:
cc @jcrist, you were asking for something like this.  It's not much, but it's a start.

PR #2675, opened 2019-05-09
Add SpecificationCluster

mrocklin:
This is intended to be a base for LocalCluster (and others) that want to
specify more heterogeneous information about workers.

Additionally, this PR does the following:

1.  Starts the use of Python 3 only code in the main codebase
2.  Cleans up a number of our intermittent testing failures (we had nannies that survived test cleanup before, and they were sending random messages to ports that were screwing up other tests)
3.  Adds a couple of new small failures, notably silent shutdown is no longer entirely silent (working on it)

Docstring
----------

Cluster that requires a full specification of workers

This attempts to handle much of the logistics of cleanly setting up and
tearing down a scheduler and workers, without handling any of the logic
around user inputs.  It should form the base of other cluster creation
functions.

Examples
--------
```python
>>> spec = {
...     'my-worker': {"cls": Worker, "options": {"ncores": 1}},
...     'my-nanny': {"cls": Nanny, "options": {"ncores": 2}},
... }
>>> cluster = SpecCluster(workers=spec)
```


mrocklin:
I'm going to try this out with some heterogeneous GPU machines.  This feels  like a nice base on which to rewrite and cleanup LocalCluster though, a prospect for which I am excited :)

distributed/deploy/spec.py
```diff
@@ -0,0 +1,198 @@
...
+        A similar mapping for a scheduler
+    asynchronous: bool
+        If this is intended to be used directly within an event loop with
+        async/await
+
+    Examples
+    --------
+    >>> spec = {
+    ...     'my-worker': {"cls": Worker, "options": {"ncores": 1}},
+    ...     'my-nanny': {"cls": Nanny, "options": {"ncores": 2}},
```
dhirschfeld: Sorry - this is probably a dumb question, but why would you want to allocate 2 cores to the nanny process but only 1 to the worker? I thought the nanny didn't do any real work and hence didn't need to be allocated significant resources?

dhirschfeld:
Is the spec intended to be per-worker - e.g.:
```python
spec = {
    'worker1': {"cls": Worker, "options": {"ncores": 1}},
    'nanny1': {"cls": Nanny, "options": {"ncores": 2}},
    'worker2': {"cls": Worker, "options": {"ncores": 1}},
    'nanny2': {"cls": Nanny, "options": {"ncores": 2}},
    'worker3': {"cls": Worker, "options": {"ncores": 1}},
    'nanny3': {"cls": Nanny, "options": {"ncores": 2}},
    ...
}
```

dhirschfeld:
I'm just wondering if now is a good time to introduce the concept of "worker pools":
https://github.com/dask/distributed/issues/2208#issuecomment-418919769

e.g. you would pass `pools` in addition to `workers` and have the `workers` dict reference the specs defined in `pools`
```python
>>> pool_specs = {
...     'default': {
...         'worker': {"cls": Worker, "options": {"ncores": 1}},
...         'nanny': {"cls": Nanny, "options": {"ncores": 2}},
...     },
...     'no-nanny': {
...         'worker': {"cls": Worker, "options": {"ncores": 1}},
...     },
... }

>>> worker_specs = {'worker1': 'default', 'worker2': 'no-nanny'}
>>> cluster = SpecCluster(workers=worker_specs, pools=pool_specs)
```

distributed/deploy/spec.py
mrocklin: Nanny is an object that starts a worker that is watched by a nanny process.  The keywords pass down to the worker.

mrocklin:
> I'm just wondering if now is a good time to introduce the concept of "worker pools":

This is related to that issue, but is lower level.  I think that it would enable other people to add things like pools more easily.  If this is something that you'd like to explore I encourage you to do so now.  I agree that now would be a good time to explore this to help guide design.

distributed/deploy/spec.py
dhirschfeld: Thanks for the clarification! So the ncores are then shared between the nanny and the worker so it would be ok to only allocate 1 core to a `Nanny` so you could have as many workers as CPUs on the box?

mrocklin: The ncores keyword specifies how many threads to add to the thread pool executor of a worker.  This keyword is passed through the nanny to the worker.  This is unrelated to this PR and is just how Nanny works.  You might find it interesting to look through the nanny.py file.

mrocklin:
OK, this PR could use review.  cc'ing the various cluster management people @jcrist @jhamman @guillaumeeb @jacobtomlinson .  In particular, I'd like to ask you all to read through the docstring of the SpecCluster class in this PR, available here:

https://github.com/dask/distributed/pull/2675/files#diff-e95e1ccb9336311eae957db482e0be4d

This PR both refactors LocalCluster to use this superclass, which I think is a useful separation, and a good step on a path towards a number of issues that have been discussed, *and* it also fixes several intermittent failures that have been affecting our CI for a long time.  As a result only about half of this PR is relevant for most people to review.  

However, this PR does also slightly break adaptivity (which I'm happy to fix afterwards) and I suspect will force some code changes downstream in dask-jobqueue, dask-kubernetes, and maybe dask-yarn.  Again, I'm happy to help with these changes, but I'd like to collect thoughts on this approach before moving forward with this.

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+
+    We can also ``.scale(...)`` the cluster, which adds new workers of a given
+    form.
+
+    >>> worker = {'cls': Worker, 'options': {}}
+    >>> cluster = SpecCluster(scheduler=scheduler, worker=worker)
+    >>> cluster.worker_spec
+    {}
+
+    >>> cluster.scale(3)
```
dhirschfeld: How does it know which spec to use for scaling? Above you create a cluster with both a worker and a nanny spec - what would scale do in that case?

mrocklin: You misunderstand the meaning of the workers_spec.  That's a full definition of all worker, not of a type of worker.

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+        self.status = "starting"
+        self.scheduler = await self.scheduler
+        self.status = "running"
+
+    def _correct_state(self):
+        if self._correct_state_waiting:
+            # If people call this frequently, we only want to run it once
+            return self._correct_state_waiting
+        else:
+            task = asyncio.Task(self._correct_state_internal())
```
jcrist: You shouldn't create `Task`s manually, but instead use `asyncio.ensure_future`.

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+        async with self._lock:
+            self._correct_state_waiting = None
+
+            pre = list(set(self.workers))
+            to_close = set(self.workers) - set(self.worker_spec)
+            if to_close:
+                await self.scheduler.retire_workers(workers=list(to_close))
+                tasks = [self.workers[w].close() for w in to_close]
+                for task in tasks:
+                    await task
```
jcrist: Perhaps use `asyncio.gather` or `asyncio.wait` here to avoid a linear iteration when things could be interleaved?

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+            for name in to_close:
+                del self.workers[name]
+
+            to_open = set(self.worker_spec) - set(self.workers)
+            workers = []
+            for name in to_open:
+                d = self.worker_spec[name]
+                cls, opts = d["cls"], d.get("options", {})
+                if "name" not in opts:
+                    opts = toolz.merge({"name": name}, opts, {"loop": self.loop})
```
jcrist: Did you mean to include the `loop` in here?

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+                worker = cls(self.scheduler.address, **opts)
+                self._created.add(worker)
+                workers.append(worker)
+            if workers:
+                await asyncio.wait(workers)
+                for w in workers:
+                    w._cluster = weakref.ref(self)
```
jcrist: What is the cluster weakref for?

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+                    if self.status == "running":
+                        await w
```
jcrist: The non running workers are never awaited, what happens to them? They're still added to the workers dict below.

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+    async def __aenter__(self):
+        await self
+        return self
+
+    async def __aexit__(self, typ, value, traceback):
+        await self.close()
+
+    async def _close(self):
+        while self.status == "closing":
+            await asyncio.sleep(0.1)
```
jcrist: Instead of polling, could have a future for the closing operation (created by the first call to `_close`), and just wait on that?

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+            raise ValueError("Cluster is closed")
+
+        self._lock = asyncio.Lock()
+        self.status = "starting"
+        self.scheduler = await self.scheduler
+        self.status = "running"
+
+    def _correct_state(self):
+        if self._correct_state_waiting:
+            # If people call this frequently, we only want to run it once
```
jcrist: I think this drops scale requests while a current scale request is processing:

- Call scale
- Spec updated
- correct state task start, task is stored as `_correct_state_waiting`
- scale returns
- Call scale again
- Spec updated
- since previous call is still in progress, state is not corrected, no new workers are started/stopped. Spec and tasks are now out of sync. Also, since there are multiple `await` calls in `_correct_state_internal`, the `worker_spec` can be different at different points in that function, leading to potential bugs.

One naive solution would be to have a background task that loops forever, waiting on an event:


```python
while self.running:
    await self._spec_updated.wait()
    # update workers to match spec
    # After updating, only clear the event if things are up to date
    # If things aren't up to date, then we loop again
    if self.spec_matches_current_state():
        self._spec_updated.clear()
```

Then `_correct_state` would look like:

```python
def _correct_state(self):
    # set the event, it's only ever cleared in the loop
    # We force synchronization here to prevent scheduling tons
    # of tasks all setting the event, this blocks until it's set.
    return self.sync(self._mark_state_updated)

async def _mark_state_updated(self):
    self._state_updated.set()
```

There are likely other ways to handle this. In dask-gateway I have a task per worker/scheduler. As the spec updates, unfinished tasks are cancelled or new ones are fired. If a previous `scale` call is still in progress for a cluster, `scale` will block until that call has finished. Note that this only blocks while we update our internal task state (cancelling/firing new tasks), not until those tasks have completed.

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+
+        if not self.asynchronous:
+            self._loop_runner.start()
+            self.sync(self._start)
+
+        self._correct_state_waiting = None
+
+    async def _start(self):
+        while self.status == "starting":
+            await asyncio.sleep(0.01)
```
jcrist: Same here as closing, could wait on the start task instead of polling.

distributed/deploy/spec.py
```diff
@@ -0,0 +1,296 @@
...
+        with ignoring(RuntimeError):  # loop closed during process shutdown
+            return self.sync(self._close)
+
+    def __del__(self):
+        if self.status != "closed":
+            self.close()
+
+    def __enter__(self):
+        self.sync(self._correct_state)
+        self.sync(self._wait_for_workers)
```
jcrist: Does this mean that `__enter__` will only complete once the initial n workers have started? What happens if we request 2, 1 worker starts and 1 fails?

distributed/deploy/local.py
```diff
@@ -188,268 +160,40 @@ def __init__(
...
+                blocked_handlers=blocked_handlers,
+            ),
+        }
+
+        worker = {
+            "cls": worker_class or (Worker if not processes else Nanny),
+            "options": worker_kwargs,
+        }
+
+        SpecCluster.__init__(
```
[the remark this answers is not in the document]
mrocklin: Thanks.  Fixed.

distributed/deploy/spec.py
mrocklin: Things get a bit tricky because of asyncio/tornado gen.coroutine.  I've fixed this now so that things should work regardless of how `Worker.close` is defined.  Thanks for flagging it.

distributed/deploy/spec.py
mrocklin: Yes, ideally we want the worker to use the IOLoop used by the cluster object.

distributed/deploy/spec.py
mrocklin: There are a lot of weakrefs around now.  They're useful when tracking down leaking references to things.

distributed/deploy/spec.py
mrocklin: This is again a tornado/asyncio difference.  I've removed the running check and made things optimal, I think for both `async def` and `gen.coroutine` style functions.

distributed/deploy/spec.py
mrocklin: Good thought.  I'm inclined to wait on this for now though if that's ok.

distributed/deploy/spec.py
jcrist: I mean that `loop` is only added if `name` is not in opts. Wouldn't you always want to pass it?

distributed/deploy/spec.py
mrocklin: > since previous call is still in progress, state is not corrected, no new workers are started/stopped. Spec and tasks are now out of sync. Also, since there are multiple await calls in _correct_state_internal, the worker_spec can be different at different points in that function, leading to potential bugs.

So, the `_correct_state_waiting` attribute isn't the currently running task, it's the currently enqueued one.  Once `_correct_state` starts running it immediately clears this attribute.  After someone calls `scale` there is a clean, not-yet-run `_correct_state_waiting` future that will run soon.

distributed/deploy/spec.py
mrocklin: Yes, this might hang.  I'm not sure we ever had a test in our test suite with this case.  I'll add something.

distributed/deploy/spec.py
mrocklin: Ah, indeed.  Looking at this again it looks like we do this in an async def function anyway, so `IOLoop.current()` should be valid regardless.  I'll remove the reference to loop entirely, which should be helpful in reducing the contract too.

distributed/deploy/spec.py
mrocklin: Added a test in 5e940691

mrocklin:
Thanks for the review @jcrist !  If you have a chance to pass through things tomorrow I would appreciate it

mrocklin:
I plan to merge this later today if there are no further comments.  Tests here are pretty decent, although I'll need to overhaul adaptive.  I'd like to do this in a separate PR though.

distributed/deploy/spec.py
jcrist: Since `_correct_state_waiting` waits on the created workers, this does mean that there's no way to cancel pending workers. This is fine for `LocalCluster`, but would be problematic if used as a base class for other cluster managers. The following would request and start 100 workers before scaling back down afaict:

```python
cluster.scale(100)
cluster.scale(2)
```

mrocklin: I think that this depends on what you mean by "waits on".  

One approach is that for a cluster manager to reach a correct state it only has to successfully submit a request to the resource manager have received an acknowledgement that the resource manager is handling it. We're not guaranteeing full deployment, merely that we've done our part of the job.  I would expect this to almost always be fairly fast.

Separately, there is now a `Client.wait_for_workers(n=10)` method that might be used for full client <-> scheduler checks.

mrocklin:
OK.  Merging this in.  I intend to be active in this area for a while, so if there are still issues please feel free to raise them.  I plan to do the following:

1.  Fix up adaptive so that it moves logic into the scheduler, and makes tests here pass
2.  Try out SpecCluster with Dask-Kubernetes.  I imagine that this will force some changes here.

PR #2686, opened 2019-05-11
Remove parse/format functions from utils

mrocklin:
These were moved to dask in https://github.com/dask/dask/pull/4793

### DIFF
--- a/distributed/tests/test_utils.py
+++ b/distributed/tests/test_utils.py
@@ -1,6 +1,5 @@
 from __future__ import print_function, division, absolute_import
 
-import datetime
 from functools import partial
 import io
 import socket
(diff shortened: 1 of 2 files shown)


TomAugspurger:
In the next distributed release, do we plan to bump the minimum dask version to something with dask/dask#4793? Else, we'll need to temporarily keep both implementations around.

mrocklin:
I'm more than happy to bump the minimum dask version aggressively, but I'm
also happy waiting on this PR until after the next release (otherwise dev
workflows will probably get more confusing).

TomAugspurger:
+1 for waiting for the next dask release and bumping the minimum version.

mrocklin:
I think that this was done elsewhere.

PR #2688, opened 2019-05-13
Wait for n workers before continuing

### LINKED ISSUE #2138: Wait for workers to join before continuing
_Moved from dask/dask-kubernetes#87_

It would be useful to have the `cluster.scale()` method block until the workers join. This was originally raised in dask-kubernetes but is applicable to dask-jobqueue and dask-yarn so would make more sense to implement here.

Suggestions from @mrocklin:

> One solution here might be to use SchedulerPlugins to explicitly trigger events whenever a new worker reaches the scheduler. Calling wait on the cluster object might register a tornado.locks.Event or tornado.locks.Condition that gets triggered when the right number of workers have arrived.

> @jhamman implemented a SchedulerPlugin for dask-jobqueue in order to improve adaptive operation (he wanted adaptive to be aware of pending jobs). This work happened in dask/dask-jobqueue#63 and might be the sort of thing we could think about generalizing. Any thoughts on this @jhamman?

danpf:
This is the easiest and I think most well supported implementation of waiting for n workers.

Initially I thought it was wrong to be in `Client`, but considering the implementation of `Cluster` -- (the bulk of which is in `LocalCluster`) this is the solution which is most immediately compatible with jobqueue and kubernetes.

The alternative would be to move `asynchronous` and `loop` from `LocalCluster` to `Cluster`,  and then make this a top level member of Cluster.  In addition, jobqueue would need to inherit from `Cluster` instead of `ClusterManager`.  This is currently tested and working on `SLURMCluster`, I will think of a way to test it without jobqueue though once we settle on the correct path.

I'm okay with both ways, but figured I would start with this since it's so low effort.
@mrocklin @guillaumeeb 

fix #2138

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1061,6 +1061,7 @@ class Client(Node):
             info = yield self.scheduler.identity()
 
     def wait_for_workers(self, n_workers=0):
+        """Blocking call to wait for n workers before continuing"""
         return self.sync(self._wait_for_workers, n_workers)
 
     def _heartbeat(self):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1053,6 +1053,14 @@ class Client(Node):
         except EnvironmentError:
             logger.debug("Not able to query scheduler for identity")
 
+    @gen.coroutine
+    def _wait_until_n_workers(self, n):
+        while n and len(self.cluster.scheduler.workers) < n:
+            yield gen.sleep(0.01)
+
+    def wait_until_n_workers(self, n):
+        return self.sync(self._wait_until_n_workers, n)
+
     def _heartbeat(self):
         if self.scheduler_comm:
             self.scheduler_comm.send({"op": "heartbeat-client"})
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1056,7 +1056,7 @@ class Client(Node):
     @gen.coroutine
     def _wait_for_workers(self, n_workers=0):
         info = yield self.scheduler.identity()
-        while n_workers and len(info['workers']) < n_workers:
+        while n_workers and len(info["workers"]) < n_workers:
             yield gen.sleep(0.1)
             info = yield self.scheduler.identity()
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1054,14 +1054,21 @@ class Client(Node):
             logger.debug("Not able to query scheduler for identity")
 
     @gen.coroutine
-    def _wait_until_n_workers(self, n):
-        info = yield self.scheduler.identity()
-        while n and len(info['workers']) < n:
-            yield gen.sleep(0.200)
-            info = yield self.scheduler.identity()
+    def _wait_until_n(self, func, n):
+        while n and (yield func()) < n:
+            yield gen.sleep(0.2)
 
-    def wait_until_n_workers(self, n):
-        return self.sync(self._wait_until_n_workers, n)
+    def wait_until_n(self, workers=0):
+        """
+        Note: If you add more arguments, you will need to return a list of
+        futures, or assert that only 1 is called at a time.
+        """
+        if workers:
+            @gen.coroutine
+            def f():
+                info = yield self.scheduler.identity()
+                raise gen.Return(len(info['workers']))
+            return self.sync(self._wait_until_n, f, workers)
 
     def _heartbeat(self):
         if self.scheduler_comm:
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1056,7 +1056,7 @@ class Client(Node):
     @gen.coroutine
     def _wait_for_workers(self, func, n):
         while n and (yield func()) < n:
-            yield gen.sleep(0.2)
+            yield gen.sleep(0.1)
 
     def wait_for_workers(self, n_workers=0):
         """
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1054,21 +1054,14 @@ class Client(Node):
             logger.debug("Not able to query scheduler for identity")
 
     @gen.coroutine
-    def _wait_for_workers(self, func, n):
-        while n and (yield func()) < n:
+    def _wait_for_workers(self, n_workers=0):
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -1053,6 +1053,14 @@ def _update_scheduler_info(self):
         except EnvironmentError:
             logger.debug("Not able to query scheduler for identity")
 
+    @gen.coroutine
+    def _wait_until_n_workers(self, n):
+        while n and len(self.cluster.scheduler.workers) < n:
+            yield gen.sleep(0.01)
```
mrocklin: I agree with @guillaumeeb 's point that we shouldn't assume the existence of either the scheduler or even the cluster attribute.  The generic way to do this is to poll the scheduler's identity function.

```python
while True:
    info = yield self.scheduler.identity()
    if len(info['workers']) >= n:
        break
    else:
        yield gen.sleep(0.200)
```

The `self.scheduler` attribute used above is a connection to the scheduler that will always be there, rather than an explicit reference to the scheduler itself.

danpf:
Maybe we could do something like `wait_until_n('workers', 3)` with respective if-statements and that function serve `_wait_until_n` a number  + a function that is evaluated until the criterea is hit? 

```
    @gen.coroutine
    def _wait_until_n(self, func, n):
        while n and (yield func()) < n:
            yield gen.sleep(0.200)

    def wait_until_n(self, thing, n):
        if thing == 'workers':
            @gen.coroutine
            def f():
                info = yield self.scheduler.identity()
                raise gen.Return(len(info['workers']))
            return self.sync(self._wait_until_n, f, n)

```

mrocklin:
yes, something like that might work.  We could probably use standard keyword arguments for this as well

```python
def wait_for_workers(n_workers=None):
    ...
```

Then as we add more things we can add more keywords

```python
def wait_for_workers(n_workers=None, memory=None, cores=None):
    ...
```

danpf:
@guillaumeeb what should we do when the scheduler isn't there? currently I have it sleep until it appears, but maybe we should raise error.... or raise error after a counter triggers.

distributed/client.py
```diff
@@ -1053,6 +1057,26 @@ def _update_scheduler_info(self):
...
+
+    def wait_until_n(self, workers=0):
+        """
+        Note: If you add more arguments, you will need to return a list of
+        futures, or assert that only 1 is called at a time.
+        """
+        if workers:
+            @gen.coroutine
+            def f():
+                info, _ = self._get_current_info_and_scheduler()
```
mrocklin: Thoughts on using `yield self.scheduler.identity()` here instead ?

distributed/client.py
```diff
@@ -1053,6 +1057,26 @@ def _update_scheduler_info(self):
         except EnvironmentError:
             logger.debug("Not able to query scheduler for identity")
 
+    @gen.coroutine
+    def _wait_until_n(self, func, n):
+        while n and (yield func()) < n:
+            yield gen.sleep(0.2)
+
+    def wait_until_n(self, workers=0):
```
mrocklin: I recommend the name `wait_for_workers` instead.  I'm not sure that a novice user will understand what `wait_until_n` means as immediately.

distributed/client.py
danpf: `_get_current_info_and_scheduler` helps make sure that we have a scheduler and a cluster, and that they are valid. it's a wrapper around the first part of `_repr_html_`.

your suggested way seems fine to me, and was the first way I did it, but I was just doing what @guillaumeeb suggested.  I definitely am less aware of all of the possible scenarios than you guys.

> Maybe you will need to add some conditionals like in 

is there no need to worry about that?

mrocklin: Please just use `yield self.scheduler.identity()`.  It's simpler and cleaner.  If you run into an issue with this then bring it up and we'll handle it.  I think that you'll be fine though.

> helps make sure that we have a scheduler and a cluster, and that they are valid

We may not ever have a cluster object locally.   That shouldn't be required.  

If you want to wait until things are set up cleanly then you could `yield self`, but that should already have run, and I don't think should be necessary.

distributed/client.py
```diff
@@ -1053,6 +1053,23 @@ def _update_scheduler_info(self):
...
 
+    @gen.coroutine
+    def _wait_for_workers(self, func, n):
+        while n and (yield func()) < n:
+            yield gen.sleep(0.1)
+
+    def wait_for_workers(self, n_workers=0):
+        """
+        Note: If you add more arguments, you will need to return a list of
+        futures, or assert that only 1 is called at a time.
```
mrocklin: This looks like it is a comment designed for other developers.  For this I recommend that you use comments (with the `#` character).  For the docstring lets follow this approach: https://docs.dask.org/en/latest/develop.html#docstrings

distributed/client.py
```diff
@@ -1053,6 +1053,23 @@ def _update_scheduler_info(self):
...
+        """
+        if n_workers:
+            @gen.coroutine
+            def f():
+                info = yield self.scheduler.identity()
+                raise gen.Return(len(info['workers']))
+            return self.sync(self._wait_for_workers, f, n_workers)
```
mrocklin: Rather than have the nested coroutine here I recommend that you unpack the definition of f in `_wait_for_workers`.  I think that this should do it.

```python
def _wait_for_workers(self, n_workers=0):
    info = yield self.scheduler.identity()
    while len(info['workers']) < n_workers:
        yield gen.sleep(0.1)
        info = yield self.scheduler.identity()

def wait_for_workers(self, n_workers=0):
    return self.sync(self._wait_for_workers, n_workers=n_workers)
```

mrocklin:
This looks good to me.  There are a couple of unrelated intermittent failures.  Also it looks like there are some linting issues.  I recommend using `black` as described in https://github.com/dask/distributed/blob/master/CONTRIBUTING.md

See also https://travis-ci.org/dask/distributed/jobs/532348315

mrocklin:
This looks good to me.  Thanks @danpf .  Mering

PR #2689, opened 2019-05-13
fix/worker-bokeh-page

quasiben:
PR is a small fix to worker url path.  Path is now at `/status` instead of `/main`


mrocklin:
Would it be possible to add a redirect to help existing users that use this page?

My guess is that this would be added to the normal Tornado routes in `distributed/bokeh/worker_html.py`

mrocklin:
> Would it be possible to add a redirect to help existing users that use this page?

Poke @quasiben

quasiben:
coming back to this now

mrocklin:
Thanks @quasiben

quasiben:
Thanks for those last commits!

PR #2690, opened 2019-05-13
RFC / ENH: Add Semaphore extension

fjetter:
This implements a distributed Semaphore. The implementation is inspired by the kazoo Semaphore implementation (c.f. [here](https://github.com/python-zk/kazoo/blob/88b657a0977161f3815657878ba48f82a97a3846/kazoo/recipe/lock.py#L426)) with some obvious modifications since we're not using zookeeper here.

The internal data structure of this semaphore is a bit complex since we do not support any notion of an _ephemeral node_, i.e. a value which expires together with the session. In `dask.distributed` context, I believe, this translates to the `Client`. Therefore, the Semaphore tracks which lease stems from which `Semaphore` client instance and stores its associated Client ID. If the client is lost/closed, the semaphore will release all it's acquired values eventually.

This behavior is quite important for resilience: If a worker is shut down ungracefully I want the leases to be released eventually, otherwise my application may be blocked. In early tests I was using a plain tornado Semaphore without the complex state logic but this didn't work out for us.

Instead of opening an issue I wanted to discuss this over a bit of code: 
* Is an implementation like this to be desired as part of the project or is this out of scope? 
* If we decide to merge this, I would propose to replace the Lock implementation (or rather merge it, `Lock ~ Semaphore(max_leases=1)`) since the complexity is quite high and maintenance of both is probably not desired

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -1,0 +1,264 @@
+from __future__ import absolute_import, division, print_function
+
+import uuid
+from collections import defaultdict, deque
+from functools import partial
+
+import dask
+import tornado.locks
+import tornado.queues
+from tornado import gen
+
+from .client import Client, _get_global_client
+from .utils import PeriodicCallback, log_errors, parse_timedelta
+from .worker import get_client, get_worker
+from toolz.dicttoolz import valmap
+from .metrics import time
+
+
+class _Watch(object):
+    def __init__(self, duration=None):
+        self.duration = duration
+        self.started_at = None
+
+    def start(self):
+        self.started_at = time()
+
+    def leftover(self):
+        if self.duration is None:
+            return None
+        else:
+            elapsed = time() - self.started_at
+            return max(0, self.duration - elapsed)
+
+
+class SemaphoreExtension(object):
+    """ An extension for the scheduler to manage Semaphores
+
+    This adds the following routes to the scheduler
+
+    * semaphore_acquire
+    * semaphore_release
+    * semaphore_create
+    """
+
+    def __init__(self, scheduler):
+        self.locks = defaultdict(tornado.locks.Lock)
+        self.scheduler = scheduler
+        self.leases = defaultdict(deque)
+        self.events = defaultdict(tornado.locks.Event)
+        self.max_leases = dict()
+        self.leases_per_client = defaultdict(partial(defaultdict, deque))
+        self.scheduler.handlers.update(
+            {
+                "semaphore_create": self.create,
+                "semaphore_acquire": self.acquire,
+                "semaphore_release": self.release,
+            }
+        )
+
+        self.scheduler.extensions["semaphores"] = self
+        self.pc_validate_leases = PeriodicCallback(
+            self._validate_leases,
+            parse_timedelta(
+                dask.config.get("distributed.scheduler.locks.lease-validation-interval")
+            ),
+            io_loop=self.scheduler.loop,
+        )
+        self.pc_validate_leases.start()
+        self._validation_running = False
+
+    @gen.coroutine
+    def create(
+        self, stream=None, name=None, client=None, timeout=None, max_leases=None
+    ):
+        if name not in self.leases:
+            assert isinstance(max_leases, int), max_leases
+            self.max_leases[name] = max_leases
+        else:
+            if max_leases != self.max_leases[name]:
+                raise ValueError(
+                    "Inconsistent max leases: %s, expected: %s"
+                    % (max_leases, self.max_leases[name])
+                )
+
+    @gen.coroutine
+    def _get_lease(self, client, name, identifier):
+        # We should make sure that the client is already properly registered with the scheduler
+        # otherwise the lease validation will mop up every acquired release immediately
+        while client not in self.scheduler.clients:
+            yield
+        with (yield self.locks[name].acquire()):
+            result = True
+            if len(self.leases[name]) < self.max_leases[name]:
+                self.leases[name].append(identifier)
+                self.leases_per_client[client][name].append(identifier)
+            else:
+                result = False
+            raise gen.Return(result)
+
+    @gen.coroutine
+    def acquire(
+        self, stream=None, name=None, client=None, timeout=None, identifier=None
+    ):
+        with log_errors():
+            if isinstance(name, list):
+                name = tuple(name)
+            w = _Watch(timeout)
+            w.start()
+
+            while True:
+                # Reset the event and try to get a release. The event will be set if the state
+                # is changed and helps to identify when it is worth to retry an acquire
+                self.events[name].clear()
+                future = self._get_lease(client, name, identifier)
+                if timeout is not None:
+                    future = gen.with_timeout(w.leftover(), future)
+                try:
+                    result = yield future
+                except gen.TimeoutError:
+                    result = False
+
+                # If acquiring fails, we wait for the event to be set, i.e. something has
+                # been released and we can try to acquire again (continue loop)
+                if not result:
+                    future = self.events[name].wait()
+                    if timeout is not None:
+                        future = gen.with_timeout(w.leftover(), future)
+                    try:
+                        yield future
+                        continue
+                    except gen.TimeoutError:
+                        result = False
+                raise gen.Return(result)
+
+    @gen.coroutine
+    def release(self, stream=None, name=None, client=None, identifier=None):
+        with log_errors():
+            if isinstance(name, list):
+                name = tuple(name)
+            if name in self.leases and identifier in self.leases[name]:
+                self.scheduler.loop.add_callback(
+                    self._release_value, name, client, identifier
+                )
+            else:
+                raise ValueError("Semaphore released too many times.")
+
+    @gen.coroutine
+    def _release_value(self, name, client, identifier):
+        with (yield self.locks[name].acquire()):
+            self.leases_per_client[client][name].remove(identifier)
+            self.leases[name].remove(identifier)
+            self.events[name].set()
+
+    def _release_client(self, client):
+        semaphore_names = list(self.leases_per_client[client])
+        for name in semaphore_names:
+            ids = list(self.leases_per_client[client][name])
+            for _id in list(ids):
+                self.scheduler.loop.add_callback(
+                    self._release_value, name=name, client=client, identifier=_id
+                )
+
+    @gen.coroutine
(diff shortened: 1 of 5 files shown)


distributed/tests/test_semaphore.py
```diff
@@ -0,0 +1,133 @@
...
+@gen_cluster(client=True)
+def test_release_semaphore_after_timeout(c, s, a, b):
+    with dask.config.set(
+        {"distributed.scheduler.locks.lease-validation-interval": "100ms"}
+    ):
+        sem = Semaphore(name="x", max_leases=2)
+        yield sem.acquire()
+        semY = Semaphore(name="y")
+
+        with Client(s.address, asynchronous=True, name="ClientB") as clientB:
```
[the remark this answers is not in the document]
fjetter: We're not defining these tests using `async def` to generate the coroutines but rather `tornado.gen.coroutine` which probably takes care of the matter somehow. With this definition `async with` it raises a `SyntaxError`.
The tests in `tests/py3_test_*.py` define the proper py3 syntax using `async def`. Do you think it is worth it/good practice to write these async tests as proper `async def` coroutines instead of relying on the tornado conversion?

fjetter:
Test failures on appveyor seem to be unrelated (timeouts for bokeh, profiling, etc.). Is this a known issue?

mrocklin:
> Test failures on appveyor seem to be unrelated (timeouts for bokeh, profiling, etc.). Is this a known issue?

There are currently known intermittent failures, yes.  I wouldn't be too worried yet if you see an unexpected test fail.

mrocklin:
@cicdw I noticed that you 👍 the comment about reviewing.  Does that mean that you're interested in reviewing this?  No pressure (everyone is busy) just seeing if we should expect something from you (which would be great) or try to find someone else to review.

cicdw:
@mrocklin I've given it a once over and will do another pass, but I wouldn't wait for any substantive feedback from me.

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+        )
+        self.pc_validate_leases.start()
+        self._validation_running = False
+
+    @gen.coroutine
+    def create(
+        self, stream=None, name=None, client=None, timeout=None, max_leases=None
+    ):
+        if name not in self.leases:
+            assert isinstance(max_leases, int), max_leases
```
cicdw: Question: is there a reason an `assert` is preferred to a more explicit `ValueError` here?

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+    def __aenter__(self):
+        yield self.acquire()
+        raise gen.Return(self)
+
+    @gen.coroutine
+    def __aexit__(self, *args, **kwargs):
+        yield self.release()
+
+    def __getstate__(self):
+        return (self.name, self.client.scheduler.address, self.max_leases)
```
cicdw: Should `id` also be serialized here, or do we want `Semaphore`s which have gone through the serialization / deserialization process to be considered different?

distributed/semaphore.py
fjetter: I would probably even remove this. I used this a bit for debugging. In we keep it, a `ValueError` or `TypeError` would make the most sense

distributed/semaphore.py
fjetter: TL;DR: I think we should not serialize the ID. I should probably add a test for this behavior.

Consider the following:

```python
from dask import delayed
from distributed import Semaphore

def access_database(sem, *args):
    with sem:
        ...
    return data

sem = Semaphore("my_database", max_leases=2)
tasks = []
for ix in range(10):
    tasks.append(
        delated(access_database)(sem, ix)
    )

dask.compute(tasks)
```

This is how I would probably write a piece of code to download data from a DB. I could rely on `dask.distributed` to serialize the class for me. Otherwise I would need to pass all initialization parameters to the function and initialize the objects myself to keep everything consistent.
When I use the serialization approach, I still require the semaphore instance for every job to be unique by ID. Otherwise the tracking is off and the mechanism for tracking leases by ID and client would not work.

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+                dask.config.get("distributed.scheduler.locks.lease-validation-interval")
+            ),
+            io_loop=self.scheduler.loop,
+        )
+        self.pc_validate_leases.start()
+        self._validation_running = False
+
+    @gen.coroutine
+    def create(
+        self, stream=None, name=None, client=None, timeout=None, max_leases=None
```
mrocklin: I recommend renaming `stream=` to `comm=`.  This switch happened a while ago but apparently not all code was updated.

mrocklin: I'm actually a little surprised that this worked.  We must be using a positional argument somewhere.

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+                    "Inconsistent max leases: %s, expected: %s"
+                    % (max_leases, self.max_leases[name])
+                )
+
+    @gen.coroutine
+    def _get_lease(self, client, name, identifier):
+        # We should make sure that the client is already properly registered with the scheduler
+        # otherwise the lease validation will mop up every acquired release immediately
+        while client not in self.scheduler.clients:
+            yield
```
mrocklin:
```suggestion
            yield gen.sleep(0.100)
```

Otherwise this stresses out the event loop.

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+        with (yield self.locks[name].acquire()):
```
mrocklin: Why do we need an async lock around this block?  There are no `yield` points within it, so I wouldn't expect there to be any chance for another coroutine to happen at the same time.

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+            if name in self.leases and identifier in self.leases[name]:
+                self.scheduler.loop.add_callback(
+                    self._release_value, name, client, identifier
+                )
+            else:
+                raise ValueError("Semaphore released too many times.")
+
+    @gen.coroutine
+    def _release_value(self, name, client, identifier):
+        with (yield self.locks[name].acquire()):
```
mrocklin: Assuming that these locks aren't necessary (I don't think they're ever used around code with yields) then my guess is that a bit of this can be simplified.

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @gen.coroutine
+    def _validate_leases(self):
+        if not self._validation_running:
+            self._validation_running = True
+            known_clients = set(self.leases_per_client.keys())
+            scheduler_clients = set(self.scheduler.clients.keys())
+            for client in known_clients - scheduler_clients:
+                client_has_leases = sum(
+                    valmap(len, self.leases_per_client[client]).values()
```
mrocklin: `valmap(func, d).values()` -> `map(func, d.values())` ?

distributed/semaphore.py
```diff
@@ -0,0 +1,264 @@
...
+                    self._release_client(client)
+            else:
+                self._validation_running = False
+
+
+class Semaphore(object):
+    def __init__(self, max_leases=1, name=None, client=None):
+        self.client = client or _get_global_client() or get_worker().client
+        self.id = uuid.uuid4().hex
+        self.name = name or "semaphore-" + uuid.uuid4().hex
```
mrocklin: reuse `id` here?

mrocklin:
Checking in @fjetter , any thoughts on my comments above?

distributed/semaphore.py
fjetter: Yes, works without locks. Will remove them again.

fjetter:
* I tried rewriting the code using `async def` instead of `@gen.coroutine` but am hitting an issue where something is complaining that coroutines are not serializable. Haven't tracked it further down, yet
* Using tornado semaphores was my first prototype but the issue about lost connections is quite important, I believe. Especially in autoscaling environments or volatile environments where a worker might die unexpectedly, I consider this extremely important. An alternative to the tracking and releasing upon loosing the client, I could imagine a maximum lease time to be a viable option. To track an acquisition and expire it after a certain amount of time is probably equally complex and probably less powerful.
* I encountered another issue which I need to track down. In our environment I receive connection timeouts when semaphores cannot be acquired. Haven't understood this issue yet, but my first guess would be that if `semaphore_timeout < connect_timeout` we get a connection timeout (worker dies, computation is canceled). I'll try to catch this with another test to be sure I'm not on the wrong track here.

Due to the complexity of this implementation, I was asking if this is considered to be in-scope of `dask.distributed`.
IMHO, a distributed semaphore needs to be able to deal with worker loss to be useful, at least for production-level services.

mrocklin:
> I tried rewriting the code using async def instead of @gen.coroutine but am hitting an issue where something is complaining that coroutines are not serializable. Haven't tracked it further down, yet

It's totally fine to leave this with `@gen.coroutine`.  There's no need to switch if you don't want to.

> Using tornado semaphores was my first prototype but the issue about lost connections is quite important, I believe. Especially in autoscaling environments or volatile environments where a worker might die unexpectedly, I consider this extremely important. An alternative to the tracking and releasing upon loosing the client, I could imagine a maximum lease time to be a viable option. To track an acquisition and expire it after a certain amount of time is probably equally complex and probably less powerful.

OK.  Mostly I wanted to verify that we weren't missing an obvious simplification here.  If you've already tried this and have some confidence that it's not the right approach then I'm happy to drop it.

> Due to the complexity of this implementation, I was asking if this is considered to be in-scope of dask.distributed.

Yes, this seems in-scope to me.  I will be glad to have it around.  Several other groups have asked for similar functionality.

fjetter:
I'll have another look at the implementation and the issues I mentioned in the following days.

@mrocklin what's your stance towards replacing the Lock implementation with a single valued semaphore? The lock implementation also looks quite complex and uses a similar logic using tornado events and I'd argue that maintaining both implementations isn't necessary. On the contrary, the lock implementation would also benefit from introduced features like "release locks on client loss" or similar.

mrocklin:
Basing the Lock implementation on top of this sounds fine to me in
principle.

fjetter:
FYI: We're currently taking another look at this. Essentially we intend to migrate the code to asyncio and cover a few more edge cases which came up in an internal review. Will ping again if there is progress.

fjetter:
Closing this now in favor of #3573

PR #2692, opened 2019-05-13
Release 1.28.1

mrocklin:
So, I've pushed this up to PyPI.  I'm not sure if we want to merge this or not (the relevant commit is already in history), but I suspect that @jakirkham knows more about how to handle this than I do, so I'm putting this up here.

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -8,8 +8,8 @@ from sys import exit
 import warnings
 
 import click
+import dask
 from distributed import Nanny, Worker
-from distributed.config import config
 from distributed.utils import get_ip_interface, parse_timedelta
 from distributed.worker import _ncores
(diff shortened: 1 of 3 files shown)


jakirkham:
So I've created a merge commit, which should do the right thing when pushed to `master`, which lives [here]( https://github.com/jakirkham/distributed/commit/e0b7b85114d61b9201dd7aef6661d7345c0b74ba ). It helps versioneer pick up the `1.28.1` tag correctly and thus version things appropriately. If it is ok, I'd like to push this directly to `master`, which will merge this PR along with it. Does this sound ok?

mrocklin:
Happy to defer to you on this :)

jakirkham:
Ok, thanks for the feedback. Have gone ahead and pushed. This PR is now merged.

Issue #2704, opened 2019-05-16
Exception with dask PubSub usage


natekupp:
Hey there, I'm getting an exception with PubSub that I can't quite figure out how to debug. 

I'm taking a topologically-sorted DAG and creating futures via `submit()` for each task in the DAG. Then, I'm publishing metadata events from each task into a PubSub, with the client subscribing to the events and forwarding them on to the rest of the application.

I've created a minimal example that represents what I'm trying to do:

https://gist.github.com/natekupp/4b8059f7b59f5ed9cfda96390c5c718d

When I run the above example, I get the following output:

```
[1, 2, 2, 3, 3, 3, 5, 5, 5, 5, 5, 4, 4, 4, 4]
tornado.application - ERROR - Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x120c29a60>)
Traceback (most recent call last):
  File "/Users/nate/.pyenv/versions/dagster-3.6.8/lib/python3.6/site-packages/tornado/ioloop.py", line 758, in _run_callback
    ret = callback()
  File "/Users/nate/.pyenv/versions/dagster-3.6.8/lib/python3.6/site-packages/tornado/stack_context.py", line 300, in null_wrapper
    return fn(*args, **kwargs)
  File "/Users/nate/.pyenv/versions/dagster-3.6.8/lib/python3.6/site-packages/distributed/pubsub.py", line 199, in cleanup
    self.client.scheduler_comm.send(msg)
  File "/Users/nate/.pyenv/versions/dagster-3.6.8/lib/python3.6/site-packages/distributed/batched.py", line 119, in send
    raise CommClosedError
distributed.comm.core.CommClosedError
```

Any ideas? Am I doing something wrong here? It seems like this is getting thrown by the cleanup on the client-side.

Thanks in advance!

mrocklin:
This actually has nothing to do with PubSub usage.  A heartbeat message was sneaking out during closing down a client.  Fixed in https://github.com/dask/distributed/pull/2705

jrbourbeau:
Fixed by #2705

PR #2705, opened 2019-05-16
Cleanly stop periodic callbacks in Client

mrocklin:
Previously we did this only in the asynchronous code, which left a gap
during which a heartbeat could sneak out.  Now we call it explicitly at
the beginning of the synchronous close command.


mrocklin:
Fixes https://github.com/dask/distributed/issues/2704

natekupp:
awesome, thanks for taking a look at this!

PR #2715, opened 2019-05-20
Fea/add proxying

### LINKED ISSUE #2160: Reuse nbserverproxy for Dask's dashboards
The nbserverproxy allows Jupyter servers to proxy other ports on localhost to anyone with access to the Jupyter server.  

We have a similar problem in dask.  People often have access to the dask-scheduler dashboard, but don't have access to each of the dask-worker dashboards.  However the dask-worker dashboards *are* visible to the scheduler.  

I'm curious if functionality from nbserverproxy could be adapted to suit Dask's needs as well.  There are some similarities

1.  In both cases we have a server at an accessible location can see an inaccessible location
2.  In both cases that server is a Tornado web server

And some differences

1.  The accessible server (scheduler) is likely on a different host than the inaccessible servers (workers)
2.  One is Jupyter, the other isn't

cc @yuvipanda, is this worth investigating?  Is it worth generalizing `nbserverproxy`?  "Nope, nbserverproxy is pretty tied to Jupyter internals" is a fine answer.  This isn't a huge pain point right now, I'm just curious about the possibility.

quasiben:
PR resolves #2160

This PR is largely done but I left some code in because I was somewhat undecided on where proxying should occur.  At first, I built a new proxying service but soon realized that the full service was unnecessary and could just run as another route in on the scheduler's dashboard.  If reviewers are ok with running the proxying as another route I can remove all of distributed/proxy/core.py  and the `Proxy` object in proxy_html.py.

## How it Works:

- uses jupyter_server_proxy (still needs to be added as an optional dependency and build proper try/except on import)
- encodes host address of the worker as a URL param
- proxy checks hosts is valid then proxies to the host and port of the worker dashboard

@yuvipanda if you have time, this would work would greatly benefit from a review from you


distributed/proxy/proxy_html.py
```diff
@@ -0,0 +1,124 @@
...
+        # route here first
+        # incoming URI /proxy/{proxy}/port
+        host = self.get_argument("host", None)
+        if not host:
+            host = "127.0.0.1"
+        self.host = host
+        worker = '%s:%s' % (self.host, str(port))
+        if not check_worker_bokeh_exits(self.scheduler, worker):
+            async def _noop():
+                return None
```
mrocklin: Debug code?

mrocklin:
Also thanks for doing this!  I'm excited about getting access to these dashboards.  I also think that once we have them more publicly available we'll probably start investing more time in adding to them.

mrocklin:
cc @jcrist who might find this interesting as well

distributed/cli/tests/test_dask_worker.py
```diff
@@ -268,5 +268,43 @@ def test_bokeh_non_standard_ports(loop):
                 except Exception:
                     sleep(0.5)
                     assert time() < start + 20
+
+        # TEST PROXYING NOT ENABLED
+        with pytest.raises(Exception):
+            url = 'http://localhost:8787/proxy/%s/status?host=127.0.0.1' % "4833"
```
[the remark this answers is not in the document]
quasiben: I believe we need these separate conventions or we need to rewrite the request before handing over to jupyter-server-proxy.  jupyter-server-proxy, is expecting `/proxy/{port}/{proxied_path}` so if we include the host in the path we need to manipulate the path before the proxy tries to make the proxied request.  I don't exactly like using the parameter in the URL but was unsure how else to nicely encode the host.  Perhaps jupyter-server-proxy devs have better ideas ?

distributed/proxy/proxy_html.py
quasiben: We need this here to ensure we only proxy to valid worker dashboards and nothing else

distributed/cli/dask_scheduler.py
```diff
@@ -67,6 +67,13 @@
     default=":8787",
     help="Address on which to listen for diagnostics dashboard",
 )
+@click.option(
+    "--use-proxy/--no-use-proxy",
+    "use_proxy",
+    default=False,
+    show_default=True,
+    help="Use proxy routes for worker dashboards from diagnostic page",
+)
```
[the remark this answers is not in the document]
quasiben: I don't think it's a security concern and thought I would spend the bit of effort to put it in?  It's easy to take out.  Networking often confuses me -- what do you think about making the default true and optionally allowing users to navigate directly to the workers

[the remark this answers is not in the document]
mrocklin: It's easier to add things in in the future than to take them out :)

If it's not very likely to be needed then I would be in favor of removing it, just to keep maintenance burden down.

distributed/cli/tests/test_dask_worker.py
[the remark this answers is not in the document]
mrocklin: **If** the proxy code is very simple then it might make sense to fork-and-vendor.  That's a big if though.

distributed/proxy/proxy_html.py
mrocklin: The function `_noop` doesn't seem to be used anywhere

distributed/cli/dask_scheduler.py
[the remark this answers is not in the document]
quasiben: Another thought which may impact proxying is if the `bokeh-prefix` is used with the dask-worker.  This information, I don't think, is passed back to the scheduler.  Using `bokeh-prefix` now is also not supported so it's not a huge impact but something which could be improved

distributed/proxy/proxy_html.py
quasiben: Oh!  Sorry, I thought you were referring to the function in general.  Yes, I'll take out `_noop`

mrocklin: We can have the worker send back its dashboard address if that would help.

distributed/cli/tests/test_dask_worker.py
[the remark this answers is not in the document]
quasiben: Backing up, is it that you prefer, `proxy/{host}/{port}`  or `?host={host}&port={port}` ?  or you just don't want to mix conventions ?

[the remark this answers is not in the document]
mrocklin: Which do you think is best?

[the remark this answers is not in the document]
mrocklin: Maybe the first?  I don't have strong thoughts here.  I'm also fine with what we have if that's best.

distributed/cli/dask_scheduler.py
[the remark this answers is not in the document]
quasiben: remove `use-proxy`.  default is now to use the proxy from the dashboard

distributed/bokeh/templates/worker-table.html
```diff
@@ -20,7 +20,7 @@
         <td> {{ len(ws.processing) }} </td>
         <td> {{ len(ws.has_what) }} </td>
         {% if 'bokeh' in ws.services %}
-        <td> <a href="http://{{ ws.host }}:{{ ws.services['bokeh'] }}">bokeh</a> </td>
+        <td> <a href="http://{{ ws.host }}:{{ scheduler.services['bokeh'].port }}/proxy/{{ ws.services['bokeh'] }}/{{ ws.host }}/status">bokeh</a> </td>
```
quasiben: @mrocklin This is the new url: `/proxy/{port}/{host}/{proxied_path}`.

quasiben:
@mrocklin I realized that I could manually overwrite the request uri (something I always had trouble with using ngxin) and a bit of regex in the handler made things smooth for having nicer URLs

distributed/bokeh/templates/worker-table.html
mrocklin: Oh wow, how did you manage that?

mrocklin: > @mrocklin I realized that I could manually overwrite the request uri (something I always had trouble with using ngxin) and a bit of regex in the handler made things smooth for having nicer URLs


Ah. Great!

yuvipanda: This is cool! 

Is ws.host coming from a trusted source? If it is user controlled, it could be used maliciously to bypass CORS / Origin checking.

mrocklin: This is the address that a Dask worker advertises as its contact address.  A user that has access to launch a Dask worker and connect it to this scheduler can manipulate this value, yes.

However, such a user already has the ability to connec to the scheduler and run arbitrary Python code there, or on any other worker-connected machine, so I suspect that at this point we're ok with this.

distributed/proxy/proxy_html.py
```diff
@@ -0,0 +1,132 @@
+from urllib import parse
```
mrocklin: Let's go ahead and remove the unnecessary files for now.

mrocklin: Ah, wait, this one *is* necessary, right?  Perhaps just the `Proxy` class at the bottom?

distributed/proxy/proxy_html.py
```diff
@@ -0,0 +1,132 @@
...
+from jupyter_server_proxy.handlers import ProxyHandler
+
+import tornado.ioloop
+import tornado.web
+from bokeh.server.server import BaseServer
+from tornado.httpserver import HTTPServer
+
+
+def get_host():
+    return "127.0.0.1"
```
mrocklin: Should this be removed?

distributed/cli/tests/test_dask_worker.py
```diff
@@ -264,9 +264,14 @@ def test_bokeh_non_standard_ports(loop):
...
                     redirect_resp.ok
+                    # TEST PROXYING WORKS
+                    url = "http://127.0.0.1:8787/proxy/4833/127.0.0.1/status"
+                    response = requests.get(url)
+                    assert response.ok
                     break
                 except Exception:
                     sleep(0.5)
                     assert time() < start + 20
+
```
mrocklin: If possible, lets also test with an async test.  The CLI tests are the least pleasant to work with when things go wrong and we have to debug.  There is a sort of hierarchy of `gen_cluster` > `c, s, a, b` fixtures > CLI for distributed tests.  See https://distributed.dask.org/en/latest/develop.html#writing-tests for more information.

The file `distributed/bokeh/tests/test_scheduler_bokeh.py` should have some good examples to pull from.

distributed/proxy/proxy_html.py
mrocklin: If this module is just a single file then lets unpack it from a directory in order to avoid hierarchy.  Also, currently most of the dashboard stuff is in the `bokeh/` folder (which we should probably consider renaming `dashboard/`).  I recommend that we put this in there for now.

distributed/cli/tests/test_dask_worker.py
quasiben: done

distributed/proxy/proxy_html.py
quasiben: remove

distributed/bokeh/tests/test_scheduler_bokeh.py
```diff
@@ -581,3 +581,37 @@ def test_root_redirect(c, s, a, b):
...
+    timeout=180,
+)
+def test_proxy_to_workers(c, s, a, b):
+    dashboard_port = s.services["bokeh"].port
+    http_client = AsyncHTTPClient()
+    response = yield http_client.fetch("http://localhost:%d/" % dashboard_port)
+    assert response.code == 200
+    assert "/status" in response.effective_url
+
+    yield [a.heartbeat(), b.heartbeat()]
```
mrocklin: Was this necessary?  If so, what data was missing?  We might want to make sure that this arrives on the first `register-worker` message.

requirements.txt
```diff
@@ -9,6 +9,7 @@ tblib
 toolz >= 0.7.4
 tornado >= 5
 zict >= 0.1.3
+jupyter-server-proxy == 1.1.0
```
mrocklin: The strict pinning makes me concerned here.  Can we do `>=`?

Also, does this bring in all of Jupyter?  If so, we might want to make this optional.

mrocklin: To be more explicit, strict pinning is awesome when making applications but leads to unhappy users when making libraries.  When multiple libraries use strict pinning then it becomes very hard to construct a valid environment.

distributed/bokeh/tests/test_scheduler_bokeh.py
```diff
@@ -581,3 +581,37 @@ def test_root_redirect(c, s, a, b):
...
+        )
+        direct_url = "http://localhost:%s/status" % port
+        http_client = AsyncHTTPClient()
+        response_proxy = yield http_client.fetch(proxy_url)
+        response_direct = yield http_client.fetch(direct_url)
+
+        assert response_proxy.code == 200
+        assert b"Crossfilter" in response_proxy.body
+        assert response_direct.code == 200
+        assert b"Crossfilter" in response_direct.body
```
mrocklin: Nice test!

quasiben:
Thanks for the review @mrocklin I think I've address all of your comments.  Another thought occurring to me as I've just added `jupyter-server-proxy` as a dependency is that it increases the size of distributed a bit.  `jupyter-server-proxy` has a hard dependency on notebook among others: https://github.com/jupyterhub/jupyter-server-proxy/blob/2726bca233af8f30e4b230b3ffce895fb1fa9b2e/setup.py#L13

For context, notebook is ~7MBs

distributed/bokeh/tests/test_scheduler_bokeh.py
quasiben: I was using this to ensure the workers were up before proceeding.  Apparently, it's not necessary

mrocklin: Right, this is handled in `gen_cluster` for you.

mrocklin:
> Another thought occurring to me as I've just added jupyter-server-proxy as a dependency is that it increases the size of distributed a bit. 

Then maybe we should make this optional.  Presumably we would try-except around the internal import of our own module when adding the route.

mrocklin:
> Then maybe we should make this optional. Presumably we would try-except around the internal import of our own module when adding the route.

Or maybe our proxy handler sends back a "Oops, couldn't proxy, please `pip install jupyter-server-proxy`" message (or something nicer)

requirements.txt
quasiben: I opted for `==` because I wasn't sure of the dev cycle and stability of jupyter-server-proxy and thus was being conservative (things work now) and we can afford to move slowly here.

It doesn't bring in all of Jupyter but it does bring in the notebook ~7MBs.  We could make this optional but it would me putting back some logic into worker-table.html or scheduler_html.py ?

distributed/bokeh/tests/test_scheduler_bokeh.py
mrocklin: Also, the conventional way to wait for workers to be in a good state is just to yield/await them

```python
yield [a, b]
```

But again, we already do this, I just wanted to show off our `__await__` methods :)

yuvipanda:
ProxyHandler also [inherits from IPythonHandler](https://github.com/jupyterhub/jupyter-server-proxy/blob/2726bca233af8f30e4b230b3ffce895fb1fa9b2e/jupyter_server_proxy/handlers.py#L31), which probably has implications around authentication - IIRC, we currently require / depend on notebook authentication to allow access.

distributed/bokeh/templates/worker-table.html
mrocklin: With the introduction of `scheduler` here we need to include that all uses of this template also include the `scheduler=` keyword, 

```diff
class Worker(RequestHandler):
    def get(self, worker):
        worker = escape.url_unescape(worker)
        with log_errors():
            self.render(
                "worker.html",
                title="Worker: " + worker,
                Worker=worker,
+               scheduler=self.server,
                **toolz.merge(self.server.__dict__, ns, self.extra)
            )
```

mrocklin: This is something that should have shown up on tests ideally.  If you felt the desire, it might be nice to add this to some test that checks the info pages (assuming that something like this already exists)

quasiben:
@yuvipanda that's a good point!  I think, though could be wrong, that as long as dask internals don't implement a LoginHandler/self.login_handler we should be ok.  `tornado.web.authenticated` is going to eventually ask for `self.current_user`:  

https://github.com/tornadoweb/tornado/blob/f65b1802a549bab73b4a6721bbfc47f82954d156/tornado/web.py#L3157-L3175

Fortunately for us the, the `AuthenticatedHandler` which is the parent of `IPythonHandler` passes `anonymous` when login_handler is None:


https://github.com/jupyter/notebook/blob/9425250005d46c9fb0057b0c322e16d07125b003/notebook/base/handlers.py#L136-L139

distributed/bokeh/templates/worker-table.html
quasiben: In my [last commit](https://github.com/dask/distributed/pull/2715/commits/d2d76e41b17f61350c99eb37787cc0c7204ccee6), I fixed the worker and task handlers.  The test were more or less already written in [test_scheduler_bokeh_html.py](https://github.com/dask/distributed/pull/2715/commits/d2d76e41b17f61350c99eb37787cc0c7204ccee6#diff-b5a38c24316b8b77073324cf2e59e29cR23) . The tests previously passed because the workers did not setup bokeh dashboards.  Now they do and tests pass again with the suggested addition above

mrocklin:
@quasiben is it ok to rename proxy_html.py to proxy.py ?

quasiben:
👍

distributed/bokeh/templates/worker-table.html
```diff
@@ -20,7 +20,7 @@
         <td> {{ len(ws.processing) }} </td>
         <td> {{ len(ws.has_what) }} </td>
         {% if 'bokeh' in ws.services %}
-        <td> <a href="http://{{ ws.host }}:{{ scheduler.services['bokeh'].port }}/proxy/{{ ws.services['bokeh'] }}/{{ ws.host }}/status">bokeh</a> </td>
+        <td> <a href="../../proxy/{{ ws.services['bokeh'] }}/{{ ws.host }}/status">bokeh</a> </td>
```
quasiben: Good idea!

mrocklin:
OK, done.  I plan to merge this after tests run.  I'll rename bokeh to dashboard shortly after

quasiben:
thanks for the edits to `scheduler.py`

mrocklin:
This is in.  Thanks @quasiben !

yuvipanda:
<3 great to see jupyter-server-proxy get used this way! \o/

PR #2719, opened 2019-05-21
Refer to LocalCluster in Client docstring

mrocklin:


martindurant:
This is long overdue, glad to see it.

PR #2721, opened 2019-05-22
Add docstring to Scheduler.check_idle_saturated

mrocklin:


martindurant:
+1

jjerphan:
Thanks @mrocklin!

jjerphan:
Just a naive follow-up: how are the behavior of `Scheduler.check_idle_saturated` has been calibrated, expecially for [determinating when a worker state is not idle and possibly saturated](https://github.com/dask/distributed/pull/2721/files#diff-6a3108098afcddf88b38652ae1e383b5R4389)?

mrocklin:
I'm not sure I fully understand the question.  Are you asking for the
reasoning behind the logic in that line?  If so, it's been a while, and I
don't personally fully remember.  Looking at it now my guess is that we
have to have more tasks than cores, more pending work than would take the
worker 400ms to do (I could be misreading that though), and more work than
almost twice the average of all workers.

jjerphan:
> I'm not sure I fully understand the question. Are you asking for the reasoning behind the logic in that line?

Yes, I am just curious.

OK, thank you for this explanation.

PR #2724, opened 2019-05-22
Move bokeh module to dashboard

mrocklin:
This is a hard break.  Previous imports will raise an informative error message.


mrocklin:
Merging this tomorrow if there are no further comments.

TomAugspurger:
I assume, but haven't checked, that the config keys will be changing from bokeh to dashboard?

```
logging:
  distributed: info
  bokeh: critical  # -> bokeh
  tornado: critical
```

If so, could I bother you for a doc update in https://github.com/dask/dask/blob/master/docs/source/configuration.rst as a followup?

mrocklin:
> I assume, but haven't checked, that the config keys will be changing from bokeh to dashboard?

The logging configuration that we do here is keyed on the name of the logger.  Bokeh's logger is still named "bokeh" and not "dashboard" so I'm inclined to keep this as-is.  It's similar to how we manage tornado.

TomAugspurger:
Sounds good to me.

Issue #2083, opened 2018-06-29
Add Client.open_dashboard


TomAugspurger:
As an alternative to clicking on the repr link.

Roughly

```python
def open_dashboard(self):
    info = client.scheduler_info()
    base = info['address']
    port = info['services']['bokeh']

    # there's probably already a way to get this URL.
    url = "http://{}:{}/status".format(urlparse(base).netloc.split(":")[0],
                                       port)
    webbrowser.open(url)
```

mrocklin:
This might be nice for tutorials

martindurant:
We found in tutorials that the URL reported by the kubernetes cluster object and the client did not match, that the former was correct - presumably because the client doesn't know it's running inside a cluster. Maybe the parallel discussions on relative addresses fixed that, but think it needs pointing out here.

TomAugspurger:
This approach will have the same limitations as the link in
`Client._repr_html_`.

Personally, I would use it for debugging sessions when I'm repeatedly
creating new clients / clusters, and would like to have a dashboard open
while the computation is running.

jakirkham:
@martindurant, have struggled with this as well. Lately my solution to that problem is [`nbserverproxy`]( https://github.com/jupyterhub/nbserverproxy ), which proxies the scheduler port under the Jupyter Notebook service. Running `dask.config.set({"distributed.dashboard.link": "/proxy/{port}/status"})` sets the URL to the scheduler relative to the current URL and seems to work well regardless of the particular network config (in my limited experience). Would be curious to know if this works for you.

mrocklin:
Conversation in this issue has separated into two equally valid points:

1.  We should have a method to open the dashboard from the client
2.  Finding the right address can sometimes be difficult

I think that conversation should continue mostly around point 1, which people seem to think is a nice idea.  Marking this as a good first issue.

douglasdavis:
Pinging this slightly old issue -- should this still be considered a feature request? If so, @TomAugspurger do you think the PR you started is a still useful to draw some inspiration?

TomAugspurger:
I think that we now have `Client.dashboard_link`. So in theory `webbrowser.open(client.dashboard_link)` should work just fine.

jakirkham:
Also if this URL is rendered in a notebook, Jupyter will create a link. So that itself might be enough

martindurant:
[...]

PR #2728, opened 2019-05-24
Add helper to open dashboard

TomAugspurger:
Closes https://github.com/dask/distributed/issues/2083

This shares all the limitations with the dashboard URL in the `Client._repr_html_`. Additionally, it won't work on things like `examples.dask.org` because the client is running on the remote machine. I still find it useful for local debugging.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -22,6 +22,7 @@ import six
 import socket
 import warnings
 import weakref
+import webbrowser
 
 import dask
 from dask.base import tokenize, normalize_token, collections_to_dsk
@@ -785,7 +786,21 @@ class Client(Node):
         else:
             return "<%s: not connected>" % (self.__class__.__name__,)
 
-    def _repr_html_(self):
+    @property
+    def _dashboard_address(self):
+        scheduler, info = self._maybe_get_scheduler_and_info()
+        if info and "bokeh" in info["services"]:
+            protocol, rest = scheduler.address.split("://")
+            port = info["services"]["bokeh"]
+            if protocol == "inproc":
+                host = "localhost"
+            else:
+                host = rest.split(":")[0]
+
+            template = dask.config.get("distributed.dashboard.link")
+            return template.format(host=host, port=port, **os.environ)
+
+    def _maybe_get_scheduler_and_info(self):
         if (
             self.cluster
             and hasattr(self.cluster, "scheduler")
@@ -803,7 +818,10 @@ class Client(Node):
         else:
             info = False
             scheduler = self.scheduler
+        return scheduler, info
 
+    def _repr_html_(self):
+        scheduler, info = self._maybe_get_scheduler_and_info()
         if scheduler is not None:
             text = (
                 "<h3>Client</h3>\n" "<ul>\n" "  <li><b>Scheduler: </b>%s\n"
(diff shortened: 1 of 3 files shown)


distributed/client.py
```diff
@@ -3864,6 +3876,19 @@ def _get_task_stream(
         else:
             raise gen.Return(msgs)
 
+    def open_dashboard(self):
+        """
+        Open the distributed dashboard.
```
mrocklin:
```suggestion
        Open the dashboard.
```

I'm trying to remove the name `distributed` from things.  I think that it's not informative for new users.

distributed/client.py
```diff
@@ -785,7 +786,21 @@ def __repr__(self):
         else:
             return "<%s: not connected>" % (self.__class__.__name__,)
 
-    def _repr_html_(self):
+    @property
+    def _dashboard_address(self):
+        scheduler, info = self._maybe_get_scheduler_and_info()
+        if info and "bokeh" in info["services"]:
```
mrocklin: Note that this is changing to `"dashboard"` in https://github.com/dask/distributed/pull/2724

distributed/tests/test_client.py
```diff
@@ -5236,6 +5237,24 @@ def test_diagnostics_link_env_variable(loop):
...
+            scheduler_kwargs={"services": {("bokeh", 12355): BokehScheduler}}
+        ) as (s, [a, b]):
+            with Client(s["address"], loop=loop) as c:
+                c.open_dashboard()
+        assert m.call_count == 1
+
+    with cluster() as (s, [a, b]):
+        with Client(s["address"], loop=loop) as c:
+            with pytest.raises(ValueError):
+                c.open_dashboard()
```
mrocklin: Regarding testing, I generally prefer to use `gen_cluster` over `with cluster` when possible.  One reason for this is speed.  A gen_cluster call takes around 100ms while a `with cluster` call takes around a second.

So I might do this as follows:

```python
@gen_cluster(client=True, scheduler_kwargs={...})
def test_open_dashboard(c, s, a, b):
    with mock.patch(...)  # or maybe we can't mock after object creation?
        c.open_dashboard()

@gen_cluster(client=True)
def test_open_dashboard_no_dashboard(c, s, a, b):
    with pytest.raises(ValueError):
       c.open_dashboard()
```

mrocklin: What is here works too and people new to working on this repository tend to prefer `with cluster` style tests.  It seems like you're interested in getting more into the dask.distributed codebase though, so I thought I'd push you towards gen_cluster.  It is, in many ways, much nicer overall.

distributed/client.py
mrocklin:
```suggestion
        if info and "dashboard" in info["services"]:
```

distributed/client.py
```diff
@@ -785,7 +786,21 @@ def __repr__(self):
...
+            protocol, rest = scheduler.address.split("://")
+            port = info["services"]["bokeh"]
```
mrocklin:
```suggestion
            port = info["services"]["dashboard"]
```

distributed/tests/test_client.py
```diff
@@ -5236,6 +5237,24 @@ def test_diagnostics_link_env_variable(loop):
                 assert link in text
 
 
+def test_open_dashboard(loop):
+    pytest.importorskip("bokeh")
+    from distributed.bokeh.scheduler import BokehScheduler
+
+    with mock.patch("distributed.client.webbrowser.open") as m:
+        with cluster(
+            scheduler_kwargs={"services": {("bokeh", 12355): BokehScheduler}}
```
mrocklin:
```suggestion
            scheduler_kwargs={"services": {("dashboard", 12355): BokehScheduler}}
```

distributed/tests/test_client.py
```diff
@@ -5236,6 +5237,24 @@ def test_diagnostics_link_env_variable(loop):
                 assert link in text
 
 
+def test_open_dashboard(loop):
+    pytest.importorskip("bokeh")
+    from distributed.bokeh.scheduler import BokehScheduler
```
mrocklin:
```suggestion
    from distributed.dashboard import BokehScheduler
```

jakirkham:
@TomAugspurger, do you have any thoughts on Matt's suggestions above?

TomAugspurger:
I think I'm going to close this. It's not actually what I want.

This was motivated by wanting to rapidly iterate on on scheduling issue. I want a way to view the memory usage of the cluster over time, and the dashboard seemed like the easiest way to get that.

What I think would actually solve my issue is something like a `Client.get_memory_stream()`, similar to `Client.get_task_stream`, which gives a history of the memory usage over time. I have another branch starting on this.

Issue #2716, opened 2019-05-20
Persist-call ignores assigned worker resources


Spiegel0:
When ```.persist(resources={...})``` is called on a Dask DataFrame, the specified resources will be ignored for scheduling. In particular, no task in the corresponding dashboard pages will outline the resources. In my setup, the following example triggers the issue (using python 3.6 and distributed 1.28.1):

```python
import pandas as pd

import dask.dataframe as dd
from dask.distributed import Client


def test_resources():
    client = Client(n_workers=1, resources={"MyRes": 1000})

    test_input_frame = dd.from_pandas(pd.DataFrame({"A": [1, 2], "B": [3, 4]}), npartitions=1)
    test_output_frame = test_input_frame.apply(lambda row: row.sum(), axis=1)

    output = test_output_frame.persist(resources={"MyRes": 1})
    print(output)

    # The task pages of the dashboard do not display the resource dependencies.
    input("Press any key to exit...")
    client.close()


if __name__ == '__main__':
    test_resources()
```

I quickly ran the debugger on the Dask distributed code and discovered that in ```client.Client._graph_to_futures(...)```, ```self._expand_resources(...)``` returns a dict where the keys are tuples identifying the task. When I manually convert the keys to their string representation, i.e. call:
```python
resources = {tokey(k): v for k,v in resources.items()}
```
after ```_expand_resources(...)```, the task dashboard will display the associated resources. (But I don't know whether converting the keys will have any other side effects)

TomAugspurger:
Sorry for the lack of response here. Can you try your example (which was great btw) on distributed master? This may be fixed. Locally, I see

<img width="953" alt="Screen Shot 2019-05-28 at 6 35 29 AM" src="https://user-images.githubusercontent.com/1312546/58475154-cec62f80-8112-11e9-9ca7-dc96d6c43985.png">

jakirkham:
@Spiegel0, have you had a chance to restest this on `master`? Please let us know if there is a way to help.

Spiegel0:
Thank you very much for your response! I tried the example on distributed master (commit d202e62) but the issue is still there. In my worker's page, the resources are also listed but on the task's page, they are not:
![2019-05-29-worker-screenshot](https://user-images.githubusercontent.com/8793444/58623614-a97e2080-82ce-11e9-8928-8e9e1c8bdb94.png)
[...]

PR #2733, opened 2019-05-30
Fix the resource key representation before sending graphs (#2716)

Spiegel0:
Convert resource key tuples to a string representation before they are submitted to the scheduler.

The commit is intended demonstrate a quick-fix of #2716. I'm not sure whether it breaks something else because I wasn't able to successfully run all test-cases, yet.


mrocklin:
Thanks for the PR @Spiegel0 !  It would be good to include a test in this PR that would have failed before this fix.  There is some information here: http://distributed.dask.org/en/latest/develop.html#writing-tests

Perhaps something like the following, with the example that you shared in the original issue filled in.

```python
@gen_cluster(client=True)
def test_resources_str(c, s, a, b):
    a.set_resources({"MyRes": 1000})
    df = ...
    df.persist(resouces=...)
    yield wait(df)

    ts = s.tasks[tokey(df.__dask_keys__()[0])]
    assert ts.resources == 
```

Spiegel0:
You are right, a test case would be very beneficial. Thank you for providing me the code snippet! I currently don't have access to my development environment but I'll try to implement the test case on Monday or Tuesday. (I guess its a good opportunity to get started with the Dask distributed internals.) I'll let you know as soon as I have the test case ready.

jakirkham:
Would you have another chance to take a look, @mrocklin? Looks like there is now a test case. Travis CI is passing (with the exception of Python 2.7, which we may not care about here). AppVeyor is failing, but it is unclear (at least to me) if the failures are related.

mrocklin:
Looks good.  Merging.  Thanks @Spiegel0 !

Spiegel0:
Thank you very much for the assistance and for merging the PR @mrocklin, @jakirkham and @TomAugspurger!

Issue #1851, opened 2018-03-22
Scheduler behaves badly when adding workers to meet resource demand 


mattilyra:
There is an issue with how the scheduler assigns tasks from the `unrannable` queue to workers who meet the resource requirements joining the scheduler.

The use case is some long running complex task where some tasks require an expensive resource, say GPUs, but those resources are only provided (through `Adaptive`) once the tasks requiring those resources are ready to be run. Say we come to a point in the computation where 5 tasks could be run, if GPUs were available. Managing the `scale_up` behaviour through adaptive is fairly straightforward and allows adding new compute nodes (for instance on AWS) with the required resources. The problem appears when the first of the new workers connects.

`Scheduler.addWorker` will go through the list of `unrunnable` tasks and check if there are workers that meet the requirements, since only the first worker has thus far connected there is only one worker that meets the requirements (some possibly positive number of workers may or may not booting up and joining shortly, but that hasn't happened yet).

            for ts in list(self.unrunnable):
                valid = self.valid_workers(ts)
                if valid is True or ws in valid:
                    recommendations[ts.key] = 'waiting'

The task goes through `released -> waiting` and then `waiting -> processing`, the `transition_waiting_processing` again calls `valid_workers` to get a list of workers where the task(s) can be run (this list still just contains a single worker because the other ones haven't yet connected).

The end result of all of this is that the worker who happens connect first and have the resource required by the tasks gets all the tasks dumped onto it with all the other workers, who potentially connect just seconds later, get nothing and are shutdown by the scheduler because they are idling.

In short, it appears to be the case that the purpose of the `resource_requirements` is to act as a hint of required peak performance (memory, GPU, whatever) from the workers, and not to be a dynamically changing resource allocation. Is this the case, and is there any interested in changing that? The resources available and resources consumed are taken into account in `transition_waiting_processing`, but only on the `worker_state` not for the scheduler in general.

If this is not the intended behaviour and should be fixed, I'm more than happy to work on this.

mrocklin:
> If this is not the intended behaviour and should be fixed, I'm more than happy to work on this.

This is not the intended behaviour and should be fixed.  Help would be welcome!

I agree with your understanding of the situation and in general I'm glad to hear from someone who is comfortable working within the scheduler code.  

My first instinct on how to solve this problem would be to look at improving work stealing with resources.  The first worker *should* get all of the tasks if its the only one around, but it should also share them once others arrive.  

[...]

PR #2740, opened 2019-05-31
Allow tasks with restrictions to be stolen

calebho:
Addresses stealing tasks with resource restrictions, as mentioned in #1851.
If a task has hard restrictions, do not just give up on stealing.
Instead, use the restrictions to determine which workers can steal it
before attempting to execute a steal operation.

A follow up PR will be needed to address the issue of long-running tasks
not being stolen because the scheduler has no information about their
runtime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ continuous_integration/hdfs-initialized
 .pytest_cache/
 dask-worker-space/
 .vscode/
+*.swp
+.ycm_extra_conf.py
+tags
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -333,7 +333,7 @@ class WorkStealing(SchedulerPlugin):
                 saturated = [
                     ws
                     for ws in saturated
-                    if combined_occupancy(ws) > 0.2 and len(ws.processing) > ws.ncores
+                    if combined_occupancy(ws) > 0.2 and len(ws.processing) > ws.nthreads
                 ]
             elif len(s.saturated) < 20:
                 saturated = sorted(saturated, key=combined_occupancy, reverse=True)
@@ -379,7 +379,7 @@ class WorkStealing(SchedulerPlugin):
                             continue
                         if combined_occupancy(sat) < 0.2:
                             continue
-                        if len(sat.processing) <= sat.ncores:
+                        if len(sat.processing) <= sat.nthreads:
                             continue
 
                         i += 1
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -430,12 +430,12 @@ def _can_steal(ws, what, from_):
     ``from_``.
     """
     if what.loose_restrictions:
-        logger.debug("Task {} has loose restrictions".format(what.key))
+        logger.debug("Task %s has loose restrictions", what.key)
         return True
     elif not (
         what.host_restrictions or what.worker_restrictions or what.resource_restrictions
     ):
-        logger.debug("Task {} has no restrictions".format(what.key))
+        logger.debug("Task %s has no restrictions", what.key)
         return True
 
     if (
@@ -443,16 +443,16 @@ def _can_steal(ws, what, from_):
         and get_address_host(ws.address) not in what.host_restrictions
     ):
         logger.debug(
-            "Candidate thief {} does not satisfy host restrictions {}".format(
-                ws.address, what.host_restrictions
-            )
+            "Candidate thief %s does not satisfy host restrictions %s",
+            ws.address,
+            what.host_restrictions,
         )
         return False
     elif what.worker_restrictions and ws.address not in what.worker_restrictions:
         logger.debug(
-            "Candidate thief {} does not satisfy worker restrictions {}".format(
-                ws.address, what.worker_restrictions
-            )
+            "Candidate thief %s does not satisfy worker restrictions %s",
+            ws.address,
+            what.worker_restrictions,
         )
         return False
 
@@ -468,16 +468,15 @@ def _has_resources(ws, required_resources):
         try:
             supplied = ws.resources[resource]
         except KeyError:
-            logger.debug(
-                'Worker {} does not have resource "{}"'.format(ws.address, resource)
-            )
+            logger.debug('Worker %s does not have resource "%s"', ws.address, resource)
             return False
         else:
             if supplied < value:
                 logger.debug(
-                    "Worker {} has fewer resources ({}) than required ({})".format(
-                        ws.address, supplied, value
-                    )
+                    "Worker %s has fewer resources (%s) than required (%s)",
+                    ws.address,
+                    supplied,
+                    value,
                 )
                 return False
     return True
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -6,6 +6,7 @@ from math import log
 from time import time
 
 import dask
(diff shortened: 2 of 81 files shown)


jakirkham:
@TomAugspurger, would you have a chance to look at this PR?

.gitignore
```diff
@@ -13,3 +13,6 @@ continuous_integration/hdfs-initialized
 .pytest_cache/
 dask-worker-space/
 .vscode/
+*.swp
```
TomAugspurger: This seems fine since we include it for other files, but I typically put these files under a global gitignore.

distributed/stealing.py
```diff
@@ -417,4 +425,62 @@ def story(self, *keys):
         return out
 
 
+def _can_steal(ws, what, from_):
+    """Determine whether worker ``ws`` can steal task ``what`` from worker
+    ``from_``.
+    """
+    if what.loose_restrictions:
+        logger.debug("Task {} has loose restrictions".format(what.key))
```
TomAugspurger: Typically logging uses old style string formatting, and the values are passed to `logger.debug`, .e.g


```
logger.debug("Task %s has loose restrictions", what.key)
```

distributed/stealing.py
```diff
@@ -417,4 +425,62 @@ def story(self, *keys):
         return out
 
 
+def _can_steal(ws, what, from_):
```
TomAugspurger: So this is going to be called relatively often. AFAICT, it'll be called once per task, thief-worker pair for every task we're considering stealing. Given that, we should ensure that it's fast. How confident are you that there aren't any surprising performance issues here? Most things look OK, but may want to look into `get_address_host`.

- `get_address_host`

calebho: In general, `get_address_host` is linear in the length of the input address, however I think it's safe to assume that `len(ws.address)` is small and bounded as a function of `ws`. 

The four collections used here are 
1. `what.host_restrictions`,
2. `what.worker_restrictions`,
3. `ws.resources`, and
4. `from_.resources`.

Key containment is checked using (1) and (2), and since they are sets these operations are constant with respect to the size of the containers. (3) and (4) are used in `_has_resources`. The work in the loop only uses (3). Because (3) is a dict, value retrieval is constant with respect to the size of (3). The loop iterates over (4) once, so the total work in this function is linear in the size of (4). So overall, `_can_steal` is O(`len(from_.resources)`).

mrocklin: Yeah, I'm also concerned about the performance here.  

This code gets called *often*.  Previously we didn't iterate through `idle`, we just indexed into it (indeed we made it a `SortedSet` object just to get this constant-time access). 

The work stealing code accounts for a non-trivial amount of the scheduler performance costs.  I would not be surprised if adding this check significantly increases those costs.  Given that resources aren't used in the common case I'm very hesitant to go this route without strong demonstration that it doesn't affect performance.

*However*, I would be much more comfortable with this if it was only engaged in the uncommon case that the task had resource constraints.  It could be that a simple test like the following resolves the issue.

```python
if ts.resources:
    theives = [...]
    theif = ...
else:
    thief = self.idle[i * len(self.idle)]
```

Then this PR is essentially a no-op in the common case, and we can be a lot more relaxed about performance.

jakirkham:
@calebho, any thoughts on Tom and Matt's suggestions above? 🙂

calebho:
So I was investigating another issue related to work stealing where all tasks were running on single worker leaving all other workers idle. Here's a demonstrative example:
```python
import time
from dask_jobqueue import SLURMCluster
from tempfile import TemporaryDirectory
import os


def long_identity(i):
    time.sleep(15)
    return i


def main():
    world_size = 5
    ntasks = 5
    with TemporaryDirectory(dir=os.getcwd()) as local_dir:
        print(local_dir)
        cluster = SLURMCluster(
            name='testing',
            queue='dev',
            cores=10,  # set to at most `ntasks` to observe work stealing
            memory='4GB',
            local_directory=local_dir,
            log_directory=local_dir,
            extra=['--resources "TASKS=1"'],
            silence_logs=False,
        )
        cluster.adapt(minimum=world_size, maximum=world_size)
        cl = Client(cluster)
        futs = []
        for i in range(ntasks):
            futs.append(cl.submit(long_identity, i, resources={'TASKS': 1}))

        for f in futs:
            f.result()


if __name__ == '__main__':
    main()
```
The interesting thing about this example is if you use more cores than there are tasks, then no work stealing occurs. If you set cores to at most `ntasks`, then work stealing occurs. After some digging, I think this is because the scheduler doesn't believe the worker holding all the tasks is saturated, even though all of its resources are consumed by a single task. As a result, the tasks are processed serially.  Looking into how `Scheduler.saturated` is computed, if the size of `WorkerState.processing` is less than `WorkerState.ncores`, then the worker is considered idle (and thus not saturated). This is a problem because many ML workflows are expressed in this manner: few long running tasks which use some number of GPUs. The number of CPUs allocated for these tasks is in general less important (just allocate enough to fully saturate the GPUs).

This gave me another idea for how to solve the current issue which would also handle the problem above; what if `Scheduler.saturated` and `Scheduler.idle` were changed to take into account the resources of the currently processing tasks?  The downside is that determining whether a worker is saturated or not now becomes much more complicated. For example, say a worker has two resources `{'foo': 2, 'bar': 4}`. What does it mean for this worker to be saturated? All resources depleted seems reasonable, but I think it would interesting to explore whether the question should instead be "Is this worker saturated with respect to these resources?". Before I get too far down this rabbit hole, I wanted to pause and sanity check with everyone here.

calebho:
> As an example, lets imagine that we have 1000 workers and 10,000,000 tasks. I think that we call balance every 100ms and we'd like stealing to take up less than 10% of our CPU scheduler overhead. My guess is that we don't currently stay under budget when there are restrictions, and my guess is that a lot of it has to do with this function. I think that it would be good to take a hard look at this function and reduce costs. Some of this might be through things like caching, but I'm not sure of the best approach here.

If this is the concern, then I don't think `_can_steal` is the issue here because it's linear in the number of victim resources and this number is generally small (< 10 seems reasonable to assume). Even if it somehow runs in constant time w.r.t. the number of victim resources it won't matter if it's being called 10 billion times.

Looking at [this if-else block](https://github.com/dask/distributed/pull/2740/commits/28b24b985b0c18cd5d7d45ff452bf78d35160213#diff-39a0814a3ec005aa2078cb3de3738c4dR352), one way to get around this cost might be to sample a constant number of workers from idle instead of using the entire population. The issue then becomes how to construct an appropriate distribution over the idle workers such that
1) the expected number of workers in the sample which can steal the given task is high, and
2) constructing such a sample scales w.r.t. the number of workers.

martindurant:
@mrocklin , @TomAugspurger , I am not following this discussion, but the OP seems to have replied to at least some concerns - did you want more specific information or work?

martindurant:
ah , @calebho , there seems to be a conflict now - sorry about that. Can you please merge from master?

calebho:
Oops I might have inadvertently polluted this PR with a bunch of commits from upstream because I rebased. I guess I can remake the branch on my fork?

martindurant:
Sure, start a new PR, if more convenient.

jrbourbeau:
Closing as this PR was superseded by https://github.com/dask/distributed/pull/3069. Thanks @calebho!

Issue #2660, opened 2019-05-06
Cannot maintain effective debug-level on distributed (parent) logger with local scheduler/cluster


javabrett:
With `config.yaml` including simple-form `logging` config:

```
logging:
  distributed: debug

>>> import logging
>>> import distributed
>>> logging.getLogger("distributed")
<Logger distributed (DEBUG)>
>>> logging.getLogger("distributed").handlers
[<StreamHandler <stderr> (NOTSET)>]
>>> logging.getLogger("distributed").debug("debug")
distributed - DEBUG - debug
>>> 
```

All looks good.  Start a `Client`:

```
>>> from distributed import Client
>>> c = Client()
>>> logging.getLogger("distributed")
<Logger distributed (DEBUG)>
>>> logging.getLogger("distributed").handlers
[<StreamHandler <stderr> (WARNING)>]
>>> logging.getLogger("distributed").debug("debug")
>>> 
```
No logging output, as the `distributed` logger has had its `Handler` (a `StreamHandler`) set to `WARNING` threshold.

Once a `Client` is started and fires-up a local cluster/scheduler, I suspect a worker:

https://github.com/dask/distributed/blob/09b959a5667a51a2dc073510c784e01d44827457/distributed/worker.py#L424-L425

... runs `silence_logging`, which does just as it is named:

https://github.com/dask/distributed/blob/09b959a5667a51a2dc073510c784e01d44827457/distributed/utils.py#L728-L742

See also #2659 .  It might be argued that nothing important is logging to the `distributed` logger, and everything is logging to a child e.g. `distributed.client` (even through the loggers are flattened with the simple config), which might be true, but it further erodes the notion of hierarchical logging.

mrocklin:
@jcrist might have thoughts here

GenevieveBuckley:
Cross referencing this broader discussion on logging: https://github.com/dask/distributed/issues/2659

Issue #2659, opened 2019-05-05
Old/simple/default logging config is opaque and hard to externally-configure


javabrett:
This issue might become either a doc-improvement one, or possibly lead to some non-backwards-compatible change in defaults.

# Summary, TL;DR

Dask `distributed` (and specifically client-side) logging should behave roughly according to [Configuring Logging for a Library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library), avoiding application-level logging-configuration decisions.  The default/absent logging configuration does not do this and configures a custom (flat, non-hierarchical) logging structure with specific `Handler`s.  There is an option to use the newer `logging` `dictConfig` schema, but this brings other considerations.

# Detail

I have an application which loads a number of libraries, including "foo", and the Dask `distributed` `Client`.  The expectation I can easily configure application logging externally, including logging from `foo` and (Dask) `distributed` and any other library.  Providing they are all good library logging citizens then this should be easy and not require special knowledge.  In actuality I need to treat the Dask `distributed` client library specifically, and have special knowledge of its logging-configuration, so it can't be treated globally as just another library.

Imagine we are running in an application's `config.py` - or we might be configuring logging in a notebook via `ipython_kernel_config.py`.  The following could also be transposed into the `logging` module YAML schema.  Testing in `daskdev/dask:latest` (`docker run -it --rm daskdev/dask python`:

```
>>> import logging
>>> from logging import StreamHandler, Formatter
```

We want all loggers to have a standard format, so we can change the formatter on the root logger's handler:

```
>>> root_handler = logging.StreamHandler()
>>> root_handler.setFormatter(Formatter("[%(levelname)1.1s %(asctime)s.%(msecs).03d %(name)s %(module)s:%(lineno)d] %(message)s"))
>>> logging.root.handlers = [root_handler]
```

Test what the `foo` logger output will look like:

```
>>> foo_logger = logging.getLogger("foo")
>>> foo_logger.warning("foo warning")
[W 2019-05-05 22:53:43,628.628 foo <stdin>:1] foo warning
```

All good so far.  Check Dask `distributed` logging:

```
>>> import distributed
>>> distributed_logger = logging.getLogger("distributed")
>>> distributed_logger.warning("distributed warning")
distributed - WARNING - distributed warning
```

Not the format we are after.  At this point we check `distributed` log config/init code and also check:

```
>>> distributed_logger.handlers
[<StreamHandler <stderr> (NOTSET)>]
>>> distributed_logger.propagate
[...]

PR #2744, opened 2019-06-03
Add nanny logs

TomAugspurger:
I had a need to collate logs from the cluster, and would like to include logs from the Nanny. This adds a method to get logs from the nanny as well. 

This isn't quite ready.

1. The keys in the returned dict are the address of the workers, not the nannies. Do we have a convention here? I assume we want the address of the nannies.
2. I may be adding the handler to the `distribted.nanny.logger` multiple times, resulting in duplicate logs. Need to verify what's going on.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2270,6 +2270,10 @@ class Client(Node):
         wait: boolean (optional)
             If the function is asynchronous whether or not to wait until that
             function finishes.
+        nanny : bool, defualt False
+            Whether to run ``function`` on the nanny. By default, the function
+            is run on the worker process.  If specified, the addresses in
+            ``workers`` should still be the worker addresses, not the nanny addresses.
 
         Examples
         --------
@@ -3364,7 +3368,7 @@ class Client(Node):
         """
         return self.sync(self.scheduler.logs, n=n)
 
-    def get_worker_logs(self, n=None, workers=None):
+    def get_worker_logs(self, n=None, workers=None, nanny=False):
         """ Get logs from workers
 
         Parameters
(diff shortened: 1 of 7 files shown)


mrocklin:
> Do we have a convention here? 

Other functions like `run` and broadcast take a `nanny=` keyword argument.  That might be appropriate here as well.

mrocklin:
So `client.get_worker_logs(nanny=True)`

In fact, I suspect that on the scheduler you could just pass through the `nanny=nanny` keyword to `Scheduler.broadcast` and everything would work out.

TomAugspurger:
Thanks @mrocklin that works out quite nicely.

I found similar behavior for Client.run. I updated the docstring there to specify that `workers` should still be the *worker* addresses when you specify `nanny=True`, not the nanny address.

TomAugspurger:
As an aside, having timestamps in logs is quite helpful for collation across the cluster (assuming the clocks synced well enough). What's our backwards compatibility story for configuration things like that?

TomAugspurger:
Actually, this may not be ready. I thought I had sorted out the duplicate logging, but apparently not.

mrocklin:
Other than the possible duplicate handler issue the implementation here looks great to me.

mrocklin:
> As an aside, having timestamps in logs is quite helpful for collation across the cluster (assuming the clocks synced well enough). What's our backwards compatibility story for configuration things like that?

I don't have any particular thoughts here.  

You may want to take a look here though:

-  https://github.com/dask/distributed/issues/2660
-  https://github.com/dask/distributed/issues/2659

mrocklin:
> Actually, this may not be ready. I thought I had sorted out the duplicate handler, but apparently not.

Can you explain this a bit more?  This seems to be the same behavior as before.  Is this correct?

TomAugspurger:
I'm not too sure what's going on.

```python
   ...: import logging
   ...:
   ...:
   ...: client = Client(n_workers=2, threads_per_worker=1)
   ...:
   ...: logger = logging.getLogger('distributed.nanny')
   ...: logger.handlers

Out[1]:
[<StreamHandler <stderr> (WARNING)>,
 <DequeHandler (NOTSET)>,
 <DequeHandler (NOTSET)>]
```

So if I get the `distributed.nanny` logger from my main processes, I see one handler per Nanny.

Oh... but I just realized this is the same behavior as `distributed.worker` with `processes=False`.

```python
   ...: client = Client(n_workers=2, threads_per_worker=1, processes=False)
   ...:
   ...: logger = logging.getLogger('distributed.worker')
   ...: logger.handlers
Out[2]: [<DequeHandler (NOTSET)>, <DequeHandler (NOTSET)>]
```

so perhaps we're OK?

mrocklin:
Looks good to me.   Merging.

PR #2745, opened 2019-06-03
Add async context managers to scheduler/worker classes

mrocklin:


mrocklin:
cc @jcrist, you were asking about cleaning up start/stop

PR #2758, opened 2019-06-07
Replace ncores with nthreads throughout codebase

mrocklin:
As long as we're breaking things, we might as well break some more.

In some places we use nthreads, in others ncores.  This moves almost all use cases to nthreads, which is a little bit more literally correct.


mrocklin:
Merging tomorrow if there are no comments

PR #2761, opened 2019-06-07
Delay lookup of allowed failures.

TomAugspurger:
This allows for setting the config after importing distributed

xref https://github.com/dask/dask-examples/pull/75#discussion_r291141404

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -73,8 +73,6 @@ from .variable import VariableExtension
 logger = logging.getLogger(__name__)
 
 
-ALLOWED_FAILURES = dask.config.get("distributed.scheduler.allowed-failures")
-
 LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
 DEFAULT_DATA_SIZE = dask.config.get("distributed.scheduler.default-data-size")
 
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -846,7 +844,9 @@ def __init__(
         self._setup_logging(logger)
 
         # Attributes
-        self.allowed_failures = allowed_failures
+        self.allowed_failures = allowed_failures or dask.config.get(
```
mrocklin: We should watch out for `allowed_failures=0`, which is a reasonable input here.

distributed/tests/test_scheduler.py
```diff
@@ -1607,3 +1607,13 @@ def test_dashboard_address():
...
+
+
+@pytest.mark.asyncio
+async def test_allowed_failures_config():
+    async with Scheduler(port=0, allowed_failures=10) as s:
+        assert s.allowed_failures == 10
+
+    with dask.config.set({"distributed.scheduler.allowed_failures": 100}):
+        async with Scheduler(port=0) as s:
+            assert s.allowed_failures == 100
```
mrocklin: Looks good to me.  Thanks @TomAugspurger

PR #2765, opened 2019-06-10
Task scheduling with memory pressure

### LINKED ISSUE #2602: an example that shows the need for memory backpressure
In my work with large climate datasets, I often concoct calculations that cause my dask workers to run out of memory, start dumping to disk, and eventually grind my computation to a halt. There are many ways to mitigate this by e.g. using more workers, more memory, better disk-spilling settings, simpler jobs, etc. and these have all been tried over the years with some degree of success. But in this issue, I would like to address what I believe is the root of my problems within the dask scheduler algorithms.

The core problem is that the tasks early in my graph generate data faster than it can be consumed downstream, causing data to pile up, eventually overwhelming my workers. Here is a self contained example:

```python
import dask.array as dsa

# create some random data
# assume chunk structure is not under my control, because it originates
# from the way the data is laid out in the underlying files
shape = (500000, 100, 500)
chunks = (100, 100, 500)
data = dsa.random.random(shape, chunks=chunks)

# now rechunk the data to permit me to do some computations along different axes
# this aggregates chunks along axis 0 and dis-aggregates along axis 1
data_rc = data.rechunk((1000, 1, 500))
FACTOR = 15


def my_custom_function(f):
    # a pretend custom function that would do a bunch of stuff along
    # axis 0 and 2 and then reduce the data heavily
    return f.ravel()[::15][None, :]

# apply that function to each chunk
c1 = math.ceil(data_rc.ravel()[::FACTOR].size / c0)
res = data_rc.map_blocks(my_custom_function, dtype=data.dtype,
                         drop_axis=[1, 2], new_axis=[1], chunks=(1, c1))

res.compute()
```

(Perhaps this could be simplified further, but I have done my best to preserve the basic structure of my real problem.)

When I watch this execute on my dashboard, I see the workers just keep generating data until they reach their memory thresholds, at which point they start writing data to disk, before `my_custom_function` ever gets called to relieve the memory buildup. Depending on the size of the problem and the speed of the disks where they are spilling, sometimes we can recover and manage to finish after a very long time. Usually the workers just stop working.

This fail case is frustrating, because often I can achieve a reasonable result by just doing the naive thing:
```
for n in range(500):
    res[n].compute()
```
and evaluating my computation in serial.

I wish the dask scheduler knew to stop generating new data before the downstream data could be consumed. I am not an expert, but I believe the term for this is [backpressure](https://www.quora.com/What-is-backpressure-in-the-context-of-data-streaming). I see this term has come up in https://github.com/dask/distributed/issues/641, and also in [this blog post](http://matthewrocklin.com/blog/work/2017/04/13/streaming) by @mrocklin regarding streaming data.

I have a hunch that resolving this problem would resolve many of the pervasive but hard-to-diagnose problems we have in the xarray / pangeo sphere. But I also suspect it is not easy and requires major changes to core algorithms.

Dask version 1.1.4

TomAugspurger:
There are two main components to the PR

1. Tracking whether a task (prefix) tends to increase or decrease memory usage.
2. Adjust the scheduling logic to prefer memory-freeing tasks when the worker is
   under memory pressure.

Part 1 is pretty straightforward. The scheduler keeps a running some per
task-prefix. When a task finishes, we compare the output size to the size of
each freed input. This strategy is biased towards 0 (underestimate the amount of
memory consumed or freed by a task), but for scheduling we really only care
whether it's positive or negative.

Part 2 is... incomplete. I'm struggling with a reasonable way to do this.
AFAICT, there are two main problems we might address:

- a: We are low on memory, and have tasks that we think will free memory. We
     re-prioritize to run those, rather than what dask.order woudl do (right now
     we also ignore user-provided priorities, which is bad).
- b: We are low on memory, and are currently running tasks that we think will
     free memory. We'd like to just not schedule new tasks.

This isn't ready yet, but it's a start. Aside from general cleanups, I'm primarily concerned with this alternative scheduling logic absolutely breaking everything :) This feels like an invasive change.

[This notebook](https://gist.github.com/eb30e78aab84f8e957ffe247d07f4790) has a couple toy problems I've been testing this with.

xref #2602


distributed/scheduler.py
```diff
@@ -3466,9 +3473,11 @@ def _add_to_memory(
             s.discard(ts)
             if not s and not dts.who_wants:
                 recommendations[dts.key] = "released"
+                self.task_net_nbytes[ts.prefix] -= dts.nbytes
```
mrocklin: Should we divide by the number of dependents a dependency has?

If a large dependency has 100 dependents, they probably shouldn't all think that they're going to release a bunch of memory.

distributed/tests/test_scheduler.py
```diff
@@ -1609,6 +1609,36 @@ def test_dashboard_address():
...
+        "b-2": (lambda t, u: t[:10], "a-1", "a-2"),
+    }
+    x = c.get(dsk, "b-2", sync=False)
+    result = yield x
+    assert result == bytearray(10)
+    assert s.task_net_nbytes["a"] > 0
+    assert s.task_net_nbytes["b"] < 0
+
+    # now that we've learned some things, submit again, and
+    # check the workers.
```
mrocklin: So we're not going to learn this on initial large computations, correct?

TomAugspurger: That's unfortunately correct. Once (if) the initial large computation completes we'll think that other tasks with the same prefix are likely to be large.

I see two possible solutions

1. (Long term) task annotations could store the information we need, avoiding the need to learn for the initial computation.
2. (Short term) Provide an API for users to inform the scheduler, though I don't think this is a good idea.

distributed/scheduler.py
TomAugspurger: Let me describe what I hoped would happen, which may not be how I coded / understood things. If you have a graph like

```python
  
        a1    a2
       /  \  /  \
     b0    b1    b2
       \  /  \  /
        c1    c2
  
  ```

then `a1` can be freed when both b0 and b1 are done. Let's suppose b0 finishes before b1. Then, I think the way I've written it things go like

1. b0 finishes. The scheduler sees that the worker running `b1` still wants `a0`, so we don't even reach this section, there aren't any dependents to free. `task_net_bytes['b']` is still 0
2. b1 finishes. The scheduler sees that `a1` can now be freed, so `tast_net_bytes['b']` is decremented by `sizeof(a1)`.

TomAugspurger:
FYI, I plan to work on some better diagnostics for memory usage next, to more accurately measure what this is doing. Something like `Client.get_task_stream`, but for memory.

MikeAlopis:
Any updates on progress here @TomAugspurger?

TomAugspurger:
Been a bit busy lately, hoping to return to it tomorrow.

If you have use cases that run into this kind of memory pressure, it'd be nice to try out this alternative scheduling regime on them (when it's ready).

TomAugspurger:

Issue #2708, opened 2019-05-17
Identify and warn about starting processes in module


mrocklin:
Users who are new to Dask and possibly new to Python often run into this error message

```
distributed.utils - ERROR - 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

This comes from starting a new process in module code not protected by `if __name__ == '__main__':` 

This error message is not informative, and seems to stop more than a few users.  We probably can't affect this error message itself, nor the behavior (both are pretty deep in Python I suspect) but perhaps we can detect the situation before we trigger the error and raise a nicer error message?

How do we identify that we're starting a process in module code?

mrocklin:
You can create this error by running any of the following scripts (running from ipython won't reproduce)

```python
from dask.distributed import Client
client = Client()
```

```python
import multiprocessing
multiprocessing.get_context('spawn').Pool()
```

```
$ python myscript.py
```

jcrist:
This is unfortunately only catchable at runtime (i.e. after the workers start). Ideally we should see this error from the first failed worker and abort all pending workers, but this is complicated by the new `SpecCluster` class which needs to be general for some future other subclasses. We also block on all pending futures waiting for them to start, rather than handling their errors as they come up (meaning that for a 64 worker local cluster we'll see the error 64 times before we get to handle the first failure), and then drop the futures so we can't cancel any pending workers (something I mentioned in the SpecCluster PR: https://github.com/dask/distributed/pull/2675#discussion_r286559419).

I could handle this by special casing methods in LocalCluster, or by redesigning the async behavior of `SpecCluster`. I'm not sure which is better, or how critical this issue is - thoughts @mrocklin? Given no other direction I'd tend towards redesigning `SpecCluster`.

mrocklin:
> This is unfortunately only catchable at runtime (i.e. after the workers start).

That's fine with me.  I think that if we can give users a better error message then that would be awesome.

> I could handle this by special casing methods in LocalCluster, or by redesigning the async behavior of SpecCluster. I'm not sure which is better, or how critical this issue is - thoughts @mrocklin? Given no other direction I'd tend towards redesigning SpecCluster.

[...]

PR #2768, opened 2019-06-11
[WIP] Better track worker failures in SpecificationCluster

jcrist:
This refactores `SpecificationCluster` to allow responding to errors
raised during worker startup, allowing better error messages to be
reported to the user. All started tasks now have their results
inspected, removing `asyncio` default handling of logging background
errors.

The end goal of this is to be able to provide better user errors during
cluster startup failure (not necessarily cluster scale up errors).

Aims to address #2708.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -1,5 +1,6 @@
 import asyncio
 import atexit
+import logging
 import weakref
 
 from tornado import gen
@@ -7,6 +8,9 @@ from tornado import gen
 from .cluster import Cluster
 from ..utils import LoopRunner, silence_logging, ignoring
 from ..scheduler import Scheduler
+
+
+logger = logging.getLogger(__name__)
 
 
 class SpecCluster(Cluster):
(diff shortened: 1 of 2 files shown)


jcrist:
This currently restructures the code to allow tracking errors (the diff is larger than it needed to be, as I reordered some methods to group things - sorry). I'm currently struggling with debugging the Nanny process, the restart behavior seems and process management seems resistant to shutting down during these error cases. I'm going to take a break from this and come back later.

mrocklin:
@jcrist this is marked as WIP.  Are there aspects here that you're still uncomfortable with?

mrocklin:
Also, do you think that this should block a 2.0 release?

jcrist:
> Also, do you think that this should block a 2.0 release?

No, this shouldn't be a blocker.

>  Are there aspects here that you're still uncomfortable with?

Yeah, this still doesn't handle detecting the original issue (some logic in the nanny is complicating things). Hope to get back to this soon.

mrocklin:
Cool.  Thanks for surfacing for the update.

PR #2770, opened 2019-06-13
Fix diagnostics page for memory_limit=None

bnaul:
Similar to #2255: w/ no memory limit some parts of the diagnostics page don't work because of this formatting code. This is one way to fix it, or I guess you could change all the callers to pass `inf` instead...?

### DIFF
--- a/distributed/dashboard/templates/worker-table.html
+++ b/distributed/dashboard/templates/worker-table.html
@@ -16,7 +16,7 @@
         <td><a href="../worker/{{ url_escape(ws.address) }}.html">{{ws.address}}</a></td>
         <td> {{ ws.name if ws.name is not None else "" }} </td>
         <td> {{ ws.nthreads }} </td>
-        <td> {{ format_bytes(ws.memory_limit) }} </td>
+        <td> {{ format_bytes(ws.memory_limit) if ws.memory_limit is not None else "" }} </td>
         <td> <progress class="progress" value="{{ ws.metrics['memory'] }}" max="{{ ws.memory_limit }}"></progress> </td>
         <td> {{ format_time(ws.occupancy) }} </td>
         <td> {{ len(ws.processing) }} </td>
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Hmm I think that changing the callers is preferred.

mrocklin:
@bnaul do you understand @TomAugspurger 's request above?  I think that he would prefer to see code something like the following:

```python
if memory_limit is not None:
    text = format_bytes(memory_limit)
else:
    text = "--"
```

bnaul:
yep, I thought about doing it that way but this was a lot easier 🙃 will take another stab when I have some time

bnaul:
Updated to just change the callers. It looks like there are ~2 dozen calls to `format_bytes`; most of these have arguments that couldn't be `None`, but enough of them could be that it still feels a little shaky to me to have this helper that will fail completely if `None` is passed. I guess I could see an argument for failing faster, but for just printing text it seems like erring on the side of ignoring rather than raising might be better. Just my 2 cents :)

mrocklin:
No strong thoughts from me on this topic

mrocklin:
If we keep the second commit there is a small linting issue.  @bnaul we started using `black` recently.

```
pip install black
black distributed
git commit -a -m "Apply black"
```

@TomAugspurger given @bnaul 's comment above do you think that format_bytes should include a path for `None`?

@bnaul if @TomAugspurger says "sure" then please be aware that you'll probably want to change this upstream in dask, since we're moving this function to `dask/utils.py` shortly (see #2686 )

TomAugspurger:
I’d prefer format_bytes to raise on invalid input. I regularly accidentally pass an array rather than array.nbytes. I wouldn’t want that to fail silently.


________________________________

bnaul:
thanks for the lint @martindurant; since I switched it back to the way @TomAugspurger prefers (changing the callers), I don't think anything else needs to change upstream in `dask`. so this should be good I think? sorry I missed the release 😢

TomAugspurger:
We'll most likely be doing another release shortly.

PR #2783, opened 2019-06-18
Typo in bokeh service_kwargs for dask-worker

TomAugspurger:
Saw this while poking around dask-worker. I suspect this was missed in the bokeh -> dashboard migration because of the typo.

I've split into a separate PR, in case you request tests @mrocklin :) If so I'll put something together.

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -346,7 +346,7 @@ def main(
             host=host,
             port=port,
             dashboard_address=dashboard_address if dashboard else None,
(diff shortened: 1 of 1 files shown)


mrocklin:
Hrm, indeed.  Tests would be great, but obviously this hasn't made anyone sad enough to file an issue, so it's probably a lower priority than other things that you're working on.

mrocklin:
+1 from me

mrocklin:
Also, I have to say that it's great to have you rooting around in the distributed codebase

Issue #2799, opened 2019-06-24
Add compatibility for --no-bokeh option


TomAugspurger:
OK with this before 2.0 @mrocklin? Making a PR now in case (Pangeo specifies --no-bokeh).

mrocklin:
Yeah, I wasn't sure how best to provide a message around a True/False flag. In Python I would set this to None and then raise if it was not None.

I'm comfortable with adding it back in as long as it raises an annoying warning.

TomAugspurger:
> In Python I would set this to None and then raise if it was not None.

I was just running into that. I think we're fine. The only weird case is if the user does `--no-bokeh --dashboard`, or something like that, to raise. There's not a clean way to check if dashboard is explicitly set. You can do `default=None`, but that messes with other things. So I think we'll just take `--no-bokeh` in that case (after warning).

PR #2800, opened 2019-06-24
Deprecate --bokeh/--no-bokeh CLI

TomAugspurger:
Closes https://github.com/dask/distributed/issues/2799

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -78,6 +78,13 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     required=False,
     help="Launch the Dashboard",
 )
+@click.option(
+    "--bokeh/--no-bokeh",
+    "bokeh",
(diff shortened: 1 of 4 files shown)


mrocklin:
Looks great.  Thanks @TomAugspurger

mrocklin:
Hrm, the tests here seem to be unhappy

TomAugspurger:
Whoops, I was running the worker one twice.

I don't really see an easy way to do the test for the scheduler.

mrocklin:
> I don't really see an easy way to do the test for the scheduler.

I don't understand.  What is the concern here?

TomAugspurger:
The click-based version I was using seemed to just be waiting in the dask-scheduler command, waiting for workers to connect. I didn’t see a way to background the call, so that we could interrupt / end it.

I didn’t have luck using the style of tests in that file already. I don’t recall the issue I ran into.

Issue #2543, opened 2019-02-25
Windows Subsystem for Linux "Could not set timeout on TCP stream: [Errno 92] Protocol not available"


jwilson8767:
`distributed/comm/tcp.py` gives the warning "Could not set timeout on TCP stream: [Errno 92] Protocol not available" when spawning workers using Windows Subsystem for Linux as it expects "TCP_USER_TIMEOUT" to work -- and it doesn't appear to in WSL. I have created an issue over here: https://github.com/Microsoft/WSL/issues/3873 as I think this is an OS type issue, but I just wanted to bring it up over here to discuss workarounds/proper documentation. Currently it appears (though I'm having trouble confirming conclusively) that this allows worker processes to be orphaned following exceptions, and while some get cleaned up with the message "distributed.comm.tcp - WARNING - Closing dangling stream in <TCP local=tcp://127.0.0.1:46476 remote=tcp://127.0.0.1:46457>
" many others remain running after the program has terminated.

What can we do to mitigate this or how do we better document this limitation?

Host OS: Windows 10.0.17134.619
Container OS: Ubuntu 18.04.2 LTS
```
jwilson@M:~$ python3
Python 3.6.7 (default, Oct 22 2018, 11:32:17)
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.version()
'#523-Microsoft Mon Dec 31 17:49:00 PST 2018'
>>> import sys
>>> sys.platform
'linux'
```

Thanks

jwilson8767:
Could this be related to reusing sockets as py-amqp was over in this issue: [celery/py-amqp#145](https://github.com/celery/py-amqp/issues/145)?

mrocklin:
Thanks for raising the issue @jwilson8767 .  

I personally have no particular experience here and am unlikely to be able to help.  I just wanted to make sure that your well-worded issue here was acknowledged and appreciated.

Perhaps @beberg has suggestions?  (no worries if not)

jwilson8767:
Just to follow up, this issue hasn't progressed on the WSL side. I'm hoping the recently announced WSL2 will offer some changes that will make multiprocessing as a whole a lot more efficient, but I'm unsure what changes to the network stack may help this particular issue.

jurreht:
[...]

PR #2802, opened 2019-06-25
Check whether running on Windows Subsystem for Linux. (#2543)

jurreht:
This works around an issue in WSL that causes Distributed with
default config to crash on initializion. Fix by checking whether
the stack is WSL. If so, proceed as if Windows is running, which
fixes the issue.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -13,6 +13,7 @@ import multiprocessing
 from numbers import Number
 import operator
 import os
+import platform
 import re
 import shutil
 import socket
@@ -62,7 +63,12 @@ no_default = "__no_default__"
 
 
 def _initialize_mp_context():
-    if PY3 and not sys.platform.startswith("win") and "PyPy" not in sys.version:
+    if (
+        PY3
+        and not sys.platform.startswith("win")
+        and "PyPy" not in sys.version
+        and "microsoft" not in platform.uname()[3].lower()
+    ):
         method = dask.config.get("distributed.worker.multiprocessing-method")
         ctx = multiprocessing.get_context(method)
         # Makes the test suite much faster


mrocklin:
Thanks for this fix @jurreht !  It's super-valuable to have Windows devs tracking work here.

To have slightly broader impact, I recommend that we add this to the definition of the `WINDOWS` constant in `compatibility.py

https://github.com/dask/distributed/blob/991391cb71492c2ecf10366bddd0dc8d526f212e/distributed/compatibility.py#L95-L96

And then let's be more consistent about using that constant throughout.  My guess is that this will stop us from running into other issues into the future as well.

I take it that WSL also doesn't support fork?

jurreht:
WSL does support fork, at least in theory. I have verified that setting distributed.worker.multiprocessing-method to fork (instead of the default forkserver) does work. More broadly, using a WINDOWS constant would not work to fix WSL issues since there are many places where it's used in the codebase to do Windows-specific stuff where on WSL the Unix variant is appropriate.

However, it seems this fix is not as useful as I thought at first... While it does resolve the error message in #2543 (and the related error messages in the test suite), performance is so slow as to be unusable when running on my own codebase. My guess is this is an issue in WSL itself with passing messages between processes, but I'm not sure tbh. Based on this, you might consider not merging this pull request since it would give the impression of things working while in fact they don't "really" work.

mrocklin:
> While it does resolve the error message in #2543 (and the related error messages in the test suite), performance is so slow as to be unusable when running on my own codebase. 

OK.  If you have any interest in pursuing this, raising upstream or whatnot please let me know.  Getting Dask to run smoothly on WSL would definitely be nice from our perspective.

mrocklin:
Closing this for now.

Issue #2804, opened 2019-06-26
distributed 2.0.0 is installed in a python 2 environment


cpaulik:
Doing a simple `pip install dask distributed` results in the following:

```
Successfully built distributed
ERROR: distributed 2.0.0 has requirement dask>=2, but you'll have dask 1.2.2 which is incompatible.
Installing collected packages: dask, cloudpickle, msgpack, sortedcontainers, tblib, toolz, heapdict, zict, distributed
Successfully installed cloudpickle-1.2.1 dask-1.2.2 distributed-2.0.0 heapdict-1.0.0 msgpack-0.6.1 sortedcontainers-2.1.0 tblib-1.4.0 toolz-0.9.0 zict-1.0.0
```
I think the metadata on `distributed 2.0.0` is not set correctly since it is still installed although it seems incompatible

TomAugspurger:
Ahh whoops. So what do we do now? First we'll need to fix master by setting a `python_requires` in setup.py. But how do we fix what's released?

IIRC, Django did the same thing. I'll see what I can find.

TomAugspurger:
https://stackoverflow.com/a/47692662

https://code.djangoproject.com/ticket/28878

So right now, as a user, you can do 

```
# requirements.txt
distributed>=1.28,<2.0; python_version<"3.5"
distributed>=2.0; python_version>="3.5"
```

TomAugspurger:
https://github.com/dask/distributed/pull/2806 fixes things going forward.

I'm not sure there's anything we can do about the bad package out on PyPI. I think you'll have to pin `distributed<2` in your requirements.txt The PyPI admins updated the metadata for Django, but it's the internet and people complained.

At least after https://github.com/dask/distributed/pull/2806 you'll get a failure at install time, rather than when attempting to import distributed.

cpaulik:
Is deleting the 2.0.0 release an option after 2.0.1 was uploaded? Or maybe a [post release](https://www.python.org/dev/peps/pep-0440/#post-releases)? No idea how pip would interpret a post release.

TomAugspurger:
Deleting is generally frowned upon.

Would a post release fix the install for python 2 users? Or would pip see that distributed 2.0.0 (not post) allegedly supports Py2, and install that?

mrocklin:
> Would a post release fix the install for python 2 users? Or would pip see that distributed 2.0.0 (not post) allegedly supports Py2, and install that?

My guess is that pip would download the 2.0.0 release.

[...]

PR #2806, opened 2019-06-26
PKG: python_requires and raise for Py2 in setup.py

TomAugspurger:
xref https://github.com/dask/distributed/issues/2804

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,45 @@
 #!/usr/bin/env python
 
 import os
+import sys
 from setuptools import setup
 import versioneer
+
+
+CURRENT_PYTHON = sys.version_info[:2]
+REQUIRED_PYTHON = (3, 5)
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,9 +1,46 @@
...
+This will install the latest version of distributed that works on your
+version of Python. If you can't upgrade your pip (or Python), request
+an older version of distributed:
+
+    $ python -m pip install "distributed<2"
+""".format(
+            *(REQUIRED_PYTHON + CURRENT_PYTHON)
+        )
+    )
+    sys.exit(1)
```
mrocklin: Do we actually need to do this though?  I would have thought that the `python_requires=` keyword would suffice.

mrocklin: The behavior with a modern version of pip on Python 2 for `dask`seems sensible to me without this.

PR #2808, opened 2019-06-26
Calls sync _correct_state() in SpecCluster.scale() if asynchronous=False

pentschev:
@mrocklin this ensures `LocalCluster` reports on workers correctly immediately after creation. I can't find a way to reproduce this issue purely here, but this is the reason why dask-cuda wasn't reporting workers correctly. To solve this quickly, I added a fix on that side https://github.com/rapidsai/dask-cuda/pull/78.

Also, I'm not sure the fix here is the most appropriate one, so feel free to suggest changes.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -275,14 +275,20 @@ class SpecCluster(Cluster):
             self.worker_spec.popitem()
 
         if self.status in ("closing", "closed"):
-            self.loop.add_callback(self._correct_state)
+            if not self.asynchronous:
+                self.sync(self._correct_state)
+            else:
+                self.loop.add_callback(self._correct_state)
             return
 
         while len(self.worker_spec) < n:
             k, spec = self.new_worker_spec()
             self.worker_spec[k] = spec
 
-        self.loop.add_callback(self._correct_state)
+        if not self.asynchronous:
+            self.sync(self._correct_state)
+        else:
+            self.loop.add_callback(self._correct_state)
 
     def new_worker_spec(self):
         """ Return name and spec for the next worker


mrocklin:
The current contract is that the scale method is supposed to return immediately, and not wait for things to finish.

pentschev:
Makes sense to me when it's asynchronous, but otherwise I would expect a synchronous behavior, no? This is what this PR is doing.

mrocklin:
> but otherwise I would expect a synchronous behavior, no? 

That definitely isn't current behavior, nor has it been.  Currently people use scale to say "I want this many machines" and then if they want to they can call `client.wait_for_workers(n=...)`, but scale returns immediately.  This is particularly important for things like adaptive scaling, where we call scale very frequently.  If you'd like to change this policy I recommend raising an issue for broader discussion (there are a number of different FooCluster projects that this would affect) but my sense is that folks won't be sold on the behavior change.

pentschev:
I don't intend to change the behavior, I merely thought this was a mistake when `asynchronous=False`. What I want is to understand what's the proper meaning of `asynchronous`, the documentation reads:

> Set to True if using this cluster within async/await functions or within Tornado gen.coroutines. This should remain False for normal use.

From my C/C++ background this sounds a bit strange. What I would expect is that `asynchronous=True` means the user is responsible for synchronization, and `asynchronous=False` means everything will run sequentially.

I guess my original understanding was incorrect. In both cases everything is asynchronous, what changes is the mechanism for synchronization, so `asynchronous=False` means the user will use async/await or Tornado to synchronize. Is this understanding right?

mrocklin:
You should use `asynchronous=True` if you plan to use async/await syntax.  Otherwise you should use `asynchronous=False`.  There are some small docs on this here: https://distributed.dask.org/en/latest/asynchronous.html

Many operations in Dask don't block regardless.  Operations like submit or persist just fire off work, and make no promises about waiting for completion.  Same with `scale`, which changes a set-point in the cluster, but doesn't guarantee waiting for that set-point to be achieved as part of its contract.  There are other functions to do this waiting.

pentschev:
Thanks for clarifying @mrocklin, closing this.

PR #2812, opened 2019-06-28
Add HTTPS support for the dashboard

### LINKED ISSUE #2775: Serve Dashboard over HTTPS
Do we have a nice way to serve the dashboard over HTTPS?

https://stackoverflow.com/questions/56602973/dask-bokeh-server-accessd-at-http-scheduler-address8787-can-this-be-run-over

jcrist:
Adds optional HTTPS support for the scheduler dashboard. This is only
available via the configuration file, by setting the following fields:

- `distributed.scheduler.tls.cert`: the certificate file
- `distributed.scheduler.tls.key`: the key file, optional if the key
  file is concatenated with the cert above
- `distributed.scheduler.tls.ca-file`: the CA file, optional

These certs *may* be the same as those used for the
scheduler/worker/client communication, but aren't required to be. The
user is responsible for making this decision and providing the proper
configuration.

Likewise, the user is responsible for providing trusted certificates, or
understanding the security implications of telling their browser "I
understand the risks, trust this certificate" (this is more likely,
given the transient nature of dask clusters).

The generated dashboard links now format on an optional `scheme`
parameter, which is either `http` or `https`, depending on if the TLS
configuration fields above are configured.

Fixes #2775

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -27,6 +27,11 @@ distributed:
         task-stream-length: 1000
       tasks:
         task-stream-length: 100000
+      tls:
+        ca-file: null
+        key: null
+        cert: null
+
 
   worker:
     blocked-handlers: []
@@ -88,7 +93,7 @@ distributed:
   ###################
 
   dashboard:
-    link: "http://{host}:{port}/status"
+    link: "{scheme}://{host}:{port}/status"
     export-tool: False
 
   ##################
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -91,6 +91,7 @@ from .utils import (
     shutting_down,
     Any,
     has_keyword,
+    format_dashboard_link,
 )
 from .versions import get_versions
 
@@ -818,8 +819,7 @@ class Client(Node):
                 host = "localhost"
             else:
                 host = rest.split(":")[0]
-            template = dask.config.get("distributed.dashboard.link")
-            address = template.format(host=host, port=port, **os.environ)
+            address = format_dashboard_link(host, port)
             text += (
                 "  <li><b>Dashboard: </b><a href='%(web)s' target='_blank'>%(web)s</a>\n"
                 % {"web": address}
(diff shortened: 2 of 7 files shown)


jcrist:
This works, but due to the transient nature of Dask clusters (with potentially non-consistent addresses), it's not easy to provide certificates that your browser will trust. In the long term (still too alpha for me to recommend), a different option would be to use [dask-gateway](https://jcrist.github.io/dask-gateway/), which provides a consistent public address for all dask dashboards, and supports HTTPS configuration. This allows a trusted certificate (e.g. from letsencrypt) to be used, since the public address is known and consistent for all clusters.

jcrist:
Failure is unrelated (is this a known flaky test?). cc @mrocklin.

distributed/distributed.yaml
```diff
@@ -88,7 +93,7 @@ distributed:
   ###################
 
   dashboard:
-    link: "http://{host}:{port}/status"
+    link: "{scheme}://{host}:{port}/status"
```
mrocklin: scheme -> protocol?

(scheme may also be a standard term here, I just haven't heard of it before)

mrocklin:
I've seen similar failures in the past.  We can probably ignore it.

mrocklin:
The failure is maybe addressed in https://github.com/dask/distributed/pull/2813 ?  I'd give it about a 40% chance of resolving the issue :/

distributed/distributed.yaml
jcrist: Scheme is what `urllib.parse.urlparse` uses.

jcrist: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax

mrocklin: Alrighty then.  Thanks for the reference.

mrocklin:
Thanks @jcrist !  Merging this in.

jeffcbecker:
I just installed the latest dask and friends. To configure https for the dashboard, do I copy distributed.yaml to my .config/dask directory, and edit it to specify certs, etc?

jcrist:
Yes. For more information on our configuration system see https://docs.dask.org/en/latest/configuration.html.

mrocklin:
cc @rcthomas @shreddd , you were asking about this.

PR #2819, opened 2019-07-02
fix case where key, rather than TaskState, could end up in ts.waiting_on

tjb900:
Ideally there would be a test for this, I know :|  Hopefully this is an uncontroversial change even without adding that, though...

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3559,7 +3559,7 @@ class Scheduler(ServerNode):
             for dts in ts.dependencies:
                 dep = dts.key
                 if not dts.who_has:
-                    ts.waiting_on.add(dep)
+                    ts.waiting_on.add(dts)
                 if dts.state == "released":
                     recommendations[dep] = "waiting"
                 else:


mrocklin:
Thanks @tjb900 .

I'm curious, how did you run across this?

tjb900:
Thanks @mrocklin 

I'm not quite sure exactly how it happened, yet.  During a run, I reached a state where nothing was happening, and by using some magic `run_on_scheduler` probes managed to find out that a task was `waiting_on` a string key - so then I went looking for something like the above.

Note that this is using a somewhat patched scheduler - I've been experimenting with adding logic to avoid sending work to workers experiencing memory pressure, and using the `no-worker` state as a fallback when no worker meets the criteria.  So it might have been impossible to hit the above in normal operation; I'm not sure.

TomAugspurger:
Thanks @tjb900 

>  I've been experimenting with adding logic to avoid sending work to workers experiencing memory pressure

FYI, you might want to follow https://github.com/dask/distributed/pull/2765. I'll be picking that up again soon, but it's a bit tricky.

Issue #2815, opened 2019-07-01
LocalCluster does not respect security configuration


rbubley:
I encountered this error in distributed 1.28.1, but from inspecting the code, I expect this to still be an issue in master.

LocalCluster accepts a parameter 'security', but defaults it to None.

By contrast, Scheduler (for example), will read the security parameter from the configs if it is not explicitly set, using 
`self.security = security or Security()`

Depending on your configuration, this can lead to the LocalCluster failing:

```
>>> distributed.deploy.LocalCluster()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/distributed/deploy/local.py", line 219, in __init__
    self.start(ip=ip, n_workers=n_workers)
  File "/usr/local/lib/python2.7/site-packages/distributed/deploy/local.py", line 257, in start
    self.sync(self._start, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/distributed/deploy/local.py", line 250, in sync
    return sync(self.loop, func, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/distributed/utils.py", line 331, in sync
    six.reraise(*error[0])
  File "/usr/local/lib/python2.7/site-packages/distributed/utils.py", line 316, in f
    result[0] = yield future
  File "/usr/local/lib/python2.7/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/usr/local/lib/python2.7/site-packages/tornado/concurrent.py", line 261, in result
    raise_exc_info(self._exc_info)
  File "/usr/local/lib/python2.7/site-packages/tornado/gen.py", line 326, in wrapper
    yielded = next(result)
  File "/usr/local/lib/python2.7/site-packages/distributed/deploy/local.py", line 283, in _start
    self.scheduler.start(address)
  File "/usr/local/lib/python2.7/site-packages/distributed/scheduler.py", line 1199, in start
    self.listen(addr_or_port, listen_args=self.listen_args)
  File "/usr/local/lib/python2.7/site-packages/distributed/core.py", line 320, in listen
    connection_args=listen_args,
  File "/usr/local/lib/python2.7/site-packages/distributed/comm/core.py", line 258, in listen
    loc, handle_comm, deserialize, **(connection_args or {})
  File "/usr/local/lib/python2.7/site-packages/distributed/comm/tcp.py", line 530, in get_listener
[...]

PR #2822, opened 2019-07-02
Respect security configuration (#2815)

rbubley:


mrocklin:
cc @jcrist

mrocklin:
Thanks for this fix @rbubley !  If it's ok with you I'd prefer to wait for @jcrist to take a look at this, but I suspect that he's off until next week due to the American holiday.

distributed/deploy/local.py
```diff
@@ -123,11 +124,15 @@ def __init__(
...
         self.processes = processes
+        security = security or Security()
 
         if protocol is None:
             if host and "://" in host:
                 protocol = host.split("://")[0]
-            elif security:
+            elif (
+                security
+                and security.get_connection_args("scheduler")["require_encryption"]
```
jcrist: This should be:

```python
elif security and security.require_encryption:
```

mrocklin:
This is in!  Thanks @rbubley for the fix and @jcrist for the review.

PR #2824, opened 2019-07-03
Remove dask-mpi

### LINKED ISSUE #2823: Deprecated bin/dask_mpi overwrites that from dask-mpi project
It appears there are two implementations of dask_mpi : The deprecated one in [distributed/cli/dask_mpi.py](https://github.com/dask/distributed/blob/master/distributed/cli/dask_mpi.py), and the newer [dask-mpi](https://github.com/dask/dask-mpi) project.

Somehow in Anaconda I managed to do the follow: (1) Install dask/distributed, (2) install dask-mpi, (3) reinstall dask/distributed. The problem is that step (3) overwrote the dask-mpi binary `bin/dask_mpi` with the deprecated one from dask/distributed.

This problem was resolved by removing and reinstalling the dask-mpi package. However, it took a while to figure out what's going on. Is there a way to avoid this problem of conflicting packages?

mrocklin:
This has moved to the dask-mpi project documented at
https://mpi.dask.org

Fixes #2823


mrocklin:
cc @andersy005 @kmpaul

Issue #2771, opened 2019-06-13
Small doc clarifications needed


chrish42:
I'm reading through the docs of Dask Distributed, and a couple points are unclear to me. I'd be happy to submit a pull request with doc improvements if someone can at least point me in the general direction of answers for my doc questions. (Actually, two of my three questions are yes/no, so at least those two should be quick to answer for someone who knows Dask Distributed well.)

1. In the "Resilience" page, the "Restart and Nanny Process" section doesn't really talk about nanny process. When would I want to use one and why? And when is it not needed?
2. Again in the "Resilience" page, for "Closed Network Connections", point 3: "Data scattered out to the workers is not kept in the scheduler (it is often quite large) and so the loss of this data is irreparable." Is that just about data send via the "scatter()" method? Or is scattering used in a broader sense here?
3. When talking about ClientState in the "Scheduling State", etc. sections, that's an internal object that describes the state of each worker, and which is not related to the Client() object that is instantiated by the user, correct?

Thanks!

TomAugspurger:
1. Nanny seems relatively undocumented. I suppose we could add an API autodoc for it in https://distributed.dask.org/en/latest/worker.html
2. I believe that's correct. May want to double check on `persist`ed collections though...
3. Correct. It's referring to https://distributed.dask.org/en/latest/scheduling-state.html#distributed.scheduler.ClientState (you could link to it with :class:`scheduler.ClientState` if you want to disambiguate.

martindurant:
@chrish42 , are you interested in providing the clarifications as a PR?

chrish42:
Yes. On vacation now, though, so I'll create the pull request beginning of July, if that's okay.

martindurant:
That's fine, just wanting to know how things stand - thanks.

chrish42:
Submitted a first cut at a pull request.

1. I added autodoc for Nanny, but there should probably also be at least a sentence or two there also. Basically, what am I missing out on when I'm not using the nanny to supervise my workers?
2. My understanding is that  data that is `persist`ed will be recreated, because the scheduler knows the Dask task graph that created it, and so can rerun that. But my understanding of the code is still pretty shallow. I think the distinction is between loading data directly (via scatter()) vs. creating data from a Dask task graph (where you can rerun the graph to recreate the data if needed), so I tweaked the doc accordingly. Please confirm, though.

PR #2826, opened 2019-07-05
Add Nanny to worker docs

chrish42:
From my experience reading through it. For bug #2771.

### DIFF
--- a/docs/source/resilience.rst
+++ b/docs/source/resilience.rst
@@ -49,7 +49,7 @@ This has some fail cases.
     called on other workers.  This function will be marked as "bad" after it
     kills a fixed number of workers (defaults to three).
 3.  Data sent out directly to the workers via a call to ``scatter()`` (instead
-    of being created via a call to ``persist()`` from a Dask task graph) is not
+    of being created from a Dask task graph via other Dask functions) is not
     kept in the scheduler, as it is often quite large, and so the loss of this
     data is irreparable.  You may wish to call ``Client.replicate`` on the data
     with a suitable replication factor to ensure that it remains long-lived or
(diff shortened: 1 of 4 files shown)


docs/source/resilience.rst
```diff
@@ -48,11 +48,12 @@ This has some fail cases.
     causes a segmentation fault, then that bad function will repeatedly be
     called on other workers.  This function will be marked as "bad" after it
     kills a fixed number of workers (defaults to three).
-3.  Data scattered out to the workers is not kept in the scheduler (it is
-    often quite large) and so the loss of this data is irreparable.  You may
-    wish to call ``Client.replicate`` on the data with a suitable replication
-    factor to ensure that it remains long-lived or else back the data off of
-    some resilient store, like a file system.
+3.  Data sent out directly to the workers via a call to ``scatter()`` (instead
+    of being created via a call to ``persist()`` from a Dask task graph) is not
```
mrocklin: This parenthetical is slightly troublesome because the `persist` function is only one of a few methods that will trigger computation (client.compute, submit, map, ...)

docs/source/worker.rst
```diff
@@ -233,6 +233,12 @@ YARN, Mesos, SGE, etc..).  After termination the nanny will restart the worker
 in a fresh state.
 
 
+Nanny
+~~~~~
+
```
mrocklin:
```suggestion
Dask workers are often launched, monitored, and managed by a small Nanny process.
```

Then, I would add more context into the Nanny docstring itself.  In particular, we should probably add a See Also section that links to the worker, and we should mention that the parameters to Nanny are mostly the same as the parameters that are in the Worker.

mrocklin:
Thanks for doing this @chrish42 .  I've added a couple of small comments with some thoughts.

chrish42:
Updated the pull request to handle all the comments I received.

chrish42:
test_gather_after_failed_worker is failing, but I don't understand how that could be caused by my changes. Flaky test?

mrocklin:
Thanks @chrish42 !

PR #2827, opened 2019-07-06
Add alternative SSHCluster implementation

mrocklin:
This is a proof of concept here for two reasons:

1.  It opens up a possible alternative for SSH deployment (which was surprisingly popular in the user survey)
2.  It is the first non-local application of `SpecCluster` and so serves as a proof of concept for other future deployments that are mostly defined by creating a remote Worker/Scheduler object

This forced some changes in `SpecCluster`, notably we now have an `rpc` object that does remote calls rather than accessing the scheduler directly.  Also, we're going to have to figure out how to handle all of the keyword arguments.  In this case we need to pass them from Python down to the CLI, and presumably we'll also want a `dask-ssh` CLI command which has to translate the other way.

cc @jcrist @jacobtomlinson @quasiben

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -225,8 +225,6 @@ class SpecCluster(Cluster):
         return _().__await__()
 
     async def _wait_for_workers(self):
-        # TODO: this function needs to query scheduler and worker state
-        # remotely without assuming that they are local
         while {
             str(d["name"])
             for d in (await self.scheduler_comm.identity())["workers"].values()
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -5,8 +5,10 @@ import weakref
 from tornado import gen
 
 from .cluster import Cluster
+from ..core import rpc
 from ..utils import LoopRunner, silence_logging, ignoring
 from ..scheduler import Scheduler
+from ..security import Security
 
 
 class SpecCluster(Cluster):
@@ -107,6 +109,7 @@ class SpecCluster(Cluster):
         worker=None,
         asynchronous=False,
         loop=None,
+        security=None,
         silence_logs=False,
     ):
         self._created = weakref.WeakSet()
@@ -125,6 +128,8 @@ class SpecCluster(Cluster):
         self.workers = {}
         self._i = 0
         self._asynchronous = asynchronous
+        self.security = security or Security()
+        self.scheduler_comm = None
 
         if silence_logs:
             self._old_logging_level = silence_logging(level=silence_logs)
@@ -156,6 +161,10 @@ class SpecCluster(Cluster):
         self._lock = asyncio.Lock()
         self.status = "starting"
         self.scheduler = await self.scheduler
+        self.scheduler_comm = rpc(
+            self.scheduler.address,
+            connection_args=self.security.get_connection_args("client"),
+        )
         self.status = "running"
 
     def _correct_state(self):
@@ -174,11 +183,13 @@ class SpecCluster(Cluster):
             pre = list(set(self.workers))
             to_close = set(self.workers) - set(self.worker_spec)
             if to_close:
-                await self.scheduler.retire_workers(workers=list(to_close))
+                if self.scheduler.status == "running":
+                    await self.scheduler_comm.retire_workers(workers=list(to_close))
                 tasks = [self.workers[w].close() for w in to_close]
                 await asyncio.wait(tasks)
                 for task in tasks:  # for tornado gen.coroutine support
-                    await task
+                    with ignoring(RuntimeError):
+                        await task
             for name in to_close:
                 del self.workers[name]
 
@@ -216,9 +227,10 @@ class SpecCluster(Cluster):
     async def _wait_for_workers(self):
         # TODO: this function needs to query scheduler and worker state
         # remotely without assuming that they are local
-        while {d["name"] for d in self.scheduler.identity()["workers"].values()} != set(
-            self.workers
-        ):
+        while {
+            str(d["name"])
+            for d in (await self.scheduler_comm.identity())["workers"].values()
+        } != set(map(str, self.workers)):
             if (
                 any(w.status == "closed" for w in self.workers.values())
                 and self.scheduler.status == "running"
@@ -240,12 +252,14 @@ class SpecCluster(Cluster):
             return
         self.status = "closing"
 
-        async with self._lock:
-            await self.scheduler.close(close_workers=True)
         self.scale(0)
         await self._correct_state()
+        async with self._lock:
+            await self.scheduler_comm.close(close_workers=True)
+        await self.scheduler.close()
         for w in self._created:
             assert w.status == "closed"
+        self.scheduler_comm.close_rpc()
 
         if hasattr(self, "_old_logging_level"):
             silence_logging(self._old_logging_level)
--- a/distributed/deploy/ssh2.py
+++ b/distributed/deploy/ssh2.py
@@ -44,6 +44,19 @@ class Process:
     def __repr__(self):
         return "<SSH %s: status=%s>" % (type(self).__name__, self.status)
 
+async def _create_connection(address, validate_host):
+    if not validate_host:
+        connection = await asyncssh.connect(address, known_hosts=None)
+    else:
+        try:
+            connection = await asyncssh.connect(address)
+        except asyncssh.misc.HostKeyNotVerifiable:
+            # improve debug message when there is more support for host validation
+            logger.debug("Explicitly validating host")
+            raise asyncssh.misc.HostKeyNotVerifiable
+
+    return connection
+
 
 class Worker(Process):
     """ A Remote Dask Worker controled by SSH
@@ -55,18 +68,21 @@ class Worker(Process):
     address: str
         The hostname where we should run this worker
     kwargs:
-        TODO
+        validate_host: bool
+            Validate host key is trusted (default is False)
     """
 
     def __init__(self, scheduler: str, address: str, **kwargs):
         self.address = address
         self.scheduler = scheduler
         self.kwargs = kwargs
+        self.validate_host = self.kwargs.get('validate_host', False)
+
 
         super().__init__()
 
     async def start(self):
-        self.connection = await asyncssh.connect(self.address)
+        self.connection = await _create_connection(self.address, self.validate_host)
         self.proc = await self.connection.create_process(
             " ".join(
                 [
@@ -99,17 +115,22 @@ class Scheduler(Process):
         The hostname where we should run this worker
     kwargs:
         TODO
+        validate_host: bool
+            Validate host key is trusted (default is False)
     """
 
     def __init__(self, address: str, **kwargs):
         self.address = address
         self.kwargs = kwargs
+        self.validate_host = self.kwargs.get('validate_host', False)
 
         super().__init__()
 
     async def start(self):
         logger.debug("Created Scheduler Connection")
-        self.connection = await asyncssh.connect(self.address)
+
+        self.connection = await _create_connection(self.address, self.validate_host)
+
         self.proc = await self.connection.create_process(
             " ".join([sys.executable, "-m", "distributed.cli.dask_scheduler"])
         )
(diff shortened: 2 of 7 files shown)


mrocklin:
Also cc @guillaumeeb @jhamman .  I could imagine rewriting Dask Jobqueue in this way as well.

guillaumeeb:
> notably we now have an rpc object that does remote calls rather than accessing the scheduler directly

That part sounds really interesting! I'm kinda overwhelmed by high level strategical questions currently at work, and I still have to go through SpecCluster... But I've not given up, and I hope I can have a closer look at all this during the summer!

mrocklin:
Mostly I encourage you to look not at SpecCluster, which is complicated,
but at the Worker class here, which I'm hoping is all we would have to
write for a project like Dask Jobqueue.

jacobtomlinson:
This is really nice to see. I think we could remove a bunch of code from `dask-kubernetes` by switching to this. 

How would adaptive behave in this situation? I notice there is no stop method implemented on the worker classes and I guess there are a fixed number of hosts anyway. It could adapt within the number of hosts?

mrocklin:
> How would adaptive behave in this situation? I notice there is no stop method implemented on the worker classes and I guess there are a fixed number of hosts anyway. 

Adaptive first asks the scheduler to kill the worker gracefully.  Eventually it then calls the `Worker.close` method on the `Worker` object (where `Worker` is now `SSHWorker` or `PodWorker` or something similar).  

> It could adapt within the number of hosts?

I'm not sure I fully understand this question.

jacobtomlinson:
> It could adapt within the number of hosts?

If I provide a list of nine host addresses I will get one scheduler and eight workers. I would then expect that it would adapt up a maximum of eight workers. Is that also what you would expect?

mrocklin:
Oh I see, right.  There isn't an obvious way to add new workers.

I would probably just not support adaptive for SSHClusters to start.

mrocklin:
@jcrist next time I run into you I'll probably have questions about your thoughts on how we should unify keywords in Python and on the CLI.  That ends up being fairly important here.

mrocklin:
I think I'm inclined to merge this in without including it in any public import, mostly as a short-term test case and with a warning on use.

Tests are failing currently because ssh'ing into localhost isn't working trivially.  My guess is that the fix to this is fairly simple, but I'm not immediately sure how to handle it.  @quasiben I suspect that you can probably solve this quickly.  Could you take a look?

```python-traceback
    def validate_server_host_key(self, key_data):
        """Validate and return the server's host key"""
    
        try:
            host_key = self._validate_host_key(self._host, self._peer_addr,
                                               self._port, key_data)
        except ValueError as exc:
>           raise HostKeyNotVerifiable(str(exc)) from None
E           asyncssh.misc.HostKeyNotVerifiable: Host key is not trusted

```

quasiben:
Sure, I'll take a look now

quasiben:
My first suspicion is that travis-ci sometimes has inconsistent behavior with regards to `localhost` vs `127.0.0.1`.  Mind if I take over this PR and get the CI sorted out?

mrocklin:
> Mind if I take over this PR and get the CI sorted out?

You are always welcome to push to any PR of mine.

quasiben:
Ah, nevermind -- the error is `hostkey trust issue` so we probably just need to pass in another flag or something.

>You are always welcome to push to any PR of mine.

Great!

quasiben:
@mrocklin pushed a commit to resolve the issue.  This is a bit of security issue (disabling host key validation) -- I added some options/comments for when things are improved

distributed/deploy/ssh2.py
```diff
@@ -44,6 +44,19 @@ def __await__(self):
...
+        connection = await asyncssh.connect(address, known_hosts=None)
+    else:
+        try:
+            connection = await asyncssh.connect(address)
+        except asyncssh.misc.HostKeyNotVerifiable:
+            # improve debug message when there is more support for host validation
+            logger.debug("Explicitly validating host")
+            raise asyncssh.misc.HostKeyNotVerifiable
+
+    return connection
```
mrocklin: It seems like we might be able to drop this function (which would be nice to keep things simple) and optionally add the `known_hosts=None` keyword where we call `asyncssh.connect`.  I think that things would be the same but we just wouldn't get the logging, is that correct?

distributed/deploy/ssh2.py
```diff
@@ -55,18 +68,21 @@ class Worker(Process):
     address: str
         The hostname where we should run this worker
     kwargs:
-        TODO
+        validate_host: bool
+            Validate host key is trusted (default is False)
```
mrocklin: Let's unpack this outside of kwargs and put it just above

jacobtomlinson:
@quasiben defaulting this to `False` makes me nervous. I would be tempted to default to `True` and just set to false in the tests.

Also happy not to as it's a POC. It just makes the sysadmin in my twitchy.

distributed/deploy/ssh2.py
```diff
@@ -55,18 +68,21 @@ class Worker(Process):
...
     """
 
     def __init__(self, scheduler: str, address: str, **kwargs):
         self.address = address
         self.scheduler = scheduler
         self.kwargs = kwargs
+        self.validate_host = self.kwargs.get('validate_host', False)
```
mrocklin: Let's pull this out from kwargs above

```python
    def __init__(self, scheduler: str, address: str, validate_host=False : bool, **kwargs):
```

mrocklin: Or should we call this `known_hosts` and then just pass it through?

Or, adding even fewer opinions, have a `connect_kwargs=` keyword that we pass down to `asyncssh.connect`?

distributed/deploy/ssh2.py
quasiben: correct

mrocklin:
> @quasiben defaulting this to False makes me nervous. I would be tempted to default to True and just set to false in the tests.

> Also happy not to as it's a POC. It just makes the sysadmin in my twitchy.

I think that we can resolve this by just setting something to pass through kwargs to connect.  That way we just expose functionality, but don't take any opinions.

quasiben:
@mrocklin tests are passing now

distributed/deploy/spec.py
```diff
@@ -240,12 +252,14 @@ def __await__(self):
             return
         self.status = "closing"
 
-        async with self._lock:
-            await self.scheduler.close(close_workers=True)
         self.scale(0)
         await self._correct_state()
+        async with self._lock:
+            await self.scheduler_comm.close(close_workers=True)
```
jacobtomlinson: If the scheduler comm has closed for some reason and resulted in a broken pipe we will get an exception here.

```python-traceback
Error in atexit._run_exitfuncs:
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/comm/tcp.py", line 194, in read
    n_frames = yield stream.read_bytes(8)
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/iostream.py", line 436, in read_bytes
    future = self._start_read()
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/iostream.py", line 797, in _start_read
    self._check_closed()  # Before reading, check that stream is not closed.
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/iostream.py", line 1009, in _check_closed
    raise StreamClosedError(real_error=self.error)
tornado.iostream.StreamClosedError: Stream is closed

  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/core.py", line 675, in send_recv_from_rpc
    result = yield send_recv(comm=comm, op=key, **kwargs)
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/core.py", line 535, in send_recv
    response = yield comm.read(deserializers=deserializers)
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/gen.py", line 209, in wrapper
    yielded = next(result)
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/comm/tcp.py", line 214, in read
    convert_stream_closed_error(self, e)
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/comm/tcp.py", line 139, in convert_stream_closed_error
    raise CommClosedError("in %s: %s: %s" % (obj, exc.__class__.__name__, exc))
distributed.comm.core.CommClosedError: in <closed TCP>: BrokenPipeError: [Errno 32] Broken pipe

  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/utils.py", line 182, in ignoring
    yield
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/deploy/spec.py", line 343, in close_clusters
    cluster.close(timeout=10)
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/deploy/spec.py", line 271, in close
    return self.sync(self._close, callback_timeout=timeout)
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/deploy/cluster.py", line 245, in sync
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/utils.py", line 332, in sync
    six.reraise(*error[0])
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/six.py", line 693, in reraise
    raise value
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/utils.py", line 317, in f
    result[0] = yield future
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/deploy/spec.py", line 258, in _close
    await self.scheduler_comm.close(close_workers=True)
  File "/home/nfs/jtomlinson/miniconda3/envs/dask-csp/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/core.py", line 678, in send_recv_from_rpc
    "%s: while trying to call remote method %r" % (e, key)
distributed.comm.core.CommClosedError: in <closed TCP>: BrokenPipeError: [Errno 32] Broken pipe: while trying to call remote method 'close'
```

This results in `self.scheduler.close()` never being called and failing to clean up.

distributed/deploy/spec.py
```diff
@@ -216,9 +227,10 @@ def __await__(self):
     async def _wait_for_workers(self):
         # TODO: this function needs to query scheduler and worker state
         # remotely without assuming that they are local
```
mrocklin: I should probably remove this TODO.  I think that now that we're accessing the comm rather than the scheduler that we're handling the remote case.

jacobtomlinson: I seem to still be getting caught in this loop forever despite it using the `scheduler_comm`.  Only when starting a cluster with workers and awaiting the cluster object. When scaling up it's fine.

mrocklin: Perhaps there is some inconsistency between `self.workers` (the cluster's expected worker set) and the return value of `identity` (what the scheduler sees)?  It might be worth printing out what these look like to see if, for example, they are using different naming conventions.

jacobtomlinson: Weirdly it seems to be that `await cluster.scheduler_comm.identity()` hangs forever every other time you call it. If I keyboard interrupt it and run it again I get the output twice.

mrocklin: That seems odd.  Do you have a branch where I can try this out?

mrocklin:
This is in.  Thanks @quasiben for pushing this over the finish line.

distributed/deploy/spec.py
jacobtomlinson: Happening for me on master of [dask-cloud](https://github.com/jacobtomlinson/dask-cloud). 

```python
In [1]: from dask_cloud import ECSCluster

In [2]: cluster = ECSCluster()

In [3]: await cluster.scheduler_comm.identity()  # hangs, press ctrl+c

In [4]: await cluster.scheduler_comm.identity()  # Returns
```

I'll look at working up a minimum reproducible example if I keep getting stuck with this.

Issue #2685, opened 2019-05-11
unable to set nanny_port programmatically 


MichaelSchreier:
When creating new workers via the `.start_worker()` method of a cluster instance it is possible to supply the `worker_port` argument to pin down which port is used by that worker.
However, the `nanny_port` argument is not supported by this method and only accessible when launching workers via the CLI.

mrocklin:
You can pass any keyword argument to `start_worker` that the `Nanny` expects.  I think that `port=` may be one of them.

MichaelSchreier:
I see that what you describe was added in #2667 a couple of days. I'll be waiting for the next release then. Thanks for the hard work !

MichaelSchreier:
I just played around with the latest release and noticed that `cluster.start_worker()` has been deprecated. The suggested alternative via `cluster.scale()`, however, appears to lack functionality to assign each worker a different set of properties (`worker_port` and `port` in my case).
I tried various methods of setting `cluster.new_spec` and `cluster.worker_spec` manually and iteratively scaling the cluster but to no avail. 

Am I missing something obvious or has the new way of starting workers made it impossible to set individual worker's properties?

TomAugspurger:
I suspect you can set this in `cluster.new_spec`. Not sure if there's an easy API for defining that when creating the LocalCluster. May have to do it after the fact?


```python
In [2]: cluster = LocalCluster(n_workers=1)

In [3]: cluster.new_spec
Out[3]:
{'cls': distributed.nanny.Nanny,
 'options': {'memory_limit': 17179869184,
  'nthreads': 8,
  'services': {},
  'dashboard_address': None,
  'interface': None,
  'protocol': 'tcp://',
  'security': None,
  'silence_logs': 30}}
```

Can you try that out.

MichaelSchreier:
After writing a long answer detailing what I've tried and what didn't work I seem to have stumbled on the root cause of the issue and a workaround.
This code doesn't work:
```python
In [2]: cluster = LocalCluster(n_workers=0, threads_per_worker=1, process=True)
In [3]:
for i in range(4):
    spec = cluster.new_spec
    spec['options']['worker_port'] = 9000 + i
    cluster.new_spec = spec
    cluster.scale(i+1)
```
Here, every other worker appears to try to use the same port as its successor and ends up failing.

Adding a time delay solves the issue but is not particularly elegant:
```python
In [2]: cluster = LocalCluster(n_workers=0, threads_per_worker=1, process=True)
In [3]:
[...]

PR #2831, opened 2019-07-09
Make await/start more consistent across Scheduler/Worker/Nanny

mrocklin:
Now every ServerNode has a start async method that returns itself.
And the __await__ method is handled in the superclass

cc @jcrist , this is in a theme of things that you've asked for

Fixes https://github.com/dask/distributed/issues/2685  cc @MichaelSchreier


mrocklin:
FYI @jacobtomlinson , this should make setup things marginally cleaner.  There isn't much for you to review here I think, I just thought I'd keep you in the loop.

Issue #2792, opened 2019-06-21
Extension of Prometheus Metrics Endpoint


sublinus:
Hi @mrocklin,

Would you be willing to accept a contribution regarding the extension of the current metrics endpoints containing:
 
- number of tasks in states
- server tick duration

Looking forward to the contribution.

mrocklin:
I personally don't know much about Prometheus, so I'm not sure that I
personally am the right person to answer this.  I don't know what level of
information is appropriate for prometheus endpoints. Perhaps @beberg can
help here ?

Fra-nk:
Prometheus by default is built to handle a huge amount of metrics and since it's intended as a systems monitoring system, exposing verbose information is (imho) feasible.

- You can see some [metrics examples in the Prometheus documentation](https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-example) where an app exposes statistics on HTTP requests - IMHO a rather verbose thing.
- [Prometheus Node exporter](https://prometheus.io/docs/guides/node-exporter/#node-exporter-metrics) exposes detailed internals with a single system usually exposing 600+ metrics.
- Even the Python Prometheus client library exposes [very detailed statistics](https://github.com/prometheus/client_python/blob/master/prometheus_client/gc_collector.py#L20) on its own internal operation.

→ I think the deciding factors if a metric should (not) be exposed should be
- the cost of _collecting_ the data (usually repeatedly)
- high-level questions like "does exposing this value have security implications"

mrocklin:
Under the criteria that @Fra-nk suggests:

>  the cost of collecting the data (usually repeatedly)

The number of tasks in states requires a full pass over the current tasks.  There are possibly 1000 to 100000 tasks on a worker at any given time.  Checking the status of each is probably fine to do every time prometheus checks in, assuming that happens every second or so.  You might want to consider the `cytoolz.frequencies` function, which is fairly fast at counting things like this.

```python
d = frequencies(self.task_state.values())
```

Tick duration seems fine.  You might consider generalizing this to all current counters and digests.  For background, we keep track of various metrics in code like this:

https://github.com/dask/distributed/blob/991391cb71492c2ecf10366bddd0dc8d526f212e/distributed/worker.py#L1880-L1884

This actually gets plotted in Dask's dashboard in the hidden [`/counters` page](
[...]

PR #2833, opened 2019-07-12
Extend prometheus metrics endpoint (#2792)

sublinus:
This PR introduces the following prometheus metrics:
- scheduler
    - number of tasks that were received
    - number of unrunnable tasks
- worker
   - tasks: stored, ready, waiting and serving 
   - number of connections to other workers
   - number of worker threads
   - latency
   - median tick duration
   - median task duration
   - median transfer bandwidth

The requirement of crick for the digest-based metrics is handled by checking if crick is available.
If crick is not available there is a log message on loglevel info regarding the missing crick and the metrics, which require crick are not exposed.

Regarding what @mrocklin mentioned in the issue(#2792) about the digest metrics not being exposed individually. The problem here is that the metrics names would not be compliant with prometheus naming. Additionally expressive descriptions would not be available either if the metrics were not exposed individually.

### DIFF
--- a/distributed/dashboard/worker_html.py
+++ b/distributed/dashboard/worker_html.py
@@ -11,7 +11,9 @@ class _PrometheusCollector(object):
             import crick  # noqa: F401
         except ImportError:
             self.crick_available = False
-            self.logger.info("Digest-based metrics require crick to be installed")
+            self.logger.info(
+                "Not all prometheus metrics available are exported. Digest-based metrics require crick to be installed"
+            )
 
     def collect(self):
         from prometheus_client.core import GaugeMetricFamily
--- a/distributed/dashboard/worker_html.py
+++ b/distributed/dashboard/worker_html.py
@@ -8,10 +8,7 @@ class _PrometheusCollector(object):
         self.logger = logging.getLogger("distributed.dask_worker")
         self.crick_available = True
         try:
-            import crick
-
-            # usage of import necessary to comply with flake8 F401
-            crick_version = crick._version
+            import crick  # noqa: F401
         except ImportError:
             self.crick_available = False
             self.logger.info("Digest-based metrics require crick to be installed")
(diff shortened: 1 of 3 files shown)


mrocklin:
@jacobtomlinson would you mind reviewing this?

distributed/dashboard/worker_html.py
```diff
@@ -1,23 +1,70 @@
...
-        self.server = server
+    def __init__(self, server):
+        self.worker = server
+        self.logger = logging.getLogger("distributed.dask_worker")
+        self.crick_available = True
+        try:
+            import crick  # noqa: F401
+        except ImportError:
+            self.crick_available = False
+            self.logger.info("Digest-based metrics require crick to be installed")
```
jacobtomlinson: Could this message contain a little more context? Given that you are logging under `distributed.dask_worker` it may not be clear that this is prometheus or monitoring related. 

Perhaps including the word prometheus in here somewhere would be useful.

distributed/dashboard/worker_html.py
```diff
@@ -1,23 +1,70 @@
...
-        while False:
-            yield None
-        #
-        # 2. yield your metrics
-        #     yield prometheus_client.core.GaugeMetricFamily(
-        #         'dask_worker_connections',
-        #         'Number of connections currently open.',
-        #         value=???,
-        #     )
+        from prometheus_client.core import GaugeMetricFamily
```
jacobtomlinson: Switching to importing here seems reasonable, but does this mean there is now a stray unused import somewhere?

sublinus: Previously the import was done at the init of PrometheusHandler when the PrometheusCollector is registered in the registry.
All imports which are there are currently still required in the PrometheusHandler, which was the only point the changed _PrometheusCollector class is used.
So there is no dangling import for prometheus_client anywhere in my opinion.

mrocklin:
Merging this in.  Thanks @sublinus !  Thanks also to @jacobtomlinson for the review.

PR #2834, opened 2019-07-12
Make spec repr generic

jacobtomlinson:
Set `SpecCluster` repr to use type for class name. Useful when subclassing.

### DIFF
--- a/distributed/deploy/tests/test_spec_cluster.py
+++ b/distributed/deploy/tests/test_spec_cluster.py
@@ -82,6 +82,15 @@ def test_loop_started():
 
 
 @pytest.mark.asyncio
+async def test_repr():
+    class MyCluster(SpecCluster):
+        pass
+
+    async with MyCluster(..., asynchronous=True) as cluster:
+        assert "MyCluster" in str(cluster)
+
+
+@pytest.mark.asyncio
 async def test_scale():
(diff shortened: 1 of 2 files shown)


mrocklin:
+1 from me.  I'll suggest the following test:

```python
async def test_repr():
    class MyCluster(SpecCluster):
        pass

    async with MyCluster(..., asynchronous=True) as cluster:
        assert "MyCluster" in str(cluster)
```

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -81,6 +81,15 @@ def test_loop_started():
     )
 
 
+@pytest.mark.asyncio
+async def test_repr():
+    class MyCluster(SpecCluster):
+        pass
+
+    async with MyCluster(..., asynchronous=True) as cluster:
```
mrocklin: Ah, blind copy-paste issue I think?  I used `...` to mean "I haven't thought about what should go in here too much, but you could probably copy-paste from similar tests"

jacobtomlinson: Ha my bad! I assumed that was some clever shorthand for defaulting everything that I hadn't seen before.

mrocklin:
Thanks @jacobtomlinson !

PR #2838, opened 2019-07-13
Use click's show_default=True in relevant places

chrish42:
For some reason, if you specify a default value for an option with click, it will not show that default value in the generated `--help` usage doc unless you also specify `show_default=True`.  So, for example, it's not clear from reading the usage doc that `dask-worker`'s default is to use a nanny process, etc. and I had to go read the code to be sure of that.

The following pull request adds `show_default=True` to relevant click options and switches some options from a hardcoded explanation of the default in the `help=` text to also use that. I think this makes the output of `--help` clearer for less experienced users.

My only question actually is for the options like `--nanny / --no-nanny`. Click will now show `[default: True]`. It might be clearer to hardcode this to be something like `[default: --nanny]`. What do folks think?

If this is generally deemed useful and merged, I'll do the same thing for the dask/dask repo.

### DIFF
--- a/distributed/cli/dask_remote.py
+++ b/distributed/cli/dask_remote.py
@@ -7,7 +7,7 @@ from distributed.submit import _remote
 
 @click.command()
 @click.option("--host", type=str, default=None, help="IP or hostname of this server")
-@click.option("--port", type=int, default=8788, help="Remote Client Port")
+@click.option("--port", type=int, default=8788, show_default=True, help="Remote Client Port")
 @click.version_option()
 def main(host, port):
     _remote(host, port)
--- a/distributed/cli/dask_remote.py
+++ b/distributed/cli/dask_remote.py
@@ -7,7 +7,9 @@ from distributed.submit import _remote
 
 @click.command()
 @click.option("--host", type=str, default=None, help="IP or hostname of this server")
-@click.option("--port", type=int, default=8788, show_default=True, help="Remote Client Port")
+@click.option(
+    "--port", type=int, default=8788, show_default=True, help="Remote Client Port"
+)
 @click.version_option()
 def main(host, port):
     _remote(host, port)
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -68,6 +68,7 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "--dashboard-address",
     type=str,
     default=":8787",
+    show_default=True,
     help="Address on which to listen for diagnostics dashboard",
 )
 @click.option(
@@ -85,7 +86,7 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     required=False,
     help="Deprecated.  See --dashboard/--no-dashboard.",
 )
-@click.option("--show/--no-show", default=False, help="Show web UI")
+@click.option("--show/--no-show", default=False, show_default=True, help="Show web UI")
 @click.option(
     "--dashboard-prefix", type=str, default=None, help="Prefix for the dashboard app"
 )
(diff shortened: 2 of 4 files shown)


mrocklin:
cc @jcrist , you might have thoughts on this

mrocklin:
Poke @jcrist

jcrist:
I think this makes sense, but I'd not add it for boolean options, instead relying on the text in the help description (as you said above).

> If this is generally deemed useful and merged, I'll do the same thing for the dask/dask repo.

There are no clis in `dask/dask`.

chrish42:
How about this? Now for example, `dask-worker --help` looks like this:

```
Usage: dask-worker [OPTIONS] [SCHEDULER] [PRELOAD_ARGV]...

Options:
  --tls-ca-file PATH            CA cert(s) file for TLS (in PEM format)
  --tls-cert PATH               certificate file for TLS (in PEM format)
  --tls-key PATH                private key file for TLS (in PEM format)
  --worker-port INTEGER         Serving computation port, defaults to random
  --nanny-port INTEGER          Serving nanny port, defaults to random
  --bokeh-port INTEGER          Deprecated.  See --dashboard-address
  --dashboard-address TEXT      Address on which to listen for diagnostics
                                dashboard
  --dashboard / --no-dashboard  Launch the Dashboard [default: --dashboard]
  --bokeh / --no-bokeh          Deprecated.  See --dashboard/--no-dashboard.
  --listen-address TEXT         The address to which the worker binds.
                                Example: tcp://0.0.0.0:9000
  --contact-address TEXT        The address the worker advertises to the
                                scheduler for communication with it and other
                                workers. Example: tcp://127.0.0.1:9000
  --host TEXT                   Serving host. Should be an ip address that is
                                visible to the scheduler and other workers.
                                See --listen-address and --contact-address if
                                you need different listen and contact
                                addresses. See --interface.
  --interface TEXT              Network interface like 'eth0' or 'ib0'
  --protocol TEXT               Protocol like tcp, tls, or ucx
  --nthreads INTEGER            Number of threads per process.
  --nprocs INTEGER              Number of worker processes to launch.
                                [default: 1]
  --name TEXT                   A unique name for this worker like 'worker-1'.
                                If used with --nprocs then the process number
                                will be appended like name-0, name-1, name-2,
                                ...
  --memory-limit TEXT           Bytes of memory per process that the worker
                                can use. This can be an integer (bytes), float
                                (fraction of total system memory), string
                                (like 5GB or 5000M), 'auto', or zero for no
                                memory management  [default: auto]
  --reconnect / --no-reconnect  Reconnect to scheduler if disconnected
                                [default: --reconnect]
  --nanny / --no-nanny          Start workers in nanny process for management
                                [default: --nanny]
  --pid-file TEXT               File to write the process PID
  --local-directory TEXT        Directory to place worker files
  --resources TEXT              Resources for task constraints like "GPU=2
                                MEM=10e9". Resources are applied separately to
                                each worker process (only relevant when
                                starting multiple worker processes with '--
                                nprocs').
  --scheduler-file TEXT         Filename to JSON encoded scheduler
                                information. Use with dask-scheduler
                                --scheduler-file
  --death-timeout TEXT          Seconds to wait for a scheduler before closing
  --dashboard-prefix TEXT       Prefix for the dashboard
  --preload TEXT                Module that should be loaded by each worker
                                process like "foo.bar" or "/path/to/foo.py"
  --version                     Show the version and exit.
  --help                        Show this message and exit.
```

jcrist:
Looks good to me. Happy to merge on tests pass.

chrish42:
The message I see for the AppVeyor run is "Build execution time has reached the maximum allowed time for your plan (60 minutes)." Search through the logs, I don't see any FAILED tests, so I assume that timeout is the problem. How do I work around that?

jrbourbeau:
Looks to be a transient issue, I just restarted the AppVeyor build

jrbourbeau:
Tests are passing now 👍 @chrish42 just for reference, there have been several recent improvements to make distributed more robust to these types of spurious failures. So you should see fewer of them in the future : ) 

Thanks @chrish42  for this contribution and @jcrist for reviewing

Issue #2788, opened 2019-06-19
dask-worker not handling KeyboardInterrupt correctly


TomAugspurger:
Connect a `dask-worker` to the scheduler and then `ctrl-c`.

That should exit cleanly.

```
2019-06-19 14:53:49,386 distributed.worker[54182] INFO -------------------------------------------------
2019-06-19 14:53:49,393 distributed.worker[54182] INFO         Registered to:    tcp://192.168.7.20:8786
2019-06-19 14:53:49,393 distributed.worker[54182] INFO -------------------------------------------------
2019-06-19 14:53:49,394 distributed.core[54182] INFO Starting established connection
^C2019-06-19 14:53:51,525 distributed.dask_worker[54155] INFO Exiting on signal 2
2019-06-19 14:53:51,526 distributed.nanny[54155] INFO Closing Nanny at 'tcp://192.168.7.20:62826'
2019-06-19 14:53:51,528 distributed.dask_worker[54155] INFO End worker
Traceback (most recent call last):
  File "/Users/taugspurger/.virtualenvs/dask-dev/bin/dask-worker", line 11, in <module>
    load_entry_point('distributed', 'console_scripts', 'dask-worker')()
  File "/Users/taugspurger/sandbox/distributed/distributed/cli/dask_worker.py", line 387, in go
    main()
  File "/Users/taugspurger/Envs/dask-dev/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/taugspurger/Envs/dask-dev/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/taugspurger/Envs/dask-dev/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/taugspurger/Envs/dask-dev/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/taugspurger/sandbox/distributed/distributed/cli/dask_worker.py", line 380, in main
    raise TimeoutError("Timed out starting worker.") from None
tornado.util.TimeoutError: Timed out starting worker.
2019-06-19 14:53:51,531 distributed.process[54155] WARNING reaping stray process <ForkServerProcess(Dask Worker process (from Nanny), started daemon)>
```

This is related to my PR. Will take a look later.

TomAugspurger:
[...]

PR #2844, opened 2019-07-15
WIP: Make workers gracefully handle sigint

jni:
Was hacking with @mrocklin trying to fix https://github.com/dask/dask-jobqueue/issues/122, ran into #2788. These are early attempts to fix both things. @TomAugspurger you might find my test code helpful, if overly verbose (I just copied the worker/scheduler creation from the test above it).

The basic idea behind the `unregister_with_scheduler` coroutine is that the cluster (apparently? @mrocklin told me this) sends SIGINT to processes before killing them for exceeding their time allocation. We can use this to close out the workers with `safe=True` so that the tasks running on them are not marked as suspicious.


jni:
(Note: I do not want to unduly associate @mrocklin with the quality of this PR (or lack thereof) — @mrocklin was just patiently sitting next to me answering my questions and pointing me to the right files and SO answers. =)

TomAugspurger:
Thanks @jni. I haven't worked any more on #2788 since a couple weeks ago when it completely stumped me.

Do you plan to continue working on this? If not, I'll try to pick it up again by the end of the week.

jni:
@TomAugspurger I'd very much appreciate you picking this up, as I don't think I can dedicate much time to it in the next few weeks...

TomAugspurger:
OK, I'll try to get #2788 sorted out today.

TomAugspurger:
Well, that whole "step away from the problem for a bit" thing actually worked.

```diff
diff --git a/distributed/cli/dask_worker.py b/distributed/cli/dask_worker.py
index e23a8ab9..5a396686 100755
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -394,6 +394,7 @@ def main(
         raise TimeoutError("Timed out starting worker.") from None
     finally:
         logger.info("End worker")
+        return 0
 
 
 def go():
```

is all I was missing for #2788 :) Now to write a test.

mrocklin:
OK, I've fiddled with this a bit.  Things seem to behave well on linux, but windows CI is unhappy.

distributed/cli/dask_worker.py
```diff
@@ -386,26 +387,32 @@ def del_pid_file():
...
     async def run():
         await asyncio.gather(*nannies)
         await asyncio.gather(*[n.finished() for n in nannies])
 
-    install_signal_handlers(loop, cleanup=on_signal)
-
+    for sig in [signal.SIGINT, signal.SIGTERM]:
+        asyncio.get_event_loop().add_signal_handler(
+            sig, functools.partial(on_signal, sig)
+        )
```
jacobtomlinson: I expect windows will not be happy with this. https://stackoverflow.com/questions/45987985/asyncio-loops-add-signal-handler-in-windows

mrocklin:
cc @jacobtomlinson

distributed/cli/dask_worker.py
```diff
@@ -386,26 +387,32 @@ def del_pid_file():
...
-        if signum != signal.SIGINT:
+        from distributed.utils import log_errors
+
+        with log_errors():
             logger.info("Exiting on signal %d", signum)
-        asyncio.ensure_future(close_all())
+            signal_fired = True
+            if signum == signal.SIGINT:
+                logger.info("Gracefully closing worker because of SIGINT call")
+                await asyncio.gather(*[n.close_gracefully() for n in nannies])
```
gerald732, line 401: Would it be reasonable to give SIGTERM the same treatment as SIGINT as well?

gerald732 (changes requested):

Issue #2525, opened 2019-02-12
Race condition between worker heartbeat and reconnect


fjetter:
I noticed a race condition between the worker heartbeat and a worker reconnect in case of a severed connection. This race condition leads to an exception raised and logged on the Scheduler. The scheduler and worker are still in a functional state but we receive a lot of misleading error log messages, in particular in combination with issue #2524

From my understanding the following is happening (please correct me if I get something wrong about the asyncio/tornado functionality)

### Worker heartbeat:
The worker heartbeat coroutine has three stages. In between all of these it is possible for another coroutine to gain control of the event loop. The stages are as followed (separated by a yield keyword in the code)
1. Setup and call to scheduler to signal heartbeat
2. handle heartbeat response and reconnect if necessary
3. finalize (nothing interesting)

### Handle scheduler / handle stream:
Continuous loop until comm is closed. Every cycle begins with a `comm.read()` and handles the response by calling a stream handler. After every cycle, control is given back to the event loop and potentially another coroutine is running.

## Race condition scenario:

Condition: Connection between scheduler is externally severed.

Scheduler: Connection to worker is broken -> remove worker
Worker `handle_schedule`r: Catch exception; Add callback to event loop to reconnect
Worker `heartbeat`: heartbeat stage 1: issue heartbeat to scheduler, response is "missing". Suspend coroutine.
Worker `handle_scheduler/reconnect`: Reconnect to scheduler, triggered by the caught exception above
Worker `heartbeat`: handle response of heartbeat with status "missing" since the worker was not registered when issuing the heartbeat -> schedule a reconnect
Worker `heartbeat`: finalize
Worker `heartbeat / reconnect`: Worker tries to reconnect but `Scheduler` raises exception because the worker is already known

Any suggestions on how to avoid this?

mrocklin:
Thank you for the detailed writeup @fjetter .  This is very useful.  Also, my apologies for the long delay in response.

Often in situations like this we set a status that other coroutines can check.  For example during this stage:

> Worker `handle_scheduler`: Catch exception; Add callback to event loop to reconnect

We might set something like 

    self.status = 'connecting'

which this stage might check to determine if it should indeed reconnect

> Worker `heartbeat`: handle response of heartbeat with status "missing" since the worker was not registered when issuing the heartbeat -> schedule a reconnect

In principle I think that something like this would work.  We would want to check all of the other uses of `status` to make sure that they remained accurate, but there are only a few of these.  

[...]

PR #2845, opened 2019-07-16
remove reconnect from heartbeat

pacman82:
addresses #2525.

Heartbeat does no longer trigger reconnect, this implies there won't be double reconnects. Is this the way to go? If so, I'd be thankful for hints how to test this.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -800,7 +800,6 @@ class Worker(ServerNode):
                 middle = (start + end) / 2
 
                 if response["status"] == "missing":
-                    await self._register_with_scheduler()
                     return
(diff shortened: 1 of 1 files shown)


mrocklin:
Presumably this was here for a reason.  Are there cases where we want the heartbeat to re-establish connections?

> Is this the way to go? 

I don't know.

I think that in order to resolve this problem someone will probably have to dive into what's going on and understand a bit more about the right way to resolve this thoroughly.  Unfortunately I don't know that anyone knows the right solution without doing a bit more study.

mrocklin:
Closing in favor of #2906

Issue #2848, opened 2019-07-16
Spec Cluster initialisation questions


jacobtomlinson:
I'm currently working on a cluster manager for [AWS ECS](https://aws.amazon.com/ecs/) called `ECSCluster` which inherits from `SpecCluster`. It takes `Scheduler` and `Worker` classes to create those resources. In order to create a scheduler or worker I need to query the AWS API ahead of time to do some bootstrapping. Currently I'm doing this in the `ECSCluster` init function and then passing the AWS IDs through to the `SpecCluster` as arguments. 

Once example bootstrapping job would be registering a template for the docker container with AWS. This only needs doing once and AWS returns the ID of the template for later use by the individual scheduler and worker tasks. Therefore doing it from within the cluster object makes sense.

However I'm now trying to make everything async and I'm running into circular issues. I want to create the template in the init and then pass the ID to the super init. However the creation task is async and the super init does the setup `self.sync`.

I can think of a few API changes that would resolve this, but would appreciate some guidance:

- Instead of doing the bootstrapping in the init, pass a reference to the cluster object to the `Scheduler` and `Worker` class instead. Then call the various `template_id = cluster.register_template()` from within the `Scheduler` and `Worker` init methods. This method would cache the ID for subsequent calls so it would only make an API call one.
```python
class Scheduler(Task):

    def __init__(self, cluster, **kwargs):
        self.cluster = cluster
        self.template_id = None
        super().__init__(**kwargs)

    def __await__(self):
        async def _():
            async with self.lock:
                await self.start()
            return self

        return _().__await__()

    def start(self):
        self.template_id = await self.cluster.get_template()


class ECSCluster(SpecCluster):

    def __init__(self, **kwargs):
        self.template = None
        scheduler = {"cls": Scheduler, "options": {"cluster": self}}
        worker = {"cls": Worker, "options": {"cluster": self}}
        super().__init__({}, scheduler, worker , **kwargs)

    async def get_template(self):
        if not self.template:
            await self.register_template()

    async def register_template(self):
        self.template = ...

```

- Instead of passing the scheduler and worker classes to the `SpecCluster` init we could pass/set them within our own init after calling super init. However we would need to tell the `SpecCluster` init that we will call `_start` instead.

```python
class ECSCluster(SpecCluster):

    def __init__(self, **kwargs):
        super().__init__(start=False, **kwargs)
        self.template = self.sync(self.get_template)
[...]

PR #2850, opened 2019-07-17
SpecCluster: move init logic into start

jacobtomlinson:
This PR builds on top of #2827 with a goal of solving #2848. Relevant changes are in commit 3e14ad6.

I've moved the scheduler creation from `__init__` to `_start`. This allows me to make async calls within my own subclassed `_start` before the scheduler object is created.

I'm also ignoring exceptions from closing the scheduler if the scheduler has already timed out.


mrocklin:
To ease review I'm going to focus first on merging in the SSH work (with
appropriate caveats) and then moving onto this.  Does that work for
you @jacobtomlinson ?

mrocklin:
In principle this looks fine to me.  I'm somewhat inclined to have `self.workers_spec` and `self.scheduler_spec` as non-None after `__init__` and `self.workers` and `self.scheduler` as non-None after `await`, but I don't have a concrete reason for this at the moment.  I'm happy to merge this as-is or make the change to move the default `self.scheduler_spec` logic back into `__init__`, while still keeping the scheduler creation in `start`

mrocklin:
Actually, nm.  I can see how the spec itself would sometimes need communication in order to determine.  Merging.

PR #2851, opened 2019-07-17
Adds docs to Client methods for resources, actors, and traverse

IPetrik:
I am kinda guessing on some of the details of accepted values to some of these, so please check that they are correct.


distributed/client.py
```diff
@@ -2629,6 +2651,21 @@ def compute(
             Higher priorities take precedence
         fifo_timeout: timedelta str (defaults to '60s')
             Allowed amount of time between calls to consider the same priority
+        traverse: bool (defaults to True)
+            By default dask traverses builtin python collections looking for
+            dask objects passed to ``compute``. For large collections this can
+            be expensive. If none of the arguments contain any dask objects,
+            set ``traverse=False`` to avoid doing this traversal.
```
IPetrik: Copied this from the dask documentation.

distributed/client.py
```diff
@@ -2629,6 +2651,21 @@ def compute(
...
+        resources: dict (defaults to {})
+            Defines the `resources` these tasks require on the worker. Can
+            specify global resources (``{'GPU': 2}``), or per-task resources
+            (``{'x': {'GPU': 1}, 'y': {'SSD': 4}}``), but not both.
+            See :doc:`worker resources <resources>` for details on defining
+            resources.
+        actors: bool or dict (default None)
+            Whether these tasks should exist on the worker as stateful actors.
+            Specified on a global (True/False) or per-task (``{'x': True,
+            'y': False}``) basis. See :doc:`actors` for additional details.
```
IPetrik: Not sure if this is true.

distributed/client.py
```diff
@@ -1379,7 +1379,7 @@ def submit(
         func: callable
         *args:
         **kwargs:
-        pure: bool (defaults to True)
+        pure: bool (defaults to opposite of `actor(s)`)
```
mrocklin: actors are very rarely used.  I'm disinclined to expect all users to check on that default value in order to understand the default value here.  I'd rather say the factually-wrong-but-almost-always-correct response of just "True" insteead.

distributed/client.py
mrocklin: Is it possible to use the `actor=` keyword with compute?

mrocklin:
Checking in here @IPetrik .  It looks like there are a few unresolved comments.  It has been a while, so I suspect that you've moved on, but I thought I'd ping you here in case you're still interested in this.

distributed/client.py
IPetrik: Do you mean, "Is it currently possible to use the `actor=` kw with `compute`?" or "Can we make it possible to use the `actor=` kw with `compute`?" or something else?

distributed/client.py
IPetrik: I'm ok with this.

IPetrik:
> 
> 
> Checking in here @IPetrik . It looks like there are a few unresolved comments. It has been a while, so I suspect that you've moved on, but I thought I'd ping you here in case you're still interested in this.

Hey, @mrocklin - sorry, got swamped with work and life for a while there...

distributed/client.py
mrocklin: I think that this is fine.  Thanks.

mrocklin:
Thanks @IPetrik .  Merging this in. 

IPetrik:
> 
> 
> Thanks @IPetrik . Merging this in.
> 

Thanks! It's my first contribution to any OSS project. 😄 Feels good.

mrocklin:
Our docs get built after every commit.  So your changes should be live
sometime in the next ten minutes or so here:
https://docs.dask.org/en/latest/futures.html#api

Issue #2524, opened 2019-02-12
Keep alive for inactive worker/scheduler connection


fjetter:
I noticed that the connection between workers and schedulers (and effectively all `Server` subclasses using [`handle_stream`](https://github.com/dask/distributed/blob/357bca686b47293062584e12543f2ce22bd8f16f/distributed/core.py#L378)) are kept open to continuously listen to incoming requests. This connection is solely used to listen to `stream_handler` requests while, for instance the `heartbeat`, uses a different connection from the connection pool. This means that for a longer time of cluster inactivity, the _primary_ connection is inactive. The issue about inactive connections is that some external tools are inclined to kill inactive connections (for instance, we're using [haproxy](http://www.haproxy.org) and we kill inactive connections after 1h) which requires for the worker to reconnect and register with the scheduler every hour during inactivity.

I was wondering if this is an issue for somebody else and whether it would be desired to either send a _keep alive_ using the open connection every X secs (simple message to keep the connection active)  or even send the hearbeat over this connection. I believe the first option could be implemented fairly easily in `handle_stream` while the latter is a bit more difficult. In either case I wanted to get some feedback before implementing anything.

(dedicated worker-scheduler connection creation see [here](https://github.com/dask/distributed/blob/0a5b8da4041414e1e621c1b8d613730720c6e684/distributed/worker.py#L575), reconnect if connection is broken see [here](https://github.com/dask/distributed/blob/0a5b8da4041414e1e621c1b8d613730720c6e684/distributed/worker.py#L660))

mrocklin:
Thank you for the excellently worded issue, and my apologies for the delay in response.

I'm hearing two possible solutions:

1.  Move the heartbeat to the long-running connection.

    I suspect that the current reason for the heartbeat to be on a separate connection is to trigger a reconnect if something goes wrong with the long-running connection.  However, as you point out in #2525 we're already handling this, so perhaps this reason is not sufficient (or in fact, problematic). 

    If so then moving the heartbeat to the long-running connection seems like a good idea

2.  Add a second keep-alive route to the server (probably in the handlers in `core.py::Server`) and add a periodic callback that sends a message every minute or so.  This is easy to do and has low impact, but slightly increases complexity. 

Either is fine.  If you were interested and have the time to investigate option 1 I think that that would be best, but it's also more work.

lr4d:
[...]

PR #2854, opened 2019-07-18
WIP: Move heartbeat to long running connection

pacman82:
Is this enough to address #2524

### DIFF
--- a/distributed/dashboard/tests/test_scheduler_bokeh.py
+++ b/distributed/dashboard/tests/test_scheduler_bokeh.py
@@ -17,7 +17,15 @@ from dask.core import flatten
 from distributed.utils import tokey, format_dashboard_link
 from distributed.client import wait
 from distributed.metrics import time
-from distributed.utils_test import gen_cluster, inc, dec, slowinc, div, get_cert
+from distributed.utils_test import (
+    gen_cluster,
+    inc,
+    dec,
+    slowinc,
+    div,
+    get_cert,
+    wait_for_heartbeat,
+)
 from distributed.dashboard.worker import Counters, BokehWorker
 from distributed.dashboard.scheduler import (
     BokehScheduler,
@@ -340,7 +348,7 @@ def test_WorkerTable_custom_metrics(c, s, a, b):
         for name, func in metrics.items():
             w.metrics[name] = func
 
-    yield [a.heartbeat(), b.heartbeat()]
+    yield wait_for_heartbeat(s, a, b)
 
     for w in [a, b]:
         assert s.workers[w.address].metrics["metric_port"] == w.port
@@ -367,7 +375,7 @@ def test_WorkerTable_different_metrics(c, s, a, b):
 
     a.metrics["metric_a"] = metric_port
     b.metrics["metric_b"] = metric_port
-    yield [a.heartbeat(), b.heartbeat()]
+    yield wait_for_heartbeat(s, a, b)
 
     assert s.workers[a.address].metrics["metric_a"] == a.port
     assert s.workers[b.address].metrics["metric_b"] == b.port
@@ -391,7 +399,7 @@ def test_WorkerTable_metrics_with_different_metric_2(c, s, a, b):
         return worker.port
 
     a.metrics["metric_a"] = metric_port
-    yield [a.heartbeat(), b.heartbeat()]
+    yield wait_for_heartbeat(s, a, b)
 
     wt = WorkerTable(s)
     wt.update()
@@ -411,7 +419,7 @@ def test_WorkerTable_add_and_remove_metrics(c, s, a, b):
 
     a.metrics["metric_a"] = metric_port
     b.metrics["metric_b"] = metric_port
-    yield [a.heartbeat(), b.heartbeat()]
+    yield wait_for_heartbeat(s, a, b)
 
     assert s.workers[a.address].metrics["metric_a"] == a.port
     assert s.workers[b.address].metrics["metric_b"] == b.port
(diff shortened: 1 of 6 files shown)


mrocklin:
I don't think so?  Perhaps you can say more about what you were intending to do here

lr4d:
This should address #2525 and #2524 (PR #2845 can probably be closed, as this builds on top of it).

The first commit removes the reconnection trigger when the worker receives `status=missing` to prevent the race condition of #2525 .

The second commit moves the heartbeat to the long running connection between worker and scheduler.
I have tested this and HAProxy no longer kills the connection between worker and scheduler after an hour, as described originally in #2524 .

Some adjustments had to be made to the tests so that they would wait for the heartbeat response from the scheduler.

@mrocklin what do you think?

mrocklin:
In principle it seems fine

> Some adjustments had to be made to the tests so that they would wait for the heartbeat response from the scheduler.

These adjustments are, I think, too invasive.  We wouldn't expect a beginner dev to know how to add these if they were making a new test on something as unrelated as dashboard code.

As an alternative, perhaps this?  https://github.com/dask/distributed/issues/2525#issuecomment-516149515

lr4d:
> These adjustments are, I think, too invasive. 

I am not too comfortable with it, either. It can be simplified however, perhaps using something like:
```
def wait_for_heartbeat(scheduler, *workers):
    [...]
```
and calling `wait_for_heartbeat(s, a, b)` instead of this (current) code block:
```
    bs = s.stream_comms[a.address]
    init = bs.batch_count
    yield [a.heartbeat(), b.heartbeat()]
    yield gen.sleep(bs.interval)
    yield gen.sleep(bs.interval)
    assert bs.batch_count > init
```
I'd appreciate suggestions in this regard.

Regarding the alternative you propose, it would probably fix the issue, but @pacman82 and I moved the heartbeat to the long running connection as we thought adding another route which would function as keep-alive would increase complexity while not adding more value

lr4d:
I've updated the last commit, now it only replaces the line `yield [a.heartbeat(), b.heartbeat()]` in the tests for ` yield wait_for_heartbeat(s, a, b)`. I think this should allow new developers to still understand and be able to create dashboard tests.

mrocklin:
> I've updated the last commit, now it only replaces the line yield [a.heartbeat(), b.heartbeat()] in the tests for yield wait_for_heartbeat(s, a, b). I think this should allow new developers to still understand and be able to create dashboard tests.

Previously we could expect a heartbeat to have occurred before a worker had started.  I don't think that we should relax this condition.  My concern isn't only on dashboard tests, but on the full system.  This seems like a regression to me.

mrocklin:
> Regarding the alternative you propose, it would probably fix the issue, but @pacman82 and I moved the heartbeat to the long running connection as we thought adding another route which would function as keep-alive would increase complexity while not adding more value

I agree that ideally heartbeats should play this role, but the complexity of https://github.com/dask/distributed/pull/2907 seems to be much less than the complexity that we've had to go through here.

lr4d:
> Previously we could expect a heartbeat to have occurred before a worker had started. I don't think that we should relax this condition. My concern isn't only on dashboard tests, but on the full system. This seems like a regression to me.

What's the rationale behind this?

lr4d:
>the complexity of #2907 seems to be much less than the complexity that we've had to go through here

Is this because of the testing code?
The reason I like this approach is because it looks to me like #2907 increases architectural complexity and maintainability relative to this PR

mrocklin:
> What's the rationale behind this?

I'm not sure that this was intentional originally, but the fact that some tests failed when we took this away means that some functionality implicitly depended on it.  I suspect that functionality outside of our test suite also depends on it.  

Looking at the tests it seems like the scheduler would like to have metrics from workers as soon as they arrive?

mrocklin:
> The reason I like this approach is because it looks to me like #2907 increases architectural complexity and maintainability relative to this PR

That PR is, from my perspective, very simple to maintain.  Adding new routes is cheap (especially if they don't do anything).  Adding PeriodicCallbacks is cheap.  Changing how we do heartbeats is more uncertain and more likely to have unexpected consequences.

lr4d:
> I'm not sure that this was intentional originally, but the fact that some tests failed when we took this away means that some functionality implicitly depended on it. I suspect that functionality outside of our test suite also depends on it.
> 
> Looking at the tests it seems like the scheduler would like to have metrics from workers as soon as they arrive?

I realized that previously the `heartbeat` method of a worker would `await` the message from the scheduler. This is causing the need for adjustments to tests as the heartbeat has not yet fully completed after `yield worker.heartbeat()` is executed.

lr4d:
Essentially, though, the heartbeat works in the same way, except that the communication channel is different and that the heartbeat waits for `worker.batched_stream.interval` before actually relaying the message to the scheduler when `worker.heartbeat` is executed (this is the logic of `batched_stream.send`).

The difficult thing (for me, at least) in this case is to wait for the heartbeat to have completed: previously a simple `yield worker.heartbeat()` would suffice and the heartbeat would have arrived at the scheduler and received a succesful response. With the current implementation, one would ideally wait for `worker.heartbeat_response` but since this a `stream_handler`, I have not yet figured out a way to do so

mrocklin:
Do the other solutions work?  If so, I'm inclined just to go with them if that's ok.

lr4d:
I didn't yet test it for the HAProxy situation but I'm pretty sure they will. I respect your decision on this matter.

martindurant:
@lr4d , @pacman82 : I have not been following the conversation here, just a gentle ping to ask where you think things currently stand

mrocklin:
This was resolved in another PR. Closing.  Thanks @pacman82 for raising this.  Thanks @martindurant for the nudge.

PR #2860, opened 2019-07-19
Document distributed.Reschedule in API docs

jrbourbeau:
Currently `distributed.Reschedule` is part of the public API, but the docstring isn't being rendered


mrocklin:
Thanks @jrbourbeau !

PR #2870, opened 2019-07-22
Remove unused variable in SpecCluster scale down

### LINKED ISSUE #2869: SpecCluster error when removing workers
When attempting to scale down a `SpecCluster` I'm getting the following error.

```python-traceback
distributed.utils - ERROR - 'Worker' object has no attribute 'worker_address'
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/utils.py", line 714, in log_errors
    yield
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/deploy/adaptive.py", line 176, in _retire_workers
    await f
  File "/home/nfs/jtomlinson/Projects/dask/distributed/distributed/deploy/spec.py", line 324, in scale_down
    if v.worker_address in workers:
AttributeError: 'Worker' object has no attribute 'worker_address'
```

jacobtomlinson:
Fixes #2869.

I've made a few changes here to avoid issues I was having when scaling down a `SpecCluster`.

- Remove unused `to_close ` variable and comment.
- Update `worker_address` to `address`.
- Change `await self` to `self.loop.add_callback(self._correct_state)`. This was to make it consistent with `scale`. Happy to update `scale` if that approach is preferred.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -319,10 +319,10 @@ class SpecCluster(Cluster):
         workers = set(workers)
 
         for k, v in self.workers.items():
-            if v.address in workers:
+            if v.worker_address in workers:
                 del self.worker_spec[k]
 
-        self.loop.add_callback(self._correct_state)
+        await self
 
     scale_up = scale  # backwards compatibility
 
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -318,13 +318,11 @@ class SpecCluster(Cluster):
     async def scale_down(self, workers):
         workers = set(workers)
 
-        # TODO: this is linear cost.  We should be indexing by name or something
-        to_close = [w for w in self.workers.values() if w.address in workers]
         for k, v in self.workers.items():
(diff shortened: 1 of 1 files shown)


mrocklin:
It looks like `scale_down` is currently only used in `Adaptive.retire_workers`, and is intended to block until the workers have been fully removed.  I think that there is just a confusion around names here.  The `scale` method changes a target (fast, nonblocking) while `scale_down` actually does the work of closing workers using the resource manager (slow).

As discussed elsewhere I think that we're planning to stick with the `worker_address` attribute.   Is that the case?

jacobtomlinson:
If the `worker_address` issue has been addressed elsewhere (#2874) then this PR is probably not useful. 

Although I am curious about the `to_close = ...` line as it doesn't seem to do anything.

mrocklin:
It's probably an artifact of history.  We should probably remove it.

jacobtomlinson:
I've taken out my changes and turned this PR into only removing the unused variable.

PR #2871, opened 2019-07-22
Replace gen.coroutine with async/await in core

mrocklin:
This replaces the use of Tornado `gen.coroutine`s and the `yield` keyword with `async def` functions and the `await` keyword.  Additionally, because these have slightly different semantics this also changes a few internal bits to smooth over this transition.

This is still a work in progress.  The current implementation introduces a couple of small bugs and the conversion is not complete.

cc @jcrist (who requested this a while ago) and @jacobtomlinson who might find this easier to work with than gen.coroutine.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -48,7 +48,6 @@ from .security import Security
 from .utils import (
     All,
     ignoring,
-    get_ip,
     get_fileno_limit,
     log_errors,
     key_split,
@@ -1178,11 +1177,9 @@ class Scheduler(ServerNode):
             return ws.host, port
 
     @gen.coroutine
-    def start(self, addr_or_port=None, start_queues=True):
+    def start(self):
         """ Clear out old state and restart all running coroutines """
         enable_gc_diagnosis()
-
-        addr_or_port = addr_or_port or self._start_address
 
         self.clear_task_state()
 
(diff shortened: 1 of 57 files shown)


mrocklin:
This is getting close to mergeable, and includes fixes for issues in master.  I'm going to try to clean things up here today or tomorrow and hopefully merge in shortly afterwards.

mrocklin:
There were some cases where I've decided to keep gen.coroutine semantics.  In particular asyncio complains if we leave a task un-awaited in a way that it is bothersome for users.  In a couple of places I didn't find a nice way to await every task (such as in the case of the TCP Comm.write method) so I left these in the tornado style.

mrocklin:
Tests are now starting to look good :)

mrocklin:
OK, I'm going to pronounce this as "good enough" and merge in.  I'll probably do a follow-up in a couple of days.

distributed/scheduler.py
```diff
@@ -1305,7 +1314,6 @@ def close_worker(self, stream=None, worker=None, safe=None):
     # Stimuli #
     ###########
 
-    @gen.coroutine
```
lr4d: `async` is not used here. Is this intentional?

mrocklin: Probably?  Is this causing some concrete issue for you?  Does this function need to be asynchronous?

PR #2872, opened 2019-07-22
Add support for CUDA serialization to host

pentschev:
This PR adds support for serialization of CUDA objects to host. The current implementation allows serialization while keeping data on the device (useful for communication with UCX, for example), but to spill memory from the device to host, we need to actually move that data device<->host.

@mrocklin @quasiben could you take a look at this? It improves the memory spilling of CUDA objects and is the simplest change for dask-cuda to operate under those circumstances.

### DIFF
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -87,5 +87,7 @@ def _register_numba():
 
 @cuda_serialize.register_lazy("cudf")
 @cuda_deserialize.register_lazy("cudf")
+@cuda_host_serialize.register_lazy("cudf")
+@cuda_host_deserialize.register_lazy("cudf")
 def _register_cudf():
     from . import cudf
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -79,6 +79,8 @@ def _register_cupy():
 
 @cuda_serialize.register_lazy("numba")
 @cuda_deserialize.register_lazy("numba")
+@cuda_host_serialize.register_lazy("numba")
+@cuda_host_deserialize.register_lazy("numba")
 def _register_numba():
     from . import numba
 
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -4,7 +4,7 @@ from functools import partial
 
 from .compression import compressions, default_compression
 from .core import dumps, loads, maybe_compress, decompress, msgpack
-from .cuda import cuda_serialize, cuda_deserialize
+from .cuda import cuda_serialize, cuda_deserialize, cuda_host_serialize, cuda_host_deserialize
 from .serialize import (
     serialize,
     deserialize,
@@ -71,6 +71,8 @@ def _register_torch():
 
 @cuda_serialize.register_lazy("cupy")
 @cuda_deserialize.register_lazy("cupy")
+@cuda_host_serialize.register_lazy("cupy")
+@cuda_host_deserialize.register_lazy("cupy")
 def _register_cupy():
     from . import cupy
 
--- a/distributed/protocol/cuda.py
+++ b/distributed/protocol/cuda.py
@@ -1,3 +1,5 @@
+from functools import partial
+
 import dask
 
 from . import pickle
@@ -6,12 +8,14 @@ from dask.utils import typename
 
 cuda_serialize = dask.utils.Dispatch("cuda_serialize")
 cuda_deserialize = dask.utils.Dispatch("cuda_deserialize")
+cuda_host_serialize = dask.utils.Dispatch("cuda_serialize")
+cuda_host_deserialize = dask.utils.Dispatch("cuda_deserialize")
 
 
-def cuda_dumps(x):
+def cuda_dumps(x, dispatcher=cuda_serialize):
     type_name = typename(type(x))
     try:
-        dumps = cuda_serialize.dispatch(type(x))
+        dumps = dispatcher.dispatch(type(x))
     except TypeError:
         raise NotImplementedError(type_name)
 
@@ -24,10 +28,15 @@ def cuda_dumps(x):
     return header, frames
 
 
-def cuda_loads(header, frames):
+def cuda_loads(header, frames, dispatcher=cuda_deserialize):
     typ = pickle.loads(header["type-serialized"])
-    loads = cuda_deserialize.dispatch(typ)
+    loads = dispatcher.dispatch(typ)
     return loads(header, frames)
 
 
 register_serialization_family("cuda", cuda_dumps, cuda_loads)
+register_serialization_family(
+    "cuda_host",
+    partial(cuda_dumps, dispatcher=cuda_host_serialize),
+    partial(cuda_loads, dispatcher=cuda_host_deserialize)
+)
--- a/distributed/protocol/cudf.py
+++ b/distributed/protocol/cudf.py
@@ -1,5 +1,7 @@
+import numpy as np
 import cudf
-from .cuda import cuda_serialize, cuda_deserialize
+from numba import cuda
+from .cuda import cuda_serialize, cuda_deserialize, cuda_host_serialize, cuda_host_deserialize
 from .numba import serialize_numba_ndarray, deserialize_numba_ndarray
 
 
@@ -47,7 +49,7 @@ def serialize_cudf_dataframe(x):
 
 
 @cuda_deserialize.register(cudf.DataFrame)
-def serialize_cudf_dataframe(header, frames):
+def deserialize_cudf_dataframe(header, frames):
     columns = header["columns"]
     n_columns = len(header["columns"])
     n_masks = len(header["null_subheaders"])
@@ -72,3 +74,17 @@ def serialize_cudf_dataframe(header, frames):
         pairs.append((name, series))
 
     return cudf.DataFrame(pairs)
+
+
+@cuda_host_serialize.register(cudf.DataFrame)
+def serialize_cudf_host_dataframe(x):
+    header, frames = serialize_cudf_dataframe(x)
+    frames = [f.copy_to_host() if isinstance(f, cuda.cudadrv.devicearray.DeviceNDArray)
+              else f for f in frames]
+    return header, frames
+
+
+@cuda_host_deserialize.register(cudf.DataFrame)
+def deserialize_cudf_host_dataframe(header, frames):
+    frames = [cuda.to_device(f) if isinstance(f, np.ndarray) else f for f in frames]
+    return deserialize_cudf_dataframe(header, frames)
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -2,7 +2,7 @@
 Efficient serialization GPU arrays.
 """
 import cupy
-from .cuda import cuda_serialize, cuda_deserialize
+from .cuda import cuda_serialize, cuda_deserialize, cuda_host_serialize, cuda_host_deserialize
 
 
 @cuda_serialize.register(cupy.ndarray)
@@ -40,3 +40,15 @@ def deserialize_cupy_array(header, frames):
         pass
     arr = cupy.asarray(frame)
     return arr
+
+
+@cuda_host_serialize.register(cupy.ndarray)
+def serialize_cupy_host_ndarray(x):
+    header, frames = serialize_cupy_ndarray(x)
+    frame, = frames
+    return header, [frame.get()]
+
+
+@cuda_host_deserialize.register(cupy.ndarray)
+def deserialize_cupy_host_array(header, frames):
+    return deserialize_cupy_array(header, frames)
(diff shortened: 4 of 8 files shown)


mrocklin:
@pentschev I'm still curious if there is a way that we can make cuda-device-host transfers work well without adding a new system into the Dask mainline library.

Currently the `cuda_serialize` serializers convert a device-memory-backed Python object (like a CuPy array or CuDF dataframe) into a header and a bunch of frames, where each frame includes the `__cuda_array_interface__` property.  

I would hope that we could combine this existing function along with a function that converted a frame-with-the-`__cuda_array_interface__` keyword into a host-side bytes-like object in order to get what we wanted.  My hope is that by combining these two more general operations we could acheive the same result that we want (spilling from device to host) without having to create a new general operation.

I could be missing things here though.  I haven't actually dived into this problem yet.

pentschev:
> @pentschev I'm still curious if there is a way that we can make cuda-device-host transfers work well without adding a new system into the Dask mainline library.

Could you clarify what you mean by "new system"? I think you mean libraries such as CuPy and Numba, but they have already been added with the original CUDA serialization code. Alternatively, we could move those to `dask-cuda`, I think?

> Currently the cuda_serialize serializers convert a device-memory-backed Python object (like a CuPy array or CuDF dataframe) into a header and a bunch of frames, where each frame includes the __cuda_array_interface__ property.

Correct, except that both Numba and CuPy use currently a single frame, only cuDF creates multiple frames.

> I would hope that we could combine this existing function along with a function that converted a frame-with-the-__cuda_array_interface__ keyword into a host-side bytes-like object in order to get what we wanted. My hope is that by combining these two more general operations we could acheive the same result that we want (spilling from device to host) without having to create a new general operation.

I assume what you mean by "new general operation" is the addition of `cuda_host`, correct me if I misunderstood. That's exactly what I couldn't do otherwise, we need a way to identify whether we're serializing for communication (e.g., UCX without copying data to host) or for spilling to host.

I may have not understood all the serialization code properly, but I didn't see a way to serialize `(header, frames)` with the existing code. I thought about adding such a function, but I think it may be more cluttered than the solution proposed in this PR.

mrocklin:
By "new system" I mean a new function, like `cuda_host_serialize`.  This is specific enough that I'd prefer to avoid promising to support it into the future if we can avoid it.

> That's exactly what I couldn't do otherwise, we need a way to identify whether we're serializing for communication (e.g., UCX without copying data to host) or for spilling to host

I'm not sure that I agree with this.

> but I didn't see a way to serialize (header, frames) with the existing code

These are, from my perspective, already serialized.  There is nothing left to do.  I would just convert each of the frames to live on the host, and then leave it at that.

I think that the one piece we're missing is a function that takes a device-side-frame and turns it into a host-side-frame.  This shouldn't need to be specific to each individual input type (cudf, cupy, ...) because the current `cuda_serialize` function should handle the work of normalizing all of those down to more simple frame objects.  My hope is that we're able to construct this function with a few lines of numba code.

So if we wanted to move a device-backed Python object (like a cupy array or cudf dataframe) into host memory then I think that the process would be something like the following:

```python
header, device_frames = cuda_serialize(obj)  # use existing serialization code to normalize things
host_frames = [move_to_host(frame) for frame in device_frames]  
del device_frames

host_dict[key] = (header, host_frames)
```

We're still missing the `move_to_host` function.  Fortuantely though, that function can be fairly simple, because it will be the same if it's operating on a cupy array or cudf dataframe.  All it needs is an object that implements the `__cuda_array_interface__` attribute.  This `move_to_host` function is the thing that I was thinking could probably be written with a few lines of Numba.

pentschev:
> I think that the one piece we're missing is a function that takes a device-side-frame and turns it into a host-side-frame.

But this function wouldn't be a new one, so the client-code (e.g., dask-cuda), would need to add code to use that function in the moment it needs to move data to host, right?

> We're still missing the move_to_host function. Fortuantely though, that function can be fairly simple, because it will be the same if it's operating on a cupy array or cudf dataframe.

A cuDF dataframe is actually composed by a bunch of Numba arrays, so it's already handled the same way.

> All it needs is an object that implements the __cuda_array_interface__ attribute. This move_to_host function is the thing that I was thinking could probably be written with a few lines of Numba.

I honestly don't see why we need to do that through Numba, instead of just using the library's implementation that should be capable of handling even the more complex cases without the need for us to potentially convert metadata. Could you elaborate why is this the design choice here?

mrocklin:
> But this function wouldn't be a new one, so the client-code (e.g., dask-cuda), would need to add code to use that function in the moment it needs to move data to host, right?

Yes, dask-cuda would need to implement and then use this function when it needs it.  I'm entirely comfortable with this living outside of the mainline dask codebase.

>> We're still missing the move_to_host function. Fortuantely though, that function can be fairly simple, because it will be the same if it's operating on a cupy array or cudf dataframe.

> A cuDF dataframe is actually composed by a bunch of Numba arrays, so it's already handled the same way.

Right.  I think that we're in agreement here.  Mostly I wanted to ensure that we don't need to make a separate one of these functions for every type that we encounter in the future.  We already do this once with `cuda_serialize` and I'd prefer to not have to do it again if possible.  

-   `cuda_serialize`: any object -> list of frames  # most of the complexity of dealing with different types is in this one function
-  `device_to_host_copy`: a frame -> host data  # so that this can be fairly simple

> I honestly don't see why we need to do that through Numba, instead of just using the library's implementation that should be capable of handling even the more complex cases without the need for us to potentially convert metadata. Could you elaborate why is this the design choice here?

Which library?  Are you proposing writing a separate "convert this thing to host memory" for every gpu-enabled library that we want to deal with?  I think that that's a reasonable approach.  I was just trying to avoid having to implement many different functions for every type that Dask has to interoperate with. We already do this once with `cuda_serialize`, so I was hoping to reuse that logic, which I think should work fine.

I'm not saying that we have to use Numba, just that Numba is probably a pretty good library today to handle device-to-host transfers of objects that implement the `__cuda_array_interface__` protocol.  One could write the same code in Cython or C I'm sure just as well.  I would be against having it in cupy or something though, because that's a bit less dependably around.

pentschev:
> > But this function wouldn't be a new one, so the client-code (e.g., dask-cuda), would need to add code to use that function in the moment it needs to move data to host, right?
> 
> Yes, dask-cuda would need to implement and then use this function when it needs it. I'm entirely comfortable with this living outside of the mainline dask codebase.
> 
> > > We're still missing the move_to_host function. Fortuantely though, that function can be fairly simple, because it will be the same if it's operating on a cupy array or cudf dataframe.
> 
> > A cuDF dataframe is actually composed by a bunch of Numba arrays, so it's already handled the same way.
> 
> Right. I think that we're in agreement here. Mostly I wanted to ensure that we don't need to make a separate one of these functions for every type that we encounter in the future. We already do this once with `cuda_serialize` and I'd prefer to not have to do it again if possible.
> 
> * `cuda_serialize`: any object -> list of frames  # most of the complexity of dealing with different types is in this one function
> * `device_to_host_copy`: a frame -> host data  # so that this can be fairly simple
> 
> > I honestly don't see why we need to do that through Numba, instead of just using the library's implementation that should be capable of handling even the more complex cases without the need for us to potentially convert metadata. Could you elaborate why is this the design choice here?
> 
> Which library? Are you proposing writing a separate "convert this thing to host memory" for every gpu-enabled library that we want to deal with? I think that that's a reasonable approach. I was just trying to avoid having to implement many different functions for every type that Dask has to interoperate with. We already do this once with `cuda_serialize`, so I was hoping to reuse that logic, which I think should work fine.
> 
> I'm not saying that we have to use Numba, just that Numba is probably a pretty good library today to handle device-to-host transfers of objects that implement the `__cuda_array_interface__` protocol. One could write the same code in Cython or C I'm sure just as well. I would be against having it in cupy or something though, because that's a bit less dependably around.

In here, I'm just asking if there's a problem with calling `cupy_array.get()`/`cupy.asarray(numpy_array)` on CuPy arrays and `numba_device_array.copy_to_host()`/`numpy_array.to_device()` for Numba arrays. There's some dependency (that we need to know about both CuPy and Numba), but we get rid of the complexity of potentially needing to convert a CuPy array's metadata into a Numba array, thus ensuring we only implement a single way of copying data back-and-forth. I'm not sure if this is exactly what you were proposing, but it is what I'm trying to avoid.

mrocklin:
Right, the `copy_device_to_host` function I'm proposing definitely would not assume that the input type is a cupy array or a numba device array (or a pytorch tensor, or a tensorflow tensor, or a ...).  It would only assume that the object it was given implemented the `__cuda_array_interface__` standard.  It would look at the information in that dictionary, and perform a device_to_host copy using lower level techniques and return something like a Python `memoryview` or `bytes` object.  It would definitely not call anything like `cupy.get`.

I think that some of our confusion arises from the accidental fact that in the cupy case `frame` is actually still a cupy array.  This is fine, but odd, and atypical.  Usually frames are just very bare containers of bytes without any additional functionality.  In general once something is a frame we should not assume that it has any functionality other than that it implements `__cuda_array_interface__`.  So even though in the cupy case we could call `.get()`, this won't be general, so we probably shouldn't use that in general code.

pentschev:
> Right, the `copy_device_to_host` function I'm proposing definitely would not assume that the input type is a cupy array or a numba device array (or a pytorch tensor, or a tensorflow tensor, or a ...). It would only assume that the object it was given implemented the `__cuda_array_interface__` standard. It would look at the information in that dictionary, and perform a device_to_host copy using lower level techniques and return something like a Python `memoryview` or `bytes` object. It would definitely not call anything like `cupy.get`.
> 
> I think that some of our confusion arises from the accidental fact that in the cupy case `frame` is actually still a cupy array. This is fine, but odd, and atypical. Usually frames are just very bare containers of bytes without any additional functionality. In general once something is a frame we should not assume that it has any functionality other than that it implements `__cuda_array_interface__`. So even though in the cupy case we could call `.get()`, this won't be general, so we probably shouldn't use that in general code.

Indeed, calling `.get()` from CuPy won't be general. However, there are various complications for not doing that. As you noticed before, we would first need a general-purpose function (perhaps in Cython) to do the `__cuda_array_interface__` handling between device and host. Second, we would need to ensure synchronization of the CUDA stream in use prior to accessing to copying a buffer to host/after copying it back to device, and that would require us to find out the CUDA stream that each backend is using at that moment.

I understand your wish for a general-purpose function, and I share that wish, but it currently it's not the case that we can ignore those complications. Besides that, we need a lot of implementation-specific logic for handling an array's metadata. That said, I don't really see why we want to avoid an implementation-specific device<->host data copying functions when we already need to handle each array type differently when doing creating the header/frames, but please do point out the potential setbacks I'm probably missing at the moment.

mrocklin:
> we need a lot of implementation-specific logic for handling an array's metadata

That is already handled in the cuda_serialize function, which places all of the array's metadata into the header.  There is nothing in the frame, other than the bytes, that we need.  If that's not the case, then it's a bug.

To be totally clear, we should think of the frames as only holding bytes and we should not depend on any other functionality or metadata that they may have.

> That said, I don't really see why we want to avoid an implementation-specific device<->host data copying functions when we already need to handle each array type differently when doing creating the header/frames, but please do point out the potential setbacks I'm probably missing at the moment.

If we have to make a separate function for each operation we want to do with these things then I think that it will become increasingly difficult for new types to work with Dask.  Copying from host to device and back is just one operation.  I think that others will arrive in the future, as well as new data types.  I'd prefer not to have to special case all of them.

Maybe for now we just depend on the presence of Numba as the simplest of the options, and use that to convert each frame to a Numba device array and use methods on that object to move the data to host memory.

pentschev:
> That is already handled in the cuda_serialize function, which places all of the array's metadata into the header. There is nothing in the frame, other than the bytes, that we need. If that's not the case, then it's a bug.
> 
> To be totally clear, we should think of the frames as only holding bytes and we should not depend on any other functionality or metadata that they may have.

The headers are only partially handled, for example, non-contiguous and 0d arrays are not handled at the moment. If you look at the implementation to handle both CuPy and Numba headers, these are 20+ lines, whereas copying is only a single line for each direction.

> If we have to make a separate function for each operation we want to do with these things then I think that it will become increasingly difficult for new types to work with Dask. Copying from host to device and back is just one operation. I think that others will arrive in the future, as well as new data types. I'd prefer not to have to special case all of them.

What would be other possible operations? In the context of serialization, I don't see any other operations that makes sense rather than moving data around.

> Maybe for now we just depend on the presence of Numba as the simplest of the options, and use that to convert each frame to a Numba device array and use methods on that object to move the data to host memory.

We have to also consider other issues, such as synchronization. Once we started passing CUDA pointers (via `__cuda_array_interface__`), unlike CPU memory views, we have to ensure that all the enqueued operations in the stream(s) accessing that buffer have finished and we're actually accessing the data at the correct point in time, given that today each library handles the CUDA stream that it operates on. If we only get a CuPy data pointer and treat it as a Numba data pointer at any time, we risk accessing corrupt data.

mrocklin:
> The headers are only partially handled, for example, non-contiguous and 0d arrays are not handled at the moment.

I would consider that a bug then.  What happens when we try to communicate a non-contiguous or 0d array across a wire?

pentschev:
We've solved this in https://github.com/rapidsai/dask-cuda itself, thus I'm closing this PR.

PR #2874, opened 2019-07-23
Add ProcessInterface

jacobtomlinson:
Added `SpecProcess` which is an interface for custom schedulers and workers to inherit from for use in `SpecCluster`.

Not particularly excited about the name of this class. Suggestions welcome!

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -11,7 +11,7 @@ from ..scheduler import Scheduler
 from ..security import Security
 
 
-class SpecProcess:
+class ProcessInterface:
     """ An interface for Scheduler and Worker processes for use in SpecCluster
 
     Parameters
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -37,11 +37,6 @@ class ProcessInterface:
 
         return _().__await__()
 
-    @property
-    def worker_address(self):
-        """ For API compatibility with Nanny. """
-        return self.address
-
     async def start(self):
         """ Start the process. """
         self.status = "running"
@@ -368,7 +363,7 @@ class SpecCluster(Cluster):
         # TODO: this is linear cost.  We should be indexing by name or something
         to_close = [w for w in self.workers.values() if w.address in workers]
         for k, v in self.workers.items():
-            if v.worker_address in workers:
+            if getattr(v, "worker_address", v.address) in workers:
                 del self.worker_spec[k]
 
         await self
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -11,6 +11,53 @@ from ..scheduler import Scheduler
 from ..security import Security
 
 
+class SpecProcess:
+    """ An interface for Scheduler and Worker processes for use in SpecCluster
+
+    Parameters
+    ----------
+    loop:
+        A pointer to the running loop.
+
+    """
+
+    def __init__(self, loop=None):
+        self.address = None
+        self.loop = loop
+        self.lock = asyncio.Lock()
+        self.status = "created"
+
+    def __await__(self):
+        async def _():
+            async with self.lock:
+                if self.status == "created":
+                    await self.start()
+                    assert self.status == "running"
+            return self
+
+        return _().__await__()
+
+    @property
+    def worker_address(self):
+        """ For API compatibility with Nanny. """
+        return self.address
+
+    async def start(self):
+        """ Start the process. """
+        self.status = "running"
+
+    async def close(self):
+        """ Close the process. """
+        self.status = "closed"
+
+    async def logs(self):
+        """ Generator which yeilds log messages. """
+        raise NotImplementedError()
+
+    def __repr__(self):
+        return "<%s: status=%s>" % (type(self).__name__, self.status)
+
+
 class SpecCluster(Cluster):
     """ Cluster that requires a full specification of workers
 
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -45,10 +45,6 @@ class ProcessInterface:
         """ Close the process. """
         self.status = "closed"
 
-    async def logs(self):
-        """ Generator which yeilds log messages. """
-        raise NotImplementedError()
-
     def __repr__(self):
         return "<%s: status=%s>" % (type(self).__name__, self.status)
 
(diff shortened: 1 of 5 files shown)


distributed/deploy/spec.py
```diff
@@ -11,6 +11,53 @@
...
+                    await self.start()
+                    assert self.status == "running"
+            return self
+
+        return _().__await__()
+
+    @property
+    def worker_address(self):
+        """ For API compatibility with Nanny. """
+        return self.address
```
mrocklin: This attribute seems to only make sense for the worker/nanny type objects.  Maybe this encourages us to have two such interface classes?

mrocklin: I guess it feels odd to split things for only this one method though.

distributed/deploy/spec.py
```diff
@@ -11,6 +11,53 @@
...
+    Parameters
+    ----------
+    loop:
+        A pointer to the running loop.
+
+    """
+
+    def __init__(self, loop=None):
+        self.address = None
+        self.loop = loop
```
mrocklin: Is a pointer to the event loop required by SpecCluster in order to run?

distributed/deploy/spec.py
```diff
@@ -11,6 +11,53 @@
 from ..security import Security
 
 
+class SpecProcess:
```
mrocklin: `ProcessInterface` ?

distributed/deploy/spec.py
jacobtomlinson: Yes. I fixed a bug on Monday in ECSCluster where the adaptive code was trying to schedule a callback on the loop.

distributed/deploy/spec.py
jacobtomlinson: I agree, it feels odd either way. Having two interfaces seems a bit wasteful though.

distributed/deploy/spec.py
jacobtomlinson: I guess that more clearly describes what it is. Still not sure though.

distributed/deploy/spec.py
mrocklin: I've made our use of `worker_address` a bit more permissive.  Hopefully this sidesteps this issue.

distributed/deploy/spec.py
mrocklin: But perhaps the ECSCluster object can have a pointer to its own event loop, rather than relying on Worker/Scheduler objects to have them.

distributed/deploy/spec.py
mrocklin: I've made this change for now.  I've also removed this from the top level API.

mrocklin:
I've pushed a few changes up to your fork.  I hope you don't mind.

distributed/deploy/spec.py
```diff
@@ -11,6 +11,48 @@
...
+        """ Start the process. """
+        self.status = "running"
+
+    async def close(self):
+        """ Close the process. """
+        self.status = "closed"
+
+    async def logs(self):
+        """ Generator which yeilds log messages. """
+        raise NotImplementedError()
```
mrocklin: I suspect that in other resource managers we won't get a stream of logs, we'll just get a dump of the logs such as they are right now.  

- @jacobtomlinson What is the situation in Kubernetes?
- @jcrist What is the situation in Yarn?

mrocklin: Also https://github.com/dask/dask-jobqueue/issues/299

mrocklin: Actually, I probably should have raised this conversation at https://github.com/dask/distributed/pull/2875

Perhaps people can respond there instead.

jacobtomlinson: Currently in kubernetes you get a dump. However when moving to async kubernetes we will get a stream. 

We can always encourage folks to present a dump as a stream for consistency anyway.

```python
for line in dump.split('\n'):
    yield line
```

distributed/deploy/spec.py
jacobtomlinson: Sure. That will require a change in adaptive where is it trying to access the loop of the scheduler.

jacobtomlinson: I can't actually seem to find it now.

distributed/deploy/spec.py
jacobtomlinson: From #2875 we should do something better than raise a `NotImplementedError` like get logs back from the scheduler.

In the case that this is the scheduler we can open a `rpc` to `self.address` and call `logs()`. In the case that this is a worker we also need the same rpc, then call `worker_logs` and figure out which is ours.

jacobtomlinson: Another option would be to put this logic up in the `SpecCluster` where we catch the `NotImplementedError` and gather these details curectly from the `cluster.scheduler_comm` up there.

jcrist: There's no public api for getting logs in yarn besides the yarn CLI (which implements all of it in the CLI instead of a public method :/). Currently users can get running logs through skein's Web UI (click on the app in the yarn web ui and you get taken here), or use `yarn logs -applicationId appid` to get logs after completion.

I likely could figure out how to get logs as a method and add that to Skein if desired. Besides a few setup scripts though the logs should be equivalent to using the `get_scheduler_logs`/`get_worker_logs` methods.

mrocklin: Currently `get_scheduler_logs`/`get_worker_logs` only grab output from the `distributed` Python logger, and nothing else.  I suspect that Yarn captures everything from stdout/stderr, which is a bit nicer.  In general I think that having a uniform way for users to get logs would have some value.  If it's easy for us to do and we can coordinate some way on how to do it cleanly then I think we should do it.

jcrist: I agree, I'm just saying that having a default method that uses the rpc for now (as @jacobtomlinson suggested) would be good, and other implementations can override with their own implementations where possible.

mrocklin: Ah great.  It sounds like we're in agreement.  I'll draft something up and submit a PR for general review by the deployment folks sometime hopefully today.

mrocklin: In the mean time, I'm going to remove the logs bit from this PR and merge this in.  OK with you @jacobtomlinson ?

jacobtomlinson: Sure :)

PR #2875, opened 2019-07-23
Add logs method to spec cluster

jacobtomlinson:
This PR adds a `logs()` method to `SpecCluster` to allow the retrieval of logs from the scheduler and all workers.

- Pulled in the `Log` and `Logs` classes from dask-kubernetes with fancy reprs. (not sure if utils is the right place for them)
- Added async generator `logs()` to SSH cluster 2 processes which iterate over stderr from the ssh connection
- Added `logs()` method to `SpecCluster` which bundles all logs together into a dictionary of strings, which are then cast to `Logs` and `Log` for nice reprs.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -272,7 +272,7 @@ class SpecCluster(Cluster):
 
     def __del__(self):
         if self.status != "closed":
-            self.close()
+            self.loop.add_callback(self.close)
 
     def __enter__(self):
         self.sync(self._correct_state)
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -6,7 +6,7 @@ from tornado import gen
 
 from .cluster import Cluster
 from ..core import rpc, CommClosedError
-from ..utils import LoopRunner, silence_logging, ignoring
+from ..utils import LoopRunner, silence_logging, ignoring, Logs, Log
 from ..scheduler import Scheduler
 from ..security import Security
 
@@ -327,6 +327,24 @@ class SpecCluster(Cluster):
         await self
 
     scale_up = scale  # backwards compatibility
+
+    def logs(self):
+        async def get_logs(task):
+            log = ""
+            async for line in task.logs():
+                log += "{}\n".format(line)
+            return Log(log)
+
+        scheduler_logs = {
+            "scheduler - {}".format(self.scheduler.address): self.sync(
+                get_logs, self.scheduler
+            )
+        }
+        worker_logs = {
+            "worker {} - {}".format(key, worker.address): self.sync(get_logs, worker)
+            for key, worker in self.workers.items()
+        }
+        return Logs({**scheduler_logs, **worker_logs})
 
     def __repr__(self):
         return "%s(%r, workers=%d)" % (
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -328,7 +328,7 @@ class SpecCluster(Cluster):
 
     scale_up = scale  # backwards compatibility
 
-    def logs(self):
+    async def _logs(self):
         async def get_logs(task):
             log = ""
             async for line in task.logs():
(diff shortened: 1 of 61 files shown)


distributed/utils.py
```diff
@@ -1489,3 +1489,29 @@ def format_dashboard_link(host, port):
...
+
+class Logs(dict):
+    """A container for multiple logs."""
+
+    def _widget(self):
+        from ipywidgets import Accordion
+
+        accordion = Accordion(children=[log._widget() for log in self.values()])
+        [accordion.set_title(i, title) for i, title in enumerate(self.keys())]
+        return accordion
```
mrocklin: Thoughts on using details and summary tags instead here?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

I think that it's nicer to use raw HTML and `_repr_html_` over widgets if possible, both for simplicity, and for static notebook rendering.

distributed/deploy/spec.py
```diff
@@ -328,6 +328,24 @@ def new_worker_spec(self):
...
+        scheduler_logs = {
+            "scheduler - {}".format(self.scheduler.address): self.sync(
+                get_logs, self.scheduler
+            )
+        }
+        worker_logs = {
+            "worker {} - {}".format(key, worker.address): self.sync(get_logs, worker)
+            for key, worker in self.workers.items()
+        }
+        return Logs({**scheduler_logs, **worker_logs})
```
mrocklin: Typically we try to support both synchronous and asynchronous modes by having two functions:

```python
async def _logs(self):
    ...

def logs(self):
    return self.sync(self._logs)
```

The `self.sync` method takes care to return the async future if we're in an asynchronous context or to run things in the event loop itself, block, and return the result if we're in a synchronous context.

mrocklin: We might also provide a default implementation using `self.scheduler_comm.logs` and `self.scheduler_comm.worker_logs`.  This queries the scheduler and workers for logs using Dask communication, but is a bit worse, because it only picks up messages logged to the `distributed` logging handler.

distributed/utils.py
jacobtomlinson: That would definitely be neater. However there is [no support for those tags in Microsoft flavor browsers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Browser_compatibility).

Is that an acceptable trade off?

distributed/deploy/spec.py
jacobtomlinson: I hadn't considered getting logs over the Dask communication. 

I would feel better if the process classes were responsible for providing their own logs on demand by the cluster. Going through the `cluster.scheduler_comm` sidesteps that. Ideally the process classes would speak directly to the resources they have created, but I guess we can't assume that the workers are routable from our Python session.

mrocklin: I agree that it would be better to get logs from the resource manager, this will have the advantage of getting things like stdout as well.

I'm just proposing using the Dask methods as a default implementation, in case the whoever implements this class doesn't bother to implement a logs method.

distributed/utils.py
mrocklin: Hrm, that's a shame. 

> Is that an acceptable trade off?

I don't know.  I suspect that you know more about the deployment of browsers and eventual adoption of these tags.

distributed/deploy/spec.py
jacobtomlinson: Ah gotcha. So instead of the `raise NotImplementedError`.

distributed/utils.py
jacobtomlinson: I expect usage of Microsoft browsers within the Dask community is low. However that is just a feeling and not backed up with any evidence, might be nice to try and find out more specifically (shame we missed the survey). If we look at [stats for the general population](https://en.wikipedia.org/wiki/Usage_share_of_web_browsers) usage of these browsers is around 15%.

Microsoft have a track record of refusing to implement some standards in their browsers. Given that all other browsers support these tags it's likely that they do not intend to implement these.

I feel like introducing something that will not work for up to 15% of users isn't good. If that number was 5% of lower I would be fine with it.

mrocklin: What happens in MS browsers?  Do they fall back sensibly?  Perhaps they just render the summaries?  If so then this might not be that bad.

distributed/deploy/spec.py
mrocklin: Yes.   That.

distributed/utils.py
mrocklin: I was curious enough to fire up my VM.  The details/summary tags seem to work, at least on IE11, even if they aren't officially supported.

mrocklin: Ah, no, details/summary don't work, but do they do seem to just render the fully expanded version, which seems good enough for me.

jacobtomlinson: > Ah, no, details/summary don't work, but do they do seem to just render the fully expanded version, which seems good enough for me.

jacobtomlinson:
Now with simpler html only repr.

image

distributed/deploy/spec.py
jacobtomlinson: Ok I'll look at that over in #2874.

mrocklin:
I've trimmed this down to include only the Log/Logs objects.  I'm continuing work in https://github.com/dask/distributed/pull/2889

PR #2877, opened 2019-07-23
Remove `Client.upload_environment`

jcrist:
This method was undocumented and broken. Modifying the user environment
is better handled by the deployment solution, the current method is not
resilient to worker additions/removals, and uploading and replacing the
current environment is an inefficient method to add additional packages.
If users really want to upload packages to each worker at runtime, they
can use `upload_file` with a zipfile/egg.

Fixes https://github.com/dask/dask/issues/5142.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -9,7 +9,6 @@ import copy
 from datetime import timedelta
 import errno
 from functools import partial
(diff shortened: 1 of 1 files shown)


mrocklin:
+1

Issue #2830, opened 2019-07-08
cuDF UCX Deserialization Failure


quasiben:
When naively trying to deserialize a distributed cuDF dataframe an error is raised in cuDF but I suspect the issue may be with the deserialization scheme in distributed.

```python
import os
import dask
import distributed
import cudf
import dask.array as da
import dask.dataframe as dd

print(f'Dask: {dask.__version__}')
print(f'Distributed: {distributed.__version__}')
print(f'cuDF: {cudf.__version__}')

n_rows = 10
n_keys = 3

ddf = dd.concat([
    da.random.random(n_rows).to_dask_dataframe(columns='x'),
    da.random.randint(0, n_keys, size=n_rows).to_dask_dataframe(columns='id'),
], axis=1).persist()

print(ddf.head())

print(ddf.map_partitions(cudf.from_pandas).head().to_pandas()) # succeeds 

from dask.distributed import Client, wait
from dask_cuda import DGX

cluster = DGX(CUDA_VISIBLE_DEVICES=[0,1,2,3], dashboard_address='0.0.0.0:8989')
client = Client(cluster)
client

ddf2 = dd.concat([
    da.random.random(n_rows).to_dask_dataframe(columns='x'),
    da.random.randint(0, n_keys, size=n_rows).to_dask_dataframe(columns='id'),
], axis=1).persist()


ddf2.map_partitions(cudf.from_pandas).head().to_pandas() # fails:
```

[Gist to Notebook](https://gist.github.com/quasiben/69c4edaa415c86b400c8ca19d78366a3)

## Error

<details>

```
distributed.protocol.core - CRITICAL - Failed to deserialize
Traceback (most recent call last):
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/protocol/core.py", line 126, in loads
    value = _deserialize(head, fs, deserializers=deserializers)
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/protocol/serialize.py", line 190, in deserialize
    return loads(header, frames)
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/protocol/cuda.py", line 30, in cuda_loads
    return loads(header, frames)
  File "/home/nfs/bzaitlen/GitRepos/cudf/python/cudf/cudf/dataframe/dataframe.py", line 181, in deserialize
    index = deserialize(index_header, index_frames)
[...]

PR #2881, opened 2019-07-24
update cudf serialization/deserialization

quasiben:
This in conjunction with:
- https://github.com/rapidsai/cudf/pull/2378
- https://github.com/rapidsai/ucx-py/pull/129

Will fix: https://github.com/dask/distributed/issues/2830

I would not merge this until https://github.com/rapidsai/cudf/pull/2378 is merged in


distributed/__init__.py
```diff
@@ -35,3 +35,19 @@
...
+def f(idx):
+    while True:
+        time.sleep(1)
+        frame = sys._current_frames()[idx]
+        print("-" * 100)
+        print("".join(call_stack(frame)))
+
+
+thread = threading.Thread(target=f, args=(threading.get_ident(),), daemon=True)
+# thread.start()
```
mrocklin: We'll want to clear this out

distributed/protocol/cudf.py
```diff
@@ -1,74 +1,20 @@
...
-        "is_cuda": len(arrays),
-        "subheaders": sub_headers,
-        # TODO: the header must be msgpack (de)serializable.
-        # See if we can avoid names, and just use integer positions.
-        "columns": x.columns.tolist(),
-        "null_counts": null_counts,
-        "null_subheaders": null_headers,
-    }
+    header, frames = x.serialize()
+    header["is_cuda"] = 1
```
mrocklin: Do we need this?  I can see that it is added in all of cudf, cupy, numba, but I'm not sure that it is actually used anywhere.

We an address that in a follow-on PR though.

quasiben:
@mrocklin I think this is ready to be merged in.  There are still issues with running a distributed join with cudf+ucx+dask.  Folks are investigating ucx-py and cudf and I think it best to get everyone on "master"

mrocklin:
OK, there is a dask release happening later today.  Should this go in as part of that release (is it critical?) or can we merge it after the release occurs?  I have a preference for afterwards, just so that it gets a bit more use by bleeding edge users before it goes into a proper release.

quasiben:
Afterwards makes sense.  Thanks!

Issue #2886, opened 2019-07-26
Dashboard Address incorrect reported


quasiben:
When launching a cluster on the CLI with UCX the dashboard address is incorrect reported after launch:

```
(cudf-dev) bzaitlen@dgx15:~$ dask-scheduler --protocol ucx  --host ucx://10.33.225.165:42841 --dashboard-address 10.33.227.165:8973
distributed.scheduler - INFO - -----------------------------------------------
distributed.scheduler - INFO - Local Directory:    /tmp/scheduler-1ebdby3z
distributed.scheduler - INFO - -----------------------------------------------
distributed.scheduler - INFO - Clear task state
distributed.scheduler - INFO -   Scheduler at: ucx://10.33.225.165:42841
distributed.scheduler - INFO -   dashboard at:        10.33.225.165:8973
```

Notice the defined address is **10.33.227.165:8973** where as the reported address is **10.33.225.165:8973** .  It should also be noted that the address does in fact launch on 10.33.227.165:8973

mrocklin:
I suspect that this is due to odd networking rules on your network.
Locally for machines within your network I suspect that the .225 address is
correct while for machines outside your network the .227 is correct.

You may also want to use the option `--dashboard-address :8973` if you are
ok with open access.

quasiben:
.225 is the network for the UCX device and .227 is the network for TCP -- both are on the same machine.  I believe https://github.com/dask/distributed/pull/2887 should resolve the issue

PR #2887, opened 2019-07-26
User Service Address for Report

quasiben:
PR fixes: https://github.com/dask/distributed/issues/2886


distributed/scheduler.py
```diff
@@ -1213,7 +1213,11 @@ def get_worker_service_addr(self, worker, service_name, protocol=False):
             self.status = "running"
             logger.info("  Scheduler at: %25s", self.address)
             for k, v in self.services.items():
-                logger.info("%11s at: %25s", k, "%s:%d" % (listen_ip, v.port))
+                if hasattr(v, "server"):
+                    ip = v.server.address
```
mrocklin: A few questions:

When does this come up?  Presumably for the BokehServer stuff?

What if this object has a `server` attribute but not an `address` attribute?  

Does address typically mean just the IP or the full ip/port pair?

Issue #2861, opened 2019-07-19
Gracefully restart a worker from a nanny


mrocklin:
A common request is that people want to restart a single worker into a clean state.  This might be to refresh the imported software environment or to clear out leaked memory.  To do this cleanly a worker needs to stop accepting work, offload its data to peers, and then close itself and let the nanny restart it.  We can do all of these steps today, but it's a bit of a manual process.  It would be nice to make it easier.

As an example use case, long-running Dask workers (days or weeks) may want to refresh themselves every hour

```
dask-worker scheduler-address:8786 --refresh-restart 1h
```

filmor:
This is also connected to #1823 in my opinion (i.e. allowing explicit restart via a `Client`).

AUtomatic restart would be nice as well, though :)

PR #2892, opened 2019-07-27
Close workers gracefully

mrocklin:
This allows workers to optionally terminate themselves gracefully after a predetermined time.  This can be helpful in a few contexts:

1.  We receive a SIGINT, and know that we need to clean up quickly (though note that the signal handlers are not implemented as part of this PR (cc @jni))
2.  We know that we'll be kicked off at a certain time, such as in one hour from now, as is often specified by HPC job schedulers (cc @lesteve @guillaumeeb @jhamman )
3.  We just want to refresh our workers every once in a while, because we know that our code leaks some memory .  Fixes https://github.com/dask/distributed/issues/2861 

This is configurable as keywords to the `Worker` or `Nanny` classes, in config values, or with CLI.  Here is an example with CLI.

### Restart to clear state

```
dask-worker scheduler:8786 --lifetime 1hr --lifetime-restart --lifetime-stagger 5m
```

This will kill the worker roughly 1 hour from now +- a range of 5 minutes (to avoid killing all of our workers at the same time).  It will also allow that worker to be restarted afterwards

### Restart to avoid walltime death

```
dask-worker scheduler:8786 --lifetime 58m 
```

Here we don't try to restart the worker (no point) and we choose a time a bit before our 60m walltime.

Also cc @jacobtomlinson @jcrist for general deployment information

For folks who want to review.  I recommend going commit-by-commit (it should be fairly clean).  (this is also a decently simple/educational PR to review)


mrocklin:
Merging this tomorrow if there are no further comments (Although comments would be quite welcome here)

TomAugspurger:
> though note that the signal handlers are not implemented as part of this PR

Do you anticipate this making fixing https://github.com/dask/distributed/issues/2788 easier? https://github.com/dask/distributed/pull/2892 makes it sound like "yes".

mrocklin:
> Do you anticipate this making fixing #2788 easier?

It's the action that I think we would want to take when handling a SIGINT.

mrocklin:
Merging this shortly if there are no further comments

distributed/worker.py
```diff
@@ -252,6 +252,16 @@ class Worker(ServerNode):
         Resources that this worker has like ``{'GPU': 2}``
     nanny: str
         Address on which to contact nanny, if it exists
+    lifetime: str
+        Amount of time like "1 hour" after which we gracefully shut down the worker.
+        This defaults to None, meaning no explicit shutdown time.
+    lifetime_stagger: str
+        Amount of time like "5 minutes" to stagger the lifetime value
+        The actual lifetime will selected uniformly at random between
```
guillaumeeb:
```suggestion
        The actual lifetime will be selected uniformly at random between
```

distributed/cli/dask_worker.py
```diff
@@ -175,6 +175,30 @@
 @click.option(
     "--dashboard-prefix", type=str, default="", help="Prefix for the dashboard"
 )
+@click.option(
+    "--lifetime",
+    type=str,
+    default="",
+    help="If provided, shut down the worker after this duration. "
+    "Note that if combined with --nanny (default) "
+    "the worker will restart after this time",
```
guillaumeeb: Even with lifetime_restart to false as default?

mrocklin: Thanks.  I've removed this.  I think that it's now clear with the `--lifetime-restart` keyword just below.

lesteve:
Just a quick comment (I don't think this is a problem for this PR, since the signal handling is only part of it AFAICT): for SGE the signal you get when the job exceed the walltime is a `SIGKILL` (e.g. `kill -9`). Not sure for other schedulers.

jacobtomlinson:
Ouch! AFAIK Slurm and PBS send a `SIGTERM` shortly before the `SIGKILL` to give the process a chance to exit cleanly.

lesteve:
Good to know that other schedulers are a bit nicer regarding job termination.

I found [this](http://disbauxes.upc.es/code/analysis/automatic-job-resubmit-using-sun-grid-engine/) which seems to indicate that for SGE `SIGKILL` is used by default but that you can configure it to be something else.

PR #2902, opened 2019-07-28
Request feed of worker information from Scheduler to SpecCluster

mrocklin:
The SpecCluster now queries the Scheduler to get up-to-date information about workers and services.  This is useful for things like ...

1.  Adaptivity
2.  Dashboard links
3.  Displaying cores/memory in the dashboard
4.  Available services

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -41,6 +41,7 @@ from .comm import (
 from .comm.addressing import address_from_user_args
 from .compatibility import finalize, unicode, Mapping, Set
 from .core import rpc, connect, send_recv, clean_exception, CommClosedError
+from .diagnostics.plugin import SchedulerPlugin
 from . import profile
 from .metrics import time
 from .node import ServerNode
@@ -1266,7 +1267,7 @@ class Scheduler(ServerNode):
         self.periodic_callbacks.clear()
 
         self.stop_services()
-        for ext in self.extensions:
+        for ext in self.extensions.values():
             with ignoring(AttributeError):
                 ext.teardown()
         logger.info("Scheduler closing all comms")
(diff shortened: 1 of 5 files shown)


mrocklin:
@jcrist here I propose that Cluster objects include a `scheduler_info` attribute that mirrors the output of the `Scheduler.identity()` function.  This would then be used by downstream consumers like dask-labextension, and a shared Widget mixin (we could move the widget code from here into a mixin if you're game).

In this PR SpecCluster tracks the scheduler's state with the following, fairly generic code: 

https://github.com/dask/distributed/pull/2902/files#diff-e95e1ccb9336311eae957db482e0be4dR224-R249

I could also move this to Cluster if desired.

mrocklin:
See for example https://github.com/dask/dask-labextension/pull/73 on how this enables dask-labextension to work with SpecCluster

distributed/scheduler.py
```diff
@@ -4963,3 +4964,52 @@ def __init__(self, task, last_worker):
...
+            del v["metrics"]
+            del v["last_seen"]
+        return ident
+
+    def add_worker(self, worker=None, **kwargs):
+        if self.started:
+            ident = self.scheduler.workers[worker].identity()
+            del ident["metrics"]
+            del ident["last_seen"]
+            self.bcomm.send(["add", {"workers": {worker: ident}}])
```
jcrist: A few questions (not familiar with `BatchedSend` objects, and skimming their implementation didn't clarify things):

- If no client has subscribed to this stream, what happens here with this `send` call?
- What happens if two different clients call `subscribe-worker-status`? I can't tell for sure, but it looks like this would likely cause problems, since it would replace the existing `comm`?
- What happens if the client disconnects after subscribing?

If this plugin is all no-ops when no client has subscribed, and works with multiple subscribers, then I think this is fine to always have in `DEFAULT_EXTENSIONS`, otherwise it may be better to have off.

mrocklin: > If no client has subscribed to this stream, what happens here with this send call?

The fact that we're in this method means that someone is subscribing.

> What happens if two different clients call subscribe-worker-status?

Good point.  After a quick fix (I'll push a bit later this afternoon), there will be one plugin per connection

It's also worth noting that, currently, this is only being used by Cluster objects, so I would only expect one connection per scheduler.

> What happens if the client disconnects after subscribing?

The send calls would generate errors, which would start to fill logs.

It's worth noting again though that clients aren't likely to call this, only cluster managers (or at least that's the intent).

I can resolve this though.

Maybe a general question, in the architecture of Dask-Yarn, is it likely that two cluster managers might both connect up to the same scheduler? (I remember something like this being possible).

mrocklin: I think that, more broadly, my question is:

*"Would it make sense for YarnCluster to also provide a `scheduler_info` attribute?"*

I'm building functionality on top of this and want to ensure that that functionality can be useful generally both for SpecCluster and YarnCluster, assuming that those two implementations remain separate.

jcrist: What functionality are you building on top of this? This could be done for YarnCluster, sure. Multiple cluster manager could be attached to the same cluster, but that's not the normal mode of operation.

The questions I'm asking above are just general "is this plugin robust" questions.

mrocklin: > What functionality are you building on top of this?

-  I'd like to make the widget depend on it, rather than assuming that a Scheduler object was present
-  Dask-Labextension: https://github.com/dask/dask-labextension/pull/73
-  I'll also probably want to use it in Adaptive so that we have a sense of 
    1. what the user asked for (`SpecCluster.worker_spec`)
    2. what we've successfully requested (`SpecCluster.workers`)
    3.  what has actually arrived (`Cluster.scheduler_info["workers"]`)

    Though I suspec that we could also get by with just 2 and 3, which are probably more generic.

mrocklin: > The questions I'm asking above are just general "is this plugin robust" questions.

Yup.  And that's appreciated.

PR #2904, opened 2019-07-29
Support slowly arriving workers in SpecCluster

mrocklin:
Previously SpecCluster waited until all workers had checked in with the
scheduler.  This made sense for LocalCluster or SSHCluster because
there isn't really a significant delay in starting things that we can't
control.  However, for other systems like dask-jobqueue or
dask-kubernetes workers might not ever start, so we need a different
system.

Now, SpecCluster still awaits the Worker object that it is passed, but
doesn't require that the worker has started in the scheduler.  We now
expect awaiting to mean

*"We have successfully handed control of starting the worker to some other robust system"*

Our job at this point is done.  We hope that the worker arrives, but
from our perspective this local Worker object is awaited and "running".

This commit also includes a minimal example of a Worker class,
`SlowWorker`, that serves as a nice minimal example for what SpecCluster
expects.

I plan to do a bit more work here and review our Adaptive policies when workers 
may take a while.  I hope that this allows us to replicate/replace a lot of the 
fine work that @jhamman did over at dask-jobqueue.

cc @jcrist @jacobtomlinson


mrocklin:
@jcrist would you mind reviewing the adaptive_core.py file?  I think that it abstracts away all of the adaptive logic from assumptions about the cluster class.  It would be both to get your perspective as a controls person, and also as someone who might want to connect this logic to another cluster class type (or two) in the future.

mrocklin:
Moving ahead with this.  I'm more than happy to change things after merging.

PR #2907, opened 2019-07-29
Add keep-alive message between worker and scheduler

### LINKED ISSUE #2524: Keep alive for inactive worker/scheduler connection

mrocklin:
This is effectively a heartbeat, but much simpler and less frequent than
our current heartbeats

Fixes #2524

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1025,7 +1025,7 @@ class Scheduler(ServerNode):
             "missing-data": self.handle_missing_data,
             "long-running": self.handle_long_running,
             "reschedule": self.reschedule,
-            "keep-alive": lambda comm: None,
+            "keep-alive": lambda *args, **kwargs: None,
         }
 
         client_handlers = {
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1025,6 +1025,7 @@ class Scheduler(ServerNode):
             "missing-data": self.handle_missing_data,
             "long-running": self.handle_long_running,
             "reschedule": self.reschedule,
+            "keep-alive": lambda comm: None,
         }
 
         client_handlers = {
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -3319,7 +3319,7 @@ def test_get_foo_lost_keys(c, s, u, v, w):
     client=True,
     Worker=Nanny,
     worker_kwargs={"death_timeout": "500ms"},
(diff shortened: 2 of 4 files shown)


lr4d:
Getting this error on the scheduler side:
```
2019-08-01 13:30:48,297 ERROR    <lambda>() got an unexpected keyword argument 'worker' (distributed.core)
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/core.py", line 477, in handle_stream
    handler(**merge(extra, msg))
TypeError: <lambda>() got an unexpected keyword argument 'worker'
```

mrocklin:
Thanks @lr4d .  Handled

Also, what is a good time for the frequency here?  Every minute?  Every ten minutes?  Every hour?

lr4d:
I'd keep it below 5-10 minutes. For HAProxy the treshold for killing a connection on which no data is sent is 60 minutes, but I don't know what this may be for similar tools.

lr4d:
This is appears to be working fine now for our HAProxy setup, I left the cluster alive for 6 hours and no disconnections or error messages took place.
Thanks @mrocklin

PR #2913, opened 2019-08-01
Move core functionality from SpecCluster to Cluster

mrocklin:
This moves standard functionality from SpecClsuter to the Cluster
superclass.  It also removes the assumption that the Scheduler will be
local to the Cluster class.

cc @jcrist

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -29,11 +29,13 @@ class Cluster(object):
 
     scheduler_comm: rpc  # need to implement this
     scheduler_address: str  # need to implement this
+    _supports_scaling = True
 
     def __init__(self, asynchronous):
         self.scheduler_info = {}
         self.periodic_callbacks = {}
         self._asynchronous = asynchronous
+
         self.status = "created"
 
     async def _start(self):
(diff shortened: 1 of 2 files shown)


mrocklin:
@mmccarty , this is what came out of our meeting.  We wanted to find the set of functionality in SpecCluster that was general enough to be used within projects like dask-yarn and dask-gateway.

Hopefully everything here is generic enough, but we'll have to test things out to see.

Also cc @jacobtomlinson , though I suspect that he is busy.

distributed/deploy/cluster.py
```diff
@@ -24,45 +26,84 @@
...
+        await self._watch_worker_status_comm.close()
+        await self._watch_worker_status_task
+
+        for pc in self.periodic_callbacks.values():
+            pc.stop()
+        self.scheduler_comm.close_rpc()
+
+        self.status = "closed"
+
+    def close(self, timeout=None):
```
jcrist: Should `_close`/`close` be idempotent? I think right now it is just because all the operations in `_close` happen to be idempotent themselves, but it would maybe be good to be explicit with this.

distributed/deploy/cluster.py
```diff
@@ -24,45 +26,84 @@
...
+        with ignoring(RuntimeError):  # loop closed during process shutdown
+            return self.sync(self._close, callback_timeout=timeout)
+
+    def __del__(self):
+        if self.status != "closed":
+            self.loop.add_callback(self.close)
```
jcrist: Is it safe to use the loop in `__del__` in all scenarios?

mrocklin: It's worked well so far (I think?)

It's a good point though.  For now I've protected it against Attribute and Runtime errors (runtime errors arise when the loop is closed, as might be the case during shutdown)

distributed/deploy/cluster.py
mrocklin: I've added a check to pass if the status is closed.  However there are still cases where this gets called twice that could be problematic.

jcrist:
I'm currently looking at using this, and found a few things that need changing (mostly around startup, shutdown, and inadequate separation from `SpecCluster`). AFAIK there's no non-me projects looking at using this class directly currently? Trying to determine if `SpecCluster` is the only public interface I need to worry about breaking.

mrocklin:
That's correct.

Issue #2336, opened 2018-11-05
Unable to use published datasets in a different client


nikhilrajrs:
I get the error "Inputs contain futures that were created by another client" when I try to join a published dataset within a client different from the one that originally published it. The whole flow can be summarised as follows:

**Client 1**
df = dd.read_csv(...)
client.persist(df)
client.publish_dataset(ds_name=df)

**Client 2**
df = client.get_dataset("ds_name")
df2 = dd.read_csv(...)
df2.join(df)
client.persist(df2)

Is this expected behavior?

For reference, this check was added in the commit https://github.com/dask/distributed/commit/c02ea63ca2d0225ef52382e69165343d02f197cd#diff-96a27223dc91b5c9ea3d03684d79ad3f] which is part of the pull request https://github.com/dask/distributed/pull/2227

nikhilrajrs:
Sorry for being not precise enough.
The code I use is indeed:
`names = dd.read_csv('/shared-store-path/names.csv')
names = names.set_index("ID")
names = client1.persist(names)
client1.publish_dataset(names=names)`
I have attached a full example that can be used to reproduce the behavior. Shown below is the output from the notebook I used to test -



```python
from dask.distributed import Client
client1 = Client("scheduler-address:8786")
client2 = Client("scheduler-address:8786")
```


```python
from dask.distributed import wait
import dask.dataframe as dd

names = dd.read_csv('/shared-store-path/names.csv')
names = names.set_index("ID")

names = client1.persist(names)
wait(names)
```




    DoneAndNotDoneFutures(done={<Future: status: finished, type: DataFrame, key: ('sort_index-6f9583110f3b77c24727c1e970735470', 0)>}, not_done=set())




```python
client1.publish_dataset(names=names)
```


```python
roles = dd.read_csv('/shared-store-path/roles.csv')
roles = roles.set_index("ID")

names_dataset = client2.get_dataset("names")

roles = roles.join(names_dataset)
```


```python
roles.head()
```




<div>
<style scoped>
    .dataframe tbody tr th:only-of-type {
        vertical-align: middle;
    }

    .dataframe tbody tr th {
        vertical-align: top;
    }

    .dataframe thead th {
        text-align: right;
    }
</style>
<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>Role</th>
      <th>Name</th>
    </tr>
    <tr>
      <th>ID</th>
      <th></th>
      <th></th>
    </tr>
[...]

PR #2918, opened 2019-08-02
Resolve scheduler address in get_client

mrocklin:
This helps to normalize scheduler addresses before comparison

Fixes https://github.com/dask/distributed/issues/2336

I'm working on a test now, but I thought I'd push this up in the meantime.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2924,7 +2924,7 @@ def get_client(address=None, timeout=3, resolve_address=True):
     except ValueError:  # could not find worker
         pass
     else:
-        if not address or worker.scheduler.address == address:
+        if not address or comm.resolve_address(worker.scheduler.address) == address:
             return worker._get_client(timeout=timeout)
 
     from .client import _get_global_client


mrocklin:
cc @jacobtomlinson

mrocklin:
So previously we would pass the address along with serialized Futures for two reasons:

1.  To verify that the client we're connecting to is connected to the right scheduler (this comes up in some advanced multi-client/multi-scheduler situations)
2.  To create a new client and connect it to a scheduler if no client is found locally

When we call `get_client()` from within a worker, we don't need to worry about 2 (we already have a fine connection to the scheduler) but we do still need to verify that this is the right cluster for us to be on.  Using the address for this verification was error prone, because network names are odd.  

Now we also pass along the scheduler ID string, and use that for verification instead.

jacobtomlinson:
Just given this a test with [this MRE](https://github.com/dask/distributed/issues/2336#issuecomment-519027492). 

On this branch it hangs for me at `client.publish_dataset(df=df)`.

mrocklin:
@quasiben in case you're interested.

PR #2921, opened 2019-08-03
Give 404 when requesting nonexistent tasks or workers

### LINKED ISSUE #2910: [UI][FEATURE] Invalid worker or task keys should return 404 instead of 500 HTML status code
E.g. http://host:8787/info/task/foobared.html returns `500: Internal Server Error`.

Should instead return a friendlier 404 page.

martindurant:
Fixes #2910

### DIFF
--- a/distributed/dashboard/scheduler_html.py
+++ b/distributed/dashboard/scheduler_html.py
@@ -27,6 +27,9 @@ class Workers(RequestHandler):
 class Worker(RequestHandler):
     def get(self, worker):
         worker = escape.url_unescape(worker)
+        if worker not in self.server.workers:
+            self.send_error(404)
+            return
         with log_errors():
             self.render(
                 "worker.html",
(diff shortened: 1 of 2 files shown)


mrocklin:
Cool.  Thanks for doing this @martindurant !  It's also nice to see more people get involved in the Tornado codebase.

One small thought, thoughts on responding with a more informative error message?  After googling around for a bit (I didn't know how to do this) it seems like this is possible by setting the status code, then writing a normal response, and then returning (I think?) 

https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.set_status
https://www.tornadoweb.org/en/stable/guide/structure.html#error-handling

No worries if you'd like to just be done though.  This is clearly a nice improvement.

martindurant:
Indeed, this error is just the simplest thing, and there are many custom 40* pages, etc.; I would say it's maybe worthwhile in most cases to include an informative message, but here that message would only be "task %s not found" or somesuch, so it doesn't actually add any information.

I'll just merge this as-is, and we can consider more general error-handling on the HTTP routes another time.

mrocklin:
Sounds good.  Just to be clear, "Task ... not found" or "Worker ... not
found" is way more informative that 404

PR #2929, opened 2019-08-06
LINT: Fixup black string normalization

jcrist:
After running black, several places in our codebase were rewritten from
something like

```
raise ValueError("part one of the message "
                 "part two")
```

to

```
raise ValueError("part one of the message " "part two")
```

This fixes those cases, removing the unnecessary two-part string.

Stumbled across this when doing some other work, and it was a
quick fix. See also https://github.com/dask/dask/pull/5227.

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -215,7 +215,7 @@ def main(
         port=port,
         dashboard_address=dashboard_address if dashboard else None,
         service_kwargs={"dashboard": {"prefix": dashboard_prefix}},
-        **kwargs,
+        **kwargs
     )
     logger.info("Local Directory: %26s", local_directory)
     logger.info("-" * 47)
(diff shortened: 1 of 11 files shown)


distributed/cli/dask_scheduler.py
```diff
@@ -139,7 +139,7 @@ def main(
     tls_cert,
     tls_key,
     dashboard_address,
-    **kwargs
+    **kwargs,
```
mrocklin: This excludes Python 3.5 .  You'll have to update your black to master, or the commit in the pre-commit config file.   :/

jcrist: Ah, didn't realize this required a non-released version of black.

jcrist: Found the lines that were confusing black and fixed them, should now work fine with the latest black release or master.

mrocklin:
Thanks @jcrist !

PR #2932, opened 2019-08-06
Pass GPU diagnostics from worker to scheduler

mrocklin:
This does a few things:

1.  Use `pynvml` to collect information about any CUDA GPUs present
2.  Optionally add those metrics to the worker's initial handshake and
    heartbeats
3.  Collect that information in the scheduler in the WorkerState object

For now these just hang out in the scheduler information,
but in the future they might be used for dashboards,
or possibly scheduling decisions in the future.

I believe that everything gpu-specific here is fairly well separated
and generalized (others should be able to follow this pattern to add
more diagnostics relatively easily) but it would be good to hear from
others on if this is out of scope.

### DIFF
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -1,0 +1,20 @@
+import pynvml
+
+pynvml.nvmlInit()
+count = pynvml.nvmlDeviceGetCount()
+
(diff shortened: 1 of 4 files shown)


TomAugspurger:
In general, this kind of special / extra information seems fine, as long as it doesn't affect the common case when the GPUs isn't present (and this one is fine).

PR #2938, opened 2019-08-07
Import from collections.abc

jcrist:
Silences deprecation warnings about importing from collections instead of collections.abc.


jcrist:
I looked at adding a test for this, but couldn't get pytest's `filterwarnings` to catch this. I think it's because they're warnings on *import*, and pytest's warning filtering only comes into affect during tests (where the warning will have already occurred).

mrocklin:
+1 if CI passes

PR #2940, opened 2019-08-07
Alternative scheduling for new tasks

TomAugspurger:
Rather than placing new tasks with no dependencies on the first
idle worker, we try placing them on a worker executing tasks they're
a co-dependency with. This helps to reduce memory usage of graphs like

```
        a-1   a-2  a-3   a-4  a-5  a-6
          \    |    /     \    |    /
              b-1             b-2
```

This is meant to address #2602. Will require some testing

I'm writing up a bunch of benchmarks on synthetic workloads now. Will try out on some real workloads as well.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3596,6 +3596,84 @@ class Scheduler(ServerNode):
                 pdb.set_trace()
             raise
 
+    def decide_worker_initial(self, ts: TaskState) -> WorkerState:
+        r"""
+        Decide the worker to assign a task to.
+
+        Parameters
+        ----------
+        ts : TaskState
+            This is a ready task that has no dependencies.
+
+        Returns
+        -------
+        WorkerState
+
+        Notes
+        -----
+        This prioritizes scheduling a task on a worker executing tasks that we're
+        a sibling dependency with.
+
+        Consider the following task graph (read top to bottom)
+
+        ::
+
+            a-1  a-2  a-3  a-4
+             \   /     \   /
+              b-1       b-2
+
+        If we have
+
+        * Two workers: `A` and `B`
+        * Task `a-1` is running on worker `A`
+        * We're currently scheduling task `a-2`
+
+        we'll choose to schedule `a-2` it on worker `A` as well because that will
+        minimize communication when `b-1` is eventually scheduled.
+
+        When we dont have any sibling tasks running, we assign `ts` to an idle worker,
+        or a worker with occupancy / a relatively low number of tasks.
+
+        See Also
+        --------
+        decide_worker
+        worker_objective
+        transition_waiting_processing
+        """
+        worker = None
+        # the config is just for ease of testing / benchmarking. Will remove
+        if dask.config.get("distributed.scheduler.lump_tasks", default=True):
+            for dts in ts.dependents:
+                # Figure out where my siblings are running. Note that we
+                # stop as soon as we find *a* sibling running *somewhere*.
+                # If time weren't an issue, we might find the worker with the
+                # most siblings. But that's expensive.
+                #
+                for sts in dts.dependencies:
+                    if sts.processing_on:
+                        # c[sts.processing_on] += 1
+                        worker = sts.processing_on
+                        break
+                    if sts.who_has:
+                        worker = random.choice(sts.who_has)
+                        break
+
+        if worker:
+            # ((worker, n_tasks),) = c.most_common(1)
+            return worker
+        elif self.idle:
+            if len(self.idle) < 20:  # smart but linear in small case
+                worker = min(self.idle, key=operator.attrgetter("occupancy"))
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -3596,6 +3596,84 @@ def transition_no_worker_waiting(self, key):
...
+        worker = None
+        # the config is just for ease of testing / benchmarking. Will remove
+        if dask.config.get("distributed.scheduler.lump_tasks", default=True):
+            for dts in ts.dependents:
+                # Figure out where my siblings are running. Note that we
+                # stop as soon as we find *a* sibling running *somewhere*.
+                # If time weren't an issue, we might find the worker with the
+                # most siblings. But that's expensive.
+                #
+                for sts in dts.dependencies:
```
mrocklin: There are several situations where a single task has very many dependents.  In these cases I think that we'll hit N^2 scaling and bring things down.

mrocklin: What about cases where we don't have siblings, but cousins n'th removed

```
a1
|
a2    b1
 \   /
   c
```

distributed/scheduler.py
```diff
@@ -3596,6 +3596,84 @@ def transition_no_worker_waiting(self, key):
...
+        This prioritizes scheduling a task on a worker executing tasks that we're
+        a sibling dependency with.
+
+        Consider the following task graph (read top to bottom)
+
+        ::
+
+            a-1  a-2  a-3  a-4
+             \   /     \   /
+              b-1       b-2
```
mrocklin: Note that all ascii art diagrams in the codebase so far have computation going from bottom to top.  This is also the way that visualize works.

mrocklin:
A long while ago we used to schedule things differently if they call came in in the same batch.  We wouldn't do things one by one, we would take all of the initially free tasks, sort them by their dask.order value, and then partition them among the workers in that order.  This worked well because nodes that have similar ordering values are likely to be closely related.

However this works poorly if...
-  There are *many* leaves and our sorting algorithm is O(nlogn)
-  New tasks come in that aren't part of this
-  These tasks actually all have a single dependency
-  The workers have some pre-existing load that we want to take into account
-  Our graph isn't just a big collections computation, but the ordering between tasks is really important, and so we want to highly prioritize the tasks that ordering suggests we prioritize by moving them to all of the workers rather than on one.

distributed/scheduler.py
TomAugspurger: Yes, my initial got a full count of where each of our co-dependencies was running. That blew up very quickly. The early `break` once we find a co-dependency was a first attempt to avoid that.

mrocklin:
As an aside, a common cause of the graphs that you're dealing with come from, I think, not doing high-level-graph fusion aggressively enough.  I think that if we had data ingestion operations fused as we currently fuse blockwise that this situation would occur much less frequently.  This is a less general solution, but handling it well would be an unambiguous benefit, while core scheduling always has tradeoffs.  

I don't know the exact operation that you're trying to deal with, but it might be better handled by bringing operations like read_parquet, from_array, and others, under the Blockwise banner.

TomAugspurger:
> A long while ago we used to schedule things differently if they call came in in the same batch. We wouldn't do things one by one, we would take all of the initially free tasks, sort them by their dask.order value, and then partition them among the workers in that order.

That's interesting to hear. I briefly looked into trying to fix things earlier on since it's so hard to satisfy the "schedule co-dependencies together" goal this late in the scheduling process (at the single-task level).  I didn't explore it much, since it seems to go against how things are done currently.

>  I think that if we had data ingestion operations fused as we currently fuse blockwise that this situation would occur much less frequently.

Does too aggressive of fusion have a negative impact when you have multiple threads per worker?  e.g. with


```
             b-1                b-2
          /  / \  \          /  / \  \
         /  |   |  \        /  |   |  \
       a-1 a-2 a-3 a-4    a-5 a-6 a-7 a-8
```

we might want to ensure that `a-1` through `a-4` end up on the same machine, but we might not want to fuse them.

---

I'll look into blockwise a bit. Perhaps updating Xarray's `open_mfdataset` to use it would yield some improvements.

distributed/scheduler.py
TomAugspurger: This approach won't help in that case (I think `a-1` and `b-1` are niblings 😄).

mrocklin:
> Does too aggressive of fusion have a negative impact when you have multiple threads per worker? e.g. with

Maybe,  but it's not common with collections (which is where we'll get high level blockwise fusion), where we commonly have far more partitions than we have threads.

> I'll look into blockwise a bit. Perhaps updating Xarray's open_mfdataset to use it would yield some improvements.

The challenge is that blockwise currently expects to operate on Dask collections.  There isn't a clean way of using it to start up a new graph.

TomAugspurger:

PR #2944, opened 2019-08-09
Add GPUCurrentLoad dashboard plots

mrocklin:
A first step towards fixing https://github.com/rapidsai/dask-cuda/issues/36

### DIFF
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -1651,28 +1651,7 @@ class BokehScheduler(BokehServer):
 
         self.server_kwargs["prefix"] = prefix or None
 
-        self.apps = {
-            "/system": systemmonitor_doc,
-            "/stealing": stealing_doc,
-            "/workers": workers_doc,
-            "/events": events_doc,
-            "/counters": counters_doc,
-            "/tasks": tasks_doc,
-            "/status": status_doc,
-            "/profile": profile_doc,
-            "/profile-server": profile_server_doc,
-            "/graph": graph_doc,
-            "/individual-task-stream": individual_task_stream_doc,
-            "/individual-progress": individual_progress_doc,
-            "/individual-graph": individual_graph_doc,
-            "/individual-profile": individual_profile_doc,
-            "/individual-profile-server": individual_profile_server_doc,
-            "/individual-nbytes": individual_nbytes_doc,
-            "/individual-cpu": individual_cpu_doc,
-            "/individual-nprocessing": individual_nprocessing_doc,
-            "/individual-workers": individual_workers_doc,
-        }
-
+        self.apps = applications
         self.apps = {k: partial(v, scheduler, self.extra) for k, v in self.apps.items()}
 
         self.loop = io_loop or scheduler.loop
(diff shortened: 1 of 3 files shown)


pentschev:
A bit of an unrelated question, but I thought you wanted to keep GPU-specific code mostly out of dask/distributed. This is just for my own understanding, what's the current policy on whether GPU code should go in here rather than dask-cuda?

mrocklin:
It's a good question.  I could imagine moving this around later in the future.

There was some non-trivial amount of work to make the dashboarding general enough to take information like this.  I was also able to find a way to keep things nicely separate.  I also have decent confidence that this code won't need much experimentation in the future, and will be low churn.

mrocklin:
For example, we defined `cuda_serialize` in distributed because the code changes needed for it are mostly the same as the code changes needed for `dask_serialize`.  The infrastructure is there.  Things like the dask-cuda-worker code is changing rapidly enough that I'd prefer to keep it external for a while.  I could imagine moving it in once it's more stable.

pentschev:
Ok, makes sense. Thanks for the clarifications @mrocklin .

Now going back to this PR, would you like me to test it out and review it?

mrocklin:
Review would be welcome

distributed/dashboard/scheduler.py
```diff
@@ -1701,3 +1680,33 @@ def listen(self, *args, **kwargs):
...
+    "/individual-nprocessing": individual_nprocessing_doc,
+    "/individual-workers": individual_workers_doc,
+}
+
+try:
+    import pynvml  # noqa: 1708
+except ImportError:
+    pass
+else:
+    from . import nvml  # noqa: 1708
```
pentschev: I don't really understand, why are the imports needed? They don't seem to be used anywhere in this PR.

distributed/dashboard/tests/test_scheduler_bokeh.py
```diff
@@ -54,22 +55,8 @@ def test_simple(c, s, a, b):
...
-        "individual-task-stream",
-        "individual-progress",
-        "individual-graph",
-        "individual-nbytes",
-        "individual-nprocessing",
-        "individual-profile",
-    ]:
-        response = yield http_client.fetch("http://localhost:%d/%s" % (port, suffix))
+    for suffix in applications:
+        response = yield http_client.fetch("http://localhost:%d%s" % (port, suffix))
```
pentschev: This can be primarily a matter of personal preference, but wouldn't it make more sense to keep the format `%d/%s` and remove the `/` prefix from the strings in `applications`? My reasoning for that is that we can never append a string to the port without the `/`, so I think it makes this more error prone (if someone adds a new suffix and forgets the `/`).

pentschev:
It looks good to me, I added a couple of minor comments. I don't have experience with bokeh, superficially it looks fine for me, but maybe someone else with bokeh experience could take a look at that, otherwise, good to be merged by me.

distributed/dashboard/scheduler.py
mrocklin: They import the `nvml` module, which modifies the `applications` dictionary as a side effect.

distributed/dashboard/tests/test_scheduler_bokeh.py
mrocklin: Perhaps, but I wasn't strongly confident about why the leading slash was present in applications to begin with, and so I decided that that choice should be made in another PR.  

Paths in web servers are sometimes odd, especially when using prefixes.  I'd rather just make the one change to avoid confusion.

distributed/dashboard/scheduler.py
pentschev: I see, thanks for clarifying!

distributed/dashboard/tests/test_scheduler_bokeh.py
pentschev: Sounds good, thanks for the info.

pentschev:
LGTM, looking forward to see this being used by more people!

PR #2955, opened 2019-08-14
dashboard: allow server_kwargs to override defaults

### LINKED ISSUE #2915: [Feature] Let allow_websocket_origin to be overridden
When distributed creates the Bokeh server for the dashboard it passes `allow_websocket_origin=["*"]` ([here](https://github.com/dask/distributed/blob/cc4fc7d8fe42c41b328ade1db54dd88da904244a/distributed/dashboard/core.py#L36)).

That's not great from a security standpoint: any webpage on the internet can run Javascript to interact with your dashboard, exploit any security holes there might be in the installed version of Tornado, exfiltrate information about your cluster etc. Changing the default would break compatibility, but it would be nice if I could at least override it for my apps. Unfortunately, passing something through service_kwargs doesn't work, because then the argument is passed twice.

I'd like to suggest changing the way the constructor is called so that first a dictionary is constructed with the internal args, then update it with server_kwargs, then `**` that dict into the call. That way the user kwargs can override the other arguments. Apologies for not providing a PR (yet) - I'm about to go on holiday and wanted to file this before I forgot.

bmerry:
Closes #2915.

### DIFF
--- a/distributed/dashboard/core.py
+++ b/distributed/dashboard/core.py
@@ -28,16 +28,16 @@ class BokehServer(object):
             ip = None
         for i in range(5):
             try:
-                self.server = Server(
-                    self.apps,
+                server_kwargs = dict(
                     port=port,
                     address=ip,
                     check_unused_sessions_milliseconds=500,
                     allow_websocket_origin=["*"],
                     use_index=False,
                     extra_patterns=[(r"/", web.RedirectHandler, {"url": "/status"})],
-                    **self.server_kwargs
                 )
+                server_kwargs.update(self.server_kwargs)
+                self.server = Server(self.apps, **server_kwargs)
                 self.server.start()
 
                 handlers = [
--- a/distributed/dashboard/tests/test_scheduler_bokeh_html.py
+++ b/distributed/dashboard/tests/test_scheduler_bokeh_html.py
@@ -6,7 +6,8 @@ import pytest
 pytest.importorskip("bokeh")
 
 from tornado.escape import url_escape
-from tornado.httpclient import AsyncHTTPClient, HTTPClientError
+from tornado.httpclient import AsyncHTTPClient, HTTPClientError, HTTPRequest
+from tornado.websocket import websocket_connect
 
 from dask.sizeof import sizeof
 from distributed.utils import is_valid_xml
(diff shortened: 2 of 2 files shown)


bmerry:
@martindurant this is what I had in mind. I'm not sure what to do about tests, because from what I can see none of the existing tests in distributed/dashboard/tests are customising service_kwargs. Is it all being tested via the CLI tests? Should I wire up a CLI option for allow_websocket_origin to test it?

martindurant:
I don't know about the testing per se, but it does seem useful to allow these overrides on the CLI. It comes with the complexity of inferring types, though, so it may not be worth the hassle.

mrocklin:
> because from what I can see none of the existing tests in distributed/dashboard/tests are customising service_kwargs

Yeah... our testing of the dashboard isn't the greatest.  Honestly, most of the Dask developers don't have much experience testing web systems.

If you have thoughts on how to test this well that would certainly be welcome in order to ensure that this improvement doesn't get removed accidentally in the future.

> Should I wire up a CLI option for allow_websocket_origin to test it?

The CLI options are getting somewhat unweildy.  I would be more inclined to use configuration for less-frequently-used features like this if possible.

bmerry:
> Yeah... our testing of the dashboard isn't the greatest. Honestly, most of the Dask developers don't have much experience testing web systems.

I'm afraid I'm largely in the same boat.

In theory I don't think this should be *that* hard to test: pass an overridden allow_websocket_origin, then make an OPTIONS request to the websocket endpoint and check the Access-Control-Allow-Origin header. I'm not sure exactly where such a test belongs in the codebase though.

mrocklin:
> I'm not sure exactly where such a test belongs in the codebase though.

I would probably use distributed/dashboard/tests/test_scheduler_bokeh.py

Hopefully there are other tests in there that would serve as examples

mrocklin:
It's also not a huge deal if you can't test this.  It would be nice though.

bmerry:
> It's also not a huge deal if you can't test this. It would be nice though.

I'll give it a go, but if it turns into a pain I'll abandon the testing.

bmerry:
It turns out websocket security doesn't work the way I thought it did (Same Origin Policy and CORS do not apply, and the server has to enforce origin policies), but I've cooked up a test. It might be touch fragile because it has to encode the bokeh-protocol-version into the URL.

mrocklin:
> It might be touch fragile because it has to encode the bokeh-protocol-version into the URL.

Do you have a sense for how fragile this is or how often it is likely to change?  Perhaps a comment here helping the next person along would reduce concerns?

bmerry:
> Do you have a sense for how fragile this is or how often it is likely to change?

The fact that it's still protocol version 1.0 after however many years Bokeh has been around suggests that it's unlikely to change frequently.

mrocklin:
Sounds sensible.  OK, merging.  Thanks @bmerry !

PR #2957, opened 2019-08-16
it's safe to add A robust case or omit the error

abaelhe:
for example,   
  worker0( 'tls://10.252.19.48:63530' ):  pub = Pub(topic_name)
  worker1( 'tls://10.252.19.49:63530' ):  sub = Sub(topic_name, worker=w)
  
if :
    del worker0
when calling:
                   scheduler.extensions["pubsub"].add_subscriber( name=topic_name, ...  )
then:
                  pubsub.scheduler.worker_send('tls://10.252.19.48:23530', {"op": "pubsub-add-subscriber", "address": 'tls://10.252.19.49:63530', "name": topic_name})
   may receive A  "KeyError"              

so it's safe to omit the error

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2423,7 +2423,6 @@ class Scheduler(ServerNode):
         """
         if isinstance(plugin, type):
             plugin = plugin(self, **kwargs)
-tls://10.252.19.48:23530
         if idempotent and any(isinstance(p, type(plugin)) for p in self.plugins):
             return
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2423,7 +2423,7 @@ class Scheduler(ServerNode):
         """
         if isinstance(plugin, type):
             plugin = plugin(self, **kwargs)
-
+tls://10.252.19.48:23530
         if idempotent and any(isinstance(p, type(plugin)) for p in self.plugins):
             return
 
@@ -2443,6 +2443,8 @@ class Scheduler(ServerNode):
             self.stream_comms[worker].send(msg)
         except (CommClosedError, AttributeError):
             self.loop.add_callback(self.remove_worker, address=worker)
+        except KeyError:
+            pass
 
     ############################
     # Less common interactions #


mrocklin:
Thank you for the pull request @abaelhe 

Two requests:

1.  Can you remove this line: `tls://10.252.19.48:23530`
2.  Can you add a test?  In the description of this PR you have a good beginning I think.

abaelhe:
> Thank you for the pull request @abaelhe
> 
> Two requests:
> 
> 1. Can you remove this line: `tls://10.252.19.48:23530`



> 2. Can you add a test?  In the description of this PR you have a good beginning I think.



> Thank you for the pull request @abaelhe
> 
> Two requests:
> 
> 1. Can you remove this line: `tls://10.252.19.48:23530`
@mrocklin 
    Accepted. 

> 2. Can you add a test?  In the description of this PR you have a good beginning I think.
@mrocklin 
    To this commit, code review is good enough. Let's keep things simple and useful.
There are so many more important things to do other than test, especially at current early stage of our project.
Actually i've found and fixed many other more important  "dask/distributed"  bugs in my job time.
and maybe submit new commits.

abaelhe:
@mrocklin @jbcrail 
it's time to  approve and merge ?

abaelhe:
@mrocklin

mrocklin:
I'm sorry, I'm not willing to make a change this large without understanding what was going wrong before.  We use `send_to_worker` very frequently, and I have not heard of problems with KeyErrors before.  If there is a specific issue around Pub/Sub then I recommend making a reproducible example so that others can see the problem, then maybe we can make a more focused change.

abaelhe:
@mrocklin that is not Pub|Sub issue; BUT may caused by
  <SCHEDULER_INSTANCE>.remove_worker(...) before <SCHEDULER_INSTANCE>.worker_send(...) ;

i experienced this issue in my work cluster with many nodes for Tensorflow training and estimation with CUDA GPU Acceleration ;

since the cluster is in AsyncStream|Actor Computation model. and there are many users at the same time;  The true reason is in deep investigation. 

with above commit we can avoid the issue, and my training and estimation work is PASSED.

mrocklin:
If we are trying to send a message to a worker that does not exist, then that is a problem.  We should identify the root problem.

I recommend raising an issue with a minimal reproducible example and a traceback.  See https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports

I'm closing this for now.  Thank you @abaelhe for the PR.  We're going to choose not to merge this.  My apologies if this causes you some inconvenience.

PR #2963, opened 2019-08-17
Add support for separate external address for SpecCluster scheduler

### LINKED ISSUE #2962: SpecCluster Multiple IP addresses
There are situations when creating a distributed cluster where the workers will need to connect to the scheduler on an internal private IP address but the client should connect on an external public IP.

One example of this would be out-of-cluster kubernetes where the workers need to connect via the internal container networking IP but the client should connect through an external load balancer or node port. Another would be on AWS ECS where the scheduler is assigned both an internal and external address, sadly workers are not able to route to the external address and clients are not able to route to the internal address.

When creating the scheduler via the various APIs both addresses will be known, but currently the `SpecCluster` only takes a single address which is passed on to the worker tasks as well as used to connect the RPC.

I propose that we make it possible for an external address to optionally also be set, which will be used for the RPC and daskboard urls. The `scheduler_comm` is created [here](https://github.com/dask/distributed/blob/master/distributed/deploy/spec.py#L228) and we could change it to `self.scheduler.external_address or self.scheduler.address` with `self.scheduler.external_address` defaulting to `None`.

jacobtomlinson:
Minimal implementation which closes #2962.

Allow `SpecCluster` scheduler tasks to set an optional `external_address` property which will be used for the RPC connection instead of the scheduler address.

Allows clusters to be created and managed externally while allowing workers to continue to talk on an internal network.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -226,7 +226,7 @@ class SpecCluster(Cluster):
         self.status = "starting"
         self.scheduler = await self.scheduler
         self.scheduler_comm = rpc(
-            self.scheduler.external_address or self.scheduler.address,
+            getattr(self.scheduler, "external_address", None) or self.scheduler.address,
             connection_args=self.security.get_connection_args("client"),
         )
         await super()._start()
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -23,6 +23,7 @@ class ProcessInterface:
 
     def __init__(self):
         self.address = None
+        self.external_address = None
         self.lock = asyncio.Lock()
         self.status = "created"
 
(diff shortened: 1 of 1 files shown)


mrocklin:
If we do this then we might also want to override the `dashboard_link` property from the Cluster superclass.

https://github.com/dask/distributed/blob/912b6f79787b778ff4cfbde67b1378b69adc9540/distributed/deploy/cluster.py#L193-L201

mrocklin:
Ah, never mind about the `dashboard_link`, that uses `scheduler_address`, which uses `scheduler_comm.address`

mrocklin:
OK, I'm +1 on this.  If you're ok with things I'll merge on passed tests.

PR #2964, opened 2019-08-19
Improve get_ip_interface error message when interface does not exist

lesteve:
Context: in `dask-jobqueue` you need some tweaking to find the right network interface to use on your cluster. People may not be familiar with what a network interface is and the current error message is a bit cryptic.

The main thing I am wondering about is whether we should keep the same error type `KeyError` with backward-compatibility in mind or switch to `ValueError` since `ValueError` is already the type of the error you get when the network interface does not have a valid IPv4 address.

My personal preference would be to switch to `ValueError` but I have kept `KeyError` for now.

Error message on master:
```
In [1]: from distributed.utils import get_ip_interface 
   ...: get_ip_interface('asdf')                                                                                                                                                   
<ipython-input-1-cf8f3124444e> in <module>
      1 from distributed.utils import get_ip_interface
----> 2 get_ip_interface('asdf')

/home/local/lesteve/dev/distributed/distributed/utils.py in get_ip_interface(ifname)
    170     import psutil
    171 
--> 172     for info in psutil.net_if_addrs()[ifname]:
    173         if info.family == socket.AF_INET:
    174             return info.address

KeyError: 'asdf'
```

Error message in this PR:
```
In [1]: from distributed.utils import get_ip_interface 
   ...: get_ip_interface('asdf')                                                                                                                                                   
<ipython-input-1-cf8f3124444e> in <module>
      1 from distributed.utils import get_ip_interface
----> 2 get_ip_interface('asdf')

/home/local/lesteve/dev/distributed/distributed/utils.py in get_ip_interface(ifname)
    175         allowed_ifnames = list(net_if_addrs.keys())
    176         raise KeyError('{!r} is not a valid network interface. '
--> 177                          'Valid network interfaces are: {}'.format(ifname, allowed_ifnames))
    178 
    179     for info in net_if_addrs[ifname]:

KeyError: "'asdf' is not a valid network interface. Valid network interfaces are: ['lo', 'enp0s25', 'br-e1ab2121f792', 'docker0', 'br-9f15079dee0d']"
```

### DIFF
--- a/distributed/tests/test_utils.py
+++ b/distributed/tests/test_utils.py
@@ -149,8 +149,16 @@ def test_get_ip_interface():
         assert get_ip_interface("lo") == "127.0.0.1"
     else:
         pytest.skip("test needs to be enhanced for platform %r" % (sys.platform,))
-    with pytest.raises(KeyError):
-        get_ip_interface("__non-existent-interface")
+
+    non_existent_interface = "__non-existent-interface"
+    expected_error_message = ("{!r} is not a valid network interface.+"
+                              "Valid network interfaces are:".format(non_existent_interface))
+    if sys.platform == 'darwin':
+        expected_error_message += ".+'lo0'"
+    elif sys.platform.startswith("linux"):
+        expected_error_message += ".+'lo'"
+    with pytest.raises(KeyError, match=expected_error_message):
+        get_ip_interface(non_existent_interface)
 
 
 def test_truncate_exception():
--- a/distributed/tests/test_utils.py
+++ b/distributed/tests/test_utils.py
@@ -151,15 +151,13 @@ def test_get_ip_interface():
         pytest.skip("test needs to be enhanced for platform %r" % (sys.platform,))
 
     non_existent_interface = "__non-existent-interface"
-    expected_error_message = (
-        "{!r} is not a valid network interface.+"
-        "Valid network interfaces are:".format(non_existent_interface)
-    )
+    expected_error_message = "{!r}.+network interface.+".format(non_existent_interface)
+
     if sys.platform == "darwin":
-        expected_error_message += ".+'lo0'"
+        expected_error_message += "'lo0'"
     elif sys.platform.startswith("linux"):
-        expected_error_message += ".+'lo'"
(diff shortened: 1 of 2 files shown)


distributed/tests/test_utils.py
```diff
@@ -149,8 +149,18 @@ def test_get_ip_interface():
...
+    non_existent_interface = "__non-existent-interface"
+    expected_error_message = (
+        "{!r} is not a valid network interface.+"
+        "Valid network interfaces are:".format(non_existent_interface)
+    )
+    if sys.platform == "darwin":
+        expected_error_message += ".+'lo0'"
+    elif sys.platform.startswith("linux"):
+        expected_error_message += ".+'lo'"
+    with pytest.raises(KeyError, match=expected_error_message):
```
mrocklin: Small nit:

Rather than test that the entire error message is exactly this, we might instead test some properties of it that we care about, like that "network interface" and the names of other interfaces are present in the message.

If someone changes the error message but includes this kind of information I think that that's ok, and shouldn't be considered an error.

distributed/utils.py
```diff
@@ -169,7 +169,16 @@ def get_ip_interface(ifname):
...
 
-    for info in psutil.net_if_addrs()[ifname]:
+    net_if_addrs = psutil.net_if_addrs()
+
+    if ifname not in net_if_addrs:
+        allowed_ifnames = list(net_if_addrs.keys())
+        raise KeyError(
+            "{!r} is not a valid network interface. "
+            "Valid network interfaces are: {}".format(ifname, allowed_ifnames)
+        )
```
mrocklin: This is great.  I really like changes like this!

mrocklin:
Thanks for this change @lesteve !  I left one small nit, but I'm totally fine with this PR as-is.

I'm fine with changing the error type if you prefer.

lesteve:
I pushed a commit:
* the type of the error is now `ValueError`
* I reduced the test dependence on the specific wording of the error message. Now I am only testing that `ifname`, `"network interface"` and "lo" (or its OSX equivalent) is in the error message

mrocklin:
Thanks @lesteve !

PR #2970, opened 2019-08-21
With Facebook's  Zstandard - Fast real-time compression algorithm

abaelhe:
References:
    https://pypi.org/project/zstandard/
    https://github.com/facebook/zstd
    http://www.zstd.net


distributed/protocol/compression.py
```diff
@@ -93,6 +93,27 @@ def _fixed_lz4_decompress(data):
...
     default_compression = "lz4"
 
+
+with ignoring(ImportError):
+    import zstandard as zstd
+
+    def zstd_compress(data):
+        if isinstance(data, (memoryview, bytearray)):
+            data = bytes(data)
+        return zstd.ZstdCompressor().compress(data)
```
crepererum: I think you don't need to create the compressor and decompressor on every compress call.

Also, the compressor can probably get more threads (like 2, which is also done for other compressors).

Furthermore, it would be nice if the compression level would be controllable via a config variable.

distributed/protocol/compression.py
```diff
@@ -93,6 +93,27 @@ def _fixed_lz4_decompress(data):
...
+    def zstd_decompress(data):
+        if isinstance(data, (memoryview, bytearray)):
+            data = bytes(data)
+        return zstd.ZstdDecompressor().decompress(data)
+
+    compressions["zstd"] = {
+        "compress": zstd_compress,
+        "decompress": zstd_decompress,
+    }
+    default_compression = "zstd"
```
mrocklin: Hrm, should zstd or lz4 be the default?

distributed/protocol/compression.py
mrocklin: +1

mrocklin:
>  I think there should also be some tests.

Agreed, but fortunately we may get this for free.  This is already in the tests.

```python
@pytest.mark.parametrize("compression", list(compressions))
@pytest.mark.parametrize("serialize", [echo_serialize, echo_no_serialize])
def test_compression(compression, serialize, loop):
    with dask.config.set(compression=compression):

        @gen.coroutine
        def f():
            server = Server({"echo": serialize})
            server.listen("tcp://")

            with rpc(server.address) as r:
                data = b"1" * 1000000
                result = yield r.echo(x=to_serialize(data))
                assert result == {"result": data}

            server.stop()

        loop.run_sync(f)
```

distributed/protocol/compression.py
crepererum: that's a good question. The [squash benchmark](https://quixdb.github.io/squash-benchmark/) kinda indicates that zstd is better, but that of course depends on the payload.

mrocklin: My previous intuition was that lz4 ran at a few GB/s bandwidth and got modest compression ratios (2-4) while zstd ran a bit slower, maybe 1GB/s, but got better compression ratios (4-10).  Obviously those numbers are highly dependent.

mrocklin:
It looks like the `black` and `flake8` tests are failing.  

-  https://docs.dask.org/en/latest/develop.html#code-formatting
-  https://travis-ci.org/dask/distributed/jobs/574754596

distributed/protocol/compression.py
abaelhe: New commit: #2970
https://github.com/dask/distributed/pull/2970/commits/8de39d744dcd96c2ddb94f0288e83a4358d1c191
 customizable compression level and many other options;

> I think you don't need to create the compressor and decompressor on every compress call.
Accepted.


> Also, the compressor can probably get more threads (like 2, which is also done for other compressors).
 threads=0, which means disabled multi-threaded compression. so it's safe in default cases.
    and also user can change the "distributed.comm.zstd.threads" value in dask.yaml  according to their requirements.


> Furthermore, it would be nice if the compression level would be controllable via a config variable.
comprehensive options are configurable now. includes compression level , import policy, ...

crepererum:
> @mrocklin fortunately, set zstd as the default compression option, will make nearly all tests reusable.

I think the code is not tested at all since zstd is not installed on travis right now.

mrocklin:
Good point @crepererum 

@abaelhe can you add zstd to our CI?  The relevant directory is here: https://github.com/dask/distributed/tree/master/continuous_integration

I suspect that there are conda packages for zstd that we can use on conda-forge, so this should mean adding the appropriate line here:

https://github.com/dask/distributed/blob/799b5a0a0c95dc757a69e440e1d70a5e3afdd56f/continuous_integration/travis/install.sh#L30-L55

abaelhe:
Windows :  [#2987](https://github.com/dask/distributed/pull/2987/commits/52f0bc9eeac956e4972c9c511b17508c6236fc7d)
MacOSX + Linux: [#2988](https://github.com/dask/distributed/pull/2988/commits/06790ca388aa9d6606d6d758a2e79ad982f6f180)

mrocklin:
@abaelhe please include these changes in this PR.  Please do not create multiple PRs for this change.

You can push more git commits to your branch here.  The first pull request will update automatically.

Also, please use conda rather than pip.  Please see the lines of code I included in my previous comment.

distributed/protocol/compression.py
```diff
@@ -93,6 +93,41 @@ def _fixed_lz4_decompress(data):
     }
     default_compression = "lz4"
 
+with ignoring(ImportError):
+    _import_policy = dask.config.get("distributed.comm.zstd.import_policy", "default")
+    os.environ["PYTHON_ZSTANDARD_IMPORT_POLICY"] = _import_policy
```
mrocklin: Why this change?

abaelhe:
> @abaelhe please include these changes in this PR. Please do not create multiple PRs for this change.
> 
> You can push more git commits to your branch here. The first pull request will update automatically.
> 
> Also, please use conda rather than pip. Please see the lines of code I included in my previous comment.

@mrocklin @crepererum 

[Windows](https://github.com/dask/distributed/pull/2970/commits/e9eb9d33844f91cd1aac0032939c1b485308e24b)
[MacOSX+ Linux](https://github.com/dask/distributed/pull/2970/commits/85ca26d9ba931f738ad691cb1ef656b609cef15d):

distributed/protocol/compression.py
```diff
@@ -93,6 +93,41 @@ def _fixed_lz4_decompress(data):
...
+
+    zstd_level = dask.config.get("distributed.comm.zstd.level", None)
+    zstd_kwargs = dict(
+        write_checksum=dask.config.get("distributed.comm.zstd.write_checksum", None),
+        write_content_size=dask.config.get(
+            "distributed.comm.zstd.write_content_size", True
+        ),
+        write_dict_id=dask.config.get("distributed.comm.zstd.write_dict_id", True),
+        threads=dask.config.get("distributed.comm.zstd.threads", 0),
+    )
```
mrocklin: Can you add the default configuration values to this file?  https://github.com/dask/distributed/blob/master/distributed/distributed.yaml

Then you can remove these default values from the code here.

distributed/protocol/compression.py
abaelhe: according to:     [Zstandard](https://pypi.org/project/zstandard/) :
zstandard provided some module import policy: use "default", "cffi_fallback", "cext", or "cffi" ;

and as a public library implementation, we let user make their best choice.

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,52 @@
...
+                        )
+                        continue
+                except Exception as e:
+                    print(
+                        "FAILED, zstandard compression, config:%s, value:%s" % (cfg, v)
+                    )
+                    raise
+
+
+test_zstd()
```
crepererum: tests are automatically called by pytest, no need to call the test function here

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,52 @@
+def test_zstd():
+    import zstandard, zstandard as zstd, six, cffi, dask, distributed.protocol.compression as compr
```
crepererum: we probably want to skip this or the entire module if zstd is not installed

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,52 @@
...
+        # but AVAILABLE by manual install zstandard with "python setup.py install"
+        ("distributed.comm.zstd.import_policy", ["default", "cffi_fallback", "cext"]),
+        ("distributed.comm.zstd.level", range(1, 23)),
+        ("distributed.comm.zstd.write_checksum", [False, True]),
+        ("distributed.comm.zstd.write_content_size", [True]),
+        ("distributed.comm.zstd.write_dict_id", [False, True]),
+        ("distributed.comm.zstd.threads", range(-1, 100, 1)),
+    ]
+
+    for cfg, values in zstd_config_kwargs:
```
crepererum: I would normally prefer to use parametrized tests here, but I think that's up to you and @mrocklin if that's needed

distributed/tests/test_zstd.py
abaelhe: @crepererum  

commit 00703010e569a842415c25389fa514d2b2e95c06 (HEAD -> patch-5, origin/patch-5)
Author: heyijun <heyijun@qianxin.com>
Date:   Fri Aug 23 21:09:00 2019 +0800

     ONLY call test_zstd() if "zstandard" installed

distributed/tests/test_zstd.py
abaelhe: commit 00703010e569a842415c25389fa514d2b2e95c06 (HEAD -> patch-5, origin/patch-5)
Author: heyijun <heyijun@qianxin.com>
Date:   Fri Aug 23 21:09:00 2019 +0800

     ONLY call test_zstd() if "zstandard" installed

abaelhe:
@mrocklin  @crepererum  change the  package name from "zstd" to "zstandard":

according to my investigation on:  https://travis-ci.org/dask/distributed/jobs/575790895,

there are two  Zstandard Python implementation( both available at pypi.org and conda ):   

      First is "zstandard" and which was listed on [Facebook's Zstandard Official Site](http://facebook.github.io/zstd/):
            zstandard-0.11.1 is the latest one from pypi.orgi; 
            zstandard-0.8.1-py35_0 is the latest one from conda-forge/linux-64  

      Second is "zstd":  on page(line#514, line#560) https://travis-ci.org/dask/distributed/jobs/575671832
              zstd-1.4.1.0 is the latest one from pypi.org;
              zstd==1.3.7=h0b5b093_0 is latest according to  https://travis-ci.org/dask/distributed/jobs/575671832

"zstandard" MUST be the sophisticated choice to us.

distributed/tests/test_zstd.py
abaelhe: @mrocklin @crepererum 

To github.com:abaelhe/distributed.git
   00703010..6db2347e  patch-5 -> patch-5
(qianxin_venv) Abael:distributed abaelhe$ pytest  distributed/tests/test_zstd.py  
=============================================================== test session starts ===============================================================
platform darwin -- Python 3.7.4, pytest-5.1.1, py-1.8.0, pluggy-0.12.0 -- /www/workspace/qianxin/qianxin_venv/bin/python3.7
cachedir: .pytest_cache
rootdir: /www/workspace/qianxin/smart_crawler/distributed, inifile: setup.cfg
collected 6 items                                                                                                                                 

distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.import_policy-values0] PASSED                                               [ 16%]
distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.level-values1] PASSED                                                       [ 33%]
distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.write_checksum-values2] PASSED                                              [ 50%]
distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.write_content_size-values3] PASSED                                          [ 66%]
distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.write_dict_id-values4] PASSED                                               [ 83%]
distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.threads-values5] PASSED                                                     [100%]

============================================================ slowest 10 test durations ============================================================
0.22s call     distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.threads-values5]
0.06s call     distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.level-values1]
0.01s call     distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.import_policy-values0]
0.01s call     distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.write_dict_id-values4]
0.01s call     distributed/tests/test_zstd.py::test_zstd[distributed.comm.zstd.write_checksum-values2]

(0.00 durations hidden.  Use -vv to show these durations.)
================================================================ 6 passed in 0.32s ================================================================
(qianxin_venv) Abael:distributed abaelhe$

abaelhe:
> Good point @crepererum
> 
> @abaelhe can you add zstd to our CI? The relevant directory is here: 
https://github.com/dask/distributed/tree/master/continuous_integration
> 

BTW, the package name is "### zstandard",  and "### zstd" ONLY its abbreviation.
already done, and review requested.


> I suspect that there are conda packages for zstd that we can use on conda-forge, so this should mean adding the appropriate line here:
> 
> https://github.com/dask/distributed/blob/799b5a0a0c95dc757a69e440e1d70a5e3afdd56f/continuous_integration/travis/install.sh#L30-L55

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    for v in values:
+        with dask.config.set({cfg: v}):
+            assert dask.config.get(cfg) == v, "check config:%s, value:%s" % (cfg, v)
+            try:
+                six.moves.reload_module(cffi)
+                six.moves.reload_module(zstandard)
+                six.moves.reload_module(zstd)
+                six.moves.reload_module(compr)
```
mrocklin: Dask has dropped Python 2.  Do we need `six`?

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,51 @@
...
+            b"\x91\xe7\x9a\x84\xe5\x84\xbf\xe5\xad\x90\xef\xbc\x8c\xe6\x88\x91\xe6",
+            b"\xb7\xb1\xe6\x83\x85\xe5\x9c\xb0\xe7\x83\xad\xe7\x88\xb1\xe7\x9d\x80",
+            b"\xe6\x88\x91\xe7\x9a\x84\xe7\xa5\x96\xe5\x9b\xbd\xe5\x92\x8c\xe4\xba",
+            b"\xba\xe6\xb0\x91\xe3\x80\x82",
+        ]
+    )
+
+    for v in values:
+        with dask.config.set({cfg: v}):
+            assert dask.config.get(cfg) == v, "check config:%s, value:%s" % (cfg, v)
```
mrocklin: This assert statement checks that dask's config operations work correctly?  

I think that this is not necessary here.  We test dask.config in other code.

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,51 @@
...
+
+    original_data = b"".join(
+        [
+            b"\xe6\x88\x91\xe6\x98\xaf\xe4\xb8\xad\xe5\x9b\xbd\xe4\xba\xba\xe6\xb0",
+            b"\x91\xe7\x9a\x84\xe5\x84\xbf\xe5\xad\x90\xef\xbc\x8c\xe6\x88\x91\xe6",
+            b"\xb7\xb1\xe6\x83\x85\xe5\x9c\xb0\xe7\x83\xad\xe7\x88\xb1\xe7\x9d\x80",
+            b"\xe6\x88\x91\xe7\x9a\x84\xe7\xa5\x96\xe5\x9b\xbd\xe5\x92\x8c\xe4\xba",
+            b"\xba\xe6\xb0\x91\xe3\x80\x82",
+        ]
+    )
```
mrocklin: Why this data in particular?  Is it just random data?  If so, lets use the following:

```python
import numpy as np

original_data = np.random.random(10).data
```

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,51 @@
...
+                compress, decompress = (
+                    compr.compressions["zstd"]["compress"],
+                    compr.compressions["zstd"]["decompress"],
+                )
+                data = decompress(compress(original_data))
+                if data == original_data:
+                    print(
+                        "PASSED, Zstandard compression, config:%s, value:%s" % (cfg, v)
+                    )
+                    continue
```
mrocklin: Please do not use `print` here.  Instead, I recommend `assert`

```python
assert data == bytes(original_data)
```

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,51 @@
...
+            except Exception as e:
+                print("FAILED, zstandard compression, config:%s, value:%s" % (cfg, v))
+                raise
```
mrocklin: Please do not use `print` here.  Just let the exception raise.  We probably don't need try-except here.

distributed/tests/test_zstd.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+
+
+@pytest.mark.parametrize(
+    "cfg,values",
+    [
+        # 'cffi' is NOT support by "pip install zstandard"
+        # but AVAILABLE by manual install zstandard with "python setup.py install"
```
mrocklin: Is this option included with the conda installation?

distributed/protocol/compression.py
```diff
@@ -93,6 +93,39 @@ def _fixed_lz4_decompress(data):
...
+    zstd_decompressor = zstd.ZstdDecompressor()
+
+    def zstd_compress(data):
+        if isinstance(data, (memoryview, bytearray)):
+            data = bytes(data)
+        return zstd_compressor.compress(data)
+
+    def zstd_decompress(data):
+        if isinstance(data, (memoryview, bytearray)):
+            data = bytes(data)
```
mrocklin: I recommend either `distributed.utils.ensure_bytes` or just call `bytes(data)` in all cases (remove the `if`).  This will also support other types that support the buffer protocol.  It is also cheap.

```python
In [1]: data = b'0' * 1000000

In [2]: %timeit bytes(data)
91.4 ns ± 2.16 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

distributed/protocol/compression.py
mrocklin: The user can set this environment variable independently.  I am not so sure that it is a good idea to do this for them.

distributed/distributed.yaml
```diff
@@ -71,6 +71,15 @@ distributed:
     socket-backlog: 2048
     recent-messages-log-length: 0  # number of messages to keep for debugging
 
+    # Facebook's Zstandard - Fast real-time compression algorithm;  reference https://pypi.org/project/zstandard/
```
mrocklin: This is probably not necessary

```suggestion
```

distributed/tests/test_zstd.py
abaelhe: sure. six is not Python2;
"test_zstd.py" is fully verified by me with Python 3.7.4, on my MacBookPro installed zstandard;
use "six" is a best practice to me to keep write once and run on as many interpreter as possible. also keep consistent interface.

distributed/tests/test_zstd.py
abaelhe: it is for engineering effectiveness:
once exception raise, I can get the corresponding values of  "cfg" and "v",
so I can make right judgement and decision according to the printed info. at most time. 
then commit again;  
 *** ALL IN ONE PASS. ***

the printed info. also give us fine grain feedback regarding to the run progress.

and the print(...) is harmless.

distributed/tests/test_zstd.py
abaelhe: ditto

distributed/tests/test_zstd.py
mrocklin: I understand.  Still, please remove these print statements.  

These tests are run by every Dask developer and it is disruptive to have print statements for all tests.

distributed/tests/test_zstd.py
mrocklin: We only officially support the CPython interpreter with Python 3.  Please remove the use of the six module.

distributed/tests/test_zstd.py
abaelhe: it's a sample bytes that long enough.
random data is ok for short size.

distributed/distributed.yaml
abaelhe: it is a reminder comment for INTUITIVE DESIGN.
this comment let user know what is it,  let user make fast judgement by intuitive:
    whether they need to change the "zstd" config group.

keep it,  Harmless too.

abaelhe: < The user can set this environment variable independently. I am not so sure that it is a good idea to do this for them.

we provide user a UNIFIED config system in ONE "dask.yaml" as UNIFIED as possible.
just like Direct Flight better than Transfer Flight to user;

Actually dask|distributed is an accurate complex system with so many configs;
UNIFIED and INTUITIVE DESIGN is very useful.

distributed/protocol/compression.py
abaelhe: commit#47d05a5,   zstandard from Conda has "cffi" import policy

abaelhe:
(qianxin_venv) Abaelcom:distributed abaelhe$ git commit   -m 'ensure_bytes make safe input'
[patch-5 7f32df93] ensure_bytes make safe input
 1 file changed, 2 insertions(+), 2 deletions(-)

distributed/protocol/compression.py
mrocklin: cc @jcrist, thoughts on things like this?

distributed/tests/test_zstd.py
abaelhe: Python 3 is a MUST option;
Python 2 is a SHOULD option;
six make dask more user the easy way.

distributed/distributed.yaml
mrocklin: > we provide user a UNIFIED config system in ONE "dask.yaml" as UNIFIED as possible.

Unfortunately, Dask connects with many other software libraries.  We can not be responsible for all configuration systems.  We need to keep scope small for maintenance reasons.  If another library already has a configuration solution then we point users to that system.  Otherwise our configuration system becomes very very large.

mrocklin:
> 1. print -> assert CONDITION, CONTEXT_INFO;
> 2. six -> importlib;
> 3. raise without try...except...;
> 4. zstandard with "cffi" available and works in Conda;
> 5. removed double check on "dask.config" system;

Thank you for making these changes.

I know that you do not agree with all of them.  So I thank you even more :)

distributed/distributed.yaml
abaelhe: unified system makes user easy. 
scope small make maintenance easy;
I choose the First one. 

large configuration system is not reason. dask with distributed is large already.

we do need to rewrite the rules.
for user,  dask team.

abaelhe:
BTW, would you ask Anaconda team to upgrade the "zstandard" version in conda-forge channel to latest. or switch to use latest "zstandard" version from pypi.org.

mrocklin:
> BTW, would you ask Anaconda team to upgrade the "zstandard" version in conda-forge channel to latest. or switch to use latest "zstandard" version from pypi.org.

I do not work for Anaconda.  Anaconda does not maintain conda-forge.  You should raise an issue on conda-forge.

distributed/protocol/compression.py
jcrist: Yeah, the user should set this themselves if they care, distributed shouldn't modify environment variables like this.

distributed/protocol/compression.py
jcrist: The question is though should `zstd` be made the default over lz4. I think we'd want to benchmark a bit on realistic workloads before making this change for users. Not all users set the configuration system, the default should be the best option and I'm not yet convinced zstd is better for our workloads.

distributed/tests/test_zstd.py
jcrist: Instead of embedding a bytestring in our tests, we should just generate one as Matt said above. As a maintainer I see this fixed bytestring as written and think "why these bytes, I wonder why they matter", whereas seeing `os.urandom(100)` it's obvious that they're just a random byte payload.

abaelhe:
> > BTW, would you ask Anaconda team to upgrade the "zstandard" version in conda-forge channel to latest. or switch to use latest "zstandard" version from pypi.org.
> 
> I do not work for Anaconda. Anaconda does not maintain conda-forge. You should raise an issue on conda-forge.

so we can switch to use LATEST VERSION of Zstandard via "pip install --upgrade zstandard" ?

abaelhe:
@mrocklin latest commit and all check have passed.

distributed/distributed.yaml
```diff
@@ -71,6 +71,15 @@ distributed:
...
+    zstd:
+      import_policy: default    # one of "default", "cffi_fallback", "cext", "cffi"; use as zstandard module import policy.
```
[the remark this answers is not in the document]
crepererum: I agree with @mrocklin here. While the thread count and compression level configs make totally sense, I would also not monkey-patch the OS env variables, especially since the user has other ways of doing so. The proposed approach might not even work, since we are now relying on the fact the `distributed` is the first importer, which might be wrong in so many cases.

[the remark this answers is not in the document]
abaelhe: @crepererum   good point. :) 
@crepererum  @mrocklin
    now we use `importlib.reload(zstd)` to do reload the zstandard module to provide user unified design and consistent interaction.


To github.com:abaelhe/distributed.git
   bca1833c..775b2975  patch-5 -> patch-5

distributed/protocol/compression.py
abaelhe: @jcrist  @mrocklin  @crepererum @jbcrail 
vote !    zstd +1 

Dask Congress count the tickets.

zstd, lz4, snappy, ...   all were selected by people in a long time.
reference:   
    https://facebook.github.io/zstd/
    https://pypi.org/project/zstandard/
    .....

----------------------------------------------------------------
Zstandard is used by :

Featured

 Linux    FreeBSD   Redshift    Mercurial
Databases

 RocksDB    Hadoop    Cassandra    WiredTiger    Redis    Presto    Clickhouse    Groonga  Tarantool    TokuDB    HBase    PostgreSQL    VictoriaMetrics
File systems

 BtrFS    SquashFS    OpenZFS    ReiserFS
Web

 caddy    HHVM    nginx (module)    wget2    django
Serialization

 FST    Blosc    bcolz    mrcz    bgen    Gecko    HDF5    Sereal    PLink2    NAF
Network

 fbthrift    Fizz    proxygen    mcrouter    Rspamd    Tor
Hardware

 Inaccel
Games

 League of Legends    Godot    Khronos's KTX    OpenSiv3D    Esenthel    Mass Effect:Andromeda
Other

 Ubuntu    IBM TS7700    folly    cmake    TaskCluster   libarchive    tar    PowerAchiver  Fog    Backup    U++    Ceph   LiveScan3D

adamchainz:
I think you didn't mean to @ me

abaelhe:
> I do not work for Anaconda. Anaconda does not maintain conda-forge. You should raise an issue on conda-forge.

@mrocklin  is this your site, really makes me confused.: 
    [I work for Anaconda] (https://matthewrocklin.com)

mrocklin:
> @mrocklin is this your site ?( really makes me confused.):

Yup.  That information is outdated.  My apologies.

mrocklin:
@abaelhe  no one doubts that zstd is commonly used, or that it is new.  The question is is it likely to perform better than lz4 for workloads that are commonly seen for Dask.  Compression technologies have a wide range of trade-offs and some are good for some applications and some for others.  

Typically decisions like this are made by running a small benchmark with a few of the candidate algorithms on data that is representative of what is commonly used with the project, and the compare the relative cost of both compression and network transfer given the various options.  

To motivate your choice of zstd I recommend that you perform a small experiment and share the results here.  Perhaps you could download some dataset like the NYC Taxicab dataset, load it with `pandas.read_csv`, see how long it takes to compress each column, and the compression ratios you get on each column.  You could then compare the total time to compress and transfer the data in a couple of common network situations, perhaps assuming 100MB/s and 500MB/s network speeds.  I suspect that different compression tools will be optimal for each column, but that if ZStandard is more often a better choice then yes, it would be good to make it the default.

If you don't want to this that's fine.  You should raise an issue about making ZStandard the default, and remove the default change in this PR.

abaelhe:
@mrocklin   the default change is based on such an intelligent conclusion:
    to those user who know and installed "zstandard", that denote "zstandard" is their choice;
and vice versa; just like other embeded "snappy",  "lz4";
BTW, vote denote the opinion and contribution from most user. this is the reason.

jcrist:
@abaelhe without a benchmark of zstandard vs lz4 for operations meaningful to dask (as @mrocklin described above) the dask maintainers are unwilling to accept a change in our default compression algorithm. Further arguments without meaningful numbers to back them up will not change this decision. Please follow the request of @mrocklin above and either provide such a benchmark  or revert this change.

abaelhe:
@jcrist  @mrocklin 
 would you please provide us benchmarks of "snappy" and "lz4" ?
 since "zstandard", "snappy", "lz4", all the projects operating in full public, and all have been approved by many business companies and users.
 we ONLY do "DOUBLE CHECK" if necessary. for efficiency. and fair and justice to publics.

> @abaelhe without a benchmark of zstandard vs lz4 for operations meaningful to dask (as @mrocklin described above) ......

mrocklin:
I personally am unlikely to perform a benchmark in the near future.  

I laid out a suggestion for a benchmark in https://github.com/dask/distributed/pull/2970#issuecomment-526446131

If you don't have time to do this that's fine.  In that case I recommend that you leave the default as it is, and raise an issue for further discussion.

distributed/distributed.yaml
mrocklin: Please stop marking these as resolved.  Please remove these comments.

distributed/protocol/compression.py
mrocklin: Please remove these lines.

distributed/distributed.yaml
abaelhe: @mrocklin
Please keep these comments. 
Just like you keep the COPYRIGHT document from PSF,  in GPL, Apache, BSD, MIT;

we are NOT ONLY OPEN SOURCE;
BUT ALSO OPEN CULTURE AND BUILD BETTER GLOBAL COMMUNITY. 

To achieve that,  you do need to 
EMBRACE these GLOBAL OPEN CULTURE and CONCEPTS of VALUES are not identical to your's.

abaelhe:
please merge these commits to provide MORE OPTIONS and let user make their BEST CHOICE.

just like we provide beef, chicken, cake, salad, cola, vegetables ... and so on; user will choose the one they love.

if you don't provide benchmarks of "snappy" and "lz4", then just do it; 

since they all operating in full public; And i can take the further relevant things.

sublinus:
As indicated above there is no reason to make zstd the default compression for dask just because everybody else is using it. Dask users have very specific workloads, which might mean that

abaelhe:
@mrocklin @sublinus
attention,  zstd become the default compression for dask in reality requires that:
1. the "zstandard" is installed;
2. default compression set to "zstd". 

we set zstd as default and let user choose which one to be remain installed. 

"zstd", "snappy", "lz4", there are fair choices.

jcrist:
@abaelhe, I've pushed a commit to this branch making the changes we requested. Merging now.

Yes open-source is a community effort, but part of being a community effort means that the needs of multiple disparate users need to be met. We as Dask maintainers need to ensure that any changes don't negatively affect existing users (whose workloads may be different than yours). In the future, please *at least* listen to our requests - there are many constraints we have to keep in mind when reviewing PRs and requests for changes are not made lightly.

PR #2974, opened 2019-08-21
Add cores= and memory= keywords to scale

mrocklin:
Dask-Jobqueue did this internally.
It seems like a decent idea to pull upstream.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -334,27 +334,23 @@ class SpecCluster(Cluster):
     def scale(self, n=0, memory=None, cores=None):
         if memory is not None:
             try:
-                n = max(
-                    n,
-                    int(
-                        math.ceil(
-                            parse_bytes(memory)
-                            / parse_bytes(self.new_spec["options"]["memory_limit"])
-                        )
-                    ),
-                )
+                limit = self.new_spec["options"]["memory_limit"]
             except KeyError:
                 raise ValueError(
                     "to use scale(memory=...) your worker definition must include a memory_limit definition"
                 )
+            else:
+                n = max(n, int(math.ceil(parse_bytes(memory) / parse_bytes(limit))))
 
         if cores is not None:
             try:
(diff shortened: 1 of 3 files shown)


mrocklin:
cc @guillaumeeb for review (I think that you did this originally?)

distributed/deploy/spec.py
```diff
@@ -330,7 +331,31 @@ def __exit__(self, typ, value, traceback):
...
+                raise ValueError(
+                    "to use scale(memory=...) your worker definition must include a memory_limit definition"
+                )
+
+        if cores is not None:
+            try:
+                n = max(n, int(math.ceil(cores / self.new_spec["options"]["nthreads"])))
+            except KeyError:
+                raise ValueError(
+                    "to use scale(threads=...) your worker definition must include an nthreads= definition"
```
jrbourbeau: Minor comment

```suggestion
                    "to use scale(cores=...) your worker definition must include an nthreads= definition"
```

guillaumeeb:
Currently testing Dask-jobqueue SpecCluster PR. This does not work because dask-jobqueue uses the custom Job class, which does not need memory or nthreads argument, or at least not in every implementations. For PBS, `cluster.new_spec` looks like this :

```
{'cls': dask_jobqueue.pbs.PBSJob,
 'options': {'cores': 4,
  'memory': '20Gb',
  'walltime': '01:00:00',
  'interface': None,
  'protocol': 'tcp://',
  'security': None}}
```

Should we modify this in dask-jobqueue's side?

We also lack the nice widget for filling cores or memory in jupyter, see #2985

mrocklin:
I believe I have resolved this latest comment.  We now check for a variety of terms, like cores, threads, nthreads, and so on

mrocklin:
Although the widget issue is still live.  Thank you for raising an issue.

guillaumeeb:
Will try to test again all this next week, but not sure I can find the time. 

Thanks a lot for all the work!

PR #2975, opened 2019-08-21
Directly import progress from diagnostics.progressbar

### LINKED ISSUE #2973: dask.distributed.progress no longer callable in 2.3.0?
We've used the progress() function from dask.distributed a bunch in the past to display a progress bar in JupyterLab, but it seems to have stopped working after upgrading to Dask 2.3.0:

```
from dask.distributed import Client, progress
import dask.dataframe as dd

df = dd.demo.make_timeseries('2010', '2016',
                             {'value': float, 'name': str, 'id': int},
                             freq='10s', partition_freq='7d', seed=1)

df = df.persist()
progress(df)
```

Executing this in a single cell in JupyterLab (with an existing Dask cluster already running) results in:

```
<ipython-input-1-16af814d7204> in <module>
      7 
      8 df = df.persist()
----> 9 progress(df)

TypeError: 'module' object is not callable
```

Let me know if I can provide any more info. Thanks!

mrocklin:
For some reason the implicit import behavior has changed.

Fixes #2973

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -3,7 +3,7 @@ from dask.config import config
 from .actor import Actor, ActorFuture
 from .core import connect, rpc
 from .deploy import LocalCluster, Adaptive, SpecCluster
-from .diagnostics import progress
+from .diagnostics.progressbar import progress
 from .client import (
     Client,
     Executor,


jrbourbeau:
FWIW, I tracked down `a55515569d4c5da734e5b14ae414cd342c37ed7b` as the commit where this import behavior changed. It wasn't immediately clear how the changes in #2932 caused the `import` to change, and I'm not sure how much value there is in tracking it down. 

The changes here make the `import` more explicit, which I'm generally in favor of

PR #2984, opened 2019-08-22
Make workers robust to bad custom metrics

mrocklin:
Fixes https://github.com/rapidsai/dask-cuda/issues/122

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1548,3 +1548,13 @@ async def test_gpu_metrics(s, a, b):
 
     assert "gpu" in a.startup_information
     assert len(s.workers[a.address].extra["gpu"]["name"]) == count
+
+
+@pytest.mark.asyncio
+async def test_bad_metrics(cleanup):
+    def bad_metric(w):
+        raise Exception("Hello")
+
+    async with Scheduler() as s:
+        async with Worker(s.address, metrics={"bad": bad_metric}) as w:
+            assert "bad" not in s.workers[w.address].metrics
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -740,7 +740,7 @@ class Worker(ServerNode):
                 if hasattr(result, "__await__"):
                     result = await result
                 custom[k] = result
-            except Exception:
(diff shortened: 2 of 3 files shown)


distributed/tests/test_worker.py
```diff
@@ -1548,3 +1548,13 @@ def test_resource_limit():
 
     assert "gpu" in a.startup_information
     assert len(s.workers[a.address].extra["gpu"]["name"]) == count
+
+
+@pytest.mark.asyncio
+async def test_bad_metrics(cleanup):
+    def bad_metric(w):
+        raise Exception("Hello")
```
jacobtomlinson: This line made me smile 😄

distributed/worker.py
```diff
@@ -735,10 +735,13 @@ def local_dir(self):
...
-            result = metric(self)
-            if hasattr(result, "__await__"):
-                result = await result
-            custom[k] = result
+            try:
+                result = metric(self)
+                if hasattr(result, "__await__"):
+                    result = await result
+                custom[k] = result
+            except Exception:
```
rlratzel: Can we catch only the exception(s) raised in the situation where it's safe to ignore them? This catch-all handler will catch syntax errors and broken things that should work, making for a potential debugging mystery.

mrocklin: Good point.   Unfortunately the `metrics` provided here can be user provided, so we don't have a strong sense of what exactly we can catch and not catch.

Normally I would handle this by logging the error properly, so that we at least report up what's wrong while we don't error.  However, this gets called frequently, and I would like to avoid dumping a ton of logging messages.  This could be resolved by having some sort of log-once check or state, but at the moment I'm a bit pressed and didn't want to build this.

Suggestions welcome.  Otherwise I'm going to merge and release shortly (there are a couple of other things that I'd like to get out there as well).

rlratzel: Okay, that's reasonable IMO.  In lieu of your logging solution, maybe a FIXME describing that would be appropriate, possibly even with a link to your response here.

jacobtomlinson:
I can confirm this resolves the issue for me.

PR #2990, opened 2019-08-23
Remove lost workers from SpecCluster.workers

### LINKED ISSUE #2968: Track killed workers in SpecCluster
I've noticed that workers created by a `SpecCluster` remain in the worker count if they are killed harshly. The client shows them as being gone.

Perhaps we should add some checks in correct state to remove any dead workers and call it intermittently?

guillaumeeb:
Closes #2968. First attempt...

Still in work in progress as I'm experiencing some unexpected difficulties. I guess this must be some kind of race conditions with async methods, The test I added is failing in a way I don't understand.

I also encountered some issues when testing with https://github.com/dask/dask-jobqueue/pull/307, which is that `cluster.scheduler_info["workers"]` objects have String `id` and `name` attribute, whereas in `cluster.workers` and `cluster.worker_spec` these are integer. But this is not always the case with other Speccluster classes.

Another thing: I assumed we must only remove the killed worker from `cluster.workers`, and not from `cluster.worker_spec`. This way, the user could always `await cluster` or call `cluster._correct_state()`to restart lost workers. Not sure if this is the best thing to do...

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -8,6 +8,7 @@ import warnings
 
 import click
 import dask
+from dask.utils import ignoring
 from distributed import Nanny, Worker
 from distributed.security import Security
 from distributed.cli.utils import check_python_3, install_signal_handlers
@@ -354,6 +355,9 @@ def main(
             "dask-worker SCHEDULER_ADDRESS:8786"
         )
 
+    with ignoring(TypeError, ValueError):
+        name = int(name)
+
     nannies = [
         t(
             scheduler,
@@ -367,7 +371,7 @@ def main(
             port=port,
             dashboard_address=dashboard_address if dashboard else None,
             service_kwargs={"dashboard": {"prefix": dashboard_prefix}},
-            name=name if nprocs == 1 or not name else name + "-" + str(i),
+            name=name if nprocs == 1 or not name else str(name) + "-" + str(i),
             **kwargs
         )
         for i in range(nprocs)
--- a/distributed/cli/tests/test_dask_worker.py
+++ b/distributed/cli/tests/test_dask_worker.py
@@ -1,3 +1,4 @@
+import asyncio
 import pytest
 from click.testing import CliRunner
 
@@ -9,11 +10,11 @@ import os
 from time import sleep
 
 import distributed.cli.dask_worker
-from distributed import Client
+from distributed import Client, Scheduler
 from distributed.metrics import time
 from distributed.utils import sync, tmpfile
 from distributed.utils_test import popen, terminate_process, wait_for_port
-from distributed.utils_test import loop  # noqa: F401
+from distributed.utils_test import loop, cleanup  # noqa: F401
 
 
 def test_nanny_worker_ports(loop):
@@ -330,3 +331,13 @@ def test_bokeh_deprecation():
         except ValueError:
             # didn't pass scheduler
             pass
+
+
+@pytest.mark.asyncio
+async def test_integer_names(cleanup):
+    async with Scheduler(port=0) as s:
+        with popen(["dask-worker", s.address, "--name", "123"]) as worker:
+            while not s.workers:
+                await asyncio.sleep(0.01)
+            [ws] = s.workers.values()
+            assert ws.name == 123
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -107,10 +107,9 @@ class Cluster(object):
             self.scheduler_info["workers"].update(workers)
             self.scheduler_info.update(msg)
         elif op == "remove":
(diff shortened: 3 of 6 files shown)


mrocklin:
>  which is that cluster.scheduler_info["workers"] objects have String id and name attribute, whereas in cluster.workers and cluster.worker_spec these are integer

I suspect that this is because we pass the integer name into the CLI, like `--name 123`, and so it gets interpreted as a string.  One option would be to consider converting names to integers if possible.  We might do this either in `dask-worker` or in `Worker.__init__`.

```python
try:
    name = int(name)
except ValueError:
    pass
```

mrocklin:
> I assumed we must only remove the killed worker from cluster.workers, and not from cluster.worker_spec. This way, the user could always await cluster or call cluster._correct_state()to restart lost workers. Not sure if this is the best thing to do...

I also don't know what the right behavior is here.  We don't have much experience with this.  Any thoughts?

mrocklin:
> Still in work in progress as I'm experiencing some unexpected difficulties. I guess this must be some kind of race conditions with async methods, The test I added is failing in a way I don't understand.

Indeed.   here is the location of the failure:

```
            pre = list(set(self.workers))
            to_close = set(self.workers) - set(self.worker_spec)
            if to_close:
                if self.scheduler.status == "running":
                    await self.scheduler_comm.retire_workers(workers=list(to_close))
                tasks = [self.workers[w].close() for w in to_close]
                await asyncio.wait(tasks)
                for task in tasks:  # for tornado gen.coroutine support
                    with ignoring(RuntimeError):
                        await task
            for name in to_close:
>               del self.workers[name]
E               KeyError: 0
```

so we define to_close at the beginning, then there are a few `await` statements, during which anything can happen, then we use `to_close` afterwards, assuming that it is still valid.  Probably we should redefine it afterwards.

```diff
+           to_close = set(self.workers) - set(self.worker_spec)
            for name in to_close:
               del self.workers[name]
```

mrocklin:
Although one might also question why this is now different

guillaumeeb:
That's different because in the PR I modify `self.workers` when the Scheduler detects its end. This is causing a problem upon test teardown when the cluster is closed.

See https://github.com/dask/distributed/pull/2990/files#diff-e95e1ccb9336311eae957db482e0be4dR290. 

> 
> > I assumed we must only remove the killed worker from cluster.workers, and not from cluster.worker_spec. This way, the user could always await cluster or call cluster._correct_state() to restart lost workers. Not sure if this is the best thing to do...
> 
> I also don't know what the right behavior is here. We don't have much experience with this. Any thoughts?

I propose to leave it like that, so that the user knows that his cluster is not in his desired state anymore.

guillaumeeb:
> I suspect that this is because we pass the integer name into the CLI, like `--name 123`, and so it gets interpreted as a string. One option would be to consider converting names to integers if possible. We might do this either in `dask-worker` or in `Worker.__init__`.
> 
> ```python
> try:
>     name = int(name)
> except ValueError:
>     pass
> ```

It feels wrong to me to modify Worker class to fix this... Maybe we should do something in SpecCluster or Jobqueue instead?

mrocklin:
> It feels wrong to me to modify Worker class to fix this... Maybe we should do something in SpecCluster or Jobqueue instead?

I agree that SpecCluster is less central than Worker, but I also think that this feature is reasonable.  I've had to work around this problem before and have considered implementing this feature.  I think that it's *probably* a good idea, but I'm not sure.  We could also implement it only in `dask-worker`, if that is more comfortable (because it is less central).

guillaumeeb:
So for Dask-jobqueue, we will have some problem if we launch jobs with several processes: we'll have worker ids that looks like "1-0", "1-1", etc... In this case, we'll need to handled Strings, but that may be an edge case.

Could we think of using String as keys for SpecCluster.workers and SpecCluster.worker_spec?

mrocklin:
Hrm, so one problem that occurs in tests is that when we restart a cluster with `Nanny`s we actually don't want to remove the worker object in `cluster.workers`.  The Nanny itself is still alive and well, and it expects to ressurect the worker soon.  In this case it is not appropriate to remove the worker from the Cluster object, even though it has been temporarily removed from the scheduler.

mrocklin:
OK, so now we remove the worker if it has not reappeared after 15s (configurable).  I think that this is likely to work well both for local Nannys and for HPC/Kubernetes jobs.

mrocklin:
@guillaumeeb if you have a moment to review the changes here, that would be welcome

mrocklin:
> This is much more complex to handle that I could have done, but this all makes sense and looks good to me!

Yes, it became more complex than I expected.  It was still good, I think, to have you involved in the PR.  Hopefully it helps to spread around some knowledge of how things work internally.

PR #2991, opened 2019-08-23
Skip exceptions in startup information

jacobtomlinson:
Turns out #2984 doesn't completely solve rapidsai/dask-cuda#122. Not sure why my local testing was working yesterday as today it doesn't.

We also need to skip exceptions when gathering startup information as that may also use pynvml.


mrocklin:
Thanks for this, can I ask for a tiny test, similar to this

https://github.com/dask/distributed/blob/4e9d5ecb373f03c4e71928c79cbabdfb3f257d26/distributed/tests/test_worker.py#L1553-L1560

You should be able to provide a `startup_information=[bad_function]` keyword to the worker.

I can also handle this is you're out

jacobtomlinson:
Added :)

Issue #2992, opened 2019-08-23
GPU metrics are raising an exception


jacobtomlinson:
Despite [being initialized when imported](https://github.com/dask/distributed/blob/4e9d5ecb373f03c4e71928c79cbabdfb3f257d26/distributed/diagnostics/nvml.py#L3) the nvml startup info and metrics methods seem to raise an exception about NVML not being initialised. (rapidsai/dask-cuda#122)

```python-traceback
distributed.nanny - ERROR - Failed to start worker
Traceback (most recent call last):
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/nanny.py", line 674, in run
    await worker
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/worker.py", line 1016, in start
    await self._register_with_scheduler()
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/worker.py", line 811, in _register_with_scheduler
    metrics=await self.get_metrics(),
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/worker.py", line 740, in get_metrics
    result = await result
  File "/home/nfs/bzaitlen/miniconda3/envs/cudf-dev/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/worker.py", line 3406, in gpu_metric
    result = yield offload(nvml.real_time)
  File "/home/nfs/bzaitlen/miniconda3/envs/cudf-dev/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/home/nfs/bzaitlen/miniconda3/envs/cudf-dev/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/utils.py", line 1489, in offload
    return (yield _offload_executor.submit(fn, *args, **kwargs))
  File "/home/nfs/bzaitlen/miniconda3/envs/cudf-dev/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/home/nfs/bzaitlen/miniconda3/envs/cudf-dev/lib/python3.7/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
[...]

PR #2993, opened 2019-08-23
Pynvml initialization fix

rjzamora:
Addressing [#2992](https://github.com/dask/distributed/issues/2992) by explicitly checking if pynvml needs initialization (and doing so if needed) inside `real_time` and  `one_time`.

After these changes, I get the following behavior:

```
docker run --runtime=nvidia --rm -it -p ... rapidsai/rapidsai-nightly:cuda9.2-runtime-ubuntu16.04
```

Then, after rebuilding distributed with `rjzamora:pynvml-fix` - **In Jupyter notebook**:

```python
from dask_cuda import LocalCUDACluster
cluster = LocalCUDACluster()
cluster.scheduler.workers[cluster.scheduler.workers.keys()[0]].metrics
```
**Output**:
```
{'gpu': {'utilization': (0, 0, 0, 0, 0, 0, 0, 0),
  'memory-used': (3447259136,
   23938531328,
   10172825600,
   10697113600,
   22126592000,
   439943168,
   439943168,
   439943168)},
 'cpu': 0.0,
 'memory': 110350336,
 'time': 1566581704.5954611,
 'read_bytes': 0.0,
 'write_bytes': 0.0,
 'num_fds': 24,
 'executing': 0,
 'in_memory': 0,
 'ready': 0,
 'in_flight': 0,
 'bandwidth': 100000000}
```
Before these changes, the `'gpu'` infromation was missing from the `metrics` dictionary.

Please let me know if we should add a new test (to assert that the `metrics` include `'gpu'`) for systems with pynvml installed.

### DIFF
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -1,21 +1,19 @@
 import pynvml
 
-pynvml_initialized = False
 handles = None
 
 
-def _ensure_pynvml_initialized():
-    global pynvml_initialized, handles
-    if not pynvml_initialized:
+def _pynvml_handles():
+    global handles
+    if handles is None:
         pynvml.nvmlInit()
         count = pynvml.nvmlDeviceGetCount()
         handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
-        pynvml_initialized = True
+    return handles
 
 
 def real_time():
-    global handles
-    _ensure_pynvml_initialized()
+    handles = _pynvml_handles()
     return {
         "utilization": [pynvml.nvmlDeviceGetUtilizationRates(h).gpu for h in handles],
         "memory-used": [pynvml.nvmlDeviceGetMemoryInfo(h).used for h in handles],
@@ -23,8 +21,7 @@ def real_time():
 
 
 def one_time():
-    global handles
-    _ensure_pynvml_initialized()
+    handles = _pynvml_handles()
     return {
         "memory-total": [pynvml.nvmlDeviceGetMemoryInfo(h).total for h in handles],
         "name": [pynvml.nvmlDeviceGetName(h).decode() for h in handles],
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -3,6 +3,7 @@ import pynvml
 need_pynvml_init = True
 handles = None
 
+
 def _initialize_pynvml():
     global need_pynvml_init, handles
     pynvml.nvmlInit()
@@ -29,4 +30,3 @@ def one_time():
         "memory-total": [pynvml.nvmlDeviceGetMemoryInfo(h).total for h in handles],
         "name": [pynvml.nvmlDeviceGetName(h).decode() for h in handles],
     }
-
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -1,12 +1,20 @@
 import pynvml
 
-pynvml.nvmlInit()
-count = pynvml.nvmlDeviceGetCount()
+need_pynvml_init = True
+handles = None
 
-handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
+def _initialize_pynvml():
+    global need_pynvml_init, handles
+    pynvml.nvmlInit()
+    count = pynvml.nvmlDeviceGetCount()
+    handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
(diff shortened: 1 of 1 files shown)


distributed/diagnostics/nvml.py
```diff
@@ -1,19 +1,31 @@
...
+def _initialize_pynvml():
+    global need_pynvml_init, handles
+    pynvml.nvmlInit()
+    count = pynvml.nvmlDeviceGetCount()
+    handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
+    need_pynvml_init = False
 
 
 def real_time():
+    global need_pynvml_init, handles
```
rlratzel: Can you explain why this is needed/works? It seems like the old code should handle the init on import.

distributed/diagnostics/nvml.py
```diff
@@ -1,19 +1,31 @@
...
+    if need_pynvml_init:
+        _initialize_pynvml()
```
mrocklin: I wonder if we might include the check inside the `_initialize_pynvml` call.  This function would then just become idempotent.  We would call it all the time, and it would decide whether or not it needed to call the `pynvml` function.

rlratzel: yes, and if that's done, I'd prefer to change the name to `_ensure_pynvml_initialized()` to indicate that it may not (re)run initialization.

distributed/diagnostics/nvml.py
rjzamora: @rlratzel - Sorry I missed your code snippet in the slack thread (long thread!).

The short answer is "I dont know yet"...  However,  it may have something to do with multiple workers fighting over a "load" lock for the `nvml` library... I should probably understand pynvm well enought to answer this, so I'll dig a bit deeper :)

distributed/diagnostics/nvml.py
```diff
@@ -1,19 +1,30 @@
...
+    if not pynvml_initialized:
+        pynvml.nvmlInit()
+        count = pynvml.nvmlDeviceGetCount()
+        handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
+        pynvml_initialized = True
 
 
 def real_time():
+    global handles
+    _ensure_pynvml_initialized()
```
mrocklin: I recommend that we remove the global here, and have the function return handles.  This again should hide some of the jankiness of the global state inside of the function.

Maybe

```python
handles = pynvml_handles()  # both inits if necessary, and returns handles
```

rlratzel: I don't want to hold up getting this merged, but here's another suggestion (untested)
```
class PynvmlMgr:
    def __init__(self):
        self.__initialized = False
        self.__handles = []

    @property
    def handles(self):
        self.__ensureInitialized()
        return self.__handles

    def __ensureInitialized(self):
        if not self.__initialized:
            pynvml.nvmlInit()
            count = pynvml.nvmlDeviceGetCount()
            self.__handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
            self.__initialized = True

pynvmlMgr = PynvmlMgr()

# Then just use handles:
pynvmlMgr.handles
```

rjzamora: Makes sense to me - made the change.

mrocklin:
Looks good to me.  Merging on passing CI

rjzamora:
@mrocklin - I can also use the [@rlratzel's suggestion](https://github.com/dask/distributed/pull/2993#discussion_r317265752). For example, this also seems to work:

```python
import pynvml


class PynvmlManager:
    def __init__(self):
        self._initialized = False
        self._handles = ()

    @property
    def handles(self):
        self._ensure_initialized()
        return self._handles

    def _ensure_initialized(self):
        if not self._initialized:
            pynvml.nvmlInit()
            count = pynvml.nvmlDeviceGetCount()
            self._handles = [pynvml.nvmlDeviceGetHandleByIndex(i) for i in range(count)]
            self._initialized = True

manager = PynvmlManager()


def real_time():
    return {
        "utilization": [pynvml.nvmlDeviceGetUtilizationRates(h).gpu for h in manager.handles],
        "memory-used": [pynvml.nvmlDeviceGetMemoryInfo(h).used for h in manager.handles],
    }


def one_time():
    return {
        "memory-total": [pynvml.nvmlDeviceGetMemoryInfo(h).total for h in manager.handles],
        "name": [pynvml.nvmlDeviceGetName(h).decode() for h in manager.handles],
    }
```

mrocklin:
I have a slight preference for what you have here if that's ok with you

Issue #2837, opened 2019-07-12
[BUG] Task state with CuPy array


matthieubulte:
Hi!

This issue has started as an issue in the `dask_cuda` project, but it seems to be more adequate here. The original issue can be found here: https://github.com/rapidsai/dask-cuda/issues/93

I am currently working on implementing some statistical estimators using `dask`, `distributed`, `cupy` and `dask_cuda` for multi-GPU support.

My issue all started with the observation I am doing in the referred issue: the tasks screen of the dashboard indicates very high and unintended `transfer-*`costs when performing some simple operations on the GPU. Please see the initial issue to have a screenshot of the task screen.

After some exploration of the execution via some custom `SchedulerPlugin` (this is a great feature!), I realized that these very high transfer cost (>500ms) didn't make sense. They were associated to `gather` actions on tasks gathering chunks smaller than 1kb. See:
![image](https://user-images.githubusercontent.com/6624306/61159113-400f4580-a4fb-11e9-9b11-7c19307c6f33.png)

After some more debugging, I realized that the reason for that is that `cupy` operations are asynchronous. You can find an explanation and example in the original issue. However, since `distributed` is not aware of this, the library is mislead in measuring very low `compute-*` times and very high `transfer-*` time since the serialization requires a synchronous copying from the GPU and the CPU, which blocks until the computation is over and the data is copied.

I know that `distributed` and `cupy` are two very separate libraries, but this clearly seems to be a bug, or at least a misleading situation for people trying to combine `cupy` and `distributed`.

So, I would be very happy to work on a fix for this, but I'm not sure if you would be interested in adding a custom support for `cupy`, or if you would prefer to have (or already have) some generic solution for handling such situations.

NOTE: this is somewhat separated, but the attached screenshot was constructed via a mix of the graph computed by dask and the execution trace obtained with a `SchedulerPlugin`. Would a cleaner / nicer version of this be interesting to a more general audience?

pentschev:
> So, I would be very happy to work on a fix for this (on the cupy side, it would be something like adding `cupy_array.device.synchronize()`, see https://docs-cupy.chainer.org/en/stable/reference/generated/cupy.cuda.Device.html#cupy.cuda.Device.synchronize ), but I'm not sure if you would be interested in adding a custom support for `cupy`

Should a specialization like this be done, (most likely) the correct place for that would be https://github.com/rapidsai/dask-cuda.

> or if you would prefer to have (or already have) some generic solution for handling such situations.

[...]

PR #2994, opened 2019-08-24
Extend Worker plugin API (task_started, task_finished, task_failed) [#2837]

matthieubulte:
As discussed in #2837 , here is the extension to the Worker plugin API with a `task_finished` callback, called whenever a task is done executing.

Note that because a task might run in very different conditions, `task_finished` should not be able to mutate the worker or plugin itself.

Let me know if you would like me to change anything.


crepererum:
I think that is already half of dask/dask#4789, which is cool. The `task_started` hooks is missing for that story but I can do that as well. Should be easy now that you already did the finished-tracking.

matthieubulte:
@crepererum: that's great! I can also do the `task_started`, it's only a couple of copy/pastes away. I would be interested on feedback on the implementation, it feels quite hacky right now and I don't know if this is something the maintainers would be comfortable integrating in the current state. Any idea how/if to improve it?

distributed/worker.py
```diff
@@ -3228,12 +3237,19 @@ def apply_function_actor(
...
     with active_threads_lock:
         del active_threads[ident]
 
     return result
 
 
+def notify_task_finished(plugins, result):
+    for plugin in plugins.values():
+        if hasattr(plugin, "task_finished"):
+            plugin.task_finished(result)
```
crepererum: it might also be helpful to have a state here (i.e. "finished successfully" vs "exception")

distributed/worker.py
```diff
@@ -3226,14 +3238,28 @@ def apply_function_actor(
...
+def notify_task_started(plugins, worker, args, kwargs):
+    for plugin in plugins.values():
+        if hasattr(plugin, "task_started"):
+            plugin.task_started(worker, args, kwargs)
+
+
+def notify_task_finished(plugins, worker, result):
+    for plugin in plugins.values():
+        if hasattr(plugin, "task_finished"):
+            plugin.task_finished(worker, result)
```
crepererum: might be helpful to have the result state (successfully vs exception) reported to the plugin as well

distributed/worker.py
```diff
@@ -2252,6 +2252,8 @@ def actor_execute(self, comm=None, actor=None, function=None, args=(), kwargs={}
 
         if iscoroutinefunction(func):
             result = yield func(*args, **kwargs)
+            notify_task_finished(self.plugins, result)
+
         elif separate_thread:
             result = yield self.executor_submit(
```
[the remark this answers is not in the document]
matthieubulte: This was a mistake, thanks. I still have to look into how testing all possible ways to call a task to make sure we don't miss any of those.

distributed/worker.py
matthieubulte: That's a good idea, but two points:

1. it would basically mean adding try/catches at every call site, which would be quite invasive.

2. maybe it could be also be done by adding a general exception handling hook to the worker plugin.

This also makes me think that we also might want to add the task key to each of these function calls. Otherwise it would make it hard to connect the dots between different logs.

crepererum: > That's a good idea, but two points ...

Then let's clarify the behavior: if a task crashes, what do we expect to happen? Does the worker plugin get:

1. a `task_done` notification and an (TBD) exception/crashed notification
2. only an (TBD) exception/crashed notification
3. only a `done` notification

> This also makes me think that we also might want to add the task key to each of these function calls.

IMHO that would be extremely helpful.

matthieubulte: > Then let's clarify the behavior: if a task crashes, what do we expect to happen?

That was my question as well ;) It's really just a question of API design, and even for the user it really doesn't make much of a difference I would say. I'll implement 3. and then we can discuss further.

> IMHO that would be extremely helpful.

This is now part of the PR.

matthieubulte:
@crepererum I have added a `task_failed` handler, let me know what you think of it.

Note that the handler is transparent, in the sense that it will not be able to block or cancel the exception.

distributed/tests/test_worker_plugins.py
```diff
@@ -18,6 +23,37 @@ def teardown(self, worker):
...
+            if _kwargs is not None:
+                assert kwargs == _kwargs
+
+    def task_finished(self, worker, key, result):
+        assert isinstance(worker, Worker)
+
+        if len(self.expected_results) > 0:
+            assert result == self.expected_results.pop(0)
+
+    def task_failed(self, worker, key, exception):
```
crepererum: would be helpful to have the failed callback somewhat aligned with the Python `__exit__` functions, so we get `type` (exception type), `value` (exception value) and `tb` (traceback). See [Python docs](https://docs.python.org/3/library/stdtypes.html#typecontextmanager) as well. A use case would be that the handler could than analyze/report the traceback.

matthieubulte: that's a good idea, thanks!

crepererum:
BTW: I am not a maintainer here, so I can only advice / review. @mrocklin I think this PR is good enough now to be reviewed by you :)

matthieubulte:
Oh, thanks for pinging a maintainer! I have the feeling that some documentation is missing here, but I'm not sure where/what to add since Worker plugins aren't documented at all at the moment.

One way might be to create a base class, the same way it was done for scheduler plugins, and have a page in the documentation for them.

crepererum:
> One way might be to create a base class, the same way it was done for scheduler plugins, and have a page in the documentation for them.

Sounds like a great idea!

mrocklin:
Hey all.  My apologies for the late review.  Thank you @matthieubulte for starting this off and @crepererum for the review.  They're both much appreciated.  

A few high level notes come to mind.  I should have raised these earlier before a lot of the conversation.  I'm definitely guilty of failing the "Don’t let reviewers drag things out too long" [maintainer responsibility](https://matthewrocklin.com/blog/2019/05/18/maintainer).  

1.  We need to watch out for what runs in the event loop and what runs outside.  In particular functions like `apply_function` are being called in the thread pool, which I suspect will screw with users' plugins.
2.  I'm a little nervous about sprinkling many `notify_foo` functions in many places.  I think that it's likely that we'll forget some of them in the future.
3.  Instead, what do you think about having a single `plugin.task_transition` method that gets called with the old and new state as well as whatever keywords get passed in.  We would then call this once in this, somewhat central code
 https://github.com/dask/distributed/blob/810808e9ce4b3efeff1afc99e8fd8a9430f08a8c/distributed/worker.py#L1503-L1512

I think that this might be simpler and easier to maintain.  It does put a bit more burden onto the plugin writer.  They now need to have checks like `if start == "executing" and end == "memory"`, but I think that this might not be too burdensome.  This is also what we do in the scheduler today.  

https://github.com/dask/distributed/blob/810808e9ce4b3efeff1afc99e8fd8a9430f08a8c/distributed/scheduler.py#L4328-L4332

matthieubulte:
Hi @mrocklin , thanks for the feedback!

> which I suspect will screw with users' plugins.

You mean the confusion from having a copy of the plugin per thread?

> I'm a little nervous about sprinkling many notify_foo functions

Agree, I must say I even missed a couple of places at first...

> Instead, what do you think about...

That's a better idea, indeed..


Alright, I'll then just force push on this branch a rewrite of this feature. @mrocklin do you have a last comment about documentation, esp. wrt. what we were discussing earlier in the thread?

mrocklin:
> You mean the confusion from having a copy of the plugin per thread?

Yes.   Actually I suspect that we will have only one plugin, but that different threads will manipulate it at the same time, which can cause issues.

> Alright, I'll then just force push on this branch a rewrite of this feature. 

We squash-merge, so I wouldn't worry much about history.  It will and up being a single commit in the end.

> @mrocklin do you have a last comment about documentation, esp. wrt. what we were discussing earlier in the thread?

Having some plugin superclass could make sense if everything aligns well.  If not then maybe a separate WorkerPlugin.  But if that doesn't make sense in terms of code, then maybe just add normal rst docs to https://distributed.dask.org/en/latest/plugins.html ?

matthieubulte:
Hi,

I implemented the changes suggested. The only thing that bothers me a little bit is that there is little to no information available in the state transition.

Taking the logger example from @crepererum , the logger would only be able to log the fact that an exception happened, but wouldn't have access to the said exception. The only piece of data that seems to be available during a transition (at least the ones I could test) is the value returned by an execution. (executing -> memory)

Do you have any thoughts on that? I'm not sure if this form of plugin would then be useful.

mrocklin:
The information should be available in the worker's state though.  For example if you need the exception ...

```python
def transition(self, key, start, finish, ...):
    if finish == "exception":
        exc = self.worker.exceptions[key]
```

matthieubulte:
So, I think I'm done with the code changes and the unit tests, but I still need to figure out a good way to adapt the documentation.

distributed/diagnostics/__init__.py
```diff
@@ -1,6 +1,6 @@
 from ..utils import ignoring
 from .graph_layout import GraphLayout
-from .plugin import SchedulerPlugin
```
mrocklin: Renaming this module may cause some pain for users who use it currently.  

We would have to either provide a period of backwards compatibility, or we just keep the SchedulerPlugin and WorkerPlugin both in plugin.py (which is my preference).

distributed/diagnostics/worker_plugin.py
```diff
@@ -0,0 +1,55 @@
+class WorkerPlugin(object):
+    """ Interface to extend the Worker
+
+    Throughout the lifecycle of a task (see :doc:`Worker <worker>`), Workers are
+    instructed by the scheduler to compute certain tasks, resulting in transitions
+    in the state of each task. The Worker owning the task is then notified of this
+    state transition.
```
mrocklin: This introductory paragraph focuses on the transition method, but I suspect that many people today will also care about setup/teardown.  It might be good to have a general introductory paragraph, followed by the setup/teardown and then the transition use case.

distributed/diagnostics/worker_plugin.py
```diff
@@ -0,0 +1,55 @@
...
+    ...     def __init__(self, logger):
+    ...         self.logger = logger
+    ...
+    ...     def setup(self, worker):
+    ...         self.worker = worker
+    ...
+    ...     def transition(self, key, start, finish, *args, **kwargs):
+    ...         if finish == 'error':
+    ...             exc = self.worker.exceptions[key]
+    ...             self.logger.error("Task '%s' has failed with exception: %s" % (key, str(exc)))
```
mrocklin: Nice example!

distributed/worker.py
```diff
@@ -1055,7 +1055,7 @@ def _close(self, *args, **kwargs):
             setproctitle("dask-worker [closing]")
 
             teardowns = [
-                plugin.teardown(self)
+                plugin.teardown()
```
mrocklin: Why this change?  This will break backwards compatibility (people who have code that uses this feature today will suddenly have their code break).  If it's not critical, I recommend leaving this the way that it was so that we don't have to bother with a deprecation cycle.

distributed/diagnostics/worker_plugin.py
```diff
@@ -0,0 +1,55 @@
...
+
+    >>> l = ErrorLogger()
```
mrocklin: I recommend that we use the variable name `plugin` rather than `l` in this example just to be more descriptive.

distributed/worker.py
matthieubulte: The idea was to unify with the interface of the `SchulderPlugin`, but you're right, let's not break anything!

distributed/diagnostics/worker_plugin.py
matthieubulte: I did this to unify with the `SchedulerPlugin` example, but maybe I should have instead changes the example there. Will do!

docs/source/plugins.rst
```diff
@@ -1,7 +1,7 @@
 Scheduler Plugins
 =================
 
-.. autoclass:: distributed.diagnostics.plugin.SchedulerPlugin
+.. autoclass:: distributed.diagnostics.scheduler_plugin.SchedulerPlugin
```
mrocklin:
```suggestion
.. autoclass:: distributed.diagnostics.plugin.SchedulerPlugin
```

mrocklin: Do we also want to include the WorkerPlugin on this page?

distributed/worker.py
```diff
@@ -2680,6 +2681,13 @@ def get_call_stack(self, comm=None, keys=None):
         result = {k: profile.call_stack(frame) for k, frame in frames.items()}
         return result
 
+    def _notify_transition(self, key, start, finish, **kwargs):
+        for plugin in self.plugins.values():
+            try:
+                plugin.transition(key, start, finish, **kwargs)
```
mrocklin: Old plugins in existing user code may not have this method set.  Let's allow this for now.

```suggestion
                plugin.transition(key, start, finish, **kwargs)
            except AttributeError:
                pass
```

distributed/worker.py
```diff
@@ -2680,6 +2681,13 @@ def get_call_stack(self, comm=None, keys=None):
...
+            except Exception:
+                logger.info("Plugin failed with exception", exc_info=True)
```
mrocklin: We might want to include the name of the plugin in the logs.  Perhaps you could iterate over `name, plugin in self.plugins.items()` above and use the name here?

crepererum:
Wow, this turned out to be very elegant. Nice job! 🎉 :)

matthieubulte:
> Wow, this turned out to be very elegant. Nice job! 🎉 :)

I think @mrocklin and yourself are to thank for this. Thank you very much for the support on this PR!

mrocklin:
This is in.  Thank you @matthieubulte for the contribution and @crepererum for review.

Also, @matthieubulte I notice that this is your first code contribution to this repository.  Welcome!

matthieubulte:
Thank you for the great mentoring on this @mrocklin and @crepererum !

I'm glad I could contribute to this fantastic project, I'm a big fan 😉

PR #2995, opened 2019-08-25
Add threads= and memory= to Cluster.__repr__

mrocklin:
We're stealing this from dask-jobqueue

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -314,13 +314,6 @@ class Cluster(object):
     def _ipython_display_(self, **kwargs):
         return self._widget()._ipython_display_(**kwargs)
 
-    def __repr__(self):
-        return "%s(%r, workers=%d)" % (
-            type(self).__name__,
-            self.scheduler_address,
-            len(self.scheduler_info["workers"]),
-        )
-
     async def __aenter__(self):
         await self
         return self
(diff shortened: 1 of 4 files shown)


mrocklin:
cc @guillaumeeb for review if he has time

distributed/client.py
```diff
@@ -753,6 +753,21 @@ def sync(self, func, *args, asynchronous=None, callback_timeout=None, **kwargs):
                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
             )
 
+    def __repr__(self):
```
guillaumeeb: You have two __repr__ in client class now, is this voluntary? This looks like a copy paste mistake.

distributed/deploy/cluster.py
```diff
@@ -331,3 +324,18 @@ def __repr__(self):
     @property
     def scheduler_address(self):
         return self.scheduler_comm.address
+
+    def __repr__(self):
+        text = "%s(%r, workers=%d, threads=%d" % (
```
guillaumeeb: You're using `cores` in client, but `threads` in cluster.

distributed/client.py
```diff
@@ -761,12 +776,18 @@ def __repr__(self):
...
             nthreads = sum(w["nthreads"] for w in workers.values())
-            return "<%s: scheduler=%r processes=%d cores=%d>" % (
+            text = "<%s: %r processes=%d cores=%d" % (
                 self.__class__.__name__,
                 addr,
                 nworkers,
                 nthreads,
             )
+            memory = [w["memory_limit"] for w in workers.values()]
+            if all(memory):
```
guillaumeeb: Why is this needed? You may encounter None values?

distributed/deploy/cluster.py
guillaumeeb: At first thought I'd prefer cores, but threads is probably more accurate.

guillaumeeb: Let's stay with threads!

PR #3012, opened 2019-08-28
Tweak `Logs` styling

jcrist:
- Not all browsers render unstyled `summary` elements with the "carrot"
indicating they can be expanded. We style these so they always look the
same across browsers.
- We sort log elements when rendering.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1453,12 +1453,13 @@ class Logs(dict):
 
     def _repr_html_(self):
         summaries = [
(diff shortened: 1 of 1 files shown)


jcrist:
Before:

Screen Shot 2019-08-28 at 4 33 08 PM

After:

Screen Shot 2019-08-28 at 4 32 33 PM

mrocklin:
This seems fine to me.  I'll merge this in the morning just to give @jacobtomlinson a chance to glance at it.

mrocklin:
Thanks @jcrist and @jacobtomlinson !

PR #3013, opened 2019-08-29
Support Spec jobs that generate multiple workers

mrocklin:
Sometimes a single entry in the worker_spec will generate multiple Dask
workers.  We add an entry, "group", to the spec that shows how
worker_spec entries map to dask-workers that connect to the scheduler


mrocklin:
cc @guillaumeeb @jcrist

mrocklin:
OK, I'm going to merge this in.  Review would still be welcome after merging if anyone has thoughts.

Issue #2965, opened 2019-08-19
Worker task deserialization and ThreadPoolExecutor interactions


jrbourbeau:
I've had users report issues where a task arrives at a worker and is deserialized on the worker's main thread (here https://github.com/dask/distributed/blob/2bff61d9bee59e0bf655937922d9d4c37e49820a/distributed/worker.py#L1328) which causes code to be executed that interacts poorly with other tasks running concurrently on the worker `ThreadPoolExecutor`. Is there an existing mechanism that could help prevent this task deserialization and task execution from happening at the same time? 

One potential development could be to have task deserialization happen on the `ThreadPoolExecutor` right before the task is executed. This won't fix the problem in general, but would allow for limiting the worker's `ThreadPoolExecutor` to a single thread so that task deserialization and task execution never happen concurrently. This approach also has the added benefit of not deserializing tasks that are stolen before being executed. That said, I'm sure this approach also has some associated down sides and may not be appropriate for the common case. 

Does this approach seem like a reasonable way forward (I'm not sure if it is)? Is improving non-threadsafe code execution in scope for `distributed`?

mrocklin:
One straightforward way to resolve this would be with a `SerializedLock`, both in the deserialization code and in execution code.  However that assumes that people have control over that code.

It is not entirely unreasonable to put task deserialization in with execution.  However, I wouldn't be surprised if this uncovers as many or more bugs than it fixes.  I don't have particularly strong thoughts on this, except for a default preference for status quo.

mrocklin:
Another complication.  I *thought* that we cached deserialization of functions.  This is particularly valuable (maybe?) for numba jitted functions, where we don't want to have to keep jitting code.

It looks like we don't currently cache deserialization of functions though (although we do cache serialization).  Anyway, this is the sort of thing that would conflict with deserializing in different threads

jrbourbeau:
Thanks for the `SerializedLock` suggestion! I'll see if it's applicable to the particular use case I'm thinking of. 

IIUC using a lock to coordinate between `_deserialize` on the worker's main loop and running tasks in the `ThreadPoolExecutor` will cause the main loop to block and be unresponsive when executing longer running tasks (e.g. the worker won't be heartbeating, transferring dependencies, etc.). Do we have a way to coordinate handlers like `distributed/worker.py::add_task` in a non-blocking way?

> I don't have particularly strong thoughts on this, except for a default preference for status quo.

[...]

PR #3015, opened 2019-08-29
Move task deserialization to immediately before task execution

jrbourbeau:
This PR moves task deserialization on workers to just before the task is marked for execution on the worker executor.

This has a couple of benefits:

- In the case that a task is stolen, there won't be any unnecessary deserialization

- In the case of single-threaded workers, this will help reduce issues with task deserialization interfering with concurrently running tasks on the executor (ref #2965). This is because tasks are only transitioned to executing when the number of executing tasks is less than the number of executor theads https://github.com/dask/distributed/blob/cf26e1a559e2c89c4a4b14b6622111eaf0954f12/distributed/worker.py#L2361 IIUC this means, for single-threaded executors, tasks will only transition to executing when there are no currently executing tasks

On the other hand, a drawback to this approach is that deserialization errors are not immediately raised.

But overall, I'm in favor of moving deserialization based on the benefits outlined above. Curious to hear what others think.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1557,26 +1557,6 @@ class Worker(ServerNode):
                 pdb.set_trace()
             raise
 
-    def _maybe_deserialize_task(self, key):
-        if not isinstance(self.tasks[key], SerializedTask):
-            return self.tasks[key]
-        try:
-            start = time()
-            function, args, kwargs = _deserialize(*self.tasks[key])
-            stop = time()
-
-            if stop - start > 0.010:
-                self.startstops[key].append(("deserialize", start, stop))
-            return function, args, kwargs
-        except Exception as e:
-            logger.warning("Could not deserialize task", exc_info=True)
-            emsg = error_message(e)
-            emsg["key"] = key
-            emsg["op"] = "task-erred"
-            self.batched_stream.send(emsg)
-            self.log.append((key, "deserialize-error"))
-            raise
-
     def transition_ready_executing(self, key):
         try:
             if self.validate:
@@ -1588,7 +1568,7 @@ class Worker(ServerNode):
                     dep in self.data or dep in self.actors
                     for dep in self.dependencies[key]
                 )
-            self.tasks[key] = self._maybe_deserialize_task(key)
+
             self.executing.add(key)
             self.loop.add_callback(self.execute, key)
         except Exception as e:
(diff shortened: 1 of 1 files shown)


mrocklin:
> On the other hand, a drawback to this approach is that deserialization errors are not immediately raised.

Can you say more about what the previous and current behaviors were here?  Is it the same kind of feedback, but just time delayed, or is the kind of feedback different?  For example, if previously the user got an exception locally and now there is only a logging message on the worker, then I would be -1.  If it's a logging message in both cases, but just a bit delayed, then I would be comfortable with that.

jrbourbeau:
> Can you say more about what the previous and current behaviors were here? 

The feedback from deserialization errors is the same on the current `master` branch as well as with the changes in this PR. In both cases, a warning is logged and the scheduler is informed an error occurred with the task (worker sends a `"task-erred"` message to the scheduler). This message is just a bit delayed here. Later, if we call `.result()` on the errored task, then the error that occured during deserialization is re-raised locally.  

I've included an example script (based on `distributed/tests/test_client.py::test_badly_serialized_input`) which illustrates this behavior. 

<details>
<summary>Example script:</summary>

```python

from time import sleep
from toolz import identity


if __name__ == '__main__':

    class BadlySerializedObject(object):
        def __getstate__(self):
            return 1

        def __setstate__(self, state):
            raise ValueError("hello!")


    with Client() as c:
        o = BadlySerializedObject()

        # Submit task that can't be deserialized
        future = c.submit(identity, o)
        print(f'future = {future}')    # status: pending
        sleep(0.1)                     # Deserialization error logged
        print(f'future = {future}')    # status: error        

        try:
            future.result()
        except ValueError as e:
            print(f"ValueError with message {e} was raised")

```

</details>


I agree we shouldn't change the behavior the user gets here

jrbourbeau:
Just wanted to note that my most recent commit (4fbb089ea6aa398612ca80154533d300c65e4cad) had a couple of seemingly unrelated test failures on Travis CI that I wasn't able to reproduce locally. I restarted the build and now all tests are passing. @mrocklin do you know of any flaky tests popping up in other PRs? (I don't want to degrade the test suite)

mrocklin:
I've seen this a bit recently: https://github.com/dask/distributed/issues/3044

mrocklin:
Merging this in.   My apologies for the delay @jrbourbeau .

PR #3023, opened 2019-09-03
Add fallback html repr

### LINKED ISSUE #3020: ipywidget error on minimal install
This issue was filed/closed in [dask-jobqueue](https://github.com/dask/dask-jobqueue/issues/271) but doesn't seem to be open here. I just ran into it today, figured I'd file a bug report.

I have a minimal install of dask (just dask and dask.distributed) on miniconda. A cell with the following code

```python
from dask.distributed import LocalCluster
cluster = LocalCluster()
cluster
```

throws the following traceback.
```python
ModuleNotFoundError                       Traceback (most recent call last)
~\AppData\Local\Continuum\miniconda3\envs\dask-tutorial\lib\site-packages\IPython\core\formatters.py in __call__(self, obj)
    916             method = get_real_method(obj, self.print_method)
    917             if method is not None:
--> 918                 method()
    919                 return True
    920 

~\AppData\Local\Continuum\miniconda3\envs\dask-tutorial\lib\site-packages\distributed\deploy\cluster.py in _ipython_display_(self, **kwargs)
    206 
    207     def _ipython_display_(self, **kwargs):
--> 208         return self._widget()._ipython_display_(**kwargs)

~\AppData\Local\Continuum\miniconda3\envs\dask-tutorial\lib\site-packages\distributed\deploy\cluster.py in _widget(self)
    143             pass
    144 
--> 145         from ipywidgets import Layout, VBox, HBox, IntText, Button, HTML, Accordion
    146 
    147         layout = Layout(width='150px')

ModuleNotFoundError: No module named 'ipywidgets'
```
Let me know if you need more configuration info.

jcrist:
This PR fixes two things:
- Cluster objects don't error when repr'd in notebooks without ipywidgets installed.
- If ipywidgets isn't installed, a fallback HTML repr is used instead of the default `repr` string. This was borrowed from my work in dask-gateway (https://github.com/jcrist/dask-gateway/pull/70).

A test is added to test both of these.

Fixes #3020

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -246,7 +246,11 @@ class Cluster(object):
         except AttributeError:
             pass
 
-        from ipywidgets import Layout, VBox, HBox, IntText, Button, HTML, Accordion
+        try:
+            from ipywidgets import Layout, VBox, HBox, IntText, Button, HTML, Accordion
+        except ImportError:
+            self._cached_widget = None
+            return None
 
         layout = Layout(width="150px")
 
@@ -258,7 +262,7 @@ class Cluster(object):
         else:
             link = ""
 
-        title = "<h2>%s</h2>" % type(self).__name__
+        title = "<h2>%s</h2>" % self._cluster_class_name
         title = HTML(title)
         dashboard = HTML(link)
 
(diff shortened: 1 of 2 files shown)


jcrist:
The dashboard address is displayed if the dashboard is running. This could likely be styled better, but I'm not a CSS person. We could also include more information, although I don't want to include anything that's dynamic (like n-workers, cores, etc...) as these won't update live like they will with the widget repr.

Screen Shot 2019-09-03 at 12 11 47 PM

mrocklin:
Merging.  Thanks @jcrist !

PR #3027, opened 2019-09-05
Fix worker preload config

byjott:
This change makes the workers honor the config file option: so far, the `Nanny` has a default of `preload=()`,which is passed on to the `Worker`. However, the `Worker`s only read the config file value in case of `preload is None`.
This PR changes the default in the `Nanny` to `preload=None`, so the worker will use the value from the config file.


quasiben:
Thanks @byjott for the PR!  The test is currently failing on the assert.  I was actually going to suggest that rather than create a global you could run a small function as outlined in the [custom-startup docs](https://docs.dask.org/en/latest/setup/custom-startup.html).  Something like:

```python
from distributed import get_worker

class MyPlugin:
        self.started = False
    def setup(self, worker: dask.distributed.Worker):
        self.started = True
    def teardown(self, worker: dask.distributed.Worker):
        pass

plugin = MyPlugin(1, 2, 3)
client.register_worker_plugin(plugin, name='my-plugin')

def f(): 
    worker = get_worker()
    plugin = worker.plugins['my-plugin']
    return plugin.started

future = client.run(f)
assert all(value == True for value in future.values())
```

byjott:
> Thanks @byjott for the PR! The test is currently failing on the assert. I was actually going to suggest that rather than create a global you could run a small function as outlined in the [custom-startup docs](https://docs.dask.org/en/latest/setup/custom-startup.html). Something like:
> [...]

Well, that's weird: The tests ran fine locally (although I ran only tests from `test_config.py`, so maybe something breaks when running it in a larger test session).

Note that what I wanted to test here is that specifying `distributed.worker.preload` in the config file works as expected if starting the `Worker` processes via a `Nanny`. What you propose would effectively be a test that the plugin's `setup` method is called (which is not what I wanted to test and assumed this is covered elsewhere).

I'll re-write the failed test and avoid the global flag. Instead, I'll use a flag on the `Worker` instance. Hopefully, this is much more robust than the module global.

TomAugspurger:
Is it possible that the lack of a `distributed/tests/__init__.py` file makes it so that test preload isn't found?

Perhaps you can put the code in a string and write it to a tempfile as part of the test setup?

quasiben:
@byjott what do you think about either:

1) parameterizing `test_worker_preload_module` to use the config you've created instead of the hardcoded `worker_kwargs`

2) copying the `test_worker_preload_module` test and passing in the config as you are in the test you currently have ?

Reusing the existing test pattern in `test_preload` should help make debugging any/all failures a bit easier

byjott:
> Is it possible that the lack of a `distributed/tests/__init__.py` file makes it so that test preload isn't found?

No, that's not what's happening, as this would lead to a visible import error when the worker is starting.

I think I now found out what is a happening: what the test *attempts* to do is to write out a customized config file, with the expectation that this is picked up by the worker processes started with `LocalCluster`.
The way it's done is that `distributed.utils_test.new_config_file` sets up the new config file in a temporary location and sets the environment variable "DASK_CONFIG" to point to that location. However, this environment variable is not visible to the workers in all cases: `LocalCluster` starts worker processes via a forkserver process. If that forkserver process is started the first time for this test, everything is fine ("DASK_CONFG" env var reached the forkserver, and it reaches the worker process forked from the forkserver process).
If, however, another tests already started the forkserver, the setting the environment variable "DASK_CONFIG" in the test is not visible in the subprocesses.

This explains why running only this test alone works fine, but running it within the whole testsuite fails.

To reproduce, observe that just running the new test I added works fine on this branch:
```
pytest -k test_preload_config
```
while this will fail, because another `LocalCluster` starts subprocesses (and a forkserver)
```
pytest -k 'test_preload_config or test_quiet_quit_when_cluster_leaves'
```

So one should really never attempt to use `distributed.utils_test.new_config_file` mixed with forkserver-started subprocesses in the tests ...

As a minimal showcase on the level of `multiprocessing`, consider this:
```python
import multiprocessing
import os

def print_var():
    print(f"{os.getpid()}: {os.environ['TEST_ENV_VAR']}")

if __name__ == '__main__':
  # multiprocessing.set_start_method('forkserver')  # see below

  os.environ["TEST_ENV_VAR"] = "VALUE_1"
  multiprocessing.Process(target=print_var).start()

  os.environ["TEST_ENV_VAR"] = "VALUE_2"
  multiprocessing.Process(target=print_var).start()
```

If run like this, it will print something like
```
13083: VALUE_1
13084: VALUE_2
```

However, when commenting in the line that switches to the "forkserver" start method, the output is something like this:
```
13090: VALUE_1
13091: VALUE_1
```

I'll have to think some more on how to make this test better ...

byjott:
> @byjott what do you think about either:
> 
> 1. parameterizing `test_worker_preload_module` to use the config you've created instead of the hardcoded `worker_kwargs`
> 2. copying the `test_worker_preload_module` test and passing in the config as you are in the test you currently have ?
> 
> Reusing the existing test pattern in `test_preload` should help make debugging any/all failures a bit easier

I'm not sure how this would work. I want to write a test that exposes the bug I fixed. This needs:
 * go through the `Nanny` (not just the worker), using a default for the "preload" argument (because this is where the bug is), and *not* passing this config via `worker_kwargs`
 * the worker should see the `preload` value of `None` and consult the configuration, so I need the config value for `distributed.worker.preload` to change, as seen in the worker.

Neither of that is done in these tests, nor do I see how to trivially add that.

TomAugspurger:
> I'm not sure how this would work. I want to write a test that exposes the bug I fixed. This needs:

@byjott try something like

```
diff --git a/distributed/tests/test_preload.py b/distributed/tests/test_preload.py
index 07ee56d8..93607e19 100644
--- a/distributed/tests/test_preload.py
+++ b/distributed/tests/test_preload.py
@@ -1,8 +1,12 @@
+import contextlib
 import os
 import shutil
 import sys
 import tempfile
 
+import pytest
+
+from dask import config
 from distributed.utils_test import cluster
 from distributed.utils_test import loop  # noqa F401
@@ -19,7 +23,9 @@ def get_worker_address():
 """
 
 
-def test_worker_preload_file(loop):
+@pytest.mark.parametrize("method", ["kwargs", "config"])
+def test_worker_preload_file(loop, method):
+
     def check_worker():
         import worker_info
 
@@ -31,13 +37,21 @@ def test_worker_preload_file(loop):
         with open(path, "w") as f:
             f.write(PRELOAD_TEXT)
 
-        with cluster(worker_kwargs={"preload": [path]}) as (s, workers), Client(
-            s["address"], loop=loop
-        ) as c:
-
-            assert c.run(check_worker) == {
-                worker["address"]: worker["address"] for worker in workers
-            }
+        if method == "kwargs":
+            worker_kwargs = dict(preload=[path])
+            context = contextlib.nullcontext()
+        else:
+            worker_kwargs = dict()
+            context = config.set({"distributed.worker.preload": [path]})
+
+        with context:
+            with cluster(worker_kwargs=worker_kwargs) as (s, workers), Client(
+                s["address"], loop=loop
+            ) as c:
+
+                assert c.run(check_worker) == {
+                    worker["address"]: worker["address"] for worker in workers
+                }
     finally:
         shutil.rmtree(tmpdir)
 ```

that fails for me on master. Presumably with your fix, it will pass.

**Comment by byjott** [comment]
> try something like

Thanks for your proposal, @TomAugspurger. Unfortunately, that can't be correct: note that the fix is a single line in `Nanny.__init__`, and this line is not exercised by this test ...

I tried some variants similar to that, but what I do get into is a hell from the global configuration in `dask.config.config`, which (if modified in the tests) are not visible in subprocesses; no good mechanism of applying such modifications s.t. it is visible in subprocesses. Or, if it is, just if you run it the first time.

So I'll just give up to write a test for something that apparently was never tested before, but keep this PR that fixes a real issue.

**Comment by TomAugspurger** [comment]
I suspect there's a way to get the config passed through to the nanny / worker process within the test. I'll try to take a look later today, but may not have time.

**Comment by mrocklin** [comment]
> I suspect there's a way to get the config passed through to the nanny / worker process within the test. I'll try to take a look later today, but may not have time.

Note that Nanny's have an `env=` keyword.  So you could pass config as an environment variable if that's what you're looking for.  

For something like a preload script though I would expect the nanny to pick that up from configuration, and then pass that value down to the worker as a normal keyword argument, overriding whatever the worker would pick up from a config file.

**Comment by mrocklin** [comment]
There was an intermittent failure.  I reported it here and have restarted: https://github.com/dask/distributed/issues/3044

**Comment by byjott** [comment]
> > I suspect there's a way to get the config passed through to the nanny / worker process within the test. I'll try to take a look later today, but may not have time.
> 
> Note that Nanny's have an `env=` keyword. So you could pass config as an environment variable if that's what you're looking for.

Yes, I saw that. But I did not get that working to override the config (I tried to set the `DASK_CONFIG` env var to a temporary location where the modified config resides).

> For something like a preload script though I would expect the nanny to pick that up from configuration, and then pass that value down to the worker as a normal keyword argument, overriding whatever the worker would pick up from a config file.

Alright, I can also make the change along these lines, i.e. read from the config only in the `Nanny` and not in the `Worker`.

**Comment by mrocklin** [comment]
> Alright, I can also make the change along these lines, i.e. read from the config only in the Nanny and not in the Worker.

To be clear, the worker also needs to check config if the provided value is None.  Workers aren't always started with Nannies.  If a Worker is started with a Nanny though, then I would expect the nanny to read the configuration, get the appropriate value, and then pass that to the worker.  Because the worker is being handed a concrete value, it won't feel the need to check with the config files.

**Comment by jrbourbeau** [comment]
Hey @byjott, just checking in here. Are you stuck on any blockers? I'd also expect the behavior outlined in https://github.com/dask/distributed/pull/3027#issuecomment-529964405

**Comment by byjott** [comment]
> Hey @byjott, just checking in here. Are you stuck on any blockers?

Yes, the blocker has not changed and is testing config changes seems not really supported right now. To be more precise, what I want to do is to change the config in a test in a way it is effective in the worker process spawned during the test via a Nanny. I got it working if running only the particular test, but not if running the whole test suite, because of the interactions via multiprocess forkserver with preloading in tests ...

As far as I can see, this would require major new features in tests, which I'm hesitant of implementing, as they all have downsides and I can't really make the call of what to do here.

**Comment by mrocklin** [comment]
The nanny itself runs in the main process, so you should be able to do something like the following:

```python
with dask.config.set(...):
    async with Scheduler(...) as s:
        async with Nanny(s.address, ...) as n:
            assert n.preload == ...
```

The nanny then takes these values and passes them to the worker through a process.  But I don't think that you to control the configuration in that process.

byjott:
@mrocklin Thanks for adding the test and getting this merged!
Last week, I just didn't find the time to do it myself; it was on my agenda today ;-)

mrocklin:
No worries.  Thanks for noticing the original bug.

PR #3034, opened 2019-09-06
Set the x_range limit of the Meory utilization plot to memory-limit

mrocklin:
This should give a better sense for how close we are to filling up
memory, which seems to be the biggest use of the memory plot.


quasiben:
This looks good!  Yeah, viz improvements 👾

PR #3039, opened 2019-09-09
Detect cgroups resource limits

### LINKED ISSUE #1712: Improve memory available detection
In https://github.com/dask/distributed/blob/5ef9c806bf341a4c06d17d24eb94838b307d4871/distributed/worker.py#L63 you use psutil to known the amount of memory available.

However, in the case of the worker is launched in a cgroup with a memory limitation (memory.limit_in_bytes) the value returned by psutil is the amount of the host, not the amount available in the cgroups, there fore the worker try to use more memory than it can and the linux kernel kill it.

jcrist:
This adds support for detecting resource (CPU and memory) limits set
using `cgroups`. This makes dask resource detection play nicer with
container systems (e.g. `docker`), rather than detecting the host memory
and cpus avaialable.

This also centralizes all queries about the host platform to a single
module (`distributed.platform`), with top-level constants defined for
common usage.

Fixes #1712

### DIFF
--- a/distributed/platform.py
+++ b/distributed/platform.py
@@ -1,0 +1,70 @@
+import os
+import sys
+
+import psutil
+
+__all__ = ("memory_limit", "cpu_count", "PLATFORM_MEMORY_LIMIT", "PLATFORM_CPU_COUNT")
+
+
+def memory_limit():
+    """Get the memory limit (in bytes) for this system.
+
+    Takes the minimum value from the following locations:
+
+    - Total system host memory
+    - Cgroups limit (if set)
+    - RSS rlimit (if set)
+    """
+    limit = psutil.virtual_memory().total
+
+    # Check cgroups if available
+    if sys.platform == "linux":
+        try:
+            with open("/sys/fs/cgroup/memory/memory.limit_in_bytes") as f:
+                cgroups_limit = int(f.read())
+            if cgroups_limit > 0:
+                limit = min(limit, cgroups_limit)
+        except Exception:
+            pass
+
+    # Check rlimit if available
+    try:
+        import resource
+
+        hard_limit = resource.getrlimit(resource.RLIMIT_RSS)[1]
+        if hard_limit > 0:
+            limit = min(limit, hard_limit)
+    except (ImportError, OSError):
+        pass
+
+    return limit
+
+
+def cpu_count():
+    """Get the available CPU count for this system.
+
+    Takes the minimum value from the following locations:
+
+    - Total system cpus available on the host.
+    - Cgroups limit (if set)
+    """
+    count = os.cpu_count()
+
+    # Check cgroups if available
+    if sys.platform == "linux":
+        try:
+            with open("/sys/fs/cgroup/cpuacct,cpu/cpu.cfs_quota_us") as f:
+                quota = int(f.read())
+            with open("/sys/fs/cgroup/cpuacct,cpu/cpu.cfs_period_us") as f:
+                period = int(f.read())
+            cgroups_count = int(quota / period)
+            if cgroups_count > 0:
+                count = min(count, cgroups_count)
+        except Exception:
+            pass
+
+    return count
+
+
+PLATFORM_MEMORY_LIMIT = memory_limit()
+PLATFORM_CPU_COUNT = cpu_count()
--- a/distributed/platform.py
+++ b/distributed/platform.py
@@ -46,9 +46,18 @@ def cpu_count():
     Takes the minimum value from the following locations:
 
     - Total system cpus available on the host.
+    - CPU Affinity (if set)
     - Cgroups limit (if set)
     """
     count = os.cpu_count()
+
+    # Check CPU affinity if available
+    try:
+        affinity_count = len(psutil.Process().cpu_affinity())
+        if affinity_count > 0:
+            count = min(count, affinity_count)
+    except Exception:
+        pass
 
     # Check cgroups if available
     if sys.platform == "linux":
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -1,6 +1,5 @@
 import atexit
 import logging
-import multiprocessing
 import gc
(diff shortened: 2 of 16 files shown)


distributed/platform.py
```diff
@@ -0,0 +1,70 @@
...
+            if cgroups_count > 0:
+                count = min(count, cgroups_count)
+        except Exception:
+            pass
+
+    return count
+
+
+PLATFORM_MEMORY_LIMIT = memory_limit()
+PLATFORM_CPU_COUNT = cpu_count()
```
mrocklin: Thoughts on removing the `PLATFORM_` prefix here?  If we wanted the name then we could use Python namespaces like 

```python
from dask.distributed import platform

platform.CPU_COUNT
```

jcrist: Yeah, I thought about that. The issue here is that `platform` is already a python module. Happy to go with a different prefix and rename the module, naming is hard.

jcrist: Other potential module names (in the module scope of "query things about the host platform")):

- `system`
- `osutils`
- `os_utils`
- `sysutils`

pentschev: As I was writing "better names", I realized they were all taken or unclear. I think in general I would prefer to leave the current names, even if they feel less Pythonic/intuitive, IMHO clarity makes a big case here.

pentschev:
On a very high-level overview, this looks good to me and I also like this idea, so thanks for working on that @jcrist.

One alternative to reading `/sys` directly is to use https://pypi.org/project/cgroup-utils/, have you checked/thought about that? Not trying to make a case in favor of it, just pointing to its existence, since adding that dependency may also be undesirable.

pentschev:
Forgot to mention, but I can take a more detailed look at this PR tomorrow.

jcrist:
> One alternative to reading /sys directly is to use https://pypi.org/project/cgroup-utils/, have you checked/thought about that?

There's a few different libraries for doing this sort of thing, but all of them seemed to be overkill for the simple queries we need, and there didn't seem to be a strong community standard lib to use. Opted for the simplest thing that works.

distributed/platform.py
mrocklin: I think that `platform.CPU_COUNT` is as clear as `PLATFORM_CPU_COUNT` or `system.CPU_COUNT` is as clear at `SYSTEM_CPU_COUNT`.  I'm slightly in favor of `system.CPU_COUNT`.

jcrist: My only issue with importing as `platform` was that it matches a stdlib module name. `system` does not, so I'm happy to switch to that and drop the prefixes.

mrocklin: That would be my preference if that's ok with you.

distributed/platform.py
```diff
@@ -0,0 +1,70 @@
...
+
+def cpu_count():
+    """Get the available CPU count for this system.
+
+    Takes the minimum value from the following locations:
+
+    - Total system cpus available on the host.
+    - Cgroups limit (if set)
+    """
+    count = os.cpu_count()
```
dhirschfeld: Since `psutil` is a dep could we use it in preference here as it's more robust that the Python implementation.

In particular I'm running into a Python bug (win64/py37) whereby it thinks I have 64 cores when my VMWare VM only has 4 VCPUs allocated:
```python
In [3]: os.cpu_count()
Out[3]: 64

In [4]: psutil.cpu_count()
Out[4]: 4
```

Furthermore, the number of cores on the system isn't representative of the number of cores available to the process.  Docker containers are only one example of this. On some HPC systems it's usual for processes to have their affinity set to lock them to particular cores. Since we're making improvements here to consider the number of cores *available* I think it would be a useful improvement to use the cpu affinity rather than the cpu count:
```python
In [5]: psutil.Process().cpu_affinity()
Out[5]: [0, 1, 2, 3]

In [6]: len(psutil.Process().cpu_affinity())
Out[6]: 4
```
```suggestion
    count = len(psutil.Process().cpu_affinity())
```

***xref:*** https://bugs.python.org/issue26692#msg310113

jacobtomlinson:
This all looks good to me. I don't feel I can add more to the existing reviews right now.

distributed/platform.py
```diff
@@ -0,0 +1,70 @@
+import os
```
TomAugspurger: Do we intend this module to be public? If so, add these to the API docs?

jcrist: No, I think this should be just another internal module. Kept it out of `utils.py`, as that module is huge.

mrocklin:
If you're good to change PLATFORM to .system then I'd be happy to merge

distributed/platform.py
TomAugspurger: Then do we want to prefix with an underscore?

jcrist: We don't do this for utils, etc..., so I don't think that's needed here.

TomAugspurger: Right, but was that a deliberate decision? I don't want to make the situation "worse" if we aren't making any API guarantees about these methods.

mrocklin: Yeah, I don't personally think that we should underscore all of the non-user-facing modules.  I think that we just shouldn't include them in `__init__.py` and not publicize them too broadly.  If people are reaching into our internal module structure I personally don't mind burning them.

TomAugspurger: OK. I'll likely open up a PR at least documenting that if they're not part of the publicly documented API then they're considered internal.

distributed/platform.py
jcrist: Cool, good to know. I've added this to the set of things we check, but we can't rely on it 100% since it's not available on all platforms.

Out of curiousity, does `os.sched_getaffinity(0)` work for your use case?

jcrist:
I believe all comments have been addressed. Should be good-to-go on tests pass.

mrocklin:
Thanks @jcrist  .  This is in.  I think that this will make folks happy.

PR #3040, opened 2019-09-09
Drop joblib shim module in distributed

jakirkham:
This already just raises an `ImportError` and has done so for a while. Seems reasonable to just drop the module and let Python raise the `ImportError` that it would anyways. Likely everyone has migrated to this convention by now.

xref: https://github.com/dask/distributed/pull/2298


TomAugspurger:
Yep, thanks.

PR #3045, opened 2019-09-10
Remove six

mrocklin:
The one I didn't know how to do was this:

```python
class Listener(six.with_metaclass(ABCMeta)):
```

Is anyone familiar enough with the syntax of metaclasses in Python 3 so that this is easy for them?  @jcrist maybe?


jcrist:
```
class Listener(metaclass=ABCMeta):
```

For this specific case though you could also derive from `abc.ABC` and skip the metaclass.

PR #3052, opened 2019-09-13
Allow full script in preload inputs

mrocklin:
Previously we allowed either a script or a module in Scheduler or
Worker preload arguments.

    worker = Worker(..., preload='myfile.py')

However for simple scripts we sometimes don't want to create a full
file, but just want to include the script as a  multi-line text value.

    script = """def dask_setup(worker):\n    worker.foo=123"""
    worker = Worker(..., preload=script)

cc @quasiben this may help with the CUDA context work

### DIFF
--- a/distributed/preloading.py
+++ b/distributed/preloading.py
@@ -8,6 +8,8 @@ from importlib import import_module
 
 import click
 
+from dask.utils import tmpfile
+
 from .utils import import_file
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


distributed/preloading.py
```diff
@@ -77,36 +78,36 @@ def _import_modules(names, file_dir=None):
...
-                copy_dst = os.path.join(file_dir, basename)
-                if os.path.exists(copy_dst):
-                    if not filecmp.cmp(name, copy_dst):
-                        logger.error("File name collision: %s", basename)
-                shutil.copy(name, copy_dst)
-                module = import_file(copy_dst)[0]
-            else:
-                module = import_file(name)[0]
-
+    if "\n" in name:  # not a name, actually the text of the script
```
jcrist: What if it's a one-liner script? Perhaps a safer check would be:

- If it ends in `.py` it's a file to import
- Elif it has no spaces, it's a module name
- Else it's a script to run

mrocklin: Nice idea.  Done.

PR #3055, opened 2019-09-15
Add small section on disabling work stealing

cicdw:
I figured this deserved a mention in the work stealing doc.

Also, this may be known but it appears that the Sphinx redirects aren't currently working, which affects the discoverability of the section on Dask configuration from the Distributed search (searching for "Configuration" on the distributed site points to [a broken link](http://distributed.dask.org/en/latest/configuration.html)).  I saw the config redirect was only added [6 days ago](https://github.com/dask/distributed/pull/3038), so maybe the docs haven't been rebuilt, but thought it was worth calling out.


mrocklin:
Thanks @cicdw !

PR #3056, opened 2019-09-16
Check multiple cgroups dirs, ceil fractional cpus

### LINKED ISSUE #3053: cpu cores estimate based on cgroups
I was reading the announcement for 2.4.0 and got interested in https://github.com/dask/distributed/pull/3039 by @jcrist 

That did lead me to this part of the code:
https://github.com/dask/distributed/blob/7d017c467590c758fa4b8cb2b1193205fe5aa7ad/distributed/system.py#L62

Just by looking at it (and although I'm not an expert I think I know what's going on), I have to observations -- half way between a question and a bugreport.

1. in my docker environment I have here (ubuntu 18.04), the filename is different:

```
~$ cat /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us 
100000
~$ cat /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us 
220000
```
in the code is that path:
```
~$ ls /sys/fs/cgroup/cpuacct,cpu/cpu.cfs_quota_us
ls: cannot access '/sys/fs/cgroup/cpuacct,cpu/cpu.cfs_quota_us': No such file or directory
```

2. The actual calculation is `int(quota / period)`. I think this should round up, not down. The point is, if you have a fraction like "2.5 cores", it will report 2 cores and keep half a core unused. It would be better if it reports 3 cores and then cgroups limits to the actual 2.5.

jcrist:
A few fixes for resource detection using cgroups:

- The directory for determining cpu availability isn't standardized
across linux distros, could be either `cpuacct,cpu`, or `cpu,cpuacct`.
We now check for both.
- When allotted fractional cpus (e.g. 1.5), we now round up.

Also adds tests for both CPU and memory limit detection under cgroups,
by monkeypatching in fake files.

Fixes #3053.

### DIFF
--- a/distributed/system.py
+++ b/distributed/system.py
@@ -1,3 +1,4 @@
+import math
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


mrocklin:
It looks like there is a failing test:

```python
        count = cpu_count()
        if dirname:
            # Rounds up
>           assert count == 201
E           assert 2 == 201
```

jcrist:
Hmmm, I wonder if `rlimit` is set. We don't monkeypatch this.

jcrist:
Yeah, that was it. All green now.

PR #3061, opened 2019-09-18
Allow specification of worker type in SSHCLuster

jacobtomlinson:
This PR is the result of a conversation with @pentschev about how `SSHCluster` could be used with multiple GPU servers.

This change allows the user to specify which worker module to run on remote workers when using `SSHCluster`. The default is `distributed.cli.dask_worker` but the user could optionally specify `dask_cuda.dask_cuda_worker` if they want GPU workers. It is also generic enough that they could specify something else if another worker type exists.

### DIFF
--- a/distributed/deploy/ssh2.py
+++ b/distributed/deploy/ssh2.py
@@ -70,15 +70,15 @@ class Worker(Process):
         self,
         scheduler: str,
         address: str,
-        worker_module: str,
         connect_kwargs: dict,
         kwargs: dict,
+        worker_module="distributed.cli.dask_worker",
         loop=None,
         name=None,
     ):
         self.address = address
         self.scheduler = scheduler
-        self.worker_module = worker_module or "distributed.cli.dask_worker"
+        self.worker_module = worker_module
         self.connect_kwargs = connect_kwargs
         self.kwargs = kwargs
         self.name = name
@@ -166,7 +166,7 @@ def SSHCluster(
     connect_kwargs={},
     worker_kwargs={},
     scheduler_kwargs={},
-    worker_module=None,
+    worker_module="distributed.cli.dask_worker",
     **kwargs
 ):
     """ Deploy a Dask cluster using SSH
@@ -226,9 +226,9 @@ def SSHCluster(
             "cls": Worker,
             "options": {
                 "address": host,
-                "worker_module": worker_module,
                 "connect_kwargs": connect_kwargs,
                 "kwargs": worker_kwargs,
+                "worker_module": worker_module,
             },
         }
         for i, host in enumerate(hosts[1:])
--- a/distributed/deploy/ssh2.py
+++ b/distributed/deploy/ssh2.py
@@ -78,7 +78,7 @@ class Worker(Process):
     ):
         self.address = address
         self.scheduler = scheduler
-        self.worker_module = worker_module
+        self.worker_module = worker_module or "distributed.cli.dask_worker"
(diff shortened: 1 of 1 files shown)


distributed/deploy/ssh2.py
```diff
@@ -68,13 +70,15 @@ def __init__(
         self,
         scheduler: str,
         address: str,
+        worker_module: str,
```
pentschev: Not sure if anyone relies on this class directly, but if it's a possibility, maybe we should make `"distributed.cli.dask_worker"` the default value?

jacobtomlinson: Good point. It is the default up in the `SSHCluster` class, and I assumed that the `Worker` class would never be used anywhere else. But maybe that's a bad assumption?

pentschev:
I added only a small comment, but this looks great, thanks a lot for taking the time to do it @jacobtomlinson !

distributed/deploy/ssh2.py
pentschev: I don't know if it's a bad assumption, maybe nobody uses it, but since it's easy to let the current behavior unchanged, probably better to play it safe.

mrocklin: I think it makes sense to set `worker_module="distributed.cli.dask_worker"` as the default value.  This is probably slightly nicer for people looking at the docstring.  It gives them a sense for what kind of value they should place here.

mrocklin:
This seems good to me.   I'm happy to merge as-is.  I'm also happy with a change of the default value.

distributed/deploy/ssh2.py
jacobtomlinson: Sure! In which class `SSHCluster` or `Worker` or both?

pentschev:
> This seems good to me. I'm happy to merge as-is. I'm also happy with a change of the default value.

The default value was already changed in the latest commit, so I would be +1 to merging as is.

distributed/deploy/ssh2.py
mrocklin: I would add the default in `SSHCluster` because that is the most visible.  I think of the `Worker` class as being mostly an implementation detail.  It could be added there though.  Adding it in both places is probably the most safe option.

pentschev:
Hmm, I just realized that `dask-ssh` is using `ssh.py` rather than `ssh2.py`, and that already allows passing a custom worker https://github.com/dask/distributed/blob/master/distributed/deploy/ssh.py#L267. What's the difference between the two?

mrocklin:
ssh2.py is a new and somewhat experimental implementation that uses the new SpecCluster machinery.  Internally it is much simpler for us to maintain and probably a bit more generic.

ssh.py is very old, but also probably more feature-rich.  If it works for you then I would use it.

pentschev:
> ssh2.py is a new and somewhat experimental implementation that uses the new SpecCluster machinery. Internally it is much simpler for us to maintain and probably a bit more generic.

Thanks for clarifying.

> ssh.py is very old, but also probably more feature-rich. If it works for you then I would use it.

I'll check them out and see, I'll try to use the newer one if possible, otherwise I'll try the older.

jacobtomlinson:
Ok I think this is good to go

pentschev:
+1 from me. Thanks @jacobtomlinson for this PR!

mrocklin:
Merging.  Thanks @jacobtomlinson !

(also thanks @pentschev for the poke)

PR #3064, opened 2019-09-18
Make Client.get_versions async

jacobtomlinson:
Switched `Client.get_versions` to be async and renamed to `_get_versions` and use `self.sync` to call it synchronously under the original name.

Also added an async client fixture for testing.

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -91,6 +91,7 @@ from distributed.utils_test import (
 from distributed.utils_test import (  # noqa: F401
     client as c,
     client_secondary as c2,
+    async_client as ac,
     cluster_fixture,
     loop,
     loop_in_thread,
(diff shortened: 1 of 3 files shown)


distributed/tests/test_client.py
```diff
@@ -3724,6 +3725,11 @@ def test_get_versions(c):
     assert dict(v["client"]["packages"]["optional"])["requests"] == requests.__version__
 
 
+@pytest.mark.asyncio
+async def test_async_get_versions(ac):
```
mrocklin: Currently we use the `gen_cluster` decorator for this:

```suggestion
@gen_cluster(client=True)
async def test_async_get_versions(c, s, a, b):
```

This does a bunch of other things that are useful.  Here are a few:

1.  It's much faster than using `LocalCluster`, probably something like 10ms rather than a couple seconds
2. It gives us some workers and the scheduler too
3.  It does a ton of checks on cleanup (see the cleanup fixture)
4.  It adds in the `pytest.mark.asyncio` decorator by default (you should also remove that line)

jacobtomlinson: Ah nice!

mrocklin:
This looks great.  Merging.  My apologies for letting this linger.

PR #3066, opened 2019-09-18
Add task state counts to Scheduler identity

jrbourbeau:
This PR adds `"task_counts"` and `"time"` keys to `Scheduler.identity`. This provides a convenient way to quickly get the distribution of task states (i.e. how many are processing, how many are in memory, etc.).

In addition, since `Client._update_scheduler_info` is used in a `PeriodicCallback`, adding `"task_counts"` to `Scheduler.identity` allows users to get the most recently cached distribution of tasks without an additional round trip to the scheduler. This is nice in situations where the scheduler becomes unresponsive for a period of time (e.g. when processing a large task graph). 

I also wanted to note that adding a `Scheduler.task_counts` attribute isn't strictly necessary as `Scheduler.tasks` contains all the necessary information to determine how many tasks are in each state. However, since `Client._update_scheduler_info` is used in a `PeriodicCallback`, I wanted to avoid repeated loops over all tasks the `Scheduler` is tracking.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -959,6 +959,7 @@ class Scheduler(ServerNode):
         self.unrunnable = set()
 
         self.n_tasks = 0
+        self.task_counts = {state: 0 for state in ALL_TASK_STATES}
         self.task_metadata = dict()
         self.datasets = dict()
 
@@ -1159,9 +1160,11 @@ class Scheduler(ServerNode):
             "id": str(self.id),
             "address": self.address,
             "services": {key: v.port for (key, v) in self.services.items()},
+            "task_counts": self.task_counts,
             "workers": {
                 worker.address: worker.identity() for worker in self.workers.values()
             },
+            "time": time(),
         }
         return d
 
@@ -4315,6 +4318,16 @@ class Scheduler(ServerNode):
                     ts.state,
                     dict(recommendations),
                 )
+            if finish2 == "forgotten":
+                # Not keeping track of forgotten tasks
+                self.task_counts[start] -= 1
+            elif start == "released" and finish2 == "waiting":
+                # Newly added tasks which are waiting to be processed
+                self.task_counts[finish2] += 1
+            else:
+                self.task_counts[start] -= 1
+                self.task_counts[finish2] += 1
+
             if self.plugins:
                 # Temporarily put back forgotten key for plugin to retrieve it
                 if ts.state == "forgotten":
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -53,7 +53,7 @@ from distributed.client import (
 from distributed.compatibility import WINDOWS
 
 from distributed.metrics import time
-from distributed.scheduler import Scheduler, KilledWorker
+from distributed.scheduler import Scheduler, KilledWorker, ALL_TASK_STATES
 from distributed.sizeof import sizeof
 from distributed.utils import (
     ignoring,
@@ -3690,6 +3690,16 @@ def test_scheduler_info(c):
     info = c.scheduler_info()
     assert isinstance(info, dict)
     assert len(info["workers"]) == 2
+
+
+def test_scheduler_info_task_counts(c):
+    info = c.scheduler_info()
+    assert info["task_counts"].keys() == ALL_TASK_STATES
+    assert all(i == 0 for i in info["task_counts"].values())
+    futures = c.map(inc, range(7))
+    wait(futures)
+    info = c.scheduler_info()
+    assert info["task_counts"]["memory"] == 7
 
 
 def test_write_scheduler_file(c):


mrocklin:
In principle this seems fine to me.  I'm curious, how long does this add to a single task's lifetime cost?  Currently we aim for (and I think fail to meet) about 200-300us per task.  If we do these dict lookups a few times for each task, how long does that add?

jrbourbeau:
Doing a quick benchmark (by calling `time.time()` before and after the block where we do the dict lookups) adds roughly 3us of overhead per task. I've included the code used for benchmarking below. Happy to try other benchmarking approaches if this isn't representative. 

<details>
<summary>Diff for adding benchmarking to the Scheduler:</summary>

```diff
diff --git a/distributed/scheduler.py b/distributed/scheduler.py
index bd22bb44..2c461487 100644
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -960,6 +960,7 @@ class Scheduler(ServerNode):

         self.n_tasks = 0
         self.task_counts = {state: 0 for state in ALL_TASK_STATES}
+        self.task_livetime_tracker = defaultdict(float)
         self.task_metadata = dict()
         self.datasets = dict()

@@ -4318,6 +4319,7 @@ class Scheduler(ServerNode):
                     ts.state,
                     dict(recommendations),
                 )
+            t_start = time()
             if finish2 == "forgotten":
                 # Not keeping track of forgotten tasks
                 self.task_counts[start] -= 1
@@ -4327,6 +4329,8 @@ class Scheduler(ServerNode):
             else:
                 self.task_counts[start] -= 1
                 self.task_counts[finish2] += 1
+            dt = time() - t_start
+            self.task_livetime_tracker[key] += dt

             if self.plugins:
                 # Temporarily put back forgotten key for plugin to retrieve it
```
</details>

<details>

```python
from time import sleep
from pprint import pprint
import numpy as np

def inc(x):
    return x + 1

if __name__ == '__main__':
    
    with Client() as client:
        futures = client.map(inc, range(100))
        result = client.submit(sum, futures)
        # Wait for computation to finish
        wait(result)
        # Remove futures
        del futures
        del result

        pprint(client.cluster.scheduler.task_livetime_tracker)
        average = np.mean(list(client.cluster.scheduler.task_livetime_tracker.values()))
        print(f'average =  {average}')
```
</details>

TomAugspurger:
FWIW, I'm currently resuccitating our dask-benchmarks repo (it's a bit out
of date).

That benchmark may be a good addition.

jrbourbeau:
Nice! I'll look into adding benchmarks

mrocklin:
So, 3us seems like it's kind of on that boundary of maybe trivial but maybe not.  Some thoughts:

0.  How did this compare to the total time spent in that block?
1.  Can we reduce this at all, perhaps by avoiding dict lookups somehow?
2.  Do we need this?  I agree that it's handy for after-the-fact debugging, but how important is it?  If it's convenient just for a single user or workload then I'm less excited by it.  If it's useful for lots of people, then it seems more exciting.

jrbourbeau:
Thanks for the feedback

> How did this compare to the total time spent in that block?

I should have specified this in my previous comment, the 3us was the total time spent in the new block over the lifetime of a task (i.e. I incremented a running total for each task whenever that block of code was executed). I also benchmarked the total time spent in the `Scheduler.transition` function for a task over it's lifetime to be ~150us. So the additions here increase the time it takes to do all the various transitions throughout a task's lifetime by roughly 2%. 

> Can we reduce this at all, perhaps by avoiding dict lookups somehow?

I can think about alternatives to dict lookups. Nothing is coming immediately to mind.

> Do we need this? I agree that it's handy for after-the-fact debugging, but how important is it? 

That's a great question. I do see this as more as an added convenience rather than enabling new functionality. In particular, I see this as useful when a `Scheduler` object isn't easily accessible to avoid querying the scheduler directly. So I think it's use extends beyond a single workload, but it ultimately is just an additional convenience.

mrocklin:
I was hoping that using a list would be faster than a dict.  It looks like Python is optimized in this case though (at least for ints)

```python
In [1]: L = [0, 0, 0, 0]

In [2]: d = {0: 0, 1: 0, 2: 0, 3:0}

In [3]: %%time
   ...: for i in range(1000000):
   ...:     L[i % 4] += 1
   ...:     L[i % 4] -= 1
   ...:
CPU times: user 217 ms, sys: 1.32 ms, total: 219 ms
Wall time: 219 ms

In [4]: %%time
   ...: for i in range(1000000):
   ...:     d[i % 4] += 1
   ...:     d[i % 4] -= 1
   ...:
CPU times: user 226 ms, sys: 1.28 ms, total: 227 ms
Wall time: 227 ms
```

jrbourbeau:
I'm going to close this PR for the time being. The scheduling overhead this introduces probably isn't worth the added convenience.

PR #3069, opened 2019-09-19
Revive #2740: Allow tasks with restrictions to be stolen

seibert:
I recently stumbled over an issue that is fixed by #2740, but noticed that review on that PR seems to have stalled because a bunch of extraneous commits were accidentally merged, and the original author hasn't opened a new PR or cleaned up the existing one.

This PR attempts to recreate that PR by cherry-picking those commits (so the original author still gets credit), and adds a minor fix.  If there are additional changes needed to finish out code review on this, I'm happy to do those.


seibert:
Ping original author @calebho, and original reviewers @mrocklin, @martindurant, @TomAugspurger.

guillaumeeb:
👍 just to say that this would be wonderfull if this issue could be solve, this has been awaited for several month in our HPC center!

distributed/stealing.py
```diff
@@ -415,4 +429,46 @@ def story(self, *keys):
...
+        and get_address_host(thief.address) not in ts.host_restrictions
+    ):
+        return False
+    elif ts.worker_restrictions and thief.address not in ts.worker_restrictions:
+        return False
+
+    return _has_resources(thief, victim.resources)
+
+
+def _has_resources(ws, required_resources):
```
TomAugspurger: This could likely be inlined in `_can_steal`

seibert: ok, I've inlined the function

calebho:
Thanks for reviving this. I haven't had the bandwidth to continue pushing for this so please feel free to take over from here. Happy to provide review input

distributed/stealing.py
```diff
@@ -351,16 +347,22 @@ def maybe_move_task(level, ts, sat, idl, duration, cost_multiplier):
                             stealable.discard(ts)
                             continue
                         i += 1
-                        if not idle:
```
mrocklin: It might still be nice to have the `if not idle: break` lines, just to avoid these checks when the cluster is under full load.

distributed/stealing.py
```diff
@@ -415,4 +425,42 @@ def story(self, *keys):
...
+        ts.host_restrictions or ts.worker_restrictions or ts.resource_restrictions
+    )
+
+
+def _can_steal(thief, ts, victim):
+    """Determine whether worker ``thief`` can steal task ``ts`` from worker
+    ``victim``.
+    """
+    if not _has_restrictions(ts):
+        return True
```
mrocklin: This check makes sense to have semantically, but it's also unnecessary given where `_can_steal` is called.

distributed/stealing.py
```diff
@@ -415,4 +425,42 @@ def story(self, *keys):
         return out
 
 
+def _has_restrictions(ts):
+    """Determine whether the given task has restrictions and whether these
+    restrictions are strict.
+    """
+    return not ts.loose_restrictions and (
+        ts.host_restrictions or ts.worker_restrictions or ts.resource_restrictions
+    )
```
mrocklin: I wonder if maybe we can cache this on the `TaskState` object itself in order to avoid recomputation.

```python
if ts.restricted:
    ...
```

mrocklin:
In general this seems fine.  I am concerned about performance.  Anecdotally I'll say that work stealing accounts for something like 20-30% of scheduler time when under heavy load.  At some point I went through and micro-optimized things a bit.  It would be good to be mindful of costs here.  

The scheduler is definitely a bottleneck in lots of larger important user workflows (Pangeo cares about this a bunch) so we need to be performance conscious.  I think that we can achieve that fairly easily here though with modest work.

mrocklin:
@seibert checking in.  Is this still something that you're likely to pursue?

seibert:
Yes, I had to switch to doing some conference prep / speaking this week, but I am still interested in solving this issue, as it reduces throughput in the Dask-based testing system we have.  I'll need some guidance how to benchmark the impact of any changes, though.

mrocklin:
>  I'll need some guidance how to benchmark the impact of any changes, though.

Unfortunately we don't have good benchmarks for the Dask scheduler (or any part of Dask really) (although this would be great work for someone in the future).

I've made a few concrete suggestions above that I think might be helpful.  Also, if anything pop's out at you it might be good to bring it up (I trust your performance intuition over pretty much anyone's).

TomAugspurger:
Added a benchmark for the improvement in https://github.com/dask/dask-benchmarks/pull/22. Things look good for the new one. Things are being stolen by the new worker with the resource.

distributed/stealing.py
TomAugspurger: It seems that sometimes these are updated

```python
# in scheduler.py L1707
                    ts = self.tasks[k]
                    ts.loose_restrictions = True
```

If that's the case, I'd prefer to not cache, so that we don't have to worry about invalidating the cache (though perhaps this is the only case?).

mrocklin: That's in `update_graph` which is the time when we construct most tasks. My guess is that that this isn't an issue.  We won't check `ts.restricted` before calling those lines.

mrocklin:
> Things seem OK over in the dask benchmarks repo.

I added a few comments there.  I don't think that that benchmark is likely to be sensitive to the performance impacts here.  I would recommend cranking the stealing interval up super high.

mrocklin:
To be more explicit, lets say that we're aiming for a 200us overhead per task.  How much are we willing to increase that overhead to support stealing restricted tasks?  10us?  20us?  This isn't a very common case, so I think that even those numbers might be fairly high.

This isn't a long term solution, but I encourage people to run a workload locally on your laptop and then look through the `/profile-server` route on the scheduler's dashboard.  You'll find stealing there as taking up 10-15% of the time (or at least this was the case last time I checked, about a year ago).  I recommend diving through this just to get a sense of what things cost today.

mrocklin:
I just did this myself to verify my previous experience.  I was surprised to learn that we seem to be spending almost all of our time sending and receiving things from sockets that are supposed to be non-blocking.  This is probably something that we should look into in the near future.

mberglundmx:
I would really benefit from this PR being merged - unless there is a work-around...?

My workload consists of short running (5-20s) GPU tasks, followed by longer running (2-10 minutes) CPU tasks. I set resource CPU or GPU on my workers to make sure tasks are run on the right hosts.

I launch the GPU tasks (.submit), followed by the CPU tasks. If I e.g. restart (or add) a CPU-worker, that worker will not steal any of the tasks from the other workers.

TomAugspurger:
> unless there is a work-around...?

For now, a workaround is to disable work stealing, and add a worker plugin with the changes from this PR.

But we're hoping to get this merged in the next day or so. Just need to verify a few performance things.

TomAugspurger:
@mrocklin do the benchmarks above sufficiently satisfy your concerns about performance?

mrocklin:
> @mrocklin do the benchmarks above sufficiently satisfy your concerns about performance?

Not entirely.  We're still iterating through every worker for every stealable task.  I think that this will get more interesting when this starts being run in production.  But, I'm at least satisfied that it doesn't seem to be slowing down the common case, and folks seem pretty excited about trying this out.  So let's merge and see what happens.

mrocklin:
Thank you @seibert for doing this work, and to @leej3 and friends for benchmarking things.

mberglundmx:
Thanks a lot everyone for getting this in, will try this release ASAP

PR #3071, opened 2019-09-19
Avoid swamping high-memory workers with data requests

TomAugspurger:
This is a workaround for scheduling issues where worker memory
*temporarily* exceeds the worker memory limit. This is common when the
worker is near it's maximum memory, and then another worker requests a
(large) bit of data. Sending the data to that worker can push the
high-memory worker process past its limit.

Handling this required two changes:

1. Workers under high memory (paused) now reply with "busy" when a
   worker requests data. They'll hopefully complete the tasks they're
   working on, which hopefully lowers memory usage, and then respond
   to the other worker's request for data.
2. We add a new configuration value controlling how long a worker is
   allowed to violate the memory limit. Once they blow past their limit,
   they're given `n` seconds to return to normal memory use.

No tests yet. Will add them if this seems like a reasonable thing.

xref https://github.com/dask/distributed/issues/2602


TomAugspurger:
Here's the test program I'm playing with. It puts a bunch of (too large) bits of data on the cluster, and then does a bunch of (memory freeing!) tasks that require data from each worker.

```python
import dask
import toolz


def gen(i, B):
    return f'{i}'.encode() * B


def f(a):
    return a[0]


if __name__ == "__main__":
    dask.config.set(distributed__worker__memory__terminate_violations='10s')
    client = Client(processes=True, threads_per_worker=1,
                    n_workers=3, memory_limit='2GB')
    B = dask.utils.parse_bytes("50 Mb")
    N = 50
    workers = list(client.scheduler_info()['workers'])

    rdata = [client.submit(gen, i, B, workers=[workers[i % len(workers)]]) for i in range(N)]
    pairs = list(toolz.sliding_window(20, rdata))

    fut = client.submit(f, pairs[0])
    wait(fut)
    print("One is OK")

    results = client.map(f, pairs)
    wait(results)
    print("done")
```

TomAugspurger:
@mrocklin do you have high-level thoughts on this?

It's a workaround, to be sure. Ideally we would have better scheduling so that we don't end up in these situations to begin with, but that's not always possible (and I've struggled with implementing something that does better).

We might also be a bit smarter about how we load data when a worker requests it and we're under memory pressure (reducing number of intermediate copies? Sending in smaller batches?).

mrocklin:
Sorry for the lack of response @TomAugspurger .  Thanks for the ping.

So, how does this problem manifest itself?  Other works hammer this machine for data, it loads a bunch in, and the machine crashes?  If so, then yes, responding with a busy signal makes sense to me.  We already have the busy signal logic written up, yes?

The timeout you have is to avoid a worker waiting forever on data that won't arrive?

I wonder if it makes sense to restrict the number of incoming data requests down to 1 when we're in a paused state, or something similar.

TomAugspurger:
>  Other works hammer this machine for data, it loads a bunch in, and the machine crashes? If so, then yes, responding with a busy signal makes sense to me.

I would think that we handle that case decently well on master with the `distributed.worker.connections.outgoing` configuration value.

This solution is focused more on the case where we're already up against our memory limit because of the tasks that have been assigned to us, and a small number of data requests pushes us over the limit.

> We already have the busy signal logic written up, yes?

Yes, I'm reusing that logic. From the point of view of the worker requesting data, the responses are identical.

> The timeout you have is to avoid a worker waiting forever on data that won't arrive?

Yes, I was somewhat concerned with a deadlock as two saturated workers are waiting on eachother for data and unable to proceed. But the timeout is also necessary for the `{'status': 'busy'}` response to be useful. Without it, I noticed that workers were still getting killed in my test script. I didn't investigate too deeply after getting the timeout working. Perhaps I should spend some more time understanding that case.

> I wonder if it makes sense to restrict the number of incoming data requests down to 1 when we're in a paused state, or something similar.

In theory, we should know the size of the data we need to send, and how much memory we have free. I wonder if that should inform how we handle these requests.

mrocklin:
> Yes, I was somewhat concerned with a deadlock as two saturated workers are waiting on eachother for data and unable to proceed. 

Restricting connections to 1 would resolve this concern, although as you mention below it opens us up to failure under very large single requests

> In theory, we should know the size of the data we need to send, and how much memory we have free. I wonder if that should inform how we handle these requests.

In general we assume that chunks of data are significantly smaller than our total memory.  If we set our pause limit at 80% then we should be safe if users restrict themselves to chunks of data that are less than 10% in size?  If they're using giant chunks then I wonder if we might educate them away from this behavior.

mrocklin:
Any thoughts on the last comment @TomAugspurger ?

TomAugspurger:
Better education is always good. Let me see if I can better understand the
original failure cases.

Say we're paused at 80%, and we try to service ~15-20 roughly simultaneous
requests for data, the we get into
trouble when the data is ~1% of our total size. And that's assuming

1. We don't already have it in memory, so we're loading from disk
2. There's no intermediate copies in dask or tornado

mrocklin:
Right, my proposal would be to only service 1 simultaneous request for data, rather than 0.  This is maybe still safe, and means that we don't have to add logic around a deadlock situation.

TomAugspurger:
> Right, my proposal would be to only service 1 simultaneous request for data, rather than 0

Sounds reasonable. Will update that next week.

distributed/worker.py
```diff
@@ -1163,10 +1163,24 @@ def send_to_worker(self, address, msg):
...
             and self.outgoing_current_count > max_connections
         ):
+            logger.warning(
+                "Worker %s has too many open connections to respond to data request from %s (%d/%d).%s",
+                self.address,
+                who,
+                self.outgoing_current_count,
+                max_connections,
+                throttle_msg,
+            )
```
mrocklin: This is the sort of thing that might happen *very* frequently and fill up logs.  I wish there was a way to throttle logging itself.

It might make sense to move this to debug level.

distributed/tests/test_nanny.py
```diff
@@ -290,6 +290,58 @@ def leak():
...
+        return b"0" * size
+
+    a = workers[0]
+    proc = a.process.pid
+    from distributed import wait
+
+    with captured_logger(logging.getLogger("distributed.worker")) as logger:
+        size = int(0.05 * a.memory_limit)
+        futures = [
+            c.submit(data, size, i, workers=[a.worker_address]) for i in range(12)
```
mrocklin: Not necessary to change, but it looks like you might have wanted the `pure=False` keyword here, rather than pass in `i`

distributed/tests/test_nanny.py
```diff
@@ -290,6 +290,58 @@ def leak():
...
+    timeout=20,
+    clean_kwargs={"threads": False},
+)
+def test_nanny_throttle(c, s, *workers):
+    def data(size, i):
+        return b"0" * size
+
+    a = workers[0]
+    proc = a.process.pid
+    from distributed import wait
```
mrocklin: Feel free to include this in the top level imports.

distributed/tests/test_nanny.py
```diff
@@ -290,6 +290,58 @@ def leak():
...
 
+@gen_cluster(
+    nthreads=[("127.0.0.1", 1)] * 8,
+    client=True,
+    Worker=Nanny,
+    worker_kwargs={"memory_limit": 2e8, "memory_pause_fraction": 0.6},
+    timeout=20,
+    clean_kwargs={"threads": False},
+)
+def test_nanny_throttle(c, s, *workers):
```
mrocklin: You might enjoy using `async def` here, and replacing `yield`s with `await`s.  It's not important, but I personally find it satisfying for some reason :)

distributed/tests/test_nanny.py
```diff
@@ -290,6 +290,58 @@ def leak():
...
+        yield wait(futures)
+
+        def check(dask_worker):
+            self = dask_worker
+            proc = self.monitor.proc
+            memory = proc.memory_info().rss
+            frac = memory / self.memory_limit
+            return frac, self.paused, self.memory_pause_fraction
+
+        start = time()
```
TomAugspurger: Is there a helper for this type of thing? "Keep repeating this check until it passes, else raise when you hit the timeout?"

TomAugspurger:
I'm having some trouble making the test reliable :/ This commit failed then passed on Travis.

distributed/tests/test_nanny.py
```diff
@@ -290,6 +290,58 @@ def leak():
...
+        memory = proc.memory_info().rss
+        frac = memory / self.memory_limit
+        return frac, self.paused, self.memory_pause_fraction
+
+    start = time()
+    while True:
+        paused = await c.run(check)
+        if not paused[a.worker_address][1]:
+            assert time() - start < 5, paused[a.worker_address]
+            futures.append(c.submit(data, size, workers=[a.worker_address], pure=False))
```
mrocklin: You might want to `await wait(future)` here too to avoid things running away.

distributed/tests/test_nanny.py
mrocklin: I think that there is, but I tend to just use a `while True` loop because I understand it more.

TomAugspurger:
Not finding a traceback right now.

One of them was a timeout while waiting for the worker to become paused, which led to adding the `futures.apppend(...)` inside the `while True` loop checking if they're paused.

Another failure seemed to only occur when `test_workspace_concurrency` also failed.

mrocklin:
> You might want to await wait(future) here too to avoid things running away.

`test_workspace_concurrency` seems to be failing intermittently again.  I'm not sure what changed.  We recently xfailed it though in https://github.com/dask/distributed/pull/3132

TomAugspurger:
Ah, good to know.

https://travis-ci.org/dask/distributed/jobs/596216678#L2762 failed, but that it perhaps a genuine failure, if the throttling didn't kick in... I'm wondering if a worker can become unpaused before we get to the task requesting data from that worker. I'll keep looking.

TomAugspurger:
@mrocklin I think this is good to go now.

I rewrote the test to just directly set `self.paused` on the one worker. It's a bit brittle, since I need to also set the `_memory_monitoring` value (a kind of check for whether we're currently doing stuff inside `memory_monitor`) otherwise we unpause the worker on the next memory-monitor interval. But it seems much less flaky now.

TomAugspurger:
Not sure what happened in  https://ci.appveyor.com/project/daskdev/distributed/builds/28099415


<details>

```
_____________________________ test_nanny_throttle _____________________________
2120Traceback (most recent call last):
2121  File "C:\projects\distributed\distributed\node.py", line 173, in wait_for
2122    await asyncio.wait_for(future, timeout=timeout)
2123  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\tasks.py", line 410, in wait_for
2124    raise futures.TimeoutError()
2125concurrent.futures._base.TimeoutError
2126
2127During handling of the above exception, another exception occurred:
2128
2129Traceback (most recent call last):
2130  File "C:\projects\distributed\distributed\utils_test.py", line 892, in coro
2131    worker_kwargs=worker_kwargs,
2132  File "C:\projects\distributed\distributed\utils_test.py", line 810, in start_cluster
2133    await asyncio.gather(*workers)
2134  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\futures.py", line 381, in __iter__
2135    yield self  # This tells Task to wait for completion.
2136  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\tasks.py", line 310, in _wakeup
2137    future.result()
2138  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\futures.py", line 294, in result
2139    raise self._exception
2140  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\tasks.py", line 240, in _step
2141    result = coro.send(None)
2142  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\tasks.py", line 585, in _wrap_awaitable
2143    return (yield from awaitable.__await__())
2144  File "C:\projects\distributed\distributed\node.py", line 178, in wait_for
2145    type(self).__name__, timeout
2146tornado.util.TimeoutError: Nanny failed to start in 5 seconds
2147
2148During handling of the above exception, another exception occurred:
2149
2150Traceback (most recent call last):
2151  File "C:\projects\distributed\distributed\utils_test.py", line 952, in test_func
2152    coro, timeout=timeout * 2 if timeout else timeout
2153  File "c:\miniconda35-x64\envs\testenv\lib\site-packages\tornado\ioloop.py", line 576, in run_sync
2154    return future_cell[0].result()
2155  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\futures.py", line 294, in result
2156    raise self._exception
2157  File "c:\miniconda35-x64\envs\testenv\lib\asyncio\tasks.py", line 242, in _step
2158    result = coro.throw(exc)
2159  File "C:\projects\distributed\distributed\utils_test.py", line 904, in coro
2160    raise Exception("Could not start cluster")
2161Exception: Could not start cluster
2162
```

</details>

Unclear if https://travis-ci.org/dask/distributed/jobs/597730582 is releated `test_open_close_many_workers[Worker-100-5]` failed.

mrocklin:
The test appears to fail on appveyor.  Also it seems to be taking 30s

mrocklin:
is it possible that you have a 5s death timeout set?

TomAugspurger:
I see `test_open_close_many_workers` failed on https://travis-ci.org/dask/distributed/jobs/597733157.

> Also it seems to be taking 30s

> is it possible that you have a 5s death timeout set?

I'm not sure, I don't specify a custom one. I do create 8 nannies / workers right now. I think that can be lowered.

mrocklin:
test_open_close_many_workers appears to also be intermittently failing on master.  See #3148 .  Give me a day to see if I can resolve it.

TomAugspurger:
Sounds good, no rush though. My test still isn't 100% reliable unfortunately. I was able to get a failure locally after many retries.

distributed/worker.py
```diff
@@ -1156,10 +1156,24 @@ def send_to_worker(self, address, msg):
...
+        if self.paused:
+            max_connections = 1
+            throttle_msg = " Throttling outgoing connections because worker is paused."
+        else:
+            throttle_msg = ""
+
         if (
             max_connections is not False
-            and self.outgoing_current_count > max_connections
+            and self.outgoing_current_count >= max_connections
```
TomAugspurger: I think this should be a `>=`.

We don't increment `outgoing_current_count` until a few lines down (L1179 in my version).

Suppose, we have `max_connections=5` and were already at that (`self.outgoing_current_count=5`), and we have a new request. To service that, we'd need a 6th connection, so we want to reply busy (i.e. `>=`).

@mrocklin can you double check my logic here?

mrocklin: Sure.  That makes sense to me

TomAugspurger:
@mrocklin thoughts on how to proceed here? I think this PR and is in a good state (Not that I'm volunteering to track down that other failure (unless you actually want me to look at it, in which case I'm happy to find some time today 😄)).

TomAugspurger:
@mrocklin think we're OK to merge this before the release?

mrocklin:
Ah, sorry to go dark before.  It seems sensible.  I'm happy if you're happy.

TomAugspurger:
Thanks.

PR #3077, opened 2019-09-20
Respect Cluster.dashboard_link in Client._repr_html_ if it exists

mrocklin:
cc @jcrist

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -824,10 +824,10 @@ class Client(Node):
 
                 address = format_dashboard_link(host, port)
 
-        text += (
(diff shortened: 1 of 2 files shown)


jcrist:
LGTM, thanks.

distributed/client.py
```diff
@@ -811,17 +811,23 @@ def _repr_html_(self):
...
+
+                port = info["services"]["dashboard"]
+                if protocol == "inproc":
+                    host = "localhost"
+                else:
+                    host = rest.split(":")[0]
+
+                address = format_dashboard_link(host, port)
+
+        text += (
```
jcrist: Looks like you need to move this under the `if` statement as well, otherwise `address` is undefined.

PR #3078, opened 2019-09-20
Checks for command parameters in ssh2

pentschev:
Check for command's parameters in ssh2, as well as the class. A custom worker, such as `dask-cuda-worker` may contain parameters that don't exist in the `Worker` class, thus we must check if they are valid.

cc @mrocklin


distributed/utils.py
```diff
@@ -1464,7 +1464,7 @@ def _repr_html_(self):
         return "\n".join(summaries)
 
 
-def cli_keywords(d: dict, cls=None):
+def cli_keywords(d: dict, cls=None, cmd=None):
```
mrocklin: I'm not sure I agree with the changes here.  This function used to do a very simple thing, now it does that thing and a more complex thing that I think was tacked on just for this one use case.

Can you say a bit more about what you're trying to accomplish, and then maybe we can find another approach.

mrocklin: Oh, I think that you're trying to change the worker_module string, like `distributed.cli.dask_worker` to an actual module?

I recommend that we make a separate convenience function for this other than `cli_keywords`, and then maybe we can use it in the `ssh2.py::Worker` class instead?  Something like that might be a cleaner separation.

Also, is it easy to add a test for this functionality?  That might make it easier for me to see what you're trying to do here.

pentschev: I'm not sure I understand your comments. But a short explanation is that `dask_cuda_worker` is a specialization of `Worker`, it takes other parameters that `Worker` doesn't know about (such as `--cpu-affinity`), but only the `dask-cuda-worker` executable knows of. I'm not sure I can find a self-contained example that doesn't depend on `dask-cuda` at the moment, but I'll look for one.

pentschev: Ok, I extended the existing tests. To clarify `nprocs` is an option from the `cli/dask_worker.py`, so `SSHCluster` would have failed to start with that option before. Also added a test to ensure unimplemented options raise an exception.

pentschev:
I'm reviving this PR. Just to be clear what it allows is to pass something like the following to `SSHCluster`:

```python
worker_options={"enable_nvlink": ""}
```

This wasn't possible before, since this parameter is part of the `dask-cuda-worker` command, and not the `Worker` class.

Besides that, I've moved most of the new complexity into a new `command_has_keyword` function. @mrocklin I hope that is a better way to address this case.

pentschev:
@mrocklin could you take a look at this when you have the time?

distributed/deploy/tests/test_ssh.py
```diff
@@ -32,7 +32,12 @@
         ["127.0.0.1"] * 3,
         connect_options=dict(known_hosts=None),
         asynchronous=True,
-        worker_options={"nthreads": 2, "memory_limit": "2 GiB", "death_timeout": "5s"},
+        worker_options={
+            "nprocs": 2,
```
jacobtomlinson: Perhaps an inline comment here to explain that `nprocs` is an option from `cli/dask_worker.py` and why that is significant would be helpful. It is a rather subtle change. Perhaps an assertion below that uses it would also help.

distributed/utils.py
```diff
@@ -1331,6 +1358,9 @@ def cli_keywords(d: dict, cls=None):
         The keywords to convert
     cls: callable
         The callable that consumes these terms to check them for validity
+    cmd: string or object
+        A string with the name of a module, or the module containing a
+        click-generated command with a "main" function, or the function itself
```
jacobtomlinson: This explains that the `cmd` option takes but not what it is for. Perhaps some reference that this is the python module to run to start the worker would be helpful.

mrocklin:
> Perhaps @mrocklin could give this another look?

Sorry for the long delay in responding here gents.  I'm going to pass on this one.  If you two both agree that this is a good change then I trust your collective judgement.

pentschev:
Thanks @jacobtomlinson for the review, I added some better documentation/commenting now, could you check if you think that's clearer now? Thanks @mrocklin as well for the response.

pentschev:
Thanks @jacobtomlinson for the review and merging!

PR #3079, opened 2019-09-20
Have Client get Security from passed Cluster

mrocklin:


distributed/client.py
```diff
@@ -658,6 +653,13 @@ def __init__(
             self.cluster = address
             with ignoring(AttributeError):
                 loop = address.loop
+            if security is None:
+                self.security = self.cluster.security
```
jcrist: We should probably handle security things all together. This is currently split between here and line 640 above.

PR #3085, opened 2019-09-25
set task duration based on current duration

danpf:
not 100% sure if i have the details correct, but as i understand it:
if you have a task that takes 1hr+ currently the scheduler adapts based on occupancy which when set can be boiled down to essentially (but not exactly)

task_duration = 0.5 seconds (until 1 task finishes)
target_duration = 5 seconds (set at the beginning of `.adapt`)
target_num_cpus = (task_duration*number_of_tasks) / target_duration

so for 200 tasks target_num_cpus == (0.5 * 200) / 5 == 20

so when adapting with jobs that are 1hr+, the scheduler will wait until one of those jobs finishes, and 
THEN it will scale up because it says 'oh now I know the duration of this function prefix, it's really long, lets get more workers'

if we track how much time a TaskState is alive since it started processing, we can then make a better guess of how long a task will take, and thus-adapt better.

however... I'm not 100% sure `"processing"` actually means the task has started, or if it's just on the worker waiting to be run.  If `"processing"`  means 'ready to run' then I should probably think of an alternative method.  does anyone have any insight on that?

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -23,6 +23,7 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []
     preload-argv: []
+    default-task-durations: {}  # How long we expect function names to run ("1h", "1s") (helps for long tasks)
     dashboard:
       status:
         task-stream-length: 1000
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -570,7 +570,6 @@ class TaskState(object):
         "loose_restrictions",
         # === Task state ===
         "state",
-        "time_started",
         # Whether some dependencies were forgotten
         "has_lost_dependencies",
         # If in 'waiting' state, which tasks need to complete
@@ -599,7 +598,6 @@ class TaskState(object):
         self.prefix = key_split(key)
         self.run_spec = run_spec
         self.state = None
-        self.time_started = None
         self.exception = self.traceback = self.exception_blame = None
         self.suspicious = self.retries = 0
         self.nbytes = None
@@ -3676,7 +3674,6 @@ class Scheduler(ServerNode):
             ws.occupancy += duration + comm
             self.total_occupancy += duration + comm
             ts.state = "processing"
-            ts.time_started = time()
             self.consume_resources(ts, ws)
             self.check_idle_saturated(ws)
             self.n_tasks += 1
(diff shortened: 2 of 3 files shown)


mrocklin:
> however... I'm not 100% sure "processing" actually means the task has started, or if it's just on the worker waiting to be run. If "processing" means 'ready to run' then I should probably think of an alternative method. does anyone have any insight on that?

"processing" just means "ready to run" or "all dependencies have been met, so the scheduler has shipped this off to a worker"

danpf:
would you be opposed to having the worker contact the scheduler once a task moves to  "executing"?  it doesn't seem like it would add too much complexity.

mrocklin:
We're pretty sensitive to per-task costs on the scheduler.  We have a fairly small time budget (200-300us) to spend on each task.  Sending and receiving an additional message is non-trivially expensive.  So yes, we could do this, but only if there was enough motivation.   My current feeling is that, while the issue you raise is important, it might not be important enough to justify this cost.

danpf:
Would you be more open to an `expected_duration` kwarg added to `submit`/`map`? ( I haven't investigated if that's even possible yet, I'm just trying to think of a less costly an alternative )

mrocklin:
> Would you be more open to an expected_duration kwarg added to submit/map? ( I haven't investigated if that's even possible yet, I'm just trying to think of a less costly an alternative )

Can I counter-propose that we have a configuration value for known times?  We already do something like this here:

https://github.com/dask/distributed/blob/dec8abead0c460fece9b9d15f216bee5ca2e0d11/distributed/scheduler.py#L965-L966

If this was, instead, a configuration value then that would be, from my perspective, relatively cheap to add

```python
self.task_duration = dask.config.get("distributed.scheduler.task-durations")
self.task_duration = {k: parse_timedelta(v) for k, v in self.task_duration.items()}
```

```yaml
distributed:
  scheduler:
    task-durations:
      my-long-task: 1hr
      my-short-task: 1ms
```

mrocklin:
Thanks @danpf .  Can I ask you for a small test?  Maybe something like the following in `distributed/tests/test_scheduler.py`?

```python
def test_foo(cleanup):
    with dask.config.set({...}):
        s = Scheduler(port=0)
        assert s.task_durations["..."] == ...
```

mrocklin:
Thanks for making these changes @danpf .

Merging this in!

danpf:
thanks for helping me with this Matt :tada:

PR #3086, opened 2019-09-25
give option for longer errors tracebacks

danpf:
I often am forced to run long 'subprocess' commands, and have for a long time now just hacked the `error_message` function to not truncate at all.  I figured it might be useful for others to adjust this in case your ridiculously long subprocess command fails and you want to know exactly what failed. 

in my case -- nothing is worse than getting a traceback back that is missing half of the error message that I need to help reproduce the error.

I personally think this is really useful, but I understand if you do not think it is worth adding.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -969,7 +969,7 @@ def error_message(e, status="error"):
     --------
     clean_exception: deserialize and unpack message into exception/traceback
     """
-    MAX_ERROR_LEN = dask.config.get("distributed.admin.max_error_len")
+    MAX_ERROR_LEN = dask.config.get("distributed.admin.max-error-length")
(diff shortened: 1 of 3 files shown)


mrocklin:
Thanks @danpf 

In principle this looks fine to me.  I recommend that we slightly rename `max_error_len` to `max-error-length`.  We tend to use hyphens rather than underscores by convention, and it seems friendly to expand `len` to `length`.

mrocklin:
Also, please note the linter errors.  You may want to look through https://docs.dask.org/en/latest/develop.html#code-formatting

mrocklin:
Thanks @danpf !  Merging this in.

Issue #1683, opened 2018-01-10
'import distributed' causes eating of bokeh server exceptions


timothydmorton:
Not sure if this issue is better for here or for bokeh (pinging @bryevdv), but the simplest bokeh app I can think of does not print exceptions to the command line if I import distributed.  The app looks like 
```python
from bokeh.io import curdoc
import distributed

def modify_doc(doc):
    raise ValueError
    return doc


doc = modify_doc(curdoc()) 
```
When you run this with `bokeh serve <filename>` and try to access the app, the exception does not get printed to the command line (as it should, and as it does when the `distributed` import does not occur).

mrocklin:
We definitely screw around with logging.  If we do this on import then I
agree that this is a problem.

timothydmorton:
Yes, this happens on import.  I've just been poking around to see if I can figure out where, but no luck yet.

timothydmorton:
OK, found it: https://github.com/dask/distributed/blob/master/distributed/config.py#L165

mrocklin:
So what is the right thing to do here?  Call initialize logging only when starting a dask process?  Find some other way to avoid mucking about with Bokeh logs?

EtsuNDmA:
Hi, this issue is still actual

philippjfr:
This is still a major issue and makes debugging bokeh apps which use distributed or any library that imports it (e.g. xarray) really difficult. I'll take a look at fixing this.

mrocklin:
Thank you @philippjfr !

philippjfr:
Not entirely sure how to proceed here, so let me clarify a few things. What's the reasoning behind raising the bokeh log levels to critical?

One simple suggestion I have is that we simply solve this by raising the log level of code execution errors to critical here: https://github.com/bokeh/bokeh/blob/master/bokeh/application/application.py#L180. That makes sense to me anyway since an error when running the bokeh app code is indeed a critical error.

That said I think raising log levels of some external library on import is not a good thing to do in general.  I get that a user may not want to see low-level errors for the dask monitoring dashboard but does the log-level really need to be set that high, would ERROR level logging be okay? Could we at least defer raising the bokeh log level until the dask monitoring dashboard is started?

mrocklin:
As long as the user experience is pleasant I don't personally have strong thoughts on what should happen.  There are two common usage patterns:

Naive use from Python.  In this case we probably want no logging:

```python
from dask.distributed import Client

client = Client()
```

[...]

PR #3087, opened 2019-09-25
Lowered default bokeh log level

philippjfr:
This PR lowers the default bokeh log level to ERROR from CRITICAL. This ensures that simply importing distributed does not hide relevant log output for users who merely import distributed or some other library that imports distributed (e.g. xarray, datashader etc.). Additionally this respects the `silence_logs` argument to ``SpecCluster`` ensuring that when starting a cluster interactively the logging is still suppressed.

Fixes https://github.com/dask/distributed/issues/1683

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -212,7 +212,7 @@ class SpecCluster(Cluster):
         if silence_logs:
             self._old_logging_level = silence_logging(level=silence_logs)
             self._old_bokeh_logging_level = silence_logging(
-                level=logging.CRITICAL, root='bokeh'
+                level=silence_logs, root='bokeh'
             )
 
         self._loop_runner = LoopRunner(loop=loop, asynchronous=asynchronous)
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -212,7 +212,7 @@ class SpecCluster(Cluster):
         if silence_logs:
             self._old_logging_level = silence_logging(level=silence_logs)
             self._old_bokeh_logging_level = silence_logging(
-                level=silence_logs, root='bokeh'
+                level=silence_logs, root="bokeh"
             )
 
         self._loop_runner = LoopRunner(loop=loop, asynchronous=asynchronous)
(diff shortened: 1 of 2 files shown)


distributed/deploy/spec.py
```diff
@@ -211,6 +211,9 @@ def __init__(
 
         if silence_logs:
             self._old_logging_level = silence_logging(level=silence_logs)
+            self._old_bokeh_logging_level = silence_logging(
+                level=logging.CRITICAL, root='bokeh'
```
mrocklin: The `silence_logs` value here can be a logging level from the user.  We should probably respect what they're asking for.

```suggestion
                level=silence_logs, root='bokeh'
```

mrocklin:
In principle this seems fine to me.  I made one small nitpicky comment.

distributed/deploy/spec.py
philippjfr: Right, I did consider that. The problem is that this probably depends on the package, e.g. it seems the default value is INFO, which would be pretty verbose for bokeh and certainly a big difference to CRITICAL.

philippjfr: So for my purposes I'd actually be pretty happy with that but it doesn't really jive with your desire to have interactive Python usage be pretty much silent. Maybe I'm missing something though and the default level is actually higher than INFO.

mrocklin: The default level for LocalCluster (the 99% usecase) is WARN

https://github.com/dask/distributed/blob/4953f81d7b6856f6a22f00e27476d4a554a120c2/distributed/deploy/local.py#L96

I can live with what's here though.  I'm happy to defer the deicsion on what's best here to you.

philippjfr: Okay, I think making it configurable is definitely better than hardcoding it so I'll apply your suggestion.

mrocklin:
Thanks @philippjfr .  If it's ok I'm going to hold onto this PR and merge it after the upcoming release (today or tomorrow).  That way we'll have a bit of time with this change in early-adopters' hands before releasing it.  We do releases every couple of weeks, so this should still be out soonish.

philippjfr:
That seems like a good idea, thanks!

mrocklin:
This is in @philippjfr .  Thanks!

PR #3088, opened 2019-09-25
Identify lost workers in SpecCluster based on address not name

### LINKED ISSUE #3062: client.restart() crashes client
dask version: 2.4.0
python version: 3.7.3 (packaged by conda-forge)

Steps to reproduce:
```python
import dask
client = Client()  
# open dashboard http://localhost:8787 and everything looks fine

x = client.submit(lambda x: x+1, 10)
x.result()  # works

client.restart()
x = client.submit(lambda x: x+1, 10)
# dashboard still looks fine

x.result()  # might work or freeze, keep repeating until it freezes
# dashboard is empty (no workers and no bytes stored)
```

Traceback:

```bash
  File "<stdin>", line 1, in <module>
  File "/home/.../python3.7/site-packages/distributed/client.py", line 219, in result
    result = self.client.sync(self._result, callback_timeout=timeout, raiseit=False)
  File "/home/.../python3.7/site-packages/distributed/client.py", line 754, in sync
    self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
  File "/home/.../python3.7/site-packages/distributed/utils.py", line 334, in sync
    e.wait(10)
  File "/home/.../python3.7/threading.py", line 552, in wait
  File "/home/.../python3.7/threading.py", line 300, in wait
KeyboardInterrupt
```

jrbourbeau:
This PR slightly modifies the logic for removing lost workers from `SpecCluster.workers` introduced in #2990 to identify workers by their address, not their `name`.

When nanny's restart workers, the restarted worker has the same name but a different address (please correct me if this is not always the case!). For example we might initially have a worker with `(name, address) = ('alice', 'tcp://127.0.0.1:53081')`, and then after the worker process is restarted by the nanny it may have `(name, address) = ('alice', 'tcp://127.0.0.1:53088')`.

So after restarting, the current

https://github.com/dask/distributed/blob/4953f81d7b6856f6a22f00e27476d4a554a120c2/distributed/deploy/spec.py#L309

logic will find the restarted worker by it's name (same as the original, pre-restarting worker) and close the restarted worker. 

The changes proposed here replace the named-based worker lookup with address-based worker lookup. Although, I'm definitely open to other suggestions. 

Note, I also updated `self.scheduler_info` -> `self.scheduler_info["workers"]` as `msg` is a worker address, which are the keys in `self.scheduler_info["workers"]`.

Fixes #3062

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -307,8 +307,12 @@ class SpecCluster(Cluster):
 
             def f():
                 if (
-                    msg in [w.worker_address for w in self.workers.values()]
+                    name in self.workers
                     and msg not in self.scheduler_info["workers"]
+                    and not any(
+                        d["name"] == name
+                        for d in self.scheduler_info["workers"].values()
+                    )
                 ):
                     self._futures.add(asyncio.ensure_future(self.workers[name].close()))
                     del self.workers[name]
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -306,7 +306,10 @@ class SpecCluster(Cluster):
             name = self.scheduler_info["workers"][msg]["name"]
 
             def f():
-                if name in self.workers and msg not in self.scheduler_info:
+                if (
+                    msg in [w.worker_address for w in self.workers.values()]
+                    and msg not in self.scheduler_info["workers"]
+                ):
                     self._futures.add(asyncio.ensure_future(self.workers[name].close()))
                     del self.workers[name]
 
--- a/distributed/deploy/tests/test_spec_cluster.py
+++ b/distributed/deploy/tests/test_spec_cluster.py
@@ -160,7 +160,7 @@ async def test_unexpected_closed_worker(cleanup):
 @pytest.mark.asyncio
 async def test_restart(cleanup):
     # Regression test for https://github.com/dask/distributed/issues/3062
-    worker = {"cls": Worker, "options": {"nthreads": 1}}
+    worker = {"cls": Nanny, "options": {"nthreads": 1}}
     with dask.config.set({"distributed.deploy.lost-worker-timeout": "2s"}):
         async with SpecCluster(
             asynchronous=True, scheduler=scheduler, worker=worker
--- a/distributed/deploy/tests/test_spec_cluster.py
+++ b/distributed/deploy/tests/test_spec_cluster.py
@@ -156,6 +156,26 @@ async def test_unexpected_closed_worker(cleanup):
             assert len(cluster.workers) == 2
 
 
+@pytest.mark.slow
+@pytest.mark.asyncio
+async def test_restart(cleanup):
+    # Regression test for https://github.com/dask/distributed/issues/3062
+    worker = {"cls": Worker, "options": {"nthreads": 1}}
+    with dask.config.set({"distributed.deploy.lost-worker-timeout": "2s"}):
+        async with SpecCluster(
+            asynchronous=True, scheduler=scheduler, worker=worker
+        ) as cluster:
+            async with Client(cluster, asynchronous=True) as client:
+                cluster.scale(2)
+                await cluster
+                assert len(cluster.workers) == 2
+
+                await client.restart()
+                await asyncio.sleep(3)
+
+                assert len(cluster.workers) == 2
+
+
 @pytest.mark.asyncio
 async def test_broken_worker():
     with pytest.raises(Exception) as info:


distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -156,6 +156,26 @@ class MyCluster(SpecCluster):
...
+        async with SpecCluster(
+            asynchronous=True, scheduler=scheduler, worker=worker
+        ) as cluster:
+            async with Client(cluster, asynchronous=True) as client:
+                cluster.scale(2)
+                await cluster
+                assert len(cluster.workers) == 2
+
+                await client.restart()
+                await asyncio.sleep(3)
```
mrocklin: Do we need this sleep here?  I've found that fixed sleeps are bad in our tests for two reasons:

1.  They have to be *very* conservative on travis-ci, otherwise we get intermittent failures
2.  If they're very conservative then they take a long time

Can you say more about the reason for this sleep?  Perhaps there is some other thing that we could wait on or poll to make this more precise.

distributed/deploy/spec.py
```diff
@@ -306,7 +306,10 @@ def _update_worker_status(self, op, msg):
             name = self.scheduler_info["workers"][msg]["name"]
 
             def f():
-                if name in self.workers and msg not in self.scheduler_info:
+                if (
+                    msg in [w.worker_address for w in self.workers.values()]
```
mrocklin: The `Worker` and `Nanny` classes have the `worker_address` attribute, but it's not guaranteed to be present.  dask-jobqueue `Job` objects don't know the address of the worker at all for example.

distributed/deploy/spec.py
```diff
@@ -306,7 +306,10 @@ def _update_worker_status(self, op, msg):
...
+                    and msg not in self.scheduler_info["workers"]
```
mrocklin: Thanks for the fix here

distributed/deploy/tests/test_spec_cluster.py
mrocklin: Thanks for marking it slow.  I suppose that if we need the sleep then that is a nice compromise.

distributed/deploy/spec.py
mrocklin: I wonder if it makes sense to close the Worker object here if no other worker has shown up with the same name?  

This seems error prone though, because we're likely to get the "removed" message before the new worker has a chance to spin up.

distributed/deploy/tests/test_spec_cluster.py
jrbourbeau: The main thin we want to check here is that, after workers have been restarted, the delayed call  to `f` doesn't remove the restarted workers. My thinking was to set `distributed.deploy.lost-worker-timeout` sufficiently large such that the delayed call to `f` happens after workers are restarted. The sleep was added so we know the `f` call has already happened. 

Now that I've typed my reasoning out, it does seem a little fragile... I'll think a little more about if there's something else more direct we could wait on

distributed/deploy/spec.py
mrocklin: Thoughts on this @jrbourbeau ?

mrocklin: Maybe something like the following?

```python
if (name in self.workers and 
    msg not in self.scheduler_info["workers"] and
    not any(d["name"] == name for d in self.scheduler_info["workers"].values()
):
```

jrbourbeau: This works if the new worker has already spun up. If the new worker has not spun up yet then this will delete the old worker (in line with https://github.com/dask/distributed/pull/2990#issuecomment-527218288), but the new workers then don't show up in `self.workers` (still not sure why yet). The new workers exist and are functional (e.g. you can compute things on them), but aren't in `self.workers`. 

```python
from time import sleep
import dask


def inc(x):
    return x + 1


if __name__ == "__main__":

    with dask.config.set({"distributed.deploy.lost-worker-timeout": "1ms"}):
        with Client(n_workers=1) as client:
            print(f"client.cluster.workers = {client.cluster.workers}")
            client.restart()
            for _ in range(5):
                print(f"client.cluster.workers = {client.cluster.workers}")
                sleep(0.5)

            wait(future)
            print(f"future.result() = {future.result()}")
            print(f"client.who_has() = {client.who_has()}")
            print(f"client.cluster.workers = {client.cluster.workers}")

```

mrocklin: > If the new worker has not spun up yet then this will delete the old worker 

I think that this is what the current delay is for?  We wait a modest time to allow the worker to come back.  If it does, then we don't delete it.

> but the new workers then don't show up in self.workers

From the cluster manager's perspective, there is no new worker.  In the `Client()` case the objects in `self.worker` are `Nanny` objects, which aren't dying or coming back.  They've been the same the entire time.

jrbourbeau: I guess I'm talking about when we wait a very small amount of time for the worker to come back (e.g. 1ms). So we're checking if the new worker has appeared before the `"add"` message (from `restart`-ing the workers) has been received by the cluster manager. 

In this case, the new worker hasn't appeared yet, so we remove the `Nanny` object from `self.workers`. But later the `"add"` message is received and the new worker appears. However the corresponding `Nanny` is not added back to `self.workers`.

Does it make sense to set a minimum allowed value for `lost-worker-timeout`?

mrocklin: > Does it make sense to set a minimum allowed value for lost-worker-timeout?

Sorry for going dark here.

I don't think so.  I think that our default should definitely be large enough so that a worker is likely to restart in this time (I think that it is).  But that if a user wants to claim that their workers restart faster, then we should let them, even if they're wrong.  In practice I doubt that many will modify this configuration value.

mrocklin: So without thinking too hard about this I think that we should do this

```python
if (name in self.workers and 
    msg not in self.scheduler_info["workers"] and
    not any(d["name"] == name for d in self.scheduler_info["workers"].values()
):
```

I think that your concern here 

> This works if the new worker has already spun up. If the new worker has not spun up yet then this will delete the old worker

Is accurate, but that we need to assume that the delay will handle this.

There are probably also other solutions here, but I think that this solution is decent, and improves things.

jrbourbeau: I agree with your points about letting users claim their workers start quickly and also that the fix here is net progress. I've updated the logic in `_update_worker_status` accordingly

mrocklin:
Merging.  Thanks @jrbourbeau !

jrbourbeau:
Thanks for all the feedback @mrocklin!

PR #3090, opened 2019-09-26
Dask spec worker

mrocklin:
This is an experimental CLI that takes in a fully configured worker spec, rather than command line arguments.  This might be helpful for systems like SSH, that are trying to route Python arguments through the CLI.

## Example

```yaml
# foo.yaml
big:
  cls: dask.distributed.Worker
  opts:
    nanny: false,
    nthreads: 12
    resources: {"FOO": 1}
small:
  cls: dask.distributed.Worker
  opts:
    nanny: false,
    nthreads: 2
```

```bash
$ dask-scheduler
distributed.scheduler - INFO -   Scheduler at:   tcp://192.168.1.88:8786

$ dask-spec-worker localhost:8786 --file foo.yaml
distributed.worker - INFO -       Start worker at:      tcp://127.0.0.1:63400
distributed.worker - INFO -          Listening to:      tcp://127.0.0.1:63400
distributed.worker - INFO -               Threads:                         12
distributed.worker - INFO -                Memory:                   17.18 GB
distributed.worker - INFO -       Local Directory: /Users/mrocklin/workspace/distributed/dask-worker-space/worker-r762e6xl
distributed.worker - INFO -       Start worker at:      tcp://127.0.0.1:63401
distributed.worker - INFO -          Listening to:      tcp://127.0.0.1:63401
distributed.worker - INFO -               Threads:                          2
distributed.worker - INFO -                Memory:                    2.86 GB
distributed.worker - INFO -       Local Directory: /Users/mrocklin/workspace/distributed/dask-worker-space/worker-h8mjhnef
distributed.worker - INFO -         Registered to:       tcp://localhost:8786
distributed.worker - INFO -         Registered to:       tcp://localhost:8786
```

I'm not sure that this is worth having, but I've sometimes wanted something like this.  It was simple to make, so I thought I'd put it up here and ask for comments.

cc @jcrist @pentschev @jacobtomlinson @guillaumeeb 

Edit: As another example, here is starting a scheduler with two protocols 

```
dask-spec --spec '{"cls": "dask.distributed.Scheduler", "opts": {"protocol": ["inproc", "tcp"]}}'
```


jacobtomlinson:
Are some files missing from this PR?

mrocklin:
Indeed there are :)

distributed/cli/dask_spec_worker.py
```diff
@@ -0,0 +1,37 @@
...
+import click
+import json
+import yaml
+
+from distributed.deploy.spec import run_workers
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.argument("scheduler", type=str, required=False)
+@click.option("--text", type=str, default="", help="")
```
jcrist: perhaps `--spec` and `--spec-file`? `--text` feels like the wrong name to me.

pentschev:
I think the general idea here, but I'm thinking of ways we can extend its usage with other projects, such as dask-cuda, for that I filed https://github.com/rapidsai/dask-cuda/issues/146.

distributed/cli/dask_spec_worker.py
pentschev: I agree, something like `--spec` sounds more appropriate. Maybe `--spec` here and `--spec-file` (instead of `--file` in the line just below)?

mrocklin: Done

dhirschfeld:
> If a spec is provided (either --spec or --spec-file), all other flags are ignored? 

I'd instead recommend being able to override the defaults in the spec file with provided CLI arguments

mrocklin:
I've generalized this from dask-spec-worker to just dask-spec.  It's capable of running both workers and schedulers, or even an arrangement of both.

mrocklin:
I'm feeling decently good about this now.  If no one has any objections I plan to merge this in tomorrow sometime.

distributed/cli/dask_spec.py
```diff
@@ -0,0 +1,37 @@
...
+
+from distributed.deploy.spec import run_spec
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.argument("args", nargs=-1)
+@click.option("--spec", type=str, default="", help="")
+@click.option("--spec-file", type=str, default=None, help="")
+@click.version_option()
+def main(args, spec: str, spec_file: str):
```
jcrist: Do you allow both `spec` and `spec-file`? I'd make them a `xor` relationship, should be only one or the other. I think mixing would be confusing and don't really see a use case for it.

mrocklin: Good ideas both.  I'll handle this tomorrow.  Thanks!

pentschev:
LGTM, I think I may have a use for it already soon. :)

PR #3095, opened 2019-09-26
Add favicon of logo to the dashboard

jrbourbeau:
This PR adds a little dask logo to browser tabs for the dashboard, like so:

Screen Shot 2019-09-26 at 9 01 32 AM


quasiben:
👍

mrocklin:
Nice.  

Should we do something similar for dask-sphinx-theme?

jrbourbeau:
> Should we do something similar for dask-sphinx-theme?

That seems like a nice idea to me (xref https://github.com/dask/dask-sphinx-theme/pull/18)

jrbourbeau:
There are a few test failures due to using the absolute path `href="/statics/images/favicon.ico"`

https://github.com/dask/distributed/blob/031e3a29edd9b4979b910c4773586881b2b26401/distributed/dashboard/tests/test_scheduler_bokeh_html.py#L50

Changing the path to be relative to the current location (`/statics/...` -> `statics/...`) works fine for pages like `localhost:8787/status` and `localhost:8787/workers`, but fails when there's additional nested directories like `localhost:8787/info/main/workers.html`. 

Is there some way to programmatically insert the correct number of `../` to get back to the `statics/` directory?

mrocklin:
cc @jacobtomlinson

jacobtomlinson:
Awesome idea! Would definitely be good to do in the sphinx theme too. 

I might suggest using a tool like https://realfavicongenerator.net/ to generate a full set of icons and markup for this. As with all things web there is a proliferation of standards for all platforms and they should hit most of them. 

It's more likely to be relevant for the website than the dashboard though as I think more people will bookmark/pin that.

mrocklin:
Is this good to go?

jrbourbeau:
Yes, this should be good to go. Merging now

PR #3099, opened 2019-09-27
Fixed SpecCluster identity in html repr

### LINKED ISSUE #3098: Client HTML repr broken with SpecClusters
From within a Jupyter Notebook.

```python
from distributed.deploy.ssh2 import SSHCluster
cluster = SSHCluster(['localhost']*2, connect_kwargs={"known_hosts": None})
client = Client(cluster)
client
```

```python-traceback
AttributeError                            Traceback (most recent call last)
~/miniconda3/envs/dask/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/Projects/dask/distributed/distributed/client.py in _repr_html_(self)
    792             and self.cluster.scheduler
    793         ):
--> 794             info = self.cluster.scheduler.identity()
    795             scheduler = self.cluster.scheduler
    796         elif (

AttributeError: 'Scheduler' object has no attribute 'identity'
```

jacobtomlinson:
Closes #3098

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -786,10 +786,12 @@ class Client(Node):
             return "<%s: not connected>" % (self.__class__.__name__,)
 
     def _repr_html_(self):
+        from .scheduler import Scheduler
+
         if (
             self.cluster
             and hasattr(self.cluster, "scheduler")
-            and self.cluster.scheduler
+            and isinstance(self.cluster.scheduler, Scheduler)
         ):
             info = self.cluster.scheduler.identity()
             scheduler = self.cluster.scheduler


distributed/client.py
```diff
@@ -786,13 +786,16 @@ def __repr__(self):
...
+        if self.cluster:
+            if hasattr(self.cluster, "scheduler") and self.cluster.scheduler:
+                scheduler = self.cluster.scheduler
+                if (
+                    hasattr(self.cluster, "scheduler_comm")
+                    and self.cluster.scheduler_comm
+                ):
+                    info = self.sync(self.cluster.scheduler_comm.identity)
+                else:
+                    info = self.cluster.scheduler.identity()
```
mrocklin: So, I think that this is worse if we have a local Scheduler, and we're in async mode.  In that case `info` is a future, rather than a dictionary.

As an alternative, I recommend the following:

```diff
diff --git a/distributed/client.py b/distributed/client.py
index 381516da..01b1248a 100644
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -786,10 +786,11 @@ class Client(Node):
             return "<%s: not connected>" % (self.__class__.__name__,)

     def _repr_html_(self):
+        from .scheduler import Scheduler
         if (
             self.cluster
             and hasattr(self.cluster, "scheduler")
-            and self.cluster.scheduler
+            and isinstance(self.cluster.scheduler, Scheduler)
         ):
             info = self.cluster.scheduler.identity()
             scheduler = self.cluster.scheduler
```

It's nice to use the local scheduler's identify function directly, because this doesn't require any async communication.  This makes it trivial to use whether or not we're in async mode.

jacobtomlinson: Is that change instead of this, or as well as? We

mrocklin: Instead of, I think.  (I haven't tested it)

My guess is that the implementation currently in this PR would fail in Jupyter if we had code like the following:

```python

client = await Client(asynchronous=True)
client
```

jacobtomlinson:
Pushed up the change proposed by @mrocklin instead.

mrocklin:
Does this resolve your issue @jacobtomlinson ?

jacobtomlinson:
Yep looks good

image

PR #3100, opened 2019-09-27
Support clusters that don't have .security or ._close methods

mrocklin:
Previously clients checked for attributes that may not be universally
present.  Now we only check on scheduler, scheduler_comm, close,
workers, and other elements defined in the `Cluster` interface.

Fixes https://github.com/dask/dask-jobqueue/issues/341

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -652,7 +652,7 @@ class Client(Node):
             with ignoring(AttributeError):
                 loop = address.loop
             if security is None:
-                security = self.cluster.security
+                security = getattr(self.cluster, "security", None)
 
         self.security = security or Security()
         assert isinstance(self.security, Security)
@@ -917,7 +917,7 @@ class Client(Node):
         if self.cluster is not None:
             # Ensure the cluster is started (no-op if already running)
             try:
-                await self.cluster._start()
+                await self.cluster
             except AttributeError:  # Some clusters don't have this method
                 pass
             except Exception:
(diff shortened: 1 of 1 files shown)


mrocklin:
FYI @jcrist

distributed/client.py
```diff
@@ -917,7 +917,7 @@ def _send_to_scheduler(self, msg):
         if self.cluster is not None:
             # Ensure the cluster is started (no-op if already running)
             try:
-                await self.cluster._start()
+                await self.cluster
             except AttributeError:  # Some clusters don't have this method
```
jcrist: Should this `AttributeError` check still be here?

distributed/client.py
```diff
@@ -1266,7 +1266,7 @@ def _handle_error(self, exception=None):
                 self._release_key(key=key)
             if self._start_arg is None:
                 with ignoring(AttributeError):
-                    await self.cluster._close()
+                    await self.cluster.close()
```
jcrist: This is only called for `LocalCluster`, correct?

mrocklin: I think it's any cluster that is given to the client in a situation like the following:

```python
cluster = FooCluster()
client = Client(cluster)
```

Generally, how do you feel about agreeing on a `close` method?

jcrist: if `self._start_arg is None` makes me think it's just `client = Client()`. IMO we shouldn't have the closure of a client shutdown a remote cluster, the cluster object can manage that itself just fine.

I'm fine with agreeing on a close method, I believe we already standardize on this.

mrocklin: Oh indeed, you're correct about `_start_arg is None` and `Client()`.

mrocklin:
I plan to merge this and release once tests pass if there are no objections.

PR #3104, opened 2019-09-30
Retry scheduler connect multiple times

### LINKED ISSUE #3084: Wait on single connection or try multiple connections
When we connect to a remote connection we currently wait on for our full timeout, often ten seconds, something like this:

```python
comm = await connect(address, timeout="10s")
```

However, @jacobtomlinson and I just ran into a situation with Kubernetes where the address that we were connecting to was created at just about the same time, so when we first tried to connect we were sent somewhere that would never receive the connection, but if we try again a second later, things are fine.

```python
comm = await connect(address, timeout="10s")  # this hangs for 10s
```
```python
for i in range(10):  # this connects after 1s
    with ignoring(TimeoutError):
        comm = await comm(address, timeout="1s")
```

This seems to work because, presumably, after the first connection fails and we try reconnecting the network now routes us to the correct location.

In general this second approach seems more robust to networks that might be fiddled with on-the-fly, which is presumably more common in cloud and Kubernetes situations.  However, it also means that we need to become better about cleaning up missed connections.

cc @jcrist @jacobtomlinson and @mmccarty 

The actual code for this is here: https://github.com/dask/distributed/blob/549660e07c0c70fdb17e07c6a18ca438933bd8ba/distributed/comm/core.py#L205-L228

jacobtomlinson:
Closes #3084 

Instead of waiting for the scheduler connection for a long time we are waiting for a shorter time but retrying until the original timeout. If the connection is never made then we raise.

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -207,7 +207,7 @@ async def connect(addr, timeout=None, deserialize=True, connection_args=None):
     # This starts a thread
     while True:
         try:
-            while deadline - time():
+            while deadline - time() > 0:
                 future = connector.connect(
                     loc, deserialize=deserialize, **(connection_args or {})
                 )
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -1,14 +1,13 @@
 from abc import ABC, abstractmethod, abstractproperty
 from datetime import timedelta
 import logging
-from warnings import ignoring
 import weakref
 
 import dask
 from tornado import gen
 
 from ..metrics import time
-from ..utils import parse_timedelta
+from ..utils import parse_timedelta, ignoring
 from . import registry
 from .addressing import parse_address
 
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -212,7 +212,9 @@ async def connect(addr, timeout=None, deserialize=True, connection_args=None):
                 )
                 with ignoring(gen.TimeoutError):
                     comm = await gen.with_timeout(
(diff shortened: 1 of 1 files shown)


distributed/comm/core.py
```diff
@@ -205,14 +206,17 @@ def _raise(error):
...
-                future,
-                quiet_exceptions=EnvironmentError,
-            )
+            while deadline - time() > 0:
+                future = connector.connect(
+                    loc, deserialize=deserialize, **(connection_args or {})
+                )
+                with ignoring(gen.TimeoutError):
+                    comm = await gen.with_timeout(
+                        timedelta(seconds=1), future, quiet_exceptions=EnvironmentError
```
mrocklin: What if the user-specified timeout is quite short (we do this in tests, for example).  

```suggestion
                        timedelta(seconds=min(deadline - time(), 1)), 
                        future, 
                        quiet_exceptions=EnvironmentError
```

jacobtomlinson: Ah! Good point.

mrocklin:
Any thoughts on how to test this?  I can't think of any, but I thought I'd ask.

jacobtomlinson:
It's a tricky one as this is trying to fix a strange edge case. 

Mocking the connect to timeout and then succeed would work, but that is very much code testing that that code does what the code does, rather than actually testing a real example.

Perhaps something fancy with iptables could do it. But I'm not sure if it is worth the effort.

mrocklin:
Yeah, that was my sense as well.  Just thought I'd ask.  Happy to merge on passed CI.

Issue #3091, opened 2019-09-26
Proposed additions to `dask-ssh` CLI


gforsyth:
I've been using `dask-ssh` the last few days and it has been super helpful. Thanks!

I've run into two small things that I think might be of benefit to others to add but wanted to check in (especially given what looks like an upcoming switchover to `asyncssh`).

1. `dask-ssh` keeps an SSH connection open while the scheduler and clients are running -- killing the process also shuts down the remote workers (this seems good). However, SSH timeouts also cause this to break and fall over. `paramiko` (which is currently driving the show here) can send some noise over the wire to keep the connection "active". This could be an optional way (given some `--keepalive` flag) to work around timeouts.  Might also be a bit of scope creep, but I thought I'd raise it.

2. Add an option to disable `StrictHostKeyChecking` -- shouldn't be the default, but if you've just spun up resources to add to a cluster, you might not worry so much about having those machines in `known_hosts`.

Let me know if either/both of these are of interest and I'm happy to take them on.

mrocklin:
Hi @gforsyth !  I'm very glad to see that you're engaging with Dask and SSH.  

One of the biggest surprises of our user survey was that SSH is the most common deployment strategy, so improvements here likely have high impact.

Have you tried the solution in distributed.deploy.ssh2 ?  This new implementation isn't as battle-tested as the one that you're currently using, but I think that it's probably easier to manipulate.  It just passes most things through to asyncssh and the workers.

https://github.com/dask/distributed/blob/f7f6bd77ab8d28d8811b94aa84528d3a470f1ff1/distributed/deploy/ssh2.py#L193-L213

gforsyth:
Thanks for the pointer, @mrocklin !  I'll take a look and see if key checking and keepalive are valid additions to the newer implementation.

mrocklin:
My hope is that we're able to defer those questions entirely to "are key checking and keepalives valid options in asyncssh" and that Dask gets to extract itself from that conversation.

gforsyth:
Looking at https://asyncssh.readthedocs.io/en/latest/api.html#asyncssh.SSHClientConnectionOptions

it would seem that setting `known_hosts` to `None` will handle host checking issues and it's also possible to set `keepalive_interval` to send a ping to keep the connection active.  I'll test both of these out -- the configs can always be added as additional examples to the docstrings.

mrocklin:
Any update here @gforsyth ?  Did this approach end up resolving your problem?

[...]

PR #3105, opened 2019-10-01
Send noise over the wire to keep connection alive

gforsyth:
This change keeps the `dask-ssh` connection alive, preventing ssh
timeouts from closing a pipe and killing workers/schedulers without user
action.

Note that for the `asyncssh` version, this won't be required as
`asyncssh` has a specific `keepalive` parameter that can be passed
through, but before that is the default, this should ease some pain.

Related to issue #3091

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -182,11 +182,16 @@ def async_ssh(cmd_dict):
             )
             return True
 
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @gforsyth !

Also, I notice that this is your first code contribution to this repository.  Welcome!  I'm very glad to have you on board.

PR #3106, opened 2019-10-01
Add Client.shutdown method

mrocklin:
This lets the client shut down the scheduler and workers

Asked for in https://stackoverflow.com/questions/50919227/is-it-possible-to-shutdown-a-dask-distributed-cluster-given-a-client-instance

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1336,9 +1336,11 @@ class Client(Node):
             self._loop_runner.stop()
 
     async def _shutdown(self):
-        with ignoring(CommClosedError):
-            await self.scheduler.terminate(close_workers=True)
-        await self.close()
+        if self.cluster:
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -1335,14 +1335,24 @@ def close(self, timeout=no_default):
         if self._should_close_loop and not shutting_down():
             self._loop_runner.stop()
 
-    def shutdown(self, *args, **kwargs):
-        """ Deprecated, see close instead
+    async def _shutdown(self):
+        if self.cluster:
+            await self.cluster.close()
```
quasiben: Might be nice to include a debug message here for tracking purposes
> logger.debug("Shutting Down Cluster...")

mrocklin:
cc @jcrist we were talking about this

PR #3110, opened 2019-10-03
Support calling `cluster.scale` as async method

### LINKED ISSUE #3107: Should Cluster.scale be an async method?
Currently when using a cluster with `asynchronous=True`, `Cluster.scale` is not an async method, which means that any work that it does either blocks, or is scheduled to run as background task. In contrast, `dask_gateway.GatewayCluster.scale` currently is an async method when running with `asynchronous=True`. Note that it doesn't wait for the workers to be running, but it does wait for them to be submitted.

```python
cluster = await ClusterClass(asynchronous=True)

# Most cluster managers don't need an await here
# Any submissions either block, or run as background tasks
cluster.scale(2)

# dask-gateway currently needs an await here
# This returns after the scale request has been sent to the gateway server,
# not when the workers have been received.
await cluster.scale(2)
```

This was discovered when trying to make dask-gateway work with dask-labextension. It's easy to make labextension work with both, or we could make dask-gateway use a background task here to send the request. Happy with either.

cc @lesteve, @jacobtomlinson, @mrocklin @guillaumeeb, @ian-r-rose for thoughts.

jcrist:
This adds optional support for calling `cluster.scale` as an async
method (i.e. `await cluster.scale(...)`). This is currently optional and
backwards compatible - perhaps in the future we may want to deprecate
calling in the non-async context.

Fixes #3107 

See also https://github.com/dask/dask-labextension/pull/98.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -77,6 +77,19 @@ class ProcessInterface:
         await self.close()
 
 
+class NoOpAwaitable(object):
+    """An awaitable object that always returns None.
+
+    Useful to return from a method that can be called in both asynchronous and
+    synchronous contexts"""
+
+    def __await__(self):
+        async def f():
+            return None
+
+        return f().__await__()
+
+
 class SpecCluster(Cluster):
     """ Cluster that requires a full specification of workers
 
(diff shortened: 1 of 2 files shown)


distributed/deploy/spec.py
```diff
@@ -77,6 +77,19 @@ def __repr__(self):
...
+    """An awaitable object that always returns None.
+
+    Useful to return from a method that can be called in both asynchronous and
+    synchronous contexts"""
+
+    def __await__(self):
+        async def f():
+            return None
+
+        return f().__await__()
```
mrocklin: We might also consider `asyncio.sleep(0)` or `tornado.gen.moment`

Although this might be nicer if either of the two return something, which would look bad in Jupyter.  I suspect that they're clean though.

mrocklin:
This looks good to me.  I made one small comment, but it's mostly cosmetic.

distributed/deploy/spec.py
jcrist: > We might also consider asyncio.sleep(0) or tornado.gen.moment

Why? In this case there's no context switch. Asyncio's event loop uses an eager scheduler - if an async function is called that doesn't await on anything that would block (e.g. `sleep`, I/O, etc...) then things run like normal function calls and return immediately. Inserting a `sleep(0)` or `gen.moment()` (which is deprecated, BTW) would allow another coroutine to resume execution, but I don't see that as necessary here.

We also only return a `NoOpAwaitable` if the cluster is running in async mode, so in Jupyter notebook contexts there'd be no output.

mrocklin:
Thanks @jcrist !  This is in.

I plan to release today, so this should be in a live release soonish.

PR #3115, opened 2019-10-04
Move death timeout logic up to Node.start

mrocklin:
Previously there were some cases where the death-timeout logic
wouldn't reliably be triggered.  Now we handle it higher up in the call
chain, where hopefully it will be more consistent.


mrocklin:
@jcrist , if you have a moment to review

Issue #3014, opened 2019-08-29
Raise exception if the user passes in unused keywords to Client


mrocklin:
For example

```python
client = Client(address, dashboard_address="foo")
```

The dashboard address keyword is useful if we're creating a `LocalCluster`, but it doesn't get used if we're connecting to an already-started scheduler.  In these cases, we should probably raise, otherwise we risk adding confusion.

cc @shwina

detroyejr:
I can try my hand at submitting a PR for this issue.

In lieu of a good existing exception, it might be useful to add a custom exception to help clarify why this is being raised such as:

```
class UnusedKeywordException(ValueError):
    pass
```

I think this exception should then be raised just after this conditional block [here](https://github.com/dask/distributed/blob/master/distributed/client.py#L641).

Something like this might work:

```
if address:
    if "dashboard_address" in kwargs:
        raise UnusedKeywordException("Keyword argument 'address' cannot be used simultaneously with 'dashboard_address'.")
```

Would this be a satisfactory solution?

mrocklin:
Most of the time we tend to use built-in exceptions, but with informative messages.  We should probably either change this policy everywhere (and make dozens of new exception types) or keep it.  I'm in favor of keeping it.

The code you provide would solve the specific issue around dashboard_address, but we would prefer a solution that covers any excess keyword.  dashboard_address was just used above as an example.

detroyejr:
Thanks for the feedback.

Took another look and implemented something more general that would go after [here](https://github.com/dask/distributed/blob/78d8cad9e8934a3b863999832b7c46dfe8162683/distributed/client.py#L917).

```
        if address is not None:
            from .deploy import LocalCluster
            unused_args = [
                arg for arg in self._startup_kwargs
                if arg in list(LocalCluster.__init__.__code__.co_varnames) and 
                arg not in ["silence_logs"]  and
                self._startup_kwargs.get(arg) is not None
            ]

            if unused_args:
                raise ValueError(
                    "{} arguments cannot be used when scheduler or LocalCluster is already running.".format(
                        ", ".join(["'{}'".format(arg) for arg in unused_args])
                    )
                )

```
Errors will be raised when users invoke.

```
from dask.distributed import Client

# Fails
client = Client(address="127.0.0.1:8786", n_workers=4)

# Fails
client = Client(address="127.0.0.1:8786", dashboard_address="127.0.0.1:8787")

# Fails
client = Client(address="127.0.0.1:8786", scheduler_port=8787)

[...]

PR #3117, opened 2019-10-04
Raise exception if the user passes in unused keywords to Client

detroyejr:
Exceptions are raised using the simple approach [suggested by mrocklin](https://github.com/dask/distributed/issues/3014#issuecomment-538349320).

It seems that some tests like [test_client_with_name](https://github.com/dask/distributed/blob/670f2e84a7607b685cbc2f8fbdbfb29aa8267c84/distributed/tests/test_client.py#L5014) fail since they pass in keywords like `silence_logs`.

In this case, are we in favor of working around these tests until they pass or change where these extra arguments are placed? What are best practices here?

#3014

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -643,6 +643,11 @@ class Client(Node):
             if address:
                 logger.info("Config value `scheduler-address` found: %s", address)
 
+        if address is not None and kwargs:
+            raise ValueError(
+                "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
+            )
+
         if isinstance(address, (rpc, PooledRPCCall)):
             self.scheduler = address
         elif hasattr(address, "scheduler_address"):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -644,9 +644,13 @@ class Client(Node):
                 logger.info("Config value `scheduler-address` found: %s", address)
 
         if address is not None and kwargs:
-            raise ValueError(
-                "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
-            )
+            unused_arg = [
+                arg for arg in kwargs if arg not in ["silence_logs"] and kwargs.get(arg) != None
+            ]
+            if unused_arg and any(unused_arg):
+                raise ValueError(
+                    "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
+                )
 
         if isinstance(address, (rpc, PooledRPCCall)):
             self.scheduler = address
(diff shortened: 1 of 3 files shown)


detroyejr:
Some of these tests I expected to fail. Not sure about all the `gen_cluster` failures. Will do some digging.

distributed/client.py
```diff
@@ -643,6 +643,17 @@ def __init__(
             if address:
                 logger.info("Config value `scheduler-address` found: %s", address)
 
+        if address is not None and kwargs:
+            unused_arg = [
+                arg
+                for arg in kwargs
+                if arg not in ["silence_logs"] and kwargs.get(arg) is not None
+            ]
```
mrocklin: I'm curious, why did you have to include these exceptions?  

My guess is that you're working around test failures somewhere in the test suite.  It's entirely possible though that those tests are wrong.

distributed/deploy/tests/test_local.py
```diff
@@ -219,6 +219,27 @@ def test_Client_kwargs(loop):
...
+def test_Client_unused_kwargs_with_address(loop):
+    with pytest.raises(Exception) as argexcept:
+        c = Client(
+            "127.0.0.1:8786", n_workers=2, dashboard_port=8000, silence_logs=None
+        )
+    assert (
+        str(argexcept.value)
+        == "Unexpected keyword arguments: ['dashboard_port', 'n_workers']"
+    )
+
```
mrocklin: Thanks for this test!

distributed/client.py
detroyejr: Yes, this is just to get the current test suite to pass. Some tests like `test_client_repr_closed` include arguments that cause the test to fail:

```
c = yield Client(s.address, asynchronous=True, dashboard_address=None)
```

While `dashboard_address` doesn't have a value, kwargs is no longer None/False so it's going to raise an exception. [Here's the example in travis logs.](https://travis-ci.org/dask/distributed/jobs/593684024#L3253)

I think there are only 2 or 3 tests that would need to be changed. I can do that if that's the better than working around them.

mrocklin: > I think there are only 2 or 3 tests that would need to be changed. I can do that if that's the better than working around them.

Yes, that would be ideal if you're willing.  Thank you for fixing more bugs :)

mrocklin:
Merging this in. Thanks @detroyejr !

Also.  I notice that this is your first code contribution to this repository.  Welcome!

PR #3118, opened 2019-10-07
Attempt to replace dead workers

danpf:
This (at least in a  simple test case) causes workers to be replaced if they have disappeared.

It's possible that this is not the correct way to fix the issue, but basically it updates 'worker_spec' to be in sync with the known workers.  I'm guessing there is a reason it wasn't done this way in the first place so please tell me if this is just bad.

to replicate this, I spawn a bunch of workers with this script (dask-jobqueue), and then manually cancel 6 or so of them after 1 min or so.   
```
 #!/usr/bin/env python

import time

from dask_jobqueue import SLURMCluster


def run_sleepy(client):
    jobs = [client.submit(lambda y: time.sleep(10), x) for x in range(100)]
    client.gather(jobs)


def main():
    with SLURMCluster(
        cores=1,
        memory=f"1GB",
        queue="myqueue",
        job_name="test_worker",
        env_extra=["source ~/.zshrc"],
        walltime="4:00:00",
    ) as cluster:
        cluster.adapt(minimum=0, maximum=10)
            run_sleepy(client)


if __name__ == "__main__":
    main()
```


mrocklin:
We've intentionally decided not to do this for now across the cluster managers.   You should be able to get something close to this using `cluster.adapt(minimum=5, maximum=5)` (I think, I'm not sure I've ever actually tried this.

danpf:
that's not really ideal if there's sections where only a few workers are working to collect, compile and then redistribute -- but either way that adapt call you mentioned doesn't appear to respawn workers (at least within a min or two) :/

PR #3124, opened 2019-10-08
Bump dask dependency

jcrist:
Due to moving some things around (e.g. https://github.com/dask/distributed/pull/3042), distributed 2.5 now relies on dask 2.5. We bump the required version to account for this.

 See https://github.com/dask/dask/issues/5465#issuecomment-538917319.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 click >= 6.6
 cloudpickle >= 0.2.2
-dask >= 2.3
+dask >= 2.5.2
 msgpack
 psutil >= 5.0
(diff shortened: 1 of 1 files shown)


jcrist:
We may want to issue a bugfix release with bumped requirements, as is users can install `distributed >= 2.5` and it won't update `dask`, resulting in a broken installation.

jcrist:
Same for the conda recipe: https://github.com/conda-forge/distributed-feedstock/blob/master/recipe/meta.yaml

mrocklin:
Looks good to me.  Thanks for taking this on so quickly @jcrist

PR #3134, opened 2019-10-10
Add Nanny(config={...}) keyword

mrocklin:
cc @pentschev


pentschev:
LGTM, will be very helpful for UCX! Thanks @mrocklin.

PR #3138, opened 2019-10-11
Refactor dashboard code

### LINKED ISSUE #3048: Cleanup dashboard code
The Dashboard code is effective... but messy.

This module has grown very organically over time, and so has some dead code, plenty of repetition, and so on.  Some plots are in components.py, some in scheduler.py.  Sometimes in both.  There are some plots that we no longer use. 

Additionally, I think that currently I'm the only person that maintains this section.  It might be nice to increase the bus factor in this module, and code cleanup might be a useful way to get accustomed to it.

jacobtomlinson:
Closes #3048.

In an attempt to tidy up the dashboard code I've done some major shuffling of things within the dashboard submodule. I think overall this puts things in more obvious places and should make it more accessible to new contributors. 

Notable changes:
- Components have been broken out in a new `distributed.dashboard.components` submodule and placed into logical groupings "scheduler specific", "worker specific", "shared" and "nvml/gpu". This could be broken down further but there is a balance to be struck between indirection and giant scary files.
- The additional server routes from `scheduler_html.py` and `worker_html.py` have been moved into the `scheduler.py` and `worker.py` files to keep all server things together.
- Shared functions and utilities have been moved around into more appropriate places.

Other things I'd like to do:
- [x] Make use of the [update source function](https://github.com/dask/distributed/pull/3138/files#diff-0db0a1f6d00335e7cc6e5e94eae3f8a1R89-R116) outside of the scheduler components
- [x] ~Re-order components to group `_doc` functions with their `DashboardComponent` classes to reduce indirection (or at least distance).~ _On second thoughts as some docs use multiple components having the docs together at the bottom actually makes more sense._
- [x] Rename functions for consistency (there are a few which should probably follow the `_doc` names)
- [x] Identify and remove unused components. (Could use some pointers on this)


distributed/dashboard/components/scheduler.py
```diff
@@ -0,0 +1,1797 @@
...
+                "destination": y,
+                "bandwidth": value,
+                "bandwidth_text": list(map(format_bytes, value)),
+            }
+            self.fig.title.text = "Bandwidth: " + format_bytes(self.scheduler.bandwidth)
+
+            update(self.source, result)
+
+
+class CurrentLoad(DashboardComponent):
```
mrocklin: This should maybe be broken out into one class per plot?  I'm not sure.

distributed/dashboard/components/scheduler.py
```diff
@@ -0,0 +1,1797 @@
...
+
+                if PROFILING:
+                    curdoc().add_next_tick_callback(
+                        lambda: self.source.stream(new, 10000)
+                    )
+                else:
+                    self.source.stream(new, 10000)
+
+
+class SchedulerTaskStream(TaskStream):
```
mrocklin: I *think* that there should be only one task stream plot, and that these two classes should be merged.  I'm not entirely sure though.

distributed/dashboard/components/scheduler.py
```diff
@@ -0,0 +1,1797 @@
...
+
+            if PROFILING:
+                curdoc().add_next_tick_callback(
+                    lambda: self.source.stream(rectangles, self.n_rectangles)
+                )
+            else:
+                self.source.stream(rectangles, self.n_rectangles)
+
+
+class GraphPlot(DashboardComponent):
```
mrocklin: Maybe rename `GraphPlot` to `TaskGraph` ?  Having plot in a name sounds odd.

distributed/dashboard/components/scheduler.py
```diff
@@ -0,0 +1,1797 @@
...
+
+            self.root.title.text = (
+                "Progress -- total: %(all)s, "
+                "in-memory: %(memory)s, processing: %(processing)s, "
+                "waiting: %(waiting)s, "
+                "erred: %(erred)s" % totals
+            )
+
+
+class MemoryUse(DashboardComponent):
```
mrocklin: I *think* that we no longer use this?  Maybe we should though (or something like it)

jacobtomlinson: Yeah it's not used, but it looks useful. I've created a new individual plot for it for now.

distributed/dashboard/components/scheduler.py
jacobtomlinson: Possibly. The nvml stuff also seems to follow this pattern.

jacobtomlinson: I've had a go at pulling this apart into two classes and ended up with quite a bit of duplication. I'm tempted to say that while this class is quite big and complex it does make sense to gather these metrics in one go.

An alternative would be to make a common base class which deals with the data updating and then subclasses which deal with plotting the parts they want. However, that introduces quite a lot of indirection.

What do you think @mrocklin?

jacobtomlinson:
This should be ready for final review and potential merging now.

distributed/dashboard/components/scheduler.py
mrocklin: Sounds fine to me

distributed/dashboard/components/nvml.py
```diff
@@ -181,14 +189,3 @@ def gpu_utilization_doc(scheduler, extra, doc):
...
-
-try:
-    import pynvml
-
-    pynvml.nvmlInit()
-except Exception:
-    pass
-else:
-    applications["/individual-gpu-memory"] = gpu_memory_doc
-    applications["/individual-gpu-utilization"] = gpu_utilization_doc
```
mrocklin: Did these last two lines go away?  We need them in order to get access to these plots in the JLab extension

jacobtomlinson: They are over here now

https://github.com/dask/distributed/pull/3138/files#diff-bb0ae02b02c7f964b31a044e6dc83b19R401-R409

mrocklin: Ah, no I see now, they're in scheduler.py

mrocklin:
Do we need to include `distributed.dashboard.components` in setup.py ?  (I honestly don't fully understand how package building works)

jacobtomlinson:
Hmm possibly, I've only ever used `find_packages` to specify the `packages` kwarg in the setup. Perhaps someone with better knowledge like @jakirkham could weigh in here?

mrocklin:
@jcrist can you take a look at this?

jcrist:
Yeah, you need to add `distributed.dashboard.components` to `packages`. `packages` needs to be a list of all directories in the library. We could swap out a static list for `find_packages` which generally does a good job of doing this automatically.

mrocklin:
https://github.com/dask/distributed/pull/3150

mrocklin:
Thanks for this @jacobtomlinson !

PR #3153, opened 2019-10-16
Don't overwrite `self.address` if it is present

gforsyth:
Testing out the new `SSHCluster` I was unable to connect to an IP that turned
out to be `None` -- the call to `super().__init__()` was overwriting
`self.address` with `None` by default.  Quick one-line fix to check if the
attribute has already been declared in a child class.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -29,7 +29,7 @@ class ProcessInterface:
     """
 
     def __init__(self, scheduler=None, name=None):
-        self.address = None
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks @gforsyth !  This seems fine to me.  Is there a test that we should add here?  It's ok if the answer is "no", but it's worth asking.

gforsyth:
Definitely worth asking.  And yes, there certainly are.  Tests added.

mrocklin:
Thanks @gforsyth !

PR #3162, opened 2019-10-22
Remove circular reference in Cluster

### LINKED ISSUE #3112: Cluster's have self-reference preventing cleanup
@jcrist just mentioned this, the `Cluster._watch_worker_status` coroutine may contain a reference to the cluster (presumably `self`) and so may keep the cluster alive.  We might want to consider making this a classmethod and a weak reference.

jcrist:
Previously there was a circular reference in
`distributed.deploy.Cluster._watch_worker_status` that would keep the
cluster around after deletion. We fix this here, and add a test. Note
that cleanup isn't instantaneous after `del` since closing down the
cluster takes some time, so the test sleeps in a loop for a bit to check
for this.

Fixes #3112

### DIFF
--- a/distributed/deploy/tests/test_spec_cluster.py
+++ b/distributed/deploy/tests/test_spec_cluster.py
@@ -1,5 +1,6 @@
 import asyncio
 import re
+import weakref
 
 import dask
 from dask.distributed import SpecCluster, Worker, Client, Scheduler, Nanny
@@ -80,6 +81,20 @@ def test_spec_sync(loop):
             assert cluster.loop is client.loop
             result = client.submit(lambda x: x + 1, 10).result()
             assert result == 11
+
+
+def test_no_circular_references(cleanup, loop):
+    import time
+
+    ref = weakref.ref(SpecCluster(scheduler=scheduler, loop=loop))
+
+    for i in range(50):
+        if ref() is None:
+            break
+        time.sleep(0.1)
+        print(i)
+    else:
+        assert False
 
 
 def test_loop_started():
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -2,6 +2,7 @@ import asyncio
 from datetime import timedelta
 import logging
 import threading
+import weakref
 
 from dask.utils import format_bytes
 from tornado import gen
@@ -89,12 +90,17 @@ class Cluster(object):
 
     async def _watch_worker_status(self, comm):
         """ Listen to scheduler for updates on adding and removing workers """
+        ref_self = weakref.ref(self)
         while True:
+            self = None
             try:
                 msgs = await comm.read()
             except OSError:
                 break
 
+            self = ref_self()
+            if self is None:
+                break
             with log_errors():
                 for op, msg in msgs:
                     self._update_worker_status(op, msg)


mrocklin:
It looks like this test is still failing even after the fixup.

TomAugspurger:
Sorry there's a merge conflict now as well @jcrist.

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -82,6 +83,19 @@ def test_spec_sync(loop):
             assert result == 11
 
 
+def test_no_circular_references(cleanup, loop):
```
crepererum: Why do we need to wait here for the cleanup? Shouldn't the following be sufficient?

```python
import gc
try:
    gc.disable()
    ref = weakref.ref(SpecCluster(scheduler=scheduler, loop=loop))
    assert ref() is None
finally:
    gc.enable()
```

mrocklin: Cleanup is generally good practice to include.  It ensures that we don't leak processes or threads between tests.

mrocklin:
```
_______________ ERROR at teardown of test_no_circular_references _______________
    @pytest.fixture
    def loop():
        with check_instances():
            with pristine_loop() as loop:
                # Monkey-patch IOLoop.start to wait for loop stop
                orig_start = loop.start
                is_stopped = threading.Event()
                is_stopped.set()
    
                def start():
                    is_stopped.clear()
                    try:
                        orig_start()
                    finally:
                        is_stopped.set()
    
                loop.start = start
    
                yield loop
    
                # Stop the loop in case it's still running
                try:
                    sync(loop, cleanup_global_workers, callback_timeout=0.500)
                    loop.add_callback(loop.stop)
                except RuntimeError as e:
                    if not re.match("IOLoop is clos(ed|ing)", str(e)):
                        raise
                except gen.TimeoutError:
                    pass
                else:
>                   is_stopped.wait()
distributed/utils_test.py:148: 
../../../miniconda/envs/test-environment/lib/python3.6/contextlib.py:88: in __exit__
    next(self.gen)
distributed/utils_test.py:185: in pristine_loop
    loop.close(all_fds=True)
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/tornado/platform/asyncio.py:228: in close
    super(AsyncIOLoop, self).close(all_fds=all_fds)
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/tornado/platform/asyncio.py:71: in close
    self.asyncio_loop.close()
../../../miniconda/envs/test-environment/lib/python3.6/asyncio/unix_events.py:63: in close
    super().close()
self = <_UnixSelectorEventLoop running=True closed=False debug=False>
    def close(self):
        if self.is_running():
>           raise RuntimeError("Cannot close a running event loop")
E           RuntimeError: Cannot close a running event loop
../../../miniconda/envs/test-environment/lib/python3.6/asyncio/selector_events.py:96: RuntimeError
----------------------------- Captured stderr call -----------------------------
distributed.scheduler - INFO -   Scheduler at:    tcp://10.20.0.99:45497
--------------------------- Captured stdout teardown ---------------------------
Unclosed Comms [<TCP  local=tcp://10.20.0.99:45497 remote=tcp://10.20.0.99:57186>, <TCP rpc local=tcp://10.20.0.99:57186 remote=tcp://10.20.0.99:45497>]
_________________________ test_no_circular_references __________________________
cleanup = None
loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7f89e7078be0>
    def test_no_circular_references(cleanup, loop):
        import time
    
        ref = weakref.ref(SpecCluster(scheduler=scheduler, loop=loop))
    
        for i in range(100):
            if ref() is None:
                break
            time.sleep(0.1)
        else:
>           assert False
E           assert False

```

PR #3163, opened 2019-10-22
Added Prometheus metric for a worker's executing tasks count.

darindf:
Added Prometheus metric for a worker's executing tasks count.


mrocklin:
Thanks @darindf !

PR #3164, opened 2019-10-22
Support automatic TLS

### LINKED ISSUE #1686: Token based security for workers
This came up when working with @mrocklin on daskernetes.

It'd be great if when launching the scheduler I can generate a cryptographically good-enough random token, and all workers then need to provide this token to connect. This allows us to use this as a simpler form of security (just disallow unauthenticated endpoints) than setting up a full blown mutually authenticated TLS setup.

jcrist:
This adds support for automatically securing cluster communication with
TLS. This can be useful for situations where you don't have existing
credentials.

This required the following changes:

- ``Security`` objects now support either paths or contents for all
`key`/`cert`/`ca` fields. Due to limitations in Python's ``ssl`` module,
if contents are provided they must be written to a temporary directory
before being loaded back in. We make sure to use secure methods for
doing this. We also change the ``__repr__`` to not show the raw cert
values in the case they're stored in memory.

- ``Security`` objects now have a classmethod ``temporary``, which can
be used to create temporary credentials using self-signed certs. This
requires ``cryptography`` to be installed. Most environments will
already have ``cryptography``, so this isn't a huge new dependency.

```python
>>> sec = Security.temporary()
```

- Both ``Client`` and ``LocalCluster`` now support passing in
``security=True``, which will generate temporary credentials
automatically for use with that cluster. This api could be supported by
other cluster managers, but for now we restrict to ``LocalCluster``
only.

```python
>>> client = Client(security=True)
>>> client
<Client: 'tls://127.0.0.1:52616' processes=4 threads=8, memory=17.18 GB>
```

Fixes #1686

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -523,8 +523,10 @@ class Client(Node):
         Claim this scheduler as the global dask scheduler
     scheduler_file: string (optional)
         Path to a file with scheduler information if available
-    security: (optional)
-        Optional security information
+    security: Security or bool, optional
+        Optional security information. If creating a local cluster can also
+        pass in ``True``, in which case temporary self-signed credentials will
+        be created automatically.
     asynchronous: bool (False by default)
         Set to True if using this client within async/await functions or within
         Tornado gen.coroutines.  Otherwise this should remain False for normal
@@ -659,8 +661,15 @@ class Client(Node):
             if security is None:
                 security = getattr(self.cluster, "security", None)
 
-        self.security = security or Security()
-        assert isinstance(self.security, Security)
+        if security is None:
+            security = Security()
+        elif security is True:
+            security = Security.temporary()
+            self._startup_kwargs["security"] = security
+        elif not isinstance(security, Security):
+            raise TypeError("security must be a Security object")
+
+        self.security = security
 
         if name == "worker":
             self.connection_args = self.security.get_connection_args("worker")
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -56,7 +56,10 @@ class LocalCluster(SpecCluster):
         like ``['feed', 'run_function']``
     service_kwargs: Dict[str, Dict]
         Extra keywords to hand to the running services
-    security : Security
+    security : Security or bool, optional
+        Configures communication security in this cluster. Can be a security
+        object, or True. If True, temporary self-signed credentials will
+        be created automatically.
     protocol: str (optional)
         Protocol to use like ``tcp://``, ``tls://``, ``inproc://``
         This defaults to sensible choice given other keyword arguments like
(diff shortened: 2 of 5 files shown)


mrocklin:
cc @rcthomas @jacobtomlinson

jcrist:
Test failures seem unrelated.

TomAugspurger:
The many workers test is failing elsewhere.

I haven't seen the `test_keywords` failure yet, but it also seems unrelated.

mrocklin:
@jacobtomlinson is this something that we would want in dask-kubernetes as well?

jacobtomlinson:
I could see this being useful in most cluster implementations. Perhaps adding it to `SpecCluster` would be good?

jrbourbeau:
What do you think about adding this to `SpecCluster` @jcrist?

jcrist:
I think the interface here could be supported by all backends. However, moving around credentials must be done in a secure way, and would be backend specific. For example, on kubernetes you'd create a secret and mount it as a volume containing the credentials. On yarn you'd upload the credentials along with your application. Since this is all very backend specific, I don't think it belongs in `SpecCluster`, but could be implemented by backends matching the same interface.

mrocklin:
> Since this is all very backend specific, I don't think it belongs in SpecCluster, but could be implemented by backends matching the same interface

Would it be reasonable to expand the SpecCluster interface to have `Worker` and `Scheduler` classes (which may be pods or jobs or whatnot) expect a `security=` keyword?  Then it would be up to the downstream libraries to expect this keyword and handle it accordingly.

jcrist:
Maybe? Some backends may not be able to safely move around credentials, and transitioning what kwargs a backend expects may break old code. FWIW handling the `security` kwarg from the cluster interface is only like 4 lines of python:

```
if security is None:
    security = Security()
elif security is True:
    security = Security.temporary()

# Go on and use the security argument as needed...
```

I think for now we should leave it out of SpecCluster, and have backends implement this as they see fit.

mrocklin:
cc @lesteve @guillaumeeb @jacobtomlinson for thoughts on downstream SpecCluster libraries.  Should we accept a `Security` object, and then find some way to safely pass around those credentials?

mrocklin:
(I'm also happy to call this done and move this discussion to an issue)

jacobtomlinson:
> (I'm also happy to call this done and move this discussion to an issue)

That's probably reasonable.

TomAugspurger:
Let's go with that then. Opened https://github.com/dask/distributed/issues/3172 for followup.


There's one failure I don't recall seeing before

```
distributed/deploy/tests/test_ssh.py::test_keywords FAILED               [ 17%]
```

I've restarted that worker to see if it goes away. The other test failures are known.

jrbourbeau:
`distributed/deploy/tests/test_ssh.py::test_keywords` passed on the restart

jcrist:
Good to merge?

mrocklin:
To the extent that I'm able to judge this seems good to me.  I say merge at will @jcrist

PR #3165, opened 2019-10-22
Updated Prometheus documentation

darindf:
Updated Prometheus documentation to include additional metrics, and to clarify and differentiate worker vs scheduler metrics.


darindf:
Formatting is fixed

PR #3166, opened 2019-10-22
Fix Numba serialization when strides is None

pentschev:
This is necessary for newer Numba versions with `__cuda_array_interface__` version 2.

The existing test fails for Numba 0.46.0, so this case is automatically covered, but I'm not sure if we can really have a specific test for this.

cc @quasiben @madsbk

### DIFF
--- a/distributed/protocol/numba.py
+++ b/distributed/protocol/numba.py
@@ -25,7 +25,7 @@ def deserialize_numba_ndarray(header, frames):
     # meaning the array is C-contiguous, so we have to calculate it.
     if strides is None:
         itemsize = np.dtype(header["typestr"]).itemsize
-        strides = tuple((np.cumprod((1, ) + shape[:0:-1]) * itemsize).tolist())
+        strides = tuple((np.cumprod((1,) + shape[:0:-1]) * itemsize).tolist())
 
     arr = numba.cuda.devicearray.DeviceNDArray(
         shape,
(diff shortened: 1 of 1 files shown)


pentschev:
The failure is unrelated to this PR.

distributed/protocol/numba.py
```diff
@@ -18,9 +18,20 @@ def serialize_numba_ndarray(x):
...
     (frame,) = frames
+    strides = header["strides"]
+
+    # Starting with __cuda_array_interface__ version 2, strides can be None,
+    # meaning the array is C-contiguous, so we have to calculate it.
+    if strides is None:
+        itemsize = np.dtype(header["typestr"]).itemsize
+        strides = list(header["shape"][1:]) + [itemsize]
+        for i in range(len(strides) - 1, 0, -1):
+            strides[i - 1] *= strides[i]
```
mrocklin: You might want something like the following?

```python
strides = np.cumprod(shape[::-1]) * itemsize
```

mrocklin: Why didn't this show up in tests?  Don't we test with a C-contiguous array?

pentschev: Thanks for the suggestion, it's indeed what I wanted and couldn't think of when I wrote the code.

There are two reasons it didn't show up in the tests:

1. CUDA code is not being tested by CI;
1. Numba >= 0.46.0 is required (which contains version 2 of `__cuda_array_interface__`.

TomAugspurger:
Did `strides = np.cumprod(shape[::-1]) * itemsize` fail for some cases?

Looks like the linting failure is real: https://travis-ci.org/dask/distributed/jobs/601702095#L491.

pentschev:
Thanks @TomAugspurger, I hadn't seen the latest failure. The new commit should fix it, let's hope it's green now!

mrocklin:
Thanks @pentschev !  Merging in.

PR #3168, opened 2019-10-23
Fix adaptive scaling losing all workers

### LINKED ISSUE #3154: Adaptive cannot recover from lost workers
It seems that in the current version of `distributed=2.6.0` (and in older versions also I think), clusters running in adaptive mode cannot (at least, in most cases) recover from workers being unexpectedly lost. This is best exemplified in the case where if all workers are killed at the same time no more workers spin up to replace them, although this issue also persists in more subtle cases. This seems related to both #2990 and #3118.

This can be easily reproduced by the running the following simple example (it currently uses `dask-jobqueue` but I believe this will be true for most `SpecCluster` derived clusters?):

```
import time
from dask_jobqueue import LSFCluster


cluster = LSFCluster(cores=1,
                     memory='1gb',
                     walltime='00:05',
                     queue='cpuqueue',
                     env_extra=['conda activate distributed'])

cluster.adapt(minimum_jobs=1, maximum_jobs=2)

client = Client(cluster)
future = client.submit(time.sleep, 60)

client.gather([future])
```

and then killing the one spawned worker before it finishes sleeping. The result is a cluster that remains active, but no new workers are spawned leaving things hanging indefinitely.

Adding some simple log statements (excuse the quick and dirty use of print here):

```
print('Adapt start.', flush=True)
print(f'Target {target}.', flush=True)

print(f'plan - {plan}')
print(f'requested - {requested}')
print(f'observed - {observed}')

print(f'workers - {self.cluster.workers}')
print(f'worker_spec - {self.cluster.worker_spec}')
```

at the top of the  `distributive.deploy.adaptive_core.AdaptiveCore.recommendations` method one can see that the cluster state becomes desynchronised with the scheduler state:

```
Adapt start.
Target 1.
plan - set()
requested - set()
observed - set()
workers - {}
worker_spec - {}
Recommendations {"status": "up", "n": 1}.
Adapt start.
Target 1.
plan - {0}
requested - {0}
observed - set()
workers - {0: <LSFJob: status=running>}
worker_spec - {0: {...}}
Recommendations {"status": "same"}.
.
. <------- The above 8 lines repeat
.
Adapt start.
Target 1.
plan - {0}
requested - {0}
observed - {0}
workers - {0: <LSFJob: status=running>}
worker_spec - {0: {...}}
Recommendations {"status": "same"}.     <---- I kill the worker just after this prints.
Adapt start.
Target 1.
plan - {0}
requested - {0}
observed - set()                        <----- The scheduler no longer sees the worker
workers - {0: <LSFJob: status=running>} <------ The cluster still sees the worker
worker_spec - {0: {...}}
Recommendations {"status": "same"}.
.
. <------- The above 8 lines repeat indefinitely
.
```

namely that the scheduler (which here is the source of `observed` I believe?) correctly detects the worker is no longer alive, but the cluster (which here  is the source of `workers`) does not see this change.

This can seemingly be corrected by setting the `distributed.deploy.lost-worker-timeout` config option to 15s (as per #2990 ):

```
... <----------- Same as previous example
Adapt start.
Target 1.
plan - {0}
requested - {0}
observed - {0}
workers - {0: <LSFJob: status=running>}
worker_spec - {0: {...}}
Recommendations {"status": "same"}.     <---- I kill the worker just after this prints.
Adapt start.
Target 1.
plan - {0}
requested - {0}
observed - set() <------ The scheduler sees the worker has been killed.
workers - {0: <LSFJob: status=running>}
worker_spec - {0: {...}}
Recommendations {"status": "same"}.
.
. <----- The above 8 lines repeat
.
Adapt start.
Target 1.
plan - {0}
requested - set()
observed - set()
workers - {}   <------- The worker is now removed correctly after some time
worker_spec - {0: {...}}
Recommendations {"status": "same"}.
.
. <----- The above 8 lines repeat
.
```

Even though now the cluster `workers` list appears be in sync with the scheduler kept worker list, it seem like this would still not be enough to correct this issue as the 'planned' workers (which are described by the `worker_spec` I think?) match the target number of workers, but the cluster does not request these.

Adding `await self.cluster` to the top of `recommendations` seems to fix this issue whereby the workers are now re-spawned (I think as it leads to `cluster._correct_state` being called?) but I don't know enough about the internals of these classes to know if this is a correct thing to do?

Sorry for the long post!

So three questions -

1) Is it safe to add `await self.cluster` to the top of `recommendations`, or is this a bad solution / just hiding a larger issue?

2) This only works if the `lost-worker-timeout` config is set manually either in code or the config file - should there be a built-in default for this value (apologies if I've missed this!)? 

3) There doesn't seem to be much information about `lost-worker-timeout` - how resilient is it? Is there something missing here which should keep the cluster + scheduler better in sync?

SimonBoothroyd:
First attempt to close #3154

### DIFF
--- a/distributed/deploy/adaptive_core.py
+++ b/distributed/deploy/adaptive_core.py
@@ -179,11 +179,6 @@ class AdaptiveCore:
 
         try:
 
-            if len(self.plan) != len(self.requested):
-                # Ensure that the number of planned and requested workers
-                # stay in sync.
-                await self.cluster
-
             target = await self.safe_target()
             recommendations = await self.recommendations(target)
 
--- a/distributed/deploy/adaptive_core.py
+++ b/distributed/deploy/adaptive_core.py
@@ -178,6 +178,12 @@ class AdaptiveCore:
         self._adapting = True
 
         try:
+
+            if len(self.plan) != len(self.requested):
+                # Ensure that the number of planned and requested workers
+                # stay in sync.
+                await self.cluster
+
             target = await self.safe_target()
             recommendations = await self.recommendations(target)
 
--- a/distributed/deploy/tests/test_spec_cluster.py
+++ b/distributed/deploy/tests/test_spec_cluster.py
@@ -1,5 +1,6 @@
 import asyncio
 import re
+from time import sleep
 
 import dask
 from dask.distributed import SpecCluster, Worker, Client, Scheduler, Nanny
@@ -129,6 +130,33 @@ async def test_scale(cleanup):
         await cluster.scale(2)
         await cluster
         assert len(cluster.workers) == 2
+
+
+@pytest.mark.asyncio
+async def test_adaptive_killed_worker(cleanup):
+    with dask.config.set({"distributed.deploy.lost-worker-timeout": 1}):
+
+        with SpecCluster(
+            worker={"cls": Nanny, "options": {"nthreads": 1}},
+            scheduler={"cls": Scheduler, "options": {"port": 0}},
+        ) as cluster:
+
+            with Client(cluster) as client:
+
+                cluster.adapt(minimum=1, maximum=1)
+
+                # Scale up a cluster with 1 worker.
+                while len(cluster.workers) != 1:
+                    await asyncio.sleep(1)
+
+                future = client.submit(sleep, 5)
+
+                # Kill the only worker.
+                worker_id = next(iter(cluster.workers))
+                await cluster.workers[worker_id].kill()
+
+                # Wait for the worker to re-spawn and finish sleeping.
+                future.result(timeout=10)
 
 
 @pytest.mark.asyncio
--- a/distributed/deploy/tests/test_spec_cluster.py
+++ b/distributed/deploy/tests/test_spec_cluster.py
@@ -132,31 +132,33 @@ async def test_scale(cleanup):
         assert len(cluster.workers) == 2
 
 
+@pytest.mark.slow
 @pytest.mark.asyncio
 async def test_adaptive_killed_worker(cleanup):
-    with dask.config.set({"distributed.deploy.lost-worker-timeout": 1}):
+    with dask.config.set({"distributed.deploy.lost-worker-timeout": 0.1}):
 
-        with SpecCluster(
+        async with SpecCluster(
+            asynchronous=True,
             worker={"cls": Nanny, "options": {"nthreads": 1}},
             scheduler={"cls": Scheduler, "options": {"port": 0}},
         ) as cluster:
 
-            with Client(cluster) as client:
+            async with Client(cluster, asynchronous=True) as client:
 
                 cluster.adapt(minimum=1, maximum=1)
 
                 # Scale up a cluster with 1 worker.
                 while len(cluster.workers) != 1:
-                    await asyncio.sleep(1)
+                    await asyncio.sleep(0.01)
 
-                future = client.submit(sleep, 5)
+                future = client.submit(sleep, 0.1)
 
                 # Kill the only worker.
-                worker_id = next(iter(cluster.workers))
+                [worker_id] = cluster.workers
                 await cluster.workers[worker_id].kill()
 
                 # Wait for the worker to re-spawn and finish sleeping.
-                future.result(timeout=10)
+                await future.result(timeout=5)
 
 
 @pytest.mark.asyncio
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -113,6 +113,14 @@ class Adaptive(AdaptiveCore):
             target_duration=self.target_duration
         )
 
+    async def recommendations(self, target: int) -> dict:
+        if len(self.plan) != len(self.requested):
+            # Ensure that the number of planned and requested workers
+            # are in sync before making recommendations.
+            await self.cluster
+
+        return await super(Adaptive, self).recommendations(target)
+
     async def workers_to_close(self, target: int):
         """
         Determine which, if any, workers should potentially be removed from


danpf:
seems much cleaner than #3118 :+1: 
I'll try this for a little bit and see if it works for me.

distributed/deploy/adaptive_core.py
```diff
@@ -178,6 +178,12 @@ def stop(self):
         self._adapting = True
 
         try:
+
+            if len(self.plan) != len(self.requested):
+                # Ensure that the number of planned and requested workers
+                # stay in sync.
+                await self.cluster
```
TomAugspurger: Do we require that `self.cluster` be awaitable for every backend?

SimonBoothroyd:
@TomAugspurger I added a simple test which hopefully should be sufficient - without the change the test will fail with a timeout error as no worker will be spawned to fulfil the request I think.

distributed/deploy/adaptive_core.py
SimonBoothroyd: That's a good point!

I've moved this logic into the main `Adaptive` class where there definitely should be a cluster provided, but I'm not sure what the best way to check if the cluster object is awaitable. Really this should only be needed if the cluster is a `SpecCluster`, but I think checking for that here will introduce a circular dependency?

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -131,6 +132,33 @@ class MyCluster(SpecCluster):
...
+                    await asyncio.sleep(1)
+
+                future = client.submit(sleep, 5)
+
+                # Kill the only worker.
+                worker_id = next(iter(cluster.workers))
+                await cluster.workers[worker_id].kill()
+
+                # Wait for the worker to re-spawn and finish sleeping.
+                future.result(timeout=10)
```
mrocklin:
```suggestion
               await future.result(timeout=10)
```

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -131,6 +132,33 @@ class MyCluster(SpecCluster):
...
+            scheduler={"cls": Scheduler, "options": {"port": 0}},
+        ) as cluster:
+
+            with Client(cluster) as client:
+
+                cluster.adapt(minimum=1, maximum=1)
+
+                # Scale up a cluster with 1 worker.
+                while len(cluster.workers) != 1:
+                    await asyncio.sleep(1)
```
mrocklin:
```suggestion
                    await asyncio.sleep(0.01)
```

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -131,6 +132,33 @@ class MyCluster(SpecCluster):
...
+
+                future = client.submit(sleep, 5)
+
+                # Kill the only worker.
+                worker_id = next(iter(cluster.workers))
```
mrocklin: If you're feeling fancy ...

```suggestion
                [worker_id] = cluster.workers
```

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -131,6 +132,33 @@ class MyCluster(SpecCluster):
         assert len(cluster.workers) == 2
 
 
+@pytest.mark.asyncio
```
mrocklin: It looks like this test takes a while (longer than a second).  If we can, it would be good to reduce the time here.  Most tests run in under a second, and it's nice to keep the test suite running reasonably fast.  This may not be possible given the nature of the test, but anything that you can do here would be welcome.

Additionally, we should mark this test with `@pytest.mark.slow`

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -131,6 +132,33 @@ class MyCluster(SpecCluster):
...
+async def test_adaptive_killed_worker(cleanup):
+    with dask.config.set({"distributed.deploy.lost-worker-timeout": 1}):
+
+        with SpecCluster(
+            worker={"cls": Nanny, "options": {"nthreads": 1}},
+            scheduler={"cls": Scheduler, "options": {"port": 0}},
+        ) as cluster:
+
+            with Client(cluster) as client:
```
mrocklin: If you are doing an async style test (which I recommend) then I encourage you to use an asynchronous cluster and client using `async with` and the `asynchronous=` keyword.  Other tests in this file should provide an example.

distributed/deploy/adaptive_core.py
mrocklin: I *believe* that this is a safe assumption.  Certainly SpecCluster does this, and I think that dask-labextension expects it, so probably @jcrist has enforced this on dask-yarn and dask-gateway.

mrocklin:
In principle this seems fine to me.  Thank you for identifying the issue and providing a few @SimonBoothroyd .  I added a bunch of minor suggestions, but they should be straightforward.

distributed/deploy/tests/test_spec_cluster.py
SimonBoothroyd: Ah that's good to know about - I've updated the tests to include the `async with`!

distributed/deploy/tests/test_spec_cluster.py
SimonBoothroyd: I've tried to cut down the time taken to run the test but couldn't get it below ~4s unfortunately so I've marked it as slow!

SimonBoothroyd:
Thanks @mrocklin - the changes should all have been made in 9035f86!

distributed/deploy/adaptive_core.py
jcrist: Dask-gateway is awaitable, but doesn't support adaptive yet. dask-yarn is currently not async at all, but should be (TODO). I consider the async interface that `distributed.deploy.Cluster` defines to be the proper spec, so anything that supports should be fine.

mrocklin:
Thanks @SimonBoothroyd .  This is in.

PR #3175, opened 2019-10-28
Nanny/Worker with unspecified protocol will get from scheduler address

pentschev:
This ensures protocol is taken from the scheduler address when unspecified. If the user doesn't pass it when using a different protocol (such as `ucx`), it will cause the worker to use `tcp` by default and make communication not work as intended. Therefore, it's safer to get from the scheduler address to prevent faulty behavior that can be difficult to detect.

cc @mrocklin

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1500,7 +1500,7 @@ async def test_interface_async(loop, Worker):
 @pytest.mark.asyncio
 @pytest.mark.parametrize("Worker", [Worker, Nanny])
 @pytest.mark.parametrize("protocol", ["tcp", "ucx"])
-async def test_protocol_from_scheduler_address(loop, Worker, protocol):
+async def test_protocol_from_scheduler_address(Worker, protocol):
     if protocol == "ucx":
         ucp = pytest.importorskip("ucp")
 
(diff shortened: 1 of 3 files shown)


pentschev:
I'm still thinking of a suitable test, but haven't yet figured one without making `protocol` an attribute of `Nanny`/`Worker` that I can use to assert later on. Suggestions welcome.

mrocklin:
Maybe make a scheduler with UCX and then verify that `worker.address.startswith("ucx://")` ?

pentschev:
> Maybe make a scheduler with UCX and then verify that `worker.address.startswith("ucx://")` ?

But that would not run on CI today since AFAIK there's no UCX available. Is that still ok?

mrocklin:
It's fine with me

pentschev:
@mrocklin I've added such test. As discussed before, CI isn't going to run it, and to run locally one needs #3177 as well.

distributed/tests/test_worker.py
```diff
@@ -1497,6 +1497,22 @@ def test_resource_limit(monkeypatch):
                 assert all("127.0.0.1" == d["host"] for d in info["workers"].values())
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("Worker", [Worker, Nanny])
+@pytest.mark.parametrize("protocol", ["tcp", "ucx"])
```
mrocklin:
```suggestion
@pytest.mark.parametrize("protocol", ["ucx"])
```

I recommend that we remove tcp here to keep tests somewhat fast.

mrocklin:
Thanks @pentschev !

pentschev:
Thanks for reviewing and merging @mrocklin !

PR #3177, opened 2019-10-29
Update UCX variables to use sockcm by default

pentschev:
Without `sockcm` workers/client can't connect to scheduler. For example `test_ucx_localcluster` won't pass without this.

@quasiben @madsbk could you please check if you agree with this change?


pentschev:
The failing test is unrelated.

pentschev:
Yeah, it sets things correctly, the issue is only when we need specific variables per process (e.g., a different `UCX_NET_DEVICES` for each worker), but for unit tests the hardcoded variables work fine.

pentschev:
CI failure is unrelated.

PR #3180, opened 2019-10-29
Efficiently serialize zero strided NumPy arrays

jrbourbeau:
This PR updates how NumPy arrays with zero-strided axes are serialized in order to avoid transmitting the entire array where possible. Instead dimensions with a zero stride are collapsed, this collapsed array is serialized, and upon deserialization the smaller array is then broadcasted to the appropriate shape.

Taking the following as an example:

```python
In [1]: import numpy as np

In [2]: a = np.arange(3)

In [3]: x = np.broadcast_to(a, (5, 3))

In [4]: x
Out[4]:
array([[0, 1, 2],
       [0, 1, 2],
       [0, 1, 2],
       [0, 1, 2],
       [0, 1, 2]])
```

Instead of serializing the entire `x` array (with `x.shape = (5,3)`), we would serialize the smaller `[0, 1, 2]` array. Then after deserializing the smaller array, we use `np.broadcast_to` to reproduce the expected `x` array with shape `(5,3)`.

### DIFF
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -2,11 +2,6 @@ import logging
 
 from dask.sizeof import sizeof
 
-try:
-    import numpy as np
-except ImportError:
-    np = None
-
 logger = logging.getLogger(__name__)
 
 
@@ -16,11 +11,6 @@ def safe_sizeof(obj, default_size=1e6):
     This returns a default size of 1e6 if the sizeof function fails
     """
     try:
-        # Handle zero-strided NumPy arrays
-        if np is not None and isinstance(obj, np.ndarray) and 0 in obj.strides:
-            from .protocol.numpy import _zero_strided_slice
-
-            obj = _zero_strided_slice(obj)
         return sizeof(obj)
     except Exception:
         logger.warning("Sizeof calculation failed.  Defaulting to 1MB", exc_info=True)
--- a/distributed/sizeof.py
+++ b/distributed/sizeof.py
@@ -2,6 +2,11 @@ import logging
 
 from dask.sizeof import sizeof
 
+try:
+    import numpy as np
+except ImportError:
+    np = None
+
 logger = logging.getLogger(__name__)
 
 
@@ -11,6 +16,11 @@ def safe_sizeof(obj, default_size=1e6):
     This returns a default size of 1e6 if the sizeof function fails
     """
     try:
+        # Handle zero-strided NumPy arrays
+        if np is not None and isinstance(obj, np.ndarray) and 0 in obj.strides:
+            from .protocol.numpy import _zero_strided_slice
+
+            obj = _zero_strided_slice(obj)
         return sizeof(obj)
     except Exception:
         logger.warning("Sizeof calculation failed.  Defaulting to 1MB", exc_info=True)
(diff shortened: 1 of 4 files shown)


TomAugspurger:
That's a neat trick.

Any reason you marked this as a WIP?

No idea what is going on with the CI:

```

  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/_pytest/config/__init__.py", line 451, in _importconftest

    return self._conftestpath2mod[key]

KeyError: PosixPath('/home/travis/build/dask/distributed/conftest.py')

```

jrbourbeau:
> Any reason you marked this as a WIP?

I wanted to run the full CI with these changes...which is currently experiencing some problems. I've not seen this problem before. Seems to only happen on the 3.5 build, I wonder if it's related to dropping Python 3.5 in dask

mrocklin:
@jrbourbeau should this go in?

mrocklin:
Or rather, in principle this seems fine to me.  If you're comfortable merging, then please do.

jrbourbeau:
Thanks for giving this a look @mrocklin! The only additional change I think we should make here is handling `sizeof(x)` when `x` is a zero-strided array. As `dask.sizeof.sizeof` can be used outside the context of `distributed`, it probably makes more sense for me to update `safe_sizeof` here in `distributed`.

mrocklin:
OK.  I'm happy to defer to your judgement here.

distributed/sizeof.py
```diff
@@ -11,6 +16,11 @@ def safe_sizeof(obj, default_size=1e6):
     This returns a default size of 1e6 if the sizeof function fails
     """
     try:
+        # Handle zero-strided NumPy arrays
+        if np is not None and isinstance(obj, np.ndarray) and 0 in obj.strides:
+            from .protocol.numpy import _zero_strided_slice
```
mrocklin: It seems off to include this special case in this function.  Is this a case that we can fix upstream?

jrbourbeau: I agree this seems a little off. My main motivation for including this special case here instead of upstream in `dask.sizeof` was because this is tied specifically to how `distributed` handles zero-strided arrays. In theory, `dask.sizeof` could be used independently of `distributed`. Although I'm not sure how common that is (I suspect probably not very common).

We could move this upstream to `dask`. Does that seem more reasonable than what I have here?

mrocklin: To be honest I'm not sure what you mean by zero-strided numpy array.  This is an array that is stored as, say a 2d array, but uses stride tricks to appear to be 3d or 4d or somethiing?

In that case does `dask.sizeof` report the correct size of the object in memory?  I guess I'm not sure how what distributed does here affects things.  If we have one of these objects in memory, how much RAM does it take up?  I would think that this would be independent of which library is asking for the size.

mrocklin:
Checking in again here @jrbourbeau .  I know that you're busy these days so no rush.  Should we remove the sizeof change here, merge the rest of this in, and change the sizeof implementation in dask/dask?

jrbourbeau:
Yeah, following our conversation the other day I think that sounds good. Just pushed a commit to revert the `sizeof.py` changes (although I did leave the added `safe_sizeof` tests)

mrocklin:
Thanks @jrbourbeau .  This is in.

PR #3183, opened 2019-10-30
check bokeh RC

TomAugspurger:
cc @bryevdv

Not sure how extensive our unit tests are, but this will hit the basics.


bryevdv:
@TomAugspurger FYI if this is just for the Bokeh-related Dask unit tests, we already run those on every build in our `downstream` job, e.g.

https://travis-ci.org/bokeh/bokeh/jobs/604671457

I was mostly hoping someone could just also do a 5 minute smoke test running real Dask tasks with the RC installed.

PR #3193, opened 2019-11-04
Run preload scripts before starting communication

pentschev:
This ensures preload script runs before starting communications, allowing any initialization of communication libraries to happen before beginning. It also makes the name more consistent, as preload probably is intended to run before anything else.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1206,6 +1206,8 @@ class Scheduler(ServerNode):
 
         self.clear_task_state()
 
+        preload_modules(self.preload, parameter=self, argv=self.preload_argv)
+
         with ignoring(AttributeError):
             for c in self._worker_coroutines:
                 c.cancel()
@@ -1242,8 +1244,6 @@ class Scheduler(ServerNode):
                     os.remove(fn)
 
             weakref.finalize(self, del_scheduler_file)
-
-        preload_modules(self.preload, parameter=self, argv=self.preload_argv)
 
         self.start_periodic_callbacks()
 
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -976,18 +976,19 @@ class Worker(ServerNode):
         enable_gc_diagnosis()
         thread_state.on_event_loop_thread = True
 
-        self.listen(self._start_address, listen_args=self.listen_args)
-        self.ip = get_address_host(self.address)
-
-        if self.name is None:
-            self.name = self.address
-
         preload_modules(
             self.preload,
             parameter=self,
             file_dir=self.local_directory,
             argv=self.preload_argv,
         )
+
+        self.listen(self._start_address, listen_args=self.listen_args)
+        self.ip = get_address_host(self.address)
+
+        if self.name is None:
+            self.name = self.address
+
         # Services listen on all addresses
         # Note Nanny is not a "real" service, just some metadata
         # passed in service_ports...


pentschev:
cc @jakirkham @madsbk @mrocklin this allows us to remove the `ucp.reset()` call from https://github.com/rapidsai/dask-cuda/blob/branch-0.11/dask_cuda/initialize.py#L73 and ensure UCX-Py gets initialized before starting communication. In the long run it's probably useful to do UCX-Py initialization as in https://github.com/dask/distributed/pull/3178 anyway, but this PR unblocks some of the efforts that have been hitting the `ucp.reset` issue.

mrocklin:
Hrm, it looks like this causes the current preload tests to fail, so this policy change (which I agree seems sensible) may break existing expectations.

jakirkham:
@bm371613, it looks like you originally authored this feature. Do you have thoughts on this proposed change?

pentschev:
Hmm, I didn't anticipate that issue. I will take a good look if this really breaks something or there's some assumption on the test itself that is causing this.

mrocklin:
It looks like the test is checking the value of the address, so it's very specifically testing against the communication system.  I don't have enough exposure to how people use preload scripts to know if this is likely to be an issue.

pentschev:
Yeah, I see that now. It seems that this may be important for the scheduler only? The change that we actually need here for dask-cuda is only on the worker side, so I could totally revert the scheduler one. Alternatively, we could also have the preload script to run before or after the communication setup based on user input, this could become too complex and cluttered though.

mrocklin:
In principle I think that this change is sensible.  The only question is whether or not we want to make it, given that it may disrupt some people.  In the original post you mention that this change my be redundant to  #3178 .  I'm now curious how hard it would be to get that in quickly instead and ignore the preload topic for now.

pentschev:
I guess we should have this in the next few days, I guess the main advantage is that this unblocks people immediately. I don't know whether the changes in #3178 will require also other changes in dask-cuda or elsewhere that may make things a bit slower.

mrocklin:
> I guess the main advantage is that this unblocks people immediately.

That's probably not a sufficient reason for a change in mainline Dask unfortunately.  If we want to make the change here then we need to have a strong argument that this is the better solution.  Putting on my Dask maintainer hat this seems ambiguous enough that I am inclined to just leave it as-is.  

Hopefully in the mean time people can work from branches (though I know, you deal with enough of these already)

pentschev:
Putting my syntactic correctness hat, the actual argument is that, as the name suggests, preload should come before anything else (which is in fact what I believed until I read the code), and not before a preselection of things that might have been so chosen for no compelling reason. I do understand that this change may break usage and if this is a hassle I don't mind working from branches for the time being. As long as nobody else has a problem, I'm fine with going forward with the changes here or closing it.

mrocklin:
> that might have been so chosen for no compelling reason

The current position is for a reason.  Very few users want to screw around with the communications system.  Instead, most users want to get the scheduler and worker after they have been hooked up to everything, but before they are assigned any work.  It's very reasonable here for user code on the worker to want to check in with the scheduler for example.  If there is only one setup method then it should probably be here.  

As you suggest, we could also have several optional functions for different stages in the lifetime of a worker.

pentschev:
I'd rather keep things simple instead of having preload, pre-preload, and so on. I can wait for #3178, unless @jakirkham is badly blocked, I would suggest we close this.

mrocklin:
Rather than `pre-preload` we would probably add a new function to the preload set of functions

```python
def initialize(...):
    ...

def setup(...):
    ...

def teardown(...):
    ...
```

But yeah, I'd perfer to hold off on this until we know that we need it.

pentschev:
I agree, let's put this on wait.

bm371613:
This change makes sense to me.

mrocklin:
@pentschev is this still necessary?

If this is important then we could also add a function like `def initialize` which we chose to run as we create the worker`.

pentschev:
I think we can live without it for now. I'm going to close this and reopen if needed in the future. Thanks for the discussion everyone, and also @bm371613 for chiming in on this.

PR #3198, opened 2019-11-05
Adding PatchedDeviceArray to drop stride attribute for cupy<7.0

rjzamora:
As discussed in [cudf#3236](https://github.com/rapidsai/cudf/pull/3236), cupy does not handle the case that `__cuda_array_interface__['strides'] == None`.  However, numba 0.46+ **does**.  This leads to deserialization failures in the `cupy.asarray` call with problematic numba/cupy package combinations.  

To avoid errors in `deserialize_cupy_array`  (e.g. `TypeError: zip argument #2 must support iteration`), I am just copying the approach used by @brandon-b-miller in cudf.

### DIFF
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -12,7 +12,7 @@ class PatchedCudaArrayInterface(object):
     def __init__(self, ary):
         vsn = LooseVersion(cupy.__version__)
         cai = ary.__cuda_array_interface__
-        if vsn < "7.0.0" and cai.get("strides") is None:
+        if vsn < "7.0.0rc1" and cai.get("strides") is None:
             cai.pop("strides", None)
         self.__cuda_array_interface__ = cai
 
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -19,7 +19,7 @@ class PatchedDeviceArray(object):
         if self.vsn >= "7.0.0" or name != "__cuda_array_interface__":
             return getattr(self.parent, name)
         else:
-            # Cupy<7.0 cannot handle 
+            # Cupy<7.0 cannot handle
             # __cuda_array_interface__['strides'] == None
             rtn = self.parent.__cuda_array_interface__
             if rtn.get("strides") is None:
@@ -41,8 +41,8 @@ def serialize_cupy_ndarray(x):
 def deserialize_cupy_array(header, frames):
     (frame,) = frames
     arr = cupy.ndarray(
-        header["shape"], dtype=header["typestr"], memptr=cupy.asarray(
-            PatchedDeviceArray(frame)
-        ).data
+        header["shape"],
+        dtype=header["typestr"],
+        memptr=cupy.asarray(PatchedDeviceArray(frame)).data,
     )
     return arr
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -10,7 +10,7 @@ class PatchedCudaArrayInterface(object):
     #       once Cupy<7.0 is no longer supported
     def __init__(self, ary):
         cai = ary.__cuda_array_interface__
-        cai_cupy_vsn = cupy.ndarray(0).__cuda_array_interface__["version"]
+        cai_cupy_vsn = cupy.ndarray(None).__cuda_array_interface__["version"]
         if cai.get("strides") is None and cai_cupy_vsn < 2:
             cai.pop("strides", None)
         self.__cuda_array_interface__ = cai
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -3,6 +3,28 @@ Efficient serialization GPU arrays.
 """
 import cupy
 from .cuda import cuda_serialize, cuda_deserialize
+from distutils.version import LooseVersion
+
+
+class PatchedDeviceArray(object):
+    # TODO: This class wont be necessary
+    #       once Cupy<7.0 is no longer supported
+    def __init__(self, ary):
+        self.parent = ary
+        self.vsn = LooseVersion(cupy.__version__)
+
+    def __getattr__(self, name):
+        if name == "parent":
+            raise AttributeError()
+        if self.vsn >= "7.0.0" or name != "__cuda_array_interface__":
+            return getattr(self.parent, name)
+        else:
+            # Cupy<7.0 cannot handle 
+            # __cuda_array_interface__['strides'] == None
+            rtn = self.parent.__cuda_array_interface__
+            if rtn.get("strides") is None:
+                rtn.pop("strides")
+            return rtn
 
 
 @cuda_serialize.register(cupy.ndarray)
@@ -19,6 +41,8 @@ def serialize_cupy_ndarray(x):
 def deserialize_cupy_array(header, frames):
     (frame,) = frames
     arr = cupy.ndarray(
-        header["shape"], dtype=header["typestr"], memptr=cupy.asarray(frame).data
+        header["shape"], dtype=header["typestr"], memptr=cupy.asarray(
+            PatchedDeviceArray(frame)
+        ).data
     )
     return arr
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -13,7 +13,7 @@ class PatchedCudaArrayInterface(object):
         vsn = LooseVersion(cupy.__version__)
         cai = ary.__cuda_array_interface__
         if vsn < "7.0.0" and cai.get("strides") is None:
-            cai.pop("strides")
+            cai.pop("strides", None)
         self.__cuda_array_interface__ = cai
 
 
@@ -30,9 +30,9 @@ def serialize_cupy_ndarray(x):
 @cuda_deserialize.register(cupy.ndarray)
 def deserialize_cupy_array(header, frames):
     (frame,) = frames
+    if not isinstance(frame, cupy.ndarray):
+        frame = PatchedCudaArrayInterface(frame)
     arr = cupy.ndarray(
-        header["shape"],
-        dtype=header["typestr"],
-        memptr=cupy.asarray(PatchedCudaArrayInterface(frame)).data,
+        header["shape"], dtype=header["typestr"], memptr=cupy.asarray(frame).data
     )
     return arr
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -10,7 +10,7 @@ class PatchedCudaArrayInterface(object):
     #       once Cupy<7.0 is no longer supported
     def __init__(self, ary):
         cai = ary.__cuda_array_interface__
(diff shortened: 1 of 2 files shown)


distributed/protocol/cupy.py
```diff
@@ -3,6 +3,28 @@
...
+            raise AttributeError()
+        if self.vsn >= "7.0.0" or name != "__cuda_array_interface__":
+            return getattr(self.parent, name)
+        else:
+            # Cupy<7.0 cannot handle
+            # __cuda_array_interface__['strides'] == None
+            rtn = self.parent.__cuda_array_interface__
+            if rtn.get("strides") is None:
+                rtn.pop("strides")
+            return rtn
```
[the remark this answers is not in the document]
rjzamora: > Is it possible to modify the __cuda_array_interface__ attribute of a cupy array directly..

I do not believe so, the `__cuda_array_interface__` property is immutable.

> What attributes does cupy.asarray need?

I'm sure we could do that instead, but is there a good reason to reimplement everything when we can just toss out one problematic `__cuda_array_interface__` attribute? (note that I might be missunderstanding)

rjzamora:
> Is there a test that we should add here, or is this covered by our existing tests when the versions are just right?

I'll look into this - I had forgotten about testing since I discovered the problem in a failing ucx-py test

distributed/protocol/cupy.py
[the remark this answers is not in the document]
mrocklin: Well, what attributes do we actually need here?  I wouldn't be surprised if we *only* needed the `__cuda_array_interface__` attribute, and didn't need to mimic everything else.  If that was the case then I would suggest that we make a simpler object that only exposed that.  It's not a big deal though, and if we're pressed for time I'd be happy to pass on this.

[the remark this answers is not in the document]
brandon-b-miller: Originally I figured forwarding the attrs was a good check in case cupy needed to use any information or code built into the numba array to create the cupy object. However it occurs to me (now a week later of course) that the `__cuda_array_interface__` is sort of "by definition" all that one needs to know, so I suspect that branch of code isn't ever used.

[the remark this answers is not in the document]
rjzamora: > I wouldn't be surprised if we only needed the __cuda_array_interface__ attribute, and didn't need to mimic everything else. If that was the case then I would suggest that we make a simpler object that only exposed that.

Good call - Doing this seems to work fine

rjzamora:
I started looking into testing, and I am a bit confused by the ucx-py test failure that is fixed by this PR.  The problem in that test ([test_send_recv_two_workers.py::test_send_recv_cu[cupy]](https://github.com/rapidsai/ucx-py/blob/19d5e6c89e5ff6cf42c2b28525631dc27839def4/tests/test_send_recv_two_workers.py#L221)) seems to be that the `frame` object is type `numba.cuda.cudadrv.devicearray.DeviceNDArray` (with `__cuda_array_interface__.strides == None`).  However, distributed has a different `deserialize` function for this type.

rjzamora:
I added a test here that serializes a numba array and then tries to deserialize into a cupy array.  It seems that this (strange and poossibly erroneous) use of serialization is the only way to run into the "strides" problem.

It seems that this type type of behavior is causing the problem in ucx-py.  @quasiben, Does it make sense that ucx-py might try to convert a cupy array into a numba device array between serialization and deserialization?

mrocklin:
I'm comfortable merging this in any time.  @quasiben @pentschev any objections?  

I'll merge in tomorrow morning my time if I don't hear anything back.

distributed/protocol/cupy.py
```diff
@@ -3,6 +3,18 @@
...
+from distutils.version import LooseVersion
+
+
+class PatchedCudaArrayInterface(object):
+    # TODO: This class wont be necessary
+    #       once Cupy<7.0 is no longer supported
+    def __init__(self, ary):
+        vsn = LooseVersion(cupy.__version__)
+        cai = ary.__cuda_array_interface__
+        if vsn < "7.0.0" and cai.get("strides") is None:
```
pentschev: Maybe makes more sense to check `ary.__cuda_array_interface__["version"] < 2` instead of `vsn < "7.0.0"`? The `__cuda_array_interface__["version"] == 2` is only introduced with CuPy 7.0.0rc1, so versions 7.0.0b4 and lower may not work properly (in case someone is using that).

pentschev:
> I'm comfortable merging this in any time. @quasiben @pentschev any objections?
> 
> I'll merge in tomorrow morning my time if I don't hear anything back.

I added one comment regarding CuPy/`__cuda_array_interface__`, but merging it even without that is ok by me, in case we're fine with telling any eventual users of the beta versions to upgrade.

distributed/protocol/cupy.py
rjzamora: Sure - My only concern here is that we would only want to pop the "strides" attribute if `__cuda_array_interface__["version"] == 2` **and** `vsn < "7.0.0"`.  The problem is that numba supplies a 2nd-version interface, but cupy won't support it until version 7.0.

rjzamora: Perhaps we should just check that `vsn < "7.0.0rc1"` (since we only care about the version of the interface if this is the case)?

pentschev: But if we only care about the version of the interface, why don't we check that instead of the version of CuPy? Anyway, either of these checks are _probably_ fine though, unless CuPy backports the `__cuda_array_interface__` which is unlikely.

rjzamora: > But if we only care about the version of the interface, why don't we check that instead of the version of CuPy?

I am probably just misunderstanding the suggestion.  However, I don't really see how we can avoid checking the cupy version.  The problem occurs when the interface version is "2", and the cupy version is < 7.0.0rc1 (note that the interface was not defined by a cupy object, but is now being used to create a cupy object).

pentschev: I hope this helps clarify what I mean:

```python
In [1]: import cupy

In [2]: cupy.__version__
Out[2]: '7.0.0rc1'

In [3]: cupy.ones(1000).__cuda_array_interface__["version"]
Out[3]: 2
```

```python
In [1]: import cupy

In [2]: cupy.__version__
Out[2]: '7.0.0b4'

In [3]: cupy.ones(1000).__cuda_array_interface__["version"]
Out[3]: 0
```

Only starting with CuPy 7.0.0rc1 is `__cuda_array_interface__["version"] == 2`, prior to that, it was on version 0. In practice, my suggestion was that the condition would be:

```python
if cay["version"] < 2 and cai.get("strides") is None:
```

But again, the way it's written now is more than fine and probably won't create any issues for anyone, so we can keep it as is.

rjzamora: Right - I think I am failing to clarify that the `__cuda_array_interface__` is **not** created by cupy in this case.  The interface is just being used to create a new cupy object.  Therefore, using `if cai["version"] < 2 and cai.get("strides") is None:` will **not** capture the problematic situation, because the interface version **is** 2  (the tests fail with that check in place).

pentschev: I understand now, thanks @rjzamora for the clarification here (and offline). So the issue is that Numba creates the `__cuda_array_interface__` that is later used for Numpy to create the object. Since Numba may be an older version (where `__cuda_array_interface__["version"] < 2"),  the check for CuPy version is needed instead.

pentschev:
Thanks for the fix and clarifications @rjzamora, we indeed need to check the CuPy version. From my side, it's good to be merged.

distributed/protocol/cupy.py
```diff
@@ -3,6 +3,18 @@
...
+from distutils.version import LooseVersion
+
+
+class PatchedCudaArrayInterface(object):
+    # TODO: This class wont be necessary
+    #       once Cupy<7.0 is no longer supported
+    def __init__(self, ary):
+        vsn = LooseVersion(cupy.__version__)
+        cai = ary.__cuda_array_interface__
+        if vsn < "7.0.0rc1" and cai.get("strides") is None:
```
mrocklin: Should this concern us?

```python
In [1]: "7.0.0" < "7.0.0rc1"
Out[1]: True
```

pentschev: Argh, I'd say so, we want "7.0.0" inclusive. What is normally used for the proper comparison in such cases? In the dask codebase I see things like the following:

```python
PANDAS_GT_0240 = PANDAS_VERSION >= LooseVersion("0.24.0rc1")
```

That seems to be the same kind of comparison, and I'm assuming that works?

rjzamora: Perhaps the newest revision is better?  I am creating an empty cupy object to check the *supported* `__cuda_array_interface__` version.  E.g:

```python
class PatchedCudaArrayInterface(object):
    # TODO: This class wont be necessary
    #       once Cupy<7.0 is no longer supported
    def __init__(self, ary):
        cai = ary.__cuda_array_interface__
        cai_cupy_vsn = cupy.ndarray(0).__cuda_array_interface__["version"]
        if cai.get("strides") is None and cai_cupy_vsn < 2:
            cai.pop("strides", None)
        self.__cuda_array_interface__ = cai
```

distributed/protocol/cupy.py
```diff
@@ -5,6 +5,17 @@
 from .cuda import cuda_serialize, cuda_deserialize
 
 
+class PatchedCudaArrayInterface(object):
+    # TODO: This class wont be necessary
+    #       once Cupy<7.0 is no longer supported
+    def __init__(self, ary):
+        cai = ary.__cuda_array_interface__
+        cai_cupy_vsn = cupy.ndarray(None).__cuda_array_interface__["version"]
```
pentschev: This is a good idea, but `cupy.ndarray(0)` is a better choice as it doesn't incur a memory allocation, whereas `cupy.ndarray(None)` does. The pointer for the first is a null pointer, the latter has a valid one, as seen below (and verifiable with nvprof):

```python
In [1]: import cupy

In [2]: cupy.ndarray(0).__cuda_array_interface__["data"]
Out[2]: (0, False)

In [3]: cupy.ndarray(None).__cuda_array_interface__["data"]
Out[3]: (140350232461312, False)
```

distributed/protocol/cupy.py
pentschev: That's a good idea indeed, thanks for coming up with this @rjzamora. I left a comment on improving this (or rather reverting or your latest change).

distributed/protocol/cupy.py
rjzamora: Thanks for pointing this out @pentschev !!   My mistake :)

pentschev: No worries, it's something I didn't know as well, but I decided to look if that would incur in any unwanted overhead. Thanks for the fix!

pentschev:
LGTM, please feel free to merge when CI passes.

rjzamora:
> LGTM, please feel free to merge when CI passes.

Thanks for your help with this @pentschev !

mrocklin:
This is in.  Thanks @rjzamora !

PR #3199, opened 2019-11-05
Use CPU count from dask

jrbourbeau:
In https://github.com/dask/dask/pull/5499 the existing `CPU_COUNT` from `distributed.system` was copied to `dask.system`. This is a follow-up PR which imports `CPU_COUNT` from `dask` to avoid duplicate code.

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -10,9 +10,9 @@ import warnings
 import click
 import dask
 from dask.utils import ignoring
+from dask.system import CPU_COUNT
 from distributed import Nanny, Worker
 from distributed.security import Security
-from distributed.system import CPU_COUNT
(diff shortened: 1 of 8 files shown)


mrocklin:
I suggest that we wait a release before doing this, just to create some version flexibility (unless we're already breaking for other reasons)

jrbourbeau:
That seems totally reasonable. I just wanted to open a PR with these changes before I forget we moved `CPU_COUNT` over to `dask/dask` : )

mrocklin:
Yup.  Makes sense to me.  Thanks @jrbourbeau !

jrbourbeau:
We've released version `2.7.0` (of `dask` and `distributed`) since this PR was opened. Should we merge this?

jrbourbeau:
Thanks @mrocklin!

PR #3200, opened 2019-11-06
Clear task stream based on recent behavior

mrocklin:
Alternative to #3190 cc @dickreuter

I tried to breifly implement my proposed solution to clearing the task stream based on recent behavior.  It only runs the check if we haven't seen an update in a suitable amount of time (which is nice for performance) and then bases the decision on the current timespan of the rectangles in the plot.  In principle this works fairly nicely. 

However, it does have a fail case when building up a larger and larger task stream plot over time.  If you start having these large gaps in the stream then things can get harder and harder to clear out.  Probably we should have some other check as well that attempts to understand the amount of whitespace in the plot currently.  Perhaps some measure of the sum of the durations / workers over the total timespan.  

@dickreuter I mostly wanted to share this to communicate what I was trying to say in the issue earlier.  Maybe it helps make my original intent more clear.

Screen Shot 2019-11-05 at 3 56 47 PM

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -916,7 +916,6 @@ class TaskStream(DashboardComponent):
                     sum(self.source.data["duration"])
                     / len(self.workers)
                     / (old_end - old_start)
-                    / 1000
                 )
 
                 # If whitespace is more than 3x the old width
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -912,11 +912,6 @@ class TaskStream(DashboardComponent):
                     )
                 )
 
-                from dask.utils import format_time
-
-                print("Whitespace", format_time(new_start - old_end))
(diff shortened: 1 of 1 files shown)


dickreuter:
Thanks for this. Yes I agree, your proposal basically does not clear out old runs if there are gaps. That's why I used (gap_to_previous > (avg_duration_of_visualized * self.clear_multiplier) > self.clear_interval*1000), which for my cases yielded better results. But I guess both solutions are better than what we currently have.

mrocklin:
Merging this in shortly if there are no objections.

mrocklin:
It looks like I was a factor of 1000 off in the density computation due to the s/ms difference.

Also while diving in with pbd it looks like "start" should really be renamed "middle".  A lot of the logic here is off by half of the duration value.

mrocklin:
We're nearing a release.  I'm going to go ahead and merge this in.  We can refine it in the future if necessary.

PR #3216, opened 2019-11-09
Add PipInstall WorkerPlugin

mrocklin:
Examples
--------

```python
>>> from dask.distributed import PipInstall
>>> plugin = PipInstall(packages=["dask", "scikit-learn"])
>>> client.register_worker_plugin(plugin)
```


mrocklin:
cc @TomAugspurger @rabernat @jhamman

distributed/diagnostics/plugin.py
```diff
@@ -126,3 +128,54 @@ def transition(self, key, start, finish, **kwargs):
...
         kwargs: More options passed when transitioning
         """
+
+
+class PipInstall(WorkerPlugin):
+    """ A Worker Plugin to pip install a set of packages
+
+    This accepts a set of packages to install on all workers.
+    You can also optionally ask for the worker to restart itself after
+    performing this installation.
```
TomAugspurger: I'd like to see a note of warning (here or elsewhere) that this is primarily intended for experimentation / interactive development. A more robust solution would be to include it in these in the actual worker environment (docker image, virtualenv/conda pack, etc.)

distributed/diagnostics/plugin.py
```diff
@@ -126,3 +128,54 @@ def transition(self, key, start, finish, **kwargs):
...
+            stdout, stderr = proc.communicate()
+
+            if self.restart and worker.nanny:
+                lines = stdout.strip().split(b"\n")
+                if not all(
+                    line.startswith(b"Requirement already satisfied") for line in lines
+                ):
+                    worker.loop.add_callback(
+                        worker.close_gracefully, restart=True
+                    )  # restart
```
TomAugspurger: Seems worth logging a warning that `restart=True` has no effect when there's no nanny. The user explicitly asked for it.

distributed/diagnostics/plugin.py
```diff
@@ -126,3 +128,54 @@ def transition(self, key, start, finish, **kwargs):
...
+            proc = subprocess.Popen(
+                [sys.executable, "-m", "pip", "install"] + self.packages,
+                stdout=subprocess.PIPE,
+            )
+            stdout, stderr = proc.communicate()
+
+            if self.restart and worker.nanny:
+                lines = stdout.strip().split(b"\n")
+                if not all(
+                    line.startswith(b"Requirement already satisfied") for line in lines
```
TomAugspurger: Feels a *bit* fragile, in that we're relying on pip to not change the output message. Perhaps an unnecessary optimization?

distributed/diagnostics/plugin.py
```diff
@@ -126,3 +128,54 @@ def transition(self, key, start, finish, **kwargs):
...
+    --------
+    >>> from dask.distributed import PipInstall
+    >>> plugin = PipInstall(packages=["dask", "scikit-learn"])
+
+    >>> client.register_worker_plugin(plugin)
+    """
+
+    name = "pip"
+
+    def __init__(self, packages=[], restart=False):
```
TomAugspurger: *Possibly* would want to accept an argument for options passed to pip. Things like `--upgrade`, `--pre`, etc.

So perhaps a `pip_options=None` that's a list of strings, just like in Popen.

distributed/diagnostics/plugin.py
TomAugspurger: On second thought, it might be nice to share the same workerplugin setup for different clusters.

TomAugspurger:
Pushed a few changes here.

@jcrist do you have any objections to the overall idea of doing this in a worker plugin?

mrocklin:
Conceptually there are issues here around restarting.  We want to restart
things if we have different versions than what we just installed. However
it's also possible that another process just updated our versions, and we
should restart, even though when we call `pip install` everything seems
like it is already installed.

kyprifog:
Whats the status of this effort?  I just ran into an issue running dask-gateway with not having s3fs on the workers and I'm trying to find the path of least resistence.

TomAugspurger:
I pushed a couple updates to

- fix merge conflicts
- add a test for an install failures (e.g. package typo)
- added the PipInstall plugin to the docs

I think this should be good to go.

mrocklin:
I think that there are likely still issues when you have multiple workers on the same machine pip/conda installing things.  However, this may be an uncommon enough case where this is likely to be used that we shouldn't care.  I'm not sure.  I think that it's totally ok to merge anyway.

kyprifog:
@mrocklin, to clarify are you saying there are issues with multiple clients trying to issue (possibly contradictory or overlapping) commands to the workers of the same cluster to update packages?

mrocklin:
I'm saying that if multiple workers are on the same filesystem (such as
happens when you say dask-worker --nprocs 2) then they will both try to
pip/conda install and both try to restart, sort of.  Things can fail easily
in that situation.

What you're suggesting is also an issue, but one that seems much less
likely to occur.

TomAugspurger:
>  I am still getting a package not found s3fs error from the workers:

I wonder if the installation / restart fails to complete before you start your computation requiring s3fs?

The issue with multiple workers sharing a filesystem is a good one I hadn't considered. I'm OK with just documenting that shortcoming for now.

kyprifog:
> > I am still getting a package not found s3fs error from the workers:
> 
> I wonder if the installation / restart fails to complete before you start your computation requiring s3fs?
> 
> The issue with multiple workers sharing a filesystem is a good one I hadn't considered. I'm OK with just documenting that shortcoming for now.

I had wondered the same thing but if that were the case running the same operation twice would result in a positive response.  I think maybe there is also a chance that the dask default configuration leverages conda so its possible that it needs to be using conda install instead of pip.   The command that finally worked (when run manually on each worker) was

```
conda install --yes -c conda-forge s3fs==0.4.2
```

So i had to use conda and downgrade s3fs (I was getting an error form fsspec about "other_paths" with 0.5.0 and 0.5.1)

For context, I'm trying to avoid forking the helm chart for daskhub which is why I'm spending more time trying to figure out how to do this via plugin.

kyprifog:
Update:  I was able to get my issue resolved by building my own conda install worker plugin.

TomAugspurger:
>  That being said, it didn't appear that the workers restarted in kubernetes

What do you mean by this? I don't believe this would be visible at the kubernetes level (other than via logs), since no pods / processes are going away.

kyprifog:
Yeah I meant via logs, I was running `stern dask` and non of the workers logged any changes.

TomAugspurger:
Gotcha, thanks for clarifying.

mrocklin:
+1

Issue #3111, opened 2019-10-03
Add CondaPipInstall WorkerPlugin


mrocklin:
Often users want to make small modifcations onto their existing environment.  For large changes, we recommend affecting docker images, NFS installations, and so on, but for small changes we often find people doing something like the following:

```python
def install_foo():
    os.system("pip install foo")

client.run(install_foo)
```

This can cause problems, particularly if workers restart, if new workers arise, or if the library is already around and imported into Python (and so the new version doesn't show up).  

We should probably provide a more robust solution for this.  I propose that we create a [`WorkerPlugin`](https://docs.dask.org/en/latest/setup/custom-startup.html#worker-lifecycle-plugins) that does this robustly.  Something like the following:

```python
class CondaPipInstall:
    def __init__(self, packages: List[Tuple[str, str]]):
        self.packages = packages

    def setup(self, worker):
        if any of the packages aren't up to date:
            install_packages(self.packages)
            restart_this_worker()

install = CondaPipInstall(packages=["numpy >= 1.17", "git+https://github.com/...@master"])
client.register_worker_plugin(install)
```

I think that this would be a good middle ground between `client.run` or `client.upload_file` and asking users to make new docker images.

(this came up in conversation with @jcrist and @TomAugspurger)

mrocklin:
A partial solution lives here: https://github.com/dask/distributed/pull/3216

PR #3217, opened 2019-11-09
Pip install packages from configuration in Nanny.start

mrocklin:
This is a different approach to #3216 and #3111 which places pip install packages in configuration.

This is mostly up here for conversation.  My guess is that @jcrist will likely veto it (which is entirely fine with me).  Mostly I wanted to talk about a few other options we have.

cc @jhamman @rabernat @TomAugspurger

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -66,6 +66,9 @@ distributed:
   client:
     heartbeat: 5s  # time between client heartbeats
 
+  nanny:
+    pip: []
+
   deploy:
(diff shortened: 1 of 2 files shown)


TomAugspurger:
I prefer https://github.com/dask/distributed/pull/3216 to adding packages in the config. I view these as helpers to tweak a base environment while iterating on something. If you're to the point where you're modifying the config, you should probably be rebuilding the image / environment :)

mrocklin:
OK.  Closing.

PR #3218, opened 2019-11-10
Use legend_field= keyword in bokeh plots

mrocklin:
The old legend= keyword has been deprecated

@bryevdv if you know off the top of your head, how old is the `legend_field=` keyword?  I'm curious how careful we need to be here with backwards compatibility.


bryevdv:
@mrocklin it is new as of 1.4

PR #3228, opened 2019-11-12
Fail fast when importing distributed.comm.ucx

mrocklin:
Fixes https://github.com/dask/dask/issues/5572

If we don't have UCX installed then we shouldn't try anything further.
In this commit we move the `import ucp` line to the top of the file.


mrocklin:
@bnaul does this solve things for you?

bnaul:
Works great :)

PR #3229, opened 2019-11-12
Avoid setting inf x_range

rockwellw:
When trying to view the dashboard, workers with no `memory_limit` lead to the following error:
`ValueError: Out of range float values are not JSON compliant`

It looks like this was fixed in #2770 , but has broken again since then.

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -585,7 +585,7 @@ class CurrentLoad(DashboardComponent):
                     or inf
                 )
 
-                if limit > max_limit:
+                if limit > max_limit and limit != inf:
                     max_limit = limit
 
                 if nb > limit:


mrocklin:
Thanks @rockwellw .  Do you have any suggestions on how we might test this change to avoid future breakages?

rockwellw:
@mrocklin it's not a very robust test, but we could catch this specific breakage by modifying `test_CurrentLoad` to something like: 
```
@gen_cluster(client=True)
def test_CurrentLoad(c, s, a, b):
    cl = CurrentLoad(s)
    cl.scheduler.workers.values()[0].memory_limit = None

    futures = c.map(slowinc, range(10), delay=0.001)
    yield wait(futures)

    cl.update()
    d = dict(cl.source.data)

    assert all(len(L) == 2 for L in d.values())
    assert all(d["nbytes"])
    assert type(cl.nbytes_figure.x_range.end) == int
```
Let me know if I should add this to the PR.

Also, this bug doesn't actually cause anything to crash, but it adds a ton of noisy error messages when trying to view the dashboard.

mrocklin:
I'm just going to go ahead and merge this in.  Thanks @rockwellw !

Issue #3232, opened 2019-11-13
Intermittent failure with test_nanny_throttle


mrocklin:
```
_____________________________ test_nanny_throttle ______________________________
    def test_func():
        result = None
        workers = []
        with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:
    
            async def coro():
                with dask.config.set(config):
                    s = False
                    for i in range(5):
                        try:
                            s, ws = await start_cluster(
                                nthreads,
                                scheduler,
                                loop,
                                security=security,
                                Worker=Worker,
                                scheduler_kwargs=scheduler_kwargs,
                                worker_kwargs=worker_kwargs,
                            )
                        except Exception as e:
                            logger.error(
                                "Failed to start gen_cluster, retrying",
                                exc_info=True,
                            )
                        else:
                            workers[:] = ws
                            args = [s] + workers
                            break
                    if s is False:
                        raise Exception("Could not start cluster")
                    if client:
                        c = await Client(
                            s.address,
                            loop=loop,
                            security=security,
                            asynchronous=True,
                            **client_kwargs
                        )
                        args = [c] + args
                    try:
                        future = func(*args)
                        if timeout:
                            future = gen.with_timeout(
                                timedelta(seconds=timeout), future
                            )
                        result = await future
                        if s.validate:
                            s.validate_state()
                    finally:
                        if client and c.status not in ("closing", "closed"):
                            await c._close(fast=s.status == "closed")
                        await end_cluster(s, workers)
                        await gen.with_timeout(
                            timedelta(seconds=1), cleanup_global_workers()
                        )
    
                    try:
                        c = await default_client()
                    except ValueError:
                        pass
                    else:
                        await c._close(fast=True)
    
                    for i in range(5):
                        if all(c.closed() for c in Comm._instances):
                            break
                        else:
                            await asyncio.sleep(0.05)
                    else:
                        L = [c for c in Comm._instances if not c.closed()]
                        Comm._instances.clear()
                        # raise ValueError("Unclosed Comms", L)
                        print("Unclosed Comms", L)
    
                    return result
    
            result = loop.run_sync(
>               coro, timeout=timeout * 2 if timeout else timeout
            )
distributed/utils_test.py:952: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[...]

PR #3235, opened 2019-11-14
Add name to Pub/Sub repr

mrocklin:


jrbourbeau:
Just noting that the test failure is unrelated and has been reported in #3232

PR #3239, opened 2019-11-15
Close attached clients as scheduler shuts down

mrocklin:
Fixes https://github.com/dask/dask-labextension/issues/85

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -704,6 +704,7 @@ class Client(Node):
             "task-erred": self._handle_task_erred,
             "restart": self._handle_restart,
             "error": self._handle_error,
+            "scheduler-shutdown": self.close,
         }
 
         self._state_handlers = {
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1275,6 +1275,9 @@ class Scheduler(ServerNode):
                 else:
                     break
 
+        for c in self.client_comms.values():
+            c.send({"op": "scheduler-shutdown"})
+
         for pc in self.periodic_callbacks.values():
             pc.stop()
         self.periodic_callbacks.clear()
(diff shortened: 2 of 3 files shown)


jcrist:
This would be useful for dask-gateway as well, to prevent lingering clients from spamming the proxy with reconnect attempts until they timeout.

jsignell:
Is this done? Does it just need a rebase?

jrbourbeau:
It looks like there were some failing tests too. I've resolved the existing merge conflicts, let's see what the CI looks like as of today

mrocklin:
```python
    @gen_cluster(client=True, nthreads=[], client_kwargs={"timeout": 0.5})
    def test_reconnect_timeout(c, s):
        with captured_logger(logging.getLogger("distributed.client")) as logger:
            yield s.close()
            start = time()
            while c.status != "closed":
                yield c._update_scheduler_info()
                yield gen.sleep(0.05)
                assert time() < start + 5, "Timeout waiting for reconnect to fail"
        text = logger.getvalue()
>       assert "Failed to reconnect" in text
E       AssertionError: assert 'Failed to reconnect' in 'Client report stream closed to scheduler\nReconnecting...\n'
```

It looks like we're failing to respect a timeout now.  If anyone has time to look into this failure that would be welcome.  My apologies for not finishing this off at the time.

TomAugspurger:
I think that test as written doesn't work now. It was relying on `Scheduler.close` not calling `Client.close`. I've added a parameter `close_clients` to `Scheduler.close`, primarily to make this easy to test. Is that OK?

TomAugspurger:
Tests are still failing here. I'm assuming they're related to the changes here, since we're potentially changing the number of clients that are created and closed. However I haven't been able to reproduce the failures locally here.

PR #3241, opened 2019-11-15
Use inspect.isawaitable where relevant

jcrist:
Noticed a few places in distributed where `hasattr(x, '__await__')` was checked. This should use `inspect.isawaitable` instead for readability/correctness. Updated accordingly.


mrocklin:
Thanks @jcrist !  Merging.

PR #3242, opened 2019-11-15
Remove `gen.coroutine` usage in scheduler

jcrist:
Use `async`/`await` and `asyncio` idioms throughout.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2006,9 +2006,8 @@ class Scheduler(ServerNode):
             return
         if ts is None or not ts.who_wants:  # no key yet, lets try again in a moment
             if retries:
-                self.loop.add_future(
-                    asyncio.sleep(0.2),
-                    lambda _: self.cancel_key(key, client, retries - 1),
+                self.loop.call_later(
+                    0.2, lambda: self.cancel_key(key, client, retries - 1)
                 )
             return
         if force or ts.who_wants == {cs}:  # no one else wants this key
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -24,7 +24,6 @@ except ImportError:
     from toolz import frequencies, merge, pluck, merge_sorted, first
 from toolz import valmap, second, compose, groupby
 from tornado import gen
-from tornado.gen import Return
 from tornado.ioloop import IOLoop
 
 import dask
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -2794,13 +2793,11 @@ def rebalance(self, comm=None, keys=None, workers=None):
                 },
             )
 
-            if not all(r["status"] == "OK" for r in result.values()):
-                raise Return(
-                    {
-                        "status": "missing-data",
-                        "keys": sum([r["keys"] for r in result if "keys" in r], []),
```
mrocklin: Was this wrong before? I would have expected this to iterate over result.values

mrocklin:
In principle everything here looks good to me.  Thanks for doing this @jcrist 

I'm a little curious why I didn't change some of these over in the past.  I wonder if there was some odd thing going on before.

distributed/scheduler.py
jcrist: Yeah, this was wrong before.

mrocklin: It's really good to have a second pair of eyes here

jcrist:
Looks like not all tests pass, so will have to dive deeper. Only ran a few checks locally since the test suite takes a while. In general what's your process with testing locally? Many things seem related, so its hard to know what tests are likely to fail.

mrocklin:
I usually run one of the files that I think is relevant, probably `test_core.py`, `test_scheduler.py`, or `test_client.py`.  I find some simple failing test in that group, hopefully one that is async, and then run that one repeatedly while i try to find the issue.  

As context, before we got very stringent about checking for thread/process/fd leaks before and after every test (the `cleanup` fixture) each of these used to take around 20s.  We might look into profiling our testing harness at some point.

mrocklin:
```python-traceback
  File "/home/travis/build/dask/distributed/distributed/scheduler.py", line 2012, in cancel_key
    lambda _: self.cancel_key(key, client, retries - 1),
  File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/tornado/ioloop.py", line 693, in add_future
    assert is_future(future)
```

It looks like this might be a relatively routine case of Tornado/Asyncio mismatch.  This is probably why I left that one there.  We could also try replacing `loop.add_callback/loop.add_future` with `asyncio.ensure_future` in places.

jcrist:
Yeah, I'm going to keep pushing on this until we can remove this aspect of tornado from the codebase. Just need to go deeper.

distributed/scheduler.py
```diff
@@ -2007,8 +2006,8 @@ def cancel_key(self, key, client, retries=5, force=False):
             return
         if ts is None or not ts.who_wants:  # no key yet, lets try again in a moment
             if retries:
-                self.loop.add_future(
-                    gen.sleep(0.2), lambda _: self.cancel_key(key, client, retries - 1)
+                self.loop.call_later(
+                    0.2, lambda _: self.cancel_key(key, client, retries - 1)
                 )
```
jcrist: Fixed. There's no reason to manually call `gen.sleep` here, using `IOLoop.call_later` works just as well.

jcrist:
All tests pass. Good to merge?

mrocklin:
Woot.  Thanks @jcrist !

PR #3245, opened 2019-11-18
More tornado cleanups

jcrist:
- Remove `gen.coroutine` from `worker.py`
- Remove tornado from `asyncprocess.py`. This spiraled out into some general cleanups/dead code removal for this file.

### DIFF
--- a/distributed/process.py
+++ b/distributed/process.py
@@ -1,31 +1,24 @@
+import asyncio
 import atexit
-from datetime import timedelta
 import logging
 import os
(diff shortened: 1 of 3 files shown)


jcrist:
Hmmm, looks like failures on tornado 5. Will investigate later.

mrocklin:
@jcrist are you planning to come back to this?  My guess is that the answer is "no", but I'm still holding out hope :)

jcrist:
Yes, sorry for dropping the ball here.

Issue #3184, opened 2019-10-30
Improve recovery time in worker failure scenarios


fjetter:
We are operating our distributed clusters in a cloud environment where we need to deal with frequently failing nodes. We usually dispatch jobs automatically and are bound to certain SLAs and therefore expect our jobs to finish in a more or less well defined time.
While distributed offers resilience in terms of graph recalculation we're facing the issue that the recalculation introduces severe performance issues for us.

We are looking for something which would allow us to recover faster in scenarios where individual workers die such that we do not need to recalculate large, expensive chunks of the graph, e.g. by persisting or replicating valuable, small intermediate results.

Ideally the solution would be handled by the scheduler itself, s.t. many different applications can benefit of it (e.g. via a scheduler plugin/extension). We were thinking about milestone/snapshotting where the user can label certain results to be worthy to be repliacated (and later forgotten once another milestone passes/completes). We also discussed some kind of automatic replication based on heuristics (e.g. `bytes_result < x and runtime of task < Y` -> replicate result) to soften the blow in case of failures.

My questions would be:

1. Does anybody have additional ideas we should take into account?
2. Could anybody already gather some experience with similar situations which might help us?
3. If we would start to implement something like this, what would be the best approach / where should we start?
4. Are we doing something fundamentally wrong? :)

Researching existing github issues, I only found #2748 which discusses this scenario briefly but is ultimately closed without a proper resolution to this topic. The only solution which is suggested is a caching library but persisting every single result is most likely not an option.

mrocklin:
Hi @fjetter ,

My apologies for the delay in response.  You raised this when I was fairly busy, and I hadn't gone through old issues until now.

Everything you say makes a lot of sense to me.  I think that Dask could really use an active memory manager that thinks about replication like how you describe.  It's unlikely that I personally will build such a thing, but I think that you and your group could handle it easily.

Here are some links to prior discussion that might be helpful in informing design.  

-  An old issue talking about retiring the current `replicate` method with a more active system like how you describe: https://github.com/dask/distributed/issues/1002
-  A very old post on opportunistic caching: https://matthewrocklin.com/blog/work/2015/08/03/Caching .  Also https://docs.dask.org/en/latest/caching.html

[...]

PR #3248, opened 2019-11-19
RFC Retire worker with periodic callback

fjetter:
## Context

We're currently experimenting with graceful termination of workers and see the issue that if we terminate multiple workers at once (or closely after each others) the replication logic will replicate to workers which are about to retire themselves, i.e. not ideal / prone to errors.

While the _proper_ solution would be a redesign of the replication logic as briefly discussed #3184 and #1002, the change in this PR could help in the meantime.

## Changes

Instead of immediately retiring upon calling `retire_workers`, only the intend to retire is logged, i.e. a list of workers "about to retire" is filled. A periodic callback will then mop up the intended retirements batch wise. The batch wise retirement should help with replication since all workers which are about to be retired won't be included in the replication attempt.

## PR state

The code is not properly cleaned up and probably a lot of test are still failing. If it is considered a feasible approach, I'll clean this up properly and add proper documentation.

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -24,6 +24,7 @@ distributed:
     preload: []
     preload-argv: []
     default-task-durations: {}  # How long we expect function names to run ("1h", "1s") (helps for long tasks)
+    worker-retirement-interval: 10s
     dashboard:
       status:
         task-stream-length: 1000
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1146,6 +1146,14 @@ class Scheduler(ServerNode):
         for ext in extensions:
             ext(self)
 
+        self._workers_to_retire = []
+        self.retire_workers_active = False
+        self.periodic_callbacks["retire-workers"] = PeriodicCallback(
+            self._retire_workers,
+            parse_timedelta(dask.config.get("distributed.scheduler.worker-retirement-interval")),
+            io_loop=loop
+        )
+
         setproctitle("dask-scheduler [not started]")
         Scheduler._instances.add(self)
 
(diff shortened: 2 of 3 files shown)


mrocklin:
Do these options satisfy your needs ?

https://docs.dask.org/en/latest/setup/cli.html#cmdoption-dask-worker-lifetime

https://github.com/dask/distributed/pull/2892

(this could use better docs)

fjetter:
The worker lifetime rather works like: If a worker is up for <worker-lifetime>s shut it down.
If I understand this correctly, this is rather to "recycle" workers and ensure that the uptime is not too long, isn't it?

We're rather looking for something where we can shutdown workers gracefully to free their resources if a cluster manager decides that they are needed elsewhere. Essentially a forced scale down in an adaptive deployment scenario.

fjetter:
Amazing response time btw :)

mrocklin:
Ah, I see.  Sorry I misread your original post (going too quickly through e-mail in the morning).  

If this is only about fixing the replicate and retire_workers interactions, and if it increases the complexity of `retire_workers` then I think I'm generally against this.  I prefer to avoid making things things more complex to work around replicate being bad.

fjetter:
The issue we want to address is indeed caused by the replication logic. However, the fundamental idea is to handle graceful worker shutdowns which are triggered on worker side better. This may currently happen in two scenarios:

1. An external handler calls the `close_gracefully` method of the worker (our use case)
2. Worker `lifetime` is configured which causes the workers to shut down after a set amount of time.

Both scenarios cannot issue multiple retirements at once on scheduler side which this change would address. Especially for the `lifetime` feature I would imagine this to happen frequently but I do not have any experience with it since we're not using it.

The assumption is that it is beneficial for the scheduler to handle multiple, almost simultaneous retirement requests in a synchronised/batched manner. Currently, the only _serious_ work the scheduler needs to due to perform the retirement is the replication which is why the issue essentially boils down to this. Are there any intentions on extending the retirement logic?

If this is not considered an issue, I can understand and we can close this issue.

mrocklin:
What do you think about adding a lock around some of these operations?  

https://github.com/dask/distributed/pull/3259

PR #3249, opened 2019-11-19
Add support for Python 3.8

jrbourbeau:
See the corresponding `dask/dask` PR at https://github.com/dask/dask/pull/5603


TomAugspurger:
Seems like a slight issue with stacktrace:

```
UnsatisfiableError: The following specifications were found

to be incompatible with the existing python installation in your environment:

Specifications:

  - stacktrace -> python[version='>=3.5,<3.6.0a0|>=3.6,<3.7.0a0|>=3.7,<3.8.0a0']

Your python: python=3.8

If python is on the left-most side of the chain, that's the version you've asked for.

When python appears to the right, that indicates that the thing on the left is somehow

not available for the python version you are constrained to. Note that conda will not

change your python version to a different minor version unless you explicitly specify

that.

The following specifications were found to be incompatible with each other:

Package setuptools conflicts for:

stacktrace -> python[version='>=3.6,<3.7.0a0'] -> pip -> setuptools

python=3.8 -> pip -> setuptools

Package certifi conflicts for:

stacktrace -> python[version='>=3.6,<3.7.0a0'] -> pip -> setuptools -> certifi[version='>=2016.09|>=2016.9.26']

python=3.8 -> pip -> setuptools -> certifi[version='>=2016.09|>=2016.9.26']

Package pip conflicts for:

python=3.8 -> pip

stacktrace -> python[version='>=3.6,<3.7.0a0'] -> pip

Package wheel conflicts for:

stacktrace -> python[version='>=3.6,<3.7.0a0'] -> pip -> wheel

python=3.8 -> pip -> wheel

Package ca-certificates conflicts for:

stacktrace -> python[version='>=3.6,<3.7.0a0'] -> openssl=1.0 -> ca-certificates

python=3.8 -> openssl[version='>=1.1.1d,<1.1.2a'] -> ca-certificates

```

I think we can remove that from the 3.8 build.

jrbourbeau:
Thanks for reviewing @TomAugspurger. Yeah, `stacktrace` isn't available for Python 3.8 yet. Just pushed a commit which skips the install for 3.8

TomAugspurger:
Failure on 3.8 in 

```
_________________________ test_cleanup_repeated_tasks __________________________

...
>       assert not list(ws)

E       AssertionError: assert not [<test_steal.test_cleanup_repeated_tasks.<locals>.Foo object at 0x7f60b5e8daf0>, <test_steal.test_cleanup_repeated_tasks.<locals>.Foo object at 0x7f606b665ca0>]

E        +  where [<test_steal.test_cleanup_repeated_tasks.<locals>.Foo object at 0x7f60b5e8daf0>, <test_steal.test_cleanup_repeated_tasks.<locals>.Foo object at 0x7f606b665ca0>] = list({<weakref at 0x7f6062ab9bd0; to 'Foo' at 0x7f60b5e8daf0>, <weakref at 0x7f6062ab9680; to 'Foo' at 0x7f606b665ca0>})
```


I don't recall seeing that before.

TomAugspurger:
Restarted the 3.8 build to check, though I think we'll want to avoid adding new flaky tests :)

jrbourbeau:
CI passed. Restarting again

PMeira:
I noticed the AppVeyor/Windows tests only include Python 3.6, so this doesn't address the Tornado situation with Python 3.8 on Windows, right? Context: https://github.com/tornadoweb/tornado/issues/2608

I'm testing Distributed with Python 3.8 on Windows (most packages from conda-forge) and it seems to be working fine for my usage (1 thread per worker) after adjusting the event loop policy. To do that, I added...
```
        if sys.platform == 'win32':
            asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
...right after:
https://github.com/dask/distributed/blob/b6fb54154e8805ffe8fbec80bcb5ee7f2d24f328/distributed/utils.py#L1194-L1198

(not sure if it's the best place for that or even if it's generic enough)

jrbourbeau:
There was a Python 3.8-related fix to tornado in version 6.0.3 (https://github.com/tornadoweb/tornado/pull/2683). This causes only one of our tests, `distributed/deploy/tests/test_local.py::test_silent_startup`, to fail and I've `xfail`ed it here.

However, when using tornado < 6.0.3 on Python 3.8 there are many `asyncio.exceptions.CancelledError` tracebacks logged to the screen:

```
Exception in callback with_timeout.<locals>.error_callback(<Future cancelled>) at /Users/jbourbeau/miniconda/envs/dist-3.8-tornado-5/lib/python3.8/site-packages/tornado/gen.py:968
handle: <Handle with_timeout.<locals>.error_callback(<Future cancelled>) at /Users/jbourbeau/miniconda/envs/dist-3.8-tornado-5/lib/python3.8/site-packages/tornado/gen.py:968>
  File "/Users/jbourbeau/miniconda/envs/dist-3.8-tornado-5/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/jbourbeau/miniconda/envs/dist-3.8-tornado-5/lib/python3.8/site-packages/tornado/gen.py", line 970, in error_callback
    future.result()
asyncio.exceptions.CancelledError
```

Are there any objections to bumping our minimum tornado version to 6.0.3 (for reference, this version was released on Jun 22, 2019)? Not sure if this is too aggressive a bump for most users cc @mrocklin

TomAugspurger:
@jrbourbeau I *think* there are way to do version-specific requirements: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies

It'll complicate our setup.py / conda recipe a bit, but is probably worth it for tornado, maybe?

mrocklin:
I've run into some systems that still don't react well to Tornado 6.  For example I think that our dask-examples binder deployment might not work due to some Jupyter/Tornado issues.  I could be wrong though.

mrocklin:
This might also be good motivation to replace `gen.with_timeout` with the asyncio equivalent as you suggested in #3367 ?

mrocklin:
In general I find that `test_silent_startup` is valuable to have active if we can manage it.  Seeing a bunch of warning messages pop up during usage is distracting, and gives a sense that we're less mature of a project than we are.

jrbourbeau:
> This might also be good motivation to replace gen.with_timeout with the asyncio equivalent as you suggested in #3367 ?

Yeah, that would help us avoid the tornado issue. I opened up #3372 to see how the test suite handles the `gen.with_timeout` -> `asyncio.wait_for` switch

> In general I find that test_silent_startup is valuable to have active if we can manage it

Agreed

mrocklin:
With the asyncio/tornado change are things better now, or is there still more to do?

mrocklin:
I pushed a small commit that I hope should help.  It looks like I accidentally allowed a merge commit to get in there as well.  My apologies.

jrbourbeau:
In Python 3.8, `asyncio.CancelledError` was changed from being a subclass of `concurrent.futures.CancelledError` to a subclass of `BaseException` (xref https://github.com/python/cpython/pull/13528). This means that before when throwing and catching `CancelledError`s, both `asyncio.CancelledError` and `concurrent.futures.CancelledError` would be caught by `except concurrent.futures.CancelledError`. With Python 3.8+ this is not longer the case so we need to be a little more careful about how we handle `CancelledError`s.

Luckily, almost all the occurrences of `CancelledError` in the codebase are intended to be `concurrent.futures.CancelledError`. So I've added `concurrent.futures.CancelledError` to the `distributed.utils` namespace so we can standardize on it throughout the project (similar to the `TimeoutError` standardization in https://github.com/dask/distributed/pull/3394). Moving forward, if a `CancelledError` from `asyncio` is needed we should explicitly use `asyncio.CancelledError`.

mrocklin:
That seems sensible to me.  Thank you for diving into this @jrbourbeau

jrbourbeau:
Np, it's a subtle change that's good to know about

A few other notes:

- I've updated the title of this PR to be a `WIP` to indicate we should hold off on merging until the [2.10.0 release](https://github.com/dask/community/issues/33) is out in the next few days. That'll give these changes some time to simmer in `master`

- This PR should be ready for a review

- I'm going to re-run the CI builds a few times to identify possible flaky tests (all green checks at the time of writing this comment)

distributed/tests/test_steal.py
```diff
@@ -650,6 +655,9 @@ class Foo:
     assert not s.who_has
     assert not any(s.has_what.values())
 
+    import gc
```
TomAugspurger: Should this be removed?

distributed/utils.py
```diff
@@ -1211,11 +1211,28 @@ def reset_logger_locks():
...
-            import tornado.platform.asyncio
 
-            asyncio.set_event_loop_policy(
-                tornado.platform.asyncio.AnyThreadEventLoopPolicy()
-            )
+            BaseEventLoopPolicy = asyncio.DefaultEventLoopPolicy
+            if WINDOWS and sys.version_info >= (3, 8):
+                # WindowsProactorEventLoopPolicy is not compatible with tornado 6
+                # fallback to the pre-3.8 default of Selector
+                # https://github.com/tornadoweb/tornado/issues/2608
```
TomAugspurger: From what I understand, tornado decided not to do this for users by default. Did Jupyter decided whether or not to do it?

Is appropriate for Dask to me making this decision?

jrbourbeau: Tornado decided not to set the `asyncio.WindowsSelectorEventLoopPolicy()` by default and instead recommends projects do that themselves. Jupyter and several other projects have done this (e.g. https://github.com/jupyter/notebook/pull/5047).

Our situation is a little different because we want to set the event loop policy to tornado's `AnyThreadEventLoopPolicy` to let us create loops inside of any thread. Tornado has decided to set the base class for `AnyThreadEventLoopPolicy` to be `asyncio.WindowsSelectorEventLoopPolicy` on Windows (https://github.com/tornadoweb/tornado/commit/d2af6a6fbe57a4d3bc06fbde6f4c1f4e8ec06dd3), but there hasn't been a release with the those changes yet so we're doing that ourselves here. I'll add a note that we can remove this once the next tornado release is out.

distributed/tests/test_steal.py
jrbourbeau: Ah yes, thanks! I added that to try and fix this test, but it didn't do the trick

jrbourbeau:
CI is passing 🎉 

For reference, there was a patch made to the conda-forge tornado feedstock that fixed the tornado event loop policy issue with Python 3.8 on Windows. Unfortunately the patch also broke things for us (hence the most recent CI failure). Fortunately, other packages have fixed things internally and the patch to the conda-forge recipe was recently removed (xref https://github.com/conda-forge/tornado-feedstock/pull/36). It's worth noting that the same conda recipe patch was also made to tornado in the default channel and is still in place, so that's why we're installing `tornado` from conda-forge in CI now

mrocklin:
Woo!

Issue #2880, opened 2019-07-24
Timed out trying to connect to host after 10 s


seanlaw:
I have a dask distributed cluster up and running on 40 workers:

```
dask_client = Client('localhost:8786')
dask_client.restart()
dask_client
```
I've restarted everything so no tasks are queued and the scheduler log shows:
    
    distributed.scheduler - INFO - Clear task state

I have a large csr sparse matrix that I am scattering to the cluster:
    
    csr_future = dask_client.scatter(csr, broadcast=True)

After a few seconds, I see:

```
distributed.scheduler - INFO - Remove worker tcp://10.157.169.65:38615
distributed.core - INFO - Removing comms to tcp://10.157.169.65:38615
distributed.scheduler - INFO - Remove worker tcp://10.157.169.65:33352
distributed.core - INFO - Removing comms to tcp://10.157.169.65:33352
distributed.scheduler - INFO - Register tcp://10.157.169.65:38051
distributed.scheduler - INFO - Starting worker compute stream, tcp://10.157.169.65:38051
distributed.core - INFO - Starting established connection
distributed.scheduler - INFO - Register tcp://10.157.169.65:46414
distributed.scheduler - INFO - Starting worker compute stream, tcp://10.157.169.65:46414
distributed.core - INFO - Starting established connection
```

So, it looks like some workers are being removed and new workers are being added back to replace those workers. Around 30 seconds after this, I see multiple tornado errors:

```
tornado.application - ERROR - Multiple exceptions in yield list
Traceback (most recent call last):
  File "/app/home/miniconda3/lib/python3.7/site-packages/distributed/comm/core.py", line 218, in connect
    quiet_exceptions=EnvironmentError,
  File "/app/home/miniconda3/lib/python3.7/site-packages/tornado/gen.py", line 729, in run
    value = future.result()
tornado.util.TimeoutError: Timeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/home/miniconda3/lib/python3.7/site-packages/tornado/gen.py", line 501, in callback
    result_list.append(f.result())
  File "/app/home/miniconda3/lib/python3.7/site-packages/tornado/gen.py", line 736, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/app/home/miniconda3/lib/python3.7/site-packages/distributed/core.py", line 736, in send_recv_from_rpc
    comm = yield self.pool.connect(self.addr)
[...]

PR #3250, opened 2019-11-19
Fix hanging worker when the scheduler leaves

TomAugspurger:
cc @mrocklin if you have a chance to review.

Closes https://github.com/dask/distributed/issues/2880

### DIFF
--- a/distributed/tests/test_nanny.py
+++ b/distributed/tests/test_nanny.py
@@ -1,3 +1,4 @@
+import asyncio
 import gc
 import logging
 import os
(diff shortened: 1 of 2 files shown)


distributed/tests/test_nanny.py
```diff
@@ -129,6 +129,20 @@ def test_run(s):
     yield n.close()
 
 
+@pytest.mark.slow
+@gen_cluster(config={"distributed.comm.timeouts.connect": "1s"})
+async def test_no_hang_when_scheduler_closes(s, a, b):
+    # https://github.com/dask/distributed/issues/2880
+    with captured_logger("tornado.application", logging.ERROR) as logger:
+        await s.close()
+        await gen.sleep(1)
```
mrocklin:
```suggestion
        await asyncio.sleep(1.2)
```

mrocklin: Should we wait a little bit longer to avoid a race condition?

PR #3255, opened 2019-11-21
Skip numba.cuda tests if CUDA is not available

### LINKED ISSUE #3254: CuPy serialization test not skipped when Numba is installed
If I run `pytest distributed/protocol/tests/test_numba.py` with `numba` installed, but without a GPU or `cudatoolkit` installed, I get the following error:

```
E           numba.cuda.cudadrv.error.CudaSupportError: Error at driver init:
E
E           CUDA driver library cannot be found.
E           If you are sure that a CUDA driver is installed,
E           try setting environment variable NUMBA_CUDA_DRIVER
E           with the file path of the CUDA driver shared library.
```

This is because we're using `importorskip` on `numba.cuda`

https://github.com/dask/distributed/blob/face9e8273ef755aa658d89d8f3dd80b1c02cd7c/distributed/protocol/tests/test_numba.py#L4

which can still be `import`ed even if `cudatoolkit` isn't installed:

```python
In [1]: import numba

In [2]: numba.__version__
Out[2]: '0.45.1'

In [3]: import numba.cuda as cuda

In [4]: cuda.is_available()
Out[4]: False
```

cc @quasiben @pentschev if you have any thoughts on a better condition we can use to skip this test. Perhaps skipping if `nvvm` or `cupy` aren't installed?

pentschev:
Fixes #3254

### DIFF
--- a/distributed/protocol/tests/test_numba.py
+++ b/distributed/protocol/tests/test_numba.py
@@ -7,6 +7,9 @@ np = pytest.importorskip("numpy")
 
 @pytest.mark.parametrize("dtype", ["u1", "u4", "u8", "f4"])
 def test_serialize_cupy(dtype):
+    if not cuda.is_available():
+        pytest.skip("CUDA is not available")
+
     ary = np.arange(100, dtype=dtype)
     x = cuda.to_device(ary)
     header, frames = serialize(x, serializers=("cuda", "dask", "pickle"))


TomAugspurger:
Thanks!

jrbourbeau:
Thanks @pentschev

jakirkham:
Adding this to the other Numba test in PR ( https://github.com/dask/distributed/pull/3446 ).

PR #3260, opened 2019-11-22
Update function serialization caches

jrbourbeau:
This PR switches to using `functools.lru_cache` for caching function serialization (and is currently used for the deserialization cache) on workers instead of `zict.LRU`. This still limits the cache from growing indefinitely but avoids an edge case where `zict.LRU` can end up in an inconsistent state (discussed in https://github.com/dask/zict/issues/29).

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1,6 +1,6 @@
 import asyncio
 import atexit
-from collections import deque
+from collections import deque, OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from datetime import timedelta
@@ -1399,3 +1399,23 @@ def deserialize_for_cli(data):
         The de-serialized data
     """
     return json.loads(base64.urlsafe_b64decode(data.encode()).decode())
+
+
+class LRU(OrderedDict):
+    """ Limited size mapping, evicting the least recently looked-up key when full
+    """
+
+    def __init__(self, maxsize):
+        self.maxsize = maxsize
+        super().__init__()
+
+    def __getitem__(self, key):
+        value = super().__getitem__(key)
+        self.move_to_end(key)
+        return value
+
+    def __setitem__(self, key, value):
+        if len(self) >= self.maxsize:
+            oldest = next(iter(self))
+            del self[oldest]
+        super().__setitem__(key, value)
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1,6 +1,6 @@
 import asyncio
 import atexit
-from collections import deque, OrderedDict
+from collections import deque, OrderedDict, UserDict
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from datetime import timedelta
@@ -1401,21 +1401,21 @@ def deserialize_for_cli(data):
     return json.loads(base64.urlsafe_b64decode(data.encode()).decode())
 
 
-class LRU(OrderedDict):
+class LRU(UserDict):
     """ Limited size mapping, evicting the least recently looked-up key when full
     """
 
     def __init__(self, maxsize):
-        self.maxsize = maxsize
         super().__init__()
+        self.data = OrderedDict()
+        self.maxsize = maxsize
 
     def __getitem__(self, key):
         value = super().__getitem__(key)
-        self.move_to_end(key)
+        self.data.move_to_end(key)
         return value
 
     def __setitem__(self, key, value):
         if len(self) >= self.maxsize:
-            oldest = next(iter(self))
-            del self[oldest]
+            self.data.popitem(last=False)
         super().__setitem__(key, value)
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -3,7 +3,6 @@ import bisect
 from collections import defaultdict, deque, namedtuple
 from collections.abc import MutableMapping
 from datetime import timedelta
-import functools
 import heapq
 from inspect import isawaitable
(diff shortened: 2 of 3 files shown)


distributed/worker.py
```diff
@@ -3187,23 +3186,15 @@ def execute_task(task):
...
+    return pickle.dumps(func)
 
 
 def dumps_function(func):
     """ Dump a function to bytes, cache functions """
     try:
-        result = cache[func]
-    except KeyError:
-        result = pickle.dumps(func)
-        if len(result) < 100000:
```
mrocklin: I think that we still need to respect this limit and only cache small functions.  Some user functions get pretty pathological.

jrbourbeau: Sure, I can see that being a nice feature to have. Do we want to include a similar condition on the function deserialization cache?

mrocklin: Hrm, indeed.  I didn't see that we weren't doing that there.  It would be good to add, but given that we're not doing it in one place already if you wanted to skip and be done then that would seem reasonable.

jrbourbeau: It's not possible to manually remove an item from the `functools.lru_cache` cache, so I added a small `LRU` class to `distributed.utils` which should avoid going into an inconsistent state if a `KeyboardInterrupt` is raised. 

Also, using an `LRU` `maxsize` of 100 + only caching small functions sets an upper bound on the size of the cache to be 10 MB (which is the current upper limit)

mrocklin:
@jcrist , you might be interested in reviewing this

distributed/utils.py
```diff
@@ -1399,3 +1399,23 @@ def deserialize_for_cli(data):
...
+
+    def __getitem__(self, key):
+        value = super().__getitem__(key)
+        self.move_to_end(key)
+        return value
+
+    def __setitem__(self, key, value):
+        if len(self) >= self.maxsize:
+            oldest = next(iter(self))
+            del self[oldest]
```
jcrist: Can use `self.popitem(last=False)` here instead.

distributed/worker.py
```diff
@@ -3144,18 +3145,25 @@ def parse_memory_limit(memory_limit, nthreads, total_cores=CPU_COUNT):
...
-def cached_function_deserialization(func):
-    return pickle.loads(func)
+
+def loads_function(bytes_object):
+    """ Load a function from bytes, cache bytes """
+    try:
+        result = cache_loads[bytes_object]
+    except KeyError:
+        result = pickle.loads(bytes_object)
+        if len(bytes_object) < 100000:
```
jcrist: Should we move this check to before the cache check? Since large keys will never be stored, the cache check will always fail.

```python
def loads_function(bytes_object):
    if bytes_object < 100000:
        try:
            result = cache_loads[bytes_object]
        except KeyError:
            cache_loads[bytes_object] = result = pickle.loads(bytes_object)
        return result
    return pickle.loads(bytes_object)
```

jrbourbeau: Yeah, good point

distributed/utils.py
jrbourbeau: Hmm I initially thought that would work, but I'm getting a `test_lru` failure locally (traceback below). It looks like `popitem` internally calls `__getitem__` which we're overloading here to include a call to `move_to_end`, which how some methods are implemented in `OrderedDict`.

I just pushed a commit which updates `LRU` to be a subclass of [`UserDict`](https://docs.python.org/3.8/library/collections.html#userdict-objects) and sets the underlying `.data` mapping to be an `OrderedDict`. This lets us still keep track or the order items are used, but avoids breaking other mapping methods like `LRU.popitem`

<details>
<summary>test_lru traceback:</summary>

```
============================================================ FAILURES ============================================================
____________________________________________________________ test_lru ____________________________________________________________

    def test_lru():

        l = LRU(maxsize=3)
        l["a"] = 1
        l["b"] = 2
        l["c"] = 3
        assert list(l.keys()) == ["a", "b", "c"]

        # Use "a" and ensure it becomes the last recently used item
        l["a"]
        assert list(l.keys()) == ["b", "c", "a"]

        # Ensure maxsize is respected
>       l["d"] = 4

distributed/tests/test_utils.py:617:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
distributed/utils.py:1427: in __setitem__
    self.popitem(last=False)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = LRU([('c', 3), ('a', 1)]), key = 'b'

    def __getitem__(self, key):
        value = super().__getitem__(key)
>       self.move_to_end(key)
E       KeyError: 'b'

distributed/utils.py:1414: KeyError
```

</details>

jrbourbeau:
The `test_workspace_concurrency` and `test_nanny_throttle` failures are known and unrelated to the changes here

mrocklin:
Thanks @jrbourbeau !  This is in.

jrbourbeau:
Thanks for reviewing @mrocklin @jcrist

PR #3261, opened 2019-11-23
Use `DeviceBuffer` from newer RMM releases

jakirkham:
Depends on PR ( https://github.com/rapidsai/rmm/pull/177 )

If a newer version of RMM is around, use `DeviceBuffer` instead. Otherwise fallback to `device_array`. This is significantly faster to allocate.

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -46,7 +46,10 @@ def init_once():
     try:
         import rmm
 
-        cuda_array = lambda n: rmm.device_array(n, dtype=np.uint8)
+        if hasattr(rmm, "DeviceBuffer"):
+            cuda_array = lambda n: rmm.DeviceBuffer(size=n)
+        else:  # pre-0.12.0
+            cuda_array = lambda n: rmm.device_array(n, dtype=np.uint8)
     except ImportError:
         try:
             import numba.cuda


jakirkham:
I'm not sure what all features of `device_array` are being used by `distributed`. So it would be good to get some context on that as well. There may be more PRs we want to add to `rmm` to make this viable.

jakirkham:
cc @quasiben (for awareness)

mrocklin:
cc also @madsbk

mrocklin:
Happy to press the merge button any time.  It looks like this is still pending upstream changes, so presumably we should wait for that.

quasiben:
I think we should wait until https://github.com/rapidsai/rmm/pull/177 is merged in.  @jakirkham may be gone for another week so I don't think there is any rush here

quasiben:
@mrocklin https://github.com/rapidsai/rmm/pull/177 was just merged in so I think this should be good to go now

jakirkham:
cc @kkraus14 (for awareness 🙂)

distributed/comm/ucx.py
```diff
@@ -46,7 +46,10 @@ def init_once():
     try:
         import rmm
 
-        cuda_array = lambda n: rmm.device_array(n, dtype=np.uint8)
+        if hasattr(rmm, "DeviceBuffer"):
+            cuda_array = lambda n: rmm.DeviceBuffer(size=n)
+        else:  # pre-0.12.0
```
jakirkham: Should be `pre-0.11.0`. Fixing in PR ( https://github.com/dask/distributed/pull/3299 ).

PR #3264, opened 2019-11-25
Fix dev requirements for pytest.

QuLogic:
The minimum version specified in `setup.cfg` is newer than `dev-requirements.txt`.


TomAugspurger:
Thanks!

PR #3268, opened 2019-11-26
Fix layout scaling on profile plots

### LINKED ISSUE #3206: Use CSS for profile and profile server plots
These plots currently only function well when they are of a certain aspect ratio.  It may be that Bokeh's layout engine isn't cut out for them.  If someone with some CSS knowledge is around, it would be great to have these operate more like the /status route (see `dashboard/components/scheduler.py::status_doc` for pointers) so that it conformed to the space that it was in.

cc @jacobtomlinson

jacobtomlinson:
This PR changes the Bokeh layouts to ensure plots a responsive with page size but some elements are fixed height (buttons and lower plot).

Fixes #3206

ezgif com-optimize

### Before
image

### After

image

### Before

image

### After

image

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1852,7 +1852,7 @@ def individual_bandwidth_workers_doc(scheduler, extra, doc):
 def profile_doc(scheduler, extra, doc):
     with log_errors():
         doc.title = "Dask: Profile"
-        prof = ProfileTimePlot(scheduler, sizing_mode="scale_width", doc=doc)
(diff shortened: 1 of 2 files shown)


mrocklin:
Whoa, that was all that was necessary?  I thought that this was going to be hard because the buttons were there.  I'm very glad to see this!

jacobtomlinson:
Me too, and I started down that path, but after learning a bunch about the Bokeh layout system it just turned out to be this :).

mrocklin:
Thanks @jacobtomlinson !

PR #3278, opened 2019-11-27
Make Listener.start asynchronous

mrocklin:
Previously the Listener.start method was synchronous,
expecting comms to be able to start themselves and get an address
immediately.  This was a challenge when we tried adding asyncio comms,
which needed to await the server creation.

Now we make Listener.start asynchronous, and make sure that we await all
calls to it within the codebase.

FYI @jcrist


jcrist:
Conceptually this seems fine to me, haven't looked through the implementation though.

mrocklin:
Yeah, there is a bit of noise in this PR because of all the yield -> await changes.  Mostly, the change here is that Listener.start is now asynchronous.  You are, I think, possibly the only person around today who has a listener outside of the main codebase, with dask-gateway.

jcrist:
We actually only need to implement a `Connector` in dask-gateway, since we're only dispatching on the client side: https://github.com/dask/dask-gateway/blob/7b8f9306b593154817b9f364ec0651c9ef86d17c/dask-gateway/dask_gateway/comm.py#L30. Thanks for the heads up though.

PR #3279, opened 2019-11-27
xfail ucx test

quasiben:
cc @madsbk @mrocklin


quasiben:
xref: https://github.com/rapidsai/cudf/issues/3481

mrocklin:
I'm happy to merge this if desired, but is it possible to resolve it upstream instead of xfailing it here?

quasiben:
I think we should xfail for now so we can test the ucx functionality in distributed.

PR #3288, opened 2019-11-30
Support multiple listeners in the scheduler

mrocklin:
This allows the scheduler to listen on multiple different listeners at
once. This can be useful when the client and workers are on different
interfaces, or have different security concerns.

So far this isn't documented. 

cc @jcrist @jacobtomlinson for review

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -136,7 +136,7 @@ class Server(object):
         self._ongoing_coroutines = weakref.WeakSet()
         self._event_finished = Event()
 
-        self.listener = None
+        self.listeners = []
         self.io_loop = io_loop or IOLoop.current()
         self.loop = self.io_loop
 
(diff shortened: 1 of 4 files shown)


distributed/core.py
```diff
@@ -221,15 +221,22 @@ def start_pcs():
...
                 # See https://github.com/tornadoweb/tornado/issues/2069
-                self.io_loop.add_callback(self.listener.stop)
+                self.io_loop.add_callback(listener.stop)
+
+    @property
+    def listener(self):
+        if self.listeners:
+            return self.listeners[0]
+        else:
+            return None
```
jacobtomlinson: Should this raise some kind of warning as it will limit folks to only accessing one listener in the case that they have multiple listeners.

mrocklin: I think that this is likely ok?  In principle we should stop using `listener` any longer, and should always iterate over listeners.  This is here mostly for backwards compatibility and convenience.  Given that this is likely to be ok 99% of the time I'm inclined not to raise a warning.

PR #3289, opened 2019-11-30
Fix distributed.wait documentation

tomzx:
The current documentation for `distributed.wait` is slightly broken (see https://docs.dask.org/en/latest/futures.html#distributed.wait). This fixes the return not being rendered properly and adds a bit of documentation for the undocumented parameter `return_when`.


mrocklin:
Thanks @tomzx !

PR #3292, opened 2019-12-02
more ucx fixes

quasiben:
These are the rest of the fixes for UCX testing in distributed 

cc @mrocklin @madsbk

### DIFF
--- a/distributed/comm/tests/test_ucx.py
+++ b/distributed/comm/tests/test_ucx.py
@@ -36,13 +36,12 @@ async def get_comm_pair(
         await q.put(comm)
 
     listener = listen(listen_addr, handle_comm, connection_args=listen_args, **kwargs)
-    await listener.start()
-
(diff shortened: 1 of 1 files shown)


distributed/comm/tests/test_ucx.py
```diff
@@ -36,12 +41,13 @@ def test_registered():
         await q.put(comm)
 
     listener = listen(listen_addr, handle_comm, connection_args=listen_args, **kwargs)
-    with listener:
-        comm = await connect(
-            listener.contact_address, connection_args=connect_args, **kwargs
-        )
-        serv_com = await q.get()
-        return comm, serv_com
+    await listener.start()
```
mrocklin: You probably want `async with listener` here instead

quasiben: Thanks!  Resolved

PR #3293, opened 2019-12-02
Avoid repeatedly adding same dependency to already in memory stack

jrbourbeau:
This PR updates the logic in `update_graph` to avoid repeatedly adding the same dependency to the already in memory `stack`. This improves our performance when there are many tasks that share dependencies which are already in `memory` or `erred`.

Below is a small example to demonstrate the performance before and after these changes:

<details>
<summary>Example snippet:</summary>

```python
import time
import random

from dask import delayed, compute

if __name__ == "__main__":

    random.seed(2)

    with Client(n_workers=1, threads_per_worker=1) as client:

        # Perform same calculation for increasing graph size
        for n in [50, 100, 150, 200]:
            # Generate 10 indepdent tasks
            x = [delayed(random.random)() for _ in range(10)]
            # Generate lots of interrelated dependent tasks
            for _ in range(10, n):
                random_subset = [random.choice(x) for _ in range(5)]
                random_max = delayed(max)(random_subset)
                x.append(random_max)

            # Persist tasks into distributed memory and wait to finish
            y = client.persist(x)
            wait(y)

            # Time computation
            t_start = time.time()
            compute(*x)
            dt = time.time() - t_start
            print(f"(n={n}) time taken = {dt:0.3f} sec")
```

</details>

<details>
<summary>Output on current master:</summary>

```
(n=50) time taken = 0.043 sec
(n=100) time taken = 0.683 sec
(n=150) time taken = 3.092 sec
(n=200) time taken = 11.783 sec
```

</details>

<details>
<summary>Output with changes here:</summary>

```
(n=50) time taken = 0.018 sec
(n=100) time taken = 0.030 sec
(n=150) time taken = 0.046 sec
(n=200) time taken = 0.061 sec
```

</details>

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1608,7 +1608,7 @@ class Scheduler(ServerNode):
                     else:
                         child_deps = self.dependencies[dep]
                     if all(d in done for d in child_deps):
-                        if dep in self.tasks:
+                        if dep in self.tasks and dep not in done:
                             done.add(dep)
                             stack.append(dep)
 


mrocklin:
Cool.  Thanks for the performance fix @jrbourbeau !

jrbourbeau:
Thanks for reviewing!

jrbourbeau:
I'd like to add a test to try to avoid a future performance regression here if possible. To avoid introducing a flaky test, it would be great to test something more directly than asserting an example snippet runs in a certain amount of time. 

I was thinking about adding a debug-level logging message when a dependency is added to the `done` set and asserting that we only ever add a given dependency once. Here's a quick estimate of the overhead that would introduce:

```python
In [1]: import logging

In [2]: logger = logging.getLogger(__name__)

In [3]: a = 1

In [4]: %timeit logger.debug("a is {}".format(a))
650 ns ± 4.63 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

@mrocklin do you think that 650 ns is a blocker for such a test? I'm trying to get a sense for if this type of performance test is suitable for the `distributed` test suite, or if it's more appropriate for `dask-benchmarks`

mrocklin:
Is that 650 ns per task?  Our general per-task budget is around 200us, so this is about 0.3%.  It's not huge, but these do add up over time.  Over time I've pretty much cleared out all of the `logger.debug` elements from the inner loop of the scheduler for this reason.  They end up taking more time than you might expect.  If this is just for a test then I suggest that no, we don't include this.

As an alternative, I wonder if this would be a good time to start pushing on the dask-benchmarks?

jrbourbeau:
> Is that 650 ns per task?

It would be 650 ns per task that's has all it's child deps in memory. So it's possible the overall cost of such a logging call could become large

>  Over time I've pretty much cleared out all of the logger.debug elements from the inner loop of the scheduler for this reason

Yeah, I noticed a there weren't many `logger.debug` calls in this portion of the codebase and figured it was worth asking about. Thanks for the feedback! 

I think we'd benefit greatly from more benchmarks. I'll go ahead and add a benchmarks to the `dask-benchmarks` repo for this instead

PR #3301, opened 2019-12-04
Connectionpool: don't hand out closed connections

byjott:
Operating long-running dask clusters (sometimes, they run for many days without interruptions), we found that connection issues we observe are likely related to a behavior of the ConnectionPool of handing out connections that are already closed by the remote end (e.g. because a connection has been established some days ago, easily above connection timeouts).

It also fixes another bug in the connecitonpool's bookkeeping of connections.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -832,6 +832,9 @@ class ConnectionPool(object):
         self._created = weakref.WeakSet()
         self._instances.add(self)
 
+    def validate(self):
+        assert self.semaphore._value == self.limit - self.open - self._n_connecting
+
     @property
     def active(self):
         return sum(map(len, self.occupied.values()))
@@ -860,9 +863,11 @@ class ConnectionPool(object):
         """
         available = self.available[addr]
         occupied = self.occupied[addr]
-        if available:
+        while available:
             comm = available.pop()
-            if not comm.closed():
+            if comm.closed():
+                self.semaphore.release()
+            else:
                 occupied.add(comm)
                 return comm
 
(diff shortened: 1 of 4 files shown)


distributed/core.py
```diff
@@ -832,6 +832,9 @@ def __init__(
         self._created = weakref.WeakSet()
         self._instances.add(self)
 
+    def validate(self):
```
jcrist: Since this is only used for testing, I'd make this `_validate`, and add a docstring for it explaining what it tests.

mrocklin:
In principle this seems fine to me.  I'm not a great expert on the comms these days though.  Thanks @byjott for the work and @jcrist for the review.

distributed/core.py
byjott: Fixed it

distributed/tests/test_client.py
```diff
@@ -3532,9 +3532,6 @@ def test_reconnect(loop):
         assert time() < start + 5
         sleep(0.01)
 
-    with pytest.raises(Exception):
```
byjott: Unfortunately, this test failed, and this change made it work again. What seemed to happen is that now the client detects connections issues earlier than it used to (yay), and tries *here* to reconnect to the scheduler. This can never succeed, as there is no scheduler process, so the client eventually gives up completely (c.state is "closed").
I think it only used to work this way because the client tried to use the (closed) scheduler conneciton, triggering an exception here instead of re-trying to reconnect until the timeout is reached.

 I think the new behavior is fine.

mrocklin:
This is in.  Thanks @byjott !

PR #3302, opened 2019-12-04
Skip Security.temporary() tests if cryptography not installed

jrbourbeau:
The use of `Security.temporary()` in `test_temporary_credentials` requires the `cryptography` package to be installed

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -74,8 +74,6 @@ if [[ $CRICK == true ]]; then
     pip install -q git+https://github.com/jcrist/crick.git
 fi;
 
-conda uninstall --force cryptography
-
 # Install distributed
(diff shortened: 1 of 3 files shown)


continuous_integration/travis/install.sh
```diff
@@ -74,6 +74,8 @@ if [[ $CRICK == true ]]; then
     pip install -q git+https://github.com/jcrist/crick.git
 fi;
 
+conda uninstall --force cryptography
```
jcrist: Why this? Leftover from testing?

jrbourbeau: Yeah, I'll remove that shortly. That's a temporary change to catch any other tests that should have `pytest.importorskip("cryptography")`

jrbourbeau:
Thanks for reviewing Jim!

PR #3304, opened 2019-12-05
Make ConnectionPool.close asynchronous

mrocklin:
Previously we would call a hard abort rather than waiting for comms to
close more gracefully.

FYI @jcrist @madsbk

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -951,11 +951,13 @@ class ConnectionPool(object):
 
     async def close(self):
         """
-        Close all communications abruptly.
+        Close all communications
         """
         for d in [self.available, self.occupied]:
             comms = [comm for comms in d.values() for comm in comms]
-            await asyncio.gather(*[comm.close() for comm in comms])
+            await asyncio.gather(
+                *[comm.close() for comm in comms], return_exceptions=True
+            )
             for _ in comms:
                 self.semaphore.release()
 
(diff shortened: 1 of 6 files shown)


mrocklin:
Merging this tomorrow morning if there are no objections

madsbk:
I can confirm that it works for me on a fully updated stack.

distributed/core.py
```diff
@@ -949,18 +949,15 @@ def remove(self, addr):
...
+    async def close(self):
         """
         Close all communications abruptly.
         """
-        for comms in self.available.values():
-            for comm in comms:
-                comm.abort()
+        for d in [self.available, self.occupied]:
+            comms = [comm for comms in d.values() for comm in comms]
+            await asyncio.gather(*[comm.close() for comm in comms])
```
byjott: Currently, this does not seem to release the semaphore in case `comm.close` raises. I think one can make i safer by using `return_exceptions=True` in `gather`.
(OTOH, if `close` really means this instance should never be used again, it does not matter, of course. However, in that case, it would be nice to be more explicit about that and fail in case one tries to use a "closed" pool).

mrocklin: Thanks @byjott !  I've added the `return_exceptions=True` keyword, which I agree is likely a good idea.  In practice you're correct in guessing that this is only used at process shutdown, but it doesn't hurt.

I haven't yet added any sort of check on reuse of a closed connection pool.  If it's ok I'd like to pass on this until it comes up (which I suspect it won't)

PR #3305, opened 2019-12-08
Add plugin entry point for out-of-tree comms library

sodre:
This PR adds the ability to include external comms plugins without
having to directly access comm.backends.

It has the advantage to not require a library to be preloaded in order
to work, e.g. dask-spec does not have preloading support.

### DIFF
--- a/distributed/comm/registry.py
+++ b/distributed/comm/registry.py
@@ -60,14 +60,8 @@ backends = {}
 def get_backend(scheme):
     """
     Get the Backend instance for the given *scheme*.
-
-    New backends can be registered using setuptools entry_points, e.g.:
-    setup(..., entry_points={
-            'distributed.comm.backends': [
-                'udp=dask_udp.backend:UDPBackend',
-            ]
-        },
-    )
+    It looks for matching scheme in dask's internal cache, and falls-back to
+    package metadata for the group name ``distributed.comm.backends``
     """
 
     backend = backends.get(scheme)
--- a/distributed/comm/registry.py
+++ b/distributed/comm/registry.py
@@ -1,7 +1,5 @@
 from abc import ABC, abstractmethod
 
-import pkg_resources
-
 
 class Backend(ABC):
     """
@@ -72,17 +70,22 @@ def get_backend(scheme):
     )
     """
 
-    backend = backends.get(scheme) or next(
-        iter(
-            backend_class_ep.load()()
-            for backend_class_ep in pkg_resources.iter_entry_points(
-                "distributed.comm.backends", scheme
-            )
-        ),
-        None,
-    )
+    backend = backends.get(scheme)
     if backend is None:
-        raise ValueError(
-            "unknown address scheme %r (known schemes: %s)" % (scheme, sorted(backends))
+        import pkg_resources
+
+        backend = next(
+            iter(
+                backend_class_ep.load()()
+                for backend_class_ep in pkg_resources.iter_entry_points(
+                    "distributed.comm.backends", scheme
+                )
+            ),
+            None,
         )
+        if backend is None:
+            raise ValueError(
+                "unknown address scheme %r (known schemes: %s)"
+                % (scheme, sorted(backends))
+            )
     return backend
--- a/distributed/comm/registry.py
+++ b/distributed/comm/registry.py
@@ -1,5 +1,7 @@
 from abc import ABC, abstractmethod
 
+import pkg_resources
+
 
 class Backend(ABC):
     """
@@ -60,8 +62,25 @@ backends = {}
 def get_backend(scheme):
     """
     Get the Backend instance for the given *scheme*.
+
+    New backends can be registered using setuptools entry_points, e.g.:
+    setup(..., entry_points={
+            'distributed.comm.backends': [
+                'udp=dask_udp.backend:UDPBackend',
+            ]
+        },
+    )
     """
-    backend = backends.get(scheme)
+
+    backend = backends.get(scheme) or next(
+        iter(
+            backend_class_ep.load()()
+            for backend_class_ep in pkg_resources.iter_entry_points(
+                "distributed.comm.backends", scheme
+            )
+        ),
+        None,
+    )
     if backend is None:
         raise ValueError(
             "unknown address scheme %r (known schemes: %s)" % (scheme, sorted(backends))
--- a/distributed/comm/registry.py
+++ b/distributed/comm/registry.py
@@ -82,4 +82,6 @@ def get_backend(scheme):
                 "unknown address scheme %r (known schemes: %s)"
                 % (scheme, sorted(backends))
             )
+        else:
+            backends[scheme] = backend
     return backend
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -1,15 +1,18 @@
 import asyncio
+import types
 from functools import partial
 import os
 import sys
 import threading
 import warnings
 
+import pkg_resources
 import pytest
 
 from tornado import ioloop, locks, queues
 from tornado.concurrent import Future
 
+import distributed
 from distributed.metrics import time
 from distributed.utils import get_ip, get_ipv6
 from distributed.utils_test import (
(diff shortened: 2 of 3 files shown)


mrocklin:
I don't know much about entry points unfortunately.  cc'ing @jakirkham , who I suspect knows more and may be able to quickly review this.

It looks like linting errors are coming up in CI.  @sodre you may want to take a look at https://docs.dask.org/en/latest/develop.html#code-formatting

sodre:
@mrocklin, a good overview of supporting plugins in python is [here.]( https://packaging.python.org/guides/creating-and-discovering-plugins/). I will fix the linting issues later today or once I get some feedback from @jakirkham.
 
I find that the least opinionated way is with `entry-points`. It allows the plugin-writer to put the comms backend in their own namespace without any further constraints. It would allow one to release new versions of a comms backend without having to cut a new release of distributed.

distributed/comm/registry.py
```diff
@@ -60,8 +62,25 @@ def get_local_address_for(self, loc):
...
-    backend = backends.get(scheme)
+
+    backend = backends.get(scheme) or next(
+        iter(
+            backend_class_ep.load()()
+            for backend_class_ep in pkg_resources.iter_entry_points(
+                "distributed.comm.backends", scheme
+            )
+        ),
+        None,
```
sodre: This is picking the "first" plugin that is registered with "scheme", in case one exists, otherwise it returns `None`.

jcrist:
I recommend using [entrypoints](https://github.com/takluyver/entrypoints) instead of pkg_resources. pkg_resources has some inefficient behavior (see https://github.com/takluyver/entrypoints/pull/37/). Jupyter, pandas, intake, etc... have all settled on using `entrypoints` instead of `pkg_resources` for finding entrypoints.

TomAugspurger:
FYI, pandas is using pkg_resources but only because we didn't wnat to add `entrypoints` as a dependency.

TomAugspurger:
Can you also add a test for this? https://github.com/pandas-dev/pandas/blob/2470690b9f0826a8feb426927694fa3500c3e8d2/pandas/tests/plotting/test_backend.py#L11-L38 has an example using `types.ModuleType` that may be helpful.

sodre:
@TomAugspurger I added the requested unit-tests. thanks for pointing me to an example.

@jcrist, Are you okay sticking with pkg_resources after @TomAugspurger comment? I am fine either way.

TomAugspurger:
FWIW, I'm fine with pkg_resources if it's imported inside the `get_backend` function. It takes ~200ms to import on my machine the first time.

Or we can just use entrypoints. Distributed already has a decent number of dependencies. I don't have a strong preference.

sodre:
I updated the code to import `pkg_resources` only if the backend is not found in the existing registry. 

I did a version of the code using `entrypoints`. Unfortunately, the unit-tests did not pass unless I mocked the entrypoints call. 

@jcrist, let me know if you feel strongly about pkg_resources and the additional 200ms delay.

jcrist:
I'm ok with the current implementation (importing only if not found), as this shouldn't affect any existing users. I do think the approach used in entrypoints is better than `pkg_resources`, but don't feel strongly enough about it at this point to ask you to make a change. I trust your judgement here.

sodre:
@jcrist, thank you. I have an implementation using entrypoints, but I did not like the unit-test for it. I will open a separate PR for comparison.

TomAugspurger:
For now my preference is to just use `pkg_resources`. If we start using entrypoints in more places, I'm fine to add `entrypoints` as a dependency.

@jakirkham, should the `distributed` feedstock be updated to include `setuptools` as a host and runtime dependency then, given that we import setuptools in our setup.py?

https://github.com/dask/dask/blob/1c39d48b452d857418c854008932fdac2608e121/setup.py#L5

https://github.com/conda-forge/distributed-feedstock/blob/master/recipe/meta.yaml#L20-L38

distributed/comm/registry.py
```diff
@@ -60,10 +60,32 @@ def get_local_address_for(self, loc):
 def get_backend(scheme):
     """
     Get the Backend instance for the given *scheme*.
+
+    New backends can be registered using setuptools entry_points, e.g.:
```
TomAugspurger: It seems that `get_backend` isn't included in `docs/source/communications.rst`. Could you add it and a `.. autofunction` for it?

distributed/comm/registry.py
```diff
@@ -60,10 +60,32 @@ def get_local_address_for(self, loc):
...
+                    "distributed.comm.backends", scheme
+                )
+            ),
+            None,
         )
+        if backend is None:
+            raise ValueError(
+                "unknown address scheme %r (known schemes: %s)"
+                % (scheme, sorted(backends))
+            )
```
TomAugspurger: I'm not sure how often `get_backend` is called, but I think we should mutate `bcakends` as a side effect of finding the backend from an entrypoint. Future calls will look it up from the cache.

distributed/comm/registry.py
```diff
@@ -60,10 +60,32 @@ def get_local_address_for(self, loc):
...
     """
     Get the Backend instance for the given *scheme*.
+
+    New backends can be registered using setuptools entry_points, e.g.:
+    setup(..., entry_points={
+            'distributed.comm.backends': [
+                'udp=dask_udp.backend:UDPBackend',
+            ]
+        },
+    )
```
TomAugspurger: Perhaps show the usage after this. It'd be

```python
>>> get_backend("udp")  # doctest: +SKIP
```

right?

distributed/comm/registry.py
```diff
@@ -60,10 +60,32 @@ def get_local_address_for(self, loc):
...
     """
+
     backend = backends.get(scheme)
     if backend is None:
-        raise ValueError(
-            "unknown address scheme %r (known schemes: %s)" % (scheme, sorted(backends))
+        import pkg_resources
+
+        backend = next(
```
TomAugspurger: Do you know how multiple entrypoints with the same name handled?

distributed/comm/registry.py
sodre: I moved the documentation to `communications.rst`

distributed/comm/registry.py
sodre: If two projects register the same name scheme, then the current code will pick the first one as far loaded  dictated by system path.

distributed/comm/registry.py
sodre: I think the calls to get_backend are all internal to distributed at the moment. I think the documentation in communications.rst should be sufficient at this point.  But let me know if you still want the usage in here.

distributed/comm/registry.py
sodre: This is taken care of ⬇️

distributed/comm/registry.py
TomAugspurger: Are you worried at all about "stale entrypoints" being left around when you upgrade a library providing an entrypoint (perhaps while developing?) Perhaps warn if there are unused entrypoints? I may be being paranoid here.

sodre: I am not at the moment. In the future, we could provide a simple script to list all the entry points registered for dask. For now, I don't think it is an issue.

TomAugspurger: Sounds good to me.

TomAugspurger:
Thanks @sodre!

jakirkham:
Thanks for working on this @sodre! This seems really useful (cc @quasiben, @madsbk, @pentschev 😉). Sorry for being somewhat absent here.

Adding `pkg_resources` seems like a reasonable thing to do for what you are using it for here.

It would be nice to move away from it when we can, but that may take more time. Python has slowly been adding functionality (along with backport packages) to move users away from `pkg_resources`. Though they haven't handled the entry point use case AFAIAA. So this remains reasonable for now. That said, Python core devs are aware of this use case. So would expect them to address it in the future. It's just a longer time horizon than we are thinking about ATM.

Also it seems you figured this out already, but adding `pkg_resources` does mean adding a runtime dependency on `setuptools`. It use to be the case that this was a bit problematic from a packaging perspective, but I think this was resolved some time ago. As long as we are clear about the dependency (as we have been), that's all that is needed. So thanks for handling that.

sodre:
@jakirkham, I made an alternative to this PR in #3314 that used the `entrypoints` package instead of `pkg_resources`. 

When the next extension point to DASK comes up it will be an ideal time to convert to the `entrypoints` library.

jakirkham:
Good to know. I think we can stick to `pkg_resources` for the time being.

PR #3307, opened 2019-12-09
All scheduler task states in prometheus

fjetter:
This exposes *all* task states as a prometheus metric. The current behaviour only captures the scheduler state partially and uses descriptions which are different than what is used internally. I believe the state machine of the cluster is already complex enough and using different names for the states only introduces confusion.

This also better aligns with what is visible on the bokeh dashboard.

If backwards compatibility is of concern we could still export the old ones or define a new metrics name, of course. I'm wondering how the whole prometheus metrics are supposed to be treated in terms of versioning/stability since IMHO some of them are _should_ probably be changed to offer real value (I haven't opened issues, yet)

cc @darindf are the received/unrunnable states something you rely upon?

A note regarding the test:
I wanted to add a proper unit test for this but am struggling to get it running on travis. I'm facing the usual problem "works on my machine" but on travis the metrics fetch *always* returns an empty metric (tried different approaches with waiting and scraping multiple times already but couldn't succeed)
Is there anything I'm doing wrong here or is there something I could try to get it running? I do not like changing stuff without a test and my creativity is running low on this one so a hint would be appreciated.

### DIFF
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -4,8 +4,12 @@ import logging
 
 import dask
 from dask.utils import format_bytes
-import toolz
-from toolz import merge
+
+try:
+    from cytoolz import merge, merge_with
+except ImportError:
+    from toolz import merge, merge_with
+
 from tornado import escape
 
 try:
@@ -42,6 +46,7 @@ from .worker import counters_doc
 from .proxy import GlobalProxyHandler
 from .utils import RequestHandler, redirect
 from ..utils import log_errors, format_time
+from ..scheduler import ALL_TASK_STATES
 
 
 ns = {
@@ -65,7 +70,7 @@ class Workers(RequestHandler):
                 "workers.html",
                 title="Workers",
                 scheduler=self.server,
-                **toolz.merge(self.server.__dict__, ns, self.extra, rel_path_statics),
+                **merge(self.server.__dict__, ns, self.extra, rel_path_statics),
             )
 
 
@@ -81,7 +86,7 @@ class Worker(RequestHandler):
                 title="Worker: " + worker,
                 scheduler=self.server,
                 Worker=worker,
-                **toolz.merge(self.server.__dict__, ns, self.extra, rel_path_statics),
+                **merge(self.server.__dict__, ns, self.extra, rel_path_statics),
             )
 
 
@@ -97,7 +102,7 @@ class Task(RequestHandler):
                 title="Task: " + task,
                 Task=task,
                 scheduler=self.server,
-                **toolz.merge(self.server.__dict__, ns, self.extra, rel_path_statics),
+                **merge(self.server.__dict__, ns, self.extra, rel_path_statics),
             )
 
 
@@ -109,7 +114,7 @@ class Logs(RequestHandler):
                 "logs.html",
                 title="Logs",
                 logs=logs,
-                **toolz.merge(self.extra, rel_path_statics),
+                **merge(self.extra, rel_path_statics),
             )
 
 
@@ -123,7 +128,7 @@ class WorkerLogs(RequestHandler):
                 "logs.html",
                 title="Logs: " + worker,
                 logs=logs,
-                **toolz.merge(self.extra, rel_path_statics),
+                **merge(self.extra, rel_path_statics),
             )
 
 
@@ -137,7 +142,7 @@ class WorkerCallStacks(RequestHandler):
                 "call-stack.html",
                 title="Call Stacks: " + worker,
                 call_stack=call_stack,
-                **toolz.merge(self.extra, rel_path_statics),
+                **merge(self.extra, rel_path_statics),
             )
 
 
@@ -156,7 +161,7 @@ class TaskCallStack(RequestHandler):
                     "call-stack.html",
                     title="Call Stack: " + key,
                     call_stack=call_stack,
-                    **toolz.merge(self.extra, rel_path_statics),
+                    **merge(self.extra, rel_path_statics),
                 )
 
 
@@ -239,7 +244,7 @@ class _PrometheusCollector(object):
         self.server = server
 
     def collect(self):
-        from prometheus_client.core import GaugeMetricFamily
+        from prometheus_client.core import GaugeMetricFamily, CounterMetricFamily
 
         yield GaugeMetricFamily(
             "dask_scheduler_clients",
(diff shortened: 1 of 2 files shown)


distributed/dashboard/tests/test_scheduler_bokeh_html.py
```diff
@@ -119,10 +120,50 @@ def test_prometheus(c, s, a, b):
...
+    families = {}
+
+    await asyncio.sleep(0)
+
+    # We may need to scrape multiple times to allow the prometheus
+
+    bokeh_scheduler = s.services["dashboard"]
+    response = await http_client.fetch(
+        f"http://{bokeh_scheduler.server.address}:{bokeh_scheduler.port}/metrics"
+    )
```
mrocklin: Perhaps the future hasn't yet had a chance to reach the scheduler?

Maybe something like this:

```python
future = client.submit(slowinc, 1, delay=0.5)

while not s.tasks:
   await asyncio.sleep(0.01)
```

Or, if you can allow for the future to be finished by the time you call this, then maybe ...

```python
await future
```

distributed/dashboard/tests/test_scheduler_bokeh_html.py
```diff
@@ -75,10 +76,10 @@ def test_worker_404(c, s):
         "services": {("dashboard", 0): (BokehScheduler, {"prefix": "/foo"})}
     },
 )
-def test_prefix(c, s, a, b):
+async def test_prefix(c, s, a, b):
```
mrocklin: Thank you for asyncifying things here

distributed/dashboard/scheduler.py
```diff
@@ -268,8 +270,9 @@ def collect(self):
             "Number of tasks known by scheduler.",
             labels=["state"],
         )
-        tasks.add_metric(["received"], len(self.server.tasks))
-        tasks.add_metric(["unrunnable"], len(self.server.unrunnable))
+        task_counter = Counter(map(lambda x: x.state, self.server.tasks.values()))
+        for state, num_tasks in task_counter.items():
+            tasks.add_metric([state], num_tasks)
```
mrocklin: This should become simpler if we merge in https://github.com/dask/distributed/pull/3262

Do you have any interest in reviewing that PR?  It would be good to get someone else's eyes on it.

distributed/dashboard/scheduler.py
```diff
@@ -268,8 +270,9 @@ def collect(self):
             "Number of tasks known by scheduler.",
             labels=["state"],
         )
-        tasks.add_metric(["received"], len(self.server.tasks))
-        tasks.add_metric(["unrunnable"], len(self.server.unrunnable))
+        task_counter = Counter(map(lambda x: x.state, self.server.tasks.values()))
```
mrocklin: If we do do it this way (rather than the PR mentioned below) then we might want to use `(cy)toolz.frequencies`, which is a bit faster.

distributed/dashboard/scheduler.py
mrocklin: In particular, I prefer to avoid looping over all tasks if we can.  It can really slow things down when we have millions of them.

distributed/dashboard/scheduler.py
StephanErb: @fjetter just to double check: Is that really a counter in the Prometheus sense (i.e. only goes up, ever goes down unless it rests to 0 due to integer overflow or server restart)? 

Given the way I understand Dask I would assume it is only a gauge (i.e. can go up or go down)

distributed/dashboard/tests/test_scheduler_bokeh_html.py
fjetter: I'll give it a shot

distributed/dashboard/scheduler.py
fjetter: >  just to double check: Is that really a counter in the Prometheus sense 

It's a `GaugeMetric`. The `Counter` is the python collections container to calculate the frequencies of the individual states.

What's _missing_ here is that we probably want to expose all existing states instead of only those currently existing, i.e. we should ensure there is a value for each state in `ALL_TASK_STATES` which makes the prometheus metric more consistent.

mrocklin: This becomes pretty easy (and much cheaper) once we merge in https://github.com/dask/distributed/pull/3262/

distributed/dashboard/tests/test_scheduler_bokeh_html.py
fjetter: Unfortunately, this also didn't work. There are scheduler tasks before and after the fetch but somehow it isn't picked up :(

fjetter:
I will hold off with this until #3262 is merged. The changes there look quite useful in general and allows to implement this change cleanly

distributed/dashboard/scheduler.py
```diff
@@ -268,8 +272,19 @@ def collect(self):
...
-        tasks.add_metric(["received"], len(self.server.tasks))
-        tasks.add_metric(["unrunnable"], len(self.server.unrunnable))
+
+        task_counter = merge_with(
+            sum, (tp.states for tp in self.server.task_prefixes.values())
+        )
+
+        yield CounterMetricFamily(
+            "dask_scheduler_tasks_forgotten",
+            "Number of tasks already forgotten by scheduler.",
```
mrocklin: I wonder if something like"total processed tasks would be more informative?"

We can still pop off the forgotten value at the end, but I suspect that people would be interested in the total of memory + forgotten + erred or something similar, more than just forgotten.

distributed/dashboard/scheduler.py
mrocklin: Ah, if we're doing this per state anyway then why treat forgotten differently?

distributed/dashboard/scheduler.py
fjetter: > I wonder if something like"total processed tasks would be more informative?"

As far as I understand, the approach with prometheus should rather be to export raw metric data. Interpretation, aggregation, etc. can then be done by the user. With these metrics everything is exposed and can be combined as needed. The way I see it, the bokeh dashboards are there to provide "easily understandable" information with context but for prometheus I would rather go for "raw data".

The reasoning for the metrics here is that the gauge `dask_scheduler_tasks` describes what the scheduler currently actually cares about and what it is working with. A more appropriate name would maybe be `dask_scheduler_active_tasks` but I'm also fine to stick to the current name.
By removing the `forgotten` state from this gauge there is also a nice invariant `sum(dask_scheduler_tasks) == len(scheduler.tasks)` and the possibility to infer progress with these states.

I split off the `forgotten` state since it is technically not a gauge in the prometheus sense (increasing *and* decreasing) but rather a count (only increasing). Ultimately it's a measure of how much work the scheduler has done since it was started.

distributed/dashboard/scheduler.py
```diff
@@ -253,40 +258,53 @@ def collect(self):
...
     def __init__(self, *args, **kwargs):
         import prometheus_client
 
         super(PrometheusHandler, self).__init__(*args, **kwargs)
 
-        if PrometheusHandler._initialized:
+        if PrometheusHandler._collector:
+            # Especially during testing, multiple schedulers are started
+            # sequentially in the same python process
+            PrometheusHandler._collector.server = self.server
```
fjetter: This is why the tests didn't run on travis. I only ran one of the prometheus tests on my machine but travis obviously executes all. The first test would register a collector with its scheduler. Every subsequent test would then receive the request handler for the scheduler of the first test instead of the current one.

This is a rather pragmatic fix since I expect these situations not to actually happen in real scenarios and I didn't want to introduce anything too complex but I'm open for suggestions.

distributed/dashboard/scheduler.py
mrocklin: > As far as I understand, the approach with prometheus should rather be to export raw metric data

Fair enough.  Is there some better term than forgotten?  This seems a bit negative, as though something has gone wrong.  I can imagine people raising issues "why did Dask forget so many of my tasks?" in the future.

mrocklin: I'm also happy to merge this as-is though.  My comments here are mostly academic.

fjetter: > Fair enough. Is there some better term than forgotten?

The term `done` comes to my mind since (if everything behaves as expected) a task is transitioned to `forgotten` once it is no longer needed, i.e. `done`. More descriptive would be `successfully processed` similar to what you already suggested.
I wanted to stick to the internal state name to avoid confusion if somebody actually bothers to look into the code :)
I will change the documentation to better describe what this actually means.

fjetter: I was initially also struggling with all of the states and am wondering if they are documented anywhere. Since these states are also visualised on the dashboard it probably makes sense to include it in the official docs, e.g. a section in https://distributed.dask.org/en/latest/journey.html

jrbourbeau: There's a description of the different task states at https://distributed.dask.org/en/latest/scheduling-state.html#task-state

mrocklin:
OK.   Merging this in.  Thanks @fjetter !

PR #3308, opened 2019-12-09
Use hostname as default IP address rather than localhost

mrocklin:
Previously if we couldn't connect to the scheduler we used localhost
this made sense for testing, but probably doesn't make sense for
operations.


jcrist:
Makes sense to me. Thanks, merging.

PR #3309, opened 2019-12-09
Use worker name in logs

StephanErb:
## Current Behaviour

Workers can get started with a `--name` property

    dask-worker --name myworker-de820c tcp://192.168.0.144:8786

This property is used on the UI but not within the scheduler logs. This can make it difficult to identify particular worker instances during debugging. 

    distributed.scheduler - INFO - Register worker tcp://192.168.0.144:39535
    distributed.scheduler - INFO - Starting worker compute stream, tcp://192.168.0.144:39535
    distributed.scheduler - INFO - Remove worker tcp://192.168.0.144:39535
    distributed.core - INFO - Removing comms to tcp://192.168.0.144:39535
    distributed.scheduler - INFO - Lost all workers

## Expected Behavior
When starting a worker with `--name` the name is consistently used within scheduler logs

    distributed.scheduler - INFO - Register worker myworker-de820c (tcp://192.168.0.144:40607)
    distributed.scheduler - INFO - Starting worker compute stream, tcp://192.168.0.144:40607
    distributed.scheduler - INFO - Remove worker myworker-de820c (tcp://192.168.0.144:40607)
    distributed.core - INFO - Removing comms to tcp://192.168.0.144:40607
    distributed.scheduler - INFO - Lost all workers

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -291,14 +291,12 @@ class WorkerState(object):
         return ws
 
     def __repr__(self):
-        return "<Worker %r, memory: %d, processing: %d>" % (
+        return "<Worker %r, name: %s, memory: %d, processing: %d>" % (
             self.address,
+            self.name,
             len(self.has_what),
             len(self.processing),
         )
-
-    def __str__(self):
-        return self.address
 
     def identity(self):
         return {
@@ -1435,7 +1433,7 @@ class Scheduler(ServerNode):
 
             ws = self.workers.get(address)
             if ws is not None:
-                raise ValueError("Worker already exists %s" % address)
+                raise ValueError("Worker already exists %s" % ws)
 
             self.workers[address] = ws = WorkerState(
                 address=address,
@@ -1518,7 +1516,7 @@ class Scheduler(ServerNode):
 
             self.log_event(address, {"action": "add-worker"})
             self.log_event("all", {"action": "add-worker", "worker": address})
-            logger.info("Register %s", str(address))
+            logger.info("Register worker %s", ws)
 
             if comm:
                 await comm.write(
@@ -1920,7 +1918,7 @@ class Scheduler(ServerNode):
                     "processing-tasks": dict(ws.processing),
                 },
             )
-            logger.info("Remove worker %s", address)
+            logger.info("Remove worker %s", ws)
             if close:
                 with ignoring(AttributeError, CommClosedError):
                     self.stream_comms[address].send({"op": "close", "report": False})
@@ -1991,7 +1989,7 @@ class Scheduler(ServerNode):
                 dask.config.get("distributed.scheduler.events-cleanup-delay")
             )
             self.loop.call_later(cleanup_delay, remove_worker_from_events)
-            logger.debug("Removed worker %s", address)
+            logger.debug("Removed worker %s", ws)
 
         return "OK"
 
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -6,6 +6,7 @@ from contextlib import contextmanager
 from datetime import timedelta
 import functools
 from hashlib import md5
+import html
 import inspect
 import json
 import logging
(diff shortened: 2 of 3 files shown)


distributed/scheduler.py
```diff
@@ -291,14 +291,18 @@ def clean(self):
         return ws
 
     def __repr__(self):
-        return "<Worker %r, memory: %d, processing: %d>" % (
+        return "<Worker %r, name: %s, memory: %d, processing: %d>" % (
```
mrocklin: Should we include this if name is None or the same as the address?

distributed/scheduler.py
```diff
@@ -291,14 +291,18 @@ def clean(self):
...
             len(self.has_what),
             len(self.processing),
         )
 
     def __str__(self):
-        return self.address
+        if self.name != self.address:
+            return "%s (%s)" % (self.name, self.address)
+        else:
+            return self.address
```
mrocklin: I wonder why we don't just use `__str__ = __repr__` here.  It might be worth trying.

distributed/scheduler.py
StephanErb: Agreed. I can change that (tomorrow)

distributed/scheduler.py
StephanErb: Sounds good to me. I will give it a shot.

jrbourbeau: FWIW by default `__str__` will call `__repr__` if it's not implemented. So we may be able to remove `__str__` altogether

StephanErb: I have addressed the implementation as you proposed by dropping  `__str__`. This seems to have the desired effect.

distributed/scheduler.py
StephanErb: I got second thoughts during the implementation of the proposal and therefore left it out for now:

* In order to add an `if` for the optional name, we would need to provide the scheduler implementation with the knowledge that worker defaults its name to its address. This would break encapsulation.
* On the UI we also always show the name field even if it is identical to the worker address.

If you insist I am happy make the name optional though.

TomAugspurger:
@StephanErb are you able to reproduce the test failure locally? https://travis-ci.org/dask/distributed/jobs/623376084#L2583

```pytb
__________________________________ test_logs ___________________________________

cleanup = None

    @pytest.mark.asyncio

    async def test_logs(cleanup):

        worker = {"cls": Worker, "options": {"nthreads": 1}}

        async with SpecCluster(

            asynchronous=True, scheduler=scheduler, worker=worker

        ) as cluster:

            cluster.scale(2)

            await cluster

    

            logs = await cluster.logs()

>           assert is_valid_xml("<div>" + logs._repr_html_() + "</div>")

distributed/deploy/tests/test_spec_cluster.py:282: 

distributed/utils.py:1356: in is_valid_xml

    return xml.etree.ElementTree.fromstring(text) is not None

text = "<div><details>\n<summary style='display:list-item'>Scheduler</summary>\n<pre><code>\ndistributed.scheduler - INFO - C...:34264\ndistributed.worker - INFO - -------------------------------------------------\n</code></pre>\n</details></div>"

parser = <xml.etree.ElementTree.XMLParser object at 0x7f32894baac8>

    def XML(text, parser=None):

        """Parse XML document from string constant.

    

        This function can be used to embed "XML Literals" in Python code.

    

        *text* is a string containing XML data, *parser* is an

        optional parser instance, defaulting to the standard XMLParser.

    

        Returns an Element instance.

    

        """

        if not parser:

            parser = XMLParser(target=TreeBuilder())

>       parser.feed(text)

E         File "<string>", line None

E       xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 6, column 55

```

distributed/scheduler.py
mrocklin: OK, I'm happy to accept the push back.

distributed/utils.py
```diff
@@ -1296,7 +1297,9 @@ class Log(str):
     """ A container for logs """
 
     def _repr_html_(self):
-        return "<pre><code>\n{log}\n</code></pre>".format(log=self.rstrip())
+        return "<pre><code>\n{log}\n</code></pre>".format(
+            log=html.escape(self.rstrip())
```
StephanErb: The change discovered a missing HTML escaping during the UI rendering. This fixes the test for me locally. Let's see if Travis agrees

mrocklin:
This looks good to me.  Thank you @StephanErb !

PR #3310, opened 2019-12-10
Update latencies with heartbeats

fjetter:
The latency is currently only calculate once during connect. The heartbeat seemed to be a good periodic callback which could update this. Also the latency seems to be a good digest candidate, similar to the tick length

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1586,3 +1586,15 @@ async def test_bad_startup(cleanup):
             w = await Worker(s.address, startup_information={"bad": bad_startup})
         except Exception:
             pytest.fail("Startup exception was raised")
+
+
+@pytest.mark.asyncio
+async def test_update_latency(cleanup):
+    async with await Scheduler() as s:
+        async with await Worker(s.address) as w:
+            original = w.latency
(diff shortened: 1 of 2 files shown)


distributed/tests/test_worker.py
```diff
@@ -1586,3 +1586,15 @@ def bad_startup(w):
             w = await Worker(s.address, startup_information={"bad": bad_startup})
         except Exception:
             pytest.fail("Startup exception was raised")
+
+
+@pytest.mark.asyncio
+async def test_update_latency():
```
mrocklin:
```suggestion
async def test_update_latency(cleanup):
```

This will enforce a variety of checks.

In particular, it will also require that your schedulers and workers are cleaned up by the end of the test.  I recommend using `async with` context managers below for both.

mrocklin:
Other than the small comment about the `cleanup` fixture this looks good to me.  Thanks @fjetter !

distributed/tests/test_worker.py
fjetter: done

mrocklin:
Thanks!  I'll merge once CI finishes.

PR #3311, opened 2019-12-10
Increase severity of various log messages

StephanErb:
By increasing the log severity for log statements outside of the happy path, we can make it significantly simpler to search for problems within a central log aggregator (e.g. elasticsearch)

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -907,7 +907,7 @@ class Scheduler(ServerNode):
             try:
                 from distributed.dashboard import BokehScheduler
             except ImportError:
-                logger.debug("To start diagnostics web server please install Bokeh")
+                logger.warning("To start diagnostics web server please install Bokeh")
             else:
                 self.service_specs[("dashboard", dashboard_address)] = (
                     BokehScheduler,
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -907,7 +907,7 @@ def __init__(
             try:
                 from distributed.dashboard import BokehScheduler
             except ImportError:
-                logger.debug("To start diagnostics web server please install Bokeh")
+                logger.warning("To start diagnostics web server please install Bokeh")
```
TomAugspurger: Perhaps keep at debug, since bokeh is an optional dependency.

distributed/worker.py
```diff
@@ -585,7 +585,7 @@ def __init__(
             try:
                 from distributed.dashboard import BokehWorker
             except ImportError:
-                logger.debug("To start diagnostics web server please install Bokeh")
+                logger.warning("To start diagnostics web server please install Bokeh")
```
TomAugspurger: Prefer to keep at `debug`.

distributed/worker.py
```diff
@@ -1191,7 +1191,7 @@ def send_to_worker(self, address, msg):
             max_connections is not False
             and self.outgoing_current_count >= max_connections
         ):
-            logger.debug(
+            logger.warning(
```
TomAugspurger: In https://github.com/dask/distributed/pull/3071#discussion_r330066652 we decided to leave this at debug level. Ideally these warnings would be throttled at a higher-level, but right now they'll show up frequently.

StephanErb: Thanks for the quick feedback. I have to admit this was the one why I created this PR in the first place. It sounded too important to be not visible at all.

StephanErb:
Your comment indicates we need another fix for that. Closing fow now to not bother you with a half-done PR. I might come back :)

PR #3314, opened 2019-12-12
Support out-of-tree comms backends using entrypoints

sodre:
This in an alternative version to #3305, but using the faster `entrypoints` library. 

cc: @jcrist


sodre:
close in favor of #3305

Issue #3318, opened 2019-12-13
New dashboard plot discussion


jacobtomlinson:
I've been experimenting with a new dashboard element to help visualise transfer between workers and overall cluster communication. This is a super early rough mockup of the kind of thing we could have. I'd like to share it and have some discussion around it before I put more time into it.

Currently it shows tasks being sent to workers and memory transfer between workers.

Things on my list of improvements so far are:
- Place workers spatially based on their network performance. Workers with higher bandwidth to each other should be closer together.
- Workers could show their memory and CPU usages.
- Clients could be shown and client-scheduler transfers could be shown. (We could have the client notify the scheduler of incoming work if it is about to serialise a lot of things, the apparent hang here is a common grumble).
- Various UCX protocols could be shown somehow instead of just the red transfer lines.
- Show disk-read and disk-write operations somehow.

Feedback/criticism would be appreciated!

![ezgif com-optimize(2)](https://user-images.githubusercontent.com/1610850/70787442-902e5480-1d86-11ea-8e7a-a0b742dca499.gif)

mrocklin:
> Place workers spatially based on their network performance. Workers with higher bandwidth to each other should be closer together.

FWIW I would personally find the networking structure to be more valuable.  For example I'd love to see all of the workers grouped by hostname.  Our measurement of bandwidth is often somewhat messy and not necessarily particularly telling.

> Clients could be shown and client-scheduler transfers could be shown. (We could have the client notify the scheduler of incoming work if it is about to serialise a lot of things, the apparent hang here is a common grumble).

+1

> Various UCX protocols could be shown somehow instead of just the red transfer lines.

For the common-case Dask user UCX use is still somewhat uncommon.  Dask also may not have good visibility into what was used.  Instead, I would encourage us to focus on the amount of transfer (maybe with duration) and the bandwidth of transfer (maybe with width), which are more universally accessible and important.

> Show disk-read and disk-write operations somehow.

It might be useful for workers to publish their disk usage as a metric to be stored back up in the scheduler.  This could be a small incremental PR if you want to get your feet wet with pushing telemtry back up to the scheduler.

mrocklin:
Also, thanks for doing this, I'm enthusiastic to see what eventually comes out of this

jacobtomlinson:
Thanks Matt!

> For example I'd love to see all of the workers grouped by hostname.

[...]

PR #3319, opened 2019-12-13
Switch startstops to dicts and add worker name to transfer

jacobtomlinson:
While working on #3318 I was unable to find out where a transfer had come from when processing task `startstops`. _(The current gifs in that issue show mocked data where the start worker is randomly selected)._

After chatting with @mrocklin we discussed switching the `startstop` tuples for dictionaries to allow passing additional metadata more easily.

This PR switches the tuples for dicts and also add a `worker` key to the transfer `startstop`. This is the worker the transfer came from which should allow me to plot the transfers correctly.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -215,7 +215,7 @@ class Worker(ServerNode):
         The exception caused by running a task if it erred
     * **tracebacks**: ``{key: traceback}``
         The exception caused by running a task if it erred
-    * **startstops**: ``{key: [(str, float, float)]}``
+    * **startstops**: ``{key: [{startstop}]}``
         Log of transfer, load, and compute times for a task
 
     * **priorities**: ``{key: tuple}``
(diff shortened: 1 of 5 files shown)


TomAugspurger:
LGTM, though there's a CI failure. Does it seem related?

jacobtomlinson:
Yeah I think it is. Just looking at it now.

distributed/worker.py
```diff
@@ -1933,11 +1935,12 @@ def select_keys_for_gather(self, worker, dep):
...
-                        (
-                            "transfer",
-                            start + self.scheduler_delay,
-                            stop + self.scheduler_delay,
-                        )
+                        {
+                            "action": "transfer",
+                            "start": start + self.scheduler_delay,
+                            "stop": stop + self.scheduler_delay,
+                            "worker": worker,
```
mrocklin: The term worker here miight be a little bit ambiguous about which side of the transfer it refers to.  Maybe `sender` or `source`?  Not a huge deal either way, just an idle comment.

jacobtomlinson: Yeah that makes sense! My preference would be `source`.

jacobtomlinson:
```suggestion
                            "source": worker,
```

mrocklin:
This is in. Thanks @jacobtomlinson !

PR #3320, opened 2019-12-13
Add setuptools to dependencies

jrbourbeau:
In #3305 we added an import of `pkg_resources`. This PR adds `setuptools` (which includes `pkg_resources`) to our requirements. 

cc @TomAugspurger


TomAugspurger:
Thanks. I think this is the correct thing to do, though perhaps @jakirkham can verify?

jakirkham:
Thanks @jrbourbeau! 😄

Issue #3256, opened 2019-11-21
Task stuck processing on non-existent worker


bnaul:
At the end of long-running jobs, I'm often seeing one or two tasks that never finish. Looking a bit more closely, it seems that the task is `processing` on a non-existent worker:
![image](https://user-images.githubusercontent.com/903655/69368737-84e87d00-0c4f-11ea-8612-a4fed55af49b.png)

Even after closing every worker, the status of this task doesn't change.

Searching through the scheduler logs for this task I found:
```
2019-11-20T18:58:16.118979408Z distributed.scheduler - INFO - Unexpected worker completed task, likely due to work stealing.  Expected: tcp://10.47.187.2:42561, Got: tcp://10.47.187.2:42561, Key: ('from_pandas-f8d9cff3cb70c117a611b7cf40be25d3', 9507)
```
Seems pretty fishy that a worker could steal a task from itself? 🤔 

Some other context:
- k8s cluster w/ ~1000 workers (before scaling to investigate the deadlock), around 100k tasks total (50k of these pandas tasks, and a processing task for each partition)
- Our workers run OOM and get evicted somewhat frequently (every few hrs on average), and we use a lot of retries to compensate
- We GC futures as we go using `dask.as_completed(futures); del futures`

Full scheduler logs related to this renegade worker, which has quite a 🎢  minute at 18:58:
```
2019-11-20T18:46:57.497189783Z distributed.scheduler - INFO - Register tcp://10.47.187.2:42561
2019-11-20T18:46:57.497605833Z distributed.scheduler - INFO - Starting worker compute stream, tcp://10.47.187.2:42561
2019-11-20T18:57:57.392080977Z distributed.scheduler - INFO - Remove worker tcp://10.47.187.2:42561
2019-11-20T18:57:57.392125386Z distributed.core - INFO - Removing comms to tcp://10.47.187.2:42561
2019-11-20T18:58:01.058431284Z distributed.scheduler - INFO - Unexpected worker completed task, likely due to work stealing.  Expected: tcp://10.47.193.16:36713, Got: tcp://10.47.187.2:42561, Key: ('to_records-ff4aa1a02b819ae4fa0769294d47599b', 21371)
[...]

PR #3321, opened 2019-12-13
Update comparison logic for worker state

rockwellw:
Addresses #3256 , based on @TomAugspurger 's comment.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2557,7 +2557,7 @@ class Scheduler(ServerNode):
         if ts is None:
             return
         ws = self.workers[worker]
-        if ts.processing_on is not ws:
+        if ts.processing_on != ws:
             return
         r = self.stimulus_missing_data(key=key, ensure=False, **msg)
         self.transitions(r)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4020,7 +4020,7 @@ class Scheduler(ServerNode):
             if ws is None:
                 return {key: "released"}
 
-            if ws is not ts.processing_on:  # someone else has this task
+            if ws != ts.processing_on:  # someone else has this task
                 logger.info(
                     "Unexpected worker completed task, likely due to"
                     " work stealing.  Expected: %s, Got: %s, Key: %s",


TomAugspurger:
Thanks. Do you think this will actually fix #3256? This being the cause was a shot in the dark.

And do you know if / how `WorkerState.__eq__` is defined? I don't see one at a glance. Does implementing `__slots__` give us one?

rockwellw:
This may just address a symptom and not the root problem, but it seems like a good starting point considering the log messages @bnaul was seeing.

TomAugspurger:
Ok. Do you know how `__eq__` is implemented ok WorkerState?

TomAugspurger:
I think `WorkerState.__eq__` may actually be using identity. I would have expected these to be equal, if we were comparing attributes

```python
In [15]: from distributed.scheduler import WorkerState

In [16]: a = WorkerState()

In [17]: b = WorkerState()

In [18]: a == b
Out[18]: False
```

This may not be a problem if WorkerState instances are expected to be singletons per worker, which I suspect they are.

StephanErb:
There is another occurrence of the same identity comparison here https://github.com/dask/distributed/blob/d747f639d73c5f9b8961f2f80e407868de7da4cb/distributed/scheduler.py#L2560 To reduce surprises, I guess both should be fixed.

While I have no proof, I could totally imagine that race conditions between `add_worker` and `remove_worker` break the single instance assumption.

bnaul:
> I think `WorkerState.__eq__` may actually be using identity. I would have expected these to be equal, if we were comparing attributes

@TomAugspurger it looks like `identity()` has the info that we would want to compare against, do you think something like this might do the trick?
```
class WorkerState():
    def __eq__(self, value):
        if hasattr(value, 'identity'):
            return self.identity() == value.identity()
        else:
            return False
```

mrocklin:
Rather than conditionals I usually do something like ...

    return type(self) == type(other) and self.identity() == other.identity()

Also, if we're going to implement `__eq__` then we'll probably also have to implement `__hash__` (should be more or less the same)

bnaul:
> Rather than conditionals I usually do something like ...
> 
> ```
> return type(self) == type(other) and self.identity() == other.identity()
> ```
Probably not at all important for this case but doesn't this pattern kind of break subclassing? How about `isinstance(other, type(self))`?

> Also, if we're going to implement `__eq__` then we'll probably also have to implement `__hash__` (should be more or less the same)
Do we currently rely on `WorkerState`s being hashable? It seems to me like they're mutable so wouldn't we want `hash` to be `None`?

mrocklin:
If they're of different classes then can they be equal?  (but yeah, in truth it doesn't matter)

Fair enough on hashability (assuming that it doesn't come up, which it may)

bnaul:
@mrocklin you were right that hashability is needed so we added that as well and changed `__eq__` to rely on that instead. it seems to me like just name+host is enough to uniquely specify the WorkerState but if you think we should add more attributes we can do that.

@TomAugspurger anecdotally we haven't seen #3256 since we started running with this change, but I wouldn't say we've conclusively proven it's resolved since unfortunately that's a random unpredictable failure. regardless this seems like the "right" check to be performing and should help w/ possible race conditions as pointed out by @StephanErb, so my vote would be to merge this if everyone's happy with it and close #3256 until someone observes that same behavior again.

TomAugspurger:
Has anyone seen this CI failure before?

```

____________________________ test_pickle_functions _____________________________

    def test_pickle_functions():

        def make_closure():

            value = 1

    

            def f(x):  # closure

                return x + value

    

            return f

    

        def funcs():

            yield make_closure()

            yield (lambda x: x + 1)

            yield partial(add, 1)

    

        for func in funcs():

            wr = weakref.ref(func)

            func2 = loads(dumps(func))

            wr2 = weakref.ref(func2)

            assert func2(1) == func(1)

            del func, func2

            gc.collect()

>           assert wr() is None

E           AssertionError: assert <function test_pickle_functions.<locals>.funcs.<locals>.<lambda> at 0x7f0ab6f535f0> is None

E            +  where <function test_pickle_functions.<locals>.funcs.<locals>.<lambda> at 0x7f0ab6f535f0> = <weakref at 0x7f0a8cc05350; to 'function' at 0x7f0ab6f535f0 (<lambda>)>()

distributed/protocol/tests/test_pickle.py:47: AssertionError`
```

I've restarted that build.

mrocklin:
Not recently, but yes, it used to be an intermittent failure.

TomAugspurger:
OK, thanks. We'll merge this and keep an eye out for that failure again.

Thanks @rockwellw!

distributed/scheduler.py
```diff
@@ -271,6 +271,12 @@ def __init__(
 
         self.extra = extra or {}
 
+    def __hash__(self):
+        return hash((self.name, self.host))
+
+    def __eq__(self, other):
+        return type(self) == type(other) and hash(self) == hash(other)
```
StephanErb: Sorry for being late to the party. There is a bug in here: It is not guaranteed that two WorkerState objects are actually the same even though there hash value is the same. Hashing is not a bijective function.

I think we need to go the extra route here and check for `(self.name, self.host) == (other.name, other.host)`

StephanErb: @TomAugspurger  @bnaul  please see above.

While I totally agree that hash collisions are very very unlikely, Dask has a massive install base with sometimes hundreds to thousands of workers per cluster. So as far as we know the once in a million chance event be next Tuesday. 🤷‍♂

PR #3322, opened 2019-12-15
Always add new TaskGroup to TaskPrefix

### LINKED ISSUE #3368: Incorrect task counts in scheduler status page
Surprised this hasn't already been posted as I started noticing it at least a month ago:
```
c = distributed.Client()
f = c.submit(sum, [1, 2])
g = c.submit(len, [1, 2])
h = c.submit(sum, [3, 4])
```
Previously the plot would show 2 `sum`s and 1 `len`; now (on 2.9.1 and master, `distributed==2.9.1+10.g44d24098`) it shows 3 of each:
image
This behavior continues indefinitely, so with e.g. 1 `len` and 10000 `sums` we'd see 10001 of each for a total of 20002. Observed the same behavior w/ bokeh 1.0.4 and 1.4.0.

jrbourbeau:
Currently a `TaskGroup` is only added to a `TaskPrefix` when a new `TaskPrefix` is created:

https://github.com/dask/distributed/blob/d747f639d73c5f9b8961f2f80e407868de7da4cb/distributed/scheduler.py#L1969

I think we want to add append a `TaskGroup` to `TaskPrefix.groups` anytime a new group is created. 

Closes #3368

### DIFF
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -280,7 +280,11 @@ class _PrometheusCollector(object):
 
         yield CounterMetricFamily(
             "dask_scheduler_tasks_forgotten",
-            "Total number of processed tasks no longer in memory and already removed from the scheduler job queue.",
+            (
+                "Total number of processed tasks no longer in memory and already "
+                "removed from the scheduler job queue. Note task groups on the "
+                "scheduler which have all tasks in the forgotten state are not included."
+            ),
             value=task_counter.get("forgotten", 0.0),
         )
 
--- a/distributed/dashboard/tests/test_scheduler_bokeh_html.py
+++ b/distributed/dashboard/tests/test_scheduler_bokeh_html.py
@@ -171,7 +171,7 @@ async def test_prometheus_collect_task_states(c, s, a, b):
     active_metrics, forgotten_tasks = await fetch_metrics()
     assert active_metrics.keys() == expected
     assert sum(active_metrics.values()) == 0.0
-    assert sum(forgotten_tasks) == 1.0
+    assert sum(forgotten_tasks) == 0.0
 
 
 @gen_cluster(
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -833,7 +833,7 @@ class TaskPrefix(object):
 
     @property
     def types(self):
-        return set.union(*[tg.types for tg in self.groups])
+        return set().union(*[tg.types for tg in self.groups])
 
 
 class _StateLegacyMapping(Mapping):
@@ -1963,7 +1963,7 @@ class Scheduler(ServerNode):
             tp = self.task_prefixes[prefix_key] = TaskPrefix(prefix_key)
         ts.prefix = tp
 
-        group_key = ts.group_key or prefix_key
+        group_key = ts.group_key
         try:
             tg = self.task_groups[group_key]
         except KeyError:
@@ -4603,6 +4603,13 @@ class Scheduler(ServerNode):
                         logger.info("Plugin failed with exception", exc_info=True)
                 if ts.state == "forgotten":
                     del self.tasks[ts.key]
+
+            if ts.state == "forgotten":
+                # Remove TaskGroup if all tasks are in the forgotten state
+                tg = ts.group
+                if not any(tg.states.get(s) for s in ALL_TASK_STATES):
+                    ts.prefix.groups.remove(tg)
+                    del self.task_groups[tg.name]
 
             return recommendations
         except Exception as e:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1963,10 +1963,11 @@ class Scheduler(ServerNode):
             tp = self.task_prefixes[prefix_key] = TaskPrefix(prefix_key)
         ts.prefix = tp
 
+        group_key = ts.group_key or prefix_key
         try:
-            tg = self.task_groups[ts.group_key]
+            tg = self.task_groups[group_key]
         except KeyError:
-            tg = self.task_groups[ts.group_key] = TaskGroup(ts.group_key)
+            tg = self.task_groups[group_key] = TaskGroup(group_key)
             tg.prefix = tp
             tp.groups.append(tg)
         tg.add(ts)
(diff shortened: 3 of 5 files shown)


mrocklin:
Thanks for catchiing this @jrbourbeau !  In principle this looks good to me, and I'm always happy to see people engaging on the scheduler.  It looks like there is a failing test here.

jrbourbeau:
Hmm the test appears to be failing due how we handle tasks like `"inc-123"`. These types of tasks belong to the `TaskPrefix` I expected (task `"inc-123"` belongs to `TaskPrefix` `"inc"`), however they all belong to the same `TaskGroup` with an empty string name:

```python
In [1]: from distributed.utils import key_split_group, key_split

In [2]: key_split("inc-123")    # task prefix name
Out[2]: 'inc'

In [3]: key_split_group("inc-123")    # task group name
Out[3]: ''

In [4]: key_split_group("add-456")    # task group name
Out[4]: ''
```

I _think_ what we want (certainly open to other suggestions) is to have a separate group for keys like `"inc-123"` such that `"inc-123"`, `"inc-456"`, etc. belong to an `"inc"` `TaskGroup` instead of the `""` `TaskGroup`. So something like:

`"inc-123"` -> `TaskGroup("inc")` -> `TaskPrefix("inc")`
`"inc-456"` -> `TaskGroup("inc")` -> `TaskPrefix("inc")`
`"(inc-789, 0)"` -> `TaskGroup("inc-789")` -> `TaskPrefix("inc")`

It's worth noting that this will result in an increased number of task groups compared to our current behavior

mrocklin:
So, *currently* task groups track dependencies, and are acyclic.  Having all `inc-123` tasks go to the `inc` task group would break this.  Today this doesn't matter (we haven't built up much around TaskGroups yet) but I imagine that it might in the near future.

Having inc-123 go to a task group of inc-123 would mean that we have *many* task groups in dask.delayed or futures workloads.  We currently let TaskGroups persist after all of their tasks are gone.

mrocklin:
The solution so far has been to just not give tasks like `inc-123` a TaskGroup (or assign them the `None` group) which I think makes some sense.  They aren't really part of a group of tasks.

jrbourbeau:
Thanks for the feedback!

The main reason I proposed having tasks like `"inc-123"` belong to a group `TaskGroup("inc")` was I was confused about how tasks are currently assigned a `TaskPrefix` and `TaskGroup`. For example, task `"inc-124"` is tracked by `TaskGroup("")` and `TaskPrefix("inc")`. But it wasn't clear to me that `TaskGroup("")` should belong to `TaskPrefix("inc")`'s `.groups` list. There can be many other tasks, e.g. `"add-456"`, which also belong to `TaskGroup("")` but have a different `TaskPrefix`, e.g. `TaskPrefix("add")`. This confusion may come from a lack of understanding on my part. 

As an alternative, we could continue to assign tasks like `"inc-123"` to `TaskGroup("")`, but also have them belong to a null task prefix like `TaskPrefix("")` that only tracks the `TaskGroup("")` group?

> So, currently task groups track dependencies, and are acyclic

Thanks for pointing that out. I wasn't thinking about how task group dependencies fit into the proposed changes. FWIW, `TaskGroup("")` can have itself as a dependency because tasks like `inc-123` and `add-456` both belong to that group and can depend on one another:

<details>
<summary>Example:</summary>

```python

if __name__ == "__main__":
    
    inc = lambda x: x + 1
    add = lambda x, y: x + y
    
    with Client() as client:
        f_inc = client.submit(inc, 1)
        f_add = client.submit(add, f_inc, 2)
        wait([f_inc, f_add])
        
        group = client.cluster.scheduler.task_groups[""]
        print(f"group = {group}")
        print(f"group in group.dependencies = {group in group.dependencies}")
```

outputs

```
group = <no-group: memory: 2>
group in group.dependencies = True
```

</details>

mrocklin:
> I was confused about how tasks are currently assigned a TaskPrefix and TaskGroup

Yeah, this is all fairly new and rough.  I'm sure that we'll find better ways as we continue down this path.

> But it wasn't clear to me that TaskGroup("") should belong to TaskPrefix("inc")'s .groups list.

This is not clear to me either

> There can be many other tasks, e.g. "add-456", which also belong to TaskGroup("") but have a different TaskPrefix, e.g. TaskPrefix("add"). This confusion may come from a lack of understanding on my part

Yup.  Same.

> FWIW, TaskGroup("") can have itself as a dependency because tasks like inc-123 and add-456 both belong to that group and can depend on one another:

Yes, I'm mostly planning to ignore that group in future work.

> As an alternative, we could continue to assign tasks like "inc-123" to TaskGroup(""), but also have them belong to a null task prefix like TaskPrefix("") that only tracks the TaskGroup("") group?

Currently the prefix is used for the `duration_average` attribute, which affects scheduling decisions.  It's important for `inc-123` to map to the TaskPrefix `inc`

jrbourbeau:
I've updated this PR to remove `TaskGroup`s once all their tasks are in the `"forgotten"` state and also included a modified version of the test added over in https://github.com/dask/distributed/pull/3370 (thanks @quasiben!)

To reiterate the main changes in this PR are:

- Tasks with keys like `"inc-123"` (e.g. from `client.submit(inc, 1)`) are now added to `TaskGroup("inc")`. Previously they were added to a single `TaskGroup("")`.

- Because we are creating more task groups, we now remove task groups once all their tasks are in the `"forgotten"` state. This is to avoid holding onto many task groups in `dask.delayed` or futures workloads.

quasiben:
Looks great @jrbourbeau  -- thanks for taking this up

bnaul:
this seems like a fairly innocuous change, @jrbourbeau @jacobtomlinson any reason to hold off on merging? just don't want it to miss the next release since IMO the dashboard is currently close to useless for a complex workflow with many different tasks

jrbourbeau:
Thanks for following up @bnaul. Let's give a little more time for an additional review. I agree this should be included in the next bi-weekly release to fix known dashboard issues

mrocklin:
If you all are comfortable with this then please don't wait up for me

jrbourbeau:
Thanks @quasiben and @jacobtomlinson for reviewing!

PR #3325, opened 2019-12-16
add performance report docs and color definitions to docs

quasiben:
PR updates docs with `performance_report` info and adds definitions of colors to web interface page

cc @mrocklin

### DIFF
--- a/docs/source/diagnosing-performance.rst
+++ b/docs/source/diagnosing-performance.rst
@@ -115,6 +115,35 @@ command on the workers:
    client.run(lambda dask_worker: dask_worker.incoming_transfer_log)
 
 
+Performance Reports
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks @quasiben !  

It might be good to also add this content to https://docs.dask.org/en/latest/diagnostics-distributed.html which lives at https://github.com/dask/dask/blob/master/docs/source/diagnostics-distributed.rst

I've tried to start consolidating docs there.  That site gets 10x the views that distributed.dask.org gets.

quasiben:
Sure, I'll add a PR over there as well

quasiben:
see: https://github.com/dask/dask/pull/5713

PR #3332, opened 2019-12-19
Stop dask scheduler gracefully

habibutsu:
Currently if you terminating dask-scheduler following exception occurs:
```
  File "/usr/local/bin/dask-scheduler", line 8, in <module>
    sys.exit(go())
  File "/usr/local/lib/python3.6/dist-packages/distributed/cli/dask_scheduler.py", line 248, in go
    main()
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/distributed/cli/dask_scheduler.py", line 237, in main
    loop.run_sync(run)
  File "/usr/local/lib/python3.6/dist-packages/tornado/ioloop.py", line 531, in run_sync
    raise TimeoutError("Operation timed out after %s seconds" % timeout)
tornado.util.TimeoutError: Operation timed out after None seconds
```

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -227,7 +227,10 @@ def main(
     logger.info("Local Directory: %26s", local_directory)
     logger.info("-" * 47)
 
-    install_signal_handlers(loop)
+    async def on_signal(signum):
+        await scheduler.close()
+
+    install_signal_handlers(loop, cleanup=on_signal)
 
     async def run():
         await scheduler


mrocklin:
Thank you for the fix @habibutsu !  I've tried this locally and confirmed both the original bug and that this fixes the issue.  I'm actually fairly surprised that this bug was here in the first place.  I thought that we had things working nicely here not too long ago.  I'm curious, do you have any thoughts on how we might test for this to make sure that things don't revert to the poor behavior in the future?

mrocklin:
Hrm, the test failure in `distributed/tests/test_client.py::test_reconnect` does seem to be genuine.

TomAugspurger:
Hopefully https://github.com/dask/distributed/pull/3332/commits/d5cd0c925768761e192c21203a60145e0f844997 fixed the test. That's roughly doing

```
- create worker
- create scheduler
- close scheduler
- create scheduler
- do stuff assuming worker has reconnected
```

The difference is that we now (IMO correctly) tell the workers to close when the scheduler closes, so the worker cleanly exited and didn't attempt to reconnect.

By closing the scheduler with `signal.SIGKILL`, which we don't have a handler for, we restore the assumption behind the test.

TomAugspurger:
Interestingly, this is failing after https://github.com/dask/distributed/pull/3706. I'm not sure why that would be. Even `dask-scheduler` is exiting with a non-zero exit code now. If I revert #3706 on this branch, then things fail.

@crusaderky do you have any guesses why that would be? If not, I can dig into things some more.

crusaderky:
yes, I noticed a few cases where you just can't replace the gen.coroutines with async def functions. Afraid I don't know enough about tornado.gen to understand why.

crusaderky:
asyncio signal handling is very particular. Maybe tornado has its own, incompatible, approach?

TomAugspurger:
OK thanks. For now I'm OK with reverting the asyncio changes in that function.

distributed/cli/utils.py
```diff
@@ -50,10 +51,13 @@ def install_signal_handlers(loop=None, cleanup=None):
     old_handlers = {}
 
     def handle_signal(sig, frame):
-        async def cleanup_and_stop():
+        @gen.coroutine
```
crusaderky, line 54:
```suggestion
        # FIXME: this breaks when changing to async def... await
        @gen.coroutine
```

TomAugspurger:
The test I added is failing on windows. Apparently the process may not be exiting cleanly there. If anyone is able to debug that it'd be welcome, but I've skipped the assertions on Windows for now.

PR #3335, opened 2019-12-20
Add websocket scheduler plugin

jacobtomlinson:
This PR adds a new `/events` API endpoint which provides a websocket plugin to the scheduler.

When a client opens a websocket connection to this endpoint a scheduler plugin is registered which forwards all plugin events (`restart`, `add_worker`, `remove_worker` and startstops from `transition` events) into the websocket. Clients can also ping the scheduler via the socket to check that it is still alive.

This enables the building of custom dashboards such as the one proposed in #3318.

I'm raising this ahead of the dashboard as I'd like some feedback on the design here, and thoughts on whether it will affect performance .

### DIFF
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from functools import partial
+import json
 import logging
 
 import dask
@@ -11,6 +12,7 @@ except ImportError:
     from toolz import merge, merge_with
 
 from tornado import escape
+from tornado.websocket import WebSocketHandler
 
 try:
     import numpy as np
@@ -45,6 +47,7 @@ from .core import BokehServer
 from .worker import counters_doc
 from .proxy import GlobalProxyHandler
 from .utils import RequestHandler, redirect
+from ..diagnostics.websocket import WebsocketPlugin
 from ..utils import log_errors, format_time
 from ..scheduler import ALL_TASK_STATES
 
@@ -319,6 +322,34 @@ class HealthHandler(RequestHandler):
         self.set_header("Content-Type", "text/plain")
 
 
+class EventsHandler(WebSocketHandler):
+    def initialize(self, server=None, extra=None):
+        self.server = server
+        self.extra = extra or {}
+        self.plugin = WebsocketPlugin(self, server)
+        self.server.add_plugin(self.plugin)
+
+    def send(self, name, data):
+        data["name"] = name
+        for k in list(data):
+            # Drop bytes objects for now
+            if isinstance(data[k], bytes):
+                del data[k]
+        self.write_message(json.dumps(data))
+
+    def open(self):
+        for worker in self.server.workers:
+            self.plugin.add_worker(self.server, worker)
+
+    def on_message(self, message):
+        message = json.loads(message)
+        if message["name"] == "ping":
+            self.send("pong", {"timestamp": str(datetime.now())})
+
+    def on_close(self):
+        self.server.remove_plugin(self.plugin)
+
+
 routes = [
     (r"info", redirect("info/main/workers.html")),
     (r"info/main/workers.html", Workers),
@@ -334,6 +365,7 @@ routes = [
     (r"individual-plots.json", IndividualPlots),
     (r"metrics", PrometheusHandler),
     (r"health", HealthHandler),
+    (r"events", EventsHandler),
     (r"proxy/(\d+)/(.*?)/(.*)", GlobalProxyHandler),
 ]
 
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -322,7 +322,7 @@ class HealthHandler(RequestHandler):
         self.set_header("Content-Type", "text/plain")
 
 
-class EventsHandler(WebSocketHandler):
+class EventstreamHandler(WebSocketHandler):
     def initialize(self, server=None, extra=None):
         self.server = server
         self.extra = extra or {}
@@ -365,7 +365,7 @@ routes = [
     (r"individual-plots.json", IndividualPlots),
     (r"metrics", PrometheusHandler),
     (r"health", HealthHandler),
-    (r"events", EventsHandler),
+    (r"eventstream", EventstreamHandler),
     (r"proxy/(\d+)/(.*?)/(.*)", GlobalProxyHandler),
 ]
 
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -335,7 +335,7 @@ class EventstreamHandler(WebSocketHandler):
             # Drop bytes objects for now
             if isinstance(data[k], bytes):
                 del data[k]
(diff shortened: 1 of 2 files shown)


distributed/dashboard/scheduler.py
```diff
@@ -334,6 +365,7 @@ def get(self):
     (r"individual-plots.json", IndividualPlots),
     (r"metrics", PrometheusHandler),
     (r"health", HealthHandler),
+    (r"events", EventsHandler),
```
mrocklin: Given the experimental nature of this, I would prefer for it to claim a less broad name.  Alternatively, is there some suffix that is common for websockets endpoints by any chance?

distributed/diagnostics/websocket.py
```diff
@@ -0,0 +1,54 @@
+import tornado.websocket
+
+from .plugin import SchedulerPlugin
+from ..utils import key_split
+from .task_stream import colors
+
+
+class WebsocketPlugin(SchedulerPlugin):
+    def __init__(self, socket: tornado.websocket.WebSocketHandler, scheduler):
```
mrocklin: I think that the type here is incorrect.  I think that this is an `EventsHandler`

distributed/dashboard/scheduler.py
```diff
@@ -319,6 +322,34 @@ def get(self):
...
+        self.plugin = WebsocketPlugin(self, server)
+        self.server.add_plugin(self.plugin)
+
+    def send(self, name, data):
+        data["name"] = name
+        for k in list(data):
+            # Drop bytes objects for now
+            if isinstance(data[k], bytes):
+                del data[k]
+        self.write_message(json.dumps(data))
```
mrocklin: It looks like we can just send a dict and Tornado will understand what to do.

mrocklin: Rather than override this method maybe we just send messages like `{"name": "add-worker", "worker": "..."}` and use the vanilla Tornado implementation?  I would prefer avoiding wrapping something like this if we can avoid it.

distributed/dashboard/scheduler.py
```diff
@@ -319,6 +322,34 @@ def get(self):
...
+        for worker in self.server.workers:
+            self.plugin.add_worker(self.server, worker)
+
+    def on_message(self, message):
+        message = json.loads(message)
+        if message["name"] == "ping":
+            self.send("pong", {"timestamp": str(datetime.now())})
+
+    def on_close(self):
+        self.server.remove_plugin(self.plugin)
```
mrocklin: Do we need to `del self.plugin` to remove the reference cycle, or does Python handle this?

distributed/dashboard/scheduler.py
```diff
@@ -319,6 +322,34 @@ def get(self):
         self.set_header("Content-Type", "text/plain")
 
 
+class EventsHandler(WebSocketHandler):
+    def initialize(self, server=None, extra=None):
+        self.server = server
+        self.extra = extra or {}
```
mrocklin: This appears to be unused.

mrocklin:
In principle this seems fine to me.  I left a bunch of nit-picky comments.  I look forward to seeing what you end up doing with this :)

distributed/dashboard/scheduler.py
jacobtomlinson: I assumed this was invoked somewhere. If I don't include this the `self.server` attribute doesn't exist.

distributed/dashboard/scheduler.py
jacobtomlinson: I think Python handles this, however I'm not 100% sure.

distributed/dashboard/scheduler.py
jacobtomlinson: That's reasonable. I'm not sure if there is a common suffix for websockets, generally they seem to be suffix-less.

How about something longer like `eventstream` or `eventsocket`?

distributed/diagnostics/websocket.py
jacobtomlinson: Good point! Not sure why I decided specify the base class instead.

distributed/dashboard/scheduler.py
jacobtomlinson: I'm not sure I follow. There is no `send` message in the base class. Do you mean to use `self.write_message` instead?

distributed/dashboard/scheduler.py
mrocklin: Yes, either of those sound good to me.  We could also elevate this route in the future to a more canonical name if it takes off.

distributed/dashboard/scheduler.py
mrocklin: Yes, rather than say 

```python
self.send("foo", {...})
```

Maybe we call instead

```python
self.write_message({"name": "foo", ...})
```

This method does three things:

1.  Insert name (which we can do manually)
2.  del bytes objects (I don't understand the motivation for this yet, but perhaps this is why this method exists)
3.  call json.dumps (this I think is not necessary, write_message does this by default)

jacobtomlinson: > Insert name (which we can do manually)

I think this was the original reason why I added the method. I wanted to ensure this was consistent. I guess we could manually do this in every place we call this from. (I may have initially overesitamated how often we would call this).

> del bytes objects (I don't understand the motivation for this yet, but perhaps this is why this method exists)

I'm only deleting these because I'm not using them in my demo dashboard and they can't be json dumped. Which isn't really a great reason 😆. Do you know what would happen if we cut out the json dump code and passed through bytes objects?

mrocklin: Ah, I actually didn't realize that bytes weren't json serializable (I just learned this).  

OK, maybe it makes sense to keep this in.

distributed/dashboard/scheduler.py
jacobtomlinson: On a second read of this comment I'm not 100% sure what you are referring to.

jacobtomlinson:
The CI issues appear to be unrelated to this change.

distributed/dashboard/scheduler.py
jacobtomlinson:
```suggestion
        self.write_message(data)
```

distributed/dashboard/scheduler.py
mrocklin: I meant `self.extra`.  I assumed by your previous comment that it actually was necessary though by some upstream class.

mrocklin:
This looks good to me.  I'm happy to merge.  I'll wait a few hours to see if @jacobtomlinson has any follow on comments.

distributed/dashboard/scheduler.py
jacobtomlinson: Ah thanks! After further inspection, I'm not using this. It may be required by something else. I just copy/pasted it from somewhere else (although I can't seem to find where now).

The `extra` kwarg seems to be necessary, so I could alternatively not store it as an attribute, or wrap up the kwargs and optionally store those.

My inclination is to leave things as they are. Are you happy with that?

Issue #3202, opened 2019-11-06
Consider not having an import time side effect (distributed.utils)


rahulporuri:
At the moment, `import distributed` has the import time side-effect of modifying the asyncio event loop policy (if the environment contains asyncio and if the tornado version >= 5).

This is because `distributed.__init__` imports from the `distributed.utils` module, which is where the import time side-effect happens. See below

https://github.com/dask/distributed/blob/bc04d7642c0439bf3757b34bd1eb802b9679ba12/distributed/utils.py#L1191-L1209

This is a problem for desktop applications built around dask/distributed which already use an event loop and aren't expecting `distributed` (or any other package for that matter) to update the event loop policy at any point during the application lifecycle, especially not implicitly.

rahulporuri:
FTR, here's the PR which originally added the relevant code - https://github.com/dask/distributed/pull/1518 - and here are the PRs which touched the relevant code - https://github.com/dask/distributed/pull/2326 and https://github.com/dask/distributed/pull/2343

mrocklin:
Thanks for raising this issue @rahulporuri .  Do you have any suggestions on how to resolve the underlying issue caused when running Dask within a Jupyter notebook?

mrocklin:
@rahulporuri any thoughts on the question above?

potpath:
@mrocklin @rahulporuri , I just encountered this problem when `import distributed` in the ipython startup script. We can reproduce by adding the following to `~/.ipython/profile_default/startup/00-first.py`
```python
import asyncio
import tornado.platform.asyncio

asyncio.set_event_loop_policy(
    tornado.platform.asyncio.AnyThreadEventLoopPolicy()
)
```
I think it is essentially the same issue with https://github.com/dask/distributed/pull/2343.
But this one is when the asyncio event loop is not started yet. I'm not sure if we could just add the `and not is_kernel()` check at https://github.com/mrocklin/distributed/blob/master/distributed/utils.py#L1325.

mrocklin:
Thanks for the comment @potpath ! 

> I'm not sure if we could just add the and not is_kernel() check

is this something that you're able to investigate?  I'll be honest that I personally don't know much about this space.  If you're able to help here that would be very welcome.

potpath:
@mrocklin Sure, opened at https://github.com/dask/distributed/pull/3336.

PR #3336, opened 2019-12-21
Avoid setting event loop policy if within IPython kernel and no running event loop

potpath:
Setting asyncio event loop policy at these two places could cause problems.
1. When policy is set in Jupyter notebook server extension.
This causes the notebook server to hang. This is fixed in https://github.com/dask/distributed/pull/2343.
2. When policy is set in iPython startup config (`~/.ipython/profile_default/startup/whatever.py`) or by setting `get_config().InteractiveShellApp.exec_lines` in `~/.ipython/profile_default/ipython_config.py`.
This causes the kernel to hang. This can be reproduced by running either `jupyter console` or `jupyter notebook`. If running In Jupyter notebook, it will struck at "Kernel starting, please wait...".
Note that manually setting the policy in notebook cell, after the kernel has started, is fine.

In both cases, running `asyncio.get_running_loop()` just before setting the policy will raise `RuntimeError`, meaning there is no running event loop yet.

This MR fix the second problem, while still keeping the first problem fixed

See https://github.com/dask/distributed/issues/3202.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1180,22 +1180,31 @@ def reset_logger_locks():
 # Only bother if asyncio has been loaded by Tornado
 if "asyncio" in sys.modules and tornado.version_info[0] >= 5:
 
-    jupyter_event_loop_initialized = False
+    is_server_extension = False
 
     if "notebook" in sys.modules:
         import traitlets
         from notebook.notebookapp import NotebookApp
 
-        jupyter_event_loop_initialized = traitlets.config.Application.initialized() and isinstance(
+        is_server_extension = traitlets.config.Application.initialized() and isinstance(
             traitlets.config.Application.instance(), NotebookApp
         )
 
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @carreau in case he has time to review this

Carreau:
I haven't touch the eventloop code in quite some time.  I can't think of any issues from the top of my head; and this will likely only affect a small subset of users so I don't see any reasons to not try.

potpath:
@mrocklin Any chance for merging this?

mrocklin:
Sure.  Given @Carreau 's belief that this is likely to be safe, I'm happy to merge.  

Thank you @potpath and my apologies for the delay.  Also, I notice that this is your first code contribution to this repository.  Welcome!

potpath:
No problem @mrocklin , The delay is expected during these time of year.
Thanks for a great experience in contributing and happy new year :fireworks: !

PR #3338, opened 2019-12-22
[WIP] Support different Security class implementations

sodre:
This PR does/will do three things:
  - [ ] Renames distributed.security.Security to distributed.security.TLSSecurity
  - [ ] Creates a new abstract base class called distributed.security.Security. This new class defines the calls to get_connection_args and get_listen_args.
  - [ ] Replaces instanceof checking against TLSSecurity, to the new abstract base class.

Other things to consider:
  - Are implementations of the security class inherently tied to a Comm object? If so, should it move to distributed.comms?
  - Investigate a TLSSecurity mode based on mutual tls authentication, scroll to [Using PKI for Services](https://blog.cloudflare.com/how-to-build-your-own-public-key-infrastructure/), with separate CAs for Client, Server and Workers.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -64,7 +64,7 @@ from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
 from .publish import Datasets
 from .pubsub import PubSubClientExtension
-from .security import Security
+from .security import TLSSecurity
 from .sizeof import sizeof
 from .threadpoolexecutor import rejoin
 from .worker import dumps_task, get_client, get_worker, secede
@@ -660,11 +660,11 @@ class Client(Node):
                 security = getattr(self.cluster, "security", None)
 
         if security is None:
-            security = Security()
+            security = TLSSecurity()
         elif security is True:
-            security = Security.temporary()
+            security = TLSSecurity.temporary()
             self._startup_kwargs["security"] = security
-        elif not isinstance(security, Security):
+        elif not isinstance(security, TLSSecurity):
             raise TypeError("security must be a Security object")
 
         self.security = security
--- a/distributed/security.py
+++ b/distributed/security.py
@@ -10,11 +10,11 @@ except ImportError:
 import dask
 
 
-__all__ = ("Security",)
+__all__ = ("TLSSecurity",)
 
 
-class Security(object):
-    """Security configuration for a Dask cluster.
+class TLSSecurity(object):
+    """TLSSecurity configuration for a Dask cluster.
 
     Default values are loaded from Dask's configuration files, and can be
     overridden in the constructor.
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -15,7 +15,7 @@ from tornado.ioloop import IOLoop
 
 from distributed import Scheduler
 from distributed.preloading import validate_preload_argv
-from distributed.security import Security
+from distributed.security import TLSSecurity
 from distributed.cli.utils import check_python_3, install_signal_handlers
 from distributed.utils import deserialize_for_cli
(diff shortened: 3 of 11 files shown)


distributed/client.py
```diff
@@ -660,11 +660,11 @@ def __init__(
                 security = getattr(self.cluster, "security", None)
 
         if security is None:
-            security = Security()
+            security = TLSSecurity()
```
mrocklin: It's not clear how to handle this case.  The user hasn't provided any security object explicitly.  Whlie assuming `TLSSecurity` makes sense today we might want to generalize this default approach in the future.

distributed/client.py
```diff
@@ -660,11 +660,11 @@ def __init__(
...
         elif security is True:
-            security = Security.temporary()
+            security = TLSSecurity.temporary()
             self._startup_kwargs["security"] = security
-        elif not isinstance(security, Security):
+        elif not isinstance(security, TLSSecurity):
```
mrocklin: My guess is that these type checks should remain against the `Security` type rather than `TLSSecurity`.  Presumably if the user is providing some other Security type then that should be ok too.

sodre: this should be resolved with commit be7bd57.

distributed/client.py
sodre: how about replacing TLSSecurity() with security.get_default_security(). At the moment, it would return TLSSecurity as expected, but could be changed to return something from the configuration files.

mrocklin:
Also cc @jcrist and @mariusvniekerk who might have thoughts

distributed/security.py
```diff
@@ -10,11 +11,70 @@
...
+        return (
+            type(self).__name__
+            + "("
+            + ", ".join("%s=%s" % (k, v) for k, v in items)
+            + ")"
+        )
+
+    @classmethod
+    def temporary(cls):
+        return TLSSecurity.temporary()
```
sodre: This class method should be marked deprecated, and direct the user to TLSSecurity.temporary().

I had to add it, to prevent out-of-tree code from breaking too badly.

sodre: This does not make sense in the new base class, but to reduce the breakage for out-of-tree codebases I added it here.

If we have a `deprecated` decorator, this would be a place to add it.

jcrist:
Can you provide more motivation for these changes? In dask-gateway we subclass the existing Security class to add our own behaviors, which worked but isn't 100% ideal. What kind of security implementation are you using that doesn't rely on TLS and thus wasn't a good fit for the existing codebase?

My preference (if possible) would be to:
- Create a class `BaseSecurity` (or something similar), and use that for any `isinstance` checks in distributed.
- Make the existing class a subclass of `BaseSecurity`

I think that the existing TLS implementation works for most users, and other codebases import `distributed.security.Security` which makes renaming it a bit of a pain.

sodre:
Hi @jcrist, thanks for the feedback! 

I looked at the GatewaySecurity code just now. It looks like inheritance worked moderately well  because the larger concern was around the storage of the TLS certificates, am I correct?

What I have in mind is different, take the case of using GSSAPI for encryption and authentication. In that scenario, the TLS fields to not align with krb5 identities, and inheritance from the existing (TLS)Security class would make very little sense.

Lastly, I see your point w.r.t. to calling the new class `BaseSecurity` instead of the renaming dance that is currently in the PR. Long-term, I think that Security should be renamed TLSSecurity, but that is not needed right now. I will undo the renaming dance.

Lastly, what are your thoughts on replacing the calls of "Security()" to "security.get_default_security()"?  IMO this still not idea, because I see a security object a function of the protocol, and the two sides of the communication, e.g. security between clients and scheduler should be very different to security between scheduler and workers.

PR #3352, opened 2020-01-06
Fix failures on mixed integer/string worker names

filmor:


mrocklin:
Thanks @filmor !  Merging.

PR #3361, opened 2020-01-10
Add cluster map dashboard

jacobtomlinson:
Cluster Map Dashboard demo

This PR adds a first version of the cluster map dashboard shown in #3318.

The dashboard is purely clientside JavaScript and uses the websocket event stream added in #3335. The animation is done with [anime.js](https://animejs.com) and the websocket lifecycle is handled with [reconnecting-websocket](https://github.com/joewalnes/reconnecting-websocket). Both are MIT licensed and bundled in this PR.

I've added the dashboard as an individual plot which will be listed in the `/individual-plots.json` file and loaded by the Jupyter Lab extension. You can also visit http://localhost:8787/statics/individual-cluster-map.html directly.

There is a lot more I would like to do here to improve this. But wanted to get a first draft in for feedback. I can always build upon it with future PRs.


mrocklin:
This is awesome.  Thanks @jacobtomlinson !  I'm looking forward to using this generally, but also specifically when talking about Dask. 

Thoughts on using the `/static` route prefix?  This seems like an internal detail leaking into the user experience. Maybe there isn't a better way though.  

One style question (which we can totally handle later, or never) is directionality of the tasks and communications.  It's not clear which is the source and sink.

mrocklin:
The test failure should be easy to work around.  If we keep the statics prefix then we probably need to exclude the `"bokeh" in body` check in that case.

jacobtomlinson:
> Thoughts on using the /static route prefix?

We could certainly set up a different endpoint for this is you prefer. I just used the existing static endpoint because it existed. As this dashboard is entirely client side and uses the websocket API it needs to be served from a static files endpoint, but we could definitely dedicate one to dashboards.

There are three files serviced to make up the dashboard, the HTML, CSS and JavaScript and then two JavaScript dependencies. Would you want them all under a different location? Or just the Dask specific stuff and still serve the JS deps from `/static`?

> One style question (which we can totally handle later, or never) is directionality of the tasks and communications. It's not clear which is the source and sink.

This may be the limitation of the GIF I provided above, the framerate is a little low. The transfer lines animate to illustrate the direction. Here is a smaller snippet at a higher framerate.

ezgif com-video-to-gif

Is that clear enough or should we think about adding more visual cues (arrows or something)?

mrocklin:
> I just used the existing static endpoint because it existed.

Yup.  Makes sense to me.  I'm just pontificating here.

> The transfer lines animate to illustrate the direction

Thanks, that is more clear once I know to look for it.  I wouldn't bother changing anything at this point.  This all seems like micro-optimization.

That test failure is genuine though.  I suggest that we resolve it and then merge this in.

jacobtomlinson:
> That test failure is genuine though. I suggest that we resolve it and then merge this in.

Sure. I was just waiting until we had decided about `/static` before fixing.

Will get onto it now.

jacobtomlinson:
Instead of testing that the body contains the content we expect in order to check that we got a valid response I've switched to checking the HTTP status code instead. I hope that makes sense.

jacobtomlinson:
Test is passing. This should be good to merge.

@mrocklin

mrocklin:
Thanks Jacob!

PR #3363, opened 2020-01-10
Add client preparing graph hook

jacobtomlinson:
When a client serializes a large task graph and submits it to the scheduler things can appear to hang for before the dashboard kicks into action.

I'd like to add some visual feedback in various places that things are happening client side within the dashboard. 

To start I've added a new `"preparing-update-graph"` RPC call (not set on the name) which a client sends when it is about to start preparing to make an `"update-graph"` call. On the scheduler side this just runs a plugin hook for now but we could expand on this later to show some feedback with the dashboard plugins.

I've also added a couple more plugin hooks for when clients connect and disconnect called `add_client` and `remove_client`, although this could also be `client_connect` and `client_disconnect` if that is preferable.

These are all just hooks for now which could be built off later.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -905,7 +905,6 @@ class Client(Node):
         if self.status in ("running", "closing"):
             try:
                 self.scheduler_comm.send(msg)
-                print("Graph sent")
             except (CommClosedError, AttributeError):
                 if self.status == "running":
                     raise
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -905,6 +905,7 @@ class Client(Node):
         if self.status in ("running", "closing"):
             try:
                 self.scheduler_comm.send(msg)
+                print("Graph sent")
             except (CommClosedError, AttributeError):
                 if self.status == "running":
                     raise
@@ -2433,6 +2434,7 @@ class Client(Node):
         actors=None,
     ):
         with self._refcount_lock:
+            self._send_to_scheduler({"op": "preparing-update-graph"})
             if resources:
                 resources = self._expand_resources(
                     resources, all_keys=itertools.chain(dsk, keys)
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -65,13 +65,7 @@ class SchedulerPlugin(object):
         """ Run when a new worker enters the cluster """
 
     def remove_worker(self, scheduler=None, worker=None, **kwargs):
-        """ Run when a worker leaves the cluster """
-
-    def add_client(self, scheduler=None, client=None, **kwargs):
-        """ Run when a new client connects """
-
-    def remove_client(self, scheduler=None, client=None, **kwargs):
-        """ Run when a client disconnects """
+        """ Run when a worker leaves the cluster"""
 
 
 class WorkerPlugin(object):
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -37,6 +37,9 @@ class SchedulerPlugin(object):
     >>> scheduler.add_plugin(plugin)  # doctest: +SKIP
     """
 
+    def preparing_update_graph(self, scheduler, client=None, **kwargs):
+        """ Run when a new graph / tasks will enter the scheduler soon """
+
     def update_graph(self, scheduler, dsk=None, keys=None, restrictions=None, **kwargs):
         """ Run when a new graph / tasks enter the scheduler """
 
@@ -62,7 +65,13 @@ class SchedulerPlugin(object):
         """ Run when a new worker enters the cluster """
 
     def remove_worker(self, scheduler=None, worker=None, **kwargs):
-        """ Run when a worker leaves the cluster"""
+        """ Run when a worker leaves the cluster """
+
+    def add_client(self, scheduler=None, client=None, **kwargs):
+        """ Run when a new client connects """
+
+    def remove_client(self, scheduler=None, client=None, **kwargs):
+        """ Run when a client disconnects """
 
 
 class WorkerPlugin(object):
--- a/distributed/diagnostics/websocket.py
+++ b/distributed/diagnostics/websocket.py
@@ -20,21 +20,9 @@ class WebsocketPlugin(SchedulerPlugin):
         """ Run when a worker leaves the cluster"""
         self.socket.send("remove_worker", {"worker": worker})
 
-    def add_client(self, scheduler=None, client=None, **kwargs):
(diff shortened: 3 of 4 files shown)


distributed/client.py
```diff
@@ -2433,6 +2433,7 @@ def _graph_to_futures(
         actors=None,
     ):
         with self._refcount_lock:
+            self._send_to_scheduler({"op": "preparing-update-graph"})
```
mrocklin: This may not do what you want.  The `_send_to_scheduler` method puts this message on the BatchedSend comm queue, which batches things for a bit (a couple of milliseconds) and then sends them along.  It's great for when we want to send lots of small messages.

However, we don't actually send that message until at least the next time that we `await`.  There are no await calls between this point and the `update_graph` message that comes at the bottom of this function, so my guess is that the first time that the scheduler gets this message will be at the same time (though just before) it also gets the `update_graph` message.

```python
[
    {"op": "preparing-update-graph"},
    {"op": "update-graph", ...},
]
```

At first thought I'm not sure how best to resolve this.  We intentionally don't have any yield/await points in submit (we want it to be low latency) and we intentionally don't separate small messages in the BatchedSend until after we've already serialized them (it's often efficient to serialize many things at once).

There probably is a good solution here.  I just don't know it yet.  I encourage you to investigate a bit into the pieces of technology mentioned above (BatchedSend, the fact that `_graph_to_futures` is synchronous, ...).  Perhaps a good solution will come to mind.

jacobtomlinson: Thanks for this, and for pointing me on for other places to look.

I've had a bit of a dig through the `_graph_to_futures` method and noticed that [the calls to `unpack_remotedata`](https://github.com/dask/distributed/blob/e7ac25a0e9655cf68f81ee5e30e423b6ab375f4b/distributed/client.py#L2462) are the first point where latency of slow submissions is noticeable. This line is completed after roughly 10% of the total client submission time has elapsed (and therefore roughly 5% of the total time before the dashboard reacts). It seems to be proportional with various graph sizes that I've tested.

Therefore perhaps it would be a good place to introduce a check to see how long has elapsed since the beginning of the function call and if it has been over a threshold (10ms?) we manually trigger the `BatchSend`.

This would involve adding a method to `BatchSend` to allow manually pushing out a batch.

I appreciate that we are going to have to give the IOLoop some control here but given that we know things are not going to be fast I don't see it as a huge problem.

@mrocklin I'd love some feedback on this before I proceed.

mrocklin: >  It seems to be proportional with various graph sizes that I've tested.

This might differ based on how easy it is to serialize things.  I wouldn't be surprised if this is commonly the case, but it's not super robust.  It doesn't need to be though in order to be effective and worthy of a change.  

Another, even less robust mechanism would be to look at the size of the graph.  Maybe for anything over 100 tasks we send it in the background.

> I appreciate that we are going to have to give the IOLoop some control here but given that we know things are not going to be fast I don't see it as a huge problem.
> This would involve adding a method to BatchSend to allow manually pushing out a batch.

The bigger challenge here I think will be making some space for the IOLoop to actually do the send, regardless of whether or not there is API for it.  The current user API doesn't make much space for this.  For example when users call the `submit/map/persist` methods they don't `await` at all. 

```python
async def f():
    df = df.persist()
```

Given this, I don't know how exactly to give the IOLoop time to breathe here.  I suppose that we could launch the submission in a separate coroutine to run in the background while we return control to the user immediately.  There might be some semantic differences here (we'll have futures pointing to tasks well before the scheduler is aware that those tasks are on their way) but that's probably resolvable with modest effort.

jacobtomlinson: I had imagined that we could pull the bulk of `BatchedSend._background_send` out into a separate method, perhaps called `BatchedSend._send_buffer`, which would be called by `_background_send`.

Then we could add an async method called `BatchedSend.send_now` that calls down to `BatchedSend._send_buffer`, which we would then run with `self.sync` from within the `Client._graph_to_futures` method.

Would that solve things here or am I misunderstanding now `self.sync` works?

mrocklin: If the user is in async mode then `self.sync` returns an `tornado.gen.Future` on which we will need to await, or call `asyncio.ensure_future` to make sure that it runs in the background the next time that there is an `await` call.

If we're in asynchronous mode then we won't be able to give up control to the event loop until after the `submit/map/persist` call has returned control to the user and they next await something.

jacobtomlinson: Ah!

jacobtomlinson: @mrocklin do you have thoughts about putting the `BatchedSend_background_send` in a separate loop on its own thread?

It feels like we are never going to escape the fact that `compute/submit/map/persist` are going to be CPU bound for large graphs and block the IO loop.

mrocklin: > do you have thoughts about putting the BatchedSend_background_send in a separate loop on its own thread?

My gut says that this is a bad move.  I would much rather keep everything in a single thread.  This makes debugging much easier.

mrocklin: > It feels like we are never going to escape the fact that compute/submit/map/persist are going to be CPU bound for large graphs and block the IO loop.

When we want to get things off of the event loop we typically use the `offload` function which uses a local concurrent.futures.ThreadPoolExecutor.  I'm not sure how to best integrate that here though.

However, you don't actually care about not blocking the event loop thread, you care about signaling before it gets blocked.  That's still possible while keeping everything on the same event loop.  In principle you could ...

1.  Send a "we're about to send a message" message
2.  Flush BatchedSend
3.  Add a serialize-and-send coroutine onto the event loop 
4.  Return control back to the user
5.  serialize-and-send runs in the background the next time the user releases control

mrocklin: (also, even if all of this fails to produce a final result, I'm quite happy that you're thinking about all of this.  I can see neural pathways developing and bus factors increasing).

jacobtomlinson: > I would much rather keep everything in a single thread. This makes debugging much easier.

mrocklin: We would need to construct the futures before calling that.  In principle this is doable.  All we need in order to construct the future is the key, which we have.  I wouldn't be surprised if implementing this is a little complex, but it should be doable if we want.

mrocklin:
I'm going to go ahead and close this for now.  I think that this is a great idea, but that we may not have the right mechanism here.  Obviously feel free to reopen @jacobtomlinson if you think that there is still something here.

jacobtomlinson:
Thanks @mrocklin. I think closing this is the right call. I definitely want to implement this, it's just a little trickier than I first thought.

PR #3364, opened 2020-01-10
Add --worker-class option to dask-worker CLI

jrbourbeau:
This allows users to specify a custom worker class when using `dask-worker`. For example:

```
dask-worker <scheduler-address> --worker-class mymodule.SpecialWorker
```

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -194,8 +194,8 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "--worker-class",
     type=str,
     default=None,
-    help="Worker class used to instantiate workers from. Defaults to "
-    "distributed.worker.Worker. Only supported when using --nanny.",
+    help="Worker class used to instantiate workers from. Defaults "
+    "to distributed.worker.Worker.",
 )
 @click.option(
     "--lifetime-restart/--no-lifetime-restart",
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -348,12 +348,11 @@ def main(
 
     worker_class = kwargs.pop("worker_class")
     if worker_class is not None:
+        worker_class = import_term(worker_class)
         if nanny:
-            kwargs["worker_class"] = import_term(worker_class)
-        else:
-            raise ValueError(
-                "Cannot use the --worker-class option without using the --nanny option"
-            )
+            kwargs["worker_class"] = worker_class
+    else:
+        worker_class = Worker
 
     if nanny:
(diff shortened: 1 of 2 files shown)


distributed/cli/tests/test_dask_worker.py
```diff
@@ -340,3 +340,48 @@ def test_bokeh_deprecation():
...
             [ws] = s.workers.values()
             assert ws.name == 123
+
+
+WORKER_CLASS_TEXT = """
+from distributed.worker import Worker
+
+class MyWorker(Worker):
+    pass
+"""
```
mrocklin: I would prefer moving this within the test function if that's OK.  Or are you trying to avoid indentation issues?

distributed/cli/dask_worker.py
```diff
@@ -339,13 +347,20 @@ def del_pid_file():
 
     loop = IOLoop.current()
 
+    if worker_class is not None:
+        worker_class = import_term(worker_class)
+        if nanny:
+            kwargs["worker_class"] = worker_class
+    else:
+        worker_class = Worker
```
mrocklin: Maybe change the default to `"dask.distributed.Worker"` and avoid the `is not None` case?

jrbourbeau: Done. Thanks for the suggestion, that's cleaner

distributed/cli/tests/test_dask_worker.py
jrbourbeau: Nope, moving into the test function body seems reasonable

distributed/cli/dask_worker.py
```diff
@@ -190,6 +190,13 @@
     show_default=True,
     help="Random amount by which to stagger lifetime values",
 )
+@click.option(
+    "--worker-class",
+    type=str,
+    default="distributed.worker.Worker",
```
mrocklin: I have a slight preference to publish this name instead long-term and treat it as the canonical location for these classes.  This is what we use in other parts of the docs as well.

```suggestion
    default="dask.distributed.Worker",
```

The `.worker` bit is purely due to our internal organization of files, and isn't something that users should necessarily see.  I'm happy to guarantee that `from dask.distributed import Worker` will always work, but I'm less sure that this class will always be in the worker.py file.

mrocklin:
+1 from me.  I have a slight suggestion on the name to publish, but I'm happy regardless.  Feel free to merge at will.

distributed/cli/dask_worker.py
jrbourbeau: Ah, sorry I misread your original comment

mrocklin:
Thanks @jrbourbeau !  This is in.

PR #3365, opened 2020-01-11
Close connection comm on retry

jrbourbeau:
Currently when making new comm connections, we do so with [`gen.with_timeout`](https://www.tornadoweb.org/en/stable/gen.html#tornado.gen.with_timeout) here

https://github.com/dask/distributed/blob/baf5903c6994a286e02c36f9df48c87776b062d7/distributed/comm/core.py#L215-L219

and retry the connection if a timeout is reached. However, `gen.with_timeout` does not cancel a task when it's timeout expires. Not cancelling these tasks can lead to a "Closing dangling stream" warning being raised here

https://github.com/dask/distributed/blob/baf5903c6994a286e02c36f9df48c87776b062d7/distributed/comm/tcp.py#L169

when `TCP` class instances from previously timed out connection task are garbage collected. 

Switching to [`asyncio.wait_for`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for) allows us to keep a similar timeout-then-retry connection behavior, but also cancels a connection task if a timeout occurs. 

This is just one proposed improvement to the situation. Alternatively we could make the currently hardcoded 1s in the timeout configurable.

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -1,14 +1,12 @@
 from abc import ABC, abstractmethod, abstractproperty
 import asyncio
-from datetime import timedelta
 import logging
 import weakref
 
 import dask
-from tornado import gen
 
 from ..metrics import time
-from ..utils import parse_timedelta, ignoring
+from ..utils import parse_timedelta
 from . import registry
 from .addressing import parse_address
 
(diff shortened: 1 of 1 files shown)


mrocklin:
This seems fine to me.  Should we swap out `gen.with_timeout` generally throughout the codebase?

cc also @jcrist , who has been interested in tornado/asyncio switching code in the past.

mrocklin:
Also, thanks for hunting this down @jrbourbeau !

jcrist:
This looks good to me, thanks @jrbourbeau. Merging.

> Should we swap out gen.with_timeout generally throughout the codebase?

I think we should strive for that, but because it has different semantics replacing it isn't always straightforward. Each case needs to be examined to determine what the intent was and if the lack of cancellation was accidental or necessary.

jcrist:
Oop, just saw you marked this as `[WIP]`. I'll let you merge when you're happy with things. The fix looks good to me as is though.

jrbourbeau:
Thanks for feedback @mrocklin @jcrist! Merging this now

I opened up a separate issue (https://github.com/dask/distributed/issues/3367) for tracking `gen.with_timeout` -> `asyncio.wait_for` switches more broadly throughout the codebase

PR #3366, opened 2020-01-13
Fix scheduler state in case of worker name collision

byjott:
This fixes two bugs that leave the state of the Scheduler in an inconsistent state in case of errors / exceptions.


quasiben:
In general this looks good to me but I am not very familiar with the scheduler codebase.  @mrocklin can i ask you to review ?

mrocklin:
This looks great to me.  Thank you for finding these bugs and fixing them @byjott !

Merging in.

PR #3370, opened 2020-01-14
fix-task-groups

quasiben:
fixes https://github.com/dask/distributed/issues/3368

Previously, tasks were grouped around a key like `("x-123", 0)`.  However, for tasks which are submitted directly to the scheduler:

> h = c.submit(sum, [3, 4])   

the group key (`group_key`) is an empty string.  We can see this more clearly by looking at the following:

```python
from distributed.utils import key_split_group
# `len-442307e1c8fb4925fb9b0fbc7e7eb56d` is a key from a submitted task
key_split_group('len-442307e1c8fb4925fb9b0fbc7e7eb56d')
''
```

This PR rectifies this issue by constructing a new group for each empty group string.  Alternatively, we could consider a non-empty string output for `key_split_group`

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1784,6 +1784,20 @@ async def test_task_prefix(c, s, a, b):
     assert s.task_prefixes["sum-aggregate"].states["memory"] == 1
 
 
+@gen_cluster(client=True)
+async def test_task_unique_groups(c, s, a, b):
+    """ This test ensure that task groups remain unique
+    when using submit
+    """
+    h = await c.submit(sum, [3, 4])
+    g = await c.submit(len, [1, 2])
+
+    assert s.task_prefixes["len"].states["memory"] == 1
+    assert s.task_prefixes["sum"].states["forgotten"] == 1
+    print(s.task_prefixes)
+    print("\n\n")
+
+
 class BrokenComm(Comm):
     peer_address = None
     local_address = None
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1794,8 +1794,6 @@ async def test_task_unique_groups(c, s, a, b):
 
     assert s.task_prefixes["len"].states["memory"] == 1
     assert s.task_prefixes["sum"].states["forgotten"] == 1
-    print(s.task_prefixes)
-    print("\n\n")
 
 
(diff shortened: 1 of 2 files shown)


distributed/tests/test_scheduler.py
```diff
@@ -1784,6 +1784,20 @@ def test_dashboard_address():
...
+    """ This test ensure that task groups remain unique
+    when using submit
+    """
+    h = await c.submit(sum, [3, 4])
+    g = await c.submit(len, [1, 2])
+
+    assert s.task_prefixes["len"].states["memory"] == 1
+    assert s.task_prefixes["sum"].states["forgotten"] == 1
+    print(s.task_prefixes)
+    print("\n\n")
```
mrocklin: It looks like you have some errant print statements here

quasiben: thanks

mrocklin:
@quasiben to aid review can I ask you to include a header comment about what you changed and why?  

Personally I typically structure these as follows, but of course different people have their own styles.

```
Previously we did X, which unfortunately caused Y.  

Now we do Z, which seems to solve Y.  It might also be bad because of Z, but that should be ok because of ...
```

quasiben:
Yup, should've done that before submitting.  Apologies.

> @quasiben to aid review can I ask you to include a header comment about what you changed and why?
> 
> Personally I typically structure these as follows, but of course different people have their own styles.
> 
> ```
> Previously we did X, which unfortunately caused Y.  
> 
> Now we do Z, which seems to solve Y.  It might also be bad because of Z, but that should be ok because of ...
> ```

mrocklin:
Right, so the reasoning behind the empty string group is that currently we don't forget old TaskGroups.  They stick around forever.  This is great because it gives us a lot of history, but would be bad if we created TaskGroups for all of the single-grouped tasks (we would make very many groups that never went away).  This also came up in https://github.com/dask/distributed/pull/3322 (cc @jrbourbeau ) and so was maybe a poor decition.

quasiben:
Thanks for pointer to @jrbourbeau <https://github.com/jrbourbeau>'s PR

bnaul:
just curious, are these two fixes independent or is this blocked by #3322?

quasiben:
closing in favor of https://github.com/dask/distributed/pull/3401

PR #3371, opened 2020-01-14
Add client join and leave hooks

jacobtomlinson:
This is a simpler version of #3363 with just the client join and leave hooks. The discussion around `_send_to_scheduler` is a little trickier than I'd first anticipated so I think it makes sense to break this up into two PRs.

In this PR I've added two scheduler hooks for when clients come and go. I've also updated the websocket plugin to pass along these hooks as well as the update graph hook.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -905,7 +905,6 @@ class Client(Node):
         if self.status in ("running", "closing"):
             try:
                 self.scheduler_comm.send(msg)
-                print("Graph sent")
             except (CommClosedError, AttributeError):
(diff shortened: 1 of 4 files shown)


mrocklin:
Everything here looks good to me.  Does this deserve a test, or is it simple enough that we shouldn't bother?

jacobtomlinson:
I'd say it was simple enough. But I'm happy to add one if you prefer.

mrocklin:
I'm happy to defer to your judgment.

Issue #3367, opened 2020-01-13
Switch gen.with_timeout to asyncio.wait_for


jrbourbeau:
There are several instances where [`gen.with_timeout`](https://www.tornadoweb.org/en/stable/gen.html#tornado.gen.with_timeout) is used in the codebase. As part of our switch from existing tornado code to asyncio, we should move to using [`asyncio.wait_for`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for).

However, one important distinction between these functions is that `gen.with_timeout` does not cancel a task when it's timeout expires, while `asyncio.wait_for`does cancel the task. So we need to be careful and determine on a case by case basis when task cancellation on timeout should happen. From @jcrist's comment (https://github.com/dask/distributed/pull/3365#issuecomment-573342606):

> replacing it isn't always straightforward. Each case needs to be examined to determine what the intent was and if the lack of cancellation was accidental or necessary.

To avoid task cancellation when using `asyncio.wait_for`, the task should first be wrapped in [`asyncio.shield`](https://docs.python.org/3/library/asyncio-task.html#asyncio.shield)

mrocklin:
My guess is that in most cases we'll be fine with cancellation.  I would
encourage a PR that tries this, just to see if the test suite passes.

jrbourbeau:
Closed by https://github.com/dask/distributed/pull/3372

PR #3372, opened 2020-01-15
Replace gen.with_timeout with asyncio.wait_for

jrbourbeau:
As suggested in https://github.com/dask/distributed/issues/3367#issuecomment-573811926, this is to see how the test suite holds up against this change

EDIT:

The changes in this PR are:

- Switched all `gen.with_timeout` occurrences to `asyncio.wait_for` (with one exception in `Client._close` that uses `asyncio.wait`)
- Made corresponding `gen.TimeoutError` -> `asyncio.TimeoutError` changes
- Increased the timeout used when disconnecting workers and the scheduler in our `cluster` testing utility. I found that a 3 second timeout worked for most tests, but did have to special case a few tests with a longer timeout


mrocklin:
It's more work, but this seems like one of those times where it might make sense to do something like ...

```python
from .utils import TimeoutError
```

This way we could change this in one place more consistently in the future.

jrbourbeau:
Good idea, I totally agree that we should do that. However, since tornado's `TimeoutError` is also raised by `IOLoop.run_sync`, I'd prefer to handle it in a seperate PR as it might require lots of additional changes and I think this PR is close to done

mrocklin:
Woot

distributed/client.py
```diff
@@ -1308,7 +1301,11 @@ def _handle_error(self, exception=None):
 
             if not fast:
                 with ignoring(TimeoutError):
-                    await gen.with_timeout(timedelta(seconds=2), list(coroutines))
+
+                    async def _(c):
+                        await asyncio.wait_for(c, 2)
+
+                    await asyncio.gather(*[_(c) for c in coroutines])
```
mrocklin: I'm curious, why the helper function here?

distributed/tests/test_client_executor.py
```diff
@@ -218,30 +227,32 @@ def test_retries(client):
...
-    e.shutdown(wait=False)
-    dt = time.time() - t1
-    assert dt < 0.5
-    time.sleep(0.1)  # wait for future outcome to propagate
-    assert fut.cancelled()
-
-    with pytest.raises(RuntimeError):
-        e.submit(time.sleep, 1.0)
+def test_shutdown(loop):
+    with cluster(disconnect_timeout=10) as (s, [a, b]):
```
mrocklin: I'm happy to see this test just because I know that it means that you dove into the utils_test.py infrastructure.  Hooray for increasing bus factors!

mrocklin:
Nice green check marks :)

jrbourbeau:
I think this should be ready for review now. I updated my original comment to summarize the changes

mrocklin:
I went through things a bit ago and I'm +1 on this (maybe even +2).  cc'ing @jcrist who was interested in this topic previously.  

As a reminder, probably no one should press the green button until we trigger the next release.

distributed/client.py
```diff
@@ -1268,12 +1265,8 @@ def _handle_error(self, exception=None):
...
             # Give the scheduler 'stream-closed' message 100ms to come through
             # This makes the shutdown slightly smoother and quieter
-            with ignoring(AttributeError, gen.TimeoutError):
-                await gen.with_timeout(
-                    timedelta(milliseconds=100),
-                    self._handle_scheduler_coroutine,
-                    quiet_exceptions=(CancelledError,),
-                )
+            with ignoring(AttributeError, CancelledError):
+                await asyncio.wait([self._handle_scheduler_coroutine], timeout=0.1)
```
jcrist: This just blocks on `_handle_scheduler_coroutine` until 0.1 seconds pass or the coroutine completes/is cancelled. Is that what's intended here? Why not cancel it here on timeout?

If this is correct, I find the following easier to read:

```python
await asyncio.wait_for(asyncio.shield(self._handle_scheduler_coroutine), 0.1)
```

distributed/client.py
jcrist: Why not

```python
await asyncio.wait_for(asyncio.gather(*coroutines), 2)
```

This blocks until:
- all coroutines complete
- or 2 seconds passes

On timeout, all coroutines in `coroutines` are cancelled and a `TimeoutError` is raised.

jrbourbeau: Ah, yes. That's much cleaner. Thanks for the suggestion @jcrist!

distributed/client.py
jrbourbeau: Using `wait_for` with an 0.1s timeout was causing timeout error test failures. I tried bumping the timeout up but was still getting non-zero failures at 30s. Then I noticed we were previously ignoring `gen.TimeoutError`s here, so assumed that timeouts were okay in some cases.

jcrist: Hmmm, ok. Either as it is now, or as I wrote above (and adding `asyncio.TimeoutError` to the `ignoring`) would be equivalent. I'm not sure which I find easier to read.

jrbourbeau: I like using `wait_for` + `shield`. It seems more explicit about what's going on, which is good in this case

distributed/tests/test_client_executor.py
jrbourbeau: `bus_factor += 1`

jrbourbeau:
Planning to merge in the next couple of hours if there are no further comments

mrocklin:
There are still a number of `gen.TimeoutError`s in the code.  Should we switch these out for `asyncio.TimeoutError` generally?  I ran into an intermittent failure in `test_utils.py::test_sync` where we sometimes throw an `asyncio.TimeoutError` rather than a `gen.TimeoutError`

PR #3378, opened 2020-01-16
Serialize cuDF.Index

jakirkham:
cc @kkraus14

### DIFF
--- a/distributed/protocol/cudf.py
+++ b/distributed/protocol/cudf.py
@@ -9,7 +9,7 @@ from ..utils import log_errors
 # here we ammend the returned headers with `is_gpu` for
 # UCX buffer consumption
 @cuda_serialize.register(
-    (cudf.DataFrame, cudf.Series, cudf.groupby.groupby._Groupby)
+    (cudf.DataFrame, cudf.Index, cudf.Series, cudf.groupby.groupby._Groupby)
 )
 def serialize_cudf_dataframe(x):
     with log_errors():
@@ -18,7 +18,7 @@ def serialize_cudf_dataframe(x):
 
 
 @cuda_deserialize.register(
-    (cudf.DataFrame, cudf.Series, cudf.groupby.groupby._Groupby)
+    (cudf.DataFrame, cudf.Index, cudf.Series, cudf.groupby.groupby._Groupby)
 )
 def deserialize_cudf_dataframe(header, frames):
     with log_errors():
--- a/distributed/protocol/cudf.py
+++ b/distributed/protocol/cudf.py
@@ -8,14 +8,18 @@ from ..utils import log_errors
 # all (de-)serializtion code lives in the cudf codebase
 # here we ammend the returned headers with `is_gpu` for
 # UCX buffer consumption
-@cuda_serialize.register((cudf.DataFrame, cudf.Series, cudf.groupby.groupby._Groupby))
+@cuda_serialize.register(
+    (cudf.DataFrame, cudf.Series, cudf.groupby.groupby._Groupby)
+)
 def serialize_cudf_dataframe(x):
     with log_errors():
         header, frames = x.serialize()
         return header, frames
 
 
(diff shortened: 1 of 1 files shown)


distributed/protocol/cudf.py
```diff
@@ -8,14 +8,18 @@
 # all (de-)serializtion code lives in the cudf codebase
 # here we ammend the returned headers with `is_gpu` for
 # UCX buffer consumption
-@cuda_serialize.register((cudf.DataFrame, cudf.Series, cudf.groupby.groupby._Groupby))
+@cuda_serialize.register(
+    (cudf.DataFrame, cudf.Index, cudf.Series, cudf.groupby.groupby._Groupby)
```
rjzamora: It seems that this file is no longer used after cudf-0.9 - `cudf.groupby.groupby._Groupby` will no longer work.

jakirkham: Should we drop the file altogether then?

rjzamora: > Should we drop the file altogether then?

I'm in support of that :)

jakirkham: @quasiben @madsbk @pentschev, any thoughts on dropping this file or do we still need it?

pentschev: I'm fine with it, I don't know if it's still relevant but note that there's a comment about UCX:

```
# all (de-)serializtion code lives in the cudf codebase
# here we ammend the returned headers with `is_gpu` for
# UCX buffer consumption
```

Maybe @quasiben can chime in.

rjzamora: Right - we should wait for Ben's opinion.  However, we do know [this file is most likely ignored](https://github.com/dask/distributed/blob/6a76ca7ac6db9f6243869057b6805f14c23e3811/distributed/protocol/__init__.py#L88) for recent cudf versions

quasiben: i think we can drop this file.  We kept it around when moving to 0.9.  The code for switching can be found here:
https://github.com/dask/distributed/blob/6a76ca7ac6db9f6243869057b6805f14c23e3811/distributed/protocol/__init__.py#L83-L91

So if we drop the file we should update the __init__

jakirkham: Yeah I saw that comment to @pentschev. Though I'm not actually seeing any use of `is_gpu` in the code. Perhaps the comment is outdated?

quasiben: The comment is outdated.  `is_gpu` was used in the previous ucx-py version when you needed to explicitly receive into a GPU buffer https://github.com/dask/distributed/blob/master/distributed/comm/ucx.py#L85-L87

madsbk: > The comment is outdated. `is_gpu` was used in the previous ucx-py version when you needed to explicitly receive into a GPU buffer https://github.com/dask/distributed/blob/master/distributed/comm/ucx.py#L85-L87

I think `is_gpu` is still used to determine when to use `cuda_array()`:
https://github.com/dask/distributed/blob/master/distributed/comm/ucx.py#L139
https://github.com/dask/distributed/blob/master/distributed/comm/ucx.py#L169

quasiben: @madsbk doesn't that use the `__cuda_array_interface__` interface instead of `is_gpus` ?

madsbk: Sorry, I thought you talked about the UCX communication protocol where the writer sends ` is_gpu` to the reader.

But yes, cuDF doesn't need the `is_gpu` amendment anymore. 
However, I think we still need to register a  serialization function for `cudf.DataFrame, cudf.Index, cudf.Series` right?

jakirkham: Accordingly to @rjzamora this is now handled internally within cuDF. So it sounds like we can drop the whole file.

jakirkham:
Replacing with PR ( https://github.com/dask/distributed/pull/3404 ). PTAL 🙂

Issue #3381, opened 2020-01-17
AttributeError: module 'asyncio' has no attribute 'get_running_loop'


haje01:
In Jupyter Lab, If I run:

`import distributed`

causes `AttributeError: module 'asyncio' has no attribute 'get_running_loop'` error.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-432545581521> in <module>()
----> 1 import distributed

~/anaconda3/lib/python3.6/site-packages/distributed/__init__.py in <module>()
      1 from . import config
      2 from dask.config import config
----> 3 from .actor import Actor, ActorFuture
      4 from .core import connect, rpc
      5 from .deploy import LocalCluster, Adaptive, SpecCluster, SSHCluster

~/anaconda3/lib/python3.6/site-packages/distributed/actor.py in <module>()
      5 from queue import Queue
      6 
----> 7 from .client import Future, default_client
      8 from .protocol import to_serialize
      9 from .utils import sync

~/anaconda3/lib/python3.6/site-packages/distributed/client.py in <module>()
     47 from asyncio import iscoroutine
     48 
---> 49 from .batched import BatchedSend
     50 from .utils_comm import (
     51     WrappedKey,

~/anaconda3/lib/python3.6/site-packages/distributed/batched.py in <module>()
      6 from tornado.ioloop import IOLoop
      7 
----> 8 from .core import CommClosedError
      9 from .utils import parse_timedelta
     10 

~/anaconda3/lib/python3.6/site-packages/distributed/core.py in <module>()
     16 from tornado.locks import Event
     17 
---> 18 from .comm import (
     19     connect,
     20     listen,

~/anaconda3/lib/python3.6/site-packages/distributed/comm/__init__.py in <module>()
----> 1 from .addressing import (
      2     parse_address,
      3     unparse_address,
      4     normalize_address,
      5     parse_host_port,

~/anaconda3/lib/python3.6/site-packages/distributed/comm/addressing.py in <module>()
      3 
      4 from . import registry
----> 5 from ..utils import get_ip_interface
      6 
      7 

~/anaconda3/lib/python3.6/site-packages/distributed/utils.py in <module>()
   1205         if is_kernel():
   1206             try:
-> 1207                 asyncio.get_running_loop()
   1208             except RuntimeError:
   1209                 is_kernel_and_no_running_loop = True

AttributeError: module 'asyncio' has no attribute 'get_running_loop'
```

My Environment:
[...]

PR #3383, opened 2020-01-17
Fix asyncio.get_running_loop import for Python < 3.7

jrbourbeau:
This PR adds a compatibility check to make sure we don't import `asyncio.get_running_loop` on Python < 3.7

I confirmed locally that `import distributed` works in Jupyter with Python 3.6

Closes https://github.com/dask/distributed/issues/3381


jrbourbeau:
Thanks for reviewing @TomAugspurger! Once CI passes, I'll merge this and cut a new release

PR #3387, opened 2020-01-18
Respect dashboard prefix when redirecting root location

### LINKED ISSUE #3405: When --dashboard-prefix is used root location is always redirected to /status without adding the prefix
The problem arises when using --dashboard-prefix, using http://DASHBOARD[:PORT]/PREFIX/ redirects to http://DASHBOARD[:PORT]/status instead of http://DASHBOARD[:PORT]/PREFIX/status.

cnanakos:
When --dashboard-prefix is used root location is always redirected to
/status without adding the prefix.

Fixes #3405

### DIFF
--- a/distributed/dashboard/core.py
+++ b/distributed/dashboard/core.py
@@ -5,6 +5,7 @@ import warnings
 import bokeh
 from bokeh.server.server import Server
 from tornado import web
+from urllib.parse import urljoin
 
 
 if LooseVersion(bokeh.__version__) < LooseVersion("0.13.0"):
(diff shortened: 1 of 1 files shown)


TomAugspurger:
@jacobtomlinson can you take a look at this? @cnanakos is there an issue associated with this PR? If not, can you provide a bit more context about what this is fixing?

cnanakos:
@jacobtomlinson sorry no, there is no associated issue with this PR. The problem arises when using the --dashboard-prefix, using http://DASHBOARD:PORT/PREFIX/ redirects to http://DASHBOARD:PORT/status instead of http://DASHBOARD:PORT/PREFIX/status.

This PR adds the prefix to the root location when redirected.

distributed/dashboard/core.py
```diff
@@ -34,7 +34,9 @@ def listen(self, addr):
                     check_unused_sessions_milliseconds=500,
                     allow_websocket_origin=["*"],
                     use_index=False,
-                    extra_patterns=[(r"/", web.RedirectHandler, {"url": "/status"})],
+                    extra_patterns=[(r"/", web.RedirectHandler, {"url":
+                                                                 self.prefix +
```
jacobtomlinson: Will there ever be situations where `self.prefix` has a trailing slash? In that situation, we would end up with `profix//status`.

Perhaps using a joiner like `urllib.parse.urljoin` or `posixpath.join` may be more robust here?

cnanakos:
@jacobtomlinson opened an issue as requested, rebased current PR, used `urllib.parse.urljoin` to avoid trailing-slashes and all tests are passing now. thanks for reviewing so far.

PR #3388, opened 2020-01-19
Use latest release of black

jrbourbeau:
We pinned `black` to a specific commit to avoid formatting issues with Python 3.5. We've since dropped support for 3.5, so we should be able to use the latest release of `black` now.

This will help avoid discrepancies where users `pip install black` and manually run `black dask` instead of using the pre-commit hook

Corresponding PR in `dask/dask`: https://github.com/dask/dask/pull/5798


mrocklin:
Thanks @jrbourbeau !

Issue #3379, opened 2020-01-16
Version checking introduced in pull/2627 fails if worker has older version of `distributed`


ig248:
The issue appears if scheduler has `distributed` at SHA>=https://github.com/dask/distributed/commit/32cb96effe9287fdc42dd73be36989be13abe99d and worker is at SHA<https://github.com/dask/distributed/commit/32cb96effe9287fdc42dd73be36989be13abe99d

PR: https://github.com/dask/distributed/pull/2627

It appears that within the spirit of adding version checking, this mis-match should be handled gracefully with an informative error message.

Minimal `docker-compose` example to reproduce:
```
# docker-compose.yaml
version: "3.1"

services:
  scheduler:
    image: daskdev/dask:latest
    command: ["dask-scheduler", "--port", "54321"]
    environment:
      - EXTRA_PIP_PACKAGES=git+https://github.com/dask/distributed@32cb96effe9287fdc42dd73be36989be13abe99d
  worker:
    image: daskdev/dask:latest
    command: ["dask-worker", "tcp://scheduler:54321", "--nthreads", "1", "--no-dashboard"]
```

error trace:
```
scheduler_1  | distributed.scheduler - INFO - Register worker <Worker 'tcp://172.23.0.3:34809', name: tcp://172.23.0.3:34809, memory: 0, processing: 0>
scheduler_1  | distributed.utils - ERROR - 'NoneType' object is not subscriptable
scheduler_1  | Traceback (most recent call last):
scheduler_1  |   File "/opt/conda/lib/python3.7/site-packages/distributed/utils.py", line 662, in log_errors
scheduler_1  |     yield
scheduler_1  |   File "/opt/conda/lib/python3.7/site-packages/distributed/scheduler.py", line 1720, in add_worker
scheduler_1  |     client_name="This Worker",
scheduler_1  |   File "/opt/conda/lib/python3.7/site-packages/distributed/versions.py", line 105, in error_message
scheduler_1  |     client_versions = client["packages"]
scheduler_1  | TypeError: 'NoneType' object is not subscriptable
scheduler_1  | distributed.core - ERROR - 'NoneType' object is not subscriptable
scheduler_1  | Traceback (most recent call last):
scheduler_1  |   File "/opt/conda/lib/python3.7/site-packages/distributed/core.py", line 412, in handle_comm
scheduler_1  |     result = await result
[...]

PR #3390, opened 2020-01-19
Make version mismatch checking more robust

ig248:
This PR address #3379 

The entire change is confined to `distributed.versions.error_message()`.

The PR should allow handling following with informative messages:
1. one (or more) nodes (scheduler/worker/client) has a mismatched version. E.g.:
`client["packages"]["distributed"] == 2.9.2`, but `client["packages"]["distributed"] == 2.9.3`.

2. one node does not appear to have a package, though same package is available on one of the other nodes - we report it as `MISSING`

3. one of the nodes does not conform to the latest version info format:
  - it could be returning `None`, as in #3379 - i.e. a worker `@distributed<=2.9.1`
  - it could be missing the ["packages"] key (older versions)
  In these cases, we record the package version as `UNKNOWN`

## Message examples
(only excerpts shown)
```
distributed
+-----------+-------------------------+
|           | version                 |
+-----------+-------------------------+
| client    | 2.9.3+4.g366851e2.dirty |
| scheduler | 2.9.3+4.g366851e2.dirty |
| worker-0  | 2.9.3+4.g366851e2.dirty |
| worker-1  | 0.0.0.dev0              |
| worker-2  | 2.9.3+4.g366851e2.dirty |
+-----------+-------------------------+
```
```
tornado
+-----------+---------+
|           | version |
+-----------+---------+
| client    | 6.0.3   |
| scheduler | UNKNOWN |
| worker-0  | 6.0.3   |
| worker-1  | 6.0.3   |
| worker-2  | 6.0.3   |
+-----------+---------+
```
```
lz4
+-----------+---------+
|           | version |
+-----------+---------+
| client    | None    |
| scheduler | UNKNOWN |
| worker-0  | None    |
| worker-1  | None    |
| worker-2  | None    |
+-----------+---------+
```
```
distributed
+-----------+-------------------------+
|           | version                 |
+-----------+-------------------------+
| client    | MISSING                 |
| scheduler | 2.9.3+4.g366851e2.dirty |
| worker-0  | 2.9.3+4.g366851e2.dirty |
| worker-1  | 2.9.3+4.g366851e2.dirty |
| worker-2  | 2.9.3+4.g366851e2.dirty |
+-----------+-------------------------+
```

### DIFF
--- a/distributed/versions.py
+++ b/distributed/versions.py
@@ -26,6 +26,12 @@ optional_packages = [
 ]
 
 
+# only these scheduler packages will be checked for version mismatch
+scheduler_relevant_packages = set(pkg for pkg, _ in required_packages) | set(
+    ["lz4", "blosc"]
+)
+
+
 def get_versions(packages=None):
     """
     Return basic information on our software installation, and our installed versions of packages.
@@ -119,7 +125,11 @@ def error_message(scheduler, workers, client, client_name="client"):
 
     errs = []
     for pkg in sorted(packages):
-        versions = set(node_packages[node][pkg] for node in nodes)
+        versions = set(
+            node_packages[node][pkg]
+            for node in nodes
+            if node != "scheduler" or pkg in scheduler_relevant_packages
+        )
         if len(versions) <= 1:
             continue
         rows = [
--- a/distributed/versions.py
+++ b/distributed/versions.py
@@ -28,7 +28,7 @@ optional_packages = [
 
 def get_versions(packages=None):
     """
-    Return basic information on our software installation, and out installed versions of packages.
+    Return basic information on our software installation, and our installed versions of packages.
     """
     if packages is None:
         packages = []
--- a/distributed/versions.py
+++ b/distributed/versions.py
@@ -100,10 +100,7 @@ def get_package_info(pkgs):
 def error_message(scheduler, workers, client, client_name="client"):
     from .utils import asciitable
 
-    MISSING, UNKNOWN = "MISSING", "UNKNOWN"
-    scheduler_name = "scheduler"
-
-    nodes = {**{client_name: client}, **{scheduler_name: scheduler}, **workers}
+    nodes = {**{client_name: client}, **{"scheduler": scheduler}, **workers}
 
     # Hold all versions, e.g. versions["scheduler"]["distributed"] = 2.9.3
     node_packages = defaultdict(dict)
@@ -113,9 +110,9 @@ def error_message(scheduler, workers, client, client_name="client"):
 
     for node, info in nodes.items():
         if info is None or not (isinstance(info, dict)) or "packages" not in info:
-            node_packages[node] = defaultdict(lambda: UNKNOWN)
+            node_packages[node] = defaultdict(lambda: "UNKNOWN")
         else:
-            node_packages[node] = defaultdict(lambda: MISSING)
+            node_packages[node] = defaultdict(lambda: "MISSING")
             for pkg, version in info["packages"].items():
                 node_packages[node][pkg] = version
                 packages.add(pkg)
--- a/distributed/versions.py
+++ b/distributed/versions.py
@@ -98,34 +98,39 @@ def get_package_info(pkgs):
 
 
 def error_message(scheduler, workers, client, client_name="client"):
-    # we care about the required & optional packages matching
-    try:
-        client_versions = client["packages"]
-        versions = [("scheduler", scheduler["packages"])]
-        versions.extend((w, d["packages"]) for w, d in sorted(workers.items()))
-    except KeyError:
-        return (
-            "Version mismatch for dask.distributed. "
-            "The scheduler has version >= 1.28.0 "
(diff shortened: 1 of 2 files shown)


ig248:
@mrocklin I noticed that version mismatch handling was not "symmetric" with respect to nodes. In this PR I made it  symmetric, though I wonder if I missed an important case that requires treating e.g. client and scheduler differently?

distributed/versions.py
```diff
@@ -98,34 +98,38 @@ def get_package_info(pkgs):
...
-
-        errs = []
-        for pkg, versions in sorted(mismatched.items()):
-            rows = [(client_name, client_versions[pkg])]
-            rows.extend(versions)
-            errs.append("%s\n%s" % (pkg, asciitable(["", "version"], rows)))
+    from .utils import asciitable
 
+    MISSING, UNKNOWN = "MISSING", "UNKNOWN"
+    scheduler_name = "scheduler"
```
mrocklin: Can we inline these rather than have constants?  I would prefer this so that future readers have one less thing to refer back to as they read the code.

distributed/versions.py
```diff
@@ -98,34 +98,38 @@ def get_package_info(pkgs):
...
+
+    nodes = {**{client_name: client}, **{scheduler_name: scheduler}, **workers}
```
mrocklin: This seems needlessly complex.  Maybe the following instead?

```suggestion
    nodes = {client_name: client, "scheduler": scheduler, **workers}
```

mrocklin:
> I noticed that version mismatch handling was not "symmetric" with respect to nodes. In this PR I made it symmetric, though I wonder if I missed an important case that requires treating e.g. client and scheduler differently?

I don't recall the asymettry that this code has in it in particular (I've forgotten most of the details about writing this code).   In general though the scheduler doesn't need to have user libraries like Numpy and Pandas around.  It only needs to have Dask and the various compression and serialization libraries.

mrocklin:
Also, when checking out this repository I found that I had an un-checked-in test_versions.py.  I'm including it here in case it's useful

```python
import pytest

from dask.distributed import Client, Worker
from distributed.versions import get_versions, error_message
from distributed.utils_test import gen_cluster


def test_error_message():
    a = get_versions()
    b = get_versions()
    b["packages"]["dask"] = "0.0.0"

    msg = error_message(a, {"w-1": b}, a)
    assert msg
    assert "dask" in msg
    assert "0.0.0" in msg


@gen_cluster()
async def test_warning(s, a, b):
    s.workers[a.address].versions["packages"]["dask"] = "0.0.0"

    with pytest.warns(None) as record:
        async with Client(s.address, asynchronous=True) as client:
            pass

    assert record
    assert any("dask" in str(r.message) for r in record)
    assert any("0.0.0" in str(r.message) for r in record)
    assert any(a.address in str(r.message) for r in record)

        assert any("This Worker" in line.message for line in w.logs)
        assert any("dask" in line.message for line in w.logs)
        assert any(
            "0.0.0" in line.message and a.address in line.message for line in w.logs
        )
```

ig248:
@mrocklin Is it accurate to say that `required_packages` must agree between all Nodes, and other packages should agree between client(s) and worker(s) only? I would be happy to handle this logic

distributed/versions.py
ig248: :+1: was just for symmetry...

mrocklin:
> Is it accurate to say that required_packages must agree between all Nodes, and other packages should agree between client(s) and worker(s) only? I would be happy to handle this logic

That is currently not accurate, but maybe it should be.  The scheduler must also agree on compression libraries like `lz4` and `blosc`, *if they are present*.

ig248:
OK, I will create a separate `scheduler_packages` list (or similar) to capture the different groups of dependencies

mrocklin:
Thank you for taking this on @ig248 !

distributed/versions.py
ig248: :heavy_check_mark:

distributed/versions.py
ig248: :heavy_check_mark: inlined

ig248:
> Thank you for taking this on @ig248 !

Thank you, I found it extremely helpful to better understand the internals of `dask`.

ig248:
@mrocklin I have added [`scheduler_relevant_packages`](https://github.com/dask/distributed/blob/44e5f24fca0fbb076018bebb2045a6c105a691b2/distributed/versions.py#L30) for the discussed purpose, and version checking now ignores scheduler packages outside this set - see [tests](https://github.com/dask/distributed/pull/3390/commits/fd46b8c325bd168a6b9e8c90c6bb6ceb7de32c5f).

mrocklin:
OK.  I'm going to merge this in.  My apologies for being late here (it has been a busy week).  Thank you @ig248 !

> Thank you, I found it extremely helpful to better understand the internals of dask.

I'm very glad to hear it!  I notice that this is your first code contribution to this repository.  I hope that it is not your last :)  Welcome!

PR #3391, opened 2020-01-20
Add lifecycle hooks to SchedulerPlugin

mrocklin:
This adds start and close async functions to Scheduler Plugins


mrocklin:
I've added some very short comments that we start at the end of setup and close at the beginning of shutdown.

TomAugspurger:
Thanks!

PR #3392, opened 2020-01-20
Support args and kwargs in offload

mrocklin:


jrbourbeau:
Just pushed a commit to bring this PR up to date with our latest `master`. That may resolve the `TimeoutError` we were seeing before

PR #3394, opened 2020-01-20
Replace gen.TimeoutError with utils.TimeoutError

mrocklin:
Previously we had a mix of gen.TimeoutError and asyncio.TimeoutError
within the code.  This import asyncio.TimeoutError within utils.py (and
our top level imports) and then uses that consistently throughout the
codebase

cc @jrbourbeau 

(disclaimer, I haven't yet fully tested this, so this may still need some love)

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -24,12 +24,10 @@ from .pubsub import Pub, Sub
 from .queues import Queue
 from .scheduler import Scheduler
 from .threadpoolexecutor import rejoin
-from .utils import sync
+from .utils import sync, TimeoutError
 from .variable import Variable
 from .worker import Worker, get_worker, get_client, secede, Reschedule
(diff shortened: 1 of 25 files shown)


jrbourbeau:
Thanks for the PR @mrocklin, I think this is moving in the right direction. My initial thought is that since [`IOLoop.run_sync`](https://www.tornadoweb.org/en/stable/util.html#tornado.util.TimeoutError) can raise `gen.TimeoutError`, we may need to have `distributed.utils.TimeoutError` be `(asyncio.TimeoutErro, gen.TimeoutError)`.

I'm out of the office today but will give this a more thorough look later this evening or tomorrow

mrocklin:
@jrbourbeau ok to merge?

cc also @jcrist

jrbourbeau:
Thanks @mrocklin, taking a look at the changes right now

PR #3398, opened 2020-01-22
development document revisions

dazza-codes:
Maybe the docs are deliberately sparse and this PR can be trashed, but a little extra detail might be useful for some folks.

While using the example unit tests from the dev-doc, ran into some problems that might find a way to get updated in this or another PR.  The following is not (yet) in this PR because it's not entirely clear how it should work, i.e. the working code with comments on failures is like:

```python
# import all the dask distributed pytest fixtures
from distributed import Future
from distributed.utils_test import *  # pylint: disable=wildcard-import


def test_sync_submit(client):
    future = client.submit(inc, 10)
    assert isinstance(future, Future)
    assert future.result() == 11  # use the synchronous/blocking API here


def test_sync_submit_full(client, s, a, b):
    """
    In this style of test you do not have access to the scheduler or workers.
    The variables s, a, b are now dictionaries holding a multiprocessing.Process
    object and a port integer. However, you can now use the normal synchronous
    API (never use yield in this style of test) and you can close processes
    easily by terminating them.
    """
    assert isinstance(client, Client)
    assert isinstance(s, dict)
    assert client.scheduler.address == s['address']
    assert isinstance(a, dict)  # worker-a
    assert isinstance(b, dict)  # worker-b

    future = client.submit(inc, 10)
    assert isinstance(future, Future)
    assert future.result() == 11  # use the synchronous/blocking API here

    fixed_test = False
    if fixed_test:
        # killing a worker fails the test cleanup checks:
        # Failed: some RPCs left active by test

        worker_process = a['proc']()  # call weakref to get worker process
        worker_process.terminate()  # kill one of the workers
        # client.retire_workers()  # also not a clean test teardown

        result = future.result()  # test that future remains valid
        assert isinstance(future, Future)
        assert result == 11
```

### DIFF
--- a/docs/source/develop.rst
+++ b/docs/source/develop.rst
@@ -78,28 +78,33 @@ The test suite contains three kinds of tests
     These are rare and mostly for testing the command line interface.
 
 If you are comfortable with the Tornado interface then you will be happiest
-using the ``@gen_cluster`` style of test
+using the ``@gen_cluster`` style of test, e.g.
 
 .. code-block:: python
 
-   from distributed.utils_test import gen_cluster
+    # tests/test_submit.py
 
-   @gen_cluster(client=True)
-   def test_submit(c, s, a, b):
-       assert isinstance(c, Client)
-       assert isinstance(s, Scheduler)
-       assert isinstance(a, Worker)
-       assert isinstance(b, Worker)
+    from distributed.utils_test import gen_cluster, inc
+    from distributed import Client, Future, Scheduler, Worker
 
-       future = c.submit(inc, 1)
-       assert future.key in c.futures
+    @gen_cluster(client=True)
+    def test_submit(c, s, a, b):
+        assert isinstance(c, Client)
+        assert isinstance(s, Scheduler)
+        assert isinstance(a, Worker)
+        assert isinstance(b, Worker)
 
-       # result = future.result()  # This synchronous API call would block
-       result = yield future
-       assert result == 2
+        future = c.submit(inc, 1)
+        assert isinstance(future, Future)
+        assert future.key in c.futures
 
-       assert future.key in s.tasks
-       assert future.key in a.data or future.key in b.data
+        # result = future.result()  # This synchronous API call would block
+        result = yield future
+        assert result == 2
+
+        assert future.key in s.tasks
+        assert future.key in a.data or future.key in b.data
+
 
 The ``@gen_cluster`` decorator sets up a scheduler, client, and workers for
 you and cleans them up after the test.  It also allows you to directly inspect


docs/source/develop.rst
```diff
@@ -78,28 +98,42 @@ The test suite contains three kinds of tests
...
-   from distributed.utils_test import gen_cluster
+    # tests/test_submit.py
 
-   @gen_cluster(client=True)
-   def test_submit(c, s, a, b):
-       assert isinstance(c, Client)
-       assert isinstance(s, Scheduler)
-       assert isinstance(a, Worker)
-       assert isinstance(b, Worker)
+    from distributed.utils_test import gen_cluster
```
jrbourbeau: Can you condense these imports to import multiple things per line. Something like:

```python
from distributed.utils_test import gen_cluster, inc
from distributed import Client, Scheduler, Worker, Future
```

docs/source/develop.rst
```diff
@@ -78,28 +98,42 @@ The test suite contains three kinds of tests
...
+
+        # result = future.result()  # This synchronous API call would block
+        result = yield future
+        assert result == 2
+
+        assert future.key in s.tasks
+        assert future.key in a.data or future.key in b.data
+
+
+.. code-block:: shell
```
jrbourbeau: I think we can remove this code snippet as this section is generally about writing tests, not running them (that's described in the Dask contributing guide)

docs/source/develop.rst
```diff
@@ -22,6 +22,26 @@ and install it from source::
...
+
+   git clone git@github.com:{your-fork}/distributed.git
+   cd distributed
+   conda create -y -n distributed python=3.6
+   conda activate distributed
+   pip install -U -r requirements.txt
+   pip install -U -r dev-requirements.txt
+   pip install -e .
+
+To keep a fork in sync with the upstream source::
```
jrbourbeau: In the Dask developer docs we link to Pandas' documentation on using `git` which is quite thorough. I'd prefer to leave this section out in favor of users going to the relevant section in the Pandas docs

docs/source/develop.rst
```diff
@@ -22,6 +22,26 @@ and install it from source::
    cd distributed
    python setup.py install
 
+Using conda, for example::
```
jrbourbeau: Since the Dask docs go through setting up a `conda` environment, could we remove this section in favor of a sentence like "A list of dependencies for `distributed` is maintained in `requirements.txt`, while additional dependencies needed for development are listed in `dev-requirements.txt`"?

docs/source/develop.rst
dazza-codes: Whie it's easier to keep things DRY from a documentation perspective, extra links does add a burden on the user to skip around and sift through the volumes of docs about pandas development to find the kernels that are needed for dask.  A few lines for copy-paste that can get it up and running in a few seconds could be useful, but not a substitute for the other docs.

docs/source/develop.rst
dazza-codes: The intention is to provide a small one-stop-shop for a copy-paste snippet that works, with no doubts or questions about how to use the files.  I understand the value of DRY documentation, so it could be removed for the sake of doc-maintenance and consistency.

docs/source/develop.rst
dazza-codes: I'll split this and the other snippet out of this PR.

docs/source/develop.rst
dazza-codes: This is removed from this PR, could be taken up in #3406

dazza-codes:
The focus of this PR is now only the test code (the git/conda changes are split out).

jrbourbeau:
Also, I noticed this is your first code contribution to this repository. Welcome!

PR #3401, opened 2020-01-22
Ensure scheduler updates task and worker states after successful worker data deletion

jrbourbeau:
In `Scheduler.replicate` we currently delete data on workers and then update the corresponding `WorkerState` and `TaskState` on the scheduler:

https://github.com/dask/distributed/blob/a891a8316a2f28eeea77145a125e56570e51bc39/distributed/scheduler.py#L3144-L3166

However in the event that an error is raised in one of the awaitables being gathered, [`asyncio.gather`](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) doesn't cancel the other tasks being gathered so they continue to run. For us this means that if one of the `delete_data` tasks fails for some reason (raises an error) then other `delete_data` tasks will continue to run, but we won't get to the next step where we update the worker and task states. So the task/worker states on the scheduler can get out of sync with the actual state of our workers in this way. 

This PR consolidates the worker data deletion + worker/task update into a single coroutine which is then gathered. This way, we'll always update worker and task states on the scheduler for keys that are successfully deleted from a worker. 

cc @jcrist if you get a moment


jrbourbeau:
Marked as a WIP for now as we may be able to make a similar update to `Scheduler.rebalance`

jrbourbeau:
I was able to make a similar change to `Scheduler.rebalance`. So now both it and `Scheduler.replicate` use the new `Scheduler._delete_worker_data` method which should ensure that when a set of keys are deleted from a worker, the corresponding worker/task states on the scheduler are always updated

mrocklin:
This looks great to me.  Thank you for doing this @jrbourbeau .  I'm curious, what would happen if one of the delete calls were to fail (say a worker dies during that process) is it possible that we would have deleted some data on a worker but not cleared up its state on the scheduler?  Clearly this would have been a problem before as well, so we should merge this regardless, I'm just curious if we can get into trouble here.

jrbourbeau:
> if one of the delete calls were to fail...is it possible that we would have deleted some data on a worker but not cleared up its state on the scheduler?

That's correct, even with the changes in this PR. If a worker dies, say, halfway through deleting a set of keys, then we won't update the scheduler state for those keys. 

Previously if an error was raised in any of the data deletion tasks, then no scheduler state is updated (even for the keys that were successfully deleted from a worker). Now, we only fail to update the scheduler state for the keys where data deletion tasks also fail.

mrocklin:
Thanks @jrbourbeau .  Merging this in.

PR #3402, opened 2020-01-22
Cleanup old Tornado compatibility code

jakirkham:
Following up on the discussion in issue ( https://github.com/rapidsai/dask-cuda/issues/198 ) to speed up cluster startup time.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -330,7 +330,7 @@ class SpecCluster(Cluster):
                 self._created.add(worker)
                 workers.append(worker)
             if workers:
-                await asyncio.wait(workers)
+                 await asyncio.gather(*workers)
                 for w in workers:
                     w._cluster = weakref.ref(self)
             self.workers.update(dict(zip(to_open, workers)))
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -308,7 +308,7 @@ class SpecCluster(Cluster):
                 if self.scheduler.status == "running":
                     await self.scheduler_comm.retire_workers(workers=list(to_close))
                 tasks = [self.workers[w].close() for w in to_close if w in self.workers]
-                await asyncio.wait(tasks)
+                await asyncio.gather(*tasks)
             for name in to_close:
                 if name in self.workers:
                     del self.workers[name]
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -330,7 +330,7 @@ class SpecCluster(Cluster):
                 self._created.add(worker)
                 workers.append(worker)
             if workers:
-                 await asyncio.gather(*workers)
+                await asyncio.gather(*workers)
                 for w in workers:
                     w._cluster = weakref.ref(self)
             self.workers.update(dict(zip(to_open, workers)))
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -594,8 +594,6 @@ async def run_spec(spec: dict, *args):
 
     if workers:
         await asyncio.gather(*workers.values())
-        for w in workers.values():
-            await w  # for tornado gen.coroutine support
     return workers
 
 
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -333,7 +333,6 @@ class SpecCluster(Cluster):
                 await asyncio.wait(workers)
                 for w in workers:
                     w._cluster = weakref.ref(self)
-                await asyncio.gather(*workers)
             self.workers.update(dict(zip(to_open, workers)))
 
     def _update_worker_status(self, op, msg):
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -309,9 +309,6 @@ class SpecCluster(Cluster):
                     await self.scheduler_comm.retire_workers(workers=list(to_close))
                 tasks = [self.workers[w].close() for w in to_close if w in self.workers]
                 await asyncio.wait(tasks)
-                for task in tasks:  # for tornado gen.coroutine support
-                    with ignoring(RuntimeError):
-                        await task
             for name in to_close:
                 if name in self.workers:
                     del self.workers[name]
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -365,7 +365,7 @@ class SpecCluster(Cluster):
             await self.scheduler
             await self._correct_state()
             if self.workers:
-                await asyncio.wait(list(self.workers.values()))  # maybe there are more
+                await asyncio.gather(*self.workers.values())  # maybe there are more
             return self
 
         return _().__await__()
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -333,7 +333,7 @@ class SpecCluster(Cluster):
                 await asyncio.wait(workers)
                 for w in workers:
                     w._cluster = weakref.ref(self)
-                    await w  # for tornado gen.coroutine support
+                await asyncio.gather(*workers)
             self.workers.update(dict(zip(to_open, workers)))
 
     def _update_worker_status(self, op, msg):


distributed/deploy/spec.py
```diff
@@ -333,7 +333,7 @@ def _correct_state(self):
                 await asyncio.wait(workers)
                 for w in workers:
                     w._cluster = weakref.ref(self)
-                    await w  # for tornado gen.coroutine support
+                await asyncio.gather(*workers)
```
jakirkham: @jcrist, mentioned in [this comment]( https://github.com/rapidsai/dask-cuda/issues/198#issuecomment-561393289 ) that we should use `return_exceptions=True`. Though it sounds like `asyncio.gather` will just raise exceptions if we don't pass that argument, which seems similar to the previous behavior (excepting order is not guaranteed). So IDK if this is good enough or if we still need to handle exceptions for some other reason. If so, please advise on what sort of handling is expected. Thanks!

jcrist: Actually, I missed the `asyncio.wait` call above here, which should already successfully wait on all workers. Can we just delete the `await w` call entirely and not replace it with `asyncio.gather`? I don't see why we'd need to individually wait each worker in the case of tornado `gen.coroutine`, the `asyncio.wait` call should handle those equally fine.

jakirkham: Sure happy to give that a try. The comment worried me, but I try your judgement here more than my own 🙂

jakirkham: @mrocklin, do you know why the `await` was needed here? I see it was added in PR ( https://github.com/dask/distributed/pull/2675 ). Maybe [this discussion]( https://github.com/dask/distributed/pull/2675#discussion_r286161172 ) is relevant?

quasiben:
It might be worth benchmarking to see if this improves launch time.  I think I tried this but it did not change anything.  Happy to be wrong about this

jakirkham:
Yeah I was planning to ask someone to try it out once we settled on the changeset 🙂

mrocklin:
This definitely came out of the great asyncification of dask.distributed.  I think that maybe we didn't raise exceptions correctly or something?  I don't recall unfortunately.

jakirkham:
No worries. Figured I'd check before removing it 😉

jakirkham:
Seeing the following test failures:

```python
2581__________________________ test_death_timeout_raises ___________________________
2582
2583loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7f295186b2d0>
2584
2585    def test_death_timeout_raises(loop):
2586        with pytest.raises(TimeoutError):
2587            with LocalCluster(
2588                scheduler_port=0,
2589                silence_logs=False,
2590                death_timeout=1e-10,
2591                dashboard_address=None,
2592                loop=loop,
2593            ) as cluster:
2594>               pass
2595E               Failed: DID NOT RAISE <class 'concurrent.futures._base.TimeoutError'>
2596
2597distributed/deploy/tests/test_local.py:534: Failed
...
2652______________________________ test_broken_worker ______________________________
2653
2654    @pytest.mark.asyncio
2655    async def test_broken_worker():
2656        with pytest.raises(Exception) as info:
2657            async with SpecCluster(
2658                asynchronous=True,
2659                workers={"good": {"cls": Worker}, "bad": {"cls": BrokenWorker}},
2660                scheduler={"cls": Scheduler, "options": {"port": 0}},
2661            ) as cluster:
2662>               pass
2663E               Failed: DID NOT RAISE <class 'Exception'>
2664
2665distributed/deploy/tests/test_spec_cluster.py:223: Failed
```

ref: https://travis-ci.org/dask/distributed/jobs/640655412

distributed/deploy/spec.py
```diff
@@ -333,7 +333,6 @@ def _correct_state(self):
                 await asyncio.wait(workers)
```
jcrist: I think you want to swap this for `gather` to ensure that all exceptions are retrieved (stops warning) and raised (should fix tests).

```suggestion
                 await asyncio.gather(*workers)
```

jakirkham: Good point. Thanks Jim! Let's give it a try 🙂

jakirkham:
Looks like that worked. Do you want to give it another look @jcrist?

distributed/deploy/spec.py
jakirkham: FWIW I dug a little further and it looks `asyncio.gather` may not have always worked with Tornado, but they did fix it and start testing with `asyncio.gather` in 5.0.0 (see [this line]( https://github.com/tornadoweb/tornado/pull/2253/files#diff-fe992c84ff138a2e164479c6b81d4668R1074 )). So this should be fine.

jakirkham:
Seeing the following error on CI:

<details>

```python
2580________________ ERROR at teardown of test_death_timeout_raises ________________
2581
2582    @pytest.fixture
2583    def loop():
2584        with check_instances():
2585            with pristine_loop() as loop:
2586                # Monkey-patch IOLoop.start to wait for loop stop
2587                orig_start = loop.start
2588                is_stopped = threading.Event()
2589                is_stopped.set()
2590    
2591                def start():
2592                    is_stopped.clear()
2593                    try:
2594                        orig_start()
2595                    finally:
2596                        is_stopped.set()
2597    
2598                loop.start = start
2599    
2600                yield loop
2601    
2602                # Stop the loop in case it's still running
2603                try:
2604                    sync(loop, cleanup_global_workers, callback_timeout=0.500)
2605                    loop.add_callback(loop.stop)
2606                except RuntimeError as e:
2607                    if not re.match("IOLoop is clos(ed|ing)", str(e)):
2608                        raise
2609                except TimeoutError:
2610                    pass
2611                else:
2612>                   is_stopped.wait()
2613
2614distributed/utils_test.py:147: 
2615_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2616../../../miniconda/envs/test-environment/lib/python3.7/contextlib.py:119: in __exit__
2617    next(self.gen)
2618_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2619
2620    @contextmanager
2621    def check_instances():
2622        Client._instances.clear()
2623        Worker._instances.clear()
2624        Scheduler._instances.clear()
2625        SpecCluster._instances.clear()
2626        # assert all(n.status == "closed" for n in Nanny._instances), {
2627        #     n: n.status for n in Nanny._instances
2628        # }
2629        Nanny._instances.clear()
2630        _global_clients.clear()
2631        Comm._instances.clear()
2632    
2633        yield
2634    
2635        start = time()
2636        while set(_global_clients):
2637            sleep(0.1)
2638            assert time() < start + 10
2639    
2640        _global_clients.clear()
2641    
2642        for w in Worker._instances:
2643            with ignoring(RuntimeError):  # closed IOLoop
2644                w.loop.add_callback(w.close, report=False, executor_wait=False)
2645                if w.status == "running":
2646                    w.loop.add_callback(w.close)
2647        Worker._instances.clear()
2648    
2649        for i in range(5):
2650            if all(c.closed() for c in Comm._instances):
2651                break
2652            else:
2653                sleep(0.1)
2654        else:
2655            L = [c for c in Comm._instances if not c.closed()]
2656            Comm._instances.clear()
2657            print("Unclosed Comms", L)
2658            # raise ValueError("Unclosed Comms", L)
2659    
2660>       assert all(n.status == "closed" or n.status == "init" for n in Nanny._instances), {
2661            n: n.status for n in Nanny._instances
2662        }
2663E       AssertionError: {<Nanny: None, threads: 1>: 'closing', <Nanny: None, threads: 1>: 'closed'}
2664E       assert False
2665E        +  where False = all(<generator object check_instances.<locals>.<genexpr> at 0x7f2bf0ee5ad0>)
2666
2667distributed/utils_test.py:1520: AssertionError
```

</details>

ref: https://travis-ci.org/dask/distributed/jobs/641102071#L2580

jakirkham:
@jcrist, if you have a moment to advise, that would be really helpful 🙂

mrocklin:
> Actually, I missed the asyncio.wait call above here, which should already successfully wait on all workers. Can we just delete the await w call entirely and not replace it with asyncio.gather? I don't see why we'd need to individually wait each worker in the case of tornado gen.coroutine, the asyncio.wait call should handle those equally fine.


I suspect that this PR doesn't significantly affect startup time.  (do we have any performance information about this?).  Given that, does it still make sense to do?  Should we close it out?

jakirkham:
Right now it fails tests, which means it doesn’t work. I don’t think it makes sense to benchmark things that don’t work.

jakirkham:
Also would still like guidance on why the tests are failing. Is this something you would be willing to do @mrocklin? 🙂

mrocklin:
> Also would still like guidance on why the tests are failing. Is this something you would be willing to do @mrocklin? slightly_smiling_face

Sure, I'll take a look.

```python-traceback
>       assert all(n.status == "closed" or n.status == "init" for n in Nanny._instances), {
            n: n.status for n in Nanny._instances
        }
E       AssertionError: {<Nanny: None, threads: 1>: 'closing', <Nanny: None, threads: 1>: 'closed'}
E       assert False
E        +  where False = all(<generator object check_instances.<locals>.<genexpr> at 0x7f2bf0ee5ad0>)
```

This looks like one of the close calls wasn't fully awaited.  Probably one of the deleted awaits that you've removed is actually necessary?  Beyond that I wouldn't know without investing a lot more time.

> Right now it fails tests, which means it doesn’t work. I don’t think it makes sense to benchmark things that don’t work.

Sure.  Mostly I'm responding to Jim's comment here:

>  Actually, I missed the asyncio.wait call above here, which should already successfully wait on all workers. Can we just delete the await w call entirely and not replace it with asyncio.gather? I don't see why we'd need to individually wait each worker in the case of tornado gen.coroutine, the asyncio.wait call should handle those equally fine.

By this I think that the original premise of the original issue was flawed.  All of the workers are started simultaneously.  It might be worth verifying that the original assumption, that we were starting workers one at a time, is actually true before investing more time here.

jakirkham:
Yeah I think closing is the right thing to do here. What we have found is, in our use case we call `cuInit` (indirectly through `numba`), which takes ~1s ( https://github.com/rapidsai/dask-cuda/issues/198#issuecomment-607535763 ). However if multiple workers call `cuInit`, this scales quadratically (not linearly). Given how long `cuInit` takes and that it scales quadratically, I doubt changes in Dask (particularly ones that scale linearly) will have much effect on the startup time.

PR #3404, opened 2020-01-24
Drop custom cuDF serialization

jakirkham:
This is since handled in cuDF 0.9.0, which was released a while ago. So go ahead and drop this from Distributed.


jakirkham:
Also cc @madsbk (as GitHub won't let me request a review from you 😛)

distributed/protocol/__init__.py
```diff
@@ -78,14 +78,3 @@ def _register_cupy():
...
     from . import numba
-
-
-@cuda_serialize.register_lazy("cudf")
-@cuda_deserialize.register_lazy("cudf")
-def _register_cudf():
-    import cudf
-
-    if LooseVersion(cudf.__version__) > "0.9":
-        from cudf.comm import serialize
```
rjzamora: Don't we still want to import the serialization definitions in cudf?

quasiben: yes, i think we still need these lines

jakirkham: Ah ok. Thought we might be doing this in cuDF too. Have added it back 🙂

jakirkham: Actually is this not handled [here]( https://github.com/rapidsai/cudf/blob/1ead2d55562b9cd1b80568b86b8b51ab616f3cb8/python/cudf/cudf/comm/serialize.py#L10-L40 )?

rjzamora: > Actually is this not handled here?

Right, but the `from cudf.comm import serialize` statement might be needed here to make sure the registration is triggered correctly.

quasiben:
Thanks @jakirkham and @rjzamora for the reviews.  Unless there are other comments I am going to merge later today

PR #3406, opened 2020-01-24
Docs - add example conda-dev-env

dazza-codes:
Split out of #3398 to be rebased later and adapted to provide a small snippet for dev-getting-started.

The intention is to provide a small one-stop-shop for a copy-paste snippet that works, with no doubts or questions about how to use the files involved.  There is great value in DRY consistent documentation, so it could be removed for the sake of doc-maintenance and consistency.  But, it currently places a burden on the reader to jump around to a few places to sift through the volumes to find a few kernels about how to get off the ground.

### DIFF
--- a/docs/source/develop.rst
+++ b/docs/source/develop.rst
@@ -22,6 +22,26 @@ and install it from source::
    cd distributed
    python setup.py install
 
+Using conda, for example::
+
+   git clone git@github.com:{your-fork}/distributed.git
+   cd distributed
+   conda create -y -n distributed python=3.6
+   conda activate distributed
+   python -m pip install -U -r requirements.txt
+   python -m pip install -U -r dev-requirements.txt
+   python -m pip install -e .
+
+To keep a fork in sync with the upstream source::
+
+   cd distributed
+   git remote add upstream git@github.com:dask/distributed.git
+   git remote -v
+   git fetch -a upstream
+   git checkout master
+   git pull upstream master
+   git push origin master
+
 Test
 ----
 


docs/source/develop.rst
```diff
@@ -22,6 +22,26 @@ and install it from source::
...
 
+Using conda, for example::
+
+   git clone git@github.com:{your-fork}/distributed.git
+   cd distributed
+   conda create -y -n distributed python=3.6
+   conda activate distributed
+   pip install -U -r requirements.txt
+   pip install -U -r dev-requirements.txt
+   pip install -e .
```
quasiben: I think the recommended way of using pip now is with `python -m pip ...`. but I can't find where are read that.

TomAugspurger: https://snarky.ca/why-you-should-use-python-m-pip/

dazza-codes: Done

docs/source/develop.rst
```diff
@@ -22,6 +22,26 @@ and install it from source::
...
+   python -m pip install -U -r dev-requirements.txt
+   python -m pip install -e .
+
+To keep a fork in sync with the upstream source::
+
+   cd distributed
+   git checkout master
+   git remote add upstream git@github.com:dask/distributed.git
+   git remote -v
+   git fetch -ap upstream
```
quasiben: One last comment here, the `-p` may unintentionally remove tags the user would like to keep.  I don't think this is the common case, but if we are going to be recommending git commands perhaps we should avoid prunng?

dazza-codes: OK, better safe than sorry.  Done.

quasiben:
Thanks @dazza-codes for the contribution -- merging in!

Issue #3399, opened 2020-01-22
ZeroDivisionError in scheduler dashboard


mdwint:
This occurs when visiting the `/status` page when no workers are running.

```python
Traceback (most recent call last):
  File "/code/.venv/lib/python3.7/site-packages/distributed/utils.py", line 652, in log_errors
    yield
  File "/code/.venv/lib/python3.7/site-packages/distributed/dashboard/components/scheduler.py", line 1754, in status_doc
    task_progress.update()
  File "/code/.venv/lib/python3.7/site-packages/bokeh/core/property/validation.py", line 97, in func
    return input_function(*args, **kwargs)
  File "/code/.venv/lib/python3.7/site-packages/distributed/dashboard/components/scheduler.py", line 1435, in update
    d = progress_quads(state)
  File "/code/.venv/lib/python3.7/site-packages/distributed/diagnostics/progress_stream.py", line 105, in progress_quads
    rl = width * r / a + l
ZeroDivisionError: float division by zero
```

(distributed: 2.9.1)

jrbourbeau:
Thanks for reporting @mdwint! I'm not able to reproduce the `ZeroDivisionError` on the current `master` branch or 2.9.1 release. I'm running `dask-scheduler` in my local terminal and then going to http://localhost:8787/status in my browser with no errors subsequently being raised in the terminal. Does that differ from what you're doing?

mdwint:
Hi @jrbourbeau. We're running Dask in Docker on AWS, but I'm not sure if that would have an impact. I'll try to create minimal example locally.

jrbourbeau:
Awesome, thanks @mdwint! That goes a long in debugging and fixing things

mdwint:
This example produces the error:

```python
>>> from distributed import Client, LocalCluster
>>> cluster = LocalCluster(n_workers=0)
>>> client = Client(cluster)
>>> client.submit(lambda: 1)
```

With this session still running, visit http://localhost:8787/status.

The key here is that a client submits work while no workers are available. I suspect that the scheduler's state then holds tasks with some fields set to zero, causing the ZeroDivisionError.

jrbourbeau:
Thanks for the nice reproducible example @mdwint! https://github.com/dask/distributed/pull/3407 should fix the `ZeroDivisionError`

PR #3407, opened 2020-01-24
Fix TaskProgress dashboard component when tasks are only in the no-worker state

jrbourbeau:
Currently the `TaskProgress` dashboard component uses `TaskPrefix.active_states` to get the number of tasks in each state. However, `TaskPrefix.active_states` includes tasks in the `no-worker` state which `TaskProgress` doesn't track. This causes issues when we only have tasks in the `no-worker` state.

This PR updates `TaskProgress` to only rely on the states it's tracking, not all the states in `TaskPrefix.active_states` (i.e. ignore the `no-worker` state in `active_states`)

cc @jacobtomlinson 

Closes https://github.com/dask/distributed/issues/3399

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1418,12 +1418,13 @@ class TaskProgress(DashboardComponent):
             }
 
             for tp in self.scheduler.task_prefixes.values():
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Thanks for reviewing @jacobtomlinson!

Issue #3409, opened 2020-01-24
WorkStealing in DEFAULT_EXTENSIONS when work stealing disabled


jakirkham:
It appears that if work stealing is not disabled before importing `dask.distributed` the `config` value is ignored. For example please take the following examples:

Work stealing still enabled:

```python
import dask
import dask.distributed
dask.config.set({'distributed.scheduler.work-stealing': False})
```

Work stealing disabled:

```python
import dask
dask.config.set({'distributed.scheduler.work-stealing': False})
import dask.distributed
```

This seems to come down to the fact that [`dask.distributed.scheduler.DEFAULT_EXTENSIONS` will have `distributed.stealing.WorkStealing` added in the former case]( https://github.com/dask/distributed/blob/3a2f8a84121d27decf432cd4f379724907ea2b3d/distributed/scheduler.py#L92-L93 ) and not the latter. It does not appear that `dask.distributed.scheduler.DEFAULT_EXTENSIONS` is updated later either.

<hr>

As a workaround, users may consider including the following code before starting up any Dask clusters.

```python
from contextlib import suppress
import dask
import dask.distributed

if not dask.config.get("distributed.scheduler.work-stealing"):
    with suppress(ValueError):
        dask.distributed.scheduler.DEFAULT_EXTENSIONS.remove(
            dask.distributed.stealing.WorkStealing
        )
```

jrbourbeau:
We could move the config check and corresponding `WorkStealing` append inside `Scheduler.__init__` here:

https://github.com/dask/distributed/blob/3a2f8a84121d27decf432cd4f379724907ea2b3d/distributed/scheduler.py#L1335-L1336

I think that's the only place where `DEFAULT_EXTENSIONS` for the `Scheduler` class is used

jrbourbeau:
So something like:

```python
if extensions is None:
    extensions = DEFAULT_EXTENSIONS
    if dask.config.get("distributed.scheduler.work-stealing"):
        extensions.append(WorkStealing)
```

What do you think?

jakirkham:
Seems reasonable to me.

jakirkham:
Submitted PR ( https://github.com/dask/distributed/pull/3410 ) with the suggested fix 🙂

PR #3410, opened 2020-01-24
Enable WorkStealing case-by-case

jakirkham:
Fixes https://github.com/dask/distributed/issues/3409

Previously the decision to enable work-stealing was made when `distributed` was imported and could not be changed after that. This delays the decision until a scheduler is spun up and then makes sure to check `dask.config`.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -89,9 +89,6 @@ DEFAULT_EXTENSIONS = [
     PubSubSchedulerExtension,
 ]
 
-if dask.config.get("distributed.scheduler.work-stealing"):
-    DEFAULT_EXTENSIONS.append(WorkStealing)
-
 ALL_TASK_STATES = {"released", "waiting", "no-worker", "processing", "erred", "memory"}
 
 
@@ -1333,7 +1330,9 @@ class Scheduler(ServerNode):
             self.periodic_callbacks["idle-timeout"] = pc
 
         if extensions is None:
-            extensions = DEFAULT_EXTENSIONS
+            extensions = list(DEFAULT_EXTENSIONS)
+            if dask.config.get("distributed.scheduler.work-stealing"):
+                extensions.append(WorkStealing)
         for ext in extensions:
             ext(self)
 
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -596,17 +596,17 @@ def test_coerce_address():
         yield [w.close() for w in [a, b, c]]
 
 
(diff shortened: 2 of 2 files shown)


distributed/scheduler.py
```diff
@@ -1333,7 +1330,9 @@ def __init__(
             self.periodic_callbacks["idle-timeout"] = pc
 
         if extensions is None:
-            extensions = DEFAULT_EXTENSIONS
+            extensions = list(DEFAULT_EXTENSIONS)
```
jakirkham: Copied to avoid modifying the `DEFAULT_EXTENSIONS` by accident here.

jrbourbeau: Ah yes, good catch

distributed/tests/test_scheduler.py
```diff
@@ -596,6 +596,19 @@ def test_coerce_address():
         yield [w.close() for w in [a, b, c]]
 
 
+@gen_test()
+def test_config_stealing():
```
jakirkham: Is it preferred to use `@gen_test()` or `@pytest.mark.asyncio`? There are a few of both here

mrocklin:
As a warning, almost every case of people disabling workstealing that I've seen has been because of some other bug, like mis-estimating memory sizes.  If you're in a situation where you're finding that disabling work stealing is helping then I recommend checking out how `sizeof` reports the size of your objects, and what your bandwidth tends to be when moving those objects.

jakirkham:
At this stage we have done a pretty good job of covering `sizeof` in many cases. It's always possible we are missing the size of some object. Though I suspect it is a different issue. Unfortunately we are unlikely to explore this until the medium term, but I agree it's worth understanding better.

In any event it seems reasonable that this configure flag works as expected. Not too mention it seems to make users happy 😉

distributed/tests/test_scheduler.py
jrbourbeau: I'm not sure if there's a reason to prefer one over the other (although others may have an opinion here). FWIW there are ~3x more occurrence of `pytest.mark.asyncio` than `gen_test`, so maybe `pytest.mark.asyncio` is preferred

mrocklin:
Yeah, I have no objection to the actual change, just a general comment/warning a bit.

distributed/tests/test_scheduler.py
jakirkham: Makes sense. Rewrote as suggested using `pytest.mark.asyncio`. PTAL 🙂

PR #3418, opened 2020-01-28
make _get_ip return an IP address when defaulting

### LINKED ISSUE #3417: test_local_tls fails on master (MacOS High Sierra, Ubuntu)
Reproducer (on a fresh `Python3.7` virtual environment, assuming you at the `distributed` root)

```
pip install -e .
python -m pytest distributed/deploy -k test_local_tls
```

By bisecting the repository it seems that the faulty commit is 3151f09e.
I am not an expert, but it seems that there relying only on `hostname` is not enough -- itself probably being linked tot the fact that we don't pass the IP version (v4 or v6) anymore. As proposed by the already existent [inline comment](https://github.com/dask/distributed/blob/4081a0e09febd9099f04a1db9772f6cd95203b2e/distributed/utils.py#L141), we should probably use `socket.getaddrinfo` or `socket.getfqdn()` on top of the `hostname`.

This currently affects the downstream builds of `cloudpickle`, (see cloudpipe/cloudpickle#324), so pinging @jakirkham :)

pierreglaser:
Fixes #3417 

Changing `socket.gethostname()` by `socket.getaddrinfo(socket.gethostname(), port, family)[0][4][0]` also works, but I don't know whether it is better than `socket.getfqdn(socket.gethostname())`. Maybe @pitrou has any clue, as the author of the useful `XXX` comment above :)

also pinging @jakirkham and @ogrisel

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -143,7 +143,14 @@ def _get_ip(host, port, family):
             "reaching %r, defaulting to hostname: %s" % (host, e),
             RuntimeWarning,
         )
-        return socket.getfqdn(socket.gethostname())
+        addr_info = socket.getaddrinfo(
+            socket.gethostname(),
+            port,
+            family,
+            socket.SOCK_DGRAM,
+            socket.IPPROTO_UDP
+        )[0]
+        return addr_info[4][0]
     finally:
         sock.close()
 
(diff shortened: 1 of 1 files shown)


pitrou:
`getaddrinfo` should definitely be used. You want an IP address, not a hostname.

mrocklin:
cc @jacobtomlinson

mrocklin:
Thanks @pierreglaser .  Merging this in.  

Thanks also @pitrou @jrbourbeau @jacobtomlinson for reviewing.

PR #3423, opened 2020-01-29
Don't send empty dependencies

Kobzol:
When a client sends the `update-graph` message, it sends a dictionary with dependencies of individual tasks. Tasks without dependencies needlessly occupy bandwidth in this message, since the scheduler anyway does something like `dependencies.get(task, ())` when it parses the message (if I'm wrong please let me know).

For example for this simple merge with a lot of tasks without dependencies:
```python
from dask import delayed

client = Client("tcp://localhost:8786")

@delayed
def do_something(x):
    return x * 10

@delayed
def merge(*args):
    return sum(args)

xs = [do_something(x) for x in range(1000)]
result = merge(*xs).compute()
```

This PR changes the dependencies in the `update-graph` message from this:
```python
{
    "merge-3cfcfa4f-fa40-45ba-b181-54eb7554377f": ('do_something-ae852807-355c-4d87-85e4-72e6f1d0a146', 'do_something-3e003b98-363b-42ab-b270-691db098f222', ...),
    "do_something-9f9820f5-fbe9-4d31-ba9d-ac17996e4c35": (),
    "do_something-34c2c63b-2d3d-4e5c-8fef-f94ab635e19f": (),
    ... # 998 additional lines
}
```

into this:
```python
{
    "merge-3cfcfa4f-fa40-45ba-b181-54eb7554377f": ('do_something-ae852807-355c-4d87-85e4-72e6f1d0a146', 'do_something-3e003b98-363b-42ab-b270-691db098f222', ...),
}
```

For the aforementioned graph, it reduces the `update-graph` total frame byte size by the following amounts:

| Merge task # | Original size | Reduced size | % of original |
| ----------------- | ----------------- | ------------------ | ----------------- | 
| 5000 | 4170630 B | 3910628 B | 94 % |
| 10000 | 8340666 B | 7820664 B | 94 % |
| 15000 | 12510702 B | 11730700 B | 94 % |
| 20000 | 16680738 B | 15640736 B | 94 % |
| 25000 | 20850774 B | 19550772 B | 94 % |

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2489,6 +2489,7 @@ class Client(Node):
             dependencies = {
                 tokey(k): [tokey(dep) for dep in deps]
                 for k, deps in dependencies.items()
+                if deps
             }
             for k, deps in future_dependencies.items():
                 if deps:
--- a/distributed/diagnostics/graph_layout.py
+++ b/distributed/diagnostics/graph_layout.py
@@ -36,16 +36,16 @@ class GraphLayout(SchedulerPlugin):
             }
             priority = {k: ts.priority for k, ts in scheduler.tasks.items()}
             self.update_graph(
-                self.scheduler, dependencies=dependencies, priority=priority
+                self.scheduler, tasks=self.scheduler.tasks, dependencies=dependencies, priority=priority
             )
 
-    def update_graph(self, scheduler, dependencies=None, priority=None, **kwargs):
-        stack = sorted(dependencies, key=lambda k: priority.get(k, 0), reverse=True)
+    def update_graph(self, scheduler, dependencies=None, priority=None, tasks=None, **kwargs):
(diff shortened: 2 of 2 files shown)


mrocklin:
Thank you for the pull request @Kobzol .  I'm curious, was this work motivated by something in particular? The performance changes here are small, so I'm trying to decide if adding this change is worth the added complexity.  Was there a motivating application?

Kobzol:
While working on the Rust scheduler, I noticed that there is a lot of empty dependencies being sent, which seemed inefficient to me. I wouldn't send this PR if it was a complicated change, but since it's a one liner, I thought that it's a nice improvement for such a small change.

I later had to modify the Graphlayout to work with tasks instead of dependencies, but even that doesn't seem like a big change.

mrocklin:
Thanks @Kobzol !  Sorry for the delay

PR #3425, opened 2020-01-29
Try to use pickle5 when available

jakirkham:
As there are some improvements for buffer serialization in pickle protocol 5 and these can be opted in using a backports package on older Python versions. Try to leverage the backport package when available and fallback to the builtin pickle otherwise.

cc @quasiben @pentschev

### DIFF
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -3,9 +3,7 @@ import sys
 
 import cloudpickle
 
-if sys.version_info.major == 2:
-    import cPickle as pickle
-elif sys.version_info.major == 3 and sys.version_info.minor < 8:
+if sys.version_info.major == 3 and sys.version_info.minor < 8:
     try:
         import pickle5 as pickle
     except ImportError:
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -5,7 +5,12 @@ import cloudpickle
 
 if sys.version_info.major == 2:
     import cPickle as pickle
-else:
+elif sys.version_info.major == 3 and sys.version_info.minor < 8:
+    try:
+        import pickle5 as pickle
+    except ImportError:
+        import pickle
+elif sys.version_info.minor:
     import pickle
 
 logger = logging.getLogger(__name__)
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -10,7 +10,7 @@ elif sys.version_info.major == 3 and sys.version_info.minor < 8:
         import pickle5 as pickle
     except ImportError:
         import pickle
-elif sys.version_info.minor:
+else:
     import pickle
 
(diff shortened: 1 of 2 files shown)


mrocklin:
When using protocol 5 I would expect us to select out the frames from the pickling process and send them along separately.  Does using pickle5 provide a speed boost if we just use dumps/loads naively?

cc @pitrou @ogrisel

mrocklin:
Closing this soonish if there are no further comments.

pitrou:
> When using protocol 5 I would expect us to select out the frames from the pickling process and send them along separately. 

That would be the best reason to use it indeed.

>  Does using pickle5 provide a speed boost if we just use dumps/loads naively?

It may, depending exactly how pickle is invoked (some copies may be avoided, though not all). The best is if you can do some measurements and post them here :-)

distributed/protocol/pickle.py
```diff
@@ -5,7 +5,12 @@
 
 if sys.version_info.major == 2:
     import cPickle as pickle
-else:
+elif sys.version_info.major == 3 and sys.version_info.minor < 8:
+    try:
+        import pickle5 as pickle
+    except ImportError:
+        import pickle
+elif sys.version_info.minor:
```
pitrou: Why this condition?

pitrou:
PS: are you still Python 2-compatible?

mrocklin:
No, we dropped Python 2 a little while ago.

distributed/protocol/pickle.py
```diff
@@ -5,7 +5,12 @@
 
 if sys.version_info.major == 2:
```
pitrou: I suppose this can be removed :-)

distributed/protocol/pickle.py
jakirkham: I think this was bad copy-pasta. Probably should be `else`.

distributed/protocol/pickle.py
jakirkham: Thanks. Dropped :)

jakirkham:
> Does using pickle5 provide a speed boost if we just use dumps/loads naively?

This is what I was finding.

> When using protocol 5 I would expect us to select out the frames from the pickling process and send them along separately.

That would certainly be better. Though it would be nice if we didn't have to maintain non-protocol 5 support as well.

jakirkham:
Replacing with PR ( https://github.com/dask/distributed/pull/3784 ), which actually handles out-of-band buffers. Admittedly it doesn't use pickle5 on older Python versions. However we would need cloudpickle to support to support pickle5 for that to work correctly ( https://github.com/cloudpipe/cloudpickle/issues/179 ) anyways.

jakirkham:
Also this has been extended to Python versions pre-3.8 with `pickle5` with PR ( https://github.com/dask/distributed/pull/3849 ).

PR #3429, opened 2020-01-30
Add dashboard_link property to Client

jacobtomlinson:
Currently if you create a Dask cluster using a cluster manager object you can grab the dashboard web address using the `cluster.dashboard_link` attribute. When a client connects to that cluster you can also grab the address via `client.cluster.dashboard_link`.

We use this in the Dask Jupyter Lab Extension to auto-populate the address by grabbing the default client and then accessing the `client.cluster.dashboard_link` property. However if the cluster is created by another method, on the command line for example, the `client.cluster` property will be set to `None` and it is not possible to get the address.

However when displaying the HTML repr for the client we infer the dashboard address and expose it, this works regardless of the cluster setup method.

This PR pulls that logic out of the HTML repr into a new property so that this address can be accessed at `client.dashboard_link`. I had to add a little indirection with the `_get_scheduler_info()`  method as we now need to auto-detect the scheduler object and scheduler information in multiple places.

My plan if this is acceptable is to [update the lab extension here](https://github.com/dask/dask-labextension/blob/3915f6a4f354c917217ed3e4d830d7f345fad9b5/src/index.ts#L476) to get the address via `client.dashboard_link` instead of `client.cluster.dashboard_link`.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -752,6 +752,22 @@ class Client(Node):
         """
         return self._asynchronous and self.loop is IOLoop.current()
 
+    @property
+    def dashboard_link(self):
+        scheduler, info = self._get_scheduler_info()
+        try:
+            return self.cluster.dashboard_link
+        except AttributeError:
+            protocol, rest = scheduler.address.split("://")
+
+            port = info["services"]["dashboard"]
+            if protocol == "inproc":
+                host = "localhost"
+            else:
+                host = rest.split(":")[0]
+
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks @jacobtomlinson .  This seems fine to me.  Can I ask for a trivial test for the `dashboard_link` attribute?

jacobtomlinson:
@mrocklin done!

PR #3432, opened 2020-01-31
Remove object from class hierarchy

### LINKED ISSUE #3428: Remove object from class hierarchy
In many cases in the codebase we inherit from `object` like this:

```python
class Foo(object):
    ...
```

This used to be important in Python 2, but we no longer support that.  It might make sense to remove these for now and change code like this to the following:

```python
class Foo:
    ...
```

andersy005:
Closes #3428


mrocklin:
Woot!

Note to self: make more easy-to-fix issues :)

PR #3434, opened 2020-02-01
Add documentation of parameters in coordination primitives

### LINKED ISSUE #3343: Unclear what client is used for coordination primitives
I'm currently working with the coordination primitives in `dask.distributed`. First, let us note that the following code blows up:

```
from dask.distributed import Queue
q = Queue()
...
AttributeError: 'NoneType' object has no attribute 'asynchronous'
```
This makes sense, since clearly we need something that can talk to the scheduler. The docs provide the following example:
```
from dask.distributed import Client, Queue
c = Client()
q = Queue()
```
With this, `q.client` is `c` which also makes sense, because what else could it be. Now, consider the following example:
```
from dask.distributed import Client, Queue
c1 = Client()
c2 = Client()
q = Queue()
```
Here, there are two candidates for what `q.client` could be. It ends up being `c2` but from the documentation, it's not clear what sort of logic is applied: Does Dask keep track of all `Client`s created in the current process, then picks the last one? In any case, it would be nice with just a couple words explaining this in the docs.

fuglede:
This closes #3343. A few questions:

* I left the `name` parameter as a `string`, because that's what was given in the docstring for `Lock`. However, `Pub`/`Sub` just specify that it should be a general msgpack-serializable object; is that supported for all of them? Then I'll go ahead and change it.
* It looks like the `maxsize` parameter in `Variable` is unused (maybe a result of copying from `Queue`?) Should that be removed? It breaks API backwards compatibility but given that it didn't do anything, chances are this is unproblematic?
* In general, there's some inconsistency in whether a period is followed by a single space or two spaces; should this be consistent?

### DIFF
--- a/distributed/queues.py
+++ b/distributed/queues.py
@@ -155,6 +155,18 @@ class Queue:
 
        This object is experimental and has known issues in Python 2
 
+    Parameters
+    ----------
+    name: string (optional)
+        Name used by other clients and the scheduler to identify the queue. If
+        not given, a random name will be generated.
+    client: Client (optional)
+        Client used for communication with the scheduler. Defaults to the
+        value of ``_get_global_client()``.
+    maxsize: int (optional)
+        Number of items allowed in the queue. If 0 (the default), the queue
+        size is unbounded.
+
     Examples
     --------
     >>> from dask.distributed import Client, Queue  # doctest: +SKIP
--- a/distributed/variable.py
+++ b/distributed/variable.py
@@ -130,6 +130,15 @@ class Variable:
 
        This object is experimental and has known issues in Python 2
 
+    Parameters
+    ----------
+    name: string (optional)
+        Name used by other clients and the scheduler to identify the variable.
+        If not given, a random name will be generated.
+    client: Client (optional)
+        Client used for communication with the scheduler. Defaults to the
+        value of ``_get_global_client()``.
+
     Examples
     --------
     >>> from dask.distributed import Client, Variable # doctest: +SKIP
--- a/distributed/lock.py
+++ b/distributed/lock.py
@@ -78,9 +78,13 @@ class Lock:
 
     Parameters
     ----------
-    name: string
+    name: string (optional)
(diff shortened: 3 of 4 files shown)


distributed/queues.py
```diff
@@ -155,6 +155,18 @@ class Queue:
...
        This object is experimental and has known issues in Python 2
 
+    Parameters
+    ----------
+    name: string (optional)
+        Name used by other clients and the scheduler to identify the queue. If
+        not given, a random name will be generated.
+    client: Client (optional)
+        Client used for communication with the scheduler. Defaults to the
+        value of ``_get_global_client()``.
```
jakirkham: Maybe instead of specifying a private function here we can just state this in words? What do you think?

distributed/variable.py
```diff
@@ -130,6 +130,15 @@ class Variable:
...
        This object is experimental and has known issues in Python 2
 
+    Parameters
+    ----------
+    name: string (optional)
+        Name used by other clients and the scheduler to identify the variable.
+        If not given, a random name will be generated.
+    client: Client (optional)
+        Client used for communication with the scheduler. Defaults to the
+        value of ``_get_global_client()``.
```
jakirkham: We can apply the same change here as well. Though let's keep the discussion in the first thread for simplicity 🙂

distributed/queues.py
fuglede: As a general principle, I agree; documentation is part of the public API and should be agnostic towards private members. I put it this way to make it consistent with the documentation of [`Future`](https://distributed.readthedocs.io/en/latest/api.html#future), so if we change it here, maybe we should take care of that one while we're at it?

I considered just replacing it with an explanation (something like "the default global client"?) but then again, @mrocklin pointing out to me the existence of the relevant function is what helped me realize what was actually happening.

Out of these, `Lock` is a bit special as it uses `self.client = client or _get_global_client() or get_worker().client`.

jakirkham: Fair enough.

jakirkham:
Thanks @fuglede! 😄

PR #3441, opened 2020-02-04
Changing default values for dask_worker cli options

### LINKED ISSUE #3440: change default value of deploy.cli.dask-worker to use .config/dask if not given
In case of using `dask-worker` from command line, the default value of local-directory is an empty string. If you want to define a default directory in `.config/dask.dask.yaml` by setting the `temporary-directory` value.
In `distributed/worker.py Worker` the configured value from config file will only be used if `local-directory is None` which is not an empty string.

The same is with some other options which should be `None` instead of `""` for consistency.

condoratberlin:
fixes #3440 This fixes the use of configured local-directory in .config/dask/dask.yaml when using dask_worker cli command without giving --local-directory option.

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -151,12 +151,12 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 )
 @click.option("--pid-file", type=str, default="", help="File to write the process PID")
 @click.option(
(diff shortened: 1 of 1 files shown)


condoratberlin:
I will try.

mrocklin:
Thanks @condoratberlin

PR #3442, opened 2020-02-05
Support serializing/deserializing `rmm.DeviceBuffer`s

jakirkham:
This adds support for serializing and deserializing `rmm.DeviceBuffer` within `distributed`. As it is a very simple object supporting `__cuda_array_interface__`, this is pretty straightforward. For the most part deserialization should be trivial if RMM is available since we prefer using that when transmitting data. Otherwise construct a new `DeviceBuffer` using a pointer to the data in the frame and its size.

### DIFF
--- a/distributed/protocol/rmm.py
+++ b/distributed/protocol/rmm.py
@@ -15,10 +15,9 @@ if hasattr(rmm, "DeviceBuffer"):
     def deserialize_rmm_device_buffer(header, frames):
         (arr,) = frames
 
-        # Copy data into new `DeviceBuffer` if needed
-        if not isinstance(arr, rmm.DeviceBuffer):
-            (ptr, _) = arr.__cuda_array_interface__["data"]
-            (size,) = header["shape"]
-            arr = rmm.DeviceBuffer(ptr=ptr, size=size)
+        # We should already have `DeviceBuffer`
+        # as RMM is used preferably for allocations
+        # when it is available (as it is here).
+        assert isinstance(arr, rmm.DeviceBuffer)
 
         return arr
--- a/distributed/protocol/rmm.py
+++ b/distributed/protocol/rmm.py
@@ -1,0 +1,24 @@
+import rmm
+from .cuda import cuda_serialize, cuda_deserialize
+
+
+# Used for RMM 0.11.0+ otherwise Numba serializers used
+if hasattr(rmm, "DeviceBuffer"):
+    @cuda_serialize.register(rmm.DeviceBuffer)
+    def serialize_rmm_device_buffer(x):
+        header = x.__cuda_array_interface__.copy()
+        frames = [x]
+        return header, frames
+
+
+    @cuda_deserialize.register(rmm.DeviceBuffer)
+    def deserialize_rmm_device_buffer(header, frames):
+        (arr,) = frames
+
+        # Copy data into new `DeviceBuffer` if needed
+        if not isinstance(arr, rmm.DeviceBuffer):
+            (ptr, _) = arr.__cuda_array_interface__["data"]
+            (size,) = header["shape"]
+            arr = rmm.DeviceBuffer(ptr=ptr, size=size)
+
+        return arr
--- a/distributed/protocol/rmm.py
+++ b/distributed/protocol/rmm.py
@@ -4,12 +4,12 @@ from .cuda import cuda_serialize, cuda_deserialize
 
 # Used for RMM 0.11.0+ otherwise Numba serializers used
 if hasattr(rmm, "DeviceBuffer"):
+
     @cuda_serialize.register(rmm.DeviceBuffer)
     def serialize_rmm_device_buffer(x):
         header = x.__cuda_array_interface__.copy()
         frames = [x]
         return header, frames
-
 
     @cuda_deserialize.register(rmm.DeviceBuffer)
     def deserialize_rmm_device_buffer(header, frames):
--- a/distributed/protocol/tests/test_cupy.py
+++ b/distributed/protocol/tests/test_cupy.py
@@ -29,3 +29,23 @@ def test_serialize_cupy_from_numba(dtype):
     y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle", "error"))
 
     assert (x_np == cupy.asnumpy(y)).all()
+
+
+@pytest.mark.parametrize("size", [0, 3, 10])
+def test_serialize_cupy_from_rmm(size):
+    np = pytest.importorskip("numpy")
+    rmm = pytest.importorskip("rmm")
+
+    x_np = np.arange(size, dtype="u1")
+
+    x_np_desc = x_np.__array_interface__
+    (x_np_ptr, _) = x_np_desc["data"]
+    (x_np_size,) = x_np_desc["shape"]
+    x = rmm.DeviceBuffer(ptr=x_np_ptr, size=x_np_size)
+
+    header, frames = serialize(x, serializers=("cuda", "dask", "pickle"))
+    header["type-serialized"] = pickle.dumps(cupy.ndarray)
+
+    y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle", "error"))
+
+    assert (x_np == cupy.asnumpy(y)).all()
--- a/distributed/protocol/tests/test_rmm.py
+++ b/distributed/protocol/tests/test_rmm.py
@@ -19,7 +19,7 @@ def test_serialize_rmm_device_buffer(size):
 
     header, frames = serialize(x, serializers=("cuda", "dask", "pickle"))
     y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle", "error"))
-    y_np = cuda.as_cuda_array(y).copy_to_host()
+    y_np = y.copy_to_host()
 
     assert (x_np == y_np).all()
 
--- a/distributed/protocol/tests/test_rmm.py
+++ b/distributed/protocol/tests/test_rmm.py
@@ -9,12 +9,11 @@ rmm = pytest.importorskip("rmm")
 
 @pytest.mark.parametrize("size", [0, 3, 10])
 def test_serialize_rmm_device_buffer(size):
-    DeviceBuffer = getattr(rmm, "DeviceBuffer", None)
-    if not DeviceBuffer:
+    if not hasattr(rmm, "DeviceBuffer"):
         pytest.skip("RMM pre-0.11.0 does not have DeviceBuffer")
 
     x_np = numpy.arange(size, dtype="u1")
-    x = DeviceBuffer(size=size)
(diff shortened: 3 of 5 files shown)


jakirkham:
cc @madsbk (please take a look as well 🙂)

distributed/protocol/rmm.py
```diff
@@ -0,0 +1,22 @@
...
+
+@cuda_deserialize.register(rmm.DeviceBuffer)
+def deserialize_rmm_device_buffer(header, frames):
+    (arr,) = frames
+
+    # Copy data into new `DeviceBuffer` if needed
+    if not isinstance(arr, rmm.DeviceBuffer):
+        (ptr, _) = arr.__cuda_array_interface__["data"]
+        (size,) = header["shape"]
+        arr = rmm.DeviceBuffer(ptr=ptr, size=size)
```
jakirkham: Not actually sure this special case is necessary/possible. If RMM is available, we already [allocate a `DeviceBuffer`]( https://github.com/dask/distributed/blob/4a8a4f3bce378406e83a099e8a12fc9bc12ef25c/distributed/comm/ucx.py#L50 ) and [fill it]( https://github.com/dask/distributed/blob/4a8a4f3bce378406e83a099e8a12fc9bc12ef25c/distributed/comm/ucx.py#L188 ). So it might be ok to drop this code. Left it in as a preventive measure just in case things get changed around later.

distributed/protocol/rmm.py
```diff
@@ -0,0 +1,24 @@
...
+        frames = [x]
+        return header, frames
+
+
+    @cuda_deserialize.register(rmm.DeviceBuffer)
+    def deserialize_rmm_device_buffer(header, frames):
+        (arr,) = frames
+
+        # Copy data into new `DeviceBuffer` if needed
+        if not isinstance(arr, rmm.DeviceBuffer):
```
TomAugspurger: Can you help me understand when this occurs?

distributed/protocol/tests/test_cupy.py
```diff
@@ -29,3 +29,23 @@ def test_serialize_cupy_from_numba(dtype):
     y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle", "error"))
 
     assert (x_np == cupy.asnumpy(y)).all()
+
+
+@pytest.mark.parametrize("size", [0, 3, 10])
+def test_serialize_cupy_from_rmm(size):
```
pentschev: Not sure I understand this test, why is the original array a NumPy one instead of a CuPy array?

pentschev: Nevermind, I just saw other tests are written in the same way, so no reason to change things.

distributed/protocol/rmm.py
madsbk: I guess must of the time?
AFAIK unless manually called, `arr` isn't a `rmm.DeviceBuffer`.

distributed/protocol/tests/test_rmm.py
```diff
@@ -0,0 +1,41 @@
...
+    if not DeviceBuffer:
+        pytest.skip("RMM pre-0.11.0 does not have DeviceBuffer")
+
+    x_np = numpy.arange(size, dtype="u1")
+    x = DeviceBuffer(size=size)
+    cuda.to_device(x_np, to=cuda.as_cuda_array(x))
+
+    header, frames = serialize(x, serializers=("cuda", "dask", "pickle"))
+    y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle", "error"))
+    y_np = cuda.as_cuda_array(y).copy_to_host()
```
pentschev: Is the call to `as_cuda_array` necessary? Shouldn't `deserialize` take care of that?

distributed/protocol/tests/test_rmm.py
```diff
@@ -0,0 +1,41 @@
...
+    if not DeviceBuffer:
+        pytest.skip("RMM pre-0.11.0 does not have DeviceBuffer")
+
+    x_np = numpy.arange(size, dtype="u1")
+    x = cuda.to_device(x_np)
+
+    header, frames = serialize(x, serializers=("cuda", "dask", "pickle"))
+    header["type-serialized"] = pickle.dumps(DeviceBuffer)
+    y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle", "error"))
+    y_np = cuda.as_cuda_array(y).copy_to_host()
```
pentschev: Ditto.

quasiben:
Can you elaborate on when users might be passing raw RMM buffers around as opposed to higher level objects like cuDF which use RMM?

distributed/protocol/rmm.py
jakirkham: Guessing you are referring to [this piece of code]( https://github.com/dask/distributed/blob/4a8a4f3bce378406e83a099e8a12fc9bc12ef25c/distributed/comm/ucx.py#L46-L63 )?

Probably this is never needed. Though maybe it comes up if someone changes that code. Perhaps it may also come up if users have different packages installed in different places. Maybe there are other cases that would be issues? FWIW we also cover this case in the tests.

That said, not attached to having this here if you think it should just be dropped.

distributed/protocol/tests/test_cupy.py
jakirkham: No worries. Yeah it seemed prudent to test building other objects from RMM `DeviceBuffer`'s more directly since this is something we implicitly rely on for all GPU objects (when RMM is available).

distributed/protocol/tests/test_rmm.py
jakirkham: Good catch! Will try that change and push if it passes locally.

distributed/protocol/tests/test_rmm.py
jakirkham: Actually that's not the case here as this will be a `rmm.DeviceBuffer`. See the `header["type-serialized"]` line above for context.

distributed/protocol/tests/test_rmm.py
jakirkham: Dropped. Thanks 🙂

jakirkham:
Well even those higher-level objects use lower-level objects to perform serialization.

Right now [they are building]( https://github.com/rapidsai/cudf/blob/1ead2d55562b9cd1b80568b86b8b51ab616f3cb8/python/cudf/cudf/core/column/column.py#L74-L76 ) a [Numba array]( https://github.com/rapidsai/rmm/blob/47cb6b876022bdf0a9205c18c21479e9238d7894/python/rmm/rmm.py#L153-L155 ) to [send over each time]( https://github.com/rapidsai/cudf/blob/1ead2d55562b9cd1b80568b86b8b51ab616f3cb8/python/cudf/cudf/core/column/column.py#L847 ). We can probably cut out this overhead and send `DeviceBuffer`'s (and other objects we already handle) directly.

Though this assumes we have a way to send `DeviceBuffer`'s in the first place. 😉

Independently this can be useful for low-level performance testing of ucx-py where we may be concerned about the overhead of building other objects. 🙂

distributed/protocol/tests/test_rmm.py
pentschev: Ah, right, that makes sense. Thanks for clarifying!

quasiben:
Merging in later today unless others have comments

distributed/protocol/rmm.py
jakirkham: Otherwise I think this is analogous to what other deserializers (like [the one used by Numba]( https://github.com/dask/distributed/blob/88c354f68761b60c11ba5613806891f78810da0c/distributed/protocol/numba.py#L34 )) do. Please let me know if there are any other questions though @TomAugspurger. 🙂

distributed/protocol/tests/test_rmm.py
jakirkham: No worries. Thanks for checking 🙂

distributed/protocol/rmm.py
mrocklin: If we know that the frames are not going to be of this type then I recommend that we remove the condition.  If someone changes some code and this breaks then that's probably good, we'll know that a contract was broken due to tests presumably and we can fix it then.

distributed/protocol/tests/test_rmm.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@pytest.mark.parametrize("size", [0, 3, 10])
+def test_serialize_rmm_device_buffer(size):
+    DeviceBuffer = getattr(rmm, "DeviceBuffer", None)
+    if not DeviceBuffer:
+        pytest.skip("RMM pre-0.11.0 does not have DeviceBuffer")
+
+    x_np = numpy.arange(size, dtype="u1")
+    x = DeviceBuffer(size=size)
```
mrocklin: I have a slight style preference to keep terms like `DeviceBuffer` namespaced and refer to it as `rmm.DeviceBuffer`.  I think that this is especially valuable when dealing with libraries that most Dask maintainers are unlikely to be familiar with.

distributed/protocol/rmm.py
jakirkham: Perhaps. Though I don't know that we test that assumption currently, which would make this code (if with the referenced section removed) an implicit test of that assumption. So I expect that would make it hard for someone to connect their change to test failures. Much better in my mind to safely handle unknown input types here (IOW leaving this in).

In any event I'm planning to break this somewhat tenuous assumption to fix another issue in a different PR. So it seems reasonable to let this stay. Besides we heavily test this `if` anyways.

Though please let me know if I'm missing other concerns. 🙂

distributed/protocol/tests/test_rmm.py
jakirkham: Sure that seems reasonable. Will update momentarily.

jakirkham: Ok have pushed this change. Please take a look 🙂

distributed/protocol/rmm.py
mrocklin: If you are concerned about type things then I would rather have assertion statements about the types than code being unnecessarily flexible.  In general in the distributed scheduler code we tend towards strictness (I recommend checking out the validation code within scheduler.py at some point).  It's important to avoid hard to diagnose bugs. 

If you are intending for ucx-py to return rmm.DeviceBuffers at some point then lets talk about this then, but for now I would appreciate it if you would remove the conditional.

jakirkham: Alright. I remain concerned about dropping this code, but have done so. Also have dropped the test associated with it. Have added an assertion and a comment explaining its intent. Please take another look.

jakirkham:
Have pushed the remaining changes requested. Please let me know if there is anything else. FWIW I've also been running the tests on a DGX-1. So can verify they do pass.

jakirkham:
Planning on merging tomorrow if no further comments.

PR #3443, opened 2020-02-05
WIP: Handle other contiguous CUDA arrays

jakirkham:
To avoid copying in more cases, check for different types of contiguous arrays and make sure they are sent through as-is. For arrays that we still cannot recognize as contiguous, go ahead and copy them to C-order (as was the case previously) before sending them over.


mrocklin:
Is this still an active effort, or should we close this out?

jakirkham:
Superseded by PR ( https://github.com/dask/distributed/pull/3475 ).

PR #3446, opened 2020-02-05
Adjust `numba.cuda` import and add check

jakirkham:
It seems `numba` now raises an `AttributeError` when accessing `numba.cuda` if it is not already imported. So go ahead and import `numba.cuda` in the test to ensure it is available. Also copies over @pentschev's fix from PR ( https://github.com/dask/distributed/pull/3255 ) to check that CUDA is available (or otherwise skip the test).

### DIFF
--- a/distributed/protocol/tests/test_cupy.py
+++ b/distributed/protocol/tests/test_cupy.py
@@ -20,6 +20,9 @@ def test_serialize_cupy_from_numba(dtype):
     cuda = pytest.importorskip("numba.cuda")
     np = pytest.importorskip("numpy")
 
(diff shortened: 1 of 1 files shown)


pentschev:
LGTM, tests failing are certainly unrelated, given CI for CUDA tests is unavailable.

quasiben:
@pentschev you are right.  I restarted the test for safe measure

jakirkham:
Looks like whatever CI issue has gone away. Please let me know if anything else is needed here 🙂

quasiben:
Thanks @jakirkham

PR #3453, opened 2020-02-06
WIP: Aggregate received data into a large allocation

jakirkham:
**Note:** This is **still** being worked. So there may be bugs, test failures, bad behavior, etc. At this stage, all of these should be expected. You have been warned. 😉

<br>

As we have seen some issues due to churn and fragmentation from small allocations, this tries to allocate one large buffer up front before receiving. This buffer is then split up into small views onto that larger allocation. Once this setup is done we receive into all of these views. The views are then passed to deserialization much as the smaller buffers were before.

We take views onto the data so send and deserialization methods can work on data they are familiar with. A follow-up to this work might be to figure out how deserialization and receive can work on larger buffers to start with to avoid this split of the buffer into views. Though the intent is to not require this at this stage and only follow-up on it as needed later.

To help illustrate the motivation of this this change. Assume we are receiving a Dataframe. Previously we would create multiple allocations for each column along with their relevant internal data (masks, indices, etc.). With this change, we should make one host (and one device) allocation for the whole Dataframe and then fill out its content.

As a bonus of performing this allocation up front and parceling it into the relevant views, we are able to get rid of the old looping behavior around each receive and instead hand all of these to asyncio. This could potentially (though this yet to be verified) get more out of the available bandwidth and benefit from things like non-blocking sends.

Have added this change to both TCP and UCX. The former to aid in testing and discussion. The latter is of interest for improving performance in that use case. Though hopefully both will get a boost from this work.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -30,10 +30,7 @@ import asyncio
 from itertools import starmap
 from operator import add
 
-try:
-    from cytoolz import accumulate, cons, sliding_window
-except ImportError:
-    from toolz import accumulate, cons, sliding_window
+from tlz import accumulate, cons, sliding_window
 
 logger = logging.getLogger(__name__)
 
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -26,6 +26,14 @@ from .addressing import parse_host_port, unparse_host_port
 from .core import Comm, Connector, Listener, CommClosedError, FatalCommClosedError
 from .utils import to_frames, from_frames, get_tcp_server_address, ensure_concrete_host
 
+import asyncio
+from itertools import starmap
+from operator import add
+
+try:
+    from cytoolz import accumulate, cons, sliding_window
+except ImportError:
+    from toolz import accumulate, cons, sliding_window
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


mrocklin:
Interesting idea.  I would be curious to see what performance differences are like.  My *hope* would be that this is not necessary with TCP, and wouldn't have much of an effect with UCX *if* we're using RMM, but I wouldn't be surprised if myintuition was wrong here.

jakirkham:
The issue is RMM uses CNMEM under-the-hood, which is know to have issues freeing lots of memory (like small chunks). Keith [mentioned this]( https://github.com/rapidsai/ucx-py/issues/402#issuecomment-580770932 ) and [added some numbers to back this up]( https://github.com/rapidsai/ucx-py/issues/402#issuecomment-580776895 ). So this effectively cuts down on the number of times free needs to be called per send, which should improve overall memory allocation performance. Of course some of this occurs on the critical path, but it also happens in other places too 😉

As to TCP, it really depends on how the Python memory manager works (given we are using `bytearray`). That said, we might want to change that to optionally use NumPy to benefit from the HUGEPAGES work ( https://github.com/numpy/numpy/pull/14216 ), at which point we would want to know how NumPy handles memory management. My guess is it would be biased towards handling large allocations efficiently, but I could always be wrong about this 🙂 In event this is largely orthogonal and should be saved for a separate discussion thread (feel free to open one if it is of interest). For now TCP is just handy for fixing bugs in my approach 😄

I can't say at this point. Still sorting out bugs.

It's worth noting that this may affect things outside of communication itself as it affects how allocations are done, which will affect other allocations that may not be in the communication path (like spilling or creating intermediates for example). So a simple benchmark UCX with/without this change is likely to miss many things or possibly focus on the wrong things. This is all conjecture, but do want to make sure we are aware of the forrest and not just one tree in it. 🙂

To summarize, the idea of the work here is to then cutdown the number of times alloc/free are called by some factor and then see how overall workflows are impacted.

Anyways probably wouldn't follow this too closely for now as it is still WIP. Trying not to distract people while things are still in an early stage 😉

mrocklin:
> To summarize, the idea of the work here is to then cutdown the number of
times alloc/free are called by some factor and then see how overall
workflows are impacted.

+1.  I look forward to seeing how this performs.

mrocklin:
Checking in here.  Is this still an active effort?  No worries either way, I'm just going through old PRs and seeing what we can close out.

jakirkham:
I have had my hands full, but yes this remains important.

jakirkham:
Closing out in favor of PR ( https://github.com/dask/distributed/pull/3732 ).

jakirkham:
Somewhat related to the approach that was tried in PR ( https://github.com/dask/distributed/pull/5487 ) except that looked at `write` in addition to `read`

Issue #1674, opened 2017-12-31
"Stream is closed"


rabernat:
I am having workers die with the following error messages.

This occurs when I am trying to `.persist()` large xarray dataset into memory. (There is more than enough memory in the cluster for the dataset by a factor of 5.)

I don't know what these errors mean, other than that the workers have died. Advice would be appreciated on how to debug more effectively.

```
distributed.worker - ERROR - failed during get data
Traceback (most recent call last):
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/distributed/comm/tcp.py", line 221, in write
    yield future
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
tornado.iostream.StreamClosedError: Stream is closed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/distributed/worker.py", line 524, in get_data
    compressed = yield comm.write(msg)
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/distributed/comm/tcp.py", line 225, in write
    convert_stream_closed_error(self, e)
  File "/rigel/ocp/users/ra2697/conda/envs/pangeo/lib/python3.6/site-packages/distributed/comm/tcp.py", line 124, in convert_stream_closed_error
    raise CommClosedError("in %s: %s: %s" % (obj, exc.__class__.__name__, exc))
[...]

PR #3455, opened 2020-02-08
Workaround RecursionError on profile data

TomAugspurger:
@mrocklin would be curious to hear of any alternatives. This is very much a workaround rather than a proper fix.

### DIFF
--- a/distributed/comm/utils.py
+++ b/distributed/comm/utils.py
@@ -1,5 +1,5 @@
-import sys
 import logging
+import math
 import socket
 
 import dask
@@ -41,7 +41,7 @@ async def to_frames(msg, serializers=None, on_error="message", context=None):
     try:
         msg_size = sizeof(msg)
     except RecursionError:
-        msg_size = sys.getsizeof(msg)
+        msg_size = math.inf
 
     if FRAME_OFFLOAD_THRESHOLD and msg_size > FRAME_OFFLOAD_THRESHOLD:
         return await offload(_to_frames)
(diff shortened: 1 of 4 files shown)


distributed/tests/test_profile.py
```diff
@@ -199,3 +202,16 @@ def stop():
...
+    # https://github.com/dask/distributed/issues/1674
+    n = 500
+    original = outer = {}
+    inner = {}
+
+    for i in range(n):
+        outer["children"] = inner
+        outer, inner = inner, {}
+    profile = Profile(original)
+    sizeof(profile)
```
TomAugspurger: This raises a RecursionError on `original`.

mrocklin:
Ah, I was walking through my e-mail chronologically.  I think that this is maybe a more robust solution to the problem?  https://github.com/dask/distributed/issues/1674#issuecomment-583698574

distributed/comm/utils.py
```diff
@@ -37,7 +38,12 @@ def _to_frames():
             logger.exception(e)
             raise
 
-    if FRAME_OFFLOAD_THRESHOLD and sizeof(msg) > FRAME_OFFLOAD_THRESHOLD:
+    try:
+        msg_size = sizeof(msg)
+    except RecursionError:
+        msg_size = sys.getsizeof(msg)
```
mrocklin: I suggest instead that we default to offloading the serialization work in this case.  This object has proven that it's odd, so it probably makes sense to move it off to another thread for safety.

```suggestion
        msg_size = math.inf
```

mrocklin:
Thanks @TomAugspurger !

PR #3464, opened 2020-02-10
Add total row to workers plot in dashboard

jsignell:
Closes https://github.com/dask/dask/issues/5275

image

Note that the `Total` is also included in the circle plots above the table. I added the worker name to the hover to make it clear which one is which:

Screenshot from 2020-02-10 13-40-10

But I would like some input on whether to take it out of those, or make it another color.

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1556,7 +1556,7 @@ class WorkerTable(DashboardComponent):
             point_policy="follow_mouse",
             tooltips="""
                 <div>
-                  <span style="font-size: 10px; font-family: Monaco, monospace;">@worker: </span>
+                  <span style="font-size: 10px; font-family: Monaco, monospace;">Worker (@name): </span>
                   <span style="font-size: 10px; font-family: Monaco, monospace;">@memory_percent</span>
                 </div>
                 """,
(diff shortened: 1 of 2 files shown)


mrocklin:
I feel like we should visually promote the total value a bit more.  Having it at the bottom of the list makes it seem unimportant, particularly if there are more workers than there are visible rows.  We could put it at the top, or if we wanted to, maybe even separate it into its own standalone table just above the current workers table?

jsignell:
Yeah I'd almost like to bold it or something. Here it is at the top. See what you think of that:

image

mrocklin:
At the top definitely seems nicer to me.

jsignell:
And do you have an opinion about including the total in the circles plots at the top?

mrocklin:
I'm personally less excited about that.  In my experience when people hover over or select those they're looking to highlight a pathological worker, and at that point don't care strongly about how many workers there are.

No strong feelings though.  I'm happy to defer to your judgment.

jsignell:
Well, it'd be more work to exclude the totals from the circles then to just leave it in, so if people tend to use those plots via hover I'll just leave it in, since it shouldn't t upset them to have the total in there.

mrocklin:
This looks good to me.  It looks like the only CI failure is linting.  You might want to install the git commit hooks with pre-commit.  https://docs.dask.org/en/latest/develop.html#code-formatting

jsignell:
Oh I only did the pip install, not the second step. Sorry about that.

mrocklin:
Looks good to me.  Thanks @jsignell !

PR #3466, opened 2020-02-11
Split dashboard host on additional slashes to handle inproc

### LINKED ISSUE #3456: processes=False breaks Client dashboard
Hi, I'm on Windows 10 with JupyterLab in Chrome. Was just going through the official sites tutorials and noticed that half of the Client dashboards wouldn't open(empty localhost page):

```
`from dask.distributed import Client, progress

client = Client(processes=False, threads_per_worker=4,
                                              n_workers=1, memory_limit='2GB')
client`
```

The culprit is **processes=False**, by just removing it you can access dashboard. 
Thanks for the modules and cheers.

jacobtomlinson:
When the Dask scheduler is run with `processes=False` and the scheduler uses an `inproc://` address the dashboard url logic fails. This is because the scheduler host contains the process pid info separated by slashes `inproc://192.168.0.11/535/1`. 

Current logic strips off the protocol but leaves the pid info. This PR strips that off too and adds a test to avoid a regression.

Closes #3456

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -5255,6 +5255,12 @@ def test_dashboard_link(loop, monkeypatch):
                 assert link in text
 
 
+def test_dashboard_link_inproc(loop, monkeypatch):
+    with Client(processes=False, loop=loop) as c:
+        with dask.config.set({"distributed.dashboard.link": "{host}"}):
+            assert "/" not in c.dashboard_link
+
+
 @gen_test()
 def test_client_timeout_2():
     with dask.config.set({"distributed.comm.timeouts.connect": "10ms"}):
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -5255,8 +5255,9 @@ def test_dashboard_link(loop, monkeypatch):
                 assert link in text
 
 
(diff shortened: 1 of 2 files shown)


distributed/tests/test_client.py
```diff
@@ -5255,6 +5255,12 @@ def test_dashboard_link(loop, monkeypatch):
                 assert link in text
 
 
+def test_dashboard_link_inproc(loop, monkeypatch):
```
mrocklin: Why monkeypatch here?

If you're comfortable with it, I recommend defaulting to asynchronous tests.  They're faster and easier to debug if anything goes wrong in the future.  Also, if you do end up using synchronous tests then I encourage you to include the `cleanup` fixture.

```python
@pytest.mark.asyncio
async def test_dashboard_link_inproc(cleanup):
    async with Client(processes=False, asynchronous=True) as c: 
        ...

jacobtomlinson: Hmm that's some bad copy/paste from the test above. Apologies.

I'll switch it to async.

mrocklin:
OK.  Merging in.  Thanks @jacobtomlinson !

PR #3468, opened 2020-02-11
add last seen column to worker table

### LINKED ISSUE #2595: `last_seen` in worker info page
When workers die or halted it can be useful to see when that worker was last seen by the scheduler.  We should bubble this information up to the dashboard

> client.scheduler_info()
```
'tcp://172.17.0.2:43161': {'type': 'Worker',
   'id': 'tcp://172.17.0.2:43161',
   'host': '172.17.0.2',
   'resources': {},
   'local_directory': '/notebooks/dask-worker-space/worker-k540965f',
   'name': 'tcp://172.17.0.2:43161',
   'ncores': 1,
   'memory_limit': 67619300864,
   'last_seen': 1554412409.6942942,
   'services': {'nanny': 46010, 'bokeh': 34120},
   'metrics': {'cpu': 101.7,
    'memory': 472141824,
    'time': 1554412409.1895998,
    'read_bytes': 10374.966431182967,
    'write_bytes': 38095.74065486942,
    'num_fds': 69,
    'executing': 1,
    'in_memory': 2,
    'ready': 2,
    'in_flight': 1}}}}
```

kaelgreco:
Had some problems with stalled workers and thought it'd be helpful to surface `last_seen` in the web interface. This change adds a column for "Last seen" in `worker-table.html`. It also might be worth coloring the row to raise visibility? I'm not super familiar with jinja templates but I'd be happy to take a stab at it if anyone has suggestions.

Fixes #2595

### DIFF
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -4,6 +4,7 @@ import os
 import os.path
 import json
 import logging
+from time import time
 
 import dask
 from dask.utils import format_bytes
@@ -55,7 +56,8 @@ from ..scheduler import ALL_TASK_STATES
 
 
 ns = {
-    func.__name__: func for func in [format_bytes, format_time, datetime.fromtimestamp]
+    func.__name__: func
+    for func in [format_bytes, format_time, datetime.fromtimestamp, time]
 }
 
 rel_path_statics = {"rel_path_statics": "../../"}
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -4,7 +4,6 @@ import os
 import os.path
 import json
 import logging
-from time import time
 
 import dask
 from dask.utils import format_bytes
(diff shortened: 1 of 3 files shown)


distributed/dashboard/scheduler.py
```diff
@@ -4,6 +4,7 @@
 import os.path
 import json
 import logging
+from time import time
```
mrocklin: In Windows `time` returns integer values, which can be unpleasant.  

We sometimes use either `IOLoop.current().time()` or `distributed.metrics.time()`.  My guess is that these values are computed with `metrics.time` and so it's probably best to use that for cross-platform support.

distributed/dashboard/templates/worker-table.html
```diff
@@ -27,6 +28,7 @@
         <td> </td>
```
mrocklin: My guess is that we would want to conditionally add a `style=` keyword to the `tr` block just above this.

I'm also not an expert in Jinja, but a bit of stack overflow searching lead to https://stackoverflow.com/a/47610447 which might be enough?

mrocklin:
Fixes https://github.com/dask/distributed/issues/2595

kaelgreco:
Managed to color the row red if time since last seen > 60s. Let me know if there's a more appropriate threshold.

here's an example of the color:
image

mrocklin:
This looks great to me.  Merging in.  Thanks @kaelgreco !

PR #3469, opened 2020-02-12
Minor gen.Return cleanup

jrbourbeau:
This PR fixes a couple of instances where `gen.Return` wasn't being used properly


mrocklin:
Thanks @jrbourbeau !

PR #3473, opened 2020-02-13
Rename logs to get_logs

jacobtomlinson:
Currently the scheduler has an RPC method called `logs` which maps to a `get_logs` method underneath.

Libraries downstream such as dask-kubernetes have followed the practice of using `logs` as they often communicate via RPC. However in instances where they interact directly they need to call `get_logs` instead which involves implementing alternative logic.

This PR renames `logs` to `get_logs` in many places and adds a shim to the `Cluster` to ensure backward compatibility. I may have missed other areas for backward compatibility though.

~I've also made the worker consistent with this as it had a `get_logs` RPC call.~

xref dask/dask-kubernetes#232

### DIFF
--- a/distributed/node.py
+++ b/distributed/node.py
@@ -145,6 +145,10 @@ class ServerNode(Node, Server):
         weakref.finalize(self, logger.removeHandler, self._deque_handler)
 
     def get_logs(self, comm=None, n=None):
+        # Deprecated use logs instead
+        return self.logs(comm=comm, n=None)
+
+    def logs(self, comm=None, n=None):
         deque_handler = self._deque_handler
         if n is None:
             L = list(deque_handler.deque)
--- a/distributed/node.py
+++ b/distributed/node.py
@@ -146,7 +146,7 @@ class ServerNode(Node, Server):
 
     def get_logs(self, comm=None, n=None):
         # Deprecated use logs instead
-        return self.logs(comm=comm, n=None)
+        return self.logs(comm=comm, n=n)
 
     def logs(self, comm=None, n=None):
         deque_handler = self._deque_handler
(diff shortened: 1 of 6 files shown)


mrocklin:
+1 from me

distributed/node.py
```diff
@@ -145,6 +145,10 @@ def _setup_logging(self, logger):
         weakref.finalize(self, logger.removeHandler, self._deque_handler)
 
     def get_logs(self, comm=None, n=None):
+        # Deprecated use logs instead
+        return self.logs(comm=comm, n=None)
```
jrbourbeau:
```suggestion
        return self.logs(comm=comm, n=n)
```

distributed/worker.py
```diff
@@ -617,7 +617,7 @@ def __init__(
             "call_stack": self.get_call_stack,
             "profile": self.get_profile,
             "profile_metadata": self.get_profile_metadata,
-            "get_logs": self.get_logs,
+            "logs": self.logs,
```
jrbourbeau: The `Worker` class seems to already have a `logs` property defined

https://github.com/dask/distributed/blob/ca88aa7327820fc934b2254b884721cea94991e5/distributed/worker.py#L731-L733

I haven't looked into the CI failures too deeply, but perhaps it's related

mrocklin: Ah, that's unfortunate.  This change might not make sense I guess.  I suspect that a few folks  *do* use this, so we might want to go through a deprecation process if we do this, which might not be worth the effort.

jacobtomlinson: Ah I missed that!

They seem to have similar functionality, although the `get_logs` does a bit more checking and formatting.

I guess the main change would be going from a property to a method.

jacobtomlinson:
If this PR doesn't make sense then perhaps changing the scheduler RPC to be `get_logs` would?

mrocklin:
Yes, that could make sense.  It's also easier to keep backwards compatibility in that case.

jacobtomlinson:
Ok I've gone the other way and renamed `logs` to `get_logs`.

quasiben:
Thanks @jacobtomlinson for the PR and @mrocklin and @jrbourbeau for reviews!

PR #3474, opened 2020-02-13
Update NumPy array serialization to handle non-contiguous slices

jrbourbeau:
Our current serialization of NumPy arrays with a zero stride fails when the slice taken here

https://github.com/dask/distributed/blob/ca88aa7327820fc934b2254b884721cea94991e5/distributed/protocol/numpy.py#L50

results in a non-contiguous memory layout. For example

```python
import numpy as np
from distributed.protocol import serialize, deserialize

x = np.broadcast_to(np.arange(5), (4, 5))[:, ::-1]
y = deserialize(*serialize(x))
print(y)
```

will currently fail with `ValueError: strides is incompatible with shape of requested array and size of buffer` because the sliced version of the array is non-contiguous and breaks when broadcasting later during deserialization

This PR adds an additional contiguous check and will create a contiguous version of an array with the appropriate strides when needed

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -44,12 +44,21 @@ def serialize_numpy_ndarray(x):
     else:
         dt = (0, x.dtype.str)
 
-    # Only serialize non-broadcasted data for arrays with zero strided axes
+    # Only serialize broadcastable data for arrays with zero strided axes
+    broadcast_to = None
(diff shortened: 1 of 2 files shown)


mrocklin:
I'll admit to not fully grokking what is going on here, but tests pass and things don't seem overly complex, so I'm merging in.  Thanks @jrbourbeau !

PR #3475, opened 2020-02-14
Serialize 1-D, contiguous, `uint8` CUDA frames

jakirkham:
When performing serialization of CUDA frames, make sure they are converted to 1-D, contiguous, `uint8` frames. This makes them significantly easier to work with in later pipeline steps that operate on this serialized data. For example checking the data is contiguous is as simple as ensuring the `strides` are `(1,)`. Similarly the number of bytes to transmit can be extracted from the `shape` as it will be `(nbytes,)`. Additionally this gives us greater confidence that generic objects that merely contain some number of bytes can adequately receive more complex objects and represent them correctly. Finally this has the added benefit of handling F-contiguous data just as easily as C-contiguous data is handled, which can be useful for algorithms expecting this particular data layout. As we now guarantee `strides` is not `None` to simplify data reconstruction, we were able to get rid of some older workarounds and hopefully clarify things a bit.

### DIFF
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -43,6 +43,7 @@ def serialize_cupy_ndarray(x):
         x = cupy.array(x, copy=True)
 
     header = x.__cuda_array_interface__.copy()
+    header["strides"] = tuple(x.strides)
     frames = [x]
 
     return header, frames
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -55,6 +55,9 @@ def deserialize_cupy_array(header, frames):
     if not isinstance(frame, cupy.ndarray):
         frame = PatchedCudaArrayInterface(frame)
     arr = cupy.ndarray(
-        shape=header["shape"], dtype=header["typestr"], memptr=cupy.asarray(frame).data
+        shape=header["shape"],
+        dtype=header["typestr"],
+        memptr=cupy.asarray(frame).data,
+        strides=header["strides"],
     )
     return arr
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -33,7 +33,7 @@ class PatchedCudaArrayInterface:
 @cuda_serialize.register(cupy.ndarray)
 def serialize_cupy_ndarray(x):
     # Making sure `x` is behaving
-    if not x.flags["C_CONTIGUOUS"]:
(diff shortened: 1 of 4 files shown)


jakirkham:
Also if you could take a look as well, @madsbk. That would be great! 😄

distributed/protocol/cupy.py
```diff
@@ -39,11 +33,18 @@ def __del__(self):
 @cuda_serialize.register(cupy.ndarray)
 def serialize_cupy_ndarray(x):
     # Making sure `x` is behaving
-    if not x.flags.c_contiguous:
+    if not (x.flags["C_CONTIGUOUS"] or x.flags["F_CONTIGUOUS"]):
```
madsbk: We could take this a step further and only make a copy if `x` is truly non-contiguous. I.e when checking contiguous, we can ignore all 0-strided (broadcasted), 1-sized dimensions, and the dimension order. 
We are properly not sending broadcasted arrays that often but this copy can makes it *very* expensive.

jakirkham: Are you thinking of PR ( https://github.com/dask/distributed/pull/3474 )? Yeah it would be helpful to leverage a lot of the same tricks we are using with NumPy on other array-like serialization methods. Not entirely sure how we do that, but open to suggestions 🙂

F-order in particular is of interest for cuML, which largely uses F-order arrays. Hence the interest in handling that case zero-copy here.

cc @jrbourbeau (in case you have ideas 😉)

pentschev:
LGTM. Thanks @jakirkham .

distributed/protocol/cupy.py
```diff
@@ -6,21 +6,15 @@
...
+        1) Makes sure that the cuda context is active
            when deallocating the base cuda array.
         Notice, this is only needed when the array to deserialize
         isn't a native cupy array.
     """
 
     def __init__(self, ary):
-        cai = ary.__cuda_array_interface__
-        cai_cupy_vsn = cupy.ndarray(0).__cuda_array_interface__["version"]
-        if cai.get("strides") is None and cai_cupy_vsn < 2:
```
rjzamora: My memory of this is fuzzy - Is there no longer a chance the cupy/numba combination can be a problem here?

jakirkham: Good question. In this PR are we are now always storing the actual `strides` in the `header`. So no it's not possible.

jakirkham:
Planning on merging in 24hrs if no further comments.

jakirkham:
Thanks all for the reviews! 😄

PR #3478, opened 2020-02-14
Register Dask cuDF serializers

jakirkham:
As cuDF is gaining support for serializing using the Dask protocol as well ( https://github.com/rapidsai/cudf/pull/4153 ), make sure to register it's serializers there as well. cuDF versions lacking this support will continue to behave the same (falling back to pickle). However cuDF versions with this support will bypass pickle. So only pay the cost of a host-to-device transfer.

### DIFF
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -88,5 +88,7 @@ def _register_rmm():
 
 @cuda_serialize.register_lazy("cudf")
 @cuda_deserialize.register_lazy("cudf")
+@dask_serialize.register_lazy("cudf")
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @jakirkham .  This looks simple enough.  At first I was concerned about what happens if someone is using an older version of cudf, but I guess that this doesn't actually register anything with dask_serialize, it just gives cudf a chance to register something.

Merging.

jakirkham:
Thanks Matt! 😄

Yeah I think you summarized it well. We will try to `import cudf` in those cases, but nothing will happen. FWIW I tried all combinations locally just to make sure 🙂

PR #3480, opened 2020-02-14
Propose fix for collection based resources docs

chrisroat:
Besides a missing end brace, the dask keys need to be flattened before they can be turned into a tuple (they can contain lists, which are not hashable as dict keys).


mrocklin:
Thanks @chrisroat !

PR #3482, opened 2020-02-15
Add dask serialization of CUDA objects

jakirkham:
This implements `"dask"` serialization for CUDA objects. All serializers work by first performing `"cuda"` serialization on the different objects and then performing a device-to-host transfer. The deserializers perform a host-to-device transfer before running the respective `"cuda"` deserializer for the object considered. Should provide an alternative way to do TCP transfers with CUDA objects that does not rely on pickling.

Note: Just like with UCX transfers, we preferentially add objects to the RMM pool if RMM is available. Otherwise we fallback to using Numba.

### DIFF
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -82,6 +82,8 @@ def _register_numba():
 
 @cuda_serialize.register_lazy("rmm")
 @cuda_deserialize.register_lazy("rmm")
+@dask_serialize.register_lazy("rmm")
+@dask_deserialize.register_lazy("rmm")
 def _register_rmm():
     from . import rmm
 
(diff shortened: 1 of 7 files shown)


mrocklin:
Thanks @jakirkham !  For testing, what do you think about calling serialize on a few objects, like a cupy array, and verifying that some of the frames are `memoryview` objects?

jakirkham:
Yep, agree this needs some tests. Those sound like reasonable suggestions. Will work on those and update when they are up. Thanks Matt! 😄

jakirkham:
Have parametrized tests around serializers so this tests `cuda`, `dask`, and `pickle` (where possible). Also have tested that frames are acceptable. Please let me know if there is anything else. 🙂

Planning on merging in 24hrs if no comments.

PR #3483, opened 2020-02-15
Check exact equality for worker state

bnaul:
Per https://github.com/dask/distributed/pull/3321#discussion_r379826000, I agree it makes more sense to check for exact equality instead of relying on the hash (just in case 🙂).

I could see an argument for wrapping this tuple in a property so it can be reused, do we think that's preferable? I'd originally wanted to use `identity` for `hash` and `eq` but that has a bit too much going on IMO.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -276,13 +276,10 @@ class WorkerState:
         self.extra = extra or {}
 
     def __hash__(self):
-        return hash((self.name, self.host))
+        return hash(self.address)
 
     def __eq__(self, other):
-        return type(self) == type(other) and (self.name, self.host) == (
-            other.name,
-            other.host,
-        )
+        return type(self) == type(other) and self.address == other.address
 
     @property
     def host(self):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -279,7 +279,10 @@ class WorkerState:
         return hash((self.name, self.host))
 
     def __eq__(self, other):
-        return type(self) == type(other) and hash(self) == hash(other)
+        return type(self) == type(other) and (self.name, self.host) == (
+            other.name,
+            other.host,
+        )
 
     @property
     def host(self):


bnaul:
cc @StephanErb

mrocklin:
Thanks for picking this up @bnaul 

This looks good to me.  I'll merge after either a +1 from @StephanErb or a couple days.

mrocklin:
My guess is that all of the test failures are intermittent.  When we changed to spawn by default it unearthed a bunch of things.  Squashing down a couple now.

mrocklin:
Well, this one is interesting

```python
    @gen_cluster(timeout=1000, client=True)
    def test_recompute_released_key(c, s, a, b):
        x = c.submit(inc, 100)
        result1 = yield x
        xkey = x.key
        del x
        import gc
    
        gc.collect()
        yield gen.moment
>       assert c.refcount[xkey] == 0
E       assert 1 == 0
E         -1
E         +0
```

distributed/scheduler.py
```diff
@@ -279,7 +279,10 @@ def __hash__(self):
         return hash((self.name, self.host))
 
     def __eq__(self, other):
-        return type(self) == type(other) and hash(self) == hash(other)
+        return type(self) == type(other) and (self.name, self.host) == (
+            other.name,
+            other.host,
+        )
```
StephanErb: I am wondering if using `name` and `host` is correct here:
* `name` is optional and could be None
* `host` is not necessary unique as multiple workers could be running on the same machine

I am lacking necessary Dask context, so I am wondering why you have not picked `address` instead? It is used for indexing the `WorkerState` and the documentation says it is `This worker's unique key.`.

bnaul: I was just pulling items from the `identity()` function, no personal reason to prefer one to the other. @TomAugspurger @mrocklin would you agree `address` makes more sense here?

mrocklin:
Yes, address is probably the best thing to do in both cases.

bnaul:
thanks @StephanErb @mrocklin , updated w/ your suggestions

Issue #3467, opened 2020-02-11
Improved layout for task graph in dashboard


MichaelSchreier:
Over the weekend I played around with improving the visuals of the task graph in the online dashboard and was wondering if you were interested in me refining this for a PR.

Some examples:
* The [custom graph from the documentation ](https://docs.dask.org/en/latest/custom-graphs.html) becomes
  ![dask_graph_2](https://user-images.githubusercontent.com/25567300/74268539-4f13db80-4d08-11ea-9d04-c6390e6dc231.jpg)

* The code 
  ```python
  from distributed import Client, LocalCluster
  import dask.array as da

  cluster = LocalCluster()
  client = Client(cluster)
  x = da.ones((1500, 1500), chunks=(300, 300))

  y = x + x.T

  client.compute(y)
  ```
  becomes
  ![dask_graph_1](https://user-images.githubusercontent.com/25567300/74268690-913d1d00-4d08-11ea-9072-e1f2c73e80a5.jpg)

* A complex custom graph from my project at work transforms from this
  ![dask_custom_graph_old](https://user-images.githubusercontent.com/25567300/74268784-be89cb00-4d08-11ea-8813-a885d2263829.png)
  to this
  ![dask_custom_graph_new](https://user-images.githubusercontent.com/25567300/74268801-c77a9c80-4d08-11ea-98fd-730bdbd94175.png)

---

To get these results I did the following
* slightly tuned the colors to [more eye friendly versions ](http://colorbrewer2.org/?type=qualitative&scheme=Set1&n=3) (I just noted that #2711 exists) - loosely following [this ](https://commons.wikimedia.org/wiki/File:Topological_Ordering.svg#/media/File:Topological_Ordering.svg) wikipedia visualization
* replaced the straight edges with bezier curves
* added arrow heads to the edges
* removed the axes as they don't really add anything to the visualization
* replaced the `update_graph()` method in `graph_layout.py` with a version leveraging on the [Grandalf](https://github.com/bdcht/grandalf) package to compute a nicer layout
  * my initial impression is that the overhead caused by proper layouting is not an issue for graphs not larger than few hundred nodes
  * for a final implementation one could use the old method as an automatic fallback in case of larger graphs or allow selecting the method via a config entry
  * dask already has graphviz as an optional dependency and it might make sense to use this instead to compute the node coordinates (though it's a heavier dependency, whereas Grandalf is self contained, pure Python)
[...]

PR #3486, opened 2020-02-16
Improved visuals for task graph

MichaelSchreier:
This implements the _visual_ changes shown in #3467. The algorithm for placing the nodes has __not__ been changed.

To address the performance concerns raised by @mrocklin the graph now uses bokeh's WebGL backend which effortlessly pushes 10k nodes on my integrated GPU. 
All elements in the graph have been chosen so that they play along nicely with WebGL.

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -27,6 +27,7 @@ from bokeh.models import (
     GroupFilter,
     CDSView,
 )
+from bokeh.models.markers import Triangle
 from bokeh.models.widgets import DataTable, TableColumn
 from bokeh.plotting import figure
 from bokeh.palettes import Viridis11
@@ -1126,7 +1127,18 @@ class TaskGraph(DashboardComponent):
         self.node_source = ColumnDataSource(
             {"x": [], "y": [], "name": [], "state": [], "visible": [], "key": []}
         )
-        self.edge_source = ColumnDataSource({"x": [], "y": [], "visible": []})
+
+        self.edge_source = ColumnDataSource({
+            "x0": [],
+            "x1": [],
+            "cx0": [],
+            "cx1": [],
+            "cy0": [],
+            "cy1": [],
+            "y0": [],
+            "y1": [],
+            "visible": []
+        })
 
         node_view = CDSView(
             source=self.node_source,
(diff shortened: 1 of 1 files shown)


mrocklin:
This is very cool to see @MichaelSchreier .  Thanks!

I tried master and this PR with the following code

```bash
export OPENBLAS_NUM_THREADS=1
export OMP_NUM_THREADS=1
export MKL_NUM_THREADS=1
```

```python
client = Client(n_workers=2, threads_per_worker=1)
import dask.array as da
x = da.random.random((10000, 10000), chunks=(1000, 1000))
y = x.dot(x.T - 1) - x.mean(axis=0); y = y.std().persist()
```

I found that, my browser performance was pegged at 100% during the computation, while on master it was closer to 40-50%.  (this computation has something like 1000 nodes total).  However, I'm also on Linux, which doesn't have the best in-browser hardware acceleration, so the benefits of WebGL may be somewhat lost in my case, I'm not sure.  It would be good if other people could give this a shot.  cc @jacobtomlinson , who I think might be interested.

mrocklin:
The CI failure is just linting.  You might want to read through this: https://docs.dask.org/en/latest/develop.html#code-formatting

MichaelSchreier:
I did some quick test on another machine (again on Windows) and the performance issues seem to be browser related. 

In Firefox I can confirm your findings that it puts the browser under heavy load. Chromium based browsers on the other hand work perfectly fine.

I'll check what options there are to remedy the performance under Firefox & friends.

jacobtomlinson:
This PR looks awesome!

I has a quick test in Firefox and Chrome on macOS. Firefox seems a little laggy but not awful compared to Chrome.

---

_Side note: I used the new [GitHub cli](https://cli.github.com/) to check out this PR, what an awesome experience 🎉._ 

`gh pr checkout 3486`

mrocklin:
> I'll check what options there are to remedy the performance under Firefox & friends.

Thanks @MichaelSchreier .  I'm curious to hear what you find.

fjetter:
On safari this doesn't work out at all, unfortunately. It draws the edges as it is supposed to be but the nodes are not shown. Also, we can see a drag of the drawing towards the lower right corner.

Firefox
image

Safari
image

That has most likely nothing to do with the changes of this PR but I have no clue where / how to report this (bokeh?).

MichaelSchreier:
I've now spend quite a bit of time profiling bokeh to see how to one could improve the performance of the task graph in general. 

__The TLDR of it is that we probably have to wait for bokeh to improve their WebGL implementation before this PR can be pursued any further.__

The longer version follows below.

- Regardless of with or without WebGL, it turns out the performance bottleneck isn't drawing the nodes of the graph, but rather managing and drawing it's edges (straight or Bezier). With every update the javascript code spends a significant amount of time just to check and set the properties of each individual edge.
image
There is currently some work being done on this specific system (https://github.com/bokeh/bokeh/pull/9615) but I'm not sure it will address the bad performance we observe.
- The workaround for now is to just use a single line glyph (separating individual segments with NaNs) to reduce the number of times the relevant code section is called to one per update.
- At any rate, the workaround requires calculating and interpolating the Bezier curves in python rather than in javascript and the management of added and removed edges becomes a bit more involved - I only did a quick test using straight lines without updating them to confirm that this fix does indeed work.
- Unfortunately due to https://github.com/bokeh/bokeh/issues/8489 this currently doesn't work under WebGL.
- I've also tested the current RC of bokeh 2.0, but didn't find any measurable improvements.
- Regardless of the above I managed to bring the performance of the new layout code to approximately 10% of each other by removing the number of elements to draw (just a single input arrow head per node) and some quick performance wins in the bokeh code (uneccessary use of regex  when converting between HEX and RGB) - but of course this doesn't solve the Safari issue discussed above (https://github.com/dask/distributed/pull/3486#issuecomment-590231552).

I suggest this topic should be revisited in half a year or so to see if there has been any progress on the bokeh side.

Issue #3488, opened 2020-02-17
distributed.comm.core.CommClosedError loop failing to Terminate worker pods


scottyhq:
We recently encountered an issue on binderhub where a dask pod failed to terminate, resulting in a node running for hours:

```
[ec2-user@ip-192-168-60-131 ~]$ kubectl get pods -A
NAMESPACE        NAME                                                            READY   STATUS      RESTARTS   AGE
binder-staging   autohttps-77465ddf6f-8mgcn                                      2/2     Running     0          9d
binder-staging   binder-58b4fbccd5-7r2qz                                         1/1     Running     0          9d
binder-staging   binder-staging-dind-fppcs                                       1/1     Running     0          9d
binder-staging   binder-staging-image-cleaner-lmg2n                              2/2     Running     0          9d
binder-staging   binder-staging-kube-lego-84c9c589cf-zwjzc                       1/1     Running     0          9d
binder-staging   binder-staging-nginx-ingress-controller-84464555db-psrdm        1/1     Running     0          9d
binder-staging   binder-staging-nginx-ingress-controller-84464555db-st8np        1/1     Running     0          9d
binder-staging   binder-staging-nginx-ingress-default-backend-6cd8b44c86-scn29   1/1     Running     0          9d
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-a2v8wz        0/1     Completed   0          3h24m
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-a6sfbh        1/1     Running     0          3h24m
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-a8s7x5        0/1     Completed   0          3h24m
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-a9csxq        0/1     Completed   0          3h24m
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-ajsm8h        0/1     Completed   0          3h24m
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-aldrmw        0/1     Completed   0          3h24m
binder-staging   dask-cgentemann-osm2020tutorial-9fm90iqn-d58bc39d-avv8qw        0/1     Completed   0          3h24m
binder-staging   hub-57b965856b-9nfc8                                            1/1     Running     3          9d
[...]

PR #3493, opened 2020-02-18
Stop keep alives when worker reconnecting to the scheduler

jacobtomlinson:
Related to #3488.

I don't think this fixes the problem,  but I think it should address the `CommClosedError` exceptions which are seen in the log.

When a worker registers with the scheduler it starts sending keepalive messages via a periodic callback. However that never seems to stop if the connection is broken. If the connection hangs for a long period of time the worker still attempts to send the keepalive messages.

This PR moves the definitions of the `BatchedSend` and `keep-alive` callback to the `__init__` and stops the callback during the reconnect. This is already being done for the heartbeat.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -382,7 +382,7 @@ class Worker(ServerNode):
         self.executed_count = 0
         self.long_running = set()
 
-        self.batched_stream = None
+        self.batched_stream = BatchedSend(interval="2ms", loop=self.loop)
         self.recent_messages_log = deque(
             maxlen=dask.config.get("distributed.comm.recent-messages-log-length")
         )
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
I'm not entirely sure what I've broken here. Assistance would be appreciated.

mrocklin:
I'll try to take a look later today

mrocklin:
If you run tests with `-s` you'll see errors in the logs that look like the following

```
>       self.batched_stream = BatchedSend(interval="2ms", loop=self.loop)
E       AttributeError: 'Worker' object has no attribute 'loop'
```

I've pushed a patch moving the construction a bit lower, but on my machine there are still some failures.  It might make sense to put this construction back in `start`` if there isn't a big reason to move it out.

jacobtomlinson:
Thanks for pushing to this @mrocklin. The CI seems happy here.

PR #3496, opened 2020-02-18
Add backoff to client connect attempts.

### LINKED ISSUE #3487: Attempts to connet to a non-running server spam it with connect attempts
This error does not make sense, on the face of it – either you time out, implying that there was no answer, or the connection was rejected, implying that the server told you there's no listener.
```
OSError: Timed out trying to connect to 'tcp://127.0.0.1:60590' after 10 s: in <distributed.comm.tcp.TCPConnector object at 0x7fce383a1ef0>: ConnectionRefusedError: [Errno 111] Connection refused
```
What actually happens is that `distributed` spams the server with connect requests for ten seconds, all of which get rejected. This is not a good idea. At all. Please either use a reasonable back-off or simply believe the server when it says ECONNREFUSED.

smurfix:
Closes #3487.

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -202,6 +202,9 @@ async def connect(addr, timeout=None, deserialize=True, connection_args=None):
         )
         raise IOError(msg)
 
+    # don't spam the server with connect attempts
+    backoff = min(0.1, timeout/20 if timeout else 9999)
+
     # This starts a thread
     while True:
         try:
@@ -221,8 +224,9 @@ async def connect(addr, timeout=None, deserialize=True, connection_args=None):
         except EnvironmentError as e:
             error = str(e)
             if time() < deadline:
-                await asyncio.sleep(0.01)
-                logger.debug("sleeping on connect")
+                logger.debug("delaying before connect retry")
+                await asyncio.sleep(backoff)
+                backoff *= 1.5
             else:
                 _raise(error)
         else:
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -202,8 +202,9 @@ async def connect(addr, timeout=None, deserialize=True, connection_args=None):
         )
         raise IOError(msg)
 
-    # don't spam the server with connect attempts
-    backoff = min(0.1, timeout/20 if timeout else 9999)
+    backoff = 0.01
+    if timeout and timeout / 20 < backoff:
+        backoff = timeout / 20
 
     # This starts a thread
     while True:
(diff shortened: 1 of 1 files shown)


mrocklin:
@jacobtomlinson could I ask you to take a look at this?  I think that we ended up doing this repeated connection bit because of new Kubernetes addresses arriving.

mrocklin:
Ping @jacobtomlinson in case you're back on work

distributed/comm/core.py
```diff
@@ -202,6 +202,9 @@ def _raise(error):
         )
         raise IOError(msg)
 
+    # don't spam the server with connect attempts
+    backoff = min(0.1, timeout/20 if timeout else 9999)
```
TomAugspurger: In the meantime @smurfix, this is causing a lint failure: https://travis-ci.org/dask/distributed/jobs/652056913#L496.

TomAugspurger: Previously, the first attempt would occur 0.01s after the previous attempt. Should the first element in the min be 0.01?

smurfix: I'll fix the lint failure.

Yes, the previous interval was 0.01; the question is, how fast would the service (re)start? I assume that it will take a non-negligible fraction of second, thus retrying almost-immediately is not going to improve things.

TomAugspurger: What scenarios result in an EnvironmentError? Could something like a temporary network hiccup cause that? If so, I'd think retrying almost immediately is appropriate.

TomAugspurger:
@smurfix can you respond to https://github.com/dask/distributed/pull/3496#discussion_r387298026?

distributed/comm/core.py
```diff
@@ -221,8 +224,9 @@ def _raise(error):
         except EnvironmentError as e:
             error = str(e)
             if time() < deadline:
-                await asyncio.sleep(0.01)
-                logger.debug("sleeping on connect")
+                logger.debug("delaying before connect retry")
+                await asyncio.sleep(backoff)
+                backoff *= 1.5
```
jacobtomlinson: I would be tempted to change this to `2` for a more exponential backoff.

```suggestion
                backoff *= 2
```

distributed/comm/core.py
mrocklin: I've gone ahead and set the minimum to 0.01

distributed/comm/core.py
mrocklin: I suggest that we keep with 1.5.  We're not in a huge rush to grow the timeout.

mrocklin:
It as faster to just make changes than wait here.  I've pushed to @smurfix 's branch

I've also instituted a maximum delay of 1s.

Merging on passed tests.

mrocklin:
This is merged in .  Thank you @smurfix .  

PR #3494, opened 2020-02-18
Msgpack 1.0 compatibility

jrbourbeau:
In msgpack 1.0 the default value for `strict_map_key` was changed from `False` to `True` (see https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10). This causes failures for us as we use map keys which are not `bytes` or `str`. This PR updates our use of `msgpack.loads` to include `strict_map_key=False`

Closes #3491

PR #3497, opened 2020-02-18
Pinned Msgpack in requirements.txt to 0.6.*

aprowe:
Msgpack 1.0.0 was released and caused breaking changes in our code base. We had to pin msgpack to 0.6.2 and noticed that dask does not specify the version of msgpack.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 click >= 6.6
 cloudpickle >= 0.2.2
 dask >= 2.9.0
-msgpack
+msgpack ~= 0.6.2
 psutil >= 5.0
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Thanks for the PR @aprowe! We just added msgpack 1.0 compatibility in over in #3494

TomAugspurger:
I don't think we should pin this here. We've fixed the incompatibility on master, and pinning it now won't do any good for currently released versions of distributed.

jrbourbeau:
Agreed. Thank you for the PR @aprowe, but I'm going to close it for the reasons @TomAugspurger  mentioned

PR #3499, opened 2020-02-19
Suppress cuML `ImportError`

jakirkham:
If cuML is present, but `cuml.comm` does not yet exist, make sure to suppress the `ImportError`. After all there is nothing to do here in this case and we don't want to raise unnecessary errors.

cc @dantegd

### DIFF
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -1,4 +1,3 @@
-from contextlib import suppress
 from functools import partial
 from distutils.version import LooseVersion
 
@@ -100,5 +99,5 @@ def _register_cudf():
 @dask_serialize.register_lazy("cuml")
 @dask_deserialize.register_lazy("cuml")
 def _register_cuml():
-    with suppress(ImportError):
+    with ignoring(ImportError):
         from cuml.comm import serialize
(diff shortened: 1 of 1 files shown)


distributed/protocol/__init__.py
```diff
@@ -99,4 +100,5 @@ def _register_cudf():
 @dask_serialize.register_lazy("cuml")
 @dask_deserialize.register_lazy("cuml")
 def _register_cuml():
-    from cuml.comm import serialize
+    with suppress(ImportError):
```
quasiben: I think in dask this is typically written as 

```python
with ignoring(ImportError):
    ...
```

jakirkham: Thanks! Fixed 😄

It would be interesting to know why that is. I would assume it is some holdover from when Python 2 was still supported. If so, would it make sense to move to the standard library equivalent? 😉

jrbourbeau: > would it make sense to move to the standard library equivalent?

Yeah! xref https://github.com/dask/distributed/issues/3500

jakirkham:
Merging to fix an issue we saw in cuML's CI addressed by this.

Issue #3452, opened 2020-02-06
Make Adaptive keyword args configurable?


willirath:
Currently, `kwargs` of `deploy.Adaptive`
https://github.com/dask/distributed/blob/105d040fb9cabfa3581dcdce73165c77cb22de9a/distributed/deploy/adaptive.py#L12 do not seem to be configurable. As tuning these parameters is depending on the use case, the platform, etc. (see https://github.com/pangeo-data/pangeo/issues/757), there should be a way of changing these default values via dask config.

mrocklin:
Adding these to the configuration makes sense to me

willirath:
cc: @kathoef

TomAugspurger:
This was fixed by https://github.com/dask/distributed/pull/3509 I think.

PR #3509, opened 2020-02-21
Add config for Adaptive arguments

sublinus:
I would like to add config options for the global task duration and the target-duration on a dask cluster basis to be able to tweak the behaviour of adaptive target for long-running tasks with many different and potentially not known function names.

I'm open to suggestions regarding a different name for the global task duration default.
Should this be documented and where would i document the config options?

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -75,11 +75,11 @@ distributed:
     lost-worker-timeout: 15s  # Interval after which to hard-close a lost worker job
 
   adaptive:
-    interval: 1s # interval between scaling evaluations
-    target-duration: 5s # time an entire graph calculation is desired to take ("1m", "30m")
-    minimum: 0 # minimum number of workers
-    maximum: 100000 # maximum number of workers
-    wait-count: 3 # number of times a worker should be suggested for removal before removing it
+    interval: 1s         # Interval between scaling evaluations
+    target-duration: 5s  # Time an entire graph calculation is desired to take ("1m", "30m")
+    minimum: 0           # Minimum number of workers
+    maximum: 100000      # Maximum number of workers
+    wait-count: 3        # Number of times a worker should be suggested for removal before removing it
 
   comm:
     retry:  # some operations (such as gathering data) are subject to re-tries with the below parameters
@@ -148,4 +148,3 @@ ucx:
   infiniband: null # enable Infiniband
   cuda_copy: null  # enable cuda-copy
   net-devices: null  # define which Infiniband device to use
-
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -78,7 +78,7 @@ distributed:
     interval: 1s         # Interval between scaling evaluations
     target-duration: 5s  # Time an entire graph calculation is desired to take ("1m", "30m")
     minimum: 0           # Minimum number of workers
-    maximum: 100000      # Maximum number of workers
+    maximum: .inf        # Maximum number of workers
     wait-count: 3        # Number of times a worker should be suggested for removal before removing it
 
   comm:
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -74,6 +74,13 @@ distributed:
   deploy:
     lost-worker-timeout: 15s  # Interval after which to hard-close a lost worker job
 
+  adaptive:
+    interval: 1s # interval between scaling evaluations
+    target-duration: 5s # time an entire graph calculation is desired to take ("1m", "30m")
+    minimum: 0 # minimum number of workers
+    maximum: 100000 # maximum number of workers
+    wait-count: 3 # number of times a worker should be suggested for removal before removing it
+
   comm:
     retry:  # some operations (such as gathering data) are subject to re-tries with the below parameters
       count: 0  # the maximum retry attempts. 0 disables re-trying.
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5199,7 +5199,7 @@ class Scheduler(ServerNode):
         if close:
             self.loop.add_callback(self.close)
 
-    def adaptive_target(self, comm=None, target_duration="5s"):
+    def adaptive_target(self, comm=None, target_duration=None):
         """ Desired number of workers based on the current workload
 
         This looks at the current running tasks and memory use, and returns a
@@ -5215,6 +5215,8 @@ class Scheduler(ServerNode):
         --------
         distributed.deploy.Adaptive
         """
+        if target_duration is None:
+            target_duration = dask.config.get("distributed.adaptive.target-duration")
         target_duration = parse_timedelta(target_duration)
 
         # CPU
--- a/distributed/deploy/adaptive.py
+++ b/distributed/deploy/adaptive.py
@@ -1,6 +1,6 @@
 from inspect import isawaitable
 import logging
-import math
+import dask.config
 
 from .adaptive_core import AdaptiveCore
 from ..utils import log_errors, parse_timedelta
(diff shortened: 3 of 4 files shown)


mrocklin:
I recommend that we have a config namespace like the following:

```yaml
distributed:
  adaptive:
    interval:
    wait-count:
    target-duration:
    minimum:
    maximum:
```

I got these names by looking at the current keyword arguments to the `Adaptive` class in distributed/deploy/adaptive.py .  I think that separating them away from scheduler into their own category should help with discoverability.  I think that it's best to avoid terms like `default-` if possible for brevity (and because these are all defaults).  I think that we should probably include all of them if we can for consistency, rather than just one or two.

sublinus:
I like the change of the namespace. Would that then currently be under the `deploy` section?

mrocklin:
Sure, it could go under deploy.  I think I probably have a slight preference for it to not be though, and just have it under distributed.  I don't think that there is likely to be another adaptive thing that we'll conflict with here.

distributed/scheduler.py
```diff
@@ -796,6 +796,14 @@ def __init__(self, name):
...
                 ]
             )
+        elif "global-duration" in dask.config.get(
+            "distributed.scheduler.default-task-durations"
+        ):
+            self.duration_average = parse_timedelta(
+                dask.config.get("distributed.scheduler.default-task-durations", {}).get(
+                    "global-duration"
+                )
+            )
```
mrocklin: I suspect that this was held over from before?

distributed/deploy/cluster.py
```diff
@@ -122,7 +123,10 @@ def adapt(self, Adaptive=Adaptive, **kwargs) -> Adaptive:
         with ignoring(AttributeError):
             self._adaptive.stop()
         if not hasattr(self, "_adaptive_options"):
-            self._adaptive_options = {}
+            self._adaptive_options = dask.config.get("distributed.adaptive")
+            self._adaptive_options = {
+                k.replace("-", "_"): v for k, v in self._adaptive_options.items()
+            }
```
mrocklin: I think that it would be best to apply these configuration defaults at the broadest possible point.  If someone calls `Adaptive` in some other way then we want to make sure that these configuration values are also applied, not only if they're called with the Cluster.adapt method.

I recommend setting the default values in the `Adaptive.__init__` constructor to None, and then doing a bunch of checks like the following:

```python
if interval is None:
    interval = dask.config.get("distributed.adaptive.interval")
...
```

This is similar to how other parts of the code base handle configuration options.

distributed/dashboard/scheduler.py
```diff
@@ -269,7 +271,7 @@ def collect(self):
         yield GaugeMetricFamily(
             "dask_scheduler_desired_workers",
             "Number of workers scheduler needs for task graph.",
-            value=self.server.adaptive_target(),
+            value=self.server.adaptive_target(adaptive_target_duration),
```
mrocklin: Rather than get the configuration value every time this method is used, lets get the configuration value within the function itself.  That is likely to affect things more consistently, and keep things a bit more maintainable so that we don't need to always remember to do this logic each time we call this function.

distributed/scheduler.py
sublinus: Yes, the default-task-durations dict already exists in the config. I don't know if we should change this only in order to add a possibility for configuring all function durations.

distributed/deploy/cluster.py
sublinus: That is sensible. Thanks for the implementation hint.

distributed/scheduler.py
mrocklin: Are you running into a situation where you need this?  If not, I recommend that we remove it for now.

sublinus:
This PR will solve issue #3452

distributed/scheduler.py
sublinus: Yes, unfortunately we have an issue as described in #3516 and don't know all function names beforehand as there is also task fusing happening. That's why we would like to be able for individual clusters to change the general default task duration.

sublinus: Can we keep it in the PR in that case @mrocklin ?

distributed/distributed.yaml
```diff
@@ -74,6 +74,13 @@ distributed:
   deploy:
     lost-worker-timeout: 15s  # Interval after which to hard-close a lost worker job
 
+  adaptive:
```
TomAugspurger: @mrocklin

distributed/distributed.yaml
```diff
@@ -74,6 +74,13 @@ distributed:
...
+    interval: 1s # interval between scaling evaluations
+    target-duration: 5s # time an entire graph calculation is desired to take ("1m", "30m")
+    minimum: 0 # minimum number of workers
+    maximum: 100000 # maximum number of workers
+    wait-count: 3 # number of times a worker should be suggested for removal before removing it
```
TomAugspurger: Style nitpicks:

```suggestion
    interval: 1s         # Interval between scaling evaluations
    target-duration: 5s  # Time an entire graph calculation is desired to take ("1m", "30m")
    minimum: 0           # Minimum number of workers
    maximum: 100000      # Maximum number of workers
    wait-count: 3        # Number of times a worker should be suggested for removal before removing it
```

Can the maximum workers be changed to `inf`?

distributed/scheduler.py
TomAugspurger: I'm having trouble understanding this change. Is `global-duration` an existing key in scheduler.default-task-durations?

This also seems independent of the changes to Adaptive, right?

TomAugspurger: I suppose the interaction is through Adaptive using the expected duration to make scaling decisiosn.

distributed/distributed.yaml
mrocklin: This config bit seems fine to me?

TomAugspurger: Whoops, thought I deleted that, sorry :)

mrocklin: For as long as you're handling everything I don't think you get to apologize :)

Thanks Tom.

distributed/scheduler.py
sublinus: Yeah, it is related through Adaptive using the expected durations in the occupancy.
Currently the key `global-duration` does not exist in the default-task-durations. As the keys are the names of the executed functions this also should never collide by accident as `global-duration` is not a valid function name.
This key is supposed to be a wildcard.

distributed/distributed.yaml
sublinus: Thanks for the change in the config. Yes, i will change that back to inf again.

sublinus:
> Can you add tests for Adaptive getting the default values from the config? I think you can just create the Adaptive class directly inside a `dask.config.set` context mananger and check the `.interval`, etc.

I like the suggestion, where would you prefer the tests to be?

TomAugspurger:
Probably test_adaptive.py or test_adaptive_core.py, though I'm not sure what the difference between the two is.

distributed/scheduler.py
TomAugspurger: OK, I'm better understanding the motivation for this change, thanks. I'm still a bit unsure about it since it's such a blunt too for the job. I can imagine several alternatives (users running optimize locally to discover keys, better control over names through optimization, task-prefix control) which might be better / conflict with the changes here.

If we do go forward with this, I'm not sure about the `global-duration` though, since duration is already in the parent key. I'd recommend putting it in a separate key like `default-duration: null`.

Regardless, you might want to consider making these changes in a new pull request so that the uncontroversial changes in Adaptive don't get held up.l

sublinus: Yes, you are right this change should go in its own PR. Thanks!

sublinus:
Sorry, your comment is unclear to me. Where should I add docs?

TomAugspurger:
Updating the docstring of Adaptive to note that the values can come from the config.

mrocklin:
This looks great.  Thank you for your perseverance on this PR @sublinus , and my apologies for the delay in review.  I've fixed a merge conflict and pushed to your branch (I hope that you don't mind).  I'll merge this in after CI runs.

PR #3510, opened 2020-02-21
Remove `import ucp` from the top of `ucx.py`

pentschev:
This is needed to ensure Dask configurations will be propagated
to UCX upon importing. Since `ucx.py` is imported upon
`import distributed`, Dask configurations passed to
`Nanny(..., config=ucx_config)` won't be read by UCX since it has
already been loaded.

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -5,8 +5,6 @@ See :ref:`communications` for more.
 
 .. _UCX: https://github.com/openucx/ucx
 """
-import ucp
-
 import logging
 
 import dask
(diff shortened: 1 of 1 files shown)


pentschev:
@mrocklin I realize you added that `import ucp` back in https://github.com/dask/distributed/commit/43b2ed77218861c34dd2c98d88ec5153bd35f5c4 for tests to be faster. However, this breaks when we try using Dask configurations when forking a process, to fix that issue the change in https://github.com/dask/distributed/blob/master/distributed/comm/ucx.py#L34-L42 was introduced, where we ensure we import ucp and immediately initialize it, otherwise it will be already initialized at this point and we can't configure it anymore.

cc @quasiben

pentschev:
Not sure, but I think the tests are timing out and seems unrelated to this PR.

quasiben:
The test failure here was a windows test on `test_thread_time` .  We can safely ignore this.

quasiben:
Apologies, forgot to merge

Issue #3512, opened 2020-02-22
Windows timing is off


mrocklin:
We've noticed some intermittent failures in a few tests when we measure time in Windows.  This is exemplified in [this test failure](https://github.com/dask/distributed/pull/3510/checks?check_run_id=461073401)

```python
______________________________ test_thread_time _______________________________

    def test_thread_time():
        start = metrics.thread_time()
        run_for(0.05)
        dt = metrics.thread_time() - start
>       assert 0.03 <= dt <= 0.2
E       assert 0.03 <= 0.015625
```

For background, on windows `time.time` returns values in integer seconds, which is problematic.  To work around this, we've built a combination `time.time` + `default_timer` solution that is both long-term stable (default_timer drifts) and precise.  This has worked for a long time, but apparently has recently decided to stop working.  Maybe this is changes upstream, maybe it's the switch from forkserver to spawn, maybe it's the change from appveyor to Github actions.  

We have some options here:

1.  Ignore timing imprecision on Windows.  It's not terrible, and is often only off by a 10ms-100ms.  To be clear, this affects things like timeouts, reported "last seen at" values, and so on.  That being said, we should be aware that there are *a lot* of silent windows users out there, and so we're probably biased towards not caring about this as much as we should.
2.  Figure out what's going on, and try to fix it

cc @pitrou , in case he has quick thoughts on what might be going wrong.

mrocklin:
Hrm, on Windows time.time now seems to return full precision floating point values.  Checking to see if maybe this was changed in recent versions of Python.

mrocklin:
From the [Python 3.6 docs](https://docs.python.org/3.6/library/time.html):

> Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second. While this function normally returns non-decreasing values, it can return a lower value than a previous call if the system clock has been set back between the two calls.

So it seems like there might still be reason to keep this code around.  I'm curious what systems there are today that don't provide precise time estimates.

pitrou:
15.6 ms is _exactly_ the default timer resolution on Windows. So I think the test is simply too reliant on precise timing. Try replacing it with something coarser like:
```python
    def test_thread_time():
        start = metrics.thread_time()
        run_for(0.1)
        dt = metrics.thread_time() - start
        assert 0.03 <= dt <= 0.3```
```

pitrou:
That said, I wouldn't be surprised if the Python doc snippet you're quoting dates back to the 90s...

[...]

PR #3513, opened 2020-02-22
Remove old windows time keeping

mrocklin:
I'm curious to see what CI says on this

See #3512


pitrou:
CI probably won't bark, but you're giving up some timing precision. The default timer precision on Windows is 15.6 ms.

Issue #3506, opened 2020-02-20
Accelerated networking options for the scheduler


quasiben:
With UCX-Py and additions to the comms code in dask, we've enable workers to communicate with accelerated hardware.  This has largely been done by passing flags/env-vars with [dask-cuda](https://github.com/rapidsai/dask-cuda/blob/474108a3b597d27339a019a200bd824f6f85ef50/dask_cuda/dask_cuda_worker.py#L178-L182) to the worker setup.

For example, one can start a dask-cuda-worker with following flags:

> dask-cuda-worker ucx://scheduler --enable-nvlink --enable-infiniband

I don't think any such options exist for the scheduler today.  While the scheduler should largely only be touching host memory and serialized strings (task graph), nothing restricts the user from inserting GPU objects in the task graph.  Furthermore, transport over enabled IB should be faster than non-enabled IB

By allowing such options directly in the distributed code base other cluster managers (Dask-K8s, Dask-SSH, Dask-jobqueue) can easily expose the options as well.

mrocklin:
Ideally we can get by with options like `--protocol ucx://` and configuration settings for UCX itself (which is presumably separate).

To me, adding UCX specific keywords into the dask-scheduler CLI is likely out of scope.

pentschev:
> Ideally we can get by with options like --protocol ucx:// and configuration settings for UCX itself (which is presumably separate).

Not really, if you're creating the scheduler from scratch, it's up to you to define what kind of transports you want enabled, otherwise you'll just use UCX's defaults and have no control over it.

mrocklin:
I get that, and I agree that there needs to be some way to configure the use of UCX.  However I am vetoing putting UCX-specific configuration options into Dask's top level CLI API.  We can't put every dependency's options into the CLI.  It would become thousands of options long.  We'll need to find some way to make these things configurable without putting all of them into the core Dask interface.

I think that maybe it makes sense to have a ucx config file?

```yaml
ucx:
  tls: all
  net-devices:...
```

If memory serves we had this conversation when @pentschev first added `--enable-nvlink` to the dask-cuda-worker API.

quasiben:
Relying on the the config seems like a good option.  I think the cluster managers could expose these options (should they want to) with reasonable defaults.  They should be able to do something like the following before the scheduler is created, right ?

something like:

```python
dask.config.set('ucx.tls': '...')
dask.config.set('ucx.net-devices': '...')
...
```

[...]

PR #3515, opened 2020-02-23
RMM/UCX Config Flags

quasiben:
closes https://github.com/dask/distributed/issues/3506

PR Is still very much a WIP but will allow users to write simplified CLI args for configuring UCX/RMM when starting scheduler/worker

> DASK_UCX__NVLINK=True dask-scheduler --protocol ucx --port 13778
> DASK_RMM__POOL_SIZE=1G dask-scheduler...

cc @pentschev


distributed/comm/ucx.py
```diff
@@ -39,6 +39,29 @@ def init_once():
     import ucp as _ucp
 
     ucp = _ucp
+    ucx_conf = dask.config.get("ucx")
+
+    enable_infiniband = str(ucx_conf.get("infiniband", None)).lower()
+    enable_nvlink = str(ucx_conf.get("nvlink", None)).lower()
+    tcp_over_ucx = str(ucx_conf.get("tcp_over_ucx", None)).lower()
+    net_devices = ucx_conf.get("net_devices", None)
```
mrocklin: For the boolean values can we use bools rather than strings?

Also, we tend to put default values into a yaml file in the project rather than put default values into code like this.  You might add a `ucx:` section to the distributed/distributed.yaml file.

So I think that ideally we would see things like the following:

```python
enable_infiniband = dask.config.get("ucx.infiniband")
```

Or perhaps instead we avoid the intermediate value entirely and just query the configuration when needed ...

```python
if dask.config.get("ucx.infiniband"):
    ...
```

I think that this is still concise, and will make it easier for future reviewers so that they won't have to refer back to what `enable_nvlink` is, and then refer back to what `ucx_conf` is.

quasiben: My concern about bool/string is that if user writes `'False` we can get into trouble without knowing it.

quasiben: The intermediate values serve two purposes:
1) check if any of the configs are set and, if so, configure TLS
2) configure specific hardware nvlink, ib, etc

quasiben:
Thanks for the comments @mrocklin .   Still in WIP but it's taking better shape

distributed/comm/ucx.py
quasiben: Tried to address much of the comments here in https://github.com/dask/distributed/pull/3515/commits/0755036ee9450a6b794ab6da00c6d6d72a90ec5a

mrocklin: > My concern about bool/string is that if user writes 'False we can get into trouble without knowing it.

Right, but that's something that if we handle we should handle centrally, not just for the UCX configuration bits.

> The intermediate values serve two purposes:

To be clear, my only request here is that we instead inline these as `dask.config.get("ucx.nvlink")` rather than have separate variables `ucx_config` and `enable_nvlink`.  I'm only suggesting a cosmetic change.

distributed/comm/tests/test_ucx.py
```diff
@@ -350,3 +354,102 @@ def test_ucx_deserialize():
...
+
+    def get_ucp_config():
+        import ucp
+
+        return ucp.get_config()
+
+    def get_rmm():
+        import rmm
+
+        return rmm.get_info()
```
mrocklin: Maybe we can replace these functions with just  `rmm.get_info` and `ucp.get_config`?

distributed/comm/ucx.py
```diff
@@ -39,7 +47,41 @@ def init_once():
...
+    # To handle high level flags we grab the full ucx configuration
+    # pop flags from the dictionary and proceed to check flags.  We do this
+    # because UCX/UCP will not nicely process flags it does not know about
+    # ucp.init(options={'foo': 'bar'}) -- `UCXConfigError`
+    ucx_conf = dask.config.get("ucx")
+
+    enable_infiniband = ucx_conf.pop("infiniband", None)
+    enable_nvlink = ucx_conf.pop("nvlink", None)
+    tcp_over_ucx = ucx_conf.pop("tcp-over-ucx", None)
+    net_devices = ucx_conf.pop("net-devices", None)
```
mrocklin: Now that you have included `null` in the yaml file you can safely drop the default values here.  In general we try not to include default values in code, so that the defaults in the yaml file are respected if they are changed in the future.

distributed/comm/ucx.py
```diff
@@ -39,7 +47,41 @@ def init_once():
...
+
+    # if any of the high level flags are set, as long as they are not Null/None,
+    # we assume we should configure basic TLS settings for UCX
+    if any([s is not None for s in (tcp_over_ucx, enable_infiniband, enable_nvlink)]):
```
mrocklin: Can maybe drop is `is not None` bit here and just have

```python
if any([
    dask.config.get("ucx.nvlink"),
    dask.config.get("ucx.infiniband"),
    dask.config.get("ucx.tcp-over-ucx"),
]):
```

distributed/comm/tests/test_ucx.py
quasiben: :). Definitely!

quasiben:
I move the tests to it's own file because we test `test_ucx` in the ucx-py CI:

https://github.com/rapidsai/ucx-py/blob/d301dc552f58bfb1dd31a8e372713c9f58e12c28/ci/gpu/build.sh#L122-L137

These tests require env vars not to be set (i tried playing patching the env var in the tests but was unsuccessful)

distributed/comm/tests/test_ucx_config.py
```diff
@@ -0,0 +1,118 @@
...
+                    assert ucx_config.get("TLS") == "rc,tcp,sockcm,cuda_copy,cuda_ipc"
+                    assert ucx_config.get("SOCKADDR_TLS_PRIORITY") == "sockcm"
+                    assert ucx_config.get("NET_DEVICES") == "mlx5_0:1"
+
+                    # worker is configured with NVLINK/IB
+                    worker_ucx_config = await c.run(ucp.get_config)
+                    ucx_config = worker_ucx_config[a.contact_address]
+                    assert ucx_config.get("TLS") == "rc,tcp,sockcm,cuda_copy,cuda_ipc"
+                    assert ucx_config.get("SOCKADDR_TLS_PRIORITY") == "sockcm"
+                    assert ucx_config.get("NET_DEVICES") == "mlx5_0:1"
```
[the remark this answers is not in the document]
quasiben: No, this test will fail.  The tests here having problems running consecutively -- still trying to understand wha the issue is.  Currently testing on a DGX machine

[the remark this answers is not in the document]
pentschev: I'm not sure you'll be able to easily test this. You'd need to find what's the main network interface somehow and use that. I suggest you skip this test if it's not a DGX or allow it to xfail. This is what we do in dask-cuda as well because we don't have a way to identify the interfaces easily.

distributed/comm/ucx.py
```diff
@@ -61,6 +109,13 @@ def cuda_array(n):
                     "In order to send/recv CUDA arrays, Numba or RMM is required"
                 )
 
+    pool_size_str = dask.config.get("rmm.pool-size")
+    if pool_size_str is not None:
+        pool_size = parse_bytes(pool_size_str)
+        rmm.reinitialize(
+            pool_allocator=True, managed_memory=False, initial_pool_size=pool_size
```
quasiben: @pentschev would you like to see options for `pool_allocator` and `managed_memory` as well ?

quasiben:
@pentschev @mrocklin I am unable to get the tests to run consecutively due to cleanup issues.  @madsbk is working on refactoring the cython layer which I am hoping will resolve shutdown/cleanup issues.   I think this is now ready for another review time permitting

distributed/comm/ucx.py
pentschev: I think for the time being we can skip that.

pentschev:
> @pentschev @mrocklin I am unable to get the tests to run consecutively due to cleanup issues.

What kind of issues are you experiencing? I am working on tests for `dask-cuda-worker` as well and noticed they won't close the connection to the scheduler, preventing the test from finishing. Is that the same issue you're seeing?

quasiben:
The tests finish, but ucx/ucp is still holding onto some state making it impossible to change the initialization.  Calling `ucp.reset()` results in everything blowing

distributed/comm/ucx.py
```diff
@@ -39,6 +43,50 @@ def init_once():
...
+        if dask.config.get("ucx.nvlink"):
+            tls = tls + ",cuda_ipc"
+
+        options = {"TLS": tls, "SOCKADDR_TLS_PRIORITY": tls_priority}
+
+        net_devices = dask.config.get("ucx.net-devices")
+        if net_devices is not None and net_devices != "":
+            options["NET_DEVICES"] = net_devices
+
+        dask.config.set({"ucx": options})
```
mrocklin: Should we be concerned about losing the user's provided options here?

```python
In [1]: import dask                                                             

In [2]: dask.config.set({"a": {1: 1}})                                          
Out[2]: <dask.config.set at 0x7f25d2507610>

In [3]: dask.config.set({"a": {2: 2}})                                          
Out[3]: <dask.config.set at 0x7f25d256d610>

In [4]: dask.config.get("a")                                                    
Out[4]: {2: 2}
```

Maybe we want `dask.config.update` here instead?

distributed/comm/ucx.py
```diff
@@ -39,6 +43,50 @@ def init_once():
...
+
+    # support other clients (dask-cuda) passing ucx configurations directly
+    # when using explicty UCX configurations assume advanced users
+    elif all(
+        [
+            dask.config.get("ucx.TLS", False),
+            dask.config.get("ucx.SOCKADDR_TLS_PRIORITY", False),
```
mrocklin: Should these default values be in the yaml file instead?

distributed/comm/ucx.py
```diff
@@ -39,6 +43,50 @@ def init_once():
...
+        ]
+    ):
+        pass
+    else:
+        dask.config.set({"ucx": {}})
```
mrocklin: This is perhaps something to think about later, but it seems odd to delete user configuration.  

Instead, I wonder if a function like this might produce some new dictionary on demand, and that dictionary gets used in the `ucp.init` call without modifying the dask.config global in-place.

distributed/comm/ucx.py
quasiben: Well, what we want to avoid is passing in options which are valid for ucp.   ucp.init({'foo': 'bar'}) throws a `UCXConfigError`.  So either we want to remove invalid options or passing in a "clean' dictionary to `ucp.init`

distributed/comm/ucx.py
quasiben: Admittedly this may be a conservative opinion, I don't think so.  There are loads of UCX config options and this could easily explode.  On there hand, maybe a minimal set _might_ be ok.   @pentschev thoughts ?

env vars which have been important for us lately would include:
- TLS
- SOCKADDR_TLS_PRIORITY
- MEMTYPE_CACHE
- CUDA_IPC_CACHE
- NET_DEVICES

distributed/comm/ucx.py
quasiben: I went back and forth on this as well.  I think a new dictionary makes sense.  i'd like to do that now

distributed/comm/ucx.py
pentschev: Hmm, the way I was handling defaults is not to have defaults, i.e., the user is forced to select at least one option. The reason for doing that is that users are then aware of what they're selecting until we are confident options work more broadly. IMO, this is the best way to do it now, as users will know they _need_ to specify what they want to use rather than assume UCX will enable everything by default -- maybe one day that will be the case, but that day is still not today, unfortunately.

pentschev: To be more specific, if the user doesn't choose at least TCP over UCX, dask-cuda will error out today.

quasiben:
I refactored the flag processing into a separate function/file. This made the testing significantly easier

distributed/comm/utils.py
```diff
@@ -113,3 +113,52 @@ def ensure_concrete_host(host):
         return get_ipv6()
     else:
         return host
+
+
+def _srub_ucx_config():
```
mrocklin: srub -> scrub

Also, I would prefer keeping this in the ucp file if we can, just to contain this work a bit.

quasiben: sure

quasiben:
@mrocklin thanks for the reviews.  @pentschev do you want to test before merging

distributed/comm/ucx.py
```diff
@@ -328,3 +343,52 @@ def get_local_address_for(self, loc):
...
+    # if any of the high level flags are set, as long as they are not Null/None,
+    # we assume we should configure basic TLS settings for UCX
+    if any(
+        [
+            dask.config.get("ucx.nvlink"),
+            dask.config.get("ucx.infiniband"),
+            dask.config.get("ucx.tcp-over-ucx"),
+            dask.config.get("ucx.net-devices"),
+        ]
+    ):
```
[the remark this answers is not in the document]
quasiben: I'll make those changes. Thanks for the comments @pentschev

quasiben:
Thanks all.  Merging now

mrocklin:
Thanks for this Ben!  I'm quite happy with how it turned out.

Issue #3426, opened 2020-01-30
WorkerPlugin loses name on restart


nsmith-:
If I make a named plugin like:
```python
from distributed import Client, WorkerPlugin, get_worker
client = Client()
client.register_worker_plugin(WorkerPlugin(), 'myplugin')
client.run(lambda: list(get_worker().plugins))
```
then I see the output of the last command has properly-named plugins:
```
{'tcp://127.0.0.1:58761': ['myplugin'],
 'tcp://127.0.0.1:58762': ['myplugin'],
 'tcp://127.0.0.1:58763': ['myplugin'],
 'tcp://127.0.0.1:58764': ['myplugin']}
```
but after issuing a `client.restart()` now the same plugin listing shows:
```
{'tcp://127.0.0.1:58806': ['<distributed.diagnostics.plugin.WorkerPlugin objec-4ecf53b0-b67d-40f5-b2e3-fcb2e76d1979'],
 'tcp://127.0.0.1:58807': ['<distributed.diagnostics.plugin.WorkerPlugin objec-596662ab-f814-481b-a6f4-dc068c65e66c'],
 'tcp://127.0.0.1:58810': ['<distributed.diagnostics.plugin.WorkerPlugin objec-a49d12a3-2f47-4672-b53f-e93299a9290d'],
 'tcp://127.0.0.1:58812': ['<distributed.diagnostics.plugin.WorkerPlugin objec-533dd071-54d4-4452-8763-557d721c7968']}
```

I would prefer if the name stuck.  Also, is there no method to unregister plugins?

P.S. `distributed.__version__` is 2.9.0

mrocklin:
Oh indeed.  It would be good if the name stuck.  I totally agree.  Is this something that you would be interested in investigating?

jrbourbeau:
Thanks for opening up an issue and providing a nice example @nsmith-! Good to see you here : )

Just a heads up, if your plugin has a `name` attribute then things should currently work. For example,

```python
from pprint import pprint
from distributed import Client, WorkerPlugin, get_worker

class MyPlugin(WorkerPlugin):
    name = "my_plugin"

if __name__ == "__main__":

    client = Client()
    client.register_worker_plugin(MyPlugin())
    print("Before restart:")
    pprint(client.run(lambda: list(get_worker().plugins)))
    client.restart()
    print("After restart:")
    pprint(client.run(lambda: list(get_worker().plugins)))
```

outputs

```
Before restart:
[...]

PR #3517, opened 2020-02-24
Assign name to Work Plugin when you register it

jsignell:
Closes: #3426

This is the most straightforward fix I could think of to make sure that the name sticks to the plugin. I'm not sure how bad it is to alter the plugin when you call register.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4021,8 +4021,15 @@ class Client(Node):
         return self.register_worker_plugin(_WorkerSetupPlugin(setup))
 
     async def _register_worker_plugin(self, plugin=None, name=None):
-        if getattr(plugin, "name", None) is None and name is not None:
-            plugin.name = name
+        if name is not None:
+            warnings.warn(
+                "This name= parameter is deprecated. "
+                "Instead set name as an attribute on your plugin class",
+                stacklevel=2,
+            )
+        if hasattr(plugin, "name"):
+            name = plugin.name
+
         responses = await self.scheduler.register_worker_plugin(
             plugin=dumps(plugin), name=name
         )
@@ -4060,7 +4067,7 @@ class Client(Node):
         plugin: WorkerPlugin
             The plugin object to pass to the workers
         name: str, optional
-            A name for the plugin.
+            Deprecated.  Set name as attribute on plugin object instead.
             Registering a plugin with the same name will have no effect.
 
         Examples
@@ -4080,10 +4087,10 @@ class Client(Node):
 
         You can get access to the plugin with the ``get_worker`` function
 
-        >>> client.register_worker_plugin(other_plugin, name='my-plugin')
+        >>> client.register_worker_plugin(other_plugin)
         >>> def f():
         ...    worker = get_worker()
-        ...    plugin = worker.plugins['my-plugin']
+        ...    plugin = worker.plugins[other_plugin.name]
         ...    return plugin.my_state
 
         >>> future = client.run(f)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4021,7 +4021,7 @@ class Client(Node):
         return self.register_worker_plugin(_WorkerSetupPlugin(setup))
 
     async def _register_worker_plugin(self, plugin=None, name=None):
-        if getattr(plugin, "name", None) is None:
+        if getattr(plugin, "name", None) is None and name is not None:
             plugin.name = name
         responses = await self.scheduler.register_worker_plugin(
             plugin=dumps(plugin), name=name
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -3961,6 +3961,8 @@ def register_worker_callbacks(self, setup=None):
         return self.register_worker_plugin(_WorkerSetupPlugin(setup))
 
     async def _register_worker_plugin(self, plugin=None, name=None):
+        if getattr(plugin, "name", None) is None:
+            plugin.name = name
```
mrocklin: Hrm, this seems a little janky.  We could also investigate changing `Scheduler.worker_plugins` from a list to a dict, or storing the plugins themselves as dictionaries like `{"plugin": my_plugin, "name": name}`.  Those probably both have problems (we lose ordering in the first case, and it's not clear what to do when we don't have a name).

mrocklin: Hrm, maybe not though.  I guess if it gets the job done?

quasiben: With the conditional check we don't need the additional `is None`

jsignell: @mrocklin - definitely janky :) but I didn't want to change existing behavior if the plugin has a name, so I think if we want to pass a dict, then we should use `plugin.name` as the default when name isn't provided as a kwarg. Which feels pretty similar to me. 

@quasiben - You don't think the name could be something that evaluates to False (like 0)?

quasiben: Oh! I wasn't thinking about numerical names.  Thanks for the correction

jsignell: So is this an ok solution? Should I go ahead and add/fix tests?

TomAugspurger: This seems OK to me. Changing the container from a list to dict also seems sensible, but probably will be better as a followup.

TomAugspurger:
Sorry, one last question here: are we OK just mutating the user-provided object inplace? That seems a bit unfriendly.

Would it be better to require that WorkerPlugin have a name attribute set before they can be registered?

jsignell:
Maybe this approach is unfriendly enough to be a bad idea and we should pursue the dict idea instead. Or maybe there shouldn't be a name kwarg on the register method at all. It is fundamentally kind of stange to allow two different methods for setting the name.

jsignell:
@TomAugspurger so your proposal is to not have a name kwarg on the register and instead check that the plugin has a name and throw an error if it doesn't?

TomAugspurger:
> @TomAugspurger so your proposal is to not have a name kwarg on the register and instead check that the plugin has a name and throw an error if it doesn't?

Yeah, though I haven't really thought through the implications of it, I was just thinking out loud.

If we think that requiring a `.name` is a good idea, I think we can deprecate the current behavior by allowing the name= keyword still, but warning if it's provided an the plugin doesn't have a name.

nsmith-:
I think requiring a `.name` is better, less confusion having two ways to set it.
Would it make sense to have the base class declare it as a property such that if a user doesn't provide the attribute, there is some fallback? Then a user can even override the property if the want to use multiple instances of the same class as plugins.

jsignell:
> Would it make sense to have the base class declare it as a property such that if a user doesn't provide the attribute, there is some fallback?

Is there a meaningful default name that we could set on the base class or would it just be an attr set to `None` such that the attr could be set on instances or overwritten by inherited classes?

nsmith-:
I cannot think of anything more meaningful than `repr()` in the case that the user has not supplied a name either by attribute or by the argument to register, so I guess setting it to None seems the most simple.

jsignell:
Ok sorry, this slipped off my radar for a bit. I've addressed the comments now.

mrocklin:
> If we think that requiring a .name is a good idea, I think we can deprecate the current behavior by allowing the name= keyword still, but warning if it's provided an the plugin doesn't have a name.

Sorry to jump in here late.  As a reminder, a common use case is to have a user provide a callable function, rather than a class, and have Dask turn that into a plugin on the fly.  In my mind one of the major use cases of naming plugins is to replace old ones.

I think that before making a decision to remove the name parameter we probably need to gather a bit of information about how this gets used and why it was added in the first place.  I honestly don't recall this information, so some sleuthing and reporting back might be good here.

jsignell:
Ok thanks for the context. I'll do some sleuthing.

jsignell:
I am now actually using worker plugins and hopefully I understand them a bit better. 

I looked back at the original PR where worker plugins were added (#2453) and I don't understand why the naming behavior is how it is. Right now if you do:

```python
client.register_worker_plugin(PipInstall(packages=["pandas"])) 
```

and then do 

```python
client.register_worker_plugin(PipInstall(packages=["intake"])) 
```

you'll never get the second set of packages because the plugins have the same name. This is intentional according to the [docs](https://github.com/dask/distributed/blob/4205280f3f6584c59b5bc35af5f1938e328654f9/distributed/client.py#L4189:L4191). 

I'm closing this PR, since I can't reproduce the original issue on master anymore.

PR #3521, opened 2020-02-26
make work stealing callback time configurable

lr4d:


mrocklin:
Did you mean to close this @lr4d ?

If you end up re-opening I recommend using a time-valued string like `100 ms` as a default and using the `dask.utils.parse_timedelta` function.  This is how we usually handle time-like values in configuration.

PR #3523, opened 2020-02-27
make work stealing callback time configurable

lr4d:
Makes it easier to benchmark #3069

### DIFF
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -6,7 +6,7 @@ from time import time
 import dask
 from .core import CommClosedError
 from .diagnostics.plugin import SchedulerPlugin
-from .utils import log_errors, PeriodicCallback
+from .utils import log_errors, parse_timedelta, PeriodicCallback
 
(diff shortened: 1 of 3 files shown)


distributed/stealing.py
```diff
@@ -40,8 +40,14 @@ def __init__(self, scheduler):
         for worker in scheduler.workers:
             self.add_worker(worker=worker)
 
+        callback_time = parse_timedelta(
+            dask.config.get("distributed.scheduler.work-stealing-interval"),
+            default="ms",
```
TomAugspurger: I think you'll need something like "100ms" here.

lr4d: AFAI see, `default` represents the default time unit (`def parse_timedelta(s, default="seconds")`)

TomAugspurger: Oh, sorry I misunderstood.

TomAugspurger:
Thanks!

PR #3527, opened 2020-02-28
Add dashboard grid layout

jacobtomlinson:
During one of the working sessions at the Dask workshop in DC I threw together a quick prototype of a configurable grid layout in the Dask dashboard.

Dash dashboard grid layout

This demo uses [gridstack.js](https://gridstackjs.com/) to enable positioning and resizing of plots. Clicking the "plus" button randomly adds a plot from the `/individual-plots.json` endpoint which is the same list that Jupyter Lab uses. Each plot is an iframe which displays one of the individual plot pages which were originally created for the Jupyter Lab integration.

Still to do:
- [ ] Add close button to each plot
- [ ] Add plot picker to plus button instead of adding a random plot
- [ ] Persist the layout in a cookie between page refreshes
- [ ] CSS tweaks and fixes to make the layout clearer

We could also potentially replace the whole dashboard with this. Currently the dashboard is a collection of fixed Bokeh pages, each of which are made from one or more plots.

Instead we could replace each page with the grid layout with the default configuration set to match the current layout. Users could then customise the layout of each page. We could also add a way to add new pages to the menu bar at the top to enable custom pages to be created.

### DIFF
--- a/distributed/dashboard/static/css/grid.css
+++ b/distributed/dashboard/static/css/grid.css
@@ -1,0 +1,69 @@
+body {
+  /* background-color: #eee; */
+}
+
+nav img {
+  height: 28px;
+  padding: 5px 15px;
+}
+
+.grid-stack {
+  height: 100%;
+  width: 100%;
+}
+
+.grid-stack-item {
+  padding: 10px;
+  background-color: white;
+}
+
+.grid-stack-item iframe {
+  width: 100%;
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
cc @mmccarty

jsignell:
This is very cool! I'm excited to see where it goes :)

PR #3529, opened 2020-02-28
Update heartbeat CommClosedError error handling

jrbourbeau:
This moves an `except CommClosedError` inside the worker heartbeat to occur before a recently added `except IOError`. The `IOError` catching was added to catch timeout errors, but since `CommClosedError` is a subclass of `IOError` we're now catching more exceptions in the `IOError` block than originally intended. 

Additionally, I noticed that workers aren't closed when a heartbeat fails and `reconnect=False`. This PR also adds a `reconnect` check and closes the worker accordingly.

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -31,7 +31,7 @@ from distributed import (
     wait,
 )
 from distributed.compatibility import WINDOWS
-from distributed.core import rpc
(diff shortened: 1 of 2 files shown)


quasiben:
thanks @jrbourbeau merging in

quasiben:
Looks good to me.  Thanks for adding test coverage as well

PR #3530, opened 2020-02-28
Pin numpydoc to avoid doubly-escaped `*`

gforsyth:
Similar to dask/dask#5961, recent changes to `numpydoc` lead to function
signatures displayed using `autosummary` to have doubly-escaped `*`s.

This commit swaps out `numpydoc` in favor of `napoleon` -- there are a
couple of benefits:
* no more double-escaped function signatures
* one less dependency for docs (napoleon is shipped with sphinx)

I've added a make target to automatically generate API docs using
`sphinx-apidoc` but this isn't required for `autosummary` compatibility.
It can always be added as a requirement of the `html` target at a later
time.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -39,6 +39,7 @@ extensions = [
 
 numpydoc_show_class_members = False
 
+
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
 
@@ -382,7 +383,7 @@ extlinks = {
 # and the Numpy documentation.
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
-    "numpy": ("http://docs.scipy.org/doc/numpy", None),
+    "numpy": ("https://docs.scipy.org/doc/numpy", None),
 }
 
 # Redirects
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,8 +1,9 @@
+# We pin numpydoc to avoid doubly-escaped *args and **kwargs in rendered docs
+numpydoc==0.8.0
 tornado
 toolz
 cloudpickle
 dask
-numpydoc
 sphinx
 dask_sphinx_theme
 sphinx-click
--- a/docs/source/resources.rst
+++ b/docs/source/resources.rst
@@ -96,6 +96,7 @@ delayed objects.  You can pass a dictionary mapping keys of the collection to
 resource requirements during compute or persist calls.
 
 .. code-block:: python
+
     from dask import core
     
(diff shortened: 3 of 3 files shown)


gforsyth:
@jrbourbeau this may be a better solution than the one I pushed up to dask/dask, but we can also pin `numpydoc` here, too, if that's preferred.

gforsyth:
> Do you know of other difference between numpydoc and napoleon?

I don't -- just that they're roughly equivalent.  `numpydoc` supports adding `::plot` directives which I think is then also used by extensions that autogenerate example galleries, but I don't think that's used in the dask docs anywhere.

I also don't have a preference on styling, although it should probably match across all dask projects -- if folks are ok with the general styling change I can swap this out in the other repos, too.

jsignell:
This change seems reasonable to me. No strong feelings on the style change.

jrbourbeau:
Sorry for the delayed reply here. I think I'd rather stick with numpydoc for the time being. Not all of our docstrings follow the official numpydoc style for their `Returns` sections and so both numpydoc and napoleon outputs aren't properly formatted. However, napoleon makes things bulleted which looks a little more off to me than the numpydoc styling. For example:

Napoleon:

Screen Shot 2020-03-13 at 5 04 47 PM

Numpydoc:

Screen Shot 2020-03-13 at 5 05 01 PM

We could also reformat our docstrings, which could be done in a future PR

gforsyth:
Good with me, I don't have strong opinions on this either way.  I'll change this PR to pin `numpydoc` instead -- also followed up in https://github.com/numpy/numpydoc/pull/221 which should fix the issue in `numpydoc` with the double-escapes.

gforsyth:
Hey @jrbourbeau -- updated this to pin `numpydoc` instead and fixed a few small nits.

Small changes to avoid a sphinx timeout looking for an intersphinx inventory, and a fixed a misformatted code-block

docs/source/conf.py
```diff
@@ -382,7 +383,7 @@
 # and the Numpy documentation.
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
-    "numpy": ("http://docs.scipy.org/doc/numpy", None),
+    "numpy": ("https://docs.scipy.org/doc/numpy", None),
```
jrbourbeau: Nice 😎

jsignell:
Good catch @jrbourbeau! Do we need a follow on issue to coerce all the docstrings to numpydoc style?

jrbourbeau:
I opened up #3578 for updating the docstring style

PR #3536, opened 2020-03-01
Support using other serializers with `register_generic`

jakirkham:
This changes `register_generic` (and surrounding machinery) to support using other protocols when traversing through an object's `__dict__`. This can be handy if another serialization protocol (like `"cuda"` 😉) is need for an object that contains content that can already be serialized using this protocol. A couple examples of this are other CuPy objects (like sparse matrices) and cuML models ( https://github.com/rapidsai/cuml/issues/1732 ).

cc @cjnolet @dantegd

### DIFF
--- a/distributed/protocol/cuda.py
+++ b/distributed/protocol/cuda.py
@@ -1,7 +1,7 @@
 import dask
 
 from . import pickle
-from .serialize import register_serialization_family
+from .serialize import ObjectDictSerializer, register_serialization_family
 from dask.utils import typename
 
 cuda_serialize = dask.utils.Dispatch("cuda_serialize")
@@ -29,3 +29,8 @@ def cuda_loads(header, frames):
 
 
 register_serialization_family("cuda", cuda_dumps, cuda_loads)
+
+
+cuda_object_with_dict_serializer = ObjectDictSerializer("cuda")
+
+cuda_deserialize.register(dict)(cuda_object_with_dict_serializer.deserialize)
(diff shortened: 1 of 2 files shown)


quasiben:
Please correct me if wrong, @jakirkham .  This will allow dask to use other serializers (should they be present) for mixed typed dictionary.   Something like:

> {msgpack, pickle, dask, cuda, cuda}

with this PR, as dask is iterating over the dictionary, dask will call out to the cuda serialization for cuda objects.   Where as before, it would have used pickle, correct ?

jakirkham:
Not quite.

Before this PR one could call `register_generic` on an object and that would allow it to go through `"dask"` serialization. After this PR one can call `register_generic` with another serialization protocol (like `"cuda"` serialization) and benefit from the same recursive serialization of the object, but with a different protocol.

Does that make more sense?

jakirkham:
As to usage, it should look something like this

```python
register_generic(MyClass, "cuda", cuda_serialize, cuda_deserialize)
```

quasiben:
Great!   Thanks for that explanation!  Would it make sense to add a test for this as well?  We could run any cuda tests in the ucx-py CI for now.  If not, let's merge

jakirkham:
Of course 😊

It might make sense. I didn't seen existing tests for it though. 😟 Maybe I'm missing something?

Am planning on adding a subsequent PR that supports a few different sparse matrix types (both SciPy and CuPy), which would be tested and leverage this functionality.

quasiben:
Thanks John.  A follow up PR with tests would be welcome

jakirkham:
Added PR ( https://github.com/dask/distributed/pull/3545 ), which includes sparse array serialization using this functionality.

PR #3538, opened 2020-03-01
Use `makedirs` when constructing `local_directory`

jakirkham:
Make sure the full path to the `temporary-directory` provided exists. Also make sure that if the path already exists no error is raised.

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -152,8 +152,7 @@ class Nanny(ServerNode):
 
         if local_directory is None:
             local_directory = dask.config.get("temporary-directory") or os.getcwd()
-            if not os.path.exists(local_directory):
-                os.mkdir(local_directory)
+            os.makedirs(local_directory, exist_ok=True)
             local_directory = os.path.join(local_directory, "dask-worker-space")
 
         self.local_directory = local_directory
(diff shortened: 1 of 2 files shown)


jakirkham:

jakirkham:
Yeah that is one reason we would want it. Another might be the user actually created this directory for us or it exists from a previous Dask session (since this path is in their config after-all 🙂).

jrbourbeau:
Gotcha, I think the current implementation is okay if the temporary directory already exists because of the `if not os.path.exists(local_directory)` condition

jakirkham:
Well the other thing this is doing is making sure all of the parent directories exist.

jrbourbeau:
Yeah, totally agree using `makedirs` is an improvement. One last nitpick: could you add back in the `if not os.path.exists(local_directory):` condition so we only call `makedirs` when the directory doesn't already exist?

jakirkham:
Sure can do. There’s no difference between the two though (just in case we are missing something).

jakirkham:
Have added back the `if` in the latest commit. Would you be able to take another look @jrbourbeau? 🙂

jrbourbeau:
Thanks @jakirkham!

Sorry for all the back and forth, but do we still want to have `exists_ok=True`? I was thinking about a similar change that was made in https://github.com/dask/zict/pull/33 to avoid a race condition with multiple workers trying to create a temp directory. If not, I'm happy to merge the current changes as is

jakirkham:
No worries. 😉 Either is fine with me 🙂

jrbourbeau:
Thanks, this is in

jakirkham:
Thanks James! 😄

PR #3539, opened 2020-03-02
Fix/more ucx config options

quasiben:
Adds two more config options:

- ucx.tcp
- ucx.cuda_copy

`tcp` is necessary when _not_ using any accelerated devices but still wanting to use UCX
`cuda_copy` is necessary when _not_ passing any CUDA objects with UCX -- host memory only.  Potentially useful for IB+UCX with numpy/pandas only data

cc @pentschev

### DIFF
--- a/distributed/comm/tests/test_ucx_config.py
+++ b/distributed/comm/tests/test_ucx_config.py
@@ -20,18 +20,30 @@ rmm = pytest.importorskip("rmm")
 @pytest.mark.asyncio
 async def test_ucx_config(cleanup):
 
-    ucx = {"nvlink": True, "infiniband": True, "net-devices": ""}
+    ucx = {
+        "nvlink": True,
+        "infiniband": True,
+        "net-devices": "",
+        "tcp": True,
+        "cuda_copy": True,
+    }
 
     with dask.config.set(ucx=ucx):
         ucx_config = _scrub_ucx_config()
         assert ucx_config.get("TLS") == "rc,tcp,sockcm,cuda_copy,cuda_ipc"
         assert ucx_config.get("NET_DEVICES") is None
 
-    ucx = {"nvlink": False, "infiniband": True, "net-devices": "mlx5_0:1"}
+    ucx = {
+        "nvlink": False,
+        "infiniband": True,
+        "net-devices": "mlx5_0:1",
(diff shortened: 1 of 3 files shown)


distributed/comm/ucx.py
```diff
@@ -359,10 +359,22 @@ def _scrub_ucx_config():
...
     # we assume we should configure basic TLS settings for UCX
-    if any([dask.config.get("ucx.nvlink"), dask.config.get("ucx.infiniband")]):
-        tls = "tcp,sockcm,cuda_copy"
+    if any(
+        [
+            dask.config.get("ucx.tcp"),
+            dask.config.get("ucx.nvlink"),
+            dask.config.get("ucx.infiniband"),
+        ]
+    ):
```
pentschev: Since we have this condition now, I think we should error out if none of them are specified. This would still mean we leave things to the UCX default, but we haven't really tested that well and we know it doesn't work in certain cases (e.g., DGX-2 where the number of transports is higher than 64).

distributed/comm/tests/test_ucx_config.py
```diff
@@ -20,32 +20,58 @@
...
+    ucx = {
+        "nvlink": False,
+        "infiniband": False,
+        "net-devices": "all",
+        "MEMTYPE_CACHE": "y",
+        "tcp": False,
+        "cuda_copy": True,
+    }
+    with dask.config.set(ucx=ucx):
+        with raises(ValueError):
```
jakirkham: Did `raises` get imported earlier or should this be `pytest.raises`? Seeing [this linting error in `master`]( https://travis-ci.org/dask/distributed/jobs/657551806#L460 ).

```
distributed/comm/tests/test_ucx_config.py:72:14: F821 undefined name 'raises'
```

jakirkham: Maybe this fixes it ( https://github.com/dask/distributed/pull/3541 )?

quasiben: Thanks for the fixes John.  I think something is not syncing correctly between my vscode sessions and the remote node. Again, thank you for fixing my mistake

jakirkham: No worries 😉

PR #3544, opened 2020-03-03
Use UCX default configuration instead of raising

pentschev:
Raising an error can be a problem with scheduler, given that we don't configure it in `LocalCluster`, thus removing that for the time being.


pentschev:
cc @quasiben

quasiben:
Thanks Peter and apologies to other dask devs while we were sorting out dask-cuda/ucx interactions with dask core.

PR #3545, opened 2020-03-03
Serialize sparse arrays

jakirkham:
Builds off of PR ( https://github.com/dask/distributed/pull/3536 )

Adds support for more efficient serialization of sparse arrays from both SciPy and CuPy. As both implementations frequently are simple classes that contain objects that are already serializable like NumPy's or CuPy's `ndarray`s, this just uses `register_generic` to recurse through the accepted object's `__dict__` serializing anything it finds. The deserialization process works more-or-less the same. As a result, the underlying data can be transformed into frames efficiently and transmitted without needing to resort to `pickle`.

Mostly this just works with the exception of SciPy's `dok_matrix`, which is a weird hybrid of a `dict` and a `spmatrix`. Unfortunately this confuses Dask as it isn't quite sure how to serialize it. So we special case this and convert to a `coo_matrix`, which is very similar in layout and a bit easier for us to serialize.

cc @dantegd @cjnolet @quasiben

### DIFF
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -4,7 +4,7 @@ Efficient serialization GPU arrays.
 import cupy
 
 from .cuda import cuda_deserialize, cuda_serialize
-from .serialize import dask_deserialize, dask_serialize
+from .serialize import dask_deserialize, dask_serialize, register_generic
 
 try:
     from .rmm import dask_deserialize_rmm_device_buffer as dask_deserialize_cuda_buffer
@@ -80,3 +80,20 @@ def dask_deserialize_cupy_ndarray(header, frames):
     frames = [dask_deserialize_cuda_buffer(header, frames)]
     arr = cuda_deserialize_cupy_ndarray(header, frames)
     return arr
+
+
+try:
+    from cupy.cusparse import MatDescriptor
+    from cupyx.scipy.sparse import spmatrix
+
+    cupy_sparse_types = [MatDescriptor, spmatrix]
+except ImportError:
+    cupy_sparse_types = []
+
+
+for t in cupy_sparse_types:
+    for n, s, d in [
+        ("cuda", cuda_serialize, cuda_deserialize),
+        ("dask", dask_serialize, dask_deserialize),
+    ]:
+        register_generic(t, n, s, d)
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -30,6 +30,12 @@ def _register_numpy():
     from . import numpy
 
 
+@dask_serialize.register_lazy("scipy")
+@dask_deserialize.register_lazy("scipy")
+def _register_scipy():
+    from . import scipy
+
+
 @dask_serialize.register_lazy("h5py")
 @dask_deserialize.register_lazy("h5py")
 def _register_h5py():
--- a/distributed/protocol/scipy.py
+++ b/distributed/protocol/scipy.py
@@ -6,3 +6,25 @@ import scipy
 from .serialize import dask_deserialize, dask_serialize, register_generic
 
 register_generic(scipy.sparse.spmatrix, "dask", dask_serialize, dask_deserialize)
+
+
+@dask_serialize.register(scipy.sparse.dok.dok_matrix)
+def serialize_scipy_sparse_dok(x):
+    x_coo = x.tocoo()
+    coo_header, coo_frames = dask_serialize(x.tocoo())
+
+    header = {"coo_header": coo_header}
+    frames = coo_frames
+
+    return header, frames
+
+
+@dask_deserialize.register(scipy.sparse.dok.dok_matrix)
+def deserialize_scipy_sparse_dok(header, frames):
+    coo_header = header["coo_header"]
+    coo_frames = frames
+    x_coo = dask_deserialize(coo_header, coo_frames)
+
+    x = x_coo.todok()
+
+    return x
--- a/distributed/protocol/scipy.py
+++ b/distributed/protocol/scipy.py
@@ -1,0 +1,8 @@
+"""
+Efficient serialization of SciPy sparse matrices.
+"""
+import scipy
+
+from .serialize import dask_deserialize, dask_serialize, register_generic
+
+register_generic(scipy.sparse.spmatrix, "dask", dask_serialize, dask_deserialize)
--- a/distributed/protocol/tests/test_cupy.py
+++ b/distributed/protocol/tests/test_cupy.py
@@ -4,6 +4,7 @@ import pytest
 from distributed.protocol import deserialize, serialize
 
 cupy = pytest.importorskip("cupy")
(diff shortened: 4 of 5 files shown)


distributed/protocol/cupy.py
```diff
@@ -80,3 +80,20 @@ def dask_deserialize_cupy_ndarray(header, frames):
...
     arr = cuda_deserialize_cupy_ndarray(header, frames)
     return arr
+
+
+try:
+    from cupy.cusparse import MatDescriptor
+    from cupyx.scipy.sparse import spmatrix
+
+    cupy_sparse_types = [MatDescriptor, spmatrix]
+except ImportError:
```
quasiben: What is the behavior here when the import fails.  Will there be no serialization defined for cupy sparse arrays ?

quasiben:
In general this looks ok to me.  Though I did leave one comment about the import try/expect behavior.  I would be good to get @mrocklin's thoughts if he is available as this starts to touch lowers levels of serialization within dask

distributed/protocol/cupy.py
mrocklin: Presumably in this case they don't exist.

This seems fine to me.  I might write it slightly differently, but this is just a subjective/style difference and not a difference worth spending much time on.

```python
try:
    import ...
except ImportError:
    pass
else:
    for typ in [MatDescriptor, spmatrix]:
        ...
```

jakirkham: If the `import` fails, we don't register these. 🙂

Yeah `else` as part of `try...except...finally...else` and `for...else` has bitten me a few times in unexpected ways before. Would prefer to avoid it as it is not really any simpler and the failure modes are annoying to debug.

quasiben: Ah ok. if the import fails then nothing is registered.  Sorry, that took me a minute.

quasiben:
Thanks for the extra eyes @mrocklin.  And thank you @jakirkham for the work!  Merging in

jakirkham:
Thanks for the reviews! 😄

madsbk:
After this PR, I get the following error when running pytest.
Notice, the CI doesn't run `test_cupy.py` or any other test that requires CUDA.

```
pytest "../distributed/distributed/protocol/tests/test_cupy.py::test_serialize_cupy_sparse[cuda-dtype0-csr_matrix]" -vsx
==================================================================================================================== test session starts =====================================================================================================================
platform linux -- Python 3.7.6, pytest-5.3.5, py-1.8.1, pluggy-0.12.0 -- /home/mkristensen/apps/miniconda3/envs/rap-nightly13/bin/python
cachedir: .pytest_cache
rootdir: /home/mkristensen/repos/distributed, inifile: setup.cfg
plugins: asyncio-0.10.0
collected 1 item                                                                                                                                                                                                                                             

../distributed/distributed/protocol/tests/test_cupy.py::test_serialize_cupy_sparse[cuda-dtype0-csr_matrix] FAILED

========================================================================================================================== FAILURES ==========================================================================================================================
_____________________________________________________________________________________________________ test_serialize_cupy_sparse[cuda-dtype0-csr_matrix] _____________________________________________________________________________________________________

sparse_type = <class 'cupyx.scipy.sparse.csr.csr_matrix'>, dtype = dtype('float32'), serializer = 'cuda'

    @pytest.mark.parametrize(
        "sparse_type",
        [
            cupy_sparse.coo_matrix,
            cupy_sparse.csc_matrix,
            cupy_sparse.csr_matrix,
            cupy_sparse.dia_matrix,
        ],
    )
    @pytest.mark.parametrize(
        "dtype",
        [numpy.dtype("<f4"), numpy.dtype(">f4"), numpy.dtype("<f8"), numpy.dtype(">f8"),],
    )
    @pytest.mark.parametrize("serializer", ["cuda", "dask",])
    def test_serialize_cupy_sparse(sparse_type, dtype, serializer):
        a_host = numpy.array([[0, 1, 0], [2, 0, 3], [0, 4, 0]], dtype=dtype)
        a = cupy.asarray(a_host)
    
        anz = a.nonzero()
        acoo = cupy_sparse.coo_matrix((a[anz], anz))
        asp = sparse_type(acoo)
    
        header, frames = serialize(asp, serializers=[serializer])
>       asp2 = deserialize(header, frames)

../distributed/distributed/protocol/tests/test_cupy.py:91: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../distributed/distributed/protocol/serialize.py:268: in deserialize
    return loads(header, frames)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

header = {'serializer': 'error'}, frames = [b'Could not serialize object of type csr_matrix.']

    def serialization_error_loads(header, frames):
        msg = "\n".join([ensure_bytes(frame).decode("utf8") for frame in frames])
>       raise TypeError(msg)
E       TypeError: Could not serialize object of type csr_matrix.

../distributed/distributed/protocol/serialize.py:80: TypeError
================================================================================================================= slowest 10 test durations ==================================================================================================================
0.80s call     distributed/protocol/tests/test_cupy.py::test_serialize_cupy_sparse[cuda-dtype0-csr_matrix]

(0.00 durations hidden.  Use -vv to show these durations.)
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
===================================================================================================================== 1 failed in 1.07s ======================================================================================================================
```

PR #3550, opened 2020-03-04
Add prometheus metric for suspicious tasks

sublinus:
This PR adds a prometheus metric for suspicious tasks parted by the task prefix name.

### DIFF
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -292,7 +292,7 @@ class _PrometheusCollector:
         suspicious_tasks = CounterMetricFamily(
             "dask_scheduler_tasks_suspicious",
             "Total number of times a task has been marked suspicious",
-            labels=["name"],
+            labels=["task_name"],
         )
 
         suspicious_by_prefix = {
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -292,11 +292,11 @@ class _PrometheusCollector:
         suspicious_tasks = CounterMetricFamily(
             "dask_scheduler_tasks_suspicious",
             "Total number of times a task has been marked suspicious",
-            labels=["task_name"],
+            labels=["task_prefix_name"],
         )
 
-        for name, val in self.server.suspicious_by_prefix.items():
-            suspicious_tasks.add_metric([name], val)
+        for tp in self.server.task_prefixes.values():
+            suspicious_tasks.add_metric([tp.name], tp.suspicious)
         yield suspicious_tasks
 
         yield CounterMetricFamily(
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -289,6 +289,24 @@ class _PrometheusCollector:
             sum, (tp.states for tp in self.server.task_prefixes.values())
         )
 
+        suspicious_tasks = CounterMetricFamily(
+            "dask_scheduler_tasks_suspicious",
+            "Total number of times a task has been marked suspicious",
+            labels=["name"],
+        )
+
+        suspicious_by_prefix = {
+            prefix.name: 0 for prefix in self.server.task_prefixes.values()
+        }
+
+        for task in self.server.tasks.values():
+            if task.suspicious > 0:
+                suspicious_by_prefix[task.prefix.name] += task.suspicious
+
+        for name, val in suspicious_by_prefix.items():
+            suspicious_tasks.add_metric([name], val)
+        yield suspicious_tasks
+
         yield CounterMetricFamily(
             "dask_scheduler_tasks_forgotten",
             (
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -1,4 +1,3 @@
-from collections import Counter
 from datetime import datetime
 from functools import partial
 import os
@@ -296,11 +295,7 @@ class _PrometheusCollector:
             labels=["task_name"],
         )
 
-        suspicious_by_prefix = Counter()
-        for task in self.server.tasks.values():
-            suspicious_by_prefix[task.prefix.name] += task.suspicious
-
-        for name, val in suspicious_by_prefix.items():
+        for name, val in self.server.suspicious_by_prefix.items():
             suspicious_tasks.add_metric([name], val)
         yield suspicious_tasks
 
--- a/distributed/dashboard/scheduler.py
+++ b/distributed/dashboard/scheduler.py
@@ -1,3 +1,4 @@
+from collections import Counter
 from datetime import datetime
 from functools import partial
 import os
@@ -295,13 +296,9 @@ class _PrometheusCollector:
             labels=["task_name"],
         )
 
-        suspicious_by_prefix = {
-            prefix.name: 0 for prefix in self.server.task_prefixes.values()
-        }
-
+        suspicious_by_prefix = Counter()
         for task in self.server.tasks.values():
-            if task.suspicious > 0:
-                suspicious_by_prefix[task.prefix.name] += task.suspicious
+            suspicious_by_prefix[task.prefix.name] += task.suspicious
 
         for name, val in suspicious_by_prefix.items():
             suspicious_tasks.add_metric([name], val)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -791,7 +791,7 @@ class TaskPrefix:
 
     .. attribute:: suspicious: int
 
-       Numbers of tasks marked as suspicious with this prefix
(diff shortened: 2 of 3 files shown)


mrocklin:
I apologize for the lack of response here.

@gforsyth any chance you have some familiarity with Prometheus metrics and have some time to review this?

distributed/dashboard/scheduler.py
```diff
@@ -292,6 +292,24 @@ def collect(self):
...
         )
 
+        suspicious_tasks = CounterMetricFamily(
+            "dask_scheduler_tasks_suspicious",
+            "Total number of times a task has been marked suspicious",
+            labels=["name"],
+        )
+
+        suspicious_by_prefix = {
+            prefix.name: 0 for prefix in self.server.task_prefixes.values()
```
gforsyth: Rather than pre-initializing a dictionary here with keys corresponding to every task, I think a `collections.Counter` is well suited to this

sublinus: Thanks for the review! 
According to [Robust Perception](https://www.robustperception.io/existential-issues-with-metrics) and the [Prometheus best practices](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics) timeseries should be initialised with zero.

> Where this usually comes up is for metrics with labels. It is recommended to initialise metrics to zero when your application starts, so that when there's an increment on a subsequent scrape we'll be able to see that the value has increased over time.

gforsyth: Ok, cool.  I think a `Counter` still makes sense here and avoids the (admittedly small) cost of iterating over every task twice.

Maybe something like

```
from collections import Counter

suspicious_by_prefix = Counter()
for task in self.server.tasks.values():
    suspicious_by_prefix[task.prefix.name] += task.suspicious
```

sublinus: I think the counter is really cool, but what benefit does it give here over a regular dictionary?

I fully agree with the need to simplify the whole collection of suspicious tasks, but I currently don't see the need for a special data structure, which extends the dict with an interface we would not really use here.

sublinus: Oh nevermind. Thanks a lot for the advice regarding the Counter.

distributed/dashboard/scheduler.py
```diff
@@ -292,6 +293,20 @@ def collect(self):
...
         )
 
+        suspicious_tasks = CounterMetricFamily(
+            "dask_scheduler_tasks_suspicious",
+            "Total number of times a task has been marked suspicious",
+            labels=["task_name"],
+        )
+
+        suspicious_by_prefix = Counter()
+        for task in self.server.tasks.values():
```
jrbourbeau: In general we try to avoid looping over all tasks where possible. This can noticeably slow things down when there are millions of tasks. Do you have any thoughts on how we might improve things here?

sublinus: The only thing I can think of right now would be keeping the `suspicious_by_prefix` counter at the scheduler similar to the unrunnable tasks and increasing this counter every time a task is marked as suspicious.

TomAugspurger: That seems reasonable to me.

fjetter: Instead of increasing the counter similar to `unrunnable` we might want to include this in the `TaskGroup`/`TaskPrefix` containers. Should be similar in complexity but the group/prefix containers seem to be the "future" way of aggregating task state

sublinus: As my counter is already called `suspicious_by_prefix` putting it into the `TaskPrefix` seems very sensible to me. Thanks @fjetter

sublinus: Are the TaskPrefixes cleared after a graph is calculated?

jrbourbeau: `TaskGroup`s are deleted up when all the tasks in the group are in the `'forgotten'` state, but `TaskPrefixes` are not removed

sublinus:
I am confused regarding use of TaskPrefix vs TaskGroup in this instance.
In the case of TaskGroups consisting of tasks in a single layer in the dask graph I can see that this would be very useful, but I can't find the encoding of the layer in the task-key right now. 
What is the use of TaskGroups at the moment? Do they have to be set manually by e.g. setting the task-keys yourself?
Which of the both abstractions should I go for regarding the suspicious counts?

distributed/dashboard/scheduler.py
sublinus: So we would have to use `TaskGroups` in order to not potentially have a wrong counter for the tasks, but with the current state of `TaskGroups` for example for `delayed` objects there seems to be one TaskGroup per task. This would not seem feasible to me and I would stay with the implementation at the Scheduler here for now. Are there any comments on this? Is the thing with the TaskGroups for delayed actually correct?

distributed/tests/test_scheduler.py
```diff
@@ -225,6 +225,21 @@ def test_remove_worker_by_name_from_scheduler(s, a, b):
...
+    s.update_graph(
+        tasks=valmap(dumps_task, dsk),
+        keys=list(dsk),
+        dependencies={k: set() for k in dsk},
+    )
+    assert s.remove_worker(address=a.address) == "OK"
+    assert sum(tp.suspicious for tp in s.task_prefixes.values()) == sum(
+        ts.suspicious for ts in s.tasks.values()
+    )
+    assert s.task_prefixes["x"].suspicious == 10
```
mrocklin:
```suggestion
    assert s.task_prefixes["inc"].suspicious == 10
```

distributed/tests/test_scheduler.py
```diff
@@ -225,6 +225,21 @@ def test_remove_worker_by_name_from_scheduler(s, a, b):
...
 
+@gen_cluster()
+def test_remove_worker_increments_suspicious(s, a, b):
+    dsk = {("x-%d" % i): (inc, i) for i in range(20)}
+    s.update_graph(
+        tasks=valmap(dumps_task, dsk),
+        keys=list(dsk),
+        dependencies={k: set() for k in dsk},
+    )
+    assert s.remove_worker(address=a.address) == "OK"
```
mrocklin: Ah, I see, you're using internal API in order to simulate a failing task.  

Instead, you might consider using a function that generally causes a worker to fail, like `sys.exit` (you'll want to make sure that you use `gen_cluster(Worker=Nanny)` so that we use separate worker processes.

distributed/scheduler.py
```diff
@@ -789,6 +789,11 @@ class TaskPrefix:
 
        An exponentially weighted moving average duration of all tasks with this prefix
 
+    .. attribute:: suspicious: int
+
+       Numbers of tasks marked as suspicious with this prefix
```
mrocklin:
```suggestion
       Numbers of times a task was marked as suspicious with this prefix
```

A single task might add a few to this count

distributed/tests/test_scheduler.py
```diff
@@ -225,6 +225,22 @@ def test_remove_worker_by_name_from_scheduler(s, a, b):
...
+
+        sys.exit()
+
+    futures = client.map(killer, range(20))
+    while len(s.workers) == 2:
+        yield gen.sleep(0.01)
+    assert sum(tp.suspicious for tp in s.task_prefixes.values()) == sum(
+        ts.suspicious for ts in s.tasks.values()
+    )
+    assert 10 <= s.task_prefixes["killer"].suspicious <= 30
```
sublinus: @mrocklin I changed the test to not depend on internal APIs, but I'm not really happy with the way the test works right now. For me it was flapping, as sometimes one worker is killed until the wait is over and sometimes two are killed which leads to either 10 tasks marked as suspicious or 30.
Do you have an idea how to solve this properly? I tried to await the submitted tasks, but then I got `KilledWorker` exceptions.

mrocklin: Maybe something like the following?

```python
@gen_cluster(client=True, Worker=Nanny, config={"distributed.scheduler.suspicious-tasks": 1})  # TODO verify config key
async def test_suspicious_prefix(client, s, a, b):
    future = client.submit(sys.exit, 0)
    await wait(future)
    assert s.task_prefixes["exit"].suspicious
```

This isn't as strong of a test as what you have here (there is only one task), but it should be simpler and run more more quickly.

sublinus: It especially takes away the flapping, because of the `allowed-failures` configuration. Thanks a lot for the help @mrocklin

sublinus:
I have addressed the feedback and think this should be ready to merge.

mrocklin:
Thank you @sublinus .  Merging.

PR #3554, opened 2020-03-05
Add 'local_directory' option to dask-ssh

abduhbm:
This PR enables `--local-directory` option in `dask-ssh` cli to change the default location where workers and scheduler place their files.

### DIFF
--- a/distributed/cli/dask_ssh.py
+++ b/distributed/cli/dask_ssh.py
@@ -81,7 +81,7 @@ from distributed.cli.utils import check_python_3
     "--local-directory",
     default=None,
     type=click.Path(exists=True),
-    help="Directory to place workers and scheduler files.",
+    help="Directory to use on to all cluster nodes to place workers and scheduler files.",
 )
 @click.option(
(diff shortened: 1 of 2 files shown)


abduhbm:
It is on the SSH machine. `dask-ssh` by default places `dask-worker-space` in the user home directory which can create disk space problems in environments with quota-enabled network home shared by NFS. 

Unknown options can be passed in Click. Please see [Forwarding Unknown Options](http://click.pocoo.org/5/advanced/#forwarding-unknown-options).

TomAugspurger:
> It is on the SSH machine. 

Can you clarify that in the docs?

abduhbm:
I updated the docs for the `--local-directory` option to: 
`Directory to use on all cluster nodes to place workers and scheduler files.
`
> Thoughts on just forwarding all additional arguments through to dask-worker & dask-scheduler?

I think it can be don. I will do another PR for this. 

Thanks @TomAugspurger

TomAugspurger:
Thanks @abduhbm!

PR #3549, opened 2020-03-04
Register Cupyx

quasiben:
Fixes https://github.com/dask/distributed/issues/3547

PR adds cupyx to list of registered (de-)serializers

This does pass but it does produce some nastiness at the end (we can probably ignore but thought it was worth mentioning)

```
distributed/protocol/tests/test_cupy.py::test_serialize_cupy_sparse[cuda-dtype0-coo_matrix] PASSED
distributed/protocol/tests/test_cupy.py::test_serialize_cupy_sparse[cuda-dtype0-csc_matrix] PASSED
distributed/protocol/tests/test_cupy.py::test_serialize_cupy_sparse[cuda-dtype0-csr_matrix] malloc_consolidate(): invalid chunk size
Fatal Python error: Aborted

Current thread 0x00007f6989001740 (most recent call first):
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/_pytest/logging.py", line 628 in _runtest_for_main
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/contextlib.py", line 81 in __enter__
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/_pytest/logging.py", line 607 in _runtest_for
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/contextlib.py", line 81 in __enter__
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/_pytest/logging.py", line 645 in pytest_runtest_setup
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/pluggy/callers.py", line 182 in _multicall
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/pluggy/manager.py", line 81 in <lambda>
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/pluggy/manager.py", line 87 in _hookexec
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/pluggy/hooks.py", line 289 in __call__
  File "/datasets/bzaitlen/miniconda3/envs/20200302/lib/python3.6/site-packages/_pytest/runner.py", line 210 in <lambda>
[...]

Issue #3547, opened 2020-03-04
Flaky serialize cupy sparse array test


TomAugspurger:
Reported in https://github.com/dask/distributed/pull/3545#issuecomment-594509462.

cc @jakirkham

jakirkham:
Thanks Tom! I think Ben's PR ( https://github.com/dask/distributed/pull/3549 ) will fix this. Sorry for missing that.

PR #3556, opened 2020-03-05
Fix-up CuPy sparse serialization

jakirkham:
Fixes https://github.com/dask/distributed/issues/3547
Closes https://github.com/dask/distributed/pull/3549

This pulls in the registration fixes from @quasiben's PR.

Also this special cases the serialization of `MatDescriptor` instead of using `register_generic`. While it appears `MatDescriptor` just contains a Python `int`. It turns out this actually a pointer. When this pointer is check in another process that doesn't contain the resource, it becomes invalid. To fix this, we just construct a new `MatDescriptor`. Turns out this is what CuPy would do anyways and what cuML did to fix a similar problem (thanks @cjnolet for the tip! 😄).

Cleans up the tests to use SciPy sparse matrices to build the CuPy sparse matrices from. This simplifies things a bit. Though some special casing was needed here due to some oddities with CuPy's `dia_matrix`. Will work with upstream to fix this so we can drop that special casing in the future.

cc @TomAugspurger

### DIFF
--- a/distributed/protocol/cupy.py
+++ b/distributed/protocol/cupy.py
@@ -1,6 +1,8 @@
 """
 Efficient serialization GPU arrays.
 """
+import copyreg
+
 import cupy
 
 from .cuda import cuda_deserialize, cuda_serialize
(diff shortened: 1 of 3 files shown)


distributed/protocol/cupy.py
```diff
@@ -85,15 +87,35 @@ def dask_deserialize_cupy_ndarray(header, frames):
...
+
+
+if MatDescriptor is not None:
+
+    def reduce_matdescriptor(other):
+        # Pickling MatDescriptor errors
+        # xref: https://github.com/cupy/cupy/issues/3061
+        return cupy.cusparse.MatDescriptor.create, ()
+
+    copyreg.pickle(MatDescriptor, reduce_matdescriptor)
```
jakirkham: `MatDescriptor` goes through normal pickling. However that will cause a segfault (as it will contain an invalid `int` that points to invalid memory). So we special case pickling to avoid this.

xref: https://github.com/cupy/cupy/issues/3061

jakirkham: `MatDescriptor` goes through normal pickling. However that will cause a segfault (as it will contain an invalid `int` that points to invalid memory). So we special case pickling to avoid this.

xref: https://github.com/cupy/cupy/issues/3061

distributed/protocol/tests/test_cupy.py
```diff
@@ -66,31 +65,37 @@ def test_serialize_cupy_from_rmm(size):
...
+    a_host = numpy.array([[0, 1, 0], [2, 0, 3], [0, 4, 0]], dtype=dtype)
+    asp_host = scipy_sparse_type(a_host)
+    if sparse_name == "dia_matrix":
+        # CuPy `dia_matrix` cannot be created from SciPy one
+        # xref: https://github.com/cupy/cupy/issues/3158
+        asp_dev = cupy_sparse_type(
+            (asp_host.data, asp_host.offsets),
+            shape=asp_host.shape,
+            dtype=asp_host.dtype,
+        )
```
jakirkham: CuPy sparse matrix types should support taking a SciPy matrix as input to the constructor. However this doesn't work correctly with CuPy's `dia_matrix`. So we special case it here.

xref: https://github.com/cupy/cupy/issues/3158

jakirkham:
Yeah that's unfortunate, but I'm not too worried about it either. Agree it would be nice to see this land 😄

jakirkham:
Thanks all! 😀

Issue #3552, opened 2020-03-04
Clarification on worker resources documentation


chrisroat:
In the (worker resources docs)[https://distributed.dask.org/en/latest/resources.html#resources-are-applied-separately-to-each-worker-process], it shows the following scenario.

> ```
> dask-worker scheduler:8786 --nprocs 3 --nthreads 2 --resources "process=1"
> ```
> 
> With the code below, there will be at most 3 tasks running concurrently and each task will run in a separate process:
> ```
> from distributed import Client
> client = Client('scheduler:8786')
> 
> futures = [client.submit(non_thread_safe_function, arg,
>                          resources={'process': 1}) for arg in args]
> ```
> 

If additional tasks are scheduled without the `process` resource, might they be scheduled on the unused threads?  The scenario I'm concerned about is when the `{process: 1}` consumes multiple threads, effectively starving these additional tasks.  

If this is so, this leads me down the road of tagging all my additional tasks (and other workers) to avoid this situation.  Consequently, I then lose out on scheduling on the `process` worker with these  tasks when no `process` tasks are running.  (Is this clear?)

I'd be happy to update the documentation to help elucidate this situation, once I feel I fully grok what is going on.

TomAugspurger:
> If additional tasks are scheduled without the process resource, might they be scheduled on the unused threads?

I think you're correct, but it'd be good to verify with tests / going through the code.

chrisroat:
Anecdotally, I was finding this to be true.  The trick of defining resources for all workers and all tasks can mitigate the problem.  

However, this leads to under-utilization because an unused thread-hostile worker will not take up thread-safe work.  Additionally, heavy use of resources requires turning off graph optimization for scheduling when intermediate steps need resources.

I added a test+doc to hopefully clarify the situation.  Please let me know if anything is unclear.

Would it be a plausible to consider special task-only resources that could take-over an entire or partial worker?  I am thinking along the lines of special resources named {"_PROCESS_": 1} or {"_THREADS_": 4}.

mrocklin:
> Would it be a plausible to consider special task-only resources that could take-over an entire or partial worker? I am thinking along the lines of special resources named {"PROCESS": 1} or {"THREADS": 4}.

Yes, it would be plausible to consider that.  That being said, I don't immediately have a plan of how it would be implemented.  Someone would have to put some thought into it, propose a plan, and then implement it.

PR #3561, opened 2020-03-08
Test to verify resources do not block all threads.

chrisroat:
Part of discussion in #3552

### DIFF
--- a/distributed/tests/test_resources.py
+++ b/distributed/tests/test_resources.py
@@ -393,3 +393,15 @@ def test_collections_get(client, optimize_graph, s, a, b):
     logs = client.run(g)
     assert logs[a["address"]]
     assert not logs[b["address"]]
+
+
+@gen_cluster(
+    client=True, nthreads=[("127.0.0.1", 2, {"resources": {"A": 1}}),],
+)
+def test_resources_do_not_block_all_threads(c, s, a):
+    constrained = c.submit(slowinc, 1, delay=0.2, resources={"A": 1})
+    unconstrained = c.submit(inc, 2)
+
+    # Fast, unconstrained work finished while slower, constrained is processing.
+    yield wait(unconstrained)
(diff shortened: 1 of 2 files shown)


distributed/tests/test_resources.py
```diff
@@ -393,3 +393,15 @@ def g(dask_worker):
...
+@gen_cluster(
+    client=True, nthreads=[("127.0.0.1", 2, {"resources": {"A": 1}}),],
+)
+def test_resources_do_not_block_all_threads(c, s, a):
+    constrained = c.submit(slowinc, 1, delay=0.2, resources={"A": 1})
+    unconstrained = c.submit(inc, 2)
+
+    # Fast, unconstrained work finished while slower, constrained is processing.
+    yield wait(unconstrained)
+    assert s.get_task_status(keys=[constrained.key]) == {constrained.key: "processing"}
```
mrocklin: We tend to discourage the use of internal API in tests when possible.  That way we're more easily able to change internals without modifying very many tests.  Maybe you wanted to test that `constrained.status == "pending"` or something similar?  

This test is also a little bit prone to intermittent failure.  The CI systems can be sluggish sometimes, and I wouldn't be surprised to see a 200ms delay caused by something like GC, which would occasionally cause this test to fail.  I'm not sure how exactly to solve the problem.  Increasing the delay to something like 500ms might be safer?

chrisroat: I agree with both points.  I didn't realize I was using an internal API.  (I simply was cribbing from what was done elsewhere in this file.)

I'm not sure I want "pending" here.  I think(?) I want to verify that it is still running.  (And, in fact, I should verify this before I start the fast tasks.)

My preference would be to examine details about how the tasks were executed -- essentially, I want to verify the tasks ran side-by-side.  Is there a way to run the tasks and verify after-the-fact what worker/thread each task ran on and it's start/end time?

chrisroat:
<sniff>

PR #3563, opened 2020-03-10
Fix old_ssh redirecting output to logfile in non-bash default shell

abduhbm:
Using `--log-directory` with `dask-ssh` command fails in non-bash default shell. This PR forces run the command in bash.


TomAugspurger:
I'm not sure I know how to evaluate the safety of this change. People could be relying on items in their login config, which would no longer be read if just using bash, right?

abduhbm:
You are right. 
How about reading `stdout` and `stderr` from `paramiko.client.exec_command` and store them in a log directory on the client machine?

PR #3567, opened 2020-03-12
Add Python version check

jrbourbeau:
This PR adds the Python version to our version mismatch check. I've also updated `get_system_info` to return a dictionary instead of a list-of-tuples. This is so the `"host"` and `"packages"` data structures returned by `get_versions` are the same, and it it also made `test_python_version_mismatch_warning` slightly cleaner to write. 

cc @mrocklin 

Closes https://github.com/dask/dask/issues/6007


jrbourbeau:
Planning to merge in a few hours if no further comments

mrocklin:
Thanks @jrbourbeau !

PR #3568, opened 2020-03-12
Add Python version to version checking

mrocklin:
Previously we would capture the Python version, but not actually raise a
warning message if it differed, like we would do with package versions.

Now we also include the Python version in the list of packages so that
it gets the same version checking treatment.

Fixes #6007

Example
-------

    distributed.worker - WARNING - Mismatched versions found

    python
    +-----------------------+---------+
    |                       | version |
    +-----------------------+---------+
    | This Worker           | 3.7.4   |
    | scheduler             | 3.7.4   |
    | tcp://127.0.0.1:42441 | 3.7.4   |
    | tcp://127.0.0.1:46325 | 3.8.1   |
    +-----------------------+---------+

### DIFF
--- a/distributed/tests/test_versions.py
+++ b/distributed/tests/test_versions.py
@@ -1,4 +1,5 @@
 import re
+import sys
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


mrocklin:
Ah whoops.  I see that @jrbourbeau just fixed this in #3567

PR #3570, opened 2020-03-12
Pin bokeh in CI builds

jrbourbeau:


jrbourbeau:
The new `bokeh` 2.0 release [sets the asyncio event loop policy](https://github.com/bokeh/bokeh/blob/dea6ca42135c0402fa4f24fa502b58fce288e067/bokeh/server/tornado.py#L689-L692) when on Windows and Python 3.8+ to avoid `tornado` incompatibilities. This impacts us as we sometime need to use a custom event loop policy, which ends up being overwritten by `bokeh`. 

For the time being, I've pinned to `bokeh=1.4.0` in CI and have filed an upstream PR https://github.com/bokeh/bokeh/pull/9776. 

Also, the `test_dont_steal_unknown_functions` failure here seems to be a flaky test which has popped up before (xref https://travis-ci.org/github/dask/distributed/jobs/650611742#L2626). I've opened up an issue for it

jakirkham:
Thanks James! 😄

Issue #864, opened 2017-02-09
Throttling tasks


dhirschfeld:
To prevent too many tasks trying to use the same resources (e.g. database, web api) at the same time I would like to control the maximum degree of parallelism of certain jobs.

I understand that it should be possible to control this using queues as documented in [Data Strems with Queues](http://distributed.readthedocs.io/en/latest/queues.html#data-streams-with-queues) and as demonstrated in [this](http://stackoverflow.com/a/38921763/2960140) SO answer.

When I try a minimal example it doesn't seem if the `maxsize` argument is having any effect?
```python
from queue import Queue
import time

from distributed import Client
from distributed import as_completed
from numpy.random import rand


client = Client("10.200.13.45:8786")


def sleep(delay):
    from time import sleep
    sleep(delay)
    return delay


delay = 5*rand(10)

in_q = Queue()
for t in delay:
    in_q.put(t)

maxsize = 10
print("maxsize =", maxsize)
out_q = client.map(sleep, in_q, maxsize=maxsize)

t0 = time.time()
print()
futs = [out_q.get() for _ in range(len(delay))]
for fut in as_completed(futs):
    print(fut.result(), flush=True)
t1 = time.time()

print()
print("Total Delay =", sum(delay))
print("Max Delay =", max(delay))
print("Time Taken =", t1 - t0)
```
When the `maxsize` is equal to the number of tasks they all run in parallel and the time taken is equal to (just over) the largest delay:
```
maxsize = 10

0.605000352854
0.87092424672
1.71378184661
1.91747345547
1.94673285822
1.98018309505
2.78900939666
3.42870653016
3.53337024226
4.46855449187

Total Delay = 23.2537365159
Max Delay = 4.46855449187
Time Taken = 4.508450746536255
```
When the `maxsize` is equal to 1 I would expect the tasks to be run sequentially and the time taken to equal the sum of all delays. It appears that the `maxsize` argument has no effect and that the tasks all run in parallel?
```
maxsize = 1

0.32444368678
0.996665256193
1.55266709866
1.86744982966
1.96963848636
2.51513007203
2.5978904159
3.15468744398
3.87090085426
4.28739030954

Total Delay = 23.1368634534
Max Delay = 4.28739030954
Time Taken = 4.312431335449219
```
![image](https://cloud.githubusercontent.com/assets/881019/22780070/095a451c-ef09-11e6-88e7-a66d4db9b3f6.png)

Am I doing something really dumb or missing something obvious?

[...]

PR #3573, opened 2020-03-13
Add Semaphore extension

lr4d:
Supersedes #2690

Feedback is appreciated

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -174,6 +174,12 @@ class SemaphoreExtension:
 
 
 class Semaphore:
+    """
+    Note: dask executes functions by default assuming they are pure, when using semaphore acquire/releases inside
+    such a function, it must be noted that there *are* in-fact side-effects, thus, the function can no longer be
+    considered pure. If this is not taken into account, this could lead to some weirdly unexpected behavior.
+    """
+
     def __init__(self, max_leases=1, name=None, client=None):
         # NOTE: the `id` of the `Semaphore` instance will always be unique, even among different
         # instances for the same resource. The actual attribute that identifies a specific resource is `name`,
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -68,9 +68,8 @@ class SemaphoreExtension:
         self.pc_validate_leases.start()
         self._validation_running = False
 
-    async def create(
-        self, stream=None, name=None, client=None, timeout=None, max_leases=None
-    ):
+    # `comm` here is required by the handler interface
+    async def create(self, comm=None, name=None, max_leases=None):
         if name not in self.leases:
             assert isinstance(max_leases, int), max_leases
             self.max_leases[name] = max_leases
@@ -98,7 +97,7 @@ class SemaphoreExtension:
         return result
 
     async def acquire(
-        self, stream=None, name=None, client=None, timeout=None, identifier=None
+        self, comm=None, name=None, client=None, timeout=None, identifier=None
     ):
         with log_errors():
             if isinstance(name, list):
@@ -134,7 +133,7 @@ class SemaphoreExtension:
                         result = False
                 return result
 
-    async def release(self, stream=None, name=None, client=None, identifier=None):
+    async def release(self, comm=None, name=None, client=None, identifier=None):
         with log_errors():
             if isinstance(name, list):
                 name = tuple(name)
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -56,8 +56,12 @@ class SemaphoreExtension:
         self.scheduler.extensions["semaphores"] = self
         self.pc_validate_leases = PeriodicCallback(
             self._validate_leases,
-            parse_timedelta(
-                dask.config.get("distributed.scheduler.locks.lease-validation-interval")
+            1000
+            * parse_timedelta(
+                dask.config.get(
+                    "distributed.scheduler.locks.lease-validation-interval"
+                ),
+                default="s",
             ),
             io_loop=self.scheduler.loop,
         )
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -1,0 +1,255 @@
+from __future__ import absolute_import, division, print_function
+
+import uuid
+from collections import defaultdict, deque
+from functools import partial
+import asyncio
+import dask
+import tornado.locks
+import tornado.queues
+from concurrent.futures import TimeoutError
+from .client import Client, _get_global_client
+from .utils import PeriodicCallback, log_errors, parse_timedelta
+from .worker import get_client, get_worker
+from toolz.dicttoolz import valmap
+from .metrics import time
+
+
+class _Watch(object):
+    def __init__(self, duration=None):
+        self.duration = duration
+        self.started_at = None
+
+    def start(self):
+        self.started_at = time()
+
+    def leftover(self):
+        if self.duration is None:
+            return None
+        else:
+            elapsed = time() - self.started_at
+            return max(0, self.duration - elapsed)
+
+
+class SemaphoreExtension(object):
+    """ An extension for the scheduler to manage Semaphores
+
+    This adds the following routes to the scheduler
+
+    * semaphore_acquire
+    * semaphore_release
+    * semaphore_create
+    """
+
+    def __init__(self, scheduler):
+        self.scheduler = scheduler
+        self.leases = defaultdict(deque)
+        self.events = defaultdict(tornado.locks.Event)
+        self.max_leases = dict()
+        self.leases_per_client = defaultdict(partial(defaultdict, deque))
+        self.scheduler.handlers.update(
+            {
+                "semaphore_create": self.create,
+                "semaphore_acquire": self.acquire,
+                "semaphore_release": self.release,
+            }
+        )
+
+        self.scheduler.extensions["semaphores"] = self
+        self.pc_validate_leases = PeriodicCallback(
+            self._validate_leases,
+            parse_timedelta(
+                dask.config.get("distributed.scheduler.locks.lease-validation-interval")
+            ),
+            io_loop=self.scheduler.loop,
+        )
+        self.pc_validate_leases.start()
+        self._validation_running = False
+
+    async def create(
+        self, stream=None, name=None, client=None, timeout=None, max_leases=None
+    ):
+        if name not in self.leases:
+            assert isinstance(max_leases, int), max_leases
+            self.max_leases[name] = max_leases
+        else:
+            if max_leases != self.max_leases[name]:
+                raise ValueError(
+                    "Inconsistent max leases: %s, expected: %s"
+                    % (max_leases, self.max_leases[name])
+                )
+
+    async def _get_lease(self, client, name, identifier):
+        # We should make sure that the client is already properly registered with the scheduler
+        # otherwise the lease validation will mop up every acquired lease immediately
+        while client not in self.scheduler.clients:
+            print("wait")
+            await asyncio.sleep(0.001)
+
+        result = True
+        if len(self.leases[name]) < self.max_leases[name]:
+            self.leases[name].append(identifier)
+            self.leases_per_client[client][name].append(identifier)
+        else:
+            result = False
+        return result
+
+    async def acquire(
+        self, stream=None, name=None, client=None, timeout=None, identifier=None
+    ):
+        with log_errors():
+            if isinstance(name, list):
+                name = tuple(name)
+            w = _Watch(timeout)
+            w.start()
+
+            while True:
+                # Reset the event and try to get a release. The event will be set if the state
+                # is changed and helps to identify when it is worth to retry an acquire
+                self.events[name].clear()
+
+                future = asyncio.wait_for(
+                    self._get_lease(client, name, identifier), timeout=w.leftover()
+                )
+
+                try:
+                    result = await future
+                except TimeoutError:
+                    result = False
+
+                # If acquiring fails, we wait for the event to be set, i.e. something has
+                # been released and we can try to acquire again (continue loop)
+                if not result:
+                    future = asyncio.wait_for(
+                        self.events[name].wait(), timeout=w.leftover()
+                    )
+                    try:
+                        await future
+                        continue
+                    except TimeoutError:
+                        result = False
+                return result
+
+    async def release(self, stream=None, name=None, client=None, identifier=None):
+        with log_errors():
+            if isinstance(name, list):
+                name = tuple(name)
+            if name in self.leases and identifier in self.leases[name]:
+                self.scheduler.loop.add_callback(
+                    self._release_value, name, client, identifier
+                )
+            else:
+                raise ValueError("Semaphore released too many times.")
+
+    async def _release_value(self, name, client, identifier):
+        self.leases_per_client[client][name].remove(identifier)
+        self.leases[name].remove(identifier)
+        self.events[name].set()
+
+    def _release_client(self, client):
+        semaphore_names = list(self.leases_per_client[client])
+        for name in semaphore_names:
+            ids = list(self.leases_per_client[client][name])
+            for _id in list(ids):
+                self.scheduler.loop.add_callback(
+                    self._release_value, name=name, client=client, identifier=_id
+                )
+
+    async def _validate_leases(self):
+        if not self._validation_running:
+            self._validation_running = True
+            known_clients = set(self.leases_per_client.keys())
+            scheduler_clients = set(self.scheduler.clients.keys())
+            for client in known_clients - scheduler_clients:
+                client_has_leases = sum(
+                    valmap(len, self.leases_per_client[client]).values()
+                )
+                if client_has_leases:
+                    self._release_client(client)
+            else:
+                self._validation_running = False
+
+
+class Semaphore(object):
+    def __init__(self, max_leases=1, name=None, client=None):
+        self.client = client or _get_global_client() or get_worker().client
+        self.id = uuid.uuid4().hex
+        self.name = name or "semaphore-" + uuid.uuid4().hex
+        self.max_leases = max_leases
+
+        if self.client.asynchronous:
+            self._started = self.client.scheduler.semaphore_create(
+                name=self.name, max_leases=max_leases
+            )
+        else:
+            self.client.sync(
+                self.client.scheduler.semaphore_create,
+                name=self.name,
+                max_leases=max_leases,
+            )
+            self._started = True
+
+    def __await__(self):
+        async def create_semaphore():
+            await self._started
+            return self
+
+        return create_semaphore().__await__()
+
+    def acquire(self, timeout=None):
+        """
+        Acquire a semaphore.
+
+        If the internal counter is greater than zero, decrement it by one and return True immediately.
+        If it is zero, wait until a release() is called and return True.
+        """
+        return self.client.sync(
+            self.client.scheduler.semaphore_acquire,
+            name=self.name,
+            timeout=timeout,
+            client=self.client.id,
+            identifier=self.id,
+        )
+
+    def release(self):
+        """
+        Release a semaphore.
+
+        Increment the internal counter by one.
+        """
+
+        """ Release the lock if already acquired """
+        return self.client.sync(
+            self.client.scheduler.semaphore_release,
+            name=self.name,
+            client=self.client.id,
+            identifier=self.id,
+        )
+
+    def __enter__(self):
+        self.acquire()
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        self.release()
+
+    async def __aenter__(self):
+        await self.acquire()
+        return self
+
+    async def __aexit__(self, *args, **kwargs):
+        await self.release()
+
+    def __getstate__(self):
+        return (self.name, self.client.scheduler.address, self.max_leases)
+
+    def __setstate__(self, state):
+        name, address, max_leases = state
+        try:
+            client = get_client(address)
+        except (AttributeError, AssertionError):
+            client = Client(address, set_as_default=False)
+        self.__init__(name=name, client=client, max_leases=max_leases)
+
+    def close(self):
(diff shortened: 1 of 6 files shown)


distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
...
+                raise ValueError(
+                    "Inconsistent max leases: %s, expected: %s"
+                    % (max_leases, self.max_leases[name])
+                )
+
+    async def _get_lease(self, client, name, identifier):
+        # We should make sure that the client is already properly registered with the scheduler
+        # otherwise the lease validation will mop up every acquired lease immediately. That's mostly relevant for tests
+        while client not in self.scheduler.clients:
+            await asyncio.sleep(0)
```
mrocklin: Should this be a non-zero value?

distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
...
+
+    async def release(self, comm=None, name=None, client=None, identifier=None):
+        with log_errors():
+            if isinstance(name, list):
+                name = tuple(name)
+            if name in self.leases and identifier in self.leases[name]:
+                self.scheduler.loop.add_callback(
+                    self._release_value, name, client, identifier
+                )
+            else:
```
mrocklin: Can we do this immediately?  Why do we need to do it in a future cycle?

distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
...
+                raise ValueError("Semaphore released too many times.")
+
+    async def _release_value(self, name, client, identifier):
```
mrocklin: Does this need to be `async def`?  We don't await in here, so could we make this a normal function?

distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
...
+
+    * semaphore_acquire
+    * semaphore_release
+    * semaphore_create
+    """
+
+    def __init__(self, scheduler):
+        self.scheduler = scheduler
+        self.leases = defaultdict(deque)
+        self.events = defaultdict(tornado.locks.Event)
```
mrocklin: We recently switched to using `asyncio.Event`s rather than `tornado.locks.Event`s.  Is this easy to change?

distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
...
+            self._started = self.client.scheduler.semaphore_create(
+                name=self.name, max_leases=max_leases
+            )
+        else:
+            self.client.sync(
+                self.client.scheduler.semaphore_create,
+                name=self.name,
+                max_leases=max_leases,
+            )
+            self._started = True
```
mrocklin: Maybe this should be an awaitable like `asyncio.sleep(0)` just in case someone awaits the Semaphore later?

(people sometime mix sync and async APIs)

distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
...
+    def __setstate__(self, state):
+        name, address, max_leases = state
+        try:
+            client = get_client(address)
+        except (AttributeError, AssertionError):
+            client = Client(address, set_as_default=False)
+        self.__init__(name=name, client=client, max_leases=max_leases)
+
+    def close(self):
+        self.client.sync(self.client.scheduler.semaphore_close, name=self.name)
```
mrocklin:
```suggestion
        return self.client.sync(self.client.scheduler.semaphore_close, name=self.name)
```

We need to return the result here.  If we're in async mode then this will be an `asyncio.Future`, and the caller will need to await it.

distributed/tests/test_semaphore.py
```diff
@@ -0,0 +1,156 @@
...
+from distributed.metrics import time
+from distributed.utils_test import cluster, gen_cluster  # noqa F401
+
+import dask
+from dask.distributed import Client
+
+
+@gen_cluster(client=True)
+async def test_semaphore(c, s, a, b):
+    semaphore = await Semaphore(max_leases=2, name="exasol_db")
```
mrocklin: Why the name `exasol_db`?

distributed/tests/test_semaphore.py
```diff
@@ -0,0 +1,156 @@
...
+async def test_acquires_with_timeout(c, s, a, b):
+    sem = await Semaphore(1, "x")
+    assert await sem.acquire(timeout=0.025)
+    assert not await sem.acquire(timeout=0.025)
+    await sem.release()
+    assert await sem.acquire(timeout=0.025)
+    await sem.release()
+
+
+def test_timeout_sync():
```
mrocklin:
```suggestion
def test_timeout_sync(c, s, a, b):
```

I believe that there are pytest fixtures that now replace the cluster/client manual construction.

distributed/tests/test_semaphore.py
```diff
@@ -0,0 +1,156 @@
...
+        with Client(scheduler["address"]) as client:
+
+            def f(x):
+                with Semaphore(name="x"):
+                    client = get_client()
+                    assert client.get_metadata("locked") is False
+                    client.set_metadata("locked", True)
+                    sleep(0.01)
+                    assert client.get_metadata("locked") is True
+                    client.set_metadata("locked", False)
```
mrocklin: Did you want to run this function many times concurrently?

distributed/tests/test_semaphore.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+@gen_cluster(client=True, timeout=20)
+async def test_release_semaphore_after_timeout(c, s, a, b):
+    with dask.config.set(
+        {"distributed.scheduler.locks.lease-validation-interval": "100ms"}
+    ):
+        sem = await Semaphore(name="x", max_leases=2)
+        await sem.acquire()  # leases: 2 - 1 = 1
+        semY = Semaphore(name="y")
```
mrocklin: Did you want to await this one?

distributed/semaphore.py
```diff
@@ -0,0 +1,259 @@
+import uuid
+from collections import defaultdict, deque
+from functools import partial
+import asyncio
+import dask
+import tornado.locks
+import tornado.queues
```
mrocklin: This doesn't appear to be used.  My guess is that you have some excess imports here.

distributed/semaphore.py
fjetter: I believe this edge case is mostly triggered in test scenarios. Non-zero should be fine either way.

distributed/semaphore.py
fjetter: Should be fine to convert to an ordinary function. I believe once there was a version with some waiting code and this was never converted back.

distributed/tests/test_semaphore.py
fjetter: That slipped in for educational purposes, I believe. We should revert this.

Background (other than we're using an Exasol)

We had a discussion about the naming of arguments and where discussing about the `name` argument and whether or not it should be named `resource` since the semaphore is likely used to limit access to a resource

distributed/tests/test_semaphore.py
```diff
@@ -0,0 +1,156 @@
...
+    with cluster() as (scheduler, workers):
+        with Client(scheduler["address"]):
+            s = Semaphore(name="x")
+            # Using the context manager already acquires a lease, so the line below won't be able to acquire another one
+            with s:
+                assert s.acquire(timeout=0.025) is False
+
+
+def test_lock_name_only():
+    with cluster() as (scheduler, workers):
```
fjetter: This could also use the async interface, couldn't it?

distributed/tests/test_semaphore.py
fjetter: Yes, somewhere a line is lost where `f` is called using, e.g. a `client.map`

fjetter:
cc @pacman82

dhirschfeld:
IIUC this closes #864

distributed/semaphore.py
lr4d: `scheduler.semaphore_close` is not implemented

lr4d:
My pleasure, @mrocklin . I'll address the review comments in the upcoming days. Any idea why the tests for Python 3.8 fail on Windows?

distributed/tests/test_semaphore.py
lr4d:
```
  def test_timeout_sync(c, s, a, b):
E       fixture 'c' not found
```

fjetter: Might be that you need to import it first. With fixtures defined in a proper module, the rules are slightly different compared to `conftest` defined fixtures

lr4d: ah my bad

mrocklin: This is in test_client.py

```python
from distributed.utils_test import (  # noqa: F401
    client as c,
    s,
    a,
    b,
)
```

lr4d:
I'm looking into an additional test which is currently failing and I have not yet figured out why.

wrt the Semaphore.close method, I don't see _that_ much value in implementing this at the moment, am I missing something?

distributed/semaphore.py
mrocklin: When we create semaphores on the client side do we create any state on the scheduler side?  Is there a need to clean up that state if the client destroys its semaphores?

If there is no need for a close method, then great, we should probably remove this entire method.  If there is some state that we would like to clean up then we should probably do something here?

lr4d: I'll look into it

distributed/tests/test_semaphore.py
lr4d: I renamed the arguments `id` and `name` to `instance_id` and `resource_name` respectively

mrocklin:
> My pleasure, @mrocklin . I'll address the review comments in the upcoming days. Any idea why the tests for Python 3.8 fail on Windows?

I'm not seeing that currently.  There are occasionally intermittently failing tests.  Note that currently Linux 3.6-3.7 are failing with TLS issues.

> wrt the Semaphore.close method, I don't see that much value in implementing this at the moment, am I missing something?

In many other situations we want to make sure that things get cleaned up.  It might be good to test that whatever collections this system maintains are cleared out if there are no active semaphores.  We just want to make sure that we don't accrue state continuously.

```python
async with Semaphore(name=...):
    ...

assert not scheduler.semaphore_state
```

distributed/tests/test_semaphore.py
mrocklin: Note that variables, locks, and queues all use a `name=` parameter.  If this parameter fulfills a similar role then it would be good to keep the name `name` for consistency.  Similarly Client uses `id` for instance id.  We might want to stick to that as well.

lr4d: I'll revert that for consistency then. I still think it wouldn't hurt to refactor that sometime for easier understanding of the code

distributed/semaphore.py
```diff
@@ -0,0 +1,327 @@
...
+    def release(self, comm=None, name=None, client=None, identifier=None):
+        with log_errors():
+            if not self._semaphore_exists(name):
+                return
+            if isinstance(name, list):
+                name = tuple(name)
+            if name in self.leases and identifier in self.leases[name]:
+                self._release_value(name, client, identifier)
+            else:
+                raise ValueError(
```
lr4d: do we not care if an exception is raised here or would a warning/log message be more appropriate so as to not block further execution? cc @fjetter

lr4d:
> In many other situations we want to make sure that things get cleaned up. It might be good to test that whatever collections this system maintains are cleared out if there are no active semaphores. We just want to make sure that we don't accrue state continuously.

I've implemented a close method and added a line to make sure we remove all state from a client that has been "released". Would appreciate any further feedback

distributed/semaphore.py
mrocklin: This code is triggered by a call from the Client side.  The mechanism that Dask uses to do these sorts of remote calls should serialize the error, send it to the client, and reraise it there.  This would be good to verify though.  Maybe something like the following?

```python
semaphore.acquire()
semaphore.release()

with pytest.raises(ValueError) as info:
    semaphore.release()

assert "already released" in str(info.value)
```

?

distributed/semaphore.py
```diff
@@ -0,0 +1,327 @@
...
+            # not naive:
+            self.leases[name].append(identifier)
+            self.leases_per_client[client][name].append(identifier)
+        else:
+            result = False
+        return result
+
+    def _semaphore_exists(self, name):
+        if name not in self.max_leases:
+            logger.exception(f"Semaphore `{name}` not known or already closed.")
```
lr4d: @mrocklin how about the handling of this error? Context: the semaphore has been closed using `Semaphore.close` and a client attempts to acquire/release/close the closed semaphore. Would the expected behavior be that there is an exception raised?

distributed/semaphore.py
fjetter: I think the question is not about how the exception is handled but rahter should we raise an exception or ignore this?

I'm inclined to say this should **not** be an exception. This path is encountered in two scenarios:
1. Releasing too often
2. Semaphore closed already

I could imagine both scenarios to happen in a distributed system due to weird timings and I wouldn't want my application to crash while cleaning up. There are different philosophies for this, I believe and I don't have a very strong opinion here.

distributed/semaphore.py
```diff
@@ -0,0 +1,327 @@
...
+            return False
+        return True
+
+    async def acquire(
+        self, comm=None, name=None, client=None, timeout=None, identifier=None
+    ):
+        with log_errors():
+            if not self._semaphore_exists(name):
+                return
```
fjetter: 1. Acquire should return a boolean. The boolean is well defined and it is clear what's going on.
2. I would argue that an acquire for a not yet create semaphore should raise

distributed/semaphore.py
fjetter: I would not log or raise here since the behaviour should be depending on where this check is performed. I think it's ok to encapsulate the _exists_ check in a method since what data structure needs to be checked is not trivial

distributed/semaphore.py
lr4d: Would you also argue that it should raise for a semaphore that has already been closed?

distributed/semaphore.py
```diff
@@ -0,0 +1,327 @@
...
+        disconnected processes to coordinate.  If not given, a random
+        name will be generated.
+    client: Client (optional)
+        Client to use for communication with the scheduler.  If not given, the
+        default global client will be used.
+
+    Examples
+    --------
+    >>> sem = Semaphore(max_leases=2, name='my_database')  # doctest: +SKIP
+    >>> sem.acquire(timeout=0.1)  # doctest: +SKIP
```
dhirschfeld: Perhaps the example should advertise the context-manager api as that's generally preferred to manually cleaning up resources.

dhirschfeld:
I'm curious if this works with [`fire_and_forget`](https://docs.dask.org/en/latest/futures.html#fire-and-forget)

In my usage a client will connect, fire off 100 tasks then exit. `fire_and_forget` ensures that even though the client goes away all the submitted tasks will be run. What I'd like is to be able to use the `Semaphore` to also limit the number of tasks running in parallel.

Just wondering if `Semaphore` implements some black magic that makes that possible?

What I do currently is to have the client submit a "manager" task which submits the 100 tasks itself using a `worker_client` and the `as_completed` trick. I imagine that I'd still have to use a similar indirection with `Semaphore`?

distributed/semaphore.py
pacman82: I'm in agreement with @fjetter on this one. The post condition is met and the semaphore released. No need to raise an exception.

distributed/semaphore.py
```diff
@@ -0,0 +1,327 @@
...
+                raise ValueError(
+                    "Inconsistent max leases: %s, expected: %s"
+                    % (max_leases, self.max_leases[name])
+                )
+
+    async def _get_lease(self, client, name, identifier):
+        # We should make sure that the client is already properly registered with the scheduler
+        # otherwise the lease validation will mop up every acquired lease immediately. That's mostly relevant for tests
+        while client not in self.scheduler.clients:
+            await asyncio.sleep(0.0005)  # This value is set somewhat arbitrarily
```
pacman82: What if the client get's deleted during waiting for a pending semaphore (can this happen?). In such a case it might block for the complete duration of a timeout in context of `acquire`.

distributed/semaphore.py
mrocklin: What is the behavior observed by the user/Client?  Is it the same as they would see with a typical Python Semaphore?  If so, then I'm happy.

distributed/semaphore.py
mrocklin: To make this concrete from the Client perspective, is this the situation that we're considering?

```python
# client 1
semaphore = Semaphore(name="my-resource")
# do things with semaphore
semaphore.close()
```

```python
# client 2
semaphore = Semaphore(name="my-resource")
ValueError("Semaphore already closed")
```

If so I agree that this shouldn't occur.  Perhaps this isn't what we're discussing though.  If folks here are able to spell out what the result of this code is that would help in review (sorry, there is a large backlog to go through, so I'm rushing through reviews today)

distributed/semaphore.py
mrocklin: Without knowing too much about the situation here, I would default to whatever the behavior is for the normal Python Semaphore

mrocklin:
> What I'd like is to be able to use the Semaphore to also limit the number of tasks running in parallel.

It won't limit the number of tasks, but it will cause them to block.  Semaphores and `fire_and_forget` don't conflict with each other in any way.  So after this PR you should be able to launch all of your tasks at once, but have all of those tasks wait on the Semaphore.   That should give you the kind of rate limiting that you're looking for without doing any work on the client side.

dhirschfeld:
Thanks, I look forward to trying it out!

distributed/semaphore.py
fjetter: 1. The python semaphore always returns bool
2. The "does not exist" does not exist in the python standard lib (since it's not distributed) so there are no references. I would argue this should raise in that scenario. 

this scenario would happen if

```
# client 1
semaphore = Semaphore(name="my-resource")

# client 2
semaphore = Semaphore(name="my-resource")

# client 1
semaphore.close()

# client 2
semaphore.acquire()
ValueError("Semaphore already closed")
```

as pointed out in the above [thread](https://github.com/dask/distributed/pull/3573/files/17ba039ff95dc84cb054e1d2872172bc1c3950ca#r396113995)

I believe this scenario should properly raise and we should make sure that Semaphore close should only be called by the process which is responsible for scheduling, e.g.

```
sem = Semaphore()
graph = my_dask_comp(sem=sem)
graph.compute()
sem.close()
```

distributed/semaphore.py
lr4d: confirmed and removed

distributed/semaphore.py
lr4d: changed this behavior in https://github.com/dask/distributed/pull/3573/commits/7a519279a2539203f5dd2f18c0506365b6f28e4e

@mrocklin no, it is for the case when any client closes a semaphore and then any client tries to acquire, release or close that semaphore. See the commit I linked for the behavior for each of these operations now. 

Happy to edit that commit one final time if we can get to a consensus

lr4d:
@fjetter Seems reasonable to me. I'm curious if this change could interfere with somebody else's usual method of execution (assuming they have valid reasons for doing so). Would also like to hear others' feedback on this

distributed/semaphore.py
lr4d: Described this in the semaphore docstring under notes

fjetter:
@mrocklin from my POV this is ready to merge. I believe there are a few more things to clean up (e.g. the above changes in interface) but I don't think that should be a blocker since the current implementation is ready for users to experiment with.

There are a few follow up tasks for which I would create github issues to track
* Semaphore close as contextmanager
* Client request timeouts separate to lease timeout
* Semaphore observability using prometheus metrics
* Lock uses semaphore implementation as backend

Is there a best practice of flagging new features as experimental to let users know that 
A) it might still be buggy, don't expect too much 
B) there may be subtle breaking changes on the horizon

distributed/semaphore.py
```diff
@@ -0,0 +1,336 @@
...
+
+    Examples
+    --------
+    >>> from distributed import Semaphore
+    >>> sem = Semaphore(max_leases=2, name='my_database')
+    >>> # This automatically acquires a lease from the semaphore (if available) which will be released when leaving
+    >>> # the context manager
+    >>> with sem:
+    >>>     pass  # use the protected resource
+    >>> sem.close()  # Close resource (clearing any residual state)
```
fjetter:
```suggestion
    >>> def access_ressource(s, sem):
    >>>     # This automatically acquires a lease from the semaphore (if available) which will be
    >>>     # released when leaving the context manager.
    >>>     with sem:
    >>>         pass
    >>>    
    >>> futures = client.map(access_ressource, range(10), sem=sem)
    >>> client.gather(futures)
    >>> # Once done, close the semaphore to clean up the state on scheduler side.
    >>> sem.close()
    
```

This example shows the (recommended) usage where the semaphore is initialised on client side and not in the worker. This usage makes the cleanup more straight forward

mrocklin:
> @mrocklin from my POV this is ready to merge

In that case I'm happy to defer to your judgement here.  I've been quite busy recently and so I haven't had time to give this PR the full attention that it deserves.  Merge at will.

Also, when merging, I recommend the following practice:

-  We use "Squash and merge" for PRs, unless the commit history is particularly concise and valuable (it has been many many months since I did anything other than squash and merge)
-  Similarly, unless the commit history is particularly informative, I recommend removing the list summary that is automatically generated.  Instead, I often copy in the header comment (if it's informative) or write up something short myself

> Is there a best practice of flagging new features as experimental to let users know that
> A) it might still be buggy, don't expect too much
> B) there may be subtle breaking changes on the horizon

The simplest thing is probably to put a note in the docstring.  We don't currently have any formal warning mechanism for this.  

You might also consider a small documentation PR to dask/dask, which holds the user documentation today.  There is a section on coordination primitives that includes Locks and Queues and such where this would fit in well.

PR #3577, opened 2020-03-16
Ability to skip errors in client.compute

whalebot-helmsman:
Let me describe my use case and what I want to achieve. Example gist https://gist.github.com/whalebot-helmsman/a86aa313c550114b096e6a4c1e5da42a

I have a graph of computations, it is very complicated graph. From time to time I want to run not a whole graph, but part of it. To keep same graph building code for different execution paths I have to use `dask.delayed`. I use `client.compute` for graph execution. See https://gist.github.com/whalebot-helmsman/a86aa313c550114b096e6a4c1e5da42a#file-my_pipeline-py-L32-L48 . Problem is my tasks fail from time to time and `client.compute` doesn't support partial failing.

Futures and `client.gather` support partial failures. It is possible to use futures in same manner as `dask.delayed`, but I have to provide code for graph dependency resolution. I see `dask` as DAG computation handler and it's strange to manage graph dependencies in graph building code. See https://gist.github.com/whalebot-helmsman/a86aa313c550114b096e6a4c1e5da42a#file-my_pipeline-py-L51-L65 . 

In this PR I propose to add `errors=skip` for `client.compute` to combine `dask.delayed` non-submitting way of graph building and ability to handle partial fails during execution.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2669,6 +2669,7 @@ class Client(Node):
         fifo_timeout="60s",
         actors=None,
         traverse=True,
+        errors="raise",
         **kwargs
     ):
         """ Compute dask collections on cluster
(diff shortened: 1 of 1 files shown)


mrocklin:
Hi @whalebot-helmsman .  I apologize for the long delay in response.  Have you considered using `dask.persist` instead of compute?  Persist will allow you to get many dask.delayed things out, each of which can succeed or fail independently.  Does that satisfy your needs?

whalebot-helmsman:
Thanks @mrocklin `client.persist` serve my purpose perfectly.

PR #3579, opened 2020-03-16
Import tlz

jakirkham:
Same change as PR ( https://github.com/dask/dask/pull/5997 ) except for Distributed.

cc @jrbourbeau

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -35,7 +35,8 @@ from bokeh.transform import factor_cmap, linear_cmap
 from bokeh.io import curdoc
 import dask
 from dask.utils import format_bytes, key_split
-from toolz import pipe
+from tlz import pipe
+from tlz.curried import map, concat, groupby
 from tornado import escape
 
 try:
@@ -62,11 +63,6 @@ from distributed.utils import log_errors, format_time, parse_timedelta
 from distributed.diagnostics.progress_stream import color_of, progress_quads
 from distributed.diagnostics.graph_layout import GraphLayout
 from distributed.diagnostics.task_stream import TaskStreamPlugin
-
-try:
-    from cytoolz.curried import map, concat, groupby
-except ImportError:
-    from toolz.curried import map, concat, groupby
 
 if dask.config.get("distributed.dashboard.export-tool"):
     from distributed.dashboard.export_tool import ExportTool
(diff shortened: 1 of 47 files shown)


distributed/dashboard/components/scheduler.py
```diff
@@ -35,7 +35,8 @@
 from bokeh.io import curdoc
 import dask
 from dask.utils import format_bytes, key_split
-from toolz import pipe
+from tlz import pipe
+from tlz.curried import map, concat, groupby
```
jrbourbeau: Can use built in `map` function

distributed/protocol/core.py
```diff
@@ -3,10 +3,7 @@
 
 import msgpack
 
-try:
-    from cytoolz import reduce
-except ImportError:
-    from toolz import reduce
+from tlz import reduce
```
jrbourbeau: Can use `reduce` from `functools`

distributed/tests/test_client.py
```diff
@@ -18,7 +18,7 @@
 import zipfile
 
 import pytest
-from toolz import identity, isdistinct, concat, pluck, valmap, partial, first, merge
+from tlz import identity, isdistinct, concat, pluck, valmap, partial, first, merge
```
jrbourbeau: Can use `partial` from `functools`

distributed/worker.py
```diff
@@ -21,10 +21,7 @@
 from dask.utils import format_bytes, funcname
 from dask.system import CPU_COUNT
 
-try:
-    from cytoolz import pluck, partial, merge, first, keymap
-except ImportError:
-    from toolz import pluck, partial, merge, first, keymap
+from tlz import pluck, partial, merge, first, keymap
```
jrbourbeau: Can use `partial` from `functools`

distributed/dashboard/components/scheduler.py
jakirkham: This is curried so idk if that would work here.

jrbourbeau: Ah, yep you're right. I overlooked the `curried` part of this `import`

jakirkham:
Thanks James! Have updated this to use `functools` where possible. Would appreciate another look when you have a chance 🙂

jrbourbeau:
Thanks @jakirkham!

jakirkham:
Thanks for the review 🙂

Issue #3580, opened 2020-03-17
CuPy (De)serialization error 


cjnolet:
I'm encountering the following exception when trying to perform a custom tree-reduce on a set of large cupy objects. I don't get this exception with smaller objects, so I have not been able to reproduce it with the normal pytests. The `n_features` in the `HashingVectorizer` that is used as input to the Naive Bayes pytest, however, can be modified to 8M in order to reproduce this exception. 

```
Traceback (most recent call last):
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/core.py", line 124, in loads
    value = _deserialize(head, fs, deserializers=deserializers)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 255, in deserialize
    deserializers=deserializers,
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 268, in deserialize
    return loads(header, frames)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/cuda.py", line 28, in cuda_loads
    return loads(header, frames)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 608, in deserialize
    v = deserialize(h, f)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 268, in deserialize
    return loads(header, frames)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/cuda.py", line 28, in cuda_loads
    return loads(header, frames)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/cupy.py", line 63, in cuda_deserialize_cupy_ndarray
    frame = PatchedCudaArrayInterface(frame)
  File "/raid/cnolet/miniconda3/envs/cuml_dev_013/lib/python3.7/site-packages/distributed/protocol/cupy.py", line 26, in __init__
    self.__cuda_array_interface__ = ary.__cuda_array_interface__
AttributeError: 'bytes' object has no attribute '__cuda_array_interface__'
[...]

PR #3584, opened 2020-03-17
Don't Split Frames for UCX

quasiben:
PR fixes https://github.com/dask/distributed/issues/3580 .

~Previously, during `merge_frames` dask would call `ensure_bytes` on a list of objects.  This would trigger a conversion to host memory and would invalidate deserialization logic downstream.  This PR fixes this issue by checking if the list is composed of cuda objects.  If it is, simply extend the output list~

Update:
PR adds functionality which allows for frames to not be split during serialization.  Additionally, communication with UCX will no longer split frames

### DIFF
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -15,7 +15,7 @@ _deserialize = deserialize
 logger = logging.getLogger(__name__)
 
 
-def dumps(msg, serializers=None, on_error="message", context=None):
+def dumps(msg, serializers=None, on_error="message", context=None, split_frames=True):
     """ Transform Python message to bytestream suitable for communication """
     try:
         data = {}
@@ -49,7 +49,9 @@ def dumps(msg, serializers=None, on_error="message", context=None):
             if "lengths" not in head:
                 head["lengths"] = tuple(map(nbytes, frames))
             if "compression" not in head:
-                frames = frame_split_size(frames)
+                # splitting frames is not the default behavior for UCX
+                if split_frames:
+                    frames = frame_split_size(frames)
                 if frames:
                     compression, frames = zip(*map(maybe_compress, frames))
                 else:
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -15,6 +15,20 @@ _deserialize = deserialize
 logger = logging.getLogger(__name__)
 
 
+def _split_and_compress(header, frames):
+    """
+    Internal function for splitting and compressing frames
+    """
+
+    frames = frame_split_size(frames)
+    if frames:
+        compression, frames = zip(*map(maybe_compress, frames))
+    else:
+        compression = []
+    header["compression"] = compression
+    return header, frames
+
+
 def dumps(msg, serializers=None, on_error="message", context=None, split_frames=True):
     """ Transform Python message to bytestream suitable for communication """
     try:
@@ -48,15 +62,13 @@ def dumps(msg, serializers=None, on_error="message", context=None, split_frames=
         for key, (head, frames) in data.items():
             if "lengths" not in head:
                 head["lengths"] = tuple(map(nbytes, frames))
-            if "compression" not in head:
-                # splitting frames is not the default behavior for UCX
-                if split_frames:
-                    frames = frame_split_size(frames)
-                if frames:
-                    compression, frames = zip(*map(maybe_compress, frames))
-                else:
-                    compression = []
-                head["compression"] = compression
+            # treat compression of collections homogenously
+            if "is-collection" in head:
+                if "compression" not in head["sub-headers"][0]:
+                    head, frames = _split_and_compress(head, frames)
+            elif "compression" not in head:
+                head, frames = _split_and_compress(head, frames)
+
             head["count"] = len(frames)
             header["headers"][key] = head
             header["keys"].append(key)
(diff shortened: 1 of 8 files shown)


distributed/protocol/utils.py
```diff
@@ -92,7 +92,11 @@ def merge_frames(header, frames):
         if len(L) == 1:  # no work necessary
             out.extend(L)
```
jakirkham: I'm interested to know why we don't run into this case. Are we seeing the frames get split? If so, should we be disabling that? It seems we should still merge them somehow if they are getting split, but if we can avoid the splitting that would be better still.

quasiben: frame splitting is part of regular serialization:
https://github.com/dask/distributed/blob/f2f82c6c2e8d36731cb3fb82fb1f80ea0323358e/distributed/protocol/core.py#L52

mrocklin: +1 to @jakirkham 's comment

Before putting a band-aid on this problem I think that we should first see if we can avoid the situation entirely.

distributed/protocol/tests/test_serialize.py
```diff
@@ -374,3 +374,16 @@ def dumps(f):
 
     msg = {"data": original}
     frames = await to_frames(msg)
+
+
+@pytest.mark.asyncio
+async def test_serialize_no_splitting():
+    cp = pytest.importorskip("cupy")
```
quasiben: I am not super happy that this requires cupy but I was unable to trigger a result with numpy.  Open to suggestions if folks have any

quasiben:
@mrocklin @jakirkham  changed the PR around based on comments to not split frames when using UCX

distributed/protocol/tests/test_serialize.py
jakirkham: Yeah there isn't one really as this is done correctly when using the `"dask"` serializer. We have to use something that uses the `"cuda"` serializer.

cjnolet:
LGTM, verified that Naive Bayes is working as expected with these changes. Thanks again, Ben!

cjnolet:
Also, thank you to @jakirkham for spending the time to help me narrow this down.

distributed/protocol/core.py
```diff
@@ -49,7 +49,9 @@ def dumps(msg, serializers=None, on_error="message", context=None):
             if "lengths" not in head:
                 head["lengths"] = tuple(map(nbytes, frames))
             if "compression" not in head:
-                frames = frame_split_size(frames)
+                # splitting frames is not the default behavior for UCX
+                if split_frames:
+                    frames = frame_split_size(frames)
```
mrocklin: It looks like we only consider splitting frames due to compression.  I wonder if we can make this a bit simpler and avoid the keyword argument if we always include `"compression": False` in the header?

quasiben: That would probably work but there may be a GPU compression story in the future.  I suppose we could handle that then if/when it becomes available

mrocklin: Give me a couple minutes to look things over

mrocklin: I still suggest that we go with setting `compression=False` for now rather than introduce the new keyword argument.  

Tying frame splitting to compression is not entirely clean, but it does happen to be the only reason to split frames today.  In the future if that continues to be the case then we might have compressors register a maximum frame size.  If that doesn't continue to be the case (perhaps some comms also have a maximum size limit) then we will have to rejigger this code regardless.  

However, looking at this a bit more, I notice that `cuda_serialize` already specifies a compression value.  My guess is that this is coming from serializing tuples/lists/dicts of objects, and that we don't move compression values through in that case.  I'll take a look at this later tonight.

mrocklin: Yeah ok, so coupling compression with serialization gets complicated, at least with our current logic.  

https://github.com/dask/distributed/blob/511427b81f599105dc1eb4d2f35fd33aa249b4b8/distributed/protocol/core.py#L51-L57

Currently a serializer can choose to pre-compress data.  If so, it adds the compression used to the header.  Great, we can pass on compression (and thus frame splitting).  The CUDA serializers avoids this by setting `compression=None` claiming "I've already handled compression, don't bother" to downstream.  Great.  If we send a single cupy array or something along then we'll be fine.

However, if we happen to include a few things in a small dict/tuple/list then we lose this information.  The headers for the individual bits get thrown into a subheader, and so we lose direct access to the compression.

https://github.com/dask/distributed/blob/511427b81f599105dc1eb4d2f35fd33aa249b4b8/distributed/protocol/serialize.py#L144-L181

Currently our logic though is "If you specify any compression anywhere, don't do anything".  But really we want to handle this on a frame-by-frame basis.  Otherwise the first time someone sends a combined result of `[numpy_array, cupy_array]` things will get messy.  

I think I'm shaving a yak here, but it could be that the right thing to do here is to rework how we handle compression on a frame by frame basis.  Submitting PR now.

jakirkham: Well the other thing I was thinking about was checking `head` for `"cuda"` not in `"serializer"`, which would also be easy and avoid this.

quasiben:
Sorry, I missed the discussion as I was sorting through the same issues of collections. Looking at https://github.com/dask/distributed/pull/3586 now

quasiben:
closing in favor of #3586

PR #3585, opened 2020-03-17
MAINT: better shows worker memory usage in histogram

### LINKED ISSUE #3582: Dashboard histogram on "memory usage" doesn't have helpful x-ticks
**Describe the bug**

When the distributed has many workers available, the dashboard page `/status` shows a histogram of the memory usage of each workers:

Screen Shot 2020-03-17 at 10 33 18 AM

This graph is not useful to me because it has one tick mark, `0.0`. To see the memory usage of the workers, I have to go to `/workers` and manually parse out the (approximate) maximum memory usage to get an idea of the histogram scale:

<details>
Screen Shot 2020-03-17 at 10 32 27 AM
</details>

**Expected Results**

I'd like to see at least one other tick mark besides `0.0`. I'd be fine with the maximum usage displaying.

This matters to me because I think I had a `KilledWorker` error because of an out-of-memory error.

**Versions**

* distributed: 2.11.0

stsievert:
**What does this PR implement?**
This adds ticks to the worker memory usage histogram.

| Before | After
|:---:|:---:|
| before-hist | after-hist |

**Reference issues/PRs**
This resolves #3582.


stsievert:
Looking at the [documentation for mantissas][1], it appears these are the possible ticks:

* Possible tick values for **old** implementation (all appropriately scaled for KB/MB/GB):
    * [1, 2, 3, 4, 5, ...].
    * [256, 512, 1024, ...]
    * [512, 1024, 1536, ...]
* Possible tick values for **new** implementation (all appropriately scaled for KB/MB/GB):
    * [1, 2, 3, 4, ...]
    * [2, 4, 6, 8, ...]
    * [4, 8, 16, 32, ...]
    * [8, 16, 24, 32, ...]
    * [16, 32, 48, 64, ...]
    * [64, 128, 192, 256, ...]
    * [128, 256, 384, 512, ...]
    * [256, 512, 768, 512, ...]
    * [512, 1024, 1536, 2048, ...]

This same code with `mantissas` scattered over this file. Typically `len(mantissas) in [3, 4]`.

[1]:https://docs.bokeh.org/en/latest/docs/reference/models/tickers.html#bokeh.models.tickers.BasicTicker.mantissas

mrocklin:
Thanks for doing this @stsievert .  Do you have a sense for what the right value we should use is?  Is what you currently have here the best choice do you think?

stsievert:
> Do you have a sense for what the right value we should use is? 

I think the right value is `mantissas=[2**i for i in range(10)]` and `base=1024`. That comes from looking at Bokeh's [adaptive_ticker.coffee][1]. 

<details>

I think a base is chosen, and ticks are generated to be one of the mantissas. If `mantissas=[1, 2, 5]` the ticks chosen from this list

* `[1, 2, 3, ...]`
* `[2, 4, 8, ...]`
*  `[5, 10, 15, ...]`.

This is because the adaptive_ticker.coffee has essentially this code:

``` python
# user specifies mantissas=[1, 2, 5]
candidate_mantissas = [0.5, 1, 2, 5, 10]
errors = np.abs(desired_n_ticks - approximate_ticks(m)) for m in candidate_mantissas)
mantissa = candidate_mantissas[np.argmin(errors)]
# say mantissa=2 is specified.
```

If `mantissas=2` is specified, it looks like ticks [2. 4, 6, 8, ...] will be specified in [abstract_ticker.coffee#L72][2].

</details>

> Is what you currently have here the best choice do you think?

I think it makes sense to have `base=1024` so KB/MB/GB are shown. I think `mantissas=[2**i for i in range(10)]` is a good choice; the other choice is `[1, 2, 5, 10, 20, 50, 100, 200, 500]`. Going up to about 1000 is required because of the KB/MB/GB range.

I can modify this PR soon to fix this in the other classes.

[1]:https://github.com/bokeh/bokeh/blob/b1c64dad030fb31d5a37f8da7d7c1dec26283091/bokehjs/src/coffee/ticking/adaptive_ticker.coffee#L54-L56
[2]:https://github.com/bokeh/bokeh/blob/b1c64dad03/bokehjs/src/coffee/ticking/abstract_ticker.coffee#L72

stsievert:
I've propagated the change in mantissa, typically from [1, 256, 512] to powers of 2. I can't see all the results, but here are two results:

| Before | After |
| :---: | :---: |
| before-current-load | after-current-load |
| before-indiv-memory-by-key | after-indiv-memory-by-key |

For these tickers, `desired_num_ticks=6` by default ([source][1]). I have left that value unchanged.

[1]:https://docs.bokeh.org/en/latest/_modules/bokeh/models/tickers.html#ContinuousTicker

mrocklin:
Awesome.  Thank you so much for propagating this change @stsievert .  Anything left to do here, or should we merge?

stsievert:
Let's merge. I don't see anything else.

mrocklin:
Woot.  Thanks @stsievert

stsievert:
Thanks @mrocklin. I'm looking forward to the next release; I'll use this to better monitor the status/health of my cluster (though I'm not sure it'll be necessary; I've resolved the out-of-memory error).

PR #3591, opened 2020-03-19
Regression in work stealing for blacklisted fast tasks

fjetter:
**This is a regression which dates back to 2.9.1**

Through the introduction of the TaskPrefix code, the work stealing for shuffle operations behaves drastically different (not for the better) which can be tracked back to the `split in fast_tasks` check which fails since the split was a `TaskPrefix` object instead of the string `shuffle-split` in the `fast_tasks` dict

### DIFF
--- a/distributed/tests/test_steal.py
+++ b/distributed/tests/test_steal.py
@@ -1,33 +1,29 @@
 import itertools
-from operator import mul
 import random
 import sys
-from time import sleep
 import weakref
-
-from unittest import mock
-import pytest
-from tlz import sliding_window, concat
-from tornado import gen
+from operator import mul
+from time import sleep
 
 import dask
+import pytest
 from distributed import Nanny, Worker, wait, worker_client
 from distributed.config import config
 from distributed.metrics import time
 from distributed.scheduler import key_split
 from distributed.system import MEMORY_LIMIT
 from distributed.utils_test import (
-    slowinc,
-    slowadd,
-    inc,
-    gen_cluster,
-    slowidentity,
     captured_logger,
+    gen_cluster,
+    inc,
+    nodebug_setup_module,
+    nodebug_teardown_module,
+    slowadd,
+    slowidentity,
+    slowinc,
 )
-from distributed.utils_test import nodebug_setup_module, nodebug_teardown_module
-
-import pytest
-
+from tlz import concat, sliding_window
+from tornado import gen
 
 # Most tests here are timing-dependent
 setup_module = nodebug_setup_module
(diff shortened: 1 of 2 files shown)


fjetter:
I believe build failures are unrelated. There is one failure connected to work stealing on Windows Python3.7 but so far I couldn't reproduce this. Maybe flaky? Are there tests known to be flaky in the `test_stealing.py` module?

fjetter:
I added the prefix.name patch to the line you mentioned but I couldn't add a test for it. In fact, I would argue it is unreachable code since the only place where we populate this dictionary is 
https://github.com/dask/distributed/blob/0d64f3a3c2f72543420b6f2967e8e789ad265a27/distributed/stealing.py#L140
but this line assumes that `ts.processing_on is None` which is an invalid state since all stealing dicts are only populated for tasks in state `processing` and as soon as a task is transitioned away from `processing` we'll remove it from `stealable`.
Is there something I'm missing? Can this be removed? (Other PR, of course)

distributed/tests/test_steal.py
```diff
@@ -145,23 +146,51 @@ def test_steal_related_tasks(e, s, a, b, c):
...
+    with mock.patch.object(
+        distributed.stealing, "fast_tasks", {fast_blacklisted.__name__}
+    ):
+
+        w = workers[0]
+        futures = c.map(fast_blacklisted, range(100), y=x)
+
+        await wait(futures)
+
+    mmock.assert_not_called()
```
mrocklin: This test is highly dependent on the internal API.  Historically we've tried to avoid tests like this, because it makes changing internal code really hard (future developers have to understand and change all tests that depend on the internal API).  Instead, I wonder if you could use something like `c.map(..., workers=[...], allow_other_workers=True)` and then test that none of the other workers got any data.  This only requires that we keep the `worker.data` API, which I think is probably more stable than the `move_task_request` API.

Also, do we need the `timeout=1000` here, or was this just for debugging?

mrocklin:
> but this line assumes that ts.processing_on is None which is an invalid state since all stealing dicts are only populated for tasks in state processing and as soon as a task is transitioned away from processing we'll remove it from stealable.
> Is there something I'm missing? Can this be removed? (Other PR, of course)

If this is true then yes, it would be good to remove this code.  It might be worth checking out `git blame` to see why this was added in the first place, and spot checking the various ways in which tasks can be removed from the `processing` state to make sure that we're not missing any stealable cleanups.

mrocklin:
Thank you for identifying and resolving this issue @fjetter .  I imagine that it was not trivial to track down.

distributed/tests/test_steal.py
fjetter: Thanks, I already suspected this to become an issue during review :)

I didn't know about the `allow_other_workers` flag, will rewrite the test, ofc.

- [ ] No mocks

> Also, do we need the timeout=1000 here, or was this just for debugging?

Yes, that was debugging. No timeout should be necessary here.

- [ ] No timeouts

fjetter:
> If this is true then yes, it would be good to remove this code. It might be worth checking out git blame to see why this was added in the first place

I'll try to investigate the history and come back with what I find.

> I imagine that it was not trivial to track down.

Indeed, this one was uncomfortable. We do not really instrument the work stealing which made me fall back to "watching dashboards", greping logs and pulling scheduler/workerr state with `client.run*`. I this particular case a mere "unique prefixes which were subject to stealing" would've shown the issue already but I suspect this is worthless in most other scenarios. I'm wondering if it's worth collecting things like average steal_time_ratio per task/taskprefix, total count of stolen tasks, most likely victims / thieves, etc. Has this issue ever come up somewhere else?

mrocklin:
Thanks @fjetter . This is in.

PR #3595, opened 2020-03-19
Add SubgraphCallables to serialization cache

jrbourbeau:
Currently when serializing `SubgraphCallable`s in `dumps_function`, because `SubgraphCallable`s aren't hashable, we can't use them as keys in the `cache_dumps` serialization cache. Instead we end up trigger the `TypeError` here

https://github.com/dask/distributed/blob/2acffc3172ec32e173547ee4c39a01b6c94e74a1/distributed/worker.py#L3288-L3289

and must re-serialize the same `SubgraphCallable` each time it is encountered. 

One solution would be to implement a `SubgraphCallable.__hash__` method, however I don't think we can reliably hash `SubgraphCallable`s as they contain a dask graph which in general can contain non-hashable things. 

This PR proposes we place `SubgraphCallable`s in the serialization cache using their `id` as a key. Having the same `SubgraphCallable` instance occur in many tasks is common when working with dask array and dask dataframe collections. This shouldn't noticeably impact workloads that don't use `SubgraphCallable`s, and will decreasing graph processing time for workloads that do.

<details>

For example, with the changes in this PR we have:

```python
In [1]: from toolz import valmap
   ...:
   ...: import dask.array as da
   ...: from distributed.worker import dumps_task
   ...:
   ...: # Do some array computation
   ...: x = da.arange(100_000, chunks=100)
   ...: y = da.mean(x) + x
   ...: z = da.cos(y)
   ...:
   ...: # Optimized graph
   ...: dsk_optimized = z.__dask_optimize__(z.dask, keys=z.__dask_keys__())
   ...:
   ...: # Serialize optimized graph
   ...: %timeit valmap(dumps_task, dsk_optimized)
25 ms ± 793 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

and on the current `master` branch the same code snippet gives:

```
3.67 s ± 311 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

</details>

cc @mrocklin if you get a chance to look at this

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -13,6 +13,7 @@ import asyncio
 
 import dask
 from dask import delayed
+from dask.optimization import SubgraphCallable
 from dask.utils import format_bytes
 from dask.system import CPU_COUNT
 import pytest
@@ -34,7 +35,14 @@ from distributed.compatibility import WINDOWS
 from distributed.core import rpc, CommClosedError
 from distributed.scheduler import Scheduler
 from distributed.metrics import time
-from distributed.worker import Worker, error_message, logger, parse_memory_limit
+from distributed.worker import (
+    Worker,
+    error_message,
+    logger,
+    parse_memory_limit,
+    dumps_function,
+    cache_dumps,
+)
 from distributed.utils import tmpfile, TimeoutError
 from distributed.utils_test import (  # noqa: F401
     cleanup,
@@ -1652,3 +1660,10 @@ async def test_heartbeat_comm_closed(cleanup, monkeypatch, reconnect):
                 else:
                     assert w.status == "closed"
     assert "Heartbeat to scheduler failed" in logger.getvalue()
+
+
+def test_dumps_function_subgraphcallable_cached():
+    dsk = {"x": 1, "y": (slowinc, "x")}
+    func = SubgraphCallable(dsk, outkey="y", inkeys=())
+    dumps_function(func)
+    assert id(func) in cache_dumps
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -42,6 +42,7 @@ from distributed.worker import (
     parse_memory_limit,
     dumps_function,
     cache_dumps,
+    _subgraphcallable_cache_key,
 )
 from distributed.utils import tmpfile, TimeoutError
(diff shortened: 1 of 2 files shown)


mrocklin:
I apologize for the delay in response @jrbourbeau .  Should we concerned about the use of `id` here.  Is it possible for `func` to be gc'ed at some point and lose a handle on that `id` number?

jrbourbeau:
Thanks for taking a look at this, apologies for the delayed response. Good point about `func` getting gc'ed, that situation hadn't occured to me. We'll want to avoid keeping entries in the cache for `SubgraphCallable`s which have been gc'ed as, in theory, a later `SubgraphCallable` could be assigned the same memory location and we'd end up pull the wrong value from the cache. 

A couple of alternatives would be to:

1. Place a reference to the `SubgraphCallable` in cached value which would prevent unwanted garbage collection. So instead of just caching the serialized `SubgraphCallable`, we would cache the tuple `(<serialized-subgraph>, subgraph)`. This would work, but is messy.

2. Use `weakref.finalize` to ensure we remove the `SubgraphCallable` from the cache when it is gc'ed

I have a preference for the second option, so I'll work on that and push up a commit to see what you think

jrbourbeau:
Turns out that one currently can't make weak references to `SubgraphCallable`s as they don't have `"__weakref__"` in their [`"__slots__"` attribute](
https://github.com/dask/dask/blob/5d5b43b4373677ff5fb7cf25eb4bdfb514cf3e76/dask/optimization.py#L956). After exploring what that means for us here (xref https://github.com/dask/distributed/pull/3595/commits/58aa113b7312f6207a98f26611ecdf88846f316f) I think I've changed my mind and prefer option 1.

In the most recent commit we now store a reference to the `SubgraphCallable` object in the cache to prevent the `SubgraphCallable` from being garbage collected prior to eviction from the LRU cache

mrocklin:
> because SubgraphCallables aren't hashable

Alternatively, can we make SubgraphCallables hashable?

andrewfulton9:
I made a [PR](https://github.com/dask/dask/pull/6424) in dask related to this PR that makes `SubgraphCallable` hashable.

jrbourbeau:
Thanks @andrewfulton9! Closing in favor of https://github.com/dask/dask/pull/6424

Issue #2986, opened 2019-08-22
Add a document "Why are my workers dying?"


mrocklin:
This is a common request.  We might add a document that includes many of the common causes, and then add a link to that document in a log message somewhere when we find that a worker has died unexpectedly.

mUtterberg:
@mrocklin’s response on [this Stack Overflow post](https://stackoverflow.com/questions/46691675/what-do-killedworker-exceptions-mean-in-dask) might be a good place to start.
I’m not sure what the common causes might be, as I’m currently working my way through a KilledWorker exception in my workflow and haven’t gotten to the bottom of it yet.

PR #3596, opened 2020-03-19
Write "why killed" docs [skip ci]

martindurant:
Fixes https://github.com/dask/distributed/issues/2986

### DIFF
--- a/docs/source/killed.rst
+++ b/docs/source/killed.rst
@@ -1,0 +1,105 @@
+.. _killed:
+
+Why did my worker die?
+----------------------
+
+A Dask worker can cease functioning for a number of reasons. These fall into the
+following categories:
+
+- the worker chooses to exit
+- an unrecoverable exception happens within the worker
+- the worker process is shut down by some external action
+
+Each of these cases will be described in more detail below. The *symptoms* you will
+experience when these things happen range from simply work not getting done anymore,
+to various exceptions appearing when you interact with your local client, such as
+``KilledWorker``, ``TimeoutError`` and ``CommClosedError``.
+
+Note the special case of ``KilledWorker``: this means that a particular task was
+tried on a worker, and it died, and then the same task was sent to another worker,
+which also died. After a configurable number of deaths, Dask decides to blame the
+task itself, and returns this exception.
+
+In every case, the first place to look for further information is the logs of
+the given worker, which may well give a complete description of what happened. These
+logs are printed by the worker to it's "standard output", which may be the text
+console from which you launched the worker, or some logging system maintained by
+the cluster infrastructure.
+
+In all cases, the scheduler will notice that the worker has gone, either because
+of an explicit de-registration, or because the worker no longer produces heartbeats,
+and it should be possible to reroute tasks to other workers and have the system
+keep running.
+
+Worker chose to exit
+''''''''''''''''''''
+
+Workers may exit in normal functioning because they have been asked to, e.g.,
+they received a keyboard interrupt (^C), or the scheduler scaled down the cluster.
+In such cases, the work that was being done by the worker will be redirected to
+other workers, if there are any left.
+
+You should expect to see the following message at the end of the worker's log:
+
+.. pull-quote::
+
+   distributed.dask_worker - INFO - End worker
+
+In these cases, there is not normally anything which you need to do, since the
+behaviour is expected.
+
+Unrecoverable Exception
+'''''''''''''''''''''''
+
+The worker is a python process, and like any other code, an exception may occur
+which causes the process to exit. One typical example of this might be a
+version mismatch between the packages of the client and worker, so that
+a message sent to the worker errors while being unpacked. There are a number of
+packages that need to match, not only ``dask`` and ``distributed``.
+
+In this case, you should expect to see the full python traceback in the worker's
+log. In the event of a version mismatch, this might be complaining about a bad
+import or missing attribute. However, other fatal exceptions are also possible,
+such as trying to allocate more memory than the system has available, or writing
+temporary files without appropriate permissions.
+
+To assure that you have matching versions, you should run (more recent versions
+of distributed may do this automatically)
+
+.. code-block::
+
+   client.get_versions(check=True)
+
+For other errors, you might want to run the computation in your local client, if
+possible, or try grabbing just the task that errored and using
+:func:`recreate_error_locally <distributed.recreate_exceptions.ReplayExceptionClient.recreate_error_locally>`,
+as you would for ordinary exceptions happening during task execution.
+
+Specifically for connectivity problems (e.g., timeout exceptions in the worker
+logs), you will need to diagnose your networking infrastructure, which is more
+complicated than can be described here. Commonly, it may involve logging into
+the machine running the affected worker
+(although you can :ref:`ipylaunch`).
+
+Sudden Exit
+'''''''''''
+
+The worker process may stop working without notice. This can happen due to
+something internal to the worker, e.g., a memory violation (common if interfacing
+with compiled code), or due to something external, e.g., the ``kill`` command, or
+stopping of the container or machine on which the worker is running.
+
+In the best case, you may have a line in the logs from the OS saying that the
+worker was shut down, such as the single word "killed"  or something more descriptive.
+In these cases, the fault may well be in your code, and you might be able to use the
+same debugging tools as in the previous section.
+
+However, if the action was initiated by some outside framework, then the worker will
+have no time to leave a logging message, and the death *may* have nothing to do with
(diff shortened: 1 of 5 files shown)


danielballan:
I recently went looking for something like this, and this would have helped me. 👍

martindurant:
@mrocklin @TomAugspurger @quasiben - any thoughts here?

martindurant:
(or anyone else...)

mrocklin:
I plan to take a look at this later tonight.

docs/source/killed.rst
```diff
@@ -0,0 +1,105 @@
...
+- the worker process is shut down by some external action
+
+Each of these cases will be described in more detail below. The *symptoms* you will
+experience when these things happen range from simply work not getting done anymore,
+to various exceptions appearing when you interact with your local client, such as
+``KilledWorker``, ``TimeoutError`` and ``CommClosedError``.
+
+Note the special case of ``KilledWorker``: this means that a particular task was
+tried on a worker, and it died, and then the same task was sent to another worker,
+which also died. After a configurable number of deaths, Dask decides to blame the
```
TomAugspurger: Mention the config key itself.

docs/source/killed.rst
```diff
@@ -0,0 +1,105 @@
...
+''''''''''''''''''''
+
+Workers may exit in normal functioning because they have been asked to, e.g.,
+they received a keyboard interrupt (^C), or the scheduler scaled down the cluster.
+In such cases, the work that was being done by the worker will be redirected to
+other workers, if there are any left.
+
+You should expect to see the following message at the end of the worker's log:
+
+.. pull-quote::
```
TomAugspurger: Does this render as fixed-width text?

docs/source/killed.rst
```diff
@@ -0,0 +1,105 @@
...
+   distributed.dask_worker - INFO - End worker
+
+In these cases, there is not normally anything which you need to do, since the
+behaviour is expected.
+
+Unrecoverable Exception
+'''''''''''''''''''''''
+
+The worker is a python process, and like any other code, an exception may occur
+which causes the process to exit. One typical example of this might be a
```
TomAugspurger: Is version mismatch likely now that we have version checking enabled on startup?

docs/source/killed.rst
```diff
@@ -0,0 +1,105 @@
...
+``KilledWorker``, ``TimeoutError`` and ``CommClosedError``.
+
+Note the special case of ``KilledWorker``: this means that a particular task was
+tried on a worker, and it died, and then the same task was sent to another worker,
+which also died. After a configurable number of deaths, Dask decides to blame the
+task itself, and returns this exception.
+
+In every case, the first place to look for further information is the logs of
+the given worker, which may well give a complete description of what happened. These
+logs are printed by the worker to it's "standard output", which may be the text
```
danielballan:
```suggestion
logs are printed by the worker to its "standard output", which may be the text
```

Also, does it go to stderr, typically? Maybe the distinction is not important here.

docs/source/killed.rst
```diff
@@ -0,0 +1,105 @@
...
+For other errors, you might want to run the computation in your local client, if
+possible, or try grabbing just the task that errored and using
+:func:`recreate_error_locally <distributed.recreate_exceptions.ReplayExceptionClient.recreate_error_locally>`,
+as you would for ordinary exceptions happening during task execution.
+
+Specifically for connectivity problems (e.g., timeout exceptions in the worker
+logs), you will need to diagnose your networking infrastructure, which is more
+complicated than can be described here. Commonly, it may involve logging into
+the machine running the affected worker
+(although you can :ref:`ipylaunch`).
```
mrocklin: I like all of the situations that you've brought up here (versions, running out of memory, network connectivity.   I think that it might make sense to enumerate these into their own small subsections and then include them in the TOC.  I think that this might also make it easier to extend this document in the future as we remember more ways in which workers die.  It will be easier, I think, to extend this list, than to add other cases into the narrative flow that you have here (although your flow reads more nicely than a list).

docs/source/killed.rst
```diff
@@ -0,0 +1,105 @@
...
+
+In the best case, you may have a line in the logs from the OS saying that the
+worker was shut down, such as the single word "killed"  or something more descriptive.
+In these cases, the fault may well be in your code, and you might be able to use the
+same debugging tools as in the previous section.
+
+However, if the action was initiated by some outside framework, then the worker will
+have no time to leave a logging message, and the death *may* have nothing to do with
+what the worker was doing at the time. For example, if kubernetes decides to evict a
+pod, or your ec2 instance goes down for maintenance, the worker is not at fault.
```
mrocklin: I think that these typically leave a kill signal that often has as much or more information than a segfault

mrocklin:
Thanks for writing this up @martindurant .  I like the content and the flow.  This is good high level narrative documentation that should help people understand the general structure of why things can fail.  

I'm curious though, if it makes sense to pull out the various kinds of failures and their remediation into a list that follows the high level explanation (I wrote about this in a comment above).  I think that in many cases users will be coming to this document frustrated because things aren't working, and they'll want to quickly find the situation that matches their experience, and read about how to resolve (or at least understand) that experience.  To that end, I wonder if we might pull some of the specific content that you have here out into a following section, and establish some structure like the following:

```
## Network Connectivity

### How it presents

Often we see messages that say that workers can't connect

... example of worker not connecting

Or we get messages like the following

... another example

### Understanding what is going on

The Dask workers need to be able to talk to the Dask scheduler.  They do this over  ....

### How to resolve the situation

The most common solution to this is to specify a network interface that can see where you've placed the Dask scheduler

    dask-worker ... --interface ib0

But this varies widely based on how your system is set up.  Your system administrator may be able to help you, especially if you connect them to this useful documentation page ...
```

In particular I want to stress the pattern of "How it presents", "Understanding what is going on", and "How to resolve the situation".  I think that if we start something structured like this then people will be comfortable adding to it in the future.

Thoughts?

docs/source/killed.rst
martindurant: Yes I think so, in the case that they don't have the latest versions anyway (which is pretty common)

martindurant:
I added one "specific case" here, but I remain unconvinced. I don't seem to be able to make this concise, since many exceptions will partially overlap: `TimeoutError`? `gaia.error`? SSL handshake error? Can they all be written out long-hand, even though many of the solutions will include the same diagnosis and remediation options?

docs/source/killed.rst
```diff
@@ -0,0 +1,142 @@
...
+- the worker chooses to exit
+- an unrecoverable exception happens within the worker
+- the worker process is shut down by some external action
+
+Each of these cases will be described in more detail below. The *symptoms* you will
+experience when these things happen range from simply work not getting done anymore,
+to various exceptions appearing when you interact with your local client, such as
+``KilledWorker``, ``TimeoutError`` and ``CommClosedError``.
+
+Note the special case of ``KilledWorker``: this means that a particular task was
```
StephanErb: The Dask scheduler does not differentiate between different task states on a worker. So, if there are 1000 tasks on a worker and only one is executed right now, then a segfault or out of memory will still lead to all 1000 tasks being suspected and having their error counter incremented. 

In practice this can lead to situations where an unlucky but totally innocent tasks first crosses the `allowed-failure` threshold even though other tasks are actually responsible. If you then try to investigate the `KilledWorker` exception you end up looking at the wrong task

Even though this is an implementation detail, this can make debugging quite difficult and might be worth mentioning here.

martindurant: I am surprised - does the worker not notify the scheduler which task it's about to process?

docs/source/killed.rst
```diff
@@ -0,0 +1,142 @@
...
+:func:`recreate_error_locally <distributed.recreate_exceptions.ReplayExceptionClient.recreate_error_locally>`,
+as you would for ordinary exceptions happening during task execution.
+
+Specifically for connectivity problems (e.g., timeout exceptions in the worker
+logs), you will need to diagnose your networking infrastructure, which is more
+complicated than can be described here. Commonly, it may involve logging into
+the machine running the affected worker
+(although you can :ref:`ipylaunch`).
+
+Sudden Exit
```
StephanErb: I think it might also make sense to at least mention the default Nanny behavior to kill workers using too much memory.

We could even recommend for anyone running a worker per container to disable the Nanny worker termination entirely, so that one can fully rely on the logging of the orchestration system (e.g. Kubernetes)

martindurant: Indeed I have missed that, thanks.
Do you happen to know what logging text is produced by the nanny when this happens? Should link to https://distributed.dask.org/en/latest/worker.html#memory-management

martindurant:
OK, everyone?

PR #3597, opened 2020-03-19
Change docs to reference adaptive_target

jsignell:
Looks like the original work was done in #2735 but these docs weren't changed.


mrocklin:
Thanks @jsignell !

PR #3598, opened 2020-03-19
Synchronize default CUDA stream before UCX send/recv

pentschev:
This is necessary to ensure CUDA buffers are in the correct state before UCX transfers data.

cc @jakirkham @quasiben @kkraus14 @jrhemstad @harrism @Akshay-Venkatesh

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -170,7 +170,11 @@ class UCX(Comm):
                 )
                 # Send frames
 
-                # It is necessary to first synchronize the device before start sending
+                # It is necessary to first synchronize the default stream before start sending
+                # We synchronize the default stream because UCX is not stream-ordered and
+                #  syncing the default stream will wait for other non-blocking CUDA streams.
+                # Note this is only sufficient if the memory being sent is not currently in use on
+                # non-blocking CUDA streams. 
                 synchronize_stream(0)
 
                 for frame in frames:
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -221,7 +221,7 @@ class UCX(Comm):
                             frames.append(b"")
 
                 # It is necessary to first populate `frames` with CUDA arrays and synchronize
-                # the device before starting receiving to ensure buffers have been allocated
+                # the default stream before starting receiving to ensure buffers have been allocated
                 synchronize_stream(0)
                 for i, (is_cuda, size) in enumerate(zip(is_cudas.tolist(), sizes.tolist())):
                     if size > 0:
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -35,6 +35,15 @@ ucp = None
 cuda_array = None
 
 
+def synchronize_stream(stream=0):
+    import numba.cuda
+
+    ctx = numba.cuda.current_context()
+    cu_stream = numba.cuda.driver.drvapi.cu_stream(stream)
+    stream = numba.cuda.driver.Stream(ctx, cu_stream, None)
+    stream.synchronize()
+
+
 def init_once():
     global ucp, cuda_array
     if ucp is not None:
@@ -160,6 +169,10 @@ class UCX(Comm):
                     np.array([nbytes(f) for f in frames], dtype=np.uint64)
                 )
                 # Send frames
+
+                # It is necessary to first synchronize the device before start sending
+                synchronize_stream(0)
+
                 for frame in frames:
                     if nbytes(frame) > 0:
                         await self.ep.send(frame)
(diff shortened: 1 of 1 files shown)


distributed/comm/ucx.py
```diff
@@ -35,6 +35,15 @@
 cuda_array = None
 
 
+def synchronize_stream(stream=0):
+    import numba.cuda
+
+    ctx = numba.cuda.current_context()
```
harrism: Wow, that's a lot of work to synchronize a stream you already have.... Is this necessary? Can you just call the CUDA driver API directly?

pentschev:
> Changes look good to me. We should do some profiling after this to see if we're losing a lot of time in Numba getting the context in which case we can add an RMM synchronization path if needed.

If I did nothing wrong, there seems to be no overhead with that, other than the synchronization itself. Running a full workflow with `nsys` doesn't present any context related calls such as `cuCtxGetCurrent` or `cuCtxCreate`. Also below is a simple code I wrote and the CUDA API calls presented by nsys.

```python
def synchronize_stream(stream=0):
    import numba.cuda

    ctx = numba.cuda.current_context()
    cu_stream = numba.cuda.driver.drvapi.cu_stream(stream)
    stream = numba.cuda.driver.Stream(ctx, cu_stream, None)
    stream.synchronize()


for i in range(3):
    synchronize_stream(0)
```

```
CUDA API Statistics (nanoseconds)

Time(%)      Total Time       Calls         Average         Minimum         Maximum  Name
-------  --------------  ----------  --------------  --------------  --------------  --------------------------------------------------------------------------------
   99.2         1930480           1       1930480.0         1930480         1930480  cuMemGetInfo_v2
    0.8           15530           3          5176.7            2225           10240  cuStreamSynchronize
```

distributed/comm/ucx.py
```diff
@@ -160,6 +169,10 @@ def peer_address(self) -> str:
                     np.array([nbytes(f) for f in frames], dtype=np.uint64)
                 )
                 # Send frames
+
+                # It is necessary to first synchronize the device before start sending
```
harrism: Maybe it's important to be very clear here:

```suggestion
                # It is necessary to first synchronize the default stream before start sending
                # We synchronize the default stream because UCX is not stream-ordered and
                #  the syncing the default stream will wait for other non-blocking CUDA streams.
                # Note this is only sufficient if the data being sent is not currently in use on
                # non-blocking CUDA streams. 
```

distributed/comm/ucx.py
jakirkham: That's probably more complex than this change. Reason being we need the logic to find that library and load it. Numba is helping us bypass that step.

distributed/comm/ucx.py
```diff
@@ -196,13 +209,20 @@ def peer_address(self) -> str:
...
-                        await self.ep.recv(frame)
                         frames.append(frame)
                     else:
                         if is_cuda:
                             frames.append(cuda_array(size))
                         else:
                             frames.append(b"")
+
+                # It is necessary to first populate `frames` with CUDA arrays and synchronize
+                # the device before starting receiving to ensure buffers have been allocated
```
harrism:
```suggestion
                # the default stream before starting receiving to ensure buffers have been allocated
```

distributed/comm/ucx.py
pentschev: In Numba 0.49.0, there's https://numba.pydata.org/numba-doc/dev/cuda-reference/host.html#numba.cuda.default_stream, but not on earlier versions unfortunately. So the code here is basically doing what the new `numba.cuda.default_stream` function does, plus synchronizing that stream.

How would I call the CUDA driver API directly from Dask with less work than that, and also not depending on some more lower-level libraries? Do you know of a way to do so?

kkraus14: > How would I call the CUDA driver API directly from Dask with less work than that, and also not depending on some more lower-level libraries? Do you know of a way to do so?

You could technically do it via ctypes but then you'd need to reimplement the status code checking and whatnot where it's absolutely not worth it.

If this has problematic performance we can add something in rmm Python to deal with it.

jakirkham: It's also usually a chore getting ctypes to load the right library. Agree it would be better to avoid ctypes if possible.

Akshay-Venkatesh:
Maybe I missed this in the thread somewhere but does this change actually fix the unsafe accesses we've seen?

distributed/comm/ucx.py
pentschev: Yeah, in that case it sounds more convoluted than what's here currently. It doesn't seem problematic performance-wise as of now, so I would prefer that we get this in to unblock people asap, and do such improvements later if they're really needed.

pentschev:
> Maybe I missed this in the thread somewhere but does this change actually fix the unsafe accesses we've seen?

It fixes the segfaults related to synchronization, allowing CUDA IPC (NVLink) and IB (with RDMA disabled) to work again without modifications to UCX core:. IB with RDMA segfaults still remain, but I don't think anymore that issue is related to synchronization.

distributed/comm/ucx.py
harrism: OK, I was just asking.

jakirkham:
FWIW the CI failure is the same unrelated certificate issue shown in PR ( https://github.com/dask/distributed/pull/3587 ). I think we are safe to ignore that.

jakirkham:
Confirmed by the team in test runs as well.

Thanks all! 😄

pentschev:
Thanks for all reviews and @jakirkham for merging as well!

PR #3599, opened 2020-03-20
TCP simplify receiving frames in `comm`s

jakirkham:
Simplifies the logic for receiving frames in `tcp` and `ucx`. Drops some legacy code on the `tcp` side. Also preallocates buffers to receive data into beforehand (and subselects the non-0-length buffers). Then hands off control to the underlying library receiving data to fill these buffers. Afterwards handles any sanity checks.

cc @quasiben @pentschev


pentschev:
LGTM @jakirkham , thanks for the work! I did some testing (not very heavy though) and it doesn't seem to create any issues. On the other hand it also doesn't seem to improve performance, but I'm guessing we will need some larger workflows to be tested to see that, which I didn't do.

quasiben:
Will this PR supersede https://github.com/dask/distributed/pull/3453 ?   If not, is it a step in that direction ?

This looks good to be.  This PR also makes me really appreciate CI due to the tcp changes I am not as familiar with.  

It would be good to get comments from @mrocklin or @TomAugspurger before merging in

jakirkham:
It's a preliminary step in that direction.

Agreed that's why I figured we could try the same thing in `tcp` and `ucx` cases.

Yeah if Matt and Tom have time to review, that would be welcome 🙂

FWIW I also tried to combining the `async` recv calls into an `asyncio.gather` call (instead of having a `for`-loop). However it seems in the `tcp` case Tornado isn't able to handle overlapping receives, so we see some test failures (maybe other people know how to make this work? 😉). In either case we may still consider it for the `ucx` case.

jakirkham:
cc @jrbourbeau (in case you have thoughts here as well 😉)

jakirkham:
FWIW the UCX tests pass for me locally except for issue ( https://github.com/rapidsai/ucx-py/issues/461 ), which I run into with or without these changes. The latter may be a hardware configuration though.

mrocklin:
> FWIW I also tried to combining the async recv calls into an asyncio.gather call (instead of having a for-loop). However it seems in the tcp case Tornado isn't able to handle overlapping receives, so we see some test failures (maybe other people know how to make this work? wink). In either case we may still consider it for the ucx case

Yeah, this is actually something that we want to handle sequentially.  We want to get one frame, and then only once that's done get the next one.  Otherwise we have lots of coroutines all trying to pull bytes off a wire at the same time, which is unwise.

distributed/comm/tcp.py
```diff
@@ -190,18 +190,12 @@ def peer_address(self):
...
-                        frame = await stream.read_bytes(length)
-                else:
-                    frame = b""
-                frames.append(frame)
+            frames = [bytearray(each_length) for each_length in lengths]
+            recv_frames = [each_frame for each_frame in frames if len(each_frame) > 0]
+            for each_frame in recv_frames:
+                each_length = len(each_frame)
+                n = await stream.read_into(each_frame)
+                assert n == each_length, (n, each_length)
```
mrocklin: It looks like this change has a semantic difference in that previously we used to include empty frames but now we don't.  Is this correct?

Is there a specific reason for this change?  Does it affect performance in some way, or is it strictly cosmetic?  If it's strictly cosmetic, and if there is a change to semantics (however minor) I'm tempted to avoid the change just because I wouldn't be surprised if it has some unforeseen effect.

jakirkham: That's not true actually. We [excluded them before]( https://github.com/dask/distributed/blob/e928cc0090015a7e07c18d2d64255d33a849224a/distributed/comm/tcp.py#L195 ) as well. It's just now done outside of the `for`-loop. 🙂

The idea was to more-or-less pass control to Tornado or UCX-Py completely while receiving frames without all of the boilerplate in Dask between each receive. Tried `asyncio.gather` to push this even further, but Tornado doesn't seem to like that.

In any event I thought it would be a nice thing to do the same thing we are doing in UCX for TCP. At the end of the day, I don't have strong feelings about it (especially if there is push back). 🤷‍♂

jakirkham: Any other thoughts here, @mrocklin? 🙂

jakirkham:
We've broken out the UCX part and integrated it separately with PR ( https://github.com/dask/distributed/pull/3651 ). This is now the change for TCP only. Would be good to discuss again when you have time Matt, but it's fine to sit in the interim.

mrocklin:
Pass from me.  You'll have to get consensus from other maintainers.

jakirkham:
Other changes have gone in that more-or-less handle what this was set out to do. Closing...

Issue #3588, opened 2020-03-18
TLS CI failures on Linux Python 3.6 and 3.7


mrocklin:
So TLS is reliably failing on Python 3.6 and 3.7 on Linux.  I have a CI testing PR up here: https://github.com/dask/distributed/pull/3587

Failing Travis CI run: https://travis-ci.org/github/dask/distributed/jobs/664007369?utm_medium=notification&utm_source=github_status

cc @mariusvniekerk @jcrist @jacobtomlinson if any of you have a moment

<details>

<summary> Tracebacks (long) </summary>

```
_________________________ test_tls_reject_certificate __________________________

addr = 'tls://10.20.0.241:40676', timeout = 2, deserialize = True

connection_args = {'ssl_context': <ssl.SSLContext object at 0x7ff0bfc34ca8>}

    async def connect(addr, timeout=None, deserialize=True, connection_args=None):

        """

        Connect to the given address (a URI such as ``tcp://127.0.0.1:1234``)

        and yield a ``Comm`` object.  If the connection attempt fails, it is

        retried until the *timeout* is expired.

        """

        if timeout is None:

            timeout = dask.config.get("distributed.comm.timeouts.connect")

        timeout = parse_timedelta(timeout, default="seconds")

    

        scheme, loc = parse_address(addr)

        backend = registry.get_backend(scheme)

        connector = backend.get_connector()

        comm = None

    

        start = time()

        deadline = start + timeout

        error = None

    

        def _raise(error):

            error = error or "connect() didn't finish in time"

            msg = "Timed out trying to connect to %r after %s s: %s" % (

                addr,

                timeout,

                error,

            )

            raise IOError(msg)

    

        # This starts a thread

        while True:

            try:

                while deadline - time() > 0:

                    future = connector.connect(

                        loc, deserialize=deserialize, **(connection_args or {})

                    )

                    with ignoring(TimeoutError):

                        comm = await asyncio.wait_for(

                            future, timeout=min(deadline - time(), 1)

                        )

                        break

                if not comm:

>                   _raise(error)

distributed/comm/core.py:218: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #3601, opened 2020-03-20
Add ndg-httpsclient as a dependency

jacobtomlinson:
See #3588


jacobtomlinson:
This seems to have caused other CI issues 😕

PR #3602, opened 2020-03-20
Pin openssl to 1.1.1d for Travis

jacobtomlinson:
See #3588

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -42,7 +42,6 @@ conda install -c conda-forge -q \
     jupyter_client \
     msgpack-python>=0.6.0 \
     netcdf4 \
-    openssl==1.1.1d \
     paramiko \
     prometheus_client \
     psutil \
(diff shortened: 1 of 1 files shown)


jrbourbeau:
It looks like installing `libunwind` was causing `openssl` to be upgraded from `1.1.1d` to `1.1.1e` (see https://travis-ci.org/github/dask/distributed/jobs/665056582#L630-L657). I've moved the `openssl=1.1.1d` pinning to the end of our installation steps to make sure we're definitely pinning `openssl`

jrbourbeau:
[Huzzah](https://travis-ci.org/github/dask/distributed/builds/665056580) 🎉

mrocklin:
Oooh, nice green check marks :)

mrocklin:
I'm going to go ahead and merge this short term.

Does this issue signal something larger that we need to be concerned with long term?

jacobtomlinson:
Ah thanks for spotting this @jrbourbeau!

@mrocklin yes this doesn't seem like a great solution, but perhaps we should wait for `openssl` `1.1.1f` and see if the problem persists before being too concerned.

PR #3611, opened 2020-03-20
Make Listeners awaitable

mrocklin:


jrbourbeau:
It appears the `test_recompute_released_key` failure has [shown up before](https://github.com/dask/distributed/pull/3483#issuecomment-586657326). I opened up #3616 to track it

PR #3615, opened 2020-03-21
Replace ncores with nthreads in work stealing tests

jrbourbeau:
This PR resolves some of the warnings that are raised when running our test suite


mrocklin:
Is this good to go @jrbourbeau ?

jrbourbeau:
Reverted some of the changes here as they're taken care of in #3662 and #3653

PR #3619, opened 2020-03-23
Remove dead stealing code

fjetter:
This code piece seems to be unreachable since `ts.processing_on` to be `None` is only possible for tasks with a different state than `processing` but tasks with a different state should not be `stealable`. The dict `stealable_unknown_durations` should therefore never be populated.

Tracing back the git history I found the following commit

https://github.com/dask/distributed/commit/840468462127f421844195b67f1da755a27b8117

which refactored the stealing code to use the taskprefix containers. There, the logic for the dict population changed semantically from catching a key error to checking for the existence of the worker state object. I would argue that my initial argument about the unreachability still holds.

before this, we can only see one commit which factors the work steal code out into an extension. Unfortunately, this commit also seemed to rewrite a huge piece of the logic which is why there is no 1:1 mapping available. Browsing the original scheduler based work stealing code, we can find a variable `stealable_unknown_durations` but it wasn't used anywhere
https://github.com/dask/distributed/blob/e866e11936cc5c2e5bb5af48a529c93236043a71/distributed/scheduler.py

ref #3591

### DIFF
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -29,8 +29,6 @@ class WorkStealing(SchedulerPlugin):
         self.stealable = dict()
         # { task state: (worker, level) }
         self.key_stealable = dict()
-        # { prefix: { task states } }
-        self.stealable_unknown_durations = defaultdict(set)
 
         self.cost_multipliers = [1 + 2 ** (i - 6) for i in range(15)]
         self.cost_multipliers[0] = 1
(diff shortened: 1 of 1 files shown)


mrocklin:
Hrm, I'm surprised by this, but not incredibly so.  Tests pass and this reduces complexity, so I'm happy to merge.  If this ends up having negative effects can I depend on you to fix this @fjetter ?

fjetter:
The good thing about this change is that if it goes south we should see it as an `AttributeError: 'NoneType' object has no attribute 'processing'` exception since for this sections to have any effect the code first needs to pass the `ws is None` guard. It should not show up as some weird, hard to track down performance degradation and I'm fairly confident that we should be able to track such an issue down easily to this change, i.e. if that's the case you may ping me and I'll take care of it :)

mrocklin:
Great.  Merging in.  Thanks for cleaning things up @fjetter !

PR #3621, opened 2020-03-23
Add link to contributing.md

prasunanand:


mrocklin:
The docs.dask.org page is kept more up to date than the distributed.docs.org
page.  It might be worthwhile including both.

jrbourbeau:
Sure, that's a fair point. I don't want to overly nitpick, but it may be useful to be clear that there are development docs for both projects. Something like:

For general information on how to contribute see the [dask development guidelines](https://docs.dask.org/en/latest/develop.html) and [distributed development guidelines](https://distributed.dask.org/en/latest/develop.html).

prasunanand:
I followed `dask-ml` to add the info. https://github.com/dask/dask-ml/blob/master/CONTRIBUTING.md

jrbourbeau:
Alright, this is in

PR #3622, opened 2020-03-23
Add linux CI for GitHub Actions

jacobtomlinson:
Adding a GitHub Actions linux CI as a result of the Travis openssl debugging we've been doing in #3588 

Curious to see if I can reproduce the same issues on GitHub Actions with openssl `1.1.1e`.

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -40,11 +40,13 @@ jobs:
           auto-activate-base: false
 
       - name: Setup SSH
+        shell: bash -l {0}
         run: |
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
The errors here appear to be related to IPv6.

mrocklin:
What is the status here?  Is this still live or should it be closed?

jacobtomlinson:
I would be keen generally to try and unify CI on Windows/Linux/Mac and GitHub actions would be a good way to do it.

This PR may not be the best implementation of that. There is no reason why the same CI script can't be used an all operating systems.

If this is still desirable I'm happy to pick it up again.

jacobtomlinson:
Closing in favour of #4218

PR #3625, opened 2020-03-23
Unpin openssl and handle exception

jacobtomlinson:
Instead of pinning the version of openssl this PR catching the broad `ssl.SSLError` and if it contains the new `unexpected eof while reading` message from `1.1.1e` it converts that to a `StreamClosedError` and treats it like that.

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -79,9 +79,6 @@ if [[ $CRICK == true ]]; then
     python -m pip install -q git+https://github.com/jcrist/crick.git
 fi;
 
-# Pin openssl==1.1.1d (see https://github.com/dask/distributed/issues/3588)
-conda install -c conda-forge openssl==1.1.1d
-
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
The exception still seems to be being raised in Python 3.6, but I don't fully understand where in the traceback it is being raised from.

Could someone take a look and give me a pointer?

mrocklin:
We need an SSL github team :)

cc @jcrist @mariusvniekerk @quasiben

jrbourbeau:
FYI Travis CI is currently experiencing some issues related to build statuses. The Python 3.6 build is still failing for this PR (https://travis-ci.org/github/dask/distributed/builds/666509937)

mrocklin:
Checking in here.  What should we do with this?

jacobtomlinson:
`1.1.1g` is now available and SHOULD resolve this. It may be sensible to exclude `1.1.1f.

PR #3628, opened 2020-03-23
Check `nbytes` and `types` before reading `data`

jakirkham:
To avoid reading `data` when it is not needed, try checking `nbytes` and `types` beforehand. If the metadata is already there, continue on without reading `data`. Otherwise fallback to the reading `data`, but do make sure to cache the results of that read to avoid doing it again.

cc @madsbk @mrocklin

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1836,13 +1836,16 @@ class Worker(ServerNode):
 
     def send_task_state_to_scheduler(self, key):
         if key in self.data or self.actors.get(key):
-            try:
-                value = self.data[key]
-            except KeyError:
-                value = self.actors[key]
-            nbytes = self.nbytes[key] or sizeof(value)
-            typ = self.types.get(key) or type(value)
-            del value
+            nbytes = self.nbytes[key]
+            typ = self.types.get(key)
+            if nbytes is None or typ is None:
+                try:
+                    value = self.data[key]
+                except KeyError:
+                    value = self.actors[key]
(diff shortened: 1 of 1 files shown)


distributed/worker.py
```diff
@@ -1836,13 +1836,16 @@ def ensure_communicating(self):
 
     def send_task_state_to_scheduler(self, key):
         if key in self.data or self.actors.get(key):
-            try:
-                value = self.data[key]
-            except KeyError:
-                value = self.actors[key]
-            nbytes = self.nbytes[key] or sizeof(value)
```
jakirkham: This appeared to just use `getitem` on `nbytes`, which AFAICT is a `dict`. So maybe getting `value` is never needed? If it is, maybe we should be changing this to `.get(...)`.

mrocklin: Yeah, good catch.  I don't know.

mrocklin:
+1 from me.  Thanks @jakirkham

distributed/worker.py
jakirkham: I went ahead and used `.get(...)` instead and raised issue ( https://github.com/dask/distributed/issues/3629 ) to determine whether the fallback to get `value` is needed.

jakirkham:
Thanks for the guidance Matt 😄

mrocklin:
Not sure I did anything, but I'll take all the credit I can get :)

quasiben:
Thanks everyone for the reviews and @jakirkham for the work.  Merging in

PR #3634, opened 2020-03-25
Drop unused line from `pack_frames_prelude`

jakirkham:
`lengths` is replaced in the following line and the previous assignment does not appear to be used. So just drop it.


quasiben:
Thanks for the cleanup!

PR #3635, opened 2020-03-25
Simplify `unpack_frames`

jakirkham:
Handles all of the `struct` work of unpacking `nframes` and `lengths` up front. Then simply iterates through the frames pulling out each one using its respective length.

### DIFF
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -1,6 +1,9 @@
 import struct
 import msgpack
 
+from itertools import accumulate
+from tlz import cons, sliding_window
+
 from ..utils import ensure_bytes, nbytes
 
 BIG_BYTES_SHARD_SIZE = 2 ** 26
@@ -132,12 +135,7 @@ def unpack_frames(b):
     (n_frames,) = struct.unpack_from(fmt, b)
     lengths = struct.unpack_from(n_frames * fmt, b, fmt_itemsize)
 
-    frames = []
     start = (1 + n_frames) * fmt_itemsize
-    for i in range(n_frames):
-        (length,) = struct.unpack(fmt, b[(i + 1) * fmt_itemsize : (i + 2) * fmt_itemsize])
-        frame = b[start : start + length]
-        frames.append(frame)
-        start += length
+    frames = [b[i:j] for i, j in sliding_window(2, accumulate(cons(start, lengths)))]
 
     return frames
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -126,12 +126,15 @@ def unpack_frames(b):
     --------
     pack_frames
     """
-    (n_frames,) = struct.unpack("Q", b[:8])
+    fmt = "Q"
+    fmt_itemsize = struct.calcsize(fmt)
+
+    (n_frames,) = struct.unpack_from(fmt, b)
 
     frames = []
-    start = 8 + n_frames * 8
+    start = (1 + n_frames) * fmt_itemsize
     for i in range(n_frames):
-        (length,) = struct.unpack("Q", b[(i + 1) * 8 : (i + 2) * 8])
+        (length,) = struct.unpack(fmt, b[(i + 1) * fmt_itemsize : (i + 2) * fmt_itemsize])
         frame = b[start : start + length]
         frames.append(frame)
         start += length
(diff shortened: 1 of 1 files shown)


distributed/protocol/utils.py
```diff
@@ -126,14 +129,13 @@ def unpack_frames(b):
...
-
-    frames = []
-    start = 8 + n_frames * 8
-    for i in range(n_frames):
-        (length,) = struct.unpack("Q", b[(i + 1) * 8 : (i + 2) * 8])
-        frame = b[start : start + length]
-        frames.append(frame)
-        start += length
+    fmt = "Q"
+    fmt_itemsize = struct.calcsize(fmt)
```
quasiben: I was not familiar with `calcsize`, this and the logic below is the equivalent of [:8], correct ?

distributed/protocol/utils.py
```diff
@@ -126,14 +129,13 @@ def unpack_frames(b):
...
+
+    (n_frames,) = struct.unpack_from(fmt, b)
+    lengths = struct.unpack_from(n_frames * fmt, b, fmt_itemsize)
+
+    start = (1 + n_frames) * fmt_itemsize
+    frames = [b[i:j] for i, j in sliding_window(2, accumulate(cons(start, lengths)))]
```
quasiben: Is this faster ?  Arguably, the for loop is more explicit though more verbose.

mrocklin: Yeah, the version in master is, I think, more readily understood by a broader set of people.

If there isn't a strong motivation for this PR, like correctness or performance, I suggest that we pass on this change.

distributed/protocol/utils.py
jakirkham: Yeah that's right.

distributed/protocol/utils.py
jakirkham: I've been reading through the code in this file pretty closely as we keep seeming to run into issues either in it or at least related to it.

TBH reading this function took me a few passes. It was unclear if the lengths and frames were interwoven somehow. Eventually I realized the lengths were coming in a header. After thinking about it what was throwing me was the messing with offsets here. It's something one might see in C/C++ code (probably with comments), but in Python without comments just felt hard to parse.

This rewrite just tries to make it clearer what's going on so other (like me) can parse what is happening here. If we think this is easier to maintain as it was, I'd be very interested in hearing why that is.

jakirkham: Should add slicing performance is secondary (at least to me), but is also worth noting. When we use objects like `bytes`, Python is smart enough to avoid a copy (`bytes` is immutable after all). However when we use objects like `bytearray`, Python will copy the data. So we can be paying a little or a lot in performance depending what comes through here (`tcp` collects `bytearray` objects for example). Here's a simple benchmark for context

<details>

```python
In [1]: b = bytes(1_000)                                                        

In [2]: %timeit b[:]                                                            
53.2 ns ± 0.345 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: b = bytes(1_000_000)                                                    

In [4]: %timeit b[:]                                                            
52.9 ns ± 0.163 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: b = bytearray(1_000)                                                    

In [6]: %timeit b[:]                                                            
192 ns ± 2.92 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: b = bytearray(1_000_000)                                                

In [8]: %timeit b[:]                                                            
30.5 µs ± 698 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

</details>

mrocklin: I don't doubt that this version is easier for you to understand, but that might not be true for others.  In particular, people now need to grok sliding_window, accumulate, and cons, which I think rules out the vast majority of people.  I'm in favor of the for-loop approach personally.

mrocklin:
Currently I'm in favor of closing this.  I don't think that it significantly improves readability (although this is subjective), and the maintainer time to benefit ratio here seems to be high.

distributed/protocol/utils.py
jakirkham: In the past reviewers have pushed for list comprehensions, so that's why I did it here. If you'd prefer rewriting the list comprehension as a `for`-loop, I don't mind tweaking it.

PR #3638, opened 2020-03-25
UCX synchronize default stream only on CUDA frames

pentschev:
Avoid unnecessary CUDA default stream synchronizations if frames don't have CUDA buffers.

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -175,7 +175,8 @@ class UCX(Comm):
                 #  syncing the default stream will wait for other non-blocking CUDA streams.
                 # Note this is only sufficient if the memory being sent is not currently in use on
                 # non-blocking CUDA streams.
-                synchronize_stream(0)
+                if any([hasattr(f, "__cuda_array_interface__") for f in frames]):
+                    synchronize_stream(0)
 
                 for frame in frames:
                     if nbytes(frame) > 0:
@@ -222,7 +223,8 @@ class UCX(Comm):
 
                 # It is necessary to first populate `frames` with CUDA arrays and synchronize
                 # the default stream before starting receiving to ensure buffers have been allocated
-                synchronize_stream(0)
+                if any(is_cudas.tolist()):
+                    synchronize_stream(0)
                 for i, (is_cuda, size) in enumerate(
                     zip(is_cudas.tolist(), sizes.tolist())
                 ):
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -158,9 +158,9 @@ class UCX(Comm):
                 )
 
                 # Send meta data
-                cuda_frames = [hasattr(f, "__cuda_array_interface__") for f in frames]
+                cuda_frames = np.array([hasattr(f, "__cuda_array_interface__") for f in frames], dtype=np.bool)
                 await self.ep.send(np.array([len(frames)], dtype=np.uint64))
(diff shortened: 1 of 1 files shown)


pentschev:
cc @quasiben @jakirkham

quasiben:
Looks good.  I'll merge when this passes

distributed/comm/ucx.py
```diff
@@ -175,7 +175,8 @@ def peer_address(self) -> str:
                 #  syncing the default stream will wait for other non-blocking CUDA streams.
                 # Note this is only sufficient if the memory being sent is not currently in use on
                 # non-blocking CUDA streams.
-                synchronize_stream(0)
+                if any([hasattr(f, "__cuda_array_interface__") for f in frames]):
```
jakirkham: We collect this earlier on line 164 as well. Thoughts on just storing that to a variable so we can reuse it here?

distributed/comm/ucx.py
```diff
@@ -222,7 +223,8 @@ def peer_address(self) -> str:
 
                 # It is necessary to first populate `frames` with CUDA arrays and synchronize
                 # the default stream before starting receiving to ensure buffers have been allocated
-                synchronize_stream(0)
+                if any(is_cudas.tolist()):
```
jakirkham: Can just use this reduction with the NumPy array.

```suggestion
                if is_cudas.any():
```

distributed/comm/ucx.py
jakirkham: We could then do an analogous thing here as below.

```python
if is_cudas.any():
    ...
```

distributed/comm/ucx.py
```diff
@@ -158,13 +158,9 @@ def peer_address(self) -> str:
                 )
 
                 # Send meta data
+                cuda_frames = [hasattr(f, "__cuda_array_interface__") for f in frames]
```
jakirkham: Would actually just go ahead and make this an array since we need that later anyways.

```suggestion
                cuda_frames = np.array([hasattr(f, "__cuda_array_interface__") for f in frames], dtype=bool)
```

distributed/comm/ucx.py
```diff
@@ -158,13 +158,9 @@ def peer_address(self) -> str:
...
                 await self.ep.send(np.array([len(frames)], dtype=np.uint64))
-                await self.ep.send(
-                    np.array(
-                        [hasattr(f, "__cuda_array_interface__") for f in frames],
-                        dtype=np.bool,
-                    )
-                )
+                await self.ep.send(np.array(cuda_frames, dtype=np.bool))
```
jakirkham: We can then just send it as-is.

```suggestion
                await self.ep.send(cuda_frames)
```

distributed/comm/ucx.py
```diff
@@ -175,7 +171,8 @@ def peer_address(self) -> str:
                 #  syncing the default stream will wait for other non-blocking CUDA streams.
                 # Note this is only sufficient if the memory being sent is not currently in use on
                 # non-blocking CUDA streams.
-                synchronize_stream(0)
+                if any(cuda_frames):
```
jakirkham: Then call the `.any()` method.

```suggestion
                if cuda_frames.any():
```

pentschev:
Thanks @jakirkham , I believe this should address your comments.

pentschev:
Thanks for the review @jakirkham .

PR #3639, opened 2020-03-26
Merge frames in `deserialize_bytes`

jakirkham:
Fixes https://github.com/dask/distributed/issues/3851

It appears that we are splitting frames in `serialize_byteslist` so that we can compress them. However we are not merging them back together afterwards during deserialization. This can cause an exception to be raised by serializers that expected their frames to be structured in a particular way.

To fix this, we make sure to call `merge_frames` after `decompress` in `deserialize_bytes`. Further we make sure to pack the `lengths` of the original `frames` in the `header` (if not already present) in `serialize_byteslist`. This should ensure deserializers get the original frame structuring back when they operate on them.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -9,12 +9,13 @@ from tlz import valmap, get_in
 import msgpack
 
 from . import pickle
-from ..utils import has_keyword, typename
+from ..utils import has_keyword, nbytes, typename
 from .compression import maybe_compress, decompress
 from .utils import (
     unpack_frames,
     pack_frames_prelude,
     frame_split_size,
+    merge_frames,
     ensure_bytes,
     msgpack_opts,
 )
@@ -473,6 +474,8 @@ def nested_deserialize(x):
 
 def serialize_bytelist(x, **kwargs):
     header, frames = serialize(x, **kwargs)
+    if "lengths" not in header:
+        header["lengths"] = tuple(map(nbytes, frames))
     frames = sum(map(frame_split_size, frames), [])
     if frames:
         compression, frames = zip(*map(maybe_compress, frames))
@@ -499,6 +502,7 @@ def deserialize_bytes(b):
     else:
         header = {}
     frames = decompress(header, frames)
+    frames = merge_frames(header, frames)
     return deserialize(header, frames)
 
 
--- a/distributed/protocol/tests/test_serialize.py
+++ b/distributed/protocol/tests/test_serialize.py
@@ -190,9 +190,12 @@ def test_empty_loads_deep():
     assert isinstance(e2[0][0][0], Empty)
 
 
-def test_serialize_bytes():
-    for x in [1, "abc", np.arange(5), b"ab" * int(40e6)]:
-        b = serialize_bytes(x)
+@pytest.mark.parametrize(
+    "kwargs", [{}, {"serializers": ["pickle"]},],
+)
+def test_serialize_bytes(kwargs):
+    for x in [1, "abc", np.arange(5), b"ab" * int(40e6), 2 ** 26 * b"ab"]:
+        b = serialize_bytes(x, **kwargs)
         assert isinstance(b, bytes)
         y = deserialize_bytes(b)
         assert str(x) == str(y)
--- a/distributed/protocol/tests/test_serialize.py
+++ b/distributed/protocol/tests/test_serialize.py
@@ -194,7 +194,14 @@ def test_empty_loads_deep():
     "kwargs", [{}, {"serializers": ["pickle"]},],
 )
 def test_serialize_bytes(kwargs):
-    for x in [1, "abc", np.arange(5), b"ab" * int(40e6), 2 ** 26 * b"ab"]:
+    for x in [
+        1,
+        "abc",
+        np.arange(5),
+        b"ab" * int(40e6),
+        2 ** 26 * b"ab",
+        (2 ** 25 * b"ab", 2 ** 25 * b"ab"),
+    ]:
         b = serialize_bytes(x, **kwargs)
         assert isinstance(b, bytes)
         y = deserialize_bytes(b)
--- a/distributed/protocol/tests/test_serialize.py
+++ b/distributed/protocol/tests/test_serialize.py
@@ -199,8 +199,8 @@ def test_serialize_bytes(kwargs):
         "abc",
         np.arange(5),
         b"ab" * int(40e6),
-        2 ** 26 * b"ab",
-        (2 ** 25 * b"ab", 2 ** 25 * b"ab"),
+        int(2 ** 26) * b"ab",
+        (int(2 ** 25) * b"ab", int(2 ** 25) * b"ab"),
     ]:
         b = serialize_bytes(x, **kwargs)
         assert isinstance(b, bytes)
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -59,9 +59,6 @@ def merge_frames(header, frames):
     """
     lengths = list(header["lengths"])
 
-    if not frames:
-        return frames
-
     assert sum(lengths) == sum(map(nbytes, frames))
 
     if all(len(f) == l for f, l in zip(frames, lengths)):
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -67,9 +67,6 @@ def merge_frames(header, frames):
     if all(len(f) == l for f, l in zip(frames, lengths)):
         return frames
 
-    if len(lengths) == 1 and len(frames) == 1:
-        return frames
-
     frames = frames[::-1]
     lengths = lengths[::-1]
 
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -67,8 +67,8 @@ def merge_frames(header, frames):
     if all(len(f) == l for f, l in zip(frames, lengths)):
         return frames
 
-    if len(lengths) == 1:
-        return [b"".join(map(ensure_bytes, frames))]
+    if len(lengths) == 1 and len(frames) == 1:
+        return frames
 
     frames = frames[::-1]
     lengths = lengths[::-1]
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -67,6 +67,9 @@ def merge_frames(header, frames):
     if all(len(f) == l for f, l in zip(frames, lengths)):
         return frames
 
+    if len(lengths) == 1:
+        return [b"".join(map(ensure_bytes, frames))]
+
     frames = frames[::-1]
     lengths = lengths[::-1]
 
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -931,7 +931,9 @@ def ensure_bytes(s):
     >>> ensure_bytes(b'123')
     b'123'
     """
-    if hasattr(s, "encode"):
+    if isinstance(s, bytes):
+        return s
+    elif hasattr(s, "encode"):
         return s.encode()
     else:
         try:


distributed/protocol/serialize.py
```diff
@@ -440,6 +441,8 @@ def replace_inner(x):
 
 def serialize_bytelist(x, **kwargs):
     header, frames = serialize(x, **kwargs)
+    if "lengths" not in header:
```
quasiben: We (you :) ) recently added lengths to CUDA object headers.  When is lengths not in header ?  Is this something we should be requiring ?

jakirkham: Yeah still wrapping my head around this. My understanding is the header from the object has already gone through msgpack at this point so is actually a frame as well. So it may be we always need to set the `length` here.

jakirkham:
@mrocklin, do you have any thoughts on this? 🙂

mrocklin:
I have no particular thoughts on this.

mrocklin:
Does this need a test?

mrocklin:
This appears to be stalled.  @jakirkham @quasiben are you all still active here? Should we close this?

jakirkham:
Yeah still working on this.

martindurant:
Another friendly ping and status request, just to keep this thread going.

jakirkham:
I’ve been out on PTO for the last week

jakirkham:
@michaelnarodovitch @gshimansky, could you please give this a try?

jakirkham:
Also sorry for the long delay Martin and Matt. This now has a test. First commit demonstrates this fails without the change. Second commit demonstrates the change fixes it.

martindurant:
In as much as this fixes something that was shown broken, I am happy; but I can't pretend to understand the reason.
Does the extra call to `merge_frames` have any cost for the typical single-frame case? I expect not.

michaelnarodovitch:
Pulled `distributed-2.17.0+6.g1313e147` from current master and reran my (previously failing) usecase with that version. Works flawless now.

The fix reveals the failure mode in my usecase. The dataframe, read from disk after it was spilled, ended up being split into multiple frames, which, in turn, caused `pickle` being called with the `buffers` keyword from `python 3.8` via
https://github.com/dask/distributed/blob/master/distributed/protocol/serialize.py#L62

gshimansky:
Yes! I checked your branch `merge_bytes_frames` and on that branch I don't have this problem any more. Please merge this change and publish a new release.

martindurant:
I'd just like to repeat, @jakirkham : `merge_frames` does not special-case one frame exactly, so it may produce a performance regression, no?

michaelnarodovitch:
#3851 for reference

jakirkham:
Yeah I'm just looking at this now, Martin. Will see if we have a sensible way to fast path it.

Had asked folks to try this out to confirm it actually works for them (before spending more time). Sounds like it is working and therefore worth spending more time on.

martindurant:
Perfect. I imagine an `if len(frames) == 1` branch in `merge_frames` would be reasonable

jakirkham:
Have pushed a commit to short circuit where we want a single frame from merging. Also added another test case where there are two frames that both get split. Finally realized that `ensure_bytes` was copying data that it shouldn't. So pushed a fix for that as well. Please let me know what you think 🙂

distributed/utils.py
```diff
@@ -931,7 +931,9 @@ def ensure_bytes(s):
     >>> ensure_bytes(b'123')
     b'123'
     """
-    if hasattr(s, "encode"):
+    if isinstance(s, bytes):
```
martindurant, line 934: I don't believe `bytes(s)` does make a copy if `s` is already `bytes`

distributed/protocol/utils.py
```diff
@@ -67,6 +67,9 @@ def merge_frames(header, frames):
     if all(len(f) == l for f, l in zip(frames, lengths)):
         return frames
 
+    if len(lengths) == 1:
+        return [b"".join(map(ensure_bytes, frames))]
```
martindurant, line 71: Does it not imply only one frame, so you can `return ensure_bytes(frames[0])` ?

distributed/protocol/serialize.py
```diff
@@ -473,6 +474,8 @@ def replace_inner(x):
...
+        header["lengths"] = tuple(map(nbytes, frames))
     frames = sum(map(frame_split_size, frames), [])
```
jakirkham, line 479: Note that the frames are already be split here. This is due to constraints caused by compression (which happens below).

distributed/protocol/utils.py
jakirkham: No it doesn't. It implies there was only one frame. However we may already have split it up during serialization. Noted the line where this happens below.

martindurant: Is it worth, then, checking for exactly one frame, to avoid the copy during `join`?

distributed/utils.py
jakirkham: Hmm...maybe I'm misremembering. In any event we seem to have [similar code in Dask]( https://github.com/dask/dask/blob/79836fe4a52c9cf0e1eb32525792ad6bfed1c62b/dask/utils.py#L772-L773 ).

distributed/protocol/utils.py
jakirkham: Not sure. Was trying to integrate your feedback above. Happy to revert it if you prefer 🙂

jakirkham, line 71: Or do you mean you would like to see something like this?

```suggestion
    if len(lengths) == 1 and len(frames) == 1:
        return frames
```

jakirkham: Tried to respond inline, but GitHub didn't like it. Moved here: https://github.com/dask/distributed/pull/3639#pullrequestreview-424799455

distributed/utils.py
martindurant: OK, well never mind - this version shouldn't hurt. I wonder why not just call the dask version? I suppose this one can work on memoryviews and bytearrays too.

distributed/protocol/utils.py
martindurant: Yes, `return ensure_bytes(frames[0])` (although I don't see how it could not be bytes).

distributed/utils.py
jakirkham: Yeah I was thinking about that as well, but didn't want to go down a rabbit hole here. Am ok pulling this out into a separate PR so we can explore orthogonally. WDYT?

martindurant: I'm saying your version is fine :)

distributed/protocol/utils.py
jakirkham: I think we have `bytearray` in the TCP code path and NumPy arrays in the UCX code path. So coercing to `bytes` can make sense. Use a `list` to wrap it as well since that seems to be expected (based on related testing before).

```suggestion
    if len(lengths) == 1 and len(frames) == 1:
        return [ensure_bytes(frames[0])]
```

martindurant: yep, agree

jakirkham: That said, I guess `merge_frames` is collecting some frames unaltered, some as `memoryview`s, and some as `bytes`. So maybe we can just pass `frames` through as-is? Would potentially save us a copy.

martindurant: I don't know about that, depends if all callers explicitly expect bytes or not.

jakirkham: Just for context, here are the lines I'm looking at:

1. [Preserve frame]( https://github.com/dask/distributed/blob/85620bf43d67807e60cdcb45d18270f131a1cb14/distributed/protocol/utils.py#L80 )
2. [Convert to `memoryview`]( https://github.com/dask/distributed/blob/85620bf43d67807e60cdcb45d18270f131a1cb14/distributed/protocol/utils.py#L85 )
3. [Convert to `bytes` (if there are multiple frames)]( https://github.com/dask/distributed/blob/85620bf43d67807e60cdcb45d18270f131a1cb14/distributed/protocol/utils.py#L90 )

My guess is the only expectation is that frames must be `bytes`-like, which they are when inputted or when coerced to `bytes` or `memoryviews`. So I think we are safe to pass the frames as-is.

martindurant: Maybe a good use case for typing - even though I'm not sure how you say "bytes-like".

martindurant: (that was just musing, not suggesting a change)

jakirkham: Good question. I guess it's an open issue ( https://github.com/python/typing/issues/593 ).

jakirkham:
Alright I think the latest changes address the last round of reviews. Please let me know if there's anything else 🙂

distributed/protocol/utils.py
```diff
@@ -67,6 +67,9 @@ def merge_frames(header, frames):
     if all(len(f) == l for f, l in zip(frames, lengths)):
         return frames
 
+    if len(lengths) == 1 and len(frames) == 1:
```
martindurant, line 70: I think I may have been stupid here: is there any way that this condition can be triggered, given the assert and the previous if? If there is only one frame and one length, that length must be the same as the length of the one frame.

If this block came before the previous, then we would avoid calculating lengths again - which should be instantaneous in any case, so should we just get rid of it?

(indeed, the `if all...` condition also catches `if not frames` above, and would be fast to iterate over a zero-length list; and the assert is only needed after that if...)

jakirkham: Actually I think you were right to request this and it can come up. For frames that are not too large and for situations where only a single frame is needed, this could be quite common. For example NumPy arrays would pass through this path. So having a fast path makes sense. This saves us going through the `while`, which does a fair bit of work.

Well we would hope the frame is the same size as the length specified. However that might not be true if the message was truncated or otherwise corrupted (and I think this is why the `assert`s are there). So having the `assert`s makes sense. If we moved this before the `assert`s, I think we should add a check to ensure the frame is the expected length.

I might not be following this last bit, but I think it could make sense to move the `if not frames` after the `assert`s or drop it all together. No strong thoughts here. Happy to leave it.

jakirkham: Ah I see what you mean (I think). [This check]( https://github.com/dask/distributed/blob/a71951bc60f13054d50af4ff479fd1fbc64b13ad/distributed/protocol/utils.py#L67-L68 ) already acts as a fast path. Yeah we can drop our case.

jakirkham: Ok pushed two commits that drop these.

martindurant:
+1 here, sorry for the back-and-forth.

distributed/protocol/tests/test_serialize.py
```diff
@@ -190,9 +190,19 @@ def test_empty_loads_deep():
...
+    "kwargs", [{}, {"serializers": ["pickle"]},],
+)
+def test_serialize_bytes(kwargs):
+    for x in [
+        1,
+        "abc",
+        np.arange(5),
+        b"ab" * int(40e6),
+        2 ** 26 * b"ab",
+        (2 ** 25 * b"ab", 2 ** 25 * b"ab"),
```
mrocklin, line 203: Beware here that I've seen Python pre-compute literals like these as an unfortunate optimization, which results in large constant values in the program code.  It might be wise to call a function around `2 **26` like `int(2 ** 26)` even though it has no semantic effect.

My bad experience on this was a long time ago, and so maybe things have improved..

mrocklin:
FWIW these changes seem fine to me

distributed/protocol/tests/test_serialize.py
jakirkham: Hmm...yeah I'm not sure. Did push some `int` wrappers around these though. One of these cases was already here (so we may have already been ok).

jakirkham:
Not at all. Thanks for the feedback! 😄

distributed/protocol/tests/test_serialize.py
mrocklin: Ah ok, probably not a problem then.  Thanks again for resolving the full issue quickly.

jakirkham: Of course 🙂

jakirkham:
At least for me the Travis CI status is not being shown on GitHub (not sure why), but it did [run and pass]( https://travis-ci.org/github/dask/distributed/builds/695137295 ).

mrocklin:
It is visible and has a green check mark to me.  Merging

jakirkham:
Great! Yeah it now shows up for me too. Not sure what was up.

Thanks all! 😄

distributed/utils.py
jakirkham: FWIW updating the Dask implementation to contain the code from Distributed in PR ( https://github.com/dask/dask/pull/9050 ). Then we should be able to switch to the Dask implementation in Distributed.

Issue #3627, opened 2020-03-23
`Adaptive` doesn't scale quickly for long-running jobs


kazimuth:
I'm using dask to coordinate some long-running machine learning jobs. I've set up an adaptive cluster (with `dask_jobqueue`) that has a minimum of 5 workers and a maximum of 10. Each task I dispatch takes about two hours to run and consistently uses ~100% of the CPU available to it. However, the adaptive cluster doesn't seem to want to add any more workers. It sits at the minimum number and never increases. Is there some way to modify the scheduling policy so that the cluster scales up more aggressively?

I'm aware this isn't exactly the sort of job dask is designed to schedule -- it wants smaller, faster tasks was my impression. I think you might be able to modify `Adaptive` to use a different policy that's better suited for long-running jobs? But I spent some time digging into the source and got kinda lost. Any pointers would be helpful :)

My current workaround is ignoring Adaptive and scaling the cluster by hand. I feel bad for taking up nodes longer than I need though.

fjetter:
The decision to scale up or down is currently coupled to the measurement of a tasks runtime. (Similar tasks are grouped, see [TaskPrefix](https://github.com/dask/distributed/blob/4d4d935f46619ad9d96d64190274201ebe894eec/distributed/scheduler.py#L772))
As long as the runtime wasn't measured at least once, the cluster cannot estimate how long the entire computation graph might take and it will not scale up (see [here](https://github.com/dask/distributed/blob/4d4d935f46619ad9d96d64190274201ebe894eec/distributed/scheduler.py#L5209), it boils down to `Scheduler.total_occupancy` if you want to go down the rabbit hole). You _should_ see the cluster scale up once the first job finishes.

We're facing the same issue, see #3516

A workaround for this is to configure default task durations which are used as long as there are no measurements available, e.g.

```
distributed:
  scheduler:
    default-task-durations:
      my-function: 2h  # This should be the same name you see on the dashboard
```

fjetter:
Will close this issue as a duplicate to #3516 

I can also recommend to have a look at #3642 which introduces an option to define a default runtime for unknown tasks which might offer a practical way to resolve your issue as well.

chrisroat:
> A workaround for this is to configure default task durations which are used as long as there are no measurements available, e.g.
> 
> ```
> distributed:
>   scheduler:
>     default-task-durations:
>       my-function: 2h  # This should be the same name you see on the dashboard
> ```

[...]

Issue #3516, opened 2020-02-24
Dynamic adaption of average task duration


sublinus:
We currently experience an issue with very long-running tasks, when using `adaptive_target` to figure out how many workers the dask graph needs.
The estimation by `adaptive_target` is currently wrong as long as no task of the TaskPrefix has been executed yet. 
Could we change the estimation to correct dynamically and propagate a new average across the cluster as soon as the execution of one task e.g. already takes twice the estimated time?

sublinus:
Do you see any chance in adding timing logic around the ThreadPoolExecutor on the worker and creating a new endpoint to the scheduler, which receives updated timing info for tasks and then propagates that to all other workers?

```python
start_time = time.time()
future = executor.submit(function, *args, **kwargs)
pc = PeriodicCallback(lambda: self.threadpool_timing_callback(start_time), 1000)
pc.start()
```
```python
def threadpool_timing_callback(self, start_time):
     if time.time() - start_time > some_boundary:
          update_average()
          notify_scheduler(task_key, new_average)
```
https://github.com/dask/distributed/blob/master/distributed/worker.py#L2309-L2319

mrocklin:
> Could we change the estimation to correct dynamically and propagate a new average across the cluster as soon as the execution of one task e.g. already takes twice the estimated time?

Sure.  I think that we currently do something similar with work stealing.  Maybe?  

It is a little bit hard because we don't know when the first task of a prefix has actually started running, only when it has been sent to a worker.

Or perhaps this is the motivation for your second comment, about including some signal in the worker that talks back to the scheduler once a task has started processing?  Was this was you meant?

mrocklin:
If so, the concern here is that we're increasing the communication by a non-trivial amount.  Sending a new message for every task is doable, but does add non-trivial cost.

sublinus:
> Or perhaps this is the motivation for your second comment, about including some signal in the worker that talks back to the scheduler once a task has started processing? Was this was you meant?

Yes, the timing around the executor would be a way to solve the problem with the information about when the tasks start running.
 
Assuming that task duration in one TaskPrefix is roughly normally distributed I think the amount of additional messages could be limited to: `O(prefixes *  task duration corrections * workers)`  by only sending messages when the currently executing task is over the boundary, which would adapt dynamically with an exponential backoff and then propagate the new expected duration to all other workers via the scheduler.

[...]

PR #3642, opened 2020-03-26
Introduce config for default task duration

sublinus:
This introduces the key `distributed.scheduler.task-duration` in the Dask config to configure a default task duration for all tasks, when there is no otherwise configured value available.

That is useful in cases of very long-running tasks to tune the occupancy of the cluster (see #3516 and #3627 )

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -24,6 +24,7 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []
     preload-argv: []
+    unknown-task-duration: 0.5s  # Default duration for all tasks with unknown durations ("15m", "2h")
     default-task-durations:  # How long we expect function names to run ("1h", "1s") (helps for long tasks)
       rechunk-split: 1us
       shuffle-split: 1us
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -24,7 +24,7 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []
     preload-argv: []
-    unknown-task-duration: 0.5s  # Default duration for all tasks with unknown durations ("15m", "2h")
+    unknown-task-duration: 500ms  # Default duration for all tasks with unknown durations ("15m", "2h")
     default-task-durations:  # How long we expect function names to run ("1h", "1s") (helps for long tasks)
       rechunk-split: 1us
       shuffle-split: 1us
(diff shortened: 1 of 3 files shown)


quasiben:
Thanks @sublinus .  Looks reasonable to me and definitely appreciate the test!  I am going to wait until the end of the day to merge should other maintainers want to comment

distributed/distributed.yaml
```diff
@@ -24,6 +24,9 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []
     preload-argv: []
+    # Assume this duration ("15m", "2h") for all tasks with unknown durations
+    # This can be useful to change occupancy estimations in Adaptive scenarios
+    task-duration: null
```
mrocklin: Should this have a non-null value?  What do we do currently?

So far we've been limiting comments for configuration values to be a bit shorter and be on the same line.  If we choose to change convention here and have much larger comments then we should probably do that across the file.  For now I recommend that we try to make the comment more brief if we can.

sublinus:
Thanks for all your comments. The question regarding the non-null value just pointed me to `scheduler.get_task_duration`, which is where I think this `unknown-task-duration` should be used instead of in the `TaskPrefix` as an average. 
Currently the default used for tasks with unknown task durations is 0.5s, which is now reflected in the config.
I totally see your point regarding the comment style.

sublinus:
The partial CI failure seems to be unrelated here.

quasiben:
@sublinus i think you are probably right but I restarted the tests job to be sure.

distributed/distributed.yaml
```diff
@@ -24,6 +24,7 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []
     preload-argv: []
+    unknown-task-duration: 0.5s  # Default duration for all tasks with unknown durations ("15m", "2h")
```
mrocklin:
```suggestion
    unknown-task-duration: 500ms  # Default duration for all tasks with unknown durations ("15m", "2h")
```

distributed/tests/test_scheduler.py
```diff
@@ -2057,3 +2057,14 @@ def reducer(x, y):
...
     assert s.aliases == {a.name: a.address}
+
+
+@gen_cluster(client=True, config={"distributed.scheduler.unknown-task-duration": "1h"})
+async def test_unknown_task_duration_config(client, s, a, b):
+    future = client.submit(inc, 1)
+    while not s.tasks:
+        await asyncio.sleep(0.001)
+    assert sum(s.get_task_duration(ts) for ts in s.tasks.values()) == 3600
+    assert len(s.unknown_durations) == 1
```
mrocklin: What if `inc` finishes before we get here?

mrocklin: You might want to consider `slowinc`

sublinus: Good point. It should be easy to switch to `slowinc` here.

sublinus:
The build shows as in progress even though GH Actions and Travis ran through fine. It looks like the notify about the successful build from Travis to Github failed.

mrocklin:
Thanks @sublinus !  This is in

PR #3649, opened 2020-03-27
Bump checkout GitHub action to v2

jrbourbeau:
I think this will fix an issue where restarting GitHub actions CI builds would fail during code checkout


jrbourbeau:
Restarting the GHA builds...

jrbourbeau:
For reference, upgrading to v2 fixed the issue where re-started builds would fail at the code checkout step

quasiben:
I saw this failure earlier today thanks @jrbourbeau for the fix

Issue #2181, opened 2018-08-13
Several millions tasks and Client.map


guillaumeeb:
Hi,

I'm facing an issue that has already been raised in other more complicated forms, see pangeo-data/pangeo/issues/266, dask/dask/issues/3514, and probabaly other places. I've read those threads, but I cannot see yet a simple solution or good workaround for my problem.

I'm using Dask to do some really basic embarrassingly parallel problem:

- I've got a scientific code, that I will call using subprocess
- I need to run int with exactly 3123120 different set of parameters.

So here is basically what I do:
````python
cmd_lines = []
for param1 in param1_set
    for param2 in param2_set
         # Some other nested loops
         cmd_lines += [' '.join(("myscript.sh", param1, param2, ... ))]

def run_process(cmd):
    os.chdir(WORK_DIR)
    process = Popen(cmd.split(' '), stdout=PIPE)
    (output, err) = process.communicate()
    exit_code = process.wait()
    return output, err

futures = client.map(run_process, cmd_lines)
results = client.gather(futures)
```` 

When taking only the first 100,000 elements of cmd_lines, this works with some minor problems:

- It takes between 40 seconds and 1m30s before my workers do anything, I imagine this is the time of building the graph and sending to worker.
- Gather takes about 20 seconds, this seems better than the scattering though.

When trying with the full 3 billions parameters, I can't see any computation starting, and the Dashboard is almost not responding anymore.

So an easy workaround is just going 100,000 tasks by 100,000, but I'm not satisfied by this, and it unnecessarily complexifies the really simple code here.

To sum up, I'd like to have a solution for two problems:
- First one is more of an annoyance: the time taken to actually start computation, without any information on the Dashboard. For such an embarrassingly parallel workload, I would even expect the tasks to start right after submission, they could be sent to workers in a streaming way, and we should not wait for 100,000 of tasks to be all sent to workers to begin.
- Second one which is probably very related, scheduler should handle several billion tasks at a time.

mrocklin:
> First one is more of an annoyance: the time taken to actually start computation, without any information on the Dashboard. For such an embarrassingly parallel workload, I would even expect the tasks to start right after submission, they could be sent to workers in a streaming way, and we should not wait for 100,000 of tasks to be all sent to workers to begin.

[...]

PR #3650, opened 2020-03-27
Add chunksize to Client.map

TomAugspurger:
An informal benchmark

```python
In [9]: %time _ = wait(c.map(inc, range(100_000), pure=False,
partition_size=1_000))
CPU times: user 31.8 s, sys: 1.07 s, total: 32.9 s
Wall time: 33.3 s

In [10]: %time _ = wait(c.map(inc, range(100_000), pure=False))

CPU times: user 45.3 s, sys: 2.13 s, total: 47.5 s
Wall time: 48.4 s
```

The difference likely increases in the size of the iterable.

Closes https://github.com/dask/distributed/issues/2181


mrocklin:
The term partition is a little bit overloaded in Dask.  I wonder if some other term, like batch might work better?

mrocklin:
Should we partition/batch by default?

jakirkham:
+1 for `batch_size`

TomAugspurger:
Batch sounds good. I don't think we should partition by default, since we're mirroring the stdlib here.

mrocklin:
>  I don't think we should partition by default, since we're mirroring the stdlib here.

The API is the same though, right?  `batch_size` is a purely internal change I thought.

TomAugspurger:
I added as a public, user-facing parameter. It could be purely internal, in which case I'd be fine with batching by default (perhaps with a config setting if people really want to control that).

The result will be the same, regardless of whether things are submitted in batches.

jrbourbeau:
FWIW the stdlib [`map` function](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map) in `concurrent.futures` has a `chunksize=` keyword. Using that would more closely align with the stdlib, but is also somewhat overloaded already. What do others think about using `chunksize=` here?

TomAugspurger:
Huh thanks @jrbourbeau, I didn't know `concurrent.futures` had that parameter.

I think we should follow the standard library exactly here.

jrbourbeau:
That's my preference as well

mrocklin:
Hey, sorry to switch late, I think that the intention of the concurrent.futures chunksize is different, and possibly orthogonal to what we are talking about here. 

I believe that the cf chunksize argument is intended to wrap many inputs into a single task in order to reduce overhead.  In that respect I think that what we're doing here is different.  We might even support a chunksize= parameter similar to what concurrent.futures does in the future (although I think that there are lower returns for us here).  We might want to chose a different name.

TomAugspurger:
To me, the behavior of `chunksize` sounds the same between Dask and concurrent.futures:

> this method chops iterables into a number of chunks which it submits to the pool as separate tasks. The (approximate) size of these chunks can be specified by setting chunksize to a positive integer. For very long iterables, using a large value for chunksize can significantly improve performance compared to the default size of 1.

mrocklin:
From a user perspective they're the same-ish.  However internally these do different things.  In CF I believe that they're placing many elements of the list into a single task.  In Dask we're pushing out many tasks at a time.

mrocklin:
I'm also happy if we just want to press the green button here.  I'm probably arguing a point that is unlikely to come up for most people.

distributed/client.py
```diff
@@ -1532,6 +1532,7 @@ def map(
         self,
         func,
         *iterables,
+        chunksize=1,
```
mrocklin, line 1535: Going along with the argument of "this isn't quite the same as CF chunksize, wouldn't our default argument here be `math.inf` (or perhaps None)?

martindurant:
@TomAugspurger , this seems to have dropped off the list. The conversation above sounds like it's up to you what you want to name things in the end.

TomAugspurger:
Still thinking through it. I think the meaning of chunksize is the same as
in concurrent.futures, but our default value is the opposite. I'm not sure
how to reconcile those yet.

TomAugspurger:
Changed to `batch_size`, with a default of `None` (no batching).

distributed/client.py
```diff
@@ -1611,6 +1612,35 @@ def map(
...
+                        priority=priority,
+                        allow_other_workers=allow_other_workers,
+                        fifo_timeout=fifo_timeout,
+                        resources=resources,
+                        actor=actor,
+                        actors=actors,
+                        pure=pure,
+                        **kwargs,
+                    )
+                    for batch in partition_all(batch_size, *iterables)
```
mrocklin, line 1635: It looks like `toolz.partition_all` doesn't accept more than one sequence argument, so this will fail for something like ...

```python
client.map(operator.add, range(100), range(100), batch_size=10)
```

(sorry for not finding this earlier)

However, what's here is still an improvement I think.  I'd be happy merging it in, especially given the long review cycle, if desired.

TomAugspurger: Good catch. Hopefully resolved now.

mrocklin:
Thanks @TomAugspurger !  This is in.

PR #3651, opened 2020-03-27
UCX simplify receiving frames in `comm`s

jakirkham:
Basically a redux of PR ( https://github.com/dask/distributed/pull/3599 ) with UCX changes only.

Simplifies the logic for receiving frames in `ucx`. Also preallocates buffers to receive data into beforehand (and subselects the non-0-length buffers). Then hands off control to the underlying library receiving data to fill these buffers. Afterwards handles any sanity checks.

cc @quasiben @pentschev

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -160,9 +160,9 @@ class UCX(Comm):
                     np.array([nbytes(f) for f in frames], dtype=np.uint64)
                 )
                 # Send frames
-                for frame in frames:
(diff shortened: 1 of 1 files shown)


quasiben:
Looks good to me.  @pentschev do you have any comments before we merge ?

pentschev:
LGTM as well. I'll just do a quick test for peace of mind and reported back here in a few minutes, but it should be fine.

jakirkham:
Thanks Ben and Peter! 😀

PR #3652, opened 2020-03-27
do not log an error on unset variable delete

### LINKED ISSUE #3492: Variable delete raises KeyError
It seems that there is an unhandled KeyError in the scheduler variable delete logic.

If using the asynchronous API only visible in the scheduler logs but it does mean that Variables will be "leaking" and staying around instead of being deleted.

It happens regardless of whether you set a value.

Shortest reproduction I can provide is as follows:
```python
from distributed import Variable


def main():
    Client()
    Variable().delete()


if __name__ == '__main__':
    main()
```

```python
tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOLoop object at 0x7f8fe9265be0>>, <Task finished coro=<VariableExtension.delete() done, defined at ./venv/lib/python3.6/site-packages/distributed/variable.py:101> exception=KeyError('variable-68a51209410248fa9f69f84b25f48343',)>)
  File "./venv/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "./venv/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "./venv/lib/python3.6/site-packages/distributed/variable.py", line 110, in delete
    del self.waiting_conditions[name]
KeyError: 'variable-68a51209410248fa9f69f84b25f48343'
```

Python 3.6.10
Reproduced with distributed 2.9.3 and 2.10.0

jjhelmus:
If a Variable is never set or accessed no entries are made in the
tracking attributes of VariableExtension.  Therefore there is no need to
raise or log an error when the variable is deleted.

closes #3492

### DIFF
--- a/distributed/tests/test_variable.py
+++ b/distributed/tests/test_variable.py
@@ -2,6 +2,7 @@ import asyncio
 import random
 from time import sleep
 import sys
+import logging
 
 import pytest
 from tornado import gen
@@ -12,6 +13,7 @@ from distributed.metrics import time
 from distributed.compatibility import WINDOWS
 from distributed.utils_test import gen_cluster, inc, div
 from distributed.utils_test import client, cluster_fixture, loop  # noqa: F401
+from distributed.utils_test import captured_logger
 
 
 @gen_cluster(client=True)
@@ -39,10 +41,15 @@ def test_variable(c, s, a, b):
         assert time() < start + 5
 
 
-def test_delete_unset_variable(client):
+@gen_cluster(client=True)
+def test_delete_unset_variable(c, s, a, b):
     x = Variable()
-    assert x.client is client
-    x.delete()
+    assert x.client is c
+    with captured_logger(logging.getLogger("distributed.utils")) as logger:
+        yield x.delete()
+        yield c.close()
+    text = logger.getvalue()
+    assert "KeyError" not in text
 
 
 @gen_cluster(client=True)
--- a/distributed/tests/test_variable.py
+++ b/distributed/tests/test_variable.py
@@ -42,7 +42,7 @@ def test_variable(c, s, a, b):
 
 
 @gen_cluster(client=True)
-def test_delete_unset_variable(c, s, a, b):
+async def test_delete_unset_variable(c, s, a, b):
     x = Variable()
     assert x.client is c
     with captured_logger(logging.getLogger("distributed.utils")) as logger:
--- a/distributed/tests/test_variable.py
+++ b/distributed/tests/test_variable.py
@@ -46,8 +46,8 @@ async def test_delete_unset_variable(c, s, a, b):
     x = Variable()
     assert x.client is c
     with captured_logger(logging.getLogger("distributed.utils")) as logger:
-        yield x.delete()
-        yield c.close()
+        await x.delete()
+        await c.close()
     text = logger.getvalue()
     assert "KeyError" not in text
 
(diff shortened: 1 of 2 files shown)


jjhelmus:
I've added a test here but it passes even without this change.  An error is logged by the scheduler but I'm not sure how to turn this into a failure in the test.

mrocklin:
> An error is logged by the scheduler but I'm not sure how to turn this into a failure in the test.

Try `grep`ping for `captured_logger`

distributed/tests/test_variable.py
```diff
@@ -39,6 +39,12 @@ def test_variable(c, s, a, b):
         assert time() < start + 5
 
 
+def test_delete_unset_variable(client):
+    x = Variable()
+    assert x.client is client
+    x.delete()
```
mrocklin: I encourage you to use `gen_cluster` whenever possible when testing.  This is for a few reasons:

1.  If we need to debug we can enter a breakpoint anywhere.  `gen_cluster` sets up everything to run in the main thread
2. It's much faster than the `client` fixture, which while more comfortable because it uses the synchronous syntax, also sets up some processes
3.  It will force you to learn some async things if you don't know them already

https://distributed.dask.org/en/latest/develop.html#writing-tests

distributed/variable.py
```diff
@@ -113,8 +113,14 @@ def __init__(self, scheduler):
...
-            del self.waiting_conditions[name]
-            del self.variables[name]
+            try:
+                del self.waiting_conditions[name]
+            except KeyError:
+                pass
+            try:
+                del self.variables[name]
+            except KeyError:
+                pass
```
mrocklin: We sometimes also use the `with ignoring(KeyError)` convention (I think that this now has some analog in the stdlib).  No pressure though.  Using straight try/except is probably understandable by a broader set of readers.

mrocklin:
Friendly ping @jjhelmus

distributed/tests/test_variable.py
jjhelmus: I modified the test to use `gen_cluster` and `captured_logger`, thanks for the suggestion.  

I needed to trigger an action on the client after the delete to ensure the error was logged.  I used `c.close` for this, was curious is there is a better way to do this.

distributed/tests/test_variable.py
```diff
@@ -39,6 +41,17 @@ def test_variable(c, s, a, b):
         assert time() < start + 5
 
 
+@gen_cluster(client=True)
+def test_delete_unset_variable(c, s, a, b):
```
mrocklin, line 45:
```suggestion
async def test_delete_unset_variable(c, s, a, b):
```

distributed/tests/test_variable.py
```diff
@@ -39,6 +41,17 @@ def test_variable(c, s, a, b):
...
+    x = Variable()
+    assert x.client is c
+    with captured_logger(logging.getLogger("distributed.utils")) as logger:
+        yield x.delete()
+        yield c.close()
```
mrocklin, line 50:
```suggestion
        await x.delete()
        await c.close()
```

distributed/tests/test_variable.py
mrocklin: The `yield` approach was common when we were also supporting Python 2.  It's still totally valid, but we've tended to default more towards using async/await.

mrocklin:
Ah, it looks like the delete method isn't asynchronous.  It just fires of a message and returns immediately.  

Hrm, that kind of mismatch makes using these things a bit rough.  Probably not something to resolve now, but something to keep in mind.

Issue #2560, opened 2019-03-12
Bokeh warning from get_task_stream.


TomAugspurger:
```
/home/nfs/an.taugspurger/miniconda3/envs/ucx-dev-test/lib/python3.7/site-packages/bokeh/io/saving.py:126: UserWarning: save() called but no resources were supplied and output_file(...) was never called, defaulting to resources.CDN
  warn("save() called but no resources were supplied and output_file(...) was never called, defaulting to resources.CDN")
         Took 14.14s
```

To avoid the warning, we need to pass a `resources=` to https://github.com/dask/distributed/blob/fb30c33562862f30864456766424b44a3e91aa5b/distributed/client.py#L3590

The bokeh default of `resources=resources.CDN` seems good, but I don't think we should warn.

I could imagine making this configurable, for when you want to view the bokeh plot behind a firewall that doesn't have access to bokeh's CDN. To make it configureable, we'll need to add a `resoruces` keyword to `Client.get_task_stream`  and `distributed.get_task_stream`. Is there a potential ambiguity with the name resources? I could imagine "resources"? http://distributed.dask.org/en/latest/resources.html?highlight=resources#worker-resources Maybe call it `bokeh_resources` to be safe.

mrocklin:
That seems reasonable to me

jakirkham:
Saw this recently. Tom's suggestion also seems reasonable to me.

jakirkham:
It looks like [Bokeh's default is `INLINE`]( https://github.com/bokeh/bokeh/blob/d53cfcc92bc5cf8b13dd59ed6d31ae2bb2c6be65/bokeh/io/saving.py#L65-L67 ). Am I missing something?

prasunanand:
I am working on it.

jrbourbeau:
Great, thanks @prasunanand. I think we want to have a call to `output_file` before `save`, like is done here:

https://github.com/dask/distributed/blob/b0c000883eb10d4801b967fe348c4d6281ca3f1d/distributed/scheduler.py#L5097-L5098

PR #3653, opened 2020-03-28
Add Resouces option to get_task_stream and call output_file

prasunanand:
* Resouces option to get_task_stream 
* call output_file
* Modify docs

Fixes https://github.com/dask/distributed/issues/2560

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -23,8 +23,6 @@ from queue import Queue as pyQueue
 import warnings
 import weakref
 
-from bokeh.resources import INLINE
-
 import dask
 from dask.base import tokenize, normalize_token, collections_to_dsk
 from dask.core import flatten, get_dependencies
@@ -3369,8 +3367,9 @@ class Client(Node):
                 filename = "dask-profile.html"
 
             if filename:
-                from bokeh.plotting import save
+                from bokeh.plotting import output_file, save
 
+                output_file(filename=filename, title="Dask Profile")
                 save(figure, title="Dask Profile", filename=filename)
             return (state, figure)
 
@@ -3858,7 +3857,7 @@ class Client(Node):
         count=None,
         plot=False,
         filename="task-stream.html",
-        resources=INLINE,
+        bokeh_resources=None,
     ):
         """ Get task stream data from scheduler
 
@@ -3887,9 +3886,8 @@ class Client(Node):
             If plot == 'save' then save the figure to a file
         filename: str (optional)
             The filename to save to if you set ``plot='save'``
-        resources: bokeh.resources.Resources (optional)
+        bokeh_resources: bokeh.resources.Resources (optional)
             Specifies if the resource component is INLINE or CDN
-            Default is INLINE
 
         Examples
         --------
@@ -3929,7 +3927,7 @@ class Client(Node):
             count=count,
             plot=plot,
             filename=filename,
-            resources=resources,
+            bokeh_resources=bokeh_resources,
         )
 
     async def _get_task_stream(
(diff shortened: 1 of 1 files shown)


distributed/client.py
```diff
@@ -23,6 +23,8 @@
 import warnings
 import weakref
 
+from bokeh.resources import INLINE
```
jrbourbeau: `bokeh` is an optional dependency, so we tend to do all `bokeh`-related imports inside functions and methods. This allows us to raise `ImportError`s when users call a function that requires `bokeh`, e.g. `get_task_stream`, but not when doing other non-`bokeh` operations like `import distributed`

distributed/client.py
```diff
@@ -3933,9 +3951,15 @@ def get_task_stream(
             source, figure = task_stream_figure(sizing_mode="stretch_both")
             source.data.update(rects)
             if plot == "save":
-                from bokeh.plotting import save
-
-                save(figure, title="Dask Task Stream", filename=filename)
+                from bokeh.plotting import save, output_file
+
+                output_file(filename=filename, title="Dask Performance Report")
```
jrbourbeau:
```suggestion
                output_file(filename=filename, title="Dask Task Stream")
```

distributed/client.py
```diff
@@ -3850,7 +3852,13 @@ def collections_to_dsk(collections, *args, **kwargs):
...
 
     def get_task_stream(
-        self, start=None, stop=None, count=None, plot=False, filename="task-stream.html"
+        self,
+        start=None,
+        stop=None,
+        count=None,
+        plot=False,
+        filename="task-stream.html",
+        resources=INLINE,
```
jrbourbeau: Could you please change this `bokeh_resources` to avoid confusion with the `resources` parameter that's commonly used throughout the codebase for specifying worker resources

jrbourbeau: IIUC the default value for `resources` can be configured using [environment variables](https://docs.bokeh.org/en/latest/docs/reference/settings.html#resources). Do you know if passing `INLINE` here will take priority over any user-configured environment variables? 

If so, we could use `bokeh_resources=None` as the default value here. That way bokeh resources can still be configured from `get_task_stream`, and we continue to fallback to `bokeh`'s default behavior, but we don't override user-configured `bokeh` environment variables.

prasunanand:
@jrbourbeau I have made the changes. Please review.

distributed/client.py
```diff
@@ -3933,9 +3949,15 @@ def get_task_stream(
...
-
-                save(figure, title="Dask Task Stream", filename=filename)
+                from bokeh.plotting import save, output_file
+
+                output_file(filename=filename, title="Dask Task Stream")
+                save(
+                    figure,
+                    title="Dask Task Stream",
+                    filename=filename,
+                    bokeh_resources=bokeh_resources,
```
jrbourbeau: Here we should just use `resources=` in bokeh's `save` function:

```suggestion
                    resources=bokeh_resources,
```

distributed/client.py
```diff
@@ -3933,9 +3949,15 @@ def get_task_stream(
...
             if plot == "save":
-                from bokeh.plotting import save
-
-                save(figure, title="Dask Task Stream", filename=filename)
+                from bokeh.plotting import save, output_file
+
+                output_file(filename=filename, title="Dask Task Stream")
+                save(
+                    figure,
+                    title="Dask Task Stream",
```
jrbourbeau: We can remove the `title` in `save` since it's included in `output_file` now

```suggestion
```

distributed/client.py
```diff
@@ -3367,8 +3367,9 @@ def profile(
                 filename = "dask-profile.html"
 
             if filename:
-                from bokeh.plotting import save
+                from bokeh.plotting import output_file, save
 
+                output_file(filename=filename, title="Dask Profile")
                 save(figure, title="Dask Profile", filename=filename)
```
jrbourbeau: Same thing here with `title`

```suggestion
                save(figure, filename=filename)
```

PR #3664, opened 2020-04-01
More documentation for Semaphore

fjetter:
This extends the documentation a bit further

cc @lr4d

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -204,6 +204,24 @@ class SemaphoreExtension:
 class Semaphore:
     """ Semaphore
 
+    This semaphore will track leases on the scheduler which can be acquired and
+    released by an instance of this class. If the maximum amount of leases are
+    already acquired, it is not possible to acquire more and the caller waits
+    until another lease has been released.
+
+    The lifetime of a lease are coupled to the ``Client`` it was acquired with.
+    Once the Client goes out of scope, the leases associated to it are freed.
+    This behaviour can be controlled with the parameter
+    `distributed.scheduler.locks.lease-validation-interval`.
+
+    A noticeable difference to the Semaphore of the python standard library is
+    that this implementation does not allow to release more often than it was
+    acquired. If this happens, a warning is emitted but the internal state is
+    not modified.
+
+    This implementation is still in an experimental state and subtle changes in
+    behaviour may occur without any change in the major version of this library.
+
     Parameters
     ----------
     max_leases: int (optional)
@@ -222,6 +240,7 @@ class Semaphore:
     --------
     >>> from distributed import Semaphore
     >>> sem = Semaphore(max_leases=2, name='my_database')
+    >>>
     >>> def access_resource(s, sem):
     >>>     # This automatically acquires a lease from the semaphore (if available) which will be
     >>>     # released when leaving the context manager.
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -212,8 +212,9 @@ class Semaphore:
 
     The lifetime of a lease is coupled to the ``Client`` it was acquired with.
     Once the Client goes out of scope, the leases associated to it are freed.
-    This behavior can be controlled with the parameter
-    `distributed.scheduler.locks.lease-validation-interval`.
+    This behavior can be controlled with the
+    ``distributed.scheduler.locks.lease-validation-interval`` configuration
+    option.
 
     A noticeable difference to the Semaphore of the python standard library is
     that this implementation does not allow to release more often than it was
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -204,14 +204,15 @@ class SemaphoreExtension:
 class Semaphore:
     """ Semaphore
 
-    This semaphore will track leases on the scheduler which can be acquired and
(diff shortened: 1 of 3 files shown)


distributed/semaphore.py
```diff
@@ -204,6 +204,24 @@ def close(self, comm=None, name=None):
 class Semaphore:
     """ Semaphore
 
+    This semaphore will track leases on the scheduler which can be acquired and
```
lr4d, line 207:
```suggestion
    This `semaphore <https://en.wikipedia.org/wiki/Semaphore_(programming)>`_
    will track leases on the scheduler which can be acquired and
```

distributed/semaphore.py
```diff
@@ -204,6 +204,24 @@ def close(self, comm=None, name=None):
...
+    released by an instance of this class. If the maximum amount of leases are
+    already acquired, it is not possible to acquire more and the caller waits
+    until another lease has been released.
+
+    The lifetime of a lease are coupled to the ``Client`` it was acquired with.
```
gforsyth, line 212:
```suggestion
    The lifetime of a lease is coupled to the ``Client`` it was acquired with.
```

distributed/semaphore.py
```diff
@@ -204,6 +204,24 @@ def close(self, comm=None, name=None):
...
+    Once the Client goes out of scope, the leases associated to it are freed.
+    This behaviour can be controlled with the parameter
```
gforsyth, line 214:
```suggestion
    This behavior can be controlled with the parameter
```
both are valid spellings but the rest of the dask docs use the American version

distributed/semaphore.py
```diff
@@ -204,6 +204,24 @@ def close(self, comm=None, name=None):
...
+    `distributed.scheduler.locks.lease-validation-interval`.
+
+    A noticeable difference to the Semaphore of the python standard library is
+    that this implementation does not allow to release more often than it was
+    acquired. If this happens, a warning is emitted but the internal state is
+    not modified.
+
+    This implementation is still in an experimental state and subtle changes in
+    behaviour may occur without any change in the major version of this library.
```
gforsyth, line 223:
```suggestion
    behavior may occur without any change in the major version of this library.
```

distributed/semaphore.py
```diff
@@ -204,6 +204,25 @@ def close(self, comm=None, name=None):
...
+    This `semaphore <https://en.wikipedia.org/wiki/Semaphore_(programming)>`_
+    will track leases on the scheduler which can be acquired and
+    released by an instance of this class. If the maximum amount of leases are
+    already acquired, it is not possible to acquire more and the caller waits
+    until another lease has been released.
+
+    The lifetime of a lease is coupled to the ``Client`` it was acquired with.
+    Once the Client goes out of scope, the leases associated to it are freed.
+    This behavior can be controlled with the parameter
+    `distributed.scheduler.locks.lease-validation-interval`.
```
jrbourbeau, line 216: Could you update "with the parameter `distributed.scheduler.locks.lease-validation-interval`" to "with the ``distributed.scheduler.locks.lease-validation-interval`` configuration option". Note that we should use two backticks (``) for code highlighting in sphinx

dhirschfeld:
Hi, just read the docs here and wanted to confirm that the `Semaphore` class as-is won't enable [`fire_and_forget`](https://github.com/dask/distributed/blob/66fe0acf041add720d5cc9d9d1ed99d9e90e7e2e/distributed/client.py#L4489) scenarios?

i.e. I would like to spin up a process, fire off 100 tasks to the scheduler using a `Semaphore` to limit the maximum parallelism to 10. The process would then call `fire_and_forget` on the futures and exit.

Once the process which created the tasks exits I'd like them to continue running until completion and to still be bound by the semaphore. From the documentation here it sounds like that once the process exits the tasks will no longer be bound by the semaphore and so will DDOS my database.

Is that correct? i.e. is my desired use-case not supported by the current implementation?

dhirschfeld:
> is my desired use-case not supported by the current implementation

That's fine if so - just trying to understand if I can move my custom code to use a `Semaphore`

lr4d:
@dhirschfeld It should still work fine for your use case. 

> From the documentation here it sounds like that once the process exits the tasks will no longer be bound by the semaphore and so will DDOS my database.

The tasks will still be bound by the semaphore, as the state of the semaphore lies on the scheduler side.
The documentation refers to a semaphore _lease_ being released when the process (referred to as `client`) who acquired the lease disappears. This _lease_ can be thought of as being part of an internal counter at the scheduler side, which ensures that there is an upper bound to the amount of parallelism (i.e. the parameter `max_leases`) you want to allow.

If you want to be sure though, I encourage you to try out the semaphore using `fire_and_forget` for some trivial tasks (e.g. sleeping for a while) and exiting the client before execution has finised, and see what happens

lr4d:
@dhirschfeld Since this feature is still experimental and there may be some (mostly internal) changes upcoming, it would be great if you could contribute a unit test for your use case so that we make sure that it remains supported

dhirschfeld:
Thanks @lr4d - good to know! I just misunderstood the documentation. I'll be giving it a go as soon as I can find the time and am happy to report back if everything works as expected... or otherwise :)

PR #3666, opened 2020-04-01
Get CUDA context to finalize Numba `DeviceNDArray`

jakirkham:
Related to issue ( https://github.com/rapidsai/dask-cuda/issues/176 )
Follow-up to PR ( https://github.com/dask/distributed/pull/3240 )

This ensures that when a Numba `DeviceNDArray` is used to back any kind of CUDA memory allocation (either for receiving CUDA frames via UCX or restoring spilled data to device), we make sure to have a valid context before finalizing the object. Should make sure all CUDA objects are handled correctly. Also supplants the need for doing this with CuPy `ndarray`s in a special case fashion.

cc @pentschev @madsbk @rjzamora

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -6,6 +6,7 @@ See :ref:`communications` for more.
 .. _UCX: https://github.com/openucx/ucx
 """
 import logging
+import weakref
 
 import dask
 import numpy as np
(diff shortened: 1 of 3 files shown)


quasiben:
Merging in. Thanks @jakirkham

PR #3670, opened 2020-04-03
Update Scheduler.rebalance return value when data is missing

jrbourbeau:
This PR improves error handling in `Scheduler.rebalance` when keys which are being rebalanced are found to be missing on workers. Currently when we attempt to return the missing data keys here

https://github.com/dask/distributed/blob/4f11509b844c3569f164704983bb0affd009dd4c/distributed/scheduler.py#L3106

we get the following error:

```
TypeError: can only concatenate list (not "dict") to list
```

because `r["keys"]` is a dictionary. We update things here to return a tuple of keys.

<details>
<summary>Example to trigger missing data in rebalance:</summary>

```python
import time
import random
from itertools import count

import dask.array as da

with Client() as client:
    counter = count()
    while next(counter) < 10:
        x = da.random.normal(size=100, chunks=1)
        x = x.persist()
        
        time.sleep(random.random() * 0.1)
        x = None
        client.rebalance()
```

</details>

Note: there were also several trailing commas that `black` wanted to add

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -587,7 +587,7 @@ class Client(Node):
         deserializers=None,
         extensions=DEFAULT_EXTENSIONS,
         direct_to_workers=None,
-        **kwargs
+        **kwargs,
     ):
         if timeout == no_default:
             timeout = dask.config.get("distributed.comm.timeouts.connect")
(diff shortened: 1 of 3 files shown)


mrocklin:
Thanks @jrbourbeau .  Should we add some variant of the example as a test?

mrocklin:
Thanks @jrbourbeau

PR #3671, opened 2020-04-03
don't make task graphs too big

### LINKED ISSUE #2979: Dashboard views that may crash your browser should be gated
If I have a dask task graph with 200k tasks then clicking on the "tasks" tab of dask dasboard will lead to doom. Specifically it will just hang my browser. This is particularly disastrous if you're using dask-jupyterlab because then you break your jupyterlab session. 

I'm wondering if we can implement a gate that, for a given threshold (say 50k for arguments sake), delays the loading of the plot and presents the user a message saying "this view has more than 50k items, are you sure you want to display" or something like that.

martindurant:
Fixes #2979

I don't yet know how to test this (tested manually).

I worry that there may be potential bad behaviour when the number of tasks hovers around the limit, as allowed updates will always be complete datasets instead of incremental.

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -35,7 +35,6 @@ from bokeh.transform import factor_cmap, linear_cmap
 from bokeh.io import curdoc
 import dask
 from dask.utils import format_bytes, key_split
-from dask.config import config
 from tlz import pipe
 from tlz.curried import map, concat, groupby
 from tornado import escape
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -35,6 +35,7 @@ from bokeh.transform import factor_cmap, linear_cmap
 from bokeh.io import curdoc
 import dask
 from dask.utils import format_bytes, key_split
+from dask.config import config
 from tlz import pipe
 from tlz.curried import map, concat, groupby
 from tornado import escape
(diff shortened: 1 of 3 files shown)


martindurant:
(heh, breaks tests, because I set a very small default value to play with)

distributed/dashboard/components/scheduler.py
```diff
@@ -1193,6 +1194,7 @@ def update(self):
 
     @without_property_validation
     def add_new_nodes_edges(self, new, new_edges, update=False):
+        max_items = config.get('distributed.dashboard.graph_max_items', 2)
```
mrocklin, line 1197: For config we typically do the following by convention

```suggestion
        max_items = dask.config.get('distributed.dashboard.graph_max_items', 2)
```

distributed/dashboard/components/scheduler.py
```diff
@@ -35,6 +35,7 @@
 from bokeh.io import curdoc
 import dask
 from dask.utils import format_bytes, key_split
+from dask.config import config
```
mrocklin, line 38:
```suggestion
```

distributed/distributed.yaml
```diff
@@ -129,6 +129,7 @@ distributed:
   dashboard:
     link: "{scheme}://{host}:{port}/status"
     export-tool: False
+    graph_max_items: 500  # maximum number of tasks to try to plot in graph view
```
mrocklin, line 132: The config generally uses hyphens rather than underscores by convention.

mrocklin: I recommend 5000 as a default value.  Hopefully we can set it to something smaller during a test.  Any thoughts on how best to test this?

martindurant: Yes, the manner in which this caused current tests to fail is illustrative :)

PR #3672, opened 2020-04-04
Add zoom tools to profile plots

### LINKED ISSUE #3626: Adding zoom tools to performance_report
In some cases ["the devil is in the detail"]( https://en.wikipedia.org/wiki/The_devil_is_in_the_detail ), it would be useful to have the zoom tooltips included in other [`performance_report`]( https://distributed.dask.org/en/latest/diagnosing-performance.html#performance-reports ) panels to allow closer investigation of where time is being spent.

jrbourbeau:
Closes #3626

cc @jakirkham

### DIFF
--- a/distributed/profile.py
+++ b/distributed/profile.py
@@ -383,7 +383,7 @@ def plot_figure(data, **kwargs):
 
     source = ColumnDataSource(data=data)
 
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @jrbourbeau

PR #3673, opened 2020-04-05
Run preload at import, start, and teardown

mrocklin:
Previously we only ran preload scripts after the server had started.
This made it difficult to modify the server before certain actions had
taken place.  For example, a test in this PR registers a new Comm
backend for the server to use.  This would not have been possible
before.

Now we run different preload functions when

1.  we first instantiate the server
2.  we first start the server
3.  we close the server (we used to run teardown at `atexit` time

cc @pentschev I think that you asked for this at one point.

### DIFF
--- a/distributed/preloading.py
+++ b/distributed/preloading.py
@@ -111,7 +111,7 @@ def _import_module(name, file_dir=None):
     }
 
 
-def on_creation(names, file_dir=None):
+def on_creation(names, file_dir: str = None) -> dict:
     """ Imports each of the preload modules
 
     Parameters
@@ -127,13 +127,13 @@ def on_creation(names, file_dir=None):
     return {name: _import_module(name, file_dir=file_dir) for name in names}
 
 
-async def on_start(names, dask_server=None, file_dir: str = None, argv=None):
-    """ Imports modules, handles `dask_setup` and `dask_teardown`.
+async def on_start(modules: dict, dask_server=None, argv=None):
+    """ Run when the server finishes its start method
 
     Parameters
     ----------
-    names: list of strings
-        Module names or file paths
+    modules: Dict[str, module]
+        The imported modules, from on_creation
     dask_server: dask.distributed.Server
         The Worker or Scheduler
     argv: [string]
@@ -141,7 +141,7 @@ async def on_start(names, dask_server=None, file_dir: str = None, argv=None):
     file_dir: string
         Path of a directory where files should be copied
     """
-    for name, interface in on_creation(names, file_dir).items():
+    for name, interface in modules.items():
         dask_setup = interface.get("dask_setup", None)
 
         if dask_setup:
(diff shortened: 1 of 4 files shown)


mrocklin:

distributed/preloading.py
```diff
@@ -111,7 +111,7 @@ def _import_module(name, file_dir=None):
     }
 
 
-def preload_modules(names, parameter=None, file_dir=None, argv=None):
+def on_creation(names, file_dir=None):
```
jrbourbeau, line 114: We can remove `parameter` and `argv` from the docstring for this function

distributed/preloading.py
```diff
@@ -128,11 +128,25 @@ def preload_modules(names, parameter=None, file_dir=None, argv=None):
     if isinstance(names, str):
         names = [names]
 
-    for name in names:
-        interface = _import_module(name, file_dir=file_dir)
+    return {name: _import_module(name, file_dir=file_dir) for name in names}
 
+
+async def on_start(names, parameter=None, file_dir=None, argv=None):
+    """ Imports modules, handles `dask_setup` and `dask_teardown`.
```
jrbourbeau, line 135:
```suggestion
    """ Imports modules and handles ``dask_setup``.
```

distributed/preloading.py
```diff
@@ -128,11 +128,25 @@ def preload_modules(names, parameter=None, file_dir=None, argv=None):
...
+
+    Parameters
+    ----------
+    names: list of strings
+        Module names or file paths
+    parameter: object
+        Parameter passed to `dask_setup` and `dask_teardown`
```
jrbourbeau, line 142:
```suggestion
        Parameter passed to ``dask_setup``
```

distributed/preloading.py
```diff
@@ -141,8 +155,25 @@ def preload_modules(names, parameter=None, file_dir=None, argv=None):
...
             else:
-                dask_setup(parameter)
+                future = dask_setup(parameter)
+                if inspect.isawaitable(future):
+                    await future
                 logger.info("Run preload setup function: %s", name)
 
+
+async def on_teardown(names, parameter=None, file_dir=None):
+    """ Imports modules, handles `dask_setup` and `dask_teardown`.
```
jrbourbeau, line 165:
```suggestion
    """ Imports modules and handles ``dask_teardown``.
```

distributed/preloading.py
```diff
@@ -141,8 +155,25 @@ def preload_modules(names, parameter=None, file_dir=None, argv=None):
...
+
+    Parameters
+    ----------
+    names: list of strings
+        Module names or file paths
+    parameter: object
+        Parameter passed to `dask_setup` and `dask_teardown`
```
jrbourbeau, line 172:
```suggestion
        Parameter passed to ``dask_teardown``
```

distributed/preloading.py
```diff
@@ -1,4 +1,4 @@
-import atexit
```
jrbourbeau, line 1: +1 I like that we're calling the teardown code explicitly in in `close` methods now instead of using `atexit`

mrocklin: Yeah, I figure that if people want to run atexit they can use `atexit` in their preload module itself.  We don't have to mandate that structure.

mrocklin:
Thanks for the review @jrbourbeau .  I believe I've handled everything.  I also changed things up slightly so that we're now passing around the imported modules.

I plan to follow up on this PR with a PR that supports preloads for the Nanny.

PR #3674, opened 2020-04-05
Pass through connection/listen_args as splatted keywords

mrocklin:
Previously we would accept a specific keyword connection_args or
listen_args for security information like `ssl_context`.

Now we pass in these keywords directly,
and let the listener handle them without intermediary.

This different is mostly cosmetic, but does allow the passing of
other keywords to third-party listeners if needed.

### DIFF
--- a/distributed/cli/tests/test_dask_scheduler.py
+++ b/distributed/cli/tests/test_dask_scheduler.py
@@ -32,7 +32,9 @@ def test_defaults(loop):
         @gen.coroutine
         def f():
             # Default behaviour is to listen on all addresses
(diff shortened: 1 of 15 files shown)


mrocklin:

mrocklin:
@jcrist as far as I'm aware you're the only other person with a Comm in another library.  I took a brief look at dask-gateway and I think that we should be fine here, but I thought I'd ping you anyway.

mrocklin:
Yeah, I guess we could have done this before by moving things through the
connection_args keyword.  I guess in my mind connection/listen_args are
mostly for things like security.  I'll step back to saying that this is
mostly a cosmetic change.

Issue #3645, opened 2020-03-27
Centralize HTTPServer in Server


mrocklin:
Currently we use the HTTPServer hidden in the Bokeh server, if we create the dashboard.  We then hang various additional routes on the Tornado web application in this server for additional things, like the info pages, and prometheus.  

I suggest that we instead create a default HTTP server for the Dask server, and then attach new Tornado applications to that server as needed (like bokeh, prometheus, ...)  This might be a bit more aligned to how we actually think about things.  Additionally, if we needed to set things up like SSL, it's a bit simpler to do this within the core dask code rather than route various keywords through the bokeh code.

cc @jacobtomlinson @jcrist 

I'm making Bokeh a little better about handling this in https://github.com/bokeh/bokeh/pull/9833 , but short term we would need to bring some of that logic into our codebase (unless we wanted to depend on latest release Bokeh, which I don't think that we want to do).

mrocklin:
I'm getting a bit further along in this.  I propose the following:

Schedulers and Workers run a Tornado HTTPServer by default.  This hosts info pages, health checks, Prometheus metrics, and we make it easy for people to add more routes with configuration  We move this code from `/dashboard` to `/http`.

For the Bokeh dashboard we no longer create a Bokeh Server, but instead make a Bokeh Tornado Application (called BokehTornado in that library), and register that application with the existing HTTP Server.  It's a bit janky to add a new application to a running server, but it doesn't seem impossible.

<details>

<summary> Example attempt </summary>

```python
from tornado import web
import tornado.httputil
from typing import Any, List


class RoutingApplication(web.Application):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.applications = []

    def find_handler(self, request: tornado.httputil.HTTPServerRequest, **kwargs: Any) ->                '_HandlerDelegate':
        handler = super().find_handler(request, **kwargs)
        if handler and not issubclass(handler.handler_class, web.ErrorHandler):
            return handler
        else:
            for app in self.applications:
                handler = app.find_handler(request, **kwargs) or handler
                if handler and not issubclass(handler.handler_class, web.ErrorHandler):
                    break
            return handler

    def add_application(self, application: web.Application):
        self.applications.append(application)
```

</details>

PR #3676, opened 2020-04-06
Use relative URL in scheduler dashboard

nsmith-:
As done for other examples of `OpenURL` in the same file. This is important for cases where the dashboard is exposed at a non-root prefix (e.g. via the `--dashboard-prefix` option to `dask-scheduler`)


mrocklin:
Grand.  Thanks @nsmith- !

You may also be interested in #3645 where a lot of this gets changed around.  I *think* that everything should be fine, but if you have an environment that cares about this, I'd love to have you try things out.

PR #3678, opened 2020-04-07
Support preload modules in Nanny

mrocklin:
Today we can run preload modules on workers or the scheduler, but not on the Nanny.

Now we can also run preload scripts on the nanny

```
dask-worker ... --preload-nanny my_module
```

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -216,6 +216,14 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 @click.argument(
     "preload_argv", nargs=-1, type=click.UNPROCESSED, callback=validate_preload_argv
 )
+@click.option(
+    "--preload-nanny",
+    type=str,
+    multiple=True,
+    is_eager=True,
+    help="Module that should be loaded by each nanny "
+    'like "foo.bar" or "/path/to/foo.py"',
+)
 @click.version_option()
 def main(
     scheduler,
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -248,6 +248,7 @@ def main(
     tls_key,
     dashboard_address,
     worker_class,
+    preload_nanny,
     **kwargs
 ):
     g0, g1, g2 = gc.get_threshold()  # https://github.com/dask/distributed/issues/1653
@@ -357,6 +358,7 @@ def main(
     worker_class = import_term(worker_class)
     if nanny:
         kwargs["worker_class"] = worker_class
+        kwargs["preload_nanny"] = preload_nanny
 
     if nanny:
         kwargs.update({"worker_port": worker_port, "listen_address": listen_address})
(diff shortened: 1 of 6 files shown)


distributed/nanny.py
```diff
@@ -247,6 +263,10 @@ def local_dir(self):
         await self.listen(self._start_address, listen_args=self.listen_args)
         self.ip = get_address_host(self.address)
 
+        await preloading.on_start(
+            self._preload_modules, self, argv=self.preload_argv,
```
jrbourbeau, line 267:
```suggestion
            self._preload_modules, self, argv=self.preload_nanny_argv,
```

distributed/worker.py
```diff
@@ -452,7 +452,7 @@ def __init__(
         # Target interface on which we contact the scheduler by default
         # TODO: it is unfortunate that we special-case inproc here
         if not host and not interface and not scheduler_addr.startswith("inproc://"):
-            host = get_ip(get_address_host(scheduler_addr))
+            host = get_ip(get_address_host(scheduler_addr.split("://")[-1]))
```
jrbourbeau, line 455: Why was this change needed?

distributed/cli/dask_worker.py
```diff
@@ -216,6 +216,14 @@
...
     "preload_argv", nargs=-1, type=click.UNPROCESSED, callback=validate_preload_argv
 )
+@click.option(
+    "--preload-nanny",
+    type=str,
+    multiple=True,
+    is_eager=True,
+    help="Module that should be loaded by each nanny "
+    'like "foo.bar" or "/path/to/foo.py"',
+)
```
jrbourbeau, line 226: Currently `preload_nanny_argv` can't be specified on the command line. Do you have any thoughts on how we could support a `--preload-nanny-argv` CLI option? To be clear, I suspect this may be difficult with our current `click` setup and don't think the lack of a `--preload-nanny-argv` option should stop us from merging this PR

distributed/worker.py
mrocklin: It's not entirely related.  I was testing this with another workload and ran into a small bug here.

distributed/cli/dask_worker.py
mrocklin: Yeah, I wasn't sure how to pass things through from click.  My sense is that we probably have to point users to the configuration to handle this.  

I've added https://github.com/dask/dask/pull/6077 to help with documentation here.

distributed/distributed.yaml
```diff
@@ -71,6 +71,10 @@ distributed:
       pause: 0.80  # fraction at which we pause worker threads
       terminate: 0.95  # fraction at which we terminate the worker
 
+  nanny:
+    preload: []
```
martindurant, line 75: Most entries have a comment describing them

mrocklin: Thanks for flagging this.

I've gone ahead and added brief comments that mostly just point to documentation (I don't think that we can reasonably describe this in one line).  Ben and I are planning to play with larger descriptions for documentation later this week though.

PR #3688, opened 2020-04-09
Fix propagating inherit config in SSHCluster for non-bash shells

### LINKED ISSUE #3687: SSHCluster fails in non-bash compatible shells
SSHCluster fails to propagate `DASK_INTERNAL_INHERIT_CONFIG` variable to scheduler and workers 
 using non-bash compatible shells. 

I suggest to pass this in the `env` keyword in `asyncssh.connection.create_process` instead of embedding it in the process command string to avoid handling different shell syntax for setting variables.

abduhbm:
Fixes #3687

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -86,10 +86,26 @@ class Worker(Process):
         import asyncssh  # import now to avoid adding to module startup time
 
         self.connection = await asyncssh.connect(self.address, **self.connect_options)
+
+        result = await self.connection.run("uname")
+        if result.exit_status == 0:
+            set_env = 'env DASK_INTERNAL_INHERIT_CONFIG="{}"'.format(
+                serialize_for_cli(dask.config.global_config)
+            )
+        else:
+            result = await self.connection.run("cmd /c ver")
+            if result.exit_status == 0:
+                set_env = "set DASK_INTERNAL_INHERIT_CONFIG={} &&".format(
+                    serialize_for_cli(dask.config.global_config)
+                )
+            else:
+                raise Exception(
+                    "Worker failed to set DASK_INTERNAL_INHERIT_CONFIG variable "
+                )
+
         cmd = " ".join(
             [
-                'env DASK_INTERNAL_INHERIT_CONFIG="%s"'
-                % serialize_for_cli(dask.config.global_config),
+                set_env,
                 sys.executable,
                 "-m",
                 self.worker_module,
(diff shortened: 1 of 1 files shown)


martindurant:
Seems to specifically break test `test_config_inherited_by_subprocess`

abduhbm:
@martindurant the environment variable was silently rejected ... now this uses `env` command before starting scheduler and worker. 
Tested on `bash` and `tcsh` with no issues.

gforsyth:
Do we explicitly `dask-ssh` usage on Windows?  I ask because it's less likely to have `env` installed.  It's present on OSX by default and I don't know a linux distro that doesn't include `coreutils`.  Re-implementing `which` to check for `env` is a bit much to do, but maybe we want to do a retry without `env` if the first go fails?

martindurant:
> a retry without env if the first go fails

This should be an exception with a clear message ("can't pass an env on Windows without installing env"). Just dropping the env seems like a good way to get strange errors.

abduhbm:
Added a retry to create scheduler and worker processes without env command for setting inherit config.

distributed/deploy/ssh.py
```diff
@@ -86,25 +86,37 @@ def __init__(
...
+                "-m",
+                self.worker_module,
+                self.scheduler,
+                "--name",
+                str(self.name),
+            ]
+            + cli_keywords(self.kwargs, cls=_Worker, cmd=self.worker_module)
         )
 
+        self.proc = await self.connection.create_process(cmd)
```
martindurant, line 103: Should have a timeout?

distributed/deploy/ssh.py
```diff
@@ -86,25 +86,37 @@ def __init__(
...
         while True:
             line = await self.proc.stderr.readline()
+            # retry without env command
+            if not line.strip() and retry_flag:
+                retry_flag = False
+                logger.debug("Retry creating process without env command")
+                self.proc = await self.connection.create_process(
+                    cmd.replace(
+                        "env DASK_INTERNAL_INHERIT_CONFIG=",
+                        "DASK_INTERNAL_INHERIT_CONFIG=",
```
martindurant, line 116: Since the original suggestion was about windows' lack of `env`, does this do the right thing?

abduhbm: Not really that familiar with Windows environment ... should I add another retry to run `set DASK_INTERNAL_CONFIG=...`? Can I determine the OS type for the host using `asyncssh` after establishing the connection?

distributed/deploy/ssh.py
abduhbm: There is no timeout parameter for asyncssh connection. Please see the note on `asyncssh.connect`:

```
.. note:: Unlike :func:`socket.create_connection`, asyncio calls
                 to create a connection do not support a `timeout`
                 parameter. However, asyncio calls can be wrapped in a
                 call to :func:`asyncio.wait_for` or :func:`asyncio.wait`
                 which takes a timeout and provides equivalent functionality.
```
Something like below ... Should this be implemented here?

```python
try:
    self.connection = yield from asyncio.wait_for(asyncssh.connect(self.address, **self.connect_options), timeout=timeout)
    # do stuff with conn
    conn.close()
except TimeoutError:
    # handle connection timeout
```

distributed/deploy/ssh.py
martindurant: Ah, actually good point, probably an ssh host isn't windows. 
I would just catch the error and repot it the first time in that case, assume that the ssh host *must* have env to work (and not force to retry due to any other error).

distributed/deploy/ssh.py
martindurant: Sorry, I don't know the right way to do it either, but would have thought a timeout would be a good idea here. Perhaps someone else has best practices on this.

abduhbm:
@martindurant I added a step to check OS and set the env based on it. This should handle Windows too.

martindurant:
OK, this is going in now - thank you @abduhbm

PR #3689, opened 2020-04-09
Dask-serialize dicts longer than five elements

rjzamora:
Some dask.dataframe/dask_cudf shuffling algorithms utilize tasks that produce a dictionary of pandas/cudf DataFrame objects.  Since these dictionaries are typically larger than five elements, they are usually pickled when the output needs to be transferred between workers.  For the cudf-backed shuffle algorithm, this can seriously degrade performance (changing the dominant communication mechanism from IPC to TPC on systems with NVLink support).

This PR increases the minimum size of a list/set/dict for which each element will be separately serialized (from 5 to 64).  If the length is >5, the iterative serialization is only performed if the "first"  element is dask-serializable.  The check clearly doesnt assert that *all* items in the list/set/dict are dask-serializable, but always capture the case when all elements are.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -132,8 +132,29 @@ def serialize(x, serializers=None, on_error="message", context=None):
     if isinstance(x, Serialized):
         return x.header, x.frames
 
+    # Check for "dask"-serializable data in large collections.
+    # We will iterate through small collections by default (<=5).
+    # We will not iterate through any collections larger than 64
+    supported = False
+    if type(x) in (list, set, tuple):
+        supported = len(x) <= 5
+        if not supported and len(x) <= 64:
+            try:
+                dask_serialize.dispatch(type(x[0]))
+                supported = True
+            except TypeError:
+                pass
+    elif type(x) is dict:
+        supported = len(x) <= 5
+        if not supported and len(x) <= 64:
+            try:
+                dask_serialize.dispatch(type(next(iter(x.items()))[1]))
+                supported = True
+            except TypeError:
+                pass
+
     # Determine whether keys are safe to be serialized with msgpack
-    if type(x) is dict and len(x) <= 5:
+    if type(x) is dict and supported:
         try:
             msgpack.dumps(list(x.keys()))
         except Exception:
@@ -143,9 +164,9 @@ def serialize(x, serializers=None, on_error="message", context=None):
 
     if (
         type(x) in (list, set, tuple)
-        and len(x) <= 5
+        and supported
         or type(x) is dict
-        and len(x) <= 5
+        and supported
         and dict_safe
     ):
         if isinstance(x, dict):
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -132,15 +132,15 @@ def serialize(x, serializers=None, on_error="message", context=None):
     if isinstance(x, Serialized):
         return x.header, x.frames
 
-    # Check for "dask"-serializable data in large collections.
-    # We will iterate through small collections by default (<=5).
-    # We will not iterate through any collections larger than 64
+    # Check for "dask"-serializable data in large data structures.
+    # We will iterate through small structures by default (<=5).
+    # We will not iterate through any structures larger than 64
     supported = False
     if type(x) in (list, set, tuple):
         supported = len(x) <= 5
         if not supported and len(x) <= 64:
             try:
-                dask_serialize.dispatch(type(x[0]))
+                dask_serialize.dispatch(type(next(iter(x))))
                 supported = True
             except TypeError:
                 pass
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -147,11 +147,12 @@ def serialize(x, serializers=None, on_error="message", context=None):
         return x.header, x.frames
 
     # Check for "dask"-serializable data in dict/list/set
-    supported = check_dask_serializable(x)
-    special_case = isinstance(x, list) and "pickle" not in serializers
+    supported = isinstance(x, list) and "pickle" not in serializers
+    if not supported:
+        supported = check_dask_serializable(x)
 
     # Determine whether keys are safe to be serialized with msgpack
-    if type(x) is dict and (supported or special_case):
+    if type(x) is dict and supported:
         try:
             msgpack.dumps(list(x.keys()))
         except Exception:
@@ -161,9 +162,9 @@ def serialize(x, serializers=None, on_error="message", context=None):
 
     if (
         type(x) in (list, set, tuple)
-        and (supported or special_case)
+        and supported
         or type(x) is dict
-        and (supported or special_case)
+        and supported
         and dict_safe
     ):
         if isinstance(x, dict):
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -91,19 +91,24 @@ register_serialization_family("error", None, serialization_error_loads)
 
 
 def check_dask_serializable_collection(x):
+    supported = False
     if type(x) in (list, set, tuple):
-        try:
-            dask_serialize.dispatch(type(next(iter(x))))
-            return True
-        except TypeError:
-            pass
+        supported = len <= 5
+        if not supported:
+            try:
+                dask_serialize.dispatch(type(next(iter(x))))
+                supported = True
+            except TypeError:
+                pass
     elif type(x) is dict:
-        try:
-            dask_serialize.dispatch(type(next(iter(x.items()))[1]))
-            return True
-        except TypeError:
-            pass
-    return False
+        supported = len <= 5
+        if not supported:
+            try:
+                dask_serialize.dispatch(type(next(iter(x.items()))[1]))
+                supported = True
+            except TypeError:
+                pass
+    return supported
 
 
 def serialize(x, serializers=None, on_error="message", context=None):
@@ -152,7 +157,7 @@ def serialize(x, serializers=None, on_error="message", context=None):
     supported = check_dask_serializable_collection(x)
 
     # Determine whether keys are safe to be serialized with msgpack
-    if type(x) is dict and (supported or len(x) <= 5):
+    if type(x) is dict and supported:
         try:
             msgpack.dumps(list(x.keys()))
         except Exception:
@@ -162,9 +167,9 @@ def serialize(x, serializers=None, on_error="message", context=None):
 
     if (
         type(x) in (list, set, tuple)
-        and (supported or len(x) <= 5)
+        and supported
         or type(x) is dict
-        and (supported or len(x) <= 5)
+        and supported
         and dict_safe
     ):
         if isinstance(x, dict):
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -90,25 +90,21 @@ register_serialization_family("msgpack", msgpack_dumps, msgpack_loads)
 register_serialization_family("error", None, serialization_error_loads)
 
 
-def check_dask_serializable_collection(x):
-    supported = False
+def check_dask_serializable(x, serializers=None):
+    serializers = serializers or []
     if type(x) in (list, set, tuple):
-        supported = len(x) <= 5
-        if not supported:
-            try:
-                dask_serialize.dispatch(type(next(iter(x))))
-                supported = True
-            except TypeError:
-                pass
+        if type(x) is list and "pickle" not in serializers:
+            return True  # msgpack seems to get this wrong
+        return check_dask_serializable(next(iter(x)))
     elif type(x) is dict:
-        supported = len(x) <= 5
-        if not supported:
-            try:
-                dask_serialize.dispatch(type(next(iter(x.items()))[1]))
-                supported = True
-            except TypeError:
-                pass
-    return supported
+        return check_dask_serializable(next(iter(x.items()))[1])
+    else:
+        try:
+            dask_serialize.dispatch(type(x))
+            return True
+        except TypeError:
+            pass
+    return False
 
 
 def serialize(x, serializers=None, on_error="message", context=None):
@@ -154,7 +150,7 @@ def serialize(x, serializers=None, on_error="message", context=None):
         return x.header, x.frames
 
     # Check for "dask"-serializable data in dict/list/set
-    supported = check_dask_serializable_collection(x)
+    supported = check_dask_serializable(x, serializers=serializers)
 
     # Determine whether keys are safe to be serialized with msgpack
     if type(x) is dict and supported:
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -93,7 +93,7 @@ register_serialization_family("error", None, serialization_error_loads)
 def check_dask_serializable_collection(x):
     supported = False
     if type(x) in (list, set, tuple):
-        supported = len <= 5
+        supported = len(x) <= 5
         if not supported:
             try:
                 dask_serialize.dispatch(type(next(iter(x))))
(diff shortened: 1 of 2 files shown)


rjzamora:
@pentschev - Any thoughts on using  logic like this to avoid pickling of the `group-shuffle` distionary output?

cc @beckernick @VibhuJawa

jakirkham:
cc @mrocklin

mrocklin:
```python
                 dask_serialize.dispatch(type(next(iter(x))))
```

This is a fun approach.  Really, what we want to say is "if this dictionary is a bunch of big-complex things that Dask knows how to handle, then please serailize each of them independently.  If it's just a big json-like blob, then please don't bother".  The previous check of `if len(d) > 5` was a very poor approximation of this check.  The check here of "does `dask_serialize` grok these objects in a special way seems a lot nicer to me.  If we can trust this approach it would be great to remove the `len(d) < 64` part of the check as well.  

How well does the `dask_serialize` approach work?  If I had a dict like `{"x": 1}` does this reliably not dive through the dict?  It might be worth pulling this functionality into a separate function, testing it with a bunch of cases that we know about 

```python
{"x": 1} -> False
{"x": np.ones(5)} -> True
{"a": 1, "x": np.ones(5)} -> True (maybe xfails today?)
{"x": [np.ones(5)} -> True
{("x", i): np.ones(5) for i in range(100)} -> True
...
```
And then make sure that it runs in constant time relative to the size of the dict.

In short, I like this approach if it works.  I'd love to see us drop the 64 length limit if we can.  I think that being able to do that would be easier if we had a function that we could all agree did the right thing all (most?) of the time, which would be easy to show through a nice test like this.

rjzamora:
> In short, I like this approach if it works. I'd love to see us drop the 64 length limit if we can. I think that being able to do that would be easier if we had a function that we could all agree did the right thing all (most?) of the time, which would be easy to show through a nice test like this.

Thanks for the feedback @mrocklin!  The approach will indeed not dive through the dict for the case you mentioned.  I will work on the test you suggested and confirm that the check runs in constant time.

mrocklin:
I think that you've done a constant time check by only checking the first
one. This means that some of the test cases will fail (like the one with an
integer and a numpy array in the dict), but makes things fast.

rjzamora:
Thanks again for your advice here @mrocklin - I moved the check into a standalone function (`check_dask_serializable_collection`), and added a test for the behavior you mentioned.  One difference from your example, however, is that I could not use a numpy array as a "dask-serializable" object.  As far as I can tell, there are no custom serialization mechanisms registered for numpy types.

rjzamora:
Okay - It seems that the removal of the `len <= 5` criteria is leading to two new test failures...

- `test_nested_types` is serializing `[[[x]]]`, and expecting to recursively iterate through the list (but no longer will).
- `test_pickle_safe` is serializing `[1, 2, 3]`, but only offering the `"msgpack"` serializer.  Since we are no longer iterating through the list, and recording the "collection" datatype, we seem to be erroneously converting from `list` to `tuple` during the dumps-loads round-trip.

rjzamora:
**Note**: I am adding back the `len <= 5` criteria for now (to allow tests to pass) - Ill need to follow up later

mrocklin:
For nested types it seems like the `check_dask_serializable` function might have to recurse if the subtype is list/dict/...

rjzamora:
> For nested types it seems like the check_dask_serializable function might have to recurse if the subtype is list/dict/...

Right - Not sure how else a "dask-serializable" object can be detected within a nested collection... Having a bit of trouble finding a balance here (between functionality and simplicity)

mrocklin:
Maybe something like this?  (I don't know really, I haven't thought too much about this

```python
def is_serializable(x):
    if isinstance(x, (tuple, list)) and x:
        return is_serializable(x[0])
    if isinstance(x, dict) and x:
        return is_serializable(toolz.first(x.values()))
    return dask_serialize.....(x)
```

rjzamora:
Sorry - I'm not seeing how that is very different from `check_dask_serializable`. It will still need to recurse, right?

mrocklin:
Oh I see, yes.  I apologize I hadn't looked at the recent code (pushing quickly through issues this morning)

> test_nested_types is serializing [[[x]]], and expecting to recursively iterate through the list (but no longer will).

O guess I'm surprised by this then

rjzamora:
> O guess I'm surprised by this then

Sorry - I was using github/CI a bit too much for debugging :) 

That test failure is no longer a problem (the recursion fixes it).  The only remaining issue is that I still need the `len <= 5` criteria to get all tests to pass. (There also seems to be a recent CI problem unrelated to these changes)

mrocklin:
What tests are failing due to the length check?  Is it possible to fix whatever is causing those tests to fail?

rjzamora:
> What tests are failing due to the length check? Is it possible to fix whatever is causing those tests to fail?

On my local system, it is only `distributed/tests/test_publish.py::test_pickle_safe` - The problem seems to be that a`msgpack` round-trip converts a list to a tuple.  I removed the `len <= 5` and added a `special_case` for when the "pickle" serializer is unavailable -- I will try to find time to look into the "root"/extent of the msgpack problem.

pentschev:
Sorry for the _very_ late reply here @rjzamora . I think your approach is great, nice that you managed to improve my very naive implementation, thanks for doing that! Also it seems that it passed now! :)

rjzamora:
@mrocklin - Do you think the current state of this PR is reasonable? (thanks again for your attention on this)

distributed/protocol/serialize.py
```diff
@@ -90,6 +90,20 @@ def register_serialization_family(name, dumps, loads):
...
 
 
+def check_dask_serializable(x):
+    if type(x) in (list, set, tuple) and len(x):
+        return check_dask_serializable(next(iter(x)))
+    elif type(x) is dict and len(x):
+        return check_dask_serializable(next(iter(x.items()))[1])
+    else:
+        try:
+            dask_serialize.dispatch(type(x))
```
jakirkham, line 100: Just a note, this requires that objects be registered with `dask_serialize` (as we don't check `cuda_serialize` for example). However we do register all CUDA objects with `dask_serialize` now and use this in other contexts (like [Dask-CUDA's spilling]( https://github.com/rapidsai/dask-cuda/blob/b5ae4fc872b057f9d71595fccaa6dfd61654e301/dask_cuda/device_host_file.py#L71 )). So this seems like a good enough check.

jakirkham:

distributed/protocol/serialize.py
rjzamora: Thanks for the note @jakirkham - The fact that you have cleaned up serialization and registered everything as `dask_serialize` definitely made my life easier here - Thanks! :)

distributed/protocol/tests/test_serialize.py
```diff
@@ -388,3 +389,27 @@ def _(x):
 
     header, frames = serialize([MyObj(), MyObj()])
     assert header["compression"] == [False, False]
+
+
+@pytest.mark.parametrize(
+    "test_pair",
```
mrocklin, line 395: You can replace this with two terms separated by a comma like the following:

```suggestion
    "data,is_serializable",
```

distributed/protocol/tests/test_serialize.py
```diff
@@ -388,3 +389,27 @@ def _(x):
...
+        pytest.param(
+            ([1, MyObj(5)], True),
+            marks=pytest.mark.xfail(reason="Only checks 0th element for now."),
+        ),
+        ([MyObj([0, 1, 2]), 1], True),
+        (tuple([MyObj(None)]), True),
+        ({("x", i): MyObj(5) for i in range(100)}, True),
+    ],
+)
+def test_check_dask_serializable(test_pair):
```
mrocklin, line 411: And then use those terms below

```suggestion
def test_check_dask_serializable(data, is_serializable):
```

distributed/protocol/serialize.py
```diff
@@ -90,6 +90,20 @@ def register_serialization_family(name, dumps, loads):
 register_serialization_family("error", None, serialization_error_loads)
 
 
+def check_dask_serializable(x):
+    if type(x) in (list, set, tuple) and len(x):
+        return check_dask_serializable(next(iter(x)))
+    elif type(x) is dict and len(x):
+        return check_dask_serializable(next(iter(x.items()))[1])
```
mrocklin, line 97: We should maybe test an empty sequence, like `[]` and an empty dict like `{}`

distributed/protocol/serialize.py
```diff
@@ -132,8 +146,13 @@ def serialize(x, serializers=None, on_error="message", context=None):
     if isinstance(x, Serialized):
         return x.header, x.frames
 
+    # Check for "dask"-serializable data in dict/list/set
+    supported = (
+        isinstance(x, list) and "pickle" not in serializers
+    ) or check_dask_serializable(x)
```
mrocklin, line 152: What's going on here?  I know that you added this because a particular test was failing, but what was the underlying cause?

rjzamora: Sorry - Didn't have time to look into this carefully yet.  A list is being converted to a tuple during a round trip.  I originally thought it was a msgpack limitation (since the failing test only provides the "msgpack" serializer).  However, the problem may be something else in `client.publish_dataset` or `get_dataset` (a simple msgpack dumps-loads test works fine).

rjzamora: The problem seems to be that the `use_list=False` argument is passed to `msgpack.loads` in [`msgpack_loads`](https://github.com/dask/distributed/blob/5243d23df9efb58d4f0d643c29a40062b365b65e/distributed/protocol/serialize.py#L72) - Not sure if there is a reason for this?

distributed/protocol/tests/test_serialize.py
rjzamora: Neat - Thanks for the tip

distributed/protocol/serialize.py
jakirkham: Looks like that might be intentional. Appears to have been added in PR ( https://github.com/dask/distributed/pull/2000 ). @mrocklin, how would you like us to proceed here?

rjzamora: Agreed @jakirkham - Removing that option allows tests to pass here, but may have uninteded consequences.

jakirkham: Setting `use_list=False` will let it pass with `tuple`s, but fail with `list`s. Maybe we should be tracking the collection type in some other kind of metadata (if that is important to us)?

mrocklin: > @mrocklin, how would you like us to proceed here?

Figure out what's going on (which it seems that you're doing), come up with an explanation of what's going on, make a recommendation, and then implement it?

I'm not sure if you're asking me to solve this problem for you or not.  If so, you might have to wait for a long while.  It looks like you all are making good progress though.

mrocklin: > Setting use_list=False will let it pass with tuples, but fail with lists. Maybe we should be tracking the collection type in some other kind of metadata (if that is important to us)?

Maybe msgpack serialization just isn't capable of handling type/list type ambiguities?

(please don't take anything I say here too seriously, I haven't looked into this problem at all and don't really know what I'm talking about here)

jakirkham: > Maybe msgpack serialization just isn't capable of handling type/list type ambiguities?

Yeah I think that is right.

> I'm not sure if you're asking me to solve this problem for you or not.

In some sense Rick already solved the problem, but it deviates from an earlier PR you submitted. So we are trying to understand if that deviation is ok.

rjzamora: I think the best solution is to revert back to something similar to the original check that initiated this discussion, but to add a comment about the reasoning for the check (including a reference to PR#2000).

It is clear that msgpack will treat collections as either lists or tuples - Since we are assuming tuple support throughout, it is best to force "dask" serialization of lists in the case that "msgpack" is in `serializers`.

jakirkham: It's true msgpack has issues.

```python
In [1]: import msgpack                                                          

In [2]: t = (0, 1, 2)                                                           

In [3]: l = [1, 2, 3]                                                           

In [4]: msgpack.loads(msgpack.dumps(t))                                         
Out[4]: [0, 1, 2]

In [5]: msgpack.loads(msgpack.dumps(l))                                         
Out[5]: [1, 2, 3]
```

Though Dask actually handles this correctly.

```python
In [1]: from distributed.protocol import serialize, deserialize                 

In [2]: t = (0, 1, 2)                                                           

In [3]: l = [1, 2, 3]                                                           

In [4]: deserialize(*serialize(t, serializers=["msgpack"]))                     
Out[4]: (0, 1, 2)

In [5]: deserialize(*serialize(l, serializers=["msgpack"]))                     
Out[5]: [1, 2, 3]
```

If one looks at the headers, it appears to be tracking the type information.

```python
In [6]: serialize(t, serializers=["msgpack"])                                   
Out[6]: 
({'sub-headers': [{'serializer': 'msgpack'},
   {'serializer': 'msgpack'},
   {'serializer': 'msgpack'}],
  'is-collection': True,
  'frame-lengths': [1, 1, 1],
  'type-serialized': 'tuple'},
 [b'\x00', b'\x01', b'\x02'])
```

Not sure why we are not seeing this behavior here. Is there still another issue lurking?

rjzamora: Master is currently iterating through the list if it is <= 5 elements long (and directly labeling the type as `'type-serialized'`.  The goal here was to throw out the 5-element rule for something more rigorous.  So far, the rule is that the collection contains a dask-serializable object as the first element (nesting allowed), or it is a list and msgpack is an available serializer (which we know will produce a tuple).

jakirkham: Good point! Wasn't paying attention to the size of the collections. Sounds like a bug in msgpack serialization/deserialization. Filed as issue ( https://github.com/dask/distributed/issues/3716 ).

distributed/protocol/serialize.py
```diff
@@ -132,8 +146,16 @@ def serialize(x, serializers=None, on_error="message", context=None):
     if isinstance(x, Serialized):
         return x.header, x.frames
 
+    # Check for "dask"-serializable data in dict/list/set
+    # Note: "msgpack" will always convert lists to tuples
+    #       list (see GitHub PR#2000), so we need to use
```
jakirkham, line 151: Suggest fixing a small typo and referencing the issue for further discussion.

```suggestion
    #       (see GitHub #3716), so we need to use
```

jakirkham:
Outside of that small comment, I think this is about as good for now. Hopefully issue ( https://github.com/dask/distributed/issues/3716 ) will garner some attention and we can determine how to improve things more in the future.

distributed/protocol/serialize.py
mrocklin: What does supported mean here?  It can be serialized?  If so, what if it is a list of unserializable objects?

rjzamora: Supported just means it contains a data type registered by dask_serialize.  Sorry - Would you prefer a more descriptive name?

mrocklin: But doesn't `[an_unsupported_object]` pass this check if the msgpack serializer is included?

rjzamora: Yes - You are right. supported just means “we should iterate through the collection”

rjzamora: @mrocklin - I just pushed some changes to (hopefully) make the logic a bit more clear.  For example, I am using the term `iterate_collection` instead of `supported`.   I am also now considering the order of the `serializers` list, because `"msgpack"` is only a problem for lists if it comes before `"pickle"` (I added `test_serialize_lists` to show/check this).

mrocklin: Ah, I see what's going on now.  OK, thanks @rjzamora !

jakirkham: With PR ( https://github.com/dask/distributed/pull/4575 ), I think we are able to workaround this MsgPack oddity. So dropped this code there.

PR #3690, opened 2020-04-09
Idempotent semaphore acquire with retries

fjetter:
This adds a lot of logging information to the semaphore but most importantly refactors the internal structures to make the acquire call idempotent.

This relieves us from using the Client itself as a lifetime anchor for the leases and exchanges this scheme with an explicit refresh of the leases and lease specific timeouts.

This gives the following benefits:
* I can create a unique ID for each acquire/lease which can be used for tracing using logs to debug the behaviour
* The acquire requests themselves are idempotent, i.e. upon connection failures this can be retried
* The requests are simpler since we no longer need to handle the client ID which was in the original implementation somehow redundant anyhow
* If any weird "proxy did not get ACK" issues arise (e.g. scheduler logged the lease but the OK was never received by the client) the system will eventually self heal / not deadlock
* The internal structure is subjectively more simple since we have only one dict to maintain which now holds timestamps

at the cost of additional complexity:
* The client will spawn a new periodic callback to refresh the leases. This replaces the implicit coupling to the client heartbeat. I think this is a complexity worth taking since it decouples this implementation a bit from assumptions about the lifetime management of the client.
* Another configuration option (I would argue this makes it more transparent to the users, though)

I sneaked in another change regarding the configurability of the retry options. Will open another PR for this but I needed it for the tests to succeed

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -1,6 +1,5 @@
 import uuid
 from collections import defaultdict, deque
-from functools import partial
 import asyncio
 import dask
 from asyncio import TimeoutError
@@ -9,6 +8,7 @@ from .worker import get_client
 from .metrics import time
 import warnings
 import logging
+from distributed.utils_comm import retry_operation
 
 logger = logging.getLogger(__name__)
 
@@ -21,12 +21,14 @@ class _Watch:
     def start(self):
         self.started_at = time()
 
+    def elapsed(self):
+        return time() - self.started_at
+
     def leftover(self):
         if self.duration is None:
             return None
         else:
-            elapsed = time() - self.started_at
-            return max(0, self.duration - elapsed)
+            return max(0, self.duration - self.elapsed())
 
 
 class SemaphoreExtension:
@@ -37,41 +39,49 @@ class SemaphoreExtension:
     * semaphore_acquire
     * semaphore_release
     * semaphore_create
+    * semaphore_close
+    * semaphore_refresh_leases
     """
 
     def __init__(self, scheduler):
         self.scheduler = scheduler
-        self.leases = defaultdict(deque)
+
+        # {semaphore_name: asyncio.Event}
         self.events = defaultdict(asyncio.Event)
+        # {semaphore_name: max_leases}
         self.max_leases = dict()
-        self.leases_per_client = defaultdict(partial(defaultdict, deque))
+        # {semaphore_name: {lease_id: lease_last_seen_timestamp}}
+        self.leases = defaultdict(dict)
+
         self.scheduler.handlers.update(
             {
                 "semaphore_create": self.create,
                 "semaphore_acquire": self.acquire,
                 "semaphore_release": self.release,
                 "semaphore_close": self.close,
+                "semaphore_refresh_leases": self.refresh_leases,
             }
         )
 
         self.scheduler.extensions["semaphores"] = self
-        self.pc_validate_leases = PeriodicCallback(
-            self._validate_leases,
-            1000
-            * parse_timedelta(
-                dask.config.get(
-                    "distributed.scheduler.locks.lease-validation-interval"
-                ),
-                default="s",
-            ),
+
+        validation_callback_time = 1000 * parse_timedelta(
+            dask.config.get("distributed.scheduler.locks.lease-validation-interval"),
+            default="s",
+        )
+        self._pc_lease_timeout = PeriodicCallback(
+            self._check_lease_timeout,
+            validation_callback_time,
             io_loop=self.scheduler.loop,
         )
-        self.pc_validate_leases.start()
-        self._validation_running = False
+        self._pc_lease_timeout.start()
+        self.lease_timeout = parse_timedelta(
+            dask.config.get("distributed.scheduler.locks.lease-timeout"), default="s",
+        )
 
     # `comm` here is required by the handler interface
     def create(self, comm=None, name=None, max_leases=None):
-        # We use `self.max_leases.keys()` as the point of truth to find out if a semaphore with a specific
+        # We use `self.max_leases` as the point of truth to find out if a semaphore with a specific
         # `name` has been created.
         if name not in self.max_leases:
             assert isinstance(max_leases, int), max_leases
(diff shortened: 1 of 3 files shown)


martindurant:
@fjetter , thanks for submitting this!

@quasiben @jakirkham , any chance for a review here?

quasiben:
Looks like we need the PR (or add the changes here) for the `retry_count` config option

marco-neumann-jdas:
So if I understand this correctly the lease will not periodically refreshed by the holding worker. How does this interact with:

- long running user payloads? (at which point do they NOT share the same thread anymore)
- GIL blocking?

fjetter:
> So if I understand this correctly the lease will not periodically refreshed by the holding worker. How does this interact with:
> 
> long running user payloads? (at which point do they NOT share the same thread anymore)
> GIL blocking?

We will face the same issues as with the old implementation. In the old implementation everything was coupled to the heartbeat of the client. Instead, we'll now have a dedicated semaphore heartbeat/refresh.

> long running user payloads

User payloads are always in another thread and don't impact the event loop of the worker unless...

> GIL blocking?

If the GIL is held we're out of luck and need to counteract this with longer timeouts

marco-neumann-jdas:
So if the user schedules a task that is guarded by a semaphore but is prone to GIL blocking, it might run into refresh timeouts and systematically overbook the semaphore?

fjetter:
For the retry configurations, see #3705

distributed/semaphore.py
```diff
@@ -180,15 +222,12 @@ def close(self, comm=None, name=None):
...
                 del self.events[name]
             if name in self.leases:
-                del self.leases[name]
-
-            for client, client_leases in self.leases_per_client.items():
-                if name in client_leases:
+                if self.leases[name]:
                     warnings.warn(
-                        f"Closing semaphore `{name}` but client `{client}` still has a lease open.",
+                        f"Closing semaphore {name} but still unreleased leases {sorted(self.leases[name])}",
```
lr4d, line 227:
```suggestion
                        f"Closing semaphore {name} but there remain unreleased leases: {sorted(self.leases[name])}",
```

distributed/semaphore.py
```diff
@@ -138,37 +173,44 @@ def release(self, comm=None, name=None, client=None, identifier=None):
...
         if not self._validation_running:
             self._validation_running = True
-            known_clients_with_leases = set(self.leases_per_client.keys())
-            scheduler_clients = set(self.scheduler.clients.keys())
-            for dead_client in known_clients_with_leases - scheduler_clients:
-                self._release_client(dead_client)
-            else:
-                self._validation_running = False
+            now = time()
+            semaphore_names = list(self.leases.keys())
```
lr4d, line 194: I'm curious, is there a reason (preserving determinism?) for casting `dict_keys -> list` here (and doing the same for `ids` a few lines below)?

distributed/semaphore.py
```diff
@@ -37,41 +39,45 @@ class SemaphoreExtension:
...
+    * semaphore_refresh_leases
     """
 
     def __init__(self, scheduler):
         self.scheduler = scheduler
-        self.leases = defaultdict(deque)
         self.events = defaultdict(asyncio.Event)
         self.max_leases = dict()
-        self.leases_per_client = defaultdict(partial(defaultdict, deque))
+        self.leases = defaultdict(dict)
```
lr4d, line 50:
```suggestion
        # `self.leases` is used to track each lease of each semaphore "client" with their latest "refresh" timestamp
        # Structure: {"client_name_1" : {"lease_id_1": `lease_timestamp_1`, "lease_id_2": ...},  "client_name_2" : {...}}
        self.leases = defaultdict(dict)  
```

distributed/semaphore.py
```diff
@@ -316,12 +392,13 @@ def release(self):
         """
 
         """ Release the lock if already acquired """
-        # TODO: what if connection breaks up?
+        if not self._leases:
+            raise RuntimeError("Released too often")
+        # popleft to release the oldes lease first
```
lr4d, line 397:
```suggestion
        # popleft to release the oldest lease first
```

distributed/semaphore.py
fjetter: Otherwise, this will raise a `dictionary changed size during iteration` exception when a lease is removed

distributed/semaphore.py
fjetter: FYI: The structure you're suggesting here is wrong. It is `semaphore_name.lease_id.timestamp`

fjetter:
> So if the user schedules a task that is guarded by a semaphore but is prone to GIL blocking, it might run into refresh timeouts and systematically overbook the semaphore?

Currently, the timeout would trigger a log on warning level and once the GIL is released on the worker again, this would trigger an exception on scheduler level (caught and logged as an error). The tasks/computations would not fail, however, and the semaphore would be overbooked, correct.

fjetter:
@marco-neumann-jdas I believe we can either protect ourselves from resource starvation and deadlocks (intention of this implementation) or from overbooking. I don't think we can pull off both within this library. If I'm wrong about this statement, I'd be happy to be educated :)

What I could suggest is that a lease refresh will not simply warn/log/raise the overbooking but actually registers the lease again, i.e. we'd be briefly in a state where there are more leases registered than allowed which would then block new leases until we're in a normal state.

For applications where the timeout is ill configured and every task breaches the timeout this would at least stop an unlimited avalanche.

fjetter:
@marco-neumann-jdas I added the treatment for oversubscription. If this is detected, the lease is registered and further acquisitions are blocked until we fall back to a _normal_ state. See test `test_oversubscribing_leases` (I documented it as well as I could since the logic is not exactly trivial)
This should at least protect scenarios where the GIL lock makes up only a fraction of a tasks runtime. If the entire runtime is a big GIL lock we're completely out of luck but I would argue that in these cases, the users should be able to adjust the configuration since we log a lot of warnings to point them to the proper options.

Once we equip the semaphore with some prometheus metrics this should also be clearly visible.

distributed/semaphore.py
```diff
@@ -263,25 +322,36 @@ class Semaphore:
...
+        )
+        self._refreshing_leases = False
+        pc = PeriodicCallback(
+            self._refresh_leases,
+            callback_time=1000 * refresh_leases_interval,
+            io_loop=self.client.io_loop,
+        )
+        self.refresh_callback = pc
+        # Registering here is important for proper cleanup
+        self.client._periodic_callbacks[f"refresh_semaphores_{self.id}"] = pc
```
lr4d, line 353: np: how about creating an attribute which has a value of `f"refresh_semaphores_{self.id}"` as it is used later on?

distributed/semaphore.py
```diff
@@ -138,37 +181,44 @@ def release(self, comm=None, name=None, client=None, identifier=None):
...
+                    time_since_refresh = now - self.leases[name][_id]
+                    if time_since_refresh > self.lease_timeout:
+                        logger.info(
+                            "Lease %s for %s timed out after %ss.",
+                            _id,
+                            name,
+                            time_since_refresh,
+                        )
+                        self._release_value(name=name, lease_id=_id)
+            self._validation_running = False
```
lr4d, line 221: shouldn't this be unintended once?

fjetter: No this is correct. Think of this as a poor-mans lock and the setting to False would be a release. If we'd unindent once we'd release every time the PC is called. This _lock_, however, is supposed to ensure that the body is only executed by one coroutine. While writing this I realise that the validation, by now, is entirely synchronous and this cannot happen at all. This guard can be removed.

lr4d: Sorry, I misread the indentation here, thought this was inside the loop

fjetter:
@martindurant @quasiben any feedback? If not, I'd like to merge this.

martindurant:
I'll defer to @quasiben here, if he has time

quasiben:
This looks great and I definitely appreciate the comments around tests.  @fjetter do you want to honors of hitting the green button ?

mrocklin:
Hi Folks, this introduced a test failure in master.  https://github.com/dask/distributed/issues/3717

Is there any chance that people here can help to resolve this?

PR #3692, opened 2020-04-09
Add Client.wait_to_workers to Client autosummary table

jrbourbeau:
It was appearing during the `Client` `autoclass`, but not in the `autosummary` table at the top of the API page

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -44,6 +44,7 @@ API
    Client.unpublish_dataset
    Client.upload_file
    Client.who_has
+   Client.wait_for_workers
 
 .. currentmodule:: distributed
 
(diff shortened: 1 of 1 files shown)


martindurant:
+1

docs/source/api.rst
```diff
@@ -44,6 +44,7 @@ API
    Client.unpublish_dataset
    Client.upload_file
    Client.who_has
+   Client.wait_for_workers
```
mrocklin, line 47:
```suggestion
   Client.wait_for_workers
   Client.who_has
```

jrbourbeau: Good catch : )

PR #3696, opened 2020-04-10
Add Configuration Schema

mrocklin:
This adds type and description information to the configuration using the jsonschema spec.

So far this is only a proof of concept, and touches only a couple of entries.

Our plan is to use this short term for documentation, but in the future we might also validate user configuration at runtime.  

cc @jcrist @quasiben

### DIFF
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -13,7 +13,7 @@ properties:
           allowed-failures:
             type: integer
             minimum: 0
-            description:
+            description: |
               The number of retries before a task is considered bad
 
               When a worker dies when a task is running that task is rerun elsewhere.
@@ -24,7 +24,7 @@ properties:
             type:
             - integer
             - string
-            description:
+            description: |
               The expected bandwidth between any pair of workers
 
               This is used when making scheduling decisions.
@@ -32,7 +32,7 @@ properties:
 
           blocked-handlers:
             type: array
-            description:
+            description: |
               A list of handlers to exclude
 
               The scheduler operates by receiving messages from various workers and clients
@@ -47,14 +47,14 @@ properties:
             type:
             - string
             - integer
-            description:
+            description: |
               The default size of a piece of data if we don't know anything about it.
 
               This is used by the scheduler in some scheduling decisions
 
           events-cleanup-delay:
             type: string
-            description:
+            description: |
               The amount of time to wait until workers or clients are removed from the event log
               after they have been removed from the scheduler
 
@@ -62,7 +62,7 @@ properties:
             type:
             - string
             - "null"
-            description:
+            description: |
               Shut down the scheduler after this duration if no activity has occured
 
               This can be helpful to reduce costs and stop zombie processes from roaming the earth.
@@ -70,7 +70,7 @@ properties:
           transition-log-length:
             type: integer
             minimum: 0
-            description:
+            description: |
               How long should we keep the transition log
 
               Every time a task transitions states (like "waiting", "processing", "memory", "released")
@@ -82,7 +82,7 @@ properties:
 
           work-stealing:
             type: boolean
-            description:
+            description: |
               Whether or not to balance work between workers dynamically
 
               Some times one worker has more work than we expected.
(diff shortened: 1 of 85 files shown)


mrocklin:
Also, for documentation, there are projects that seem to marry sphinx with jsonschema

mrocklin:
I tried using sphinx-jsonschema, but I'm not getting any output

mrocklin:
OK, I tried out writing more of these.  It feels decently natural after a few minutes.

If we wanted to get clever we could start doing regex testing and such.  I'm going to leave this for future work though.

At this point I'm looking for approval from some other Dask maintainers that this is a direction that we want to go.  This is something that we might apply across many of the projects.  It would be good to have buy-in.

docs/source/configuration.rst
```diff
@@ -0,0 +1,6 @@
+Configuration
+=============
+
+.. jsonschema:: https://raw.githubusercontent.com/jupyterhub/zero-to-jupyterhub-k8s/master/jupyterhub/schema.yaml
```
mrocklin, line 4: This is me just trying to get anything to print out.

distributed/distributed-schema.yaml
```diff
@@ -0,0 +1,356 @@
...
+        type: integer
+
+      scheduler:
+        type: object
+        properties:
+
+          allowed-failures:
+            type: integer
+            minimum: 0
+            description:
```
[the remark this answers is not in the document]
mrocklin: Ah, thanks.  I was wondering about this.  I remember now this from the helm days, but it's been a while and I definitely needed the reminder.

mrocklin:
> I think this definitely makes sense for documentation for the core libraries (provided we can generate nice docs from it). If it's nice we might apply it elsewhere as well, but I don't think we should make it mandatory.

+1.  Fortunately it looks like jsonschema.validate passes on missing schema entries.

Any thoughts on a convention for where to place schema files?  Currently this PR does a `foo.yaml` -> `foo-schema.yaml`

jcrist:
> Any thoughts on a convention for where to place schema files? Currently this PR does a foo.yaml -> foo-schema.yaml

If we're using it for testing and docs only, I might leave it out of the package. If you'd want to use it as part of the library (I'm not 100% sold on this yet, but could be) then adding a `schema` suffix or prefix makes sense.

mrocklin:
> If we're using it for testing and docs only, I might leave it out of the package. If you'd want to use it as part of the library (I'm not 100% sold on this yet, but could be) then adding a schema suffix or prefix makes sense.

I'm mostly talking about a naming convention for the file.  My intent was for this to be an orthogonal consideration to "how much do we depend on this?".  Although my answer to the latter is "include it in the package, use it for docs, and later on figure out how to use it for run-time error detection."  I want to establish a convention on where to place this early on though so that we don't have to rename a bunch of files later on.

foo-schema.yaml ?

jcrist:
> foo-schema.yaml

Makes sense to me.

mrocklin:
OK, the process of describing config options here is complete.

docs/source/configuration.rst
quasiben: I think we should remove the configuration page here in favor of https://github.com/dask/dask/pull/6069 then merge this in

docs/requirements.txt
```diff
@@ -7,3 +7,4 @@ dask
 sphinx
 dask_sphinx_theme
 sphinx-click
+sphinx-jsonschema
```
quasiben, line 10: we can remove this if not using jsonschema

docs/source/configuration.rst
mrocklin: Yup.  Thanks for flagging this.  I've removed all of the work around making an actual docpage and now this PR is just about the distributed-schema.yaml file, and testing its accuracy against the config file.  We don't do anything with documentation.

quasiben:
Not sure why this is failing on windows.  This passes for me locally on OSX

jrbourbeau:
It looks like the failure is related to the recently added `distributed.client.scheduler-info-interval` config option (see https://github.com/dask/distributed/pull/3723). We also added a `distributed.deploy.cluster-repr-interval` option in that PR

mrocklin:
It is because the config was recently modified in a recent PR.  In this repository we're testing to ensure completeness of the schema.  I don't think that we necessarily want to do this in all of the repositories, but I think that it makes sense to do here.

I've modified this PR to now include an informative error message pointing future developers on how to resolve syncing issues when they arise.

```python-traceback
E           ValueError: 
E           The distributed.yaml and distributed-schema.yaml files are not in sync.
E           This usually happens when we add a new configuration value,
E           but don't add the schema of that value to the distributed-schema.yaml file
E           Please modify these files to include the missing values: 
E           
E               distributed.yaml:        ['cluster-repr-interval', 'lost-worker-timeout']
E               distributed-schema.yaml: ['lost-worker-timeout']
E           
E           Examples in these files should be a good start, 
E           even if you are not familiar with the jsonschema spec
```

quasiben:
Thanks @mrocklin.  Merging in

PR #3698, opened 2020-04-11
Easier local installation

crusaderky:
This PR streamlines the CI installation script, with the main purpose of making it easier to replicate on a Linux/MacOSX dev box the exact same environment that exists on CI.

- Apply safe defaults for environment variables if the script is not invoked by travis
- Do not install miniconda if the 'conda' command is already available
- Explicitly state conda channel every time to avoid relying on ~/.condarc
- Recommend using conda in developer docs

### DIFF
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -20,7 +20,7 @@ jobs:
           miniconda-version: "latest"
           python-version: ${{ matrix.python-version }}
           environment-file: continuous_integration/environment-windows.yml
-          activate-environment: distributed
+          activate-environment: dask-distributed
           auto-activate-base: false
 
       - name: Install tornado
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -19,8 +19,8 @@ jobs:
         with:
           miniconda-version: "latest"
           python-version: ${{ matrix.python-version }}
-          environment-file: continuous_integration/environment.yml
-          activate-environment: testenv
+          environment-file: continuous_integration/environment-windows.yml
+          activate-environment: distributed
           auto-activate-base: false
 
       - name: Install tornado
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ dist: trusty
 
 env:
   matrix:
-    - PYTHON=3.6 TESTS=true COVERAGE=true PACKAGES="scikit-learn lz4" TORNADO=5 CRICK=true
-    - PYTHON=3.7 TESTS=true PACKAGES="scikit-learn python-snappy python-blosc" TORNADO=6
-    - PYTHON=3.8 TESTS=true PACKAGES="scikit-learn python-snappy python-blosc" TORNADO=6
+    - PYTHON=3.6 TESTS=true COVERAGE=true PACKAGES="lz4" TORNADO=5 CRICK=true
+    - PYTHON=3.7 TESTS=true PACKAGES="python-snappy python-blosc" TORNADO=6
+    - PYTHON=3.8 TESTS=true PACKAGES="python-snappy python-blosc" TORNADO=6
 
 matrix:
   fast_finish: true
@@ -18,7 +18,7 @@ matrix:
     python: 3.6
     env: LINT=true
   - os: osx
-    env: PYTHON=3.7 TESTS=true PACKAGES="scikit-learn python-snappy python-blosc" TORNADO=6
+    env: PYTHON=3.7 TESTS=true PACKAGES="python-snappy python-blosc" TORNADO=6
     if: type != pull_request OR commit_message =~ test-osx  # Skip on PRs unless the commit message contains "test-osx"
 
   allow_failures:
--- a/continuous_integration/environment-windows.yml
+++ b/continuous_integration/environment-windows.yml
@@ -1,4 +1,4 @@
-name: testenv
+name: distributed
 channels:
   - conda-forge
 dependencies:
--- a/continuous_integration/environment-windows.yml
+++ b/continuous_integration/environment-windows.yml
@@ -1,4 +1,4 @@
-name: distributed
+name: dask-distributed
 channels:
   - conda-forge
 dependencies:
--- a/continuous_integration/environment-windows.yml
+++ b/continuous_integration/environment-windows.yml
@@ -1,6 +1,7 @@
 name: dask-distributed
 channels:
   - conda-forge
+  - defaults
 dependencies:
   - zstandard
   - bokeh!=2.0.0
@@ -16,18 +17,18 @@ dependencies:
   - prometheus_client
   - psutil
   - pytest
+  - pytest-asyncio
+  - pytest-repeat
+  - pytest-timeout
+  - pytest-faulthandler
   - requests
+  - sortedcollections
   - toolz
   - tblib
   - zict
   - fsspec
   - pip
   - pip:
-      - pytest-repeat
-      - pytest-timeout
-      - pytest-faulthandler
-      - sortedcollections
-      - pytest-asyncio
       - git+https://github.com/dask/dask
       - git+https://github.com/joblib/joblib.git
       - git+https://github.com/dask/zict
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -5,6 +5,12 @@
 
 # Note we disable progress bars to make Travis log loading much faster
 
+# Set default variable values if unset
+# (useful when this script is not invoked by Travis)
+: ${PYTHON:=3.8}
+: ${TORNADO:=6}
+: ${PACKAGES:=python-snappy python-blosc}
+
 # Install conda
 case "$(uname -s)" in
     'Darwin')
@@ -16,18 +22,21 @@ case "$(uname -s)" in
     *)  ;;
 esac
 
-wget https://repo.continuum.io/miniconda/$MINICONDA_FILENAME -O miniconda.sh
-bash miniconda.sh -b -p $HOME/miniconda
-export PATH="$HOME/miniconda/bin:$PATH"
-conda config --set always_yes yes --set changeps1 no
-conda update -q conda
+if ! which conda; then
+  wget https://repo.continuum.io/miniconda/$MINICONDA_FILENAME -O miniconda.sh
+  bash miniconda.sh -b -p $HOME/miniconda
+  export PATH="$HOME/miniconda/bin:$PATH"
+fi
+
+conda config --set always_yes yes --set quiet yes --set changeps1 no
+conda update conda
 
 # Create conda environment
-conda create -q -n test-environment python=$PYTHON
-source activate test-environment
+conda create -n distributed -c pkgs/main python=$PYTHON
+source activate distributed
 
 # Install dependencies
-conda install -c conda-forge -q \
+conda install -c conda-forge \
     asyncssh \
     bokeh \
     click \
@@ -39,17 +48,17 @@ conda install -c conda-forge -q \
     ipywidgets \
     joblib \
     jupyter_client \
-    msgpack-python>=0.6.0 \
+    msgpack-python \
     netcdf4 \
     paramiko \
     prometheus_client \
     psutil \
-    pytest>=4 \
+    pytest \
     pytest-timeout \
-    python=$PYTHON \
     requests \
+    scikit-learn \
     scipy \
-    tblib>=1.5.0 \
+    tblib \
     toolz \
     tornado=$TORNADO \
     zstandard \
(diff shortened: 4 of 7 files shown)


.github/workflows/ci-windows.yaml
```diff
@@ -19,8 +19,8 @@ jobs:
         with:
           miniconda-version: "latest"
           python-version: ${{ matrix.python-version }}
-          environment-file: continuous_integration/environment.yml
-          activate-environment: testenv
+          environment-file: continuous_integration/environment-windows.yml
+          activate-environment: distributed
```
mrocklin, line 23: Note that we try to avoid using the name `distributed` in isolation.  It's a bit overly generic.  I recommend `dask-distributed` instead.

.travis.yml
```diff
@@ -6,9 +6,9 @@ dist: trusty
 
 env:
   matrix:
-    - PYTHON=3.6 TESTS=true COVERAGE=true PACKAGES="scikit-learn lz4" TORNADO=5 CRICK=true
-    - PYTHON=3.7 TESTS=true PACKAGES="scikit-learn python-snappy python-blosc" TORNADO=6
-    - PYTHON=3.8 TESTS=true PACKAGES="scikit-learn python-snappy python-blosc" TORNADO=6
+    - PYTHON=3.6 TESTS=true COVERAGE=true PACKAGES="lz4" TORNADO=5 CRICK=true
+    - PYTHON=3.7 TESTS=true PACKAGES="python-snappy python-blosc" TORNADO=6
+    - PYTHON=3.8 TESTS=true PACKAGES="python-snappy python-blosc" TORNADO=6
```
mrocklin, line 11: This seems quite sensible.  Thank you.

continuous_integration/environment-windows.yml
```diff
@@ -1,4 +1,4 @@
-name: testenv
+name: distributed
```
mrocklin, line 1:
```suggestion
name: dask-distributed
```

continuous_integration/travis/install.sh
```diff
@@ -39,17 +48,17 @@ conda install -c conda-forge -q \
...
     psutil \
-    pytest>=4 \
+    pytest \
     pytest-timeout \
-    python=$PYTHON \
     requests \
+    scikit-learn \
     scipy \
-    tblib>=1.5.0 \
+    tblib \
```
mrocklin, line 61: Why remove these version pinnings?  I'm not disagreeing with the action, I'm just curious.

continuous_integration/travis/install.sh
```diff
@@ -60,12 +69,11 @@ conda install -c conda-forge -q \
 if [[ $PYTHON != 3.8 ]]; then
     # For low-level profiler, install libunwind and stacktrace from conda-forge
     # For stacktrace we use --no-deps to avoid upgrade of python
-    conda install -c defaults -c conda-forge libunwind
-    conda install --no-deps -c defaults -c numba -c conda-forge stacktrace
-fi;
+    conda install -c pgks/main -c conda-forge libunwind
+    conda install --no-deps -c pgks/main -c numba -c conda-forge stacktrace
```
mrocklin, line 73: Why this change?  Also, cc @jakirkham

docs/source/develop.rst
```diff
@@ -7,31 +7,45 @@ and documentation and style standards are available at the `Dask developer
...
 
-After setting up an environment as described in the `Dask developer
-guidelines`_ you can clone this repository with git::
+Clone this repository with git::
 
    git clone git@github.com:dask/distributed.git
+   cd distributed
 
-and install it from source::
+Install all dependencies to match CI:
```
mrocklin, line 19: Some devs may not know the term CI.   Also, I don't think that they should necessarily be thinking about CI when setting up their development environment.

docs/source/develop.rst
```diff
@@ -7,31 +7,45 @@ and documentation and style standards are available at the `Dask developer
...
+On Windows:
 
-Using conda, for example::
+1. Install anaconda or miniconda
+2. ::
+
+    conda create -n distributed python=3.8 tornado=6
+    conda activate distributed
+    conda env update --file continuous_integration/environment-windows.yml
+    python -m pip install .
```
mrocklin, line 33: Why the difference between linux/mac and windows here?  Is there not also a script for windows?

docs/source/develop.rst
```diff
@@ -7,31 +7,45 @@ and documentation and style standards are available at the `Dask developer
...
+
+Alternative instructions, without using conda::
 
    git clone git@github.com:{your-fork}/distributed.git
    cd distributed
-   conda create -y -n distributed python=3.6
-   conda activate distributed
+   python -m venv $HOME/distributed
+   export PATH=$HOME/distributed/bin:$PATH
```
mrocklin, line 40: Historically we haven't been supporting venv developer workloads.  I'm not sure folks are around to debug issues here if they arise.

continuous_integration/travis/install.sh
crusaderky: They were being interpreted by bash as stdout redirection, creating files such as ``=1.5.0``. They were never received by conda to begin with.

I don't think minimum versions should be pinned in conda files because conda will already install the very latest available version anyway (save conflicts); the only exception I can think of is when you have ``channels: [pkgs/main, conda-forge]``; in such a case you may want to explicitly request a version that is in conda-forge but not in pkg/main yet.

continuous_integration/travis/install.sh
crusaderky: To my understanding 'defaults' means whatever channels are in ~/.condarc. In travis, you are always guaranteed to start without the file existing; on a dev box however you could have anything.

docs/source/develop.rst
crusaderky: As a matter of fact, there isn't. The "script" for Windows is ``.github/workflows/ci-windows.yaml``. To make it clean I should extract these lines from it and write a .BAT file, but I don't have a Windows box readily available and I'm not comfortable with DOS scripting. So I chose to be conservative here.

docs/source/develop.rst
crusaderky: I'm happy to completely remove the paragraph if you wish. That would mean removing ``dev-requirements.txt`` too.

.github/workflows/ci-windows.yaml
crusaderky, line 23:
```suggestion
          activate-environment: dask-distributed
```

continuous_integration/travis/install.sh
```diff
@@ -16,18 +22,21 @@ case "$(uname -s)" in
...
+fi
+
+conda config --set always_yes yes --set quiet yes --set changeps1 no
+conda update conda
 
 # Create conda environment
-conda create -q -n test-environment python=$PYTHON
-source activate test-environment
+conda create -n distributed -c pkgs/main python=$PYTHON
+source activate distributed
```
crusaderky, line 36:
```suggestion
conda create -n dask-distributed -c pkgs/main python=$PYTHON
source activate dask-distributed
```

docs/source/develop.rst
crusaderky, line 19:
```suggestion
Install all dependencies:
```

docs/source/develop.rst
```diff
@@ -7,31 +7,45 @@ and documentation and style standards are available at the `Dask developer
...
+    source continuous_integration/travis/install.sh
+
+On Windows:
 
-Using conda, for example::
+1. Install anaconda or miniconda
+2. ::
+
+    conda create -n distributed python=3.8 tornado=6
+    conda activate distributed
```
crusaderky, line 31:
```suggestion
    conda create -n dask-distributed python=3.8 tornado=6
    conda activate dask-distributed
```

.github/workflows/ci-windows.yaml
crusaderky: fixed

docs/source/develop.rst
crusaderky: fixed

martindurant:
@mrocklin , are you happy to keep reviewing here?

mrocklin:
I'm probably turning into a pumpkin here, at least for the next few days.

continuous_integration/travis/install.sh
gforsyth: no, `defaults` refers to a group of channels, historically `conda-main`, `conda-free` and `conda-r`. It can be explicitly overridden to mean something else in a `condarc` but without any user intervention is points at anaconda repos.

continuous_integration/travis/install.sh
```diff
@@ -60,12 +69,11 @@ conda install -c conda-forge -q \
 if [[ $PYTHON != 3.8 ]]; then
     # For low-level profiler, install libunwind and stacktrace from conda-forge
     # For stacktrace we use --no-deps to avoid upgrade of python
-    conda install -c defaults -c conda-forge libunwind
-    conda install --no-deps -c defaults -c numba -c conda-forge stacktrace
-fi;
+    conda install -c pgks/main -c conda-forge libunwind
```
gforsyth, line 72:
```suggestion
    conda install -c pkgs/main -c conda-forge libunwind
```

continuous_integration/travis/install.sh
gforsyth, line 73:
```suggestion
    conda install --no-deps -c pkgs/main -c numba -c conda-forge stacktrace
```

gforsyth:
> explicitly calling out to `pkgs/main` should be fine, but leaving it as defaults is also ok

`pkgs/main` should lead to a slightly faster solve, I think, since `defaults` will also pull down the `conda-r` repodata file

docs/source/develop.rst
crusaderky: @mrocklin I'd need you to deliberate on this one

mrocklin: You don't need me personally.  Any of the maintainers should be able to make this decision.  I think that @martindurant volunteered to be on this week.  I'd also trust @gforsyth  who has been active here.

In general, requiring my activity for something these days in a volunteer capacity is not a winning solution  :)

martindurant: Lets indeed keep with the setup most like the current one, without explicit venv stuff. That could be offered in a later PR, if anyone wants it.

continuous_integration/travis/install.sh
jakirkham: Yeah `defaults` is defined in `conda` itself.

In any even why is `pgks/main` needed?

continuous_integration/travis/install.sh
danielballan: I have anecdotally observed the conda solver failing to grab the latest version. If a minimum compatible version is known, it can be helpful to document that in this way. Would putting quotes around the full expression resolve the bash issue?

docs/source/develop.rst
crusaderky: @martindurant so should I simply remove the paragraph, but keep dev-requirements.txt there for the time being?

martindurant: Sounds reasonable to me

jakirkham:
> > explicitly calling out to `pkgs/main` should be fine, but leaving it as defaults is also ok
> 
> `pkgs/main` should lead to a slightly faster solve, I think, since `defaults` will also pull down the `conda-r` repodata file

I'm a little nervous about this. Slightly faster solve time, but deviating from what Conda's definition of `defaults` is can lead to unexpected behavior later (and likely undetectable by the average user). Any objection to just using `defaults`?

jakirkham:
Would suggest keeping `defaults` (just lower priority than `conda-forge`). This is how `conda-forge` builds, tests, and deploys as well. Most issues I see come from tinkering with channels incorrectly.

continuous_integration/travis/install.sh
crusaderky: done

crusaderky:
I'm changing everything to ``-c conda-forge -c defaults``

docs/source/develop.rst
crusaderky: done

crusaderky:
All review notes incorporated; ready for further comments or merge

jakirkham:
I may be missing something, but the change now appears to be related to how publish works. Is that expected?

crusaderky:
What the... looks like I horribly fumbled a git push --force. Recovered from backup.

mrocklin:
That test is known to be flaky.  Restarting

docs/source/develop.rst
```diff
@@ -7,30 +7,30 @@ and documentation and style standards are available at the `Dask developer
 guidelines`_ in the main documentation.
 
 .. _Dask: http://dask.org
-.. _`Dask developer guidelines`: http://docs.dask.org/en/latest/develop.html
```
mrocklin, line 10: Do we want to add this back in?  There is a broken link just above otherwise.

crusaderky: that's a link back to the very same document, it didn't make any sense to me

crusaderky: what do you mean with broken link? http://dask.org works just fine

gforsyth: Across lines 6-7 above there's a reference to 

```
`Dask developer guidelines`_
```

You are removing the target, so that link will break.

crusaderky: Ah I see it now

crusaderky:
Anything outstanding here?

mrocklin:
Thank you for the work @crusaderky and for the review @gforsyth .  Merging.

PR #3701, opened 2020-04-12
Add remote_python keyword to the new SSHCluster

### LINKED ISSUE #3700: SSHCluster fails if client and nodes have different Python paths
I tested this on my laptop and a remote server. Python installations are on different paths.
```python
    with SSHCluster(
            ["remote-node"] * 2,
            scheduler_options={"idle_timeout": "10s"},
            worker_options={"death_timeout": "10s"},
    ) as cluster:
            ....
```
output:
```
distributed.deploy.ssh - INFO - bash: /client/path/bin/python: No such file or directory
...
```

`dask-ssh` overcomes this by allowing passing `remote_python` keyword to nodes. This can be easily done in the new `SSHCluster` implementation. I will work on a new PR.

abduhbm:
Fixes #3700


distributed/deploy/ssh.py
```diff
@@ -58,6 +58,8 @@ class Worker(Process):
         The python module to run to start the worker.
     connect_options: dict
         kwargs to be passed to asyncssh connections
+    remote_python: str
+        Path to Python on remote node to run this worker.
```
mrocklin, line 62: I like that you've chosen to be consistent with the previous value in dask-ssh.  I'm curious, without that constraint do you think that there is a better name here?

distributed/deploy/ssh.py
```diff
@@ -70,6 +72,7 @@ def __init__(
         connect_options: dict,
         kwargs: dict,
         worker_module="distributed.cli.dask_worker",
+        remote_python=None,
```
mrocklin, line 75: What do you think about making the default value here `sys.executable`?

```suggestion
        remote_python=sys.executable,
```

I have mixed feelings about this. It would simplify code a little, and also help users understand the expectations for this value because it would be in the docstring. However, on the other hand it would be hard coded in the docs as wherever the python path was in the machine that generated the docs.

mrocklin:
Thanks for doing this @abduhbm

distributed/deploy/ssh.py
abduhbm: I like `remote_python` ... also `python_path` can be an alternative.

distributed/deploy/ssh.py
abduhbm: I don't like the idea that this can be hard coded in the docs. I prefer to keep it None rather than possibly confusing the reader.

martindurant:
(following the merger of your PR #3688 , you'll have to clean up the conflict here)

abduhbm:
@martindurant resolved. Thanks,

Issue #3656, opened 2020-03-28
Create MutableMapping for automatic compression


mrocklin:
In some workloads with highly compressible data we would like to trade off some computation time for more in-memory storage automatically.  Dask workers store data in a MutableMapping (the superclass of `dict`).  So in principle all we would need to do is make a MutableMapping subclass that overrides the getitem and setitem methods to compress and decompress data on demand.  

This would be an interesting task for someone who wants to help Dask, wants to learn some internals, but doesn't know a lot just yet.  I'm marking this as a good first issue.  This is an interesting and useful task that doesn't require deep incidental Dask knowledge.

Here is a conceptual prototype of such a `MutableMapping`.  This is completely untested, but maybe gives a sense of how I think about this problem.  It's probably not ideal though so I would encourage others to come up with their own design.

```python
import collections
from typing import Dict, Tuple, Callable

class TypeCompression(collections.MutableMapping):
    def __init__(
        self,
        types: Dict[type, Tuple[Callable, Callable]],
        storage=dict
    ):
        self.types = type
        self.storage = collections.defaultdict(storage)

    def __setitem__(self, key, value):
        typ = type(key)
        if typ in self.types:
            compress, decompress = self.types[typ]
            value = compress(value)
        self.storage[typ] = value

    def __getitem__(self, key):
        for typ, d in self.storage.items():
            if key in d:
                value = d[key]
                break
        else:
            raise KeyError(key)

        if typ in self.types:
            compress, decompress = self.types[typ]
            value = decompress(value)

        return value
```

This came up in https://github.com/dask/distributed/pull/3624 . cc @madsbk and @jakirkham from that PR.  cc also @eric-czech who was maybe curious about automatic compression/decompression.

People looking at compression might want to look at and use Dask's serializations and comrpession machinery in `distributed.protocol` (maybe start by looking at the `dumps`, `serialize` and `maybe_compress` functions).

mrocklin:
Also cc @prasunanand and @andersy005 who have both asked about good first issues in the past.  I think that this would be fun.

prasunanand:
@mrocklin I would love to work in it. :)

mrocklin:
Great!

prasunanand:
Hi, I need a little help.

Do I need to modify the logic in `dumps`, `loads` ( [link](https://github.com/dask/distributed/blob/2aa3ee7be1b11e4326d679d9d3c7bf906ea0623d/distributed/protocol/core.py#L18) ) ? 

[...]

PR #3702, opened 2020-04-12
WIP: Autocompression

prasunanand:
Fixes: https://github.com/dask/distributed/issues/3656

* Add TypeCompression class
* Added setitem and getitem
* Need to get rid of maybe_compress in `distributed/protocol/core.py` and `distributed/protocol/serialize.py`(currently trying to understand headers and comm)

cc: @mrocklin


mrocklin:
Thanks for starting on this @prasunanand .  I've added more context to the original issue.  I hope that that helps to clarify the objectives: https://github.com/dask/distributed/issues/3656

prasunanand:
Thanks @mrocklin :) .

prasunanand:
@mrocklin I request you to review again

distributed/protocol/compression.py
```diff
@@ -214,3 +214,35 @@ def decompress(header, frames):
...
+    def __init__(self, compressions):
+        self.storage = {}
+        self.compressions = compressions
+
+    def __setitem__(self, key, value):
+        compression_type = dask.config.get("distributed.comm.compression")
+        compression_type = "zlib"
+        if compression_type == "auto":
+            compression_type = default_compression
+        self.storage[key] = self.compressions[compression_type]["compress"](value)
```
mrocklin, line 232: I suspect that this won't work for values that don't satisfy the buffer protocol.  

Compression functions generally consume bytes-like objects and return bytes-like objects.  Your test probably passes because Numpy arrays happen to satisfy this protocol, but other objects, like custom user classes, probably wouldn't.  

I recommend trying out your solution with a wider variety of workloads.  I suspect that it will have difficulty on many of them.

You will probably need to look into how Dask serializes objects to understand how to convert arbitrary Python objects into bytes-like objects.

prasunanand:
Thanks @mrocklin. I have a doubt. Does Serialization and compression mean the same for `large results or Python specific data, like NumPy arrays or Pandas. If yes, Should I call serialize() function on Python specific datasets which looks for different serializers registered ? 

```
if type(key) is in [np.ndarray, pd.Dataframe]:
    compress_func = serialize
    decompress_func = serialize
```

Sorry for replying late .

prasunanand:
Looking at the code in [`dumps`](https://github.com/dask/distributed/blob/master/distributed/protocol/core.py#L18) , are both serialize() and compress() called on the data ?

TomAugspurger:
@prasunanand I would have a look at https://github.com/dask/distributed/blob/6db09f32d8ca58a56a50385a49150ef16a5d51b0/distributed/worker.py#L233-L234.

I think the idea is to have `Worker.data` be the mutable mapping that implements automatic compression.

Compressing data prior to communication (in `distributed/protocol`) is also an interesting idea, but might be more involved.

distributed/protocol/__init__.py
```diff
@@ -117,3 +117,34 @@ def _register_cudf():
...
+
+
+class TypeCompressor(collections.abc.MutableMapping):
+    def __init__(self, compressions):
+        self.storage = {}
+        self.compressions = compressions
+
+    def __setitem__(self, key, value):
+        header, frames = serialize(value)
+        self.storage[key] = {"header": header, "data": maybe_compress(frames)}
```
TomAugspurger, line 134: Can this just store a tuple of `(header, data)`? Seems more natural than a dict with always 2 keys.

distributed/protocol/__init__.py
```diff
@@ -117,3 +117,34 @@ def _register_cudf():
...
+
+    def __getitem__(self, key):
+        header = self.storage[key]["header"]
+        compression, compressed = self.storage[key]["data"]
```
TomAugspurger, line 138: Then here we can do `header, (compression, compressed) = self.storage[key]`.

distributed/protocol/__init__.py
```diff
@@ -117,3 +117,34 @@ def _register_cudf():
...
+
+        frames = decompress({"compression": {compression}}, compressed)
+        return deserialize(header, frames)
+
+    def __delitem__(self, key):
+        return self.storage.__delitem__(key)
```
TomAugspurger, line 144:
```suggestion
        return del self.storage[key]
```

mrocklin:
> Thanks @mrocklin. I have a doubt. Does Serialization and compression mean the same for `large results or Python specific data, like NumPy arrays or Pandas. If yes, Should I call serialize() function on Python specific datasets which looks for different serializers registered ?

```
if type(key)  == np.ndarray or type(key)  ==  pd.Dataframe:
    compress_func = serialize
    decompress_func = serialize
```

It should work on any python object that is passed in.  To do this well you will need to understand more about how dask handles serialization and compression.

In general, I get the sense that this issue may be beyond your current understanding of Dask.  I was probably wrong to mark this as a good first issue.

prasunanand:
Apologies @mrocklin . I studied `zict` codebase and `worker.py`  as I was unable to get my head around what this feature demanded. Now I have a better understanding. Hence, I have tried once more and hope it solves the issue(or the solution is close).

If not I will close this PR and work on other issues.

distributed/worker.py
```diff
@@ -535,7 +541,11 @@ def __init__(
                 "distributed.worker.memory.pause"
             )
 
-        if isinstance(data, MutableMapping):
+        if isinstance(data, MutableMapping) and not isinstance(data, TypeCompressor):
+            self.data = TypeCompressor()
+            for k, v in data.values():
+                self.data[k] = data[k]
+        elif isinstance(data, TypeCompressor):
```
mrocklin, line 548: I'm not sure what's going on here.  Was it necessary to special-case `TypeCompressor` here?  Ideally we only depend on the MutableMapping interface and don't have to worry too much about every possible type that someone might provide individually.  That can make future maintenance an issue.

distributed/protocol/__init__.py
```diff
@@ -117,3 +117,47 @@ def _register_cudf():
...
+
+    def __setitem__(self, key, value):
+        header, frames = serialize(value)
+        self.storage[key] = (header, maybe_compress(frames))
+
+    def __getitem__(self, key):
+        header, (compression, compressed) = self.storage[key]
+
+        frames = decompress({"compression": {compression}}, compressed)
+        return deserialize(header, frames)
```
mrocklin, line 143: In principle this looks good @prasunanand .  Thank you for sticking with this.

The next thing to do is probably to try it out in practice a bit and see how it performs.  There are optimizations I can think of, like it might make sense to only compress values that we know are small using the `sizeof` function, but that would just be a guess.  If you have time to try out a few different kinds of computations, maybe taken from examples.dask.org that might be a good way to get more information here.

martindurant: @prasunanand , have you had any time to try the suggestions here?

jakirkham: I thought this is what `maybe_compress` already does. Meaning it checks to see whether something is worth compressing. If so, it compresses it.

jakirkham:
cc @madsbk (as this may be of interest 🙂)

distributed/protocol/__init__.py
```diff
@@ -117,3 +117,34 @@ def _register_cudf():
...
     with ignoring(ImportError):
         from cuml.comm import serialize
+
+
+import collections.abc
+
+import dask
+
+
+class TypeCompressor(collections.abc.MutableMapping):
```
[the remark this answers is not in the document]
jakirkham: It's probably tricky to do that as this depends on `maybe_compress`, which is in Distributed. That said, maybe there are ways to build this out of functionality in Zict. Thus making this more compact (maybe not even needing this class).

jakirkham:
Any thoughts or questions on the feedback so far @prasunanand ? 🙂

prasunanand:
Sorry I have been busy at work. Last week I have been trying to measure the performance using different workloads. Found out that Numpy array were not getting deserialized properly. Regarding measung performance for different workloads I have been trying to use dask diagnostics.


In the following code, I am trying to compute over an ndarray of size (1000, 1000).
```
import asyncio

async def test_compression():
	import numpy as np
	import dask.array as da
	from distributed.protocol import TypeCompressor
	from distributed import Scheduler, Worker, Client, wait, performance_report

	x = np.ones(1000000)  # a large but compressible piece of data

	d = TypeCompressor()

	d["x"] = x  # put Python object into d
	out = d["x"]  # get the object back out
	d["z"] = 3

	assert str(out) == str(x)
	np.testing.assert_allclose(
	    np.frombuffer(out, x.dtype), x
	)  # the two objects should match

	# assuming here that the underlying bytes are
	# stored in something like a `.storage` attribute, but this isn't required
	# we check that the amount of actual data stored is small
	assert sum(map(len, d.storage.values())) < x.nbytes

	async with Scheduler() as s:
	    async with Worker(s.address, data=d) as worker:
	        async with Client(s.address, asynchronous=True) as c:
	        	async with performance_report(filename="dask-report.html"): #measure performance
	        		x = da.ones((1000, 1000))
	        		y = await x.persist()  # put data in memory
	        		y = await (x + x.T).mean().persist()  # do some work
	        		future = c.compute(y)
	        		await wait(future)
	        		assert sum(map(len, worker.data.storage.values())) < x.nbytes

asyncio.run(test_compression())

```

Attached is an html file that I got as result.(To view change the extension from .txt to .html)
[dask-report.txt](https://github.com/dask/distributed/files/4920356/dask-report.txt)

prasunanand:
Apart from this:
What are the other data-types I should investigate with (numpy.ndarray, string, pandas.Series) ?

jakirkham:
I tried another approach to this problem (hope that is ok 🙂) in PR ( https://github.com/dask/distributed/pull/3968 ). We need not go with that. Just trying to see if we are able to get the behavior we want out of Zict alone. Would be curious to hear what people think of it.

PR #3704, opened 2020-04-13
Allow range of ports to be specified for Workers

jrbourbeau:
This PR allows multiple ports to be specified for `Worker` and `Nanny` processes created with the `dask-worker` command. Here we allows users to specify the starting and stopping points for a sequential range of available ports like so `<first-port>:<last-port>`. So, for example, `--worker-ports 9500:9600` would use ports `9500`, `9501`, ..., `9599`, `9600`. 

```
dask-worker <scheduler-address> --nprocs 10 --worker-port 9500:9600 --nanny-port 9700:9800
```

This can be useful when using systems where a specific range of ports have been whitelisted for use. There was some discussion over in https://github.com/dask/distributed/issues/899 where there seemed to be interest in adding this capability 

Fixes https://github.com/dask/distributed/issues/899

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -1,4 +1,5 @@
 import asyncio
+import errno
 import logging
 from multiprocessing.queues import Empty
 import os
@@ -33,7 +34,7 @@ from .utils import (
     ignoring,
     TimeoutError,
 )
-from .worker import run, parse_memory_limit, Worker
+from .worker import run, parse_memory_limit, Worker, parse_worker_ports
 
 logger = logging.getLogger(__name__)
 
@@ -91,7 +92,7 @@ class Nanny(ServerNode):
         port=None,
         protocol=None,
         config=None,
-        **worker_kwargs
+        **worker_kwargs,
     ):
         self._setup_logging(logger)
         self.loop = loop or IOLoop.current()
@@ -212,13 +213,10 @@ class Nanny(ServerNode):
         ):
             host = get_ip(get_address_host(self.scheduler.address))
 
-        self._start_address = address_from_user_args(
-            host=host,
-            port=port,
-            interface=interface,
-            protocol=protocol,
-            security=security,
-        )
+        self._start_port = port
+        self._start_host = host
+        self._interface = interface
+        self._protocol = protocol
 
         self._listen_address = listen_address
         Nanny._instances.add(self)
@@ -259,9 +257,33 @@ class Nanny(ServerNode):
 
         await super().start()
 
-        await self.listen(
-            self._start_address, **self.security.get_listen_args("worker")
-        )
+        ports = parse_worker_ports(self._start_port)
+        for port in ports:
+            start_address = address_from_user_args(
+                host=self._start_host,
+                port=port,
+                interface=self._interface,
+                protocol=self._protocol,
+                security=self.security,
+            )
+            try:
+                await self.listen(
+                    start_address, **self.security.get_listen_args("worker")
+                )
+            except OSError as e:
+                if len(ports) > 1 and e.errno == errno.EADDRINUSE:
+                    continue
+                else:
+                    raise e
+            else:
+                self._start_address = start_address
+                break
+        else:
+            raise ValueError(
+                f"Could not start nanny on host {self._start_host}"
+                f"with port {self._start_port}"
+            )
+
         self.ip = get_address_host(self.address)
 
         await preloading.on_start(
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -12,7 +12,7 @@ import weakref
 
 import dask
 from dask.system import CPU_COUNT
-from tornado.ioloop import IOLoop
+from tornado.ioloop import IOLoop, PeriodicCallback
 from tornado import gen
 
 from .comm import get_address_host, unparse_host_port
@@ -29,7 +29,6 @@ from .utils import (
     mp_context,
     silence_logging,
     json_load_robust,
-    PeriodicCallback,
     parse_timedelta,
     parse_ports,
     ignoring,
@@ -204,7 +203,7 @@ class Nanny(ServerNode):
         self.scheduler = self.rpc(self.scheduler_addr)
 
         if self.memory_limit:
-            pc = PeriodicCallback(self.memory_monitor, 100, io_loop=self.loop)
+            pc = PeriodicCallback(self.memory_monitor, 100)
             self.periodic_callbacks["memory"] = pc
 
         if (
@@ -786,4 +785,6 @@ class WorkerProcess:
             # Loop was stopped before wait_until_closed() returned, ignore
             pass
         except KeyboardInterrupt:
-            pass
+            # At this point the loop is not running thus we have to run
+            # do_stop() explicitly.
+            loop.run_sync(do_stop)
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -256,15 +256,15 @@ class Nanny(ServerNode):
 
         await super().start()
 
-        self._start_address = addresses_from_user_args(
+        _start_addresses = addresses_from_user_args(
             host=self._start_host,
             port=parse_worker_ports(self._start_port),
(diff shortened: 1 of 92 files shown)


mrocklin:
Thanks for this @jrbourbeau .  It's not every day that you're able to fix an issue below `#1000` :)

I wonder if it's possible to make this change lower down, in the `Worker` class.  About a year ago there was a push to remove logic from the `dask-foo` commands and push it into the classes themselves.  This makes it easier for the other Dask subprojects to reuse this logic, even if they don't use the same `dask-foo` commands (I started this for dask-cuda, but I think that it ended up being useful for dask-gateway, and the more recent dask-ssh command).

jrbourbeau:
>  It's not every day that you're able to fix an issue below #1000

An oldie, but a goody : )

> This makes it easier for the other Dask subprojects to reuse this logic

Yeah, that's a great point. We could push this down into the `Worker` class so `Worker(..., port="9500:9600")` would automatically select a port between 9500 and 9600 (inclusive). Since this isn't specifying an exact port, but rather a range of possible ports, we'll need to check that a given port is available and, if not, then try another port in the allowed range

jrbourbeau:
I've moved the port range logic down into the `Worker` and `Nanny` classes, so we can now specify a range of ports both at the command line

```
dask-worker <scheduler-address> --nprocs 10 --worker-port 9500:9600
```

and using the Python API

```python
w = await Worker(..., port="7000:8000")
```

The CI test failures are from a known flaky test (`test_dont_steal_unknown_functions`) and `test_pid_file` which I noticed popped up in `master` so I believe is unrelated (xref https://github.com/dask/distributed/issues/3734 for tracking)

distributed/worker.py
```diff
@@ -1028,9 +1013,52 @@ async def start(self):
         enable_gc_diagnosis()
         thread_state.on_event_loop_thread = True
 
-        await self.listen(
-            self._start_address, **self.security.get_listen_args("worker")
+        ports = parse_worker_ports(self._start_port)
```
TomAugspurger, line 1016: Slightly concerned with the amount of duplication between here and the Nanny. What are your thoughts?

distributed/nanny.py
```diff
@@ -259,9 +257,33 @@ async def start(self):
 
         await super().start()
 
-        await self.listen(
-            self._start_address, **self.security.get_listen_args("worker")
-        )
+        ports = parse_worker_ports(self._start_port)
+        for port in ports:
```
TomAugspurger, line 261: Just for my understanding, does each nanny attempt to connect to the first port? One succeeds and the rest try the second? And so on?

distributed/worker.py
jrbourbeau: I spent some time trying to push this logic down into the underlying `Server.listen` call, but things got somewhat complicated so I decided to punt on it for now. I'll revisit this to see if we can do it in a simple way

distributed/nanny.py
jrbourbeau: Yes, that's exactly what happens

distributed/worker.py
TomAugspurger: I'm completely fine with things as is if it's not straightforward to dedpulicate. I didn't look too closely.

jsignell:
It looks like the failing tests are known flakey tests.

distributed/worker.py
```diff
@@ -3182,6 +3193,25 @@ def parse_memory_limit(memory_limit, nthreads, total_cores=CPU_COUNT):
...
+        port_start, port_stop = map(int, port.split(":"))
+        if port_stop <= port_start:
+            raise ValueError(
+                "When specifying a range of ports like port_start:port_stop, "
+                "port_stop must be greater than port_start, but got "
+                f"port_start={port_start} and port_stop={port_stop}"
+            )
+        port = list(range(port_start, port_stop + 1))
+
+    return port
```
mrocklin, line 3212: This could be used for nannies schedulers or other servers I think.  Maybe it should move to utils and we should remove `worker` from the name.  Also, can I ask for a small docstring, maybe with a couple of examples?

Should it always return a list of integers for consistency, or would that complicate other parts of the codebase?

distributed/nanny.py
```diff
@@ -259,9 +256,17 @@ async def start(self):
 
         await super().start()
 
+        self._start_address = addresses_from_user_args(
+            host=self._start_host,
+            port=parse_worker_ports(self._start_port),
```
mrocklin, line 261: I'm not sure that this is what we want.  If I provide a range like "8000:9000" then I think that this will provide a list of 1000 addresses on which to listen.  Is that correct?  If so, that could quickly become problematic. 

Instead, I would expect the port logic to occur around when we try to listen on a port, and find it to be busy.

jrbourbeau: It will provide a list of 1000 possible addresses to listen on but when we then make the call to `self.listen` it will start to loop through the 1000 addresses until we find one that is open, create a single listener for that address, and then return. So each `listen` call results in, at most, one listener being created. There's an added test, `distributed/tests/test_core.py::test_multiple_addresses`, which demonstrates this behavior. 

I agree if we created 1000 listeners here that would be problematic. Do you think the behavior outlined above is still problematic?

distributed/worker.py
jrbourbeau: +1 to moving to utils and adding documentation

Right now this function either returns a list of integers or the 1-element list `[None]`. I decided to keep the `None` option as to not break existing code paths where if `None` is received as a port, then some configured [default port is used](https://github.com/dask/distributed/blob/ecdcb33720b2794c21b5bd322e4d2ccd8fec2787/distributed/core.py#L301-L302)

distributed/nanny.py
mrocklin: OK, I think that there might be a slight conflict here that may be my fault.  

In https://github.com/dask/distributed/pull/3288 I added the ability for the Scheduler to listen on many addresses at once.  This is where the `addresses_from_user_args` function came from.  However I never took the time to extend this functionality over to worker/nanny.  The intention here was that we would iterate over many addresses from `self._start_address` and listen on all of them.

In this PR we're now reusing that same mechanism, but for a different purpose saying "listen on the first one that works".  Everything today is fine, but the moment that we try to either have multiple networks active on the workers, or use a port range on the scheduler, we'll run into a biggish conflict I think.

jrbourbeau: Alright, thanks for the additional context. We would indeed have a conflict if we wanted to use, say, a port range and multiple interfaces on a `Worker`. 

I'll roll back the recent changes that moved the "listen on the first address that works" logic to `Server.listen` and instead keep an explicit `for`-loop over multiple ports in the `Worker`/`Nanny` classes like in https://github.com/dask/distributed/pull/3704/commits/2d7162e305bfc531c23f07391ba2c7e14efaaa4e. I think having a single input address for `Server.listen` will help us better isolate any port range and/or multiple interface logic moving forward.

distributed/worker.py
```diff
@@ -1017,9 +1003,52 @@ async def start(self):
...
+                else:
+                    raise e
+            else:
+                self._start_address = start_address
+                break
+        else:
+            raise ValueError(
+                f"Could not start Worker on host {self._start_host}"
+                f"with port {self._start_port}"
+            )
```
mrocklin, line 1031: Does it make sense to try to move this logic into `Server.listen`?  It would mean changing the interface of that method to something like the following I think.

```python
async def listen(host, port, interface, protocol, security, ...)
```

Then maybe we can pull `address_from_user_args in, and apply this logic to the Scheduler and other future systems as well?

mrocklin: (sorry to be late in reviewing by the way, if you want to just move forward and merge this that's fine with me too)

jrbourbeau: For now I'd prefer to have this merged in as is, and then address the move to `Server.listen` in a follow-up PR

mrocklin: +1 from me

PR #3705, opened 2020-04-14
Allow modification of distributed.comm.retry at runtime

fjetter:
This allows the modification of the retry options at runtime such that the ordinary `dask.config` context manager can be used. Makes tests much more intuitive and also makes sense to be configurable to give users more control about this behaviour.

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -7,7 +7,6 @@ import operator
 import re
 import sys
 from time import sleep
-from unittest import mock
 import logging
 
 import dask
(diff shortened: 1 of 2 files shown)


martindurant:
Got `OSError: [Errno 98] Address already in use` on test_nanny for py35. I suggest that *might* be related.

fjetter:
I'll try to reproduce it locally. In the meantime I re-triggered the run in case it is unrelated / flaky.

fjetter:
couldn't reproduce locally, had issues setting up py3.5 :/

@martindurant seems like a flaky test and I assume it is unrelated since the retries do not start/restart or affect the nanny in any way.

martindurant:
OK, the failure went away, so I'm happy :)

Issue #3574, opened 2020-03-13
test_dont_steal_unknown_functions failure


jrbourbeau:
xref https://travis-ci.org/github/dask/distributed/jobs/662107052

<details>
<summary>Full traceback:</summary>

```python
______________________ test_dont_steal_unknown_functions _______________________

    def test_func():

        result = None

        workers = []

        with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:

    

            async def coro():

                with dask.config.set(config):

                    s = False

                    for i in range(5):

                        try:

                            s, ws = await start_cluster(

                                nthreads,

                                scheduler,

                                loop,

                                security=security,

                                Worker=Worker,

                                scheduler_kwargs=scheduler_kwargs,

                                worker_kwargs=worker_kwargs,

                            )

                        except Exception as e:

                            logger.error(

                                "Failed to start gen_cluster, retrying",

                                exc_info=True,

                            )

                        else:

                            workers[:] = ws

                            args = [s] + workers

                            break

                    if s is False:

                        raise Exception("Could not start cluster")

                    if client:

                        c = await Client(

                            s.address,

                            loop=loop,

                            security=security,

                            asynchronous=True,

                            **client_kwargs

                        )

                        args = [c] + args

                    try:

                        future = func(*args)

                        if timeout:

                            future = asyncio.wait_for(future, timeout)

                        result = await future

                        if s.validate:

                            s.validate_state()

                    finally:

                        if client and c.status not in ("closing", "closed"):

                            await c._close(fast=s.status == "closed")

                        await end_cluster(s, workers)

                        await asyncio.wait_for(cleanup_global_workers(), 1)

    

                    try:

                        c = await default_client()

                    except ValueError:

                        pass

                    else:

                        await c._close(fast=True)

    

                    for i in range(5):

                        if all(c.closed() for c in Comm._instances):

                            break

                        else:

                            await asyncio.sleep(0.05)

                    else:

                        L = [c for c in Comm._instances if not c.closed()]

                        Comm._instances.clear()

                        # raise ValueError("Unclosed Comms", L)

                        print("Unclosed Comms", L)

    

                    return result

    

            result = loop.run_sync(

>               coro, timeout=timeout * 2 if timeout else timeout

            )

distributed/utils_test.py:957: 

[...]

PR #3706, opened 2020-04-15
async... await in tests

crusaderky:
- Update all tests to replace @gen.coroutine... yield with async... await
- Replace all gen.sleep with asyncio.sleep
- Clean up ``@skip`` markers for Python < 3.6
- Fix bug in ``Client(asynchronous=True).close()`` that, if invoked twice, on the second round would return ``None``; it now returns ``Awaitable[None]``.
- Decorators ``@gen_cluster`` and ``@gen_test`` now accept only async functions
- ``@gen_cluster`` now properly waits for all Comm and Client objects to be closed, and crashes in case some keep lingering
- Fix flaky tests

### DIFF
--- a/distributed/dashboard/tests/test_worker_bokeh.py
+++ b/distributed/dashboard/tests/test_worker_bokeh.py
@@ -28,20 +28,20 @@ from distributed.dashboard.components.worker import (
     worker_kwargs={"dashboard": True},
     scheduler_kwargs={"dashboard": True},
 )
-def test_routes(c, s, a, b):
+async def test_routes(c, s, a, b):
     port = a.http_server.port
 
     future = c.submit(sleep, 1)
-    yield gen.sleep(0.1)
+    await gen.sleep(0.1)
 
     http_client = AsyncHTTPClient()
     for suffix in ["status", "counters", "system", "profile", "profile-server"]:
-        response = yield http_client.fetch("http://localhost:%d/%s" % (port, suffix))
+        response = await http_client.fetch("http://localhost:%d/%s" % (port, suffix))
         body = response.body.decode()
         assert "bokeh" in body.lower()
         assert not re.search("href=./", body)  # no absolute links
 
-    response = yield http_client.fetch(
+    response = await http_client.fetch(
         "http://localhost:%d/info/main/workers.html" % s.http_server.port
     )
 
@@ -49,16 +49,16 @@ def test_routes(c, s, a, b):
 
 
 @gen_cluster(client=True, worker_kwargs={"dashboard": True})
-def test_simple(c, s, a, b):
+async def test_simple(c, s, a, b):
     assert s.workers[a.address].services == {"dashboard": a.http_server.port}
     assert s.workers[b.address].services == {"dashboard": b.http_server.port}
 
     future = c.submit(sleep, 1)
-    yield gen.sleep(0.1)
+    await gen.sleep(0.1)
 
     http_client = AsyncHTTPClient()
     for suffix in ["crossfilter", "system"]:
-        response = yield http_client.fetch(
+        response = await http_client.fetch(
             "http://localhost:%d/%s" % (a.http_server.port, suffix)
         )
         assert "bokeh" in response.body.decode().lower()
@@ -67,12 +67,12 @@ def test_simple(c, s, a, b):
 @gen_cluster(
     client=True, worker_kwargs={"dashboard": True},
 )
-def test_services_kwargs(c, s, a, b):
+async def test_services_kwargs(c, s, a, b):
     assert s.workers[a.address].services == {"dashboard": a.http_server.port}
 
 
 @gen_cluster(client=True)
-def test_basic(c, s, a, b):
+async def test_basic(c, s, a, b):
     for component in [
         StateTable,
         ExecutingTimeSeries,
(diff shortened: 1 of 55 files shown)


mrocklin:
Oh wow, this is amazing.  Thank you @crusaderky !

distributed/dashboard/tests/test_worker_bokeh.py
```diff
@@ -28,37 +28,37 @@
     worker_kwargs={"dashboard": True},
     scheduler_kwargs={"dashboard": True},
 )
-def test_routes(c, s, a, b):
+async def test_routes(c, s, a, b):
     port = a.http_server.port
 
     future = c.submit(sleep, 1)
-    yield gen.sleep(0.1)
+    await gen.sleep(0.1)
```
quasiben, line 35: can gen.sleep also be replaced with asyncio.sleep ?

crusaderky: sure, why not

crusaderky: done!

crusaderky:
I'm done, however travis is randomly failing. The tests are concentrated (but not exclusive to) Python 3.6, and are completely random - run again, and they just move somewhere else. I can't reproduce them locally.
Help is welcome.

For reviewers: since reading 2800 lines of changes can make your eyes cross, I presented the change in 3 commits:
1. everything else. The code after this change is not valid but it's where all manual interventions are.
2. replace all yield with await when used in coroutines; change 'def' to 'async def'
3. replace gen.sleep with asyncio.sleep

mrocklin:
Thanks again for doing this @crusaderky .  This is great.  

> For reviewers: since reading 2800 lines of changes can make your eyes cross, I presented the change in 3 commits:

That's super thoughtful.  Thanks!

> The failures are concentrated in (but not exclusive to) Python 3.6, and are completely random - run again, and they just move somewhere else. I can't reproduce them locally.
> Help is welcome.

Hrm, that looks frustrating.  At first glance I don't know what's going on.  Those do appear to be new intermittent failures though, so I suspect that they are related to some change here.  There are some subtle differences between asyncio and tornado that can come up very rarely, like `asyncio.sleep(0)` behaving differently than `gen.sleep(0)` or similar topics.

One rather unpleasant way to figure out what is going on is to do a binary search with CI.  For example you could commit the changes in half of the files and see if CI keeps complaining.  That is a really painful iteration cycle though.

crusaderky:
Ok I brought it down to two flaky tests.

- test_oversubscribing_leases randomly fails on all environment. I did not touch this test at all.
- test_dont_steal_unknown_functions is deterministically broken on Python 3.6 Linux *only*, and randomly (rarely) fails in other environments. Note how I changed the test to be less susceptible to randomness.

Not sure what to do with either but I'm fairly convinced it's not my fault...

mrocklin:
See https://github.com/dask/distributed/issues/3717 for one

`test_dont_steal_unknown_functions` seems new to me, but I'm not entirely sure how it could be related.  My recommendation here would be to raise an empty PR to test CI and see if it's failing on master.

jrbourbeau:
`test_dont_steal_unknown_functions` is a known flaky test (xref https://github.com/dask/distributed/issues/3574)

mrocklin:
Well great.  In that case I'm +1 here.  Any objection to merging @jrbourbeau ?

crusaderky:
should I mark them as xfail?

mrocklin:
Merging this in.  Thanks again @crusaderky !

chinmaychandak:
A bunch of `@gen_cluster` tests in [streamz](https://github.com/python-streamz/streamz) are also failing. 

An example test is the one below (link [here):](https://github.com/python-streamz/streamz/blob/994b7308387ae6f9113adc55dbcc9cf558d85327/streamz/tests/test_dask.py#L17)

```
@gen_cluster(client=True)
def test_map(c, s, a, b):
    source = Stream(asynchronous=True)
    futures = scatter(source).map(inc)
    futures_L = futures.sink_to_list()
    L = futures.gather().sink_to_list()

    for i in range(5):
        yield source.emit(i)

    assert L == [1, 2, 3, 4, 5]
    assert all(isinstance(f, Future) for f in futures_L)
```

Just from the looks of it, and by reading the comments here, upgrading from Python 3.6 and adding an `async def` for test with `@gen_cluster` should work. 

But, even after making these changes, all `@gen_cluster` tests are still failing on travisCI with the error: `ValueError: @gen_cluster should wrap async def functions`. 

Also note that I had to add `pytest-tornasync` to travis since pytest alone does not handle async tests.

Any help is appreciated since this is blocking another PR as well.

mrocklin:
@crusaderky do you have time to revert the check that causes these failures downstream?

crusaderky:
@mrocklin yes I will

crusaderky:
> But, even after making these changes, all @gen_cluster tests are still failing on travisCI with the error: ValueError: @gen_cluster should wrap async def functions.

@chinmaychandak could you post the output of ``sys.version`` and ``platform.platform()`` when running on travis?

chinmaychandak:
> @chinmaychandak could you post the output of `sys.version` and `platform.platform()` when running on travis?

Does this help — https://travis-ci.org/github/python-streamz/streamz/builds/677897003? It show a Linux platform with Python 3.8.

PR #3709, opened 2020-04-15
Add remote_python option in ssh cmd

### LINKED ISSUE #3707: remote_path is ignored in SSHCluster
Current State:

In the doc: https://docs.dask.org/en/latest/setup/ssh.html#cmdoption-dask-ssh-remote-python
It is not specified that one can use `remote_python` in SSHCluster, although both `worker_options` and `scheduler_options` support it:
- https://github.com/dask/distributed/blob/master/distributed/deploy/ssh.py#L87
- https://github.com/dask/distributed/blob/master/distributed/deploy/ssh.py#L163

Furthermore, these parameters are ignored when passed, dask always using `sys.executable` in both remote scheduler and workers.

Expected behavior:
I would like to pass `remote_python` to both `worker_options` and `scheduler_options`.

Implementation details:

- https://github.com/dask/distributed/blob/master/distributed/deploy/ssh.py#L113
- https://github.com/dask/distributed/blob/master/distributed/deploy/ssh.py#L189

Both calls should replaced with `self.remote_python or sys.executable`, as it is in [old_ssh.py](https://github.com/dask/distributed/blob/master/distributed/deploy/old_ssh.py#L295)

abduhbm:
Fixes #3707

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -110,7 +110,7 @@ class Worker(Process):
         cmd = " ".join(
             [
                 set_env,
-                sys.executable,
+                self.remote_python or sys.executable,
                 "-m",
                 self.worker_module,
                 self.scheduler,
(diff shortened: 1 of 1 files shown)


abduhbm:
Referencing #3701 

CI 3.7 failure `test_dont_steal_unknown_functions` is presumably unrelated.

martindurant:
Thanks, @abduhbm

PR #3710, opened 2020-04-15
Use `bytearray` for trivial frames too

jakirkham:
Related to issue ( https://github.com/dask/dask/issues/6071 )

PR #3711, opened 2020-04-15
Always use `readinto` in TCP

jakirkham:
This code path existed before Tornado 5 to provide a way to workaround the absence of `readinto`. However as we require Tornado 5 at a minimum, there is no need to handle the case where `readinto` is not defined.

Further there is no need to special case empty frames with this change. So incorporate the change from PR ( https://github.com/dask/distributed/pull/3710 ).

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -13,7 +13,7 @@ except ImportError:
 
 import dask
 from tornado import netutil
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @jrbourbeau

jakirkham:
Thanks for catching that! 😄

That's right.

jakirkham:
Thanks James! 😄

PR #3715, opened 2020-04-16
Default threads_per_worker value causes test failure on large CPUs

crusaderky:
On my Linux dev box mounting a 16 cores / 32 threads CPU, ``LocalCluster(0)`` defaults to threads_per_worker=2, while on Travis it defaults to 1. This caused a test to fail on my box.


mrocklin:
Seems fine to me.  Thanks for the fix @crusaderky !

Issue #3665, opened 2020-04-01
Configurable polling interval for cluster widget 


jsignell:
At the moment, if you render a cluster in the notebook using the widget mechanism, it polls the backend ever 500ms for changes in `scheduler_info`:

https://github.com/dask/distributed/blob/2129b740c1e3f524e5ba40a0b6a77b239d4c1f94/distributed/deploy/cluster.py#L322-L324

It seems like it'd be pretty easy to make this configurable. Perhaps `_widget_callback_interval` (I am open to other names :upside_down_face: ) could be an attr on the Cluster class?

jrbourbeau:
Making this configurable seems fine. These types of values have typically been added as config options in `distributed/distributed.yaml`. Out of curiosity, has the 500ms update time been an issue?

jsignell:
It hasn't been an issue yet, but I am writing a Cluster that does a GET for `scheduler_info` so I am a little worried about doing a done of requests.

jrbourbeau:
Gotcha, thanks for the additional context. How about adding a `distributed.deploy.cluster-repr-update` (other names welcome) config option? It would then be read in when creating the `cluster-repr` `PeriodicCallback` like so:

https://github.com/dask/distributed/blob/7c9da106b2edd543d3b420ad8270261c6a04ec5e/distributed/deploy/spec.py#L354-L356

PR #3723, opened 2020-04-17
Configurable polling interval for cluster widget

jsignell:
I did the scheduler_info for the client as well since it is a very similar situation.

Closes: #3665

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -677,9 +677,13 @@ class Client(Node):
             heartbeat_interval = dask.config.get("distributed.client.heartbeat")
         heartbeat_interval = parse_timedelta(heartbeat_interval, default="ms")
 
+        scheduler_info_update = parse_timedelta(
+            dask.config.get("distributed.client.scheduler-info-update")
+        )
+
         self._periodic_callbacks = dict()
         self._periodic_callbacks["scheduler-info"] = PeriodicCallback(
-            self._update_scheduler_info, 2000, io_loop=self.loop
+            self._update_scheduler_info, scheduler_info_update * 1000, io_loop=self.loop
         )
         self._periodic_callbacks["heartbeat"] = PeriodicCallback(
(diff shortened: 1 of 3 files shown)


distributed/distributed.yaml
```diff
@@ -96,10 +96,12 @@ distributed:
     preload-argv: []        # See https://docs.dask.org/en/latest/setup/custom-startup.html
 
   client:
-    heartbeat: 5s  # time between client heartbeats
+    heartbeat: 5s  # Interval between client heartbeats
+    scheduler-info-update: 2s  # Interval between scheduler-info updates
```
jrbourbeau, line 100:
```suggestion
    scheduler-info-interval: 2s  # Interval between scheduler-info updates
```

distributed/distributed.yaml
```diff
@@ -96,10 +96,12 @@ distributed:
...
 
   deploy:
     lost-worker-timeout: 15s  # Interval after which to hard-close a lost worker job
+    cluster-repr-update: 500ms  # Interval between calls to update cluster-repr for the widget
```
jrbourbeau, line 104:
```suggestion
    cluster-repr-interval: 500ms  # Interval between calls to update cluster-repr for the widget
```

distributed/client.py
```diff
@@ -677,9 +677,13 @@ def __init__(
             heartbeat_interval = dask.config.get("distributed.client.heartbeat")
         heartbeat_interval = parse_timedelta(heartbeat_interval, default="ms")
 
+        scheduler_info_update = parse_timedelta(
+            dask.config.get("distributed.client.scheduler-info-update")
+        )
```
jrbourbeau, line 682: By default `parse_timedelta` will assume an input numeric value is in seconds, so the `* 1000` factor will be off in that case. I think we should specify `default="ms"` here, similar to when we parse `heartbeat_interval`

jsignell: Oh I see. I misunderstood what that default was for.

PR #3725, opened 2020-04-17
Use PeriodicCallback class from tornado

jrbourbeau:
We have a `PeriodicCallback` compatibility class which wraps tornado's `PeriodicCallback` and has some logic for dealing with tornado versions < 5

https://github.com/dask/distributed/blob/07b0cfeef4d2515361c1ee89222c18d30ad26a67/distributed/utils.py#L1120-L1128

Since today we require `tornado >= 5`, we can drop the compatibility class and just use `PeriodicCallback` from tornado

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -37,7 +37,7 @@ try:
 except ImportError:
     single_key = first
 from tornado import gen
-from tornado.ioloop import IOLoop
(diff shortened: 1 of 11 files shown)


mrocklin:
Woot

jrbourbeau:
Awesome, thanks for reviewing! I'll plan to merge this once CI has passed (except for a few known flaky tests)

PR #3726, opened 2020-04-18
Fix flaky test_oversubscribing_leases

### LINKED ISSUE #3717: Test failure in test_oversubscribing_leases
This showed up in a recent PR.  I copy-pasted the traceback and then restarted the job. 
 Unfortunately my attempt to copy-paste the trace turned out to be incomplete.  My apologies.  Hopefully this helps though.

cc @fjetter 

```python-traceback
>   assert not sem.acquire(timeout=0.05)
E   assert not True
E    +  where True = <bound method Semaphore.acquire of <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>>(timeout=0.05)
E    +    where <bound method Semaphore.acquire of <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>> = <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>.acquire

distributed\tests\test_semaphore.py:396: AssertionError
---------------------------- Captured stderr call -----------------------------
distributed.worker - WARNING -  Compute Failed
Function:  observe_state
args:      ()
kwargs:    {'sem': <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>}
Exception: AssertionError('assert not True\n +  where True = <bound method Semaphore.acquire of <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>>(timeout=0.05)\n +    where <bound method Semaphore.acquire of <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>> = <distributed.semaphore.Semaphore object at 0x0000027BF23DE438>.acquire',)

```

fjetter:
This should finally fix #3717 by introducing two changes

* A new method get_value to observe the state. This eliminates any timing based assumption on the oversubscription but allows us to wait until that condition is given for certain
* For some reason, the PeriodicCallback start/stop within the test did not work as expected. I got stop the callback, I could start it again but for whatever reason it was not doing anything. That left the entire test in more or less random state and I'm wondering why it worked in the first place. To eliminate the randomness I control the periodic callback via the `refresh_leases` variable now instead of start-/stopping it. Also, I explicitly verify that the internals do what they are supposed to do by capturing the logs. Has anybody else encountered a PC behaviour like this before? Is start/stop/start something we are not supposed to do with PCs?

I couldn't reproduce the race condition in the first point but I am now more confident that the test is doing what it was intended to do when running it locally and I don't see any more room for race conditions atm.

Sorry for the flaky test, I hope this fixes the issue now

cc @mrocklin

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -106,7 +106,7 @@ class SemaphoreExtension:
             for id_ in lease_ids:
                 if id_ not in self.leases[name]:
                     logger.critical(
-                        f"Trying to refresh an unknown lease ID {id_} for {name}. This might be due to leases "
+                        f"Refreshing an unknown lease ID {id_} for {name}. This might be due to leases "
                         f"timing out and may cause overbooking of the semaphore!"
                         f"This is often caused by long-running GIL-holding in the task which acquired the lease."
                     )
@@ -353,6 +353,7 @@ class Semaphore:
         self._periodic_callback_name = f"refresh_semaphores_{self.id}"
         self.client._periodic_callbacks[self._periodic_callback_name] = pc
         pc.start()
+        self.refresh_leases = True
 
     def __await__(self):
         async def create_semaphore():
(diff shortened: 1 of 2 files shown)


mrocklin:
Thank you for following up on this @fjetter !  I appreciate it.

PR #3729, opened 2020-04-20
Client.get_dataset to always create Futures attached to itself

### LINKED ISSUE #2336: Unable to use published datasets in a different client

### LINKED ISSUE #3227: Published Future retrieved from async client is synchronous
distributed-2.7.0
Possibly related to https://github.com/dask/distributed/issues/2336.

I'm using an asynchronous client connected to a remote cluster.
A publish_dataset -> get_dataset roundtrip on a Future causes it to be attached to a synchronous client.

```python
client = await distributed.Client("localhost:8786", asynchronous=True)
a, = client.compute([delayed(1, pure=True)])
print(a)  # <Future: status: pending, key: int-62645d78d66e2508256b7ab60a38b944>
print(a.result())  # <coroutine object Future._result at 0x11ad10dd0>
print(await a)  # 1

await client.publish_dataset(foo=a)
b = await client.get_dataset("foo")

print(b)  # <Future: status: pending, key: int-62645d78d66e2508256b7ab60a38b944>
print(client.asynchronous, b.client.asynchronous) # True, False
print(b.result())  # 1
print(await b)  # AttributeError: 'int' object has no attribute '__await__'
```

Switching to a LocalCluster weirdly makes the issue disappear.

Workarounds:
- Force ``b.client = client``
- Use ``client.gather([b])`` instead, which (involuntarily?) seems to ignore b.client.

crusaderky:
- Fix error "Inputs contain futures that were created by another client"
- Closes #3227
- Closes #2336

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -102,7 +102,6 @@ def _get_global_client():
             return c
         else:
             del _global_clients[k]
-    del L
     return None
 
 
@@ -1339,6 +1338,10 @@ class Client(Node):
             timeout = self._timeout * 2
         # XXX handling of self.status here is not thread-safe
         if self.status == "closed":
+            if self.asynchronous:
+                future = asyncio.Future()
+                future.set_result(None)
+                return future
             return
         self.status = "closing"
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -30,7 +30,7 @@ from dask.optimization import SubgraphCallable
 from dask.compatibility import apply
 from dask.utils import ensure_dict, format_bytes, funcname
 
-from tlz import first, groupby, merge, valmap, keymap
+from tlz import first, groupby, merge, valmap, keymap, partition_all
 
 try:
     from dask.delayed import single_key
@@ -1044,7 +1044,7 @@ class Client(Node):
 
         try:
             comm = await connect(
-                self.scheduler.address, timeout=timeout, **self.connection_args,
+                self.scheduler.address, timeout=timeout, **self.connection_args
             )
             comm.name = "Client->Scheduler"
             if timeout is not None:
@@ -1540,6 +1540,7 @@ class Client(Node):
         actor=False,
         actors=False,
         pure=None,
+        batch_size=None,
         **kwargs,
     ):
         """ Map a function on a sequence of arguments
@@ -1579,6 +1580,11 @@ class Client(Node):
             See :doc:`actors` for additional details.
         actors: bool (default False)
             Alias for `actor`
+        batch_size : int, optional
+            Submit tasks to the scheduler in batches of (at most) ``batch_size``.
+            Larger batch sizes can be useful for very large ``iterables``,
+            as the cluster can start processing tasks while later ones are
+            submitted asynchronously.
         **kwargs: dict
             Extra keywords to send to the function.
             Large values will be included explicitly in the task graph.
@@ -1596,11 +1602,6 @@ class Client(Node):
         --------
         Client.submit: Submit a single function
         """
-        key = key or funcname(func)
-        actor = actor or actors
-        if pure is None:
-            pure = not actor
-
         if not callable(func):
             raise TypeError("First input to map must be a callable function")
 
@@ -1611,6 +1612,38 @@ class Client(Node):
                 "Dask no longer supports mapping over Iterators or Queues."
                 "Consider using a normal for loop and Client.submit"
             )
+        total_length = sum(len(x) for x in iterables)
+
+        if batch_size and batch_size > 1 and total_length > batch_size:
+            batches = list(
+                zip(*[partition_all(batch_size, iterable) for iterable in iterables])
+            )
+            return sum(
+                [
+                    self.map(
+                        func,
+                        *batch,
+                        key=key,
+                        workers=workers,
+                        retries=retries,
+                        priority=priority,
+                        allow_other_workers=allow_other_workers,
+                        fifo_timeout=fifo_timeout,
+                        resources=resources,
+                        actor=actor,
+                        actors=actors,
+                        pure=pure,
+                        **kwargs,
+                    )
+                    for batch in batches
+                ],
+                [],
+            )
+
+        key = key or funcname(func)
+        actor = actor or actors
+        if pure is None:
+            pure = not actor
 
         if allow_other_workers and workers is None:
             raise ValueError("Only use allow_other_workers= if using workers=")
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2248,11 +2248,10 @@ class Client(Node):
     async def _get_dataset(self, name):
         out = await self.scheduler.publish_get(name=name, client=self.id)
         if out is None:
-            raise KeyError("Dataset '%s' not found" % name)
+            raise KeyError(f"Dataset '{name}' not found")
 
         with temp_default_client(self):
-            data = out["data"]
-        return data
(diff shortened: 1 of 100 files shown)


distributed/client.py
```diff
@@ -90,6 +91,7 @@
 _global_clients = weakref.WeakValueDictionary()
 _global_client_index = [0]
 
+_deserialize_client = ContextVar("_deserialize_client", default=None)
```
jcrist, line 94: Do we need to use contextvars here? We currently use a thread local for storing these kind of things, but that doesn't always play nicely with async. Is there a way we could refactor part of distributed to avoid passing around this implicit state?

In general I think`contextvars` can be quite useful where required, but want to ensure it's needed here before bringing it in. We also already have a fair bit of implicit state, I'm a bit hesitant to add more if we can avoid it.

distributed/tests/test_publish.py
```diff
@@ -229,3 +246,34 @@ async def test_pickle_safe(c, s, a, b):
 
         with pytest.raises(TypeError):
             await c2.get_dataset("z")
+
+
+@gen_cluster(client=True)
+async def test_deserialize_client(c, s, a, b):
```
jcrist, line 252: Nice test case.

distributed/client.py
crusaderky: To avoid using contextvars you need to explicitly propagate the client until you stop calling async functions and just call sync functions. In this case, this means that

- ``client.Client._get_dataset`` needs to pass ``self`` to
- ``core.send_recv``, which in turn needs to pass it to
- ``comm.tcp.TCP.read`` (plus its variants ``comm.inproc.InProc.read`` and the 100% untested ``comm.ucx.UCX.read``), which in turn needs to pass it to
- ``comm.utils.from_frames``, which can finally put a context manager (setting a thread-local variable) around
https://github.com/dask/distributed/blob/8376f227e288757c3fd9b1e8742ca350e6e57b25/distributed/comm/utils.py#L60-L62
- which will alter the behaviour of ``client.Future.__getstate__``

and it should be really self-evident why I don't recommend doing it.

> We currently use a thread local for storing these kind of things

For this one you were using a global, ``client._global_clients``.
contextvars exactly fits the use case of thread locals when you use coroutines.

jcrist: Makes sense to me. Should this logic replace how we access implicit clients everywhere? If so, perhaps we should rename the context var key?

crusaderky: Are there other cases where a method of Client loses ``self`` along the way and re-acquires it later from a global variable?

Also note the very big fat caveat of the race conditions on Python 3.6. Not sure how much of the dask.distributed user base still uses Python 3.6 AND has more than one client running.

TomAugspurger:
@jcrist are you able to shepherd this PR through?

distributed/client.py
jcrist: Sorry, let me clarify.

It feels a bit weird to handle setting the contextvar in the `_get_dataset` method. It might be cleaner to have a parent context over all coroutines started by a client that includes a reference back to the client. As it stands this is baking in functionality to `Future` that is only used by one method, when we likely want this everywhere.

crusaderky: > It might be cleaner to have a parent context over all coroutines started by a client that includes a reference back to the client. 

Are you talking to some sort of blanket decorator applied around all methods of Client? If so, it feels a bit overkill to me, as I do not know of another method that needs it.

>  As it stands this is baking in functionality to Future that is only used by one method, when we likely want this everywhere.

You definitely want to reuse this same variable every time you have a ``self`` reference to a Client instance, invoke a method, lose the reference (because it would be too cumbersome to propagated by hand), and then you need to reacquire it deeper down the line _before the end of the method_. If you can point me to any other function that does this, I'll be happy to look into it.

crusaderky: @jcrist ping

crusaderky: I'm redesigning the change to make it more generic, hold on

crusaderky:
@jcrist I redesigned the PR to address your concerns - see if you like it now

crusaderky:
@jcrist ping

crusaderky:
@jcrist any feedback please?

distributed/client.py
```diff
@@ -727,10 +732,42 @@ def __init__(
...
+        """Thread-local, Task-local context manager that causes the Client.current class
+        method to return self. This is used when a method of Client needs to propagate a
+        reference to self deep into the stack through generic methods that shouldn't be
+        aware of this class.
+        """
+        # Python 3.6; contextvars are thread-local but not Task-local.
+        # We can still detect a race condition.
+        if sys.version_info < (3, 7) and _current_client.get() not in (self, None):
+            raise RuntimeError(
+                "Detected race condition where get_dataset() is invoked in "
```
jcrist, line 746: This error message still mentions `get_dataset`, should probably be made a bit more generic.

distributed/client.py
```diff
@@ -727,10 +732,42 @@ def __init__(
 
         ReplayExceptionClient(self)
 
+    @contextmanager
+    def as_current(self):
+        """Thread-local, Task-local context manager that causes the Client.current class
+        method to return self. This is used when a method of Client needs to propagate a
+        reference to self deep into the stack through generic methods that shouldn't be
+        aware of this class.
```
jcrist, line 740: Since this only sets the contextvar (and thus only works in an async context), I think this method may be confusing for most users. Perhaps we should keep this as a private method for now (i.e. move to `_as_current(self)`)?

distributed/client.py
crusaderky: Fixed

distributed/client.py
crusaderky: The big use case for a final user is when he pickles a dask collection that internally contains futures, e.g. to dump it to disk for forensics purposes, and then wants to unpickle it and he has more than one Client open. I updated the description to reflect this use case, see if you like it.

distributed/tests/test_client.py
```diff
@@ -3876,38 +3876,148 @@ async def test_scatter_compute_store_lose_processing(c, s, a, b):
...
+    await c1.close()
+    await c2.close()
 
-    with temp_default_client(f):
-        assert default_client() is f
-        assert default_client(c) is c
 
-    await c.close()
-    await f.close()
+def test_as_current_is_thread_local(s):
```
jcrist, line 3935: Clever test

crusaderky:
Getting a completely unrelated failure on Windows since the latest merge from master

jcrist:
Yeah, the test failures are unrelated.

jcrist:
Test failure is unrelated. Thanks @crusaderky, merging.

pborgen:
Thanks for this fix. .Cannot wait to try it out...

PR #3731, opened 2020-04-20
Relax NumPy requirement in UCX

jakirkham:
This relaxes the NumPy requirement in UCX communication. We do this by leveraging things like `struct.pack` and `struct.unpack` for metadata about the frames, which gives us simple `bytes` objects to send over the wire. Also we create a `host_array` function that will use NumPy when available, but will fallback to things like `bytearray` if not.

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -239,9 +239,7 @@ class UCX(Comm):
             else:
                 # Recv frames
                 frames = [
-                    device_array(each_size)
-                    if is_cuda
-                    else host_array(each_size)
+                    device_array(each_size) if is_cuda else host_array(each_size)
                     for is_cuda, each_size in zip(cuda_frames, sizes)
                 ]
                 recv_frames = [
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -10,7 +10,6 @@ import struct
 import weakref
 
 import dask
-import numpy as np
 
 from .addressing import parse_host_port, unparse_host_port
 from .core import Comm, Connector, Listener, CommClosedError
@@ -34,6 +33,7 @@ logger = logging.getLogger(__name__)
 # required to ensure Dask configuration gets propagated to UCX, which needs
 # variables to be set before being imported.
 ucp = None
+host_array = None
 device_array = None
 
 
@@ -47,7 +47,7 @@ def synchronize_stream(stream=0):
 
 
 def init_once():
-    global ucp, device_array
+    global ucp, host_array, device_array
     if ucp is not None:
         return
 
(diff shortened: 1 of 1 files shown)


quasiben:
Is there any expected performance improvement by using struct vs numpy ?

cc @madsbk

jakirkham:
Mostly hoping this simpler implementation avoids red herrings while debugging issues.

jakirkham:
In terms of performance this does improve things a little bit, but it probably doesn't impact the overall send/receive process meaningfully (though I could be wrong about that). Again this isn't really the motivation for it.

```python
In [1]: import struct                                                           

In [2]: import numpy                                                            

In [3]: l = 4 * [True, False]                                                   

In [4]: %timeit struct.pack(len(l) * "?", *l)                                   
389 ns ± 4.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit numpy.array(l)                                                  
838 ns ± 2.67 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

distributed/comm/ucx.py
```diff
@@ -59,34 +60,42 @@ def init_once():
 
     ucp.init(options=ucx_config, env_takes_precedence=True)
 
+    # Find the function, `host_array()`, to use when allocating new host arrays
+    try:
```
quasiben, line 64: I don't think we need the try except as ucx-py depends on numpy: https://github.com/rapidsai/ucx-py/blob/branch-0.14/setup.py#L73

jakirkham: That may be so. However that would be like saying NumPy can rely on Pandas dependencies for code that Pandas uses from NumPy. In general we can't expect that to hold and it's a bit fragile. We should either make NumPy a requirement here or add a fallback. Seems safer to add this fallback here and it doesn't cost us much. Plus it should save us suffering later should things change in UCX-Py and we forget about this line here 🙂

quasiben:
That's a good point.

jakirkham:
Went ahead and pulled the header consolidation commit from PR ( https://github.com/rapidsai/rmm/issues/318 ) and pushed it here. That doesn't really change how the frames are serialized, but it does allow us to send all of the per frame metadata in one message.

quasiben:
test_ucx.py passed locally with one exception:
> distributed/comm/tests/test_ucx.py::test_ucx_deserialize FAILED

though this is probably due to hostname issues.  

While I don't think this changes the overall scheme described in the notes, lines like:
> struct.pack(nframes * "?" + nframes * "Q", *cuda_frames, *sizes)

will be hard to read tomorrow.  Would you mind adding a note or two describe what packing is doing ?

distributed/comm/ucx.py
quasiben: i thought i hit send last night on this comment but I didn't.  It said,

That's a good point!  Thanks

quasiben:
Thank you @jakirkham for taking the time to add those comments

jakirkham:
Thanks Ben! 😄

Yeah I've never gotten that test to run successfully. 😞 Probably we should figure out how the config issue should be fixed and talk to OPS so it can be integrated into their provisioning scripts.

Was just about to ask if those comments seemed reasonable. Happy to adjust if needed 🙂

Issue #3489, opened 2020-02-17
client.get() on an unknown key hangs forever


OliverHofkens:
### Issue
Calling `client.get(graph, [keys, ...])` with a key that does not exist hangs forever. The timeout argument is not respected in this case. 

### Expected behavior
A clear error, the threaded scheduler throws a friendly `KeyError` for example.

### Reproducible example

```python
import operator as op
from distributed import Client, wait
from dask import delayed

client = Client()

result = delayed(op.add)(1, 2)
graph = result.__dask_graph__()

client.get(graph, ["does_not_exist"], timeout="1s")  # <--- hangs
```

Interrupting gives the following stack trace:

```python-traceback
---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-3-c3eb58815d0d> in <module>
      1 result = delayed(op.add)(1, 2)
      2 graph = result.__dask_graph__()
----> 3 client.get(graph, ["does_not_exist"], timeout="1s")

/usr/local/lib/python3.7/site-packages/distributed/client.py in get(self, dsk, keys, restrictions, loose_restrictions, resources, sync, asynchronous, direct, retries, priority, fifo_timeout, actors, **kwargs)
   2590                     should_rejoin = False
   2591             try:
-> 2592                 results = self.gather(packed, asynchronous=asynchronous, direct=direct)
   2593             finally:
   2594                 for f in futures.values():

/usr/local/lib/python3.7/site-packages/distributed/client.py in gather(self, futures, errors, direct, asynchronous)
   1888                 direct=direct,
   1889                 local_worker=local_worker,
-> 1890                 asynchronous=asynchronous,
   1891             )
   1892 

/usr/local/lib/python3.7/site-packages/distributed/client.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    766         else:
    767             return sync(
--> 768                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
    769             )
    770 

/usr/local/lib/python3.7/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    340     else:
    341         while not e.is_set():
--> 342             e.wait(10)
    343     if error[0]:
    344         typ, exc, tb = error[0]

/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py in wait(self, timeout)
    550             signaled = self._flag
    551             if not signaled:
[...]

PR #3736, opened 2020-04-22
add timeout option with default to 10s

prasunanand:
Fixes https://github.com/dask/distributed/issues/3489


distributed/client.py
```diff
@@ -1871,7 +1871,7 @@ async def _gather_remote(self, direct, local_worker):
 
         return response
 
-    def gather(self, futures, errors="raise", direct=None, asynchronous=None):
+    def gather(self, futures, errors="raise", direct=None, asynchronous=None, callback_timeout=None):
```
TomAugspurger, line 1874: I think we call this `timeout` elsewhere

```suggestion
    def gather(self, futures, errors="raise", direct=None, asynchronous=None, timeout=None):
```

distributed/client.py
```diff
@@ -2573,6 +2574,7 @@ def get(
         direct=None,
         retries=None,
         priority=0,
+        timeout=10,
```
TomAugspurger, line 2577: Does this match the current behavior, or should it be `None`?

prasunanand:
Thanks @TomAugspurger, I have modified the Pull Request. Now I raise KeyError if key is not found in the graph.

Thanks @jrbourbeau, I got confused because of [line](https://github.com/dask/distributed/blob/master/distributed/client.py#L776) . I thought a callback timeout was excepted.

sublinus:
Considering the original issue maybe you could go for a test similar to this one:
```python
def test_get_unknown_key(c):
    result = delayed(add)(1, 2)
    graph = result.__dask_graph__()
    with pytest.raises(KeyError):
        c.get(graph, ["does_not_exist"])
```

distributed/tests/test_client.py
```diff
@@ -625,6 +625,10 @@ async def test_get(c, s, a, b):
 
 def test_get_sync(c):
     assert c.get({"x": (inc, 1)}, "x") == 2
+    try:
+        c.get({"y": (inc, 1)}, "z")
+    except Exception as e:
+        assert "not found in the dask graph" in str(e)
```
jrbourbeau, line 631: As @sublinus pointed out, we tend to use `pytest.raises` to check that an error has been raised. This provides useful feedback when the expected error isn't raised and has nice error message checking functionality. In general we also try to be as specific as we can when catching errors. Here that would mean catching a `KeyError` instead of `Exception`. Something like the following would be the  `pytest.raises` equivalent: 

```python
with pytest.raises(KeyError, match="not found in the dask graph"):
    c.get({"y": (inc, 1)}, "z")
```

distributed/client.py
```diff
@@ -2620,6 +2620,11 @@ def get(
             user_priority=priority,
             actors=actors,
         )
+
+        for key in set(flatten([keys])):
+            if not key in dict(dsk):
```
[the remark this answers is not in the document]
jrbourbeau: Also, there's a linting error here in the CI were we want to use `key not in dsk` instead of `not key in dsk`

distributed/client.py
```diff
@@ -2620,6 +2620,11 @@ def get(
             user_priority=priority,
             actors=actors,
         )
+
+        for key in set(flatten([keys])):
```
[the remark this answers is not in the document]
jrbourbeau: A small point, but instead of doing `set(flatten([keys]))` twice (once here and once in the `_graph_to_futures` call, could you define `keys_flatten = set(flatten([keys]))` first and then use that variable in both places

prasunanand:
Thanks @sublinus and @jrbourbeau , I have made the changes as you suggested.

TomAugspurger:
Just to verify the desired behavior here... On master, if I do

```python
# process 1
>>> client1 = Client("path/to/scheduler")
>>> client1.get({}, "a")
```

```python
# process 2
>>> client2 = Client("path/to/scheduler")
>>> client2.get({"a": 1}, "a")
```

Then the behavior I see is that `client1` gets it's result now that the scheduler knows about that key.

IIUC, this PR will change that behavior. Is that something we would want to support, or should we require that the requested `key` be in the actual graph submitted?

jrbourbeau:
That's a good point, thanks for bringing it up @TomAugspurger. I personally find the current behavior in `master` non-intuitive - though this is subjective. Instead I'd prefer to guide users towards creating a `Future` for a key and then calling `result()` on that `Future`. I'm certainly open to hearing about what others think though

TomAugspurger:
Agreed that the current behavior is unintuitive. I'm comfortable claiming that it "accidentally worked" unless we document that behavior somewhere.

distributed/client.py
```diff
@@ -2609,9 +2609,15 @@ def get(
         --------
         Client.compute: Compute asynchronous collections
         """
+        keys_flatten = set(flatten([keys]))
+
+        for key in keys_flatten:
+            if key not in dsk:
+                raise KeyError("Key '%s' not found in the dask graph" % key)
```
TomAugspurger, line 2616: Can you check the performance of this vs. 

```python
if keys_flatten - set(dsk):
```

I suppose the performance will depend on the relative  sizes of `keys_flatten` and `dsk`.

prasunanand: I get the following metrics:
```
def flatten(seq, container=list):
    if isinstance(seq, str):
        yield seq
    else:
        for item in seq:
            if isinstance(item, container):
                for item2 in flatten(item, container=container):
                    yield item2
            else:
                yield item

dsk = {"x": (op.add, 1, 2)}

keys = ["x", "y"]
                
keys_flatten = set(flatten([keys]))
```
```
def f1():
    for key in keys_flatten:
        if key not in dsk:
#             raise KeyError("Key '%s' not found in the dask graph" % key)
            pass
            
%timeit f1()
```
`147 ns ± 1.86 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)`

```
def f2():
    not_found = keys_flatten - set(dsk)
#     raise KeyError("Key '%s' not found in the dask graph" % key)
    pass
    
%timeit f2()
```
`230 ns ± 8.82 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
​`

TomAugspurger: Thanks. I'm more concerned when `dsk` is very large.

It seems like this might fit a bit better in `_graph_to_futures`, where we already iterate over `dsk` a few times, and already have the set of keys (called `keyset`). Can you see if there's a good spot somewhere there?

prasunanand: This [line](https://github.com/dask/distributed/blob/2d54ef947a7fe372b1fafc40d2f2b89ff88e9449/distributed/client.py#L2496) could be a spot, but its not possible because that defeats the purpose of using `futures`. And a key can exist in dsk but not in keyset.
Another spot is [line](https://github.com/dask/distributed/blob/2d54ef947a7fe372b1fafc40d2f2b89ff88e9449/distributed/client.py#L2544).

This line could be moved to the top of `_graph_to_futures` function body:
```
futures = {}
for key in keyset:
            if key in dsk:
                futures[key] = Future(key, self, inform=False)
            else:
                raise KeyError("Key '%s' not found in the dask graph" % key)
```

prasunanand: Adding the lines to `_graph_to_futures` breaks `client#submit` calls.

PR #3740, opened 2020-04-25
Replace OrderedDict with dict in scheduler

mrocklin:
This is no longer necessary (I think) now that we require Python 3.6+


TomAugspurger:
dicts didn't become officially ordered in Python till 3.7. Them being ordered in 3.6 was a CPython implementation details. As long as we're comfortable with that this can be merged.

Other projects like pandas just treat dicts as ordered in 3.6 and we haven't had any issues.

jrbourbeau:
Thanks @mrocklin!

PR #3742, opened 2020-04-26
Fix confusing error if compression not available

necaris:
Currently, if a compression method isn't available on the client, it raises a `KeyError` with the name of the method used (e.g. `lz4`). This updates the decompress function to raise `InvalidCompressionMethodError` when that happens, which should be easier to interpret.

It also adds an obtrusive "Danger!" warning when connecting to a scheduler if the client doesn't support the scheduler's default compression method. 

I would appreciate some pointers as to how to make this configurable to throw an error ([see note in the diff](https://github.com/dask/distributed/pull/3742/files#diff-6a3108098afcddf88b38652ae1e383b5R2532)) which I think might be a good better-safe-than-sorry behavior. I'm also not sure how best to test this -- thanks in advance for the help!

### DIFF
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -211,14 +211,12 @@ def maybe_compress(payload, min_size=1e4, sample_size=1e4, nsamples=5):
 
 def decompress(header, frames):
     """ Decompress frames according to information in the header """
-    return [
-        _decompress_frame(c, frame) for c, frame in zip(header["compression"], frames)
-    ]
-
-
-def _decompress_frame(method, frame):
-    """Decompress a single frame"""
-    compressor = compressions.get(method, None)
-    if compressor is None:
-        raise ValueError("Invalid compression method: %s" % method)
-    return compressor["decompress"](frame)
+    try:
+        return [
+            compressions[c]["decompress"](c, frame)
+            for c, frame in zip(header["compression"], frames)
+        ]
+    except KeyError as e:
+        # while it's not documented behavior, the presence of the offending key
+        # at `e.args[0]` is pretty consistent across versions of Python
+        raise ValueError("Invalid compression method: %s" % e.args[0])
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -218,8 +218,7 @@ def maybe_compress(payload, min_size=1e4, sample_size=1e4, nsamples=5):
 def decompress(header, frames):
     """ Decompress frames according to information in the header """
     return [
-        _decompress_frame(c, frame)
-        for c, frame in zip(header["compression"], frames)
+        _decompress_frame(c, frame) for c, frame in zip(header["compression"], frames)
     ]
 
 
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -213,7 +213,7 @@ def decompress(header, frames):
     """ Decompress frames according to information in the header """
     try:
         return [
-            compressions[c]["decompress"](c, frame)
+            compressions[c]["decompress"](frame)
             for c, frame in zip(header["compression"], frames)
         ]
     except KeyError as e:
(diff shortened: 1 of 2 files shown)


distributed/protocol/compression.py
```diff
@@ -10,6 +10,13 @@
 import dask
 from tlz import identity
 
+
+class InvalidCompressionMethodError(ValueError):
+    def __init__(self, method):
+        self.method = method
+        super().__init__("Invalid compression method: %s" % method)
```
mrocklin, line 17: We have historically not created custom error classes within the Dask projects, and have instead tended towards the base Value/Type/ImportErrors.  This isn't a rule by any means, but making a type for this is a bit odd for the project.

distributed/protocol/compression.py
```diff
@@ -211,6 +218,13 @@ def maybe_compress(payload, min_size=1e4, sample_size=1e4, nsamples=5):
 def decompress(header, frames):
     """ Decompress frames according to information in the header """
     return [
-        compressions[c]["decompress"](frame)
-        for c, frame in zip(header["compression"], frames)
+        _decompress_frame(c, frame) for c, frame in zip(header["compression"], frames)
```
mrocklin, line 221: Rather than have the helper function here I wonder if this could instead be ...

```python
try:
    return [
        compressions[c]["decompress"](frame)
        for c, frame in zip(header["compression"], frames)
    ] 
except KeyError:
    raise ValueError("Invalid compression method: %s" % method)
```

This is a little less fancy, but may be easier for future maintainers to quickly understand without having to track down the `_decompress_frame` method and the `InvalidCompressionMethodError` type.  Hrm, I guess that's hard though because `c` is no longer in scope.  Maybe we go even less fancy and do this in a for loop?  Or perhaps I'm optimizing too much for low-tech.

distributed/scheduler.py
```diff
@@ -2523,11 +2523,27 @@ async def add_client(self, comm, client=None, versions=None):
...
+                from .protocol import compression
+
+                if not client_versions.get(compression.default_compression, None):
+                    version_warning = version_warning or ""
+                    version_warning += """
+
+DANGER! The scheduler's default compression method is not available
+in this client! You may not be able to receive large result sets!
+                    """
+
```
mrocklin, line 2546: I like the idea of calling out specific library versions that may be particularly concerning.

If we do this though then we might want to be more comprehensive.  The compression mismatch can cause a serious issue if any two components (clients, schedulers, or workers) disagree with each other, not just the client/scheduler relationship.  In an ideal world two peers would do a little handshake in order to determine which compression to use.  We're not there yet today though.

Short term though, if we want to put this somewhere I wonder if we want it to be in `versions.py::error_message`.  I haven't actually dug into this yet, so please don't rush and change a bunch of things.  This is mostly a question.

distributed/protocol/compression.py
necaris: :+1: thanks for the feedback! Simplified this to a `ValueError` in https://github.com/dask/distributed/pull/3742/commits/d712c1d699436f1ae6b8bbcab52b2cf0993561ce

distributed/protocol/compression.py
necaris: Sure thing -- simplified out the helper method in https://github.com/dask/distributed/pull/3742/commits/78443f547b0d81310559a4898d336e5df3bb1ee7 , in favor of pulling the erroring key from the `KeyError.args`.

distributed/scheduler.py
necaris: I'd be totally fine putting it in `versions::error_message` for the moment, and rely on the `Invalid compression method` message for the user to recognize the error even if they aren't paying attention to the warning. I'd like to note that while I understand the warning _is_ shown by default, it's really easy to skim over, especially when most of the mismatches are likely to be minor version issues that probably won't cause problems. So I'd ideally like to have some way to fail fast on connecting, if an error is likely.

 I'll hold off on making changes for now, though -- thank you for taking a look!

necaris:
Closing in favor of https://github.com/dask/distributed/pull/3936 which addresses the more general case of mismatched libraries / Python versions.

Issue #3640, opened 2020-03-26
Efficient serialization of `memoryview`s


jakirkham:
It would be useful to have efficient serialization of `memoryview`s as these show up in lots of places or are easy enough for other codes to generate (given they are builtin component of Python). Ideally this would be very similar to NumPy's `ndarray` serialization with some tweaks. It also has the potential to simplify a bunch of serialization code as it would then be able to pass things off to/take things from `memoryview` serialization.

Currently this is getting handled by cloudpickle, which is [performing a copy to a `bytes` object before pickling]( https://github.com/cloudpipe/cloudpickle/blob/d8452cc3a6f4bcd797b4edaf13b349e5440f5425/cloudpickle/cloudpickle_fast.py#L274-L275 ). An efficient serialization method of `memoryview` objects should be able to avoid this copy as well as pickling.

```python
In [1]: from distributed.protocol.serialize import serialize, deserialize       

In [2]: b = b"abc"                                                              

In [3]: serialize(b)                                                            
Out[3]: 
({'type': 'builtins.bytes',
  'type-serialized': b'\x80\x04\x95\x16\x00\x00\x00\x00\x00\x00\x00\x8c\x08builtins\x94\x8c\x05bytes\x94\x93\x94.',
  'serializer': 'dask'},
 [b'abc'])

In [4]: m = memoryview(b)                                                       

In [5]: serialize(m)                                                            
Out[5]: 
({'serializer': 'pickle'},
 [b'\x80\x04\x95\x07\x00\x00\x00\x00\x00\x00\x00C\x03abc\x94.'])

In [6]: deserialize(serialize(m)) 
```

mrocklin:
Yeah, passing through memoryviews would be ideal.  Thanks for bringing this up @jakirkham .  I assumed, incorrectly, that this was already happening.

jakirkham:
Me too.

I started a bit of work on this locally. Will try to get this in a WIP PR at some point so we can discuss 🙂

PR #3743, opened 2020-04-27
Memoryview serialisation

martindurant:
Fixes https://github.com/dask/distributed/issues/3640 (for discussion)

Does not create an intermediate bytes object for serialization, passes through the memoryview. That, of course, still gets sent as bytes on the wire, so for deserialization, we just wrap it.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -570,7 +570,11 @@ def _deserialize_bytes(header, frames):
 
 @dask_deserialize.register(memoryview)
 def _serialize_memoryview(header, frames):
-    return memoryview(b"".join(frames))
+    if len(frames) == 1:
+        out = frames[0]
+    else:
+        out = b"".join(frames)
+    return memoryview(out)
 
 
 #########################
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -556,7 +556,7 @@ def normalize_Serialized(o):
 
 
 # Teach serialize how to handle bytestrings
-@dask_serialize.register((bytes, bytearray))
+@dask_serialize.register((bytes, bytearray, memoryview))
 def _serialize_bytes(obj):
     header = {}  # no special metadata
     frames = [obj]
@@ -566,6 +566,11 @@ def _serialize_bytes(obj):
 @dask_deserialize.register((bytes, bytearray))
 def _deserialize_bytes(header, frames):
     return b"".join(frames)
+
+
+@dask_deserialize.register(memoryview)
+def _serialize_memoryview(header, frames):
+    return memoryview(b"".join(frames))
 
 
 #########################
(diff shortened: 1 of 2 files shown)


mrocklin:
+1

distributed/protocol/serialize.py
```diff
@@ -568,6 +568,11 @@ def _deserialize_bytes(header, frames):
     return b"".join(frames)
 
 
+@dask_deserialize.register(memoryview)
+def _serialize_memoryview(header, frames):
+    return memoryview(b"".join(frames))
```
TomAugspurger, line 573: Is it worth special casing length-1 `frames`, since then you can just return `frames[0]` and avoid the copy? Maybe?

martindurant: Good point

martindurant: Also true for `_deserialize_bytes`, right?

jakirkham:
Thanks Martin!

So one thing I'm curious about is there's a [fair bit of logic in NumPy serialization]( https://github.com/dask/distributed/blob/3de9973cafaf85809f712a719111b8db3839975d/distributed/protocol/numpy.py ), which we seem to be skipping here. For example it is possible to have Python objects much like one can in NumPy arrays. So it might make sense to [pickle in that case]( https://github.com/dask/distributed/blob/3de9973cafaf85809f712a719111b8db3839975d/distributed/protocol/numpy.py#L24-L27 ). Similarly we may want logic to [flatten N-D data]( https://github.com/dask/distributed/blob/3de9973cafaf85809f712a719111b8db3839975d/distributed/protocol/numpy.py#L63-L74 ), [cast larger types]( https://github.com/dask/distributed/blob/3de9973cafaf85809f712a719111b8db3839975d/distributed/protocol/numpy.py#L76-L77 ), and [split large frames]( https://github.com/dask/distributed/blob/3de9973cafaf85809f712a719111b8db3839975d/distributed/protocol/numpy.py#L90-L93 ). Though we [don't need to worry about `datetime` or `timedelta` objects]( https://github.com/dask/distributed/blob/3de9973cafaf85809f712a719111b8db3839975d/distributed/protocol/numpy.py#L79-L83 ) as they are not supported types in `memoryview`s. Given this, what do you think about repurposing this logic for `memoryview`s and allowing NumPy serialization to just piggyback on `memoryview` serialization?

martindurant:
> t is possible to have Python objects

I am surprised to find that `memoryview.format == "O"` is indeed allowed. Does this actually ever happen? I suppose this just failed previously, or it never ever happened, but I am not sure.

I think the same goes for complex types and multiple dimensions - those should be already using numpy. I daresay that in every case here, we have a bytes-like thing (from arrow, whatever) that we don't want to accidentally copy.

In short, I am not keen to change the numpy side of things, even if it just amounts to moving code around a little, to support memoryviews that I'd be surprised if they got used. Prepard to be proven wrong...

jakirkham:
I think this goes back to the fact that the Python Buffer Protocol allows things like arrays of pointers. This was to make PIL developers happy with the PEP IIRC. So `memoryview.format == "O"` is one such case as one would have `PyObject*` in the array. In practice, this might come up when serializing a column of strings? 🤷‍♂️

Was mostly thinking we could get more out of the same code. Though you may be right this seldom comes up in practice.

Yeah I guess we wait and see if there's an issue. 🙂

martindurant:
>  In practice, this might come up when serializing a column of strings?

Since pandas and numpy already have their own serialisation, someone would *really* have to be trying to break Dask :)

Issue #3677, opened 2020-04-06
Improve discoverability of dashboard pages


mrocklin:
There are many pages hosted by the scheduler and workers' dashboard servers:

1.  The bokeh pages pointed to by the top navbar
2.  The secret bokeh pages aren't pointed to by anything currently (like profile-server, work stealing, and counters/digests)
3.  The info pages, which are pointed to by the top navbar
4.  The json API endpoints, which are currently secret, and not particularly well developed
5.  The prometheus metrics endpoint, currently secret, but in a predictable place
6.  The pew-pew-pew plot by @jacobtomlinson (I just like saying the name)

I suspect that no human other than myself is aware of all of these.  We should probably do something about that.

One solution would be to make an index page.

Issue #3734, opened 2020-04-21
test_pid_file failure


jrbourbeau:
I've started seeing `distributed/cli/tests/test_dask_scheduler.py::test_pid_file` sporadically fail in CI builds (xref https://travis-ci.org/github/dask/distributed/jobs/677794876#L1468-L1549)

<details>
<summary>Traceback:</summary>

```
________________________________ test_pid_file _________________________________

loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7f07c59270b8>

    def test_pid_file(loop):

        def check_pidfile(proc, pidfile):

            start = time()

            while not os.path.exists(pidfile):

                sleep(0.01)

                assert time() < start + 5

    

            text = False

            start = time()

            while not text:

                sleep(0.01)

                assert time() < start + 5

                with open(pidfile) as f:

                    text = f.read()

            pid = int(text)

            if sys.platform.startswith("win"):

                # On Windows, `dask-XXX` invokes the dask-XXX.exe

                # shim, but the PID is written out by the child Python process

                assert pid

            else:

                assert proc.pid == pid

    

        with tmpfile() as s:

            with popen(["dask-scheduler", "--pid-file", s, "--no-dashboard"]) as sched:

>               check_pidfile(sched, s)

distributed/cli/tests/test_dask_scheduler.py:207: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

../../../miniconda/envs/dask-distributed/lib/python3.6/contextlib.py:88: in __exit__

    next(self.gen)

distributed/utils_test.py:1041: in popen

    terminate_process(proc)

distributed/utils_test.py:1005: in terminate_process

    proc.wait(10)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <subprocess.Popen object at 0x7f07c592cda0>, timeout = 10

endtime = 190.384243671

    def wait(self, timeout=None, endtime=None):

        """Wait for child process to terminate.  Returns returncode

        attribute."""

        if self.returncode is not None:

            return self.returncode

    

        if endtime is not None:

            warnings.warn(

                "'endtime' argument is deprecated; use 'timeout3m'.",

                DeprecationWarning,

                stacklevel=2)

[...]

PR #3744, opened 2020-04-27
Discover HTTP endpoints

martindurant:
Fixes https://github.com/dask/distributed/issues/3677


martindurant:
@mrocklin , I have made a list here, but I'm not sure where it should go. This could be documentation, a live page, a JSON endpoint... 

Also, I'm sure I don't have them all yet. Although I'm not done, I have no idea what pew-pew is.

TomAugspurger:
>  I have no idea what pew-pew is.

That's the Official Name for the "Cluster map" dashboard added in https://github.com/dask/distributed/pull/3361

mrocklin:
Two thoughts:

1.  The different routes can change for a few reasons:
    - People can turn off different pieces of functionality in config
    - Hardware like GPUs can show up, which enables others
    - People can add in their own with config
2.  We should probably have links from the directory

For point 1, I wonder if there is some way that we can discover pages (maybe see distributed.http.utils.get_handlers plus whatever bokeh does) and grab descriptions of them from docstrings?

martindurant:
Something dynamic might be nice, but take a little working out - it means having a list somewhere of the places to look.

I also think that a static list like this, for the docs, is very useful, though

mrocklin:
> Something dynamic might be nice, but take a little working out - it means having a list somewhere of the places to look.

Yes, see `distributed.http.utils.get_handlers`

> I also think that a static list like this, for the docs, is very useful, though

No objection, but I suspect that less than 10% of our users will check that doc page, while I think that very many of them may check something like a collapsable sidebar on the dashboard.

martindurant:
Interestingly, bokeh endpoints come out like:
```
"/individual-graph/?$",
"/individual-graph/autoload.js$",
"/individual-graph/metadata$",
"/individual-graph/ws$",
```
It may be worthwhile to consolidate these (i.e., treat the bokeh application object as special), since it is "/individual-graph" that we expect people to navigate to. 
Also, it may look nicer to remove the "$"s, although they appear always to be part of the regex.

martindurant:
OK did that - I'll make a few tests before finishing up here.

martindurant:
Still didn't link the doc page to anywhere - happy to hear suggestions. @jrbourbeau , you mentioned somewhere it would fit?

martindurant:
Failures:
- test_preload_remote_module in github tests
- test_pid_file on travis

no idea why...

jrbourbeau:
I thought there was a list of some of the endpoints in the distributed docs, but I can't seem to find it now...

Re: failues.`test_preload_remote_module` has been fixed (sorry that's my fault) and `test_pid_file` is a [known flaky test](https://github.com/dask/distributed/issues/3734).

martindurant:
@jrbourbeau , any thought?

martindurant:
Ping here - any further requests?

jakirkham:
cc @jacobtomlinson (in case this is of interest 😉)

jsignell:
I wonder if there'd be a way to list what's included in statics ("/statics/individual-cluster-map.html" for example).  

Also, I think a top-level index would be very nice. Ideally, it'd have thumbnails, but that is more involved. I am happy for that to be follow on work though. I think this is a big step in the right direction for discoverability of these plots.

martindurant:
@jsignell , are you saying you'd like to see that change in this PR? I'm not sure how to go about listing statics, except really a file-list (which is easy and doable)

jsignell:
Yeah I'd like to see this in this PR.

martindurant:
@jsignell : done. This method may be specific to standard configuration of the tornado static handler, but works fine for the dashboard servers as they are right now.

martindurant:
@jsignell : finally :)

PR #3755, opened 2020-04-29
Support web addresses in preload

mrocklin:
This commit does two things:

1.  We support passing web addresses as preloads,
    allowing for for functionality like the following:

    dask-scheduler --preload http://my-web-address/myfile.py

2.  We refactor preloads into a class structure.
    I'm usually against this, but I think that in this case it cleans
    things up.
    I think that it will also make it easier to rewrite how we handle
    preload and preload_argv in configuration later

cc @jcrist I think that this might interest you for Dask-Gateway work

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -131,12 +131,10 @@ class Nanny(ServerNode):
         if self.preload_argv is None:
             self.preload_argv = dask.config.get("distributed.worker.preload-argv")
 
-        self.preload_nanny = preload_nanny
-        if self.preload_nanny is None:
-            self.preload_nanny = dask.config.get("distributed.nanny.preload")
-        self.preload_nanny_argv = preload_nanny_argv
-        if self.preload_nanny_argv is None:
-            self.preload_nanny_argv = dask.config.get("distributed.nanny.preload-argv")
+        if preload_nanny is None:
+            preload_nanny = dask.config.get("distributed.nanny.preload")
+        if preload_nanny_argv is None:
+            preload_nanny_argv = dask.config.get("distributed.nanny.preload-argv")
 
         self.Worker = Worker if worker_class is None else worker_class
         self.env = env or {}
@@ -168,8 +166,8 @@ class Nanny(ServerNode):
 
         self.local_directory = local_directory
 
-        self._preload_modules = preloading.on_creation(
-            self.preload_nanny, file_dir=self.local_directory
+        self.preloads = preloading.process_preloads(
+            self, preload_nanny, preload_nanny_argv, file_dir=self.local_directory
         )
 
         self.services = services
@@ -263,9 +261,8 @@ class Nanny(ServerNode):
         )
         self.ip = get_address_host(self.address)
 
-        await preloading.on_start(
-            self._preload_modules, self, argv=self.preload_nanny_argv,
-        )
+        for preload in self.preloads:
+            await preload.start()
 
         logger.info("        Start Nanny at: %r", self.address)
         response = await self.instantiate()
@@ -465,7 +462,8 @@ class Nanny(ServerNode):
         self.status = "closing"
         logger.info("Closing Nanny at %r", self.address)
 
-        await preloading.on_teardown(self._preload_modules, self)
+        for preload in self.preloads:
+            await preload.teardown()
 
         self.stop()
         try:
--- a/distributed/preloading.py
+++ b/distributed/preloading.py
@@ -3,10 +3,12 @@ import logging
 import os
 import shutil
 import sys
+from typing import List
 import filecmp
 from importlib import import_module
 
 import click
+from tornado.httpclient import AsyncHTTPClient
 
 from dask.utils import tmpfile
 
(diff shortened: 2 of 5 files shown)


distributed/tests/test_preload.py
```diff
@@ -121,3 +123,23 @@ async def test_preload_import_time(cleanup):
...
+""".strip()
+            )
+
+    app = web.Application([(r"/preload", MyHandler)])
+    server = app.listen(12345)
+    try:
+        async with Scheduler(preload=["http://localhost:12345/preload"]) as s:
+            assert s.foo == 1
+    finally:
+        server.stop()
```
mrocklin, line 145: This was the primary motivation for this PR

distributed/preloading.py
```diff
@@ -111,65 +131,59 @@ def _import_module(name, file_dir=None):
...
-    for name, interface in modules.items():
-        dask_setup = interface.get("dask_setup", None)
+
+    def __init__(self, dask_server, name, argv, file_dir):
+        self.dask_server = dask_server
+        self.name = name
+        self.argv = argv
+        self.file_dir = file_dir
+
+        self.module = _import_module(name, file_dir)
```
jcrist, line 153: I would move this to under `start`, no reason to do it here afaict.

distributed/tests/test_preload.py
jcrist: I actually don't see the motivation for this, can you expand? `exec`'ing code that is loaded from an http response seems like a risky thing to enable users to do - without proper security steps to prevent MITM attacks, this could load and exec a malicious script.

distributed/preloading.py
mrocklin: In between now and start lots of decisions are made.  For example, it's useful to import the module before we decide on protocol (this is an old request from the ucx folks, who want to `import ucx` to depend on the side effects of loading `"ucx://"` into the registry of protocols before we try to use that protocol.

mrocklin:
> I actually don't see the motivation for this, can you expand? exec'ing code that is loaded from an http response seems like a risky thing to enable users to do - without proper security steps to prevent MITM attacks, this could load and exec a malicious script.

At this point they're running preload scripts which run arbitrary Python code anyway.  If they want to avoid MITM attacks then they could use https?

distributed/preloading.py
jcrist: Then maybe set this to `None` in the case of a remote address? Creating and stashing coroutine that is later run seems like a bit convoluted logic that may be cleaner if the logic wasn't split between `__init__` and `start`. idk.

jcrist:
Right, but I'd argue many users probably won't bother, which exposes them. I think we don't want to enable poor security practices without good reason. Can you provide a motivating example for this feature?

mrocklin:
> Can you provide a motivating example for this feature?

Dask gateway could run user-provided docker images, but append `--preload http://gateway-address/scheduler-preload.py` to the command

mrocklin:
> If I understand correctly, the only user-facing change here is support for getting preload scripts from a remote endpoint? The class refactor is internal only?

Correct.  I started doing this in order to switch out the preload/preload_argv to a single preloads config value, but decided that was a bit too much for this PR.  I still think that it's a good move though.

distributed/preloading.py
mrocklin: Yeah, I agree that it's a little convoluted.  Worse is that behavior for web-downloaded preloads is different than those for file-loaded ones (we don't get on-import behavior for web-downloaded preloads).  It's a choice between this and poor testing though, and at least for now I'd prefer to optimize for testability.

In terms of how to handle the magic coroutine I don't really have a good answer here, other than to store state about if it's a web address, and to explicitly call this out in `def start`.  I'll do that now.  I guess we're being more explicit, which is good.

mrocklin: OK, we're now more explicit.  The behavior is the same, but things are hopefully a bit cleaner.

mrocklin:
@jcrist any thoughts on recent changes?  Are your concerns resolved (or resolved enough)?

mrocklin:
Merging tomorrow if there are no further comments.

distributed/nanny.py
```diff
@@ -263,9 +261,8 @@ async def start(self):
         )
         self.ip = get_address_host(self.address)
 
-        await preloading.on_start(
-            self._preload_modules, self, argv=self.preload_nanny_argv,
-        )
+        for preload in self.preloads:
+            await preload.start()
```
jcrist, line 265: Do you want these to run serially, or could the be done concurrently (e.g. using `asyncio.gather`)?

distributed/nanny.py
```diff
@@ -465,7 +462,8 @@ async def close(self, comm=None, timeout=5, report=None):
         self.status = "closing"
         logger.info("Closing Nanny at %r", self.address)
 
-        await preloading.on_teardown(self._preload_modules, self)
+        for preload in self.preloads:
+            await preload.teardown()
```
jcrist, line 466: Same here (and also in the worker file).

distributed/nanny.py
jcrist: If a call to `preload.start` fails, this will cause the scheduler startup to also fail, correct?

mrocklin: Making them sequential was intentional.  I figured it was more likely that someone would complain about preloads interacting poorly with each other than they would care about preloads being slow because they're not being run concurrently.

And yes, currently things will fail if a preload fails.  I think that I'm ok with that.

jcrist: > And yes, currently things will fail if a preload fails. I think that I'm ok with that.

I think that's the right behavior.

jcrist:
LGTM, merging.

mrocklin:
Thanks @jcrist !

PR #3757, opened 2020-04-30
Add prometheus metrics for semaphore

lr4d:
This introduces two new data structures at the scheduler side: `metrics`, and `pending_leases`.

For `http/semaphore/prometheus.py` I copied the `PrometheusHandler` and the basics of `_PrometheusCollector` from `http/scheduler/prometheus.py`, I'd be happy to change this if there is a more correct way of doing so.

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -1,15 +1,17 @@
-import uuid
-from collections import defaultdict, deque
 import asyncio
-import dask
+import logging
+import uuid
+import warnings
 from asyncio import TimeoutError
+from collections import defaultdict, deque
+
+import dask
 from tornado.ioloop import PeriodicCallback
+
+from distributed.utils_comm import retry_operation
+from .metrics import time
 from .utils import log_errors, parse_timedelta
 from .worker import get_client
-from .metrics import time
-import warnings
-import logging
-from distributed.utils_comm import retry_operation
 
 logger = logging.getLogger(__name__)
 
@@ -67,16 +69,24 @@ class SemaphoreExtension:
 
         self.scheduler.extensions["semaphores"] = self
 
+        # {metric_name: {semaphore_name: metric}}
+        self.metrics = {
+            "acquire_total": defaultdict(int),  # counter
+            "release_total": defaultdict(int),  # counter
+            "average_pending_lease_time": defaultdict(float),  # gauge
+            "pending": defaultdict(int),  # gauge
+        }
+
         validation_callback_time = parse_timedelta(
             dask.config.get("distributed.scheduler.locks.lease-validation-interval"),
             default="s",
         )
         self._pc_lease_timeout = PeriodicCallback(
-            self._check_lease_timeout, validation_callback_time * 1000,
+            self._check_lease_timeout, validation_callback_time * 1000
         )
         self._pc_lease_timeout.start()
         self.lease_timeout = parse_timedelta(
-            dask.config.get("distributed.scheduler.locks.lease-timeout"), default="s",
+            dask.config.get("distributed.scheduler.locks.lease-timeout"), default="s"
         )
 
     async def get_value(self, comm=None, name=None):
@@ -122,6 +132,7 @@ class SemaphoreExtension:
             now = time()
             logger.info("Acquire lease %s for %s at %s", lease_id, name, now)
             self.leases[name][lease_id] = now
+            self.metrics["acquire_total"][name] += 1
         else:
             result = False
         return result
@@ -141,6 +152,7 @@ class SemaphoreExtension:
             w = _Watch(timeout)
             w.start()
 
+            self.metrics["pending"][name] += 1
             while True:
                 logger.info(
                     "Trying to acquire %s for %s with %ss left.",
@@ -172,6 +184,12 @@ class SemaphoreExtension:
                     result,
                     w.elapsed(),
                 )
+                # We're about to return, so the lease is no longer "pending"
+                self.metrics["average_pending_lease_time"][name] = (
+                    self.metrics["average_pending_lease_time"][name] + w.elapsed()
+                ) / 2
+                self.metrics["pending"][name] -= 1
+
                 return result
 
     def release(self, comm=None, name=None, lease_id=None):
@@ -196,6 +214,7 @@ class SemaphoreExtension:
         # Everything needs to be atomic here.
         del self.leases[name][lease_id]
         self.events[name].set()
+        self.metrics["release_total"][name] += 1
 
     def _check_lease_timeout(self):
         now = time()
@@ -235,6 +254,16 @@ class SemaphoreExtension:
                         RuntimeWarning,
                     )
                 del self.leases[name]
+            if name in self.metrics["pending"]:
+                if self.metrics["pending"][name]:
+                    warnings.warn(
+                        f"Closing semaphore {name} but there remain pending leases",
+                        RuntimeWarning,
+                    )
+            # Clean-up state of semaphore metrics
+            for _, metric_dict in self.metrics.items():
+                if name in metric_dict:
+                    del metric_dict[name]
 
 
 class Semaphore:
@@ -394,7 +423,7 @@ class Semaphore:
     async def _acquire(self, timeout=None):
         lease_id = uuid.uuid4().hex
         logger.info(
-            "%s requests lease for %s with ID %s", self.client.id, self.name, lease_id,
+            "%s requests lease for %s with ID %s", self.client.id, self.name, lease_id
         )
 
         # Using a unique lease id generated here allows us to retry since the
@@ -433,7 +462,7 @@ class Semaphore:
         lease_id = self._leases.popleft()
         logger.info("%s releases %s for %s", self.client.id, lease_id, self.name)
         return self.client.sync(
-            self.client.scheduler.semaphore_release, name=self.name, lease_id=lease_id,
+            self.client.scheduler.semaphore_release, name=self.name, lease_id=lease_id
         )
 
     def get_value(self):
--- a/distributed/tests/test_semaphore.py
+++ b/distributed/tests/test_semaphore.py
@@ -1,9 +1,10 @@
+import asyncio
 import pickle
 import dask
 import pytest
 from dask.distributed import Client
-from time import sleep
-from distributed import Semaphore
+from time import time, sleep
+from distributed import Semaphore, fire_and_forget
 from distributed.comm import Comm
 from distributed.core import ConnectionPool
 from distributed.metrics import time
@@ -205,10 +206,26 @@ async def test_close_async(c, s, a, b):
     ):
         await sem.acquire()
 
+    sem2 = await Semaphore(name="t2", max_leases=1)
+    assert await sem2.acquire()
+
+    def f(sem_):
+        return sem_.acquire()
+
     semaphore_object = s.extensions["semaphores"]
+    fire_and_forget(c.submit(f, sem_=sem2))
+    while not semaphore_object.metrics["pending"]["t2"]:  # Wait for the pending lease
+        await asyncio.sleep(0.01)
+    with pytest.warns(
+        RuntimeWarning, match="Closing semaphore .* but there remain pending leases"
+    ):
+        await sem2.close()
+
     assert not semaphore_object.max_leases
     assert not semaphore_object.leases
     assert not semaphore_object.events
+    for metric_dict in semaphore_object.metrics.values():
+        assert not metric_dict
 
 
 def test_close_sync(client):
@@ -420,7 +437,7 @@ async def test_oversubscribing_leases(c, s, a, b):
     assert await sem.get_value() == 0
 
 
-@gen_cluster(client=True,)
+@gen_cluster(client=True)
 async def test_timeout_zero(c, s, a, b):
     # Depending on the internals a timeout zero cannot work, e.g. when the
     # initial try already includes a wait. Since some test cases use this, it is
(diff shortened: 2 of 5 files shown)


distributed/http/semaphore/prometheus.py
```diff
@@ -0,0 +1,143 @@
...
+        prometheus_client.REGISTRY.register(PrometheusHandler._collector)
+
+    def get(self):
+        import prometheus_client
+
+        self.write(prometheus_client.generate_latest())
+        self.set_header("Content-Type", "text/plain; version=0.0.4")
+
+
+routes = [("/metrics", PrometheusHandler, {})]
```
mrocklin, line 143: I think that this route is already taken by `distributed.http.scheduler.prometheus`.  Probably we should do some runtime checks to make sure that there aren't collisions like this.

lr4d: It's also in https://github.com/dask/distributed/blob/c8ee44bebe40e4db4495be1e2e0340fafbb41a37/distributed/http/worker/prometheus.py#L97. It seems to work fine when accessing the `/metrics`endpoint (I see the metrics from both files)

mrocklin: Those are never included in the same server.  The scheduler file is for the scheduler and the worker file is for the worker.   But presumably you want to include this route in one or the other of these two, and so there would be some conflict there.  When Prometheus comes by it will only be able to access one of these Handlers.

More generally though I like the idea of making our Prometheus metrics extensible.  I think that this particular approach isn't exactly the right way, but I would very much like to encourage you to think about how to resolve this problem.

lr4d: > I would very much like to encourage you to think about how to resolve this problem

Sure, I'll try to come up with something

distributed/http/scheduler/prometheus.py
```diff
@@ -2,6 +2,9 @@
 
 from ..utils import RequestHandler
 from ...scheduler import ALL_TASK_STATES
+from ..semaphore.prometheus import SemaphoreMetricExtension
+
+DEFAULT_METRIC_EXTENSIONS = (SemaphoreMetricExtension,)
```
lr4d, line 7: @mrocklin Opinions on this?

mrocklin: This could work.  Maybe we can make this slightly more generic and have something like ...


```python
from ..semaphore.prometheus import SemaphoreMetricExtension


class _PrometheusCollector:
    ...

collectors = [
    _PrometheusCollector, 
    SemaphoreMetricCollector,
]

...
```

This way everything is on the same baseline?

distributed/http/semaphore/prometheus.py
```diff
@@ -0,0 +1,112 @@
+import numpy as np
```
mrocklin, line 1: If this is the only file here then maybe it doesn't make sense to make this a separate directory.  Maybe it should live in http/scheduler/prometheus.py ?  Or maybe http/scheduler/prometheus_semaphore.py?  Or maybe we make a prometheus/ directory rather than a semaphore directory.  That seems more likely to see further activity in the future.

Although perhaps not, from my perspective semaphores aren't a big deal, but perhaps they are more important to your workloads than I understand.

mrocklin:
cc @fjetter for review in case he hasn't already looked at this one.

distributed/http/semaphore/prometheus.py
fjetter: Unfortunately we have some workloads which rely quite strongly on some sort of semaphore since they are quite entangled with a central payload DB and we can't allow the jobs to kill the DB.

So far, we've been using a semaphore in our internal/proprietary job scheduler which is currently in sunset and is being replaced by dask.distributed. The contributions to the semaphore here are essentially the condensation of our past experiences with the matter. Observation and resilience where a big deal for us in the past and we don't want to fall back to old error patterns.

These metrics will most likely wrap up our immediate concerns about the semaphore and I don't expect any major changes or additions incoming after this; assuming nothing breaks :). In particular, I don't see any additions incoming w.r.t the `http` module.

Wrapping up, I like the suggestion of putting this into a `http/scheduler/prometheus` directory since this fits best with the notion of extensions

distributed/semaphore.py
```diff
@@ -141,6 +153,10 @@ async def acquire(self, comm=None, name=None, timeout=None, lease_id=None):
             w = _Watch(timeout)
             w.start()
 
+            # We create a new uuid in case of multiple acquire retries with the same `lease_id`
```
fjetter, line 156: I would actually not introduce a new UUID here. On the one side it introduces even more complexity to the setup and on the other side the retries are more or less an implementation detail we should not necessarily expose via the prometheus metrics. Best case, we'd expose a counter of how often something was retried but I'm not sure if this is valuable since I'd expect this to be an edge case and we have logging info for this as well

distributed/semaphore.py
```diff
@@ -172,6 +188,16 @@ async def acquire(self, comm=None, name=None, timeout=None, lease_id=None):
                     result,
                     w.elapsed(),
                 )
+
+                time_to_acquire_lease = (
+                    time() - self.pending_leases[name][pending_lease_id]
```
fjetter, line 193: I'd like to keep the code instrumentation as close as possible to the actual code it is supposed to observe. In particular this timing is mostly relevant in connection to the timeouts and the timeouts use the watch `w`. I'd prefer if we dont introduce another time measurement and just use `w.elapsed()` as for the log above

distributed/semaphore.py
```diff
@@ -67,16 +71,23 @@ def __init__(self, scheduler):
 
         self.scheduler.extensions["semaphores"] = self
 
+        # {metric_name: {semaphore_name: metric}}
+        self.metrics = {
+            "acquire_total": defaultdict(int),  # counter
+            "release_total": defaultdict(int),  # counter
+            "time_to_acquire_lease": defaultdict(list),  # summary-like
```
fjetter, line 78: this list is unbound and might pose a problem. I suggest to either use a limited `deque` or even better, use a `tdigest` since essentially we're interested in a histogram, aren't we?

distributed/http/semaphore/prometheus.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class SemaphoreMetricExtension:
+    def __init__(self, dask_server):
+        self.server = dask_server
+
+    def collect(self):
+        from prometheus_client.core import (
```
fjetter, line 9: why the local import?

distributed/http/semaphore/prometheus.py
```diff
@@ -0,0 +1,112 @@
...
+            "Total number of leases released per semaphore.\n"
+            "Note: if a semaphore is closed while there are still leases active, this count will not equal "
+            "`semaphore_acquired_total` after execution.",
+            labels=["name"],
+        )
+        for semaphore_name, count in sem_ext.metrics["release_total"].items():
+            semaphore_release_total.add_metric([semaphore_name], count)
+        yield semaphore_release_total
+
+        semaphore_time_to_acquire_lease = HistogramMetricFamily(
```
fjetter, line 69: I don't really have a good feeling about what this histogram values will actually look like and I would assume the size of the buckets is strongly depending on the applications running. For instance, our applications using a semaphore usually query data from a DB where these jobs usually take 10-20s or longer, including data submissions, etc. This would all fall into the max bin already. OTOH, I can see valid use cases in a much smaller range of 1-5s. For even smaller lease times I'm not even sure if this semaphore is the best fit.

I think a more stable / independent approach would be to expose some measure or quantile of the distribution. I'd start with min,max,mean,median or similar. That could be exposed in simple gauges, like `samphore_time_to_acquire_min`, etc.

distributed/http/scheduler/prometheus.py
```diff
@@ -68,6 +69,9 @@ def collect(self):
         yield tasks
 
 
+COLLECTORS = [_PrometheusCollector, SemaphoreMetricExtension]
```
fjetter, line 72: I like the approach since it's simple and very similar to how the scheduler extensions are defined

lr4d: yeah I got my inspiration from the scheduler extensions

distributed/http/semaphore/prometheus.py
lr4d: Same as in `http/scheduler/prometheus.py` (and `worker`). I assumed there must be some reason for having it there. Perhaps somebody else can shine some light on that

distributed/semaphore.py
lr4d: Yes, we are. Thanks for the suggestion. Will change this accordingly

distributed/semaphore.py
lr4d: - [ ] substitute this for `w.elapsed()`

distributed/semaphore.py
lr4d: My rationale was that when this edge case occurs, we don't want this metric to say that a pending_lease is pending for a shorter amount of time than is the actual case. To clarify, you consider this too improbable to be worth implementing?

distributed/http/semaphore/prometheus.py
lr4d: One alternative would be to change the `HistogramMetricFamily` to a `Summary` (https://prometheus.io/docs/concepts/metric_types/#summary) and then add `min`,`max`, `median`. What do you think?

distributed/http/semaphore/prometheus.py
lr4d: - [ ] move this under `http/scheduler/`

distributed/http/semaphore/prometheus.py
lr4d: Ignore my last comment. I took a look at the prometheus_client code and there is no quantile information when using `SummaryMetricFamily` (which is the custom metric we would use instead of `Summary` in this case).
However, one can generate such quantiles with the current implementation by using https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile on the server/dashboard that is processing the metrics.

fjetter: You'd still need to define explicit buckets and we cannot do this here without very strong assumptions

distributed/semaphore.py
fjetter: Setting consistency and complexity arguments aside for a moment, I would argue that you can achieve an identical behaviour with a simple counter

```
self.metrics["time_to_acquire_lease"]["pending_leases"] += 1

# do stuff

self.metrics["time_to_acquire_lease"]["pending_leases"] -= 1
self.metrics["time_to_acquire_lease"][name].update(w.elapsed())
```

distributed/http/semaphore/prometheus.py
fjetter: probably about having prometheus as an optional dependency. That's fine I guess.

distributed/tests/test_semaphore.py
```diff
@@ -443,3 +460,33 @@ async def test_getvalue(c, s, a, b):
...
+    sem_ext = s.extensions["semaphores"]
+
+    actual = sem_ext.metrics.copy()
+    assert (
+        expected_average_pending_lease_time - epsilon
+        <= actual.pop("average_pending_lease_time")["test"]
+        <= expected_average_pending_lease_time + epsilon
+    )
+    expected = {
+        "acquire_total": defaultdict(int, {"test": 2}),
```
fjetter, line 488: Is the metrics dict cleaned up once the semaphore is closed?

distributed/tests/test_semaphore.py
```diff
@@ -205,10 +206,26 @@ async def test_close_async(c, s, a, b):
...
+        await asyncio.sleep(0.01)
+    with pytest.warns(
+        RuntimeWarning, match="Closing semaphore .* but there remain pending leases"
+    ):
+        await sem2.close()
+
     assert not semaphore_object.max_leases
     assert not semaphore_object.leases
     assert not semaphore_object.events
+    for metric_dict in semaphore_object.metrics.values():
```
lr4d, line 227: @fjetter regarding clean-up, yes, see this test or `SemaphoreExtension.close`

distributed/semaphore.py
```diff
@@ -99,6 +99,10 @@ def create(self, comm=None, name=None, max_leases=None):
         if name not in self.max_leases:
             assert isinstance(max_leases, int), max_leases
             self.max_leases[name] = max_leases
+            # Initialize metrics
```
lr4d, line 102: I just added this to make sure all metrics are set when the semaphore is created. @fjetter I could also move this to a new PR to not block the commits that you have already reviewed

fjetter: This logic somehow defeats the purpose of using defaultdicts, doesn't it? If we initialize these data structure here explicitly, we don't need defaultdicts.
This feels rather like a shortcoming of the prometheus collector and I don't understand why this is needed. If the collector calls `len(self.leases[name])` or something similar this should be properly handled already, shouldn't it?

lr4d: Sure, I'll move this logic to the prometheus collector. I'll remove the commit for now

mrocklin:
The module structure looks fine to me.  I probably won't dive into the technical changes here, but please feel free to merge if you both have confidence.

fjetter:
@mrocklin We receive continuously `OSError: [Errno 98] Address already in use` errors and I don't think this is connected. What's the process here? Open a ticket and merge anyhow or should we first troubleshoot this?

mrocklin:
> We receive continuously OSError: [Errno 98] Address already in use errors and I don't think this is connected. What's the process here? Open a ticket and merge anyhow or should we first troubleshoot this?

If you are confident that it is not related then yes, you can feel free to open a ticket and merge anyway.

PR #3758, opened 2020-04-30
Connect to dashboard when address provided

TomAugspurger:
It seems like if the caller provides a `dashboard_address` then they
want the dashboard running. I'm not sure why `dashboard` is required,
looking for guidance on that if you recall the reason for the keyword @mrocklin.

Closes https://github.com/dask/dask-gateway/issues/261

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1133,14 +1133,14 @@ class Scheduler(ServerNode):
         )
         self.start_http_server(routes, dashboard_address, default_port=8787)
 
-        if dashboard:
+        if dashboard or (dashboard is None and dashboard_address):
             try:
                 import distributed.dashboard.scheduler
             except ImportError:
                 logger.debug("To start diagnostics web server please install Bokeh")
             else:
                 distributed.dashboard.scheduler.connect(
-                    self.http_application, self.http_server, self, prefix=http_prefix,
+                    self.http_application, self.http_server, self, prefix=http_prefix
                 )
 
         # Communication state


mrocklin:
Bad naming on my end, due to a desire to keep things backwards compatible.

Dashboard address refers to the address used by the http server, which is
active regardless.  The dashboard keyword might by more precisely "bokeh",
as it turns on the bokeh part of the dashboard.  It is reasonable for
people to want the http server running but not the bokeh dashboard.

TomAugspurger:
Mmm, OK. Do you think that the check `(dashboard is None and dashboard_address)` is OK for enabling the bokeh dashboard? I think that captures the previous behavior.

Users wishing to disable it can specify `dashboard=False`. This will make `Scheduler` match the behavior of `dask-scheduler` where the default is to enable the dashboard.

mrocklin:
Sure.  I think that our default for the workers was not to enable it,
although I don't recall exactly.

TomAugspurger:
@jrbourbeau have you seen `pytest distributed/tests/test_client.py::test_repr_localcluster` fail? It failed at https://travis-ci.org/github/dask/distributed/jobs/681652905#L1562.

I can't reproduce locally, but it did fail in `start_http_server` so potentially related to this change. Will rerun it a few times.

jrbourbeau:
Hmm, no I haven't seen that fail before (also wasn't able to reproduce locally). Do you recall what the failure was?

TomAugspurger:
Sorry, I thought I pasted it. I don't recall the details.

TomAugspurger:
I'm also happy to do the release. Up to you.

jrbourbeau:
Okay, sounds good. I'll go ahead and merge this in and do a `2.15.2` release

jrbourbeau:
Just pushed `2.15.2` to PyPI, should be on conda-forge in an hour-ish

TomAugspurger:
Thanks!

PR #3760, opened 2020-04-30
[WIP] Reduce number of loop computations

galipremsagar:
This PR tries to reduce the number of times we loop on the same objects and also tries to squeeze into as few loops as possible. This is a first pass of the chanes, hence this is still WIP.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2498,15 +2498,26 @@ class Client(Node):
             if values:
                 dsk = subs_multiple(dsk, values)
 
-            d = {k: unpack_remotedata(v, byte_keys=True) for k, v in dsk.items()}
-            extra_futures = set.union(*[v[1] for v in d.values()]) if d else set()
-            extra_keys = {tokey(future.key) for future in extra_futures}
-            dsk2 = str_graph({k: v[0] for k, v in d.items()}, extra_keys)
-            dsk3 = {k: v for k, v in dsk2.items() if k is not v}
+            d = {}
+            extra_futures_values = []
+            dsk2_dict = {}
+            dependencies = {}
+            for k, v in dsk.items():
+                d[k] = unpack_remotedata(v, byte_keys=True)
+                extra_futures_values.append(d[k][1])
+                dsk2_dict[k] = d[k][0]
+                dependencies[k] = get_dependencies(dsk, k)
+
+            extra_futures = set.union(*extra_futures_values) if d else set()
+            extra_keys = set()
             for future in extra_futures:
                 if future.client is not self:
                     msg = "Inputs contain futures that were created by another client."
                     raise ValueError(msg)
+                extra_keys.add(tokey(future.key))
+            dsk2 = str_graph(dsk2_dict, extra_keys)
+            dsk3 = {k: v for k, v in dsk2.items() if k is not v}
+            
 
             if restrictions:
                 restrictions = keymap(tokey, restrictions)
@@ -2514,17 +2525,6 @@ class Client(Node):
 
             if loose_restrictions is not None:
                 loose_restrictions = list(map(tokey, loose_restrictions))
-
-            future_dependencies = {
-                tokey(k): {tokey(f.key) for f in v[1]} for k, v in d.items()
-            }
-
-            for s in future_dependencies.values():
-                for v in s:
-                    if v not in self.futures:
-                        raise CancelledError(v)
-
-            dependencies = {k: get_dependencies(dsk, k) for k in dsk}
 
             if priority is None:
                 priority = dask.order.order(dsk, dependencies=dependencies)
@@ -2534,9 +2534,15 @@ class Client(Node):
                 tokey(k): [tokey(dep) for dep in deps]
                 for k, deps in dependencies.items()
             }
-            for k, deps in future_dependencies.items():
-                if deps:
-                    dependencies[k] = list(set(dependencies.get(k, ())) | deps)
+
+            for key, value in d.items():
+                _key = tokey(key)
+                _value = {tokey(f.key) for f in value[1]}
+                for v in _value:
+                    if v not in self.futures:
+                        raise CancelledError(v)
+                if _value:
+                    dependencies[_key] = list(set(dependencies.get(_key, ())) | _value)
 
             if isinstance(retries, Number) and retries > 0:
                 retries = {k: retries for k in dsk3}
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2517,7 +2517,6 @@ class Client(Node):
                 extra_keys.add(tokey(future.key))
             dsk2 = str_graph(dsk2_dict, extra_keys)
             dsk3 = {k: v for k, v in dsk2.items() if k is not v}
-            
 
             if restrictions:
                 restrictions = keymap(tokey, restrictions)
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -1823,17 +1823,21 @@ def update_graph(
...
             if k in deps:
                 deps.remove(k)
             dependencies[k] = deps
 
-        # Avoid computation that is already finished
-        already_in_memory = set()  # tasks that are already done
-        for k, v in dependencies.items():
-            if v and k in self.tasks and self.tasks[k].state in ("memory", "erred"):
+            if (
+                dependencies[k]
```
kkraus14, line 1837: You already have this as `v` from the loop I believe.
```suggestion
                v
```

kkraus14, line 1837: Since you're assigning it above we can avoid the extra `__getitem__` call here into the dict.
```suggestion
                deps
```

mrocklin:
Thank you for working on this @galipremsagar .  At some point it will be useful to see if this has a significant impact on timing in real-world applications.

kkraus14:
@galipremsagar it would be good to test with the example here https://github.com/dask/dask/issues/6163#issuecomment-622571290 and post results to see if this is making any meaningful improvements.

galipremsagar:
> @galipremsagar it would be good to test with the example here [dask/dask#6163 (comment)](https://github.com/dask/dask/issues/6163#issuecomment-622571290) and post results to see if this is making any meaningful improvements.

For the sample you pointed to the improvement in performance is about 0.3s. But as the number of records is increased we see 1s improvement.

```python
from dask.datasets import timeseries
from dask.dataframe.shuffle import shuffle


client = Client()

ddf_d = timeseries(start='2000-01-01', end='2015-01-01', partition_freq='1d')
ddf_d_2 = shuffle(ddf_d, "id", shuffle="tasks")

%time ddf_d_2 = ddf_d_2.persist()
```

**Master**
CPU times: user 45.5 s, sys: 1.07 s, total: 46.6 s
Wall time: 46.2 s

**This PR**
CPU times: user 44.2 s, sys: 1.46 s, total: 45.6 s
Wall time: 45.2 s

kkraus14:
How many times have you run this? ~1s could just be noise 😄

distributed/client.py
```diff
@@ -1621,10 +1621,7 @@ def map(
         total_length = sum(len(x) for x in iterables)
 
         if batch_size and batch_size > 1 and total_length > batch_size:
-            # PREM
-            batches = list(
-                zip(*[partition_all(batch_size, iterable) for iterable in iterables])
-            )
+            batches = zip(*[partition_all(batch_size, iterable) for iterable in iterables])
```
kkraus14: This creates a temporary list just to unpack it? Could we just create a zip iterator from a generator expression here to avoid materializing an intermediate list unnecessarily?

distributed/client.py
```diff
@@ -3319,7 +3316,7 @@ def call_stack(self, futures=None, keys=None):
         keys = keys or []
         if futures is not None:
             futures = self.futures_of(futures)
-            keys += list(map(tokey, {f.key for f in futures}))
+            keys.extend(list(map(tokey, {f.key for f in futures})))
```
kkraus14, line 3321: You should be able to just pass an iterable here so you shouldn't have to wrap what's passed to `.extend(...)` in a list.

distributed/client.py
galipremsagar: @kkraus14 So the actual materialization is removed with the removal of the outer `list`. Now there is no materialization with the usage of inner `[]`, we will need it to create `partition_all` objects list(like.. `[partition_all, partition_all,...]`)

Also, there seems to be a need for `zip` in this particular logic as to me it looks like the iterables are being expanded into repetitions.

kkraus14: You're creating an intermediate list via list comprehension in `[partition_all(batch_size, iterable) for iterable in iterables]` which you then unpack into the zip iterator. I'm saying you should be able to just use a generator to avoid materializing that list:
`zip(*(partition_all(batch_size, iterable) for iterable in iterables))`

mrocklin: I think that we may be at the point of diminishing returns here.  If this isn't happening within a tight for loop then it probably doesn't matter.

There is also some debuggability cost to using generators over lists.  The Python debugger doesn't descend as well well when we unpack generators than when we unpack lists.

mrocklin:
> How many times have you run this? ~1s could just be noise smile

This would be good to know

It looks like things have stalled here.  Is this still a priority?

galipremsagar:
> > How many times have you run this? ~1s could just be noise smile
> 
> This would be good to know
> 
> It looks like things have stalled here. Is this still a priority?

@mrocklin @kkraus14 

From the benchmark numbers it looks more or less the same against master:

```python
   ...: from dask.datasets import timeseries 
   ...: from dask.dataframe.shuffle import shuffle 
   ...:  
   ...:  
   ...: client = Client() 
   ...:  
   ...: ddf_d = timeseries(start='2000-01-01', end='2015-01-01', partition_freq='1d') 
   ...: ddf_d_2 = shuffle(ddf_d, "id", shuffle="tasks")   
```

Master
```python
In [2]: %timeit ddf_d_2.persist()                                                                   
34.8 s ± 2.61 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This PR
```python
In [2]: %timeit ddf_d_2.persist()                                                                   
34.5 s ± 2.72 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

quasiben:
@galipremsagar does this PR reduce temporary memory usage ? If not, would you consider these style changes then given similar performance ?

mrocklin:
It sounds like we're looking for a reason to continue this work. It's also totally ok to not continue it, and focus efforts elsewhere.  My guess is that there are plenty of things to work on when it comes to accelerating the scheduler.

galipremsagar:
It reduces memory usages, here is a memory profile. Not a significant reduction but similar noticeable difference across runs.

This PR
```
In [2]: %memit ddf_d_2.persist()                                                              
peak memory: 1383.89 MiB, increment: 1029.93 MiB
```
Master
```
In [2]: %memit ddf_d_2.persist()                                                              
peak memory: 1438.21 MiB, increment: 1083.59 MiB
```

quasiben:
> It sounds like we're looking for a reason to continue this work. It's also totally ok to not continue it, and focus efforts elsewhere. My guess is that there are plenty of things to work on when it comes to accelerating the scheduler.

Agreed.  I don't think we should push on this PR anymore.  There is some gain but not a whole lot.  Due to the large number of changes perhaps we should abandon. 

If others feel strongly that this should be merged in they should feel free to comment.

kkraus14:
I will defer to @galipremsagar's opinion here, but I am in favor of abandoning this.

galipremsagar:
Given that these are micro-optimizations, here are both ways to it.

We can abandon this PR as the speed-ups are not noticeable and changes are more.
We can merge this PR based on the minor reduction in memory foot-print reported above.

jacobtomlinson:
Just to weigh in here I think many of the changes are style changes, some make things more readable, but others make things less readable. I would actually argue that small increases in contributor/maintainer performance with readable code is more important than small increases in code performance.

However, a ~4% reduction in memory usage is not insignificant. (Also TIL about `%memit` so thanks!).

Therefore I would probably go either way with this. And generally in OSS maintenance if I'm on the fence about a change like this I tend to merge it, mainly for the community benefit of getting another developer with changes in the project.

Issue #2710, opened 2019-05-18
System page on dashboard regularly awkward


birdsarah:
I often end up in an aspect ratio where the system page of the dashboard looks like this:

![Screenshot from 2019-05-18 12-28-41](https://user-images.githubusercontent.com/1796208/57973071-a139f300-7968-11e9-92d4-0705c697a22b.png)

Bokeh's layout has had a big overhaul of late, maybe there's an opportunity to take advantage and get a more robust layout with less code. I would suggest a grid of four plots should always be fine.

mrocklin:
cc @mathdugre , who has been looking into Bokeh plotting issues recently.  (though it looks like this one might be more serious than the previous two).

mathdugre:
Hi @mrocklin, I would definitely like to work on this. Will give it a shot during the week :)

PR #3764, opened 2020-05-04
Fix System page on dashboard

prasunanand:
* Fixes https://github.com/dask/distributed/issues/2710
* Modified the flex property of `#system-fluid` and `.system-item`

Attached screenshot
Screenshot from 2020-05-04 19-18-22

### DIFF
--- a/distributed/http/static/css/system.css
+++ b/distributed/http/static/css/system.css
@@ -18,9 +18,9 @@
  * half of the width */
 @media (min-width: 992px) {
     #system-fluid {
-        flex-direction: row;
+        flex-direction: column;
     }
     #system-fluid .system-item {
-        flex: 1 50%;
+        flex: 1;
     }
 }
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @dsludwig in case he has time

mrocklin:
Or @jsignell perhaps?  This should be a quick review if you're around and are comfortable with CSS (I am not unfortunately)

jsignell:
Thanks for the ping; I will take a look.

jsignell:
An alternate approach that is more similar to the proposal in the original issue is at #3781.

prasunanand:
Thanks @jsignell . Made the changes.

jsignell:
Thank you for making the changes @prasunanand. Since you have brought up this fix I went back to the original issue and I think an even better solution is to use bokeh layouts and not write our own css for this view at all. I implemented that in #3781.

PR #3766, opened 2020-05-04
Fix TypeError in broadcast when using TLS

fjetter:
This fixes an issue where the broadcast fails if using TLS with

```
`TypeError: TLS expects a `ssl_context` argument of type ssl.SSLContext (perhaps check your TLS configuration?)  Instead got None`
```

The issue is that the connect passes the connection_args as one argument but it should pass it as kwargs. While fixing this, I realised that this is the only place where we connect without using the connection pool. I opted to use the connection pool instead but if there is a reason for having dedicated connections for the broadcast, I can revert this, of course.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -45,7 +45,7 @@ from .comm import (
     unparse_host_port,
 )
 from .comm.addressing import addresses_from_user_args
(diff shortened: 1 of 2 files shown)


mrocklin:
+1  .  Thanks @fjetter

PR #3768, opened 2020-05-04
Don't validate http preloads locally

necaris:
`validate_preload_argv` can't validate web address preloads for
`click.Command` compliance, so it shouldn't try.

### DIFF
--- a/distributed/cli/tests/test_dask_scheduler.py
+++ b/distributed/cli/tests/test_dask_scheduler.py
@@ -299,6 +299,36 @@ def test_preload_module(loop):
         shutil.rmtree(tmpdir)
 
 
+def test_preload_remote_module(loop):
+    def check_scheduler():
+        import scheduler_info
+
+        return scheduler_info.get_scheduler_address()
+
+    tmpdir = tempfile.mkdtemp()
+    try:
+        path = os.path.join(tmpdir, "scheduler_info.py")
+        with open(path, "w") as f:
+            f.write(PRELOAD_TEXT)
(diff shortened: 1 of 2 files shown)


necaris:
@jrbourbeau yes, that's basically the issue: because `validate_preload_argv` is run as a callback when `click` parses command-line arguments, any of the other processing hasn't run yet.

I _presume_ that this could be tested by something along the lines of https://github.com/dask/distributed/blob/master/distributed/cli/tests/test_dask_scheduler.py#L269 -- does that seem like the right approach?

mrocklin:
> is the issue here that at the time the validation code is run, we haven't yet downloaded and parsed the remote module into a ModuleType object yet?

Yes.  We're using tornado's AsyncHTTPClient to download the module, and at the point where we validate we haven't yet spun up the event loop, and we're not in an asynchronous context.  There are other inconsistencies with web-based preloads.  In particular we don't import the module at setup time, but rather at start time.  

This is more-or-less necessary if we want to make it easy to test web-based preloads in an asynchronous context (which I think that we do).  I think that testability is more important here than the feature mismatch.  I'm inclined to wait for a user to arrive before we bend too much to make this fully consistent.

necaris:
@jrbourbeau @mrocklin added https://github.com/dask/distributed/pull/3768/commits/5b8c52692d485ac0cef3536bce5d560bc1a4d527 -- please let me know if that's a sensible testing approach.

mrocklin:
I've modified it a bit.  I'm +1 on this and will plan to merge tomorrow if there are no objections.

PR #3780, opened 2020-05-06
Add missing options to dask-ssh cli

### LINKED ISSUE #3763: Missing options in `dask-ssh` cli
Currently, `dask-ssh` can handle forwarding only 16 options to `dask-scheduler` and `dask-worker` commands. Other options, supported by these two commands, cannot be passed through.

Having `dask-ssh` to support dynamically all options from `dask-scheduler` and `dask-worker` can be very handy in dask deployments with `old_ssh`. 

If this is OK with you, I can work around with `Click` and propose a PR for this.

abduhbm:
Closes #3763

This PR add missing options from `dask-scheduler` and `dask-worker` commands to `dask-ssh`.

Example:
<pre>
$ dask-ssh --nprocs 2 --nthreads 1 --worker-port 9000:9010  --hostfile ../nodes \
<b>--scheduler-dashboard-address :8989</b> <b>--scheduler-preload /path/to/foo.py</b>
</pre>

### DIFF
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -49,15 +49,21 @@ worker_options = [
     ),
     click.Option(
         ["--worker-port"],
-        type=int,
-        default=0,
-        help="Serving computation port, defaults to random",
+        default=None,
+        help="Serving computation port, defaults to random. "
+        "When creating multiple workers with --nprocs, a sequential range of "
+        "worker ports may be used by specifying the first and last available "
+        "ports like <first-port>:<last-port>. For example, --worker-port=3000:3026 "
+        "will use ports 3000, 3001, ..., 3025, 3026.",
     ),
     click.Option(
         ["--nanny-port"],
-        type=int,
-        default=0,
-        help="Serving nanny port, defaults to random",
+        default=None,
+        help="Serving nanny port, defaults to random. "
+        "When creating multiple nannies with --nprocs, a sequential range of "
+        "nanny ports may be used by specifying the first and last available "
+        "ports like <first-port>:<last-port>. For example, --nanny-port=3000:3026 "
+        "will use ports 3000, 3001, ..., 3025, 3026.",
     ),
     click.Option(
         ["--bokeh-port"],
--- a/distributed/cli/dask_worker.py
+++ b/distributed/cli/dask_worker.py
@@ -26,205 +26,231 @@ from tornado.ioloop import IOLoop, TimeoutError
 
 logger = logging.getLogger("distributed.dask_worker")
 
-
 pem_file_option_type = click.Path(exists=True, resolve_path=True)
 
+worker_options = [
+    click.Option(
+        ["--tls-ca-file"],
+        type=pem_file_option_type,
+        default=None,
+        help="CA cert(s) file for TLS (in PEM format)",
+    ),
+    click.Option(
+        ["--tls-cert"],
+        type=pem_file_option_type,
+        default=None,
+        help="certificate file for TLS (in PEM format)",
+    ),
+    click.Option(
+        ["--tls-key"],
+        type=pem_file_option_type,
+        default=None,
+        help="private key file for TLS (in PEM format)",
+    ),
+    click.Option(
+        ["--worker-port"],
+        type=int,
+        default=0,
+        help="Serving computation port, defaults to random",
+    ),
+    click.Option(
+        ["--nanny-port"],
+        type=int,
+        default=0,
+        help="Serving nanny port, defaults to random",
+    ),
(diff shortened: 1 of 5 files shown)


distributed/cli/dask_worker.py
```diff
@@ -26,214 +26,237 @@
...
+    click.Option(
+        ["--contact-address"],
+        type=str,
+        default=None,
+        help="The address the worker advertises to the scheduler for "
+        "communication with it and other workers. "
+        "Example: tcp://127.0.0.1:9000",
+    ),
+    click.Option(
+        ["--host"],
```
jcrist, line 108: Not all worker arguments make sense to add to dask-ssh (where they're used as a single flag to provide to all workers). `--host` for example is likely to differ between workers. Same with `--name` below, and likely a few others.

distributed/cli/dask_worker.py
```diff
@@ -26,214 +26,237 @@
...
+        "When creating multiple nannies with --nprocs, a sequential range of "
+        "nanny ports may be used by specifying the first and last available "
+        "ports like <first-port>:<last-port>. For example, --nanny-port=3000:3026 "
+        "will use ports 3000, 3001, ..., 3025, 3026.",
+    ),
+    click.Option(
+        ["--bokeh-port"],
+        type=int,
+        default=None,
+        help="Deprecated.  See --dashboard-address",
```
jcrist, line 72: We probably don't need to add deprecated options to `dask-ssh` (since they didn't exist on that command before).

abduhbm: Add the following to ignore options with `deprecated` in their help:
```python
    if hasattr(opt, "help") and "deprecated" in opt.help.lower():
        continue
```

distributed/cli/dask_worker.py
abduhbm: Added `"pid_file", "host", "name"` in the `ignored_options` list. 
Any others that I missed?

abduhbm:
Thanks @jcrist for your feedback. 

`test_bokeh_deprecation` keeps failing in CI checks. It passes on my local setup. Any idea why?

jcrist:
> test_bokeh_deprecation keeps failing in CI checks. It passes on my local setup. Any idea why?

Not sure, but since you've changed the CLI here I'd expect this to be due to your changes. Usually when I run into a test failing I do the following:

- I find the name of the test, and use `py.test` locally run just that test:

```python
# The `-k` flag tells pytest to run just that test
$ py.test distributed/cli/tests/test_dask_worker.py -k test_bokeh_deprecation
```

- If looking at the failure doesn't help, I might also run py.test with `--pdb`, to drop into a debugger on failure

```python
# The `-k` flag tells pytest to run just that test
$ py.test distributed/cli/tests/test_dask_worker.py -k test_bokeh_deprecation --pdb
```

- If that still isn't useful, I might make a separate script (or in this case CLI command) to reproduce and iterate from there. If needed, I might add a `breakpoint()` call somewhere in the code where I want to stop and inspect what's going on. In this case, it looks like the `--bokeh-port` and `--bokeh` flags are possibly no longer setting the `bokeh_port` and `bokeh` parameters properly. You might start looking here: https://github.com/dask/distributed/blob/master/distributed/cli/dask_worker.py#L269-L279 aren't being forwarded properly.

If after looking through this you're still having trouble debugging, let me know and I'll take a closer look.

abduhbm:
@jcrist it seems that it was an import order issue for Click options. `pytest` runs `test_dask_ssh` before `test_dask_worker` and will import and register Click Options for `distributed.cli.dask_ssh.main` before `distributed.cli.dask_worker.main`.

I modified some cli tests to run them using `subprocess.Popen` instead of `CliRunner.invoke` to avoid this import issue.

distributed/cli/utils.py
```diff
@@ -64,3 +65,49 @@ async def cleanup_and_stop():
 
     for sig in [signal.SIGINT, signal.SIGTERM]:
         old_handlers[sig] = signal.signal(sig, handle_signal)
+
+
+def prepare_dask_ssh_options(entity, opt):
```
jcrist, line 70: Oh, I missed that you were mutating the options list here. Mutating globals is generally bad practice, instead I recommend copying (or recreating) the options here. You might make use of `copy.deepcopy`, or there may be a better way.

```python
opt = copy.deepcopy(opt)
# you can now mutate `opt`, as it's a unique copy.
```

abduhbm: Thanks @jcrist for your recommendation. I reverted the test changes.

abduhbm:
@jcrist please let me know if anything else is needed here. Thanks,

PR #3781, opened 2020-05-06
Use bokeh column for /system instead of custom css

jsignell:
Closes: #2710
image

Please weigh in if you think the columns should stacked in a single column rather than in a grid.


mrocklin:
The dashboard is often used in an orientation where it takes up half of the screen.  In this setting a stacked column is nice.  But when it's run full-screen this can be a bit much.  About a year ago @dsludwig tried using CSS Grid to make things switch around based on aspect ratio I think.

jsignell:
Ah, thanks for the context. If switching between column and grid based on screen size is still desirable then I don't think this can be solved from the bokeh side. Personally I think that always stacked or always grid will be less brittle.

mrocklin:
Between the two, always stacked seems better to me for time series.  This way you can easily compare between the different plots.  Also we tend not to need too much precision in the y-axis in order to make decisions.

jsignell:
It is surprising to me how much I prefer the bokeh stacking - which aligns the axes - to the flex stacking that we had before. 

image

mrocklin:
> It is surprising to me how much I prefer the bokeh stacking - which aligns the axes - to the flex stacking that we had before.

I'm entirely happy to defer the question of using this solution or the other to you.

jsignell:
Ok I think this is the better approach.

jsignell:
This is ready to merge.

mrocklin:
Thanks @jsignell !

PR #3782, opened 2020-05-06
Remove dev-requirements since it is unused

jsignell:
AFAICT dev-requirements.txt is not used in any of docs or CI. So while it is a theoretically useful thing, it is likely to drift out of date. I propose removing it.


jcrist:
Makes sense to me. Will merge later today if no push-back.

PR #3784, opened 2020-05-07
Support Pickle's protocol 5

jakirkham:
This ensures that objects that do support Pickle's protocol 5 go through out-of-band serialization of buffers. Should avoid some extra memory usage and memory copies that would otherwise be encountered when going through Pickle's older protocols without this feature. Requires Python 3.8+ to work (as that is when this protocol was introduced).

Had considered supporting `pickle5` as well to ensure users could get this functionality with earlier versions of Python. However this would involve supporting `pickle5` in `cloudpickle`, which hasn't been done yet ( https://github.com/cloudpipe/cloudpickle/issues/179 ). So am side stepping that for now. Could be handled in a follow-up once that is addressed.

### DIFF
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -24,7 +24,7 @@ def _always_use_pickle_for(x):
         return False
 
 
-def dumps(x):
+def dumps(x, *, buffer_callback=None):
     """ Manage between cloudpickle and pickle
 
     1.  Try pickle
@@ -36,25 +36,25 @@ def dumps(x):
         result = pickle.dumps(x, **dump_kwargs)
         if len(result) < 1000:
             if b"__main__" in result:
-                return cloudpickle.dumps(x, **dump_kwargs)
-            else:
-                return result
+                result = cloudpickle.dumps(x, **dump_kwargs)
         else:
-            if _always_use_pickle_for(x) or b"__main__" not in result:
-                return result
-            else:
-                return cloudpickle.dumps(x, **dump_kwargs)
+            if not (_always_use_pickle_for(x) or b"__main__" not in result):
+                result = cloudpickle.dumps(x, **dump_kwargs)
     except Exception:
         try:
-            return cloudpickle.dumps(x, **dump_kwargs)
+            result = cloudpickle.dumps(x, **dump_kwargs)
         except Exception as e:
             logger.info("Failed to serialize %s. Exception: %s", x, e)
             raise
+    return result
 
 
-def loads(x):
+def loads(x, *, buffers=()):
     try:
-        return pickle.loads(x)
+        if buffers:
+            return pickle.loads(x, buffers=buffers)
+        else:
+            return pickle.loads(x)
     except Exception:
         logger.info("Failed to deserialize %s", x[:10000], exc_info=True)
         raise
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -36,18 +36,18 @@ def dumps(x, *, buffer_callback=None):
     if HIGHEST_PROTOCOL >= 5 and buffer_callback is not None:
         dump_kwargs["buffer_callback"] = buffers.append
     try:
-        del buffers[:]
+        buffers.clear()
         result = pickle.dumps(x, **dump_kwargs)
         if len(result) < 1000:
             if b"__main__" in result:
-                del buffers[:]
+                buffers.clear()
                 result = cloudpickle.dumps(x, **dump_kwargs)
         elif not _always_use_pickle_for(x) and b"__main__" in result:
-            del buffers[:]
+            buffers.clear()
             result = cloudpickle.dumps(x, **dump_kwargs)
     except Exception:
         try:
-            del buffers[:]
+            buffers.clear()
             result = cloudpickle.dumps(x, **dump_kwargs)
         except Exception as e:
             logger.info("Failed to serialize %s. Exception: %s", x, e)
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -31,23 +31,30 @@ def dumps(x, *, buffer_callback=None):
     2.  If it is short then check if it contains __main__
     3.  If it is long, then first check type, then check __main__
     """
+    buffers = []
     dump_kwargs = {"protocol": HIGHEST_PROTOCOL}
     if HIGHEST_PROTOCOL >= 5:
-        dump_kwargs["buffer_callback"] = buffer_callback
+        dump_kwargs["buffer_callback"] = buffers.append
     try:
+        del buffers[:]
         result = pickle.dumps(x, **dump_kwargs)
         if len(result) < 1000:
             if b"__main__" in result:
+                del buffers[:]
                 result = cloudpickle.dumps(x, **dump_kwargs)
         else:
             if not (_always_use_pickle_for(x) or b"__main__" not in result):
+                del buffers[:]
                 result = cloudpickle.dumps(x, **dump_kwargs)
     except Exception:
         try:
+            del buffers[:]
             result = cloudpickle.dumps(x, **dump_kwargs)
         except Exception as e:
             logger.info("Failed to serialize %s. Exception: %s", x, e)
             raise
+    for b in buffers:
+        buffer_callback(b)
     return result
 
 
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -31,21 +31,22 @@ def dumps(x):
     2.  If it is short then check if it contains __main__
     3.  If it is long, then first check type, then check __main__
     """
+    dump_kwargs = {"protocol": HIGHEST_PROTOCOL}
     try:
-        result = pickle.dumps(x, protocol=HIGHEST_PROTOCOL)
+        result = pickle.dumps(x, **dump_kwargs)
         if len(result) < 1000:
             if b"__main__" in result:
-                return cloudpickle.dumps(x, protocol=HIGHEST_PROTOCOL)
+                return cloudpickle.dumps(x, **dump_kwargs)
             else:
                 return result
         else:
             if _always_use_pickle_for(x) or b"__main__" not in result:
                 return result
             else:
-                return cloudpickle.dumps(x, protocol=HIGHEST_PROTOCOL)
+                return cloudpickle.dumps(x, **dump_kwargs)
     except Exception:
         try:
-            return cloudpickle.dumps(x, protocol=HIGHEST_PROTOCOL)
+            return cloudpickle.dumps(x, **dump_kwargs)
         except Exception as e:
             logger.info("Failed to serialize %s. Exception: %s", x, e)
             raise
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -33,7 +33,7 @@ def dumps(x, *, buffer_callback=None):
     """
     buffers = []
     dump_kwargs = {"protocol": HIGHEST_PROTOCOL}
-    if HIGHEST_PROTOCOL >= 5:
+    if HIGHEST_PROTOCOL >= 5 and buffer_callback is not None:
         dump_kwargs["buffer_callback"] = buffers.append
     try:
         del buffers[:]
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -52,8 +52,9 @@ def dumps(x, *, buffer_callback=None):
         except Exception as e:
             logger.info("Failed to serialize %s. Exception: %s", x, e)
             raise
-    for b in buffers:
-        buffer_callback(b)
+    if buffer_callback is not None:
+        for b in buffers:
+            buffer_callback(b)
     return result
 
 
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -32,6 +32,8 @@ def dumps(x, *, buffer_callback=None):
     3.  If it is long, then first check type, then check __main__
     """
     dump_kwargs = {"protocol": HIGHEST_PROTOCOL}
+    if HIGHEST_PROTOCOL >= 5:
+        dump_kwargs["buffer_callback"] = buffer_callback
     try:
         result = pickle.dumps(x, **dump_kwargs)
         if len(result) < 1000:
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -42,7 +42,7 @@ def dumps(x, *, buffer_callback=None):
             if b"__main__" in result:
                 del buffers[:]
                 result = cloudpickle.dumps(x, **dump_kwargs)
-        elif not (_always_use_pickle_for(x) or b"__main__" not in result):
+        elif not _always_use_pickle_for(x) and b"__main__" in result:
             del buffers[:]
             result = cloudpickle.dumps(x, **dump_kwargs)
     except Exception:
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -42,10 +42,9 @@ def dumps(x, *, buffer_callback=None):
             if b"__main__" in result:
                 del buffers[:]
                 result = cloudpickle.dumps(x, **dump_kwargs)
-        else:
-            if not (_always_use_pickle_for(x) or b"__main__" not in result):
-                del buffers[:]
-                result = cloudpickle.dumps(x, **dump_kwargs)
+        elif not (_always_use_pickle_for(x) or b"__main__" not in result):
+            del buffers[:]
+            result = cloudpickle.dumps(x, **dump_kwargs)
     except Exception:
         try:
             del buffers[:]
--- a/distributed/protocol/pickle.py
+++ b/distributed/protocol/pickle.py
@@ -1,5 +1,6 @@
 import logging
 import pickle
+from pickle import HIGHEST_PROTOCOL
 
 import cloudpickle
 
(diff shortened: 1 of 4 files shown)


distributed/protocol/pickle.py
```diff
@@ -23,36 +25,43 @@ def _always_use_pickle_for(x):
...
+def dumps(x, *, buffer_callback=None):
     """ Manage between cloudpickle and pickle
 
     1.  Try pickle
     2.  If it is short then check if it contains __main__
     3.  If it is long, then first check type, then check __main__
     """
+    pickle_dumps = functools.partial(pickle.dumps, protocol=HIGHEST_PROTOCOL)
+    cloudpickle_dumps = functools.partial(cloudpickle.dumps, protocol=HIGHEST_PROTOCOL)
+    if pickle.HIGHEST_PROTOCOL >= 5:
```
jcrist, line 37: Nitpick for readability - should use either `HIGHEST_PROTOCOL` or `pickle.HIGHEST_PROTOCOL`, not both. Slight preference to keep namespaced version, but really doesn't matter.

jcrist: There's also a slight readability overhead IMO by building up nesting `functools.partial` calls. Slight preference to reorg to avoid this if possible, or handle by building up `kwargs` dicts instead.

jakirkham: Thanks Jim! 😄

Yeah that slipped. Sorry about that. Will fix.

Good point. Will update.

jakirkham: Fixed in recent commits. Also added a couple tests 🙂

jakirkham:

distributed/protocol/pickle.py
```diff
@@ -23,36 +24,39 @@ def _always_use_pickle_for(x):
...
+def dumps(x, *, buffer_callback=None):
     """ Manage between cloudpickle and pickle
 
     1.  Try pickle
     2.  If it is short then check if it contains __main__
     3.  If it is long, then first check type, then check __main__
     """
+    dump_kwargs = {"protocol": HIGHEST_PROTOCOL}
+    if HIGHEST_PROTOCOL >= 5:
+        dump_kwargs["buffer_callback"] = buffer_callback
```
jcrist, line 36: Thinking a bit more, I think this current structure isn't correct. Passing in a buffer callback to this function can result in the buffer callback being called an incorrect number of times. For example, say we're pickling a tuple:

```
(object_that_supports_pickle_5, object_that_requires_cloudpickle)
```

With the current logic of trying pickle, then maybe trying cloudpickle, the following happens

- The buffer callback is used on the first call to `pickle.dumps` when it hits `object_that_supports_pickle_5`
- pickle errors when it hits `object_that_requires_cloudpickle`
- cloudpickle is then tried, resulting in the buffer callback being used on `object_that_supports_pickle_5` again
- The list of buffers contains erroneous items

To avoid that, I think you'll probably want to manage the list of buffers (and the callback) all internally in this file, and only return a list of frames.

jakirkham: That's a fair concern. I'll take a look. Though am out this week. So it likely won't be until next week.

jakirkham: Ok have pushed a fix to address this. Please let me know what you think 🙂

jakirkham:

jcrist:
Could you wait a bit? I still want to review this, just a bit swamped right now.

jakirkham:
Sure. Do you have a (rough) sense of when you might be able to look at it (for my own planning)?

jcrist:
Probably sometime tomorrow.

distributed/protocol/pickle.py
```diff
@@ -23,36 +24,46 @@ def _always_use_pickle_for(x):
...
     2.  If it is short then check if it contains __main__
     3.  If it is long, then first check type, then check __main__
     """
+    buffers = []
+    dump_kwargs = {"protocol": HIGHEST_PROTOCOL}
+    if HIGHEST_PROTOCOL >= 5 and buffer_callback is not None:
+        dump_kwargs["buffer_callback"] = buffers.append
     try:
-        result = pickle.dumps(x, protocol=pickle.HIGHEST_PROTOCOL)
+        del buffers[:]
```
jcrist, line 39: This isn't needed here.

distributed/protocol/pickle.py
```diff
@@ -23,36 +24,46 @@ def _always_use_pickle_for(x):
...
-            else:
-                return result
+                del buffers[:]
+                result = cloudpickle.dumps(x, **dump_kwargs)
         else:
-            if _always_use_pickle_for(x) or b"__main__" not in result:
-                return result
-            else:
-                return cloudpickle.dumps(x, protocol=pickle.HIGHEST_PROTOCOL)
+            if not (_always_use_pickle_for(x) or b"__main__" not in result):
```
jcrist, line 46: could be `elif not ...`, dropping one indentation level.

distributed/protocol/pickle.py
```diff
@@ -23,36 +24,46 @@ def _always_use_pickle_for(x):
...
     except Exception:
         try:
-            return cloudpickle.dumps(x, protocol=pickle.HIGHEST_PROTOCOL)
+            del buffers[:]
+            result = cloudpickle.dumps(x, **dump_kwargs)
         except Exception as e:
             logger.info("Failed to serialize %s. Exception: %s", x, e)
             raise
+    for b in buffers:
+        buffer_callback(b)
```
jcrist, line 57: `buffer_callback` could be `None` per the signature, it's just never not none in usage.

```suggestion
    if buffer_callback is not None:
        for b in buffers:
            buffer_callback(b)
```

jcrist: Just noticed that `buffers` will never be appended to if `buffer_callback` is None, so the original code is correct. I find the explicit `is not None` check more readable, but up to you.

distributed/protocol/pickle.py
jakirkham: Yeah I was worried people might add lines before this and forget to add the deletion. Though maybe changes here are rare. No strong feelings either way. Can drop if you prefer 🙂

distributed/protocol/pickle.py
jakirkham: Yeah tried both ways and went with the current form as it was less verbose. Happy to change if you prefer.

distributed/protocol/pickle.py
jcrist: That's a fair point, fine keeping it. I do find `buffers.clear()` easier to read than `del buffers[:]`, so if you'd be up for changing that, that'd be my preference. Up to you I guess.

distributed/protocol/pickle.py
jakirkham: Thanks! Done. Also applied the `not` through to simplify the conditional a bit.

distributed/protocol/pickle.py
jcrist: In this case I think an explicit check is easier to understand. It took me a sec to fully understand the current logic.

distributed/protocol/pickle.py
jakirkham: Sounds good to me. Pushed a commit using `.clear()` instead. Please let me know if that looks ok.

distributed/protocol/pickle.py
jakirkham: Yeah that makes sense. Included the change. Thanks!

jakirkham:
I think everything has been addressed with the recent changes. Please let me know if there's anything else :)

jakirkham:
Thanks Jim! Really appreciate the feedback. Feel like this is a cleaner change as a result :)

jacobtomlinson:
Tests passing 🎉

gshimansky:
Is there any check that this code should run only on Python 3.8+?
I am getting this exception on Python 3.7:
```
  File "/home/teamcity/.local/lib/python3.7/site-packages/distributed/worker.py", line 2511, in execute
    data[k] = self.data[k]
  File "/home/teamcity/.local/lib/python3.7/site-packages/zict/buffer.py", line 78, in __getitem__
    return self.slow_to_fast(key)
  File "/home/teamcity/.local/lib/python3.7/site-packages/zict/buffer.py", line 65, in slow_to_fast
    value = self.slow[key]
  File "/home/teamcity/.local/lib/python3.7/site-packages/zict/func.py", line 38, in __getitem__
    return self.load(self.d[key])
  File "/home/teamcity/.local/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 502, in deserialize_bytes
    return deserialize(header, frames)
  File "/home/teamcity/.local/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 302, in deserialize
    return loads(header, frames)
  File "/home/teamcity/.local/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 64, in pickle_loads
    return pickle.loads(x, buffers=buffers)
  File "/home/teamcity/.local/lib/python3.7/site-packages/distributed/protocol/pickle.py", line 64, in loads
    return pickle.loads(x, buffers=buffers)
TypeError: 'buffers' is an invalid keyword argument for loads()
```
and downgrading to distributed 2.16.0 made it go away.

jcrist:
That looks like you might have some clients/workers/scheduler on 3.8 and some on < 3.8. Dask nodes (client/worker/scheduler) use the highest pickle protocol available to them, if not all nodes support pickle 5 you'll get that error.

If you don't have a mix of python versions, then this is a bug, but looking at the code I don' see how that's possible.

jakirkham:
Agreed. Would try running [`client.get_versions(check=True)`]( https://distributed.dask.org/en/latest/api.html#distributed.Client.get_versions ) on your cluster to make sure you have a working environment. Should help you pinpoint any version mismatches.

gshimansky:
I am running on one system only and it has only Python 3.6.8 and 3.7.5 installed from Ubuntu distro. I checked with `get_versions` and they all return `'python': '3.7.5.final.0'` as expected.
I will try to debug maybe and will let you know if I find how this bug happens.

jcrist:
Hmmm, that is concerning. Can you reproduce it on a local cluster?

```python
client = Client()
client.submit(lambda: "test function").result()
```

jakirkham:
Might need to use a NumPy array and force `pickle` for communication to create a reproducer. Alternatively could use the `MemoryviewHolder` class from the tests here. Something like that should generate a non-trivial number of buffers.

gshimansky:
@jcrist Yes I am getting back `'test function'`.
This is a very subtle bug and for some reason it is reproducible only in test environment. But am able to do this consistently, so I will try to find the reason.

jakirkham:
It would be good to start debugging with something where out-of-band pickling would be used. NumPy arrays would work for example (though we have to force them to go through pickle). Alternatively something that implements pickle protocol 5 (like [`MemoryviewHolder`]( https://github.com/dask/distributed/blob/ddc6377b76f957ef4a3a607f1dc2cf2ba772abc7/distributed/protocol/tests/test_pickle.py#L26-L34 )) that Dask doesn't otherwise know how to serialize. Here's an example with NumPy:

```python
import numpy as np

client = Client(serializers=["pickle"])
client.submit(lambda a, b: a + b, np.arange(3), 2).result()
```

This may serve as a good starting point for debugging further. Though it may need further adjustment. Like using `client.run` to hit all workers. Or only returning or only passing a NumPy array instead of both (as done above). Additionally there may be other flags we need to set to ensure pickling happens (though using `MemoryviewHolder` should avoid that).

Hopefully that gives some ideas on how to narrow this down further.

mrocklin:
Do Pandas dataframes support the new protocol?  To me that seems like the
highest impact application of this work.

jcrist:
Yes, since pickle recurses, anything that has a `__reduce__` implementation that returns bumpy arrays (or other objects that directly support pickle 5) will also then use the new protocol. No special casing in pandas needed.

mrocklin:
Well then great.  Perhaps that gives us our test case

jakirkham:
Yeah this is what @TomAugspurger and I were learning in issue ( https://github.com/pandas-dev/pandas/issues/34244 ). Also brought this up in comment ( https://github.com/dask/distributed/issues/614#issuecomment-631033227 ). It would be good to have some tests in Pandas that cover pickling with protocol 5 as well. Though based on local exploration this does seem to efficiently serialize Pandas objects.

While more efficient serialization of Pandas objects is an immediate win with impact, I think the bigger win is that any object that supports this protocol is now serialized efficiently. So it lowers the bar for getting efficient serialization with Dask. We've done this sort of work for RAPIDS for example. Possibly other libraries like spaCy will do this as well ( https://github.com/explosion/spaCy/issues/5472 ). It might be nice to identify some other pain points like this one. They should now be low hanging fruit.

Additionally it would be nice to cover the last mile of Python pre-3.8 support through the [pickle5 backport package]( https://github.com/pitrou/pickle5-backport ). This would allow Dask to have efficient serialization on any supported Python version. Did some work to make cloudpickle compatible with pickle5 in PR ( https://github.com/cloudpipe/cloudpickle/pull/370 ) thanks to help from @pierreglaser. Stuck on a couple remaining test failures (if anyone has thoughts there 😉). Expect the changes for pickle5 support in Dask to be much simpler once cloudpickle supports it.

Once we wrap that up, it would be good to share this work more openly. Was thinking of doing a blogpost to discuss this along the lines of your, ["Pickle isn't slow, it's a protocol" blogpost]( http://matthewrocklin.com/blog/work/2018/07/23/protocols-pickle ). It would be nice to point out to the community this avenue for efficient serialization in distributed computing (whether or not they use Dask). In fact it seems some people are already taking notice ( https://github.com/ray-project/ray/pull/8577 ) 😉

gshimansky:
Well, I am actually experiencing the problem when I run tests on Modin project https://github.com/modin-project/modin and Pandas DataFrames and Series are the objects that are passed to Dask workers.

jcrist:
Are you able to reproduce without modin?

gshimansky:
> Are you able to reproduce without modin?
Not yet. It is quite fragile and easy to lose when making changes.

jakirkham:
@gshimansky, have you tried [my suggestion above]( https://github.com/dask/distributed/pull/3784#issuecomment-635761343 )?

gshimansky:
I tried to debug `distributed` code. So far I found that `HIGHEST_PROTOCOL` value is always 4. I also found out that `pickle_loads` gets called with `frames` of length 2 which results in `buffers` being not-None. `frames` of length 2 come from `deserialize_bytes` but there the trace is lost because I don't know whether `decompress` should return such value or not.

jakirkham:
Maybe Dask is splitting up the frames somehow when serializing and not joining them correctly when deserializing? That's my best guess. Though an example would really help (in addition to being a useful test case for a fix 😉)

gshimansky:
My question is, how should it work correctly when `HIGHEST_PROTOCOL` equals 4 and therefore `loads` doesn't have a `buffers` argument? If `frames` argument comes of length 2, this results in `buffers` not-None and `loads` gets called with `buffers` which it doesn't support.

jakirkham:
We don't collect `buffers` when protocol is `4`. So it should be empty.

However Dask sometimes splits up frames for various reasons (networking, limitations on compression, etc.) independently of whether pickling was used. If Dask didn't join them back together correctly, that might present as the bug you are seeing.

Again we would need an MRE to confirm that though.

gshimansky:
Ok so do you mean that if protocol is `4` then `decompress` in `deserialize_bytes` should not return `frames` list of length greater than `1`? I need to know if this behavior is allowed or not to investigate farther.

distributed/protocol/serialize.py
```diff
@@ -52,11 +52,16 @@ def dask_loads(header, frames):
...
+    header = {"serializer": "pickle"}
+    frames = [None]
+    buffer_callback = lambda f: frames.append(memoryview(f))
+    frames[0] = pickle.dumps(x, buffer_callback=buffer_callback)
+    return header, frames
 
 
 def pickle_loads(header, frames):
-    return pickle.loads(b"".join(frames))
+    x, buffers = frames[0], frames[1:]
```
jakirkham, line 63: My expectation is that `frames[1:]` would be an empty list here. It might not be though. If it's not, that would be good to fix. Whether that has to do with `deserialize_bytes` behavior or not, I'm less sure. Maybe someone else knows?

gshimansky: `frames` in `pickle_loads` comes from `deserialize_bytes` for me (take a look at exception stack trace). And for me `frames[1:]` is not empty. That's why I am asking if this is expected or not.

jakirkham: Right. So do we have a reproducer? That should help us debug this further.

jakirkham: FWIW I've been trying to reproduce this too, but haven't had any luck. Though I believe you that there could be a problem here. It's just tricky to come up with fixes in the dark. So anything you can come up with would help 🙂

gshimansky: If you say that Dask is able to split frames, may it happen when large pandas `DataFrames` are serialized? My test is a benchmark which operates on such objects, about 20M of records long and 40 columns wide.

jakirkham: Indeed it might. In particular am eyeing [this function]( https://github.com/dask/distributed/blob/fbdb067ed514024aad14dbdc3790bae9344548df/distributed/protocol/utils.py#L22 ). Though I have tried creating fairly large NumPy arrays and haven't managed to trigger that case. Still it seems like that is probably what is happening. Just need to narrow it down a bit more.

jakirkham: Maybe something like PR ( https://github.com/dask/distributed/pull/3841 ) will help? Still lack a test case though. So I have no way to confirm whether it actually helps (or ensure we don't accidentally regress).

jcrist: Even when sending large bytes that do get split by `frame_split_size`, I'm unable to reproduce an error. Things are properly reassembled on the other side. Not sure what's going on there.

jakirkham: Yeah me neither unfortunately. Happy to dig deeper once we identify a reproducer 🙂

jakirkham: A similar issue was raised recently ( https://github.com/dask/distributed/issues/3851 ). Not sure if that is the same. Suspect PR ( https://github.com/dask/distributed/pull/3639 ) fixes this. Would be good if you can try though.

jakirkham: This was fixed by PR ( https://github.com/dask/distributed/pull/3639 ).

jakirkham:
This has been extended to Python versions pre-3.8 with `pickle5` with PR ( https://github.com/dask/distributed/pull/3849 ).

PR #3788, opened 2020-05-09
Await self before handling comms

mrocklin:
This stops servers from handling messages until they are finished
starting.

Additionally, we centralize the status management to Server.__await__

Lots of things break currently, but I didn't want to lose this.


quasiben:
Did you end up resolve the things which broke ?

mrocklin:
Yes.

Issue #3745, opened 2020-04-27
Compute/Transfer Ratio


quasiben:
Often I want to see the a ratio of time spent computing vs transferring.  At the moment, I can get to it (help from @mrocklin ) by doing something like the following:

```python
with get_task_stream() as data:
    u, s, v = da.linalg.svd_compressed(x, k=10, seed=rs)
    v.compute();

timespent = defaultdict(int)

for d in data.data:
    for x in d['startstops']:
        start = x['start']
        stop = x['stop']
        compute = x['action']
        timespent[compute] += stop-start

defaultdict(int,
            {'compute': 49.57789707183838,
             'transfer': 152.50884366035461,
             'disk-read': 0.009233713150024414})
```

I think this is a useful enough measurement that we should consider including this is a dask dashboard plot.  Thoughts ?

mrocklin:
This is somewhat similar to the `Memory Use (operation)` plot.  A chart
like that that talked about compute time, and included transfer would
probably be useful.

mrocklin:
(this is a barchart that includes one element per key prefix, like
"read-csv", or "tensordot")

mrocklin:
I think that this information might be present already in 

```python
{k: p.duration for k, p in Scheduler.prefixes.items()}
```

For transfer times we'll have to do something else.  This might be a good opportunity to rethink how we look at bandwidth as well.

PR #3792, opened 2020-05-12
Action Dashboard

quasiben:
This PR is intended to resolve https://github.com/dask/distributed/issues/3745 .  

This is a new dashboard is intended to give feedback to the user for how tasks are spending their time.  Are they computing / transferring / deserializing ?

Screen Shot 2020-05-12 at 8 54 53 AM


mrocklin:
Some thoughts:

1.  It would be good to use the color scheme that we use for the other plots.  See the `color_of`  function.  For transfer, and other components we use colors defined here:

```python
def color_of_message(msg):
    if msg["status"] == "OK":
        split = key_split(msg["key"])
        return color_of(split)
    else:
        return "black"


colors = {
    "transfer": "red",
    "disk-write": "orange",
    "disk-read": "orange",
    "deserialize": "gray",
    "compute": color_of_message,
}
```

2.  I think that while the transfer/compute ratio is most critical to you all, I'm not confident that this will be the common case.  I'm not sure that it should be the title of the plot.

3.  When I originally envisioned this plot I thought of it as one bar per task prefix, and then one bar for transfer.  Breaking up transfer costs per task prefix is nice I think.  It makes the plot a little bit busier (it's harder to see everything at once) but I can imagine situations where it would be useful.  

    Along these same lines, I'm curious if there are ways to make this plot more informative while also simpler.  For example maybe we can remove the transfer/compute prefixes and rely on colors to help here, allowing then the task prefix labels to be tilted slightly to fit more comfortably.  

These are all just thoughts.  They're intended to encourage other thoughts, not say exactly what we should do.

distributed/dashboard/scheduler.py
```diff
@@ -83,6 +84,7 @@ def connect(application, http_server, scheduler, prefix=""):
     "/individual-bandwidth-types": individual_bandwidth_types_doc,
     "/individual-bandwidth-workers": individual_bandwidth_workers_doc,
     "/individual-memory-by-key": individual_memory_by_key_doc,
+    "/individual-ratio-by-key": individual_ratio_by_key_doc,
```
mrocklin, line 87: Personally I don't think that the compute/transfer _ratio_ is the main point here.  I think that it's mostly compute time by key that most people will think about.  Transfer is in there just as a little extra.

distributed/dashboard/components/scheduler.py
```diff
@@ -438,6 +439,116 @@ def name(address):
...
+                for data in startstops:
+
+                    act = data["action"]
+
+                    start = data["start"]
+                    stop = data["stop"]
+                    time = stop - start
+
+                    key_vals[(name, act)] += time
+                    totals[act] += time
```
mrocklin, line 532: This walks over all tasks and recomputes things every 500ms.  This might be ok, but it might be nicer to instead track some of these metrics on the `TaskPrefix` task in the scheduler during normal operation, and then reflect those numbers here as necessary.

In general with dashboard plots it's better if they do as little work assembling data as possible.  The same things that we humans will want to know are also things that the scheduler will want to know to make better scheduling decisions in the future.  If we can do this work in the scheduler itself just as cheaply (or more cheaply in this case) then I recommend that we do so.

It looks like we already track the duration in every TaskGroup.  Here is a snippet of code from distributed/scheduler.py

```python
            if startstops:
                L = [
                    (startstop["start"], startstop["stop"])
                    for startstop in startstops
                    if startstop["action"] == "compute"
                ]
                if L:
                    compute_start, compute_stop = L[0]
                else:  # This is very rare
                    compute_start = compute_stop = None
            else:
                compute_start = compute_stop = None

            #############################
            # Update Timing Information #
            #############################
            if compute_start and ws.processing.get(ts, True):
                # Update average task duration for worker
                old_duration = ts.prefix.duration_average or 0
                new_duration = compute_stop - compute_start
                if not old_duration:
                    avg_duration = new_duration
                else:
                    avg_duration = 0.5 * old_duration + 0.5 * new_duration

                ts.prefix.duration_average = avg_duration
                ts.group.duration += new_duration
```

Maybe we can do something similar there with transfers (and all other components) and manage state that way.

mrocklin:
> Along these same lines, I'm curious if there are ways to make this plot more informative while also simpler. For example maybe we can remove the transfer/compute prefixes and rely on colors to help here, allowing then the task prefix labels to be tilted slightly to fit more comfortably.

A legend might also help here, perhaps one also where we can select and deselect different components.  My by default we keep transfers off, but people can click on the transfer element in the legend to turn them on in the plot?  I'm not sure how baked into Bokeh this functionality.

quasiben:
> 1.  It would be good to use the color scheme that we use for the other plots.  See the `color_of`  function.  For transfer, and other components we use colors defined here:

Totally agree

> 2.  I think that while the transfer/compute ratio is most critical to you all, I'm not confident that this will be the common case.  I'm not sure that it should be the title of the plot.

Also agree, I started with just transfer/compute then expanded to all actions.  I'll update the title and URL

> 3.  When I originally envisioned this plot I thought of it as one bar per task prefix, and then one bar for transfer.  Breaking up transfer costs per task prefix is nice I think.  It makes the plot a little bit busier (it's harder to see everything at once) but I can imagine situations where it would be useful.  

I had similar thoughts as well, I briefly considered doing stacked bars here but I thought it easier to visually compare actions across a prefix if they were grouped as they are now

>    Along these same lines, I'm curious if there are ways to make this plot more informative while also simpler.  For example maybe we can remove the transfer/compute prefixes and rely on colors to help here, allowing then the task prefix labels to be tilted slightly to fit more comfortably.  

Moving to color with a legend is doable.  Note, it's not just transfer/compute but all the actions so it can get a lot busier with more complete workflows

quasiben:
Still working on improvements but wanted to post latest style:
Screen Shot 2020-05-12 at 5 08 48 PM

One thing that bothers me is that the `compute` color in the legend will be variable as.  I asked bokeh devs if it [was possible to have a rainbow in the legend](https://discourse.bokeh.org/t/rainbow-spectrum-in-legend-box/5449) and @jsignell was kind of enough to respond but nothing so far.  This detail is probably not worth the amount of time I put into it :/

quasiben:
In commit https://github.com/dask/distributed/pull/3792/commits/332345fd51b0996defd525e9031bec0de803758f, I swapped out get_task_stream for a newly create `add_durations` dict within TaskGroups. Below is a demo of this running.  Two things I want to point out:

1) The plots is only running when during the computation.  This is the same for many of the `/individual-*` plots.  How can we capture this plot or something like it during a call to `performance_report`
2) After the first run, (and I have no idea why this happens), the bokeh plot auto zooms in (refreshing the page resolves the issue -- still, it is strange). You can see this if you watch the gif long enough and wait for the second call.

action-dashboard-1

mrocklin:
> The plots is only running when during the computation. This is the same for many of the /individual-* plots. How can we capture this plot or something like it during a call to performance_report

Currently the TaskGroup objects go away, but the TaskPrefix objects stick around.  Maybe it makes sense to record this data there instead?  That would give us permanent history rather than very little history, which might also be better but non-ideal?  The other model for this kind of thing is what we do with bandwidths, which have exponential moving averages.  Maybe we do something like exponential decay?  

Bandwidths are, to my understanding, also not ideal currently.  Maybe we could use some larger thinking here about how to handle this problem more holistically.

mariusvniekerk:
So one kind of view that might be interesting to consider for this is a more heatmap oriented view that that you can see distribution of durations?  And have a toggle to include/exclude transfer time in that?

quasiben:
I switched to TaskPrefixes and that worked nicely.  Below is the plot of:

> y = (x + x.T) - x.mean(axis=0); y.sum().compute()

Screen Shot 2020-05-14 at 9 54 18 PM

I also double checked that the action times from the task stream generally lined up with the plot:

```
Out[29]: 
defaultdict(int,
            {'compute': 470.11822295188904,
             'transfer': 457.4781596660614,
             'disk-read': 59.76479458808899,
             'disk-write': 29.56607961654663,
             'deserialize': 0.04147481918334961})
```

Heat map would be interesting as would toggles.  The only toggles I have seen thus far are dropdowns in the `crossfilter` plot.  I think I need some help in determining what is useful here.  Originally, I wanted to answer questions around, "where is my workflow spending time ?  Computing ? Transferring ?"  The barchart helps answer that question but the prefixes diffuse the aggregate behavior a bit.

mrocklin:
> Originally, I wanted to answer questions around, "where is my workflow spending time ? Computing ? Transferring ?" The barchart helps answer that question but the prefixes diffuse the aggregate behavior a bit. 

Another option here is to factor out this information and make two bar-charts.  One for compute-time-per-key and another which is about aggregate compute/transfer/disk-read over all keys.

mrocklin:
> Another option here is to factor out this information and make two bar-charts. One for compute-time-per-key and another which is about aggregate compute/transfer/disk-read over all keys.

Did this make sense?

The current plot seems busy to me.  It's not immediately clear what is going on without first staring at it for a while.  I wonder if this is a sign that we're trying to break out too much information.  Two plots with more aggreagted information might be more readily digestible.

quasiben:
@mrocklin Does something like the following capture what you are describing ?


Screen Shot 2020-06-01 at 9 33 48 AM

I like what shown above.  The user can clearly see at a high level where in the workflow the computation is spending time and how other actions: transfer, read/write, etc. can impact performance.  When we read from a cloud storage (gcsfs, s3, etc) I don' think that will show up here, will it?

mrocklin:
> Does something like the following capture what you are describing ?

Yes

> When we read from a cloud storage (gcsfs, s3, etc) I don' think that will show up here, will it?

Correct.

quasiben:
Things are now functional.  I don't love that the bars are huge for actions when full screen:

Screen Shot 2020-06-02 at 5 43 49 PM

Screen Shot 2020-06-02 at 5 43 39 PM

mrocklin:
Cool.  Thoughts on making this two different `individual-foo` plots?  That way you could arrange them separate in JupyterLab.

distributed/scheduler.py
```diff
@@ -4127,12 +4131,19 @@ def transition_processing_memory(
...
+                for startstop in startstops:
+                    stop = startstop["stop"]
+                    start = startstop["start"]
+                    action = startstop["action"]
+                    if action == "compute":
+                        L.append((start, stop))
+
+                    # record timings of all actions -- a cheaper way of
+                    # getting timing info compared with get_task_stream()
+                    ts.prefix.all_durations[action] += stop - start
```
mrocklin, line 4144: I imagine that in the future we'll also want to accumulate counts so that we can get things like average transfer duration.  That should be future work though once we have a need.

mrocklin:
This looks great to me.  Some small style suggestions:

1.  Replace "key" with "type of task" or something that a novice user might better understand.
2.  Axis label of `Time (s)` ?
3.  Increase font size on the x-axis?  (although maybe this is just due to you having a large monitor, probably better to use the defaults)
4.  As above, I think that these might feel natural as two independent plots.  Happy to be overruled here though.

quasiben:
> Cool. Thoughts on making this two different individual-foo plots? That way you could arrange them separate in JupyterLab.

I thought about this but opt-ed against because you would then be passing over the same data twice and was trying to do things as cheaply as possible.  If you think this is not too expensive I can it breakout into separate plots

mrocklin:
We're iterating over the task prefixes, which I suspect will be relatively few (~10).  I think that this won't matter much. 

It would be a bigger deal if we were iterating over workers (~100) or recent tasks (~1000 or ~100000), but prefixes are likely to be trivial I think.

quasiben:
Thanks, I'll breakout the plots then.  Do you think we should also have a page with both or leave that for jupyterlab ?

mrocklin:
Recently my approach has been to leave the construction of composite
layouts to JupyterLab.  This might not be best though, especially because
not everyone uses JupyterLab. My recommendation would be to leave composite
pages for future work.  We should figure out how to handle this broadly
across all plots I think.  (cc @jacobtomlinson )

quasiben:
- Split plots into separate endpoints: `/individual-aggregate-time-per-action`, `individual-compute-time-per-key`
- Increased Font size for the aggregate time plot
- Added `Time (s)` axis label
- Title changed to: `Compute Time Per Task`

Screen Shot 2020-06-03 at 1 29 45 PM
Screen Shot 2020-06-03 at 1 29 57 PM

distributed/dashboard/components/scheduler.py
```diff
@@ -438,6 +440,206 @@ def name(address):
...
+            # order by largest time first
+            agg_times = sorted(agg_times.items(), key=lambda x: x[1], reverse=True)
+
+            agg_colors = list()
+            agg_names = list()
+            agg_time = list()
+            for action, t in agg_times:
+                agg_names.append(action)
+                if action == "compute":
+                    agg_colors.append("purple")
```
quasiben, line 630: I arbitrarily chose `purple` as the general compute color

jacobtomlinson:
> My recommendation would be to leave composite pages for future work.

I think leaving composite pages out of the scope of this PR makes sense.

I did play around a while ago with an editable grid layout for the regular dashboard, so folks can lay things out however they want even if they aren't using Jupyter Lab. Maybe something like that would be valuable rather than producing more opinionated composite pages.

mrocklin:
This looks good to me.  Thanks for this @quasiben !  I look forward to playing with it.  Merging.

Issue #3356, opened 2020-01-08
Local test failures


crusaderky:
I can't seem to create a working environment on my MacOS Catalina dev box. CONTRIBUTING.md could use some attention...

```
export PYTHON=3.7 TESTS=true PACKAGES="scikit-learn python-snappy python-blosc" TORNADO=6
source continuous_integration/travis/install.sh
source continuous_integration/travis/setup-ssh.sh
source continuous_integration/travis/run_tests.sh
```
Test failures:

# 1.
Many (all?) tests failed on the ``check_thread_leak()`` helper function, e.g. test_scheduler_bokeh.test_simple:
```
>               assert False, (thread, call_stacks)
E               AssertionError: (<Thread(Profile, started daemon 123145338511360)>, ['  File "/Users/crusaderky/miniconda3/envs/distributed/lib/python...ile "/Users/crusaderky/PycharmProjects/distributed/distributed/profile.py", line 269, in _watch
E                 \tsleep(interval)
E                 '])
E               assert False
```

# 2.
- test_dask_scheduler.test_defaults
- test_dask_scheduler.test_hostport

failed with:
OSError: Timed out trying to connect to 'tcp://[Guidos-MacBook-Pro.local]:8786' after 5 s: Timed out trying to connect to 'tcp://[Guidos-MacBook-Pro.local]:8786' after 5 s: connect() didn't finish in time

# 3.
- test_dask_scheduler.test_no_dashboard
- test_dask_scheduler.test_dashboard

failed with:
```
                with pytest.raises(Exception):
>                   requests.get("http://127.0.0.1:8787/status/")
E                   Failed: DID NOT RAISE <class 'Exception'>
```

# 4.
- test_comms.test_default_client_server_ipv6
- test_comms.test_tcp_client_server_ipv4
- test_comms.test_tcp_client_server_ipv6
- test_comms.test_tls_client_server_ipv4

failed with:
OSError: Timed out trying to connect to 'tls://127.0.0.1:61915' after 5 s: Timed out trying to connect to 'tls://127.0.0.1:61915' after 5 s: connect() didn't finish in time

What am I missing or doing wrong? Have the tests been run recently on MacOSX?

mrocklin:
Interesting.  I no longer have access to an OSX machine.  If you or someone
else is able to diagnose and resolve these issues that would be great.

quasiben:
[...]

PR #3795, opened 2020-05-14
Fix profile thread leakage during test teardown on some platforms

fjetter:
When running tests locally (in a debian9 docker container; docker on osx) I receive errors during test cleanup like

```
AssertionError: (<Thread(Profile, started daemon 139685301057280)>, ['  File "/home/fjetter/miniconda/envs/dask-distributed/lib/python...*self._kwargs)
E                 ', '  File "/workspace/distributed/distributed/profile.py", line 269, in _watch
```

Tracking this down, the profile watch thread has a termination condition which is coupled to the `ioloop` being alive. Therefore, this fixture needed another `pristine_loop` context manager for me to succeed. Ultimately, the construct looks incredibly similar to the contextmanager `clean` and I replaced it with this.

Apparently, the build failures can be reproduced when running on travis.**com** instead of travis.**org**. I didn't try to determine what the differences of the environment are.

Looks like this is one of the errors reported in #3356

### DIFF
--- a/distributed/cli/dask_ssh.py
+++ b/distributed/cli/dask_ssh.py
@@ -175,7 +175,7 @@ def main(
             version=distributed.__version__
         )
     )
-    print("Worker nodes:".format(n=len(hostnames)))
+    print("Worker nodes: {n}".format(n=len(hostnames)))
     for i, host in enumerate(hostnames):
         print("  {num}: {host}".format(num=i, host=host))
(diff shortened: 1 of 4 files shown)


fjetter:
Build failure caused by the linter issue on master

fjetter:
I included the commit with the flake8 changes here in hopes this fixes all builds 🤞

mrocklin:
Looks great.  Thank you @fjetter .

PR #3797, opened 2020-05-14
Fix flake8

fjetter:
Should fix the linter issue on master

### DIFF
--- a/distributed/deploy/old_ssh.py
+++ b/distributed/deploy/old_ssh.py
@@ -212,7 +212,7 @@ def start_scheduler(
     logdir, addr, port, ssh_username, ssh_port, ssh_private_key, remote_python=None,
 ):
     cmd = "{python} -m distributed.cli.dask_scheduler --port {port}".format(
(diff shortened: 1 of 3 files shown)


distributed/deploy/old_ssh.py
```diff
@@ -212,7 +212,7 @@ def start_scheduler(
     logdir, addr, port, ssh_username, ssh_port, ssh_private_key, remote_python=None,
 ):
     cmd = "{python} -m distributed.cli.dask_scheduler --port {port}".format(
-        python=remote_python or sys.executable, port=port, logdir=logdir
```
fjetter, line 215: This confused me a bit but the dask-scheduler CLI doesn't offer to configure logging so this should not have been here. afaik, this cmd arg only creates the log dir and pipes the output there using bash (see 220ff) but doesn't modify cluster startup/logging config.

fjetter:
Tests seem to fail consistently with the error of #3795

fjetter:
This is redundant with #3795

Issue #3800, opened 2020-05-15
Travis failures with OSError: [Errno 98] Address already in use


fjetter:
We're haunted by a few unstable tests which cause more or less reproducible failures on travis but I'm failing to reproduce the failures locally.

We receive the `OSError` in https://github.com/dask/distributed/blob/c5d2e4baf0e03585f04acb6d4aa6293c70f6ed04/distributed/node.py#L135
where the HTTP server is instructed to use a random, available port which should avoid this exact issue.

This can be seen in
#3757 with https://travis-ci.org/github/dask/distributed/jobs/687341481
or 
#3773 with https://travis-ci.org/github/dask/distributed/jobs/683708829

The two issues are completely unrelated and so far, I've only seen this happening with Tornado5 and python3.6.

<details>

<summary>Traceback</summary>

```python
___________________________ test_repr_localcluster ____________________________
1482
1483self = <[AttributeError("'Scheduler' object has no attribute '_address'",) raised in repr()] Scheduler object at 0x7faf614bd898>
1484routes = [('/metrics', <class 'distributed.http.scheduler.prometheus.PrometheusHandler'>, {'dask_server': <[AttributeError("'Sc...eError("'Scheduler' object has no attribute '_address'",) raised in repr()] Scheduler object at 0x7faf614bd898>}), ...]
1485dashboard_address = None, default_port = 8787, ssl_options = None
1486
1487    def start_http_server(
1488        self, routes, dashboard_address, default_port=0, ssl_options=None,
1489    ):
1490        """ This creates an HTTP Server running on this node """
1491    
1492        self.http_application = RoutingApplication(routes,)
1493    
1494        # TLS configuration
1495        tls_key = dask.config.get("distributed.scheduler.dashboard.tls.key")
1496        tls_cert = dask.config.get("distributed.scheduler.dashboard.tls.cert")
1497        tls_ca_file = dask.config.get("distributed.scheduler.dashboard.tls.ca-file")
1498        if tls_cert:
1499            import ssl
1500    
1501            ssl_options = ssl.create_default_context(
1502                cafile=tls_ca_file, purpose=ssl.Purpose.SERVER_AUTH
1503            )
1504            ssl_options.load_cert_chain(tls_cert, keyfile=tls_key)
1505            # We don't care about auth here, just encryption
1506            ssl_options.check_hostname = False
1507            ssl_options.verify_mode = ssl.CERT_NONE
1508    
[...]

PR #3801, opened 2020-05-15
Add retries to server listen

jacobtomlinson:
I thought I'd have a go at adding a little retry logic to the server listen.

As raised in #3800 we occasionally see issues with ports not being available even if they are a random high port. This is likely to be some quirk of the CI system environment, but I thought it wouldn't hurt to retry a couple of times.

There may be a more elegant way of writing this, I'm a little tired 😅.

cc @fjetter


fjetter:
I was hoping for some wizard with a magical wand and an explanation appearing but the practical fix works for me as well :)

PR #3805, opened 2020-05-15
Set colour of nbytes pane based on thresholds

### LINKED ISSUE #3799: Proposal: Update colour scheme for `bytes stored` dashboard component
Currently, the dashboard component that shows the number of bytes stored by each worker, has a simple colour scheme:
- red if it has exceeded the set limits
- orange if it has exceeded 50% of limits
- blue otherwise
https://github.com/dask/distributed/blob/95228dc2d480aa35dc6521cb46bdb34983583faa/distributed/dashboard/components/scheduler.py#L790-L795

I would find it more useful if the colours reflected where the workers sat compared to the `target`/`spill`/`pause`/`terminate` limits that can be set in Dask's configuration.

https://distributed.dask.org/en/latest/worker.html#memory-management

I would be happy to contribute a PR to change this behaviour if you agree this would be a useful change. (I would also appreciate it if you offered opinions on what colours to use)

KrishanBhasin:
Fixes #3799


jacobtomlinson:
I like the idea of warning with yellow and then showing spilling with orange and paused with red.

However as @mrocklin said in #3799 spilling occurs when both the target and spill thresholds are reached.

Therefore I suggest we change orange to be when target is reached (the first occurrence of spilling). We could use yellow to warn at perhaps 50% or 75% of the target threshold?

KrishanBhasin:
Weirdly, running the below I don't see the colour of the bars changing

```python
import dask.array as da
import dask
import dask.dataframe as dd
import dask.config

if __name__=="__main__":
    dask.config.set({"distributed.worker.memory.target":0.2})
    dask.config.set({"distributed.worker.memory.spill":0.3})

    client = Client()


    x = da.random.random((30000, 30000), chunks=(300, 300))
    y = x + x.T
    z = y[::2, 50:].mean(axis=1)

    c = z.compute()
```

KrishanBhasin:
> 
> 
> I like the idea of warning with yellow and then showing spilling with orange and paused with red.
> 
> However as @mrocklin said in #3799 spilling occurs when both the target and spill thresholds are reached.
> 
> Therefore I suggest we change orange to be when target is reached (the first occurrence of spilling). We could use yellow to warn at perhaps 50% or 75% of the target threshold?

I set it at 85% of the target, because otherwise I think the bars will go yellow too low (85% * 60% is ~51%)

jacobtomlinson:
This looks good. Perhaps @mrocklin could comment on whether this is acceptable?

KrishanBhasin:
I'm not sure what caused the Travis CI to fail - it looks unrelated to this PR

jacobtomlinson:
Agree that it looks unrelated. I'll restart it to see if it works this time.

mrocklin:
> This looks good. Perhaps @mrocklin could comment on whether this is acceptable?

Sorry for the delay in response.  I'm going to defer judgement on this entirely to you if that's ok @jacobtomlinson .

jacobtomlinson:
In that case I'm happy with this! Thanks @KrishanBhasin

Issue #3716, opened 2020-04-16
Serialization with `"msgpack"` doesn't preserve `list`s


jakirkham:
Currently if a `list` is serialized (and is small enough), it will be handled correctly.

```python
In [1]: from distributed.protocol import serialize, deserialize                 

In [2]: t = (0, 1, 2)                                                           

In [3]: deserialize(*serialize(t, serializers=["msgpack"]))                     
Out[3]: (0, 1, 2)

In [4]: l = [0, 1, 2]                                                           

In [5]: deserialize(*serialize(l, serializers=["msgpack"]))                     
Out[5]: [0, 1, 2]
```

However for larger `list`s, this breaks down and a `tuple` is returned instead.

```python
In [1]: from distributed.protocol import serialize, deserialize                 

In [2]: t = (0, 1, 2, 3, 4, 5, 6)                                               

In [3]: deserialize(*serialize(t, serializers=["msgpack"]))                     
Out[3]: (0, 1, 2, 3, 4, 5, 6)

In [4]: l = [0, 1, 2, 3, 4, 5, 6]                                               

In [5]: deserialize(*serialize(l, serializers=["msgpack"]))                     
Out[5]: (0, 1, 2, 3, 4, 5, 6)
```

jakirkham:
It's worth noting msgpack itself by default will always return a `list`.

```python
In [1]: import msgpack                                                          

In [2]: t = (0, 1, 2)                                                           

In [3]: msgpack.loads(msgpack.dumps(t))                                         
Out[3]: [0, 1, 2]

In [4]: l = [0, 1, 2]                                                           

In [5]: msgpack.loads(msgpack.dumps(l))                                         
Out[5]: [0, 1, 2]
```

That said, we actually force it to return a `tuple`, which presents its own set of issues.

```python
In [6]: msgpack.loads(msgpack.dumps(l), use_list=False)                         
Out[6]: (0, 1, 2)
```

This choice goes back to PR ( https://github.com/dask/distributed/pull/2000 ). Though it seems we knew this could present a problem ( https://github.com/dask/distributed/pull/2000#issuecomment-396327604 ). Not sure if the reasoning behind that change still holds today or IOW what are constraints are now.

Kobzol:
It would be nice if this was resolved and there were no auto conversions, it also bit us when we were implementing the scheduler in Rust.

jakirkham:
[...]

PR #3809, opened 2020-05-18
Simplify frame encoding

Kobzol:
This PR attempts to simplify the low-level encoding of Dask protocol messages into frames to make it easier to implement handling of the protocol in a statically typed language (like Rust). There are of course multiple ways of solving this issue, in this PR I'll attempt to describe the issue and offer a possible solution that we have used for our Rust scheduler.

### The issue
The current implementation of the `dumps` function in `protocol/cores.py` takes a Python message (usually a dictionary), extracts serialized/serializable values out of the message, puts them into a series of frames and creates headers that describe how to reconstruct the original message from the frame stream. For example, if the input message has a "function" attribute which contains serialized data, Dask will remove the attribute, create a frame with its serialized data and describe how to put the "function" attribute back into the message in a header which is stored into a separate frame. This is displayed here in the top right corner:
image

The problematic part is message reconstruction during deserialization. A natural representation of Dask message types in a statically typed language is to use data structures which strictly describe the various message types, as opposed to loosely typed dictionaries with arbitrary types inside. An example from Rust (simplified):
```rust
struct DaskUpdateGraphTask {
   function: FunctionDefinition,
   arguments: ArgumentsDefinition,
   ...
}
```
If Dask clients/workers can remove arbitrary attributes from these messages and we have to support arbitrary data structure modification queries like `put <something> into "key1" field of the 0th element of the "key2" field of the input message`, we are pretty much left with no choice than to go back to untyped dictionaries, which defeats one of the benefits of statically typed languages, especially for such critical component of the scheduler. It's trivial to support these queries in a dynamic language, but in e.g. Rust it's incredibly painful.

### Proposed solution
Instead of removing the attributes of the Dask protocol messages and then putting them back in, we propose to keep the original message structure, but replace the attribute values with placeholders (see example image above, bottom right corner). A value which has to be offloaded will be replaced with a placeholder dictionary on the wire, which describes from which frames should the value be reconstructed (frame index + frame count) and how (header). With this approach we can model the original message structure and use a generic enumeration (like `InPlace`/`Offloaded`) type for attributes that can be offloaded (potentially for all attributes). Most importantly, we don't have to implement arbitrary data structure modification queries. In our specific implementation, we have two versions of message types that can contain offloadable attributes which can be converted between each other if you provide them a frame array. Here's a pseudocode of how that could look like:
```rust
struct TaskDefinitionOffloaded {
     function: Offloadable<FunctionDefinition>
}
struct TaskDefinition {
     function: FunctionDefinition
}

impl TaskDefinitionOffloaded {
     fn to_task_def(frames: &[Frame]) {
         // reconstruct TaskDefinition from TaskDefinitionOffloaded by reading data
         // from `frames`
    }
}
```

### Implementation
In our implementation, the placeholder is a dictionary with three "magic" keys, the number of frames of the offloaded value, the index of the first frame in the resulting frame stream and the Dask header containing information about compression and other stuff.

During serialization, we replace each offloaded value with this placeholder.
During deserialization, we check if a value if offloaded (by looking for these magic keys) and if yes, we read the corresponding frame range and reconstruct the original value.

The names of the magic keys are rather arbitrary, they can be probably improved. In our implementation, we send one frame less than in the current encoding, because we do not need the intermediate frame with headers describing how to reconstruct offloaded values. The headers are stored directly in the placeholders. This should not increase message size by itself, as all of the headers are also present in the current encoding, just in a different location. I'm not sure about the frame size limitations though. So far in our experiments it has not caused problems.

Even though our changes were motivated by our inability to implement the old protocol encoding elegantly and performantly in Rust, I'm certain that potential implementations in other statically typed languages (like C++) would face identical problems. I am not sure if our proposed solution is the right one, but I think that in order to make implementations of the scheduler (or other Dask parts) easier in statically typed languages, Dask should use a less "hostile" encoding. It would also help in documenting and understanding of the protocol itself. 

In an (slightly unrelated, but still relevant) attempt to unify what comes out of deserialization, the code in this PR also normalizes both lists and tuples to tuple (https://github.com/dask/distributed/issues/3716).

### DIFF
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -11,8 +11,41 @@ from ..utils import nbytes
 
 _deserialize = deserialize
 
-
 logger = logging.getLogger(__name__)
+
+"""
+Some values are sent in a serialized form, which does not need to be understood by the central server.
+These values are offloaded to separate MessagePack frames which are sent after the main message.
+The offloaded values in the main message are replaced by a placeholder object:
+key: {
+    [OFFLOAD_HEADER_KEY]: <message-pack encoded header of this offloaded value>,
+    [OFFLOAD_FINDEX_KEY]: <frames of this value start at this index in the frames array>,
+    [OFFLOAD_FCOUNT_KEY]: <number of frames of this offloaded value> 
+}
+
+The frames for a specific offloaded value can be found at frames[OFFLOAD_FINDEX_KEY:OFFLOAD_FINDEX_KEY+OFFLOAD_FCOUNT_KEY].
+"""
+
+OFFLOAD_HEADER_KEY = "_$header"
+OFFLOAD_FINDEX_KEY = "_$findex"
+OFFLOAD_FCOUNT_KEY = "_$fcount"
+
+
+def _make_offload_value(header, frame_index, frame_count):
+    return {OFFLOAD_HEADER_KEY: header, OFFLOAD_FINDEX_KEY: frame_index,
+            OFFLOAD_FCOUNT_KEY: frame_count}
+
+
+def _extract_offload_value(value):
+    if not isinstance(value, dict) or len(value) != 3:
+        return None
+    frame_index = value.get(OFFLOAD_FINDEX_KEY)
+    if frame_index is None:
+        return None
+    header = value.get(OFFLOAD_HEADER_KEY)
+    if header is None:
+        return None
+    return (header, frame_index)
 
 
 def dumps(msg, serializers=None, on_error="message", context=None):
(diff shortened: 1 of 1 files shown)


mrocklin:
Thank you for the PR @Kobzol and for the informative description.  I apologize for the delay in response.  

In principle I don't object to this change.  However, it would be good to get feedback from the NVIDIA folks on this topic.  I would prefer to avoid changing the core protocol frequently if we can avoid it, so it would be good to get their thoughts before moving forward.

cc @kkraus14 @quasiben @jacobtomlinson

jacobtomlinson:
I understand the motivation behind this and it seems like a good thing to do from that viewpoint. I can't comment on whether we should make other changes. I'll let others weigh in.

cc @madsbk

Issue #2007, opened 2018-05-25
Add Semaphore, Condition, Event


mrocklin:
Dask currently has some distributed coordination primitives like Locks and Queues that match the API of normal local Locks and Queues.  There are other primitives like these that we do not yet support, notable `Semaphore`, `Condition`, and `Event`.

These might be good training for people who want to become familiar with extending the dask.distributed system.  The files `lock.py`, `queues.py`, and `variable.py` might be a good model for development here.

palash25:
Hi, I've been writing python for some time but I'm new to distributed computing. This seems like the right issue to get started. Can I work on this?

mrocklin:
Yes!  Anyone can work on any issue.  You may want to familiarize yourself
with http://distributed.readthedocs.io/en/latest/develop.html

palash25:
Thanks will try to submit a PR within a week's time

andersy005:
- It appears that https://github.com/dask/distributed/pull/2690 is/was meant to implement one (Semaphore) of the missing features listed here. Are there any updates on the status of https://github.com/dask/distributed/pull/2690? 

- https://github.com/dask/distributed/pull/3397 mentions Events and Conditions. Is it the case that these two features have also been implemented already?

I am interested in working on the other features namely (Event, Condition), and wanted to make sure that I am not duplicating work that has already been done.

mrocklin:
> It appears that #2690 is/was meant to implement one (Semaphore) of the missing features listed here. Are there any updates on the status of #2690?

All communication is happening on github.  I recommend checking in on that PR.

> #3397 mentions Events and Conditions. Is it the case that these two features have also been implemented already?

This is different.  Dask uses Tornado/asyncio objects within its codebase.  We would also like to expose distributed versions of these objects to users so that they can make their own applications.

andersy005:
> All communication is happening on github. I recommend checking in on that PR.

👍  will do 

> This is different. Dask uses Tornado/asyncio objects within its codebase. We would also like to expose distributed versions of these objects to users so that they can make their own applications.

Thank you for the clarification!

nadzhou:
Can i work on the `Condition` part, it seems like `Event` has been implemented.

PR #3821, opened 2020-05-21
Implementation of `Event` using `asyncio.Event`

nils-braun:
This PR solves part of #2007 (I assumed no progress so far, since the last update was approx. half a year ago) . It adds a new class `Event` which has the same interface as the `asyncio.Event` and the needed scheduler extension.

Actually, the implementation would have been quite simple: just keeping a `dict` of events and forwarding everything to the `asyncio.Event`. However, this will never clean up the created events in the dictionary, so I added some kind of reference counter additionally.

I am very happy for suggestions - I do not know if this is the best way to do it.

### DIFF
--- a/distributed/event.py
+++ b/distributed/event.py
@@ -77,6 +77,12 @@ class EventExtension:
             finally:
                 self._waiter_count[name] -= 1
 
+                if not self._waiter_count[name] and not event.is_set():
+                    # No one is waiting for this
+                    # and as the default flag for an event is false
+                    # we can safely remove it
+                    self._delete_event(name)
+
             return True
 
     async def event_set(self, stream=None, name=None):
@@ -99,10 +105,7 @@ class EventExtension:
                 # No one is waiting for this
                 # and as the default flag for an event is false
                 # we can safely remove it
-                with suppress(KeyError):
-                    del self._waiter_count[name]
-                with suppress(KeyError):
-                    del self._events[name]
+                self._delete_event(name)
 
             else:
                 # There are waiters
@@ -139,6 +142,13 @@ class EventExtension:
             self._events[name] = asyncio.Event()
 
         return self._events[name]
+
+    def _delete_event(self, name):
+        """ Helper function to delete an event """
+        with suppress(KeyError):
+            del self._waiter_count[name]
+        with suppress(KeyError):
+            del self._events[name]
 
 
 class Event:
--- a/distributed/event.py
+++ b/distributed/event.py
@@ -47,12 +47,14 @@ class EventExtension:
         # we can remove the event
         self._waiter_count = defaultdict(int)
 
-        self.scheduler.handlers.update({
-            "event_wait": self.event_wait,
-            "event_set": self.event_set,
-            "event_clear": self.event_clear,
-            "event_is_set": self.event_is_set
-        })
+        self.scheduler.handlers.update(
+            {
+                "event_wait": self.event_wait,
+                "event_set": self.event_set,
+                "event_clear": self.event_clear,
+                "event_is_set": self.event_is_set,
+            }
+        )
 
         self.scheduler.extensions["events"] = self
 
@@ -122,7 +124,7 @@ class EventExtension:
         with log_errors():
             name = self._normalize_name(name)
             # the default flag value is false
-            if not name in self._events:
+            if name not in self._events:
                 return False
 
             return self._events[name].is_set()
@@ -178,6 +180,7 @@ class Event:
     >>> event_2.set() # doctest: +SKIP
     >>> # now event_1 will stop waiting
     """
+
     def __init__(self, name=None, client=None):
         try:
             self.client = client or Client.current()
@@ -205,9 +208,7 @@ class Event:
         True if the event was set of false, if a timeout happend
         """
         result = self.client.sync(
-            self.client.scheduler.event_wait,
-            name=self.name,
-            timeout=timeout,
+            self.client.scheduler.event_wait, name=self.name, timeout=timeout,
         )
         return result
 
@@ -216,9 +217,7 @@ class Event:
 
         All waiters will now block.
         """
-        result = self.client.sync(
-            self.client.scheduler.event_clear, name=self.name,
-        )
+        result = self.client.sync(self.client.scheduler.event_clear, name=self.name,)
         return result
 
     def set(self):
@@ -226,16 +225,12 @@ class Event:
 
         All waiters will now be released.
         """
-        result = self.client.sync(
-            self.client.scheduler.event_set, name=self.name,
-        )
+        result = self.client.sync(self.client.scheduler.event_set, name=self.name,)
         return result
 
     def is_set(self):
         """ Check if the event is set """
-        result = self.client.sync(
-            self.client.scheduler.event_is_set, name=self.name,
-        )
+        result = self.client.sync(self.client.scheduler.event_is_set, name=self.name,)
         return result
 
     def __enter__(self):
--- a/distributed/event.py
+++ b/distributed/event.py
@@ -165,11 +165,11 @@ class Event:
     ----------
     name: string (optional)
         Name of the event.  Choosing the same name allows two
-        disconnected processes to coordinate an event.  If not given, a random
-        name will be generated.
+        disconnected processes to coordinate an event.  
+        If not given, a random name will be generated.
     client: Client (optional)
-        Client to use for communication with the scheduler.  If not given, the
-        default global client will be used.
+        Client to use for communication with the scheduler.  
+        If not given, the default global client will be used.
 
     Examples
     --------
--- a/distributed/event.py
+++ b/distributed/event.py
@@ -241,19 +241,5 @@ class Event:
         result = self.client.sync(self.client.scheduler.event_is_set, name=self.name,)
         return result
 
-    def __enter__(self):
-        self.set()
-        return self
-
-    def __exit__(self, *args, **kwargs):
-        self.clear()
-
-    async def __aenter__(self):
-        await self.set()
-        return self
-
-    async def __aexit__(self, *args, **kwargs):
-        await self.clear()
-
     def __reduce__(self):
         return (Event, (self.name,))
--- a/distributed/event.py
+++ b/distributed/event.py
@@ -87,7 +87,7 @@ class EventExtension:
 
             return True
 
-    async def event_set(self, comm=None, name=None):
+    def event_set(self, comm=None, name=None):
         """ Set the event with the given name to true.
 
         All waiters on this event will be notified.
@@ -97,9 +97,9 @@ class EventExtension:
             # No matter if someone is listening or not,
             # we set the event to true
             event = self._get_or_create_event(name)
-            self.scheduler.loop.add_callback(event.set)
+            event.set()
 
-    async def event_clear(self, comm=None, name=None):
+    def event_clear(self, comm=None, name=None):
         """Set the event with the given name to false."""
         with log_errors():
             name = self._normalize_name(name)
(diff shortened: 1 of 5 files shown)


distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+        """ Set the event with the given name to true.
+
+        All waiters on this event will be notified.
+        """
+        with log_errors():
+            name = self._normalize_name(name)
+            # No matter if someone is listening or not,
+            # we set the event to true
+            event = self._get_or_create_event(name)
+            self.scheduler.loop.add_callback(event.set)
```
mrocklin, line 100: Why not call this immediately?

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+            self._events[name] = asyncio.Event()
+
+        return self._events[name]
+
+    def _delete_event(self, name):
+        """ Helper function to delete an event """
+        with suppress(KeyError):
+            del self._waiter_count[name]
+        with suppress(KeyError):
+            del self._events[name]
```
mrocklin, line 153: Are these `suppress` contexts needed?  If so, why?

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+        return name
+
+    def _get_or_create_event(self, name):
+        """ Helper function to return or create and return an event """
+        if name not in self._events:
+            # It seems we are the first one accessing it,
+            # so lets create a new event
+            self._events[name] = asyncio.Event()
+
+        return self._events[name]
```
mrocklin, line 146: I *think* that `self._events = collections.defaultdict(asyncio.Event)`  would work here, but I could be wrong.

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+                # This can happen if an event is "double-cleared"
+                # In principle, the event should be unset at this point
+                # (because if it is set, all waiters should have been
+                # notified). But to prevent race conditions
+                # do to unlucky timing, we clear anyways
+                assert name in self._events
+                event = self._events[name]
+                self.scheduler.loop.add_callback(event.clear)
+
+    async def event_is_set(self, stream=None, name=None):
```
mrocklin, line 123: I think that we started calling these comm rather than stream a while ago.  I'm curious, where did you get the `stream=` name from?  

```suggestion
    async def event_is_set(self, comm=None, name=None):
```

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+            else:
+                # There are waiters
+                # This can happen if an event is "double-cleared"
+                # In principle, the event should be unset at this point
+                # (because if it is set, all waiters should have been
+                # notified). But to prevent race conditions
+                # do to unlucky timing, we clear anyways
+                assert name in self._events
+                event = self._events[name]
+                self.scheduler.loop.add_callback(event.clear)
```
mrocklin, line 121: If possible I would prefer running `event.clear` immediately.  Running it later is the sort of thing that can lead to race conditions.  Right now we know that we can safely clear it so great, let's clear it now rather than later.

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+        All waiters on this event will be notified.
+        """
+        with log_errors():
+            name = self._normalize_name(name)
+            # No matter if someone is listening or not,
+            # we set the event to true
+            event = self._get_or_create_event(name)
+            self.scheduler.loop.add_callback(event.set)
+
+    async def event_clear(self, stream=None, name=None):
```
mrocklin, line 102: It looks like this function doesn't have any `await`s in it, so we can probably make it a normal function.
```suggestion
    def event_clear(self, stream=None, name=None):
```

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+
+                if not self._waiter_count[name] and not event.is_set():
+                    # No one is waiting for this
+                    # and as the default flag for an event is false
+                    # we can safely remove it
+                    self._delete_event(name)
+
+            return True
+
+    async def event_set(self, stream=None, name=None):
```
mrocklin, line 90:
```suggestion
    def event_set(self, stream=None, name=None):
```

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+    The flag can be set to true (using the set() call) or back to false
+    (with the clear() call).
+    Every call to wait() blocks until the event flag is set to true.
+
+    Parameters
+    ----------
+    name: string (optional)
+        Name of the event.  Choosing the same name allows two
+        disconnected processes to coordinate an event.  If not given, a random
+        name will be generated.
```
mrocklin, line 169:
```suggestion
        disconnected processes to coordinate an event.  
        If not given, a random name will be generated.
```

This is perhaps pedantic, but I slightly prefer aligning sentences this way if it fits well.  I think it's a bit easier to read than having to deal with the line break in a sentence.

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+    client: Client (optional)
+        Client to use for communication with the scheduler.  If not given, the
+        default global client will be used.
```
mrocklin, line 172:
```suggestion
        Client to use for communication with the scheduler.  
        If not given, the default global client will be used.
```

Plus this way I get to play with multi-line suggestions :)

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+        --------
+        >>> event = Event('a')  # doctest: +SKIP
+        >>> event.wait(timeout=1)  # doctest: +SKIP
+
+        Returns
+        -------
+        True if the event was set of false, if a timeout happend
+        """
+        result = self.client.sync(
+            self.client.scheduler.event_wait, name=self.name, timeout=timeout,
```
mrocklin, line 211: We might want to use `dask.utils.parse_timedelta` here.  I notice that this wasn't done in Lock (where I think you are copying from).  If you're interested in doing it there too that would be welcome.

(although all of this could also be done in a follow up).

This would allow for functionality like `await event.wait(timeout="100 ms")`

distributed/event.py
```diff
@@ -0,0 +1,251 @@
...
+        )
+        return result
+
+    def clear(self):
+        """ Clear the event (set its flag to false).
+
+        All waiters will now block.
+        """
+        result = self.client.sync(self.client.scheduler.event_clear, name=self.name,)
+        return result
```
mrocklin, line 221:
```suggestion
        return self.client.sync(self.client.scheduler.event_clear, name=self.name)
```

distributed/tests/test_events.py
```diff
@@ -0,0 +1,165 @@
...
+    wait_futures = c.map(wait_for_it_ok, range(10))
+    await c.gather(wait_futures)
+
+    assert not s.extensions["events"]._events
+    assert not s.extensions["events"]._waiter_count
+
+
+def test_default_event(client):
+    event = Event("x")
+    assert not event.is_set()
```
mrocklin, line 53: With small tests like these we prefer using the async API if you're comfortable with it.  They are easier to debug if something goes wrong and they're also much faster to run, which helps to keep the overall test suite duration down.

I usually test everything in async mode, and then have one or two tests specifically to make sure that we're calling `client.sync` correctly.

distributed/tests/test_events.py
```diff
@@ -0,0 +1,165 @@
...
+    assert not s.extensions["events"]._waiter_count
+
+
+@gen_cluster(client=True)
+async def test_serializable(c, s, a, b):
+    def f(x, event=None):
+        assert event.name == "x"
+        return x + 1
+
+    event = Event("x")
```
mrocklin, line 127: I think that people may call `event = await Event()` in many cases (this is common with other coordination primitives).  Can I ask that we support this as well, even if it doesn't do anything?

distributed/event.py
nils-braun: From [lock.py](https://github.com/dask/distributed/blob/master/distributed/lock.py#L33) and [variable.py](https://github.com/dask/distributed/blob/master/distributed/variable.py#L42) which I basically used as blueprint (also for other design decisions). 
But no problem to change of course.

distributed/event.py
nils-braun: Yes, because a user could clear an event without ever having used it in waiting or something. Then the code has never created an `asyncio.Event` at all, so I need to suppress this. 
However, using your proposed `defaultdict` solution might make things easier.

distributed/event.py
nils-braun: Yes, makes a lot of sense!

distributed/event.py
nils-braun: Sorry, that was stupid. Using or not using a `defaultdict` does of course not change the fact... I would want to call this function without needing to check of the key is present before, therefore the suppression.

distributed/event.py
nils-braun: I was a bit "scared" that setting all the futures of the event might take too long and (sorry for using this always as an excuse) it is done like this in [lock.py](https://github.com/dask/distributed/blob/master/distributed/lock.py#L69) :-) 
But I guess it is fast enough to call it directly.

distributed/tests/test_events.py
nils-braun: Thanks for the suggestion! I did not understood this distinction in other tests in the first place, but it makes a lot of sense.

distributed/event.py
nils-braun: Yes, that looks good.
I have created #3823 and will do this after this PR.

distributed/tests/test_events.py
nils-braun: Sure!

nils-braun:
Thanks for the valuable feedback @mrocklin! I implemented your suggestions (or created an issue for them for a follow up).
I have one two last questions concerning usage patterns:

If a user calls something like `event = await Event("x")` would the user suggest it to be equivalent to ```python
event = Event("x")
event.wait()
```
or only to 
```
event = Event("x")
```
I had a look into the `queue` and the `semaphore` - but they await until the construct is registered with the server - something which I do not have to do here.

And: if a user calls 
```python
with Event("x") as e:
    some stuff
```
I currently assume the user means "set the event on enter" and "clear the event on exit". One could also argue that it should mean "wait for the event on enter". 
What is your opinion?

distributed/event.py
mrocklin: Raised here: https://github.com/dask/distributed/issues/3824

mrocklin:
> If a user calls something like event = await Event("x") would the user suggest it to be equivalent to ...

Only to `event = Event()`.  

> I had a look into the queue and the semaphore - but they await until the construct is registered with the server - something which I do not have to do here.

Yes, but users who are familiar with queue and semaphore might reasonably assume that events worked similarly and await them when creating them in async mode.  It might be good to mimic that behavior for consistency.

mrocklin:
> And: if a user calls

```python
with Event("x") as e:
    some stuff
```

> I currently assume the user means "set the event on enter" and "clear the event on exit". One could also argue that it should mean "wait for the event on enter".
> What is your opinion?

I recommend that we do whatever Python/asyncio does.  If they don't support this behavior then we shouldn't either.  If they do support it, we should follow those semantics.

nils-braun:
Ok, I have removed the contextmanager as `asyncio` also does not have it.
From my side, I am now finished with the PR.

distributed/tests/test_events.py
```diff
@@ -0,0 +1,220 @@
...
+
+    wait_futures = c.map(wait_for_it_ok, range(10))
+    await c.gather(wait_futures)
+
+    assert not s.extensions["events"]._events
+    assert not s.extensions["events"]._waiter_count
+
+
+@gen_cluster()
+async def test_default_event(s, a, b):
```
mrocklin, line 54:
```suggestion
@gen_cluster(client=True)
async def test_default_event(c, s, a, b):
```

I recommend including the client in these tests, otherwise it's a bit confusing about which client the `Event` is using by default.

In this case I suspect that this is only working because there is a worker in the same process, so the `get_worker()` path is saving us.  In principle though we shouldn't depend on this.  I recommend adding clients to most of the async tests here.

nils-braun: Sounds very reasonable, thanks.

mrocklin:
Thanks @nils-braun !  This is in.

PR #3822, opened 2020-05-22
Include total number of tasks in the performance report

### LINKED ISSUE #3753: [FEA] Include total number of tasks in the performance report summary
Dask performance reports can be invaluable when profiling complex workloads. When evaluating multiple workloads at once, it can be difficult to remember the size of the task graphs in each workload. Including the total number of tasks in the Summary tab of the performance report would be a nice addition, perhaps under the Duration section.

abduhbm:
This PR tries to close #3753
Still not sure if this is a correct approach, please let me know if more work is needed. 

Screen Shot 2020-05-22 at 4 24 51 AM

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5050,6 +5050,7 @@ class Scheduler(ServerNode):
 
         # Task stream
         task_stream = self.get_task_stream(start=start)
+        total_tasks = len(task_stream)
         from .diagnostics.task_stream import rectangles
         from .dashboard.components.scheduler import task_stream_figure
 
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @beckernick

quasiben:
I think this resolves the stated issue but I am wondering if this useful generally ? Perhaps instead we want a summary breakdown of how tasks spent their time.  Similar to https://github.com/dask/distributed/issues/3745 you could print out timings of various tasks actions: compute, transfer, disk-read, serialization, etc.

abduhbm:
Thanks @quasiben for your comment. 
Timings of tasks actions and total number of tasks are now printed out in the performance report under Tasks Information header. 

Screen Shot 2020-05-22 at 8 59 18 PM

quasiben:
Sorry for the delay here @abduhbm .  Thanks for doing this!  Merging in

Issue #3592, opened 2020-03-19
Easy way to restrict task and all dependencies to certain workers with dask.compute()


cbyrohl:
Hi there,

currently dask.compute(task, workers="worker1") will force "task" to run on "worker1" but not its dependencies. Is there an easy way to enforce this for the dependencies as well?

I guess there should be a hacky solution using task.dask.dependencies or task.dask.layers? Those return the names of the dependencies. If one can get the corresponding objects from those strings you could feed those into 'workers=(dependencies: "worker1")'.

Thanks,
Chris

mrocklin:
No, there is not currently a way to do this.  I agree with you though that it would be nice if there was.

jbouzaglou:
We are interested in this as well, it would really help in terms of debugging infrastructure related issues

mrocklin:
If anyone wants to resolve this issue, that would be welcome

PR #3826, opened 2020-05-23
Allow to pass in task key strings in the worker restrictions

nils-braun:
The `workers` parameter of `Client.compute` and `Client.persist` allows to restrict the workers where the calculations of the tasks should take place.

Advanced usages of this keyword (for example #3592) might need passing the task key
(string) instead of the dask collection or dask task. Internally, collections and task (lists) are turned into string keys anyways - so we just need to pass the key of the dict in this case.

This allows now to do things like:

```python
task = ... long dependencies ...
all_dependencies = task.dask.dependencies.keys()

workers = {d: worker_address for d in all_dependencies}
result = client.compute(last_task, workers=workers)
```

to have a task and all of its dependencies being calculated on a single worker.

This does not directly solve #3592, but at least makes it much easier to handle.

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -923,9 +923,9 @@ async def test_restrictions_ip_port_task_key(c, s, a, b):
     last_task = tasks[-1]
 
     # calculate all dependency keys
-    all_dependencies = list(last_task.__dask_graph__())
+    all_tasks = list(last_task.__dask_graph__())
     # only restrict to a single worker
-    workers = {d: a.address for d in all_dependencies}
+    workers = {d: a.address for d in all_tasks}
     result = c.compute(last_task, workers=workers)
     await result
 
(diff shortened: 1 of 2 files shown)


distributed/tests/test_client.py
```diff
@@ -913,6 +913,31 @@ async def test_restrictions_ip_port(c, s, a, b):
...
+async def test_restrictions_ip_port_task_key(c, s, a, b):
+    # Create a long dependency list
+    tasks = [delayed(inc)(1)]
+    for _ in range(100):
+        tasks.append(delayed(add)(tasks[-1], random.choice(tasks)))
+
+    last_task = tasks[-1]
+
+    # calculate all dependency keys
+    all_dependencies = list(last_task.dask.dependencies.keys())
```
mrocklin, line 926:
```suggestion
    all_dependencies = list(last_task.__dask_graph__())
```

This should be a more stable API to depend on.

mrocklin: And maybe also rename `all_dependencies` to `all_tasks`?

nils-braun:
Thanks @mrocklin - I implemented your suggestions.

PR #3829, opened 2020-05-26
Add notification of the origin of stack traces from other machines

### LINKED ISSUE #3828: Notify where exception originated
I probably spent an afternoon looking for a bug in the wrong place. When this function is executed an info is needed where the error originated. It is particular frustrating, because the tracebook actually reads the code from the local machine. It might be obvious for most, but it wasn't for me.

https://github.com/dask/distributed/blob/0041cc1f23fff4dd64c644c243ba8404122786c8/distributed/core.py#L602-L608

An easy fix would be to write something like this before line 605:

```python
print("Error originated from", comm.peer_address)
```

JulianWgs:
fix #3828

The problem is that this notification is only readable by the a human user and should rather be an exception. I don't know if anything from https://docs.python.org/3.8/library/traceback.html could help.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -602,6 +602,7 @@ async def send_recv(comm, reply=True, serializers=None, deserializers=None, **kw
     if isinstance(response, dict) and response.get("status") == "uncaught-error":
         if comm.deserialize:
             typ, exc, tb = clean_exception(**response)
+            print("Error originated from", comm.peer_addr)
             raise exc.with_traceback(tb)
         else:
             raise Exception(response["text"])


jacobtomlinson:
Perhaps a warning would be better?

mrocklin:
In general we don't print things from Dask.  Instead, I would expect this information to be included in the exception itself.  That way people can catch/interact with the exception and handle it however they like.  Printing is something that is rarely done in library code.  

> Perhaps a warning would be better?

Personally I'm against this as well.  I agree that it would be better.  I still don't think that it would be good though.

If we want to include this information then I think that we need to include it on the exception itself.  However, that becomes tricky because the exception is created by user code, and we probably want to be faithful to whatever their code creates.

jacobtomlinson:
Yeah those are good points. I wonder if we could extend the traceback somehow to indicate where it originated.

mrocklin:
Yeah, in general diving into the traceback would be great.  It would be
especially nice if we could also remove the Dask part of the traceback for
users, and then add a frame saying where it was coming from.

PR #3830, opened 2020-05-26
Record the time since the last run task on the scheduler

mrocklin:
This is used for the idle-timeout keyword.
This makes the logic a bit simpler,
and also gives more informative state.


quasiben:
The failures are unrelated. Thanks @mrocklin

PR #3832, opened 2020-05-27
Register `dask_cudf` serializers

jakirkham:
As `dask_cudf` simply relies on the same serialization machinery that is in `cudf` ( plus whatever is implemented by `dask_cudf` objects themselves, https://github.com/rapidsai/cudf/pull/5294 ), just reuse the `cudf` serialization import to handle `dask_cudf` object serialization for simplicity.

### DIFF
--- a/distributed/protocol/__init__.py
+++ b/distributed/protocol/__init__.py
@@ -105,6 +105,10 @@ def _register_rmm():
 @cuda_deserialize.register_lazy("cudf")
 @dask_serialize.register_lazy("cudf")
 @dask_deserialize.register_lazy("cudf")
+@cuda_serialize.register_lazy("dask_cudf")
+@cuda_deserialize.register_lazy("dask_cudf")
+@dask_serialize.register_lazy("dask_cudf")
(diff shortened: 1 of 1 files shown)


mrocklin:
This seems fine to me.

quasiben:
Thanks all.  Merging in

quasiben:
Oh, looks like @jakirkham needs to move out of draft mode

jakirkham:
Thanks all! Though I'm starting to second guess this change. Could use some help [understanding when pickling of collections is used]( https://stackoverflow.com/q/62054825/3877089 ) to better inform next steps.

jacobtomlinson:
Gah for some reason @jakirkham's last message didn't display when I hit merge. Only that it was marked as ready.

> Though I'm starting to second guess this change.

Do you want to revert it?

jakirkham:
No worries 🙂

I’m not aware of any downside to this change. At worst it simply doesn’t do anything.

I think we can leave it in. If we determine we don’t need it, we can always come back and cleanup.

Issue #3833, opened 2020-05-27
Rebalance with dask.array fails


RAMitchell:
Hi, I am trying to rebalance a dask array evenly across workers with the following script: 

```python
import dask.array as da
from dask.distributed import Client, wait, LocalCluster
import dask
import numpy as np


def main():
    print(dask.__version__)
    cluster = LocalCluster(n_workers=8)
    client = Client(cluster)
    n = 5000
    m = 10000
    X_np = np.random.random((m, n))
    X = da.from_array(X_np, chunks=(100, n)).persist()
    wait(X)
    client.rebalance()
    cluster.close()
    client.close()


if __name__ == "__main__":
    main()
```
This results in:
```python-traceback
2.16.0
[...]

PR #3834, opened 2020-05-27
Fix key check in rebalance missing keys

jacobtomlinson:
This check was removed in #3670 and seemed to be causing the exception in #3833.

There still seems to be a problem with rebalancing keys so #3833 is not fixed, but now the error message comes back with more useful information.

```python-traceback
<ipython-input-4-d4d2d986b89e> in <module>
----> 1 client.rebalance()

~/Projects/dask/distributed/distributed/client.py in rebalance(self, futures, workers, **kwargs)
   3117             A list of workers on which to balance, defaults to all workers
   3118         """
-> 3119         return self.sync(self._rebalance, futures, workers, **kwargs)
   3120 
   3121     async def _replicate(self, futures, n=None, workers=None, branching_factor=2):

~/Projects/dask/distributed/distributed/client.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    822         else:
    823             return sync(
--> 824                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
    825             )
    826 

~/Projects/dask/distributed/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    337     if error[0]:
    338         typ, exc, tb = error[0]
--> 339         raise exc.with_traceback(tb)
    340     else:
    341         return result[0]

~/Projects/dask/distributed/distributed/utils.py in f()
    321             if callback_timeout is not None:
    322                 future = asyncio.wait_for(future, callback_timeout)
--> 323             result[0] = yield future
    324         except Exception as exc:
    325             error[0] = sys.exc_info()

~/miniconda3/envs/dask/lib/python3.7/site-packages/tornado/gen.py in run(self)
    733 
    734                     try:
--> 735                         value = future.result()
    736                     except Exception:
    737                         exc_info = sys.exc_info()

~/Projects/dask/distributed/distributed/client.py in _rebalance(self, futures, workers)
   3095         if result["status"] == "missing-data":
   3096             raise ValueError(
-> 3097                 f"During rebalance {len(result['keys'])} keys were found to be missing"
   3098             )
   3099         assert result["status"] == "OK"

ValueError: During rebalance 1 keys were found to be missing
```

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3125,7 +3125,7 @@ class Scheduler(ServerNode):
                     return {
                         "status": "missing-data",
                         "keys": tuple(
-                            concat(r["keys"].keys() for r in result if "keys" in r)
+                            concat(r["keys"].keys() for r in result if r["status"] == "missing-data")
                         ),
                     }
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3125,7 +3125,11 @@ class Scheduler(ServerNode):
                     return {
                         "status": "missing-data",
                         "keys": tuple(
-                            concat(r["keys"].keys() for r in result if r["status"] == "missing-data")
+                            concat(
+                                r["keys"].keys()
+                                for r in result
+                                if r["status"] == "missing-data"
+                            )
                         ),
                     }
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3124,7 +3124,9 @@ class Scheduler(ServerNode):
                 if not all(r["status"] == "OK" for r in result):
                     return {
                         "status": "missing-data",
-                        "keys": tuple(concat(r["keys"].keys() for r in result)),
+                        "keys": tuple(
+                            concat(r["keys"].keys() for r in result if "keys" in r)
+                        ),
                     }
 
                 for sender, recipient, ts in msgs:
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -3124,7 +3124,9 @@ async def rebalance(self, comm=None, keys=None, workers=None):
                 if not all(r["status"] == "OK" for r in result):
                     return {
                         "status": "missing-data",
-                        "keys": tuple(concat(r["keys"].keys() for r in result)),
+                        "keys": tuple(
+                            concat(r["keys"].keys() for r in result if "keys" in r)
```
mrocklin, line 3128:
```suggestion
                            concat(r["keys"].keys() for r in result if r["status"] == "missing-data")
```

This seems like it might be more explicit / easier for future maintainers to understand what is going on

mrocklin:
> This check was removed in #3670 and seemed to be causing the exception in #3833.

Given that this was taken away before, is there some test that we can add to make sure that it doesn't go away in the future?

mrocklin:
Also, thanks for identifying and resolving this.

jacobtomlinson:
I did try writing a test. However the rebalance still fails, this was more about cleaning up the exception.

I'd be tempted to include a test for the rebalance in a future PR which will check both cases.

mrocklin:
OK.  Merging.  

Also, FYI, you had a small linting error.  It might be worth verifying that pre-commit hooks are installed.

jacobtomlinson:
> Also, FYI, you had a small linting error. It might be worth verifying that pre-commit hooks are installed.

I do have them, but I hit commit with your suggestion on GitHub in the GUI. I wonder if we could automate these kinds of small linting issues. Have a bot which pushes a lint fixing commit.

PR #3837, opened 2020-05-28
Fix daskboard nbytes colours

### LINKED ISSUE #3799: Proposal: Update colour scheme for `bytes stored` dashboard component

KrishanBhasin:
In making #3805 to fix #3799 , I left some bad copy-pastes in which result in the dashboard pane staying blue.

I'm not sure whats wrong with my local setup (or, indeed, this code), but I cannot seem to get the colours to _ever_ change from blue.

cc @jacobtomlinson


KrishanBhasin:
I'm not sure what I'm doing wrong, but when I use 
```python
def dump(obj):
  for attr in dir(obj):
    print("obj.%s = %r" % (attr, getattr(obj, attr)))
```
to inspect the workers, I can't see any attributes related to memory other than `memory_limit`, so the bars stay perpetually blue - even when attempting to set them via:

```python
dask.config.set({"distributed.worker.memory.spill":0.15})
```
or modifying the `~/.config/dask/distributed.yml` file.

Looking at the Worker code, I think that I should expect to be able to look up the attributes?
https://github.com/dask/distributed/blob/9db7368bae6f5598324b2b0c8beaf292dfd382fc/distributed/worker.py#L516-L533

Can anyone help me spot what I am overlooking?

PR #3838, opened 2020-05-28
Fix typo in journey of a task documentation

jrbourbeau:
I think this dates back to the `Executor` days : )


mrocklin:
They were good times :)

PR #3840, opened 2020-05-29
Wrong variable name example.

Carreau:
Spotted by James Bourbeau during a demo of dask/distributed.


Carreau:
cc @jrbourbeau

Carreau:
Yes, happy to be back and doing more Python. Slowly ramping up. Will be a bit more on the Zarr side.

jrbourbeau:
Woo, thanks @Carreau

PR #3842, opened 2020-05-31
Validate address parameter in client constructor

### LINKED ISSUE #3839: Add helpful error message when passing a bad argument to Client
Today I made a typo which resulted in an unhelpful error message. I created a `Client` from a `LocalCluster` but forgot to actually instantiate the `LocalCluster`.

```python
from dask.distributed import LocalCluster, Client
cluster = LocalCluster  # Note the missing `()`
client = Client(cluster)
```

This results in a `TypeError` complaining about expecting a string but get a property.

```python-traceback
<ipython-input-3-82a29784114b> in <module>
      1 cluster = LocalCluster
----> 2 client = Client(cluster)

~/Projects/dask/distributed/distributed/client.py in __init__(self, address, loop, timeout, set_as_default, scheduler_file, security, asynchronous, name, heartbeat_interval, serializers, deserializers, extensions, direct_to_workers, connection_limit, **kwargs)
    734             ext(self)
    735 
--> 736         self.start(timeout=timeout)
    737         Client._instances.add(self)
    738 

~/Projects/dask/distributed/distributed/client.py in start(self, **kwargs)
    938             self._started = asyncio.ensure_future(self._start(**kwargs))
    939         else:
--> 940             sync(self.loop, self._start, **kwargs)
    941 
    942     def __await__(self):

~/Projects/dask/distributed/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    337     if error[0]:
    338         typ, exc, tb = error[0]
--> 339         raise exc.with_traceback(tb)
    340     else:
    341         return result[0]

~/Projects/dask/distributed/distributed/utils.py in f()
    321             if callback_timeout is not None:
    322                 future = asyncio.wait_for(future, callback_timeout)
--> 323             result[0] = yield future
    324         except Exception as exc:
    325             error[0] = sys.exc_info()

~/miniconda3/envs/dask/lib/python3.7/site-packages/tornado/gen.py in run(self)
    733 
    734                     try:
--> 735                         value = future.result()
    736                     except Exception:
    737                         exc_info = sys.exc_info()

~/Projects/dask/distributed/distributed/client.py in _start(self, timeout, **kwargs)
   1031 
   1032         if self.scheduler is None:
-> 1033             self.scheduler = self.rpc(address)
   1034         self.scheduler_comm = None
   1035 

~/Projects/dask/distributed/distributed/core.py in __call__(self, addr, ip, port)
    915     def __call__(self, addr=None, ip=None, port=None):
    916         """ Cached rpc objects """
--> 917         addr = addr_from_args(addr=addr, ip=ip, port=port)
    918         return PooledRPCCall(
    919             addr, self, serializers=self.serializers, deserializers=self.deserializers

~/Projects/dask/distributed/distributed/core.py in addr_from_args(addr, ip, port)
    616     if isinstance(addr, tuple):
    617         addr = unparse_host_port(*addr)
--> 618     return normalize_address(addr)
    619 
    620 

~/Projects/dask/distributed/distributed/comm/addressing.py in normalize_address(addr)
     52     'tcp://[::1]'
     53     """
---> 54     return unparse_address(*parse_address(addr))
     55 
     56 

~/Projects/dask/distributed/distributed/comm/addressing.py in parse_address(addr, strict)
     19     """
     20     if not isinstance(addr, str):
---> 21         raise TypeError("expected str, got %r" % addr.__class__.__name__)
     22     scheme, sep, loc = addr.rpartition("://")
     23     if strict and not sep:

TypeError: expected str, got 'property'
```

It would be nice to catch this and raise a more helpful error about the kind of things that can be passed to `Client`. Perhaps even test if the thing passed in is a `Cluster` class rather than an instance of the class.

joshreback:
Closes #3839 

-Validate address parameter.
-Provide helpful error message if `address` is a `Cluster` class rather than an instance of the class.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -657,7 +657,7 @@ class Client:
                 loop = address.loop
             if security is None:
                 security = getattr(self.cluster, "security", None)
-        elif not isinstance(address, str):
+        elif address is not None and not isinstance(address, str):
             raise ValueError(
                 "Scheduler address must be a string or a Cluster instance, got {}".format(
                     address
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -660,7 +660,7 @@ class Client:
         elif address is not None and not isinstance(address, str):
             raise ValueError(
                 "Scheduler address must be a string or a Cluster instance, got {}".format(
-                    address
+                    type(address)
                 )
             )
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -648,24 +648,21 @@ class Client:
                 "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
             )
 
-        from .deploy import Cluster
-
         if isinstance(address, (rpc, PooledRPCCall)):
             self.scheduler = address
-        elif isinstance(address, Cluster):
+        elif isinstance(getattr(address, "scheduler_address", None), str):
             # It's a LocalCluster or LocalCluster-compatible object
             self.cluster = address
             with suppress(AttributeError):
                 loop = address.loop
             if security is None:
                 security = getattr(self.cluster, "security", None)
-        elif address is not None and not isinstance(address, str):
-            if issubclass(address, Cluster):
-                raise ValueError(
-                    "address is a subclass of Cluster rather than an instance.\n"
-                    "Did you forget to instantiate the cluster?"
+        elif not isinstance(address, str):
+            raise ValueError(
+                "Scheduler address must be a string or a Cluster instance, got {}".format(
+                    address
                 )
-            raise ValueError("address must be a string or Cluster instance.")
+            )
 
         if security is None:
             security = Security()
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -648,15 +648,24 @@ class Client:
                 "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
             )
 
+        from .deploy import Cluster
+
         if isinstance(address, (rpc, PooledRPCCall)):
             self.scheduler = address
-        elif hasattr(address, "scheduler_address"):
+        elif isinstance(address, Cluster):
             # It's a LocalCluster or LocalCluster-compatible object
             self.cluster = address
             with suppress(AttributeError):
                 loop = address.loop
             if security is None:
                 security = getattr(self.cluster, "security", None)
+        elif address is not None and not isinstance(address, str):
+            if issubclass(address, Cluster):
+                raise ValueError(
+                    "address is a subclass of Cluster rather than an instance.\n"
+                    "Did you forget to instantiate the cluster?"
+                )
+            raise ValueError("address must be a string or Cluster instance.")
 
         if security is None:
             security = Security()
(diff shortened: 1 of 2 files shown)


distributed/client.py
```diff
@@ -648,15 +648,24 @@ def __init__(
                 "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
             )
 
+        from .deploy import Cluster
+
         if isinstance(address, (rpc, PooledRPCCall)):
             self.scheduler = address
-        elif hasattr(address, "scheduler_address"):
+        elif isinstance(address, Cluster):
```
mrocklin, line 655: We don't require that downstream implementations of Dask Cluster types inherit from our type.  Perhaps maybe you want something like `isinstance(getattr(address, "scheduler_address", None), str)`?

distributed/client.py
```diff
@@ -648,15 +648,24 @@ def __init__(
...
             # It's a LocalCluster or LocalCluster-compatible object
             self.cluster = address
             with suppress(AttributeError):
                 loop = address.loop
             if security is None:
                 security = getattr(self.cluster, "security", None)
+        elif address is not None and not isinstance(address, str):
+            if issubclass(address, Cluster):
```
mrocklin, line 663: The second condition seems to imply the first.  Also, the second instance seems very specific.  I wonder if maybe this could be more broad like the following:

```python
elif not isinstance(address, str):
    raise ValueError("Scheduler address must be a string or a Cluster instance, got {}".format(address))
```

joshreback: thank you for the feedback. I made the error message more broad, as suggested, but had to also check that `address is not None` otherwise this would raise an error when address was None which caused some tests to fail. I'm not sure if it still makes sense for `address` to be None at this point in the setup, but I just preserved the behavior that was expected by the failing tests.

joshreback:
hmm, i see the build failed, but I cannot reproduce the failure locally and I don't see a way to retry the build. Not sure how to proceed 😕

jrbourbeau:
Thanks @joshreback. That was a known flaky test, I've restarted the failing CI build

distributed/client.py
```diff
@@ -650,13 +650,19 @@ def __init__(
...
             # It's a LocalCluster or LocalCluster-compatible object
             self.cluster = address
             with suppress(AttributeError):
                 loop = address.loop
             if security is None:
                 security = getattr(self.cluster, "security", None)
+        elif address is not None and not isinstance(address, str):
+            raise ValueError(
+                "Scheduler address must be a string or a Cluster instance, got {}".format(
+                    address
```
jacobtomlinson, line 663: Maybe we just want to report back to the user the type of object we were passed, not the actual object.

```suggestion
                    type(address)
```

joshreback: good thought - pushed the recommended change and the CI build now passes.

mrocklin:
Thanks @joshreback .  The implementation here looks good to me.  If you have time it would also be helpful to add a test case for the original bad behavior that we're trying to handle.  Maybe something like the following?

```python
def test_informative_error_on_cluster_type():
    with pytest.raises(TypeError) as info:
        Client(LocalCluster)

    assert things about info.value
```

mrocklin:
Great.  Thanks @joshreback !  This is in.

PR #3844, opened 2020-05-31
install vanilla status route if bokeh dependency is not satisfied

### LINKED ISSUE #3836: Display informative message if attempting to use the dashboard and bokeh isn't installed
From https://github.com/dask/distributed/issues/3741#issuecomment-635396442:

> Perhaps we should include a vanilla route under `/status` (the common
entry point for Bokeh users) that says "please install bokeh" with some
conda/pip install lines if Bokeh isn't installed.  The relevant code is
probably in distributed/http/scheduler/info.py .  I suspect that someone
even very mildly familiar with web programming could accomplish this by
copying examples that are already there.

joshreback:
Fixes #3836 

Not entirely sure how to test this..

### DIFF
--- a/distributed/http/scheduler/info.py
+++ b/distributed/http/scheduler/info.py
@@ -189,6 +189,16 @@ class EventstreamHandler(WebSocketHandler):
         self.server.remove_plugin(self.plugin)
 
 
+class MissingBokeh(RequestHandler):
+    def get(self):
+        with log_errors():
+            self.write(
+                "<p>Dask needs bokeh >= 0.13.0 for the dashboard.</p>"
+                "<p>Install with conda: conda install bokeh>=0.13.0</p>"
+                "<p>Install with pip: pip install bokeh>=0.13.0</p>"
+            )
+
+
 routes = [
     (r"info", redirect("info/main/workers.html"), {}),
     (r"info/main/workers.html", Workers, {}),
@@ -201,3 +211,8 @@ routes = [
     (r"individual-plots.json", IndividualPlots, {}),
     (r"eventstream", EventstreamHandler, {}),
 ]
+
+try:
+    import distributed.dashboard.scheduler  # noqa: F401
+except ImportError:
+    routes.append((r"status", MissingBokeh, {}))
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
Tested locally on my Mac in a fesh conda environment and it seems to be working.

image

I'm not sure if the CI failures are related. I've restarted them.

I also noticed that just visiting http://localhost:8787/ results in a 404. I personally would expect this to still redirect to `/status` and then show the bokeh warning.

Is the failure to redirect also a result of Bokeh missing?

joshreback:
@jacobtomlinson oh good point. I didn't think of the root route. I'll need to append that route - in addition to `/status` - to the tornado server to show this page if bokeh is not installed.

joshreback:
I pushed a new commit that shows this page for both `/status` and `/` if bokeh is not installed...I'm not sure what to make of the CI failure though. The failure - `distributed\cli\tests\test_dask_worker.py:135: AssertionError`- didn't fail when i ran that test locally.

mrocklin:
In principle this seems fine to me.  Thank you @joshreback .

The test failures here are odd.  They don't seem related, but they are popping up more frequently than I would expect.  I'm sending up a test PR here to see if this is on master.  https://github.com/dask/distributed/pull/3850

jacobtomlinson:
It does seem to be `test_no_nanny` which is consistently failing on Windows with this change.

joshreback:
The `test_no_nanny` test failed consistently on Windows w/Python version 3.6. I could not track down exactly why but it seemed to be due to double-importing `distributed.dashboard.scheduler`.  

So I reworked the approach a bit to put the backup status routes in a different file, and only install those URLs if the dashboard params were set and bokeh was not installed.

joshreback:
Hi @mrocklin / @jacobtomlinson just wanted to follow up to see if there's any feedback on this one. Looks like it is now passing in CI.

mrocklin:
Hi!  Sorry about the delay.  I'll take a look in the next 24 hours.

quasiben:
Also tested and works for me.  I'll merge in tomorrow morning unless @mrocklin has more thoughts

joshreback:
@quasiben anything further to do here?

quasiben:
Thank you for the ping.  Sorry, merging now.

PR #3847, opened 2020-06-02
Allow async add_worker remove_worker plugin methods

jrbourbeau:
This PR is to add support for asynchronous `SchedulerPlugin.add_worker` and `SchedulerPlugin.remove_worker` methods (which are currently supported for `start` and `close`)

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -516,6 +516,7 @@ class Server:
                             handler = self.stream_handlers[op]
                             if is_coroutine_function(handler):
                                 self.loop.add_callback(handler, **merge(extra, msg))
+                                await gen.sleep(0)
                             else:
                                 handler(**merge(extra, msg))
                         else:
--- a/distributed/diagnostics/tests/test_scheduler_plugin.py
+++ b/distributed/diagnostics/tests/test_scheduler_plugin.py
@@ -70,6 +70,41 @@ async def test_add_remove_worker(s):
     assert events == []
 
 
+@gen_cluster(nthreads=[], client=False)
+async def test_async_add_remove_worker(s):
+    events = []
+
+    class MyPlugin(SchedulerPlugin):
+        async def add_worker(self, worker, scheduler):
+            assert scheduler is s
+            events.append(("add_worker", worker))
+
+        async def remove_worker(self, worker, scheduler):
+            assert scheduler is s
+            events.append(("remove_worker", worker))
+
+    plugin = MyPlugin()
+    s.add_plugin(plugin)
+    assert events == []
+
+    async with Worker(s.address) as a:
+        async with Worker(s.address) as b:
+            pass
+
+    assert set(events) == {
+        ("add_worker", a.address),
+        ("add_worker", b.address),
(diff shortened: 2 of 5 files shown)


distributed/core.py
```diff
@@ -516,6 +516,7 @@ async def handle_stream(self, comm, extra=None, every_cycle=[]):
                             handler = self.stream_handlers[op]
                             if is_coroutine_function(handler):
                                 self.loop.add_callback(handler, **merge(extra, msg))
+                                await gen.sleep(0)
```
jrbourbeau, line 519: For reference, when looping through messages here:

https://github.com/dask/distributed/blob/a6db3a1967b936d8942cd1c7b98bd683f117e37d/distributed/core.py#L508

handlers which are synchronous functions are executed in order, while coroutine handlers are added to the event loop as a callback then executed in reserve order. This causes issues with our task stealing code as we ended up releasing keys before they were stolen.

To avoid executing operations in reverse order I added `gen.sleep(0)` to force the next I/O loop iteration where the most recently added callback is run. I _think_ this is okay.

For my particular use case there's a workaround where I no longer need asynchronous `Scheduler.add_worker`/`.remove_worker` methods. That said, I think the changes here are still an improvement to the current situation

mrocklin:
To me this seems like it might have been an exercise in yak-shaving.  A simple objective ended up having many knock-on effects.  However, all of those effects seem pretty ok to me, and I think that we have a nicely shaved yak here :)  +1 from me.

I suggest that we let this sit here for 24 hours before merging (unless you're in a rush).

jrbourbeau:
Thanks for taking a look! Letting this sit for a bit is totally fine, no rush from me

distributed/core.py
pentschev: We have a few tests in dask-cuda that check the behavior of Device<->Host<->Disk spilling and I noticed after the 2.19 release one of them has broken, I managed to track it down to this line specifically. The test in question happens in https://github.com/rapidsai/dask-cuda/blob/branch-0.15/dask_cuda/tests/test_spill.py#L409-L411, where we assert that the zict dictionaries are empty after deleting `cdf2`, which is the object being spilled. It seems that this is because we're not awaiting for `Worker.delete_data` somewhere, as per the warning below that doesn't happen if I comment `await gen.sleep(0)` out:

```python
dask_cuda/tests/test_spill.py::test_cudf_device_spill[params0]
  /datasets/pentschev/miniconda3/envs/r-102-0.14/lib/python3.7/inspect.py:732: RuntimeWarning: coroutine 'Worker.delete_data' was never awaited
    for modname, module in list(sys.modules.items()):
```

I _think_ that the only place where `Worker.delete_data` would be called and should be awaited is in https://github.com/dask/distributed/blob/4f878b420b349ee725de5ef64fd5e664dedb8aba/distributed/scheduler.py#L2791-L2800, but I don't have anything better than my guess at this time because it's really hard for me to understand all the async black magic. I'm gonna continue trying to figure this out, but any suggestions on how to pinpoint that are appreciated!

PR #3848, opened 2020-06-02
Allow proxying dashboard for any current worker (#3846)

### LINKED ISSUE #3846: Dashboard cannot proxy Worker Dashboards in distributed cluster
**What happened**:

Set up a distributed Dask cluster (1 scheduler and 4 workers) each running on different hosts and accessed the dashboard:

Screenshot 2020-06-01 at 17 42 17

Then tried to access the Worker Dashboards via the Info tab and received the following error:

Screenshot 2020-06-02 at 09 44 37

Expected to be able to access the worker dashboards successfully

On initial attempt I got a different error advising that I needed to install the additional `jupyter-sever-proxy` package to enable worker dashboard proxying.  However once installed you get the above error because the `jupyter-server-proxy` package enforces a whitelist that only has `localhost` and `127.0.0.1` in it by default.

There does not appear to be any way for users to customise this whitelist so it is hard to see how this feature is ever expected to work except for a fully local cluster i.e. all workers on same physical machine as the scheduler.

Ideally the Dask scheduler dashboard should automatically update the proxy whitelist whenever a new worker is registered.

**Environment**:

- Dask version: 2.17.0
- Python version: 3.8
- Operating System: OS X
- Install method: conda

rvesse:
The default behaviour of `jupyter-server-proxy` that is used by the
dashboard to proxy the worker dashboards is to whitelist only `localhost`
and `127.0.0.1`.  This means that in a distributed cluster the scheduler
dashboard can never successfully proxy worker dashboards unless those
workers happen to be on the same physical host as the scheduler and
listening on `localhost/127.0.0.1`

This commit changes the proxy handler to set a whitelist checking method
that verifies the requested hostname is a currently registered worker.  This
takes advantage of the fact that `jupyter-server-proxy` allows the whitelist
to be a function and thus on every request we can verify that the hostname
we are trying to proxy for is a current worker.  This avoids needing to 
manually configure the whitelist and ensures that the whitelist check is always
correct wrt. the current state of the Dask cluster.

Closes #3846

### DIFF
--- a/distributed/http/proxy.py
+++ b/distributed/http/proxy.py
@@ -13,9 +13,42 @@ try:
         from a port to any valid endpoint'.
         """
 
+        def is_worker_host(self, proxy_handler, host=None):
+            """
+            Check whether host is a valid worker hostname
+
+            This is necessary to ensure that the Dashboard can actually proxy requests
+            to worker dashboards when running a distributed cluster where each worker
+            may be on a different hostname than the scheduler host
+
+            Parameters
+            ----------
+
+            host : str
+
+            Returns
+            -------
+            bool
+            """
+            if host is None:
+                return False
+            if self.scheduler is None:
+                return False
+            workers = list(self.scheduler.workers.values())
+            for w in workers:
+                if w.host == host:
+                    return True
+            return False
+
         def initialize(self, dask_server=None, extra=None):
             self.scheduler = dask_server
             self.extra = extra or {}
+
+            # Customise the host whitelist checking function
+            # Otherwise jupyter-server-proxy will only allow proxying
+            # localhost/127.0.0.1 which effectively means we can't proxy
+            # worker dashboards in distributed clusters
+            self.host_whitelist = self.is_worker_host
 
         async def http_get(self, port, host, proxied_path):
             # route here first


distributed/http/proxy.py
```diff
@@ -13,10 +13,43 @@ class GlobalProxyHandler(ProxyHandler):
...
+
+            Returns
+            -------
+            bool
+            """
+            if host is None:
+                return False
+            if self.scheduler is None:
+                return False
+            workers = list(self.scheduler.workers.values())
```
dhirschfeld, line 37: If all you need is an iterable I think the `list` is redundant:
```suggestion
            workers = self.scheduler.workers.values()
```

mrocklin: Or you could inline this with the line below

```python
for w in self.scheduler.workers.values():
```

This is how most of the other code looks so far

mrocklin: Actually it looks like you're maybe looking for the following

```python
return any(w.host == host for w in self.scheduler.workers.values())
```

distributed/http/proxy.py
```diff
@@ -13,10 +13,43 @@ class GlobalProxyHandler(ProxyHandler):
...
+
         def initialize(self, dask_server=None, extra=None):
             self.scheduler = dask_server
             self.extra = extra or {}
 
+            # Customise the host whitelist checking function
+            # Otherwise jupyter-server-proxy will only allow proxying
+            # localhost/127.0.0.1 which effectively means we can't proxy
+            # worker dashboards in distributed clusters
+            self.host_whitelist = self.is_worker_host
```
mrocklin, line 51: Should we define the `is_worker_host` function above as just `host_whitelist`?

mrocklin:
What's up with the function `check_worker_dashboard_exits` ?  Is that not doing the same job here?  Was it not working before?  If so, should it be removed?

mrocklin:
Thanks for taking this on @rvesse .  I appreciate it.

rvesse:
@mrocklin Yes I will try and clean this up likely next week.

We have a big feature freeze release milestone in $dayjob this week so that's taking priority

PR #3853, opened 2020-06-03
Use Enum for worker, scheduler and nanny status.

Carreau:
Looking at the code it was hard to find all the states a worker, nanny
scheduler could be
in, and in particular it is easy to forget 'closing_gracefully' for
example when skimming over a `self.status.startswith('clos')`, and it
also hard to fine that in some places the status can be "None", "init",
"starting" and "stopped"

So I just took a pass at seeing in how many places status was
assigned/compared in worker, nanny and scheduler and replace that by an
Enum I'm assuming the status each can see is not shared  so we end up

I thus tried to do separate Status Enums for Scheduler/Nanny/Worker.

Note that even if Enums have values for their variants, the variants _do
not_ compare equal to their values. That is to say `Status.running =!
'running'`, but `Status.running.value == 'running'`,

So I implemented a custom `__eq__` method that can compare with
strings/None for backward compatibility. It works both when the LHS or
RHS is str/None as the default impl of str/None forward to our
implementation. This also make sure the value you compare with is also
in the list of possible state.

The main other code changes were updating some of the comparisons to use
the enums instead of the strings directly, and replace the "startswith"
with the `in` operator, which is slightly more verbose.

I believe despite being a bit more verbose, it is a bit more explicit and
may help navigate the codebase and if you want to go this route you
probably want to remove the custom comparison but that would be an API
change that could break other project.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -64,7 +64,7 @@ class Status(Enum):
         if isinstance(other, str) or (other is None):
             assert other in [s.value for s in type(self)]
             return other == self.value
-        elif isinstance(other, type(self)):
+        elif isinstance(other, Enum):
             return self.value == other.value
         raise TypeError(
             f"'==' not supported between instances of"
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -42,16 +42,21 @@ from . import protocol
 
 class Status(Enum):
     """
-    This Enum contains the various states a worker can be.
-    Those states can be observed and used in worker, scheduler and nanny.
-
+    This Enum contains the various states a worker, scheduler and nanny can be
+    in. Some of the status can only be observed in one of nanny, scheduler or
+    worker but we put them in the same Enum as they are compared with each
+    other.
     """
 
-    undefined = None
-    running = "running"
     closed = "closed"
     closing = "closing"
     closing_gracefully = "closing-gracefully"
+    init = "init"
+    running = "running"
+    starting = "starting"
+    stopped = "stopped"
+    stopping = "stopping"
+    undefined = None
 
     def __eq__(self, other):
         """
@@ -59,13 +64,15 @@ class Status(Enum):
 
         If other object instance is string, we compare with the values, but we
         actually want to make sure the value compared with is in the list of
-        possible Status.
+        possible Status, this avoid comparison with non-existing status.
         """
-        if isinstance(other, str) or (other is None):
-            assert other in [s.value for s in type(self)]
-            return other == self.value
-        elif isinstance(other, Enum):
+        if isinstance(other, type(self)):
             return self.value == other.value
+        elif isinstance(other, str) or (other is None):
+            assert other in [
+                s.value for s in type(self)
+            ], f"comparison with non-existing states {other}"
+            return other == self.value
         raise TypeError(
             f"'==' not supported between instances of"
             f" {type(self).__module__+'.'+type(self).__qualname__!r} and"
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -1,6 +1,7 @@
 import asyncio
 from collections import defaultdict, deque
 from contextlib import suppress
+from enum import Enum
 from functools import partial
 import inspect
 import logging
@@ -37,6 +38,39 @@ from .utils import (
     TimeoutError,
 )
 from . import protocol
+
+
+class Status(Enum):
+    """
+    This Enum contains the various states a worker can be.
+    Those states can be observed and used in worker, scheduler and nanny.
+
+    """
+
+    undefined = None
+    running = "running"
+    closed = "closed"
+    closing = "closing"
+    closing_gracefully = "closing-gracefully"
+
+    def __eq__(self, other):
+        """
+        Implement equality checking with backward compatibility.
+
+        If other object instance is string, we compare with the values, but we
+        actually want to make sure the value compared with is in the list of
+        possible Status.
+        """
+        if isinstance(other, str) or (other is None):
+            assert other in [s.value for s in type(self)]
+            return other == self.value
+        elif isinstance(other, type(self)):
+            return self.value == other.value
+        raise TypeError(
+            f"'==' not supported between instances of"
+            f" {type(self).__module__+'.'+type(self).__qualname__!r} and"
+            f" {type(other).__module__+'.'+type(other).__qualname__!r}"
+        )
 
 
 class RPCClosed(IOError):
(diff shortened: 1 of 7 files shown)


Carreau:
Mostly an experiment/suggestion, I'm perfectly ok if you are not open to such a change. 

This was mostly inspired by trying to figure out if it is possible to "drain" the workers, and wait until they are in a idle state, to try to fix the issue when retiring worker cannot replicate all the keys and fails. 

I ended up trying to list all the "status" and where they were used and this helped  me a bit.

I think I would also probably be more explicit with "WorkerStatus", "SchedulerStatus" and "NannyStatus", unless some of these states need to be unions of the current ones. 

The other routes would have been properties, though I like the explicitness of Enums.

distributed/scheduler.py
```diff
@@ -5535,6 +5568,7 @@ class WorkerStatusPlugin(SchedulerPlugin):
     """
 
     def __init__(self, scheduler, comm):
+        self._status = None
```
Carreau, line 5571:
```suggestion
```

Carreau:
Actually it seem from the more extensive testsuite than the one I ran the actually values get compared between Worker and Scheduler. So I'll have to revise this.

mrocklin:
In principle an approach like this seems sensible to me.  It probably makes sense to make things a bit more formal today.  This might be a good opportunity for us to think about our statuses more broadly and see if there are opportunities to unify them (it would be nice to have only one Status type).

cc'ing @jcrist , who I think will be made happy by this PR

jcrist:
>  This might be a good opportunity for us to think about our statuses more broadly and see if there are opportunities to unify them (it would be nice to have only one Status type).

If possible, I agree this would be a good thing to do. The scheduler, worker, and nanny all should have similar states they transition through during execution. The `Cluster` objects also have a status field, but those may have slightly different states.

mrocklin:
> If possible, I agree this would be a good thing to do. The scheduler, worker, and nanny all should have similar states they transition through during execution. The Cluster objects also have a status field, but those may have slightly different states.

My hope is that we can have one set of states that is consistently used across all components.  That would also solve this:

> Actually it seem from the more extensive testsuite than the one I ran the actually values get compared between Worker and Scheduler. So I'll have to revise this.

Carreau:
Thanks, I wasn't sure whether the states were supposed to be shared or at least have a common subset. We Could have a core Enum, and deved enums is my guess (though they will need to not inherit from Enum), and/or restrict the stated on Scheduler/Nany/Worker by making status a property. Let me update that with a single unified status type first and then let's see if we can shave some states or refactor.

mrocklin:
@lesteve @jacobtomlinson @jcrist thoughts on applying this same change to downstream dask-foo cluster managers?

jcrist:
Dask-Gateway has a status field that's mostly for debouncing close operations - would be fine to use the enum here instead. Dask-Yarn currently has no status field, state is tracked via a couple internal event flags.

mrocklin:
It looks like dask-kubernetes doesn't modify status, but dask-cloudprovider
does.  I wonder if we might be able to handle some of this by using things
like `super().start()` and `super().close()` and relying on behavior there.

Carreau:
A safe first step would be to make all of those `status` fields properties which convert from strings to the corresponding enum when set and emit a warning, that would be completely backward compatible, and non invasive.

Carreau:
Which is also relatively easy as they are all subclass of `Server`.

Carreau:
Can one of you kick the failing travis build ?

lesteve:
> @lesteve @jacobtomlinson @jcrist thoughts on applying this same change to downstream dask-foo cluster managers?

Fine with me, we rely on `SpecCluster` heavily so that would not affect `dask-jobqueue` clusters too much. A quick `git grep` shows `.status` is used twice in all of Dask-Jobqueue (and one of these two occurences happens in a test).

jacobtomlinson:
> @lesteve @jacobtomlinson @jcrist thoughts on applying this same change to downstream dask-foo cluster managers?

Yeah I support making these changes downstream. And also pulling dask-kubernetes inline.

distributed/core.py
```diff
@@ -39,6 +40,46 @@
...
+
+    closed = "closed"
+    closing = "closing"
+    closing_gracefully = "closing-gracefully"
+    init = "init"
+    running = "running"
+    starting = "starting"
+    stopped = "stopped"
+    stopping = "stopping"
+    undefined = None
```
mrocklin, line 59: It looks like closing/closed and stopping/stopped might be redundant.  I would be in favor of unifying around one pair of values, presumably closing/closed.

mrocklin: I don't have a sense for how often users check these values and if there is a way that we should make this change smoothly.  I'm inclined to make a small breaking change here, just because I don't think that many people depend on these values.

distributed/nanny.py
```diff
@@ -56,7 +57,7 @@ class Nanny(ServerNode):
 
     _instances = weakref.WeakSet()
     process = None
-    status = None
+    status = Status.undefined
```
mrocklin, line 60: I wonder if `Status.undefined` and `Status.init` should be unified.

Carreau:
Do you want to do all the unification as part of this PR, or do you want to get something smaller in master and then bikesheb on which one is renamed how later ? 

My guess is that the renaming/dropping is backward incompatible and easier to discuss on its own/revert if necessary.

distributed/core.py
Carreau: There are many ways of handling this, in the properties you can emit a warning if you get a given value assigned tell users to use the Enum, and same in the `__eq__` method, if you see comparison with the deprecated values, you warn. 

Those warnings can be turned into error in tests. 

You can also help refactor with having two variants of the enum map to the same value.

mrocklin:
> Do you want to do all the unification as part of this PR, or do you want to get something smaller in master and then bikesheb on which one is renamed how later ?

> My guess is that the renaming/dropping is backward incompatible and easier to discuss on its own/revert if necessary.

Good point.  I'm happy to continue on this PR without these changes.

mrocklin:
I think that there is a release in progress today.  Let's let that pass through and then we can merge this and let it sit in master for a while to see if any issues arise.

Carreau:
No problem, and no hurry.

mrocklin:
OK, merging this in 24 hours if there are no further comments.

Carreau:
Thanks, sounds great, once done I can change in more places and start adding warnings.

jacobtomlinson:
Once this is in we should open issues on downstream projects to track implementing this there too. [This blog post](https://github.com/dask/dask-blog/pull/64) I've submitted should be a good overview of the existing cluster managers.

mrocklin:
OK, this is in.  I'm curious to see how much activity this generates :)

PR #3854, opened 2020-06-04
Allow SSHCluster to take a list of connect_options

### LINKED ISSUE #2350: Support different usernames when starting an SSH cluster
I wanted to start an SSH cluster where my username is not same on all the nodes.  It would be great if this was supported.  The most obvious way to configure this would be to allow specifying hosts as `user@hostname`, either on the command line, or the hosts file.

jacobtomlinson:
Fixes #2350

In #2350 there is a requirement for different usernames to be passed to different ssh workers. This PR enables you to optionally pass a list of dictionaries to `connect_options` where that list must be equal length to the list of addresses and the two will be zipped.

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -1,6 +1,6 @@
 import logging
 import sys
-from typing import List
+from typing import List, Union
 import warnings
 import weakref
 
@@ -230,7 +230,7 @@ old_cluster_kwargs = {
 
 def SSHCluster(
     hosts: List[str] = None,
-    connect_options: List[dict] or dict = {},
+    connect_options: Union[List[dict], dict] = {},
     worker_options: dict = {},
     scheduler_options: dict = {},
     worker_module: str = "distributed.cli.dask_worker",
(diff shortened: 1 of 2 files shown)


distributed/deploy/ssh.py
```diff
@@ -230,7 +230,7 @@ async def start(self):
 
 def SSHCluster(
     hosts: List[str] = None,
-    connect_options: dict = {},
+    connect_options: List[dict] or dict = {},
```
jrbourbeau, line 233: I think we want `Union[List[dict], dict]`

distributed/deploy/ssh.py
```diff
@@ -311,11 +311,20 @@ def SSHCluster(
         kwargs.setdefault("worker_addrs", hosts)
         return OldSSHCluster(**kwargs)
 
+    if isinstance(connect_options, list):
+        if len(connect_options) != len(hosts):
```
jrbourbeau, line 315: This may be slightly more readable as a single `if`-statement

```python
if isinstance(connect_options, list) and len(connect_options) != len(hosts):
```

jacobtomlinson:
Thanks for the review @jrbourbeau. I've made the suggested changes.

PR #3855, opened 2020-06-05
Use SpecCluster name in worker names

lesteve:
So this comes from a long discussion about using job arrays in Dask-Jobqueue iin https://github.com/dask/dask-jobqueue/issues/196.

I'll try to summarise it:
- the emphasis is for now to provide a work-around to people who have a strong constraint to use job arrays
- when using a job array the job submission script looks like this (SGE here but would be similar for other job schedulers):
```bash
#!/usr/bin/env bash
# some other header things
# job array option
#$ -t 1-3

# 0 comes from `self._i` in SpecCluster
dask-worker <scheduler-address> --name 0
```
the problem is that this script will be run 3 times (the size of the job array) and workers will have the same name causing an error
- we would like to do:
```py
cluster = SGECluster(name='whatever_$SGE_TASK_ID', cores=1, memory='1GB')
```

so that the job submission script would look like this:
```bash
#!/usr/bin/env bash
# some other header things
# job array option
#$ -t 1-3

# 0 comes from `self._i` in SpecCluster
dask-worker <scheduler-address> --name whatever_$SGE_TASK_ID-0
```

This PR is an attempt at doing this. Suggestions on how to improve it would be more than welcome!

This modification is only motivated by Dask-Jobqueue so it could be argued that it should stay in Dask-Jobqueue. My feeling is that:
- having the SpecCluster name be used in the worker names is a rather limited change (to be fair I am not sure what SpecCluster name is used for outside of the `__repr__`)
- I have to say that my general approach is to keep as little tricky code in Dask-Jobqueue so this PR feels a better thing to do than overriding `new_worker_spec` in Dask-Jobqueue to do what we want. The latter approach feels more fragile and could break if some of the SpecCluster internals changed.

If I am being honest the Dask-Jobqueue use case feels a little bit kludgy because the key of the spec would be `wathever_$SGE_TASK_ID-0`, `wathever_$SGE_TASK_ID-1` and the worker names would be `whatever_1-0`, `whatever_2-0`, `whatever_3-0`, `whatever_1-1`, etc ... not sure it that could cause a problem later down the line.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -462,6 +462,14 @@ class SpecCluster(Cluster):
         if self.asynchronous:
             return NoOpAwaitable()
 
+    def _new_worker_name(self, worker_number):
+        """ Returns new worker name.
+
+        This can be overriden in SpecCluster derived classes to customise the
+        worker names.
+        """
+        return worker_number
+
     def new_worker_spec(self):
         """ Return name and spec for the next worker
 
@@ -473,13 +481,12 @@ class SpecCluster(Cluster):
         --------
         scale
         """
-        worker_name_template = f"{self._name}-{{}}"
-        worker_name = worker_name_template.format(self._i)
-        while worker_name in self.worker_spec:
+        new_worker_name = self._new_worker_name(self._i)
+        while new_worker_name in self.worker_spec:
             self._i += 1
-            worker_name = worker_name_template.format(self._i)
+            new_worker_name = self._new_worker_name(self._i)
 
-        return {worker_name: self.new_spec}
+        return {new_worker_name: self.new_spec}
 
     @property
     def _supports_scaling(self):
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -473,10 +473,13 @@ class SpecCluster(Cluster):
         --------
         scale
         """
-        while self._i in self.worker_spec:
+        worker_name_template = f'{self._name}-{{}}'
+        worker_name = worker_name_template.format(self._i)
+        while worker_name in self.worker_spec:
             self._i += 1
+            worker_name = worker_name_template.format(self._i)
 
-        return {self._i: self.new_spec}
+        return {worker_name: self.new_spec}
 
     @property
     def _supports_scaling(self):
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -538,7 +538,7 @@ class SpecCluster(Cluster):
         maximum_cores: int = None,
         minimum_memory: str = None,
         maximum_memory: str = None,
-        **kwargs
+        **kwargs,
     ) -> Adaptive:
         """ Turn on adaptivity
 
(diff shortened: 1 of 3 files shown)


lesteve:
Slightly tangential: something I would love to get input on as well is what to do with the spec `group`.

I have read the SpecCluster docstring and this certainly looks like something that applies here (a single entry in the spec creates multiple Dask workers). Looking at the code it feels the spec `group` is only used in adaptive. Did I miss something? If it is only used in adaptive, I think the reasonable thing to do is to ignore it for the Dask-Jobqueue use case because the scaling is going to be weird anyway and the goal is to provide a work-around for now.

Note that in Dask-Jobqueue there would be two reasons of multiple Dask workers here:
- one because of the job id indside the job array
- one for multiple Dask workers created by a single `dask-worker` command e.g. `dask-worker --nprocs 3`

distributed/deploy/spec.py
```diff
@@ -473,10 +473,13 @@ def new_worker_spec(self):
         --------
         scale
         """
-        while self._i in self.worker_spec:
+        worker_name_template = f"{self._name}-{{}}"
+        worker_name = worker_name_template.format(self._i)
```
mrocklin, line 477: What if instead we generalize this to having a `SpecCluster._new_worker_name` method or something similar?  Today this would be the following by default

```python
def _new_worker_name(self):
    while self._i in self.worker_spec:
        self._i += 1
    return self._i
```

But Dask-JobQueue could overwrite it on its own method.

This is just one idea, but shows how we might generailize something on SpecCluster that would make it easier to modify downstream without changing default behaviors on SpecCluster itself.

lesteve: This seems completely reasonable (I thought about adding a `worker_name_func` as an argument to `SpecCluster.__init__` which seems slightly overcomplicated but not the good old object-oriented approach), I'll do this and add some tests as well.

lesteve: I added the `_new_worker_name(self, worker_number)` method. This feels simpler to customise in Dask-Jobqueue (no need for the `self._i` updating logic).

I also added some simple tests.

distributed/deploy/spec.py
```diff
@@ -535,7 +545,7 @@ def adapt(
         maximum_cores: int = None,
         minimum_memory: str = None,
         maximum_memory: str = None,
-        **kwargs
+        **kwargs,
```
lesteve, line 548: Interestingly I had to change this to make black happy (I tend to try to avoid unrelated changes like this in general) ... I would have thought all the files were black-compliant but maybe black changed its mind about this at one point.

mrocklin:
This looks good to me.  There was an unrelated test failure (I think).  I'm going to increase a timeout and see if that helps (I hope that you do not mind my pushing to your branch).

lesteve:
> I hope that you do not mind my pushing to your branch

I actually think this is a super useful feature and allow more productive iterations (both from the contributor and the maintainer point of view), so clearly pushes to my branches more than welcome!

distributed/deploy/spec.py
mrocklin: Black adds trailing commas if the code is identified to be Python 3.6+ (it isn't legal in Python 3.5) so maybe you added some other feature that changed how black perceived the code?  Regardless, this is fine.

PR #3860, opened 2020-06-08
Replace stream= with comm= in handlers

### LINKED ISSUE #3824: Replace stream= with comm= in handlers
Over time we switched from using the keyword `stream=` to `comm=` (this happened when we generalized comms beyond just tornado streams).  

It looks like there are still some modules which use the term `stream=`.  I *think* that we should replace these with `comm=`, although it presumably doesn't matter too much.

jjerphan:
Fixes #3824.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -33,17 +33,17 @@ logger = logging.getLogger(__name__)
 MAX_BUFFER_SIZE = MEMORY_LIMIT / 2
 
 
-def set_tcp_timeout(stream):
+def set_tcp_timeout(comm):
     """
     Set kernel-level TCP timeout on the stream.
     """
-    if stream.closed():
+    if comm.closed():
         return
 
     timeout = dask.config.get("distributed.comm.timeouts.tcp")
     timeout = int(parse_timedelta(timeout, default="seconds"))
 
-    sock = stream.socket
+    sock = comm.socket
 
     # Default (unsettable) value on Windows
     # https://msdn.microsoft.com/en-us/library/windows/desktop/dd877220(v=vs.85).aspx
@@ -92,15 +92,15 @@ def set_tcp_timeout(stream):
         logger.warning("Could not set timeout on TCP stream: %s", e)
 
 
-def get_stream_address(stream):
+def get_stream_address(comm):
     """
     Get a stream's local address.
     """
-    if stream.closed():
+    if comm.closed():
         return "<closed>"
 
     try:
-        return unparse_host_port(*stream.socket.getsockname()[:2])
+        return unparse_host_port(*comm.socket.getsockname()[:2])
     except EnvironmentError:
         # Probably EBADF
         return "<closed>"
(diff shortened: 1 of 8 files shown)


quasiben:
Thanks for doing this @jjerphan .  What do you think about things like:
https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/comm/tcp.py#L95

https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/comm/tcp.py#L36

I found those by doing:
> git grep 'stream)'

distributed/comm/ucx.py
```diff
@@ -37,13 +37,13 @@
...
+def synchronize_stream(comm=0):
     import numba.cuda
 
     ctx = numba.cuda.current_context()
-    cu_stream = numba.cuda.driver.drvapi.cu_stream(stream)
-    stream = numba.cuda.driver.Stream(ctx, cu_stream, None)
-    stream.synchronize()
+    cu_stream = numba.cuda.driver.drvapi.cu_stream(comm)
+    comm = numba.cuda.driver.Stream(ctx, cu_stream, None)
+    comm.synchronize()
```
mrocklin, line 46: This is actually a different kind of stream (CUDA streams rather than Dask streams) and we should probably leave it as is.

distributed/comm/tcp.py
```diff
@@ -160,10 +160,10 @@ def _read_extra(self):
         pass
 
     def _get_finalizer(self):
-        def finalize(stream=self.stream, r=repr(self)):
-            if not stream.closed():
+        def finalize(comm=self.stream, r=repr(self)):
+            if not comm.closed():
                 logger.warning("Closing dangling stream in %s" % (r,))
-                stream.close()
+                comm.close()
```
mrocklin, line 166: This is actually a different kind of stream (Torando IOStream rather than Dask stream) and we should probably leave it as-is.

jjerphan:
Hum, I am trying to understand [the failing test on the Travis CI build](https://travis-ci.org/github/dask/distributed/jobs/696552213#L1429): it does not fail for me (it is skipped when running it locally).

mrocklin:
Yeah, that isn't your fault.  I don't know what happened, but that test has become flaky recently, and only on Python 3.6.  I'm going to go ahead and skip it for now.  It's a very peripheral feature.

See https://github.com/dask/distributed/pull/3888

mrocklin:
My apologies for the testing uncleanliness .  Thank you for your patience with it.

mrocklin:
I'm merging this in.  Thank you for your work here @jjerphan !

jjerphan:
@mrocklin: it really is fine: don't worry. :)
Thanks for the review!

jjerphan:
In fact, I [once studied a problem between in the `DaskBackendConnector`](https://github.com/jjerphan/joblib_dask_deadlock) but I have not been able fixed it. [joblib#914](https://github.com/joblib/joblib/pull/914) did however!

PR #3865, opened 2020-06-08
Drop no longer needed local `import pickle`

jakirkham:
We already [`import pickle` at the top of the module]( https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/scheduler.py#L15 ). So there is no need to import it locally here as well. So drop the local `import pickle`.


quasiben:
Thanks @jakirkham merging in

jakirkham:
Thanks Ben! 😄

PR #3868, opened 2020-06-09
Drop old frame splitting in NumPy serialization

jakirkham:
This is leftover from when Blosc compression was still being handled with the NumPy serialization code path itself ( https://github.com/dask/distributed/pull/777 ) ( https://github.com/dask/distributed/pull/830 ) ( https://github.com/dask/distributed/pull/832 ) ( https://github.com/dask/distributed/pull/998 ). These days frame splitting already happens as part of serialization more generally before compressing data. Also this cutoff is not relevant for when splitting occurs as that is determined by [a limit in the size of buffers Blosc can handle]( https://github.com/dask/distributed/blob/6144eb1a2f93274148cf84299f0eb0a2d93fc509/distributed/protocol/utils.py#L6 ), which [`frame_split_size` already knows]( https://github.com/dask/distributed/blob/6144eb1a2f93274148cf84299f0eb0a2d93fc509/distributed/protocol/utils.py#L22 ). So go ahead and drop this special casing in NumPy serialization and rely on these commonly shared code paths that all serialization passes through.


mrocklin:
Looks good to me.  Thanks for sifting through this part of the codebase @jakirkham 

Linting error:

    distributed/protocol/numpy.py:4:1: F401 '.utils.frame_split_size' imported but unused

If memory serves you prefer that others don't push to your branches, so I'll hold off on pushing a flake8 commit.

jakirkham:
Good catch. Thanks! Fixed in the latest commit.

Also we shouldn't need the `merge_frames` call in deserialization of NumPy arrays any more for the same reason. So have dropped that as well.

quasiben:
Thanks @jakirkham for the work and @mrocklin for the review.  Merging in

jakirkham:
Thanks Ben and Matt! 😄

PR #3871, opened 2020-06-09
Use pickle protocol 5 with NumPy object arrays

jakirkham:
Depends on PR ( https://github.com/dask/distributed/pull/3868 )

Builds off the work in PR ( https://github.com/dask/distributed/pull/3784 ) to support pickle protocol 5. Supports out-of-band pickling when serializing NumPy object arrays. This can be useful when serializing ragged arrays for example or really any NumPy object arrays containing data that supports out-of-band pickling.

### DIFF
--- a/distributed/protocol/numpy.py
+++ b/distributed/protocol/numpy.py
@@ -4,7 +4,7 @@ import numpy as np
 from .serialize import dask_serialize, dask_deserialize
 from . import pickle
 
-from ..utils import log_errors
+from ..utils import log_errors, nbytes
 
 
 def itemsize(dt):
@@ -22,7 +22,10 @@ def itemsize(dt):
 def serialize_numpy_ndarray(x):
     if x.dtype.hasobject:
         header = {"pickle": True}
-        frames = [pickle.dumps(x)]
+        frames = [None]
+        buffer_callback = lambda f: frames.append(memoryview(f))
+        frames[0] = pickle.dumps(x, buffer_callback=buffer_callback)
+        header["lengths"] = tuple(map(nbytes, frames))
         return header, frames
 
     # We cannot blindly pickle the dtype as some may fail pickling,
@@ -96,10 +99,10 @@ def serialize_numpy_ndarray(x):
 @dask_deserialize.register(np.ndarray)
 def deserialize_numpy_ndarray(header, frames):
     with log_errors():
-        (frame,) = frames
-
         if header.get("pickle"):
-            return pickle.loads(frame)
+            return pickle.loads(frames[0], buffers=frames[1:])
+
+        (frame,) = frames
 
         is_custom, dt = header["dtype"]
         if is_custom:
--- a/distributed/protocol/tests/test_numpy.py
+++ b/distributed/protocol/tests/test_numpy.py
@@ -81,9 +81,10 @@ def test_dumps_serialize_numpy(x):
         assert isinstance(frame, (bytes, memoryview))
     y = deserialize(header, frames)
 
-    np.testing.assert_equal(x, y)
     if x.flags.c_contiguous or x.flags.f_contiguous:
         assert x.strides == y.strides
+
+    np.testing.assert_equal(x, y)
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


distributed/protocol/numpy.py
```diff
@@ -23,7 +23,10 @@ def itemsize(dt):
 def serialize_numpy_ndarray(x):
     if x.dtype.hasobject:
         header = {"pickle": True}
-        frames = [pickle.dumps(x)]
+        frames = [None]
+        buffer_callback = lambda f: frames.append(memoryview(f))
```
mrocklin, line 27: Why are we converting into a `memoryview` here?

jakirkham: The `buffer_callback` is called on values of type `PickleBuffer`. Here's an example:

```python
In [1]: import pickle                                                           

In [2]: import numpy as np                                                      

In [3]: a = np.array([np.arange(3), np.arange(4, 6)], dtype=object)             

In [4]: l = []; d = pickle.dumps(a, protocol=5, buffer_callback=l.append)       

In [5]: l                                                                       
Out[5]: [<pickle.PickleBuffer at 0x115a64740>, <pickle.PickleBuffer at 0x115a647c0>]
```

Though `PickleBuffer` is a wrapper that expects to be passed some [`bytes`-like object]( https://docs.python.org/3.8/glossary.html#term-bytes-like-object ), which it wraps zero-copy. In the example above these are the nested NumPy arrays.

However as the rest of the serialization code expects frames to be `bytes` or `memoryview`s. We would either need to retool it to support `PickleBuffer` (with special casing when it is not available) or we can simply convert it to a `memoryview`, which is also zero-copy.

We've gone for the latter path of converting to a `memoryview` similar to what was done in PR ( https://github.com/dask/distributed/pull/3784 ). Note this is done without loss of generality.

jakirkham:
This should be ready on my end. Please let me know if anything else is needed 🙂

mrocklin:
One more dumb question (sorry for drawing this out): how does this work if the user doesn't have a more modern version of Python?  If pickle5 isn't installed and we're using an older version of Python then what happens here?

jakirkham:
Not at all. The discussion is welcome. 🙂 Was more noting this is not in draft status any more.

In the case pickle protocol 5 is not supported, we would [skip collecting buffers]( https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/protocol/pickle.py#L37-L38 ) when calling `pickle.dumps` or `cloudpickle.dumps` and would [skip calling the `buffer_callback`]( https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/protocol/pickle.py#L56-L58 ) ([as there would be no `buffers`]( https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/protocol/pickle.py#L35 )). As a result we would have [only one frame here]( https://github.com/dask/distributed/blob/e07d26ba8136703a5896ebffd18167b0a980eaf1/distributed/protocol/numpy.py#L27 ), which results from calling `pickle.dumps` or `cloudpickle.dumps` as before. When it comes to unpickle the data, we would have [the first frame as an argument to `loads`]( https://github.com/dask/distributed/blob/e07d26ba8136703a5896ebffd18167b0a980eaf1/distributed/protocol/numpy.py#L103 ) and an empty `list`/`tuple` for `buffers` so would [just calling `pickle.loads` without `buffers`]( https://github.com/dask/distributed/blob/8480b4fde57d16b27b3079f25b95e61b862b79ae/distributed/protocol/pickle.py#L67 ).

TL;DR it behaves the same as the code does before this change.

mrocklin:
Ah, ok I was thrown by seeing `pickle.dumps(...)` without realizing that this was *our* wrapping around pickle.  Grand.  Thanks for the detailed explanation @jakirkham .

jakirkham:
Yeah can see how that would be confusing. Yep, we are [importing it here]( https://github.com/jakirkham/distributed/blob/e07d26ba8136703a5896ebffd18167b0a980eaf1/distributed/protocol/numpy.py#L5 ).

Of course! Thanks for the review 😄

PR #3874, opened 2020-06-09
Allow dictionaries in security= keywords

mrocklin:
Previously Dask Server and Client objects had to be instantiated with a
Security object for security.  Now we also allow for dictionary inputs
that will be splatted out to the Security object.  This helps when using
systems like dask-spec with Security.

cc @jcrist if you have a moment


jcrist:
No comment on the implementation, but the idea seems fine.

mrocklin:
Thanks for the speedy reply @jcrist

PR #3875, opened 2020-06-09
Emmit warning when assign/comparing string with Status Enum.

Carreau:
Turn this into an error in the test suite.

We want to make sure of various things:

  1) original behavior of comparing/assign with string still works
  but emit warnings.
  1b) assigning strings convert to proper enum variant.
  2) assign/comparison with invalid strings should fail
  2) warnings are errors in test suite to make sure we don't re-introduce
  strings.

This is the continuation of #3853

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -9,6 +9,7 @@ import threading
 import traceback
 import uuid
 import weakref
+import warnings
 
 import dask
 import tblib
@@ -52,11 +53,13 @@ class Status(Enum):
     closing = "closing"
     closing_gracefully = "closing-gracefully"
     init = "init"
+    created = "created"
     running = "running"
     starting = "starting"
     stopped = "stopped"
     stopping = "stopping"
     undefined = None
+    dont_reply = "dont-reply"
 
     def __eq__(self, other):
         """
@@ -69,6 +72,11 @@ class Status(Enum):
         if isinstance(other, type(self)):
             return self.value == other.value
         elif isinstance(other, str) or (other is None):
+            warnings.warn(
+                f"Since distributed 2.19 `.status` is now an Enum, please compare with `Status.{other}`",
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
             assert other in [
                 s.value for s in type(self)
             ], f"comparison with non-existing states {other}"
@@ -261,9 +269,16 @@ class Server:
         if isinstance(new_status, Status):
             self._status = new_status
         elif isinstance(new_status, str) or new_status is None:
+            warnings.warn(
+                f"Since distributed 2.19 `.status` is now an Enum, please assign `Status.{new_status}`",
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
             corresponding_enum_variants = [s for s in Status if s.value == new_status]
             assert len(corresponding_enum_variants) == 1
             self._status = corresponding_enum_variants[0]
+        else:
+            raise TypeError(f"expected Status or str, got {new_status}")
 
     async def finished(self):
         """ Wait until the server has finished """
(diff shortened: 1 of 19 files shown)


Carreau:
I'm expecting the tests to fail with many more areas where status is still compared with or assingned a string.

mrocklin:
I'm inclined to not warn just yet, and give downstream libraries a chance to react to this change.  Thoughts?

Carreau:
I'm fine not warning but I think that warnings (wel phrased with the right class) are a good and important communication channel, this is also why I did a PendingDeprecationWarning and not a DeprecationWarning.

I'll also mostly use CI failure to find all the place that need updates.

I can also update the message to say "We are thinking about deprecating ... blah blah", and let the deprecation in master but not release them so that downstread have  chance to pitch in on non-release versions.

mrocklin:
Yeah, let me rephrase.  I think that we should not impose warnings on users yet, at least not until we fix the other Dask sub-packages and have a release cycle or two.  If you're using a warning class that is invisile to users under default settings then that seems sensble to me.

Carreau:
> If you're using a warning class that is invisile to users under default settings then that seems sensible to me.

In that case PendingDeprecationWarnings fits that definition. It is not visible in Repl or scripts. I believe since 3.7 only DeprecationWarnings are shown in REPL by default.

distributed/deploy/spec.py
```diff
@@ -36,19 +36,39 @@ class ProcessInterface:
     It should implement the methods below, like ``start`` and ``close``
     """
 
+    @property
```
mrocklin, line 39: Maybe instead we should implement this on the `Cluster` superclass in `distributed/deploy/cluster.py`

mrocklin: That would also affect most of the downstream projects automatically, which would be useful.

Carreau: We could, but then ssh.Worker and ssh.Scheduler would not inherit this and I believe they were the reason I did this originally. 

I think I might end up pulling this into a decorator that can be applied to classes maybe.

distributed/core.py
```diff
@@ -519,7 +534,10 @@ async def handle_comm(self, comm, shutting_down=shutting_down):
                         logger.exception(e)
                         result = error_message(e, status="uncaught-error")
 
-                if reply and result != "dont-reply":
+                # result is not type stable:
+                # when LHS is not Status then RHS must not be Status or it raises.
+                # when LHS is Status then RHS must be status or it raises in tests
+                if reply and ininstance(result, Status) and result != Status.dont_reply:
```
Carreau, line 540: I'm not super happy with this either.

distributed/deploy/spec.py
mrocklin: Oh, my apologies.  I saw the `spec.py` filename and assumed that this was the SpecCluster class.  I see now that it is the `ProcessInterface` class instead.  Happy to retract the comment.

I'd prefer avoiding the iterator and being explicit for now if that's ok, even given the duplication.

Carreau:
Seem like the last few failures are around Status not being serialisable for msgpack. I'm not too sure how to write a custom serializer for that, and if that could be a deal breaker.

Carreau:
I found [here](https://github.com/msgpack/msgpack-python#user-content-packingunpacking-of-custom-data-type) how to do custom packing.

Carreau:
Yeah ! All tests passing.

There are other places where messagepack is used, but those seem like the only two when those were needed.  I defined in serialize as this is where most of msgpack serialisation lies. 

If you are happy with this, let me know and I can rebase/squash and update the version number in PeindingDeprecationWarning + something in the changelog.

Carreau:
Going to ping on this; if you are still +1 on the approach, and if there is anything you want me to change.

mrocklin:
Sure.  Let's give it a try.  We're a couple weeks away from a release so we can see who this affects for a while and we'll have plenty of time to revert if necessary.

PR #3883, opened 2020-06-10
Check hosts for None value in SSH cluster.

Carreau:
Then mypy is happy and does not complain in the rest of the function.
If host was None or empty list it would fail anyway either on the next
line `len(None)`, or further down `host[0]`.

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -235,7 +235,7 @@ def SSHCluster(
     scheduler_options: dict = {},
     worker_module: str = "distributed.cli.dask_worker",
     remote_python: str = None,
-    **kwargs
+    **kwargs,
 ):
     """ Deploy a Dask cluster using SSH
 
@@ -315,6 +315,10 @@ def SSHCluster(
         kwargs.setdefault("worker_addrs", hosts)
         return OldSSHCluster(**kwargs)
 
+    if not hosts:
+        raise ValueError(
+            f"`hosts` must be a non empty list, value {repr(hosts)!r} found."
+        )
     if isinstance(connect_options, list) and len(connect_options) != len(hosts):
         raise RuntimeError(
             "When specifying a list of connect_options you must provide a "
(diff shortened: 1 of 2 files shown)


mrocklin:
This seems fine to me.  If you wanted to add a test that would be welcome, but this isn't critical.

The linter is complaining.  If you're not already aware we have git pre-commit hooks in the repositories that you might want to activate.  

https://docs.dask.org/en/latest/develop.html#code-formatting

(although I suspect that you already know this)

Carreau:
> This seems fine to me. If you wanted to add a test that would be welcome, but this isn't critical.

Done, It would have failed anyway but with  a different error category. 

> If you're not already aware we have git pre-commit hooks

yeah, I don't like pre commit hooks, I prefer to have my editor do that but I still need to find a better setup to have black activated on a per folder/project basis. I also have our bot which can make github comments with black formatting  code suggestion you can apply from within github UI (e.g: `https://github.com/ipython/traitlets/pull/558`) to not have to ask new contributors to install black and run it. It's still non optimal though.

Forced push to clean the changes. Would you be opposed to have mypy as a linter as well (on currently non failing files only of course).

mrocklin:
> Forced push to clean the changes

We squash-and-merge by the way, so I wouldn't spend too much energy on commit cleanliness.

> Would you be opposed to have mypy as a linter as well (on currently non failing files only of course).

Personally I would be fine with it, but this should be a larger issue.  Some of the maintainers aren't yet sold on mypy.  Maybe open something on github.com/dask/community  ?

Carreau:
> We squash-and-merge by the way, so I wouldn't spend too much energy on commit cleanliness.

That still give a cleaner commit message doesn't it ? I'm not sure how the squash merge handle multiple commits message at squash time.

mrocklin:
It turns them into a bulletted list.  However, these are usually very messy, and so I often remove the auto-generated commit message and either write something myself, or use the posting from the header comment on the PR.  This is only if you don't have good commit messages though.  If you take the time to write good commit messages then we'll keep them, but for maybe 90% of contributions this isn't the case, so I handle things manually.

Carreau:
Thanks.

Issue #3420, opened 2020-01-28
daskssh cluster does not close


condoratberlin:
Hi,

when starting a sshcluster via:

```
from distributed.deploy.ssh import SSHCluster

def start_cluster():
    hosts = ['localhost']
    nthreads = 1
    nprocs = 2
    c = SSHCluster(hosts * (nprocs+1),
                   connect_options={"known_hosts": None},
                   worker_options={"nthreads": nthreads, },
                   scheduler_options={"port": 0, "dashboard_address": ":8787"},
                   )
    c.scale(nprocs)
    c.close()

if __name__ == '__main__':
    start_cluster()
```

the scheduler and the worker are still alive even if the calling python process is finished. The dashboard is reachable and the python processes are shown in "top".

Is there another way to close the ssh cluster?

Using Linux: 
distributed 2.9.1
dask 2.9.1
python 3.6.9

jakirkham:
Perhaps you could `scale` back to `0` and poll until the workers are shutdown before closing out the client.

condoratberlin:
`c.scale(0)` has no effect. scale seems not to work for SSHCluster. The number of worker is determined by the number of hosts -1 for the scheduler.

`from distributed.deploy.ssh import SSHCluster

def start_cluster():
    hosts = ['localhost']
    nthreads = 1
    nprocs = 2
    c = SSHCluster(hosts * (nprocs+1),
                   connect_options={"known_hosts": None},
                   worker_options={"nthreads": nthreads, },
                   scheduler_options={"port": 0, "dashboard_address": ":8787"},
                   )

if __name__ == '__main__':
    start_cluster()`

creates also 2 workers.

condoratberlin:
I figured out that it could be an issue with openssh as described [here](https://github.com/ronf/asyncssh/issues/112) 

> ...the problem is that OpenSSH actually doesn't implement the "signal" channel request...

The reason is that the 

`self.proc.kill()` command in deploy.ssh Process.close() is not executed and the process does not stop. There is also a workaround described [here](https://github.com/ronf/asyncssh/issues/135#issuecomment-380312533)

> Possible workarounds include requesting a PTY and sending a ^C or ^\ on the SSH channel's stdin to trigger the remote shell to send a SIGINT or SIGTERM to the process you started..

If got OpenSSH_7.2 on my machine.

condoratberlin:
[...]

PR #3885, opened 2020-06-11
SSHCluster send kill signal

jacobtomlinson:
There is a discussion in #3420 around older versions of OpenSSH not respecting the kill command, which results in SSH connections being left open after `SSHCluster` is done.

This issue (ronf/asyncssh#112) discusses the problem and explains that upgrading your OpenSSH version will resolve it. However as not all users are able to do this is also explains a workaround which involves sending a keyboard interrupt character to stdin on the process to manually kill it.

This PR implements that workaround.

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -32,12 +32,15 @@ class Process(ProcessInterface):
     async def start(self):
         assert self.connection
         weakref.finalize(
-            self, self.proc.kill
+            self, self.proc.stdin.write, "\x03"
         )  # https://github.com/ronf/asyncssh/issues/112
         await super().start()
 
     async def close(self):
-        self.proc.kill()  # https://github.com/ronf/asyncssh/issues/112
+        # https://github.com/ronf/asyncssh/issues/112
+        await self.proc.stdin.write("\x03")
+        await self.proc.wait_closed()
+
         self.connection.close()
         await super().close()
 
@@ -120,7 +123,7 @@ class Worker(Process):
             + cli_keywords(self.kwargs, cls=_Worker, cmd=self.worker_module)
         )
 
-        self.proc = await self.connection.create_process(cmd)
+        self.proc = await self.connection.create_process(cmd, term_type="ansi")
 
         # We watch stderr in order to get the address, then we return
         while True:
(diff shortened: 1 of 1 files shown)


Keou0007:
Tested this on macOS and it works as expected, but mac worked before anyway...

Tried on CentOS 7 and I run into an issue where the scheduler process is spawned, but the worker processes never appear. It sits stalled seemingly indefinitely, and when I killed with CTRL-C, I get this traceback.

```
>>> from distributed import SSHCluster 
>>> cluster = SSHCluster(['localhost']*4, connect_options={'known_hosts':None}) 
File “<stdin>”, line 1, in <module>
File "/opt/miniconda3/lib/python3.8/site-packages/distributed/deploy/ssh.py", line 353, in SSHCluster 
	return SpecCluster(workers, scheduler, name="SSHCluster", **kwargs) 
File "/opt/miniconda3/lib/python3.8/site-packages/distributed/deploy/spec.py", line 256, in __init__ 
	self.sync(self._start) 
File "/opt/miniconda3/lib/python3.8/site-packages/distributed/deploy/cluster.py", line 163, in sync 
File "/opt/miniconda3/lib/python3.8/site-packages/distributed/utils.py", line 336, in sync 
	e.wait(10) 
File "/opt/miniconda3/lib/python3.8/threading.py", line 558, in wait 
File "/opt/miniconda3/lib/python3.8/threading.py", line 306, in wait 
KeyboardInterrupt 
```

Seems specifying `term_type="ansi"` is preventing the workers from launching. I'm assuming the workers are being launched by the scheduler, so perhaps the Linux version (or older version) of OpenSSH doesn't allow creating cascading ansi sessions?!

Possible solution being to launch the scheduler on the local host without it being controlled via an ssh connection? Is there a need to be able to spawn the scheduler on a remote machine?

Keou0007:
I made a mistake, I also see the stalling behaviour on macOS. I had applied the workaround to dask in a different environment and was testing the old version.

mrocklin:
Checking in on this.  Do you think that this is worth continuing @jacobtomlinson ?

jacobtomlinson:
Thanks for trying this out @Keou0007. Sorry it didn't work out.

PR #3889, opened 2020-06-12
Docs: Resolve PEP-3148/3184 typo

DomHudson:
Link incorrectly titled PEP-3184 when it refers to https://www.python.org/dev/peps/pep-3148/.


mrocklin:
Thanks @DomHudson !

PR #3895, opened 2020-06-15
Add support for port forwarding in SSHCluster

### LINKED ISSUE #3881: Support port forwarding in SSHCluster
Some users may have SSH access to a machine and want to use `SSHCluster` with it, but are not able to access ports other than 22 due to firewall restrictions.

It would be nice to be able to forward the scheduler ports locally through the SSH tunnel.

This would require:
- Add kwargs to `SSHCluster` to specify local port forwards
- Set up port forward in [scheduler object](https://github.com/dask/distributed/blob/6ee7f899d1bc686ba4910c08509c7f2e5345506d/distributed/deploy/ssh.py#L170) after connection is made ([docs](https://asyncssh.readthedocs.io/en/latest/api.html#asyncssh.SSHClientConnection.forward_remote_port))
- Set the scheduler address to the locally forwarded port [here](https://github.com/dask/distributed/blob/6ee7f899d1bc686ba4910c08509c7f2e5345506d/distributed/deploy/ssh.py#L206)

**Example**

```python
>>> from dask.distributed import Client, SSHCluster
>>> cluster = SSHCluster(
...     ["host1", "host2", "host3", "host4"],
...     forward_scheduler_port=8786,
...     forward_dashboard_port=8787
... )
>>> client = Client(cluster)
```

There is an assumption here that the restrictive firewall rules would not limit worker-scheduler and worker-worker communication.

jacobtomlinson:
Closes  #3881

Added `forward_scheduler_port` and `forward_dashboard_port` kwargs to `SSHCluster` to automatically handle port forwarding. This means you only need access to port `22` on the remote system and you'll still be able to connect to the cluster.

Not set on the kwarg names, happy to take input on those.

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import sys
 from typing import List, Union
@@ -150,10 +151,20 @@ class Scheduler(Process):
     kwargs: dict
         These will be passed through the dask-scheduler CLI to the
         dask.distributed.Scheduler class
+    forward_scheduler_port: int, optional
+        Local port to forward the remote scheduler port to.
+    forward_dashboard_port: int, optional
+        Local port to forward the remote dashboard port to.
     """
 
     def __init__(
-        self, address: str, connect_options: dict, kwargs: dict, remote_python=None
+        self,
+        address: str,
+        connect_options: dict,
+        kwargs: dict,
+        remote_python=None,
+        forward_scheduler_port: int = None,
+        forward_dashboard_port: int = None,
     ):
         super().__init__()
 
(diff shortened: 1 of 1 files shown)


distributed/deploy/ssh.py
```diff
@@ -204,10 +218,37 @@ async def start(self):
...
+                        await self.connection.forward_remote_port(
+                            "localhost",
+                            int(self.forward_scheduler_port),
+                            "localhost",
+                            int(port),
+                        )
+                    )
+                    self.address = "tcp://localhost:{}".format(
+                        self.forward_scheduler_port
+                    )
```
mrocklin, line 233: What if we're using a different protocol, like tls?

mrocklin:
This looks slick to me.  It might be nice to have an example in the docstring (or a follow-on PR to docs.dask.org)

distributed/deploy/ssh.py
jacobtomlinson: Ah whoops I meant to fix that.

jacobtomlinson: Resolved, now auto detects in the same way it detects the port on the remote end.

mrocklin:
Any thoughts on the TravisCI failures here?  They appear to be related.

mrocklin:
Thanks for keeping this up to date @jacobtomlinson 

It looks like the Travis CI failures here are still live

Issue #3877, opened 2020-06-10
Make heartbeat easier to configure


TomAugspurger:
As noted by @mmohrhard  in https://github.com/dask/distributed/issues/3691#issuecomment-641845231, handing heartbeats can be an issue for very large clusters. Increasing the heartbeat interval is one workaround, but it's not straightforward right now.

https://github.com/dask/distributed/blob/6ee7f899d1bc686ba4910c08509c7f2e5345506d/distributed/scheduler.py#L5561, https://github.com/dask/distributed/blob/6ee7f899d1bc686ba4910c08509c7f2e5345506d/distributed/scheduler.py#L5301

We do have a config for the client heartbeat

```yaml
distributed:
  client:
    heartbeat: 5s  # Interval between client heartbeats
```

And a related `distributed.scheduler.worker-ttl` key. Otherwise the worker heartbeat seems to be hardcoded to 1000 (ms?) at https://github.com/dask/distributed/blob/6ee7f899d1bc686ba4910c08509c7f2e5345506d/distributed/worker.py#L643. We should add that to the config at `distributed.worker.heartbeat`.

PR #3896, opened 2020-06-15
update heartbeat checks in scheduler

quasiben:
attempt to fix https://github.com/dask/distributed/issues/3877

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5303,7 +5303,9 @@ class Scheduler(ServerNode):
     async def check_worker_ttl(self):
         now = time()
         for ws in self.workers.values():
-            if ws.last_seen < now - self.worker_ttl:
+            if (ws.last_seen < now - self.worker_ttl) and (
+                10 * heartbeat_interval(len(self.workers))
+            ):
                 logger.warning(
                     "Worker failed to heartbeat within %s seconds. Closing: %s",
                     self.worker_ttl,
@@ -5571,7 +5573,7 @@ def heartbeat_interval(n):
     elif n < 200:
         return 2
     else:
-        return 5
+        return 10
 
 
 class KilledWorker(Exception):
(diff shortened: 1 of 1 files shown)


quasiben:
I was a little confused by this statement:
> I recommend modifying the distributed/scheduler.py::heartbeat_interval
function to remove the maximum

For now, I increased the maximum heartbeat to 10s

mrocklin:
> For now, I increased the maximum heartbeat to 10s

At the 20k worker level even this results in heartbeats every 500us, which might still be a burden on the scheduler.  I recommend that we drop the maximum interval entirely.

quasiben:
> At the 20k worker level even this results in heartbeats every 500us, which might still be a burden on the scheduler. I recommend that we drop the maximum interval entirely.

Sorry for being dense here but if we removed the maximum what should we be doing for messages like the following when len(workers)>200

```
return {
            "status": "OK",
            "time": time(),
            "heartbeat-interval": heartbeat_interval(len(self.workers)),
        }

```

mrocklin:
Lets say that we want no more than 200 heartbeats a second or something
like that, and adjust the heartbeat interval accordingly to match that
based on the number of workers.

distributed/scheduler.py
```diff
@@ -5303,7 +5303,9 @@ def _reevaluate_occupancy_worker(self, ws):
     async def check_worker_ttl(self):
         now = time()
         for ws in self.workers.values():
-            if ws.last_seen < now - self.worker_ttl:
+            if (ws.last_seen < now - self.worker_ttl) and (
+                10 * heartbeat_interval(len(self.workers))
```
jonashaag, line 5307: Terribly confused about this change. Doesn't `10 * heartbeat_interval()` always evaluate to true since `heartbeat_interval` is strictly positive?

quasiben: No, you're right.  i messed up the logic here.

garanews:
Is heartbeat_interval something in the future can be set in config like other Parameters?

mrocklin:
@garanews this PR has been closed.  I encourage you to raise a new issue at https://github.com/dask/distributed/issues/new

PR #3897, opened 2020-06-15
Fix total aggregated metrics in dashboard

lesteve:
Originally reported in https://github.com/dask/dask-jobqueue/issues/440.

To reproduce:
```py
import webbrowser
from distributed import LocalCluster

cluster = LocalCluster(n_workers=4)
webbrowser.open(cluster.dashboard_link.replace("status", "workers"))
```
image

The total memory usage is `250MiB`, the total memory limit is `16GiB`, so the total memory percentage should be `250/16e3 ~= 1.6%` and not `6.4%` (the memory percentage gets summed across all the workers rather than averaged).

I believe `memory_percent` is the only column affected by this.

Not sure how to add a test for this and I was not able to find some a good inspiration in the existing tests, but suggestions more than welcome!

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1864,7 +1864,7 @@ class WorkerTable(DashboardComponent):
                     total_data = (
                         sum(ws.metrics["cpu"] for ws in self.scheduler.workers.values())
                         / 100
-                        / sum(1 for ws in self.scheduler.workers.values())
+                        / len(self.scheduler.workers.values())
                     )
                 elif name == "cpu_fraction":
                     total_data = (
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -535,7 +535,7 @@ class ComputerPerKey(DashboardComponent):
             self.fig.title.text = "Compute Time Per Task"
 
             compute_result = dict(
-                times=compute_time, color=compute_colors, names=compute_names,
+                times=compute_time, color=compute_colors, names=compute_names
             )
 
             update(self.compute_source, compute_result)
@@ -635,7 +635,7 @@ class AggregateAction(DashboardComponent):
             self.fig.x_range.factors = agg_names
             self.fig.title.text = "Aggregate Time Per Action"
 
-            action_result = dict(times=agg_time, color=agg_colors, names=agg_names,)
+            action_result = dict(times=agg_time, color=agg_colors, names=agg_names)
 
             update(self.action_source, action_result)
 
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @jsignell who I think worked on this before

jsignell:
Wow, yeah. I guess it shouldn't be summing that column. CPU is wrong as well right?

distributed/dashboard/components/scheduler.py
```diff
@@ -1856,7 +1856,10 @@ def update(self):
                 )
                 continue
             try:
-                data[name].insert(0, sum(data[name]))
+                total_data = sum(data[name])
+                if name == "memory_percent":
+                    total_data /= len(data[name])
```
jsignell, line 1861: I don't actually think this is right if different workers have different memory limits. That is probably not super common, but it's probably more correct to do: 

```suggestion
                    n = len(data[name])
                    total_data = sum(data["memory"][-n:]) / sum(data["memory_limit"][-n:])
```

lesteve:
I fixed the computation for cpu, cpu_fraction and memory_percent when workers have different memory_limit. I used `ws.metrics` because it felt simpler than using `data` (one of the problem with `data`, the column you need may not have been inserted)

distributed/dashboard/components/scheduler.py
```diff
@@ -1856,7 +1856,26 @@ def update(self):
...
-                data[name].insert(0, sum(data[name]))
+                if name == "memory_percent":
+                    total_data = sum(
+                        ws.metrics["memory"] for ws in self.scheduler.workers.values()
+                    ) / sum(ws.memory_limit for ws in self.scheduler.workers.values())
+                elif name == "cpu":
+                    total_data = (
+                        sum(ws.metrics["cpu"] for ws in self.scheduler.workers.values())
+                        / 100
+                        / sum(1 for ws in self.scheduler.workers.values())
```
jsignell, line 1867: Slight simplification:

```suggestion
                        / len(self.scheduler.workers.values())
```

lesteve: Good point I went too much over the top with this `sum(... for ws in self.scheduler.worker.values())`! I pushed a commit your suggestion.

lesteve:
It seems like the CI failures are not related to this PR and should be fixed by https://github.com/dask/distributed/pull/3904.

mrocklin:
Merged and re-triggered CI here

lesteve:
I am seeing errors like these when exiting my IPython session. I guess they happen because at the beginning there are no workers, should I try to get rid of these, or should I ignore them (maybe by catching ZeroDivisionError in the try/catch a few lines below) ?

```
bokeh.util.tornado - ERROR - Error thrown from periodic callback:
bokeh.util.tornado - ERROR - Traceback (most recent call last):
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/tornado/gen.py", line 501, in callback
    result_list.append(f.result())
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/bokeh/server/session.py", line 67, in _needs_document_lock_wrapper
    result = func(self, *args, **kwargs)
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/bokeh/server/session.py", line 195, in with_document_locked
    return func(*args, **kwargs)
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/bokeh/document/document.py", line 1164, in wrapper
    return doc._with_self_as_curdoc(invoke)
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/bokeh/document/document.py", line 1150, in _with_self_as_curdoc
    return f()
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/bokeh/document/document.py", line 1163, in invoke
    return f(*args, **kwargs)
  File "/home/lesteve/dev/distributed/distributed/dashboard/components/__init__.py", line 77, in <lambda>
    doc.add_periodic_callback(lambda: update(ref), interval)
  File "/home/lesteve/dev/distributed/distributed/dashboard/components/__init__.py", line 84, in update
    comp.update()
  File "/home/lesteve/miniconda3/lib/python3.7/site-packages/bokeh/core/property/validation.py", line 93, in func
    return input_function(*args, **kwargs)
  File "/home/lesteve/dev/distributed/distributed/dashboard/components/scheduler.py", line 1867, in update
    / len(self.scheduler.workers.values())
ZeroDivisionError: float division by zero
```

mrocklin:
Yes, I think that we should handle the no-workers case, even if the warnings/logs are delayed.

lesteve:
> Yes, I think that we should handle the no-workers case, even if the warnings/logs are delayed.

Fixed. Actually this was causing a HTTP 500 in the dashboard when there was no workers which I did not realise until I tested my change.

mrocklin:
Great.  Thanks @lesteve !

PR #3904, opened 2020-06-16
Support Bokeh 2.1

mrocklin:
The `create_hosts_whitelist` function was renamed to `create_hosts_allowlist`

Fixes https://github.com/dask/distributed/issues/3741

FYI @bryevdv

PR #3906, opened 2020-06-16
Show cpu_fraction on hover for dashboard workers circle plot

lesteve:
A [few lines below](https://github.com/lesteve/distributed/blob/e696ec9ea356af9c48c44b5982dcb4a2c5e1bd2e/distributed/dashboard/components/scheduler.py#L1812) "cpu_fraction" is used to position the dot but "cpu" is shown rather than "cpu_fraction" on hover.

```py
cpu_plot.circle(
	source=self.source, x="cpu_fraction", y=0, size=10, fill_alpha=0.5
)
```

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1794,7 +1794,7 @@ class WorkerTable(DashboardComponent):
             tooltips="""
                 <div>
                   <span style="font-size: 10px; font-family: Monaco, monospace;">Worker (@name): </span>
-                  <span style="font-size: 10px; font-family: Monaco, monospace;">@cpu</span>
+                  <span style="font-size: 10px; font-family: Monaco, monospace;">@cpu_fraction</span>
                 </div>
                 """,
         )


mrocklin:
The test failures here are unrelated.  Resolved in https://github.com/dask/distributed/pull/3904

mrocklin:
cc @jsignell

jsignell:
Are we sure that this isn't intentional? Maybe people are interested in knowing the exact CPU rather than the fraction?

lesteve:
I have to admit I am not sure what people would find more useful but at the moment there is some inconsistency between:
- the horizontal position of the circle plot shows `cpu_fraction` (CPU usage per worker thread). The range is between 0 and 1 as shown [here](https://github.com/lesteve/distributed/blob/e696ec9ea356af9c48c44b5982dcb4a2c5e1bd2e/distributed/dashboard/components/scheduler.py#L1805) and the vertical bars are for 20%, 40%, 60%, 80% and 100%.
- the hover tooltip on the circle shows `cpu` (CPU usage per worker)

Maybe this snapshot helps:
image

Here `cpu` is roughly 100% as shown in the table, `cpu_fraction` is roughly 33% as shown by the circle plot (the dot is between the first and second vertical bar) but the hover shows roughly 100%.

jsignell:
Oh yeah, I think you are right with your initial assumption that the hover should display cpu_fraction.

PR #3907, opened 2020-06-16
Use dask-spec in SSHCluster

### LINKED ISSUE #3903: SSHCluster can't turn dashboard off
Trying to turn off the dashboard results in a traceback at launch for SSHCluster
```python
from distributed import SSHCluster
cluster = SSHCluster(['localhost']*4, scheduler_options={'dashboard_address':None})
```
<details>
  <summary>Traceback - Click to expand!</summary>
  
```
distributed.deploy.ssh - INFO - distributed.scheduler - INFO - -----------------------------------------------
distributed.deploy.ssh - INFO - distributed.http.proxy - INFO - To route to workers diagnostics web server please install jupyter-server-proxy: python -m pip install jupyter-server-proxy
distributed.deploy.ssh - INFO - Traceback (most recent call last):
distributed.deploy.ssh - INFO - File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/node.py", line 232, in start_http_server
distributed.deploy.ssh - INFO - self.http_server.listen(**http_address)
distributed.deploy.ssh - INFO - File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/tornado/tcpserver.py", line 151, in listen
distributed.deploy.ssh - INFO - sockets = bind_sockets(port, address=address)
distributed.deploy.ssh - INFO - File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/tornado/netutil.py", line 117, in bind_sockets
distributed.deploy.ssh - INFO - socket.getaddrinfo(address, port, family, socket.SOCK_STREAM, 0, flags),
distributed.deploy.ssh - INFO - File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/socket.py", line 918, in getaddrinfo
distributed.deploy.ssh - INFO - for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
distributed.deploy.ssh - INFO - socket.gaierror: [Errno 8] nodename nor servname provided, or not known
  File "/Users/shannon/Library/Application Support/JetBrains/PyCharm2020.1/scratches/scratch_1.py", line 19, in <module>
    cluster = SSHCluster(['localhost']*4, scheduler_options={'dashboard_address':None})
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/ssh.py", line 336, in SSHCluster
    return SpecCluster(workers, scheduler, name="SSHCluster", **kwargs)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/spec.py", line 256, in __init__
    self.sync(self._start)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/cluster.py", line 163, in sync
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/utils.py", line 347, in sync
    raise exc.with_traceback(tb)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/utils.py", line 331, in f
    result[0] = yield future
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/spec.py", line 284, in _start
    self.scheduler = await self.scheduler
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/spec.py", line 50, in _
    await self.start()
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/ssh.py", line 203, in start
    raise Exception("Worker failed to start")
Exception: Worker failed to start
distributed.deploy.spec - WARNING - Cluster closed without starting up
Error in atexit._run_exitfuncs:
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/utils.py", line 198, in ignoring
    yield
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/spec.py", line 611, in close_clusters
    cluster.close(timeout=10)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/cluster.py", line 84, in close
    return self.sync(self._close, callback_timeout=timeout)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/cluster.py", line 163, in sync
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/utils.py", line 347, in sync
    raise exc.with_traceback(tb)
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/utils.py", line 331, in f
    result[0] = yield future
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/spec.py", line 400, in _close
    await super()._close()
  File "/Users/shannon/opt/miniconda3/envs/pywakes/lib/python3.8/site-packages/distributed/deploy/cluster.py", line 73, in _close
    await self._watch_worker_status_comm.close()
AttributeError: 'SpecCluster' object has no attribute '_watch_worker_status_comm'
```

</details>

**Environment**:

- Dask version: 2.17.2
- Python version: 3.8.3
- Operating System: macOS

mrocklin:
Fixes #3903

Previously we used to specify CLI commands like dask-worker.
Now we specify Python classes like Worker or Nanny.
This makes it more consistent to pass through keywords without having to
go from Python to CLI back to Python.

However, this will also probably force dask-cuda to push its logic
from a CLI command into a Python class.
This commit speculatively changes documentation here in order to suggest
a CUDAWorker class type.  

I *think* that using dask-spec is probably a clean way forward generally.
Most systems that have switched over to it tend to have fewer complaints.

cc @quasiben @pentschev @jacobtomlinson

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import sys
 from typing import List, Union
@@ -7,9 +8,6 @@ import weakref
 import dask
 
 from .spec import SpecCluster, ProcessInterface
-from ..utils import cli_keywords
-from ..scheduler import Scheduler as _Scheduler
-from ..worker import Worker as _Worker
 from ..utils import serialize_for_cli
 
 logger = logging.getLogger(__name__)
@@ -71,7 +69,8 @@ class Worker(Process):
         address: str,
         connect_options: dict,
         kwargs: dict,
-        worker_module="distributed.cli.dask_worker",
+        worker_module="deprecated",
+        worker_class="dask.distributed.Nanny",
         remote_python=None,
         loop=None,
         name=None,
@@ -80,7 +79,13 @@ class Worker(Process):
 
         self.address = address
         self.scheduler = scheduler
-        self.worker_module = worker_module
+        if worker_module != "deprecated":
+            raise ValueError(
+                "The worker_module= keyword has been deprecated. "
+                "Please use worker_class= and instead. "
+                "You will have to specify a Python class rather than a CLI command."
+            )
+        self.worker_class = worker_class
         self.connect_options = connect_options
         self.kwargs = kwargs
         self.name = name
@@ -112,12 +117,17 @@ class Worker(Process):
                 set_env,
                 self.remote_python or sys.executable,
                 "-m",
-                self.worker_module,
+                "distributed.cli.dask_spec",
                 self.scheduler,
-                "--name",
-                str(self.name),
+                "--spec",
+                "'%s'"
+                % json.dumps(
+                    {
+                        "cls": self.worker_class,
+                        "opts": {**self.kwargs, "name": self.name,},
+                    }
+                ),
             ]
-            + cli_keywords(self.kwargs, cls=_Worker, cmd=self.worker_module)
         )
 
         self.proc = await self.connection.create_process(cmd)
@@ -190,9 +200,13 @@ class Scheduler(Process):
                 set_env,
                 self.remote_python or sys.executable,
                 "-m",
-                "distributed.cli.dask_scheduler",
+                "distributed.cli.dask_spec",
+                "--spec",
+                "'%s'"
+                % json.dumps(
+                    {"cls": "dask.distributed.Scheduler", "opts": self.kwargs,}
+                ),
             ]
-            + cli_keywords(self.kwargs, cls=_Scheduler)
         )
         self.proc = await self.connection.create_process(cmd)
 
@@ -233,7 +247,8 @@ def SSHCluster(
     connect_options: Union[List[dict], dict] = {},
     worker_options: dict = {},
     scheduler_options: dict = {},
-    worker_module: str = "distributed.cli.dask_worker",
+    worker_module: str = "deprecated",
+    worker_class: str = "dask.distributed.Nanny",
     remote_python: str = None,
     **kwargs,
 ):
@@ -271,8 +286,8 @@ def SSHCluster(
         Keywords to pass on to workers.
     scheduler_options: dict, optional
         Keywords to pass on to scheduler.
-    worker_module: str, optional
-        Python module to call to start the worker.
+    worker_class: str, optional
+        Python class to use to create the worker
     remote_python: str, optional
         Path to Python on remote nodes.
 
@@ -288,14 +303,14 @@ def SSHCluster(
     >>> client = Client(cluster)
 
     An example using a different worker module, in particular the
-    ``dask-cuda-worker`` command from the ``dask-cuda`` project.
+    ``CUDAWorker`` command from the ``dask-cuda`` project.
 
     >>> from dask.distributed import Client, SSHCluster
     >>> cluster = SSHCluster(
     ...     ["localhost", "hostwithgpus", "anothergpuhost"],
     ...     connect_options={"known_hosts": None},
     ...     scheduler_options={"port": 0, "dashboard_address": ":8797"},
-    ...     worker_module='dask_cuda.dask_cuda_worker')
+    ...     worker_class='dask_cuda.CUDAWorker')
     >>> client = Client(cluster)
 
     See Also
@@ -304,6 +319,12 @@ def SSHCluster(
     dask.distributed.Worker
     asyncssh.connect
     """
+    if worker_module != "deprecated":
+        raise ValueError(
+            "The worker_module= keyword has been deprecated. "
+            "Please use worker_class= and instead. "
+            "You will have to specify a Python class rather than a CLI command."
+        )
     if set(kwargs) & old_cluster_kwargs:
         from .old_ssh import SSHCluster as OldSSHCluster
 
@@ -345,7 +366,7 @@ def SSHCluster(
                 if isinstance(connect_options, dict)
                 else connect_options[i + 1],
                 "kwargs": worker_options,
-                "worker_module": worker_module,
+                "worker_class": worker_class,
                 "remote_python": remote_python,
             },
         }
--- a/distributed/deploy/tests/test_ssh.py
+++ b/distributed/deploy/tests/test_ssh.py
@@ -42,12 +42,7 @@ async def test_keywords():
         ["127.0.0.1"] * 3,
         connect_options=dict(known_hosts=None),
         asynchronous=True,
-        worker_options={
-            "nprocs": 2,  # nprocs checks custom arguments with cli_keywords
-            "nthreads": 2,
-            "memory_limit": "2 GiB",
-            "death_timeout": "5s",
-        },
+        worker_options={"nthreads": 2, "memory_limit": "2 GiB", "death_timeout": "5s",},
         scheduler_options={"idle_timeout": "10s", "port": 0},
     ) as cluster:
         async with Client(cluster, asynchronous=True) as client:


pentschev:
> This commit speculatively changes documentation here in order to suggest a CUDAWorker class type.

If only I had done this already in the past... 😆 

https://github.com/rapidsai/dask-cuda/pull/35/files#diff-a77f0c6f19d8d34d59aede5e31455719R39

distributed/deploy/ssh.py
```diff
@@ -80,7 +79,13 @@ def __init__(
 
         self.address = address
         self.scheduler = scheduler
-        self.worker_module = worker_module
+        if worker_module != "deprecated":
+            raise ValueError(
+                "The worker_module= keyword has been deprecated. "
+                "Please use worker_class= and instead. "
+                "You will have to specify a Python class rather than a CLI command."
+            )
```
pentschev, line 87: Would it be possible to deprecate without raising an error/removing this for now?

mrocklin:
Ha!  OK, well great.  More broadly, do you know people who use dask-ssh
with dask-cuda?  If so, do you have any concerns about how this change
might affect them?

distributed/deploy/ssh.py
mrocklin: This would require keeping around the old implementation.  Given how rare modifying this parameter is, and the current lack of maintenance efforts, I'm inclined to make this a hard change if that is palatable to others.

pentschev:
> Ha!  OK, well great.  More broadly, do you know people who use dask-ssh with dask-cuda?  If so, do you have any concerns about how this change might affect them?

We have some internal uses, as well as multi-node testing/benchmarking: https://github.com/rapidsai/dask-cuda/blob/branch-0.15/dask_cuda/benchmarks/utils.py#L124-L151 . With this change and until we can implement a new `CUDAWorker` class it seems that we'll lose the ability to run `SSHCluster`s entirely, which isn't great for us.

distributed/deploy/ssh.py
pentschev: Actually, IIRC, this was specifically introduced for dask-cuda by @jacobtomlinson , which is used whenever we need an `SSHCluster` with CUDA capability: https://github.com/rapidsai/dask-cuda/blob/branch-0.15/dask_cuda/benchmarks/utils.py#L147 .

I understand your point in wanting to get rid of it and I support that, but I would prefer to have some time before that gets completely dropped. Is there an immediate use case for `worker_class`? If there isn't, I would argue in favor of keeping the current existing use case against of only supporting future uses.

mrocklin: > Is there an immediate use case for worker_class?

Yes, the keyword handling is awkward otherwise.  This came up in a user issue that is referenced in the header comment of this issue.

> but I would prefer to have some time before that gets completely dropped

Yes, time makes sense.  How much time would you need?

pentschev: I'm guessing we'll have 2.19 release this week, so I'd be hoping not to drop it until 2.21-2.22, this should give us some 4 weeks to work on that. Given that we have some very pressing priorities at the moment, I don't think any of us working on dask-cuda would have time to adapt to this change before that.

mrocklin: This is probably a one-line change though I would think?

```diff
-  worker_module="dask_cuda.cli.dask_cuda",
+  worker_class="dask_cuda.CUDAWorker",
```

When is the next RAPIDS release?

quasiben:
If dask-cuda added a CUDAWorker, would that help other dask deployment tools ? For example, dask-gateway users are interested in cuda things: https://github.com/dask/dask-gateway/issues/177 -- or possibly dask-jobqueue (something I am personally interested in)

This is probably a direction we would want to take but as @pentschev this would break things short term

mrocklin:
> If dask-cuda added a CUDAWorker, would that help other dask deployment tools ? For example, dask-gateway users are interested in cuda things: dask/dask-gateway#177 -- or possibly dask-jobqueue (something I am personally interested in)

Right now most dask-foo projects can take in an alternative CLI command, dask-gateway being the exception (but this is possible with some work). My experience has been that there is some friction about using the Python API to spell out CLI commands, like converting `{"memory_limit": "4GiB"}` to `--memory-limit 4GiB`.  I've found that dask-spec handles this more cleanly because it accepts arguments on the command line like `{"memory_limit": "4GiB"}`, and I wouldn't mind converting everyone over to using that.  In general I think that this reduces user friction a bit.

In general this is fine, except in the rare case where people have custom `dask-foo-worker` CLI commands.  The only public one that I know of that composes with these projects is `dask-cuda-worker`.

If we need to wait a few weeks to ensure a smooth transition with RAPIDS projects that makes sense to me.  If RAPIDS folks are too busy and this needs to wait ten weeks then I think that that seems off, especially given how small I expect this change will be on the RAPIDS side (we may spend more time discussing this than will take implementing the change).

mmccarty:
@gforsyth @dankerrigan - I've confirmed that our `dask-ssh` deployment won't be impacted by this change, but one of you should weigh in on `dask-gateway` + `dask-cuda`.

mrocklin:
dask-gateway uses its own dask-worker command that is, today, pretty deeply tied to dask-gateway.  There is an issue that talks about decoupling these.  It shouldn't be that hard but probably needs someone who needs it done to motivate the work.

quasiben:
> If we need to wait a few weeks to ensure a smooth transition with RAPIDS projects that makes sense to me. If RAPIDS folks are too busy and this needs to wait ten weeks then I think that that seems off, especially given how small I expect this change will be on the RAPIDS side (we may spend more time discussing this than will take implementing the change).

Definitely agreed!

distributed/deploy/ssh.py
quasiben: Next RAPIDS release is Aug 5th but we should be able to get this done well before then.

pentschev: Sorry, I think I was misleading in https://github.com/dask/distributed/pull/3907#issuecomment-644830382 , what I meant was that I implemented it in the past and we decided not to use it to reduce the maintenance cost. So it's not just a one line change, but we will need to implement a whole new class to match this new requirement.

mrocklin: Ah, I see.  We went through this process in dask-distributed where we took a lot of the logic in the CLI and moved it into the Worker/Nanny classes.  In general it was a nice win.  It made things a lot easier for us to test, and helped to reduce code duplication with other projects.

pentschev: Yes, I think we can do the same, I'll try to work on that soon, it may not be very complicated to do so, the difficult part is ensuring we don't introduce new bugs -- which we often do on such changes -- due to the spawning of new processes. This is particularly painful for UCX support, as we don't yet have full CI capabilities for multi-GPU/multi-node, so we have to test parts of it manually.

mrocklin: My hope is that is mostly a copy-paste job from the existing dask-cuda CLI code and is mostly around changing how we handle inputs from Click to normal Python keywords.  *Hopefully* all of the tricky logic can be moved over without touching it too much.

jacobtomlinson: I'd be happy to take a run that this. @pentschev do you have any pointers on testing to avoid regressions.

pentschev: First and foremost we should ensure the current tests pass on a DGX-1/DGX-2, then we'll need to do some multi-node/multi-GPU testing. I'd get a PR ready first and then we sync on additional tests we should run, as we don't have many of them automated yet.

pentschev:
Thanks to @jacobtomlinson for the work in https://github.com/rapidsai/dask-cuda/pull/324, dask-cuda should now be capable to handle changes from this PR, please feel free to move on with it! 😄

mrocklin:
Thanks for the update @pentschev and the work @jacobtomlinson .  Is there a release date that we should wait for on the RAPIDS side?

pentschev:
> Is there a release date that we should wait for on the RAPIDS side?

No, we can work with that change immediately, we can target upstream dask/distributed without problems.

distributed/deploy/ssh.py
```diff
@@ -112,12 +117,17 @@ async def start(self):
...
+                "distributed.cli.dask_spec",
                 self.scheduler,
-                "--name",
-                str(self.name),
+                "--spec",
+                "'%s'"
+                % json.dumps(
+                    {
+                        "cls": self.worker_class,
+                        "opts": {**self.kwargs, "name": self.name,},
```
pentschev, line 127: Is the `"name"` parameter here necessary? This is causing errors, probably because `CUDAWorker` will create multiple workers at once, one per GPU:

```python
distributed.nanny - ERROR - Failed while trying to start worker process: Unexpected response from register: {'status': 'error', 'message': 'name taken, 0', 'time': 1593107200.1947834}
distributed.nanny - ERROR - Failed to start worker
  File "/datasets/pentschev/miniconda3/envs/rn-102-0.15.0b200625/lib/python3.7/site-packages/distributed/nanny.py", line 766, in run
    await worker
  File "/datasets/pentschev/miniconda3/envs/rn-102-0.15.0b200625/lib/python3.7/site-packages/distributed/core.py", line 290, in _

    await self.start()

  File "/datasets/pentschev/miniconda3/envs/rn-102-0.15.0b200625/lib/python3.7/site-packages/distributed/worker.py", line 1093, in start
    await self._register_with_scheduler()
  File "/datasets/pentschev/miniconda3/envs/rn-102-0.15.0b200625/lib/python3.7/site-packages/distributed/worker.py", line 847, in _register_with_scheduler
    raise ValueError("Unexpected response from register: %r" % (response,))
```

The `--name` parameter was also being used before but wasn't causing any issues, I don't know why that happens with the spec. I tried just removing `"name": self.name,` and that alone suffices to resolve the issue with `CUDAWorker`.

mrocklin: Typically the cli commands will take a name parameter and expand it out to name-0, name-1, and so on.  I suspect that this is what the dask-cuda cli was doing before?  Maybe we could replicate that logic in dask-cuda?

pentschev: Yes, we're doing that. I now found that the name we're passing is actually an integer instead of a string, that's why it fails now, it was safe before, even though `name` is indeed expected to be a string. I will fix that on our end, but maybe it's still worth to keep the `str(self.name)` in case someone does the same mistake we did?

pentschev: Actually I can't fix it on our end. We never pass the `name` argument from `SSHCluster` to the `Worker` class. We would either need to keep the `str(self.name)` or add a way to pass it via `SSHCluster`.

pentschev: Or even better, just get rid of `name` in `Worker` ? This would then allow `name` from `kwargs`, if I'm not missing anything.

pentschev: Argh, no that won't work either, `name` is part of the spec. I'm really not sure what's the correct way of handling this.

pentschev: Ok, so I took another look and the `name` argument passed to `Worker` is defined and coming from https://github.com/dask/distributed/blob/44b2358e33a0738c4c70ca96db4242636245e07d/distributed/deploy/spec.py#L318-L325 , as an integer. We need to cast it to a string, either here or in dask-cuda. I have no strong feelings but it seems like here would be more appropriate, perhaps someone else would have a similar issue should they use a custom class as well.

pentschev: @mrocklin any news on this PR ?

pentschev:
@mrocklin do you still intend to work on this? We're fine on the dask-cuda side if you do.

mrocklin:
I'm not actively working on it right now, but only due to lack of time, not
lack of interest

pentschev:
Since #5191 has been merged, we can probably close this one, what do you think @jacobtomlinson @mrocklin @quasiben ?

jrbourbeau:
Thanks for following-up here @pentschev. I think it's safe to close this PR

PR #3910, opened 2020-06-17
Prune virtual client on variable deletion

### LINKED ISSUE #3899: Variables leak virtual clients
**What happened**:

```python
def _get_number_of_clients(dask_scheduler: Optional[Scheduler] = None) -> Optional[int]:
    if dask_scheduler is None:
        return None
    else:
        return len(dask_scheduler.clients)

n_clients1 = client.run_on_scheduler(_get_number_of_clients)
assert n_clients1 is not None

for i in range(100):
    future = create_some_dummy_future()
    var = Variable(f"var-{i}", client)
    var.set(future)
    future.cancel(force=True)
    var.delete()

n_clients2 = client.run_on_scheduler(_get_number_of_clients)
assert n_clients2 is not None

assert n_client1 == n_clients2
```

This fails, because here:

https://github.com/dask/distributed/blob/1d7640b0172febf9ceef37c2c31241c66ac165eb/distributed/scheduler.py#L2333-L2347

a new virtual client for each and every variable is created but here:

https://github.com/dask/distributed/blob/1d7640b0172febf9ceef37c2c31241c66ac165eb/distributed/scheduler.py#L2349-L2361

we never clean up / prune these clients again.

Scheduler-side state is not leaky.

**Environment**:

- Dask version: 2.17.2
- Distributed version: 2.17.0
- Python version: 3.6
- Operating System: Linux

marco-neumann-jdas:
Fixes #3899

### DIFF
--- a/distributed/tests/test_variable.py
+++ b/distributed/tests/test_variable.py
@@ -263,3 +263,22 @@ def test_future_erred_sync(client):
 
     with pytest.raises(ZeroDivisionError):
         future2.result()
+
+
+@gen_cluster(client=True)
+async def test_variables_do_not_leak_client(c, s, a, b):
+    # https://github.com/dask/distributed/issues/3899
+    clients_pre = set(s.clients)
+
+    # setup variable with future
+    x = Variable("x")
+    future = c.submit(inc, 1)
+    await x.set(future)
+
+    # complete teardown
+    x.delete()
+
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks @marco-neumann-jdas !  This looks great!

PR #3912, opened 2020-06-18
Fix exception handling in `_wait_until_connected`

sk1p:
In the nanny, if there is an exception in the `init_result_q`, the dict was raised, instead of only the exception, leading to a `TypeError: exceptions must derive from BaseException`

### DIFF
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -693,7 +693,7 @@ class WorkerProcess:
                     "Failed while trying to start worker process: %s", msg["exception"]
                 )
                 await self.process.join()
-                raise msg
+                raise msg["exception"]
             else:
                 return msg
 
--- a/distributed/tests/test_nanny.py
+++ b/distributed/tests/test_nanny.py
@@ -557,3 +557,20 @@ async def test_nanny_closed_by_keyboard_interrupt(cleanup, protocol):
             await n.process.stopped.wait()
             # Check that the scheduler has been notified about the closed worker
             assert len(s.workers) == 0
+
+
+class StartException(Exception):
+    pass
+
+
+class BrokenWorker(worker.Worker):
+    async def start(self):
+        raise StartException("broken")
+
+
+@pytest.mark.asyncio
+async def test_worker_start_exception(cleanup):
+    # make sure this raises the right Exception:
+    with pytest.raises(StartException):
+        async with Nanny("tcp://localhost:1", worker_class=BrokenWorker) as n:
+            await n.start()


sk1p:
Not sure what the CI failure is caused by - msgpack trying to serialize some object graph with a cycle?

mrocklin:
Thank you for the PR @sk1p .  I agree that the CI failure is likely unrelated.

The code fix looks good to me.  I'm curious, is it easy to make a test for this behavior?  It would be good to ensure that we don't lose this fix in the future.

sk1p:
> I'm curious, is it easy to make a test for this behavior? It would be good to ensure that we don't lose this fix in the future.

True, I didn't have an idea how to do this yet, but I'll have a look.

sk1p:
Added a test case. When only running this test, I get these messages after the pytest run completes:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-2' coro=<Nanny._on_exit() done, defined at /home/alex/source/distributed/distributed/nanny.py:440> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f8a00dc8760>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/alex/.virtualenvs/distributed/lib/python3.8/site-packages/tornado/ioloop.py:690]>
Task was destroyed but it is pending!
task: <Task pending name='Task-3' coro=<BaseTCPConnector.connect() done, defined at /home/alex/source/distributed/distributed/comm/tcp.py:337> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f8a00dc8e50>()]> cb=[_release_waiter(<Future pendi...a00dc8760>()]>)() at /usr/lib/python3.8/asyncio/tasks.py:429]>
```

Is there some cleanup for the `Nanny` I'm missing?

mrocklin:
I apologize for the long delay in response @sk1p .  This looks good to me.  I'm going to go ahead and merge. 

> Is there some cleanup for the Nanny I'm missing?

This isn't your fault.  My guess is that there are some coroutines that aren't cleaned up if the Nanny doesn't clean it self up smoothly.  This would be nice to fix, but is probably ok though.  If you wanted to raise it as a follow-on issue that would be welcome, but not obligatory.

sk1p:
> I apologize for the long delay in response @sk1p . This looks good to me. I'm going to go ahead and merge.

No worries, and thanks for merging!

> This isn't your fault. My guess is that there are some coroutines that aren't cleaned up if the Nanny doesn't clean it self up smoothly. This would be nice to fix, but is probably ok though. If you wanted to raise it as a follow-on issue that would be welcome, but not obligatory.

Submitted as #3940

Thanks!

PR #3915, opened 2020-06-22
Fix race condition in repeated calls to cluster.adapt()

### LINKED ISSUE #3863: Calling "cluster.adapt()" multiple times causes strange behavior and `KilledWorker` errors
**What happened**:

We tried scaling an adaptive cluster twice like this:
```python
cluster.adapt(maximum=16)
cluster.adapt(maximum=8)
```

After submitting tasks to the cluster workers will crash with
```
KilledWorker: ('sample_pi_monte_carlo-c4b060d1-c020-4aa1-b1ea-9f7fadeb881a', <Worker 'tcp://10.56.4.77:40643', name: 7, memory: 0, processing: 94>)
```
and logs on the scheduler show that it scales up the cluster then immediately retires the workers:
```
distributed.scheduler - INFO - Retire worker names (1, 2, 3, 4, 5, 6, 7, 8)
distributed.deploy.adaptive - INFO - Retiring workers [1, 2, 3, 4, 5, 6, 7, 8]
distributed.scheduler - INFO - Register worker <Worker 'tcp://10.56.11.43:33153', name: 12, memory: 0, processing: 0>
distributed.scheduler - INFO - Starting worker compute stream, tcp://10.56.11.43:33153
distributed.scheduler - INFO - Register worker <Worker 'tcp://10.56.6.29:35277', name: 6, memory: 0, processing: 0>
distributed.scheduler - INFO - Starting worker compute stream, tcp://10.56.6.29:35277
[more register worker / starting worker messages]
...
[more remove worker / removing comms messages]
distributed.scheduler - INFO - Remove worker <Worker 'tcp://10.56.6.29:35277', name: 6, memory: 14, processing: 85>
distributed.core - INFO - Removing comms to tcp://10.56.6.29:35277
distributed.scheduler - INFO - Remove worker <Worker 'tcp://10.56.4.77:40643', name: 7, memory: 14, processing: 94>
distributed.core - INFO - Removing comms to tcp://10.56.4.77:40643
```

Calling `cluster.adapt()` multiple times should have no effect if the cluster size is within bounds since you're only changing the minimum and maximum number of workers.

```python
import time

import dask
import numpy as np
from dask_kubernetes import KubeCluster

cluster = KubeCluster.from_dict(...)
client = Client(cluster)

cluster.adapt(maximum=16)
cluster.adapt(maximum=8)

def sample_pi_monte_carlo(sleep_time):
    x = np.random.uniform(-1, 1)
    y = np.random.uniform(-1, 1)
    time.sleep(sleep_time)
    return (np.sqrt(x**2 + y**2) < 1)

results = client.compute([
    dask.delayed(sample_pi_monte_carlo)(0.1)
    for _ in range(1000)
], sync=True)

print("pi is:", 4 * np.mean(results))
```

This uses `dask_kubernetes` to implement adaptive scaling. I originally reported this at https://github.com/dask/dask-kubernetes/issues/250 but it might be an upstream issue (their `KubeCluster` class inherits from `SpecCluster` doesn't add any logic to `.adapt()`)

**Environment**:

- Dask version: 2.15.0
- Python version: 3.7
- Operating System: Ubuntu 18

Let me know if there's anything else I can do to help!

jacobtomlinson:
Closes #3863.

There was a race condition where an `Adaptive` periodic callback was being created and started in a deferred function. This meant that if `adapt` was called again in quick succession the `stop` method which removed the periodic callback would be called before the `start` method, resulting in multiple periodic callbacks being scheduled.

The reference to the old `Adaptive` object was being replaced so this bug was intermittent depending on how fast the old object was garbage collected.

This PR instantiates the `PeriodicCallback` sequentially and only defers the `start` call. We also catch when `start` raises an `AttributeError` due to the callback already being stopped and nullified.

### DIFF
--- a/distributed/deploy/adaptive_core.py
+++ b/distributed/deploy/adaptive_core.py
@@ -75,11 +75,13 @@ class AdaptiveCore:
         self.maximum = maximum
         self.wait_count = wait_count
         self.interval = parse_timedelta(interval, "seconds") if interval else interval
-        self.periodic_callback = None
+        self.periodic_callback = PeriodicCallback(self.adapt, self.interval * 1000)
 
         def f():
-            self.periodic_callback = PeriodicCallback(self.adapt, self.interval * 1000)
-            self.periodic_callback.start()
+            try:
+                self.periodic_callback.start()
+            except AttributeError:
+                pass
 
         if self.interval:
             try:
--- a/distributed/deploy/adaptive_core.py
+++ b/distributed/deploy/adaptive_core.py
@@ -75,7 +75,7 @@ class AdaptiveCore:
         self.maximum = maximum
         self.wait_count = wait_count
         self.interval = parse_timedelta(interval, "seconds") if interval else interval
-        self.periodic_callback = PeriodicCallback(self.adapt, self.interval * 1000)
+        self.periodic_callback = None
 
         def f():
             try:
@@ -84,6 +84,7 @@ class AdaptiveCore:
                 pass
 
         if self.interval:
+            self.periodic_callback = PeriodicCallback(self.adapt, self.interval * 1000)
             try:
                 self.loop.add_callback(f)
             except AttributeError:
--- a/distributed/deploy/tests/test_adaptive.py
+++ b/distributed/deploy/tests/test_adaptive.py
@@ -430,3 +430,22 @@ def test_adaptive_config():
         assert adapt.maximum == math.inf
         assert adapt.interval == 5
         assert adapt.wait_count == 8
+
+
+@pytest.mark.asyncio
+async def test_update_adaptive(cleanup):
+    async with LocalCluster(
+        0,
+        threads_per_worker=2,
+        memory_limit="3 GB",
+        scheduler_port=0,
+        silence_logs=False,
+        processes=False,
+        dashboard_address=None,
+        asynchronous=True,
+    ) as cluster:
+        first = cluster.adapt(maxmimum=1)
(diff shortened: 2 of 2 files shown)


distributed/deploy/adaptive_core.py
```diff
@@ -78,10 +78,13 @@ def __init__(
...
         def f():
-            self.periodic_callback = PeriodicCallback(self.adapt, self.interval * 1000)
-            self.periodic_callback.start()
+            try:
+                self.periodic_callback.start()
+            except AttributeError:
+                pass
 
         if self.interval:
+            self.periodic_callback = PeriodicCallback(self.adapt, self.interval * 1000)
```
mrocklin, line 87: Do we ever call this from a thread that isn't the IOLoop?  If so, does this pick up the IOLoop at instantiation time or at start time?  Usually when we use `add_callback` it's because we're handling something that is sensitive to where it gets started and we're not always on the event loop.

mrocklin:
Thanks for taking this on @jacobtomlinson

distributed/deploy/adaptive_core.py
jacobtomlinson: The loop is picked up at start time. 

https://www.tornadoweb.org/en/stable/_modules/tornado/ioloop.html#PeriodicCallback.start

There is nothing in the `PeriodicCallback.__init__` that makes me think this would be a problem.

mrocklin: Oh awesome.  Merging then.

PR #3922, opened 2020-06-23
Make `Worker.delete_data` sync

### LINKED ISSUE #3920: Await on handle_stream raises missing delete_data await warning
For increased visibility, I'm reposting https://github.com/dask/distributed/pull/3847/files#r443766556 as an issue here:

We have a few tests in dask-cuda that check the behavior of Device<->Host<->Disk spilling and I noticed after the 2.19 release one of them has broken, I managed to track it down to one specific line of code in https://github.com/dask/distributed/blob/44b2358e33a0738c4c70ca96db4242636245e07d/distributed/core.py#L573, introduced by https://github.com/dask/distributed/pull/3847. The test in question happens in https://github.com/rapidsai/dask-cuda/blob/branch-0.15/dask_cuda/tests/test_spill.py#L409-L411, where we assert that the zict dictionaries are empty after deleting `cdf2`, which is the object being spilled. It seems that this is because we're not awaiting for `Worker.delete_data` somewhere, as per the warning below that doesn't happen if I comment `await gen.sleep(0)` out:

```python
dask_cuda/tests/test_spill.py::test_cudf_device_spill[params0]
  /datasets/pentschev/miniconda3/envs/r-102-0.14/lib/python3.7/inspect.py:732: RuntimeWarning: coroutine 'Worker.delete_data' was never awaited
    for modname, module in list(sys.modules.items()):
```

I _think_ that the only place where `Worker.delete_data` would be called and should be awaited is in https://github.com/dask/distributed/blob/4f878b420b349ee725de5ef64fd5e664dedb8aba/distributed/scheduler.py#L2791-L2800, but I don't have anything better than my guess at this time because it's really hard for me to understand all the async black magic. I'm gonna continue trying to figure this out, but any suggestions on how to pinpoint that are appreciated!

pentschev:
This should fix issues where `Worker.delete_data` isn't awaited.

Fixes #3920

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1634,3 +1634,28 @@ async def test_heartbeat_comm_closed(cleanup, monkeypatch, reconnect):
                 else:
                     assert w.status == "closed"
     assert "Heartbeat to scheduler failed" in logger.getvalue()
+
+
+@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
+async def test_delete_data(client, s, w):
+    df = (
+        dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s")
+        .reset_index(drop=True)
+    )
+
+    await client.compute(df.map_partitions(lambda df: df.__sizeof__()))
+
+    df2 = df.persist()
+    await df2
+
+    def _check_data(empty):
+        if empty is True:
+            assert len(get_worker().data) == 0
+        else:
+            assert len(get_worker().data) > 0
+
+    del df
+    await client.run(_check_data, False)
+
+    del df2
+    await client.run(_check_data, True)
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1634,22 +1634,3 @@ async def test_heartbeat_comm_closed(cleanup, monkeypatch, reconnect):
                 else:
                     assert w.status == "closed"
     assert "Heartbeat to scheduler failed" in logger.getvalue()
-
-
-@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
-async def test_delete_data(c, s, w):
-    df = dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s").reset_index(
-        drop=True
-    )
-
-    df2 = df.persist()
-    await df2
-
-    del df
-    assert len(w.data) > 0
-
-    del df2
-    start = time()
-    while w.data:
-        await asyncio.sleep(0.01)
-        assert time() < start + 2
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1637,24 +1637,19 @@ async def test_heartbeat_comm_closed(cleanup, monkeypatch, reconnect):
 
 
 @gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
-async def test_delete_data(client, s, w):
+async def test_delete_data(c, s, w):
     df = dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s").reset_index(
         drop=True
     )
 
-    await client.compute(df.map_partitions(lambda df: df.__sizeof__()))
-
     df2 = df.persist()
     await df2
 
-    def _check_data(empty):
-        if empty is True:
-            assert len(get_worker().data) == 0
-        else:
-            assert len(get_worker().data) > 0
-
     del df
-    await client.run(_check_data, False)
+    assert len(w.data) > 0
 
     del df2
-    await client.run(_check_data, True)
+    start = time()
+    while w.data:
+        await asyncio.sleep(0.01)
+        assert time() < start + 2
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1638,9 +1638,8 @@ async def test_heartbeat_comm_closed(cleanup, monkeypatch, reconnect):
 
 @gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
 async def test_delete_data(client, s, w):
-    df = (
-        dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s")
-        .reset_index(drop=True)
+    df = dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s").reset_index(
+        drop=True
     )
 
     await client.compute(df.map_partitions(lambda df: df.__sizeof__()))
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1352,12 +1352,6 @@ class Worker(ServerNode):
                     self.release_dep(key)
 
             logger.debug("Deleted %d keys", len(keys))
-            if report:
-                logger.debug("Reporting loss of keys to scheduler")
-                # TODO: this route seems to not exist?
-                self.scheduler.remove_keys(
-                    address=self.contact_address, keys=list(keys)
-                )
         return "OK"
 
     async def set_resources(self, **resources):
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1341,7 +1341,7 @@ class Worker(ServerNode):
         info = {"nbytes": {k: sizeof(v) for k, v in data.items()}, "status": "OK"}
         return info
 
-    async def delete_data(self, comm=None, keys=None, report=True):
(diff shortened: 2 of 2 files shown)


distributed/worker.py
```diff
@@ -1355,7 +1355,7 @@ async def delete_data(self, comm=None, keys=None, report=True):
             if report:
                 logger.debug("Reporting loss of keys to scheduler")
                 # TODO: this route seems to not exist?
-                await self.scheduler.remove_keys(
+                self.scheduler.remove_keys(
                     address=self.contact_address, keys=list(keys)
                 )
```
mrocklin, line 1360: If we don't await a coroutine then it won't occur.  So this code probably doesn't make sense.  

However, the TODO note just above it is interesting.  Maybe this code isn't necessary?  It might be worth looking at git blame here to see why it was added.

distributed/tests/test_worker.py
```diff
@@ -1634,3 +1634,27 @@ def bad_heartbeat_worker(*args, **kwargs):
...
     assert "Heartbeat to scheduler failed" in logger.getvalue()
+
+
+@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
+async def test_delete_data(client, s, w):
+    df = dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s").reset_index(
+        drop=True
+    )
+
+    await client.compute(df.map_partitions(lambda df: df.__sizeof__()))
```
mrocklin, line 1645: Is everything here necessary?

distributed/tests/test_worker.py
```diff
@@ -1634,3 +1634,27 @@ def bad_heartbeat_worker(*args, **kwargs):
...
+        if empty is True:
+            assert len(get_worker().data) == 0
+        else:
+            assert len(get_worker().data) > 0
+
+    del df
+    await client.run(_check_data, False)
+
+    del df2
+    await client.run(_check_data, True)
```
mrocklin, line 1660: To simplify this logic I recommend that you just check `w.data`

distributed/worker.py
pentschev: It's even harder to await a function that doesn't exist, which is the case here. 😄 

With `git log -c -S remove_keys distributed/scheduler.py`, I managed to find that this function was removed somewhere in between https://github.com/dask/distributed/commit/dc5474815788f75b88be38a4d4242e515c48ef2b and https://github.com/dask/distributed/commit/2c175e86e0b1d43705510cf21ede408ccd71110d . Unfortunately that dates back to 2016, I believe this was a leftover from some commit around that time, and I can remove that line if it makes sense, but unfortunately I can't dig in deeper.

distributed/tests/test_worker.py
pentschev: Seems like getting rid of `await client.compute(df.map_partitions(lambda df: df.__sizeof__()))` works just the same, so just the dataframe creation suffices.

distributed/tests/test_worker.py
pentschev: I assume you're suggesting `assert len(w.data) > 0`/`assert len(w.data) == 0`, is that right? If so, that unfortunately fails on the second assertion, could this be some async black magic? E.g., `client.run` will force some syncing while `w.data` is still outdated?

distributed/worker.py
jakirkham: Yeah I'd be +1 to just dropping this and the `TODO`.

distributed/tests/test_worker.py
mrocklin: client.run doesn't force any synchronization, but maybe the GC hasn't run yet at this point.  Regardless we should probably make this robust to GC taking a while (GC delays are a major source of intermittent testing failures for us).  Most tests of this form look like the following:

```python
start = time()
while w.data:
    await asyncio.sleep(0.01)
    assert time() < start + 2
```

distributed/worker.py
pentschev: I tentatively dropped it on the newest commit, if there are any concerns with that I'll revert it.

distributed/tests/test_worker.py
pentschev: Removed in latest commits.

distributed/tests/test_worker.py
pentschev: Thanks for the explanation @mrocklin , your suggestion works and that's fixed in latest commits.

distributed/tests/test_worker.py
```diff
@@ -1634,3 +1634,22 @@ def bad_heartbeat_worker(*args, **kwargs):
                 else:
                     assert w.status == "closed"
     assert "Heartbeat to scheduler failed" in logger.getvalue()
+
+
+@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
+async def test_delete_data(c, s, w):
+    df = dask.datasets.timeseries(dtypes={"x": int, "y": float}, freq="1s").reset_index(
+        drop=True
+    )
```
mrocklin, line 1643: Do we need a dataset this complex?  Do we need these datatypes or the reset_index call?  If not, lets remove them to keep things as simple as possible for future reviewers.

Almost all tests in this file intentionally use very simple computations, like `future = client.submit(inc, 1)`.  I recommend that we do the same here unless having a dataframe with a reset index is critical to test the delete_data operation.

distributed/tests/test_worker.py
```diff
@@ -1634,3 +1634,22 @@ def bad_heartbeat_worker(*args, **kwargs):
...
+
+    df2 = df.persist()
+    await df2
+
+    del df
+    assert len(w.data) > 0
```
mrocklin, line 1649: Note that this deletion almost certainly hasn't gone through yet.  I'm not sure what you're trying to test here.

mrocklin: Actually, can you help explain the objective of this test?  Are there not already tests that verify that data gets deleted from workers?

distributed/tests/test_worker.py
pentschev: Actually, no. I can simplify that a bit.

distributed/tests/test_worker.py
pentschev: What I'm trying to test here is that `Worker.delete_data` is awaited (see https://github.com/dask/distributed/issues/3920#issue-643386582), or sync in the case of this PR. The test here is a simplification of the test in https://github.com/rapidsai/dask-cuda/blob/5c7942d6a62d630475fa51dd6b80fcf469bca142/dask_cuda/tests/test_spill.py#L358-L411, but I also noticed that the test here does not reproduce that error 100% accurately. I'm thus not sure that error will be easy to reproduce, so I'm open to suggestions on a better way to handle this here, should we remove this test and just make `delete_data` sync, as is the focus of this PR?

pentschev: Friendly ping here @mrocklin . :)

mrocklin: Hi @pentschev .  Thanks for the friendly ping.  My apologies for the delay.

What I'm hearing is that you want to test that Dask deletes data once references to that data are deleted.  I would be very surprised if we didn't already have tests for this.  It might be worth looking around in the test_scheduler.py or test_client.py file if that is your goal.

After looking at the test that you pointed to briefly I'm afraid to say that I wasn't able to understand the intent of that test.  I'm not sure how to help on that topic.

pentschev: As mentioned, making `delete_data` sync seems to be the right thing to do. Before that there were corner cases where it would not be awaited and current tests in distributed seem not to pick this case up, I was trying to come up with a small test to ensure that, but it has proven very challenging, even though the dask-cuda test did pick that up. I'm fine with just removing the test completely but keeping the change (removing `async` from `delete_data`). What do you say about this?

mrocklin: > What do you say about this?

I have high confidence that we're already testing data deletion.  I'm entirely ok dropping this test entirely.

pentschev:
@mrocklin I dropped the test and left only the actual code changes. From my side it's good to go, let me know if there are other things to be addressed.

distributed/tests/test_worker.py
pentschev: Done.

mrocklin:
This is in.  Thanks @pentschev

pentschev:
Thanks @mrocklin for reviewing and merging!

PR #3924, opened 2020-06-23
WIP: Add option to enable RMM logging

jakirkham:
Related to PR ( https://github.com/rapidsai/dask-cuda/pull/322 )

Includes an option to enable RMM logging for Dask-CUDA workers.

cc @pentschev @quasiben

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -101,10 +101,15 @@ def init_once():
                 )
 
     pool_size_str = dask.config.get("rmm.pool-size")
+    enable_logging = dask.config.get("rmm.logging")
     if pool_size_str is not None:
         pool_size = parse_bytes(pool_size_str)
         rmm.reinitialize(
-            pool_allocator=True, managed_memory=False, initial_pool_size=pool_size
+            pool_allocator=True,
+            managed_memory=False,
+            initial_pool_size=pool_size,
+            logging=enable_logging,
+            log_file_name="",  # Something more useful here ;)
         )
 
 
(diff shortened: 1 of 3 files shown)


distributed/distributed-schema.yaml
```diff
@@ -770,6 +770,12 @@ properties:
         - "null"
         description:
           The size of the memory pool in bytes
+      logging:
+        type:
+        - bool
```
quasiben, line 775: I think the errors are related to the schema checking. 
```suggestion
        - boolean
```

jakirkham: Thanks! :)

quasiben:
@jakirkham I pushed a commit which should fix the schema issue

quasiben:
@jakirkham can you move this PR from Draft to Review ?

distributed/comm/ucx.py
```diff
@@ -101,10 +101,15 @@ def device_array(n):
...
+    enable_logging = dask.config.get("rmm.logging")
     if pool_size_str is not None:
         pool_size = parse_bytes(pool_size_str)
         rmm.reinitialize(
-            pool_allocator=True, managed_memory=False, initial_pool_size=pool_size
+            pool_allocator=True,
+            managed_memory=False,
+            initial_pool_size=pool_size,
+            logging=enable_logging,
+            log_file_name="",  # Something more useful here ;)
```
jakirkham, line 112: We still need to fill this in with something. This is a bit tricky to get right as we would presumably want a different location depending on whether we are setting this on a worker, on the scheduler, or potentially on the client. However I'm not sure how we access that information from a context that is not inside one of those classes. Welcome thoughts on that though 🙂

pentschev: Maybe we should force the user to pass a path+filename prefix if logging is enabled and append the PID to it?

PR #3926, opened 2020-06-26
Test that tasks holding the GIL for too long are marked suspicious

jonashaag:
Using `usleep` in `PyDLL` mode is a simple and reliable way to hold the GIL using only "pure Python" code.

cc @fjetter

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1890,6 +1890,40 @@ async def test_failing_task_increments_suspicious(client, s, a, b):
     )
 
 
+@gen_cluster(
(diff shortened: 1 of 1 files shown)


fjetter:
I'd be curious what other people think about the usefulness of such a thing (we started building tests with this in our infrastructure since this is a problem for us).
If we want to include this here, a good place for a `hold_gil` function would probably be in https://github.com/dask/distributed/blob/master/distributed/utils_test.py

mrocklin:
At first glance I like it.  I've wanted a hold_gil function before.  I would be happy putting that in `distributed/utils_test.py`

PR #3927, opened 2020-06-26
update logic for worker removal in check ttl

quasiben:
PR updates #3896 

cc @jonashaag

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5304,8 +5304,11 @@ class Scheduler(ServerNode):
         now = time()
         for ws in self.workers.values():
             if (ws.last_seen < now - self.worker_ttl) and (
-                10 * heartbeat_interval(len(self.workers))
+                ws.last_seen < now - 10 * heartbeat_interval(len(self.workers))
             ):
+                print(
+                    now, ws.last_seen, now - 10 * heartbeat_interval(len(self.workers))
+                )
                 logger.warning(
                     "Worker failed to heartbeat within %s seconds. Closing: %s",
                     self.worker_ttl,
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -5304,8 +5304,11 @@ async def check_worker_ttl(self):
         now = time()
         for ws in self.workers.values():
             if (ws.last_seen < now - self.worker_ttl) and (
-                10 * heartbeat_interval(len(self.workers))
+                ws.last_seen < now - 10 * heartbeat_interval(len(self.workers))
             ):
+                print(
+                    now, ws.last_seen, now - 10 * heartbeat_interval(len(self.workers))
+                )
```
mrocklin, line 5311:
```suggestion
```

distributed/tests/test_failed_workers.py
```diff
@@ -417,14 +417,19 @@ async def test_restart_timeout_on_long_running_task(c, s, a):
...
+
     a.periodic_callbacks["heartbeat"].stop()
     await asyncio.sleep(0.010)
     assert set(s.workers) == {a.address, b.address}
 
     start = time()
     while set(s.workers) == {a.address, b.address}:
-        await asyncio.sleep(0.050)
-        assert time() < start + 2
+        await asyncio.sleep(interval)
```
mrocklin, line 432: How long does this test take to run?  If it's longer than a second can we mark it with `pytest.mark.slow`.  If it's longer than 10 seconds then we might want to find some other way around this.

mrocklin:
Thanks for doing this @quasiben !

distributed/scheduler.py
quasiben: Oops.  Thanks!

distributed/tests/test_failed_workers.py
quasiben: This takes ~7 seconds. 
> real    0m7.363s

Happy to make as slow if that's close enough to 10

mrocklin:
Let's definitely mark it as slow.

PR #3928, opened 2020-06-27
Make local directory if it does not exist

mrocklin:
Fixes https://stackoverflow.com/questions/62603609/create-local-directory-for-dask-jobqueue

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1460,6 +1460,15 @@ async def test_local_directory(s):
             assert "dask-worker-space" in w.local_directory
 
 
+@gen_cluster(nthreads=[])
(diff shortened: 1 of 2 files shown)


lsterzinger:
Hi @mrocklin, I'm the author of the SO question you linked. Looks like this small fix resolved my issue, thanks!

mrocklin:
Thank you for the well phrased question @lsterzinger and for taking the time to try this out and report success.

mrocklin:
Also, cc @lesteve who might have thoughts on if this policy change is good or bad.

quasiben:
I've probably bit but this in the past and unconsciously created the directory and moved on. +1 for merging in though I have limited dask-jq experience

PR #3933, opened 2020-06-29
Fix/task per key updates

### LINKED ISSUE #3930: Compute time per task with too many task types
I'm loving the new compute time per task plot.  I think that we could improve it in two ways:

1.  Format times better.  Often I see times like `1.528528e4 s`.  We might do this on the server side with `dask.utils.format_time`, or on the Bokeh side, however they manage formatting. 
2.  We should probably only show the top few entries:

image

Actually, this may be one situation where a pie chart makes more sense.  We don't need to give the same horizontal space to every task type.

quasiben:
Attempt to resolve #3930

This PR _keeps_ the original compute-per-task bar chart (I am a bit partial to the bar chart but I can remove) but limits which tasks are displayed as @mrocklin suggested:

Screen Shot 2020-06-29 at 12 29 02 PM

While the removal of keys provides easier digestion of data I wonder if users would wonder where the other keys were.  If we keep the bar chart I'd like to address this some how -- maybe a note on the bottom of the page ?

I also added the pie chart which was pleasant to try out:
Screen Shot 2020-06-29 at 12 16 54 PM

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -591,31 +591,32 @@ class ComputePerKey(DashboardComponent):
             )
 
             # keep only time which are 2% of max or greater
-            max_time = compute_times[0][1] * 0.02
-            compute_times = [(n, t) for n, t in compute_times if t > max_time]
-            compute_colors = list()
-            compute_names = list()
-            compute_time = list()
-            total_time = 0
-            for name, t in compute_times:
-                compute_names.append(name)
-                compute_colors.append(ts_color_of(name))
-                compute_time.append(t)
-                total_time += t
+            if compute_times:
+                max_time = compute_times[0][1] * 0.02
+                compute_times = [(n, t) for n, t in compute_times if t > max_time]
+                compute_colors = list()
+                compute_names = list()
+                compute_time = list()
+                total_time = 0
+                for name, t in compute_times:
+                    compute_names.append(name)
+                    compute_colors.append(ts_color_of(name))
+                    compute_time.append(t)
+                    total_time += t
 
-            angles = [t / total_time * 2 * math.pi for t in compute_time]
+                angles = [t / total_time * 2 * math.pi for t in compute_time]
 
-            self.fig.x_range.factors = compute_names
+                self.fig.x_range.factors = compute_names
 
-            compute_result = dict(
-                angles=angles,
-                times=compute_time,
-                color=compute_colors,
-                names=compute_names,
-                formatted_time=[format_time(t) for t in compute_time],
-            )
+                compute_result = dict(
+                    angles=angles,
+                    times=compute_time,
+                    color=compute_colors,
+                    names=compute_names,
+                    formatted_time=[format_time(t) for t in compute_time],
+                )
 
-            update(self.compute_source, compute_result)
+                update(self.compute_source, compute_result)
 
 
 class AggregateAction(DashboardComponent):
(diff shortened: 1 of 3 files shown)


distributed/dashboard/components/scheduler.py
```diff
@@ -535,7 +535,112 @@ def update(self):
...
+                color=compute_colors,
+                names=compute_names,
+                formatted_time=[format_time(t) for t in compute_time],
+            )
+
+            update(self.compute_source, compute_result)
+
+
+class ComputePerKeyPie(DashboardComponent):
+    """ Pie!!!!! chart showing time spend in action by key prefix"""
```
mrocklin, line 548: If we wanted to keep both I recommend that you take a look at [Bokeh tabs](http://docs.bokeh.org/en/1.0.2/docs/reference/models/widgets.panels.html) and then include two figures that are backed by the same `ColumnDataSource`

quasiben:
In https://github.com/dask/distributed/pull/3933/commits/146607465490437faa2634e52f158ff93d8b6937 I opted to use different `ColumnDataSource` objects as the `Bar Chat` is filtered before updating -- maybe there is a nice way of doing this in bokeh?  

Bar Chart             |  Pie Chart
:-------------------------:|:-------------------------:
Screen Shot 2020-06-29 at 1 50 42 PM  |  Screen Shot 2020-06-29 at 1 51 02 PM

Any other ideas on better names for tabs ?

mrocklin:
> Any other ideas on better names for tabs ?

They seem fine to me

> I opted to use different ColumnDataSource objects as the Bar Chat is filtered before updating

Thoughts on filtering the pie chart as well?  We might want to anyway if there is a legend?

mrocklin:
Looks like a test failure in CI

```python-traceback
 File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\tornado\web.py", line 1703, in _execute
    result = await result
  File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\bokeh\server\views\doc_handler.py", line 52, in get
    session = await self.get_session()
  File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\bokeh\server\views\session_handler.py", line 120, in get_session
    session = await self.application_context.create_session_if_needed(session_id, self.request, token)
  File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\bokeh\server\contexts.py", line 218, in create_session_if_needed
    self._application.initialize_document(doc)
  File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\bokeh\application\application.py", line 171, in initialize_document
    h.modify_document(doc)
  File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\bokeh\application\handlers\function.py", line 132, in modify_document
    self._func(doc)
  File "D:\a\distributed\distributed\distributed\dashboard\components\scheduler.py", line 2199, in individual_compute_time_per_key_doc
    component.update()
  File "C:\Miniconda3\envs\dask-distributed\lib\site-packages\bokeh\core\property\validation.py", line 93, in func
    return input_function(*args, **kwargs)
  File "D:\a\distributed\distributed\distributed\dashboard\components\scheduler.py", line 578, in update
    max_time = compute_times[0][1] * 0.02
```

quasiben:
Thanks @mrocklin for CI review.  I also added a small note on the bottom of the plots:

> Note: tasks less than 2% of max are not displayed

Screen Shot 2020-06-30 at 10 48 14 AM

Perhaps i am being overly concerned here

quasiben:
If there are no more comments I'll plan on merging tomorrow morning

mrocklin:
Thanks @quasiben !  I look forward to playing with this.

PR #3934, opened 2020-06-29
Restore Scheduler.time_started for Dask Gateway

TomAugspurger:
Removed in https://github.com/dask/distributed/pull/3830. Dask Gateway was using it. I'll make a PR to it to use the new name in the meantime.


mrocklin:
+1

My apologies for the churn

Issue #3932, opened 2020-06-29
Raise on Python major version mismatch?


mrocklin:
Currently we warn when the user has mismatched versions of numpy/pandas/etc..  We treat Python the same.

Currently this raises issues because of pickle-5 in some cases.

We may want to err hard in the case of a major Python version mismatch.

PR #3936, opened 2020-06-29
Err hard when Dask has mismatched versions or lz4 installed

mrocklin:
Fixes https://github.com/dask/distributed/issues/3932

### DIFF
--- a/distributed/tests/test_versions.py
+++ b/distributed/tests/test_versions.py
@@ -34,7 +34,7 @@ def kwargs_matching():
 
 
 def test_versions_match(kwargs_matching):
-    assert error_message(**kwargs_matching) == ""
+    assert error_message(**kwargs_matching)["warning"] == ""
 
 
 @pytest.fixture(params=["client", "scheduler", "worker-1"])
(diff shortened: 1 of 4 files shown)


distributed/tests/test_versions.py
```diff
@@ -142,3 +144,31 @@ def test_python_version():
...
     assert required["python"] == ".".join(map(str, sys.version_info))
+
+
+@gen_cluster()
+async def test_python_version_error(s, a, b):
+    s.workers[a.address].versions["packages"]["python"] = "3.5.1"
+
+    with pytest.raises(ImportError) as info:
+        async with Client(s.address, asynchronous=True) as client:
+            pass
```
mrocklin, line 155: For some reason this `pytest.raises` fails to capture the error on my machine. I'm not sure why.

abduhbm: It fails on my machine too. 

How about switching to `error_message` approach?
```python
def test_python_version_error(kwargs_matching):
    kwargs_matching["workers"]["worker-0"]["packages"]["python"] = "3.5.1"
    msg = error_message(**kwargs_matching)
    assert "Python" in msg["error"]
    assert "major" in msg["error"]
```

mrocklin:
Yes, we could, but I would feel more comfortable testing the user behavior
rather than the internal function if we can find a way to make this work.

abduhbm:
@mrocklin This works too

```python
def test_python_version_error():
    with pytest.raises(ImportError) as info:
        with LocalCluster(
            1,
            processes=False,
            scheduler_port=8786,
            silence_logs=False,
            dashboard_address=None,
        ) as c:
            c.scheduler.workers[c.workers[0].address].versions["packages"][
                "python"
            ] = "3.5.1"
            with Client(c.scheduler.address) as client:
                pass

    assert "Python" in str(info.value)
    assert "major" in str(info.value).lower()
```

PR #3937, opened 2020-06-29
Ensure `Worker.run*` handles `kwargs` correctly

jakirkham:
Make sure that `kwargs` is always an empty `dict` in `Worker.run*` methods.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2361,11 +2361,10 @@ class Worker(ServerNode):
         # logger.info("Finish job %d, %s", i, key)
         raise gen.Return(result)
 
-    def run(self, comm, function, args=(), wait=True, kwargs=None):
-        kwargs = kwargs or {}
+    def run(self, comm, function, args=(), wait=True, kwargs={}):
         return run(self, comm, function=function, args=args, kwargs=kwargs, wait=wait)
 
-    def run_coroutine(self, comm, function, args=(), kwargs=None, wait=True):
+    def run_coroutine(self, comm, function, args=(), kwargs={}, wait=True):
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @jakirkham .  

It looks like the run function does mutate kwargs, so we may have to be careful here.  

I suggest that we handle this in the top-level run function, perhaps by setting its kwargs to None as well, and then doing?

```python
kwargs = kwargs or {}
```

That should help to centralize things as well.

I'm curious, can you share how this came up?  Is there a test that we should add?

jakirkham:
> It looks like the run function does mutate kwargs, so we may have to be careful here.

Would that affect the default argument though? I'm not sure I follow though concern being raised.

> I suggest that we handle this in the top-level run function, perhaps by setting its kwargs to None as well, and then doing?

Sure this should be handled in the latest change.

> I'm curious, can you share how this came up?

Was reading the code and thought it odd that `run` and `run_coroutine` differed in this way. Guessing that was unintentional; so figured it deserved a fix.

> Is there a test that we should add?

Maybe? Could you please advise on where such a test would live and what it would look like?

mrocklin:
> Would that affect the default argument though? I'm not sure I follow though concern being raised.

Yes.  People often avoid using default mutable arguments for this reason.   

> Maybe? Could you please advise on where such a test would live and what it would look like?

I couldn't advise on this no.  I don't know when this would come up.  I had assumed that this came up for a reason, and I was asking about what that reason was.  If this hasn't come up though then no thoughts from me.

mrocklin:
+1

PR #3941, opened 2020-07-01
Make dashboard server listens on all IPs by default even when interface is set explicitly.

lesteve:
This is causing a HTTP 500 in the dashboard in a dask-jobqueue context (when interface is explicitly specified).

Hacky way to reproduce the problem:
```sh
dask-scheduler --interface lo&
sleep 2
ss -nlput | grep 8787
kill %
```

distributed 2.19 (useful par of the output)
```
tcp    LISTEN     0      128    127.0.0.1:8787                  *:*                   users:(("dask-sched
```

So my reading of this is that the dashboard is only listening on the IP address corresponding to the `lo` interface (i.e. the one specified for the scheduler).

distributed 2.14 (useful part of the output)
```
tcp    LISTEN     0      128       *:8787                  *:*                   users:(("dask-scheduler",pid=19693,fd=9))
tcp    LISTEN     0      128      :::8787                 :::*                   users:(("dask-scheduler",pid=19693,fd=10))
```

Very likely introduced by #3658.

When `dashboard_address` was `":8787"` `http_address` was `""` so the HTTP server was not listening on all IP interfaces.

Not sure if this is the right fix and how to test this. Feed-back more than welcome!

### DIFF
--- a/distributed/node.py
+++ b/distributed/node.py
@@ -121,7 +121,7 @@ class ServerNode(Server):
         self.http_server = HTTPServer(self.http_application, ssl_options=ssl_options)
         http_address = clean_dashboard_address(dashboard_address or default_port)
 
-        if not http_address["address"]:
+        if http_address["address"] is None:
             address = self._start_address
             if isinstance(address, (list, tuple)):
                 address = address[0]
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1205,8 +1205,8 @@ async def test_service_hosts():
     port = 0
     for url, expected in [
         ("tcp://0.0.0.0", ("::", "0.0.0.0")),
-        ("tcp://127.0.0.1", "127.0.0.1"),
-        ("tcp://127.0.0.1:38275", "127.0.0.1"),
+        ("tcp://127.0.0.1", "0.0.0.0"),
+        ("tcp://127.0.0.1:38275", "0.0.0.0"),
     ]:
         async with Scheduler(host=url) as s:
             sock = first(s.http_server._sockets.values())
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1205,8 +1205,8 @@ async def test_service_hosts():
     port = 0
     for url, expected in [
         ("tcp://0.0.0.0", ("::", "0.0.0.0")),
-        ("tcp://127.0.0.1", "0.0.0.0"),
-        ("tcp://127.0.0.1:38275", "0.0.0.0"),
+        ("tcp://127.0.0.1", ("::", "0.0.0.0")),
+        ("tcp://127.0.0.1:38275", ("::", "0.0.0.0")),
     ]:
         async with Scheduler(host=url) as s:
             sock = first(s.http_server._sockets.values())
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -1042,7 +1042,7 @@ async def test_service_hosts_match_worker(s):
 
     async with Worker(s.address, host="tcp://127.0.0.1") as w:
         sock = first(w.http_server._sockets.values())
-        assert sock.getsockname()[0] == "0.0.0.0"
+        assert sock.getsockname()[0] in ("::", "0.0.0.0")
 
 
 @gen_cluster(nthreads=[])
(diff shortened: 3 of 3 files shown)


mrocklin:
@quasiben if you have a moment this could use a second opinion, and I personally would value yours

mrocklin:
@lesteve thank you for tracking this down.  I don't want to know how many hours when into this one-line code change :)

If we wanted to test this then maybe we could do the following?

```python
@pytest.mark.asyncio
async def test_dashboard_open_address(cleanup):
    async with Scheduler(interface="lo") as s:
        client = await tornado.tcpclient.AsyncHTTPClient(...)  # I'm a little fuzzy on how this works
        await client.start()  # but grepping in the codebase should give better examples

        response = await client.get(f"http://127.0.0.2:{s.http_server.port}")
        assert response.status == 200
```

mrocklin:
@lesteve if a user wants to limit the access to the dashboard is that still possible with this change?

lesteve:
> I don't want to know how many hours when into this one-line code change :)

That was not that bad actually once I convinced myself that something must have changed in the treatment of `dashboard_address` (thanks to the `ss -nlput` commands) ...

> @lesteve if a user wants to limit the access to the dashboard is that still possible with this change?

Not an expert on this, but I guess you can always set the dashboard server IP address explicitly (e.g. with `--dashboard-address`) if you want to restrict dashboard access to only one specific IP address.

What is true though is that the default behaviour changed in distributed 2.15.
1. dask 2.14 `--interface` is used by the scheduler, the dashboard server listens on all interfaces by default
2. dask 2.15 `--interface` is used both by the scheduler and the dashboard server

Not sure whether 1. or 2. is preferable.

I always assumed that `:8787` (e.g. in `dask-scheduler --help`) meant all IP addresses on the port 8787 but I may be wrong.

In dask-jobqueue we probably want the behaviour of running the dashboard server on all IP addresses by default to avoid hard-to-debug dashboard problems.

lesteve:
> I always assumed that :8787 (e.g. in dask-scheduler --help) meant all IP addresses on the port 8787 but I may be wrong.

The [Python socket documentation](https://docs.python.org/3.8/library/socket.html#socket-families) seems to agree that empty string for an IP address means "bind to all interfaces":

> For IPv4 addresses, two special forms are accepted instead of a host address: `''` represents `INADDR_ANY`, which is used to bind to all interfaces.

quasiben:
As @mrocklin said, thanks for tracking this down @lesteve!  

I think with this change we also get the dashboard connecting to all IPs even when we start with ucx/IB:

> dask-scheduler --protocol ucx --interface ib1

Current:
```
(base) bzaitlen@dgx12:~$ ss -nlput | grep 8787
tcp  LISTEN  0       128                10.33.225.172:8787         0.0.0.0:*     users:(("dask-scheduler",pid=48327,fd=7))
```

With this PR:
```
(base) bzaitlen@dgx12:~$ ss -nlput | grep 8787
tcp  LISTEN  0       128                      0.0.0.0:8787         0.0.0.0:*     users:(("dask-scheduler",pid=48099,fd=7))
tcp  LISTEN  0       128                         [::]:8787            [::]:*     users:(("dask-scheduler",pid=48099,fd=8))
```

If this assessment is correct, this will make our Infiniband users happy.

There some test failures which probably need to be fixed -- though nothing to significant:

```
    async def test_service_hosts():
        port = 0
        for url, expected in [
            ("tcp://0.0.0.0", ("::", "0.0.0.0")),
            ("tcp://127.0.0.1", "127.0.0.1"),
            ("tcp://127.0.0.1:38275", "127.0.0.1"),
        ]:
            async with Scheduler(host=url) as s:
                sock = first(s.http_server._sockets.values())
                if isinstance(expected, tuple):
                    assert sock.getsockname()[0] in expected
                else:
>                   assert sock.getsockname()[0] == expected
E                   AssertionError: assert '0.0.0.0' == '127.0.0.1'
E                     - 127.0.0.1
E                     + 0.0.0.0
```

mrocklin:
This looks fine to me, although I'm not particularly knowledgable here.  @quasiben is this something that I could leave in your hands?

quasiben:
I think we're good here.  Merging in.  Thanks for the ping @mrocklin

lesteve:
Great, I forgot about that but is there some need for a changelog somewhere?

mrocklin:
No, whoever does the release handles that at release time.  We generally
copy over the commit title, which is taken from the the title of the PR.

PR #3944, opened 2020-07-06
Avoid too aggressive retry of connections.

Carreau:
Originally retrying was implemented to avoid timeout expiring for
addresses not yet existing (#3084). Though giving up after 1 second
maximum can be annoying for slow system and can trigger a thundering
herd-like problem.

Here progressively increase the inner timeout to give a chance to a slow
systems on th other side to respond especially when they are slow.

Here also jitter a bit when we are going to retry, and for how long
before we give up in the inner timeout to try-to again mitigate teh
thundering herd.

This is relatively difficult to test; so no tests are included.

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -219,7 +219,7 @@ async def connect(addr, timeout=None, deserialize=True, **connection_args):
     if timeout and timeout / 20 < backoff:
         backoff = timeout / 20
 
-    retry_timeout_backoff = 1
+    retry_timeout_backoff = random.randrange(140, 160) / 100
 
     # This starts a thread
     while True:
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -3,6 +3,7 @@ import asyncio
 from contextlib import suppress
 import inspect
 import logging
+import random
 import weakref
 
 import dask
@@ -218,6 +219,8 @@ async def connect(addr, timeout=None, deserialize=True, **connection_args):
     if timeout and timeout / 20 < backoff:
         backoff = timeout / 20
 
+    retry_timeout_backoff = 1
+
     # This starts a thread
     while True:
         try:
@@ -227,7 +230,7 @@ async def connect(addr, timeout=None, deserialize=True, **connection_args):
                 )
                 with suppress(TimeoutError):
                     comm = await asyncio.wait_for(
-                        future, timeout=min(deadline - time(), 1)
+                        future, timeout=min(deadline - time(), retry_timeout_backoff)
                     )
                     break
             if not comm:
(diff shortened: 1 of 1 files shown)


quasiben:
Retry with backoff makes sense to me.  I'm not sure jitter-ing adds something which the backoff is already doing ? Did jitter-ing particularly help a use case for you ?

mrocklin:
cc @jacobtomlinson who I think has touched this before

Carreau:
> Did jitter-ing particularly help a use case for you ?

I'll see if I can experiment.

Carreau:
The other possibility would to maybe figure out a way to stagger a large number of connections, 
but then I would likely need guidance.

jacobtomlinson:
I'm generally in favour of jittering when retrying a connection, although in many situations I can imagine the whatever is scheduling providing enough jitter when starting work.

The problem I was tackling when touching this code was kubernetes creating the compute resource concurrently with the network resource which resulted in a race condition. That is a nuance with the scheduler on a single connection (this could happen on many connections, but each connection with experience it's own isolated race condition) so I can't imagine it resulting in a thundering herd situation.

It may help to know a little more background about the problem being experienced that resulted in this PR.

Carreau:
I can't give all of the informations as I don't have them all but the use case is the following:

>  there are times when I have to initiate a connection to each of thousands of workers (e.g. Client.run()). It takes more than 1 second to handle thousands of connections/handshakes, so a large fraction of attempts are timed out and retried.

> random initial retry_timeout_backoff helps [make initial retry_timeout_backoff = random....], as does jitter.

> I still see lots of retries and performance doesn't seem close to what is possible if there are no retries. But this is already much better than before.

The problem with retrying every one second (of any fixed amount) is for a system that can't deal with all the connections you give it you'll basically abort all of them at the same time ; sleep for an amount of time where connection could be processed but are not, and then all connections are retired at the same time.

There is intrinsec jittering of the scheduler, but likely not enough.

With this patch:

> I think if you have a burst of connect attempts at t=0, then you'll get a lot of failures at 1.4-1.6, which are then rescheduled for 1.96-2.56 seconds later, i.e. over a span of 0.2+0.6=0.8 seconds, which is still pretty close. Of course, it's close in my use-case, but for the rest of the world, maybe not.

I pushed a patch that also make the _initial_ retry_timeout_backoff random; 

I'm of course happy to change the values of the randomness, use gaussian instead of uniform, or whatever you see fit. Another alternative is to make this `retry_timeout_backoff` configurable, though if we can find something that does not needs configuration, that would be great.

jacobtomlinson:
In general I'm in favour of this patch.

quasiben:
@jacobtomlinson feel free to merge if you are in favor :)

jacobtomlinson:
Ok let's go for this. @Carreau if you want to extend this further let's do it in another PR.

Carreau:
> Ok let's go for this. @Carreau if you want to extend this further let's do it in another PR.

Thanks for the quick turnaround and I'll try to think of a better solution or improve even more.

PR #3945, opened 2020-07-06
Move scheduler_comm into Cluster.__init__

mrocklin:
Following on from https://github.com/dask/distributed/pull/3935
This makes sure that not-yet-started clusters close quietly, even if
they are not SpecClusters

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -55,6 +55,7 @@ class Cluster:
         self._asynchronous = asynchronous
         self._watch_worker_status_comm = None
         self._watch_worker_status_task = None
+        self.scheduler_comm = None
 
         self.status = "created"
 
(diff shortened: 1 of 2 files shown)


quasiben:
Looks good. 

Do you think we need to update the docstring in the cluster object as `scheduler_comm` ?

https://github.com/dask/distributed/blob/c76fc83af548a633bc5602738937a1e7e0933b17/distributed/deploy/cluster.py#L34-L35

I suppose that is still necessary for objects which implement `Cluster`

mrocklin:
Yes, the nature of the scheduler_comm attribute hasn't changed.  We're just ensuring that the attribute exists in case we close a cluster before starting it.

quasiben:
Thanks!  Merging in.

PR #3949, opened 2020-07-07
Synchronize on non-trivial CUDA frame transmission

jakirkham:
Further restrict when CUDA synchronization occurs by checking when non-trivial (non-zero length) CUDA frames are being transmitted and only synchronize in that case. Should avoid synchronization when only meta objects or objects with primarily host frames are involved.

cc @pentschev @quasiben


jakirkham:
Have tested with the UCX tests in Distributed. Also have run with UCX-Py's and Dask-CUDA's test suites. Seems to work with all of them based on my testing.

quasiben:
Thanks John

PR #3960, opened 2020-07-14
Skip coercing to `bytes` in `merge_frames`

jakirkham:
As the frames we receive are typically mutable, non-`bytes` objects like `bytearray`s or NumPy `ndarray`s, coercing to `bytes` at this stage triggers a copy of all frames. As we are going to toss those copied versions anyways when joining them into a larger `bytes` object, this ends up being wasteful with memory. Fortunately `bytes.join(...)` accepts any and all `bytes`-like objects. So instead just pass them all through as-is to `bytes.join(...)`, which is free and doesn't require a copy. Should cutdown on the memory usage in this part of the code.

### DIFF
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -1,7 +1,7 @@
 import struct
 import msgpack
 
-from ..utils import ensure_bytes, nbytes
+from ..utils import nbytes
 
 BIG_BYTES_SHARD_SIZE = 2 ** 26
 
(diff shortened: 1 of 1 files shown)


jakirkham:
FWIW here's a little example showing this works.

```python
In [1]: import numpy as np                                                      

In [2]: b"".join([ 
   ...:     b"abc", 
   ...:     bytearray(b"def"), 
   ...:     memoryview(b"ghi"), 
   ...:     np.arange(106, 109, dtype="u1") 
   ...: ])                                                                      
Out[2]: b'abcdefghijkl'
```

quasiben:
I kicked the CI here so test should be passing now

PR #3963, opened 2020-07-14
Fix replication error

andrewfulton9:
fixes error where a retired worker fails to replicate remaining tasks to other workers due to garbage collection. Fixes issue by removing tasks from the tasks to replicate list that have a `forgotten` state

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -63,6 +63,7 @@ conda create -n dask-distributed -c conda-forge -c defaults \
     toolz \
     tornado=$TORNADO \
     zstandard \
+    pip \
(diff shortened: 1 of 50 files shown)


quasiben:
@jacobtomlinson would you have time to take a look at this ?

andrewfulton9:
There isn't a related issue that I know of. I could make one if need be.

jacobtomlinson:
@andrewfulton9 don't worry about it this time. It just makes reviewing easier when there is a good description of the bug somewhere.

I've tested this locally and can confirm the test fails without the change to `scheduler.py` and passes with the change. The test also looks reasonable.

Thanks for taking the time to raise this PR!

PR #3964, opened 2020-07-15
Treat falsey local directory as None

TomAugspurger:
Closes https://github.com/pangeo-data/pangeo-cloud-federation/issues/655

Dask-Gateway uses local-directory="" to launch workers. I think the
intent is that it's interpreted as False, and should fall back to the
config / current directory.

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -867,8 +867,11 @@ def test_worker_dir(worker):
 
 @gen_cluster(nthreads=[])
 async def test_false_worker_dir(s):
-    async with Worker(s.address, loop=s.loop, local_directory=""):
-        pass
+    async with Worker(s.address, local_directory="") as w:
+        local_directory = w.local_directory
+
+    cwd = os.getcwd()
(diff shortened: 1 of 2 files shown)


distributed/tests/test_worker.py
```diff
@@ -865,6 +865,12 @@ async def test_worker_dir(c, s, a, b):
         test_worker_dir()
 
 
+@gen_cluster(nthreads=[])
+async def test_false_worker_dir(s):
+    async with Worker(s.address, loop=s.loop, local_directory=""):
```
mrocklin, line 870:
```suggestion
    async with Worker(s.address, local_directory=""):
```

Specifying `loop=` isn't necessary if you're in the same event loop for a Worker.  Really, we should consider removing the keyword.

mrocklin:
This seems fine to me.  In the meantime, would it make sense to change dask-gateway as well?

TomAugspurger:
Doing that in https://github.com/dask/dask-gateway/pull/283

TomAugspurger:
Maybe a flaky test `test_service_hosts`. I restarted that worker.

Issue #3943, opened 2020-07-05
Exception: ValueError('buffer source array is read-only')


astrojuanlu:
(Comes from https://github.com/dask/distributed/issues/1978#issuecomment-645869748)

**What happened**:

```
$ ipython
Python 3.8.3 (default, May 20 2020, 12:50:54) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.15.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: # coding: utf-8 
   ...: from dask import dataframe as dd 
   ...: import pandas as pd 
   ...: from distributed import Client 
   ...: client = Client() 
   ...: df = dd.read_csv("../data/yellow_tripdata_2019-*.csv", parse_dates=["tpep_pickup_datetime", "tpep_dropoff_datetime"]) 
   ...: payment_types = { 
   ...:     1: "Credit Card", 
   ...:     2: "Cash", 
   ...:     3: "No Charge", 
   ...:     4: "Dispute", 
   ...:     5: "Unknown", 
   ...:     6: "Voided trip" 
   ...: } 
   ...: payment_names = pd.Series( 
   ...:     payment_types, name="payment_name" 
   ...: ).to_frame() 
   ...: df2 = df.merge( 
   ...:     payment_names, left_on="payment_type", right_index=True 
   ...: ) 
   ...: op = df2.groupby("payment_name")["tip_amount"].mean() 
   ...: client.compute(op) 
   ...:                                                                                                                                                                                                                                       
Out[1]: <Future: pending, key: finalize-85edcc1f23785545f628c932abd19768>

In [2]: distributed.worker - WARNING -  Compute Failed                                                                                                                                                                                        
Function:  _apply_chunk
args:      (        VendorID tpep_pickup_datetime tpep_dropoff_datetime  passenger_count  trip_distance  RatecodeID store_and_fwd_flag  ...  mta_tax  tip_amount  tolls_amount  improvement_surcharge  total_amount  congestion_surcharge  payment_name
0              1  2019-01-04 14:08:46   2019-01-04 14:18:10                1           1.70           1                  N  ...      0.5         0.0          0.00                    0.3          9.30                   NaN          Cash
1              1  2019-01-04 14:20:33   2019-01-04 14:25:10                1           0.90           1                  N  ...      0.5         0.0          0.00                    0.3          6.30                   NaN          Cash
[...]

PR #3918, opened 2020-06-22
WIP: Use bytearray in serialization

jakirkham:
Requires PR ( https://github.com/dask/distributed/pull/3960 )

Make sure we use `bytearray`s for frames in `merge_frames`. This is needed to ensure we have frames that are writable as opposed to `bytes`, which is readonly.

xref: https://github.com/dask/distributed/issues/1978#issuecomment-646724477

Issue #1978, opened 2018-05-12
Tackle "ValueError: buffer source array is read-only"


crusaderky:
One of the biggest pitfalls of running dask.array with a distributed scheduler is the dreaded ``ValueError: buffer source array is read-only``.
This error is typical when one runs a memoryview-based Cython kernel in distributed, and it's particularly insidious as it will never show up in unit tests performed with the local dask multithreaded scheduler. It might even not show up when you run your problem on distributed and the arrays just happen to never transit across nodes or to the disk cache (which is exactly what the scheduler will try to achieve if enough RAM and CPU power are available).

In other words, this is a textbook example of an issue that risks appearing in production for the first time!

I can see a few ways of tackling the problem:
- in Cython: if you ain't writing to an array, it should just work. As this issue has been around for the longest time, I suspect it might not be trivial?
- in distributed, making sure that all arrays passed to all kernels are writeable
- in dask.array, making sure that all arrays passed to all kernels are NOT writeable, which actually makes a lot of sense regardless of distributed. This will make the error crop up immediately in any naive unit tests. It will also wreak havoc for many existing dask.array users though. Possibly a opt-in setting?
- in the distributed docs, with a thorough tutorial on how to reproduce the problem in unit testing and how to change your kernels to fix it, so that it becomes the first result when anybody googles the exception.

On the last point, I personally solved the problem as follows:

In the kernels:
```
def _memoryview_safe(x):
    """Make array safe to run in a Cython memoryview-based kernel. These
    kernels typically break down with the error ``ValueError: buffer source
    array is read-only`` when running in dask distributed.
    """
    if not x.flags.writeable:
        if not x.flags.owndata:
            x = x.copy(order='C')
        x.setflags(write=True)
    return x


def splev(x_new, t, c, k=3, extrapolate=True):
    x_new = _memoryview_safe(x_new)
    t = _memoryview_safe(t)
    c = _memoryview_safe(c)
    spline = scipy.interpolate.BSpline.construct_fast(t, c, k, axis=0, extrapolate=extrapolate)
    return spline(x_new)
```

In the unit test:
```
def test_distributed():
    def ro_array(a):
        a = np.array(a)
        a.setflags(write=False)
        # Return a view of a, so that setting the write flag on the view is not enough
        return a[:]

    t = ro_array([1, 2])
    c = ro_array([10, 20])
    x_new = ro_array([1.5, 1.8])
    splev(x_new, t, c, k=1)
```
[...]

PR #3967, opened 2020-07-17
Ensure writable frames

jakirkham:
Fixes https://github.com/dask/distributed/issues/1978
Fixes https://github.com/dask/distributed/issues/3943

Closes https://github.com/dask/distributed/pull/3918 (superseded)

User code often expects NumPy and Pandas objects to be writable. However this fails in various cases. After some investigation this appears to be due to some frames being read-only (as they coerced to `bytes` via compression, logic in `merge_frames`, etc.). To fix this we update the logic in `merge_frames` to check frames are writable. If frames are read-only, we copy them to a writable object (like `bytearray`). We include tests of `merge_frames` and NumPy array serialization to make sure this is working correctly.


mrocklin:
Thanks for handling this @jakirkham .  I'm only seeing tests in this PR so far.  Is that correct?

jakirkham:
Yeah I'm just trying to demonstrate the failure on CI. Will push the rest of the fix after.

jakirkham:
Alright have pushed the changes to fix the tests. Please let me know what you think 🙂

distributed/protocol/utils.py
```diff
@@ -58,34 +58,36 @@ def merge_frames(header, frames):
...
-                l = 0
-        if len(L) == 1:  # no work necessary
-            out.extend(L)
-        else:
-            out.append(b"".join(L))
-    return out
+                out.append(memoryview(bytearray().join(L)))
+        frames = out
+
+    frames = [memoryview(bytearray(f)) if f.readonly else f for f in frames]
```
mrocklin, line 88: For fun I decided to micro-benchmark this: 

```python
In [1]: L = [b'0' * i for i in [10, 100, 10, 100, 1000] * 2]                                                                                                                                                      

In [2]: %timeit b"".join(L)                                                                                                                                                                                       
203 ns ± 1.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit memoryview(bytearray().join(L))                                                                                                                                                                   
463 ns ± 5.9 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

260ns doesn't really matter today I don't think

mrocklin:
In principle this looks good to me.  Thank you for identifying and resolving this @jakirkham !

jakirkham:
Thanks Matt! 😄

jakirkham:
Adding a test for a Pandas Series as well in PR ( https://github.com/dask/distributed/pull/3995 ).

PR #3971, opened 2020-07-20
Use `merge_frames` with host memory only

jakirkham:
Currently `merge_frames` is only really designed to work with memory on host. So limit calling `merge_frames` to cases where host memory is used (as opposed to cases with device memory).

### DIFF
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -126,7 +126,8 @@ def loads(frames, deserialize=True, deserializers=None):
             if deserialize or key in bytestrings:
                 if "compression" in head:
                     fs = decompress(head, fs)
-                fs = merge_frames(head, fs)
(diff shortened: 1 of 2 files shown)


quasiben:
Thanks @jakirkham tested locally and works for me

quasiben:
This is failing with errors like:

```
    if head["serializer"] in ("dask", "pickle"):
KeyError: 'serializer'
```

In [test_TaskGraph](https://github.com/dask/distributed/pull/3971/checks?check_run_id=891383559#step:8:115)

jakirkham:
Changed the check to simply see that no frames are on device before calling `merge_frames`. Should fix it. Also retested with [this reproducer]( https://github.com/rapidsai/tpcx-bb/issues/51#issuecomment-661256713 ).

quasiben:
Thanks @jakirkham

Issue #3969, opened 2020-07-20
Concurrent os.makedirs for dask-worker-space can lead to worker failure


jendrikjoe:
**What happened**:

This bug occurs randomly as it is one that is raised due to multiple processes creating the `dask-worker-space` folder if it does not exist. I only stumble over it every ~100 scripts I launch. 
If you run:
```
from dask.distributed import Client
client = Client(memory_limit=0, n_workers=Environment.DASK_PROCESSES)
```
a lot of times with a high enough number for `Environment.DASK_PROCESSES` (16+), the following error will occur.

```
Process Dask Worker process (from Nanny):
Traceback (most recent call last):
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/site-packages/distributed/process.py", line 191, in _run
    target(*args, **kwargs)
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/site-packages/distributed/nanny.py", line 728, in _run
    worker = Worker(**worker_kwargs)
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/site-packages/distributed/worker.py", line 489, in __init__
    os.makedirs(local_directory)
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/dask-worker-space'
distributed.utils - ERROR - addresses should be strings or tuples, got None
Traceback (most recent call last):
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/site-packages/distributed/utils.py", line 656, in log_errors
    yield
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/site-packages/distributed/scheduler.py", line 2208, in remove_worker
    address = self.coerce_address(address)
  File "/home/user/.pyenv/versions/3.8.2/lib/python3.8/site-packages/distributed/scheduler.py", line 4946, in coerce_address
    raise TypeError("addresses should be strings or tuples, got %r" % (addr,))
TypeError: addresses should be strings or tuples, got None
distributed.core - ERROR - addresses should be strings or tuples, got None
[...]

PR #3972, opened 2020-07-20
Ignore existing local directory when booting worker

jendrikjoe:
This PR addresses #3969. 
It resolves an issue where concurrently booting workers could try to create a folder at the same time with one of the two failing because the folder already exists. It resolves that issue, by simply trying to create the local_directory `dask-worker-space` on each workers boot time. If the folder already exists, it simply proceeds without raising an error. This way folder creation is ensured while avoiding concurrency issues. I as well removed the `if` statement, as that is essentially what `os.makedirs` does internally, so it seemed redundant to me. I can add it back in though 😉 

I did not add a test to reproduce the error either. It is really hard to reproduce the state of two processes trying to create the same folder at the same time and I really just found the bug after launching a lot of scripts running dask. I could take a shot at it, but it would probably end up being a for loop with a really large number of loops and non-deterministic nonetheless 😉

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -485,8 +485,7 @@ class Worker(ServerNode):
         if not local_directory:
             local_directory = dask.config.get("temporary-directory") or os.getcwd()
 
-        if not os.path.exists(local_directory):
-            os.makedirs(local_directory)
+        os.makedirs(local_directory, exist_ok=True)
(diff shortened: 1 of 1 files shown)


jakirkham:
Looks great. Thanks Jendrik! 😄

Agree the `if` seems fine to drop. Wouldn't worry about an explicit test. 😉

PR #3979, opened 2020-07-21
Fixes #3978, enabling ActorFuture to work with `as_completed()`

### LINKED ISSUE #3978: ActorFuture doesn't work with `as_completed(...)`
**What happened**:
Running:
```python
    class Counter:
        n = 0

        def __init__(self):
            self.n = 0

        def increment(self):
            self.n += 1
            return self.n

        def add(self, x):
            self.n += x
            return self.n

    future = client.submit(Counter, actor=True)
    counter = future.result()

    futures = []
    for num in [0, 1, 2]:
        futures.append(counter.add(num))

    for future in as_completed(futures):
        print(future.result())
```

yielded the error:

```python
TypeError: Input must be a future, got <ActorFuture>
```

It to not raise and error (and print 0, 1, 3)

I was able to fix my problem by commenting out [these 2 liines](https://github.com/dask/distributed/blob/6eab3ce464e9ff79e4192de4f1c778d3ef89379b/distributed/client.py#L4382-L4383). 

I have submitted a pull request to fix this.

**Environment**:

- Dask version: master
- Python version: v3.8.4
- Operating System: Ubuntu

asmith26:
* Add test.
* Move import in actor.py (to prevent circular import - wasn't sure if this is the best way to achieve this, all feedback welcome :)).

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -40,6 +40,7 @@ except ImportError:
 from tornado import gen
 from tornado.ioloop import IOLoop, PeriodicCallback
 
+from.actor import ActorFuture
 from .batched import BatchedSend
 from .utils_comm import (
     WrappedKey,
@@ -4379,8 +4380,8 @@ class as_completed:
         The added futures will emit from the iterator once they finish"""
         with self.lock:
             for f in futures:
-                if not isinstance(f, Future):
-                    raise TypeError("Input must be a future, got %s" % f)
+                if not isinstance(f, Future) and not isinstance(f, ActorFuture):
+                    raise TypeError("Input must be a Future or ActorFuture, got %s" % f)
                 self.futures[f] += 1
                 self.loop.add_callback(self._track_future, f)
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -40,7 +40,7 @@ except ImportError:
 from tornado import gen
 from tornado.ioloop import IOLoop, PeriodicCallback
 
-from.actor import ActorFuture
+from .actor import ActorFuture
 from .batched import BatchedSend
 from .utils_comm import (
     WrappedKey,
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -63,6 +63,7 @@ class Actor(WrappedKey):
                 self._worker = None
             try:
                 from .client import Future, default_client
+
                 self._client = default_client()
                 self._future = Future(key)
             except ValueError:
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -3,7 +3,6 @@ import functools
 import threading
 from queue import Queue
 
-from .client import Future, default_client
 from .protocol import to_serialize
 from .utils import sync
 from .utils_comm import WrappedKey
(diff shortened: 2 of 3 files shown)


distributed/client.py
```diff
@@ -4379,8 +4380,8 @@ def update(self, futures):
         The added futures will emit from the iterator once they finish"""
         with self.lock:
             for f in futures:
-                if not isinstance(f, Future):
-                    raise TypeError("Input must be a future, got %s" % f)
+                if not isinstance(f, Future) and not isinstance(f, ActorFuture):
```
mrocklin, line 4383:
```suggestion
                if not isinstance(f, (Future, ActorFuture):
```

mrocklin:
Thanks for working on this @asmith26 .  I suspect that this may not work out of the box, and that you might have to modify the `as_completed._track_future` method a bit, which today probably expects a normal Dask Future.  Some of the code in that method probably won't work on ActorFutures.  

However, the good news here is that making a similar method for ActorFutures probably isn't that hard because ActorFutures are a bit less complex.  You can probably `await future` rather than calling `_wait` and `future._result`.  `ActorFutures` always pull results back, so we don't need to be too careful here.  Most of the rest of the code should be similar.

distributed/client.py
mrocklin: Actually we should probably keep this as two if statements, and maybe include a second `_track_actorfuture` method

asmith26:
Hi @mrocklin, Thanks for your help :)

Changing my test seems to have fixed the failing checks (I realised that my old test would not always return `4` due to the asynchronous nature of it). Consequently I think we may be done here (I can still have a go at creating a `_track_actorfuture` if useful, I thought this might be unnecessary/redundant code now though?)

mrocklin:
Can I ask you to try another test with actor futures that take a little bit of time?  Maybe put a `time.sleep(0.1)` in there?  Looking at the track_futures implementation I don't see how it would work on ActorFuture objects.

asmith26:
I tried using `with_results=True` in my test which seemed to cause my test to hang.

mrocklin:
The code in `_track_future` expects a `Future`, not an `ActorFuture`.  You'll have to make a separate, hopefully simpler method to handle `ActorFutures` correctly.  I recommend reading through the `_track_future` method to understand what it does, and then using `await future` instead of `_wait` and `_result`.

distributed/client.py
asmith26: I think this is now done (please see ff62b0b).

asmith26:
> The code in `_track_future` expects a `Future`, not an `ActorFuture`. You'll have to make a separate, hopefully simpler method to handle `ActorFutures` correctly. I recommend reading through the `_track_future` method to understand what it does, and then using `await future` instead of `_wait` and `_result`.

Thanks for you help @mrocklin 

I think I understand the `_track_future` method now. I've currently included lots of comments with my current understanding of how (I think) this works (please feel free to correct me, best way to learn is from my mistakes :) ).

Trying your suggestion regarding [await future](https://github.com/dask/distributed/pull/3979/commits/ff62b0b5722939a50937097caf377d96cb2f1e7a#diff-96a27223dc91b5c9ea3d03684d79ad3fR4379) seems to hang (I have no idea why, I've probably implemented it incorrectly). I haven't really used `asyncio` before (so any recommended reading or resources greatly appreciated :) ).

Issue #3739, opened 2020-04-24
client.submit not working with client.upload_file after first time


huseinzol05:
Below are my local and cloud Dask version.

Dask version: 2.14
Distributed version: 2.14

We have a production Dask cluster in GKE, and shared among our engineers. Now, we want to create a python script for boilerplate purpose, simply name, `boilerplate.py`,

```python
def inc(x):
    if x > 10:
        raise Exception('x > 10')
    return x + 1
```

And our client script,
```python
from dask.distributed import Client
import boilerplate as bp
bootstrap = # proxied
client = Client(bootstrap)
client.upload_file('boilerplate.py')

client.submit(bp.inc,7).result() # no error
client.submit(bp.inc,11).result() # got error
```

Now we fixed `boilerplate.py`,

```python
def inc(x):
    if x > 20:
        raise Exception('x > 20')
    return x + 1
```

And rerun client script to reupload fixed `boilerplate.py`,
```python
from dask.distributed import Client
Exception: x > 10
bootstrap = # proxied
client = Client(bootstrap)
client.upload_file('boilerplate.py')

client.submit(bp.inc,7).result() # no error
client.submit(bp.inc,11).result() # got error
```

It is still same exception,
```
Exception: x > 10
````

The problem here, we cannot restart our Dask cluster, it might disturb other important tasks. It might seems got some caching layer here, this is only happened if we use `file_upload` + `submit`, if we use `file_upload` + `run`, it use latest update `boilerplate.py`.

**Comment by TomAugspurger** [comment]
What happens if you include a `client.run(importlib.reload(boilerplate)` to reload the module? Reloading python code at runtime is always fraught (what do you do with existing classes?) but may work here.

**Comment by TomAugspurger** [comment]
You might also need to manually clear some data off the workers. I don' know if the function gets cached in the worker's `.data` attribute.

**Comment by huseinzol05** [comment]
Tried,

```python
import boilerplate as bp
def reload_package(package):
    import sys
    import importlib
    importlib.reload(sys.modules[package])
    return list(sys.modules.keys())

r = client.run(reload_package, 'boilerplate')
import boilerplate

x = client.submit(boilerplate.inc, 21)
x.result() # error
```

Some of workers reloaded, some are not, weird.

TomAugspurger:
Let us know if you're able to do more debugging.

Given the complexities here, I don't really see how Dask can support re-uploading modules with `upload_file`. I'd recommend investigating alternative ways of doing this.

[...]

PR #3981, opened 2020-07-22
disable worker caching

huseinzol05:
https://github.com/dask/distributed/issues/3739


jsignell:
Thanks for opening a pull request @huseinzol05! It looks like you have some automatic syntax formatting set up that makes it look like there are many more changes to the code than I think you are suggesting. Can you turn that off and make sure that the changes you are suggesting only target the specific issue you are trying to address?

Issue #3975, opened 2020-07-21
Serialization Error Messages Hidden


jfenc91:
I am running a map_partitions that returns a dataframe with rows that have a large memory footprint. I dug out the following the exception below from the error log.

At first glance it doesn't really make sense because doing a simple test shows that dask can serialize dataframes:
`
>>> from distributed.protocol.serialize import serialize as s
>>> s(pd.DataFrame())
`

However, taking a closer look, it appears that there is a generic Exception catch here: 
https://github.com/dask/distributed/blob/2.21.0/distributed/protocol/serialize.py#L230

and if the on_error is set to raise like in the case of of moving partitions to disk during a map_partitions computation, we never get to see the message that is in the traceback format exception as shown here:

https://github.com/dask/distributed/blob/2.21.0/distributed/protocol/serialize.py#L234-L244

I suspect I am actually getting some sort of memory error, although I really have no way of knowing because the exception is reported as a TypeError

`
2020-07-21T17:12:03.599113035Z Traceback (most recent call last):
2020-07-21T17:12:03.599138316Z   File "/usr/local/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
2020-07-21T17:12:03.599143053Z     ret = callback()
2020-07-21T17:12:03.599146719Z   File "/usr/local/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
2020-07-21T17:12:03.599150027Z     future.result()
2020-07-21T17:12:03.599153015Z   File "/usr/local/lib/python3.6/site-packages/distributed/worker.py", line 2688, in memory_monitor
2020-07-21T17:12:03.599161122Z     k, v, weight = self.data.fast.evict()
2020-07-21T17:12:03.599164768Z   File "/usr/local/lib/python3.6/site-packages/zict/lru.py", line 89, in evict
2020-07-21T17:12:03.599192248Z     cb(k, v)
2020-07-21T17:12:03.599196364Z   File "/usr/local/lib/python3.6/site-packages/zict/buffer.py", line 60, in fast_to_slow
2020-07-21T17:12:03.599200149Z     self.slow[key] = value
2020-07-21T17:12:03.599203704Z   File "/usr/local/lib/python3.6/site-packages/zict/func.py", line 41, in __setitem__
2020-07-21T17:12:03.599207077Z     self.d[key] = self.dump(value)
[...]

PR #3982, opened 2020-07-23
fix: don't hide serialization exceptions

brycedrennan:
addresses https://github.com/dask/distributed/issues/3975


mrocklin:
Hi @brycedrennan !  

Sorry for the delay in response.  Could I ask you to include a small test case?  Your example from the issue would work, although ideally we could reproduce it from the standard serializers.  Maybe create an object that fails to serialize with pickle using `getstate/setstate` and verify that we see the right exception?

mrocklin:
Hi @brycedrennan !  

Any thoughts on how best to test this?  I'm not sure if you're still around or not on this any longer.

Also cc @jakirkham who has been digging around in serialization code recently.

PR #3983, opened 2020-07-24
disable worker caching

huseinzol05:
https://github.com/dask/distributed/issues/3739


huseinzol05:
@jsignell , created a new PR.

distributed/worker.py
```diff
@@ -3259,13 +3259,12 @@ async def _get_data():
 
 job_counter = [0]
 
-
 cache_loads = LRU(maxsize=100)
-
+enable_caching = dask.config.get("distributed.worker.cache", False)
```
jsignell, line 3263: I can't find where this is defined. I would expect it to be in this file: https://github.com/dask/distributed/blob/master/distributed/distributed-schema.yaml#L228 and in here https://github.com/dask/distributed/blob/master/distributed/distributed.yaml#L58. Plus is should be `True` by default to preserve existing behavior.

jsignell:
Can you add a test please?

distributed/worker.py
mrocklin: If we want to add this then it will have to have a more precise name.  Worker caching may refer to many different things other than this.

distributed/worker.py
```diff
@@ -3259,13 +3259,12 @@ async def _get_data():
...
 
 def loads_function(bytes_object):
     """ Load a function from bytes, cache bytes """
-    if len(bytes_object) < 100000:
+    if len(bytes_object) < 100000 and enable_caching:
```
mrocklin, line 3267: Perhaps a more generic way of doing this would be to make the `100000` number configurable?  That way you could specify a value of `0` to turn off function caching, and others could use this same value to tune the size in bytes that are allowed.  If we do this I encourage you to set the config value to a string and use the `dask.utils.parse_bytes` function

huseinzol05: I prefer `dask.utils.parse_bytes`, so we can set to 0 to disable it.

jxeagle:
@huseinzol05 @mrocklin @jsignell I independently saw the same behavior in our dask cluster that primarily runs delayed jobs, which rely on the module files that are updated to workers. 

fyi: I came up with this approach that is slightly different in that it clears the LRU cache when and only when a new version of the module is uploaded. see https://github.com/dask/distributed/pull/3993 the goal is to only invalidate the cache only when it is necessary, for performance considerations.

jrbourbeau:
@huseinzol05 are you blocked on anything here? No rush, just wanted to check in

huseinzol05:
I need to test https://github.com/dask/distributed/pull/3993 first, if that branch already solved our problem, I will close it, I will try to do it ASAP.

jrbourbeau:
@huseinzol05 just wanted to check in again to see if https://github.com/dask/distributed/pull/3993 solves the issue you were running into

jrbourbeau:
Closing for now due to inactivity. @huseinzol05 feel free to re-open if you'd like to pick this up again

PR #3984, opened 2020-07-24
Handle sum of memory percentage when memory_limit is 0

jsignell:
Closes https://github.com/dask/dask/issues/6452

To test do: 

```python

cluster = LocalCluster(n_workers=2, memory_limit=0)
client = Client(cluster)

client
```

and navigate to http://127.0.0.1:8787/workers

### DIFF
--- a/distributed/dashboard/tests/test_scheduler_bokeh.py
+++ b/distributed/dashboard/tests/test_scheduler_bokeh.py
@@ -431,6 +431,19 @@ async def test_WorkerTable_custom_metric_overlap_with_core_metric(c, s, a, b):
 
 
 @gen_cluster(client=True)
+async def test_WorkerTable_with_memory_limit_as_0(c, s, a, b):
+    s.workers[a.address].memory_limit = 0
+    s.workers[b.address].memory_limit = 0
+
+    wt = WorkerTable(s)
+    wt.update()
+    assert all(wt.source.data.values())
+    assert wt.source.data["name"][0] == "Total (2)"
+    assert wt.source.data["memory_limit"][0] == 0
+    assert wt.source.data["memory_percent"][0] == ""
+
+
+@gen_cluster(client=True)
 async def test_TaskGraph(c, s, a, b):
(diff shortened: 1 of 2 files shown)


quasiben:
Tested this out and it worked nicely.  Thanks @jsignell!
Screen Shot 2020-07-24 at 9 56 32 AM

mrocklin:
Thanks @jsignell !

Do we want to add a small test?  There are a variety of WorkerTable tests in `distributed/dashboard/tests/test_scheduler_bokeh.py`

jsignell:
Ah! Yes I always forget that these have tests :grimacing: I'll add one now.

lesteve:
This does seems like a edge case I missed in #3897, thanks for the fix @jsignell!

distributed/dashboard/tests/test_scheduler_bokeh.py
```diff
@@ -430,6 +430,19 @@ def metric(worker):
     assert s.workers[a.address].metrics["metric"] == -999
 
 
+@gen_cluster(client=True)
+async def test_WorkerTable_with_memory_limit_as_0(c, s, a, b):
+    s.workers[a.address].memory_limit = 0
+    s.workers[b.address].memory_limit = 0
```
mrocklin, line 436: I think that something like this should work instead, and probably relies only on public API rather than reaching in

```suggestion
@gen_cluster(client=True, worker_kwargs={"memory_limit": 0})
async def test_WorkerTable_with_memory_limit_as_0(c, s, a, b):
```

mrocklin: +1 from me otherwise

jsignell: Oh nice! I didn't look closely at what was going on in the test setup :facepalm:

mrocklin:
Thanks @jsignell !

rcthomas:
Thanks @jsignell !

PR #3986, opened 2020-07-24
What to do when threads_per_worker is 0

jsignell:
*tl;dr Maybe `threads_per_worker` should throw an error if set to 0.*

While working on https://github.com/dask/distributed/pull/3984  I ran into some odd behaviors when `threads_per_worker` is set to 0 in `LocalCluster`. I am not sure what the intended behavior should be or if this PR makes things better. 

### Some examples:
On master the following results in a divide by 0 error. This PR sets n_workers to 0 which I think is a good solution:

```python
cluster = LocalCluster(threads_per_worker=0)
client = Client(cluster)

client
# <Client: 'tcp://127.0.0.1:46121' processes=0 threads=0, memory=0 B> 
```

For the next case, I am not sure that the behavior should be. Here the `nthreads` is being passed int the `worker_kwargs` as 0. So we don't really have control over what happens after that. In this case it sets nthreads for each worker to `CPU_COUNT`

```python

cluster = LocalCluster(n_workers=2, threads_per_worker=0)
client = Client(cluster)

client
# <Client: 'tcp://127.0.0.1:46509' processes=2 threads=24, memory=33.37 GB>
```

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -161,8 +161,10 @@ class LocalCluster(SpecCluster):
             else:
                 n_workers = 1
                 threads_per_worker = CPU_COUNT
-        if n_workers is None and threads_per_worker is not None:
+        if n_workers is None and threads_per_worker:
             n_workers = max(1, CPU_COUNT // threads_per_worker)
+        if n_workers is None:
+            n_workers = 0
         if n_workers and threads_per_worker is None:
             # Overcommit threads per worker, rather than undercommit
             threads_per_worker = max(1, int(math.ceil(CPU_COUNT / n_workers)))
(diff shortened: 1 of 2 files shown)


jsignell:
@gforsyth who I was looking at this with.

mrocklin:
I think that we should raise if `threads_per_worker < 1`

When testing in the distributed repository async tests are preferred, if they're easy to use.  You'll also want to use the `cleanup` fixture.

jsignell:
> I think that we should raise if threads_per_worker < 1

That makes sense to me and should result in unambiguous behavior.

jsignell:
@mrocklin I made it a warning instead of an error since it is easy enough to set it to None.

mrocklin:
Thanks @jsignell !  Merged.

PR #3993, opened 2020-07-28
enable uploaded delayed functions to refresh

jackxxu:
### tl;dr

This PR enables delayed functions in newer versions of uploaded python modules to be applied without having to restart workers.

### background

`delayed` has been instrumental in helping us to parallelize our custom algorithms to run in the cloud. Because our custom algorithms are rather involved, we organize them into modules and upload them using `upload_file`, and then in the main module decorate them with `delayed` before sending it to `dask.compute`. 

### problem

We found out that subsequent uploads of modules with new versions of the functions don't refresh in the worker. In other words, after running `upload_file`, only the very first version of these functions are being used in delayed, until we reboot the worker.

This is highly undesirable operationally because we often times don't get the permission to restart workers in the cloud, which understandably is obtrusive to other tenants of dask cluster. 

### cause 

it turns out that each dask worker maintains a `LRU` OrderedDict of pickle loaded delayed functions, likely for performance reasons. The cause of the problem is that these references don't refresh when a new version of these functions are uploaded.

### what is involved
1. added a new method in the `utils.LRU` class called `clear` to clear the cache
2. when a new file is uploaded, `clear` method is called to invalidate the cache and allow the cache to refresh.
3. added a unit test and a "functional" test.

### DIFF
--- a/distributed/tests/test_utils.py
+++ b/distributed/tests/test_utils.py
@@ -625,10 +625,6 @@ def test_lru():
     assert len(l) == 3
     assert list(l.keys()) == ["c", "a", "d"]
 
-    # clear
-    l.clear()
-    assert len(l) == 0
-    assert list(l.keys()) == []
 
 @pytest.mark.asyncio
 async def test_offload():
--- a/distributed/tests/test_utils.py
+++ b/distributed/tests/test_utils.py
@@ -625,6 +625,10 @@ def test_lru():
     assert len(l) == 3
     assert list(l.keys()) == ["c", "a", "d"]
 
+    # clear
+    l.clear()
+    assert len(l) == 0
+    assert list(l.keys()) == []
 
 @pytest.mark.asyncio
 async def test_offload():
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1555,6 +1555,9 @@ class LRU(UserDict):
             self.data.popitem(last=False)
         super().__setitem__(key, value)
 
+    def clear(self):
+        self.data.clear()
+
 
 def clean_dashboard_address(addr, default_listen_ip=""):
     """
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1555,9 +1555,6 @@ class LRU(UserDict):
             self.data.popitem(last=False)
         super().__setitem__(key, value)
 
-    def clear(self):
-        self.data.clear()
-
 
 def clean_dashboard_address(addr, default_listen_ip=""):
     """
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -1547,7 +1547,7 @@ async def test_upload_file_refresh_delayed(c, s, a, b):
             with tmp_text("myfile.py", "def f():\n    return {}".format(value)) as fn:
                 await c.upload_file(fn)
 
-            sys.path.append(fn)
+            sys.path.append(os.path.dirname(fn))
             from myfile import f
 
             b = delayed(f)()
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -1541,6 +1541,22 @@ async def test_upload_file(c, s, a, b):
 
 
 @gen_cluster(client=True)
+async def test_upload_file_refresh_delayed(c, s, a, b):
+    with save_sys_modules():
+        for value in [123, 456]:
+            with tmp_text("myfile.py", "def f():\n    return {}".format(value)) as fn:
+                await c.upload_file(fn)
+
+            sys.path.append(fn)
+            from myfile import f
+
+            b = delayed(f)()
+            bb = c.compute(b, sync=False)
+            result = await c.gather(bb)
+            assert result == value
+
+
+@gen_cluster(client=True)
 async def test_upload_file_no_extension(c, s, a, b):
     with tmp_text("myfile", "") as fn:
         await c.upload_file(fn)
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -960,7 +960,7 @@ class Worker(ServerNode):
         if load:
             try:
                 import_file(out_filename)
-                cache_loads.clear()
+                cache_loads.data.clear()
             except Exception as e:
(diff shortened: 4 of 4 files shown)


distributed/utils.py
```diff
@@ -1555,6 +1555,9 @@ def __setitem__(self, key, value):
             self.data.popitem(last=False)
         super().__setitem__(key, value)
 
+    def clear(self):
```
quasiben, line 1558: This seems dangerous -- in `self.data` is where workers will store a MutabbleMapping of keys and values -- essentially all the data on a worker.  Perhaps `importlib.reload` might be a better/safer option here

jxeagle: @quasiben thanks for the feedback!

in this particular case, the `LRU.clear` method is only called on `worker.cache_loads`, a LRU instance.

`cache_loads` is a **pass-thru cache** of mapping of pickled functions to the their loaded (unpickled) function objects. Besides the `clear` method, here is the only place it is set and accessed: https://github.com/dask/distributed/blob/master/distributed/worker.py#L3270-L3273

`cache_loads` seems to be a redundant state of functions only, and pickled function and args are passed in with each delayed call, but this cache aims to alleviate the burden of loading the pickled function every time by caching the already loaded function object. 

the problem, which this PR is meant to address, is that some of these function objects can be outdated after each file upload. 

Perhaps I am missing something that is potential harmful?

jackxxu: @quasiben thanks for the feedback!

in this particular case, the LRU.clear method is only called on worker.cache_loads, a LRU instance.

cache_loads is a pass-thru cache of mapping of pickled functions to the their loaded (unpickled) function objects. Besides the clear method, here is the only place it is set and accessed: https://github.com/dask/distributed/blob/master/distributed/worker.py#L3270-L3273

cache_loads seems to be a redundant state of functions only, and pickled function and args are passed in with each delayed call, but this cache aims to alleviate the burden of loading the pickled function every time by caching the already loaded function object.

the problem, which this PR is meant to address, is that some of these function objects can be outdated after each file upload.

Perhaps I am missing something that is potential harmful?

quasiben: Nope, you're good. I misread the the self.data.clear line and thought this was happening in the worker, not the LRU cache.

quasiben:
In general this looks good.  The CI failures here are unrelated to the PR.  @jrbourbeau if you have a minute to also review that would be appreciated as I think you built LRU setup.  My concern earlier was that we may be inadvertently clearing keys from LRU, I now think we this is only affect function caching (as @jackxxu has said) and thus we should be ok.

jrbourbeau:
Thanks for the PR @jackxxu! I'm a little concerned about clearing `cache_loads` as there may be a performance regression in other workflows (e.g. those where `upload_file` is used only once). Alternatively, you could manually clear `cache_loads` without restarting your workers. For example:

```python

# Upload file
client.upload_file(fn)

def clear_cache(dask_worker):
    from distributed.worker import cache_loads
    cache_loads.data.clear()

# Clear cache
result = client.run(clear_cache)
```

Additionally, there's another PR which attempts to disable worker function caching altogether (xref https://github.com/dask/distributed/pull/3983). Perhaps that would also be of benefit to you too?

jackxxu:
@jrbourbeau Thanks for your feedback and your code snippet. It essentially implements the logic in this PR on the client side, and it definitely works. 

However, I wonder if this cache invalidation logic is specific to my use case or it is better suited to be generalized and implemented in worker side.

To be clear, this PR will NOT cause performance regression in a workflow where upload_file is used only once as cache is cleared only when file is uploaded. in other words, in an environment where file is uploaded one time, [this cache_loads.clear method](https://github.com/jackxxu/distributed/blob/clear_Load_cache/distributed/worker.py#L963) will never be invoked after the initial upload. 

There is a spectrum of caching of the function objects: 
* The current master branch is one extreme where cache for a particular pickled byte sequence is set once and once only. While it surely is performant, this causes outdated cache problem with subsequent file uploads.
* https://github.com/dask/distributed/pull/3983 is the other end of the spectrum where it never caches and the pickled function in every client method call is loaded every time. I will NOT use this option in production because, while it surely solve the problem with the outdated cache, it will inevitably cause performance regression that you mentioned for every method call, at an extent that largely depends on the size of the pickled object. 

This PR essentially implements a go-between that invalidates the cache only when a new file is uploaded, which presumably doesn't occur every often, especially in production. 

I know this is one of the 2 hardest things in computer science. :) 

What are your thoughts?

distributed/utils.py
mrocklin: I recommend that we remove this method and just use `cache_loads.data.clear()` instead.  

Motivation: http://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection

distributed/tests/test_utils.py
```diff
@@ -625,6 +625,10 @@ def test_lru():
     assert len(l) == 3
     assert list(l.keys()) == ["c", "a", "d"]
 
+    # clear
+    l.clear()
+    assert len(l) == 0
+    assert list(l.keys()) == []
```
mrocklin, line 631: I wouldn't mind dropping this part of the test and just relying on the integration test that you have in `test_client.py`.  I think that that will capture the intent as effectively and be robust to future internal changes.

jxeagle: addressed w/ https://github.com/dask/distributed/pull/3993/commits/80895f627c9abf2517c0d337994ed9f466c02b40

jxeagle:
@mrocklin thanks for taking your time to review and I have fixed it per your recommendation.  i get why you recommend it, particularly in community code.

distributed/tests/test_utils.py
jackxxu: addressed w/ 80895f6

jackxxu:
@mrocklin thanks for taking your time to review and I have fixed it per your recommendation. i get why you recommend it, particularly in community code.

mrocklin:
Thanks @jackxxu .  This is in.

PR #3996, opened 2020-07-28
Only call `frame_split_size` when there are frames

jakirkham:
This just becomes a no-op if an empty list is given (returning an empty list). However there is no need to do this work if we don't need it. So simply restrict calling `frame_split_size` to the case where frames are present and compression will occur.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -475,8 +475,8 @@ def serialize_bytelist(x, **kwargs):
     header, frames = serialize(x, **kwargs)
     if "lengths" not in header:
         header["lengths"] = tuple(map(nbytes, frames))
-    frames = sum(map(frame_split_size, frames), [])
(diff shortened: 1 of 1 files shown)


quasiben:
Ive been seeing the `test_bandwidth` in a few other PRs.  Would you mind merging with master and resubmitting ?

jakirkham:
Unfortunately this is already based on the latest commit in `dask/master`.

quasiben:
I restarted CI to see if that fixes things

jakirkham:
Yeah `test_bandwidth` also fails on `master`. Submitted PR ( https://github.com/dask/distributed/pull/3999 ) in an attempt to fix this.

quasiben:
Looks good.  Thanks @jakirkham

jakirkham:
Thanks Ben 🙂

PR #3998, opened 2020-07-28
Delay deserialization of Data in workers until actual usage.

Carreau:
In some case deserialization is not thread safe and it would be good to
only deserialize if no work is being done at the same time.

This is trying to achieve this in part by delaying until unpicking
until the data is actually needed to be sent to the Executor. In the
case of a single thread TPE that would mean non-thread safe code can be
unpickled. Maybe we even want to move unpicking into the Executor, but
then we need the executor to support this; and we might want to make
sure we don't unpickle many times.

This of course needs a gross hack by looking at frames, but at least I
hope this will get some conversation started.

This also screws up the computation of per-type bandwidth; unless we
delay the BW calculation per type.

There is also some issues as clients will also call indirectly `get_data_from_worker`, and for example `await client.submit(lambda x: x + 1, 10)` will return a `Serialized` object instead of expected result...

### DIFF
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -123,7 +123,12 @@ def loads(frames, deserialize=True, deserializers=None):
             else:
                 fs = []
 
-            if deserialize or key in bytestrings:
+            import inspect
+
+            stack = inspect.stack()
+            get_data = any([s.function == "get_data_from_worker" for s in stack])
+
+            if (deserialize or key in bytestrings) and not get_data:
                 if "compression" in head:
                     fs = decompress(head, fs)
                 if not any(hasattr(f, "__cuda_array_interface__") for f in fs):
--- a/distributed/utils_comm.py
+++ b/distributed/utils_comm.py
@@ -12,6 +12,7 @@ from tlz import merge, concat, groupby, drop
 
 from .core import rpc
 from .utils import All, tokey
+from .protocol.serialize import Serialized
 
 logger = logging.getLogger(__name__)
 
@@ -84,7 +85,12 @@ async def gather_from_workers(who_has, rpc, close=True, serializers=None, who=No
                     )
                     missing_workers.add(worker)
                 else:
-                    response.update(r["data"])
+                    response.update(
+                        {
+                            k: (v.deserialize() if isinstance(v, Serialized) else v)
+                            for k, v in r["data"].items()
+                        }
+                    )
         finally:
             for r in rpcs.values():
                 await r.close_rpc()
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -40,6 +40,7 @@ from .node import ServerNode
 from . import preloading
 from .proctitle import setproctitle
 from .protocol import pickle, to_serialize, deserialize_bytes, serialize_bytelist
+from .protocol.serialize import Serialized
 from .pubsub import PubSubWorkerExtension
 from .security import Security
 from .sizeof import safe_sizeof as sizeof
@@ -2021,7 +2022,7 @@ class Worker(ServerNode):
                     self.bandwidth = self.bandwidth * 0.95 + bandwidth * 0.05
                     bw, cnt = self.bandwidth_workers[worker]
                     self.bandwidth_workers[worker] = (bw + bandwidth, cnt + 1)
-
+                    # this is going to be wrong as all types will be "Serialized", unless we delay this.
                     types = set(map(type, response["data"].values()))
                     if len(types) == 1:
                         [typ] = types
@@ -2031,9 +2032,11 @@ class Worker(ServerNode):
                 if self.digests is not None:
                     self.digests["transfer-bandwidth"].add(total_bytes / duration)
                     self.digests["transfer-duration"].add(duration)
+                # that's still ok
                 self.counters["transfer-count"].add(len(response["data"]))
                 self.incoming_count += 1
 
+                # that's still ok
                 self.log.append(("receive-dep", worker, list(response["data"])))
             except EnvironmentError as e:
                 logger.exception("Worker stream died during communication: %s", worker)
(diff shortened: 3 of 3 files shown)


Carreau:
My other idea was to use a custom serializer/deserializer that would be lazy, the problems being that: 

1) This still needs hooks in the right place to deserialize just in time. 
2) This appear to make other things beyond just the deps/data use the deserializer, but it's unclear as this was mostly failing in a number of places.


I'm also wondering if keeping the serialized data in worker, if it's needed somewhere else as then worked don't need to re-serialise it to be sent...

mrocklin:
Yeah, this seems like a big enough hack that we would probably reject it without a lot of evidence that it was necessary in a variety of situations.  In general I would recommend that any group facing things like this look into improving their serialization, or using locks to protect finicky resources.  You could also look into using a synchronous executor if you wanted everything to be in one thread.

The closest common situation I can recall that sounds like this one is dealing with TensorFlow graphs, which don't like being deserialized in one thread and then used in another.  

Any fix like this would have, I think, enough unexpected results that I'd be very hesitant to consider it.

Carreau:
> Yeah, this seems like a big enough hack that we would probably reject it without a lot of evidence that it was necessary in a variety of situations.

Do you mean the inspect stack, or keeping obj as serialized until necessary ?

I can likely find another way to pass the information that data should be lazyly deserialized; but mostly wanted to know if it's worth looking into.

Also yes, the group is looking into improving thread safety and (de)serialisation of objects, and already on a single thread TPE to minimize issues, and looking into locks.

mrocklin:
In general lazy deserialization.  Or if we do want to do this that's fine, but we need to survey other use cases first, see how important it is, figure out what a convention might be that makes sense, etc..  This feels like tacking on a feature for a single user's use case fairly deep into the guts of Dask.  

My guess is that there is some more general solution somewhere that lets this group get what they want, and doesn't add an odd corner case into the core.

mrocklin:
Alternatively, if this *is* the right solution, then I think that we need to build a case for it.  The best case I can think of that is similar to this is TensorFlow/Keras, as mentioned above.

quasiben:
I thought we could already control serialization/deserialization with a separate thread with the config option: [distributed.comm.offload](https://docs.dask.org/en/latest/configuration-reference.html#distributed.comm.offload) .  Does that not work ?

Carreau:
> Does that not work ?

I don't believe it does, the core of the problem being that some types objects don't like having work being done at the same time and other instances being deserialized. 

deactivating offload already force the deserialisation to be done in the main loop, but then the ThreadPool executor is still running; And that's not good. My goal here is to reduce the chance of deserialisation happening while the TPE is doing work.

Either the objects need to have a both when in use that block deserialisation; OR I need dask/distributed to not deserialize while computing. 

Right now this is hard as unpacking messages unpacks data as well.

jakirkham:
There is a shared interest in delaying deserialization I think. For example here we discussed delaying deserialization to reduce memory usage ( https://github.com/rapidsai/dask-cuda/issues/342#issuecomment-656609142 ). I think how that would look may differ a bit from how it is currently implement here, but the overall objective seems agreeable.

cc @madsbk @pentschev

madsbk:
> 
> 
> There is a shared interest in delaying deserialization I think. For example here we discussed delaying deserialization to reduce memory usage ( [rapidsai/dask-cuda#342 (comment)](https://github.com/rapidsai/dask-cuda/issues/342#issuecomment-656609142) ). I think how that would look may differ a bit from how it is currently implement here, but the overall objective seems agreeable.
> 
> cc @madsbk @pentschev

Yes, in [rapidsai/dask-cuda#342 (comment)](https://github.com/rapidsai/dask-cuda/issues/342#issuecomment-656609142) the plan is to delay the deserialization until the data is accessed by the task. It even enables tasks to coordinate deserializations with other work explicitly.
 
I currently am on vacation but I will start working on this when I get back next week.

Carreau:
The problem with not having dask being aware of this is then user get exposed to those proxy objects you suggest in https://github.com/rapidsai/dask-cuda/issues/342#issuecomment-656609142 ;

I think it would be good to have a solution that works regardless of the types; or the serializer/deserializer involved.

I completely agree that the implementation here is horrible with stack inspection.

Please ping me on any work you are doing on this on the dask-cuda, I'm happy to help make it more generic.

jakirkham:
I wonder if we can push deserialization into [`apply_function`]( https://github.com/dask/distributed/blob/8ebd396bc28a829ae9a928ed590ac2a8ef7dc301/distributed/worker.py#L3389 ) and [`apply_function_actor`]( https://github.com/dask/distributed/blob/8ebd396bc28a829ae9a928ed590ac2a8ef7dc301/distributed/worker.py#L3436 ), which are where the actual function winds up being called for computation on the `Executor`. This should delay deserialization until the data is going to be used.

Carreau:
> I wonder if we can push deserialization into apply_function and apply_function_actor,

Does this have the potential to stall some task/duplicate work if the same deserialized data is required by 2 tasks.

mrocklin:
I think that deserializing data when we get it is more sensible in the common case situation.  I want to make sure that we're not mucking about with sensible behaviors because of a few odd cases.

madsbk:
> The problem with not having dask being aware of this is then user get exposed to those proxy objects you suggest in [rapidsai/dask-cuda#342 (comment)](https://github.com/rapidsai/dask-cuda/issues/342#issuecomment-656609142) ;

Agree, this is are disadvantages with the approach. We can mitigate this by make the proxy object as transparent as possible but you are right, it shouldn't be on by default. 
However, if the application is really sensitive to serialization and deserialization you properly want to coordinate the work explicitly anyways?

jrbourbeau:
https://github.com/dask/distributed/pull/4307 should allow users to not deserialize and run tasks at the same time

Carreau:
Closing as stale.

PR #3999, opened 2020-07-29
Fix failing `test_bandwidth`

jakirkham:
It seems this test has been failing in a few places. Here's [one build]( https://travis-ci.org/github/dask/distributed/jobs/713023658#L3259 ) that failed. This tweaks the test to get it to pass again. Not sure if this is the right solution. So am open to other ideas.

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1635,7 +1635,7 @@ async def test_idle_timeout(c, s, a, b):
 @gen_cluster(client=True, config={"distributed.scheduler.bandwidth": "100 GB"})
 async def test_bandwidth(c, s, a, b):
     start = s.bandwidth
-    x = c.submit(operator.mul, b"0", 1000000, workers=a.address)
+    x = c.submit(operator.mul, b"0", 1000001, workers=a.address)
     y = c.submit(lambda x: x, x, workers=b.address)
     await y
     await b.heartbeat()
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1639,7 +1639,7 @@ async def test_bandwidth(c, s, a, b):
     y = c.submit(lambda x: x, x, workers=b.address)
     await y
     await b.heartbeat()
-    assert s.bandwidth < start  # we've learned that we're slower
+    assert s.bandwidth <= start  # we've learned that we're slower
     assert b.latency
     assert typename(bytes) in s.bandwidth_types
     assert (b.address, a.address) in s.bandwidth_workers
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1639,7 +1639,7 @@ async def test_bandwidth(c, s, a, b):
     y = c.submit(lambda x: x, x, workers=b.address)
     await y
     await b.heartbeat()
-    assert s.bandwidth <= start  # we've learned that we're slower
+    assert s.bandwidth < start  # we've learned that we're slower
     assert b.latency
     assert typename(bytes) in s.bandwidth_types
     assert (b.address, a.address) in s.bandwidth_workers


jakirkham:
Looks like there is a deeper issue...

```
>       assert typename(bytes) in s.bandwidth_types
1695E       assert 'builtins.bytes' in defaultdict(<class 'float'>, {})
1696E        +  where 'builtins.bytes' = typename(bytes)
1697E        +  and   defaultdict(<class 'float'>, {}) = <Scheduler: "tcp://127.0.0.1:38669" processes: 2 cores: 3>.bandwidth_types
```

https://travis-ci.org/github/dask/distributed/jobs/713046159#L1582-L1585

jakirkham:
I'm guessing we expect to exercise [this code path]( https://github.com/dask/distributed/blob/8ebd396bc28a829ae9a928ed590ac2a8ef7dc301/distributed/worker.py#L2020-L2029 ), but that doesn't seem to be happening for some reason.

jakirkham:
Should add I'm not able to reproduce this issue outside of CI.

quasiben:
Thanks for looking at this John.  I am also investigating -- I don't see why this would start failing

quasiben:
After a bit more investigation I think this issue might come down to sizeof calculations.  In https://github.com/dask/dask/pull/6457 we changes the sizeof measurement for bytes objects from `sys.getsizeof` to `len`

```python
In [1]: import sys

In [2]: arr = b"0" * 1000000

In [3]: len(arr)
Out[3]: 1000000

In [4]: sys.getsizeof(arr)
Out[4]: 1000033
```

And, if you checkout the commit just prior to PR https://github.com/dask/dask/pull/6457 everything passes. I suspect the change here is ok.   And will merge late afternoon unless others want to comment

distributed/tests/test_scheduler.py
```diff
@@ -1635,11 +1635,11 @@ async def test_idle_timeout(c, s, a, b):
 @gen_cluster(client=True, config={"distributed.scheduler.bandwidth": "100 GB"})
 async def test_bandwidth(c, s, a, b):
     start = s.bandwidth
-    x = c.submit(operator.mul, b"0", 1000000, workers=a.address)
+    x = c.submit(operator.mul, b"0", 1000001, workers=a.address)
     y = c.submit(lambda x: x, x, workers=b.address)
     await y
     await b.heartbeat()
-    assert s.bandwidth < start  # we've learned that we're slower
+    assert s.bandwidth <= start  # we've learned that we're slower
```
mrocklin, line 1642: This change seems to remove the intent of the test, which I think is implied by the comment here

distributed/tests/test_scheduler.py
```diff
@@ -1635,11 +1635,11 @@ async def test_idle_timeout(c, s, a, b):
 @gen_cluster(client=True, config={"distributed.scheduler.bandwidth": "100 GB"})
 async def test_bandwidth(c, s, a, b):
     start = s.bandwidth
-    x = c.submit(operator.mul, b"0", 1000000, workers=a.address)
+    x = c.submit(operator.mul, b"0", 1000001, workers=a.address)
```
mrocklin, line 1638: Why this change?

quasiben: When using `len` the `metrics["bandwidth"]["total"]` is reported to be `100000000000`, always.  When using `getsize` of the `metrics["bandwidth"]["total"]` at the end is `95012655208`.

distributed/tests/test_scheduler.py
quasiben: Agreed, we don't need this change

distributed/tests/test_scheduler.py
mrocklin: Understood.  I think I'm still confused on why we changed 1000000 to 1000001 here though.

jakirkham:
To answer your questions, I don't know. Just trying things to see how we fix this. Happy to try something different. Suggestions welcome 🙂

distributed/tests/test_scheduler.py
quasiben: After a bit more digging I we need to increase here because we only update the bandwidth if the total data is larger than  1_000_000: https://github.com/dask/distributed/blob/7cdba1d7e17df7be005f4d4b8ca528065a3f83a7/distributed/worker.py#L2020-L2029

So to wrap this all up, when using `getsize` we are just over the line and we update bandwidth.  My guess is that we picked that value because we didn't want to update when passing small data such as heartbeats

mrocklin: Ah ok, that makes sense.  Thanks for the explanation @quasiben !

quasiben:
@jakirkham I've update the test in the last commit I think after it passes CI we are good to merge.  Thank you for your work here!

quasiben:
Woot!  Tests pass.  Thanks again @jakirkham for the work and @mrocklin for the reviews

jakirkham:
Thanks all! 😄

PR #3995, opened 2020-07-28
Test Pandas `Series` is writeable

jakirkham:
Follow-up to PR ( https://github.com/dask/distributed/pull/3967 ) and issue ( https://github.com/dask/distributed/issues/3943 )

Includes a test to ensure a Pandas `Series` is writeable after round-trip serialization.

Issue #3994, opened 2020-07-28
Copy for mutable frames can introduce a slowdown


jakirkham:
Recently we made a fix to ensure all frames are mutable in PR ( https://github.com/dask/distributed/pull/3967 ). This solves some issues where immutable frames were used to back Python objects.

One downside of that change is it effectively requires a copy of the frames be taken for all objects. This can be unnecessary in cases where the data may already get copied as is the case with Python builtin objects that like to own their memory or GPU objects that require data be moved from host to device (effectively a copy). As a result this copy can introduce a slowdown.

Included below is an example with and without that change using cuDF:

<details>
<summary>Without mutable frames:</summary>

```python
In [1]: import cupy 
   ...: import cudf 
   ...: import rmm 
   ...:  
   ...: from distributed.protocol import serialize_bytes, deserialize_bytes     

In [2]: rmm.reinitialize(pool_allocator=True, 
   ...:                  initial_pool_size=int(30 * 2**30))                     

In [3]: df = cudf.DataFrame({ 
   ...:     k: cupy.random.random(1_000_000) 
   ...:     for i, k in enumerate(map(chr, range(ord("A"), ord("K")))) 
   ...: })                                                                      

In [4]: b = serialize_bytes(df)                                                 

In [5]: %timeit deserialize_bytes(b)                                            
10.1 ms ± 61.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

</details>

<details>
<summary>With mutable frames:</summary>

```python
In [1]: import cupy 
   ...: import cudf 
   ...: import rmm
   ...: from distributed.protocol import serialize_bytes, deserialize_bytes

In [2]: rmm.reinitialize(pool_allocator=True,
   ...:                  initial_pool_size=int(30 * 2**30))

In [3]: df = cudf.DataFrame({
   ...:     k: cupy.random.random(1_000_000)
   ...:     for i, k in enumerate(map(chr, range(ord("A"), ord("K"))))
   ...: })

In [4]: b = serialize_bytes(df)

In [5]: %timeit deserialize_bytes(b)
45.4 ms ± 65.3 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

</details>

jsignell:
That is a pretty sizable performance hit. Do you have a proposed solution? I suppose the user could pick whether frames are mutable or not. Is there a smart way for us to determine that for them so that we can provide the best default experience?

jakirkham:
[...]

PR #4004, opened 2020-07-31
Track mutable frames

jakirkham:
Fixes https://github.com/dask/distributed/issues/3994
Closes https://github.com/dask/distributed/pull/3995 (as it's included)

Instead of making all frames writeable as was done in PR ( https://github.com/dask/distributed/pull/3967 ), this tracks which frames were writeable and only makes sure those are writeable. This also fixes a performance regression. Thanks @jsignell for the suggestion 🙂

```python
In [1]: import cupy
   ...: import cudf
   ...: import rmm
   ...: 
   ...: from distributed.protocol import serialize_bytes, deserialize_bytes

In [2]: rmm.reinitialize(pool_allocator=True,
   ...:                  initial_pool_size=int(30 * 2**30))

In [3]: df = cudf.DataFrame({
   ...:     k: cupy.random.random(1_000_000)
   ...:     for i, k in enumerate(map(chr, range(ord("A"), ord("K"))))
   ...: })

In [4]: b = serialize_bytes(df)

In [5]: %timeit deserialize_bytes(b)
10.1 ms ± 50.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Edit: Had a typo in my benchmark before. It has been updated.

Note: The `merge_frames` code in this PR is now closer to what it looked like in `2.21.0` (thanks to some simplifications :). So it may be easier to compare what is here against the code in that version for a cleaner diff.

cc @quasiben

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1108,7 +1108,12 @@ def nbytes(frame, _bytes_like=(bytes, bytearray)):
 
 
 def is_writeable(frame):
-    """ Check whether frame is writeable """
+    """
+    Check whether frame is writeable
+
+    Will return ``True`` if writeable, ``False`` if readonly, and
+    ``None`` if undetermined.
+    """
     try:
         return not memoryview(frame).readonly
     except TypeError:
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1107,6 +1107,14 @@ def nbytes(frame, _bytes_like=(bytes, bytearray)):
             return len(frame)
 
 
+def is_writeable(frame):
+    """ Check whether frame is writeable """
+    try:
+        return not memoryview(frame).readonly
+    except TypeError:
+        return None
+
+
 @contextmanager
 def time_warn(duration, text):
     start = time()
--- a/distributed/protocol/core.py
+++ b/distributed/protocol/core.py
@@ -7,7 +7,7 @@ import msgpack
 from .compression import compressions, maybe_compress, decompress
 from .serialize import serialize, deserialize, Serialize, Serialized, extract_serialize
 from .utils import frame_split_size, merge_frames, msgpack_opts
-from ..utils import nbytes
+from ..utils import is_writeable, nbytes
 
 _deserialize = deserialize
 
@@ -47,9 +47,7 @@ def dumps(msg, serializers=None, on_error="message", context=None):
 
         for key, (head, frames) in data.items():
             if "writeable" not in head:
-                head["writeable"] = tuple(
-                    not f.readonly for f in map(memoryview, frames)
-                )
+                head["writeable"] = tuple(map(is_writeable, frames))
             if "lengths" not in head:
                 head["lengths"] = tuple(map(nbytes, frames))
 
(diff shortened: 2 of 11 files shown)


jakirkham:
Additionally have managed to back out the requirement that all frames be `memoryview`s. Also included tests to make sure NumPy arrays and Pandas objects built from read-only frames are reconstructed with writeable frames (if they were originally).

quasiben:
Thanks for doing continuing to push here @jakirkham. I have two comments:

1) I think updating a docstring somewhere on how `writeable` is used would be helpful.  Maybe in `core.py` -- this is only used for GPU data correct ?  
2) Would you be willing to confirm test_cupy/numba/rmm/test_collection_cuda/test_ucx pass as well.

distributed/utils.py
```diff
@@ -1107,6 +1107,14 @@ def nbytes(frame, _bytes_like=(bytes, bytearray)):
             return len(frame)
 
 
+def is_writeable(frame):
+    """ Check whether frame is writeable """
+    try:
+        return not memoryview(frame).readonly
+    except TypeError:
+        return None
```
quasiben, line 1115: This is a small nitpick but should this be False instead of None ?

jakirkham: No it should return `None`. `True` means it is writeable, `False` means it is readonly, and `None` means we can't tell so ignore. This impacts whether we need to copy or not later. So `None` effectively means don't try to perform a copy to make it writeable or readonly.

jakirkham: Added to the docstring of this function. Please let me know if that makes sense 🙂

jakirkham:
> 1. I think updating a docstring somewhere on how `writeable` is used would be helpful.  Maybe in `core.py` -- this is only used for GPU data correct ?

Hopefully `is_writeable` addresses this. Though happy to expand on this further if there are particular points that remain unclear

> 2. Would you be willing to confirm test_cupy/numba/rmm/test_collection_cuda/test_ucx pass as well.

I've been running the full protocol test suite on a DGX-1, which has worked well. Haven't been testing comms though. However it seems there are other issues without this change as evidenced by PR ( https://github.com/rapidsai/ucx-py/pull/567 ).

jakirkham:
>> 2. Would you be willing to confirm test_cupy/numba/rmm/test_collection_cuda/test_ucx pass as well.
>
> I've been running the full protocol test suite on a DGX-1, which has worked well. Haven't been testing comms though. However it seems there are other issues without this change as evidenced by PR ( https://github.com/rapidsai/ucx-py/pull/567 ).

Ok the UCX comms tests pass for me locally too. Though we should still figure out what is up with UCX-Py's CI.

jakirkham:
> > > 1. Would you be willing to confirm test_cupy/numba/rmm/test_collection_cuda/test_ucx pass as well.
> > 
> > 
> > I've been running the full protocol test suite on a DGX-1, which has worked well. Haven't been testing comms though. However it seems there are other issues without this change as evidenced by PR ( [rapidsai/ucx-py#567](https://github.com/rapidsai/ucx-py/pull/567) ).
> 
> Ok the UCX comms tests pass for me locally too. Though we should still figure out what is up with UCX-Py's CI.

Turns out to be related to how the test was written. PR ( https://github.com/rapidsai/ucx-py/pull/568 ) should fix this.

quasiben:
Thanks John.  Merging

PR #4008, opened 2020-08-01
Make GPU plots robust to not having GPUs

mrocklin:
cc @quasiben @jacobtomlinson for review

### DIFF
--- a/distributed/dashboard/components/nvml.py
+++ b/distributed/dashboard/components/nvml.py
@@ -134,7 +134,10 @@ class GPUCurrentLoad(DashboardComponent):
             i = 0
 
             for ws in workers:
-                info = ws.extra["gpu"]
+                try:
+                    info = ws.extra["gpu"]
+                except KeyError:
+                    continue
                 metrics = ws.metrics["gpu"]
                 for j, (u, mem_used, mem_total) in enumerate(
                     zip(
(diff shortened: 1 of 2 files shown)


distributed/dashboard/components/nvml.py
```diff
@@ -134,7 +134,10 @@ def update(self):
             i = 0
 
             for ws in workers:
-                info = ws.extra["gpu"]
+                try:
+                    info = ws.extra["gpu"]
+                except KeyError:
+                    continue
                 metrics = ws.metrics["gpu"]
```
quasiben, line 141: Should `ws.metrics["gpu"]` also be in the try ?

quasiben:
This looks good to me. Thanks @mrocklin! 

What will be displayed when GPUs are not installed ?

mrocklin:
An empty plot

quasiben:
What do you think about including a message saying no GPUs were detected ?

mrocklin:
I'm open to that.  We might do the same on all of the plots that show usage
like CPU usage, memory usage, GPU utilization / memory, and so on.

I'm not sure how best to do this with Bokeh though

quasiben:
Maybe this could be done with a template like


```
{% extends main.html %}
{% block content %}
   {% if not {{GPU-FLAG}} %}
     <--GPU IS MISSING MESSAGE--> 
   {% endif %} 
{% end %}
```

Maybe this can be followed up with a subsequent PR

mrocklin:
What if GPU machines showed up later?  I think that we would have to do
this dynamically.

Issue #4011, opened 2020-08-03
Increase flexibility around Python versions and compression


mrocklin:
Currently Dask fails to communicate when we try to cross Python versions or have mismatched compression.  The Python version mismatch is new-ish ever since we added Pickle protocol 5, the compression is a long-standing issue.  In principle we ask that users provide consistent versions across all machines, however we may still be able to accomodate things here with a little bit of work on our end.  

When establishing a connection between two endpoints we might have the two sides publish their capabilities.  This might inform the other side which compression libraries they have available, whether or not they support Pickle5, or whether or not they have a GPU.  This might make communication in heterogeneous situations a bit easier.  

We already have some work here.  There is a `context=` keyword that is passed down from `Comm.write` to `to_frames`, which in turn handles serialization.  We could start populating that with more information and using it in serialization/compression.  

This would allow the following situations:

1.  More relaxation around Python versions
2.  More relaxation around compression libraries
3.  Downgrading GPU results smoothly to CPU clients

cc @quasiben @jakirkham

jakirkham:
Messaging capabilities makes sense. Though I would think we would want to cache these somewhere (just to avoid sending this info repeatedly).

As all GPU things implement `"cuda"` and `"dask"` serialization, it should be simple matter to serialize them with either as appropriate. This is how things like TCP and spilling work today.

Though currently compression is not that configurable (or at least if it is it is not clear to me how to do this). How would you envision compression being handled here?

To pickle protocol 5, I wonder if we shouldn't just give users a hook to enable or disable this ( https://github.com/dask/distributed/pull/4012 ). Seems pretty lightweight.

jakirkham:
That said, one thing I'm curious about is in what cases you envision sending GPU objects through a comm to a worker that lacks a GPU. In particular what would that worker do with that message? Or am I misunderstanding something here?

mrocklin:
> Messaging capabilities makes sense. Though I would think we would want to cache these somewhere (just to avoid sending this info repeatedly).

Agreed, I think that we would send this information when first connecting.

> That said, one thing I'm curious about is in what cases you envision sending GPU objects through a comm to a worker that lacks a GPU. In particular what would that worker do with that message? Or am I misunderstanding something here?

[...]

PR #4012, opened 2020-08-03
Allow pickle protocol to be overridden

jakirkham:
This provides the option to override the pickle protocol used. The default behavior is still to use the `HIGHEST_PROTOCOL`. However this provides users an easy way to customize the pickle protocol through the `distributed.yaml` config.

cc @mrocklin @quasiben

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -170,6 +170,8 @@ distributed:
     log-length: 10000  # default length of logs to keep in memory
     log-format: '%(name)s - %(levelname)s - %(message)s'
     pdb-on-err: False       # enter debug mode on scheduling error
+pickle:
+  protocol: null  # specify the pickle protocol to use
 rmm:
   pool-size: null
 ucx:
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -759,6 +759,18 @@ properties:
             type: boolean
             description: Enter Python Debugger on scheduling error
 
+  pickle:
+    type: object
+    description: |
+      Configuration for pickle serialization
+    properties:
+      protocol:
+        type:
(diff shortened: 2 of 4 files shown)


distributed/distributed.yaml
```diff
@@ -170,6 +170,8 @@ distributed:
     log-length: 10000  # default length of logs to keep in memory
     log-format: '%(name)s - %(levelname)s - %(message)s'
     pdb-on-err: False       # enter debug mode on scheduling error
+pickle:
+  protocol: null  # specify the pickle protocol to use
```
jakirkham, line 174: Just placed this at the global level as it is not really worker or scheduler specific. Not picky on where it lives though.

mrocklin:
Interesting.  I'm curious, what was the motivation?

jakirkham:
Issue ( https://github.com/dask/distributed/issues/4011 ) encouraged me to put it in a PR. Albeit that issue is discussing a different solution. Could also be useful for debugging/helping users.

mrocklin:
Thanks for the speedy response to that issue.  If we were to go this approach (which seems sensible regardless) I would then ask the question

*what should we make the default?  4 or 5?*

I suspect that your answer is likely "5, because it's faster" while I probably tend towards "4, because it's robust".  I'm curious, did we ever speed test pickle protocol 5 on a real-world workload?  Would something like `x + x.T` be made considerably faster by it?  Given the cost around Python version mismatches it might be worth verifying that the performance benefit is felt in real-world situations

jakirkham:
Closing since this seems to be addressed by PR ( https://github.com/dask/distributed/pull/4019 ).

mrocklin:
Woot.  Thanks for the early effort @jakirkham

PR #4013, opened 2020-08-03
Serialize builtin array

jakirkham:
Implements `"dask"` serialization for Python builtin `array`s.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -1,3 +1,4 @@
+from array import array
 from functools import partial
 import traceback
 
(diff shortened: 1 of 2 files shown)


distributed/protocol/serialize.py
```diff
@@ -596,6 +597,25 @@ def _deserialize_bytearray(header, frames):
...
+
+
+@dask_deserialize.register(array)
+def _deserialize_array(header, frames):
+    a = array(header["typecode"])
+    for f in map(memoryview, frames):
+        try:
+            f = f.cast("B")
+        except TypeError:
+            f = f.tobytes()
```
jakirkham, line 614: This is needed because `array.frombytes(...)` expects that [`itemsize == 1`]( https://github.com/python/cpython/blob/580fbb018fd0844806119614d752b41fc69660f9/Modules/arraymodule.c#L1597 ). So we perform a cast to `uint8`, which meets this requirement. If that fails, we copy to `bytes`, which also meets this requirement.

quasiben:
Thanks @jakirkham

PR #4015, opened 2020-08-04
Just extend `frames2` by `frames`

jakirkham:
Instead of coercing `frames` to a `list` to then later copy it again, just include `frames` in whatever iterable form it may be in `frames2`. This avoids the extra copy with the same net effect.


quasiben:
Nice!

PR #4017, opened 2020-08-04
Install PyTorch on CI

jakirkham:
In order to run PyTorch tests on CI, make sure that it gets installed.

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -32,7 +32,7 @@ conda config --set always_yes yes --set quiet yes --set changeps1 no
 conda update conda
 
 # Create conda environment
-conda create -n dask-distributed -c pytorch -c conda-forge -c defaults \
+conda create -n dask-distributed -c conda-forge -c defaults \
     asyncssh \
     bokeh \
     click \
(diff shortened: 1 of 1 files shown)


jakirkham:
Any thoughts on this?

cc @mrocklin @jrbourbeau @quasiben

jrbourbeau:
Thanks for the PR @jakirkham! I'm +0.5 on adding `pytorch` to CI. It'd be great to run `pytorch`-related tests. However it looks like `conda install pytorch` adds a little over 2 minutes to our Travis builds. That's probably not prohibitive given that builds take ~40 minutes today, but small additions do add up over time. Do you know if there have been issues recently where having `pytorch` in CI would have helped?

jakirkham:
We saw some issues where the PyTorch serializers were not being used. So wanted to make sure they are working and tests pass. Additionally we are interested in extending these to handle more efficiently GPU serialization. So it would help to have a sanity check to make sure we don't break what already works in the process.

Would it be better if we only install PyTorch on one job as opposed to all of them?

jakirkham:
Should add I think part of the issue is PyTorch pulls in MKL, which is quite large and takes some time to install.

jrbourbeau:
> Would it be better if we only install PyTorch on one job as opposed to all of them?

Yeah, I think that would make the situation better. 

I'll plan to merge at the end of the day if there are no objections

jakirkham:
Thanks James! Moved to the Python 3.8 job. Though happy to move to a different job if there is a preference.

jakirkham:
Thanks James! 😄

PR #4018, opened 2020-08-04
Try getting cluster dashboard_link before asking scheduler

mrocklin:
This is very slightly more robust


mrocklin:
I'm playing around with the JLab auto-detect button here: https://github.com/dask/dask-labextension/issues/126

PR #4019, opened 2020-08-05
Add compression, pickle protocol to comm contexts

mrocklin:
Currently we expect a great degree of uniformity among servers in a Dask
cluster.  This is especially apparent with pickle protocol and
compression formats.  This PR includes a handshake when connecting that
exchanges a bit of information.  Then this information is passed to
serialization functions through the existing context= mechanism.

This allows serialization functions to have more information about the
other side of the connection, and to make choices accordingly.

This isn't yet smooth, but I thought I'd throw it up early so that
people can see what I'm thinking and comment early.

cc @quasiben @jakirkham @jrbourbeau

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2363,7 +2363,10 @@ class Client:
 
     async def _run_on_scheduler(self, function, *args, wait=True, **kwargs):
         response = await self.scheduler.run_function(
-            function=dumps(function), args=dumps(args), kwargs=dumps(kwargs), wait=wait
+            function=dumps(function, protocol=4),
+            args=dumps(args, protocol=4),
+            kwargs=dumps(kwargs, protocol=4),
+            wait=wait,
         )
         if response["status"] == "error":
             typ, exc, tb = clean_exception(**response)
@@ -2409,10 +2412,10 @@ class Client:
         responses = await self.scheduler.broadcast(
             msg=dict(
                 op="run",
-                function=dumps(function),
-                args=dumps(args),
+                function=dumps(function, protocol=4),
+                args=dumps(args, protocol=4),
                 wait=wait,
-                kwargs=dumps(kwargs),
+                kwargs=dumps(kwargs, protocol=4),
             ),
             workers=workers,
             nanny=nanny,
@@ -4084,7 +4087,7 @@ class Client:
 
     async def _register_worker_plugin(self, plugin=None, name=None):
         responses = await self.scheduler.register_worker_plugin(
-            plugin=dumps(plugin), name=name
+            plugin=dumps(plugin, protocol=4), name=name
         )
         for response in responses.values():
             if response["status"] == "error":
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1310,8 +1310,9 @@ class Client:
 
         self.status = "closing"
 
-        for pc in self._periodic_callbacks.values():
-            pc.stop()
+        with suppress(AttributeError):
+            for pc in self._periodic_callbacks.values():
+                pc.stop()
 
         with log_errors():
             _del_global_client(self)
@@ -1405,8 +1406,9 @@ class Client:
             return
         self.status = "closing"
 
-        for pc in self._periodic_callbacks.values():
-            pc.stop()
+        with suppress(AttributeError):
+            for pc in self._periodic_callbacks.values():
+                pc.stop()
 
         if self.asynchronous:
             future = self._close()
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -13,7 +13,7 @@ from ..metrics import time
 from ..utils import parse_timedelta, TimeoutError
 from . import registry
 from .addressing import parse_address
-from ..protocol.compression import default_compression
+from ..protocol.compression import get_default_compression
 from ..protocol import pickle
 
 
@@ -127,7 +127,7 @@ class Comm(ABC):
     @staticmethod
     def handshake_info():
         return {
-            "compression": default_compression,
+            "compression": get_default_compression(),
             "python": tuple(sys.version_info)[:3],
             "pickle-protocol": pickle.HIGHEST_PROTOCOL,
         }
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -281,8 +281,8 @@ async def connect(
                     )
                     local_info = {**comm.handshake_info(), **handshake_overrides}
                     try:
-                        write = await asyncio.wait_for(comm.write(local_info), 1)
                         handshake = await asyncio.wait_for(comm.read(), 1)
+                        write = await asyncio.wait_for(comm.write(local_info), 1)
                         # This would be better, but connections leak if worker is closed quickly
                         # write, handshake = await asyncio.gather(write, handshake)
                     except Exception:
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -205,10 +205,12 @@ class Listener(ABC):
     async def on_connection(self, comm: Comm, handshake_overrides={}):
         local_info = {**comm.handshake_info(), **handshake_overrides}
         try:
-            write = await asyncio.wait_for(comm.write(local_info), 1)
-            handshake = await asyncio.wait_for(comm.read(), 1)
+            # write = await asyncio.wait_for(comm.write(local_info), 1)
+            # handshake = await asyncio.wait_for(comm.read(), 1)
             # This would be better, but connections leak if worker is closed quickly
-            # write, handshake = await asyncio.gather(write, handshake)
+            write, handshake = await asyncio.wait_for(
+                asyncio.gather(comm.write(local_info), comm.read()), 1
+            )
         except Exception:
             with suppress(Exception):
                 await comm.close()
@@ -282,10 +284,12 @@ async def connect(
                     )
                     local_info = {**comm.handshake_info(), **handshake_overrides}
                     try:
-                        handshake = await asyncio.wait_for(comm.read(), 1)
-                        write = await asyncio.wait_for(comm.write(local_info), 1)
+                        # handshake = await asyncio.wait_for(comm.read(), 1)
+                        # write = await asyncio.wait_for(comm.write(local_info), 1)
                         # This would be better, but connections leak if worker is closed quickly
-                        # write, handshake = await asyncio.gather(write, handshake)
+                        write, handshake = await asyncio.wait_for(
+                            asyncio.gather(comm.write(local_info), comm.read()), 1
+                        )
                     except Exception:
                         with suppress(Exception):
                             await comm.close()
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -4,6 +4,7 @@ from contextlib import suppress
 import inspect
 import logging
 import random
+import sys
 import weakref
 
 import dask
@@ -12,6 +13,8 @@ from ..metrics import time
 from ..utils import parse_timedelta, TimeoutError
 from . import registry
 from .addressing import parse_address
+from ..protocol.compression import default_compression
+from ..protocol import pickle
 
 
 logger = logging.getLogger(__name__)
@@ -43,6 +46,9 @@ class Comm(ABC):
         self._instances.add(self)
         self.allow_offload = True  # for deserialization in utils.from_frames
         self.name = None
+        self.local_info = {}
+        self.remote_info = {}
+        self.handshake_options = {}
 
     # XXX add set_close_callback()?
 
@@ -118,6 +124,27 @@ class Comm(ABC):
         """
         return {}
 
+    @staticmethod
+    def handshake_info():
+        return {
+            "compression": default_compression,
+            "python": tuple(sys.version_info)[:3],
+            "pickle-protocol": pickle.HIGHEST_PROTOCOL,
+        }
+
+    @staticmethod
+    def handshake_configuration(local, remote):
+        out = {
+            "pickle-protocol": min(local["pickle-protocol"], remote["pickle-protocol"])
+        }
+
+        if local["compression"] == remote["compression"]:
+            out["compression"] = local["compression"]
+        else:
+            out["compression"] = None
+
+        return out
+
     def __repr__(self):
         clsname = self.__class__.__name__
         if self.closed():
@@ -174,6 +201,20 @@ class Listener(ABC):
             return self
 
         return _().__await__()
+
+    async def on_connection(self, comm: Comm):
+        write = comm.write(comm.handshake_info())
+        handshake = comm.read()
+        write, handshake = await asyncio.gather(write, handshake)
+
+        comm.remote_info = handshake
+        comm.remote_info["address"] = comm._peer_addr
+        comm.local_info = comm.handshake_info()
+        comm.local_info["address"] = comm._local_addr
+
+        comm.handshake_options = comm.handshake_configuration(
+            comm.local_info, comm.remote_info
+        )
 
 
 class Connector(ABC):
@@ -249,6 +290,19 @@ async def connect(addr, timeout=None, deserialize=True, **connection_args):
                 _raise(error)
         else:
             break
+
+    write = comm.write(comm.handshake_info())
+    handshake = comm.read()
+    write, handshake = await asyncio.gather(write, handshake)
+
+    comm.remote_info = handshake
+    comm.remote_info["address"] = comm._peer_addr
+    comm.local_info = comm.handshake_info()
+    comm.local_info["address"] = comm._local_addr
+
+    comm.handshake_options = comm.handshake_configuration(
+        comm.local_info, comm.remote_info
+    )
 
     return comm
 
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -202,8 +202,8 @@ class Listener(ABC):
 
         return _().__await__()
 
-    async def on_connection(self, comm: Comm, handshake_overrides={}):
-        local_info = {**comm.handshake_info(), **handshake_overrides}
+    async def on_connection(self, comm: Comm, handshake_overrides=None):
+        local_info = {**comm.handshake_info(), **(handshake_overrides or {})}
         try:
             write = await asyncio.wait_for(comm.write(local_info), 1)
             handshake = await asyncio.wait_for(comm.read(), 1)
@@ -236,7 +236,7 @@ class Connector(ABC):
 
 
 async def connect(
-    addr, timeout=None, deserialize=True, handshake_overrides={}, **connection_args
+    addr, timeout=None, deserialize=True, handshake_overrides=None, **connection_args
 ):
     """
     Connect to the given address (a URI such as ``tcp://127.0.0.1:1234``)
@@ -280,7 +280,10 @@ async def connect(
                     comm = await connector.connect(
                         loc, deserialize=deserialize, **connection_args
                     )
-                    local_info = {**comm.handshake_info(), **handshake_overrides}
+                    local_info = {
+                        **comm.handshake_info(),
+                        **(handshake_overrides or {}),
+                    }
                     try:
                         handshake = await asyncio.wait_for(comm.read(), 1)
                         write = await asyncio.wait_for(comm.write(local_info), 1)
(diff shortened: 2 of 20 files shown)


distributed/comm/tcp.py
```diff
@@ -552,5 +586,27 @@ class TLSBackend(BaseTCPBackend):
...
+    return {
+        "compression": default_compression,
+        "python": tuple(sys.version_info)[:3],
+    }
+
+
+def handshake_configuration(local, remote):
+    out = {}
+    if local["python"] >= (3, 8) and remote["python"] >= (3, 8):
+        out["pickle_protocol"] = 5
```
jakirkham, line 599: Would we be able to check whether `pickle5` is available?

distributed/protocol/serialize.py
```diff
@@ -54,11 +54,15 @@ def dask_loads(header, frames):
...
-def pickle_dumps(x):
+def pickle_dumps(x, context=None):
     header = {"serializer": "pickle"}
     frames = [None]
     buffer_callback = lambda f: frames.append(memoryview(f))
-    frames[0] = pickle.dumps(x, buffer_callback=buffer_callback)
+    frames[0] = pickle.dumps(
+        x,
+        buffer_callback=buffer_callback,
+        protocol=context["pickle-protocol"] if context else None,
```
jakirkham, line 64: Should this be `_` or `-`. Just noting it is `_` above

distributed/protocol/pickle.py
```diff
@@ -33,15 +33,15 @@ def _always_use_pickle_for(x):
         return False
 
 
-def dumps(x, *, buffer_callback=None):
+def dumps(x, *, buffer_callback=None, protocol=HIGHEST_PROTOCOL):
```
jakirkham, line 36: Is this suppose to be `None`? Asking as below we have `protocol or HIGHEST_PROTOCOL`, which seems unneeded if this is `HIGHEST_PROTOCOL`.

jakirkham:
Thanks Matt! 😄 Had a couple minor questions above.

mrocklin:
> Thanks Matt! smile Had a couple minor questions above.

Thanks for the careful review @jakirkham .  In general does this approach seem ok to you?  I'm looking for high level thoughts before I jump in too deeply (there are still some details to iron out before things work smoothly).

distributed/comm/tcp.py
mrocklin: Yeah, I was curious about how best to handle `pickle5`.  Good idea.  Done.

distributed/protocol/serialize.py
mrocklin: Thanks, done I think

distributed/protocol/pickle.py
mrocklin: Either one is fine with me.  We sometimes pass in `protocol=None`, hence the `or HIGHEST_PROTOCOL` bit, but I figured having the default be informative wouldn't hurt in either case.

mrocklin:
@jcrist the handshake on connect mechanism here may interest you

distributed/protocol/pickle.py
jakirkham: Yeah no strong feelings from me either. Just noting it's a little odd to have `HIGHEST_PROTOCOL or HIGHEST_PROTOCOL` potentially ;)

jakirkham:
> > Thanks Matt! smile Had a couple minor questions above.
> 
> Thanks for the careful review @jakirkham . In general does this approach seem ok to you? I'm looking for high level thoughts before I jump in too deeply (there are still some details to iron out before things work smoothly).

Yeah I think so. IIUC this is just the implementation of what was discussed in issue ( https://github.com/dask/distributed/issues/4011 ). So yeah it seems good to me.

Haven't thought too deeply about how the compression side would look. So am trusting you know what you want there ;) Happy to look at that in more detail when that makes sense :)

mrocklin:
> Haven't thought too deeply about how the compression side would look. So am trusting you know what you want there ;) Happy to look at that in more detail when that makes sense :)

Our current policy here isn't great.  It's "if both sides agree on the same default compression then compress, otherwise don't".  In the future we an pass around a set of allowable compressions and then choose the jointly maximal pair.  I figured I'd wait for later to deal with those details though.

mrocklin:
I've just pushed up a commit which moves this from the TCP comm to Comms generally.  Everything seems fine on TLS/Inproc but if someone from RAPIDS can check out UCX I would appreciate that.

jakirkham:
cc-ing @madsbk (in case you have thoughts here 🙂)

distributed/comm/tests/test_comms.py
```diff
@@ -649,7 +646,8 @@ async def handle_comm(comm):
     if os.name != "nt":
         try:
             # See https://serverfault.com/questions/793260/what-does-tlsv1-alert-unknown-ca-mean
-            assert "unknown ca" in str(excinfo.value)
+            # assert "unknown ca" in str(excinfo.value)
+            pass
```
mrocklin, line 650: This is a regression.  Our error reporting around bad TLS degraded a bit.  It now shows up as a timeout error with a more generic "could not connect" message.  I tried figuring out what was going on here for a while but couldn't come up with a resolution.  I could use help here.

jakirkham:
> I've just pushed up a commit which moves this from the TCP comm to Comms generally. Everything seems fine on TLS/Inproc but if someone from RAPIDS can check out UCX I would appreciate that.

My guess is this is fine. I'm not sure how much usage we will get out of this in UCX (we ensure `pickle5` is installed as needed and don't currently use compression). That said, it is worth thinking about where we might use this (one idea below).

We probably need to update `to_frames` and `from_frames` to take the `context` in there. Would imagine this is the same for TCP. Though I could always be missing something here.

One thing that might be interesting in the UCX context (and this can be done in another PR) is whether we want to enable/disable `"cuda"` serialization (potentially moving all frames to host) depending on where/how the data is routed. Maybe @pentschev has thoughts on this? 🙂

mrocklin:
Yeah, more generally I encourage RAPIDS folks to think of this PR as two separate features, a specific and a general one

Specific: It's now easier to handle mismatched workers/schedulers with regards to compression/Python versions (maybe not that interesting to you)

General: Comms now have a mechanism to know arbitrary information about both ends of the connection.  If there are questions that you want to know like "which GPU are you?" or "what architecture are you running?" then there is infrastructure here to collect and transfer that information as well as deliver it to any serialization function.

distributed/protocol/serialize.py
```diff
@@ -54,11 +54,15 @@ def dask_loads(header, frames):
...
+def pickle_dumps(x, context=None):
     header = {"serializer": "pickle"}
     frames = [None]
     buffer_callback = lambda f: frames.append(memoryview(f))
-    frames[0] = pickle.dumps(x, buffer_callback=buffer_callback)
+    frames[0] = pickle.dumps(
+        x,
+        buffer_callback=buffer_callback,
+        protocol=context.get("pickle-protocol", None) if context else None,
+    )
```
jakirkham, line 65: We will also want to do this [for NumPy `object` arrays]( https://github.com/dask/distributed/blob/6ee7f899d1bc686ba4910c08509c7f2e5345506d/distributed/protocol/numpy.py#L27 ).

jakirkham:
> General: Comms now have a mechanism to know arbitrary information about both ends of the connection. If there are questions that you want to know like "which GPU are you?" or "what architecture are you running?" then there is infrastructure here to collect and transfer that information as well as deliver it to any serialization function.

On this general question, I think there is another point that will come up (though again this would be future work), is can we choose where tasks goes based on this information? For example knowing a worker is "closer" and free might be a useful thing when distributing tasks and moving data, which would be helpful for RAPIDS, but could have broader value. Though there could be other cases like directing tasks based on the capabilities of the worker referenced above.

pentschev:
> One thing that might be interesting in the UCX context (and this can be done in another PR) is whether we want to enable/disable `"cuda"` serialization (potentially moving all frames to host) depending on where/how the data is routed. Maybe @pentschev has thoughts on this? 🙂

I don't really see what how moving all frames to host would benefit UCX. Apart from that, today we only use UCP layer, meaning all routing is done internally by UCX without any capability to expose that to UCX-Py, and consequently to the outside world, such as Dask. That is in itself a limiting factor for doing that.

jakirkham:
> > One thing that might be interesting in the UCX context (and this can be done in another PR) is whether we want to enable/disable `"cuda"` serialization (potentially moving all frames to host) depending on where/how the data is routed. Maybe @pentschev has thoughts on this? 🙂
> 
> I don't really see how moving all frames to host would benefit UCX. Apart from that, today we only use UCP layer, meaning all routing is done internally by UCX without any capability to expose that to UCX-Py, and consequently to the outside world, such as Dask. That is in itself a limiting factor for doing that.

Yeah I wasn't sure if this might be helpful when using IB (especially if we have a more efficient way of doing device-to-host transfers).

pentschev:
> Yeah I wasn't sure if this might be helpful when using IB (especially if we have a more efficient way of doing device-to-host transfers).

For IB this wouldn't be helpful as we rely on GPUDirectRDMA for optimal performance, without that IB performance for CUDA frames isn't that great.

mrocklin:
```python
In [1]: %%time 
   ...: from dask.distributed import Client 
   ...: client = Client("localhost:8786") 
   ...: import dask.array as da 
   ...: x = da.random.random((20000, 20000)) 
   ...: (x + x.T).sum().compute() 
   ...:  
   ...:                                                                                                                                                                                                                                                                                                                                                                                                                                
/home/mrocklin/workspace/distributed/distributed/client.py:1136: VersionMismatchWarning: Mismatched versions found

+---------+---------------+---------------+-------------------------------------+
| Package | client        | scheduler     | workers                             |
+---------+---------------+---------------+-------------------------------------+
| blosc   | 1.7.0         | 1.7.0         | None                                |
| dask    | 2.18.1        | 2.18.1        | 2.20.0                              |
| lz4     | 3.0.2         | 3.0.2         | None                                |
| msgpack | 0.6.2         | 0.6.2         | 1.0.0                               |
| numpy   | 1.17.5        | 1.17.5        | 1.19.1                              |
| python  | 3.7.6.final.0 | 3.7.6.final.0 | {'3.8.5.final.0', '3.6.10.final.0'} |
+---------+---------------+---------------+-------------------------------------+
Notes: 
-  msgpack: Variation is ok, as long as everything is above 0.6
  warnings.warn(version_module.VersionMismatchWarning(msg[0]["warning"]))
CPU times: user 1.38 s, sys: 129 ms, total: 1.51 s
Wall time: 15.4 s
Out[1]: 399987145.9078545
```

mrocklin:
OK, some issues:

1.  I had to turn off pickle5 when sending data to and from the scheduler.  Because the scheduler commonly operates as a proxy between clients and workers it's hard to predict where things will go and the capabilities on the other side.  I think that this probably isn't a *huge* deal because this is rarely where we care about hardcore performance (like worker-worker) and users shouldn't be sending that much data along this wire anyway
2.   I turned off pickle protocol 5 on all of the small serializations like types and such.  This seems more unambiguously ok to me.
3.  We no longer compress msgpack encoded messages.  These are usually small anyway

I'm still fine with these changes, but I thought I'd highlight them in case they make others nervous.

It's pretty cool to see things working in a more heterogenous context though :)

jakirkham:
I'm curious why the Scheduler ends up being a special case. Doesn't it use the same Comms as Workers do? What differs there?

mrocklin:
Because the scheduler commonly operates as a proxy between clients and workers it's hard to predict where things will go and the capabilities on the other side

jakirkham:
Fair enough. At least the usages I see here don't handle out-of-band frames, so would figure using protocol 4 instead of 5 there don't matter much.

The other use case I could see for this functionality would be tracking endianness of workers. Maybe easy to add (though we can defer this as well).

mrocklin:
> The other use case I could see for this functionality would be tracking endianness of workers. Maybe easy to add (though we can defer this as well).

Yeah, I think that this fits under the "general" part of the PR.  We can add whatever we want into the `Comm.handshake_info` method.  That data gets collected from each side and fed into the serialize functions so that we can make well-informed decisions there.  There are probably lots of different use cases. 

As an example someone a long time ago asked for basic token-based security .  That would probably be easy to add with this mechansim.  (cc @Carreau folks you know may be interested)

mrocklin:
The other challenge to this PR is that now that we're sending a couple of messages along at connection time we're uncovering some issues with our stress tests.  So far I don't think that we've identified any issues that we've created, we're just now a bit more sensitive to tests that create and destroy a lot of workers and comms, and so need to be a bit more diligent.  I'm tracking these down, but afterwards I plan to merge.

jakirkham:
SGTM. Thanks Matt! 😄

distributed/client.py
```diff
@@ -2361,7 +2363,10 @@ def get_dataset(self, name, **kwargs):
 
     async def _run_on_scheduler(self, function, *args, wait=True, **kwargs):
         response = await self.scheduler.run_function(
-            function=dumps(function), args=dumps(args), kwargs=dumps(kwargs), wait=wait
+            function=dumps(function, protocol=4),
+            args=dumps(args, protocol=4),
+            kwargs=dumps(kwargs, protocol=4),
```
jrbourbeau, line 2368: Why does `protocol=4` need to be specified here (and several other places)?

distributed/comm/core.py
```diff
@@ -175,6 +202,26 @@ async def _():
 
         return _().__await__()
 
+    async def on_connection(self, comm: Comm, handshake_overrides={}):
```
jrbourbeau, line 205: Nitpick: I'd feel a little better about initializing `handshake_overrides=None` and then assigning it to be `{}` if `handshake_overrides is None` to avoid any issues with using mutable objects as a keyword default

distributed/comm/core.py
```diff
@@ -187,7 +234,9 @@ def connect(self, address, deserialize=True):
         """
 
 
-async def connect(addr, timeout=None, deserialize=True, **connection_args):
+async def connect(
+    addr, timeout=None, deserialize=True, handshake_overrides={}, **connection_args
```
jrbourbeau, line 238: Same comment here about `handshake_overrides`

distributed/protocol/compression.py
```diff
@@ -124,16 +124,22 @@ def zstd_decompress(data):
...
     else:
-        raise ValueError(
-            "Default compression '%s' not found.\n"
-            "Choices include auto, %s"
-            % (default, ", ".join(sorted(map(str, compressions))))
-        )
+        return default_compression
+
+
+get_default_compression()
```
jrbourbeau, line 142: For my own knowledge, we're just calling `get_default_compression` here to potentially trigger the `ValueError` above?

distributed/tests/test_client.py
```diff
@@ -6137,3 +6137,22 @@ async def test_as_completed_condition_loop(c, s, a, b):
 def test_client_connectionpool_semaphore_loop(s, a, b):
     with Client(s["address"]) as c:
         assert c.rpc.semaphore._loop is c.loop.asyncio_loop
+
+
+@pytest.mark.slow
+@pytest.mark.asyncio
+async def test_mixed_compression(cleanup):
```
jrbourbeau, line 6144: Nice!

distributed/client.py
```diff
@@ -1310,8 +1310,9 @@ async def _close(self, fast=False):
 
         self.status = "closing"
 
-        for pc in self._periodic_callbacks.values():
-            pc.stop()
+        with suppress(AttributeError):
+            for pc in self._periodic_callbacks.values():
+                pc.stop()
```
jrbourbeau, line 1315: Why was this change needed?

distributed/client.py
```diff
@@ -1405,8 +1406,9 @@ def close(self, timeout=no_default):
             return
         self.status = "closing"
 
-        for pc in self._periodic_callbacks.values():
-            pc.stop()
+        with suppress(AttributeError):
+            for pc in self._periodic_callbacks.values():
+                pc.stop()
```
jrbourbeau, line 1411: Same question

distributed/comm/core.py
mrocklin: In general I agree, but in this case I think that we're safe.

distributed/protocol/compression.py
mrocklin: Yes.  I'm mostly mimicing the previous behavior in order to not have to think too much about it :)

distributed/client.py
mrocklin: That's a good question.  I'm getting tracebacks like the following, which I don't really understand.  I guess Python can delete an object while it's initializing?  Or maybe this is an at shutdown time thing.

```python
distributed.http.proxy - INFO - To route to workers diagnostics web server please install jupyter-server-proxy: python -m pip install jupyter-server-proxy
distributed.scheduler - INFO -   Scheduler at:     tcp://127.0.0.1:41303
distributed.scheduler - INFO -   dashboard at:           127.0.0.1:36750
Exception ignored in: <function Client.__del__ at 0x7f2d672f2440>
  File "/home/travis/build/dask/distributed/distributed/client.py", line 1189, in __del__
    self.close()
  File "/home/travis/build/dask/distributed/distributed/client.py", line 1413, in close
    if self.asynchronous:
  File "/home/travis/build/dask/distributed/distributed/client.py", line 802, in asynchronous
    return self._asynchronous and self.loop is IOLoop.current()
AttributeError: 'Client' object has no attribute '_asynchronous'
```

mrocklin: But they're probably not needed.  I can remove them.  I'm just trying a few things to try to placate the stress tests.

Carreau:
> As an example someone a long time ago asked for basic token-based security. That would probably be easy to add with this Mechanism. (cc @Carreau folks you know may be interested)

I can think of  @lheagy, @rcthomas with the Jupyter & Dask deployment at LBL and Nersc (and Shreyas, but I can't find his github handle), And @rpwagner (now at UCSD I believe?).

rpwagner:
> > As an example someone a long time ago asked for basic token-based security. That would probably be easy to add with this Mechanism. (cc @Carreau folks you know may be interested)
> 
> I can think of @lheagy, @rcthomas with the Jupyter & Dask deployment at LBL and Nersc (and Shreyas, but I can't find his github handle), And @rpwagner (now at UCSD I believe?).

Would this be for internal security, like for authenticating communication to and from subtasks, or external security like for public REST API? I'd be able to comment more on the public REST API case.

mrocklin:
Sorry, I was thinking of someone specific that @Carreau and I both know separately.  I didn't mean to bring other folks into this conversation.  I've since followed up with Matthias off-list.

rpwagner:
No problem. If it's something I can contribute to, let me know.

distributed/client.py
jcrist: > I guess Python can delete an object while it's initializing?

Once `__new__` is called and the object instance is created, the `__del__` is always called. So if an error occurs in `__init__` and initialization fails the `__del__` will still be called.

distributed/protocol/compression.py
jrbourbeau: +1

distributed/client.py
mrocklin: Because these pickle.dumps calls aren't regulated by the same handshake context that exists in the normal serialization process.  We don't know anything about the recipient side here, so we remain conservative.  

In practice I think that this is ok.  These arguments are typically small, and this isn't in high-performance codepaths like direct worker-worker communication during execution.

distributed/comm/core.py
mrocklin: I'm interpretting @jcrist 's :+1: as a vote towards safety.  I've gone ahead and changed these to `None`.

mrocklin:
OK, I've had to relax one test (which is pretty pedantic anyway) and I've resolved all comments.  I'm going to go ahead and merge after passed tests.

Thank you everyone for the careful review.

jakirkham:
Thanks Matt! 😄

jakirkham:
As a result of this, we need to fix the UCX tests as well. Submitted PR ( https://github.com/dask/distributed/pull/4036 ) to do just that.

distributed/comm/tests/test_comms.py
jakirkham: Filed as issue ( https://github.com/dask/distributed/issues/4084 ) for tracking and broader visibility.

PR #4022, opened 2020-08-05
Upstream Cluster.get_client method from dask-gateway

mrocklin:
Fixes https://github.com/dask/dask-gateway/issues/217

This seems harmless enough to try.  We can see if people prefer it or not.

cc @jcrist @jacobtomlinson

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -11,6 +11,7 @@ from dask.utils import format_bytes
 from .adaptive import Adaptive
 
 from ..core import Status
+from ..client import Client
 from ..utils import (
     log_errors,
     sync,
@@ -412,3 +413,21 @@ class Cluster:
     @property
     def observed(self):
         return {d["name"] for d in self.scheduler_info["workers"].values()}
+
+    def get_client(self, set_as_default=True):
+        """Get a ``Client`` for this cluster.
+
+        Returns
+        -------
+        client : dask.distributed.Client
+
+        Examples
+        --------
+        >>> client = cluster.get_client()
+        """
+        return Client(
+            self,
+            set_as_default=set_as_default,
+            asynchronous=self.asynchronous,
+            loop=self.loop,
+        )
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -1050,3 +1050,32 @@ async def test_async_with():
         assert w
 
     assert not w
+
+
+@pytest.mark.asyncio
+async def test_get_client_async(cleanup):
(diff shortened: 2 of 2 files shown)


jcrist:
The main point of `cluster.get_client` was to ensure that when `cluster.shutdown()` happens, all attached clients also close (`get_client()` attaches the client to the cluster object to be cleaned up on shutdown). Otherwise distributed ended up spamming the dask-gateway proxy until the timeout was reached trying to reconnect, which isn't terrible but also not ideal.

A different solution would be to add a registration step when wrapping a `cluster` object with a `Client` so that the cluster knows about the client. Something like:

```python
# inside `Client.__init__` or something
if hasattr(cluster, "register_client"):
    cluster.register_client(self)
```

This would allow people to keep doing `Client(cluster)`, but dask-gateway (or other projects) could still get the same benefit. 

https://github.com/dask/distributed/pull/3239 was another attempt at solving the same issue (but possibly more general, as it might also handled the case of the scheduler failure).

mrocklin:
There is presumably also some value in typing.

### Before

```python
from dask_foo import FooCluster
cluster = FooCluster()

client = Client(cluster)
```

### After

```python
from dask_foo import FooCluster
client = FooCluster().get_client()
```

I could go either way on this.  Today, I'm leaning slightly towards the latter.

mrocklin:
> A different solution would be to add a registration step when wrapping a cluster object with a Client so that the cluster knows about the client. Something like:

```python
# inside `Client.__init__` or something
if hasattr(cluster, "register_client"):
    cluster.register_client(self)
```

I'm pretty comfortable with that too.  We could even enforce it on the `Cluster` superclass to avoid the `hasattr`

jcrist:
> I could go either way on this. Today, I'm leaning slightly towards the latter.

I also could go either way on this. I appreciate not needing an extra import line, but most of our docs go the other way. I suspect we'll likely want to make both work equally.

> We could even enforce it on the Cluster superclass to avoid the hasattr

I think I shouldn't enforce it on the cluster base class (at least right away), as not all dask cluster implementations subclass from `distributed.deploy.Cluster`. Putting a default on the base class sounds good though.

jacobtomlinson:
I personally prefer this approach.

```
# inside `Client.__init__` or something
if hasattr(cluster, "register_client"):
    cluster.register_client(self)
```

I feel like cluster and client instantiation is already a little magic and I'd rather not head further in that direction.

gjoseph92:
As noted in https://github.com/coiled/feedback/issues/144, this does still feel relevant. The current syntax would still need to work, but I, for one, would like the option of less typing :)

PR #4025, opened 2020-08-06
DOC: Update PendingDeprecationWarning with correct version number.

Carreau:
It does not matter much as those warnings are silent by default,
but it might still be good to give the right version number.


mrocklin:
This seems fine to me.  The test failure seems genuine though.

PR #4030, opened 2020-08-09
[FIX #3992] [MODIFY] angle on the x-axis labels.

### LINKED ISSUE #3992: Rotate text of Bandwidth type chart
image

We should probably slightly rotate the labels on the x axis so that they pack a bit more nicely.

This requires understanding a bit of Bokeh (or looking at their docs) and modifying the BandwidthTypes class here

https://github.com/dask/distributed/blob/907452b1bc0f7876a01de5ff39cc2dc64eb3c4a4/distributed/dashboard/components/scheduler.py#L274-L322

mathdugre:
This make it easier to read when the labels are close to each other.

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -297,6 +297,7 @@ class BandwidthTypes(DashboardComponent):
                 y_range=["a", "b"],
                 **kwargs,
             )
+            fig.xaxis.major_label_orientation = -0.5
             rect = fig.rect(
                 source=self.source,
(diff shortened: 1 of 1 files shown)


mrocklin:
Thank you for the PR @mathdugre 

There are some test failures.  I'm guessing that they are unrelated.  I'm testing CI on master now over here: https://github.com/dask/distributed/pull/4031

quasiben:
Looks like things are passing now. Merging in.  Thanks @mathdugre !

PR #4034, opened 2020-08-10
WIP: Use PTDS for synchronization in UCX communication

jakirkham:
This changes the stream we synchronize on to support the per-thread default stream.

Requires `numba` version `0.50.0+` ( https://github.com/numba/numba/pull/5347 )

<br>

**Note:** This is pending other changes in RAPIDS. Just putting this up to show what changes would be needed here at this point.

<br>

cc @shwina @kkraus14 @jrhemstad (as we discussed this earlier)
cc @quasiben @pentschev @harrism (for awareness)

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -39,12 +39,10 @@ ucx_create_endpoint = None
 ucx_create_listener = None
 
 
-def synchronize_stream(stream=0):
+def synchronize_ptds():
     import numba.cuda
 
-    ctx = numba.cuda.current_context()
-    cu_stream = numba.cuda.driver.drvapi.cu_stream(stream)
-    stream = numba.cuda.driver.Stream(ctx, cu_stream, None)
+    stream = numba.cuda.per_thread_default_stream()
     stream.synchronize()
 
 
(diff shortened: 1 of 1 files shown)


distributed/comm/ucx.py
```diff
@@ -221,13 +219,13 @@ async def write(
 
                 # Send frames
 
-                # It is necessary to first synchronize the default stream before start sending
-                # We synchronize the default stream because UCX is not stream-ordered and
-                #  syncing the default stream will wait for other non-blocking CUDA streams.
+                # It is necessary to first synchronize the per-thread default stream before start sending
+                # We synchronize the per-thread default stream because UCX is not stream-ordered and
+                #  syncing the per-thread default stream will wait for other non-blocking CUDA streams.
```
harrism, line 224: It won't wait for other threads' default per-thread default streams...

jakirkham: So should we leave this as-is? Or is there something else we should be doing here?

kkraus14: I think Mark is referring to the section of `syncing the per-thread default stream will wait for other non-blocking CUDA streams` as that line is incorrect. Syncing the per-thread default stream will only wait for the current stream WITHOUT blocking work on other non-blocking CUDA streams.

jakirkham: Right, I just wasn't sure if this implied changes beyond the text in the comment.

harrism: Assuming syncing only the PTDS is correct behaviour, I think only the comment is incorrect.

kkraus14:
One concern, what if someone is doing non-synchronized work on the legacy default stream and then goes to send here? We'll synchronize only the per-thread default stream and not synchronize the legacy default stream which can leave UCX in a bad state.

jakirkham:
So what would the counterproposal be? Leave things as-is?

kkraus14:
> So what would the counterproposal be? Leave things as-is?

I think medium term is work with the community to get the `stream` included in `__cuda_array_interface__`. Short term, I'm not sure.

PR #4039, opened 2020-08-11
Always set RMM's strides in the `header`

jakirkham:
Follow-up to PR ( https://github.com/rapidsai/rmm/pull/477 )
Fixes this test failure ( https://github.com/rapidsai/ucx-py/pull/575#issuecomment-672359289 )

As newer versions of RMM have `DeviceBuffer` setting `strides` in `__cuda_array_interface__` to `None` (as they are C-contiguous) and some deserializers rely on this to be a non-`None` value, make sure to just explicitly set `strides` to a non-`None` value.


jakirkham:
Going to go ahead and merge to fix UCX-Py CI. Only affects RMM serialization.

mrocklin:
> Going to go ahead and merge to fix UCX-Py CI. Only affects RMM
serialization.

From my perspective RAPIDS folks should feel free reign to self-merge in
UCX/RMM land

PR #4042, opened 2020-08-13
Tidy up exception traceback

KrishanBhasin:
I've been having trouble scaling my LocalCluster to use all the cores on an `r5a.24xlarge` and have having lots of trouble with `CommClosedError`.

I'm still working through the causes of that, but while doing so I noticed that the error messages were being propegated as 
```
```
instead of 

```
The above exception was the direct cause of the following exception:
```

I think this PR should address that.


mrocklin:
Looks good to me.  Thanks for the fix @KrishanBhasin !

PR #4046, opened 2020-08-14
Worker dashboard clean up

jsignell:
I got rid of the size option on the CrossFilter view since it was hardcoded to not do anything and it was tricky to make it have the desired behavior. To get it working properly I think we'd want to use radius, but then the axes might not match so it gets odd.

I also moved off widgetbox since it is raising a deprecated warning. 

Before:
image
`BokehDeprecationWarning: 'WidgetBox' is deprecated and will be removed in Bokeh 3.0, use 'bokeh.models.Column' instead`

After:
image

### DIFF
--- a/distributed/dashboard/components/worker.py
+++ b/distributed/dashboard/components/worker.py
@@ -2,7 +2,7 @@ import logging
 import math
 import os
 
-from bokeh.layouts import row, column, widgetbox
+from bokeh.layouts import row, column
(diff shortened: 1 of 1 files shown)


quasiben:
If you have time can you post a screenshot of the before/after ?

quasiben:
Looks good. Sorry for the delay.  Thanks @jsignell !

PR #4047, opened 2020-08-14
Make behavior clearer for how to get worker dashboard

jsignell:
I ran into this when smoke testing the dashboard. The docs say that all worker kwargs are passed along, but that isn't quite true.

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -45,6 +45,10 @@ class LocalCluster(SpecCluster):
         'localhost:8787' or '0.0.0.0:8787'.  Defaults to ':8787'.
         Set to ``None`` to disable the dashboard.
         Use ':0' for a random port.
+    worker_dashboard_address: str
+        Address on which to listen for the Bokeh worker diagnostics server like
+        'localhost:8787' or '0.0.0.0:8787'.  Defaults to None which disables the dashboard.
+        Use ':0' for a random port.
     diagnostics_port: int
         Deprecated.  See dashboard_address.
     asynchronous: bool (False by default)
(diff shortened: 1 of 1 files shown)


distributed/deploy/local.py
```diff
@@ -133,6 +137,13 @@ def __init__(
             )
             threads_per_worker = None
 
+        if "dashboard" in worker_kwargs:
+            warnings.warn(
+                "Setting `dashboard` is discouraged. "
+                "Please set `worker_dashboard_address` or `dashboard_address` to "
+                "get best behavior."
```
mrocklin, line 144: worker_dashboard_address here is a normal kwarg, while dashboard_address would refer to the scheduler's dashboard address I think if it was used as a normal kwarg, rather than within worker_kwargs.

jsignell: Right that's why `dashboard` is potentially ambiguous. Do they mean to pass `dashboard` to `work_kwargs` or to scheduler?

mrocklin: Ah.  I understood this issue, but misunderstood your recommendation because I didn't realize that `**worker_kwargs` was splatted.  I felt like we were encouraging them to say `worker_dashboard_address=...` or `worker_kwargs={"dashboard_address": ...}`.  

I suspect that 99% of the time that this is used the user will intend to affect the scheduler dashboard.  Because of that we might start with `dashboard_address` rather than `worker_dashboard_address`, maybe calling out something like "use the dashboard_address= keyword to affect the scheduler and the worker_dashboard_address= keyword for the worker"

jsignell: Yeah that makes sense. I added more common/less common to try to nip that in the bud.

mrocklin:
Thanks @jsignell !

PR #4049, opened 2020-08-14
Match pre-commit in dask

jsignell:
This PR copies the pre-commit from dask which seems to work fine

### Motivation
After cleaning out my dev env, I was getting errors from the pre-commit in distributed. I think it has to do with my dev env being python 3.8.

The error looked like:

```
[INFO] Installing environment for https://github.com/ambv/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/home/julia/conda/envs/dask-dev/bin/python3.8', '-mvirtualenv', '/home/julia/.cache/pre-commit/repoyh1z3_q8/py_env-python3.7', '-p', 'python3.7')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'
    
stderr: (none)
Check the log at /home/julia/.cache/pre-commit/pre-commit.log
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
-  -   repo: https://github.com/python/black
+  -   repo: https://github.com/psf/black
       rev: stable
       hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3
   -   repo: https://gitlab.com/pycqa/flake8
-      rev: 3.7.9
+      rev: 3.8.3
       hooks:
       - id: flake8
-        language_version: python3.7
+        language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ repos:
       hooks:
       - id: black
         language_version: python3
+        exclude: versioneer.py
   -   repo: https://gitlab.com/pycqa/flake8
       rev: 3.8.3
       hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,11 @@
 repos:
--   repo: https://github.com/ambv/black
-    rev: stable
-    hooks:
-    - id: black
-      language_version: python3.7
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
-    hooks:
-    - id: flake8
+  -   repo: https://github.com/python/black
+      rev: stable
+      hooks:
+      - id: black
+        language_version: python3.7
+  -   repo: https://gitlab.com/pycqa/flake8
+      rev: 3.7.9
+      hooks:
+      - id: flake8
+        language_version: python3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ env:
 matrix:
   fast_finish: true
   include:
-  - os: linux
-    language: python
-    python: 3.6
   - os: osx
     env: PYTHON=3.7 TESTS=true PACKAGES="python-snappy python-blosc" TORNADO=6
     if: type != pull_request OR commit_message =~ test-osx  # Skip on PRs unless the commit message contains "test-osx"
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ matrix:
   - os: linux
     language: python
     python: 3.6
-    env: LINT=true
   - os: osx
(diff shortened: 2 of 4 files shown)


jsignell:
@jrbourbeau does this seem reasonable?

.pre-commit-config.yaml
```diff
@@ -1,10 +1,11 @@
...
-    rev: stable
-    hooks:
-    - id: black
-      language_version: python3.7
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
-    hooks:
-    - id: flake8
+  -   repo: https://github.com/python/black
+      rev: stable
```
jrbourbeau, line 3: I've been running into `black` version mismatches between local and CI dev environments here and in other projects. What do you think about pinning to a specific release of `black`?

```suggestion
      rev: 20.8b1 
```

.pre-commit-config.yaml
```diff
@@ -1,10 +1,11 @@
...
+      hooks:
+      - id: black
+        language_version: python3.7
```
jrbourbeau, line 6:
```suggestion
        language_version: python3
```

.pre-commit-config.yaml
```diff
@@ -1,10 +1,11 @@
...
+  -   repo: https://gitlab.com/pycqa/flake8
+      rev: 3.7.9
+      hooks:
+      - id: flake8
+        language_version: python3.7
```
jrbourbeau, line 11:
```suggestion
        language_version: python3
```

jrbourbeau (commented):
Thanks for the PR @jsignell! Apologies for the delayed reply. I'm still running into `RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'` when running locally in a Python 3.8 conda environment. Perhaps we should just [use `python3` for the `language_version`](https://github.com/pre-commit/pre-commit/issues/1375#issuecomment-603906811)? That works for me locally and I suspect is general enough to work for most contributors

.pre-commit-config.yaml
```diff
@@ -1,10 +1,11 @@
...
--   repo: https://github.com/ambv/black
-    rev: stable
-    hooks:
-    - id: black
-      language_version: python3.7
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
-    hooks:
-    - id: flake8
+  -   repo: https://github.com/python/black
```
jsignell, line 2:
```suggestion
  -   repo: https://github.com/psf/black
```

.pre-commit-config.yaml
jsignell: I think we should use this pre-commit stuff in ci instead as in https://github.com/dask/dask/pull/6645

.pre-commit-config.yaml
```diff
@@ -1,10 +1,11 @@
...
+      rev: stable
+      hooks:
+      - id: black
+        language_version: python3.7
+  -   repo: https://gitlab.com/pycqa/flake8
+      rev: 3.7.9
```
jsignell, line 8:
```suggestion
      rev: 3.8.3
```

.travis.yml
```diff
@@ -16,7 +16,6 @@ matrix:
   - os: linux
     language: python
     python: 3.6
```
jrbourbeau, line 18: I _think_ we can remove this job altogether as previously it was just running `flake8` and `black` (for an example build see https://travis-ci.org/github/dask/distributed/jobs/727483894)

jrbourbeau (commented):
Thanks @jsignell! Overall this looks really good to me, just left one small comment about the changes to `.travis.yml`

jsignell:
Ok I merged a non-blacked change to make sure this fails properly.

jsignell:
@jrbourbeau I did `pre-commit run --all-files` locally as recommended by the test failure:https://github.com/dask/distributed/runs/1124209877 and I got a big diff. Would you prefer that in a follow on PR?

jsignell:
I actually can't quite get the same thing to happen on master so maybe things are ok there?

jsignell:
Yeah I think https://github.com/dask/distributed/pull/4081 fixed the formatting on master so this PR should be good to go @jrbourbeau.

jrbourbeau (approved):
Woo, thanks @jsignell!

PR #4051, opened 2020-08-19
Add cluster log method

jacobtomlinson:
In many downstream cluster managers it may take time for the cluster manager to be instantiated. Often this is because the scheduler needs to be spawned on some external resource like a batch system, and the cluster cannot be created until the scheduler exists.

In `dask-cloudprovider.AzureMLCluster` we have taken the approach to print output to the user to let them know where progress is, because it can take tens of minutes depending on the complexity of your environment.

This PR adds a standardised way to log messages to the user with the `Cluster._log` method which is intended for use by subclasses. It also adds a `quiet` kwarg which can be set to `True` to disable printing. And it stores all log messages in a list so that they can be included in the `Cluster.get_logs()` method if a user wished to retrieve all logs later.

I'm not 100% set on the design for this so would love some feedback, but I think the main thing I want to achieve is to standardise the way cluster managers print output to the user and provide options to silence it and retrieve that output later.

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -1,4 +1,5 @@
 import asyncio
+import datetime
 from contextlib import suppress
 import logging
 import threading
@@ -50,12 +51,14 @@ class Cluster:
 
     _supports_scaling = True
 
-    def __init__(self, asynchronous):
+    def __init__(self, asynchronous, quiet=False):
         self.scheduler_info = {"workers": {}}
         self.periodic_callbacks = {}
         self._asynchronous = asynchronous
         self._watch_worker_status_comm = None
         self._watch_worker_status_task = None
+        self._cm_logs = []
+        self.quiet = quiet
         self.scheduler_comm = None
 
         self.status = Status.created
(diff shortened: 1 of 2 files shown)


distributed/deploy/cluster.py
```diff
@@ -50,12 +51,14 @@ class Cluster:
...
     _supports_scaling = True
 
-    def __init__(self, asynchronous):
+    def __init__(self, asynchronous, quiet=False):
         self.scheduler_info = {"workers": {}}
         self.periodic_callbacks = {}
         self._asynchronous = asynchronous
         self._watch_worker_status_comm = None
         self._watch_worker_status_task = None
+        self._cm_logs = []
```
dantheman39, line 60: Curious what "cm" stands for? Oh, I guess cluster manager logs? Maybe include a little comment above saying what the name is, or give a clearer name?

distributed/deploy/cluster.py
```diff
@@ -170,9 +173,27 @@ def sync(self, func, *args, asynchronous=None, callback_timeout=None, **kwargs):
...
+        >>> self._log("Submitted job X to batch scheduler")
+        """
+        self._cm_logs.append((datetime.datetime.now(), log))
+        if not self.quiet:
+            print(log)
+
     async def _get_logs(self, scheduler=True, workers=True):
         logs = Logs()
 
+        logs["Cluster"] = Log("\n".join(line[1] for line in self._cm_logs))
```
dantheman39, line 195: We don't want to include the timestamps in the output somehow?

distributed/deploy/cluster.py
jacobtomlinson: Yeah it was cluster manager, perhaps `_cluster_manager_logs` or just `_cluster_logs` would be better.

distributed/deploy/cluster.py
jacobtomlinson: I'm not sure. I made a point to store them, but I don't think the scheduler or worker logs include them.

jacobtomlinson:
Thanks @mrocklin. I've cleaned things up a little and updated the tests.

PR #4052, opened 2020-08-19
Add a default argument to the datasets and a possibility to override datasets

nils-braun:
Following the discussion in https://github.com/dask/dask/issues/6531, that PR adds two new features:
* if the get datasets method returns nothing, a new default keyword argument on the client can prevent the KeyError (similar to `dict.get`). I also added a `get` method to the `client.datasets`.
* A new keyword argument `override` in `publish_dataset` lets you also prevent the KeyError there on duplicated entries (server side)

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2342,24 +2342,28 @@ class Client:
         """
         return self.sync(self.scheduler.publish_list, **kwargs)
 
-    async def _get_dataset(self, name):
+    async def _get_dataset(self, name, default):
         with self.as_current():
             out = await self.scheduler.publish_get(name=name, client=self.id)
 
         if out is None:
-            raise KeyError(f"Dataset '{name}' not found")
+            if default is None:
+                raise KeyError(f"Dataset '{name}' not found")
+            else:
+                return default
         return out["data"]
 
-    def get_dataset(self, name, **kwargs):
+    def get_dataset(self, name, default=None, **kwargs):
         """
-        Get named dataset from the scheduler
+        Get named dataset from the scheduler if present.
+        Return the default or raise a KeyError if not present.
 
         See Also
         --------
         Client.publish_dataset
         Client.list_datasets
         """
-        return self.sync(self._get_dataset, name, **kwargs)
+        return self.sync(self._get_dataset, name, default=default, **kwargs)
 
     async def _run_on_scheduler(self, function, *args, wait=True, **kwargs):
         response = await self.scheduler.run_function(
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -98,6 +98,8 @@ _global_client_index = [0]
 _current_client = ContextVar("_current_client", default=None)
 
 DEFAULT_EXTENSIONS = [PubSubClientExtension]
+# Placeholder used in the get_dataset function(s)
+NO_DEFAULT_PLACEHOLDER = "_no_default_"
 
 
 def _get_global_client():
@@ -2281,7 +2283,8 @@ class Client:
         ----------
         args : list of objects to publish as name
         name : optional name of the dataset to publish
-        override : if true, override any already present dataset with the same name
+        override : bool (optional, default False)
+            if true, override any already present dataset with the same name
         kwargs: dict
             named collections to publish on the scheduler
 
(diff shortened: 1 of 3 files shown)


distributed/client.py
```diff
@@ -2280,6 +2281,7 @@ def publish_dataset(self, *args, **kwargs):
         ----------
         args : list of objects to publish as name
         name : optional name of the dataset to publish
+        override : if true, override any already present dataset with the same name
```
martindurant, line 2284: If false, will raise when the key already exists.
These argument descriptions could use a better style, like
```
override : bool (optional, default False)
    if true, override ...
```

distributed/client.py
```diff
@@ -2342,24 +2344,28 @@ def list_datasets(self, **kwargs):
...
         return self.sync(self.scheduler.publish_list, **kwargs)
 
-    async def _get_dataset(self, name):
+    async def _get_dataset(self, name, default):
         with self.as_current():
             out = await self.scheduler.publish_get(name=name, client=self.id)
 
         if out is None:
-            raise KeyError(f"Dataset '{name}' not found")
+            if default is None:
```
martindurant, line 2352: What if we want a default of `None`? Perhaps use use placeholder like `"_no_default_"`.

distributed/client.py
```diff
@@ -2342,24 +2344,28 @@ def list_datasets(self, **kwargs):
...
+            else:
+                return default
         return out["data"]
 
-    def get_dataset(self, name, **kwargs):
+    def get_dataset(self, name, default=None, **kwargs):
         """
-        Get named dataset from the scheduler
+        Get named dataset from the scheduler if present.
+        Return the default or raise a KeyError if not present.
```
martindurant, line 2361: Could do with a full update to args here too

distributed/publish.py
```diff
@@ -116,3 +116,7 @@ def __len__(self):
                 "please use 'len(await client.list_datasets())' instead"
             )
         return len(self._client.list_datasets())
+
+    def get(self, key, default=None):
```
martindurant, line 120: Why is this needed?

distributed/tests/test_publish.py
```diff
@@ -208,6 +222,18 @@ def test_datasets_contains(client):
...
+def test_datasets_republish(client):
+    key, value, value2 = "key", "value", "value2"
+    client.publish_dataset(key=value)
+    assert client.get_dataset(key) == value
+
+    with pytest.raises(KeyError) as exc_info:
+        client.publish_dataset(key=value)
+
+    client.publish_dataset(key=value2, override=True)
+    assert client.get_dataset(key) == value2
```
mrocklin, line 234: This behavior seems sensible to me

mrocklin:
This all seems good to me.  Deferring to @martindurant for a decision on the behavior

martindurant:
Fixes  dask/dask#6531

distributed/publish.py
nils-braun: Yes, you are totally correct. It will pick up the default argument anyways. Thanks for the hint!

nils-braun:
Thanks for your review @martindurant and @mrocklin. I have implemented your feedback

distributed/client.py
```diff
@@ -2342,24 +2347,36 @@ def list_datasets(self, **kwargs):
         """
         return self.sync(self.scheduler.publish_list, **kwargs)
 
-    async def _get_dataset(self, name):
+    async def _get_dataset(self, name, default=NO_DEFAULT_PLACEHOLDER):
```
martindurant, line 2350: @mrocklin , is there a utility variable or other standard way to do this?

Otherwise, happy with the PR.

nils-braun: I did a quick grep through the `distributed` repo and have found

```python
_empty_namedtuple = collections.namedtuple("_empty_namedtuple", ())
```
in `metrics.py` - but I do not know if this is the standard. Happy to follow one!

mrocklin: I am personally more familiar with using `__no_default__` over a named tuple.  I think that what is here is fine.

martindurant:
(also requires running `black`, unfortunately)

nils-braun:
Thanks @martindurant for fixing the style

PR #4054, opened 2020-08-20
Cleanup more exception tracebacks

KrishanBhasin:
I've found a few more places where the error message could be confusing for users:

e.g.

```

  File "/home/krishan/.local/lib/python3.7/site-packages/distributed/comm/core.py", line 295, in _
    raise CommClosedError()
distributed.comm.core.CommClosedError
```

I've changed in places that look like they are correct, but I have not personally encountered errors for all of the diffs I have created.

If preferred, I'm happy to limit this PR to only the instances I have seen myself.

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -209,10 +209,10 @@ class Listener(ABC):
             handshake = await asyncio.wait_for(comm.read(), 1)
             # This would be better, but connections leak if worker is closed quickly
             # write, handshake = await asyncio.gather(comm.write(local_info), comm.read())
-        except Exception:
+        except Exception as e:
             with suppress(Exception):
                 await comm.close()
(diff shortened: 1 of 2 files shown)


distributed/comm/core.py
```diff
@@ -263,7 +263,7 @@ def _raise(error):
             timeout,
             error,
         )
-        raise IOError(msg)
+        raise IOError(msg) from error
```
KrishanBhasin, line 266: This is one change I have made speculatively and am conscious may not be preferred - e.g. I'm not confident I know what happens with the below snippet (I'm not sure what the value of `error` is there.

https://github.com/dask/distributed/blob/76de87bb49faf00a84947bf0b2c6708c509d89c0/distributed/comm/core.py#L312-L313

PR #4067, opened 2020-08-24
Simplify `frame_split_size`

jakirkham:
As we already check `nbytes` and `return`ed if it was `<= n`, we already know `nbytes(frame) > n`. So there is no need to check it again. Thus we get rid of this second check. Also assign some repeatedly used values to variables to make things a little easier to follow. Finally just always use a `memoryview` to split frames as it views instead of copying data and has easy access to things like `itemsize` and `nbytes`.

### DIFF
--- a/distributed/protocol/utils.py
+++ b/distributed/protocol/utils.py
@@ -28,9 +28,8 @@ def frame_split_size(frame, n=BIG_BYTES_SHARD_SIZE) -> list:
     if frame.nbytes <= n:
         return [frame]
 
-    itemsize = frame.itemsize
-    nitems = frame.nbytes // itemsize
-    items_per_shard = n // itemsize
+    nitems = frame.nbytes // frame.itemsize
+    items_per_shard = n // frame.itemsize
 
     return [frame[i : i + items_per_shard] for i in range(0, nitems, items_per_shard)]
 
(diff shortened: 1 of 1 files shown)


jakirkham:
This has a notable improvement on runtime as well.

Before:

```python
In [1]: from distributed.protocol.utils import frame_split_size                 

In [2]: b = 2 ** 30 * b"ab"                                                     

In [3]: %timeit frame_split_size(b)                                             
6.07 µs ± 32.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

After:

```python
In [1]: from distributed.protocol.utils import frame_split_size                 

In [2]: b = 2 ** 30 * b"ab"                                                     

In [3]: %timeit frame_split_size(b)                                             
4.86 µs ± 78.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

distributed/protocol/utils.py
```diff
@@ -23,21 +23,15 @@ def frame_split_size(frame, n=BIG_BYTES_SHARD_SIZE) -> list:
     >>> frame_split_size([b'12345', b'678'], n=3)  # doctest: +SKIP
     [b'123', b'45', b'678']
     """
-    if nbytes(frame) <= n:
+    frame = memoryview(frame)
```
quasiben, line 26: Are we always guaranteed that will pass ?  Previously we had a try/except

jakirkham: Yeah

The previous `try...except...` was guarding against the absence of an `itemsize` attribute. Though `memoryview` always has this. So that is not an issue.

quasiben:
Thanks @jakirkham merging in

PR #4073, opened 2020-08-25
Drop `Serialized.deserialize()` method

jakirkham:
This method was added in PR ( https://github.com/dask/distributed/pull/638 ). However it appears it wasn't used in that change and does not appear to be used today. So go ahead and drop it.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -381,12 +381,6 @@ class Serialized:
         self.header = header
         self.frames = frames
 
-    def deserialize(self):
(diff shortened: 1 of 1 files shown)


quasiben:
This looks to good me.  @mrocklin, is/was the `deserialize` func used outside of serialize.py ?

jakirkham:
FWIW I tried to look for this, but didn't find anything.

```bash
$ git grep '\.deserialize()'
```

mrocklin:
Thanks @jakirkham !

Issue #4069, opened 2020-08-24
SpecCluster + Dask JobQueue does not retire workers when scaling down.


bbockelm:
I am currently chasing what appears to be a bug in the combination of `distributed` and `dask-jobqueue`.

Particularly, I find that this piece of code almost never completes when running a dask-jobqueue `HTCondorCluster` with adaptive scaling enabled:
```
import time
def square(x):
    time.sleep(0.5)
    return x ** 2

def neg(x):
    return -x

A = client.map(square, range(1000))
B = client.map(neg, A)
total = client.submit(sum, B)
total.result()
```

What happens is that, whenever the tasks get almost finished, one of the workers is killed.  This results in lost keys, sets the progress back a few hundred tasks, and repeats indefinitely.

Putting some debugging in the decision making, it appears that the adaptive code is deciding to scale down whenever it decides there is insufficient work to support all the workers.  This results in the invocation of the following code in `SpecCluster`:

https://github.com/dask/distributed/blob/2.24.0/distributed/deploy/spec.py#L327-L329

Particularly, the list `to_close` comes from `self.workers` and appears to be integers (e.g., `0`) and invokes the `scheduler.retire_workers` method.  Adding a print statement, I see the following worker dictionary:

```
{0: <CoffeaCasaJob: status=Status.running>, 1: <CoffeaCasaJob: status=Status.running>,
 2: <CoffeaCasaJob: status=Status.running>, 3: <CoffeaCasaJob: status=Status.running>,
 4: <CoffeaCasaJob: status=Status.running>, 5: <CoffeaCasaJob: status=Status.running>}
```

However, that method filters passed workers on a different set of keys:

https://github.com/dask/distributed/blob/93a8d28933ebe4b82d715e9f6948be3c38c49953/distributed/scheduler.py#L3502

Examples from adding a print statement:

```
SortedDict({'tls://129.93.X.Y:32806': <Worker 'tls://129.93.X.Y:32806', name: htcondor--4523817.0--,
 memory: 223, processing: 46>, 'tls://129.93.X.Y:32883': <Worker 'tls://129.93.X.Y:32883', name: htcondor--4523815.0--,
 memory: 214, processing: 47>, 'tls://129.93.X.Y:32768': <Worker 'tls://129.93.X.Y:32768', name: htcondor--4523813.0--,
 memory: 740, processing: 33>, 'tls://129.93.X.Y:32769': <Worker 'tls://129.93.X.Y:32769', name: htcondor--4523814.0--,
[...]

PR #4074, opened 2020-08-25
Retire workers in SpecCluster by name

jrbourbeau:
Currently there's a mismatch between how workers are stored on the scheduler and spec cluster. `Scheduler.workers` uses worker addresses for keys while `SpecCluster.workers` uses the output of `SpecCluster._new_worker_name` for keys (which is often an integer that's incremented when a new worker is added). This mismatch results in workers not being retired properly here (xref https://github.com/dask/distributed/issues/4069):

https://github.com/dask/distributed/blob/586ded30a8f15b91bd4e45e50c832ef21faea79c/distributed/deploy/spec.py#L328

This PR updates how we call `retire_workers` to use worker names instead of instead of addresses which will ensure the `Scheduler` and `SpecCluster` are on the same page about which workers should be retired. 

Note that since the key stored in `SpecCluster.workers` isn't _always_ used as the corresponding worker's name:

https://github.com/dask/distributed/blob/586ded30a8f15b91bd4e45e50c832ef21faea79c/distributed/deploy/spec.py#L343-L345

I added a new `SpecCluster._spec_name_to_worker_name` mapping which maps between the name stored in `SpecCluster.workers` and the actual name used for the worker that's created. 

Closes https://github.com/dask/distributed/issues/4069

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -247,6 +247,7 @@ class SpecCluster(Cluster):
         name=None,
     ):
         self._created = weakref.WeakSet()
+        self._spec_name_to_worker_name = {}
 
         self.scheduler_spec = copy.copy(scheduler)
         self.worker_spec = copy.copy(workers) or {}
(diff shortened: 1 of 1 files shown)


jrbourbeau:
cc @bbockelm

quasiben:
@jacobtomlinson if you have a moment would you be able to look this over ?

mrocklin:
> Note that since the key stored in SpecCluster.workers isn't always used as the corresponding worker's name:

We could also make this a requirement.  I don't recall why this distinction was made, but it might not be strictly required, and this sounds like the kind of thing that would make things more consistent.

Also, a corner case to watch out for with SpecCluster is multi-worker jobs (grep for `MultiWorker` for a test).  This comes up with dask-cuda and with dask-jobqueue.

jacobtomlinson:
This looks reasonable to me. But I want to echo what Matt said about multiple workers. A worker object in respect to `SpecCluster` may reference mutliple tightly coupled workers, like in `dask-cuda`.

When you run `dask-cuda-worker` it inspects the number of GPUs and spawns one worker per GPU. So it may not always be possible to reconsile a 1-to-1 relationship.

LunarLanding:
Hi, I'm interested in fixing this too, I updated it and added support for MultiWorkers ( https://github.com/dask/distributed/pull/6065 ), locally the only failing tests seem unrelated, at this point if I could have some guidance on how to fix them that would be great. Maybe using a stable version to base the patch on?

PR #4076, opened 2020-08-25
Provide informative error when connecting an older version of Dask

mrocklin:
This seems to be coming up with the new protocol changes.
In the future we'll know which versions are present and so will be able
to give better messages.  This should stop the bleeding short term
though.


jakirkham:
Ah good catch. So essentially mismatched Distributed version cause this. Is that right?

mrocklin:
Mismatch across the 2.22 and 2.23 barrier.  Consistently above or
consistently below should be ok.

PR #4077, opened 2020-08-25
Refactor out `deserialize_bytelist`

jakirkham:
Extracts the `deserialize_bytelist` function from `deserialize_bytes`. Under-the-hood `deserialize_bytes` does a minimal amount of prep work before calling into `deserialize_bytelist`. Thus maintaining the same code path.

This can be a useful place to hook into this part of serialization pipeline for things like in-memory compressed spilling in combination with `serialize_bytelist`. Also this avoids unnecessary copying in this case (unlike `deserialize_bytes`).

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -520,18 +520,11 @@ def serialize_bytelist(x, **kwargs):
 
     header = msgpack.dumps(header, use_bin_type=True)
     frames2 = [header, *frames]
-    frames2.insert(0, pack_frames_prelude(frames2))
     return frames2
 
 
-def serialize_bytes(x, **kwargs):
-    L = serialize_bytelist(x, **kwargs)
-    return b"".join(L)
-
-
-def deserialize_bytes(b):
-    frames = unpack_frames(b)
-    header, frames = frames[0], frames[1:]
+def deserialize_bytelist(L):
+    header, frames = L[0], L[1:]
     if header:
         header = msgpack.loads(header, raw=False, use_list=False)
     else:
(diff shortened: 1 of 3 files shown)


distributed/protocol/serialize.py
```diff
@@ -541,6 +534,21 @@ def deserialize_bytes(b):
...
 
+def serialize_bytes(x, **kwargs):
+    L = serialize_bytelist(x, **kwargs)
+    return b"".join([pack_frames_prelude(L), *L])
+
+
+def deserialize_bytes(b):
+    try:
+        b = memoryview(b)
+    except TypeError:
```
quasiben, line 545: I'm sorry, I think I've asked this before, can you remind me when you think we would hit a `TypeError`?

jakirkham: If the object is not [`bytes`-like]( https://docs.python.org/3/glossary.html#term-bytes-like-object ), this will raise. We end up performing this coercion in [`unpack_frames`]( https://github.com/dask/distributed/blob/a848ee0c20d4b3df261da562c4c66cb762d2a3ec/distributed/protocol/utils.py#L116 ) later. So this merely performs this coercion a bit sooner.

The non-exception case is for things like spilling to disk (analogous to `deserialize_bytes(serialize_bytes(obj))`) and the exception case is for things like compressed in-memory spilling (analogous to `deserialize_bytes(serialize_bytelist(obj))`). IOW both cases can rely on using `deserialize_bytes`.

quasiben: Thank you!

quasiben:
There are errors in [test_spill_to_disk](https://github.com/dask/distributed/pull/4077/checks?check_run_id=1028278817#step:8:250) which seem related

>   lengths = struct.unpack_from(f"{n_frames}{fmt}", b, fmt_size)
E   struct.error: total struct size too long

jakirkham:
Yeah getting this right is a bit fiddly based on how the code works. I'll give it another look. Should add this is why I marked this as WIP for the moment.

PR #4078, opened 2020-08-26
Exclude 'fire-and-forget' client from metrics

TomAugspurger:
Right now we include the "fire-and-forget" client in our prometheus
metrics. So after connecting a `Client(scheduler)` prometheus will report 2
clients rather than 1. I think it's best to leave that one out, to only have "user"
clients in the reported metrics.

### DIFF
--- a/distributed/http/scheduler/prometheus/__init__.py
+++ b/distributed/http/scheduler/prometheus/__init__.py
@@ -12,9 +12,6 @@ class _PrometheusCollector:
     def collect(self):
         from prometheus_client.core import GaugeMetricFamily, CounterMetricFamily
 
-        print("server", self.server)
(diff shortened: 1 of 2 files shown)


distributed/http/scheduler/prometheus/__init__.py
```diff
@@ -12,10 +12,13 @@ def __init__(self, dask_server):
     def collect(self):
         from prometheus_client.core import GaugeMetricFamily, CounterMetricFamily
 
+        print("server", self.server)
+        print("clients", self.server.clients)
+
```
mrocklin, line 17:
```suggestion
```

TomAugspurger:
Whoops, thanks for fixing those.

PR #4090, opened 2020-08-31
Support zero-worker clients

mrocklin:
Previously we used to wait until workers arrived before handing back
control.  However this stopped zero-worker clients from starting up

I don't think that we need this any more.  I'm guessing that SpecCluster
handles this now.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1028,13 +1028,6 @@ class Client:
                     **self._startup_kwargs,
                 )
 
-            # Wait for all workers to be ready
-            # XXX should be a LocalCluster method instead
-            while not self.cluster.workers or len(self.cluster.scheduler.workers) < len(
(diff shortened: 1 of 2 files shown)


quasiben:
This seems fine to me.  Tested by also scaling after client creation:


```python

In [2]: client = Client(n_workers=0)

In [3]: client
Out[3]: <Client: 'tcp://127.0.0.1:50763' processes=0 threads=0, memory=0 B>

In [4]: client.cluster
Out[4]: LocalCluster('tcp://127.0.0.1:50763', workers=0, threads=0, memory=0 B)

In [5]: client.cluster.scale(2)

In [6]: client
Out[6]: <Client: 'tcp://127.0.0.1:50763' processes=2 threads=24, memory=34.36 GB>
```

PR #4105, opened 2020-09-10
Fix registering a worker plugin with name arg

### LINKED ISSUE #4100: Worker plugin registration with name is not idempotent after cluster scaling
**What happened**:

Worker plugin registration with name is not idempotent after cluster scaling.

Worker plugin registration with name is always idempotent.

```python
import time

class FooWorkerPlugin:
    # name = "foo"  # Workaround.

    def setup(self, worker):
        if hasattr(worker, "foo"):
            raise RuntimeError(f"Worker {worker.address} already has foo!")

        worker.foo = True


cluster = LocalCluster(
    n_workers=1,
    threads_per_worker=1,
    processes=False)

client = Client(cluster)

for i in range(3):
    print("Workers:", list(cluster.scheduler.workers))

    responses = client.register_worker_plugin(
        name="foo",
        plugin=FooWorkerPlugin())

    print("Register foo plugin:", responses)

    if i == 1:
        cluster.scale(2)

    time.sleep(1)
    print()
```
```
Workers: ['inproc://192.168.10.20/34912/3']
Register foo plugin: {'inproc://192.168.10.20/34912/3': {'status': 'OK'}}

Workers: ['inproc://192.168.10.20/34912/3']
Register foo plugin: {'inproc://192.168.10.20/34912/3': {'status': 'repeat'}}

Workers: ['inproc://192.168.10.20/34912/10', 'inproc://192.168.10.20/34912/3']
  ...
RuntimeError: Worker inproc://192.168.10.20/34912/10 already has foo!
```
With the workaround:
```
Workers: ['inproc://192.168.10.20/32776/3']
Register foo plugin: {'inproc://192.168.10.20/32776/3': {'status': 'OK'}}

Workers: ['inproc://192.168.10.20/32776/3']
Register foo plugin: {'inproc://192.168.10.20/32776/3': {'status': 'repeat'}}

Workers: ['inproc://192.168.10.20/32776/10', 'inproc://192.168.10.20/32776/3']
Register foo plugin: {'inproc://192.168.10.20/32776/10': {'status': 'repeat'}, 'inproc://192.168.10.20/32776/3': {'status': 'repeat'}}
```

I propose that `Scheduler.worker_plugins` should be a dict `{name: plugin}` instead of a list of plugins. This is a change to a public API.

**Environment**:

- Dask version: 2.25.0
- Python version: 3.8.5
- Operating System: Windows

nre:
Fixes #4100.

Please say if `utils` is not the best place for `get_plugin_name`, which is used by `scheduler` and `worker`.

### DIFF
--- a/distributed/diagnostics/tests/test_worker_plugin.py
+++ b/distributed/diagnostics/tests/test_worker_plugin.py
@@ -1,6 +1,6 @@
 import pytest
 
-from distributed import Client, LocalCluster, Worker, WorkerPlugin
+from distributed import Worker, WorkerPlugin
 from distributed.utils_test import async_wait_for, gen_cluster, inc
 
 
@@ -136,8 +136,8 @@ async def test_release_dep_called(c, s, w):
     await async_wait_for(lambda: not (w.task_state or w.dep_state), timeout=10)
 
 
-@pytest.mark.asyncio
-async def test_registering_with_name_arg():
+@gen_cluster(nthreads=[("127.0.0.1", 1)], client=True)
+async def test_registering_with_name_arg(c, s, w):
     class FooWorkerPlugin:
         def setup(self, worker):
             if hasattr(worker, "foo"):
@@ -145,20 +145,10 @@ async def test_registering_with_name_arg():
 
             worker.foo = True
 
-    async with LocalCluster(
-        1,
-        scheduler_port=0,
-        processes=False,
-        silence_logs=False,
-        dashboard_address=None,
-        asynchronous=True,
-    ) as cluster, Client(cluster, asynchronous=True) as c:
-        responses = await c.register_worker_plugin(FooWorkerPlugin(), name="foo")
-        assert list(responses.values()) == [{"status": "OK"}]
+    responses = await c.register_worker_plugin(FooWorkerPlugin(), name="foo")
+    assert list(responses.values()) == [{"status": "OK"}]
 
-        cluster.scale(2)
-        await cluster
-
+    async with Worker(s.address, loop=s.loop):
         responses = await c.register_worker_plugin(FooWorkerPlugin(), name="foo")
         assert list(responses.values()) == [{"status": "repeat"}] * 2
 
--- a/distributed/diagnostics/tests/test_worker_plugin.py
+++ b/distributed/diagnostics/tests/test_worker_plugin.py
@@ -1,6 +1,6 @@
 import pytest
 
-from distributed import Worker, WorkerPlugin
+from distributed import Client, LocalCluster, Worker, WorkerPlugin
 from distributed.utils_test import async_wait_for, gen_cluster, inc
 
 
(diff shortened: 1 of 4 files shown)


distributed/diagnostics/tests/test_worker_plugin.py
```diff
@@ -136,6 +136,33 @@ async def test_release_dep_called(c, s, w):
...
+@pytest.mark.asyncio
+async def test_registering_with_name_arg():
+    class FooWorkerPlugin:
+        def setup(self, worker):
+            if hasattr(worker, "foo"):
+                raise RuntimeError(f"Worker {worker.address} already has foo!")
+
+            worker.foo = True
+
+    async with LocalCluster(
```
mrocklin, line 148: Rather than use LocalCluster I recommend using `gen_cluster`, and then following up with another worker

```python
@gen_cluster(client=True)
async def test_registering_with_name_arg(c, s, a, b):
    ...
        ...

```

See https://distributed.dask.org/en/latest/develop.html#writing-tests for more information

distributed/worker.py
```diff
@@ -2369,13 +2369,8 @@ async def plugin_add(self, comm=None, plugin=None, name=None):
...
                 plugin = pickle.loads(plugin)
-            if not name:
-                if hasattr(plugin, "name"):
-                    name = plugin.name
-                else:
-                    name = funcname(plugin) + "-" + str(uuid.uuid4())
-
-            assert name
+            if name is None:
+                name = get_plugin_name(plugin)
```
mrocklin, line 2373: It looks like the behavior here is to not overwrite the existing plugin.  I wonder if maybe we should overwrite instead?  I'm guessing that this will more often be what the user wants.

mrocklin (commented):
Thanks @nre .  I've left a couple of comments below.

distributed/diagnostics/tests/test_worker_plugin.py
nre: Thank you. I didn't know how to add a worker after `gen_cluster`.

distributed/worker.py
nre: The behaviour here is the [documented behaviour](https://docs.dask.org/en/latest/futures.html#distributed.Client.register_worker_plugin): "a plugin with that name has already registered then any future plugins will not run". This behaviour was already enforced by the worker, although it is documented for the client.

Perhaps my mistake is in the scheduler, where I do overwrite any existing plugin with the same name.

nre:
The 59 failed tests in one build are not related to this PR.

nre:
Sorry, I've changed my mind.

When `Scheduler.worker_plugins` was `[plugin]`, the problem was that it did not record a name arg.

The problem with `{name: plugin}` is that it should be identical to `Worker.plugins`.

My new version is `[{"plugin": plugin, "name": name}]`, which, like the original, is a record of calls of `Scheduler.register_worker_plugin` but now also records other args.

mrocklin:
This looks sensible to me.  Thanks @nre .  My apologies for the delay.

Issue #4097, opened 2020-09-05
Add TaskState class to worker.py


mrocklin:
Currently we record state in the Worker class with a variety of dictionaries that hold information like the function to be run, the type of the output, the status, the dependencies that we're waiting on, the number of bytes of the output, and so on.  We might consider turning all of these collections into a single collection of a `TaskState` type.

We did this transformation in the Scheduler a while ago see `scheduler.py::TaskState` and that seems to have been a good choice.  We should probably repeat it for the worker.  Maybe something like the following (incomplete)

```python
class TaskState:
    def __init__(self, key, run_spec):
        self.key = key
        self.dependencies = ()
        self.dependents = ()
        self.duration = None
        self.priority = None
        self.state = None
        self.dep_state = None
        self.who_has = None
        self.nbytes = None
        self.resource_restrictions = None
        self.actor = None
        self.exception = None
        self.traceback = None
        self.type = None
```

This has the advantage of being a good learning task

mrocklin:
@gforsyth if you wanted a warmup ...

gforsyth:
Sounds good. I'll pick this up today.

PR #4107, opened 2020-09-10
Add TaskState to worker.py

gforsyth:
This is some initial progress towards #4097 -- dependents, dependencies and dep_state are still being tracked in a single dict on the Worker class but those come next.

Still lots to do on this, so no need for anyone to review just yet.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -420,7 +420,6 @@ class Worker(ServerNode):
         self.validate = validate
 
         self._transitions = {
-            ("ready", "waiting"): self.transition_ready_waiting,
             ("waiting", "ready"): self.transition_waiting_ready,
             ("waiting", "memory"): self.transition_waiting_done,
             ("waiting", "error"): self.transition_waiting_done,
@@ -1606,25 +1605,6 @@ class Worker(ServerNode):
                 pdb.set_trace()
             raise
 
-    def transition_ready_waiting(self, ts, worker, **kwargs):
-        # TODO: do we actually want this?  or is this a bandaid for
-        # bad bookkeeping elsewhere?
-        try:
-            if self.validate:
-                assert ts.state == "ready"
-                assert ts.waiting_for_data
-                assert ts.key in self.ready
-
-            self.ready.pop(ts.key)
-            ts.state = "waiting"
-        except Exception as e:
-            logger.exception(e)
-            if LOG_PDB:
-                import pdb
-
-                pdb.set_trace()
-            raise
-
     def transition_waiting_done(self, ts, value=None):
         try:
             if self.validate:
@@ -2087,7 +2067,7 @@ class Worker(ServerNode):
                         self.transition(ts, "memory", value=data[d])
                     elif ts is None or ts.state == "executing":
                         self.release_key(d)
-                    elif ts.state != "memory":
+                    elif ts.state not in ("ready", "memory"):
                         self.transition(ts, "waiting", worker=worker, remove=not busy)
 
                     if not busy and d not in data and self.tasks[d].dependents:
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -114,7 +114,6 @@ class TaskState:
         return "<Task %r %s>" % (self.key, self.state)
 
 
-
 class Worker(ServerNode):
     """Worker node in a Dask distributed cluster
 
@@ -360,11 +359,11 @@ class Worker(ServerNode):
     ):
         self.tasks = dict()
         # self.task_state = dict()
-        #self.dep_state = dict()
-        #self.dependencies = dict()
-        #self.dependents = dict()
-        #self.waiting_for_data = dict()
-        #self.who_has = dict()
+        # self.dep_state = dict()
+        # self.dependencies = dict()
+        # self.dependents = dict()
+        # self.waiting_for_data = dict()
+        # self.who_has = dict()
         self.has_what = defaultdict(set)
         self.pending_data_per_worker = defaultdict(deque)
         self.nanny = nanny
@@ -1364,7 +1363,7 @@ class Worker(ServerNode):
             for key in list(keys):
                 self.log.append((key, "delete"))
                 # TODO: make sure this doesn't blow anything up but I think this check is redundant
-                #if key in self.tasks:
+                # if key in self.tasks:
                 self.release_key(key)
 
             logger.debug("Worker %s -- Deleted %d keys", self.name, len(keys))
@@ -1554,7 +1553,6 @@ class Worker(ServerNode):
                 pdb.set_trace()
             raise
 
-
         # NOTES
         #
         # I've combined deps and tasks and it's leading to this horrible loop
@@ -1626,7 +1624,8 @@ class Worker(ServerNode):
                 assert ts.state == "waiting"
                 assert not ts.waiting_for_data
                 assert all(
-                    dep.key in self.data or dep.key in self.actors for dep in ts.dependencies
+                    dep.key in self.data or dep.key in self.actors
+                    for dep in ts.dependencies
                 )
                 assert all(dep.state == "memory" for dep in ts.dependencies)
                 assert ts.key not in self.executing
@@ -1672,7 +1671,8 @@ class Worker(ServerNode):
                 assert ts.state in READY
                 assert ts.key not in self.ready
                 assert all(
-                    dep.key in self.data or dep.key in self.actors for dep in ts.dependencies
+                    dep.key in self.data or dep.key in self.actors
+                    for dep in ts.dependencies
                 )
 
             self.executing.add(ts.key)
@@ -1873,7 +1873,9 @@ class Worker(ServerNode):
                         worker = random.choice(local)
                     else:
                         worker = random.choice(list(workers))
-                    to_gather, total_nbytes = self.select_keys_for_gather(worker, dep.key)
+                    to_gather, total_nbytes = self.select_keys_for_gather(
+                        worker, dep.key
+                    )
                     self.comm_nbytes += total_nbytes
                     self.in_flight_workers[worker] = to_gather
                     for d in to_gather:
@@ -2020,7 +2022,6 @@ class Worker(ServerNode):
                 )
                 stop = time()
 
-
                 if response["status"] == "busy":
                     self.log.append(("busy-gather", worker, deps))
                     for ts in deps_ts:
@@ -2038,9 +2039,7 @@ class Worker(ServerNode):
                         }
                     )
 
-                total_bytes = sum(
-                    self.nbytes.get(key, 0) for key in response["data"]
-                )
+                total_bytes = sum(self.nbytes.get(key, 0) for key in response["data"])
                 duration = (stop - start) or 0.010
                 bandwidth = total_bytes / duration
                 self.incoming_transfer_log.append(
@@ -2050,8 +2049,7 @@ class Worker(ServerNode):
                         "middle": (start + stop) / 2.0 + self.scheduler_delay,
                         "duration": duration,
                         "keys": {
-                            key: self.nbytes.get(key, None)
-                            for key in response["data"]
+                            key: self.nbytes.get(key, None) for key in response["data"]
                         },
                         "total": total_bytes,
                         "bandwidth": bandwidth,
@@ -2102,9 +2100,7 @@ class Worker(ServerNode):
                     if not busy and d in data:
                         self.transition(d, "memory", value=data[d])
                     elif self.tasks[d].state != "memory":
-                        self.transition(
-                            d, "waiting", worker=worker, remove=not busy
-                        )
+                        self.transition(d, "waiting", worker=worker, remove=not busy)
 
                     if not busy and d not in data and self.tasks[d].dependents:
                         self.log.append(("missing-dep", d))
@@ -2128,7 +2124,6 @@ class Worker(ServerNode):
                     # See if anyone new has the data
                     await self.query_who_has(dep.key)
                     self.ensure_communicating()
-
 
     def bad_dep(self, dep):
         exc = ValueError("Could not find dependent %s.  Check worker logs" % str(dep))
@@ -2264,7 +2259,10 @@ class Worker(ServerNode):
             for dependency in ts.dependencies:
                 # remove task as dependent
                 dependency.dependents.discard(ts)
-                if not dependency.dependents and dependency.state in ("waiting", "flight"):
+                if not dependency.dependents and dependency.state in (
+                    "waiting",
+                    "flight",
+                ):
                     self.release_key(dependency.key)
 
             if ts.key in self.suspicious_deps:
@@ -2292,7 +2290,7 @@ class Worker(ServerNode):
             if report and ts.state in PROCESSING:  # not finished
                 # If not finished, restore TaskState to tasks dict
                 # TODO: need this?
-                #self.tasks[key] = ts
+                # self.tasks[key] = ts
                 self.batched_stream.send({"op": "release", "key": key, "cause": cause})
 
             self._notify_plugins("release_key", key, ts.state, cause, reason, report)
@@ -2335,7 +2333,9 @@ class Worker(ServerNode):
                     self.release_key(deptask.key, cause=dep.key)
 
             if report and dep.state == "memory":
-                self.batched_stream.send({"op": "release-worker-data", "keys": [dep.key]})
+                self.batched_stream.send(
+                    {"op": "release-worker-data", "keys": [dep.key]}
+                )
 
             self._notify_plugins("release_dep", dep.key, dep.state, report)
         except Exception as e:
@@ -2360,7 +2360,9 @@ class Worker(ServerNode):
                 dependency.dependents.remove(key)
                 # If the dependent is now without purpose (no dependencies), remove it
                 if not dependency.dependents:
-                    self.release_key(dependency.key, reason="All dependent keys rescinded")
+                    self.release_key(
+                        dependency.key, reason="All dependent keys rescinded"
+                    )
 
         except Exception as e:
             logger.exception(e)
@@ -3005,7 +3007,7 @@ class Worker(ServerNode):
                     # Might need better bookkeeping
                     assert dep.state == "memory" or dep.key in self.tasks
                     self.validate_key(dep.key)
-                #for depkey in ts.waiting_for_data:
+                # for depkey in ts.waiting_for_data:
                 #    assert (
                 #        depkey in self.in_flight_tasks
                 #        or depkey in self._missing_dep_flight
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1363,8 +1363,9 @@ class Worker(ServerNode):
         if keys:
             for key in list(keys):
                 self.log.append((key, "delete"))
-                if key in self.tasks:
-                    self.release_key(key)
+                # TODO: make sure this doesn't blow anything up but I think this check is redundant
+                #if key in self.tasks:
+                self.release_key(key)
 
             logger.debug("Worker %s -- Deleted %d keys", self.name, len(keys))
         return "OK"
@@ -1449,9 +1450,7 @@ class Worker(ServerNode):
                 assert workers
                 if dependency not in self.tasks:
                     self.tasks[dependency] = dep_ts = TaskState(key=dependency)
-                    # What state should be set here?
                     dep_ts.state = "waiting"
-                    #self.send_task_state_to_scheduler(dep_ts)
 
                 dep_ts = self.tasks[dependency]
                 self.log.append((dependency, "new-dep", dep_ts.state))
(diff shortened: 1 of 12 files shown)


mrocklin:
Hooray for PRs with incremental progress!

FWIW I really like the approach of starting with just the task-wise state.  That seems sanity-preserving.

mrocklin:
Checking in here @gforsyth .  How are things going?  Have you run into tricky bits that I can help with?

gforsyth:
> How are things going?

Definitely hit some tricky bits, but there's something I thought of I want to try.  
Mostly working, just some bookkeeping issues w.r.t. updating `data_needed`, `waiting_for_data`, etc. in a sane way.

I'll ping if the next few hours don't resolve things.

mrocklin:
It's cool seeing all of the activity here :)  

How are things going for you @gforsyth ?

gforsyth:
> How are things going for you @gforsyth ?

I think I'm closing in on a decent first cut of this.  I have a few more test failures I'm trying to trace out.

distributed/worker.py
```diff
@@ -1602,6 +1606,25 @@ def transition_waiting_ready(self, ts):
                 pdb.set_trace()
             raise
 
+    def transition_ready_waiting(self, ts, worker):
+        # TODO: do we actually want this?  or is this a bandaid for
+        # bad bookkeeping elsewhere?
```
mrocklin, line 1611: I'm curious.  What caused the need for this?

distributed/worker.py
```diff
@@ -1485,7 +1486,10 @@ def add_task(
             raise
 
     def transition(self, key, finish, **kwargs):
-        ts = self.tasks[key]
```
mrocklin: I'm curious, what cause the need for this?

mrocklin (commented):
Some minor comments. I realized that I had some other comments queued up here from a previous review that I neglected to send.  It looks like a few of those are obsolete now.

distributed/worker.py
gforsyth: I think this is actually a result of some bad bookkeeping, but I stuck it in there to both give me a place to set a breakpoint, but also to see where else things broke if there wasn't a `KeyError` in `self._transitions`.  It should not be required once I've sorted that out.

distributed/worker.py
```diff
@@ -1435,34 +1436,34 @@ def add_task(
...
 
             who_has = who_has or {}
-            self.waiting_for_data[ts.key] = set()
 
             #    msg["who_has"] = {
             #        dep.key: [ws.address for ws in dep.who_has] for dep in deps
             #    }
             # dep -> dependency (parent)
-            for parent, workers in who_has.items():
+            for dependency, workers in who_has.items():
```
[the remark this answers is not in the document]
gforsyth: ahh, yeah, this was a previous bit of disambiguation -- elsewhere I've stuck to being a little more verbose and saying `dependent` and `dependency`, I'll also make that change here.

distributed/worker.py
```diff
@@ -1338,7 +1339,7 @@ async def get_data(
 
     def update_data(self, comm=None, data=None, report=True, serializers=None):
         for key, value in data.items():
-            if key in self.tasks:
+            if getattr(self.tasks.get(key), "state", None) is not None:
```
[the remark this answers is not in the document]
gforsyth: This shouldn't happen anymore.  It was possible for a task to be instantiated with a default state of `None`, but I think setting it to `"waiting"` at instantiation nixes this issue.

mrocklin:
Ooh, nice green check marks :heavy_check_mark:

mrocklin:
@gforsyth are we ready to remove the WIP label from this PR?  Would it be reasonable to merge it?

gforsyth:
> @gforsyth are we ready to remove the WIP label from this PR?  Would it be reasonable to merge it?

I can remove the WIP. I think it's ready for an initial review. There are comments to clean up and some docstrings to expand a bit. 

I think there's more that _could_ be done, but probably not much more that _needs_ to be done. And given that I'm going to try out speculative task assignment it might make more sense to merge this in closer to as-is (pending review) than to get it "perfect" before tearing it apart again.

distributed/worker.py
```diff
@@ -1545,18 +1539,22 @@ def transition_dep_flight_waiting(self, dep, worker=None, remove=True):
...
 
-            del self.in_flight_tasks[dep]
-            if self.dependents[dep]:
-                self.dep_state[dep] = "memory"
-                self.put_key_in_memory(dep, value)
-                self.batched_stream.send({"op": "add-keys", "keys": [dep]})
+            del self.in_flight_tasks[ts.key]
+            if ts.dependents:
+                self.put_key_in_memory(ts, value)
+                assert ts.key in self.data
```
mrocklin, line 1550: If the value is large enough to be placed into slow storage immediately (unlikely) then this might not be free.  We might want to drop it.

distributed/worker.py
```diff
@@ -1663,52 +1634,52 @@ def transition_ready_executing(self, key):
...
             self.available_resources[resource] -= quantity
 
         if self.validate:
             assert all(v >= 0 for v in self.available_resources.values())
 
-    def transition_executing_done(self, key, value=no_value, report=True):
+    def transition_executing_done(self, ts, value=no_value, report=True):
         try:
             if self.validate:
-                assert key in self.executing or key in self.long_running
```
mrocklin, line 1680: Is it possible to reinstate this check?

distributed/worker.py
```diff
@@ -1663,52 +1634,52 @@ def transition_ready_executing(self, key):
...
+                    ts.traceback = msg["traceback"]
+                    ts.state = "error"
                     out = "error"
 
-                if key in self.dep_state:
-                    self.transition_dep(key, "memory")
+                # Don't release the dependency keys, but do remove them from `dependents`
+                for dependency in ts.dependencies:
+                    dependency.dependents.discard(ts)
+                ts.dependencies = set()
```
mrocklin, line 1679: Tiny nit, but maybe `.clear()`?

If this was in the the scheduler I would recommend this due to what I would expect to be a small performance gain.  It probably doesn't matter as much here though.

distributed/worker.py
```diff
@@ -1792,31 +1767,32 @@ def ensure_communicating(self):
...
                     self.log.append((key, "communication pass"))
                     self.data_needed.popleft()
                     changed = True
                     continue
 
-                deps = self.dependencies[key]
+                deps = ts.dependencies
                 if self.validate:
-                    assert all(dep in self.dep_state for dep in deps)
+                    assert all(hasattr(dep, "state") for dep in deps)
```
mrocklin, line 1779: I would think that TaskState objects always have a `.state` attribute

distributed/worker.py
```diff
@@ -1792,31 +1767,32 @@ def ensure_communicating(self):
...
+                if ts.state != "waiting":
                     self.log.append((key, "communication pass"))
                     self.data_needed.popleft()
                     changed = True
                     continue
 
-                deps = self.dependencies[key]
+                deps = ts.dependencies
                 if self.validate:
-                    assert all(dep in self.dep_state for dep in deps)
```
mrocklin, line 1803: Should this be `dep in self.tasks` now?

distributed/worker.py
```diff
@@ -1843,18 +1819,21 @@ def ensure_communicating(self):
...
                         worker = random.choice(list(workers))
-                    to_gather, total_nbytes = self.select_keys_for_gather(worker, dep)
+                    to_gather, total_nbytes = self.select_keys_for_gather(
+                        worker, dep.key
+                    )
                     self.comm_nbytes += total_nbytes
                     self.in_flight_workers[worker] = to_gather
                     for d in to_gather:
-                        self.transition_dep(d, "flight", worker=worker)
+                        self.transition(self.tasks.get(d), "flight", worker=worker)
```
mrocklin, line 1828: Is it safe to do `self.tasks[d]` here, or should we be concerned about `d not in self.tasks` as a possibility?

distributed/worker.py
```diff
@@ -1884,73 +1863,70 @@ def send_task_state_to_scheduler(self, key):
...
+        assert isinstance(dep, str)
         deps = {dep}
 
         total_bytes = self.nbytes[dep]
         L = self.pending_data_per_worker[worker]
 
         while L:
             d = L.popleft()
-            if self.dep_state.get(d) != "waiting":
+            if getattr(self.tasks.get(d), "state", None) != "waiting":
```
mrocklin, line 1929: Do we still need this `getattr` call or can we assume this?

Oh, or is this for if `d not in self.tasks`?

distributed/worker.py
```diff
@@ -1960,6 +1936,12 @@ def select_keys_for_gather(self, worker, dep):
         return deps, total_bytes
 
     async def gather_dep(self, worker, dep, deps, total_nbytes, cause=None):
+        """
+        worker: str
+        dep: TaskState object
+        deps: list of keys of dependencies to gather
```
mrocklin, line 1942: I support docstrings generally, but when we add them we should probably adhere to numpdoc standard.  

I know that you know this stuff better than 99.999999% of the population (probably an underestimate) but I'll do what I always do, and copy in this doc: https://docs.dask.org/en/latest/develop.html#docstrings

distributed/worker.py
```diff
@@ -1970,9 +1952,11 @@ async def gather_dep(self, worker, dep, deps, total_nbytes, cause=None):
 
                 # dep states may have changed before gather_dep runs
                 # if a dep is no longer in-flight then don't fetch it
-                deps = tuple(dep for dep in deps if self.dep_state.get(dep) == "flight")
+                deps_ts = [self.tasks[key] for key in deps]
+                deps_ts = tuple(ts for ts in deps_ts if ts.state == "flight")
+                deps = [d.key for d in deps_ts]
```
mrocklin, line 1957: This really makes you want variable scope in list comprehensions :)

distributed/worker.py
```diff
@@ -2037,9 +2022,9 @@ async def gather_dep(self, worker, dep, deps, total_nbytes, cause=None):
                 logger.exception("Worker stream died during communication: %s", worker)
                 self.log.append(("receive-dep-failed", worker))
                 for d in self.has_what.pop(worker):
-                    self.who_has[d].remove(worker)
-                    if not self.who_has[d]:
-                        del self.who_has[d]
+                    self.tasks[d].who_has.remove(worker)
+                # If the worker has died, `return` here so we don't hit `finally`
+                return
```
mrocklin, line 2027: ??  What happened here?

distributed/worker.py
```diff
@@ -2465,28 +2396,37 @@ def ensure_computing(self):
...
             while self.ready and len(self.executing) < self.nthreads:
-                _, key = heapq.heappop(self.ready)
-                if self.task_state.get(key) in READY:
+                priority, key = heapq.heappop(self.ready)
+                ts = self.tasks.get(key)
+                if ts is None:
+                    # It is possible for tasks to be released while still remaining on `ready`
+                    # The scheduler might have re-routed to a new worker and told this worker
+                    # to release.  If the task has "disappeared" just continue through the heap
+                    continue
```
mrocklin, line 2420: Thanks for taking the time to encode this explanation

distributed/worker.py
```diff
@@ -2843,86 +2790,46 @@ def _notify_plugins(self, method_name, *args, **kwargs):
...
-        try:
-            state = self.dep_state[dep]
-            if state == "waiting":
-                self.validate_dep_waiting(dep)
-            elif state == "flight":
-                self.validate_dep_flight(dep)
-            elif state == "memory":
-                self.validate_dep_memory(dep)
-            else:
-                raise ValueError("Unknown dependent state", state)
```
mrocklin, line 2925: Just checking, can you verify that these checks were moved to the other validate methods?

Most of our testing confidence comes from running these methods while operating Dask in different situations.  It's very important to us that we not lose validation here.

mrocklin (commented):
A bunch of tiny comments!

mrocklin:
Also, kudos on adding capability while dropping line-count :)

mrocklin:
(unless that line count was due to dropping validation functions, in which case you will be hunted down by future generations of Dask maintainers)

distributed/worker.py
gforsyth: Yep!

distributed/worker.py
gforsyth: Sounds good.

distributed/worker.py
gforsyth: Yes, I think so.  Thanks

distributed/worker.py
gforsyth: yeah, I think this was from an earlier debugging pass, cleaning it up now.

distributed/worker.py
gforsyth: Yep, these were notes to myself that I forgot to remove, but I'll make it a proper docstring instead (notes for my future self)

distributed/worker.py
gforsyth: YES

distributed/worker.py
gforsyth: I think it should be safe to do `self.tasks[d]` here, since the dependencies from `dep` should be created at this point.  I think I was worried about something dying in flight but that doesn't apply here.

distributed/worker.py
gforsyth: > Oh, or is this for if `d not in self.tasks`?

Yeah, I should add a comment about that

gforsyth: Or just make it more explicit in the first place

distributed/worker.py
gforsyth: Thanks for the call out.  Deps are gone now, but all of these checks (except for `flight`) are covered in `validate_task_state`.  I'm adding an additional check for `flight` now, which should then cover all of these bases.

gforsyth:
Thank you for the many small notes, @mrocklin.  It is my hope and belief that the line-count drop is mostly due to combining various validation functions, rather than removing them.  :crossed_fingers: that I am not the subject of the most dangerous game.

I think I've addressed all of your comments, but I'll take another cleaning pass now, too.

distributed/worker.py
gforsyth: I genuinely don't remember why I thought this was necessary but tests pass locally when I remove it.

distributed/worker.py
```diff
@@ -2521,58 +2477,62 @@ async def execute(self, key, report=False):
...
-                        key,
+                        ts.key,
                         self.active_threads,
                         self.active_threads_lock,
                         self.scheduler_delay,
                     ),
                 )
+                # TODO: it seems insane that this needs to be here
+                # but it is occasionally cleared by something in executor_submit
+                self.executing.add(ts.key)
```
mrocklin, line 2507: Why does this need to be here?  What happens if it is removed?

Grepping through the codebase I see three paths for this:

1.  The task has been released during execution.  Maybe the user no longer needs the result
2.  The task is long-running, such as is called with `worker_client` or `secede`
3.  The same task is run twice (??)

In some of these situations I would expect the following `ts.state not in ("executing", "long-running")` check to clean things up.

distributed/worker.py
```diff
@@ -2253,70 +2241,24 @@ def release_key(self, key, cause=None, reason=None, report=True):
                 pdb.set_trace()
             raise
 
-    def release_dep(self, dep, report=False):
```
jrbourbeau, line 2256: It looks like we've replaced `release_dep` calls with `release_keys`. What was the reasoning for this? It'll impact worker plugins

jrbourbeau (commented):
Apologies for the delayed review @gforsyth. Thanks for all your careful work here!

distributed/worker.py
```diff
@@ -706,7 +725,7 @@ def __init__(
     ##################
 
     def __repr__(self):
-        return "<%s: %r, %s, %s, stored: %d, running: %d/%d, ready: %d, comm: %d, waiting: %d>" % (
+        return "<%s: %r, %s, %s, stored: %d, running: %d/%d, ready: %d, comm: %d>" % (
```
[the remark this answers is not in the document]
gforsyth: Restoring that now, I had removed `waiting` entirely from the `Worker` class but restored it so we don't have to do a loop over all pending tasks when we update this

distributed/worker.py
```diff
@@ -91,6 +91,73 @@
...
+    * **waiting_for_data**: ``set(keys of dependencies)``
+        A dynamic verion of dependencies.  All dependencies that we still don't
+        have for a particular key.
+    * **resource_restrictions**: ``{str: number}``
+        Abstract resources required to run a task
+    * **exception**: ``str``
+        The exception caused by running a task if it erred
+    * **traceback**: ``str``
+        The exception caused by running a task if it erred
+    * **type_**: ``type``
```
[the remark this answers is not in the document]
gforsyth: Yeah, I didn't want to use the name `type`, even as an attribute.  Happy to change this if it reads weird.

distributed/worker.py
```diff
@@ -1508,35 +1506,31 @@ def transition_dep_waiting_flight(self, dep, worker=None):
...
 
-            if not self.who_has.get(dep):
-                if dep not in self._missing_dep_flight:
-                    self._missing_dep_flight.add(dep)
-                    self.loop.add_callback(self.handle_missing_dep, dep)
-            for key in self.dependents.get(dep, ()):
-                if self.task_state[key] == "waiting":
+            if not ts.who_has:
+                if ts.key not in self._missing_dep_flight:
+                    self.loop.add_callback(self.handle_missing_dep, ts)
```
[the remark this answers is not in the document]
gforsyth: Nice catch!  Thanks!

distributed/worker.py
```diff
@@ -1566,55 +1566,26 @@ def transition_dep_flight_memory(self, dep, value=None):
...
+                    for dep in ts.dependencies
                 )
-                assert key not in self.executing
-                assert key not in self.ready
+                assert all(dep.state == "memory" for dep in ts.dependencies)
+                assert ts.key not in self.executing
+                assert ts.key not in self.ready
 
-            self.waiting_for_data.pop(key, None)
+            ts.waiting_for_data = set()
```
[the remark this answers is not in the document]
gforsyth: Definitely.  I thought I'd changed all of these but missed this one.  Thanks!

distributed/worker.py
```diff
@@ -1792,31 +1770,32 @@ def ensure_communicating(self):
...
-                        if dep not in self._missing_dep_flight
+                        if dep.key not in self._missing_dep_flight
                     }
                     for dep in missing_deps2:
-                        self._missing_dep_flight.add(dep)
+                        self._missing_dep_flight.add(dep.key)
                     self.loop.add_callback(self.handle_missing_dep, *missing_deps2)
 
-                    deps = [dep for dep in deps if dep not in missing_deps]
+                    deps = [dep for dep in deps if dep.key not in missing_deps]
```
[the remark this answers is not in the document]
gforsyth: Yep, tested and confirmed this.

distributed/worker.py
gforsyth: There's no longer a distinction between a `dep` and a `Task` which is why I removed it.  But if it's going to impact the plugins I can restore it and have it wrap `release_keys` (or we can update the plugins?)

distributed/worker.py
gforsyth: Hmm, 
I had put this in because there were tests in `test_failed_worker.py` that would throw assertion errors (but not fail the test in question) but I've just removed it and can no longer get those errors to show up, so it may have been due to a problem elsewhere that is now fixed.

distributed/worker.py
mrocklin: > It'll impact worker plugins

Do we know of any worker plugins that trigger on `release_dep`?

Also, If we're jumping from version 2 to 2020 in the next release then we can also break some things.  Breaking `WorkerPlugin.release_dep` sounds ok to me.

distributed/worker.py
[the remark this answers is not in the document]
jrbourbeau: Gotcha. FWIW it reads a little weird to me, so I have a slight preference for switching to `type` instead of `type_` (similar to what we have in the [`FutureState` class](https://github.com/dask/distributed/blob/946c6a46abb04615051537f1bf39b81163aa91f0/distributed/client.py#L432))

distributed/worker.py
jrbourbeau: > Do we know of any worker plugins that trigger on release_dep?

None come to mind, but there may be user-defined plugins that use `release_dep`

> Breaking WorkerPlugin.release_dep sounds ok to me

Yeah, I think we're treating all the HLG changes + the bump to calver as a major version change, so removing `release_dep` should be okay

jrbourbeau (commented):
Thanks for the updates @gforsyth! It looks like there are a few remaining `executing` -> `executing_count` updates, but other this looks good to go

mrocklin:
Passed tests here are encouraging.  I'm going to restart Github actions CI to see if this is consistent.

jrbourbeau (approved):
Thanks @gforsyth! I'm +1 on merging the changes here. I'll leave this open for a bit to allow for last minute feedback, but then I plan on merging

PR #4110, opened 2020-09-12
Update super usage

rahulporuri:
On Python 3, `super` can be called without any arguments where as on Python 2, `super` needs to be called with the class object and an instance e.g. `super(Executor, self)`. Given that distributed does not support Python 2 anymore on master, the `super` usage can be updated.

Additionally, it looks like a few uses of `super()` had already crept into the codebase so these changes also help with consistency in the codebase.

Note that this was an automated regex-based search and replace. The regex used was - `(super)+\(+(\w+)(,)+\s(self)+\)+`. After the automated search and replace happened, each change was individually checked and committed.

See also sister PR ins dask - https://github.com/dask/dask/pull/6630

Ref : https://docs.python.org/3/library/functions.html#super

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4202,7 +4202,7 @@ class Executor(Client):
 
     def __init__(self, *args, **kwargs):
         warnings.warn("Executor has been renamed to Client")
-        super(Executor, self).__init__(*args, **kwargs)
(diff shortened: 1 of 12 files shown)


mrocklin:
Nice improvement.  Thanks for the cleanup @rahulporuri !

Merging!

PR #4111, opened 2020-09-14
Allow scheduler to preserve worker hostnames

### LINKED ISSUE #4070: Preserve hostnames in worker addresses
The scheduler internally converts an address with a hostname, e.g.,

```
tls://worker1.example.com:8786
```

into an IP address:

```
tls://129.93.2.2:8786
```

For the `tcp://` protocol, this is fine -- probably saves a few cycles in address translation.

However, for `tls://`, this drops an important piece of information: the desired hostname.  A TLS router (such as Traefik) can use the desired hostname information to internally route requests to the right location.

For example, we run a Dask cluster per user that's partially split across Kubernetes and a HTCondor cluster.  The scheduler is run behind a TLS router (Traefik) which exposes the scheduler externally.  A client request to connect to the scheduler `tls://scheduler1.example.com` preserves the hostname.  Hence, SNI (https://en.wikipedia.org/wiki/Server_Name_Indication) allows the router to proxy the client to the correct scheduler.  We do this as we don't have sufficient public IPs to allocate one to each scheduler.  [Yes, we are aware of dask-gateway; it doesn't  quite fit our needs - I don't want to derail the ticket on that though.]

Given the majority of the data comes from a distributed storage system and we aren't trying to push GB/s through the proxy, this setup with the scheduler works swimmingly well.

However, it doesn't work for the workers as the scheduler coerces the address of the worker almost immediately to an IP address.

SO:

1. Is there a good reason why this is done?  It seems purposeful; what's the advantage that using IPs conveys?
2. Would there be interest in a patch that preserves the hostname instead?  [Preserving the hostname might also convey some advantage to dual IPv6/IPv4 hosts]
3. If not, what about keeping both?  That is, address based on IP but keep around the hostname so it can be passed to the TLS layer.

bbockelm:
This provides a configuration option allowing the scheduler to preserve worker hostnames (as opposed to the current & default behavior of resolving hostnames immediately to IP address).

Preserving the hostname should convey advantages in two situations:

1. Dual-network setups where some hosts have IPv4-only or IPv6-only. This allows the host that wants to communicate with the worker to determine the best network to route over (as opposed to the scheduler).
2. Clusters that route TLS connections based on SNI.  If the hostname is kept internally, it will be pased to OpenSSL when making a connection.  This allows the OpenSSL client to embed the desired hostname in a SNI header in the TLS handshake.  If the worker is behind a TLS-router (not entirely uncommon in a Kubernetes setup), then this SNI is required to successfully route the request.

Closes #4070

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -24,6 +24,7 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []             # Run custom modules with Scheduler
     preload-argv: []        # See https://docs.dask.org/en/latest/setup/custom-startup.html
+    resolve-worker-hostname: true # Whether or not the worker hostname is resolved to an IP address at connection time
     unknown-task-duration: 500ms  # Default duration for all tasks with unknown durations ("15m", "2h")
     default-task-durations:  # How long we expect function names to run ("1h", "1s") (helps for long tasks)
       rechunk-split: 1us
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -128,6 +128,16 @@ properties:
 
               See https://docs.dask.org/en/latest/setup/custom-startup.html for more information
 
+          resolve-worker-hostname:
+            type: boolean
+            description: |
+              Whether the scheduler should resolve the worker hostname when joining.
+
+              If true, the worker's contact string is resolved to an IP address when it connects
+              (and the IP address is used to identify the worker internally).  Otherwise, the
+              resolution is only done when a socket is connected.  Disabling hostname resolution
+              may be advantageous when using TLS security.
+
           unknown-task-duration:
             type: string
             description: |
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -102,6 +102,8 @@ DEFAULT_DATA_SIZE = parse_bytes(
     dask.config.get("distributed.scheduler.default-data-size")
 )
 
+RESOLVE_WORKER_HOSTNAME = dask.config.get("distributed.scheduler.resolve-worker-hostname")
+
 DEFAULT_EXTENSIONS = [
     LockExtension,
     PublishExtension,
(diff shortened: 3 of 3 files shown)


distributed/distributed.yaml
```diff
@@ -24,6 +24,7 @@ distributed:
     pickle: True            # Is the scheduler allowed to deserialize arbitrary bytestrings
     preload: []             # Run custom modules with Scheduler
     preload-argv: []        # See https://docs.dask.org/en/latest/setup/custom-startup.html
+    resolve-worker-hostname: true # Whether or not the worker hostname is resolved to an IP address at connection time
```
mrocklin, line 27: What should the default be here?  Should we change this to false?

cc @jacobtomlinson

bbockelm: Eventually, yes, I would argue this should become `False` given the potential benefit to TLS and dual-stack deploys.  Seems prudent to ease into it though - have a release or two with this set to `True`.

mrocklin:
Also cc'ing @jcrist who might have experience here to suggest a default choice.

distributed/distributed.yaml
mrocklin: Prudence?  Bah.

jacobtomlinson: I would argue that we should default to false. @mrocklin mentioned in #4070 that this is the result of being developed on a system with a poor DNS service. In general applications shouldn't resolve and cache DNS like this, so I think making that thge default would be good. But given that this has been an issue in the past it makes sense to leave a confgurable option to turn it on.

mrocklin: Great, let's be aggressive here and change the default to false

mrocklin:
```suggestion
    resolve-worker-hostname: false # Whether or not the worker hostname is resolved to an IP address at connection time
```

oshadura:
@bbockelm I opened https://github.com/dask/distributed/pull/4938  to replace your PR since it has conflicts and it was not anymore easy to rebase on master. I took patch directly from our distributed.git fork.

PR #4112, opened 2020-09-14
Correctly tear down LoopRunner in Client

### LINKED ISSUE #4072: Client should call LoopRunner.stop()
**What happened**:
`distributed.Client` doesn't cleanup `LoopRunner`. Because of that `tornado.ioloop.IOLoop` and related threads remain running after closing the `Client`. The issue appears if run multiple clients with `asynchronous=False`. This happens because `Client` calls `LoopRunner.start()` but doesn't call `LoopRunner.stop()`.

`distributed.Client` should stop `LoopRunner` to cleanup resources.

```python
import threading
import time
from distributed.utils import LoopRunner


def main():
    threads_before = threading.enumerate()

    runner = LoopRunner(asynchronous=False)
    runner.start()

    cluster = LocalCluster(loop=runner.loop, asynchronous=False)
    client = Client(cluster, loop=runner.loop, asynchronous=False)

    # NOTE: uncomment to correctly cleanup resources.
    # client._should_close_loop = True

    client.close()
    cluster.close()

    runner.stop()

    # Wait until daemon threads finished.
    time.sleep(5)

    threads_after = threading.enumerate()

    if threads_before != threads_after:
        print('Before:', threads_before)
        print('After:', threads_after)
        raise AssertionError


if __name__ == '__main__':
    main()
```
If you set `client._should_close_loop = True` then it will call `LoopRunner.stop()`. Check https://github.com/dask/distributed/blob/master/distributed/client.py#L1435-L1436.

I use old distributed/dask version, but the latest sources has the same code in this part. `LoopRunner` counts `start()` calls, so there must be the same number start/stop calls to correctly stop IO loop.

**Environment**:

- Dask version: 1.2.2
- Python version: 2.7.17 (Anaconda)
- Operating System: Linux Mint 18.3 Sylvia
- Install method (conda, pip, source): `conda/pip`

skozlovf:
This PR partially fixes #4072 and adds a tests for the fix.

`SpecCluster` requires few changes to fully fix the issue. They will be provided in a separate PR since I'm not sure about impact on other parts.

### DIFF
--- a/distributed/tests/test_client_loop.py
+++ b/distributed/tests/test_client_loop.py
@@ -1,49 +1,34 @@
 import pytest
-import threading
-from time import sleep
 from distributed import LocalCluster, Client
 from distributed.utils import LoopRunner
 
 
-def run_client(loop):
-    with LocalCluster(loop=loop, asynchronous=False) as cluster:
-        client = Client(cluster, loop=loop, asynchronous=False)
-        sleep(1)  # Allow IO loop process events.
-        client.close()
-
-
-# Test if Client correctly tears down LoopRunner on close.
+# Test if Client stops LoopRunner on close.
 @pytest.mark.parametrize("with_own_loop", [True, False])
 def test_close_loop_sync(with_own_loop):
     loop_runner = loop = None
 
-    # Number of threads before and after the Cluster running.
-    threads_before = []
-    threads_after = []
+    # Setup simple cluster with one threaded worker.
+    # Complex setup is not required here since we test only IO loop teardown.
+    cluster_params = dict(n_workers=1, dashboard_address=None, processes=False)
 
-    for i in range(3):
-        threads_before.append(threading.enumerate())
+    loops_before = LoopRunner._all_loops.copy()
 
-        if with_own_loop:
-            loop_runner = LoopRunner(asynchronous=False)
-            loop_runner.start()
-            loop = loop_runner.loop
+    # Start own loop or use current thread's one.
+    if with_own_loop:
+        loop_runner = LoopRunner()
+        loop_runner.start()
+        loop = loop_runner.loop
 
-        run_client(loop)
+    with LocalCluster(loop=loop, **cluster_params) as cluster:
+        with Client(cluster, loop=loop) as client:
+            client.run(max, 1, 2)
 
-        if with_own_loop:
-            loop_runner.stop()
+    # own loop must be explicitly stopped.
+    if with_own_loop:
+        loop_runner.stop()
 
-        # Allow IO loop process events and wait until daemon threads finished.
-        sleep(1)
-        threads_after.append(threading.enumerate())
-
-    # TCP backend starts globally and runs few own threads on demand.
-    # 'threads_before' populates before that and contains less threads on first
-    # iteration. That's why we drop first values.
-    threads_before = threads_before[1:]
-    threads_after = threads_after[1:]
-
-    # In all other iterations there must be same number of threads before and
-    # after the Client running.
-    assert all([x == y for (x, y) in zip(threads_before, threads_after)])
+    # Internal loops registry must the same as before cluster running.
+    # This means loop runners in LocalCluster and Client correctly stopped.
+    # See LoopRunner._stop_unlocked().
+    assert loops_before == LoopRunner._all_loops
--- a/distributed/tests/test_client_loop.py
+++ b/distributed/tests/test_client_loop.py
@@ -1,0 +1,49 @@
+import pytest
+import threading
+from time import sleep
+from distributed import LocalCluster, Client
+from distributed.utils import LoopRunner
+
+
+def run_client(loop):
+    with LocalCluster(loop=loop, asynchronous=False) as cluster:
+        client = Client(cluster, loop=loop, asynchronous=False)
+        sleep(1)  # Allow IO loop process events.
+        client.close()
+
+
+# Test if Client correctly tears down LoopRunner on close.
+@pytest.mark.parametrize('with_own_loop', [True, False])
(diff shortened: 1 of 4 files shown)


mrocklin:
Thanks @skozlovf !

It looks like there is a test failure here:

```
================================== FAILURES ===================================
_____________________ test_client_doesnt_close_given_loop _____________________

loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x000002216D8A3BE0>
s = {'address': 'tcp://127.0.0.1:58987'}
a = {'address': 'tcp://127.0.0.1:58988', 'proc': <weakref at 0x000002216F8AC2C0; to 'Process' at 0x00000221704122E0>}
b = {'address': 'tcp://127.0.0.1:58995', 'proc': <weakref at 0x000002216F8AC310; to 'Process' at 0x000002216F988610>}

    def test_client_doesnt_close_given_loop(loop, s, a, b):
        with Client(s["address"], loop=loop) as c:
            assert c.submit(inc, 1).result() == 2
>       with Client(s["address"], loop=loop) as c:

distributed\tests\test_client.py:5409: 
distributed\client.py:744: in __init__
    self.start(timeout=timeout)
distributed\client.py:949: in start
```

skozlovf:
@mrocklin I'm analyzing. The change breaks some tests, because `pristine_loop()` and `loop` fixture create an `IOLoop` without starting it but they close the loop at the end. Outermost `LoopRunner` also closes same loop. Thinking how to improve.

mrocklin:
Thanks @skozlovf .  It looks like there are also some linting issues.  You might want to try installing the pre-commit hooks for Dask.

See https://docs.dask.org/en/latest/develop.html#code-formatting

distributed/tests/test_client_loop.py
```diff
@@ -0,0 +1,49 @@
...
+from time import sleep
+from distributed import LocalCluster, Client
+from distributed.utils import LoopRunner
+
+
+def run_client(loop):
+    with LocalCluster(loop=loop, asynchronous=False) as cluster:
+        client = Client(cluster, loop=loop, asynchronous=False)
+        sleep(1)  # Allow IO loop process events.
+        client.close()
```
mrocklin, line 12: Is there a reason why this is pulled out into a separate function?  If there isn't a specific need for this I would prefer that this be included directly as normal code.

Also, you can drop the `asynchronous=False` keywords (these are normal) but I would prefer using `with Client` context managers.

`sleep(1)` makes me nervous.  Is there a specific event that you're waiting for?  Would it be enough to run a small computation like `await client.submit(inc, 1)`.  This would both be faster and also ensure that something happened.

Can we add `processes=True` to the `LocalCluster` call?  This should shave off a couple of seconds from the time of this test (we try to keep tests fast when possible).

distributed/tests/test_client_loop.py
```diff
@@ -0,0 +1,49 @@
...
+            loop = loop_runner.loop
+
+        run_client(loop)
+
+        if with_own_loop:
+            loop_runner.stop()
+
+        # Allow IO loop process events and wait until daemon threads finished.
+        sleep(1)
+        threads_after.append(threading.enumerate())
```
mrocklin, line 39: It looks like as-is this test takes something like 10s.  Is there a way to reduce this?  Most tests that we run take less than a second.  We've found that as the test suite gets longer people run it less while developing.

distributed/tests/test_client_loop.py
skozlovf: > Is there a reason why this is pulled out into a separate function? If there isn't a specific need for this I would prefer that this be included directly as normal code.

I used parts from py2 tests where it's in a separate function, will move to test body.

> Also, you can drop the asynchronous=False keywords (these are normal) but I would prefer using with Client context managers.

Yep, also there could be two tests: context manager and explicit, to check related methods. I'll change to use context manager to be more compact. Later will add explicit `close()` if everything will be fine.

> sleep(1) makes me nervous. Is there a specific event that you're waiting for? Would it be enough to run a small computation like await client.submit(inc, 1). This would both be faster and also ensure that something happened.

I used it to give some time to IO loop, probably better to make the test async and use `await` as you suggested.

> Can we add processes=True to the LocalCluster call?

I'll try.

Seems my changes still break some tests. I created py3.6 conda env and ran tests there. One test failed: `test_dask_worker.py:test_nanny_worker_port_range_too_many_workers_raises`. There are more test failed in CI. Not sure why it happened.

PS. I tried to run tests separately and found that if run in a different order, previously passed tests will fail:

```
# Ok
$ py.test -m "not avoid_travis" -r s --timeout-method=thread --timeout=300 --durations=20 distributed/cli/tests/test_dask_scheduler.py distributed/deploy/tests/test_ssh.py -k "test_basic or test_defaults"

# Fail
$ py.test -m "not avoid_travis" -r s --timeout-method=thread --timeout=300 --durations=20 distributed/deploy/tests/test_ssh.py distributed/cli/tests/test_dask_scheduler.py -k "test_basic or test_defaults"
```

Tested my changes and 'master`  in py 3.6/3.8 (miniconda)

distributed/tests/test_client_loop.py
skozlovf: I'll improve, seems better to to replace `sleep`-way with  async test.

martindurant:
Ping @mrocklin , the OP seems t have tried to address your concerns

mrocklin:
Thanks for the ping @martindurant .  I think that I intended to merge this last week, but decided to wait until the release.  Then I promptly forgot about it.

@skozlovf thank you so much for identifying this issue, and for providing a fix.  I'm going to go ahead and merge this now.

PR #4116, opened 2020-09-18
Return right away in Cluster.close if cluster is already closed

tomzx:
If the cluster encountered an exception for example it may already
be closed. Calling close again should not at this point do anything.

I have a test suite (other project) where I call `cluster.close()` in an object that contains both a client and a LocalCluster.

Here's a simplified version.

```
class Scheduler:
    def __init__():
        self.cluster = None
        self.client = None

    def __del__():
        if self.cluster is not None:
            self.cluster.close()

        if self.client is not None:
            self.client.close()

    def submit(....):
        if self.cluster is None:
            self.cluster = LocalCluster()
            self.client = Client(self.cluster)
```

While this code has worked great, for some reason after an exception has been raised and we should have excited the program, the `self.cluster.close()` call is stuck in the `while not e.is_set():` loop of `distributed.utils.sync`.

One simple fix appears to be to not go through the async system when calling `self.cluster.close()` if the cluster is already closed due to the exception. I'm open to receiving more guidance on how to write a test case that would identify such case if it is pertinent.

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -91,6 +91,15 @@ class Cluster:
         self.status = Status.closed
 
     def close(self, timeout=None):
+        # If the cluster is already closed, we're already done
+        if self.status == Status.closed:
+            if self.asynchronous:
+                future = asyncio.Future()
+                future.set_result(None)
+                return future
+            else:
+                return
+
         with suppress(RuntimeError):  # loop closed during process shutdown
             return self.sync(self._close, callback_timeout=timeout)
 


mrocklin:
> While this code has worked great, for some reasone after an exception has been raised and we should have excited the program, the self.cluster.close() call is stuck in the while not e.is_set(): loop of distributed.utils.sync.

This may also be improved by https://github.com/dask/distributed/pull/4096/ , which merged recently.

> One simple fix appears to not go through the async system when calling self.cluster.close() if the cluster is already closed due to the exception. I'm open to receiving more guidance on how to write a test case that would identify such case if it is pertinent.

That's fine, but if we're in a synchronous mode then we shouldn't return a future at all.  We'll have to be  a little bit more careful here.

tomzx:
> That's fine, but if we're in a synchronous mode then we shouldn't return a future at all. We'll have to be a little bit more careful here.

True. My initial implementation didn't have the future but it of course failed to pass a few tests that were expecting the method to be asynchronous. I've updated the code to only return a `Future` if the cluster was started as `asynchronous`.

Another possibly (?) related PR would be https://github.com/dask/distributed/pull/3162 since I also observed that the clusters would stay up even though the object containing them was out of scope and the clusters would live until the end of the program, at which point they'd be reaped by the `close_clusters` in `distributed/deploy/spec.py`.

distributed/deploy/cluster.py
```diff
@@ -91,6 +91,15 @@ async def _close(self):
         self.status = Status.closed
 
     def close(self, timeout=None):
+        # If the cluster is already closed, we're already done
+        if self.status == Status.closed:
+            if self.asynchronous:
```
martindurant, line 96: Why is this section required? It seems like it could use testing.

mrocklin: @martindurant this is covered in the discussion above.  `Cluster.close` is used in both a sync and an async manner.  If it is asynchronous then it should return an awaitable future.  If it is synchronous then it can return None.

mrocklin:
This is in.  Thanks @tomzx !

PR #4122, opened 2020-09-22
Fix SSL connection_args for progressbar connect

jennalc:
Fix the connection_args for the client to work with SSL


mrocklin:
This looks great.  Thanks @jennalc !

PR #4123, opened 2020-09-23
Stop writing config files by default

mrocklin:
Follows on #6647 upstream


quasiben:
There was a small issue with CI and I restarted the job.  Should we merge when this passes ?

PR #4125, opened 2020-09-24
Cleanup new exception traceback

KrishanBhasin:
I was browsing the latest changelog and stumbled across this tiny change, as per my most recent PRs

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -145,7 +145,7 @@ class Comm(ABC):
                 "Your Dask versions may not be in sync. "
                 "Please ensure that you have the same version of dask "
                 "and distributed on your client, scheduler, and worker machines"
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @KrishanBhasin .  I'll admit to not knowing much about when this change would or would not be preferred. Can I ask you to explain the motivation a little?

KrishanBhasin:
Hi Matt, this is very similar to a previous PR I have raised at https://github.com/dask/distributed/pull/4042

When the error is raised, the message will be slightly clearer for the user:

```The above exception was the direct cause of the following exception:```

instead of

```During handling of the above exception, another exception occurred:```

mrocklin:
Thanks @KrishanBhasin !

Issue #4080, opened 2020-08-27
Timed out trying to connect ... : connect() didn't finish in time


michaelnarodovitch:
Whereas the following is not a bug we can pinpoint, it really bothers us. After weeks of experimentation and investigation we didn't find a real solution.

We run something like
```python
(a_bag
  .repartition(100_000) # outputs small parttitions
  .map(very_cpu_heavy_computation) # outputs ~4MB partitions
  .repartition(1000).to_dataframe().to_parquet() # writes ~30MB parquet files from ~400MB in-memory-partitions
)
```
on 100 workers with 1500 cpus on Kubernetes pods.

When it comes to the last re-partition step, dask starts blocking the event loop and the workers start spamming the following logs
```
distributed.utils_comm - INFO - Retrying get_data_from_worker after exception in attempt 0/2: Timed out trying to connect to 'tcp://10.2.3.20:38449' after 20 s: Timed out trying to connect to 'tcp://10.2.3.20:38449' after 20 s: connect() didn't finish in time
```
=> 5000 events

```
distributed.core - INFO - Event loop was unresponsive in Worker for 4.95s. This is often caused by long-running GIL-holding functions or moving large chunks of data. This can cause timeouts and instability.
```
=> 26,094 events

We are working around that, by setting
```
# less connections to avoid potential network saturation and connection timeouts
export DASK_DISTRIBUTED__WORKER__CONNECTIONS__OUTGOING=25
export DASK_DISTRIBUTED__WORKER__CONNECTIONS__INCOMING=5
export DASK_DISTRIBUTED__COMM__SOCKET_BACKLOG=16384

# graceful timeout and retry policy
export DASK_DISTRIBUTED__SCHEDULER__ALLOWED_FAILURES=20
export DASK_DISTRIBUTED__COMM__TIMEOUTS__CONNECT=20
export DASK_DISTRIBUTED__COMM__RETRY__COUNT=2
```
sothat, the job finally finishes, despite the issues in the communication.

Anyways, sometimes workers even start hanging and have to be restarted manually. Also the progress significantly slows down to the end of the job.

I'm sharing this with the community and hoping, that somebody may give pointers what to try, or even ideas for a resolution.

**Environment**:

- Dask version: 2.23.0
- Python version: 3.6
- Operating System: ubuntu on kubernetes
- Install method (conda, pip, source): pip

michaelnarodovitch:
I have the suspicion, that this is related to the handshake, which is failing, when the event loop is blocked
[...]

Issue #2519, opened 2019-02-09
version 1.25.3 in docker env stuck in "Waiting to connect to: ..."


haraldschilly:
I've a situation, where I'm basically running the scheduler and workers in a docker environment (it's a project at https://cocalc.com ). Version 1.25.2 it works fine, but starting with 1.25.3 I only get
```
distributed.worker - INFO - Waiting to connect to:       tcp://localhost:8786
```
and well, it never does. There are only [few changes](https://github.com/dask/distributed/compare/1.25.2...1.25.3), though… so maybe this help to narrow down this regression.

scheduler started via `dask-scheduler` and worker via `dask-worker tcp://localhost:8786 --local-directory ~/dask-worker --nthreads 1 --nprocs 1 --memory-limit 256M`

mrocklin:
Interesting.  I just went through all of those changes (thanks for the convenient link) and there really wasn't much that was changed in how things connect.  I don't immediately see anything that would explain the change in behavior that you're seeing.  

The next thing I recommend trying would be using `git bisect` to better identify the cause of the change in behavior.

mrocklin:
See also https://github.com/dask/distributed/issues/2506

also cc'ing @danpf if you have any suggestions here it looks like the previous work is affecting a lot of people.  Do you have any thoughts on how to address this quickly?

haraldschilly:
Ha! Thank you for pointing out git bisect, I haven't used it in at least a year. It's the first commit after 1.25.2, where it no longer connects for me: 70c5129428e0352d322c74dcc83a0c09fcd9ce39

haraldschilly:
ok, after reading #2506 I think we can close my ticket here. I am just confirming it independently.

danpf:
I know you're using docker so it might be a little hard to get at, but would you be able to try:

upping this number to something like 3, 6, or 10?
[70c5129#diff-c853b5c26508bfb36840c2e32e303886R326](https://github.com/dask/distributed/commit/70c5129428e0352d322c74dcc83a0c09fcd9ce39#diff-c853b5c26508bfb36840c2e32e303886R326)

haraldschilly:
No, that's not a problem, I work interactively inside of it. I changed it to 10 and 100 (based on 1.25.3) and it still doesn't connect. So, just upping the thread pool doesn't work.

haraldschilly:
[...]

Issue #2506, opened 2019-02-05
Issues starting workers 


bpmweel:
I'm trying to start a dask cluster on my local machine. I can use the LocalCluster just fine, but starting the scheduler and client from the command line I run into issues:

```
$ dask-scheduler
distributed.scheduler - INFO - -----------------------------------------------
distributed.scheduler - INFO - Clear task state
distributed.scheduler - INFO -   Scheduler at:  tcp://145.90.225.10:8786
distributed.scheduler - INFO -       bokeh at:                     :8787
distributed.scheduler - INFO - Local Directory: /var/folders/h6/ck24x_854wd94jzwy9r7gvl40000gn/T/scheduler-gjtgzvz2
distributed.scheduler - INFO - -----------------------------------------------
```

```
$ dask-worker 145.90.225.10:8786
distributed.nanny - INFO -         Start Nanny at: 'tcp://145.90.225.10:51273'
distributed.worker - INFO -       Start worker at:  tcp://145.90.225.10:51274
distributed.worker - INFO -          Listening to:  tcp://145.90.225.10:51274
distributed.worker - INFO -              nanny at:        145.90.225.10:51273
distributed.worker - INFO -              bokeh at:        145.90.225.10:51275
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -               Threads:                          8
distributed.worker - INFO -                Memory:                   17.18 GB
distributed.worker - INFO -       Local Directory: /Users/bweel/worker-73osop8n
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
distributed.worker - INFO - Waiting to connect to:   tcp://145.90.225.10:8786
...etc
```

I have the feeling this has to do with having multiple interfaces and the nanny process. The following combination works on the command line:

```
$ dask-scheduler --interface en0
distributed.scheduler - INFO - -----------------------------------------------
[...]

Issue #1704, opened 2018-01-23
BufferError:  - ERROR - Existing exports of data: object cannot be re-sized


softmanu:
I am running a dask-scheduler on one node and my dask-worker is running on another node.. And I submit a task to the dask-scheduler from a third node.

it sometimes throws the below error. I am using python 2.7, tornado 4.5.2, tensorflow 1.3.0

Following is the minimal script that can be used to reproduce the mentioned error which appears more often than not.

```
import os, sys
import subprocess

from dask.distributed import Variable, Client
#import psutil

import time, json, shlex

## the following function/task will be executed on the dask-worker node running on a separate node in the cluster.
def my_task(stop, is_alive):
  proc = None
  proc_started = False
  try:
    while(True):
      if stop.get():
        proc.terminate()
        return
      else:
        if not proc_started:
        
          ### The following train_image_classifier script which does the training on the set of images for classification,
          ### which is running smoothly when executing as a standalone script.
          
          ###starting a child process where the train_image_classifier i.e. training will run.
          proc = subprocess.Popen("python train_image_classifier.py") 
      is_alive.set(proc.poll())
  except:
    #####
    is_alive.set(proc.poll())
  finally:
    is_alive.set(proc.poll())
    return

### the following or this script will basically launch the task to run on the dask-worker through dask-scheduler
### dask-client script which will submit the task on dask-worker and listens the running status of the task and will send stop signal to dask-worker to stop the live task.
if __name__ == '__main__':

  client = Client("198.152.1.2:8786") # creating a dask client

  ### these two distributed variables will be used for two way communication between dask-client and the dask-worker
  stop = Variable("stop_", client = client)
  is_alive = Variable("is_alive_", client = client)

  future = client.submit(my_task)

  ###polling whether the running task is alive or not.
  ###if not then send the stop signal and come out of the execution
  while(True):
    if is_alive.get():
      stop.set(True)
      break;
  print("Execution over.! returning back to the caller..")
```

And here is the error description from the trace. it appears during the execution of the training process. it sometimes appear or sometimes not, letting the training completes.. 

**distributed.utils - ERROR - Existing exports of data: object cannot be re-sized
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/distributed/utils.py", line 238, in f
    result[0] = yield make_coro()**
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
[...]

PR #4128, opened 2020-09-25
Fixed stuck BatchedSend comm

TomAugspurger:
As reported in https://github.com/pangeo-data/pangeo/issues/788, users
were seeing tasks just not completing. After some debugging, I
discovered that the scheduler had assigned the "stuck" tasks to a worker,
but the worker never received the message. A bit more digging showed
that

1. The message was stuck in the worker BatchedSend comm's buffer
2. The BatchedSend.waker event was clear (awaiting it would wait)
3. The BatchedSend.next_deadline was set

I couldn't determine *why*, but this state is consistent with us
"missing" a deadline, i.e. the `BatchedSend.next_deadline` is set, but
the `_background_send` is already `awaiting` the event (maybe with no timeout?).
So I'm very shaky on the cause, but I'm hoping that this fixes the issue. Doing
some more manual testing.

@mrocklin I'm extra confused here, since this code hasn't been touched in a while. I'd have thought we would have seen more reports of this, but I don't recall any.

### DIFF
--- a/distributed/batched.py
+++ b/distributed/batched.py
@@ -119,7 +119,7 @@ class BatchedSend:
         self.message_count += 1
         self.buffer.append(msg)
         # Avoid spurious wakeups if possible
-        if self.next_deadline is None:
+        if self.next_deadline is None or self.next_deadline < self.loop.time():
             self.waker.set()
 
     @gen.coroutine
--- a/distributed/tests/test_batched.py
+++ b/distributed/tests/test_batched.py
@@ -253,3 +253,20 @@ async def test_serializers():
 
         with pytest.raises(TimeoutError):
             msg = await asyncio.wait_for(comm.read(), 0.1)
+
+
+@pytest.mark.asyncio
+async def test_missed_deadline():
+    async with EchoServer() as e:
+        comm = await connect(e.address)
+        b = BatchedSend(interval=10)
+        b.start(comm)
+
+        await asyncio.sleep(0.020)
+        # We've missed the deadline.
+        b.next_deadline = b.loop.time() - b.interval
+        b.send("hello")
+        b.send("world")
+        await asyncio.sleep(0.020)
+        assert len(b.buffer) == 0
+        await b.close()


TomAugspurger:
Oh, I should have looked at the logs sooner. From the scheduler pod

```
dask_gateway.dask_cli - INFO - Requesting scale to 30 workers from 1
distributed.batched - ERROR - Error in batched write
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/distributed/batched.py", line 93, in _background_send
    payload, serializers=self.serializers, on_error="raise"
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/distributed/comm/tcp.py", line 241, in write
    stream.write(b)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/tornado/iostream.py", line 553, in write
    self._write_buffer.append(data)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/tornado/iostream.py", line 177, in append
    b += data  # type: ignore
BufferError: Existing exports of data: object cannot be re-sized
```

which does sound familiar (https://github.com/dask/distributed/issues/2519, https://github.com/dask/distributed/issues/1704, https://github.com/dask/distributed/issues/2506). This is with tornado 6.0.4

TomAugspurger:
John's been looked into this a bit a https://github.com/dask/distributed/issues/4080#issuecomment-689285147.

mrocklin:
OK, so things failed in the batched_send, so the next send didn't trigger?  We should be robust to this, which I take it this PR helps with, but I honestly wouldn't expect things to finish if the error in #4080 persists .  We're not robust to dropped messages.  

One thing to check here is if an older version of distributed has this problem, ideally something just before all of the serialization / bytes / memoryviews changes that went in a few months ago.  Git bisect might help highlight an issue here.

mrocklin:
(if it's easy to reproduce that is)

TomAugspurger:
> OK, so things failed in the batched_send, so the next send didn't trigger?

I think (but I'm pretty hazy here) the rough sequence was

```python
        while not self.please_stop:
            try:
                nbytes = yield self.comm.write(
                    payload, serializers=self.serializers, on_error="raise"  # this raised a BufferError
                )
             ....
            except CommClosedError as e:
                logger.info("Batched Comm Closed: %s", e)
                break
            except Exception:
                logger.exception("Error in batched write")
                break   # <------------ so we break out of the for loop
        # and set self.stopped
        self.stopped.set()

```


>  We should be robust to this, which I take it this PR helps with,

I'm less sure that it actually helps with things. I'll need to dig more. But what would help is retrying.

> but I honestly wouldn't expect things to finish if the error in #4080 persists

The error *is* transient I think, since when I manually wake up the Comm things finish fine. I think that tornado's internal buffer is drained and we're OK to go again?

I'd consider this a WIP for now. I haven't been able to make a small reproducer yet but will give it some more time on Monday.

michaelnarodovitch:
I gave it a try with the reproducer reported in #4080 :
```python
b = bag.from_sequence([1]*100_000, partition_size=1)
bb = b.map_partitions(lambda *_, **__: [b'1'*2**20]*5).persist()
bc = bb.repartition(1000).persist()
```
Unfortunately, this does not fix the issue. Workers still use to loose connection, as reported in #4080 .

mrocklin:
Thanks @michaelnarodovitch .  Would you be willing to use `git bisect` to see if there was a time when this used to work, and maybe try to isolate which commit caused the issue?

michaelnarodovitch:
For me, there was no time, where this used to work. Our production code is still working around with aggressive retries.

distributed.comm.timeout=80s
distributed.scheduler.allowed-failures=999

Ran the reproducer with dask from `pip install git+https://github.com/TomAugspurger/distributed.git@pangeo-788`. Tried with some other versions back then, as reported in #4080 .

mrocklin:
Folks were digging around in the serialization code a bit this summer.  If
you're willing, I would encourage you to go back about 6-12 months and see
if things worked then.

michaelnarodovitch:
> Folks were digging around in the serialization code a bit this summer. If you're willing, I would encourage you to go back about 6-12 months and see if things worked then.
Same problem with dask/distributed 2.7.0 and dask/distributed 2.0.0.

mrocklin:
OK.  Thanks for checking

TomAugspurger:
@mrocklin, at a high-level, what do you expect to happen when a `Scheduler.stream_comms[wroker].send(msg)` fails? We handle the `CommClosedError` (aka "this worker is probably gone forever) well by removing the worker.

For other exceptions like this `BufferError`, I'd expect us to retry some number of times. But eventually, we should likely eventually give up and handle this like the `CommClosedError` by closing the worker?

mrocklin:
> For other exceptions like this BufferError, I'd expect us to retry some number of times. But eventually, we should likely eventually give up and handle this like the CommClosedError by closing the worker?

Yes, that would make sense to me

> Edit: Hmm, this is maybe complicated by the fact that we don't actually await things from BatchedSend._background_send, it just happens in the background. So even if we were to raise from BatchedSend._background_send no one would be there to catch it.

Ah indeed.  Well maybe we can still retry there, and it fails we close the underlying Comm, so that the next time the BatchedSend tries to write it fails and restarts the worker?

TomAugspurger:
I'm going to close this for now in favor of #4135.

This PR really only fixed things for the case where we missed a deadline (failed to await `self.waker.wait(self.next_deadline)` by the time `next_deadline` passed. We don't have any examples of this actually occurring when there isn't a deeper issue like in #4135, and so this is just a theoretical concern. Let's revisit it if we get an actual example where we miss a deadline.

PR #4130, opened 2020-09-27
Log

michaelnarodovitch:
#4080 

did more digging with the reproducer and identified the failure mode, which is causing connection timeouts and failed tasks on a 50 worker cluster.

```python
b = bag.from_sequence([1]*100_000, partition_size=1)
bb = b.map_partitions(lambda *_, **__: [b'1'*2**20]*5).persist()
bc = bb.repartition(1000).persist()
```

1. For the repartition step, the workers will hit the scheduler with `100_000` `who_has` requests in a quite short period in time. 
2. The hits will go through the `ConnectionPool` of `Worker.scheduler`, which has the default connection limit of 512. This will result in ~25,000 concurrent connection attempts and saturate the scheduler listener socket.
Logs would typically output, and first tasks start to fail / worker start to crash
`Connect start: <SCHEDULER>` 170,000 events - Worker side
`Connect done` 16,000 events - Worker side
`On connection` 13,000 events - Scheduler side
`handshake:` 6,000 events - Successful handshakes
``
3. Setting a limit of 8 concurrent connections to the scheduler per worker resolved the issue

I doubt that this is a good solution and do not intend to do further work on this PR. Nevertheless, I believe that this 'fix' is helps to shed light on the root-cause of the issue.

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -286,7 +286,7 @@ async def connect(
             while deadline - time() > 0:
 
                 async def _():
-                    logger.info('Connect start')
+                    logger.info(f'Connect start: {loc}')
                     comm = await connector.connect(
                         loc, deserialize=deserialize, **connection_args
                     )
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -286,6 +286,7 @@ async def connect(
             while deadline - time() > 0:
 
                 async def _():
+                    logger.info('Connect start')
                     comm = await connector.connect(
                         loc, deserialize=deserialize, **connection_args
                     )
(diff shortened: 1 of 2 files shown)


mrocklin:
Thank you for submitting your thoughts here @michaelnarodovitch .  This is helpful.

I'm curious why there are 100,000 who_has calls.  Looking through the Worker implementation it looks like we call `who_has` only when there is something unexpected, and even then we're careful to exponentially back off on these calls.  

Regardless, it seems like we might want to ease back on the 512 connection limit on workers generally, given that they might all be talking to the scheduler simultaneously.

michaelnarodovitch:
Thanks again for looking into that :) Maybe, I was not entirely correct with the explanation. What I see clear evidence for in the logs, are 51,988 stack-traces from
```
File "/usr/local/lib/python3.6/site-packages/distributed/worker.py", line 2161, in query_who_has
response = await retry_operation(self.scheduler.who_has, keys=deps)
await self.query_who_has(dep)
```
thrown from `gather_dep` https://github.com/dask/distributed/blob/b2f594e69751e0733ac595026e04181c55119ae9/distributed/worker.py#L2078-L2085 .

The way, I understand it, is that the workers would query the scheduler where to get the keys for the re-partition step.

Maybe it would be a good idea to refine the `ConnectionPool`, that the semaphores limit connections per host? i.e. max 8 concurrent connections per host?
Or process the `who_has` queries in a `BatchedComm` ?

Issue #4113, opened 2020-09-15
Async client problems


jacobtomlinson:
I've noticed a couple of issues with using clusters and clients asynchronously. These may be user error requiring docs/exceptions or bugs.

#### Sync client with async cluster hangs

```python
from dask.distributed import Client, LocalCluster

cluster = LocalCluster(asynchronous=True)
await cluster

client = Client(cluster)  # Hangs forever

client = Client(cluster, asynchronous=True)  # Works as expected
```

#### Compute with async client reports `TypeError: zip argument #1 must support iteration`

```python
from dask.distributed import Client
import dask.array as da

async with Client(asynchronous=True) as client:
    arr = da.random.random((1000, 1000), chunks=(1000, 100))
    assert arr.mean().compute() < 1
```

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-b9635e77df95> in <module>
      1 arr = da.random.random((1000, 1000), chunks=(1000, 100))
----> 2 assert arr.mean().compute() < 1

~/Projects/dask/dask/dask/base.py in compute(self, **kwargs)
    164         dask.base.compute
    165         """
--> 166         (result,) = compute(self, traverse=False, **kwargs)
    167         return result
    168 

~/Projects/dask/dask/dask/base.py in compute(*args, **kwargs)
    443 
    444     results = schedule(dsk, keys, **kwargs)
--> 445     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    446 
    447 

TypeError: zip argument #1 must support iteration
```

- Dask version: master
- Python version: 3.7.4
- Operating System: macOS
- Install method (conda, pip, source): source

mrocklin:
> Compute with async client reports TypeError: zip argument #1 must support iteration

Yeah, the `get` method used within `.compute()` only operates synchronously.  I don't know a way to make `.compute()` work today.

The best way to do this today is to use `await client.compute(arr.mean())`

jacobtomlinson:
Ah ok, thanks for the workaround!

martindurant:
Suggest close, unless there is a good place to document the situation

quasiben:
We could document this on the [asynchronous page](https://distributed.dask.org/en/latest/asynchronous.html) .  Something like.  We have some nice text [in an example](https://examples.dask.org/applications/async-await.html#Collections)

[...]

PR #4137, opened 2020-09-29
update async doc with example on .compute() vs client.compute()

quasiben:
Fixes https://github.com/dask/distributed/issues/4113


quasiben:
cc @martindurant

martindurant:
+1

PR #4143, opened 2020-10-02
Fix for missing `:` in autosummary docs

gforsyth:
Same fix as in dask/dask#6700

Shouldn't be merged until dask/dask-sphinx-theme#43 is merged and a
release is cut

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,5 +4,5 @@ toolz
 cloudpickle
 dask
 sphinx
-dask-sphinx-theme>=1.3.3
+dask-sphinx-theme>=1.3.4
 sphinx-click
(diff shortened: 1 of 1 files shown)


mrocklin:
@gforsyth should we merge this now?

docs/requirements.txt
```diff
@@ -4,5 +4,5 @@ toolz
 cloudpickle
 dask
 sphinx
-dask_sphinx_theme
+dask-sphinx-theme>=1.3.3
```
jrbourbeau, line 7: We recently pushed out a new `dask-sphinx-theme` release, so I think we want: 

```suggestion
dask-sphinx-theme>=1.3.4
```

gforsyth:
If CI passes, then I think this is ready to go @mrocklin.  Something along the way also seems to have fixed the little house icon, but I'm not rightly sure what did it...

Issue #4114, opened 2020-09-15
Worker clients don't respect timeout config settings


jcrist:
Clients created with `worker_client` or `get_client` don't respect the timeout settings (e.g. `distributed.comm.timeouts.connect`. The timeout is available to set programmatically, but defaults to `3` rather than falling back to the config file. I think this should be as simple as replacing `timeout=3` with `timeout=None` throughout that code path.

jrbourbeau:
Closed by https://github.com/dask/distributed/pull/4146

PR #4146, opened 2020-10-04
Configurable timeouts for worker_client and get_client.

geethanjalieswaran:
Issue: https://github.com/dask/distributed/issues/4114

Previously, connection timeout for worker_client and get_client was
hard-coded to 3s by default.
With this change, the default timeout value is fetched from the
dask config 'distributed.comm.timeouts.connect' if set. And if
this config is not set, '3s' is taken as default.


distributed/worker.py
```diff
@@ -3111,6 +3122,14 @@ def get_client(address=None, timeout=3, resolve_address=True):
     worker_client
     secede
     """
+
+    # Get timeout value from dash config if not passed.
+    # Defaults to 3s if the config is not set.
+    if timeout is None:
+        timeout = parse_timedelta(
+            dask.config.get("distributed.comm.timeouts.connect", "3s"), "s"
+        )
```
mrocklin, line 3131: We probably want to support `get_client(timeout="10s")` as well, so it might be best to wrap the `parse_timedelta` outside of the conditional.  The same is probably true elsewhere as well.

geethanjalieswaran: Thanks for the suggestion, Made the requested changes.

jsignell:
This looks good to me! Ready for merge @dask/maintenance?

martindurant:
Would be good to mention in the docstrings the relevant config keys, and should update distributed.yaml and distributed-schema.yaml with the new entries.

jsignell:
@geethanjalieswaran are you able to make the changes that @martindurant suggests?

distributed/worker.py
```diff
@@ -2993,6 +2993,14 @@ def _get_client(self, timeout=3):
         --------
         get_client
         """
+
+        # Get timeout value from dash config if not passed.
+        # Defaults to 3s if the config is not set.
+        if timeout is None:
+            timeout = dask.config.get("distributed.comm.timeouts.connect", "3s")
```
mrocklin, line 3000: There will always be a default, which is taken from the base distributed.yaml file in this directory.  The default 3s value here won't be used.

```suggestion
            timeout = dask.config.get("distributed.comm.timeouts.connect")
```

distributed/worker.py
```diff
@@ -3111,6 +3123,14 @@ def get_client(address=None, timeout=3, resolve_address=True):
     worker_client
     secede
     """
+
+    # Get timeout value from dash config if not passed.
+    # Defaults to 3s if the config is not set.
```
mrocklin, line 3128: I think that we can probably drop these comments.  I think that the code here is clear enough.

geethanjalieswaran: Ack! Done.

geethanjalieswaran:
> Would be good to mention in the docstrings the relevant config keys, and should update distributed.yaml and distributed-schema.yaml with the new entries.

Thanks for the suggestion. I have updated the doctoring and the entry 'distributed.comm.timeouts.connect' already exists in the distributed.yaml and distributed-schema.yaml (I am reusing the existing config key).

martindurant:
OK, +1 from me 
(except something seems to have timed out in a test, possibly unrelated)

distributed/worker.py
```diff
@@ -3111,6 +3120,12 @@ def get_client(address=None, timeout=3, resolve_address=True):
     worker_client
     secede
     """
+
+    if timeout is None:
+        timeout = dask.config.get("distributed.comm.timeouts.connect", "3s")
```
mrocklin, line 3125: Missed a spot

```suggestion
        timeout = dask.config.get("distributed.comm.timeouts.connect")
```

jrbourbeau (approved):
Thanks @geethanjalieswaran! This is in

jrbourbeau:
Also, I noticed this was your first code contribution to this repository. Welcome!

Issue #4147, opened 2020-10-05
Semaphore gets released too often


lfdversluis:
To avoid generating too much NFS traffic, I am using a semaphore like this to copy data from NFS to a local file storage to read from later:

```
cluster = SLURMCluster(cores=16, memory="64 GB", #processes=1,
                       local_directory="/var/scratch/me/dask_scheduler_spill",
                       interface='ib0', walltime='24:00:00')

# Create a client to submit to.
client = Client(cluster)

# Allocate 10 nodes in the cluster
cluster.scale(10)

print("Waiting for workers")
# Wait until they are ready
client.wait_for_workers(10)
print("Workers are ready!")

sem = Semaphore(max_leases=4, name='data_copy')

def copy_data(location, target, sem):
    with sem:
         copy_tree(location, target, update=True)

client.run(copy_data, root_dataset, local_dataset, sem)
sem.close()  # Clean up the semaphore at the scheduler side
```

Now if I run it I get the error that 
```
site-packages/distributed/semaphore.py", line 472, in release
    raise RuntimeError("Released too often")
RuntimeError: Released too often
```

I tried debugging, I found out that setting `processes=1,` in the construction of SLURMCluster does **not ** fix this. So it looks like a threading issue to me. https://github.com/dask/distributed/issues/4057 should've fixed that, but clearly it didn't? Creating the semaphore with `sem = Semaphore(max_leases=4, name='data_copy', client=client, register=True)` also did not work.

lfdversluis:
Probably important details:

```
>>> import dask
>>> dask.__version__
'2.28.0'
```

and 
```
>>> import dask_jobqueue
>>> dask_jobqueue.__version__
'0.7.1'
```

mrocklin:
cc @fjetter @lr4d

fjetter:
Can you try defining the semaphore instance within the function instead of passing it via the function

```
def copy_data(location, target):
    with Semaphore(max_leases=4, name='data_copy'):
         copy_tree(location, target, update=True)
```

[...]

PR #4151, opened 2020-10-06
Increase robustness of Semaphore.release

lr4d:
The Semaphore.release method does not handle connection errors gracefully. If a connection error appears, the exception may cause a catastrophic failure of the computation.

This PR adds retries to the semaphore release method, which must be
configured using "distributed.comm.retry.count". This may come to
use when the scheduler is under high load and some release
calls may fail.
This PR also avoids raising an exception in case a release fails,
instead a warning is logged at ERROR level, and the Semaphore.release
method will return `False`. The lease will eventually be cleaned up
during the semaphore lease validation check, which may be configured using
"distributed.scheduler.locks.lease-validation-interval" and
"distributed.scheduler.locks.lease-timeout".

Wrt the "Released too often" RuntimeError that affects #4147, one may argue that exception could also be converted into a log message, but I have decided not to change that until we have some more clarity on the background of that issue.

cc @fjetter

### DIFF
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -486,12 +486,18 @@ class Semaphore:
 
     def release(self):
         """
-        Release a semaphore.
+        Release the semaphore.
 
-        Increment the internal counter by one.
+        This will increment the amount of available leases by one.
+
+        Returns
+        -------
+        bool
+            This value indicates whether a lease was released immediately or not. Note that a user should  *not* retry
+            this operation. Under certain circumstances (e.g. scheduler overload) the lease may not be released
+            immediately, but it will always be automatically released after a specific interval configured using
+            "distributed.scheduler.locks.lease-validation-interval" and "distributed.scheduler.locks.lease-timeout".
         """
-
-        """ Release the lock if already acquired """
         if not self._leases:
             raise RuntimeError("Released too often")
 
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -488,8 +488,6 @@ class Semaphore:
         """
         Release the semaphore.
 
-        This will increment the amount of available leases by one.
-
         Returns
         -------
         bool
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -477,8 +477,9 @@ class Semaphore:
             )
         except OSError:  # Too many broken connections, release fails
             logger.error(
-                "Release failed for client=%s, lease_id=%s, name=%s. Cluster network might be unstable."
+                "Release failed for client=%s, lease_id=%s, name=%s. Cluster network might be unstable?"
                 % (self.client.id, lease_id, self.name),
+                exc_info=True,
             )
             return False
 
--- a/distributed/semaphore.py
+++ b/distributed/semaphore.py
@@ -437,6 +437,8 @@ class Semaphore:
             name=self.name,
             timeout=timeout,
             lease_id=lease_id,
+            operation="semaphore acquire: client=%s, lease_id=%s, name=%s"
+            % (self.client.id, lease_id, self.name),
         )
         if result:
             self._leases.append(lease_id)
@@ -460,6 +462,26 @@ class Semaphore:
         timeout = parse_timedelta(timeout)
         return self.client.sync(self._acquire, timeout=timeout)
 
+    async def _release(self):
+        # popleft to release the oldest lease first
+        lease_id = self._leases.popleft()
+        logger.info("%s releases %s for %s", self.client.id, lease_id, self.name)
+
+        try:
+            return await retry_operation(
+                self.client.scheduler.semaphore_release,
+                name=self.name,
+                lease_id=lease_id,
+                operation="semaphore release: client=%s, lease_id=%s, name=%s"
+                % (self.client.id, lease_id, self.name),
+            )
+        except OSError:  # Too many broken connections, release fails
+            logger.error(
+                "Release failed for client=%s, lease_id=%s, name=%s. Cluster network might be unstable."
+                % (self.client.id, lease_id, self.name),
+            )
+            return False
+
     def release(self):
         """
         Release a semaphore.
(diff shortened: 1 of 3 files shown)


distributed/semaphore.py
```diff
@@ -460,6 +462,26 @@ def acquire(self, timeout=None):
...
+
+        try:
+            return await retry_operation(
+                self.client.scheduler.semaphore_release,
+                name=self.name,
+                lease_id=lease_id,
+                operation="semaphore release: client=%s, lease_id=%s, name=%s"
+                % (self.client.id, lease_id, self.name),
+            )
+        except OSError:  # Too many broken connections, release fails
```
lr4d, line 478: Change this for a generic `Exception`?

distributed/semaphore.py
```diff
@@ -460,6 +462,26 @@ def acquire(self, timeout=None):
...
+            logger.error(
+                "Release failed for client=%s, lease_id=%s, name=%s. Cluster network might be unstable."
```
fjetter, line 480: We should log the original exception. Otherwise this is impossible to debug if things don't go smooth. Suspecting the cluster network is only one possibility and I wouldn't jump to any conclusions but just log what happened.

distributed/semaphore.py
```diff
@@ -460,6 +462,26 @@ def acquire(self, timeout=None):
...
+                % (self.client.id, lease_id, self.name),
+            )
+            return False
```
fjetter, line 483: The release itself doesn't return anything. The `False` return value is not used anywhere. Why introduce this?

fjetter: At the very least I would then return a True if it was released successfully

distributed/tests/test_semaphore.py
```diff
@@ -518,3 +522,80 @@ def access_limited(val, sem):
...
+        c.scheduler = rpc
+        semaphore = await Semaphore(
+            max_leases=2, name="resource_we_want_to_limit", client=c
+        )
+        await semaphore.acquire()
+        pool.activate()  # Comm chaos starts
+        with captured_logger("distributed.utils_comm") as caplog:
+            await semaphore.release()
+        logs = caplog.getvalue().split("\n")
+        print(logs)
```
fjetter, line 542: remove print

fjetter (changes requested):
For the record, this has nothing to do with #4147. As it looks right now, #4147 might be an issue that the leases are never properly acquired. This is effectively the scenario this exception should protect you from, therefore, I'm inclined to keep this exception raising for now

distributed/semaphore.py
lr4d: > Why introduce this?

So that a user/machine can easily figure out if the release did not succeed

distributed/tests/test_semaphore.py
lr4d: whoops

distributed/semaphore.py
lr4d: > At the very least I would then return a True if it was released successfully

I agree, this way we keep the logic consistent. I will change the appropriate tests then, unless you suggest a better alternative

fjetter:
While we're at it, the semaphore refresh leases coroutine might also have the potential to fail due to flaky connections. Since it is scheduler via the `add_callback` functionality, I'm not sure how the behaviour there will be. Can you have a look into this as well? (Might be more difficult to provoke in test cases, though)

distributed/semaphore.py
lr4d: yes, and change the "default" return value of release to true

lr4d:
@fjetter I'll have a look but would probably open a different PR for that

distributed/semaphore.py
```diff
@@ -460,6 +462,28 @@ def acquire(self, timeout=None):
...
+            logger.error(
+                "Release failed for client=%s, lease_id=%s, name=%s. Cluster network might be unstable?"
+                % (self.client.id, lease_id, self.name),
+                exc_info=True,
+            )
+            return False
+
     def release(self):
         """
         Release a semaphore.
```
fjetter, line 489: Please add a comment about the meaning of the return value. In particular, this should instruct users how to proceed in case of a failed release. I'm still wondering if we should return anything for the release function. After all, what the user is not supposed to do is release again in this scenario since the most likely scenario is that the connection broke up but the lease ID was already removed from internal tracking. Releasing again would cause the "Released too many times" exception

distributed/semaphore.py
```diff
@@ -460,22 +462,46 @@ def acquire(self, timeout=None):
...
+            return False
+
     def release(self):
         """
-        Release a semaphore.
+        Release the semaphore.
 
-        Increment the internal counter by one.
-        """
+        This will increment the amount of available leases by one.
```
fjetter, line 491: This is not necessarily true. This is true for the stdlib semaphore but not this one. the exception is that we cannot increase the number of available leases to a larger value than initially specified. This is also documented in the class documentation but I'm afraid this might be slightly misleading

lr4d: Can you suggest a better alternative?

lr4d:
btw @fjetter while adding retries to the refresh-leases method, I've seen that these flaky connections can also raise an exception in `Semaphore.get_value()`. We can probably refactor the code a bit to have retry operations for all methods that involve scheduler communication in a cleaner way than it is done currently

distributed/semaphore.py
lr4d: I've just removed that line for now, I don't think the previous line was very informative either

fjetter:
> We can probably refactor the code a bit to have retry operations for all methods that involve scheduler communication in a cleaner way than it is done currently

One could probably put this directly in the `client.scheduler` hook but I suggest to do this in another PR. I guess this has non-trivial implications

fjetter:
Test failures seem to be unrelated. I re-triggered the run and reported #4163

fjetter (approved):

TomAugspurger:
Hmm the failure happened again... I don't recall seeing it on other branches, but I agree that it doesn't look related.

I'm looking into it this morning.

TomAugspurger:
Maybe good news, that test failed in https://github.com/dask/distributed/pull/4164/checks?check_run_id=1242648960#step:8:115, so I think it can be ignored here. I'll continue to try and debug it.

Issue #4153, opened 2020-10-06
test_gpu_metrics fails


jakirkham:
**What happened**:

When running the `test_gpu_metrics` tests, found that they failed due to an `ImportError`.

**What you expected to happen**:

Would have expected these tests to pass.

**Minimal Complete Verifiable Example**:

```python
$ python -m pytest distributed/tests/test_gpu_metrics.py
```

**Anything else we need to know?**:

<details>
<summary>Test failure:</summary>

```python
============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
rootdir: /datasets/jkirkham/devel/distributed, configfile: setup.cfg
plugins: asyncio-0.12.0
collected 1 item                                                               

distributed/tests/test_gpu_metrics.py F                                  [100%]

=================================== FAILURES ===================================
_______________________________ test_gpu_metrics _______________________________

    def test_func():
        result = None
        workers = []
        with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:
    
            async def coro():
                with dask.config.set(config):
                    s = False
                    for i in range(5):
                        try:
                            s, ws = await start_cluster(
                                nthreads,
                                scheduler,
                                loop,
                                security=security,
                                Worker=Worker,
                                scheduler_kwargs=scheduler_kwargs,
                                worker_kwargs=worker_kwargs,
                            )
                        except Exception as e:
                            logger.error(
                                "Failed to start gen_cluster, retrying",
                                exc_info=True,
                            )
                            await asyncio.sleep(1)
                        else:
                            workers[:] = ws
                            args = [s] + workers
                            break
                    if s is False:
                        raise Exception("Could not start cluster")
                    if client:
                        c = await Client(
                            s.address,
                            loop=loop,
                            security=security,
                            asynchronous=True,
                            **client_kwargs,
                        )
                        args = [c] + args
                    try:
                        future = func(*args)
                        if timeout:
                            future = asyncio.wait_for(future, timeout)
                        result = await future
                        if s.validate:
                            s.validate_state()
                    finally:
                        if client and c.status not in ("closing", "closed"):
                            await c._close(fast=s.status == Status.closed)
                        await end_cluster(s, workers)
                        await asyncio.wait_for(cleanup_global_workers(), 1)
    
                    try:
                        c = await default_client()
                    except ValueError:
                        pass
                    else:
                        await c._close(fast=True)
    
                    def get_unclosed():
[...]

PR #4154, opened 2020-10-06
Fix `test_gpu_metrics` failure

jakirkham:
Fix https://github.com/dask/distributed/issues/4153

The `test_gpu_metrics` test fails. This attempts to fix it.

cc @quasiben


distributed/tests/test_gpu_metrics.py
```diff
@@ -6,7 +6,7 @@
 
 @gen_cluster()
 async def test_gpu_metrics(s, a, b):
-    from distributed.diagnostics.nvml import handles
+    from distributed.diagnostics.nvml import _pynvml_handles as handles
 
     assert "gpu" in a.metrics
     assert len(s.workers[a.address].metrics["gpu"]["memory-used"]) == len(handles)
```
jakirkham, line 12: When running this, I encounter another issue. Short snippet from the failure included below for brevity. Not sure if this is an indication of a bug or another issue in the test.

```python
    @gen_cluster()
    async def test_gpu_metrics(s, a, b):
        from distributed.diagnostics.nvml import _pynvml_handles as handles
    
        assert "gpu" in a.metrics
>       assert len(s.workers[a.address].metrics["gpu"]["memory-used"]) == len(handles)
E       TypeError: object of type 'int' has no len()
```

quasiben:
Perhaps we should remove this test.  Most of this is covered in https://github.com/dask/distributed/blob/master/distributed/diagnostics/tests/test_nvml.py .  Thoughts ?

distributed/tests/test_gpu_metrics.py
jakirkham: We only return one handle from `_pynvml_handles` as well. So I'm not sure if `len` was suppose to be here or if this is a typo.

https://github.com/dask/distributed/blob/a1dc5f437b39c1b35a9b05cbc048e3a793b89715/distributed/diagnostics/nvml.py#L29-L30

jakirkham: It looks like this function use to return a `list`, but PR ( https://github.com/dask/distributed/pull/3810 ) changed it to return a single value. So maybe this test doesn't make sense (at least how it is written) any more.

quasiben: I changed the way nvml/handles work -- we are now only handing back one handle per worker

quasiben:
We could also add a check for `memory-used` in `test_nvml.py`

jakirkham:
Yeah I'm fine dropping it. After doing a bit of digging (really not familiar with this piece of code), looks like it is just a historical artifact.

Also happy to consolidate into our more recent working tests.

jakirkham:
Ok this now passes for me locally.

quasiben:
Cool!  Maybe let's leave this as the solution and check back in with @jacobtomlinson when he returns before removing

jakirkham:
Looks like the tests in [`distributed/diagnostics/tests/test_nvml.py`]( https://github.com/dask/distributed/blob/6efd5b6147d29ec613bd83c72cdbd6a5a45a1438/distributed/diagnostics/tests/test_nvml.py ) also don't pass for the same reason.

```python
____________________________ test_1_visible_devices ____________________________

    def test_1_visible_devices():
        os.environ["CUDA_VISIBLE_DEVICES"] = "0"
        output = nvml.one_time()
>       assert len(output["memory-total"]) == 1
E       TypeError: object of type 'int' has no len()

distributed/diagnostics/tests/test_nvml.py:21: TypeError
```

jakirkham:
Ok have now fixed that error too.

jakirkham:
> Cool! Maybe let's leave this as the solution and check back in with @jacobtomlinson when he returns before removing

Well the test is still there. Have just moved it to join the other NVML tests. Thought this is what was being suggested.

Mainly my interest here is to get Dask-CUDA's CI to stop failing. Don't have strong feelings about what we do in the future, but maybe we can get this in to fix those issues?

distributed/diagnostics/tests/test_nvml.py
```diff
@@ -17,11 +18,15 @@ def test_one_time():
...
-    assert len(output["memory-total"]) == 1
+    h = nvml._pynvml_handles()
+    assert output["memory-total"] == pynvml.nvmlDeviceGetMemoryInfo(h).total
 
 
 @pytest.mark.parametrize("CVD", ["1,0", "0,1"])
 def test_2_visible_devices(CVD):
+    if pynvml.nvmlDeviceGetCount() <= 1:
+        pytest.skip("Machine only has a single GPU")
+
```
jakirkham, line 29: Added this to skip the multi-GPU test on single GPU machines based on our conversation offline, can you please take a look and see if I got this right, Ben? :)

quasiben:
Thanks @jakirkham.  All this passes now including the other GPU tests within distributed on a DGX machine.  I am getting a failure on a workstation (I think it's a GPU configuration issue:

```python
ipdb> pynvml.nvmlDeviceGetSerial(h)
*** pynvml.nvml.NVMLError_NotSupported: Not Supported
```

I'll look at this early tomorrow morning.

jakirkham:
Yeah looking at [the docs]( https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g56b0c288a1c2eb60b9949aff60d64486 ) I see the following...

> For all products with an inforom.

An [earlier version of the docs]( http://developer.download.nvidia.com/compute/DevZone/NVML/doxygen/group__group4.html#g56b0c288a1c2eb60b9949aff60d64486 ) says...

> For Fermi-class products only.

So maybe this GPU is not supported for this operation? If so, maybe there is an alternative to this API call that would work more generally?

Though it's also possible there's some configuration issue we don't understand (as you have noted). Might be worth verify whether this specific GPU can run this function or not before digging in too deep.

quasiben:
After reviewing/changing [permissions from pynvml](https://github.com/gpuopenanalytics/pynvml#nvml-permissions)  everything works as expected.  Merging in.

Thanks @jakirkham !

jakirkham:
Ah ok. Good to know. Thanks for checking Ben! 😄

jacobtomlinson:
Thanks for dealing with this folks!

PR #4155, opened 2020-10-07
adaptive_target: no more workers then runnable tasks

### LINKED ISSUE #4108: Adaptive scaling went above the number of available tasks
Currently Dask supports [adaptive scaling](https://docs.dask.org/en/latest/setup/adaptive.html).  This allows it to scale the number of workers based on load.  The logic for how many workers to ask for is based on many things, like how much work we have, how much memory we need, and so on.  

I recently observed a cluster scale up beyond the number of available tasks.  It shouldn't do this.  I had three tasks, each of which took several minutes, and I saw my cluster scale up to ten workers.  Probably there should be a maximum applied in there somewhere that wasn't being set.  There appears to already be some code to do this (the intent is there) but perhaps there is a but.  The relevant code is here

https://github.com/dask/distributed/blob/4115f55c655000b04a8132c9ca62fa07bd6df6f5/distributed/scheduler.py#L5337-L5390

twoertwein:
I think that the memory condition in `Scheduler.adaptive_target` is swapped. If the total memory of all workers is less than half (or for some reason less than 60%) of the memory limit, then you should spawn more workers (and not the other way round)?

If I understand the code correctly, `adapt` worked only when the memory limit was set to zero.

Fixes https://github.com/dask/dask-jobqueue/issues/463

Edit:
Limit the number of workers by the number of runnable tasks
Fixes  #4108

### DIFF
--- a/distributed/deploy/tests/test_adaptive.py
+++ b/distributed/deploy/tests/test_adaptive.py
@@ -6,10 +6,10 @@ from time import sleep
 import dask
 import pytest
 
-from distributed import Client, wait, Adaptive, LocalCluster, SpecCluster, Worker
-from distributed.utils_test import gen_test, slowinc, clean
-from distributed.utils_test import loop, nodebug, cleanup  # noqa: F401
+from distributed import Adaptive, Client, LocalCluster, SpecCluster, Worker, wait
 from distributed.metrics import time
+from distributed.utils_test import clean, cleanup, gen_test, loop, slowinc  # noqa: F401
+from distributed.worker_client import get_worker
 
 
 @pytest.mark.asyncio
@@ -449,3 +449,44 @@ async def test_update_adaptive(cleanup):
         await asyncio.sleep(0.2)
         assert first.periodic_callback is None
         assert second.periodic_callback.is_running()
+
+
+@pytest.mark.asyncio
+async def test_adaptive_too_many_workers(cleanup):
+    """ Make sure that not more workers are spawned than runnable tasks."""
+    n_tasks = 3
+    n_max_workers = 10
+
+    async with LocalCluster(
+        n_workers=0, threads_per_worker=1, memory_limit=0, asynchronous=True
+    ) as cluster:
+        cluster.adapt(minimum=0, maximum=n_max_workers, interval="100 ms")
+        async with Client(cluster, asynchronous=True) as client:
+            futures = client.map(sleep, range(n_tasks))
+            await client.gather(futures)
+
+        log = cluster._adaptive.log
+        for state in log:
+            if state[1]["status"] != "up":
+                continue
+            assert state[1]["n"] <= n_tasks
+
+
+@pytest.mark.parametrize("memory_limit", (0, "1G"))
+@pytest.mark.asyncio
+async def test_adaptive_memory_limit(memory_limit, cleanup):
+    """Make sure that adapt() works even when a memory limit is set."""
+
+    def _test_fun(x):
+        sleep(3)
+        return get_worker().name
+
+    n_tasks = 3
+
+    async with LocalCluster(
+        n_workers=0, threads_per_worker=1, memory_limit=memory_limit, asynchronous=True
+    ) as cluster:
+        async with Client(cluster, asynchronous=True) as client:
+            cluster.adapt(minimum=1, maximum=n_tasks, interval="100 ms")
+            futures = client.map(_test_fun, range(n_tasks))
+            assert max(await client.gather(futures)) > 0
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6201,31 +6201,27 @@ class Scheduler(ServerNode):
             self.total_occupancy / target_duration
         )  # TODO: threads per worker
 
-        # Avoid a few long tasks from asking for many cores
-        ws: WorkerState
-        tasks_processing = 0
-        for ws in self.workers.values():
-            tasks_processing += len(ws._processing)
-
-            if tasks_processing > cpu:
-                break
-        else:
-            cpu = min(tasks_processing, cpu)
-
         if self.unrunnable and not self.workers:
             cpu = max(1, cpu)
 
-        # Memory
-        limit_bytes = {addr: ws._memory_limit for addr, ws in self.workers.items()}
-        worker_bytes = [ws._nbytes for ws in self.workers.values()]
-        limit = sum(limit_bytes.values())
-        total = sum(worker_bytes)
-        if total > 0.6 * limit:
+        # add more workers if more than 60% of memory is used
+        limit = sum(ws.memory_limit for ws in self.workers.values())
+        used = sum(ws.nbytes for ws in self.workers.values())
+        memory = 0
+        if used > 0.6 * limit:
             memory = 2 * len(self.workers)
-        else:
-            memory = 0
 
         target = max(memory, cpu)
+
+        # avoid having more workers than runnable tasks
+        runnable_states = ("processing", "no-worker")
+        tasks = 0
+        for task_group in self.task_groups.values():
+            tasks += sum(task_group.states[state] for state in runnable_states)
+            if tasks >= target:
+                break
+        target = min(target, tasks)
+
         if target >= len(self.workers):
             return target
         else:  # Scale down?


mrocklin:
`limit` refers to our target maximal amount of memory. `total` refers to the amount of memory used currently.  I think that the logic was correct before.  Are you running into something that caused you to investigate this?

twoertwein:
> `limit` refers to our target maximal amount of memory. `total` refers to the amount of memory used currently.

Doesn't this mean that we want `limit > total` to create new workers - currently it is the opposite.

> Are you running into something that caused you to investigate this?

Yes, the link dask_jobqueue issue but the same also happens for the LocalCluster. The important part is that a non-zero memory limit is set for the workers (it works if it is zero):
```py
import time

from dask.distributed import Client, LocalCluster
from distributed.worker_client import get_worker


def test_fun():
    time.sleep(5)
    return get_worker().name


if __name__ == "__main__":
    # create simple task graph
    n = 4
    graph = {f"test_{i}": (test_fun,) for i in range(n)}
    targets = [f"test_{i}" for i in range(n)]

    workers = 10  # we shouldn't need more then n=4 workers

    # LocalCluster
    kwargs = {
        "n_workers": 1,
        "diagnostics_port": None,
        "threads_per_worker": 1,
        "memory_limit": "2G",
        "processes": True,
    }
    with LocalCluster(**kwargs) as cluster:
        cluster.adapt(minimum=1, maximum=workers)
            client.register_worker_callbacks(
                lambda dask_worker: print("setup", dask_worker.name)
            )
            print(client.get(graph, targets))  # prints only 0s. If memory_limit=0, more works are spawned
```

Do you have recommendations to debug what is going on with the failing tests?

twoertwein:
I think the remaining tests would pass if #4108 was fixed.

distributed/scheduler.py
```diff
@@ -5379,16 +5379,27 @@ def adaptive_target(self, comm=None, target_duration=None):
...
-        limit = sum(limit_bytes.values())
+        limit = sum(limit_bytes)
         total = sum(worker_bytes)
-        if total > 0.6 * limit:
+        if total < 0.6 * limit or limit <= 0:
             memory = 2 * len(self.workers)
         else:
             memory = 0
 
         target = max(memory, cpu)
```
twoertwein, line 5391: shouldn't this be `min(memory, cpu)`? If we have not enough CPUs/memory for more workers, we shouldn't spawn more workers

distributed/scheduler.py
```diff
@@ -5379,16 +5379,27 @@ def adaptive_target(self, comm=None, target_duration=None):
...
+
+        # runnable tasks
+        tasks = len(
+            [
+                task
+                for task in self.tasks.values()
+                if task.state in ("processing", "no-worker") and not task.waiting_on
```
twoertwein, line 5398: do we also need `released` and `memory` here? I don't think we need `erred` and `forgotten`.

Thank you for having a nice overview of the states here :) https://distributed.dask.org/en/latest/scheduling-state.html

twoertwein: does iterating over all tasks take potentially too much time? We could stop as soon as we have more than `target` tasks.

distributed/deploy/tests/test_adaptive.py
```diff
@@ -449,3 +449,40 @@ async def test_update_adaptive(cleanup):
         await asyncio.sleep(0.2)
         assert first.periodic_callback is None
         assert second.periodic_callback.is_running()
+
+
+def test_adaptive_too_many_workers():
```
twoertwein, line 454: I'm not too familiar with `async/await` but I assume you would like to have tests with `sleep` to by `async`. I tried to async'ify this test but I get the following error:

```
_____________________________________________________________________________ ERROR at teardown of test_adaptive_too_many_workers ______________________________________________________________________________

    @pytest.fixture
    def cleanup():
        with clean():
>           yield

distributed/utils_test.py:1541: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../local/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
distributed/utils_test.py:1535: in clean
    del thread_state.on_event_loop_thread
../../local/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    @contextmanager
    def check_thread_leak():
    
        yield
    
        start = time()
        while True:
            bad = [
                t
                for t, v in threading._active.items()
                if t not in active_threads_start
                and "Threaded" not in v.name
                and "watch message" not in v.name
                and "TCP-Executor" not in v.name
            ]
            if not bad:
                break
            else:
                sleep(0.01)
            if time() > start + 5:
                from distributed import profile
    
                tid = bad[0]
                thread = threading._active[tid]
                call_stacks = profile.call_stack(sys._current_frames()[tid])
>               assert False, (thread, call_stacks)
E               AssertionError: (<Thread(Profile, started daemon 140229604820736)>, ['  File "/results/twoertwe/home/local/lib/python3.8/threading.py"...', '  File "/results/twoertwe/home/git/distributed/distributed/profile.py", line 269, in _watch
E                 \tsleep(interval)
E                 '])
E               assert False

distributed/utils_test.py:1435: AssertionError
```
using the following code (not yet pushed):

```py
@pytest.mark.asyncio
async def test_adaptive_too_many_workers(cleanup):
    """
    Make sure that not more workers are spawned than runnable tasks.

    GH 4155
    """
    n = 4
    graph = {f"test_{i}": (slowinc, 1) for i in range(n)}

    def _assert_not_too_many_worker(dask_worker):
        assert dask_worker.name < n

    workers = 10  # we shouldn't need more than n=4 workers
    async with LocalCluster(
        n_workers=0,
        processes=True,
        threads_per_worker=1,
        memory_limit=0,
        asynchronous=True,
    ) as cluster:
        async with Client(cluster, asynchronous=True) as client:
            client.register_worker_callbacks(_assert_not_too_many_worker)
            cluster.adapt(minimum=1, maximum=workers)
            client.get(graph, list(graph.keys()))

```

twoertwein:
@mrocklin if you do not agree with the memory-related part, I can separate the "not spawning more jobs than runnable tasks" into its own PR

twoertwein:
bump @mrocklin @guillaumeeb

Other people seem to encounter the same issue (SLURMCluster not adapting) https://github.com/esi-neuroscience/acme/issues/14

I have been using this patch for more than a month on a SLURM cluster and it addressed my main concern (wasting resources) using dask on a cluster:

- do not spawn more workers than necessary and
- down-scale number of workers: I have typically one long'ish task left at the end, no need to block resources with idling workers.

distributed/scheduler.py
mrocklin: I'm not sure I understand.  If we need more memory then we should spawn more workers.  If we need more CPUs then we should spawn more workers.  At least that's my perception of common desire today.

distributed/scheduler.py
```diff
@@ -6216,16 +6216,25 @@ def adaptive_target(self, comm=None, target_duration=None):
...
+        if total < 0.6 * limit or limit <= 0:
             memory = 2 * len(self.workers)
         else:
             memory = 0
 
         target = max(memory, cpu)
+
+        # runnable tasks
+        tasks = 0
+        for task in self.tasks.values():
```
mrocklin, line 6230: There may be millions of tasks.  In general we can't iterate through all tasks every time we want to ask for the target.  Instead I recommend that you iterate through `TaskGroup`s and look at the `.states` attribute.

mrocklin:
>> limit refers to our target maximal amount of memory. total refers to the amount of memory used currently.

> Doesn't this mean that we want limit > total to create new workers - currently it is the opposite.

I don't think so.  For example, if I have a cluster with 100TB of memory (limit=1e14) and it stores no data (total=0) then I think that we don't want to scale up.

twoertwein:
thank you @mrocklin for your comments! I think I understand my misconception :)

I will create a separate PR for the number of runnable task (after I switched to `TaskGroup`s) and then spend more time to analyze why  Local/SLURMCluster do not spawn more workers in my setup. Based on my current understanding it shouldn't be related to memory at all (must be the CPU part).

distributed/scheduler.py
twoertwein: sorry, yes you are right.

distributed/scheduler.py
```diff
@@ -6201,31 +6201,27 @@ def adaptive_target(self, comm=None, target_duration=None):
...
-        # Avoid a few long tasks from asking for many cores
-        ws: WorkerState
-        tasks_processing = 0
-        for ws in self.workers.values():
-            tasks_processing += len(ws._processing)
-
-            if tasks_processing > cpu:
-                break
-        else:
-            cpu = min(tasks_processing, cpu)
```
twoertwein, line 6213: this code prevented more worker to spawn for me. It is not needed anymore as the code now limits the number of targeted workers by the number of runnable tasks.

distributed/scheduler.py
```diff
@@ -6201,31 +6201,27 @@ def adaptive_target(self, comm=None, target_duration=None):
...
-        limit_bytes = {addr: ws._memory_limit for addr, ws in self.workers.items()}
-        worker_bytes = [ws._nbytes for ws in self.workers.values()]
-        limit = sum(limit_bytes.values())
-        total = sum(worker_bytes)
-        if total > 0.6 * limit:
+        # add more workers if more than 60% of memory is used
+        limit = sum(ws.memory_limit for ws in self.workers.values())
+        used = sum(ws.nbytes for ws in self.workers.values())
+        memory = 0
+        if used > 0.6 * limit:
```
twoertwein, line 6211: it might be good to change this to `used > 0.6 * limit and limit > 0`. With no memory limit set, it tries to always increase the number of workers (that is why I initially thought it was related to memory).

twoertwein:
I think that all changes in this PR are not addressing the root cause: `self.total_occupancy` doesn't seem to get updated well enough (neither for SLURMCluster nor LocalCluster). I would expect that if a task is sleeping for 30s, that `total_occupancy` will increase and at some point reach 30. By default,  the targeted duration is 5s. As soon as the `total_occupancy` is above 5s, dask should spawn more workers. Unfortunately, `total_occupancy` doesn't ever go above 2.0 in my test examples even though wall time is much larger.

If all tasks take much longer than spawning a worker, something like the following is an ugly workaround:
```py
with dask.config.set({"distributed.adaptive.target-duration": 0.1}):
    ...
```

The only reason we don't see this behavior when no memory limit is set (`limit = 0`), is that the if condition for memory is always True for `limit = 0`.

Iterating over `TaskGroup.status` does not seem to contain all tasks. Might just be missing the `'no-worker'` cases?

guillaumeeb:
For the record, the initial problem was that the graph used for the computation had not enough information to enable adaptive scaling. See dask/dask-jobqueue#463.

Issue #4061, opened 2020-08-21
Dask workers spilling to disk too early


KrishanBhasin:
**What happened**:

I have a LocalCluster with ~50 workers. I use pure Dask Dataframe operations to create a dataframe with a large task graph (~3 million tasks).

When this dataframe is submitted to the LocalCluster (e.g. `ddf.to_csv()`), the compute finishes successfully, however at no point does RAM usage go above ~25% (the highest I have seen an individual worker go is ~35%). When I monitor the file system, I can see that `dask-worker-space` gets very large (>100GB).

I believe this is costing me lots of performance due to de/serialisation costs (I see tasks in the dashboard like e.g. `disk-write-read-csv`...).

I have set the dask config to as below (terminate at >100% as my data is skewed across workers - a single one breaking 100% isn't an issue):

```yaml
distributed:
  worker:
    memory:
      target: 0.7
      spill: 0.9
      pause: 1.0
      terminate: 1.2
```

When I have the cluster computing, I have attached a new client to it and run checks that the workers have been configured correctly (it appears they have):
```py

#returns 0.7 as expected
client.run(lambda dask_worker:dask_worker.memory_target_fraction)

#returns 0.9 as expected
client.run(lambda dask_worker:dask_worker.memory_spill_fraction)

#returns 1.0 as expected
client.run(lambda dask_worker:dask_worker.memory_pause_fraction)
```

I have DEBUG logs turned on, and also do not see any lines indicating spill-to-disk has started as per https://github.com/dask/distributed/blob/6a32208717c04aa9ccb4defdf253b8a00cab32b7/distributed/worker.py#L2667-L2669

If I set `worker.memory.target` to `False`, I continue to get low RAM usage (~25% per worker) and large `dask-worker-space`.

If I set `worker.memory.spill` to `False`, I continue to get low RAM usage (~25% per worker, apart from 1 of my workers which is at ~50%) and large `dask-worker-space`.

If I set `worker.memory.target` and `worker.memory.spill` to False, I see `dask-worker-space` remains empty and RAM usage steadily climbs. This unfortunately is not a solution for me, as my data doesn't wholly fit into memory, so this will eventually fail.

**What you expected to happen**:

- My workers to wait until they are at ~70% RAM util before `dask-worker-space` starts filling up.

**Minimal Complete Verifiable Example**:

I think colleague of mine got one using `dask.datasets.timeseries` - I will ask him to share it here.

```python
# Put your MCVE code here
```

[...]

PR #4156, opened 2020-10-08
Adjust value of zict buffer's `target` to ~inf when `memory_target_fraction` is False

KrishanBhasin:
Helps address #4061 

If `self.memory_target_fraction` is set to False, then `target` becomes `0`, which appears to trigger immediate spill-to-disk.

I believe what we actually want is for target to be ~infinite, so that this measure is not used to trigger spill-to-disk.

When running my code with this code diff, I no longer observe unexpectedly early spill-to-disk.


mrocklin:
This code block is in a conditional `if self.memory_limit` so I'm not sure that this will have an effect.  I might be reading things wrong though.

KrishanBhasin:
Are you referring to this block?
https://github.com/dask/distributed/blob/0e88c5ba3c2f9ba3e15f19f6ea75d90b8b3693ea/distributed/worker.py#L545-L547

I can't see any other block that it sits in (but its been a long day today so apologies if I'm missing something obvious)

mrocklin:
Yes.  If self.memory_limit is zero then we'll proceed down to the else:
block, where self.data = dict()

KrishanBhasin:
My understanding is that `memory_limit` is the overall limit on the worker, then the `memory_*_fraction` is a multiplier we apply to that value.

My use case is keeping the limit for my workers (e.g. 20GB each), but setting `memory_target_fraction` to False, and using `memory_spill_fraction` to decide when to get my workers to disk.

What I saw was that in this block if `memory_target_fraction` was set to `False`, (but `memory_limit` was a valid memory limit), multiplying them together returned 0


```python
>>> 233434 * False
0
```

Running this code diff with `memory_target_fraction=False`, I am no longer seeing my early spill-to-disk problems

KrishanBhasin:
An MVCE for the problem this fixes is below:

```python
import dask

client = Client(threads_per_worker=1, n_workers=4, memory_target_fraction=False)

df = dask.datasets.timeseries(
    start="2000-01-01",
    end="2010-01-31",
    freq="1s",
    partition_freq="1d",
)
df.persist()
```
On `2.30.0`, the `Status` dashboard tab immediately shows spill-to-disk occurring, and the `Worker` tab shows the workers barely breaking 3% memory use.

Using this diff, the memory use ramp up to ~70% as expected, before spill to disk occurs.

KrishanBhasin:
Using `sys.maxsize` feels a bit hacky to me, but it looks like given the way that Zict appears to work, I think this is the best way to implement this.

If I'm reading the below snippet correctly, every time a new key is added, the LRU will call `evict()` as many times as needed to bring the memory use down below `n`. As `n` is a primitive integer, the only 'fix' I can see right now is to make it the maximum possible value so this process is triggered as late/rarely as possible.

https://github.com/dask/zict/blob/3a9ea886583ba9920be8ab8ecec4adde23401a1a/zict/lru.py#L52-L70
```python
def __setitem__(self, key, value):
        if key in self.d:
            del self[key]

        weight = self.weight(key, value)

        if weight <= self.n:
            self.d[key] = value
            self.i += 1
            self.heap[key] = self.i

            self.weights[key] = weight
            self.total_weight += weight
        else:
            for cb in self.on_evict:
                cb(key, value)

        while self.total_weight > self.n:
            self.evict()
```

mrocklin:
I see, so you still want to have spill-to-disk behavior, you just want to rely on the psutil process memory check exclusively.  OK, that makes sense to me.  Merging.

KrishanBhasin:
Thank you!

I have since had a quick thought about using `float("inf")` instead of `sys.maxsize` - if you prefer that (as it guarantees to always be greater than the `memory_limit`, and Zict should handle the `float` instead of an `int` just fine), I can do another quick PR to update this.

PR #4158, opened 2020-10-09
WIP: Add Semaphore.refresh_leases retries

lr4d:
Builds on top of #4151 , only the last commit is new.

The test added is still flaky.


lr4d:
The rationale here is similar to the issue linked, to increase the robustness of the Semaphore during unstable periods. In case the refresh_leases method fails, we retry the operation. In case all retries fail, we generate an appropriate log message

PR #4167, opened 2020-10-14
Fix distributed comm connect timeout logic

jcrist:
The previous timeout was limited to too short a timeout per attempt for a connect & handshake on a highly loaded system (e.g. one oversubscribed with worker threads). We do three things here:

- Move the connect retry loop out to its own separate task. The loop with backoff was to retry to work with things like dynamic DNS (e.g. k8s), which should only affect the initial connect. The handshake can then happen with a separate timeout. This shortens the amount of work being done in the capped timeout, which helps a bit.

- We up the initial minimum timeout to something larger. Originally it was 1.5 with some jitter, we up it to a minimum of 5 seconds or a fraction of the total timeout. Since a DNS race condition should only happen rarely, this shouldn't cause degradation in that case but gives a longer initial connect attempt in the common case.

- Call `comm.abort` on handshake timeout in `inproc` comms, rather than letting the comm finalizer handle it (squashes an innocuous warning).

Fixes #4165.

Note that at some level of inproc workers things will still fail due to limits of the python VM, so this isn't really a real use case (users should have multiple threads for a single inproc worker, not multiple inproc workers). However, I think the fix provided here is still useful for some more realistic systems, as before the full connect process (connect & handshake) would get the following timeout attempts by default: `[1.5, 2.25, 3.375, 2.875]`, none of which may have been long enough to complete in a distributed setting.

PR #4176, opened 2020-10-21
Remove hard coded connect handshake timeouts

fjetter:
This code section, in particular the handshake (#4019), is causing issues for me in "high load" scenarios. I see broken connections popping up once I reach about ~200 workers and these errors tear down the entire system. From the exceptions I cannot infer whether the handshake runs into a timeout or whether it is completely broken (mostly because our internal user reports are not as thorough as I'd like them to be but I'm investigating).
Whenever the handshake fails, it is raised as a `CommClosedError` which is, unfortunately, not an `EnvironmentError`. Therefore, what I *wanted* to change is the exception type upon which we retry to be more inclusive.
Then I had a look at the code and was really confused about the retry behaviour and the individual timeouts and started to (subjectively) simplify this piece of code and write a test for it. The test is a bit messy, works but I'd appreciate suggestions on how this can be implemented cleaner.

W.r.t the implementation, I am open for suggestions and can revert anything/everything/nothing depending on the feedback here. 

Here a quick dump of my thoughts to this

* We should not only retry `EnvironmentError`s but rather more inclusive error classes (at least the `CommClosed` we raise ourselves)
* We should retry with (an exponential) backoff
* Ideally with a jitter
* I don't have a strong opinion about individual timeouts of the read/write/connect. Therefore I chose this approach, where every step may take time until the deadline is reached. I guess one could argue that this should somehow be split up but I want for the reduced complexity approach instead.
* I removed the backoff cap since I figured we wouldn't really need it here. This is just a gut feeling. Happy to introduce something
* I oriented the implementation on https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ and chose the FullJitter approach. Not sure if this applies 100% but it made me feel as if my choices where "data driven" and this sounded similar enough to our situation :)

In case anybody wonders, with the chosen base of 0.01 this results in (non randomised) backoffs

```python
In [1]: backoff_base = 0.01

In [2]: [backoff_base * 2 ** ix for ix in range(10)]
Out[2]: [0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12]

In [3]: sum([backoff_base * 2 ** ix for ix in range(10)])
Out[3]: 10.23
```
With the jitter this likely adds up to a probably significantly larger number of max retries

There is currently an alternative fix for this section open, see https://github.com/dask/distributed/pull/4167 cc @jcrist

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -274,68 +274,59 @@ async def connect(
         )
         raise IOError(msg)
 
-    backoff = 0.01
-    if timeout and timeout / 20 < backoff:
-        backoff = timeout / 20
+    backoff_base = 0.01
+    attempt = 0
 
-    retry_timeout_backoff = random.randrange(140, 160) / 100
-
-    # This starts a thread
     while True:
         try:
-            while deadline - time() > 0:
+            comm = await asyncio.wait_for(
+                connector.connect(loc, deserialize=deserialize, **connection_args),
+                timeout=deadline - time(),
+            )
 
-                async def _():
-                    comm = await connector.connect(
-                        loc, deserialize=deserialize, **connection_args
-                    )
-                    local_info = {
-                        **comm.handshake_info(),
-                        **(handshake_overrides or {}),
-                    }
-                    try:
-                        handshake = await asyncio.wait_for(comm.read(), 1)
-                        write = await asyncio.wait_for(comm.write(local_info), 1)
-                        # This would be better, but connections leak if worker is closed quickly
-                        # write, handshake = await asyncio.gather(comm.write(local_info), comm.read())
-                    except Exception as e:
-                        with suppress(Exception):
-                            await comm.close()
-                        raise CommClosedError() from e
+            local_info = {
+                **comm.handshake_info(),
+                **(handshake_overrides or {}),
+            }
 
-                    comm.remote_info = handshake
-                    comm.remote_info["address"] = comm._peer_addr
-                    comm.local_info = local_info
-                    comm.local_info["address"] = comm._local_addr
+            # This would be better, but connections leak if worker is closed quickly
+            # write, handshake = await asyncio.gather(comm.write(local_info), comm.read())
+            handshake = await asyncio.wait_for(comm.read(), deadline - time())
+            await asyncio.wait_for(comm.write(local_info), deadline - time())
 
-                    comm.handshake_options = comm.handshake_configuration(
-                        comm.local_info, comm.remote_info
-                    )
-                    return comm
+            comm.remote_info = handshake
+            comm.remote_info["address"] = comm._peer_addr
+            comm.local_info = local_info
+            comm.local_info["address"] = comm._local_addr
 
-                with suppress(TimeoutError):
-                    comm = await asyncio.wait_for(
-                        _(), timeout=min(deadline - time(), retry_timeout_backoff)
-                    )
-                    break
-            if not comm:
-                _raise(error)
+            comm.handshake_options = comm.handshake_configuration(
+                comm.local_info, comm.remote_info
+            )
+            return comm
         except FatalCommClosedError:
             raise
-        except EnvironmentError as e:
+        # CommClosed, EnvironmentError inherit from OSError
+        except (OSError, TimeoutError) as e:
+            if comm:
+                try:
+                    comm.close()
+                except Exception:
+                    comm.abort()
+                    comm = None
+
             error = str(e)
             if time() < deadline:
-                logger.debug("Could not connect, waiting before retrying")
+                # FullJitter see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+                backoff = random.uniform(0, backoff_base * 2 ** attempt)
+                # If there is no more time left, don't wait longer than the initial timeout
+                backoff = min(backoff, deadline - time())
+                attempt += 1
+                logger.debug(
+                    "Could not connect, waiting for %s before retrying", backoff
+                )
                 await asyncio.sleep(backoff)
-                backoff *= random.randrange(140, 160) / 100
-                retry_timeout_backoff *= random.randrange(140, 160) / 100
-                backoff = min(backoff, 1)  # wait at most one second
             else:
                 _raise(error)
-        else:
-            break
-
-    return comm
 
 
 def listen(addr, handle_comm, deserialize=True, **kwargs):
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -213,9 +213,15 @@ class Listener(ABC):
 
     async def on_connection(self, comm: Comm, handshake_overrides=None):
         local_info = {**comm.handshake_info(), **(handshake_overrides or {})}
+
+        timeout = dask.config.get("distributed.comm.timeouts.connect")
+        timeout = parse_timedelta(timeout, default="seconds")
         try:
-            write = await comm.write(local_info)
-            handshake = await comm.read()
+            # Timeout is to ensure that we'll terminate connections eventually.
+            # Connector side will employ smaller timeouts and we should only
+            # reach this if the comm is dead anyhow.
+            write = await asyncio.wait_for(comm.write(local_info), timeout=timeout)
+            handshake = await asyncio.wait_for(comm.read(), timeout=timeout)
             # This would be better, but connections leak if worker is closed quickly
             # write, handshake = await asyncio.gather(comm.write(local_info), comm.read())
         except Exception as e:
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -307,7 +307,7 @@ async def connect(
     except Exception as exc:
         comm.abort()
         raise IOError(
-            f"Timed out trying to connect to {addr} after {timeout} s"
+            f"Timed out during handshake while connecting to {addr} after {timeout} s"
         ) from exc
 
     comm.remote_info = handshake
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -262,38 +262,40 @@ async def connect(
     comm = None
 
     start = time()
-    deadline = start + timeout
+
+    def time_left():
+        deadline = start + timeout
+        return max(0, deadline - time())
 
     backoff_base = 0.01
     attempt = 0
 
     # Prefer multiple small attempts than one long attempt.
     intermediate_cap = timeout / 5
-    while True:
+    active_exception = None
+    while time_left() > 0:
         try:
             comm = await asyncio.wait_for(
                 connector.connect(loc, deserialize=deserialize, **connection_args),
-                timeout=min(intermediate_cap, deadline - time()),
+                timeout=min(intermediate_cap, time_left()),
             )
             break
         except FatalCommClosedError:
             raise
         # CommClosed, EnvironmentError inherit from OSError
         except (TimeoutError, OSError) as exc:
+            active_exception = exc
+            # FullJitter see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
 
-            if time() < deadline:
-                # FullJitter see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
-                backoff = random.uniform(0, backoff_base * 2 ** attempt)
-                # If there is no more time left, just raise the exception
-                if backoff > deadline - time():
-                    raise IOError(
-                        f"Timed out trying to connect to {addr} after {timeout} s"
-                    ) from exc
-                attempt += 1
-                logger.debug(
-                    "Could not connect, waiting for %s before retrying", backoff
-                )
-                await asyncio.sleep(backoff)
+            upper_cap = min(time_left(), backoff_base * (2 ** attempt))
+            backoff = random.uniform(0, upper_cap)
+            attempt += 1
+            logger.debug("Could not connect, waiting for %s before retrying", backoff)
+            await asyncio.sleep(backoff)
+    else:
+        raise IOError(
+            f"Timed out trying to connect to {addr} after {timeout} s"
+        ) from active_exception
 
     local_info = {
         **comm.handshake_info(),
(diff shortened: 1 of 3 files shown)


distributed/comm/core.py
```diff
@@ -274,68 +274,59 @@ def _raise(error):
...
-                with suppress(TimeoutError):
-                    comm = await asyncio.wait_for(
-                        _(), timeout=min(deadline - time(), retry_timeout_backoff)
-                    )
-                    break
-            if not comm:
-                _raise(error)
+            comm = await asyncio.wait_for(
+                connector.connect(loc, deserialize=deserialize, **connection_args),
+                timeout=deadline - time(),
```
jcrist, line 284: This drops the change added in https://github.com/dask/distributed/pull/3104 where the initial connect timeout was set lower than the full timeout to kill connect attempts that hung due to dynamic DNS race conditions. In #4167 I made the initial attempt longer (the existing progression was too short for my test case, no attempt was long enough to succeed). I think we still want our *connect* timeouts to be shorter than the full deadline (to allow a few connect attempts), and that we should only retry on connects not on handshakes here (once data is able to send over the wire, we should abort retrying to the application code above this). Perhaps the connect timeouts should start as a fraction of the full timeout (with a min of a couple seconds), and backoff quickly to the remaining timeout?

jcrist (commented):
Thanks @fjetter , this looks better than my attempt (apologies for never finishing that up). A few comments.

distributed/comm/core.py
fjetter: > This drops the change added in #3104 where the initial connect timeout was set lower than the full timeout to kill connect attempts that hung due to dynamic DNS race conditions

Sure, wasn't aware of what exactly was going on here. I can cap these intermediates with a fraction of the total

> and that we should only retry on connects not on handshakes here (once data is able to send over the wire, we should abort retrying to the application code above this)

I disagree on this point. The handshake is a rather internal detail of the connect. How is the application supposed to deal with a failed handshake, let alone detect it?

fjetter:
Tests fail because I changed the way exceptions are reraised. Will need to change this logic again

distributed/comm/core.py
jcrist: > How is the application supposed to deal with a failed handshake, let alone detect it?

It's not a firm stance. The retries here on connect were to deal with connection-level issues. If a socket connects, but the handshake fails, I'd argue that retrying here isn't worth it (we had a valid connection, then we lost it), and the retry should happen higher up at the application level. Not a big deal though, adding a retry on handshake failure adds no harm, I just don't think it's necessary.

quasiben:
@pentschev and I have been looking at these issues as well.   @pentschev are you interested in testing out this PR for our UCX case ? If not, I can check it out

pentschev:
@quasiben I tested this now and unfortunately this doesn't resolve our issues, but this particular piece of code seems to be causing trouble in different situations which is a bit worrying. I'm still not certain where's the problem on our side, but it seems like it's not really related to a timeout for us, we mostly see connections getting closed during read/write even if we increase the `wait_for` timeout to a very large number, so perhaps we're missing an `await` or something analogous to there.

quasiben:
Thank you for testing @pentschev

distributed/comm/core.py
fjetter: understood. I think this is true for actual failures but with the code on master, there is also the possibility that these handshakes simply timeout. At the very least we should increase the timeouts of these handshakes since the 1s might just not be enough if the system is under stress.

I figured I'd just deal with all at once and not distinguish between the cases for the sake of reduced complexity.

jcrist: >  At the very least we should increase the timeouts of these handshakes since the 1s might just not be enough if the system is under stress.

I agree with this, and agree with including them with the larger timeout. I'm just arguing that in the case that we get a successful connection but fail to handshake, we shouldn't retry here. Something like (pseudocode):

```python
while time_remaining_until_deadline:
    try:
        comm = try_to_connect_with_an_increasing_connect_timeout()
        break
    except TimeoutError:
        continue

handle_handshake_in_remaining_time_until_deadline_or_timeout(comm)
```

If the connection succeeds, but then fails during handshake, that indicates either:

- The handshake will always fail, some incompatibility between nodes
- The connection was valid, but then was dropped. This could happen after the comm is returned just as easily, I don't feel like we should handle that here.
- The handshake timed out, which means we've reached the full deadline

None of these IMO should be retried here inside the comm code.

That said, if it's easier to express the timeout logic this way that's fine with me, but I don't think we should expend effort to loop on handshake failure here.

fjetter: Ok, finally understood this. I think removing the handshakes from the retries is ok-ish. There may be other comm errors, of course but they can probably be handled further up the stack.
with the `deadline - time()` logic, removing them from retries does not really make anything more complex imo. I changed the code accordingly

fjetter:
After implementing the suggestion of @jcrist to not include the handshakes in the retries, I added a test checking for "slow handshakes" and stumbled over the listener timeouts as well. I completely removed the timeouts for the handshake in the listener since I figured, it is fine on the connector side to enforce timeouts but I'm not entirely sure about this. If we require a timeout there as well, we'll need to configure it somehow since 1s is not enough.

distributed/comm/core.py
```diff
@@ -214,8 +214,8 @@ async def _():
     async def on_connection(self, comm: Comm, handshake_overrides=None):
         local_info = {**comm.handshake_info(), **(handshake_overrides or {})}
         try:
-            write = await asyncio.wait_for(comm.write(local_info), 1)
```
fjetter, line 217: Instead of introducing another config option, would `distributed.comm.timeouts.connect` be also applicable here?

pentschev:
With the most recent changes I've also seen `CommClosedError`s go away in our Dask-CUDA+UCX use case. However, I see some cleanup issues that I'm not sure yet whether they're related to the handshake connection. Regardless of that, this seems like great progress and we can continue looking up for the cleanup issues down the road. I'm definitely +1 on this PR.

fjetter:
I still had an issue in the code where I encountered negative backoffs. That might've been the cause for the failing builds. The retry logic itself didn't change otherwise. I think the important change in the last commits was to remove the timeout from the listeners but I'm not sure if it is safe to not have any there.

Other than this, I'm also wondering what to do with the `comm` in case the handshake fails. I am now trying to close it but what happens if the close fails or gets stuck for whatever reason?
@jcrist you added a `comm.abort` for this case. any reason why the `come.close()` isn't sufficient? Shall I add a `comm.abort` as well?

jcrist:
> added a comm.abort for this case. any reason why the come.close() isn't sufficient? Shall I add a comm.abort as well?

What you have here is fine. I added a `comm.abort` in that location since it was being called from a synchronous context (`comm.abort` isn't an `async` method).

distributed/comm/core.py
```diff
@@ -262,79 +262,65 @@ async def connect(
...
-                _raise(error)
-        else:
-            break
-
+        # CommClosed, EnvironmentError inherit from OSError
+        except (TimeoutError, OSError) as exc:
+            active_exception = exc
+            # FullJitter see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+
+            upper_cap = min(time_left(), backoff_base * (2 ** attempt))
```
jcrist, line 290: We should also bump the `intermediate_cap` by some fraction, in case the initial size is too small. As is right now, no connect attempt can last longer than `timeout/5`. Perhaps 1.5 x it every attempt?

distributed/comm/core.py
```diff
@@ -262,79 +262,65 @@ async def connect(
...
+            backoff = random.uniform(0, upper_cap)
```
jcrist, line 291: The backoff should get progressively larger each attempt, as is this is adding a random sleep with progressively larger ranges (but 0 is still valid). I liked the old algorithm better of roughly 1.5x the previous backoff with some jitter. Why did you make this change?

distributed/comm/core.py
jcrist: I think we still want a timeout on the listener side, but it could be large (we just want to ensure servers can kill stuck connections). `distributed.comm.timeouts.connect` seems fine to me, both the `comm.write` and `comm.read` should complete within that time.

distributed/comm/core.py
fjetter: The AWS blog post does not include timeouts which is why I slightly changed the logic but I wouldn't classify it as a bug

I oriented myself on the FullJitter approach which is defined by

1. `sleep = random_between(0, min(cap, base * 2 ** attempt))`

where I chose the remaining time left as the cap. The reason why I chose this as a cap is because I don't want the coroutine to unnecessarily block for too much longer if there is no chance for it to complete anyhow.

In a previous iteration I was doing it the other way round, namely

2. `sleep = min(cap, random_between(0, base * 2 ** attempt))`

which is slightly different but what they both have in common is that they do not progressively produce larger backoffs. Only the average expected backoff is increasing progressively but 0 is theoretically still valid for all attempts. Only the algorithm EqualJitter offers the guarantee that zero is never chosen but it performs slightly worse than the others.

Would you prefer 1. over 2. or have I an error in my logic?

distributed/comm/core.py
fjetter: I don't have a good feeling about this intermediate cap but increasing it by a factor each attempt should be a safe default. I'll add the x1.5

fjetter: Maybe, I'll add a smaller factor? Otherwise, we'd effectively limit ourselves to effectively three tries

```
In [1]: sum([0.2 * 1.5 ** attempt for attempt in range(3)])
Out[1]: 0.95
```

as I said, I don't have a good feeling about how important these intermediate caps really are

distributed/comm/core.py
jcrist: > The AWS blog post does not include timeouts which is why I slightly changed the logic

But the calculation here is still for a backoff in attempts, not timeouts (so it seems the blogpost should apply as written?).

I missed the FullJitter option, which does look like what you've implemented here. For connect failures though, I think we do want to ensure some amount of backoff is used in case the server is still starting up or unavailable for other reasons. Both the "Equal Jitter" and "Decorrelated Jitter" options should (IIUC) provide a guarantee of non-zero backoff times (Decorrelated looking slightly better), but we could also use what you have here with a non-zero min (perhaps 0.05 or something small). What you have here seems fine too though on further thought, thanks for the explanation.

distributed/comm/core.py
jcrist: I agree on this, running into DNS race conditions feels like the sign of larger problems elsewhere, but for now we should at least match the existing behavior. My goal with increasing the value here is that depending on the value of `distributed.comm.timeouts.connect`, no `intermediate_cap` may be large enough to complete if it's set at 1/5 the timeout. I'm not too worried about limiting to 3 attempts (note this is only true if the timeout is hit each time, not some other error), as more than that is likely a sign of deeper issues. 1.5 or 1.25 both seem fine, I wouldn't want to go lower than that.

distributed/comm/core.py
fjetter: > (so it seems the blogpost should apply as written?).

There is the one scenario where the new backoff would breach the timeout

(numbers only for clarity)
* Initial timeout maybe 5s
* We're in attempt 10 with about 100ms remaining
* New backoff would calculate to 200ms
* We'd wait and try to connect again but the 11th attempt will then have a timeout of zero, i.e. it is guaranteed to fail

-> We'd wait for 5.1s, i.e. longer than the configured timeout

I introduced the cap since it would, at least give the marginal chance of another try with 100ms timeout and it would not breach the total amount of time waited. The analysis of the blog post assumes that we'll retry indefinitely (or a fixed amount of max attempts) until we're successful but we're in a slightly different scenario here.

> "Decorrelated Jitter" options should (IIUC) provide a guarantee of non-zero backoff times (Decorrelated looking slightly better)

Correct, this was a false statement of mine. The decorrelated jitter is always lower capped by the base in this example.

fjetter:
Finally, I'm wondering if the default connect timeout should be increased. We currently have 10s as a default connect timeout but I guess this was set at a time where we had a simpler retry mechanism (w/out intermediate capping) and without handshakes. Considering that multiple people encountered the `CommClosed` exceptions since the handshake was hard coded to 1s this might indicate that a more conservative value should be set for the overall timeout (My issues did not appear again with the default, so I'd be fine either way, just asking the question)

jcrist:
I'd be fine increasing the default to something higher (perhaps 30s?), but don't think that necessarily needs to be done here. Unless others think otherwise, I think we should get this fix in but leave the timeout the same.

jcrist:
This generally looks good to me, but there's a test failure at `test_worker_who_has_clears_after_failed_connection` that looks related.

fjetter:
Yes, I'm looking into the tests and am currently suspecting the intermediate_cap to be too small and am trying to increase it. I'm also considering to remove the cap entirely after the initial failure but am waiting for the build.

However, I don't think it is actually caused by this change but somehow amplified. I noticed an awful lot of error logs which are apparently retried and may be responsible for an overall flakiness of the system at the moment #4199

TomAugspurger:
How's this looking @fjetter? Does https://github.com/dask/distributed/pull/4200 / https://github.com/dask/distributed/issues/4199 need to be resolved before this can be merged?

For reference, we're planning to backport this fix and issue a release, ideally today but we can push if it isn't ready. Does this build in any way on https://github.com/dask/distributed/pull/4200, so that it would need to be backported too, or are they likely independent?

fjetter:
I'm pretty certain #4200 / #4199 was introduced by #4107 which is not part of distributed==2.30.0, therefore the backport is not necessary. I'm struggling to find a reason for all of the failures and am just suspecting this to be connected to #4199 but I cannot confirm it.

TomAugspurger:
Thanks. https://github.com/dask/distributed/pull/4204 is testing out this diff against 2.30.x. If CI passes there then we can be confident that the test failures here are unrelated.

We plan to merge this to master, cherry-pick & backport it to 2.30.x, and then release 2.30.1.

jennakwon06:
Hello - I see that 2.30.1 is in the changelog ( https://distributed.dask.org/en/latest/changelog.html ) but not available in PyPi yet.

We are blocked on the connect timeout fix and was wondering when it would be available on PyPI.

Thanks!

jcrist:
The 2.30.1 release isn't out yet, see https://github.com/dask/community/issues/105 for more info.

jrbourbeau (commented):
Tests have passed over in https://github.com/dask/distributed/pull/4204, so I'm going to merge this PR and include it in the 2.30.1 release. Thanks @fjetter (and @jcrist @quasiben @pentschev for reviewing)!

Issue #4181, opened 2020-10-22
`AsyncProcess` exit handlers execute too early


pentschev:
I've been working on identifying some issues with UCX cleanup and found one of the issues to be with with exit handlers, in this particular case, it comes from `AsyncProcess`' handler in https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/process.py#L340-L348

For better understanding, consider the log below -- I deliberately removed some lines that weren't relevant to make things easier to understand, order is preserved.

<details>

```
distributed.process - INFO - reaping stray process <SpawnProcess(Dask Worker process (from Nanny), started daemon)>
distributed.process - INFO - reaping stray process <SpawnProcess(Dask Worker process (from Nanny), started daemon)>
distributed.core - DEBUG - Connection from 'ucx://127.0.0.1:57264' to Scheduler
distributed.core - DEBUG - Message from 'ucx://127.0.0.1:57264': {'op': 'retire_workers', 'workers': (0, 1), 'reply': True}
distributed.core - DEBUG - Calling into handler retire_workers
distributed.nanny - INFO - Closing Nanny at 'ucx://127.0.0.1:53785'
distributed.nanny - INFO - Closing Nanny at 'ucx://127.0.0.1:60064'
distributed.process - DEBUG - [<AsyncProcess Dask Worker process (from Nanny)>] process 44169 exited with code -15
distributed.nanny - INFO - Worker process 44169 was killed by signal 15
distributed.process - DEBUG - [<AsyncProcess Dask Worker process (from Nanny)>] got message {'op': 'stop'}
distributed.process - DEBUG - [<AsyncProcess Dask Worker process (from Nanny)>] process 44165 exited with code -15
distributed.nanny - INFO - Worker process 44165 was killed by signal 15
distributed.process - DEBUG - [<AsyncProcess Dask Worker process (from Nanny)>] got message {'op': 'stop'}
distributed.core - DEBUG - Message from 'ucx://127.0.0.1:57264': {'op': 'close', 'close_workers': True, 'reply': True}
distributed.scheduler - INFO - Scheduler closing...
distributed.scheduler - INFO - Scheduler closing all comms
distributed.batched - INFO - Batched Comm Closed: While writing, the connection was closed
distributed.batched - INFO - Batched Comm Closed: While writing, the connection was closed
```

</details>

[...]

Issue #4173, opened 2020-10-20
Flaky test_broken_worker_during_computation


jrbourbeau:
I've noticed `distributed/tests/test_failed_workers.py::test_broken_worker_during_computation` has been sporadically failing in recent PRs (e.g. https://travis-ci.org/github/dask/distributed/jobs/737168606#L1607)

martindurant:
Good news! I can get this to fail locally - occasionally.

martindurant:
Git bisect has the test passing (20 runs) up to and not including dcb46d07318282d8a8976aadb0e4003144a42ae7 (cc @gforsyth ). The funny thing is, that the failure comes after 13 successful runs at that commit, but gets slowly worse with following commits - if I got my bisecting order right.

martindurant:
(that commit was the merger of the rather large #4107 , which passed in the PR but failed in py36 on merger)

gforsyth:
Thanks for digging that up, @martindurant -- I had run this test locally a bunch, but not 20 successive times.

Can you confirm what kind of failure you're seeing when it does fail?  Is it an asyncio timeout or something else?

martindurant:
Luckily, it failed on the third run once I had tracebacks back on
```
(py37) /Users/mdurant/code/distributed ((dcb46d07...)|BISECTING)> pytest --count=20 distributed/tests/test_failed_workers.py::test_broken_worker_during_computation -x
========================================================================================== test session starts ===========================================================================================
platform darwin -- Python 3.7.8, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /Users/mdurant/code/distributed, configfile: setup.cfg
plugins: repeat-0.8.0, flaky-3.7.0, hypothesis-5.26.0, cov-2.10.1, profiling-1.7.0
collected 20 items

distributed/tests/test_failed_workers.py ..F

================================================================================================ FAILURES ================================================================================================
______________________________________________________________________________ test_broken_worker_during_computation[3-20] _______________________________________________________________________________

    def test_func():
        result = None
        workers = []
        with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:

            async def coro():
                with dask.config.set(config):
                    s = False
                    for i in range(5):
                        try:
                            s, ws = await start_cluster(
                                nthreads,
                                scheduler,
                                loop,
                                security=security,
                                Worker=Worker,
                                scheduler_kwargs=scheduler_kwargs,
                                worker_kwargs=worker_kwargs,
[...]

PR #4184, opened 2020-10-23
Replace `AsyncProcess` exit handler by weakref.finalize

pentschev:
This is an attempt at solving #4181 . I believe this is a more reliable approach, given we can ensure the process will not get destroyed before it's garbage collected.

### DIFF
--- a/distributed/process.py
+++ b/distributed/process.py
@@ -82,24 +82,15 @@ class AsyncProcess:
             ),
         )
         self._name = self._process.name
-        self._proc_finalizer = weakref.finalize(self, self._get_finalizer())
+        self._proc_finalizer = weakref.finalize(
+            self, _asyncprocess_finalizer, self._process
+        )
         self._watch_q = PyQueue()
         self._exit_future = Future()
         self._exit_callback = None
         self._closed = False
 
         self._start_threads()
-
-    def _get_finalizer(self):
-        def finalize(proc=self._process, r=repr(self)):
-            if proc.is_alive():
-                try:
-                    logger.info("reaping stray process %s" % (proc,))
-                    proc.terminate()
-                except OSError:
-                    pass
-
-        return finalize
 
     def __repr__(self):
         return "<%s %s>" % (self.__class__.__name__, self._name)
@@ -343,3 +334,12 @@ class AsyncProcess:
     @daemon.setter
     def daemon(self, value):
         self._process.daemon = value
+
+
+def _asyncprocess_finalizer(proc):
+    if proc.is_alive():
+        try:
+            logger.info("reaping stray process %s" % (proc,))
+            proc.terminate()
+        except OSError:
+            pass
--- a/distributed/process.py
+++ b/distributed/process.py
@@ -81,14 +81,25 @@ class AsyncProcess:
                 dask.config.global_config,
             ),
         )
-        _dangling.add(self._process)
         self._name = self._process.name
+        self._proc_finalizer = weakref.finalize(self, self._get_finalizer())
         self._watch_q = PyQueue()
         self._exit_future = Future()
         self._exit_callback = None
         self._closed = False
 
         self._start_threads()
+
+    def _get_finalizer(self):
+        def finalize(proc=self._process, r=repr(self)):
+            if proc.is_alive():
+                try:
+                    logger.info("reaping stray process AAA %s" % (proc,))
+                    proc.terminate()
+                except OSError:
+                    pass
+
+        return finalize
 
     def __repr__(self):
         return "<%s %s>" % (self.__class__.__name__, self._name)
@@ -118,8 +129,8 @@ class AsyncProcess:
             # We don't join the thread here as a finalizer can be called
             # asynchronously from anywhere
 
-        self._finalizer = weakref.finalize(self, stop_thread, q=self._watch_q)
-        self._finalizer.atexit = False
+        self._thread_finalizer = weakref.finalize(self, stop_thread, q=self._watch_q)
+        self._thread_finalizer.atexit = False
 
     def _on_exit(self, exitcode):
         # Called from the event loop when the child process exited
@@ -292,7 +303,7 @@ class AsyncProcess:
         immediately and does not ensure the child process has exited.
         """
         if not self._closed:
-            self._finalizer()
+            self._thread_finalizer()
             self._process = None
             self._closed = True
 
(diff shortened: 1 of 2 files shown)


quasiben:
This does resolve a long standing cleanup issue for UCX and we think this is a cleaner approach to halt workers.  @jcrist if you have time can you look this over ?

quasiben:
@fjetter your thoughts here would also be helpful

distributed/process.py
```diff
@@ -81,15 +81,26 @@ def __init__(self, loop=None, target=None, name=None, args=(), kwargs={}):
...
+        self._proc_finalizer = weakref.finalize(self, self._get_finalizer())
         self._watch_q = PyQueue()
         self._exit_future = Future()
         self._exit_callback = None
         self._closed = False
 
         self._start_threads()
 
+    def _get_finalizer(self):
+        def finalize(proc=self._process, r=repr(self)):
```
jcrist, line 94: `r` is never referenced in the function, could be removed

distributed/process.py
```diff
@@ -81,15 +81,26 @@ def __init__(self, loop=None, target=None, name=None, args=(), kwargs={}):
                 dask.config.global_config,
             ),
         )
-        _dangling.add(self._process)
         self._name = self._process.name
+        self._proc_finalizer = weakref.finalize(self, self._get_finalizer())
```
jcrist, line 85: It's important that a finalizer contain no references to the object being finalized, otherwise a cycle will prevent it from ever being run.

Right now, `_get_finalizer` does the right thing, but IMO it'd be clearer to define the finalizer as an external function and pass in the `process` as an arg, making it crystal clear that the finalizer contains no references to `self`.

jcrist (commented):
A few quick comments. No thoughts on whether this is the right approach over other methods - generally I'd hope we could explicitly manage closing these processes rather than requiring a finalizer to implicitly handle them.

pentschev:
> A few quick comments. No thoughts on whether this is the right approach over other methods - generally I'd hope we could explicitly manage closing these processes rather than requiring a finalizer to implicitly handle them.

I think the main problem we have today is that using exit handlers as globals as is the case in Distributed seems incorrect, as the order will not be respected. I haven't looked at all exit handlers but the few I looked at will all have the same problem, a couple examples:

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/client.py#L4838-L4850
https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/deploy/spec.py#L636-L641

I don't know whether this is the right approach either, I was hoping to suggest this and begin a discussion on potential pitfalls and alternatives. At the moment I don't see a much better approach,

distributed/process.py
pentschev: > It's important that a finalizer contain no references to the object being finalized, otherwise a cycle will prevent it from ever being run.

Sorry for the syntactical annoyance, a cycle _will_ or _may_ prevent it? I'm asking because then we may need to fix most of the uses we have of it in distributed today, including in `AsyncProcess` itself, a few examples:

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/process.py#L121

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/comm/tcp.py#L143

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/scheduler.py#L1512

It's not super difficult to change that, so I can probably work on a separate PR for that if you think that has a big potential for problems.

distributed/process.py
pentschev: Good catch, perks of using copy/paste. :) 

Fixed in https://github.com/dask/distributed/pull/4184/commits/d61f92932a9aa775d436f05125f5096a8cd2cf1a .

distributed/process.py
pentschev: > Right now, `_get_finalizer` does the right thing, but IMO it'd be clearer to define the finalizer as an external function and pass in the `process` as an arg, making it crystal clear that the finalizer contains no references to `self`.

Fixed in https://github.com/dask/distributed/pull/4184/commits/d61f92932a9aa775d436f05125f5096a8cd2cf1a

jcrist: > a cycle will or may prevent it

*Will* prevent it. Finalizers that contain references to `obj` will never run automatically (manually calling them will still work), as the cycle will prevent automatic collection.

https://docs.python.org/3/library/weakref.html#weakref.finalize

To be more specific, given the signature `finalize(obj, func, *args, **kwargs)`, none of `func`, `args`, or `kwargs` can contain a reference to `obj`. The examples you posted above are all fine.

jcrist (approved):
Provided tests pass, this looks good to me.

distributed/process.py
pentschev: > The examples you posted above are all fine as they don't break this rule.

Were you referring to `self._process` that I had in `finalize(proc=self._process)`? If so, then we do have that issue in the TCP example I pasted above if you look at the finalizer:

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/comm/tcp.py#L164-L170

jcrist: That code is fine since no where in `def finalize` is `self` bound as part of the closure. If we were writing that code today, I'd pull that function out to a top level function and pass in the arguments as `args`/`kwargs to `weakref.finalize` (as I asked you to do here), but since it already exists and is correct (if a bit opaque) I wouldn't mess with it.

fjetter (approved):
I agree with @jcrist and would prefer a world where we would not need to cleanup with finalizers but instead would be able to close everything intentionally.

However, as far as finalizers go I think this approach is slightly better than the one before since it deals with each object/process individually instead of in bulk. In particular, as already mentioned, it respects the ordering in the order the objects where created. Before the order was determined on module import time. Not sure if this makes any difference but this approach seems also simpler to reason about than the module level finalizer.

fjetter:
There are a lot of failures on travis/py3.6. Most seem unrelated but I'm not entirely sure.

pentschev:
I'm also not sure whether the 3.6 failures are related to this, it seems that they aren't, but happy to look more into that if there's some suspicion this PR is the cause.

quasiben:
I restarted the 3.6 CI job

pentschev:
Seems like there's still one test failing with a timeout: `test_broken_worker_during_computation`. By looking at builds from other PRs in https://travis-ci.org/github/dask/distributed/jobs/738345888 and https://travis-ci.org/github/dask/distributed/jobs/738412961, the same test is failing, so I believe it's not a side-effect of the changes here.

fjetter:
The `test_broken_worker_during_computation` failure is already reported in #4173

pentschev:
Thanks @fjetter , I hadn't seen that. In this case, I really think this PR seems safe.

jrbourbeau (approved):
Thanks @pentschev! (and @jcrist @fjetter for reviewing)

pentschev:
Thanks everyone for reviews and merging! :)

PR #4185, opened 2020-10-23
Managing priorities manually

SultanOrazbayev:
When defining priorities manually, somehow they are added with a negative sign. In the example below, last task should have the highest priority, but ends up with a lower priority (-4):

```python
from dask.distributed import Client, LocalCluster
cluster = LocalCluster()
client = Client(cluster)

def test_func(i):
    return

futures = [client.submit(test_func, i, priority=i) for i in range(5)]
```

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1961,7 +1961,7 @@ class Scheduler(ServerNode):
         for key in set(priority) & touched_keys:
             ts = self.tasks[key]
             if ts.priority is None:
-                ts.priority = (-(user_priority.get(key, 0)), generation, priority[key])
+                ts.priority = ((user_priority.get(key, 0)), generation, priority[key])
 
         # Ensure all runnables have a priority
         runnables = [ts for ts in touched_tasks if ts.run_spec]


jrbourbeau (commented):
Thanks for the PR @SultanOrazbayev. IIUC the `-` is intentional as when the scheduler constructs task state transition recommendations we add tasks to the `recommendations` dictionary in _reverse_ order of their priority

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/scheduler.py#L2018-L2020

This is so the highest priority tasks, those with the largest `priority=` value in `client.submit`, will be added to the `recommendations` dict last. This is important because later on when we go to actually transition the tasks (e.g. to the "waiting" state) using `recommendations.popitem`

https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/scheduler.py#L4787-L4790

higher priority tasks will be processed first.

jrbourbeau:
@SultanOrazbayev I should have asked, did you find tasks were being run in the opposite order you expected?

SultanOrazbayev:
> @SultanOrazbayev I should have asked, did you find tasks were being run in the opposite order you expected?

Thank you for the quick feedback, yes I noticed the reverse ordering in the code but got confused by the values in the task details of the dashboard. With regards to the task completion... I think it's wrong, but maybe I am doing it the wrong way. So, consider a modification of the above:

```python
from dask.distributed import Client, LocalCluster
cluster = LocalCluster(n_workers=1, resources={'foo': 1})
client = Client(cluster)

def test_func(i):
    import time
    time.sleep(i)
    return i

futures = [client.submit(test_func, i, priority=i, resources={'foo': 1}) for i in range(5)]
```
This should complete the longest running task first, but in the dashboard I see that the completion of the first task is instant. Or, am I coding it wrong?

mrocklin:
Probably the scheduler just got the first task first and so started running
it right away.  The scheduler then got the subsequent tasks a fraction of a
millisecond later but the first task was already running.  If you want to
submit many tasks at once then you might consider using the dask.delayed
interface.

SultanOrazbayev:
> Probably the scheduler just got the first task first and so started running it right away. The scheduler then got the subsequent tasks a fraction of a millisecond later but the first task was already running.

OK, thanks! Just to make sure I understand it correctly, with futures it's not possible to affect the order of execution? Changing priority sign in the code below gives the same execution order (in the order of submission):

```python
from dask.distributed import Client, LocalCluster
cluster = LocalCluster(n_workers=1, resources={'foo': 1})
client = Client(cluster)

# a small delay is added to check if the scheduler needs more time to accept all tasks before sorting them
def test_func(i):
    import time
    time.sleep(1+i)
    return i

# executed in the order of submission (priorities are in the same order)
futures = [client.submit(test_func, i, priority=-i, resources={'foo': 1}) for i in range(5)]

# also executed in the order of submission, even though priorities of later tasks are higher
futures = [client.submit(test_func, i, priority=i, resources={'foo': 1}) for i in range(5)]

```

mrocklin:
You can affect the order of execution, but Dask will also start working on
things right away.  Dask won't stop things that have already started
running.

SultanOrazbayev:
Hmm, after removing `resources`, the order of execution is affected by the priorities (so things run as expected). Not sure if this is a feature.
```python
# this executes in the order of priorities (long-running task first)
futures = [client.submit(test_func, i, priority=i) for i in range(5)]
```

SultanOrazbayev:
I got confused by the negative sign in the task dashboard, but right now priorities work as expected. The only problem I have is combining resources and priority tags, but this is more a matter of understanding how to use resources, so I opened a discussion here: https://github.com/dask/distributed/discussions/4209

mrocklin:
If you want to submit a small patch negating the priority again in the info
page that would be welcome.

https://github.com/dask/distributed/blob/cee4e3c99c34c1e515b9e80032d19c24b5f390ed/distributed/http/templates/task.html#L42-L45

PR #4190, opened 2020-10-27
Add to FAQ section about how to get workers to import right

martindurant:
Fixes https://github.com/dask/dask/issues/6702

This section may now be too long to fit int he FAQ.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -6,6 +6,44 @@ More questions can be found on StackOverflow at http://stackoverflow.com/search?
 How do I use external modules?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Most functions you will run on your cluster will require imports. This
+is even true for passing any object which is not a python builtin -
+the pickle serialisation method will save references to imported modules
+rather than trying to send all of your source code.
+
+You therefore must ensure that workers have access to all of the modules
+you will need, and ideally with exactly the same versions.
+
+Maintain consistent environments
+````````````````````````````````
+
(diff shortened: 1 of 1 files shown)


martindurant:
By the way, does [skip ci] still work for docs-only PRs?

quasiben:
Do you think we should also mention the [PipInstall plugin](https://github.com/dask/distributed/pull/3216/) as a temporary solution ?

martindurant:
> we should also mention the PipInstall plugin

Probably - I had missed that. I did *not* want to say "do `client.run(subprocess.call())`. Nice to see considerations of restarts. However, the plugin isn't mentioned in any docs (yet).

mrocklin:
distributed.dask.org gets far less traffic than docs.dask.org 

I wonder if this would be better suited as a new deployment best practices page on docs.dask.org

martindurant:
@mrocklin , that being the case, we should consider more information that might be more visible if moved; but not in this PR...

mrocklin:
Yeah, I don't have strong thoughts on what goes into this FAQ file.  My preference in the future is to remove it.  (I think that FAQs should be fairly comprehensive or else not exist).  If your goal is to solve an education problem then I don't think that editing docs here will do a ton of good.  I think that starting a best practices doc in dask/dask would have high value but yes, I agree that this isn't necessary.

mrocklin:
Should we still consider merging this, or are we relying on the dask/dask PR?

martindurant:
I think the one on dask/dask is better, and is marked as closing this one. But I'll close this now anyway.

mrocklin:
Thanks for both Martin!

Issue #4188, opened 2020-10-26
TaskState metadata


jrbourbeau:
Recently I was working with a group that computed user-defined attributes for each task (e.g. the start and stop time when running the task) and stored them in an external database. However, after https://github.com/dask/distributed/pull/4107 we now have a `TaskState` object on both the worker and scheduler which could be used to store custom task-level information directly on the cluster.

For example, we could add a new `TaskState.metadata` mapping to both the scheduler and worker `TaskState` classes and synchronize them when `Worker.send_task_state_to_scheduler` is called. This would give users an option to store rich information relevant to their particular workloads

This seems like useful functionality to me and I wanted to see if other groups may also benefit from adding task-level metadata

mrocklin:
Thinking again about this this might be simple enough to implement that it might make sense to just create a PR.  It might make things more clear and I'm guessing it's only a few lines.

quasiben:
I think this would be great!  It'd be cool to optionally store timeseries info about tasks in addition to the aggregate behavior we currently store now.  Specifically, with this kind of tooling we may be able to resolve https://github.com/dask/distributed/issues/4159

jrbourbeau:
I opened up a PR to implement this over in https://github.com/dask/distributed/pull/4191

PR #4191, opened 2020-10-27
Add TaskState metadata

jrbourbeau:
This PR adds a new `.metadata` attribute to the `TaskState` classes used by the scheduler and workers. When a task is finished the metadata stored on worker `TaskState`s synchronized with the corresponding `TaskState` on the scheduler   

Closes https://github.com/dask/distributed/issues/4188

### DIFF
--- a/distributed/http/templates/task.html
+++ b/distributed/http/templates/task.html
@@ -43,7 +43,7 @@
               <th> Priority </th>
               <td>{{ts.priority}}</td>
           </tr>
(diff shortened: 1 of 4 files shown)


mrocklin:
This seems pretty straightforward to me.  +1

mrocklin:
If you wanted to extend things very slightly it might also be interesting to add `metadata` to the task info pages if metadata is present

jrbourbeau:
Good idea! Just pushed a commit which does that 

Also, cc @quasiben as you expressed an interest in this over in https://github.com/dask/distributed/issues/4188

jrbourbeau:
Thanks for reviewing @mrocklin! 

(for reference, `test_broken_worker_during_computation` is a known flaky test https://github.com/dask/distributed/issues/4173)

PR #4192, opened 2020-10-27
Send active task durations from worker to scheduler

jrbourbeau:
This PR updates the worker's heartbeat to the scheduler to include the duration of actively running tasks. This information can help us make more informed scheduling decisions when there's a discrepancy between the expected task duration (stored on the corresponding `TaskPrefix.duration_average`) and the actual duration of a running task. For example, this situation can arise in work stealing scenarios when a task takes much longer than expected based on the average task duration. Additionally, sending this information to the scheduler via worker heartbeat helps avoid significant message-per-task overhead. 

cc @mrocklin @sheer-coiled for thoughts

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -126,6 +126,10 @@ class TaskState:
         The number of times a dependency has not been where we expected it
     * **startstops**: ``[{startstop}]``
         Log of transfer, load, and compute times for a task
+    * **start_time**: ``float``
+        Time at which task begins running
+    * **stop_time**: ``float``
+        Time at which task finishes running
 
     Parameters
     ----------
@@ -156,6 +160,8 @@ class TaskState:
         self.type = None
         self.suspicious_count = 0
         self.startstops = list()
+        self.start_time = None
+        self.stop_time = None
 
     def __repr__(self):
         return "<Task %r %s>" % (self.key, self.state)
@@ -763,6 +769,7 @@ class Worker(ServerNode):
         return self.local_directory
 
     async def get_metrics(self):
+        now = time()
         core = dict(
             executing=self.executing_count,
             in_memory=len(self.data),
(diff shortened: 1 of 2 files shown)


distributed/worker.py
```diff
@@ -773,6 +780,10 @@ async def get_metrics(self):
                 "workers": dict(self.bandwidth_workers),
                 "types": keymap(typename, self.bandwidth_types),
             },
+            active_durations={
+                key: now - self.tasks[key].start_time
+                for key in self.active_threads.values()
+            },
```
mrocklin, line 786: Maybe we should reclaim the term `executing` from above?  That seems to be derivative state from this

fjetter:
Another situation where this might be useful is if the submitted tasks are very long running but the cluster hasn't seen them before. In the current situation this forces users to configure some estimation of a tasks runtime in the configuration to get reasonable responsive scaling behaviour using the `Scheduler.adaptive_target`. This would require an additional change to the `WorkerState.processing` dict, though.

mrocklin:
> Another situation where this might be useful is if the submitted tasks are very long running but the cluster hasn't seen them before. In the current situation this forces users to configure some estimation of a tasks runtime in the configuration to get reasonable responsive scaling behaviour using the Scheduler.adaptive_target. This would require an additional change to the WorkerState.processing dict, though.

Agreed.  I think/hope that this should remove most of the need for the config values in the future.

jrbourbeau:
That's a great point @fjetter, I totally agree with you. I suspect there are a few situations where we can use this duration information to make more informed scheduling decisions. This PR is just a first step to get active task durations to the scheduler

jrbourbeau:
Not sure why `test_close_gracefully` failed as the changes here seem unrelated. All CI builds passed after pushing an empty commit, so I'm inclined to mark it as a flaky test and merge this (I opened up https://github.com/dask/distributed/issues/4201 to track the test failure)

PR #4197, opened 2020-10-28
Don't inherit from WorkerPlugin class

jsignell:
It seems like WorkerPlugins should not inherit from the WorkerPlugin class, but just use it as a spec. Otherwise I don't think:

https://github.com/dask/distributed/blob/24007c2b81e12d98fb229504f858c8f28b6063ba/distributed/worker.py#L1158-L1162

can work if the user hasn't defined a teardown method.

### DIFF
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -185,7 +185,7 @@ class WorkerPlugin:
         """
 
 
-class PipInstall(WorkerPlugin):
+class PipInstall:
     """A Worker Plugin to pip install a set of packages
 
     This accepts a set of packages to install on all workers.
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -104,9 +104,12 @@ class WorkerPlugin:
     the plugin to your client in order to have it attached to every existing and
     future workers with ``Client.register_worker_plugin``.
 
+    Note: Do not inherit from this class. Just use it as a specification and implement
+    the methods that you need.
+
     Examples
     --------
-    >>> class ErrorLogger(WorkerPlugin):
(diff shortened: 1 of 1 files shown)


mrocklin:
I'm not sure that I understand.  Can you explain why this would fail?

jsignell:
Yeah sorry, if we inherit from the class `WorkerPlugin` then the subclass will have attr `teardown` even if it's not implemented. And it'll try to run `plugin.teardown(self)` you get an error:

```python-traceback
<ipython-input-16-486128a0031d> in <module>
----> 1 plugin.teardown()

TypeError: teardown() missing 1 required positional argument: 'worker'
``

But maybe this isn't the solution.

jrbourbeau:
I think in that all the methods are no-ops on the base `WorkerPlugin` class. So there won't be any breakage, just a few extra method calls that don't do anything

jsignell:
You don't get any errors in the worker logs?

jsignell:
Maybe the right thing is to make worker optional in the `WorkerPlugin` class.

mrocklin:
@jsignell maybe the right thing to do here is to provide a failing test case first?  That might help to guide development and conversation.

jsignell:
Oh! I am not thinking about this right at all. Sorry for the noise. I got confused about what self meant in that snippet above (hint - it means worker :facepalm: ).

jsignell:
What I was doing is I kept thinking that `register_worker_plugin` accepted a class rather than an instantiated object. And then I'd have all sorts of errors about self being missing...

jsignell:
Thank you both for engaging with me on this.

mrocklin:
Yeah, no sweat.  Good luck

PR #4198, opened 2020-10-28
Instantiate plugin if needed in register_worker_plugin

jsignell:
From the conversation in #4197 I realized that if you register a plugin wrong then that name is used up and you cannot register another plugin with the same name. This PR is just a little earlier catch to try to protect people from that. If this seems like overkill that is fine and we can close, just wanted to propose it since it would have helped me figure out what was going on sooner.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4223,6 +4223,9 @@ class Client:
         --------
         distributed.WorkerPlugin
         """
+        if isinstance(plugin, type):
+            raise ValueError("Plugin should be an instantiated object.")
+
         return self.sync(self._register_worker_plugin, plugin=plugin, name=name)
 
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4160,7 +4160,7 @@ class Client:
                 raise exc.with_traceback(tb)
         return responses
 
-    def register_worker_plugin(self, plugin=None, name=None):
+    def register_worker_plugin(self, plugin=None, name=None, **kwargs):
         """
         Registers a lifecycle worker plugin for all current and future workers.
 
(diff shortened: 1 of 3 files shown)


jrbourbeau (commented):
Thanks for the PR @jsignell! Alternatively, perhaps we could just instantiate the plugin if users pass in a class instead of an instance? This is what we do for scheduler plugins:

https://github.com/dask/distributed/blob/24007c2b81e12d98fb229504f858c8f28b6063ba/distributed/scheduler.py#L2795-L2796

I don't have strong thoughts either way. What do you think?

jsignell:
> Alternatively, perhaps we could just instantiate the plugin if users pass in a class instead of an instance? 

Yes I love this! Don't know how I missed it :facepalm: I'll change this PR to do that instead.

jsignell:
@jrbourbeau I think this is good to go.

distributed/client.py
```diff
@@ -4189,6 +4189,9 @@ def register_worker_plugin(self, plugin=None, name=None):
         name: str, optional
             A name for the plugin.
             Registering a plugin with the same name will have no effect.
+        **kwargs: optional
+            If you do pass an class name as the plugin, it will be instantiated
+            with any extra keyword arguments.
```
jrbourbeau, line 4194: A small suggestion on this parameter description, but otherwise this LGTM 

```suggestion
            If you pass a class as the plugin, instead of a class instance, then the
            class will be instantiated with any extra keyword arguments.
```

jsignell: Thanks!

jrbourbeau (approved):
Thanks @jsignell!

PR #4204, opened 2020-10-30
(TEST) Apply Florian's patch to 2.30.x

TomAugspurger:
Trying out https://github.com/dask/distributed/pull/4176 against 2.30 to check if the failures on that PR are only present on master. 

If this passes, I don't think we merge this PR. Instead, we should merge https://github.com/dask/distributed/pull/4176 despite the failing tests and then just backport the merged PR to 2.30.x. That should play more nicely with versioneer.

### DIFF
--- a/continuous_integration/environment-windows.yml
+++ b/continuous_integration/environment-windows.yml
@@ -17,7 +17,7 @@ dependencies:
   - prometheus_client
   - psutil
   - pytest
-  - pytest-asyncio
+  - pytest-asyncio<0.14.0
   - pytest-repeat
   - pytest-timeout
   - pytest-faulthandler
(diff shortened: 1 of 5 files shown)


jrbourbeau:
I opened a separate PR over in https://github.com/dask/distributed/pull/4212 to see what test failures we currently have in the main branch and saw there are many failures on both Windows and Travis CI builds. Lots of the test failures (maybe all?) look the same as the ones we're seeing in this PR, but it's difficult to directly compare the builds when there are > 20 failures.

I ended up not releasing 2.30.1 last Friday as, I think, the Travis builds weren't completed late in the evening and then on Saturday I saw more tests were failing than expected. Given the current state of our test suite, I'm inclined to not release right now and instead focus on fixing tests. Thoughts @TomAugspurger?

Also, I can cross-post this over in the release issue on the community tracker

TomAugspurger:
> Given the current state of our test suite, I'm inclined to not release right now and instead focus on fixing tests. Thoughts @TomAugspurger?

Agreed.

jrbourbeau:
Woo, tests pass 🎉 I'll carry on with releasing 2.30.1

TomAugspurger:
Great, thanks!

PR #4211, opened 2020-11-01
Display priority values as specified by user (removing the negative sign)

SultanOrazbayev:
Only the first element in the tuple is negated (it contains the user-specified priority), other components are unchanged.

This is a follow-up from https://github.com/dask/distributed/pull/4185

### DIFF
--- a/distributed/http/templates/task.html
+++ b/distributed/http/templates/task.html
@@ -41,7 +41,7 @@
           {% end %}
           <tr>
               <th> Priority </th>
-              <td>{{ts.priority}}</td>
+              <td>{{(-ts.priority[0],*ts.priority[1:])}}</td>
           </tr>
           {% for attr in ['has_lost_dependencies', 'host_restrictions', 'worker_restrictions', 'resource_restrictions', 'loose_restrictions', 'suspicious', 'retries', 'metadata'] %}
           {% if getattr(ts, attr) %}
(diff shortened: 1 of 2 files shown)


jrbourbeau (commented):
Thanks for the PR @SultanOrazbayev. Apologies for the CI build failures, they are unrelated to the changes in this PR. I've merged the main branch into this branch which should fix the test failures (hope that's okay)

mrocklin:
> Only the first element in the tuple is negated (it contains the user-specified priority), other components are unchanged.

Hrm, that's interesting.  We probably can't negate only one element, otherwise the presented value doesn't actually represent the priority.  We probably have to negate them all or none.

SultanOrazbayev:
Uh oh. I thought just the first element should be negated because of these lines:

https://github.com/dask/distributed/blob/debe1b80e74e43ed88001c40a9cf07cfefa2fc47/distributed/scheduler.py#L1967-L1970

In the dashboard right now (without this commit), the negative sign appears only on the user-specified values... so I thought this would fix it.

mrocklin:
I suspect that the other priorities are just genuinely negative values.  It
doesn't matter so much to the system if we use negative or positive
values.  It matters if things are higher or lower than each other.  By
reversing just some of these values it throws things out of alignment

PR #4212, opened 2020-11-02
Pin pytest-asyncio version

jrbourbeau:
Checking whether or not the Travis CI failures in https://github.com/dask/distributed/pull/4204 are related to the changes there

### DIFF
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -34,7 +34,7 @@ jobs:
         shell: bash -l {0}
         run: |
           if [[ "${{ matrix.python-version }}" = "3.8" ]]; then
-            conda install -c conda-forge tornado=6,<6.1
+            conda install -c conda-forge "tornado=6,<6.1"
           else
             conda install -c conda-forge tornado=5
           fi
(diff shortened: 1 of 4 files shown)


jrbourbeau:
In its latest release, `pytest-asyncio` began modifying the default asyncio `event_loop_policy` which broke tests here as we need to explicitly set our event loop policy

https://github.com/dask/distributed/blob/cee4e3c99c34c1e515b9e80032d19c24b5f390ed/distributed/utils.py#L1237

There is a discussion of how to proceed on the `pytest-asyncio` side (xref https://github.com/pytest-dev/pytest-asyncio/issues/168) but for now we should just set an upper limit on our allowed version of `pytest-asycnio`

jrbourbeau:
Tests on Windows (mostly) pass 🎉 I've been `test_tls_specific` fail before on `master` (xref https://github.com/dask/distributed/runs/1332842736#step:8:116) and opened up an issue for tracking the failure (https://github.com/dask/distributed/issues/4215)

TomAugspurger:
Thanks.

We'll want to backport this to 2.30.x as well once it's merged.

jrbourbeau:
Merging as the two test failures are known flaky tests

PR #4213, opened 2020-11-02
Better task duration estimates for outliers.

selshowk:
This PR implements more adaptive worker duration estimates for "outlier" tasks (defined as those whose on-going runtime exceeds twice the current task average duration).  Rather than use the average duration as an estimate, we estimate 2x the current runtime if the task is an outlier.  See discussion in #4192 as well.

@jrbourbeau I had to add an additional check to the edits from #4192 because otherwise my test below generated a missing key error.  Can you comment if it makes sense or suggestions a deeper problem?

CCing @jrbourbeau @mrocklin for thoughts.


distributed/tests/test_worker.py
```diff
@@ -1715,6 +1714,68 @@ def bad_heartbeat_worker(*args, **kwargs):
...
+                        if td < sleep_time:
+                            await r
+                    for ts in ws.processing:
+                        #logger.info(f"***************************************************************")
+                        #logger.info(f"TIMES ***  {time() - t0}  t={sleep_time}, avg[{ts}] = {ws.processing[ts]}")
+                        #logger.info(f"***************************************************************")
+                        expected_duration = ws.processing[ts]
+                        # expected duraction should be close to d0
+                        assert expected_duration > (1-tolerance) * exp_avg
+                        assert expected_duration < (1+tolerance) * exp_avg
```
mrocklin, line 1777: I think that we can test this more simply, and with less reference to internal API (most Dask distributed tests try very hard not to reference the internal API, which gives us freedom to change that API smoothly in the future).  In general we want to test behaviors, not state.  In this case the behavior that we want to test is, I think, that Dask does load balancing even when the estimate is very wrong.

Here is maybe a short attempt at a test.

```python
@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)] * 2)
def test_balance_with_longer_task(c, s, a, b):
    await client.submit(slowinc, 0, delay=0)  # scheduler learns that slowinc is very fast

    x = client.scatter(np.arange, 100000, workers=[a.address])  # put some weight on a
    y = client.submit(slowinc, 1, delay=5, workers=[a.address])  # a surprisingly long task
    z = client.submit(inc, x, workers=[a.address], allow_other_workers=True)  # a task after y, suggesting a, but open to b

    await z
    assert z.key in b.data
```

I *think* that this will fail today (it would be good to be sure)

distributed/scheduler.py
```diff
@@ -5597,6 +5613,9 @@ def heartbeat_interval(n):
         return n / 200 + 1
 
 
+
+
+
```
mrocklin, line 5618: You'll need to install the git pre-commit hooks.  See https://docs.dask.org/en/latest/develop.html#code-formatting

distributed/scheduler.py
```diff
@@ -4068,6 +4068,28 @@ def decide_worker(self, ts):
...
+        duration = self.get_task_duration(ts)
+        comm = self.get_comm_cost(ts, ws)
+        new_duration = duration + comm
+        exec_time = -1
+        if 'executing' in ws.metrics and ts.key in ws.metrics['executing']:
+            exec_time = ws.metrics['executing'][ts.key]
+        if support_outliers and exec_time > threshold * duration:
+            new_duration = 2 * exec_time
+        ws.processing[ts] = new_duration 
+        return ws.processing[ts]
```
mrocklin, line 4091: Some thoughts here:

-  I wonder if we should place `executing` on the WorkerState class directly, rather than including it in the catch-all metrics attribute.  
-  Until we decide to generalize things, I'd prefer to assume values of `support_outliers` and `threshold`.  I don't think that we should generalize pre-maturely.  I suspect that by inlining these this code will become a bit simpler and so be easier for future maintainers to quickly understand.

Thoughts?

distributed/worker.py
```diff
@@ -784,7 +784,8 @@ async def get_metrics(self):
             },
             executing={
                 key: now - self.tasks[key].start_time
-                for key in self.active_threads.values()
+                for key in self.active_threads.values() 
+                if key in self.tasks
```
mrocklin, line 788: Ah, an interesting addition.  That makes sense.  I'm curious, how did this issue present?

distributed/scheduler.py
selshowk: For `executing`, I agree it makes sense though I'm not familiar enough with that code to speak meaningfully about it.

For the second point, I had considered the possibility that those parameters would become configuration options so users could tweak them for weird cases.  I was initially thinking of the possibly of fat gaussians (where the variance is larger than the mean) but then I realized that can't happen with purely positive data.  Given that, its probably fine to hard-code default values.

I still think the having a separate function for this makes sense though since the duration is computed in two places.  It will also make it easier to change the logic later if needed and I would think the function name makes the code more navigable.

mrocklin: > For executing, I agree it makes sense though I'm not familiar enough with that code to speak meaningfully about it.

The relevant code on the scheduler side is here:

https://github.com/dask/distributed/blob/cee4e3c99c34c1e515b9e80032d19c24b5f390ed/distributed/scheduler.py#L1614-L1623

We might add an `executing={}` dict to that signature

The relevant code on the worker side is in two places

https://github.com/dask/distributed/blob/cee4e3c99c34c1e515b9e80032d19c24b5f390ed/distributed/worker.py#L911-L916

https://github.com/dask/distributed/blob/cee4e3c99c34c1e515b9e80032d19c24b5f390ed/distributed/worker.py#L785-L788

I would remove the executing bit from get_metrics, and send it as a direct parameter to `heartbeat_worker`.  I would then modify the WorkerState object on the Scheduler directly.

This could be a separate PR, or we could just roll it into this one if that's simpler (personally I would probably just roll it in).

mrocklin: > For the second point, I had considered the possibility that those parameters would become configuration options so users could tweak them for weird cases. I was initially thinking of the possibly of fat gaussians (where the variance is larger than the mean) but then I realized that can't happen with purely positive data. Given that, its probably fine to hard-code default values.

Yeah, I think that these are valid points, and something that we might want to do in the future.  In general though there are a *lot* of different things that we've said we might want to do and never did.  If we generalized things when possible like this then I think that the codebase would quickly become unreadable.  Something I've learned after a while of doing this is to optimize strongly for future reviewer readability.  If we can make this a five-line function with only one conditional at the cost of generality then I think that that is a clear win.

jrbourbeau: +1 for holding off on generalizing

distributed/worker.py
jrbourbeau: Thanks for fixing this @selshowk. I'd also be curious to know what failure prompted this addition

distributed/scheduler.py
selshowk: Ok sounds good but do you agree that it makes sense to keep it a separate function rather than inline?

distributed/worker.py
selshowk: I'm trying to reproduce it but am not able to.  It might have been with a different version of the test function which didn't work so never got committed.  I'll try a bit more to reproduce it.

distributed/scheduler.py
mrocklin: Separate function seems fine to me

pentschev:
We run some Distributed tests that require CUDA support in dask-cuda, and I just found out that the failures we see there are due to `pytest-asyncio=0.14.0` and the most recent was `pytest-asyncio=0.12.0` until 3 days ago in Anaconda: https://anaconda.org/conda-forge/pytest-asyncio/files .

Locally I can run the tests from dask-cuda by just reverting to `pytest-asyncio=0.12.0`, and I noticed the builds here are picking `pytest-asyncio=0.14.0`, in contrast to `pytest-asyncio=0.12.0` in https://github.com/dask/distributed/pull/4211 where builds are passing. I'm not yet sure if there's something that needs to be changed in Distributed or if there's a bug in `pytest-asyncio=0.14.0`.

pentschev:
Only saw that older version of `pytest-asyncio` was pinned in https://github.com/dask/distributed/pull/4212, maybe running tests here again will fix failures?

jrbourbeau:
Yeah, running tests again here should pick up the changes in https://github.com/dask/distributed/pull/4212, which will fix test failures (up to a few known flaky tests)

distributed/tests/test_worker.py
jrbourbeau: @selshowk and I were able to talk about this test offline on Friday (summarizing the output of that conversation here). 

There are a few different [types of priorities that are used to determine the order in which tasks are executed](https://distributed.dask.org/en/latest/priority.html) (e.g. user-provided priority, priority based on graph structure, etc.). For tasks which are submitted rapidly, and all other priorities are the same, we use the order in which tasks arrive at a worker as a tie breaker. This tie breaking is implemented using a last-in-first-out policy.

So in the above test, `z` will actually start running before `y` (due to LIFO tie breaking). To ensure `y` starts running before `z`, we can specify our own user-provided priority: 

```python
@gen_cluster(client=True)
async def test_balance_with_longer_task(c, s, a, b):
    np = pytest.importorskip("numpy")

    await c.submit(slowinc, 0, delay=0)  # scheduler learns that slowinc is very fast
    x = await c.scatter(np.arange(10000), workers=[a.address])
    y = c.submit(
        slowinc, 1, delay=5, workers=[a.address], priority=1
    )  # a surprisingly long task
    z = c.submit(
        inc, x, workers=[a.address], allow_other_workers=True, priority=0
    )  # a task after y, suggesting a, but open to b
    await z
    assert z.key in b.data
```

> I think that this will fail today (it would be good to be sure)

I can confirm this fails on the main branch, but passes here

jrbourbeau:
@selshowk I pushed a couple of commits to resolve some merge conflicts and update the test added here (hope that's okay). Let me know what you think

distributed/scheduler.py
```diff
@@ -286,6 +292,7 @@ class WorkerState:
     _actors: set
     _address: str
     _bandwidth: double
+    _executing: dict
```
jrbourbeau, line 295: We're adding a new attribute to the `WorkerState` class which we recently added type annotations to for Cythonization. @jakirkham I think I took care of everything needed for this addition, but is there some kind of check I can run to make sure Cython is happy? For instance, is being able to successfully build the C-extensions a sufficient check?

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
 
         return worker
 
+    def set_duration_estimate(self, ts, ws):
```
jakirkham, line 4419:
```suggestion
    def set_duration_estimate(self, ts: TaskState, ws: WorkerState):
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+        """Estimate task duration using worker state and task state.
+
+        If a task takes longer than twice the current average duration we
+        estimate the task duration to be 2x current-runtime, otherwise we set it
+        to be the average duration.
+        """
+        duration = self.get_task_duration(ts)
+        comm = self.get_comm_cost(ts, ws)
+        total_duration = duration + comm
+        if ts in ws.executing:
```
jakirkham, line 4429:
```suggestion
        if ts in ws._executing:
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+            exec_time = ws.executing[ts]
```
jakirkham, line 4430:
```suggestion
            exec_time = ws._executing[ts]
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+            if exec_time > 2 * duration:
+                total_duration = 2 * exec_time
+        ws.processing[ts] = total_duration
```
jakirkham, line 4433:
```suggestion
        ws._processing[ts] = total_duration
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+        return ws.processing[ts]
```
jakirkham, line 4434:
```suggestion
        return ws._processing[ts]
```

jakirkham (commented):
Mostly things look good on the Cython front. Just a couple suggestions here on the new method (assuming we keep that).

distributed/scheduler.py
jakirkham: Thanks James! Looks like everything is good so far.

Had a few comments on the new method. IDK if we plan to keep that though or not (if not they can just serve as an example for new functions).

jakirkham: Oh sorry missed your second question. We build with Cython on the Python 3.7 GitHub Action job and run the tests there. If they pass, we should be good 🙂

mrocklin:
This brings up an interesting question about how we maintain Cython
performance.  The tests will make sure that things are valid, but I
wouldn't be surprised if we slip a little with each PR, especially among
those contributed by drive-by contributors.

I suspect that for the near-term we'll want to do little Cython "tune-ups"
before release, and that long-term we may want to see if some sort of
Cython coverage tool, perhaps one that uses the annotated HTML output, is
possible.

mrocklin:
This isn't something to think about too much right now (let's make sure
that Cython is a good idea first), but it's a fun conversation to hold for
the future.

jakirkham:
Well I know @quasiben did some work setting up benchmarks that run nightly. I don't think we can promise to run them on every PR. Maybe we can eyeball those and see how things are going? Perhaps Ben has thoughts on these sorts of things 😉

Generally I think writing code that Cython can do well with is not difficult. It's really a matter of annotating variables with types and sticking to them. This has much less to do with Cython and more generally being thoughtful while programming. Would describe this similarly as to how we advise people to write code that will run well with Dask. First think about how it can be written to parallelize well and then running with Dask isn't so hard. Cython isn't too different 🙂

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
         return worker
 
+    def set_duration_estimate(self, ts: TaskState, ws: WorkerState):
+        """Estimate task duration using worker state and task state.
+
+        If a task takes longer than twice the current average duration we
+        estimate the task duration to be 2x current-runtime, otherwise we set it
+        to be the average duration.
+        """
+        duration = self.get_task_duration(ts)
```
jakirkham, line 4426:
```suggestion
        duration: double = self.get_task_duration(ts)
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+        comm = self.get_comm_cost(ts, ws)
```
jakirkham, line 4427:
```suggestion
        comm: double = self.get_comm_cost(ts, ws)
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+        total_duration = duration + comm
```
jakirkham, line 4428:
```suggestion
        total_duration: double = duration + comm
```

distributed/scheduler.py
```diff
@@ -4396,6 +4416,23 @@ def decide_worker(self, ts):
...
+        if ts in ws._executing:
+            exec_time = ws._executing[ts]
```
jakirkham, line 4430:
```suggestion
            exec_time: double = ws._executing[ts]
```

jakirkham (commented):
Guessing the numbers `set_duration_estimate` is working with are floating point. If so, we can do a little more annotation through here. This will get us some straight C code for computation when Cythonized.

Skipped `transition_waiting_processing` as I'm planning to go back over that one in detail later.

jakirkham:
Took the liberty of fixing conflicts with `master` (pretty minor changes really), hope that is ok 🙂

distributed/scheduler.py
```diff
@@ -512,7 +525,8 @@ def clean(self):
             extra=self._extra,
         )
         ts: TaskState
-        ws._processing = {ts._key: cost for ts, cost in self._processing.items()}
+        ws._processing = {ts.key: cost for ts, cost in self._processing.items()}
+        ws._executing = {ts.key: duration for ts, duration in self._executing.items()}
```
jakirkham, line 529: This fetches the C-level attributes of `TaskState` when using Cython, which can be quite a bit faster.

```suggestion
        ws._processing = {ts._key: cost for ts, cost in self._processing.items()}
        ws._executing = {ts._key: duration for ts, duration in self._executing.items()}
```

jrbourbeau: Thanks @jakirkham!

distributed/scheduler.py
```diff
@@ -5976,10 +6009,7 @@ def _reevaluate_occupancy_worker(self, ws: WorkerState):
         new = 0
         nbytes = 0
         for ts in ws._processing:
-            duration = self.get_task_duration(ts)
-            comm = self.get_comm_cost(ts, ws)
-            ws._processing[ts] = duration + comm
```
jakirkham, line 5981: Do we still want to set this?

distributed/scheduler.py
```diff
@@ -4710,14 +4747,10 @@ def transition_waiting_processing(self, key):
                 return {}
             worker = ws._address
 
-            duration = self.get_task_duration(ts)
-            comm = self.get_comm_cost(ts, ws)
-            occupancy = duration + comm
-
-            ws._processing[ts] = occupancy
```
jakirkham, line 4717: Same question as above

distributed/scheduler.py
jrbourbeau: Yes we still want to set `ws._processing`, but it's been moved into a `set_duration_estimate` method:

https://github.com/dask/distributed/blob/308c4df69e250d90a159885fd566d253228b48a7/distributed/scheduler.py#L4714-L4729

jrbourbeau (approved):
Thanks for your work on this @selshowk @jakirkham! 

(FWIW the failures observed on Travis CI have been seen on other PRs and are unrelated to the changes here)

jakirkham:
Thanks James! 😄 I think @selshowk did all the real work here. Just tried to keep merge conflicts to a minimum 😉

PR #4216, opened 2020-11-03
Add get_task_metadata context manager

jrbourbeau:
In https://github.com/dask/distributed/pull/4191 we added a new `.metadata` attribute to the `TaskState` classes used by the scheduler and workers. However, after a tasks have been computed and are no longer needed, the corresponding `TaskState` objects are removed from the scheduler / workers. This makes it difficult to collect custom task-level metadata in practice.

This PR adds a new `get_task_metadata` context manager to collect `TaskState` metadata within a given context (similar to our existing `get_task_stream` and `get_performance_report` context managers). This enables users to do something like:

```python
>>> from distributed import get_task_metadata
>>> with get_task_metadata() as tasks:
...     x.compute()
>>> tasks.metadata
{...}
```

to get a key-to-metadata mapping for all the tasks which were submitted and completed within the `get_task_metadata` context.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3876,7 +3876,7 @@ class Scheduler(ServerNode):
 
         plugin = plugins[0]
         self.remove_plugin(plugin)
-        return plugin.metadata
+        return {"metadata": plugin.metadata, "state": plugin.state}
 
     async def register_worker_plugin(self, comm, plugin, name=None):
         """ Registers a setup function, and call it on every worker """
@@ -5667,6 +5667,7 @@ class CollectTaskMetaDataPlugin(SchedulerPlugin):
         self.name = name
         self.keys = set()
         self.metadata = {}
+        self.state = {}
 
     def update_graph(self, scheduler, dsk=None, keys=None, restrictions=None, **kwargs):
         self.keys.update(keys)
@@ -5676,4 +5677,5 @@ class CollectTaskMetaDataPlugin(SchedulerPlugin):
             ts = self.scheduler.tasks.get(key)
             if ts is not None and ts.key in self.keys:
                 self.metadata[key] = ts.metadata
+                self.state[key] = finish
                 self.keys.discard(key)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1363,6 +1363,8 @@ class Scheduler(ServerNode):
             "adaptive_target": self.adaptive_target,
             "workers_to_close": self.workers_to_close,
             "subscribe_worker_status": self.subscribe_worker_status,
+            "start_task_metadata": self.start_task_metadata,
+            "stop_task_metadata": self.stop_task_metadata,
         }
 
         self._transitions = {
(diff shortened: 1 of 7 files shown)


distributed/scheduler.py
```diff
@@ -3855,6 +3857,27 @@ def get_task_stream(self, comm=None, start=None, stop=None, count=None):
...
+        plugin = CollectTaskMetaDataPlugin(scheduler=self, name=name)
+
+        self.add_plugin(plugin)
+
+    def stop_task_metadata(self, comm=None, name=None):
+        plugins = [
+            p
+            for p in self.plugins
+            if isinstance(p, CollectTaskMetaDataPlugin) and p.name == name
+        ]
```
mrocklin, line 3870: At some point we should consider turning `self.plugins` into a dict

distributed/scheduler.py
```diff
@@ -5636,3 +5659,21 @@ def remove_worker(self, worker=None, **kwargs):
...
+        self.scheduler = scheduler
+        self.name = name
+        self.keys = set()
+        self.metadata = {}
+
+    def update_graph(self, scheduler, dsk=None, keys=None, restrictions=None, **kwargs):
+        self.keys.update(keys)
+
+    def transition(self, key, start, finish, *args, **kwargs):
+        if start == "processing" and (finish == "memory" or finish == "erred"):
```
mrocklin, line 5675: Consider also `long-running`.  I think that it might make sense to do this for all tasks that end up in a finished state, regardless of their start state.

distributed/scheduler.py
```diff
@@ -5636,3 +5659,21 @@ def remove_worker(self, worker=None, **kwargs):
...
+            ts = self.scheduler.tasks.get(key)
+            if ts is not None and ts.key in self.keys:
+                self.metadata[key] = ts.metadata
```
mrocklin, line 5678: Are there other parts of the `TaskState` state that would be valuable to include?

mrocklin:
I had a couple of small comments but they are small.  This looks good to me.

distributed/scheduler.py
jrbourbeau: IIUC long running tasks on a worker are kept in the `processing` state on the scheduler, so this should include those tasks. But I agree with your larger point, this should handle all finished tasks regardless of their start state

distributed/scheduler.py
jrbourbeau: +1, but I'd prefer to include that in a separate PR

distributed/scheduler.py
mrocklin: Ah!  I just checked and you're right.  My information is out-dated here.

distributed/scheduler.py
jrbourbeau: That's a good point. The `finish` state would also be useful to have.

jrbourbeau:
Thanks for reviewing @mrocklin!

PR #4225, opened 2020-11-06
Allow actors to call actors on the same worker

### LINKED ISSUE #4224: Actors on same worker cause deadlock
Consider
```python
client = dask.distributed.Client()

class Minimal:
    
    def thetime(self):
        import time
        return time.time()

class UsesMinimal():
     
     def getthetime(self, ac):
         return ac.thetime().result()
```

The following works
```python
ac = client.submit(Minimal, actor=True, workers=["127.0.0.1:61047"]).result()
ac2 = client.submit(UsesMinimal, actor=True, workers=['tcp://127.0.0.1:61046']).result()
ac2.getthetime(ac).result()
```

but if they are on the same worker
```python
ac2 = client.submit(UsesMinimal, actor=True, workers=['tcp://127.0.0.1:61047']).result()
ac2.getthetime(ac).result()
```
... deadlock

- Dask version: master (dask/distributed)
- Python version: 3.7
- Operating System: osx
- Install method (conda, pip, source): source

martindurant:
Fixes #4224

### DIFF
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -119,14 +119,18 @@ class Actor(WrappedKey):
         return sorted(o)
 
     def __getattr__(self, key):
-        print("###getattr", file=open("temp", "at"))
 
         if self._future and self._future.status not in ("finished", "pending"):
             raise ValueError(
                 "Worker holding Actor was lost.  Status: " + self._future.status
             )
 
-        if self._worker and self._worker.address == self._address:
+        if (
+            self._worker
+            and self._worker.address == self._address
+            and threading.current_thread().name.startswith("Dask-Actor-Threads")
+        ):
+            # actor calls actor on same worker
             actor = self._worker.actors[self.key]
             attr = getattr(actor, key)
 
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -6,7 +6,7 @@ from queue import Queue
 
 from .client import Future, default_client
 from .protocol import to_serialize
-from .utils import sync
+from .utils import thread_state, sync
 from .utils_comm import WrappedKey
 from .worker import get_worker
 
@@ -128,7 +128,7 @@ class Actor(WrappedKey):
         if (
             self._worker
             and self._worker.address == self._address
-            and threading.current_thread().name.startswith("Dask-Actor-Threads")
+            and getattr(thread_state, "actor", False)
         ):
             # actor calls actor on same worker
             actor = self._worker.actors[self.key]
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -131,7 +131,7 @@ class Actor(WrappedKey):
             attr = getattr(actor, key)
 
             if iscoroutinefunction(attr):
-                return lambda *args, **kwargs: attr(*args, **kwargs)
+                return attr
 
             elif callable(attr):
                 return lambda *args, **kwargs: ActorFuture(
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+from inspect import iscoroutinefunction
 import threading
 from queue import Queue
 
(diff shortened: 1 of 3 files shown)


martindurant:
I tested manually with the following code, to show that blocking, async methods and properties work:
```python
import dask.distributed
client = dask.distributed.Client(n_workers=1)

class Minimal:
    def __init__(self):
        self.value = 0

    def inc(self):
        self.value += 1
        return self.value

    async def ainc(self):
        self.value += 1
        return self.value

    @property
    def val(self):
        return self.value

class UsesMinimal:

    def do_inc(self, ac, N):
        return ac.inc().result()

    async def ado_inc(self, ac, N):
        return await ac.ainc()

    def do_val(self, ac):
        return ac.val

ac = client.submit(Minimal, actor=True).result()
ac2 = client.submit(UsesMinimal, actor=True, workers=[ac._address]).result()
```

mrocklin:
Cool.  I'm glad to see this come together so fast.  Do you think it would be doable to include your manual test as a pytest test?

martindurant:
Yes of course - I wasn't certain whether what i wrote made sense or not.

On November 6, 2020 8:38:03 PM EST, Matthew Rocklin <notifications@github.com> wrote:
>Cool.  I'm glad to see this come together so fast.  Do you think it
>would be doable to include your manual test as a pytest test?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/dask/distributed/pull/4225#issuecomment-723370437

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

distributed/actor.py
```diff
@@ -118,13 +119,27 @@ def __dir__(self):
...
             raise ValueError(
                 "Worker holding Actor was lost.  Status: " + self._future.status
             )
 
+        if self._worker and self._worker.address == self._address:
+            actor = self._worker.actors[self.key]
+            attr = getattr(actor, key)
+
+            if iscoroutinefunction(attr):
+                return lambda *args, **kwargs: attr(*args, **kwargs)
```
mrocklin, line 134: I'm confused by this.   It seems like a no-op?  Would `return attr` be equivalent?

mrocklin:
> Yes of course - I wasn't certain whether what i wrote made sense or not.

I don't see anything that rings any warning bells, but all of my experience here is somewhat dated.  And hey, "if it passes tests ... " :)

distributed/actor.py
martindurant: That seems ... likely :). It must have been a little more complex as I tried&errored here.

martindurant:
This took a little more work, because the new code path I made only applies to workers calling workers, not normal submitted tasks calling workers (because they go via the scheduler and get executed in the normal pool).

martindurant:
This passed on Travis at https://travis-ci.com/github/martindurant/distributed/builds/199375854 but not updating here.

martindurant:
The following also fails - in fact, I think any exception in an actor might kill the worker. I'll put the fix in a new PR as soon as I have it
```python
def test_exception():
    class MyException(Exception):
        pass

    class Broken:
        def method(self):
            raise MyException

    with cluster(nworkers=2) as (cl, w):
        client = Client(cl["address"])
        ac = client.submit(Broken, actor=True).result()
        acfut = ac.method()
        with pytest.raises(MyException):
            acfut.result()
```

distributed/actor.py
```diff
@@ -118,13 +119,33 @@ def __dir__(self):
...
 
         if self._future and self._future.status not in ("finished", "pending"):
             raise ValueError(
                 "Worker holding Actor was lost.  Status: " + self._future.status
             )
 
+        if (
+            self._worker
+            and self._worker.address == self._address
+            and threading.current_thread().name.startswith("Dask-Actor-Threads")
```
mrocklin, line 131: Typically we use threadlocals for this.  You might want to look into the use of the `thread_state` variable in worker.py

distributed/worker.py
```diff
@@ -3416,6 +3416,7 @@ def apply_function_actor(
 
     thread_state.execution_state = execution_state
     thread_state.key = key
+    thread_state.actor = True
```
mrocklin, line 3419: Maybe we should set this to `False` just after calling the function?

martindurant: The intent, I think, is to state "yes, this is an actor-specific thread" rather than "we happen to be running an actor function right now". This use is more similar to checking the thread name.

PR #4229, opened 2020-11-09
Special case profile thread in leaked thread check

jrbourbeau:
In our test suite we check that threads are not leaked in between tests. However, I've been running to an issue where the [profiling thread we create](https://github.com/dask/distributed/blob/49044c3c3092d0bc3c98663a57c9f95b3ef38fd8/distributed/profile.py#L308-L319) isn't cleaned up for some tests (see the traceback from running `distributed/tests/test_scheduler.py::test_async_context_manager` below), which makes it difficult to run the test suite locally when developing.

This has been reported before when trying to run tests on a mac (xref https://github.com/dask/distributed/issues/3356) and I just had a co-worker run into the same issue when running tests on linux. This PR adds the profile thread to our list of threads which we ignore when checking for thread leaks as a temporary patch. I'll open up an issue so we can track down/fix the underlying problem and then remove the changes in this PR. 

<details>
<summary>Traceback:</summary>

```
=================================================================== ERRORS ====================================================================
_______________________________________________ ERROR at teardown of test_async_context_manager _______________________________________________

    @pytest.fixture
    def cleanup():
        with clean():
>           yield

distributed/utils_test.py:1543:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../miniforge3/envs/distributed/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
distributed/utils_test.py:1537: in clean
    del thread_state.on_event_loop_thread
../../../miniforge3/envs/distributed/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    @contextmanager
    def check_thread_leak():

        yield

        start = time()
        while True:
            bad = [
                t
                for t, v in threading._active.items()
                if t not in active_threads_start
                and "Threaded" not in v.name
                and "watch message" not in v.name
                and "TCP-Executor" not in v.name
                # TODO: Make sure profile thread is cleaned up
                # and remove the line below
                # and "Profile" not in v.name
            ]
            if not bad:
                break
            else:
                sleep(0.01)
            if time() > start + 5:
                from distributed import profile

                tid = bad[0]
                thread = threading._active[tid]
                call_stacks = profile.call_stack(sys._current_frames()[tid])
>               assert False, (thread, call_stacks)
E               AssertionError: (<Thread(Profile, started daemon 123145547345920)>, ['  File "/Users/james/miniforge3/envs/distributed/lib/python3.8/t...', '  File "/Users/james/projects/dask/distributed/distributed/profile.py", line 269, in _watch
E                 \tsleep(interval)
E                 '])
E               assert False

distributed/utils_test.py:1438: AssertionError
```
</details>

### DIFF
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -1421,6 +1421,9 @@ def check_thread_leak():
             and "Threaded" not in v.name
             and "watch message" not in v.name
             and "TCP-Executor" not in v.name
+            # TODO: Make sure profile thread is cleaned up
+            # and remove the line below
+            and "Profile" not in v.name
         ]
         if not bad:
             break


gforsyth:
I'll just add that I've also been hitting this issue locally on both OSX and Arch, but thought it was due to enterprise issues and Arch, respectively.

jrbourbeau:
👍 thanks for the additional data point @gforsyth!

Issue #4222, opened 2020-11-05
Aggregated topic logs


mrocklin:
It would be useful to have a global ledger of events, both Dask and user generated, that occured over time, indexed by topic.  This would be useful in a variety of situations:

1.  Reporting the progress of some externally-tracked system
2.  Reporting objective score values while training
3.  Tracking worker or scheduler events for future debugging
4.  Seeing when every worker joined and left the system
5.  ...

There are a few ways to do this today.  

1.  The `Server.log_event("topic_name", {"record": ...})` method, available on the Scheduler and Worker class independently.  This makes it easy to query based on topics like worker names or user defined terms, but doesn't aggregate records across all of the workers.
2.  The Pub/Sub system, which includes topics, and aggregates information, but doesn't keep things over time (`log_event` uses `deques`), and for some reason feels heavy-weight.
3.  Logging, and log-aggregators like DataDog

Neither log_event nor pub/sub get much love today.  I'm going to propose a hybrid.

As with `log_event`, we maintain a dict of deques of messages, keyed by topic name.  Current topics include every worker and client, but can also be user provided.

```python
>>> scheduler.events
{
    "127.0.0.1:7525": [(timestamp, "start"), (timestamp, {"op": "heartbeat", ...}), ...]
    "my-experiment-1": [...]
}
```

As with pub/sub, we provide user accessible API to append to these logs from anywhere in the system.  Here is an example where we track the start and stop of a computation.

```python
def mytask(x):
    start = time()

    ...

    stop = time()
    
    dask.distributed.log(topic="runtimes", {"start": start, "stop": stop})

futures = client.map(mytask, range(100))

>>> dask.distributed.logs("runtimes")
[]

wait(futures)

>>> dask.distributed.logs("runtimes")
[
 {"start": 100, "stop": 102}, 
 ...,
]
```

Some other extensions / ideas:

1.  These could be iterable
2.  Users might want to capture stdout/stderr and pipe them to logs
3.  We could have dashboard plots for these
4.  We might want to write these to a file (Althogh this would be pretty easy to combine with the iterable approach, assuming that they're also async-iterable)

Additionally, I can imagine combining this with other systems, like `WorkerPlugins` to provide some interesting diagnostic systems.

PR #4230, opened 2020-11-09
Add aggregated topic logs

jrbourbeau:
This PR adds the ability to log events under a specified topic to a centralized ledger of events that lives on the scheduler. Currently the centralized ledger contains some administrative events which were previously logged using `Scheduler.log_event`. With the changes in this PR, the client and workers now also have a `log_event` method to add their own events to the ledger. Additionally, the client also has a `get_events` method which allows users to retrieve events from the ledger.

Closes https://github.com/dask/distributed/issues/4222

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3579,15 +3579,16 @@ class Client:
         """
         return self.sync(self.scheduler.worker_logs, n=n, workers=workers, nanny=nanny)
 
-    def log_event(self, topic: str, msg):
+    def log_event(self, topic, msg):
         """Log an event under a given topic
 
         Parameters
         ----------
-        topic: str
-            Name of the topic under which to log an event
+        topic: str, list
+            Name of the topic under which to log an event. To log the same
+            event under multiple topics, pass a list of topic names.
         msg
-            Event to log. Note this must be a msgpack serializable.
+            Event message to log. Note this must be msgpack serializable.
 
         Examples
         --------
@@ -3597,12 +3598,13 @@ class Client:
         return self.sync(self.scheduler.log_event, topic=topic, msg=msg)
 
     def get_events(self, topic: str = None):
-        """Retrieve structured logs
+        """Retrieve structured topic logs
 
         Parameters
         ----------
-        topic: str
-            Name of topic log to retrieve events for
+        topic: str, optional
+            Name of topic log to retrieve events for. If no ``topic`` is
+            provided, then logs for all topics will be returned.
         """
         return self.sync(self.scheduler.events, topic=topic)
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3579,10 +3579,20 @@ class Client:
         """
         return self.sync(self.scheduler.worker_logs, n=n, workers=workers, nanny=nanny)
 
-    def log_event(self, topic, msg):
+    def log_event(self, topic: str, msg: dict):
+        """Log an event under a given topic
+
+        Parameters
+        ----------
+        topic: str
+            Name of the topic under which to log an event
+        msg: dict
+            Event to log. Note that this must be a msgpack serializable dictionary.
+        """
         return self.sync(self.scheduler.log_event, topic=topic, msg=msg)
 
-    def get_events(self, topic=None):
+    def get_events(self, topic: str = None):
+        """Retrieve structured logs"""
         return self.sync(self.scheduler.events, topic=topic)
 
     def retire_workers(self, workers=None, close_workers=True, **kwargs):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3579,20 +3579,31 @@ class Client:
         """
         return self.sync(self.scheduler.worker_logs, n=n, workers=workers, nanny=nanny)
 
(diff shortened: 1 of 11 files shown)


distributed/scheduler.py
```diff
@@ -1298,6 +1298,8 @@ def __init__(
         self.log = deque(
             maxlen=dask.config.get("distributed.scheduler.transition-log-length")
         )
+        self.events = defaultdict(lambda: deque(maxlen=100000))
+        self.event_counts = defaultdict(lambda: 0)
```
mrocklin, line 1302: If you wanted to be ultra-concise

```suggestion
        self.event_counts = defaultdict(int)
```

distributed/tests/test_client.py
```diff
@@ -6251,3 +6251,32 @@ async def test_get_task_metadata_multiple(c, s, a, b):
...
+
+    # Log an event from inside a task
+    def foo():
+        get_worker().log_event("topic1", {"foo": "bar"})
+
+    assert not await c.get_events("topic1")
+    await c.submit(foo)
+    events = await c.get_events("topic1")
+    assert len(events) == 1
+    assert events[0]["foo"] == "bar"
```
mrocklin, line 6267: Why not something like this?

```suggestion
    assert events = ({"foo": "bar"},) 
```

Are we trying to support list/tuple ambiguity?  If so, I recommend calling `list(events)` or `tuple(events)` explicitly.

distributed/tests/test_client.py
```diff
@@ -6251,3 +6251,32 @@ async def test_get_task_metadata_multiple(c, s, a, b):
...
+
+    # Log an event while on the scheduler
+    def log_scheduler(dask_scheduler):
+        dask_scheduler.log_event("topic2", {"woo": "hoo"})
```
mrocklin, line 6271: A fine contender to displace the foo-bar hegemony

distributed/scheduler.py
```diff
@@ -3736,6 +3741,9 @@ async def feed(
                 if teardown:
                     teardown(self, state)
 
+    def log_worker_event(self, worker=None, topic=None, msg=None):
+        self.log_event(topic, msg)
```
mrocklin, line 3745: Open question, should we include the worker address in a list of topics?

distributed/client.py
```diff
@@ -3579,6 +3579,12 @@ def get_worker_logs(self, n=None, workers=None, nanny=False):
         """
         return self.sync(self.scheduler.worker_logs, n=n, workers=workers, nanny=nanny)
 
+    def log_event(self, topic, msg):
+        return self.sync(self.scheduler.log_event, topic=topic, msg=msg)
```
mrocklin, line 3583: Open question, what should the name of this be?  `log_event` is perfectly fine.  Do we want something that claims more space, like `log`?

mrocklin (commented):
Some small comments, none of which are very important.  In general this seems good to me.  It would be useful to get some feedback from others, and maybe also include an example in a doc page somewhere.

distributed/scheduler.py
jrbourbeau: +1

distributed/tests/test_client.py
jrbourbeau: There's an extra`"time"` timestamp key in each event which makes hard-coding the expected event tuple slightly more tricky. We could pop the `"time"` key off each event and then `assert` the observed tuple is as expected &mdash; which would be totally fine. I'm also okay with the current style here

distributed/client.py
jrbourbeau: I have a slight preference for `log_event` as (to me) it conveys there's some additional structure beyond, for example, just logging a string in Python's normal logging system. Additionally `log` and `get_logs` are already in use elsewhere. We could certainly move things around, but I'm just not sure if it'd be worth the effort to do it right now

jrbourbeau:
> also include an example in a doc page somewhere

I didn't see an obvious place to document this, so I added a new "Logging" doc page with information on the event logs we're adding here and a couple other logging systems already in place. This seemed reasonable to me, but I'm happy to move things around too

After thinking about it a bit more, I also updated the structure of the events we log on the scheduler to be `(timestamp, event-msg)` (this turns out to be consistent with your original proposal in https://github.com/dask/distributed/issues/4222). I like this for a couple of reasons:

1. This allows non-`dict` messages to be logged which might be preferred in some situations. For example, if logging an ML model's score I'd prefer to do:

```python
worker.log_event("my-model-score", 0.9978)
```

instead of 

```python
worker.log_event("my-model-score", {"score": 0.9978})
```

which has a bit more structure than needed.

2. This way we don't mutate the event message that is logged (previously we were adding a `"time"` key).

distributed/scheduler.py
jrbourbeau: I don't have strong thoughts on this. This could, for example, be useful to get all the events logged on specific workers. Though I'm inclined to hold off on that for the time being just to keep things simple initially

mrocklin:
This seems like it's in a good state.  Let's merge now.  We can modify in the future if necessary.

Issue #4133, opened 2020-09-28
Computation deadlocks after inter worker communication error


fjetter:
Before I start, sincere apologies for not having a reproducible example but the issue is a bit weird and hard to reproduce. I mostly querying here to see if anybody has an idea what might be going on.

**What happened**:

We have seen multiple times now that some individual computations are stuck in the `processing` state and are not being processed. A closer investigation revealed in all occurrences that the worker the processing task is on, let's call it WorkerA, is in the process of fetching the tasks dependencies from WorkerB. While fetching these, an exception seems to be triggered on WorkerB and for some reason this exception is lost. In particular, WorkerA never gets any signal that something was wrong and it waits indefinitely for the dependency to arrive while WorkerB already forgot about the error. This scenario ultimately leads to a deadlock where the single dependency fetch is blocking the cluster. It looks like the cluster never actually self heals. The last time I could observe this issue, the computation was blocked for about 18hrs. After manually closing all connections, the workers retried and the graph finished successfully.

I've seen this deadlock happen multiple times already but only during the last instance, I could glimps an error log connecting to issue #4130 where the tornado stream seems to cause this error. Irregardless of what caused the error, I believe there is an issue on distributed side since the error was not properly propagated, the connections where not closed, etc. suggesting to WorkerA that everything was fine.

<details>
<summary>Traceback</summary>

```
Traceback (most recent call last):

  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/core.py", line 513, in handle_comm
    result = await result

  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/worker.py", line 1284, in get_data
    compressed = await comm.write(msg, serializers=serializers)

  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/comm/tcp.py", line 243, in write
    future = stream.write(frame)

  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/tornado/iostream.py", line 553, in write
    self._write_buffer.append(data)

  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/tornado/iostream.py", line 177, in append
    b += data  # type: ignore

BufferError: Existing exports of data: object cannot be re-sized
```

</details>

[...]

PR #4235, opened 2020-11-11
Close comm on internal handler errors

jochen-ott-by:
This improves the robustness in case of internal errors in handlers that leave `comm` in an invalid state, see also #4133

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -531,8 +531,11 @@ class Server:
                             logger.info("Lost connection to %r: %s", address, e)
                         break
                     except Exception as e:
-                        logger.exception(e)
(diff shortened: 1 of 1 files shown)


fjetter:
By this reasoning we should also remove 
https://github.com/dask/distributed/blob/7d769b8d87197211f6bd7c91335f9712fc4da949/distributed/core.py#L476

and the handling code on listener side

https://github.com/dask/distributed/blob/7d769b8d87197211f6bd7c91335f9712fc4da949/distributed/core.py#L679

jochen-ott-by:
Obviously, this breaks a lot of stuff and thus required major refactoring or a change of how to approach this. I'm therefore closing this and will probably create another PR soon.

PR #4238, opened 2020-11-11
Upload file worker plugin

### LINKED ISSUE #3560: Upload file to new workers only
Currently, `upload_file` will only send files to workers connected to scheduler at an instant = t.
If a worker connects to the scheduler afterwards, it does not have the data send with `upload_file` beforehand.

This is a problem because then we cannot assume that all the workers have the file available.

1. One simple workaround is to store the identity of the connected workers just before calling `upload_file` and only send jobs requiring this file to these workers in the future. But this prevent scalability.

2. An other possibility would be to register a `SchedulerPlugin` and use its `add_worker` event function to trigger there resending the file with  `upload_file` to all the workers when a new worker joins. If this solution is preferred, then it might be good that `upload_file` provides a `worker` parameter, so that we can target the newly connected worker and avoid reuploading the file to other workers already having the file.

3. As an extension of solution 2, add a `future_workers` bool parameter to `upload_file` and extend `upload_file` to automatically register the `SchedulerPlugin` described above.

4. Other ideas?

ian-r-rose:
Fixes #3560, based on @jrbourbeau's snippet.

This reimplements `client.upload_file` as a worker plugin, allowing it to function correctly when new workers are added after the initial upload.

### DIFF
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -4,7 +4,7 @@ from .actor import Actor, ActorFuture
 from .core import connect, rpc
 from .deploy import LocalCluster, Adaptive, SpecCluster, SSHCluster
 from .diagnostics.progressbar import progress
-from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall, UploadFile
+from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall
 from .client import (
     Client,
     Executor,
--- a/distributed/__init__.py
+++ b/distributed/__init__.py
@@ -4,7 +4,7 @@ from .actor import Actor, ActorFuture
 from .core import connect, rpc
 from .deploy import LocalCluster, Adaptive, SpecCluster, SSHCluster
 from .diagnostics.progressbar import progress
-from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall
+from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall, UploadFile
 from .client import (
     Client,
     Executor,
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3034,12 +3034,6 @@ class Client:
         """
         return self.sync(self._restart, **kwargs)
 
-    async def _upload_file(self, filename):
-        await self._register_worker_plugin(
-            UploadFile(filename),
-            name=filename + str(uuid.uuid4()),
-        )
-
     async def _upload_large_file(self, local_filename, remote_filename=None):
         if remote_filename is None:
             remote_filename = os.path.split(local_filename)[1]
@@ -3083,7 +3077,10 @@ class Client:
         >>> from mylibrary import myfunc  # doctest: +SKIP
         >>> L = client.map(myfunc, seq)  # doctest: +SKIP
         """
-        return self.sync(self._upload_file, filename)
+        return self.register_worker_plugin(
+            UploadFile(filename),
+            name=filename + str(uuid.uuid4()),
+        )
 
     async def _rebalance(self, futures=None, workers=None):
         await _wait(futures)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3036,8 +3036,8 @@ class Client:
 
     async def _upload_file(self, filename, raise_on_error=True):
         await self._register_worker_plugin(
-                UploadFile(filename),
-                name=filename+str(uuid.uuid4()),
+            UploadFile(filename),
+            name=filename + str(uuid.uuid4()),
         )
 
     async def _upload_large_file(self, local_filename, remote_filename=None):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3034,8 +3034,11 @@ class Client:
         """
         return self.sync(self._restart, **kwargs)
 
-    async def _upload_file(self, filename):
-        await self._register_worker_plugin(UploadFile(filename), name=filename)
+    async def _upload_file(self, filename, raise_on_error=True):
+        await self._register_worker_plugin(
+                UploadFile(filename, raise_on_error=raise_on_error),
+                name=filename,
+        )
 
     async def _upload_large_file(self, local_filename, remote_filename=None):
         if remote_filename is None:
@@ -3080,7 +3083,7 @@ class Client:
         >>> from mylibrary import myfunc  # doctest: +SKIP
         >>> L = client.map(myfunc, seq)  # doctest: +SKIP
         """
-        return self.sync(self._upload_file, filename)
+        return self.sync(self._upload_file, filename, raise_on_error=self.asynchronous)
 
     async def _rebalance(self, futures=None, workers=None):
         await _wait(futures)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3036,8 +3036,8 @@ class Client:
 
     async def _upload_file(self, filename, raise_on_error=True):
         await self._register_worker_plugin(
-                UploadFile(filename, raise_on_error=raise_on_error),
-                name=filename,
+                UploadFile(filename),
+                name=filename+str(uuid.uuid4()),
         )
 
     async def _upload_large_file(self, local_filename, remote_filename=None):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -69,7 +69,7 @@ from .security import Security
 from .sizeof import sizeof
 from .threadpoolexecutor import rejoin
 from .worker import get_client, get_worker, secede
-from .diagnostics.plugin import WorkerPlugin
+from .diagnostics.plugin import UploadFile, WorkerPlugin
 from .utils import (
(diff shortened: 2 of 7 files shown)


distributed/worker.py
```diff
@@ -999,7 +999,7 @@ def func(data):
                 cache_loads.data.clear()
             except Exception as e:
                 logger.exception(e)
-                return {"status": "error", "exception": to_serialize(e)}
+                return {"status": "error", "exception": e}
```
ian-r-rose, line 1002: Exception handling is now handled by the worker plugin system, so this no longer should be serialized.

distributed/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .core import connect, rpc
 from .deploy import LocalCluster, Adaptive, SpecCluster, SSHCluster
 from .diagnostics.progressbar import progress
-from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall
+from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall, UploadFile
```
jrbourbeau, line 7: As most users will interact with this plugin via `Client.upload_file`, I don't think we need to include it in the top-level `distributed` namespace

```suggestion
from .diagnostics.plugin import WorkerPlugin, SchedulerPlugin, PipInstall
```

distributed/diagnostics/plugin.py
```diff
@@ -258,3 +259,38 @@ async def setup(self, worker):
                     worker.loop.add_callback(
                         worker.close_gracefully, restart=True
                     )  # restart
+
+
+# Adapted from https://github.com/dask/distributed/issues/3560#issuecomment-596138522
+class UploadFile(WorkerPlugin):
+    """A WorkerPlugin to upload a local python module to workers.
```
jrbourbeau, line 266: Small nitpick

```suggestion
    """A WorkerPlugin to upload a local file to workers.
```

distributed/tests/test_client.py
```diff
@@ -1684,6 +1684,21 @@ def test_upload_file_exception_sync(c):
...
+        import myfile
+
+        return myfile.x
+
+    with tmp_text("myfile.py", "x = 123") as fn:
+        await c.upload_file(fn)
+        await Worker(s.address, loop=s.loop)
+        x = await c.submit(g)
+
+        assert x == 123
```
jrbourbeau, line 1699: Nice test!

distributed/client.py
```diff
@@ -3035,22 +3035,11 @@ def restart(self, **kwargs):
...
 
     async def _upload_file(self, filename, raise_on_error=True):
-        with open(filename, "rb") as f:
-            data = f.read()
-        _, fn = os.path.split(filename)
-        d = await self.scheduler.broadcast(
-            msg={"op": "upload_file", "filename": fn, "data": to_serialize(data)}
+        await self._register_worker_plugin(
+            UploadFile(filename),
+            name=filename + str(uuid.uuid4()),
```
jrbourbeau, line 3040: Are we making the `name` here unique so that if a file is updated locally and then re-uploaded that the new version of the file will be present on the workers? If so, could you add a test for this case (if one doesn't already exist)

distributed/worker.py
jrbourbeau: Nice! I think we can simplify a little further and just raise the exception here directly

```suggestion
                raise e
```

jrbourbeau (commented):
Thanks for the PR @ian-r-rose! Generally this looks good, I'm looking forward to seeing this merged

distributed/__init__.py
ian-r-rose: :+1:

distributed/client.py
ian-r-rose: Yes, that's right. Without this unique name, re-uploading the same file with changes has no effect. This is already tested by existing tests, which do exactly that.

jrbourbeau: Great!

distributed/tests/test_client.py
mrocklin, line 1699: I suspect that tests will yell at you for leaving a lingering Worker .  We might want to try `async with Worker(...) as w` instead.  Also, I think that the `loop=` keyword is probably no longer necessary.

distributed/diagnostics/plugin.py
```diff
@@ -258,3 +259,38 @@ async def setup(self, worker):
...
+        self.filename = os.path.basename(filepath)
+        with open(filepath, "rb") as f:
+            self.data = f.read()
+
+    async def setup(self, worker):
+        response = await worker.upload_file(
+            comm=None, filename=self.filename, data=self.data, load=True
+        )
+        if response["status"] == "error":
+            raise response["exception"]
```
jrbourbeau, line 295: Since we started raising an exception directly in `Worker.upload_file`, we can drop these lines too 

```suggestion
```

distributed/client.py
```diff
@@ -3035,22 +3035,11 @@ def restart(self, **kwargs):
         return self.sync(self._restart, **kwargs)
 
     async def _upload_file(self, filename, raise_on_error=True):
```
jrbourbeau, line 3037: It looks like `raise_on_error` isn't used anymore. Since we're maintaining our previous behavior (`test_upload_file_exception` and `test_upload_file_exception_sync` both pass still) this might be a good opportunity to remove `raise_on_error` as a keyword argument.

distributed/diagnostics/plugin.py
ian-r-rose: ah, good catch

distributed/tests/test_worker.py
```diff
@@ -1665,6 +1665,32 @@ async def test_pip_install_fails(c, s, a, b):
...
+@gen_cluster(client=True)
+async def test_upload_file_plugin_exception(c, s, a, b):
+    def g():
+        import myfile
+
+        return myfile.x
+
+    with tmp_text("myfile.py", "!syntax_error") as fn:
+        with pytest.raises(SyntaxError):
+            await c.register_worker_plugin(UploadFile(fn))
```
mrocklin, line 1691: I'm curious if there is ever an occasion when we would want users to interact with `UploadFile` directly, rather than using `Client.upload_file`.  If not, then we might want to only use the `upload_file` method in tests and treat `UploadFile` as an implementation detail.  Historically I've found that this makes it easier to change internals in the future (as you've just done here).  Thoughts?

ian-r-rose: The only difference I can think of is that the user could give the plugin a different name, thereby changing the behavior of re-uploading the same file multiple times. I've chosen `fname + uuid` for the plugin name to allow re-uploads to take effect.

I'm not sure why one might want to do that, however.

distributed/client.py
```diff
@@ -3034,23 +3034,12 @@ def restart(self, **kwargs):
...
-    async def _upload_file(self, filename, raise_on_error=True):
-        with open(filename, "rb") as f:
-            data = f.read()
-        _, fn = os.path.split(filename)
-        d = await self.scheduler.broadcast(
-            msg={"op": "upload_file", "filename": fn, "data": to_serialize(data)}
+    async def _upload_file(self, filename):
+        await self._register_worker_plugin(
+            UploadFile(filename),
+            name=filename + str(uuid.uuid4()),
```
mrocklin, line 3040: I suspect that we no longer need the split `upload_file/_upload_file` here, and can simplify this by using `register_worker_plugin` directly

```python
def upload_file(self, filename):
    return self.register_worker_plugin(
        UploadFile(filename),
        name=filename + str(uuid.uuid4()),
    )
```

The background here is that the `_underscored_functions` are all async while the `normal_functions` use the `Client.sync` method, which determines if we're in the same thread as an event loop or not, and calls the `_underscored_function` directly if we are, returning a future that the calling code then awaits, or if we are in a different thread it triggers the `_underscored_function` to be called on the event loop, and then blocks until it is done.

In our case, this magic is already handled by `register_worker_plugin` and so we can probably rely on that method's use of the `sync`.  We probably don't need to do this dance again.  Hooray for basing methods on top of other methods.

mrocklin: (This sync magic is historically annoying to first get used to, my apologies.  It *is* nice once you start blending async and sync code though)

distributed/tests/test_worker.py
mrocklin: Yeah, I'm inclined to stick with just the upload_file method.  This PR is also fine as is though if you want to be done.  I'm mostly splitting hairs here.

ian-r-rose: I love deleting code

jrbourbeau (approved):
Thanks for your work here @ian-r-rose. The changes here LGTM, I'll merge once CI passes

jrbourbeau:
Also, I noticed this was your first code contribution to this repository. Welcome! It's good to have you here : )

PR #4239, opened 2020-11-12
comm: close comm on low-level errors

jochen-ott-by:
Close `comm.stream` on low-level errors, such as BufferErrors. In such cases, we do not really know what was written to / read from the underlying socket, so the only safe way forward is to immediately close the connection.
See also #4133

### DIFF
--- a/distributed/tests/test_batched.py
+++ b/distributed/tests/test_batched.py
@@ -297,11 +297,5 @@ async def test_handles_exceptions():
             b.send("hello")
             b.send("world")
             await asyncio.sleep(0.020)
-            result = await comm.read()
-            assert result == ("hello", "hello", "world")
-
-            b.send("raises when flushed")
-            await asyncio.sleep(0.020)  # CommClosedError hit in callback
-
             with pytest.raises(CommClosedError):
-                b.send("raises when sent")
+                await comm.read()
--- a/distributed/tests/test_batched.py
+++ b/distributed/tests/test_batched.py
@@ -1,6 +1,5 @@
 import asyncio
 import random
-from unittest import mock
 
 import pytest
 from tlz import assoc
@@ -8,10 +7,9 @@ from tlz import assoc
 from distributed.batched import BatchedSend
 from distributed.core import listen, connect, CommClosedError
 from distributed.metrics import time
-from distributed.utils import All, TimeoutError
+from distributed.utils import All
 from distributed.utils_test import captured_logger
 from distributed.protocol import to_serialize
-from distributed.compatibility import WINDOWS, TORNADO6
 
 
 class EchoServer:
@@ -250,52 +248,4 @@ async def test_serializers():
         assert "type" in value
         assert "function" in value
 
-        msg = await comm.read()
-        assert list(msg) == [{"x": 123}, {"x": "hello"}]
-
-        with pytest.raises(TimeoutError):
-            msg = await asyncio.wait_for(comm.read(), 0.1)
-
-
-@pytest.mark.asyncio
-@pytest.mark.skipif(WINDOWS and not TORNADO6, reason="failing on windows, tornado 5.")
-async def test_handles_exceptions():
-    # Ensure that we properly handle exceptions in BatchedSend.
-    # https://github.com/pangeo-data/pangeo/issues/788
-    # mentioned in https://github.com/dask/distributed/issues/4080, but
-    # possibly distinct.
-    #
-    # The reported issues (https://github.com/tornadoweb/tornado/pull/2008)
-    # claim that the BufferError *should* only happen when the application
-    # is incorrectly using threads. I haven't been able to construct an
-    # actual example, so we mock IOStream.write to raise and ensure that
-    # BufferedSend handles things correctly. We don't (yet) test that
-    # any *users* of BatchedSend correctly handle BatchedSend dropping
-    # messages.
-    async with EchoServer() as e:
-        comm = await connect(e.address)
-        b = BatchedSend(interval=10)
-        b.start(comm)
-        await asyncio.sleep(0.020)
-        orig = comm.stream.write
-
-        n = 0
-
-        def raise_buffererror(*args, **kwargs):
-            nonlocal n
-            n += 1
-
-            if n == 1:
-                raise BufferError("bad!")
-            elif n == 2:
-                orig(*args, **kwargs)
-            else:
-                raise CommClosedError
-
-        with mock.patch.object(comm.stream, "write", wraps=raise_buffererror):
-            b.send("hello")
-            b.send("hello")
-            b.send("world")
-            await asyncio.sleep(0.020)
-            with pytest.raises(CommClosedError):
-                await comm.read()
+        assert comm.closed()
--- a/distributed/batched.py
+++ b/distributed/batched.py
@@ -99,28 +99,17 @@ class BatchedSend:
                     self.recent_message_log.append("large-message")
                 self.byte_count += nbytes
             except CommClosedError as e:
-                # If the comm is known to be closed, we'll immediately
-                # give up.
                 logger.info("Batched Comm Closed: %s", e)
                 break
             except Exception:
-                # In other cases we'll retry a few times.
-                # https://github.com/pangeo-data/pangeo/issues/788
-                if self._consecutive_failures <= 5:
-                    logger.warning("Error in batched write, retrying")
-                    yield gen.sleep(0.100 * 1.5 ** self._consecutive_failures)
-                    self._consecutive_failures += 1
-                    # Exponential backoff for retries.
-                    # Ensure we don't drop any messages.
-                    if self.buffer:
-                        # Someone could call send while we yielded above?
-                        self.buffer = payload + self.buffer
-                    else:
-                        self.buffer = payload
-                    continue
-                else:
-                    logger.exception("Error in batched write")
-                    break
+                # We cannot safely retry self.comm.write, as we have no idea
+                # what (if anything) was actually written to the underlying stream.
+                # Re-writing messages could result in complete garbage (e.g. if a frame
+                # header has been written, but not the frame payload), therefore
+                # the only safe thing to do here is to abort the stream without
+                # any attempt to re-try `write`.
+                logger.exception("Error in batched write")
+                break
             finally:
                 payload = None  # lose ref
         else:
@@ -128,9 +117,10 @@ class BatchedSend:
             self.stopped.set()
             return
 
-        # If we've reached here, it means our comm is known to be closed or
-        # we've repeatedly failed to send a message. We can't close gracefully
-        # via `.close()` since we can't send messages. So we just abort.
+        # If we've reached here, it means `break` was hit above and
+        # there was an exception when using `comm`.
+        # We can't close gracefully via `.close()` since we can't send messages.
+        # So we just abort.
         # This means that any messages in our buffer our lost.
         # To propagate exceptions, we rely on subsequent `BatchedSend.send`
         # calls to raise CommClosedErrors.
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -200,6 +200,13 @@ class TCP(Comm):
             self.stream = None
             if not shutting_down():
                 convert_stream_closed_error(self, e)
+        except Exception:
+            # Some OSError or a another "low-level" exception. We do not really know what
+            # was already read from the underlying socket, so it is not even safe to retry
+            # here using the same stream. The only safe thing to do is to abort.
+            # (See also GitHub #4133).
+            self.abort()
+            raise
         else:
             try:
                 msg = await from_frames(
@@ -253,13 +260,18 @@ class TCP(Comm):
                         await future
                         bytes_since_last_yield = 0
         except StreamClosedError as e:
-            stream = None
-            convert_stream_closed_error(self, e)
-        except TypeError as e:
+            self.stream = None
+            if not shutting_down():
+                convert_stream_closed_error(self, e)
+        except Exception:
+            # Some OSError or a another "low-level" exception. We do not really know what
+            # was already written to the underlying socket, so it is not even safe to retry
+            # here using the same stream. The only safe thing to do is to abort.
+            # (See also GitHub #4133).
             if stream._write_buffer is None:
                 logger.info("tried to write message %s on closed stream", msg)
-            else:
-                raise
+            self.abort()
+            raise
 
         return sum(lengths)
 
--- a/distributed/comm/tests/test_comms.py
+++ b/distributed/comm/tests/test_comms.py
@@ -790,6 +790,25 @@ async def test_inproc_comm_closed_explicit_2():
     await comm.close()
 
 
+@pytest.mark.asyncio
+async def test_comm_closed_on_buffer_error():
+    # Internal errors from comm.stream.write, such as
+    # BufferError should lead to the stream being closed
+    # and not re-used. See GitHub #4133
+    reader, writer = await get_tcp_comm_pair()
+
+    def _write(data):
+        raise BufferError
+
+    writer.stream.write = _write
+    with pytest.raises(BufferError):
+        await writer.write("x")
+    assert writer.stream is None
+
+    await reader.close()
+    await writer.close()
+
+
 #
 # Various stress tests
 #
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -472,9 +472,12 @@ class Server:
                         )
                     break
                 except Exception as e:
-                    logger.exception(e)
-                    await comm.write(error_message(e, status="uncaught-error"))
-                    continue
+                    logger.exception("Exception while reading from %s", address)
+                    if comm.closed():
+                        raise
+                    else:
+                        await comm.write(error_message(e, status="uncaught-error"))
+                        continue
                 if not isinstance(msg, dict):
                     raise TypeError(
                         "Bad message type.  Expected dict, got\n  " + str(msg)
@@ -531,8 +534,11 @@ class Server:
                             logger.info("Lost connection to %r: %s", address, e)
                         break
                     except Exception as e:
-                        logger.exception(e)
-                        result = error_message(e, status="uncaught-error")
+                        logger.exception("Exception while handling op %s", op)
+                        if comm.closed():
+                            raise
+                        else:
+                            result = error_message(e, status="uncaught-error")
 
                 # result is not type stable:
                 # when LHS is not Status then RHS must not be Status or it raises.


distributed/tests/test_batched.py
```diff
@@ -297,11 +297,5 @@ def raise_buffererror(*args, **kwargs):
             b.send("hello")
             b.send("world")
             await asyncio.sleep(0.020)
-            result = await comm.read()
```
jochen-ott-by, line 300: @TomAugspurger You wrote this test I touched here, so you may want to look at it: On BufferError, the first `comm.read` already fails.

jochen-ott-by:
The only build failure is a for `test_broken_worker_during_computation` on python3.6 on linux.

As this is a known issue (see #4173) and only affected one of the 6 builds, so I think we can consider all tests passed for the purpose of this PR.

distributed/tests/test_batched.py
fjetter: The test now catches another exception.

The original test was testing for https://github.com/dask/distributed/blob/eda9bcca8baa502481bf929f39340857da1ebf09/distributed/batched.py#L146

but with your implementation we're now hitting https://github.com/dask/distributed/blob/eda9bcca8baa502481bf929f39340857da1ebf09/distributed/comm/tcp.py#L184

fjetter: After looking into it a bit, I think this is a clash of philosophy and this test actually tests the opposite to what you are proposing here.
You suggest that e.g. a BufferError must not be retried but the implementation introduced in #4135 explicitly implements a retry mechansim for these failures.

What happens is: the first instance of the mock raises a `BufferError` which was supposed to be retried. Your implementation now, however, immediately closes the comm before anything can be retried.

If we allow for retries in other exception cases, the "solution" would be to change the mock to use a different exception class than BufferError. However, I think the intention was to explicitly deal with a buffer error :/

jochen-ott-by: > After looking into it a bit, I think this is a clash of philosophy and this test actually tests the opposite to what you are proposing here.

I think the code that re-tries here is deeply flawed, for one simply reason: one cannot safely re-try `write` if one does not know how much was written in the previous attempt. So I'd rather go ahead and delete all the retry code for the batched send. It is just plain wrong to retry here.

> You suggest that e.g. a BufferError must not be retried but the implementation introduced in #4135 explicitly implements a retry mechansim for these failures.

Yes, re-trying is just wrong.

jochen-ott-by: I added a second commit which removes the re-tries in `distributed.batched`.

I re-based on the latest master, so tests have a chance to run.

fjetter:
I believe the reasoning in here is fine. We should, however, double check that this doesn't cause further trouble. The retry implementation apparently fixed things for the pangeo people, see https://github.com/pangeo-data/pangeo/issues/788

For me there are two questions connected to this change
* Is the failure / closure of the BatchedSend being dealt with gracefully? (E.g. worker<->scheduler should just trigger a reconnect)
* What happens to the buffered payload in the failure case? If it is lost, what would this mean? I have the impression that most of what we're transmitting over batched send relies on a guaranteed delivery (although this is not true, not even now)

I would appreciate another opinion, cc @TomAugspurger @mrocklin @jrbourbeau

TomAugspurger:
I don’t really have anything to add here, other than BatchedSend being fire and forget seemed a bit strange. It wasn’t clear to me how the system handled messages that were lost in the buffer.

mrocklin:
> I have the impression that most of what we're transmitting over batched send relies on a guaranteed delivery 

Dask would not be robust to dropped messages along the BatchedSend channel.  I am generally curious why buffers are being dropped.  If they are, resending seems like a good idea (maybe?) or otherwise we could disconnect and reconnect the worker and hope that the general worker-loss resilience mechanism handle things.

mrocklin:
@TomAugspurger who had the connection issues in Pangeo?  Would they be open to testing out this alternative fix?

TomAugspurger:
The failing workload is at https://github.com/pangeo-data/pangeo/issues/788 <https://github.com/pangeo-data/pangeo/issues/788> (though it failed maybe 1/5 or 1/10 times).

If anyone has time & interest they can run the example on pangeo’s binder, at https://binder.pangeo.io <https://binder.pangeo.io/>, pointing to this branch, similar to https://github.com/TomAugspurger/pangeo-binder-test/blob/master/binder/Dockerfile

martindurant:
Is there anything left to do here?

jochen-ott-by:
> Is there anything left to do here?

FWIW, I think we can merge this.

martindurant:
@fjetter did you have more thoughts?

mrocklin:
I think that it would be useful for someone to try the pangeo workload that Tom mentions here: https://github.com/dask/distributed/pull/4239#issuecomment-730511897

Maybe @rabernat has someone he can point to this?

rabernat:
The example in https://github.com/pangeo-data/pangeo/issues/788 originated with @paigem. Maybe should could try to give it a spin?

paigem:
@rabernat I'm happy to test out this PR with the example from [pangeo-data/pangeo#788](https://github.com/pangeo-data/pangeo/issues/788). I just tried to recreate the error from that example with no luck, but was using a different dataset, so I'll have another try tomorrow.

@TomAugspurger you mention testing the example on Pangeo binder - is that preferable to running it on Pangeo Cloud (where the original error occurred)?

TomAugspurger:
With the binder at binder.pangeo.io you can test this branch by making a dockerfile like https://github.com/TomAugspurger/pangeo-binder-test/blob/master/binder/Dockerfile. We can't easily test unreleased packages on pangeo cloud. But in the end they're the same, from this PR's point of view. Just a Dask cluster created with Dask Gateway.

paigem:
Thanks for the tip @TomAugspurger. I started up a Pangeo Binder instance pointing to this branch, but I don't seem to be able to start up a Dask Gateway cluster. I get a "GatewayClusterError: Cluster failed to start". Not sure why it's not working - the notebook I'm running and the Dockerfile can be found here: [paigem/test_dask_comms](https://github.com/paigem/test_dask_comms). Sorry I'm not of more help at the moment!

TomAugspurger:
Right now you need to manually ensure that the version of dask-gateway in your image is compatible with dask-gateway-server running in the client. I think that’s 0.9 right now, but it should be documented in the binder docs.

________________________________

paigem:
I can now start up a Dask Gateway cluster (thanks @TomAugspurger) with dask-gateway version 0.9.0 as discussed in the Pangeo binder docs. But I get a `CancelledError` when trying to run even a basic example calculation:

```python
import dask.array as dsa
a = dsa.ones((1000,1000,100))
print(a.mean().compute())
```

However, when I start up a binder instance with the default Dask-distributed (i.e. I don't point it to this PR branch), then I am able to run the above (as well as more complicated calculations) just fine. So it seems like the `CancelledError` I am getting is either due to how I'm setting up my binder instance to point to this branch or an issue with this PR. It's likely the former, since I'm a newbie and still struggling to understand how to debug here, but happy to provide more details if someone is able to help me figure out why I'm getting a `CancelledError`. The Dockerfile I'm using to set up Binder can be found [here](https://github.com/paigem/test_dask_comms/Dockerfile).

TomAugspurger:
Ah, most likely dask needs to be on master as well.

paigem:
@TomAugspurger you're right, it looks like dask is not on master in my binder instance. How do I ensure that my binder uses the dask master branch, with this unreleased version of distributed? Do I need to add a line to my [Dockerfile](https://github.com/paigem/test_dask_comms/blob/master/Dockerfile)?

TomAugspurger:
You can pip install it from GitHub too, if you still want to try this.

paigem:
@TomAugspurger Sorry I haven't been of any help here. I'm happy to keep trying at this, as I'm learning a lot, but I also don't want to stall this PR any further. I'm still having trouble getting the correct version of both dask and distributed into my Binder instance, so if you or anyone else has any specific pointers for how to make this work, I'm happy to keep at it. 

<details><summary>Dockerfile contents:</summary>

```
FROM pangeo/pangeo-notebook
RUN /srv/conda/envs/notebook/bin/python -m pip install -U --no-deps git+https://github.com/jochen-ott-by/distributed.git@close-comm-on-internal-errors
RUN /srv/conda/envs/notebook/bin/python -m pip install -U --no-deps git+https://github.com/dask/dask.git
```

</details>

<details><summary>Result of <tt>client.get_versions()</tt>:</summary>

```
{'scheduler': {'host': {'python': '3.8.6.final.0',
   'python-bits': 64,
   'OS': 'Linux',
   'OS-release': '4.19.112+',
   'machine': 'x86_64',
   'processor': 'x86_64',
   'byteorder': 'little',
   'LC_ALL': 'C.UTF-8',
   'LANG': 'C.UTF-8'},
  'packages': {'python': '3.8.6.final.0',
   'dask': '2.30.0+84.gb8131ea1',
   'distributed': '2.3.2+606.g06a3f2a0',
   'msgpack': '1.0.0',
   'cloudpickle': '1.6.0',
   'tornado': '6.1',
   'toolz': '0.11.1',
   'numpy': '1.19.4',
   'lz4': '3.1.0',
   'blosc': '1.9.2'}},
 'workers': {'tcp://127.0.0.1:32959': {'host': {'python': '3.8.6.final.0',
    'python-bits': 64,
    'OS': 'Linux',
    'OS-release': '4.19.112+',
    'machine': 'x86_64',
    'processor': 'x86_64',
    'byteorder': 'little',
    'LC_ALL': 'C.UTF-8',
    'LANG': 'C.UTF-8'},
   'packages': {'python': '3.8.6.final.0',
    'dask': '2.30.0+84.gb8131ea1',
    'distributed': '2.3.2+606.g06a3f2a0',
    'msgpack': '1.0.0',
    'cloudpickle': '1.6.0',
    'tornado': '6.1',
    'toolz': '0.11.1',
    'numpy': '1.19.4',
    'lz4': '3.1.0',
    'blosc': '1.9.2'}},
  'tcp://127.0.0.1:40157': {'host': {'python': '3.8.6.final.0',
    'python-bits': 64,
    'OS': 'Linux',
    'OS-release': '4.19.112+',
    'machine': 'x86_64',
    'processor': 'x86_64',
    'byteorder': 'little',
    'LC_ALL': 'C.UTF-8',
    'LANG': 'C.UTF-8'},
   'packages': {'python': '3.8.6.final.0',
    'dask': '2.30.0+84.gb8131ea1',
    'distributed': '2.3.2+606.g06a3f2a0',
    'msgpack': '1.0.0',
    'cloudpickle': '1.6.0',
    'tornado': '6.1',
    'toolz': '0.11.1',
    'numpy': '1.19.4',
    'lz4': '3.1.0',
    'blosc': '1.9.2'}},
  'tcp://127.0.0.1:43929': {'host': {'python': '3.8.6.final.0',
    'python-bits': 64,
    'OS': 'Linux',
    'OS-release': '4.19.112+',
    'machine': 'x86_64',
    'processor': 'x86_64',
    'byteorder': 'little',
    'LC_ALL': 'C.UTF-8',
    'LANG': 'C.UTF-8'},
   'packages': {'python': '3.8.6.final.0',
    'dask': '2.30.0+84.gb8131ea1',
    'distributed': '2.3.2+606.g06a3f2a0',
    'msgpack': '1.0.0',
    'cloudpickle': '1.6.0',
    'tornado': '6.1',
    'toolz': '0.11.1',
    'numpy': '1.19.4',
    'lz4': '3.1.0',
    'blosc': '1.9.2'}},
  'tcp://127.0.0.1:44805': {'host': {'python': '3.8.6.final.0',
    'python-bits': 64,
    'OS': 'Linux',
    'OS-release': '4.19.112+',
    'machine': 'x86_64',
    'processor': 'x86_64',
    'byteorder': 'little',
    'LC_ALL': 'C.UTF-8',
    'LANG': 'C.UTF-8'},
   'packages': {'python': '3.8.6.final.0',
    'dask': '2.30.0+84.gb8131ea1',
    'distributed': '2.3.2+606.g06a3f2a0',
    'msgpack': '1.0.0',
    'cloudpickle': '1.6.0',
    'tornado': '6.1',
    'toolz': '0.11.1',
    'numpy': '1.19.4',
    'lz4': '3.1.0',
    'blosc': '1.9.2'}}},
 'client': {'host': {'python': '3.8.6.final.0',
   'python-bits': 64,
   'OS': 'Linux',
   'OS-release': '4.19.112+',
   'machine': 'x86_64',
   'processor': 'x86_64',
   'byteorder': 'little',
   'LC_ALL': 'C.UTF-8',
   'LANG': 'C.UTF-8'},
  'packages': {'python': '3.8.6.final.0',
   'dask': '2.30.0+84.gb8131ea1',
   'distributed': '2.3.2+606.g06a3f2a0',
   'msgpack': '1.0.0',
   'cloudpickle': '1.6.0',
   'tornado': '6.1',
   'toolz': '0.11.1',
   'numpy': '1.19.4',
   'lz4': '3.1.0',
   'blosc': '1.9.2'}}}
```

</details>

TomAugspurger:
Probably not worth holding the PR up for that test, if it's otherwise ready to go.

paigem:
Ok, I finally figured this out and am now able to test this PR on Pangeo Binder. Apologies for being so slow to get this up and running @TomAugspurger. 

I am still getting similar behavior as in [pangeo-data/pangeo#788](https://github.com/pangeo-data/pangeo/issues/788), where one worker seems to have all the tasks and the computation stalls, though it doesn't look like the worker is out of memory in this case (see screenshot below). However, I only got this behavior twice in about 20 tries. 

Screen Shot 2020-12-02 at 00 24 56


Here is the code I am running:

```python
from dask_gateway import Gateway
gateway = Gateway()
options = gateway.cluster_options()
options.worker_memory = 6
cluster = gateway.new_cluster(options)
cluster.adapt(minimum=1, maximum=30)

import gcsfs
import dask.array as dsa
import os
scratch_path = 'gs://pangeo-scratch/paigem/'
gcs = gcsfs.GCSFileSystem(requester_pays=True)
varname = 'SHF_2'
url = f'{scratch_path}/CESM_POP_hires_control/{varname}_target.zarr'
target_mapper = gcs.get_mapper(url)
target_array = dsa.from_zarr(target_mapper)

target_array.mean(axis=0).compute()
```

Dockerfile and dask/distributed versions:

<details><summary>Dockerfile contents:</summary>

```
FROM pangeo/pangeo-notebook
RUN /srv/conda/envs/notebook/bin/python -m pip install -U --no-deps git+https://github.com/jochen-ott-by/distributed.git@close-comm-on-internal-errors
RUN /srv/conda/envs/notebook/bin/python -m pip install -U git+https://github.com/dask/dask.git
```

</details>

<details><summary>Result of <tt>client.get_versions()</tt>:</summary>

```
{'scheduler': {'host': {'python': '3.8.6.final.0',
   'python-bits': 64,
   'OS': 'Linux',
   'OS-release': '4.19.112+',
   'machine': 'x86_64',
   'processor': 'x86_64',
   'byteorder': 'little',
   'LC_ALL': 'C.UTF-8',
   'LANG': 'C.UTF-8'},
  'packages': {'python': '3.8.6.final.0',
   'dask': '2.30.0+84.gb8131ea1',
   'distributed': '2.3.2+606.g06a3f2a0',
   'msgpack': '1.0.0',
   'cloudpickle': '1.6.0',
   'tornado': '6.1',
   'toolz': '0.11.1',
   'numpy': '1.19.4',
   'lz4': '3.1.0',
   'blosc': '1.9.2'}},
 'workers': {'tls://10.48.127.134:34669': {'host': {'python': '3.8.6.final.0',
    'python-bits': 64,
    'OS': 'Linux',
    'OS-release': '4.19.112+',
    'machine': 'x86_64',
    'processor': 'x86_64',
    'byteorder': 'little',
    'LC_ALL': 'C.UTF-8',
    'LANG': 'C.UTF-8'},
   'packages': {'python': '3.8.6.final.0',
    'dask': '2.30.0+84.gb8131ea1',
    'distributed': '2.3.2+606.g06a3f2a0',
    'msgpack': '1.0.0',
    'cloudpickle': '1.6.0',
    'tornado': '6.1',
    'toolz': '0.11.1',
    'numpy': '1.19.4',
    'lz4': '3.1.0',
    'blosc': '1.9.2'}}},
 'client': {'host': {'python': '3.8.6.final.0',
   'python-bits': 64,
   'OS': 'Linux',
   'OS-release': '4.19.112+',
   'machine': 'x86_64',
   'processor': 'x86_64',
   'byteorder': 'little',
   'LC_ALL': 'C.UTF-8',
   'LANG': 'C.UTF-8'},
  'packages': {'python': '3.8.6.final.0',
   'dask': '2.30.0+84.gb8131ea1',
   'distributed': '2.3.2+606.g06a3f2a0',
   'msgpack': '1.0.0',
   'cloudpickle': '1.6.0',
   'tornado': '6.1',
   'toolz': '0.11.1',
   'numpy': '1.19.4',
   'lz4': '3.1.0',
   'blosc': '1.9.2'}}}
```

</details>

jochen-ott-by:
Ok, so how do we proceed here?

I think it's pretty obvious that the retry code in batched_send is flawed in the general case, because it can lead to garbage on the comm if re-trying on any exception after a partial write. For me, this is enough reason to remove it. This is true even if a problem re-appears where this re-try helped: instead, we should fix the problem properly.

But I sense this is not the consensus here? So what would be required to get this merged?

mrocklin:
@paigem to be clear, are you saying that before things worked well, and with this PR things work less well?

rabernat:
My interpretations of @paigem's comment is that she was finally able to test this PR via a properly-configured binder. She found that this PR _did not resolve the intermittent failure issue (with tasks all stuck on one worker) we experienced with earlier versions of distributed_.

mrocklin:
> For me, this is enough reason to remove it. This is true even if a problem re-appears where this re-try helped: instead, we should fix the problem properly.

@jochen-ott-by it sounds like you are confident that this is an improvement.  Correct?

> But I sense this is not the consensus here? So what would be required to get this merged?

I suspect that the problem here is that no one feels comfortable making a final decision.  Historically I was the one who handled tricky comms issues and I think people tended to wait until I weighed in.  That probably makes less sense now that I've been sidelined a bit recently organizing other things.  My apologies, and thank you @jochen-ott-by for continuing to follow up here.  

If this PR doesn't negatively impact the Pangeo workload (see question above to @paigem) , then I think that we're good to go and should merge.  If this PR does negatively impact the Pangeo workload then it might be useful to figure out what that failure looks like (@paigem do you happen to have a traceback?) and see what we can do there.

mrocklin:
> My interpretations of @paigem's comment is that she was finally able to test this PR via a properly-configured binder. She found that this PR did not resolve the intermittent failure issue (with tasks all stuck on one worker) we experienced with earlier versions of distributed.

Got it.  Just to be clear, the current released version of distributed *does* resolve that issue, correct?   Merging this in would degrade performance for Pangeo folks?

rabernat:
I reported the original issue in https://github.com/pangeo-data/pangeo/issues/788. There is no traceback to report, because the computation never finishes or errors.

> the current released version of distributed _does_ resolve that issue, correct?

I _thought_ it was resolved by https://github.com/dask/distributed/pull/4135. I don't have enough knowledge to understand how this PR interacts with that one.

This problem is not particularly severe, since it only happens on 1 out of 20 trials. If merging this PR helps move development forward, go for it! But it would be good to keep track of this lingering issue somewhere.

TomAugspurger:
I was reasonably confident that https://github.com/dask/distributed/pull/4135 fixed the issue with computations completely stalling, though it's hard to say for sure. So *if* that's correct then a stalled computation on this branch would be a regression.

That said, I really know how confident I should be that this was actually fixed by https://github.com/dask/distributed/pull/4135. If we're able to reproduce the stalled computation with distributed 2.30.1, then we can conclude that https://github.com/dask/distributed/pull/4135 didn't fix the original issue, and can proceed with this.

Or we determine that this PR is a net improvement and move forward with it anyway, and try to re-debug the stalled computation issue afterwards.

rabernat:
Maybe, when @paigem comes online (she's down under 🙃 ), she could share the binder repo / url she is using to debug this. It may be useful to have a shared reality where the problem is reproducible.

paigem:
> I thought it was resolved by #4135. I don't have enough knowledge to understand how this PR interacts with that one.

Is the update from #4135 currently running on Pangeo Cloud? If so, then my thinking was also that this comms issue was resolved in that update, as I wasn't able to reproduce the error. However, I probably only reran the calculation ~ 10 tries, which isn't conclusive. If it would be useful, I could test the #4135 version again either on Pangeo Cloud or in a Binder instance, but I don't want to delay this PR more than I already have... 🙂

As @rabernat suggested, here is the url I used to start my Pangeo binder `https://github.com/paigem/test_dask_comms` on the `master` branch. I am reading in data from my scratch bucket, so someone will need permissions to access that.

paigem:
Also, quick question to verify that I had the correct Dask/distributed versions on my binder: when running `client.get_versions()` it showed that the dask version installed appears to be an unreleased version `'dask': '2.30.0+84.gb8131ea1'`. No need to respond if this looks right - just wanted to highlight this in case I got something wrong here! (You can see all other package versions in an earlier [comment](https://github.com/dask/distributed/pull/4239#issuecomment-736581044) of mine.)

jochen-ott-by:
> > For me, this is enough reason to remove it. This is true even if a problem re-appears where this re-try helped: instead, we should fix the problem properly.
> 
> @jochen-ott-by it sounds like you are confident that this is an improvement. Correct?

Yes, I am confident it is an improvement: It removes an execution path that potentially blocks forever.

One execution path it removes are the "unusual exception from socket.send", which would look like this:

 * socket.send raises an exception that does *not* inherit from IOError (the last part is a bit unusual and I cannot fully explain that, but not completely unheard of for, esp. of wrapped sockets like ssl sockets). BTW, this exception would (via the exception traceback) grab a ref to the `data` argument which might be an important ingredient to creating BufferErrors.
https://github.com/tornadoweb/tornado/blob/b4e39e52cd27d6b3b324a399dff046f71545c4a5/tornado/iostream.py#L1169
 * this exception is not handled by tornado, because tornado only catches IOError and closely related exception types
https://github.com/tornadoweb/tornado/blob/b4e39e52cd27d6b3b324a399dff046f71545c4a5/tornado/iostream.py#L993
 * the exception bubbles up the chain and arrives at distributed. What happens next depends on where the comm.write happened. One likely possibility is that it happened within a `Server.handlers` invocation, so the exception would be caught here: https://github.com/dask/distributed/blob/7ea9696cc6861e5a0d524025c2f0ff251e881bf6/distributed/core.py#L519
   ... where the exception is converted to a message.
 * the message is sent to the `comm` (that just emitted an exception!) here:
    https://github.com/dask/distributed/blob/7ea9696cc6861e5a0d524025c2f0ff251e881bf6/distributed/core.py#L532

So overall, there is the situation that `comm.write` is called twice despite there was an exception the first time. In general, such cases can lead to weird behavior, as the first write might have partially succeeded and (maybe) the second write works somehow. This would lead to a garbage byte stream on the read end where the read end of that connection might wait forever.

Note that this fix is only about handling low-level comm errors by more aggressively closing the connection; it does not fix the actual source of the first error assumed above. I hoped these exception would become more visible, though.

Even if what I described is not happening right now "in the wild", I still think this change would be a net improvement, because I think the distributed code should be robust against such cases.

mrocklin:
This is a difficult decision.  We're choosing between two incomplete solutions to a problem that affects two different groups differently.  We're also making that decision without diving into and fully understanding the underlying issue.  Diagnosing the underlying issue requires significant setup work to reproduce.  Given that, what do we do?

> Or we determine that this PR is a net improvement and move forward with it anyway, and try to re-debug the stalled computation issue afterwards.

This seems like the lead bad option to me.  I'm going to go ahead with merging this.  It looks like this will degrade performance a bit on a Pangeo workload, but it sets us up to better understand and resolve the problem in the future I think.  @jochen-ott-by if Pangeo folks come back with a traceback I might try to engage you in thinking about a solution in the future if you're ok with that.

@paigem thank you for the small example.  That's very helpful.  If you're able to verify that the problem also exists without private data that would make it easier for other maintainers to reproduce this and see what is going on.  

> There is no traceback to report, because the computation never finishes or errors.

Also, while you may not get a traceback in Jupyter, it might still be helpful to get logs from the workers (which may include tracebacks).  I *think* that Dask-Gateway will get these for you if you run something like `cluster.get_logs()`

martindurant:
> Also, while you may not get a traceback in Jupyter, it might still be helpful to get logs from the workers (which may include tracebacks). I think that Dask-Gateway will get these for you if you run something like cluster.get_logs()

I was going to ask elsewhere, but it seems to me a good idea to be able to set scheuler/worker logging to DEBUG from the client. This doesn't yet exist, right (without a hand-crafter `run()`) ?

mrocklin:
I suspect that a failed network connection will show up at the ERROR level.  My read from @rabernat 's comment above is that they're not seeing a traceback in Jupyter, but aren't yet checking the worker logs.

mrocklin:
Not that I know of, but yes, I would prefer to continue that conversation
elsewhere if that's ok

rabernat:
> If you're able to verify that the problem also exists without private data that would make it easier for other maintainers to reproduce this and see what is going on.

The data are accessible from Pangeo Binder, so they are not fully private.

mrocklin:
Does this require GCS?  Does it require Zarr?  Can this be done with random data, or is it something about using those libraries at the same time that causes the issue?

fjetter:
Thanks for merging this @mrocklin ! If the pangeo issue bubbles up again, feel free to ping me as well, in case @jochen-ott-by is not available

PR #4254, opened 2020-11-19
Fuse steps in `extract_serialize`

jakirkham:
When Distributed prepares a message with [`dumps`]( https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/protocol/core.py#L26 ), it makes a call to [`extract_serialize`]( https://github.com/dask/distributed/blob/18fff8b0ca0297e100695519931c78e579cb55a6/distributed/protocol/core.py#L38 ), which in turn [extracted all of the binary objects to serialize alongside]( https://github.com/dask/distributed/blob/09d9799e401da695981600fe98ab4d4de4ec419e/distributed/protocol/serialize.py#L459 ), [took a copy of the message with binary objects removed]( https://github.com/dask/distributed/blob/09d9799e401da695981600fe98ab4d4de4ec419e/distributed/protocol/serialize.py#L460-L467 ), and [then extracted all of the `bytes` and `bytearray` objects]( https://github.com/dask/distributed/blob/09d9799e401da695981600fe98ab4d4de4ec419e/distributed/protocol/serialize.py#L469-L473 ). This resulted in a full copy of the message along with a few passes over the message and it contents to refine it. As a result `extract_serialize` and notably `container_copy` showed up when profiling a shuffle workload with Dask.

To improve the performance of `extract_serialize`, we handle all of these steps in one pass of the recursive function `_extract_serialize` avoiding the need for copying of the data or later mutation. While we are collecting the objects to serialize, we flag any `bytes` or `bytearray` objects for special handling as well. In addition we construct a new message from elements in the old message that don't require special handling. Finally `_extract_serialize` had two code paths depending on whether a `list` or `dict` was involved, which varied slightly between the two cases. With a small amount of prep work at the beginning of `_extract_serialize`, we were able to fuse these two blocks of code together. Thus cutting a bunch of lines of code in the process.

With these changes, we are able to cut the runtime of this function roughly in half as demonstrated by the simple example below.

<details>
<summary>Before:</summary>

```python

   ...: msg = 1_000 * [data]

1.1 ms ± 7.05 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

</details>

<details>
<summary>After:</summary>

```python

   ...: msg = 1_000 * [data]

612 µs ± 14.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

</details>

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -447,17 +447,12 @@ def extract_serialize(x):
     """
     x2 = type(x)()
     ser = {}
-    _extract_serialize(x, x2, ser)
-
     bytestrings = set()
-    for k, v in ser.items():
-        if type(v) in (bytes, bytearray):
-            ser[k] = to_serialize(v)
-            bytestrings.add(k)
+    _extract_serialize(x, x2, ser, bytestrings)
     return x2, ser, bytestrings
 
 
-def _extract_serialize(x, x2, ser, path=()):
+def _extract_serialize(x, x2, ser, bytestrings, path=()):
     typ_x = type(x)
     if typ_x is dict:
         x_items = x.items()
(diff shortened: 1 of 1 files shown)


mrocklin:
I haven't checked over things, but if tests pass then I'm happy.  +1 from me.

mrocklin:
Thanks @jakirkham !

jakirkham:
Thanks Matt! 😀

PR #4258, opened 2020-11-20
WIP: Avoid recursion in `_extract_serialize`

jakirkham:
This rewrites `_extract_serialize` to avoid recursion. Does this by making use of a [`collections.deque`]( https://docs.python.org/3/library/collections.html#collections.deque ) to track all of the nested `dict`s and `list`s that still need to be visited. A pass through each `dict` or `list` is completed before passing through any nested ones. Otherwise it behaves mostly like a depth-first search. These are all looped through until no more can be found. Then it returns to `extract_serialize`.

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -1,6 +1,5 @@
 from array import array
 from functools import partial
-from itertools import repeat
 import traceback
 import importlib
 from enum import Enum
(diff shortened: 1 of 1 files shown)


jakirkham:
The `type` check improvements themselves are useful. So have broken that out as PR ( https://github.com/dask/distributed/pull/4281 ).

That said, eliminating recursion here doesn't seem to have much impact. At this point just calling `to_serialize` is more expensive than anything else here, which isn't saying much.

Going to go ahead and close this out. If things change, we can always revisit.

distributed/protocol/serialize.py
```diff
@@ -445,34 +445,46 @@ def extract_serialize(x):
...
+        x2 = len(x) * [None]
+
+    ser = {}
+    bytestrings = set()
+    _extract_serialize(x_items, x2, ser, bytestrings)
+    return x2, ser, bytestrings
+
+
+def _extract_serialize(x_items, x2, ser, bytestrings, path=()):
+    q = deque()
```
mrocklin, line 463: If we're using this as only a stack then maybe this should be a list rather than a deque

jakirkham: Yeah I tried both. This performed better based on the benchmarking I did. That said, it seems that recursion itself is already handled quite efficiently.

PR #4259, opened 2020-11-20
Tracking Child Processes (#2568)

fbriol:
Modification allowing to take into account the child processes created from the workers.

### DIFF
--- a/distributed/system_monitor.py
+++ b/distributed/system_monitor.py
@@ -8,6 +8,7 @@ from .metrics import time
 class SystemMonitor:
     def __init__(self, n=10000):
         self.proc = psutil.Process()
+        self.children = set()
 
         self.time = deque(maxlen=n)
         self.cpu = deque(maxlen=n)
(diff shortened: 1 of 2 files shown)


distributed/system_monitor.py
```diff
@@ -45,6 +46,24 @@ def update(self):
         with self.proc.oneshot():
             cpu = self.proc.cpu_percent()
             memory = self.proc.memory_info().rss
+            children = set(self.proc.children(True))
```
mrocklin, line 49: Hrm, it looks like it takes a few milliseconds in order to compute this.  This may be too expensive to have on by default.  We run this fairly frequently.

```python
In [1]: import psutil

In [2]: proc = psutil.Process()

In [3]: %timeit proc.children()
4.87 ms ± 161 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

fbriol: The time-consuming step is the reading of the process table. First line of the ``psutil.Process.children`` function.

```python
In [1]: import psutil

In [2]: proc = psutil.Process()

In [3]: %timeit proc.children()
8.79 ms ± 36.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit psutil._psplatform.ppid_map()
8.59 ms ± 101 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
One of the possible solutions, maybe if you have a better idea, is to delegate the calculation to a thread. For example,

````python
import threading
import psutil
import time


class TrackChildren(threading.Thread):
    def __init__(self, proc, *args, **kwargs) -> None:
        self._proc = proc
        self._cpu = 0
        self._mem = 0
        self._children = set()
        self._lock = threading.RLock()
        self._event = threading.Event()

    def stop(self):
        self._event.set()

    def values(self):
        with self._lock:
            return self._cpu, self._mem

    def run(self):
        while not self._event.is_set():
            cpu = 0
            mem = 0
            children = set(self._proc.children(True))
            # "self.children" tracks the subprocesses to calculate the CPU
            # usage correctly. Otherwise, the computed CPU usage would always
            # be 0 as no time interval would exist for the calculation
            # (cf. psutil.Process.cpu_percent).
            if children:
                new_children = children - self._children
                if new_children:
                    self._children.update(new_children)
                for child in list(self._children):
                    # The inspected process may die during its introspection.
                    try:
                        with child.oneshot():
                            cpu += child.cpu_percent()
                            mem += child.memory_info().rss
                    except (psutil.NoSuchProcess, psutil.ZombieProcess):
                        self._children.discard(child)
            with self._lock:
                self._cpu, self._mem = cpu, mem
            time.sleep(0.1)
````
In this case, the reading of the numbers is much faster.
````python
t = TrackChildren(psutil.Process())
t.start()
%timeit t.values()
306 ns ± 5.68 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
````

PR #4264, opened 2020-11-23
Speculative task assignment

### LINKED ISSUE #3974: Populate workers with downstream tasks
### Motivation

Currently when a task becomes ready to run the scheduler finds the best worker for that task distributes that task to that worker to be enqueued.  The worker then handles whatever communication is necessary in order to collect the dependencies for that task and once it's ready it puts it in a queue to be run in a local ThreadPoolExecutor.  When it finishes the worker informs the scheduler and goes on to its next task.  When the scheduler receives news that the task has finished it marks all of its dependents, each of which go through this process again.

This is occasionally sub-optimal.  Consider the following graph:

```
A1 -> B1
A2 -> B2
```

If we have one worker then both of the A's will be sent to that worker.  The worker will finish A1, send the report to the scheduler that it is finished, and begin work on A2.  It will then get information from the scheduler that B1 is ready to compute and it will work on that next.  As a result, the resulting order of execution looks like the following:

```
A1, A2, B1, B2
```

When really we would have preferred ...

```
A1, B1, A2, B2
```

Today we often avoid this situation by doing task fusion on the client side, so that we really only have two tasks

```
AB1
AB2
```

If we remove low-level task fusion (which we may want to do for performance reasons) then it would be nice to capture this same A, B, A, B behavior some other way

### Send tasks to workers before dependencies are set

One way to capture this same behavior is to send more tasks down to the worker, even before they are ready to run.  Today we only send a task to a worker once we have high confidence that that is where it should run, which typically we only know after we understand the data sizes of all of its inputs.  We only know this once its dependencies are done, so we only send tasks to workers once all of their dependencies are complete.

However, we could send not-yet-ready-to-run tasks to a worker with high confidence if all of that task's dependencies are also present or running on that worker.  This happens to fully consume the low-level task fusion case.  If `A1` is running on a worker then we believe with high probability that `B1` will also run on that same worker.

### Changes

This would be a significant change to the worker.  We would need to add states to track dependencies, much like how we do in the dask.local scheduler, or a very stripped down version of the dask.distributed scheduler.  

This will also require managing interactions with other parts of Dask like ...

1.  Work stealing / load balancing:  What happens when to a dependent task when its dependency is taken? (my guess is that it probably goes back to the scheduler and we let the scheduler sort it out)
2.  Resources / restrictions:  We should make sure to only speculatively schedule a dependent on a worker if that worker is valid for that worker
3.  Exceptions / failures:  Probably we just kick the task back up to the scheduler if anything bad happens.

### Learning

I think that this task is probably a good learning task for someone who has some moderate exposure to the distributed scheduler and wants to level up a bit.  Adding dependencies to the worker will, I think, force someone to think a lot about the systems that help make Dask run while mostly implementing minor versions of them.  cc @quasiben @jacobtomlinson @madsbk

gforsyth:
This is a preliminary cut at introducing speculative task assignment to the distributed scheduler and closes #3974 

As @mrocklin mentions in #3974, if tasks are speculatively assigned to a worker that contains all of their dependencies (which may still be processing or awaiting execution), this removes the need for task fusion.

The current approach I've started sketching out here is to use the existing `recommendation` system in the scheduler to examine a tasks dependent(s) -- (I'm going to switch to parent/child terminology here to make this less confusing):
  if a task has only one child and all parents of that child task are on the same worker, speculatively assign that task to the same worker.

Still TODO:
- [ ] so many more tests
- [ ] figure out if we can disable fusion for an entire test run (to make sure tasks are being assigned as expected)
- [ ] Update scheduling policy documentation to include possibility of speculative assignment
- [x] Figure out what to do with Actors (ignore spec assignment in this case?): don't use speculative assignment with actors
- [ ] Handle resource restrictions
- [ ] Handle work stealing

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1413,6 +1413,7 @@ class Scheduler(ServerNode):
             ("waiting", "memory"): self.transition_waiting_memory,
             ("waiting", "speculative"): self.transition_waiting_speculative,
             ("speculative", "processing"): self.transition_speculative_processing,
+            ("speculative", "released"): self.transition_speculative_released,
             ("processing", "released"): self.transition_processing_released,
             ("processing", "memory"): self.transition_processing_memory,
             ("processing", "erred"): self.transition_processing_erred,
@@ -4246,6 +4247,38 @@ class Scheduler(ServerNode):
             ts.state = "processing"
 
             return {}
+
+        except Exception as e:
+            logger.exception(e)
+            if LOG_PDB:
+                import pdb
+
+                pdb.set_trace()
+            raise
+
+    def transition_speculative_released(self, key):
+        try:
+            ts = self.tasks[key]
+
+            if self.validate:
+                assert ts.state == "speculative"
+                assert ts.processing_on
+
+            ts.waiters.clear()
+            ts.waiting_on.clear()
+
+            self._remove_from_processing(
+                ts, send_worker_msg={"op": "release-task", "key": key}
+            )
+            ts.state = "released"
+
+            recommendations = {}
+
+            for dts in ts.dependents:
+                if dts.state in ("erred", "speculative"):
+                    recommendations[dts.key] = "released"
+
+            return recommendations
 
         except Exception as e:
             logger.exception(e)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -653,7 +653,6 @@ class TaskState:
         "exception",
         "traceback",
         "exception_blame",
-        "speculative",
         "suspicious",
         "retries",
         "nbytes",
@@ -688,7 +687,6 @@ class TaskState:
         self.group_key = key_split_group(key)
         self.group = None
         self.metadata = {}
-        self.speculative = False
 
     @property
     def state(self) -> str:
@@ -2520,8 +2518,8 @@ class Scheduler(ServerNode):
         assert ts in ws.processing
         assert not ts.who_has
         for dts in ts.dependencies:
-            assert dts.who_has or ts.speculative
-            assert ts in dts.waiters or ts.speculative
+            assert dts.who_has
+            assert ts in dts.waiters
 
     def validate_memory(self, key):
         ts = self.tasks[key]
@@ -2746,16 +2744,16 @@ class Scheduler(ServerNode):
                 msg["resource_restrictions"] = ts.resource_restrictions
             if ts.actor:
                 msg["actor"] = True
-            if ts.speculative:
+            if ts.state == "speculative":
                 msg["speculative"] = True
 
             deps = ts.dependencies
-            if deps and not ts.speculative:
+            if deps and not ts.state == "speculative":
                 msg["who_has"] = {
                     dep.key: [ws.address for ws in dep.who_has] for dep in deps
                 }
                 msg["nbytes"] = {dep.key: dep.nbytes for dep in deps}
-            elif ts.speculative:
+            elif ts.state == "speculative":
                 msg["who_has"] = {dep.key: dep.processing_on.address for dep in deps}
                 msg["nbytes"] = {dep.key: dep.nbytes for dep in deps}
 
@@ -4208,12 +4206,9 @@ class Scheduler(ServerNode):
 
             ws = list(ts.dependencies)[0].processing_on
 
-            worker = ws.address
-
             duration = self.get_task_duration(ts)
-            # There are no comm costs if this is speculative
-            # comm = self.get_comm_cost(ts, ws)
 
+            # There are no comm costs if this is speculative
             ws.processing[ts] = duration
             ts.processing_on = ws
             ws.occupancy += duration
@@ -4223,12 +4218,11 @@ class Scheduler(ServerNode):
             self.check_idle_saturated(ws)
             self.n_tasks += 1
 
-            ts.speculative = True
-
             for dts in ts.dependencies:
                 ts.waiting_on.add(dts)
+                dts.waiters.add(ts)
 
-            self.send_task_to_worker(worker, key)
+            self.send_task_to_worker(ws.address, key)
 
             return _recommend_speculative_assignment(ts)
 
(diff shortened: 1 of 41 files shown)


distributed/scheduler.py
```diff
@@ -2467,14 +2506,18 @@ def validate_waiting(self, key):
...
+        )
         ws = ts.processing_on
         assert ws
         assert ts in ws.processing
         assert not ts.who_has
         for dts in ts.dependencies:
-            assert dts.who_has
-            assert ts in dts.waiters
+            assert dts.who_has or ts.speculative
+            assert ts in dts.waiters or ts.speculative
```
mrocklin, line 2520: I would expect there to be a separate `validate_speculative` method that was applied to tasks in the `speculative` state rather than reusing this method.

distributed/scheduler.py
```diff
@@ -4138,6 +4193,52 @@ def decide_worker(self, ts):
...
+
+            duration = self.get_task_duration(ts)
+            # There are no comm costs if this is speculative
+            # comm = self.get_comm_cost(ts, ws)
+
+            ws.processing[ts] = duration
+            ts.processing_on = ws
+            ws.occupancy += duration
+            self.total_occupancy += duration
+            ts.state = "processing"
```
mrocklin, line 4217: Maybe the state should be "speculative" instead?  This would remove the need for the `.speculative` attribute, and maybe make some of the other handling in the scheduler more consistent.

distributed/scheduler.py
```diff
@@ -4138,6 +4193,52 @@ def decide_worker(self, ts):
...
         return worker
 
+    def transition_waiting_speculative(self, key):
+        try:
+
+            ts = self.tasks[key]
+
+            if self.validate:
+                # All dependencies are on the same worker
+                assert len({dts.processing_on for dts in ts.dependencies}) == 1
```
mrocklin, line 4203: We probably also want that at least one of the dependencies is in a "processing" state

distributed/scheduler.py
```diff
@@ -4138,6 +4193,52 @@ def decide_worker(self, ts):
...
+
+            ws = list(ts.dependencies)[0].processing_on
+
+            worker = ws.address
```
mrocklin, line 4207: I have a slight preference to use `ws.address` inline.  I find that having few root terms reduces mental indirection when reviewing / reading later.  "Wait, what is worker?  Was this a parameter that was passed in?  Oh no, I see, it's just the address of the workers state."

distributed/scheduler.py
```diff
@@ -4138,6 +4193,52 @@ def decide_worker(self, ts):
...
+            if ts.actor:
+                ws.actors.add(ts)
+
+            ts.speculative = True
+
+            # logger.debug("Send job to worker: %s, %s", worker, key)
+
+            self.send_task_to_worker(worker, key)
+
+            return _recommend_speculative_assignment(ts)
```
mrocklin, line 4231: Ah, so we would send long chains down to a worker all at once?

distributed/scheduler.py
```diff
@@ -3837,7 +3885,14 @@ def get_comm_cost(self, ts, ws):
         Get the estimated communication cost (in s.) to compute the task
         on the given worker.
         """
-        return sum(dts.nbytes for dts in ts.dependencies - ws.has_what) / self.bandwidth
+        # TODO: How is it possible for nbytes to be None when there's a getter that is supposed to
+        # stop that from happening?
```
mrocklin, line 3889: This looks like an excellent "who-dunnit?" mystery to enjoy

distributed/tests/test_speculative.py
```diff
@@ -0,0 +1,93 @@
...
+async def test_speculative_assignment_simple(c, s, a, b):
+    x = delayed(inc)(1)
+    y = delayed(inc)(x)
+    z = delayed(dec)(y)
+
+    zz = c.compute(z)
+    result = await zz
+    assert result == 2
+    assert (y.key, "speculative", "ready") in a.story(y.key)
+    assert (z.key, "speculative", "ready") in a.story(z.key)
```
mrocklin, line 16: Another way to test this would be to have a slowinc, followed by two fast-incs.  Once the slow-inc is found on the worker we assert that all three tasks are in `worker.tasks`.  This seems like a test that is slightly less likely to go bad as people change around internals.  (what we have here is great too though, I'm just spit-balling)

distributed/worker.py
```diff
@@ -1441,7 +1443,7 @@ def add_task(
                 self.tasks[key] = ts = TaskState(
                     key=key, runspec=SerializedTask(function, args, kwargs, task)
                 )
-                ts.state = "waiting"
+                ts.state = "waiting" if not speculative else "speculative"
```
mrocklin, line 1446: Hrm, what if the dependency has just finished running?  Maybe there is a way to trust the local state here more than what the scheduler said.

mrocklin (commented):
Eight comments.  Ah! Ah! Ah!

distributed/scheduler.py
gforsyth: that's what currently happens and I can see it leading to less than stellar performance, especially if a cluster is scaling up -- 
We could add a config item about # of speculative tasks to assign to a given worker?  a la `max_height` in the fuse options?

mrocklin: Oh, I think it's fine.  Let's leave it like this until there is a problem.

distributed/scheduler.py
```diff
@@ -1371,9 +1407,12 @@ def __init__(
 
         self._transitions = {
             ("released", "waiting"): self.transition_released_waiting,
+            ("released", "speculative"): self.transition_waiting_speculative,
             ("waiting", "released"): self.transition_waiting_released,
             ("waiting", "processing"): self.transition_waiting_processing,
             ("waiting", "memory"): self.transition_waiting_memory,
+            ("waiting", "speculative"): self.transition_waiting_speculative,
+            ("speculative", "processing"): self.transition_speculative_processing,
```
mrocklin, line 1415: We'll eventually want to see what happens when a speculative task gets pushed back to released/waiting again, such as if a worker goes down.  The catch all behavior is, I think, to move things back to released.  I think that if you handle that that things should be ok, we may also want a speculative->waiting though.  Same with errors.  I suspect that when a processing task errs it will mark all of its dependents as errored, and so the scheduler will try to enact a speculative->erred transition.  

In some of these it may be that the current `self.transition_processing_*` methods will work without modification, but we'll probably have to verify this.

distributed/worker.py
gforsyth: It shouldn't make a difference, when `add_task` finishes we'll have noted that the dependency is in memory and then the task will transition from speculative -> ready

PR #4267, opened 2020-11-23
Find the set of workers instead of their frequency

jakirkham:
Since we don't use the frequency of workers in any way and this takes a bit of time to compute, simply determine the set of workers instead. This gives us the information we use in the end without any loss of functionality and speeds up a bottleneck here.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -20,7 +20,6 @@ import psutil
 import sortedcontainers
 
 from tlz import (
-    frequencies,
     merge,
     pluck,
     merge_sorted,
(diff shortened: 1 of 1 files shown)


jakirkham:
Should add here's a simple benchmark that illustrates the difference

```python
In [1]: import random
   ...: 
   ...: from tlz import frequencies

In [2]: L = [random.randint(0, 10) for i in range(1_000)]

In [3]: %timeit frequencies(L)
27.8 µs ± 183 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %timeit set(L)
7.7 µs ± 15.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

PR #4268, opened 2020-11-24
Use `set.update` to include other `comms`

jakirkham:
Instead of coercing other collections to `set`s before performing a union with the `set` of `comms`, which can take a bit longer, simply call `comms.update(...)` on the provided collection. This avoids the need to build a separate `set` before performing the union and still has the same effect of including new values in `comms` where relevant.

Included a link to relevant Python docs for this method (for those unfamiliar). Note that both `frozenset` and `set` are documented together hence the reference to `frozenset` in the link even though it applies to `set` also.

https://docs.python.org/3/library/stdtypes.html#frozenset.update

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2589,14 +2589,14 @@ class Scheduler(ServerNode):
             ts = self.tasks.get(msg["key"])
         if ts is None:
             # Notify all clients
-            comms |= set(self.client_comms.values())
+            comms.update(self.client_comms.values())
         else:
             # Notify clients interested in key
-            comms |= {
+            comms.update(
                 self.client_comms[c.client_key]
                 for c in ts.who_wants
                 if c.client_key in self.client_comms
-            }
+            )
         for c in comms:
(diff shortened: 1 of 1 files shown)


jakirkham:
Here's a short benchmark illustrating the difference. Admittedly this may not come as much of a surprise upon reflection.

```python
In [1]: import random
   ...: 
   ...: from tlz import frequencies

In [2]: L1 = [random.randint(0, 1_000) for i in range(100)]
   ...: L2 = [random.randint(0, 1_000) for i in range(100)]

In [3]: %%timeit S = set(L1)
   ...: S |= set(L2)
   ...: 
   ...: 
2.62 µs ± 34.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: %%timeit S = set(L1)
   ...: S.update(L2)
   ...: 
   ...: 
952 ns ± 19.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

PR #4269, opened 2020-11-24
Use `Status` `Enum` in `remove_worker`

jakirkham:
Looks like this can use the `Status` `Enum`, which we already use elsewhere. So make use of it here too.


mrocklin:
cc @Carreau

jakirkham:
AFAIU this was added in PR ( https://github.com/dask/distributed/pull/3853 ). Here we are just catching one case that was missed that should be using the `Enum`.

PR #4270, opened 2020-11-24
Drop custom `__eq__` from `Status`

jakirkham:
This appears to eat up a good chunk of time in places where the `Status` needs to be checked. So try dropping this method in favor of the default implementation.

That said, I'm sure this is here for some reason. So marking this as WIP so we can discuss whether removing is ok or if we need to figure out some alternative solution.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -61,32 +61,6 @@ class Status(Enum):
     undefined = None
     dont_reply = "dont-reply"
 
-    def __eq__(self, other):
-        """
-        Implement equality checking with backward compatibility.
-
-        If other object instance is string, we compare with the values, but we
-        actually want to make sure the value compared with is in the list of
-        possible Status, this avoid comparison with non-existing status.
-        """
-        if isinstance(other, type(self)):
-            return self.value == other.value
-        elif isinstance(other, str) or (other is None):
-            warnings.warn(
-                f"Since distributed 2.23 `.status` is now an Enum, please compare with `Status.{other}`",
-                PendingDeprecationWarning,
-                stacklevel=1,
-            )
-            assert other in [
-                s.value for s in type(self)
-            ], f"comparison with non-existing states {other}"
-            return other == self.value
-        raise TypeError(
-            f"'==' not supported between instances of"
-            f" {type(self).__module__+'.'+type(self).__qualname__!r} and"
-            f" {type(other).__module__+'.'+type(other).__qualname__!r}"
-        )
-
 
 class RPCClosed(IOError):
     pass
--- a/distributed/tests/test_core.py
+++ b/distributed/tests/test_core.py
@@ -100,15 +100,6 @@ def test_server_status_assign_non_variant_raises():
             server.status = "I do not exists"
 
 
-def test_server_status_compare_non_variant_raises():
-    server = Server({})
-    # turn off warnings into error for assertion checking.
-    with warnings.catch_warnings(record=True) as w:
-        warnings.simplefilter("default")
-        with pytest.raises(AssertionError):
-            server.status == "You can't compare with me"
-
-
 def test_server_status_assign_with_variant_warns():
     server = Server({})
     with warnings.catch_warnings(record=True) as w:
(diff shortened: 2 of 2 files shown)


mrocklin:
cc @Carreau

jakirkham:
For context this `Enum` was added in PR ( https://github.com/dask/distributed/pull/3853 ). AIUI this custom `__eq__` was about having a smooth change from using `str`s to `Enum`s in downstream libraries. Not sure if that already happened.

cc @lesteve @jacobtomlinson @jcrist

jacobtomlinson:
From a search through the Dask org it looks like strings are still used in dask-cloudprovider and dask-jobqueue.
We will need to update those before merging this.

jakirkham:
What sort of places are these `str`s used? Unfortunately searching for things like `"finished"` on GitHub were less helpful than expected. Though do see some use of [`Status` here and below]( https://github.com/dask/dask-cloudprovider/blob/09c28979fd283d483f46ffec7e5165d887a9083a/dask_cloudprovider/aws/tests/test_ec2.py#L91 ).

distributed/core.py
```diff
@@ -61,32 +61,6 @@ class Status(Enum):
...
     dont_reply = "dont-reply"
 
-    def __eq__(self, other):
-        """
-        Implement equality checking with backward compatibility.
-
-        If other object instance is string, we compare with the values, but we
-        actually want to make sure the value compared with is in the list of
-        possible Status, this avoid comparison with non-existing status.
-        """
```
mrocklin, line 71: Given the need for some backwards compatibility, I wonder if we might first try the following:

```python
try:
    return self.value == other.value
except AttributeError:
    ...
```

My experience is that try-except tends to be quite a bit faster than if/else in the try case, and this avoids the type check as well (which should be fine).

We can also migrate over all of the other projects if that will provide a large boost.  It isn't hard to do.

lesteve:
This should be easily fixable in dask-jobqueue, if I believe my original comment in https://github.com/dask/distributed/pull/3853#issuecomment-639282877. I have not done anything about it (and probably not planning to do anything about it in the next few days ...) though. Don't hesitate ping me if you open a PR in dask-jobqueue!

mrocklin:
See https://github.com/dask/dask-jobqueue/pull/476

mrocklin:
And https://github.com/dask/dask-cloudprovider/pull/200

Also https://github.com/dask/distributed/pull/4280 for the future

jakirkham:
Thanks for filing all of these Matt! 😄

jakirkham:
Looks like those PRs have been merged. Thanks all! 😄

Marking this as ready for review 🙂

jakirkham:
Is anything else needed here? Or are we good to merge?

mrocklin:
@jacobtomlinson @lesteve @guillaumeeb it would be convenient for the Scheduler performance work to drop backwards compatibility for string-valued statuses in the core dask.distributed package.  This would _break_ interoperability with dask-cloudprovider and dask-jobqueue prior to the recent PRs sent in last week.

Are you comfortable releasing soonish and dealing with the inevitable user complaints?

jakirkham:
It's worth noting the `Enum` object has been around since `distributed` version `2.19.0` ( https://github.com/dask/distributed/commit/1408ab59e691378f5a5c0299c8ac0bb4f42d4b22 ).

I don't think we are planning on releasing Distributed soon, but I could be wrong about that.

mrocklin:
Yeah, the enum has been around, but we didn't do a good job originally in
actually making the change throughout the subprojects consistently until
now.  This puts the users and maintainers of the subprojects in a
suboptimal position.

jakirkham:
Yep I understand. Just sharing info for the sake of transparency.

Carreau:
Sorry for the late reply; this does looks good to me; 

If you are a bit worries, you can make one release where you turn this warning on by default. It will raise on users' code but they will still be able to turn the error off with a warning filter to ignore.

jacobtomlinson:
Dask cloudprovider 0.5.1 is now on PyPI. Thanks @mrocklin for pushing this forwards.

jakirkham:
Dask-Jobqueue 0.7.2 has also been released.

Since it sounds like we are planning to do a release of Dask + Distributed soon. Maybe we can merge this after that release? That would allow for more uptake of the downstream projects' releases before this change shows up. Thoughts?

mrocklin:
That seems like a good idea.  Thank you for being flexible with the timing
on this one.  I know you'd like to have it out of your profiles

jakirkham:
Now that Dask + Distributed have been released and a week has passed, am curious are we comfortable merging this change? Or would we like to wait longer?

jakirkham:
Thoughts @mrocklin? 🙂

mrocklin:
Sure.  Let's go for it.

PR #4271, opened 2020-11-24
Precompute `hash` for `WorkerState`

jakirkham:
Despite `self.address` being a simple `str`. It appears calling `hash(...)` on it results in a bit of overhead. As we know that `self.address` should not be changing, we can precompute this once in `__init__` and simply `return` that result in `__hash__`. In all a simple change that should cutdown on the time spent selecting a `Worker` in `decide_worker`.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -239,6 +239,7 @@ class WorkerState:
         "bandwidth",
         "extra",
         "has_what",
+        "_hash",
         "last_seen",
         "local_directory",
         "memory_limit",
@@ -281,6 +282,7 @@ class WorkerState:
         self.versions = versions or {}
         self.nanny = nanny
 
+        self._hash = hash(address)
         self._status = Status.running
         self.nbytes = 0
         self.occupancy = 0
@@ -298,7 +300,7 @@ class WorkerState:
         self.extra = extra or {}
 
     def __hash__(self):
-        return hash(self.address)
+        return self._hash
 
     def __eq__(self, other):
         return type(self) == type(other) and self.address == other.address


distributed/scheduler.py
```diff
@@ -281,6 +281,7 @@ def __init__(
         self.versions = versions or {}
         self.nanny = nanny
 
+        self._hash = hash(address)
```
jrbourbeau, line 284: Could you add `"_hash"` to `WorkerState.__slots__`?

jakirkham: Thanks James! I completely missed that. Added above 🙂

jakirkham:
Please let us know if anything else is needed here :)

jrbourbeau (approved):
Thanks @jakirkham. Will merge after CI finishes 

For reference here's a quick benchmark. On the current main branch:

```python
In [1]: from distributed.scheduler import WorkerState

In [2]: ws = WorkerState("127.0.0.1")

In [3]: %timeit ws.__hash__()
125 ns ± 2.02 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

with the changes here:

```python
In [3]: %timeit ws.__hash__()
96.7 ns ± 1.08 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

jakirkham:
Thanks James! 😄

jrbourbeau:
Thanks @jakirkham! FWIW `test_broken_worker_during_computation` is a known flaky test (xref https://github.com/dask/distributed/issues/4173)

jakirkham:
Ah ok. Good to know. Thanks for flagging that 🙂

PR #4274, opened 2020-11-24
Move `nbytes` from `WorkerState` to `TaskState`

gforsyth:
Removing one less extra dictionary to track -- this was the source of
some tricky-to-debug issues over in #4264 and I thought I'd break it out
into a separate PR instead of jamming everything into one enormous PR.

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -573,7 +573,6 @@ async def test_clean(c, s, a, b):
     collections = [
         a.tasks,
         a.data,
-        a.nbytes,
         a.threads,
     ]
     for c in collections:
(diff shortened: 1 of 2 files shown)


jrbourbeau (commented):
Thanks @gforsyth! In general I'm in favor of this change as it further aligns the `TaskState` class on the scheduler and workers. It looks like there are some [tests that need to be updated](https://github.com/dask/distributed/pull/4274/checks?check_run_id=1449963129#step:8:613) to use `TaskState.nbytes` instead of `Worker.nbytes`

gforsyth:
> It looks like there are some [tests that need to be updated](https://github.com/dask/distributed/pull/4274/checks?check_run_id=1449963129#step:8:613) to use `TaskState.nbytes` instead of `Worker.nbytes`

Whoops!  Thanks, @jrbourbeau -- still having trouble with the test suite running cleanly on `$WORK_LAPTOP`

Issue #4273, opened 2020-11-24
Implementing `__hash__` and `__eq__` for `BatchedSend`


jakirkham:
In [`Scheduler.report`]( https://github.com/dask/distributed/blob/bbb23eb4a68ee0301db35dd18a916cb6b1cbafe4/distributed/scheduler.py#L2578 ) we [collect a `set` of `comms` for the `Client`s]( https://github.com/dask/distributed/blob/bbb23eb4a68ee0301db35dd18a916cb6b1cbafe4/distributed/scheduler.py#L2594-L2603 ). Looking more closely at the contents of [`self.client_comms`, it appears to just contain `BatchedSend` objects]( https://github.com/dask/distributed/blob/bbb23eb4a68ee0301db35dd18a916cb6b1cbafe4/distributed/scheduler.py#L2630-L2632 ). Putting these objects in a `set` will wind up using `__eq__` and `__hash__`. Neither of these appear to be defined by `BatchedSend`. As a result a good chunk of time is spent hashing these objects by some fallback mechanism before adding them to the `set` in `Scheduler.report`. Would be good to have implementations of `__eq__` and `__hash__` for `BatchedSend` to avoid this issue.

jakirkham:
Alternatively we might investigate another way to write this code so it doesn't rely on `hash`ing and comparing `BatchedSend` objects.

jakirkham:
Maybe another way to look at this is, when would a `BatchedSend` object be used for more than 1 `Client`?

cc @mrocklin (in case you have thoughts here 😉)

mrocklin:
It seems reasonable that there would be only one BatchedSend per Client.  In this situation I would try implementing this (hopefully it's cheap to do), and then seeing if any tests fail.  If no tests fail then we're good to go.

mrocklin:
We can also ask @jrbourbeau to do this if y'all are less familiar here.  James is fairly adept at these sorts of changes.

jakirkham:
Found a good way to do this with a `dict` ( https://github.com/dask/distributed/pull/4275 ), which should get us past this issue. It may be possible to simplify this further to a `list`, but that might not matter much.

PR #4275, opened 2020-11-24
Collect `report`'s `client_key`s in a `list`

jakirkham:
Fixes https://github.com/dask/distributed/issues/4273

Instead of collecting a `set` of `comms` where `BatchedSend` objects will need to be `hash`ed, use a `dict` with keys based on `Client` IDs and their `BatchedSend` object as values.  This should still ensure that we only pick up each `Client`'s `BatchedSend` object once. Though this will hash the `Client` IDs (`str`s) instead of `BatchedSend` objects, which should be more performant.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2590,13 +2590,13 @@ class Scheduler(ServerNode):
         if ts is None:
             # Notify all clients
             client_keys = list(self.client_comms)
-        elif client is not None:
+        elif client is None:
+            # Notify clients interested in key
+            client_keys = [c.client_key for c in ts.who_wants]
+        else:
             # Notify clients interested in key (including `client`)
             client_keys = [c.client_key for c in ts.who_wants if c.client_key != client]
             client_keys.append(client)
-        else:
-            # Notify clients interested in key
-            client_keys = [c.client_key for c in ts.who_wants]
 
         for k in client_keys:
             try:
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @jakirkham! This very looks sensible. Another option would be to keep a `set` of string client IDs and then use that set to index `self.client_comms` when sending `msg`. But similar to what you touched on in https://github.com/dask/distributed/issues/4273#issuecomment-733293561, I'm not sure how much that would speed things up. 

Just a heads up, there are some [linting issues](https://github.com/dask/distributed/pull/4275/checks?check_run_id=1450657183) in CI. You may consider installing `pre-commit` hooks

jakirkham:
Yep that seems like a good solution as well. As long as we are not calling `hash` on `BatchedSend`, I think we get past the main bottleneck. Noticed we were effectively checking `client_comms` for each key twice. Have rewritten it to check only once.

Oops sorry about that. Hadn't run `black` as I didn't expect any changes there. Have run with `black` now and it seems to have fixed the issue.

jakirkham:
Ok managed to rewrite this to collect the `client_key`s instead as suggested. Also managed to get this to use a `list` instead of a `set`. Since the contents being added to the `set` were either keys from a `dict` or values from a `set`, we already could guarantee there were no duplicates in either of those collections. The only potential duplicate is the `client` argument (when specified), which was easily to avoid duplicating by simply filtering that value from any other collection before adding to the `list`.

PR #4276, opened 2020-11-25
Add `__hash__` and `__eq__` for `ClientState`

jakirkham:
Since we periodically collect `ClientState` objects in `set`s or as keys in `dict`, we need to make sure they are hashable, which means implementing these methods. Of course there is a fallback if we don't, but it is preferable to have these implemented as we would expect instead.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -150,7 +150,7 @@ class ClientState:
         return self._hash
 
     def __eq__(self, other):
-        return type(self) == type(other) and self.client == other.client
+        return type(self) == type(other) and self.client_key == other.client_key
 
     def __repr__(self):
         return "<Client %r>" % (self.client_key,)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -137,13 +137,20 @@ class ClientState:
 
     """
 
-    __slots__ = ("client_key", "wants_what", "last_seen", "versions")
+    __slots__ = ("client_key", "_hash", "wants_what", "last_seen", "versions")
 
     def __init__(self, client, versions=None):
         self.client_key = client
+        self._hash = hash(client)
         self.wants_what = set()
         self.last_seen = time()
         self.versions = versions or {}
+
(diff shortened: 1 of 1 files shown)


jakirkham:
This just follows what we do with `WorkerState`, which is basically what we would want here as well.

https://github.com/dask/distributed/blob/f3488988c6ad31a888b66f6ce7de69afd66ee1ce/distributed/scheduler.py#L302-L306

distributed/scheduler.py
```diff
@@ -137,14 +137,21 @@ class ClientState:
...
+        self._hash = hash(client)
         self.wants_what = set()
         self.last_seen = time()
         self.versions = versions or {}
 
+    def __hash__(self):
+        return self._hash
+
+    def __eq__(self, other):
+        return type(self) == type(other) and self.client == other.client
```
jrbourbeau, line 153:
```suggestion
        return type(self) == type(other) and self.client_key == other.client_key
```

jrbourbeau (commented):
Thanks @jakirkham! Just one comment

distributed/scheduler.py
jrbourbeau: I'm somewhat surprised this didn't result in a test failure. I suppose we never do an equality check between `ClientState`s?

mrocklin: That doesn't surprise me

jrbourbeau: Fair enough. Thinking about it more, we're using these objects to keep track of state for a given client, not comparing across clients

jakirkham: Oops. Thanks James! Added the suggestion.

Yeah this has more to do with meeting the [requirements of a hashable object]( https://docs.python.org/3/glossary.html#term-hashable ).

> An object is hashable if it has a hash value which never changes during its lifetime (it needs a __hash__() method), and can be compared to other objects (it needs an __eq__() method). Hashable objects which compare equal must have the same hash value.

PR #4279, opened 2020-11-25
Transmit Layer annotations to scheduler

sjperkins:


sjperkins:
Depends on https://github.com/dask/dask/pull/6889

mrocklin:
I have an alternative here: https://github.com/dask/distributed/pull/4288/

It's not quite as comprehensive (annotations other than priority/retires/workers don't get transmitted) but it's simple and this approach covers what we can do today I think.

distributed/client.py
```diff
@@ -2552,7 +2552,86 @@ def _graph_to_futures(
...
+            if isinstance(dsk, HighLevelGraph):
+                for layer in dsk.layers.values():
+
+                    if layer.annotations is None:
+                        continue
+
+                    layer_priority = layer.annotations.get("priority", False)
+
+                    if layer_priority is not False:
+                        if isinstance(layer_priority, Number):
```
mrocklin, line 2576: FWIW I found using `if "priority" in layer.annotations:` more clear.  There is a bit less to follow.

I think that in general I optimize highly for adding as few terms as possible to code, perhaps more than most (you can probably see this in my previous comments.  This mostly comes from having to debug a high volume of code.  Every time someone introduces a new class or function or even variable that I need to learn it adds significantly to the time that it takes to quickly understand a piece of foreign code.  

I think that spelling everything out, like you do here, makes sense as we write code (it's nice to give everything a name and be explicit) but that it ends up hurting future reviewers who are in a rush.  I write more about this topic here: https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection

distributed/client.py
```diff
@@ -2552,7 +2552,86 @@ def _graph_to_futures(
...
+                            )
+
+                    layer_resources = layer.annotations.get("resources", False)
+
+                    if layer_resources:
+                        if isinstance(layer_resources, dict):
+                            for k in layer:
+                                annotation_resources[stringify(k)] = layer_resources
+                        else:
+                            raise TypeError("Expected resources to be a dict")
```
mrocklin, line 2633: If we're going to send along the annotations to the scheduler anyway then we might consider doing this work on the scheduler side.

mrocklin (commented):
A few comments.  Some are somewhat pedantic.  I hope that that is ok.

distributed/protocol/tests/test_highlevelgraph.py
```diff
@@ -82,3 +92,83 @@ async def test_shuffle(c, s, a, b):
...
+        assert_array_equal(r, 2)
+
+    assert isinstance(B.__dask_graph__().layers[A.name], BasicLayer)
+    assert isinstance(B.__dask_graph__().layers[B.name], Blockwise)
+
+    assert plugin.correct_qux == 25
+    assert plugin.correct_priorities == 25
+    assert plugin.correct_resources == 25
+
+    df = dd.from_pandas(
```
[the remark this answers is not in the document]
mrocklin: To be more clear, my goal is that when a test breaks, a maintainer who doesn't understand the underlying feature can see what is wrong within a minute or two.  This requires some care on our part to craft tests that are thorough, but also very simple.  We don't have to be perfect here, but taking some care is a great service to future maintainers.

distributed/protocol/tests/test_highlevelgraph.py
```diff
@@ -82,3 +92,83 @@ async def test_shuffle(c, s, a, b):
...
+    assert plugin in s.plugins
+
+    with dask.annotate(priority=fn, qux="baz"):
+        A = da.ones((10, 10), chunks=(2, 2))
+
+    with dask.annotate(resource="widget"):
+        B = A + 1
+
+    # TODO: replace with client.compute when it correctly supports HLG transmission
+    ret = c.get(B.__dask_graph__(), list(flatten(B.__dask_keys__())), sync=False)
```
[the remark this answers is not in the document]
sjperkins: Thanks, this is useful to know, I've changed the test cases to use this style.

distributed/protocol/tests/test_highlevelgraph.py
```diff
@@ -82,3 +92,83 @@ async def test_shuffle(c, s, a, b):
...
+
+    for r in await c.gather(ret):
+        assert_array_equal(r, 2)
+
+    assert isinstance(B.__dask_graph__().layers[A.name], BasicLayer)
+    assert isinstance(B.__dask_graph__().layers[B.name], Blockwise)
+
+    assert plugin.correct_qux == 25
+    assert plugin.correct_priorities == 25
+    assert plugin.correct_resources == 25
```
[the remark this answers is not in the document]
sjperkins: I've updated the TestAnnotationsPlugin to take expected values for matching annotations. In addition the above asserts now look like:

```python
assert plugin.priority_matches == B.npartitions
assert plugin.qux_matches == B.npartitions
assert plugin.resource_matches == B.npartitions
```

distributed/protocol/tests/test_highlevelgraph.py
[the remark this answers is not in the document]
sjperkins: I've split this into two test cases, an array test and a dataframe test. The TestAnnotationPlugin has been modified to take expected annotation values and count the number of matches.

distributed/client.py
sjperkins: I've removed the layer extraction code from ``Client._graph_to_futures`` and reintroduced it into ``Scheduler.update_graph``

distributed/client.py
sjperkins: Yep, no need to optimise per layer.

I've implemented this in the associated PR https://github.com/dask/dask/pull/6889/commits/e0e1dbf86e18a2fe2038fa470045cdea1f4fd155 where checks are performed on the existing distributed Client.{submit, persist, compute} keyword taxonomy.

distributed/protocol/tests/test_highlevelgraph.py
```diff
@@ -82,3 +92,83 @@ async def test_shuffle(c, s, a, b):
...
+        B = A + 1
+
+    # TODO: replace with client.compute when it correctly supports HLG transmission
+    ret = c.get(B.__dask_graph__(), list(flatten(B.__dask_keys__())), sync=False)
+
+    for r in await c.gather(ret):
+        assert_array_equal(r, 2)
+
+    assert isinstance(B.__dask_graph__().layers[A.name], BasicLayer)
+    assert isinstance(B.__dask_graph__().layers[B.name], Blockwise)
```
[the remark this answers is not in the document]
sjperkins: Removed

sjperkins:
> A few comments. Some are somewhat pedantic. I hope that that is ok.

No complaints here. The above makes sense to me and I think I understand where you're coming from, as I've been labelled "fussy about how things are done" upon occasion. ;-)

distributed/tests/test_client.py
```diff
@@ -6297,3 +6297,27 @@ def log_scheduler(dask_scheduler):
...
+        assert all(ts.retries == 2 for ts in s.tasks.values())
+
+        with dask.annotate(
+            workers=["fake"], allow_other_workers=True, resources={"GPU": 1}
+        ):
+            x = da.ones(10, chunks=(5,))
+
+        x = await x.persist()
+        print([{"GPU": 1} == ts.resource_restrictions for ts in s.tasks.values()])
+        assert b.data or a.data
```
mrocklin, line 6323: I recommend that we pull this off to a separate test, remove the print statement (or convert it to an assertion).  I'm also confused by the last assertion `assert b.data or a.data`.  What is the intent here?

distributed/scheduler.py
```diff
@@ -2012,6 +2014,31 @@ def update_graph(
...
+                    restrictions[k] = a["workers"]
+
+                if a.get("allow_other_workers", False):
+                    loose_restrictions.append(k)
+
+                if "retries" in a:
+                    retries[k] = a["retries"]
+
+                if "resources" in a:
+                    resources[k] = a["resources"]
```
mrocklin, line 2040: See comment in https://github.com/dask/dask/pull/6889/files#r534262218

distributed/tests/test_client.py
sjperkins: I've split these into 5 separate case, one for each distributed kwarg taxonomy (retries, priority, workers, allow_other_workers, resources).

> I'm also confused by the last assertion assert b.data or a.data.

The point was to prove that data ended up somewhere. I've removed it because its superfluous if the persist succeeded

distributed/scheduler.py
sjperkins: Updated with tranposed annotations! Thanks for the suggestion.

mrocklin:
This all seems good to me.  I'm happy to merge.

For future work (or now, if you feel like you're on a roll) we might consider assigning the annotations to the `TaskState` objects, such that the following might work:

```python
>>> ts.annotations
{"foo": 1}
```

The test failure here seems unrelated.  I'll push up a separate PR with a fix.

sjperkins:
> For future work (or now, if you feel like you're on a roll) we might consider assigning the annotations to the `TaskState` objects, such that the following might work:
> 
> ```python
> >>> ts.annotations
> {"foo": 1}
> ```

Done.

One of the future additions I'm considering is transmitting annotations to the workers, but I haven't looked into any of the details yet. I think it would be good to leave that to the scope of a new PR.

jakirkham:
It looks like this checked an empty file, `distributed/tests/test_highgraph.py`. Is that needed for some reason or should we drop that?

jrbourbeau:
Thanks for catching that, I believe it can be dropped

PR #4281, opened 2020-11-25
Drop extra type check in `_extract_serialize`

jakirkham:
Currently `_extract_serialize` performs a `type` check on its arguments and then when walking through any collections it performs `type` checks on any values it encounters. If any of those values are collections, it effectively performs the same `type` check twice. Thus for more deeply nested structures, these additional `type` checks will add up for every layer. To fix this, we perform a `type` check on the arguments provided to `extract_serialize` and do a bit of preparation on the arguments to `_extract_serialize`. Additionally when looping over values in `_extract_serialize`, we leverage this one `type` check to do any other preparation we may need to do with that info. As a result we are able to eliminate this overhead.

Here's a rough benchmark showing the effect of this change:

<details>
<summary>Before:</summary>

```python

   ...: msg = 11 * [10 * [2 * [5 * [data]]]]

943 µs ± 39 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

</details>

<details>
<summary>After:</summary>

```python

   ...: msg = 11 * [10 * [2 * [5 * [data]]]]

786 µs ± 9.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

</details>

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -1,6 +1,5 @@
 from array import array
 from functools import partial
-from itertools import repeat
 import traceback
 import importlib
 from enum import Enum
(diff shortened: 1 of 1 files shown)


distributed/protocol/serialize.py
```diff
@@ -445,27 +444,32 @@ def extract_serialize(x):
...
-
-def _extract_serialize(x, x2, ser, bytestrings, path=()):
     typ_x = type(x)
     if typ_x is dict:
         x_items = x.items()
+        x2 = dict()
     elif typ_x is list:
         x_items = enumerate(x)
-        x2.extend(repeat(None, len(x)))
+        x2 = len(x) * [None]
```
jakirkham, line 453: Also it is worth noting that this change of `list` construction is significantly faster on its own. By reducing the `type` checks performed, we are able to take advantage of this improvement by creating the `list` as we intend without filling up an existing `list` later.

```python
In [1]: %%timeit l = []
   ...: l.extend(repeat(None, 100))
   ...: 
   ...: 
641 ns ± 7.04 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [2]: %%timeit l = []
   ...: l = 100 * [None]
   ...: 
   ...: 
348 ns ± 10.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

mrocklin:
Fine by me

PR #4285, opened 2020-11-29
Make gather_dep robust to missing tasks

mrocklin:
Previously the gather_dep function assumed that tasks valid when
starting the function were still valid after an await call.  This may
not be the case.  This was fine, the rest of the work in this functions
was just about book-keeping for diagnostics, so errors didn't stop
things from halting.

Now we don't count the data sent for keys that we no longer know about.
This isn't entirely accurate, but should be fine.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2046,7 +2046,7 @@ class Worker(ServerNode):
                     )
 
                 total_bytes = sum(
-                    self.tasks[key].nbytes or 0
+                    self.tasks[key].get_nbytes()
                     for key in response["data"]
                     if key in self.tasks
                 )
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -749,17 +749,20 @@ class Worker(ServerNode):
     ##################
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @gforsyth

gforsyth (approved):
This looks good to me.  Non-blocking, but should we do `self.tasks[key].nbytes or DEFAULT_DATA_SIZE` instead?

distributed/worker.py
```diff
@@ -2046,7 +2046,9 @@ async def gather_dep(self, worker, dep, deps, total_nbytes, cause=None):
                     )
 
                 total_bytes = sum(
-                    self.tasks[key].nbytes or 0 for key in response["data"]
+                    self.tasks[key].nbytes or 0
```
jrbourbeau, line 2049: @gforsyth what about this?

```suggestion
                    self.tasks[key].get_nbytes()
```

gforsyth: Heh, that _would_ make a great deal of sense

jrbourbeau (approved):
Thanks for the PR @mrocklin and for reviewing @gforsyth! I'll merge after CI finishes

PR #4286, opened 2020-11-30
Print Traceback for an Exception on a worker

manuels:
If there is no traceback, it's a bit hard to debug the error, so let's add the traceback to the logging message.


manuels:
So it build looks fine now, but some CI test time out for no reason. Looks like hickups in the travis infratstructure, but I have no idea how to restart these tests.

PR #4289, opened 2020-12-01
Annotate `check_idle_saturated` for Cythonization

jakirkham:
In our profiling of the scheduler, we identified to main transitions that took a good chunk of time. These are `transition_processing_memory` and `transition_waiting_processing`. The former takes slightly longer than the other, but both are easily 2x slower than any transition that follows them. Both of them directly or indirectly make a call to `check_idle_saturated`. While this is not necessarily the worst bottleneck for either of them, it does stick out on the callgraph and stands a good chance of improving both transitions runtimes at once. Additionally `check_idle_saturated` includes a fair bit of code that simply crunches numbers and does not touch Python objects as much. It also isn't as dependent on the Cythonization of other classes as other functions in the profile are. So this makes it easier to Cythonize this piece of code without needing to touch too much other code.

Here we go through and annotate the local variables with types. Also we assign non-local variables accessed through attributes to local variables, which we type. Additionally we changed default argument values to be more friendly with C-style types. Initially we tried to type `self.idle`. However as [`self.idle` is a `sortedcontainers.SortedSet`]( https://github.com/dask/distributed/blob/9460e3fe1e0bcdb2daf6ebafe5335d536fa4f492/distributed/scheduler.py#L1292 ), this didn't work (as we need an actual Python `set` to type it). So we left this as untyped.

Should add as a good chunk of the time in `check_idle_saturated` is just spent doing `ws.status == Status.closed`, we still need PR ( https://github.com/dask/distributed/pull/4270 ) to cutdown on the time spent in this method.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4961,17 +4961,19 @@ class Scheduler(ServerNode):
         total_occupancy: double = self.total_occupancy
         avg: double = total_occupancy / total_nthreads
 
+        idle: set = self.idle
+        saturated: set = self.saturated
         if p < nc or occ / nc < avg / 2:
-            self.idle.add(ws)
-            self.saturated.discard(ws)
+            idle.add(ws)
+            saturated.discard(ws)
         else:
-            self.idle.discard(ws)
+            idle.discard(ws)
 
             pending: double = occ * (p - nc) / p / nc
             if p > nc and pending > 0.4 and pending > 1.9 * avg:
-                self.saturated.add(ws)
+                saturated.add(ws)
             else:
-                self.saturated.discard(ws)
+                saturated.discard(ws)
 
     def valid_workers(self, ts):
         """Return set of currently valid workers for key
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -85,6 +85,14 @@ from .stealing import WorkStealing
 from .variable import VariableExtension
 from .protocol.highlevelgraph import highlevelgraph_unpack
 
+try:
+    from cython import double, Py_ssize_t
+except ImportError:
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @mrocklin @quasiben

mrocklin:
The sortedcontainers issue is interesting.  It is convenient for scheduling logic to keep that around, but I'll be curious to learn how much that hurts us long-term.  

The one other issue to consider is if we want to start including cython/ctypes level typing in the scheduler.  This is still an experimental effort and so maybe a step backwards (readability gets mangled a bit).  I think it's probably still ok because it'll be easy to strip out later on if we decide that to go in a different direction.  I'm cc'ing @jcrist though, who I think is the person most likely to object to adding annotations like these.

jakirkham:
The `Enum` comparison presents a bigger issue ( https://github.com/dask/distributed/pull/4270 ).

Sure. Would be interested to get thoughts here and PR ( https://github.com/dask/distributed/pull/4290 ), Jim :)

jcrist:
> I'm cc'ing @jcrist though, who I think is the person most likely to object to adding annotations like these.

No objections from me. I find running mypy in CI a bit onerous, but adding annotations for optional cythonization seems totally fine.

mrocklin:
OK, in that case this looks good to me.  

@jakirkham I'm happy to merge this.  I notice that you still have the WIP flag up so I'll wait a bit unless you want to give the go-ahead.  Alternatively, please merge at will.

jakirkham:
Thanks all! 😄

I think this is ready. May come back to this function after we have typed some of the `*State` objects as well, but that isn't a blocker for this change (just a future improvement 😉).

PR #4290, opened 2020-12-01
[WIP] Annotate `ClientState` for Cythonization

jakirkham:
From profiling the `Scheduler`, we know that there are a series of `transition_*` functions that take the bulk of the time. These functions and functions they call primarily work with the various `*State` objects. In these functions they spend time accessing their attributes, hashing them, etc. At the Python level we have optimized these operations about as much as we can. Already we use `__slots__` on the `*State` objects to speed up access. Also we pre-compute their `hash` and store it during construction to speed up later `hash`ing of these objects. That said, working with the `*State` objects still takes a fair bit of time.

Thus it seems reasonable to try and optimize these objects with Cython. Analogous to PR ( https://github.com/dask/distributed/pull/4283 ) and PR ( https://github.com/dask/distributed/pull/4289 ) we try to leverage Cython's pure Python mode to minimize disruption. Just to get a taste for these kinds of changes, we pick the simplest `*State` type to start, `ClientState`. This only has a handful of attributes and while it does see a fair bit of usage it is not as heavily used as `WorkerState` or `TaskState`.

In order to benefit from annotating `class`es, we need to use one feature from Cython, `cclass`, to decorate the `class` in question. This has the same effect as using `cdef class` in Cython. We then define and type attributes using type annotations, which Cython converts into C level attributes in a `struct` with very fast access. By default these act like `cdef` attributes, so are only visible in C/Cython code when compiled (there is no affect in pure Python). We can make them visible in Python by either using Cython's `declare`, a `.pxd` file, or rolling our own `@properties` (which is also Cython's `declare` does for us). Have forgone this atm as it doesn't appear `ClientState` is used externally. This also helped with the next part. Additionally have typed a few method variables used by `ClientState`.

When trying to running with these changes to `ClientState`, we found any spots where a `ClientState` was assigned to a variable and annotated that variable accordingly. This was done by picking up on convention `cs` typically meant `ClientState`. Sometimes `c` meant `ClientState`, but those have been changed to `cs` for clarity. Finally in running the profile anything we might have missed was easy to pick up as there would be a `ClientState` attribute access error (since attributes are not visible to Python currently). So any other variable needing typing was caught and fixed that way.

After getting a sense of what kinds of issues would crop up and how to fix them, this went pretty smoothly. The code changes also remained fairly light. Some changes to `ClientState`. A few `import`s with fallbacks. Otherwise just typing a `ClientState` variable when it came up. This should help us prepare for the other `*State` objects as well. Would be interested to get others takes on these changes. 🙂


mrocklin:
From my perspective this looks pretty lightweight.  I'm pleasantly surprised :)

jakirkham:
Thanks! This is basically done on my end if we are ok with it. Wasn't sure to what extent the `*State` objects (and their attributes specifically) are expected to be visible outside the `Scheduler` (if at all). If that's not needed, will just fill in the missing docstring attributes and mark it ready.

jakirkham:
cc @quasiben (for vis)

mrocklin:
> Wasn't sure to what extent the *State objects (and their attributes specifically) are expected to be visible outside the Scheduler 

When you say "outside the scheduler" do you mean

- Outside of the scheduler process?  Like "do client processes ever consume TaskState objects?"  If so, the answer is "probably no"
-  Outside of the Scheduler class?  Then yes, the dashboard is a good example of another system in Dask that uses internal Scheduler state

jakirkham:
Sorry I mean does anything outside of this `scheduler.py` file do `from ... import ClientState` and then do something with a `ClientState`'s attributes. Like does `cs.client_key` (or any other attribute) get used outside of this file.

This applies to other `*State` objects as well. So am curious both in terms of this specific type and then generally the others.

mrocklin:
There are several uses of TaskState and WorkerState in the dashboard code.
Here is an example file, I recommend grepping for "ts."

https://github.com/dask/distributed/blob/master/distributed/dashboard/components/scheduler.py

jakirkham:
Ok thanks for that data point. We might want to adjust this approach for them.

To the second part, is `ClientState` used outside of this file (or even `distributed`)?

mrocklin:
In general scheduler state is used outside of scheduler.py.  Scheduler logic is spread to other files like stealing.py or locks.py intentionally to try to keep the core somewhat lightweight.  Is it possible to draw the Cython box a big bigger around multiple files?

It is probably doable to keep things contained within this repository though

jakirkham:
Ok thanks that helps. I'll keep that in mind going forward. Also will push some changes to address that here.

We could use Cython in more places. I don't think we need to yet, but it could happen. Stealing is a good example of where we may consider this eventually.

Independently we might want to start tracking what is intended to be Cythonized and how to enable it. Initially was thinking we could add an optional flag to `setup.py` to Cythonize a few files. If not passed, nothing would happen. If passed, it would Cythonize a few files (with a fallback if Cython isn't around).

Sure. Was more wondering whether we need to be cognizant of breakage. We are not entering that territory yet (and may not at all). Just trying to understand the constraints a bit more since we are discussing them.

jakirkham:
Have pushed a couple of commits to add `property`s for these attributes. This is one way to make sure they are visible to Python when Cythonizing this file. Though they are not the only way as noted in the OP (quoted below)

> We can make them visible in Python by either using Cython's `declare`, a `.pxd` file, or rolling our own `@properties` (which is also Cython's declare does for us).

The advantage of using `property`s is the code still looks pretty Pythonic. Also it ensures if the `_` attributes are used when Cythonized, we get an error if the variable using those attributes was not typed. IOW instead of potentially writing poor perf code, we just get an error. The downside is this a bit more boilerplate. Also there is a little more overhead experienced by anything using the `property`s in pure Python vs. the `_` attributes, which are in the slots.

If we prefer something more succinct while still remaining rather Pythonic, Cython's `declare` is a good option. The downside is it will mask inefficient attribute access in Cython as opposed to erroring as we have now with `property`s.

Using `.pxd` would move the declarative bits of Cython (`cdef class` and `cdef`/`cpdef`) into a separate file. On the one hand, this might avoid adding some typing bits to the `.py` file. However it increases the maintenance burden as we now have 2 files to keep up-to-date and in-sync. So seems to be the most error prone in the long run. It also raises the question of whether we should just scrap pure Python mode and use a `.pyx` file instead of a `.py` file.

Out of all of the options slightly prefer the `property` approach, but could go with the `declare` approach if we prefer something succinct. Would like to avoid the `.pxd` approach as this seems hard to maintain and moves us in a direction I don't think we want to go.

mrocklin:
This seems like a sensible path forward.  I'll be curious to see if this has an impact when we hit TaskState and WorkerState, which I suspect will be more significant and more work.

jakirkham:
For comparison have included both approaches in this PR's commits. Here is the `@property` approach ( https://github.com/dask/distributed/pull/4290/commits/a0e0c1c35e6b57989a117cfdd5acaff832d16f09 ) and here is the `declare` approach ( https://github.com/dask/distributed/pull/4290/commits/9000361f9b714e6564870d90e7d0afe1ca32d1b5 ).

jakirkham:
Ok this is ready for review. Please let me know if you have any other questions 🙂

mrocklin:
This looks good to me.  Thanks @jakirkham

distributed/scheduler.py
```diff
@@ -86,13 +86,18 @@
 from .protocol.highlevelgraph import highlevelgraph_unpack
 
 try:
-    from cython import double, Py_ssize_t
+    from cython import cclass, double, Py_hash_t, Py_ssize_t
 except ImportError:
     from ctypes import (
         c_double as double,
+        c_ssize_t as Py_hash_t,
```
jakirkham, line 93: For context, this is reasonable as [Python does the same thing itself]( https://github.com/python/cpython/blob/b3b98082c5431e77c64cab2c85525a804436b505/Include/pyport.h#L114 ). Though it shouldn't matter whether we get this right in the pure Python case as this doesn't have any effect on the implementation besides documentation. Just noting this to provide background as to how this was decided.

Issue #4263, opened 2020-11-23
Cluster lifecycle management


jacobtomlinson:
**TL;DR I'm going to play around with adding lifecycle features for Dask clusters for the next couple of days. This will likely result in design discussions and eventually many PRs to many repos, so wanted to track the overall conversation here.**

**My primary goal is to enable this. Then expand to many other uses.**

```console
$ dask-cluster list
CLUSTER ID        TYPE                              WORKERS       CREATED         STATUS
dask-123abc       dask_kubernetes.KubeCluster       6             4 hours ago     Running

$ dask-cluster scale dask-123abc 10
Scaled dask-123abc to 10 workers

$ dask-cluster delete dask-123abc
Are you sure Y/n: Y
Cluster dask-123abc deleted
```

---

I've spent a lot of time recently working on various cluster managers and feel like I'm in a good headspace to think about Dask cluster management as a whole. This issue may become a little sprawling as I try and unpack what I've been thinking about, so apologies!

Cluster lifecycle management is currently tricky and fractured. I would like to add some new features which will act as building blocks to improve things and gradually bring things in line.

### Current state

In the Dask ecosystem `LocalCluster` is the prototypical cluster implementation. It is often the first that folks come into contact with and it is used as a reference implementation for others. However it contains one core assumption that I've found to be problematic compared to other cluster managers; the lifecycle of the cluster is tied to the process which created it. If you create a `LocalCluster` in ipython and then close ipython the cluster will close because any subprocesses will exit.

Cluster implementations such as jobqueue, kubernetes, cloudprovider, gateway, etc all create external resources which are not tied to the lifecycle of the process. Jobs are submitted to scheduling systems, cloud resources are created, etc. Many of these projects attempt to replicate the functionality of `LocalCluster` by handling lifecycle within the cluster object.

- Dask Kubernetes and Dask Cloudprovider attempt to use `finalize` hooks to destroy any dangling resources. However this can be flaky and leave things running in some failure modes.
- Some of the newer libraries in Dask Cloudprovider do not attempt clean up at all and instead heavily document closing the cluster and the use of context managers. Some have manual CLI clean up commands for recovery from failure.
- Dask gateway has more advanced lifecycle management, but has reimplemented many existing cluster managers in order to do this.
- The `HelmCluster` manager in Dask Kubernetes is completely divorced from the lifecycle of the cluster, leaving the user to create and destroy the cluster with the `helm` CLI tool and only provides utility features such as connection, scaling and logging.
[...]

PR #4291, opened 2020-12-01
Add initial cluster lifecycle functionality

jacobtomlinson:
Related to #4263 which contains a long description of what I am hoping to achieve.

This PR has the following goals:
- Give all clusters names (defaults to a UUID)
- Add a high-level discovery method for finding existing Dask clusters
- Extend `SpecCluster` to support reconstructing a cluster object by name
- Add utility methods to get, list, scale and delete clusters by name
- Add a command-line tool to list, scale and delete clusters by name

### Discovery

Discovery works by inspecting the `dask_cluster_discovery` entrypoint and calling all registered methods. Each method should return a list of cluster names and the classes that they can be reconstructed with.

As a toy example in `dask-kubernetes` I've experimented with a discovery method which searched the Kubernetes API for Dask schedulers created by the Helm Chart and lists their name and the `HelmCluster` manager. I've also added a `from_name` classmethod to `HelmCluster` which will create a new object from that name. See dask/dask-kubernetes#285.

This was reasonably straight forward because `HelmCluster` already works by constructing an object referencing existing Kubernetes resources. This will be more involved to implement on other cluster managers and does rely on the scheduler running remotely.

Discovery is broken into two functions, `discover_cluster_names` and `discover_clusters`. The former calls all registered entrypoints and passes the results straight back, the latter goes a step further and reconstructs all cluster objects and passes those.

### Utilities

The `list_clusters`, `get_cluster`, `scale_cluster` and `delete_cluster` methods are inspired by the Dask Gateway API and make use of discovery to perform these operations.

The list method uses `discover_clusters` to get all cluster objects and then returns key information about that cluster, mostly derived from `scheduler_info`.

The other methods use `discover_cluster_names` and then only reconstruct the cluster with the matching name. Scale and delete then use the `.scale` and `.close` methods on the cluster object to perform their actions.

### Command line

The above utilities are also exposed via a `dask-cluster` CLI.

```
$ dask-cluster list                                                        
NAME            ADDRESS                   TYPE         WORKERS  THREADS  MEMORY     CREATED  STATUS 
mycluster.dask  tcp://10.51.100.43:32314  HelmCluster        3       36  299.72 GB  Unknown  Running
```

```
$ dask-cluster scale mycluster.dask 1
Scaled cluster mycluster.dask to 1 workers.
```

You can also list the registered discovery methods.

```
$ dask-cluster list-discovery        
NAME          PACKAGE          VERSION                   PATH                                                             
helmcluster   dask-kubernetes  0.11.0+7.g5211ebe.dirty   /Users/jtomlinson/Projects/dask/cluster-lifecycle/dask-kubernetes
proxycluster  distributed      2.30.1+1.g81b306c3.dirty  /Users/jtomlinson/Projects/dask/cluster-lifecycle/distributed
```

### Proxy Cluster

Reconstructing all cluster managers is not possible. For those like `LocalCluster` and `SSHCluster` the scheduler and workers are created as child processed of the Python process, so reconstructing them doesn't make sense.

Other cluster managers like `KubeCluster` and those in dask-jobqueue can also optionally run a local scheduler which is tightly coupled to the Python process that created them.

To add some functionality for these cluster managers I've also added `ProxyCluster`.

This is a very thin cluster manager which subclasses `Cluster` and therefore can be used in the same situations (such as 
connecting a client or calling `get_logs`). However you cannot scale or delete them. 

This is useful as it allows you to see them in the list command and also connect to them by name.


jacobtomlinson:
Closing in favour of https://github.com/dask-contrib/dask-ctl

PR #4292, opened 2020-12-01
Provide option to build Distributed with Cython

jakirkham:
This takes an initial pass at providing an option to build some portions of Distributed with Cython. Basically if we encounter `--with-cython` in `sys.argv`, we remove it and wrap the modules to compile in `Extension`s. If for whatever reason Cython isn't around, we add it `setup_requires` (though we could skip this and error instead if preferred). If we don't encounter `--with-cython`, we don't do anything.

### DIFF
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -47,6 +47,13 @@ jobs:
         shell: bash -l {0}
         run: conda list
 
+      - name: Optionally Cythonize
+        shell: bash -l {0}
+        run: |
+          if [[ "${{ matrix.python-version }}" = "3.7" ]]; then
+            python setup.py build_ext --with-cython
+          fi
+
       - name: Run tests
         shell: bash -l {0}
         env:
--- a/continuous_integration/environment-windows.yml
+++ b/continuous_integration/environment-windows.yml
@@ -7,6 +7,7 @@ dependencies:
   - bokeh!=2.0.0
   - click
   - cloudpickle
+  - cython
   - dask
   - lz4
   - ipykernel
--- a/setup.py
+++ b/setup.py
@@ -34,18 +34,24 @@ if cython:
     except ImportError:
         setup_requires.append("cython")
 
-    ext_modules.extend(
-        [
-            Extension(
-                "distributed.scheduler",
-                sources=["distributed/scheduler.py"],
-            ),
-            Extension(
-                "distributed.protocol.serialize",
-                sources=["distributed/protocol/serialize.py"],
-            ),
-        ]
-    )
+    cyext_modules = [
+        Extension(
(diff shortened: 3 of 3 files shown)


jakirkham:
cc @mrocklin @quasiben

jakirkham:
We also may want to pick a CI job or add a new one to test the Cythonized version. Please let me know if there is any preference.

mrocklin:
To the extent that I'm able to judge this seems fine.  Adding Cython to one of the builds sounds fine to me.  I suggest that we leave 3.8 alone, but otherwise I don't have  preference.

jakirkham:
Sounds good. Was thinking we could add this to 3.7.

jakirkham:
Have pushed commits to test Cython on GitHub Action CI's Python 3.7 job. Seems to be working ok.

Didn't do anything with Travis CI as that is being phased out. Plus it seems to take a really long time for queued jobs to start running.

mrocklin:
To the extent that I'm able to judge this seems fine to me.

jakirkham:
Thanks Matt! If anything else comes up, please let us know. Happy to follow up in new PRs as needed.

gforsyth:
I've never seen `setup_requires` do anything except ugly fail -- since it's only happening when `--with-cython` is passed, nbd I think, but if we start cythonizing more regularly (:+1: on that) we should move over to `pyproject.toml` for handling the setup dependencies

jakirkham:
For clarity, it is only passed if the user passes `--with-cython` and `cython` isn't installed. If the user already has Cython, it also does nothing (in addition to doing nothing if the flag is excluded).

Interesting it's normally worked for very simple cases like this for me. That said, happy to remove it and replace it with an error instead. Just figured we should have some kind of fallback behavior if the user tries to use Cython and it isn't there.

Yeah was looking to see if `pyproject.toml` had any way to specify an optional dependency, but didn't see one. Do you know of one? Alternatively I suppose we could just always install Cython. Though maybe some folks on more rare hardware might not like that if Cython isn't already built for their platform.

gforsyth:
`pyproject.toml` supports optional and conditional dependencies, but I don't _think_ that that works with the `[build-system]` section, which is where we'd want `cython` to live.  This must have been a thing somewhere before, I'll see if I can dig anything up.  
I think your current setup is all good, I just wanted to put in a warning that it might not be fine if we start making it a bit more aggressive.

jakirkham:
Yeah that last bit about how these pertain to `[build-system]` was unclear to me as well. Please let me know what you find out.

For sure, if we wanted to start requiring Cython (I don't think we are there yet), then agree we should just move to `pyproject.toml`.

Should add am happy to just drop the `setup_requires`. If we are worried about it, that seems like the right thing to do anyways.

PR #4294, opened 2020-12-02
Annotate `WorkerState` for Cythonization

jakirkham:
Analogous to PR ( https://github.com/dask/distributed/pull/4290 ) except instead of `ClientState`, this is annotating `WorkerState` and usages thereof. This is a bit longer simply because of how many attributes `WorkerState` has. That said, this just uses the same strategy here as was used for `ClientState`.


distributed/scheduler.py
```diff
@@ -370,54 +481,62 @@ def status(self, new_status):
...
+            name=self._name,
+            nthreads=self._nthreads,
+            memory_limit=self._memory_limit,
+            local_directory=self._local_directory,
+            services=self._services,
+            nanny=self._nanny,
+            extra=self._extra,
         )
-        ws.processing = {ts.key for ts in self.processing}
+        ws._processing = {ts.key for ts in self._processing}
```
jakirkham, line 508: This gives us an error now as this is typed as `dict` based on the docstring above and the default value, which is also a `dict`. However here we are assigning it a `set`. Submitted PR ( https://github.com/dask/distributed/pull/4295 ) to provide one way to correct this. Though I don't know that it is the right solution necessarily.

distributed/scheduler.py
```diff
@@ -282,30 +283,54 @@ class WorkerState:
 
     # XXX need a state field to signal active/removed?
 
+    _actors: set
+    _address: str
```
jakirkham, line 287: It seems like `address` should be a `str`, but ran into a few issues in the tests with that. Submitted PR ( https://github.com/dask/distributed/pull/4296 ) to hopefully address those.

distributed/scheduler.py
```diff
@@ -3513,8 +3646,9 @@ def workers_to_close(
         if n is None and memory_ratio is None:
             memory_ratio = 2
 
+        ws: WorkerState
         with log_errors():
-            if not n and all(ws.processing for ws in self.workers.values()):
+            if not n and all([ws._processing for ws in self.workers.values()]):
```
jakirkham, line 3651: Ideally this wouldn't need to instantiate the `list`, but ran into some issues with the type being picked up in Cython ( https://github.com/cython/cython/issues/3928 ). So decided to do it this way instead. Expect it doesn't matter much either way since we already track all `Worker`s anyways.

As a side note, we may get more value from the Cython side by converting this to a `for`-loop. That said, will save that for future work should we find this part of the code is slow.

distributed/scheduler.py
mrocklin: Merged.  Thanks

distributed/scheduler.py
mrocklin: Sounds like a sensible plan

distributed/scheduler.py
jakirkham: Thanks Matt! 😄

distributed/scheduler.py
jakirkham: It sounds like `int` is allowable in some cases. Though we can work to constrain this to a `str`. For now maybe we can relax this to `object` while we figure that out and then tighten this down later.

```suggestion
    _address: object
```

jakirkham:
This is ready for review!

cc @mrocklin @quasiben

distributed/scheduler.py
```diff
@@ -201,11 +201,12 @@ def versions(self):
         return self._versions
 
 
+@cclass
 class WorkerState:
     """
     A simple object holding information about a worker.
 
-    .. attribute:: address
+    .. attribute:: address: str | int
```
jakirkham, line 209: [From]( https://github.com/dask/distributed/pull/4302#discussion_r535324551 ) @mrocklin:

> I'm surprirsed that this can be an int. I would expect the variation to exist here for name, but address I would always expect to be a string today.

distributed/scheduler.py
```diff
@@ -282,78 +283,188 @@ class WorkerState:
 
     # XXX need a state field to signal active/removed?
 
+    _actors: set
+    _address: object
+    _bandwidth: int
```
jakirkham, line 288: [From]( https://github.com/dask/distributed/pull/4302#discussion_r535325646 ) @mrocklin:

> Sometimes we use int and sometimes Py_ssize_t, can you help explain why we choose one over the other? From a readability perspective I would prefer int everywhere. I'm guessing that this varies a bit on 32bit systems?

distributed/scheduler.py
jakirkham: Yeah I was surprised as well. Was running into some failures if this was restricted to `str` though. So just opted to type this as an `object` for now instead of getting too hung up on fixing the typing initially.

distributed/scheduler.py
jakirkham: Oops this should be `Py_ssize_t`. Thanks for catching that.

<hr>

Unfortunately Cython treats [`int` as a Python object]( https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#pep-484-type-annotations ):

> Note the use of cython.int rather than int - Cython does not translate an int annotation to a C integer by default since the behaviour can be quite different with respect to overflow and division.

For now I've just used `Py_ssize_t` as it is the largest signed integer available. So we are unlikely to run into issues with it. Though yeah this may vary between 32-bit and 64-bit systems. Figure if it is important to restrict this to a particular sized integer (like `int16`) we can do that in follow up work.

In general haven't been too focused on typing everything perfectly, but have instead tried to provide some type for everything and try to avoid `object` when possible. The goal being to generally get types across as much of the code as possible and tune afterwards.

jakirkham: Fixed in latest comment

jakirkham: Reverted and marked unresolved. Now am remembering why I used `int`. It seems like this should be an `int`, but sometimes it is a `float`.

For example, [the `Scheduler.bandwidth` is an `int` to start]( https://github.com/dask/distributed/blob/dcdb4651eb0331bb99caf3362c84d4878d94ba5b/distributed/scheduler.py#L1204 ). Though [`Scheduler.bandwidth_workers` contains `float`s]( https://github.com/dask/distributed/blob/dcdb4651eb0331bb99caf3362c84d4878d94ba5b/distributed/scheduler.py#L1205 ). Later on [the `Scheduler.bandwidth` becomes a `float`]( https://github.com/dask/distributed/blob/dcdb4651eb0331bb99caf3362c84d4878d94ba5b/distributed/scheduler.py#L1712-L1714 ).

Though the `float` value of [`Scheduler.bandwidth` infects values like `TaskState.nbytes`]( https://github.com/dask/distributed/blob/b7a2274f7ee8ac6a370604fbdd538c7a56d447af/distributed/tests/test_steal.py#L496 ), which isn't good.

Could use some guidance here. Do we expect `bandwidth` to be an `int`? If so, we should clamp down on some of these cases where it becomes a `float` by rounding it somehow and coercing it back to an `int`. If not, we should be more careful about how we mix `bandwidth` with other variables to ensure we get an `int` where we expect one.

jakirkham: Submitted PR ( https://github.com/dask/distributed/pull/4311 ) for that specific issue.

distributed/scheduler.py
jakirkham: Was originally trying to fix this issue along with `name` in PR ( https://github.com/dask/distributed/pull/4296 ). Wasn't sure if we wanted to fix `address` based on that discussion. As it sounds like we do, broke out that piece of the fix into PR ( https://github.com/dask/distributed/pull/4312 ).

jakirkham: Ok sorry I have no clue why this was causing issues before. Have switched back to `str`. Will see how it goes 🍀

distributed/scheduler.py
```diff
@@ -282,78 +283,188 @@ class WorkerState:
...
+    _address: str
+    _bandwidth: double
+    _extra: dict
+    _has_what: set
+    _hash: Py_hash_t
+    _last_seen: double
+    _local_directory: str
+    _memory_limit: Py_ssize_t
+    _metrics: dict
+    _name: object
```
jakirkham, line 296: This is an `object` as it could be a `str` or an `int`. See discussion in PR ( https://github.com/dask/distributed/pull/4296 ) for more context.

mrocklin:
This is in.  Thanks @jakirkham

jakirkham:
Thanks Matt! 😄

jakirkham:
Found a couple minor things that needed fixing. Submitted as PR ( https://github.com/dask/distributed/pull/4321 ).

PR #4295, opened 2020-12-02
Set `WorkerState.processing` w/`dict` in `clean`

jakirkham:
Based on [the documentation about `WorkerState.processing`]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L213-L226 ), this should be a `dict`. However here we are assigning it a `set`. This replaces it with a copy of the `dict`.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -386,7 +386,7 @@ class WorkerState:
             nanny=self.nanny,
             extra=self.extra,
         )
-        ws.processing = {ts.key for ts in self.processing}
+        ws.processing = {ts.key: cost for ts, cost in self.processing.items()}
         return ws
 
     def __repr__(self):


jakirkham:
I'm honestly not sure if this is the right thing to replace it with, but am a little clueless as to what this should be instead. Would be good if someone could check this and either confirm or suggest an alternative.

cc @mrocklin

jakirkham:
AFAICT this `.clean()` method is only used to create a [`KilledWorker`]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L5773 ) exception where [the `WorkerState` object returned from `.clean()` is passed as the `last_worker` object]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L2382 ), which is then [`pickle`d]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L2381 ). Guessing this is then communicated back to the `Client`.

Based on [the only test of `.clean()`]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/tests/test_scheduler.py#L1670-L1675 ), it seems we are mostly interested in making sure this object is compact before passing it over the wire.

jakirkham:
Based on the observations above, IIUC `.clean()` is simply trying to avoid passing a full `TaskState` object and instead pass along the `TaskState.key` attribute (a `str`). These are the keys of `processing`. The values are simply numbers. So shouldn't be too burdensome to `pickle` and may just have been left out accidentally. Have included these as well in the latest commit. Hopefully that captures the intent here, but please let me know if I'm missing something 🙂

mrocklin:
Thanks for catching this @jakirkham .  I suspect that it was just an oversight

jakirkham:
Ofc. Yeah that's my guess as well. Just wasn't sure initially how we should fix it. Think we came to a reasonable solution in the end.

PR #4296, opened 2020-12-02
[WIP] Use `str` for `Worker` `name` & `address` in tests

jakirkham:
It seems a few tests specified the `Worker`'s `name` and `address` with `int`s. However `Scheduler.coerce_address` (used in the tests) [doesn't actually allow `int`s]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L5113 ). Also [the `WorkerState`'s docs for `address` imply it is a `str`]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L208-L211 ), which is [also said of the `nanny` address]( https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/scheduler.py#L266-L268 ). Here we attempt to resolve this discrepancy by fixing these tests to use `str`s instead of `int`s.

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -579,7 +579,7 @@ async def test_coerce_address():
         s = await Scheduler(validate=True, port=0)
         print("scheduler:", s.address, s.listen_address)
         a = Worker(s.address, name="alice")
-        b = Worker(s.address, name=123)
+        b = Worker(s.address, name="123")
         c = Worker("127.0.0.1", s.port, name="charlie")
         await asyncio.gather(a, b, c)
 
@@ -598,12 +598,12 @@ async def test_coerce_address():
         assert s.coerce_address(a.address) == a.address
         # Aliases
         assert s.coerce_address("alice") == a.address
-        assert s.coerce_address(123) == b.address
+        assert s.coerce_address("123") == b.address
         assert s.coerce_address("charlie") == c.address
 
         assert s.coerce_hostname("127.0.0.1") == "127.0.0.1"
         assert s.coerce_hostname("alice") == a.ip
-        assert s.coerce_hostname(123) == b.ip
+        assert s.coerce_hostname("123") == b.ip
         assert s.coerce_hostname("charlie") == c.ip
         assert s.coerce_hostname("jimmy") == "jimmy"
 
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -787,7 +787,7 @@ async def start_cluster(
         Worker(
             s.address,
             nthreads=ncore[1],
-            name=i,
+            name=str(i),
             security=security,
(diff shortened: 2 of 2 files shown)


mrocklin:
Worker names can technically be ints.  This comes up with LocalCluster I suspect.  We can change this to make things always stringy.

mrocklin:
@jacobtomlinson I'm not sure if you have time these days, but if so you might be well positioned to think about this problem

jacobtomlinson:
In `SpecCluster` worker names are ints. When creating new workers they are also enumerated to find the next number.

https://github.com/dask/distributed/blob/f018007fba3b5d82e7b1a484ba758d2c302abd25/distributed/deploy/spec.py#L493-L509

@mrocklin do these numbers have any semantic meaning? Or are we just looking for unique names here? If its the latter perhaps this logic should be replaced with uuids?

mrocklin:
They need to be unique.  It's nice if they're easy to call out, as in 

```python
client.submit(func, x, workers=[0, 1, 2, 3])
client.submit(func, y, workers=[4, 5, 6, 7])
```

I would be against UUIDs for this reason.  My guess/hope is that there is some way to keep using ints in the spec, but to stringify names once they get to the worker

jacobtomlinson:
> It's nice if they're easy to call out

I'm curious how often this comes up in practice. I've never used this pattern myself.

Also I wonder if this could also be achieved more robustly with something like

```python
worker_names = [w for w in cluster.workers]
client.submit(func, x, workers=worker_names[:4])
client.submit(func, y, workers=worker_names[4:8])
```

jakirkham:
Interesting. Thanks for letting me know this is an expected case. Wouldn't have assumed that initially.

Using UUIDs or stringifying `int`s seem fine to me.

Mainly am interested in something with a fixed type here as opposed to several. The latter makes it harder to type things under-the-hood.

mrocklin:
> I'm curious how often this comes up in practice. I've never used this pattern myself.

It's not super common, but I've used it from time to time.  If you're familiar with MPI style programming it's also pretty natural.

> Also I wonder if this could also be achieved more robustly with something like

Yes, that would be more robust.  

I'm inclined towards stringification personally.  It's also the least disruptive.  I'm also personally unlikely to do this work short term though.  If others who are going to do this work have strong thoughts then I'm all for those thoughts.

jakirkham:
Broke out the `address` bit in PR ( https://github.com/dask/distributed/pull/4312 ) as it seems like we want that fix anyways.

jakirkham:
Going to abandon this. Not going to open a new issue since issue ( https://github.com/dask/distributed/issues/5743 ) already encapsulates this issue well.

Issue #4297, opened 2020-12-02
Handling exceptions in Worker plugins


pentschev:
I have noticed that `Worker` plugins exceptions are currently not handled, but they are checked and returned as a message to the caller in https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/worker.py#L2392-L2398

Specifically, plugins are awaited in `Worker.start()`: https://github.com/dask/distributed/blob/a192b22dece37b9e151b71aaaa129a409f5bc630/distributed/worker.py#L1142-L1144

I'm not sure what is supposed to happen with the error message, is it supposed to be passed to the scheduler somehow? Currently it seems to me that these message are silently ignored and not logged at all, but I am still not sure what is the expected way to handle this. For me, it looks as if we should check the status of message of each plugin, and if it raised an exception we should re-raise them within `Worker.start()` itself. Is that the right solution or is there something else we should do here? I'm happy to work on a fix either way, but I wanted to check what's the right way to fix it before spending the time on a potentially flawed fix.

mrocklin:
Failing on start seems sensible to me.

PR #4298, opened 2020-12-02
Raise plugin exceptions on Worker.start()

pentschev:
This PR raises plugin exception in `Worker.start()` that were so far being silently ignored.

Fixes https://github.com/dask/distributed/issues/4297


distributed/worker.py
```diff
@@ -1139,9 +1139,12 @@ async def start(self):
 
         setproctitle("dask-worker [%s]" % self.address)
 
-        await asyncio.gather(
+        plugins_msgs = await asyncio.gather(
             *[self.plugin_add(plugin=plugin) for plugin in self._pending_plugins]
         )
+        for msg in plugins_msgs:
+            if msg["status"] != "OK":
+                raise msg["exception"].data
```
pentschev, line 1147: I'm not sure if this is the right way of raising the exception that's being created by `distributed.core.error_message`. The object in `msg["exception"]` is delivered as a `distributed.protocol.Serialized`, but there seems to be no way exposed to extract its data other than just addressing the `data` attributed directly.

I'm not sure if this is also being handled correctly or if this is a bug in `error_message`. Any other suggestions here are appreciated.

distributed/tests/test_worker.py
```diff
@@ -407,6 +407,29 @@ def __str__(self):
...
+            self.msg = msg
+
+        def __str__(self):
+            return "MyException(%s)" % self.msg
+
+    class MyPlugin:
+        def setup(self, worker=None):
+            raise MyException("Foo")
+
+    s = await Scheduler(port=8007)
```
mrocklin, line 423: I recommend using `async with Scheduler(port=0) as s:

distributed/worker.py
mrocklin: I suspect that if you just remove the try-except and special handling in the plugin_add function that things may just work.  I suspect that the standard machinery around rpc will handle the exception comfortably.

mrocklin: I'm not sure though, this is just a guess.  I give it a 50/50 chance.

distributed/tests/test_worker.py
```diff
@@ -407,6 +407,29 @@ def __str__(self):
...
+    class MyException(Exception):
+        def __init__(self, msg):
+            self.msg = msg
+
+        def __str__(self):
+            return "MyException(%s)" % self.msg
+
+    class MyPlugin:
+        def setup(self, worker=None):
+            raise MyException("Foo")
```
jrbourbeau, line 421: Could we use a built-in exception instead, for example `raise ValueError("setup failed")`? Then later ensure that the expected error was raised with `pytest.raises(ValueError, match="setup failed")`

distributed/tests/test_worker.py
```diff
@@ -407,6 +407,29 @@ def __str__(self):
         assert "Bar" in str(e.__cause__)
 
 
+@pytest.mark.asyncio
+async def test_plugin_exception():
```
jrbourbeau, line 411:
```suggestion
async def test_plugin_exception(cleanup):
```

distributed/tests/test_worker.py
jrbourbeau, line 423: Generally we try to create `Scheduler` instance with a context manager in tests to ensure things are cleaned up properly.

```python
async with Scheduler() as s:
```

Also, is there a reason we need to specify `port=8007` here?

distributed/tests/test_worker.py
```diff
@@ -407,6 +407,29 @@ def __str__(self):
...
+        def __str__(self):
+            return "MyException(%s)" % self.msg
+
+    class MyPlugin:
+        def setup(self, worker=None):
+            raise MyException("Foo")
+
+    s = await Scheduler(port=8007)
+    with pytest.raises(MyException):
+        await Worker(
```
jrbourbeau, line 425: Same comment about using a context manager here

```python
```

jrbourbeau (commented):
Thanks for the PR @pentschev!

distributed/tests/test_worker.py
jrbourbeau: Oops, I see @mrocklin made a similar comment. Marking this one as resolved

pentschev:
@jrbourbeau @mrocklin you both had pretty much the same comments at the same time on the tests. Responding here, the test was mostly a copy of another test, given my first attempts with `async with ...` had failed and the previous solution with `await` worked, I chose to use that one. Now with both of your suggestions everything works, all test-related changes should be in, but please let me know if I missed something.

distributed/worker.py
pentschev: I think your suggestion would work in `Worker.start()`, but `plugin_add` is used also as a handler in https://github.com/dask/distributed/blob/master/distributed/worker.py#L662 and I think that would break it. Of course, I could have two different functions, one that raises exceptions in place for `Worker.start()` and another to be used as a handler.

There's also a chance that I'm overlooking something and your suggestion would work in all situations, is that the case?

mrocklin: Functions used as handlers handle exceptions in the same way that this function is doing manually.  I recommend trying things, seeing if everything passes, and reporting back.

pentschev: Sure, I did that now. What I wanted to prevent is breaking some untested behavior, just like this isn't being tested currently I believe that may break some other use case out there.

mrocklin: Ah, perhaps we could add a small test then?  I suspect that this would be the same as your current test, but you would use a pre-existing worker (just use gen_cluster rather than create things manually) and then call the client.register_plugin method (I think that that's the name but am not sure)

pentschev: Thanks Matt, I really wasn't aware how to use that properly. I now added a test for `register_worker_plugin`, but I had to revert my previous commit, as the exception type doesn't seem to get propagated with the `try`/`except` block you suggested removing. I guess the serialization step is indeed required for handlers.

If this is still too dirty, the only idea I have right now to avoid the `raise msg["exception"].data` is to have two different functions for `plugin_add`, one that serializes the exception, and another that doesn't.

pentschev:
Any other suggestions on improving this PR or are we ok with its current state? The test seems unrelated too, but correct me if I'm mistaken.

pentschev:
Friendly ping here, do we have anything left/other suggestions to improve this PR?

mrocklin:
Sorry for the delay @pentschev .  I and others have been busy recently.  I would like for someone (maybe me) to take a look at why the exception isn't being passed through.  I don't want our code to manually pack and unpack exceptions if we don't have to.  I personally will put this on my holiday TODO list (starts at the end of next week).  Hopefully I have time to look at it before then.

pentschev:
No worries, thanks @mrocklin . I'm also happy to do things here, I just want to be sure on what's the general "right" way of handling such cases. Also this isn't very high-priority, so no pressure, I just wanted to make sure it wasn't forgotten. :)

pentschev:
@jrbourbeau @mrocklin if there are no further changes needed, could we merge this?

The failing tests are https://github.com/dask/distributed/issues/5494 and the other pytest process seems to have died without reporting back final results in https://github.com/dask/distributed/runs/4304808476?check_suite_focus=true, but it seems unrelated to me.

quasiben:
@florian-jetter-by as you've done some work on the worker recently, would you have any issue with us merging this in today or tomorrow ?

mrocklin:
Pinging me these days probably isn't effective.

Also, cc'ing @fjetter (Florian's more standard username)

distributed/worker.py
```diff
@@ -1494,9 +1494,12 @@ async def start(self):
 
         setproctitle("dask-worker [%s]" % self.address)
 
-        await asyncio.gather(
+        plugins_msgs = await asyncio.gather(
             *(self.plugin_add(plugin=plugin) for plugin in self._pending_plugins)
         )
+        for msg in plugins_msgs:
+            if msg["status"] != "OK":
+                raise pickle.loads(msg["exception"].data)
```
fjetter, line 1502: I'm wondering what happens if multiple plugins raise an exception. do we at least log all of them? If not, should we iterate over all, log them and only afterwards raise?

FWIW, I think this would be a great application for exception groups proposed (and accepted) in [PEP654](https://www.python.org/dev/peps/pep-0654/). It's not implemented, yet, just a heads up.

distributed/worker.py
```diff
@@ -1494,9 +1494,12 @@ async def start(self):
 
         setproctitle("dask-worker [%s]" % self.address)
 
-        await asyncio.gather(
+        plugins_msgs = await asyncio.gather(
             *(self.plugin_add(plugin=plugin) for plugin in self._pending_plugins)
         )
```
fjetter, line 1499: It's also possible for `plugin_add` to raise other exceptions (e.g. unpickling errors). What would happen in these cases? I assume it is failing hard but I would like to confirm. Do we have tests for this?

pentschev: Added test for that in https://github.com/dask/distributed/pull/4298/commits/d3239133d1b0adc7c4732b5b60fd8fb60231927c .

distributed/worker.py
pentschev: Done in https://github.com/dask/distributed/pull/4298/commits/7552423149be6a123c35d6f642a969f976d0fbbc .

fjetter (approved):
LGTM. Once green-ish we can merge

pentschev:
Failing gpuCI tests are being addressed by https://github.com/dask/distributed/pull/5540 and https://github.com/dask/distributed/pull/5562 .

pentschev:
It seems Windows failures are unrelated to this. Failures from https://github.com/dask/distributed/runs/4409717678?check_suite_focus=true are https://github.com/dask/distributed/issues/5494, and `test_worker_reconnects_mid_compute`/`test_worker_reconnects_mid_compute_multiple_states_on_scheduler` that don't rely on worker plugins. The other failure from https://github.com/dask/distributed/runs/4409717724?check_suite_focus=true seems like a general timeout without most tests even running. And gpuCI failures are as per https://github.com/dask/distributed/pull/4298#issuecomment-985624323.

Are we good to merge this @fjetter ?

pentschev:
Can we merge this @fjetter @jrbourbeau ?

distributed/worker.py
```diff
@@ -1494,9 +1494,23 @@ async def start(self):
...
+        plugins_exceptions = []
+        for msg in plugins_msgs:
+            if msg["status"] != "OK":
+                exc = pickle.loads(msg["exception"].data)
```
gjoseph92, line 1503: The pickle-unpickle here feels a little cumbersome to me. I'm not sure I love this idea, but what if you add a `catch_errors=True` to `plugin_add`, and in `start` just set `catch_errors=False`?

```diff
diff --git a/distributed/worker.py b/distributed/worker.py
index 6e2fc56b..f4e3ed03 100644
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1494,22 +1494,12 @@ class Worker(ServerNode):
 
         setproctitle("dask-worker [%s]" % self.address)
 
-        plugins_msgs = await asyncio.gather(
-            *(self.plugin_add(plugin=plugin) for plugin in self._pending_plugins)
+        await asyncio.gather(
+            *(
+                self.plugin_add(plugin=plugin, catch_errors=False)
+                for plugin in self._pending_plugins
+            )
         )
-        plugins_exceptions = []
-        for msg in plugins_msgs:
-            if msg["status"] != "OK":
-                exc = pickle.loads(msg["exception"].data)
-                plugins_exceptions.append(pickle.loads(msg["exception"].data))
-        if len(plugins_exceptions) >= 1:
-            if len(plugins_exceptions) > 1:
-                logger.error(
-                    "Multiple plugin exceptions raised. All exceptions will be logged, the first is raised."
-                )
-                for exc in plugins_exceptions:
-                    logger.error(repr(exc))
-            raise plugins_exceptions[0]
 
         self._pending_plugins = ()
 
@@ -3182,7 +3172,7 @@ class Worker(ServerNode):
     def run_coroutine(self, comm, function, args=(), kwargs=None, wait=True):
         return run(self, comm, function=function, args=args, kwargs=kwargs, wait=wait)
 
-    async def plugin_add(self, comm=None, plugin=None, name=None):
+    async def plugin_add(self, comm=None, plugin=None, name=None, catch_errors=True):
         with log_errors(pdb=False):
             if isinstance(plugin, bytes):
                 plugin = pickle.loads(plugin)
@@ -3204,6 +3194,8 @@ class Worker(ServerNode):
                     if isawaitable(result):
                         result = await result
                 except Exception as e:
+                    if not catch_errors:
+                        raise
                     msg = error_message(e)
                     return msg
```
If you wanted to maintain the behavior you have here where multiple plugin exceptions can be handled and logged, then you could pass `gather(..., return_exceptions=True)` and handle them similarly as here, just without the unpickling. But I'm not sure if this functionality is actually necessary.

distributed/worker.py
```diff
@@ -1504,9 +1504,22 @@ async def start(self):
...
-        await asyncio.gather(
-            *(self.plugin_add(plugin=plugin) for plugin in self._pending_plugins)
+        plugins_exceptions = await asyncio.gather(
+            *(
+                self.plugin_add(plugin=plugin, catch_errors=False)
+                for plugin in self._pending_plugins
+            ),
+            return_exceptions=True,
         )
+        if len(plugins_exceptions) >= 1:
```
gjoseph92, line 1514: I think `plugins_exceptions` is not just the exceptions; it contains all the values returned by `plugin_add`, including OK messages. I think you'd want `plugins_exceptions = [x for x in plugins_results if isinstance(x, Exception)]` first.

pentschev: I don't think so, that was the case with `plugins_msgs` (the previous pickled version), but now adding a line just above this `print(f"plugins_exceptions: {plugins_exceptions}")`, I see the following for `distributed/tests/test_worker.py::test_plugin_exception`:

```
plugins_exceptions: [ValueError('Setup failed')]
```

and for `distributed/tests/test_worker.py::test_plugin_internal_exception` I see:

```
plugins_exceptions: [UnicodeDecodeError('utf-8', b'orrupting pickle\x80\x04\x95e\x02\x00\x00\x00\x00\x00\x00\x8c\x17cloudpickle.cloudpickle\x94\x8c\r_builtin_type\x94\x93\x94\x8c', 16, 17, 'invalid start byte')]
```

gjoseph92: Try having the plugin _not_ raise an exception and see what happens.

pentschev: Yes, you're absolutely right, thanks for pointing it out. This is fixed now by https://github.com/dask/distributed/pull/4298/commits/dc3aaeb6932103f4b29df3ee0b6ad956b98eb48f. Could you take another look when you have a chance?

gjoseph92 (commented):
LGTM besides this small nit

distributed/worker.py
```diff
@@ -1504,9 +1504,26 @@ async def start(self):
...
+            *(
+                self.plugin_add(plugin=plugin, catch_errors=False)
+                for plugin in self._pending_plugins
+            ),
+            return_exceptions=True,
         )
+        plugins_exceptions = []
+        for msg in plugins_msgs:
+            if isinstance(msg, Exception):
+                plugins_exceptions.append(msg)
```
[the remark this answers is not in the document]
pentschev: Good idea, applied suggestion.

pentschev:
A quick look at the failing tests seem to point all failures to timeouts. Once the release is complete, can we get this merged?

quasiben:
With the release completed for 2022.01.0  and the last comments addressed I'm going to merge in this PR.  

Thank your @pentschev for the work and @fjetter / @gjoseph92 for the reviews

pentschev:
Thanks everyone who reviewed and @quasiben for merging!

PR #4299, opened 2020-12-02
Check traceback object can be unpickled

jakirkham:
Just as we check if the exception can be round-tripped through pickle, check the traceback can as well. If that does, work proceed to fallback handling as before for the traceback.


jakirkham:
Basing this off of what we do a few lines above

https://github.com/dask/distributed/blob/d114d3a0cc401a8aea0e7a0832aca5b4404f3c66/distributed/core.py#L1137-L1138

jakirkham:

PR #4302, opened 2020-12-03
Annotate `Task*` objects for Cythonization

jakirkham:
Analogous to PR ( https://github.com/dask/distributed/pull/4290 ) ( https://github.com/dask/distributed/pull/4294 ) except this is annotating `TaskState`, `TaskPrefix`, and `TaskGroup`. Plus all usages thereof. This is a bit longer simply because of how many attributes `TaskState` has and how frequently it is used. That said, it follows the same pattern as was seen with the other two and so shouldn't be too surprising.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -512,7 +512,7 @@ class WorkerState:
             extra=self._extra,
         )
         ts: TaskState
-        ws._processing = {ts.key: cost for ts, cost in self._processing.items()}
+        ws._processing = {ts._key: cost for ts, cost in self._processing.items()}
         return ws
 
     def __repr__(self):
@@ -694,7 +694,7 @@ class TaskGroup:
     def add(self, o):
         ts: TaskState = o
         self.states[ts.state] += 1
-        ts.group = self
+        ts._group = self
 
     def __repr__(self):
         return (
@@ -1211,7 +1211,7 @@ def _legacy_task_key_set(tasks):
     Transform a set of task states into a set of task keys.
     """
     ts: TaskState
-    return {ts.key for ts in tasks}
+    return {ts._key for ts in tasks}
 
 
 def _legacy_client_key_set(clients):
@@ -1235,7 +1235,7 @@ def _legacy_task_key_dict(task_dict):
     Transform a dict of {task state: value} into a dict of {task key: value}.
     """
     ts: TaskState
-    return {ts.key: value for ts, value in task_dict.items()}
+    return {ts._key: value for ts, value in task_dict.items()}
 
 
 def _task_key_or_none(task):
@@ -2034,7 +2034,7 @@ class Scheduler(ServerNode):
             for ts in list(self.unrunnable):
                 valid = self.valid_workers(ts)
                 if valid is True or ws in valid:
-                    recommendations[ts.key] = "waiting"
+                    recommendations[ts._key] = "waiting"
 
             if recommendations:
                 self.transitions(recommendations)
@@ -2218,8 +2218,8 @@ class Scheduler(ServerNode):
             ts = self.tasks.get(k)
             if ts is None:
                 ts = self.new_task(k, tasks.get(k), "released")
-            elif not ts.run_spec:
-                ts.run_spec = tasks.get(k)
+            elif not ts._run_spec:
+                ts._run_spec = tasks.get(k)
 
             touched_keys.add(k)
             touched_tasks.append(ts)
@@ -2230,7 +2230,7 @@ class Scheduler(ServerNode):
         # Add dependencies
         for key, deps in dependencies.items():
             ts = self.tasks.get(key)
-            if ts is None or ts.dependencies:
+            if ts is None or ts._dependencies:
                 continue
             for dep in deps:
                 dts = self.tasks[dep]
@@ -2268,14 +2268,14 @@ class Scheduler(ServerNode):
             for a, kv in annotations.items():
                 for k, v in kv.items():
                     ts = self.tasks[k]
-                    ts.annotations[a] = v
+                    ts._annotations[a] = v
 
         # Add actors
         if actors is True:
             actors = list(keys)
         for actor in actors or []:
             ts = self.tasks[actor]
-            ts.actor = True
+            ts._actor = True
 
         priority = priority or dask.order.order(
             tasks
@@ -2284,7 +2284,7 @@ class Scheduler(ServerNode):
         if submitting_task:  # sub-tasks get better priority than parent tasks
             ts = self.tasks.get(submitting_task)
             if ts is not None:
-                generation = ts.priority[0] - 0.01
+                generation = ts._priority[0] - 0.01
             else:  # super-task already cleaned up
                 generation = self.generation
         elif self._last_time + fifo_timeout < start:
@@ -2296,14 +2296,14 @@ class Scheduler(ServerNode):
 
         for key in set(priority) & touched_keys:
             ts = self.tasks[key]
-            if ts.priority is None:
-                ts.priority = (-(user_priority.get(key, 0)), generation, priority[key])
+            if ts._priority is None:
+                ts._priority = (-(user_priority.get(key, 0)), generation, priority[key])
 
         # Ensure all runnables have a priority
-        runnables = [ts for ts in touched_tasks if ts.run_spec]
+        runnables = [ts for ts in touched_tasks if ts._run_spec]
         for ts in runnables:
-            if ts.priority is None and ts.run_spec:
-                ts.priority = (self.generation, 0)
+            if ts._priority is None and ts._run_spec:
+                ts._priority = (self.generation, 0)
 
         if restrictions:
             # *restrictions* is a dict keying task ids to lists of
@@ -2314,21 +2314,21 @@ class Scheduler(ServerNode):
                 ts = self.tasks.get(k)
                 if ts is None:
                     continue
-                ts.host_restrictions = set()
-                ts.worker_restrictions = set()
+                ts._host_restrictions = set()
+                ts._worker_restrictions = set()
                 for w in v:
                     try:
                         w = self.coerce_address(w)
                     except ValueError:
                         # Not a valid address, but perhaps it's a hostname
-                        ts.host_restrictions.add(w)
+                        ts._host_restrictions.add(w)
                     else:
-                        ts.worker_restrictions.add(w)
+                        ts._worker_restrictions.add(w)
 
             if loose_restrictions:
                 for k in loose_restrictions:
                     ts = self.tasks[k]
-                    ts.loose_restrictions = True
+                    ts._loose_restrictions = True
 
         if resources:
             for k, v in resources.items():
@@ -2338,7 +2338,7 @@ class Scheduler(ServerNode):
                 ts = self.tasks.get(k)
                 if ts is None:
                     continue
-                ts.resource_restrictions = v
+                ts._resource_restrictions = v
 
         if retries:
             for k, v in retries.items():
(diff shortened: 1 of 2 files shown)


mrocklin:
I'm excited about seeing this one in.  It's probably too early, but I'd be
curious to see how this changes profiles, if at all.

jakirkham:
Both PR ( https://github.com/dask/distributed/pull/4294 ) and this help based on my own local profiling and looking at the call graphs. Though the other thing they do is unlock further optimizations in the Scheduler transition methods. Since all of those methods are just working with these 3 objects `ClientState`, `WorkerState`, and `TaskState`. I think once we get these in, we will want to take another look at line profiling of the transition methods and see what sticks out. There are different ways we can structure the code and make additional annotations so that Cython can optimize it more effectively.

Even just the `ClientState` and `check_idle_saturated` changes helped as evidenced by the nightly profile ( https://github.com/quasiben/dask-scheduler-performance/issues/39 ).

Anyways if you have time to review PR ( https://github.com/dask/distributed/pull/4294 ) tomorrow that will be really helpful. This PR is 95% there, but there is one or two more things that it may still need. So will be working on them as well.

distributed/scheduler.py
```diff
@@ -201,11 +201,12 @@ def versions(self):
         return self._versions
 
 
+@cclass
 class WorkerState:
     """
     A simple object holding information about a worker.
 
-    .. attribute:: address
+    .. attribute:: address: str | int
```
mrocklin, line 209: I'm surprirsed that this can be an int.  I would expect the variation to exist here for name, but address I would always expect to be a string today.

distributed/scheduler.py
```diff
@@ -282,78 +283,188 @@ class WorkerState:
 
     # XXX need a state field to signal active/removed?
 
+    _actors: set
+    _address: object
+    _bandwidth: int
```
mrocklin, line 288: Sometimes we use int and sometimes Py_ssize_t, can you help explain why we choose one over the other?  From a readability perspective I would prefer int everywhere.  I'm guessing that this varies a bit on 32bit systems?

distributed/scheduler.py
```diff
@@ -657,134 +778,277 @@ class TaskState:
     :      The group of tasks to which this one belongs.
     """
 
+    _key: str
+    _hash: Py_hash_t
+    _prefix: object
```
mrocklin, line 783: TaskPrefix maybe

distributed/scheduler.py
```diff
@@ -657,134 +778,277 @@ class TaskState:
...
+    _run_spec: object
```
mrocklin, line 784: bytes maybe

distributed/scheduler.py
```diff
@@ -657,134 +778,277 @@ class TaskState:
...
+    _priority: tuple
+    _state: str
```
mrocklin, line 786: Should we make this into an enum?  Would Cython prefer that?

distributed/scheduler.py
```diff
@@ -657,134 +778,277 @@ class TaskState:
...
+    _traceback: object
+    _exception_blame: object
+    _suspicious: Py_ssize_t
+    _host_restrictions: set
+    _worker_restrictions: set
+    _resource_restrictions: dict
+    _loose_restrictions: bool
+    _metadata: dict
+    _actor: bool
+    _group: object
```
mrocklin, line 808: TaskGroup maybe

mrocklin (commented):
Some minor comments here.  In general all of this looks straightforward.  

I'm curious about a couple of the types, but those can wait as well.  

Also beware, you're likely to get a few conflicts when we merge in the annotations work https://github.com/dask/distributed/pull/4279

jakirkham:
Since this just includes the `WorkerState` PR, maybe we should work on that one first? 😉 Unless we’d rather just work with both of them in this PR? 🙂

mrocklin:
I don't have a preference where the work is done.  It'd be nice to merge one and then the other for git history.  It was easier for me to comment here just because everything was here.

jakirkham:
Understood. If we'd like to merge two separate PRs, would suggest to have `WorkerState` comments in that PR at least. Otherwise I think once that PR gets merged the comments here may wind up being artificially "resolved" since that code is no longer here, which may make hard for us to ensure we have addressed all of them.

That said, if we find it simply more convenient to work here, it may make more sense to close out that PR and do all the work here. It probably gets harder to disambiguate where changes apply if we start pushing both `TaskState` and `WorkerState` changes here.

distributed/scheduler.py
jakirkham: Moving to this thread ( https://github.com/dask/distributed/pull/4294/files#r535542496 )

distributed/scheduler.py
jakirkham: Moved to [this thread]( https://github.com/dask/distributed/pull/4294/files#r535543341 )

jakirkham:
Copied the `WorkerState` comments to PR ( https://github.com/dask/distributed/pull/4294 ). Let's follow up on those over there 🙂

distributed/scheduler.py
jakirkham: Since we haven't Cythonized that, it won't have any effect. Also we run into issues as it hasn't been defined yet at this point in the file.

distributed/scheduler.py
jakirkham: Interesting. This came from the [docstring above]( https://github.com/dask/distributed/blob/dcdb4651eb0331bb99caf3362c84d4878d94ba5b/distributed/scheduler.py#L438 ). Later in the code it was implied this [could be a `dict`, but might not be]( https://github.com/dask/distributed/blob/dcdb4651eb0331bb99caf3362c84d4878d94ba5b/distributed/scheduler.py#L2822-L2823 ). Seems like it is not very clear what it may be so perhaps it is best to leave as `object` for now.

distributed/scheduler.py
jakirkham: Cython can't do much with a Python `Enum`. There are `enum`s in Cython that it does expose to the Python layer optionally. Though these are not available in pure Python mode unfortunately ( https://github.com/cython/cython/issues/3923 ).

For now I think it is ok to just type this as a `str` as part of our broad effort to type things. We can then revisit specifics once everything is in and we have had an opportunity to profile more closely.

distributed/scheduler.py
jakirkham: Same story as with `TaskPrefix`. Since we haven't Cythonized that, it won't have any effect. Also we run into issues as it hasn't been defined yet at this point in the file.

distributed/scheduler.py
mrocklin: Sounds good.  We check and change state pretty often.  I wouldn't be surprised if this has some effect in the future. 

Are pure Python enums something that would make sense to request upstream, or is this likely to be hard to achieve with cython?

jakirkham: Yeah it may. I think line profiling after this lands makes sense and should hopefully guide us to which of several things we should further optimize. It could be `Status` or it could be something else.

Seems like a reasonable request to me (though maybe I'm biased as I already made that request 😄). Have no idea if it will be easy or not. Certainly hope it is doable.

In any event, I think at a first pass having `str` here is fine. Am more interested in broadly typing things first and worrying about more tuning in subsequent PRs after more profiling.

jakirkham:
Rebased on `master` to get rid of the `WorkerState` bits.

distributed/scheduler.py
jakirkham: I went ahead and put this in quotes, which avoids definition ordering issue. It's still treated as `object` for now though.

distributed/scheduler.py
jakirkham: I went ahead and put this in quotes, which avoids definition ordering issue. It's still treated as `object` for now though.

jakirkham:
Do we want to handle Cythonization of `TaskPrefix` and `TaskGroup` in this PR? They are pretty small classes and don't see as much use as `TaskState`. So think it should be straightforward to do. That said, don't want to overload this PR since it already has a lot of changes. WDYT?

Also would be good to get your thoughts on `run_spec` as the typing seems unclear ( https://github.com/dask/distributed/pull/4302#discussion_r535771684 ). Am also ok just leaving it as `object` if we are not sure. We can always follow up later if needed.

Thanks for the feedback thus far 🙂

mrocklin:
No preference on TaskPrefix/TaskGroup.  I think that you probably care more about git history than I do.  

It would be good to make `run_spec` more consistent, but it's rarely accessed, so I don't think that this is a priority.

jakirkham:
Was more concerned about how that affects the review load for you. No strong feelings on commit history from me.

Ok let's skip `run_spec` for now and we can revisit after this lands if it becomes important.

mrocklin:
Review of this stuff is pretty simple.  There isn't any tricky logic to go over.  I'm generally happy.

jakirkham:
Found I was having issues getting Cython to use the type definitions correctly when quoted. So am proposing just swapping the ordering of `Task*` `class` definitions around to make things easier. ( https://github.com/dask/distributed/pull/4302 )

jakirkham:
Need to flip the order of `TaskPrefix` and `TaskGroup`. Didn't realize the latter has an attribute that holds the former. Fixing with PR ( https://github.com/dask/distributed/pull/4319 ).

mrocklin:
I was toying around with this PR, looking at the results of `cython distributed/scheduler.py --annotate` it was fun to play around with some of the `transition_*` functions, drop in a couple of type annotations, and watch the yellow go away.  

I'm curious, what's next after this?  The full `Scheduler` class?

mrocklin:
Ah, after toying around a bit I see that we might need to do `core.py::Server` and `node.py::ServerNode`.  OK, I'm going to step back and leave this to the professionals :)

jakirkham:
Yeah it's definitely fun to play with. Things like `ts._prefix._groups` turn into `__pyx_v_ts->_prefix->_groups` at the C level. So attribute access in Python becomes the same as accessing fields in C `struct`s.

At least based on what we have seen with the call graphs, the next steps would be going back through the `transition` functions and annotating those more thoroughly. We might also transform some generator expressions into `for`-loops. Cython is pretty good at turning them into C `for`-loops.

Yeah Cython doesn't like having Cython extension types inherit from Python classes. Mentioned to Ben that we could look at refactoring out some of the `Scheduler` content into a Cython extension type class, which the `Scheduler` would inherit from. IDK how well this would work in practice, but it's an idea.

Right now have just been fixing up test failures here. Am still wrapping my head around where else `TaskState` is used.

mrocklin:
> Mentioned to Ben that we could look at refactoring out some of the Scheduler content into a Cython extension type class, which the Scheduler would inherit from. IDK how well this would work in practice, but it's an idea.

In principle the scheduler would be split into a state management system (cythonized) and as event management system (networking code).  That was never part of the design though and so things are pretty tangled.  I would expect getting this right would take someone a while.  I would love to be surprised otherwise though :)

The transition functions are the thing to look at if that's the short-term goal.  My guess is that methods like `report_on_key` will have to put a message on some other system, like a Queue, that the networking code then waits on.

jakirkham:
Ok this should be good to go. Please let me know if anything else is needed 🙂

quasiben:
I looked over the scheduler and seems like you are following the same techniques you laid out earlier -- _var with type and property decorators for Python access of the variables.  Thanks @jakirkham !!

Merging in now

jakirkham:
Thanks Ben! 😄

If anything else comes up, happy to follow up in a new issue/PR as appropriate.

distributed/scheduler.py
jakirkham: One other thing I forgot to note, Cython does `intern` any `str` literals it sees by default as part of module initialization (IOW at `import` time), which it hangs onto globally (so they don't go out of scope). That way, when we do something like `ts._state = "running"`, Cython will already have `intern`ed `"running"` and assigned the `intern`ed `str` variable to `ts._state`. Further when we later do something like `assert ts._state == "running"`, this `"running"` will also refer to the same `intern`ed `str`. As a result comparisons of `str`s defined with literals should benefit from the `intern` speedup.

Though there is a caveat. This does not apply for any dynamically generated `str`s. So `"My name is %s" % s` will construct a `str` at runtime that is not `intern`ed.

PR #4307, opened 2020-12-03
WIP - Optionally use offload executor in worker

mrocklin:
Normally we have one separate thread pool executor for deserializing data
and another thread pool executor in a worker for execution.
Sometimes this presents a problem because some data types want to be
used in the thread in which they were created.  One example of this is
TensorFlow graphs, but there are others.

One way to resolve this is to reuse the same executor in both
situations, and ensure that it has only one thread.  This means that
execution and deserialization will block each other (not great) but that
user data will always be operated on in one thread only.

This commit implements that, and drafts up a small test.  However, it is
still broken because Dask can be clever in some situations and
deserialize directly on the event loop.  This happens for a few reasons
today:

1.  For small messages we deserialize on the event loop for performance
    reasons.  The user can control this with the
    distributed.comm.offload configuration value.

    I recommend the value of 1, meaning a single byte

2.  The scheduler-client comms intentionally do not offload today
    (grep for the `allow_offload=False`)

### DIFF
--- a/distributed/comm/utils.py
+++ b/distributed/comm/utils.py
@@ -13,14 +13,6 @@ from ..utils import get_ip, get_ipv6, nbytes, offload
 logger = logging.getLogger(__name__)
 
 
-# Offload (de)serializing large frames to improve event loop responsiveness.
-# We use at most 4 threads to allow for parallel processing of large messages.
-
-FRAME_OFFLOAD_THRESHOLD = dask.config.get("distributed.comm.offload")
-if isinstance(FRAME_OFFLOAD_THRESHOLD, str):
-    FRAME_OFFLOAD_THRESHOLD = parse_bytes(FRAME_OFFLOAD_THRESHOLD)
-
-
 async def to_frames(
     msg, serializers=None, on_error="message", context=None, allow_offload=True
 ):
@@ -40,7 +32,11 @@ async def to_frames(
             logger.exception(e)
             raise
 
-    if FRAME_OFFLOAD_THRESHOLD and allow_offload:
+    # Offload serializing large frames to improve event loop responsiveness.
+    frame_offload_threshold = dask.config.get("distributed.comm.offload")
+    if isinstance(frame_offload_threshold, str):
+        frame_offload_threshold = parse_bytes(frame_offload_threshold)
+    if frame_offload_threshold and allow_offload:
         try:
             msg_size = sizeof(msg)
         except RecursionError:
@@ -48,7 +44,7 @@ async def to_frames(
     else:
         msg_size = 0
 
-    if allow_offload and FRAME_OFFLOAD_THRESHOLD and msg_size > FRAME_OFFLOAD_THRESHOLD:
+    if allow_offload and frame_offload_threshold and msg_size > frame_offload_threshold:
         return await offload(_to_frames)
     else:
         return _to_frames()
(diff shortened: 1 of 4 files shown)


jrbourbeau:
Nice, this is really interesting. We also don't offload deserialization of tasks on the worker

https://github.com/dask/distributed/blob/3407aa31be08ce100bd96e42acb99e1d7daf996e/distributed/worker.py#L3280-L3294

so for this to work, we'll need to add some additional offloading logic there

distributed/worker.py
```diff
@@ -2449,7 +2452,16 @@ def _maybe_deserialize_task(self, ts):
             return ts.runspec
         try:
             start = time()
-            function, args, kwargs = _deserialize(*ts.runspec)
+            # Offload deserializing large tasks
+            offload_threshold = get_offload_threshold()
+            if sizeof(ts.runspec) > offload_threshold:
```
jrbourbeau, line 2457: I want to think about if there's a better way to do this. This may lead to a slowdown if we start submitting task deserialization to a busy offloading thread pool.

distributed/comm/utils.py
```diff
@@ -40,15 +39,17 @@ def _to_frames():
             logger.exception(e)
             raise
 
-    if FRAME_OFFLOAD_THRESHOLD and allow_offload:
+    # Offload serializing large frames to improve event loop responsiveness.
+    offload_threshold = get_offload_threshold()
+    if offload_threshold and allow_offload:
```
mrocklin, line 44: Hrm, this may be expensive to do on every message. Thoughts?

jrbourbeau: It looks like this adds ~3ms:

```python
In [1]: import dask

In [2]: from distributed.comm.utils import get_offload_threshold

In [3]: %timeit get_offload_threshold()
2.96 µs ± 44.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

This change was so we could set `distributed.comm.offload` at runtime (i.e. `with dask.config.set(distributed__comm__offload=1):`) instead of just at startup time. Though I agree since is run for every message, the increased overhead may not be worth the extra flexibility.

jrbourbeau: I reverted the addition of `get_offload_threshold` and we now just pull in `distributed.comm.offload` at startup time

distributed/tests/test_worker.py
```diff
@@ -1755,3 +1756,30 @@ async def test_taskstate_metadata(cleanup):
...
+async def test_executor_offload(cleanup, monkeypatch):
+    class SameThreadClass:
+        def __getstate__(self):
+            return ()
+
+        def __setstate__(self, state):
+            self._thread_ident = threading.get_ident()
+            return self
+
+    monkeypatch.setattr("distributed.worker.OFFLOAD_THRESHOLD", 1)
```
mrocklin, line 1771: I'm not used to `monkeypatch`.  Does this clear out at the end of this test automatically, or do we need to do a try-finally here?

jrbourbeau: It clears at the end of this test

distributed/worker.py
```diff
@@ -2444,12 +2450,16 @@ def meets_resource_constraints(self, key):
...
-    def _maybe_deserialize_task(self, ts):
+    async def _maybe_deserialize_task(self, ts):
         if not isinstance(ts.runspec, SerializedTask):
             return ts.runspec
         try:
             start = time()
-            function, args, kwargs = _deserialize(*ts.runspec)
+            # Offload deserializing large tasks
+            if sizeof(ts.runspec) > OFFLOAD_THRESHOLD:
+                function, args, kwargs = await offload(_deserialize, *ts.runspec)
```
jrbourbeau, line 2460: +1 this is definitely nicer

mrocklin:
OK.  This seems ok to me.  Merging in.

jakirkham:
cc @Carreau (who IIRC had a similar use case in the past)

Issue #4251, opened 2020-11-18
Flaky test_profile_server


jrbourbeau:
We ran into a failure of `test_profile_server` in CI over in https://github.com/dask/distributed/pull/4248 that was unrelated to the changes in the PR. Full traceback below. 

<details>
<summary>Full traceback:</summary>

```
2020-11-17T21:43:17.4637490Z ================================== FAILURES ===================================
2020-11-17T21:43:17.4638364Z _____________________________ test_profile_server _____________________________
2020-11-17T21:43:17.4638784Z 
2020-11-17T21:43:17.4639380Z self = <closed TCP>, deserializers = None
2020-11-17T21:43:17.4639838Z 
2020-11-17T21:43:17.4640417Z     async def read(self, deserializers=None):
2020-11-17T21:43:17.4641238Z         stream = self.stream
2020-11-17T21:43:17.4641733Z         if stream is None:
2020-11-17T21:43:17.4642434Z             raise CommClosedError
2020-11-17T21:43:17.4642878Z     
2020-11-17T21:43:17.4643274Z         try:
2020-11-17T21:43:17.4643935Z >           n_frames = await stream.read_bytes(8)
2020-11-17T21:43:17.4644789Z E           tornado.iostream.StreamClosedError: Stream is closed
2020-11-17T21:43:17.4645373Z 
2020-11-17T21:43:17.4645937Z distributed\comm\tcp.py:187: StreamClosedError
2020-11-17T21:43:17.4646472Z 
2020-11-17T21:43:17.4647051Z The above exception was the direct cause of the following exception:
2020-11-17T21:43:17.4647545Z 
2020-11-17T21:43:17.4647915Z     def test_func():
2020-11-17T21:43:17.4648375Z         result = None
2020-11-17T21:43:17.4648771Z         workers = []
2020-11-17T21:43:17.4649377Z         with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:
2020-11-17T21:43:17.4649894Z     
2020-11-17T21:43:17.4650314Z             async def coro():
2020-11-17T21:43:17.4650810Z                 with dask.config.set(config):
2020-11-17T21:43:17.4651328Z                     s = False
2020-11-17T21:43:17.4651735Z                     for i in range(5):
2020-11-17T21:43:17.4652177Z                         try:
2020-11-17T21:43:17.4652622Z                             s, ws = await start_cluster(
2020-11-17T21:43:17.4653132Z                                 nthreads,
[...]

PR #4308, opened 2020-12-03
Remove offload try/except for thread_name_prefix keyword

jrbourbeau:
The `thread_name_prefix=` keyword for `ThreadPoolExecutor` was added in Python 3.6. Since we no longer support Python < 3.6 we can safely remove this `try`/`except` block


mrocklin:
If CI is happy then I'm happy :)

jrbourbeau:
CI is just as happy as it usually is https://github.com/dask/distributed/issues/4251

Issue #4305, opened 2020-12-03
Clean Up/Closing issues


quasiben:
**What happened**:
I've been tracking what I originally thought was a [streamz clean up issue](https://github.com/python-streamz/streamz/issues/385) but have since built a reproducer with just dask/distributed.  The example runs correctly but has a noisy clean up

```
bzaitlen@exp02:/tmp/bzaitlen$ python test-cleanup.py
dask.array<rechunk-merge, shape=(1500, 1500), dtype=float32, chunksize=(500, 500), chunktype=numpy.ndarray>
finished
Exception ignored in: <function Future.__del__ at 0x7f28c8f6ac10>
Traceback (most recent call last):
  File "/datasets/bzaitlen/miniconda3/envs/20201201/lib/python3.8/site-packages/distributed/client.py", line 385, in __del__
  File "/datasets/bzaitlen/miniconda3/envs/20201201/lib/python3.8/site-packages/distributed/client.py", line 360, in release
  File "/datasets/bzaitlen/miniconda3/envs/20201201/lib/python3.8/site-packages/tornado/platform/asyncio.py", line 222, in add_callback
AttributeError: 'NoneType' object has no attribute 'get_ident'
```

**What you expected to happen**:

Run without Exceptions on exit

**Minimal Complete Verifiable Example**:

```python
import dask.array as da
import numpy
from distributed import Client, wait
if __name__ == "__main__":
    client = Client(n_workers=2)
    a = numpy.arange(1500 * 1500).reshape((1500, 1500)).astype(numpy.float32)
    a = da.from_array(a).persist()
    a = a.rechunk((500, 500))
    print(a)
    b = a.persist()
    wait((b.dot(b.T)).persist())
    print('finished')
```

**Anything else we need to know?**:

I am not able to reproduce on my laptop but I can reproduce easily on a DGX machine

**Environment**:

- Dask version: latest
- Python version: 3.8
- Operating System: Linux/OSX
- Install method (conda, pip, source): source

PR #4309, opened 2020-12-03
When releasing futures except AttributeError for early closing clients

quasiben:
PR resolves https://github.com/dask/distributed/issues/4305

I think this is a bit better than checking for `AttributeError`.  If you have time, @martindurant I'd appreciate your thoughts here

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -382,11 +382,11 @@ class Future(WrappedKey):
     def __del__(self):
         try:
             self.release()
-        except AttributeError as e:
+        except AttributeError:
             # Ocassionally we see this error when shutting down the client
             # https://github.com/dask/distributed/issues/4305
             if not shutting_down():
-                raise e
+                raise
         except RuntimeError:  # closed event loop
             pass
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -382,10 +382,11 @@ class Future(WrappedKey):
     def __del__(self):
         try:
             self.release()
-        except AttributeError:
+        except AttributeError as e:
             # Ocassionally we see this error when shutting down the client
             # https://github.com/dask/distributed/issues/4305
-            pass
+            if not shutting_down():
+                raise e
         except RuntimeError:  # closed event loop
             pass
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -353,11 +353,7 @@ class Future(WrappedKey):
     def release(self, _in_destructor=False):
         # NOTE: this method can be called from different threads
         # (see e.g. Client.get() or Future.__del__())
-        if (
-            not self._cleared
-            and self.client.generation == self._generation
-            and self.client.scheduler is not None
-        ):
+        if not self._cleared and self.client.generation == self._generation:
             self._cleared = True
             try:
                 self.client.loop.add_callback(self.client._dec_ref, stringify(self.key))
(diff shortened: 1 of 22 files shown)


martindurant:
This seems eminently sensible to me.
Can it be triggered in a test, assign a Future to a client, then close the scheduler before calling `release()` explicitly? I've has a 2-minute attempt at this and didn't find a way to make it fail. If you can't do it in 5min, then we should merge this.

quasiben:
I haven't been able to trigger on my laptop.  Just on the DGX :/ I'll take a look some more tomorrow

distributed/client.py
```diff
@@ -353,7 +353,11 @@ def type(self):
...
         # (see e.g. Client.get() or Future.__del__())
-        if not self._cleared and self.client.generation == self._generation:
+        if (
+            not self._cleared
+            and self.client.generation == self._generation
+            and self.client.scheduler is not None
+        ):
             self._cleared = True
             try:
                 self.client.loop.add_callback(self.client._dec_ref, stringify(self.key))
```
mrocklin, line 363: The _dec_ref method does other things, like reducing the current refcount on the client.  If we're closing then sure, I'm happy for us not to this, but if we're in a state where we might connect again then this seems not quite right.

I'm curious, how does this situation come about?

If we're woried about the scheduler connection being down, then maybe there is a more comprehensive / central place to handle this, like in `_send_to_scheduler`.

martindurant: I believe the alternative is to catch multiple exceptions in `Future.release()`. 
The ignored exception is seen during interpreter shutdown, when it seems the client has closed its comm first. I don't think its been seen in ordinary cleanup of futures.

quasiben: I originally saw this issue in streamz test suite but only when running on DGXs.  I can reproduce on my laptop.  With @martindurant help, I was able to build a [reproducer](https://github.com/dask/distributed/issues/4305) which only relied on dask/distributed but only occurs when exiting the program or closing all the connections

martindurant:
@quasiben , do you want to simplify by adding the exception to the `except` clause in `Future.__del__`, instead of this solution, with a comment that we think this only happens during interpreter shutdown.

quasiben:
> @quasiben , do you want to simplify by adding the exception to the except clause in Future.__del__, instead of this solution, with a comment that we think this only happens during interpreter shutdown.

Sure, I'll do that

quasiben:
This keeps failing in various Travis tests.  I am restarting CI again

martindurant:
when do we squash Travis?! :)

quasiben:
Update:

I have restarted this job a number of times -- it keeps failing on:

> distributed/tests/test_worker.py::test_wait_for_outgoing

Locally, this passes.  Restarting CI again

quasiben:
Grr, this time it failed on:

> test_broken_worker_during_computation

And another one segfaulted.  Restarting again

mrocklin:
That one is known to fail on Travis

mrocklin:
I think that there is also a check somewhere of if we're shutting down the Python process.  I think that you might want `distributed.utils.shutting_down`

distributed/client.py
```diff
@@ -382,10 +382,11 @@ def __setstate__(self, state):
     def __del__(self):
         try:
             self.release()
-        except AttributeError:
+        except AttributeError as e:
             # Ocassionally we see this error when shutting down the client
             # https://github.com/dask/distributed/issues/4305
-            pass
+            if not shutting_down():
+                raise e
```
mrocklin: I think you want to drop the `as e` here and just call `raise`

If done within an except block this raises the original exception with a clean traceback I think.

quasiben: thanks @mrocklin -- fixed in #0455fa8

PR #4311, opened 2020-12-03
Coerce new `TaskState.nbytes` value to `int`

jakirkham:
Make sure that whatever value we assign to `nbytes` is an `int`. This is important as `Schedule.bandwidth` may be a `float`. So the value here could be a `float`. Though we only measure `nbytes` in whole values.

### DIFF
--- a/distributed/tests/test_steal.py
+++ b/distributed/tests/test_steal.py
@@ -493,7 +493,7 @@ async def assert_balanced(inp, expected, c, s, *workers):
                 ts = s.tasks[dat.key]
                 # Ensure scheduler state stays consistent
                 old_nbytes = ts.nbytes
-                ts.nbytes = s.bandwidth * t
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @jakirkham! This LGTM. Just wanted to point out that `nbytes` starts out as `None`

https://github.com/dask/distributed/blob/1c86be3208d73df3074a153fbad59324a804831e/distributed/scheduler.py#L721

Will that impact any Cython-related type checking?

jakirkham:
It does, but we are already fixing that in PR ( https://github.com/dask/distributed/pull/4302 ).

jrbourbeau (approved):
+1, thanks for the explanation @jakirkham

jakirkham:
Ofc. Thanks for the review! 😄

PR #4312, opened 2020-12-03
Use `str` for `Worker` `address` in tests

jakirkham:
Broken out from PR ( https://github.com/dask/distributed/pull/4296 )
Following up on this comment ( https://github.com/dask/distributed/pull/4294#pullrequestreview-544353865 )

This test uses an `int` address though we would expect it to be a `str`. So here we change it to a `str` to fix that issue.


jakirkham:
Hmm...think I misunderstood what was going on here. Looks like this is actually an alias following the `name=123` set earlier.

PR #4310, opened 2020-12-03
Fix `pip` install issue on CI

jakirkham:
Fixes an issue we are [seeing on CI]( https://github.com/dask/distributed/pull/4294/checks?check_run_id=1495023367 ). This came up for us in RAPIDS and this solved the issue there as well. Basically `pip`'s solver seems to be trying to get involved here even though we don't need it to. Specifying the repo branches seems to bypass the issue.

This also came up in dask-cloudprovider recently ( https://github.com/dask/dask-cloudprovider/pull/203 ).

PR #4313, opened 2020-12-04
Remove empty test_highgraph.py file

jrbourbeau:
This remove an empty file that was committed (xref https://github.com/dask/distributed/pull/4279#issuecomment-738500472)


jakirkham (approved):

jakirkham:
Thanks James! 😄

I restarted CI to see if that clears things up, but we may need to rebased this on latest `master` to get the `pip` fix in commit ( https://github.com/dask/distributed/commit/1c86be3208d73df3074a153fbad59324a804831e ). For context some CIs (like Travis CI, Azure, etc.), create a merge commit between the PR and the base branch the PR is targeting, `master` in this case. However some CIs (like CircleCI) don't do this. As a result, sometimes users need to explicitly perform this merge with the base branch itself.

I'm not sure where GitHub Actions fits between these two. If it is in the former group, then this shouldn't be needed. Though if it is in the later, we may need to merge or rebase to ensure we have the fix. Based on a couple of PRs yesterday, I think manual action may be needed, but could be wrong.

jrbourbeau:
Thanks for the additional context and CI bump @jakirkham!

I had merged the `master` branch locally (which contained changes from https://github.com/dask/distributed/pull/4310) before creating a branch for this PR. Looking at my local git history for this branch I get:

```
(distributed) ➜  distributed git:(remove-test-file) ✗ git log --pretty=format:"%h - %an : %s" | head -n 5
65558264 - James Bourbeau : Remove empty test_highgraph.py file
7ea9696c - jakirkham : Coerce new `TaskState.nbytes` value to `int` (#4311)
3407aa31 - James Bourbeau : Remove offload try/except for thread_name_prefix keyword (#4308)
1c86be32 - jakirkham : Fix `pip` install issue on CI (#4310)
dcdb4651 - Simon Perkins : Transmit Layer annotations to scheduler (#4279)
```

So I would have expected the fix from 1c86be3208d73df3074a153fbad59324a804831e to be included regardless of if GitHub actions creates a merge commit with the base branch or not

jakirkham:
Interesting, well it doesn't look like we are seeing the issue on GitHub Actions any more. Not sure why that one job still had issues.

Happy to merge now. I'm guessing we don't wait for Travis CI any more due to the slow queue and credit limit issues. Is that right?

jrbourbeau:
Yeah, for this PR in particular I think it's safe to not wait for Travis : )

PR #4315, opened 2020-12-04
Ensure that retire_workers returns a dict

mrocklin:
A long while ago we changed retire_workers to return a dict, but didn't
change all of the cases, particularly those for empty results, which
still returned a list.  Now we reliably return a dict in all cases.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3616,7 +3616,7 @@ class Scheduler(ServerNode):
         names=None,
         lock=True,
         **kwargs,
-    ):
+    ) -> dict:
         """Gracefully retire workers from cluster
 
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):
Good catch, thanks @mrocklin

mrocklin:
Thanks go to @selshowk

jakirkham:
Thanks all! 😄

Just want to add this increased consistency also helps with Cythonization as well.

jrbourbeau:
🎉

jakirkham:
Found one more case where this can happen. Submitted PR ( https://github.com/dask/distributed/pull/4323 ) to fix that case as well.

PR #4319, opened 2020-12-05
Swap order of `TaskGroup` and `TaskPrefix`

jakirkham:
Required for PR ( https://github.com/dask/distributed/pull/4302 )

Sorry for not realizing this earlier. Didn't notice that `TaskGroup` should have a `TaskPrefix` attribute (see code below). So this just flips the code of these 2 classes so we can annotate that attribute in `TaskGroup`.

https://github.com/dask/distributed/blob/b23f7478e08e7022ec6102082eb8b7144731526d/distributed/scheduler.py#L2354

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -539,76 +539,6 @@ class WorkerState:
         return self._nthreads
 
 
-class TaskGroup:
-    """Collection tracking all tasks within a group
-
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):

jakirkham:
That was fast! Thanks James 😄 Again sorry for not catching this earlier 😞

PR #4324, opened 2020-12-07
Use `operator.attrgetter` on `WorkerState.address`

jakirkham:
Make use of `operator.attrgetter` to get `WorkerState`'s `address` attribute instead of a `lambda`. This winds up being a bit faster. Also `operator.attrgetter` is honed for this purpose. Not too mention it is already used frequently for this purpose in this code.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3696,7 +3696,7 @@ class Scheduler(ServerNode):
                 return []
 
             if key is None:
-                key = lambda ws: ws.address
+                key = operator.attrgetter("address")
             if isinstance(key, bytes) and dask.config.get(
                 "distributed.scheduler.pickle"
             ):


jakirkham:
Here's a simple benchmark to illustrate the difference:

```python
In [1]: import operator

In [2]: class Obj:
   ...:     def __init__(self, k):
   ...:         self.k = k
   ...: 

In [3]: L = lambda o: o.k

In [4]: ag = operator.attrgetter("k")

In [5]: o = Obj("key")

In [6]: %timeit L(o)
81.3 ns ± 0.396 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: %timeit ag(o)
61.5 ns ± 0.315 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

quasiben:
Do you see us implementing `operator. attrgetter` in many places throughout the scheduler code ?   Mostly I want to make sure we (dask maintainers) are ok with tradeoff of readability and performance for this particular change.  This isn't so bad in my opinion and it does come at a modest performance benefit

jakirkham:
Yeah that is already the case. This use of a `lambda` was an exception. So if anything, this makes the code more consistent.

quasiben:
> Yeah that is already the case.

Thanks

PR #4326, opened 2020-12-08
[WIP] Fix-up Cython build

jakirkham:
As the modules we Cythonize are `.py` file, both the compiled extension and the `.py` file get installed. On macOS and Linux, this doesn't appear to present an issue as the `.so` will be loaded before the `.py` file. So the `.py` files are effectively ignored on these platforms. However it seems Windows loads the `.py` file before the `.pyd` (Windows compiled extension).

To fix this, we make some tweaks to the build process. Namely we rename the `.py` file to a `.pyx` file, which exclude from packaging (just as we exclude the `.c` file Cython generates). Thus we avoid having two files with the same name and different file extensions, which may be loaded in arbitrary order. We make sure to return the source directory to the original state by reverting the renaming when we are done.

For this to work effectively on CI, we start installing into the Conda environment instead of doing an editable install. This way we know only the compiled extension was installed.

### DIFF
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -43,9 +43,9 @@ jobs:
         shell: bash -l {0}
         run: |
           if [[ "${{ matrix.python-version }}" = "3.7" ]]; then
-            python -m pip install -q --no-deps --install-option="--with-cython" -e .
+            python -m pip install -q --no-deps --install-option="--with-cython" .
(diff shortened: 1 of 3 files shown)


mrocklin:
Happy to defer entirely to you on this one :)

jakirkham:
Addressing with PR ( https://github.com/dask/distributed/pull/4764 )

PR #4329, opened 2020-12-08
Optimize `valid_workers`

jakirkham:
Annotates `valid_workers` for Cython optimization. In particular ensure it always returns a `set` or `None` (in case all workers are viable instead of `True` as before). Also use unique variables when they have already been assigned or if new results may have different types than before. This makes it easier to annotate these variables and for Cython to then optimize usage thereof.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5544,7 +5544,7 @@ class Scheduler(ServerNode):
                 s |= ss
 
         if ts._resource_restrictions:
-            w = {
+            dw = {
                 resource: {
                     w
                     for w, supplied in self.resources[resource].items()
@@ -5553,7 +5553,7 @@ class Scheduler(ServerNode):
                 for resource, required in ts._resource_restrictions.items()
             }
 
-            ww = set.intersection(*w.values())
+            ww = set.intersection(*dw.values())
             if s is True:
                 s = ww
             else:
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -4655,9 +4655,9 @@ def decide_worker(self, ts: TaskState):
         """
         Decide on a worker for task *ts*.  Return a WorkerState.
         """
-        valid_workers = self.valid_workers(ts)
+        valid_workers: set = self.valid_workers(ts)
 
-        if not valid_workers and not ts._loose_restrictions and self.workers:
+        if valid_workers == set() and not ts._loose_restrictions and self.workers:
```
mrocklin, line 4660: I'm surprised that it makes sense to create an empty set each time.

jakirkham: Yeah I think you are right; we can do better. Will push a fix.

```python
In [4]: %timeit s == set()
91.3 ns ± 0.747 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: %timeit s is not None and not s
37.3 ns ± 0.373 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

mrocklin:
+1

mrocklin:
+1

jakirkham:
Thanks Matt! 😄

PR #4332, opened 2020-12-09
Optimize `decide_worker`

jakirkham:
Annotate variables in `decide_worker` (both method and function). Deduplicate some code within the `decide_worker` method. Also tune the `decide_worker` function fast path.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4653,7 +4653,7 @@ class Scheduler(ServerNode):
                 pdb.set_trace()
             raise
 
-    def decide_worker(self, ts: TaskState):
+    def decide_worker(self, ts: TaskState) -> WorkerState:
         """
         Decide on a worker for task *ts*.  Return a WorkerState.
         """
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -4667,36 +4668,31 @@ def decide_worker(self, ts: TaskState):
...
-                worker,
+            assert ws is None or isinstance(ws, WorkerState), (
+                type(ws),
+                ws,
             )
-            assert worker.address in self.workers
+            assert ws.address in self.workers
 
-        return worker
+        return ws
```
mrocklin, line 4695: Thank you for standardizing pronouns along the way

mrocklin:
It looks like we're going the route of doing easy optimizations before splitting things out to a cythonized class and a non-cythonized class.  Is this correct?

distributed/scheduler.py
jakirkham: It's certainly made it a lot easier to understand where things can be annotated 🙂

jakirkham:
Yes and no. I started a refactoring branch as well. Though I think the particular transition these changes are associated with, `transition_waiting_processing`, is probably the hardest to refactor in a meaningful way atm. It also happens to be the 2nd slowest transition. So wanted to make a concerted effort to unblock performance for this transition before going ahead with the refactor.

PR #4335, opened 2020-12-09
Handle `WorkerState` `memory_limit` of `None`

jakirkham:
If a `memory_limit` of `None` is passed to `add_worker`, make sure to replace it with `0` before constructing `WorkerState` as `memory_limit` in `WorkerState` is typed as `Py_ssize_t` so `None` is not permissible.


jakirkham:
As `memory_limit` can be set to `None` in the `Worker` constructor...

https://github.com/dask/distributed/blob/d54388c92f5cb6643180749f7d30f580609379a2/distributed/worker.py#L560

...and `parse_memory_limit` will preserve the value of `None`...

https://github.com/dask/distributed/blob/d54388c92f5cb6643180749f7d30f580609379a2/distributed/worker.py#L3201-L3203

...this can wind up being passed to the `Scheduler` during worker registration.

https://github.com/dask/distributed/blob/d54388c92f5cb6643180749f7d30f580609379a2/distributed/worker.py#L878

Hence this replaces `None` with an integer value instead that `WorkerState` can handle.

https://github.com/dask/distributed/blob/d54388c92f5cb6643180749f7d30f580609379a2/distributed/scheduler.py#L294

This matches the default `WorkerState` would have anyways.

https://github.com/dask/distributed/blob/d54388c92f5cb6643180749f7d30f580609379a2/distributed/scheduler.py#L342

PR #4336, opened 2020-12-09
Relax `TaskState` typing of `key` to `object`

jakirkham:
While this usually is a `str`, it seems sometimes it can be a `bytes` object. So relax the typing of `key` to object. That way it won't cause issues when given a `bytes` object.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1026,7 +1026,7 @@ class TaskState:
         Task annotations
     """
 
-    _key: str
+    _key: object
     _hash: Py_hash_t
     _prefix: TaskPrefix
     _run_spec: object
@@ -1107,7 +1107,7 @@ class TaskState:
         "_annotations",
     )
 
-    def __init__(self, key: str, run_spec: object):
+    def __init__(self, key: object, run_spec: object):
         self._key = key
         self._hash = hash(key)
         self._run_spec = run_spec


jakirkham:
This came up as a result of this test.

https://github.com/dask/distributed/blob/d54388c92f5cb6643180749f7d30f580609379a2/distributed/tests/test_client.py#L5059-L5067

mrocklin:
I'm inclined to either coerce the key on the client side.  I think that we do this with other objects as well.

```python

In [2]: client = Client()

In [3]: future = client.submit(lambda x: x + 1, 10, key=("x", 1))

In [4]: future
Out[4]: <Future: finished, type: builtins.int, key: ('x', 1)>

In [5]: future.key
Out[5]: "('x', 1)"
```

I understand that this may break existing workloads.  I'm ok with that.   `TaskState.key` is heavily used on the scheduler.  My guess is that we want to keep this more strictly typed.

jakirkham:
Yeah I agree with that as well. Was mainly trying to surface the issue discovered. Will look into coercion on the `Client`.

jakirkham:
One option might be to ensure `stringify` coerces `bytes` to `str` ( https://github.com/dask/dask/pull/6944 ). That function seems to already be used throughout on the `Client`. Not sure to what extent we rely on its current behavior though. So open to other options.

PR #4341, opened 2020-12-09
Annotation of some comm related methods in the Scheduler

jakirkham:
This does some annotation around comm related methods in the scheduler: `report_on_key`, `report`, and `worker_send`. Should help speed through the intermediate code needed to send out a message.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4200,7 +4200,7 @@ class Scheduler(ServerNode):
             if client:
                 self.client_desires_keys(keys=list(who_has), client=client)
 
-    def report_on_key(self, key=None, ts: TaskState = None, client=None):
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -4201,7 +4201,7 @@ def update_data(
                 self.client_desires_keys(keys=list(who_has), client=client)
 
     def report_on_key(self, key: str = None, ts: TaskState = None, client: str = None):
-        assert (key is None) + (ts is None) == 1, (key, ts)
+        assert (key is None) != (ts is None), (key, ts)
```
jakirkham, line 4204: Found another case in the scheduler like this, addressing in PR ( https://github.com/dask/distributed/pull/4342 ).

PR #4342, opened 2020-12-09
Optimize `assert` in `validate_waiting`

jakirkham:
Makes a few tweaks to this `assert` in `validate_waiting`. First `not not` ends up being faster than `bool(...)` in Python and Cython. The latter because it can skip constructing a Python `bool` object and continue operating in straight C. Second `!=` behaves like `xor` between `bool`s, which Cython can also do directly in C.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2950,7 +2950,7 @@ class Scheduler(ServerNode):
         assert ts not in self.unrunnable
         for dts in ts._dependencies:
             # We are waiting on a dependency iff it's not stored
-            assert bool(dts._who_has) != (dts in ts._waiting_on)
+            assert (not not dts._who_has) != (dts in ts._waiting_on)
             assert ts in dts._waiters  # XXX even if dts._who_has?
 
     def validate_processing(self, key):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2950,7 +2950,7 @@ class Scheduler(ServerNode):
         assert ts not in self.unrunnable
         for dts in ts._dependencies:
             # We are waiting on a dependency iff it's not stored
-            assert bool(dts._who_has) + (dts in ts._waiting_on) == 1
+            assert bool(dts._who_has) != (dts in ts._waiting_on)
             assert ts in dts._waiters  # XXX even if dts._who_has?
 
     def validate_processing(self, key):


jakirkham:
Couple of quick benchmarks below.

First `bool` vs. `not not`:

```python
In [1]: s0 = set()

In [2]: s1 = {1, 2, 3}

In [3]: %timeit bool(s0)
66.1 ns ± 0.123 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit bool(s1)
70.5 ns ± 0.203 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: %timeit (not not s0)
22.6 ns ± 0.25 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [6]: %timeit (not not s1)
28.3 ns ± 1.35 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Second adding `bool`s vs. comparing them:

```python
In [1]: b0 = False

In [2]: b1 = True

In [3]: %timeit b0 + b1
34.7 ns ± 0.174 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit b0 != b1
33.2 ns ± 0.137 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

This gets better in Cython of course where these are turned into C, but that should give an idea of how pure Python is impacted.

jakirkham:
Oops 😮 Sorry messed something up locally. Reopened

mrocklin:
Not a big issue, but I have a slight preference to accumulate these smaller changes into a larger PR if you're game.

jakirkham:
Yeah this seemed unrelated to other changes so erred on the side of keeping it separate, but ok with including in a larger PR in the future. Thanks!

mrocklin:
Yeah, totally fine. I think I saw a few PRs come in, saw this one first and thought "oh dear, this is going to be a bunch of micro-changes, one change per PR" but then I saw that that wasn't the case and regretted my comment 

Thanks again for all of the work here.

jakirkham:
Not at all. It's good to be reminded of people's preferences when it comes to reviewing :)

More often than not the feedback I've heard over the years has been to keep PRs focused hence my bias in that direction. That said, this approach doesn't suit everyone. So completely understand lumping more things together

Thanks for all the reviews :)

PR #4343, opened 2020-12-10
[WIP] Refactor communication out of transition functions

jakirkham:
Currently transition functions alternate between working on the task graph a bit and communicate out to workers and clients. As it is, this is pretty interwoven. Sometimes going back and forth between these two modes of processing multiple times in the same function. This makes it a little tricky to optimize the transition functions further as often the communication is handled in high-level Python and doesn't improve much through Cythonization. Further this makes a bit tricky to pull apart the transitions functions and task graph state from the Scheduler as it is a bit too reliant on these communication steps.

To attempt to remedy this situation, this PR tries to move all of the communication out of the transition functions. It does this by trying to return all messages to be communicated and handling those in `transition` after each transition function completes. While transition functions and communication still effectively interleave at a higher level, this should make it easier to separate out the transition functions along with relevant state for further optimization.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4568,7 +4568,7 @@ class Scheduler(ServerNode):
     # State Transitions #
     #####################
 
-    def _remove_from_processing(self, ts: TaskState, send_worker_msg=None):
+    def _remove_from_processing(self, ts: TaskState) -> str:
         """
         Remove *ts* from the set of processing tasks.
         """
@@ -4586,8 +4586,9 @@ class Scheduler(ServerNode):
                 ws._occupancy -= duration
             self.check_idle_saturated(ws)
             self.release_resources(ts, ws)
-            if send_worker_msg:
-                self.worker_send(w, send_worker_msg)
+            return w
+        else:
+            return None
 
     def _add_to_memory(
         self,
@@ -5179,9 +5180,9 @@ class Scheduler(ServerNode):
                 assert not ts._waiting_on
                 assert self.tasks[key].state == "processing"
 
-            self._remove_from_processing(
-                ts, send_worker_msg={"op": "release-task", "key": key}
-            )
+            w: str = self._remove_from_processing(ts)
+            if w:
+                self.worker_send(w, {"op": "release-task", "key": key})
 
             ts.state = "released"
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4671,6 +4671,7 @@ class Scheduler(ServerNode):
             workers: dict = cast(dict, self.workers)
             ts: TaskState = tasks[key]
             dts: TaskState
+            worker_msgs: dict = {}
 
             if self.validate:
                 assert ts._run_spec
@@ -4680,7 +4681,7 @@ class Scheduler(ServerNode):
                 assert not any([dts._state == "forgotten" for dts in ts._dependencies])
 
             if ts._has_lost_dependencies:
-                return {key: "forgotten"}
+                return {key: "forgotten"}, worker_msgs
 
             ts.state = "waiting"
 
@@ -4691,7 +4692,7 @@ class Scheduler(ServerNode):
                 if dts._exception_blame:
                     ts._exception_blame = dts._exception_blame
                     recommendations[key] = "erred"
-                    return recommendations
+                    return recommendations, worker_msgs
 
             for dts in ts._dependencies:
                 dep = dts._key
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -4566,10 +4571,14 @@ def _add_to_memory(
         if ts in cs._wants_what:
             self.client_releases_keys(client="fire-and-forget", keys=[ts._key])
```
jakirkham, line 4572: This seems like something we need to move out of here as well, but am less sure how we do that.

jakirkham: Have tried to address this by collecting the `recommendations` from `client_releases_keys` and passing those along back to the caller to in turn return back to `transition` to handle.

jakirkham:
This still has some warts. That said, it would be good to get a sense of whether this change seems reasonable and what issues we might expect going with this approach (which may in turn inform how the warts are addressed 😉).

distributed/scheduler.py
```diff
@@ -4556,20 +4567,25 @@ def _add_to_memory(
...
+            )
+
+        return msg
 
     def transition_released_waiting(self, key):
         try:
             ts: TaskState = self.tasks[key]
             dts: TaskState
+            worker_msgs: dict = {}
+            report_msg: dict = {}
```
mrocklin, line 4588: Should these be `None` for the seemingly common case of no messages to report?

mrocklin:
At first glance this seems sensible to me.  I would prefer that we not merge it until we get further along down this path.  It seems like a decent path to go down though.  (I'm thinking that we merge once we have a Python class that manages the scheduler state machine).

distributed/scheduler.py
jakirkham: Well as we have `for`-loops back in `transition` that try to iterate over these, we would need to add logic to check for `None` somehow. Atm this seems pretty straightforward even if there are cases where one of these doesn't contain something.

jakirkham:
Thanks Matt! That makes sense. Wanted to share my thought process before going too far down this path.

Atm am seeing some hangs in the test suite. Not sure why that is yet. Am curious does it matter what state the Scheduler's task graph is in before it communicates with workers? Also should we be concerned about message ordering at all?

jakirkham:
> Atm am seeing some hangs in the test suite. Not sure why that is yet. Am curious does it matter what state the Scheduler's task graph is in before it communicates with workers? Also should we be concerned about message ordering at all?

Figured it out. Was overlooking another call to `_add_to_memory` so was not handling the `report` calls needed there. Fixing that fixed the issue.

distributed/scheduler.py
jakirkham: Should add that ultimately my hope is this becomes a message object, which we pass from transition functions and the communication side of the Scheduler. For now it is sort of an arbitrary list of things we have found need to be passed around. So I think now that we have identified what those things that we message are, this should be easier to handle.

jakirkham:
Here's the call graph from the Scheduler that I see with these changes. This is best compared with the recent benchmark results here ( https://github.com/quasiben/dask-scheduler-performance/issues/51 ).

From this we see that `transition_memory_released` also drops out (as it was only in the call graph at this point due to communication). The only relevant transitions are `transition_processing_memory` and `transition_waiting_processing`. The latter also takes significantly less time as part of its work was communication as well (though not entirely).

Additionally we now see communication that stems from `transition`, which is expected. It also takes basically the same amount of time. Again this is expected. We've only shifted where the communication happens. Not how much of it occurs.

prof_68726 pstat dot

jakirkham:
Have refactored things away from passing `TaskState` objects to things like `report_on_key` after they (and the Scheduler graph as a whole) have been modified. Instead we now construct messages for each Client in the same way we do for each worker as part of the transition. As a result we are now returning `dict`s of addresses with messages for them. One for the clients and one for the workers.

jakirkham:
Is there anything special we should be doing with the "fire-and-forget" Client? Does it receive messages? IIUC this Client is really just on the Scheduler (maybe not even a full Client in the same sense as a user created one). Though want to make sure there's nothing I'm missing here.

distributed/scheduler.py
```diff
@@ -2970,7 +2970,7 @@ def client_desires_keys(self, keys=None, client=None):
             if ts._state in ("memory", "erred"):
                 self.report_on_key(ts=ts, client=client)
 
-    def client_releases_keys(self, keys=None, client=None):
+    def _client_releases_keys(self, keys=None, client=None, recommendations: dict = {}):
```
mrocklin, line 2973: Should we be concerned about a default mutable argument like this?  Or does it not matter because everyone always provides the recommendations= parameter?

jakirkham: Good point. Yeah we can make this `None`, which would be safer. Thanks!

```suggestion
    def _client_releases_keys(self, keys=None, client=None, recommendations: dict = None):
```

jakirkham:
Closing and continuing in PR ( https://github.com/dask/distributed/pull/4365 ).

PR #4344, opened 2020-12-10
Drop serialize extension

jakirkham:
This disables Cythonizing `serialize`. While we have annotated one function in that module, most of `serialize` is not Cythonized. Additionally there are some `ImportError`s in development mode that have been getting in the way. So suggest we just drop this for now and revisit if this becomes more important in the future.


quasiben:
Thanks @jakirkham

Issue #4306, opened 2020-12-03
Respect dask.annotate during compute/persist calls


mrocklin:
Now that https://github.com/dask/dask/pull/6889 and https://github.com/dask/distributed/pull/4279 are in (thanks @sjperkins !) we can use the `dask.annotate` context manager to mark layers built within a context

```python
with dask.annotate(foo="bar"):
    x = da.ones(10)
```

Some folks may also want to use this same context when calling compute/persist

```python
with dask.annotate(foo="baz"):
    x.compute()
```

Some thoughts on how we might do this:

- This would probably require that we check `dask.config.get("annotations")` inside `Client._graph_to_futures` and then send those annotations up to the scheduler.  
- We would probably also want to handle those annotations in the `Scheduler.update_graph_hlg` or `Scheduler.update_graph`.  
- We'll need to figure out the relative priority when the layer annotations and the compute-time annotations disagree.  My guess is that compute time annotations should take precedence.

@sjperkins does this interest you?

jrbourbeau:
> We'll need to figure out the relative priority when the layer annotations and the compute-time annotations disagree. My guess is that compute time annotations should take precedence.

+1 to having compute-time annotations taking precedence over construction-time annotations

sjperkins:
> Some thoughts on how we might do this:
> 
>     * This would probably require that we check `dask.config.get("annotations")` inside `Client._graph_to_futures`

Agreed

> and then send those annotations up to the scheduler.

>     * We would probably also want to handle those annotations in the `Scheduler.update_graph_hlg` or `Scheduler.update_graph`.

An easy way to do this would be to update existing HLG layers which have their own annotation transmission mechanism.

A generic "apply annotations to all keys transmitted to the scheduler" mechanism smells like it may have some gremlins/inefficiencies that would need careful consideration.

>     * We'll need to figure out the relative priority when the layer annotations and the compute-time annotations disagree.  My guess is that compute time annotations should take precedence.

Agreed. With the first approach above, this simply involves overriding the existing layer annotations.

> 
> @sjperkins does this interest you?

If the first approach is acceptable I think I could achieve a quick win.

mrocklin:
> An easy way to do this would be to update existing HLG layers which have their own annotation transmission mechanism.

I see two possible problems with this:

1.  Mutating the layers may unintended side effects
2.  Operations like submit and map may not be covered here

sjperkins:
[...]

PR #4347, opened 2020-12-10
Use annotations machinery for workers/priority/...

mrocklin:
Previously we had two systems to send per-task metadata like retries or
workers or priorities to the scheduler.

1.  Older system with explicit workers= keywords and expand_foo functions
2.  Newer system with annotations

The annotations system is nicer for a few reasons:

1.  It's more generic
2.  It's more consistent (there were some bugs in the expand foo
    functions, especially when dealing with collections)
3.  We ship values up on a per-layer basis rather than a per-key basis

This work-in-progress commit rips out the old system and uses the new system,
but it still missing a lot:

1.  It only implements this for the Client.compute method.
    We need to repeat this for persist, submit, and map
2.  It doesn't handle the allow_other_workers -> loose_restrictions
    conversion anywhere yet.  (this will need to be added to the
    scheduler)

cc @sjperkins if you have time I'd love your thoughts on this approach.  Also, if you agree with this approach and want to take over this PR that would also be welcome :)

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -11,10 +11,9 @@ import errno
 from functools import partial
 import html
 import inspect
-import itertools
 import json
 import logging
-from numbers import Number, Integral
+from numbers import Number
 import os
 import sys
 import uuid
@@ -2543,8 +2542,8 @@ class Client:
         self,
         dsk,
         keys,
-        restrictions=None,
-        loose_restrictions=None,
+        workers=None,
+        allow_other_workers=None,
         priority=None,
         user_priority=0,
         resources=None,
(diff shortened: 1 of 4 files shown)


sjperkins:
Firstly, I think placing the existing taxonomy into an annotation dictionary for scheduler transmission is the right way forward as it's a cleaner mechanism than having separate arguments to update-graph-hlg.

My initial reaction to global annotations (as described in https://github.com/dask/distributed/issues/4306#issue-756406852):

```python
with dask.annotate(foo="baz"):
    x.compute()
```
is that they shouldn't be supported as I've always thought of annotations as a fine-grained mechanism for describing a task. I think that "global " annotations should be discouraged. Perhaps my view is too narrow? If really desired the following pattern would also work:

```python
with dask.annotate(foo="baz"):
   x = construct_graph()

x.compute()
```

However, the existing API should probably be supported for backwards compatibility (as is started in this PR)

```python
client.compute(x, retries=3)
```

I may have time to look into this in more detail next week.

mrocklin:
This PR doesn't have an opinion on global vs local annotations in the compute call.  I suggested this in a previous issue, but so far all we're doing here is replacing the internal machinery.

mrocklin:
@ian-r-rose if you have time to carry on this PR that would be welcome.

ian-r-rose:
> @ian-r-rose if you have time to carry on this PR that would be welcome.

Sounds good

dhirschfeld:
> Perhaps my view is too narrow?

One use case I'd like to enable is to keep track of different jobs running on the cluster:
```python
with dask.annotate(job_name="weather-etl", job_id=uuid4().hex):
   x.compute()
```

I think that would require global annotations?

jrbourbeau:
Closing as this work was superseded by https://github.com/dask/distributed/pull/4406

PR #4348, opened 2020-12-10
Some common optimizations across transitions

jakirkham:
Makes sure `self.tasks` is treated as a `dict` by assigning to a locally typed variable. Similarly type `recommendations` as a `dict` throughout since that is commonly what it is. This allows Cython to use the Python C API for `dict`s with these objects.

Also uses `TaskState`'s `_state` attribute for faster access when just getting it and not setting it.

Annotate variables (both in arguments and locally) of the private methods used by various transitions.

Finally make a few other minor arrangements to favorably affect the C generated by Cython.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6221,7 +6221,7 @@ def validate_task_state(ts: TaskState):
         assert dts.state != "forgotten"
 
     assert (ts._processing_on is not None) == (ts.state == "processing")
-    assert bool(ts._who_has) == (ts.state == "memory"), (ts, ts._who_has)
+    assert (not not ts._who_has) == (ts.state == "memory"), (ts, ts._who_has)
 
     if ts.state == "processing":
         assert all([dts._who_has for dts in ts._dependencies]), (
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -2215,9 +2215,11 @@ async def add_worker(
                 except Exception as e:
                     logger.exception(e)
 
+            recommendations: dict
             if nbytes:
                 for key in nbytes:
-                    ts: TaskState = self.tasks.get(key)
+                    tasks: dict = self.tasks
+                    ts: TaskState = tasks.get(key)
```
mrocklin, line 2222: Can we not type `self.tasks` at the class level?  Or does this need the `@cclass` treatment to work?

jakirkham: Yeah it needs the `@cclass` treatment.

jakirkham:
Also switched to using `TaskState._state` instead of `TaskState.state` when just getting (not setting) the value. Think this was already a `@property` beforehand. So didn't make the change earlier. Though realized that it is just returning the attribute, which was eating up some time. So this fixes that issue as well.

The setting case for `TaskState.state` is a little more complicated and doesn't show up in the call graph. So left that alone for now.

Issue #3590, opened 2020-03-19
client.compute() not properly working with "workers" and dask.array


cbyrohl:
Hi there,

client.compute() properly works with a delayed object and "workers" specified, but not with a dask array.

For an existing (distributed) client

```python
from dask import delayed
from operator import add
x = delayed(add)(1, 2)
xx = client.compute([x],workers=['tcp://130.183.183.160:34655'])
```

is fine, but

```python
import dask.array as da
y = da.random.random(10)
y = client.compute([y],workers=['tcp://130.183.183.160:34655'])
```

fails with:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/freya/u/cbyrohl/repos/voroILTIS/production_LAE_SBr/haloprofiles.py in <module>
      1 from dask import delayed
      2 y = da.random.random(10)
----> 3 y = client.compute([y],workers=['tcp://130.183.183.160:34655'])

~/anaconda3/envs/lya/lib/python3.7/site-packages/distributed/client.py in compute(self, collections, sync, optimize_graph, workers, allow_other_workers, resources, retries, priority, fifo_timeout, actors, traverse, **kwargs)
   2767 
   2768         restrictions, loose_restrictions = self.get_restrictions(
-> 2769             collections, workers, allow_other_workers
   2770         )
   2771 

~/anaconda3/envs/lya/lib/python3.7/site-packages/distributed/client.py in get_restrictions(cls, collections, workers, allow_other_workers)
   3814         """ Get restrictions from inputs to compute/persist """
   3815         if isinstance(workers, (str, tuple, list)):
-> 3816             workers = {tuple(collections): workers}
   3817         if isinstance(workers, dict):
   3818             restrictions = {}

TypeError: unhashable type: 'Array'
```

Easy workaround is to create a delayed object from the array, but I think client.compute() is supposed to work with all dask collections?

Thanks,
Chris

mrocklin:
Yes indeed, this appears to be a bug.  Thank you for raising @cbyrohl

mrocklin:
Is this something that you're interesting in helping to resolve?

andrewfulton9:
Is this issue already resolved? I'm not able to reproduce it.

mrocklin:
Great.  Thanks for checking @andrewfulton9 .  Closing.

jrbourbeau:
Reopening as 

```python

import dask.array as da
from dask.distributed import Client

if __name__ == "__main__":
[...]

PR #4350, opened 2020-12-11
fixes client.compute not properly working with workers keywork argument and dask.arry

andrewfulton9:
addresses issue #3590

It was failing because `client.get_restrictions` was trying to use the `dask.array` collection as a key and it isn't hashable. The fix I have here, simply adds the restrictions from an except clause if it fails to build the dictionary. Another option would be to make `dask.array` hashable so it can be used as a dictionary key, though numpy arrays aren't hashable so I'm not sure if that makes sense


mrocklin:
Thanks for the work here @andrewfulton9 

FYI I'm also trying a more general solution over at https://github.com/dask/distributed/pull/4347

jrbourbeau:
@andrewfulton9 just wanted to give you a heads up that we ended up going down a more general path in https://github.com/dask/distributed/pull/4406 to specify worker restrictions, task priority, etc. with Dask collection. See my comment over in the original issue https://github.com/dask/distributed/issues/3590#issuecomment-803397314

PR #4355, opened 2020-12-12
Cast `SortedDict`s to `dict`s in a few key places & other minor changes

jakirkham:
* Use a `SortedDict` for `self.idle` (can handle keys faster ourselves; plus leads to the next thing)
* Cast `SortedDict`s to `dict`s when viewing them (allows for the Python C API to be used on them)
* Type some variables containing integers for faster checks and mathematical operations
* Use `WorkerState` C-level attribute for faster access (missed previously)

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4710,7 +4710,7 @@ class Scheduler(ServerNode):
                 type(ws),
                 ws,
             )
-            assert ws.address in self.workers
+            assert ws._address in self.workers
 
         return ws
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -86,7 +86,7 @@ from .variable import VariableExtension
 from .protocol.highlevelgraph import highlevelgraph_unpack
 
 try:
-    from cython import bint, cclass, double, Py_hash_t, Py_ssize_t
+    from cython import bint, cast, cclass, double, Py_hash_t, Py_ssize_t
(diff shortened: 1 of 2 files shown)


jakirkham:
Should add this resulted in a 25% speed up of `decide_worker`, which was the main bottleneck of `transition_waiting_processing`, but is no longer after this change.

distributed/scheduler.py
```diff
@@ -4693,24 +4700,26 @@ def decide_worker(self, ts: TaskState) -> WorkerState:
...
             )
         else:
-            worker_pool = self.idle or self.workers.values()
-            n_workers = len(worker_pool)
+            worker_pool = self.idle or self.workers
+            worker_pool_dv = cast(dict, worker_pool)
+            n_workers: Py_ssize_t = len(worker_pool_dv)
             if n_workers < 20:  # smart but linear in small case
-                ws = min(worker_pool, key=operator.attrgetter("occupancy"))
+                ws = min(worker_pool_dv.values(), key=operator.attrgetter("occupancy"))
```
jakirkham, line 4712: This would shave off a bit more time as it is faster to call `.values()` on a `dict` (one C call) as opposed to a `SortedDict`. Though this would mean the `sort` doesn't preserve the `SortedDict`'s order. However it appears `test_correct_bad_time_estimate` expects the order to be preserved. So will go ahead and drop this, which fixes that issue.

PR #4358, opened 2020-12-13
Optimizations of `*State` and `Task*` objects and stealing

jakirkham:
* Mark all `*State` and `Task*` `class`es as `final` (optimizes method class by disabling inheritance)
* Use `ccall` decorator  `*State` and `Task*` (provide C API functions that can be called quickly in Cython) 
* Use `declare` to type some module variables ( annotation doesn't work at the module level https://github.com/cython/cython/issues/3944 ) 
* Add `TaskState.get_nbytes_deps` to speed up computation in stealing
* Annotate `get_task_duration` used both to evaluate and steal (also used in `transition_waiting_processing`)
* Annotate `bandwidth` in `get_comm_cost`
* Some tweaks to `check_idle_saturated` to minimize zero division checks
* Optimize `get_nbytes` to work directly in C and inline itself when possible

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -145,11 +145,12 @@ logger = logging.getLogger(__name__)
 
 
 LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
-DEFAULT_DATA_SIZE = parse_bytes(
-    dask.config.get("distributed.scheduler.default-data-size")
+DEFAULT_DATA_SIZE = declare(
+    Py_ssize_t, parse_bytes(dask.config.get("distributed.scheduler.default-data-size"))
 )
-UNKNOWN_TASK_DURATION = parse_timedelta(
-    dask.config.get("distributed.scheduler.unknown-task-duration")
+UNKNOWN_TASK_DURATION = declare(
+    double,
+    parse_timedelta(dask.config.get("distributed.scheduler.unknown-task-duration")),
 )
 
 DEFAULT_EXTENSIONS = [
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -86,7 +86,7 @@ from .variable import VariableExtension
 from .protocol.highlevelgraph import highlevelgraph_unpack
 
 try:
-    from cython import bint, cast, cclass, double, Py_hash_t, Py_ssize_t
+    from cython import bint, cast, ccall, cclass, double, Py_hash_t, Py_ssize_t
 except ImportError:
     from ctypes import (
         c_double as double,
@@ -98,6 +98,9 @@ except ImportError:
 
     def cast(T, v, *a, **k):
         return v
+
+    def ccall(func):
+        return func
 
     def cclass(cls):
         return cls
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4419,8 +4419,9 @@ class Scheduler(ServerNode):
             s: set = self.unknown_durations[ts._prefix._name]
             s.add(ts)
             if default < 0:
-                default = UNKNOWN_TASK_DURATION
-            return default
+                duration = UNKNOWN_TASK_DURATION
+            else:
+                duration = default
 
         return duration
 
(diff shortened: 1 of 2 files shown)


jakirkham:
This results in a 30% speed up of `steal_time_ratio`, which is the main bottleneck in `put_key_in_stealable` (followed by the time to log).

distributed/scheduler.py
```diff
@@ -1328,6 +1382,13 @@ def validate(self):
 
                 pdb.set_trace()
 
+    def get_nbytes_deps(self):
+        nbytes: Py_ssize_t = 0
+        ts: TaskState
+        for ts in self._dependencies:
+            nbytes += ts.get_nbytes()
+        return nbytes
```
mrocklin, line 1390: If this is used only once then I wouldn't mind avoiding the function definition and moving this code to the call site.  Thoughts?

distributed/scheduler.py
```diff
@@ -116,8 +153,12 @@ def cclass(cls):
...
 LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
-DEFAULT_DATA_SIZE = parse_bytes(
-    dask.config.get("distributed.scheduler.default-data-size")
+DEFAULT_DATA_SIZE = declare(
+    Py_ssize_t, parse_bytes(dask.config.get("distributed.scheduler.default-data-size"))
+)
+UNKNOWN_TASK_DURATION = declare(
+    double,
+    parse_timedelta(dask.config.get("distributed.scheduler.unknown-task-duration")),
 )
```
jakirkham, line 162: Ideally we would annotate these and that would be sufficient. Unfortunately that didn't appear to work ( https://github.com/cython/cython/issues/3944 ). So used `declare` here instead, which does handle this correctly.

distributed/scheduler.py
```diff
@@ -503,6 +546,7 @@ def used_resources(self):
     def versions(self):
         return self._versions
 
+    @ccall
     def clean(self):
```
jakirkham, line 550: Need to wait for Python 3.6 to be dropped before we can annotate this with the `return` type since we need to reference `WorkerState`, which is still being defined. Once that happens we can add `from __future__ import annotations`. That said, this method is used infrequently and knowing the `return` type doesn't look particularly useful in those places. So this is non-pressing. Just noting for future reference.

https://github.com/cython/cython/issues/3935#issuecomment-739115351

distributed/scheduler.py
```diff
@@ -528,7 +572,9 @@ def __repr__(self):
             len(self._processing),
         )
 
-    def identity(self):
+    @ccall
+    @exceptval(check=False)
+    def identity(self) -> dict:
```
jakirkham, line 577: As we are annotating the `return` with a Python object and using `ccall`, Cython gets a bit confused. Cython typically wants to add exception handling code when return a non-Python object (like `double`), but this isn't needed with Python objects as these are pointers, which will be set to `NULL` on an exception. Cython already checks for this afterwards (as it does with all Python objects). Hence we add `exceptval(check=False)` to stop Cython from adding this extra code, which fixes the issue.

distributed/scheduler.py
jakirkham: We are Cythonizing this function as it is in `scheduler.py`, which would not be the case if it lived in `stealing.py`. Of course we could consider Cythonizing `stealing.py`, but for one line of code I'm not sure it is worth it.

distributed/stealing.py
```diff
@@ -120,7 +120,7 @@ def steal_time_ratio(self, ts):
         if not ts.dependencies:  # no dependencies fast path
             return 0, 0
 
-        nbytes = sum(dep.get_nbytes() for dep in ts.dependencies)
+        nbytes = ts.get_nbytes_deps()
```
jakirkham, line 123: JFYI this is where we are using `get_nbytes_deps`

distributed/scheduler.py
mrocklin: OK, sounds good.  Thanks for the explanation.

Issue #3898, opened 2020-06-15
Scheduler memory leak / large worker footprint on simple workload


chrisroat:
**What happened**:

Running an embarrassingly parallel map_overlap workload may be causing a memory leak in the scheduler.  Upon completion, releasing the tasks and restarting the client does not reclaim the memory.  The example below, with 200k tasks, shows a jump in scheduler memory from 100MB to 1.3GB while running the graph.  After `client.restart`, it remains at 1.1GB.

In addition, the memory of the workers climb into the "yellow", where I believe swapping to disk begins to happen.  Given the parallel nature of this workload, workers ought to be able discard pieces when they are done with them.

From a performance perspective,  during `client.compute`, the scheduler gets unresponsive (it takes 20ish seconds to start), presumably because its loading a large graph.  I have seen this cause already running computations to start erroring.  I've seen lost keys and KilledWorkers.

And finally, anecdotally, it sometimes happens that one worker runs hot, getting 10x the tasks of other workers.  Eventually, forward progress halts.  I now watch for this, and then kill that worker, which redistributes the work and finishes the job.  (I'm using dask-gateway on K8s).

**What you expected to happen**:

- The scheduler should not use up additional memory once a computation is done.
- Workers should shard a parallel job so that each shard can be discarded when done, keeping a low worker memory profile
- Loading a graph should not disrupt ongoing computation

**Minimal Complete Verifiable Example**:

```python
import dask.array as da
import distributed
client = distributed.Client(n_workers=4, threads_per_worker=1, memory_limit='10GB')
arr = da.zeros((50, 2, 8192, 8192), chunks=(1, -1, 512, 512))
result = arr.map_overlap(lambda x: x, depth=(0,0,200,200))
store = result.to_zarr('/media/ssd/test/memory_scheduler.zarr', compute=False, overwrite=True)
future = client.compute(store)
```

**Environment**:

- Dask version: 2.18.1
- Distributed version: 2.18.0
- Python version: 3.7
- Operating System: Ubuntu 18.04
- Install method (conda, pip, source): conda

chrisroat:
I also see a lot of this in the logs
```
distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 7.03 GB -- Worker memory limit: 10.00 GB
```

mrocklin:
For memory issues on the worker I recommend that you look through https://github.com/dask/dask/issues/3530

[...]

PR #4360, opened 2020-12-14
Deadlocks and infinite loops connected to failed dependencies

fjetter:
Two fixes in here which are unfortunately a bit intertwined.

1. if `send_rcv` fails for whatever reason (e.g. deserialization error) but doesn't close the comm, the other end might still wait for a response. in the meantime this comm is resused, causing some weird behaviour, see also #4316 I'll still need to write a test capturing the behaviour described in there

2. If we are in the case of a bad_dep, the current implementation never detects this and we're stuck in an infinite loop since the condition `dep.suspicious_count > 5` is *never* hit. It is never hit because there are multiple calls to `release-keys`issued causing the worker to loose all its memory, therefore the suspicious count never climbs up but the task instance is reinitialized

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -77,8 +77,13 @@ LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
 
 no_value = "--no-value-sentinel--"
 
-PENDING = ("waiting", "ready", "constrained")
-PROCESSING = ("waiting", "ready", "constrained", "executing", "long-running")
+
+# All valid error-free states for which the keys are not in memory yet but we
+# can assume the tasks are available eventually. Not to be confused with the scheduler "processing" state
+PROCESSING = ("waiting", "ready", "constrained", "executing", "long-running", "flight")
+
+# Ready keys are supposed to be computed on this worker. All dependencies are
+# available and the task is waiting for the next available slot of an executor or resource
 READY = ("ready", "constrained")
 
 
@@ -116,7 +121,7 @@ class TaskState:
         The worker that current task data is coming from if task is in flight
     * **waiting_for_data**: ``set(keys of dependencies)``
         A dynamic verion of dependencies.  All dependencies that we still don't
-        have for a particular key.
+        have for a particular key. Some of these keys may be computed on this worker while others are gathered.
     * **resource_restrictions**: ``{str: number}``
         Abstract resources required to run a task
     * **exception**: ``str``
@@ -1477,6 +1482,8 @@ class Worker(ServerNode):
                     )
                     self.send_task_state_to_scheduler(ts)
                     return
+                self.log.append((key, "dependency to runnable"))
+                logger.debug("Asked to compute dependency: %s: %s", key, ts.state)
             else:
                 self.log.append((key, "new"))
                 self.tasks[key] = ts = TaskState(key=key)
@@ -1622,10 +1629,6 @@ class Worker(ServerNode):
             if self.validate:
                 assert ts.state == "flight"
                 assert not ts.waiting_for_data
-                assert all(
-                    dep.key in self.data or dep.key in self.actors
-                    for dep in ts.dependencies
-                )
                 assert all(dep.state == "memory" for dep in ts.dependencies)
                 assert ts.key not in self.ready
                 assert ts.runspec is not None
@@ -1649,10 +1652,8 @@ class Worker(ServerNode):
                 assert ts.state == "waiting"
                 assert not ts.waiting_for_data
                 assert all(
-                    dep.key in self.data or dep.key in self.actors
-                    for dep in ts.dependencies
-                )
-                assert all(dep.state == "memory" for dep in ts.dependencies)
+                    dep.state == "memory" for dep in ts.dependencies
+                ), ts.dependencies
                 assert ts.key not in self.ready
                 assert ts.runspec is not None
 
@@ -1843,6 +1844,8 @@ class Worker(ServerNode):
 
                 key = self.data_needed[0]
 
+                # Key might've been already cancelled since it was put into
+                # data_needed. Worker.release_key doesn't clean up data_needed
                 if key not in self.tasks:
                     self.data_needed.popleft()
                     changed = True
@@ -2139,7 +2142,7 @@ class Worker(ServerNode):
                     if ts.key in data:
                         self.transition(ts, "memory", value=data[d])
 
-                    if ts.state != "memory":
+                    if ts.state == "flight":
                         self.transition(ts, "waiting", worker=worker)
                         missing.add(ts)
 
@@ -2148,9 +2151,6 @@ class Worker(ServerNode):
                 if missing:
                     await self.handle_missing_dep(*missing, worker=worker)
                     if cause not in self.data_needed:
-                        if self.validate:
-                            assert isinstance(cause, str)
-                            assert cause in self.tasks
                         self.data_needed.append(cause)
 
                 if self.validate:
@@ -2953,11 +2953,7 @@ class Worker(ServerNode):
                     assert ts in dep.dependents
                 for key in ts.waiting_for_data:
                     ts_wait = self.tasks[key]
-                    assert (
-                        ts_wait.state == "flight"
-                        or ts_wait.state == "waiting"
-                        or ts_wait.who_has.issubset(self.in_flight_workers)
-                    )
+                    assert ts_wait.state in PROCESSING, ts_wait
                     waiting_keys.add(key)
                 if ts.state == "memory":
                     assert isinstance(ts.nbytes, int)
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -454,6 +454,8 @@ class Worker(ServerNode):
             ("ready", "executing"): self.transition_ready_executing,
             ("ready", "memory"): self.transition_ready_memory,
             ("ready", "error"): self.transition_ready_error,
+            ("ready", "waiting"): self.transition_ready_waiting,
+            ("constrained", "waiting"): self.transition_ready_waiting,
             ("constrained", "executing"): self.transition_constrained_executing,
             ("executing", "memory"): self.transition_executing_done,
             ("executing", "error"): self.transition_executing_done,
@@ -1720,7 +1722,25 @@ class Worker(ServerNode):
         self.send_task_state_to_scheduler(ts)
 
     def transition_ready_memory(self, ts, value=None):
+        if value:
+            self.put_key_in_memory(ts, value=value)
         self.send_task_state_to_scheduler(ts)
+
+    def transition_ready_waiting(self, ts):
+        """
+        This transition is common for work stealing
+        """
+        ts.state = "waiting"
+        ts.runspec = None
+        # FIXME GH4413 Replace with recommended transition to forgotten
+        # If there are no dependents anymore, this has been a mere dependency
+        # and it was not intended to be executed on this worker. In this case we
+        # need to release the key ourselves since the scheduler will no longer
+        # propagate the deletion to this worker
+        if not ts.dependents:
+            self.release_key(ts.key)
+            return
+        return ts.state
 
     def transition_constrained_executing(self, ts):
         self.transition_ready_executing(ts)
@@ -2259,8 +2279,9 @@ class Worker(ServerNode):
         # There may be a race condition between stealing and releasing a task.
         # In this case the self.tasks is already cleared. The `None` will be
         # registered as `already-computing` on the other end
-        if key in self.tasks:
-            state = self.tasks[key].state
+        ts = self.tasks.get(key)
+        if ts:
+            state = ts.state
         else:
             state = None
 
@@ -2268,7 +2289,11 @@ class Worker(ServerNode):
         self.batched_stream.send(response)
 
         if state in ("ready", "waiting", "constrained"):
-            self.release_key(key, reason="steal_request")
+            # FIXME GH4413 We no longer want to execute this task but there is
+            # no state to identify this. We still need to track it,
+            # otherwise this worker may never fetch it in case it is still
+            # required
+            self.transition(ts, "waiting")
 
     def release_key(self, key, reason=None, cause=None, report=True):
         try:
@@ -2582,7 +2607,17 @@ class Worker(ServerNode):
                 executor_error = e
                 raise
 
-            if ts.state not in ("executing", "long-running"):
+            # We'll need to check again for the task state since it may have
+            # changed since the execution was kicked off. In particular, it may
+            # have been canceled and released already in which case we'll have
+            # to drop the result immediately
+            key = ts.key
+            ts = self.tasks.get(key)
+
+            if ts is None:
+                logger.debug(
+                    "Dropping result for %s since task has already been released." % key
+                )
                 return
 
             result["key"] = ts.key
@@ -2877,7 +2912,7 @@ class Worker(ServerNode):
 
     def validate_task_memory(self, ts):
         assert ts.key in self.data or ts.key in self.actors
-        assert ts.nbytes is not None
+        assert isinstance(ts.nbytes, int)
         assert not ts.waiting_for_data
         assert ts.key not in self.ready
         assert ts.state == "memory"
@@ -2904,7 +2939,7 @@ class Worker(ServerNode):
     def validate_task_waiting(self, ts):
         assert ts.key not in self.data
         assert ts.state == "waiting"
-        if ts.dependencies:
+        if ts.dependencies and ts.runspec:
             assert not all(dep.key in self.data for dep in ts.dependencies)
 
     def validate_task_flight(self, ts):
@@ -2940,6 +2975,9 @@ class Worker(ServerNode):
         try:
             assert len(self.data_needed) == len(set(self.data_needed))
             waiting_keys = set()
+
+            assert self.executing_count >= 0
+
             for ts in self.tasks.values():
                 assert ts.state is not None
                 # check that worker has task
@@ -2958,10 +2996,6 @@ class Worker(ServerNode):
                     ts_wait = self.tasks[key]
                     assert ts_wait.state in PROCESSING, ts_wait
                     waiting_keys.add(key)
-                if ts.state == "memory":
-                    assert isinstance(ts.nbytes, int)
-                    assert not ts.waiting_for_data
-                    assert ts.key in self.data or ts.key in self.actors
 
             # FIXME: Tracking of waiting_for_data_count is broken. Most likely
             # since tasks may be required/waited on by multiple tasks but the
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2164,6 +2164,8 @@ class Worker(ServerNode):
                     # Exponential backoff to avoid hammering scheduler/worker
                     self.repetitively_busy += 1
                     await asyncio.sleep(0.100 * 1.5 ** self.repetitively_busy)
+                    # See if anyone new has the data
+                    await self.query_who_has(*deps)
                 self.ensure_communicating()
 
     def bad_dep(self, ts):
@@ -2221,7 +2223,8 @@ class Worker(ServerNode):
             self.transition(dep, "waiting", worker=worker)
 
     async def query_who_has(self, *deps):
-        # FIXME: If this is improperly called, the query fails. We should ensure that deps is a flat list here
+        if self.validate:
+            assert all(k in self.tasks for k in deps)
         with log_errors():
             response = await retry_operation(self.scheduler.who_has, keys=deps)
             self.update_who_has(response)
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -77,7 +77,6 @@ LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
 
 no_value = "--no-value-sentinel--"
 
-IN_PLAY = ("waiting", "ready", "executing", "long-running")
 PENDING = ("waiting", "ready", "constrained")
 PROCESSING = ("waiting", "ready", "constrained", "executing", "long-running")
 READY = ("ready", "constrained")
@@ -151,10 +150,10 @@ class TaskState:
 
     """
 
-    def __init__(self, key, runspec=None):
+    def __init__(self, key):
         assert key is not None
         self.key = key
-        self.runspec = runspec
+        self.runspec = None
         self.dependencies = set()
         self.dependents = set()
         self.duration = None
@@ -405,7 +404,7 @@ class Worker(ServerNode):
         self.data_needed = deque()  # TODO: replace with heap?
 
         self.in_flight_tasks = 0
-        self.in_flight_workers = dict()
+        self.in_flight_workers = defaultdict(set)
         self.total_out_connections = dask.config.get(
             "distributed.worker.connections.outgoing"
         )
@@ -414,7 +413,6 @@ class Worker(ServerNode):
         )
         self.total_comm_nbytes = 10e6
         self.comm_nbytes = 0
-        self._missing_dep_flight = set()
 
         self.threads = dict()
 
@@ -450,6 +448,7 @@ class Worker(ServerNode):
             ("waiting", "flight"): self.transition_waiting_flight,
             ("ready", "executing"): self.transition_ready_executing,
             ("ready", "memory"): self.transition_ready_memory,
+            ("ready", "error"): self.transition_ready_error,
             ("constrained", "executing"): self.transition_constrained_executing,
             ("executing", "memory"): self.transition_executing_done,
             ("executing", "error"): self.transition_executing_done,
@@ -459,7 +458,7 @@ class Worker(ServerNode):
             ("long-running", "memory"): self.transition_executing_done,
             ("long-running", "rescheduled"): self.transition_executing_done,
             ("flight", "memory"): self.transition_flight_memory,
-            ("flight", "ready"): self.transition_flight_memory,
+            ("flight", "ready"): self.transition_flight_ready,
             ("flight", "waiting"): self.transition_flight_waiting,
         }
 
@@ -867,7 +866,14 @@ class Worker(ServerNode):
                         keys=list(self.data),
                         nthreads=self.nthreads,
                         name=self.name,
-                        nbytes={ts.key: ts.get_nbytes() for ts in self.tasks.values()},
+                        nbytes={
+                            # FIXME: Tasks and data should not be allowed to
+                            # diverge. Data should always be a subset of tasks.
+                            # For all practical purposes this should only happen
+                            # if someone modifies the dicts manually
+                            k: self.tasks.get(k, TaskState(k)).get_nbytes()
+                            for k in self.data.keys()
+                        },
                         types={k: typename(v) for k, v in self.data.items()},
                         now=time(),
                         resources=self.total_resources,
@@ -1336,6 +1342,11 @@ class Worker(ServerNode):
                     from .actor import Actor
 
                     data[k] = Actor(type(self.actors[k]), self.address, k)
+        if len(data) != len(keys):
+            logger.debug(
+                "Data request from %s but keys %s are not available."
+                % (who, set(keys) - set(data))
+            )
 
         msg = {"status": "OK", "data": {k: to_serialize(v) for k, v in data.items()}}
         nbytes = {k: self.tasks[k].nbytes for k in data if k in self.tasks}
@@ -1348,6 +1359,16 @@ class Worker(ServerNode):
             compressed = await comm.write(msg, serializers=serializers)
             response = await comm.read(deserializers=serializers)
             assert response == "OK", response
+
+        except CommClosedError:
+            logger.exception(
+                "Other end hung up during get_data with %s -> %s",
+                self.address,
+                who,
+                exc_info=True,
+            )
+            comm.abort()
+            raise
         except EnvironmentError:
             logger.exception(
                 "failed during get data with %s -> %s", self.address, who, exc_info=True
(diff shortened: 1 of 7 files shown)


mrocklin:
As always, thanks for identifying these tricky issues and reporting upstream.  I haven't yet had a time to go through things deeply here (and that time may not come), but if you're confident in things here I'm happy to defer to you.

However, currently tests are failing, so something is going wrong I think.

fjetter:
Indeed, I will need to look deeper into the tests. I opened the PR a bit prematurely but I figured this is an area where a lot is happening lately and somebody else might have an opinion

mrocklin:
+1 on opening PRs before they're polished

distributed/worker.py
```diff
@@ -1452,19 +1462,30 @@ def add_task(
...
+                    # dependencies which are waiting to be fetched (no runspec)
+                    # and for tasks ready to be computed (with runspec)
+                    if ts.state == "waiting":
+                        ts.runspec = runspec
                 if ts.state == "erred":
                     ts.exception = None
                     ts.traceback = None
                 else:
-                    ts.state = "waiting"
+                    self.transition(ts, "waiting", runspec=runspec)
```
fjetter, line 1484: Not setting the runspec for these cases can actually lead to a deadlock. I've been hunting this deadlock for a while now and I am very confident this is the root cause for #4133

I also have a suspicion that this is causing #4173 but I cannot reproduce locally

Fixes are still incomplete and I will need more time fixing CI since I cannot reproduce everything. I'm a bit confused about what the "waiting" state is actually supposed to flag since this is slightly different for dependencies and runnable tasks (see in code comment). Any pointers and opinions appreciated

distributed/nanny.py
```diff
@@ -144,7 +144,7 @@ def __init__(
 
         self.Worker = Worker if worker_class is None else worker_class
         self.env = env or {}
-        self.config = config or {}
+        self.config = config or dask.config.config
```
fjetter, line 147: Will create another PR for this. This prohibits setting worker configs in the tests

- [ ] PR for Nanny/config

distributed/worker.py
```diff
@@ -2338,8 +2317,8 @@ def rescind_key(self, key):
 
     # FIXME: this breaks if changed to async def...
     # xref: https://github.com/dask/distributed/issues/3938
-    @gen.coroutine
-    def executor_submit(self, key, function, args=(), kwargs=None, executor=None):
+    # @gen.coroutine
+    async def executor_submit(self, key, function, args=(), kwargs=None, executor=None):
```
fjetter, line 2321: This should not be part of this PR. I put this in while debugging the deadlock, trying to figure out if it was a problem with the threadpool. Changes might be viable anyhow but should be part of another PR

- [ ] PR for executor_submit changes

mrocklin:
cc @gforsyth would you be comfortable reviewing this?

fjetter:
Seems like the infinite loop I initially wanted to fix is appearing on CI (but not locally). Will need to investigate a bit further. There are also some broken tests which are failing on master as well #4374

distributed/worker.py
```diff
@@ -2451,7 +2428,9 @@ def meets_resource_constraints(self, key):
         return True
 
     async def _maybe_deserialize_task(self, ts):
+        assert ts.runspec is not None
         if not isinstance(ts.runspec, SerializedTask):
+            # FIXME: What is this case??
```
gforsyth, line 2433: Yeah, I don't know -- I updated it when adding `TaskState`.  I guessed that there was a possibility of the `runspec` already being deserialized?

gforsyth (commented):
Thanks for diving in to this @fjetter !  The changes here make sense to me overall -- just a few comments.

And we can certainly add a `transition_new_waiting` function -- most of that functionality is currently wrapped up in `add_task`

distributed/worker.py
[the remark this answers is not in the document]
fjetter: `start == finish` is treated in Worker.transition

https://github.com/dask/distributed/blob/3fb3d8c8c14800ff8f1cd76d0052985f355ecfcc/distributed/worker.py#L1532-L1537

This might still not be correct, though, since `waiting` is currently ambiguous since it refers to a dependency waiting to be computed and fetched and simultaneously to a runnable task. Both cases allow for very different transitions.

This ambiguity doesn't feel correct and I'm wondering if there is something we can do about this. IMHO, the `state` attribute should uniquely define which transition a task is allowed to perform and which not but this is currently not the case. I was already thinking about introducing another state, like `waiting_runnable` to deal with this ambiguity but I would like to only introduce a new state if absolutely necessary

distributed/worker.py
fjetter: Currently the runspec is deserialized right before it is being transitioned to executing

https://github.com/dask/distributed/blob/3fb3d8c8c14800ff8f1cd76d0052985f355ecfcc/distributed/worker.py#L2491-L2496

and 

https://github.com/dask/distributed/blob/3fb3d8c8c14800ff8f1cd76d0052985f355ecfcc/distributed/worker.py#L2510-L2515

Therefore, I would assume that as this is deserialized for all tasks in the `executing` state. We currently only allow transitions from `executing` to `long-running` and `done`. Therefore, if we trust that all task transitions are properly performed using `Worker.transition`, this case should not be possible.

Unless, of course, the transitioning itself hits any kind of exception after deserializing, i.e. essentially in `Worker.execute` but as far as I understand it we are not robust to an exception there at the moment, anyhow.

fjetter:
Ok, I can reproduce the CI failure locally... just took me 589 tries :/

distributed/worker.py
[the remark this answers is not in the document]
gforsyth: > `start == finish` is treated in Worker.transition

Ahh, yeah.  

> currently ambiguous since it refers to a dependency waiting to be computed and fetched and simultaneously to a runnable task.

yeah, this is ambiguous.  Ideally tasks that are runnable should be marked `ready`, but the default starting state is `waiting`, which leads to this ambiguity.  We might be better served by having a default task state of `new` (as you referenced in a code comment, there are no `new->?` transition functions but that might be worth doing.
I can try to put together a draft of what that might look like and see how it works (unless you're interested in tackling that yourself)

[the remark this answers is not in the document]
fjetter: >  Ideally tasks that are runnable should be marked ready

I believe it's a bit more tricky than this. I'm not necessarily talking about "ready" tasks in a sense that they can be calculated *now* but with "runnable" I am referring to tasks, this worker is supposed to calculate at some point vs dependencies the worker is only supposed to fetch. This might include speculative assignments (let's include this for a moment since this is still WIP) but also tasks which were ready at some point but one of their dependencies got lost while the worker was trying to fetch them. Therefore, there must be a state for _runnable_ tasks before the `ready` state. On scheduler side this is `released` (although imo `released` is not the best name for this state)

I believe it is worth the effort putting more though into the state mapping and a new state might be the way to go here (not sure, whether it should be called "new" or not)

I am trying to stabilize this PR without introducing any new state(s). I found another race condition in the way missing deps are handled and I believe this is doable without a new state. I would still propose to discuss this additional state topic but on a different PR/issue. If you already have an idea, feel free to suggest something. We might also want to look into more documentation of what the states are _supposed_ to do (also out of scope for this PR)

distributed/nanny.py
fjetter: https://github.com/dask/distributed/pull/4378

fjetter:
Ok the (hopefully) last race condition is pretty bad and I don't have a resolution for it, yet, but I believe I figured out what's happening.

We have two competing mechanisms to deal with missing dependencies (the cluster doesn't distinguish between actually missing deps, e.g. a worker died, and deps which could not be fetched due to some sort of exceptions, comm related or otherwise; everything is a missing dep)

### 1 Worker internal

If the worker is unsuccessful in fetching the dependencies after 5 times, it flags the dependency as bad and gives up. This is what my test intended to capture

https://github.com/dask/distributed/blob/3fb3d8c8c14800ff8f1cd76d0052985f355ecfcc/distributed/worker.py#L2167-L2170

### 2 Escalate to scheduler

The competing mechanism escalates immediately to the scheduler via 
https://github.com/dask/distributed/blob/3fb3d8c8c14800ff8f1cd76d0052985f355ecfcc/distributed/worker.py#L2128-L2130

https://github.com/dask/distributed/blob/3fb3d8c8c14800ff8f1cd76d0052985f355ecfcc/distributed/scheduler.py#L3334-L3351

which transitions the faulty dependency to `released`. In the scenario where we have two sequential tasks A -> B and A is the faulty dependency, this means A is transitioned `memory -> released` which also triggers As depentent B to be transitioned, namely from `processing` to `waiting`. Since there is no `processing -> waiting` transition, it transitions B via `processing -> released -> waiting` where the first `processing->released` triggers a `release-key` which resets the suspicious counter of the worker internal process -> infinite loop

I am trying to find a solution which does not require me to rewrite this entire thing but it feels to me as if this mechanism requires a redesign and cleanup w/ appropriate tests (testing this is hard)

jrbourbeau (commented):
Just merged the main branch to resolve some unrelated CI failures

fjetter:
Working on a few more changes and will update by tomorrow. I hope I have stabilised this by now but I'm not certain, yet.

fjetter:
Changes got a bit bigger than I was hoping for. Here are the highlights

* Instead of implementing multiple ways to deal with missing dependencies there is now one. How this works can be discussed but this proposes to query `who_has` from scheduler and make a decision based on the result. If the who_has of the scheduler is different to what the worker knows, that's a valid case for a retry and the worker updates it's state of `who_has` and tries again. If the `who_has` is in sync with the scheduler, I deduce something is broken with the world and escalate to the scheduler to fix this, i.e. scheduler will release and reschedule the broken key
* If something happens during dependency collection which is not a connection error or a missing one, we will immediately flag the task as faulty, working under the assumption that this is unrecoverable and retrying won't solve anything. This is **much stricter** than before and we should discuss this. It helped me to avoid the last race condition I was talking about but this might also be avoided differently if this behaviour seems to drastic

In terms of changed behaviour that _should_ be it. Most code changes where some kind of missing or improper transitions or places I consider dead code. Some of these dead code places I just deleted, others raise exceptions.


Apologies for the big PR but as you can imagine this is very hard to break apart

distributed/worker.py
```diff
@@ -1444,56 +1460,50 @@ def add_task(
...
+                    # workers = tuple(w_addr for w_addr in workers if w_addr != self.address)
                 if dependency not in self.tasks:
                     self.tasks[dependency] = dep_ts = TaskState(key=dependency)
-                    dep_ts.state = (
-                        "waiting" if dependency not in self.data else "memory"
-                    )
-
-                dep_ts = self.tasks[dependency]
-                self.log.append((dependency, "new-dep", dep_ts.state))
+                    dep_ts.state = "waiting"
```
gforsyth, line 1499: I'm certainly not sure that my approach was the correct one, but if we call `release_key` on a task that has any dependents, then the key will end up popped out of `self.tasks` without also clearing the corresponding item out of `self.data` -- so if we set the state to `waiting` here without checking for presence in `self.data` it can fail the check in `validate_task_waiting`

fjetter: Sure, will check this section again. In general I'm not very happy with the approach of modifying the state attribute manually and maybe this should also by done via a proper transition step. tbh, I haven't fully understood your comment, yet, but if multiple tasks / dependencies must be adjusted this sounds like something which should be handled by a transition method anyhow I suppose

gforsyth: My comment was very unclear (having just re-read it).

It is possible at this point to have a task whose result is present in `self.data` even though there is no corresponding `TaskState` object in `self.tasks`.  If that is the case, then the correct `state` is `memory`.  If we blanket set it to `waiting` it will fail the assertions in `validate_task_waiting`.

fjetter: How can data end up in memory without it being a task? (other than manual manipulation of the dict)

fjetter: I indeed am seeing this error case you are describing but I'm wondering if this is an allowed state altogether. Even when effectively setting data _manually_ (i.e. using the respective API) a `TaskState` object is created if it doesn't exist.

https://github.com/dask/distributed/blob/49e19c010b101c629691c73d62caee7140bb92d9/distributed/worker.py#L1387-L1398

 I was working under the assumption that `data` is always a subset of `tasks`

fjetter: I can find a section of code which doesn't delete data even though the key is released if there are dependents of the task. 
https://github.com/dask/distributed/blob/49e19c010b101c629691c73d62caee7140bb92d9/distributed/worker.py#L2255-L2268
I can somehow understand this reasoning but am wondering how the data is then cleaned up once it is no longer required since `release_key` is the only interface on how other servers can interact with the tasks and the data.

Even `delete_data` itself uses this, see https://github.com/dask/distributed/blob/49e19c010b101c629691c73d62caee7140bb92d9/distributed/worker.py#L1405-L1412

Doesn't this introduce memory leakage?

gforsyth: > Doesn't this introduce memory leakage?

Probably?  We could try removing that case (in `release_key`) and see what breaks.

fjetter: Ok I think I found what mechanism allowed this to not be discovered / function properly. There is something like a reference counting ongoing, i.e. the dependents and dependencies are all removed from the task states individually, e.g. 
https://github.com/dask/distributed/blob/49e19c010b101c629691c73d62caee7140bb92d9/distributed/worker.py#L1734-L1737
and 
https://github.com/dask/distributed/blob/49e19c010b101c629691c73d62caee7140bb92d9/distributed/worker.py#L2271-L2277

This allows for a cleanup of the data even though the data is no longer tracked as a task. I deleted some of the code since I considered dependencies/dependents as a immutable property of a task and not a runtime dependent one. That confused me quite a bit. I don't necessarily agree with the implementation right now but I'll restore everything as it was to reduce the number of changes.

gforsyth: > That confused me quite a bit. I don't necessarily agree with the implementation right now but I'll restore everything as it was to reduce the number of changes.

I'm definitely game to sync up at some point and sketch out a cleaner implementation for this.

fjetter: > I'm definitely game to sync up at some point and sketch out a cleaner implementation for this.

Sounds good. Once this PR is finally stable, I'll open an issue and summarise the points I've seen where I think we can improve and we can sync up on the topic.

fjetter: I updated the PR with my current state. Unfortunately, reinstating the full dependency/release/refcounting as described above seemed to make things worse :(

gforsyth:
Looking over the changes now -- 

I've run the two `test_worker` failures from travis on my OSX laptop 50 times each and they've passed every time, so I'm fairly certain those are spurious.

The test_stress failure also fails on my laptop, but my work laptop has several networking issues due to security stuff, so it may not be relevant.

gforsyth (commented):
The changes overall look good to me.  I'm seeing a few assertion errors when I run `test_failed_workers.py::test_broken_worker_during_computation`.  Every 5th time or so, there's an assertion error when a worker tries to transition a task from `ready -> waiting` which results in a `KeyError`.  
I'm fairly certain this only happens _before_ the worker in question is killed, so it may just be some messiness as the worker is coming down before a new worker takes its place.

```
tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornad
o.platform.asyncio.AsyncIOLoop object at 0x7f873d8dc5b0>>, <Task finished name='Task-722' coro=<Worker.gather_dep() done, de
fined at /Users/gil/github.com/dask/distributed/distributed/worker.py:2005> exception=KeyError(('ready', 'waiting'))>)
  File "/Users/gil/miniforge3/envs/dask-distributed/lib/python3.8/site-packages/tornado/ioloop.py", line 743, in _run_cal
lback
    ret = callback()
  File "/Users/gil/miniforge3/envs/dask-distributed/lib/python3.8/site-packages/tornado/ioloop.py", line 767, in _discard
_future_result
    future.result()
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2143, in gather_dep
    self.transition(ts, "waiting", worker=worker)
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 1551, in transition
    func = self._transitions[start, finish]
KeyError: ('ready', 'waiting')
distributed.nanny - WARNING - Restarting worker
```

and occasional assertion errors after the new workers come up (these two problems never occur during the same test run, it's either one or the other)

```
distributed.worker - ERROR - 
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2965, in validate_state
    assert ts.key in self.data or ts.key in self.actors
AssertionError
distributed.worker - ERROR - 
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2274, in release_key
    self.validate_state()
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2965, in validate_state
    assert ts.key in self.data or ts.key in self.actors
AssertionError
distributed.core - ERROR - 
  File "/Users/gil/github.com/dask/distributed/distributed/core.py", line 563, in handle_stream
    handler(**merge(extra, msg))
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 1430, in delete_data
    self.release_key(key, reason="delete")
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2274, in release_key
    self.validate_state()
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2965, in validate_state
    assert ts.key in self.data or ts.key in self.actors
AssertionError
distributed.worker - ERROR - 
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 985, in handle_scheduler
    await self.handle_stream(
  File "/Users/gil/github.com/dask/distributed/distributed/core.py", line 563, in handle_stream
    handler(**merge(extra, msg))
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 1430, in delete_data
    self.release_key(key, reason="delete")
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2274, in release_key
    self.validate_state()
  File "/Users/gil/github.com/dask/distributed/distributed/worker.py", line 2965, in validate_state
    assert ts.key in self.data or ts.key in self.actors
AssertionError
```

I'm trying to figure out the root cause (of mostly the second error) but I don't think it's any worse than the existing behavior and I wouldn't consider it a blocker.

fjetter:
Rechecked the mentioned tests myself and I also can see a 

```
  File "/Users/fjetter/miniconda3/envs/dask-dev/lib/python3.8/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "/Users/fjetter/miniconda3/envs/dask-dev/lib/python3.8/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
    future.result()
  File "/Users/fjetter/workspace/distributed/distributed/worker.py", line 2153, in gather_dep
    assert cause in self.tasks
```

in `test_broken_worker_during_computation`

back to it...

fjetter:
I believe the first AssertionError mentioned above about the state transition ready->waiting is merely a wrong transition and should be easily fixed.

The second looks a bit more troublesome. While I agree that it's probably not worse than before, having a task in state "memory" w/out any data sounds quite serious, even if it is very rare. I'll have another look

fjetter:
Currently, I am observing a few failures on travis which seem unrelated. I'll retrigger the runs to see if they are persistent

Py3.6

* `distributed/tests/test_stess::test_cancel_stress` with an `RuntimeError: ('Unclosed Comms', [])` error 
* `distributed/tests/test_stess:: test_close_connections` with a simple timeout
* `distributed/tests/test_worker::test_wait_for_outgoing` with `RuntimeError: cannot schedule new futures after shutdown`

Py3.7

* `test_stop_doing_unnecessary_work` timeout
* `distributed/tests/test_stess:: test_close_connections` with a simple timeout
* `distributed/tests/test_worker::test_wait_for_outgoing` with `RuntimeError: cannot schedule new futures after shutdown`

Py3.8 a few failures which look similar to above but eventually segfault https://github.com/dask/distributed/issues/4387

mrocklin:
I recommend opening an empty PR against master to test CI

fjetter:
Wow, I'm really falling into a rabbit hole with this. I was investigating the `test_stop_doing_unnecessary_work` since its failure seemed out of place. Turns out this is connected. I'm not sure if it is connected to my actual change but it is connected to the same root cause.

Actually looking at the logs reveals that the `executing_count` of the workers become negative (see the running status of the repr of the workers) and the while loop of this test keeps on spinning for a negative value.

<details>
```
c = <Client: 'tcp://127.0.0.1:41064' processes=2 threads=3, memory=16.74 GB>
1839s = <Scheduler: "tcp://127.0.0.1:41064" processes: 0 cores: 0>
1840a = <Worker: 'tcp://127.0.0.1:40927', 0, Status.closed, stored: 1, running: -1/1, ready: 0, comm: 0, waiting: 0>
1841b = <Worker: 'tcp://127.0.0.1:45391', 1, Status.closed, stored: 2, running: -2/2, ready: 0, comm: 0, waiting: 0>
1842
1843    @gen_cluster(client=True)
1844    async def test_stop_doing_unnecessary_work(c, s, a, b):
1845        futures = c.map(slowinc, range(1000), delay=0.01)
1846        await asyncio.sleep(0.1)
1847    
1848        del futures
1849    
1850        start = time()
1851        while a.executing_count:
1852            await asyncio.sleep(0.01)
1853>           assert time() - start < 0.5
1854E           assert (1610380514.3005424 - 1610380510.597802) < 0.5
1855E            +  where 1610380514.3005424 = time()
```

</details>

This negative value comes from a kind of "doublefree". The counter is decreased once during `release_key` (which acts as a transition function when triggered by the scheduler) and once by the `execution_done` transition. I.e. there is again a race condition for cases where tasks are released while actually being executed.

This connects to https://github.com/dask/distributed/blob/658fccc3268bd16226cc5815bf00eed349fe1553/distributed/worker.py#L2584-L2585 being improperly changed in https://github.com/dask/distributed/pull/4107 . Before the change in #4107 this was fetching the "current" state of the task instead of the "stale" state of the task instance in that scope. If I fix this, the test fails even harder...

------

I'm slowly wondering how we want to proceed with this since I continue to find issues and am trying to plug all the holes at once in this PR. That makes the entire thing very messy and almost impossible to review, I assume. Merging anything might make things worse in the meantime, though. Any suggestions on how to proceed here?

mrocklin:
> That makes the entire thing very messy and almost impossible to review, I assume. Merging anything might make things worse in the meantime, though. Any suggestions on how to proceed here?

That is a fair concern.  Ideally incremental PRs would be better, but we're also fixing bugs after a larger change (`worker.py::TaskState`) and so that may or may not be possible.  FWIW, tests are sufficiently robust in distributed that if both you and tests agree that the fix is a good one then I'm reasonably confident merging it in, even without diligent review.  

Of course, that sort of relaxed behavior is also what caused this problem in the first place.

gforsyth:
> If I fix this, the test fails even harder...

@fjetter do you have this fix pushed up to a branch somewhere?

fjetter:
Ok, the last one was connected to work stealing. I mentioned this as a potential issue over in https://github.com/dask/distributed/issues/4413 already. My solution is, for now, to put a stolen task into "waiting for dependencies" which translates to the actual state `waiting` + `runspec == None` if there are no dependents, otherwise it is released.

That was the cause for the breakage of at least `test_stop_doing_unnecessary_work` and `test_close_connections`. While my fixes run through smoothly locally, CI still seems to have issues with it.

fjetter:
I think the CI failures are finally "expected". However, doing some _real_ stress tests with this still reveal instabilities.

There are some failures which are probably invalid in-memory task states

<details>
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/worker.py", line 2572, in execute
    data[k] = self.data[k]
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/zict/buffer.py", line 80, in __getitem__
    raise KeyError(key)
KeyError: "('repartition-415-5a8ec5d8ce9f7fa507c212dd1c024d87', 83)"

  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/worker.py", line 2576, in execute
    data[k] = Actor(type(self.actors[k]), self.address, k, self)
</details>

and some invalid state transitions triggered by some scheduler <-> worker interaction/rescheduling, I suspect again work stealing of some kind

<details>
('executing', 'ready')
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/worker.py", line 1538, in add_task
    self.transition(ts, "ready")
  File "/mnt/mesos/sandbox/venv/lib/python3.6/site-packages/distributed/worker.py", line 1560, in transition
    func = self._transitions[start, finish]
KeyError: ('executing', 'ready')
</details>

Unfortunately, these are causing hard deadlocks and I don't trust this to be merged just yet

mrocklin:
Hi @fjetter is there something that I can run in order to expose these issues?

With regards to unexpected transitions, what I would probably do in this situation is call `Worker.story(key)` on any such situation in order to see what happened.  That might help to pinpoint the issue.

jrbourbeau:
@mrocklin I ran into the first issue @fjetter listed running the example at https://github.com/dask/distributed/issues/4430#issue-786248149

mrocklin:
Yeah, I was hoping to find something reproducible locally, but I'll try taking that and putting that into a stress test.

gforsyth:
> Yeah, I was hoping to find something reproducible locally, but I'll try taking that and putting that into a stress test.

I think Florian fixed all of my local reproducers -- I've run the previous failing tests 50 times each and can't get them to fail, so it may require something at scale

fjetter:
@mrocklin Yes I am aware of the `Worker.story` but this is often not enough to really understand what's going on (at least I'm still often in the dark) Sometimes it's cross interaction of the scheduler when dealing with missing dependencies, sometimes it's workstealing, sometimes a wrong transition and it's very hard to distinguish.

I'm fairly certain that most, if not all, problems are connected to how `Worker.add_task` and `Worker.release_key` works and am currently considering starting from scratch.

What's most frustrating is that most issues cannot reliably reproduced, at least not by unit tests. Unit tests hit these cases but only every few 100 times. (bash script with an endless loop...)

fjetter:
What I noticed is that a lot of the internal state machine is effectively tested using the `Worker.validate` and respective `TaskState` methods. What's a bit unfortunate is that I occasionally observe these assertions to actually raise but tests run through anyhow. I believe we should figure out another mechanism such that the tests fail if one of these special validation errors are raised.
Don't have any good ideas how at the moment, but in the future this might give us an opportunity to detect these issues earlier

mrocklin:
> Unit tests hit these cases but only every few 100 times. (bash script with an endless loop...)

You might also consider iterating from within an ipython session.  For fast unit tests you can often run the test several times a second.

I have some time to look into this.  I'm curious how best I can spend this time.  Do you have any suggestions?  

Also, I know that it's late on a Friday for you, but if you want to connect over video on this let me know.  I would be curious to learn more about what is going on and your thoughts on how to fix it.  

I'm free 6-7pm Central European time if that's interesting to you.

mrocklin:
I would expect the validation tests to stop a `gen_cluster` test, but maybe not a stress test that used Nannys

fjetter:
I opened another smaller PR https://github.com/dask/distributed/pull/4432 with a minimal set of changes. With this I can still see assertion errors (but the tests are not failing ;) e.g.`test_worker_who_has_clears_after_failed_connection` no Nanny )

Where this PR escalated was when I tried reworking the missing keys mechanism, I omitted in #4432. I believe there is a lot of ambiguity in how tasks are managed and released, in particular in cases of missing keys, comm failures or missing workers (which also happens in stress situations a lot). This afternoon I was investigating scheduler handlers and the way the worker escalates missing or erroneous tasks where I noticed the handlers `missing-data` and `release` to be very similar but are used in very different circumstances. At least the `missing-data` handler is triggering the race condition I was sketching out in https://github.com/dask/distributed/pull/4360#issuecomment-748195179 and I suspect cleaning this up / clarifying the intentions would help a bit


@mrocklin I would love to connect but my internet is dropping every few minutes today and its getting late already. Maybe we can find a spot next week, assuming nobody else fixed it by then :)

mrocklin:
Sounds good.   In the mean time I'm curious, are there particular things that you're doing to replicate failures?  Or is this mostly just the normal test suite?  If so, which tests in particular?

gforsyth:
in #4432 you can get some errors (although the test passes) running `pytest test_failed_workers.py::test_worker_who_has_clears_after_failed_connection -s`

mrocklin:
Thanks @gforsyth .  That's helpful.  Reproduced.

distributed/worker.py
chrisroat: > Doesn't this introduce memory leakage?

Seeing the question about memory leak, I'm curious if this could be related to https://github.com/dask/distributed/issues/3898.

fjetter:
I'll close this PR now since it escalated pretty badly. The deadlock issue was (at least partially) addressed in https://github.com/dask/distributed/pull/4432 and I suggest to open new PRs for follow up issues. Thanks everyone for review and input!

distributed/worker.py
fjetter: If my argumentation about the memory leak was true, the leak would happen on the worker since the worker would not release the data anymore. From my understanding, #3898 is about a potential leak on the scheduler.
Also, the leak I was referring to does not exist and I was mistaken. The worker applies some refcounting by tracking the `dependents` of each task and releases the data properly once all dependents are gone, i.e. the data is no longer needed. I missed this mechanism since it's a bit spread out in this module but from my understanding it works as intended.

Issue #925, opened 2017-03-02
upload directories


jacksonloper:
Sometimes I'm working on a python package which is just a directory with contents constantly being altered.  I just want to send that directory to my dask workers without making a zipfile or egg or whatever.  Currently I use these functions to get the job done:

    def fn_to_targz_string(fn):
        with io.BytesIO() as bt:
            with tarfile.open(fileobj=bt,mode='w:gz') as tf:
                tf.add(fn,arcname=os.path.basename(fn))
            bt.seek(0)
            s=bt.read()
        return s

    def extract_targz_string(s,*args,**kwargs):
        import io,tarfile
        with io.BytesIO() as bt:
            bt.write(s)
            bt.seek(0)
            with tarfile.open(fileobj=bt,mode='r:gz') as tf:
                tf.extractall(*args,**kwargs)

If we used tarfile in this way for Client.upload_file, it wouldn't matter whether the user wanted to upload a file or a directory.  It should just work.

I believe the only necessary changes would be replacing the beginning of `Client._upload_file` with something like `fn_to_targz_string` (shown above) and the beginning of `Worker.upload_file` with something like `extract_targz_string` (shown above).   If you wanted to be complete, you could also add something to the belly of `Worker.upload_file` so that the module would be properly inserted into `names_to_import`.

mrocklin:
Generally adding directory support to upload file seems sensible to me.  

However I think that relying on tar-gz might be problematic just because these aren't universally deployed.  (Dask runs on on windows clusters).  Even though it's quite old, perhaps the `zipfile` module in the stdlib could be used here?  We could also just open up all of the files and send all of their contents in a big nested dict.

jacksonloper:
tarfile has been in the [standard library](https://docs.python.org/2/library/tarfile.html) since python 2.3... do you know if it one of the optional components or is it mandatory?  I kind of assumed that even if the computer doesn't have its own zlib library, the zlib code gets baked into CPython.   I don't have a windows machine so I can't easily check.

mrocklin:
Ah, I didn't realize that.  Yeah, if it's in the standard library then I'm
all for tar files.

mrocklin:
@jacksonloper is uploading directories something that you have time to contribute?

jacksonloper:
Sure, but I can't test windows machines :(.

jacksonloper:
Woops wrong button.  I'll submit a PR.

mrocklin:
That's just fine

[...]

PR #4361, opened 2020-12-14
Add upload_dir function #925

sonicxml:
### Overview:
As discussed in #925, add an `upload_dir` function to `client` that allows users to batch-submit files. This function has two behaviors:
1) If given a directory with many top level files (see below), it will add all files in that directory to the Python imports on the workers (essentially, as if you had called `upload_file` for each file in that directory).
```
src/
  a.py
  b.py
```
would allow you to call `import a` on the workers after calling `upload_dir("src")`.

2) If `remote_path` is specified, the contents of the directory are stored under that path on each worker. This allows for using Python's package import syntax (e.g. `import src.mypkg.subpkg.file`) on the workers.
```
src/
  foo/
  bar/
  baz/
    a.py
    b.py
```
would allow you to call `import src.mypkg.a` after calling, for example, `upload_dir("src/baz", remote_path="src/mypkg")`.

### Implementation:
This function makes a compressed tarfile of the given folder and sends that to each worker. It is built using the same plugin system that `upload_file` now uses, allowing it to handle additional workers later. I tried to re-use the `upload_file` code flow where I could. I also fixed the `save_sys_modules` test util since I don't think it was working as intended.

### Design Decisions:
- The tarfile is stored in memory on the local computer, similar to how `UploadFile` works. This means that there could be issues if the contents of the provided directory take up more space than available memory, but I assumed that in this case users would just be building a custom Docker image with their files in it. 
- As mentioned, I tried to reuse the `upload_file` code path, so `upload_dir` shares the functions in `distributed/worker.py` and `distributed/utils.py`, though I had to make some modifications to handle the additional logic.
- When uploading a directory with package format again, which would force a re-import of the modules, I discovered that `pkgutil.iter_modules` won't search recursively. To force an `import.reload` I had to search `sys.modules` for the directory's package name. This was my first time getting this deep into the Python import system so there might be a better way to do it. I unfortunately couldn't figure out another way that also passed all tests.


sonicxml:
CI failures look unrelated

sonicxml:
@mrocklin is there anyone to tag that would be good to review this?

andrzejnovak:
Things seem to have been a bit quiet here for a while, @sonicxml have you found an alternate way of doing this without modifying dask source directly?

GPUtester:

jrbourbeau:
@sonicxml does the `UploadDirectory` plugin work for you too?

andrzejnovak:
@jrbourbeau is there a way to query paths/environ of the workers?

jrbourbeau:
You can use `client.run` to run any function directory on the workers and forward the results back to the client process. For example, something like (the code below is untested):

```python
def get_env_vars():
    import os
    return list(os.environ.keys())

client.run(get_env_vars)
```

andrzejnovak:
@jrbourbeau Cheers that helped me debug a bit I think. Here's what I am dealing with. Indeed, it doesn't appear the directory gets uploaded to the worker then.
image

jrbourbeau:
Looking at the `UploadDirectory` plugin source, the directory you're uploading is placed in the nanny's `local_directory`

https://github.com/dask/distributed/blob/25881517d101b72944a7352acdcff8bd657bc0b4/distributed/diagnostics/plugin.py#L399-L413

Could you check in `nanny.local_directory` instead? You can retrieve that location using something like

```python
client.run(lambda dask_worker: dask_worker.local_directory, nanny=True)
```

andrzejnovak:
@jrbourbeau alright cool, calling the above with `nanny=True` shows `workflows` uploaded and in the path so that bit works. I assumed these were propagating from the nanny to the worker but guess not. Strangely switching `UploadPlugin(..., nanny=False)` throws `TypeError: setup() got an unexpected keyword argument 'worker'` on the first cell execution (I guess worker is not assigned), passes on the second and uploads the files, but doesn't update path. 

image

oshadura:
@jrbourbeau I tried @andrzejnovak code snippet as well on the same facility with `nanny=True` and I see that workflow directory is getting uploaded everywhere including HTCondor workers but I am keep on getting: `OSError: Timed out trying to connect to tls://172.19.0.5:42245 after 30 s` where `172.19.0.5:42245` is address of nanny.

oshadura:
@jrbourbeau I see that it works out of the box  if we have only 1 k8s worker (we are running it as sidecar), but if there are HTCondor workers then I see timeout.

oshadura:
@jrbourbeau, also actually even I see that directory `workflows` is  uploaded on k8s dask workers, `client.run(get_info)` still not able to import it.

```def get_info():
    import sys, os
    #import workflows.SUEP_coffea
    return os.listdir('dask-worker-space/workflows'), sys.path

client.run(get_info)
{'tls://oksana-2eshadura-40cern-2ech.dask-worker.coffea.casa:8788': (['SUEP_coffea.py',
   '.ipynb_checkpoints',
   '__pycache__',
   '__init__.py'],
  ['/home/cms-jovyan/dask-worker-space/worker-galajhw9',
   '/opt/conda/bin',
   '/opt/conda/lib/python38.zip',
   '/opt/conda/lib/python3.8',
   '/opt/conda/lib/python3.8/lib-dynload',
   '/opt/conda/lib/python3.8/site-packages'])}
```

oshadura:
> @jrbourbeau I see that it works out of the box if we have only 1 k8s worker (we are running it as sidecar), but if there are HTCondor workers then I see timeout.
> 
> Update: @jthiltges confirmed that 172.19 address range is used by the condor workers (in docker containers). It won't be reachable from outside the node where it's running.

@jrbourbeau could you explain us please how it actually suppose to work,  if timeout OSError is because we need to connect to the worker or worker is trying to connect to the scheduler to retrieve the data? It is not clear from message...

IlyaMaksheevQuantori:
Is there any info about that feature? I am also lacking ability to send code with complex path to files.

Danferno:
For anyone stumbling upon this, the `register_worker_plugin` function has been deprecated in favour of `register_plugin`

PR #4362, opened 2020-12-14
Provide `setup.py` option to profile Cython code

jakirkham:
As it can be useful for development to profile the Cythonized code and this requires an argument be provided to the extensions to enable, go ahead and add some logic to `setup.py` to parse this option if provided. If it is not provided, profiling is turned off, which is preferable as this would impact the performance of the compiled code.

cc @quasiben

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,18 +21,26 @@ for r in requires:
     else:
         install_requires.append(r)
 
-try:
-    sys.argv.remove("--with-cython")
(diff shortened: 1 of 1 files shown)


quasiben:
I think the failure of `test_release_failure` is unrelated

quasiben:
With this passing GH CI I'll merge in.  Thanks @jakirkham

PR #4364, opened 2020-12-15
Dask internal inherit config

### LINKED ISSUE #4363: Impossible to override default config with DASK_INTERNAL_INHERIT_CONFIG
**What happened**:

When passing config with `DASK_INTERNAL_INHERIT_CONFIG` it appears it is not possible to override the default values.

I expected the config passed with `DASK_INTERNAL_INHERIT_CONFIG` to be applied.

Generate config
```python
>>> from distributed.utils import serialize_for_cli
>>> serialize_for_cli({"distributed": {"comm": {"default-scheme": "tls"}}})
'eyJkaXN0cmlidXRlZCI6IHsiY29tbSI6IHsiZGVmYXVsdC1zY2hlbWUiOiAidGxzIn19fQ=='
```

Start scheduler with `DASK_INTERNAL_INHERIT_CONFIG`.

```console
$ DASK_INTERNAL_INHERIT_CONFIG="eyJkaXN0cmlidXRlZCI6IHsiY29tbSI6IHsiZGVmYXVsdC1zY2hlbWUiOiAidGxzIn19fQ==" dask-scheduler
```

Connect a client and check the config.

```python
>>> from dask.distributed import Client
>>> import dask.config
>>> client = Client("tcp://localhost:8786")
>>> client.run_on_scheduler(lambda: dask.config.get("distributed.comm.default-scheme"))
'tcp'
```

It appears that existing config is prioritised, even if the existing config is just detected from the defaults.

https://github.com/dask/distributed/blob/f2b09c6b55188f6b633cba5422454b689b0fd7b4/distributed/cli/dask_scheduler.py#L170-L173

**Environment**:

- Dask version: 2020.12.0+6.g1d669480 (master)
- Distributed version: 2020.12.0+7.gc2d87738 (master)
- Python version: 3.8.5
- Operating System: macOS
- Install method (conda, pip, source): Source

jacobtomlinson:
Setting `priority="old"` when updating config with `DASK_INTERNAL_INHERIT_CONFIG` means that default config options can never be overridden.

When used in [`SSHCluster` to package up the local config](https://github.com/dask/distributed/blob/c2d87738828ddad421e4766cc9521b9e2c471857/distributed/deploy/ssh.py#L178-L180) to send to the scheduler and workers any value which has a [default set in `distributed.yaml`](https://github.com/dask/distributed/blob/master/distributed/distributed.yaml) will be returned to the default.

This change resolved my issue, however I'm guessing this option was included for a reason. @madsbk do you know why this was added?

Closes #4363 

---

I also added a test for the serialization as I ended up writing one as part of putting together the MVRE for #4363.

### DIFF
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -170,7 +170,7 @@ def main(
     if "DASK_INTERNAL_INHERIT_CONFIG" in os.environ:
         config = deserialize_for_cli(os.environ["DASK_INTERNAL_INHERIT_CONFIG"])
         # Update the global config given priority to the existing global config
-        dask.config.update(dask.config.global_config, config, priority="old")
+        dask.config.update(dask.config.global_config, config)
 
     if not host and (tls_ca_file or tls_cert or tls_key):
         host = "tls://"
--- a/distributed/cli/dask_scheduler.py
+++ b/distributed/cli/dask_scheduler.py
@@ -169,7 +169,6 @@ def main(
 
     if "DASK_INTERNAL_INHERIT_CONFIG" in os.environ:
         config = deserialize_for_cli(os.environ["DASK_INTERNAL_INHERIT_CONFIG"])
-        # Update the global config given priority to the existing global config
         dask.config.update(dask.config.global_config, config)
 
     if not host and (tls_ca_file or tls_cert or tls_key):
--- a/distributed/cli/dask_spec.py
+++ b/distributed/cli/dask_spec.py
@@ -1,10 +1,13 @@
 import asyncio
 import click
 import json
+import os
 import sys
 import yaml
 
+import dask.config
 from distributed.deploy.spec import run_spec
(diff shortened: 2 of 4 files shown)


jacobtomlinson:
CI failures seem to be unrelated (they are the same on master)

madsbk (commented):
I do not remember exactly why we gave priority to the existing global config. My guess is that we didn't want to overwrite an option given specifically to a worker. I don't know if that is an issue?

jacobtomlinson:
Thanks @madsbk. From my reading of the code I don't think this would be an problem. 

It actually appears that the worker gets some values from the config before this is applied. So it may be that today some of these values are missed.

I wonder if upstreaming this to `dask.config` all together would be the best approach?

madsbk (approved):
> I wonder if upstreaming this to dask.config all together would be the best approach?

I think it is fine to merge this for now.

jacobtomlinson:
Thanks @madsbk

PR #4365, opened 2020-12-15
Refactor `SchedulerState` from `Scheduler`

jakirkham:
Requires PR ( https://github.com/dask/distributed/pull/4343 )

This refactors out the `SchedulerState` as `@cclass` including info about endpoints and tasks. Also includes the `transition_*` functions. This should make it easier to more thoroughly Cythonize these components.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -135,6 +135,7 @@ class Server:
         connection_args=None,
         timeout=None,
         io_loop=None,
+        **kwargs,
     ):
         self.handlers = {
             "identity": self.identity,
@@ -229,6 +230,8 @@ class Server:
         )
 
         self.__stopped = False
+
+        super().__init__(**kwargs)
 
     @property
     def status(self):
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1646,11 +1646,10 @@ class SchedulerState:
         """
         Remove *ts* from the set of processing tasks.
         """
-        workers: dict = cast(dict, self._workers)
         ws: WorkerState = ts._processing_on
         ts._processing_on = None
         w: str = ws._address
-        if w in workers:  # may have been removed
+        if w in self._workers_dv:  # may have been removed
             duration = ws._processing.pop(ts)
             if not ws._processing:
                 self._total_occupancy -= ws._occupancy
@@ -1730,9 +1729,7 @@ class SchedulerState:
 
     def transition_released_waiting(self, key):
         try:
-            tasks: dict = self._tasks
-            workers: dict = cast(dict, self._workers)
-            ts: TaskState = tasks[key]
+            ts: TaskState = self._tasks[key]
             dts: TaskState
             worker_msgs: dict = {}
             client_msgs: dict = {}
@@ -1770,7 +1767,7 @@ class SchedulerState:
             ts._waiters = {dts for dts in ts._dependents if dts._state == "waiting"}
 
             if not ts._waiting_on:
-                if workers:
+                if self._workers_dv:
                     recommendations[key] = "processing"
                 else:
                     self._unrunnable.add(ts)
@@ -1787,9 +1784,7 @@ class SchedulerState:
 
     def transition_no_worker_waiting(self, key):
         try:
-            tasks: dict = self._tasks
-            workers: dict = cast(dict, self._workers)
-            ts: TaskState = tasks[key]
+            ts: TaskState = self._tasks[key]
             dts: TaskState
             worker_msgs: dict = {}
             client_msgs: dict = {}
@@ -1819,7 +1814,7 @@ class SchedulerState:
             ts.state = "waiting"
 
             if not ts._waiting_on:
-                if workers:
+                if self._workers_dv:
                     recommendations[key] = "processing"
                 else:
                     self._unrunnable.add(ts)
@@ -1838,7 +1833,6 @@ class SchedulerState:
         """
         Decide on a worker for task *ts*.  Return a WorkerState.
         """
-        workers: dict = cast(dict, self._workers)
         ws: WorkerState = None
         valid_workers: set = self.valid_workers(ts)
 
@@ -1846,7 +1840,7 @@ class SchedulerState:
             valid_workers is not None
             and not valid_workers
             and not ts._loose_restrictions
-            and workers
+            and self._workers_dv
         ):
             self._unrunnable.add(ts)
             ts.state = "no-worker"
@@ -1855,7 +1849,7 @@ class SchedulerState:
         if ts._dependencies or valid_workers is not None:
             ws = decide_worker(
                 ts,
-                workers.values(),
+                self._workers_dv.values(),
                 valid_workers,
                 partial(self.worker_objective, ts),
             )
@@ -1866,15 +1860,14 @@ class SchedulerState:
             if n_workers < 20:  # smart but linear in small case
                 ws = min(worker_pool.values(), key=operator.attrgetter("occupancy"))
             else:  # dumb but fast in large case
-                n_tasks: Py_ssize_t = self._n_tasks
-                ws = worker_pool.values()[n_tasks % n_workers]
+                ws = worker_pool.values()[self._n_tasks % n_workers]
 
         if self._validate:
             assert ws is None or isinstance(ws, WorkerState), (
                 type(ws),
                 ws,
             )
-            assert ws._address in workers
+            assert ws._address in self._workers_dv
 
         return ws
 
@@ -1897,8 +1890,7 @@ class SchedulerState:
 
     def transition_waiting_processing(self, key):
         try:
-            tasks: dict = self._tasks
-            ts: TaskState = tasks[key]
+            ts: TaskState = self._tasks[key]
             dts: TaskState
             worker_msgs: dict = {}
             client_msgs: dict = {}
@@ -1944,10 +1936,8 @@ class SchedulerState:
 
     def transition_waiting_memory(self, key, nbytes=None, worker=None, **kwargs):
         try:
-            workers: dict = cast(dict, self._workers)
-            ws: WorkerState = workers[worker]
-            tasks: dict = self._tasks
-            ts: TaskState = tasks[key]
+            ws: WorkerState = self._workers_dv[worker]
+            ts: TaskState = self._tasks[key]
             worker_msgs: dict = {}
             client_msgs: dict = {}
 
@@ -1997,8 +1987,7 @@ class SchedulerState:
         worker_msgs: dict = {}
         client_msgs: dict = {}
         try:
-            tasks: dict = self._tasks
-            ts: TaskState = tasks[key]
+            ts: TaskState = self._tasks[key]
             assert worker
             assert isinstance(worker, str)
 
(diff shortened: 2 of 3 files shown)


jakirkham:
cc @mrocklin @quasiben

jakirkham:
Should add that after doing this, I can see a path to typing all attributes of the `Scheduler` and using them throughout. This would speed up attribute access and would also give Cython the opportunity to use targeted Python C APIs with those attributes (assuming they have a builtin type more specific than `object`).

mrocklin:
FYI, I'm slammed today and tomorrow.  I'm not going to be very responsive
this week.

jakirkham:
No worries. Just wanted to make sure you are aware of this. Mostly just interested in your general high-level thoughts as opposed to a detailed review at this stage.

jakirkham:
Some of the transition and supporting methods take `**kwargs`, but don't seem to do anything with them. As this hinders optimizing these methods, I've just dropped them. Please let me know if that is issue and they are needed by something else somewhere so we can devise alternative strategies (if needed).

jakirkham:
Hmm...seems those `**kwargs` exist to just ignore some arguments (see traceback below)? Confused as to why they are showing up in the first place.

```python
distributed.core - ERROR - transition_processing_memory() got an unexpected keyword argument 'status'
  File "/Users/jkirkham/Developer/distributed/distributed/core.py", line 592, in handle_stream
    handler(**merge(extra, msg))
  File "distributed/scheduler.py", line 4707, in distributed.scheduler.Scheduler.handle_task_finished
  File "distributed/scheduler.py", line 4115, in distributed.scheduler.Scheduler.stimulus_task_finished
  File "distributed/scheduler.py", line 6079, in distributed.scheduler.Scheduler.transition
  File "distributed/scheduler.py", line 6015, in distributed.scheduler.Scheduler.transition
  File "distributed/scheduler.py", line 2022, in distributed.scheduler.SchedulerState.transition_processing_memory
TypeError: transition_processing_memory() got an unexpected keyword argument 'status'
```

jakirkham:
After some toying around, I think these are coming from here for finished tasks:

https://github.com/dask/distributed/blob/c2d87738828ddad421e4766cc9521b9e2c471857/distributed/worker.py#L1920-L1929

Also from errored tasks this appears to come from here:

https://github.com/dask/distributed/blob/c2d87738828ddad421e4766cc9521b9e2c471857/distributed/worker.py#L1931-L1938

Perhaps the `stimulus_` functions should be grabbing these extra arguments and handling them. For now am just grabbing them in `transition` until we come to some agreement here on what to do.

jakirkham:
Here's the Scheduler call graph with these changes so far. Some functions (like `_add_to_memory`) are being handled with a virtual table, which we don't need as they won't be inherited. So we might break them out into separate functions to make them independent of the class (and thus the virtual table).

Scheduler call graph

jakirkham:
Yeah messages including unused arguments and extra arguments being passed to `transition_` functions seems like the main issue we need to address here (details in this comment https://github.com/dask/distributed/pull/4365#issuecomment-747107706 and above it). Once we figure that out it should be more straightforward to optimize these.

jakirkham:
There's other odd things like `nbytes` being set to `None` (`-1` would be better for typing). `worker` being optional where a function absolutely needs it. And so on. Though these are second order problems we can fix if we fix the first order ones (above).

jakirkham:
This should now be ready for review. May tweak a few things still, but am interested in getting feedback on this at this point.

quasiben:
@madsbk would you have time to review (or review part of) this PR ?

distributed/scheduler.py
```diff
@@ -1576,3919 +1566,4196 @@ class Scheduler(ServerNode):
...
+        except Exception as e:
+            logger.exception(e)
+            if LOG_PDB:
+                import pdb
 
-    ##################
-    # Administration #
-    ##################
+                pdb.set_trace()
+            raise
```
madsbk, line 1786: I suggest we implement all the exception logging using a context manager as described in https://stackoverflow.com/a/28158006

madsbk (approved):
Overall I like the refactoring; encapsulate the state of the scheduler in a different class makes a lot of sense. Actually, I think we should go even further and encapsulate all states of the scheduler and tasks with clearly defined invariants but that is future work :)

I have read through all of the changes and as far as I can see, it looks good but because of the amount of changes I am not confident that it is bug free :)

As a side note I agree, we should get rid of all the unused `**kwargs` in a future PR. It makes a harder to follow variables through the code.

jakirkham:
Thanks for looking through this Mads! 😄

> Overall I like the refactoring; encapsulate the state of the scheduler in a different class makes a lot of sense. Actually, I think we should go even further and encapsulate all states of the scheduler and tasks with clearly defined invariants but that is future work :)

Thanks Mads! 😄 Yeah for sure. I think this opens a lot of doors for us in terms of future optimizations. Just tried to find the right balance where we do enough to get some benefit without pushing everything into one PR.

> I have read through all of the changes and as far as I can see, it looks good but because of the amount of changes I am not confident that it is bug free :)

Completely understand that. Have really been relying on the CI and tests to catch any issues. Also have been doing some profiling locally, which has helped as well. Plus regex has been my friend (most changes represent a common pattern).

> As a side note I agree, we should get rid of all the unused `**kwargs` in a future PR. It makes a harder to follow variables through the code.

Agreed. I tried to get rid of them in this PR, but it wound up being a little too complicated and touching too much other code. Also as we are still calling the `transition_*` functions with their Python APIs, we don't get much benefit yet of adding C APIs. Though this may change with future PRs.

distributed/core.py
```diff
@@ -230,6 +231,8 @@ def set_thread_ident():
 
         self.__stopped = False
 
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
jakirkham: Yeah that's fair. It's a little confusing.

The call to `super` is needed to call the parent constructors of `Scheduler`. These are `ServerNode` and `SchedulerState`. We also want to affect when these get called relative to other things in this constructor (as this constructor needs some of the attributes of the parent classes to be setup in later steps). 

As `Scheduler` inherits from two base classes, we need to make sure that arguments for those constructors get passed through and picked up by the right parent class. There are some more details in [this SO answer]( https://stackoverflow.com/a/34885285 ) that may help clarify this further.

distributed/scheduler.py
jakirkham: Yeah I agree that something like that makes sense. Alternatively we might want to push this into `transition` where all of these methods are called. Held off on that for now as I didn't want to disrupt the code too much, but agree this is something we should follow up on.

mrocklin:
>> Actually, I think we should go even further and encapsulate all states of the scheduler and tasks with clearly defined invariants but that is future work :)

> Thanks Mads! smile Yeah for sure. I think this opens a lot of doors for us in terms of future optimizations. Just tried to find the right balance where we do enough to get some benefit without pushing everything into one PR.

Yeah, I have the same question.  What do we include in SchedulerState and what do we include in Scheduler?  I'm totally fine doing this in the future, but some methods, like `new_task` or `transition` or `remove_worker` I expected to be primarily state-focused while methods like `stimulus_task_finished` or `handle_worker` I would expect to remain in the Scheduler.

jakirkham:
Yeah that makes sense. Generally have been thinking anything that relates the management of the task graph or transformations of it seem like good candidates for `SchedulerState`. Methods involved in communication or general upkeep/maintenance of the `Scheduler` seem better there. IOW I think we have the same idea in mind that's just broadly what I've been thinking. Agree that these things can be tackled in future work. Am mainly trying to provide a framework and initial first step here to allow us to dive into that work (possibly from multiple directions 😉).

jakirkham:
Refactored out a few more minor utility functions used by the `transition_*`s. AFAICT that's all of them. Please let us know if there's anything else needed here 🙂

jakirkham:
FWIW here's the profile using [Ben's shuffle benchmark]( https://github.com/quasiben/dask-scheduler-performance ) with these changes.

<details>

prof_476 pstat dot

</details>

lr4d:
Just wondering, is it necessary that `Scheduler` inherits from `SchedulerState`? From a logical level, it seems like it would make more sense if `Scheduler` had one or mulitple attribute/property provided by `SchedulerState`

jakirkham:
After having done the work it's actually not that clear to me that it would be that doable. `SchedulerState` and `Scheduler` overlap on a fair number of attributes that they both need. Subclassing makes more sense as they both access these as opposed to letting the `Scheduler` muck with `SchedulerState`'s internal state, which may actually make it harder to refactor out more functionality later. As the goal was to turn some of the `Scheduler`'s methods into C functions using Cython, this also was easier to do with minimal duplication or disruption via subclassing. Also some of the methods like `worker_objective` seem to be designed to be overridden in subclasses, which we continue to allow with this structure. Maybe over time as the boundaries between these two become more clearly defined we can revisit, but at least today I'm not sure this is doable.

distributed/scheduler.py
```diff
@@ -5543,9 +5858,14 @@ def transition(self, key, finish, *args, **kwargs):
                     "Impossible transition from %r to %r" % (start, finish)
                 )
 
+            for worker, msg in worker_msgs.items():
+                self.worker_send(worker, msg)
+            for client, msg in client_msgs.items():
+                self.client_send(client, msg)
```
mrocklin, line 5864: It might not help any, but any thoughts on pulling these out of the `transition` function into `transitions` ?

mrocklin: It looks like they're a decent chunk of the current time spent

image

jakirkham: It's an interesting suggestion and agree it's worth trying. That said, I feel like this PR has gotten quite long and am worried that has slowed down the reviewing of it. Instead would propose we focus on getting this in. That would make it easier to make smaller more focused PRs on moving communication further out and consolidating messages further as well as allowing us to more effectively Cythonize any remaining functions we still see popping up in the profiles. Thoughts? 🙂

(Should add if we want to do this after the patch release tomorrow that also makes sense to me)

mrocklin:
OK, we're good to go here.  I expect that over time we'll move more things over, but I agree with @jakirkham that what is here now is a good firm base.

jakirkham:
Thanks Matt! 😄

jakirkham:
> Yeah, I have the same question. What do we include in SchedulerState and what do we include in Scheduler? I'm totally fine doing this in the future, but some methods, like `new_task` or `transition` or `remove_worker` I expected to be primarily state-focused while methods like `stimulus_task_finished` or `handle_worker` I would expect to remain in the Scheduler.

FWIW moving `new_task` over in PR ( https://github.com/dask/distributed/pull/4527 ). This was pretty easy to do.

I think `transition` should eventually be doable. Would be good to more strongly type the arguments to this function first though.

Looked at `remove_worker` and some pieces may be amenable to moving, but it is also an `async` method with some IO bits. So we might need to split it into a helper method that lives in `SchedulerState` and leave the rest in `Scheduler`. Will need to think about this more.

distributed/scheduler.py
jakirkham: This was mostly done in PR ( https://github.com/dask/distributed/pull/4451 ). Handling the remaining ones in PR ( https://github.com/dask/distributed/pull/4526 )

PR #4370, opened 2020-12-17
Grouped task stream

### LINKED ISSUE #4346: Improve dask dashboard display nested tasks execution
# Description
`dask`'s dashboard `Task Stream` pane (in the `status` page) seems to map horizontal sections within the pane to unique worker thread ids. 

On the other hand, when a `client.submit` call is made **within** a `submit`-ed task ("nested task" situation), additional worker threads must be created through `secede` calls, which in turns creates additional horizontal section.
However,
- the two (outer and inner) task status are not independent: the `outer` task will complete only if the `inner` task has already completed.
- as far as I can tell, the use case for `secede` is only to make sure a passive, waiting worker thread does not occupy the thread pool while it waits for other tasks. Therefore, even though an additional worker thread is created at each `secede` calls, the number of **active** threads remains unchanged throughout the program execution.

In my opinion both of those points are not made visually clear by the `dask` dashboard. If `n` working threads exist initially in the cluster and  `m` `secede` calls are made, `n+m` horizontal sections (suggesting `n+m` active threads) will displayed, while the arrangement does not take into account the hierarchical structure of the program (nested `submit` calls)

See screenshot:
nested_submit_dask_dashboard

<details>
<summary> That was obtained from the following reproducer </summary>

```python
import logging
import time

import numpy as np

from distributed import (
    LocalCluster,
    Client,
    get_client,
    secede,
    rejoin,
    get_worker,
    as_completed,
)


def sum_and_sleep(array):
    time.sleep(0.5)
    return np.sum(array)


def outer_function(array, i):
    print(f"running outer task {i}")
    client = get_client()

    futures = client.map(
        sum_and_sleep, [array[i + j :] for j in range(NUM_INNER_TASKS)]
    )
    secede()

    results = [f.result() for f in futures]
    return sum(results)


if __name__ == "__main__":
    NUM_OUTER_TASKS = 10
    NUM_INNER_TASKS = 10

    my_arrays = [np.ones(100000) for _ in range(NUM_OUTER_TASKS)]

    cluster = LocalCluster(
        n_workers=4, threads_per_worker=1, silence_logs=logging.WARNING
    )
    client = Client(cluster)

    list_of_res = []
    results = client.map(outer_function, my_arrays, range(len(my_arrays)))
    print(client.gather(results))
```
</details>  

Although 4 worker threads are active at the same time, many more line exists (13 in this situation).

# Real life use case

Nested scatter calls typically happen in `joblib`, in the context of nested `Parallel` calls.

# Next steps proposal

Would you be opened to improve this situation? I'm thinking about a `config` entry allowing toggling between mapping horizontal sections to thread is, and mapping horizontal sections to  {process id +/hostnames}, although there may surely be better display options. I can make a PR, I'm mostly waiting for some kind for prior consent by core developers.  

cc'ing @ogrisel.

pierreglaser:
Fixes #4346 

Changes the tasks stream displaying logic:

- All bands related to the same workers are now placed one after the other
- The total band for a worker is computed statically at the beginning: if new threads are created from `secede` calls,
  tasks completed by such threads will all be allocated to the upper-most sub-band of their worker
- Added a slight hovering effect to know which sub-bands are associated with each worker

Screencasts to come, but @mrocklin or anyone else, feel free to check it out locally.

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1348,7 +1348,6 @@ class TaskStream(DashboardComponent):
                     this_worker_positions['nthreads']
                 ) / 2
                 this_worker_positions['threads_y_coord'][thread] = y
-                print(worker, thread, y)
 
     def _group_tasks_vetrically_by_workers(self, rects):
         # Modifies the 'y' field of the completed tasks data structure so that
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -1266,6 +1266,9 @@ class TaskStream(DashboardComponent):
         boxes = {}
 
         def highlight_worker_band(event):
+            if event.x is None or event.y is None:
+                return
+
             # highlights the horizontal band corresponding to the tasks
             # completed by all threads of a given worker
             nonlocal last_box
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -5,6 +5,8 @@ from numbers import Number
 import operator
 import os
 
+from bokeh.models.annotations import BoxAnnotation
+from bokeh.events import MouseMove, MouseLeave
 from bokeh.layouts import column, row
 from bokeh.models import (
     ColumnDataSource,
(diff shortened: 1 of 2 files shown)


distributed/dashboard/components/scheduler.py
```diff
@@ -1255,6 +1257,113 @@ def __init__(self, scheduler, n_rectangles=1000, clear_interval="20s", **kwargs)
...
+                # Any unexpected additional thread is redirected to the
+                # uppermost area of the associated worker.
+                y = this_worker_positions['y_min'] + min(
+                    len(this_worker_positions['threads_y_coord']),
+                    this_worker_positions['nthreads']
+                ) / 2
+                this_worker_positions['threads_y_coord'][thread] = y
+                print(worker, thread, y)
+
+    def _group_tasks_vetrically_by_workers(self, rects):
```
pierreglaser, line 1353: (typo)

mrocklin:
I look forward to these screencasts with great excitement :)

distributed/scheduler.py
```diff
@@ -403,8 +405,9 @@ def __init__(
         self._services = services or {}
         self._versions = versions or {}
         self._nanny = nanny
-
+        self._uuid = uuid4().hex
         self._hash = hash(address)
+
```
pierreglaser, line 410: (rebase leftovers)

mrocklin:
I couldn't wait

image

```python
import dask.array as da

client = Client()

x = da.random.random((100, 2000, 2000))
workers = list(client.scheduler_info()["workers"])
x.sum().compute()
_ = client.map(f, range(30), workers=workers[0])
```

mrocklin:
One style comment is that it might make sense to reduce the size of the separation between worker groups.  A smaller/more subtle difference might look nicer.  This may become more important especially in the case of many single-threaded workers.  

However, this might work less well if we want to include long running tasks or other information in the gaps.  

I haven't thought too much about this yet.  I'm just throwing out ideas.

distributed/dashboard/components/scheduler.py
```diff
@@ -1255,6 +1257,113 @@ def __init__(self, scheduler, n_rectangles=1000, clear_interval="20s", **kwargs)
...
+                # first computed task, this function allocates the first
+                # available sub-band of the thread's worker band, from bottom to top.
+                # Any unexpected additional thread is redirected to the
+                # uppermost area of the associated worker.
+                y = this_worker_positions['y_min'] + min(
+                    len(this_worker_positions['threads_y_coord']),
+                    this_worker_positions['nthreads']
+                ) / 2
+                this_worker_positions['threads_y_coord'][thread] = y
+                print(worker, thread, y)
```
mrocklin, line 1351:
```suggestion
```

distributed/dashboard/components/scheduler.py
```diff
@@ -1255,6 +1257,113 @@ def __init__(self, scheduler, n_rectangles=1000, clear_interval="20s", **kwargs)
         # Required for update callback
         self.task_stream_index = [0]
 
+        self._worker_position_info = {}
+        self._compute_workers_y_range()
+        self._add_worker_band_hover_effect(self.root)
+
+    def _add_worker_band_hover_effect(self, root):
```
mrocklin, line 1264: Ooh, this is fun.

pierreglaser: Suddenly I know `bokeh` very well :)

pierreglaser:
> One style comment is that it might make sense to reduce the size of the separation between worker groups. A smaller/more subtle difference might look nicer. This may become more important especially in the case of many single-threaded workers.

The separation is done in case of threads created after a `secede` call : i allocate one extra task bands for newly created threads. It maybe is not a good default, but it happens in nested parallel calls in `joblib`.

Previously, when an extra thread is created after a `secede` call, its task was displayed in a new band, the uppormost one. But now, we are trying to group things up, so this task should be displayed in a band associated to its worker. Since I don't want to move all rects all the time for each new thread created after `secede`, I pre-allocate one band for each worker for all extra threads (threads that come after `WorkerState.nthreads` threads, thus surely from `secede` calls - although there may be other cases where extra threads are created), hence the big separation.

In general, there is a trade-off between how much logic we add / how much data updates we send to bokeh, and how adaptable the dashboard can be. If we want to make the dashboard more dynamic (thus, dont create band statically, but dynamically, we have to send full history updates to re-draw the whole task diagram). If I'm too succint, which makes it unclear, please ask me to expand.

distributed/dashboard/components/scheduler.py
mrocklin: It's a fun tool to have in your toolkit

pierreglaser:
cc @ogrisel if you want to test it out. It's probably a bit rough around the edges, but probably testable. Still some spurious errors sometimes that I have to check.

pierreglaser:
also @mrocklin, let me know what you think about the "worker" hover system, where all bands associated to a worker are highlighted when the user hovers the bands in the task stream.

mrocklin:
> also @mrocklin, let me know what you think about the "worker" hover system, where all bands associated to a worker are highlighted when the user hovers the bands in the task stream.

That's what I was referring to when I said "Ooh, this is fun" :)

Stylistically I suspect that we could reduce the visual weight of the hover (make it more transparent) and still have the same highlighting effect.

pierreglaser:
> Stylistically I suspect that we could reduce the visual weight of the hover (make it more transparent) and still have the same highlighting effect.

Sure, I've not yet tuned any style parameter yet.

ogrisel:
I gave this pull request a try with this script: https://gist.github.com/ogrisel/4d27a3fa19fcbaf2d069f617e059bce8 (just change the parameter to `n_samples` to `int(1e3)` instead of `int(1e3)` to get shorter tasks) as it creates many new threads per worker because of `secede` / `rejoin` calls. After a while, all the tasks in the stream are displayed in the lane dedicated to new threads.

Maybe it would be better to allocate just one lane per original thread per worker. And then when displaying tasks that complete with a thread id that is not one of the original thread ids, just display them either of the lanes of the worker of that task using either a round robin strategy or a some kind of greedy strategy that tries to minimize overlap with existing rectangle.

This way for work loads that do not use secede / rejoin, we get the same deterministic behavior as previously and for secede / rejoin heavy workloads, one would get a display that approximately reflect the OS thread scheduling behavior inside each worker.

ogrisel:
Also, nitpick: +1 for removing the horizontal gray lines that are unrelated to worker boundaries.

mrocklin:
Another option for displaying seceded threads would be to overlay them over the entire vertical section for a worker, but with more transparency.  Any task that had seceded would cover the entire vertical space, and then the next unknown thread to arrive would take the place of that thread in the vertical ordering.  

This visual effect maybe aligns better with our understanding of what is happening it is a thread in the background both computationally and visually.  We would need to make it very transparent in order to not make things feel too cluttered.  It might look nice though?

ogrisel:
That's also an option. It might be a bit surprising to the users though. Suppose that you have 4 workers with 2 threads each: at the beginning of the processing they will see 8 streams of opaque, mostly non-overlapping tasks that progressively transition to 4 (wider) lanes of transparent and overlapping tasks. But maybe this is fine.

mrocklin:
> That's also an option. It might be a bit surprising to the users though. Suppose that you have 4 workers with 2 threads each: at the beginning of the processing they will see 8 streams of opaque, mostly non-overlapping tasks that progressively transition to 4 (wider) lanes of transparent and overlapping tasks. But maybe this is fine.

I think that we would want to play with opacity a bit here.  

Also, the surprise reflects what is actually going on in this case.  The current experience of suddenly seeing a lot of very long bars completely disconnected in the task stream is also surprising, appropriately so I think.

I'm not good at this, but this seems like the kind of thing where we might want to visually design something before coding it up (unless of course coding it up is faster for us (which may be))

mrocklin:
@jacobtomlinson should you have spare time, reviewing this might be interesting.

pierreglaser:
I currently have some time to work on this - I gave this topic one more thought, and re-read the discussion. My current opinion is that given the current layout of the task stream, which is optimized for displaying "unitary/atomic" tasks, there is no way of displaying "composite/hierarchical" tasks (such as tasks including a  `secede` call) that will correctly represent their hierarchical nature.  I am unsure whether displaying such tasks using a transparent hover effect will both scale well (see https://github.com/dask/distributed/pull/4370#issuecomment-747535325), and provide more information than confusion to the user.

I recently thought of a solution that to my knowledge has not yet been discussed. This solution includes two main changes:
- make the task stream displaying logic simply discard tasks that include a `secede` call
- assign to each thread joining a dask worker executor a unique `effective_subworker_id` (better naming yet to come) that would always be within `range(0, executor._max_workers)`.

Thanks to these two changes, the task stream should be able to display all work done by one executor within `executor._max_workers` horizontal bars, while not having to struggle with displaying seceding tasks. We could possibly include a warning message in the case where a `secede` event happens, indicating that tasks including a `secede` call are not displayed in the task stream.

cc @mrocklin @ogrisel

distributed/scheduler.py
ogrisel, line 410:
```suggestion
```

pierreglaser:
I would support closing this in favour of #5757

PR #4372, opened 2020-12-17
Upstream config serialization and inheritance

jacobtomlinson:
Counterpart to dask/dask#6987.

Upstreaming config serialization to `dask.config`. This PR removes all code and replaces them with imports for backward compatibility.


jacobtomlinson:
dask/dask#6987 is now merged. Hopefully that means CI against master will pass.

distributed/deploy/ssh.py
```diff
@@ -95,13 +95,13 @@ async def start(self):
         result = await self.connection.run("uname")
         if result.exit_status == 0:
             set_env = 'env DASK_INTERNAL_INHERIT_CONFIG="{}"'.format(
-                serialize_for_cli(dask.config.global_config)
+                serialize(dask.config.global_config)
```
jrbourbeau, line 98: Nitpick: We might consider using the full `dask.config.serialize` here instead of just `serialize` in order to not get confused with `distributed.protocol.serialize`. At least for me personally, when I see `serialize` in `distributed` code I immediately think about `distributed.protocol.serialize`

jrbourbeau (commented):
Thanks @jacobtomlinson! Generally this looks good. We'll also need to bump the minimum version of Dask to make sure we pull in the changes from https://github.com/dask/dask/pull/6987

jacobtomlinson:
Thanks @jrbourbeau! I've updated the import.

Agree about bumping the minimum. I assumed we generally tightly couple this dependency each release anyway. Given that dask/dask#6987 is so far unreleased what should we change?

jrbourbeau:
> I assumed we generally tightly couple this dependency each release anyway.

Often we can release `distributed` without bumping the minimum `dask` version. For example, before all the recent HighLevelGraph work, the minimum `dask` version was 2.9.0 (release December 2019).

> Given that dask/dask#6987 is so far unreleased what should we change?

Previously we've bumped our dask version to point to the latest development branch and added a note that we should update to use the latest `dask` release before releasing `distributed`

```
# requirements.txt

# TODO: Update our dask version requirement to
# the latest dask release before releasing distributed
dask @ git+https://github.com/dask/dask.git@master
```

jrbourbeau (approved):
Thanks @jacobtomlinson!

Issue #4359, opened 2020-12-14
Disabling logging in stealing


jakirkham:
Through profiling of the Scheduler, there are 2 parts of stealing that stick out. First is a generator expression that is being fixed in PR ( https://github.com/dask/distributed/pull/4358 ). Second is this logging line (below). Am wondering given this logging is pretty granular (logging each key to steal), whether it would make sense to disable this logging with an option to re-enable. Thoughts? 🙂

https://github.com/dask/distributed/blob/07970728386788eace213487b54562b3af077b84/distributed/stealing.py#L88

mrocklin:
Yes, that makes sense I think.  Most of the other log_event events in
scheduler.py occur at the worker level (like every time a worker comes or
goes), or the client interaction level (like every time the client submits
a graph).  I think that it is sensible to avoid per-task operations like
this.

jakirkham:
Should we just drop it entirely then?

jakirkham:
Proposing just dropping it in PR ( https://github.com/dask/distributed/pull/4375 )

jakirkham:
Fixing another extra `log` line that we missed before in PR ( https://github.com/dask/distributed/pull/4609 )

Issue #4374, opened 2020-12-17
BlockwiseIO is not always msgpack serializable 


fjetter:
Multiple TypeErrors are raised during (de-)serialization on master CI

```
TypeError: can not serialize 'CreateArraySubgraph' object
```

E.g. https://travis-ci.org/github/dask/distributed/jobs/750189428

mrocklin:
cc @ian-r-rose @rjzamora

ian-r-rose:
It looks the new `BlockwiseIO` functionality is not fully msgpack-serializable (on both the dataframe and array side).

Two MWEs, which don't optimize the high-level-graph in order to delay materialization:
```python
import dask.dataframe
import distributed

client = distributed.Client()
ddf = dask.dataframe.read_parquet("test.parquet")
ddf.compute(optimize_graph=False)
```

```python
import dask.array
import distributed

client = distributed.Client()
arr = dask.array.ones((1000, 1000), chunks=(100,100))
arr.compute(optimize_graph=False)
```

We need to (a) make sure that these mappings can be serialized, and (b) make sure that is tested in core to catch it sooner in the future.

rjzamora:
Interesting - I had not tested any distributed code with `optimize_graph=False`.  Thanks for pointing this out @ian-r-rose !

rjzamora:
Sorry about this!

As Ian already mentioned, none of the `BlockwiseIO` functionality is serializable.  This is because the `__dask_distributed_pack__`/`__dask_distributed_unpack__` definitions do not handle the dictionary of IO dependencies (`io_deps`).  I am trying to figure out if it makes more more sense to require subgraphs (like `CSVSubgraph`) to implement their own serialization, or if `BlockwiseIO` should just materialize the IO subgraph to perform serialization.  The latter might be better, because it would be agnostic to the underlying IO subgraph.

I will try to think through a solution tonight, but I am also curious to know if @madsbk has any ideas :)

ian-r-rose:
Thanks for looking into this @rjzamora. I suspect it would be easier for downstream implementers if it was handled closer to `BlockwiseIO`, rather than needing to pass packing/unpacking info down the chain. The good news is that subgraphs like `CSVSubgraph` are by design not very stateful, so rehydrating them on command is probably not too difficult.

I'd be interested to hear what you think best practices are for serializing functions (as I did with `ones`/`zeros` in `BlockwiseCreateArray`). `distributed.worker.dumps_function`?

[...]

PR #4375, opened 2020-12-18
Fix compilation warnings, `decide_worker` now a C func, stealing improvements

jakirkham:
Fixes https://github.com/dask/distributed/issues/4359

* Use `cython.compiled` to skip pure Python fallback functions in Cython case (fixes unused symbol warnings from C compiler)
* Move `decide_worker` function into C (removes overhead of Python API, which is unneeded)
* Some optimization to stealing math and other misc. stealing improvements

Note: There are still some warnings that predate this PR and are not handled currently. See issue ( https://github.com/cython/cython/issues/3474 ) for details.

### DIFF
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -122,6 +122,7 @@ class WorkStealing(SchedulerPlugin):
         split = ts.prefix.name
         if split in fast_tasks:
             return None, None
+
         ws = ts.processing_on
         compute_time = ws.processing[ts]
         if compute_time < 0.005:  # 5ms, just give up
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -84,7 +84,6 @@ class WorkStealing(SchedulerPlugin):
         ws = ts.processing_on
         worker = ws.address
         cost_multiplier, level = self.steal_time_ratio(ts)
-        self.log(("add-stealable", ts.key, worker, level))
         if cost_multiplier is not None:
             self.stealable_all[level].add(ts)
             self.stealable[worker][level].add(ts)
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -134,7 +134,9 @@ class WorkStealing(SchedulerPlugin):
             return None, None
 
         level = int(round(log2(cost_multiplier) + 6))
-        level = max(1, level)
+        if level < 1:
+            level = 1
+
         return cost_multiplier, level
 
     def move_task_request(self, ts, victim, thief):
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -119,9 +119,6 @@ class WorkStealing(SchedulerPlugin):
         if not ts.dependencies:  # no dependencies fast path
             return 0, 0
 
-        nbytes = ts.get_nbytes_deps()
-
-        transfer_time = nbytes / self.scheduler.bandwidth + LATENCY
         split = ts.prefix.name
         if split in fast_tasks:
             return None, None
@@ -129,6 +126,9 @@ class WorkStealing(SchedulerPlugin):
         compute_time = ws.processing[ts]
         if compute_time < 0.005:  # 5ms, just give up
             return None, None
+
+        nbytes = ts.get_nbytes_deps()
+        transfer_time = nbytes / self.scheduler.bandwidth + LATENCY
         cost_multiplier = transfer_time / compute_time
         if cost_multiplier > 100:
             return None, None
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -1,6 +1,6 @@
 from collections import defaultdict, deque
 import logging
-from math import log
+from math import log2
 from time import time
 
 from tornado.ioloop import PeriodicCallback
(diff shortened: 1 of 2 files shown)


jakirkham:
@quasiben, this should fix some of the compile warnings we were seeing earlier 🙂

jakirkham:
CI failures seem to be the same as those in issue ( https://github.com/dask/distributed/issues/4374 ).

distributed/stealing.py
```diff
@@ -134,8 +133,10 @@ def steal_time_ratio(self, ts):
         if cost_multiplier > 100:
             return None, None
 
-        level = int(round(log(cost_multiplier) / log_2 + 6, 0))
-        level = max(1, level)
+        level = int(round(log2(cost_multiplier) + 6))
```
jakirkham, line 136: `log2` was added in Python 3.3+ and is faster than hand rolling by 2x

```python
In [3]: %timeit log(3) / log(2)
401 ns ± 5.33 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit log(3, 2)
234 ns ± 2.72 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

`round` already rounds to `0` digits after the decimal place. Using the default is 2x faster than specifying one

```python

In [7]: %timeit round(1.5, 0)
393 ns ± 2.58 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [8]: %timeit round(1.5)
195 ns ± 1.77 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

distributed/stealing.py
```diff
@@ -134,8 +133,10 @@ def steal_time_ratio(self, ts):
...
+        if level < 1:
+            level = 1
```
jakirkham, line 138: Using `if` is roughly an order of magnitude faster than `max`.

```python
In [1]: %timeit max(1, 3)
222 ns ± 1.66 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [2]: %timeit max(1, 0)
222 ns ± 1.41 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %%timeit
   ...: L = 3
   ...: if L < 1:
   ...:     L = 1
   ...: 
25.5 ns ± 0.165 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %%timeit
   ...: L = 0
   ...: if L < 1:
   ...:     L = 1
   ...: 
29.7 ns ± 0.231 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

distributed/stealing.py
```diff
@@ -120,22 +119,24 @@ def steal_time_ratio(self, ts):
         if not ts.dependencies:  # no dependencies fast path
             return 0, 0
 
-        nbytes = ts.get_nbytes_deps()
-
-        transfer_time = nbytes / self.scheduler.bandwidth + LATENCY
```
jakirkham, line 125: We don't even use these until the 3rd `if`. So just move them later. We could already have `return`ed before needing these.

jakirkham:
Understand if this doesn't get looked at for a while. Just making sure you are aware 😉

jakirkham:
Restarted CI now that issue ( https://github.com/dask/distributed/issues/4374 ) is fixed

jakirkham:
Planning to merge tomorrow if no comments.

distributed/stealing.py
mrocklin: No objection here, but I do want us to look at the full picture rather than just micro-optimizing.  200ns may not matter in the grand scheme.  At some point it might make more sense to keep code around that is nanoseconds slower because it is more readable.  

I'm not making a judgement in this case (I haven't looked at this much).   I just want to avoid us going overboard here.

jakirkham: Mainly came here since `put_key_in_stealable` takes as much runtime as `transition_waiting_processing` and `transition_processing_memory`. With the latter 2 we have already put a lot of time into optimizing them, but this part of the code has largely been untouched. So am trying to tackle some of the low hanging fruit here.

jakirkham: Should add that collectively these changes more than halve the time spent in `put_key_in_stealable`.

mrocklin: Glad to hear it.  My comment was just general.  I have no particular concern with this PR.

jakirkham: Understood. Just wanted to provide context :)

quasiben:
Thanks @jakirkham -- looks good

distributed/stealing.py
```diff
@@ -85,7 +84,6 @@ def put_key_in_stealable(self, ts):
         ws = ts.processing_on
         worker = ws.address
         cost_multiplier, level = self.steal_time_ratio(ts)
-        self.log(("add-stealable", ts.key, worker, level))
```
jakirkham, line 88: Doing the same thing for `remove_key_from_stealable` ( https://github.com/dask/distributed/pull/4609 )

PR #4396, opened 2020-12-31
Support Websocket communication protocols

marcosmoyano:
* Add `ws` backend
* Add `wss` backend

### DIFF
--- a/distributed/comm/ws.py
+++ b/distributed/comm/ws.py
@@ -1,0 +1,372 @@
+import asyncio
+import logging
+import struct
+import weakref
+
+from typing import Callable
+from ssl import SSLError
+
+from tornado.iostream import StreamClosedError
+from tornado.httpserver import HTTPServer
+from tornado.httpclient import HTTPRequest
+from tornado import web
+from tornado.websocket import WebSocketClosedError, websocket_connect, WebSocketHandler
+
+from ..utils import ensure_bytes, nbytes
+
+from .registry import backends
+from .addressing import parse_host_port, unparse_host_port
+from .core import Comm, CommClosedError, Listener, Connector, FatalCommClosedError
+from .tcp import BaseTCPBackend, _expect_tls_context, convert_stream_closed_error
+from .utils import ensure_concrete_host, to_frames, from_frames, get_tcp_server_address
+
+
+logger = logging.getLogger(__name__)
+
+
+class WSHandler(WebSocketHandler):
+    def __init__(
+        self,
+        application,
+        request,
+        handler=None,
+        deserialize=None,
+        listener=None,
+        **kwargs,
+    ):
+        self.handler = handler
+        self.deserialize = deserialize
+        self.request = request
+        self.listener = listener()
+        super().__init__(application, request, **kwargs)
+
+    async def open(self):
+        self.set_nodelay(True)
+        self.q = asyncio.Queue()
+        self.closed = False
+        self.comm = WSHandlerComm(self, deserialize=self.deserialize)
+        # `on_message` won't get called until `open` returns
+        # we need `open` to return to finish the handshake
+        asyncio.ensure_future(self.on_open())
+
+    async def on_open(self):
+        try:
+            await self.listener.on_connection(self.comm)
+        except CommClosedError:
+            logger.debug("Connection closed before handshake completed")
+        await self.handler(self.comm)
+
+    async def on_message(self, msg):
+        await self.q.put(msg)
+
+    def on_close(self):
+        self.closed = True
+        self.q.put_nowait(CommClosedError)
+
+    def close(self):
+        super().close()
+        self.closed = True
+
+
+class WSHandlerComm(Comm):
+    def __init__(self, handler, deserialize=True):
+        self.handler = handler
+        self.deserialize = deserialize
+        super().__init__()
+
+    async def read(self, deserializers=None):
+        try:
+            n_frames = await self.handler.q.get()
+        except RuntimeError:  # Event loop is closed
+            raise CommClosedError()
+
+        if n_frames is CommClosedError:
+            raise CommClosedError()
+        else:
+            n_frames = struct.unpack("Q", n_frames)[0]
+        frames = [(await self.handler.q.get()) for _ in range(n_frames)]
+        return await from_frames(
+            frames, deserialize=self.deserialize, deserializers=deserializers
+        )
+
+    async def write(self, msg, serializers=None, on_error=None):
+        frames = await to_frames(
+            msg,
+            serializers=serializers,
+            on_error=on_error,
+            context={
+                "sender": self.local_info,
+                "recipient": self.remote_info,
+                **self.handshake_options,
+            },
+        )
+        n = struct.pack("Q", len(frames))
+        try:
+            await self.handler.write_message(n, binary=True)
+            for frame in frames:
+                await self.handler.write_message(ensure_bytes(frame), binary=True)
+        except WebSocketClosedError as e:
+            raise CommClosedError(str(e))
+
+        return sum(map(nbytes, frames))
+
+    def abort(self):
+        self.handler.close()
+
+    @property
+    def local_address(self):
+        return self.handler.request.host
+
+    @property
+    def peer_address(self):
+        return self.handler.request.remote_ip + ":0"
+
+    @property
+    def _local_addr(self):
+        return self.handler.request.host
+
+    @property
+    def _peer_addr(self):
+        return self.handler.request.remote_ip + ":0"
+
+    def closed(self):
+        return (
+            self.handler.closed
+            or not self.handler.ws_connection
+            or self.handler.request.connection.stream
+            and self.handler.request.connection.stream.closed
+        )
+
+    async def close(self):
+        self.handler.close()
+
+
+class WS(Comm):
+    prefix = "ws://"
+
+    def __init__(self, sock, deserialize=True):
+        self._closed = False
+        Comm.__init__(self)
+        self.sock = sock
+        self._peer_addr = f"{self.prefix}{self.sock.parsed.netloc}"
+        self._local_addr = f"{self.prefix}{self.sock.parsed.netloc}"
+        self.deserialize = deserialize
+        self._finalizer = weakref.finalize(self, self._get_finalizer())
+        self._extra = {}
+        self._read_extra()
+
+    def _get_finalizer(self):
+        def finalize(sock=self.sock, r=repr(self)):
+            if not sock.close_code:
+                logger.info("Closing dangling websocket in %s", r)
+                sock.close()
+
+        return finalize
+
+    async def read(self, deserializers=None):
+        try:
+            n_frames = await self.sock.read_message()
+            if n_frames is None:
+                # Connection is closed
+                self.abort()
+                raise CommClosedError()
+            n_frames = struct.unpack("Q", n_frames)[0]
+        except WebSocketClosedError as e:
+            raise CommClosedError(e)
+
+        frames = [(await self.sock.read_message()) for _ in range(n_frames)]
+
+        return await from_frames(
+            frames, deserialize=self.deserialize, deserializers=deserializers
+        )
+
+    async def write(self, msg, serializers=None, on_error=None):
+        # TODO: Add
+        frames = await to_frames(
+            msg,
+            serializers=serializers,
+            on_error=on_error,
+            context={
+                "sender": self.local_info,
+                "recipient": self.remote_info,
+                **self.handshake_options,
+            },
+        )
+        n = struct.pack("Q", len(frames))
+        try:
+            await self.sock.write_message(n, binary=True)
+            for frame in frames:
+                await self.sock.write_message(ensure_bytes(frame), binary=True)
+        except WebSocketClosedError as e:
+            raise CommClosedError(e)
+
+        return sum(map(nbytes, frames))
+
+    async def close(self):
+        if not self.sock.close_code:
+            self._finalizer.detach()
+            self.sock.close()
+        self._closed = True
+
+    def abort(self):
+        if not self.sock.close_code:
+            self._finalizer.detach()
+            self.sock.close()
+        self._closed = True
+
+    def closed(self):
+        return not self.sock or self.sock.close_code or self._closed
+
+    @property
+    def local_address(self):
+        return f"{self.prefix}{self.sock.parsed.netloc}"
+
+    @property
+    def peer_address(self):
+        return f"{self.prefix}{self.sock.parsed.netloc}"
+
+    def _read_extra(self):
+        pass
+
+    @property
+    def extra_info(self):
+        return self._extra
+
+
+class WSS(WS):
+    prefix = "wss://"
+
+    def _read_extra(self):
+        WS._read_extra(self)
+        sock = self.sock.stream.socket
+        if sock is not None:
+            self._extra.update(peercert=sock.getpeercert(), cipher=sock.cipher())
+            cipher, proto, bits = self._extra["cipher"]
+            logger.debug(
+                "TLS connection with %r: protocol=%s, cipher=%s, bits=%d",
+                self._peer_addr,
+                proto,
+                cipher,
+                bits,
+            )
+
+
+class WSListener(Listener):
+    prefix = "ws://"
+
+    def __init__(
+        self,
+        address: str,
+        handler: Callable,
+        deserialize: bool = False,
+        **connection_args,
+    ):
+        if not address.startswith(self.prefix):
+            address = f"{self.prefix}{address}"
+
+        self.ip, self.port = parse_host_port(address, default_port=0)
+        self.handler = handler
+        self.deserialize = deserialize
+        self.connection_args = connection_args
+        self.bound_address = None
+        self.server_args = self._get_server_args(**connection_args)
+
+    def _get_server_args(self, **connection_args):
+        return {}
+
+    @property
+    def address(self) -> str:
+        return f"{self.prefix}{self.ip}:{self.port}"
+
+    async def start(self):
+        routes = [
+            (
+                r"/",
+                WSHandler,
+                {
+                    "handler": self.handler,
+                    "deserialize": self.deserialize,
+                    "listener": weakref.ref(self),
+                },
+            )
+        ]
+        try:
+            self.server = self.handler.__self__.http_server
+            self.handler.__self__.http_application.add_handlers(r".*", routes)
+        except AttributeError:
+            self.server = HTTPServer(web.Application(routes), **self.server_args)
+            self.server.listen(self.port)
+
+    async def stop(self):
+        self.server.stop()
+
+    def get_host_port(self):
+        """
+        The listening address as a (host, port) tuple.
+        """
+        if self.bound_address is None:
+            self.bound_address = get_tcp_server_address(self.server)
+        # IPv6 getsockname() can return more a 4-len tuple
+        return self.bound_address[:2]
+
+    @property
+    def listen_address(self) -> str:
+        return self.prefix + unparse_host_port(*self.get_host_port())
+
+    @property
+    def contact_address(self) -> str:
+        host, port = self.get_host_port()
+        host = ensure_concrete_host(host)
+        return self.prefix + unparse_host_port(host, port)
+
+
+class WSSListener(WSListener):
+    prefix = "wss://"
+
+    def _get_server_args(self, **connection_args):
+        ctx = _expect_tls_context(connection_args)
+        return {"ssl_options": ctx}
+
+
+class WSConnector(Connector):
+    prefix = "ws://"
+    comm_class = WS
+
+    async def connect(self, address, deserialize=True, **connection_args):
+        kwargs = self._get_connect_args(**connection_args)
+        try:
+            request = HTTPRequest(f"{self.prefix}{address}", **kwargs)
+            sock = await websocket_connect(request, max_message_size=10_000_000_000)
+            if sock.stream.closed() and sock.stream.error:
+                raise StreamClosedError(sock.stream.error)
+        except StreamClosedError as e:
+            convert_stream_closed_error(self, e)
+        except SSLError as err:
+            raise FatalCommClosedError() from err
+        return self.comm_class(sock)
+
+    def _get_connect_args(self, **connection_args):
+        return {}
+
+
+class WSSConnector(WSConnector):
+    prefix = "wss://"
+    comm_class = WSS
+
+    def _get_connect_args(self, **connection_args):
+        ctx = _expect_tls_context(connection_args)
+        return {"ssl_options": ctx}
+
+
+class WSBackend(BaseTCPBackend):
+    _connector_class = WSConnector
+    _listener_class = WSListener
+
+
+class WSSBackend(BaseTCPBackend):
+    _connector_class = WSSConnector
+    _listener_class = WSSListener
+
+
+backends["ws"] = WSBackend()
+backends["wss"] = WSSBackend()
--- a/distributed/comm/__init__.py
+++ b/distributed/comm/__init__.py
@@ -16,6 +16,7 @@ from .utils import get_tcp_server_address
 def _register_transports():
     from . import inproc
     from . import tcp
+    from . import ws
 
     try:
         from . import ucx
--- a/distributed/comm/tests/test_ws.py
+++ b/distributed/comm/tests/test_ws.py
@@ -1,0 +1,122 @@
+import warnings
+import pytest
+
+from distributed import Client, Scheduler, Worker
+from distributed.comm.registry import backends, get_backend
+from distributed.comm import ws, listen, connect
+from distributed.utils_test import (
+    get_client_ssl_context,
+    get_server_ssl_context,
+    cleanup,
+    inc,
+    captured_logger,
+)  # noqa: F401
+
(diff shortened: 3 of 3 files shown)


mrocklin:
Thank you for submitting this @marcosmoyano .  

@jcrist and @quasiben this may interest you

mrocklin:
@marcosmoyano when i try this locally I get a lot of pending task warnings

```python
In [2]: client = Client(protocol="ws")
In [3]: client
Out[3]: <Client: 'ws://192.168.0.2:8787' processes=4 threads=12, memory=16.44 GB>

In [4]: client.close()
In [5]:                                                                            
Do you really want to exit ([y]/n)? 
```

```
Task was destroyed but it is pending!
task: <Task pending name='Task-931' coro=<HTTP1ServerConnection._server_request_loop() running at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/http1connection.py:823> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f26ac43f9a0>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/ioloop.py:688]>
Task was destroyed but it is pending!
task: <Task pending name='Task-1219' coro=<HTTP1ServerConnection._server_request_loop() running at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/http1connection.py:823> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f26ac43f550>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/ioloop.py:688]>
Task was destroyed but it is pending!
task: <Task pending name='Task-930' coro=<HTTP1ServerConnection._server_request_loop() running at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/http1connection.py:823> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f26ac43f100>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/ioloop.py:688]>
Task was destroyed but it is pending!
task: <Task pending name='Task-1217' coro=<HTTP1ServerConnection._server_request_loop() running at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/http1connection.py:823> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f26c8048070>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/ioloop.py:688]>
Task was destroyed but it is pending!
task: <Task pending name='Task-1218' coro=<HTTP1ServerConnection._server_request_loop() running at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/http1connection.py:823> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f26ac43f430>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/mrocklin/miniconda/lib/python3.8/site-packages/tornado/ioloop.py:688]>
```

I know that in other PR we had fixed this.  Perhaps some of those changes didn't come through here?

marcosmoyano:
@mrocklin mmmh I'm not seeing that behavior
```python
In [2]: client = Client(protocol="ws")
In [3]: client
Out[3]: <Client: 'ws://192.168.68.111:8787' processes=4 threads=8, memory=33.52 GB>
In [4]: client.close()
In [5]:
Do you really want to exit ([y]/n)?
```
I'll double check the changes and make sure I didn't leave anything behind.

distributed/comm/ws.py
```diff
@@ -0,0 +1,372 @@
...
+                {
+                    "handler": self.handler,
+                    "deserialize": self.deserialize,
+                    "listener": weakref.ref(self),
+                },
+            )
+        ]
+        try:
+            self.server = self.handler.__self__.http_server
+            self.handler.__self__.http_application.add_handlers(r".*", routes)
```
marcosmoyano, line 295: I'm trying to reuse the dashboard's server, but in the context of a secure websocket connection I ran into an issue trying to do something like: 
```python
In [2]: client = Client(protocol="wss", security=True)
```
Looks like the dashboard ssl context (if present) is coming from `dask.config.get("distributed.scheduler.dashboard.tls.key") ...` and not from the self-signed cert. Since those were different servers this wasn't an issue, but now that I'm trying to use a single one for both dashboard and comms this is a problem.
The `start_http_server` method has the `ssl_options` kwarg (https://github.com/dask/distributed/blob/master/distributed/node.py#L99-L121) but it's not used by the scheduler/workers. And even if they did, those might get overridden from the config file.
Playing around with the scheduler code I got it to work with the self-signed cert by changing the `start_http_server` call from
```python
self.start_http_server(routes, dashboard_address, default_port=8787)
```
to
```python
self.start_http_server(
    routes, dashboard_address, default_port=8787, 
    ssl_options=self.connection_args["ssl_context"]
)
```
But in a more production settings this will probably not work.

I'm struggling to find a good approach with minimal changes.

Thoughts?

mrocklin: @jcrist if you have time do you have thoughts  on this?

quasiben:
This is really cool!  Do you have an idea of what performance is like when using WS compared with TCP ? According to [this post](https://medium.com/kifi-engineering/websockets-vs-regular-sockets-b3b8e7ea0708#:~:text=WebSockets%20performs%20quite%20well%2C%20with,an%20order%20of%20magnitude%20less.), we should generally expect network traffic to be 10x slower when using Websockets vs TCP

mrocklin:
I tried testing performance by sending a numpy array around, and ran into a bug.

```python

In [2]: client = Client(protocol="ws")

In [3]: import numpy as np

In [4]: x = np.arange(10000000)

In [5]: %time _  = client.scatter(x).result()
distributed.protocol.core - CRITICAL - Failed to Serialize
  File "/home/mrocklin/workspace/distributed/distributed/protocol/core.py", line 39, in dumps
    small_header, small_payload = dumps_msgpack(msg, **compress_opts)
  File "/home/mrocklin/workspace/distributed/distributed/protocol/core.py", line 184, in dumps_msgpack
    payload = msgpack.dumps(msg, default=msgpack_encode_default, use_bin_type=True)
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/msgpack/__init__.py", line 35, in packb
    return Packer(**kwargs).pack(o)
  File "msgpack/_packer.pyx", line 286, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 292, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 289, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 225, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 225, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 283, in msgpack._cmsgpack.Packer._pack
TypeError: can not serialize 'numpy.ndarray' object
distributed.comm.utils - ERROR - can not serialize 'numpy.ndarray' object
  File "/home/mrocklin/workspace/distributed/distributed/comm/utils.py", line 32, in _to_frames
    protocol.dumps(
  File "/home/mrocklin/workspace/distributed/distributed/protocol/core.py", line 39, in dumps
    small_header, small_payload = dumps_msgpack(msg, **compress_opts)
  File "/home/mrocklin/workspace/distributed/distributed/protocol/core.py", line 184, in dumps_msgpack
    payload = msgpack.dumps(msg, default=msgpack_encode_default, use_bin_type=True)
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/msgpack/__init__.py", line 35, in packb
    return Packer(**kwargs).pack(o)
  File "msgpack/_packer.pyx", line 286, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 292, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 289, in msgpack._cmsgpack.Packer.pack
  File "msgpack/_packer.pyx", line 225, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 225, in msgpack._cmsgpack.Packer._pack
  File "msgpack/_packer.pyx", line 283, in msgpack._cmsgpack.Packer._pack
TypeError: can not serialize 'numpy.ndarray' object
CommClosedError                           Traceback (most recent call last)
<timed exec> in <module>

~/workspace/distributed/distributed/client.py in result(self, timeout)
    219 
    220         # shorten error traceback
--> 221         result = self.client.sync(self._result, callback_timeout=timeout, raiseit=False)
    222         if self.status == "error":
    223             typ, exc, tb = result

~/workspace/distributed/distributed/client.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    834             return future
    835         else:
--> 836             return sync(
    837                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
    838             )

~/workspace/distributed/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    338     if error[0]:
    339         typ, exc, tb = error[0]
--> 340         raise exc.with_traceback(tb)
    341     else:
    342         return result[0]

~/workspace/distributed/distributed/utils.py in f()
    322             if callback_timeout is not None:
    323                 future = asyncio.wait_for(future, callback_timeout)
--> 324             result[0] = yield future
    325         except Exception as exc:
    326             error[0] = sys.exc_info()

~/miniconda/lib/python3.8/site-packages/tornado/gen.py in run(self)
    760 
    761                     try:
--> 762                         value = future.result()
    763                     except Exception:
    764                         exc_info = sys.exc_info()

~/workspace/distributed/distributed/client.py in _result(self, raiseit)
    244                 return exception
    245         else:
--> 246             result = await self.client._gather([self])
    247             return result[0]
    248 

~/workspace/distributed/distributed/client.py in _gather(self, futures, errors, direct, local_worker)
   1882                 else:
   1883                     self._gather_future = future
-> 1884                 response = await future
   1885 
   1886             if response["status"] == "error":

~/workspace/distributed/distributed/client.py in _gather_remote(self, direct, local_worker)
   1933 
   1934             else:  # ask scheduler to gather data for us
-> 1935                 response = await retry_operation(self.scheduler.gather, keys=keys)
   1936 
   1937         return response

~/workspace/distributed/distributed/utils_comm.py in retry_operation(coro, operation, *args, **kwargs)
    382         dask.config.get("distributed.comm.retry.delay.max"), default="s"
    383     )
--> 384     return await retry(
    385         partial(coro, *args, **kwargs),
    386         count=retry_count,

~/workspace/distributed/distributed/utils_comm.py in retry(coro, count, delay_min, delay_max, jitter_fraction, retry_on_exceptions, operation)
    367                 delay *= 1 + random.random() * jitter_fraction
    368             await asyncio.sleep(delay)
--> 369     return await coro()
    370 
    371 

~/workspace/distributed/distributed/core.py in send_recv_from_rpc(**kwargs)
    876             name, comm.name = comm.name, "ConnectionPool." + key
    877             try:
--> 878                 result = await send_recv(comm=comm, op=key, **kwargs)
    879             finally:
    880                 self.pool.reuse(self.addr, comm)

~/workspace/distributed/distributed/core.py in send_recv(comm, reply, serializers, deserializers, **kwargs)
    659         await comm.write(msg, serializers=serializers, on_error="raise")
    660         if reply:
--> 661             response = await comm.read(deserializers=deserializers)
    662         else:
    663             response = None

~/workspace/distributed/distributed/comm/ws.py in read(self, deserializers)
    170                 # Connection is closed
    171                 self.abort()
--> 172                 raise CommClosedError()
    173             n_frames = struct.unpack("Q", n_frames)[0]
    174         except WebSocketClosedError as e:

CommClosedError: 

In [6]: client = Client(protocol="tcp")
/home/mrocklin/workspace/distributed/distributed/node.py:151: UserWarning: Port 8787 is already in use.
Hosting the HTTP server on port 42385 instead
  warnings.warn(

In [7]: %time _  = client.scatter(x).result()
CPU times: user 168 ms, sys: 54.6 ms, total: 222 ms
Wall time: 237 ms
```

mrocklin:
It looks like it's just on gathering.  On the plus side, performance is fine.

```
In [17]: client = Client(protocol="ws")
/home/mrocklin/workspace/distributed/distributed/node.py:151: UserWarning: Port 8787 is already in use.
Hosting the HTTP server on port 44933 instead
  warnings.warn(

In [18]: %timeit client.scatter(x, hash=False)
65.8 ms ± 482 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [19]: client = Client(protocol="tcp")
/home/mrocklin/workspace/distributed/distributed/node.py:151: UserWarning: Port 8787 is already in use.
Hosting the HTTP server on port 46115 instead
  warnings.warn(

In [20]: %timeit client.scatter(x, hash=False)
65.5 ms ± 2.11 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

marcosmoyano:
@mrocklin I just fixed the bug you reported. Here's a timing comparison between ws vs tcp
```python
In [1]: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:from dask.distributed import Client
client = Client(protocol="ws")
import numpy as np
x = np.arange(10000000)
%time _  = client.scatter(x).result()::::
:--
CPU times: user 146 ms, sys: 62.5 ms, total: 208 ms
Wall time: 249 ms
```
```python
In [1]: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:from dask.distributed import Client
client = Client(protocol="tcp")
import numpy as np
x = np.arange(10000000)
%time _  = client.scatter(x).result()::::
:--
CPU times: user 153 ms, sys: 52 ms, total: 205 ms
Wall time: 248 ms
```

mrocklin:
Thanks for adding the fix.  Can I ask you to also add a test for the failure that we saw here?  I recommend using the `gen_cluster` style tests if that's easy for you.  https://distributed.dask.org/en/latest/develop.html#writing-tests

mrocklin:
In particular I would make a test that creates a small numpy array and then scatters and gathers it to ensure that we can roundtrip data through the client-scheduler-worker-scheduler-client path.

gen_cluster tests are async by default (this is way easier for future debugging) so you'll want something like ...

```python
    x = ...
    future = await c.scatter(x)
    y = await x
    assert (x == y).all()
```

jacobtomlinson:
Drive by question.

With TCP we create multiple connections from client->scheduler, scheduler->client, worker->scheduler, worker->worker, etc. Will all of that remain the same with this implementation?

mrocklin:
> With TCP we create multiple connections from client->scheduler, scheduler->client, worker->scheduler, worker->worker, etc. Will all of that remain the same with this implementation?

I think so?  I don't see any reason why that wouldn't be the case here.  Tornado websockets should handle multiple connections the same as the Tornado TCP server.

distributed/comm/ws.py
mrocklin: @jacobtomlinson you also seem like someone who might be able to help here.

jacobtomlinson:
Sure that makes sense. I guess my question was more around "do we have to?".

My understanding of websockets is that once a connection is created either end can send a message through the socket (maybe this is true for TCP too). So is the `scheduler->worker` connection necessary once the `worker->scheduler` connection has been established?

mrocklin:
I'm inclined to not mess with this too much yet.

There is some tricky logic around establishing connections, handling when
they break, and so on that I'd prefer not to get into if we don't have to :)

Although I guess that this might be useful in situations where only the
scheduler is visible on the network, and no one else has an accessible
network address.  I noticed your AWS Lambda conversation on Twitter.  This
might be an interesting solution for that?

mrocklin:
Ah, checking in on that thread on Twitter again I see that yes indeed, this
is where you were going.

This sounds like a fun thing to do.  It would probably take some care to do
well.  You'd have to hijack the current connection dance to all reuse the
same connection.  Totally doable, but requires some effort is my guess.

jacobtomlinson:
> Ah, checking in on that thread on Twitter again I see that yes indeed, this
is where you were going.

Yup that's where I was going with this. But it's definitely a tangent to this PR so thanks for all the answers, I'll have more of a think about it some time.

mrocklin:
Thinking about this again, while this would solve the problem for
scheduler-worker communications, we still don't have a good solution for
worker-worker communications.

jacobtomlinson:
Yeah I think the comm would end up being treated as a bus, with the scheduler connection becoming a bottleneck.

distributed/comm/ws.py
mrocklin: Ping @jacobtomlinson or @quasiben in case either of you have time/expertise to share here.

jacobtomlinson: Am I right in thinking that it is possible to configure different sets of certs for the dask comm and the dashboard web server?

And if so has anyone seen folks intentionally use different certs for the comm and dashboard in the past? If not perhaps some work so consolidate this would make things easier here.

marcosmoyano: > Am I right in thinking that it is possible to configure different sets of certs for the dask comm and the dashboard web server?

You are absolutely right, which is the problem I'm running into trying to share the same server.
> And if so has anyone seen folks intentionally use different certs for the comm and dashboard in the past?

I can't really speak to this. @mrocklin ?

> If not perhaps some work so consolidate this would make things easier here.

Definitely :)

mrocklin: > And if so has anyone seen folks intentionally use different certs for the comm and dashboard in the past?

Well, no one has used this yet, so no.  However, maybe a scenario here would be that we want the Dask communication to be secure, but don't mind if the dashboard is publicly accessible.

jcrist: Most users securing dask with TLS are using self-signed-certs that won't be usable via the browser (the browser won't have a compatible CA). As such, it's important to have the option for using different certs for Dask communication and the dashboard, so the browser can still work (either no TLS, or a cert the browser will accept).

mrocklin: Maybe a dumb question, but is it possible to have one HTTP server that uses different certs for different routes?

jcrist: The routes are encrypted, so by the time they get to the server the certs have to already be chosen.

Perhaps a good design here would be:

- By default spin up a 2nd server when using Websocket comms, unless the dask comm port is specified as the same as the dashboard port (then share the server).
- If sharing the server, use the dask comm certs first (if configured), fall back to the dashboard certs (if configured), then no certs.

mrocklin: @marcosmoyano , should you find yourself with free time before you head out (low chance, I know) this seems like a sensible approach (actually, I have no idea, but if Jim and Jacob both agree then I have high confidence).

marcosmoyano: Ok, so if the ports don't match behave like any other comm and fire up a 2nd server. +1
When the ports match try to share the server. It looks like the dashboard server fires up before the scheduler comm so I would have bring some protocol awareness before that happens to do the protocol/port checking and make sure that the `ssl_options` argument is respected. I'm not sure how clean it'll end up. I'll play with it and see how that goes. A couple of high level questions 
* What happens if the comm port is the same as the dashboard but no protocol was specified? Should it raise an exception? Default to using websockets?
* Looks like the ssl context for the dashboard server is created with `purpose=ssl.Purpose.SERVER_AUTH` while all other comms start with `purpose=ssl.Purpose.CLIENT_AUTH` (the naming of ssl.Purpose is confusing). If sharing the server those would have to match. Also, if anyone can explain why the dashboard server start with `SERVER_AUTH` that would be great. 
* When sharing the server, the dashboard won't work with self-signed certs. Should it default to no certs at all? Does the dashboard take precedence over having security communication within the cluster? Based on @jcrist comment it looks like it. Just verifying.

Alternatively, this could be handled through documentation. "If using the secure websocket protocol, Dask shares a single server for the dashboard and the comm so both certs have to match" 🤷🏼‍♂️

marcosmoyano: > Most users securing dask with TLS are using self-signed-certs that won't be usable via the browser
>> If sharing the server, use the dask comm certs first (if configured), fall back to the dashboard certs (if configured), then no certs.

So self-signed-certs from comms will take precedence over the dashboard. I think the dashboard will still work, but won't be over ssl.

Security concern: dask comms use bidirectional authentication with `ssl_context.verify_mode = ssl.CERT_REQUIRED` while the dashboard uses `ssl_context.verify_mode = ssl.CERT_NONE`. For the dashboard to keep working we need `ssl.CERT_NONE` but when sharing the server that means that the scheduler comm will also have `ssl.CERT_NONE`. Thoughts?

mrocklin:
OK, so an issue arose with security and the dashboard here.  It's convenient to have the dashboard and the comms running on the same server.  However, the dashboard probably wants to run with one-way-SSL (typical for HTTPS servers) while the Dask comms want to run on two-way-SSL (good for secure data communication).  If we run everything on one server then we have to choose one:

1.  If we choose two-way SSL by default then comms are secure, but the the dashboard becomes hard to access through a normal browser, which then won't have the client cert installed by default.

    To remediate this we would want to figure out ways to insert the client cert when possible (more on this below).

2.  If we choose one-way SSL by default then the dashboard would be visible by default, but we would need to add security to Dask comms after the fact

    To remediate this we would maybe want to write some Tornado middleware that checked for client certs in the HTTPS header and rejected the connection if they weren't correct (I am totally faking my understanding of this by the way, so this may or may not make sense).

Short term I think that we should start with option 1, mostly because it's simpler and less error-prone.  In this situation we're using secure websocket connections for Dask but losing easy access to the dashboard.  

But losing the dashboard is sad ;(  How can we resolve this?

1.  Well first, we can choose not to use HTTPS
2.  We can inject client certs in some cases when we proxy maybe?
    1. Can the Dask-Labextension do this?  (cc @ian-r-rose )
    2. Probably systems that deploy Dask (like Dask-Gateway, Coiled) could handle this 

I'm hopeful that as this gets used more other folks with more experience speak up.  If those folks are around now, I encourage them to speak up.

jacobtomlinson:
Of those choices I agree that option 1 makes the most sense here. 

But I feel the simplest and least error prone option is to also just continue to run two servers. I think the primary use case for this is on clusters where access is done via a Layer 7 reverse proxy and DNS (nginx, haproxy, traefik, etc) and therefore the TCP comm is not proxyable. We currently run the dashboard and TCP comm on two different ports, adding websockets could just keep that the same but use HTTP/WS for the comm allowing it to be proxied.

I worry that attempting to unify the two servers has added complexity that is blocking progress here when the benefit is only marginal.

jacobtomlinson:
I also have some slight concern that the reverse proxy will attempt to terminate the TLS connection anyway, because I think TLS can only be directly proxied at Layer 4.

mrocklin:
> But having users install client side certificates in their browser is not going to be a great experience for them.

I agree.  My hope was that short term this would either be solved by various proxy solutions, or that the Jlab extension might be able to assist here.  Long term my hope was that someone with a better understanding of this space would show up and say what we should actually do.  Maybe this is happening now :)

Yeah, so the situation that I'm trying to address is one where, as you suggest, network policies only allow in a few select ports, 443 among them, and various proxy solutions exist, but mostly just for things that look like web traffic.  OK, so this establishes that websockets are useful.

I have no objection to running two servers internally within the Dask scheduler, but it's important that they're both accessible to the user through a single port, 443.  Perhaps proxies can handle this for us, routing different endpoints on the same port to different servers hosting on different ports.  If so, great, that's a fine way to start.

jacobtomlinson:
> it's important that they're both accessible to the user through a single port, 443. Perhaps proxies can handle this for us

This could be totally possible with a Layer 7 proxy, but in this case the proxy would need to hold both sets of certificates (dashboard and comm) in order to inspect the HTTP headers and correctly route the traffic.

In the use case I'm thinking of the proxy will be owned and managed by the HPC/Service operator and we will have no control over it.

Alternatively you can route at Layer 4, Load balancers typically do this. However this will require using multiple ports as the traffic cannot be inspected. Ports are used to determine which upstream the traffic should be sent to. 

I guess the most complete solution here would be option 2 that you mentioned. Have tornado act as the proxy, and write some middleware to handle this. We could possibly do something where a specific URL path is proxied through to the comm and requires the client certificate too.

I feel like we are trying to achieve two things here:
- Support comm over websocket for scenarios where traffic is being reverse proxied for us and we have no control over it. Checking client certs will never be possible in this scenario.
- Support dashboard and comm on a single port for scenarios where we can only expose a single port.

I would suggest we focus on the first and perhaps worry about the second in some follow up work. But I base that on an assumption that the first will be much more common than the second.

marcosmoyano:
@mrocklin / @jacobtomlinson I'd appreciate your 👀 on the state of this PR after my latest changes.

mrocklin:
To the extent I'm able to judge (which is not much) I'm ready to merge this.  

There was an odd test failure in a rarely used module (cfexecutor).  I've pushed up a small fix to address it.  

@jacobtomlinson if you have time to take a look at this tomorrow that would be welcome.  If not I suggest that we just merge and continue with updates over time.  This code is not yet in anyone's workflow, and this PR has been up for a long while.

jacobtomlinson:
Thanks for merging this @mrocklin. I had a sync with @marcosmoyano the other day and agreed it was best to get things merged as they are and then we can iterate on it in follow up PRs.

PR #4397, opened 2021-01-04
adtive_target: do not mindlessly spawn workers when no memory limit is set

twoertwein:
`adaptive_target` wants to increase the number of workers when more than 60% of the memory is used. If no memory limit is set (`limit = 0`), it will always try to spawn more workers.


mrocklin:
This seems sensible to me.  Can you think of a simple test to add?

twoertwein:
I added a test and unfortunately rebased (lots of failing tests)

guillaumeeb:
@twoertwein do you think it is still an issue in Dask with the answer to dask/dask-jobqueue#463?

twoertwein:
@guillaumeeb The only connection between dask/dask-jobqueue#463 and this PR is that the issue made me look at this code :)  

I will rebase and test locally whether the test in this PR still fails when the other changes in this PR are not applied.

twoertwein:
The test fails locally when `... and limit > 0` is not part of the if condition.

I assume there might be a quicker test to verify that not more worker are spawned when no memory limit is set: the current test takes on my machine 5.59s to complete.

rene-rex:
Any progress on this? I really would like to use memory=0...

distributed/deploy/tests/test_adaptive.py
```diff
@@ -460,3 +460,19 @@ async def test_update_adaptive(cleanup):
...
+        n_workers=0, threads_per_worker=1, memory_limit=0, asynchronous=True
+    ) as cluster:
+        cluster.adapt(minimum=1, maximum=10, interval="100 ms")
+        async with Client(cluster, asynchronous=True) as client:
+            await client.gather(client.map(sleep, range(3)))
+        assert all(
+            state[1]["n"] <= 3
+            for state in cluster._adaptive.log
+            if state[1]["status"] == "up"
+        )
```
fjetter, line 478: Test runtime is not a huge concern. By using `slowinc` instead of sleep we can reduce runtime down to about 2s. That's effectively just a smaller sleep interval.

One thing I'd like to get in, though, is to change the assert statement. You are testing that there is no individual scaling beyond three instance but I think we should rather assert that there are not more than three in total. 

```suggestion
            await client.gather(client.map(slowinc, range(3)))
        assert (
            sum(
                state[1]["n"]
                for state in cluster._adaptive.log
                if state[1]["status"] == "up"
            )
            <= 3
        )
```

fjetter (commented):
Changes look otherwise good to me. @twoertwein would you mind adding the suggestion, then we're good to go.

distributed/deploy/tests/test_adaptive.py
twoertwein: Thank you! I will locally test `slowinc`. I think I had tried it at some point, but then the test did not fail when I removed the other parts of this PR. Probably needs some fine-tuning of `interval="100 ms"` and the number of `tasks`.

fjetter: you can also tweak the sleep time by calling `client.map(sleep, range(3), seconds=0.05)`. Same for `incslow` but the argument is called differently

twoertwein: Thanks! I assume I will need to also use the git version of dask and not 2021.04.0 as dask seems to import a non-exiting function:
```py
pytest -k test_adaptive_no_memory_limit distributed/deploy/tests/test_adaptive.py
[...]
>       from distributed.protocol.core import dumps_msgpack
E       ImportError: cannot import name 'dumps_msgpack' from 'distributed.protocol.core' (/results/twoertwe/home/git/distributed/distributed/protocol/core.py)

../../.cache/pypoetry/virtualenvs/python-tools-0mbQS0_g-py3.9/lib/python3.9/site-packages/dask/highlevelgraph.py:946: ImportError
```

fjetter: Yes, this was breaking in dask. We'll have a release upcoming this week, I believe which should resolve this. For now, you'll need master :/

fjetter:
Test failures appear to be #4651

distributed/deploy/tests/test_adaptive.py
twoertwein: I wasn't able to bring the `sleep`/`slowinc` below 0.35s, but now it fails when the actual patch isn't applied :)

twoertwein: @fjetter is there anything more to do?

distributed/scheduler.py
```diff
@@ -6898,17 +6898,12 @@ def adaptive_target(self, comm=None, target_duration=None):
...
-        # Memory
-        limit_bytes = {
-            addr: ws._memory_limit for addr, ws in parent._workers_dv.items()
-        }
-        worker_bytes = [ws._nbytes for ws in parent._workers_dv.values()]
-        limit = sum(limit_bytes.values())
-        total = sum(worker_bytes)
-        if total > 0.6 * limit:
+        # add more workers if more than 60% of memory is used
+        limit = sum(ws._memory_limit for ws in parent._workers_dv.values())
```
twoertwein, line 6902: Some tests fail with:
`AttributeError: 'distributed.scheduler.WorkerState' object has no attribute '_memory_limit'`
The existing code should have the same issue?!

twoertwein: mh, tests pass on the main branch even though it also accesses the same attribute? Any ideas @fjetter

twoertwein: The failing test with python 3.7 + cython. cython doesn't seem to like `sum(generator)`. similar to the previous code, I'm now using `sum(list comprehension)`

fjetter:
Thanks a lot @twoertwein!

PR #4402, opened 2021-01-06
Fix `test_as_current_is_thread_local`

jakirkham:
`s` is a `dict`, but we were treating it as the `Scheduler` object. This fixes that.

### DIFF
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -4035,7 +4035,7 @@ def test_as_current_is_thread_local(s):
                     l4.release()
 
     def run2():
-        with Client(s.address) as c:
+        with Client(s["address"]) as c:
             with c.as_current():
                 l1.release()
                 l2.acquire()
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -4011,6 +4011,7 @@ async def test_as_current(c, s, a, b):
     await c2.close()
 
 
+@gen_cluster(client=False)
 def test_as_current_is_thread_local(s):
     l1 = threading.Lock()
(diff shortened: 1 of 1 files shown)


jakirkham:
Was getting some messages locally about `s` being a `dict` so not having an `address` attribute. Noticed `gen_cluster` was missing here. So this adds it in the hope of fixing that issue. Though I could just be misunderstanding how this test works.

jrbourbeau (commented):
There's an `s` fixture over in `utils_test.py`

https://github.com/dask/distributed/blob/76ef459fbb0af0bf0f3c6f16c86c489c1bbafbb0/distributed/utils_test.py#L525-L528

which isn't quite being used correctly here as the `s` fixture returns a dictionary 

https://github.com/dask/distributed/blob/76ef459fbb0af0bf0f3c6f16c86c489c1bbafbb0/distributed/utils_test.py#L670-L673

not a `Scheduler` instance.

So here we want to keep the `s` fixture as is and change `with Client(s.address) as c:`  to `with Client(s["address"]) as c:`

jakirkham:
Ah great thanks for the advice James! Thought there might be a fixture, but didn't think to look outside this test file. Have made the suggested change. Please let me know if there is anything else 🙂

jrbourbeau (approved):
Thanks @jakirkham! I pushed a small commit with another `s.address` - > `s["address"]` update. This should be good to go, will merge once CI finishes

jakirkham:
Great thanks for catching that 😄

jrbourbeau:
Thanks @jakirkham! This is in

PR #4403, opened 2021-01-06
Make tests pass after 2028

bmwiedemann:
This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


jacobtomlinson:
Thanks for this. Seems sensible to bump these dates out.

Out of curiosity how does this change make builds reproducible?

bmwiedemann:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our `python-distributed` package build.
See https://reproducible-builds.org/ for why this matters.

So in the end it is about getting identical build results anytime anywhere and expirations break the anytime part.

jacobtomlinson:
Really interesting thanks! 

CI failures appear to be related to #4404.

PR #4406, opened 2021-01-06
Unify annotations

### LINKED ISSUE #4262: Use Layer annotations
In https://github.com/dask/dask/pull/6767 task annotations were added to `HighLevelGraph` `Layers`. We can now start supporting `priority`, `resources`, `retries`, etc. annotations here in the distributed scheduler.

@sjperkins you've been leading the task annotation effort and expressed an interested in incorporating them into `distributed`. Some colleagues and I have time to dedicate to this if now is a busy time for you, but I first wanted to check in with you to see if you're still interested in working on this

ian-r-rose:
Continues/supersedes #4347, using the task annotation machinery to specify workers/priority/retries/etc when submitting code to the scheduler. Fixes #4262

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1570,12 +1570,6 @@ class Client:
 
         if isinstance(workers, (str, Number)):
             workers = [workers]
-        if workers is not None:
-            restrictions = {skey: workers}
-            loose_restrictions = [skey] if allow_other_workers else []
-        else:
-            restrictions = {}
-            loose_restrictions = []
 
         if kwargs:
             dsk = {skey: (apply, func, list(args), kwargs)}
@@ -1585,8 +1579,8 @@ class Client:
         futures = self._graph_to_futures(
             dsk,
             [skey],
-            restrictions,
-            loose_restrictions,
+            workers=workers,
+            allow_other_workers=allow_other_workers,
             priority={skey: 0},
             user_priority=priority,
             resources={skey: resources} if resources else None,
@@ -1762,26 +1756,8 @@ class Client:
 
         if isinstance(workers, (str, Number)):
             workers = [workers]
-        if isinstance(workers, (list, set)):
-            if workers and isinstance(first(workers), (list, set)):
-                if len(workers) != len(keys):
-                    raise ValueError(
-                        "You only provided %d worker restrictions"
-                        " for a sequence of length %d" % (len(workers), len(keys))
-                    )
-                restrictions = dict(zip(keys, workers))
-            else:
-                restrictions = {k: workers for k in keys}
-        elif workers is None:
-            restrictions = {}
-        else:
+        if workers is not None and not isinstance(workers, (list, set)):
             raise TypeError("Workers must be a list or set of workers or None")
-        if allow_other_workers not in (True, False, None):
-            raise TypeError("allow_other_workers= must be True or False")
-        if allow_other_workers is True:
-            loose_restrictions = set(keys)
-        else:
-            loose_restrictions = set()
 
         internal_priority = dict(zip(keys, range(len(keys))))
 
@@ -1793,8 +1769,8 @@ class Client:
         futures = self._graph_to_futures(
             dsk,
             keys,
-            restrictions,
-            loose_restrictions,
+            workers=workers,
+            allow_other_workers=allow_other_workers,
             priority=internal_priority,
             resources=resources,
             retries=retries,
@@ -2605,8 +2581,8 @@ class Client:
         self,
         dsk,
         keys,
-        restrictions=None,
-        loose_restrictions=None,
+        workers=None,
+        allow_other_workers=None,
         resources=None,
         sync=True,
         asynchronous=None,
@@ -2652,8 +2628,8 @@ class Client:
         futures = self._graph_to_futures(
             dsk,
             keys=set(flatten([keys])),
-            restrictions=restrictions,
-            loose_restrictions=loose_restrictions,
+            workers=workers,
+            allow_other_workers=allow_other_workers,
             resources=resources,
             fifo_timeout=fifo_timeout,
             retries=retries,
@@ -2969,18 +2945,14 @@ class Client:
 
         names = {k for c in collections for k in flatten(c.__dask_keys__())}
 
-        restrictions, loose_restrictions = self.get_restrictions(
-            collections, workers, allow_other_workers
-        )
-
         if not isinstance(priority, Number):
             priority = {k: p for c, p in priority.items() for k in self._expand_key(c)}
 
         futures = self._graph_to_futures(
             dsk,
             names,
-            restrictions,
-            loose_restrictions,
+            workers=workers,
+            allow_other_workers=allow_other_workers,
             resources=resources,
             retries=retries,
             user_priority=priority,
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -11,9 +11,10 @@ import errno
 from functools import partial
 import html
 import inspect
+import itertools
 import json
 import logging
-from numbers import Number
+from numbers import Integral, Number
 import os
 import sys
 import uuid
@@ -2536,6 +2537,11 @@ class Client:
         actors=None,
     ):
         with self._refcount_lock:
+            if retries:
+                retries = self._expand_retries(
+                    retries, all_keys=itertools.chain(dsk, keys)
+                )
+
             if actors is not None and actors is not True and actors is not False:
                 actors = list(self._expand_key(actors))
 
@@ -2544,6 +2550,8 @@ class Client:
             # Make sure `dsk` is a high level graph
             if not isinstance(dsk, HighLevelGraph):
                 dsk = HighLevelGraph.from_collections(id(dsk), dsk, dependencies=())
+
+            dsk = highlevelgraph_pack(dsk, self, keyset)
 
             annotations = {}
             if user_priority:
@@ -2559,7 +2567,7 @@ class Client:
             if resources:
                 annotations["resources"] = resources
 
-            dsk = highlevelgraph_pack(dsk, self, keyset)
+            annotations = merge(dask.config.get("annotations", {}), annotations)
 
             # Create futures before sending graph (helps avoid contention)
             futures = {key: Future(key, self, inform=False) for key in keyset}
@@ -2833,7 +2841,10 @@ class Client:
         layers.update(dsk.layers)
         dependencies = {finalize_name: set(dsk.layers.keys())}
         dependencies.update(dsk.dependencies)
+        print("A")
+        print(layers)
         dsk = HighLevelGraph(layers, dependencies)
+        print("B")
 
         futures_dict = self._graph_to_futures(
             dsk,
@@ -3836,6 +3847,28 @@ class Client:
                     yield stringify(kkk)
             else:
                 yield stringify(kk)
+
+    @classmethod
+    def _expand_retries(cls, retries, all_keys):
+        """
+        Expand the user-provided "retries" specification
+        to a {task key: Integral} dictionary.
+        """
+        if retries and isinstance(retries, dict):
+            result = {
+                name: value
+                for key, value in retries.items()
+                for name in cls._expand_key(key)
+            }
+            result["__expanded_annotations__"] = True
+        elif isinstance(retries, Integral):
+            # If a simple number, it can be expanded on the scheduler
+            return retries
+        else:
+            raise TypeError(
+                "`retries` should be an integer or dict, got %r" % (type(retries))
+            )
+        return keymap(stringify, result)
 
     @staticmethod
     def collections_to_dsk(collections, *args, **kwargs):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2825,9 +2825,6 @@ class Client:
             else:
                 dsk2[name] = (func, keys) + extra_args
 
-        if not isinstance(priority, Number):
-            priority = {k: p for c, p in priority.items() for k in self._expand_key(c)}
-
         if not isinstance(dsk, HighLevelGraph):
             dsk = HighLevelGraph.from_collections(id(dsk), dsk, dependencies=())
 
@@ -2944,9 +2941,6 @@ class Client:
         dsk = self.collections_to_dsk(collections, optimize_graph, **kwargs)
 
         names = {k for c in collections for k in flatten(c.__dask_keys__())}
-
-        if not isinstance(priority, Number):
-            priority = {k: p for c, p in priority.items() for k in self._expand_key(c)}
 
         futures = self._graph_to_futures(
             dsk,
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2562,6 +2562,8 @@ class Client:
                 annotations["workers"] = workers
             if retries:
                 annotations["retries"] = retries
+            if allow_other_workers not in (True, False, None):
+                raise TypeError("allow_other_workers= must be True or False")
             if allow_other_workers:
                 annotations["allow_other_workers"] = allow_other_workers
             if resources:
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -11,10 +11,9 @@ import errno
 from functools import partial
 import html
 import inspect
-import itertools
 import json
 import logging
-from numbers import Integral, Number
+from numbers import Number
 import os
 import sys
 import uuid
@@ -2537,11 +2536,6 @@ class Client:
         actors=None,
     ):
         with self._refcount_lock:
-            if retries:
-                retries = self._expand_retries(
-                    retries, all_keys=itertools.chain(dsk, keys)
-                )
-
             if actors is not None and actors is not True and actors is not False:
                 actors = list(self._expand_key(actors))
 
@@ -3846,28 +3840,6 @@ class Client:
                     yield stringify(kkk)
             else:
                 yield stringify(kk)
-
-    @classmethod
-    def _expand_retries(cls, retries, all_keys):
-        """
-        Expand the user-provided "retries" specification
-        to a {task key: Integral} dictionary.
-        """
-        if retries and isinstance(retries, dict):
-            result = {
-                name: value
-                for key, value in retries.items()
-                for name in cls._expand_key(key)
-            }
-            result["__expanded_annotations__"] = True
-        elif isinstance(retries, Integral):
-            # If a simple number, it can be expanded on the scheduler
-            return retries
-        else:
-            raise TypeError(
-                "`retries` should be an integer or dict, got %r" % (type(retries))
-            )
-        return keymap(stringify, result)
 
     @staticmethod
     def collections_to_dsk(collections, *args, **kwargs):
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1522,9 +1522,8 @@ class Client:
         fifo_timeout: str timedelta (default '100ms')
             Allowed amount of time between calls to consider the same priority
         resources: dict (defaults to {})
-            Defines the `resources` these tasks require on the worker. Can
-            specify global resources (``{'GPU': 2}``), or per-task resources
-            (``{'x': {'GPU': 1}, 'y': {'SSD': 4}}``), but not both.
+            Defines the `resources` each instance of this mapped task requires
+            on the worker; e.g. ``{'GPU': 2}``.
             See :doc:`worker resources <resources>` for details on defining
             resources.
         actor: bool (default False)
@@ -1641,8 +1640,8 @@ class Client:
             Allowed amount of time between calls to consider the same priority
         resources: dict (defaults to {})
             Defines the `resources` each instance of this mapped task requires
-            on the worker; e.g. ``{'GPU': 2}``. See
-            :doc:`worker resources <resources>` for details on defining
+            on the worker; e.g. ``{'GPU': 2}``.
+            See :doc:`worker resources <resources>` for details on defining
             resources.
         actor: bool (default False)
             Whether these tasks should exist on the worker as stateful actors.
@@ -2612,9 +2611,8 @@ class Client:
             Optional prioritization of task.  Zero is default.
             Higher priorities take precedence
         resources: dict (defaults to {})
-            Defines the `resources` these tasks require on the worker. Can
-            specify global resources (``{'GPU': 2}``), or per-task resources
-            (``{'x': {'GPU': 1}, 'y': {'SSD': 4}}``), but not both.
+            Defines the `resources` each instance of this mapped task requires
+            on the worker; e.g. ``{'GPU': 2}``.
             See :doc:`worker resources <resources>` for details on defining
             resources.
         sync: bool (optional)
@@ -2763,9 +2761,8 @@ class Client:
             be expensive. If none of the arguments contain any dask objects,
             set ``traverse=False`` to avoid doing this traversal.
         resources: dict (defaults to {})
-            Defines the `resources` these tasks require on the worker. Can
-            specify global resources (``{'GPU': 2}``), or per-task resources
-            (``{'x': {'GPU': 1}, 'y': {'SSD': 4}}``), but not both.
+            Defines the `resources` each instance of this mapped task requires
+            on the worker; e.g. ``{'GPU': 2}``.
             See :doc:`worker resources <resources>` for details on defining
             resources.
         actors: bool or dict (default None)
@@ -2912,9 +2909,8 @@ class Client:
         fifo_timeout: timedelta str (defaults to '60s')
             Allowed amount of time between calls to consider the same priority
         resources: dict (defaults to {})
-            Defines the `resources` these tasks require on the worker. Can
-            specify global resources (``{'GPU': 2}``), or per-task resources
-            (``{'x': {'GPU': 1}, 'y': {'SSD': 4}}``), but not both.
+            Defines the `resources` each instance of this mapped task requires
+            on the worker; e.g. ``{'GPU': 2}``.
             See :doc:`worker resources <resources>` for details on defining
             resources.
         actors: bool or dict (default None)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -11,10 +11,9 @@ import errno
 from functools import partial
 import html
 import inspect
-import itertools
 import json
 import logging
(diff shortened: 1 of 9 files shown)


distributed/protocol/highlevelgraph.py
```diff
@@ -136,12 +136,18 @@ def _materialized_layer_unpack(state, dsk, dependencies, annotations):
...
+        new_annotations = {}
+        for k, v in expanded.items():
+            if isinstance(v, dict) and k in annotations:
+                new_annotations[k] = merge(annotations[k], v)
+            else:
+                new_annotations[k] = v
+        annotations.update(new_annotations)
 
 
 def highlevelgraph_unpack(dumped_hlg):
```
jrbourbeau, line 149:
```suggestion
```

ian-r-rose: oop, merge issue

jrbourbeau (commented):
Thanks for your work here @ian-r-rose. It looks like there are some merge conflicts which are preventing CI from running. When you get a chance, would you mind merging /rebasing `master`?

ian-r-rose:
We've had some out-of-band discussion about the ability to annotate specific collections with things like `workers` or `retries` using commands like
```python
c.compute(x, retries={x: 2})
```
This worked for some collections, but not others. In 6e0695b I remove that ability, so that should be a good overview of what some of the consequences are. This is a breaking change, but perhaps of some functionality that was not widely used.

These per-collection annotations can be accomplished with the new syntax using something like
```python
with dask.annotate(retries=2):
    x = da.ones((1,1)
    c.compute(x)
```

This largely works, with the caveat that dask/dask#7036 should be fixed to get around some surprising results.

mrocklin (commented):
Whoops.  It looks like I might have had some lingering comments in review.  Submitting now, although they may be out of date.

distributed/client.py
```diff
@@ -3890,78 +3856,16 @@ def _expand_retries(cls, retries, all_keys):
...
-            loose_restrictions = list(restrictions)
-        elif allow_other_workers:
-            loose_restrictions = list(
-                {k for c in flatten(allow_other_workers) for k in c.__dask_keys__()}
-            )
-        else:
-            loose_restrictions = []
-
-        return restrictions, loose_restrictions
-
```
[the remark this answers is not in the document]
ian-r-rose: :+1:

distributed/tests/test_client.py
```diff
@@ -948,16 +923,6 @@ async def test_restrictions_map(c, s, a, b):
...
 
-    L = c.map(inc, [10, 11, 12], workers=[{a.ip}, {a.ip, b.ip}, {b.ip}])
-    await wait(L)
-
-    assert s.host_restrictions[L[0].key] == {a.ip}
-    assert s.host_restrictions[L[1].key] == {a.ip, b.ip}
-    assert s.host_restrictions[L[2].key] == {b.ip}
-
-    with pytest.raises(ValueError):
-        c.map(inc, [10, 11, 12], workers=[{a.ip}])
```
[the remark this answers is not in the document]
ian-r-rose: I was generally trying to avoid any expansion of keys/collections on the client, instead preferring to *only* allow a worker-or-iterable-of-workers. With that API restriction, the unpacking/expanding logic on the scheduler is much simpler. So I view this as philosophically aligned with removing the ability to expand `{ collection: worker }` on the client.

ian-r-rose (commented):
Thanks for taking a look @mrocklin, I've responded to your comments inline.

I would also flag: there are a lot of tests which now have `optimize_graph=False`. This is due to 
https://github.com/dask/dask/issues/7036, and should give an indication of what kinds of situations that affects

distributed/client.py
```diff
@@ -2587,8 +2550,23 @@ def _graph_to_futures(
...
 
-            if isinstance(retries, Number) and retries > 0:
-                retries = {k: retries for k in dsk}
+            annotations = {}
+            if user_priority:
+                annotations["priority"] = user_priority
+            if workers:
+                if not isinstance(workers, (list, tuple, set)):
+                    workers = [workers]
+                annotations["workers"] = workers
```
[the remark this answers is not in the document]
ian-r-rose: Yeah, looks like the `set` case is tested in a few places, e.g., https://github.com/ian-r-rose/distributed/blob/e63cbbecd1beffd96b2cab0927a2743edb4651c8/distributed/tests/test_client.py#L905-L914 , so I think we are covered

mrocklin:
Checking in.  How are we doing here?

ian-r-rose:
I think this is ready for review from my perspective. Tests are passing (except some that seem flaky and unrelated?), the major question in my mind is whether the backwards-incompatibilties introduced are worth it, and what to do about dask/dask#7036 .

mrocklin:
@sjperkins are you around to review?

sjperkins:
Sure will take a look over the next day or two. Thanks for working on this @ian-r-rose

distributed/protocol/highlevelgraph.py
```diff
@@ -130,12 +130,17 @@ def _materialized_layer_unpack(state, dsk, dependencies, annotations):
...
-            Layer.expand_annotations(state["annotations"], state["dsk"].keys())
-        )
+        expanded = Layer.expand_annotations(state["annotations"], state["dsk"].keys())
+        new_annotations = {}
+        for k, v in expanded.items():
+            if isinstance(v, dict) and k in annotations:
+                new_annotations[k] = merge(annotations[k], v)
+            else:
+                new_annotations[k] = v
+        annotations.update(new_annotations)
```
sjperkins, line 140: I understand this change provides backwards compatibility for the existing per-key Client.{submit, persist} kwarg taxonomy (retries, priority, workers)?

The functionality in **_materialized_layer_unpack** mirrors [Layer.\_\_dask_distributed_unpack\_\_](https://github.com/dask/dask/blob/d05508163086a0d09c65cb7d91a3aaab35e79a2e/dask/highlevelgraph.py#L224-L231) and it all it's subclasses. They also need to expand and unpack annotations.

I wonder if it would be better to place this functionality in Layer.expand_annotations so that it automatically is applied through the Layer class hierarchy? It would require a sister dask PR though.

distributed/protocol/highlevelgraph.py
```diff
@@ -157,18 +167,21 @@ def highlevelgraph_unpack(dumped_hlg):
...
+            if layer["state"]["annotations"] is None:
+                layer["state"]["annotations"] = {}
+            layer["state"]["annotations"].update(annotations)
         if layer["__module__"] is None:  # Default implementation
             unpack_func = _materialized_layer_unpack
         else:
             mod = import_allowed_module(layer["__module__"])
             unpack_func = getattr(mod, layer["__name__"]).__dask_distributed_unpack__
-        unpack_func(layer["state"], dsk, deps, annotations)
+        unpack_func(layer["state"], dsk, deps, out_annotations)
```
sjperkins, line 185: To me this implies that Layer annotations always take precedence over the existing Client.{submit, persist} kwarg taxonomy. I think this makes sense in terms of backwards compatibility (specific annotations override global annotations)

distributed/client.py
```diff
@@ -2632,14 +2597,23 @@ def get(
...
             Number of allowed automatic retries if computing a result fails
         priority: Number
             Optional prioritization of task.  Zero is default.
             Higher priorities take precedence
+        resources: dict (defaults to {})
+            Defines the `resources` these tasks require on the worker. Can
+            specify global resources (``{'GPU': 2}``), or per-task resources
+            (``{'x': {'GPU': 1}, 'y': {'SSD': 4}}``), but not both.
+            See :doc:`worker resources <resources>` for details on defining
+            resources.
```
sjperkins, line 2616: @ian-r-rose @mrocklin What do you think about adding a deprecation warning to the existing Client.{persist, compute} kwarg taxonomy?

sjperkins (commented):
I've added some comments: I'm most interested in the one concerning annotation expansion.

> I would also flag: there are a lot of tests which now have optimize_graph=False. This is due to
dask/dask#7036, and should give an indication of what kinds of situations that affects.

I might suggest commenting each optimize_graph=False case to indicate that it should be removed when dask/dask#7036 is fixed. Its a bit laborious and ugly but I can't think of a better way to handle this at present.

distributed/protocol/highlevelgraph.py
ian-r-rose: Thanks for pointing this out. This is actually not for backwards compatibility of per-key/collection annotations, but instead is needed to handle the case where different HLG Layers have different retries/priority/workers.

For example, if we are unpacking a two-layer HLG, and Layer A has `retries:2` while Layer B has `retries:4`, we need to make sure that the unpacking keeps both of those annotations. The current implementation has a shallow merge, which makes later-unpacked layers overwrite the annotations of previously-unpacked layers. To make the above case work, we need a bit deeper of a merge.

For the same reason, I also think this logic needs to be one step above `Layer.expand_annotations`.

distributed/client.py
ian-r-rose: Good point, this per-task/collection doc is no longer correct.

I don't currently see a clean way to keep it around for a deprecation cycle without doing the work to expand these annotations client-side.

distributed/protocol/highlevelgraph.py
sjperkins: Yes, you're right. Thanks for correcting my understanding.

```python
new_annotations[k] = {**annotations[k], **v}
```

may be faster than tlz.merge now that dask supports >= Python 3.6

mrocklin: cc'ing @madsbk

ian-r-rose:
@mrocklin You mentioned offline that passing in `{key: priority}` for this functionality still works, but `{collection: priority}` does not. At least as it currently stands, that is not the case here.

For simplicity's sake, this PR has tried to completely move such annotations to live at the HLG Layer level, and remove the older ways of doing it, including `{key: priority}`. This is partially because the scheduler needs to know when to unpack the annotations, and when to leave them alone. For callables, this is accomplished by adding a sentinel dict key `__expanded_annotations__`, e.g. https://github.com/dask/dask/blob/ede43a536cc25ad8e2611c59058f415436ae8764/dask/highlevelgraph.py#L100-L105

I've tried to avoid a proliferation such special cases in the pack/unpack logic. We could revisit that, but at least at the moment, retries/priority/workers/etc only take very simple non-graph objects.

madsbk (commented):
Overall this look great @ian-r-rose but I think we should consider simplifying the annotation design in order avoid duplicate annotation implementations in each Layer (see comment).

distributed/protocol/highlevelgraph.py
ian-r-rose: Thanks @madsbk, this is a great point. There is also the `Layer.expand_annotations` function called in the default implementation, which it may make sense to remove for the same reasons. @sjperkins are you aware of any major downsides to doing this?

sjperkins: @ian-r-rose @madsbk I do have some objections around [low code cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)), but I'm trying to provide some designs to rectify this. I'll try answer in more detail today.

sjperkins: I'm concerned about [low code cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)) in the sense that functionality for serialising and deserialising layers is spread across dask and dask distributed  (_materialized_layer_pack and _materialized_layer_unpack) in particular. Moving annotation expansion into dask distributed further exacerbates this trend.

`__dask_distributed_pack__` and `__dask_distributed_unpack__` appear to describe an API for serialising/deserialising layers similar to pickle, but `__dask_distributed_unpack__` does not reconstruct the original Layer, instead returning a low-level dask graph and associated dependencies for ingest by the scheduler.

I suggest:

1. Modifying `__dask_distributed_pack__` to return a `(cls, args, kwargs)` tuple (similar to the `__reduce__` protocol) for encoding by msgpack instead of a state dict . `__dask_distributed_pack__` could be renamed to `__dask_distributed_reduce__`.

2. This can be used to reconstruct the Layer server-side:

    ```python
   # client-side
   state = mspack.pack(layer.__dask_distributed_reduce__(client))
   # server-side
    cls, args, kwargs = msgpack.unpack(state)
    layer = cls(*args, **kwargs)
    ```

3. Replace classmethod  `__dask_distributed_unpack__(cls, state)` with instance method `__dask_distributed_unpack__(self)` which provides the raw `dsk` and `deps` dicts.

4.  Replace `_materalized_layer_pack` and `_materialized_layer_unpack` in dask distributed with a `class MaterializedLayer(Layer)` in dask that encapsulates the same functionaliy.

With this design, annotations can be packed in `(cls, args, kwargs)` automatically by Layer subclass implementers calling super() in `__dask_distributed_reduce__` (see extended example below).

Also MaterializedLayer gets placed in dask so there's no split across dask distributed.

This design prevents further low cohesion across the code base.

The code below illustrates most of these ideas:


```python
stringify = lambda x: x

class Layer:
    def __init__(self, annotations=None):
        self.annotations = annotations

    def __dask_distributed_reduce__(self, client):
        """
        Returns
        -------
        cls : class
        *args : Arguments for constructing a layer
        *kwargs : Keyword arguments for constructing a layer
        """
        return Layer, (), {"annotations": self.pack_annotations()}

    def __dask_distributed_unpack__(self):
        """
        Returns
        -------
        dsk : dict
        deps : dict
        """
        raise {stringify(k): v for k, v in self.items()}, {}


    def pack_annotations(self):
        """ This should actually do annotation packing """
        return self.annotations


    def materialize(self, all_keys, known_key_dependencies, client, client_keys):
        """ Implement _materialized_layer_pack """
        return MaterializedLayer({"dask": "graph"}, {"deps": "keys"},
                                 annotations=self.annotations)


class BasicLayer(Layer):
    def __init__(self, mapping, annotations=None):
        super().__init__(annotations=annotations)
        self.mapping = mapping

    def __dask_distributed_reduce__(self, client):
        _, args, kwargs = super().__dask_distributed_reduce__(client)
        assert not args
        return self.__class__, (self.mapping,), kwargs

    def __dask_distributed_unpack__(self):
        dsk, deps = super().__dask_distributed_unpack__()
        # update deps
        return dsk, deps


class Blockwise(Layer):
    def __init__(self, output, output_indices, dsk, indices,
                 numblocks, concatenate, new_axes, output_blocks,
                 annotations=None):
        super().__init__(annnotations=annotations)
        self.output = output
        self.output_indices = output_indices
        self.dsk = dsk
        self.indices = indices
        self.numblocks
        self.concatenate = concatenate
        self.new_axes = new_axes
        self.output_blocks = output_blocks

    def __dask_distributed_reduce__(self, client):
        _, args, kwargs = super().__dask_distributed_reduce__(client)
        assert not args
        args = (self.output, self.output_indices, self.dsk, self.indices,
                self.numblocks, self.concatenate, self.new_axes,
                self.output_blocks)
        return Blockwise, args, kwargs

    def __dask_distributed_unpack__(self):
        raw, raw_deps = make_blockwise_graph()
        dsk = {stringify(k): v for k, v in raw.items()}
        deps = {stringify(k): v for k, v in raw_deps.items()}
        return dsk, deps


class MaterializedLayer(Layer):
    def __init__(self, dsk, deps, annotations=None):
        super().__init__(annotations=annotations)
        self.dsk = dsk
        self.deps = deps

    def __dask_distributed_reduce__(self, client):
        cls, args, kwargs = super().__dask_distributed_reduce__(client)

        # Update args, kwargs etc.
        assert not args
        args = args + (self, dsk, self.deps)
        return MaterializedLayer, args, kwargs

    def __dask_distributed_unpack__(self):
        dsk = {stringify(k): v for k, v in self.dsk.items()}
        deps = {stringify(k): v for k, v in self.deps.items()}
        return dsk, deps


def encode_layer(client, layers):
    """ Client-side """
    return [msgpack.pack(l.__dask_distributed_reduce__(client)) for l in layers]


def decode_layers(layers):
    """ Server side """"
    decoded_layers = []

    for layer in layers:
        cls, args, kwargs = msgpack.unpack(layer)
        decoded_layers.append(cls(*args, **kwargs))

    return decode_layers

def reify_layers(layers):
    """ Server side """
    dsk = {}
    deps = {}
    annotations = {}

    for layer in layers:
        layer_dsk, layer_deps = layer.__dask_distributed_unpack__()
        dsk.update(layer_dsk)
        deps.update(layer_deps)
        annotations.update(Layer.expand_annotations(layer.annotations, layer.keys()))

    return dsk, deps, annotations
```

madsbk: Initially, we decided to implement the functionality that depended on Distributed in Distributed in order to avoid importing distributed in Dask. However, since we have to do that anyway e.g. in Blockwise I agree that it makes sense to move everything to Dask. I also agree that implementing the default pack/unpack in `Layer` makes a lot of sense. 

However, I don't think we need to change the basic logical flow and because of the asymmetric nature of `__dask_distributed_pack__` and `__dask_distributed_unpack__` and because we have very [specific requirement for `__dask_distributed_pack__`](https://github.com/dask/dask/blob/30c9a3a34240e08cccd7e38b115141661ca198cf/dask/highlevelgraph.py#L202-L207), I don't we should rename it to `__dask_distributed_reduce__`. 

Also notice that:
`__dask_distributed_unpack__` needs `dsk` and `dependencies` from the preceding layers.
`__dask_distributed_reduce__` for security reasons we cannot de-serializes class/modules directly.

sjperkins: > However, I don't think we need to change the basic logical flow and because of the asymmetric nature of `__dask_distributed_pack__` and `__dask_distributed_unpack__` and because we have very specific requirement for `__dask_distributed_pack__`, I don't we should rename it to `__dask_distributed_reduce__`.

I read the requirements and understand this better now.

> Also notice that:
`__dask_distributed_unpack__` needs dsk and dependencies from the preceding layers.

Ah, I see. Then, to separate annotation expansion from `__dask_distributed_unpack__` I would say that `__dask_distributed_unpack__` should not modfiy the dsk and deps arguments, but instead construct layer specific dsk and deps for return (Layer.expand_annotations needs to know the keys for the associated layer).

I almost might suggest modifying `__dask_distributed_unpack__` to return a dict for future-proofing return of any additional state required on the scheduler. I think the dict looks a bit aesthetically ugly, but I can't think of anything better off-hand. Thoughts?

```python
class SomeLayer(Layer):
    def __dask_distributed_unpack__(cls, state, dsk, deps):
        layer_dsk = {... using state }
        layer_deps = {... using state and dsk and deps }

        return {"dsk": layer_dsk, "deps": layer_deps, "keys": layer_dsk.keys()}

def unpack(layer_state):
    dsk  = {}
    deps = {}
    annotations = {}

    for state in layer_state:
        layer_cls = state["class"]
        state = state["state"]
        data = layer_cls.__dask_distributed_unpack(cls, state, dsk, deps)
        dsk.update(data["dsk"])
        deps.update(data["deps"])
        # Layer.expand_annotations must expand over the layer keys only
        annotations.update(Layer.expand_annotations(state["annotations"], data["dsk"].keys())
    
   return dsk, deps, annotations
```

`__dask_distributed_reduce__` for security reasons we cannot de-serializes class/modules directly.

:+1:

sjperkins: An alternative where `__dask_distributed_unpack__` returns a dict and its possible to defer to the superclass for annotation unpacking

```python
class Layer:
    def __dask_distributed_pack__(self, client):
        return {"annotations": self.pack_annotations}

    def __dask_distributed_unpack__(cls, state, dsk, deps):
        try:
            annotations = state["annotations"]
        except KeyError:
            return {}

        try:
            layer_dsk = state["dsk"]
        except KeyError:
            raise ValueError("dsk must be in state to expand annotations over layer keys")
        else:
            keys = layer_dsk.keys()

        return {"annotations": cls.expand_annotations(annotations, keys)}

class SomeLayer(Layer):
    def __dask_distributed_unpack__(cls, state, dsk, deps):
        layer_dsk = {... using state }
        layer_deps = {... using state and dsk and deps }

        # Defer to superclass for automatic layer unpacking.
        return {"dsk": layer_dsk,
                    "deps": layer_deps":
                    "keys": layer_dsk.keys(),
                   **super().__dask_distributed_unpack__({"dsk": layer_dsk, **state}, dsk, deps)}


def unpack(layer_state):
    dsk  = {}
    deps = {}
    annotations = {}

    for state in layer_state:
        layer_cls = state["class"]
        state = state["state"]
        data = layer_cls.__dask_distributed_unpack(cls, state, dsk, deps)
        dsk.update(data["dsk"])
        deps.update(data["deps"])
        annotations.update(data["annotations"])
    
   return dsk, deps, annotations
```

ian-r-rose: Thanks for the detailed discussion @sjperkins and @madsbk. I've tried to do something like a compromise solution in dask/dask#7102, and I'd love to hear your thoughts on it. It takes a somewhat minimal approach to API changes (so `__dask_distributed_unpack__()` still takes the annotations), and avoids further imports of `distributed` there. Overall the strategy I'm trying out is

1. Move the default implementation of layer unpacking into `Layer`.
2. Move the logic of doing a deeper merge on annotations into a new `Layer.merge_annotations`, similar to `Layer.expand_annotations`. I am not overly attached to this division of responsibility, so if there are suggestions for how to make it cleaner, I'd be happy to hear them.
3. Ensure that `Layer`, `Blockwise`, and `ShuffleLayer` all unpack annotations properly (previously the latter two failed).

madsbk: LGTM @ian-r-rose but I think @sjperkins has a point, we should unify everything in Dask and change the API such that layers can delegate the annotation unpacking to `Layer.__dask_distributed_unpack__()`.

I suggest that we split this work into multiple PRs. Let's rollback this PR to the state that didn't require any significant change to the API and get it merged. A followup PR can then modify the API based on the design thoughts we have been discussing here. I will be happy to give that a try.

sjperkins: Thanks for the discussion @madsbk and @ian-r-rose . I agree with moving Layer API changes forward to a new PR, so don't let my concerns hold this one up. LGTM

ian-r-rose: @madsbk I'd be happy to have this be across multiple PRs. Can you specify what API changes you would want to roll back? This now depends on dask/dask#7102, but that is somewhat more of a bugfix than a major API change.

I could bring the default implementation of `Layer.__dask_distributed_unpack__()` back here and then xfail the test that catches dask/dask#7102 , if that's what you mean.

madsbk: Sorry my bad, I thought you changed the API in the last commit. I am totally fine with this PR+dask/dask#7102

ian-r-rose:
n.b., now that this depends on dask/dask#7102, tests are failing, but they should pass once that PR is resolved

mrocklin:
How are we doing here?

ian-r-rose:
> How are we doing here?

dask/dask#7102 (on which this now depends) has been merged, I think this is ready (test failures look unrelated to me)

jrbourbeau:
My sense is this is good for a final review and merge

mrocklin:
> (test failures look unrelated to me)

Note, I've been ignoring travis failures for a while now, for better or worse.

distributed/tests/test_client.py
```diff
@@ -4110,20 +4134,44 @@ async def test_persist_workers(e, s, a, b, c):
...
+@gen_cluster(nthreads=[("127.0.0.1", 1)] * 3, client=True)
+async def test_compute_workers_annotate(e, s, a, b, c):
+    with dask.annotate(workers=a.address, allow_other_workers=True):
+        L1 = [delayed(inc)(i) for i in range(4)]
+    with dask.annotate(workers=b.address, allow_other_workers=True):
+        total = delayed(sum)(L1)
+    with dask.annotate(
+        workers=[
+            c.address,
+        ]
```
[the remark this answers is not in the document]
ian-r-rose: Wow, don't know why `black` let this through

distributed/tests/test_client.py
```diff
@@ -4100,6 +4101,29 @@ async def run2():
...
+    await wait(out)
+    assert all(v.key in a.data for v in L1)
+    assert total.key in b.data
+
+    assert s.loose_restrictions == {total2.key} | {v.key for v in L2}
+
+
 @nodebug  # test timing is fragile
 @gen_cluster(nthreads=[("127.0.0.1", 1)] * 3, client=True)
 async def test_persist_workers(e, s, a, b, c):
```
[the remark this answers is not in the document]
ian-r-rose: I've updated them to persist/compute on a given set of workers, but it is no longer possible to delegate certain tasks/keys to a specific worker in a persist/compute call. In order to do that, you now need to use the annotation framework.

> We're now covering the certain collections on certain workers functionality with new `test_persist_workers_annotate` and `test_compute_workers_annotate` tests. Is that correct?

Correct

distributed/tests/test_client.py
```diff
@@ -6363,6 +6424,37 @@ async def test_annotations_retries(c, s, a, b):
     assert all(ts.annotations == {"retries": 2} for ts in s.tasks.values())
 
 
+@gen_cluster(client=True)
+async def test_annotations_blockwise_unpack(c, s, a, b):
```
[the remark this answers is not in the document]
ian-r-rose: Yeah, I copied this test over there. I could remove it here, but maybe useful to have some level of duplication?

jrbourbeau:
Just wanted to point out that the CI is starting to fail here (xref https://github.com/dask/distributed/pull/4467/checks?check_run_id=1786579413) because of the updates in https://github.com/dask/dask/pull/7102. Merging this PR will resolve the 

```python
AttributeError: type object 'Layer' has no attribute 'expand_annotations'
```

issue

jrbourbeau (approved):
Thanks @ian-r-rose for working on this and @sjperkins @madsbk @mrocklin for reviewing!

ian-r-rose:
Thanks for the reviews, all!

sjperkins:
Thanks @ian-r-rose  and congratulations on treading a path through all the discussion :-D

Issue #4395, opened 2020-12-29
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte


haraldschilly:
**What happened**: crash with msgpack 0.6.2 – the `requirements.txt` file specifies minimum 0.6, so, it should work. Maybe it's too low, IDK.

**What you expected to happen**: no crash

**Minimal Complete Verifiable Example**:

```python
from dask.distributed import Client, LocalCluster
import dask.array as da
cluster = LocalCluster(
    n_workers=1,
    threads_per_worker=1,
    processes=True,
    diagnostics_port=None,
)
client = Client(cluster)
x = da.random.random((300, 300), chunks=(10, 10))
y = x + x.T
z = (y.mean(axis=1) / y.shape[0]).std()
print(z.compute())
```

causing: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte`

```
distributed.core - ERROR - 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte                                   
Traceback (most recent call last):                                                                                                  
  File "/usr/local/lib/python3.8/dist-packages/distributed/core.py", line 589, in handle_stream                                     
    handler(**merge(extra, msg))                                                                                                    
  File "/usr/local/lib/python3.8/dist-packages/distributed/scheduler.py", line 2286, in update_graph_hlg                            
    dsk, dependencies, annotations = highlevelgraph_unpack(hlg)                                                                     
  File "/usr/local/lib/python3.8/dist-packages/distributed/protocol/highlevelgraph.py", line 125, in highlevelgraph_unpack          
    hlg = msgpack.loads(                                                                                                            
  File "msgpack/_unpacker.pyx", line 201, in msgpack._cmsgpack.unpackb                                                              
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte                                          
distributed.core - ERROR - Exception while handling op register-client                                                              
Traceback (most recent call last):                                                                                                  
  File "/usr/local/lib/python3.8/dist-packages/distributed/core.py", line 517, in handle_comm                                       
    result = await result                                                                                                           
  File "/usr/local/lib/python3.8/dist-packages/distributed/scheduler.py", line 3149, in add_client                                  
    await self.handle_stream(comm=comm, extra={"client": client})                                                                   
[...]

PR #4407, opened 2021-01-06
Set use_bin_type=True in highlevelgraph_pack

jrbourbeau:
This PR ensures we use `use_bin_type=True` in `highlevelgraph_pack`. For `msgpack` >= 1 this is the default behavior, so doesn't need to be explicitly specified. However older versions of `msgpack` use `use_bin_type=False` by default, which was causing serialization errors.

cc @haraldschilly @bnaul 

Closes https://github.com/dask/distributed/issues/4395

### DIFF
--- a/distributed/protocol/highlevelgraph.py
+++ b/distributed/protocol/highlevelgraph.py
@@ -127,7 +127,9 @@ def highlevelgraph_pack(hlg: HighLevelGraph, client, client_keys):
             }
         )
 
-    return msgpack.dumps({"layers": layers}, default=msgpack_encode_default)
+    return msgpack.dumps(
+        {"layers": layers}, use_bin_type=True, default=msgpack_encode_default
+    )
 
 
 def _materialized_layer_unpack(state, dsk, dependencies, annotations):


distributed/protocol/highlevelgraph.py
```diff
@@ -127,7 +127,9 @@ def highlevelgraph_pack(hlg: HighLevelGraph, client, client_keys):
             }
         )
 
-    return msgpack.dumps({"layers": layers}, default=msgpack_encode_default)
+    return msgpack.dumps(
+        {"layers": layers}, use_bin_type=True, default=msgpack_encode_default
+    )
```
jakirkham, line 132: Should we just use [`dumps_msgpack`]( https://github.com/dask/distributed/blob/7fb5d4d691df91cb08702509aeebc366bac6f5a0/distributed/protocol/core.py#L175 )?

jakirkham:
Thanks for looking into this James! 😄

Had one thought above. Cc-ing @madsbk in case he has more thoughts 🙂

distributed/protocol/highlevelgraph.py
madsbk: Yes great point. 
Initially, I didn't use `dumps_msgpack` because I was using a new ` default=` instead of `msgpack_encode_default`. Later in the development I rolled back to using `msgpack_encode_default` thus using `dumps_msgpack` now makes a lot of sense!

madsbk (changes requested):
Good catch @jrbourbeau, I didn't know that they changed the default. 

I agree with @jakirkham that we should use [`dumps_msgpack`](https://github.com/dask/distributed/blob/7fb5d4d691df91cb08702509aeebc366bac6f5a0/distributed/protocol/core.py#L175) and [`loads_msgpack`](https://github.com/dask/distributed/blob/7fb5d4d691df91cb08702509aeebc366bac6f5a0/distributed/protocol/core.py#L198) in `highlevelgraph_pack` and `highlevelgraph_unpack` to make the code more simple.

We might even set `compression=dask.config.get("distributed.comm.compression")` ?

jacobtomlinson:
@madsbk do you have time to raise an alternative PR?

madsbk:
> @madsbk do you have time to raise an alternative PR?

Sure,  https://github.com/dask/distributed/pull/4409

jrbourbeau:
Closing in favor of https://github.com/dask/distributed/pull/4409

PR #4408, opened 2021-01-07
Type `TaskGroup` in `active_states`

jakirkham:
This was missing a type annotation before, which results in issues when accessing `_states` when compiled. This fixes that issue.

cc @quasiben


jakirkham:
This was the only method in `TaskPrefix` missing this annotation. The rest already include it.

jakirkham:
Merging this minor change so that we have the fix in. Hope that is ok 🙂

PR #4425, opened 2021-01-13
Add time started to scheduler info

jacobtomlinson:
It is useful from a cluster lifecycle management perspective to include the scheduler start time in the scheduler info.

Extracted from #4291

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1963,6 +1963,7 @@ class Scheduler(ServerNode):
             "id": str(self.id),
             "address": self.address,
             "services": {key: v.port for (key, v) in self.services.items()},
+            "time_started": self.time_started,
             "workers": {
                 worker.address: worker.identity() for worker in self.workers.values()
             },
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -1963,6 +1963,7 @@ def identity(self, comm=None):
             "id": str(self.id),
             "address": self.address,
             "services": {key: v.port for (key, v) in self.services.items()},
+            "time_started": self.time_started,
```
mrocklin, line 1966: Maybe just `"started"` to make things shorter?  Definitely not something I care a ton about though.

distributed/tests/test_client.py
```diff
@@ -3789,6 +3789,7 @@ def test_scheduler_info(c):
     info = c.scheduler_info()
     assert isinstance(info, dict)
     assert len(info["workers"]) == 2
+    assert "time_started" in info
```
mrocklin, line 3792: Maybe a type check instead

jacobtomlinson:
Thanks for the review @mrocklin. Addressed both points.

jrbourbeau (approved):
Thanks @jacobtomlinson!

PR #4426, opened 2021-01-13
Give clusters names

jacobtomlinson:
This PR gives all `Cluster` objects and subclasses a unique name.

This name can be provided with the `name` kwarg, if this is omitted it will default to a UUID.

I have also stubbed out a `from_name` class method in `SpecCluster` with the intention that cluster objects that can be reconstructed by name (Cloud, Kubernetes, etc) should implement this method.

I've also added the `shutdown_on_close` argument to `SpecCluster` as proposed by @mrocklin as creating a cluster from name requires the ability to detach from the cluster in the first place.

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -4,6 +4,7 @@ from contextlib import suppress
 import logging
 import threading
 import warnings
+import uuid
 from tornado.ioloop import PeriodicCallback
 
 import dask.config
@@ -50,8 +51,9 @@ class Cluster:
     """
 
     _supports_scaling = True
+    name = str(uuid.uuid4())[:8]
 
-    def __init__(self, asynchronous, quiet=False):
+    def __init__(self, asynchronous, quiet=False, name=None):
         self.scheduler_info = {"workers": {}}
         self.periodic_callbacks = {}
         self._asynchronous = asynchronous
@@ -61,6 +63,8 @@ class Cluster:
         self.quiet = quiet
         self.scheduler_comm = None
 
+        if name is not None:
+            self.name = name
         self.status = Status.created
 
     async def _start(self):
@@ -424,8 +428,9 @@ class Cluster:
         return getattr(self, "_name", type(self).__name__)
 
     def __repr__(self):
-        text = "%s(%r, workers=%d, threads=%d" % (
+        text = "%s(%s, %r, workers=%d, threads=%d" % (
             self._cluster_class_name,
+            self.name,
             self.scheduler_address,
             len(self.scheduler_info["workers"]),
             sum(w["nthreads"] for w in self.scheduler_info["workers"].values()),
@@ -449,3 +454,9 @@ class Cluster:
     @property
     def observed(self):
         return {d["name"] for d in self.scheduler_info["workers"].values()}
+
+    def __eq__(self, other):
+        return self.name == other.name
+
+    def __hash__(self):
+        return id(self)
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -51,7 +51,7 @@ class Cluster:
     """
 
     _supports_scaling = True
-    name = str(uuid.uuid4())[:8]
+    name = None
 
     def __init__(self, asynchronous, quiet=False, name=None):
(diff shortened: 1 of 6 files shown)


distributed/deploy/local.py
```diff
@@ -94,6 +94,7 @@ class LocalCluster(SpecCluster):
 
     def __init__(
         self,
+        name=None,
```
jrbourbeau, line 97: The position of this keyword is breaking some tests which specify `n_workers` by positions (e.g. `distributed/deploy/tests/test_local.py::test_scale_up_and_down`)

```
    async with LocalCluster(
        0,
        scheduler_port=0,
        processes=False,
        silence_logs=False,
        dashboard_address=None,
        asynchronous=True,
    ) as cluster:
```

Perhaps we should start switching some classes to use keyword-only arguments in their constructors to avoid this in the future?

jacobtomlinson: Thanks @jrbourbeau. I think that would be sensible.

I'm happy to go through and update the tests to use explicit kwargs. Perhaps one day when our minimum Python version is 3.8 we could use the PEP570 syntax and disable positional args on many classes altogether.

jacobtomlinson:
Failures here are now unrelated. Ready for final review/merge @dask/maintenance.

distributed/deploy/cluster.py
```diff
@@ -50,8 +51,9 @@ class Cluster:
     """
 
     _supports_scaling = True
+    name = str(uuid.uuid4())[:8]
```
martindurant, line 54: So each instance of an unnamed cluster gets the same name - why is that useful? This is then used for equality, but I would have thought it takes much more than that for two clusters to be considered equal; I *do* agree on equality in the case that the name is meaningful.

Wouldn't it be better to leave it as `None` when the name is not meaningful, and use `hash()` for equality in this case?

jacobtomlinson: > So each instance of an unnamed cluster gets the same name

Each instance of an unnamed cluster is given a unique name. The idea is that you can attach to an existing cluster by name. 

It may help to consider a cluster as more than the Python class which represents it. In many cases it is a collection of resources created in some other infrastructure (pods on kubernetes for example). Giving all of these things a common cluster name allows you to implement discovery downstream where we look at all the bits of an existing cluster, then construct a Python object to represent those.

martindurant: > Each instance of an unnamed cluster is given a unique name.

Shouldn't this line be in __init__ in that case? 

Adding the following fails the test
```diff
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -1059,3 +1059,5 @@ async def test_cluster_names():
             assert unnamed_cluster == unnamed_cluster
             assert named_cluster == named_cluster
             assert unnamed_cluster != named_cluster
+        async with LocalCluster(processes=False, asynchronous=True) as unnamed_cluster2:
+            assert unnamed_cluster2 != unnamed_cluster
```

jacobtomlinson: Ah good catch thanks. I've updated the test and code.

martindurant:
Failure is flakiness on Travis?

martindurant:
Otherwise +1 from me

jacobtomlinson:
Failures are consistent with flakiness.

Thanks for the review @martindurant. Merging.

distributed/deploy/local.py
mrocklin: Others may be using `LocalCluster(n_workers)` in the wild.  This may cause some user frustration. 

Thoughts on moving the `name=` parameter to later in this list of parameters?

distributed/deploy/cluster.py
```diff
@@ -449,3 +456,9 @@ def requested(self):
     @property
     def observed(self):
         return {d["name"] for d in self.scheduler_info["workers"].values()}
+
+    def __eq__(self, other):
+        return self.name == other.name
+
+    def __hash__(self):
+        return id(self)
```
mrocklin, line 464: This caused me to think for a bit, but I convinced myself that it was probably ok.

PR #4431, opened 2021-01-15
Set port 8786 as default port for scheduler

### LINKED ISSUE #4429: LocalCluster not using default ports
**What happened**:

When creating a `LocalCluster` object the comm is started on a random high port, even if there are no other clusters running.

Should use port `8786`.

```console
$ conda create -n dask-lc-test -c conda-forge -y python=3.8 ipython dask distributed
```

The `dask-scheduler` command works as expeted.

```console
$ dask-scheduler
distributed.scheduler - INFO -   Scheduler at:   tcp://10.51.100.43:8786
distributed.scheduler - INFO -   dashboard at:                     :8787
```

But `LocalCluster` does not.

```python
>>> from dask.distributed import LocalCluster
>>> cluster = LocalCluster()
>>> cluster.scheduler_address
'tcp://127.0.0.1:45083'
```

- Dask version: 2020.12.0 (tried latest source too)
- Python version: 3.8
- Operating System: macOS and Ubuntu 18.04
- Install method (conda, pip, source): conda and source

FabioRosado:
Hello,

This is my attempt at a fix #4429.  I followed Benjamin suggestion and changed the port number in the constructor, I've also added a tiny (silly) test to check if the port is the right one 😄 

Let me know if you would like me to change something 🤔

### DIFF
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -1045,3 +1045,10 @@ async def test_no_workers(cleanup):
         n_workers=0, silence_logs=False, dashboard_address=None, asynchronous=True
     ) as c:
         pass
+
+
+@pytest.mark.asyncio
+async def test__default_scheduler_port():
+    async with LocalCluster(n_workers=2, processes=False, asynchronous=True) as cluster:
+
+        assert cluster.scheduler_port == 8786
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -16,6 +16,7 @@ import pytest
 
 from dask.system import CPU_COUNT
 from distributed import Client, Worker, Nanny, get_client
+from distributed.comm.addressing import parse_host_port
 from distributed.core import Status
 from distributed.deploy.local import LocalCluster
 from distributed.metrics import time
@@ -1048,7 +1049,8 @@ async def test_no_workers(cleanup):
 
 
 @pytest.mark.asyncio
-async def test__default_scheduler_port():
+async def test_default_scheduler_port():
     async with LocalCluster(n_workers=2, processes=False, asynchronous=True) as cluster:
 
-        assert "8786" in cluster.scheduler_address
+        _, port = parse_host_port(cluster.scheduler_adress)
+        assert port == 8786
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -1052,5 +1052,5 @@ async def test_no_workers(cleanup):
 async def test_default_scheduler_port():
     async with LocalCluster(n_workers=2, processes=False, asynchronous=True) as cluster:
 
-        _, port = parse_host_port(cluster.scheduler_adress)
+        _, port = parse_host_port(cluster.scheduler_address)
         assert port == 8786
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -122,7 +122,7 @@ def test_move_unserializable_data():
     transports.
     """
     with LocalCluster(
-        processes=False, silence_logs=False, dashboard_address=None, scheduler_port=0
+        processes=False, silence_logs=False, dashboard_address=None
     ) as cluster:
         assert cluster.scheduler_address.startswith("inproc://")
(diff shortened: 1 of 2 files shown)


distributed/deploy/tests/test_local.py
```diff
@@ -1045,3 +1045,10 @@ async def test_no_workers(cleanup):
         n_workers=0, silence_logs=False, dashboard_address=None, asynchronous=True
     ) as c:
         pass
+
+
+@pytest.mark.asyncio
+async def test__default_scheduler_port():
+    async with LocalCluster(n_workers=2, processes=False, asynchronous=True) as cluster:
+
+        assert cluster.scheduler_port == 8786
```
jrbourbeau, line 1054: `LocalCluster` doesn't have a `scheduler_port` attribute. Though it does have a `scheduler_address` attribute from which we can extract the port using `parse_host_port` in `distributed/comm/addressing.py` with something like 

```diff
diff --git a/distributed/deploy/tests/test_local.py b/distributed/deploy/tests/test_local.py
index 79ae3a7f..fc034ab0 100644
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -16,6 +16,7 @@

 from dask.system import CPU_COUNT
 from distributed import Client, Worker, Nanny, get_client
+from distributed.comm.addressing import parse_host_port
 from distributed.core import Status
 from distributed.deploy.local import LocalCluster
 from distributed.metrics import time
@@ -1050,5 +1051,5 @@ async def test_no_workers(cleanup):
 @pytest.mark.asyncio
 async def test__default_scheduler_port():
     async with LocalCluster(n_workers=2, processes=False, asynchronous=True) as cluster:
-
-        assert cluster.scheduler_port == 8786
+        _, port = parse_host_port(cluster.scheduler_address)
+        assert port == 8786
```

distributed/deploy/tests/test_local.py
```diff
@@ -1045,3 +1045,10 @@ async def test_no_workers(cleanup):
         n_workers=0, silence_logs=False, dashboard_address=None, asynchronous=True
     ) as c:
         pass
+
+
+@pytest.mark.asyncio
+async def test__default_scheduler_port():
```
jrbourbeau, line 1051: Small nitpick

```suggestion
async def test_default_scheduler_port():
```

jrbourbeau (commented):
Thanks @FabioRosado! Overall this looks good, looking forward to seeing this merged

distributed/deploy/tests/test_local.py
FabioRosado: Oops, I forgot to push the changes that I've done to that test, but your suggestion looks way better than what I had (I was checking if 8786 were in `cluster.scheduler_address` 😬  I'll update this PR in a second

FabioRosado:
This last push I added a `scheduler_port=0` to the failing tests and they pass on windows. This makes me feel that perhaps we need to do something else to set 8687 as default port, looking at the failing tests I'm not sure what though 🤔

jrbourbeau:
Thanks for the updates @FabioRosado, this is looking close! Regarding explicitly passing `scheduler_port=0`, in some cases (i.e. no explicit scheduler port is requested for and `processes` is `False`) we want to automatically select the `inproc` protocol over `tcp`. The relevant logic is here:

https://github.com/dask/distributed/blob/9442d9b3f2847bf6d0252a8ed671d342a5379501/distributed/deploy/local.py#L164-L165

Now that we're using `8687` for the default scheduler port, the `and not scheduler_port` logic is no longer working and we're not setting `protocol = "inproc://"` as we would like to. 

Instead of using `8687` as the default `scheduler_port` and explicitly passing `scheduler_port=0` we might update the default `scheduler_port` to be `None` and then add some additional logic to assign it `8687` or `0` depending on whether or we select the `tcp` or `inproc` protocol, respectively.

FabioRosado:
I thought about this issue a bit more and the failing test (` test_Client_twice`). I think that we could take one of the two approaches here:

- Decide that the port 8786 is the default one and if we try to run two clients the second one will fail while trying to connect to the default port (so we need to use `port=0`)
- Replace the default port to be whatever `port=0` assigns and update the code that says that the cluster is being run on 8786

Issue #4430, opened 2021-01-14
Profile Cythonization work


mrocklin:
We've been optimizing the scheduler for larger workloads recently.  We've found that profiling the scheduler is challenging to do well.  A lot of the recent profiling has happened on NVIDIA systems, which may or may not be representative of typical hardware.

I recently ran a small experiment on AWS with the following code:

```python
import coiled
coiled.create_software_environment(
    "dev", 
    conda=["python=3.8", "dask", "lz4", "python-blosc"], 
    pip=["git+https://github.com/dask/dask", "git+https://github.com/dask/distributed"]
)

cluster = coiled.Cluster(n_workers=100, software="dev")

from dask.distributed import Client, performance_report
client = Client(cluster)

import dask
import dask.dataframe as dd
df = dask.datasets.timeseries(start="2020-01-01", end="2020-12-31", partition_freq="1h", freq="60s")
df2 = dd.shuffle.shuffle(df, "x")
with performance_report(filename="report.html"):
    df3 = df2.sum().compute()
```

I found that

1. The scheduler CPU was pegged at 100%
2. The majority of time was spent in communication
3. The computation actually didn't finish, I suspect because of the worker TaskState issue in 2020.12.0, but I can't be sure

It would be good to try a few things here:

1. Try bringing in @fjetter's recent PR https://github.com/dask/distributed/pull/4360 to see if it resolves the deadlocks
2. Try compiling with Cython and seeing if that helps at all, this will probably require making a docker image that installs `distributed` with the appropriate flag (cc @jakirkham do we have instructions for this somewhere?) and using that instead of a Coiled-built image 
3. Actually produce and publish performance reports and publish them on this issue

After that we need to consider what to do about TLS communication performance.  Is this just because we're using TLS rather than TCP?  If so, is there anything that we can do to accelerate this?  Maybe asyncio is faster?  Maybe Tornado can be improved?  

NVIDIA devs @quasiben and @jakirkham report that UCX doesn't have this problem (although that may be because it's hard to profile).

jakirkham:
> The computation actually didn't finish, I suspect because of the worker TaskState issue in 2020.12.0, but I can't be sure

Do you have a link for this issue?

mrocklin:
https://github.com/dask/distributed/pull/4360

mrocklin:
[...]

PR #4432, opened 2021-01-15
Set runspec on all new tasks to avoid deadlocks

fjetter:
This is a more minimalistic version of #4360 with fixes I am pretty confident with. I still can see local issue, though and this is incomplete.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -450,6 +450,9 @@ class Worker(ServerNode):
             ("waiting", "flight"): self.transition_waiting_flight,
             ("ready", "executing"): self.transition_ready_executing,
             ("ready", "memory"): self.transition_ready_memory,
+            ("ready", "error"): self.transition_ready_error,
+            ("ready", "waiting"): self.transition_ready_waiting,
+            ("constrained", "waiting"): self.transition_ready_waiting,
             ("constrained", "executing"): self.transition_constrained_executing,
             ("executing", "memory"): self.transition_executing_done,
             ("executing", "error"): self.transition_executing_done,
@@ -1445,8 +1448,10 @@ class Worker(ServerNode):
         **kwargs2,
     ):
         try:
+            runspec = SerializedTask(function, args, kwargs, task)
             if key in self.tasks:
                 ts = self.tasks[key]
+                ts.runspec = runspec
                 if ts.state == "memory":
                     assert key in self.data or key in self.actors
                     logger.debug(
@@ -1475,6 +1480,7 @@ class Worker(ServerNode):
             if actor:
                 self.actors[ts.key] = None
 
+            ts.runspec = runspec
             ts.priority = priority
             ts.duration = duration
             if resource_restrictions:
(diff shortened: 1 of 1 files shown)


mrocklin:
I notice that this is still in WIP status.  Are the changes here an improvement over what is in master?  If so, I might press the "Ready for review" button

gforsyth:
I'll note that the failures in `test_failed_workers.py::test_worker_who_has_clears_after_failed_connection` are connected to work stealing -- if stealing is disabled the assertion errors go away

fjetter:
I'm currently testing the latest commit on a few clusters in parallel. So far, no funny logs, no deadlocks, no problems at all.

fjetter:
The only thing I could observe, so far is that the waiting_for_data_count is off but I checked the workers, both data and tasks are empty and I suspected a bug in this counter already. I believe this is not critical, though, since the counter is not used internally, merely for instrumentation

image

mrocklin:
That's great news

fjetter:
Needed to push another time. There was actually a typo on my side causing a test to fail.

fjetter:
Unfortunately, I needed to push another small change but an expected one where we didn't clear `who_has` on the task state during `release_key`. That's not a big deal, I believe, and my tests so far showed this works still nicely.

I'm dropping out now but if there is no big CI thing popping up again, I feel confident merging this now. I'm not sure if it fixes all issues but it should provide a significant improvement to master from what I've seen.

mrocklin:
@fjetter are there other issues that you know exist?  If so I'm happy to take a look at them today.

distributed/worker.py
```diff
@@ -1691,9 +1699,26 @@ def transition_ready_executing(self, ts):
...
         self.send_task_state_to_scheduler(ts)
 
+    def transition_ready_waiting(self, ts):
+        """
+        This transition is common for work stealing
+        """
+        ts.state = "waiting"
+        ts.runspec = None
+
+        return ts.state
```
mrocklin, line 1720: Do we need to transition to waiting in this case?  I'm curious if this was necessary before due to the transition when we were releasing keys, but maybe it's not necessary now?

mrocklin: Also, in both cases where we call this the final state is "waiting".  Do we need to set the state explicitly, or can we let the `transition` function handle this?

fjetter:
On the other PR I originally wanted to fix an issue with erroneous tasks which led me to discover this runspec thing. I added a test case with an object with broken deserialization. That should hit the worker internal suspicious counter (but it doesn't/didn't) something else might reveal itself there. Otherwise I don't have anything on my mind right now

distributed/worker.py
fjetter: This transition is probably not necessary. The only case I am aware of where we hit this is work stealing.

mrocklin:
OK, once tests pass here I'm going to merge.  If things are ok for a bit I suggest that we release.  cc @jrbourbeau

mrocklin:
I briefly tried things with the exercise in #4430 and things didn't complete. I need to double check that work though.

fjetter:
I had already a few successful runs of the example in #4430 and that worked out great. That was using an earlier commit of this PR and I'll double check with master now.

Really happy to see this merged. Thanks for your help @mrocklin !

mrocklin:
Ah, indeed, things with that workflow are good for me too as well now.  Good for a release whenever makes sense @jrbourbeau .  I suspect that you're aiming for this Friday.

jrbourbeau:
xref https://github.com/dask/community/issues/121

PR #4433, opened 2021-01-17
Don't require network to inspect tests

mrocklin:
Previously the get_ipv6 function was called at import time in test_core.py
This stopped any tests from running if a network adapter was not
attached

The results of this function were used only in one test,
so we move this call within that test and skip if it fails

### DIFF
--- a/distributed/tests/test_core.py
+++ b/distributed/tests/test_core.py
@@ -42,11 +42,6 @@ from distributed.utils_test import (
 from distributed.utils_test import loop  # noqa F401
 
 
-EXTERNAL_IP4 = get_ip()
-if has_ipv6():
-    EXTERNAL_IP6 = get_ipv6()
-
-
 def echo(comm, x):
     return x
 
@@ -194,6 +189,15 @@ async def test_server_listen():
     """
     Test various Server.listen() arguments and their effect.
     """
+    import socket
+
+    try:
+        EXTERNAL_IP4 = get_ip()
+        if has_ipv6():
+            EXTERNAL_IP6 = get_ipv6()
+    except socket.gaierror:
+        raise pytest.skip(reason="no network access")
+
     from contextlib import asynccontextmanager
 
     @asynccontextmanager


jrbourbeau (commented):
Thanks @mrocklin! 

We also do something similar over in `distributed/distributed/comm/tests/test_comms.py`

https://github.com/dask/distributed/blob/467bbc2c7d1c23328571a418f4d19f0110a3c18c/distributed/comm/tests/test_comms.py#L43-L47

If you want to update there too, that'd be welcome. Though this PR is a definite improvement over our current behavior and can be merged as is too

distributed/tests/test_core.py
```diff
@@ -194,6 +189,15 @@ async def test_server_listen():
...
     Test various Server.listen() arguments and their effect.
     """
+    import socket
+
+    try:
+        EXTERNAL_IP4 = get_ip()
+        if has_ipv6():
+            EXTERNAL_IP6 = get_ipv6()
+    except socket.gaierror:
+        raise pytest.skip(reason="no network access")
```
bnavigator, line 199: [`pytest.skip`](https://docs.pytest.org/en/latest/reference.html#pytest-skip) does not have a `reason` kwarg. Just use `raise pytest.skip("no network access")` 

```
[  346s]     @pytest.mark.asyncio
[  346s]     async def test_server_listen():
[  346s]         """
[  346s]         Test various Server.listen() arguments and their effect.
[  346s]         """
[  346s]         import socket
[  346s]     
[  346s]         try:
[  346s]             EXTERNAL_IP4 = get_ip()
[  346s]             if has_ipv6():
[  346s]                 EXTERNAL_IP6 = get_ipv6()
[  346s]         except socket.gaierror:
[  346s] >           raise pytest.skip(reason="no network access")
[  346s] E           TypeError: skip() got an unexpected keyword argument 'reason'
[  346s] 
[  346s] distributed/tests/test_core.py:199: TypeError

```

jrbourbeau: Thanks @bnavigator! See https://github.com/dask/distributed/pull/4467

Issue #4413, opened 2021-01-08
Internal worker state transitions


fjetter:
In #4360 various deadlock situations are investigated where some of them are connected to the way valid states and state transitions of the **worker**[1] `TaskState` objects are defined.

The [current documentation](https://distributed.dask.org/en/latest/worker.html#internal-scheduling) is slightly outdated since dependencies and runnable tasks where consolidated in https://github.com/dask/distributed/pull/4107. The current state transitions are now following the pipeline (omitting `long-running`, `error`, `constrained` for the sake of clarity [2])

![worker-state-master](https://user-images.githubusercontent.com/8629629/104005407-599f5500-51a5-11eb-84ff-4478b97afdc5.png)

What's remarkable about this transition pipeline is that virtually all states allow a transition to `memory` and there are multiple allowed transition paths which are only allowed in very specific circumstances and only upon intervention via the scheduler. For instance, a task in state `flight` may be transitioned via `ready` to `executing` but this is only possible if the worker actually possess the knowledge about how to execute a given task, i.e. the `TaskState` object possesses a set attribute `runspec`. This attribute is usually only known to the worker if the scheduler intents for this worker to actually execute the task. This transition path is, however, allowed since a _dependency_, i.e. a task without the knowledge of how to compute it, is reassigned by the scheduler for computation on this worker. This may happen if the worker where the task was intended to be computed on originally is shut down.

This ambiguity is essentially introduced by not distinguishing between dependencies and runnable tasks anymore. What I would propose is to make this distinction explicit in the `state` of the tasks. Consider the following pipeline 

![worker-state-proposed](https://user-images.githubusercontent.com/8629629/104005414-5c01af00-51a5-11eb-857c-26a5acda0f77.png)

Every task starts off in `new`. This is effectively a dummy state and could be omitted. It represents a known task which hasn't been classified into "can the worker compute this task or not". Based on the answer of this question it is put into the states 

`waiting_for_dependencies` : This task is intended to be computed by this worker. Once all dependencies are available on *this* worker, it is supposed to be transitioned to `ready` to be queued up for execution. (No dependencies is a special subset of this case)

`waiting_to_fetch` : This task is *not* intended to be computed on this worker but the `TaskState` on this worker merely is a reference to a remote data key we are about to fetch.

The red transition is only possible via scheduler interference once the scheduler reassigns a task to be computed on this worker. This is relatively painless as long as the `TaskState` is in a valid state (in particular `runspec` is set)

[...]

PR #4440, opened 2021-01-19
Update TaskState documentation about dependents attribute

fjetter:
I think this mechanism should be refined but currently `dependents` is a dynamic data structure and this should be reflected in the docs.

For dependencies we have already a dynamic equivalent with `waiting_for_data` and I believe we should establish a similar mechanism for the dependents since this is primarily used to determine whether or not a key is allowed to be forgotten. In the meantime, a more accurate docstring should help

xref #4413

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -104,7 +104,10 @@ class TaskState:
     * **dependencies**: ``set(TaskState instances)``
         The data needed by this key to run
     * **dependents**: ``set(TaskState instances)``
-        The keys that use this dependency
+        The keys that use this dependency. Only keys which are not available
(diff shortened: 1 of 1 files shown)


fjetter:
The scheduler uses the attribute `waiters` for this dynamic set of dependencies and applies otherwise the same mechanism

jakirkham:
IIUC Gil this is code that you have been working with lately. Maybe you have thoughts on this change? 🙂

gforsyth (approved):
This looks good and accurately describes the current behavior of `dependents`

jakirkham:
Thanks Florian for the PR and Gil for the review! 😄

PR #4441, opened 2021-01-19
Performance Report Additions

quasiben:
PR adds a date to the report and uses `self.n_tasks` when reporting number tasks.  Previously, the report used `len(self.get_task_stream(start=start))` which has a maximum of 100000

## Example Output
 
Screen Shot 2021-01-19 at 11 24 16 AM

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3,7 +3,7 @@ from collections import defaultdict, deque
 
 from collections.abc import Mapping, Set
 from contextlib import suppress
-from datetime import timedelta
+from datetime import timedelta, date
 from functools import partial
 import inspect
 import itertools
@@ -5952,7 +5952,7 @@ class Scheduler(ServerNode):
 
         # Task stream
         task_stream = self.get_task_stream(start=start)
-        total_tasks = len(task_stream)
+        total_tasks = self.n_tasks
         timespent = defaultdict(int)
         for d in task_stream:
             for x in d.get("startstops", []):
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -5952,7 +5952,7 @@ def profile_to_figure(state):
 
         # Task stream
         task_stream = self.get_task_stream(start=start)
-        total_tasks = len(task_stream)
+        total_tasks = self.n_tasks
```
mrocklin, line 5955: Ideally the performance report would capture only the number of tasks during execution, rather than for the lifetime of the scheduler.


So for example in the following situation we would want only the number of tasks from `y.compute()` not also `x.compute()`

```python
x.compute()

with performance_report():
    y.compute()
```

quasiben: good point!

quasiben:
While exploring this it's now occurring to me that with the deque and max of 100_000 for the task stream, we're probably not calculating timings correctly when exploring large graphs and _not_ changing the max:
https://github.com/dask/distributed/blob/9442d9b3f2847bf6d0252a8ed671d342a5379501/distributed/scheduler.py#L5957-L5959

mrocklin:
Yeah, in the future I think that we need to use tasks less and TaskGroups
more.  These provide aggregation at a good level where we can accumulate a
lot more information a lot more cheaply

quasiben:
Closing.  As @jrbourbeau said:

> The total number of tasks is capped at 100k. I think this is to limit the total amount of memory taken up on the scheduler by the data that goes into the task stream.

There is a `distributed.scheduler.dashboard.tasks.task-stream-length` configuration value

https://github.com/dask/distributed/blob/42e3f22ef1d9d3dcb494bd4e546f9ca6f2d54891/distributed/distributed.yaml#L37

which allows you to increase that 100k limit

PR #4445, opened 2021-01-20
Additional optimizations to stealing

jakirkham:
Annotates variables in `reevaluate_occupancy` to improve efficiency there. As well as delaying construction of values that may not be needed.

Further optimizes `put_key_in_stealable` by only retrieving `ws` and `worker` if `cost_multiplier` is not `None`, which is the only case where they are used. Avoiding grabbing them otherwise.

Finally simplify check for executing task in `set_duration_estimate` by just using `.get(...)` instead of checking for the key and then getting the value.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6106,7 +6106,6 @@ class Scheduler(ServerNode):
         lets us avoid this fringe optimization when we have better things to
         think about.
         """
-        DELAY = 0.1
         try:
             if self.status == Status.closed:
                 return
@@ -6131,7 +6130,7 @@ class Scheduler(ServerNode):
                         next_time = timedelta(seconds=duration * 5)  # 25ms gap
                         break
             else:
-                next_time = timedelta(seconds=DELAY)
+                next_time = timedelta(seconds=0.1)
 
             self.loop.add_timeout(
                 next_time, self.reevaluate_occupancy, worker_index=worker_index
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4831,10 +4831,9 @@ class Scheduler(ServerNode):
         duration: double = self.get_task_duration(ts)
         comm: double = self.get_comm_cost(ts, ws)
         total_duration: double = duration + comm
-        if ts in ws._executing:
-            exec_time: double = ws._executing[ts]
-            if exec_time > 2 * duration:
-                total_duration = 2 * exec_time
+        exec_time: double = ws._executing.get(ts, 0)
+        if exec_time > 2 * duration:
+            total_duration = 2 * exec_time
         ws._processing[ts] = total_duration
         return total_duration
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6111,9 +6111,8 @@ class Scheduler(ServerNode):
             if self.status == Status.closed:
                 return
 
-            last = time()
-
             if self.proc.cpu_percent() < 50:
+                last = time()
                 workers = list(self.workers.values())
                 for i in range(len(workers)):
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -6111,13 +6111,13 @@ def reevaluate_occupancy(self, worker_index=0):
...
                 return
 
-            last = time()
-            next_time = timedelta(seconds=DELAY)
-
             if self.proc.cpu_percent() < 50:
-                workers = list(self.workers.values())
-                for i in range(len(workers)):
-                    ws: WorkerState = workers[worker_index % len(workers)]
+                last = time()
```
jakirkham, line 6115: FWIW this takes `~15us`. So is insignificant at the level milliseconds, which is what we compare `duration` with later.

```python
In [1]: import psutil

In [2]: proc = psutil.Process()

In [3]: %%timeit
   ...: if proc.cpu_percent() < 50:
   ...:     pass
   ...: 
14.9 µs ± 139 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

distributed/stealing.py
```diff
@@ -74,17 +74,16 @@ def transition(
         ts = self.scheduler.tasks[key]
         if finish == "processing":
             self.put_key_in_stealable(ts)
-
-        if start == "processing":
+        elif start == "processing":
```
jakirkham, line 77: IIUC these can't both be true. So changed this to `elif`

fjetter: Indeed 
https://github.com/dask/distributed/blob/e98d57dcad72d4a87fb177c4b2b9c98b2b2c50e1/distributed/scheduler.py#L5518-L5519

out of curiosity, is `elif` faster?

fjetter (approved):

distributed/stealing.py
jakirkham: If one matches the first case initially, yes. Though admittedly it is not a huge difference

```python
In [1]: %%timeit s = "a"
   ...: if s == "a":
   ...:     pass
   ...: elif s == "b":
   ...:     pass
   ...: 
   ...: 
20.9 ns ± 0.166 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [2]: %%timeit s = "a"
   ...: if s == "a":
   ...:     pass
   ...: if s == "b":
   ...:     pass
   ...: 
   ...: 
39.3 ns ± 0.172 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

distributed/scheduler.py
```diff
@@ -4828,13 +4828,14 @@ def set_duration_estimate(self, ts: TaskState, ws: WorkerState):
         estimate the task duration to be 2x current-runtime, otherwise we set it
         to be the average duration.
         """
+        exec_time: double = ws._executing.get(ts, 0)
```
quasiben, line 4831: Is it true that if `ts` is not in `ws._executing`, this will return 0 ?

jakirkham: The main point is that we don't want to enter the `if` below. `0` was just a sensible default. Though we could also consider `-1`, which we have used elsewhere (as that is also an obviously bogus value for time).

`duration` is guaranteed to be positive (though maybe there is a vanishingly small chance that it is `0`). In any event `exec_time > 2 * duration` would be `False` even if `exec_time` is `0`, which is what we are trying to accomplish here.

quasiben: Thanks @jakirkham

quasiben:
Looks good! Thanks @jak

jakirkham:
Thanks all! 😄

PR #4447, opened 2021-01-21
Make system monitor interval configurable

mrocklin:
Sometimes we want to reduce this to improve performance


jakirkham:
cc @quasiben (think you were looking for this knob recently 😉)

jakirkham (approved):
Thanks Matt! 😄

Issue #4376, opened 2020-12-18
Simplifying serialization of status messages


jakirkham:
In the Scheduler and to some extent the Workers, we send status messages that are MsgPack serializable. Though currently we go through all the same serialization steps we go through when serializing data, which adds a fair bit of overhead. It might be worth looking into a fast path for sending these status messages to improve overall message transmission runtime.

jakirkham:
Related to this I've been wondering if there is a way to incrementally build up a MsgPack serialization of an object. IOW if some part of a message lives in a `dict`, can we serialize that part first when we have it and use the serialized form to build the serialized `dict` later. Haven't seen an API like this in the MsgPack library we use, but maybe this exists out there somewhere. Though this is an additional optimization beyond what is mentioned in the OP.

jakirkham:
cc @mrocklin @quasiben

jakirkham:
Actually that's a different point than what is being articulated here. Instead it is that we do additional work on top of MsgPack that is causing the slow down. This likely can be avoided when sending simple status messages (IOW only using MsgPack).

The use of a different library instead of MsgPack may also help, but that is an orthogonal issue. The optimizations suggested here would still be valuable regardless of what library is used under-the-hood.

jakirkham:
cc @madsbk (since this came up in our discussion earlier)

mrocklin:
Quick note, for the scheduler in particular another option is to just not
call extract_serialize .  In principle it should never be used.

jakirkham:
Yep that's the idea 🙂

mrocklin:
It looks like we do include Serialized objects in messages in the scheduler.  These invariably just wrap lists of bytestrings, but they're certainly not msgpack serializable.  Custom msgpack protocols seem like the right next step to me.  I'm also trying out the Cython build now.

jakirkham:
Ah ok. Good to know.

For those following along custom msgpack protocols are outlined in issue ( https://github.com/dask/distributed/issues/4379 ).

jakirkham:
A fast path for MsgPack serializable messages is being handled in PR ( https://github.com/dask/distributed/pull/4480 )

PR #4449, opened 2021-01-22
[WIP] Build `serialize` with Cython

jakirkham:
Enables Cythonization of `serialize`, this should improve functions like `extract_serialize`, which have already been annotated for this purpose.

Note: Have sometimes encountered `ImportError`s locally when doing this. Plus we've only done the work to annotate `extract_serialize` and nothing else. Hence why this is disabled. In the end we may end up optimizing Scheduler communication to bypass this function ( https://github.com/dask/distributed/issues/4376 ). So this may not be needed in the end.

cc @quasiben @mrocklin @madsbk

### DIFF
--- a/distributed/protocol/serialize.py
+++ b/distributed/protocol/serialize.py
@@ -157,14 +157,16 @@ def serialization_error_loads(header, frames):
 families = {}
 
 
-def register_serialization_family(name, dumps, loads):
-    families[name] = (dumps, loads, dumps and has_keyword(dumps, "context"))
+def register_serialization_family(name, dumps, loads, context=None):
+    if context is None:
+        context = has_keyword(dumps, "context")
+    families[name] = (dumps, loads, dumps and context)
 
 
-register_serialization_family("dask", dask_dumps, dask_loads)
-register_serialization_family("pickle", pickle_dumps, pickle_loads)
-register_serialization_family("msgpack", msgpack_dumps, msgpack_loads)
-register_serialization_family("error", None, serialization_error_loads)
+register_serialization_family("dask", dask_dumps, dask_loads, True)
+register_serialization_family("pickle", pickle_dumps, pickle_loads, True)
+register_serialization_family("msgpack", msgpack_dumps, msgpack_loads, False)
+register_serialization_family("error", None, serialization_error_loads, False)
 
 
 def check_dask_serializable(x):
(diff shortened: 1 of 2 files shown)


mrocklin:
I tried playing around with this briefly.  Made a small change.  Still not working sadly :/

jakirkham:
Are you seeing an `ImportError` or something else?

mrocklin:
`Unable to serialize bytes`.  The serializers aren't working correctly.  I may have screwed up the context bit that I added.  I added that by the way because `inspect` no longer works on the cythonized functions

jakirkham:
Yeah maybe Cythonizing this whole file is overkill. Let me try something

mrocklin:
```python
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/distributed/batched.py", line 93, in _background_send
    nbytes = yield self.comm.write(
  File "/home/mrocklin/workspace/tornado/tornado/gen.py", line 766, in run
    value = future.result()
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/distributed/comm/tcp.py", line 230, in write
    frames = await to_frames(
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/distributed/comm/utils.py", line 52, in to_frames
    return _to_frames()
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/distributed/comm/utils.py", line 32, in _to_frames
    protocol.dumps(
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/distributed/protocol/core.py", line 50, in dumps
    data = {
  File "/home/mrocklin/miniconda/lib/python3.8/site-packages/distributed/protocol/core.py", line 51, in <dictcomp>
    key: serialize(
  File "distributed/protocol/serialize.py", line 318, in distributed.protocol.serialize.serialize
TypeError: ('Could not serialize object of type bytes.', '
```

I tried moving the extract, Serialize, and Serialized classes to a new extract.py, but it's not behaving for me.  It's entirely possible that I'm not clearing out some old files or something though.

jakirkham:
Yeah was thinking something similar, but actually just lumping them in `scheduler.py` just in case the `ImportError` turns up (though maybe that isn't the concern).

jakirkham:
We could also simplify the code further such that `Serialize` and `Serialized` don't need to move (like checking `typ_v.__name__` instead). Just in case that is causing issues.

jakirkham:
Superseded by PR ( https://github.com/dask/distributed/pull/4531 )

Issue #4443, opened 2021-01-19
SSL Performance


mrocklin:
When running Dask with TLS security turned on, the scheduler can easily become bottlenecked by SSL communication.  See the performance report below, in particular the "Scheduler Profile" tab.

https://gistcdn.githack.com/mrocklin/1d24fbac2c66364d8717952ee8827c64/raw/9c87343cb358b54eb26dbfe8b0714120c0f5ad81/report.html

I ran this on my laptop with the following code

```python
import dask
from dask.distributed import Client, performance_report

client = Client(security=True)

dask.config.set({"optimization.fuse.active": False})

df = dask.datasets.timeseries(start="2020-01-01", end="2020-01-31", partition_freq="1h", freq="60s").persist()

with performance_report("report.html"):
    df2 = df.set_index("x").persist()
```

When this is run on a larger cluster with many workers this problem becomes significantly worse.  What are some ways that we can reduce the cost of secure communication?

mrocklin:
Hrm, here is another performance report for when security is turned off.  Tornado is still a significant cost.

https://gist.githubusercontent.com/mrocklin/5b1e870bc37875f03bf0a6fe0aaec4ba/raw/35cb4a2a803ebe6683f551fed1a17f07fb515c32/insecure.html

mrocklin:
OK, I think that this is just down to  python's `socket.send` costs.

I put a timer around the socket.send calls in `tornado/iostream.py` and came away with 

     3.37 s / 10764 = 313.11 us

We're spending 300-500us per call and making lots of calls.  I can try to batch things a little bit on the worker side but that will only give us a factor increase.  I'm curious how we can take communication overhead off of the main thread.  

For reference, I got these numbers by instrumenting Tornado in the following way

```diff
diff --git a/tornado/iostream.py b/tornado/iostream.py
index 768b404b..f3d000bc 100644
--- a/tornado/iostream.py
+++ b/tornado/iostream.py
@@ -81,6 +81,38 @@ if sys.platform == "darwin":
 _WINDOWS = sys.platform.startswith("win")
 
 
+
+import contextlib
+from collections import defaultdict
+from time import time
+from dask.utils import format_time
+
+total_time_data = defaultdict(float)
+counts_data = defaultdict(int)
+
+
+@contextlib.contextmanager
+def duration(name: str) -> None:
+    start = time()
+
+    yield
+
+    stop = time()
+
+    total_time_data[name] += stop - start
+    counts_data[name] += 1
+
+
[...]

PR #4451, opened 2021-01-22
Optimize transitions

jakirkham:
Closes https://github.com/dask/distributed/issues/4454
~~Requires https://github.com/dask/distributed/pull/4452~~ (merged! 🎉)

This more thoroughly optimizes the higher-level `transition` and `transitions` functions. Does this by going through and annotating the variables used. Also avoids contains checks when it is possible to retrieve with a fallback (like with `dict.get(...)`). Tries to remove any unneeded copies where possible.

Also this collects all messages to send to workers and clients from transitions and waits to send them until the end of the transition where it lumps multiple messages together.

Note: Still need to move communication calls from `transition` to `transitions`.

### DIFF
--- a/distributed/batched.py
+++ b/distributed/batched.py
@@ -127,7 +127,7 @@ class BatchedSend:
         self.stopped.set()
         self.abort()
 
-    def send(self, msg):
+    def send(self, *msgs):
         """Schedule a message for sending to the other side
 
         This completes quickly and synchronously
@@ -135,8 +135,8 @@ class BatchedSend:
         if self.comm is not None and self.comm.closed():
             raise CommClosedError
 
-        self.message_count += 1
-        self.buffer.append(msg)
+        self.message_count += len(msgs)
+        self.buffer.extend(msgs)
         # Avoid spurious wakeups if possible
         if self.next_deadline is None:
             self.waker.set()
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -182,7 +182,10 @@ DEFAULT_EXTENSIONS = [
     EventExtension,
 ]
 
-ALL_TASK_STATES = {"released", "waiting", "no-worker", "processing", "erred", "memory"}
+ALL_TASK_STATES = declare(
+    set, {"released", "waiting", "no-worker", "processing", "erred", "memory"}
+)
+globals()["ALL_TASK_STATES"] = ALL_TASK_STATES
 
 
 @final
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5892,11 +5892,8 @@ class Scheduler(SchedulerState, ServerNode):
             if self.plugins:
                 # Temporarily put back forgotten key for plugin to retrieve it
                 if ts._state == "forgotten":
-                    try:
-                        ts._dependents = dependents
-                        ts._dependencies = dependencies
-                    except KeyError:
-                        pass
+                    ts._dependents = dependents
+                    ts._dependencies = dependencies
                     parent._tasks[ts._key] = ts
                 for plugin in list(self.plugins):
                     try:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5839,19 +5839,19 @@ class Scheduler(SchedulerState, ServerNode):
         dependents: set
         dependencies: set
         try:
+            recommendations = {}
             try:
                 ts = parent._tasks[key]
             except KeyError:
-                return {}
+                return recommendations
             start = ts._state
             if start == finish:
-                return {}
+                return recommendations
 
             if self.plugins:
                 dependents = set(ts._dependents)
                 dependencies = set(ts._dependencies)
 
-            recommendations = {}
             worker_msgs = {}
             client_msgs = {}
             if (start, finish) in self._transitions:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5858,7 +5858,8 @@ class Scheduler(SchedulerState, ServerNode):
             client_msgs = {}
             func = self._transitions.get(start_finish)
             if func is not None:
-                recommendations, worker_msgs, client_msgs = func(key, *args, **kwargs)
+                t: tuple = func(key, *args, **kwargs)
+                recommendations, worker_msgs, client_msgs = t
             elif "released" not in start_finish:
                 func = self._transitions["released", finish]
                 assert not args and not kwargs
@@ -5867,7 +5868,8 @@ class Scheduler(SchedulerState, ServerNode):
                 if v is not None:
                     func = self._transitions["released", v]
                 b: dict
-                b, worker_msgs, client_msgs = func(key)
+                t: tuple = func(key)
+                b, worker_msgs, client_msgs = t
                 recommendations.update(a)
                 recommendations.update(b)
                 start = "released"
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5839,12 +5839,12 @@ class Scheduler(SchedulerState, ServerNode):
     # State Transitions #
     #####################
 
-    def transition(self, key, finish: str, *args, **kwargs):
+    def _transition(self, key, finish: str, *args, **kwargs):
         """Transition a key from its current state to the finish state
 
         Examples
         --------
-        >>> self.transition('x', 'waiting')
+        >>> self._transition('x', 'waiting')
         {'x': 'processing'}
 
         Returns
@@ -5889,7 +5889,7 @@ class Scheduler(SchedulerState, ServerNode):
             elif "released" not in start_finish:
                 func = self._transitions["released", finish]
                 assert not args and not kwargs
-                a: dict = self.transition(key, "released")
+                a: dict = self._transition(key, "released")
                 v = a.get(key)
                 if v is not None:
                     func = self._transitions["released", v]
@@ -5950,6 +5950,24 @@ class Scheduler(SchedulerState, ServerNode):
                 pdb.set_trace()
             raise
 
+    def transition(self, key, finish: str, *args, **kwargs):
+        """Transition a key from its current state to the finish state
+
+        Examples
+        --------
+        >>> self.transition('x', 'waiting')
+        {'x': 'processing'}
+
+        Returns
+        -------
+        Dictionary of recommendations for future transitions
+
+        See Also
+        --------
+        Scheduler.transitions: transitive version of this function
+        """
+        return self._transition(key, finish, *args, **kwargs)
+
     def transitions(self, recommendations: dict):
         """Process transitions until none are left
 
@@ -5963,7 +5981,7 @@ class Scheduler(SchedulerState, ServerNode):
         while recommendations:
             key, finish = recommendations.popitem()
             keys.add(key)
-            new = self.transition(key, finish)
+            new = self._transition(key, finish)
             recommendations.update(new)
 
         if parent._validate:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -5834,6 +5834,7 @@ class Scheduler(SchedulerState, ServerNode):
         ts: TaskState
         start: str
         start_finish: tuple
+        finish2: str
         recommendations: dict
         worker_msgs: dict
         client_msgs: dict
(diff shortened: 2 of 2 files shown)


distributed/scheduler.py
```diff
@@ -5894,7 +5899,7 @@ def transition(self, key, finish, *args, **kwargs):
             if ts._state == "forgotten" and ts._group._name in self.task_groups:
                 # Remove TaskGroup if all tasks are in the forgotten state
                 tg: TaskGroup = ts._group
-                if not any([tg._states.get(s) for s in ALL_TASK_STATES]):
+                if not (tg._states.keys() & ALL_TASK_STATES):
```
jakirkham, line 5902: As key views are `Set`-like objects, things like intersections can be computed with `set`s. More details in [this SO answer]( https://stackoverflow.com/a/18554039/3877089 ) along with [a link to the Python docs]( https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects ) where they note this:

> Keys views are set-like since their entries are unique and hashable. If all values are hashable, so that `(key, value)` pairs are unique and hashable, then the items view is also set-like. (Values views are not treated as set-like since the entries are generally not unique.) For set-like views, all of the operations defined for the abstract base class [`collections.abc.Set`]( https://docs.python.org/3/library/collections.abc.html#collections.abc.Set ) are available (for example, `==`, `<`, or `^`).

jakirkham: In addition to simplifying the code a bit, this ends up being a bit more performant

```python
In [1]: d1 = {i: i for i in range(1_000)}
   ...: s2 = {i for i in range(1_000, 2_000)}

In [2]: %timeit any([d1.get(k) for k in s2])
177 µs ± 456 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [3]: %timeit d1.keys() & s2
15.8 µs ± 157 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Plus Cython can replace `&` with a quick C call to `PyNumber_And`.

distributed/scheduler.py
```diff
@@ -182,7 +182,11 @@ def nogil(func):
     EventExtension,
 ]
 
-ALL_TASK_STATES = {"released", "waiting", "no-worker", "processing", "erred", "memory"}
+ALL_TASK_STATES = declare(
+    set,
+    {"released", "waiting", "no-worker", "processing", "erred", "memory"},
+)
+sys.modules[__name__].ALL_TASK_STATES = ALL_TASK_STATES
```
jakirkham, line 189: Was unable to `import` outside of this module without this last line. Hoping we can get rid of it. Asked here ( https://github.com/cython/cython/issues/3944#issuecomment-767338699 )

distributed/batched.py
```diff
@@ -127,16 +127,16 @@ def _background_send(self):
...
 
         This completes quickly and synchronously
         """
         if self.comm is not None and self.comm.closed():
             raise CommClosedError
 
-        self.message_count += 1
-        self.buffer.append(msg)
+        self.message_count += len(msgs)
+        self.buffer.extend(msgs)
```
mrocklin, line 139: Ah, this is cleaner than I expected :)

mrocklin:
If you have time I encourage you to try out viztracer to see the results.  I would be happy to help walk you through first use of it if you want to get together some time tomorrow.

jakirkham:
Yeah I did run with cProfile last night and looked at the call graph. Though that was before the changes to `BatchedSend`.

Trying to root out a remaining bug. Probably just a small typo I'm overlooking.

mrocklin:
> Yeah I did run with cProfile last night and looked at the call graph

I think that it's worth adding viztracer to your bag of tricks.  I think that you especially might really find value from it.

distributed/scheduler.py
```diff
@@ -4705,6 +4709,29 @@ def client_send(self, client, msg):
...
+        stream_comms: dict = self.stream_comms
+        client_comms: dict = self.client_comms
+        msgs: list
+
+        for worker, msgs in worker_msgs.items():
+            try:
+                w = stream_comms[worker]
+                w.send(*msgs)
+            except (CommClosedError, AttributeError):
+                self.loop.add_callback(self.remove_worker, address=worker)
```
fjetter, line 4723: This should not be necessary. The BatchedSend uses the comm we're using in `handle_stream` which already deals with the case of a broken up connection. If we call the `remove_worker` here unjustified, it closes the worker which would be a very radical way of dealing with an error, in particular an AttributeError (where is this coming from?)

distributed/scheduler.py
```diff
@@ -5882,16 +5960,13 @@ def transition(self, key, finish, *args, **kwargs):
                     start,
                     finish2,
                     ts._state,
-                    dict(recommendations),
+                    str(recommendations),
```
fjetter, line 5963: Any reason for this str conversion?

distributed/scheduler.py
jakirkham: This just came from `worker_send` where this was being done. Just copied that logic here. Happy to drop it if it is not needed.

https://github.com/dask/distributed/blob/2089e79b3afbc3e974981f3e3411bbcbb9c4d0a6/distributed/scheduler.py#L4693-L4694

distributed/scheduler.py
jakirkham: There seems to be a fair bit of preventive copying in this code. I'm assuming this is out of concern that `recommendations` might get unintentionally modified. Thought that as this is building a logging message and `str` is immutable this would maintain the intent of the original code while avoiding copying the `dict`. Am ok just dropping the copy and using `recommendations` if that is preferred

fjetter: I was mostly curious what triggers this change. Which approach is best depends a bit on how these logs are used. For instance, the `{Scheduler,Worker}.story` uses these to filter the messages for keys. I believe the filtering there uses some kind of set logic, so I would assume we'll need to adjust it there as well.
I find working with these things programatically a lot easier in a structured container like a dict. If we want it to be immutable, something like `tuple(recommendations.items())` would preserve the structure, or at least would allow easier programatic access w/out needing to parse `str(dict)`. reverting this is surprisingly difficult :)

I would also expect the str conversion to be more expensive than the dict copy. A shallow copy should only copy the hash table and since the recommendations are mostly small-ish dicts I would hope the shallow copy is ok. otoh, a str conversion would need to copy the entire dict content, wouldn't it?

distributed/scheduler.py
fjetter: You are right of course. I was mostly confused about the `AttributeError`.

The most radical escalation at this point would be to submit a close to the worker but this is not happening if I followed to code path due to an abort in the handle_worker

https://github.com/dask/distributed/blob/b5c36b587f0e3295fe59db330603d5c78b39331f/distributed/scheduler.py#L4662-L4664

I don't think the current solution is clean or nice but since it works I suggest not to change anything. Thanks for checking

jakirkham: Understandable maybe it could be `None` in some cases? Though looking through the Scheduler I don't see it being set that way (unless I'm missing something). So it may not be needed.

Yeah IIUC this is just recognizing the worker doesn't exist and can't be reached so is pruning. Though I agree it wasn't obvious that was what was happening when looking at the code.

distributed/scheduler.py
jakirkham: Yeah I don't have strong feelings here if you would prefer it reverted. Was just thinking copying to a `dict` might not be needed. Am ok dropping the `str(...)` bit or reverting, whichever you prefer.

jakirkham: Went ahead and just reverted this

jakirkham:
Seeing a new failure I haven't seen before about layer annotations. Not seeing how we may have affected that here. Wondering if it is related to some other change like commit ( https://github.com/dask/dask/commit/9bb586a6b8fac1983b7cea3ab399719f93dbbb29 ) possibly?

jakirkham:
Should add I'm not really seeing anything socket related (except maybe `socket.close`) in the first 21 items sorted by self time when running on my Mac.

Screen Shot 2021-01-29 at 8 58 04 AM

mrocklin:
I suspect that you've lost history of the intense part of the shuffle.  viztracer only captures the last N events, so if you don't shut things down relatively quickly you miss the fun part of the computation.  I'm guessing this because of the focus on transitions to the forgotten state.

Also, when selecting the region of interest, I recommend aiming low so that you don't get the events like `Loop.run_forever`.  This should help focus in on only the interesting part.

jakirkham:
Yeah I'll admit I'm not that familiar with this tool. Also was only able to view the results in Chrome (no other browser worked). It was also pretty slow to navigate through. Would encourage others to play with this if they have something in mind that they would like to see.

mrocklin:
If you're interested I'd be very happy to give you a brief tour.  I think that pairing briefly here would be worth the time.

jakirkham:
Ok I gave this one last try. Though honestly I don't plan to spend more time with this tool (am just finding to be too slow; it was causing my computer to freeze)

Am seeing `socket.send` around the 20th item. `send` from `batched` is similarly low on the list. It does still show time spent in things like `decide_worker` and stealing, which seems to suggest it is still doing work. Transitions and serialization are higher in the list.

Screen Shot 2021-01-29 at 10 50 44 AM

jakirkham:
Here's the call graph I get from this change. This can be compared to the recent nightly benchmark ( https://github.com/quasiben/dask-scheduler-performance/issues/98 ). This seems to cut a decent chunk of time `_background_send` relative to what it was before. There is a little bit of time cut from `write`, etc. Though it looks like serialization is the main bottleneck there.

prof_97676 pstat dot

jakirkham:
Ok one last time with viztracer (still struggling with sluggish UI). Just killed the Scheduler roughly 20-30s into working. Guessing that is in the middleish somewhat near the end, which also lines up with the logging messages seen there and some of the transitions popping up here.

Seeing `socket.send` as the 25th item in the list ordered by self time. More time is actually spent reading it seems than writing. This may just be where I happened to terminate the job though. So not sure whether that is indicative of anything.

On the reading point since that came up here, there may be changes that can be made in Tornado that would help. In particular some code changes to rely on asyncio for `sock_recv_into`, which in turn could use uvloop's `sock_recv_into` when enabled in Distributed, may help. See comment ( https://github.com/dask/distributed/issues/4443#issuecomment-769946490 ) and linked issues for more details.

Also time is being spent deserializing messages. Though this goes hand-in-hand with reading and is not something addressed in this PR (though maybe we can look at that after we merge this PR).

Screen Shot 2021-01-29 at 11 57 51 AM

quasiben:
@jakirkham asked me to run this PR and compare with latest in master with Py-Spy with the following code:

```python
ddf_h = timeseries(start='2000-01-01', end='2000-02-01', partition_freq='5min')
    #ddf_h = ddf_h.map_partitions(cudf.from_pandas)
    ddf_h = ddf_h.map_partitions(lambda df: df.head(0))
    ddf_h = ddf_h.persist()
    print(ddf_h)
    _  = wait(ddf_h)
    result = shuffle(ddf_h, "id", shuffle="tasks")
    ddf = client.persist(result)
    _ = wait(ddf)
```

Total tasks: 648270

## Latest (2021-01-29)
- write: 7.88%
- extract serialize: 6.07%
Screen Shot 2021-01-29 at 4 23 24 PM

## This PR
- write: 10.04%
- extract 7.77%
Screen Shot 2021-01-29 at 4 21 23 PM

jakirkham:
To summarize the above results, we have moved a good chunk of time out of `transitions` and into `tcp`'s `write`. Most of the time in `write` is actually just spent serializing data, which we already knew. So that seems like the next thing to work on once this is in.

jakirkham:

gaogaotiantian:
@jakirkham happened to saw this thread about sluggish viztracer :) 

I'm aware that with default size of circular buffer, if you fill all the buffer, you will experience performance issue if your RAM is not large enough. One thing to solve this is to reduce the buffer size with ``--tracer_entries 100000`` (for example). I know it would be good to always have smooth experience with the tool, but it's also important to log as many entries as possible, so I picked a larger number.

However, the latest viztracer now supports perfetto, which has a much better performance than Chrome Trace Viewer, which was developed years ago and was already deprecated.

You can use viztracer with perfetto by
```
# You need to output to json in order to use perfetto
viztracer -o result.json --open your_script.py
```
This will automatically open your browser for the report.

Or you can log to a json file and open it with ``vizviewer``
```
viztracer -o result.json your_script.py
vizviewer result.json
```

I'm sorry that viztracer did not perform as you expected, but if you wanted to give it another try, let me know how the latest UI works :)

PR #4453, opened 2021-01-23
Use a callback to close TCP Comms, rather than check every time

mrocklin:
In a recent trace this relieves about 30ms of a 3s shuffle computation
resulting in around a 1% overall speedup

cc @jakirkham this might help a bit with the worker_send/client_send code in transition.  I think that the real solution is still to try to batch things there though.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -1,4 +1,5 @@
 import errno
+import functools
 import logging
 import socket
 from ssl import SSLError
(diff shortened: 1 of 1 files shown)


jakirkham (approved):

mrocklin:
I recommend that we merge this after the upcoming release.  cc @jrbourbeau

jakirkham:
Sorry Matt maybe I should have merged this earlier. Am ok with waiting until after the release as well

PR #4456, opened 2021-01-25
Added worker resources from config

TomAugspurger:
This allows specifying worker resources through Dask's configuration system.

```python
   with dask.config.set({"distributed.worker.resources.GPU": 2}):
       cluster = LocalCluster()
```

My primary motivation was using worker resources with dask-gateway, which spawns workers with `$ dask-worker ...`. There isn't an especially clean way to pass additional arguments to that command, so I'm making the fix here instead, which may be helpful for other cluster backends that are starting processes for workers. With this change I can set the environment variable `DASK_DISTRIBUTED__WORKER__RESOURCES__GPU=2` and have it be picked up by the workers.

### DIFF
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -305,6 +305,14 @@ properties:
               Whether or not to run consistency checks during execution.
               This is typically only used for debugging.
 
+          resources:
+            type: object
+            description: |
+              A dictionary specifying resources for workers.
+
+              See https://distributed.dask.org/en/latest/resources.html for more information.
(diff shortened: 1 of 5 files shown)


TomAugspurger:
At least one of the failures is related. Will fix that up in a bit.

jakirkham:
cc @jacobtomlinson (in case this is of interest 🙂)

mrocklin:
This looks good to me.  Thanks @TomAugspurger

TomAugspurger:
Thanks!

PR #4457, opened 2021-01-25
Fix var name in worker validation func

gforsyth:
Small fix for an undefined attribute in one of the validation checks


jrbourbeau (approved):
Good catch @gforsyth! Will merge after CI finishes

PR #4459, opened 2021-01-26
Update task duration estimates for unknown tasks while executing

### LINKED ISSUE #3516: Dynamic adaption of average task duration

fjetter:
Since the workers submit their currently running tasks duration, we can update the estimated average for unknown tasks, allowing for fast upscaling and work stealing in case of long running unknown tasks. I opted for not including any measurements anymore if the task is known since any intermediate measurement would skew the average.

I'm a bit on the fence regarding the exponential average for this application but I kept it since this is the way we update this average in other places. I don't think we need something formally correct for this, rather something which updates the value and converges to the true value reasonably fast.

Closes #3516

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3422,9 +3422,20 @@ class Scheduler(SchedulerState, ServerNode):
         ws._last_seen = time()
 
         if executing is not None:
-            ws._executing = {
-                parent._tasks[key]: duration for key, duration in executing.items()
-            }
+            ws._executing = {}
+            for key, duration in executing.items():
+                ts = parent.tasks[key]
+                ws._executing[ts] = duration
+                if ts._prefix._name in self.unknown_durations:
+                    old_duration = ts._prefix._duration_average
+                    new_duration = duration
+                    if old_duration < 0:
+                        avg_duration = new_duration
+                    else:
+                        avg_duration = 0.5 * old_duration + 0.5 * new_duration
+
+                    ts._prefix._duration_average = avg_duration
+                    ts._group._duration = new_duration
 
         if metrics:
             ws._metrics = metrics
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -17,7 +17,7 @@ import pytest
 from distributed import Nanny, Worker, Client, wait, fire_and_forget
 from distributed.comm import Comm
 from distributed.core import connect, rpc, ConnectionPool, Status
-from distributed.scheduler import Scheduler
+from distributed.scheduler import Scheduler, UNKNOWN_TASK_DURATION
 from distributed.client import wait
 from distributed.metrics import time
 from distributed.protocol.pickle import dumps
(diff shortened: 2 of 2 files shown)


distributed/scheduler.py
```diff
@@ -3422,9 +3422,20 @@ def heartbeat_worker(
         ws._last_seen = time()
 
         if executing is not None:
-            ws._executing = {
-                parent._tasks[key]: duration for key, duration in executing.items()
-            }
+            ws._executing = {}
+            for key, duration in executing.items():
+                ts = parent.tasks[key]
```
jakirkham, line 3427: The `_` is for typed attributes available from the `parent` class (`SchedulerState`). If we annotate `ts` as `TaskState`, Cython knows the type and will optimize attribute access, etc.

```suggestion
                ts: TaskState = parent._tasks[key]
```

distributed/scheduler.py
```diff
@@ -3422,9 +3422,20 @@ def heartbeat_worker(
         ws._last_seen = time()
 
         if executing is not None:
-            ws._executing = {
-                parent._tasks[key]: duration for key, duration in executing.items()
-            }
+            ws._executing = {}
+            for key, duration in executing.items():
+                ts: TaskState = parent._tasks[key]
```
fjetter, line 3427: @jakirkham Thanks for your pointer to the underscored attributes. I haven't been working with cython a lot so far and am wondering when I do need to declare types of my variables (probably always :P ) and when not. In particular I'm defining variables like old/new/avg_duration below. For full cythonization I would assume these should be declared as double or does cython pick this up automagically for primitive types?

jakirkham: Ofc. Generally I try to type them. That said, Cython can pick up on things if one is assigning one typed variable to a new variable

distributed/scheduler.py
```diff
@@ -3422,9 +3422,20 @@ def heartbeat_worker(
         ws._last_seen = time()
 
         if executing is not None:
-            ws._executing = {
-                parent._tasks[key]: duration for key, duration in executing.items()
-            }
+            ws._executing = {}
+            for key, duration in executing.items():
```
jakirkham, line 3426: IIUC this is a `double`. I think Cython may be able to figure out the rest once this is typed.

```suggestion
            duration: double
            for key, duration in executing.items():
```

mrocklin:
I'm curious if this will have a negative impact on performance.  This introduces a check on every running task on every heartbeat.  That seems like the kind of thing that might be costly at scale.

One way to partially reduce this cost is by first seeing if there are any unknown duration tasks?  I suspect that in the common case there are not.  I wonder if there are other ways to reduce the frequency of this check.  For example maybe we could have a periodic callback that occasionally checks to see if there are any unknown duration tasks, and if so it checks the prefix for those tasks and some state stored on that prefix.  Just a thought.

fjetter:
> his introduces a check on every running task on every heartbeat.

True. I figured this is not a big deal but here are a few things happening here we could optimize
1. `if executing is not None` We can probably avoid this by ensuring the other end guarantees to submit a dict, even if it is empty. That was already here, though :)
2. setting the `WorkerState._executing` which was previously a list comprehension, now its a proper for loop. Without looking into it, my assumption would be that after cythonization this is effectively the same (assuming everything is typed; I could be wrong here ofc)
3. What's actually new is `ts._prefix._name in self.unknown_durations`, i.e. three attribute accesses and a hash table lookup. All decently fast I assume. we could change this to `if self.unknown_durations and ts._prefix._name in self.unknown_durations` since the second clause should only evaluated if the first is true, I guess this is what you were suggesting?

The other suggestions of introducing a PC would also work. We'd need to loop over all workers but we could do it only very rarely (I think something like 1000ms or even more rarely)

Bottom line. I'm in either way. I have only followed the recent performance optimizations on a high level and am not sure if these kind of micro optimizations are already relevant but I'll keep this in mind for the future (doesn't hurt after all)

I'll think I'll check out the PC approach since it is rather easy to implement and should eliminate all of the above mentioned concerns

mrocklin:
Yes, I was thinking of option (3).  I think that that would significantly limit the number of hash lookups (which I expect to be the expensive thing here).  Also, with this functionality `unknown_duration` will now be empty most of the time :)

fjetter:
> Also, with this functionality unknown_duration will now be empty most of the time :)

Actually, it would only be cleared up once a task finishes. I didn't want to update the moving average with a lot of partial runtime information since this would always drag the average down.

Therefore, the keys are kept in unknown duration until they are finished once. As long as they are "still unknown" we're taking the hearbeat update, otherwise only after they are finished. I think in an ideal world we'd want to have continuous updates to know if there are outliers but I have no nice idea about how to do this without completely skewing the average (every new task would drag the average towards zero)

fjetter:
For the curious, I had cython annotate the code I wrote here and there are a few (to me) surprising python object interactions, see screen below for some examples

image

I'll try a few things out to see how this can be optimized. @jakirkham this might interest you or you might have some pointers or maybe you are not even surprised at all :)

jakirkham:
Thanks for the info Florian! 😄

Yeah the number Python C API calls should be fixed with this suggested change ( https://github.com/dask/distributed/pull/4459#discussion_r564352486 ). I don't think we need more annotations than that, but may if we still see those Python C API calls.

The `None` check is expected. We could disable it with a Cython directive, but it can be handy to catch those nicely if they show up. Plus a `None` check is a just pointer comparison in C (analogous to checking for a `NULL` pointer though `None` is non-`NULL`), which is really fast.

The calls to the Python `dict` C API is expected. The fact that it is using the Python `dict` C API is a good sign that it understands the type of that object and is not using generic Python C APIs. The only way we get rid of the Python `dict` C API is to starting using C or C++ objects in place of Python ones. Not sure that we need to worry about doing that yet.

distributed/scheduler.py
```diff
@@ -3422,9 +3422,20 @@ def heartbeat_worker(
...
+                ts: TaskState = parent._tasks[key]
+                ws._executing[ts] = duration
+                if ts._prefix._name in self.unknown_durations:
+                    old_duration = ts._prefix._duration_average
+                    new_duration = duration
```
jakirkham, line 3431:
```suggestion
                    avg_duration: double
```

jakirkham, line 3431:
```suggestion
                    new_duration = duration
                    avg_duration: double
```

PR #4448, opened 2021-01-21
Add uvloop config value

mrocklin:
Fixes https://github.com/dask/distributed/issues/4369

PR #4461, opened 2021-01-26
MPI4Dask - MVAPICH2-GDR based Communication Backend for the Dask Distributed Library

aamirshafi:
This document outlines installation and usage of MPI4Dask, which provides an MVAPICH2-GDR based MPI communication backend for the Dask Distributed library. MPI4Dask embodies enhancements to the Dask Distributed package. 

A paper titled ["Efficient MPI-based Communication for GPU-Accelerated Dask Applications"](https://arxiv.org/abs/2101.08878) describes MPI4Dask and its performance against existing communication devices in detail.

## Install MPI4Dask

We start off with the installation of MPI4Dask.

### Installation Pre-Requisites

Install [Miniconda](https://docs.conda.io/en/latest/miniconda.html), which is a free minimal installer for the conda package manager. The following commands download and install latest version of Miniconda3 for Python 3.x.

```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
chmod a+x Miniconda3-latest-Linux-x86_64.sh
./Miniconda3-latest-Linux-x86_64.sh
```

After installing the conda package manager, create a new conda environment named mpi4dask (suggested name). 
```
conda create --name mpi4dask
conda activate mpi4dask
```

Make sure that relevant/required modules are loaded in the environment. One of the required module is the CUDA toolkit (10.2 in this case). A list of all installed modules on the system can be seen as follows. Obviously the version of CUDA toolkit might be different on the user system.

```
module list
Currently Loaded Modulefiles:
  1) cuda/10.2
```

Install pre-requisite libraries including build tools, Python (version 3.8 here), Cython, CUDA tookit (version 10.2 here), Dask packages, RAPIDS libraries, and array packages. Users may install specific versions of these packages. Also, these packages are installed from three conda channels, including conda-forge, rapidsai, and nvidia, as follows:

```
conda install -c conda-forge -c rapidsai -c nvidia automake make libtool pkg-config libhwloc psutil python=3.8 setuptools cython cudatoolkit=10.2 cupy dask-cudf dask distributed numpy rmm
```

### Install MVAPICH2-GDR 

Download [the MVAPICH2-GDR library](http://mvapich.cse.ohio-state.edu). Detailed installation instructions are 
available from [here](http://mvapich.cse.ohio-state.edu/userguide/gdr/). Please become familiar with advanced configurations of the MVAPICH2-GDR software --- this is important to get maximum performance from the hardware system. This guide assumes that an environment variable `$MPILIB` points to the installation directory of the MVAPICH2-GDR software.

```
export MPILIB=/path/to/MVAPICH2-GDR/install/directory
```

It is important to make sure that the MVAPICH2-GDR is the MPI library used with MPI4Dask. Multiple installations of MPI libraries can result in errors. In order to avoid such situations, it is important to make sure that the correct version of the MVAPICH2-GDR library is available on the `$PATH` variable. This can be checked as follows.

```
$MPILIB/bin/mpiname -a
```

If the `$PATH` is set correctly, the output of the following command should match the previous command output.

```
mpiname -a
```

### Install mpi4py

Download and install the mpi4py library. Details about mpi4py can be seen [here](https://mpi4py.readthedocs.io). Download the software as follows.

```
git clone https://github.com/mpi4py/mpi4py.git
cd mpi4py
```

As part of the installation, the configuration file (mpi.cfg) in the root directory of mpi4py software needs to be updated to point to the MVAPICH2-GDR library. It is recommended to make a new section---in the mpi.cfg file---for MVAPICH2-GDR as follows:

```
# ------------
[MVAPICH2-GDR]
mpi_dir              = /path/to/MVAPICH2-GDR/install/directory
mpicc                = %(mpi_dir)s/bin/mpicc
mpicxx               = %(mpi_dir)s/bin/mpicxx
include_dirs         = %(mpi_dir)s/include
library_dirs         = %(mpi_dir)s/lib
runtime_library_dirs = %(library_dirs)s
```

After updating the mpi.cfg configuration file, install mpi4py with following commands:

```
python setup.py build --mpi=MVAPICH2-GDR
pip install .
```
### Install Dask-MPI

Download and Install Dask-MPI package from the github repository. MPI4Dask relies on Dask-MPI to start execution of Dask scheduler, client, and workers. Details about Dask-MPI package can be seen [here](http://mpi.dask.org). It is recommended not to install Dask-MPI through the conda package manager as it installs some conflicting packages as dependencies.

```
git clone https://github.com/dask/dask-mpi.git
cd dask-mpi
python setup.py build
pip install .
```
## Download MPI4Dask

MPI4Dask has been shared as a pull request. Enhancements to the Dask Distributed library can be summarized as follows.

1. The main enhancement to the the distributed package is the `distributed/comm/mpi.py` file
2. The dask-apps folder that contains two representative Dask test/benchmarking applications to show the MPI backend

## Install MPI4Dask

MPI4Dask can be installed using the following commands:

```
python setup.py build
pip install .
```

At this time, MPI4Dask with all its dependencies should be installed. Verify this by running the following commands:
```
conda list
conda list | grep distributed
conda list | grep dask
```

## Basic Usage Instructions

The MPI4Dask package contains two sample applications that can be used to test and benchmark performance of the MPI backend. These applications include: 1) Sum of cuPy Array and its Transpose, and 2) cuDF Merge.

### Sum of cuPy Array and its Transpose

This benchmark creates a cuPy array and distributes its chunks across Dask workers. The benchmark adds these
distributed chunks to their transpose. This benchmark is originally taken from the [dask-cuda package repository](https://github.com/rapidsai/dask-cuda/tree/branch-0.18/dask_cuda/benchmarks).

The next step is to start execution of the Dask program using Dask-MPI.

Write the hosts files. This file contains names of the compute nodes, with GPUs, that will execute the parallel Dask program.

```
cd dask-apps
vim hosts
.. write name of the compute nodes ..
cat hosts
machine1
machine2
machine3
machine4
```

Note that users need to customize the hosts file as per their environment. This benchmark can be executed as follows:

```
LD_PRELOAD=$MPILIB/lib/libmpi.so $MPILIB/bin/mpirun_rsh -export-all -np 4 -hostfile hosts MV2_USE_CUDA=1 MV2_USE_GDRCOPY=1 MV2_GPUDIRECT_GDRCOPY_LIB=/opt/gdrcopy2.0/lib64/libgdrapi.so MV2_CPU_BINDING_LEVEL=SOCKET MV2_CPU_BINDING_POLICY=SCATTER python cupy_sum_mpi.py
```

This command is starting the execution of the Dask application using MPI --- this is enabled by the Dask-MPI package. A total of 4 MPI processes are started with process 0 and 1 assuming the role of Dask scheduler and client respectively. All other processes with ranks greater than 1 becoming Dask workers. Here processes 2 and 3 are Dask workers. The `LD_PRELOAD` environment variable is required for Python applications to run correctly with the MVAPICH2-GDR library.

Details on various options, passed to the MVAPICH2-GDR library, are as follows:

- `-np 4` --- specifies the total number of parallel processes started by the MPI library
- `-hostfile hosts` --- specifies the names of machines where to start parallel processes in the ``hosts'' file
- `MV2_USE_CUDA=1` --- ensures that MVAPICH2-GDR library supports GPU-to-GPU communication
- `MV2_USE_GDRCOPY=1` --- turns on the GDRCopy protocol
- `MV2_GPUDIRECT_GDRCOPY_LIB` --- specifies path to the GDRCopy dynamic library
- `MV2_CPU_BINDING_LEVEL=SOCKET MV2_CPU_BINDING_POLICY=SCATTER` --- selects a socket-level scatter process binding policy

This is not an exhaustive list of options that can be passed to the MVAPICH2-GDR library. Details on these --- and other flags for the MVAPICH2-GDR library --- can be seen [here](http://mvapich.cse.ohio-state.edu/userguide/gdr/). As mentioned earlier, it is important to configure the MVAPICH2-GDR installation correctly to ensure optimal performance.

The `cupy_sum_mpi.py` script has some configuration parameters that can be modified, in the source-code, by the users. These include:

- `DASK_PROTOCOL = 'mpi'` --- specifies the protocol used by the Dask Distributed library. Options include `mpi', `tcp', `ucx'
- `DASK_INTERFACE = 'ib0'` --- specifies the interface used by the Dask Distributed library
- `GPUS_PER_NODE = 1` --- specifies the number of GPUs in the system
- `THREADS_PER_NODE = 28` --- specifies the number of threads (cores) in the system

### cuDF Merge

cuDF DataFrames are table-like data-structure that are stored in the GPU memory. As part of this application, the merge operation is carried out for multiple cuDF data frames. This benchmark is originally taken from the
[dask-cuda package repository](https://github.com/rapidsai/dask-cuda/tree/branch-0.18/dask\_cuda/benchmarks).

This application can be executed as follows.

```
cd dask-apps
LD_PRELOAD=$MPILIB/lib/libmpi.so $MPILIB/bin/mpirun_rsh -export-all -np 4 -hostfile hosts MV2_USE_CUDA=1 MV2_USE_GDRCOPY=1 MV2_GPUDIRECT_GDRCOPY_LIB=/opt/gdrcopy2.0/lib64/libgdrapi.so MV2_CPU_BINDING_LEVEL=SOCKET MV2_CPU_BINDING_POLICY=SCATTER python cudf_merge_mpi.py --type gpu --protocol mpi --runs 20 --chunk-size 1_000_000_00
```

The `cudf_merge_mpi.py` script accepts some runtime arguments. Some of the important ones include:

- `--type gpu` --- specifies that the data frame is GPU-based. Currently only GPU has been tested
- `--protocol mpi` --- specifies the communication backend used by Dask Distributed. Options include `mpi', `tcp', `ucx'
- `--runs 20` --- specifies the number of repetitions for the experiment
- `--chunk-size 1_000_000_00` --- specifies the chunk size for the data frame

In addition, the `cudf_merge_mpi.py` script has some configuration parameters that can be modified by the users. These include:

- `DASK_INTERFACE = 'ib0'` --- specifies the interface used by the Dask Distributed library
- `GPUS_PER_NODE = 1` --- specifies the number of GPUs in the system
- `THREADS_PER_NODE = 28` --- specifies the number of threads (cores) in the system

### DIFF
--- a/dask-apps/cudf_merge_mpi.py
+++ b/dask-apps/cudf_merge_mpi.py
@@ -1,0 +1,594 @@
+"""
+
+Taken from: https://github.com/rapidsai/dask-cuda/blob/branch-0.18/dask_cuda/benchmarks/local_cudf_merge.py
+
+This is the cuDF merge benchmarking application. It has been 
+modified to work for the MVAPICH2-based (http://mvapich.cse.ohio-state.edu) 
+communication backend for the Dask Distributed library using the 
+dask-mpi package.
+
+"""
+
+import math
+from collections import defaultdict
+from time import perf_counter as clock
+
+import numpy
+
+from dask.base import tokenize
+from dask.dataframe.core import new_dd_object
+from dask.distributed import Client, performance_report, wait
+from dask.utils import format_bytes, format_time, parse_bytes
+from dask_mpi import initialize
+
+import argparse
+import os
+
+from dask.distributed import SSHCluster
+
+#adjust as per compute system
+GPUS_PER_NODE       =   1    # number of GPUs in the system  
+DASK_INTERFACE      = 'ib0'  # interface to use for communication
+THREADS_PER_NODE    =  28    # number of threads per node.
+
+rank = os.environ['MV2_COMM_WORLD_LOCAL_RANK']
+device_id = int(rank) % GPUS_PER_NODE
+os.environ["CUDA_VISIBLE_DEVICES"]=str(device_id)
+
+#from dask_cuda.local_cuda_cluster import LocalCUDACluster
+
+#from dask_cuda import explicit_comms
+#from utils import (
+#    get_cluster_options,
+#    get_scheduler_workers,
+#    parse_benchmark_args,
+#    setup_memory_pool,
+#)
+
+# Benchmarking cuDF merge operation based on
+# <https://gist.github.com/rjzamora/0ffc35c19b5180ab04bbf7c793c45955>
+
+def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match, gpu):
+    #print("generate_chunk")
+    # Setting a seed that triggers max amount of comm in the two-GPU case.
+    if gpu:
+        import cupy as xp
+
+        import cudf as xdf
+    else:
+        import numpy as xp
+        import pandas as xdf
+
+    xp.random.seed(2 ** 32 - 1)
+
+    chunk_type = chunk_type or "build"
+    frac_match = frac_match or 1.0
+    if chunk_type == "build":
+        # Build dataframe
+        #
+        # "key" column is a unique sample within [0, local_size * num_chunks)
+        #
+        # "shuffle" column is a random selection of partitions (used for shuffle)
+        #
+        # "payload" column is a random permutation of the chunk_size
+
+        start = local_size * i_chunk
+        stop = start + local_size
+
+        parts_array = xp.arange(num_chunks, dtype="int64")
+        suffle_array = xp.repeat(parts_array, math.ceil(local_size / num_chunks))
+        
+        df = xdf.DataFrame(
+            {
+                "key": xp.arange(start, stop=stop, dtype="int64"),
+                "shuffle": xp.random.permutation(suffle_array)[:local_size],
+                "payload": xp.random.permutation(xp.arange(local_size, dtype="int64")),
+            }
+        )
+    else:
+        #print("chunk type is other")
+        # Other dataframe
+        #
+        # "key" column matches values from the build dataframe
+        # for a fraction (`frac_match`) of the entries. The matching
+        # entries are perfectly balanced across each partition of the
+        # "base" dataframe.
+        #
+        # "payload" column is a random permutation of the chunk_size
+
+        # Step 1. Choose values that DO match
+        sub_local_size = local_size // num_chunks
+        sub_local_size_use = max(int(sub_local_size * frac_match), 1)
+        arrays = []
+        for i in range(num_chunks):
+            bgn = (local_size * i) + (sub_local_size * i_chunk)
+            end = bgn + sub_local_size
+            ar = xp.arange(bgn, stop=end, dtype="int64")
+            arrays.append(xp.random.permutation(ar)[:sub_local_size_use])
+        key_array_match = xp.concatenate(tuple(arrays), axis=0)
+
+        # Step 2. Add values that DON'T match
+        missing_size = local_size - key_array_match.shape[0]
+        start = local_size * num_chunks + local_size * i_chunk
+        stop = start + missing_size
+        key_array_no_match = xp.arange(start, stop=stop, dtype="int64")
+
+        # Step 3. Combine and create the final dataframe chunk (dask_cudf partition)
+        key_array_combine = xp.concatenate(
+            (key_array_match, key_array_no_match), axis=0
+        )
+
+        df = xdf.DataFrame(
+            {
+                "key": xp.random.permutation(key_array_combine),
+                "payload": xp.random.permutation(xp.arange(local_size, dtype="int64")),
+            }
+        )
+    return df
+
+
+def get_random_ddf(chunk_size, num_chunks, frac_match, chunk_type, args):
+
+    parts = [chunk_size for i in range(num_chunks)]
+    device_type = True if args.type == "gpu" else False
+    meta = generate_chunk(0, 4, 1, chunk_type, None, device_type)
+    divisions = [None] * (len(parts) + 1)
+
+    name = "generate-data-" + tokenize(chunk_size, num_chunks, frac_match, chunk_type)
+
+    graph = {
+        (name, i): (
+            generate_chunk,
+            i,
+            part,
+            len(parts),
+            chunk_type,
+            frac_match,
+            device_type,
+        )
+        for i, part in enumerate(parts)
+    }
+
+    ddf = new_dd_object(graph, name, meta, divisions)
+
+    if chunk_type == "build":
+        if not args.no_shuffle:
+            divisions = [i for i in range(num_chunks)] + [num_chunks]
+            return ddf.set_index("shuffle", divisions=tuple(divisions))
+        else:
+            del ddf["shuffle"]
+
+    return ddf
+
+
+def merge(args, ddf1, ddf2, write_profile):
+    #print("merge called")
+    # Lazy merge/join operation
+    ddf_join = ddf1.merge(ddf2, on=["key"], how="inner")
+    if args.set_index:
+        ddf_join = ddf_join.set_index("key")
+
+    # Execute the operations to benchmark
+    if write_profile is not None:
+        with performance_report(filename=args.profile):
+            t1 = clock()
+            wait(ddf_join.persist())
+            took = clock() - t1
+    else:
+        t1 = clock()
+        wait(ddf_join.persist())
+        took = clock() - t1
+    return took
+
+
+#def merge_explicit_comms(args, ddf1, ddf2):
+#    t1 = clock()
+#    wait(explicit_comms.dataframe_merge(ddf1, ddf2, on="key").persist())
+#    took = clock() - t1
+#    return took
+
+
+def run(client, args, n_workers, write_profile=None):
+    # Generate random Dask dataframes
+    ddf_base = get_random_ddf(
+        args.chunk_size, n_workers, args.frac_match, "build", args
+    ).persist()
+
+    ddf_other = get_random_ddf(
+        args.chunk_size, n_workers, args.frac_match, "other", args
+    ).persist()
+
+    wait(ddf_base)
+    wait(ddf_other)
+    client.wait_for_workers(n_workers)
+
+    assert len(ddf_base.dtypes) == 2
+    assert len(ddf_other.dtypes) == 2
+    data_processed = len(ddf_base) * sum([t.itemsize for t in ddf_base.dtypes])
+    data_processed += len(ddf_other) * sum([t.itemsize for t in ddf_other.dtypes])
+
+    if args.backend == "dask":
+        took = merge(args, ddf_base, ddf_other, write_profile)
+    else:
+        took = None#merge_explicit_comms(args, ddf_base, ddf_other)
+
+    return (data_processed, took)
+
+
+def main(args):
+    cluster_options = get_cluster_options(args)
+    Cluster = cluster_options["class"]
+    cluster_args = cluster_options["args"]
+    cluster_kwargs = cluster_options["kwargs"]
+    scheduler_addr = cluster_options["scheduler_addr"]
+
+    if args.sched_addr:
+        initialize(
+            interface=DASK_INTERFACE,
+            protocol=args.protocol,
+            nanny=False,
+            nthreads=THREADS_PER_NODE
+        )
+        import time
+        time.sleep(5)
+
+        client = Client()
+        print(client)
+    else:
+        cluster = Cluster(*cluster_args, **cluster_kwargs)
+        if args.multi_node:
+            import time
+
+            # Allow some time for workers to start and connect to scheduler
+            # TODO: make this a command-line argument?
+            time.sleep(15)
+
+        client = Client(scheduler_addr if args.multi_node else cluster)
+
+    if args.type == "gpu":
+        client.run(setup_memory_pool, disable_pool=args.no_rmm_pool)
+        # Create an RMM pool on the scheduler due to occasional deserialization
+        # of CUDA objects. May cause issues with InfiniBand otherwise.
+        client.run_on_scheduler(setup_memory_pool, 1e9, disable_pool=args.no_rmm_pool)
+
+    scheduler_workers = client.run_on_scheduler(get_scheduler_workers)
+    n_workers = len(scheduler_workers)
+
+    took_list = []
+    for _ in range(args.runs - 1):
+        took_list.append(run(client, args, n_workers, write_profile=None))
+    took_list.append(
+        run(client, args, n_workers, write_profile=args.profile)
+    )  # Only profiling the last run
+
+    # Collect, aggregate, and print peer-to-peer bandwidths
+    incoming_logs = client.run(lambda dask_worker: dask_worker.incoming_transfer_log)
+    bandwidths = defaultdict(list)
+    total_nbytes = defaultdict(list)
+    for k, L in incoming_logs.items():
+        for d in L:
+            if d["total"] >= args.ignore_size:
+                bandwidths[k, d["who"]].append(d["bandwidth"])
+                total_nbytes[k, d["who"]].append(d["total"])
+    bandwidths = {
+        (scheduler_workers[w1].name, scheduler_workers[w2].name): [
+            "%s/s" % format_bytes(x) for x in numpy.quantile(v, [0.25, 0.50, 0.75])
+        ]
+        for (w1, w2), v in bandwidths.items()
+    }
+    total_nbytes = {
+        (scheduler_workers[w1].name, scheduler_workers[w2].name,): format_bytes(sum(nb))
+        for (w1, w2), nb in total_nbytes.items()
+    }
+
+    t_runs = numpy.empty(len(took_list))
+    if args.markdown:
+        print("```")
+    print("Merge benchmark")
+    print("-------------------------------")
+    print(f"backend        | {args.backend}")
+    print(f"merge type     | {args.type}")
+    print(f"rows-per-chunk | {args.chunk_size}")
+    print(f"protocol       | {args.protocol}")
+    print(f"device(s)      | {args.devs}")
+    print(f"rmm-pool       | {(not args.no_rmm_pool)}")
+    print(f"frac-match     | {args.frac_match}")
+    if args.protocol == "ucx":
+        print(f"tcp            | {args.enable_tcp_over_ucx}")
+        print(f"ib             | {args.enable_infiniband}")
+        print(f"nvlink         | {args.enable_nvlink}")
+    print(f"data-processed | {format_bytes(took_list[0][0])}")
+    print("===============================")
+    print("Wall-clock     | Throughput")
+    print("-------------------------------")
+    for idx, (data_processed, took) in enumerate(took_list):
+        throughput = int(data_processed / took)
+        m = format_time(took)
+        m += " " * (15 - len(m))
+        print(f"{m}| {format_bytes(throughput)}/s")
+        t_runs[idx] = float(format_bytes(throughput).split(" ")[0])
+    print("===============================")
+    if args.markdown:
+        print("\n```")
+
+    #if args.plot is not None:
+    #    plot_benchmark(t_runs, args.plot, historical=True)
+
+    if args.backend == "dask":
+        if args.markdown:
+            print("<details>\n<summary>Worker-Worker Transfer Rates</summary>\n\n```")
+        print("(w1,w2)     | 25% 50% 75% (total nbytes)")
+        print("-------------------------------")
+        for (d1, d2), bw in sorted(bandwidths.items()):
+            fmt = (
+                "(%s,%s)     | %s %s %s (%s)"
+                if args.multi_node or args.sched_addr
+                else "(%02d,%02d)     | %s %s %s (%s)"
+            )
+            print(fmt % (d1, d2, bw[0], bw[1], bw[2], total_nbytes[(d1, d2)]))
+        if args.markdown:
+            print("```\n</details>\n")
+
+    if args.multi_node:
+        client.shutdown()
+        client.close()
+
+
+def parse_args():
+    special_args = [
+        {
+            "name": ["-b", "--backend",],
+            "choices": ["dask", "explicit-comms"],
+            "default": "dask",
+            "type": str,
+            "help": "The backend to use.",
+        },
+        {
+            "name": ["-t", "--type",],
+            "choices": ["cpu", "gpu"],
+            "default": "gpu",
+            "type": str,
+            "help": "Do merge with GPU or CPU dataframes",
+        },
+        {
+            "name": ["-c", "--chunk-size",],
+            "default": 1_000_000,
+            "metavar": "n",
+            "type": int,
+            "help": "Chunk size (default 1_000_000)",
+        },
+        {
+            "name": "--ignore-size",
+            "default": "1 MiB",
+            "metavar": "nbytes",
+            "type": parse_bytes,
+            "help": "Ignore messages smaller than this (default '1 MB')",
+        },
+        {
+            "name": "--frac-match",
+            "default": 0.3,
+            "type": float,
+            "help": "Fraction of rows that matches (default 0.3)",
+        },
+        {
+            "name": "--no-shuffle",
+            "action": "store_true",
+            "help": "Don't shuffle the keys of the left (base) dataframe.",
+        },
+        {
+            "name": "--markdown",
+            "action": "store_true",
+            "help": "Write output as markdown",
+        },
+        {"name": "--runs", "default": 3, "type": int, "help": "Number of runs",},
+        {
+            "name": ["-s", "--set-index",],
+            "action": "store_true",
+            "help": "Call set_index on the key column to sort the joined dataframe.",
+        },
+    ]
+
+    return parse_benchmark_args(
+        description="Distributed merge (dask/cudf) benchmark", args_list=special_args
+    )
+
+def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]):
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument(
+        "-d", "--devs", default="0", type=str, help='GPU devices to use (default "0").'
+    )
+    parser.add_argument(
+        "-p",
+        "--protocol",
+        choices=["tcp", "ucx", "mpi"],
+        default="tcp",
+        type=str,
+        help="The communication protocol to use.",
+    )
+    parser.add_argument(
+        "--profile",
+        metavar="PATH",
+        default=None,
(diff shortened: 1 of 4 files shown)


quasiben:
I'm sorry, I don't see a PR/Link/Repo for MPI4Dask.  Can you please add this ?

aamirshafi:
@quasiben https://github.com/aamirshafi/distributed/tree/mpi-comm-device. Does that answer your question?

aamirshafi:
@hashmij FYI.

quasiben:
@aamirshafi , sorry I guess I missed the codes changes when the PR was first submitted

jakirkham:
It might be worth comparing notes with the implementation in PR ( https://github.com/dask/distributed/pull/4142 )

aamirshafi:
Sure @jakirkham. I can have a go at comparing notes with PR (https://github.com/dask/distributed/pull/4142). 

@ianthomas23: you are welcome to chime in please.

It will be good to have this PR, called MPI4Dask, or Ian's MPI device merged in distributed so we can plan future implementations based on these. 

MPI4Dask has been implemented and testing using the [MVAPICH2](http://mvapich.cse.ohio-state.edu/) family of MPI libraries.

MPI4Dask is based on the UCX communication device and hopefully that makes it easier to review.

### Dependency on dask-mpi and Initializing MPI Library ###

Both PRs depend on `dask-mpi` for starting the scheduler, client, and workers. The native MPI library has already been initialized by the `dask-mpi` package. 

### Connection Establishment ###

One major difference between the two implementations is managing connection establishment in the MPI communication device. 

The API dictated by the Dask Distributed communication device is based on the client-server model. MPI, at the application-level, does not follow this model where all processes are started together. Both implementations have not used Dynamic Process Management (DPM) feature of MPI. For this reason, both devices will support static number of workers and dynamic addition and removal of workers is not supported. This feature can be added in the future.

In this PR (MPI4Dask), we have used out-of-band connection establishment using `asyncio.open_connection()` and `asyncio.start_server()`. Every time a worker/client/scheduler makes or accepts a connection to its peer, a `core.Comm` object is created and returned. This is done in, for example, the `connect()` method:

https://github.com/dask/distributed/blob/0e10478ee9a69da2d064d56d20bb142d1fd49a06/distributed/comm/mpi.py#L583-L585

The newly created `core.Comm` object reads and store the MPI rank of its peer. This is important since everytime a message is communicated through this object, MPI's send and receive method will require this peer rank in the MPI's `COMM_WORLD` communicator.

Ian's code does not use any out-of-band connection establishment. I had a quick glance and as far as I can tell, the first message received with the source wildcard `MPI.ANY_SOURCE` helps determine which comm object is connected to which remote process (it's rank). This is obtained from the status object.

### Message Sending/Receiving: ###

Both PRs use the `COMM_WORLD` communicator for all communication. It uses mpi4py's `Send()` and `Recv()` methods. Ian's MPI code uses mpi4py's `send()` and `recv()` functions. mpi4py's `Send()/Recv()` methods communicate buffer-like data compared to `send()/recv()`, which are used to communicate Python objects. Communicating Python objects will include serialization/de-serialization of data (pickle/unpickle) and hence will have overhead. Ian's code has comments about that:

https://github.com/dask/distributed/blob/2f7f8d08aea3b37609fbdff35db776747becce07/distributed/comm/mpi.py#L61-L65

### Large Messages with MPI ###

While developing MPI4Dask, we observed that Dask can exchange really large messages (multiples of GBs). The MPI point-to-point communication uses an `int count` to express the size of data being sent and received in bytes. Hence the actual data sent/received using MPI cannot be larger than the largest positive integer that fits in an `int`. For this reason, we had to implementing chunking of data if the message is larger than an adjustable threshold.

Ian's MPI device does not handle that because it has not been tested for such applications.

The functions `mpi_send_large` and `mpi_recv_large`, in MPI4Dask, exist because of this.

https://github.com/dask/distributed/blob/0e10478ee9a69da2d064d56d20bb142d1fd49a06/distributed/comm/mpi.py#L395-L453

### Dask-based GPU Applications ###

MPI4Dask was tested with GPU-based applications.

As far as I can tell, Ian's device currently does not have support for GPU-based applications.

### Polling Interval ###

Overall the MPI send and receive co-routine works as follows:

```
request = comm.Isend([buf, size], dest, tag)
status = request.Test()

while status is False:
    await asyncio.sleep(0)
    status = request.Test()
```

There is a discussion on polling interval at PR (https://github.com/dask/distributed/pull/4142). We have not found this to be an overhead. Essentially having `await asyncio.sleep(0)` is yielding the CPU for other tasks that might be executed by the asyncio main loop. Not having this line or simply having `sleep(0)` will obviously lead to a deadlock. I do realize though there might be applications where this polling interval might become an important configuration parameter. More investigation needs to be done with applications to figure out the impact.

Ian's MPI device has 5 or 10 ms polling interval.

### Tags per Connection ### 

MPI uses tags to segregate messages communicated on a connection. Here connection refers to a point-to-point connection between a specific sender and a receiver. In order to differentiate these messages, tags are used.

In the MPI backends, one connection is used to send multiple messages. Also, there can be multiple connections between MPI processes (workers/clients/scheduler). Hence in order to provide safe-space to these concurrent messages between a pair of workers (for example), it is important to assign unique tags.

Both MPI backends assign dynamic tags everytime a new connection is built. MPI4Dask assigns a fixed number tags per connection. This is defined here:

https://github.com/dask/distributed/blob/0e10478ee9a69da2d064d56d20bb142d1fd49a06/distributed/comm/mpi.py#L44

The reason for doing this is related to implementing chunking for large messages. This needs a better solution in the future since it is possible that one connection attempts to send more messages than this constant. 

Ian's code has a different tag generation mechanism that I do not fully understand. 

### MPI4Dask Performance ###

Couple of application-level comparisons for MPI4Dask can be seen here:

- [cuDF Merge](http://hibd.cse.ohio-state.edu/performance/mpi4dask-cudf/)
- [cuPy Sum and Transpose](http://hibd.cse.ohio-state.edu/performance/mpi4dask-cupy/)

These evaluations were done on an internal cluster (at OSU) and TACC Frontera. Graphs for both pages can be seen at links above.

A paper with these results can be seen [here](https://arxiv.org/pdf/2101.08878.pdf).

There are also some performance numbers between UCX-Py and async send/recv co-routines on top of MPI. Figures 5 and 6 in the [MPI4Dask paper](https://arxiv.org/pdf/2101.08878.pdf) have those results.

jakirkham:
Thanks for writing that up Aamir! 😄 Let's see what Ian adds to that 🙂

Just to touch on a few things, generally I think there is interest in having some sort of MPI based communication layer for Dask. So I think there is agreement on that. Ideally this is something that both yourself, Ian, and colleagues of both of you are happy with. Hoping this discussion can help us bridge these two implementations and determine what that final result looks like.

It's interesting that you have made use of asyncio to handle connection establishment. We currently use Tornado for this on TCP, but have recently been looking into removing that and using asyncio directly ( https://github.com/dask/distributed/issues/4513 ). A question that comes from this is whether we can establish a general framework for different communication protocols to use that is more asyncio-centric. Again we are still early in thinking about this problem. So would be curious to hear any thoughts you or others may have on this. Feel free to chime in on that issue as well

IIUC you are referring to Dask's serialization mechanisms `to_frames` and `from_frames` to handle object serialization and collect out-of-band frames for transmission. Is that right? If so, yes using this is preferable. We've been making some improvements to cutdown overhead here where possible ( mainly by handling more stuff as part of MsgPack serialization https://github.com/dask/distributed/pull/4531 ). There are likely more improvements to come. Though would expect communication using `to_frames` and `from_frames` is agnostic to these changes. Just some information that may be of interest to you.

Thanks for pointing that out. So how are you handling GPU base communication? Does this use NVLink and IB? Do you need to check whether something is on device? Also worth noting there's a special `"cuda"` serializer (see [usage here]( https://github.com/dask/distributed/blob/6f95c6323454825f1a64b554ffc7267faeb27de5/distributed/comm/ucx.py#L191 ) and [framework here]( https://github.com/dask/distributed/blob/5ea74cd7ec3cf5b90c5e6d5a31fb7dc3888ea242/distributed/protocol/cuda.py )) that will keep objects on device as part of serialization. The frames will then be device buffers. Depending on how MPI is handling the serialization it may be possible to avoid synchronization (for example we often use [RMM]( https://github.com/rapidsai/rmm ) to back GPU memory, which is stream ordered).

Interesting would be curious to know more about the applications Ian had in mind where polling overhead is an issue. Something else to point out here is we added support for using uvloop for the event loop ( https://github.com/dask/distributed/pull/4448 ). So if one has uvloop installed and enable it in the Distributed config, uvloop will be used for the event loop. Not sure if this is relevant for the overhead Ian observed, but thought it worth mentioning.

Anyways not sure how relevant that context is for you or Ian, but I hope that sheds some light on what is going on under-the-hood in Dask. Hope that helps.

Excited to see this work progress! 😄

aamirshafi:
Thanks @jakirkham for your response.

> Just to touch on a few things, generally I think there is interest in having some sort of MPI based communication layer for Dask. So I think there is agreement on that. Ideally this is something that both yourself, Ian, and colleagues of both of you are happy with. Hoping this discussion can help us bridge these two implementations and determine what that final result looks like.
>

We are happy to contribute.

> IIUC you are referring to Dask's serialization mechanisms `to_frames` and `from_frames` to handle object serialization and collect out-of-band frames for transmission. Is that right? If so, yes using this is preferable. We've been making some improvements to cutdown overhead here where possible ( mainly by handling more stuff as part of MsgPack serialization #4531 ). There are likely more improvements to come. Though would expect communication using `to_frames` and `from_frames` is agnostic to these changes. Just some information that may be of interest to you.

Both implementations make use of `to_frames` and `from_frames' to serialize/de-serialize messages. In that sense, these devices should be able to make use of new developments (#4531).

Out-of-band in the context of MPI4Dask meant that we used the connection provided by `asyncio.open_connection()` and `asyncio.start_server()` to exchange messages needed for initial connection setup. In-band messages for MPI4Dask are obviously data exchanged using MPI's send and receive methods. 

Another related point was that mpi4py provides two variants of send and receive methods. The first variant begins with capital letters (like `Send()` and `Recv()`) and the second variant  begins with small letters (like `send()` and `recv()`).  The second variant is used to communicate data to/from Python objects and hence would naturally have overheads associated to serialization/de-serialization. My point was not related to `to_frames` and `from_frames` -- both implementations support that.

> Thanks for pointing that out. So how are you handling GPU base communication? Does this use NVLink and IB? Do you need to check whether something is on device? 

For the native MPI library, we have used the [MVAPICH-GDR](http://mvapich.cse.ohio-state.edu/userguide/gdr/) library that supports GPUDirect RDMA, CUDA IPC, and GDRCopy. This library figures out the optimal communication path (NVLink or IB) between GPUs.

> Also worth noting there's a special `"cuda"` serializer (see [usage here](https://github.com/dask/distributed/blob/6f95c6323454825f1a64b554ffc7267faeb27de5/distributed/comm/ucx.py#L191) and [framework here](https://github.com/dask/distributed/blob/5ea74cd7ec3cf5b90c5e6d5a31fb7dc3888ea242/distributed/protocol/cuda.py)) that will keep objects on device as part of serialization. The frames will then be device buffers. Depending on how MPI is handling the serialization it may be possible to avoid synchronization (for example we often use [RMM](https://github.com/rapidsai/rmm) to back GPU memory, which is stream ordered).

MPI4Dask is using this mechanism since we used UCX device as the starting point and tried to make minimal changes to it.

> Interesting would be curious to know more about the applications Ian had in mind where polling overhead is an issue. Something else to point out here is we added support for using uvloop for the event loop ( #4448 ). So if one has uvloop installed and enable it in the Distributed config, uvloop will be used for the event loop. Not sure if this is relevant for the overhead Ian observed, but thought it worth mentioning.
> 
> Anyways not sure how relevant that context is for you or Ian, but I hope that sheds some light on what is going on under-the-hood in Dask. Hope that helps.

Thanks for your comments.

> 
> Excited to see this work progress! 😄

ianthomas23:
@jakirkham and @aamirshafi I did post a comment on my PR a few weeks ago that you may not have read but is relevant here.  Since writing that PR I have moved on to other things and all of my available open source contribution time is spent on those (non-dask) projects.  So I am happy to answer a few questions, but I don't envisage returning to this work.

Firstly I should say that my implementation was something that I wrote one week sat on my sofa to prove that it was possible.  There is no corresponding academic paper and I had no co-workers to help me.  It certainly isn't as polished, documented or tested as this PR.

> MPI4Dask is based on the UCX communication device and hopefully that makes it easier to review.

Mine is based on the TCP comms code.

> Ian's code does not use any out-of-band connection establishment.

If I understand you correctly then I think mine works in exactly the same way, just like TCP that it is based on.

> Ian's MPI code uses mpi4py's send() and recv() functions

That's right, so I am using mpi4py's pickling and unpickling.  I have the option of using dask's `to_frames` and `from_frames` (like TCP comms) but that proved to be slower.

> Ian's MPI device does not handle that because it has not been tested for such applications.

Correct, not tested using big data sets requiring large messages.

> As far as I can tell, Ian's device currently does not have support for GPU-based applications.

Correct, mine in "plain vanilla" MPI.

> Ian's MPI device has 5 or 10 ms polling interval.

I don't know where these numbers come from.  Sending has a 1 ms polling interval (but would probably need to be tuned):

https://github.com/dask/distributed/blob/2f7f8d08aea3b37609fbdff35db776747becce07/distributed/comm/mpi.py#L72-L74

Receiving is either blocking where this is acceptable, or non-blocking with a polling interval that starts at 1 ms and then doubles each time

https://github.com/dask/distributed/blob/2f7f8d08aea3b37609fbdff35db776747becce07/distributed/comm/mpi.py#L125-L126

I have a debug option to record the time taken between receiving an MPI message by the single `MessageReceiver` and its consumption by the appropriate recipient.  In my tests this was usually less than a millisecond, but occasionally would be ~15 ms, which would definitely need investigation.

> Ian's code has a different tag generation mechanism that I do not fully understand.

`MPIConnector`s obtain a unique tag (from that MPI rank's point of view) from the `TagGenerator` and pass this to the `MPIListener` so that both ends use the same tag.  Hence the combination of (tag, other_rank) is unique from each MPI rank's point of view.

I hope this helps clarify!

aamirshafi:
@jakirkham can you please advise on next steps to have this merged? Any particular requirements?

dask-apps/cudf_merge_mpi.py
```diff
@@ -0,0 +1,594 @@
+"""
```
jakirkham, line 1: Could we move this somewhere else? Maybe [dask-examples]( https://github.com/dask/dask-examples ) would make sense?

dask-apps/cupy_sum_mpi.py
```diff
@@ -0,0 +1,91 @@
+"""
```
jakirkham, line 1: Same question here

distributed/comm/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 
 def _register_transports():
+    from . import mpi
```
jakirkham, line 17: This should probably be in a `try...except...` like `ucx` below as `mpi4py` is likely not always available. Also would suggest moving this next to `ucx` below since `inproc` and `tcp` are always available and should probably be registered first

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+# of tags.
+tag_table = dict()
+
+def synchronize_stream(stream=0):
+    import numba.cuda
+
+    ctx = numba.cuda.current_context()
+    cu_stream = numba.cuda.driver.drvapi.cu_stream(stream)
+    stream = numba.cuda.driver.Stream(ctx, cu_stream, None)
+    stream.synchronize()
```
jakirkham, line 73: Does MPI need to synchronize on the stream before sending? Many things with RMM and cuDF are stream-ordered. So this might not be needed if MPI respects that

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+    size = MPI.COMM_WORLD.Get_size()
+
+    # each process maintains an entry for every other process in the tag_table
+    for peer in range(size):
+        if rank != peer: 
+            tag_table.update({peer : INITIAL_TAG_OFFSET})
+
+    logger.debug(tag_table)
+
+def init_once():
```
jakirkham, line 88: Since a lot of this repeats code in the UCX module, maybe it makes sense to refactor all the common stuff into a utility function that both can call on initialization

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+    global initialized, host_array, device_array
+
+    if initialized is True:
+        return
+     
+    initialized = True
+
+    random.seed(random.randrange(sys.maxsize))
```
jakirkham, line 96: Why is this needed?

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+        if last_blk_size is not 0:
+            num_of_reqs = num_of_reqs + 1
+
+        reqs = []        
+
+        for i in range(num_of_blks):
+            s_idx = (i)   * blk_size
+            e_idx = (i+1) * blk_size
+
+            if type(buf) == rmm._lib.device_buffer.DeviceBuffer:
```
jakirkham, line 331: Why do we need to check the type? This also seems to be reaching into a private module

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+                # need this if because rmm.DeviceBuffer is not subscriptable
+                shadow_buf = rmm.DeviceBuffer(ptr=(buf.ptr+s_idx), \
```
jakirkham, line 333: Any reason not to use `device_array` here as that is defined above?

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+    prefix = "mpi://"
+    comm_class = MPI4Dask
+    encrypted = False
+
+    async def connect(self, address: str, \
+                      deserialize=True, \
+                      **connection_args) -> MPI4Dask:
+
+        rand_num = 0.001 * random.randint(0, 1000)
+        await asyncio.sleep(rand_num)
```
jakirkham, line 588: Why is a random sleep needed?

distributed/comm/mpi.py
```diff
@@ -0,0 +1,781 @@
...
+        host = ensure_concrete_host(host)  # TODO: ensure_concrete_host
+        return self.prefix + unparse_host_port(host, port)
+
+    @property
+    def bound_address(self):
+        # TODO: Does this become part of the base API? Kinda hazy, since
+        # we exclude in for inproc.
+        return self.get_host_port()
+
+class MPI4DaskBackend(Backend):
```
jakirkham, line 750: Is it possible to create an abstract base class that handles MPI and TCP's needs? This should avoid the duplication here

dask-apps/cudf_merge_mpi.py
aamirshafi: Yes we can do that. The only reason to put them here was to have sample code for the MPI backend. 

My colleague @ZackWRyan will help with this PR.

dask-apps/cupy_sum_mpi.py
aamirshafi: This can be moved as well. 

@ZackWRyan

distributed/comm/__init__.py
aamirshafi: Makes sense.

distributed/comm/mpi.py
aamirshafi: Yes that can be done as well. The common code inside `init_once` is mostly related to initializing `host_array` and `device_array`.

distributed/comm/mpi.py
aamirshafi: This is related to a followup comment. The MPI backend currently has a random wait before connecting to other workers. This should not be needed as far as I can tell. We are investigating this further and will get back here.

distributed/comm/mpi.py
aamirshafi: MPI cannot send messages larger than `2GB - 1` because it uses a signed integer to specify the number of bytes sent/received. For this reason, the MPI device needs to chunk data into blocks and communicate. 

This chunking works as long as the datatype communicated is sub-scriptable (like cuPy for example). Sub-scriptable here means using the `:` operator to index into specific regions of the buffer. However if the data being communicated is `rmm._lib.device_buffer.DeviceBuffer`, then its not sub-scriptable and hence a special case in the code. In this case, the device directly initializes a shadow buffer (of type `rmm.DeviceBuffer`) and communicate the next chunk.  

Is there a better way to do this?

distributed/comm/mpi.py
aamirshafi: I explained this earlier. We should be able to avoid this sleep. I will update here on this.

distributed/comm/mpi.py
aamirshafi: This duplicate comment is coming from UCX device. We can come to this after addressing other comments.

distributed/comm/mpi.py
aamirshafi: Same reason as above. If there's a suggestion to improve, please let us know.

@ZackWRyan

distributed/comm/mpi.py
aamirshafi: We are investigating and will update here.

distributed/comm/mpi.py
ZackWRyan: A brief search didn't find any other ways around the special case for RMM buffers.  One thought would be to convert RMM's buffer to some other type of buffer (e.g. PyBuffer, which is sub-scriptable/sliceable), but as far as I can tell, the other buffer types that I considered require extra communication and/or fully transferring the data back to CPU memory.  That would have negative impacts on performance, so I'd advice just leaving it like this.

distributed/comm/mpi.py
ZackWRyan: To add to what Aamir said, to replace subscripting/slicing for RMM buffers it's necessary to create these 'shadow' buffers that are actually just chunks of the larger RMM buffer, and in order to do that without copying we need to specify a pointer to an intermediate spot in the buffer.  `device_array` doesn't allow us to specify a pointer, only a size, partially because that `device_array` might actually be a Numba array (not RMM), and as far as I can tell, Numba doesn't support creating device buffers at specific pointer locations.

dask-apps/cudf_merge_mpi.py
ZackWRyan: I'll remove both of these examples from this PR, we can make a separate PR on dask-examples to go along with this one later.

distributed/comm/mpi.py
ZackWRyan: Initial tests are showing that the random seed and connection delay are indeed not necessary, once more results come back and we're confident it still works I'll make the commit to remove them.

sanjeev-one:
Bump on this. Is mpi able to be used for dask to communicate between workers?

GPUtester:

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

PR #4465, opened 2021-01-28
Compress serialized config data with zlib

jacobtomlinson:
While working on `dask-cloudprovider` I've noticed there is a limit to the amount of [user_data](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html) you can pass to AWS. Among other things, a serialized copy of the local Dask config is passed via the user_data and depending on how much config a user has can result in the AWS API rejecting the call.

dask/dask-cloudprovider#249

In an attempt to mitigate this I've added `zlib` compression to the config serialization utility methods. From the limited testing I've done locally I've seen a ~60% reduction in size.

```python
>>> from distributed.utils import serialize_for_cli
>>> import dask.config
>>> len(serialize_for_cli(dask.config.global_config))  # Before change
5452
>>> len(serialize_for_cli(dask.config.global_config))  # After change
2256
```

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -28,6 +28,7 @@ import pkgutil
 import base64
 import tblib.pickling_support
 import xml.etree.ElementTree
+import zlib
 
 try:
     import resource
@@ -1467,7 +1468,7 @@ def serialize_for_cli(data):
     serialized_data: str
         The serialized data as a string
     """
-    return base64.urlsafe_b64encode(json.dumps(data).encode()).decode()
+    return base64.urlsafe_b64encode(zlib.compress(json.dumps(data).encode())).decode()
 
 
 def deserialize_for_cli(data):
@@ -1482,7 +1483,7 @@ def deserialize_for_cli(data):
     deserialized_data: obj
         The de-serialized data
     """
-    return json.loads(base64.urlsafe_b64decode(data.encode()).decode())
+    return json.loads(zlib.decompress(base64.urlsafe_b64decode(data.encode())).decode())
 
 
 class EmptyContext:


quasiben:
Thanks @jacobtomlinson -- do you think we need to handle cases where zlib is not installed ?

jacobtomlinson:
@quasiben I was under the impression that `zlib` was in the standard library?

quasiben:
@jacobtomlinson , apologies.  It is in the standard library!

mrocklin:
Maybe we should consider some more general purpose solution to this?  Maybe
we should write things into a file and use the config system?

jacobtomlinson:
I agree there is probably a better way to pass this config along. This PR is just treating a symptom. 

Passing a file as part of a VM creation on any cloud is non-trivial though. I could imagine passing it through some storage like S3, but that adds complexity.

ZirconCode:
Coming from [here](https://stackoverflow.com/questions/65982439/dask-aws-cluster-error-when-initializing-user-data-is-limited-to-16384-bytes) (thanks @quasiben), I can confirm that this PR solved my issue, at least on this side of things there is no more error. However my machine is stuck at `Waiting for scheduler to run at ip:port`, while on aws it is "shutting down", I believe I also have to pull this change into my packer build, but am somewhat at a loss on how to do this.

jacobtomlinson:
@ZirconCode glad this PR got you to the next error. Could you raise an issue on dask-cloudprovider about this?

eric-valente:
Is there anything I can do in the meantime to get the EC2Cluster class to work?  Running into this issue no matter what I try.

Issue #4446, opened 2021-01-21
Resource constraints and work stealing not working as expected, the cluster ends up stuck


dtenchurin:
Hello,

Since it is pretty hard to reproduce this at the moment, but the problem is happening often enough for us to stop ignoring it. 

Setup: 
1) 100 16 core boxes with single worker each started with --resources 'cpu=16'
2) 3 task types: A: {'cpu':16}, B: {'cpu':4}, C: {'cpu':1}, tasks B and C are python subprocess calls, and task A is a function with a call to sklearn .fit().
3) work stealing is enabled
4) there are about 2000+ tasks in the queue normally

Observed bug:
At some point the available_resources dictionary in the worker object becomes incorrect. Not sure what leads to this state but something is broken in the state transition mechanism so that it would look like a particular worker has more available resources than it should. This cases overscheduling of tasks into the worker.

Example:
2021-01-21 07:52:02.021 INFO:  {'tcp://172.24.0.144:39033': (9,)}, available_resources: {'cpu': 3.0}, used_calculated_outside_dask: 24
2021-01-21 07:52:02.021 INFO:   i('A-f29492ccfd701faea20c09e4193c0db9', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-2636fbf9c5cf07db2eb3088985835730', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-28deaee4c9494ef2b376f8246894e5cb', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-2561aa2b50164e905a0582810c91d8c7', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-c75dcc64cdc7a19890fd516fede9fa28', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-75251f721ba1b411c80e7bb08fa79245', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-3d87f41690c6f1a08941628371c6e491', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-2fc85a2cee66cefff07fa4d6db5e38bc', 'executing')
2021-01-21 07:52:02.021 INFO:   i('C-7d4de834d8bb02db9a4f097e8815c5cc', 'executing')

An even bigger problem is that after some time of running these many tasks, the worker becomes unresponsive, and the scheduler starts reporting:
`OSError: Timed out during handshake while connecting to tcp://172.24.0.147:35091 after 25 s`
while clients get:
[...]

PR #4470, opened 2021-01-29
Add explicit `fetch` state to worker TaskState

gforsyth:
Taking a pass at the new control flow that @fjetter laid out in #4413.  Many thanks to him for the very detailed write-up and diagnosis of the issue.

To disambiguate a task that is to be executed vs. a task which is a data container (a dependency to be fetched from another worker) I've added a new state called `fetch`.  The existing state `waiting` now only refers to tasks which will be executed on the worker.

I think the `add_task` function needs to be cleaned up and I also agree with @fjetter that we should have a similar set of chained recommendations like the scheduler does, but hopefully this is a step in that direction.

`test_failed_workers.py::test_broken_worker_during_computation` fails one every 30 runs or so on my machine, I'd like to fix that before this gets merged in, but the more explicit states makes hunting down inconsistent state a fair bit easier.

Also needs updates to docs for the new state and the new worker flow


fjetter:
First pass looks good. Will give it a deeper review once you are more confident with the changes. If there are still failing tests this may trigger a bunch or more changes.
I can highly recommend the suggestion to run the flaky tests in an ipython shell to speed up development. At least the `gen_cluster` annotated tests can be simply imported and executed as an ordinary function

gforsyth:
Thanks for the initial pass, @fjetter -- I ran `test_broken_worker_after_computation` and test_who_has_clears_after_failed_connection` 100x each without any issue, so I think this is stable and ready for a review.

I'd like to make the state validation methods a bit more detailed but I think for now that will have to wait until we have the chained recommendation system in place and I'm trying not to do too much in this PR.  

Thanks again for all the work in diagnosing these issues!

mrocklin:
Any interest in rewriting this doc section?
https://distributed.dask.org/en/latest/worker.html#internal-scheduling

gforsyth:
> Any interest in rewriting this doc section? https://distributed.dask.org/en/latest/worker.html#internal-scheduling

yes, on my list

distributed/worker.py
```diff
@@ -1496,26 +1505,45 @@ def add_task(
...
+
+                    # link up to child / parents
+                    ts.dependencies.add(dep_ts)
+                    dep_ts.dependents.add(ts)
+
+                    # possible that a new task has asked for a dependency we're already
+                    # planning on fetching -- make sure that task knows it has to
+                    # wait for its dependencies to arrive
+                    if dep_ts.state in ("fetch", "flight"):
+                        ts.waiting_for_data.add(dep_ts.key)
```
fjetter, line 1538: IIRC, `waiting_for_data` to be empty is one of the conditions for the transition `waiting -> executing` and I'm wondering if this condition shouldn't also include `waiting` and `executing`

distributed/worker.py
```diff
@@ -1496,26 +1505,45 @@ def add_task(
...
+
+                if dep_ts.state in ("fetch", "flight"):
```
fjetter, line 1540: The same condition is checked in the body of the else clause above. I'm wondering if this cannot be fused

distributed/worker.py
```diff
@@ -1496,26 +1505,45 @@ def add_task(
...
+                    # if we _need_ to grab data or are in the process
+                    dep_ts.who_has.update(workers)
+
+                    for worker in workers:
+                        self.has_what[worker].add(dep_ts.key)
+                        # if dep_ts.state != "memory":
```
fjetter, line 1546: dead code/comment

distributed/worker.py
```diff
@@ -1555,10 +1586,104 @@ def transition(self, ts, finish, **kwargs):
...
+    def transition_new_fetch(self, ts):
+        try:
+            if self.validate:
+                assert ts.state == "new"
+                assert ts.runspec is None
+
+            for dependent in ts.dependents:
+                dependent.waiting_for_data.add(ts.key)
+
+            self.data_needed.append(ts.key)
```
fjetter, line 1612: I noticed the `data_needed` to be occasionally populated by duplicates. Since it is a list, we cannot really guarantee anything else. I believe this is not really harmful but we might want to consider a `key in data_needed` check

distributed/worker.py
```diff
@@ -1555,10 +1586,104 @@ def transition(self, ts, finish, **kwargs):
...
+                import pdb
+
+                pdb.set_trace()
+            raise
+
+    def transition_fetch_waiting(self, ts, runspec=None):
+        """This is a rescheduling transition that occurs after a worker failure.
+        A task was available from another worker but that worker died and the
+        scheduler reassigned the task for computation here.
+        """
```
fjetter, line 1626: Just a comment

I believe this is one of the crucial transitions which mess up the state occasionally. Especially since if this transition is done, chances are there is still a coroutine in the background trying to fetch the task. If, for instance, this transition was just triggered because a worker was unresponsive, that background task may still fetch the result and we're faced with a `fetch -> waiting -> memory` or worse 
```
fetch -> waiting     -> memory
                    -> executing -> memory
```

I believe there is little we can do about these things other than handling them properly. In particular this means we might have a result already in memory even though its task is executing still

distributed/worker.py
```diff
@@ -1555,10 +1586,104 @@ def transition(self, ts, finish, **kwargs):
...
+            self.waiting_for_data_count += 1
+        except Exception as e:
+            logger.exception(e)
+            if LOG_PDB:
+                import pdb
+
+                pdb.set_trace()
+            raise
+
+    def transition_fetch_waiting(self, ts, runspec=None):
```
fjetter, line 1622: `runspec` is not actually allowed to be `None` here, is it?

distributed/worker.py
```diff
@@ -1555,10 +1586,104 @@ def transition(self, ts, finish, **kwargs):
...
+                dependent.waiting_for_data.discard(ts.key)
+        except Exception as e:
+            logger.exception(e)
+            if LOG_PDB:
+                import pdb
+
+                pdb.set_trace()
+            raise
+
+    def transition_flight_waiting(self, ts, runspec=None):
```
fjetter, line 1652: same here, is `runspec` allowed to be `None`?

distributed/worker.py
```diff
@@ -1718,7 +1844,19 @@ def transition_ready_waiting(self, ts):
         """
         This transition is common for work stealing
         """
-        pass
+        ts.runspec = None
+
+    def transition_waiting_new(self, ts):
```
fjetter, line 1849: is a transition back to `new` actually allowed? I consider `new` as a pseudo/virtual state but I wouldn't expect a transition to actually move a task to `new`, with the sole exception of a "catch all" mechanism, similar to the scheduler task state machine where the `start -> release -> * -> finish` is a a catch all transition path but we haven't implemented this kind of thing for the worker, yet.

distributed/worker.py
```diff
@@ -1718,7 +1844,19 @@ def transition_ready_waiting(self, ts):
...
+        """
+        Common in work stealing
+        """
+        ts.runspec = None
+
+    def transition_constrained_waiting(self, ts):
```
fjetter, line 1855: I'm not sure if this is possible. Constrained is more or less the twin state of a `waiting` task with the distinction that the constrained task has resource constraints to fulfil. If that's the case, this transition would mean we're changing the definition of the task by stripping the resource constraints

distributed/worker.py
```diff
@@ -2149,7 +2287,8 @@ async def gather_dep(self, worker, dep, deps, total_nbytes, cause=None):
                         self.release_key(d)
                         continue
                     elif ts.state not in ("ready", "memory"):
-                        self.transition(ts, "waiting", worker=worker, remove=not busy)
+                        # "waiting" or "fetch"?
+                        self.transition(ts, "fetch", worker=worker, remove=not busy)
```
fjetter, line 2291: afaiu it should be `fetch` but the `remove` toggle plays a crucial role here. It might no longer be necessary, in fact.

distributed/worker.py
gforsyth: I think it could, but not if we add `waiting` and `executing` to the check above (which I think is the right things to do)

distributed/worker.py
gforsyth: Yeah, this is a leftover I forgot to remove, it's not reachable via the `_transitions` dict.  I was starting down the state machine path and pulled back to leave that for a later PR.  I'll rip this out.  Thanks!

distributed/worker.py
gforsyth: No, it isn't, I can add a check for that

fjetter: My point is rather why the default value if it must be provided anyhow

distributed/worker.py
gforsyth: This should only be called if constrained task is stolen -- the current logic is there is that stolen tasks revert to a state of `waiting` and are then released. 
i think in this case we're only stripping the resource constraints because they're being enacted on a different worker.

distributed/worker.py
gforsyth: yeah, I agree it's a `fetch` here -- I also tried removing the `remove` toggle in 2a8c0e2 and it looks (tentatively) that you're correct that it is no longer needed now that `waiting` and `fetch` are disambiguated.

distributed/worker.py
gforsyth: Ah, gotcha.  I was just following the convention in the other transition functions, but you're right.  Much better to be explicit here that it is a required value.

distributed/worker.py
gforsyth: fused

distributed/worker.py
gforsyth: I think that if I include `waiting` and `executing` in here that it might cause weird transitions -- if the dependency is `executing` and then we add it to `waiting_for_data` while it is moving to memory, it will cause us to try to re-run this and we'll end up with a `waiting -> memory` transition.

distributed/worker.py
fjetter: Even if it is stolen, the resource constraint of a task is defined on task initialisation, effectively by the user. ressource counting is done by the worker but the task should not be modfiied

it is managed here
https://github.com/dask/distributed/blob/98570fbcd662bc0fe468ccba9d57d582d2a3c678/distributed/worker.py#L1725-L1726
and here
https://github.com/dask/distributed/blob/98570fbcd662bc0fe468ccba9d57d582d2a3c678/distributed/worker.py#L1740-L1741
but I would consider the field `resource_restrictions` to be immutable (if only that was easily possible in python :) )

Therefore, if my reasoning is correct, this might overshadow another wrong transition (maybe key is forgotten and improperly resubmitted?)

fjetter: >the current logic is there is that stolen tasks revert to a state of waiting and are then released.

Do you mean this line?

https://github.com/dask/distributed/blob/98570fbcd662bc0fe468ccba9d57d582d2a3c678/distributed/worker.py#L2285

I think it is wrong... :( This should show up somewhere in a verify task method. I think there should be something like

```python
def validate_task_waiting(self, ts):
    ...
    assert ts.resource_restrictions is None # // or empty dict, dont know
    ...
```

distributed/worker.py
fjetter: sorry, I think my comment was wrong. This is an issue about semantics. I perceived `waiting_for_data` to be the same as "dependencies which are not yet in memory" but it is rather a "dependency which is not yet in memory and must be fetch from another worker". If this is true, your code changes make sense and I think there should be no other state in here

fjetter:
The constrained->waiting thing might connect to #4446 

everything else LGTM

distributed/worker.py
gforsyth: That's a very good point re: `resource_restrictions` -- I'll look again at the stealing code.
I don't know why the transition to `waiting` was in there -- it seems like after we have confirmation that the task has been stolen, we should release the key on the victim side (along with whatever `available_resources` might have been allocated for use)

gforsyth: Also, it looks like `steal_request` doesn't need to worry about releasing `available_resources` because they aren't "taken" until the task transitions from `ready -> executing`, at which point it's no longer eligible to be stolen

gforsyth:
Damn, moving to the non-transition `steal_request` flow lead to a deadlock and I can't track it down locally.

@fjetter you added this in 1fc1b4645670aae978168d06886556195dfdec26 as a part of #4432 -- do you remember why that transition was there?  Was it solely to remove the `runspec`?  Or to handle dependents a little more gracefully?

gforsyth:
Hmm, conversely, it might have been a fluke? It looks like perhaps `pytest test_client_executor.py::test_cancellation` hung, which seems possibly unrelated

fjetter:
> @fjetter you added this in 1fc1b46 as a part of #4432 -- do you remember why that transition was there? Was it solely to remove the runspec? Or to handle dependents a little more gracefully?

Well, the runspec is reset a few lines above the release and yes, I tried to map as much as possible via transitions instead of inplace mutations to ensure all connected state information is properly modified (dependnts, dependencies, waiting for, etc.). I am still convinced that the way we're releasing tasks is also a bit buggy or at the very least not very transparent. I added this commit trying to handle this more transparently. It is currently otherwise very difficult to know what the release actually does. For example, is the key still in self.tasks or not?? That is determined by `ts.dependents` but why do I need to check this attribute myself before calling that method? Sometimes you want to be _some_ of release_key but all of it but it's not very clear to me when we do need what. Eventually we might want to implement this more clearly by having another virtual state like `forgotten`, similar to how the scheduler deals with this but that's yet another change. I'm drifting off...

Ultimately, I can't tell you which direction is the right one, that's why my first PR escalated so badly :(

-----

> Damn, moving to the non-transition steal_request flow led to a deadlock and I can't track it down locally.

Do you have an idea which test deadlocked? I can try to run it on my machine and see if I can reproduce

>  It looks like perhaps pytest test_client_executor.py::test_cancellation hung, which seems possibly unrelated

What my last efforts debugging this taught me is that there is no such thing as an unrelated deadlock. It's all just a big ball of spaghetti ;)
I'll try to reproduce (if I can, I'll try the same on master and we'll see where we are)

gforsyth:
> there is no such thing as an unrelated deadlock

oof.  too true.  I was thinking it might've been a known flaky test but github actions doesn't hang on to logs for super long, I went looking for a previous similar timeout but didn't find one

gforsyth:
Hey @fjetter -- just rebased against master and should be up to date now.  If you have the opportunity to run this in one of your bigger workloads and see how it performs, that would be very helpful.  I've run the test-suite locally a few hundred times at this point (focusing on work-stealing, worker failure, and regular worker tests) and things seem to be stable.

fjetter (commented):
I'll test this on our workloads and will report back.

fjetter:
Unfortunately, I hit https://github.com/dask/distributed/issues/4538 when trying this so no results, yet

fjetter:
I reverted the commit which likely caused #4538 to run some tests with this PR. Unfortunately, the very first try already hit a deadlock. I'm in an upscaling scenario where I start with one worker and am adding more and more, i.e. heavy work stealing. After some upscaling, the cluster distributes some of the tasks and works on them but only on about half. View from the scheduler (after it became stale) shows that the worker is supposedly still processing ~70k tasks and doesn't have any in memory. However, when asking the worker about its task, I can see that it has two tasks in memory. The state between scheduler and worker seems to have drifted significantly.

Asking scheduler/worker about one of the processing tasks story, I get only a single entry  `[("<key>", 'release-key')]`. I'll try to dig a bit if I can find more logs or anything.

gforsyth:
Thanks for the report @fjetter -- and for the description of the failing test.  I'll also try to work up a test case that follows that pattern and see if I can reproduce locally.

fjetter:
I had another look, couldn't find any more useful logs but I noticed that our scheduler was overloaded the entire time, for whatever reason. that at least explains why the work stealing was working so poorly, but not why the initial worker was lazy and didn't process any results, nor why everything was entirely out of sync. Regardless, what I've seen didn't look healthy. I'll try to perform a few more test runs later today/tomorrow

fjetter:
Had another test run which ran through but I received a lot of logs like

```
('fetch', 'memory')
  File "/.../lib/python3.6/site-packages/distributed/utils.py", line 655, in log_errors
    yield
  File "/.../lib/python3.6/site-packages/distributed/worker.py", line 2256, in gather_dep
    self.transition(ts, "memory", value=data[d])
  File "/.../lib/python3.6/site-packages/distributed/worker.py", line 1576, in transition
    func = self._transitions[start, finish]
KeyError: ('fetch', 'memory')
```

gforsyth:
I hit the `KeyError` reliably when running `tests/test_steal.py::test_steal_twice` -- trying to figure out the root cause now, but it definitely occurs as a knock-on effect from a dependency being stolen, so that's encouraging.

gforsyth:
Changes:
- If there's a knock-on `release_key` call but the task is in flight, just leave it alone.  Worst case is an extra entry in the self.tasks dictionary
- Reserve `self.data_needed` for tasks which will be executed locally (that is, only list the tasks for which we need to fetch data, don't include the tasks to-be-fetched)
- Don't call `release_key` in a transition function, ever.

gforsyth:
@fjetter -- if you have an opportunity later this week to run your heavy stealing workload against this branch that would be a big help.  Thanks!

fjetter:
I am repeating myself but I believe we need to eventually rework the release mechanism. I have the feeling this is mostly educated guessing on when we *must* release a key, when we *should* and when we *must not*  release it. Not sure how you feel about this. Either way, I'm glad if it works and reworking this is definitely out of scope for this PR

> if you have an opportunity later this week to run your heavy stealing workload against this branch that would be a big help. Thanks!

Yes, I think I can find time for this later today

gforsyth:
> I am repeating myself but I believe we need to eventually rework the release mechanism.

Oh, definitely.  I think reworking the release mechanism and adding in a chained-transition system are the next steps once this is stable.

fjetter (approved):
I gave this another spin and couldn't find anything wrong with it. That doesn't necessarily mean a lot but it's a very good start.

I'd be willing to merge this now but there is a release scheduled for tomorrow and I'm wondering if we want to let this sit a while before releasing. Opinions? cc @gforsyth  @jrbourbeau

gforsyth:
> I gave this another spin and couldn't find anything wrong with it. That doesn't necessarily mean a lot but it's a very good start.

:tada: wooooooo!

I'll defer to @jrbourbeau on whether we want to try to sneak this in to the next release or let it sit on master for a little bit

jrbourbeau:
I'd prefer to wait until after releasing to merge this in if that's alright

gforsyth:
Hey @jrbourbeau -- are we good to merge this now?

gforsyth:
Just a gentle ping on this to see if anyone has time to review further and/or merge

fjetter:
I would go ahead and merge this. I'm just a bit hesitant since I'm lately not very involved and don't know if this interferes with any other change. @jrbourbeau can you give a quick "go ahead" / merge unless you are aware of any other big change which could interfere?

distributed/worker.py
```diff
@@ -1473,13 +1472,20 @@ def add_task(
...
                 self.log.append((key, "new"))
                 self.tasks[key] = ts = TaskState(
                     key=key, runspec=SerializedTask(function, args, kwargs, task)
                 )
-                ts.state = "waiting"
+                self.transition(ts, "waiting")
+
+            # TODO: move transition of `ts` to end of `add_task`
+            # This will require a chained recommendation transition system like
+            # the scheduler
```
jrbourbeau, line 1488: Does this need to be done in this PR or is this follow-up work?

distributed/worker.py
```diff
@@ -1499,26 +1505,42 @@ def add_task(
...
-                self.log.append((dependency, "new-dep", dep_ts.state))
+                    # link up to child / parents
+                    ts.dependencies.add(dep_ts)
+                    dep_ts.dependents.add(ts)
 
-                if dep_ts.state != "memory":
-                    ts.waiting_for_data.add(dep_ts.key)
-                    self.waiting_for_data_count += 1
+                    # check to ensure task wasn't already executed and partially released
+                    # # TODO: make this less bad
```
jrbourbeau, line 1517: I'm not sure if this is bad ascetically or from a logic perspective. Is there anything else we should do here in this PR?

jrbourbeau (commented):
Thanks @gforsyth and @fjetter for your work here -- apologies for the delayed reply. After a quick look this seems good to merge (there are a couple of `TODO`s where it'd be good to confirm if they are talking about future work or things we should include here)

gforsyth:
Hey @jrbourbeau! Yeah, both of those Todos are for the next PR which is going to add a chained recommendation system for transitions.

jrbourbeau (approved):
Thanks @gforsyth @fjetter!

PR #4472, opened 2021-01-31
DOC: typo, directives ends with 2 colons `::`.

Carreau:
It's mischievous as those are interpreted as comments so perfectly valid syntax; but completely removed from the output. 

Side question, you seem to be using Numpy Doc format in many places, why not use the [Attributes](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) section and maybe less prone to errors ?

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1089,19 +1089,19 @@ class TaskState:
        into the "processing" state and be sent for execution to another
        connected worker.
 
-    .. attribute: metadata: dict
+    .. attribute:: metadata: dict
 
        Metadata related to task.
 
-    .. attribute: actor: bool
+    .. attribute:: actor: bool
 
        Whether or not this task is an Actor.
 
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):
Good catch, thanks for fixing @Carreau! 

I'm not sure why we're not using an "Attributes" section here. I suspect it is not intentional. Any interest is moving these attributes into an "Attributes" section? If not, no worries, the current set of changes is a definite improvement with what we have today

Carreau:
> I'm not sure why we're not using an "Attributes" section here. I suspect it is not intentional. Any interest is moving these attributes into an "Attributes" section?

Maybe later; I'm working on some documentation project and that was triggering errors. I'll try to not get distracted for now but the fact that a `.. attribute::` directive exists will make my other project break later on; at that point i'll likely fix it :-)

jrbourbeau:
Sounds good! Thanks again for cleaning this up 👍

PR #4473, opened 2021-01-31
DOC: Proper numpydoc syntax for distributed/protocol/*.py

Carreau:
Numpydoc return semantically different informations when there is no
space in front of :, in particular it will put the whole line as the
name of the parameter.

This might not be visible in the sphinx theme depending on the theme
chosen.

-- 

Note that those changes have been automatically generated;  
I can easily send a PRs that add on top of this  about 250 other fixes, but don't want to send too big of an update;

### DIFF
--- a/distributed/protocol/compression.py
+++ b/distributed/protocol/compression.py
@@ -147,10 +147,10 @@ def byte_sample(b, size, n):
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


jrbourbeau (approved):
Thanks @Carreau! Any follow up PRs are welcome

PR #4390, opened 2020-12-23
Drop support for Python 3.6

jrbourbeau:
For additional context see the discussion over in dask/community#66. There's a corresponding PR for dropping Python 3.6 in Dask at https://github.com/dask/dask/pull/7006.

Issue #2803, opened 2019-06-25
Add mild type annotations


mrocklin:
We should investigate adding mild type annotations to this project.  By this I mean simple types like `str` and `int` that aren't contentious, and nothing complex.  

We might then investigate the value of adding `mypy` to CI.

I think that this would be a good first step before adding typing annotations to the core dask project.  I think that it makes sense to start here because:

1.  We're more tied to Python 3 due to the async-await syntax
2.  We're slightly less core

jrbourbeau:
It's been a while since this issue was originally opened and since then some type annotations have made there way into the `distributed` code base. If we're going to have type annotations (which I think there's interest in) then we should consider adding a `mypy` check in our CI to make sure they are correct (maybe in our existing `pre-commit` check?)

@crusaderky you and I have talked about this a few times offline. IIRC you mentioned, since `mypy` will effectively ignore non-annotated functions, adding an initial `mypy` check may be relatively straightforward. If that's the case, is this something you're interested in taking on?

PR #4474, opened 2021-02-01
[WIP] Async stream handlers

### LINKED ISSUE #4469: Avoid is_coroutine_function calls in Server
We currently support handlers that are either synchronous or asynchronous 

https://github.com/dask/distributed/blob/1297b18ff09276f6ad1553d40ab3ce77acf0fc0e/distributed/core.py#L567-L572

In #4443 we learned that these checks take up some time.  In a recent run it takes 100ms out of about a 2.5s run, so around 4%

There are a couple of ways to address this:

1. We can try to find some other cheaper check
2. We can make all of our handlers `async def` functions and asyncify the entire codebase (we should also check that this doesn't incur other costs if so)

cc @dask/scheduler-performance

ian-r-rose:
Fixes #4469.

I haven't done much profiling yet to see what kind of performance improvements this has (if any), I'll take a look at that next and post some results here.

### DIFF
--- a/distributed/pubsub.py
+++ b/distributed/pubsub.py
@@ -9,6 +9,8 @@ from .core import CommClosedError
 from .metrics import time
 from .utils import sync, TimeoutError, parse_timedelta
 from .protocol.serialize import to_serialize
+from .scheduler import Scheduler
+from .worker import Worker
 
 logger = logging.getLogger(__name__)
 
@@ -16,7 +18,7 @@ logger = logging.getLogger(__name__)
 class PubSubSchedulerExtension:
     """ Extend Dask's scheduler with routes to handle PubSub machinery """
 
-    def __init__(self, scheduler):
+    def __init__(self, scheduler: Scheduler):
         self.scheduler = scheduler
         self.publishers = defaultdict(set)
         self.subscribers = defaultdict(set)
@@ -35,7 +37,7 @@ class PubSubSchedulerExtension:
 
         self.scheduler.extensions["pubsub"] = self
 
-    def add_publisher(self, comm=None, name=None, worker=None):
+    async def add_publisher(self, comm=None, name=None, worker=None):
         logger.debug("Add publisher: %s %s", name, worker)
         self.publishers[name].add(worker)
         return {
@@ -44,7 +46,7 @@ class PubSubSchedulerExtension:
             and len(self.client_subscribers[name]) > 0,
         }
 
-    def add_subscriber(self, comm=None, name=None, worker=None, client=None):
+    async def add_subscriber(self, comm=None, name=None, worker=None, client=None):
         if worker:
             logger.debug("Add worker subscriber: %s %s", name, worker)
             self.subscribers[name].add(worker)
@@ -62,7 +64,7 @@ class PubSubSchedulerExtension:
                 )
             self.client_subscribers[name].add(client)
 
-    def remove_publisher(self, comm=None, name=None, worker=None):
+    async def remove_publisher(self, comm=None, name=None, worker=None):
         if worker in self.publishers[name]:
             logger.debug("Remove publisher: %s %s", name, worker)
             self.publishers[name].remove(worker)
@@ -71,7 +73,7 @@ class PubSubSchedulerExtension:
                 del self.subscribers[name]
                 del self.publishers[name]
 
-    def remove_subscriber(self, comm=None, name=None, worker=None, client=None):
+    async def remove_subscriber(self, comm=None, name=None, worker=None, client=None):
         if worker:
             logger.debug("Remove worker subscriber: %s %s", name, worker)
             self.subscribers[name].remove(worker)
@@ -100,14 +102,14 @@ class PubSubSchedulerExtension:
             del self.subscribers[name]
             del self.publishers[name]
 
-    def handle_message(self, name=None, msg=None, worker=None, client=None):
+    async def handle_message(self, name=None, msg=None, worker=None, client=None):
         for c in list(self.client_subscribers[name]):
             try:
                 self.scheduler.client_comms[c].send(
                     {"op": "pubsub-msg", "name": name, "msg": msg}
                 )
             except (KeyError, CommClosedError):
-                self.remove_subscriber(name=name, client=c)
+                await self.remove_subscriber(name=name, client=c)
 
         if client:
             for sub in self.subscribers[name]:
@@ -119,7 +121,7 @@ class PubSubSchedulerExtension:
 class PubSubWorkerExtension:
     """ Extend Dask's Worker with routes to handle PubSub machinery """
 
-    def __init__(self, worker):
+    def __init__(self, worker: Worker):
         self.worker = worker
         self.worker.stream_handlers.update(
             {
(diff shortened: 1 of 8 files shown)


distributed/pubsub.py
```diff
@@ -3,20 +3,25 @@
...
 import threading
+import typing
 import weakref
 
 from .core import CommClosedError
 from .metrics import time
 from .utils import sync, TimeoutError, parse_timedelta
 from .protocol.serialize import to_serialize
 
+if typing.TYPE_CHECKING:
```
ian-r-rose, line 14: I found adding type annotations to the `stream_handlers` object was useful in implementing this (cf #2803), allowing type checkers to catch which handlers were async or sync. *However*, I needed this pretty unsightly hack to avoid a circular import, just to get the type name in the module scope. I'm not happy about it, and it could be removed. On the other hand, it's kind of nice for refactoring.

distributed/utils.py
```diff
@@ -289,6 +290,21 @@ def quiet():
     return results
 
 
+T = TypeVar("T")
+
+
+def asyncify(func: Callable[..., T]) -> Callable[..., Awaitable[T]]:
```
ian-r-rose, line 296: In some places I've used an async wrapper to have a lighter footprint on the API, but more functions could just be made async in the first place.

mrocklin: Yeah, I almost raised a question about this until I looked at how these methods were used.  I agree that making them sync by default is probably the right choice.  This seemed like a good solution to me.

mrocklin:
> I haven't done much profiling yet to see what kind of performance improvements this has (if any), I'll take a look at that next and post some results here.

Ooh, fun.  I look forward to seeing which of the many different ways of profiling you choose :)

distributed/pubsub.py
jakirkham: Does `from __future__ import annotations` help at all here? Admittedly that is Python 3.7+ only. Though we are planning to drop Python 3.6 soon ( https://github.com/dask/distributed/pull/4390 )

ian-r-rose: Yeah, I had to remove `annotations` in 4f1944b for 3.6 support. Though it didn't help that much, it mostly meant I didn't have to quote the type names below. The circular import problem remained.

mrocklin:
So, a thought just came to me.  I wonder if all of this could be solved by
lru cache.

distributed/pubsub.py
crusaderky: @jakirkham it doesn't help - you still need those imports. ``from __future__ import annotations`` helps when you have a class accepting itself or returning itself in its methods, or when class A in a module accepts or returns class B which is declared afterwards in the same module.

distributed/pubsub.py
```diff
@@ -3,20 +3,25 @@
 from contextlib import suppress
 import logging
 import threading
+import typing
```
crusaderky, line 6: as a personal esthetic preference, I find that using ``from typing import ...`` is always a nicer option. It does lead to collisions with ``from collections.abc import ...``, which can however be worked around after you drop Python 3.6 support, add ``from __future__ import annotations`` everywhere, and you run a Python 3.9-compatible version of mypy (on python 3.7+).

mrocklin:
> So, a thought just came to me.  I wonder if all of this could be solved by lru cache.

Expanding on this a bit more (that thought came to me last night and I wanted to get it out quickly).  I think that we actually prefer being able to use both sync and async handlers, but we're choosing not to do this for performance reasons, mostly because the `inspect.is_coroutine_function` function is surprisingly slow.  An alternative that might make it fast is to cache the result of this function.  This might be a one-line fix

```python
is_coroutine_function = functools.lru_cache(inspect.is_coroutine_function)
```

This does open up questions about how expensive `lru_cache` is, but I suspect that it's significantly faster.

ian-r-rose:
> > So, a thought just came to me.  I wonder if all of this could be solved by lru cache.
> 
> Expanding on this a bit more (that thought came to me last night and I wanted to get it out quickly). I think that we actually prefer being able to use both sync and async handlers, but we're choosing not to do this for performance reasons, mostly because the `inspect.is_coroutine_function` function is surprisingly slow. An alternative that might make it fast is to cache the result of this function. This might be a one-line fix

This is a good point, I'll test it out

ian-r-rose:
Okay, here is a look at some profiling results. I've mostly followed the same strategy detailed by @mrocklin in https://github.com/dask/distributed/issues/4443#issuecomment-765504479 using viztracer and shuffling the same large timeseries. The specific durations for the functions in question are small, but they are typically called thousands of times a second, so it does add up.

I've looked at three cases:
1. The current `main`
2. The async-only stream handler implemented here.
3. Using an lru cache for `is_coroutine_function`

Specifically, I'm looking at the profile of `Server.handle_stream` and `is_coroutine_function`. The latter only applies in cases (1) and (3).

#### `is_coroutine_function`:

Here is a histogram of durations for all sampled calls of `is_coroutine_function`:
image

As expected, using the LRU cache is significantly faster bringing the average duration from 15-20 ns down to 2-3 ns. This is a one-line solution, and keeps the API unchanged.

#### `handle_stream`:

Here is a histogram of duraitons for all sampled calls of `handle_stream`:

image

A few observations about this:
* The  `main` and LRU cases both have pretty wide distributions, depending on which synchronous op comes in. They average about 500 ns.
* It's hard to tell from the image alone, but the *average* duration of the LRU case is in fact shorter than that of the `main` case, by some tens of ns. It also looks peakier, maybe because of the more consistent duration of the cached `is_coroutine_function`
* The fully async case has a completely different distribution. In this case all handlers are pushed onto the async stack, so we aren't blocking on anything. It has a shorter average duration than the other two (~60 ns). Now, that time difference should be paid for later by the ioloop scheduling the previously-synchronous handlers at some other point.

Overall, it seems to me like the LRU version is indeed a quick win if we don't want to change any of the API, and continue to allow for synchronous handlers. Going fully async *may* be worthwhile, but involves a breaking API change, and maybe some pain for downstream plugin authors.

distributed/pubsub.py
ian-r-rose: I generally agree with you here @crusaderky. In this case I kept it this way because the thing I'm using is `typing.TYPE_CHECKING`, and I didn't really want that floating around the module namespace. No strong preference either way, though.

jakirkham:
Thanks for the detailed results Ian 😄

Do we know how many plugins are developed for Distributed?

jakirkham:
Oh sorry one more question, was the asyncio case using Tornado or uvloop? Wouldn't expect a difference, but would be good to know if there happened to be one (as we are using the latter in performance sensitive cases atm)

ian-r-rose:
> Do we know how many plugins are developed for Distributed?

Good question. I'm not aware of any third-party ones myself, but I don't see any reason why ones couldn't do something similar to [this one](https://github.com/dask/distributed/blob/98570fbcd662bc0fe468ccba9d57d582d2a3c678/distributed/stealing.py#L57).



> Oh sorry one more question, was the asyncio case using Tornado or uvloop? Wouldn't expect a difference, but would be good to know if there happened to be one (as we are using the latter in performance sensitive cases atm)

This was with `uvloop`, I haven't done extensive profiling with tornado at the moment

mrocklin:
Thank you for the detailed analysis @ian-r-rose 

It sounds like we're going with lru?

ian-r-rose:
> It sounds like we're going with lru?

I'm happy to go with that, it's a simpler change for the time being (though I generally like more predictable async vs sync APIs where possible).

mrocklin:
I'm happy either way with a mild preference for lru, just because it seems
less likely to have unintended consequences.

I'm also totally happy to be overruled here.  The work in this PR helps to
make things more predictable, which has real value and which we could use
more of.

jakirkham:
Went ahead and merged Ian's LRU cache PR ( https://github.com/dask/distributed/pull/4481 ). It's a pretty light PR. This coroutine check also only happens in a handful of places.

That said, I don't think it rules this async approach. There may be other reasons (as noted above) for going this route.

ian-r-rose:
I'll close this as #4474 is in, but will keep the branch around in case we want to revisit later

PR #4475, opened 2021-02-01
DOC: Autoreformat all functions docstrings.

Carreau:
Prompted by comment in #4473

This uses my prototype package `velin` to auto format all the functions
docstrings, from the root of the repo:

        $ velin distributed/ --write

This should fix some common issues/mistakes with numpydoc format often
around the presence/absence of space and/or colon, that can have
semantic implications.

Note that this easy to redo per-file / folder; so let me know if this is
too big


jrbourbeau (approved):
Thanks again @Carreau!

PR #4481, opened 2021-02-03
Cache is_coroutine_function

### LINKED ISSUE #4469: Avoid is_coroutine_function calls in Server

### LINKED ISSUE #4482: Duplicate iscoroutinefunction/is_coroutine_function function
Appears we define this same function with slightly different names twice. If we need the two names, maybe one can be an alias. Otherwise maybe we drop one

https://github.com/dask/distributed/blob/3407aa31be08ce100bd96e42acb99e1d7daf996e/distributed/utils.py#L1346-L1347

https://github.com/dask/distributed/blob/3407aa31be08ce100bd96e42acb99e1d7daf996e/distributed/utils.py#L1249-L1250

cc @ian-r-rose @mrocklin

### LINKED ISSUE #4483: Use iscoroutinefunction utility function in actor
Appears `actor` grabs `iscoroutinefunction` from `inspect` even though we have the utility function defined in `distributed.utils`. Maybe we should just use our utility function here as well

https://github.com/dask/distributed/blob/3407aa31be08ce100bd96e42acb99e1d7daf996e/distributed/actor.py#L3

ian-r-rose:
Fixes #4469 , fixes #4482, fixes #4483, alternative to #4474 
Uses cached version of `is_coroutine_function` in stream handling to mitigate re-checking the same function many times.

- [ ] Tests added / passed

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -1246,6 +1246,7 @@ def color_of(x, palette=palette):
     return palette[n % len(palette)]
 
 
+@functools.lru_cache(None)
 def iscoroutinefunction(f):
     return inspect.iscoroutinefunction(f) or gen.is_coroutine_function(f)
 
@@ -1343,8 +1344,7 @@ def parse_ports(port):
     return ports
 
 
-def is_coroutine_function(f):
-    return asyncio.iscoroutinefunction(f) or gen.is_coroutine_function(f)
+is_coroutine_function = iscoroutinefunction
 
 
 class Log(str):
--- a/distributed/actor.py
+++ b/distributed/actor.py
@@ -1,12 +1,11 @@
 import asyncio
 import functools
-from inspect import iscoroutinefunction
 import threading
 from queue import Queue
 
 from .client import Future, default_client
 from .protocol import to_serialize
-from .utils import thread_state, sync
+from .utils import iscoroutinefunction, thread_state, sync
 from .utils_comm import WrappedKey
 from .worker import get_worker
 


distributed/core.py
```diff
@@ -83,6 +83,11 @@ def _raise(*args, **kwargs):
 LOG_PDB = dask.config.get("distributed.admin.pdb-on-err")
 
 
+# Cached version of is_coroutine_function for server stream handlers,
+# allowing for quicker checks on on frequent code paths.
+_is_coroutine_function = lru_cache(1000)(is_coroutine_function)
```
jakirkham, line 88: Should we just decorate [the definition]( https://github.com/dask/distributed/blob/3407aa31be08ce100bd96e42acb99e1d7daf996e/distributed/utils.py#L1346 )?

Also how did we conclude 1,000 is the right size here?

jakirkham (commented):
Thanks Ian! 😄

This looks good generally. Had a couple questions on the details

distributed/core.py
ian-r-rose: > Should we just decorate [the definition](https://github.com/dask/distributed/blob/3407aa31be08ce100bd96e42acb99e1d7daf996e/distributed/utils.py#L1346)?

Sure, I'm happy to do that as well, I just usually try to default to minimal API changes (though this seems pretty harmless) :)

> Also how did we conclude 1,000 is the right size here?

No reason, really. Seemed unlikely to me that more than those would be checked, especially in this context. Could also be `None`.

distributed/utils.py
```diff
@@ -1184,7 +1184,7 @@ def get_event_loop(self):
         asyncio.set_event_loop_policy(AnyThreadEventLoopPolicy())
 
 
-@functools.lru_cache(1000)
+@functools.cache
```
jakirkham, line 1187:
```suggestion
@functools.lru_cache
```

distributed/utils.py
```diff
@@ -1246,6 +1246,7 @@ def color_of(x, palette=palette):
     return palette[n % len(palette)]
 
 
+@functools.cache
```
jakirkham, line 1249:
```suggestion
@functools.lru_cache
```

distributed/utils.py
```diff
@@ -1343,6 +1344,7 @@ def parse_ports(port):
     return ports
 
 
+@functools.lru_cache
 def is_coroutine_function(f):
     return asyncio.iscoroutinefunction(f) or gen.is_coroutine_function(f)
```
jakirkham, line 1349: Thoughts on just aliasing this?

```suggestion
is_coroutine_function = iscoroutinefunction
```

jakirkham (commented):
Thanks Ian! 😄

Looks like there is small naming error on CI. Also made one minor aliasing suggestion

distributed/utils.py
ian-r-rose: Thoughts on why the default 128 is better than no max size here?

distributed/utils.py
ian-r-rose: I'm in favor. Looking through the history, it seems there was once some py3.5 compatibillity-related difference between the two, and there is no longer.

distributed/utils.py
jakirkham: Oh I think when I made this suggestion there was no size. No strong feelings from me :)

ian-r-rose: To be clear -- I've chosen the case of no-max-size, since this is intended to be a pure function called with relatively few arguments, and sometimes on hot code paths.

jakirkham: Yep that makes sense. Have updated my suggestion to match that usage model.

Should add the suggestion is included as `cache` was only added in Python 3.9

ian-r-rose: > Should add the suggestion is included as `cache` was only added in Python 3.9

Yeah, I missed that earlier, thought I had learned about a fun old functool today, rather than a fun new one.

jakirkham (approved):

jakirkham:
Thanks Ian! 😄

ian-r-rose:
Thanks for the review @jakirkham!

PR #4487, opened 2021-02-06
`transition_processing_memory` optimizations, etc.

jakirkham:
- [ ] Closes #xxxx
- [ ] Tests added / passed

Goes through `transition_processing_memory` and optimizes it more carefully. In particular simplifies the code around starts tops, durations, and occupancies. Also types variables to get more performant Cython code. Makes use of intermediate variables to group work and allow easier reuse in later expressions. Tries to simplify data generated and collected to cutdown on general overhead associated. Additionally really cuts down on the work needed to produce messages for clients and workers.

Also includes various small optimizations like using `.get(...)`, typing variables, dropping unneeded intermediate variables, constructing values once, using typed attributes, etc.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6559,7 +6559,7 @@ def _remove_from_processing(state: SchedulerState, ts: TaskState) -> str:
     ts._processing_on = None
     w: str = ws._address
     if w in state._workers_dv:  # may have been removed
-        duration = ws._processing.pop(ts)
+        duration: double = ws._processing.pop(ts)
         if not ws._processing:
             state._total_occupancy -= ws._occupancy
             ws._occupancy = 0
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4623,8 +4623,9 @@ class Scheduler(SchedulerState, ServerNode):
             return
 
         if compute_duration:
-            old_duration = ts._prefix._duration_average
-            new_duration = compute_duration
+            old_duration: double = ts._prefix._duration_average
+            new_duration: double = compute_duration
+            avg_duration: double
             if old_duration < 0:
                 avg_duration = new_duration
             else:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -6781,13 +6781,15 @@ def _task_to_report_msg(state: SchedulerState, ts: TaskState) -> dict:
 @exceptval(check=False)
 def _task_to_client_msgs(state: SchedulerState, ts: TaskState) -> dict:
     report_msg: dict = _task_to_report_msg(state, ts)
+    client_msgs: dict
     if ts is None:
         # Notify all clients
-        return {k: [report_msg] for k in state._clients}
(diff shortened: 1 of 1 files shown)


jakirkham:

distributed/scheduler.py
```diff
@@ -2057,35 +2058,34 @@ def transition_processing_memory(
...
                 )
-                return {}, worker_msgs, client_msgs
+                return recommendations, worker_msgs, client_msgs
 
+            has_compute_startstop: bool = False
+            compute_start: double
+            compute_stop: double
             if startstops:
-                L = list()
+                startstop: dict
```
quasiben, line 2067: I thought we did more with `L` but I don't see it being used anywhere except for recording `compute_start, compute_stop`

distributed/scheduler.py
```diff
@@ -6778,22 +6776,14 @@ def _task_to_report_msg(state: SchedulerState, ts: TaskState) -> dict:
 @cfunc
 @exceptval(check=False)
 def _task_to_client_msgs(state: SchedulerState, ts: TaskState) -> dict:
-    cs: ClientState
-    client_keys: list
+    report_msg: dict = _task_to_report_msg(state, ts)
     if ts is None:
         # Notify all clients
-        client_keys = list(state._clients)
+        return {k: [report_msg] for k in state._clients}
```
quasiben, line 6782: Should this be first typed as dict rather than building on the fly as part of the return .  I am seeing that technique used for `recommendations` above

jakirkham: We can go either way. Did it on the fly here as it kept the code succinct. Open to changing it though

With `recommendations` am trying to line it up with how we handle other return arguments (like `client_msgs` and `worker_msgs`), which are also assigned up front. This also avoids accidental duplication (like assigning `recommendations` with `{}` in a few places). Thus allowing us to pay the cost of creating the `dict` only once and then using that `dict` or modifying it later

distributed/scheduler.py
jakirkham: Yeah exactly that's what I found as well, which is why I rewrote this to avoid creating an explicit `list`. Plus we only seem to use the first element of that `list`. So we just capture that in the `for`-loop below

jakirkham:
Seeing an unrelated test failure. Am able to reproduce without these changes. Filed as issue ( https://github.com/dask/distributed/issues/4493 )

distributed/scheduler.py
quasiben: Thanks for the explanation.   I don't have a strong opinion here and would defer to you.  I brought it up naively because of the techniques I saw earlier

jakirkham: Makes sense. Went ahead and assigned it to a variable for consistency

jakirkham:
Looks like CI has a different issue unrelated to this PR. Fixing with PR ( https://github.com/dask/distributed/pull/4499 )

jakirkham:

PR #4490, opened 2021-02-08
Attempt to get client from worker in Queue and Variable

jrbourbeau:
Today if a user attempts to create a `Variable` or `Queue` from within a task they will get the following error:

```python
ValueError: No clients found
Start a client and point it to the scheduler address
  client = Client('ip-addr-of-scheduler:8786')
```

This PR adds additional fallback logic to attempt to get a `Client` from the worker running the task, which allows users to do the following

```python
In [6]: x = Variable("my-var")

In [7]: def bar():
   ...:     from distributed import Variable
   ...:     v = Variable("my-var")
   ...:     return v.get()
   ...:

In [8]: x.set(987)

In [9]: client.submit(bar).result()
987
```

This behavior is in line with how `Lock`s, `Event`s, and `Semaphore`s work today:

https://github.com/dask/distributed/blob/7d2a22f94332974825e599dd779b4d98f44dbca2/distributed/lock.py#L96-L101

### DIFF
--- a/distributed/queues.py
+++ b/distributed/queues.py
@@ -7,7 +7,7 @@ from dask.utils import stringify
 
 from .client import Future, Client
 from .utils import sync, thread_state
-from .worker import get_client
+from .worker import get_client, get_worker
 from .utils import parse_timedelta
 
 logger = logging.getLogger(__name__)
@@ -150,8 +150,8 @@ class Queue:
         Name used by other clients and the scheduler to identify the queue. If
         not given, a random name will be generated.
     client: Client (optional)
-        Client used for communication with the scheduler. Defaults to the
-        value of ``Client.current()``.
+        Client used for communication with the scheduler.
+        If not given, the default global client will be used.
     maxsize: int (optional)
         Number of items allowed in the queue. If 0 (the default), the queue
         size is unbounded.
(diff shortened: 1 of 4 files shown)


mrocklin:
Two quick comments:

1.  We should have a test for this
2. @fjetter noticed that we're less scalable around having many clients as we are around having many workers.  We don't need to fix this right now, but in the future we may want to either make clients more scalable, or else make it easier to create variables/queues from workers without a client present.

jrbourbeau:
Thanks for reviewing -- test added

> @fjetter noticed that we're less scalable around having many clients as we are around having many workers

Good to know. Is there a discussion around this somewhere? Or more just a theme @fjetter has noticed when running at scale ?

fjetter:
There is no big discussion on the subject since I changed the implementation of the semaphore to not use clients anymore, see https://github.com/dask/distributed/pull/4195

The scenario back then was that we ended up with a 100% CPU utilization on scheduler side once we reached about 200 workers w/ an active semaphore. I'm not entirely certain if it is solely due to the clients themselves or if it were too many connections, maybe both.

One low hanging fruit which should help for this is already to adjust the heartbeat intervals of the clients. They do not scale dynamically with increasing number of connected clients and this might already be it but I decided to refactor the semaphore instead of triaging the client scalability.

fjetter (approved):

jrbourbeau:
Thanks for reviewing @fjetter @mrocklin!

Also, xref https://github.com/dask/distributed/issues/4498 for the failing CI builds

PR #4495, opened 2021-02-09
Change default branch from master to main

jsignell:
I'm realizing that doing distributed and dask at the same time is unfeasible since it'll mean that the tests never pass.  I am going to split this up into just the distributed part.

### DIFF
--- a/docs/release-procedure.md
+++ b/docs/release-procedure.md
@@ -1,3 +1,3 @@
 Distributed follows a similar procedure for releasing as the core Dask project.
 
-See https://github.com/dask/dask/blob/main/docs/release-procedure.md for instructions.
+See https://github.com/dask/dask/blob/master/docs/release-procedure.md for instructions.
--- a/docs/release-procedure.md
+++ b/docs/release-procedure.md
@@ -1,3 +1,3 @@
 Distributed follows a similar procedure for releasing as the core Dask project.
 
-See https://github.com/dask/dask/blob/master/docs/release-procedure.md for instructions.
+See https://github.com/dask/dask/blob/main/docs/release-procedure.md for instructions.
--- a/.github/workflows/ci-pre-commit.yml
+++ b/.github/workflows/ci-pre-commit.yml
@@ -2,9 +2,9 @@ name: Linting
 
 on:
   push:
-    branches: master
+    branches: main
   pull_request:
-    branches: master
+    branches: main
 
 jobs:
   checks:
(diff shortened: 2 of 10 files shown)


jakirkham:
Just a note, we might want to add `main` to `pip install` commands that do development installs that are not otherwise specifying a branch. We needed to do this for `zict` recently as noted in PR ( https://github.com/dask/distributed/pull/4499 )

jsignell:
Thanks @jakirkham. I just tried pip installing from git without specifying the branch and it seemed to work fine. 

```bash
conda create -n test pip python=3.7
conda activate test
pip install git+https://github.com/dask/zict.git
```

jakirkham:
Interesting. Well here is the CI failure we saw

https://github.com/dask/distributed/runs/1868031469

jakirkham:
Maybe a non-zero exit code is being set (even though it happens to work)?

jsignell:
Definitely the windows part of #4499 replacing master with main needed to happen. I am just wondering if the travis part did.

jakirkham:
Right was wondering if that happened when running locally

jsignell:
I just rebased and I expect this to pass.

docs/release-procedure.md
```diff
@@ -1,3 +1,3 @@
 Distributed follows a similar procedure for releasing as the core Dask project.
 
-See https://github.com/dask/dask/blob/master/docs/release-procedure.md for instructions.
\ No newline at end of file
+See https://github.com/dask/dask/blob/master/docs/release-procedure.md for instructions.
```
jrbourbeau, line 3:
```suggestion
See https://github.com/dask/dask/blob/main/docs/release-procedure.md for instructions.
```

jrbourbeau (commented):
Thanks @jsignell. Given the recent release, and no reported issues with the branch name change in other `dask/*` repos, I think we're in a good place to merge this (and the corresponding PR in `dask/dask`) soon. I'll comment over in the community issue just to give people a heads up

docs/release-procedure.md
jsignell: I am purposefully only doing distributed in this PR. I wanted to be able to merge this one first, wait a bit and then do dask/dask.

jrbourbeau: Gotcha, thanks for clarifying

jsignell:
I think the failure is unrelated, so this should be good to merge.

jrbourbeau (approved):
Thanks @jsignell!

jakirkham:
What's the ETA on fixing the Dask branch as well? We need to update some CI environments and would ideally like to make this change once

PR #4218, opened 2020-11-04
Migrate to GitHub Actions and change install script to environment files

jacobtomlinson:
Closes #4214 

This PR migrates CI to GitHub Actions.

Changes:
- All python versions are now tested on macos
- Environments are now built from environment files instead of many conda install commands
- Coverage is run on Python 3.8 Ubuntu only

PR #4504, opened 2021-02-12
Migrate from travis to github actions

### LINKED ISSUE #4214: Migrate CI to GitHub Actions
Due to changes in the Travis CI billing, the Dask org is migrating CI to GitHub Actions.

This repo contains a `.travis.yml` file which needs to be replaced with an equivalent `.github/workflows/ci.yml` file.

See dask/community#107 for more details.

### LINKED ISSUE #4387: Segmentation fault on travis
There appears to be a segfault on CI. Haven't debugged further. 

See for example https://travis-ci.org/github/dask/distributed/jobs/750955092

crusaderky:
- Closes #4214 
- Closes #4387
- Supersedes #4218 
- Follow-ups: #4509, #4512, #4514, #4519, #4543

Changes:
- Moved Linux and MacOS CI from travis to github actions
- Now running MacOS CI on Python 3.8 in addition to 3.7
- Run coverage everywhere
- Added new github workflow to debug ssh-related issues (disabled by default)
- Removed python-snappy and python-blosc from 3.7 (they're already tested on 3.8)
- Moved lz4 from 3.6 to 3.8
- Moved crick from 3.6 to 3.7
- Use git tip of s3fs, zict, filesystem_spec, and joblib on the latest version of Python only
- "slow" tests are now running on Windows too
- Added verbose flag to pytest (vital in figuring out pytest-timeout failures)
- Relaxed a wealth of timeouts to prevent random failures
- Marked several of tests as flaky or xfail to prevent random failures
- Introduced pytest_rerunfailures
- Disabled ipv6 testing on Linux - see #4514. I could not figure out the issue.

Stress tests evidence:
https://docs.google.com/spreadsheets/d/1BN_85gjT9AoGJUGPvhIKSiEcIzfS3VV6O_1zSAtBhBM/edit?usp=sharing


distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -209,11 +209,9 @@ async def test_restart(cleanup):
                 cluster.scale(2)
                 await cluster
                 assert len(cluster.workers) == 2
-
                 await client.restart()
-                await asyncio.sleep(3)
-
-                assert len(cluster.workers) == 2
+                while len(cluster.workers) != 2:
+                    await asyncio.sleep(0.5)
```
mrocklin, line 215: Thanks for doing this.  Explicit sleep calls are the bane of CI. 

Question, is there anything here to cause this test to time out if it fails?  If not, this is the sort of thing that can cause CI to hang without feedback.  pytest-asyncio might do this for us.  I'm not sure.  if not then we might want to verify that `time() < start + 10` or something.

distributed/deploy/tests/test_ssh.py
```diff
@@ -19,6 +19,7 @@ def test_ssh_hosts_empty_list():
         SSHCluster(hosts=[])
 
 
+@pytest.mark.ssh
```
mrocklin, line 22: I wonder if we can mark the entire file with `ssh` somehow?

distributed/tests/test_scheduler.py
```diff
@@ -2168,7 +2169,7 @@ async def test_unknown_task_duration_config(s, a, b):
     assert s.idle_since == s.time_started
 
 
-@gen_cluster(client=True, timeout=1000)
+@gen_cluster(client=True, timeout=None)
```
mrocklin, line 2172: This seems off.  `timeout=None` can cause CI to hang without giving us much feedback.

distributed/utils_test.py
```diff
@@ -886,10 +886,11 @@ async def coro():
                                 )
                             except Exception as e:
                                 logger.error(
-                                    "Failed to start gen_cluster, retrying",
+                                    "Failed to start gen_cluster: "
+                                    f"{e.__class__.__name__}: {e}, retrying",
                                     exc_info=True,
                                 )
-                                await asyncio.sleep(1)
+                                await asyncio.sleep(5)
```
mrocklin, line 893: Personally I would rather have more short sleeps than long ones.  This change may have been necessary for some reason though.

distributed/deploy/tests/test_spec_cluster.py
mrocklin: Ah, I see in setup.cfg that you've added a timeout option.  Great.

crusaderky: pytest-timeout will kill off the whole test suite (not just the single test) after 5 minutes

distributed/deploy/tests/test_ssh.py
crusaderky: I wondered that too but found nothing on the pytest docs

distributed/tests/test_scheduler.py
crusaderky: pytest-timeout will kill everything after 300 seconds. You'll never reach the 1000 seconds listed there.

distributed/utils_test.py
crusaderky: I was systematically getting failures here. I'll try dropping down the sleep again and increasing the number of retries to compensate.

mrocklin:
Looking at the recent failed test run in https://github.com/dask/distributed/runs/1919308651 , two things stand out

There are several IO Loop and Profile threads still active.  There are about 10 of these at the end of the report.  So something is leaking here.

```python-traceback
~~~~~~~~~~~~~~~~~~~~~~ Stack of IO loop (139674081294080) ~~~~~~~~~~~~~~~~~~~~~~
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/threading.py", line 890, in _bootstrap
    self._bootstrap_inner()
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/runner/work/distributed/distributed/distributed/utils.py", line 417, in run_loop
    loop.start()
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/site-packages/tornado/platform/asyncio.py", line 199, in start
    self.asyncio_loop.run_forever()
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/asyncio/base_events.py", line 541, in run_forever
    self._run_once()
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/asyncio/base_events.py", line 1750, in _run_once
    event_list = self._selector.select(timeout)
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/selectors.py", line 468, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
```

There is a small error at the beginning with `Client._asynchronous` not being set before `__del__` is called

```python-traceback
Exception ignored in: <function Client.__del__ at 0x7f089a528f80>
  File "/home/runner/work/distributed/distributed/distributed/client.py", line 1197, in __del__
    self.close()
  File "/home/runner/work/distributed/distributed/distributed/client.py", line 1421, in close
    if self.asynchronous:
  File "/home/runner/work/distributed/distributed/distributed/client.py", line 806, in asynchronous
    return self._asynchronous and self.loop is IOLoop.current()
AttributeError: 'Client' object has no attribute '_asynchronous'
```

This is probably minor and doesn't impact the main problem here, but can probably also be fixed easily by moving the `Client._asynchronous = asynchronous` line before we possibly break out of the constructor, such as with a raised exception.

mrocklin:
Feel free to ignore my comments though.  I'm just glancing at this in hopes that I stumble upon something useful.  I estimate the probability of that happening at less than 20%

distributed/deploy/tests/test_spec_cluster.py
crusaderky: @mrocklin I didn't add it; I just moved it from the command-line arguments to setup.cfg

crusaderky: added individual timeout to the test

distributed/utils_test.py
```diff
@@ -886,7 +886,8 @@ async def coro():
                                 )
                             except Exception as e:
                                 logger.error(
-                                    "Failed to start gen_cluster, retrying",
+                                    "Failed to start gen_cluster: "
+                                    f"{e.__class__.__name__}: {e}; retrying",
```
crusaderky, line 890: exc_info=True doesn't print out anything for some reason

distributed/deploy/tests/test_ssh.py
jrbourbeau: It looks like there's a special `pytestmark` variable we can set at the module-level https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules

.github/workflows/ipv6_debug.yaml
```diff
@@ -0,0 +1,41 @@
...
+        shell: bash -l {0}
+        run: python -m pip install --no-deps -e .
+
+      - name: Test
+        shell: bash -l {0}
+        run: pytest distributed/cli/tests/test_dask_scheduler.py::test_defaults
+
+      - name: Debug with tmate on failure
+        if: ${{ failure() }}
+        uses: mxschmitt/action-tmate@v3
```
crusaderky, line 41: Unsure if this file should go to master of if I should remove it here and paste it into #4514

.github/workflows/ssh_debug.yaml
```diff
@@ -0,0 +1,47 @@
...
+        shell: bash -l {0}
+        run: ssh -vvv 127.0.0.1 'echo hello world'
+
+      - name: Test SSH vs. hostname
+        shell: bash -l {0}
+        run: ssh -vvv $(hostname) 'echo hello world'
+
+      - name: Debug with tmate on failure
+        if: ${{ failure() }}
+        uses: mxschmitt/action-tmate@v3
```
crusaderky, line 47: I think this should get into master; it will save *a lot* of work next time SSH breaks.

distributed/deploy/tests/test_ssh.py
crusaderky: fixed

crusaderky:
> This is probably minor and doesn't impact the main problem here, but can probably also be fixed easily by moving the Client._asynchronous = asynchronous line before we possibly break out of the constructor, such as with a raised exception.

@mrocklin I don't think it's that simple. We should revise all ``__del__`` methods to cope with the use case of an arbitrary attribute not being there anymore, which to my understanding it is something that can and will happen when the method is invoked by the interpreter shutdown process.

distributed/cli/tests/test_dask_scheduler.py
```diff
@@ -187,9 +187,7 @@ def test_interface(loop):
                 assert all("127.0.0.1" == d["host"] for d in info["workers"].values())
 
 
-@pytest.mark.skipif(
-    sys.version_info < (3, 7), reason="Intermittent failure on old Python version"
-)
+@pytest.mark.flaky(reruns=10, reruns_delay=5)
```
crusaderky, line 190: Found frequent failures on 3.7 and 3.8

distributed/comm/tcp.py
```diff
@@ -260,6 +259,8 @@ async def write(self, msg, serializers=None, on_error="message"):
             for each_frame in frames:
                 each_frame_nbytes = nbytes(each_frame)
                 if each_frame_nbytes:
+                    if stream._write_buffer is None:
+                        raise StreamClosedError()
```
crusaderky, line 263: Treat ``'NoneType' object has no attribute 'append'``; see handler below:
- ``if shutting_down()``, silently suppress the error
- otherwise, re-raise it as CommClosedError

distributed/tests/test_asyncprocess.py
```diff
@@ -48,7 +49,7 @@ def threads_info(q):
     q.put(threading.current_thread().name)
 
 
-@pytest.mark.xfail(reason="Intermittent failure")
+@pytest.mark.xfail()
```
crusaderky, line 52: deterministic failure

crusaderky:
@jrbourbeau @mrocklin this is now ready for review.

Failure rate on Windows/Linux is < 5% per job; on MacOS it's 20~30%.
MacOS is currently disable on PRs and only running on push.
See opening post for full summary of changes.

Stress tests evidence:
https://docs.google.com/spreadsheets/d/1BN_85gjT9AoGJUGPvhIKSiEcIzfS3VV6O_1zSAtBhBM/edit?usp=sharing

jacobtomlinson (approved):
This looks great! Thanks so much for picking this up.

jrbourbeau (approved):
Thanks for all your work on this @crusaderky! In particular, thanks for opening up several issues to inform follow up work to further improve our CI. This is in

jakirkham:
Seeing this get merged made my day! Thanks for all of the hard work here! 😄

PR #4506, opened 2021-02-12
Read & write all frames in one pass

jakirkham:
- [ ] Closes #xxxx
- [x] Tests added / passed

Instead of doing multiple reads with `async`, just allocate one big chunk of memory for all of the frames and read into it. Should cutdown on the number of passes through Tornado needed to fill these frames.

Also knowing that `IOStream` has an internal queue of buffers for writing, we are able to push all of the frames into that queue beforehand. Then ask Tornado to `write` after they are in the queue. This also cuts down on the number of passes through Tornado by simply entering the write handling code once and writing all the buffers.

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -189,14 +189,14 @@ class TCP(Comm):
             raise CommClosedError
 
         try:
-            all_nbytes = await stream.read_bytes(8)
-            (all_nbytes,) = struct.unpack("Q", all_nbytes)
+            frames_nbytes = await stream.read_bytes(8)
+            (frames_nbytes,) = struct.unpack("Q", frames_nbytes)
 
-            all_frames = bytearray(all_nbytes)
-            n = await stream.read_into(all_frames)
-            assert n == all_nbytes, (n, all_nbytes)
+            frames = bytearray(frames_nbytes)
+            n = await stream.read_into(frames)
+            assert n == frames_nbytes, (n, frames_nbytes)
 
-            frames = unpack_frames(all_frames)
+            frames = unpack_frames(frames)
         except StreamClosedError as e:
             self.stream = None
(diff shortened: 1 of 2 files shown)


mrocklin:
Hrm, neat. :)

jakirkham:
Yeah we might be able to do better with `sendmsg` and `recvmsg`. Am still wrapping my head around how we can use those

mrocklin:
My _guess_ is that going beyond this will have diminishing returns, at least on the scheduler side where we generally have small messages.  I could easily be wrong though.

jakirkham:
Made a few more changes. I think this captures the same idea as what `sendmsg`/`recvmsg` would give us. By doing this we are able to go from 2 `read_bytes` calls down to 1 `read_bytes` calls. Should add this is in addition to having one big buffer receiving all frames.

jakirkham:
Simplified this a bit more using `pack_frames_prelude` to build the header for us. Then we just tack on the message size before that so the receiving end gets that first.

On the receiving end, we just get the message size first. Then use that to preallocate a buffer to hand off to Tornado to fill. The remaining unpacking is just handled by `unpack_frames`. Deserializing steps can then be handled after that per usual.

jakirkham:
Also figured out how to offload all frames to Tornado. So it now only uses one `write` call there

jakirkham:
Some profiling details in issue ( https://github.com/quasiben/dask-scheduler-performance/issues/108 ). Most notably this is giving us a `7%` improvement in transfer time ( https://github.com/quasiben/dask-scheduler-performance/issues/108#issuecomment-780120023 )

jakirkham:

jakirkham:
Thanks Matt! 😄

mrocklin:
All I did was press the green button :)

PR #4507, opened 2021-02-13
Skip `stream.write` call for empty frames

jakirkham:
Sometimes a frame winds up have `0` bytes. In these cases, make sure to skip calling `stream.write` to avoid incurring the overhead. After all it is trivial to create an empty frame on the receiving end as long as we know it exists (the size of all frames are tracked) without involving IO.

- [ ] Closes #xxxx
- [ ] Tests added / passed


jakirkham:
Should add we already do this on the `read` side and make sure to create the empty buffers as well. So this is just extending that logic to `write` as well

https://github.com/dask/distributed/blob/ff56bd1ebf7ab1da30b3ec314651f78b25118dd8/distributed/comm/tcp.py#L198-L199

PR #4511, opened 2021-02-15
Fix settings file not work bug

daimon99:
Fix preload settings cannot be load bug when using dask-ssh start the worker.
The preload parameter in the worker init function will be an empty tuple when using dask-ssh start the worker, then it will not read the yml settings files.
But in the schedule.py file, it is ok for "if not preload". 
So I change the worker code to "if not preload"

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -525,7 +525,7 @@ class Worker(ServerNode):
             self._workdir = self._workspace.new_work_dir(prefix="worker-")
             self.local_directory = self._workdir.dir_path
 
-        if preload is None:
+        if not preload:
             preload = dask.config.get("distributed.worker.preload")
         if preload_argv is None:
             preload_argv = dask.config.get("distributed.worker.preload-argv")
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -527,7 +527,7 @@ class Worker(ServerNode):
 
         if not preload:
             preload = dask.config.get("distributed.worker.preload")
-        if preload_argv is None:
+        if not preload_argv:
             preload_argv = dask.config.get("distributed.worker.preload-argv")
         self.preloads = preloading.process_preloads(
             self, preload, preload_argv, file_dir=self.local_directory
--- a/distributed/cli/tests/test_dask_scheduler.py
+++ b/distributed/cli/tests/test_dask_scheduler.py
@@ -314,6 +314,19 @@ def test_preload_remote_module(loop, tmp_path):
                 )
 
 
+def test_preload_config(loop):
+    # Ensure dask-scheduler pulls the preload from the Dask config if
+    # not specified via a command line option
+    with tmpfile() as fn:
+        env = {"DASK_DISTRIBUTED__SCHEDULER__PRELOAD": PRELOAD_TEXT}
+        with popen(["dask-scheduler", "--scheduler-file", fn], env=env):
+            with Client(scheduler_file=fn, loop=loop) as c:
+                assert (
+                    c.run_on_scheduler(lambda dask_scheduler: dask_scheduler.foo)
+                    == "bar"
+                )
+
+
 PRELOAD_COMMAND_TEXT = """
 import click
 _config = {}
--- a/distributed/cli/tests/test_dask_worker.py
+++ b/distributed/cli/tests/test_dask_worker.py
@@ -453,3 +453,26 @@ class MyWorker(Worker):
 
                 worker_types = await c.run(worker_type)
                 assert all(name == "MyWorker" for name in worker_types.values())
+
+
+@pytest.mark.asyncio
+async def test_preload_config(cleanup):
+    # Ensure dask-worker pulls the preload from the Dask config if
+    # not specified via a command line option
+    preload_text = """
+def dask_setup(worker):
+    worker.foo = 'setup'
+"""
+    async with Scheduler(port=0) as s:
+        async with Client(s.address, asynchronous=True) as c:
+            env = {"DASK_DISTRIBUTED__WORKER__PRELOAD": preload_text}
+            with popen(
+                [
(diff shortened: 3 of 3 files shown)


jrbourbeau (commented):
Thanks for the PR @daimon99! Could you add a test for this?

> The preload parameter in the worker init function will be an empty tuple when using dask-ssh start the worker

Looking briefly at the `dask-ssh` source, I wasn't able to see why `preload=()` would be used when creating workers

daimon99:
I just have a try, The problem is here, because setting multiple=True in the preload option in the dask_worker.py line 223.
image

daimon99:
I had tried to write a test case for my pr, but it is difficult for me to do it... The logic is clear, but the test case is hard to write. Sorry.

distributed/worker.py
```diff
@@ -554,7 +554,7 @@ def __init__(
             self._workdir = self._workspace.new_work_dir(prefix="worker-")
             self.local_directory = self._workdir.dir_path
 
-        if preload is None:
+        if not preload:
             preload = dask.config.get("distributed.worker.preload")
         if preload_argv is None:
```
jrbourbeau, line 559: It looks like `preload_argv` will also be an empty tuple if no value is used for the `--preload-argv` CLI option for `dask-worker`

```suggestion
        if not preload_argv:
```

jrbourbeau (commented):
Ah, I see. The underlying issue is `dask-worker` being called via `dask-ssh`. Thanks for the screenshot @daimon99, that helped clarify things for me. 

I agree switching to `not preload` instead of `preload is None` is a reasonable fix for this. It looks like we're already doing that for the scheduler:

https://github.com/dask/distributed/blob/02a85f7891b5872aa8e7a0fc62c2891f503a7963/distributed/scheduler.py#L2900-L2903

Here's a test that can be added to `distributed/cli/tests/test_dask_worker.py` which should ensure `dask-worker` picks up preloads specified as configuration values when no `--preload` CLI option is used:

```diff
diff --git a/distributed/cli/tests/test_dask_worker.py b/distributed/cli/tests/test_dask_worker.py
index 0055c38e..c4c2cdb4 100644
--- a/distributed/cli/tests/test_dask_worker.py
+++ b/distributed/cli/tests/test_dask_worker.py
@@ -10,6 +10,7 @@ import os
 from time import sleep
 from multiprocessing import cpu_count

+import dask
 import distributed.cli.dask_worker
 from distributed import Client, Scheduler
 from distributed.deploy.utils import nprocesses_nthreads
@@ -460,3 +461,20 @@ class MyWorker(Worker):

                 worker_types = await c.run(worker_type)
                 assert all(name == "MyWorker" for name in worker_types.values())
+
+
+@pytest.mark.asyncio
+async def test_preload_config(cleanup):
+    # Ensure dask-worker pulls the preload from the Dask config if
+    # not specified via a command line option
+    preload_text = """
+def dask_setup(worker):
+    worker.foo = 'setup'
+"""
+    async with Scheduler(port=0) as s:
+        async with Client(s.address, asynchronous=True) as c:
+            env = {"DASK_DISTRIBUTED__WORKER__PRELOAD": preload_text}
```

and similarly for `dask-scheduler` (which should already have the the correct behavior, but adding a test is still a good thing to do):

```diff
diff --git a/distributed/cli/tests/test_dask_scheduler.py b/distributed/cli/tests/test_dask_scheduler.py
index 301f46fd..175f55dc 100644
--- a/distributed/cli/tests/test_dask_scheduler.py
+++ b/distributed/cli/tests/test_dask_scheduler.py
@@ -335,6 +335,19 @@ def test_preload_remote_module(loop, tmp_path):
                 )


+def test_preload_config(loop):
+    # Ensure dask-scheduler pulls the preload from the Dask config if
+    # not specified via a command line option
+    with tmpfile() as fn:
+        env = {"DASK_DISTRIBUTED__SCHEDULER__PRELOAD": PRELOAD_TEXT}
+        with popen(["dask-scheduler", "--scheduler-file", fn], env=env):
+            with Client(scheduler_file=fn, loop=loop) as c:
+                assert (
+                    c.run_on_scheduler(lambda dask_scheduler: dask_scheduler.foo)
+                    == "bar"
+                )
+
+
 PRELOAD_COMMAND_TEXT = """
 import click
 _config = {}
```

distributed/worker.py
daimon99: Study a lot from your reply :)
I just know it can insert code snippet  here.
Thank you~~

jrbourbeau (commented):
@daimon99 we recently had a large overhaul of our CI. Could you try merging the `master` branch into your PR to include those updates here?

martindurant:
(we would happily do the merge for you, but you have PR-ed from your master branch, and we normally do not/cannot push there)

GPUtester:

daimon99:
Sorry for the delay~
I merge the latest main commit to the pr now. And pr again.

jrbourbeau:
ok to test

crusaderky:
This is weird.
I reproduced the Windows issue locally.
The scheduler crashes without ever reaching the Scheduler.start() method. As a consequence, the scheduler-file is never written and the Client is stuck waiting forever.

```python
< with popen(["dask-scheduler", "--scheduler-file", fn], env=env):
> from subprocess import PIPE
> with popen(["dask-scheduler", "--scheduler-file", fn], env=env, stderr=PIPE) as proc:
>     sleep(5)
>     print(proc.stderr.read())
>     assert False
```
Output:
b'Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python\nPython runtime state: preinitialized\n\n'

The crash disappears if I remove ``env=env``.

jrbourbeau:
Closing in favor of https://github.com/dask/distributed/pull/5365

PR #4526, opened 2021-02-19
Batch more Scheduler sends

jakirkham:
Factors out `_transitions` from `transitions` to allow collecting of messages to send for the caller to handle as opposed to sending immediately. This then leveraged in a few stimulus and handle functions to collect all messages and send in one batch. Also this is leveraged in `add_worker` and `gather`. All of which makes a few calls to `transition` or `transitions`. The result is in these cases we should send fewer messages.

- [ ] Closes #xxxx
- [x] Tests added / passed


jakirkham:
Planning on merging end of day Tuesday if no comments

Issue #4522, opened 2021-02-19
Accessing `ipython` on workers through `client.start_ipython_workers()`


SultanOrazbayev:
I am trying to see if I can run ipython on workers interactively. The first example in the documentation (https://distributed.dask.org/en/latest/ipython.html) works, while the second one is not working for me. Here's the minimal code to reproduce:

```python
from dask.distributed import Client
client = Client()

info = client.start_ipython_workers(magic_names="w*")
%w0 1+1
# this gives TimeoutError: Timeout waiting for IPython output
```

This is more of a curiosity for me, just trying to explore what is possible. I posted it also as a question on StackOverflow: https://stackoverflow.com/questions/66167619/how-to-use-client-start-ipython-workers-in-dask-distributed

**Environment**:

- Dask version: 2021.02.0
- Python version: Python 3.7.9
- Operating System: MacOS Catalina
- Install method (conda, pip, source): conda

mrocklin:
This used to work and be wonderful. Then I think when Tornado updated we
ended up xfailing the tests for this because it was hard to sustain.  I'm
not sure what the status is now.

@minrk do you happen to have anyone around who would be interested in
getting this back up to speed?

SultanOrazbayev:
Just to elaborate on one potential use-case. I wanted to run a given set of commands against multiple datasets, expecting that the commands will fail at different stages due to missing values. Some of the calculations are time-intensive and I would like to have the ability to get the state of calculations at the time of failure so I can iterate interactively. However, this was only for a very specific case, so probably not of broad interest.

minrk:
#4530 should do it. It had to do with how we created the event-loop for the IPython thread, which involved multiple deprecated things (ZMQIOLoop - deprecated in pyzmq 17, IOLoop.instance - deprecated in tornado 5). I'm not 100% sure why the deprecated way isn't working. I think it should, and this is perhaps a bug in pyzmq's implementation of these deprecations, but since it's long-since deprecated and removing it gets it working again, I don't plan to dig into it further. The bug boiled down to this: no response to events is ultimately due to hooking up a listener to one eventloop, but starting a different one. Like this:

```python
loop = asyncio.new_event_loop()
loop.add_listener(fd, callback)

other_loop = asyncio.new_event_loop() # <-- not the eventloop with a listener on fd
other_loop.run_forever() # never handles events on fd!
```

> ended up xfailing the tests for this because it was hard to sustain

[...]

PR #4530, opened 2021-02-22
remove deprecated ioloop workarounds

minrk:
- There are no longer process-global loops, use standard IOLoop.current
- ZMQIOLoop is no longer needed, pyzmq >= 17 uses the standard tornado IOLoop

- [ ] Closes #xxxx
- [ ] Tests added / passed


distributed/_ipython_utils.py
```diff
@@ -234,20 +226,17 @@ def _start():
...
         app.kernel.start()
-        app.kernel.loop = IOLoop.instance()
         # save self in the IPython namespace as 'worker'
         # inject things into the IPython namespace
         if ns:
             app.kernel.shell.user_ns.update(ns)
         evt.set()
-        zmq_loop.start()
+        # start the app's IOLoop in its thread
+        IOLoop.current().start()
```
minrk, line 235: Because IPython no longer needs to run the zmq ioloop implementation, this thread can be done away with in most cases, and IPython could just share the main IOLoop. However! When event-loop integration is enabled in ipykernel (e.g. `%matplotlib qt`), the main event-loop becomes subordinate to the gui toolkit event-loop, and may not wake correctly for all dask events. So it may be safer to keep IPython in a thread anyway, despite not being *strictly* necessary.

distributed/tests/test_ipython.py
```diff
@@ -15,11 +14,7 @@ def need_functional_ipython(func):
         import jupyter_client  # noqa: F401
     except ImportError:
         return pytest.mark.skip("need ipykernel and jupyter_client installed")(func)
-    if tornado.version_info >= (5,):
```
minrk, line 18: tornado 5 is required by both ipykernel and distributed, so this condition is always True.

The upstream bug has been fixed for some time.

martindurant:
The failure on travis does look to me like it might be caused by this - the worker repeatedly crashed and restarts

minrk:
Rebased. It doesn't entirely make sense to me that the failing many-worker test was related to this PR, since it tested code not touched by these changes. It is conceivable that there were some leftover FDs contaminating from one test to another, though.


Hopefully the rebase will solve it.

jrbourbeau (approved):
Thank you for taking the time to fix this @minrk! I ran the original code snippet posted in https://github.com/dask/distributed/issues/4522 and everything works as expected with the code changes here

distributed/tests/test_ipython.py
jakirkham: Yeah we also [require Tornado 5+ as well (if not newer)]( https://github.com/dask/distributed/blob/4ce8c65039e9b531878f9d77cf884920690a3745/requirements.txt#L9-L10 ). So this was probably just missed when we bumped the requirement

distributed/_ipython_utils.py
jakirkham: Isn't there something special about Qt needing to run in the main thread? It's been a while since I thought about this so could be misremembering or that could be outdated

Issue #4542, opened 2021-02-23
CancelledError with collections annotations


TomAugspurger:
Some background: I'm using `Array.map_blocks` in combination with the new `dask.annotations` to set resource restrictions, followed by a getitem, I get a KeyError / CancelledError.

```python
import dask
import dask.array
from distributed import Client, LocalCluster

cluster = LocalCluster(resources={"gpu": 1})
client = Client(cluster)
client

def f(x):
    return x + 1

a = dask.array.ones((4, 100, 100), chunks=(4, 10, 10))


with dask.annotate(resources={"gpu": 1}):
    b = a.map_blocks(f, meta=a._meta)

b[0, 0, 0].compute(optimize_graph=False)
```

```pytb
## -- End pasted text --
distributed.core - ERROR - "('f-3406b3798f8ce70c11bdf0b38247241f', 0, 0, 1)"
Traceback (most recent call last):
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/core.py", line 572, in handle_stream
    handler(**merge(extra, msg))
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/scheduler.py", line 3658, in update_graph_hlg
    return self.update_graph(
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/scheduler.py", line 3824, in update_graph
    ts = parent._tasks[k]
KeyError: "('f-3406b3798f8ce70c11bdf0b38247241f', 0, 0, 1)"
distributed.core - ERROR - Exception while handling op register-client
Traceback (most recent call last):
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/core.py", line 500, in handle_comm
    result = await result
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/scheduler.py", line 4519, in add_client
    await self.handle_stream(comm=comm, extra={"client": client})
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/core.py", line 572, in handle_stream
    handler(**merge(extra, msg))
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/scheduler.py", line 3658, in update_graph_hlg
    return self.update_graph(
  File "/mnt/c/Users/taugspurger/src/dask/distributed/distributed/scheduler.py", line 3824, in update_graph
    ts = parent._tasks[k]
KeyError: "('f-3406b3798f8ce70c11bdf0b38247241f', 0, 0, 1)"
[...]

PR #4544, opened 2021-02-24
Handle annotations for culled tasks

TomAugspurger:
Closes https://github.com/dask/distributed/issues/4542

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3668,7 +3668,6 @@ class Scheduler(SchedulerState, ServerNode):
         fifo_timeout=0,
         annotations=None,
     ):
-
         dsk, dependencies, annotations = highlevelgraph_unpack(hlg, annotations)
 
         # Remove any self-dependencies (happens on test_publish_bag() and others)
@@ -3854,8 +3853,11 @@ class Scheduler(SchedulerState, ServerNode):
 
             for a, kv in annotations.items():
                 for k, v in kv.items():
-                    ts = parent._tasks[k]
-                    ts._annotations[a] = v
+                    # Tasks might have been culled, in which case
+                    # we have nothing to annotate.
+                    if k in parent._tasks:
+                        ts = parent._tasks[k]
+                        ts._annotations[a] = v
 
         # Add actors
         if actors is True:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3668,6 +3668,7 @@ class Scheduler(SchedulerState, ServerNode):
         fifo_timeout=0,
         annotations=None,
     ):
+
         dsk, dependencies, annotations = highlevelgraph_unpack(hlg, annotations)
 
         # Remove any self-dependencies (happens on test_publish_bag() and others)
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3856,8 +3856,8 @@ class Scheduler(SchedulerState, ServerNode):
                 for k, v in kv.items():
                     # Tasks might have been culled, in which case
                     # we have nothing to annotate.
-                    if k in parent._tasks:
-                        ts = parent._tasks[k]
+                    ts = parent._tasks.get(k)
+                    if ts is not None:
                         ts._annotations[a] = v
 
         # Add actors
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -6471,6 +6471,27 @@ async def test_annotations_resources(c, s, a, b):
     assert all([{"resources": {"GPU": 1}} == ts.annotations for ts in s.tasks.values()])
 
 
+@gen_cluster(
+    client=True,
+    nthreads=[
+        ("127.0.0.1", 1),
+        ("127.0.0.1", 1, {"resources": {"GPU": 1}}),
+    ],
+)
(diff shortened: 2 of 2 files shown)


distributed/scheduler.py
```diff
@@ -3854,8 +3853,11 @@ def update_graph(
 
             for a, kv in annotations.items():
                 for k, v in kv.items():
-                    ts = parent._tasks[k]
-                    ts._annotations[a] = v
+                    # Tasks might have been culled, in which case
+                    # we have nothing to annotate.
+                    if k in parent._tasks:
```
TomAugspurger, line 3859: @jakirkham I haven't followed the performance stuff too closely. Are we to the point where this `if` statement needs profiling?

distributed/scheduler.py
```diff
@@ -3854,8 +3854,11 @@ def update_graph(
...
+                        ts = parent._tasks[k]
```
jakirkham, line 3860: Typically we are handling these things with `.get` that avoids 2 retrievals and does justo one. Plus the `is None` check is a pointer comparison in Cython (so quite fast). Pure Python also does well with this kind of code

```suggestion
                    ts = parent._tasks.get(k)
                    if ts is not None:
```

distributed/scheduler.py
jakirkham: Answered a separate thread ( https://github.com/dask/distributed/pull/4544#pullrequestreview-597956717 ) since it was easier to suggest code that way

We are not asking anyone to do profiling though

TomAugspurger: Thanks!

jakirkham (approved):
LGTM. Thanks Tom! 😄

jakirkham:
Also if you can take a look Simon, that would be great! 😄

jrbourbeau (approved):
Thanks for the fix @TomAugspurger!

FWIW if possible, it'd be nice to only transmit annotations that apply to the tasks we're sending to the scheduler. Though that can be investigated in a later PR as the changes here are certainly an improvement on our current behavior.

sjperkins (approved):
This looks good to me. Thanks for cc'ing me on this @TomAugspurger, because I think I'll also have to handle similar concerns in Scheduler Plugins.

TomAugspurger:
> FWIW if possible, it'd be nice to only transmit annotations that apply to the tasks we're sending to the scheduler

Let me check on that today or tomorrow. It may be as simple as calling `unpack_annotations` a bit later.

TomAugspurger:
Well, it doesn't look that easy unfortunately. Even at https://github.com/dask/distributed/blob/d24d62f1dafa482b04209c1ef25a2305d6c41539/distributed/scheduler.py#L3733 in `update_graph`, we have the full set of `keys`. But it seems like (unless I'm mistaken) some of them are just not added to the schedulers internal state. So this looks like culling, but isn't using `dask.optimize` or `HighLevelGraph.cull`. Or I could be wrong about all that.

Anyway, it seems like we unpack the annotations somewhere well before we "cull" the unused keys. So I think we should be culling earlier (maybe before serialization?) and then we'll only unpack and expand annotations for keys we'll actually use.

sjperkins:
> Well, it doesn't look that easy unfortunately. Even at
> 
> https://github.com/dask/distributed/blob/d24d62f1dafa482b04209c1ef25a2305d6c41539/distributed/scheduler.py#L3733
> in `update_graph`, we have the full set of `keys`. But it seems like (unless I'm mistaken) some of them are just not added to the schedulers internal state. So this looks like culling, but isn't using `dask.optimize` or `HighLevelGraph.cull`. Or I could be wrong about all that.

You've probably considered https://github.com/dask/distributed/blob/d24d62f1dafa482b04209c1ef25a2305d6c41539/distributed/scheduler.py#L3745-L3759. Does anything show up in the logs though?

> Anyway, it seems like we unpack the annotations somewhere well before we "cull" the unused keys. So I think we should be culling earlier (maybe before serialization?) and then we'll only unpack and expand annotations for keys we'll actually use.

IIRC Annotation unpacking happens during this call

https://github.com/dask/distributed/blob/d24d62f1dafa482b04209c1ef25a2305d6c41539/distributed/scheduler.py#L3672

and there's a `update_graph` call almost directly after it, so I'm also a bit baffled.

jakirkham:
Understand we are looking to optimize things further (which is great btw! 😄). Though am curious, are we ok merging for now and optimizing afterwards? (have no strong feelings here) Mainly am interested in knowing whether we'd like to get the bug fix in the release tomorrow or whether it should wait a bit. Thoughts? 🙂

TomAugspurger:
I'd suggest merging this to fix the issue, since I don't see a straightforward reason why the set of `keys` passed to `update_graph` includes tasks that don't end up in `SchedulerState._tasks`.

jakirkham:
Sounds good. I'll plan on merging tomorrow morning. As I'm in US West that would be EOD for European timezones, so hopefully that gives Simon plenty of time to respond if there is an issue.

sjperkins:
> I'd suggest merging this to fix the issue, since I don't see a straightforward reason why the set of `keys` passed to `update_graph` includes tasks that don't end up in `SchedulerState._tasks`.

Seconded

jakirkham:
Great thanks all! 😀

Issue #4494, opened 2021-02-09
Timeseries Resource Usage Tracking


quasiben:
In Dask's dashboards we have good live tracking of many system resources including: CPU Utilization, Memory, Bandwidth, Num FDs, GPU Utilization, and GPU Memory:

### Examples of  Resource Tracking
<img width="1891" alt="Screen Shot 2021-02-09 at 9 51 57 AM" src="https://user-images.githubusercontent.com/1403768/107381052-74bef680-6abc-11eb-81c9-1abea9e552e1.png">

<img width="1912" alt="Screen Shot 2021-02-09 at 9 52 40 AM" src="https://user-images.githubusercontent.com/1403768/107381137-8dc7a780-6abc-11eb-8b6f-abc7887e2264.png">

Some times however, we want to ask questions of a workflow like what was the peak memory usage ?  What was the CPU/GPU utilization like over the course of a query ?  In trying to answer these and other questions, I think it would beneficial for this information to be include in performance reports.  Not sure how best to build this.  Perhaps starting with a generic timeseries tracking plot which sampled a resource. My first thought about attacking this problem was using a [scheduler plugin](https://docs.dask.org/en/latest/setup/custom-startup.html#example) for timeseries tracking which created a periodic callback to sample a particular resource.

We could also use [TaskState Metadata](https://github.com/dask/distributed/pull/4191/) for storing resource usage information

Here is a list of endpoints the scheduler is aware of:
https://github.com/dask/distributed/blob/7d2a22f94332974825e599dd779b4d98f44dbca2/distributed/dashboard/scheduler.py#L82-L92

jrbourbeau:
Would adding a new tab to the performance report which contains the "System" plot from the dashboard be a good start? 

> My first thought about attacking this problem was using a scheduler plugin for timeseries tracking which created a periodic callback to sample a particular resource

We might try hooking into the existing `SystemMonitor` periodic callback which runs on the scheduler and workers

https://github.com/dask/distributed/blob/f3f4bffea0640c01fc54f49c3219cf5807d14c66/distributed/core.py#L203-L209

charlesbluca:
> Would adding a new tab to the performance report which contains the "System" plot from the dashboard be a good start?

Made a quick attempt at this:

- https://github.com/charlesbluca/distributed/tree/report-system-tab

Pretty new to working with the dashboard so not really sure how to set the start time to when the performance report was called.

[...]

PR #4561, opened 2021-03-04
Add System tab to performance reports

charlesbluca:
This PR adds a System tab to performance reports matching the System plot from the dashboard, as discussed in #4494 as part of extending resource tracking in the performance reports:

image

Still a work in progress, as currently this outputs the entirety of the System plots, rather than just the section contained in the `performance_report` context.

- [x] Tests passed


jrbourbeau (commented):
Cool! Thanks for the screenshot @charlesbluca, this is nice to see

cc @quasiben @jacobtomlinson

charlesbluca:
Not sure what to make of the error on on ubuntu-3.9:

<details>
  <summary>Details</summary>
```
________________________________ test_terminate ________________________________

    def test_func():
        with clean() as loop:
            if iscoroutinefunction(func):
                cor = func
            else:
                cor = gen.coroutine(func)
>           loop.run_sync(cor, timeout=timeout)

distributed/utils_test.py:758: 

self = <tornado.platform.asyncio.AsyncIOLoop object at 0x7ff58115fbb0>
func = <function test_terminate at 0x7ff5cbde4430>, timeout = 10

    def run_sync(self, func: Callable, timeout: Optional[float] = None) -> Any:
        """Starts the `IOLoop`, runs the given function, and stops the loop.
    
        The function must return either an awaitable object or
        ``None``. If the function returns an awaitable object, the
        `IOLoop` will run until the awaitable is resolved (and
        `run_sync()` will return the awaitable's result). If it raises
        an exception, the `IOLoop` will stop and the exception will be
        re-raised to the caller.
    
        The keyword-only argument ``timeout`` may be used to set
        a maximum duration for the function.  If the timeout expires,
        a `tornado.util.TimeoutError` is raised.
    
        This method is useful to allow asynchronous calls in a
        ``main()`` function::
    
            async def main():
                # do stuff...
    
            if __name__ == '__main__':
                IOLoop.current().run_sync(main)
    
        .. versionchanged:: 4.3
           Returning a non-``None``, non-awaitable value is now an error.
    
        .. versionchanged:: 5.0
           If a timeout occurs, the ``func`` coroutine will be cancelled.
    
        """
        future_cell = [None]  # type: List[Optional[Future]]
    
        def run() -> None:
            try:
                result = func()
                if result is not None:
                    from tornado.gen import convert_yielded
    
                    result = convert_yielded(result)
            except Exception:
                fut = Future()  # type: Future[Any]
                future_cell[0] = fut
                future_set_exc_info(fut, sys.exc_info())
            else:
                if is_future(result):
                    future_cell[0] = result
                else:
                    fut = Future()
                    future_cell[0] = fut
                    fut.set_result(result)
            assert future_cell[0] is not None
            self.add_future(future_cell[0], lambda future: self.stop())
    
        self.add_callback(run)
        if timeout is not None:
    
            def timeout_callback() -> None:
                # If we can cancel the future, do so and wait on it. If not,
                # Just stop the loop and return with the task still pending.
                # (If we neither cancel nor wait for the task, a warning
                # will be logged).
                assert future_cell[0] is not None
                if not future_cell[0].cancel():
                    self.stop()
    
            timeout_handle = self.add_timeout(self.time() + timeout, timeout_callback)
        self.start()
        if timeout is not None:
            self.remove_timeout(timeout_handle)
        assert future_cell[0] is not None
        if future_cell[0].cancelled() or not future_cell[0].done():
>           raise TimeoutError("Operation timed out after %s seconds" % timeout)
E           tornado.util.TimeoutError: Operation timed out after 10 seconds

/usr/share/miniconda3/envs/dask-distributed/lib/python3.9/site-packages/tornado/ioloop.py:529: TimeoutError
```
</details>

jacobtomlinson:
I've triggered the CI again to see if this is a flaky issue.

charlesbluca:
Should the second CI run show up in checks here? I'm currently only seeing the failed runs from ~5 days ago.

charlesbluca:
These changes make it so the System tab now only plots the data gathered within the `performance_report` context.

This is done by grabbing event count from the scheduler's system monitor, and passing this through `performance_report` to the system monitor dashboard component, where it is used to filter the data being plotted.

One downside is that if the operation being profiled takes less than 250 ms, the System tab will have no data points to go off of and appear blank:

image

charlesbluca:
Looks like the tests are failing on `test_performance_report` because the client in the tests doesn't have an associated cluster. This could probably be circumvented by computing `last` within the dashboard component using the start time, something like

```python
self.last = next(i for i, t in enumerate(worker.monitor.time) if t >= start)
```

But that seems a little convoluted and its performance is determined by how large `worker.monitor.time` is - is there any other way to acces the system monitor from the `performance_report` class, or is that generally not recommended?

distributed/dashboard/components/shared.py
```diff
@@ -454,11 +454,14 @@ def trigger_update(self):
 
 
 class SystemMonitor(DashboardComponent):
-    def __init__(self, worker, height=150, **kwargs):
+    def __init__(self, worker, height=150, last=None, **kwargs):
         self.worker = worker
```
jacobtomlinson, line 458: Given that this is a reference to the scheduler in this case (not a worker, we should probably fix the name) could we not just grab `self.worker.monitor.count` here instead of having to have a client and cluster object?

charlesbluca: Since this `SystemMonitor` is being created after the operation being profiled has finished, I think using `self.worker.monitor.count` here would end up setting the start time of the system plot to after the operation has finished?

My idea was that we want to capture what the value of `self.worker.monitor.count` would have been around the time `performance_report.__aenter__()` is called, since this is also where the start time is set.

jacobtomlinson: Ah right I see. Yeah that makes sense.

quasiben:
> Looks like the tests are failing on test_performance_report because the client in the tests doesn't have an associated cluster. This could probably be circumvented by computing last within the dashboard component using the start time, something like

Dask users often run where the cluster object is not fully associated with the Client.  For example, connecting with a URL or create a client like `client = Client()`.  If we need information from the scheduler we can use the [run_on_scheduler` function](https://distributed.dask.org/en/latest/api.html#distributed.Client.run_on_scheduler).   

> client.run_on_scheduler(lambda dask_scheduler: dask_scheduler.monitor.count)

distributed/client.py
```diff
@@ -4614,6 +4614,7 @@ def __init__(self, filename="dask-report.html"):
 
     async def __aenter__(self):
         self.start = time()
+        self.last = get_client().cluster.scheduler.monitor.count
```
quasiben, line 4617: Can you think of a more descriptive name here ?  perhaps, `last_tick` or `last_count` ?

requirements.txt
```diff
@@ -1,10 +1,6 @@
 click >= 6.6
 cloudpickle >= 1.5.0
-contextvars;python_version<'3.7'
-# TODO: Update our dask version requirement to
-# the latest dask release before releasing distributed
-# xref https://github.com/dask/distributed/pull/4489
-dask @ git+https://github.com/dask/dask.git@master
+dask >= 2021.03.0
```
quasiben, line 3: Did you intend to edit this file ?  Probably should fetch latest from main

quasiben (commented):
Thanks for getting this together @charlesbluca !  What do you think about also listing max/min/avg for the 4 plots or maybe just for CPU and Memory?

distributed/client.py
charlesbluca: Sure - `last_count` sounds good! Is it okay if I apply this change across `SystemMonitor`, since I got this name from there?

quasiben: I think that would be fine.  Thanks!

requirements.txt
charlesbluca: Nope - I did this to fix local dev builds while the main/master switch was happening and committed unnecessarily - although this does seem to match the current upstream requirements?

https://github.com/dask/distributed/blob/7c6709fd8d2ee6356ba43ceff64cd04e9803b2a1/requirements.txt#L1-L4

Should I change this to pull the latest from the main branch?

quasiben: Yeah, I would pull from latest in main.

charlesbluca:
Thanks for the review and scheduler tip! Sure, I can look into adding that.

charlesbluca:
@quasiben just to make sure we're on the same page, was something like this what you had in mind?

image

I've also considered adding these stats to the worker tab as their own columns, and then adding the worker tab to the performance report.

quasiben:
Yeah, something like that!  I am having trouble visualizing what that might look like on the worker page but if you have thoughts/prototype that would be great to see.

requirements.txt
```diff
@@ -1,10 +1,6 @@
 click >= 6.6
 cloudpickle >= 1.5.0
-contextvars;python_version<'3.7'
-# TODO: Update our dask version requirement to
-# the latest dask release before releasing distributed
-# xref https://github.com/dask/distributed/pull/4489
-dask @ git+https://github.com/dask/dask.git@master
+dask @ git+https://github.com/dask/dask.git@main
```
charlesbluca, line 3: It looks like this is causing a merge conflict; should I just resolve this by going with the latest from main?

distributed/dashboard/components/shared.py
```diff
@@ -532,12 +569,22 @@ def __init__(self, worker, height=150, **kwargs):
...
     def update(self):
         with log_errors():
             self.source.stream(self.get_data(), 1000)
+            self.label_source.data["cpu"] = list(
+                "{}: {:.1f}%".format(f.__name__, f(self.source.data["cpu"]))
+                for f in [min, max, mean]
+            )
+            self.label_source.data["memory"] = list(
+                "{}: {}".format(
+                    f.__name__, dask.utils.format_bytes(f(self.source.data["memory"]))
```
charlesbluca, line 587: Using `format_bytes` here results in slightly different output of the human readable memory from the y-ticks - is this okay?

distributed/dashboard/components/shared.py
```diff
@@ -486,6 +511,18 @@ def __init__(self, worker, height=150, **kwargs):
         )
         self.mem.line(source=self.source, x="time", y="memory")
         self.mem.yaxis.axis_label = "Bytes"
+        self.mem.add_layout(
+            LabelSet(
```
charlesbluca, line 515: I didn't really configure the label text at all; is there anything that could be done here to make it more visually consistent with the plots?

distributed/dashboard/components/shared.py
```diff
@@ -454,12 +456,23 @@ def trigger_update(self):
...
-        self.last = 0
+        self.last_count = 0
+        if last_count is not None:
+            names = worker.monitor.range_query(start=last_count)
+            self.last_count = last_count
         self.source = ColumnDataSource({name: [] for name in names})
+        self.label_source = ColumnDataSource(
+            {
+                "x": [5] * 3,
+                "y": [70, 55, 40],
```
charlesbluca, line 471: These coordinates place the labels roughly in the middle of the left side of the plots - is there a better place they could go? I was thinking the top/bottom right corners, but that seemed to block the plot outputs more often.

charlesbluca (commented):
Just added the max/min/mean plot labels, here are some thoughts:

charlesbluca:
Also for the worker tab, I forgot that the scheduler isn't listed there - my idea would make sense if the scheduler was listed alongside the workers, then columns like `max(cpu)` could be used to get the aggregated stats of the workers & scheduler in a sortable view. I can also work on this, but I think it might be better for a separate PR.

requirements.txt
quasiben: Yes, pull latest from main

distributed/dashboard/components/shared.py
quasiben: What are the differences ?

distributed/dashboard/components/shared.py
quasiben: Hmm, that's a good question.  What do you think of adding a table below the plots ?  Does that result in occupying too much space on the page?  I like the idea of inlining the info but perhaps that clutter will cause confusion and/or difficulty reading.  If you don't think so that's ok

distributed/dashboard/components/shared.py
charlesbluca: `123.4MiB` (y-ticks) versus `123.4 MB` (`format_bytes`)

distributed/dashboard/components/shared.py
charlesbluca: I personally feel like a table below the plots would be equivalent to having a table in a separate tab, since in either case you'd only be able to look at one or the other (unless you're thinking the tables be directly under the plot they correspond to).

Maybe this info would be suited for a `HoverTool` so that it only pops up when you mouse over the plot?

distributed/dashboard/components/shared.py
quasiben: I don't have a strong opinion here.  What do you think ?

charlesbluca: My preference is `format_bytes`, although I'm sure if this is a problem later on it shouldn't be hard to find an alternative formatter.

quasiben:
@charlesbluca can you post a screenshot of the latest version ?  After posting, @mrocklin would you have some time to look this over ?  Ideally someone other than me could look this over and give objective feedback on design/implementation before merging in

charlesbluca:
Sure! Here's how it looks in the dashboard:

image

And in a performance report:

image

charlesbluca:
Revisiting this - is this still something we think would be a good addition to the dashboard / perf reports?

jacobtomlinson:
I think this will be really useful. 

I'm tempted to get CI passing (last fail seemed unrelated by I've restarted) and get this merged. Any other feedback can be addressed in follow up PRs.

cc @quasiben

quasiben:
The last failure here is due to a [conda/mamba thing](https://github.com/dask/distributed/pull/4561/checks?check_run_id=2673445175#step:3:6440).  Given the other tests passed I think it's safe to merge in.  Thank you so much @charlesbluca !  Thanks or the review @jacobtomlinson

PR #4569, opened 2021-03-08
Dask master -> main

jsignell:
This is the distributed side of https://github.com/dask/dask/pull/7198. There are probably similar PRs to be done on other repos, but it's not super urgent because github will redirect.


jakirkham (approved):

jrbourbeau (approved):
Thanks @jsignell!

PR #4576, opened 2021-03-10
Remove outdated macOS build badge from README

jrbourbeau:
We no longer run our macOS tests in a separate GitHub actions workflow

https://github.com/dask/distributed/blob/93bcc943bf380d7a774fd96ea64c2ba554e22301/.github/workflows/tests.yaml#L12

so we can remove the corresponding badge from our README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,15 +1,13 @@
 Distributed
 ===========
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks James! 😄

PR #4579, opened 2021-03-11
use `tobytes` instead of decode in serialization example

### LINKED ISSUE #4562: dask_serialize.register not working as described in the tutorial
Bug reports that follow these guidelines are easier to diagnose, and so are often handled much more quickly.
-->

**What happened**:
I'm following guide in the [Dask website](https://distributed.dask.org/en/latest/serialization.html) for registering a serialization but I have received the following error:

> Exception: No dispatch for <class '__main__.Human'>

I expected it to serialize the Human object.

```python
from distributed.protocol import dask_serialize, dask_deserialize

class Human:
    def __init__(self, name):
        self.name = name

@dask_serialize.register(Human)
def serialize(human):
    header = {}
    frames = [human.name.encode()]
    return header, frames

@dask_deserialize.register(Human)
def deserialize(header, frames):
    return Human(frames[0].decode())

if __name__ == '__main__':
    
    client = Client('localhost:8786')    
    futures = client.scatter([Human('Only')])

```

**Environment**:

- Dask version:'2021.02.0'
- Python version:3.6.12
- Operating System:Mac OS 10.14

quasiben:
- [x] Closes #4562
- [ ] Tests added / passed

Now that we are passing memoryviews we should use `tobytes` in the serialization example instead of `decode`

cc @jakirkham

### DIFF
--- a/docs/source/serialization.rst
+++ b/docs/source/serialization.rst
@@ -168,7 +168,7 @@ register them with Dask.
 
     @dask_deserialize.register(Human)
     def deserialize(header: Dict, frames: List[bytes]) -> Human:
-        return Human(frames[0].decode())
+        return Human(frames[0].tobytes())
 
 
 Traverse attributes


jrbourbeau (commented):
Thanks @quasiben! I think I'm missing why we would want to use `tobytes()` here instead of `decode()` -- since we're using `encode()` in the custom `serialize` function. 

For example, using the example as it currently is in our docs the following runs successfully for me

```python

from typing import Tuple, List, Dict
from distributed.protocol import dask_serialize, dask_deserialize, serialize, deserialize


class Human:
    def __init__(self, name):
        self.name = name

@dask_serialize.register(Human)
def serialize(human: Human) -> Tuple[Dict, List[bytes]]:
    header = {}
    frames = [human.name.encode()]
    return header, frames

@dask_deserialize.register(Human)
def deserialize(header: Dict, frames: List[bytes]) -> Human:
    return Human(frames[0].decode())

human = Human("bob")

# Roundtrip through serialization
human2 = deserialize(*serialize(human))

assert type(human2) is type(human)
assert human2.name == human2.name
```

docs/source/serialization.rst
```diff
@@ -168,7 +168,7 @@ register them with Dask.
 
     @dask_deserialize.register(Human)
     def deserialize(header: Dict, frames: List[bytes]) -> Human:
-        return Human(frames[0].decode())
+        return Human(frames[0].tobytes())
```
jakirkham, line 171: This would need an `import codecs` above, but would work for a `memoryview` or a `bytes` object

```suggestion
        return Human(codecs.decode(frames[0]))
```

jakirkham:
I think the missing piece is that frames are now `memoryview`s, which don't have a `decode` method. That said, agree the change is not quite right. Made a suggestion above that should address that

docs/source/serialization.rst
jakirkham: Also just to demonstrate this works...

```python
In [1]: import codecs

In [2]: mv = memoryview(b"abc")

In [3]: codecs.decode(mv)
Out[3]: 'abc'
```

jrbourbeau:
> frames are now memoryviews

Hrm playing around with this example it looks like `frames` is a `list` of `bytes`. For example the the `assert` below passes

```python
@dask_deserialize.register(Human)
def deserialize(header: Dict, frames: List[bytes]) -> Human:
    # Check whether or not frames is a list of bytes
    assert all(type(f) is bytes for f in frames)
    return Human(frames[0].decode())
```

quasiben:
> I think the missing piece is that frames are now memoryviews, which don't have a decode method. That said, agree the change is not quite right. Made a suggestion above that should address that

Yes.  Apologies @jrbourbeau for lack of context.  Frames is a list of bytes, but after passing through scheduler machinery it is coerced into a memoryview

jakirkham:
Right this may involve testing with `main`. Things changed after the last release

quasiben:
Now I don't know what's going on.  This reproducer show things working nicely with `encode/decode`:


```python
from distributed.utils import tmp_text

class Human:
    def __init__(self, name):
        self.name = name


if __name__ == '__main__':
    client = Client(n_workers=2)
    with tmp_text("myfile.py","""

from distributed.protocol import dask_serialize, dask_deserialize
class Human:
    def __init__(self, name):
        self.name = name

@dask_serialize.register(Human)
def serialize(human):
    header = {}
    frames = [human.name.encode()]
    return header, frames

@dask_deserialize.register(Human)
def deserialize(header, frames):
    return Human(frames[0].decode())
    """) as fn:
        client.upload_file(fn)
        [fut] = client.scatter([Human('Only')])
        print(fut)
        assert isinstance(fut.result(), Human)

```

jakirkham:
And that is using `main`?

In any event `codecs.decode` will work with `bytes` and `memoryview`s. So it should be fine regardless of what it gets for input

quasiben:
Yeah, this is using `main`.  I like the simplicity of encode/decode so I think I'll close this for now and the corresponding issue.  Apologies for the noise

PR #4581, opened 2021-03-11
Automatically cancel previous CI builds

jrbourbeau:
This PR adds a new GitHub actions workflow for canceling previous CI builds when rapidly pushing commits. We recently adding something similar over in `dask/dask` which has helped reduce our CI queue wait time (xref https://github.com/dask/dask/pull/7348)

cc @jakirkham

PR #4582, opened 2021-03-11
Test auto-merge 

jrbourbeau:
As title


jrbourbeau:
Alright, I see the "Enable auto-merge" button. I'm going to add a review approval to see what that does (but let's not actually merge this PR)

jrbourbeau:
Ah, PR authors can't approve their own PRs -- @jakirkham would you mind doing the honors?

jakirkham:
Prior to do anything, am already seeing auto-merge as an option. Though it does show merging is blocked without getting the approval. Curious what would happen if we told GH to automerge without the review being completed

Screen Shot 2021-03-11 at 11 16 03 AM

jakirkham (approved):

jakirkham:
That's weird. Apparently approving makes auto-merge go away?

Screen Shot 2021-03-11 at 11 18 30 AM

jrbourbeau:
Okay, so it looks like GitHub's auto-merge feature is tied to branch protection rules (these are things like requiring a certain number of review approvals or having required CI checks).

Before these rules are satisfied we can use auto-merge -- with the idea being that eventually the branch protection rules _will_ be satisfied (e.g. we're waiting on a CI build to finish). It looks after branch protection rules have been satisfied (i.e. in this case once there's a single review approval) the auto-merge button is replaced with the original merge button as the "wait until CI finishes" condition is no longer needed.

jakirkham (approved):

jakirkham:
Yep that sounds right

FWIW tried the same re-request review idea here as done in the other PR ( https://github.com/dask/distributed/pull/4581#issuecomment-796982486 ), but had the same result ( https://github.com/dask/distributed/pull/4581#issuecomment-796983134 ). IOW auto-merge does not reappear

jrbourbeau:
In our case, if we want PRs in the "This LGTM, will merge after CI finishes" state to be automatically merged, it seems like the natural branch protection rule to use is that all CI checks have passed. Though this won't always work when we have flaky tests in the test suite, which happens from time to time. 

I also wonder what happens when we've hit the auto-merge button on a PR that we're approved but is just wanting on CI and then CI fails due to a flaky test. Do we have to keep restarting CI until the test suite happens or can we override things and just manually merge the PR in w/o restarting CI?

jakirkham:
Yeah we might want to add a condition that CI passes

There's an adminstrative merge feature that owners have access to. This can be used to override and requirements and merge

jrbourbeau:
Closing in favor of continuing the conversation over in https://github.com/dask/community/issues/134. I think we have a good idea of how GitHub's automerge feature works, we now need to decide if we want to use it. 

For the time being, I'll also remove the PR review branch protection rule while we determine our next steps

PR #4584, opened 2021-03-13
Update PipInstall plugin command

jrbourbeau:
I tried out our `PipInstall` plugin for the first time and it didn't install the package I requested. It turns out we were putting the `pip install` CLI options in the wrong place. Today, for example if `pip_options = ["--upgrade"]`, we do

```
pip --upgrade install <packages>
```

we when should be doing

```
pip install --upgrade <packages>
```


quasiben:
Thanks @jrbourbeau .  Merging in

quasiben (approved):

PR #4592, opened 2021-03-15
Add support for a list of keys when using batch_size in client.map

### LINKED ISSUE #4588: Incorrect number of futures with client.map if both keys and batch_size are specified
If both `keys` and `batch_size` are specified, then the number of `futures` created with `client.map` is incorrect. Here's a reproducible example:

```python
client = Client()
futs = client.map(lambda x: x**2, list(range(20)), batch_size=5, key=[f"{x}" for x in list(range(20))])
print(len(futs)) # prints 80
```

If you try to collect these futures with `client.gather` then client will hang indefinitely.

This has to do with simultaneously specifying `batch_size` and custom `key` list, removing either of them will generate the correct number of futures.

I suspect the problem is in line 1701 here: https://github.com/dask/distributed/blob/4ce8c65039e9b531878f9d77cf884920690a3745/distributed/client.py#L1692-L1716

Specifically, it seems that the same keys are passed to every batch, this is problematic if key is a list. My suggested fix for this is to modify it as follows (code removed, it was faulty, see PR for working code).

Would be happy to submit PR if this code makes sense.

**Environment**:

- Dask version: 2021.02.0
- Python version: 3.6
- Operating System: MacOS

SultanOrazbayev:
This addresses issue #4588

- [x] Closes #4588
- [x] Tests passed (no new tests added)
- [x] Passes `black distributed` / `flake8 distributed` (flake8 finds some existing violations, but not related to changes in this PR)

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1693,7 +1693,7 @@ class Client:
             batches = list(
                 zip(*[partition_all(batch_size, iterable) for iterable in iterables])
             )
-            if type(key) == list:
+            if isinstance(key, list):
                 keys = [list(element) for element in partition_all(batch_size, key)]
             else:
                 keys = [key for _ in range(len(batches))]
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1693,12 +1693,10 @@ class Client:
             batches = list(
                 zip(*[partition_all(batch_size, iterable) for iterable in iterables])
             )
-            
             if type(key) == list:
                 keys = [list(element) for element in partition_all(batch_size, key)]
             else:
                 keys = [key for _ in range(len(batches))]
-            
             return sum(
                 [
                     self.map(
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1693,6 +1693,12 @@ class Client:
             batches = list(
                 zip(*[partition_all(batch_size, iterable) for iterable in iterables])
             )
+            
+            if type(key) == list:
+                keys = [list(element) for element in partition_all(batch_size, key)]
+            else:
+                keys = [key for _ in range(len(batches))]
+            
             return sum(
                 [
                     self.map(
@@ -1710,7 +1716,7 @@ class Client:
                         pure=pure,
                         **kwargs,
                     )
-                    for batch in batches
+                    for key, batch in zip(keys, batches)
                 ],
                 [],
             )
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -6530,3 +6530,4 @@ async def test_workers_collection_restriction(c, s, a, b):
     await future
     assert a.data and not b.data
 
+
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -6529,5 +6529,3 @@ async def test_workers_collection_restriction(c, s, a, b):
     future = c.compute(da.arange(10), workers=a.address)
     await future
     assert a.data and not b.data
-
-
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -230,6 +230,20 @@ async def test_map_batch_size(c, s, a, b):
 
 
 @gen_cluster(client=True)
+async def test_custom_key_with_batches(c, s, a, b):
+    """ Test of <https://github.com/dask/distributed/issues/4588>"""
+
+    futs = c.map(
+        lambda x: x ** 2,
+        range(10),
+        batch_size=5,
+        key=[str(x) for x in range(10)],
+    )
+    assert len(futs) == 10
+    await wait(futs)
+
+
+@gen_cluster(client=True)
 async def test_compute_retries(c, s, a, b):
     args = [ZeroDivisionError("one"), ZeroDivisionError("two"), 3]
 
@@ -6516,24 +6530,3 @@ async def test_workers_collection_restriction(c, s, a, b):
     await future
     assert a.data and not b.data
 
-
-@gen_cluster(client=True)
-async def test_custom_key_with_batches(c, s, a, b):
-    """ Test of <https://github.com/dask/distributed/issues/4588>"""
-
-    futs = c.map(
-        lambda x: x ** 2,
-        range(10),
-        batch_size=5,
-        key=[f"{x}" for x in range(10)],
-    )
-    assert len(futs) == 10
-    await wait(futs)
-
-
-    futs = c.map(
-        lambda x: x ** 2, list(range(10)), key=[f"{x}" for x in list(range(10))]
-    )
-    assert len(futs) == 10
-    res = c.gather(futs)  # this is to ensure that the futures are computed
-    assert len(res) == 10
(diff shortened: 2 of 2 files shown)


jacobtomlinson (changes requested):
Thanks for picking this up @SultanOrazbayev.

Would you be able to add a small test based on your reproducer from the issue?

SultanOrazbayev:
> Thanks for picking this up @SultanOrazbayev.
> 
> Would you be able to add a small test based on your reproducer from the issue?

Sure, I added a couple of tests.

SultanOrazbayev:
Oops, the test failed. Hmm, is there a way to see why? I can't figure out the response in the logs and my local test version doesn't execute (it throws an error: `ImportError: Error importing plugin "distributed.pytest_resourceleaks": cannot import name 'BasicLayer' from 'dask.highlevelgraph' (/Users/myuser/miniconda/envs/distributed/lib/python3.8/site-packages/dask/highlevelgraph.py)`

jrbourbeau (commented):
What version of `dask` are you using locally? Today `distributed` requires [`dask >= 2021.03.0`](https://github.com/dask/distributed/blob/b57d7239b94d733c768b7c6597a0c06e35b079ee/requirements.txt#L3)

jrbourbeau:
I also updated the title for this PR to be a bit more descriptive -- hope that's okay

distributed/tests/test_client.py
```diff
@@ -6515,3 +6515,30 @@ async def test_workers_collection_restriction(c, s, a, b):
...
+
+    futs = c.map(
+        lambda x: x ** 2,
+        list(range(10)),
+        batch_size=5,
+        key=[f"{x}" for x in list(range(10))],
+    )
+    assert len(futs) == 10
+    res = c.gather(futs)  # this is to ensure that the futures are computed
+    assert len(res) == 10
```
jrbourbeau, line 6532: This is equivalent

```suggestion
    await wait(futs)
```

distributed/tests/test_client.py
```diff
@@ -6515,3 +6515,30 @@ async def test_workers_collection_restriction(c, s, a, b):
...
+
+@gen_cluster(client=True)
+async def test_custom_key_with_batches(c, s, a, b):
+    """ Test of <https://github.com/dask/distributed/issues/4588>"""
+
+    futs = c.map(
+        lambda x: x ** 2,
+        list(range(10)),
+        batch_size=5,
+        key=[f"{x}" for x in list(range(10))],
```
jrbourbeau, line 6528:
```suggestion
        key=[f"{x}" for x in range(10)],
```

distributed/tests/test_client.py
```diff
@@ -6515,3 +6515,30 @@ async def test_workers_collection_restriction(c, s, a, b):
...
     assert a.data and not b.data
+
+
+@gen_cluster(client=True)
+async def test_custom_key_with_batches(c, s, a, b):
+    """ Test of <https://github.com/dask/distributed/issues/4588>"""
+
+    futs = c.map(
+        lambda x: x ** 2,
+        list(range(10)),
```
jrbourbeau, line 6526:
```suggestion
        range(10),
```

distributed/tests/test_client.py
```diff
@@ -6515,3 +6515,30 @@ async def test_workers_collection_restriction(c, s, a, b):
...
+        key=[f"{x}" for x in list(range(10))],
+    )
+    assert len(futs) == 10
+    res = c.gather(futs)  # this is to ensure that the futures are computed
+    assert len(res) == 10
+
+    futs = c.map(lambda x: x ** 2, list(range(10)), batch_size=5)
+    assert len(futs) == 10
+    res = c.gather(futs)  # this is to ensure that the futures are computed
+    assert len(res) == 10
```
jrbourbeau, line 6537: The case where no `key=` is specified is already tested in `test_map_batch_size` -- so I think we should be okay to remove this part of the test here 

```suggestion
```

distributed/tests/test_client.py
```diff
@@ -6515,3 +6515,30 @@ async def test_workers_collection_restriction(c, s, a, b):
...
+
+    futs = c.map(
+        lambda x: x ** 2, list(range(10)), key=[f"{x}" for x in list(range(10))]
+    )
+    assert len(futs) == 10
+    res = c.gather(futs)  # this is to ensure that the futures are computed
+    assert len(res) == 10
```
jrbourbeau, line 6544: The case where no `batch_size` is given and `key=` is a list is tested in `test_map_keynames` -- so I think we can remove this part of the test here too

distributed/client.py
```diff
@@ -1693,6 +1693,10 @@ def map(
             batches = list(
                 zip(*[partition_all(batch_size, iterable) for iterable in iterables])
             )
+            if type(key) == list:
```
jrbourbeau, line 1696:
```suggestion
            if isinstance(key, list):
```

distributed/tests/test_client.py
```diff
@@ -6515,3 +6515,30 @@ async def test_workers_collection_restriction(c, s, a, b):
     future = c.compute(da.arange(10), workers=a.address)
     await future
     assert a.data and not b.data
+
+
+@gen_cluster(client=True)
+async def test_custom_key_with_batches(c, s, a, b):
```
jrbourbeau, line 6521: Would you mind moving the contents of this test over to `test_map_batch_size` so all the checks for `client.map` + `batch_size=` are co-located?

jrbourbeau (commented):
Thanks for the fix @SultanOrazbayev!

distributed/tests/test_client.py
SultanOrazbayev: This one won't work, right?

SultanOrazbayev: Interesting, I thought that `range` was a generator, so wouldn't work with `client.map` but it works.

distributed/tests/test_client.py
SultanOrazbayev: Yes, that makes sense.

distributed/tests/test_client.py
SultanOrazbayev: Done.

SultanOrazbayev:
> What version of `dask` are you using locally? Today `distributed` requires [`dask >= 2021.03.0`](https://github.com/dask/distributed/blob/b57d7239b94d733c768b7c6597a0c06e35b079ee/requirements.txt#L3)

Yes, I'm using the latest dask version but it seems to be specific to my laptop, because I was able to start test run on my desktop with the same environment settings.

distributed/tests/test_client.py
jrbourbeau: `client.map` should work with anything that's an iterator

SultanOrazbayev:
Thank you very much for your help, @jrbourbeau !

jrbourbeau (approved):
Great, thanks for the updates @SultanOrazbayev!

jrbourbeau:
Also, I noticed this was your first code contribution to this repository. Welcome!

PR #4596, opened 2021-03-17
Move creation of SpecCluster _lock attribute to init

jacobtomlinson:
The asyncio lock attribute for `SpecCluster` is currently being created in `start`.

In some cases the lock may be accessed before `start` is called. This came up when implementing `from_name` in `dask-kubernetes`.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -256,6 +256,7 @@ class SpecCluster(Cluster):
         self._i = 0
         self.security = security or Security()
         self._futures = set()
+        self._lock = asyncio.Lock()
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Hmm I suspect that moving the `asyncio.Lock()` creation from `_start` to `__init__` is causing the lock to attach to the wrong event loop. At least there have been other cases where this was the case and we needed to move things like `asyncio.Lock` into a coroutine

jacobtomlinson:
Hrm that makes sense. I'll try and work around it another way downstream.

PR #4603, opened 2021-03-18
Update develop.rst

fjetter:
The `install.sh` script does no longer exit. This is different in the `dask/dask` repo

Would be great if somebody could confirm if this is working for windows.


quasiben (approved):
Thanks @fjetter

PR #4604, opened 2021-03-18
Make length of events deque configurable

fgebhart:
Hi there 👋 

this change fixes an OOM issue on the scheduler because of the increasing size of the events deque. The default value seems a bit too high. This issue can be avoided by making the value configurable.

~- [ ] Closes #xxxx~
- [x] Tests added / passed (Note: Some tests failed on my local machine with Timeout errors, but I assume this is related to my machine)

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -3279,7 +3279,11 @@ class Scheduler(SchedulerState, ServerNode):
         self.log = deque(
             maxlen=dask.config.get("distributed.scheduler.transition-log-length")
         )
-        self.events = defaultdict(lambda: deque(maxlen=100000))
+        self.events = defaultdict(
+            lambda: deque(
+                maxlen=dask.config.get("distributed.scheduler.transition-log-length")
+            )
+        )
         self.event_counts = defaultdict(int)
         self.worker_plugins = []
 
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -2184,3 +2184,19 @@ async def test_retire_state_change(c, s, a, b):
         step = c.compute(foo)
         c.gather(step)
     await c.retire_workers(workers=[a.address])
+
+
+@gen_cluster(client=True, config={"distributed.scheduler.transition-log-length": 3})
+async def test_configurable_events_log_length(c, s, a, b):
+    s.log_event("test", "dummy message 1")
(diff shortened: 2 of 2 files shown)


fjetter (approved):
LGTM. We should eventually adjust the default value as well (can be a different PR). We had a look at a single msg size (e.g. heartbeat) which already has about 3KB. With a deque size of 100k that makes a whooping 300MB per worker before the deque limit is hit.

Thank you @fgebhart for your first contribution!

distributed/scheduler.py
```diff
@@ -3279,7 +3279,11 @@ def __init__(
         self.log = deque(
             maxlen=dask.config.get("distributed.scheduler.transition-log-length")
         )
-        self.events = defaultdict(lambda: deque(maxlen=100000))
+        self.events = defaultdict(
+            lambda: deque(
+                maxlen=dask.config.get("distributed.scheduler.transition-log-length")
```
jrbourbeau, line 3284: Since the transition and events logging systems are independent of one another and store different types of data, users may want their corresponding `deque`s to have different maximum lengths. We could accomplish this by adding a separate `distributed.scheduler.events-log-length` config value to accompany the existing `distributed.scheduler.transition-log-length` value.

jrbourbeau (commented):
Thanks @fgebhart! 

I left one comment about using a separate config value for this. As the changes here are strictly an improvement on the current situation, we can merge this PR as is and leave adding a separate config value as follow-up work if you'd like

chrisroat:
This is a great find.  Keeping this smaller will help a lot.  Do you have recommendations on how small this can be set?

distributed/scheduler.py
fgebhart: Sounds reasonable to me. However, the idea was to provide the user with the possibility to avoid the OOM issue and I assumed somebody tweaking the `transition-log-length` settings would probably also want to reduce the overall memory footprint. I totally agree, that it makes sense to provide the possibility to configure these settings individually, but I would indeed prefer to keep this PR as is and add `distributed.scheduler.events-log-length` in a follow-up PR, if that is okay for you.

fgebhart:
> Do you have recommendations on how small this can be set?

Not yet. I will have to play around a bit with the configuration to see what could be a good default value.

jrbourbeau (approved):
Thanks @fgebhart!

PR #4609, opened 2021-03-19
Drop `log` from `remove_key_from_stealable`

jakirkham:
We already dropped the `log` line from `put_key_in_stealable` ( https://github.com/dask/distributed/pull/4375 ) because it was too granular of logging and was slowing things down there ( https://github.com/dask/distributed/issues/4359 ). This does the same thing for `remove_key_from_stealable` for the same reason.

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/stealing.py
+++ b/distributed/stealing.py
@@ -95,7 +95,6 @@ class WorkStealing(SchedulerPlugin):
             return
 
         worker, level = result
-        self.log(("remove-stealable", ts.key, worker, level))
         try:
             self.stealable[worker][level].remove(ts)
         except KeyError:


quasiben:
In scheduler.py, we sometimes have a debug logger note the event.  Do you think it's safe to remove entirely or should we add logging to help back out when/why these events occured

mrocklin (approved):

jakirkham:
> In scheduler.py, we sometimes have a debug logger note the event. Do you think it's safe to remove entirely or should we add logging to help back out when/why these events occured

Yeah we discussed in issue ( https://github.com/dask/distributed/issues/4359 ) and concluded it was ok to drop logging for the adding a single key case as it is rather granular. Removing a single key is the same level of granularity. So it seems reasonable to drop here as well

Suppose stealing could use the `verbose` attribute from the `Scheduler` to guard more detailed logging if that is needed (or handle this some other way). The main point is people shouldn't need to pay the cost of logging when they are not debugging

Do you have a need for this feature?

PR #4615, opened 2021-03-22
Introduce events log length config option

fgebhart:
Follow-up PR to #4604.

This adds the config parameter `distributed.scheduler.events-log-length` as a separate config option to accompany the existing `distributed.scheduler.transition-log-length` config option.

Note, that I also changed the default length of both `events-log-length` and `transition-log-length` to `5000` instead of `100000`, since it seems like a more reasonable default lengths to me. Let me know if you have any other suggestion for the default log length value.

~- [ ] Closes #xxxx~
- [x] Tests added / passed

### DIFF
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -18,6 +18,7 @@ distributed:
     events-cleanup-delay: 1h
     idle-timeout: null      # Shut down after this duration, like "1h" or "30 minutes"
     transition-log-length: 100000
+    events-log-length: 100000
     work-stealing: True     # workers should steal tasks from each other
     work-stealing-interval: 100ms  # Callback time for work stealing
     worker-ttl: null        # like '60s'. Time to live for workers.  They must heartbeat faster than this
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -80,6 +80,18 @@ properties:
               we will clear out old entries after a certain length.
               This is that length.
 
+          events-log-length:
+            type: integer
+            minimum: 0
+            description: |
+              How long should we keep the events log
+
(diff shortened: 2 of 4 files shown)


distributed/distributed.yaml
```diff
@@ -17,7 +17,8 @@ distributed:
     # after they have been removed from the scheduler
     events-cleanup-delay: 1h
     idle-timeout: null      # Shut down after this duration, like "1h" or "30 minutes"
-    transition-log-length: 100000
+    transition-log-length: 5000
```
mrocklin, line 20: If we want to change tihs default value then let's do so in another PR.  I recommend reverting this for now.

distributed/config.py
```diff
@@ -23,6 +23,7 @@
     "bandwidth": "distributed.scheduler.bandwidth",
     "default-data-size": "distributed.scheduler.default-data-size",
     "transition-log-length": "distributed.scheduler.transition-log-length",
+    "events-log-length": "distributed.scheduler.events-log-length",
```
mrocklin, line 26: These aliases are for backwards compatibility.  You don't need to add this here (no one was using the "events-log-length" config value previously.

mrocklin (commented):
This seems ok to me.  Two small comments.  Thanks @fgebhart

distributed/config.py
fgebhart: good to know, will remove it, thanks!

distributed/distributed.yaml
fgebhart: fine for me 👍

fgebhart:
Removed modifying the default value for the deque lengths.

jrbourbeau (approved):
Thanks for following up here @fgebhart!

PR #4623, opened 2021-03-23
Handle async clients when closing

### LINKED ISSUE #4620: Locks and LocalCluster
**What happened**:
Using Locks with a LocalCluster generates RuntimeWarnings (does not seem to generate the same warnings with other kinds of cluster).

```
.../lib/python3.8/site-packages/distributed/client.py:4717: RuntimeWarning: coroutine 'wait_for' was never awaited
  c.close(timeout=3)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
.../lib/python3.8/site-packages/distributed/client.py:4717: RuntimeWarning: coroutine 'Client._close' was never awaited
  c.close(timeout=3)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
.../lib/python3.8/site-packages/distributed/client.py:4717: RuntimeWarning: coroutine 'wait_for' was never awaited
  c.close(timeout=3)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

No warnings !

```python
import time

from distributed import Client, LocalCluster, Lock


def task(lock: Lock):
    with lock:
        time.sleep(1)


if __name__ == "__main__":
    cluster = LocalCluster()
    client = Client(cluster)

    lock_a = Lock(name="a", client=client)

    futures = []
    for i in range(2):
        f = client.submit(func=task, lock=lock_a)
        futures.append(f)

    client.gather(futures=futures)

    cluster.close()

```

**Environment**:

- Dask version: 2021.3.0
- Python version: 3.8.5
- Operating System: Ubuntu 16.04.7

mrocklin:
Fixes #4620

This probably occurs because a client created on a worker has its event loop thread, which in the worker case happens to be the main thread.  So the atexit call ends up happening in a situation where client.asynchronous is true.  

The close call may not ever actually occur (the event loop may be toast at this point) but we're doing best effort here at least.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4726,7 +4726,10 @@ def _close_global_client():
     if c is not None:
         c._should_close_loop = False
         with suppress(TimeoutError, RuntimeError):
-            c.close(timeout=3)
+            if c.asynchronous:
+                c.loop.add_callback(c.close, timeout=3)
+            else:
+                c.close(timeout=3)
 
 
 atexit.register(_close_global_client)


jrbourbeau (approved):
Thanks @mrocklin -- I confirmed this takes care of the warning in the original issue

PR #4624, opened 2021-03-23
Remove `utils.shutting_down` in favor of `sys.is_finalizing`

jrbourbeau:
This PR removes our `utils.shutting_down` utility in favor of using [`sys.is_finalizing`](https://docs.python.org/3/library/sys.html#sys.is_finalizing) directly throughout the codebase. IIUC `shutting_down` is a compatibility utility to support versions of Python that don't have `sys.if_finalizing`, which was introduced in Python 3.5 (our current minimum is 3.7).

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -5,6 +5,7 @@ from enum import Enum
 from functools import partial
 import inspect
 import logging
+import sys
 import threading
 import traceback
 import uuid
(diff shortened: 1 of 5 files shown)


distributed/core.py
```diff
@@ -406,7 +406,7 @@ async def listen(self, port_or_addr=None, allow_offload=True, **kwargs):
         )
         self.listeners.append(listener)
 
-    async def handle_comm(self, comm, shutting_down=shutting_down):
```
jrbourbeau, line 409: Grepping through the codebase I didn't see us ever utilize this `shutting_down=` keyword, so I've opted to inline `sys.is_finalizing` instead

mrocklin:
Nice.

PR #4627, opened 2021-03-23
Add GPU monitor to dashboard

charlesbluca:
In an attempt to address #4494, this PR adds a time series view of GPU memory and utilization to the dashboard, similar to the System tab:

image

Some things that could be improved:

- I'd like to add a hover tool with the worker name + current value of the metric; this seemingly requires this to be a Bokeh `multi_line` plot, which I tried out but wasn't able to get working just yet.
- Right now, the colors of the lines are generated by iterating through a 256-color Bokeh palette - this caps the number of lines to 256, which could cause problems for cases using more GPUs than this.

A plot like this would be nice in the performance reports, but I don't think this is possible right now as there isn't a way (accessible to the scheduler, at least) to collect logs for the workers outside of the dashboard component itself.

cc @quasiben 

- [ ] Tests added / passed

### DIFF
--- a/distributed/dashboard/components/nvml.py
+++ b/distributed/dashboard/components/nvml.py
@@ -47,7 +47,6 @@ class GPUMonitor(DashboardComponent):
         data = {f"{ws}_{metric}": [] for ws in workers for metric in ["mem", "util"]}
         data.update({"time": []})
         self.source = ColumnDataSource(data)
-        self.update()
 
         x_range = DataRange1d(follow="end", follow_interval=20000, range_padding=0)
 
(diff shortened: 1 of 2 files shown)


charlesbluca:
Converting this to draft as there's still a lot that should be sorted out.

jakirkham:
cc @jacobtomlinson

charlesbluca:
The failures here are related to the fact that I'm making a column for each worker in the CDS, which isn't Bokeh best practice - instead, I should be opting to use 2-dimensional columns corresponding to the timestamps and metric values for all workers, to be plotted using `bokeh.MultiLine`. However, a CDS formatted this way doesn't support incremental data streaming yet (bokeh/bokeh#11101), so until then this PR is blocked.

charlesbluca:
Closing this as the general consensus from #4668 was that we probably don't want multi-worker time plots like this unless they are more aggregated.

PR #4643, opened 2021-03-28
DOC: Update out-of-date references to `config.yaml`

hristog:
- [ ] ~~Closes #xxxx~~ N/A
- [ ] ~~Tests added / passed~~ N/A
- [ ] ~~Passes `black distributed` / `flake8 distributed`~~ N/A (locally, `black==19.10b0` suggests that there should be an unrelated update elsewhere in `distributed/client.py`, but I'm not convinced that's an actual issue, because it would've been addressed beforehand, otherwise)

Extracted out of https://github.com/dask/distributed/pull/4642.
The proposed PR updates docstring contents (outdated references to config file and associated config items) in two instances in `distributed/client.py`.

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3495,7 +3495,8 @@ class Client:
         ----------
         n : int
             Number of logs to retrive.  Maxes out at 10000 by default,
-            confiruable in config.yaml::log-length
+            configurable via the ``distributed.admin.log-length``
+            configuration value.
 
         Returns
         -------
@@ -3510,7 +3511,8 @@ class Client:
         ----------
         n : int
             Number of logs to retrive.  Maxes out at 10000 by default,
-            confiruable in config.yaml::log-length
+            configurable via the ``distributed.admin.log-length``
+            configuration value.
         workers : iterable
             List of worker addresses to retrieve.  Gets all workers by default.
         nanny : bool, default False


distributed/client.py
```diff
@@ -3494,8 +3494,8 @@ def get_scheduler_logs(self, n=None):
         Parameters
         ----------
         n : int
-            Number of logs to retrive.  Maxes out at 10000 by default,
-            confiruable in config.yaml::log-length
+            Number of logs to retrive.  Maxes out at 100000 by default,
+            confiruable in distributed.yaml::transition-log-length
```
jrbourbeau, line 3498: Both the scheduler and worker log length are configured using the [`"distributed.admin.log-length"`](https://github.com/dask/distributed/blob/25a2867e8e950d9cb51b071337ffc43ddaf0bb52/distributed/distributed.yaml#L176) configuration value (see [here](https://github.com/dask/distributed/blob/25a2867e8e950d9cb51b071337ffc43ddaf0bb52/distributed/node.py#L80-L97) for where this happens) 

```suggestion
            configurable with the ``distributed.admin.log-length``
            configuration value. 
```

distributed/client.py
```diff
@@ -3494,8 +3494,8 @@ def get_scheduler_logs(self, n=None):
         Parameters
         ----------
         n : int
-            Number of logs to retrive.  Maxes out at 10000 by default,
-            confiruable in config.yaml::log-length
+            Number of logs to retrive.  Maxes out at 100000 by default,
```
jrbourbeau, line 3497: The previous value is correct

distributed/client.py
```diff
@@ -3509,8 +3509,8 @@ def get_worker_logs(self, n=None, workers=None, nanny=False):
         Parameters
         ----------
         n : int
-            Number of logs to retrive.  Maxes out at 10000 by default,
-            confiruable in config.yaml::log-length
+            Number of logs to retrive.  Set to 0 by default,
+            confiruable in distributed.yaml::recent-messages-log-length
```
jrbourbeau, line 3513: We can use the same description here as in `get_scheduler_logs`

jrbourbeau (commented):
Thanks for the PR @hristog

jrbourbeau (approved):
Thanks for the updates @hristog! This is in

PR #4644, opened 2021-03-29
[docs] expand documentation on pure functions

jameslamb:
- [ ] Closes #xxxx
- [x] Tests added / passed

## Changes in this PR

* adds plain-language definition of "pure function" to the Client documentation
    - image

* links to the expanded doc on pure functions from the API docs on `Client.submit()` and `Client.map()`
    - image

## Background

I found myself working on a problem tonight that involved carefully controlling the behavior of `Client.submit()`, so I was reading the API docs for that method (https://distributed.dask.org/en/latest/api.html#distributed.Client.submit) to understand what could be controlled.

The documentation for `pure` doesn't provide any information about what "pure" actually means.

> Whether or not the function is pure. Set pure=False for impure functions like np.random.random.

I think that relies on users already being familiar with the concept of "pure functions". I personally was not familiar with that phrase, so I was left wondering "...is my function pure?".

Searching the rest of the documentation, I did find https://distributed.dask.org/en/latest/client.html#pure-functions-by-default which expands on this concept. But in my opinion, that section also doesn't define what "pure" is and is written for an audience that already understands that phrase. I didn't notice, until I looked at the raw `.rst` file for that page, that the word "pure" in the first sentence is a link out to [the `toolz` documentation](https://toolz.readthedocs.io/en/latest/purity.html).

image

I suspect that other people using `distributed` might not be familiar with the phrase "pure function".

I believe this PR's proposed changes make the documentation for the argument `pure` a bit more approachable, and make it easier to find the expanded docs about pure functions from the API docs for `Client.submit()`.

## Notes for reviewers

I tested the documentation locally with the following code, and didn't noticed any new warnings introduced by this change.

```shell
python setup.py install
cd docs
pip install -r requirements.txt
make clean html
```

Thanks for your time and consideration!

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1512,6 +1512,7 @@ class Client:
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
+            See "Pure Functions by Default" in :doc:`client` for more details.
         workers : string or iterable of strings
             A set of worker addresses or hostnames on which computations may be
             performed. Leave empty to default to all workers (common case)
@@ -1631,6 +1632,7 @@ class Client:
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
+            See "Pure Functions by Default" in :doc:`client` for more details.
         workers : string or iterable of strings
             A set of worker hostnames on which computations may be performed.
             Leave empty to default to all workers (common case)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1512,7 +1512,7 @@ class Client:
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
-            See "Pure Functions by Default" in :doc:`client` for more details.
+            See :ref:`pure functions` for more details.
         workers : string or iterable of strings
             A set of worker addresses or hostnames on which computations may be
             performed. Leave empty to default to all workers (common case)
@@ -1632,7 +1632,7 @@ class Client:
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
-            See "Pure Functions by Default" in :doc:`client` for more details.
+            See :ref:`pure functions` for more details.
         workers : string or iterable of strings
             A set of worker hostnames on which computations may be performed.
             Leave empty to default to all workers (common case)
--- a/docs/source/client.rst
+++ b/docs/source/client.rst
@@ -108,6 +108,7 @@ which are used for larger and smaller result sets respectively.
 
 For more information see the page on :doc:`Managing Computation <manage-computation>`.
 
+.. _pure functions:
 
 Pure Functions by Default
 -------------------------
@@ -119,7 +120,7 @@ Pure functions:
 * always return the same output for a given set of inputs
 * do not have side effects, like modifying global state or creating files
 
-If this is not the case, you should use the ``pure=False`` keyword argument.
+If this is not the case, you should use the ``pure=False`` keyword argument in methods like ``Client.map()`` and ``Client.submit()``.
 
 The client associates a key to all computations.  This key is accessible on
 the Future object.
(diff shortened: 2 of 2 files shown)


docs/source/client.rst
```diff
@@ -112,8 +112,14 @@ For more information see the page on :doc:`Managing Computation <manage-computat
 Pure Functions by Default
 -------------------------
 
-By default we assume that all functions are pure_.  If this is not the case we
-should use the ``pure=False`` keyword argument.
+By default, ``distributed`` assumes that all functions are pure_.
+
+Pure functions:
```
jrbourbeau, line 117: Small formatting nitpick

```suggestion
By default, ``distributed`` assumes that all functions are pure_. Pure functions:
```

docs/source/client.rst
```diff
@@ -112,8 +112,14 @@ For more information see the page on :doc:`Managing Computation <manage-computat
...
+
+* always return the same output for a given set of inputs
+* do not have side effects, like modifying global state or creating files
+
+If this is not the case, you should use the ``pure=False`` keyword argument.
```
jrbourbeau, line 122: I realize this predates this PR, but could you mention where users can specify the `pure=` keyword (i.e. `client.submit` and `client.map`)

distributed/client.py
```diff
@@ -1512,6 +1512,7 @@ def submit(
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
+            See "Pure Functions by Default" in :doc:`client` for more details.
```
jrbourbeau, line 1515: What do you think about linking directly to this section?

jrbourbeau (commented):
Thanks @jameslamb! I left a few small comments, but overall this is a nice improvement

distributed/client.py
jameslamb: I think it's preferable to link directly to the section, but I actually don't know how to do it with `:doc` directives, and I thought it would be a deviation from this repo's style to do something like this:

```text
`Pure Functions by Default <./client.html#pure-functions-by-default>`_
```

If you know how to link to an anchor with a `:doc:` directive, could you show me how? The word "doc" is especially hard to search for in Sphinx-related documentation :stuck_out_tongue:

jrbourbeau: I'm not sure if there's a way to do this with the `:doc:` directive. However something: 

```diff
diff --git a/distributed/client.py b/distributed/client.py
index cd60ace2..9d0f2ec1 100644
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1512,7 +1512,7 @@ class Client:
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
-            See "Pure Functions by Default" in :doc:`client` for more details.
+            See :ref:`pure-functions` for more details.
         workers : string or iterable of strings
             A set of worker addresses or hostnames on which computations may be
             performed. Leave empty to default to all workers (common case)
@@ -1632,7 +1632,7 @@ class Client:
         pure : bool (defaults to True)
             Whether or not the function is pure.  Set ``pure=False`` for
             impure functions like ``np.random.random``.
-            See "Pure Functions by Default" in :doc:`client` for more details.
+            See :ref:`pure-functions` for more details.
         workers : string or iterable of strings
             A set of worker hostnames on which computations may be performed.
             Leave empty to default to all workers (common case)
diff --git a/docs/source/client.rst b/docs/source/client.rst
index 508d8349..2679164c 100644
--- a/docs/source/client.rst
+++ b/docs/source/client.rst
@@ -108,6 +108,7 @@ which are used for larger and smaller result sets respectively.

 For more information see the page on :doc:`Managing Computation <manage-computation>`.

+.. _pure-functions:

 Pure Functions by Default
 -------------------------
```

should work (we use this lots of other places throughout the docs)

jameslamb: oooo nice, that works!

image

changed in https://github.com/dask/distributed/pull/4644/commits/7a207260b9f31efd0c0a5d7d36ccbf836a519135

docs/source/client.rst
jameslamb: added in https://github.com/dask/distributed/pull/4644/commits/7a207260b9f31efd0c0a5d7d36ccbf836a519135

jrbourbeau (approved):
Thanks @jameslamb! 

Note the CI failures are unrelated to the changes here

PR #4648, opened 2021-03-29
Install pytorch from conda-forge

jrbourbeau:
xref https://github.com/dask/distributed/pull/4585#issuecomment-809636227

cc @jakirkham @crusaderky


jakirkham:
Ah it looks like we don't have `pytorch` on Windows. So don't think this is going to work after all unfortunately

jrbourbeau:
Worth a shot, closing

jakirkham:
Yep thanks for trying. Sorry wasn't thinking about Windows

PR #4638, opened 2021-03-26
Avoid repeatedly using the same worker on first task with quiet cluster.

douglasdavis:
This adds an additional check to avoid quiet clusters (as Matt describes in #4637) repeatedly using the same worker (the new test fails without the added sum of occupancies check). I understand this is a CPU sensitive area, and perhaps the sum of occupancies is too expensive a task to use here. Very happy to hear suggestions to try alternatives.

- [x] Closes #4637
- [x] Tests added / passed
- [x] Passes `black distributed` / `flake8 distributed`

Issue #4513, opened 2021-02-17
Using asyncio directly in TCP


jakirkham:
As part of the Scheduler benchmarking work, we have seen a fair bit of time still spent in Tornado. While we can now use uvloop ( https://github.com/dask/distributed/pull/4448 ) and have optimized TCP communication ( https://github.com/dask/distributed/pull/4506 ), this still remains. In figuring out how we might fix this, have done some reading on [asyncio's Transports and Protocols]( https://docs.python.org/3/library/asyncio-protocol.html ), which uvloop in turn uses to provide [significant performance gains]( https://github.com/MagicStack/uvloop#performance ).

It seem doable (though may be a bit of work) to reimplement TCP communication using this strategy, which would also allow us to benefit from uvloop's performance improvements in this area. Hoping after we do this with TCP, we can apply something similar to UCX as well (though this may need a bit of generalization).

Am curious if this is something we would be open to. Any thoughts? 🙂

cc @mrocklin @quasiben

mrocklin:
I'm totally open to this.  I had an asyncio Comm PR a while back, but it stalled.  This may not be what you're referring to with transports and protocols though.  @jcrist has done some work here in side projects (I think https://github.com/jcrist/ery) and may have thoughts.

However, after looking again at profiles recently I think that `extract_serialize` is probably our biggest bottleneck on the comms side (or perhaps overall).  Getting that done should be, I think, our first priority.

jakirkham:
Good to know. Yeah it would be great to get Jim's and others thoughts here

IIUC that is something @madsbk is looking into

jakirkham:
Though it sounds like Jim has also done work on the serialization side ( https://github.com/jcrist/msgspec ). Am wondering if it makes sense to discuss how we could (optionally) use Jim's work on both these fronts in Distributed. Maybe it's too early to ask that question, but since we know these are bottlenecks we'd like to address it seems worth surfacing

mrocklin:
My inclination is to wait on these decisions until we get `extract_serialize` done.  For example, if Mads is occupied this week then I wonder if `extract_serialize` is something that you could take over.  If Mads is doing it and if it's on track to be done soon then great, let's figurre out the next largest bottleneck.  The sense I've gotten recently is that Mads might be doing a few things, which is why I bring this up.

[...]

PR #4650, opened 2021-03-30
Add C APIs for `transition_*` functions

jakirkham:
To cutdown on the overhead of calling `transition_*` functions, add C APIs for these functions to allow Cython to call these with C.

- [ ] Closes #xxxx
- [ ] Tests added / passed


jakirkham:
Seeing this on CI

```python
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4828, in handle_task_finished
    r: tuple = self.stimulus_task_finished(key=key, worker=worker, **msg)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4214, in stimulus_task_finished
    r: tuple = parent._transition(key, "memory", worker=worker, **kwargs)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1864, in _transition
    a: tuple = func(key, *args, **kwargs)
TypeError: transition_processing_memory() got an unexpected keyword argument 'status'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2224608533

jakirkham:
Seeing this on CI

```python
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4828, in handle_task_finished
    r: tuple = self.stimulus_task_finished(key=key, worker=worker, **msg)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4214, in stimulus_task_finished
    r: tuple = parent._transition(key, "memory", worker=worker, **kwargs)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1864, in _transition
    a: tuple = func(key, *args, **kwargs)
TypeError: transition_processing_memory() got an unexpected keyword argument 'thread'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2224796235

jakirkham:
Seeing this on CI

```python
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4836, in handle_task_finished
    r: tuple = self.stimulus_task_finished(key=key, worker=worker, **msg)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4222, in stimulus_task_finished
    r: tuple = parent._transition(key, "memory", worker=worker, **kwargs)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1872, in _transition
    a: tuple = func(key, *args, **kwargs)
TypeError: transition_processing_memory() got an unexpected keyword argument 'metadata'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2224960448

jakirkham:
Seeing this on CI

```python
  File "/home/runner/work/distributed/distributed/distributed/utils.py", line 668, in log_errors
    yield
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 3873, in add_worker
    await self.handle_worker(comm=comm, worker=address)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4959, in handle_worker
    await self.handle_stream(comm=comm, extra={"worker": worker})
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4848, in handle_task_erred
    r: tuple = self.stimulus_task_erred(key=key, **msg)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4271, in stimulus_task_erred
    **kwargs,
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1873, in _transition
    a: tuple = func(key, *args, **kwargs)
TypeError: transition_processing_erred() got an unexpected keyword argument 'worker'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2225283910

jakirkham:
Seeing this on CI

```python
distributed.core - ERROR - transition_processing_erred() got an unexpected keyword argument 'startstops'
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4848, in handle_task_erred
    r: tuple = self.stimulus_task_erred(key=key, **msg)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4264, in stimulus_task_erred
    r = parent._transition(
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1873, in _transition
    a: tuple = func(key, *args, **kwargs)
TypeError: transition_processing_erred() got an unexpected keyword argument 'startstops'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2225419119

jakirkham:
Seeing this on CI

```python
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4854, in handle_task_erred
    r: tuple = self.stimulus_task_erred(key=key, **msg)
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4270, in stimulus_task_erred
    r = parent._transition(
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1873, in _transition
    a: tuple = func(key, *args, **kwargs)
TypeError: transition_processing_erred() got an unexpected keyword argument 'text'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2225540455

jakirkham:
@mrocklin @quasiben above are some examples of the extra arguments that are being passed to `transition_` functions that are not needed

jakirkham:
@mrocklin could you please weigh in on how we should handle these extra arguments?

mrocklin:
In some of these cases I suspect that we do actually use these keyword arguments, but we use them in plugins.  For example the `thread=` keyword is used by the task stream plot.  https://github.com/dask/distributed/blob/main/distributed/diagnostics/task_stream.py

Maybe we need to send an extras dict?

mrocklin:
Also, thanks to @jrbourbeau for pointing me to this.  

For reference @jakirkham my e-mail is no longer reliably read.  If you're blocked on me specifically then you're probably better served by trying to connect out-of-band.  My apologies for the lack of responsiveness.

mrocklin:
> Maybe we need to send an extras dict?

Or, maybe easier would be to just include them as explicit keywords even if they aren't used in these particular methods.

mrocklin:
I'm enthusiastic about this change by the way.  I look forward to seeing if it has an impact.

jakirkham:
If these are meant for plugins called in `_transition`, would it make sense to extract these at that level?

mrocklin:
Oh yeah, maybe.  That makes sense.  Presumably the SchedulerState class is intended to be stripped down anyway.  That sounds like a great solution.

jakirkham:
Yeah I think that is next thing is making a clean break between Scheduler and SchedulerState (instead of inheriting from Scheduler)

mrocklin:
To be clear, my understanding is that that change would not be necessary in order to implement the changes here though.  That's more future-looking.  Is that correct?

jakirkham:
Yeah that's future looking and unrelated to this

douglasdavis:
> Yeah I think that is next thing is making a clean break between Scheduler and SchedulerState (instead of Scheduler inheriting from SchedulerState)

@jakirkham chiming in to say I'd be interested in helping in this area (if there's room for it!). Would be happy to have a quick chat about status of/plans for the effort if you think it would be worthwhile

jakirkham:
Communication is one area we could improve that would help. Shared some thoughts in this comment ( https://github.com/dask/distributed/issues/4513#issuecomment-817049499 )

distributed/scheduler.py
```diff
@@ -1940,7 +1943,19 @@ def _transition(
...
+                            finish2,
+                            *args,
+                            status=status,
+                            thread=thread,
+                            metadata=metadata,
+                            worker=worker,
+                            startstops=startstops,
+                            text=text,
+                            **kwargs,
+                        )
```
fjetter, line 1958: I don't think this is a dealbreaker but that _could_ be perceived as a breaking change, couldn't it? At least this is a subtle change in the required signature. So far we document this as `*args, **kwargs`, see https://distributed.dask.org/en/latest/plugins.html#distributed.diagnostics.plugin.SchedulerPlugin.transition

jakirkham: I don't think so. These are already passed as keyword arguments and we are continuing to pass them that way

That said, Idk if this is how things will shake out yet. Things are still changing here. IOW I'm not sure this is worth reviewing yet

mrocklin: +1 to what John said.  I think that this is a no-op if you consider the extraction of these from kwargs above.

jakirkham:
Ok this is ready for review! 😄

Am seeing some CI failures though, but they appear related to flaky tests seen elsewhere ( https://github.com/dask/distributed/pull/4638#pullrequestreview-634750417 )

jrbourbeau (commented):
It looks like `test_eventstream` and `test_TaskStreamPlugin` are genuine (non-flaky) test failures

jakirkham:
Think the other issues are fixed. Though started seeing `test_dashboard` failing. Saw there were some changes in `main` to this test. So tried to merge those in, but it looks like that might not work either. Appears `test_dashboard` has been having issues elsewhere though ( https://github.com/dask/distributed/issues/4697 ). So maybe it's not related to these changes

fjetter:
I see errors connecting to the transitioning code in the ubu-py3.7 tests

```
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 1963, in _transition
    raise RuntimeError(f"Impossible transition from {start} to {finish}")
RuntimeError: Impossible transition from no-worker to released
distributed.scheduler - ERROR - Error transitioning 'lambda-df0fe8d7604e475821f9b929791d1a43' from 'no-worker' to 'waiting'
```

https://github.com/dask/distributed/pull/4650/checks?check_run_id=2340272649

jakirkham:
Yeah I pushed some more changes last night after fixing some errors. So there may be some new ones

distributed/scheduler.py
```diff
@@ -1826,7 +1817,57 @@ def new_task(self, key: str, spec: object, state: str) -> TaskState:
...
+        elif start is "memory":  # noqa: F632
+            if finish is "released":  # noqa: F632
+                return self.transition_memory_released(*args, **kwargs)
+            elif finish is "forgotten":  # noqa: F632
+                return self.transition_memory_forgotten(*args, **kwargs)
+        elif start is "erred":  # noqa: F632
+            if finish is "released":  # noqa: F632
+                return self.transition_erred_released(*args, **kwargs)
+            elif finish is "forgotten":  # noqa: F632
+                return self.transition_released_forgotten(*args, **kwargs)
```
jakirkham, line 1859: From @mrocklin's comment ( https://github.com/dask/distributed/commit/8ff10d95b319b82282b6dc8742d889b939952fe7#r49507747 ).

> Does this change result in a meaningful speedup? This seems significantly more complex to me. I agree that it's worth doing, but only if it has an impact. If possible I think that we should try to measure the impact of this PR before merging. Thoughts?

(pasted here as the comment wasn't in the PR)

jakirkham: So the goal of this change is to try and cutdown on time spent calling the `transition_*` functions themselves by moving those calls to C. To benefit from adding C APIs to those functions we need to call them directly. Doing it indirectly will just go through Python so won't have any benefit. IOW something like this will be needed

We could flatten the `if`s by comparing a single `str` (maybe like `"{start} {finish}"`?) to make this a bit simpler/more readable. May play with this a bit. Am open to other suggestions to improve readability/maintainability here

It's worth noting that today this lives in a `dict` that is roughly as long as this code block

mrocklin: Yeah, I understand why, from first principles, this is a good idea.  I'm curious what impact is has on benchmarks and performance.  

If this improves benchmarks by a percent or two then awesome, let's good for it.  If it's only shaving off a couple of nanoseconds then maybe we shouldn't do this.  

I think that right now our development flow is to make some changes, merge them in, and then see how benchmarks respond.  I'm suggesting that we may want to roll those benchmarks in earlier in the process so that we make changes, look at how they change benchmarks/profiling, and then based on those improvements we decide to merge.  

Thoughts?

jakirkham: Yeah I think that is fine. Was just trying to provide more context as well as make sure we are comfortable with this kind of change before diving deeper. Sound like we are ok with it in principle, but would like to make sure it is delivering value, which makes sense to me

jakirkham:
Regarding the `"no-worker"` issue noted above. Python literals are normally `intern`ed automatically in pure Python. Cython also will `intern` and manage these globally for the module. For example...

```python
In [1]: s1 = "processing"

In [2]: s2 = "processing"

In [3]: s1 is s2
Out[3]: True
```

However it seems `str`s with a `"-"` don't get `intern`ed automatically.

```python
In [1]: s1 = "no-worker"

In [2]: s2 = "no-worker"

In [3]: s1 is s2
Out[3]: False
```

This is a non-issue when using `"_"`.

```python
In [1]: s1 = "no_worker"

In [2]: s2 = "no_worker"

In [3]: s1 is s2
Out[3]: True
```

So replaced `"-"` with `"_"` in the dispatch table to fix this issue

distributed/scheduler.py
mrocklin: @jakirkham I've noticed you active on github again.  Welcome back from vacation?  I hope that you had a relaxing time.

I'm curious to learn how much impact this PR had.  I suspect that this is already on your queue of work coming back (which I imagine is long) but I thought I'd throw a :+1: on it anyway.

jakirkham: Yeah I had another idea on how to improve things here, but as guessed other things are at the forefront of the queue atm 😅

distributed/scheduler.py
```diff
@@ -1975,7 +1966,50 @@ def new_task(self, key: str, spec: object, state: str) -> TaskState:
     # State Transitions #
     #####################
 
-    def _transition(self, key, finish: str, *args, **kwargs):
+    def _transition_dispatch(self, start: str, finish: str, *args, **kwargs):
+        start_finish: str = sys_intern(f"{start}_{finish}".replace("-", "_"))
```
martindurant, line 1970: I'd be interested in a speed test of this approach versus `hash`/dict lookup or actual `strcmp` (which is what I assume cython does when both operands are `char*`). For the latter, you don't need the string manipulation to make the compound, and the if statements can be grouped.

distributed/scheduler.py
```diff
@@ -2033,11 +2064,12 @@ def _transition(self, key, finish: str, *args, **kwargs):
                 a_recs, a_cmsgs, a_wmsgs = a
 
                 v = a_recs.get(key, finish)
-                func = self._transitions_table["released", v]
                 b_recs: dict
                 b_cmsgs: dict
                 b_wmsgs: dict
-                b: tuple = func(key)
+                b: tuple = self._transition_dispatch(
```
martindurant, line 2070: Don't need `b` here, can combine with next line.

distributed/scheduler.py
```diff
@@ -2070,7 +2102,7 @@ def _transition(self, key, finish: str, *args, **kwargs):
 
                 start = "released"
             else:
-                raise RuntimeError("Impossible transition from %r to %r" % start_finish)
+                raise RuntimeError(f"Impossible transition from {start} to {finish}")
```
martindurant, line 2105: Would change the logic a little, but would make sense to me to have `_transition_dispatch` raise this error if none of the cases match.

distributed/scheduler.py
```diff
@@ -1975,7 +1966,50 @@ def new_task(self, key: str, spec: object, state: str) -> TaskState:
...
+        elif start_finish is "memory_forgotten":  # noqa: F632
+            return self.transition_memory_forgotten(*args, **kwargs)
+        elif start_finish is "erred_released":  # noqa: F632
+            return self.transition_erred_released(*args, **kwargs)
+        elif start_finish is "erred_forgotten":  # noqa: F632
+            return self.transition_released_forgotten(*args, **kwargs)
+
+    def _transition(
+        self,
+        key,
```
martindurant, line 2005: Doesn't `key` have a type?

distributed/scheduler.py
```diff
@@ -2213,7 +2256,9 @@ def transition_released_waiting(self, key):
                 pdb.set_trace()
             raise
 
-    def transition_no_worker_waiting(self, key):
+    @ccall
```
martindurant, line 2259: why all `ccall` rather than `cfunc` - can they be called from anywhere but `_transition`?

distributed/scheduler.py
```diff
@@ -2423,27 +2472,28 @@ def transition_waiting_memory(
...
         recommendations: dict = {}
         client_msgs: dict = {}
         worker_msgs: dict = {}
         try:
             ts: TaskState = self._tasks[key]
-            assert worker
-            assert isinstance(worker, str)
 
             if self._validate:
+                assert worker
```
martindurant, line 2495: good move

martindurant (commented):
Forgive some naive questions here.

From a new perspective, I think this is largely good. I wonder if we can prove that the string concatenation is actually faster.

PR #4651, opened 2021-03-30
Split RAM measure into dask keys/other old/other new

### LINKED ISSUE #4634: Visualize split between dask keys and memory leaks/other
I'm working on a change to the GUI that will show, for each worker, how much of the worker's RSS is being allocated to dask keys (as reported by the sum of the sizeof() outputs) vs. everything else.
I also intend to split the "everything else" into "old" and "new" memory, defined as follows:

- "dask keys": WorkerState.nbytes, in other words the sum of the output of sizeof() for each dask key stored (not counting those spilled to disk)
- "other old": minimum across the last 30 seconds (configurable) of the measures of (RSS - dask keys)
- "other new": latest measure of RSS - dask keys - "other old"

The idea is that this should let us smooth out temporary peaks caused by delays in gc, delays in the python memory manager releasing the RAM to the OS, and memory temporarily allocated by the task functions.

I intend to use this more stable measure of (dask keys + other old) in all future non-time-critical heuristics that require to measure the total RAM usage of a worker - namely, the larger rebalance() rewrite I'm busy with.

In the GUI, the top-left graph would change as follows:

mockup

At the moment, the bars are fully blue and represent the whole RSS.
After the change, for each worker you'd have 3 stacked bars (blue: keys, dark grey: other old, light grey: other new) which add up to the RSS.

The hover tooltip for the individual workers will also change to match the top textbox.
The bar currently changes color to yellow and red to alert of high memory usage; I'd change it to a yellow/red box around it or something similar.

In the "workers" tab, I plan to add 3 new columns next to "memory" to break it down.

XREF https://github.com/dask/distributed/pull/4614 for the implementation of a drop-down to opt in/out of the extra columns to avoid making the table too crowded.

CC @jacobtomlinson, @jsignell , @jrbourbeau for opinions.

Note that all this excludes the dask keys currently spilled to disk. I'm considering adding a fourth bar on top for them.

crusaderky:
Closes #4634

DONE: backend
DONE: frontend
DONE: backend unit tests
DONE: frontend unit tests
DONE: stress testing
Out of scope (left to later PRs): use these measures in heuristics
Out of scope (left to later PRs): frontend design for >= 50 workers

**Note:** please switch to incognito mode or wipe your cache before testing this PR. If the dashboard from the main branch is in your browser's cache, bokeh will misbehave.


distributed/scheduler.py
```diff
@@ -361,6 +447,8 @@ class WorkerState:
     _last_seen: double
     _local_directory: str
     _memory_limit: Py_ssize_t
+    _memory_other_history: "deque[tuple[float, Py_ssize_t]]"
```
crusaderky, line 450: sadly, Cython doesn't seem to support ``from __future__ import annotations`` yet; it would have made the quotes unnecessary.

fjetter:
One component which takes up a surprising amount of memory is various metadata on the scheduler/worker. in particular transition and event logs. 
These are usually configurable but it is not transparent to users that this is a source of trouble, e.g. https://github.com/dask/distributed/pull/4604, https://github.com/dask/distributed/pull/2394

With this implementation, this should show up in `unmanaged_old` but would result in this to grow over time. I'm wondering if tracking the metadata memory usage delicately would be worth the effort (unfortunately `sizeof(SchedulerInstance)` is not possible without further engineering due to recursion errors).
This would help from a UX perspective and also reduce maintenance effort if people start reporting "memory leaks".

If this sounds like a good idea, I can open a dedicated issue, and we can implement this later, of course.

distributed/scheduler.py
```diff
@@ -256,6 +260,125 @@ def versions(self):
...
+        - global variables
+        - memory temporarily allocated by the dask tasks that are currently running
+        - memory fragmentation
+        - memory leaks
+        - memory not yet garbage collected
+        - memory not yet free()'d by the Python memory manager to the OS
+
+    unmanaged_old
+        Minimum of the "other" memory measures over the last
+        ``distributed.memory.new_to_old_lag`` seconds
```
fjetter, line 290: What's the motivation behind tracking the minimum? Wouldn't a max/mean/median be more informative for scheduling decisions?

crusaderky: The idea is that, short of memory leaks, most of the difference between the rolling minimum and the current value ("unmanaged recent") should disappear very soon, as long as your new_to_old_lag is configured to be longer than the runtime of a task and long enough to let gc run

crusaderky:
> One component which takes up a surprising amount of memory is various metadata on the scheduler/worker. in particular transition and event logs.
These are usually configurable but it is not transparent to users that this is a source of trouble, e.g. #4604, #2394

Under the default config, does this eventually plateau or end up balooning towards infinity? If it plateaus, how much are we talking about, and how long does it typically take to get there?

crusaderky:
Status update:

- GUI prototype for less than 50 workers finished and working as intended.
- FIXME: widget alignment for >= 50 workers is improvable
- FIXME: tooltips can get "too" transparent and be hard to read, or end up covered by the top navbar
- Known issue (not a regression): https://github.com/dask/distributed/issues/4675
- Known issue (not a regression): https://github.com/dask/distributed/issues/4676

fjetter:
> Under the default config, does this eventually plateau or end up balooning towards infinity? If it plateaus, how much are we talking about, and how long does it typically take to get there?

The default configuration atm is too high. In the case of #4604 we investigated that, for instance, the heartbeat event takes up already 3KB per event, per worker. With the addition of more information in this PR this may become larger. The default values for these deques are at 100k events, so naively we're at a threshold of 300MB for the event log. Probably similar for the transition log, etc.
It is also an option to simply reduce this default since I believe these logs are not super informative for the default user

distributed/scheduler.py
fjetter: makes sense. If this assumption is important, we might want to think about something like an automated warning if average runtime of a task is much larger than the configured value. If this is a recurring issue, we could, of course also define some heuristic, similar to the hearbeat interval.
My experience with task runtimes is simply that they can be very large, especially for custom graphs (think 10s of minutes) and finding a suitable value for everyone is probably quite difficult.

crusaderky:
@fjetter sounds like you're making a good point about an iterative improvement after this PR that shows this kind of memory usage. However I'm worried about the CPU consumption of massively running sizeof() on what sounds like it's very fine-grained pure-python data structures?

distributed/scheduler.py
crusaderky: We could configure it to "since the oldest task started running +X seconds", but it would not work for seceded tasks and for users that have genuinely long-running tasks that take very little RAM mixed with short bursts of memory-hungry ones.

fjetter:
Rethinking this, these logs are primarily important on the scheduler and this PR focuses a lot on the workers. I believe on the workers these logs are currently still under control so I suggest to ignore my comment for the moment :)

> However I'm worried about the CPU consumption of massively running sizeof() on what sounds like it's very fine-grained pure-python data structures?

Well, the `sizeof` of lists is pretty fast because it simply samples 10 elements of the list and extrapolates to the list length. We do not have a dedicated treatment for `deque`s but that should be similarly simple. Therefore, I'm not too worried about the CPU consumption. This could, for instance, be measured in the `memory_monitor` callback and I would expect this to be much faster than the proc system call we're doing there

distributed/http/static/css/status.css
```diff
@@ -6,45 +6,53 @@
 @media (min-width: 0px) {
   #status-fluid {
     grid-template-columns: 1fr 1fr;
-    grid-template-rows: 1fr 3fr 1fr;
+    grid-template-rows: 1fr 2fr 6fr 2fr;
```
crusaderky, line 9: FIXME the new cluster-wide RAM usage widget gets completely invisible if the browser window is too short. Need to find a way to set a minimum absolute height.

distributed/http/static/css/status.css
```diff
@@ -6,45 +6,53 @@
...
-    grid-row: 3;
+    grid-row: 4;
   }
 }
 
 @media (min-width: 992px) {
   #status-fluid {
     grid-template-columns: 1fr 3fr;
-    grid-template-rows: 1fr 1fr 1fr 1fr 1fr 1fr;
+    grid-template-rows: 1fr 4fr 1fr 1fr 4fr;
```
crusaderky, line 36: FIXME the new cluster-wide RAM usage widget gets completely invisible if the browser window is too short. Need to find a way to set a minimum absolute height.

quasiben:
Some screenshots of the work -- lighter shading is the "unmanaged" memory

Screen Shot 2021-04-06 at 11 50 50 AM

Screen Shot 2021-04-06 at 11 53 19 AM

Screen Shot 2021-04-06 at 11 56 18 AM

distributed/http/static/css/status.css
mrocklin: cc @ian-r-rose

crusaderky:
I think I'm satisfied with the state of both backend and frontend for this PR.
I left the graphs for >=50 workers unaltered for now.
Moving to testing.

jrbourbeau:
I just bumped CI as I think the previous builds here were being impacted some unrelated CI failures

jrbourbeau (commented):
Thanks for all your work on this @crusaderky. I'm looking forward to seeing this merged.

distributed/distributed.yaml
```diff
@@ -83,9 +83,15 @@ distributed:
...
                             # Requires https://github.com/numba/stacktrace
 
-    # Fractions of worker memory at which we take action to avoid memory blowup
-    # Set any of the lower three values to False to turn off the behavior entirely
     memory:
+      # When there is an increase in process memory (as observed by the operating
+      # system) that is not accounted for by the dask keys stored on the worker, ignore
+      # it for this long before considering it in non-critical memory measures.
+      # This should be set to be longer than the duration of most dask tasks.
+      new_to_old_lag: 30s
```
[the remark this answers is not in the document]
crusaderky: how about recent_to_old_time?

distributed/dashboard/components/scheduler.py
```diff
@@ -222,8 +218,239 @@ def update(self):
...
+            self.root.xaxis.minor_tick_line_alpha = 0
+            self.root.yaxis.visible = False
+            self.root.ygrid.visible = False
+
+            self.root.toolbar.logo = None
+            self.root.toolbar_location = None
+            self.root.yaxis.visible = False
+
+            hover = HoverTool()
+            hover.tooltips = "@proc_memory{0.00 b} (@width{0.00 b} @memtype)"
```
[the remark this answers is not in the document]
crusaderky: done

distributed/dashboard/components/scheduler.py
```diff
@@ -222,8 +218,239 @@ def update(self):
...
+            hover.point_policy = "follow_mouse"
+            self.root.add_tools(hover)
+
+    @without_property_validation
+    def update(self):
+        with log_errors():
+            workers = self.scheduler.workers.values()
```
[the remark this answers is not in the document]
crusaderky: done

distributed/dashboard/components/scheduler.py
```diff
@@ -222,8 +218,239 @@ def update(self):
...
+            limit = sum(getattr(ws, "memory_limit", 0) for ws in workers)
+            meminfo = self.scheduler.memory
+            if limit and meminfo.process > limit:
+                color = "red"
+            elif limit and meminfo.process > limit / 2:
+                color = "orange"
+            else:
+                color = "blue"
```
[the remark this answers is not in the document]
crusaderky: done

distributed/dashboard/components/scheduler.py
```diff
@@ -222,8 +218,239 @@ def update(self):
...
+            # FIXME https://github.com/dask/distributed/issues/4675
+            #       This causes flickering after adding workers and when enough memory
+            #       is spilled out
+            self.root.x_range.end = max(
+                limit, meminfo.process + meminfo.managed_spilled
+            )
+
+            title = f"Bytes stored: {format_bytes(meminfo.process)}"
+            if meminfo.managed_spilled:
+                title += f" + {format_bytes(meminfo.managed_spilled)} spilled"
```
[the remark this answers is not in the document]
crusaderky: changed to "spilled to disk"

distributed/scheduler.py
```diff
@@ -3672,6 +3794,29 @@ def heartbeat_worker(
 
         ws._metrics = metrics
 
+        # Calculate RSS - dask keys, separating "old" and "new" usage
+        # See MemoryInfo for details
+        max_memory_unmanaged_old_hist_age = local_now - MEMORY_NEW_TO_OLD_LAG
+        memory_unmanaged_old = ws._memory_unmanaged_old
+        while ws._memory_other_history:
+            ts, size = ws._memory_other_history[0]
```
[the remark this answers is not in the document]
crusaderky: fixed

distributed/scheduler.py
```diff
@@ -256,6 +260,105 @@ def versions(self):
         return self._versions
 
 
+class MemoryInfo:
```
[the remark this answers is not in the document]
crusaderky: works for me

distributed/scheduler.py
```diff
@@ -256,6 +260,105 @@ def versions(self):
...
+        return self.process - self.managed_in_memory
+
+    @property
+    def unmanaged_recent(self) -> int:
+        # This is never negative thanks to __init__
+        return self.process - self.managed_in_memory - self.unmanaged_old
+
+    @property
+    def optimistic(self) -> int:
+        return self.managed_in_memory + self.unmanaged_old
```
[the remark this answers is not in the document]
crusaderky: it will be used by rebalance() and all other heuristics

distributed/worker.py
```diff
@@ -609,25 +609,16 @@ def __init__(
...
-                File(path),
+                or sys.maxsize,
             )
-            target = (
-                int(float(self.memory_limit) * self.memory_target_fraction)
-                or sys.maxsize
-            )
-            self.data = Buffer({}, storage, target, weight)
-            self.data.memory = self.data.fast
-            self.data.disk = self.data.slow
```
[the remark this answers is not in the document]
crusaderky: reinstated as properties of SpillBuffer

crusaderky:
For failures on Python 3.9 / Windows, see https://github.com/dask/distributed/pull/4713

crusaderky:
@jrbourbeau Ready for final review and merge.
Ran all tests twice to verify stability. For failures on Python 3.9 / Windows, see #4713

distributed/http/static/css/status.css
jacobtomlinson: I am seeing this with the dashboard full screen on a 15" MacBook Pro. I expect this will affect a lot of users.

image

distributed/http/static/css/status.css
jacobtomlinson:
```suggestion
    grid-template-rows: 2fr 2fr 6fr 2fr;
```

distributed/http/static/css/status.css
jacobtomlinson:
```suggestion
    grid-template-rows: 2fr 4fr 1fr 1fr 4fr;
```

jacobtomlinson (commented):
Changing the height of the first row makes things look ok on my laptop and also on an iPad size display.

I think with boked we are always going to lose the bars if the graph is too short. But this at least makes the graphs more comparable in size.

distributed/http/static/css/status.css
jacobtomlinson: image

jacobtomlinson:
Ah I don't have one of those to test on.

But you could add another media query to handle that case and set it as `1rf` on 4k displays.

```css
@media (min-width: 2400px) {
  #status-fluid {
    grid-template-rows: 1fr 4fr 1fr 1fr 4fr;
  }
}
```

crusaderky:
@jacobtomlinson see how you like it now

jacobtomlinson:
I see no change with that. I wonder if this is a scaling thing. High density displays are scaled at 2x. Setting those numbers to `100px` and `120px` displays nicely on my screen.

crusaderky:
I don't get it - this is how 1920x1080 looks like in my screen with 60px (using chrome dev tools):

Screenshot from 2021-04-19 16-41-06

is this a problem of you having a much larger font?

crusaderky:
@jacobtomlinson try now...

jacobtomlinson:
No that also didn't do anything.

I notice looking at yours that the plots don't have any axis labels. Can you try adding some workers and checking it again?

I am just testing with `python -c 'from dask.distributed import LocalCluster, Client; client = Client(); from time import sleep; sleep(3600);'`.

crusaderky:
retested height with the axis labels

jacobtomlinson:
Perfect!

image

jacobtomlinson (approved):
Thanks for iterating with me on the status page @crusaderky.

Everything else looks good to me.

jacobtomlinson:
Let's get this in. It looks like all feedback has been addressed and it's been open for a while now. If there is anything outstanding we can tweak it in follow-up PRs.

Thanks for your efforts here @crusaderky!

mrocklin:
Woo!

jrbourbeau:
It looks like `test_memory` is failing consistently both here and in the `main` branch

crusaderky:
I'm onto it. It *WAS* working. :$

crusaderky:
Fix in #4719

distributed/scheduler.py
jakirkham: I think it will in Cython 3 ( https://github.com/cython/cython/pull/3829 ). That's probably still a ways off though

PR #4660, opened 2021-04-01
Remove incorrect comment regarding default LocalCluster creation

cameron16:
This PR resolves an erroneous comment describing the default behavior of the LocalCluster() class.

When the caller does not pass arguments to the LocalCluster() class, the number of workers created does not necessarily equal the number of cores. Specifically, if the machine has more than 4 cores, then the number of workers may not equal the number of cores.

The `nprocesses_nthreads` function determines the number of workers and number of threads, as shown here: https://github.com/dask/distributed/blob/main/distributed/deploy/local.py#L180

If the number of cores is greater than 4, then the `nprocesses_nthreads` may not set the number of cores equal to the number of workers. See the implementation of the `nprocesses_nthreads` function here: https://github.com/dask/distributed/blob/main/distributed/deploy/utils.py#L30

### DIFF
--- a/distributed/deploy/local.py
+++ b/distributed/deploy/local.py
@@ -78,7 +78,7 @@ class LocalCluster(SpecCluster):
 
     Examples
     --------
-    >>> cluster = LocalCluster()  # Create a local cluster with as many workers as cores  # doctest: +SKIP
+    >>> cluster = LocalCluster()  # Create a local cluster  # doctest: +SKIP
     >>> cluster  # doctest: +SKIP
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):
Thanks @cameron16! It looks like we forgot to update this comment when we added `nprocesses_nthreads`

cameron16:
@jrbourbeau Thanks for the quick review! I also just opened a PR in the dask repo to update a similar line in the Dask documentation: https://github.com/dask/dask/pull/7497

PR #4670, opened 2021-04-02
Add code coverage

jrbourbeau:
As title. xref https://github.com/dask/community/issues/146


hristog (commented):
Perhaps, this could benefit from a Coverage badge, as the one in the [dask repo](https://github.com/dask/dask/blame/main/README.rst#L18-L20).

jrbourbeau:
Yep, agreed we should add a badge in the README. There seem to be some issues with the Windows CI, so I'm going to focus on those first.

jakirkham:
It should be possible to [collect coverage of Cython code]( https://cython.readthedocs.io/en/latest/src/tutorial/profiling_tutorial.html#enabling-coverage-analysis ). Have not had a lot of luck with it personally. So am not sure it is worth doing initially, but wanted to note this in case we are wanting to exploring that option

jrbourbeau:
That's a fair point -- good to know. I agree it's probably okay to ignore for now and add in later in a follow-up PR if needed. I think this PR should be good to go. If you or @fjetter get a chance to review, that would be welcome : )

fjetter (approved):
Thanks for looking into this again. Last time I checked the tests got severely slowed down by coverage report.

This is something we should keep in mind going forward but for now I think coverage is important enough to suffer performance hits. If this becomes a problem we might want to consider to not do it for osx builds

jacobtomlinson (approved):
I agree this is really nice to have.

jakirkham (approved):

jakirkham:
Thanks James! 😄 Let's get this in :shipit:

PR #4671, opened 2021-04-02
Add timing information to TaskGroup

mrocklin:
I would like for us to start thinking about visualizing TaskGroups as an alternative to the task stream.

This adds timeline information to those objects as a precursor to that work.

Example
-------

```python
client = Client()
client

import dask.array as da
x = da.random.random((10000, 10000))
y = x + x.T - x.mean(axis=0)
y = y.persist()
wait(y)

groups = client.cluster.scheduler.task_groups
tg = list(groups.values())[0]
tg.__dict__
```

```python
{'_name': 'sub-2005fd772a2da340e72f1de1df1ac06d',
 '_prefix': <sub: memory: 16>,
 '_states': {'waiting': 0,
  'erred': 0,
  'released': 0,
  'memory': 16,
  'no-worker': 0,
  'processing': 0,
  'forgotten': 0},
 '_dependencies': {<mean_agg-aggregate-3f80179c8a4d92eabf1c0046dbbabd74: released: 4>,
  <random_sample-87d33bde1d8da2f2c8edc106b69514a0: released: 16>},
 '_nbytes_total': 800000000,
 '_nbytes_in_memory': 800000000,
 '_duration': 4.233909368515015,
 '_types': {'numpy.ndarray'},
 '_start': 1617389525.5363212,
 '_stop': 1617389526.3259144,
 '_all_durations': defaultdict(float,
             {'compute': 4.233909368515015, 'transfer': 1.8818869590759277})}
```

This should be enough to give us information for a rich visual

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -846,6 +846,18 @@ class TaskGroup:
     def types(self):
         return self._types
 
+    @property
+    def all_durations(self):
+        return self._all_durations
+
+    @property
+    def start(self):
+        return self._start
+
+    @property
+    def stop(self):
+        return self._stop
+
     @ccall
     def add(self, o):
         ts: TaskState = o
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -796,6 +796,9 @@ class TaskGroup:
     _nbytes_in_memory: Py_ssize_t
     _duration: double
     _types: set
+    _start: double
+    _stop: double
+    _all_durations: object
 
     def __init__(self, name: str):
         self._name = name
@@ -807,6 +810,9 @@ class TaskGroup:
         self._nbytes_in_memory = 0
         self._duration = 0
         self._types = set()
+        self._start = 0.0
+        self._stop = 0.0
+        self._all_durations = defaultdict(float)
 
     @property
     def name(self):
@@ -2305,6 +2311,7 @@ class SchedulerState:
                     # record timings of all actions -- a cheaper way of
                     # getting timing info compared with get_task_stream()
                     ts._prefix._all_durations[action] += stop - start
+                    ts._group._all_durations[action] += stop - start
 
             #############################
             # Update Timing Information #
@@ -2321,6 +2328,9 @@ class SchedulerState:
 
                 ts._prefix._duration_average = avg_duration
                 ts._group._duration += new_duration
+                ts._group._start = ts._group._start or compute_start
+                if ts._group._stop < compute_stop:
+                    ts._group._stop = compute_stop
 
                 s: set = self._unknown_durations.pop(ts._prefix._name, None)
                 tts: TaskState
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1830,7 +1830,23 @@ async def test_no_danglng_asyncio_tasks(cleanup):
     assert tasks == start
 
 
-@gen_cluster(client=True)
+class NoSchedulerDelayWorker(Worker):
+    """Custom worker class which does not update `scheduler_delay`.
+
+    This worker class is useful for some tests which make time
+    comparisons using times reported from workers.
+    """
+
+    @property
+    def scheduler_delay(self):
+        return 0
+
+    @scheduler_delay.setter
+    def scheduler_delay(self, value):
+        pass
+
+
+@gen_cluster(client=True, Worker=NoSchedulerDelayWorker)
 async def test_task_groups(c, s, a, b):
     start = time()
     da = pytest.importorskip("dask.array")
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1832,10 +1832,12 @@ async def test_no_danglng_asyncio_tasks(cleanup):
 
 @gen_cluster(client=True)
 async def test_task_groups(c, s, a, b):
+    start = time()
     da = pytest.importorskip("dask.array")
     x = da.arange(100, chunks=(20,))
     y = (x + 1).persist(optimize_graph=False)
     y = await y
+    stop = time()
 
     tg = s.task_groups[x.name]
     tp = s.task_prefixes["arange"]
@@ -1870,6 +1872,9 @@ async def test_task_groups(c, s, a, b):
     assert tg.states["forgotten"] == 5
     # Ensure TaskGroup is removed once all tasks are in forgotten state
     assert tg.name not in s.task_groups
+    assert tg._start > start
+    assert tg._stop < stop
+    assert "compute" in tg._all_durations
     assert sys.getrefcount(tg) == 2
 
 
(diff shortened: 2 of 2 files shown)


mrocklin:
I was playing around a little bit with visualizing task groups.  Adding the timing information from this PR really helps.  However placement remains an open question.  I may not get back to this work for a while.  Dumping what I have up here for now.

https://gist.github.com/2f37440e52420b0b2899f72c6f61b802

mrocklin:
https://nbviewer.jupyter.org/gist/mrocklin/2f37440e52420b0b2899f72c6f61b802

distributed/tests/test_scheduler.py
```diff
@@ -1870,6 +1872,9 @@ async def test_task_groups(c, s, a, b):
     assert tg.states["forgotten"] == 5
     # Ensure TaskGroup is removed once all tasks are in forgotten state
     assert tg.name not in s.task_groups
+    assert tg._start > start
+    assert tg._stop < stop
```
fjetter, line 1876: This fails in CI here since `stop` is a few ms shorter than `tg._stop`. The startstops of the compute are using the clock of the workers. I assume this is somehow connected. Although I would think that the clock for all workers, scheduler and test are the same considering they share the same loop

jrbourbeau: It looks like the start/stop times the worker reports are augmented by the the worker's `scheduler_delay` estimate

https://github.com/dask/distributed/blob/843c704d0bc2f3fe576ed5b7411ca7da2c4162a4/distributed/worker.py#L3625-L3626

which might be why things are a few milliseconds off.

One approach would be to include this scheduler delay in the `start` / `stop` times used in `test_task_groups`:

```diff
diff --git a/distributed/tests/test_scheduler.py b/distributed/tests/test_scheduler.py
index 38621770..5bfedefe 100644
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1830,14 +1830,15 @@ async def test_no_danglng_asyncio_tasks(cleanup):
     assert tasks == start


-@gen_cluster(client=True)
-async def test_task_groups(c, s, a, b):
-    start = time()
+
+@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)])
+async def test_task_groups(c, s, a):
     da = pytest.importorskip("dask.array")
+    start = time() + a.scheduler_delay
     x = da.arange(100, chunks=(20,))
     y = (x + 1).persist(optimize_graph=False)
     y = await y
-    stop = time()
+    stop = time() + a.scheduler_delay

     tg = s.task_groups[x.name]
     tp = s.task_prefixes["arange"]
```

While this gets the test passing for me locally, I'll admit it doesn't feel like a great solution as the `Worker.scheduler_delay` is periodically updated during the worker's heartbeat

fjetter: That makes much more sense than differing clocks :)

Once could instead force the value to be zero, e.g.

```diff
diff --git a/distributed/tests/test_scheduler.py b/distributed/tests/test_scheduler.py
index 38621770..f21bde7a 100644
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -1829,8 +1829,17 @@ async def test_no_danglng_asyncio_tasks(cleanup):
     tasks = asyncio.all_tasks()
     assert tasks == start

+class NoDelayWorker(Worker):
+    @property
+    def scheduler_delay(self):
+        return 0

-@gen_cluster(client=True)
+    @scheduler_delay.setter
+    def scheduler_delay(self, value):
+        pass
+
+
+@gen_cluster(client=True, Worker=NoDelayWorker)
 async def test_task_groups(c, s, a, b):
     start = time()
     da = pytest.importorskip("dask.array")
```

distributed/scheduler.py
```diff
@@ -796,6 +796,9 @@ class TaskGroup:
     _nbytes_in_memory: Py_ssize_t
     _duration: double
     _types: set
+    _start: double
```
jrbourbeau, line 799: @jakirkham just wanted to note that we're not adding corresponding `property`s here like 

```python
@property
def start(self):
    return self._start
```

I think this should be okay as IIUC these `property`s were added so we could gradually cythonize the scheduler's internals. As `TaskGroup._start`, `._stop`, and `._all_durations` are new attributes, we don't need to worry about being backwards compatible in the same way. Please feel free to correct me if that's not the case

jakirkham: Unfortunately that could result in breakage when Cythonizing. `@property` is needed for pure Python (non-typed) access of attributes

distributed/scheduler.py
```diff
@@ -807,6 +810,9 @@ def __init__(self, name: str):
         self._nbytes_in_memory = 0
         self._duration = 0
         self._types = set()
+        self._start = 0.0
+        self._stop = 0.0
+        self._all_durations = defaultdict(float)
```
jakirkham, line 815: Would it be possible to not use `defaultdict`? When using `dict`, we can optimize calls to the object when Cythonizing

distributed/tests/test_scheduler.py
```diff
@@ -1870,6 +1888,9 @@ async def test_task_groups(c, s, a, b):
...
+    assert "compute" in tg._all_durations
```
jakirkham, line 1893: @jrbourbeau here are cases where `@property` is needed. Cythonization of the Scheduler would result in this test failing as-is

jrbourbeau: It looks like tests passed for the [CI build which cythonizes the scheduler](https://github.com/dask/distributed/pull/4671/checks?check_run_id=2280844688). Just to double check, should we expect this `assert` to fail, or just result in less-than-optimal cythonization?

jakirkham: I'm not sure the CI is working correctly. Had tried to get it to work on Windows, but ran into issues. That was back when only GitHub Actions covered Windows. Have not revisited since getting other OSes got added to GitHub Actions

distributed/scheduler.py
mrocklin: We could, but it inceases code complexity.  I'm inclined to wait until we see that this is clearly slowing things down before optimizing away from code complexity if that's ok.  I'm enthusiastic to see the C API PR get in.

distributed/scheduler.py
mrocklin: I've added properties

distributed/scheduler.py
mrocklin: If we do find that this is important we can maybe replace this with a class or namedtuple (maybe)

jakirkham: Yeah this is more important with objects seeing a lot of use. Given the limited use of this attribute, this is probably ok for now

jakirkham (approved):

jrbourbeau (approved):
Thanks @mrocklin for adding this and @jakirkham for reviewing! This is in

PR #4672, opened 2021-04-02
Avoid creating IOLoop at import time

### LINKED ISSUE #4616: Import of distributed created an IOLoop
In `distributed.comm.tcp`:
```python
class BaseTCPConnector(Connector, RequireEncryptionMixin):
    _executor = ThreadPoolExecutor(2, thread_name_prefix="TCP-Executor")
    _resolver = netutil.ExecutorResolver(close_executor=False, executor=_executor)
    client = TCPClient(resolver=_resolver)
```
the `ExecutorResolver`, which is evaluated and instantiated at import time, creates a tornado IOLoop and whichever asyncio loop underneath, and sets these as the "default" for the current thread, usually the main thread. I don't immediately know the consequences of this, but it seems like a bad idea.

mrocklin:
Fixes #4616

@martindurant , does this resolve your issue?

### DIFF
--- a/distributed/comm/tcp.py
+++ b/distributed/comm/tcp.py
@@ -367,8 +367,24 @@ class RequireEncryptionMixin:
 
 class BaseTCPConnector(Connector, RequireEncryptionMixin):
     _executor = ThreadPoolExecutor(2, thread_name_prefix="TCP-Executor")
-    _resolver = netutil.ExecutorResolver(close_executor=False, executor=_executor)
-    client = TCPClient(resolver=_resolver)
+
+    @property
+    def _resolver(self):
+        try:
+            return self._cached_resolver
+        except AttributeError:
+            self._cached_resolver = netutil.ExecutorResolver(
+                close_executor=False, executor=self._executor
+            )
+            return self._cached_resolver
+
+    @property
+    def client(self):
+        try:
+            return self._cached_client
+        except AttributeError:
+            self._cached_client = TCPClient(resolver=self._resolver)
+            return self._cached_client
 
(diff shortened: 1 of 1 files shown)


distributed/comm/tcp.py
```diff
@@ -367,8 +367,24 @@ def _check_encryption(self, address, connection_args):
 
 class BaseTCPConnector(Connector, RequireEncryptionMixin):
     _executor = ThreadPoolExecutor(2, thread_name_prefix="TCP-Executor")
-    _resolver = netutil.ExecutorResolver(close_executor=False, executor=_executor)
-    client = TCPClient(resolver=_resolver)
+
+    @property
```
martindurant, line 371: Wouldn't this set an instance attribute, so you get a new Resolver for every instance? Previously, by having it be on the class, there is only ever one for all instances of this class and descendants. You would need a [class property](https://stackoverflow.com/a/5191224) or descriptor for the same effect, or have class attribute `_cached_resolver = [None]`, above, as in other cases.

martindurant (commented):
Thanks for getting to this so quickly!

Yes, it would solve the creation on import. However, I don't think has the original intended effect any more.

Also, I don't know if there's any possible downside to the case where multiple objects try to get cached from multiple threads at once.

distributed/comm/tcp.py
martindurant: Or I guess just `BaseTCPConnector._cached_resolver = ` in the Except block

jcrist:
Superseded by #5593. Closing.

PR #4679, opened 2021-04-07
DOC: Attempt at autoreformatting docstrings to conform to Numpydoc.

Carreau:
I'm still working on my reformatter, but as distributed in one of the
project using numpydoc, I gave it a shot.

Most of the spacing around colons slightly change the semantic and allow
sphix/NumpyDoc to properly find the parameter names and type.

The one contentious question is whether *args, **kwargs should have */**
in front of them in docstrings (or not), and wether those should be
escaped.

Numpydoc reference does not say so but provide an example where stars
are present; and escaping seem unnecessary to render properly in sphinx.
Plus \*\* looks bad when using `?` in IPython.

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/cfexecutor.py
+++ b/distributed/cfexecutor.py
@@ -99,7 +99,7 @@ class ClientExecutor(cf.Executor):
         ----------
         fn : A callable that will take as many arguments as there are
             passed iterables.
-        iterables : One iterable for each parameter to *fn*.
+        *iterables : One iterable for each parameter to *fn*.
         timeout : The maximum number of seconds to wait. If None, then there
             is no limit on the wait time.
         chunksize : ignored.
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -1624,7 +1624,7 @@ class Client:
         Parameters
         ----------
         func : callable
-        iterables : Iterables
+        *iterables : Iterables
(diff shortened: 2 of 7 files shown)


fjetter:
LGTM. Do you intend to change more of it?

Test failure should be unrelated

Carreau:
> Do you intend to change more of it?

Not currently; that's all the fixes [vélin](https://github.com/Carreau/velin) was able to automatically apply.
I keep implementing better fixes, so when those are implemented I will likely run it against distributed again.

It did find other errors though, but I'm not sure yet how to auto-fix those:

Issues with parameter names:

```
distributed/scheduler.py:workers_to_close
  missing: {'memory_ratio', 'comm'}
  extra: {'memory_factor'}
```

So likely a typo ratio/factor, though the fact that `comm` is undocumented ment it was not able to auto-rename.

```
distributed/utils.py:seek_delimiter
  missing: {'blocksize', 'delimiter'}
  extra: {"'`` or message sentinel"}
```

Likely a indent or spacing issue around `:` that triggers Numpydoc to misparse and about [two dozen undocumented function arguments](https://gist.github.com/Carreau/c96e7b5e20a55c975f4747f809cde424).

PR #4682, opened 2021-04-07
Add support for extra conn args for HTTP protocols

marcosmoyano:
Support sending extra connection information to
websockets comms, ie: headers but can map to any
init argument of tornado.httpclient.HTTPRequest

This allows me to authenticate, identify and track calls to a running Dask cluster with websockets behind a proxy, from the proxy itself.

General thoughts: 
* Couldn't find a better place for this than the Security class. That said, it place nicely with `get_connection_args`. Curious to see what other people thing about this.
* We could do something similar for `get_listener_args` if we want to add extra args to `tornado.httpserver.HTTPServer` than `ssl_options`. Is that something worth exploring? Shouldn't be too hard

/cc @mrocklin @jacobtomlinson @jcrist 

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/comm/tests/test_ws.py
+++ b/distributed/comm/tests/test_ws.py
@@ -156,6 +156,29 @@ async def test_http_and_comm_server(cleanup, dashboard, protocol, security, port
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "protocol,security",
+    [
+        (
+            "ws://",
+            Security(extra_conn_args={"headers": {"Authorization": "Token abcd"}}),
+        ),
+        (
+            "wss://",
+            Security.temporary(
+                extra_conn_args={"headers": {"Authorization": "Token abcd"}}
+            ),
+        ),
+    ],
+)
+async def test_connection_made_with_extra_conn_args(cleanup, protocol, security):
+    async with Scheduler(protocol=protocol, security=security) as s:
+        connection_args = security.get_connection_args("worker")
+        comm = await connect(s.address, **connection_args)
+        assert comm.sock.request.headers.get("Authorization") == "Token abcd"
+
+
+@pytest.mark.asyncio
 async def test_quiet_close(cleanup):
     with warnings.catch_warnings(record=True) as record:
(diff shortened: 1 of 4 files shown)


mrocklin:
I don't have much to say about the code architecture here.  @jcrist is probably the best person for this, although he may be out for a few weeks.  @jacobtomlinson if you have time here that would be welcome.

I am curious though if there is a good way to test the ultimate desired behavior, which in this case I think is a secure connection using traditional web security techniques.  My guess is that it's hard to write a test for this without creating a full web auth stack.  I thought I would ask though in case people have any ideas.  For example, does Tornado have a very simple auth system that we could use on the Scheduler side and then connect with from the Client side?  Something like that might help with motivation here outside of services like Coiled.

jacobtomlinson (commented):
This seems fine to me.

I don't see anywhere in the changes that the header is being checked against anything. Is the intention that some proxy is checking this header?

distributed/comm/tests/test_ws.py
```diff
@@ -155,6 +155,29 @@ async def test_http_and_comm_server(cleanup, dashboard, protocol, security, port
...
+                extra_conn_args={"headers": {"Authorization": "Token abcd"}}
+            ),
+        ),
+    ],
+)
+async def test_connection_made_with_extra_conn_args(cleanup, protocol, security):
+    async with Scheduler(protocol=protocol, security=security) as s:
+        connection_args = security.get_connection_args("worker")
+        comm = await connect(s.address, **connection_args)
+        assert comm.sock.request.headers.get("Authorization") == "Token abcd"
```
mrocklin, line 178: There is a failing test in CI that is about some resource not being cleaned up.  It could be that this would be resolved by closing this comm

Maybe the following would work

```suggestion
        assert comm.sock.request.headers.get("Authorization") == "Token abcd"
        await comm.close()
```

mrocklin:
Given @jacobtomlinson 's approval I'm going to merge this tomorrow if there are no objections.

I would not be surprised if @jcrist shows up in a few weeks with requested changes to the API. I would encourage us to informally treat this API as experimental until then.

PR #4690, opened 2021-04-08
Fix timeout in client.restart

mdwint:
I noticed `client.restart()` would hang sometimes, and then I discovered the timeout passed to `asyncio.wait_for` being expressed as a "deadline" instead. Am I correct in assuming this is wrong?

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -2980,13 +2980,14 @@ class Client:
             timeout = self._timeout * 2
         if timeout is not None:
             timeout = parse_timedelta(timeout, "s")
+
         self._send_to_scheduler({"op": "restart", "timeout": timeout})
         self._restart_event = asyncio.Event()
         try:
             await asyncio.wait_for(self._restart_event.wait(), timeout)
         except TimeoutError:
             logger.error("Restart timed out after %.2f seconds", timeout)
-            pass
+
         self.generation += 1
         with self._refcount_lock:
             self.refcount.clear()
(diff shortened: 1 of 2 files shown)


jrbourbeau (commented):
Good catch @mdwint! You're correct that `timeout` in [`asyncio.wait_for`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for) is the number of seconds to wait.

Could you add a small test to `distributed/tests/test_client.py` to ensure that we're timing out as expected?

Also, I know it predates the changes in this PR, but could you add a call to `parse_timedelta` so `timeout=` can accept inputs like "10min", "13s" instead of just an integer or float number of seconds? For instance, here's an example of where we use `parse_timedelta` in another location in `client.py`:

https://github.com/dask/distributed/blob/be2d49569bbbba56377c6a7690d174d6175c5991/distributed/client.py#L605

mdwint:
Thanks for your guidance @jrbourbeau. I've added that test, as well as the call to `parse_timedelta`. Let me know if this is ok.

jrbourbeau (approved):
This looks great! Just a heads up, I pushed a small commit to use `"0.5s"` instead of `0.5` in the test to added to make sure the `parse_timedelta` addition is working as expected -- hope that's okay. Will merge once CI finishes

jrbourbeau:
Also I noticed this was your first code contribution to this repository. Welcome!

mdwint:
Thanks, @jrbourbeau!

PR #4695, opened 2021-04-12
Remove WSSConnector TLS presence check

marcosmoyano:
Let it fail at connection time if TLS is not properly configured

This allows me to authenticate against a proxy using other techniques like http headers but still use `wss`. The proxy has access to the set of TLS certs and can, based upon authentication, send them along.

/cc @mrocklin @jacobtomlinson 

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/comm/tests/test_ws.py
+++ b/distributed/comm/tests/test_ws.py
@@ -9,8 +9,8 @@ import dask
 
 from distributed import Client, Scheduler, Worker
 from distributed.comm import connect, listen, ws
-from distributed.comm.registry import backends, get_backend
 from distributed.comm.core import FatalCommClosedError
+from distributed.comm.registry import backends, get_backend
 from distributed.security import Security
 from distributed.utils_test import (  # noqa: F401
     cleanup,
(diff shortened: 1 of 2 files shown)


mrocklin:
This seems fine to me.  It's also isolated to the WS module, which Marcos knows best.  Merging after passed tests.

marcosmoyano:
@mrocklin tests are not passing but I don't think those connect back to my change 🤔

mrocklin:
@jrbourbeau something is up here with the tests.  Also, I find that running `pytest distributed/comms/tests/test_ws.py` fails for me due to leaked threads.  If you have a few minutes can I ask you to try to push this in, and maybe also see what is up with tests generally in this module?

mrocklin:
> @mrocklin tests are not passing but I don't think those connect back to my change thinking

Yeah, I'm not sure what's up here.  James may be able to help.

jrbourbeau:
Yeah, the test failures here look totally unrelated to the changes in this PR. I'm running CI against the current `main` branch over in https://github.com/dask/distributed/pull/4696 to see if there are test failures there too

mrocklin:
If that's the case are we good to merge here?

jrbourbeau:
It looks like those were just [transient test failures](https://github.com/dask/distributed/pull/4696#issuecomment-818388957) ([`test_dashboard` has become flaky recently](https://github.com/dask/distributed/issues/4697)). Merging, thanks @marcosmoyano!

mrocklin:
Woo!

PR #4696, opened 2021-04-12
Test CI

jrbourbeau:
As title


jrbourbeau:
For reference, I'm wanting to check whether the test failures we're seeing over in https://github.com/dask/distributed/pull/4695 are related to the changes in that PR, or if they're unrelated and in the current `main` branch. Although if tests fail here, since CI for the [latest commit to `main`](https://github.com/dask/distributed/commit/053f99b85a56a2a412f27b1950f65eb93b5bc0be) passed, it would mean some external change has caused things to start failing

crusaderky:
#4651 is also mass-failing with TimeoutError

jrbourbeau:
Alright, after bumping CI a few times here tests started to no longer fail for all CI builds. I compared the CI software environments between the failing and passing CI builds and got the following diff

```diff
16c16
<   - backports.functools_lru_cache=1.6.3=pyhd8ed1ab_0
---
>   - backports.functools_lru_cache=1.6.4=pyhd8ed1ab_0
60c60
<   - jupyter_client=6.1.12=pyhd8ed1ab_0
---
>   - jupyter_client=6.2.0=pyhd8ed1ab_0
73,74c73,74
<   - libgfortran=5.0.0=9_3_0_h6c81a4c_21
<   - libgfortran5=9.3.0=h6c81a4c_21
---
>   - libgfortran=5.0.0=9_3_0_h6c81a4c_22
>   - libgfortran5=9.3.0=h6c81a4c_22
76c76
<   - libnetcdf=4.7.4=nompi_h9d8a93f_107
---
>   - libnetcdf=4.8.0=nompi_h0bf7110_100
97c97
<   - netcdf4=1.5.6=nompi_py38h1f593c7_100
---
>   - netcdf4=1.5.6=nompi_py38h8ce633d_101
```

The only thing that stands out is `jupyter_client` which recently yanked a couple of release (6.1.13 and 6.2.0). It could be that we were negatively impacted by these releases, though since they've been yanked I don't think there's anything more for us to do here.

PR #4702, opened 2021-04-14
Simplify transition `func` retrieval for `finish` case

jakirkham:
The transition `func` that was being grabbed depended on whether `key` was found in the recommendations from the last transition. If `key` was found, one function based on `v` was called. If not, `finish` was used in place of `v`. To make this a bit more explicit, simple, and efficient, assign `finish` to `v` if `key` is not found. This way `v` can be used in all cases with one call to retrieve the `func`.

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1876,16 +1876,15 @@ class SchedulerState:
                 a: tuple = func(key, *args, **kwargs)
                 recommendations, client_msgs, worker_msgs = a
             elif "released" not in start_finish:
-                func = self._transitions_table["released", finish]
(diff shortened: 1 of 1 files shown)


jakirkham:
`test_dashboard` is failing, but is also known to be flaky ( https://github.com/dask/distributed/issues/4697 ). So probably not related

jakirkham:

jrbourbeau (approved):
Thanks @jakirkham!

jakirkham:
Thanks James! 😄

PR #4703, opened 2021-04-14
Maint faster tests_semaphore::test_worker_dies

fjetter:
This allows to pass config to the cluster dict

I'm using this to pass the lease-timeout to the worker dies test which shaves off 30s off our test runtime. Doesn't change  the world (and won't make things more stable) but it's a start

### DIFF
--- a/distributed/tests/test_semaphore.py
+++ b/distributed/tests/test_semaphore.py
@@ -140,7 +140,11 @@ async def test_async_ctx(s, a, b):
 
 @pytest.mark.slow
 def test_worker_dies():
-    with cluster() as (scheduler, workers):
+    with cluster(
+        config={
+            "distributed.scheduler.locks.lease-timeout": "0.1s",
+        }
+    ) as (scheduler, workers):
         with Client(scheduler["address"]) as client:
             sem = Semaphore(name="x", max_leases=1)
 
@@ -156,11 +160,11 @@ def test_worker_dies():
                     return x
 
             futures = client.map(
-                f, range(100), sem=sem, kill_address=workers[0]["address"]
+                f, range(10), sem=sem, kill_address=workers[0]["address"]
             )
             results = client.gather(futures)
 
-            assert sorted(results) == list(range(100))
+            assert sorted(results) == list(range(10))
 
 
 @gen_cluster(client=True)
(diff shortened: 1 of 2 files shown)


distributed/tests/test_semaphore.py
```diff
@@ -140,7 +140,11 @@ async def test_async_ctx(s, a, b):
 
 @pytest.mark.slow
 def test_worker_dies():
-    with cluster() as (scheduler, workers):
+    with cluster(
+        config={
+            "distributed.scheduler.locks.lease-timeout": "0.1s",
+        }
+    ) as (scheduler, workers):
```
jrbourbeau, line 147: I realize this predates this PR, but out of curiosity is there a reason to not use [`gen_cluster`](https://github.com/dask/distributed/blob/de6db936e0cbfcd7d71c4258fd72b11994ec37ed/distributed/utils_test.py#L829) here instead of the `cluster` context manager? Maybe it's something to do with `gen_cluster` and the `pytest.mark.slow` marker not working well together? 

I'm wondering because `gen_cluster` is used more broadly throughout the codebase and already has a `config=` keyword argument.

distributed/utils_test.py
```diff
@@ -595,6 +598,7 @@ def cluster(
     active_rpc_timeout=10,
     disconnect_timeout=20,
     scheduler_kwargs={},
+    config={},
```
jrbourbeau, line 601: We're probably okay for now, but in general we should probably avoid having mutable default values for keyword arguments

jrbourbeau (approved):
Thanks @fjetter! This is nice to see. I've left a couple of non-blocking comments

distributed/tests/test_semaphore.py
fjetter: Most (all?) APIs we can use for the async setup is a somehow graceful shutdown. In this case, we're actually terminating the process, removing any possibility for a final release or graceful disconnect. Closest thing to a real VM dying. 

> Maybe it's something to do with gen_cluster and the pytest.mark.slow marker not working well together?

I'm not aware of anything like this, it's just the async/sync thing

distributed/utils_test.py
fjetter: absolutely right. I usually do not do this but I guess the scheduler_kwargs a line above pulled me towards the dark side. we might want to integrate smth like https://pypi.org/project/flake8-mutable/ eventually

PR #4705, opened 2021-04-14
[NEVER MERGE] Initial experiment into adding chained recommendation to workers

gforsyth:
This currently deadlocks the test-suite, breaks work-stealing, and probably also breaks some other stuff.

Opening this mostly as a pointer to the various spots in the worker that may need to be touched / updated to work with the proposed state-machine that @fjetter proposed in https://github.com/dask/distributed/issues/4413#issuecomment-817820890

I'll add comments inline if I remember why a given dirty-hack is in place.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1401,6 +1401,7 @@ class Worker(ServerNode):
             ts = self.tasks.get(key)
             if ts is None:
                 self.tasks[key] = ts = TaskState(key)
+                ts.state = "flight"
             recommendations.update({ts: ("memory", {"value": value})})
 
             self.log.append((key, "receive-from-scatter"))
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1396,17 +1396,16 @@ class Worker(ServerNode):
     ###################
 
     def update_data(self, comm=None, data=None, report=True, serializers=None):
+        recommendations = {}
         for key, value in data.items():
             ts = self.tasks.get(key)
-            if getattr(ts, "state", None) is not None:
-                self.transition(ts, "memory", value=value)
-            else:
+            if ts is None:
                 self.tasks[key] = ts = TaskState(key)
-                self.put_key_in_memory(ts, value)
-                ts.priority = None
-                ts.duration = None
+            recommendations.update({ts: ("memory", {"value": value})})
 
             self.log.append((key, "receive-from-scatter"))
+
+        self.transitions(recommendations)
 
         if report:
             self.batched_stream.send({"op": "add-keys", "keys": list(data)})
@@ -1456,6 +1455,7 @@ class Worker(ServerNode):
         **kwargs2,
     ):
         try:
+            recommendations = {}
             runspec = SerializedTask(function, args, kwargs, task)
             if key in self.tasks:
                 ts = self.tasks[key]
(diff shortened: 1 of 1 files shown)


distributed/worker.py
```diff
@@ -1396,18 +1396,18 @@ async def get_data(
...
-            if getattr(ts, "state", None) is not None:
-                self.transition(ts, "memory", value=value)
-            else:
+            if ts is None:
                 self.tasks[key] = ts = TaskState(key)
-                self.put_key_in_memory(ts, value)
-                ts.priority = None
-                ts.duration = None
+                ts.state = "flight"
+            recommendations.update({ts: ("memory", {"value": value})})
```
gforsyth, line 1405: This hack was replacing the direct call to `put_key_in_memory` -- the `flight->memory` transition does the right thing, so I tagged it that way.  Gross.

distributed/worker.py
```diff
@@ -1475,17 +1476,13 @@ def add_task(
                     # This is a scheduler re-assignment
                     # Either `fetch` -> `waiting` or `flight` -> `waiting`
                     self.log.append((ts.key, "re-adding key, new TaskState"))
-                    self.transition(ts, "waiting", runspec=runspec)
+                    recommendations.update({ts: ("waiting", {"runspec": runspec})})
```
gforsyth, line 1479: This pattern is in a lot of places -- where previously we were passing through `kwargs` via `transition`, now they're bundled as a separate dictionary that gets passed in to `self.transition` by `self.transitions` while iterating over the recommended transitions.

distributed/worker.py
```diff
@@ -1519,11 +1515,15 @@ def add_task(
 
                     # transition from new -> fetch handles adding dependency
                     # to waiting_for_data
-                    self.transition(dep_ts, state)
+                    # TODO: `state` should be "fetch" and let recommender
+                    # make it "memory" if needed
+                    recommendations.update({dep_ts: state})
```
gforsyth, line 1520: The code that defines `state` isn't visible in my view here, but this current check is for handling tasks that get re-added, e.g. we've nuked the corresponding `TaskState` object but the _result_ still lives in `self.data`

This would go away (one hopes) if we stop removing `TaskState` instances.

distributed/worker.py
```diff
@@ -1569,26 +1570,52 @@ def add_task(
...
-        state = func(ts, **kwargs)
-        self.log.append((ts.key, start, state or finish))
-        ts.state = state or finish
-        if self.validate:
-            self.validate_task(ts)
-        self._notify_plugins("transition", ts.key, start, state or finish, **kwargs)
+        recommendations = func(ts, **kwargs)
+        self.log.append((ts.key, start, ts.state))
+        self._notify_plugins("transition", ts.key, start, ts.state, **kwargs)
+        return recommendations
```
gforsyth, line 1601: All changes to `TaskState.state` are handled inside of the respective transition functions now, instead of here.  Instead the individual transition functions return any recommendations to be passed back into the transition loop inside `self.transitions`

(I think the only place the state is mutated outside of a transition function is the dirty hack above in `update_data`)

jrbourbeau:
Thanks for the indicator on when this should be merged : )

distributed/worker.py
```diff
@@ -1569,26 +1570,52 @@ def add_task(
...
 
-    def transition_new_waiting(self, ts):
+    def transition_new_waiting(self, ts, runspec=None):
         try:
             if self.validate:
                 assert ts.state == "new"
-                assert ts.runspec is not None
+                # assert ts.runspec is not None
```
gforsyth, line 1607: This is no longer a valid check because the `runspec` is optionally set below via keyword argument

PR #4719, opened 2021-04-20
Fix regressions in #4651

crusaderky:
Fix regressions introduced in #4651:
- failing test
- if a worker sends a heartbeat before its SystemMonitor has a chance to run, metrics["memory"] will be None and that will cause a crash in the scheduler.


pentschev:
I confirmed locally this fixes the problem for me too, thanks so much for the quick fix @crusaderky !

crusaderky:
Stress test outcome:
- 76x successful
- 1x an unrelated test failed
- 22x CI fell over with no log whatsoever, very fast and definitely before reaching the test suite

distributed/scheduler.py
```diff
@@ -3833,7 +3835,12 @@ def heartbeat_worker(
             if size == memory_unmanaged_old:
                 memory_unmanaged_old = 0  # recalculate min()
 
-        size = max(0, metrics["memory"] - ws._nbytes + ws._metrics["spilled_nbytes"])
+        # metrics["memory"] is None if the worker sent a heartbeat before its
+        # SystemMonitor ever had a chance to run.
+        # ws._nbytes is updated at a different time and sizeof() may not be accurate,
+        # so size may be (temporarily) negative; floor it to zero.
+        size = max(0, (metrics["memory"] or 0) - ws._nbytes + metrics["spilled_nbytes"])
+
```
crusaderky, line 3843: I can't see a way to write a unit test for this short of spoofing Client.hearbeat?

pentschev: It seems like it's going to be the only way, another test does exactly that: https://github.com/dask/distributed/blob/e4b534af5722a7aebb8d5b086081984e93e1784f/distributed/tests/test_worker.py#L1697-L1717

pentschev:
Would people be ok with merging this even without a test and leaving a test for a different PR? This would unblock CI in Dask-CUDA and UCX-Py (potentially other RAPIDS projects too).

distributed/scheduler.py
fjetter: There is `distributed.admin.system-monitor.interval` which controls how often the monitor runs. You could set it to incredibly high values such that it is never executed during the test runtime

----

Another patch version w/ using monkeypatch, you could remove the PC before it is even started. Something like

```python
@pyters.mark.asyncio
async def test_foo():
    s = Scheduler()
    s.periodic_callbacks["monitor"] = None
    w = Worker(s)
    w.periodic_callbacks["monitor"] = None
    await s
    await w
    async with Client(s):
        ....

```

fjetter:
I'm fine with postponing tests since our CI is also failing hard. @crusaderky your call

crusaderky:
I need an extra 2-3 hours to cook a unit test. I'm ok merging without

distributed/tests/test_scheduler.py
```diff
@@ -2379,34 +2383,45 @@ def test_memory():
             ]
             sleep(2)
         assert_memory(s, "managed_spilled", 1, 999)
+        # Wait for the spilling to finish. Note that this does not make the test take
+        # longer as we're waiting for recent_to_old_time anyway.
+        sleep(10)
```
jrbourbeau, line 2388: Is there something more direct we can probe here instead of sleeping for 10 seconds?

jrbourbeau (approved):
Thanks for your work here @crusaderky. I'm going to merge this as is to unblock CI, but left a comment below

distributed/tests/test_scheduler.py
crusaderky: Not really because the unmanaged memory is very volatile, so we don't know how many keys are going to be spilled out exactly. Also, as noted it doesn't slow the test down.

distributed/scheduler.py
crusaderky: 1.
Setting ``distributed.admin.system-monitor.interval`` to a very high value, before I create the Scheduler has no effect.

2.
Setting 
```
s.periodic_callbacks["monitor"] = None
w.periodic_callbacks["monitor"] = None
```
fails with 
```
        for pc in self.periodic_callbacks.values():
>           pc.stop()
E           AttributeError: 'NoneType' object has no attribute 'stop'
```

3.
Setting
```
pc = PeriodicCallback(lamba: None, 999999999)
s.periodic_callbacks["monitor"] = pc
w.periodic_callbacks["monitor"] = pc
```
has no effect.

PR #4720, opened 2021-04-20
Ensure adaptive scaling is properly awaited and closed

fjetter:
* Primarily this ensures that scale_up/down is properly awaited in case the `sync` method is used which is more common than the `loop.add_callback`.
* It also ensures the adaptive PC is stopped properly once a cluster shuts down.
* A bit cleanup. There has been a test which raises error logs and I had issues figuring out what it actually was doing. It also was refactored a few times since its original inception.

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -5,6 +5,7 @@ import threading
 import uuid
 import warnings
 from contextlib import suppress
+from inspect import isawaitable
 
 from tornado.ioloop import PeriodicCallback
 
@@ -81,6 +82,9 @@ class Cluster:
     async def _close(self):
         if self.status == Status.closed:
             return
+
+        with suppress(AttributeError):
+            self._adaptive.stop()
 
         if self._watch_worker_status_comm:
             await self._watch_worker_status_comm.close()
@@ -419,7 +423,9 @@ class Cluster:
         return self
 
     async def __aexit__(self, typ, value, traceback):
-        await self.close()
+        f = self.close()
+        if isawaitable(f):
+            await f
 
     @property
     def scheduler_address(self):
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -6,6 +6,7 @@ import math
 import warnings
 import weakref
 from contextlib import suppress
+from inspect import isawaitable
 
 from tornado import gen
 
@@ -413,7 +414,15 @@ class SpecCluster(Cluster):
             return
         if self.status == Status.running or self.status == Status.failed:
             self.status = Status.closing
-            self.scale(0)
+
+            # Need to call stop here before we close all servers to avoid having
+            # dangling tasks in the ioloop
+            with suppress(AttributeError):
+                self._adaptive.stop()
+
+            f = self.scale(0)
+            if isawaitable(f):
+                await f
             await self._correct_state()
             for future in self._futures:
                 await future
--- a/distributed/deploy/tests/test_adaptive.py
+++ b/distributed/deploy/tests/test_adaptive.py
@@ -481,17 +481,11 @@ async def test_adaptive_stopped():
     stops.
     """
     async with LocalCluster(n_workers=0, asynchronous=True) as cluster:
-        async with Client(cluster, asynchronous=True) as client:
-            instance = cluster.adapt(interval="10ms")
+        instance = cluster.adapt(interval="10ms")
+        assert instance.periodic_callback is not None
 
-            await async_wait_for(
-                lambda: instance.periodic_callback is not None, timeout=5
-            )
+        await async_wait_for(lambda: instance.periodic_callback.is_running(), timeout=5)
 
-            await async_wait_for(
-                lambda: instance.periodic_callback.is_running() is not None, timeout=5
-            )
+        pc = instance.periodic_callback
 
-            pc = instance.periodic_callback
-
-    await async_wait_for(lambda: pc.is_running() is not None, timeout=5)
+    await async_wait_for(lambda: not pc.is_running(), timeout=5)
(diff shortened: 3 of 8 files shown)


distributed/tests/test_scheduler.py
```diff
@@ -713,7 +713,7 @@ async def test_scheduler_sees_memory_limits(s):
     await w.close()
 
 
-@gen_cluster(client=True, timeout=1000)
+@gen_cluster(client=True)
```
fjetter, line 716: This is an unrelated change which slipped in. This test usually runs a second. If we run into a timeout, I'd like to know since this may hint towards an issue

fjetter: If desired I'll remove this. I'll follow up with a few things related to worker closing/retiring later on and can include it there

distributed/deploy/tests/test_adaptive.py
```diff
@@ -19,43 +20,6 @@
 )
 
 
-@pytest.mark.asyncio
-async def test_simultaneous_scale_up_and_down(cleanup):
```
fjetter, line 23: This test was added in https://github.com/dask/distributed/pull/1608 to disallow simultaneous up-/down-scaling but the current implementation doesn't work at all since the API changed. Instead it only raises error logs

distributed/deploy/cluster.py
```diff
@@ -82,6 +83,8 @@ async def _close(self):
         if self.status == Status.closed:
             return
 
+        self._adaptive.stop()
```
jrbourbeau, line 86: Since `self._adaptive` is only set after `Cluster.adapt` is called, we might want to add a `with suppress(AttributeError):` here

fjetter:
Test failure is #4651 / #4719

distributed/deploy/tests/test_adaptive.py
```diff
@@ -460,3 +424,58 @@ async def test_update_adaptive(cleanup):
         await asyncio.sleep(0.2)
         assert first.periodic_callback is None
         assert second.periodic_callback.is_running()
+
+
+from distributed.utils_test import slowinc
```
jrbourbeau, line 429: This appears to be already imported at the top of this module

distributed/tests/test_scheduler.py
jrbourbeau: +1 for removing

distributed/deploy/tests/test_adaptive.py
```diff
@@ -460,3 +424,58 @@ async def test_update_adaptive(cleanup):
...
+            cluster.adapt()
+
+            await client.gather(futures)
+
+            del futures
+            await async_wait_for(lambda: not cluster.workers, 10)
+
+
+@pytest.mark.asyncio
+async def test_adaptive_stopped():
```
jrbourbeau, line 462: Given that this uses an async cluster, do we need the `async_wait_for` + timeout or can we just `assert` the various attributes directly? For example, when I make the following changes locally this test still passes:

```diff
diff --git a/distributed/deploy/tests/test_adaptive.py b/distributed/deploy/tests/test_adaptive.py
index 39284302..8cf4f381 100644
--- a/distributed/deploy/tests/test_adaptive.py
+++ b/distributed/deploy/tests/test_adaptive.py
@@ -466,16 +466,8 @@ async def test_adaptive_stopped():
     """
     async with LocalCluster(n_workers=0, asynchronous=True) as cluster:
         async with Client(cluster, asynchronous=True) as client:
-            instance = cluster.adapt(interval="10ms")
+            pc = cluster.adapt(interval="10ms").periodic_callback
+            assert pc is not None
+            assert pc.is_running() is not None

-            await async_wait_for(
-                lambda: instance.periodic_callback is not None, timeout=5
-            )
-
-            await async_wait_for(
-                lambda: instance.periodic_callback.is_running() is not None, timeout=5
-            )
-
-            pc = instance.periodic_callback
-
-    await async_wait_for(lambda: pc.is_running() is not None, timeout=5)
+    assert pc.is_running() is not None
```

jrbourbeau (commented):
Thanks @fjetter! I left a few small comments, but overall the changes here look good to me

distributed/deploy/tests/test_adaptive.py
fjetter: Well, at the very least we'll need to have one wait in between which actually waits for the entire thing to start. If the PC was never started, the conditions are trivially true. I'll add a comment and what I can remove for it to still work

distributed/deploy/tests/test_adaptive.py
fjetter: Wondering why me pre-commit didn't catch that... will fix of course

distributed/deploy/tests/test_adaptive.py
fjetter: All of the `is not None` checks where useless which is why the test did not fail. I properly assert for bool now and now I need to wait

distributed/deploy/spec.py
```diff
@@ -413,7 +414,15 @@ async def _close(self):
             return
         if self.status == Status.running or self.status == Status.failed:
             self.status = Status.closing
-            self.scale(0)
+
+            # Need to call stop here before we close all servers to avoid having
+            # dangling tasks in the ioloop
+            with suppress(AttributeError):
+                self._adaptive.stop()
```
fjetter, line 421: This is a bit unfortunate since I need to call it up here and not down in `Cluster`. If I do this only in cluster, the even loop seems to close too soon and we have still pending tasks from `AdaptiveCore.adapt`.

I'm wondering if we ever considered adding `PYTHONASYNCIODEBUG=1` to our test suite which would raise in these instances. Not sure how much would break or if this is a bad idea in general

PR #4713, opened 2021-04-19
ipython causes test_profile_nested_sizeof crash on windows

crusaderky:
test_profile_nested_sizeof has started crashing again on Windows 3.9.
Previous discussion at #4585

CC @jrbourbeau

PR #4722, opened 2021-04-20
Enable configuration of prometheus metrics namespace

jacobtomlinson:
This PR adds the ability to configure the prometheus metric namespace prefix.

All applications should prefix their metrics with a consistent name. In most of our metrics we prefix them with `dask_` but not always.

Adding a config option `distributed.dashboard.prometheus.namespace` which defaults to `dask` (the `_` is added automatically) but allows users to override this if they have a system-specific requirement for this.

Given there are three collector classes which have had this added it may make sense to base class these and consolidate duplicated code in the future.


distributed/http/scheduler/prometheus/__init__.py
```diff
@@ -9,24 +11,33 @@
 class _PrometheusCollector:
     def __init__(self, dask_server):
         self.server = dask_server
+        self.namespace = dask.config.get("distributed.dashboard.prometheus.namespace")
+        self.subsystem = "scheduler"
 
     def collect(self):
         from prometheus_client.core import CounterMetricFamily, GaugeMetricFamily
+        from prometheus_client.metrics import _build_full_name
```
fjetter, line 19: Is there any way this can be accessed via a public API call?

jacobtomlinson: Sadly no. I was also concerned about this.

Perhaps vendoring in the method would make sense here, it's pretty small and Apache 2.0 licensed.

distributed/http/scheduler/prometheus/semaphore.py
```diff
@@ -1,36 +1,52 @@
...
         from prometheus_client.core import CounterMetricFamily, GaugeMetricFamily
+        from prometheus_client.metrics import _build_full_name
 
         sem_ext = self.server.extensions["semaphores"]
 
         semaphore_max_leases_family = GaugeMetricFamily(
-            "semaphore_max_leases",
+            _build_full_name(
+                "max_leases", namespace=self.namespace, subsystem=self.subsystem
+            ),
```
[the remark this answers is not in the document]
jacobtomlinson: Yeah it is. Although I consider not having a prefix on these metrics to be a bug.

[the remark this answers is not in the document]
sriramch: is it possible to emit both metrics for some time, deprecating the existing one (without namespace/subsystem) before removing the old one completely?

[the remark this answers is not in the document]
jacobtomlinson: I suspect the number of users that track these metrics via prometheus are pretty small. There isn't really a good way to mark it as deprecated either, other than via release notes. So I think a break here isn't too big of an issue.

distributed/http/scheduler/prometheus/__init__.py
jacobtomlinson: AFAIK we could just copy the code into somewhere like `distributed.http.utils` and include a comment which links to the original source and Apache license on `prometheus_client`.

jacobtomlinson:
I've done a little more housekeeping here. 

- Added `distributed.http.prometheus.PrometheusCollector` for other collectors to base from
- Added a much simpler replacement of `_build_full_name` to the base class
- Restructured some files/names/attributes to be consistent with each other

fjetter:
cc @StephanErb @fgebhart this will break your dashboards ;)

distributed/http/scheduler/prometheus/semaphore.py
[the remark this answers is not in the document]
fjetter: > not having a prefix on these metrics to be a bug.

Yes. Sounds about right. I'm mostly wondering if there is anything we can do to communicate this effectively but you answered that already.

jacobtomlinson:
Metrics being renamed in this PR are:

```
semaphore_max_leases                   -> dask_semaphore_max_leases
semaphore_active_leases                -> dask_semaphore_active_leases
semaphore_pending_leases               -> dask_semaphore_pending_leases
semaphore_acquire                      -> dask_semaphore_acquire
semaphore_release                      -> dask_semaphore_release
semaphore_average_pending_lease_time_s -> dask_semaphore_average_pending_lease_time_s
```

For backward compatibility you folks could rename them to use the original names in the scrape config. If I were better at regex I expect you could do this in one rule.

```yaml
- source_labels: [__name__]
  regex:  'dask_semaphore_max_leases'
  target_label: __name__
  replacement: 'semaphore_max_leases'
- source_labels: [__name__]
  regex:  'dask_semaphore_active_leases'
  target_label: __name__
  replacement: 'semaphore_active_leases'
- source_labels: [__name__]
  regex:  'dask_semaphore_pending_leases'
  target_label: __name__
  replacement: 'semaphore_pending_leases'
- source_labels: [__name__]
  regex:  'dask_semaphore_acquire'
  target_label: __name__
  replacement: 'semaphore_acquire'
- source_labels: [__name__]
  regex:  'dask_semaphore_release'
  target_label: __name__
  replacement: 'semaphore_release'
- source_labels: [__name__]
  regex:  'dask_semaphore_average_pending_lease_time_s'
  target_label: __name__
  replacement: 'semaphore_average_pending_lease_time_s'
```

jacobtomlinson:
Python 3.9 / Windows failures seem related to #4713

jacobtomlinson:
CI failures appear related to #4719.

This is ready for final review and merge.

distributed/http/worker/prometheus/core.py
```diff
@@ -20,52 +22,54 @@ def collect(self):
...
-        tasks.add_metric(["stored"], len(self.worker.data))
-        tasks.add_metric(["executing"], self.worker.executing_count)
-        tasks.add_metric(["ready"], len(self.worker.ready))
-        tasks.add_metric(["waiting"], self.worker.waiting_for_data_count)
-        tasks.add_metric(["serving"], len(self.worker._comms))
+        tasks.add_metric(["stored"], len(self.server.data))
+        tasks.add_metric(["executing"], self.server.executing_count)
+        tasks.add_metric(["ready"], len(self.server.ready))
+        tasks.add_metric(["waiting"], self.server.waiting_for_data_count)
+        tasks.add_metric(["serving"], len(self.server._comms))
```
fjetter, line 33: Since we're cleaning up the namespace and have breaking changes in here anyhow, I suggest to change this one as well. This is published as a task gauge but, in fact, it is the number of comms in the comm pool. I'm not even sure if this serves any actual value, I'd be fine with removing it. Moving it to a different gauge would also be ok for me

distributed/http/worker/prometheus/core.py
```diff
@@ -20,52 +22,54 @@ def collect(self):
...
         yield tasks
 
         yield GaugeMetricFamily(
-            "dask_worker_connections",
+            self.build_name("connections"),
```
fjetter, line 37: Similar here. Technically this is not the number of connections but the number of concurrent fetch requests. There are many more connections out there

jacobtomlinson:
Many thanks for your input here @fjetter. I've updated those two.

fjetter (approved):
LGTM. Thanks @jacobtomlinson . This is really nice

PR #4723, opened 2021-04-20
Reintroduce `weight` function

jrbourbeau:
This PR adds the `weight` function (which was recently removed) back into `distributed/worker.py`. I heard this was being used by RAPIDS so it'd be good to keep it around for a bit before removing it in a future release. 

cc @quasiben @kkraus14

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -56,6 +56,7 @@ from .utils import (
     LRU,
     TimeoutError,
     _maybe_complex,
+    deprecated,
     get_ip,
     has_arg,
     import_file,
@@ -3718,7 +3719,7 @@ def convert_kwargs_to_str(kwargs, max_len=None):
         return "{{{}}}".format(", ".join(strs))
 
 
-# Keep `weight` for backwards compatibility
+@deprecated(version_removed="2021.06.0")
 def weight(k, v):
     return sizeof(v)
 
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -3718,7 +3718,13 @@ def convert_kwargs_to_str(kwargs, max_len=None):
         return "{{{}}}".format(", ".join(strs))
 
 
+# Keep `weight` for backwards compatibility
+def weight(k, v):
+    return sizeof(v)
+
+
 async def run(server, comm, function, args=(), kwargs=None, is_coro=None, wait=True):
+
     kwargs = kwargs or {}
     function = pickle.loads(function)
     if is_coro is None:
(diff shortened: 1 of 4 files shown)


distributed/worker.py
```diff
@@ -3718,6 +3718,11 @@ def convert_kwargs_to_str(kwargs, max_len=None):
         return "{{{}}}".format(", ".join(strs))
 
 
+# Keep `weight` for backwards compatibility
+def weight(k, v):
+    return sizeof(v)
```
fjetter, line 3723: Maybe raise a deprecation warning and remove it in X releases? Do we have a process for smth like this?

quasiben:
Thank you @jrbourbeau !

pentschev:
Just FYI, we have no problem in moving to `safe_sizeof`, which is what we're doing in https://github.com/rapidsai/dask-cuda/pull/584 . For the upcoming 0.19 release we're pinning Dask/Distributed 2021.04.0 in https://github.com/rapidsai/dask-cuda/pull/585, which is something we had already decided to do some weeks back in RAPIDS. So if we're reintroducing this only because of dask-cuda that may not be necessary.

distributed/worker.py
jrbourbeau: Good idea. We don't have a formal process for deprecation cycles today. This is definitely an area in which we can improve. I pushed a commit which adds a `deprecated` decorator which will hopefully make future deprecation cycles easier to manage.

jrbourbeau:
Thanks for the additional context @pentschev -- we're primarily adding this back for `dask-cuda` but, as the `dask-cuda` CI has pointed out, there _could_ be other folks using `weight` too. Temporarily adding it back isn't a big deal from `distributed`s perspective. 

Also, when is the 0.19 release due to be out? With `dask` / `distributed` being [release this Friday](https://github.com/dask/community/issues/150), it may be beneficial to add `weight` back in to avoid any breakages between `distributed` 2021.04.1 and `dask-cuda` 0.19 being released

pentschev:
> Thanks for the additional context @pentschev -- we're primarily adding this back for `dask-cuda` but, as the `dask-cuda` CI has pointed out, there _could_ be other folks using `weight` too. Temporarily adding it back isn't a big deal from `distributed`s perspective.

Makes sense to add the deprecation then.

> Also, when is the 0.19 release due to be out? With `dask` / `distributed` being [released this Friday](https://github.com/dask/community/issues/150), it may be beneficial to add `weight` back in to avoid any breakages between `distributed` 2021.04.1 and `dask-cuda` 0.19 being released

It's scheduled for release tomorrow, or Thursday depending on the timezone I guess, for reference we have a public [RAPIDS Release Schedule](https://docs.rapids.ai/maintainers). Parts of RAPIDS are already pinning Dask 2021.04.0 which is something we have decided to do some weeks ago anyway, so even if the release was after Friday I don't think it would impact us much, but thanks a lot for taking that into consideration, it's really appreciated! 😄

Issue #2155, opened 2018-08-02
test_workspace_concurrency failing regularly


mrocklin:
We've started to see failures in `test_diskutils.py::test_workspace_concurrency` that I'm unable to understand

```
_________________________ test_workspace_concurrency __________________________
Traceback (most recent call last):
  File "C:\projects\distributed\distributed\tests\test_diskutils.py", line 260, in test_workspace_concurrency
    _test_workspace_concurrency(tmpdir, 2.0, 6)
  File "C:\projects\distributed\distributed\tests\test_diskutils.py", line 255, in _test_workspace_concurrency
    assert n_purged >= 0.5 * n_created > 0
AssertionError: assert 0 >= (0.5 * 1200)
```

cc @pitrou , author of that test, in case he's able to explain what is going on

mrocklin:
Oh, I should point out that this is really only regularly failing on windows

pitrou:
Can you pinpoint to a specific commit?

mrocklin:
Not yet, no.  "Regularly failing" is a probably too strong to use here.
"Fails intermittently, but relatively frequently" would be better.

mrocklin:
I agree that running this in a for loop on a Windows VM and doing git
bisect would be a good approach to isolate things.  I was hoping that you
might be able to provide some information before doing that, especially
because I'm unlikely to get to this near term.

I've xfailed this short term in
https://github.com/dask/distributed/pull/2152

fjetter:
We've seen this popping up occasionally again on OSX

PR #4725, opened 2021-04-21
Improve graceful shutdown if nanny is involved

fjetter:
These fixes affect test runtimes and general graceful downscaling performance if a Nanny is involved. There are two fixes in here

* When using the test fixture `cluster`, there are explicit calls to `disconnect`, i.e. close every individual worker. This call involves the `rpc` implementation which, by default, waits for the remote handler to reply. This reply always times out since the worker terminated already and closed it's comm. Therefore, waiting for a reply is useless. This timeout is not big but a few seconds which we always have to wait when this thing stops. By not waiting for the reply, this finishes almost immediately. For most tests using this fixture, this makes up almost the entire test runtime
* The second fix involves the nanny. The graceful shutdown of a nanny worker is a bit messy at the moment and happens in two stages (I won't address this right now since it requires changing a bunch of handlers with subtle side effects). See for example https://github.com/dask/distributed/blob/e4b534af5722a7aebb8d5b086081984e93e1784f/distributed/scheduler.py#L3702-L3704 . In the graceful scaledown event, the nanny tried to unregister the worker unnecessarily which also runs into a timeout and slows down the shutdown. this has real world impact, not only test runtime

### DIFF
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -301,7 +301,9 @@ async def connect(
             upper_cap = min(time_left(), backoff_base * (2 ** attempt))
             backoff = random.uniform(0, upper_cap)
             attempt += 1
-            logger.debug("Could not connect, waiting for %s before retrying", backoff)
+            logger.debug(
+                "Could not connect to %s, waiting for %s before retrying", loc, backoff
+            )
             await asyncio.sleep(backoff)
     else:
         raise IOError(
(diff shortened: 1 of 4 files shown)


fjetter:
Unfortunately, didn't change a lot regarding overall test runtime but small, individual tests are now much faster :/

fjetter:
Test failure is https://github.com/dask/distributed/issues/2155

fjetter:
While CI doesn't seem to be impressed, real world impact for local testing is huge.

For instance, running `distributed/tests/test_client.py` on my machine

this branch: ~2:20min
main (56aed448): ~4:50min

jacobtomlinson (approved):
This looks great! The speedup I see locally is less than yours, but still significant:

Before - 0:05:54
After - 0:03:24

PR #4726, opened 2021-04-21
Add thread interrupt logic

### LINKED ISSUE #4694: Allow workers to cancel running task?
Currently, if a task is already running on a thread of a worker, it will always run to completion/exception even if the work is no longer required (e.g., the future was explicitly cancelled). We usually say "you cannot interrupt a python thread", so Dask has no option here. However, you [can](https://blog.finxter.com/how-to-kill-a-thread-in-python/#Method_4_Using_ctypes_To_Raise_Exceptions_In_A_Thread) set the thread state to exception using `ctypes.pythonapi.PyThreadState_SetAsyncExc`. Yes, this is a C-api call, but it is public and stable. Is this worth pursuing? It would allow not only early cancellation of unneeded work, but "interrupts" and timeouts from the worker itself, perhaps even better debugging if it can break deadlocks with a useful traceback.

(Note that the setting of an exception only has effect at the next python bytecode instruction, so some long-running C-level function or even `wait(big)` will not raise immediately)

martindurant:
- [X] Closes #4694
- [ ] Tests added / passed

### DIFF
--- a/distributed/_concurrent_futures_thread.py
+++ b/distributed/_concurrent_futures_thread.py
@@ -51,6 +51,28 @@ def _python_exit():
 atexit.register(_python_exit)
 
 
+class WorkerThreadInterrupt(threading.ThreadError):
+    pass
+
+
+def _async_raise(tid, exctype=WorkerThreadInterrupt):
+    """raise exception in given thread"""
+    import ctypes
+
+    if not issubclass(exctype, Exception):
+        raise TypeError("Must pass an exception type")
+    res = ctypes.pythonapi.PyThreadState_SetAsyncExc(
+        ctypes.c_long(tid), ctypes.py_object(exctype)
+    )
+    if res == 0:
+        raise threading.ThreadError("No thread got set")
+    elif res != 1:
+        # if res>1, somehow more than one thread was set, so immediately roll
+        # back and raise
+        ctypes.pythonapi.PyThreadState_SetAsyncExc(tid, 0)
+        raise threading.ThreadError("PyThreadState_SetAsyncExc failed")
+
+
 class _WorkItem:
     def __init__(self, future, fn, args, kwargs):
         self.future = future
@@ -65,6 +87,8 @@ class _WorkItem:
         try:
             result = self.fn(*self.args, **self.kwargs)
         except BaseException as e:
+            if isinstance(e, WorkerThreadInterrupt):
+                print("In-task interrupt")
             self.future.set_exception(e)
         else:
             self.future.set_result(result)
@@ -73,22 +97,27 @@ class _WorkItem:
 def _worker(executor_reference, work_queue):
     try:
         while True:
-            work_item = work_queue.get(block=True)
-            if work_item is not None:
-                work_item.run()
-                # Delete references to object. See issue16284
-                del work_item
-                continue
-            executor = executor_reference()
-            # Exit if:
-            #   - The interpreter is shutting down OR
-            #   - The executor that owns the worker has been collected OR
-            #   - The executor that owns the worker has been shutdown.
-            if _shutdown or executor is None or executor._shutdown:
-                # Notice other workers
-                work_queue.put(None)
-                return
-            del executor
+            try:
+                work_item = work_queue.get(block=True)
+                if work_item is not None:
+                    work_item.run()
+                    # Delete references to object. See issue16284
+                    del work_item
+                    continue
+                executor = executor_reference()
+                # Exit if:
+                #   - The interpreter is shutting down OR
+                #   - The executor that owns the worker has been collected OR
+                #   - The executor that owns the worker has been shutdown.
+                if _shutdown or executor is None or executor._shutdown:
+                    # Notice other workers
+                    work_queue.put(None)
+                    return
+                del executor
+            except WorkerThreadInterrupt:
+                # thread interrupt outside of a task
+                print("Interrupt within thread!")
+                pass
     except BaseException:
         _base.LOGGER.critical("Exception in worker", exc_info=True)
 
@@ -156,6 +185,9 @@ class ThreadPoolExecutor(_base.Executor):
             t.start()
             self._threads.add(t)
             _threads_queues[t] = self._work_queue
+
+    def interrupt(self, ident):
+        _async_raise(ident)
 
     def shutdown(self, wait=True):
         with self._shutdown_lock:
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -2454,6 +2454,9 @@ class Worker(ServerNode):
 
             if ts.state == "executing":
                 self.executing_count -= 1
+                th = [th for th, k in self.active_threads.items() if k == key]
+                if th:
+                    self.executor.interrupt(th[0])
 
             if ts.resource_restrictions is not None:
                 if ts.state == "executing":
--- a/distributed/threadpoolexecutor.py
+++ b/distributed/threadpoolexecutor.py
@@ -40,23 +40,27 @@ def _worker(executor, work_queue):
 
     try:
         while thread_state.proceed:
-            with executor._rejoin_lock:
-                if executor._rejoin_list:
-                    rejoin_thread, rejoin_event = executor._rejoin_list.pop()
-                    executor._threads.add(rejoin_thread)
-                    executor._threads.remove(threading.current_thread())
-                    rejoin_event.set()
-                    break
             try:
-                task = work_queue.get(timeout=1)
-            except queue.Empty:
-                continue
-            if task is not None:  # sentinel
-                task.run()
-                del task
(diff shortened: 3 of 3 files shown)


martindurant:
This shows the basic working. The following script runs successfully:
```
import dask.distributed
import time
import threading
client = dask.distributed.Client(processes=False, n_workers=1, threads_per_worker=1)
client.wait_for_workers(1)

def runme(N):
    for i in range(N):
        time.sleep(1)

fut = client.submit(runme, 10000000)
time.sleep(0.01)  # make sure fut is allocated first
fut0 = client.submit(lambda x: x, True)

fut.cancel()
fut0.result()
```

As coded, this happens whenever an executing key is released for whatever reason, and there is no config for it. Also, if the task in question captures exceptions, it will keep running.

(please ignore print statements)

mrocklin:
This is cool to see.  I know that this is still just a draft status, and I'm sure that this is in your plan, but I would encourage you to write down tests to demonstrate that some of the concerns raised in the issue are handled by this, for example that finally blocks are respected and such.

martindurant:
@mrocklin : first pass at a pair of tests, but we will still need to be very careful and thorough here. btw: it took me ages to realist that `fut.cancel()` is a coroutine needing await.

I would still like some feedback here on whether this is a good idea for the general case, or should be a config, or only exposed via some explicit client call. The latter is hard, since the user may not have a handle to the particular task that is hanging (or just taking longer than expected).

martindurant:
(windows error is in `test_str`, not obviously related)

jrbourbeau:
cc @crusaderky

fjetter:
FWIW, I would be comfortable enough with calling C API directly for this to be merged. The API seems to be stable since py3.7 (https://docs.python.org/3/c-api/init.html#c.PyThreadState_SetAsyncExc)
If we add more tests and document this behaviour, I'd be fine with it. In terms of complexity, this is also justified considering the benefits. If this were to make trouble down the road, we could think about toggles but for now I'd like to avoid more options.

I'm wondering, would this be something we could contribute to CPython directly? I understand that we had the need to vendor the threadpool for the seceding feature but every modification raises maintenance cost (or the risk of it)

martindurant:
I have made the default False for now (no interrupt) and added the key to the config schema. It is pretty hidden! We could add this to the docs somewhere, or softly launch it by suggesting some users try it when facing long-running released tasks.

distributed/worker.py
```diff
@@ -2953,6 +2966,8 @@ def release_key(
             if key in self.threads:
                 del self.threads[key]
 
+            # task is released but still running - set thread exception
+            # whether or not to do this automatically should be configurable
```
fjetter, line 2970: This comment feels a bit out of place. Should this be part of `transition_executing_released`?

distributed/worker.py
```diff
@@ -2139,6 +2145,13 @@ def transition_executing_released(self, ts, *, stimulus_id):
         # See https://github.com/dask/distributed/pull/5046#discussion_r685093940
         ts.state = "cancelled"
         ts.done = False
+
+        if self.interruptor:
+            th = [th for th, k in self.active_threads.items() if k == ts.key]
+            if th:
+                logger.info("Interrupting thread %i for task %s", th[0], ts.key)
+                self.executor.interrupt(th[0])
```
fjetter, line 2153: I would argue it is not possible for `th` to be empty. If that is true, there should probably be a `else: RuntimeError`. Although I'm a bit ambivalent since on the one hand I'd like us to avoid implicit fail cases but on the other hand, we are not raising exceptions anywhere in the transition engine, so far.

distributed/_concurrent_futures_thread.py
```diff
@@ -50,6 +50,28 @@ def _python_exit():
 atexit.register(_python_exit)
 
 
+class WorkerThreadInterrupt(threading.ThreadError):
+    pass
+
+
+def _async_raise(tid: int, exctype: type = WorkerThreadInterrupt):
```
fjetter, line 57: Maybe `_set_thread_exception`? The async prefix is a bit misleading (I know the proper ABI uses it) since it is mixing with asyncio async/coroutines.

distributed/worker.py
martindurant: I imagine it's possible for the task to end at the same time as it is released for another reason. If it turns out there is no longer a thread associated, that doesn't seem like an error to me.

distributed/_concurrent_futures_thread.py
```diff
@@ -50,6 +50,28 @@ def _python_exit():
...
+    res = ctypes.pythonapi.PyThreadState_SetAsyncExc(
+        ctypes.c_long(tid), ctypes.py_object(exctype)
+    )
+    if res == 0:
+        raise threading.ThreadError("No thread got set")
+    elif res != 1:  # pragma: no cover
+        # if res>1, somehow more than one thread was set, so immediately roll
+        # back and raise. Should never happen.
+        ctypes.pythonapi.PyThreadState_SetAsyncExc(tid, 0)
+        raise threading.ThreadError("PyThreadState_SetAsyncExc failed")
```
fjetter, line 72: I don't think we actually need this block. The docs specifically mention zero and one. res>1 is not defined and it is unclear to me what we should even do in this case, let alone how to "roll back" that situation.

Even peaking into the implementation, this case is not possible, see https://github.com/python/cpython/blob/7f61d9d84843e3445f62eb00c47902f0daa30a72/Python/pystate.c#L1165-L1202

distributed/_concurrent_futures_thread.py
```diff
@@ -156,6 +182,9 @@ def weakref_cb(_, q=self._work_queue):
             self._threads.add(t)
             _threads_queues[t] = self._work_queue
 
+    def interrupt(self, ident):
+        _async_raise(ident)
```
fjetter, line 186: functionally no difference but for readability, I would prefer `WorkerThreadInterrupt` being mentioned here. If the `_async_raise` is renamed that may also be fine but at this point we rely specifically to raise this exception and not something else

distributed/tests/test_worker.py
```diff
@@ -1778,6 +1779,55 @@ async def test_story(c, s, w):
...
+
+@gen_cluster(client=True, nthreads=[("127.0.0.1", 1)], timeout=2)
+async def test_interrupt(c, s, w):
+    import concurrent.futures
+
+    w.interruptor = True
+    fut0 = c.submit(donothing)
+    await async_wait_for(lambda: bool(w.active_threads), 1)
+    fut1 = c.submit(lambda: True)  # blocked
+    assert not fut1.done()
```
fjetter, line 1803: I believe you should insert a small sleep here `await asyncio.sleep(0.01)` since even if this function returns almost immediately, the future cannot be done in the same loop cycle as it has been submitted

distributed/worker.py
```diff
@@ -766,6 +766,12 @@ def __init__(
 
         self.memory_limit = parse_memory_limit(memory_limit, self.nthreads)
 
+        if "thread_auto_interrupt" in kwargs:
+            self.interruptor = kwargs.pop("thread_auto_interrupt")
```
fjetter, line 770: imo, if we want to support this in the constructor, we should add it explicitly and add a doc string to it. I would be fine with supporting this merely via config, though.

distributed/worker.py
fjetter: Good point. Can we test this?

fjetter: At the very least, we should add a comment here. It's not straight forward to figure out and it feels like an important detail.

I'm also wondering now what happens if the thread finishes properly while we're trying to interrupt vs the task raised an exception before and we're trying to interrupt. What will be the final task state if either one happens? (I'm mostly concerned about invalid states. If we rerun anything, I don't mind as long as the state machine isn't messed up)

martindurant: > Can we test this?

I'm don't see a way we can emulate it without horrible race problems.

> what happens if the thread finishes properly while we're trying to interrupt

The thread doesn't finish as such, it goes into the loop to pick more work, which is what [this](https://github.com/dask/distributed/pull/4726/files#diff-894cd176005fcf92cb8b6a46d6a4efdb3644d23ef81928ad2ec91c1190e671d9R114) is for. Having said that, the _WorkItem s don't have a state beyond what's encoded in the corresponding future.

> task raised an exception before and we're trying to interrupt

If the task is already in an Except block, you would expect to see "During handling of the above exception, another exception occurred:" as the return, but `finally:` blocks would still run.

> If we rerun anything, I don't mind as long as the state machine isn't messed up

Fundamentally, we are cleaning up a task which is set for release, and so the output of the task and the Item corresponding object's state don't matter.

martindurant:
Anything left here?

martindurant:
ping?

PR #4727, opened 2021-04-21
Unit test for metrics["memory"]=None

crusaderky:
Follow up unit test for #4651


fjetter (approved):

fjetter:
ooof, tests fail with a very low level tornado assert

```
    def _start_read(self) -> Future:
        if self._read_future is not None:
            # It is an error to start a read while a prior read is unresolved.
            # However, if the prior read is unresolved because the stream was
            # closed without satisfying it, it's better to raise
            # StreamClosedError instead of AssertionError. In particular, this
            # situation occurs in harmless situations in http1connection.py and
            # an AssertionError would be logged noisily.
            #
            # On the other hand, it is legal to start a new read while the
            # stream is closed, in case the read can be satisfied from the
            # read buffer. So we only want to check the closed status of the
            # stream if we need to decide what kind of error to raise for
            # "already reading".
            #
            # These conditions have proven difficult to test; we have no
            # unittests that reliably verify this behavior so be careful
            # when making changes here. See #2651 and #2719.
            self._check_closed()
>           assert self._read_future is None, "Already reading"
E           AssertionError: Already reading
```

crusaderky:
> ooof, tests fail with a very low level tornado assert
> E           AssertionError: Already reading
> ```

That's a random failure in test_map; nothing to do with this PR.

distributed/tests/test_scheduler.py
```diff
@@ -2442,3 +2443,23 @@ async def test_memory_no_zict(c, s, a, b):
...
+
+
+@gen_cluster(client=True, nthreads=[])
+async def test_memory_is_none(c, s):
+    """If Worker.heartbeat() runs before Worker.monitor.update(), then
+    Worker.metrics["memory"] will be None and will need special handling in
+    Worker.memory and Scheduler.heartbeat_worker().
+    """
+    with mock.patch("distributed.system_monitor.SystemMonitor.update"):
+        w = await Worker(s.address, nthreads=1)
```
jrbourbeau, line 2455: Could you use a context manager here to ensure the `Worker` is shut down properly (or `await w.close()`)

distributed/tests/test_scheduler.py
```diff
@@ -2442,3 +2443,23 @@ async def test_memory_no_zict(c, s, a, b):
...
     assert s.memory.managed == 0
+
+
+@gen_cluster(client=True, nthreads=[])
+async def test_memory_is_none(c, s):
+    """If Worker.heartbeat() runs before Worker.monitor.update(), then
+    Worker.metrics["memory"] will be None and will need special handling in
+    Worker.memory and Scheduler.heartbeat_worker().
+    """
+    with mock.patch("distributed.system_monitor.SystemMonitor.update"):
```
jrbourbeau, line 2454: Do either of @fjetter's suggestions in https://github.com/dask/distributed/pull/4719#discussion_r617537532 allow us to test the case where the worker heartbeats before the system monitor has updated? It'd be nice to avoid `mock`ing if possible

jrbourbeau (commented):
Thanks for adding this test @crusaderky. Left a couple of small comments, but overall this looks great

distributed/tests/test_scheduler.py
crusaderky: done

distributed/tests/test_scheduler.py
jrbourbeau: Alright, thanks for trying them (and a couple other options) out 👍

jrbourbeau (approved):
Thanks @crusaderky -- will merge after CI finishes

Issue #4721, opened 2021-04-20
KeyError: ('ready', 'memory')


mdering:
**What happened**:

    ERROR - 2021-04-20 08:43:37,220 - distributed.worker.execute.l2800 - ('ready', 'memory')
    Traceback (most recent call last):
      File "\miniconda3\lib\site-packages\distributed\worker.py", line 2765, in execute
        self.transition(ts, "memory", value=value)
      File "\miniconda3\lib\site-packages\distributed\worker.py", line 1584, in transition
        func = self._transitions[start, finish]
    KeyError: ('ready', 'memory')

We see this key error occasionally during execution of our application. I checked through the worker code and theres already a function defined for this transition, its just not in the `_transitions` table. Is this an oversight? https://github.com/dask/distributed/blob/d5fc324bdef22f19d77c3d36e63a3778ceaac0b0/distributed/worker.py#L1840

**What you expected to happen**:
I would expect this to not happen

**Minimal Complete Verifiable Example**:

I don't have one, we just see it sometimes. I'm opening this issue just to see if this is an oversight or not.

**Anything else we need to know?**:

**Environment**:

- Dask version: 2021.04.0
- Python version: 3.7
- Operating System: windows
- Install method (conda, pip, source): conda

fjetter:
Thanks for the report. We've seen issues around this lately. There is a bigger issue open https://github.com/dask/distributed/issues/4413

I believe for this particular instance it indeed looks like an oversight. You don't happen to have a minimal example which reproduces this error, do you?

gforsyth:
Looks like I missed out adding it to the `_transitions` table -- thanks for flagging it @mdering

mdering:
I'm sorry, I don't. We have a largeish application and we just see this in the logs sometimes, the error message doesn't even tell me which future it is so I can't deduce anything about it from that either. I appreciate the quick responses!

mdering:
Hi everyone, I was able to reproduce this finally. Our experience with dask had lead us to notice that some of the workers would jam up inexplicably, and this would prevent our application from proceeding. These workers would be given tasks, but the tasks would never start nor be completed. This lead me to use a method based on calling `reschedule` on the scheduler, if a key took too long to progress. This seems to have mostly worked, though the workers eventually end up in strange states. I may raise a different issue about that based on the feedback I get here.

Anyway, on the example
```python
import distributed
import time
client = distributed.Client("127.0.0.1:8786") # i have a scheduler with 4 workers here locally
[...]

PR #4728, opened 2021-04-21
Add "ready->memory" to transitions in worker

gforsyth:
This should remove the error message in #4721 but is more of a band-aid
than a fix.  Ready->memory transitions shouldn't happen, but since they
do occassionally crop up, we might as well dispatch them appropriately
until the worker state machine is ready.

- [ ] Closes https://github.com/dask/distributed/issues/4721
- [ ] Tests added / passed

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1840,7 +1840,7 @@ class Worker(ServerNode):
         self.send_task_state_to_scheduler(ts)
 
     def transition_ready_memory(self, ts, value=no_value):
-        if value:
+        if value is not no_value:
             self.put_key_in_memory(ts, value=value)
         self.send_task_state_to_scheduler(ts)
 
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -1839,7 +1839,7 @@ class Worker(ServerNode):
             assert ts.traceback is not None
         self.send_task_state_to_scheduler(ts)
 
-    def transition_ready_memory(self, ts, value=None):
+    def transition_ready_memory(self, ts, value=no_value):
         if value:
             self.put_key_in_memory(ts, value=value)
         self.send_task_state_to_scheduler(ts)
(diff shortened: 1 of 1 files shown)


gforsyth:
cc @fjetter

jrbourbeau (commented):
Thanks @gforsyth! Temporarily adding transition transition handler in seems like a good thing to do.

I realize this pre-dates this PR, but I'm wondering if the default value for `value=` in `transition_ready_memory`

https://github.com/dask/distributed/blob/e57538374e566e6c264a0929937458c17a95e843/distributed/worker.py#L1842-L1845

should be `no_value`

https://github.com/dask/distributed/blob/e57538374e566e6c264a0929937458c17a95e843/distributed/worker.py#L88

instead of `None` to cover the cases when the actual result of a task is falsey (e.g. `None`, `False`, `[]`, etc.)

gforsyth:
Good catch @jrbourbeau -- I've updated accordingly.

distributed/worker.py
```diff
@@ -1837,7 +1839,7 @@ def transition_ready_error(self, ts):
             assert ts.traceback is not None
         self.send_task_state_to_scheduler(ts)
 
-    def transition_ready_memory(self, ts, value=None):
+    def transition_ready_memory(self, ts, value=no_value):
         if value:
```
jrbourbeau, line 1843: Sorry, I should have included this with my earlier comment

```suggestion
        if value is not no_value:
```

gforsyth: Thanks, I should've caught that.

jrbourbeau (approved):
Thanks @gforsyth! 

@mdering are you able to try this PR out to confirm that it resolves the `KeyError` you were running into?

fjetter (approved):
@gforsyth thanks! 

Did you, by any chance, find a test where this is triggered? The only scenario I can come up with involves some back and forth work stealing but I wouldn't know how to synthesise this

PR #4729, opened 2021-04-21
Avoid `active_threads` changing size during iteration

jrbourbeau:
Closes https://github.com/dask/dask/issues/7581. @bsesar would you mind trying this PR out to see if it resolves the error you were seeing?

Note that we could also acquire `active_threads_lock` during the heartbeat to avoid `self.active_threads` changing size during iteration. However since we acquire `active_threads_lock` during task execution and heartbeat often I opted to not acquire the lock here to avoid coupling task execution and heartbeating. Happy to acquire `active_threads_lock` instead if others prefer.

### DIFF
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -945,16 +945,16 @@ class Worker(ServerNode):
         logger.debug("Heartbeat: %s", self.address)
         try:
             start = time()
+            with self.active_threads_lock:
+                active_keys = list(self.active_threads.values())
             response = await retry_operation(
                 self.scheduler.heartbeat_worker,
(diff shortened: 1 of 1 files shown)


fjetter:
I'm not entirely sure if this is sufficient. The dict comprehension _should_ do the same thing as the copy, shouldn't it?


We are acquiring the lock here during profiling as well, see https://github.com/dask/distributed/blob/f492aa7e7a39270aaf0b58e37ed855c149b68a8f/distributed/worker.py#L2940-L2941

fjetter (approved):

PR #4633, opened 2021-03-25
ignore oserror exception when scaling down (dask#4591)

gerald732:
Adaptive stop() is called if workers are preempted. From the discussion #4591 with @jacobtomlinson , it's reasonable to ignore the exception and carry on with adaptive when scaling down

PR #4731, opened 2021-04-22
Fix `UnboundLocalError` in `AdaptiveCore.adapt()`

andersy005:
This PR addresses the `UnboundLocalError`  accidentally introduced by https://github.com/dask/distributed/pull/4633

This error occurs when the exception is raised before the assignment of `status` variable:

```python

  File "/glade/work/abanihi/opt/miniconda/envs/playground/lib/python3.8/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "/glade/work/abanihi/opt/miniconda/envs/playground/lib/python3.8/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
    future.result()
  File "/glade/work/abanihi/opt/miniconda/envs/playground/lib/python3.8/site-packages/distributed/deploy/adaptive_core.py", line 204, in adapt
    if status != "down":
UnboundLocalError: local variable 'status' referenced before assignment
```

- [ ] Tests added / passed


distributed/deploy/tests/test_adaptive_core.py
```diff
@@ -89,3 +92,21 @@ async def test_interval():
...
+        logger.addHandler(fh)
+        mock_safe_target.side_effect = OSError
+        adapt = MyAdaptive(minimum=1, maximum=4, wait_count=2)
+        await adapt.adapt()
+        fh.flush()
+        with open(fn) as f:
+            text = f.read()
+            assert "Adaptive stopping due to error" in text
+            assert "Adaptive stop" in text
+        assert not adapt._adapting
```
fjetter, line 112: For some reason, the tests are failing sporadically since there is no log msg observered. I'm not sure if this is connected to the filehandler or not but we do have the util function `captured_logger` which registers a streamhandler as the only logger and offers the stream buffer as the ctx mgr return value. This is how we write most tests which are sensitive to logging

```suggestion
    with captured_logger("distributed.deploy.adaptive_core") as log:
        mock_safe_target.side_effect = OSError
        adapt = MyAdaptive(minimum=1, maximum=4)
        await adapt.adapt()
    text = log.getvalue()
    assert "Adaptive stopping due to error" in text
    assert "Adaptive stop" in text
    assert not adapt._adapting
    assert not adapt.periodic_callback
```

fjetter (commented):
Thanks for the submission of this fix. Code looks good to me but we should try to find out why the test fails on some CI jobs

distributed/deploy/tests/test_adaptive_core.py
andersy005: Thank you, @fjetter

jrbourbeau (commented):
Thanks @andersy005! This looks great -- just left one nitpicky comment below

distributed/deploy/tests/test_adaptive_core.py
```diff
@@ -89,3 +91,17 @@ async def test_interval():
     adapt._target = 10
     await asyncio.sleep(0.020)
     assert len(adapt.plan) == 1  # last value from before, unchanged
+
+
+@pytest.mark.asyncio
+@mock.patch("distributed.deploy.adaptive_core.AdaptiveCore.safe_target")
```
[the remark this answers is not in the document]
andersy005: > Nitpick: Instead of using mock could we instead create an explicit adaptive subclass (similar to MyAdaptive) which raises an OSError when attempting to adapt (see the diff below)?

I like this idea. I too find mocking confusing sometimes and hard to reason about for users who didn't write the mocking code... I will update the test shortly

jrbourbeau (approved):
Thanks for fixing this @andersy005! Will merge after CI finishes

Issue #4640, opened 2021-03-26
Incorrect arguments to Worker/LocalCluster gives a confusing error


gjoseph92:
On 2.30.0, passing an invalid argument to Worker gave you an error telling you what you did wrong. This is easy for users to do via LocalCluster/Client (for example, giving `num_workers` instead of `n_workers`, which is exactly what I mistyped):

```python
# used to get this
In [1]: import distributed

In [2]: distributed.__version__
Out[2]: '2.30.0'

In [3]: distributed.Worker("tcp://localhost", 8000, foo=1)
TypeError: __init__() got an unexpected keyword argument 'foo'
```

On main (likely anything after #4365), the error message is less helpful:
```python
# now get this
In [1]: import distributed

In [2]: distributed.__version__
Out[2]: '2021.03.0+36.g8c2c6738.dirty'

In [3]: distributed.Worker("tcp://localhost", 8000, foo=1)
TypeError: object.__init__() takes exactly one argument (the instance to initialize)
```

This is coming from `Server.__init__` here: https://github.com/dask/distributed/blob/25a2867e8e950d9cb51b071337ffc43ddaf0bb52/distributed/core.py#L241

because now that `Scheduler` is using multiple-inheritance, we wanted the base classes to be able to play nicely with each other and propagate kwargs up the hierarchy. However, the fact that none of the base classes "bottom out" and assume that they are, in fact, the base of the MRO means we lose the opportunity to decide when some kwargs are actually wrong, versus meant for the class above us.

FWIW, `Server` does (at the moment) appear a safe place to not take `**kwargs` or call `super().__init__`, since it does come last in the MRO for all of its subclasses:

```python
In [8]: distributed.Worker.__mro__
Out[8]: 
(distributed.worker.Worker,
 distributed.node.ServerNode,
 distributed.core.Server,
 object)

In [9]: distributed.Scheduler.__mro__
Out[9]: 
(distributed.scheduler.Scheduler,
 distributed.scheduler.SchedulerState,
 distributed.node.ServerNode,
 distributed.core.Server,
 object)

In [10]: distributed.node.ServerNode.__mro__
Out[10]: (distributed.node.ServerNode, distributed.core.Server, object)
```

However, it could be brittle to assume this will always be the case. And since all we really care about is just presenting a better error message, maybe there's a different approach to take here?

Some discussion of this already in https://github.com/dask/distributed/pull/4365/files/2212edc11b424c17f2f656eb73b2ca206742705a#r557630488

[...]

PR #4733, opened 2021-04-22
Ensure exceptions are raised if workers are incorrectly started

fjetter:
`Server` inherits from `object` which has no sensible initialiser. In particular, when provided with kwargs, it raises an obscure `TypeError`

`TypeError: object.__init__() takes exactly one argument (the instance to initialize)`

Instead of raising, this would issue a warning indicating that some arguments were not used

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -136,7 +136,6 @@ class Server:
         connection_args=None,
         timeout=None,
         io_loop=None,
-        **kwargs,
     ):
         self.handlers = {
             "identity": self.identity,
@@ -237,8 +236,6 @@ class Server:
         )
 
         self.__stopped = False
-
-        super().__init__(**kwargs)
 
     @property
     def status(self):
--- a/distributed/deploy/tests/test_local.py
+++ b/distributed/deploy/tests/test_local.py
@@ -1063,3 +1063,21 @@ async def test_cluster_names():
 
         async with LocalCluster(processes=False, asynchronous=True) as unnamed_cluster2:
             assert unnamed_cluster2 != unnamed_cluster
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("nanny", [True, False])
+async def test_local_cluster_redundant_kwarg(nanny):
+    with pytest.raises(TypeError, match="unexpected keyword argument"):
+        # Extra arguments are forwarded to the worker class. Depending on
+        # whether we use the nanny or not, the error treatment is quite
+        # different and we should assert that an exception is raised
+        async with await LocalCluster(
+            typo_kwarg="foo", processes=nanny, n_workers=1
+        ) as cluster:
+
+            # This will never work but is a reliable way to block without hard
+            # coding any sleep values
+            async with Client(cluster) as c:
+                f = c.submit(sleep, 0)
+                await f
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -9,6 +9,7 @@ import warnings
 import weakref
 from contextlib import suppress
 from multiprocessing.queues import Empty
+from time import sleep as sync_sleep
 
 import psutil
 from tornado import gen
@@ -362,7 +363,6 @@ class Nanny(ServerNode):
                 config=self.config,
             )
 
-        self.auto_restart = True
         if self.death_timeout:
             try:
                 result = await asyncio.wait_for(
@@ -524,6 +524,9 @@ class Nanny(ServerNode):
 
 
 class WorkerProcess:
+    # The interval how often to check the msg queue for init
+    _init_msg_interval = 0.05
+
     def __init__(
         self,
         worker_kwargs,
@@ -589,9 +592,14 @@ class WorkerProcess:
         except OSError:
             logger.exception("Nanny failed to start process", exc_info=True)
             self.process.terminate()
-            return
-
-        msg = await self._wait_until_connected(uid)
+            self.status = Status.failed
+            return self.status
+        try:
+            msg = await self._wait_until_connected(uid)
+        except Exception:
+            self.status = Status.failed
+            self.process.terminate()
+            raise
         if not msg:
             return self.status
         self.worker_address = msg["address"]
@@ -688,14 +696,15 @@ class WorkerProcess:
                 logger.error("Failed to kill worker process: %s", e)
 
     async def _wait_until_connected(self, uid):
-        delay = 0.05
         while True:
             if self.status != Status.starting:
                 return
+            # This is a multiprocessing queue and we'd block the event loop if
+            # we simply called get
             try:
                 msg = self.init_result_q.get_nowait()
             except Empty:
-                await asyncio.sleep(delay)
+                await asyncio.sleep(self._init_msg_interval)
                 continue
 
             if msg["uid"] != uid:  # ensure that we didn't cross queues
@@ -705,7 +714,6 @@ class WorkerProcess:
                 logger.error(
                     "Failed while trying to start worker process: %s", msg["exception"]
                 )
-                await self.process.join()
                 raise msg["exception"]
             else:
                 return msg
@@ -783,6 +791,14 @@ class WorkerProcess:
                     logger.exception("Failed to start worker")
                     init_result_q.put({"uid": uid, "exception": e})
                     init_result_q.close()
+                    # If we hit an exception here we need to wait for a least
+                    # one interval for the outside to pick up this message.
+                    # Otherwise we arrive in a race condition where the process
+                    # cleanup wipes the queue before the exception can be
+                    # properly handled. See also
+                    # WorkerProcess._wait_until_connected (the 2 is for good
+                    # measure)
+                    sync_sleep(cls._init_msg_interval * 2)
                 else:
                     try:
                         assert worker.address
@@ -804,6 +820,12 @@ class WorkerProcess:
             logger.exception("Failed to initialize Worker")
             init_result_q.put({"uid": uid, "exception": e})
             init_result_q.close()
+            # If we hit an exception here we need to wait for a least one
+            # interval for the outside to pick up this message. Otherwise we
+            # arrive in a race condition where the process cleanup wipes the
+            # queue before the exception can be properly handled. See also
+            # WorkerProcess._wait_until_connected (the 2 is for good measure)
+            sync_sleep(cls._init_msg_interval * 2)
         else:
             try:
                 loop.run_sync(run)
--- a/distributed/nanny.py
+++ b/distributed/nanny.py
@@ -378,7 +378,11 @@ class Nanny(ServerNode):
                 raise
 
         else:
-            result = await self.process.start()
+            try:
+                result = await self.process.start()
+            except Exception:
+                await self.close()
+                raise
         return result
 
     async def restart(self, comm=None, timeout=2, executor_wait=True):
@@ -414,9 +418,10 @@ class Nanny(ServerNode):
         """ Track worker's memory.  Restart if it goes above terminate fraction """
         if self.status != Status.running:
             return
+        if self.process is None or self.process.process is None:
+            return None
         process = self.process.process
-        if process is None:
-            return
+
         try:
             proc = self._psutil_process
             memory = proc.memory_info().rss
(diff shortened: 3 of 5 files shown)


distributed/core.py
```diff
@@ -238,7 +238,8 @@ def set_thread_ident():
 
         self.__stopped = False
 
-        super().__init__(**kwargs)
+        if kwargs:
+            logger.warning("Provided unused arguments %s", list(kwargs.keys()))
```
jrbourbeau, line 242: Hmm I think the call to `super().__init__` was intentional -- see https://github.com/dask/distributed/pull/4365/files/2212edc11b424c17f2f656eb73b2ca206742705a#r557630488 for an earlier discussion on this point

distributed/deploy/tests/test_local.py
```diff
@@ -1061,3 +1063,17 @@ async def test_cluster_names():
 
         async with LocalCluster(processes=False, asynchronous=True) as unnamed_cluster2:
             assert unnamed_cluster2 != unnamed_cluster
+
+
+@pytest.mark.asyncio
+async def test_local_cluster_redundant_kwarg():
```
jrbourbeau, line 1069: Could you add the `cleanup` fixture here to make sure threads, processes, `rpc`s, etc. are cleaned up

```suggestion
async def test_local_cluster_redundant_kwarg(cleanup):
```

distributed/deploy/tests/test_local.py
```diff
@@ -1061,3 +1063,17 @@ async def test_cluster_names():
...
+    with captured_logger("distributed.core", level=logging.WARNING) as log:
+        cluster = await LocalCluster(
+            asynchronous=True, processes=False, typo_kwarg="foo"
+        )
+        client = await Client(cluster, asynchronous=True)
```
jrbourbeau, line 1074: Could you switch these over to use context managers in order to avoid calling methods like `.close()` later?

distributed/deploy/tests/test_local.py
```diff
@@ -1061,3 +1063,17 @@ async def test_cluster_names():
...
+        fut = client.submit(identity, 1)
+        assert await fut.result() == 1
+        cluster.close()
+    log = log.getvalue()
+    assert "unused arguments" in log
```
jrbourbeau, line 1079:
```suggestion
    assert "unused arguments" in log
    assert "typo_kwarg" in log
```

distributed/tests/test_core.py
```diff
@@ -880,3 +881,12 @@ async def sleep(comm=None):
...
         assert not len(server._ongoing_coroutines)
+
+
+@pytest.mark.asyncio
+async def test_server_redundant_kwarg():
+    with captured_logger("distributed.core", level=logging.WARNING) as log:
+        server = await Server({}, typo_kwarg="foo")
+        server.stop()
+    log = log.getvalue()
+    assert "unused arguments" in log
```
jrbourbeau, line 892:
```suggestion
    assert "unused arguments" in log
    assert "typo_kwarg" in log
```

jrbourbeau (commented):
Thanks @fjetter -- this would close https://github.com/dask/distributed/issues/4640 (cc @gjoseph92)

distributed/core.py
fjetter: multiple inheritance is a wonderful monstrosity :D

```python
In [1]: class A:
   ...:     def __init__(self, a, **kwargs):
   ...:         print(f"A: {kwargs}")
   ...:

In [2]: class B(A):
   ...:     def __init__(self, b, **kwargs):
   ...:         print(f"B: {kwargs}")
   ...:         super().__init__(**kwargs)
   ...:

In [3]: class C(A):
   ...:     def __init__(self, c, **kwargs):
   ...:         print(f"C: {kwargs}")
   ...:         super().__init__(**kwargs)
   ...:

In [4]: class D(B, C):
   ...:     def __init__(self, d, **kwargs):
   ...:         print(f"D: {kwargs}")
   ...:         super().__init__(**kwargs)
   ...:

In [5]: D.mro()
Out[5]: [__main__.D, __main__.B, __main__.C, __main__.A, object]

In [6]: D(a="a", b="b", c="c", d="d")
D: {'a': 'a', 'b': 'b', 'c': 'c'}
B: {'a': 'a', 'c': 'c'}
C: {'a': 'a'}
A: {}
```

kwargs will trickle down the mro. therefore we need to allow kwargs everwhere. This is true for all classes but the root. the root cannot call super with kwargs since object doesn't allow arguments. afaik, the init on object is a no-op

gjoseph92 (commented):
Thanks for this @fjetter. I agree with you that removing `super().__init__` is fine, since `Server` is always at the root of the MRO—so long as we don't significantly change the inheritance structure in the future.

jrbourbeau:
If we're removing the `super` call then perhaps we should also remove `kwargs` in the signature and error loudly when an unexpected keyword is given

distributed/core.py
gjoseph92: We could consider adding a test like this (untested):
```python
def test_server_inheritance():
    subclasses = Server.__subclasses__()
    for subcls in subclasses:
        assert subcls.__mro__[-2:] == [Server, object], f"Server must always be the last class in the MRO. {subcls} has MRO: {subcls.__mro__}"
```
Obviously this would depend on all Server subclasses being imported at the time the test runs.

Alternatively, if we wanted to over-engineer it a bit, we could use [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html#customizing-class-creation) (added in 3.6) to enforce this. I've done something like this before [here](https://github.com/descarteslabs/descarteslabs-python/blob/master/descarteslabs/workflows/types/widget/widget.py#L42-L57). Not sure how it would play with Cython though?

fjetter:
> If we're removing the super call then perhaps we should also remove kwargs in the signature and error loudly when an unexpected keyword is given

Sure. I was assuming we had this to allow for easier forward compatible code but I would prefer to be strict as well.

fjetter:
I removed the kwargs entirely. I was about to delete the tests but discovered this interesting edge case we didn't properly cover. If the local cluster is provided with foo, it is passed to the worker. If one uses a nanny this is entirely swallowed and the nanny simply doesn't come up. I'm not even sure if an exception was logged...

jrbourbeau (commented):
cc @jakirkham for visibility

jacobtomlinson (approved):

distributed/core.py
```diff
@@ -238,8 +237,6 @@ def set_thread_ident():
 
         self.__stopped = False
 
-        super().__init__(**kwargs)
-
```
jakirkham, line 242: These were needed because of [the inheritance in `Scheduler`]( https://github.com/dask/distributed/blob/233ec8840671b5d705a06d0dcfe1ff42b771230b/distributed/scheduler.py#L3176 ) ( https://github.com/dask/distributed/pull/4365#discussion_r557630488 ). Is there some reason to remove them? Would suspect they are still needed

fjetter: See https://github.com/dask/distributed/pull/4733#discussion_r618547785 I believe this is only needed for classes which are sandwiched, i.e. are the intermediates in the diamond inheritance. This class is always root and therefore will not need it. In fact, calling super here *with* arguments will raise an exception since _super_ is `object` which cannot be initialized with arguments

jrbourbeau: I think this was added to make us more resilient to the inheritance order (see [this earlier comment](https://github.com/dask/distributed/pull/4733#discussion_r618603420) and @gjoseph92's nice write up over in https://github.com/dask/distributed/issues/4640). One thing this PR is proposing is that `Server` is always at the bottom of the mro for any classes which inherit from `Server`. This is already the case today and we could add a test to make sure continues to be true in the future.

fjetter: In this specific case, we need it for `SchedulerState` and `ServerNode` but not for `Server`. Our case is actually even a tiny bit simpler since we do not have a diamond

```
Scheduler -> SchedulerState -> object
Scheduler -> ServerNode -> Server -> object
```

with mro


```python
In [1]: from distributed import Scheduler

In [2]: Scheduler.mro()
Out[2]:
[distributed.scheduler.Scheduler,
 distributed.scheduler.SchedulerState,
 distributed.node.ServerNode,
 distributed.core.Server,
 object]
```

fjetter:
One thing which came up in two comments is to add test about whether `Server` has a superclass other than `object`. I feel this is overengineered and I would expect this to pop up quite naturally if somebody constructs a more complicated inheritance. 
IMHO if somebody wants to create more complicated inheritance model I believe this would not go unnoticed. That would be a major change in the software design and I guess more often than not the [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) principle should then rather guide us than introducing a deeper, branched off inheritance.

However, if people feel this is useful, I don't mind adding a test, that should be easy enough

distributed/core.py
jakirkham: Oops sorry for the 2nd thread 😬

Unfortunately with a Cythonized class I think we need it first as opposed to second. Trying the latter resulted in compilation errors. That said, one of the things I've discussed with Matt before is refactoring `SchedulerState` into its own object (to make inheritance unneeded), which should be useful for a few reasons. Though this is a bit more involved change so may not happen right away and could uncover other issues

jrbourbeau: > Unfortunately with a Cythonized class I think we need it first as opposed to second. Trying the latter resulted in compilation errors

Sorry, I don't quite follow this point. Given that we're cythonizing the scheduler in some CI builds, I suspect the current class structure we have today is okay (though I could be missing something)

fjetter:
There are some related CI failures on OSX I'll need to check out :(

distributed/nanny.py
```diff
@@ -718,88 +731,108 @@ def _run(
         config,
         Worker,
     ):  # pragma: no cover
-        os.environ.update(env)
-        dask.config.set(config)
         try:
```
fjetter, line 734: This big diff is mostly wrapping the function in a try/except

distributed/nanny.py
```diff
@@ -718,88 +731,108 @@ def _run(
...
-                            "address": worker.address,
-                            "dir": worker.local_directory,
-                            "uid": uid,
-                        }
+                    await worker.close(
+                        report=True,
+                        nanny=False,
+                        safe=True,  # TODO: Graceful or not?
+                        executor_wait=executor_wait,
+                        timeout=timeout,
```
fjetter, line 759: This is new

distributed/nanny.py
```diff
@@ -718,88 +731,108 @@ def _run(
...
-            # do_stop() explicitly.
-            loop.run_sync(do_stop)
+                    # If we hit an exception here we need to wait for a least
+                    # one interval for the outside to pick up this message.
+                    # Otherwise we arrive in a race condition where the process
+                    # cleanup wipes the queue before the exception can be
+                    # properly handled. See also
+                    # WorkerProcess._wait_until_connected (the 2 is for good
+                    # measure)
+                    sync_sleep(cls._init_msg_interval * 2)
```
fjetter, line 801: This resolves the race condition in `test_worker_start_exception`

distributed/nanny.py
```diff
@@ -718,88 +731,108 @@ def _run(
...
+        except Exception as e:
+            logger.exception("Failed to initialize Worker")
+            init_result_q.put({"uid": uid, "exception": e})
+            init_result_q.close()
+            # If we hit an exception here we need to wait for a least one
+            # interval for the outside to pick up this message. Otherwise we
+            # arrive in a race condition where the process cleanup wipes the
+            # queue before the exception can be properly handled. See also
+            # WorkerProcess._wait_until_connected (the 2 is for good measure)
+            sync_sleep(cls._init_msg_interval * 2)
```
fjetter, line 828: This resolves the race condition in test_failure_during_worker_initialization

distributed/core.py
jakirkham: All I meant was `SchedulerState` needs to have higher precedence. I don't think anything here is changing that (was just unclear on whether changes to this order were planned)

IOW the code as changed seem ok to me

jakirkham:
I haven't looked at the Nanny or Worker code, but the constructor change seems ok to me. Updated that thread and marked resolved

jrbourbeau (approved):
Thanks @fjetter! This is in

PR #4734, opened 2021-04-22
Ensure exceptions in handlers are handled equally for sync and async

fjetter:
I do not have a strong opinion at the moment about what should happen if a handler raises an exception but regardless, I believe an async and sync handler should show the same behaviour. Everything else is pretty hard to distinguish for any developer.

One easy way to achieve this is to simply await the handler during handle_stream. I am not sure if it was implemented this way hoping that messages would be faster worked off or what the motivation behind this was.
If we want to stick to the old `add_callback` we need to implement exception handling on the returned future/task.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -14,7 +14,6 @@ from functools import partial
 
 import tblib
 from tlz import merge
-from tornado import gen
 from tornado.ioloop import IOLoop, PeriodicCallback
 
 import dask
@@ -37,6 +36,7 @@ from .utils import (
     get_traceback,
     has_keyword,
     is_coroutine_function,
+    shielded,
     truncate_exception,
 )
 
@@ -161,6 +161,7 @@ class Server:
         self.counters = None
         self.digests = None
         self._ongoing_coroutines = weakref.WeakSet()
+
         self._event_finished = asyncio.Event()
 
         self.listeners = []
@@ -264,10 +265,18 @@ class Server:
 
     def __await__(self):
         async def _():
+            if self.status == Status.running:
+                return self
+
+            if self.status in (Status.closing, Status.closed):
+                # We should never await an object which is already closing but
+                # we should also not start it up again otherwise we'd produce
+                # zombies
+                await self.finished()
+                return
+
             timeout = getattr(self, "death_timeout", 0)
             async with self._startup_lock:
-                if self.status == Status.running:
-                    return self
                 if timeout:
                     try:
                         await asyncio.wait_for(self.start(), timeout=timeout)
@@ -426,7 +435,7 @@ class Server:
 
         logger.debug("Connection from %r to %s", address, type(self).__name__)
         self._comms[comm] = op
-        await self
+
         try:
             while True:
                 try:
@@ -569,17 +578,15 @@ class Server:
                                 break
                             handler = self.stream_handlers[op]
                             if is_coroutine_function(handler):
-                                self.loop.add_callback(handler, **merge(extra, msg))
-                                await gen.sleep(0)
+                                await handler(**merge(extra, msg))
                             else:
                                 handler(**merge(extra, msg))
                         else:
                             logger.error("odd message %s", msg)
-                    await asyncio.sleep(0)
 
                 for func in every_cycle:
                     if is_coroutine_function(func):
-                        self.loop.add_callback(func)
+                        await func()
                     else:
                         func()
 
@@ -597,32 +604,39 @@ class Server:
             await comm.close()
             assert comm.closed()
 
-    @gen.coroutine
-    def close(self):
+    @shielded
+    async def close(self):
+        self.status = Status.closing
+        self.stop()
+        await self.rpc.close()
+
         for pc in self.periodic_callbacks.values():
             pc.stop()
         self.__stopped = True
         for listener in self.listeners:
             future = listener.stop()
             if inspect.isawaitable(future):
-                yield future
-        for i in range(20):
+                await future
+        for _ in range(20):
             # If there are still handlers running at this point, give them a
             # second to finish gracefully themselves, otherwise...
             if any(self._comms.values()):
-                yield asyncio.sleep(0.05)
+                await asyncio.sleep(0.05)
             else:
                 break
-        yield [comm.close() for comm in list(self._comms)]  # then forcefully close
+        await asyncio.gather(
+            *[comm.close() for comm in list(self._comms)]
+        )  # then forcefully close
         for cb in self._ongoing_coroutines:
             cb.cancel()
-        for i in range(10):
+        for _ in range(10):
             if all(c.cancelled() for c in self._ongoing_coroutines):
                 break
             else:
-                yield asyncio.sleep(0.01)
+                await asyncio.sleep(0.01)
 
         self._event_finished.set()
+        self.status = Status.closed
 
 
 def pingpong(comm):
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -293,9 +293,12 @@ class Server:
 
         return _().__await__()
 
+    async def _start(self):
+        self.status = Status.running
+
     async def start(self):
         await self.rpc.start()
-        self.status = Status.running
+        await self._start()
 
     async def __aenter__(self):
         await self
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -438,6 +438,10 @@ class Server:
         logger.debug("Connection from %r to %s", address, type(self).__name__)
         self._comms[comm] = op
 
+        # The server might already be listening even though it is not properly
+        # started, yet. (e.g. preload modules not done)
+        await self
+
         try:
             while True:
                 try:
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -275,20 +275,18 @@ class Server:
             timeout = getattr(self, "death_timeout", None)
             if self._startup_fut is None:
                 self._startup_fut = asyncio.ensure_future(
-                    asyncio.wait_for(self.start(), timeout=timeout)
+                    asyncio.wait_for(self._start(), timeout=timeout)
                 )
+            await self.rpc.start()
 
             try:
                 await self._startup_fut
             except Exception:
-                # This timeout is very arbitrary but the close method itself
-                # already may wait more than a second to close gracefully.
-                # before we interrupt we should give it enough time to finish
-
+                await self.rpc.close()
                 # Suppress all exception since the objects might not have been
                 # properly initialized for close to be successful.
                 with suppress(Exception):
-                    await asyncio.wait_for(self.close(), timeout=2)
+                    await self.close()
                 if timeout:
                     raise TimeoutError(
                         f"{type(self).__name__} failed to start in {timeout} seconds"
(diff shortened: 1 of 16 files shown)


fjetter:
Test failure in `test_no_danglng_asyncio_tasks` is a genuine failure and I can reproduce locally

distributed/core.py
```diff
@@ -594,29 +592,33 @@ async def handle_stream(self, comm, extra=None, every_cycle=[]):
...
             if all(c.cancelled() for c in self._ongoing_coroutines):
                 break
             else:
-                yield asyncio.sleep(0.01)
+                await asyncio.sleep(0.01)
 
         self._event_finished.set()
+        # Allow event listeners a cycle to react before we return
+        # See also https://github.com/dask/distributed/pull/4734
+        await asyncio.sleep(0)
```
fjetter, line 621: I received errors with still pending tasks from the worker after shutdown

```
  <Task pending name='Task-41' coro=<Worker.ensure_computing() running at
```

letting the loop run for another cycle here seemed to do the trick but frankly, I don't understand why. Especially not what this has to do with my proposed changes. Is anybody able to shed light on this?

mrocklin: I think that `await asyncio.sleep(0)` may not actually let the event loop cycle.  I think that `sleep(0)` might special-case and continue directly with the same coroutine.  You might want to try something like `sleep(1e-6)` to see if that makes a difference.

fjetter: It is a special case but it does allow one cycle, see https://github.com/python/cpython/blob/accea7dc2bd30a6e8e1b0334acfca9585cbd7f8a/Lib/asyncio/tasks.py#L585-L601 

If this doesn't work I'll try a fixed sleep but this feels fragile

Do you have an idea how this could be connected?

mrocklin: I'm not sure.

Maybe there was a final task submission that previously was ignored but is now actually submitted?

distributed/core.py
```diff
@@ -577,7 +575,7 @@ async def handle_stream(self, comm, extra=None, every_cycle=[]):
 
                 for func in every_cycle:
                     if is_coroutine_function(func):
-                        self.loop.add_callback(func)
```
fjetter, line 580: I tracked down the dangling task and this was the `ensure_compute` scheduled here. In particular it concerned the `ensure_computing` which uses the async interface to offload deserialisation. This was introduced in https://github.com/dask/distributed/pull/4307 just before the infamous `2020.12.0` release

fjetter:
This situation is a bit awkward and I see two options

1. We await both handler and every-cycle-callbacks and let asyncio/python deal with the exceptions themselves. This puts us at risk of having a slow handler or slow cycle-callback. This could, for instance, cause the worker to accumulate messages in the queue faster than it could work off. Even worse, I believe there is a chance for the comm / tcp socket to overflow which either results in weird exceptions or dropped messages, I'd need to research this a bit. In either case this would probably only happen if there is a quite significant load on the entire system in which case I'd wonder if we're coming out of this alive, anyhow. Not an expert though.
2. We implement the task offloading to the loop ourselves and take care that the exceptions are properly handled. The way tornado deals with this is not 100% well suited to our use case here. In the cases where we put tasks on the loop, we often would like to check if they completed, erred or cancel them even. Currently we do not have this possibility.

Note: this problem pops up in many places in the worker so solving 2.) once might benefit other situations as well.

@mrocklin Any thoughts on this? Do we have any asyncio/tornado experts we can loop into this discussion?

---

Note: The PR grew again in scope. I ignored the warning in the gen.coroutine comment and figured it was easy to rewrite. Turns out this introduces race conditions during closing since the behaviour of gen.coroutine and plain async coroutines differ in the way they are cancelled (or not). Will break this off into another PR

mrocklin:
> This could, for instance, cause the worker to accumulate messages in the queue faster than it could work off

I'm not sure that this is true.  There is a handle_stream coroutine for every comm.  Comms are most often used with the `rpc` objects as follows:

```python
## caller side
await worker.some_handler(arg=foo)
```

The caller side generally waits until the call to the handler completes before moving on, so the backup won't be inside the socket on the receiver side, it will be in application code on the caller side, which seems safe.  Previously this call would have completed very quickly and now it will take longer.  That's probably ok.

In summary, if things aren't breaking here I think that we're ok.  In principle this change feels safe to me.  I wouldn't be surprised if it did break some things, but even if it did, those things were probably broken beforehand and relying on bad logic in the worker.  For example, if the  caller side was doing the following:

```python
await worker.some_handler(arg=1)
await worker.some_handler(arg=2)
await worker.some_handler(arg=3)
await worker.some_handler(arg=4)
```

Then with async handlers this would have previously run much more quickly than it should have.  In a situation like the following:

```python
await asyncio.gather(
    worker.some_handler(arg=1),
    worker.some_handler(arg=2),
    ...
)
```

The caller side will actually open many comms to the server to handle the requests in parallel, and so we'll get many `handle_stream` coroutines, and everything should behave nicely.

(Usual disclaimer of "I'm a bit rusty on all of this, so please don't believe me")

fjetter:
> I'm not sure that this is true. There is a handle_stream coroutine for every comm. Comms are most often used with the rpc objects as follows:

I'm mostly concerned about the primary link between scheduler and worker which is based on a single comm. Most of the communication between the two is done via this single comm using a `BatchedSend`. For instance all "compute-task", "forget-task" msgs are moving along this single comm and I'm wondering if this becomes an issue.

However, with the exception of `Worker.close` all of our stream handlers seem to be sync anyhow, so I'm worrying unnecessarily 🤦

mrocklin:
> However, with the exception of Worker.close all of our stream handlers seem to be sync anyhow, so I'm worrying unnecessarily

Yeah, anything sent by a BatchedSend is intended to run immediately.  

If you wanted to be careful though you could add a read into the `BatchedSend._background_send` method.  This would close the loop and establish back pressure.  Really we would want to allow a few of these to be in flight before awaiting.  I'm not too worried about this though.

distributed/core.py
```diff
@@ -422,7 +430,7 @@ async def handle_comm(self, comm):
 
         logger.debug("Connection from %r to %s", address, type(self).__name__)
         self._comms[comm] = op
-        await self
```
fjetter, line 425: This `await self` without the guard I introduced in `__await__` would allow an incoming comm to restart a currently shutting down scheduler. There is a time window while the scheduler is handling its affairs, before the actual server closes, where it still accepts incoming comms. If during this time a new comm is opened, the scheduler is restarted although it is currently trying to shutdown.

https://github.com/dask/distributed/blob/d9bc3c634a4a3983cdc75e5c39cbdac2afbf6184/distributed/scheduler.py#L3759-L3808

fjetter: I didn't add a dedicated test for this since there are actually many failing tests once the `close` coroutines are shielded

distributed/utils.py
```diff
@@ -202,6 +203,22 @@ def get_ip_interface(ifname):
...
+    function is invoked.
+
+    See also https://docs.python.org/3/library/asyncio-task.html#asyncio.shield
+    """
+
+    @wraps(func)
+    def _(*args, **kwargs):
+        return asyncio.shield(func(*args, **kwargs))
+
+    return _
```
fjetter, line 219: This is not absolutely required but with this decorator we can define methods which are always shielded (e.g. close) instead of relying on remembering it on every invocation

distributed/deploy/spec.py
```diff
@@ -338,15 +338,8 @@ async def _correct_state_internal(self):
             if to_close:
                 if self.scheduler.status == Status.running:
                     await self.scheduler_comm.retire_workers(workers=list(to_close))
-                tasks = [
-                    asyncio.create_task(self.workers[w].close())
-                    for w in to_close
-                    if w in self.workers
-                ]
+                tasks = [self.workers[w].close() for w in to_close if w in self.workers]
```
mrocklin, line 341: Can we call `asyncio.wait` here and avoid the tasks variable entirely?

distributed/nanny.py
```diff
@@ -506,18 +509,15 @@ async def close(self, comm=None, timeout=5, report=None):
...
             await preload.teardown()
 
-        self.stop()
         try:
             if self.process is not None:
                 await self.kill(timeout=timeout)
         except Exception:
             pass
         self.process = None
-        await self.rpc.close()
```
mrocklin, line 516: Why this change?

distributed/nanny.py
```diff
@@ -506,18 +509,15 @@ async def close(self, comm=None, timeout=5, report=None):
         for preload in self.preloads:
             await preload.teardown()
 
-        self.stop()
```
mrocklin, line 509: Why this change?

distributed/scheduler.py
```diff
@@ -3808,10 +3809,6 @@ async def close(self, comm=None, fast=False, close_workers=False):
         for comm in self.client_comms.values():
             comm.abort()
 
-        await self.rpc.close()
-
-        self.status = Status.closed
-        self.stop()
```
mrocklin, line 3814: Why this change?

distributed/scheduler.py
```diff
@@ -6257,7 +6259,10 @@ async def retire_workers(
                     )
                 if remove:
                     await asyncio.gather(
-                        *[self.remove_worker(address=w, safe=True) for w in worker_keys]
+                        *[
+                            self.remove_worker(address=w, safe=True, close=False)
+                            for w in worker_keys
+                        ]
```
mrocklin, line 6265: Why this change?

mrocklin (commented):
I'm curious about the reason for a few of the changes here.  Also, do we know why tests aren't happy yet?

distributed/deploy/spec.py
fjetter: sure

distributed/nanny.py
fjetter: It is redundant since we do the same in `Server.close`. it looked like being redundant and I felt removing it would reduce complexity. If there is some purpose to stopping earlier, I can revert this logic. This is only my sense of _cleaning up_

distributed/scheduler.py
fjetter: this is performed by `super().close()` since there are no other things happening in between, letting the server close itself felt like the right thing to do to reduce complexity.

distributed/nanny.py
fjetter: same argument as above about cleaning up

distributed/scheduler.py
fjetter: This change adds `close=False` (default is True). This causes in most situations to issue a close_worker call twice since there is a dedicated close_worker call above. 

It _might_ be the correct thing to change this to `close=not close_workers` to avoid the duplication

jrbourbeau:
Sorry @fjetter, it looks like https://github.com/dask/distributed/pull/4966 caused some merge conflicts here

fjetter:
I'm hitting a lot of known "flaky" tests here (cannot reproduce locally, yet)

distributed/deploy/tests/test_spec_cluster.py
```diff
@@ -416,6 +415,9 @@ def __str__(self):
 
     __repr__ = __str__
 
+    def __await__(self):
+        return self.start().__await__()
+
```
fjetter, line 420: This `MultiWorker` is a bit strange since it isn't a Server although `Worker`s are always `Server`s

PR #4739, opened 2021-04-23
Reduce complexity of test_gather_allow_worker_reconnect

fjetter:
This test was originally authored by myself and was marked flaky which is because the sleep time is not sufficiently long. This should address this. I would prefer to have as few of the flaky marks as possible in our test base and would suggest to remove it again.

I also reduced the complexity by removing retry logic here


fjetter:
Test failures are https://github.com/dask/distributed/issues/3574 and https://github.com/dask/distributed/issues/4740

fjetter:
Test failure is unrelated. Since I am the original author of this test, I'll just go ahead and merge this

PR #4741, opened 2021-04-23
Pyupgrade

fjetter:
I personally **love** automated formatting and I discovered this gem in the pandas pre-commit hooks. While deprecations are less common in py3 (opposed to the py2->py3 migration) they do happen. Often this boils down to slightly less verbose syntax, other times exception classes or functions are renamed, etc.

Some changes are merely for style while others, e.g. around exception classes, imho clarify things better, e.g. what's the distinction between `OSError`, `IOError`, `EnvironmentError`, `socket.error`? Nothing... since [py3.3](https://docs.python.org/3/library/exceptions.html#OSError)

Is this kind of thing something people would feel comfortable with or is this too much?

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -7,7 +7,6 @@ import inspect
 import json
 import logging
 import os
-import socket
 import sys
 import threading
 import uuid
@@ -399,9 +398,9 @@ class Future(WrappedKey):
                 typ = self.type.__module__.split(".")[0] + "." + self.type.__name__
             except AttributeError:
                 typ = str(self.type)
-            return "<Future: %s, type: %s, key: %s>" % (self.status, typ, self.key)
+            return f"<Future: {self.status}, type: {typ}, key: {self.key}>"
         else:
-            return "<Future: %s, key: %s>" % (self.status, self.key)
+            return f"<Future: {self.status}, key: {self.key}>"
 
     def _repr_html_(self):
         text = "<b>Future: %s</b> " % html.escape(key_split(self.key))
@@ -494,7 +493,7 @@ class FutureState:
         await asyncio.wait_for(self._get_event().wait(), timeout)
 
     def __repr__(self):
-        return "<%s: %s>" % (self.__class__.__name__, self.status)
+        return f"<{self.__class__.__name__}: {self.status}>"
 
 
 async def done_callback(future, callback):
@@ -660,9 +659,7 @@ class Client:
                 logger.info("Config value `scheduler-address` found: %s", address)
 
         if address is not None and kwargs:
-            raise ValueError(
-                "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
-            )
+            raise ValueError(f"Unexpected keyword arguments: {str(sorted(kwargs))}")
 
         if isinstance(address, (rpc, PooledRPCCall)):
             self.scheduler = address
@@ -907,12 +904,12 @@ class Client:
             return text
 
         elif self.scheduler is not None:
-            return "<%s: scheduler=%r>" % (
+            return "<{}: scheduler={!r}>".format(
                 self.__class__.__name__,
                 self.scheduler.address,
             )
         else:
-            return "<%s: No scheduler connected>" % (self.__class__.__name__,)
+            return f"<{self.__class__.__name__}: No scheduler connected>"
 
     def _repr_html_(self):
         scheduler, info = self._get_scheduler_info()
@@ -1073,7 +1070,7 @@ class Client:
                     asynchronous=self._asynchronous,
                     **self._startup_kwargs,
                 )
-            except (OSError, socket.error) as e:
+            except OSError as e:
                 if e.errno != errno.EADDRINUSE:
                     raise
                 # The default port was taken, use a random one
@@ -1123,7 +1120,7 @@ class Client:
                 try:
                     await self._ensure_connected(timeout=timeout)
                     break
-                except EnvironmentError:
+                except OSError:
                     # Wait a bit before retrying
                     await asyncio.sleep(0.1)
                     timeout = deadline - self.loop.time()
@@ -1203,7 +1200,7 @@ class Client:
             return
         try:
             self._scheduler_identity = SchedulerInfo(await self.scheduler.identity())
-        except EnvironmentError:
+        except OSError:
             logger.debug("Not able to query scheduler for identity")
 
     async def _wait_for_workers(self, n_workers=0, timeout=None):
--- a/distributed/comm/core.py
+++ b/distributed/comm/core.py
@@ -157,9 +157,9 @@ class Comm(ABC):
     def __repr__(self):
         clsname = self.__class__.__name__
         if self.closed():
-            return "<closed %s>" % (clsname,)
+            return f"<closed {clsname}>"
         else:
-            return "<%s %s local=%s remote=%s>" % (
+            return "<{} {} local={} remote={}>".format(
                 clsname,
                 self.name or "",
                 self.local_address,
@@ -307,7 +307,7 @@ async def connect(
             )
             await asyncio.sleep(backoff)
     else:
-        raise IOError(
+        raise OSError(
             f"Timed out trying to connect to {addr} after {timeout} s"
         ) from active_exception
 
@@ -323,7 +323,7 @@ async def connect(
     except Exception as exc:
         with suppress(Exception):
             await comm.close()
-        raise IOError(
+        raise OSError(
             f"Timed out during handshake while connecting to {addr} after {timeout} s"
         ) from exc
 
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -164,9 +164,11 @@ class Occupancy(DashboardComponent):
                     color.append("blue")
 
             if total:
-                self.root.title.text = "Occupancy -- total time: %s  wall time: %s" % (
-                    format_time(total),
-                    format_time(total / self.scheduler.total_nthreads),
+                self.root.title.text = (
+                    "Occupancy -- total time: {}  wall time: {}".format(
+                        format_time(total),
+                        format_time(total / self.scheduler.total_nthreads),
+                    )
                 )
             else:
                 self.root.title.text = "Occupancy"
@@ -2640,9 +2642,7 @@ class WorkerTable(DashboardComponent):
 
         for name in self.names + self.extra_names:
             if name == "name":
-                data[name].insert(
-                    0, "Total ({nworkers})".format(nworkers=len(data[name]))
-                )
+                data[name].insert(0, f"Total ({len(data[name])})")
                 continue
             try:
                 if len(self.scheduler.workers) == 0:
--- a/distributed/deploy/old_ssh.py
+++ b/distributed/deploy/old_ssh.py
@@ -100,7 +100,7 @@ def async_ssh(cmd_dict):
             print(
                 "               "
                 + bcolors.FAIL
-                + "Retrying... (attempt {n}/{total})".format(n=retries, total=3)
+                + f"Retrying... (attempt {retries}/{3})"
                 + bcolors.ENDC
             )
 
@@ -152,7 +152,7 @@ def async_ssh(cmd_dict):
                 cmd_dict["output_queue"].put(
                     "[ {label} ] : ".format(label=cmd_dict["label"])
                     + bcolors.FAIL
-                    + "{output}".format(output=line)
+                    + f"{line}"
                     + bcolors.ENDC
                 )
                 line = stderr.readline()
@@ -215,18 +215,14 @@ def start_scheduler(
 
     # Optionally re-direct stdout and stderr to a logfile
     if logdir is not None:
-        cmd = "mkdir -p {logdir} && ".format(logdir=logdir) + cmd
+        cmd = f"mkdir -p {logdir} && " + cmd
         cmd += "&> {logdir}/dask_scheduler_{addr}:{port}.log".format(
             addr=addr, port=port, logdir=logdir
         )
 
     # Format output labels we can prepend to each line of output, and create
     # a 'status' key to keep track of jobs that terminate prematurely.
-    label = (
-        bcolors.BOLD
-        + "scheduler {addr}:{port}".format(addr=addr, port=port)
-        + bcolors.ENDC
-    )
+    label = bcolors.BOLD + f"scheduler {addr}:{port}" + bcolors.ENDC
 
     # Create a command dictionary, which contains everything we need to run and
     # interact with this command.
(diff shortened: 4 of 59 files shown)


distributed/_version.py
```diff
@@ -84,7 +84,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
                 stderr=(subprocess.PIPE if hide_stderr else None),
             )
             break
-        except EnvironmentError:
+        except OSError:
             e = sys.exc_info()[1]
             if e.errno == errno.ENOENT:
```
jakirkham, line 89: Maybe this could be simplified like so?

```suggestion
        except OSError as e:
            if e.errno == errno.ENOENT:
```

distributed/client.py
```diff
@@ -1027,7 +1024,7 @@ async def _start(self, timeout=no_default, **kwargs):
                     asynchronous=self._asynchronous,
                     **self._startup_kwargs,
                 )
-            except (OSError, socket.error) as e:
+            except OSError as e:
```
jakirkham, line 1027: Are we sure this should be dropped?

distributed/comm/core.py
```diff
@@ -320,7 +320,7 @@ def time_left():
     except Exception as exc:
         with suppress(Exception):
             await comm.close()
-        raise IOError(
+        raise OSError(
```
jakirkham, line 323: Did we intend for these exceptions to be changed?

jakirkham:
There are a few places where this tooling seems to be changing the exceptions we are raising. This seems a little unexpected and may have surprising consequences. Is there a way to disable that?

distributed/client.py
jrbourbeau: This should be okay as `IOError`, `EnvironmentError`, `socket.error`, and a few other error types were all unified in Python 3.3 

```python
Python 3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:20)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import socket

In [2]: OSError is socket.error
Out[2]: True
```

I think this type of simplification is what `pyupgrade` helps with

jacobtomlinson (commented):
This looks really useful. 

I especially enjoyed seeing the `"%s" % var` and `"{}".format(var)` calls being upgraded to use f-strings. Although it is interesting to note that not everything is being switched to f-string. Some things with more complex expressions seem to be stopping at `.format()`. Perhaps we want to refactor those so that everything is f-string?

Seeing the exceptions being changed was a little surprising. But it's my understanding that these days `OSError is IOError is EnvironmentError is socket.error` so switching them all out to be `OSError` makes sense.

The only other change I really noticed was `set([1, 2, 3])` being replaced with `{1, 2, 3}`.

My one concern about this change is adding it to the CI via the pre-comimt check. Expecting every contributor to always write 100% modern Python just adds a little more friction. We experience this today with folks not running Black. Unless everyone uses pre-commit (which is opt-in) they will have to also run `pyupgrade` themselves, or maintainers will need to push another commit for them. This is only a small concern, but wanted to raise it.

fjetter:
> Some things with more complex expressions seem to be stopping at .format(). Perhaps we want to refactor those so that everything is f-string?

Yes, I guess the magic is limited to some extend. Maybe this is only a bug, or the AST is, for whatever reason, too complex, idk. If it boils down to a few remaining `format` calls, I could live with it

> There are a few places where this tooling seems to be changing the exceptions we are raising. This seems a little unexpected and may have surprising consequences. Is there a way to disable that?

@jakirkham The exceptions we are raising are indeed altered but according to the python docs these are identical to a `OSError` since python 3.3, see https://docs.python.org/3/library/exceptions.html#OSError

I think it is not possible to disable this https://github.com/asottile/pyupgrade#oserror-aliases

> Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and mmap.error have been merged into OSError, and the constructor may return a subclass.


The referred subclasses are listed in https://docs.python.org/3/library/exceptions.html#os-exceptions and include things like `ConnectionError`, `BrokenPipeError`, `TimeoutError` and many more but the above mentioned exceptions are aliases to OSError now, e.g.

```python
# Python 3.8.8
In [2]: OSError is IOError
Out[2]: True

In [3]: IOError is OSError
Out[3]: True

In [4]: IOError == OSError
Out[4]: True

In [5]: OSError == IOError
Out[5]: True
```


> Unless everyone uses pre-commit (which is opt-in) they will have to also run pyupgrade themselves, or maintainers will need to push another commit for them. This is only a small concern, but wanted to raise it. Perhaps we should just run this periodically ourselves instead of making it a hard requirement?

That is a valid concern but I figured this was already the case for black. IMHO it is the same for isort and flake8. While flake8 does not auto-format things, I personally am not able to simply write flake8 compatible code, if only for too long comments, etc.
If this is of concern, we could add a script which calls the hooks. Alternatively, `pre-commit run (-a)` also does the trick without actually installing the commit hook, afaik.

distributed/_version.py
fjetter: definitely but I didn't want to actually change things in here manually. This PR is merely for the automation part

fjetter:
If this is accepted, I'll recreate the last commit which applies pyupgrade. No need for merging.

mrocklin:
This seems fine to me

jrbourbeau (commented):
Yeah, agreed that this is fine by me. At some point I'll want to update our contributing docs to say "use `pre-commit` for code styling" instead of recommending it as an optional nice-to-have

distributed/client.py
```diff
@@ -645,9 +644,7 @@ def __init__(
                 logger.info("Config value `scheduler-address` found: %s", address)
 
         if address is not None and kwargs:
-            raise ValueError(
-                "Unexpected keyword arguments: {}".format(str(sorted(kwargs)))
-            )
+            raise ValueError(f"Unexpected keyword arguments: {str(sorted(kwargs))}")
```
crusaderky, line 647:
```suggestion
            raise ValueError(f"Unexpected keyword arguments: {sorted(kwargs)}")
```

distributed/dashboard/components/scheduler.py
```diff
@@ -154,9 +154,11 @@ def update(self):
                     color.append("blue")
 
             if total:
-                self.root.title.text = "Occupancy -- total time: %s  wall time: %s" % (
-                    format_time(total),
-                    format_time(total / self.scheduler.total_nthreads),
+                self.root.title.text = (
+                    "Occupancy -- total time: {}  wall time: {}".format(
+                        format_time(total),
+                        format_time(total / self.scheduler.total_nthreads),
```
crusaderky, line 160:
```suggestion
                    f"Occupancy -- total time: {format_time(total)} "
                    f"wall time: {format_time(total / self.scheduler.total_nthreads)}"
```

distributed/deploy/old_ssh.py
```diff
@@ -100,7 +100,7 @@ def async_ssh(cmd_dict):
             print(
                 "               "
                 + bcolors.FAIL
-                + "Retrying... (attempt {n}/{total})".format(n=retries, total=3)
+                + f"Retrying... (attempt {retries}/{3})"
```
crusaderky, line 103:
```suggestion
                + f"Retrying... (attempt {retries}/3)"
```

distributed/deploy/old_ssh.py
```diff
@@ -152,7 +152,7 @@ def read_from_stderr():
                 cmd_dict["output_queue"].put(
                     "[ {label} ] : ".format(label=cmd_dict["label"])
                     + bcolors.FAIL
-                    + "{output}".format(output=line)
+                    + f"{line}"
```
crusaderky, line 155:
```suggestion
                    + line
```

fjetter:
@crusaderky I'll update the PR but it is probably not worth it to review individual lines. This is automatically changed. I'll rebase and we can have a look

distributed/deploy/old_ssh.py
```diff
@@ -215,18 +215,14 @@ def start_scheduler(
 
     # Optionally re-direct stdout and stderr to a logfile
     if logdir is not None:
-        cmd = "mkdir -p {logdir} && ".format(logdir=logdir) + cmd
+        cmd = f"mkdir -p {logdir} && " + cmd
```
crusaderky, line 218:
```suggestion
        cmd = f"mkdir -p {logdir} && {cmd}"
```

distributed/deploy/old_ssh.py
```diff
@@ -215,18 +215,14 @@ def start_scheduler(
...
         )
 
     # Format output labels we can prepend to each line of output, and create
     # a 'status' key to keep track of jobs that terminate prematurely.
-    label = (
-        bcolors.BOLD
-        + "scheduler {addr}:{port}".format(addr=addr, port=port)
-        + bcolors.ENDC
-    )
+    label = bcolors.BOLD + f"scheduler {addr}:{port}" + bcolors.ENDC
```
crusaderky, line 225:
```suggestion
    label = f"{bcolors.BOLD}scheduler {addr}:{port}{bcolors.ENDC}"
```

distributed/deploy/old_ssh.py
```diff
@@ -309,12 +305,12 @@ def start_worker(
 
     # Optionally redirect stdout and stderr to a logfile
     if logdir is not None:
-        cmd = "mkdir -p {logdir} && ".format(logdir=logdir) + cmd
+        cmd = f"mkdir -p {logdir} && " + cmd
```
crusaderky, line 308:
```suggestion
        cmd = f"mkdir -p {logdir} && {cmd}"
```

distributed/tests/test_scheduler.py
```diff
@@ -1983,10 +1983,10 @@ def abort(self):
         pass
 
     def read(self, deserializers=None):
-        raise EnvironmentError
+        raise OSError
```
crusaderky, line 1986:
```suggestion
        raise OSError()
```

distributed/tests/test_scheduler.py
```diff
@@ -1983,10 +1983,10 @@ def abort(self):
...
 
     def write(self, msg, serializers=None, on_error=None):
-        raise EnvironmentError
+        raise OSError
```
crusaderky, line 1989:
```suggestion
        raise OSError()
```

distributed/tests/test_semaphore.py
```diff
@@ -293,10 +293,10 @@ def abort(self):
         pass
 
     def read(self, deserializers=None):
-        raise EnvironmentError
+        raise OSError
```
crusaderky, line 296:
```suggestion
        raise OSError()
```

distributed/tests/test_semaphore.py
```diff
@@ -293,10 +293,10 @@ def abort(self):
...
 
     def write(self, msg, serializers=None, on_error=None):
-        raise EnvironmentError
+        raise OSError
```
crusaderky, line 299:
```suggestion
        raise OSError()
```

distributed/tests/test_steal.py
```diff
@@ -537,7 +537,7 @@ async def assert_balanced(inp, expected, c, s, *workers):
 
         if result2 == expected2:
             return
-    raise Exception("Expected: {}; got: {}".format(str(expected2), str(result2)))
+    raise Exception(f"Expected: {str(expected2)}; got: {str(result2)}")
```
crusaderky, line 540:
```suggestion
    raise Exception(f"Expected: {expected2}; got: {result2}")
```

distributed/tests/test_utils_comm.py
```diff
@@ -56,10 +56,10 @@ def abort(self):
         pass
 
     def read(self, deserializers=None):
-        raise EnvironmentError
+        raise OSError
```
crusaderky, line 59:
```suggestion
        raise OSError()
```

distributed/tests/test_utils_comm.py
```diff
@@ -56,10 +56,10 @@ def abort(self):
...
 
     def write(self, msg, serializers=None, on_error=None):
-        raise EnvironmentError
+        raise OSError
```
crusaderky, line 62:
```suggestion
        raise OSError()
```

fjetter:
I just rebased and updated. We can either

* merge as is and update contribution guidelines (where can I find that?)
* remove pyupgrade from pre-commit and run this occasionally but not enforce it
* do nothing

I suggest to do further refinements in a follow up PR

distributed/_version.py
```diff
@@ -84,7 +84,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
                 stderr=(subprocess.PIPE if hide_stderr else None),
```
crusaderky, line 84: This file is auto-generated by versioneer and IMHO should be left untouched and excluded from pyupgrade

crusaderky:
> remove pyupgrade from pre-commit and run this occasionally but not enforce it

I think pyupgrade in the pre-commit is a bit overkill

crusaderky (changes requested):

distributed/_version.py
fjetter: I'll revert. FWIW we already formatted the file in a black formatting step ~2years ago

distributed/scheduler.py
```diff
@@ -7204,8 +7204,8 @@ def adaptive_target(self, comm=None, target_duration=None):
             cpu = max(1, cpu)
 
         # add more workers if more than 60% of memory is used
-        limit = sum([ws._memory_limit for ws in parent._workers_dv.values()])
-        used = sum([ws._nbytes for ws in parent._workers_dv.values()])
+        limit = sum(ws._memory_limit for ws in parent._workers_dv.values())
```
fjetter, line 7207: This seems to cause some problems with cythonization

distributed/tests/test_scheduler.py
fjetter: out of curiosity, is this a style note or is there a functional difference?

crusaderky: it's only a style note. ``except OSError as e`` will set e to an instance to OSError in both cases.

distributed/scheduler.py
fjetter: Looks like cython doesn't like the generator expressions

fjetter:
I reverted the used pyupgrade version to 2.13.0 since this version worked before. The reformatting of `sum([expression])` to `sum(<genexpression>)` is not performed by this version and that seems to create issues with cythonization

I also removed it from the pre-commit config, i.e. it would not be enforced now.

fjetter:
Based off a brief discussion on our dask maint weekly, we think this will not cause major merge conflicts so I'll go ahead and merge this now.

jrbourbeau:
Thanks @fjetter

PR #4743, opened 2021-04-23
Add context manager to check for dangling tasks

fjetter:
This adds a context manager to `utils_test` which is there to ensure that there are no dangling tasks. The logic is a bit more complex since it ensure that it is actually using the correct loop, i.e. it ensures that there are also not any tasks added to _other_ loops (check for `all_tasks` w/ and w/out loop).
In particular with `pytest-asyncio`, which creates the loop itself, we're sometimes not using the correct loop in our tests and certain assertions are not working. For instance, the `pristine_loop` ctx mgr doesn't work properly with `pytest-asyncio` and therefore the clean/cleanup fixture is not working as intended. One step at a time...

### DIFF
--- a/distributed/tests/test_utils_test.py
+++ b/distributed/tests/test_utils_test.py
@@ -12,6 +12,7 @@ from distributed.core import rpc
 from distributed.metrics import time
 from distributed.utils import get_ip
 from distributed.utils_test import (  # noqa: F401
+    check_dangling_tasks,
     cleanup,
     cluster,
     gen_cluster,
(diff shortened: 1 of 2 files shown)


jrbourbeau (commented):
Thanks @fjetter, this is nice to see. Happy to merge this in as is. Do you want to start using `check_dangling_tasks` elsewhere in this PR, or leave that for follow-up work?

fjetter:
I'm actually not that sure anymore if this logic is even useful and would close this, for now, until I have a concrete use case. I might add the code in a later PR. Sorry for the PR spam :)

PR #4748, opened 2021-04-23
Unregister worker plugin

### LINKED ISSUE #4237: ENH: Add client.unregister_worker_plugin
The `Scheduler` class has `add_plugin` and `remove_plugin` methods for `SchedulerPlugin`s. But for worker plugins there is only `client.register_worker_plugin`. I think it could be useful to add a `client.unregister_worker_plugin`. 

In practice, I have found myself wanting this when I have accidentally created a broken plugin and then need to restart my cluster to get rid of it. It could also be useful if you want to overwrite a plugin with a new one that has the same name.

The argument to `client.unregister_worker_plugins` could be the plugin name.

ncclementi:
This PR adds the possibility to do `client.unregister_worker_plugin`. 
So far we added two tests:
- Given a register plugin it "unregisters" and checks there are no worker plugins (pass)
- Given a non-existing plugin raises KeyError (pass)

But I would like to add one more test that given two plugins we unregister one of them and check that case. 

- [x] Closes #4237
- [x] Tests added / passed

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4055,6 +4055,21 @@ class Client:
 
         return self.sync(self._register_worker_plugin, plugin=plugin, name=name)
 
+    async def _unregister_worker_plugin(self, name):
+        responses = await self.scheduler.unregister_worker_plugin(name=name)
+
+        for response in responses.values():
+            if response["status"] == "error":
+                exc = response["exception"]
+                typ = type(exc)
+                tb = response["traceback"]
+                raise exc.with_traceback(tb)
+        return responses
+
+    def unregister_worker_plugin(self, name):
+        """Unregisters a lifecycle worker plugin"""
+        return self.sync(self._unregister_worker_plugin, name=name)
+
 
 class _WorkerSetupPlugin(WorkerPlugin):
     """ This is used to support older setup functions as callbacks """
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4050,6 +4050,7 @@ class Client:
         See Also
         --------
         distributed.WorkerPlugin
+        unregister_worker_plugin
         """
         if isinstance(plugin, type):
             plugin = plugin(**kwargs)
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4078,7 +4078,7 @@ class Client:
 
         Parameters
         ----------
-        name: str
+        name : str
             Name of the plugin as it was registered.
             See ``register_worker_plugin`` docstrings for more reference.
             If there was not a name assigned when the plugin was added, it will
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4013,6 +4013,7 @@ class Client:
         name : str, optional
             A name for the plugin.
             Registering a plugin with the same name will have no effect.
+            If there is no name assigned it automatically assign ``name = funcname(plugin) + "-" + str(uuid.uuid4())``
         **kwargs : optional
             If you pass a class as the plugin, instead of a class instance, then the
             class will be instantiated with any extra keyword arguments.
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4067,7 +4067,44 @@ class Client:
         return responses
 
     def unregister_worker_plugin(self, name):
-        """Unregisters a lifecycle worker plugin"""
+        """Unregisters a lifecycle worker plugin
+
+        This unregister a worker plugin using its name attribute by calling
+        the plugin's teardown method. (See the ``dask.distributed.WorkerPlugin``
+        class or the examples below for the interface and docstrings.)
+
+        Parameters
+        ----------
+        name: str
+            Name of the plugin as it was registered.
+            See ``register_worker_plugin`` docstrings for more reference.
+            If there was not a name assigned when the plugin was added, it will
+            automatically assign ``name = funcname(plugin) + "-" + str(uuid.uuid4())``
+
+        Examples
+        --------
+        >>> class MyPlugin(WorkerPlugin):
+        ...     def __init__(self, *args, **kwargs):
+        ...         pass  # the constructor is up to you
+        ...     def setup(self, worker: dask.distributed.Worker):
+        ...         pass
+        ...     def teardown(self, worker: dask.distributed.Worker):
+        ...         pass
+        ...     def transition(self, key: str, start: str, finish: str, **kwargs):
+        ...         pass
+        ...     def release_key(self, key: str, state: str, cause: Optional[str], reason: None, report: bool):
+        ...         pass
+        ...     def release_dep(self, dep: str, state: str, report: bool):
+        ...         pass
+
+        >>> plugin = MyPlugin(1, 2, 3)
+        >>> client.register_worker_plugin(plugin, name='foo')
+        >>> client.unregister_worker_plugin(name='foo')
+
+        See Also
+        --------
+        register_worker_plugin
+        """
         return self.sync(self._unregister_worker_plugin, name=name)
 
 
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4073,9 +4073,8 @@ class Client:
     def unregister_worker_plugin(self, name):
         """Unregisters a lifecycle worker plugin
 
-        This unregister a worker plugin using its name attribute by calling
-        the plugin's teardown method. (See the ``dask.distributed.WorkerPlugin``
-        class or the examples below for the interface and docstrings.)
+        This unregisters an existing worker plugin. As part of the unregistration process
+        the plugin's ``teardown`` method will be called.
 
         Parameters
         ----------
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4061,7 +4061,6 @@ class Client:
         for response in responses.values():
             if response["status"] == "error":
                 exc = response["exception"]
-                typ = type(exc)
                 tb = response["traceback"]
                 raise exc.with_traceback(tb)
         return responses
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4013,7 +4013,7 @@ class Client:
         name : str, optional
             A name for the plugin.
             Registering a plugin with the same name will have no effect.
-            If there is no name assigned it automatically assign ``name = funcname(plugin) + "-" + str(uuid.uuid4())``
+            If plugin has no name attribute a random name is used.
         **kwargs : optional
             If you pass a class as the plugin, instead of a class instance, then the
             class will be instantiated with any extra keyword arguments.
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -4079,10 +4079,8 @@ class Client:
         Parameters
         ----------
         name : str
-            Name of the plugin as it was registered.
-            See ``register_worker_plugin`` docstrings for more reference.
-            If there was not a name assigned when the plugin was added, it will
-            automatically assign ``name = funcname(plugin) + "-" + str(uuid.uuid4())``
+            Name of the plugin to unregister. See the :meth:`Client.register_worker_plugin`
+            docstring for more information.
 
         Examples
         --------
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -51,7 +51,7 @@ from .core import (
     connect,
     rpc,
 )
-from .diagnostics.plugin import UploadFile, WorkerPlugin
+from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
@@ -4052,6 +4052,11 @@ class Client:
         """
         if isinstance(plugin, type):
             plugin = plugin(**kwargs)
+
+        if name is None:
+            name = _get_worker_plugin_name(plugin)
+
+        assert name
 
         return self.sync(self._register_worker_plugin, plugin=plugin, name=name)
 
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -190,8 +190,8 @@ class WorkerPlugin:
 
 
 def _get_worker_plugin_name(plugin) -> str:
-    """Returns the worker plugin name. If plugin has no name
-    it assigns a random name using funcname and uuid"""
+    """Returns the worker plugin name. If plugin has no name attribute
+    a random name is used."""
     if hasattr(plugin, "name"):
         return plugin.name
     else:
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -188,13 +188,14 @@ class WorkerPlugin:
             Whether the worker should report the released dependency to the scheduler.
         """
 
-    def _get_worker_plugin_name(plugin) -> str:
-        """Returns the worker plugin name. If plugin has no name
-        it assigns a random name using funcname and uuid"""
-        if hasattr(plugin, "name"):
-            return plugin.name
-        else:
-            return funcname(plugin) + "-" + str(uuid.uuid4())
+
+def _get_worker_plugin_name(plugin) -> str:
+    """Returns the worker plugin name. If plugin has no name
+    it assigns a random name using funcname and uuid"""
+    if hasattr(plugin, "name"):
+        return plugin.name
+    else:
+        return funcname(plugin) + "-" + str(uuid.uuid4())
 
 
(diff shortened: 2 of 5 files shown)


distributed/diagnostics/tests/test_worker_plugin.py
```diff
@@ -53,6 +53,32 @@ async def test_create_with_client(c, s):
...
 
+@gen_cluster(client=True, nthreads=[])
+async def test_remove_with_client(c, s):
+    await c.register_worker_plugin(MyPlugin(123), name="foo")
+
+    worker = await Worker(s.address, loop=s.loop)
+    assert worker._my_plugin_status == "setup"
+    assert worker._my_plugin_data == 123
+
+    await c.unregister_worker_plugin("bar")
```
jrbourbeau, line 64:
```suggestion
    await c.unregister_worker_plugin("foo")
```

distributed/diagnostics/tests/test_worker_plugin.py
```diff
@@ -53,6 +53,32 @@ async def test_create_with_client(c, s):
     assert worker._my_plugin_status == "teardown"
 
 
+@gen_cluster(client=True, nthreads=[])
+async def test_remove_with_client(c, s):
+    await c.register_worker_plugin(MyPlugin(123), name="foo")
+
+    worker = await Worker(s.address, loop=s.loop)
+    assert worker._my_plugin_status == "setup"
+    assert worker._my_plugin_data == 123
```
jrbourbeau, line 62: Thinking about it a bit more, I think we can remove these `assert`s since they are testing that the plugin's `setup` method ran which is not the point of this test and is being tested elsewhere (in `test_create_with_client`)

distributed/diagnostics/tests/test_worker_plugin.py
```diff
@@ -53,6 +53,32 @@ async def test_create_with_client(c, s):
...
+    assert not worker.plugins
+
+
+@gen_cluster(client=True, nthreads=[])
+async def test_remove_with_client_raises(c, s):
+    await c.register_worker_plugin(MyPlugin(123), name="foo")
+
+    worker = await Worker(s.address, loop=s.loop)
+    assert worker._my_plugin_status == "setup"
+    assert worker._my_plugin_data == 123
```
jrbourbeau, line 76: Similar comment here about removing these `assert`s

distributed/scheduler.py
```diff
@@ -6304,6 +6305,17 @@ async def register_worker_plugin(self, comm, plugin, name=None):
         )
         return responses
 
+    async def unregister_worker_plugin(self, comm, name):
+        """ Unregisters a worker plugin"""
+
+        for idx, entry in enumerate(self.worker_plugins):
+            if entry["name"] == name:
+                break
```
jrbourbeau, line 6313: We might want to add an `else` condition here (to make this a [`for`/`else` loop](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops)) so we can raise an informative error if `self.worker_plugins` doesn't contain an entry for a plugin with the name `name`

jrbourbeau: Alternatively, we might also want to think about updating `self.worker_plugins` to be a dictionary (similar to `Worker.plugins`) so we could just simply `pop` the desired plugin off `self.worker_plugins`. Though that's a larger change, which may have additional knock off effects

jrbourbeau (commented):
Thanks for your work on this @ncclementi! Even though this PR is still a WIP, I left a few initial thoughts below

distributed/diagnostics/tests/test_worker_plugin.py
ncclementi: Thanks for the comments James. 
Sounds good, in that case should we re-factor the test to include the case of having two plugins and remove one, or do we want to have it as a separate one?

jrbourbeau: Yeah, registering two worker plugins and then removing them one at a time would be good to include here. That way we can make sure that the logic on the scheduler and worker which selects which plugin should be removed is correct.

distributed/scheduler.py
ncclementi: Added an else to the for/else loop to raise a `keyError` with a more descriptive message.

distributed/diagnostics/tests/test_worker_plugin.py
ncclementi: Removed the asserts and made the test to create two plugins and delete one at the time with the proper asserts in between.

distributed/client.py
```diff
@@ -4055,6 +4055,21 @@ class will be instantiated with any extra keyword arguments.
 
         return self.sync(self._register_worker_plugin, plugin=plugin, name=name)
 
+    async def _unregister_worker_plugin(self, name):
+        responses = await self.scheduler.unregister_worker_plugin(name=name)
+
+        for response in responses.values():
+            if response["status"] == "error":
+                exc = response["exception"]
+                typ = type(exc)
```
jrbourbeau, line 4064: It looks like we're not using the `typ` variable here 

```suggestion
```

distributed/client.py
```diff
@@ -4055,6 +4055,21 @@ class will be instantiated with any extra keyword arguments.
...
+                tb = response["traceback"]
+                raise exc.with_traceback(tb)
+        return responses
+
+    def unregister_worker_plugin(self, name):
+        """Unregisters a lifecycle worker plugin"""
```
jrbourbeau, line 4070: Could you expand this docstring a little bit? You can check out the existing `Client.register_worker_plugin` docstring for an example of a similar docstring. In particular, it'd be good to mention that a plugin's `teardown` method is called when it's unregistered.

distributed/scheduler.py
```diff
@@ -6304,6 +6305,20 @@ async def register_worker_plugin(self, comm, plugin, name=None):
...
+        else:
+            raise KeyError(f"The worker plugin {name} does not exists")
```
jrbourbeau, line 6315: Nitpick: raising a `ValueError` probably makes more sense here as this is more about an incorrect value (for `name`) rather than for a missing value in a mapping (though this is a very minor point)

distributed/diagnostics/tests/test_worker_plugin.py
```diff
@@ -53,6 +53,40 @@ async def test_create_with_client(c, s):
...
+    await c.unregister_worker_plugin("foo")
+    assert worker._my_plugin_status == "teardown"
+
+    # check that on the scheduler regitered worker plugins we only have 'bar'
+    assert len(s.worker_plugins) == 1
+    assert s.worker_plugins[0]["name"] == "bar"
+
+    # check on the worker plugins that we only have 'bar'
+    assert "foo" not in worker.plugins
+    assert "bar" in worker.plugins
```
jrbourbeau, line 72: Instead of checking that `"foo"` isn't in the worker's `dict` of plugins, could we check that `len(worker.plugins) == 1` and that `"bar"` is in `worker.plugins` (similar to what we're doing above with the scheduler)

jrbourbeau (commented):
Thanks for the updates @ncclementi! I left a few final comments, but overall this is looking great

distributed/scheduler.py
ncclementi: Regarding updating the `self.worker_plugins` to be a dictionary I'm having an error when trying to do so. I pushed some comments with the changes I would do to make this work. But when trying to run the test that registers a plugin `test_create_with_client` it throws an error in `worker.py` line 926.

jrbourbeau: Thanks for pushing up those changes, they're definitely heading in the right direction! In addition to making changes to the scheduler, we'll also need to make sure that the workers can handle this message

https://github.com/dask/distributed/blob/8dfaca632403b14bf73350780201394d1dd0e5ba/distributed/scheduler.py#L4004-L4009

that the scheduler sends to a new worker when it's added to the cluster. Specifically here

https://github.com/dask/distributed/blob/8dfaca632403b14bf73350780201394d1dd0e5ba/distributed/worker.py#L922-L927

is where the worker processes the `"worker-plugins"` part of this message. We'll need make an update like the following since `response["worker-plugins"]` is now a `dict` instead of a `list`:

```diff
diff --git a/distributed/worker.py b/distributed/worker.py
index 413425b5..05ba5c41 100644
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -922,8 +922,8 @@ class Worker(ServerNode):
         else:
             await asyncio.gather(
                 *[
-                    self.plugin_add(**plugin_kwargs)
-                    for plugin_kwargs in response["worker-plugins"]
+                    self.plugin_add(name=name, plugin=plugin)
+                    for name, plugin in response["worker-plugins"].items()
                 ]
             )
```

ncclementi: Thanks @jrbourbeau. I modify this and the tweak the test to work properly, now all test are green after changing the data structure of `worker_plugins` to a `dict()`.

distributed/client.py
ncclementi: I expanded this docstring, not sure if it's enough let me know if something is unclear.

jrbourbeau (commented):
It looks like switch `Scheduler.worker_plugins` from a `list` to a `dict` means we have to handle changes where `Client.register_worker_plugin` is called multiple times without a `name=` keyword specified. Today, when `Scheduler.worker_plugins` is a `list`, if we do something like 

```python
client.register_worker_plugin(MyPlugin(123))
client.register_worker_plugin(MyPlugin(123))
```

that will result in `Scheduler.worker_plugins` having two entries, each that looks like `{"plugin": ..., "name": None}`. However, when we switch to using a `dict` for `Scheduler.worker_plugins`, we'll only have a single item in the `worker_plugins` mapping. 

I still think it's a good idea to move `worker_plugins` to a `dict`, but this will require us to make a few more changes to keep `distributed`s current behavior. In particular, let's add a new `_get_worker_plugin_name` utility function which handles generating a `name` for a given worker plugin:

```diff
diff --git a/distributed/diagnostics/plugin.py b/distributed/diagnostics/plugin.py
index 58b5adbc..a2068c45 100644
--- a/distributed/diagnostics/plugin.py
+++ b/distributed/diagnostics/plugin.py
@@ -3,6 +3,9 @@ import os
 import socket
 import subprocess
 import sys
+import uuid
+
+from dask.utils import funcname

 logger = logging.getLogger(__name__)

@@ -186,6 +189,13 @@ class WorkerPlugin:
         """


+def _get_worker_plugin_name(plugin) -> str:
+    if hasattr(plugin, "name"):
+        return plugin.name
+    else:
+        return funcname(plugin) + "-" + str(uuid.uuid4())
+
+
 class PipInstall(WorkerPlugin):
     """A Worker Plugin to pip install a set of packages
```

And then use `_get_worker_plugin_name` in `client.py` and `worker.py` like the following.

`client.py`:
```diff
diff --git a/distributed/client.py b/distributed/client.py
index 695c9dd3..2e98a506 100644
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -51,7 +51,7 @@ from .core import (
     connect,
     rpc,
 )
-from .diagnostics.plugin import UploadFile, WorkerPlugin
+from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
@@ -4053,6 +4053,11 @@ class Client:
         if isinstance(plugin, type):
             plugin = plugin(**kwargs)

+        if name is None:
+            name = _get_worker_plugin_name(plugin)
+
+        assert name
+
         return self.sync(self._register_worker_plugin, plugin=plugin, name=name)
```

`worker.py`:
```diff
diff --git a/distributed/worker.py b/distributed/worker.py
index eb87d797..2ad5f5bb 100644
--- a/distributed/worker.py
+++ b/distributed/worker.py
@@ -7,7 +7,6 @@ import os
 import random
 import sys
 import threading
-import uuid
 import warnings
 import weakref
 from collections import defaultdict, deque, namedtuple
@@ -41,6 +40,7 @@ from .core import (
     pingpong,
     send_recv,
 )
+from .diagnostics.plugin import _get_worker_plugin_name
 from .diskutils import WorkSpace
 from .http import get_handlers
 from .metrics import time
@@ -2551,11 +2551,8 @@ class Worker(ServerNode):
         with log_errors(pdb=False):
             if isinstance(plugin, bytes):
                 plugin = pickle.loads(plugin)
-            if not name:
-                if hasattr(plugin, "name"):
-                    name = plugin.name
-                else:
-                    name = funcname(plugin) + "-" + str(uuid.uuid4())
+            if name is None:
+                name = _get_worker_plugin_name(plugin)

             assert name
```

ncclementi:
Thanks @jrbourbeau. That makes a lot of sense, added the changes you suggested. Run all the tests in `test_worker_plugin.py` they all passed.

distributed/client.py
```diff
@@ -4053,8 +4053,64 @@ class will be instantiated with any extra keyword arguments.
...
+                tb = response["traceback"]
+                raise exc.with_traceback(tb)
+        return responses
+
+    def unregister_worker_plugin(self, name):
+        """Unregisters a lifecycle worker plugin
+
+        This unregister a worker plugin using its name attribute by calling
+        the plugin's teardown method. (See the ``dask.distributed.WorkerPlugin``
+        class or the examples below for the interface and docstrings.)
```
jrbourbeau, line 4078:
```suggestion
        This unregisters an existing worker plugin. As part of the unregistration process
        the plugin's ``teardown`` method will be called.
```

distributed/client.py
```diff
@@ -4053,8 +4053,64 @@ class will be instantiated with any extra keyword arguments.
...
+
+        Parameters
+        ----------
+        name: str
```
jrbourbeau, line 4082: Nitpick: `distributed` (and `dask`) use [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) for formatting docstrings. numpydoc's style guide has a space between the parameter name and colon here. 

```suggestion
        name : str
```

distributed/client.py
```diff
@@ -4053,8 +4053,64 @@ class will be instantiated with any extra keyword arguments.
...
+            Name of the plugin as it was registered.
+            See ``register_worker_plugin`` docstrings for more reference.
+            If there was not a name assigned when the plugin was added, it will
+            automatically assign ``name = funcname(plugin) + "-" + str(uuid.uuid4())``
```
jrbourbeau, line 4086: We can link directly to the `register_worker_plugin` API docs by using `:meth:`

```suggestion
            Name of the plugin to unregister. See the :meth:`Client.register_worker_plugin`
            docstring for more information.
```

Additionally, I think the part about the random name generation should be moved to the `register_worker_plugin` docstring since that's where that logic is actually executed. Maybe something like "If not provided, a random name will be generated".

distributed/client.py
```diff
@@ -4053,8 +4053,64 @@ class will be instantiated with any extra keyword arguments.
...
+        ...     def release_dep(self, dep: str, state: str, report: bool):
+        ...         pass
+
+        >>> plugin = MyPlugin(1, 2, 3)
+        >>> client.register_worker_plugin(plugin, name='foo')
+        >>> client.unregister_worker_plugin(name='foo')
+
+        See Also
+        --------
+        register_worker_plugin
```
jrbourbeau, line 4110: Could you add `unregister_worker_plugin` to the "See Also" section for `register_worker_plugin`?

distributed/diagnostics/plugin.py
```diff
@@ -186,6 +189,15 @@ def release_dep(self, dep, state, report):
         """
 
 
+def _get_worker_plugin_name(plugin) -> str:
+    """Returns the worker plugin name. If plugin has no name
+    it assigns a random name using funcname and uuid"""
```
jrbourbeau, line 194: Nitpick: Explicitly mentioning `funcname` and `uuid` in this docstring might be a little too specific as these are really implementation details. Maybe just mention that a random name is generated? 

```suggestion
    """ Returns the worker plugin name. If plugin has no name attribute a random name is used. """
```

jrbourbeau (commented):
Some final comments about the docs added here (which are generally great)

distributed/client.py
ncclementi: Makes sense, I added the random name generation part to the register_worker_plugin docstring

jrbourbeau (approved):
Thanks for all your work on this @ncclementi! This is in

PR #4753, opened 2021-04-26
Ensure deps are actually logged in worker

fjetter:
Ensure dependencies in our logs are actually logged. The set is popped a few lines below and we need to create a copy for the logs

### DIFF
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -54,7 +54,14 @@ from distributed.utils_test import (  # noqa: F401
     s,
     slowinc,
 )
-from distributed.worker import Worker, error_message, logger, parse_memory_limit, weight
+from distributed.worker import (
+    TaskState,
+    Worker,
+    error_message,
+    logger,
+    parse_memory_limit,
+    weight,
+)
 
 
 @pytest.mark.asyncio
@@ -1787,6 +1794,36 @@ async def test_story(c, s, w):
     ts = w.tasks[future.key]
     assert ts.state in str(w.story(ts))
     assert w.story(ts) == w.story(ts.key)
+
+
+@gen_cluster(client=True)
+async def test_story_with_deps(c, s, a, b):
+    """
+    Assert that the structure of the story does not change unintentionally and
+    expected subfields are actually filled
+    """
+    future = c.map(inc, range(10))
+    res = c.submit(sum, future)
+    await res
+
+    for w in [a, b]:
+        try:
+            ts = w.tasks[res.key]
+        except KeyError:
+            continue
+        story = w.story(ts)
+        for msg in story:
+            # There are multiple types of messages
+            # Transition
+            if isinstance(msg[-1], str):
+                assert msg[0] == ts.key
+            # events, e.g. gather-deps
+            elif msg[0] == "gather-dependencies":
+                assert msg[1] == ts.key
+                dependencies = msg[2]
+                assert isinstance(dependencies, set)
+                assert dependencies
+                assert all(isinstance(dep, TaskState) for dep in dependencies)
 
 
 def test_weight_deprecated():
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -54,14 +54,7 @@ from distributed.utils_test import (  # noqa: F401
     s,
     slowinc,
 )
-from distributed.worker import (
-    TaskState,
-    Worker,
-    error_message,
-    logger,
-    parse_memory_limit,
-    weight,
-)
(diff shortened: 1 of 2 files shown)


distributed/tests/test_worker.py
```diff
@@ -1789,6 +1796,36 @@ async def test_story(c, s, w):
...
+        try:
+            ts = w.tasks[res.key]
+        except KeyError:
+            continue
+        story = w.story(ts)
+        for msg in story:
+            # There are multiple types of messages
+            # Transition
+            if isinstance(msg[-1], str):
+                assert msg[0] == ts.key
```
jrbourbeau, line 1819: We don't seem to include the 

```python
            if isinstance(msg[-1], str):
                assert msg[0] == ts.key
```

logic in how we filter `Worker.log` to determine what log entries below to a key's story

https://github.com/dask/distributed/blob/74de05c783e99e547513bdc4d11169daf783d72e/distributed/worker.py#L1946-L1958

I agree this `assert` will pass for the example in this computation today, but it might not be in the future. Could we generalize the check we're doing here to be more future-proof?

distributed/tests/test_worker.py
```diff
@@ -1789,6 +1796,36 @@ async def test_story(c, s, w):
...
+            # events, e.g. gather-deps
+            elif msg[0] == "gather-dependencies":
```
jrbourbeau, line 1821: We might also want to `assert` earlier on that that at least one of the messages is a `gather-dependencies` message so we know for sure we go into this branch

jrbourbeau (commented):
Nice catch @fjetter!

Briefly looking through the other `Worker.log.append` occurrences in `worker.py` I think we should also add a `copy` here where we could run into a similar situation as the `gather-dependencies` message: 

https://github.com/dask/distributed/blob/74de05c783e99e547513bdc4d11169daf783d72e/distributed/worker.py#L2315

fjetter:
The `self.log.append(("handle-missing", deps))`  _could_  indeed also be subject to this problem but since the `deps` in this case is a tuple (as part of the function signature) and therefore immutable this is not an issue in this case. 

https://github.com/dask/distributed/blob/233ec8840671b5d705a06d0dcfe1ff42b771230b/distributed/worker.py#L2315-L2316

Fixing this indefinitely would probably require some standardization, (xref https://github.com/dask/distributed/issues/4762). That does not necessarily need to be as complicated as actual structured logging but would at least require something more complex than "append anything to the list"

distributed/tests/test_worker.py
fjetter: Honestly, I think this should even be much stricter and I tend to put the entire story, as is, as an expected story in this 
test. The tokens here are deterministic as well as the order of transitions. If anything in this representation changes something quite fundamental changed and I believe it should be justified to adjust this test if either the state machine and/or the logging changes fundamentally.

Writing this up, I also realised that I was, accidentally, not on purpose, logging the actual `TaskState` objects which I think is wrong since they are mutable and require more memory than the simple key.


```python
        story = w.story(ts)
        expected_story = [
            ("sum-6e43962b5fde8e7d28875ebbdb8e7bc7", "new"),
            ("sum-6e43962b5fde8e7d28875ebbdb8e7bc7", "new", "waiting"),
            (
                "gather-dependencies",
                "sum-6e43962b5fde8e7d28875ebbdb8e7bc7",
                {
                    "inc-aa226ae3a9f799819e1e685fba467442",
                    "inc-cfa5385a7159c89ba41e2744bb25c644",
                    "inc-9212ebc1516c49fcfd26f9434f14c524",
                    "inc-0c3f7739638b70e403ab21f80b09970b",
                    "inc-aa0231dc9b3c3d81591e1fe036cfdf54",
                },
            ),
            ("sum-6e43962b5fde8e7d28875ebbdb8e7bc7", "waiting", "ready"),
            ("sum-6e43962b5fde8e7d28875ebbdb8e7bc7", "ready", "executing"),
            ("sum-6e43962b5fde8e7d28875ebbdb8e7bc7", "put-in-memory"),
            ("sum-6e43962b5fde8e7d28875ebbdb8e7bc7", "executing", "memory"),
        ]
        assert story == expected_story
```

fjetter: Well, at least the tokens _should_ be deterministic. There seems to be a bug or smth in the map function :/

fjetter: Wasn't a bug, just that the map doesn't dispatch deterministically. I went along and made it more explicit without hard coding the tokens

fjetter:
I assume this is ok to go in. If something pops up, I'll take care of it afterwards

Issue #4730, opened 2021-04-21
Handle `ts=None` in `_task_to_report_msg`


jrbourbeau:
We've started seeing some tests fail when `ts=None` is passed to `_task_to_report_msg` (e.g. see [this CI build](https://github.com/dask/distributed/pull/4720/checks?check_run_id=2403923283)). Currently if we pass `ts=None` to `_task_to_report_msg` we will attempt to retreive its `._key` attribute:

https://github.com/dask/distributed/blob/f492aa7e7a39270aaf0b58e37ed855c149b68a8f/distributed/scheduler.py#L7063-L7067

which doesn't exist for `None` (see the traceback below). 

It looks like the `_task_to_report_msg` function was added in https://github.com/dask/distributed/pull/4365 (cc @jakirkham) although the corresponding logic may predate that PR. Also cc'ing @fjetter @gforsyth 

<details>
<summary>Traceback:</summary>

```python
tornado.application - ERROR - Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x7f505ad304d0>, <Task finished coro=<BaseTCPListener._handle_stream() done, defined at /home/runner/work/distributed/distributed/distributed/comm/tcp.py:476> exception=AttributeError("'NoneType' object has no attribute '_key'")>)
Traceback (most recent call last):
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/site-packages/tornado/ioloop.py", line 758, in _run_callback
    ret = callback()
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/site-packages/tornado/stack_context.py", line 300, in null_wrapper
    return fn(*args, **kwargs)
  File "/usr/share/miniconda3/envs/dask-distributed/lib/python3.7/site-packages/tornado/tcpserver.py", line 297, in <lambda>
    lambda f: f.result())
  File "/home/runner/work/distributed/distributed/distributed/comm/tcp.py", line 493, in _handle_stream
    await self.comm_handler(comm)
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 501, in handle_comm
    result = await result
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4919, in add_client
    await self.handle_stream(comm=comm, extra={"client": client})
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 573, in handle_stream
    handler(**merge(extra, msg))
[...]

PR #4755, opened 2021-04-26
Handle missing `key` case in `report_on_key`

jakirkham:
Handle missing `key` case in `report_on_key`. This [was previously handled]( https://github.com/dask/distributed/blob/dfbe171905b349fdd933424ae7af2b6217a28133/distributed/scheduler.py#L3672-L3676 ), but was accidentally dropped during refactoring. Here we add it back.

Some dead code was found in `_task_to_client_msgs` because of this change and was dropped as result. Also some unneeded code was dropped from `_task_to_report_msg`.

- [x] Closes https://github.com/dask/distributed/issues/4730
- [x] Tests added / passed

### DIFF
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -413,6 +413,7 @@ async def test_delete_data(c, s, a, b):
 async def test_delete(c, s, a):
     x = c.submit(inc, 1)
     await x
+    assert x.key in s.tasks
     assert x.key in a.data
 
     await c._cancel(x)
@@ -421,6 +422,10 @@ async def test_delete(c, s, a):
     while x.key in a.data:
         await asyncio.sleep(0.01)
         assert time() < start + 5
+
+    assert x.key not in s.tasks
+
+    s.report_on_key(key=x.key)
 
 
 @gen_cluster()
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -7082,15 +7082,9 @@ def _task_to_report_msg(state: SchedulerState, ts: TaskState) -> dict:
 @cfunc
 @exceptval(check=False)
 def _task_to_client_msgs(state: SchedulerState, ts: TaskState) -> dict:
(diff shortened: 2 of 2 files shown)


jakirkham:
cc @jrbourbeau

jrbourbeau (commented):
Thanks @jakirkham, overall this looks good! Could you add a small test to make sure `Scheduler.report_on_key` can handle the `ts is None` case when we start to construct  `report_msg`?

jakirkham:
Sure is there an existing test we can pull from? Searching for `report_on_key` and `"report-key"` doesn't show any existing tests. Is there another similar test that might work?

jakirkham:
Alright found that `test_delete` was doing enough of the right things to work for this purpose. Have added a test.

Without the changes in this PR, I see the following error

```python

    @cfunc
    @exceptval(check=False)
    def _task_to_report_msg(state: SchedulerState, ts: TaskState) -> dict:
        if ts is None:
>           return {"op": "cancelled-key", "key": ts._key}
E           AttributeError: 'NoneType' object has no attribute '_key'

distributed/scheduler.py:7062: AttributeError
```

With the changes in this PR, it passes! 🎉

distributed/tests/test_scheduler.py
```diff
@@ -422,6 +423,10 @@ async def test_delete(c, s, a):
         await asyncio.sleep(0.01)
         assert time() < start + 5
 
+    assert x.key not in s.tasks
+
+    s.report_on_key(key=x.key)
```
jakirkham, line 428: Note that we have to confirm that `x.key` is no longer in `s.tasks`. Otherwise it would just grab the `TaskState` object from there even with the `key` specified. IOW now we are sure this will test the case where no `TaskState` can be found and the `key` must be used

jrbourbeau (approved):
Great sleuthing to find `test_delete` @jakirkham! Thanks for your work on this, this is in

PR #4758, opened 2021-04-27
Drop temporary `set` variables `s`

jakirkham:
These were needed before `TaskState` was annotated. Now that `TaskState` is annotated, Cython already knows these are `set`s and so handles them properly.

- [ ] Closes #xxxx
- [ ] Tests added / passed


jrbourbeau (approved):
Thanks @jakirkham!

PR #4759, opened 2021-04-27
Add info tooltips to Dashboard plots

### LINKED ISSUE #4746: Add an info tooltip to dashboard plots
We expose many diagnostic dashboard plots to users. This is great as it provides a detailed, real-time look into a cluster's state.

However, it may not be immediately clear to users (especially new users) what information is actually being displayed in each plot. For example, users may not know that the central circle in the "cluster map" plot (a.k.a the "pew-pew plot") represents  the scheduler while the surrounding circles represent workers. 

One way we could better inform users would be to add an info icon ℹ️ to our dashboard plots which, when users hover their mouse over it, displays information about what the plot is displaying. 

cc @charlesbluca who has been thinking about dashboard visualizations recently

charlesbluca:
Adds a Bokeh `HelpTool` to most plots with a visible toolbar, allowing users to hover over it for a brief explanation of the plot. Several of these still need to be filled in (marked with `description="TODO"`), but I figured that might be something easier to hash out with members of the community.

- [ ] Closes #4746
- [ ] Tests added / passed

### DIFF
--- a/distributed/dashboard/components/nvml.py
+++ b/distributed/dashboard/components/nvml.py
@@ -3,6 +3,7 @@ import math
 from bokeh.models import (
     BasicTicker,
     ColumnDataSource,
+    HelpTool,
     HoverTool,
     NumeralTickFormatter,
     OpenURL,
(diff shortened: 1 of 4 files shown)


charlesbluca:
Some minor concerns here:

- The `HelpTool` redirects to a page on click, which AFAIK cannot be disabled. This could be useful for some plots if we wanted to link to some informative relevant documentation, but may be redundant for others. Are we okay with linking to a general Dask/Distributed page in some cases, or would this be a good case for a custom Bokeh tool that doesn't have any redirect?
- I refrained from adding a `HelpTool` to plots that didn't have a toolbar due to spacing issues; for example, most of the plots on the status page:

image

charlesbluca:
Closing this as bokeh/bokeh#11389 could offer a better solution than the tooltips I'm using here.

PR #4767, opened 2021-04-29
[WIP] Fix Cythonized Scheduler test failures

jakirkham:
Reverts the tests skipped/xfailed in PR ( https://github.com/dask/distributed/pull/4764 ) due to issues encountered with the Cythonized Scheduler. Doing this to fix those tests and get them passing in the Cythonized Scheduler case as well.

**Note:** This PR mainly exists for tracking purposes while we sort out the different tests failures.

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/cli/tests/test_dask_spec.py
+++ b/distributed/cli/tests/test_dask_spec.py
@@ -4,12 +4,10 @@ import pytest
 import yaml
 
 from distributed import Client
-from distributed.scheduler import COMPILED
 from distributed.utils_test import cleanup  # noqa: F401
 from distributed.utils_test import popen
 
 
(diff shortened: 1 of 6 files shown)


jakirkham:
Was able to fix a couple of these without too much difficulty. Broke out in PR ( https://github.com/dask/distributed/pull/4768 ).

jakirkham:
One avenue to addressing these remaining issues would be to continue with splitting apart `Scheduler` & `SchedulerState` with Martin's PR ( https://github.com/dask/distributed/pull/5176 ), which opens the door to moving `SchedulerState` into its own module. If `SchedulerState` could live in its own module that gets Cythonized without the `Scheduler`, there would be less need to worry about how Cython handles some of these other test failures that cropped up with the `Scheduler`.

PR #4768, opened 2021-04-29
Fixes some pickling issues in the Cythonized Scheduler

jakirkham:
A few of the test failures discovered in PR ( https://github.com/dask/distributed/pull/4764 ) were related to pickling issues. One relating to pickling unbound methods, which Cython has a fix for in a pending release ( https://github.com/cython/cython/issues/2972 ). Another related pickling to `lambda`s ( https://github.com/cython/cython/issues/3499 ).

Here we workaround these issues by using other techniques. This fixes a couple of test failures on the Cython Scheduler. So are now re-enabled in this PR.

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/diagnostics/progress_stream.py
+++ b/distributed/diagnostics/progress_stream.py
@@ -22,6 +22,8 @@ def counts(scheduler, allprogress):
 
 
 def remove_plugin(*args, **kwargs):
+    # Wrapper function around `Scheduler.remove_plugin` to avoid raising a
+    # `PicklingError` when using a cythonized scheduler
     return Scheduler.remove_plugin(*args, **kwargs)
 
 
--- a/distributed/diagnostics/progress_stream.py
+++ b/distributed/diagnostics/progress_stream.py
@@ -21,6 +21,10 @@ def counts(scheduler, allprogress):
     )
 
 
+def remove_plugin(*args, **kwargs):
+    return Scheduler.remove_plugin(*args, **kwargs)
+
+
 async def progress_stream(address, interval):
     """Open a TCP connection to scheduler, receive progress messages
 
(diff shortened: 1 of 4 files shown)


distributed/diagnostics/progress_stream.py
```diff
@@ -47,7 +51,7 @@ async def progress_stream(address, interval):
             "setup": dumps_function(AllProgress),
             "function": dumps_function(counts),
             "interval": interval,
-            "teardown": dumps_function(Scheduler.remove_plugin),
+            "teardown": dumps_function(remove_plugin),
```
fjetter, line 54: Maybe dumb question but `Scheduler.remove_plugin` isn't a local scope function. Why does this change anything?

fjetter (approved):
Just had a questions to help me understand the problem. Changes look good

distributed/diagnostics/progress_stream.py
jrbourbeau: I believe this is a workaround for https://github.com/cython/cython/issues/2972. An upstream fix in `cython` (https://github.com/cython/cython/pull/2969) should be included in the `cython=0.29.24` release

jrbourbeau (approved):
Thanks for these fixes @jakirkham! I pushed a tiny commit to add a comment about `remove_plugin`. Will merge after CI finishes

distributed/diagnostics/progress_stream.py
jakirkham: Yeah exactly

Pickling of the local `remove_plugin` function should only include the function and module to `import`. The code itself won't need to be pickled, which avoids needing to pickle `Scheduler.remove_plugin`

jakirkham:
Thanks all! 😄

PR #4784, opened 2021-05-04
Forget erred tasks // Fix deadlocks on worker

fjetter:
I need some feedback because the cluster behaves differently than I would expect and I would like to know for sure what we expect to happen.

The test case constructs a very simple graph consisting of three tasks, F, G and Res where F and G are simple dependencies executed on different workers while Res is task which will fail. I pinned the workers down explicitly to ensure determinism but otherwise this is not relevant.

What happens is that the task errs which is expected. We keep all tasks in memory, still expected. Afterwards, two things happen which surprised me and I would like to know if those are bugs or expected behaviour I misunderstood

1. I release refs to `F` and `G` and would expect them to be _released_, i.e. the memory is freed because `Res` is done and we no longer hold a reference to the dependencies `F` and `G`. Instead, both stay in memory. Is this expected?
2. After I move on, I release refs to `Res` and would expect the cluster to reset, i.e. forget about all tasks. This isn't happening but instead `F` and `G` are released and `Res` stays erred. The tasks are never forgotten.

cc @jrbourbeau 

----

Edit:

While investigating the free error tasks, I found two deadlock situations which were more likely to be triggered in this branch than in others. This increased the scope of this PR

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1205,6 +1205,10 @@ class TaskState:
        failed task is stored here (possibly itself).  Otherwise this
        is ``None``.
 
+    .. attribute:: erred_on: set(str)
+
+        Worker addresses on which errors appeared causing this task to be in an error state.
+
     .. attribute:: suspicious: int
 
        The number of times this task has been involved in a worker death.
@@ -1293,6 +1297,7 @@ class TaskState:
     _exception: object
     _traceback: object
     _exception_blame: object
+    _erred_on: set
     _suspicious: Py_ssize_t
     _host_restrictions: set
     _worker_restrictions: set
@@ -1343,6 +1348,7 @@ class TaskState:
         "_who_wants",
         "_exception",
         "_traceback",
+        "_erred_on",
         "_exception_blame",
         "_suspicious",
         "_retries",
@@ -1381,6 +1387,7 @@ class TaskState:
         self._group = None
         self._metadata = {}
         self._annotations = {}
+        self._erred_on = set()
 
     def __hash__(self):
         return self._hash
@@ -1527,6 +1534,10 @@ class TaskState:
     @property
     def prefix_key(self):
         return self._prefix._name
+
+    @property
+    def erred_on(self):
+        return self._erred_on
 
     @ccall
     def add_dependency(self, other: "TaskState"):
@@ -1842,7 +1853,6 @@ class SchedulerState:
             ("no-worker", "waiting"): self.transition_no_worker_waiting,
             ("released", "forgotten"): self.transition_released_forgotten,
             ("memory", "forgotten"): self.transition_memory_forgotten,
-            ("erred", "forgotten"): self.transition_released_forgotten,
             ("erred", "released"): self.transition_erred_released,
             ("memory", "released"): self.transition_memory_released,
             ("released", "erred"): self.transition_released_erred,
@@ -2722,7 +2732,6 @@ class SchedulerState:
 
             if self._validate:
                 with log_errors(pdb=LOG_PDB):
-                    assert all([dts._state != "erred" for dts in ts._dependencies])
                     assert ts._exception_blame
                     assert not ts._who_has
                     assert not ts._waiting_on
@@ -2735,6 +2744,11 @@ class SchedulerState:
             for dts in ts._dependents:
                 if dts._state == "erred":
                     recommendations[dts._key] = "waiting"
+
+            w_msg = {"op": "release-task", "key": key}
+            for w in ts._erred_on:
+                worker_msgs[w] = [w_msg]
+            ts._erred_on.clear()
 
             report_msg = {"op": "task-retried", "key": key}
             cs: ClientState
@@ -2835,7 +2849,7 @@ class SchedulerState:
             raise
 
     def transition_processing_erred(
-        self, key, cause=None, exception=None, traceback=None, **kwargs
+        self, key, cause=None, exception=None, traceback=None, worker=None, **kwargs
     ):
         ws: WorkerState
         try:
@@ -2856,8 +2870,9 @@ class SchedulerState:
                 ws = ts._processing_on
                 ws._actors.remove(ts)
 
-            _remove_from_processing(self, ts)
+            w = _remove_from_processing(self, ts)
 
+            ts._erred_on.add(w or worker)
             if exception is not None:
                 ts._exception = exception
             if traceback is not None:
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2639,9 +2639,9 @@ class SchedulerState:
             # XXX factor this out?
             ts_nbytes: Py_ssize_t = ts.get_nbytes()
             worker_msg = {
-                "op": "delete-data",
+                "op": "free-keys",
                 "keys": [key],
-                "report": False,
+                "reason": f"Memory->Released {key}",
             }
             for ws in ts._who_has:
                 del ws._has_what[ts]
@@ -2745,7 +2745,7 @@ class SchedulerState:
                 if dts._state == "erred":
                     recommendations[dts._key] = "waiting"
 
-            w_msg = {"op": "release-task", "key": key}
+            w_msg = {"op": "free-keys", "keys": [key], "reason": "Erred->Released"}
             for w in ts._erred_on:
                 worker_msgs[w] = [w_msg]
             ts._erred_on.clear()
@@ -2819,7 +2819,9 @@ class SchedulerState:
 
             w: str = _remove_from_processing(self, ts)
             if w:
-                worker_msgs[w] = [{"op": "release-task", "key": key}]
+                worker_msgs[w] = [
+                    {"op": "free-keys", "keys": [key], "reason": "Processing->Released"}
+                ]
 
             ts.state = "released"
 
@@ -4471,7 +4473,9 @@ class Scheduler(SchedulerState, ServerNode):
                 ts._who_has,
             )
             if ws not in ts._who_has:
-                worker_msgs[worker] = [{"op": "release-task", "key": key}]
+                worker_msgs[worker] = [
+                    {"op": "free-keys", "keys": [key], "reason": "Stimulus Finished"}
+                ]
 
         return recommendations, client_msgs, worker_msgs
 
@@ -5128,7 +5132,7 @@ class Scheduler(SchedulerState, ServerNode):
     def release_worker_data(self, comm=None, keys=None, worker=None):
         parent: SchedulerState = cast(SchedulerState, self)
         ws: WorkerState = parent._workers_dv[worker]
-        tasks: set = {parent._tasks[k] for k in keys}
+        tasks: set = {parent._tasks[k] for k in keys if k in parent._tasks}
         removed_tasks: set = tasks.intersection(ws._has_what)
 
         ts: TaskState
@@ -5534,8 +5538,11 @@ class Scheduler(SchedulerState, ServerNode):
             List of keys to delete on the specified worker
         """
         parent: SchedulerState = cast(SchedulerState, self)
+
         await retry_operation(
-            self.rpc(addr=worker_address).delete_data, keys=list(keys), report=False
+            self.rpc(addr=worker_address).free_keys,
+            keys=list(keys),
+            reason="rebalance/replicate",
         )
 
         ws: WorkerState = parent._workers_dv[worker_address]
@@ -6286,6 +6293,7 @@ class Scheduler(SchedulerState, ServerNode):
         if worker not in parent._workers_dv:
             return "not found"
         ws: WorkerState = parent._workers_dv[worker]
+        superfluous_data = []
         for key in keys:
             ts: TaskState = parent._tasks.get(key)
             if ts is not None and ts._state == "memory":
@@ -6294,9 +6302,16 @@ class Scheduler(SchedulerState, ServerNode):
                     ws._has_what[ts] = None
                     ts._who_has.add(ws)
             else:
-                self.worker_send(
-                    worker, {"op": "delete-data", "keys": [key], "report": False}
-                )
+                superfluous_data.append(key)
+        if superfluous_data:
+            self.worker_send(
+                worker,
+                {
+                    "op": "superfluous-data",
+                    "keys": superfluous_data,
+                    "reason": f"Add keys which are not in-memory {superfluous_data}",
+                },
+            )
 
         return "OK"
 
@@ -7323,7 +7338,13 @@ def _propagate_forgotten(
         ws._nbytes -= ts_nbytes
         w: str = ws._address
         if w in state._workers_dv:  # in case worker has died
-            worker_msgs[w] = [{"op": "delete-data", "keys": [key], "report": False}]
+            worker_msgs[w] = [
+                {
+                    "op": "free-keys",
+                    "keys": [key],
+                    "reason": f"propagate-forgotten {ts.key}",
+                }
+            ]
     ts._who_has.clear()
 
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2746,8 +2746,8 @@ class SchedulerState:
                     recommendations[dts._key] = "waiting"
 
             w_msg = {"op": "free-keys", "keys": [key], "reason": "Erred->Released"}
-            for w in ts._erred_on:
-                worker_msgs[w] = [w_msg]
+            for ws_addr in ts._erred_on:
+                worker_msgs[ws_addr] = [w_msg]
             ts._erred_on.clear()
 
             report_msg = {"op": "task-retried", "key": key}
--- a/distributed/tests/test_failed_workers.py
+++ b/distributed/tests/test_failed_workers.py
@@ -522,3 +522,101 @@ async def test_worker_time_to_live(c, s, a, b):
         assert time() < start + interval + 0.1
 
     set(s.workers) == {b.address}
+
+
+class SlowDeserialize:
+    def __init__(self, data, delay=0.1):
+        self.delay = delay
+        self.data = data
+
+    def __getstate__(self):
+        return self.delay
+
+    def __setstate__(self, state):
+        delay = state
+        import time
+
+        time.sleep(delay)
+        return SlowDeserialize(delay)
+
+    def __sizeof__(self) -> int:
+        # Ensure this is offloaded to avoid blocking loop
+        import dask
+        from dask.utils import parse_bytes
+
+        return parse_bytes(dask.config.get("distributed.comm.offload")) + 1
+
+
+@gen_cluster(client=True, timeout=None)
+async def test_handle_superfluous_data(c, s, a, b):
+    """
+    See https://github.com/dask/distributed/pull/4784#discussion_r649210094
+    """
+
+    def slow_deser(x, delay):
+        return SlowDeserialize(x, delay=delay)
+
+    futA = c.submit(
+        slow_deser, 1, delay=1, workers=[a.address], key="A", allow_other_workers=True
+    )
+    futB = c.submit(inc, 1, workers=[b.address], key="B")
+    await wait([futA, futB])
+
+    def reducer(*args):
+        return
+
+    assert len(a.tasks) == 1
+    assert futA.key in a.tasks
+
+    assert len(b.tasks) == 1
+    assert futB.key in b.tasks
+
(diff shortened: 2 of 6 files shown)


distributed/tests/test_worker.py
```diff
@@ -1792,3 +1792,75 @@ async def test_story(c, s, w):
...
+        await asyncio.sleep(0.01)
+
+    assert_task_states_on_worker(expected_states, a)
+
+    res.release()
+
+    expected_states = {
+        f.key: "memory",  # expected to be gone
+        g.key: "memory",  # expected to be gone
+        res.key: "error",
```
fjetter, line 1856: This state is actually never left so I do consider this a bug. Eventually the tasks should be released (I commented out the wait below)

distributed/tests/test_worker.py
```diff
@@ -1792,3 +1792,75 @@ async def test_story(c, s, w):
...
+    # A raised the exception therefore we should hold on to the erroneous task
+    assert res.key in a.tasks
+    ts = a.tasks[res.key]
+    assert ts.state == "error"
+
+    expected_states = {
+        # A was instructed to compute this result and we're still holding a ref via `f`
+        f.key: "memory",
+        # This was fetched from another worker. While we hold a ref via `g`, the scheduler
+        g.key: "memory",
```
fjetter, line 1831: I would expect `g` to be released on worker A since it was only on worker A as a dependency. However, it is kept after the task erred. Do we consider this sane behaviour? Do we intentionally keep this dependency in memory to allow for retries?

distributed/tests/test_worker.py
```diff
@@ -1792,3 +1792,75 @@ async def test_story(c, s, w):
...
+        res.key: "error",
+    }
+    assert_task_states_on_worker(expected_states, a)
+    # Expected states after we release references to the futures
+    expected_states = {
+        f.key: "memory",
+        g.key: "memory",
```
fjetter, line 1838: Even after explicit refs to f and g are released, we still keep hold on to them since there is the dependency `res` which erred. I'm wondering what the expected behaviour should be here

distributed/tests/test_worker.py
fjetter: For reference, the scheduler decides in this case that the keys should be kept, or rather it doesn't decide to release them. I am questioning whether this is the correct decision

distributed/tests/test_worker.py
fjetter: For reference, the scheduler forgot all tasks at this point

distributed/worker.py
```diff
@@ -2437,10 +2437,7 @@ def release_key(self, key, cause=None, reason=None, report=True):
                 # don't boot keys that are in flight
                 # we don't know if they're already queued up for transit
                 # in a gather_dep callback
-                if not dependency.dependents and dependency.state in (
-                    "waiting",
-                    "fetch",
-                ):
```
fjetter, line 2443: anybody knows why we were skipping this for specific states? cc @gforsyth

gforsyth: We were skipping it for `flight` since there's a potential race condition if we remove the task and then the `gather_dep` callback fires

distributed/scheduler.py
```diff
@@ -7025,7 +7040,7 @@ def _client_releases_keys(
     state: SchedulerState, keys: list, cs: ClientState, recommendations: dict
 ):
     """ Remove keys from client desired list """
-    logger.debug("Client %s releases keys: %s", cs._client_key, keys)
+    logger.critical("Client %s releases keys: %s", cs._client_key, keys)
```
fjetter, line 7043: - [ ] rollback. this was for debugging

distributed/scheduler.py
```diff
@@ -1522,6 +1529,10 @@ def group_key(self):
     def prefix_key(self):
         return self._prefix._name
 
+    @property
+    def erred_on(self):
+        return self._erred_on
```
fjetter, line 1534: One of the reasons why the erred task state is never forgotten is that keys are only "remembered" in `processing_on (processing)` or `who_has (memory)`, depending on the state of the task. This would be the equivalent for erred tasks to allow us to tell the worker to forget the task.

@jakirkham you have been involved in the scheduler state machine a lot recently. Just pinging in case you have thoughts about adding more state here or if you see other options

distributed/tests/test_scheduler.py
```diff
@@ -1872,18 +1872,33 @@ async def test_task_groups(c, s, a, b):
...
+    assert tg.nbytes_in_memory == in_memory_ts
 
     tg = s.task_groups[y.name]
     assert tg.states["memory"] == 5
 
     assert s.task_groups[y.name].dependencies == {s.task_groups[x.name]}
 
     await c.replicate(y)
+    # TODO: Are we supposed to track repliacted memory here? See also Scheduelr.add_keys
     assert tg.nbytes_in_memory == y.nbytes
```
fjetter, line 1902: I ran into some issues with the nbytes counting and wondered if this is correct. While the test states this explicitly, I'm a bit surprised and believe there are code areas where this is treated differently

distributed/worker.py
fjetter: Right, there is even a comment about this a few lines above 🤦

distributed/worker.py
```diff
@@ -2019,6 +2030,9 @@ def ensure_communicating(self):
                     dep = deps.pop()
                     if dep.state != "fetch":
                         continue
+                    # FIXME: It should never be possible for `not dep.who_has`
+                    # since it was thrown out in the missing_deps filter above
+                    # and control of the loop was never passed off.
```
fjetter, line 2035: Different PR

fjetter:
To not release too many tasks but, release those which are _free_ I ended up introducing additional state on the scheduler `erred_on` and there is one additional attribute on worker TaskState side, `scheduler_holds_ref`. This is a subtle problem I was trying to solve in https://github.com/dask/distributed/pull/4772 slightly differently but the end result is the same. The worker needs to know whether the scheduler still requires a task or not. Unless we replicate the entire graph we otherwise do not have a chance to make the right decision. Even if we were to replicate the entire graph we probably would need to introduce something like this to give the scheduler ultimate authority. That problem will pop up again once we approach speculative task assignment.

This additional reference count is required because in the past, when tasks and dependencies were clearly separated, this reference count was implicit. Tasks were always referred to by the scheduler, dependencies were just a side product and could be forgotten asap. You will therefore notice that all "compute-task" TaskState objects have this attribute set and dependencies don't.

Tests look promising and while this is not the end of the story for the stability issues (I still have reproducing tests which deadlock and/or never finish) this is an important step.

Gentle ping to @jrbourbeau and @gforsyth for review

distributed/worker.py
fjetter: https://github.com/dask/distributed/pull/4793

distributed/scheduler.py
```diff
@@ -1199,6 +1199,10 @@ class TaskState:
        failed task is stored here (possibly itself).  Otherwise this
        is ``None``.
 
+    .. attribute:: erred_on: set(str)
+
+        Worker addresses on which errors appeard causing this task to be in an error state.
```
gforsyth, line 1204:
```suggestion
        Worker addresses on which errors appeared causing this task to be in an error state.
```

gforsyth (approved):
Hey @fjetter -- this looks good to me.  I flagged one small typo in the scheduler attributes but that doesn't need to block this.

fjetter:
Hitting an issue in the py3.8 ubu test which might be connected although I'm not sure, yet.

```
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1720, in transition_fetch_flight
    assert ts.dependents
AssertionError
distributed.worker - ERROR - 
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 2057, in ensure_communicating
    self.transition(self.tasks[d], "flight", worker=worker)
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1615, in transition
    state = func(ts, **kwargs)
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1720, in transition_fetch_flight
    assert ts.dependents
AssertionError
distributed.core - ERROR - 
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 582, in handle_stream
    func()
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 2057, in ensure_communicating
    self.transition(self.tasks[d], "flight", worker=worker)
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1615, in transition
    state = func(ts, **kwargs)
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1720, in transition_fetch_flight
    assert ts.dependents
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1615, in transition
    state = func(ts, **kwargs)
  File "/home/runner/work/distributed/distributed/distributed/worker.py", line 1773, in transition_flight_memory
    self.batched_stream.send({"op": "add-keys", "keys": [ts.key]})
  File "/home/runner/work/distributed/distributed/distributed/batched.py", line 136, in send
    raise CommClosedError()
distributed.comm.core.CommClosedError
distributed.utils - ERROR - "('sum-partial-5b2c0f6d9d691a15dc449febe739d843', 0, 1)"
  File "/home/runner/work/distributed/distributed/distributed/utils.py", line 671, in log_errors
    yield
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4042, in add_worker
    typename=types[key],
KeyError: "('sum-partial-5b2c0f6d9d691a15dc449febe739d843', 0, 1)"
distributed.core - ERROR - Exception while handling op register-worker
  File "/home/runner/work/distributed/distributed/distributed/core.py", line 501, in handle_comm
    result = await result
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4042, in add_worker
    typename=types[key],
```

This also triggers a similar error as reported in https://github.com/dask/distributed/issues/4800 with

```
distributed.utils - ERROR - "('sum-partial-5b2c0f6d9d691a15dc449febe739d843', 0, 1)"
  File "/home/runner/work/distributed/distributed/distributed/utils.py", line 671, in log_errors
    yield
  File "/home/runner/work/distributed/distributed/distributed/scheduler.py", line 4042, in add_worker
    typename=types[key],
KeyError: "('sum-partial-5b2c0f6d9d691a15dc449febe739d843', 0, 1)"
```

at least the very first assert could be related

fjetter:
I assume the above is connected since all three failures (ubu 3.8 and the two osx failures) are the same failing test. However, it's one of these incredibly rare events which is hard to reproduce locally

fjetter:
All tests green 🎉

fjetter:
friendly ping to @gforsyth and @jrbourbeau for another review. 

What changed since the last time I asked for reviews are the handlers which control task releases on the worker and how and when the scheduler invokes them. I changed the names for all of them since I consider "release" to be much too ambiguous but maybe the new names are equally confusing. Happy to give everything different names if smbd comes up with a suggestion

gforsyth (commented):
quick first pass -- I'll try to find time to dig in a bit deeper

distributed/worker.py
```diff
@@ -1671,15 +1727,21 @@ def transition_new_waiting(self, ts):
                 pdb.set_trace()
             raise
 
-    def transition_new_fetch(self, ts):
+    def transition_new_fetch(self, ts, who_has):
         try:
             if self.validate:
                 assert ts.state == "new"
                 assert ts.runspec is None
+                # assert who_has
```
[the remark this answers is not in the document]
fjetter: Right, I had some code in `add_task` at some point which removes `self.address` from who_has in case the task was not im memory. I should check this again. This is important.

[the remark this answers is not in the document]
fjetter: The assert is back in. This edge case is exclusively handled in update_who_has and we will emit a log if this is encountered. At least I couldn't reproduce any wrongful state anymore

fjetter:
@jrbourbeau and @gforsyth can you please indicate whether I can expect a review? If you are too busy that's fine. I just would like to know if I am waiting for something to happen. Any further work on the worker is kind of blocked by this PR and I am confident that this improves stability.

fjetter:
OSX test failure is unrelated and connected to unclosed RPC after startup/teardown of an adaptive cluster

jrbourbeau:
I began reviewing the changes here yesterday, but didn't have much dedicated time to focus on this today. If you're confident in the changes here, then feel free to merge them in. Otherwise, I will have time to review tomorrow

distributed/scheduler.py
```diff
@@ -2736,6 +2745,11 @@ def transition_erred_released(self, key):
                 if dts._state == "erred":
                     recommendations[dts._key] = "waiting"
 
+            w_msg = {"op": "free-keys", "keys": [key], "reason": "Erred->Released"}
+            for w in ts._erred_on:
+                worker_msgs[w] = [w_msg]
```
mrocklin, line 2750: Minor nit, we've been pretty consistent around the pronoun `ws` for WorkerState objects.

```suggestion
            for ws in ts._erred_on:
                worker_msgs[ws] = [w_msg]
```

distributed/worker.py
```diff
@@ -1477,21 +1477,66 @@ def update_data(self, comm=None, data=None, report=True, serializers=None):
...
         return info
 
-    def delete_data(self, comm=None, keys=None, report=True):
-        if keys:
-            for key in list(keys):
-                self.log.append((key, "delete"))
-                self.release_key(key, cause="delete data")
+    def free_keys_comm(self, comm, keys, reason):
+        """This is a comm handler calling `free_keys`"""
+        return self.handle_free_keys(keys, reason)
```
mrocklin, line 1495: Minor nit, but maybe we can remote this method, and add a `comm=None` keyword to the `handle_free_keys` method?  It would be unfortunate if we had to have two methods for every handler.

distributed/worker.py
```diff
@@ -1477,21 +1477,66 @@ def update_data(self, comm=None, data=None, report=True, serializers=None):
...
+
+        For stronger guarantees, see handler free_keys
+        """
+        self.log.append(("Handle superfluous data", keys, reason))
+        for key in list(keys):
+            ts = self.tasks.get(key)
+            if ts and not ts.scheduler_holds_ref:
+                self.release_key(key, reason=f"delete data: {reason}", report=False)
+
+        logger.debug("Worker %s -- Deleted %d keys", self.name, len(keys))
```
mrocklin, line 1539: I'm curious about why there should be multiple methods for this.  In principle I think of our current approach as the scheduler saying 

> I don't need you to have this data 

The worker though, as you say, may disagree and hold on to it.  I don't imagine a situation where the scheduler would override the worker, and so I don't see a reason to have multiple methods for this signal. 

Thoughts?

mrocklin (commented):
A couple of minor nits, and one genuine question :)

distributed/worker.py
fjetter: At first, I had this in for debugging purposes to track where the scheduler actually called this. That's now obsolete since I am providing reasons everywhere (if we want to keep that).

Apart from this I liked this refactoring since it didn't require me to make all arguments optional which is nice for type checking. however, no strong opinions here.

On a deeper level, I'm wondering if we should allow the same functionality via stream and comm handlers. I'm inclined to say no since the semantics should then also be slightly different. For instance, comm handlers are useful if you actually require a response, otherwise a stream handler is preferable, for performance. Both is kind of weird. Thoughts?

mrocklin: To be clear, the functionality you're talking about here is access to the `comm`?  If so, I guess the question reduces to "is there ever a time when we need the comm in either case?"  It's a valid question.  I don't have a strong thought here.

distributed/worker.py
fjetter: First of all, I think this particular signal I called "superfluous data" should not exist and I believe this is what you're saying as well.
This signal is issued here (link to main branch) https://github.com/dask/distributed/blob/ddec3c9c7b5a5d3cb3b806ff25ee36ef640d8dda/distributed/scheduler.py#L6276-L6284

I am actually a bit fuzzy myself how this particular race condition occurs but something like this can happen but this is how I think it happens

```
Task A (Worker A) -->  Task C (Worker B)
           - ^
         /        
        /
Task B (Worker B)
```


* Worker B tries to execute C but needs to fetch Task A to do so
* Worker A submits data to B
* Data arrives at B and is being deserialized or is waiting for a loop, whatever
* Worker A dies and the scheduler notices
* Scheduler thinks all is lost and transitions Task A to released and tries to reschedule on Worker B
* ... In the meantime Worker B is done deserializing and puts key in memory, notifies scheduler usign `add-keys`
* Worker B receives the `compute-task` signal from scheduler and thinks he's lucky since he got the data already, cool. He tells the schdeuler -> `task-finished` signal
* Scheduler receives the add-keys response but thinks "This task is not in memory any more, please forget it" -> delete-data signal (what I call superfluous data)
* Worker B receives the `delete-data` signal and forgets *everything* (this is where my superfluous-data signal has a safeguard)
* Scheduler receives the `task-finished` signal from before and thinks everything is OK and proceeds as before, however, worker B does not have the data.


I assume this only is a problem due to another bug in how the worker corrects missing data (which I am also trying to fix) which would eventually heal the cluster but I'm not entirely certain

distributed/worker.py
fjetter: yes. some handlers are merely signals which don't require a response. for instance "hey, scheduler. just wanted to let you know that I finished computing X. Cheers." 
Signals without a need for a response, i.e. handlers which do not require a comm, should be implemented as stream handlers since there we can batch submit all messages.

Anyhow, we should not mix this into the PR. I think I will simply remove the second method :)

distributed/tests/test_failed_workers.py
```diff
@@ -393,30 +413,84 @@ async def test_worker_who_has_clears_after_failed_connection(c, s, a, b):
...
     assert not any(n_worker_address in s for ts in a.tasks.values() for s in ts.who_has)
 
     await n.close()
 
 
+@gen_cluster(
+    client=True,
+    nthreads=[("127.0.0.1", 1), ("127.0.0.1", 2), ("127.0.0.1", 3)],
+)
+async def test_worker_same_host_replicas_missing(c, s, a, b, x):
```
fjetter, line 453: This is a test case for the last deadlock we were made aware of. It requires a bit of patching since it is otherwise not possible to stimulate the same-host with multiple who-has edge case this is hitting

This happens if there are multiple who_has such that the scheduler doesn't retrigger a clean computation but the worker always pick the same erroneous worker. If the worker doesn't forget about that wrong worker after first try, we're running in a loop.

https://github.com/dask/distributed/blob/ddec3c9c7b5a5d3cb3b806ff25ee36ef640d8dda/distributed/worker.py#L2089-L2093

distributed/worker.py
mrocklin: Ha, that's a fun comedy of errors.

Yeah, so I agree that this feels like a different bug.  In principle it'd be nice to fix that bug rather than add this signal, but I can imagine that it might be best to do that in future work.

distributed/tests/test_failed_workers.py
```diff
@@ -448,3 +522,101 @@ async def test_worker_time_to_live(c, s, a, b):
...
+    def __sizeof__(self) -> int:
+        # Ensure this is offloaded to avoid blocking loop
+        import dask
+        from dask.utils import parse_bytes
+
+        return parse_bytes(dask.config.get("distributed.comm.offload")) + 1
+
+
+@gen_cluster(client=True, timeout=None)
+async def test_handle_superfluous_data(c, s, a, b):
```
fjetter, line 551: This test is rather close to the implementation since we somehow need to wait for worker B to actually start deserializing and we do not have any sane hooks exposing this at the moment, therefore I'm watching the connection pool to make this work.
see https://github.com/dask/distributed/pull/4784#discussion_r649210094

This test proves the race condition (replace handle-superfluous-data with handle-free-keys and it is stuck) but I couldn't prove that we require any handler like this at all. I kept this path to preserver "previous functionality" and added a really stupid test below which constructs the situation without any actual work, just to have the code path covered. I will try further to reconstruct a case for this but I don't want this to block the PR (see `test_forget_data_not_supposed_to_have`)

fjetter:
Win-py38: https://github.com/dask/distributed/issues/4859
osx-py37: https://github.com/dask/distributed/issues/4905

fjetter:
Since the test failures appear to be unrelated, I will go ahead and merge this now. Feel free to leave review comments afterwards, I'll address everything that comes up.

PR #4785, opened 2021-05-05
Fix doctests in utils.py

jacobtomlinson:
I was about to start some TDD with `pytest --doctest-modules distributed/utils.py` and noticed there were some failures.

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -640,7 +640,9 @@ def key_split_group(x):
     >>> key_split_group('<module.submodule.myclass object at 0xdaf372')
     'myclass'
     >>> key_split_group('x')
+    'x'
     >>> key_split_group('x-1')
+    'x'
     """
     typ = type(x)
     if typ is tuple:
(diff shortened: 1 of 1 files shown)


distributed/utils.py
```diff
@@ -1451,7 +1453,7 @@ def import_term(name: str):
 
     Examples
     --------
-    >>> import_term("math.sin")
+    >>> import_term("math.sin") # doctest: +SKIP
```
jacobtomlinson, line 1456: I chose to skip here because the actual result was `<built-in function sin>`. But we could always go the other way.

fjetter (approved):
Big +1 for TDD

distributed/utils.py
fjetter: I'm fine either way. I personally would probably only skip if it cannot be executed for some reason but I don't think this is really crucial

jacobtomlinson:
Failing test here appears to be unrelated. I'm going to go ahead and merge as it's only a docstring fix.

PR #4786, opened 2021-05-05
Allow the dashboard to run on multiple ports

### LINKED ISSUE #4782: Run dashboard on multiple ports
In some constrained environments (like one I am currently working with) it may be desirable to run the scheduler dashboard on multiple ports.

The use case I have is that ports on the cluster can only be exposed internally or externally, and it is desirable for the user to connect to the dashboard either way. So the only option is to run the dashboard on two ports and expose one internally and one externally.

My initial though for an implementation would be to allow `dashboard_address` to be a list of comma separated string and to update `start_http_server` and `clean_dashboard_address` to support that.

https://github.com/dask/distributed/blob/233ec8840671b5d705a06d0dcfe1ff42b771230b/distributed/node.py#L99

https://github.com/dask/distributed/blob/233ec8840671b5d705a06d0dcfe1ff42b771230b/distributed/utils.py#L1507

jacobtomlinson:
- [ ] Closes #4782
- [ ] Tests added / passed

### DIFF
--- a/distributed/comm/__init__.py
+++ b/distributed/comm/__init__.py
@@ -10,7 +10,7 @@ from .addressing import (
     unparse_host_port,
 )
 from .core import Comm, CommClosedError, connect, listen
-from .utils import get_tcp_server_address
+from .utils import get_tcp_server_address, get_tcp_server_addresses
 
 
 def _register_transports():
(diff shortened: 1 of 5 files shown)


jacobtomlinson:
CI failures appear unrelated.

quasiben:
Planning on merging tomorrow unless there are other comments

PR #4787, opened 2021-05-05
Support for UCX 1.10+

pentschev:
Select correct transports for UCX 1.10+ and ensure the CUDA context is created before initializing UCX when support for that is requested.

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -476,8 +476,8 @@ def _scrub_ucx_config():
             tls = "tcp,rdmacm"
             tls_priority = "rdmacm"
         else:
-            tls = "tcp,sockcm"
-            tls_priority = "sockcm"
+            tls = "tcp"
+            tls_priority = "tcp"
 
         # CUDA COPY can optionally be used with ucx -- we rely on the user
         # to define when messages will include CUDA objects.  Note:
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -53,12 +53,15 @@ def init_once():
     if ucp is not None:
         return
 
+    # remove/process dask.ucx flags for valid ucx options
+    ucx_config = _scrub_ucx_config()
+    if "TLS" in ucx_config and "cuda_copy" in ucx_config["TLS"]:
+        import numba.cuda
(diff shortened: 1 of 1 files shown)


pentschev:
cc @quasiben

pentschev:
I don't know the cause of the failing tests but I don't think they are related to this PR, could someone confirm/deny?

distributed/comm/ucx.py
```diff
@@ -53,13 +53,22 @@ def init_once():
...
+    ucx_config = _scrub_ucx_config()
+    if "TLS" in ucx_config and "cuda_copy" in ucx_config["TLS"]:
+        try:
+            import numba.cuda
+        except ImportError:
+            raise ImportError(
+                "CUDA support with UCX requires Numba for context management"
+            )
+
+        numba.cuda.current_context()
```
quasiben, line 66: I think we should put a comment here explaining why we need to create a cuda context before starting communications

pentschev: Done in https://github.com/dask/distributed/pull/4787/commits/7c199af218e3808bf6e7d144986d4e951aeb134b

quasiben: Thanks @pentschev !

pentschev:
Failure is:

```python
_______________________________ test_thread_time _______________________________

    def test_thread_time():
        start = metrics.thread_time()
        run_for(0.05)
        dt = metrics.thread_time() - start
>       assert 0.03 <= dt <= 0.2
E       assert 0.03 <= 0.020078000000012253

distributed/tests/test_metrics.py:48: AssertionError
```

Seems like a very tight timeout. Not sure what's the general policy, should we increase it slightly to `0.3` or ignore it for the time being?

quasiben:
I think we should raise that as an issue but it's safe to ignore for this PR

pentschev:
Makes sense, I raised https://github.com/dask/distributed/issues/4790 for that.

quasiben:
This also failed on `distributed/tests/test_nanny.py::test_num_fds FAILED ` and but only on mac 3.7 but also due to a timeout.  I added to issue #4790.  I think it's safe to merge this PR

PR #4789, opened 2021-05-06
Remove passing of kwargs to super in Server

### LINKED ISSUE #4788: Server node tried to pass kwargs to object
While debugging an issue in dask-cloudprovider I noticed that `distributed.core.Server` tries to pass any remaining kwargs to `object` which doesn't take any kwargs.

https://github.com/dask/distributed/blob/7bfd1ae202bb973aea36da0e0d41a7377d2a47ab/distributed/core.py#L241

This came about from `dask-spec` being given kwargs which do not exist in `distributed.Worker`.

jacobtomlinson:
- [x] Closes #4788
- [ ] Tests added / passed

The `super()` of `distributed.core.Server` is `object` which doesn't take any kwargs. So if there are any kwargs left over in `Server` it raises an unhelpful error.

This PR raises more helpful exceptions, but we may want to consider just removing  the `**kwargs` from the `Server.__init__` all together.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -167,6 +167,10 @@ class Server:
         self.io_loop = io_loop or IOLoop.current()
         self.loop = self.io_loop
 
+        if kwargs:
+            for key in kwargs:
+                raise NameError(f"{self.__class__.__name__} has no such option {key}.")
+
         if not hasattr(self.io_loop, "profile"):
             ref = weakref.ref(self.io_loop)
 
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @jacobtomlinson! This is similar to https://github.com/dask/distributed/pull/4733 where it was proposed that we remove extra `kwargs` to `Server` altogether.

jacobtomlinson:
Ah awesome, I hadn't spotted that. Closing this in favour of #4733.

Issue #4058, opened 2020-08-20
Cluster HTML repr doesn't respect JupyterLab themes


gjoseph92:
Transferring https://github.com/coiled/coiled-issues/issues/51 to this project.

**What happened**:

Without ipywidgets installed, the HTML repr of a `Cluster` is hard to read in a dark theme:
![image](https://user-images.githubusercontent.com/11656932/90829505-7b3a9580-e305-11ea-991f-66650a9d245e.png)

**What you expected to happen**:

The HTML repr would display using colors from the Jupyter theme, for easier readability.

**Minimal Complete Verifiable Example**:

Set the JupyterLab theme to dark (Settings > JupyterLab theme) and run in a notebook:

```python
from dask.distributed import LocalCluster

cluster = LocalCluster()
cluster
```

**Anything else we need to know?**:

Rather than hardcoding colors, you could pull them from CSS variables defined by the theme; something like

```css
background-color: var(--jp-cell-editor-background);
border: var(--jp-border-width) solid var(--jp-cell-editor-border-color);
display: inline-block;
padding: 10px;
```

(I'm sure there are canonically correct CSS variables to use here, I just grabbed whatever looked right.)

Using that produces:
![image](https://user-images.githubusercontent.com/3309802/90827137-e7fb6300-e2f8-11ea-9e3b-121c94c64e87.png)

![image](https://user-images.githubusercontent.com/3309802/90827324-488aa000-e2f9-11ea-9483-4c05b0631cf7.png)

Downside is that these variables don't work for classic notebooks, though there are probably some other variables (or fallbacks) you could use.

**Environment**:

- Dask version: 2.23.0
- Python version: 3.7.5
- Operating System: Linux
- Install method (conda, pip, source): pip

mrocklin:
Thank you for the issue @gjoseph92 , and my apologies for the delay in response.  cc'ing @jacobtomlinson and @ian-r-rose

jacobtomlinson:
I support this, it would be useful.

This is a Jupyter Lab specific change, so we would need to test against other notebook compatible tools like classic notebooks, VSCode and Nteract to ensure things have sensible defaults.

ian-r-rose:
Sorry, I missed this the first time around -- it should be fairly straightforward to do this. We would just have to inline the CSS variables in the IPython reprs (with fallback values so it will be rendered correctly in non-JupyterLab environments)

PR #4795, opened 2021-05-07
Add HTML reprs to some scheduler classes

jrbourbeau:
Occasionally, in particular when using a `LocalCluster`, I find myself printing things `WorkerState` and `TaskState` objects when in a Jupyter notebook. This PR adds HTML reprs to some of the classes used by the scheduler to improve user's experience when in Jupyter. 

I'm sure there are further improvements we can make, but even just porting their existing `__repr__`s over to HTML has a more polished look. 

Screen Shot 2021-05-07 at 4 16 17 PM
Screen Shot 2021-05-07 at 4 16 35 PM
Screen Shot 2021-05-07 at 4 16 52 PM

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1,4 +1,5 @@
 import asyncio
+import html
 import inspect
 import itertools
 import json
@@ -732,12 +733,21 @@ class WorkerState:
         return ws
 
     def __repr__(self):
-        return "<Worker %r, name: %s, memory: %d, processing: %d>" % (
+        return "<WorkerState %r, name: %s, memory: %d, processing: %d>" % (
             self._address,
             self._name,
             len(self._has_what),
             len(self._processing),
         )
+
+    def _repr_html_(self):
+        text = (
+            f"<b>WorkerState: {html.escape(self._address)}</b> "
+            f'<font color="gray">name: </font>{self.name} '
+            f'<font color="gray">memory: </font>{len(self._has_what)} '
+            f'<font color="gray">processing: </font>{len(self._processing)}'
+        )
+        return text
 
     @ccall
     @exceptval(check=False)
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -1549,7 +1559,18 @@ def set_nbytes(self, nbytes: Py_ssize_t):
         self._nbytes = nbytes
 
     def __repr__(self):
-        return "<Task %r %s>" % (self._key, self._state)
+        return "<TaskState %r %s>" % (self._key, self._state)
```
jrbourbeau, line 1562: Also changed the name here so it matches the actual class name, which I have a preference for even though it's a little longer. Though this isn't a strong preference

ian-r-rose:
Always good to see more rich reprs. It might be nice to give these colors drawn from JupyterLab theme variables, e.g, [here](https://github.com/dask/dask-labextension/blob/d771e0f4834b96ca83c5769fe8585ec523d5321e/style/index.css#L10-L12). This way the text would show up as light when using darker themes.

(For non-JupyterLab notebook renderers, you could add [fallback variable values](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#custom_property_fallback_values))

mrocklin:
I would de-emphasize the task name from bold to even possibly greyed out.  It's mostly a lot of noise.  

With infinite dev time (not the case currently), I might even try to hide them or only make them show on hover or something (Although this may not be possible with straight HTML/CSS)

I wonder if maybe we should do summary/details tags where the current information that we have is in the summary and details includes a larger table (also future looking)

jrbourbeau:
>  It might be nice to give these colors drawn from JupyterLab theme variables, e.g, here. This way the text would show up as light when using darker themes.

Totally agree, however I'd prefer to have us handle that in a follow-up PR. If you have thoughts on a good way we can pull colors from the JupyterLab theme, could you leave a comment over in https://github.com/dask/distributed/issues/4058?

ian-r-rose:
> Totally agree, however I'd prefer to have us handle that in a follow-up PR. If you have thoughts on a good way we can pull colors from the JupyterLab theme, could you leave a comment over in #4058?

Sure, happy to follow-up over there. No actual packaging should be involved, just inlining the CSS variables in the HTML string.

jrbourbeau:
I did some de-emphasizing and moved the `TaskState` key to the end of the repr (though I think it's still useful to keep around)

Screen Shot 2021-05-07 at 5 11 42 PM

jrbourbeau:
> I wonder if maybe we should do summary/details tags where the current information that we have is in the summary and details includes a larger table (also future looking)

+1 having a details section would be nice

PR #4796, opened 2021-05-07
Use JupyterLab theme variables for HTML reprs

### LINKED ISSUE #4058: Cluster HTML repr doesn't respect JupyterLab themes

ian-r-rose:
Includes fallback colors for other notebook viewers.

- [x] Closes #4058

image

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -399,19 +399,27 @@ class Future(WrappedKey):
     def _repr_html_(self):
         text = "<b>Future: %s</b> " % html.escape(key_split(self.key))
         text += (
-            '<font color="gray">status: </font>'
-            '<font color="%(color)s">%(status)s</font>, '
+            '<font style="color: var(--jp-ui-font-color2, gray)">status: </font>'
+            '<font style="color: %(color)s">%(status)s</font>, '
         ) % {
             "status": self.status,
-            "color": "red" if self.status == "error" else "black",
+            "color": "var(--jp-error-color0, red)"
+            if self.status == "error"
+            else "var(--jp-ui-font-color0, black)",
         }
         if self.type:
             try:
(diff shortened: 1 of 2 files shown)


mrocklin:
Nice.

jacobtomlinson (approved):
This is awesome thanks @ian-r-rose.

Ci failures appear unrelated.

jacobtomlinson:
@ian-r-rose out of curiosity are these CSS variables documented somewhere?

ian-r-rose:
> @ian-r-rose out of curiosity are these CSS variables documented somewhere?

Not particularly well -- the main documentation is probably the [comments](https://github.com/jupyterlab/jupyterlab/blob/master/packages/theme-light-extension/style/variables.css) in the default theme.

Issue #4766, opened 2021-04-29
Flaky `test_workspace_concurrency_intense`


jrbourbeau:
For example, see [this CI build](https://github.com/dask/distributed/pull/4763/checks?check_run_id=2459092140)

<details>
<summary>Full traceback:</summary>

```python
______________________ test_workspace_concurrency_intense ______________________

tmpdir = local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_workspace_concurrency_int10')

    @pytest.mark.flaky(reruns=10, reruns_delay=5, condition=MACOS)
    @pytest.mark.slow
    def test_workspace_concurrency_intense(tmpdir):
>       n_created, n_purged = _test_workspace_concurrency(tmpdir, 8.0, 16)

distributed/tests/test_diskutils.py:287: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tmpdir = local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_workspace_concurrency_int10')
timeout = 8.0, max_procs = 16

    def _test_workspace_concurrency(tmpdir, timeout, max_procs):
        """
        WorkSpace concurrency test.  We merely check that no exception or
        deadlock happens.
        """
        base_dir = str(tmpdir)
    
        err_q = mp_context.Queue()
        purged_q = mp_context.Queue()
        stop_evt = mp_context.Event()
        ws = WorkSpace(base_dir)
        # Make sure purging only happens in the child processes
        ws._purge_leftovers = lambda: None
    
        # Run a bunch of child processes that will try to purge concurrently
        NPROCS = 2 if sys.platform == "win32" else max_procs
        processes = [
            mp_context.Process(
                target=_workspace_concurrency, args=(base_dir, purged_q, err_q, stop_evt)
            )
            for i in range(NPROCS)
        ]
        for p in processes:
            p.start()
    
        n_created = 0
        n_purged = 0
        try:
            t1 = time()
            while time() - t1 < timeout:
                # Add a bunch of locks, and simulate forgetting them.
                # The concurrent processes should try to purge them.
                for i in range(50):
                    d = ws.new_work_dir(prefix="workspace-concurrency-")
                    d._finalizer.detach()
[...]

PR #4805, opened 2021-05-11
Let servers close faster if there are no active handlers

fjetter:
This is another marginal fix for test runtimes.

we have a hard coded grace period allowing incoming connections to finish on their own in case we want to close before we pull out the hammer.
A few things are a bit off here
1. If there is a remote Connectionpool, these comms will always remain open and they will not close themselves unless the remote closes as well. Therefore, this will *always* wait no matter what
2. If there are a bunch of comms just waiting for the remote to do stuff, there is frankly no big point in waiting. There is the edge case where we might be just in the process of reading an incoming stream but chances are very high that we will need to close something like that off anyhow
3. The only valid reason I can see for this is that if there are active handlers still running (if that is even sensible) we'll give them a chance.

This hard coded wait period causes effectively every single `await Worker.close()` to block for a second if the worker has had any incoming connections before. For some tests this makes the difference between subsecond and multi second runtime.

I think we should kick this waiting time out completely but either way, this give the entire process tests and a bit of a bigger purpose than before.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -525,6 +525,8 @@ class Server:
                             e,
                         )
                         break
+
+                self._comms[comm] = None
                 msg = result = None
                 if close_desired:
                     await comm.close()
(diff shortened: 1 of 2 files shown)


fjetter:
osx failure again https://github.com/dask/distributed/issues/4766

jrbourbeau (approved):
Thanks @fjetter! There's a small merge conflict (after merging https://github.com/dask/distributed/pull/4733), but otherwise this PR looks good to me

PR #4808, opened 2021-05-11
Add support for registering scheduler plugins from Client

### LINKED ISSUE #4745: Add `Client.register_scheduler_plugin`
Today scheduler plugins are added to a cluster by calling the `Scheduler.add_plugin` method. I often see plugins added at scheduler startup through a preload script.

However, it's also useful to be able to add a scheduler plugin to an existing, already running scheduler. In these cases one often doesn't have direct access to the `Scheduler` instance and ends up calling `client.run_on_scheduler` to add their plugin:

```python
def register_plugin(plugin):
    await plugin.start()
    dask_scheduler.add_plugin(plugin)

client.run_on_scheduler(register_plugin)
```

This works, but is a little awkward. Additionally, some users might not be aware that `client.run_on_scheduler` exists.

It would be nice if we had a `client.register_scheduler_plugin`, which was similar to our existing `client.register_worker_plugin`, to allow users to register their own scheduler plugins in a more straightforward way. 

cc @douglasdavis who may be interested in adding a `client.register_scheduler_plugin` method

douglasdavis:
- [x] Closes #4745
- [x] Tests added / passed

@martindurant


martindurant:
You will want to ping a couple of specific people for feedback here, or likely no one will notice so long as it is marked draft.

douglasdavis:
Thanks for the tip! -- pinging @jrbourbeau

distributed/client.py
```diff
@@ -3960,6 +3960,31 @@ async def _get_task_stream(
...
+
+        """
+
+        async def f(dask_scheduler=None, plugin=None, idempotent=None, **kwargs):
+            if isinstance(plugin, type):
+                plugin = plugin(**kwargs)
+            await plugin.start()
+            dask_scheduler.add_plugin(plugin=plugin, idempotent=idempotent, **kwargs)
+
+        self.run_on_scheduler(f, plugin=plugin, idempotent=idempotent, **kwargs)
```
fjetter, line 3983: I would suggest to define a new _handler_ in the scheduler (see `Scheduler.handlers`, there is also the same thing for the worker plugin) and the logic of `f` is then executed by the handler. This way, the client call here is actually quite trivial

Defining `f` in this local scope here is indeed sometimes problematic since it sometimes introduces serializability issues and a global scope (class method or global function) is usually better when serializing stuff

fjetter (commented):
> Worker plugins use a name which is useful for unregistering. As of now the scheduler plugins are stored in a Python list, so right now I can image removal perhaps being a bit tricky without a name. Should we work on supporting named scheduler plugins?

The idempotency logic currently uses types to infer whether a plugin is already there or not. The remove, however, is a bit more simple and only works if you are trying to remove the identical instance so there is room for improvement.
I'd be open for a named plugin system but am a bit afraid of the idempotency problem and am wondering if this is worth it or if it makes everything more difficult to use if we introduce names. What if people try to register the same plugin but using a different name? What if there are different plugin types trying to use the same name?

distributed/client.py
```diff
@@ -3961,6 +3961,39 @@ async def _get_task_stream(
...
+
+    def register_scheduler_plugin(self, plugin=None, idempotent=False, **kwargs):
+        """Register a scheduler plugin.
+
+        See https://distributed.readthedocs.io/en/latest/plugins.html
+
+        Parameters
+        ----------
+        plugin : SchedulerPlugin
+            Plugin class (not instance) to pass to the scheduler.
```
jrbourbeau, line 3980: It looks like `plugin` can be either a class or instance (which is the behavior we want)

distributed/client.py
```diff
@@ -3961,6 +3961,39 @@ async def _get_task_stream(
...
+            plugin = plugin(**kwargs)
+
+        return await self.scheduler.register_scheduler_plugin(
+            plugin=dumps(plugin, protocol=4)
+        )
+
+    def register_scheduler_plugin(self, plugin=None, idempotent=False, **kwargs):
+        """Register a scheduler plugin.
+
+        See https://distributed.readthedocs.io/en/latest/plugins.html
```
jrbourbeau, line 3975: Nitpick: we might instead link directly to the scheduler plugin section https://distributed.readthedocs.io/en/latest/plugins.html#scheduler-plugins

distributed/scheduler.py
```diff
@@ -5205,6 +5207,10 @@ def remove_plugin(self, plugin):
         """Remove external plugin from scheduler"""
         self.plugins.remove(plugin)
 
+    def register_scheduler_plugin(self, comm=None, plugin=None):
+        """TODO docstring."""
+        self.add_plugin(plugin=loads(plugin))
```
jrbourbeau, line 5212: Since this is one of the few places we run `pickle` on the scheduler, we'll want to first check the `distributed.scheduler.pickle` configuration value to make sure users haven't explicitly disallowed the scheduler from running `pickle` (which can result in arbitrary code being executed). For an example of another place where this occurs, see `Scheduler.run_function`:

https://github.com/dask/distributed/blob/b0a0eea957c8f473475a7699607643e55aecb700/distributed/scheduler.py#L6529-L6534

distributed/diagnostics/tests/test_scheduler_plugin.py
```diff
@@ -125,3 +125,16 @@ async def close(self):
...
+
+@gen_cluster(client=True)
+async def test_register_scheduler_plugin(c, s, a, b):
+    temp_list = []
+
+    class Dummy(SchedulerPlugin):
+        def start(self, scheduler):
+            temp_list.append(1)
+
+    await c._register_scheduler_plugin(Dummy)
```
jrbourbeau, line 138: The use `self.sync` in `Client.register_scheduler_plugin` will allow us to use it in both sync and async contexts:

```suggestion
    await c.register_scheduler_plugin(Dummy)
```

distributed/diagnostics/tests/test_scheduler_plugin.py
```diff
@@ -125,3 +125,16 @@ async def close(self):
...
+
+    async_wait_for(temp_list, 1)
```
jrbourbeau, line 140: Since `async_wait_for` is a coroutine, we still need to `await` it 

```suggestion
    await async_wait_for(temp_list, 1)
```

distributed/scheduler.py
jrbourbeau, line 5212: We should add some additional logic here such that, similar to `WorkerPlugin.setup`, `SchedulerPlugin.start` also runs when a new plugin is dynamically added to the scheduler through `Client.register_scheduler_plugin`. FWIW that's one reason why `test_register_scheduler_plugin` isn't currently passing

jrbourbeau (commented):
Thanks for the updates @douglasdavis! Apologies for the delayed review, this PR is looking good

> Worker plugins use a name which is useful for unregistering. As of now the scheduler plugins are stored in a Python list, so right now I can image removal perhaps being a bit tricky without a name. Should we work on supporting named scheduler plugins?

That's a good point. I think we should work on supporting named scheduler plugins similar to how we already treat worker plugins. Though initially we'll want to have some compatibility code to support the current behavior so there's a smooth transition for any current users of `Scheduler.remove_plugin`.

jrbourbeau:
Also, there is a small linting issue in CI. You can [install pre-commit hooks](https://docs.dask.org/en/latest/develop.html#code-formatting) to make sure things like `flake8`, `black`, etc. are automatically run when you make a `git commit` locally

distributed/scheduler.py
douglasdavis: I'm not exactly following here. What part of the worker plugin registration API is connected to `WorkerPlugin.setup`? Does the `_WorkerSetupPlugin` class down at https://github.com/dask/distributed/blob/05c56214cca6bccb0de0cbf85d7f01990f6923d8/distributed/client.py#L4129
have something to due with you're mentioning here?

jrbourbeau: Today the `setup` method for `WorkerPlugin`s are [run in two scenerios](https://distributed.dask.org/en/latest/plugins.html#distributed.diagnostics.plugin.WorkerPlugin.setup):

1. When a new worker starts up, the `setup` method for all currently registered `WorkerPlugin`s will be run
2. When a new `WorkerPlugin` is registered, the corresponding `setup` method is run on all existing workers

https://github.com/dask/distributed/blob/05c56214cca6bccb0de0cbf85d7f01990f6923d8/distributed/worker.py#L2796-L2804

For `SchedulerPlugin`s we want to have a similar behavior. Since there's only ever a single scheduler, we don't need to worry about the first case above because we won't ever have new schedulers arriving in the cluster. But we do want to make sure that, like the second case above, when a new `SchedulerPlugin` is registered, that the corresponding `start` method (there's a slight discrepancy in the `start` / `setup` method name used between scheduler and worker plugins) is run on the scheduler. Does that help clarify things?

jrbourbeau: To be clear, the reason we want to make sure `WorkerPlugin.setup` / `Scheduler.start` are run when a new plugin is registered is because often times those `setup`/`start` methods contain setup code that must be run before other plugin methods (e.g. `add_worker` or `transition`) will work

douglasdavis: Ah yes I see thanks for the detailed explanation! OK I just added the logic to call `plugin.start(self)` in `Scheduler.register_scheduler_plugin`. I'm still trying to figure out how best to test this, but do let me know if the `start` logic is going in the right direction.

douglasdavis: After rereading the commit, should the logic be in `Scheduler.add_plugin` over `Scheduler.register_scheduler_plugin`?

martindurant: That might break existing usage where, presumably, the caller had to ensure `start()` happened. I don't know if there are any of these in practice; or indeed how you could tell whether `start` has been called. It probably should be called immediately after instantiation (or deserialisation).

distributed/diagnostics/tests/test_scheduler_plugin.py
```diff
@@ -125,3 +125,16 @@ async def close(self):
...
+async def test_register_scheduler_plugin(c, s, a, b):
+    temp_list = []
+
+    class Dummy(SchedulerPlugin):
+        def start(self, scheduler):
+            temp_list.append(1)
+
+    await c.register_scheduler_plugin(Dummy)
+
+    await async_wait_for(temp_list, 1)
```
jrbourbeau, line 140: I think we can fix/simplify this test a bit by having `Dummy.start` set some attribute on the scheduler and then `assert` that the attribute is as expected after `register_scheduler_plugin` has been called. For example, something like:

```diff
diff --git a/distributed/diagnostics/tests/test_scheduler_plugin.py b/distributed/diagnostics/tests/test_scheduler_plugin.py
index 2baa914b..ef846455 100644
--- a/distributed/diagnostics/tests/test_scheduler_plugin.py
+++ b/distributed/diagnostics/tests/test_scheduler_plugin.py
@@ -129,12 +129,11 @@ async def test_lifecycle(cleanup):

 @gen_cluster(client=True)
 async def test_register_scheduler_plugin(c, s, a, b):
-    temp_list = []

     class Dummy(SchedulerPlugin):
         def start(self, scheduler):
-            temp_list.append(1)
+            scheduler.foo = "bar"

+    assert not hasattr(s, "foo")
     await c.register_scheduler_plugin(Dummy)
-
-    await async_wait_for(temp_list, 1)
+    assert s.foo == "bar"
```

distributed/scheduler.py
```diff
@@ -5220,6 +5230,27 @@ def remove_plugin(self, plugin):
         """Remove external plugin from scheduler"""
         self.plugins.remove(plugin)
 
+    def register_scheduler_plugin(self, comm=None, plugin=None):
```
jrbourbeau, line 5233: Let's make this handler `async` (similar to `register_worker_plugin`) which will allow us to directly `await` `plugin.start` below instead of using `self.sync`:

```suggestion
    async def register_scheduler_plugin(self, comm=None, plugin=None):
```

distributed/scheduler.py
```diff
@@ -5220,6 +5230,27 @@ def remove_plugin(self, plugin):
...
+                "'distributed.scheduler.pickle' configuration setting."
+            )
+        plugin = loads(plugin)
+        self.add_plugin(plugin=plugin)
+
+        if hasattr(plugin, "start"):
+            try:
+                result = plugin.start(self)
+                if isawaitable(result):
+                    result = self.sync(result)
```
jrbourbeau, line 5249:
```suggestion
                    result = await result
```

distributed/scheduler.py
jrbourbeau: >  I'm still trying to figure out how best to test this, but do let me know if the start logic is going in the right direction.

Yeah, your updates look great 👍 The `test_register_scheduler_plugin` test you've added should already ensure that `Client.register_scheduler_plugin` results in `SchedulerPlugin.start` being called.

> That might break existing usage where, presumably, the caller had to ensure start() happened

Yep, agreed that could break some existing setups where users previously had to call `SchedulerPlugin.start` manually

douglasdavis:
Thanks for all of the help here @jrbourbeau, I've gone ahead and removed the draft label as I think this is getting pretty close. Also added support for unregistering scheduler plugins, happy to hear thoughts on that.

jrbourbeau (commented):
Thanks @douglasdavis! If we’re going to have a `Client.unregister_scheduler_plugin`, then I’d like to see us move towards using named scheduler plugins / storing them in a dictionary on the scheduler instead of a list (similar to how we treat worker plugins today) and removing them by name instead of by instance. I think this will make thing easier to reason about once users are able to register scheduler plugins from the `Client` process. 

Also, depending on how you're feeling, we can not worry about `Client.unregister_scheduler_plugin` for now. We could just limit the scope of this PR to registering scheduler plugins and take care of unregistering them in a later PR. We can definitely include that functionality here, but I just wanted to give you an exit opportunity in case there are other things you're interested in working on.

douglasdavis:
Hey @jrbourbeau -- thanks for the explanation! After reading, I agree unregistering plugins via reorganization with a dictionary (like worker plugins) is a good idea. Since the registering functionality here is ready to go, I've gone ahead removed the unregistering code in this PR (going in the limited scope direction that you mentioned). I'm happy to take on the dictionary + unregistering feature in a subsequent PR

distributed/client.py
```diff
@@ -4007,6 +4007,37 @@ async def _get_task_stream(
...
+            If multiple registrations of the plugin induce no changes.
+        **kwargs : Any
+            Arguments passed to the Plugin class (if Plugin is an
+            instance kwargs are unused).
+
+        """
+        return self.sync(
+            self._register_scheduler_plugin,
+            plugin=plugin,
+            idempotent=idempotent,
```
jrbourbeau, line 4037: Since this is a new method, and we'll probably end up deprecating `idempotent=` anyways when we go refactor to handle unregistering plugins, I'll suggest we don't support `idempotent=` from the start with `Client.register_scheduler_plugin`

jrbourbeau (commented):
Just left some final comments, but overall this looks good

distributed/client.py
douglasdavis: Sounds good, removed it.

distributed/scheduler.py
```diff
@@ -5210,6 +5219,27 @@ def remove_plugin(self, plugin):
...
+        self.add_plugin(plugin=plugin)
+
+        if hasattr(plugin, "start"):
+            try:
+                result = plugin.start(self)
+                if inspect.isawaitable(result):
+                    result = await result
+            except Exception as e:
+                msg = error_message(e)
+                return msg
```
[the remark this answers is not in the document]
douglasdavis: Moved out try block and added this to the test suite

distributed/scheduler.py
```diff
@@ -5210,6 +5219,27 @@ def remove_plugin(self, plugin):
...
+        """Register a plugin on the scheduler."""
+        if not dask.config.get("distributed.scheduler.pickle"):
+            raise ValueError(
+                "Cannot register a scheduler plugin as the scheduler "
+                "has been explicitly disallowed from deserializing "
+                "arbitrary bytestrings using pickle via the "
+                "'distributed.scheduler.pickle' configuration setting."
+            )
+        plugin = loads(plugin)
+        self.add_plugin(plugin=plugin)
```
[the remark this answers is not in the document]
douglasdavis: Ah yes definitely makes sense, I moved this line to the end of the method.

distributed/scheduler.py
```diff
@@ -5210,6 +5219,27 @@ def remove_plugin(self, plugin):
         """Remove external plugin from scheduler"""
         self.plugins.remove(plugin)
 
+    async def register_scheduler_plugin(self, comm=None, plugin=None):
+        """Register a plugin on the scheduler."""
+        if not dask.config.get("distributed.scheduler.pickle"):
```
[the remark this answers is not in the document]
douglasdavis: Added the test :+1:

douglasdavis:
Thanks @jrbourbeau! Let me know if there are any final dangling issues

distributed/diagnostics/tests/test_scheduler_plugin.py
```diff
@@ -125,3 +125,43 @@ async def close(self):
...
+    raise_match = (
+        "Cannot register a scheduler plugin as the scheduler "
+        "has been explicitly disallowed from deserializing "
+        "arbitrary bytestrings using pickle via the "
+        "'distributed.scheduler.pickle' configuration setting."
+    )
+
+    n_plugins = len(s.plugins)
+    with pytest.raises(ValueError, match=raise_match):
+        await c.register_scheduler_plugin(Dummy1)
```
jrbourbeau, line 166: Nitpick: Typically we don't explicitly check the full error message which is raised and instead check that one or more key phrases are contained within the error message. This tends to make tests more robust (though not always) to small changes in the message that is raised. 

For this specific test the main things we want to make sure is in the error message is that the scheduler is configured in a way such that it can't deserialize with pickle. To handle this, we might check that `disallowed from deserializing` and `distributed.scheduler.pickle` (the corresponding configuration setting) are included in the error message which is raised: 

```diff
diff --git a/distributed/diagnostics/tests/test_scheduler_plugin.py b/distributed/diagnostics/tests/test_scheduler_plugin.py
index e19c0a34..73d380af 100644
--- a/distributed/diagnostics/tests/test_scheduler_plugin.py
+++ b/distributed/diagnostics/tests/test_scheduler_plugin.py
@@ -154,14 +154,11 @@ async def test_register_scheduler_plugin_pickle_disabled(c, s, a, b):
         def start(self, scheduler):
             scheduler.foo = "bar"

-    raise_match = (
-        "Cannot register a scheduler plugin as the scheduler "
-        "has been explicitly disallowed from deserializing "
-        "arbitrary bytestrings using pickle via the "
-        "'distributed.scheduler.pickle' configuration setting."
-    )
-
     n_plugins = len(s.plugins)
-    with pytest.raises(ValueError, match=raise_match):
+    with pytest.raises(ValueError) as excinfo:
         await c.register_scheduler_plugin(Dummy1)
+    msg = str(excinfo.value)
+    assert "disallowed from deserializing" in msg
+    assert "distributed.scheduler.pickle" in msg
+
     assert n_plugins == len(s.plugins)
```

jrbourbeau (commented):
Thanks @douglasdavis! Just one minor comment, otherwise this is good to merge

distributed/diagnostics/tests/test_scheduler_plugin.py
douglasdavis: thanks for explaining!

distributed/client.py
```diff
@@ -4007,6 +4007,36 @@ async def _get_task_stream(
...
+        """Register a scheduler plugin.
+
+        See https://distributed.readthedocs.io/en/latest/plugins.html#scheduler-plugins
+
+        Parameters
+        ----------
+        plugin : SchedulerPlugin
+            Plugin class or object to pass to the scheduler.
+        idempotent : bool
+            If multiple registrations of the plugin induce no changes.
```
jrbourbeau, line 4028: Whoops sorry I missed that this description was still here 

```suggestion
```

distributed/client.py
```diff
@@ -4007,6 +4007,36 @@ async def _get_task_stream(
...
 
+    async def _register_scheduler_plugin(self, plugin=None, **kwargs):
+        if isinstance(plugin, type):
+            plugin = plugin(**kwargs)
+
+        return await self.scheduler.register_scheduler_plugin(
+            plugin=dumps(plugin, protocol=4)
+        )
+
+    def register_scheduler_plugin(self, plugin=None, **kwargs):
```
jrbourbeau, line 4018: Users should always need to specify `plugin` (whether it's a class or an instance)

```suggestion
    def register_scheduler_plugin(self, plugin, **kwargs):
```

distributed/client.py
```diff
@@ -4007,6 +4007,36 @@ async def _get_task_stream(
         else:
             return msgs
 
+    async def _register_scheduler_plugin(self, plugin=None, **kwargs):
```
jrbourbeau, line 4010: Similar comment here 

```suggestion
    async def _register_scheduler_plugin(self, plugin, **kwargs):
```

douglasdavis:
Ah, thanks for catching-- adjustments made

jrbourbeau (approved):
Thanks @douglasdavis! This looks great, will merge after CI finishes

Issue #4809, opened 2021-05-12
dask-scheduler and dask-worker CLI break with `click==8.0.0` 


ManuelBahr:
**What happened**:

When using `click=8.0.0` which was released on May 11th, the scheduler as well as worker CLI don't come up.

**What you expected to happen**:

Working scheduler and worker CLIs

**Minimal Complete Verifiable Example**:

```bash
$ dask-scheduler --version
ValueError: 'default' must be a list when 'multiple' is true.
$ dask-worker --version
AttributeError: module 'click._unicodefun' has no attribute '_verify_python3_env'
```

**Anything else we need to know?**:

**Probable root cause for dask-scheduler**:

Click changelog: 

> The default value for a parameter with nargs > 1 and multiple=True must be a list of tuples. #1649

https://github.com/dask/distributed/blob/5e150aa6a3bd7decd38193a3d754497f8914d83c/distributed/cli/dask_scheduler.py#L104

`dask-worker` doesn't set a default for multiple options and therefore is not affected by this, but shows a different failure.

**Probable root cause for dask-worker**:

https://github.com/dask/distributed/blob/5e150aa6a3bd7decd38193a3d754497f8914d83c/distributed/cli/utils.py#L22

**Environment**:

- `pip install distributed` reproduces the issue
- Click version needs to be: 8.0.0

quasiben:
Thanks for the report @ManuelBahr .  Confirmed that this is an issue:

```
mamba create -n dask-click-test -c conda-forge python
conda activate dask-click-test
python -m pip install distributed
```

Do you know how challenging it would be to support older and latest versions of click ?

ManuelBahr:
The fix for `multiple` is trivial since the new version of clicks seems to enforce a restriction that would have made sense earlier. The issue is that all older versions of `dask` will suffer from this as well with the newest `click`. Not sure how you would go about that.

For the `dask-worker` I believe that code fragment that `distributed` used was something internal to click. Not sure how that looks like nowadays.

quasiben:
Probably the easiest thing to do right now is set a pin for pip / conda and restrict to less than 8.X.

jrbourbeau:
Thanks for reporting @ManuelBahr and investigating @quasiben! I opened up a PR over in https://github.com/dask/distributed/pull/4810 which proposes we add some compatibility code to handle the changes in `click` 8

ManuelBahr:
Do you plan to backport those changes to the last version that supported Python3.6?

[...]

PR #4810, opened 2021-05-12
Add support for `click` 8

jrbourbeau:
This PR adds some compatibility code needed for the recent `click=8` release. Specifically:

- The name of internal `click` function we were using changed
- What constituents a valid default value when using `multiple=True` changed. Though, we should be able to just drop the existing default we're using and maintain the current behavior. 

cc @quasiben 

Closes https://github.com/dask/distributed/issues/4809

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,6 +60,12 @@ jobs:
               python -c "from distributed.scheduler import COMPILED; assert not COMPILED"
           fi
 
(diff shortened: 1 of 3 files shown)


quasiben:
Thanks @jrbourbeau !

jrbourbeau:
CI failures was `test_AllProgress_lost_key` (a known flaky test). Planning to merge after CI finishes since tests passed with `click=8` and `click<8`

PR #4813, opened 2021-05-13
Locally rerun successfully completed futures

ArtinSarraf:
- [ ] Closes #4804 
- [ ] Tests added / passed

Initial concept for locally rerunning a successfully completed future (rerunning failed futures functionality already exists). (Discussed in ticket #4804). Reason you'd want to do this is, for example, if the task completes successfully but is giving results you wouldn't expect. With this PR you can now do something like this:

```
>>> import pdb

>>> pdb.runcall(client.recreate_task_locally, future)
(Pdb) b path/to/file:<line no>
(Pdb) c
<etc.>
```

I believe this functionality is a superset of recreate_error_locally, but leaving that in place, in case you want to ensure the future you're calling actually did fail.

In the interest of reusing as much code as possible I reimplemented some of the existing methods with a more general version that takes a callback so the general recreate_tasks/error can pass in their relevant differences with the callback. Not married to this idea though.

### DIFF
--- a/distributed/recreate_tasks.py
+++ b/distributed/recreate_tasks.py
@@ -2,7 +2,7 @@ import logging
 
 from dask.utils import stringify
 
-from .client import futures_of, wait, Future
+from .client import futures_of, wait
 from .utils import sync
 from .utils_comm import pack_data
 from .worker import _deserialize
@@ -41,10 +41,7 @@ class ReplayTaskScheduler:
     def get_runspec(self, *args, key=None, **kwargs):
         key = self._process_key(key)
         ts = self.scheduler.tasks.get(key)
-        return {
-            "task": ts.run_spec,
-            "deps": [dts.key for dts in ts.dependencies]
-        }
+        return {"task": ts.run_spec, "deps": [dts.key for dts in ts.dependencies]}
 
 
 class ReplayTaskClient:
@@ -61,7 +58,9 @@ class ReplayTaskClient:
         self.client = client
         self.client.extensions["replay-tasks"] = self
         # monkey patch
-        self.client._get_raw_components_from_future = self._get_raw_components_from_future
+        self.client._get_raw_components_from_future = (
+            self._get_raw_components_from_future
+        )
         self.client._prepare_raw_components = self._prepare_raw_components
         self.client._get_components_from_future = self._get_components_from_future
         self.client._get_errored_future = self._get_errored_future
@@ -77,8 +76,12 @@ class ReplayTaskClient:
         For a given future return the func, args and kwargs and future
         deps that would be executed remotely.
         """
-        await wait(future)
-        spec = await self.scheduler.get_runspec(key=future.key)
+        if isinstance(future, str):
+            key = future
+        else:
+            await wait(future)
+            key = future.key
+        spec = await self.scheduler.get_runspec(key=key)
         deps, task = spec["deps"], spec["task"]
         if isinstance(task, dict):
             function, args, kwargs = _deserialize(**task)
@@ -150,8 +153,8 @@ class ReplayTaskClient:
         futures = [f.key for f in futures_of(future) if f.status == "error"]
         if not futures:
             raise ValueError("No errored futures passed")
-        cause = await self.scheduler.get_error_cause(keys=futures)
-        return Future(cause)
+        cause_key = await self.scheduler.get_error_cause(keys=futures)
+        return cause_key
 
     def recreate_error_locally(self, future):
         """
@@ -196,7 +199,5 @@ class ReplayTaskClient:
         Nothing; the function runs and should raise an exception, allowing
         the debugger to run.
         """
-        errored_future = sync(
-            self.client.loop, self._get_errored_future, future
-        )
-        return self.recreate_task_locally(errored_future)
+        errored_future_key = sync(self.client.loop, self._get_errored_future, future)
+        return self.recreate_task_locally(errored_future_key)
--- a/distributed/recreate_tasks.py
+++ b/distributed/recreate_tasks.py
@@ -15,15 +15,14 @@ class ReplayTaskScheduler:
 
     This adds the following routes to the scheduler
 
-    *  cause_of_failure
     *  get_runspec
+    *  get_error_cause
     """
 
     def __init__(self, scheduler):
         self.scheduler = scheduler
         self.scheduler.handlers["get_runspec"] = self.get_runspec
         self.scheduler.handlers["get_error_cause"] = self.get_error_cause
-        self.scheduler.handlers["_process_key"] = self._process_key
         self.scheduler.extensions["replay-tasks"] = self
 
     def _process_key(self, key):
@@ -52,13 +51,10 @@ class ReplayTaskClient:
     """
     A plugin for the client allowing replay of remote tasks locally
 
-    Adds the following methods (and their async variants)to the given client:
+    Adds the following methods to the given client:
 
     - ``recreate_error_locally``: main user method for replaying failed tasks
     - ``recreate_task_locally``: main user method for replaying any task
-    - ``get_futures_error``: gets the task, its details and dependencies,
-        responsible for failure of the given future.
-    - ``get_futures_components``: gets the task, its details and dependencies.
     """
 
     def __init__(self, client):
@@ -71,17 +67,17 @@ class ReplayTaskClient:
         self.client._get_errored_future = self._get_errored_future
         self.client.recreate_task_locally = self.recreate_task_locally
         self.client.recreate_error_locally = self.recreate_error_locally
-        self.client._execute_task_components = self._execute_task_components
 
     @property
     def scheduler(self):
         return self.client.scheduler
 
     async def _get_raw_components_from_future(self, future):
+        """
+        For a given future return the func, args and kwargs and future
+        deps that would be executed remotely.
+        """
         await wait(future)
-        # one reason not to pass spec into this function is in case we want to
-        # expose a method to get all the components/deps in the future. This
-        # way it will be easier to do.
         spec = await self.scheduler.get_runspec(key=future.key)
         deps, task = spec["deps"], spec["task"]
         if isinstance(task, dict):
@@ -92,6 +88,9 @@ class ReplayTaskClient:
             return (function, args, kwargs, deps)
 
     async def _prepare_raw_components(self, raw_components):
+        """
+        Take raw components and resolve future dependencies.
+        """
         function, args, kwargs, deps = raw_components
         futures = self.client._graph_to_futures({}, deps)
         data = await self.client._gather(futures)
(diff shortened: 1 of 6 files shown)


distributed/recreate_tasks.py
```diff
@@ -10,17 +10,19 @@
...
 
     *  cause_of_failure
+    *  get_runspec
     """
 
     def __init__(self, scheduler):
         self.scheduler = scheduler
         self.scheduler.handlers["cause_of_failure"] = self.cause_of_failure
+        self.scheduler.handlers["get_runspec"] = self.get_runspec
         self.scheduler.extensions["exceptions"] = self
```
fjetter, line 26: We might want to rename this extension here, e.g. `replay-tasks`?

fjetter (commented):
A second opinion about this new feature would be appreciated but I think it makes sense. I'm also not entirely sure how backwards/forwards compatible we typically want these protocols to be, cc @jrbourbeau 

regarding the implementation I would suggest a slightly different approach. With the `component_getter`, `details_extractor` and `runspec_getter` arguments, I feel we are overdoing it a bit in terms of flexibility and are creating a more complex plugin than necessary.

Is there a particular reason why we'd need these hooks? My impression is that they require a very thorough understanding of the extension and the scheduler and are not well suited for a high level API. Therefore, I imagine them to be actually only used by us in this implementation. If this is correct, I would propose a slightly different interface to reduce complexity. pseudo code below but all necessary pieces to put this together are already in this PR.

```python

class ReplayTaskSchedulerExtension:
    ...

    def get_runspec(self, key) -> Dict:
        """
        Returns the runspec of a single key
        """
        ts = self.tasks[key]
        ...
        return {
            "task": ts.run_spec,
            "deps": [dts.key for dts in ts.dependencies]
        }

    def get_error_cause(self, key) -> Optional[str]:
        """
        Returns the
        """
        ts = self.tasks[key]
        return ts.exception_blame.key


class ReplayTaskClient:

    async def recreate_error_locally(self, future):
        cause = await self.scheduler.get_error_cause(future.key)
        spec_dct = await self.scheduler.get_runspec(cause.key)
        return await self._execute_spec(spec_dct)

    async def recreate_task_locally(self, future):
        spec_dct = await self.scheduler.get_runspec(future.key)
        return await self._execute_spec(spec_dct)


    async def _execute_spec(self, spec_dct):
        """Unpack and execute the response of
        `ReplayTaskSchedulerExtension.get_runspec`
        """
        # Graph to future, pack, unpack, etc
        pass

```

ArtinSarraf:
The only reason for the hooks was to maintain as much of the original structure/code as possible, while allowing for different implementations. Not sure of a proper term for it, but kinda like a functional style inheritance without classes. In any case, if we're ok with changing up the structure of the code (and API slightly), then I think your suggestions make sense. I'll take a look over the weekend.

distributed/recreate_tasks.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @property
+    def scheduler(self):
+        return self.client.scheduler
+
+    async def _get_raw_components_from_future(self, future):
+        await wait(future)
+        # one reason not to pass spec into this function is in case we want to
+        # expose a method to get all the components/deps in the future. This
+        # way it will be easier to do.
```
ArtinSarraf, line 84: @fjetter - I deviated a bit from your suggestion to reference/fetch the components by spec, for the reason stated above. This comment is referring to a method similar to the current "get_futures_error" which I removed for now for the sake of removing some noise from the PR while restructuring, but can add back. Although I think the recreate_task_locally(run=False) would provide most of that functionality (with the deps resolved instead of raw)

distributed/recreate_tasks.py
ArtinSarraf: renamed. does this need to be changed elsewhere? kinda hard to tell where references to "exceptions" in the code base are relevant to this or not.

ArtinSarraf:
@fjetter - any thoughts on the general direction of these new changes? Think it's worth pushing this current setup forward with some tests, minor api tweaks, and docstrings?

distributed/recreate_tasks.py
fjetter: I do not mind a different structure. I do care about API surface, though, and would like to only introduce features as needed.

For instance, the `run` argument is something I feel is not necessary. It adds some clutter to the function signature and changes the behaviour of the function significantly. Opposed to returning a tuple with well defined semantics, it instead will execute a function and will return _something_.
For the debugging use case it sounds like the `run=False` is actually what you are looking for. What about just exposing this as part of the API and tell users to call the function themselves?

distributed/recreate_tasks.py
```diff
@@ -0,0 +1,172 @@
...
+
+    *  cause_of_failure
+    *  get_runspec
+    """
+
+    def __init__(self, scheduler):
+        self.scheduler = scheduler
+        self.scheduler.handlers["get_runspec"] = self.get_runspec
+        self.scheduler.handlers["get_error_cause"] = self.get_error_cause
+        self.scheduler.handlers["_process_key"] = self._process_key
```
fjetter, line 26: is this handler actually used?

distributed/recreate_tasks.py
fjetter: I think not. If in doubt keep as is. Once you have a green CI you can still rename in the end

distributed/recreate_tasks.py
ArtinSarraf: Got it. I'll remove the `run` arg, so that each method only has one way of processing things. I think it would be worth it to make a final decision on the set of publicly exposed methods that we would want before I add any more commits. We have all the pieces now to make it work so deciding on this set is all that should be left.

In my eyes this is the superset of what we could potentially expose (names can be shortened, potentially overly descriptive currently),
```
get_raw_components_from_future - returns func, and raw args/kwargs including raw unresolved future inputs/deps
get_raw_components_from_errored_future - same as above but implicitly calls get_errored_future
                                         equiavelent to current implementation of get_futures_error
get_components_from_future - similar to raw_components, but args/kwargs/deps are resolved/gathered
get_components_from_errored_future - same as above but implicitly calls get_errored_future
get_errored_future - takes a future collection object and returns the future responsible for causing error
recreate_task_locally - runs the code for a future on your local machine
recreate_error_locally - same as above but implicitly calls get_errored_future
```

Personally I think the only things we should expose are: `recreate_task_locally` and `recreate_error_locally`. I feel like the case where you want just the components and not execute the future code is not common for something that's supposed to be a debugging tool, especially since you can inspect the args/kwargs through pdb, when using recreate_*_locally.

fjetter: sounds good to me. Keeping the `recreate_*_locally` methods seems intuitive to me

ArtinSarraf: @fjetter - made these changes. also added some analogous recreate_task_locally tests to the existing recreate_error_locally tests (which I also slightly updated to account for private api changes). Also added some preliminary docstrings, which could probably use some work.

distributed/recreate_tasks.py
ArtinSarraf: removed.

ArtinSarraf:
Commits above under the name "darle" are from me. Committed those from a different computer with the global name not set to mine. Looks like there's some ways to ammend this and replace the commit, but don't want to mess anything up, will leave as is.

distributed/tests/test_client.py
```diff
@@ -4723,6 +4699,98 @@ def test_recreate_error_not_error(c):
...
+
+    f = c.compute(df3)
+    function, args, kwargs = await c._get_components_from_future(f)
+    assert function(*args, **kwargs).equals(expected)
+
+
+@gen_cluster(client=True)
+async def test_recreate_task_array(c, s, a, b):
+    da = pytest.importorskip("dask.array")
+    pytest.importorskip("scipy")
```
fjetter, line 4776: this skip isn't necessary, is it?

distributed/tests/test_client.py
```diff
@@ -4723,6 +4699,98 @@ def test_recreate_error_not_error(c):
...
+@gen_cluster(client=True)
+async def test_recreate_task_collection(c, s, a, b):
+    b = db.range(10, npartitions=4)
+    b = b.map(lambda x: int(3628800 / (x + 1)))
+    b = b.persist()
+    f = c.compute(b)
+
+    function, args, kwargs = await c._get_components_from_future(f)
+    assert function(*args, **kwargs) == [3628800, 1814400, 1209600, 907200, 725760,
+                                         604800, 518400, 453600, 403200, 362880]
```
fjetter, line 4748: This will probably collide with black. I recommend setting up [pre-commit](https://pre-commit.com/) or run the formatting yourself before pushing

distributed/recreate_tasks.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        For a given future collection, return the first future that raised
+        an error.
+        """
+        await wait(future)
+        futures = [f.key for f in futures_of(future) if f.status == "error"]
+        if not futures:
+            raise ValueError("No errored futures passed")
+        cause = await self.scheduler.get_error_cause(keys=futures)
+        return Future(cause)
```
fjetter, line 154: Is this future instatiation necessary? Futures are a bit brittle objects and leak some abstraction. In particular they notify the scheduler about this and perform some reference counting. In particular the reference counting usually needs to be protected by a threading lock, see `Client._refcount_lock`. I'm wondering if this all can be implemented by just passing keys around without messing with ref counting

ArtinSarraf: Not necessary. updated to work with they key (or future) instead.

distributed/tests/test_client.py
ArtinSarraf: Copied this from the existing test_recreate_error_array test. Looks like it's necessary in that one (not sure why), but not this one. Removed.

distributed/tests/test_client.py
ArtinSarraf: updated with black formatting

ArtinSarraf:
@fjetter - the only failures from the previous CI didn't seem to be related to the changes I had made. Merged in main in case that resolves those issues. Also, any thoughts on the latest changes?

fjetter (approved):
Changes look good. I triggered CI. If nothing is popping up there, this is in. Thanks for the thorough work @ArtinSarraf !

ArtinSarraf:
@fjetter seems like the 2 failing tests seem to be hanging on something. Don't think it's related to these changes, or if it is, not obvious to me how so. Any ideas?

ArtinSarraf:
@fjetter seems like the 2 failing tests seem to be hanging on something. Don't think it's related to these changes, or if it is, not obvious to me how so. Any ideas?

ArtinSarraf:
@fjetter seems like the 2 failing tests are hanging on something. Don't think it's related to these changes, or if it is, not obvious to me how so. I can intermittently recreate the error in local tests (this one: `tornado.util.TimeoutError: Operation timed out after 60 seconds`), but it happens on random tests, and not when I rerun that specific test independently.
 Any ideas?

fjetter:
Sorry for letting this sit for so long. The seem to be unrelated. One of the failures is https://github.com/dask/distributed/issues/4859 and for the other I opened https://github.com/dask/distributed/issues/4914

Thanks for your contribution @ArtinSarraf !

ArtinSarraf:
Great, thanks for your time and help @fjetter

PR #4821, opened 2021-05-14
Increase timeout of gen_test to 30s

fjetter:
I've noticed some tests using `gen_test` are running into timeouts. We're using a timeout of 30s for all gen_cluster tests by default and think we should align here. Hopefully this helps with some flakiness

### DIFF
--- a/distributed/deploy/tests/test_adaptive.py
+++ b/distributed/deploy/tests/test_adaptive.py
@@ -312,7 +312,7 @@ async def test_adapt_down():
                 assert time() < start + 60
 
 
-@gen_test(timeout=30)
+@gen_test()
 async def test_no_more_workers_than_tasks():
     with dask.config.set(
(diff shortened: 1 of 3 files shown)


jrbourbeau (commented):
Thanks @fjetter! I'm definitely in favor of using a consistent timeout value for tests -- good catch that `gen_test` and `gen_cluster` are using different timeout durations.

I pushed a small commit to update a couple of tests which where using an larger timeout with `gen_test` which will no longer be necessary once we bump to timeout to 30 seconds. What do you think about defining a `_TEST_TIMEOUT=30` variable (or some other name) in `utils_test.py` that we can use as the default value for both `gen_cluster` and `gen_test`? That way, we'll be more likely to maintain a consistent `timeout` value in the future

fjetter:
Set up a constant as you suggested

jrbourbeau (approved):
Thanks @fjetter!

PR #4836, opened 2021-05-21
Update handling of UCX exceptions on endpoint closing

pentschev:
The changes in this PR provide an update for UCX 1.11, where we will enable endpoint error handling by default in UCX-Py. It's necessary to update exceptions raised in cases where endpoints disconnect.

The UCX-Py endpoint reuse is not anymore necessary, so we also disable that for UCX 1.11+. The primary reason it was introduced was to circumvent an issue with CUDA IPC that was resolved by https://github.com/openucx/ucx/pull/6360. Using the endpoint reuse class has also proven to be very slow, taking a long time to initialize for clusters with just a few dozen workers and pretty much unusable for a cluster in the order of 100 workers.

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -645,7 +645,7 @@ async def send_recv(comm, reply=True, serializers=None, deserializers=None, **kw
             response = await comm.read(deserializers=deserializers)
         else:
             response = None
-    except EnvironmentError:
+    except (EnvironmentError, CommClosedError):
         # On communication errors, we should simply close the communication
         force_close = True
         raise
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -860,5 +860,5 @@ properties:
           Enable UCX-Py reuse endpoints mechanism if ``True`` or if it's not specified and
           UCX < 1.11 is installed, otherwise disable reuse endpoints. This was primarily
           introduced to resolve an issue with CUDA IPC that has been fixed in UCX 1.10, but
-          can cause establishing endpoints to be very slow, particularly noticeable in
+          can cause establishing endpoints to be very slow, this is particularly noticeable in
           clusters of more than a few dozen workers.
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -855,11 +855,10 @@ properties:
           and compiled with hwloc support. Unexpected errors can occur when using
           ``"auto"`` if any interfaces are disconnected or improperly configured.
       reuse-endpoints:
+        type: [boolean, 'null']
+        description: |
           Enable UCX-Py reuse endpoints mechanism if ``True`` or if it's not specified and
           UCX < 1.11 is installed, otherwise disable reuse endpoints. This was primarily
           introduced to resolve an issue with CUDA IPC that has been fixed in UCX 1.10, but
           can cause establishing endpoints to be very slow, particularly noticeable in
           clusters of more than a few dozen workers.
-        type: [boolean, 'null']
-        description: |
-          Whether to reuse endpoints or not.
--- a/distributed/distributed-schema.yaml
+++ b/distributed/distributed-schema.yaml
@@ -855,6 +855,6 @@ properties:
           and compiled with hwloc support. Unexpected errors can occur when using
           ``"auto"`` if any interfaces are disconnected or improperly configured.
       reuse-endpoints:
-        type: boolean
-        description: | 
+        type: [boolean, 'null']
+        description: |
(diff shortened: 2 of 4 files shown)


distributed/core.py
```diff
@@ -645,7 +645,7 @@ async def send_recv(comm, reply=True, serializers=None, deserializers=None, **kw
             response = await comm.read(deserializers=deserializers)
         else:
             response = None
-    except EnvironmentError:
+    except (EnvironmentError, CommClosedError):
```
pentschev, line 648: Note here that `CommClosedError` was not being handled in this function, but even the TCP case may raise that in both [`read`](https://github.com/dask/distributed/blob/main/distributed/comm/tcp.py#L190) and [`write`](https://github.com/dask/distributed/blob/main/distributed/comm/tcp.py#L233) methods.

distributed/distributed-schema.yaml
```diff
@@ -855,6 +855,6 @@ properties:
           and compiled with hwloc support. Unexpected errors can occur when using
           ``"auto"`` if any interfaces are disconnected or improperly configured.
       reuse-endpoints:
-        type: boolean
-        description: |
```
quasiben, line 859: I think it would be helpful to include the language in this PR (at the top) for the description.

quasiben:
Thank you for continuing to push here @pentschev !  This is has been a long slog to get UCX and Dask in a better state.  @jakirkham would you mind taking a look at this PR before merging ?

distributed/distributed-schema.yaml
pentschev: Good idea, done in https://github.com/dask/distributed/pull/4836/commits/7d5db9c5ae241079d23aba64816e2f9a120e268f .

pentschev:
[The first failure is on test_AllProgress_lost_key](https://github.com/dask/distributed/pull/4836/checks?check_run_id=2640098334#step:10:2174) , which seems to be a known flaky test: https://github.com/dask/distributed/pull/4810#issuecomment-839939734 .

[The second failure is on test_map](https://github.com/dask/distributed/pull/4836/checks?check_run_id=2640098528#step:10:2164), which seems unrelated but I'm not entirely sure. It seems to raise `CancelledError` which isn't handled but is the same block where [this PR now handles `CommClosedError`](https://github.com/dask/distributed/pull/4836/files#diff-581957c552b88dd04319efca1429c0c0827daff509a115542e70b661c4c04914R648). It seems that this change couldn't be responsible for that error (but correct me if I'm mistaken), in any case it may indicate that we should also handle `CancelledError` there too. @jakirkham it seems that you were the last one to touch the [line where `CancelledError` is raise](https://github.com/dask/distributed/blob/main/distributed/comm/tcp.py#L196), would you know if we are just missing catching that exception or if this is a different issue?

distributed/core.py
jakirkham: Though it seems setting `force_close` impacts the `finally` case below. Are we sure we can remove this?

jakirkham:
Not offhand no. It would seem that error starts somewhere else and happens to be picked up on that line. IDK why the error isn't raised earlier.

distributed/core.py
pentschev: When you say "remove", what exactly are you referring to? Here I'm adding a new previously uncatched exception, which may affect UCX. I was actually comparing to conditions that may happen in TCP comms as well, then I saw this was uncatched but is totally valid (i.e., can occur).

pentschev:
> Not offhand no. It would seem that error starts somewhere else and happens to be picked up on that line. IDK why the error isn't raised earlier.

Yeah, it's possible as well. I just pushed another commit to trigger CI again, if the issue doesn't happen again it's probably just a flaky test, in which case I'll just raise an issue for this.

pentschev:
Ok, seems like all of them are just flaky tests. None of the previous ones failed this time, but the windows build for Python 3.8 seems to have timed out, which seems like a different from other flaky tests.

jakirkham:
Yeah `test_target_duration` https://github.com/dask/distributed/pull/4836/checks?check_run_id=2641782806

Going to restart CI

pentschev:
Same two flaky tests as in https://github.com/dask/distributed/pull/4836#issuecomment-846057485 now. 🙁

jakirkham:
Let's go ahead and merge. Really don't think they are related. If that turns out to be wrong, we can revisit.

jakirkham:
Thanks Peter! 😄

pentschev:
Thanks @jakirkham and @quasiben for reviews!

PR #4837, opened 2021-05-21
Memory measures for LocalCluster(worker_class=Worker)

### LINKED ISSUE #4833: Unmanaged and max memory is double-counted when workers share the same process
```python
from dask.distributed import LocalCluster, Client, Worker
cluster = LocalCluster(worker_class=Worker)
client = Client(cluster)
```
In this use case, all workers share the same process. Note that this is not the same as a single Worker with multiple threads.

The total memory usage graph at the top left of the GUI is wrong, because it counts the unmanaged and maximum memory multiple times, bluntly summing up each worker under the blind assumption each is in a different process, whereas in reality psutil.Process().get_memory_info() returns for each the data of the same PID.

crusaderky:
Closes #4833

- ``LocalCluster(worker_class=Worker)`` will not split the max_memory by the number of CPUs anymore
- GUI top-left bar will now show the correct max_memory, unmanaged_old, and unmanaged_recent
- plain-text and html repr of Client and Cluster will now show the correct max_memory

DONE: implementation
TODO: unit tests

I wonder however how much we actually care about and want to support the use case of multiple Worker objects in the same process? Who would want to use it, other than unit tests? As opposed ``LocalCluster(worker_class=Worker, processes=False)``, which spawns a single worker with multiple threads and definitely has its merit.
I wonder if, instead of all this, we should just add a check to ``LocalCluster.__init__`` that sets processes=False if the user sets worker_class=Worker?

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -889,9 +889,15 @@ class Client:
                 nworkers,
                 nthreads,
             )
-            memory = [w["memory_limit"] for w in workers.values()]
-            if all(memory):
-                text += ", memory=" + format_bytes(sum(memory))
+
+            if self.cluster is not None:
+                memory = self.cluster.memory_limit
+            else:
+                memory = [w["memory_limit"] for w in workers.values()]
+                memory = sum(memory) if all(memory) else 0
+            if memory:
+                text += ", memory=" + format_bytes(memory)
+
             text += ">"
             return text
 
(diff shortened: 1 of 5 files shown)


crusaderky:
CC @ncclementi @jrbourbeau @fjetter

fjetter:
> I wonder however how much we actually care about and want to support the use case of multiple Worker objects in the same process? 

I'm wondering the same. I'm not willing to introduce a lot of fringe edge case logic to support this use case. imho, to leverage concurrency in the same process, people should be using the worker internal threadpool.

> I wonder if, instead of all this, we should just add a check to LocalCluster.__init__ that sets processes=False if the user sets worker_class=Worker?

I think that makes the most sense. One should also not forget that this is probably also an edge case which is mostly relevant for tests (and tutorials :P)
tbh, the `processes` kwarg feels almost redundant. In it's current implementation it feels more like an alias for the Worker vs Nanny choice which may imply subprocesses but its documenation and what it actually does is rather opaque and if we found a way to remove it entirely, I'd be all in favour. I'm fine with your suggestion, though. I guess there are more important topics around than pinning down the exact semantics of that argument.

crusaderky:
Superseded by #4852

PR #4840, opened 2021-05-21
Add `performance_report` to API docs

jrbourbeau:
This is comfortably in our public interface and let's us link to the corresponding API docs throughout our documentation


quasiben:
Thanks @jrbourbeau

PR #4850, opened 2021-05-25
Fix UCX scrub config logging

### LINKED ISSUE #4845: DEBUG log messages imply UCX config options are invalid
When debugging is enabled, distributed outputs the following messages:
```
distributed.comm.ucx - DEBUG - Key: cuda_copy with value: False not a valid UCX configuration option.
distributed.comm.ucx - DEBUG - Key: tcp with value: False not a valid UCX configuration option.
distributed.comm.ucx - DEBUG - Key: nvlink with value: False not a valid UCX configuration option.
distributed.comm.ucx - DEBUG - Key: infiniband with value: False not a valid UCX configuration option.
distributed.comm.ucx - DEBUG - Key: rdmacm with value: False not a valid UCX configuration option.
distributed.comm.ucx - DEBUG - Key: net-devices with value: None not a valid UCX configuration option.
distributed.comm.ucx - DEBUG - Key: reuse-endpoints with value: None not a valid UCX configuration option.
```
When the corresponding env vars as set, such as `export DASK_UCX__CUDA_COPY=True`, the code in `ucx.py` [here](https://github.com/dask/distributed/blob/d41169bf231949f7dfb69e052a9690f7381e62f7/distributed/comm/ucx.py#L500) seems to behave correctly, but the check [here](https://github.com/dask/distributed/blob/d41169bf231949f7dfb69e052a9690f7381e62f7/distributed/comm/ucx.py#L517) still fails and prints the debug messages.

This can be seen simply by setting:
```
export DASK_LOGGING__DISTRIBUTED="DEBUG"
```
and running the scheduler like so:
```
python -m distributed.cli.dask_scheduler
```
Since the default values even cause the debug messages to be printed, there's no need to set the UCX config options to see the debug messages above.

@quasiben @pentschev

pentschev:
- [x] Closes #4845 

I believe this check was originally intended to work as a map from `DASK_UCX__*` to the actual `UCX_*` configurations. Since the original implementation the UCX configurations in Dask have evolved to aliases, so the mapping doesn't match. However, the `options` variable is what is passed to UCX, so we should verify that we're not setting variables that don't match those that UCX exposes.

### DIFF
--- a/distributed/comm/ucx.py
+++ b/distributed/comm/ucx.py
@@ -512,11 +512,9 @@ def _scrub_ucx_config():
             options["NET_DEVICES"] = net_devices
 
     # ANY UCX options defined in config will overwrite high level dask.ucx flags
-    valid_ucx_keys = list(get_config().keys())
-    for k, v in dask.config.get("ucx").items():
-        if k in valid_ucx_keys:
-            options[k] = v
-        else:
+    valid_ucx_vars = list(get_config().keys())
+    for k, v in options.items():
+        if k not in valid_ucx_vars:
             logger.debug(
                 "Key: %s with value: %s not a valid UCX configuration option" % (k, v)
             )


quasiben:
Thanks @pentschev .  I'll merge after passing

fjetter:
Is there anything we can put into a test here?

quasiben:
You could put in an invalid UCX config option then check debug output.

pentschev:
Not really, because we don't let users specify those manually and the current code will never encounter a condition where it can fail. We could do a mock test, but I don't think it's worth the time to be honest, we might just wipe that check out instead as in its current state it will never fail (unless there's a change in UCX).

jakirkham (approved):

fjetter:
I'm not familiar with the UCX code and trust your judgment. If this is fine without a test, go ahead.

quasiben:
Thank you for the comments and review @fjetter

pentschev:
Thanks all for the reviews!

PR #4853, opened 2021-05-26
Add HTML reprs for Client.who_has and Client.has_what

jacobtomlinson:
@quasiben mentioned folks are regularly asking about how to better understand what workers have what keys.

Threw together some quick HTML reprs for `Client.who_has` and `Client.has_what` to try and make things more understandable.

I followed the same approach as `distributed.utils.Logs` by subclassing a `dict` and adding an HTML repr to it. I wasn't sure if `distributed.utils` was the right place for this though, so I started `distributed.objects` for a place to put custom objects with fancy reprs. I expect this to grow and contain more variations on `dict`, `str`, etc with fancy reprs.

Happy to move `Log` and `Logs` there if that makes sense.

**has_what**
image

**who_has**
image

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -53,6 +53,7 @@ from .core import (
 )
 from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
+from .objects import HasWhat, WhoHas
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
 from .publish import Datasets
@@ -3205,7 +3206,7 @@ class Client:
             keys = list(map(stringify, {f.key for f in futures}))
         else:
             keys = None
-        return self.sync(self.scheduler.who_has, keys=keys, **kwargs)
+        return WhoHas(self.sync(self.scheduler.who_has, keys=keys, **kwargs))
 
     def has_what(self, workers=None, **kwargs):
         """Which keys are held by which workers
@@ -3239,7 +3240,7 @@ class Client:
             workers = list(workers)
         if workers is not None and not isinstance(workers, (tuple, list, set)):
             workers = [workers]
-        return self.sync(self.scheduler.has_what, workers=workers, **kwargs)
+        return HasWhat(self.sync(self.scheduler.has_what, workers=workers, **kwargs))
 
     def processing(self, workers=None):
         """The tasks currently running on each worker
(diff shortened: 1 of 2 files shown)


jakirkham:
That's really neat Jacob! Thanks for putting this together 😄

mrocklin:
If you wanted to go big here we could also improve the HTML reprs for
WorkerState and TaskState, and then start sending/using those objects on
the client side.  This would be an API change though, and so possibly not
worth it for this method.

jacobtomlinson:
I'm keen to improve HTML reprs on ALL THE THINGS! 

This seemed like a nice and simple place to start.

jacobtomlinson:
Test failure seems flaky and unrelated.

Given this is a small change to reprs only I'm going to hit merge as I want to build upon it. We can iterate on this later.

crusaderky:
This breaks has_what and who_has for asynchronous clients.
CC @jrbourbeau I understand there should be a release this friday? This looks like a showstopper.

jrbourbeau:
Thanks for flagging @crusaderky! Do you have a link to a failing CI build you can point other to?

crusaderky:
https://github.com/dask/distributed/runs/2684366276

>       return HasWhat(self.sync(self.scheduler.has_what, workers=workers, **kwargs))
E       TypeError: 'coroutine' object is not iterable

The issue is that there are no tests in master that run who_has or has_what with an asynchronous Client. I just happened to add two in my PR.

jrbourbeau:
Thanks @crusaderky -- linking to the corresponding PR for others https://github.com/dask/distributed/pull/4774

quasiben:
We are poking at this now.  Thanks @crusaderky for flagging

jacobtomlinson:
Thanks for flagging this folks.

I expect this is because `self.sync` returns an `asyncio` future when in async mode. Therefore the type conversion fails.

The fix will likely be to move the call further down into the scheduler code, as long as it plays nicely with the comm.

I'm dealing with a family emergency this afternoon, so if you want to revert this in the meantime I'm happy with that.

quasiben:
Sorry you are having troubles.  We'll take care of it.

jrbourbeau:
Just cross-referencing https://github.com/dask/distributed/pull/4860 which fixes the `TypeError: 'coroutine' object is not iterable` we were running into

PR #4858, opened 2021-05-27
WIP Relieve pytest asyncio pinning

fjetter:
This removes the pinning for pytest-asyncio

Nice side benefit is that we'll wrap all asyncio marked tests automagically with our `clean` ctx manager.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -38,9 +38,9 @@ pytest_plugins = ["distributed.pytest_resourceleaks"]
 
 @pytest.fixture()
 def event_loop():
-    from distributed.utils_test import pristine_loop
+    from distributed.utils_test import clean
 
-    with pristine_loop() as loop:
+    with clean() as loop:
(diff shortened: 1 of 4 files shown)


QuLogic:
Is there some possibility of moving this forward soon?

fjetter:
> Is there some possibility of moving this forward soon?

It's unlikely that I'll find time for this soon. If you are up for it, feel free to have a look

graingert:
might be better to use `@gen_test` instead of `@pytest.mark.asyncio` and drop the dep on pytest-asyncio

fjetter:
> might be better to use @gen_test instead of @pytest.mark.asyncio and drop the dep on pytest-asyncio

I'm open for this as well. In the past, usage of pytest-asyncio was useful because we could more easily use pytest fixtures. gen_test *does* support this by now so I don't see a strong reason for a continued usage of pytest-asyncio.

PR #4860, opened 2021-05-27
Fix WhoHas/HasWhat async usage

quasiben:
- [ ] Closes #xxxx
- [x] Tests added / passed

Fixes an issue first identified by @crusaderky https://github.com/dask/distributed/pull/4853#issuecomment-849709357

cc @jrbourbeau

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -53,7 +53,6 @@ from .core import (
 )
 from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
-from .objects import HasWhat, WhoHas
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
 from .publish import Datasets
@@ -3203,11 +3202,7 @@ class Client:
             keys = list(map(stringify, {f.key for f in futures}))
         else:
             keys = None
-        result = self.sync(self.scheduler.who_has, keys=keys, **kwargs)
-        if self.asynchronous:
-            return result
-        else:
-            return WhoHas(result)
+        return self.sync(self.scheduler.who_has, keys=keys, **kwargs)
 
     def has_what(self, workers=None, **kwargs):
         """Which keys are held by which workers
@@ -3241,11 +3236,7 @@ class Client:
             workers = list(workers)
         if workers is not None and not isinstance(workers, (tuple, list, set)):
             workers = [workers]
-        result = self.sync(self.scheduler.has_what, workers=workers, **kwargs)
-        if self.asynchronous:
-            return result
-        else:
-            return HasWhat(result)
+        return self.sync(self.scheduler.has_what, workers=workers, **kwargs)
 
     def processing(self, workers=None):
         """The tasks currently running on each worker
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -3241,12 +3241,11 @@ class Client:
             workers = list(workers)
         if workers is not None and not isinstance(workers, (tuple, list, set)):
             workers = [workers]
+        result = self.sync(self.scheduler.has_what, workers=workers, **kwargs)
         if self.asynchronous:
-            return self.sync(self.scheduler.has_what, workers=workers, **kwargs)
+            return result
         else:
-            return HasWhat(
-                self.sync(self.scheduler.has_what, workers=workers, **kwargs)
-            )
+            return HasWhat(result)
 
     def processing(self, workers=None):
         """The tasks currently running on each worker
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -53,6 +53,7 @@ from .core import (
 )
 from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
+from .objects import HasWhat, WhoHas
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
 from .publish import Datasets
@@ -3202,7 +3203,12 @@ class Client:
             keys = list(map(stringify, {f.key for f in futures}))
         else:
             keys = None
-        return self.sync(self.scheduler.who_has, keys=keys, **kwargs)
+
+        result = self.sync(self.scheduler.who_has, keys=keys, **kwargs)
+        if self.asynchronous:
+            return result
+
+        return WhoHas(result)
 
     def has_what(self, workers=None, **kwargs):
         """Which keys are held by which workers
@@ -3236,7 +3242,12 @@ class Client:
             workers = list(workers)
         if workers is not None and not isinstance(workers, (tuple, list, set)):
             workers = [workers]
-        return self.sync(self.scheduler.has_what, workers=workers, **kwargs)
+        result = self.sync(self.scheduler.has_what, workers=workers, **kwargs)
+
+        if self.asynchronous:
+            return result
+
+        return HasWhat(result)
 
     def processing(self, workers=None):
         """The tasks currently running on each worker
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -53,7 +53,6 @@ from .core import (
 )
 from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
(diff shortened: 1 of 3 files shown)


distributed/scheduler.py
```diff
@@ -1450,7 +1454,10 @@ def who_wants(self):
 
     @property
     def who_has(self):
-        return self._who_has
+        try:
```
quasiben, line 1457: As a small short term hack I added this try/except because we will sometime return WorkerState objects in the dictionary:

```
ipdb> WhoHas(self._who_has)
*** TypeError: cannot convert dictionary update sequence element #0 to a sequence
ipdb> pp self._who_has
{<WorkerState 'tcp://127.0.0.1:57682', name: 0, memory: 1, processing: 0>}
ipdb>
```

You can see this in the `distributed/tests/test_client.py::test_idempotence` test

quasiben:
There are a number of places inside of scheduler.py where we rely on dictionary methods which didn't play nicely.  For example:

> recipient.has_what.add(ts)

In b9c42fcec0f63e543aae325c901ed635c676dcce, I thought it best just to handle the return explicitly for async clients. So we could keep the fancy repr for ipython while we sort through a nice model for leveraging HTML in the repr

distributed/client.py
```diff
@@ -3203,7 +3203,10 @@ def who_has(self, futures=None, **kwargs):
             keys = list(map(stringify, {f.key for f in futures}))
         else:
             keys = None
-        return WhoHas(self.sync(self.scheduler.who_has, keys=keys, **kwargs))
+        if self.asynchronous:
+            return self.sync(self.scheduler.who_has, keys=keys, **kwargs)
+        else:
+            return WhoHas(self.sync(self.scheduler.who_has, keys=keys, **kwargs))
```
jrbourbeau, line 3209: Nitpick: To avoid repeated code we could do something like (note, I've not tested the diff below): 

```suggestion
        result = self.sync(self.scheduler.who_has, keys=keys, **kwargs)
        if self.asynchronous:
            return result
        else:
            return WhoHas(result)
```

distributed/client.py
```diff
@@ -3237,7 +3240,12 @@ def has_what(self, workers=None, **kwargs):
             workers = list(workers)
         if workers is not None and not isinstance(workers, (tuple, list, set)):
             workers = [workers]
-        return HasWhat(self.sync(self.scheduler.has_what, workers=workers, **kwargs))
+        if self.asynchronous:
+            return self.sync(self.scheduler.has_what, workers=workers, **kwargs)
+        else:
+            return HasWhat(
+                self.sync(self.scheduler.has_what, workers=workers, **kwargs)
+            )
```
jrbourbeau, line 3248: Similar comment here

jrbourbeau (commented):
Thanks for the quick fix @quasiben

jrbourbeau:
> There are a number of places inside of scheduler.py where we rely on dictionary methods which didn't play nicely

I think that's because `has_what` is a `set` on the scheduler 

https://github.com/dask/distributed/blob/2e01873444067fc04939fe2ea50f49a40b2eeb1d/distributed/scheduler.py#L482

distributed/tests/test_client.py
```diff
@@ -3613,6 +3613,19 @@ async def test_status():
     await s.close()
 
 
+@gen_cluster(client=False, timeout=None)
```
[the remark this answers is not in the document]
quasiben: Not needed -- I'll take it out

distributed/tests/test_client.py
```diff
@@ -3613,6 +3613,19 @@ async def test_status():
...
+async def test_async_whowhat(s, a, b):
+    c = await Client(s.address, asynchronous=True)
```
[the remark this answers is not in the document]
quasiben: 👍

jakirkham:
Should it be a `dict` or a `set`?

Or is it both? If the latter, maybe it should be `object`

It's worth noting when variables are too flexible it really makes it hard to Cythonize things efficiently

jrbourbeau:
Things like `TaskState._who_has` are `set`s

https://github.com/dask/distributed/blob/2e01873444067fc04939fe2ea50f49a40b2eeb1d/distributed/scheduler.py#L1292

However, client-side methods like `Client.who_has` result in some additional formatting code being called

https://github.com/dask/distributed/blob/2e01873444067fc04939fe2ea50f49a40b2eeb1d/distributed/scheduler.py#L6174-L6189

where the thing that is actually returned to the user is a `dict`. Perhaps we should have the `dict` returned by `Scheduler.get_who_has` be the place we add the `WhoHas` `dict` subclass? What do you think @jakirkham @quasiben?

quasiben:
> where the thing that is actually returned to the (client-side) user is a dict. Perhaps we should have the dict returned by Scheduler.get_who_has be the place we add the WhoHas dict subclass? What do you think @jakirkham @quasiben?

Admittedly, I didn't know the `get_who_has` method existed.  I'll test that out

jakirkham:
Appears to be working. @jrbourbeau do you have any more thoughts on this? 🙂

jrbourbeau:
Hmm, unfortunately it looks like the `HasWhat` / `WhoHas` subclasses aren't surviving the trip from the scheduler to the client. Instead we just get normal `dict`s at the client. For example, this change

```diff
diff --git a/distributed/tests/test_client.py b/distributed/tests/test_client.py
index 93aae7aa..c4b0c96f 100644
--- a/distributed/tests/test_client.py
+++ b/distributed/tests/test_client.py
@@ -58,6 +58,7 @@ from distributed.comm import CommClosedError
 from distributed.compatibility import MACOS, WINDOWS
 from distributed.core import Status
 from distributed.metrics import time
+from distributed.objects import HasWhat, WhoHas
 from distributed.scheduler import (
     COMPILED,
     CollectTaskMetaDataPlugin,
@@ -3620,6 +3621,9 @@ async def test_async_whowhat(c, s, a, b):
     who_has = await c.who_has()
     has_what = await c.has_what()

+    assert type(who_has) is WhoHas
+    assert type(has_what) is HasWhat
+
     assert who_has == {x.key: (a.address,)}
     assert has_what == {a.address: (x.key,), b.address: ()}
```

results in `test_async_whowhat` failing:

```python
    @gen_cluster(client=True)
    async def test_async_whowhat(c, s, a, b):
        [x] = await c.scatter([1], workers=a.address)

        who_has = await c.who_has()
        has_what = await c.has_what()

>       assert type(who_has) is WhoHas
E       AssertionError: assert <class 'dict'> is WhoHas
E        +  where <class 'dict'> = type({'int-c0a8a20f903a4915b94db8de3ea63195': ('tcp://127.0.0.1:59649',)})

distributed/tests/test_client.py:3624: AssertionError
```

I'm not sure why the `HasWhat` / `WhoHas` subclasses aren't making it to the client process right now. If we don't have time to dig in further, I think we can revert https://github.com/dask/distributed/pull/4860/commits/ac502283afd639e586c4da5d773afde29426ad12 and merge to fix things for the upcoming release.

quasiben:
reverted but also added a test based on your last comment

jrbourbeau (approved):
Thanks @quasiben, the changes here look good to me. Let's merge after CI finishes.

> I'm not sure why the HasWhat / WhoHas subclasses aren't making it to the client process right now

FWIW I think this has to do with `msgpack`'s handling of `dict` subclasses

```python
In [1]: from distributed.objects import HasWhat

In [2]: x = HasWhat()

In [3]: type(x)
Out[3]: distributed.objects.HasWhat

In [4]: from distributed.protocol.serialize import msgpack_dumps, msgpack_loads

In [5]: x_roundtrip = msgpack_loads(*msgpack_dumps(x))

In [6]: type(x_roundtrip)
Out[6]: dict
```

@jakirkham @madsbk may find the above snippet interesting

jakirkham:
Yeah was wondering if we needed to do something special with serialization (like defining how to serialize this object). That said, this seems like we are getting into the weeds right before a release. So probably best to punt and revisit after

quasiben:
CI finished green.  Thanks @jrbourbeau and @jakirkham

PR #4863, opened 2021-05-28
Drop usage of `WhoHas` & `WhatHas` from `Client`

jakirkham:
PR ( https://github.com/dask/distributed/pull/4853 ) ran into some issues, which was largely reverted by PR ( https://github.com/dask/distributed/pull/4860 ). However there are some lingering issues, this tries to revert the remaining pieces.

cc @dantegd @quasiben @jrbourbeau

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/client.py
+++ b/distributed/client.py
@@ -53,7 +53,6 @@ from .core import (
 )
 from .diagnostics.plugin import UploadFile, WorkerPlugin, _get_worker_plugin_name
 from .metrics import time
-from .objects import HasWhat, WhoHas
 from .protocol import to_serialize
 from .protocol.pickle import dumps, loads
 from .publish import Datasets
@@ -3203,12 +3202,7 @@ class Client:
             keys = list(map(stringify, {f.key for f in futures}))
         else:
             keys = None
-
-        result = self.sync(self.scheduler.who_has, keys=keys, **kwargs)
-        if self.asynchronous:
-            return result
-
-        return WhoHas(result)
+        return self.sync(self.scheduler.who_has, keys=keys, **kwargs)
 
     def has_what(self, workers=None, **kwargs):
         """Which keys are held by which workers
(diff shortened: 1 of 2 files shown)


jrbourbeau (commented):
I think we need to comment out / revert these checks too 

https://github.com/dask/distributed/blob/93e286987431381d995c93bbacea1ccb07da576b/distributed/tests/test_client.py#L3633-L3634

jakirkham:
Thanks marked that `xfail`. Guessing we will want to come back and straighten this out after the release. Though if we want to change more, happy to do so.

FWIW here's [a CI log]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci-v0.20/job/cuml/job/prb/job/cuml-gpu-test/CUDA=11.0,GPU_LABEL=gpu,OS=ubuntu16.04,PYTHON=3.7/281/console ) where we were seeing this issue, which traces back to [here in cuML]( https://github.com/rapidsai/cuml/blob/4d069915b9d12fa8b301d1c5e4bb4abe5198d1a4/python/cuml/dask/common/part_utils.py#L160 ):

```python
>       return WhoHas(result)
07:29:48 E       TypeError: 'coroutine' object is not iterable
```

Dante and I have been trying to reproduce with variations of this:

```python
from time import sleep

def inc(i):
    sleep(60)
    return i + 1

with Client(n_workers=1) as c:
    f = c.submit(inc, 0)
    d = c.who_has(f.key)
    print(d)
```

However we haven't quite narrowed down a reproducer. Guessing something like that would trigger the error. Though there is likely something simple we are overlooking.

Sorry this isn't more helpful. Hopefully that at least gives some context though 🙂

jrbourbeau:
Hmm that's interesting, it looks like `self.sync(self.scheduler.who_has, keys=keys, **kwargs)` is returning a coroutine even though `self.asynchronous` is `False`-y. I wonder if using `inspect.isawaitable` instead of `self.asynchronous` would be a more robust check

jakirkham:
Yeah it's odd. Also confused as to why we haven't been able to reproduce it. That makes sense. I think for now we are ok just reverting, but happy to try that after the release

jrbourbeau (commented):
The changes here look good for getting the release out today. @jakirkham could you remove the draft status of this PR when you're ready for it to be merged?

jakirkham:
Should be good to go. Sorry we were just testing on our end to make sure things were cleared up

jakirkham:
FWIW this [CI log]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci-v0.20/job/cuml/job/prb/job/cuml-gpu-test/CUDA=11.2,GPU_LABEL=gpu,OS=centos7,PYTHON=3.8/287/console ) shows the issue is fixed 🎉

PR #4866, opened 2021-06-01
Move SystemMonitor's GPU initialization back to constructor

pentschev:
- [ ] Closes https://github.com/rapidsai/dask-cuda/issues/634

### DIFF
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -1,3 +1,5 @@
+import os
+
 import pynvml
 
 nvmlInit = None
@@ -15,7 +17,18 @@ def init_once():
 
 
 def _pynvml_handles():
-    return pynvml.nvmlDeviceGetHandleByIndex(0)
+    count = pynvml.nvmlDeviceGetCount()
+    try:
+        cuda_visible_devices = [
+            int(idx) for idx in os.environ.get("CUDA_VISIBLE_DEVICES", "").split(",")
+        ]
+    except ValueError:
+        # CUDA_VISIBLE_DEVICES is not set
+        cuda_visible_devices = False
+    if not cuda_visible_devices:
+        cuda_visible_devices = list(range(count))
+    gpu_idx = cuda_visible_devices[0]
+    return pynvml.nvmlDeviceGetHandleByIndex(gpu_idx)
 
 
 def real_time():
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -1,5 +1,3 @@
-import os
-
 import pynvml
 
 nvmlInit = None
@@ -17,18 +15,7 @@ def init_once():
 
 
 def _pynvml_handles():
-    count = pynvml.nvmlDeviceGetCount()
-    try:
-        cuda_visible_devices = [
-            int(idx) for idx in os.environ.get("CUDA_VISIBLE_DEVICES", "").split(",")
-        ]
-    except ValueError:
-        # CUDA_VISIBLE_DEVICES is not set
-        cuda_visible_devices = False
-    if not cuda_visible_devices:
-        cuda_visible_devices = list(range(count))
-    gpu_idx = cuda_visible_devices[0]
-    return pynvml.nvmlDeviceGetHandleByIndex(gpu_idx)
+    return pynvml.nvmlDeviceGetHandleByIndex(0)
 
 
 def real_time():
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -14,13 +14,9 @@ def init_once():
     nvmlInit()
 
 
-def _pynvml_handles():
-    return pynvml.nvmlDeviceGetHandleByIndex(0)
-
-
 def real_time():
     init_once()
-    h = _pynvml_handles()
+    h = pynvml.nvmlDeviceGetHandleByIndex(0)
     return {
         "utilization": pynvml.nvmlDeviceGetUtilizationRates(h).gpu,
         "memory-used": pynvml.nvmlDeviceGetMemoryInfo(h).used,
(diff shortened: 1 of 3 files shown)


pentschev:
cc @charlesbluca @quasiben

distributed/diagnostics/nvml.py
```diff
@@ -17,18 +15,7 @@ def init_once():
...
-            int(idx) for idx in os.environ.get("CUDA_VISIBLE_DEVICES", "").split(",")
-        ]
-    except ValueError:
-        # CUDA_VISIBLE_DEVICES is not set
-        cuda_visible_devices = False
-    if not cuda_visible_devices:
-        cuda_visible_devices = list(range(count))
-    gpu_idx = cuda_visible_devices[0]
-    return pynvml.nvmlDeviceGetHandleByIndex(gpu_idx)
+    return pynvml.nvmlDeviceGetHandleByIndex(0)
```
charlesbluca, line 18: Nitpicky but at this point could we just call `pynvml.nvmlDeviceGetHandleByIndex(0)` in the places where we used to call `nvml._pynvml_handles()`?

distributed/system_monitor.py
```diff
@@ -92,10 +93,6 @@ def update(self):
 
         # give external modules (like dask-cuda) a chance to initialize CUDA context
         if nvml is not None and nvml.nvmlInit is not None:
```
charlesbluca, line 95: I _think_ the `nvml.nvmlInit` check here is redundant now, though it shouldn't cause any problems to leave it here.

charlesbluca (commented):
Thanks for doing this @pentschev 😄

distributed/system_monitor.py
pentschev: It's not redundant, this is what I mentioned earlier when I also thought it was. It refers to the object in https://github.com/dask/distributed/blob/9d4f0bf2fc804f955a869febd3b51423c4382908/distributed/diagnostics/nvml.py#L5 and not to the `pynvml.nvmlInit` method. I think that's a confusing naming choice nevertheless but I won't touch it right now.

distributed/diagnostics/nvml.py
pentschev: Done in https://github.com/dask/distributed/pull/4866/commits/ddf9a434154fad4802bc75915b61957fbbf13600

distributed/system_monitor.py
charlesbluca: I agree with that; I'm referring to the fact that by the time we call `update()`, we will have also called `nvml.one_time()`, meaning that `nvml.nvmlInit` will always not be `None` if `nvml` is not `None`.

pentschev: Ah sorry, you're right, good catch. I've updated that in https://github.com/dask/distributed/pull/4866/commits/79b315b660a1f045fbb7216d322a4b25a09bc801 .

jrbourbeau (commented):
Thanks @pentschev! Is there a regression test we should add here or in `dask-cuda`?

pentschev:
@jrbourbeau since it requires PyNVML and a GPU, unfortunately we can't test it in Distributed right now, but @charlesbluca is working on testing that in Dask-CUDA in https://github.com/rapidsai/dask-cuda/pull/635 .

quasiben:
I think there is an issue here still.  The changes around `pynvml.nvmlDeviceGetHandleByIndex` i _think_, need to be reverted.  In https://github.com/dask/distributed/pull/3810 we saw that pynvml doesn't respect CUDA_VISIBLE_DEVICES so we get in correct reporting. Here's a small reproducer of the issue:


```python



In [1]: from dask.distributed import Client, fire_and_forget, wait
   ...: from dask_cuda import LocalCUDACluster
   ...: from dask.utils import parse_bytes
   ...: import dask

In [2]: cluster = LocalCUDACluster()

In [4]: import rmm

In [5]: rmm.reinitialize(pool_allocator=1e9) # create data on the client/GPU 0

In [6]: for w in cluster.scheduler.workers:
    ...:     print(cluster.scheduler.workers[w].metrics['gpu_memory_used'])
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
17728536576
```

In the above, we should only see one GPU with a large allocation

jrbourbeau:
Thanks for pointing me to https://github.com/rapidsai/dask-cuda/pull/635 @pentschev -- that's what I was looking for. I knew `dask-cuda` ran some subset of the `distributed` tests, I just wanted to make sure something for this issue was included which it looks like https://github.com/rapidsai/dask-cuda/pull/635 is handling

pentschev:
> I think there is an issue here still. The changes around `pynvml.nvmlDeviceGetHandleByIndex` i _think_, need to be reverted.

You're right. I've reverted the changes now. However, this breaks https://github.com/rapidsai/dask-cuda/blob/81bbc6f85575826b13b3fb45894b54135514e668/dask_cuda/tests/test_dask_cuda_worker.py#L21-L59 , which is a test that ensures we can verify `CUDA_VISIBLE_DEVICES` behavior even on a single-GPU setup (e.g., gpuCI), but fails because we now try to address a GPU index beyond the existing ones. I'm still trying to think of a way to fix that.

pentschev:
Alright, this will break the test I mentioned above but there's not much we can do right now to prevent it without adding considerable complexity to Distributed or Dask-CUDA. I say we should merge this as is and then xfail those tests in Dask-CUDA for now, and I'll file an issue to figure out a solution later.

jrbourbeau:
cc @quasiben

quasiben:
Thanks @pentschev .  I'm good with merging in as well and I'll help (as best I can) with the failing dask-cuda test

pentschev:
I'm ok with that. The failed test doesn't seem to be related, so it's good to merge from my side.

quasiben:
Thanks again @pentschev !

pentschev:
Thanks everyone for reviews!

Issue #4619, opened 2021-03-22
Flaky `test_profile`


jrbourbeau:
Over in https://github.com/dask/distributed/pull/4615 we [observed](https://github.com/dask/distributed/pull/4615/checks?check_run_id=2167375104) `test_profile` had an intermittent failure 

<details>
<summary>Full traceback:</summary>

```
2021-03-22T16:47:54.5827441Z ================================== FAILURES ===================================
2021-03-22T16:47:54.5828133Z ________________________________ test_profile _________________________________
2021-03-22T16:47:54.5833735Z 
2021-03-22T16:47:54.5834308Z     def test_func():
2021-03-22T16:47:54.5834877Z         result = None
2021-03-22T16:47:54.5835825Z         workers = []
2021-03-22T16:47:54.5836525Z         with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:
2021-03-22T16:47:54.5837831Z     
2021-03-22T16:47:54.5844201Z             async def coro():
2021-03-22T16:47:54.5844957Z                 with dask.config.set(config):
2021-03-22T16:47:54.5845485Z                     s = False
2021-03-22T16:47:54.5846816Z                     for _ in range(60):
2021-03-22T16:47:54.5847460Z                         try:
2021-03-22T16:47:54.5851799Z                             s, ws = await start_cluster(
2021-03-22T16:47:54.5853714Z                                 nthreads,
2021-03-22T16:47:54.5854442Z                                 scheduler,
2021-03-22T16:47:54.5854958Z                                 loop,
2021-03-22T16:47:54.5855590Z                                 security=security,
2021-03-22T16:47:54.5863190Z                                 Worker=Worker,
2021-03-22T16:47:54.5864011Z                                 scheduler_kwargs=scheduler_kwargs,
2021-03-22T16:47:54.5864754Z                                 worker_kwargs=worker_kwargs,
2021-03-22T16:47:54.5865280Z                             )
2021-03-22T16:47:54.5866867Z                         except Exception as e:
2021-03-22T16:47:54.5874386Z                             logger.error(
2021-03-22T16:47:55.5832993Z                                 "Failed to start gen_cluster: "
2021-03-22T16:47:55.5835158Z                                 f"{e.__class__.__name__}: {e}; retrying",
2021-03-22T16:47:55.5835680Z                                 exc_info=True,
2021-03-22T16:47:55.5836123Z                             )
[...]

PR #4876, opened 2021-06-02
Mark `test_server_comms_mark_active_handlers` with `pytest.mark.asyncio`

jrbourbeau:
I noticed `distributed/tests/test_core.py::test_server_comms_mark_active_handlers` is currently being unintentionally skipped because it's an `async def` tests which isn't marked with `pytest.mark.asyncio`. This PR ensures the test is run


jrbourbeau:
Test failure is https://github.com/dask/distributed/issues/4619

PR #4878, opened 2021-06-03
Rename nbytes widgets

### LINKED ISSUE #4871: Rename "nbytes" chart to "worker memory"
The current chart in the dashboard that is named "nbytes" should probably be renamed to something else.  I suggest "worker memory", which I think is more easily identifiable

crusaderky:
Rename widgets:
- NBytes -> WorkersMemory
- NBytesHistogram -> WorkersMemoryHistogram
- NBytesCluster -> ClusterMemory

Rename URIs:
- /individual-nbytes -> /individual-workers-memory
- /individual-nbytes-cluster -> /individual-cluster-memory

Closes #4871


jacobtomlinson (commented):
Big +1 here, this is a good change.

I expect this will break the Jupyter Lab Extension downstream. There is a [default list of individual panels](https://github.com/dask/dask-labextension/blob/3e35dd277e503e13dfb0950379eab1400abcd729/src/dashboard.tsx#L443) which is used if the panel discovery fails.

mrocklin:
In particular I think that changing the button in JLab is what I'm most after here.  It always take me a while to find it because I'm always looking for "Worker Memory"

jacobtomlinson:
Yeah I totally agree @mrocklin. But I guess we need a related PR to the lab extension to change the default URLs.

jrbourbeau:
Gentle ping for @ian-r-rose

crusaderky:
ping @ian-r-rose

jrbourbeau:
xref https://github.com/dask/dask-labextension/issues/198

ian-r-rose:
Sorry to miss the earlier ping. It's a simple fix to update the names of the default charts, and I'd be happy to do that. But we might consider removing the defaults all-together. They used to be more useful, but today they are mostly used to show the greyed out buttons when no dashboard is connected. When the dashboard is live, it should be able to pick up whatever is provided by the `individual-plots.json` endpoint.

crusaderky:
I've reintroduced the old API endpoints; this should hopefully mean that that this PR can be merged without needing to wait for the matching change in dask-labextension?

jacobtomlinson (approved):
Awesome thanks @crusaderky, good idea.

I've made a comment on the dask-labextension issue to come back here and remove the backward compatibility again.

crusaderky:
@mrocklin @jrbourbeau ready for merge

ian-r-rose:
> I've reintroduced the old API endpoints; this should hopefully mean that that this PR can be merged without needing to wait for the matching change in dask-labextension?

Sorry, I suppose I was unclear in my comment above and the linked issue.  What I meant was: the labextension will already pick up name the name changes that you have made. By reintroducing the old name, it now shows up twice in the button listing: 
image

The *only* time that we'll see the old names is when there is not a cluster connected to the UI (which will be resolved with dask/dask-labextension#198).

I think we'll want to remove the duplicated individual plots from e7305ea

PR #4873, opened 2021-06-02
Ensure PyNVML works correctly when installed with no GPUs

pentschev:
- [ ] Closes #4868
- [ ] Depends on #4866

PR #4879, opened 2021-06-03
Add range_query tests to NVML test suite

charlesbluca:
Adds a test similar to [`test_get_worker_monitor_info()`](https://github.com/dask/distributed/blob/abed5f4b5be667c88661a1ce950bf85cfafa5e29/distributed/tests/test_scheduler.py#L2251-L2260) to `test_nvml.py` for use in Dask-CUDA's gpuCI tests; the idea here is that we can hopefully catch issues like rapidsai/dask-cuda#634 preemptively.

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/diagnostics/tests/test_nvml.py
+++ b/distributed/diagnostics/tests/test_nvml.py
@@ -57,7 +57,7 @@ async def test_gpu_metrics(s, a, b):
 
 
 @gen_cluster()
-async def test_gpu_monitoring(s, a, b):
+async def test_gpu_monitoring_recent(s, a, b):
     h = nvml._pynvml_handles()
     res = await s.get_worker_monitor_info(recent=True)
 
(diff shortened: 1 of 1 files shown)


pentschev:
We probably need https://github.com/dask/distributed/pull/4866 for this, but Distributed CI likely won't catch this because it doesn't have GPU support yet.

pentschev:
As discussed with @charlesbluca earlier, it's probably best to wait until https://github.com/dask/distributed/pull/4873 is merged and then add the following check to this test:

```python
if nvml.device_get_count() < 1:
    pytest.skip("No GPUs available")
```

pentschev:
@charlesbluca https://github.com/dask/distributed/pull/4873 is in now, you can now test with that.

charlesbluca:
Looks like that fixes things - thanks @pentschev! I'll merge upstream and add the check.

jrbourbeau (commented):
Thanks @charlesbluca! I noticed all six tests in this module require `nvml.device_get_count() >= 1`. In order to minimize code duplication could we just skip this entire module if `nvml.device_get_count() == 0`?

charlesbluca:
I believe so - I discussed this briefly with @pentschev but wasn't 100% sure if we needed to do a check per-test to ensure that the tests have proper NVML initialization.

jrbourbeau:
Fair point. FWIW my impression was that `device_get_count` handled initialization properly:

https://github.com/dask/distributed/blob/2bdec05aefa1df5332dc9b52f93a543a4d382434/distributed/diagnostics/nvml.py#L26-L31

But I'll let @pentschev confirm whether or not this is the case

pentschev:
I just didn't want to risk introducing another issue, but if it works with that change I'm fine with doing that.

jrbourbeau (approved):
Thanks @charlesbluca, this is in

PR #4882, opened 2021-06-04
No longer cancel result future in async process when using timeouts

fjetter:
Minor fix I came across. From what I can see the async join is actually never used except for this test. Since the test was marked xfail it would still run, generate weird warnings, run into timeouts, etc.


mrocklin:
Woot.

Issue #4839, opened 2021-05-21
test_metrics.py:test_process_time() failing assertions


stefanor:
@andrewshadura reports in [Debian Bug #987816](https://bugs.debian.org/987816) that he's getting some assertion failures in `test_process_time()` in his build environment:

```
    ______________________________ test_process_time _______________________________
    
        def test_process_time():
            start = metrics.process_time()
            run_for(0.05)
            dt = metrics.process_time() - start
    >       assert 0.03 <= dt <= 0.2
    E       assert 0.03 <= 0.02570413899999835
    
    distributed/tests/test_metrics.py:27: AssertionError
    _______________________________ test_thread_time _______________________________
    
        def test_thread_time():
            start = metrics.thread_time()
            run_for(0.05)
            dt = metrics.thread_time() - start
    >       assert 0.03 <= dt <= 0.2
    E       assert 0.03 <= 0.027913339999997788
    
    distributed/tests/test_metrics.py:48: AssertionError
    …
```

I haven't been able to reproduce it, but those numbers feel like they were pulled out of thin air in #1625. @pitrou?
Should we just reduce the 0.03 to 0.02?

**Environment**:

- Dask version: 2021.01.0
- Python version: 3.9.2
- Operating System: Debian
- Install method (conda, pip, source): Debian

fjetter:
Thanks for the report! We've seen a similar issue popping up in https://github.com/dask/distributed/issues/4790

PR #4883, opened 2021-06-04
Don't strip scheduler protocol when determining host

jrbourbeau:
When determining the scheduler `host` on a worker we currently strip off the protocol used by the scheduler:

https://github.com/dask/distributed/blob/611923e2d718350b8e8443d0d94c853c18b91574/distributed/worker.py#L533

Since [`distributed.addressing.get_address_host` internally calls the corresponding comm backend `get_address_host` method](https://github.com/dask/distributed/blob/611923e2d718350b8e8443d0d94c853c18b91574/distributed/comm/addressing.py#L140-L142), I would have expected that if the scheduler is using a custom protocol, we would want to use the custom comm backend `get_address_host` method when determining the scheduler's host instead of whatever the default is on the worker (`tcp` by default). 

It looks like we began stripping the protocol intentionally in https://github.com/dask/distributed/pull/3678, though I'm not sure what the reasoning behind the change was (xref https://github.com/dask/distributed/pull/3678#discussion_r404941210). This PR proposes we keep the scheduler address protocol so we can use that information in `get_address_host`, though I certainly could be missing some logic behind why we began stripping the protocol in the first place.

cc @mrocklin

### DIFF
--- a/distributed/comm/addressing.py
+++ b/distributed/comm/addressing.py
@@ -5,8 +5,6 @@ import dask
 from ..utils import get_ip_interface
 from . import registry
 
-DEFAULT_SCHEME = dask.config.get("distributed.comm.default-scheme")
-
 
 def parse_address(addr, strict=False):
     """
@@ -28,7 +26,7 @@ def parse_address(addr, strict=False):
         )
         raise ValueError(msg)
     if not sep:
-        scheme = DEFAULT_SCHEME
+        scheme = dask.config.get("distributed.comm.default-scheme")
     return scheme, loc
 
 
--- a/distributed/tests/test_worker.py
+++ b/distributed/tests/test_worker.py
@@ -29,6 +29,8 @@ from distributed import (
     get_worker,
     wait,
 )
+from distributed.comm.registry import backends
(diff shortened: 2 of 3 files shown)


mrocklin:
No recollection on this one.  If tests pass then I say go for it.

fjetter (commented):
Test failure is https://github.com/dask/distributed/issues/4839

Can we write a test for this such that we know for certain if this works or not? fwiw, using the protocol feels correct and I'd be fine without a test if this is too much effort.

jrbourbeau:
Thanks for reviewing @fjetter! Just included a test

I started writing a test to ensure that the worker uses the same comm backend as the scheduler when determining the host to connect to. However this involved more mocking and referring to implementation details than I'd usually prefer, so I backed off from that approach and wrote a more indirect test that makes sure the worker doesn't use the default backend (which is the current behavior on `main` that led to this PR). Let me know if you'd like to see any other updates

fjetter (approved):
Thanks for adding the test. I think the test failure is unrelated.

jrbourbeau:
Thanks for reviewing @fjetter!

PR #4888, opened 2021-06-08
Flake8 config cleanup

fjetter:
This shrinks our flake8 error ignore config to a more reasonable level. Most error codes are irrelevant since they concern formatting which is handled by black, nowadays. However, there are two error codes in particular which triggered me doing this.

## F841: local variable is assigned to but never used

This is an error code which detects assigned variables if they are unused. While this is _usually_ not a big deal for tests (although not desired), this can be a frequent source for bugs in the actual code.

An example in this PR are changes in distributed/profile.py which likely indicates that we lost _something_ at some point in time or distributed/scheduler.py l 3820ff which is definitely a bug (not fixed, yet)

We can still ignore this for tests, which is now reflected in the setup.cfg

## F811 Redefinition of unused name from line n

This warning is raised if we are redefining symbols, e.g. functions or global variables, etc. Mostly, this is harmless. However, for test suites in particular this causes the problem of allowing duplicate test function names. I found 5-10 tests which were duplicated and therefore the first definition was ignored. These are typically cases where the definition of the two functions is hundreds or thousands of lines apart and it is impossible for reviewers to detect something like this. All instances where I found something like this, I simply appended a `_2` to the function name or similar.

Enabling this error code in particular caused most of the change in this PR since every fixture import we were doing would trigger this, e.g. the following would raise a warning in line 3 because loop was redefined in the local scope although it was already defined in global scope but remains unused.

The _easy_ solution to this is to use pytests fixture collection system via `conftest.py` which is why I added one import there to make all fixtures available to all functions. pytest will figure out the rest.

```python
from distributed.utils_test import loop

def test_loop(loop):
    pass
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -16,6 +16,9 @@ else:
     except Exception:
         pass
 
+# Make all fixtures available
+from distributed.utils_test import *  # noqa
+
 
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true", help="run slow tests")
--- a/distributed/cli/tests/test_dask_scheduler.py
+++ b/distributed/cli/tests/test_dask_scheduler.py
@@ -16,7 +16,6 @@ import distributed.cli.dask_scheduler
 from distributed import Client, Scheduler
 from distributed.metrics import time
 from distributed.utils import get_ip, get_ip_interface, tmpfile
-from distributed.utils_test import loop  # noqa: F401
 from distributed.utils_test import (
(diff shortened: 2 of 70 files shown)


mrocklin:
+1

fjetter:
Test failure is an unrelated timeout in `test_AllProgress_lost_key`

jrbourbeau (approved):
Thanks @fjetter!

PR #4893, opened 2021-06-08
Add configuration to enable/disable NVML diagnostics

### LINKED ISSUE #4885: Simple local cluster on laptop fails with deactivated GPU
**What happened**:
Thinkpad T470p with dual Intel HD Graphics 630 / Nvidia GeForce 940MX GPU ('optimus' configuration). I use [bumblebee](https://wiki.debian.org/Bumblebee) to switch on/off the Nvidia GPU only when I need it for CUDA computational work to save power. The Nvidia is still visible but inactive when switched off this way.

When I start below code snippet with a deactivated Nvidia, it crashes:

```
$ sudo echo OFF > /proc/acpi/bbswitch
$ ./test_distributed.py 
  File "./test_distributed.py", line 6, in <module>
    client = dask.distributed.Client()
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/client.py", line 754, in __init__
    self.start(timeout=timeout)
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/client.py", line 967, in start
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/utils.py", line 354, in sync
    raise exc.with_traceback(tb)
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/utils.py", line 337, in f
    result[0] = yield future
  File "/home/sbauer/.local/lib/python3.7/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/client.py", line 1035, in _start
    **self._startup_kwargs,
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/deploy/spec.py", line 401, in _
    await self._start()
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/deploy/spec.py", line 311, in _start
    self.scheduler = cls(**self.scheduler_spec.get("options", {}))
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/scheduler.py", line 3382, in __init__
    import distributed.dashboard.scheduler
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/dashboard/scheduler.py", line 11, in <module>
    from .components.nvml import gpu_doc  # noqa: 1708
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/dashboard/components/nvml.py", line 22, in <module>
    NVML_ENABLED = nvml.device_get_count() > 0
  File "/home/sbauer/.local/lib/python3.7/site-packages/distributed/diagnostics/nvml.py", line 31, in device_get_count
    return pynvml.nvmlDeviceGetCount()
  File "/home/sbauer/.local/lib/python3.7/site-packages/pynvml/nvml.py", line 1568, in nvmlDeviceGetCount
    _nvmlCheckReturn(ret)
  File "/home/sbauer/.local/lib/python3.7/site-packages/pynvml/nvml.py", line 743, in _nvmlCheckReturn
    raise NVMLError(ret)
pynvml.nvml.NVMLError_Uninitialized: Uninitialized
```

When I activate the Nvidia GPU, all is fine:
```
$ sudo echo ON > /proc/acpi/bbswitch
$ ./test_distributed.py 
<Client: 'tcp://127.0.0.1:42147' processes=4 threads=8, memory=15.55 GiB>
```

Even specifically setting CUDA_VISIBLE_DEVICES to empty doesn't work (I use this env in Tensorflow to explicitly ignore the GPU):
```
$ CUDA_VISIBLE_DEVICES='' ./test_distributed.py
  File "./test_distributed.py", line 6, in <module>
...
```

Local cluster starts regardless of active or deactive GPU.

It would help if there were some possibility to tell *distributed* to ignore any present GPU, but I didn't find anything in the documentation.

```python
#!/usr/bin/env python3

import dask.distributed

if __name__ == "__main__":
    client = dask.distributed.Client()
    print(client)
```

**Environment**:

- Dask version: dask 2021.6.0 / distributed 2021.6.0
- Python version: 3.7.3
- Operating System: Linux 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux

pentschev:
- [ ] Closes #4885

In some cases, users may want or need to explicitly disable NVML diagnostics, this PR adds support for that via Dask config.

### DIFF
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -7,17 +7,15 @@ try:
 except ImportError:
     pynvml = None
 
-nvmlEnabled = True
 nvmlInitialized = False
 nvmlLibraryNotFound = False
 nvmlOwnerPID = None
 
 
 def init_once():
-    global nvmlEnabled, nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
+    global nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
 
-    nvmlEnabled = dask.config.get("distributed.diagnostics.nvml")
-    if nvmlEnabled is False:
+    if dask.config.get("distributed.diagnostics.nvml") is False:
         nvmlInitialized = False
         return
 
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -15,11 +15,13 @@ nvmlOwnerPID = None
 
 def init_once():
     global nvmlEnabled, nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
-    if pynvml is None or (nvmlInitialized is True and nvmlOwnerPID == os.getpid()):
-        return
 
     nvmlEnabled = dask.config.get("distributed.diagnostics.nvml")
     if nvmlEnabled is False:
+        nvmlInitialized = False
+        return
+
+    if pynvml is None or (nvmlInitialized is True and nvmlOwnerPID == os.getpid()):
         return
 
     nvmlInitialized = True
--- a/distributed/diagnostics/nvml.py
+++ b/distributed/diagnostics/nvml.py
@@ -1,18 +1,25 @@
 import os
 
+import dask
+
 try:
     import pynvml
 except ImportError:
     pynvml = None
 
+nvmlEnabled = True
 nvmlInitialized = False
 nvmlLibraryNotFound = False
 nvmlOwnerPID = None
 
 
 def init_once():
-    global nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
+    global nvmlEnabled, nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
     if pynvml is None or (nvmlInitialized is True and nvmlOwnerPID == os.getpid()):
+        return
+
+    nvmlEnabled = dask.config.get("distributed.diagnostics.nvml")
+    if nvmlEnabled is False:
         return
 
     nvmlInitialized = True
--- a/distributed/distributed.yaml
+++ b/distributed/distributed.yaml
@@ -195,6 +195,8 @@ distributed:
         key: null
         cert: null
 
+  diagnostics:
+    nvml: True
 
   ###################
   # Bokeh dashboard #
--- a/distributed/diagnostics/tests/test_nvml.py
+++ b/distributed/diagnostics/tests/test_nvml.py
@@ -31,12 +31,10 @@ def test_enable_disable_nvml():
 
     with dask.config.set({"distributed.diagnostics.nvml": False}):
         nvml.init_once()
-        assert nvml.nvmlEnabled is False
         assert nvml.nvmlInitialized is False
 
     with dask.config.set({"distributed.diagnostics.nvml": True}):
         nvml.init_once()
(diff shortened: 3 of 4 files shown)


pentschev:
cc @quasiben @jrbourbeau @SteffenBauer

pentschev:
Notice that I introduced a new section `diagnostics` for lack of a better place to add `nvml`, but feel free to suggest a more suitable section or name to replace `diagnostics`.

quasiben:
What do you think about placing this under [Dashboard section](https://docs.dask.org/en/latest/configuration-reference.html#dashboard) instead of creating a new section ?

cc @jacobtomlinson

pentschev:
> What do you think about placing this under [Dashboard section](https://docs.dask.org/en/latest/configuration-reference.html#dashboard) instead of creating a new section ?

I thought about that, but I reach the conclusion it could be more confusing, since the NVML module is under `distributed/diagnostics` and there's another directory `distributed/dashboard` where the `distributed.dashboard` configs apply.

quasiben:
That seems reasonable.  I'm @jrbourbeau how do you feel about adding a new config section ?

distributed/diagnostics/nvml.py
```diff
@@ -1,17 +1,26 @@
...
 nvmlInitialized = False
 nvmlLibraryNotFound = False
 nvmlOwnerPID = None
 
 
 def init_once():
-    global nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
+    global nvmlEnabled, nvmlInitialized, nvmlLibraryNotFound, nvmlOwnerPID
+
+    nvmlEnabled = dask.config.get("distributed.diagnostics.nvml")
```
jrbourbeau, line 19: IIUC this `nvmlEnabled` variable appears to be equivalent to the the `distributed.diagnostics.nvml` config value. If that's the case, I'd prefer to not introduce a new variable and instead just use the config value

distributed/distributed.yaml
```diff
@@ -195,6 +195,8 @@ distributed:
         key: null
         cert: null
 
+  diagnostics:
+    nvml: True
```
jrbourbeau, line 199: Since this is GPU specific, we might want to put this option in it's own section like `rmm` and `ucx`

https://github.com/dask/distributed/blob/0c154324bdfabfc28b1a773d878c759ad35252cc/distributed/distributed.yaml#L228-L237

distributed/diagnostics/nvml.py
pentschev: This is a good point, addressed in https://github.com/dask/distributed/pull/4893/commits/fed9e9770c89b0b0909130cb607942cef3db0d60 .

distributed/distributed.yaml
pentschev: I don't mind moving it there if you prefer, but it seems like the general feel about those is that they have been a mistake to be outside of the `distributed` schema. For example https://github.com/dask/distributed/issues/4904 is attempting to move `ucx` into `distributed` to comply with the general config schema.

jacobtomlinson:
I think putting this under diagnostics makes sense.

jrbourbeau (approved):
Thanks @pentschev!

Issue #4814, opened 2021-05-13
Attempting to launch with SSHCluster for gpu workers doesn't appear to work


drobison00:
**What happened**:
Attempting to launch a local cluster with SSHCluster fails.

*Note*:
Setting worker_module to `dask_cuda.CUDAWorker` or `dask_cuda.dask_cuda_worker`, as indicated in the SSHCluster docs results in a 'no module named ...' error.

**What you expected to happen**:
A functioning cluster to come up.

**Minimal Complete Verifiable Example**:

```python
from dask.distributed import SSHCluster

# May need to put id_rsa.pub in ~/.ssh/authorized_keys
ssh_cluster = SSHCluster(['localhost', 'localhost'],
                        connect_options={"known_hosts": None, 
                                         "client_keys": "~/.ssh/id_rsa.pub"},
                        scheduler_options={"port": 8786, "dashboard_address": 8787},
                        worker_module='dask_cuda.cuda_worker')
client = Client(ssh_cluster)

distributed.deploy.ssh - INFO - distributed.scheduler - INFO - -----------------------------------------------
distributed.deploy.ssh - INFO - distributed.scheduler - INFO - -----------------------------------------------
distributed.deploy.ssh - INFO - distributed.scheduler - INFO - Clear task state
distributed.deploy.ssh - INFO - distributed.scheduler - INFO -   Scheduler at:   tcp://10.10.100.33:8786
distributed.deploy.ssh - INFO - /home/drobison/anaconda3/envs/rapids-0.19/lib/python3.8/runpy.py:127: RuntimeWarning: 'dask_cuda.cuda_worker' found in sys.modules after import of package 'dask_cuda', but prior to execution of 'dask_cuda.cuda_worker'; this may result in unpredictable behaviour
distributed.deploy.ssh - INFO - warn(RuntimeWarning(msg))
Task exception was never retrieved
future: <Task finished name='Task-12' coro=<_wrap_awaitable() done, defined at /home/drobison/anaconda3/envs/rapids-0.19/lib/python3.8/asyncio/tasks.py:688> exception=Exception('Worker failed to start')>
Traceback (most recent call last):
  File "/home/drobison/anaconda3/envs/rapids-0.19/lib/python3.8/asyncio/tasks.py", line 695, in _wrap_awaitable
    return (yield from awaitable.__await__())
  File "/home/drobison/anaconda3/envs/rapids-0.19/lib/python3.8/site-packages/distributed/deploy/spec.py", line 71, in _
[...]

PR #4894, opened 2021-06-08
Update `dask_cuda` usage in `SSHCluster` docstring

jrbourbeau:
cc @quasiben @pentschev

Closes https://github.com/dask/distributed/issues/4814

### DIFF
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -302,7 +302,7 @@ def SSHCluster(
     ...     ["localhost", "hostwithgpus", "anothergpuhost"],
     ...     connect_options={"known_hosts": None},
     ...     scheduler_options={"port": 0, "dashboard_address": ":8797"},
-    ...     worker_module="dask_cuda.CUDAWorker")
+    ...     worker_module="dask_cuda.cli.dask_cuda_worker")
     >>> client = Client(cluster)
 
     See Also
--- a/distributed/deploy/ssh.py
+++ b/distributed/deploy/ssh.py
@@ -295,7 +295,7 @@ def SSHCluster(
     >>> client = Client(cluster)
 
     An example using a different worker module, in particular the
-    ``CUDAWorker`` worker from the ``dask-cuda`` project.
+    ``dask-cuda-worker`` command from the ``dask-cuda`` project.
 
     >>> from dask.distributed import Client, SSHCluster
     >>> cluster = SSHCluster(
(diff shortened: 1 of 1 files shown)


distributed/deploy/ssh.py
```diff
@@ -295,14 +295,14 @@ def SSHCluster(
...
-    ``dask-cuda-worker`` command from the ``dask-cuda`` project.
+    ``CUDAWorker`` worker from the ``dask-cuda`` project.
 
     >>> from dask.distributed import Client, SSHCluster
     >>> cluster = SSHCluster(
     ...     ["localhost", "hostwithgpus", "anothergpuhost"],
     ...     connect_options={"known_hosts": None},
     ...     scheduler_options={"port": 0, "dashboard_address": ":8797"},
-    ...     worker_module='dask_cuda.dask_cuda_worker')
+    ...     worker_module="dask_cuda.CUDAWorker")
```
pentschev, line 305:
```suggestion
    ...     worker_module="dask_cuda.cli.dask_cuda_worker")
```

pentschev: `dask_cuda.CUDAWorker` is the class that creates the `Nanny` object. I think what's expected here is the CLI interface, which is what I suggested above.

distributed/deploy/ssh.py
```diff
@@ -295,14 +295,14 @@ def SSHCluster(
     >>> client = Client(cluster)
 
     An example using a different worker module, in particular the
-    ``dask-cuda-worker`` command from the ``dask-cuda`` project.
+    ``CUDAWorker`` worker from the ``dask-cuda`` project.
```
pentschev, line 298: We could revert this change as per my suggestion below.

pentschev (changes requested):
Thanks @jrbourbeau for doing this, I added a few comments.

distributed/deploy/ssh.py
jrbourbeau: Thanks @pentschev -- I thought this was a `worker_class=` keyword which is more commonly used than `worker_module=`. Updating now

pentschev: Yeah, it can feel a bit counterintuitive as in most cases we're working with Python without a CLI, but to launch over SSH this takes a pretty straightforward path of just launching the command remotely.

pentschev (approved):
LGTM, thanks @jrbourbeau !

jrbourbeau:
Thanks for reviewing @pentschev!

Issue #4806, opened 2021-05-11
[CI] Warnings about used ports


fjetter:
Lately we are observing hundreds of warnings about the dashboard ports being occupied. While I understand that we occasionally may have an overlap, I would expect every test to close the HTTP server properly such that these overlaps happen only infrequently.
I can't help but wonder if this connects to our recent CI instabilities and/or CI performance.

Example https://github.com/dask/distributed/pull/4805/checks?check_run_id=2556648953

<details>

```
distributed/tests/test_multi_locks.py::test_multiple_locks
  /home/runner/work/distributed/distributed/distributed/node.py:160: UserWarning: Port 8787 is already in use.
  Perhaps you already have a cluster running?
  Hosting the HTTP server on port 46611 instead
    warnings.warn(

distributed/tests/test_multi_locks.py::test_num_locks
  /home/runner/work/distributed/distributed/distributed/node.py:160: UserWarning: Port 8787 is already in use.
  Perhaps you already have a cluster running?
  Hosting the HTTP server on port 46079 instead
    warnings.warn(

distributed/tests/test_nanny.py::test_lifetime
  /home/runner/work/distributed/distributed/distributed/node.py:160: UserWarning: Port 8787 is already in use.
  Perhaps you already have a cluster running?
  Hosting the HTTP server on port 37239 instead
    warnings.warn(

distributed/tests/test_nanny.py::test_nanny_closes_cleanly

```

</details>

jrbourbeau:
Thanks for raising an issue about this @fjetter -- I agree it seems like there are a lot more of these warnings lately. Perhaps there's a new test which leaves a server running and this impacts all the subsequent tests? Regardless of why, someone going through and resolving these warnings would be very welcome

martindurant:
The top of the warnings:

```distributed/comm/tests/test_ws.py::test_http_and_comm_server[True-wss://-security1-8787]
distributed/comm/tests/test_ws.py::test_http_and_comm_server[False-wss://-security3-8787]
  /home/runner/work/distributed/distributed/distributed/comm/ws.py:318: RuntimeWarning: Dashboard and Scheduler are using the same server on port 8787, defaulting to the Scheduler's ssl context. Your dashboard could become inaccessible
    warnings.warn(

distributed/dashboard/tests/test_scheduler_bokeh.py::test_NBytesHistogram
[...]

PR #4896, opened 2021-06-09
Dont start http server in Scheduler.__init__

fjetter:
This refactors the scheduler init such that the HTTP server is only started once the server is actually up. Otherwise every init call starts an http server and binds sockets, etc.

I tried testing for leaking sockets but that's not easy. Depending on OS this requires root and even with root, I had trouble asserting this properly using psutil.

Closes https://github.com/dask/distributed/issues/4806

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -86,6 +86,13 @@ from .utils_perf import disable_gc_diagnosis, enable_gc_diagnosis
 from .variable import VariableExtension
 
 try:
+    import distributed.dashboard.scheduler
+
+    HAS_BOKEH = True
+except ImportError:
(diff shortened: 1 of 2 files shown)


fjetter:
This needs much more work. Ran the wrong tests locally :/

mrocklin:
I was curious what was going to happen there.  I remember having to put the
HTTPServer in init for some reason.  I was happy to see your initial PR,
but not surprised to see it closed ;(

fjetter:
One reason why it is in init is that we can then align security and ensure that if either scheduler comm or http server is secured, the other one must be as well. At least this is one not-trivial thing I found locally. Why the CI deadlocks immediately is beyond my understanding atm

PR #4899, opened 2021-06-09
WIP co-assign related root-ish tasks

### LINKED ISSUE #4892: Co-assign neighboring tasks to neighboring workers
**Edit:** This proposal is now defunct(ish).  Read below for updated proposal.

Sometimes we assign tasks that are closely related to different machines, resulting in unnecessary communication.  The most simple case is when we have a couple of siblings

```
   c
  / \
 /   \
a     b
```

Because neither `a` nor `b` have dependencies, they are assigned somewhat randomly onto workers.  However because `c` requires them both one or the other will have to move across the network.  This can cause excess communication which can bog down the system.  

This has come up several times, notably in https://github.com/dask/distributed/issues/2602 and https://github.com/ocean-transport/coiled_collaboration/issues/3 and https://github.com/dask/distributed/pull/4864 .  There are a few different things that we can do to solve these issues.  This issue contains one part of a solution, assigning sibling tasks to the same worker.  

### Straightforward (but slow) solution

@fjetter provided a solution here https://github.com/dask/distributed/pull/4864#issuecomment-855228497

which involves iterating up through all dependents of a task, and then back down through all of those tasks dependencies, in order to find similar machines.  This doesn't work in the general case (we have to navigate through k-squared links, where k is the number of dependencies/dependents) but it might be ok in a restricted case. 

### Restricted case

In particular we seem to be interested in the following conditions:

1.  The task has no dependencies (otherwise we would schedule wherever those are)
2.  The task has very few dependents, possibly only one
3.  That dependent has very few dependencies, maybe less than five? (ten?)
4.  There are many more tasks than workers, so that we're comfortable losing some concurrency here
5.  The task we're thinking about is sizeable, and not something that we would want to move, relative to it's computation time

I suspect that there is an order by which we could check these conditions that is fast in all of the cases where it matters.  Additionally, if we have several dependencies to check then we might consider just accepting the location of the very first dependency that is already assigned, and skip checking the rest.

gjoseph92:
In `decide_worker`, rather than spreading out root tasks as much as possible, schedule consecutive (by priority order) root(ish) tasks on the same worker. This ensures the dependencies of a reduction start out on the same worker, reducing future data transfer.

- [x] Closes #4892
- [ ] Tests added / passed

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -952,6 +952,7 @@ class TaskGroup:
     _all_durations: object
     _last_worker: WorkerState
     _last_worker_tasks_left: int  # TODO Py_ssize_t?
+    _last_worker_priority: tuple  # TODO remove (debugging only)
 
     def __init__(self, name: str):
         self._name = name
@@ -968,6 +969,7 @@ class TaskGroup:
         self._all_durations = defaultdict(float)
         self._last_worker = None
         self._last_worker_tasks_left = 0
+        self._last_worker_priority = ()
 
     @property
     def name(self):
@@ -1024,6 +1026,14 @@ class TaskGroup:
     @last_worker_tasks_left.setter
     def last_worker_tasks_left(self, n: int):
         self._last_worker_tasks_left = n
+
+    @property
+    def last_worker_priority(self):
+        return self._last_worker_priority
+
+    @last_worker_priority.setter
+    def last_worker_priority(self, x: tuple):
+        self._last_worker_priority = x
 
     @ccall
     def add(self, o):
@@ -2393,6 +2403,7 @@ class SchedulerState:
             ts._group._last_worker_tasks_left = math.floor(
                 len(ts._group) / self._total_nthreads
             )
+            ts._group._last_worker_priority = ts._priority
 
         if self._validate:
             assert ws is None or isinstance(ws, WorkerState), (
@@ -7568,11 +7579,21 @@ def decide_worker(
             )
         ):
             group._last_worker_tasks_left -= 1
+            if group._last_worker_priority >= ts.priority:
+                print(
+                    f"decide_worker called out of priority order: {group._last_worker_priority} >= {ts.priority}.\n"
+                    f"{ts=}\n"
+                    f"{group.last_worker=}\n"
+                    f"{group.last_worker_tasks_left=}\n"
+                    f"{group_tasks_per_worker=}\n"
+                )
+            group._last_worker_priority = ts.priority
             return ws
 
         ws = min(candidates, key=objective)
         group._last_worker = ws
         group._last_worker_tasks_left = math.floor(group_tasks_per_worker)
+        group._last_worker_priority = ts.priority
 
     return ws
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2401,8 +2401,9 @@ class SchedulerState:
             # TODO repeated logic from `decide_worker`
             print(f"nodeps / no last worker fastpah - {ts.group_key} -> {ws.name}")
             ts._group._last_worker = ws
+            group_tasks_per_thread = len(ts._group) / self._total_nthreads
             ts._group._last_worker_tasks_left = (
-                math.floor(len(ts._group) / self._total_nthreads) - 1
+                math.floor(group_tasks_per_thread * ws._nthreads) - 1
             )
             ts._group._last_worker_priority = ts._priority
 
@@ -7577,7 +7578,7 @@ def decide_worker(
     if valid_workers is not None:
         total_nthreads = sum(wws._nthreads for wws in valid_workers)
 
-    group_tasks_per_worker = len(group) / total_nthreads
+    group_tasks_per_thread = len(group) / total_nthreads
     ignore_deps_while_picking: bool = False
 
     # Try to schedule sibling root-like tasks on the same workers, so subsequent reduction tasks
@@ -7588,7 +7589,7 @@ def decide_worker(
         # `decide_worker` hasn't previously been called out of priority order
         and group._last_worker_priority is not None
         # group is larger than cluster
-        and group_tasks_per_worker > 1
+        and group_tasks_per_thread > 1
         # is a root-like task (task group is large, but depends on very few tasks)
         and sum(map(len, group._dependencies)) < 5  # TODO what number
     ):
@@ -7667,7 +7668,9 @@ def decide_worker(
 
     if group._last_worker_priority is not None:
         group._last_worker = ws
-        group._last_worker_tasks_left = math.floor(group_tasks_per_worker) - 1
+        group._last_worker_tasks_left = (
+            math.floor(group_tasks_per_thread * ws._nthreads) - 1
+        )
         group._last_worker_priority = ts.priority
     return ws
 
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -2388,11 +2388,10 @@ class SchedulerState:
                 ws = wp_vals[self._n_tasks % n_workers]
 
             # TODO repeated logic from `decide_worker`
-            print("fastpath")
             ts._group._last_worker = ws
-            # ts._group._last_worker_tasks_left = math.floor(
-            #     len(ts._group) / self._total_nthreads
-            # )
+            ts._group._last_worker_tasks_left = math.floor(
+                len(ts._group) / self._total_nthreads
+            )
 
         if self._validate:
             assert ws is None or isinstance(ws, WorkerState), (
@@ -7546,9 +7545,7 @@ def decide_worker(
             break
     else:
         group: TaskGroup = ts._group
-        old: WorkerState = group._last_worker
-
-        ws = min(candidates, key=objective)
+        ws = group._last_worker
 
         total_nthreads = sum(
             wws._nthreads for wws in candidates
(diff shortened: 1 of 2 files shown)


distributed/scheduler.py
```diff
@@ -7477,7 +7504,33 @@ def decide_worker(
...
+        if (
+            ws is not None  # there is a previous worker
+            and group._last_worker_tasks_left > 0  # previous worker not fully assigned
+            and ts._dependents  # task has dependents
+            and group_tasks_per_worker > 1  # group is larger than cluster
+            and (  # is a root-like task (task group depends on very few tasks)
+                sum(map(len, group._dependencies)) < 5  # TODO what number
+            )
+        ):
+            group._last_worker_tasks_left -= 1
```
mrocklin, line 7527: In general I think that having a counter like this opens us up to failures in some complex situations. 

For example, what happens when we lose a few workers and need to recompute these tasks and this count goes negative?

I would encourage us to keep thinking about comparing occupancy between the best and recent workers.  That feels like a strictly local decision that might be more robust.

mrocklin: If the problem is that duration_average is unknown that's fine, we should replace it with whatever is used to increment the worker's occupancy when the previous tasks were added.

distributed/scheduler.py
```diff
@@ -7477,7 +7504,33 @@ def decide_worker(
         for ws in candidates:
             break
     else:
+        group: TaskGroup = ts._group
+        ws = group._last_worker
+
+        total_nthreads = sum(
+            wws._nthreads for wws in candidates
+        )  # TODO get `self._total_threads` from scheduler? Though that doesn't account for worker restrictions.
```
mrocklin, line 7512: Common case, if `len(candidates) < len(self.workers)` ... else `self._total_threads`

distributed/scheduler.py
```diff
@@ -7477,7 +7504,33 @@ def decide_worker(
...
+
+        group_tasks_per_worker = len(group) / total_nthreads
+
+        # Try to schedule sibling root-like tasks on the same workers, so subsequent reduction tasks
+        # don't require data transfer. Assumes `decide_worker` is being called in priority order.
+        if (
+            ws is not None  # there is a previous worker
+            and group._last_worker_tasks_left > 0  # previous worker not fully assigned
+            and ts._dependents  # task has dependents
```
mrocklin, line 7521: I recommend skipping the dependents check.  This can change in the future.  Consider the following case:

```python
df = dd.read_parquet(...).persist()

df.x.mean()
```

distributed/scheduler.py
gjoseph92: Yeah, this will probably get thrown off by both new workers arriving and workers leaving. I'll think more about how to use something less stateful.

distributed/scheduler.py
gjoseph92: That's cool (and the first non-performance-related advantage I see to this approach over finding relatives in the graph structure). So neighboring partitions of `df` would be stored on the same workers, before we even know what we're going to do with them.

distributed/scheduler.py
gjoseph92: Yeah. More that this function isn't a method of `SchedulerState`, so we'll have to pass in `self._total_threads` to here (fine, just slightly awkward).

distributed/scheduler.py
```diff
@@ -7477,7 +7509,31 @@ def decide_worker(
...
+
+        # Try to schedule sibling root-like tasks on the same workers, so subsequent reduction tasks
+        # don't require data transfer. Assumes `decide_worker` is being called in priority order.
+        if (
+            ws is not None  # there is a previous worker
+            and group._last_worker_tasks_left > 0  # previous worker not fully assigned
+            and group_tasks_per_worker > 1  # group is larger than cluster
+            and (  # is a root-like task (task group depends on very few tasks)
+                sum(map(len, group._dependencies)) < 5  # TODO what number
+            )
```
fjetter, line 7528: Does this mean that only groups of size `N (=5)` are grouped together and everything above is distributed _randomly_?

If my understanding here is correct, we might want to choose 32 (+/-1?) since this is the default branching factor for tree reductions in dask, last time I checked

mrocklin: No, this is restricing this behavior to task groups where the size of the task group is large, but the number of dependencies of that group is small.  This is true in situations like the following:

```python
df = dd.read_parquet(...)  # thousands of tasks, but zero dependenies

x = da.from_zarr(...)  # thousands of tasks, but one dependency (the zarr file)
```

But not in situations like the following:

```python
y = x + 1  # thousands of tasks, but thousands of dependencies
```

gjoseph92:
@mrocklin here's the xarray/zarr issue I'm running into.

Running this code on my branch behaves the same as on `main`. Note that this is the exact same array-sum test as #4864, just on a zarr opened in xarray. I haven't tried taking xarray out of the equation yet and using plain zarr. This is exactly the sort of case we hoped to make an exception for as you described in https://github.com/dask/distributed/pull/4899#discussion_r649950148.
[mydask.pdf](https://github.com/dask/distributed/files/6640185/mydask.pdf).

What I'm struggling to debug is that (from adding print statements to `decide_worker`; see latest commit), `decide_worker` doesn't even seem to be _called_ on the `sum-` root tasks, so we have no chance to smartly schedule them. I don't understand how this can happen.

```python
# test_zarr.py
import xarray as xr
import dask.array as da


if __name__ == "__main__":
    cluster = LocalCluster(
        processes=True, n_workers=4, threads_per_worker=1, memory_limit=0
    )
    client = Client(cluster)

    # Write a zarr array to disk (requires 100GB free disk space!)
    # Comment this out once you've run it once.
    data = da.zeros((12500000, 1000), chunks=(12500000, 1))
    ds = xr.Dataset({"data": (("x", "y"), data)})
    ds.to_zarr("test.zarr")
    print("Saved zarr")

    # Do the same array-sum example, but from zarr.
    # This _should_ behave just as well, but on the dashboard we can see lots of transfers.
    # In fact, the dashboard looks the same as running this test on main.
    ds_zarr = xr.open_zarr("test.zarr")
    ds_zarr.sum("y").data.data.compute()
    # TODO where is `sum-` getting scheduled?? Doesn't seem to be going through `decide_worker`
```
 Here's the full (optimized) graph with order coloring. You can see that it looks just like the non-zarr one, except every sum task depends on the one zarr array task: [mydask.pdf](https://github.com/dask/distributed/files/6640189/mydask.pdf)

Screen Shot 2021-06-11 at 11 42 29 AM (2)

distributed/scheduler.py
```diff
@@ -7488,36 +7532,92 @@ def decide_worker(
...
     """
-    ws: WorkerState = None
     wws: WorkerState
-    dts: TaskState
+
+    group: TaskGroup = ts._group
+    ws: WorkerState = group._last_worker
+
+    if valid_workers is not None:
+        total_nthreads = sum(wws._nthreads for wws in valid_workers)
```
mrocklin, line 7541: This walks through all workers for all tasks.  We may not be able to do this.

distributed/scheduler.py
```diff
@@ -7488,36 +7532,92 @@ def decide_worker(
...
+
+    group_tasks_per_worker = len(group) / total_nthreads
```
mrocklin, line 7543: What happens here if `valid_workers is None`?  It looks like total_nthreads might be undefined

distributed/scheduler.py
```diff
@@ -7488,36 +7532,92 @@ def decide_worker(
...
         for ws in candidates:
             break
+        print(f"1 candidate - {ts.group_key} -> {ws.name}")
     else:
         ws = min(candidates, key=objective)
+        print(f"picked worker - {ts.group_key} -> {ws.name}")
+
+    group._last_worker = ws
+    group._last_worker_tasks_left = math.floor(group_tasks_per_worker)
+    group._last_worker_priority = ts.priority
```
mrocklin, line 7620: Do we want to do this when we're not in a root-ish task situation?

distributed/scheduler.py
gjoseph92: Then we use the `self._total_nthreads` that `SchedulerState.decide_worker` passed in (see https://github.com/dask/distributed/pull/4899/files#diff-bbcf2e505bf2f9dd0dc25de4582115ee4ed4a6e80997affc7b22122912cc6591R2376). So in the common case of no worker restrictions, we won't have to do the `total_nthreads` calculation.

distributed/scheduler.py
gjoseph92: Yes, because we're also using this as our base case when `group._last_worker` is not set. So for the first root-ish task, we fall through to here, and then need to store our decision for the next `decide_worker` cycle. We could skip doing this if we've identified that the task is not root-like (which would save us from ever checking that again), though it might make the code a little more tangled.

distributed/scheduler.py
gjoseph92: See below; I believe `valid_workers is None` is the common case? Agreed that this isn't ideal though. But if there are worker restrictions, ignoring them and just using `self._total_nthreads` could be wildly wrong (imagine 10 GPU workers and 100 CPU workers for a task group of 50 that needs to run on GPUs). Maybe there's a cheaper measurement?

distributed/scheduler.py
mrocklin: > So for the first root-ish task, we fall through to here

We can maybe guard this code block with a condition that we're root-ish then?

distributed/scheduler.py
mrocklin: OK, grand.

distributed/scheduler.py
gjoseph92: Yes—though we currently skip determining if we're root-ish when `last_worker` is None. So we'd need some other sentinel value to distinguish "no last worker because we haven't checked yet if it's root-ish" from "no last worker because it's not root-ish".

mrocklin: We currently check the following:

```python
        and group_tasks_per_worker > 1  # group is larger than cluster
        and (  # is a root-like task (task group is large, but depends on very few tasks)
            sum(map(len, group._dependencies)) < 5  # TODO what number
        )
```

I recommend that we define that as a variable at the top and then use it in a couple of places.  Would that work?

mrocklin:
Alternate solution that uses occupancy here: https://github.com/dask/distributed/pull/4922

distributed/scheduler.py
```diff
@@ -7520,9 +7520,46 @@ def decide_worker(
...
+        e   f   g   h
+        |   |   |   |
+        a   b   c   d
+        \   \  /   /
+             X
+
+    In the above case, we want ``a`` and ``b`` to run on the same worker,
+    and ``c`` and ``d`` to run on the same worker, reducing future
+    data transfer. We can also ignore the location of ``X``, because
+    as a common dependency, it will eventually get transferred everywhere.
```
mrocklin, line 7547: <3 the ascii art

Comment/question: Do we want to explain all of this here?  Historically I haven't put the logic behind heuristics in the code.  This is a subjective opinion, and far from universal, but I find that heavily commented/documented logic makes it harder to understand the code at a glance.  I *really* like that the current decide_worker implementation fits in a terminal window.  I think that single-line comments are cool, but that long multi-line comments would better be written as documentation.  

Thoughts?  If you are not in disagreement then I would encourage us to write up a small docpage or maybe a blogpost and then link to that external resource from the code.

gjoseph92: I was also planning on updating https://distributed.dask.org/en/latest/scheduling-policies.html#choosing-workers, probably with this same ascii art. So just linking to that page in the docstring seems appropriate.

gjoseph92:
Closing in favor of #4967, which is similar logic but more concise. It also removes `_last_worker_tasks_left` and `_last_worker_priority`, and puts the root-ish checks in `SchedulerState.decide_worker`, where they're a little easier to do.

PR #4900, opened 2021-06-10
Handle HTTPClientError in websocket connector

marcosmoyano:
- [ ] Closes #xxxx
- [ ] Tests added / passed

Catch websocket client connection errors and re-raise them as `CommClosedError` so that they are handled correctly by the Client.

Scenario:
A Client tries to reconnect to a closed Cluster behind a proxy. The proxy closes the connection back to the client since it can't connect to the Cluster which leads to a HTTPClientError that gets bubbled up.

Not exactly sure how to write an efficient test for this 🤔 . Thoughts?

### DIFF
--- a/distributed/comm/ws.py
+++ b/distributed/comm/ws.py
@@ -7,7 +7,7 @@ from ssl import SSLError
 from typing import Callable
 
 from tornado import web
-from tornado.httpclient import HTTPRequest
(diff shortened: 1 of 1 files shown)


mrocklin:
cc @jacobtomlinson

mrocklin:
This is in.  Thanks @marcosmoyano

PR #4909, opened 2021-06-11
Add scheduler log tab to performance reports

### LINKED ISSUE #4781: Include scheduler logs in performance_report
We might want to add a new tab in the performance report that includes the logs from the scheduler

charlesbluca:
Adds a tab to the performance report with the logs of the scheduler. Note that these are all the logs currently contained in the scheduler's log deque. and not just the logs generated in the `performance_report` context:

image

- [ ] Closes #4781
- [ ] Tests added / passed

### DIFF
--- a/distributed/dashboard/components/scheduler.py
+++ b/distributed/dashboard/components/scheduler.py
@@ -32,10 +32,12 @@ from bokeh.models import (
     value,
 )
 from bokeh.models.widgets import DataTable, TableColumn
+from bokeh.models.widgets.markups import Div
 from bokeh.palettes import Viridis11
 from bokeh.plotting import figure
 from bokeh.themes import Theme
 from bokeh.transform import cumsum, factor_cmap, linear_cmap
+from jinja2.environment import Template
 from tlz import curry, pipe
 from tlz.curried import concat, groupby, map
 from tornado import escape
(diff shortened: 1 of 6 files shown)


mrocklin:
This looks good to me.  cc @jacobtomlinson

fjetter:
> Note that these are all the logs currently contained in the scheduler's log dequ
we could add timestamps to the [get_logs](https://github.com/dask/distributed/blob/b707c296dd529b7b669531b7bc7dad96387e4a80/distributed/node.py#L90-L97)

e.g.
```python
    def get_logs(self, comm=None, n=None, timestamps=False):
        deque_handler = self._deque_handler
        if n is None:
            L = list(deque_handler.deque)
        else:
            L = deque_handler.deque
            L = [L[-i] for i in range(min(n, len(L)))]
        if timstamps:
            return [(msg.created, msg.levelname, deque_handler.format(msg)) for msg in L]
        else:
            return [(msg.levelname, deque_handler.format(msg)) for msg in L]
```
with the toggle this is fully backwards compat and the report could filter

This can be obviously done in another PR

jacobtomlinson (commented):
This looks great.

I just want to draw your attention to the existing `Log` and `Logs` objects which we use to give HTML reprs to `get_logs` calls today.

https://github.com/dask/distributed/blob/b707c296dd529b7b669531b7bc7dad96387e4a80/distributed/utils.py#L1272-L1292

I wonder if these can be ehnanced/reused here?

quasiben:
Would you mind also adding a test or an assertion line to the [performance_report test](https://github.com/dask/distributed/blob/e690e82cd26b39fe87a4dd1886b71e6dc31c5a26/distributed/tests/test_client.py#L6347-L6348) ?  

Thanks for picking up this issue @charlesbluca !

charlesbluca:
Sure, thanks for catching that @quasiben 🙂 

Thanks for the tip @fjetter! Timestamps could also be useful as part of the log output itself I'd imagine - would definitely be happy to open up a follow up PR implementing that.

I like the idea of reusing the HTML reprs here, I think a potential refactor to match the behavior of this Jinja template would be to change these objects such that:

- the `Log` object returns a single line styled with inline CSS conditional on the log level
- the `Logs` object returns multiple log lines concatenated together (what `Log` is currently doing)
- a new `MultiLogs`(?) object returns the log lines wrapped in `<detail>` blocks (what `Logs` is currently doing)

jacobtomlinson:
Flaky test, rerunning.

mrocklin:
Are we good to merge here?

charlesbluca:
Think so, the test failures seem to be flaky - @quasiben is the added assertion in the perf report test sufficient?

PR #4912, opened 2021-06-13
Remove tmpfile utility context manager

### LINKED ISSUE #4910: Remove `tmpfile` utility context manager
Similar to https://github.com/dask/dask/issues/7800, we can replace our `tmpfile` context manager

https://github.com/dask/distributed/blob/5e5b9835a5e6929bb97acf94f0a2eecccc0b6a2f/distributed/utils.py#L862-L863

with [`tempfile.TemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) from the standard library.

Patil2099:
Fixes #4910

### DIFF
--- a/distributed/utils.py
+++ b/distributed/utils.py
@@ -26,6 +26,7 @@ from importlib.util import cache_from_source
 from time import sleep
 from typing import Any as AnyType
 from typing import Dict, List
+import tempfile
 
 import click
 import tblib.pickling_support
@@ -861,21 +862,8 @@ def read_block(f, offset, length, delimiter=None):
 
 @contextmanager
 def tmpfile(extension=""):
-    extension = "." + extension.lstrip(".")
-    handle, filename = tempfile.mkstemp(extension)
-    os.close(handle)
-    os.remove(filename)
+    return tempfile.TemporaryFile(extension)
 
-    yield filename
-
-    if os.path.exists(filename):
-        try:
-            if os.path.isdir(filename):
-                shutil.rmtree(filename)
-            else:
-                os.remove(filename)
-        except OSError:  # sometimes we can't remove a generated temp file
-            pass
 
 
 def ensure_bytes(s):


distributed/utils.py
```diff
@@ -861,21 +862,8 @@ def read_block(f, offset, length, delimiter=None):
 
 @contextmanager
 def tmpfile(extension=""):
-    extension = "." + extension.lstrip(".")
-    handle, filename = tempfile.mkstemp(extension)
-    os.close(handle)
-    os.remove(filename)
-
-    yield filename
+    return tempfile.TemporaryFile(extension)
```
fjetter, line 865: This raises when used with the default

```python
In [1]: import tempfile

In [2]: tempfile.TemporaryFile("")
<ipython-input-2-44ed0ba8a626> in <module>
----> 1 tempfile.TemporaryFile("")

~/mambaforge/envs/dask-distributed/lib/python3.9/tempfile.py in TemporaryFile(mode, buffering, encoding, newline, suffix, prefix, dir, errors)
    615         try:
    616             _os.unlink(name)
--> 617             return _io.open(fd, mode, buffering=buffering,
    618                             newline=newline, encoding=encoding, errors=errors)
    619         except:

ValueError: Must have exactly one of create/read/write/append mode and at most one plus
```

jrbourbeau (commented):
Thanks for the pull request @Patil2099! Instead of using `tempfile.TemporaryFile` from within `distributed.utils.tmpfile`, https://github.com/dask/distributed/issues/4910 is suggesting that we replace all the times we use `tmpfile` in `distributed` with calls to `tempfile.TemporaryFile` instead. You can use

```
git grep "tmpfile("
```

from within the root of the `distributed` repository to find all the times we call `tmpfile` internally.

tharris72:
> Thanks for the pull request @Patil2099! Instead of using `tempfile.TemporaryFile` from within `distributed.utils.tmpfile`, #4910 is suggesting that we replace all the times we use `tmpfile` in `distributed` with calls to `tempfile.TemporaryFile` instead. You can use
> 
> ```
> git grep "tmpfile("
> ```
> 
> from within the root of the `distributed` repository to find all the times we call `tmpfile` internally.

I made the changes and made a pull request. Can you let me know if the changes were good? It should be straight forward. I removed the tmpfile function in utils.py and replaced all calls with tempfile.TemporaryFile after I added the import statement.
Thanks

GPUtester:

ncclementi:
Checkin in here. The issue #4910 mentions a similar issue on the Dask end which is https://github.com/dask/dask/issues/7800 and the corresponding PR was closed (not merged) due to `tmpdir` and `tmpfile` still being used in some doctests, see https://github.com/dask/dask/pull/7803#issuecomment-893899855 
Would this also apply in distributed?  If that is the cased we might want to close this PR too. 

cc: @jrbourbeau

jrbourbeau:
Thanks for following-up here @ncclementi! I agree https://github.com/dask/dask/pull/7803#issuecomment-893899855 also applies to `distributed.utils.tmpfile`. That said, both `distributed` and `dask` define their own (largely duplicated) `tmpfile` utilities. I pushed up https://github.com/dask/distributed/pull/5361 which proposes we consolidate utilities and deprecate `distributed.utils.tmpfile` in favor of `dask.utils.tmpfile` -- comments are welcome.

jrbourbeau:
Superseded by https://github.com/dask/distributed/pull/5361

PR #4916, opened 2021-06-14
Move UCX/RMM config variables to Distributed namespace 

### LINKED ISSUE #4904: Prefixing UCX/RMM config variables with `distributed.`
As discussed in dask/dask#7579, the configuring variables for UCX/RMM are not contained in the namespace of Distributed, despite only being used there. With that in mind, it would make sense moving forward to prefix these variables with `distributed.`, and maybe have some deprecation period to handle scripts using UCX/RMM with the old variables. In general, we should be able to prefix old variables with a new alias:

```diff
diff --git a/distributed/config.py b/distributed/config.py
index 4b1cfb35..bebc0560 100644
--- a/distributed/config.py
+++ b/distributed/config.py
@@ -48,6 +48,7 @@ aliases = {
     "log-length": "distributed.admin.log-length",
     "log-format": "distributed.admin.log-format",
     "pdb-on-err": "distributed.admin.pdb-on-err",
+    "ucx": "distributed.ucx",
 }

 dask.config.rename(aliases)
```

But I'm not sure if that would cover all the cases where UCX/RMM is configured for Dask-CUDA.

cc @pentschev

charlesbluca:
Moves UCX/RMM configuring variables to be nested under Distributed's config namespace, including aliases for compatibility with the old variable names. Will probably need a PR in Dask-CUDA to update handling there as well.

- [ ] Closes #4904
- [ ] Tests added / passed


jrbourbeau (commented):
This restructure seems sensible to me. We should also update [`distributed.yaml`](https://github.com/dask/distributed/blob/42d631dea9f0e64f4553cf53230167457525eaae/distributed/distributed.yaml) and [`distributed-schema.yaml`](https://github.com/dask/distributed/blob/42d631dea9f0e64f4553cf53230167457525eaae/distributed/distributed-schema.yaml) to reflect the updated reorganization

cc @pentschev @quasiben

charlesbluca:
Is there anything else that needs to be done here, or can we rerun tests and merge?

jrbourbeau (commented):
Yeah, I think this should be good to go after the merge conflict is resolved 

Also, can you confirm that users who were previously setting the old config values (e.g. `ucx.tcp`) will have them properly converted to the new version (e.g. `distributed.comm.ucx.tcp`)? The config alias should handle this, but it'd be good to double check

pentschev:
Thanks for the reminder @jrbourbeau , indeed we'll have to make an announcement to our users when this is merged. Today, UCX settings are generally set via environment variables by most of our users, so `DASK_UCX__TCP` will have to be written as `DASK_DISTRIBUTED__COMM__UCX__TCP`.

jrbourbeau:
> Today, UCX settings are generally set via environment variables by most of our users, so DASK_UCX__TCP will have to be written as DASK_DISTRIBUTED__COMM__UCX__TCP

Since @charlesbluca is setting config alias between the old and new config names in `distributed/config.py`, users should still be able to use the old environment variables and things will work with the changes in this PR. For example, running

```
# Use old name for env variable, but use new name for getting the config value
$ DASK_UCX__TCP="foo" python -c "import dask, distributed; print(dask.config.get('distributed.comm.ucx.tcp'))"
```

locally (with this PR checked out) outputs

```
foo
```

pentschev:
Indeed, but I believe the idea is to deprecate the `DASK_UCX__*` structure and eventually remove it, so it's best to let users know they should move to the new structure.

jrbourbeau:
> Indeed, but I believe the idea is to deprecate the DASK_UCX__* structure and eventually remove it, so it's best to let users know they should move to the new structure

Ah, sorry I don't think I was super clear in my previous comment. I meant to say that guiding users towards the new config names is definitely a good thing to do 👍 Though things shouldn't totally break because of the config alias

This looks good to go. @charlesbluca @pentschev we're releasing `distributed` tomorrow. Would you like this to be included in that release, or would more time be desirable to test things out on the dask-cuda side?

charlesbluca:
IMO I would prefer this not go into tomorrow's release, as I imagine we'd want to make some changes to dask-cuda's docs to announce this change, and that project is currently in burn down. @pentschev does this sound good?

pentschev:
Yup, I agree with @charlesbluca , delaying this for next release (not the one tomorrow) is probably the best choice.

jrbourbeau:
Sounds good. I'll plan on merging this PR _after_ the release tomorrow then 👍

pentschev:
Awesome, thanks @jrbourbeau !

pentschev:
@charlesbluca could you confirm how you verified the `ucx`->`distributed.comm.ucx` functionality? To me it doesn't seem to be working. For example `DASK_UCX__NVLINK=True` didn't have any effect for me but `DASK_DISTRIBUTED__COMM__UCX__NVLINK=True` worked, and we also had to update using UCX config variables in Dask-CUDA in https://github.com/rapidsai/dask-cuda/pull/695 because the old `ucx` namespace didn't work anymore.

jrbourbeau:
Just checking in here. @pentschev I see https://github.com/rapidsai/dask-cuda/pull/695 was merged. Does that handle the fallout from this config migration, or are there still concerns you have?

pentschev:
> Does that handle the fallout from this config migration, or are there still concerns you have?

It does in Dask-CUDA internally, but my understanding was that the alias should have worked with that. In that particular case it's not a big deal, but it will be for users as we have _a lot_ of them specifying variables like `DASK_UCX__NVLINK=True` that doesn't work anymore, and they will all need to change to `DASK_DISTRIBUTED__COMM__UCX__NVLINK=True`.

Ideally, we would have that alias in place with a deprecation warning. Is this possible to do with Dask configuration schema?

jrbourbeau:
> my understanding was that the alias should have worked with that

Sorry, I might have been unclear in my previous comments. What I meant to convey is that with changes like those in https://github.com/rapidsai/dask-cuda/pull/695, things should be backwards compatible. For example, on the current `main` branch of `distributed` if I set the old environment variable, but try to read in the new config value things work:

```
# Use old name for env variable, but use new name for getting the config value
$ DASK_UCX__TCP="foo" python -c "import dask, distributed; print(dask.config.get('distributed.comm.ucx.tcp'))"
foo
```

This means that `dask-cuda` needs to be updated to use the new config names (like https://github.com/rapidsai/dask-cuda/pull/695 does), but users who might still be using the old environment variables won't be broken when they upgrade their packages

jrbourbeau:
> Ideally, we would have that alias in place with a deprecation warning. Is this possible to do with Dask configuration schema?

The alias will handle the internal mapping of the old -> new name, but won't emit a deprecation warning. Though maybe we should be. That sounds like a nice upstream improvement

jrbourbeau:
Actually warning users is possible. See https://github.com/dask/dask/pull/7956

pentschev:
You're absolutely right @jrbourbeau , the problem is very simple: I'm an idiot. When I tested, I was on Distributed 2021.7.1 instead of main. 🤦 

Sorry about the noise.

jrbourbeau:
No worries at all. If I had a nickel for every time I had the wrong commit of a local repo checked out, I'd be a rich man...

PR #4920, opened 2021-06-15
Improve work stealing for scaling situations

### LINKED ISSUE #4471: Poor work scheduling when cluster adapts size
**What happened**:

When a cluster is autoscaled in increments while running, as can happen in a GKE cluster, the work concentrates on few workers and uses the cluster inefficiently.  This seems to be worse when there are long running tasks.  The example below simulates this by adjusting a local cluster's size as it is processing a graph with 10-second tasks.

The image below shows the final look of the task graph, and the animated gif shows the status screen as the cluster processes the graph.  Many workers do zero long tasks, and only a few workers seem to be fully utilised.  

If the cluster is initially set to 20 workers with no changes, work is distributed evenly and all workers are efficiently used.

@fjetter In this example, the workers under load are green a lot.

After the cluster scales up, the work should be evenly divided among all the workers.

```python
import time
import distributed
import dask.array as da

client = distributed.Client(n_workers=1, threads_per_worker=1, memory_limit='1G')
print(client.dashboard_link)

# Slow task.
def func1(chunk):
    if sum(chunk.shape) != 0: # Make initialization fast
        time.sleep(10)
    return chunk

def func2(chunk):
    return chunk

data = da.zeros((30, 30, 30), chunks=5)
result = data.map_overlap(func1, depth=1, dtype=data.dtype)
result = result.map_overlap(func2, depth=1, dtype=data.dtype)
future = client.compute(result)

print('started computation')

time.sleep(11)
print('scaling to 4 workers')
client.cluster.scale(4)

time.sleep(5)
print('scaling to 20 workers')
client.cluster.scale(20)

_ = future.result()
```

In a real cluster with my real load (not in the simulation above), I also will see the scheduler CPU pegged near 100% (possibly due to #3898), even when all workers are working on the long tasks.  This seems odd, since nothing is being actively changed in the scheduling.

**Environment**:

- Dask (and distributed) version: 2021.01.1
- Python version: 3.8.5
- Operating System: Ubuntu 20.04

Final task screenshot:
Screen Shot 2021-01-30 at 7 23 00 AM

Movie of workload (at 10x speed):
Kapture 2021-01-30 at 07 28 34

fjetter:
These are a few preliminary fixes which would close #4471. While this does not necessarily account for opportunity cost as I described in https://github.com/dask/distributed/issues/4471#issuecomment-861483765 it still remedies the situation by

1. Allow idle workers to be part of the decision when scheduling a task. the worker objective will take care of the rest and _should_ assign it to the proper worker if it is not overloaded. That might backlash and should be tested in another context
2. The work stealing ratio calculation emits sentinel values in situations which are arguable not situations we should skip on. As I argued in https://github.com/dask/distributed/issues/4471#issuecomment-861483765  there are situations where we should consider stealing even if the task is incredibly cheap to compute since it might allow for further parallelism. In particular if the cluster is partially idling. 
3. This is a minor tuning and not connected to the problem but I chose to not pick the thief based on round robin but on the worker objective, as we do for the initial decision. We might need to tweak this since it doesn't account for in-flight occupancy and might cause another overload of a worker.

Tests missing (TODO), feedback welcome

- [x] Closes #4471 
- [ ] Tests added / passed


fjetter:
I completely forgot, performance reports

Before

https://gistcdn.githack.com/fjetter/6e04f5ced602378977c8fea0037e2c31/raw/b8dbbd5ae962ede8c0853926649619714ae1ded3/gh4471.html

After

https://gistcdn.githack.com/fjetter/6e04f5ced602378977c8fea0037e2c31/raw/b8dbbd5ae962ede8c0853926649619714ae1ded3/gh4471_combined.html

distributed/stealing.py
```diff
@@ -113,27 +114,21 @@ def steal_time_ratio(self, ts):
         For example a result of zero implies a task without dependencies.
         level: The location within a stealable list to place this value
         """
-        if not ts.dependencies:  # no dependencies fast path
-            return 0, 0
```
mrocklin, line 117: Why not keep this?  My understanding is that this line says "if you don't have any dependencies then we can move you any time". This seems great to me :)

distributed/stealing.py
```diff
@@ -344,7 +339,10 @@ def maybe_move_task(level, ts, sat, idl, duration, cost_multiplier):
                             thieves = idle
                         if not thieves:
                             break
-                        thief = thieves[i % len(thieves)]
+
+                        thief = min(
+                            thieves, key=partial(self.scheduler.worker_objective, ts)
+                        )
```
mrocklin, line 345: How many thieves can there be here?  If the answer is "maybe a lot" then we might need to be careful here.

mrocklin: I'm also curious how much impact this choice in particular had on the workload we're testing against.

distributed/scheduler.py
```diff
@@ -7471,7 +7472,11 @@ def _reevaluate_occupancy_worker(state: SchedulerState, ws: WorkerState):
 @cfunc
 @exceptval(check=False)
 def decide_worker(
-    ts: TaskState, all_workers, valid_workers: set, objective
+    ts: TaskState,
+    all_workers: set,
+    valid_workers: Optional[set],
+    idle: set,
+    objective,
```
mrocklin, line 7479: Everything here seems fine cython-wise, but pinging @jakirkham just in case

distributed/stealing.py
fjetter: That's very likely still happening. The only difference is that this now accounts for latencies.

distributed/stealing.py
fjetter: > How many thieves can there be here? If the answer is "maybe a lot" then we might need to be careful here.

Worst case situation would be if all but one worker are idling. then we'd have `W-1` potential thieves. I thought about sampling this down to 20 or smth like that to avoid any extreme cases. Sampling and picking the best of the sample still sounds better than random. However...

> I'm also curious how much impact this choice in particular had on the workload we're testing against.

Nothing, at all. This is the one chance which didn't actually impact this particular example. It just _felt_ right? idk. That's one of the things I want to test on different workloads

distributed/stealing.py
fjetter: I figured: if the outcome is the same regardless, we would we need two if-clauses? (this one and the one further below). We do not really gain in speed either since it's all cheap, constant time operations below

fjetter:
FYI tests break because the stealing now behaves different and the tests reflect that (good!). just wanted to get this out early in case smbd has input already.

distributed/stealing.py
fjetter: In particular for workloads with actual payload data. If the dependents don't have data, this is a rather extreme stealing edge case and I'd like to verify that this also behaves nicely for proper payloads

mrocklin: In decide_worker we choose the best worker if there are less than 20 workers and we choose an incrementing worker if there are more than 20 workers.  

In practice though I doubt that it matters much.  The difference in occupancy between thieves and victims means that we're making the world significantly better with the dumb change.  If this results in a suboptimal world that's ok... we'll just improve it again in a future cycle if it becomes a problem.

fjetter: > In decide_worker we choose the best worker if there are less than 20 workers and we choose an incrementing worker if there are more than 20 workers.

I also thought about just using decide_worker. After all, why would the logic be any different (given that decide_worker takes a whitelist <-> thieves). However, the important logic is in the objective function which is why I picked that. If this things turns out to be valuable, we can have a closer look

> In practice though I doubt that it matters much. The difference in occupancy between thieves and victims means that we're making the world significantly better with the dumb change. 

True. As I said, the current problem doesn't actually trigger anything where this would matter. If I can not prove an improvement, I'll remove it again.

> The difference in occupancy

The only thing I would like to leverage here is to pick a thief with dependencies, if that exists. that's implicitly done by the worker objective but could also be made explicit. My gut tells me this can make a difference but you might be right that this is only marginal.

> we'll just improve it again in a future cycle if it becomes a problem.

There will not be another cylce to correct this unless the worker becomes saturated soon-ish, will there?. Non-saturated workers are not victimized.

mrocklin: > There will not be another cylce to correct this unless the worker becomes saturated soon-ish, will there?. Non-saturated workers are not victimized.

Correct, but the stolen task probably won't be run immediately anyway.  It's probably at the end of the queue.  As the computation continues progressing and some workers become more free before others the saturated/non-saturated split will grow and trigger another cycle of transfers if necessary.

distributed/scheduler.py
```diff
@@ -7471,7 +7472,11 @@ def _reevaluate_occupancy_worker(state: SchedulerState, ws: WorkerState):
 @cfunc
 @exceptval(check=False)
 def decide_worker(
-    ts: TaskState, all_workers, valid_workers: set, objective
+    ts: TaskState,
+    all_workers: set,
+    valid_workers: Optional[set],
```
jakirkham, line 7477: Cython doesn't handling `typing` objects like `Optional` yet. Though it already happily accepts `None` and checks for that

```suggestion
    valid_workers: set,
```

distributed/scheduler.py
jakirkham: There's one issue. Commented inline here ( https://github.com/dask/distributed/pull/4920/files#r652002105 )

distributed/scheduler.py
```diff
@@ -2340,8 +2340,9 @@ def decide_worker(self, ts: TaskState) -> WorkerState:
         if ts._dependencies or valid_workers is not None:
             ws = decide_worker(
                 ts,
-                self._workers_dv.values(),
+                set(self._workers_dv.values()),
```
jakirkham, line 2343: AIUI this could be kind of expensive if we don't need it, which is why the original code didn't do this. However it was that way even prior to Cythonizing the Scheduler

jakirkham, line 2343: In particular this line

distributed/scheduler.py
```diff
@@ -2340,8 +2340,9 @@ def decide_worker(self, ts: TaskState) -> WorkerState:
...
                 valid_workers,
+                set(self._idle_dv),
```
jakirkham, line 2345: ...and this one

jakirkham (commented):
AIUI these could be kind of expensive if we don't need them, which is why the original code didn't coerce them to `set`s. However it was that way even prior to Cythonizing the Scheduler

distributed/scheduler.py
fjetter: I added optional mostly for type checking, e.g. in IDEs. is cython having trouble with this or does it simply ignore it?

distributed/stealing.py
mrocklin: > We do not really gain in speed either since it's all cheap, constant time operations below

I think that it's reasonable to avoid these constant time operations.  Our budget per-task is currently around 200us over the lifetime of the task (and we're trying to reduce that).  There are enough attribute access and dict access calls that it seems non-trivial to me.

A fast path for the common case of no dependencies seems sensible here to me, although I haven't done any measurement to back that up.

mrocklin:
Summarizing things a bit here, I'm seeing two changes:

1.  We don't only look at workers with relevant dependencies, we also consider idle workers.  I agree that this is a big benefit.
2.  We don't do a round robin, but instead use the worker objective.  I agree that this is often a benefit, I am concerned about introducing checks that depend on the number of workers for each task.  I think that we may want to defend against this.  There are a few options:
    - Round-robin defends against this but results in suboptimal scheduling.  One answer is that we keep doing round-robin until we identify that it results in user problems.  I wouldn't be surprised if it didn't over the lifetime of a computation because we'll continue correcting things.
    - Use worker objective if the number of theives is small (less than twenty) but fall back to round robin otherwise (we do this elsewhere)
    - Select a random sample of the theives (maybe ten?) and then use the worker objective on that sample (idea from @gjoseph92 ).  Maybe this gives us the best of both worlds?

fjetter:
Change 2B: relax the sentinel values in `steal_time_ratio`. While it is ok do use a short path for no-dependency tasks to save compute cost, we should not disallow work stealing for generally small or expensive tasks (<0.005 or cost_multiplier > 100). I believe these two cases should not be short-circuited but rather be dealt with using the usual logic. If it is too expensive and doesn't payoff they are not stolen but there is a remote chance that it would still pay off to move them in extreme situations (as in this example). If we do not remove these sentinels, they are forever blacklisted and will never be stolen.


> Select a random sample of the theives (maybe ten?) and then use the worker objective on that sample (idea from

I like this idea so much that I wish it was my own ;) https://github.com/dask/distributed/pull/4920#discussion_r651876940

mrocklin:
> we should not disallow work stealing for generally small or expensive tasks (<0.005 or cost_multiplier > 100). I believe these two cases should not be short-circuited but rather be dealt with using the usual logic

That's ok with me

> I like this idea so much that I wish it was my own ;) #4920 (comment)

Whoops!  Well then it sounds like it must be a great idea :)

fjetter:
> Whoops! Well then it sounds like it must be a great idea :)

I'm not entirely certain to be honest. This is one of the cases where I would really love some data. After all, what are the chances that a sampling would keep the one or two workers which carry dependencies? If I cannot back this up with data, I'm inclined to roll this back.

However, overall I'd really love to have some micro benchmarks on this to know how bad this actually is. The comments the two of you left behind about performance make 100% sense but it is all a bit fuzzy and sometimes hard to spot.

For instance, I would've thought that the following operations would be A) faster and B) identical
image but I was wrong. Twice.

Anyhow, this is offtopic but I would love to discuss the topic of benchmarking with a few ppl (maybe next dev meeting)

mrocklin:
> After all, what are the chances that a sampling would keep the one or two workers which carry dependencies? 

Ah, my thinking (or rather Gabe's thinking) was that you would keep the workers with dependencies in the set, but then mix in a few others from outside of the set.

fjetter:
> Ah, my thinking (or rather Gabe's thinking) was that you would keep the workers with dependencies in the set

Right... that sounds straight forward. there is a saying in germany "Manchmal sieht man den Wald for lauter Baeumen nicht" which translates loosely to "Sometimes you loose sight of the forest because there are too many trees."
No idea if this makes sense to you but I think that's a good idea :D (and better than mine, after all)

mrocklin:
> Sometimes you loose sight of the forest because there are too many trees

This saying is commonly known in English as well.  It's a good saying.

mrocklin:
https://www.dictionary.com/browse/can-t-see-the-forest-for-the-trees#:~:text=An%20expression%20used%20of%20someone,the%20bill%20could%20never%20pass.%E2%80%9D

distributed/scheduler.py
```diff
@@ -7471,7 +7472,11 @@ def _reevaluate_occupancy_worker(state: SchedulerState, ws: WorkerState):
 @cfunc
 @exceptval(check=False)
 def decide_worker(
-    ts: TaskState, all_workers, valid_workers: set, objective
+    ts: TaskState,
+    all_workers: set,
+    valid_workers: Optional[set],
+    idle: set,
```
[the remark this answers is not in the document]
fjetter: oops. forget it while moving stuff around. should be

```diff
diff --git a/distributed/scheduler.py b/distributed/scheduler.py
index d916d8de..be86bf52 100644
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -7503,6 +7503,7 @@ def decide_worker(
         candidates = all_workers
     else:
         candidates = {wws for dts in deps for wws in dts._who_has}
+        candidates |= idle
     if valid_workers is None:
         if not candidates:
             candidates = all_workers
```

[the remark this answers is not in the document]
gjoseph92: BTW this is covered by #4925

fjetter:
(Code still needs to be updated)

I removed all the changes around decide_worker and solely focused on the blacklisting of `cost_multiplier > 100` and `duration < 0.005` tasks (which are often the same). Below you can see the stealing tab (tab includes fixes, PR upcoming) which shows the stealing activitiy and occupancy over time

## main / default

Screenshot 2021-06-21 at 17 31 32


## without blacklisting "expensive" tasks


Screenshot 2021-06-21 at 17 27 38


What we can see is that the occupancy (timeseries chart, top) is much more effectively balanced if the small tasks are not blacklisted. I do not have proper perf reports or screenshots but the task stream density behaves similarly that the well balanced occupancy is much denser.

mrocklin:
I spoke with @fjetter about this after replicating some of this work in https://github.com/dask/distributed/pull/6115 .  

It sounds like there isn't a major blocker here.  I'm going to see if I can push it through today.  @fjetter if there are any major concerns you have on this PR that weren't listed above then please let me know.

distributed/stealing.py
```diff
@@ -450,7 +453,10 @@ def maybe_move_task(level, ts, sat, idl, duration, cost_multiplier):
                         thieves = _potential_thieves_for(ts, idle)
                         if not thieves:
                             continue
-                        thief = thieves[i % len(thieves)]
+                        thief = self._maybe_pick_thief(ts, thieves)
+                        if not thief:
+                            thief = thieves[i % len(thieves)]
```
gjoseph92, line 458: nit: might be easier to read if this was moved into `_maybe_pick_thief` (and it was no longer `maybe`).

distributed/stealing.py
```diff
@@ -464,6 +470,32 @@ def maybe_move_task(level, ts, sat, idl, duration, cost_multiplier):
...
+            who_has = set()
+            for dep in ts._dependencies:
+                who_has.update(dep.who_has)
+
+            thieves_with_data = list(who_has & set(thieves))
+
+            # If there are potential thieves with dependencies we
+            # should prefer them and pick the one which works best.
+            # Otherwise just random/round robin
+            if thieves_with_data:
```
gjoseph92, line 491: If, say, just 1 idle worker holds dependencies, is it possible for that worker to get slammed with new tasks because of this? What has to happen between picking a worker as a thief and it getting removed from the idle set?

distributed/stealing.py
fjetter: right, I figured this is nicer because it felt weird passing `i` to `_maybe_pick_thief`.

distributed/stealing.py
fjetter: > If, say, just 1 idle worker holds dependencies, is it possible for that worker to get slammed with new tasks because of this

Yes, this may hurt, particularly for sub-topologies like

```mermaid
flowchart BT
A --> B1
A --> B2
A --> B3
```

where A is very small and B* is very compute heavy, i.e. steal_ratio will always be small and tasks are even allowed to be stolen from "public", i.e. non-saturated workers. I believe for other cases, i.e. if tasks are just stolen from saturated workers, this should be OK since decide_worker is much smarter with initial task placement.

> What has to happen between picking a worker as a thief and it getting removed from the idle set?

Steal request must be confirmed which updates occupancies and recalculates the idle set. Somewhere on this PR a comment of mine is suggesting that this `_mabye_pick_thief` should incorporate `in_flight_occupancy` to reflect for the time until this happens.

fjetter:
> if there are any major concerns you have on this PR that weren't listed above then please let me know.

Dropping the sentinels will emphasize the importance of proper measurements (https://github.com/dask/distributed/pull/6115#issuecomment-1098961206) which is where I ran out of time last year.

I myself would've only picked this up again after having a few solid benchmarks up and running. alternatively a lot of manual testing, maybe both.

PR #4921, opened 2021-06-15
Ensure test_scheduler does not leak open sockets

fjetter:
Poor mans fix for #4806 by manually closing the scheduler in the affected instances. Haven't added a test since testing this either requires root (getting all socks of a proc) or very brittle hard coding of ports.

### DIFF
--- a/distributed/deploy/spec.py
+++ b/distributed/deploy/spec.py
@@ -399,10 +399,16 @@ class SpecCluster(Cluster):
             if self.status == Status.created:
                 await self._start()
             await self.scheduler
-            await self._correct_state()
-            if self.workers:
-                await asyncio.wait(list(self.workers.values()))  # maybe there are more
-            return self
+            try:
+                await self._correct_state()
+                if self.workers:
+                    await asyncio.wait(
+                        list(self.workers.values())
+                    )  # maybe there are more
+                return self
+            except Exception:
+                await self.scheduler.close()
+                raise
 
         return _().__await__()
 
--- a/distributed/tests/test_scheduler.py
+++ b/distributed/tests/test_scheduler.py
@@ -313,10 +313,15 @@ async def test_blocked_handlers_are_respected(s, a, b):
     await comm.close()
 
 
-def test_scheduler_init_pulls_blocked_handlers_from_config():
+@pytest.mark.asyncio
+async def test_scheduler_init_pulls_blocked_handlers_from_config():
+    # This test is async to allow us to properly close the scheduler since the
+    # init is already opening sockets for the HTTP server See also
+    # https://github.com/dask/distributed/issues/4806
     with dask.config.set({"distributed.scheduler.blocked-handlers": ["test-handler"]}):
         s = Scheduler()
     assert s.blocked_handlers == ["test-handler"]
+    await s.close()
 
 
 @gen_cluster()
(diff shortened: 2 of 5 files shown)


distributed/tests/test_scheduler.py
```diff
@@ -313,10 +313,15 @@ def func(scheduler):
...
-def test_scheduler_init_pulls_blocked_handlers_from_config():
+@pytest.mark.asyncio
+async def test_scheduler_init_pulls_blocked_handlers_from_config():
+    # This test is async to allow us to properly close the scheduler since the
+    # init is already opening sockets for the HTTP server See also
+    # https://github.com/dask/distributed/issues/4806
     with dask.config.set({"distributed.scheduler.blocked-handlers": ["test-handler"]}):
         s = Scheduler()
     assert s.blocked_handlers == ["test-handler"]
+    await s.close()
```
mrocklin, line 324:
```suggestion
    with dask.config.set({"distributed.scheduler.blocked-handlers": ["test-handler"]}):
        async with Scheduler() as s:
            assert s.blocked_handlers == ["test-handler"]
```

jrbourbeau: Based on the comment earlier in this test, it looks like we want to check that `Scheduler.__init__` handles blocking handlers. Using `async with Scheduler() as s:` will also call `start` which makes it difficult to isolate `Scheduler.__init__` behavior

mrocklin: OK

setup.cfg
```diff
@@ -45,6 +45,9 @@ parentdir_prefix = distributed-
 addopts = -v -rsxfE --durations=20
 filterwarnings =
     error:Since distributed.*:PendingDeprecationWarning
+
+    # See https://github.com/dask/distributed/issues/4806
+    error:Port:UserWarning:distributed.node
```
fjetter, line 50: This would raise and fail the tests whenever the warning is raised during CI. AFAIU, we should not hit this condition accidentally if we properly close all servers properly.

distributed/deploy/spec.py
```diff
@@ -399,10 +399,16 @@ async def _():
...
+            try:
+                await self._correct_state()
+                if self.workers:
+                    await asyncio.wait(
+                        list(self.workers.values())
+                    )  # maybe there are more
+                return self
+            except Exception:
+                await self.scheduler.close()
+                raise
```
fjetter, line 411: This is new and was also a cause for this warning message. If an exception during worker startup would fail, the scheduler was not cleaned up properly. This should've been implemented regardless of the warning, I believe

distributed/tests/test_scheduler.py
```diff
@@ -1406,6 +1416,7 @@ def test_deque_handler():
     msg = deque_handler.deque[-1]
     assert "distributed.scheduler" in deque_handler.format(msg)
     assert any(msg.msg == "foo123" for msg in deque_handler.deque)
+    await s.close()
```
mrocklin, line 1419: We can probably use `async with Scheduler() as s` here?

distributed/deploy/spec.py
mrocklin: +1

fjetter:
closing this in favour of https://github.com/dask/distributed/pull/4928

PR #4924, opened 2021-06-17
Use deprecation utility from Dask

jrbourbeau:
To avoid duplicate deprecation utilities, this PR removes the `deprecated` in `distributed` in favor of the new `_deprecated` decorator recently added to `dask` (xref https://github.com/dask/dask/pull/7810)

### DIFF
--- a/distributed/deploy/cluster.py
+++ b/distributed/deploy/cluster.py
@@ -3,14 +3,13 @@ import datetime
 import logging
 import threading
 import uuid
-import warnings
 from contextlib import suppress
 from inspect import isawaitable
 
 from tornado.ioloop import PeriodicCallback
 
 import dask.config
-from dask.utils import format_bytes
+from dask.utils import _deprecated, format_bytes
 
 from ..core import Status
 from ..objects import SchedulerInfo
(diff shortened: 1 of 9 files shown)


distributed/worker.py
```diff
@@ -4017,11 +4016,6 @@ def convert_kwargs_to_str(kwargs, max_len=None):
         return "{{{}}}".format(", ".join(strs))
 
 
-@deprecated(version_removed="2021.06.0")
-def weight(k, v):
-    return sizeof(v)
```
jrbourbeau, line 4022: I went ahead and removed `weight` since `2021.06.0` is already out

distributed/deploy/cluster.py
```diff
@@ -248,8 +247,8 @@ def get_logs(self, cluster=True, scheduler=True, workers=True):
             self._get_logs, cluster=cluster, scheduler=scheduler, workers=workers
         )
 
+    @_deprecated(use_instead="get_logs")
```
jrbourbeau, line 250: This changes the warning that's emitted from a `DeprecationWarning` to a `FutureWarning`. This is intentional as `Cluster.logs` is a user-facing method we want users to move away from.

Issue #4637, opened 2021-03-26
Repeatedly use the same worker on first task


mrocklin:
Apparently in a quiet cluster we end up sending new tasks to the same worker repeatedly.  So probably the following test would fail:

```python
@gen_cluster(client=True)
def test_round_robin(c, s, a, b):
    await c.submit(inc, 1)
    await c.submit(inc, 2)
    await c.submit(inc, 3)
    assert a.transition_log and b.transition_log
```

This outcome is determined here:

https://github.com/dask/distributed/blob/bf9ddab3fd67c33b0d87da4f18bcdc7e47262393/distributed/scheduler.py#L2122-L2136

It looks like when there are no dependencies and we have only a few workers we currently choose the worker with minimum occupancy.  In a quiet cluster all workers have zero occupancy, so probably we're getting whatever Python uses to break a tie in this setting.

In the case where the occupancy is zero we might do something like a round-robin (this is done just below in the case where we have greater than 20 workers) among the worker pool for as long as that worker has zero occupancy.

Reported in conversation by @crusaderky

douglasdavis:
Tinkering with this. The test indeed fails ;) but perhaps the reason is not what you were trying to trigger? I'm seeing `transition_log` is not an attribute of `distributed.worker.Worker`. Searching through the repo I only see instances of `s.transition_log` scattered in a bunch of tests.

Is your `assert` trying to test that both workers `a` and `b` have some evidence of usage? I might be barking up the wrong tree.

mrocklin:
> Is your assert trying to test that both workers a and b have some evidence of usage?

Yes, this was exactly my intent.  The issue was authored in haste, my apologies for the messiness here.  Looking again at worker.py it looks like we might look at `Worker.log` instead.

douglasdavis:
Thanks for the clarification!

PR #4925, opened 2021-06-17
Consider candidates that don't hold any dependencies in `decide_worker`

gjoseph92:
- [ ] Closes #xxxx
- [ ] Tests added / passed


distributed/tests/test_scheduler.py
```diff
@@ -126,6 +126,24 @@ async def test_decide_worker_with_restrictions(client, s, a, b, c):
...
+    root = await client.scatter(1)
+    assert sum(root.key in worker.data for worker in [a, b, c]) == 1
+
+    start = time()
+    tasks = client.map(sleep, [root] * 6, pure=False)
+    await wait(tasks)
+    elapsed = time() - start
+
+    assert elapsed <= 4
+    assert all(root.key in worker.data for worker in [a, b, c])
```
mrocklin, line 144: Three seconds sounds long for a test.  Maybe we can make this function `inc`, and instead check that `len(a.data) == len(b.data) == len(c.data) == 2` ?

distributed/scheduler.py
```diff
@@ -7495,16 +7495,28 @@ def decide_worker(
...
         candidates = set(all_workers)
     else:
-        candidates = {wws for dts in deps for wws in dts._who_has}
+        if len(random_workers_set) <= N_RANDOM_WORKERS:
+            candidates = random_workers_set
+        else:
+            candidates = {wws for dts in deps for wws in dts._who_has}
+            # TODO the ordering of this set is likely not actually random.
+            # Ideally it would be ordered by occupancy (but that's more work than we want to do).
+            # Can we at least ensure idle workers come first? And that it's reasonably random after that?
```
mrocklin, line 7518: `Scheduler.idle` and `Scheduler.workers` are both SortedDicts, and so should support random access.  You may need to stop using the `workers_dv` attribute, which was brought in for cythonization.  Let's just drop that optimization here, keep things simple, and keep using the sorted dicts.  Then you should be able to do random selections.

gjoseph92: What are they sorted by?

distributed/scheduler.py
```diff
@@ -7495,16 +7495,28 @@ def decide_worker(
...
     candidates: set
     assert all([dts._who_has for dts in deps])
     if ts._actor:
         candidates = set(all_workers)
     else:
-        candidates = {wws for dts in deps for wws in dts._who_has}
+        if len(random_workers_set) <= N_RANDOM_WORKERS:
+            candidates = random_workers_set
+        else:
+            candidates = {wws for dts in deps for wws in dts._who_has}
```
mrocklin, line 7515: I'm not sure I understand this.  I would have expected something like the following:

```python
if ts._actor:
    ...
else:
    candidates = {wws for dts in deps for wws in dts._who_has}
    candidates |= random_sample(valid_workers or all_workers)
```

Also, as a warning, any time we start getting nested if/elses I'll probably become slightly more critical :)

distributed/scheduler.py
mrocklin: Also, I would first try idle, and then if nothing is there try workers. 

It may be that keeping this second decide_workers out of the Scheduler class isn't worth it.  We shouldn't let the Cythonization effort stop us from making wise scheduling decisions.

distributed/scheduler.py
gjoseph92: This is an optimization for when there are less than 20 workers, meaning our random sample of workers would actually contain all workers. If we're going to be adding all workers into the candidates set anyway, then candidates will end up containing all workers at the end, so we can skip building candidates from `_who_has`.

distributed/scheduler.py
mrocklin: > What are they sorted by?

Just their name.  Really though it doesn't matter what they're sorted by.  What matters is that these collections are indexable, and so it's possible to do a proper random sample from them in short time.  

You may also want to see how `Scheduler.decide_worker` handles idle workers.

```python
            worker_pool = self._idle or self._workers
            worker_pool_dv = cast(dict, worker_pool)
            wp_vals = worker_pool.values()
            n_workers: Py_ssize_t = len(worker_pool_dv)
            if n_workers < 20:  # smart but linear in small case
                ws = min(wp_vals, key=operator.attrgetter("occupancy"))
                if ws._occupancy == 0:
                    # special case to use round-robin; linear search
                    # for next worker with zero occupancy (or just
                    # land back where we started).
                    wp_i: WorkerState
                    start: Py_ssize_t = self._n_tasks % n_workers
                    i: Py_ssize_t
                    for i in range(n_workers):
                        wp_i = wp_vals[(i + start) % n_workers]
                        if wp_i._occupancy == 0:
                            ws = wp_i
                            break
            else:  # dumb but fast in large case
                ws = wp_vals[self._n_tasks % n_workers]
```

And again, I'm more than happy to walk back the cythonization dict_values / `_dv` stuff if it gets in the way.

gjoseph92: I was just asking in case what they were sorted by was random enough (or correlated to occupancy in some way) that just iterating through them would be a reasonable way to take the sample. Name is too stable, so I'll use that indexing logic.

mrocklin: Yeah, we don't update the sorted collection on every task update.  Live sorting on occupancy would probably be non-viable.  Idle/Saturated is the best we have, and even that ends up taking up non-trivial CPU resources.

mrocklin: But generally I think that the decide_worker function probably needs to have a worker pool of `self._idle or self._workers`.  Short term we can probably pass that in as yet-another-parameter.

distributed/tests/test_scheduler.py
```diff
@@ -126,6 +126,22 @@ async def test_decide_worker_with_restrictions(client, s, a, b, c):
...
+)
+async def test_decide_worker_select_candidate_holding_no_deps(client, s, a, b, c):
+    root = await client.scatter(1)
+    assert sum(root.key in worker.data for worker in [a, b, c]) == 1
+
+    tasks = client.map(inc, [root] * 6, pure=False)
+    await wait(tasks)
+
+    assert all(root.key in worker.data for worker in [a, b, c])
+    assert len(a.data) == len(b.data) == len(c.data) == 3
```
mrocklin, line 142: Actually, now that I look at this again, maybe we do need the sleep.

If we're scheduling `inc`s then it's not unreasonable to schedule six of these things on one worker.  That's where the data is.   This computation is very cheap.  But in contrast, the dependency is very small.  We may need the scheduler to learn that `slowinc` is slow, and so heavy relative to the `int` dependency.  

Maybe 

```python
await client.submit(slowinc, 10, delay=0.1)  # learn that slowinc is slow
root = await client.scatter(1)

futures = client.submit(slowinc, [root] * 6, delay=0.1, pure=False)
await wait(futures)
```

This way we're confident that the computational cost of the future will far outweigh the communciation cost, but things are still fast-ish.

gjoseph92: Yeah, that's why I originally went with `sleep`. It also _feels_ more like the case we're trying to test for. Though the test still passes with 6d91816 (and note that that test fails on main), so I'm not sure what we're thinking about wrong?

mrocklin: Currently 28 bytes / 100MB/s is just still smaller than however fast inc ran in (a few microseconds probably). It would be safer to bump this up to milliseconds though.  

gen_cluster tests can run in less than a second, but more than 100ms.  So sleeps of around that 100ms time are more welcome than 1s sleeps.

gjoseph92:
This is causing `test_task_groups` to fail:
```python
    @gen_cluster(client=True, Worker=NoSchedulerDelayWorker)
    async def test_task_groups(c, s, a, b):
        start = time()
        da = pytest.importorskip("dask.array")
        x = da.arange(100, chunks=(20,))
        y = (x + 1).persist(optimize_graph=False)
        y = await y
        stop = time()
    
        tg = s.task_groups[x.name]
        tp = s.task_prefixes["arange"]
        repr(tg)
        repr(tp)
        assert tg.states["memory"] == 0
        assert tg.states["released"] == 5
        assert tp.states["memory"] == 0
        assert tp.states["released"] == 5
        assert tp.groups == [tg]
        assert tg.prefix is tp
        # these must be true since in this simple case there is a 1to1 mapping
        # between prefix and group
        assert tg.duration == tp.duration
        assert tg.nbytes_in_memory == tp.nbytes_in_memory
        assert tg.nbytes_total == tp.nbytes_total
        # It should map down to individual tasks
        assert tg.nbytes_total == sum(
            [ts.get_nbytes() for ts in s.tasks.values() if ts.group is tg]
        )
        in_memory_ts = sum(
            [
                ts.get_nbytes()
                for ts in s.tasks.values()
                if ts.group is tg and ts.state == "memory"
            ]
        )
>       assert tg.nbytes_in_memory == in_memory_ts
E       assert -480 == 0
E         +-480
E         -0
```
I'm wondering if this is exposing an existing bug where we aren't counting `nbytes_in_memory` correctly when a task is replicated on multiple workers. [Here](https://github.com/gjoseph92/distributed/blob/646b12be37ef82e2afa2baf2827716746fa93173/distributed/scheduler.py#L2650), the value is decremented for all workers when the key is released, but it feels like it's not being correspondingly incremented when the key gets duplicated. Stepping through the test, I haven't seen the `diff` in `set_nbytes` be 0, so maybe `set_nbytes` just isn't getting called when a dependency is transferred from one worker to another?

gjoseph92:
Good news though—this helps a lot with #4471.

1) It runs in 2:20 vs 3:46 on my machine
2) Kinda hard to tell in these videos (60x sped up), but you can see the tasks per worker graph stays unbalanced on `main`, but evens out on this branch. Also (when watching this at normal speed) I could tell that the `sleep(10)` tasks clearly were  more evenly distributed across workers.

#### `main`:
main

#### This branch:


cc @chrisroat

Also just for my own sake, here's the incantation I used to make the GIFs from screen recordings:
```bash
ffmpeg -itsscale 0.01666 -i main.mov -filter_complex "[0:v] palettegen" palette.png
ffmpeg -itsscale 0.01666 -i main.mov -i palette.png -filter_complex "[0:v] fps=24,scale=600:-1 [i]; [i][1:v] paletteuse" -r 24 main.gif
```

mrocklin:
cc'ing @fjetter who I think will find this interesting

distributed/scheduler.py
```diff
@@ -7476,6 +7482,9 @@ def decide_worker(
     of bytes sent between workers.  This is determined by calling the
     *objective* function.
     """
+    # TODO should it be a bounded fraction of `len(all_workers)`?
+    N_RANDOM_WORKERS: Py_ssize_t = 20
```
mrocklin, line 7486: Can we move this into the `else` block scope?  Mostly I'm optimizing to not distract readers until necessary, or to bound the distraction.

distributed/scheduler.py
```diff
@@ -7486,6 +7495,17 @@ def decide_worker(
         candidates = set(all_workers)
     else:
         candidates = {wws for dts in deps for wws in dts._who_has}
+        # Add some random workers to into `candidates`, starting with idle ones
+        # TODO shuffle to prevent hotspots?
+        candidates.update(idle_workers[:N_RANDOM_WORKERS])
```
mrocklin, line 7500: We *may* want to randomly select here, I'm not sure.  It probably doesn't matter that much because we'll naturally cycle the chosen ones out of the group.

mrocklin: We wouldn't shuffle, to be clear, we would index.

Also, maybe 5 or 10 workers is enough?

distributed/tests/test_scheduler.py
```diff
@@ -100,14 +100,16 @@ async def test_recompute_released_results(c, s, a, b):
     assert result == 1
 
 
-@gen_cluster(client=True)
+@gen_cluster(client=True, config={"distributed.scheduler.bandwidth": "1mb"})
```
mrocklin, line 103:
```suggestion
@gen_cluster(client=True, config={"distributed.scheduler.bandwidth": "1Mb"})
```

distributed/scheduler.py
```diff
@@ -7486,6 +7495,17 @@ def decide_worker(
...
+        if len(idle_workers) < N_RANDOM_WORKERS:
+            sample_from = (
+                list(valid_workers) if valid_workers is not None else all_workers
+            )
```
mrocklin, line 7504: Would it make sense to lump `idle_workers` into here instead of having the line above?

Maybe something like this would get the job done?

```python
sample_from = list(valid_workers) if valid_workers else (idle_workers or all_workers)
if len(sample_from) > 10:
    candidates |= random.choices(10, sample_from)
else:
    candidates |= sample_from
```

mrocklin: Oops, forgot to add the `.values`, which makes the sortedcollection.SortedDicts indexable.

```python
sample_from = list(valid_workers) if valid_workers else (idle_workers.values() or all_workers.values())
if len(sample_from) > 10:
    candidates |= random.choices(10, sample_from)
else:
    candidates |= sample_from
```

distributed/scheduler.py
gjoseph92: > It probably doesn't matter that much because we'll naturally cycle the chosen ones out of the group.

That's what I figured. I was more thinking about a regression on #4637 for the "ish" tasks of root-ish that wouldn't go down the current round-robin code path, since they do have dependencies.

distributed/tests/test_scheduler.py
gjoseph92: The test is passing as-is; do you still think I should change it back to `sleep`?

distributed/scheduler.py
gjoseph92: I had wanted to be sure that if there was 1 idle worker, we'd still mix in some other non-idle workers. But perhaps that's pointless, since the current objective function would never choose a non-idle worker with 0 deps over an idle worker with 0 deps, right?

distributed/scheduler.py
gjoseph92: Done

distributed/tests/test_scheduler.py
mrocklin: It would become a problem if the `inc` ran in `28 bytes / 100MB/s = 0.2us` or less.  Inc does actually run this fast.

```python
In [1]: 28 / 100e6
Out[1]: 2.8e-07

In [2]: def inc(x):
   ...:     return x + 1
   ...: 

In [3]: %timeit inc(1)
64.2 ns ± 0.914 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

But presumably the infrastructure around the inc is taking longer.  If our timing infrsatructure got much faster then this might result in intermittent errors.  It *probably* doesn't matter at this scale, but it's probably a good idea if it takes only a few minutes.

We might also introduce a penalty of something like 1ms for *any* communication (I'm surprised that we don't do this today actually) which might tip the scales in the future.

distributed/scheduler.py
mrocklin: Yeah, probably.  I also prefer the version I posted just because there are fewer moving pieces.  Moving pieces scare the crap out of me this deep in the code :)

distributed/tests/test_scheduler.py
gjoseph92: I'll switch back to a short sleep for now.

> We might also introduce a penalty of something like 1ms for _any_ communication

That seems very worth doing.

mrocklin: We maybe used to do this?  It might be worth checking logs.  The `worker_objective` function is pretty clean if you wanted to add this.  We should probably wait until we find a case where it comes up though in an impactful way just to keep from cluttering things.

distributed/scheduler.py
gjoseph92: What would you think of:
```python
# Select up to 10 idle or random workers
sample_from = (
    list(valid_workers)
    if valid_workers is not None
    else idle_workers or all_workers
)
candidates = set(random.sample(sample_from, min(10, len(sample_from))))
# If that didn't cover all valid workers, also add all workers holding deps of this task
if len(sample_from) < len(valid_workers if valid_workers is not None else all_workers):
    candidates = {wws for dts in deps for wws in dts._who_has}
```
which keeps the (maybe pointless) optimization of skipping the `who_has` when all workers are already accounted for, plus gets us the idle shuffle, without many more moving pieces.

> Oops, forgot to add the `.values`

We're actually already passing those in: b050d14d465e9ffaeb75258ac582f94ee92e9402

Also I'm curious why you're using `choices` instead of `sample`. Though we are making a set so it won't affect anything, picking the same random worker multiple times doesn't seem that helpful?

distributed/tests/test_scheduler.py
gjoseph92: Yeah I think it should be a separate PR for cleanliness

distributed/scheduler.py
mrocklin: > Also I'm curious why you're using choices instead of sample. Though we are making a set so it won't affect anything, picking the same random worker multiple times doesn't seem that helpful?

I just forgot the difference.  sample sounds great.

> which keeps the (maybe pointless) optimization of skipping the who_has when all workers are already accounted for

Yeah, currently I like the simpler one until it becomes a performance concern.  I've been cut badly by small performance optimization tricks like this having unintended consequences and being very hard to track down.  If it's ok I'd prefer to do the simplest thing possible and then revisit this code and optimize once it shows up in profiles.

I think I get pretty particular about code simplicity within the state-machine logic.  This code had hundreds of opportunities to grow in complexity.  If we had taken them it would be, today, incomprehensible.  I think that we really want the logic in here to be very easy to reason about very quickly.  I'm looking to minimize code elements here.

gjoseph92: Makes sense. Since these optimizations would only be active with <10 workers anyway, the numbers are small enough to not worry about yet.

I'm going with:
```python
        # Select all workers holding deps of this task
        candidates = {wws for dts in deps for wws in dts._who_has}
        # Add up to 10 random workers into `candidates`, preferring idle ones.
        sample_from = (
            list(valid_workers)
            if valid_workers is not None
            else idle_workers or all_workers
        )
        candidates.update(random.sample(sample_from, min(10, len(sample_from))))
```

distributed/scheduler.py
```diff
@@ -7470,7 +7470,7 @@ def decide_worker(
     Decide which worker should take task *ts*.
 
     We consider all workers which hold dependencies of *ts*,
-    plus a sample of 20 random workers (with preference for idle ones).
+    plus a sample of up to 20 random workers (with preference for idle ones).
```
mrocklin, line 7473:
```suggestion
    plus a sample of up to 10 random workers (with preference for idle ones).
```

gjoseph92: thanks!

distributed/scheduler.py
```diff
@@ -7491,19 +7491,16 @@ def decide_worker(
...
-                # ^ NOTE: `min` because `random.sample` errors if `len(sample) < k`
-            )
+        # Add up to 10 random workers into `candidates`, preferring idle ones.
+        sample_from = (
+            list(valid_workers)
+            if valid_workers is not None
+            else idle_workers or all_workers
+        )
+        candidates.update(random.sample(sample_from, min(10, len(sample_from))))
+        # ^ NOTE: `min` because `random.sample` errors if `len(sample) < k`
```
mrocklin, line 7503:
```python
import random
import sortedcontainers

d = sortedcontainers.SortedDict()
for i in range(100):
    d[str(i)] = i

candidates = set()

%timeit candidates.update(random.sample(d.values(), min(10, len(d))))
```
```
12.4 µs ± 652 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

You're not wrong to be concerned.  12us is a decent chunk of time.  

It looks like `random.sample` isn't cheap

```python
%timeit random.sample(d.values(), min(10, len(d)))
# 11.7 µs ± 61.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit d.values()[:10]
# 2.09 µs ± 27.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

mrocklin: This might help with efficiency while also up with simplicity?

```python
        candidates = {wws for dts in deps for wws in dts._who_has}

        extra = (
            list(valid_workers)
            if valid_workers is not None
            else (idle_workers or all_workers)
        )
        if len(extra) > 10:
            extra = random.sample(extra, 10)
        
        candidates.update(extra)
```

jakirkham: Some time is spent just working with `sortedcontainers`.

```python
import random

d = dict()
for i in range(100):
    d[str(i)] = i

candidates = set()

%timeit candidates.update(random.sample(list(d.values()), min(10, len(d))))
```

```
7.21 µs ± 31.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

As a side note it seems `random.sample` won't take a `dict` or its iterators directly. So this includes coercing to a `list`

gjoseph92:
```python
l = list(d.values())

%timeit candidates.update(random.sample(l, min(10, len(l))))
# 5.09 µs ± 179 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit l = list(d.values()[:10])
1.98 µs ± 34.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit l[:10]
# 79.3 ns ± 2.2 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

SortedValuesView [does say](https://github.com/grantjenks/python-sortedcontainers/blob/master/sortedcontainers/sorteddict.py#L781) lookups are O(log(n)), but it seems like it's more than that.

mrocklin: It takes any indexable, which `sortedcontainers.SortedDict.values` provides.  Indexability is the reason why we use sortedcontainers.

jakirkham: Right the point is that results in it taking ~2x longer than just copying to `list` and feeding it in

mrocklin: Oh interesting

mrocklin:
```python
        candidates = {wws for dts in deps for wws in dts._who_has}

        if valid_workers is not None:
            extra = valid_workers
        elif idle_workers:
            extra = idle_workers[:10]
        else:
            extra = all_workers

        if len(extra) > 10:
            extra = random.sample(extra, 10)
        
        candidates.update(extra)
```

This might be getting too deep into the weeds though.

gjoseph92: For comparison:
```python
from distributed.scheduler import TaskState, WorkerState

N = 5
extra = 2
workers = [WorkerState(str(i)) for i in range(N)]
deps = [TaskState(str(i), None) for i in range(N)]
for dts, ws in zip(deps, workers):
    dts.who_has.add(ws)

for i in range(extra):
    deps[i].who_has.add(workers[-i])
deps = set(deps)

%timeit candidates = {wws for dts in deps for wws in dts._who_has}
971 ns ± 24 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
```
N = 50
extra = 50
12 µs ± 202 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

N = 100
extra = 10
17.2 µs ± 214 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

So just to be clear that the random sample is the thing to try to avoid, not the `who_has` selection. In the uncommon case of many many dependencies duplicated over many workers, that selection is still on par with the random sample.

gjoseph92: So sampling the keys and then using them to look up values is 2x faster with SortedDict, and the same as what we'd get with a plain dict:
```python
%timeit candidates.update(d[k] for k in random.sample(list(d), 10))
6.95 µs ± 227 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

pd = dict(d)
%timeit candidates.update(pd[k] for k in random.sample(list(pd), 10))
6.36 µs ± 95.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Also `random.choices` is 2x faster than `random.sample` (makes sense since it doesn't have to hold onto any state):
```python
%timeit candidates.update(d[k] for k in random.choices(list(d), k=10))
3.65 µs ± 11.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit candidates.update(pd[k] for k in random.choices(list(pd), k=10))
3.17 µs ± 36.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

I'd feel okay about `random.choices`; we'd get slightly fewer extra workers in the pool, but it's probably worth the time savings.

gjoseph92: If we're okay with duplicates then we can go much faster.

Iterating values of a SortedDict is slow:
```python
%timeit list(d.values())
8.43 µs ± 85.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

But it's fast on a plain dict:
```python
pd = dict(d)
%timeit list(pd.values())
691 ns ± 5.07 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
So we just have to make sure we're passing in `self._workers_dv.values()`, not `self._workers.values()`.

This is decent:
```python
%timeit candidates.update(random.choices(list(pd.values()), k=10))
2.45 µs ± 15.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

But we can go much faster and avoid a copy with an iterator:

```python
import itertools
import random

def random_choices_iter(xs: "Iterator", k: int) -> "Iterator":
    return itertools.islice(itertools.takewhile(lambda _: random.random() < 0.5, xs), k)

%timeit candidates.update(random_choices_iter(pd.values(), k=10))
643 ns ± 12.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

gjoseph92:
@gforsyth @fjetter I happened to notice that this PR makes `Worker.waiting_for_data_count` go negative in some tests. I'd guess the logic for that count doesn't account for replicated keys, and it's getting double-decremented (much like #4927). Like in #4927, `waiting_for_data_count` doesn't seem to be used for any decisions, only for the worker dashboard. Should we remove it, or is there a simple way to keep this count accurate?

```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

client = <Client: No scheduler connected>, s = <Scheduler: "tcp://127.0.0.1:52603" workers: 0 cores: 0, tasks: 0>
a = <Worker: 'tcp://127.0.0.1:52604', 0, Status.closed, stored: 0, running: 0/1, ready: 0, comm: 0, waiting: 0>
b = <Worker: 'tcp://127.0.0.1:52606', 1, Status.closed, stored: 0, running: 0/1, ready: 0, comm: 0, waiting: -1>
c = <Worker: 'tcp://127.0.0.1:52608', 2, Status.closed, stored: 0, running: 0/1, ready: 0, comm: 0, waiting: -2>

    @gen_cluster(
        client=True,
        nthreads=[("127.0.0.1", 1)] * 3,
        config={"distributed.scheduler.work-stealing": False},
    )
    async def test_decide_worker_select_candidate_holding_no_deps(client, s, a, b, c):
        await client.submit(slowinc, 10, delay=0.1)  # learn that slowinc is slow
        root = await client.scatter(1)
        assert sum(root.key in worker.data for worker in [a, b, c]) == 1
    
        start = time()
        tasks = client.map(slowinc, [root] * 6, delay=0.1, pure=False)
        await wait(tasks)
        elapsed = time() - start
    
        assert elapsed <= 4
        assert all(root.key in worker.data for worker in [a, b, c]), [
            list(worker.data.keys()) for worker in [a, b, c]
        ]
>       assert (
            a.waiting_for_data_count
            == b.waiting_for_data_count
            == c.waiting_for_data_count
            == 0
        )
E       assert 0 == -1
E         +0
E         --1

distributed/tests/test_scheduler.py:150: AssertionError
```

distributed/scheduler.py
jakirkham: It looks `random.sample` will take `d.keys()` from playing around with it locally. So we can probably skip the coercion to a `list` in that case. Though maybe we've already moved well beyond this at this point

I'd probably shy away from getting to clever around random item selection. It's probably more likely to have surprising behavior that we only discover later. Just my 2 cents 🙂

distributed/scheduler.py
```diff
@@ -2324,7 +2324,10 @@ def decide_worker(self, ts: TaskState) -> WorkerState:
         if ts._dependencies or valid_workers is not None:
             ws = decide_worker(
                 ts,
-                self._workers_dv.values(),
+                self._workers_dv.values() if compiled else dict.values(self._workers),
+                self._idle_dv.values() if compiled else dict.values(self._workers),
+                # ^ NOTE: For performance, these must be actual `dict_values`, not `SortedDictValues`.
+                # In Cython, `_workers_dv` is a plain dict, but in plain Python, it's still a `SortedDict`.
```
jakirkham, line 2330: Is it bad to always coerce? How much does that cost us?

gjoseph92: I suppose they're actually exactly equivalent statements, so I think it costs us nothing (unless Cython interprets it differently?). We could just do `dict.values(self._workers_dv)`. Just a bit hacky. Feels more like a note to self that if we ever had Cython be the default / distributed compiled wheels, we could maybe remove the conditional.

distributed/scheduler.py
gjoseph92: I think we're beyond that, since we're only taking `dict_values` objects in this function now. Making it able to support both dicts and sets was a bit of a mess and not worth it IMO.

I didn't go with the too-clever itertools option because it did have surprising behavior. I did have to add another conditional though because `choices` can repeat items: fcb165e. So perhaps even that's too clever?

distributed/scheduler.py
jakirkham: Yeah I don't see a real alternative. Cython can `cast` whereas Python can't. So we are just handling that inherent awkwardness

I suppose we could just use an ordinary `dict` instead of a `SortedDict` if that were acceptable

gjoseph92:
I think this is ready for final review. Note that it does need #4937 to be merged first though.

The tests failing right now are a flaky one (#4957) and `test_access_semaphore_by_name`, which I've never seen fail before on this PR, so I think it's flaky too.

distributed/scheduler.py
```diff
@@ -7485,7 +7495,17 @@ def decide_worker(
     if ts._actor:
         candidates = set(all_workers)
     else:
+        # Select all workers holding deps of this task
         candidates = {wws for dts in deps for wws in dts._who_has}
+        # Add up to 10 random workers into `candidates`, preferring idle ones.
+        worker_pool = valid_workers if valid_workers is not None else all_workers
+        if len(candidates) < len(worker_pool):
+            sample_from = idle_workers or worker_pool
```
mrocklin, line 7503: What if `valid_workers and idle_workers`?

mrocklin: It seems like maybe we want to use `idle_workers or all_workers` above in line 7501

distributed/scheduler.py
```diff
@@ -7485,7 +7495,17 @@ def decide_worker(
...
+            candidates.update(
+                random.choices(list(sample_from), k=min(10, len(sample_from)))
+                if len(sample_from) > 10
+                else sample_from
+            )
     if valid_workers is None:
         if not candidates:
             candidates = set(all_workers)
```
mrocklin, line 7511: Now that we have idle_workers in this function should this be `idle_workers or all_workers`?

distributed/scheduler.py
```diff
@@ -7485,7 +7495,17 @@ def decide_worker(
...
         candidates = set(all_workers)
     else:
+        # Select all workers holding deps of this task
         candidates = {wws for dts in deps for wws in dts._who_has}
+        # Add up to 10 random workers into `candidates`, preferring idle ones.
+        worker_pool = valid_workers if valid_workers is not None else all_workers
+        if len(candidates) < len(worker_pool):
+            sample_from = idle_workers or worker_pool
+            candidates.update(
+                random.choices(list(sample_from), k=min(10, len(sample_from)))
```
mrocklin, line 7505: Thoughts on reducing 10 to 2 or 5 or something lower?  This might be premature, but I'm curious how much adding more workers helps here in aggregate.  Part of me thinks that adding just one random worker into the mix probably does 80% of the good in aggregate.

gjoseph92: I'd kind of like to do that later once we've gotten more of a sense of the impact of this, and hopefully have some more examples to measure that change against. When a cluster is mostly quiet, picking 2 random workers has a high chance of finding a good one, but in a large cluster with mostly busy workers, it's a big haystack to find the few that are underutilized (but not idle).

gjoseph92:
I'm leaning away from merging this currently. As it stands, this makes all the tests for https://github.com/dask/distributed/pull/4967 fail. I think this is an indicator of what it would do in real-life scenarios too, though I haven't tested it.

We've just done this work to ensure that future tasks run on workers that minimize data transfer; this PR now breaks our assumptions by randomizing the scheduling of the future tasks.

Basically, this _would_ work if our `worker_objective` function took a more holistic view of scheduling. But it's too short-sighted currently. I'd want to at least:

1. Consider how the placement of this task would affect the placement of its dependents, and how much data would have to move in the future for the dependents to also run on the given worker. (This can be expensive to calculate.)
1. Consider the cost of data transfer on the sender. Serialization is not free. If the worker holding the dependencies is busy, it may take a longer time to transfer them to a new idle worker.

There might be some clever, cheap heuristic for whether a future task will be a shoe-in to schedule on one particular worker, or whether it'll be fair game to run anywhere, but I haven't been able to come up with one.

----

Since https://github.com/dask/distributed/issues/4471 was originally about new workers joining the cluster, we could also consider restricting the random workers just to idle ones? Note though that with that change, both the tests for #4967 and this PR fail, so it may not be a very useful change.

mrocklin:
> There might be some clever, cheap heuristic for whether a future task will be a shoe-in to schedule on one particular worker, or whether it'll be fair game to run anywhere, but I haven't been able to come up with one

So, let's think about this for a moment.

My read of this issue is that it seems like we like the idea in principle, but that in cases where we know there is some better-than-greedy placement of tasks (such as in rootish tasks) that this greedy approach is likely to make things worse.  Is that a correct read?

If so, it seems like maybe we should go ahead with this, but block doing this on any rootish tasks.  Today rootish tasks are a (very useful) anomaly in our scheduling.  We should be ok working around them in anomalous ways I think.  

So, I'd be inclined to add a flag to TaskGroups about being rootish, and check that flag when making this decision.  Thoughts?

gjoseph92:
The problem wasn't about the root-ish tasks. It was about the _dependents_ of root-ish tasks. We might place the root tasks perfectly, but if the downstream tasks are then free to select from any random worker (instead of just workers that hold their input data), it's quite likely they'll disregard our plans and just run on whichever worker is the least busy, even if that means transferring all the data.

So really, the problem is that our `worker_objective` turns out to not be a very good objective, especially when the set of candidates is unrestricted. It's heavily biased towards low occupancy, and has low penalties for data transfer.

gjoseph92:
I'll add that though I also find this problem very interesting, and I'd love to work on it, I don't think it's what we should focus on right now. Getting lower-level stability under control is more important.

Between this, https://github.com/dask/distributed/pull/5325, and https://github.com/dask/distributed/pull/5326, we both spent a bit of time thinking about these problems last summer. We found that they were more challenging and subtle than we expected.

I don't think we'll find a quick tweak to solve this in a couple days that doesn't introduce unintended consequences. We should probably save these delicate scheduling changes for when we can dedicate more time to focusing carefully on them.

mrocklin:
> I don't think we'll find a quick tweak to solve this in a couple days

I remain, as always, more optimistic than you here :) 

> I'll add that though I also find this problem very interesting, and I'd love to work on it, I don't think it's what we should focus on right now. Getting lower-level stability under control is more important.

Fair point.  Shifting my focus over to benchmarking.

PR #4928, opened 2021-06-18
Dont start http server in Scheduler.__init__

fjetter:
This should deal with the CI port already in use warning situation reported in #4806 

Highlights
* We have a few places where we start up a cluster (e.g. spec cluster or gen_cluster) where we do not close the scheduler if workers fail to start. This would leave ports still allocated
* Previously HTTP server sockets were opened during `Scheduler.__init__` which made some tests really awkward. This PR moves the start of the HTTP server to the `Scheduler.start` method.
* A few tests intentionally start two clusters/schedulers where I circumvented this warning by simply allowing random ports. This revealed a shortcoming of our API since there is currently no way to *not* start an HTTP server
* I added a filterwarnings statement in our setup.cfg to err whenever this warning is raised in our test setup. This is very strict but I would prefer keeping it and removing it again if it causes any issues. Preferably we'd have a test fixture which checks if the test left any open socks but I'm not aware of any code which can do this without root

Supersedes https://github.com/dask/distributed/pull/4896 and https://github.com/dask/distributed/pull/4921

- [x] Closes https://github.com/dask/distributed/issues/4806
- [x] Tests added / passed

### DIFF
--- a/distributed/core.py
+++ b/distributed/core.py
@@ -426,7 +426,7 @@ class Server:
 
         logger.debug("Connection from %r to %s", address, type(self).__name__)
         self._comms[comm] = op
-        await self
+
         try:
             while True:
                 try:
--- a/distributed/deploy/utils_test.py
+++ b/distributed/deploy/utils_test.py
@@ -25,10 +25,14 @@ class ClusterTest:
         assert future.result() == 2
 
     def test_context_manager(self):
-        with self.Cluster(**self.kwargs) as c:
+        kwargs = self.kwargs.copy()
+        kwargs.pop("dashboard_address")
+        with self.Cluster(dashboard_address=":54321", **kwargs) as c:
             with Client(c) as e:
                 assert e.nthreads()
 
     def test_no_workers(self):
-        with self.Cluster(0, scheduler_port=0, **self.kwargs):
+        kwargs = self.kwargs.copy()
+        kwargs.pop("dashboard_address")
+        with self.Cluster(0, dashboard_address=":54321", scheduler_port=0, **kwargs):
             pass
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -84,6 +84,13 @@ from .utils_perf import disable_gc_diagnosis, enable_gc_diagnosis
 from .variable import VariableExtension
 
 try:
+    import bokeh  # noqa: F401
+
+    HAS_BOKEH = True
+except ImportError:
+    HAS_BOKEH = False
+
+try:
     from cython import compiled
 except ImportError:
     compiled = False
@@ -3438,24 +3445,11 @@ class Scheduler(SchedulerState, ServerNode):
             default_port=self.default_port,
         )
 
-        http_server_modules = dask.config.get("distributed.scheduler.http.routes")
-        show_dashboard = dashboard or (dashboard is None and dashboard_address)
-        missing_bokeh = False
-        # install vanilla route if show_dashboard but bokeh is not installed
-        if show_dashboard:
-            try:
-                import distributed.dashboard.scheduler
-            except ImportError:
-                missing_bokeh = True
-                http_server_modules.append("distributed.http.scheduler.missing_bokeh")
-        routes = get_handlers(
-            server=self, modules=http_server_modules, prefix=http_prefix
+        self._show_dashboard: bool = dashboard or (
+            dashboard is None and dashboard_address
         )
-        self.start_http_server(routes, dashboard_address, default_port=8787)
-        if show_dashboard and not missing_bokeh:
-            distributed.dashboard.scheduler.connect(
-                self.http_application, self.http_server, self, prefix=http_prefix
-            )
+        self._dashboard_address = dashboard_address
+        self._http_prefix = http_prefix
 
         # Communication state
         self.loop = loop or IOLoop.current()
@@ -3758,6 +3752,23 @@ class Scheduler(SchedulerState, ServerNode):
         enable_gc_diagnosis()
 
         self.clear_task_state()
+
+        http_server_modules = dask.config.get("distributed.scheduler.http.routes")
+        assert isinstance(http_server_modules, list)
+
+        # install vanilla route if show_dashboard but bokeh is not installed
+        if self._show_dashboard and not HAS_BOKEH:
+            http_server_modules.append("distributed.http.scheduler.missing_bokeh")
+        routes = get_handlers(
+            server=self, modules=http_server_modules, prefix=self._http_prefix
+        )
+        self.start_http_server(routes, self._dashboard_address, default_port=8787)
+        if self._show_dashboard and HAS_BOKEH:
+            import distributed.dashboard.scheduler
+
+            distributed.dashboard.scheduler.connect(
+                self.http_application, self.http_server, self, prefix=self._http_prefix
+            )
 
         with suppress(AttributeError):
             for c in self._worker_coroutines:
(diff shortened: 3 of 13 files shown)


distributed/core.py
```diff
@@ -422,7 +422,7 @@ async def handle_comm(self, comm):
 
         logger.debug("Connection from %r to %s", address, type(self).__name__)
         self._comms[comm] = op
-        await self
```
fjetter, line 425: This is a very interesting change. I encountered race condition where this `await` would actually trigger the warning after I moved the HTTP server startup code to `Scheduler.start`. Apparently the start is not idempotent and sometimes we'd restart / start a scheduler twice but only under some strange timing sensitive conditions (I believe the handler was already active while the scheduler is still starting...). I can dig deeper here but I figured there shouldn't be a reason to `await self` here since this handler should only be registered after/during the start anyhow.

mrocklin: My _hope_ would be that `await self` would be idempotent.  If in the future we find that there is an easy way to make this true I would be in support.

fjetter: I'd assume a `if status==running; return` would already do the trick. If we prefer this route, I can revert the `await self` thingy here

distributed/scheduler.py
```diff
@@ -5393,7 +5405,7 @@ async def gather(self, comm=None, keys=None, serializers=None):
     def clear_task_state(self):
         # XXX what about nested state such as ClientState.wants_what
         # (see also fire-and-forget...)
-        logger.info("Clear task state")
+        logger.debug("Clear task state")
```
fjetter, line 5408: I changed this log level since it felt kind of wrong. This prints a `Clear task state` to the console whenever I call `dask-scheduler`. That's currently failing the tests. I assume this is OK?

distributed/core.py
mrocklin: That, combined with using an async lock to avoid many `await self` calls at the same time.  I *think* that we do this in other situations, like Client and Cluster.  I'm surprised that we don't do it in Server.

fjetter: The idempotence you are referring to is implemented once for `Server`, see https://github.com/dask/distributed/blob/d9bc3c634a4a3983cdc75e5c39cbdac2afbf6184/distributed/core.py#L264-L269

for a reason I haven't understood, yet, this still caused issues for me. my suspicion is that the scheduler already arrived in a non running but not properly closed state when I see this message. This would then try to revive a kind of dead scheduler and cause this warning. I'll let CI run on this a few times and see if I can reproduce.

fjetter: I remembered running into this await before. I had a similar problem over in https://github.com/dask/distributed/pull/4734

fjetter: Indeed, it turns out a properly timed connection request to a dead or dying scheduler can revive it. hrhr, I guess this is merely an edge case relevant to our async test suite. Regardless, imho this is much more cleanly fixed in #4734 and I suggest to merge that one before.

distributed/tests/test_utils_test.py
```diff
@@ -225,3 +227,27 @@ def test_tls_cluster(tls_client):
 async def test_tls_scheduler(security, cleanup):
     async with Scheduler(security=security, host="localhost") as s:
         assert s.address.startswith("tls")
+
+
+from distributed.core import Status
```
jrbourbeau, line 232: Nitpick: Can you move this import to the top of the module with existing `distributed.core` import?

distributed/deploy/utils_test.py
```diff
@@ -25,10 +25,14 @@ def test_submit(self):
         assert future.result() == 2
 
     def test_context_manager(self):
-        with self.Cluster(**self.kwargs) as c:
+        kwargs = self.kwargs.copy()
+        kwargs.pop("dashboard_address")
```
jrbourbeau, line 29: This would silently ignore any `dashboard_address` which has been specified in a test. Since this is only ever used in one place, perhaps we should just specify `dashboard_address=":0"` there instead? This isn't worth spending too much time on, it was just slightly surprising to see us manually setting `dashboard_address=":54321"`

distributed/deploy/utils_test.py
```diff
@@ -25,10 +25,14 @@ def test_submit(self):
...
+        with self.Cluster(dashboard_address=":54321", **kwargs) as c:
             with Client(c) as e:
                 assert e.nthreads()
 
     def test_no_workers(self):
-        with self.Cluster(0, scheduler_port=0, **self.kwargs):
+        kwargs = self.kwargs.copy()
+        kwargs.pop("dashboard_address")
```
jrbourbeau, line 36: Similar comment here

distributed/tests/test_utils_test.py
fjetter: sure, I'm somehow used for either linters or formatters to make me aware of this. for some reason our config considers this perfectly fine 🤷‍♂️

distributed/deploy/utils_test.py
fjetter: yes, I think I changed this before I thought about the idea to simply use a random port which ignores the warning. I'll fix that

distributed/scheduler.py
```diff
@@ -85,6 +85,13 @@
 from .utils_perf import disable_gc_diagnosis, enable_gc_diagnosis
 from .variable import VariableExtension
 
+try:
+    import bokeh  # noqa: F401
+
+    HAS_BOKEH = True
+except ImportError:
+    HAS_BOKEH = False
```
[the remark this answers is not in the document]
fjetter: If we do this in other places already, I'll change it. I think keeping a consistent style in a codebase is worth the effort

setup.cfg
```diff
@@ -45,6 +45,9 @@ parentdir_prefix = distributed-
 addopts = -v -rsxfE --durations=20
 filterwarnings =
     error:Since distributed.*:PendingDeprecationWarning
+
+    # See https://github.com/dask/distributed/issues/4806
+    error:Port:UserWarning:distributed.node
```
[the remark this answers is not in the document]
fjetter: This _might_ be a bit too optimistic but let's wait and see. It may definitely interfere if the code base is run suing `pytest-xdist`. We will always get port collisions when running in parallel

mrocklin:
I'm curious, what is the status here?

fjetter:
> I'm curious, what is the status here?

Status here is a bit awkward. I needed/wanted to remove the `await` in the `handle_comm` which feels unmotivated and fragile in this PR. If I keep it, sometimes the scheduler restarts due to a race condition while closing and triggers the warning. I hit the same issue over in https://github.com/dask/distributed/pull/4734 where, imho, I implemented this much safer with a proper status check in `__await__` so I wanted to merge #4734 first. However, over there I'm hitting a lot of the flaky tests quite reliably and I cannot get the build green-ish.

Options:

1. We leave the `await` in handle_comm and remove the strict filterwarnings. We occasionally will get the port warning but much more rarely. Occasionally means that some CI runs won't have any while others might have thousands. 
2. I try to cherry-pick the code around this await onto this PR or a dedicated smaller PR to see if I can make this work and then we can merge this. That would be #4734 without the shielding stuff (prob a good idea either way)
3. I will continue to dig in #4734 since I don't believe it is a coincidence that it triggers our flaky tests more likely.

I think today I will try 2. and if that fails we'll go for 1. since having no warnings in most of the runs is still a win and we can follow up once #4734 is properly resolved

distributed/core.py
fjetter: Note after I became a bit smarter: This await is here to ensure that the server is actually up. problem is that the start is not idempotent, stumbled over this in https://github.com/dask/distributed/pull/4734 as well

PR #4929, opened 2021-06-18
Fix deadlock in handle_missing_dep if additional replicas are available

fjetter:
Yet another deadlock (the for-loop was not tested, my coverage PR is also landing shortly)

What happens is that the missing dependency management does currently only work properly if there are *no* replicas of the failed fetch are available. otherwise it can happen that the fetch is never rescheduled (add dependent to self.data_needed)

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/tests/test_failed_workers.py
+++ b/distributed/tests/test_failed_workers.py
@@ -620,3 +620,51 @@ async def test_forget_data_not_supposed_to_have(s, a, b):
     assert a.data
     while a.data:
         await asyncio.sleep(0.001)
+
+
+@gen_cluster(
+    client=True,
+    nthreads=[("127.0.0.1", 1) for _ in range(3)],
+    config={"distributed.comm.timeouts.connect": "1s"},
+    Worker=Nanny,
+)
+async def test_failing_worker_with_additional_replicas_on_cluster(c, s, *workers):
+    """
+    If a worker detects a missing dependency, the scheduler is notified. If no
+    other replica is available, the dependency is rescheduled. A reschedule
+    typically causes a lot of state to be reset. However, if another replica is
+    available, we'll need to ensure that the worker can detect outdated state
+    and correct its state.
+    """
+
+    def slow_transfer(x, delay=0.1):
+        return SlowTransmitData(x, delay=delay)
+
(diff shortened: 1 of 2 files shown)


distributed/tests/test_failed_workers.py
```diff
@@ -620,3 +620,51 @@ async def test_forget_data_not_supposed_to_have(s, a, b):
...
+    # We'll schedule tasks on two workers, s.t. f1 is replicated. We will
+    # suspend one of the workers and kill the origin worker of f1 such that a
+    # comm failure causes the worker to handle a missing dependency. It will ask
+    # the schedule such that it knows that a replica is available on f2 and
+    # reschedules the fetch
+    f2 = c.submit(dummy, f1, pure=False, key="f2", workers=[workers[1].worker_address])
+    f3 = c.submit(dummy, f1, pure=False, key="f3", workers=[workers[2].worker_address])
+
+    await wait(f1)
+    proc.suspend()
```
fjetter, line 664: I couldn't think of any other way to provoke this race condition than actually suspending the process. This is incredibly timing sensitive. Happy to investigate further angles if smbd has an idea

distributed/worker.py
```diff
@@ -2532,7 +2532,7 @@ async def handle_missing_dep(self, *deps, **kwargs):
                     logger.debug("New workers found for %s", dep.key)
                     self.log.append((dep.key, "new workers found"))
                     for dependent in dep.dependents:
-                        if dependent.key in dep.waiting_for_data:
+                        if dep.key in dependent.waiting_for_data:
```
fjetter, line 2535: I think I will plainly reject PRs which suggest using `dep` as a variable name in the future 🤔 

BTW this `if` clause in general seems a bit weird. there is no actual reason why one should not reschedule the task. however, I am not entirely sure but I believe this is here to protect `Worker.data_needed` from overflowing. We should consider replacing this with an ordered-set-like thingy (e.g. a dict) or some other data structure. Ideally we'd like to have a priority sorted set-like iterator here but well... that's a problem for another day

jrbourbeau (approved):
Thanks @fjetter!

Issue #4927, opened 2021-06-18
`TaskGroup.nbytes_in_memory` miscounted for replicated keys


gjoseph92:
I think there is a logic error with bookkeping for `TaskGroup.nbytes_in_memory`. There's a discrepancy between how we increment it and decrement it when multiple workers hold the same key.

In `transition_memory_released`, we decrement it by `nbytes` once for every worker that holds that task: https://github.com/dask/distributed/blob/6340b5b34a582ff1ffe3eaf9dd7f3aac33d0a983/distributed/scheduler.py#L2646-L2649
Whereas in `_propagate_forgotten`, we decrement it once by `nbytes` if there are any workers holding the task, regardless of how many. This doesn't match with `transition_memory_released`:
https://github.com/dask/distributed/blob/646b12be37ef82e2afa2baf2827716746fa93173/distributed/scheduler.py#L7339-L7341

On the creation side, in `TaskState.set_nbytes`, we only increment it by the _diff_ between the last known value and the current value. If the key is being copied to multiple workers, this difference is usually 0: https://github.com/dask/distributed/blob/6340b5b34a582ff1ffe3eaf9dd7f3aac33d0a983/distributed/scheduler.py#L1556-L1562

In short, I think `TaskGroup.nbytes_in_memory` is incremented once per key, but decremented once per _copy_ of the key.

If `nbytes` can be different for different workers, then to do this bookkeeping correctly, I think we'd also need to track `TaskState.total_nbytes` (size of all copies of the key), then decrement by that once in `transition_memory_released` and `_propagate_forgotten`.

Discovered in https://github.com/dask/distributed/pull/4925#issuecomment-863600812. I think #4925 made this more apparent, since it encourages more data replication.

cc @crusaderky since you know more about replicated keys.

fjetter:
I stumbled over this myself recently, see https://github.com/dask/distributed/blob/fced981adf6b8d95900cc3e089237201251f803f/distributed/tests/test_scheduler.py#L1919-L1920 where this behaviour is intentionally pinned but I believe this was by mistake. Intuitively, I would also expect this to be different.

From what I can see, the lack of counting is introduced in `Scheduler.add_keys` where workers let the scheduler know once they have a replica. There, the group nbytes is not increased
https://github.com/dask/distributed/blob/fced981adf6b8d95900cc3e089237201251f803f/distributed/scheduler.py#L6300-L6303 

I noticed this in my deadlock PR which already grew without bounds so I didn't fix it.

[...]

PR #4930, opened 2021-06-18
Remove nbytes_in_memory

mrocklin:
This doesn't appear to be necessary for any scheduling decisions today, and is causing issues.  Let's remove it.

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -873,11 +873,6 @@ class TaskPrefix:
         )
 
     @property
-    def nbytes_in_memory(self):
-        tg: TaskGroup
-        return sum([tg._nbytes_in_memory for tg in self._groups])
-
-    @property
     def nbytes_total(self):
         tg: TaskGroup
         return sum([tg._nbytes_total for tg in self._groups])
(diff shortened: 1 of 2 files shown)


mrocklin:
git blame puts the blame on me on this one.  This was added as part of the original TaskGroup/TaskPrefix PR, which did carry the sin of speculative work.  My apologies to all.

gjoseph92:
I was thinking of a similar solution 😄

What about `nbytes_total`? I'm not sure what that's supposed to mean, but it might have similar problems.

I _could_ see `nbytes_in_memory` become useful in the TaskGroup visualization, but if we actually want it we can add it back for that.

mrocklin:
I recommend that we wait on merging this PR until after the release today, just in case someone was depending on this internal state.

mrocklin:
In the meantime, cherry-picking this PR might be a good idea :)

jakirkham:
Also happy to see this go away. I don't think it was used anywhere

jakirkham:
Out of curiosity what issues was it causing?

jrbourbeau:
This happened to pop up and cause issues in a couple of places over the past few days. https://github.com/dask/distributed/pull/4925#issuecomment-863600812 is one of those places. https://github.com/dask/distributed/issues/4927 has a nice writeup from @gjoseph92

gjoseph92:
https://github.com/dask/distributed/issues/4927, https://github.com/dask/distributed/issues/4927#issuecomment-863876698, https://github.com/dask/distributed/pull/4925#issuecomment-863600812. Oddly, after a long period of dormancy, a few of us all ran into this from different angles at the same time.

jakirkham:
Ok thanks all 😄

PR #4932, opened 2021-06-18
Add option for timestamps from output of `Node.get_logs`

charlesbluca:
As discussed in #4909, it would be nice if the output of `Node.get_logs` had optional timestamps, which could be used to filter the logs in the performance report and/or be added as information for the user.

- [ ] Tests added / passed


distributed/node.py
```diff
@@ -87,13 +87,17 @@ def _setup_logging(self, logger):
         logger.addHandler(self._deque_handler)
         weakref.finalize(self, logger.removeHandler, self._deque_handler)
 
-    def get_logs(self, comm=None, n=None):
+    def get_logs(self, comm=None, n=None, timestamps=False):
```
fjetter, line 90: It would be great if you can start off with a doc string for this. Start simple with the changes you contributed. We can add on to it later on.

distributed/dashboard/components/scheduler.py
```diff
@@ -2164,10 +2164,19 @@ def update(self):
 
 
 class SchedulerLogs:
-    def __init__(self, scheduler):
-        logs = Logs(scheduler.get_logs())._repr_html_()
+    def __init__(self, scheduler, start=None):
+        logs = scheduler.get_logs(timestamps=True)
+        if start is not None:
+            logs = [log for log in logs if log[0] > start]
```
fjetter, line 2170: We could, of course, also add a start/end kwarg to the get_logs and do the filtering in one pass.

fjetter (commented):
CI failures are introduced since the performance report tests assert very specifically how the logs look like. You might want to relax the asserts over there.

charlesbluca:
Are there any other user-facing places where we might want to add timestamps for logs? Currently this only impacts the perf report and Jupyter notebook representation.

charlesbluca:
Will probably want to wait on #4941 before we merge this

charlesbluca:
Could we rerun tests on this? Should be good now that #4941 is merged

distributed/node.py
```diff
@@ -87,13 +87,35 @@ def _setup_logging(self, logger):
...
             L = deque_handler.deque
-            L = [L[-i] for i in range(min(n, len(L)))]
+            L = [msg for msg in L if msg.created > start]
+        if n is not None:
+            L = [L[-i - 1] for i in range(min(n, len(L)))]
+        if timestamps:
+            return [
+                (msg.created, msg.levelname, deque_handler.format(msg)) for msg in L
+            ]
         return [(msg.levelname, deque_handler.format(msg)) for msg in L]
```
fjetter, line 119: For performance, I would suggest to do this all in one go. how about

```python
if start is None:
    start = -1
L = []
for count, msg in enumerate(deque_handler.deque):
    if n and count >= n or msg.created < start:
        break
    if timestamps:
        L.append((msg.created, msg.levelname, deque_handler.format(msg)))
    else:
        L.append((msg.levelname, deque_handler.format(msg)))
return L
```

This way we can do everything in one pass instead of three. by not using a list comprehension we can also short-circuit the `start`since the logs should be always sorted. (haven't tested this)

fjetter (commented):
LGTM. Please let me know if you want to try my suggestion or whether we shall merge

distributed/node.py
charlesbluca: This looks good - I will try this out

charlesbluca:
Thanks for the suggestion @fjetter, think this should be good to merge now

fjetter (approved):
Thanks @charlesbluca and sorry for the long wait

PR #4933, opened 2021-06-18
Print detail in `validate_*` assertions

gjoseph92:
When validation failed for whatever reason, I found myself wanting this (since pytest may not alway pretty-print these errors, if they happen to be caught by a try-except elsewhere in the scheduler for example).

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -4782,69 +4782,77 @@ class Scheduler(SchedulerState, ServerNode):
         parent: SchedulerState = cast(SchedulerState, self)
         ts: TaskState = parent._tasks[key]
         dts: TaskState
-        assert ts._state == "released"
-        assert not ts._waiters
-        assert not ts._waiting_on
-        assert not ts._who_has
-        assert not ts._processing_on
-        assert not any([ts in dts._waiters for dts in ts._dependencies])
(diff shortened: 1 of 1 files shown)


distributed/scheduler.py
```diff
@@ -4782,69 +4782,77 @@ def validate_released(self, key):
...
-        assert not any([ts in dts._waiters for dts in ts._dependencies])
-        assert ts not in parent._unrunnable
+        assert ts._state == "released", ts._state
+        assert not ts._waiters, ts._waiters
+        assert not ts._waiting_on, ts._waiting_on
+        assert not ts._who_has, ts._who_has
+        assert not ts._processing_on, ts._processing_on
+        assert not any(ts in dts._waiters for dts in ts._dependencies), [
+            dts._waiters for dts in ts._dependencies
+        ]
```
jrbourbeau, line 4792: Should we also print `ts` (or `ts.key`) so we can parse the `ts in dts._waiters` part of this `assert`? 

I have similar questions for all the other `assert X in Y` / `assert X not in Y` statements here

jrbourbeau (commented):
Thanks @gjoseph92! Overall this seems like a nice addition

PR #4934, opened 2021-06-18
Add transition counter to Scheduler

mrocklin:
This helps us identify if things have changed recently

- [ ] Closes #xxxx
- [ ] Tests added / passed

### DIFF
--- a/distributed/scheduler.py
+++ b/distributed/scheduler.py
@@ -1770,6 +1770,7 @@ class SchedulerState:
     _validate: bint
     _workers: object
     _workers_dv: dict
+    _transition_counter: Py_ssize_t
 
     # Variables from dask.config, cached by __init__ for performance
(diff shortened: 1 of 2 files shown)


fjetter:
> identify if things have changed recently

IIUC, we'd want to detect regressions with this counter? Does it make sense to add this to some kind of dashboard, report, etc. or is this merely intended for internal debugging?

mrocklin:
My goal is to use it in dashboard plots to see if we have done any work recently ...

```python
while True:
    if self._old_transition_counter == self.scheduler._transition_counter:
        return
    else:
        ... do a bunch of work because things have changed
        self._old_transition_counter = self.scheduler._transition_counter

    await asyncio.sleep(...)
```

jrbourbeau (approved):
Thanks @mrocklin. Looking forward to using this in dashboard plots

PR #4942, opened 2021-06-21
Disable worker dashboard by default in tests

gjoseph92:
Messages like
```
  /home/runner/work/distributed/distributed/distributed/node.py:160: UserWarning: Port 8787 is already in use.
  Hosting the HTTP server on port 36717 instead
```
take a a _lot_ of lines in CI. Running the worker dashboard during tests seems unnecessary?

### DIFF
--- a/distributed/utils_test.py
+++ b/distributed/utils_test.py
@@ -870,7 +870,8 @@ def gen_cluster(
         nthreads = ncores
 
     worker_kwargs = merge(
-        {"memory_limit": system.MEMORY_LIMIT, "death_timeout": 10}, worker_kwargs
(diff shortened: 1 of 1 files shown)


fjetter:
This is not connected to worker dashboads since worker dashboards use random ports. See https://github.com/dask/distributed/issues/4806 and 
https://github.com/dask/distributed/pull/4928 for a fix (hopefully)

Also, even with dashboard=False we are actually starting an HTTP server. There is currently no way to not start an HTTP server

gjoseph92:
Thanks, yes this is not actually the problem.

